PR #416, opened 2023-11-24
Redesigned stream/topic recipient headers

sirpengi:
Design from the following two sources:
- https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=538%3A20849&mode=dev
- https://github.com/zulip/zulip-mobile/issues/5511

This PR updates the stream/topic recipient headers to the new design and a PR to #373 will follow this up very shortly. The style of dates were updated, but the format of the text itself is still left for #411 

This fixes #372 and #220.
I made an attempt to tackle #282 using combinations of `Row`/`Expanded`/`Flexible` but wasn't able to succeed. I suspect we will need to implement some custom layout code to handle that. 

Simulator Screenshot - iPhone 14 - 2023-11-24 at 18 16 28

Simulator Screenshot - iPhone 14 - 2023-11-24 at 18 19 59

PR #426, opened 2023-11-29
Redesigned DM recipient headers

sirpengi:
This is built on top of #416.

This PR updates the DM recipient headers to the new style. Text style and spacing was taken from the same sources as #416:

- https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=538%3A20849&mode=dev
- https://github.com/zulip/zulip-mobile/issues/5511

But given there were no mockups that included a DM recipient header the style used here also looked to how web renders these headers (in particular the color to use for the background, and icon to use for display).

dm_header

Here showing behavior when the system requests a larger font:
exploded_font

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -194,6 +194,21 @@
       "filename": {"type": "String", "example": "file.txt"}
     }
   },
+  "unknownUserName": "(unknown user)",
+  "@unknownUserName": {
+    "description": "Name placeholder to use for a user when we don't know their name."
+  },
+  "messageListGroupYouAndOthers": "You and {others}",
+  "@messageListGroupYouAndOthers": {
+    "description": "Message list recipient header for a DM group with others.",
+    "placeholders": {
+      "others": {"type": "String", "example": "Alice, Bob"}
+    }
+  },
+  "messageListGroupYouWithYourself": "You with yourself",
+  "@messageListGroupYouWithYourself": {
+    "description": "Message list recipient header for a DM group that only includes yourself."
+  },
   "contentValidationErrorTooLong": "Message length shouldn't be greater than 10000 characters.",
   "@contentValidationErrorTooLong": {
     "description": "Content validation error message when the message is too long."
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -680,12 +680,16 @@ class DmRecipientHeader extends StatelessWidget {
             child: Text(style: const TextStyle(color: Colors.white),
               title)),
           RecipientHeaderDate(message: message,
-            color: _kRecipientHeaderDateColor),
+            color: _kDmRecipientHeaderDateColor),
         ])));
   }
 }
 
-final _kDmRecipientHeaderColor = const HSLColor.fromAHSL(1, 0, 0, 0.27).toColor();
+// TODO(#95): web uses different color in dark mode
+// Header color from `color-background-private-message-header`
+//   in zulip:web/styles/zulip.css .
+final _kDmRecipientHeaderColor = const HSLColor.fromAHSL(1, 46, 0.35, 0.93).toColor();
+final _kDmRecipientHeaderDateColor = Colors.black.withOpacity(0.4);
 
 class RecipientHeaderDate extends StatelessWidget {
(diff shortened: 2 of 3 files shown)


sirpengi:
@gnprice rebased on main and ready for review!

lib/widgets/message_list.dart
```diff
@@ -511,147 +534,202 @@ class _UnreadMarker extends StatelessWidget {
...
-      title = 'You and $otherNames';
+        .join(", "));
     } else {
-      title = 'You with yourself'; // TODO pick string; web has glitchy "You and $yourname"
+      // TODO pick string; web has glitchy "You and $yourname"
+      title = zulipLocalizations.messageListGroupYouWithYourself;
     }
 
     return GestureDetector(
-      behavior: HitTestBehavior.translucent,
```
gnprice, line 589: msglist: Remove HitTestBehavior.translucent on DmRecipientHeader
    
    This seems unintended as the header should capture all
    clicks within its bounds.

Capturing all clicks within its bounds is the reason this was `translucent`. Here's that value's doc:
```dart
  /// Translucent targets both receive events within their bounds and permit
  /// targets visually behind them to also receive events.
  translucent,
```

So with that value, the header does capture all taps within its bounds.

The default is `deferToChild`. That one's doc says:
```dart
  /// Targets that defer to their children receive events within their bounds
  /// only if one of their children is hit by the hit test.
  deferToChild,
```

So with that default, the header would *not* necessarily capture all taps within its bounds — it only does so if its child is hit.

This parameter is unnecessary now, though, so it's good to remove it. The reason it's unnecessary is that the child fills the whole box.

Looking at the history with `git log -L` (`git log -L 668,693:lib/widgets/message_list.dart` as of the tip of your branch, to cover this whole `return` statement), I believe the key change was 412233c9cdd87062bb63dba2a7d6da85cdddf440. Before that, and when this line was added in f7feabf773a02d4b34ada9956889d55e96f285f8 as part of making the DM recipient header touchable at all, the child was an `Align` and so it only occupied the left-hand portion of the parent. That made `HitTestBehavior.translucent` necessary in order for touching everywhere in the header to work.

assets/l10n/app_en.arb
```diff
@@ -194,6 +194,21 @@
       "filename": {"type": "String", "example": "file.txt"}
     }
   },
+  "messageListGroupUnknownUser": "(unknown user)",
+  "@messageListGroupUnknownUser": {
+    "description": "Name placeholder to use for user when we don't know their name."
+  },
+  "messageListGroupYouAndOthers": "You and {others}",
+  "@messageListGroupYouAndOthers": {
+    "description": "Message list sticky header for a dm group with others.",
```
gnprice, line 203:
```suggestion
    "description": "Message list recipient header for a DM group with others.",
```

"DM" is an initialism, so in normal English text (as opposed to identifiers in code) it's all caps.

Also although these headers are sticky, I think "recipient header" works better as a name for them because it points more to what they're about.

assets/l10n/app_en.arb
```diff
@@ -194,6 +194,21 @@
       "filename": {"type": "String", "example": "file.txt"}
     }
   },
+  "messageListGroupUnknownUser": "(unknown user)",
```
gnprice, line 197:
```suggestion
  "unknownUserName": "(unknown user)",
```

I think it makes sense to use the same form for this pretty consistently across the app — seems helpful for reducing confusion about whether it literally means an unknown user or the name some actual user has picked. So we can make the identifier more general.

assets/l10n/app_en.arb
```diff
@@ -194,6 +194,21 @@
...
+  "@messageListGroupUnknownUser": {
+    "description": "Name placeholder to use for user when we don't know their name."
```
gnprice, line 199:
```suggestion
    "description": "Name placeholder to use for a user when we don't know their name."
```

The headline-ese "user" without an article sounds like it might mean "*the* user", i.e. the self-user.

gnprice (commented):
Thanks @sirpengi! Generally this looks good — a few small comments.

sirpengi:

gnprice:

Issue #377, opened 2023-11-11
msglist: Mark-as-read button takes a long time (batch-processing already-read messages)


chrisbobbe:
In `#test here` on CZO, if I let there be just one unread message, then I tap the recently added "Mark 1 message as read" button (#364), it regularly takes 4 to 5 seconds for that single message to get marked as read.

It looks like the server is doing some work on (maybe just accessing) a large number of messages that are already read. Here is some console output, where you'll see response data from the mark-as-read requests, alternating with the corresponding mark-as-read events:

<details>

```
flutter: POST https://chat.zulip.org/api/v1/messages/flags/narrow
flutter: {result: success, msg: , processed_count: 1000, updated_count: 0, first_processed_id: 6592, last_processed_id: 111506, found_oldest: true, found_newest: false}
flutter: POST https://chat.zulip.org/api/v1/messages/flags/narrow
flutter: {type: update_message_flags, op: add, operation: add, flag: read, messages: [], all: false, id: 349}
flutter: server event: update_message_flags/add read
flutter: GET https://chat.zulip.org/api/v1/events?queue_id=50c5a217-6f4b-45ce-bcfa-3888e2b3ba88&last_event_id=349
flutter: {result: success, msg: , processed_count: 1000, updated_count: 0, first_processed_id: 111507, last_processed_id: 140521, found_oldest: false, found_newest: false}
flutter: POST https://chat.zulip.org/api/v1/messages/flags/narrow
flutter: {type: update_message_flags, op: add, operation: add, flag: read, messages: [], all: false, id: 350}
flutter: server event: update_message_flags/add read
flutter: GET https://chat.zulip.org/api/v1/events?queue_id=50c5a217-6f4b-45ce-bcfa-3888e2b3ba88&last_event_id=350
flutter: {result: success, msg: , processed_count: 1000, updated_count: 0, first_processed_id: 140522, last_processed_id: 195418, found_oldest: false, found_newest: false}
flutter: POST https://chat.zulip.org/api/v1/messages/flags/narrow
flutter: {type: update_message_flags, op: add, operation: add, flag: read, messages: [], all: false, id: 351}
flutter: server event: update_message_flags/add read
flutter: GET https://chat.zulip.org/api/v1/events?queue_id=50c5a217-6f4b-45ce-bcfa-3888e2b3ba88&last_event_id=351
[...]

PR #427, opened 2023-11-29
Unconditionally add ApiNarrowIsUnread to markNarrowAsRead

sirpengi:
Fixes: #377

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -824,13 +824,13 @@ Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
   int responseCount = 0;
   int updatedCount = 0;
 
-  final apiNarrow = switch (narrow) {
-    // Since there's a database index on is:unread, it's a fast
-    // search query and thus worth using as an optimization
-    // when processing all messages.
-    AllMessagesNarrow() => [ApiNarrowIsUnread()],
-    _                   => narrow.apiEncode(),
-  };
+  // Include `is:unread` in the narrow.  That has a database index, so
+  // this can be an important optimization in narrows with a lot of history.
+  // The server applies the same optimization within the (deprecated)
+  // specialized endpoints for marking messages as read; see
+  // `do_mark_stream_messages_as_read` in `zulip:zerver/actions/message_flags.py`.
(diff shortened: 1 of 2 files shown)


lib/widgets/message_list.dart
```diff
@@ -734,13 +734,17 @@ Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
...
-  };
+  // This process can take a long time in narrows with a lot
+  // of history given the server ends up processing a large
+  // number of already read messages. Unconditionally add
+  // [ApiNarrowIsUnread] here to take advantage of a database
+  // index on `is:unread`.
+  // This optimization is also used server side for the
+  // (deprecated) specialized endpoints for marking messages
+  // as read. See `do_mark_stream_messages_as_read` in
+  // `zulip:zerver/actions/message_flags.py`.
```
gnprice, line 745:
```suggestion
  // Include `is:unread` in the narrow.  That has a database index, so
  // this can be an important optimization in narrows with a lot of history.
  // The server applies the same optimization within the (deprecated)
  // specialized endpoints for marking messages as read; see
  // `do_mark_stream_messages_as_read` in `zulip:zerver/actions/message_flags.py`.
```

I think the main thing I wanted to adjust in this comment is that the other version, with "This process can take a long time" and "the server ends up processing", sounds like it's describing what *does* happen, whereas in fact it's describing what *would* happen if the code worked in a different way.

A direct fix for that could look like "If we used `narrow.apiEncode()` directly, this process could …", or like "This process could … given the server would end up processing … messages. Avoid that by adding …".

(A related, smaller, point is that although "unconditionally" is a highly relevant term for the commit message, it's a bit out of place in the code comment. The reader of the code comment is trying to understand the code as it is, and not starting from how the code previously was, so they're left to wonder what the hypothetical other way of doing this was that would have been conditional.)

gnprice (commented):
Thanks @sirpengi! This all looks good modulo some revisions to the explanatory code comment, below.

I'll go ahead and merge with that change.

gnprice:
(Oh also I tweaked the commit summary line: "Unconditionally add ApiNarrowIsUnread *in* markNarrowAsRead", rather than "to". To my mind it's being added to the ApiNarrow, or to the request the ApiNarrow is part of, but the function markNarrowAsRead is doing the adding.)

PR #429, opened 2023-11-30
Recipient header date format

### LINKED ISSUE #411: Date formats for recipient headers
Currently the recipient headers always show the date in full, like "2023-11-22". We should instead use the same formats we have in web:
* "Today" and "Yesterday" when accurate;
* Otherwise month and day, or year, month, and day, with certain logic for deciding between those.

For the exact logic, see the (oddly-named) [`render_now` function](https://github.com/zulip/zulip/blob/411d0d7863421276904d3dfa247b2bfce4e9ca6b/web/src/timerender.ts#L158-L193) in `web/src/timerender.ts` in Zulip web.

The web implementation ultimately relies on the JS `Intl.DateTimeFormat` class. We should be able to get equivalent results using Dart's [`DateFormat` from `package:intl`](https://pub.dev/documentation/intl/latest/intl/DateFormat-class.html). As a bonus that will also take care of internationalizing these date formats, part of #278.

sirpengi:
This PR is built on top of #426 

Fixes #411

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -710,6 +710,7 @@ class RecipientHeaderDate extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
     return Padding(
       padding: const EdgeInsets.fromLTRB(10, 0, 16, 0),
       child: Text(
@@ -725,12 +726,48 @@ class RecipientHeaderDate extends StatelessWidget {
           //   https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps#all-small-caps
           fontFeatures: const [FontFeature.enable('c2sc'), FontFeature.enable('smcp')],
         ).merge(weightVariableTextStyle(context)),
-        _kRecipientHeaderDateFormat.format(
-          DateTime.fromMillisecondsSinceEpoch(message.timestamp * 1000))));
+        formatHeaderDate(
+          zulipLocalizations,
+          DateTime.fromMillisecondsSinceEpoch(message.timestamp * 1000),
+          now: DateTime.now())));
   }
 }
 
-final _kRecipientHeaderDateFormat = DateFormat('y-MM-dd', 'en_US'); // TODO(#278)
+@visibleForTesting
+String formatHeaderDate(
+  ZulipLocalizations zulipLocalizations,
+  DateTime dateTime, {
+  required DateTime now,
+}) {
+  assert(!dateTime.isUtc && !now.isUtc,
+    '`dateTime` and `now` need to be in local time.');
+
+  if (dateTime.year == now.year &&
+      dateTime.month == now.month &&
+      dateTime.day == now.day) {
+    return zulipLocalizations.today;
+  }
+
+  final yesterday = now
+    .copyWith(hour: 12, minute: 0, second: 0, millisecond: 0, microsecond: 0)
+    .add(const Duration(days: -1));
+  if (dateTime.year == yesterday.year &&
+      dateTime.month == yesterday.month &&
+      dateTime.day == yesterday.day) {
+    return zulipLocalizations.yesterday;
+  }
+
+  // If it is Dec 1 and you see a label that says `Dec 2`
+  // it could be misinterpreted as Dec 2 of the previous
+  // year. For times in the future, those still on the
+  // current day will show as today (handled above) and
+  // any dates beyond that show up with the year.
+  if (dateTime.year == now.year && dateTime.isBefore(now)) {
+    return DateFormat.MMMd().format(dateTime);
+  } else {
+    return DateFormat.yMMMd().format(dateTime);
+  }
+}
 
 /// A Zulip message, showing the sender's name and avatar if specified.
 class MessageWithPossibleSender extends StatelessWidget {
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -366,7 +366,7 @@ void main() {
     testWidgets('show dates', (tester) async {
       await setupMessageListPage(tester, messages: [
         eg.streamMessage(timestamp: 1671409088),
-        eg.dmMessage(timestamp: 1692755322, from: eg.selfUser, to: []),
+        eg.dmMessage(timestamp: 1661219322, from: eg.selfUser, to: []),
       ]);
       // We show the dates in the user's timezone.  Dart's standard library
       // doesn't give us a way to control which timezone is used — only to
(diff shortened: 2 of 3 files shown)


lib/widgets/message_list.dart
```diff
@@ -710,35 +725,39 @@ class RecipientHeaderDate extends StatelessWidget {
...
+      dateTime.month == now.month &&
+      dateTime.day == now.day) {
+    return "Today";
+  }
 
-  final Color color;
-  final Widget child;
+  final yesterday = now
+    .copyWith(hour: 12, minute: 0, second: 0, millisecond: 0, microsecond: 0)
+    .add(const Duration(days: -1));
```
gnprice, line 750: I believe I've heard of a locale where this wouldn't work — IIRC in parts of Ethiopia, the days run from midnight to midnight but the hours are counted from noon to noon.

But I'm not certain how to write correct logic for that: partly because I'm not sure I remember the details right, partly because I'm not sure how other software handles it, including `DateTime.copyWith`, and partly because it's just complicated to think about. So it wouldn't make sense to try to handle that case without writing tests for it. Since we don't have a reasonable way to control the timezone (as commented on that existing test below), I think it's best to just ignore that case at present.

gnprice (commented):
Thanks @sirpengi! Generally this looks good. Comments below, mainly on the tests.

test/widgets/message_list_test.dart
```diff
@@ -371,11 +377,32 @@ void main() {
...
+    final testCases = [
+      [DateTime(2023, 1, 10, 12),     zulipLocalizations.temporalNounToday],
+      [DateTime(2023, 1, 10, 0),      zulipLocalizations.temporalNounToday],
+      [DateTime(2023, 1, 10, 23, 59), zulipLocalizations.temporalNounToday],
+      [DateTime(2023, 1, 9, 23, 59),  zulipLocalizations.temporalNounYesterday],
+      [DateTime(2023, 1, 9, 0),       zulipLocalizations.temporalNounYesterday],
+      [DateTime(2023, 1, 8),          "Jan 8"],
+      [DateTime(2022, 12, 31),        "Dec 31, 2022"],
+      // Times in the future should show up as a normal date.
+      [DateTime(2023, 1, 11, 12),     "Jan 11"],
```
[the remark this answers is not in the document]
gnprice: > … Oh I see. This actually matches the logic we _currently_ have in web. What happened is that we accidentally altered that behavior in a commit that was described as just replacing a dependency, [zulip/zulip@9896782](https://github.com/zulip/zulip/commit/9896782fd14de1e3b573cb232762988af0e1fe6b) . This condition had matched the comment, but no longer does:
> 
> ```diff
> -    } else if (is_older_year) {
> +    } else if (time.getFullYear() !== today.getFullYear()) {
>          // For long running servers, searching backlog can get ambiguous
>          // without a year stamp. Only show year if message is from an older year
> ```

FYI @andersk as the author of that change. Though the root issue here is that we apparently never had a test for that case.

lib/widgets/message_list.dart
```diff
@@ -710,35 +725,39 @@ class RecipientHeaderDate extends StatelessWidget {
           //   https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps#all-small-caps
           fontFeatures: const [FontFeature.enable('c2sc'), FontFeature.enable('smcp')],
         ).merge(weightVariableTextStyle(context)),
-        _kRecipientHeaderDateFormat.format(
-          DateTime.fromMillisecondsSinceEpoch(message.timestamp * 1000))));
+        formatHeaderDate(
+          zulipLocalizations,
+          DateTime.fromMillisecondsSinceEpoch(message.timestamp * 1000,
+            isUtc: false),
```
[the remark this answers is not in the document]
sirpengi: I was overgeneralizing the exception we have for tests, as here we need to ensure the times being passed in are both in local-time. I thought to be explicit here about that but I see the point about the defaults being part of the api. I'll remove the redundant default but add a comment in the code reminding us along with some asserts.

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
sirpengi: I dug into this a bit this morning. It looks like `format_time_modern` actually hasn't been used since it was reverted back in March 2022 in 7bc0e70693f49ea8faef2578158362491fe5470c . See https://chat.zulip.org/#narrow/stream/137-feedback/topic/recent.20topics.20timestamps/near/1337670 for background.

Web is currently using `relative_time_string_from_date` for this label (the date for last message time in recent topics).

```js
export function relative_time_string_from_date({
    date,
    current_date = new Date(),
}: {
    date: Date;
    current_date?: Date; 
}): string {
    const minutes = differenceInMinutes(current_date, date);
    if (minutes <= 2) {
        return $t({defaultMessage: "Just now"});
    }
```
and sadly it looks like this function will always return `Just now` for any times from the future. I can also confirm by changing my computer clock to the previous day and every `time` in my inbox shows `Just now`.

I think showing times in the future is weird. Even for `format_time_modern` it shows the date label if the future time even a minute ahead, despite being on the same day (so it could be Dec 1 and the label will say `Dec 1, 2023` where I think it should continue to say `Today`). I think it's quite likely for devices to drift a few minutes and hit this condition.

Open to further discussion, but for now I'm making a change so that for future dates, dates that are today still show up as `Today` but dates further off will have the year attached.

sirpengi:
@gnprice this is ready for another round!

test/widgets/message_list_test.dart
```diff
@@ -377,11 +377,34 @@ void main() {
...
+      tester.widget(find.textContaining(RegExp("Dec 1[89], 2022")));
+      tester.widget(find.textContaining(RegExp("Aug 2[23], 2022")));
     });
   });
 
+  group('formatHeaderDate', () {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+    final now = DateTime(2023, 1, 10, 12);
+    final testCases = [
+      ("2023-01-10 12:00", zulipLocalizations.today),
```
gnprice, line 389:
```suggestion
    final now = DateTime.parse("2023-01-10 12:00");
    final testCases = [
      ("2023-01-10 12:00", zulipLocalizations.today),
```

lib/widgets/message_list.dart
```diff
@@ -725,12 +726,47 @@ class RecipientHeaderDate extends StatelessWidget {
...
 }
 
-final _kRecipientHeaderDateFormat = DateFormat('y-MM-dd', 'en_US'); // TODO(#278)
+@visibleForTesting
+String formatHeaderDate(
+  ZulipLocalizations zulipLocalizations,
+  DateTime dateTime, {
+  required DateTime now,
+}) {
+  assert(!dateTime.isUtc && !dateTime.isUtc,
```
gnprice, line 742:
```suggestion
  assert(!dateTime.isUtc && !now.isUtc,
```

:-)

lib/widgets/message_list.dart
```diff
@@ -725,12 +726,47 @@ class RecipientHeaderDate extends StatelessWidget {
...
+    '`dateTime` and `now` need to be in local time.');
```
gnprice, line 743: This worried me for a bit — what if local time happens to be UTC?

But it looks like it's fine: `isUtc` is true only if the explicit `isUtc` was set to true at construction. Here's a test program:
```dart
void main() {
  final t1 = DateTime.fromMillisecondsSinceEpoch(1701739446000);
  final t2 = DateTime.fromMillisecondsSinceEpoch(1701739446000, isUtc: true);
  print("$t1 timeZoneName: ${t1.timeZoneName} isUtc: ${t1.isUtc}");
  print("$t2 timeZoneName: ${t2.timeZoneName} isUtc: ${t2.isUtc}");
}
```

And here's output:
```
$ TZ=UTC dart /tmp/time.dart 
2023-12-05 01:24:06.000 timeZoneName: UTC isUtc: false
2023-12-05 01:24:06.000Z timeZoneName: UTC isUtc: true
```

lib/widgets/message_list.dart
```diff
@@ -725,12 +726,47 @@ class RecipientHeaderDate extends StatelessWidget {
...
+  if (dateTime.year == now.year &&
```
gnprice, line 744: nit:
```suggestion
    '`dateTime` and `now` need to be in local time.');

  if (dateTime.year == now.year &&
```

The assert is less closely related to the same-day conditional than the latter is to what's after it. So since there's a blank line separating afterward, there should be a blank line separating before.

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
gnprice: > t looks like `format_time_modern` actually hasn't been used since it was reverted back in March 2022 in [zulip/zulip@7bc0e70](https://github.com/zulip/zulip/commit/7bc0e70693f49ea8faef2578158362491fe5470c) .

Hmm, yeah, I remember that conversation now. Ah well.


> and sadly it looks like this function will always return `Just now` for any times from the future.

Wow, yeah, that seems wrong.


> for future dates, dates that are today still show up as `Today` but dates further off will have the year attached.

Sure. Given that we're only showing to calendar-day granularity anyway, that's the behavior I'd expect here.

gnprice:
Thanks @sirpengi for the revision!

A few small comments above. In order to get this into a v0.0.7 today, and since it's already nighttime for you, I'll go ahead and fix those and merge.

Issue #428, opened 2023-11-29
Fix alignment of server version on home screen


alya:
There is a horizontal alignment issue with the server version on the home screen. We should center it like all the other items on the page, and make sure that there is a reasonable margin around it.

gnprice:
As a debugging clue: I think this issue appears specifically when the version line is longer than the width of the screen, and therefore wraps. (Which is typical if you're on chat.zulip.org or Zulip Cloud, rather than a numbered release, and if you have a phone with default display settings.) When it fits as a single line, it gets centered as intended.

PR #432, opened 2023-12-01
ui: Center-align actual text (not just column children) on beta homescreen

chrisbobbe:
Fixes: #428


gnprice:
Thanks for the fix! Looks good; merging.

PR #436, opened 2023-12-04
Add Noto Color Emoji back again; use it in message-list reaction chips on Android

### LINKED ISSUE #404: Try bundling an emoji font
Since #403, we don't bundle a font for emoji; instead we rely on the system fonts for showing Unicode emoji. This has advantages but also disadvantages:
* zulip/zulip-mobile#3885

The reason that since #403 we don't bundle a font is that we haven't found a good emoji font that works in Flutter on iOS, and although the one we were shipping seems to largely work on Android it's not clear it always does. Specifically there's this upstream issue:
* https://github.com/flutter/flutter/issues/134897#issuecomment-1821594910

and the symptom when it doesn't work is that using the font produces a blank space, even when the system font would have handled the emoji just fine.

If that upstream issue gets fixed, then we should be able to just resume shipping Noto Color Emoji in its COLRv1 form.

### LINKED ISSUE #435: AboutZulipPage: "Licenses" page broken
D16E5638-1896-4FB7-9D68-A5098408F5D3

chrisbobbe:
- Bundle Noto Color Emoji, with a TODO to not bundle it on iOS (I'll file an issue for that TODO)
  - I'll say this fixes #404.
  - The "Licenses" page works again; this fixes #435
- Use Noto Color Emoji on Android (but not iOS) for reaction chips
  - I'll file an issue to use it for text elsewhere (on Android). But using it for reaction chips is easier and more urgent.

Screenshots from the office Android device (Samsung Galaxy S9 running Android 9):

(ignore/pardon the battery-life popup in the "After" screenshot)

| Before | After |
| --- | --- |
| image | image |

Fixes: #404
Fixes: #435

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -111,11 +111,6 @@ flutter:
     - assets/Source_Sans_3/LICENSE.md
 
   fonts:
-    # Zulip's custom icons.  To use or edit, see class ZulipIcons.
-    - family: Zulip Icons
-      fonts:
-        - asset: assets/icons/ZulipIcons.ttf
-
     - family: Source Code Pro
       fonts:
         - asset: assets/Source_Code_Pro/SourceCodeVF-Upright.otf
@@ -127,5 +122,10 @@ flutter:
         - asset: assets/Source_Sans_3/SourceSans3VF-Upright.otf
         - asset: assets/Source_Sans_3/SourceSans3VF-Italic.otf
           style: italic
+
+    # Zulip's custom icons.  To use or edit, see class ZulipIcons.
+    - family: Zulip Icons
+      fonts:
+        - asset: assets/icons/ZulipIcons.ttf
 
     # If adding a font, remember to account for its license in lib/licenses.dart.
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -107,10 +107,27 @@ flutter:
   uses-material-design: true
 
   assets:
+    - assets/Noto_Color_Emoji/LICENSE
     - assets/Source_Code_Pro/LICENSE.md
     - assets/Source_Sans_3/LICENSE.md
 
   fonts:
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -107,14 +107,26 @@ flutter:
...
+    - assets/Noto_Color_Emoji/LICENSE
     - assets/Source_Code_Pro/LICENSE.md
     - assets/Source_Sans_3/LICENSE.md
 
   fonts:
-    # Zulip's custom icons.  To use or edit, see class ZulipIcons.
-    - family: Zulip Icons
+    # Google's emoji font. (Web uses these emoji for the "Google" emojiset.)
+    #
+    # This should always be used under a condition that excludes iOS.
```
gnprice, line 117:
```suggestion
    # This should not be used on iOS.
```

I kept wanting to read this as being in the direction of saying it *should* be used, then realized the point was the situations where it shouldn't.

gnprice (commented):
Thanks! One comment below. I'll go ahead and fix that and merge.

> I'll file an issue to use it for text elsewhere (on Android). But using it for reaction chips is easier and more urgent.

Sure. I'd take #404 to cover using it for emoji in messages too; but happy for this to close that issue if we split out a new one.

> with a TODO to not bundle it on iOS (I'll file an issue for that TODO)

(Just copying this here too so it's mentioned alongside the other one.)

pubspec.yaml
gnprice: Also tweaked the next line:
```diff
-    # This should always be used under a condition that excludes iOS.
-    # iOS doesn't support this font in any of its available formats,
+    # This should not be used on iOS.
+    # iOS doesn't support any of the formats this font is available in,
```

because it sounded like "its available formats" might refer to iOS rather than the font.

Issue #438, opened 2023-12-06
ui: Consistently use "Noto Color Emoji" for emojis on Android (i.e., always include it as a fallback font)


chrisbobbe:
In #436, we brought back Noto Color Emoji, but just for reaction chips. We should use it everywhere else that emojis can appear.

Issue #294, opened 2023-08-31
ui: Use "Source Sans 3" font for most UI text


chrisbobbe:
Since Source Sans 3 is a variable-weight font, we'll have to take some extra care not to regress on support for the user's "use bold text" accessibility setting. We've been doing that by styling text with a `TextStyle` object that comes from `weightVariableTextStyle` (or is merged with one that comes from that). Pass a `BuildContext` to that function to let it conditionalize on that accessibility setting.

One might normally set a "default font" for the app without using `TextStyle` objects: by setting `fontFamily` or `fontFamilyFallback` on an app-wide `ThemeData`. But `ThemeData` doesn't also have a `fontVariations` field, which controls text weight for weight-variable fonts. 

Discussion: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20text.20all.20weird/near/1633648

PR #439, opened 2023-12-06
text: Apply "Source Sans 3" and (on Android) "Noto Color Emoji" widely across the app

chrisbobbe:
(As usual please pardon the battery-life indicator in some of these screenshots. Ugh.)

In the second set of screenshots, note that the 🤯 emoji is made to look the same in the message list and the reaction chip. (This is taken on the office Android device.)

| Before | After |
| --- | --- |
| image | image |
| image | image |

Fixes: #294
Fixes: #438

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -101,7 +101,7 @@ class Paragraph extends StatelessWidget {
   final ParagraphNode node;
 
   static TextStyle getTextStyle(BuildContext context) => const TextStyle(
-    fontFamily: 'Source Sans 3',
+    fontFamily: kDefaultFontFamily,
     fontSize: 14,
     height: (17 / 14),
   ).merge(weightVariableTextStyle(context));
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -100,8 +100,9 @@ class Paragraph extends StatelessWidget {
 
   final ParagraphNode node;
 
-  static TextStyle getTextStyle(BuildContext context) => const TextStyle(
+  static TextStyle getTextStyle(BuildContext context) => TextStyle(
     fontFamily: kDefaultFontFamily,
+    fontFamilyFallback: defaultFontFamilyFallback,
     fontSize: 14,
     height: (17 / 14),
   ).merge(weightVariableTextStyle(context));
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -103,8 +103,8 @@ class Paragraph extends StatelessWidget {
   static TextStyle getTextStyle(BuildContext context) => TextStyle(
     fontFamily: kDefaultFontFamily,
     fontFamilyFallback: defaultFontFamilyFallback,
-    fontSize: 14,
-    height: (17 / 14),
+    fontSize: kBaseFontSize,
+    height: (17 / kBaseFontSize),
   ).merge(weightVariableTextStyle(context));
 
   @override
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -100,7 +100,7 @@ class Paragraph extends StatelessWidget {
 
   final ParagraphNode node;
 
-  static TextStyle getTextStyle(BuildContext context) => const TextStyle(
+  static const textStyle = TextStyle(
     fontSize: kBaseFontSize,
     height: (17 / kBaseFontSize),
   );
@@ -113,7 +113,7 @@ class Paragraph extends StatelessWidget {
 
     final text = _buildBlockInlineContainer(
       node: node,
-      style: getTextStyle(context),
+      style: textStyle,
     );
 
     // If the paragraph didn't actually have a `p` element in the HTML,
@@ -646,7 +646,7 @@ class UserMention extends StatelessWidget {
         // One hopes an @-mention can't contain an embedded link.
         // (The parser on creating a UserMentionNode has a TODO to check that.)
         linkRecognizers: null,
-        style: Paragraph.getTextStyle(context),
+        style: Paragraph.textStyle,
         nodes: node.nodes));
   }
 
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -100,12 +100,10 @@ class Paragraph extends StatelessWidget {
 
   final ParagraphNode node;
 
-  static TextStyle getTextStyle(BuildContext context) => TextStyle(
-    fontFamily: kDefaultFontFamily,
-    fontFamilyFallback: defaultFontFamilyFallback,
+  static TextStyle getTextStyle(BuildContext context) => const TextStyle(
     fontSize: kBaseFontSize,
     height: (17 / kBaseFontSize),
-  ).merge(weightVariableTextStyle(context));
+  );
 
   @override
   Widget build(BuildContext context) {
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -637,7 +637,7 @@ class StreamMessageRecipientHeader extends StatelessWidget {
       fontSize: 16,
       letterSpacing: 0.02 * 16,
       height: (18 / 16),
-    ).merge(weightVariableTextStyle(context, wght: 600, wghtIfPlatformRequestsBold: 900));
+    ).merge(weightVariableTextStyle(context, wght: 600));
 
     final Widget streamWidget;
     if (!showStream) {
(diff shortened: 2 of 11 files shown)


chrisbobbe:
I would probably write a test of `zulipTypography` to make sure it matches an expected value, which would be hard-coded in the test. I'll go ahead with that if the direction taken in this PR seems like one we want to pursue. 🙂

gnprice:
Thanks!

I see there's a lot going on here, so I'm afraid I won't be able to get to it today before my vacation. But I'll be glad to take a look in a couple of weeks when I'm back :slightly_smiling_face:

gnprice:
Oops, I didn't previously get back to this PR in my catching-up after vacation. Looking now; just rebased it onto current main.

lib/widgets/content.dart
```diff
@@ -99,10 +99,9 @@ class Paragraph extends StatelessWidget {
   final ParagraphNode node;
 
   static TextStyle getTextStyle(BuildContext context) => const TextStyle(
-    fontFamily: 'Source Sans 3',
-    fontSize: 14,
-    height: (17 / 14),
-  ).merge(weightVariableTextStyle(context));
+    fontSize: kBaseFontSize,
+    height: (17 / kBaseFontSize),
+  );
```
gnprice, line 104: For a further nice cleanup (probably in a further cleanup commit at the end):
```dart
  static const textStyle = TextStyle(
    fontSize: kBaseFontSize,
    height: (17 / kBaseFontSize));
```

Since we're no longer looking at `context`, this can become a constant.

lib/widgets/message_list.dart
```diff
@@ -433,11 +433,15 @@ class MarkAsReadWidget extends StatelessWidget {
...
-              ).merge(weightVariableTextStyle(context)),
+              textStyle:
+                // Restate [FilledButton]'s default, which inherits from
+                // [zulipTypography]…
+                Theme.of(context).textTheme.labelLarge!
+                // …then clobber some attributes to follow Figma:
+                .merge(const TextStyle(
+                  fontSize: 18,
+                  height: (23 / 18))
+                .merge(weightVariableTextStyle(context, wght: 400))),
```
gnprice, line 444: Hmm interesting. I guess this will be a case to consider when thinking about how we ultimately want to arrange the data flow for the text styles.

lib/widgets/text.dart
```diff
@@ -1,5 +1,69 @@
...
+/// This is a variable-weight font, so any [TextStyle] that uses this should be
+/// merged with the result of calling [weightVariableTextStyle].
+const kDefaultFontFamily = 'Source Sans 3';
+
+/// The [TextStyle.fontFamilyFallback] for use with [kDefaultFontFamily].
+final kDefaultFontFamilyFallback = [
+  // iOS doesn't support any of the formats this font is available in.
+  // If we use it on iOS, we'll get blank spaces where we could have had Apple-
+  // style emojis.
+  if (defaultTargetPlatform == TargetPlatform.android) 'Noto Color Emoji',
```
gnprice, line 65: This is a bit of a perilous pattern because `defaultTargetPlatform` can change over time. Not likely to happen in a running app — but it happens all the time in tests, which indeed is the point of `defaultTargetPlatform`.

So as a static final, `kDefaultFontFamilyFallback` will get computed once when it's first accessed, and subsequent accesses will get the same value even if `defaultTargetPlatform` has changed. If we have a test that cares about the value, then the test could pass or fail depending on which other tests ran first. That in turn can be a big pain for debugging why some test is behaving the way it is (e.g. for figuring out why some other change broke a test).

I think a good fix would be to make it a getter, just like `defaultTargetPlatform` itself:
```dart
List<String> get defaultFontFamilyFallback => [
```

There's no performance need to memoize it, because at the tip of the PR we're only calling this twice per build of `ZulipApp`.

gnprice (commented):
Thanks! Glad to see the text styling get more centralized like this. Generally these changes all look good — various small comments below.

lib/widgets/text.dart
```diff
@@ -100,3 +179,75 @@ FontWeight clampVariableFontWeight(double wght) {
...
+/// - this function's [BuildContext] param, and
+/// - for `wght`:
+///   - if present, the field's [TextStyle.fontWeight]
+///     (via [wghtFromFontWeight]), else
+///   - null.
+///
+/// If the field is null, gives null.
+///
+/// This assumes that if a non-null field omits [TextStyle.fontWeight], then
+/// the normal font weight is implied, and it will be expressed in the output.
```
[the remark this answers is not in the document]
chrisbobbe: Great; thanks for the suggestion!

lib/widgets/content.dart
chrisbobbe: Ah indeed; I meant to do this as a followup here but forgot.

chrisbobbe:

chrisbobbe:
Oh, and what do you think of https://github.com/zulip/zulip-flutter/pull/439#issuecomment-1841998420 :

> I would probably write a test of `zulipTypography` to make sure it matches an expected value, which would be hard-coded in the test. I'll go ahead with that if the direction taken in this PR seems like one we want to pursue. 🙂

? I haven't done that yet but could. A [`Typography`](https://api.flutter.dev/flutter/material/Typography-class.html) with all its details given literally will be very verbose.

chrisbobbe:
Just pushed a revision with a test that spot-checks `zulipTypography`; PTAL. 🙂

test/widgets/text_test.dart
```diff
@@ -7,6 +7,63 @@ import 'package:zulip/widgets/text.dart';
...
+          })));
+      return result;
+    }
+
+    final matchesFonts = it<TextStyle>()
+      ..fontFamily.equals(kDefaultFontFamily)
+      ..fontFamilyFallback.isNotNull().deepEquals(defaultFontFamilyFallback);
+
+    matchesWeight(FontWeight weight) => it<TextStyle>()
+      ..fontWeight.isNotNull().equals(weight)
```
gnprice, line 29: nit:
```suggestion
      ..fontWeight.equals(weight)
```

test/widgets/text_test.dart
```diff
@@ -7,6 +7,63 @@ import 'package:zulip/widgets/text.dart';
...
+          child: Builder(builder: (context) {
+            result = zulipTypography(context);
+            return const SizedBox.shrink();
+          })));
+      return result;
+    }
+
+    final matchesFonts = it<TextStyle>()
+      ..fontFamily.equals(kDefaultFontFamily)
+      ..fontFamilyFallback.isNotNull().deepEquals(defaultFontFamilyFallback);
```
gnprice, line 26: nit:
```suggestion
    final matchesFontFamilies = it<TextStyle>()
      ..fontFamily.equals(kDefaultFontFamily)
      ..fontFamilyFallback.isNotNull().deepEquals(defaultFontFamilyFallback);
```

since it's not concerned with the font weight, etc., only the font families

test/widgets/text_test.dart
```diff
@@ -7,6 +7,63 @@ import 'package:zulip/widgets/text.dart';
...
+    test('Typography has the assumed fields', () {
+      check(Typography().toDiagnosticsNode().getProperties().map((n) => n.name).toList())
+        .deepEquals(['black', 'white', 'englishLike', 'dense', 'tall']);
+    });
+  });
+
+  test('_convertTextTheme: TextTheme has the assumed fields', () {
+    check(const TextTheme().toDiagnosticsNode().getProperties().map((n) => n.name).toList())
+      .deepEquals([
+        'displayLarge',  'displayMedium',  'displaySmall',
```
gnprice, line 59: nit: can make this a bit more robust to possible upstream changes that we shouldn't need to care about:
```suggestion
    check(const TextTheme().toDiagnosticsNode().getProperties().map((n) => n.name).toList())
      .unorderedEquals([
        'displayLarge',  'displayMedium',  'displaySmall',
```

gnprice (commented):
Thanks for the revision! Sorry for the belated review. This all looks good, with a few nits.

I think there'll also be a few spots to update when you rebase atop the `package:checks` upgrade you're doing right now.

chrisbobbe:
Thanks for the review! Revision pushed, atop my latest revision for that `checks` upgrade, #502.

lib/widgets/text.dart
```diff
@@ -1,5 +1,139 @@
...
+///
+/// The same [TextStyle] should also specify [defaultFontFamilyFallback]
+/// for [TextStyle.fontFamilyFallback].
+///
+/// This is a variable-weight font, so any [TextStyle] that uses this should be
+/// merged with the result of calling [weightVariableTextStyle].
+const kDefaultFontFamily = 'Source Sans 3';
+
+/// The [TextStyle.fontFamilyFallback] for use with [kDefaultFontFamily].
+List<String> get defaultFontFamilyFallback => [
```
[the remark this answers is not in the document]
gnprice: nit: this now refers to a `defaultFontFamily`, which doesn't exist :slightly_smiling_face: :

    If there's some corner of the app where the two fonts aren't getting
    applied, we'll eventually find it, and apply defaultFontFamily and
    defaultFontFamilyFallback according to their doc comments, to fix.

gnprice (commented):
Thanks for the revision! One fresh nit below — then please go ahead and merge.

chrisbobbe:
Thanks! Merged, with that fix.

Issue #441, opened 2023-12-06
reactions: Text emoji have 2px less height than unicode and image emoji


chrisbobbe:
<img width="561" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/90cd8e61-b969-4bc3-a951-70d9634c0535">

We have code with this comment:

```dart
                    // So text-emoji chips are at least as tall as square-emoji
                    // ones (probably a good thing).
```

but that code fails to account for 1px of padding at the top and bottom. I'll send a fix soon.

PR #442, opened 2023-12-06
reactions: (Actually) make text-emoji chip's height consistent with others

chrisbobbe:
If we don't like the refactor in the second commit, we can drop it and #441 will still be fixed. 🙂

| Before | After |
| --- | --- |
| image | image |

Fixes: #441

### DIFF
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -160,7 +160,8 @@ class ReactionChip extends StatelessWidget {
                   children: [
                     // So text-emoji chips are at least as tall as square-emoji
                     // ones (probably a good thing).
-                    SizedBox(height: _squareEmojiScalerClamped(context).scale(_squareEmojiSize)),
+                    // 2 is for the `vertical: 1` padding around `emoji`.
+                    SizedBox(height: _squareEmojiScalerClamped(context).scale(_squareEmojiSize) + 2),
                     Flexible( // [Flexible] to let text emojis expand if they can
                       child: Padding(padding: const EdgeInsets.symmetric(horizontal: 3, vertical: 1),
                         child: emoji)),
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -178,8 +178,7 @@ class ReactionChip extends StatelessWidget {
                             color: labelColor,
                           ).merge(weightVariableTextStyle(context,
                               wght: selfVoted ? 600 : null)),
-                          label),
-                      )),
+                          label))),
                   ]);
                 })))));
   }
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -138,7 +138,7 @@ class ReactionChip extends StatelessWidget {
           child: Padding(
             // 1px of this padding accounts for the border, which Flutter
             // just paints without changing size.
-            padding: const EdgeInsetsDirectional.fromSTEB(4, 2, 5, 2),
+            padding: const EdgeInsetsDirectional.fromSTEB(4, 3, 5, 3),
             child: LayoutBuilder(
               builder: (context, constraints) {
                 final maxRowWidth = constraints.maxWidth;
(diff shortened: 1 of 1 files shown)


sirpengi:
There is a `the heard-coded value 2` typo in the commit message, but I do prefer the refactor in the second commit over the first one for two reasons: the fix is more straightfoward and I think this version is more robust to different emoji sizes. So my nit about the commit message is probably moot as I would suggest to use the second method.

lib/widgets/emoji_reaction.dart
```diff
@@ -159,12 +159,15 @@ class ReactionChip extends StatelessWidget {
...
-                    SizedBox(height: _squareEmojiScalerClamped(context).scale(_squareEmojiSize)),
                     Flexible( // [Flexible] to let text emojis expand if they can
                       child: Padding(padding: const EdgeInsets.symmetric(horizontal: 3, vertical: 1),
-                        child: emoji)),
+                        child: ConstrainedBox(
+                          constraints: BoxConstraints(
+                            // minHeight so text-emoji chips are at least as tall
+                            // as square-emoji ones (probably a good thing).
+                            minHeight: _squareEmojiScalerClamped(context).scale(_squareEmojiSize)),
+                          child: Center(widthFactor: 1,
```
gnprice, line 169: Hmm ­— it feels a little awkward to have this here, because it's basically saying the same thing as the `CrossAxisAlignment.center` on the Row.

lib/widgets/emoji_reaction.dart
```diff
@@ -159,12 +159,15 @@ class ReactionChip extends StatelessWidget {
                   mainAxisSize: MainAxisSize.min,
                   crossAxisAlignment: CrossAxisAlignment.center,
                   children: [
-                    // So text-emoji chips are at least as tall as square-emoji
-                    // ones (probably a good thing).
-                    SizedBox(height: _squareEmojiScalerClamped(context).scale(_squareEmojiSize)),
                     Flexible( // [Flexible] to let text emojis expand if they can
                       child: Padding(padding: const EdgeInsets.symmetric(horizontal: 3, vertical: 1),
```
gnprice, line 163: After this change, there's 1px vertical padding on each child of the Row. Can that padding move outward?

It seems like it'd be equivalent to have instead 1px vertical padding around the Row; and then in fact it should commute with the LayoutBuilder too, and be able to merge into the outer Padding.

After that, we wouldn't need the "+ 2" adjustment, so I think we could drop the second commit and that redundant-looking `Center`.

gnprice (commented):
Thanks for fixing this! Comments below.

> and I think this version is more robust to different emoji sizes.

@sirpengi Can you elaborate on this? On their face, both the old and new code look to me equally robust to changing the emoji size ­— they both refer to the `_squareEmojiSize` constant.

chrisbobbe:

lib/widgets/emoji_reaction.dart
```diff
@@ -162,11 +162,9 @@ class ReactionChip extends StatelessWidget {
                     // ones (probably a good thing).
                     SizedBox(height: _squareEmojiScalerClamped(context).scale(_squareEmojiSize)),
                     Flexible( // [Flexible] to let text emojis expand if they can
-                      child: Padding(padding: const EdgeInsets.symmetric(horizontal: 3, vertical: 1),
+                      child: Padding(padding: const EdgeInsets.symmetric(horizontal: 3),
                         child: emoji)),
-                    // Added vertical: 1 to give some space when the label is
-                    // taller than the emoji (e.g. because it needs multiple lines)
```
gnprice, line 168: As Greg suggested:
      https://github.com/zulip/zulip-flutter/pull/442/commits/ec4a21fdccdb21774ca656082fa921161d6b1ae0#r1442492101

Link is broken — GitHub links like …/pull/NNN/commits/HHH/… are unstable, because GitHub breaks them if the commit is no longer in the latest revision of the PR. Instead link to discussions on the "Conversation" tab of the PR, like https://github.com/zulip/zulip-flutter/pull/442#discussion_r1442492101 .

gnprice (commented):
Thanks! All LGTM with one nit; please go ahead and merge after fixing that.

lib/widgets/emoji_reaction.dart
chrisbobbe: Ah indeed. I knew that, but missed it in this case; thanks for the catch. 🙂

Issue #443, opened 2023-12-06
reactions: In text-emoji reactions, extra space on the left when emoji text is multi-line


chrisbobbe:
I noticed this while investigating #441.

Screenshots from iOS, with no text scaling:

| Current | Should be |
| --- | --- |
| <img width="449" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/20cc6c51-3ffb-4283-a562-a401da55ad9f"> | <img width="449" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/f5ced3da-d998-49e8-8959-d6e4e8913562"> |

PR #444, opened 2023-12-06
reactions: For text emoji, remove extra space on the left

chrisbobbe:
Fixes: #443


sirpengi:
I've reproduced and checked that this fixes the issue. It works in theory and practice. No notes!

aarishshahmohsin:
@zulipbot claim

gnprice:
Thanks @chrisbobbe for fixing this! And thanks @sirpengi for validating the repro and for the first review. The code LGTM; merging.

Issue #445, opened 2023-12-08
Improve threshold behavior for fetching new batch of messages on scroll


sirpengi:
As the user scrolls through the message list, we fetch a new batch of messages before they reach the end of the loaded history. If the user has a good network connection this is an amazing experience given they can continuously fling through history.

Currently we have a static value for the threshold at which we load extra messages (see `kFetchMessagesBufferPixels` in `lib/widgets/message_list.dart`) which currently evaluates to 4000. When it was first introduced in bccaead94e messages always had a recipient header attached, and since then not only are messages more compact (moreso due to work towards #157 ) but messages are grouped together by recipient header. A simple static value by itself is no longer sufficient, and we should probably dynamically adjust this value or change what we rely on to fetch new messages.

gnprice:
Before this will be actionable, we'll need to have an idea of what problem the existing behavior has that we want to solve.

I'm not actually sure the threshold shouldn't ultimately remain a fixed distance in pixels, as it is now. That corresponds to a fixed amount of scrolling, which seems more likely to be relevant in varying circumstances than the number of messages involved. (With #446, it'll actually become quite difficult for the user to even tell how many messages are on their screen.) It's possible that distance would ideally be larger or smaller than the value we have now, but the first step in making a change there would be to identify a problem that it might solve.

I believe the prompt for filing this issue was the test discussed at https://github.com/zulip/zulip-flutter/pull/446#discussion_r1433497393 , which can be fragile and need adjustment when the message layout changes. I'm not sure different threshold behavior here would actually fix that, though. As written, the test starts a fling-scroll at full speed and then fast-forwards to the end of it. That seems likely to skip right over the point where some other threshold would fire, just as it can skip right over the point where the existing threshold fires. It'd be good to make the test more robust, but that's probably best done by changes to the test.

gnprice:
Closing as not actionable. We can always open a new issue if there arises a concrete problem in this area that we want to fix.

Issue #170, opened 2023-06-09
Star and unstar messages; show starred status


gnprice:
This Zulip feature:
https://zulip.com/help/star-a-message

Out of scope for this issue is to show a message list for the narrow consisting of starred messages. We'll file a separate issue for that.

terpimost:
The basic idea here is Long tap > Star message > show icon on the right gutter (the same as in desktop zulip)
After that we could think of something better (swipe the message to a side to see buttons like iOS does)

gnprice:
Cool, that sounds good. That long-tap interaction is the same as we currently have in zulip-mobile, then.

For the icon in the right gutter, is the current design in Figma all set, or are there any updates you'd like to make from that?

Issue #157, opened 2023-06-07
Convert message list to new, 2023+ design


gnprice:
In the current prototype, the visual design of the message list is based closely on Zulip web's design as it was last December, when I started on the prototype and wrote most of the code we have for that so far. This also makes it broadly similar to the zulip-mobile RN app, which uses a design based on the same generation of Zulip web's design.

Since then Zulip web has undergone a substantial redesign, to designs by @terpimost:
https://blog.zulip.com/2023/05/31/zulip-7-0-released/#redesign
and some of the major changes there are in the message list. Among them:
* Stream/topic recipient bars have the stream color the whole way across, are taller, have the stream name and topic in bold, and other changes.
* Background colors for DMs and for messages with @-mentions have changed:
  * #647
  * #681
* The styling of @-mention pills themselves has changed. (#646)

We should target the new design, instead of the old.

Probably the way to do this is to basically go from scratch, rather than incrementally modify the existing design. (This is one advantage we have as a prototype compared to making the same change in web.) That means:
* In all the places in `widgets/message_list.dart` and `widgets/content.dart` where we're controlling styling and layout, expect to rip that code out and replace it with newly-written code. (On the other hand, the code that's about finding appropriate data to show, or walking a `ContentNode` tree, should mostly not need to change for this.)
* To write the new versions of that code, repeat the reverse-engineering work I did back in 2022-12 on web's design, but this time on the new design. Look at the corresponding rendering in web using browser devtools; refer to the CSS styles for exact values for spacing, a rounded corner's radius, and so on; and use Flutter desktop to put the prototype side by side with web on screen, and Flutter's fast hot reload to dial details in exactly.

We won't want to match web's design *exactly*:
* Notably there's that wide gutter at the right of the message content, which is costly on a mobile-sized screen and in any case serves no purpose on a touchscreen because it's just reserving space for UI to appear on hover. So we'll cut that gutter.
* The "blue box" cursor is for the sake of keyboard shortcuts, and we'll leave it out.
* There's probably some other aspect of the message list where there'll be a good reason for mobile to differ from web, but I'm not thinking of one right now.

terpimost:
If confused about anything or willing to simplify design because of any reason please tell me and I try my best to help

gnprice:
A note for when we take this on: I think it will make sense to pursue both the light theme and dark theme at the same time.

[...]

PR #446, opened 2023-12-08
msglist: Update layout in messages

sirpengi:
~~Rebased on top of #447 as that impacted tests passing.~~

As a foundation for my work on #170, I needed to update the layout of messages so that the star icon could be positioned in the appropriate place. This also enables the placement of future decorations to messages (like markers for moved/edited) and moves us closer to #157 

Screenshot of the new message layout:
Simulator Screenshot - iPhone 14 - 2023-12-07 at 15 20 51

Because the new layout is more compact, this caused the `basic` test of the `fetch older messages on scroll` group to fail as a second request for new messages is triggered. This is due to the first batch of new messages not being tall enough to be beyond the threshold (see `kFetchMessagesBufferPixels` in `lib/widgets/message_list.dart`) for loading another batch. A proper fix to this would be to tweak the threshold behavior, but instead opted to fix the test by adjusting the response size and opened #445.

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -857,60 +857,64 @@ class MessageWithPossibleSender extends StatelessWidget {
 
     Widget? senderRow;
     if (item.showSender) {
-      senderRow = GestureDetector(
-        onTap: () => Navigator.push(context,
-          ProfilePage.buildRoute(context: context,
-            userId: message.senderId)),
-        child: Text(message.senderFullName, // TODO get from user data
-          style: const TextStyle(
-            fontFamily: 'Source Sans 3',
-            fontSize: 18,
-            height: (22 / 18),
-          ).merge(weightVariableTextStyle(context, wght: 600,
-                    wghtIfPlatformRequestsBold: 900))));
+      senderRow = Row(
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        crossAxisAlignment: CrossAxisAlignment.baseline,
+        textBaseline: TextBaseline.alphabetic,
+        children: [
+          Flexible(
+            child: GestureDetector(
+              onTap: () => Navigator.push(context,
+                ProfilePage.buildRoute(context: context,
+                  userId: message.senderId)),
+              child: Row(
+                children: [
+                  Avatar(size: 32, borderRadius: 3,
+                    userId: message.senderId),
+                  const SizedBox(width: 8),
+                  Flexible(
+                    child: Text(message.senderFullName, // TODO get from user data
+                      style: const TextStyle(
+                        fontFamily: 'Source Sans 3',
+                        fontSize: 18,
+                        height: (22 / 18),
+                      ).merge(weightVariableTextStyle(context, wght: 600,
+                                wghtIfPlatformRequestsBold: 900)),
+                      overflow: TextOverflow.ellipsis)),
+                ]))),
+          const SizedBox(width: 4),
+          Text(time,
+            style: TextStyle(
+              color: _kMessageTimestampColor,
+              fontFamily: 'Source Sans 3',
+              fontSize: 16,
+              height: (18 / 16),
+              fontFeatures: const [FontFeature.enable('c2sc'), FontFeature.enable('smcp')],
+            ).merge(weightVariableTextStyle(context))),
+        ]);
     }
 
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
       onLongPress: () => showMessageActionSheet(context: context, message: message),
       child: Padding(
-        padding: const EdgeInsets.only(top: 2, bottom: 3, left: 8, right: 8),
+        padding: const EdgeInsets.symmetric(vertical: 4),
         child: Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
-          item.showSender
-            ? Padding(
-                padding: const EdgeInsets.fromLTRB(3, 6, 11, 0),
-                child: GestureDetector(
-                  onTap: () => Navigator.push(context,
-                    ProfilePage.buildRoute(context: context,
-                      userId: message.senderId)),
-                  child: Avatar(size: 35, borderRadius: 4,
-                    userId: message.senderId)))
-            : const SizedBox(width: 3 + 35 + 11),
+          const SizedBox(width: 16),
           Expanded(
             child: Column(
               crossAxisAlignment: CrossAxisAlignment.stretch,
               children: [
-                if (item.showSender) ...[
-                  const SizedBox(height: 3),
-                  senderRow!,
+                if (senderRow != null) ...[
+                  const SizedBox(height: 2),
+                  senderRow,
                   const SizedBox(height: 4),
                 ],
                 MessageContent(message: message, content: item.content),
                 if ((message.reactions?.total ?? 0) > 0)
                   ReactionChipsList(messageId: message.id, reactions: message.reactions!)
               ])),
-          Container(
-            width: 80,
-            padding: const EdgeInsets.only(top: 4, right: 16 - 8),
-            alignment: Alignment.topRight,
-            child: Text(time,
-              style: TextStyle(
-                color: _kMessageTimestampColor,
-                fontFamily: 'Source Sans 3',
-                fontSize: 16,
-                height: (18 / 16),
-                fontFeatures: const [FontFeature.enable('c2sc'), FontFeature.enable('smcp')],
-              ).merge(weightVariableTextStyle(context)))),
+          const SizedBox(width: 16),
         ])));
   }
 }
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -899,22 +899,22 @@ class MessageWithPossibleSender extends StatelessWidget {
       onLongPress: () => showMessageActionSheet(context: context, message: message),
       child: Padding(
         padding: const EdgeInsets.symmetric(vertical: 4),
-        child: Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
-          const SizedBox(width: 16),
-          Expanded(
-            child: Column(
-              crossAxisAlignment: CrossAxisAlignment.stretch,
-              children: [
-                if (senderRow != null) ...[
-                  const SizedBox(height: 2),
-                  senderRow,
-                  const SizedBox(height: 4),
-                ],
-                MessageContent(message: message, content: item.content),
-                if ((message.reactions?.total ?? 0) > 0)
-                  ReactionChipsList(messageId: message.id, reactions: message.reactions!)
-              ])),
-          const SizedBox(width: 16),
+        child: Column(children: [
+          if (senderRow != null)
+            Padding(padding: const EdgeInsets.fromLTRB(16, 2, 16, 4),
+              child: senderRow),
+          Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
+            const SizedBox(width: 16),
+            Expanded(
+              child: Column(
+                crossAxisAlignment: CrossAxisAlignment.stretch,
+                children: [
+                  MessageContent(message: message, content: item.content),
+                  if ((message.reactions?.total ?? 0) > 0)
+                    ReactionChipsList(messageId: message.id, reactions: message.reactions!)
+                ])),
+            const SizedBox(width: 16),
+          ]),
         ])));
   }
 }
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -855,6 +855,21 @@ class MessageWithPossibleSender extends StatelessWidget {
     final time = _kMessageTimestampFormat
       .format(DateTime.fromMillisecondsSinceEpoch(1000 * message.timestamp));
 
+    Widget? senderRow;
+    if (item.showSender) {
+      senderRow = GestureDetector(
+        onTap: () => Navigator.push(context,
+          ProfilePage.buildRoute(context: context,
+            userId: message.senderId)),
+        child: Text(message.senderFullName, // TODO get from user data
+          style: const TextStyle(
+            fontFamily: 'Source Sans 3',
+            fontSize: 18,
+            height: (22 / 18),
+          ).merge(weightVariableTextStyle(context, wght: 600,
+                    wghtIfPlatformRequestsBold: 900))));
+    }
+
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
       onLongPress: () => showMessageActionSheet(context: context, message: message),
@@ -877,17 +892,7 @@ class MessageWithPossibleSender extends StatelessWidget {
               children: [
                 if (item.showSender) ...[
                   const SizedBox(height: 3),
-                  GestureDetector(
-                    onTap: () => Navigator.push(context,
-                      ProfilePage.buildRoute(context: context,
-                        userId: message.senderId)),
-                    child: Text(message.senderFullName, // TODO get from user data
-                      style: const TextStyle(
-                        fontFamily: 'Source Sans 3',
-                        fontSize: 18,
-                        height: (22 / 18),
-                      ).merge(weightVariableTextStyle(context, wght: 600,
-                                wghtIfPlatformRequestsBold: 900)))),
+                  senderRow!,
                   const SizedBox(height: 4),
                 ],
                 MessageContent(message: message, content: item.content),
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -852,11 +852,11 @@ class MessageWithPossibleSender extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final message = item.message;
-    final time = _kMessageTimestampFormat
-      .format(DateTime.fromMillisecondsSinceEpoch(1000 * message.timestamp));
 
     Widget? senderRow;
     if (item.showSender) {
+      final time = _kMessageTimestampFormat
+        .format(DateTime.fromMillisecondsSinceEpoch(1000 * message.timestamp));
       senderRow = Row(
         mainAxisAlignment: MainAxisAlignment.spaceBetween,
         crossAxisAlignment: CrossAxisAlignment.baseline,
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -841,6 +841,9 @@ String formatHeaderDate(
 }
 
 /// A Zulip message, showing the sender's name and avatar if specified.
+// Design referenced from:
+//   - https://github.com/zulip/zulip-mobile/issues/5511
+//   - https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=538%3A20849&mode=dev
 class MessageWithPossibleSender extends StatelessWidget {
   const MessageWithPossibleSender({super.key, required this.item});
 
@@ -855,7 +858,6 @@ class MessageWithPossibleSender extends StatelessWidget {
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
       onLongPress: () => showMessageActionSheet(context: context, message: message),
-      // TODO clean up this layout, by less precisely imitating web
       child: Padding(
         padding: const EdgeInsets.only(top: 2, bottom: 3, left: 8, right: 8),
         child: Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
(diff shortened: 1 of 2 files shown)


chrisbobbe (commented):
Thanks @sirpengi! Small comments below. It's nice to have more horizontal room for message contents. It could feel kind of cramped sometimes, in the previous design.

I did notice some potential downsides, though. Not to block merging this, but just maybe something we might be interested in soliciting feedback about in the alpha/beta.

- It seems like it might be harder to tell who sent a given piece of message content, if there's a lot of message content above it and below the corresponding sender information. When scanning upward for the sender information, message content will be more heterogenous and distracting than pure empty space, if that makes sense.
- It will be more difficult (sometimes a bit, sometimes a lot) to tell where message boundaries are. Before, this was made easy by the date markers; now, I think they have to be identified by the insertion of an 8px gap, between blocks that usually don't have clearly marked boundaries, and may contain their own vertical whitespace. To exaggerate the point:

  | Before | After |
  | --- | --- |
  | A7F80BC2-4130-4C31-96DC-E2C1A4845EBA | 891D4CD4-C7F6-40E3-9E7D-ACD875F28E9B |

- We've lost the information of when a message was sent, except if it's the first message in a run of messages from the same sender, or if it's the first message after a date boundary. So for example:

  | Before | After |
  | --- | --- |
  | B6705706-6CAD-4B69-B6DE-32069DE412E7 | 85E31DBF-EB1F-4D13-B377-ED7E83894BA6 |
  
  This might be something we can fix as a followup, but it seems like it might be complicated. The Figma has a design to support checking the date/time for any message; it looks like it wants us to let the user slide a message to the left, to reveal the timestamp: https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=147-8975&mode=design&t=P97AuMzjt47Mudyg-0 . (I wonder if we can make the slide gesture reveal dates for all visible messages at once, like iMessage on my iPhone, so you don't have to slide once for each message you want to check.)

lib/widgets/message_list.dart
```diff
@@ -781,58 +783,81 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                            fontSize: 18,
+                            height: (22 / 18),
+                          ).merge(weightVariableTextStyle(context, wght: 600,
+                              wghtIfPlatformRequestsBold: 900)),
+                          overflow: TextOverflow.ellipsis),
+                      ),
+                    ]),
+                )),
+            ),
+            Text(time,
```
gnprice, line 819: The only place we use the `time` local is now inside this `item.showSender` conditional. So that local's definition can be moved inside the conditional too.

lib/widgets/message_list.dart
```diff
@@ -785,66 +785,72 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-                if ((message.reactions?.total ?? 0) > 0)
-                  ReactionChipsList(messageId: message.id, reactions: message.reactions!)
-              ])),
-          const SizedBox(width: 16),
-        ])));
+      child: Column(
+        children: [
+          senderWidget,
+          Padding(
+            padding: const EdgeInsets.symmetric(vertical: 4),
```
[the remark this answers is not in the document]
gnprice: I agree, and I think those extra 4px don't correspond to anything in the Figma design.

AFAICT this last commit is meant to have no effect on the visible layout. Is that right? I'm not sure the effect on the source code is an improvement; so perhaps that commit can just be left out. If something like it is needed for a followup change like indicating starred messages, then it can appear as a prep commit (but with those extra 4px fixed) in that PR, where it'll be easier to assess what it's doing for us.

lib/widgets/message_list.dart
```diff
@@ -781,58 +783,81 @@ class MessageWithPossibleSender extends StatelessWidget {
     final time = _kMessageTimestampFormat
       .format(DateTime.fromMillisecondsSinceEpoch(1000 * message.timestamp));
 
+    final Widget senderWidget;
+    if (item.showSender) {
+      senderWidget = Padding(
```
gnprice, line 788: In the end this widget is only partly about the sender, because it also has the timestamp; so `senderWidget` is a bit misleading. A good name might be `senderRow` — it's the row with the sender (which also happens to have a bit more than the sender).

lib/widgets/message_list.dart
```diff
@@ -781,58 +783,81 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-              ])),
-          Container(
-            width: 80,
-            padding: const EdgeInsets.only(top: 4, right: 16 - 8),
-            alignment: Alignment.topRight,
-            child: Text(time, style: _kMessageTimestampStyle)),
-        ])));
+      child: Column(
+        children: [
+          senderWidget,
```
gnprice, line 837: nit:
```suggestion
          if (senderWidget != null) senderWidget,
```

and then instead of `SizedBox.shrink`, can set the variable to null. I think that'd make it a bit more evident when looking at this spot that it's conditional whether there's really any sender anything being shown.

test/widgets/message_list_test.dart
```diff
@@ -439,7 +445,7 @@ void main() {
         ..statusCode = HttpStatus.ok
         ..content = kSolidBlueAvatar;
 
-      await setupMessageListPage(tester, messageCount: 10);
+      await setupMessageListPage(tester, messageCount: 1);
```
gnprice, line 448: One message instead of ten seems fine — but also I don't actually get any failure if I revert this change. So perhaps simplify the explanation by just leaving it out.

test/widgets/message_list_test.dart
```diff
@@ -85,17 +85,23 @@ void main() {
...
+      //   Message batch sizes here changed to 200 -- an unnatural
+      //   result in the app given `kMessageListFetchBatchSize`
+      //   is 100. This is necessary in this test as the new messages
+      //   loaded are shorter than the threshold for fetching more messages
+      //   causing an extra request to be made.
       await setupMessageListPage(tester, foundOldest: false,
-        messages: List.generate(200, (i) => eg.streamMessage(id: 950 + i, sender: eg.selfUser)));
-      check(itemCount(tester)).equals(203);
+        messages: List.generate(150, (i) => eg.streamMessage(id: 950 + i, sender: eg.selfUser)));
+      check(itemCount(tester)).equals(153);
```
gnprice, line 96: Here's a simpler solution:

```suggestion
        messages: List.generate(300, (i) => eg.streamMessage(id: 950 + i, sender: eg.selfUser)));
      check(itemCount(tester)).equals(303);
```

and then the line at the end of the test expects 403 items instead of 303.

This way the new batch when it fetches more can remain 100 messages like it is now. That also makes the apologetic comment unnecessary — the message list starts with 300 messages, which is more than 100, but it's perfectly normal for a message list to have 300 messages in it if the user has already scrolled up some.

I think the key to what's going on here isn't actually anything untoward in how we're deciding when to fetch more messages. Rather, it's this line of the test:
```dart
      await tester.pump(const Duration(seconds: 3)); // Fast-forward to end of fling.
```
where we don't let any app code run between when the user starts a fling scroll (one at full speed, 8000 px/sec), and three seconds later when the fling scroll should be complete. With 200 very compact example messages (short contents, and all from the same sender to the same recipient on the same day), it's likely we've scrolled past all of them by that point. In normal operation we'd have started the next fetch long before then, but with this fast-forward we don't get the chance.

gnprice (commented):
Thanks @sirpengi, and thanks @chrisbobbe for the initial review!

Generally these changes look good. Some comments below, and I also agree with Chris's comments above.

One branch-level observation: I think a lot of these substantive changes don't make a lot of sense done separately. So I think a good structure probably looks like: some prep commits, then one commit that does all the main layout changes, then perhaps some followup NFC commits. For example:
* tweak that one test to have only one message instead of 10 (… or perhaps skip that change entirely, see below)
* maybe pull out the `senderWidget` (or `senderRow`) local, an NFC change
* maybe change the fonts
* then all the other changes squashed together, except for the last one "4f04046f9 msglist: Move message sender out of row containing message contents" (probably leave it out entirely, as discussed below; but also it's one that rearranges more of the code than the other changes)
* then the Row → Padding change Chris points out above, an NFC change

gnprice:
> * It will be more difficult (sometimes a bit, sometimes a lot) to tell where message boundaries are. Before, this was made easy by the date markers; […]
> * We've lost the information of when a message was sent, except if it's the first message in a run of messages from the same sender, or if it's the first message after a date boundary. […]

Yeah, I don't love those aspects either. But both are basically already true in zulip-mobile — the information is there if you know where to look, but I think it's awfully subtle and wouldn't call our design there a success. So I'm OK with losing that information for now, as the price of making good use of the available width, and trying to get it back later as a followup:

> The Figma has a design to support checking the date/time for any message; it looks like it wants us to let the user slide a message to the left, to reveal the timestamp […] (I wonder if we can make the slide gesture reveal dates for all visible messages at once, like Messages on my iPhone, so you don't have to slide once for each message you want to check.)

And I agree that last part sounds like a nice touch.

gnprice:
Filed a followup issue for revealing the message timestamps:
* #457

lib/widgets/message_list.dart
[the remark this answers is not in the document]
sirpengi: From: https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=134%3A8312&mode=dev
messages_4px_padding

The mockup here shows each distinct message item has 4px vertical margins. The sender widget is complicated in the design as it has 2px of vertical padding (making a total of 6px top margin if the sender is shown) and there is a 2px margin above the message content and the sender:

message_avatar_content_padding

So taking the figma as is, message contents would have dynamic padding depending if the sender is visible or not.  But I think to simplify things, given there is effectively 4px spacing between the contents and the sender (2px coming from the sender padding and 2px between a message content and the sender), there ends up being 4px vertical padding for message contents in all instances. If we then give the sender 6px top margin we achieve the same layout.

An alternative would be to put that 4px padding around the entire `MessageWithPossibleSender` (then giving the sender widget 2px top and 4px bottom padding), but my intention here with the layout is to leave the left/right gutters as space to add decorations (like the incoming star-marker) and having the padding in the container would limit how we could place such decorations, for example we could not align anything to the corner. There's no need for this currently but I do think a layout that gives us more flexibility to adjust those decorations is better.

(This discussion ignores that bottom padding on messages are expanded when it is before a new recipient header, but that case is already handled by introducing extra space in those situations)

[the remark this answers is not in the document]
gnprice: > So taking the figma as is, message contents would have dynamic padding depending if the sender is visible or not. But I think to simplify things, given there is effectively 4px spacing between the contents and the sender (2px coming from the sender padding and 2px between a message content and the sender), there ends up being 4px vertical padding for message contents in all instances. If we then give the sender 6px top margin we achieve the same layout.

Cool. Yeah, in general it's fine for the layout to be expressed differently from how it is in Figma as long as it produces an equivalent result. We've done that in a fair number of places already. For us the Figma design expresses a visual result, but not the details of how to implement it.

lib/widgets/message_list.dart
```diff
@@ -796,7 +796,7 @@ class MessageWithPossibleSender extends StatelessWidget {
                   onTap: () => Navigator.push(context,
                     ProfilePage.buildRoute(context: context,
                       userId: message.senderId)),
-                  child: Avatar(size: 35, borderRadius: 4,
+                  child: Avatar(size: 32, borderRadius: 3,
                     userId: message.senderId)))
             : const SizedBox(width: 3 + 35 + 11),
```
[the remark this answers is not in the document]
sirpengi: I've ended up consolidating the commits so the orphaned `35` no longer exists

sirpengi:
@gnprice ready for review again! The commits were re-ordered, and some adjustments were made for comments and to polish up the layout. I was also able to minimize the test changes thanks to your comments.

gnprice:
I'm confused by the outcome of https://github.com/zulip/zulip-flutter/pull/446#discussion_r1433463085 . It looks like the change we were talking there about removing is still in the PR, and in fact has gotten squashed into the main commit. That's a change I specifically asked at https://github.com/zulip/zulip-flutter/pull/446#pullrequestreview-1792026138 to leave unsquashed: 
> * then all the other changes squashed together, except for the last one "4f04046f9 msglist: Move message sender out of row containing message contents" (probably leave it out entirely, as discussed below; but also it's one that rearranges more of the code than the other changes)

The main commit is pretty hard to read now, and I suspect it's because that change has gotten squashed into it. Can you please take it back out?

sirpengi:
@gnprice apologies! In my revision I found the change important to include (it separates the `senderRow` from the element containing the gutters so that future items in the gutter can be laid out relative to the message), but it was hard to detangle it from the other changes I wanted to make. I think it was problematic in the first iteration because I also tweaked the layout when I moved elements around. Here I've got another attempt, with the same result, but tried to be cleaner with combining conceptual changes.

gnprice:
We looked at this in our call today, as suggested above, and the plan is that you're going to make another revision of those last two commits, in which the change that had been the last commit of the December revision becomes its own commit again, this time an NFC commit. (That change is the one that rearranges the 16px horizontal padding, so that it applies separately to the sender vs. the content-plus-reactions.)

sirpengi:
@gnprice ready for review again! The order of operations was shifted around a bit in this latest version, but I'm quite satisfied with the separation of changes here.

lib/widgets/message_list.dart
```diff
@@ -849,61 +852,74 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-        padding: const EdgeInsets.only(top: 2, bottom: 3, left: 8, right: 8),
-        child: Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
-          item.showSender
-            ? Padding(
-                padding: const EdgeInsets.fromLTRB(3, 6, 11, 0),
+    final Widget? senderRow = item.showSender
+      ? Padding(
+          padding: const EdgeInsets.only(top: 2, bottom: 4, left: 16, right: 16),
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
```
gnprice, line 860: This `spaceBetween` is missing in the commit
1d3d41d20 msglist: Consolidate sender avatar, name and message time into same row

and as a result the timestamp goes immediately after the sender name:
image

gnprice: More generally, it seems like a lot of what this later commit:
53edaad70 msglist: Update message sender style and layout

is doing is just fixing bugs in the earlier commit — things that became necessities as a consequence of creating this `Row` for the sender avatar, name, and timestamp.

The lack of `spaceBetween` is the most visible, but for another example: similarly to `spaceBetween`, if we're going to have the name and the timestamp in one Row then it seems like we naturally want `CrossAxisAlignment.baseline` for that. And we need `Flexible` and `TextOverflow.ellipsis` so that things don't break if a sender has a long name.

As discussed in Zulip's general commit style guide:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#each-commit-must-be-coherent
each commit should leave things in a state it would be reasonable for us to deploy as is. (I.e. to release as is, to use the verb that's applicable for a client.) So if part of a PR introduces a bug and then a later part fixes the bug, the fix should instead get squashed into the commit that introduced it.

lib/widgets/message_list.dart
```diff
@@ -849,61 +852,74 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                style: TextStyle(
+                  color: _kMessageTimestampColor,
+                  fontFamily: 'Source Sans 3',
+                  fontSize: 16,
+                  height: (18 / 16),
+                  fontFeatures: const [FontFeature.enable('c2sc'), FontFeature.enable('smcp')],
+                ).merge(weightVariableTextStyle(context))),
+            ]))
+      : null;
+    return GestureDetector(
```
gnprice, line 896: nit: these are giant statements, so separate by a blank line:
```suggestion
      : null;

    return GestureDetector(
```

gnprice (commented):
Thanks for the revision! My high-level comments are again about the branch structure.

The main one (details below) is about early commits introducing a broken state that later commits fix up, which is something we avoid in Zulip's style of a clean commit history.

This goes back to the overall comment I made in my original review above at https://github.com/zulip/zulip-flutter/pull/446#pullrequestreview-1792026138 :
> One branch-level observation: I think a lot of these substantive changes don't make a lot of sense done separately. So I think a good structure probably looks like: some prep commits, then one commit that does all the main layout changes, then perhaps some followup NFC commits.

That structure would mean that all the changes in these three commits:
1d3d41d20 msglist: Consolidate sender avatar, name and message time into same row
[…]
53edaad70 msglist: Update message sender style and layout
30dbcc5df msglist: Adjust layout and spacing in messages

would go into a single commit. Those correspond to this item in my original review:
> * then all the other changes squashed together, except for the last one "4f04046f9 msglist: Move message sender out of row containing message contents" (probably leave it out entirely, as discussed below; but also it's one that rearranges more of the code than the other changes)

Your previous revision (the one at https://github.com/zulip/zulip-flutter/pull/446#issuecomment-1883295359 ) successfully squashed those together, but it also squashed in that other change that I asked to be kept separate. So what I'd still like to see is a branch structure of the kind I requested in my original review.

Once we have that, I think there'll still be a few comments to make about the code and the layout, but it'll be a lot easier to review and discuss in that form.

sirpengi:
@gnprice updated again!

lib/widgets/message_list.dart
```diff
@@ -849,61 +852,75 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+            ]))
+      : null;
+
+    return GestureDetector(
+      behavior: HitTestBehavior.translucent,
+      onLongPress: () => showMessageActionSheet(context: context, message: message),
+      child: Padding(
+        padding: const EdgeInsets.symmetric(vertical: 4),
+        child: Column(children: [
+          if (senderRow != null) senderRow,
```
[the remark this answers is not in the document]
gnprice: Also on this commit message, a nit: "move into a separate row" doesn't seem to me to really capture the point. After all, it's already a `Row`, right? (As the name `senderRow` suggests.) And if by "row" one means "element of a `Column`"… it was already that, too.

I think a better summary line would be
`msglist [nfc]: Move senderRow to an outer Column`

That is, the thing that's changing is that now `senderRow` appears as an element of a new *outer* Column, one that's outside of the Row that provides the gutters.

Or perhaps better yet:
`msglist [nfc]: Group message and sender as Column of Rows, not vice versa`

gnprice, line 903: I think the layout here is somewhat easier to read and understand if the padding around senderRow stays here:
```suggestion
          if (senderRow != null)
            Padding(padding: const EdgeInsets.fromLTRB(16, 2, 16, 4),
              child: senderRow),
```

That way the "16" here is just a couple of lines away from the "16" that gives the widths of the gutters around the message body, below. And similarly the vertical 2px and 4px are close by the 4px vertical padding around the whole thing, which they interact with closely.

(The version I've given also contains a smaller change: the use of `EdgeInsets.fromLTRB` is helpful for keeping things compact.)

lib/widgets/message_list.dart
```diff
@@ -849,61 +852,75 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-      child: Padding(
-        padding: const EdgeInsets.only(top: 2, bottom: 3, left: 8, right: 8),
-        child: Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
-          item.showSender
-            ? Padding(
-                padding: const EdgeInsets.fromLTRB(3, 6, 11, 0),
+    final Widget? senderRow = item.showSender
+      ? Padding(
+          padding: const EdgeInsets.only(top: 2, bottom: 4, left: 16, right: 16),
+          child: Row(
```
gnprice, line 859: nit: This is a real long conditional expression. I think things get a bit cleaner if we instead use a normal `if` conditional:
```suggestion
    Widget? senderRow;
    if (item.showSender) {
      senderRow = Row(
```

(That version is combined with my other comment about keeping the `Padding` below.)

Then we can also leave out the `null` case; it's covered by the implicit initialization in `Widget? senderRow;`.

A follow-up bonus to using `if` rather than a conditional expression is that we can then have local variables within the `if` block to organize things. In particular, it's then possible to keep a local variable `time`, in the way I suggested in my original review:
https://github.com/zulip/zulip-flutter/pull/446#discussion_r1433448956
rather than inlining it and making the expression lengthier.

lib/widgets/message_list.dart
```diff
@@ -849,61 +852,75 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                          style: const TextStyle(
+                            fontFamily: 'Source Sans 3',
+                            fontSize: 18,
+                            height: (22 / 18),
+                          ).merge(weightVariableTextStyle(context, wght: 600,
+                                    wghtIfPlatformRequestsBold: 900)),
+                          overflow: TextOverflow.ellipsis)),
+                    ]))),
+              Text(
+                _kMessageTimestampFormat.format(
```
gnprice, line 885: This should get a spacer too, in case the sender name is long. Otherwise the name can run right into the timestamp:
image

I think 4px does it.

gnprice (commented):
Great, thanks — with this branch structure, the PR is a lot easier to read and review now.

Comments below, but all of them small; I'm glad this is now getting close to merge.

I looked into the question of the vertical spacing between sender name/avatar and message body that we discussed above at https://github.com/zulip/zulip-flutter/pull/446#discussion_r1430618216, and I believe this version faithfully matches what's in the Figma design. I'm not sure if the behavior here differs from what was in the original revision, or if it's purely a matter of the cleaner branch structure in this revision making it clear what's intended, but either way I'm happy with the result.

sirpengi:

gnprice:
Thanks! All now looks good — merging.

PR #387, opened 2023-11-17
deps: Upgrade Flutter to latest main, 3.17.0-10.0.pre.45

chrisbobbe:
And update Flutter's supporting libraries to match.

PR #447, opened 2023-12-08
Remove unnecessary imports of dart:ui

sirpengi:
Upstream Flutter (in flutter/flutter@c6741614041 ) changed `flutter/material.dart` to additionally export many types from `dart:ui`. This caused a lot of our imports to trigger an analysis warning of `unnecessary_import`.

Please make sure you update your Flutter SDK after pulling in these changes!

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1,5 +1,4 @@
 import 'dart:math';
-import 'dart:ui';
 
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -1,5 +1,4 @@
 import 'dart:io';
-import 'dart:ui';
 import 'package:flutter/widgets.dart';
 
 /// A mergeable [TextStyle] with 'Source Code Pro' and platform-aware fallbacks.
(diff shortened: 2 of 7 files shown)


chrisbobbe:
Thanks!

If you're on a Flutter version that doesn't have that commit, won't the imports still be necessary? I think we should remove the imports alongside a bump of our lower bound on the Flutter version. For how to bump that lower bound, see: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#upgrading-flutter , and for an example, see #387.

I guess it would make sense to do the min-version bump in the same commit as we remove the imports. Removing them becomes possible at the new Flutter version, but it also becomes necessary at the same time, because of the analysis rules we use. For the commit message, I would probably copy the form of a34b9e576, but with a brief note about removing the imports.

sirpengi:
@chrisbobbe thanks for the comment! I've bumped the version and reframed the commit message.

gnprice:
Thanks to you both! Looks good; merging, with small tweaks to the commit message.

(In particular it's not about `flutter/material` but about a lower-level library, `flutter/painting`. In that upstream commit, the key change is [in `packages/flutter/lib/src/painting/basic_types.dart`](https://github.com/flutter/flutter/commit/c6741614041#diff-39ae07d01da16ee2b05a56c9726c6fe5de97e77b1cd5507cff3cfe70f0efa78e). If it were about `flutter/material` then our `lib/widgets/text.dart` wouldn't be affected, because the latter imports only `flutter/widgets` which doesn't depend on `flutter/material`.)

Issue #440, opened 2023-12-06
Message user just sent should not be unread


alya:
When a user sends a message from the app, that message appears as unread (unread marker on the left; unread count in the bottom banner increases). Instead, a message you send should immediately be marked as read.

PR #450, opened 2023-12-13
Send hard-coded User-Agent in requests

sirpengi:
Zulip server will automatically mark messages as read if the request appears to come from a recognized client. We simply need to identify using the `User-Agent` header as an approved client.
See `sent_by_human` in https://github.com/zulip/zulip/blob/main/zerver/models.py

This PR has two commits, the first is a simple hard-coded user agent that will result in the desired behavior. The second calculates the user agent from the device, but this might not be how we want it calculated (espeically since `ZulipBinding` also contains an interface into device info).

A further improvement would be to get upstream to accept `ZulipFlutter` as a legitimate user agent so we aren't reusing the `ZulipMobile` name.

Fixes: #440

### DIFF
--- a/lib/api/core.dart
+++ b/lib/api/core.dart
@@ -80,6 +80,7 @@ class ApiConnection {
     assert(debugLog("${request.method} ${request.url}"));
 
     addAuth(request);
+    request.headers.addAll(userAgentHeader());
 
     final http.StreamedResponse response;
     try {
@@ -195,6 +196,13 @@ String _authHeaderValue({required String email, required String apiKey}) {
 Map<String, String> authHeader({required String email, required String apiKey}) {
   return {
     'Authorization': _authHeaderValue(email: email, apiKey: apiKey),
+  };
+}
+
+// TODO(#453): Create real User-Agent string
+Map<String, String> userAgentHeader() {
+  return {
+    'User-Agent': 'ZulipMobile/?.?.? (Android 14)',
   };
 }
 
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1,5 +1,4 @@
 import 'dart:math';
-import 'dart:ui';
 
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -1,5 +1,4 @@
 import 'dart:io';
-import 'dart:ui';
 import 'package:flutter/widgets.dart';
 
 /// A mergeable [TextStyle] with 'Source Code Pro' and platform-aware fallbacks.
--- a/lib/widgets/unread_count_badge.dart
+++ b/lib/widgets/unread_count_badge.dart
@@ -1,4 +1,3 @@
-import 'dart:ui';
 
 import 'package:flutter/material.dart';
 
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -768,10 +768,10 @@ packages:
     dependency: "direct main"
     description:
       name: path
-      sha256: "8829d8a55c13fc0e37127c29fedf290c102f4e40ae94ada574091fe0ff96c917"
+      sha256: "087ce49c3f0dc39180befefc60fdb4acd8f8620e5682fe2476afd0b3688bb4af"
       url: "https://pub.dev"
     source: hosted
-    version: "1.8.3"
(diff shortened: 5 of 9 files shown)


sirpengi:
It seemed impossible to detangle the broken tests introduced after calling `package_info_plus` from inside of `lib/api/core.dart`. I've readjusted the PR here to have a hard-coded value only and opened #453 to handle crafting that value appropriately.

lib/api/core.dart
```diff
@@ -198,6 +199,13 @@ Map<String, String> authHeader({required String email, required String apiKey})
   };
 }
 
+// TODO(#453): Create real User-Agent string
+Map<String, String> userAgentHeader() {
+  return {
+    'User-Agent': 'ZulipMobile/?.?.? (Android 14)',
```
gnprice, line 205: Let's work in something here to make clear this is actually the Flutter app, not the RN app. E.g., saying "ZulipFlutter".

Tim suggests at https://github.com/zulip/zulip-flutter/issues/453#issuecomment-1856146265 that it's enough for "ZulipMobile" to appear somewhere in the value, not necessarily at the start. Can we add something like "(like ZulipMobile)" at the end instead, and put "ZulipFlutter" at the start?

Or if we do need it at the start, in the venerable tradition of every modern browser claiming to be Mozilla 5.0:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent
then in that same venerable tradition I hope we can at least put the more informative name somewhere later.

gnprice (commented):
Thanks @sirpengi! Comments below.

Since the explicit flag to control the read-by-self behavior is already built, the main thing we can do for #440 is to just use that. That should also be pretty simple, I think.

Then we'll still want to do this as a follow-up, in order to support 7.x and older servers. But we'll no longer need to feel rushed for it.

sirpengi:
#456 was opened for support of the new `read_for_sender` flag added in https://github.com/zulip/zulip/pull/28204 . That should be merged first (when ready) and I will update this PR to go on top of that change.

lib/api/core.dart
sirpengi: Tested with live server and dug through https://github.com/zulip/zulip/blob/main/zerver/lib/user_agent.py . The User-Agent is parsed with this:
```python
pattern = re.compile(
    """^ (?P<name> [^/ ]* [^0-9/(]* )
    (/ (?P<version> [^/ ]* ))?
    ([ /] .*)?
    $""",
    re.X,
)
```

and the whitelist from https://github.com/zulip/zulip/blob/main/zerver/models/clients.py is:
```python
            sending_client
            in (
                "zulipandroid",
                "zulipios",
                "zulipdesktop",
                "zulipmobile",
                "zulipelectron",
                "zulipterminal",
                "snipe",
                "website",
                "ios",
                "android",
            )
            or "desktop app" in sending_client
            # Since the vast majority of messages are sent by humans
            # in Zulip, treat test suite messages as such.
            or (sending_client == "test suite" and settings.TEST_SUITE)
```

To trigger a match we could not start the string with "ZulipFlutter". I think given the options we should present as "ZulipMobile" and stick on a "(like ZulipMobile)" at the end.

We could possibly start with "ZulipFlutter; like desktop app" or "ZulipFlutter(like desktop app)" (but NOT "ZulipFlutter (like desktop app)") but I don't like these options.

gnprice: OK, let's do a string like `ZulipMobile/1.2.3 (Android 14) (actually ZulipFlutter)`. Where instead of "Android 14" we have the actual OS name and version (and we can leave that part out for a first pass), and instead of "1.2.3" we have the actual version (and we can leave that out too for a first pass).

gnprice: In fact, better yet: once we do have the version number, it'll be from a different sequence from the zulip-mobile version numbers anyway. So rather than saying "ZulipMobile/1.2.3", we can let the version number remain absent at that spot and instead say a string like `ZulipMobile (Android 14) (actually ZulipFlutter/1.2.3)`.

chrisbobbe:
With #460 having been prioritized (it has the `beta feedback` label), and Shu being on vacation (till 2024-01-08, from the calendar), Greg requested that I take over this PR, which I'm happy to do. 🙂 Looking now.

gnprice:
Closing this, to continue discussion on the revised version #471. Thanks again @sirpengi!

PR #456, opened 2023-12-19
New `read_for_sender` flag when sending a new message

sirpengi:
PR is built on top of #447 in order for tests to pass.

Adds a new flag introduced in https://github.com/zulip/zulip/pull/28204 .

Fixes: #440

### DIFF
--- a/lib/api/route/messages.dart
+++ b/lib/api/route/messages.dart
@@ -177,6 +177,7 @@ Future<SendMessageResult> sendMessage(
   required String content,
   String? queueId,
   String? localId,
+  bool? readBySender,
 }) {
   final supportsTypeDirect = connection.zulipFeatureLevel! >= 174; // TODO(server-7)
   return connection.post('sendMessage', SendMessageResult.fromJson, 'messages', {
@@ -193,6 +194,7 @@ Future<SendMessageResult> sendMessage(
     'content': RawParameter(content),
     if (queueId != null) 'queue_id': queueId, // TODO should this use RawParameter?
     if (localId != null) 'local_id': localId, // TODO should this use RawParameter?
+    if (readBySender != null) 'read_by_sender': readBySender,
   });
 }
 
--- a/lib/api/route/messages.dart
+++ b/lib/api/route/messages.dart
@@ -191,8 +191,8 @@ Future<SendMessageResult> sendMessage(
       throw Exception('impossible destination') // TODO(dart-3) show this statically
     ),
     'content': RawParameter(content),
-    if (queueId != null) 'queue_id': queueId,
-    if (localId != null) 'local_id': localId,
+    if (queueId != null) 'queue_id': queueId, // TODO should this use RawParameter?
+    if (localId != null) 'local_id': localId, // TODO should this use RawParameter?
   });
 }
 
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -377,13 +377,13 @@ class PerAccountStore extends ChangeNotifier with StreamStore {
     }
   }
 
-  Future<void> sendMessage({required MessageDestination destination, required String content, bool? readBySender}) {
+  Future<void> sendMessage({required MessageDestination destination, required String content}) {
     // TODO implement outbox; see design at
     //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3881.20Sending.20outbox.20messages.20is.20fraught.20with.20issues/near/1405739
     return _apiSendMessage(connection,
       destination: destination,
       content: content,
-      readBySender: readBySender,
+      readBySender: true,
     );
   }
 
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -377,10 +377,14 @@ class PerAccountStore extends ChangeNotifier with StreamStore {
     }
   }
 
-  Future<void> sendMessage({required MessageDestination destination, required String content}) {
+  Future<void> sendMessage({required MessageDestination destination, required String content, bool? readBySender}) {
     // TODO implement outbox; see design at
     //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3881.20Sending.20outbox.20messages.20is.20fraught.20with.20issues/near/1405739
-    return _apiSendMessage(connection, destination: destination, content: content);
+    return _apiSendMessage(connection,
+      destination: destination,
+      content: content,
+      readBySender: readBySender,
+    );
   }
 
   static List<CustomProfileField> _sortCustomProfileFields(List<CustomProfileField> initialCustomProfileFields) {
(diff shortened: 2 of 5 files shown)


sirpengi:
For the time being, CZO is still showing FL 226 (that is, the changes in the linked PR have not been deployed yet). So this might be premature to merge at the moment as I haven't confirmed the functionality with a live server.

gnprice:
> For the time being, CZO is still showing FL 226 (that is, the changes in the linked PR have not been deployed yet).

Huh — that's surprising, because IIUC those server changes went into Zulip Server 8.0. And I'd certainly expect anything released in 8.0 to have already been deployed on CZO and on Zulip Cloud before that release happened (which was at the end of last week).

In any case, it certainly seems to be deployed now. Looking at https://chat.zulip.org/api/changelog , it shows this change as FL 236, and as the latest API change deployed:
image

(Same story on zulip.com, i.e. Zulip Cloud.)

So this should be ripe for testing against a live server.

test/api/route/messages_test.dart
```diff
@@ -314,6 +314,20 @@ void main() {
...
+            'type': 'stream',
+            'to': streamId.toString(),
+            'topic': topic,
+            'content': content,
+            'read_by_sender': 'true',
+          });
+      });
+    });
+
+    test('to server without read_by_sender support', () {
```
gnprice, line 329: Conveniently, the Zulip server actually ignores extra parameters sent in a request. So we can simplify by always sending this parameter, without checking whether the server is new enough to understand it.

lib/api/route/messages.dart
```diff
@@ -193,6 +194,7 @@ Future<SendMessageResult> sendMessage(
     'content': RawParameter(content),
     if (queueId != null) 'queue_id': queueId,
     if (localId != null) 'local_id': localId,
+    if (supportsReadBySender) 'read_by_sender': true,
```
gnprice, line 197: Rather than hardcoding this to true within the API bindings, let's make it a parameter of this function, and then the caller passes true for it.

That way, the binding remains a thin layer exposing basically the same API as the server does itself, just in Dart. That in turn means the call sites in the application code (the code outside `lib/api/`) can be read more or less directly against the Zulip Server API. I think that's helpful for reducing the number of layers one needs to look through and making it easier to understand what the code is doing.

(This function does have one significant deviation from the underlying API, in the `destination` argument. What's happening there is basically that this part of the Zulip Server API isn't written in a way that's readily compatible with good clear static types, and so some rearrangement is needed in order to do a solid job of translating the API into Dart. But hard-coding this parameter is a different kind of matter, which is more about the specific needs and use cases we expect our application to have rather than about translating into Dart.)

gnprice (commented):
Thanks @sirpengi! Two comments below.

Let's also test this PR live against a server that has this feature — either chat.zulip.org or Zulip Cloud should work, as discussed above.

sirpengi:
@gnprice updated PR as per your comments and have checked this against CZO (I looked into the issue with the outdated FL and posted about it in CZO: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/zulipFeatureLevel.20out.20of.20sync )

lib/model/store.dart
```diff
@@ -392,10 +392,10 @@ class PerAccountStore extends ChangeNotifier {
     }
   }
 
-  Future<void> sendMessage({required MessageDestination destination, required String content}) {
+  Future<void> sendMessage({required MessageDestination destination, required String content, bool? readBySender}) {
     // TODO implement outbox; see design at
     //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3881.20Sending.20outbox.20messages.20is.20fraught.20with.20issues/near/1405739
-    return _apiSendMessage(connection, destination: destination, content: content);
+    return _apiSendMessage(connection, destination: destination, content: content, readBySender: readBySender);
```
gnprice, line 398: nit: line too long

lib/widgets/compose_box.dart
```diff
@@ -733,7 +733,7 @@ class _SendButtonState extends State<_SendButton> {
 
     final store = PerAccountStoreWidget.of(context);
     final content = widget.contentController.textNormalized;
-    store.sendMessage(destination: widget.getDestination(), content: content);
+    store.sendMessage(destination: widget.getDestination(), content: content, readBySender: true);
```
gnprice, line 736: nit: line too long

test/api/route/messages_test.dart
```diff
@@ -304,7 +304,7 @@ void main() {
     }) async {
       connection.prepare(json: SendMessageResult(id: 42).toJson());
       final result = await sendMessage(connection,
-        destination: destination, content: content);
+        destination: destination, content: content, readBySender: true);
```
gnprice, line 307: I'd like to write this test a little differently; but probably the simplest way to express how is by just writing it. So I'll do that before merging.

(And in fact, when I go to write the style of test I'm thinking of, I discover a pre-existing behavior that's likely wrong! So I'll end up with a prep commit adding that test and a comment, then your main commit tweaked to add this new parameter to that test.)

lib/model/store.dart
```diff
@@ -392,10 +392,10 @@ class PerAccountStore extends ChangeNotifier {
     }
   }
 
-  Future<void> sendMessage({required MessageDestination destination, required String content}) {
+  Future<void> sendMessage({required MessageDestination destination, required String content, bool? readBySender}) {
```
gnprice, line 395: So in contrast to the API binding layer, this method on PerAccountStore I think actually is a good place to hard-code this parameter to true. I'll just add a commit on top making that change.

Here's the commit message I wrote for that commit, explaining why:

    send [nfc]: Always pass readBySender true
    
    I think we're always going to want this parameter to be true
    for all the messages we send from the app.
    
    In the API bindings themselves, i.e. in lib/api/ , I generally
    prefer not to hard-code this sort of thing.  That lets the API
    bindings be a faithful reflection of the Zulip Server API, with
    a minimum of ad-hoc specializations for the needs of our app.
    So that's why the `sendMessage` binding just takes the parameter
    from its caller.
    
    But this method PerAccountStore.sendMessage is already something
    specific to our app rather than an API binding; it doesn't do much
    now beyond wrap the API, but the reason it exists is to be the
    future home of our "outbox" logic.  That means (a) recording the
    message locally so that we can display it immediately, (b) possibly
    retrying the send, and (c) if the send fails or times out, recording
    that fact so that our UI can present the user with that information
    and the option to retry.
    
    So that makes it a natural home for encapsulating any other logic
    that this app will always want when sending a message.
    In particular, passing this parameter.

gnprice (commented):
Thanks for the revision!

Generally this looks good; a few comments below. Some are probably to explain by showing rather than telling (and others are trivial, line length); so I'll fix these up myself, including a couple of additional commits, and merge.

Issue #327, opened 2023-10-18
Sticky header hit-tests at bottom of message list, though paints at top


gnprice:
In the message list, if you tap near the very *bottom* of the viewport, you'll get navigated to the conversation or stream corresponding to the message at the *top*.

(I discovered this just now while reviewing #318. It had me very confused for a while because the effect is the same as following a certain internal link would be… but the issue is pre-existing and unrelated to #318.)

From a bit of debugging, what's happening is that the `GestureDetector` that belongs to a recipient header is firing — specifically the one for the recipient header that's currently sticky, shown at the top of the viewport.

Presumably this is a bug in our own `sticky_header.dart`. It'd probably be in the implementation of `hitTestChildren`.

It looks like the tests I wrote don't ever attempt to tap on the list items, or the headers, to check that the gesture gets routed to the right place. They go through a comprehensive set of scenarios with the different options, and with scrolling to different positions, but only inspect the reported layout and not the response to gestures. So one thing to include in a fix for this issue will be to add some gestures to those tests.

chrisbobbe:
I assume this is the bug I've run into a few times, when tapping the new "Mark X as read" button from #364. That button is at the bottom.

gnprice:
Yeah, sounds right. I guess that button raises the visibility of this issue. This is already probably the first issue I'll pick up myself after Beta 1 is out, so I guess this just reinforces that priority.

Until then, the workaround is that you tap on the upper half or so of the mark-as-read button — the button is taller than the recipient headers, so only the lower portion of it will be covered by this ghost of the sticky header.

PR #468, opened 2023-12-29
sticky_header: Fix hit-testing of header

gnprice:
Fixes: #327

Also enable a bit of test config that I ended up not relying on, but that should be enabled in any case.


chrisbobbe (approved):
Thanks, this is great to fix! Merging.

Issue #173, opened 2023-06-09
Date separators


gnprice:
When successive messages span a date boundary, we should show a separator indicating the new date, like in Zulip web and zulip-mobile.

This complements the dates in recipient headers (#172) by highlighting when the answer changes. (Plus, when we start properly suppressing repeat recipient headers, it makes it possible to do so spanning dates without becoming misleading.)

gnprice:
In the implementation we merged for #174, we effectively included date separators by having there be a new recipient header when the date changes (even if the recipient didn't).

It'd still be nice to have a distinct design for purely separating dates, like we do in Zulip web and zulip-mobile.

gnprice:
Early beta feedback reports that our current stopgap use of additional recipient headers as date separators is "distracting":
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flutter.20feedback/near/1700169

So bumping up to Beta 2.

gnprice:
@terpimost I'm not finding any examples of date separators when I browse the existing Figma design. So I'll base this initially off of what's in web, but I'd be glad to adjust later if you have other thoughts.

PR #469, opened 2023-12-29
Add date separators

gnprice:
Fixes: #173

This is a draft because it needs tests and also some cleanups in the code. But from a user perspective it already works great — so I'm sharing this in order to get feedback on the visual design. As mentioned at https://github.com/zulip/zulip-flutter/issues/173#issuecomment-1871744164 , I don't see an existing design for this in Figma, so for a first draft I've taken what's in web and adapted it.

## Design questions

* In web the dates have color `hsl(0deg 0% 15% / 75%)`, which is the same as the dates in the recipient headers. Here, our dates in recipient headers are lighter than that: they're black with 40% alpha. So I used that color for these, though it makes the dates lighter than in web. @terpimost do you think this is the color we want?
* In web the dates are slightly smaller in the date separators than they are in the recipient headers. But the CSS that results in this on the recipient headers side looks like a bug: there's both a `calc(12em / 14)` and a `calc(15em / 14)` applying multiplicatively. (This bug illustrates why in the CSS for zulip-mobile's webview, [we use `rem` and not `em`](https://github.com/zulip/zulip-mobile/blob/main/docs/background/webview.md#why-rem-and-not-em).) So I'm not sure that difference is intended, and in this draft I've made the size the same as in the recipient headers.
* ~~The vertical alignment here is just that everything's centered. But that seems to get the dividers a bit too high compared to the text. I'll have to fiddle with this (perhaps by adjusting the text's line-height).~~ I fixed this by adding some bottom padding to the text.

## Other adjustments from web

* In web the dates and timestamps are 10px from the right, but in the current design they're 16px in. So I put this at 16px to match.
* In web there's 2px horizontal padding before and after each divider line (though web's implementation of that is a bit complex). I copied that.
* I made the dividers one physical pixel tall (not one logical pixel, which would typically be several physical pixels).
* I made the dividers black, like in web.

## Screenshot

Here's a screenshot of the current revision:

Screenshot_1703883619

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -25,6 +25,12 @@ class MessageListRecipientHeaderItem extends MessageListItem {
   MessageListRecipientHeaderItem(this.message);
 }
 
+class MessageListDateSeparatorItem extends MessageListItem {
+  final Message message;
+
+  MessageListDateSeparatorItem(this.message);
+}
+
 /// A message to show in the message list.
 class MessageListMessageItem extends MessageListItem {
   final Message message;
@@ -118,6 +124,7 @@ mixin _MessageSequence {
           case MessageListDirection.older:       return -1;
         }
       case MessageListRecipientHeaderItem(:var message):
+      case MessageListDateSeparatorItem(:var message):
         return (message.id <= messageId) ? -1 : 1;
       case MessageListMessageItem(:var message): return message.id.compareTo(messageId);
     }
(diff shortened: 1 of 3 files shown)


gnprice:
I've pushed a revision that adjusts the vertical alignment between the text and the divider lines.

I also edited the PR description with another design question, one I'd forgotten about when I first posted the PR, about the font size for the dates compared to the dates in recipient headers.

gnprice:
OK, now actually ready.

There are still a couple of design questions open (mentioned above in the PR description), but they're pretty subtle — so I think the best thing is to go ahead and merge without worrying about those, and we can always adjust later if @terpimost has comments once he's back.

chrisbobbe:
Thanks, LGTM! Merged.

PR #383, opened 2023-11-16
api: Handle surprise removal of email_address from Subscription

chrisbobbe:
This API change was made in Zulip Server 7.5, without discussion of a deprecation period first.

A process failure, but fortunately easy to recover from, since this project is still in alpha. (And zulip-mobile is unaffected by the change.)

Discussion:
  https://chat.zulip.org/#narrow/stream/378-api-design/topic/Removing.20.60email_address.60.20from.20subscription.20objects.3F/near/1683612

PR #470, opened 2023-12-29
api: On parse failure, include details in exception

gnprice:
If the server's response is malformed, as with the server issue that led to #383, the existing logging is unhelpful for working out what the problem is, even in debug: we learn what request failed, but not where in the response the misparse was.

When that issue happened, I debugged it with a cruder version of this change, just augmenting this `catch` block to print the exception and its stack trace.  But we can do better, using the magic of Error.throwWithStackTrace.


chrisbobbe:
Cool! Glad to have this; merging.

Issue #460, opened 2023-12-22
Login notifications say "An unknown browser on an unknown operating system"


gnprice:
When you log in through the Flutter app, the email the server sends you to say there was a new login has the line:
> Device: An unknown browser on an unknown operating system.

I believe this will be fixed by #406. But after we've done that, we should check this is fixed and adjust as necessary.

Issue #406, opened 2023-11-21
Set the HTTP user agent


gnprice:
Currently when we make an HTTP request, the `User-Agent` header we send is just whatever the default is that Dart's HTTP implementation gives us. This means e.g. that log lines on the server look like this, saying "via Dart":

```
2023-11-21 22:32:07.353 INFO [zr] 192.168.0.188   POST    200 113ms (db: 35ms/41q) /api/v1/register (11@zulip via Dart)
```

Instead we should set a user-agent string that identifies this app. Ideally it should include the app's version number.

There are at least two different kinds of HTTP requests the app makes, both of which should have the user agent set:
* Zulip API calls, which go through `ApiConnection.send`;
* fetching images, including avatars, image emoji, and images included in messages, all of which go through `RealmContentNetworkImage`.

It's likely those will need to be handled separately.

gnprice:
The value should probably say "ZulipMobile" somewhere, because apparently (https://github.com/zulip/zulip-flutter/issues/453#issuecomment-1856146265) for send-message requests that will activate the server's heuristic for deciding we're a UI client and therefore the newly sent message should be marked as read for its author, solving #440.

For some discussion on how to implement this, see #453.

gnprice:
> The value should probably say "ZulipMobile" somewhere

But this should be marked with a `TODO(server-8)` comment — once we're supporting only Zulip Server 8+, our use of `read_by_sender` (from #456) should always get respected, so we should no longer need to cater to that old heuristic and we can use a more transparent user-agent of our choice.

Issue #467, opened 2023-12-28
In user-agent, include platform, platform version, and app version


chrisbobbe:
In the first pass for

- #406

, Greg [suggests](https://github.com/zulip/zulip-flutter/pull/450#discussion_r1433366608) leaving out the platform, platform version, and app version. So I expect the user agent will look something like 

```
ZulipMobile (actually ZulipFlutter)
```

instead of e.g.

```
ZulipMobile (Android 14) (actually ZulipFlutter/1.2.3)
```

. This issue is for filling in those missing things.

gnprice:
In particular, after #471 there will be (as discussed at https://github.com/zulip/zulip-flutter/pull/471#issuecomment-1874756255) a sort of residual version of the following issue:
* #460

which is that the login emails will say "Device: Zulip on an unknown operating system" rather than the actual operating system.

Fixing this issue should fix that. When we have a PR for this issue, one thing we should manually test in it is that it does indeed fix that.

PR #471, opened 2023-12-30
api: Send simple User-Agent header

chrisbobbe:
Revises: #450
Fixes: #406
Fixes: #460

### DIFF
--- a/lib/api/core.dart
+++ b/lib/api/core.dart
@@ -80,6 +80,7 @@ class ApiConnection {
     assert(debugLog("${request.method} ${request.url}"));
 
     addAuth(request);
+    request.headers.addAll(userAgentHeader());
 
     final http.StreamedResponse response;
     try {
@@ -198,6 +199,13 @@ String _authHeaderValue({required String email, required String apiKey}) {
 Map<String, String> authHeader({required String email, required String apiKey}) {
   return {
     'Authorization': _authHeaderValue(email: email, apiKey: apiKey),
+  };
+}
+
+Map<String, String> userAgentHeader() {
+  return {
+    // TODO(#467) include platform, platform version, and app version
+    'User-Agent': 'ZulipFlutter',
   };
 }
 
--- a/lib/api/core.dart
+++ b/lib/api/core.dart
@@ -74,13 +74,16 @@ class ApiConnection {
   bool _isOpen = true;
 
   Future<T> send<T>(String routeName, T Function(Map<String, dynamic>) fromJson,
-      http.BaseRequest request) async {
+      http.BaseRequest request, {String? overrideUserAgent}) async {
     assert(_isOpen);
 
     assert(debugLog("${request.method} ${request.url}"));
 
     addAuth(request);
     request.headers.addAll(userAgentHeader());
+    if (overrideUserAgent != null) {
+      request.headers['User-Agent'] = overrideUserAgent;
+    }
 
     final http.StreamedResponse response;
     try {
@@ -137,13 +140,13 @@ class ApiConnection {
   }
 
   Future<T> post<T>(String routeName, T Function(Map<String, dynamic>) fromJson,
-      String path, Map<String, dynamic>? params) async {
+      String path, Map<String, dynamic>? params, {String? overrideUserAgent}) async {
     final url = realmUrl.replace(path: "/api/v1/$path");
     final request = http.Request('POST', url);
     if (params != null) {
       request.bodyFields = encodeParameters(params)!;
     }
-    return send(routeName, fromJson, request);
+    return send(routeName, fromJson, request, overrideUserAgent: overrideUserAgent);
   }
 
   Future<T> postFileFromStream<T>(String routeName, T Function(Map<String, dynamic>) fromJson,
--- a/test/api/route/messages_test.dart
+++ b/test/api/route/messages_test.dart
@@ -339,11 +339,13 @@ void main() {
       return FakeApiConnection.with_((connection) async {
         await checkSendMessage(connection,
           destination: StreamDestination(streamId, topic), content: content,
+          readBySender: true,
           expectedBodyFields: {
             'type': 'stream',
             'to': streamId.toString(),
             'topic': topic,
             'content': content,
+            'read_by_sender': 'true',
           });
       });
     });
@@ -352,10 +354,12 @@ void main() {
       return FakeApiConnection.with_((connection) async {
         await checkSendMessage(connection,
           destination: DmDestination(userIds: userIds), content: content,
+          readBySender: true,
           expectedBodyFields: {
             'type': 'direct',
             'to': jsonEncode(userIds),
             'content': content,
+            'read_by_sender': 'true',
           });
       });
     });
@@ -364,10 +368,12 @@ void main() {
       return FakeApiConnection.with_(zulipFeatureLevel: 173, (connection) async {
         await checkSendMessage(connection,
           destination: DmDestination(userIds: userIds), content: content,
+          readBySender: true,
           expectedBodyFields: {
             'type': 'private',
             'to': jsonEncode(userIds),
             'content': content,
+            'read_by_sender': 'true',
           });
       });
     });
--- a/test/api/route/messages_test.dart
+++ b/test/api/route/messages_test.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'package:checks/checks.dart';
 import 'package:http/http.dart' as http;
 import 'package:test/scaffolding.dart';
+import 'package:zulip/api/core.dart';
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/model/narrow.dart';
 import 'package:zulip/api/route/messages.dart';
@@ -304,6 +305,7 @@ void main() {
       String? localId,
       bool? readBySender,
       required Map<String, String> expectedBodyFields,
+      String? expectedUserAgent,
     }) async {
       connection.prepare(json: SendMessageResult(id: 42).toJson());
       final result = await sendMessage(connection,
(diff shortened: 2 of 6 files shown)


lib/api/core.dart
```diff
@@ -198,6 +199,21 @@ Map<String, String> authHeader({required String email, required String apiKey})
...
 
+Map<String, String> userAgentHeader() {
+  return {
+    // Putting "ZulipMobile" in the string helps users on pre-8 servers:
+    //   - When a user sends a message from the app,
+    //     the message is read instead of unread, avoiding #440.
+    //     (With servers 8+, this is accomplished
+    //     with a new param in the send-message API; see #456.)
+    //   - Login notifications say something specific instead of
+    //     "an unknown browser on an unknown operating system", avoiding #460.
```
gnprice, line 210: Does this latter part work if we say "ZulipFlutter" instead of "ZulipMobile"? (If not, then we might need further server changes but we'll want to figure those out.)

test/widgets/content_test.dart
```diff
@@ -252,20 +257,20 @@ void main() {
...
-      return headers['Authorization']?.single;
+      return httpClient.request.headers.values;
     }
 
     testWidgets('includes auth header if `src` on-realm', (tester) async {
-      check(await actualAuthHeader(tester, Uri.parse('https://chat.example/image.png')))
-        .isNotNull().equals(authHeaders['Authorization']!);
+      check(await actualHeaders(tester, Uri.parse('https://chat.example/image.png')))
+        ..['Authorization'].single.equals(authHeaders['Authorization']!)
+        ..['User-Agent'].single.equals(userAgentHeader()['User-Agent']!);
```
gnprice, line 266: Rather than check only these two headers, I think I'd prefer to continue to have a check that there aren't any unexpected additional headers. That can be done by a `deepEquals` on the headers map.

gnprice (commented):
Thanks @chrisbobbe for picking this up! I edited the description to cross-reference it with the previous PR #450.

The code generally all looks good; small comments below.

What manual testing have you done to confirm it resolves #460, and resolves #440 for older servers? (For the latter, I guess it'd suffice to test on a current server but without passing `read_by_sender`.)

chrisbobbe:
> manual testing have you done to confirm it resolves https://github.com/zulip/zulip-flutter/issues/460

Oh, hmm, doing this now…

Visiting `/emails/` on my dev server, this is what I see when I log into an account from zulip-flutter before these PR commits:

> Device: An unknown browser on an unknown operating system.

And this is what I see at the tip of the PR branch:

> Device: Zulip on an unknown operating system.

Unfortunate that it doesn't even say Android or iOS, but it makes sense; in this first pass, the user-agent header doesn't carry that information (#467).

lib/api/core.dart
chrisbobbe: Yes, I just tested manually by changing the string from 'ZulipMobile (actually ZulipFlutter)' to just 'ZulipFlutter' and seeing what the login-notification email said, by visiting `/emails` on the dev server. It said:

> Device: Zulip on an unknown operating system.

, the same as it does with 'ZulipMobile (actually ZulipFlutter)' (https://github.com/zulip/zulip-flutter/pull/471#issuecomment-1874756255).

gnprice:
> Unfortunate that it doesn't even say Android or iOS, but it makes sense; in this first pass, the user-agent header doesn't carry that information (#467).

Cool. I think the change from "An unknown browser" to "Zulip" is enough to say this resolves #460 — I think that "unknown browser" is probably the most disconcerting part.

There'll still be a sort of followup to #460 which is to fix the operating-system part. But we can track that as just a note on #467 of something to test in a PR that fixes it.

chrisbobbe:
Ah hmm, well here's a problem. The string used in this PR, `'ZulipMobile (actually ZulipFlutter)'`, doesn't actually parse:

https://github.com/zulip/zulip/blob/08bee0f1809a8a6a074f77d40fd67e10f665377e/zerver/lib/user_agent.py
```python
pattern = re.compile(
    """^ (?P<name> [^/ ]* [^0-9/(]* )
    (/ (?P<version> [^/ ]* ))?
    ([ /] .*)?
    $""",
    re.X,
)
```

gnprice:
> As a consequence, `/register` is erroring.

That seems pretty busted. The server will fail to respond to a request if its bundle of heuristics for interpreting user-agent strings doesn't understand your user-agent?

chrisbobbe:
```
2024-01-03 02:31:42.019 ERR  [django.request] Internal Server Error: /api/v1/register
Traceback (most recent call last):
  File "/srv/zulip/zerver/middleware.py", line 402, in process_exception
    raise exception  # Ensure correct sys.exc_info().
  File "/srv/zulip-py3-venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/srv/zulip/zerver/lib/rest.py", line 39, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/srv/zulip-py3-venv/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
  File "/srv/zulip/zerver/lib/rest.py", line 203, in rest_dispatch
    return target_function(request, **kwargs)
  File "/srv/zulip-py3-venv/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
  File "/srv/zulip/zerver/decorator.py", line 779, in _wrapped_func_arguments
    raise err
  File "/srv/zulip/zerver/decorator.py", line 776, in _wrapped_func_arguments
    return view_func(request, user_profile, *args, **kwargs)
  File "/srv/zulip/zerver/lib/request.py", line 464, in _wrapped_req_func
    return_value = req_func(request, *args, **kwargs)
  File "/srv/zulip/zerver/views/events_register.py", line 133, in events_register_backend
    pronouns_field_type_supported = is_pronouns_field_type_supported(
  File "/srv/zulip/zerver/lib/compatibility.py", line 157, in is_pronouns_field_type_supported
    if version_lt(user_agent["version"], FIRST_VERSION_TO_SUPPORT_PRONOUNS_FIELD):
  File "/srv/zulip/zerver/lib/compatibility.py", line 69, in version_lt
    num1, rest1 = pop_numerals(ver1)
  File "/srv/zulip/zerver/lib/compatibility.py", line 46, in pop_numerals
    match = re.search(r"^( \d+ (?: \. \d+ )* ) (.*)", ver, re.X)
  File "/usr/lib/python3.8/re.py", line 201, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or bytes-like object
```

Aha:

```python
def is_pronouns_field_type_supported(user_agent_str: Optional[str]) -> bool:
    # In order to avoid users having a bad experience with these
    # custom profile fields disappearing after applying migration
    # 0421_migrate_pronouns_custom_profile_fields, we provide this
    # compatibility shim to show such custom profile fields as
    # SHORT_TEXT to older mobile app clients.
    #
    # TODO/compatibility(7.0): Because this is a relatively minor
    # detail, we can remove this compatibility hack once most users
    # have upgraded to a sufficiently new mobile client.
    if user_agent_str is None:
        return True

    user_agent = parse_user_agent(user_agent_str)
    if user_agent["name"] != "ZulipMobile":
        return True

    FIRST_VERSION_TO_SUPPORT_PRONOUNS_FIELD = "27.192"
    if version_lt(user_agent["version"], FIRST_VERSION_TO_SUPPORT_PRONOUNS_FIELD):
        return False

    return True
```

gnprice:
Hmm, I see.

Let's discuss in `#api design` how to pick a value that navigates the various server-side hacks:
https://chat.zulip.org/#narrow/stream/378-api-design/topic/zulip-flutter.20user.20agent/near/1712186

lib/api/core.dart
chrisbobbe: Er, and I've just tested this on a dev server at 7.x (zulip/zulip@c4a826461b5d8534ad4653a289623ec7f5c4e70f), and got the same result. I realized that my original test here must have been done on an 8+ dev server.

chrisbobbe:
Revision pushed, following CZO discussion: https://chat.zulip.org/#narrow/stream/378-api-design/topic/zulip-flutter.20user.20agent/near/1712928

> What manual testing have you done to confirm it resolves #460, and resolves #440 for older servers? (For the latter, I guess it'd suffice to test on a current server but without passing `read_by_sender`.)

See: https://github.com/zulip/zulip-flutter/pull/471#discussion_r1454735431

Also, while I had that 7.x dev server handy, I went ahead and tested the 'ZulipMobile/flutter' override on send-message requests, as another way to confirm that it fixes #440 for older servers. With the override, the messages were marked as read; without it, they were not marked as read.

test/api/route/messages_test.dart
```diff
@@ -364,11 +371,45 @@ void main() {
...
+        await checkSendMessage(connection,
+          destination: StreamDestination(streamId, topic), content: content,
+          expectedBodyFields: {
+            'type': 'stream',
+            'to': streamId.toString(),
+            'topic': topic,
+            'content': content,
+          },
+          readBySender: null,
+          expectedUserAgent: 'ZulipMobile/flutter');
```
gnprice, line 396: nit:
```suggestion
          readBySender: null,
          expectedBodyFields: {
            'type': 'stream',
            'to': streamId.toString(),
            'topic': topic,
            'content': content,
          },
          expectedUserAgent: 'ZulipMobile/flutter');
```

because `readBySender` specifies what to do, and so comes logically before what to expect from the result

gnprice (commented):
Thanks for the revision! This all looks good, with one nit below.

Since you're out today, I'll just fix that and merge.

Issue #461, opened 2023-12-23
Trust user-added CAs on Android


gnprice:
This is a configuration change that was requested a few times over the years for zulip-mobile, as https://github.com/zulip/zulip-mobile/issues/3312, before we got a contribution implementing it as https://github.com/zulip/zulip-mobile/pull/5493 . We should do the same thing here.

The implementation for zulip-mobile (see https://github.com/zulip/zulip-mobile/commit/85c3a716a7bbf00c936beaf34bc84a0e6bf5c885) is entirely at the Android metadata layer, so we should be able to pretty much just copy-paste that.

gnprice:
This feature is in [the new beta release v0.0.9](https://chat.zulip.org/#narrow/stream/1-announce/topic/mobile.20beta/near/1716421). @shrizza please give it a try — I'd be glad to have confirmation of whether it works for you.

gnprice:
Seems like that version doesn't yet work — but there is hope that a small further tweak to the manifest will make it work. Details in chat:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/flutter.3A.20user-added.20certs/near/1716835

gnprice:
Rereading the thread, it looks like the next step is to make that small further tweak to our manifest, and then put that in the next beta release so @shrizza can try it out. I think we have another beta release (v0.0.15) coming up this week, so I'll aim to include this there.

(The last round of this was just before I went travelling for part of a week and then got covid; I guess by the time I was back at full speed, this had fallen off my radar.)

shrizza:
I appreciate the proactive follow-up, and look forward to the next release.

gnprice:
Well, unfortunately it looks like I'm going to have to revert the take-2 fix I made as #671.

After that PR, in a release build on Android, the app crashes at startup with SIGSEGV and a stack trace like:
```
#00 pc 00000000006b39e8  /apex/com.android.art/lib64/libart.so (art::JNI<false>::GetStringChars(_JNIEnv*, _jstring*, unsigned char*)+232) (BuildId: 7ece79c15d80914c83e60c9e93ac1684)
#01 pc 00000000000ddbd8  /system/lib64/libandroid_runtime.so (android::android_content_XmlBlock_nativeGetAttributeIndex(_JNIEnv*, _jobject*, long, _jstring*, _jstring*)+72) (BuildId: 2c0c526c96b2f0f37ce26c968792b5fe)
[...]

PR #474, opened 2024-01-04
android: Trust user-supplied CAs, except on known servers

gnprice:
This is copied from the corresponding config in zulip-mobile, which was added in the following commit:
  https://github.com/zulip/zulip-mobile/commit/85c3a716a7bbf00c936beaf34bc84a0e6bf5c885

Unlike in zulip-mobile, though, we don't need any adjustments in debug (or profile) builds, because we don't use the `android:usesCleartextTraffic` attribute which this would clobber.

(React Native debug builds reach out to the development host over cleartext HTTP in order to support hot updates; but the Flutter development tools have the host contact the app instead, using the Dart VM protocol.  It's also not clear that any Dart libraries respect the `android:usesCleartextTraffic` flag anyway.)

Fixes: #461

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -11,6 +11,7 @@
             android:launchMode="singleTop"
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
+            android:networkSecurityConfig="@xml/network_security_config"
             android:hardwareAccelerated="true"
             android:windowSoftInputMode="adjustResize">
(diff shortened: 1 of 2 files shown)


gnprice:
I've tested that the resulting app works normally (in a debug build), but haven't attempted to test that it actually resolves #461. Just like in zulip-mobile, then (https://github.com/zulip/zulip-mobile/pull/5493#issuecomment-1298009071), we'll rely in the first instance on the fact that this file is pretty short and the docs seem clear about its semantics — plus it's copied from one that's been working in zulip-mobile.

@shrizza I'll be glad for you to try this when it goes out in an upcoming zulip-flutter beta release, probably in the next week or so — please let us know then whether it does indeed make things work for you.

chrisbobbe:
LGTM; please merge after rebasing to fix the tests.

Issue #475, opened 2024-01-04
Flutter update breaks link interaction tests


gnprice:
After a Flutter update, some tests are failing in main. This is visible also in the recent PRs #472 and #474, as here:
https://github.com/zulip/zulip-flutter/actions/runs/7391944552/job/20109637276?pr=472

The failure looks like this:
```
$ flutter test test/widgets/content_test.dart --name 'can tap a link'
00:02 +0: LinkNode interactions can tap a link to open URL (variant: TargetPlatform.android)
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure was thrown running a test:
Expected: a List<({LaunchMode mode, Uri url})> that:
  has single element
Actual: []
Which: has no elements

When the exception was thrown, this was the stack:
#0      check.<anonymous closure> (package:checks/src/checks.dart:72:9)
#1      _TestContext.nest (package:checks/src/checks.dart:760:12)
#2      IterableChecks.single (package:checks/src/extensions/iterable.dart:33:36)
#3      main.<anonymous closure>.<anonymous closure> (file:///home/greg/z/flutterz/test/widgets/content_test.dart:90:10)
<asynchronous suspension>
#4      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:168:15)
<asynchronous suspension>
#5      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1017:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

The test description was:
  can tap a link to open URL (variant: TargetPlatform.android)
════════════════════════════════════════════════════════════════════════════════════════════════════
00:02 +0 -1: LinkNode interactions can tap a link to open URL (variant: TargetPlatform.android) [E]
  Test failed. See exception logs above.
  The test description was: can tap a link to open URL (variant: TargetPlatform.android)
  

To run this test again: /home/greg/n/flutter/flutter/bin/cache/dart-sdk/bin/dart test /home/greg/z/flutterz/test/widgets/content_test.dart -p vm --plain-name 'LinkNode interactions can tap a link to open URL (variant: TargetPlatform.android)'
```

The test renders some content that has a link in it, then taps the link and checks that the tap made it to the binding for `launchUrl`. The check fails — apparently the tap didn't make it to the binding.

[...]

Issue #214, opened 2023-07-07
Tappable links' touch target can extend through whitespace


chrisbobbe:
If you have a short paragraph ending in a link, the space between the link and the end margin is also tappable. Shown here in the app as built for macOS, so you can see the tappable area when the cursor changes on hover:

![Jul-07-2023 13-32-51](https://github.com/zulip/zulip-flutter/assets/22248748/d790c20b-0218-4696-97dd-2310263bb3da)

gnprice:
Wacky! Thanks for the report. Copying what I said at https://github.com/zulip/zulip-flutter/pull/204#issuecomment-1624225216 :

A variation on that scenario would be when the paragraph wraps over multiple lines, and the last line is short and ends with a link. I'd guess the same bug would reproduce there.

It sounds like the last TextSpan's `recognizer` is inappropriately being applied through the end of the line when the TextSpan visually doesn't extend that far. So this may be an upstream Flutter bug.

chrisbobbe:
Ah indeed; the "short" part of "[i]f you have a short paragraph ending in a link" is not required. 🙂

gnprice:
It looks like this was the following upstream issue:
* https://github.com/flutter/flutter/issues/131435

That was closed by https://github.com/flutter/flutter/pull/140621 (the same upstream change that induced #475, which is why I came across it). So it's likely this issue no longer reproduces. If not, we can close it.

PR #476, opened 2024-01-04
content test: Aim taps on links more precisely at the text

gnprice:
By saying `tester.tapAt(find.text('hello'))`, we had been aiming at the center of the Text widgets, and expecting that to hit the recognizer we've put on the span, even though the widget is much wider than the span and the latter doesn't reach the former's center. Effectively we were relying on the presence of issue #214.

But with an upstream change yesterday:
* https://github.com/flutter/flutter/pull/140621
* https://github.com/flutter/flutter/issues/43400

such a tap no longer hits the span.  That broke these tests.

To fix them, aim the taps near the start of the widget instead.

---

Also bump the Flutter version to require that upstream change, fixing #214.

Fixes: #475
Fixes: #214

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -664,26 +664,26 @@ packages:
     dependency: transitive
     description:
       name: leak_tracker
-      sha256: e45c31f458d01fd9ef4a214feb2e153b72d5b1907435f4332b1637a2f348c021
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Great, thanks! Merged.

PR #478, opened 2024-01-05
ios: Take automatic updates to ios/macos files

chrisbobbe:
I'm seeing these changes applied when I build for iOS and macOS. Presumably we should have committed this in a recent Flutter upgrade, 8c4b1ed66. We didn't expect these kinds of changes, but Greg reports that he didn't build for iOS or macOS when writing that commit.


gnprice:
Cool, thanks.

Are you using the same version that that previous upgrade set as the new minimum? If not, then these changes could also just as easily be from more recent upstream changes. In that case the easiest thing will be to make this its own Flutter upgrade commit, and just note in the commit message that the ios/macos updates might have belonged to the previous upgrade.

chrisbobbe:
Thanks for the review! 

Ah, no I was using a later version.

> In that case the easiest thing will be to make this its own Flutter upgrade commit, and just note in the commit message that the ios/macos updates might have belonged to the previous upgrade.

Revision pushed, doing this.

ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -157,7 +157,7 @@
 		97C146E61CF9000F007C117D /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastUpgradeCheck = 1430;
+				LastUpgradeCheck = 1510;
```
gnprice, line 160: This change (and a similar one in other files) looks like it comes from upgrading your Xcode, rather than from any change in Flutter.

gnprice (commented):
Thanks! These changes all look good, but one group of them should probably be its own commit — see below.

chrisbobbe:
> These changes all look good, but one group of them should probably be its own commit — see below.

Thanks! Merged, with those separated out into a different commit.

Issue #479, opened 2024-01-08
Date separators push sticky header off screen


gnprice:
When a date separator is scrolled near the top of the message list's viewport, it pushes the sticky header up off the top of the screen:
![image](https://github.com/zulip/zulip-flutter/assets/28173/59470000-3d7b-49b9-8713-e7ec261d994b)

In other words it does the same thing that a new recipient header would do. (Once the date separator reaches the top of the viewport, though, the header reappears — this time the one provided by the date separator, so it looks the same as the old recipient header except for its date.)

But a date separator doesn't have all the information a recipient header does, so it doesn't replace the old recipient header and shouldn't push it off the screen. Instead, it should behave the same as a message, and let the preceding item's sticky header overlap into it.

This is a sort of followup to #173 / #469, where we introduced date separators in the first place.

PR #480, opened 2024-01-08
    msglist: Let recipient header overlap date, just like a message

gnprice:
Fixes: #479

Also update some tests that probably should have been updated in #469.

---
## Screenshots

Before:
image

After:
image

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -191,16 +191,19 @@ mixin _MessageSequence {
     if (index == 0 || !haveSameRecipient(messages[index - 1], message)) {
       items.add(MessageListRecipientHeaderItem(message));
       canShareSender = false;
-    } else if (!messagesSameDay(messages[index - 1], message)) {
-      items.add(MessageListDateSeparatorItem(message));
-      canShareSender = false;
     } else {
       assert(items.last is MessageListMessageItem);
       final prevMessageItem = items.last as MessageListMessageItem;
       assert(identical(prevMessageItem.message, messages[index - 1]));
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Thanks! I noticed the behavior of #479 too, when reviewing #469, but it didn't stand out to me as a problem. I see what you're saying about how the #479 behavior allows some information (like a message's topic) to get nudged off the screen.

This alternative approach seems mostly fine, except I think it now allows the appearance of incorrect information: if I'm looking at the first message on a new day, I can scroll up slowly until it looks like that same message is being labeled with a previous day (like "Yesterday"):

https://github.com/zulip/zulip-flutter/assets/22248748/701c6836-1f09-4d66-a216-0cbc94ddc85e

I don't yet see a nice way to fix that and #479 at the same time, so if you think this is OK, please merge at will.

gnprice:
Thanks for the review!

> This alternative approach seems mostly fine, except I think it now allows the appearance of incorrect information: if I'm looking at the first message on a new day, I can scroll up slowly until it looks like that same message is being labeled with a previous day (like "Yesterday"):

Yeah. I'm pretty sure Zulip web had an equivalent bug for years; it's not great, but I think it's a lot less conspicuous than #479. The latter not only pushes some information off the screen, but looks like a significant visual glitch because (a) if you're moving quickly, there's a big flash of white where the sticky header should be, and (b) when the sticky header finishes getting pushed off the screen, "it" (really the new one with the new date) suddenly reappears all at once.

I should file an issue for that new problem, though, so we can track it. I think fixing it will probably involve adding a further wrinkle to the `sticky_header` library.

Issue #342, opened 2023-10-26
Notifications delayed when app not running, on Android


gnprice:
In my draft implementation of #320, notifications appear just fine when the app is in the foreground, and also when it's running in the background. But when the app is not running at all — for example if you've swiped it away in the app switcher — and then get a message, it seems like the notification will usually not promptly appear.

On at least one occasion, the notification appeared several minutes later, after I'd gone and launched the app for the sake of other testing. I'm not sure how long it might be delayed if the user doesn't happen to independently launch the app; just now I tried it again, and it's been over 10 minutes and counting.

We don't have this problem in zulip-mobile, as far as I've ever noticed. So it should be entirely a client-side issue to solve.

The fix will probably involve taking over some of the work that the Java code in `package:firebase_messaging` does, for listening to the FCM service and acting on those messages. In zulip-mobile we do that work ourselves, in Kotlin, and then act on them within Kotlin as well. In a worst case we can always do the same here, but I'm hopeful that we can keep the application logic in Dart instead — we'd just need to be more eager in passing over to Dart than `package:firebase_messaging` is. But that will require investigation.

(Even in that worst case where we do end up keeping application logic in Kotlin for notifications, we'll be in a much better position to share its application *data* with our Dart code (cf https://github.com/zulip/zulip-mobile/issues/5117) than we were with React Native in zulip-mobile. Our accounts data is already in a nice clean SQLite table; and when we start storing from-the-server data, that will also be in reasonable tabular form from the beginning. And a critical aspect of any such data sharing would be integration tests spanning both sides (cf https://github.com/zulip/zulip-mobile/issues/5589), which Flutter supports well and RN just catastrophically didn't.)

gnprice:
It looks like the symptoms of this may vary between devices. My observations above were on my Pixel 5 running Android 13. @chrisbobbe reports at https://github.com/zulip/zulip-flutter/pull/352#pullrequestreview-1708313636 :
> FWIW, in some brief testing on the office Android device (Samsung Galaxy S9 running Android 9), I don't notice long delays in the case where the app isn't running at all. After I swipe it away in the app switcher, I can send myself DMs from a test account and still get a notification for each one within a few seconds.

So that's something to bear in mind when investigating.

One thing I wonder is if the difference is in the app switcher UI — if maybe the one in Samsung's Android 9 leaves the app still in the background, in a way that the one in Google/Pixel's Android 13 doesn't.

n-92:
Hello, 

[...]

PR #482, opened 2024-01-09
Fixes: Fix notification delay in the android when app is not running in background

Itsamanjain:
Fix: #342
fixes notification delay when app is not running in the background

### DIFF
--- a/lib/notifications.dart
+++ b/lib/notifications.dart
@@ -158,6 +158,7 @@ class NotificationService {
     _onRemoteMessage(message);
   }
 
+  @pragma('vm:entry-point')
   static Future<void> _onBackgroundMessage(FirebaseRemoteMessage message) async {
     // This callback will run in a separate isolate from the rest of the app.
     // See docs:


Itsamanjain:
Moving forward, need to focus on the notification sound. It currently works on Zulip web, but need to ensure it functions correctly in Zulip Flutter as well. The associated file, named 'zulip.mp3,' is already present in Zulip mobile.

gnprice:
Thanks.

What did you do to test that this change successfully fixes #342? That's the main work to be done here.

The notification sound seems unrelated to this PR's change, so let's keep the PR thread focused on the PR. There's an existing issue for playing Zulip's custom sound, but it should already be the case that we play the system default sound. If you're able to reproduce the issue reported at https://github.com/zulip/zulip-flutter/issues/342#issuecomment-1874716442 where that doesn't happen, please start a new thread and include details about how you reproduce it and what you observe.

Itsamanjain:
@gnprice to check these changes. I sent a direct message to myself from another ID. Also, try it in the test section of chat.zulip.org. I noticed notifications are faster there than on Zulip mobile 
IMG_20240110_102129

Itsamanjain:
Here is test result  in test section [notification ](https://chat.zulip.org/#narrow/stream/7-test-here/topic/notification.20delay)

Itsamanjain:
closed it by mistake. I will create a new one.

PR #484, opened 2024-01-10
fixes: Fix Notification delay when app is not running in background

Itsamanjain:
Fix: #342 
This PR is created after #482 . It was closed by mistake, so I created this. I tried DMing myself and also checked it in the "Test Here" section of chat.zulip.org. Now, I'm getting notifications on time even after the app is not running in the background. The test results are in the thread of #482 .

### DIFF
--- a/lib/notifications.dart
+++ b/lib/notifications.dart
@@ -158,6 +158,7 @@ class NotificationService {
     _onRemoteMessage(message);
   }
 
+  @pragma('vm:entry-point')
   static Future<void> _onBackgroundMessage(FirebaseRemoteMessage message) async {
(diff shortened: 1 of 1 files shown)


Itsamanjain:
@gnprice PR Is ready to merge.

gnprice:
The original issue #342 doesn't always reproduce, so it's already the case in main that you'll sometimes get notifications promptly after putting the app in the background.

To effectively test that something fixes that issue, what's needed is to find a pattern that reliably reproduces the issue, then try that after the fix and find that it no longer reproduces.

gnprice:
I tried this change myself, and it didn't fix the issue. I'll put details in the issue thread #342.

PR #485, opened 2024-01-17
changed the logo

chinmay7016:
this is my design
WhatsApp Image 2024-01-17 at 19 39 15


gnprice:
Thanks. I'm closing this because this design isn't one that I'll want to merge as is, and I think a PR thread won't be the right place to have a design discussion about it — better to use the issue thread, or a chat thread in #mobile.

It also looks like this PR doesn't actually supply a new icon anywhere. Rather it deletes all the files for the icon on iOS, and leaves the Android files untouched.

PR #489, opened 2024-01-19
deps: Upgrade packages (`flutter pub upgrade`)

chrisbobbe:
I ran `flutter pub upgrade`, then updated the CocoaPods lockfiles. This required some fiddling; ultimately the commands were:

    ( cd ios && pod update sqlite3 )
    ( cd macos && pod update sqlite3 )
    flutter build ios --config-only
    flutter build macos --config-only

Without the `pod update` commands, there was a misleading error message claiming the problem was that one pod required a higher minimum deployment target than another. Much like in 0dbc0c06d.

I also ran `flutter pub upgrade --major-versions`, but it didn't make any changes.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -45,7 +45,7 @@ PODS:
   - firebase_core (2.24.2):
     - Firebase/CoreOnly (= 10.18.0)
     - Flutter
-  - firebase_messaging (14.7.9):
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Tested on my iPhone and the office Android device, and things seemed fine.

gnprice:
Thanks for taking care of this! LGTM; merging.

Issue #487, opened 2024-01-19
compose: Message input (and not topic input) should auto-capitalize sentences


chrisbobbe:
See discussion: https://chat.zulip.org/#narrow/stream/48-mobile/topic/capitalization/near/1718941

@sirpengi found that we can accomplish this by passing [`textCapitalization: TextCapitalization.sentences`](https://api.flutter.dev/flutter/material/TextField/textCapitalization.html) to the message-input `TextField`.

(We shouldn't also pass that value to the topic input. It's common for topics to be lowercased, so autocapitalization would be inconvenient there.)

Greg adds:

> It'd be good to test this, because a major purpose of tests is to prevent accidentally reverting a change.
> 
> In particular in this case, the behavior is pretty subtle. For that reason I feel like if we did in some future PR (that was aimed at some other thing) accidentally break this behavior, it'd be pretty easy not to notice just from manual testing of the PR. That's reinforced by the fact that it took a few weeks after the first beta before we heard this report.

> I think a good test for this can be fairly simple:
> * make a compose box
> * check that the TextField for the content input has this property
> * check that the TextField for the topic input *doesn't* have this property

PR #490, opened 2024-01-24
Message compose box presents a sentence case keyboard

sirpengi:
Fixes: #487

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -302,6 +302,7 @@ class _ContentInput extends StatelessWidget {
               style: TextStyle(color: colorScheme.onSurface),
               decoration: InputDecoration.collapsed(hintText: hintText),
               maxLines: null,
+              textCapitalization: TextCapitalization.sentences,
             );
           }),
         ));
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -1,9 +1,22 @@
 import 'package:checks/checks.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:zulip/api/route/messages.dart';
+import 'package:zulip/model/narrow.dart';
 import 'package:zulip/widgets/compose_box.dart';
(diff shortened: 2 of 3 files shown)


lib/widgets/compose_box.dart
```diff
@@ -302,6 +302,7 @@ class _ContentInput extends StatelessWidget {
               style: TextStyle(color: colorScheme.onSurface),
               decoration: InputDecoration.collapsed(hintText: hintText),
               maxLines: null,
+              textCapitalization: TextCapitalization.sentences,
```
gnprice, line 305: nit in commit message:

    compose: Message compose box presents a sentence case keyboard
    
    Fixes: #487

The summary line should be in the imperative, not the indicative:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-summary-part-2

So a direct rewording would be "compose: Present a sentence case keyboard for message compose box".

With some further tweaks, a good version would be "compose: Use sentence-case keyboard for message input".

(In particular the "compose box" is what we call the whole widget at the bottom for composing and sending a message. It includes the message input, the topic input when present, the file/photo/etc buttons, and the send button.)

test/widgets/compose_box_test.dart
```diff
@@ -105,4 +118,70 @@ void main() {
...
+      ).toJson());
+
+      await tester.pumpWidget(
+        MaterialApp(
+          localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+          supportedLocales: ZulipLocalizations.supportedLocales,
+          home: GlobalStoreWidget(
+            child: PerAccountStoreWidget(
+              accountId: eg.selfAccount.id,
+              child: MessageListPage(narrow: narrow)))));
```
gnprice, line 150: Instead of `MessageListPage`, this can use just `ComposeBox` directly.

That also simplifies the setup — e.g. it means we no longer need to prepare a `GetMessagesResult`.)

gnprice (commented):
Thanks @sirpengi! Generally this looks good.

A few comments below. I think the most efficient thing here is if I make the changes myself in order to show, not tell, so I'll do that and go ahead and merge. (The one most nontrivial change, using ComposeBox in the test instead of MessageListPage, I had to try out anyway in order to be sure how well it would work.) Other than the commit message, I'll make the changes as additional commits on top.

PR #491, opened 2024-01-26
deps: Update Flutter to latest main, 3.19.0-9.0.pre.169

gnprice:

I also ran the usual two followup commands:
  $ flutter build ios --config-only
  $ flutter build macos --config-only

and nothing changed.


chrisbobbe:
Thanks! Merging.

Issue #192, opened 2023-06-15
Handle headings `<h1>`…`<h5>`, as well as `<h6>`


gnprice:
Currently the prototype handles `<h6>` elements, but not headings of higher levels `<h5>` up to `<h1>`. We should handle the higher-level headings too.

This means parsing them in `lib/model/content.dart`, and then styling them appropriately in `lib/widgets/content.dart`.

For appropriate styling, see the Zulip web app's `web/styles/rendered_markdown.css`.

Issue #162, opened 2023-06-08
Sibling layout interactions in content, like `p+ul`


gnprice:
Zulip web has a couple of CSS rules that adjust spacing between sibling block elements. From `web/styles/rendered_markdown.css`:
```css
    /* The spacing between two paragraphs is significantly larger.  We
       arrange things so that this spacing matches the spacing between
       paragraphs in two consecutive 1-line messages. */
    & p + p {
        margin-top: 10px;
    }
// …

    /* Reduce top-margin when a paragraph is followed by an ordered or bulleted list */
    & p + ul,
    p + ol {
        margin-top: 0;
    }
```

(The details have changed occasionally over the years, as in zulip/zulip@d7aa186dabb8587d48c1ce076c92b049e4fbba02 and
zulip/zulip@62f2396ee27f132a25f906107b60fa19b35c97ab.)

We should either match that behavior, or determine there's a good reason for mobile to differ from web here.

gnprice:
@karlstolley writes last week that he'll be working soon on cleaning up that CSS, simplifying things. (For example, that rule about `p + ul` has no effect, because the `p` has a bottom margin of 3px anyway and the top margin of a `ul` without that rule would be only 2px.)

So this task should become simpler after that work is complete.

gnprice:
Some details here:
* https://github.com/zulip/zulip/issues/28586

In particular one planned simplification is:
>  Shift inter-element Markdown element spacing to a single dimension (`margin-top`)

(whereas the status quo involves `margin-bottom` too — which in turn means that its behavior involves CSS margin-collapsing). So that will provide a significantly more straightforward basis for us to translate from.

gnprice:
Another rule with a similar character, about an absence-of-sibling rather than siblings, is the one with `:first-child` here:
```css
    /* Headings */
    & h1,
    h2,
    h3,
    h4,
    h5,
    h6 {
        // …
        margin-top: 15px;
        // …
    }

    /* Headings: Ensure that messages that start with a heading don't have
       a weirdly blank area at the very start of the message. */
    & h1:first-child,
    h2:first-child,
    h3:first-child,
    h4:first-child,
    h5:first-child,
    h6:first-child {
        margin-top: 0;
    }
```

As @sirpengi points out at #492, we don't yet implement that rule either. Probably we should consider it as part of this issue.

karlstolley:
[...]

PR #492, opened 2024-01-29
Implement h1-h5

sirpengi:
Fixes: #192

An unresolved issue in this PR is that headers at the beginning of a message should not have a top margin. See from
https://github.com/zulip/zulip/blob/main/web/styles/rendered_markdown.css :
```css
    /* Headings: Ensure that messages that start with a heading don't have
       a weirdly blank area at the very start of the message. */
    & h1:first-child,
    h2:first-child,
    h3:first-child,
    h4:first-child,
    h5:first-child,
    h6:first-child {
        margin-top: 0;
    }
```

We have an open issue #162 to handle content layout where this should be addressed in the future.

Screenshot from android:

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -927,11 +927,10 @@ class _ZulipContentParser {
       case 'h5': headingLevel = HeadingLevel.h5; break;
       case 'h6': headingLevel = HeadingLevel.h6; break;
     }
-    if (headingLevel == HeadingLevel.h6 && classes.isEmpty) {
-      // TODO(#192) handle h1, h2, h3, h4, h5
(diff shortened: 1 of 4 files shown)


gnprice (commented):
Thanks @sirpengi! Generally this looks good; one nit below, which I'll fix and then I'll merge.

> An unresolved issue in this PR is that headers at the beginning of a message should not have a top margin. […] We have an open issue #162 to handle content layout where this should be addressed in the future.

Yeah, thanks for flagging this. I've just added a note about it on that issue: https://github.com/zulip/zulip-flutter/issues/162#issuecomment-1918238045

Issue #36, opened 2023-03-30
login: Support web-based auth methods


gnprice:
I expect the first iteration of a login flow #11 will offer only email/password auth. We should add support for the rest.

In particular this means web-based auth, which covers Google auth and GitHub auth among others. ~~We can split out other methods as separate issues, if we don't find they're easy to just do at the same time.~~ Other forms of auth are tracked separately:
* #405 
* #462

chrisbobbe:
For Apple, there's a "Flutter Favorite" plugin [`sign_in_with_apple`](https://pub.dev/packages/sign_in_with_apple).

Probably it makes sense to start with Apple; as I recall, they'll reject apps in App Store review if they have third-party auth but no Apple auth.

chrisbobbe:
We should remember that although [`authentication_methods`](https://zulip.com/api/get-server-settings#response) is marked in the doc as deprecated in 2.1.0, it still seems necessary for writing a correct client. Discussion: https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/audit.20for.20change.20entries.20vs.2E.20central.20changelog/near/1404115

gnprice:
> For Apple, there's a "Flutter Favorite" plugin [`sign_in_with_apple`](https://pub.dev/packages/sign_in_with_apple).

Good to know, thanks.


> Probably it makes sense to start with Apple; as I recall, they'll reject apps in App Store review if they have third-party auth but no Apple auth.

That means we'll need Apple auth before a full launch, or even an open beta in the form of an alternate app in the App Store. But I think we'll need web-based auth before that milestone, too, because lots of people rely on that.

One thing that could cause us to prioritize web-based auth or any other auth method sooner than that milestone is if it's an obstacle for a significant number of users who might join an earlier stage of the beta. I think that's unlikely for Apple auth, because I suspect not many people are using it, but is more likely for web-based auth because of people using Google or GitHub auth.

gnprice:
I've split this issue up, as foreshadowed in the original description, and edited the description accordingly.

gnprice:
@fizy069 Hello, welcome!

This will be a complex issue partly because of the several different APIs the implementation needs to integrate with, and partly because it's security-sensitive and so requires an especially close review and a high level of making the code and commits clear so they're easy to review. So it won't be a good fit for one of the first few issues in Zulip to work on.

[...]

PR #495, opened 2024-01-29
login: Prep for web-auth

chrisbobbe:
Some prep commits toward #36, with:

- Making a transitive dependency into a direct dependency
- Improvements to our API-modeling code
- An NFC refactor of the username-password page


lib/widgets/login.dart
```diff
@@ -305,46 +363,31 @@ class _PasswordLoginPageState extends State<PasswordLoginPage> {
       }
 
       // TODO(server-7): Rely on user_id from fetchApiKey.
-      final int userId = result.userId ?? await _getUserId(result);
+      final int userId = result.userId ?? await widget.loginPageState._getUserId(result.email, result.apiKey);
```
gnprice, line 366: nit: overlong line
```suggestion
      final int userId = result.userId
        ?? await widget.loginPageState._getUserId(result.email, result.apiKey);
```

lib/widgets/login.dart
```diff
@@ -305,46 +363,31 @@ class _PasswordLoginPageState extends State<PasswordLoginPage> {
...
       // ignore: use_build_context_synchronously
       if (!context.mounted) {
         return;
       }
 
-      final globalStore = GlobalStoreWidget.of(context);
-      // TODO(#108): give feedback to user on SQL exception, like dupe realm+user
-      final accountId = await globalStore.insertAccount(AccountsCompanion.insert(
-        realmUrl: realmUrl,
+      widget.loginPageState._tryInsertAccountAndNavigate(
```
gnprice, line 373:
```suggestion
      await widget.loginPageState._tryInsertAccountAndNavigate(
```

Otherwise the `finally` block happens sooner than it had been.

lib/widgets/login.dart
```diff
@@ -400,28 +443,19 @@ class _PasswordLoginPageState extends State<PasswordLoginPage> {
...
+      child: AutofillGroup(
+        child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
+          usernameField,
+          const SizedBox(height: 8),
+          passwordField,
+          const SizedBox(height: 8),
+          ElevatedButton(
+            onPressed: widget.loginPageState._inProgress
+              ? null
+              : _submit,
```
gnprice, line 457: nit: this can still fit on one line:
```suggestion
            onPressed: widget.loginPageState._inProgress ? null : _submit,
```

and I think it's slightly easier to read that way, since the `null` is trivial and this brings the `_submit` (which is the main answer) closer to the `onPressed:` (which is the question).

lib/widgets/login.dart
```diff
@@ -245,6 +245,73 @@ class PasswordLoginPage extends StatefulWidget {
...
+        bottom: _inProgress
+          ? const PreferredSize(preferredSize: Size.fromHeight(4),
+              child: LinearProgressIndicator(minHeight: 4)) // 4 restates default
+          : null),
+      body: SafeArea(
+        minimum: const EdgeInsets.all(8),
+        child: Center(
+          child: ConstrainedBox(
+            constraints: const BoxConstraints(maxWidth: 400),
+            child: _UsernamePasswordForm(loginPageState: this)))));
```
gnprice, line 301: For now, we'll plan to put the username/password form and all the
    web-auth buttons together on the same page. We don't have to keep it
    this way forever, but it has some advantages:
    - My feeling is that most web apps do this, and I don't see an
      obvious reason to deviate from that pattern here, probably
      because:
    - In the common case of username/password authentication, users
      won't need to tap a "Sign in with password" button and then
      context-switch to a new page. The form is just there, right after
      your realm URL has been accepted.
    - Also, it'll be more convenient to share code among the various
      ways of authenticating. Code that manages the server-settings
      data, and code that inserts an account in the database and
      navigates to the authenticated part of the app.

Sure, happy to try this.

On the third point, I think we can get that pretty much just as well if they're on separate pages too — we can have the code in one place but it's a mixin, or a base class. But I think the other two reasons are good ones.

gnprice (commented):
Thanks! Generally looks good; small comments, with just one affecting behavior.

chrisbobbe:
Thanks for the review! Revision pushed (and it looks like I forgot to comment about that last week).

gnprice:
Thanks! Looks good; merging.

Issue #82, opened 2023-04-21
Open a message list in the middle of history, when requested


gnprice:
When the user takes an action that should navigate to a particular "anchor" message (like through #73 internal links, #76 deep links, or a notification), we should take them straight to that message, rather than to the latest message or first unread message in the narrow.

This is a feature the zulip-mobile RN app lacks: https://github.com/zulip/zulip-mobile/issues/3604 . (That's for internal links, but the app doesn't have it for notifications either — see `narrowToNotification` which passes no anchor to `doNarrow` — and doesn't have deep links.)

There are two parts to this:
* Fetch the messages around the desired anchor, rather than around the latest message or the first unread.
* Control the scroll position so that we show the desired anchor message.

This has a lot in common with:
* #80

This issue has fewer prerequisites technically than #80 — no need for #79 first. On the other hand it has more prerequisites before there's a UX context where it makes sense: it'd need either #73, #76, or notifications, all of which are more complex than #79.

E-m-i-n-e-n-c-e:
Started working on this.

https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Control.20scroll.20position.20on.20new.20and.20newly-fetched.20messages

https://github.com/user-attachments/assets/a379256e-3208-48b5-960c-7740a06fe497

https://github.com/user-attachments/assets/9feec67c-380e-4e92-9631-37a58ec8d8ab

chrisbobbe:
This came up in beta feedback: [#mobile > Flutter Beta: Some Feedback @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20Beta.3A.20Some.20Feedback/near/2153404)

gnprice:
> When the user takes an action that should navigate to a particular "anchor" message (like through #73 internal links, #76 deep links, or a notification), we should take them straight to that message,

I've filed #1565 for the notifications case of this. For deep links #76, we'll handle it as part of implementing deep links in the first place (whenever we eventually do so — it's not a near-term feature).

This issue is then reduced to the case of following an internal link with /near/. (We implemented #73 long ago.) The bulk of the work for implementing it is independent anyway of how the user does the navigation — it's about making the message list support being in the middle of history, so scrolling and fetching in both directions. That work is now nearly complete.

Issue #80, opened 2023-04-21
Open message list at first unread, rather than latest message


gnprice:
When navigating to a message list, our default behavior (if there isn't a specific message we're asked to go to) should be to show the first unread message, if any, and the messages around it.

This means
* fetching around the `first_unread` anchor (https://zulip.com/api/get-messages), rather than `latest`;
* controlling the scroll position so that we show the first unread, if any, and the latest messages otherwise.

This should probably come after showing unread markers, #79.

PR #496, opened 2024-01-30
anchors 1/n: Handle GrowthDirection.reverse in sticky_header

gnprice:
This is the first of several PRs leading us to #80 and #82, opening the message list at an anchor other than the latest message in history.

For this PR, the main work is to lift one of the simplifying assumptions I originally made for an MVP of the `sticky_header` library, allowing the SliverStickyHeaderList to be a sliver that counts backward in the enclosing ScrollView rather than forward.

We also rearrange some things in how MessageList builds the list, in order to exercise that new flexibility and to get closer to how we'll want things when implementing #80 and #82.

## Main commit messages

sticky_header: Handle GrowthDirection.reverse

This resolves the last of the "TODO dir" comments in this library.

We had originally written this library to rely on the assumption
that `constraints.growthDirection` was `GrowthDirection.forward`,
i.e. that the direction in which `constraints.scrollOffset` increases
was the same as `constraints.axisDirection`, in order to keep down
the number of distinct directions one needs to keep track of in
writing and reading (and debugging) the implementation.

But now that this logic is relatively solid within that case --
and now that we have a solid test suite for this library -- we can
go back and identify the handful of places that need to be flipped
around for `GrowthDirection.reverse`, and do so.

This support isn't yet especially *useful*: there are a couple of
unrelated bugs which are triggered when the area allotted to this
sliver isn't the entire viewport, and there isn't a lot of reason
to have a sliver with GrowthDirection.reverse when no other sliver
is potentially occupying part of the viewport.  We'll fix those
separately.

---

msglist [nfc]: Let ScrollView count forward, and sliver handle reverse

This means that the overall scroll view has AxisDirection.down
as the scroll direction, rather than AxisDirection.up, and the
SliverStickyHeader is passed GrowthDirection.reverse instead
of GrowthDirection.forward in order to get the same effect.

This makes the scrolling a little easier to think about when focused
on MessageList's code, because for example (as seen in this diff)
`scrollMetrics.extentBefore` now refers to the messages that are
older than the ones on screen, rather than those that are newer.

This also brings us closer to the setup we'll want for #80 and #82,
opening the message list at an anchor other than the end: the empty
placeholder sliver at the bottom will become the list of messages
after the starting anchor, while the sliver at the top will remain
the list of messages above the anchor.

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -218,8 +218,7 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
       _scrollToBottomVisibleValue.value = true;
     }
 
-    final extentRemainingAboveViewport = scrollMetrics.maxScrollExtent - scrollMetrics.pixels;
-    if (extentRemainingAboveViewport < kFetchMessagesBufferPixels) {
+    if (scrollMetrics.extentAfter < kFetchMessagesBufferPixels) {
       // TODO: This ends up firing a second time shortly after we fetch a batch.
       //   The result is that each time we decide to fetch a batch, we end up
       //   fetching two batches in quick succession.  This is basically harmless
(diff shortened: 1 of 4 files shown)


chrisbobbe:
Thanks! Merged.

Issue #499, opened 2024-02-01
content: **bold text** should be formatted with `weightVariableTextStyle`


chrisbobbe:
Like other text that we want to present in Source Sans 3 (a variable-weight font), we should style `**bold text**` in content with `weightVariableTextStyle`. Currently, it is styled with just `fontWeight`:

```dart
  InlineSpan _buildStrong(StrongNode node) => _buildNodes(node.nodes,
    style: const TextStyle(fontWeight: FontWeight.w600));
```

gnprice:
Writing down a discussion we had in the office: it looks like the symptoms of this bug are currently subtle, but only because of the upstream Flutter issue discussed at #500.

The TextStyle that ultimately applies to the text inside one of these spans inherits a `fontVariations` from an ancestor, which in a paragraph is going to be the one `weightVariableTextStyle` gives for normal weight. In the absence of that upstream issue, the `wght` variation from there would control (probably, depending on how the issue gets resolved), and the text would have normal weight (on default system settings) despite being from a `StrongNode`.

But with that issue, the `fontWeight` combines with the `wght` variation and we get a bold effective weight. Effectively this bug roughly cancels out with #500 (… at least on Android, where #500 applies). I say "roughly" because even in the scenario where they both trigger, the effect of #500 seems to make the glyphs heavier but leave their spacing unchanged, so it's not quite the same effect as getting the actual bolder version of the font.

Issue #500, opened 2024-02-01
text: Some bold text is way bolder than intended


chrisbobbe:
See how bold the sender names and topic headers are in this screenshot on the office Android device, compared to a better alternative:

| Current | Better alternative |
| --- | --- |
| ![image](https://github.com/zulip/zulip-flutter/assets/22248748/5a4e6dae-3d7e-40b2-9c77-ff61af8eb88c) | ![image](https://github.com/zulip/zulip-flutter/assets/22248748/db74ef97-567c-40a2-bfa5-c67ecca49e98) |

The "better alternative" can be reached in this case by commenting out a line in `weightVariableTextStyle`:

```diff
diff --git lib/widgets/text.dart lib/widgets/text.dart
index 1a80cee4b..04d001bbd 100644
--- lib/widgets/text.dart
+++ lib/widgets/text.dart
@@ -69,7 +69,7 @@ TextStyle weightVariableTextStyle(BuildContext? context, {
     // This use of `fontWeight` shouldn't affect glyphs in the preferred,
     // "wght"-axis font. If it does, see for debugging:
     //   https://github.com/zulip/zulip-flutter/issues/65#issuecomment-1550666764
-    fontWeight: clampVariableFontWeight(value),
+    // fontWeight: clampVariableFontWeight(value),
 
     inherit: true);
 }
```

In other words, the comment about "This use of `fontWeight` shouldn't [etc.]" has proven helpful: the use of `fontWeight` is indeed affecting glyphs in Source Sans 3. The "see for debugging" link is relevant, but I plan to replace it with a link to an upstream Flutter issue with the `triaged-engine` label: that's https://github.com/flutter/flutter/issues/136779 .

Removing that `fontWeight` line would be sad; it exists for the sake of glyphs that can't be rendered in the primary font family (like Source Sans 3) and might need to be rendered in a fallback font that doesn't have a "wght" axis.

chrisbobbe:
We've seen this on Greg's Android device, and (as screenshotted above) on the office Android device. I don't think we've yet had reports from iOS.

gnprice:
Thanks for spotting this. I'd actually noticed that the recipient headers and sender names seemed bolder than I'd expect; but our source code seemed to clearly call for the same weights that Vlad specified in the design, so I figured that was just his design choice. Apparently not!

> I don't think we've yet had reports from iOS.

(And in particular it doesn't seem to reproduce on your iPhone.)

[...]

PR #501, opened 2024-02-01
content: Fix **`bold code`** rendering with regular weight

chrisbobbe:
Also format `**bold text**` using weightVariableTextStyle, fixing #499, and update a comment to point to #500.

| Before | After |
| --- | --- |
| image | image |

Fixes: #498
Fixes: #499

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -454,7 +454,7 @@ class InlineContent extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Text.rich(_builder.build());
+    return Text.rich(_builder.build(context));
   }
 }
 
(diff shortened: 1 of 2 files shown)


chrisbobbe:
If #439 is merged first, I should update the commit message for

content: Fix **`bold code`** rendering with regular weight

because at that point the needed `weightVariableTextStyle` won't come from paragraph-content styles, but rather from a custom `Typography` value.

chrisbobbe:
I expect the CI failure to be resolved with #502.

chrisbobbe:
(Rebased atop the merged #502; PTAL.)

gnprice:
(Plus this, I guess:)

> If #439 is merged first, I should update the commit message […]

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

Issue #497, opened 2024-01-30
content model: Avoid `classes.contains`, which does extra work like creating a Set


chrisbobbe:
Greg and I discovered on our call today that `classes.contains` does some extra work:

```dart
  /// Determine if this element contains the class [value].
  ///
  /// This is the Dart equivalent of jQuery's
  /// [hasClass](http://api.jquery.com/hasClass/).
  @override
  bool contains(Object? value) => readClasses().contains(value);
```

```dart
  @override
  Set<String> readClasses() {
    final s = LinkedHashSet<String>();
    final classname = _element.className;

    for (var name in classname.split(' ')) {
      final trimmed = name.trim();
      if (trimmed.isNotEmpty) {
        s.add(trimmed);
      }
    }
    return s;
  }
```

When we're checking that there are no classes, we should just check if `className` is empty. When we're checking that there is one class, we should just do string equality on `className`. When we're checking that there is more than one class, we might want to do a regular expression or something.

Issue #488, opened 2024-01-19
content: Support Zulip content outside messages (even outside per-account contexts)


chrisbobbe:
The web app supports Zulip Markdown rendering in contexts other than just message content:

See for example `realm_description` from the server settings on CZO, shown on the login page:

<img width=200 src="https://github.com/zulip/zulip-flutter/assets/22248748/95f057fe-961b-4200-b25a-72f67524b62b">

And see `rendered_description` on stream objects in the `/register` response, shown in stream settings. This screenshot shows links (like `[text](link)`) being rendered in some stream descriptions:

<img width="400" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/380ed378-1c7f-419f-ac76-6a13b4ed5d39">

This kind of thing would be easy to support in our app, except that some of the widgets in `lib/widgets/content.dart` have grown dependencies on data that won't be available in these other contexts. For example, `MessageImage` uses `RealmContentNetworkImage`, which requires a `PerAccountStoreWidget` ancestor. It also links to the lightbox, which is also per-account and additionally pulls data about the message's sender.

Currently, the following single line of code (atop #281) actually succeeds in visually reproducing CZO's realm description:

```dart
BlockContentList(nodes: parseContent(widget.serverSettings.realmDescription).nodes)
```

<img width=400 src="https://github.com/zulip/zulip-flutter/assets/22248748/c111a198-12af-4dc2-b566-5bd6d014c526">

But the links don't work; an error prints to the console when I tap one of them.

### References
* Issues that require this as a prerequisite:
  * #1896 
  * #1945 
  * #178

chrisbobbe:
So, we need the content widgets to work even when their message- and account-data dependencies are absent. Ideally, we would have a way to enforce that such strict dependencies aren't reintroduced.

chrisbobbe:
I noticed that this repo's `a-content` label has this description: "Parsing and rendering message contents"

Should we generalize that to something like "Parsing and rendering Zulip content, like in messages"?

gnprice:
> Should we generalize that

Sure, done: "Parsing and rendering Zulip HTML content, notably message contents".

alya:
However we end up presenting stream description, we'll probably want to show descriptions for views the same way as well. Web app issue: https://github.com/zulip/zulip/issues/29769.

PR #503, opened 2024-02-02
content: Support spoilers!

chrisbobbe:
TODO tests, but here's my code so far.

Fixes: #358

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -258,6 +258,21 @@ class QuotationNode extends BlockContentNode {
   }
 }
 
+class SpoilerNode extends BlockContentNode {
+  const SpoilerNode({super.debugHtmlNode, required this.header, required this.content});
+
+  final List<BlockContentNode> header;
+  final List<BlockContentNode> content;
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return [
+      _BlockContentListNode(header).toDiagnosticsNode(name: 'header'),
+      _BlockContentListNode(content).toDiagnosticsNode(name: 'content'),
+    ];
+  }
+}
+
 class CodeBlockNode extends BlockContentNode {
   const CodeBlockNode(this.spans, {super.debugHtmlNode});
 
@@ -809,6 +824,26 @@ class _ZulipContentParser {
     return ListNode(listStyle!, items, debugHtmlNode: debugHtmlNode);
   }
 
+  BlockContentNode parseSpoilerNode(dom.Element divElement) {
+    assert(_debugParserContext == _ParserContext.block);
+    assert(divElement.localName == 'div'
+        && divElement.className == 'spoiler-block');
+
+    if (divElement.nodes case [
+      dom.Element(
+        localName: 'div', className: 'spoiler-header', nodes: var headerNodes),
+      dom.Element(
+        localName: 'div', className: 'spoiler-content', nodes: var contentNodes),
+    ]) {
+      return SpoilerNode(
+        header: parseBlockContentList(headerNodes),
+        content: parseBlockContentList(contentNodes),
+      );
+    } else {
+      return UnimplementedBlockContentNode(htmlNode: divElement);
+    }
+  }
+
   BlockContentNode parseCodeBlock(dom.Element divElement) {
     assert(_debugParserContext == _ParserContext.block);
     final mainElement = () {
@@ -975,6 +1010,10 @@ class _ZulipContentParser {
     if (localName == 'blockquote' && className.isEmpty) {
       return QuotationNode(debugHtmlNode: debugHtmlNode,
         parseBlockContentList(element.nodes));
+    }
+
+    if (localName == 'div' && className == 'spoiler-block') {
+      return parseSpoilerNode(element);
     }
 
     if (localName == 'div' && className == 'codehilite') {
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:html/dom.dart' as dom;
 import 'package:html/parser.dart';
@@ -113,6 +114,20 @@ class UnimplementedBlockContentNode extends BlockContentNode
   // No ==/hashCode, because htmlNode is a whole subtree.
 }
 
+class _BlockContentListNode extends DiagnosticableTree {
+  const _BlockContentListNode(this.nodes);
+
+  final List<BlockContentNode> nodes;
+
+  @override
+  String toStringShort() => 'BlockContentNode list';
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return nodes.map((node) => node.toDiagnosticsNode()).toList();
+  }
+}
+
 /// A block content node whose children are inline content nodes.
 ///
 /// A node of this type expects a block layout context from its parent,
@@ -226,22 +241,9 @@ class ListNode extends BlockContentNode {
   @override
   List<DiagnosticsNode> debugDescribeChildren() {
     return items
-      .map((nodes) => _ListItemDiagnosticableNode(nodes).toDiagnosticsNode())
+      .mapIndexed((i, nodes) =>
+        _BlockContentListNode(nodes).toDiagnosticsNode(name: 'item $i'))
       .toList();
-  }
-}
-
-class _ListItemDiagnosticableNode extends DiagnosticableTree {
-  _ListItemDiagnosticableNode(this.nodes);
-
-  final List<BlockContentNode> nodes;
-
-  @override
-  String toStringShort() => 'list item';
-
-  @override
-  List<DiagnosticsNode> debugDescribeChildren() {
-    return nodes.map((node) => node.toDiagnosticsNode()).toList();
   }
 }
 
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -125,6 +125,79 @@ class ContentExample {
     const ImageEmojiNode(
       src: '/static/generated/emoji/images/emoji/unicode/zulip.png', alt: ':zulip:'));
 
+  static const spoilerDefaultHeader = ContentExample(
+    'spoiler with default header',
+    '```spoiler\nhello world\n```',
+    expectedText: 'Spoiler', // or a translation
+    '<div class="spoiler-block"><div class="spoiler-header">\n'
+        '</div><div class="spoiler-content" aria-hidden="true">\n'
+        '<p>hello world</p>\n'
+        '</div></div>',
+    [SpoilerNode(
+      header: [],
+      content: [ParagraphNode(links: null, nodes: [TextNode('hello world')])],
+    )]);
+
+  static const spoilerPlainCustomHeader = ContentExample(
+    'spoiler with plain custom header',
+    '```spoiler hello\nworld\n```',
+    expectedText: 'hello',
+    '<div class="spoiler-block"><div class="spoiler-header">\n'
+        '<p>hello</p>\n'
+        '</div><div class="spoiler-content" aria-hidden="true">\n'
+        '<p>world</p>\n'
+        '</div></div>',
+    [SpoilerNode(
+      header: [ParagraphNode(links: null, nodes: [TextNode('hello')])],
+      content: [ParagraphNode(links: null, nodes: [TextNode('world')])],
+    )]);
+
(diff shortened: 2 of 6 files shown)


gnprice:
Thanks! This direction looks good. I like how the open/close animations are very smooth.

The two prep commits toward #497 look good. I've just sent #505 incorporating those and finishing up the rest of #497.

chrisbobbe:
Thanks for the review! Revision pushed, this time with content-parsing tests; PTAL. Widget tests still TODO.

test/model/content_test.dart
```diff
@@ -324,6 +324,58 @@ void main() {
...
+      // ```spoiler **bold** [czo](https://chat.zulip.org/)\n*italic* [zulip](https://zulip.com/)\n```
+      '<div class="spoiler-block"><div class="spoiler-header">\n'
+          '<p><strong>bold</strong> <a href="https://chat.zulip.org/">czo</a></p>\n'
+          '</div><div class="spoiler-content" aria-hidden="true">\n'
+          '<p><em>italic</em> <a href="https://zulip.com/">zulip</a></p>\n'
+          '</div></div>',
+      [const SpoilerNode(
+        header: [ParagraphNode(
+          links: [LinkNode(url: 'https://chat.zulip.org/', nodes: [TextNode('czo')])],
+          nodes: [
```
gnprice, line 363:
```suggestion
        header: [
          ParagraphNode(links: null, nodes: [
```

This convenience is in a comment on `equalsNode`, which is the engine of these parsing tests. It should probably be in `model/content_test.dart` itself so as to be more visible:
```dart
  // In [expected], for the `links` field of [ParagraphNode] or
  // any other [BlockInlineContainerNode] subclass, use `null`.
  // This field will be ignored in [expected], and instead the
  // field's value in [actual] will be checked for accuracy against
  // the [BlockInlineContainerNode.nodes] field on the same node.
  void equalsNode(ContentNode expected) {
```

test/model/content_test.dart
```diff
@@ -324,6 +324,58 @@ void main() {
...
+          '<p>world</p>\n'
+          '</div></div>',
+      [const SpoilerNode(
+        header: [ParagraphNode(links: null, nodes: [TextNode('hello')])],
+        content: [ParagraphNode(links: null, nodes: [TextNode('world')])],
+      )],
+    );
+
+    testParse('with rich header and content',
+      // ```spoiler **bold** [czo](https://chat.zulip.org/)\n*italic* [zulip](https://zulip.com/)\n```
```
gnprice, line 354: Did you find a way to provoke the header into being a list of more than one BlockContentNode? That'd be fun to also test, if so.

If not, then a test case where the header is like `1. * ## details`, showing that it's not necessarily a paragraph, would be good. (I guess more generally: it'd be good to demonstrate not-a-paragraph, either as the only BlockContentNode or not.)

gnprice (commented):
Thanks! The tests generally look good too; a few comments below.

lib/model/content.dart
```diff
@@ -256,6 +256,49 @@ class QuotationNode extends BlockContentNode {
...
+  const SpoilerNode({super.debugHtmlNode, required this.header, required this.content});
+
+  final List<BlockContentNode> header;
+  final List<BlockContentNode> content;
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return [
+      _SpoilerHeaderDiagnosticableNode(header).toDiagnosticsNode(),
+      _SpoilerContentDiagnosticableNode(content).toDiagnosticsNode(),
```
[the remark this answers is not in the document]
gnprice: That commit is:
cd0ed9c3e content [nfc]: Use toDiagnosticsNode(name: ...)

gnprice:
I recommend rebasing atop #511, which I just sent, and converting the content-parsing tests into the new `ContentExample` form from that PR — that should help simplify writing the widget tests. In particular it means you can get widget smoke tests very compactly, by reusing the examples from content parsing.

You'll also want some interaction tests, though (for showing and hiding on tap), and those might not benefit from `ContentExample`; see the comment in `widgets/content_test.dart` added at the tip of #511. But when it comes to those, the examples can be kept very simple, with the header and content made of just plain text. I think all the wild and wooly other stuff that can go in the headers doesn't affect the show/hide behavior.

(Hmm I guess if there's a link in the header, it'd be nice to check that tapping the link works and isn't pre-empted by the header's gesture detector. But that can still be pretty concise, along the lines of the HTML examples in the "LinkNode interactions" tests.)

lib/model/content.dart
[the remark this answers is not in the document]
chrisbobbe: Awesome; glad to have this more satisfying way. Thanks!

test/model/content_test.dart
chrisbobbe: > Did you find a way to provoke the header into being a list of more than one BlockContentNode? That'd be fun to also test, if so.

No, try as I might, sadly I did not manage to do this. 😛

chrisbobbe: No wait! I did! Here: https://chat.zulip.org/#narrow/stream/7-test-here/topic/content/near/1735802

chrisbobbe: But that includes an image, and I don't think we can build our content widget for image previews, `MessageImage`, in a `testContentSmoke` call. That's because `MessageImage` needs `PerAccountStoreWidget.of` and `InheritedMessage.of`, and `testContentSmoke` doesn't give support for those. When we think about those dependencies, we should have this issue in mind:

- #488

gnprice: Cool.

Yeah, it's fine that `testContentSmoke` won't handle that; it's just a shortcut to enable a lot of those tests to be one line each. Instead you can write a test that's similar to what you'd get by inlining `testContentSmoke` and `prepareContentBare`, but that provides a few more enclosing widgets in the `pumpWidget` call.

(And if we find ourselves with a lot of tests like that doing similar things, we can abstract those into further helpers similar to `prepareContentBare` and/or `testContentSmoke`, or perhaps have the latter take an optional flag.)

chrisbobbe:
OK, revision pushed!

The feature is still imperfect:

- I haven't tested the screen-reader interface and I suspect it will need special care with `Semantics`
- We need to think more about the direction in which the spoiler opens, and how that interacts with the message-list sliver's `GrowthDirection`.

test/widgets/content_test.dart
```diff
@@ -89,6 +94,97 @@ void main() {
...
+          ..value.equals(expected ? 1 : 0)
+          ..status.equals(expected ? AnimationStatus.completed : AnimationStatus.dismissed);
+      }
+
+      const example = ContentExample.spoilerHeaderHasImage;
+
+      testWidgets('tap image', (tester) async {
+        final pushedRoutes = await prepareContent(tester, example.html);
+
+        await tester.tapAt(tester.getCenter(find.byType(RealmContentNetworkImage)));
```
gnprice, line 149: nit: can simplify to:
```suggestion
        await tester.tap(find.byType(RealmContentNetworkImage));
```

test/widgets/page_checks.dart
```diff
@@ -9,3 +9,7 @@ extension WidgetRouteChecks on Subject<WidgetRoute> {
 extension AccountPageRouteMixinChecks on Subject<AccountPageRouteMixin> {
   Subject<int> get accountId => has((x) => x.accountId, 'accountId');
 }
+
+extension PageRouteChecks on Subject<PageRoute> {
```
gnprice, line 13: nit: let's keep the checks extensions for Flutter upstream classes gathered in `flutter_checks.dart`

test/model/content_test.dart
```diff
@@ -125,6 +125,78 @@ class ContentExample {
...
+        '<p><em>italic</em> <a href="https://zulip.com/">zulip</a></p>\n'
+        '</div></div>',
+    [SpoilerNode(
+      header: [ListNode(ListStyle.ordered, [
+          [ListNode(ListStyle.unordered, [
+            [HeadingNode(level: HeadingLevel.h2, links: null, nodes: [
+              TextNode('hello'),
+            ])]
+          ])],
+        ])],
```
gnprice, line 171: nit: two-space indents, here and below
```suggestion
      header: [ListNode(ListStyle.ordered, [
        [ListNode(ListStyle.unordered, [
          [HeadingNode(level: HeadingLevel.h2, links: null, nodes: [
            TextNode('hello'),
          ])]
        ])],
      ])],
```

test/model/content_test.dart
```diff
@@ -125,6 +125,78 @@ class ContentExample {
...
+          LinkNode(url: 'https://zulip.com/', nodes: [TextNode('zulip')])
+        ])],
+    )]);
+
+  static const spoilerHeaderHasImage = ContentExample(
+    'spoiler a header that has an image in it',
+    '```spoiler [image](https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3)\nhello world\n```',
+    '<div class="spoiler-block"><div class="spoiler-header">\n'
+      '<p><a href="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3">image</a></p>\n'
+      '<div class="message_inline_image"><a href="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3" title="image"><img src="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3"></a></div></div><div class="spoiler-content" aria-hidden="true">\n'
```
gnprice, line 184: nit: this line finishes an element it didn't begin (the `div.spoiler-header`), so should have a newline before starting the next:
```suggestion
      '<div class="message_inline_image"><a href="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3" title="image"><img src="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3"></a></div></div>'
      '<div class="spoiler-content" aria-hidden="true">\n'
```

(Ideally the indentation would help show the nesting of elements, but this part is easy and an improvement regardless)

gnprice (commented):
Thanks! All looks good, with just small comments below.

lib/model/content.dart
```diff
@@ -226,33 +241,35 @@ class ListNode extends BlockContentNode {
   @override
   List<DiagnosticsNode> debugDescribeChildren() {
     return items
-      .map((nodes) => _ListItemDiagnosticableNode(nodes).toDiagnosticsNode())
+      .mapIndexed((i, nodes) =>
+        _BlockContentListNode(nodes).toDiagnosticsNode(name: 'item $i'))
```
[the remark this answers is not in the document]
chrisbobbe: Ah indeed; thanks for catching this!

gnprice:
Oh, and forgot that I'd meant to reply to these:

> The feature is still imperfect:
> 
> * I haven't tested the screen-reader interface and I suspect it will need special care with `Semantics`
> * We need to think more about the direction in which the spoiler opens, and how that interacts with the message-list sliver's `GrowthDirection`.
> 
> but we can leave that for later work. I can file issues, unless one or both of these falls under an umbrella issue that you have a vision for.

Yeah, please go ahead and file an issue for the second one. The first one is probably best handled as part of a future sweep through our semantics and screen-reader behavior; I'll go file an issue for that and include this as an item to remember to check when doing that.

test/model/content_test.dart
chrisbobbe: Yeah, I wasn't sure how to compromise between being helpful about element nesting and not stretching out the code too much over a lot of lines. Like toward the latter extreme, this could instead be:

```dart
    '<div class="spoiler-block">'
      '<div class="spoiler-header">\n'
        '<p>'
          '<a href="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3">image</a>'
        '</p>\n'
        '<div class="message_inline_image">'
          '<a href="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3" title="image">'
            '<img src="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=3">'
          '</a>'
        '</div>'
      '</div>'
      '<div class="spoiler-content" aria-hidden="true">\n'
        '<p>hello world</p>\n'
      '</div>'
    '</div>\n',
```

or something like that, right?

chrisbobbe:
Sure: #536

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

test/model/content_test.dart
gnprice: Yeah. And like you suggested, going all the way there would be a lot more lines and probably not worth the trade-off in making it hard to see several of these on the screen at once.

Issue #354, opened 2023-11-01
Handle `<time>` elements ("global times")


gnprice:
We should handle `<time>` elements. These are the output of the Zulip "global times" feature: https://zulip.com/help/global-times

Handling these means parsing them in `lib/model/content.dart`, and then styling them appropriately in `lib/widgets/content.dart`. For appropriate styling, see the Zulip web app's `web/styles/rendered_markdown.css`.

PR #504, opened 2024-02-05
Global Times

sirpengi:
Fixes: #354 

Screenshot_20240205_162052


lib/widgets/content.dart
```diff
@@ -687,6 +692,41 @@ class UserMention extends StatelessWidget {
 //     borderRadius: BorderRadius.all(Radius.circular(3))));
 }
 
+class GlobalTime extends StatelessWidget {
```
gnprice, line 695: nit: put this after MessageImageEmoji, to match the if/else-if chain above (which matches the node-class definitions)

test/widgets/content_test.dart
```diff
@@ -93,6 +93,18 @@ void main() {
     tester.widget(find.text(r'\lambda'));
   });
 
+  testWidgets('GlobalTime smoke', (tester) async {
```
gnprice, line 96: nit: put this after the MathInlineNode test below, to match the code under test in `lib/widgets/content.dart`

gnprice (commented):
Thanks @sirpengi! Generally this looks good; just small comments below.

sirpengi:

PR #505, opened 2024-02-05
Cut inefficient CssClassSet out of content parsing

gnprice:
Fixes: #497

The first two commits here are by @chrisbobbe, taken from #503. The later commits go in the same direction for checking a class list is empty, and then handle the remaining couple of places that are a bit trickier, applying regexps to `className` to cover the different possibilities.


chrisbobbe:
Thanks, merging!

PR #506, opened 2024-02-06
msglist test [nfc]: Cut some redundant per-case store and connection locals

gnprice:
These already exist as late variables at the top of `main`, and at each of these lines they've already been set by the setupMessageListPage helper to exactly the same values found here.


chrisbobbe:
Thanks! Merged.

Issue #384, opened 2023-11-16
Inbox: Show @-mention indicator beside stream unread count badge


chrisbobbe:
Waiting for the Inbox page implementation to land in the first place:

- #381

PR #509, opened 2024-02-08
@-mention markers

sirpengi:
This PR is built on top of #510 

Fixes: #384 

Simulator Screenshot - iPhone 14 - 2024-02-08 at 16 50 20


lib/widgets/inbox.dart
```diff
@@ -340,6 +356,7 @@ class _DmItem extends StatelessWidget {
     return StickyHeaderItem(
       header: _AllDmsHeaderItem(
         count: allDmsCount,
+        isMentioned: isMentioned,
```
gnprice, line 359: This seems wrong — it's using the is-mentioned flag for the individual DM conversation, but passing it to use for the all-DMs header.

Hmm but I can't actually trigger any symptom from that bug.

… Aha. This header has no effect! That's a pre-existing latent bug in this file. What's happening is that the items of the list view are the entire sections — all DMs, then each stream — and not the individual conversations, so the `StickyHeaderItem` widgets on the individual conversations never get consulted.

I'll send a quick PR removing those `StickyHeaderItem`s, to avoid causing us similar confusion in the future. Then you can rebase atop that PR, and I think it will slightly simplify some of this code.

lib/widgets/inbox.dart
```diff
@@ -491,10 +518,25 @@ class _TopicItem extends StatelessWidget {
...
-              // TODO(#384) show @-mention indicator when it applies
+              if (isMentioned) const _AtMentionMarker(),
               Padding(padding: const EdgeInsetsDirectional.only(end: 16),
                 child: UnreadCountBadge(backgroundColor: subscription.colorSwatch(),
                   count: count)),
             ])))));
   }
 }
+
+class _AtMentionMarker extends StatelessWidget {
```
gnprice, line 529: These two commits are best squashed together:
ce34ec011 inbox: Show at-mention marker on topics
b6ac52216 inbox [nfc]: Pull out _AtMentionMarker

because the latter basically amounts to rethinking how to arrange some of the code added by the former, and so it's simplest to go straight to the revised way of arranging it.

gnprice: In fact let's squash those together with the next commit, handling streams:
ce34ec011 inbox: Show at-mention marker on topics
b6ac52216 inbox [nfc]: Pull out _AtMentionMarker
6e74d2a43 inbox: Show at-mention marker on streams

That way there isn't an intermediate state where we have @-mention markers on the individual topics but missing on the streams, which feels a little glitchy given expectations set by web. And conversely the squashed commit is still a quite reasonable size.

gnprice (commented):
Thanks @sirpengi! Generally this looks good. Small comments below.

lib/widgets/inbox.dart
gnprice: #510

sirpengi:

lib/widgets/inbox.dart
```diff
@@ -246,7 +257,7 @@ abstract class _HeaderItem extends StatelessWidget {
               overflow: TextOverflow.ellipsis,
               title))),
           const SizedBox(width: 12),
-          // TODO(#384) for streams, show @-mention indicator when it applies
+          if (hasMention) const _AtMentionMarker(),
```
gnprice, line 260: nit in commit message:

    inbox: Show at-mention marker on topics

Should say "topics and streams", since the same commit does it for streams too.

gnprice (commented):
Thanks for the revision! Generally this all looks good now, pending #510. One nit below.

sirpengi:
@gnprice Commit message updated per comment, now just waiting on #510

Issue #389, opened 2023-11-18
Inbox: Reduce some scrolling jank by making one `StickyHeaderListView` item per topic / DM conversation


chrisbobbe:
Instead of having a whole section (all-DMs or a stream) be an item.

Quoting Greg from https://github.com/zulip/zulip-flutter/pull/381#pullrequestreview-1735781702:

> As I mentioned in our call today, I played with this on my Pixel 5 to see the performance, using a test user on chat.zulip.org who has a ton of unreads. In debug mode the performance was pretty choppy — occasional pauses while scrolling that I believe reflected a new section scrolling into view, so that with the current version the whole section has to be built and laid out, which can be hundreds of items for different topics. Those pauses in debug mode were sometimes around a second long.
> 
> It's fine for debug mode to be choppy, though, so I tried it in profile mode. There there's still enough jank to be noticeable, but it's a lot milder — mostly a couple of dropped frames at a time, sometimes like 150ms. So this level of performance is fine for Beta 1, but it's something we'll want to revisit for Beta 2.
> 
> Probably that means I'll rework something in StickyHeaderListView to make it possible to have a separate list item for each conversation, just like you did in a previous draft, and without the glitchy interactions that that led to between one section's header and the next. (Effectively the current StickyHeaderListView assumes that each header is no taller than the item it belongs to; that's true in the message list, and it's true here by making each entire section into an item, but it's not true if each conversation is its own item.)

chrisbobbe:
> Probably that means I'll rework something in StickyHeaderListView to make it possible to have a separate list item for each conversation, just like you did in a previous draft, and without the glitchy interactions that that led to between one section's header and the next. (Effectively the current StickyHeaderListView assumes that each header is no taller than the item it belongs to; that's true in the message list, and it's true here by making each entire section into an item, but it's not true if each conversation is its own item.)

@gnprice do you know if there are still changes needed in `StickyHeaderListView` for this?

gnprice:
Yeah, I don't think this assumption has been lifted:
> Effectively the current StickyHeaderListView assumes that each header is no taller than the item it belongs to

PR #510, opened 2024-02-08
inbox [nfc]: Cut superfluous StickyHeaderItems on individual conversations

gnprice:
These were never getting consulted, because these aren't entire items of the enclosing list view; rather the items are the whole sections, i.e. either a whole stream or all DMs.  Because they weren't doing anything, they're confusing for understanding how this code works and making changes to it.

I think this happened because in an early draft of #381, the list-view items were individual conversations.  That's the arrangement that would be better for performance/smoothness reasons anyway, but it won't work until we make some changes to the sticky_header library; see #389.  When we go to do #389, we can always add these StickyHeaderItem widgets back.

### DIFF
--- a/lib/widgets/inbox.dart
+++ b/lib/widgets/inbox.dart
@@ -301,8 +301,6 @@ class _AllDmsSection extends StatelessWidget {
           return _DmItem(
             narrow: narrow,
             count: count,
-            allDmsCount: data.count,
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Ah interesting, thanks for noticing this and cleaning it up! Merged.

PR #511, opened 2024-02-09
content test: Share example data between parsing tests and widgets tests

gnprice:
The main goal here is to allow us to reuse in the widget tests in `test/widgets/content_test.dart` the example content blobs that we've written down in the parsing tests in `test/model/content_test.dart`. Currently we basically have to duplicate the examples, which is OK for short examples and gets more annoying as the example HTML gets longer than a line or so.

A nice followup would be to convert all our existing parsing tests into the new `ContentExample` form, and then use `testContentSmoke` to have smoke widget tests for all of them. (We could even add an "all content examples are tested" meta-test for widget tests, analogous to the one this PR adds for parsing tests.) But the conversions in this PR, focusing on the lengthiest HTML examples, are as much as I have time for at the moment. Meanwhile this provides the framework for future tests to be on `ContentExample` from the start.

Along the way this branch takes care of a few housekeeping items:
* sorting our dependencies alphabetically, to reduce the conflicts that come from every change adding to the end of the list;
* configuring IntelliJ (aka Android Studio) to suppress some spurious complaints, notably in the HTML of the content tests;
* and along the way to those two, cutting a couple of extraneous items left over from the `flutter create` template.


chrisbobbe:
Thanks, LGTM! I've added a small commit to use ContentExample for quotations, in the model tests, and also to add a corresponding widget smoke test. If I've understood the pattern and it looks good, please merge at will. 🙂

test/model/content_test.dart
```diff
@@ -1,16 +1,261 @@
...
+    const ImageEmojiNode(
+      src: '/static/generated/emoji/images/emoji/unicode/zulip.png', alt: ':zulip:'));
+
+  static const quotation = ContentExample(
+    'quotation',
+    "```quote\nwords\n```",
+    expectedText: 'words',
+    '<blockquote>\n<p>words</p>\n</blockquote>', [
+        QuotationNode([ParagraphNode(links: null, nodes: [TextNode('words')])])
+      ]);
```
gnprice, line 127: nit: this should get just one level of indentation
```suggestion
    '<blockquote>\n<p>words</p>\n</blockquote>', [
      QuotationNode([ParagraphNode(links: null, nodes: [TextNode('words')])])
    ]);
```

gnprice (commented):
Thanks! Yep, that looks good, with just an indentation nit. I'll fix that and merge.

Issue #184, opened 2023-06-14
Retry long-poll failures, with backoff


gnprice:
A Zulip client long-polls an event queue on the server to get events. In the current prototype, if that request fails for any reason, the long-poll loop just aborts:
```dart
  void poll() async {
    while (true) {
      final result = await getEvents(connection,
        queueId: queueId, lastEventId: lastEventId);
      // …
```

Instead, if the request fails with an error that we can expect to be transient, we should retry, with appropriate backoff.

split off as #514: [~~We should also ensure that if the request doesn't come back within an appropriate timeout, then we treat it as having failed and retry. The Zulip server should always respond to a long-poll within at most about 60 seconds, with [a heartbeat event](https://zulip.com/api/get-events#heartbeat) if there's no information to convey. So if it's been much longer than that and we haven't gotten a response, then we should assume none is coming — the request or response must have been lost somewhere. (This part might get split out as a separate issue.)~~]

For which types of errors we can expect to be transient and should therefore retry, see zulip-mobile at `tryFetch`:
```js
            if (!(shouldRetry && (e instanceof Server5xxError || e instanceof NetworkError))) {
              throw e;
```

For how to control backoff, see zulip-mobile at `BackoffMachine`, background discussion at https://github.com/zulip/zulip-mobile/pull/3841#issuecomment-588590747 , and the use of `BackoffMachine` in `tryFetch`. I think we can basically transcribe `BackoffMachine` from JS to Dart and reuse that.

gnprice:
This continues to happen sometimes, even after fixing #184 which covered the most common reason for failure. When it does, the symptoms are pretty confusing:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.22mark.20as.20read.22.20issue.20.28Flutter.29/near/1726533

So we should fix this.

The main complexity in the code here will be in doing appropriate backoff. But hopefully that will be a small amount of work, through reusing the work we did in zulip-mobile:

> I think we can basically transcribe `BackoffMachine` from JS to Dart and reuse that.

PR #515, opened 2024-02-13
Retry polling event queue always, with backoff

gnprice:
Fixes: #184 

Along the way, add some infrastructure to help us better write tests for asynchronous code, and in particular to better use `package:fake_async`.


chrisbobbe:
This now has some conflicts; would you mind resolving those?

gnprice:
Sure — done.

test/api/fake_api.dart
```diff
@@ -68,9 +68,9 @@ class FakeHttpClient extends http.BaseClient {
 
   @override
   Future<http.StreamedResponse> send(http.BaseRequest request) {
+    lastRequest = request;
     final response = _nextResponse!;
     _nextResponse = null;
-    lastRequest = request;
```
chrisbobbe, line 73:
```
fake_api: Record premature requests, too, in lastRequest
```

Ah I think I see: a request is "premature" because it's made before its response has been prepared. So the difference is that in those cases (the cases where `_nextResponse!` fails), the premature request will be recorded as `lastRequest`, vs. not recorded.

chrisbobbe (approved):
Thanks! This is an important fix. One small comment to check my understanding on something, then please merge at will.

test/api/fake_api.dart
gnprice: Yep, exactly.

gnprice:
Merged. Thanks for the review!

Issue #520, opened 2024-02-14
Notifications require manually giving permission, on Android


gnprice:
On a fresh install on Android, the app never asks for permission to show notifications, and as a result it never does successfully show a notification (at least on current versions of Android).

The workaround is for the user to go into the system's notification settings for the app (long-press the app icon > App info > Notifications) and enable notifications.

When we first landed Android notification support in #344/#320, I didn't think this problem was present; not sure if I missed it, or if something changed since then. It is something that's easy to miss in manual testing, because after you've touched the setting once, the issue doesn't arise unless you uninstall the app completely in order to install again from scratch.

I reproduce the issue on Android 14 on my Pixel 8. Because it'd be annoying to delete the app and lose my logins, I tweaked it to use a different app ID (and also a different label, to help me tell the different installs apart):
```diff
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -42,2 +42,3 @@ android {
     defaultConfig {
-        applicationId "com.zulip.flutter"
+//        applicationId "com.zulip.flutter"
+        applicationIdSuffix ".tmp"
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -4,2 +4,2 @@
    <application
-        android:label="Zulip beta"
+        android:label="Zulip beta 2"
```

We've had a couple of reports of this issue: [this chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications.20on.20zulip.20and.20zulip.20beta/near/1736488) was quite helpful, and there was also [this comment](https://github.com/zulip/zulip-flutter/issues/342#issuecomment-1942724263) on a different issue.

PR #521, opened 2024-02-14
notif android: Request permission at startup

gnprice:
This is necessary (on Android 13+) in order to be able to actually show a notification in the UI.

Fixes: #520


chrisbobbe:
Thanks for the catch and fix!

PR #523, opened 2024-02-15
deps: Upgrade Flutter, pods, and packages; automate upgrades more

gnprice:
The main commit is:

---

tools/upgrade: Add script to upgrade our dependencies

This leaves open one significant gap, which is upgrades to
our Flutter version.  I'm leaving that for another time
because it involves making changes in another repository
(the user/developer's Flutter checkout), and so requires care
to avoid clobbering any local changes they have there.

For updating the CocoaPods lockfiles after an upgrade to our
Dart dependencies in pubspec.lock, we'd previously been making
an actual build, or using `flutter build ios --config-only`
(and ditto for macOS) to strip that process down somewhat; but
then we've sometimes also had to run `pod update POD...` to
unwedge CocoaPods on certain pods, as in 105727e15 and several
later commits like 0dbc0c06d and most recently 1b0061e42.

That last step would have been annoying to automate.  But
fortunately it can be subsumed by simply `pod update` with no
list of pods, which causes CocoaPods to update all the pods.
Moreover, if we do that, it turns out to update some other
pods that we hadn't been updating: our transitive dependencies
that don't directly correspond to any Flutter plugin and
therefore don't get upgraded by changes in our pubspec.lock.
So start doing `pod update` on its own, too, even apart from
`flutter pub upgrade`.

And then in fact it turns out that a full `pod update` also
subsumes all the changes that `flutter build` had been making,
so it replaces that in the followup to `flutter pub upgrade`.

---

Two of the later commits are then generated by that tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,10 +161,20 @@ To update the version bounds:
 When adding or upgrading dependencies, try to keep our generated files
 updated atomically with them.
 
-In particular the files `ios/Podfile.lock` and `macos/Podfile.lock`
+In particular the CocoaPods lockfiles
+`ios/Podfile.lock` and `macos/Podfile.lock`
 frequently need an update when dependencies change.
-To update those, run (on a Mac) the commands
-`flutter build ios --config-only && flutter build macos --config-only`.
+This can only be done in a macOS development environment.
+
+If you have access to a Mac,
+then for upgrading dependencies, use the script `tools/upgrade`.
+Or after adding a new dependency, run the commands
+`(cd ios && pod update) && (cd macos && pod update)`
+to apply any needed updates to the CocoaPods lockfiles.
+
+If you don't have convenient access to a Mac, then just mention
+clearly in your PR that the upgrade needs syncing for CocoaPods,
+and someone else can do it before merging the PR.
 
 (Ideally we would validate these automatically in CI: [#329][].
 Several other kinds of generated files are already validated in CI.)
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -213,7 +213,7 @@ SPEC CHECKSUMS:
   GoogleDataTransport: 57c22343ab29bc686febbf7cbb13bad167c2d8fe
   GoogleUtilities: 0759d1a57ebb953965c2dfe0ba4c82e95ccc2e34
   image_picker_ios: 99dfe1854b4fa34d0364e74a78448a0151025425
-  integration_test: 13825b8a9334a850581300559b8839134b124670
+  integration_test: ce0a3ffa1de96d1a89ca0ac26fca7ea18a749ef4
   nanopb: d4d75c12cd1316f4a64e3c6963f879ecd4b5e0d5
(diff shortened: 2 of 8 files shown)


chrisbobbe:
When I ran `tools/upgrade` at the tip of this branch (actually after rebasing onto a slightly newer main; 58fcc74), the `tools/upgrade pod` made changes like this:

```diff
diff --git ios/Podfile.lock ios/Podfile.lock
index 6a62a69bd..c2c13353f 100644
--- ios/Podfile.lock
+++ ios/Podfile.lock
@@ -37,19 +37,19 @@ PODS:
   - file_picker (0.0.1):
     - DKImagePickerController/PhotoGallery
     - Flutter
-  - Firebase/CoreOnly (10.20.0):
-    - FirebaseCore (= 10.20.0)
-  - Firebase/Messaging (10.20.0):
+  - Firebase/CoreOnly (10.18.0):
+    - FirebaseCore (= 10.18.0)
+  - Firebase/Messaging (10.18.0):
     - Firebase/CoreOnly
-    - FirebaseMessaging (~> 10.20.0)
-  - firebase_core (2.25.4):
-    - Firebase/CoreOnly (= 10.20.0)
+    - FirebaseMessaging (~> 10.18.0)
+  - firebase_core (2.24.2):
+    - Firebase/CoreOnly (= 10.18.0)
     - Flutter
-  - firebase_messaging (14.7.15):
-    - Firebase/Messaging (= 10.20.0)
+  - firebase_messaging (14.7.10):
+    - Firebase/Messaging (= 10.18.0)
     - firebase_core
     - Flutter
-  - FirebaseCore (10.20.0):
+  - FirebaseCore (10.18.0):

[etc.]
```

Is there a caching situation that we should be aware of? I was surprised that a `tools/upgrade` command would make a commit that downgraded anything. These downgrades ended up being reverted in the `tools/upgrade pub` commit.

gnprice:
Hmm. What if you do `flutter pub get` before the `tools/upgrade`?

Specifically I suspect `pod update` (inside `tools/upgrade`) is working from a file that `flutter pub get` updates based on `pubspec.lock`.

If that's what's happening, then probably `tools/upgrade pod` needs to start by doing `flutter pub get`.

chrisbobbe:
Indeed; I tried this again, being sure to run `flutter pub get` before the `tools/upgrade`. A `tools/upgrade pod` commit wasn't produced this time.

> If that's what's happening, then probably `tools/upgrade pod` needs to start by doing `flutter pub get`.

Cool; I'll await that change, then.

gnprice:
Great. Fix made.

chrisbobbe:
Ah it looks like shellcheck is giving some feedback, failing CI.

gnprice:
… Huh indeed. Fixed.

chrisbobbe:
Yeah, perhaps. It may not be too bad if something like this gets caught in CI but not locally, though. Even the core team doesn't spend tons of time working on the shell scripts. 🙂

Merging.

gnprice:
True. Thanks for the reviews!

Issue #516, opened 2024-02-13
Go directly on launch to inbox / home view


gnprice:
Currently when you launch the app, you get the "Choose account" screen; then when you choose an account, there's a loading screen followed by the "Home" screen, the one with "🚧 Under construction 🚧" and several buttons; then you can tap one of those like "Inbox" to get to a view that actually tells you about what messages you have.

This is a bit cumbersome, and particularly so because that loading screen before "Home" can take several seconds as it's loading the whole register-queue response from the server. (It's several seconds for a large realm like chat.zulip.org from a good network connection; a fraction of a second for a small realm; and can of course be longer on a slow network connection.) So you launch the app, tap a button, then have to wait several seconds before you can tap the next button to get to the inbox.

It'd be good to instead open directly to whatever account you were last looking at (bypassing the choose-account screen) and to the inbox or another informative home view (bypassing the "Home" screen). See [chat discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/zulip.20flutter.20login.20screen/near/1714661), and later [followup](https://chat.zulip.org/#narrow/stream/48-mobile/topic/zulip.20flutter.20login.20screen/near/1735077).

Specifically, a minimal version of this can just always go to the inbox. A future extension might be to let the user choose between Inbox, Recent conversations, and All messages, akin to the setting https://chat.zulip.org/help/configure-home-view used on web; we can track that as a separate issue.

A couple of things we'll want to take care with in doing this:

* We'll need to make sure the user can easily get to the other views and to other accounts, by navigating "back" out of the initial screen. This should be straightforward to do by putting "Choose account" and "Home" on the navigation stack from the beginning.

  (Later we'll do something else for navigation — #311 — but that will work well enough for now.)

* Because the new initial screen requires server data, it'll initially just display as a loading screen. If that server can't be reached, the loading may never finish. In case what the user actually wanted was to use some other server, we'll want to make sure they can navigate away even while loading is ongoing.

The latter point probably means adjusting the loading screen so that it has an app bar which can hold a back button, even if there's little other meaningful information on the screen. We have this TODO already in the implementation of `PerAccountStoreWidget`:
```dart
  Widget build(BuildContext context) {
    // TODO: factor out the use of LoadingPage to be configured by the widget, like [widget.child] is
    if (store == null) return const LoadingPage();
```
[...]

Issue #386, opened 2023-11-17
InboxPage: Design / implement appearance when empty


chrisbobbe:
When the user doesn't have any unreads, `InboxPage` (titled "Inbox") should have a message saying so, instead of just being blank:

<img width=300 src="https://github.com/zulip/zulip-flutter/assets/22248748/557d4997-cb90-49ac-ae60-5a585d48de3c">

### Design

The design for this is in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5966-176662&m=dev

Screenshot:
![image](https://github.com/user-attachments/assets/28c4d5e1-3ca5-477f-a244-d77e9e61b006)

### Related issues
- #385

jassirr:
@chrisbobbe can i take this up? and can you please provide me with the clarity of what actually needs to be done here.
Thanks :).

jassirr:
@gnprice any update for above discussion?

chrisbobbe:
I think this issue isn't ripe to work on yet. We'll want our designer to suggest how this should look.

baimamboukar:
Hello @chrisbobbe @gnprice 

I would like to work on this issue, please. Thanks for assigning it to me if it has not been addressed yet.

gnprice:
This issue isn't ripe to work on yet; see the previous comment https://github.com/zulip/zulip-flutter/issues/386#issuecomment-1897854406 .

alya:
Note from user feedback: we should make sure that suggested next steps / where to find messages are clear on this screen. A user reported being at first worried that all their messages have disappeared when viewing the empty inbox screen in the legacy app.

chrisbobbe:
Alya [mentioned](https://chat.zulip.org/#narrow/channel/48-mobile/topic/iOS.20feedback.20for.20Julia.20community.20member/near/1964130) on CZO:

> For reference, the web app includes a link to recent conversations. I guess either the combined feed or the channels list could work in the mobile app, since we don't have a recent conversations view?

<img width=300 src="https://github.com/user-attachments/assets/2a25659d-4cda-4847-ac94-f896b0f74bd0">

We have #315 for Recent Conversations in zulip-flutter. But that's milestoned post-launch, which is later than this issue. So I think a first iteration that uses Combined Feed or Channels makes sense.

gnprice:
This issue is ready now for someone to pick up; see the design in the description.

[...]

PR #525, opened 2024-02-17
Go straight to inbox on launch, when an account available

gnprice:
Fixes: #516

Along the way, make some improvements to our abstractions for describing a page that needs a `PerAccountStoreWidget` (like `MaterialAccountWidgetRoute`). In particular:
* extend them to cover the case where the account ID we want isn't found ambiently on the current context (and then use them more comprehensively, as we now can do);
* have them tell `PerAccountStoreWidget` to use a loading placeholder that has a `Scaffold` and therefore a back button.


chrisbobbe:
This looks great! Merging.

It probably bumps the priority of #386, because the inbox page will be featured more prominently now. But even an empty inbox page has a title and a back button, so users shouldn't be totally stuck, and we do still clearly mark the app as beta.

gnprice:
Thanks!

> It probably bumps the priority of #386, because the inbox page will be featured more prominently now.

Good point, yeah, agreed.

Issue #458, opened 2023-12-21
Update database with feature level etc. on each register


gnprice:
On the Account class, corresponding to a row in our local database, we have several fields for values that may in reality change over time: most notably `zulipFeatureLevel`, also `zulipVersion` and `zulipMergeBase`.

Currently we take the values of those fields when first logging in, and don't ever update them later. We should do so. In particular when we have compatibility logic that conditions on `zulipFeatureLevel`, that will allow that logic to correctly use the server's current version rather than the version it had when we first logged in (perhaps months ago). Discussion here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/zulipFeatureLevel.20out.20of.20sync/near/1708193

Specifically this issue is for updating these values using the `registerQueue` response. We should also do so on `restart` events; that's tracked as part of #135.

I have a draft branch for this which I wrote one morning a few months ago and then didn't finish. (That branch handled `restart` events, which aren't so much a priority in themselves, but most of the work involved was for the ability to update account records in the database at all.)

gnprice:
Another effect of this is that it will serve as an early small trial run of the sort of code changes we'll need for #477, because it'll involve writing changes into the database in tandem with making them in our Dart data structures.

gnprice:
> I have a draft branch for this which I wrote one morning a few months ago and then didn't finish

I've just sent #526, built around one commit from that branch but then with 14 other commits of followups it led to.

I've also rebased the branch as a whole and updated the rest of it a bit, and have it still as a draft branch.

gnprice:
In baea6d74627047ea69d947c3e662341e15af76de / #701 I wrote that with this issue #458 we'd want `PerAccountStore.handleEvent` to write changes to the database on certain events, and so that method needed to become async.

That wasn't quite right (or was imprecise, I guess, depending how one reads "with"): this issue only needed us to write such changes on register, not on any events. Hence this issue was completed last year, while we don't yet actually write any database changes caused by events (cf https://github.com/zulip/zulip-flutter/pull/1327#discussion_r1957206592).

The part that will call for database changes in handling events is when we handle `restart` events:
* #1271

Issue #322, opened 2023-10-16
Dedupe sending notification token to server


gnprice:
In my initial implementation of #122, I plan to have the app just tell the server its notification token again and again on each launch.

As a follow-up, we should have it instead remember when it's already successfully told the server its token, and then not tell it again. We do this in zulip-mobile.

Actually, there's a refinement to make. Google's docs recommend that the client should repeat itself to the server, but only after a month:
https://firebase.google.com/docs/cloud-messaging/manage-tokens#ensuring-registration-token-freshness
That then allows the server to conclude when a token hasn't been re-registered in, say, two months, that it's stale — that the client has gone away — and Google recommends that the server stop sending notifications to it in that case. (I don't recall seeing that recommendation the last time I studied these docs, several years ago. Not sure if Apple also recommends something similar these days.)

gnprice:
One wrinkle in this plan: in the future, with end-to-end encryption of push notifications, we expect it to be possible for a client to have working push notifications, but then a realm admin changes a setting on the server, and now the client no longer gets push notifications (even though they remain logged in and able to use Zulip through opening the app as usual). Details here:
https://github.com/zulip/zulip-flutter/issues/323#issuecomment-2756064616

That specific scenario will only apply when the client app is a version from before the implementation of end-to-end encrypted notifications. I'm not sure if we'll have other scenarios of this kind in the future.

In any case this probably means that at least before we implement E2EE notifications, we should dedupe these requests only when the last one was within a short window, like the past day. That way (once we have #323) the user will still learn reasonably quickly if their push notifications have stopped working because the realm admins started requiring E2EE and they need to update the app.

PR #526, opened 2024-02-21
Prepare for updating accounts in database

gnprice:
Currently we insert `Account` records into the database on login, and never update them. We'll need to update them in order to do:
* #458
* #322

and other things.

This branch doesn't yet make any new changes in the database, but it does some refactoring on PerAccountStore and friends which we'll need in order to do so. It also takes care of various things that became noticeable along the way:
* adding some handy getters like `store.selfUserId` and `self.realmUrl`;
* fixing where we'd been crashing if the server somehow sent us an invalid URL (one `Uri.parse` rejects) in the HTML or as a user's avatar;
* adding more content-widget tests, to test that latter change, and using the facilities recently added in #511.

The main commit is this one near the beginning:

---

store [nfc]: Keep just one copy of Account object

We're going to want to update these, when things like
the server version or the acked push token change.
To avoid endless bugs, it'll then be important that there
be just one copy to be updated.

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -9,7 +9,6 @@ import '../api/model/model.dart';
 import '../model/binding.dart';
 import '../model/content.dart';
 import '../model/internal_link.dart';
-import '../model/store.dart';
 import 'code_block.dart';
 import 'dialog.dart';
 import 'icons.dart';
@@ -261,10 +260,11 @@ class MessageImage extends StatelessWidget {
     final src = node.srcUrl;
 
     final store = PerAccountStoreWidget.of(context);
-    final resolvedSrc = resolveUrl(src, store.account);
+    final resolvedSrc = store.tryResolveUrl(src);
+    // TODO if src fails to parse, show an explicit "broken image"
 
     return GestureDetector(
-      onTap: () {
+      onTap: resolvedSrc == null ? null : () { // TODO(log)
         Navigator.of(context).push(getLightboxRoute(
           context: context, message: message, src: resolvedSrc));
       },
@@ -281,7 +281,7 @@ class MessageImage extends StatelessWidget {
               child: SizedBox(
                 height: 100,
                 width: 150,
-                child: LightboxHero(
+                child: resolvedSrc == null ? null : LightboxHero(
                   message: message,
                   src: resolvedSrc,
                   child: RealmContentNetworkImage(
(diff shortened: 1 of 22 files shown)


gnprice:
The CI failure is a pair of analyzer failures that appear with a Flutter update today. Chris noticed them earlier today and is working on a PR:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20some.20.60ColorScheme.60.20fields.20deprecated/near/1741410

So this PR is otherwise ready, and should pass CI once it's rebased atop that upcoming other PR.

lib/widgets/content.dart
```diff
@@ -730,12 +730,13 @@ class MessageImageEmoji extends StatelessWidget {
...
-            width: size,
-            height: size,
-          )),
+          child: resolvedSrc == null ? const SizedBox.shrink() // TODO(log)
+            : RealmContentNetworkImage(
+                resolvedSrc,
+                filterQuality: FilterQuality.medium,
+                width: size,
+                height: size,
+              )),
```
chrisbobbe, line 739: For emoji reactions, if the src doesn't parse, we fall back to showing the emoji name. (We use the same fallback if the src parses but the image fails to load.) This would be nice to do, or we could add a TODO for it.

I wonder if it would be hard to factor the emoji-reaction widget code into something to also use in content. Another thing I notice is that `_UnicodeEmoji` in widgets/emoji_reaction.dart has careful logic so that the emoji isn't awkwardly small on iOS. (In content, unicode emojis are awkwardly small on iOS.)

chrisbobbe (commented):
Thanks! Just one small comment below.

lib/widgets/content.dart
gnprice: Yeah, it'd be good to reuse the work that went into those reaction emoji so that emoji in content also get the benefit of it. It looks like features that would bring include:
* making Unicode emoji bigger on iOS, counteracting that Flutter-and/or-iOS bug;
* disabling animations based on the platform setting;
* text fallbacks based on the user's emojiset choice;
* text fallbacks when an image doesn't work out.

At the moment we don't have an implementation of emoji-as-text-name at all for content, so I think doing that as a fallback is out of scope for this PR. The URL failing to parse should happen rarely to never — it'd be a bug in the server, and doesn't seem like a real likely kind of bug.

chrisbobbe: Makes sense.

chrisbobbe:
Did you mean to comment further or merge this? I see you force-pushed after the discussion above.

gnprice:
Ah, those pushes were just rebasing and resolving conflicts. I didn't make any changes in content, just replied to your comment above. Shall we merge this?

chrisbobbe:
Sure, please go ahead!

PR #527, opened 2024-02-21
ui: Stop using deprecated `ColorScheme.background` and `ColorScheme.surfaceVariant`

chrisbobbe:
After upgrading Flutter I get this feedback from the analyzer:

```
Analyzing zulip-flutter...                                              

   info • 'background' is deprecated and shouldn't be used. Use surface instead. This feature
          was deprecated after v3.18.0-0.1.pre • lib/widgets/app.dart:105:9 •
          deprecated_member_use
   info • 'surfaceVariant' is deprecated and shouldn't be used. Use surfaceContainerHighest
          instead. This feature was deprecated after v3.18.0-0.1.pre •
          lib/widgets/compose_box.dart:815:26 • deprecated_member_use
```

Discussion on CZO: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20some.20.60ColorScheme.60.20fields.20deprecated/near/1741410

This fixes those analyzer issues. 

While investigating, I discovered that the same upstream change also did something else: it re-introduced a color change in our default app bar when content scrolls under it.

That's fixed straightforwardly by setting a constant background color ourselves, instead of deferring to Material's defaults. We choose a background color from the Figma, naturally…and since that color is very similar to the scaffold background color, also from the Figma, I go ahead and apply the Figma's bottom border to the app bar. There's some subtlety around removing that bottom border for some cases in the message list. Here are some screenshots of the relevant cases.

I also set the appropriate app-bar background for DM narrows, following the Figma.

| Comment | Before | After |
| --- | --- | --- |
| Default app bar (Inbox, no unreads; incidentally shows #386) | image | image |
| Default app bar (Inbox, with unreads) | image | image |
| DM narrow | image | image |
| Stream narrow (no change intended) | image | image |
| All-messages narrow | image | image |


lib/widgets/message_list.dart
```diff
@@ -56,21 +57,32 @@ class _MessageListPageState extends State<MessageListPage> {
...
       case TopicNarrow(:final streamId):
         backgroundColor = store.subscriptions[streamId]?.colorSwatch().barBackground
-          ?? _kFallbackStreamColor;
+          ?? _kUnsubscribedStreamRecipientHeaderColor;
+        // All recipient headers will match this color; remove distracting line
+        // (but are recipient headers even needed for topic narrows?)
+        removeAppBarBottomBorder = true;
       case DmNarrow():
-        backgroundColor = _kFallbackStreamColor;
+        backgroundColor = _kDmRecipientHeaderColor;
```
gnprice, line 73: Ah good catch, thanks. I definitely see the difference in the DM narrow's app bar in the screenshots, and this looks better.

lib/widgets/message_list.dart
```diff
@@ -56,21 +57,32 @@ class _MessageListPageState extends State<MessageListPage> {
...
+        backgroundColor = null; // i.e., inherit
       case StreamNarrow(:final streamId):
       case TopicNarrow(:final streamId):
         backgroundColor = store.subscriptions[streamId]?.colorSwatch().barBackground
-          ?? _kFallbackStreamColor;
+          ?? _kUnsubscribedStreamRecipientHeaderColor;
+        // All recipient headers will match this color; remove distracting line
+        // (but are recipient headers even needed for topic narrows?)
+        removeAppBarBottomBorder = true;
       case DmNarrow():
```
gnprice, line 72: nit: separate the cases (here and above) visually, now that they're longer:
```suggestion
        removeAppBarBottomBorder = true;

      case DmNarrow():
```

gnprice (commented):
Thanks for taking care of these! All looks good, with one nit; then please go ahead and merge.

chrisbobbe:
Thanks! Done.

Issue #528, opened 2024-02-23
Notifications missing in release build on Android


gnprice:
After the PR #521 which fixed #520, I can no longer reproduce any trouble receiving notifications on Android, regardless of whether the app is already running in the foreground, in the background, or not at all… so long as I'm using a debug build of the app.

But when I run a release build of the app instead — which of course is what we distribute to users — I don't get any notifications at all. Not when the app is in the foreground, not when it's in the background, and not when it's not running.

I'm investigating this now. A quick first observation is that it's unaffected by the `@pragma('vm:entry-point')` change suggested at https://github.com/zulip/zulip-flutter/issues/342#issuecomment-1874716442 ; it reproduces the same way after I add the pragma.

gnprice:
Hmm! When I reproduce the issue while the app is in the foreground, if I then go to the notification history (system settings > Notifications > Notification history), the notification does appear there! It's under the heading "Recently dismissed".

Still a mystery why the actual visible notification never shows up. But that does mean that our request to show a notification is reaching the relevant Android API. And adding debug logging confirms that the whole chain is getting successfully executed:
* FCM message (about a notification) reaching the app's broadcast receiver
* → various `package:flutter_firebase` logic within Java
* → corresponding logic in Dart
* → our notifications code in Dart
* → `package:flutter_local_notifications` code in Dart
* → corresponding logic in Java
* → the point where that code calls the relevant Android API method, `androidx.core.app.NotificationManagerCompat.notify`.

In light of that, I can also now observe that `vm:entry-point` pragma making a difference. Without the pragma (i.e. from main), when a notification arrives while in the *background*, these errors appear on the `flutter run --release` console:
```
E/flutter ( 2640): [ERROR:flutter/lib/ui/dart_runtime_hooks.cc(38)] Dart Error: Dart_LookupLibrary: library 'package:zulip/notifications.dart' not found.
E/flutter ( 2640): [ERROR:flutter/shell/common/shell.cc(117)] Dart Error: Dart_LookupLibrary: library 'package:zulip/notifications.dart' not found.
E/flutter ( 2640): [ERROR:flutter/shell/common/shell.cc(117)] Dart Error: Dart_LookupLibrary: library 'package:zulip/notifications.dart' not found.
```
and the information about the notification never reaches Dart, and no notification appears in the system notification history. When I add that pragma, those errors don't appear, and the behavior when in the background is just like the behavior in the foreground.

[...]

PR #529, opened 2024-02-23
notif: Add `vm:entry-point` pragma to _onBackgroundMessage

gnprice:
This fixes part of #528, following the suggestion @n-92 made here:
  https://github.com/zulip/zulip-flutter/issues/342#issuecomment-1874716442

Even with this change, notifications still don't yet work on Android in release builds.  But after this change, when the app wasn't in the foreground, they get farther than before.  Specifically they get just as far as they get, with or without this change, when the app was in the foreground.  Details here:
  https://github.com/zulip/zulip-flutter/issues/528#issuecomment-1960646800

### DIFF
--- a/lib/notifications.dart
+++ b/lib/notifications.dart
@@ -174,6 +174,12 @@ class NotificationService {
     _onRemoteMessage(message);
   }
 
+  // This pragma `vm:entry-point` is needed in release mode, when this method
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Great, thanks! Merged, and looking forward to the rest of the fix for #528.

PR #532, opened 2024-02-23
notif android: Keep the notification icon from being resource-shrunk away

gnprice:
Details in the new comment and in the issue thread.

Fixes: #528


chrisbobbe:
Interesting! Thanks for the fix; merging.

Issue #301, opened 2023-09-07
Request avatars of the appropriate size


chrisbobbe:
With #195 fixed with PR #287, we now have a user-profile screen! 🎉

I noticed that on many profiles on CZO, the avatar is more blurry than on the corresponding screen in `zulip-mobile`:

| zulip-flutter | zulip-mobile |
| --- | --- | 
| ![image](https://github.com/zulip/zulip-flutter/assets/22248748/66896560-9b27-4f0d-813d-2164e7aaafca) | ![image](https://github.com/zulip/zulip-flutter/assets/22248748/644603ff-c376-45b4-b50e-dc0f0d1839bc) |

This is because we don't yet have special logic for requesting a larger image when we need one. This issue is for adding such logic.

In `zulip-mobile`, we made an `AvatarURL` class with a `get` method that gives you a URL for an image of the specified size in physical pixels (or at least the best possible approximation). `AvatarURL` has [three subclasses](https://github.com/zulip/zulip-mobile/blob/09754f0a5d7408fd490a631339eced3cf77a28e0/src/utils/avatar.js) that each implement `get` very differently.

On CZO in `#api documentation`, I've [summarized](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/avatar_url.3A.20Request.20specific.20size/near/1636546) zulip-mobile's logic for determining a URL for any given size in physical pixels. I'm hoping the API doc can start specifying that logic, and we can use the new documentation for zulip-flutter's implementation.

To convert a logical-pixel value to a physical-pixel value, we'll want to multiply by the device pixel ratio. As in zulip-mobile:
```js
avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size))
```

I think a solution will likely have lots of interactions with #254 and #255, so the best approach might be to consider these three issues together.

Related: #254
Related: #255

chrisbobbe:
An observant user noticed the blurriness and reported it in chat today: https://chat.zulip.org/#narrow/stream/48-mobile/topic/flutter.3A.20blurry.20profile.20picture/near/1735359

saisumith770:
@gnprice I'm currently working on this feature. Finished with most of the task using gravatar's. I would like to know which formatter you are using because as I save my code most of the file gets re-formatted causing several changes in the git history. It would be really helpful if you linked the contributing guidelines as well.

gnprice:
@saisumith770 Cool — assigned you the issue, and I'll look at your PR in the next few days.

[...]

Issue #255, opened 2023-08-03
Compute Gravatar URLs where needed (handle `client_gravatar` true)


gnprice:
We should properly handle this optimization:
https://zulip.com/api/register-queue#parameter-client_gravatar

At the moment we don't pass that `client_gravatar` parameter, so we get the default behavior of `true`, which signs us up for the server to pass `null` as a user's `avatar_url` to express that it's a Gravatar URL we can compute from the email address. But we haven't actually written down the bit of code to compute that Gravatar URL, so we end up with a missing avatar.

I think we'll shortly flip that for the moment to pass `client_gravatar` as false, so that we get correct behavior. But we should follow up by actually implementing the client side of that — it's not hard — and then re-enabling the optimization.

For code demonstrating the logic, see zulip-mobile's `GravatarURL` in [src/utils/avatar.js](https://github.com/zulip/zulip-mobile/blob/c979530d6804db33310ed7d14a4ac62017432944/src/utils/avatar.js).

Another piece we'll need as part of implementing this issue is to deserialize the `User.avatarUrl` field in a way that distinguishes whether the server set the `avatar_url` field to null or left it out entirely. The former indicates a Gravatar, but the latter comes from `user_avatar_url_field_optional` (#254) and carries no information about what the avatar is.

That will likely involve a class `AvatarUrl`, with subclasses for a provided URL string or indicating that the avatar is a computable Gravatar or that it was omitted. This resembles the `AvatarURL` class hierarchy in zulip-mobile — but one thing I would like to do differently here is to avoid putting redundant information on these avatar-URL objects, in order to avoid unnecessary allocation. (After all, the list of users is a fairly hot path both for startup CPU time and for memory consumption, when in a large realm.) It's fine for the `AvatarUrl` value to only be possible to fully interpret by using other data from the `User`.

gnprice:
> I think we'll shortly flip that for the moment to pass `client_gravatar` as false,

(This was done in caa9603784e1a7a7aee0af4682553eed6b1d0033 in #249. Implementing the optimization, which is what the present issue is about, is still open.)

gnprice:
> Another piece we'll need as part of implementing this issue is to deserialize the `User.avatarUrl` field in a way that distinguishes whether the server set the `avatar_url` field to null or left it out entirely.

[...]

Issue #254, opened 2023-08-03
Handle avatar-URL optimization `user_avatar_url_field_optional`


gnprice:
That is, the `user_avatar_url_field_optional` client capability in the Zulip API:
https://zulip.com/api/register-queue#parameter-client_capabilities

At the moment, we claim that capability in our registerQueue requests; but we don't actually implement what we need on our side in order to handle it, so we end up with missing avatars. We'll shortly flip that so that we no longer claim the capability, getting correct behavior. But then we should actually implement the logic (there's not much to it), and re-enable the optimization.

For how the implementation can work, see `FallbackAvatarURL` in zulip-mobile's [src/utils/avatar.js](https://github.com/zulip/zulip-mobile/blob/c979530d6804db33310ed7d14a4ac62017432944/src/utils/avatar.js). In particular see `validateAndConstructInstance` and `get`. I think we can probably do the implementation with less code than is in zulip-mobile, though.

At the moment our `User.avatarUrl` field gets deserialized in a way that doesn't distinguish whether the server left `avatar_url` out or set it to null. We'll need to distinguish those in order to handle `client_gravatar` (a separate issue)… but the fallback avatar URLs should work for all users. So for this issue we should be fine without that distinction.

gnprice:
> We'll shortly flip that so that we no longer claim the capability, getting correct behavior.

(This was done in caa9603784e1a7a7aee0af4682553eed6b1d0033 in #249. Implementing the optimization, which is what the present issue is about, is still open.)

NishchayRajput:
Hi, I noticed that this issue is mentioned inside Zulip Flutter. Should the fix be implemented in Zulip Flutter or Zulip Mobile?
Since the reference code (FallbackAvatarURL) is from Zulip Mobile, should we replicate a similar implementation in Zulip Flutter, or does it need changes in both apps?

Abdelmohaymn:
Hi! I would love to pick this up. I've reviewed the **FallbackAvatarURL** implementation in **zulip-mobile/src/utils/avatar.js**. 
I plan to port this logic to Dart by constructing the **/avatar/{user_id} (and /medium)** URLs when the **avatar_url** is absent. 
Once the fallback generation is in place, I will re-enable the **user_avatar_url_field_optional** capability in the register request. I'll also add the necessary tests. Can I go ahead and claim this?

PR #534, opened 2024-02-25
Fix avatar blurriness in profile screen

saisumith770:
## Description
Updated the avatar URL used in the profile screen to request a higher-resolution image to reduce blurriness.
Refer to #301 for more details.

### Gravatar Image:
WhatsApp Image 2024-02-25 at 16 20 00_eab8d2d3

### Uploaded Image:
WhatsApp Image 2024-02-25 at 16 20 00_f8f05f9e


saisumith770:
@gnprice could you review this PR and let me know if any changes are required.

gnprice (commented):
Thanks for the contribution!

Before we can review this in detail, it will need tests, and both the new and existing tests will need to be passing.

On running the tests and writing tests, see the project README.

For this code, because it's based on existing code in zulip-mobile, it would be good to also transcribe that code's tests and make sure that this new code passes all the same tests. If there are any tests of the zulip-mobile code that for some reason don't make sense in this version, then note that with a comment.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,148 @@
...
+
+/// Pixel dimensions of different size choices we have (they're all
+/// square) when requesting an uploaded avatar.
+/// DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
+const int defaultUploadSizePx = 100;
+
+/// MEDIUM_AVATAR_SIZE in zerver/lib/upload.py.
+const int mediumUploadSizePx = 500;
+
+abstract class AvatarURL {
```
gnprice, line 16: nit: follow [camel-casing rules](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#capitalize-identifiers-consistent-with-their-spelling):

```suggestion
abstract class AvatarUrl {
```

I know the corresponding names in zulip-mobile are spelled like "AvatarURL", but we're being more consistent in zulip-flutter.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,148 @@
...
+  static AvatarURL fromUserOrBotData({
```
gnprice, line 17: nit:
```suggestion
  static AvatarURL fromUserData({
```

The "or bot" thing in zulip-mobile corresponds to the `UserOrBot` type there, which in turn is a legacy of some quirks in the Zulip API that zulip-mobile only partly papers over and zulip-flutter does so more thoroughly. In zulip-flutter, the corresponding type is just `User`.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,148 @@
...
+    if (sizePhysicalPx > defaultUploadSizePx) {
+      result = Uri.parse('$standardUri/medium');
+    }
+
+    return result;
+  }
+}
+
+class UploadedAvatarURL extends AvatarURL {
+  static UploadedAvatarURL validateAndConstructInstance({
```
gnprice, line 98: The Dart idiom here is to make this a factory constructor:
```suggestion
  factory UploadedAvatarURL({
```

And then the other constructor below can be private:
```dart
  UploadedAvatarURL._({required this.standardUri});
```

See our `Unreads` for an example. Or find other examples with: `git grep -P '\w+\._\('`

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,148 @@
...
+const int mediumUploadSizePx = 500;
+
+abstract class AvatarURL {
+  static AvatarURL fromUserOrBotData({
+    String? rawAvatarUrl,
+    required Uri realm,
+    required int userId,
+    required String email,
+  }) {
+    if (rawAvatarUrl == null || rawAvatarUrl.isEmpty) {
```
gnprice, line 23: Why the special condition for an empty string?

In general the empty string is a string like other strings; it's not the same thing as `null` and generally shouldn't be treated the same.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,148 @@
...
+      return GravatarURL.validateAndConstructInstance(email: email);
+    } else if (rawAvatarUrl.startsWith(GravatarURL.origin)) {
+      return GravatarURL.validateAndConstructInstance(
+          urlFromServer: rawAvatarUrl, email: email);
+    } else if (['.png','.jpg','.jpeg'].contains(extension(rawAvatarUrl))) {
```
gnprice, line 28: What's the basis for this condition?

gnprice:
> When `User.avatarUrl` is null, we should handle that by computing a gravatar.

Ah, no, I've now gone and reread #255 and I see that's not the correct behavior here.

Currently there should not be any null values in `User.avatarUrl`. (There will be some after #254 and/or #255.) Therefore, if we see a null value there, we should just show `SizedBox.shrink()` as we do now.

If we were to turn on both of the features described in #254 and #255 (and not do anything else), then we'd have two different cases *both* show up as null there: the case where we should compute a gravatar, and the case where we should compute a fallback avatar URL. That's because one of those would be represented by `"avatar_url": null` in the JSON, and the other by `avatar_url` being absent in the JSON. Both of those conditions are, by default, rendered as `avatarUrl: null` by Dart's JSON deserializer. (Which is a good thing — it stems from the fact that Dart does not share JS's distinctive bad design choice of having two different values `undefined` and `null`.)

When we do go to take care of those two issues, then, we'll need to add a wrinkle in the `User` class to represent that distinction. That's all out of scope for this issue, so for this PR let's leave out both the gravatar case and the fallback case.

test/model/avatar_url_test.dart
```diff
@@ -0,0 +1,147 @@
...
+        userId: egUser.userId,
+        email: egUser.email,
+      ))
+      .has((obj) => obj.get(defaultSz), 'computed uri')
+      .has((uri) =>
+        uri.toString() == avatarUrl,
+        'verify avatar',
+      )
+      .equals(true);
+    });
```
gnprice, line 89: Let's use `check` in the following style instead:
```suggestion
  group('UploadedAvatarURL', () {
    test('png image url', () {
      const url = 'https://zulip.example/image.png';
      final avatarUrl = AvatarURL.fromUserOrBotData(
        rawAvatarUrl: url,
        realm: egUser.realm,
        userId: egUser.userId,
        email: egUser.email,
      );
      check(avatarUrl.get(defaultSz).toString()).equals(url);
    });
```

In general `has` is good for defining extension getters on `Subject`, as seen in many examples in our test suite, but rarely inside individual test cases.

Also as that illustrates:
* this is straightforwardly a unit test of this code, not a [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)), so doesn't need the qualifier "smoke" in the name
* local variables are final or const if there's no intention to mutate them
* local variables don't need type annotations unless there's a reason type inference wouldn't give the same answer

test/model/avatar_url_test.dart
```diff
@@ -0,0 +1,147 @@
...
+    required this.email,
+  });
+}
+
+void main() {
+  var egUser = ExampleUser(
+    realm: Uri.parse("https://zulip.example.org"),
+    userId: 1417,
+    email: "sample@zulip.org",
+  );
```
gnprice, line 22: Better to use an actual `User` object. See `example_data.dart` for ways to conveniently make those for tests. See existing tests that import that file for examples to follow of the style for using it.

test/model/avatar_url_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+  double defaultSz = 30;
```
gnprice, line 24: "size", not "sz": https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-abbreviations

gnprice (commented):
Thanks @saisumith770 for the contribution! Comments below, and see also high-level comments above.

saisumith770:
> Sure, thanks for the clear questions.
> 
> When `User.avatarUrl` is null, we should handle that by computing a gravatar. That doesn't have to be part of this PR, though, as it's a separate issue we're tracking as #255.
> 
> There's another case where `resolvedUrl` is null, though: if `User.avatarUrl` is a string that doesn't parse as a URL. In that case we should continue to return `SizedBox.shrink()`, just like if the user wasn't found. I think the cleanest way to fix that will probably be to push the `tryResolveUrl` call inside `AvatarURL.fromUserOrBotData` (note there's a free-standing function `tryResolveUrl`, so you don't need to pass the store around in order to do that).
> 
> For the size, the AvatarImage widget hasn't previously called for a size because it hasn't previously had anything it would do with that information — it's been always fetchinig the same image regardless of what size it'll be shown at. Now that it does need to know the size, though, it should take the size as a parameter.
> 
> I don't think there's a good default that AvatarImage can use for the size — sometimes it'll be used in a place where it's going to be shown large, and sometimes where it's going to be shown small, and there's no reasonable way it can guess. So the parameter should just be required.

@gnprice You mentioned that size should be a required parameter for AvatarImage, but in the (recent_dm_conversations.dart)[https://github.com/zulip/zulip-flutter/blob/main/lib/widgets/recent_dm_conversations.dart] file AvatarImage is used without providing the size param. 
Untitled
This is in the previous version of the app. How should this be handled !

lib/model/avatar_url.dart
saisumith770: [c6715ea](https://github.com/zulip/zulip-flutter/pull/534/commits/c6715ea2f8091ddfa06162cc41e32035ea7f72a4)

lib/model/avatar_url.dart
saisumith770: [fa5e14b](https://github.com/zulip/zulip-flutter/pull/534/commits/fa5e14b9a1d5c49886f9bfaccb1210b9d10e56cd)

lib/model/avatar_url.dart
saisumith770: [d1cd89f](https://github.com/zulip/zulip-flutter/pull/534/commits/d1cd89f04490ac4053ab7ce995d83baad1656f3f)

test/model/avatar_url_test.dart
saisumith770: [ecec588](https://github.com/zulip/zulip-flutter/pull/534/commits/ecec58836637ac686e88dade6d241dbd31bcbefc)

test/model/avatar_url_test.dart
saisumith770: [ecec588](https://github.com/zulip/zulip-flutter/pull/534/commits/ecec58836637ac686e88dade6d241dbd31bcbefc)

lib/model/avatar_url.dart
saisumith770: I've removed the check for null and empty string. As per the previous suggestion, if the resolvedUrl is null the SizedBox.shrink() widget is returned instead of recomputing the url using AvatarUrl.
[ea85867](https://github.com/zulip/zulip-flutter/pull/534/commits/ea858676f5d012214f4dc3113ba2df9540491d80)

lib/model/avatar_url.dart
saisumith770: I was confused with the implementation in zulip-mobile where the get() method in UploadAvatarUrl would replace .png url with -medium.png. But a few avatar url's in zulip like the bots would not have a .png file extension, hence I thought this would be handled by the FallbackAvatarUrl.

Now I've written a way to handle both cases in the UploadAvatarUrl class itself, as to my understanding FallbackAvatarUrl should only be used when an avatarUrl is not sent by the rest api.
[869bd80](https://github.com/zulip/zulip-flutter/pull/534/commits/869bd8005ee10010db251e20083604358695d9c3)

gnprice:
At those call sites, the caller should pass the actual size the AvatarImage will have.

If you look around at the surrounding code, the caller does know the size. It just hasn't been passing it to AvatarImage because there hasn't previously been anything that AvatarImage would do with that information.

lib/model/avatar_url.dart
gnprice: Yeah, FallbackAvatarUrl is only for the case when the server doesn't tell us where the avatar is.

It looks like your new version still has special logic for `.jpg` and `.jpeg` extensions. That doesn't correspond to anything in the zulip-mobile code; the logic there around adding "-medium" is:
```js
      result.pathname = result.pathname.replace(
        // `.png` is optional: s3 uploads don't use it, local ones do.
        // See TODO in zulip/zulip@c03615b98.
        /(?:\.png)?$/,
        '-medium.png',
      );
```

For most things where we're interpreting data from the server or deciding what requests to make to the server, we follow the Zulip API docs as a specification for how to interact with the server. Here, as @chrisbobbe said in the #301 description:
> On CZO in `#api documentation`, I've [summarized](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/avatar_url.3A.20Request.20specific.20size/near/1636546) zulip-mobile's logic for determining a URL for any given size in physical pixels. I'm hoping the API doc can start specifying that logic, and we can use the new documentation for zulip-flutter's implementation.

So these avatar URLs are a part of the API that isn't yet documented, but we should do the next best thing to following the API docs:
* follow @chrisbobbe's summary in that chat thread, which is effectively a draft of what the API docs should say;
* follow the behavior of the zulip-mobile implementation, which encodes what we've previously learned about the server's behavior (and is what @chrisbobbe drew on for that chat thread).

saisumith770:
> At those call sites, the caller should pass the actual size the AvatarImage will have.
> 
> If you look around at the surrounding code, the caller does know the size. It just hasn't been passing it to AvatarImage because there hasn't previously been anything that AvatarImage would do with that information.

Refer commit [533c6ca](https://github.com/zulip/zulip-flutter/pull/534/commits/533c6ca97319106c59694538b94eec80fa5b2606)

test/model/avatar_url_test.dart
saisumith770: [df73a95](https://github.com/zulip/zulip-flutter/pull/534/commits/df73a950bda05a9d97388b757c2019a4c1d1e276)

lib/model/avatar_url.dart
saisumith770: [f4bd2a9](https://github.com/zulip/zulip-flutter/pull/534/commits/f4bd2a94630b52f45ada6ed0ef6202766efb31cb)

saisumith770:
@gnprice could you review the changes! Thank you for your efforts 🎉

lib/widgets/content.dart
```diff
@@ -967,9 +970,22 @@ class AvatarImage extends StatelessWidget {
...
       var avatarUrl => store.tryResolveUrl(avatarUrl),
     };
-    return (resolvedUrl == null)
-      ? const SizedBox.shrink()
-      : RealmContentNetworkImage(resolvedUrl, filterQuality: FilterQuality.medium, fit: BoxFit.cover);
+
+    if(resolvedUrl == null) {
+      return const SizedBox.shrink();
+    } else{
+      final avatarUrl = AvatarUrl.fromUserData(
```
gnprice, line 977: Instead of `else`, just let the normal non-null case come after the `if` statement entirely:

```suggestion
      return const SizedBox.shrink();
    }

    final avatarUrl = AvatarUrl.fromUserData(
```

The null case already returns, so this has the same effect.

This is called the "early return" pattern. It's helpful whenever you have a trivial case, or an error case, and then a normal or non-trivial case that's where the bulk of the interesting logic is. This way, the normal case with the interesting logic avoids getting indented deeper, and instead reads nicely straight down in the same column throughout the function. That's helpful for keeping the reader's attention on the main narrative of the function when that's what they want to be looking at, and for helping the reader systematically look at each of the error cases in turn when that's what they want to be looking at.

lib/widgets/content.dart
```diff
@@ -967,9 +970,22 @@ class AvatarImage extends StatelessWidget {
...
+        rawAvatarUrl: resolvedUrl.toString(),
```
gnprice, line 978: If we already have a parsed `Uri` object, we should be passing that instead of a string.

That will also let us simplify some of the downstream logic within AvatarUrl and its subclasses. This `Uri` object is always an absolute URL (because `tryResolveUrl` always returns an absolute URL), so we don't need any of the logic specific to relative URLs.

test/widgets/content_test.dart
```diff
@@ -471,6 +471,20 @@ void main() {
...
+      const avatarUrl = 'https://example/avatar.png';
+      check(await actualUrl(tester, avatarUrl, 50)).isNotNull()
+        .asString.equals(avatarUrl.replaceAll('.png', '-medium.png'));
+      debugNetworkImageHttpClientProvider = null;
+    });
+
+    testWidgets('smoke with relative URL and size(pixelratio * 50)', (tester) async {
+      const avatarUrl = '/avatar.png';
+      check(await actualUrl(tester, avatarUrl, 50))
+        .equals(store.tryResolveUrl(avatarUrl.replaceAll('.png', '-medium.png'))!);
```
gnprice, line 484: This is clearer if you just spell out the result:
```suggestion
        .equals(store.tryResolveUrl('/avatar-medium.png')!);
```

The `replaceAll` version is already relying on the details of how this particular `avatarUrl` looks — for example if the URL were `/avatars.pngset/avatar.png`, then this `replaceAll` would rewrite it in two places and get the wrong result.

test/widgets/content_test.dart
```diff
@@ -471,6 +471,20 @@ void main() {
       debugNetworkImageHttpClientProvider = null;
     });
 
+   testWidgets('smoke with absolute URL and size(pixelratio * 50)', (tester) async {
+      const avatarUrl = 'https://example/avatar.png';
+      check(await actualUrl(tester, avatarUrl, 50)).isNotNull()
+        .asString.equals(avatarUrl.replaceAll('.png', '-medium.png'));
```
gnprice, line 477: The fact that this produces the "medium" URL is dependent on (as your description string suggests) the size in physical pixels, and therefore on what the device pixel ratio is.

There is some default value that tests use for the device pixel ratio, but I don't know offhand what it is. Neither will other likely readers of this test case.

In general, a test case should be self-contained in expressing the key pieces of data that are essential to the test behaving correctly and testing the thing it's meant to test. So this test should explicitly set the device pixel ratio.

Somewhere Flutter will provide a way to do that in tests — might be a method on `tester.binding`, or might be a global with a name starting with "debug", like the `debugNetworkImageHttpClientProvider` you see in the existing tests here. So you should find that and use it. If after looking around the docs you're not finding it, ask in `#mobile-team` and mention where you've looked so far, and we can help look.

test/model/avatar_url_test.dart
```diff
@@ -0,0 +1,54 @@
...
+    test('gravatar url', () {
+      String url = '${GravatarUrl.origin}/avatar/1234';
+      final avatarUrl = AvatarUrl.fromUserData(
+        rawAvatarUrl: url,
+        realm: account.realmUrl,
+        userId: user.userId,
+        email: user.email,
+      );
+
+      check(avatarUrl.get(defaultSize).toString()).equals('$url?s=30.0');
```
gnprice, line 24: Hmm, this ".0" seems undesired. Is that what's seen in the zulip-mobile tests?

test/model/avatar_url_test.dart
```diff
@@ -0,0 +1,54 @@
...
+void main() {
+  final user = eg.user();
+  final account = eg.account(user: user);
+  
+  double defaultSize = 30;
+  double largeSize = 120;
+
+  group('GravatarUrl', () {
+    test('gravatar url', () {
+      String url = '${GravatarUrl.origin}/avatar/1234';
```
gnprice, line 16: These numbers and this string are more examples of these points from https://github.com/zulip/zulip-flutter/pull/534#discussion_r1503273346 :
> * local variables are final or const if there's no intention to mutate them
> * local variables don't need type annotations unless there's a reason type inference wouldn't give the same answer

In general, when you get a piece of code-review feedback in one place, it's important to look for other places in your PR (or other open and future PRs) where the same comment applies. That's a key ingredient in making it easy for maintainers to review and merge your PRs.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+/// Pixel dimensions of different size choices we have (they're all
+/// square) when requesting an uploaded avatar.
+/// DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
+const int defaultUploadSizePx = 100;
+
+/// MEDIUM_AVATAR_SIZE in zerver/lib/upload.py.
+const int mediumUploadSizePx = 500;
+
+abstract class AvatarUrl {
+  static AvatarUrl fromUserData({
```
gnprice, line 16: Let's make this a factory constructor too.

In fact it can be named as just the default constructor:
```suggestion
  factory AvatarUrl({
```

(That way will require a couple of small fixes for the sake of the subclasses' constructors. If you're having trouble working those out, you can make it `factory AvatarUrl.fromUserData`.)

gnprice (commented):
Thanks for the revision! Here's a fresh round of review.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+  Uri get(double sizePhysicalPx) {
+    Uri result = standardUri.replace(queryParameters: {
+      ...standardUri.queryParameters,
+      's': sizePhysicalPx.toString(),
+    });
+    return result;
+  }
+}
+
+class FallbackAvatarUrl extends AvatarUrl {
```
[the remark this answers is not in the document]
saisumith770: [b3a542c](https://github.com/zulip/zulip-flutter/pull/534/commits/b3a542c80b33f702c4a313a80c1175fc2be5f5e2)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+    // }
+
+    // TODO(#254): handle computing Fallback avatar
+    // if (rawAvatarUrl == undefined) {
+    //   return FallbackAvatarUrl(
+    //       realm: realm, userId: userId);
+    // }
+
+
+    if (rawAvatarUrl.startsWith(GravatarUrl.origin)) {
```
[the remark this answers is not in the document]
saisumith770: [b3a542c](https://github.com/zulip/zulip-flutter/pull/534/commits/b3a542c80b33f702c4a313a80c1175fc2be5f5e2)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+
+import '../log.dart';
+
+/// Pixel dimensions of different size choices we have (they're all
+/// square) when requesting an uploaded avatar.
+/// DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
+const int defaultUploadSizePx = 100;
+
+/// MEDIUM_AVATAR_SIZE in zerver/lib/upload.py.
+const int mediumUploadSizePx = 500;
```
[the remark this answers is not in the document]
saisumith770: [d6c8073](https://github.com/zulip/zulip-flutter/pull/534/commits/d6c8073109cdd127ec818767ec837c71b6b61d56)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+  factory GravatarUrl({
+    required String email,
+    String? urlFromServer,
+  }) {
+    if (urlFromServer != null) {
+      return GravatarUrl._(standardUri: Uri.parse(urlFromServer));
+    } else {
+      return GravatarUrl._(
+          standardUri: Uri.parse(
+              '${GravatarUrl.origin}/avatar/${md5.convert(utf8.encode(email.toLowerCase()))}?d=identicon'));
```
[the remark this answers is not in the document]
saisumith770: [0732587](https://github.com/zulip/zulip-flutter/pull/534/commits/0732587e579262337a8c5bb3200e46d0898f551c)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+class GravatarUrl implements AvatarUrl {
+  factory GravatarUrl({
+    required String email,
+    String? urlFromServer,
+  }) {
+    if (urlFromServer != null) {
+      return GravatarUrl._(standardUri: Uri.parse(urlFromServer));
+    } else {
+      return GravatarUrl._(
+          standardUri: Uri.parse(
```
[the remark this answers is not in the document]
saisumith770: [0732587](https://github.com/zulip/zulip-flutter/pull/534/commits/0732587e579262337a8c5bb3200e46d0898f551c)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+      result = Uri.parse('$standardUri/medium');
+    }
+
+    return result;
+  }
+}
+
+class UploadedAvatarUrl extends AvatarUrl {
+  factory UploadedAvatarUrl({
+    required Uri realm,
```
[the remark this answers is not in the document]
saisumith770: [2053a09](https://github.com/zulip/zulip-flutter/pull/534/commits/2053a09306473df8d0ac7337cf2a4aec9d8f8a66)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+
+  @override
+  Uri get(double sizePhysicalPx) {
+    Uri result = Uri.parse(standardUri.toString());
+
+    if (sizePhysicalPx > defaultUploadSizePx) {
+      result = result.replace(path: result.path.replaceFirst(RegExp(r'(?:\.png)?$'),"-medium.png"));
+    }
+
+    return result;
```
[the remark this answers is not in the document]
saisumith770: [26bf305](https://github.com/zulip/zulip-flutter/pull/534/commits/26bf305cb7a9ae95f2f85dcf76f453089744c351)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+  Uri standardUri;
+
+  UploadedAvatarUrl._({required this.standardUri});
+
+  @override
+  Uri get(double sizePhysicalPx) {
+    Uri result = Uri.parse(standardUri.toString());
+
+    if (sizePhysicalPx > defaultUploadSizePx) {
+      result = result.replace(path: result.path.replaceFirst(RegExp(r'(?:\.png)?$'),"-medium.png"));
```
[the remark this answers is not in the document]
saisumith770: [26bf305](https://github.com/zulip/zulip-flutter/pull/534/commits/26bf305cb7a9ae95f2f85dcf76f453089744c351)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+    return result;
+  }
+}
+
+class UploadedAvatarUrl extends AvatarUrl {
+  factory UploadedAvatarUrl({
+    required Uri realm,
+    required String absoluteOrRelativeUrl,
+  }) {
+    String? absoluteUrl;
```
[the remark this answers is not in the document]
saisumith770: [26bf305](https://github.com/zulip/zulip-flutter/pull/534/commits/26bf305cb7a9ae95f2f85dcf76f453089744c351)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+          standardUri: Uri.parse(
+              '${GravatarUrl.origin}/avatar/${md5.convert(utf8.encode(email.toLowerCase()))}?d=identicon'));
+    }
+  }
+
+  static String origin = 'https://secure.gravatar.com';
+
+  Uri standardUri;
+
+  GravatarUrl._({required this.standardUri});
```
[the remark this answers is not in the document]
saisumith770: [da5d662](https://github.com/zulip/zulip-flutter/pull/534/commits/da5d66280f0d5eb0225a6318cb33f03930c0914c)

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,141 @@
...
+    //       realm: realm, userId: userId);
+    // }
+
+
+    if (rawAvatarUrl.startsWith(GravatarUrl.origin)) {
+      return GravatarUrl(
+          urlFromServer: rawAvatarUrl, email: email);
+    } else {
+      return UploadedAvatarUrl(
+          realm: realm, absoluteOrRelativeUrl: rawAvatarUrl);
```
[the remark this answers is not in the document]
saisumith770: [92e8870](https://github.com/zulip/zulip-flutter/pull/534/commits/92e8870907df5a6f6ac61c9a5c96413090c80e4c)

lib/widgets/recent_dm_conversations.dart
```diff
@@ -90,24 +90,24 @@ class RecentDmConversationsItem extends StatelessWidget {
     final selfUser = store.users[store.selfUserId]!;
 
     final String title;
-    final Widget avatar;
+    final Widget Function({required double size}) generateAvatarImage;
     switch (narrow.otherRecipientIds) { // TODO dedupe with DM items in [InboxPage]
       case []:
         title = selfUser.fullName;
-        avatar = AvatarImage(userId: selfUser.userId);
+        generateAvatarImage = ({required double size}) => AvatarImage(userId: selfUser.userId, size: size);
```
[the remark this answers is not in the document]
saisumith770: [aeb11d4](https://github.com/zulip/zulip-flutter/pull/534/commits/aeb11d4ffcfe04407265c832395821165d545351)

lib/widgets/content.dart
```diff
@@ -937,7 +938,7 @@ class Avatar extends StatelessWidget {
     return AvatarShape(
       size: size,
       borderRadius: borderRadius,
-      child: AvatarImage(userId: userId));
+      child: AvatarImage(userId: userId,size: size));
```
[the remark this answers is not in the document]
saisumith770: [2033a0d](https://github.com/zulip/zulip-flutter/pull/534/commits/2033a0d1faa2fd6d32f29380783819a4fc5740f5)

lib/widgets/content.dart
```diff
@@ -967,9 +970,22 @@ class AvatarImage extends StatelessWidget {
       null          => null, // TODO(#255): handle computing gravatars
       var avatarUrl => store.tryResolveUrl(avatarUrl),
     };
-    return (resolvedUrl == null)
-      ? const SizedBox.shrink()
-      : RealmContentNetworkImage(resolvedUrl, filterQuality: FilterQuality.medium, fit: BoxFit.cover);
+
+    if(resolvedUrl == null) {
```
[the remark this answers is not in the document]
saisumith770: [2033a0d](https://github.com/zulip/zulip-flutter/pull/534/commits/2033a0d1faa2fd6d32f29380783819a4fc5740f5)

test/model/avatar_url_test.dart
gnprice: OK, seems like the zulip-mobile tests don't check this question — but in that case, what's the actual behavior in zulip-mobile? Does it ever put something like "30.0" into the URL, or is it always an integer?

---

I just took a look, and it looks like all the call sites of AvatarURL.get are like this one (in `src/common/UserAvatar.js`):
```js
          uri: avatarUrl.get(PixelRatio.getPixelSizeForLayoutSize(size)).toString(),
```
in that they use `PixelRatio.getPixelSizeForLayoutSize` to get the value to pass as `sizePhysicalPx`.

And if you click through in your IDE to the definition of that method (it's part of React Native), its jsdoc says:
```js
  /**
   * Converts a layout size (dp) to pixel size (px).
   *
   * Guaranteed to return an integer number.
   */
```

So it always returns an integer, which means we always pass an integer to the `get` method, which means we always say something like `s=30` here, not `s=30.0`.

Thinking about the semantics of this `get` method and its parameter, we're talking about the size of a desired image. The image is a raster (typically a PNG, perhaps always so), which means it's represented as a grid of pixels. So the image's actual size in each dimension, as a raster, is an integer — there might be 30 rows of pixels in the grid, or 31, but there can't be 30.1 rows of pixels. So: the parameter type for `AvatarUrl.get` should be `int`, not `double`.

That distinction doesn't exist in JavaScript — there's just numbers, meaning floating-point numbers — which is why it's `number` in zulip-mobile.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,138 @@
+import '../log.dart';
+
+/// Pixel dimensions of different size choices we have (they're all
+/// square) when requesting an uploaded avatar.
+/// DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
+const int defaultUploadSizePx = 100;
```
gnprice, line 6: This text doesn't make sense when applied to just one of these definitions. It says  "different size choices we have", but this is just one choice.

Here's a revised version:
```dart
/// The size threshold above which is "medium" size for an avatar.
///
/// This is DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
const int defaultUploadSizePx = 100;
```

And then we don't actually seem to use `mediumUploadSizePx` at all, so can just drop that one.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  Uri get(double sizePhysicalPx);
+}
+
+class GravatarUrl implements AvatarUrl {
+  GravatarUrl._({required this.standardUri});
+
+  factory GravatarUrl({
```
gnprice, line 36: nit: put this private constructor after the public one

(this public one is also the "default" constructor, i.e. the one with no name beyond the class's name — so that's the constructor the Flutter style guide I linked previously says to put first)

gnprice (commented):
I started doing another round of review, but then I realized you hadn't actually said the PR was ready for another round. And indeed there are a number of comments you haven't yet gotten to — which is fine, please take as much time as you need for addressing the review feedback.

Just reply here to say when it's ready for another round. I'm also happy to answer specific questions in between, like at https://github.com/zulip/zulip-flutter/pull/534#discussion_r1507703894 above.

Anyway, here are the couple of items I spotted as I was just beginning a review.

test/model/avatar_url_test.dart
saisumith770: [8e9cf2e](https://github.com/zulip/zulip-flutter/pull/534/commits/8e9cf2ebfb70130ec8c00136d758e2dfdcb4677d)

lib/model/avatar_url.dart
saisumith770: [ef7582d](https://github.com/zulip/zulip-flutter/pull/534/commits/ef7582dc24572904bd186034795977cec0e0886a)

lib/widgets/content.dart
saisumith770: https://github.com/zulip/zulip-flutter/pull/534/commits/4c24c3e6c25f12b792f947839baa8e85c5305e91

lib/model/avatar_url.dart
saisumith770: https://github.com/zulip/zulip-flutter/pull/534/commits/3230320ec22a46485e6d1c8c7829f3c5207aef5e

lib/model/avatar_url.dart
saisumith770: https://github.com/zulip/zulip-flutter/pull/534/commits/3230320ec22a46485e6d1c8c7829f3c5207aef5e

test/model/avatar_url_test.dart
saisumith770: [7a0b600](https://github.com/zulip/zulip-flutter/pull/534/commits/7a0b60054e634aa5120ee821b80fe8a554ed0a77)

test/widgets/content_test.dart
saisumith770: [f74e3d4](https://github.com/zulip/zulip-flutter/pull/534/commits/f74e3d4e53efee364cad4f2180351fbf72d01b02)

saisumith770:
@gnprice I've made the suggested changes, do let me know if any revisions are required. I think the CI is currently stuck, is it possible for you to manually restart it!

gnprice (commented):
Thanks for the revision! This is getting a lot closer.

There's a variety of comments below. One other important thing we'll need before we can merge this is that the commits in the branch need to be squashed and reorganized into clear and coherent commits, following the Zulip project's Git style: https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html

For this PR, I think that will mean two commits:
* One commit adding `size` to `AvatarImage`, and making the necessary changes where that widget is used. This commit should be [an NFC commit](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc).
* Then one commit with the rest of the changes: adding `AvatarUrl` and subclasses, using that from `AvatarImage`, adding the tests.

To get there from your current structure, my suggestion would be:
* First, `git rebase` to get on top of current main and eliminate the merge commit. (If this hits any merge conflicts, the easiest solution might be to abort the rebase and use `git merge origin/main` instead. You'll still have conflicts but all at once, which in this case is probably easiest.)
* Then `git reset origin/main`, so that your current commit is `origin/main` but you have all your changes to files.
* Then `git add -p` followed by `git commit`, to make the first commit.
* Then `git add` and `git commit` to make the second commit.

But if you have another procedure you're comfortable with that gets the right result, that's good too. If you have any questions with the Git manipulations or want any help, just ask in `#git help` on chat.zulip.org.

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,69 @@
...
+    required Uri urlFromServer,
+  }) {
+    return GravatarUrl._(standardUri: urlFromServer);
+  }
+
+  GravatarUrl._({required this.standardUri});
+
+  static String origin = 'https://secure.gravatar.com';
+
+  Uri standardUri;
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/d425673f04c81a15b996c65538579aee1453204a

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,69 @@
...
+}
+
+class GravatarUrl implements AvatarUrl {
+  factory GravatarUrl({
+    required Uri urlFromServer,
+  }) {
+    return GravatarUrl._(standardUri: urlFromServer);
+  }
+
+  GravatarUrl._({required this.standardUri});
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/062006fa7544a1513dc6795b97e658c76f7cf173

test/widgets/content_test.dart
```diff
@@ -471,6 +471,34 @@ void main() {
...
+      // size is (2.5 * 50) 125px that exceeds the defaultUploadSizePx
+      tester.view.devicePixelRatio = 2.5;
+
+      const avatarUrl = 'https://example/avatar.png';
+      check(await actualUrl(tester, avatarUrl, 50)).isNotNull()
+        .asString.equals(avatarUrl.replaceAll('.png', '-medium.png'));
+      debugNetworkImageHttpClientProvider = null;
+
+      // reset the device pixel ratio after test completion
+      addTearDown(tester.view.resetDevicePixelRatio);
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/81d79516d00c25ff9e97d19354a632dd11f09d34

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,69 @@
...
+    return UploadedAvatarUrl._(standardUri: absoluteUrl);
+  }
+
+  UploadedAvatarUrl._({required this.standardUri});
+
+  Uri standardUri;
+
+  @override
+  Uri get(int sizePhysicalPx) {
+    final result = Uri.parse(standardUri.toString());
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/b0c2b2cac25671d0776e4384eb858051a1c6f0a7

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  Uri standardUri;
+
+  @override
+  Uri get(int sizePhysicalPx) {
+    final result = standardUri.replace(queryParameters: {
+      ...standardUri.queryParameters,
+      's': sizePhysicalPx.toString(),
+    });
+    return result;
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/b0c2b2cac25671d0776e4384eb858051a1c6f0a7

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,69 @@
...
+///
+/// This is DEFAULT_AVATAR_SIZE in zerver/lib/upload.py.
+const int defaultUploadSizePx = 100;
+
+abstract class AvatarUrl {
+  factory AvatarUrl.fromUserData({
+    required Uri rawAvatarUrl,
+  }) {
+    // TODO(#255): handle computing gravatar's
+    // TODO(#254): handle computing Fallback avatar
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/c4cdbff0884c9993e1ed8b71e53887eaf09902e8

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,69 @@
...
+    if (rawAvatarUrl.toString().startsWith(GravatarUrl.origin)) {
+      return GravatarUrl(
+        urlFromServer: rawAvatarUrl);
+    } else {
+      return UploadedAvatarUrl(
+        absoluteUrl: rawAvatarUrl);
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/b2112db2d9be92856d357b302814cda0c10788e8

lib/widgets/content.dart
```diff
@@ -967,9 +970,21 @@ class AvatarImage extends StatelessWidget {
...
+    if (resolvedUrl == null) {
+      return const SizedBox.shrink();
+    }
+
+    final avatarUrl = AvatarUrl.fromUserData(
+      rawAvatarUrl: resolvedUrl,
+    );
+    return RealmContentNetworkImage(
+      avatarUrl.get(
+        (MediaQuery.of(context).devicePixelRatio * size).toInt(),
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/655bbd301ec984cc9e0b5dee017933e1551d5797

lib/widgets/content.dart
```diff
@@ -967,9 +970,21 @@ class AvatarImage extends StatelessWidget {
...
+      ),
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/655bbd301ec984cc9e0b5dee017933e1551d5797

test/widgets/content_test.dart
```diff
@@ -471,6 +471,34 @@ void main() {
       debugNetworkImageHttpClientProvider = null;
     });
 
+   testWidgets('smoke with absolute URL and size(pixelratio(2.5) * 50) 125px', (tester) async {
```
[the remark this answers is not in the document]
saisumith770: https://github.com/zulip/zulip-flutter/commit/fba0ddf39624df9fa292eee546754360bc5ee57e

saisumith770:
@gnprice I've made the changes, if it's alright then I will start squashing the commits as requested

gnprice:
Sure — please go ahead.

saisumith770:
> Thanks for the revision! This is getting a lot closer.
> 
> There's a variety of comments below. One other important thing we'll need before we can merge this is that the commits in the branch need to be squashed and reorganized into clear and coherent commits, following the Zulip project's Git style: https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html
> 
> For this PR, I think that will mean two commits:
> 
> * One commit adding `size` to `AvatarImage`, and making the necessary changes where that widget is used. This commit should be [an NFC commit](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc).
> * Then one commit with the rest of the changes: adding `AvatarUrl` and subclasses, using that from `AvatarImage`, adding the tests.
> 
> To get there from your current structure, my suggestion would be:
> 
> * First, `git rebase` to get on top of current main and eliminate the merge commit. (If this hits any merge conflicts, the easiest solution might be to abort the rebase and use `git merge origin/main` instead. You'll still have conflicts but all at once, which in this case is probably easiest.)
> * Then `git reset origin/main`, so that your current commit is `origin/main` but you have all your changes to files.
> * Then `git add -p` followed by `git commit`, to make the first commit.
> * Then `git add` and `git commit` to make the second commit.
> 
> But if you have another procedure you're comfortable with that gets the right result, that's good too. If you have any questions with the Git manipulations or want any help, just ask in `#git help` on chat.zulip.org.

you mentioned there should be a separate NFC commit for the AvatarImage size property, but I would need the size property to compute the physicalSize in content.dart, also in the content_test.dart the size prop is required for the tests. Hence I'm currently adding all the changes in a single commit. Let me know if there is some other workaround!

lib/model/avatar_url.dart
```diff
@@ -0,0 +1,50 @@
...
+  @override
+  Uri get(int sizePhysicalPx) {
+    if (sizePhysicalPx > defaultUploadSizePx) {
+      return standardUrl.replace(
+        path: standardUrl.path.replaceFirst(RegExp(r'(?:\.png)?$'), "-medium.png"));
+    }
+
+    return standardUrl;
+  }
+}
```
gnprice, line 50: This final line is incomplete — it lacks the terminating newline.

That's expressed by the red ⊖ icon that GitHub shows here. You can also see it in `git log -p`, which expresses that fact like this:
```
+    return standardUrl;
+  }
+}
\ No newline at end of file
```

gnprice (commented):
Thanks @saisumith770 for the revision! This is close to merging now.

A few comments below on the code. The commit structure all looks good.

The commit messages are close but need a few changes. I can make a final pass on those before merging, but please also try to give them another round of editing yourself. Take a look again in particular at this section of the Zulip commit style guide:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-summary-part-2

For the main commit (the second one, the one making most of the changes), see also the paragraph about GitHub issues in this section:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-description
and see this section about what information goes in the description:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#the-purpose-of-the-commit-description

Finally, looking back at the screenshots in the description: please include "before" versions as well as "after". That helps make it easy to see from the screenshots what's changed in the UI.

lib/widgets/content.dart
```diff
@@ -1057,9 +1060,19 @@ class AvatarImage extends StatelessWidget {
...
-    return (resolvedUrl == null)
-      ? const SizedBox.shrink()
-      : RealmContentNetworkImage(resolvedUrl, filterQuality: FilterQuality.medium, fit: BoxFit.cover);
+
+    if (resolvedUrl == null) {
+      return const SizedBox.shrink();
+    }
+
+    final avatarUrl = AvatarUrl.fromUserData(resolvedUrl: resolvedUrl);
+    final physicalSize = (MediaQuery.of(context).devicePixelRatio * size).ceil();
```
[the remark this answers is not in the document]
saisumith770: [77de482](https://github.com/zulip/zulip-flutter/pull/534/commits/77de482d797a995c2eabdd8eff5d12e6f1b1d2c0)

lib/widgets/recent_dm_conversations.dart
```diff
@@ -81,6 +81,8 @@ class RecentDmConversationsItem extends StatelessWidget {
     required this.unreadCount,
   });
 
+  static const double _avatarSize = 32;
+
   final DmNarrow narrow;
   final int unreadCount;
```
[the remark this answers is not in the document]
saisumith770: [5d95c2a](https://github.com/zulip/zulip-flutter/pull/534/commits/5d95c2ac087d582fca4a56270a86bc87883e4dcb)

test/widgets/content_test.dart
```diff
@@ -571,6 +571,32 @@ void main() {
       debugNetworkImageHttpClientProvider = null;
     });
 
+   testWidgets('absolute URL, larger size', (tester) async {
+      // setting the device pixel ratio to 2.5, effective image
+      // size is (2.5 * 50) 125px that exceeds the defaultUploadSizePx
+      tester.view.devicePixelRatio = 2.5;
```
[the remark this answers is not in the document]
saisumith770: [4f356e8](https://github.com/zulip/zulip-flutter/pull/534/commits/4f356e8831e885e6ffacc45aea177ae0a5d17f3d)

test/widgets/content_test.dart
```diff
@@ -571,6 +571,32 @@ void main() {
...
+      // reset the device pixel ratio after test completion
+      addTearDown(tester.view.resetDevicePixelRatio);
```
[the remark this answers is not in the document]
saisumith770: [4f356e8](https://github.com/zulip/zulip-flutter/pull/534/commits/4f356e8831e885e6ffacc45aea177ae0a5d17f3d)

lib/model/avatar_url.dart
saisumith770: https://github.com/zulip/zulip-flutter/pull/534/commits/11833f1fb12bed32481a9ba558b573999c80a69c

lib/widgets/content.dart
```diff
@@ -1057,9 +1060,19 @@ class AvatarImage extends StatelessWidget {
...
-    return (resolvedUrl == null)
-      ? const SizedBox.shrink()
-      : RealmContentNetworkImage(resolvedUrl, filterQuality: FilterQuality.medium, fit: BoxFit.cover);
+
+    if (resolvedUrl == null) {
+      return const SizedBox.shrink();
+    }
+
+    final avatarUrl = AvatarUrl.fromUserData(resolvedUrl: resolvedUrl);
+    final physicalSize = ((MediaQuery.maybeDevicePixelRatioOf(context) ?? 1) * size).ceil();
```
gnprice, line 1069: Hmm, interesting. Where does 1 come from as a default?

I think if we really had to use a default here, it'd be better to be more toward the high end of likely values. That'd mean we get an image that has enough pixels it shouldn't look blurry.

But I don't know a reason why we should need a default for this — there should always be a `MediaQuery` ancestor. And the previous revision didn't have a default. So instead of `?? 1`, that would be spelled
```dart
MediaQuery.maybeDevicePixelRatioOf(context)!
```
or more concisely
```dart
MediaQuery.devicePixelRatioOf(context)
```
.

gnprice (commented):
Thanks for the revision! This is very close now to merge.

Just one comment on the revised code, below.

The commit messages also look good, except that the main commit's message should say it's closing the issue. (Hmm — and the PR description, at the top of this thread, should do so too.) It could also benefit from just a few more words saying what it's doing. So here's a revised version:

```
avatar: Fix blurriness on profile screen by adjusting physical image size

Fixes: #301 
```

Please also add "before" as well as "after" screenshots as mentioned above. (Rereading, I see you didn't actually say this was ready for re-review, so perhaps you were already planning to do that.)

saisumith770:
@gnprice I've made the suggested changes!

gnprice:
Looks good — merging. Thanks @saisumith770 for all your work on this!

Issue #538, opened 2024-02-29
content: Inline code uses paragraph font size, even in headings (h1, etc.)


chrisbobbe:
![image](https://github.com/zulip/zulip-flutter/assets/22248748/379e852c-4972-44e5-98d1-d5e4c6b555e7)

chrisbobbe:
Oh, this is also true of user mentions and timezone-aware times:

![image](https://github.com/zulip/zulip-flutter/assets/22248748/3b81339b-4075-4868-8f04-c00d6b52da12)

(Also, for those elements, they aren't getting bolded as they should.)

gnprice:
Oops, #541 had a classic GitHub accidental-close phrase :slightly_smiling_face:: "doesn't fix #538."

chrisbobbe:
Agh! Thanks for the catch!

Issue #512, opened 2024-02-09
Increase content font size


gnprice:
This is a sub-issue of:
* #157

Currently the app's font size for plain text in message bodies is 14px. That dates from the early prototype phase, when I was copying web's design meticulously (to demonstrate that we could).

The corresponding size in the zulip-mobile legacy RN app is 16px, I believe.

The corresponding size in @terpimost's design in Figma is 17px. For example, see [here](https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=538-20849&mode=design&t=8UJkvvqjxtix3FQO-0).

So we should increase the size, either to 16px or 17px.

If we're unsure which of those sizes to prefer, we could add a setting (even if only for the beta period) — that would make it easy for beta users to experiment with the different values and take crisp apples-to-apples pairs of screenshots for discussion.

Chat thread: https://chat.zulip.org/#narrow/stream/48-mobile/topic/density.20in.20message.20list/near/1733351

vijay-kesanakurthi:
I want to work on this, to get a better understanding of the project for GSOC.

gnprice:
@vijay-kesanakurthi Ah, we should have marked this sooner — this is actually an issue that a member of the core team is planning to take care of within the next week. I'll mark it as assigned now.

There are plenty of other issues [on our project board](https://github.com/orgs/zulip/projects/5/views/4), though, that don't have anyone assigned, so I'd encourage browsing through those. And if you haven't already, please take a look at the Zulip project's guide for picking an issue to work on: https://zulip.readthedocs.io/en/latest/contributing/contributing.html#picking-an-issue-to-work-on

PR #540, opened 2024-02-29
content: Use 17 for base font size, and fix code/mentions/times in headings

chrisbobbe:
| Before | After |
| --- | --- |
| image | image | 
| image | image |
| image | image |
| image | image |

Fixes: #512
Fixes: #538

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -19,7 +19,7 @@ import 'store.dart';
 import 'text.dart';
 
 /// The font size for message content in a plain unstyled paragraph.
-const double kBaseFontSize = 14;
+const double kBaseFontSize = 17;
 
 /// The entire content of a message, aka its body.
 ///
@@ -112,7 +112,7 @@ class Paragraph extends StatelessWidget {
 
   static const textStyle = TextStyle(
     fontSize: kBaseFontSize,
-    height: (17 / kBaseFontSize),
+    height: (22 / kBaseFontSize),
   );
 
   @override
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -477,7 +477,7 @@ class MathBlock extends StatelessWidget {
 //
 
 Widget _buildBlockInlineContainer({
-  required TextStyle? style,
+  required TextStyle style,
   required BlockInlineContainerNode node,
 }) {
   if (node.links == null) {
@@ -493,7 +493,7 @@ class _BlockInlineContainer extends StatefulWidget {
     {required this.links, required this.style, required this.nodes});
 
   final List<LinkNode> links;
-  final TextStyle? style;
+  final TextStyle style;
   final List<InlineContentNode> nodes;
 
   @override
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks for cleaning these up! The code changes all look good to me. A couple of tests are failing, though.

chrisbobbe:
Those tests are helpful indeed! Hmm. So the problem is that we're trying to apply the 0.825 `fontSize` multiplier for inline code in places that don't actually have a `fontSize` value to apply the multiplier to, e.g. (by the end of the branch), 17 for paragraphs and 23.8 for h1 headings. That will happen when the paragraph/h1/etc. isn't the thing directly surrounding the code span, such as when there's a strong (bold) or emphasis (italic) span around the code span.

For example, this content fails to render:

[hello ***`world`***](https://chat.zulip.org/)

In a CZO message: https://chat.zulip.org/#narrow/stream/7-test-here/topic/Chris/near/1748844

The error is:

```
The following assertion was thrown building InlineContent(dirty, dependencies: [MediaQuery]):
'package:flutter/src/painting/text_style.dart': Failed assertion: line 990 pos 12: 'fontSize != null || (fontSizeFactor == 1.0 && fontSizeDelta == 0.0)': is not true.
```

The code span is expecting to find a font size on the `TextStyle` of the surrounding span. That span is an emphasis (italic) span with just `const TextStyle(fontStyle: FontStyle.italic))`, so no `fontSize`.

chrisbobbe:
To fix, we could maybe:

- For every span that might sit between a code span and a paragraph/h1/etc., make sure the span merges into its own style the font size of the span that surrounds it. This seems error-prone.
- In paragraphs/h1s/etc., use a `DefaultTextStyle` that provides the appropriate font size, and then the code-span code can consume it using `DefaultTextStyle.of`.

gnprice:
I like the second direction, and I think we can simplify it: `DefaultTextStyle` is a way of specifying a style as an inherited widget, to be found from throughout that widget's subtree. But paragraphs and headings etc. are leaves in the widget tree — the further recursion that happens in the ContentNode tree is all producing spans for the one widget, not additional descendant widgets.

So instead of `DefaultTextStyle`, I think we can pass a style to _InlineContentBuilder as the default style, and then the code-span code can consume that and do the same thing it would do with the result of `DefaultTextStyle.of`.

chrisbobbe:
Interesting; thank you!

chrisbobbe:
> So instead of `DefaultTextStyle`, I think we can pass a style to _InlineContentBuilder as the default style

Doesn't `_InlineContentBuilder` already have this, kind of, and call it `widget.style`? (Except that `InlineContent.style` is optional, but could easily be made required?)

gnprice:
Yeah, that seems right. So then that approach comes down to taking `widget.style` and drawing from it in more places, in the same way you might have done from `DefaultTextStyle.of`.

chrisbobbe:
Since #541, we're now using 17 for the base font size.

I plan to close this and open a new PR for aimed at #538, with fresh screenshots (that will show the font-size 17 in the "before" column).

PR #541, opened 2024-02-29
content: Use 17 for base font size (and height 22), like in Figma

### LINKED ISSUE #538: content: Inline code uses paragraph font size, even in headings (h1, etc.)

chrisbobbe:
This is an alternative to #540 that doesn't fix #538. Screenshots coming soon.

Fixes: #512


gnprice:
Thanks! Looks good; merging.

Issue #513, opened 2024-02-09
Excess spacing between sender and paragraph


gnprice:
From a thread on white space vs. density in the message list:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/density.20in.20message.20list/near/1735032

We currently have 8px of vertical space between the bottom of a sender avatar and the top of a paragraph, if that's the start of the message. The design shows 4px.

Specifically we have 4px between the avatar and the message content, and then our `Paragraph` widget has 4px top and bottom padding around the underlying `Text`.

I think fixing this properly will basically involve taking care of #162 — going through and properly allocating all the vertical padding between messages, between sender and body, and around and between paragraphs, list items, quotes, images, and all the rest. That's one aspect of copying web's design that I skipped in the early prototype, and we haven't yet returned to (which is why it's an open issue).

But e.g. one easy hack would be to cut the 4px below the sender, and count on the content's first child to have its own top padding. That might be worth doing, pending #162. (Which is a task we're holding off on for the moment because there's some work happening now/soon on web that should have a side effect of making the task easier for us; details there.)

PR #542, opened 2024-03-01
msglist: Cut excess spacing between sender and paragraph

chrisbobbe:
This is the "easy hack" Greg mentions in the description of #513.

Fixes: #513

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -941,7 +941,7 @@ class MessageWithPossibleSender extends StatelessWidget {
         padding: const EdgeInsets.symmetric(vertical: 4),
         child: Column(children: [
           if (senderRow != null)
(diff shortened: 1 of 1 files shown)


chrisbobbe:
| Before | After |
| --- | --- |
| image | image |

chrisbobbe:
(I didn't mark this with `Fixes:` because it doesn't do the "fixing this properly" thing that you described in the issue.)

gnprice:
Thanks!

I think this can actually be best described as fixing #513. This does fix the excess spacing, just in a way that potentially goes too far for certain content. And the work that it leaves open is basically part of #162.

gnprice:
Looks good; merging, after that metadata tweak.

chrisbobbe:
Cool, thanks!

Issue #171, opened 2023-06-09
Show edited/moved marker on messages


gnprice:
When a message has been edited or moved, Zulip web shows this with a small "EDITED" or "MOVED" marker. This can be important for understanding a conversation; we should do either the same thing or another way of conveying the same information.

The Help Center describes this feature on the following page:
https://zulip.com/help/view-a-messages-edit-history
though that page also covers showing the details of the edits or moves, which is a separate issue #531.

### UI design

See first comments below, and see this later chat thread:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/check-x.20icons.20in.20left.20margin/near/1908288

### Implementation

We don't currently store the information of whether and how a message has been edited or moved. The first phase of work on this issue will be to start storing and maintaining that information. This will involve adding a field to the `Message` class, and then updating that field as appropriate in response to events.

Concretely, this field should look like `MessageEditState edited;` where `MessageEditState` is an enum with three values: none, moved, and edited. In other words it will store only the information needed to control the edited/moved marker; storing the full edit history is out of scope for this issue (and will be left for #531).

After that we can start showing the information in the UI.

### Related issues

These related features are all out of scope for this issue:
* #126 
* #530
* #531

(The details of the edits or moves can be important information too, but for a lot of purposes just the fact that the message *was* edited or moved, or the ability to confirm that it wasn't, is very helpful.)

Note that some of these have the "Post-launch" milestone — those are things we expect to get to much later than this issue.

terpimost:
The main idea was to use the left gutter to put a small icon of `edited` or `moved` + allow to swipe/drag the message horizontally to reveal full label:
<img width="915" alt="image" src="https://github.com/zulip/zulip-flutter/assets/1903309/e5d6c9bb-45b1-4095-a061-b2241f0fdca0">
This design isn't final, but let me know what do you think about interaction. Would it be hard to do or not?

gnprice:
Interesting, thanks.

I think that gesture should be doable from the implementation side. There may not be anything drop-in we can use for the whole swipe interaction; but we can give each message a [GestureDetector](https://api.flutter.dev/flutter/widgets/GestureDetector-class.html) that responds to [onHorizontalDragStart](https://api.flutter.dev/flutter/widgets/GestureDetector/onHorizontalDragStart.html) and its update/end/etc. friends, and build the interaction out of those pieces.

[...]

PR #543, opened 2024-03-01
#171 : Show edited/moved marker on messages

fizy069:
Haven't added the zulip icons yet, since I wasn't able to run the script to auto generate the icons.dart file.

Fixes: #171


fizy069:
CI failed probably because I did not add the generated files to the commit. Will do so soon.

gnprice:
Thanks for the contribution!

In addition to getting the existing tests passing, before we can review this PR in detail you'll need to write tests for it — see the note in the README file.

If you have questions as you try to do that, we're happy to discuss in `#mobile-team` on chat.zulip.org.

gnprice:
Please also post screenshots. That helps people review the UI changes.

gnprice:
I've edited the PR description to link it up with the issue, to make it easy to find when looking at the issue.

When the PR is ready for review, just post a comment here saying so and then I can take a look.

fizy069:
#F171
Here is how it looks currently.

Also, the PR is ready for review, could you please give your inputs?

lib/api/model/model.dart
```diff
@@ -664,6 +665,37 @@ enum MessageFlag {
   String toJson() => _$MessageFlagEnumMap[this]!;
 }
 
+@JsonSerializable(fieldRename: FieldRename.snake)
+class MessageEditHistory {
+
+  final String? prevContent;
```
gnprice, line 671: nit:
```suggestion
class MessageEditHistory {
  final String? prevContent;
```

lib/api/model/model.dart
```diff
@@ -664,6 +665,37 @@ enum MessageFlag {
   String toJson() => _$MessageFlagEnumMap[this]!;
 }
 
+@JsonSerializable(fieldRename: FieldRename.snake)
+class MessageEditHistory {
```
gnprice, line 669: nit: Let's put this at the end of this file, after `DmMessage`. That way it doesn't get between reading `Message` and reading `StreamMessage`, or `DmMessage`.

(Most of the time, the edit history isn't part of what one needs to think about when thinking about a message.)

lib/api/model/model.dart
```diff
@@ -664,6 +665,37 @@ enum MessageFlag {
...
+
+  final String? prevContent;
+  final String? prevRenderedContent;
+  final int? prevRenderedContentVersion;
+  final int? prevStream;
+  final String? prevTopic;
```
gnprice, line 675: The API docs say this field was renamed in Zulip Server 5.0. This codebase currently still supports back to Zulip Server 4.0 — see the "Editing API types" section in the README. So this should handle the old name too.

To avoid the rest of the code needing to look for both names, we should have the class use only the new name, but have the deserialization code look for the old name if it doesn't find the new name. See `User.isSystemBot` for an example of how to do that.

(Then that will also be a case that should definitely get a test.)

lib/api/model/model.dart
```diff
@@ -664,6 +665,37 @@ enum MessageFlag {
...
+    required this.prevStream,
+    required this.prevTopic,
+    required this.stream,
+    required this.timestamp,
+    required this.topic,
+    required this.userId,
+  });
+
+  factory MessageEditHistory.fromJson(Map<String, dynamic> json) =>
+      _$MessageEditHistoryFromJson(json);
```
gnprice, line 694: nit: two-space indents
```suggestion
  factory MessageEditHistory.fromJson(Map<String, dynamic> json) =>
    _$MessageEditHistoryFromJson(json);
```

lib/api/model/model.dart
```diff
@@ -801,4 +835,4 @@ class DmMessage extends Message {
 
   @override
   Map<String, dynamic> toJson() => _$DmMessageToJson(this);
-}
+}
```
gnprice, line 838: This final line is incomplete. See https://github.com/zulip/zulip-flutter/pull/534#discussion_r1515353889 for discussion.

lib/api/model/model.dart
```diff
@@ -569,7 +569,7 @@ sealed class Message {
   String content;
   final String contentType;
 
-  // final List<MessageEditHistory> editHistory; // TODO handle
+  final List<MessageEditHistory>? editHistory; // TODO handle
```
gnprice, line 572: This TODO comment should be removed — it's calling for us to do what this PR is doing :slightly_smiling_face:

lib/widgets/message_list.dart
```diff
@@ -892,6 +893,16 @@ class MessageWithPossibleSender extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final message = item.message;
+    bool messageEdited = false, messageMoved = false;
+    if(message.editHistory != null) {
+      if((message.editHistory?[message.editHistory!.length-1].topic !=
+          message.editHistory?[message.editHistory!.length-1].prevTopic) ||
```
gnprice, line 899: This is pretty messy to read. The repeated subexpression here should be pulled out as a local variable.

Also see the `Iterable.last` getter :-)

lib/widgets/message_list.dart
```diff
@@ -892,6 +893,16 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+          (message.editHistory?[message.editHistory!.length-1].stream !=
+          message.editHistory?[message.editHistory!.length-1].prevStream)){
+            messageMoved = true;}
+      if(message.editHistory?[message.editHistory!.length-1].prevContent!=null){
+        messageEdited = true;
```
gnprice, line 904: What does this do if the message is edited, and then it's moved? What's the desired behavior in that case?

lib/api/model/model.dart
gnprice, line 572: This new field needs to be kept up to date in response to server events. From #171:
> The first phase of work on this issue will be to start storing and maintaining that information. This will involve adding a field to the `Message` class, and then updating that field as appropriate in response to events.

windows/flutter/CMakeLists.txt
```diff
@@ -10,6 +10,11 @@ include(${EPHEMERAL_DIR}/generated_config.cmake)
 # https://github.com/flutter/flutter/issues/57146.
 set(WRAPPER_ROOT "${EPHEMERAL_DIR}/cpp_client_wrapper")
 
+# Set fallback configurations for older versions of the flutter tool.
+if (NOT DEFINED FLUTTER_TARGET_PLATFORM)
+  set(FLUTTER_TARGET_PLATFORM "windows-x64")
```
gnprice, line 15: This change looks unrelated to the rest of the PR. As part of making clear and coherent commits:
it'll therefore need to be its own commit, with a commit message explaining why we want to make this change.

gnprice (commented):
Thanks for the revision!

Some quick comments below on the code so far.

One thing we'll need before we can fully review this PR is for the changes to be organized into [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html). See the relevant README section:
https://github.com/zulip/zulip-flutter#submitting-a-pull-request

From the screenshots, it looks like this UI doesn't yet match the design in at least one important way: pay close attention to where the edited/moved markers are when at rest, before swiping to show the word "Edited" or "Moved".

lib/widgets/message_list.dart
fizy069: If a message has been edited and moved, or has been moved and then edited - I believe it should prioritize showing edited over moved in both cases. Please correct me if I'm wrong.

gnprice: Yeah, that's right.

PR #544, opened 2024-03-01
content: Show code, mentions, and times properly even in headings etc.

chrisbobbe:
Also, use `weightVariableTextStyle` in one place we forgot to.

Fixes: #538

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -497,7 +497,7 @@ class MathBlock extends StatelessWidget {
 //
 
 Widget _buildBlockInlineContainer({
-  required TextStyle? style,
+  required TextStyle style,
   required BlockInlineContainerNode node,
 }) {
   if (node.links == null) {
@@ -513,7 +513,7 @@ class _BlockInlineContainer extends StatefulWidget {
     {required this.links, required this.style, required this.nodes});
 
   final List<LinkNode> links;
-  final TextStyle? style;
+  final TextStyle style;
   final List<InlineContentNode> nodes;
 
   @override
@@ -576,7 +576,7 @@ class InlineContent extends StatelessWidget {
 
   final GestureRecognizer? recognizer;
   final Map<LinkNode, GestureRecognizer>? linkRecognizers;
-  final TextStyle? style;
+  final TextStyle style;
   final List<InlineContentNode> nodes;
 
   late final _InlineContentBuilder _builder;
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -179,8 +179,9 @@ class Heading extends StatelessWidget {
       child: _buildBlockInlineContainer(
         style: TextStyle(
           fontSize: kBaseFontSize * emHeight,
-          fontWeight: FontWeight.w600,
-          height: 1.4),
+          height: 1.4,
+        )
+          .merge(weightVariableTextStyle(context, wght: 600)),
         node: node));
   }
 }
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -1198,8 +1198,9 @@ InlineSpan _errorUnimplemented(UnimplementedNode node) {
   }
 }
 
-const errorStyle = TextStyle(fontWeight: FontWeight.bold, color: Colors.red);
+const errorStyle = TextStyle(
+  fontSize: kBaseFontSize, fontWeight: FontWeight.bold, color: Colors.red);
 
 final errorCodeStyle = kMonospaceTextStyle
-  .merge(const TextStyle(color: Colors.red))
+  .merge(const TextStyle(fontSize: kBaseFontSize, color: Colors.red))
   .merge(weightVariableTextStyle(null)); // TODO(a11y) pass a BuildContext
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -571,12 +571,19 @@ class InlineContent extends StatelessWidget {
     required this.style,
     required this.nodes,
   }) {
+    assert(style.fontSize != null);
     _builder = _InlineContentBuilder(this);
   }
 
   final GestureRecognizer? recognizer;
   final Map<LinkNode, GestureRecognizer>? linkRecognizers;
+
+  /// A [TextStyle] applied to this content and provided to descendants.
+  ///
+  /// Must set [TextStyle.fontSize]. Some descendant spans will consume it,
+  /// e.g., to make their content slightly smaller than surrounding text.
   final TextStyle style;
+
   final List<InlineContentNode> nodes;
 
   late final _InlineContentBuilder _builder;
@@ -650,18 +657,21 @@ class _InlineContentBuilder {
       return _buildInlineCode(node);
     } else if (node is UserMentionNode) {
       return WidgetSpan(alignment: PlaceholderAlignment.middle,
-        child: UserMention(node: node));
+        child: UserMention(surroundingTextStyle: widget.style, node: node));
     } else if (node is UnicodeEmojiNode) {
       return TextSpan(text: node.emojiUnicode, recognizer: _recognizer);
     } else if (node is ImageEmojiNode) {
       return WidgetSpan(alignment: PlaceholderAlignment.middle,
         child: MessageImageEmoji(node: node));
     } else if (node is MathInlineNode) {
-      return TextSpan(style: _kInlineMathStyle,
+      return TextSpan(
+        style: widget.style
+          .merge(_kInlineMathStyle)
+          .apply(fontSizeFactor: _kInlineCodeFontSizeFactor),
         children: [TextSpan(text: node.texSource)]);
     } else if (node is GlobalTimeNode) {
       return WidgetSpan(alignment: PlaceholderAlignment.middle,
-        child: GlobalTime(node: node));
+        child: GlobalTime(node: node, surroundingTextStyle: widget.style));
     } else if (node is UnimplementedInlineContentNode) {
       return _errorUnimplemented(node);
     } else {
@@ -713,8 +723,13 @@ class _InlineContentBuilder {
 
     // TODO `code`: find equivalent of web's `unicode-bidi: embed; direction: ltr`
 
-    // Use a light gray background, instead of a border.
-    return _buildNodes(style: _kInlineCodeStyle, node.nodes);
+    return _buildNodes(
+      // Use a light gray background, instead of a border.
+      style: widget.style
+        .merge(_kInlineCodeStyle)
+        .apply(fontSizeFactor: _kInlineCodeFontSizeFactor),
+      node.nodes,
+    );
 
     // Another fun solution -- we can in fact have a border!  Like so:
     //   TextStyle(
@@ -734,17 +749,27 @@ class _InlineContentBuilder {
   }
 }
 
+/// The [TextStyle] for inline math, excluding font-size adjustment.
+///
+/// Inline math should use this and also apply [_kInlineCodeFontSizeFactor]
+/// to the font size of the surrounding text
+/// (which might be a Paragraph, a Heading, etc.).
 final _kInlineMathStyle = _kInlineCodeStyle.merge(TextStyle(
   backgroundColor: const HSLColor.fromAHSL(1, 240, 0.4, 0.93).toColor()));
 
+const _kInlineCodeFontSizeFactor = 0.825;
+
+/// The [TextStyle] for inline code, excluding font-size adjustment.
+///
+/// Inline code should use this and also apply [_kInlineCodeFontSizeFactor]
+/// to the font size of the surrounding text
+/// (which might be a Paragraph, a Heading, etc.).
 // Even though [kMonospaceTextStyle] is a variable-weight font,
 // it's acceptable to skip [weightVariableTextStyle] here,
 // assuming the text gets the effect of [weightVariableTextStyle]
 // through inheritance, e.g., from a [DefaultTextStyle].
 final _kInlineCodeStyle = kMonospaceTextStyle
-  .merge(const TextStyle(
-    backgroundColor: Color(0xffeeeeee),
-    fontSize: 0.825 * kBaseFontSize));
+  .merge(const TextStyle(backgroundColor: Color(0xffeeeeee)));
 
 final _kCodeBlockStyle = kMonospaceTextStyle
   .merge(const TextStyle(
@@ -775,8 +800,13 @@ final _kCodeBlockStyle = kMonospaceTextStyle
 // const _kInlineCodeRightBracket = '⟩';
 
 class UserMention extends StatelessWidget {
-  const UserMention({super.key, required this.node});
+  const UserMention({
+    super.key,
+    required this.surroundingTextStyle,
+    required this.node,
+  });
 
+  final TextStyle surroundingTextStyle;
   final UserMentionNode node;
 
   @override
@@ -790,7 +820,7 @@ class UserMention extends StatelessWidget {
         // One hopes an @-mention can't contain an embedded link.
         // (The parser on creating a UserMentionNode has a TODO to check that.)
         linkRecognizers: null,
-        style: Paragraph.textStyle,
+        style: surroundingTextStyle,
         nodes: node.nodes));
   }
 
@@ -857,9 +887,14 @@ class MessageImageEmoji extends StatelessWidget {
 }
 
 class GlobalTime extends StatelessWidget {
-  const GlobalTime({super.key, required this.node});
+  const GlobalTime({
+    super.key,
+    required this.node,
+    required this.surroundingTextStyle,
+  });
 
   final GlobalTimeNode node;
+  final TextStyle surroundingTextStyle;
 
   static final _backgroundColor = const HSLColor.fromAHSL(1, 0, 0, 0.93).toColor();
   static final _borderColor = const HSLColor.fromAHSL(1, 0, 0, 0.8).toColor();
@@ -867,6 +902,8 @@ class GlobalTime extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final surroundingFontSize = surroundingTextStyle.fontSize!;
+
     // Design taken from css for `.rendered_markdown & time` in web,
     //   see zulip:web/styles/rendered_markdown.css .
     final text = _dateFormat.format(node.datetime.toLocal());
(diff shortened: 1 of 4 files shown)


lib/widgets/content.dart
```diff
@@ -750,9 +771,14 @@ final _kCodeBlockStyle = kMonospaceTextStyle
...
 class UserMention extends StatelessWidget {
-  const UserMention({super.key, required this.node});
+  const UserMention({
+    super.key,
+    required this.node,
+    required this.surroundingTextStyle,
+  });
 
   final UserMentionNode node;
+  final TextStyle surroundingTextStyle;
```
gnprice, line 781: nit: I think of the `node` field of these content widgets as somewhat akin to `child` — it's the payload of the widget, describing what content it should show, whereas this other field is some metadata about how to show it. So let's put the node at the end.

(From a quick search for "node;" in this file, it looks like these are actually the first examples of a content widget that takes any other field in addition to a "node".)

gnprice (commented):
Thanks! The code all looks good with one small nit below.

Then: can you write a test for this?

For example, we could render the content
```
# one `two`

three `four`
```
and check that the font sizes on the spans reading "one" and "two" are in the same ratio to each other as the font sizes on the spans reading "three" and "four". Then similarly for @-mentions and global times.

chrisbobbe:
Thanks for the review! Still working on the test.

> The code all looks good with one small nit below.

Cool. I'll go ahead and merge the first five commits, then:

content test: Add ContentExample.globalTime
content test: Add ContentExample.strong, and its widget smoke test
content test: Add ContentExample.emphasis, and its widget smoke test
content test: Add ContentExample.inlineCode, and its widget smoke test
content test: Add ContentExamples for @-mentions, also widget smoke tests

since they're helpful independently of the interesting changes for this bugfix.

gnprice:
Those sound useful! But I don't actually see them in this PR — is there somewhere else you're thinking of?

chrisbobbe:
Oh goodness! 😅 Thanks for noticing that. It turns out I was looking at my local branch, and I forgot that these commits were not included in what you had reviewed already. I'll hold off until I'm ready with a new revision, which will include tests for the bugfix (with the complicated recurse-through-spans-merging-styles helper).

chrisbobbe:
OK, revision pushed, with those complicated tests as described in https://github.com/zulip/zulip-flutter/pull/544#pullrequestreview-1915167215.

…Actually this revision doesn't yet test @-mentions or global times (or inline math) in that way; just inline code. But if the current implementation looks good for testing inline code, I can then apply it to those other things too.

test/model/content_test.dart
```diff
@@ -633,30 +683,16 @@ void main() {
...
-      // "@_**Greg Price**"
-      '<p><span class="user-mention silent" data-user-id="2187">Greg Price</span></p>',
-      const UserMentionNode(nodes: [TextNode('Greg Price')]));
+    testParseExample(ContentExample.userMentionPlain);
+    testParseExample(ContentExample.userMentionSilent);
 
     testParseInline('silent user @-mention, class order reversed',
       // "@_**Greg Price**" (hypothetical server variation)
       '<p><span class="silent user-mention" data-user-id="2187">Greg Price</span></p>',
       const UserMentionNode(nodes: [TextNode('Greg Price')]));
```
gnprice, line 692: I think these classes-flipped examples are a bit easier to read if they're right next to their unflipped counterparts — that way the reader can easily see that that's the only thing different. So it'd be good to convert them at the same time.

lib/widgets/content.dart
```diff
@@ -752,22 +778,27 @@ final _kCodeBlockStyle = kMonospaceTextStyle
...
 
+  final TextStyle surroundingTextStyle;
   final UserMentionNode node;
 
   @override
   Widget build(BuildContext context) {
     return Container(
       decoration: _kDecoration,
-      padding: const EdgeInsets.symmetric(horizontal: 0.2 * kBaseFontSize),
+      padding: EdgeInsets.symmetric(horizontal: 0.2 * surroundingTextStyle.fontSize!),
```
gnprice, line 794: Interesting, it looks like the revision I previously reviewed didn't have this change. Nor the similar padding change for dates, or the one for the clock icon size.

Would you update the screenshot to reflect this revision?

test/widgets/content_test.dart
```diff
@@ -290,6 +297,79 @@ void main() {
...
+
+  group('InlineCodeNode', () {
+    testContentSmoke(ContentExample.inlineCode);
+
+    testWidgets('maintains font-size ratio with surrounding text', (tester) async {
+      /// Simulate a nested span's style by merging all ancestor-span styles,
+      /// starting from the root.
+      ///
+      /// This isn't how the style actually gets applied in the code under test,
+      /// but it should hopefully simulate it closely enough.
```
gnprice, line 312:
```suggestion
      /// This isn't how the style actually gets applied in the code under test
      /// (because that happens inside the engine, in SkParagraph),
      /// but it should hopefully simulate it closely enough.
```

That helps answer the question of why we *don't* just do what the code under test actually does.

test/widgets/content_test.dart
```diff
@@ -290,6 +297,79 @@ void main() {
...
+          if (span == innerSpan) {
+            return false;
+          }
+          final notInSubtree = span.visitDirectChildren(recurse);
+          if (notInSubtree) {
+            styles.removeLast();
+          }
+          return notInSubtree;
+        }
+        recurse(outerSpan);
```
gnprice, line 326: Let's have this use the return value to assert that `innerSpan` was actually found somewhere.

gnprice: Also the dartdoc can say explicitly that `innerSpan` is required to be a descendant of `outerSpan` — that should help a reader fit the pieces together of what this function is doing.

test/widgets/content_test.dart
```diff
@@ -290,6 +297,79 @@ void main() {
...
+        final styles = <TextStyle?>[];
+        bool recurse(InlineSpan span) {
+          styles.add(span.style);
+          if (span == innerSpan) {
+            return false;
+          }
+          final notInSubtree = span.visitDirectChildren(recurse);
+          if (notInSubtree) {
+            styles.removeLast();
+          }
```
gnprice, line 323: Here's a variant that's slightly more efficient:
```suggestion
          if (span == innerSpan) {
            styles.add(span.style);
            return false;
          }
          final notInSubtree = span.visitDirectChildren(recurse);
          if (!notInSubtree) {
            styles.add(span.style);
          }
```

This way we only start adding to the list when we find the inner span, and never add anything we later have to remove.

The list goes from inner to outer when built this way, rather than outer to inner; we can then say `styles.reversed.reduce(…)` below.

test/widgets/content_test.dart
```diff
@@ -290,6 +297,79 @@ void main() {
...
+          return notInSubtree;
+        }
+        recurse(outerSpan);
+
+        return styles.reduce((value, element) => switch ((value, element)) {
+          (TextStyle(), TextStyle()) => value!.merge(element!),
```
gnprice, line 329: After the comment above about the algorithm, I notice that these tests pass just the same if I reverse the order these get merged in. It's fairly subtle to get that ordering the right way around, so it'd be good to have a check to prevent it accidentally breaking. (When merged in the wrong order, all the fonts are the same — so the ratio check always passes, without successfully testing what we want to test.)

Probably the simplest way to do that is to check that the font size is bigger in the header than the paragraph, and bigger in the plain text than the inline code — say by at least 10% in each case. That's definitely a solid expectation for header vs. paragraph, and mostly solid for plain vs. code; and if someday we change the code style so the latter no longer holds, we can figure out a different check.

test/widgets/content_test.dart
```diff
@@ -290,6 +297,79 @@ void main() {
...
+      /// Simulate a nested span's style by merging all ancestor-span styles,
+      /// starting from the root.
+      ///
+      /// This isn't how the style actually gets applied in the code under test,
+      /// but it should hopefully simulate it closely enough.
+      TextStyle? mergeStylesOuterToInner(InlineSpan outerSpan, InlineSpan innerSpan) {
+        final styles = <TextStyle?>[];
+        bool recurse(InlineSpan span) {
+          styles.add(span.style);
+          if (span == innerSpan) {
```
gnprice, line 317:
```suggestion
          if (identical(span, innerSpan)) {
```

I.e. the equivalent of JS `===`.

Otherwise this calls the `operator ==` override, which can accept more things as equal.

test/widgets/content_test.dart
```diff
@@ -290,6 +297,79 @@ void main() {
...
+          (null, TextStyle())        => element,
+          (null, null)               => null,
+        });
+      }
+
+      TextStyle? mergedStyleOfSubstring(InlineSpan rootSpan, String substring) {
+        final substringSpan = rootSpan.getSpanForPosition(
+          TextPosition(offset: rootSpan.toPlainText().indexOf(substring)));
+        check(substringSpan).isNotNull();
+        return mergeStylesOuterToInner(rootSpan, substringSpan!);
```
gnprice, line 340: This works fine for this test, and if you find it works fine for the other test cases then it can stay just as it is.

A potentially cleaner, and certainly more flexible, version would have the recursion take not an InlineSpan to look for, but a predicate to call on each descendant span to see if it's the one we're looking for. Then the predicate here can be like `(span) => span.text == substring`. That way we recurse through the tree only once, and it opens the way for different predicates for other test cases if needed.

gnprice (commented):
Thanks! This looks like a good structure; comments below.

lib/widgets/content.dart
chrisbobbe: Ah, yeah. Done.

gnprice: Cool, thanks.

Looking at that screenshot, I think it's probably best to go ahead and make this adjustment for the clock icon, but *not* for the padding. The expanded padding feels kind of out of scale (even though in a literal sense it is precisely to scale).

It's reminiscent of the situation described at https://github.com/zulip/zulip-mobile/blob/dff4d4d5ed495aba0e97f105215ce425909e6d53/docs/background/webview.md#why-mostly-px-and-not-rem :
> Typical advice for web design is to use `rem` or `em` for nearly all lengths. This means that when the font size grows or shrinks, not only the text but the whole layout scales with it, including padding, margin, and images.
>
> But on mobile, if you go to a device's settings and [change the font size](https://github.com/zulip/zulip-mobile/blob/dff4d4d5ed495aba0e97f105215ce425909e6d53/docs/background/layout.md), you'll find this is not at all what apps do. When you make the font size giant, the text in an app grows with it -- but everything that _isn't_ text stays the same. Icons in the UI; user avatars; padding around text; padding or margin between elements; indentation of items in a list; all stay exactly the same. This is true on Android and iOS, and of flagship apps from Google, Apple, Facebook, and others.

Arguably that'd call for not scaling the clock icon either; but I think the clock icon looks best when it does get scaled. Effectively it feels like part of the text — like a sort of custom emoji.

chrisbobbe:
Thanks for the review! Revision pushed, and I've added tests for the sizes of inline math, @-mentions, and global times as well. Those are all still pretty verbose even with the helper code…I'm not sure if that's worth spending time on right now.

test/widgets/content_test.dart
```diff
@@ -290,6 +339,88 @@ void main() {
...
+        '</h1>\n'
+        '<p>'
+          'paragraph-plain '
+          '<span class="user-mention" data-user-id="13313">@Chris Bobbe</span>'
+        '</p>');
+
+      final headerRootSpan = tester.renderObject<RenderParagraph>(find.textContaining('header')).text;
+      final headerPlainStyle = mergedStyleOfSubstring(headerRootSpan, 'header-plain ');
+      final headerMentionStyle = mergeSpanStylesOuterToInner(headerRootSpan,
+        (span) => span is WidgetSpan && span.child is UserMention);
```
gnprice, line 401: This test seems not to actually work — if I check out the old code in `lib`, the test still passes.

gnprice (commented):
Thanks! Generally this looks good.

Two small comments below, and I also took a bit of time just now to try factoring out the common logic from these tests. Here's the diffstat, including some comments:
```
 test/widgets/content_test.dart | 210 +++++++++++++++-------------------------------
 1 file changed, 66 insertions(+), 144 deletions(-)
```
I'll push that to the branch as additional commits on top; PTAL.

Then when I was testing those test changes, I found that one of these four test cases doesn't work as a test — and that's already true in the PR branch. So that'll be good to fix.

chrisbobbe:
> Two small comments below, and I also took a bit of time just now to try factoring out the common logic from these tests. Here's the diffstat, including some comments:
> 
> ```
>  test/widgets/content_test.dart | 210 +++++++++++++++-------------------------------
>  1 file changed, 66 insertions(+), 144 deletions(-)
> ```
> 
> I'll push that to the branch as additional commits on top; PTAL.

Ah I think you meant to push this but haven't yet; is that right?

gnprice:
Ah indeed, thanks for the ping! Pushed now.

test/widgets/content_test.dart
gnprice: There's ultimately a text span that has the mentioned user's name somewhere inside that `UserMention` widget, and the text style will live there, right?

Specifically I think if we find the `RenderParagraph` that text span is part of, and merge together its styles in the same way as before, then that should tell us the effective style for that span. In particular I believe all the defaulting and merging that happens outside of `RenderParagraph` and SkParagraph winds up going into the `TextStyle` that gets placed at the root of that `RenderParagraph`'s span tree.

chrisbobbe:
Thanks for the review and those helpful refactoring commits! Revision pushed, with those squashed into the main commit, along with some refactoring to test the things that use `WidgetSpan`, following https://github.com/zulip/zulip-flutter/pull/544#discussion_r1548731285 .

test/widgets/content_test.dart
```diff
@@ -442,18 +583,69 @@ void main() {
...
-      '<p><time datetime="2024-01-30T17:33:00Z">2024-01-30T17:33:00Z</time></p>'
-    ).nodes)));
+  group('GlobalTimeNode', () {
     // The time is shown in the user's timezone and the result will depend on
-    // the timezone of the environment running this test. Accept here a wide
+    // the timezone of the environment running these tests. Accept here a wide
     // range of times. See comments in "show dates" test in
     // `test/widgets/message_list_test.dart`.
-    tester.widget(find.textContaining(RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d [AP]M$')));
+    final renderedTextRegexp = RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d [AP]M$');
```
gnprice, line 605: This regexp only really makes sense when juxtaposed with the particular time value used in these tests. So if we're factoring out the regexp, let's factor out that time HTML too:
```suggestion
    final renderedTextRegexp = RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d [AP]M$');
    const timeSpanHtml = '<time datetime="2024-01-30T17:33:00Z">2024-01-30T17:33:00Z</time>';
```

(or perhaps better, HTML first and then comment and regexp)

This means the smoke test will splice that HTML inside a `<p>` element; but that's already pretty much what `checkFontSizeRatio` is going to do.

test/widgets/content_test.dart
```diff
@@ -25,6 +28,55 @@ import 'dialog_checks.dart';
...
+/// The "merged style" ([mergeSpanStylesOuterToInner]) of a nested span.
+TextStyle? mergedStyleOfSubstring(InlineSpan rootSpan, Pattern substringPattern) {
+  return mergeSpanStylesOuterToInner(rootSpan,
+    (span) {
+      if (span is! TextSpan) return false;
+      final text = span.text;
+      if (text == null) return false;
+      return switch (substringPattern) {
+        String() => text == substringPattern,
+        _ => substringPattern.allMatches(text).isNotEmpty,
```
gnprice, line 75: These two cases don't agree in behavior — a string as a Pattern means "has this as a substring", not "equals this".

It looks like this revision doesn't end up actually using the more general Pattern case. So I think the cleanest thing is to go back to having the parameter be `String substring`, as in the previous revision.

test/widgets/content_test.dart
```diff
@@ -278,6 +335,90 @@ void main() {
...
+  /// the text around them.
+  ///
+  /// The given `targetHtml` should be a self-contained HTML fragment
+  /// that parses as an [InlineContentNode].
+  ///
+  /// Exactly one of `targetString` and `targetPredicate` should be provided,
+  /// and should match the [InlineSpan] rendered from that [InlineContentNode].
+  Future<void> checkFontSizeRatio(WidgetTester tester, {
+    required String targetHtml,
+    required double Function(InlineSpan rootSpan) targetFontSizeFinder,
```
gnprice, line 354: doc should be updated to reflect `targetFontSizeFinder`

gnprice (commented):
Thanks for the revision! Generally this all looks good — some small comments below.

test/widgets/content_test.dart
```diff
@@ -40,10 +92,15 @@ void main() {
...
-      localizationsDelegates: ZulipLocalizations.localizationsDelegates,
-      supportedLocales: ZulipLocalizations.supportedLocales,
-      home: BlockContentList(nodes: parseContent(html).nodes),
+    await tester.pumpWidget(Builder(
+      builder: (context) {
+        return MaterialApp(
+          theme: ThemeData(typography: zulipTypography(context)),
+          localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+          supportedLocales: ZulipLocalizations.supportedLocales,
+          home: Scaffold(body: BlockContentList(nodes: parseContent(html).nodes)),
```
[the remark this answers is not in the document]
chrisbobbe: Huh interesting. I can't remember why I added the Scaffold.

I guess it does bring the tests a little closer to being representative, so it doesn't hurt to keep it there…

test/widgets/content_test.dart
chrisbobbe: This revision uses the more general `Pattern` case to find text matching the RegExp we use to recognize a rendered global time:

```dart
RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d [AP]M$')
```

chrisbobbe: Maybe one option is to make the cases agree in behavior, by only considering matches that cover the whole string?

```dart
      return switch (substringPattern) {
        String() => text == substringPattern,
        _ => substringPattern.allMatches(text)
          .any((match) => match.start == 0 && match.end == text.length),
      };
```

chrisbobbe:
Thanks for the review! Revision pushed, and please see one reply above: https://github.com/zulip/zulip-flutter/pull/544#discussion_r1591781933

test/widgets/content_test.dart
gnprice: Hmm, I see, yeah.

That solution LGTM.

Issue #545, opened 2024-03-01
Zero out letter spacing in message content


gnprice:
It's currently 0.25px, a default we inherited [from Material](https://m3.material.io/styles/typography/type-scale-tokens) (for the `bodyMedium` style). That's too wide for reading prose, at least with the font we're using.

Instead we want `letterSpacing: 0`.

This should apply to both the message content in the message list, and the message content the user types in the compose box.

[Chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/letter.20spacing/near/1749389)

A followup issue is:
* #548

gnprice:
Two other areas, from further discussion on the chat thread:
* [Let's set](https://chat.zulip.org/#narrow/stream/48-mobile/topic/letter.20spacing/near/1750529) letter spacing in the compose box the same way (i.e. to 0px).
* [Let's set](https://chat.zulip.org/#narrow/stream/48-mobile/topic/letter.20spacing/near/1751385) letter spacing to 1% (that is, .01 times the font size) for buttons.

Issue #548, opened 2024-03-05
Set letter spacing to 1% in buttons


gnprice:
This is a followup to #545. That issue is focused on the one most conspicuous kind of text in the app; this one is for the rest of the app.

Specifically we want:
* letter spacing of 1% (i.e., 0.01 times `fontSize`) in buttons;
* ~~`letterSpacing: 0` everywhere else, except where the Figma design specifies otherwise.~~ (This part was handled in #549.)

[Chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/letter.20spacing/near/1751385).

Gosplex:
Hello @gnprice 
I am interested in working on this issue
Please could i get started?

gnprice:
Hi @Gosplex, welcome.

I think we're not going to want to do this by setting `letterSpacing` individually on each of the buttons in the app, or on each of the other pieces of text. That would make it error-prone because it'd be easy to forget to include it in a given place when adding a new piece of UI, and it'd also make for additional noise when reading the code for each place text appears.

Instead, we'll use the "theme" features of Flutter's Material Design implementation. See `TextTheme` and related docs:
https://main-api.flutter.dev/flutter/material/TextTheme-class.html
https://main-api.flutter.dev/flutter/material/Typography-class.html

Your first step will be to find the existing code in the app that works with those APIs, and start to get an idea of how you'd approach addressing this issue there.

There's also several button-related "theme" classes, which control the text (and other things) in buttons specifically. So you'll want to find those and get an idea of how you'd address the buttons part of this issue using those.

Gosplex:
After reviewing the resources on Flutter's theming, I suggest centralizing `textTheme` and `buttonTheme` declarations using `ThemeData` in a dedicated file. This ensures consistent letter-spacing and style throughout the app, improving maintainability. I'm open to suggestions for further refinement.

Gosplex:
Good day
So I have carefully gone through the codebase  and the help material provided by @gnprice  and I found out that the `ThemeData` of the app is managed at `app.dart` file. So any changes made there would be made available throughout the app

gnprice:
Cool, please go ahead. Note that part of this issue is now done, so the remaining part is the part about 1% letter spacing in buttons.

For an example of how we adjust themes systematically, take a look at #549 and the file that PR modified. We'll need to do something similar for the button themes.

Gosplex:
@gnprice  Sir
I have created a pull request for this issue.
Please review. 
[...]

PR #549, opened 2024-03-07
text: Set 0 letter spacing throughout zulipTypography

chrisbobbe:
Screenshots coming soon.

After this, there's still more to do for #548:

- letter spacing of 1% (i.e., 0.01 times `fontSize`) in buttons
- perhaps audit for UI code that's meant to follow Figma already,
  but where we failed to set a non-zero letter spacing that was
  specified in Figma

Fixes: #545
Fixes-partly: #548

### DIFF
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -36,6 +36,10 @@ Typography zulipTypography(BuildContext context) {
 
   convertGeometry(TextTheme inputTextTheme) {
     TextTheme result = _weightVariableTextTheme(context, inputTextTheme);
+
+    result = _convertTextTheme(result, (maybeInputStyle, _) =>
+      maybeInputStyle?.merge(const TextStyle(letterSpacing: 0)));
+
     return result;
   }
 
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -34,6 +34,11 @@ import 'package:flutter/material.dart';
 Typography zulipTypography(BuildContext context) {
   final typography = Theme.of(context).typography;
 
+  convertGeometry(TextTheme inputTextTheme) {
+    TextTheme result = _weightVariableTextTheme(context, inputTextTheme);
+    return result;
+  }
+
   Typography result = typography.copyWith(
     black: typography.black.apply(
       fontFamily: kDefaultFontFamily,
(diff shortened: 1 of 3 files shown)


chrisbobbe:
| Before | After |
| --- | --- |
| image | image |

gnprice:
Thanks! This looks good; merging.

From the commit message:

    It's worth mentioning that letter spacing is one of the parameters
    that are carefully chosen for the 15 TextThemes that make up the
    default Typography in Material 3:
    
      {display,headline,title,label,body} × {small,medium,large}
    
    So this might be considered a regression...if the Material spec were
    our primary or only source of truth for text styles. But it isn't:
    we have a designer (Vlad), and this change brings us closer to his
    specified design. See #545 and #548.

Yeah. I also navigated around the app a bit and looked at various screens, and there wasn't a lot of text that was affected that wasn't similar in character to what's in your screenshots above of the message list. Stream names in the streams screen; the miscellaneous text at top of HomeScreen (which we should probably cut anyway — it's from literally when I was demonstrating to myself that registerQueue was working, before there was even a message list); things like that.

In particular the titles in app bars seem unaffected, like in your screenshots.

FTR, here's the Material default font styles that we're overriding:
https://m3.material.io/styles/typography/type-scale-tokens#a734c6ed-634c-4abb-adb2-35daf0aed06a
Bit annoying to click through and find (it's not a nice presentation like a table), but the information is there, called "font tracking".

    After this, there's still more to do for #548:
    
    - letter spacing of 1% (i.e., 0.01 times `fontSize`) in buttons
    - perhaps audit for UI code that's meant to follow Figma already,
      but where we failed to set a non-zero letter spacing that was
      specified in Figma

I don't know a reasonable way to do that audit, unfortunately. If we could get an export of everything in such a way that we could search through for "letter-spacing" and scan for matches that are in UI we've already implemented from Figma, then it'd be pretty manageable — but I definitely don't want to try clicking through opening a bunch of nodes across the Figma design to look for letter spacing. So I take #548 as being about the buttons, and the places where a letter-spacing *isn't* specified, and if there are any other discrepancies then we can just file and fix those as and when we notice them.

(I'll adjust this last part of the commit message accordingly before merge.)

PR #553, opened 2024-03-08
Upgrade Gradle, AGP, other Android build deps and metadata

gnprice:
There are currently some warnings that `flutter run` prints for Android targets. After the Android notifications work I did a couple of weeks ago, I sat down to resolve those:
* migrating how we invoke Flutter's Gradle code
* upgrading the Kotlin version we use

and then went on to upgrade our versions of Gradle and the Android Gradle Plugin too. Most of this PR comes from that time; then I came back and finished up the last AGP upgrade, which involved a couple of extra wrinkles.

Other than clearing warnings, some of these changes are supposed to reduce the build time, which is nice. I haven't measured the effect, though — it's possible our project isn't large enough to really see the difference.

I validated things by building and running the app and checking that
* it seems to work on a cursory check;
* it successfully receives notifications, both in a debug and a release build, and in all three conditions of running in the foreground, running in the background, and not already running at all.

  (This last is the one area where I have any particular concerns about some part of the app failing to work due to Android build changes, without the whole app failing to build in the first place. Few other places in our code interact with the Android layer at all, and basically none of the most important functionality.)

---

This doesn't go quite to the very latest AGP (8.2.2), because that upgrade causes a couple of build errors. They look like they might not be that deep; but I've spent enough time on this for now, so shipping the upgrades I have rather than investigating. I don't know of anything we'd currently gain by taking the upgrade from 8.1.x to 8.2.x anyway.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -20,6 +20,8 @@ try {
 }
 
 android {
+    namespace "com.zulip.flutter"
+
     compileSdkVersion flutter.compileSdkVersion
     ndkVersion flutter.ndkVersion
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -1,3 +1,9 @@
+plugins {
+    id 'com.android.application'
+    id 'kotlin-android'
+    id 'dev.flutter.flutter-gradle-plugin'
+}
+
 def localProperties = new Properties()
 def localPropertiesFile = rootProject.file('local.properties')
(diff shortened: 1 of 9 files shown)


chrisbobbe:
Hmm, shellcheck is reporting an error.

gnprice:
Fixed, thanks.

chrisbobbe:
LGTM, thanks! Merged.

PR #557, opened 2024-03-08
Added letterSpacing to buttons

Gosplex:
This pull request introduces changes to the theme data configuration in the Zulip app. Specifically, it updates the button theme to apply consistent letter spacing across all types of buttons: `elevated buttons`, `outlined buttons,` and `text buttons`. The `letterSpacing ` is set to `0.01`  times the font size for improved readability and visual consistency.

Fixes: #548

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -94,6 +94,36 @@ class ZulipApp extends StatelessWidget {
 
         shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
       ),
+      elevatedButtonTheme: ElevatedButtonThemeData(
+        // Apply letter spacing to the text style for elevated buttons
+        style: ButtonStyle(
+          textStyle: MaterialStateProperty.all<TextStyle>(
+            const TextStyle(
+              letterSpacing: 0.01,
+            ),
+          ),
+        ),
+      ),
+      outlinedButtonTheme: OutlinedButtonThemeData(
+        // Apply letter spacing to the text style for outlined buttons
+        style: ButtonStyle(
+          textStyle: MaterialStateProperty.all<TextStyle>(
+            const TextStyle(
+              letterSpacing: 0.01,
+            ),
+          ),
+        ),
+      ),
+      textButtonTheme: TextButtonThemeData(
+        // Apply letter spacing to the text style for text buttons
+        style: ButtonStyle(
+          textStyle: MaterialStateProperty.all<TextStyle>(
+            const TextStyle(
+              letterSpacing: 0.01,
+            ),
+          ),
+        ),
+      ),
       // This applies Material 3's color system to produce a palette of
       // appropriately matching and contrasting colors for use in a UI.
       // The Zulip brand color is a starting point, but doesn't end up as
@@ -112,25 +142,24 @@ class ZulipApp extends StatelessWidget {
       tooltipTheme: const TooltipThemeData(preferBelow: false),
     );
 
-    return GlobalStoreWidget(
-      child: Builder(builder: (context) {
-        final globalStore = GlobalStoreWidget.of(context);
-        // TODO(#524) choose initial account as last one used
-        final initialAccountId = globalStore.accounts.firstOrNull?.id;
-        return MaterialApp(
+    return GlobalStoreWidget(child: Builder(builder: (context) {
+      final globalStore = GlobalStoreWidget.of(context);
+      // TODO(#524) choose initial account as last one used
+      final initialAccountId = globalStore.accounts.firstOrNull?.id;
+      return MaterialApp(
           title: 'Zulip',
           localizationsDelegates: ZulipLocalizations.localizationsDelegates,
           supportedLocales: ZulipLocalizations.supportedLocales,
           theme: theme,
-
           navigatorKey: navigatorKey,
           navigatorObservers: navigatorObservers ?? const [],
           builder: (BuildContext context, Widget? child) {
             if (!ready.value) {
-              SchedulerBinding.instance.addPostFrameCallback(
-                (_) => _declareReady());
+              SchedulerBinding.instance
+                  .addPostFrameCallback((_) => _declareReady());
             }
-            GlobalLocalizations.zulipLocalizations = ZulipLocalizations.of(context);
+            GlobalLocalizations.zulipLocalizations =
+                ZulipLocalizations.of(context);
             return child!;
           },
 
@@ -142,7 +171,6 @@ class ZulipApp extends StatelessWidget {
           // handles startup, and then we always push whole routes with methods
           // like [Navigator.push], never mere names as with [Navigator.pushNamed].
           onGenerateRoute: (_) => null,
-
           onGenerateInitialRoutes: (_) {
             return [
               MaterialWidgetRoute(page: const ChooseAccountPage()),
@@ -152,7 +180,7 @@ class ZulipApp extends StatelessWidget {
               ],
             ];
           });
-        }));
+    }));
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/widgets/app.dart
```diff
@@ -94,6 +94,36 @@ class ZulipApp extends StatelessWidget {
 
         shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
       ),
+      elevatedButtonTheme: ElevatedButtonThemeData(
+        // Apply letter spacing to the text style for elevated buttons
+        style: ButtonStyle(
+          textStyle: MaterialStateProperty.all<TextStyle>(
+            const TextStyle(
+              letterSpacing: 0.01,
```
gnprice, line 102: > The letterSpacing  is set to 0.01 times the font size

If you look at the documentation for `TextStyle.letterSpacing`, what does it say about what the value means?

lib/widgets/app.dart
```diff
@@ -112,25 +142,24 @@ class ZulipApp extends StatelessWidget {
...
           theme: theme,
-
           navigatorKey: navigatorKey,
           navigatorObservers: navigatorObservers ?? const [],
           builder: (BuildContext context, Widget? child) {
             if (!ready.value) {
-              SchedulerBinding.instance.addPostFrameCallback(
-                (_) => _declareReady());
+              SchedulerBinding.instance
+                  .addPostFrameCallback((_) => _declareReady());
```
gnprice, line 159: Avoid changing unrelated code. See Zulip's commit style guide:

If your editor made this change automatically, I'd be interested in hearing what editor you use. Then if you can find how to configure it to not do that, I'd be glad for a PR to make that change.

lib/widgets/app.dart
```diff
@@ -94,6 +94,36 @@ class ZulipApp extends StatelessWidget {
 
         shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
       ),
+      elevatedButtonTheme: ElevatedButtonThemeData(
```
gnprice, line 97: This PR covers themes for three types of buttons. Where did you look to try to get a list of the different types of button themes that are needed?

gnprice (commented):
Thanks @Gosplex for the contribution! Comments below.

Before we can fully review this change, it will also need tests. See the README.

gnprice:
(I've also just edited the PR description so that it cross-references the issue. That makes it possible to find one of them from the other.)

lib/widgets/app.dart
gnprice: Cool, I see.

That doc describes a migration that happened at a certain time, and I guess those were probably the only three button themes at that time. But there are more of them now.

Take a look at the documentation for the `ThemeData` class we're constructing here. If you search for "buttontheme" in that documentation (like with Ctrl+F on the [web page](https://api.flutter.dev/flutter/material/ThemeData-class.html) for that doc), you'll find there are several more properties with names like "foo button theme" for some "foo". So I think each of those is one that we'll want to set.

(This probably isn't one of the best-designed parts of Flutter's API — there should probably be a way to set this sort of thing for all of them together when desired. But for the moment we'll work with it as it is.)

gnprice: Because there's a lot of them, we'll also want to avoid duplicating the parts that are the same. (This was already true when there were just three, but it's more so when there are so many.) So please see what you can do in that direction.

lib/widgets/app.dart
Gosplex: @gnprice  Sir
I am using vscode
But I think that was a mistake from my side

lib/widgets/app.dart
Gosplex: @gnprice  Sir
So I have implemented the letterSpacing for all the `button Theme`

Gosplex:
@gnprice Sir
I have updated the pull request 
And I have also created `test file` - `button_letter_spacing_test.dart`.
I am open to any query and changes need.
Gracias

lib/widgets/app.dart
gnprice: I see. We do have that configuration for VS Code:
```
$ cat .vscode/settings.json 
// Visual Studio Code settings for the zulip-flutter repo.
{
  // VS Code formats files on save by default.
  // Like Flutter itself, we don't use the Dart autoformatter,
  // so disable that VS Code behavior.
  "[dart]": {
    // This appears redundant with the all-language setting,
    // but it's not: if the user's personal config has a Dart-specific
    // setting, then we need a Dart-specific setting to override it.
    "editor.formatOnSave": false,
    "editor.formatOnType": false,
    "editor.formatOnPaste": false,
  },

  // For that matter, don't go reformatting whole files in any language.
  "editor.formatOnSave": false,

  // This much more focused automatic fix is helpful, though.
  "files.trimTrailingWhitespace": true,
}
```

so yeah, I think it would not have done this automatically.

In any case you'll need to remove this change from your PR. For that, `git restore -p` will be helpful — see `git help restore`.

gnprice:
Thanks for the revision. Several comments above still apply:
* https://github.com/zulip/zulip-flutter/pull/557#discussion_r1518409854
* https://github.com/zulip/zulip-flutter/pull/557#discussion_r1518411316
* https://github.com/zulip/zulip-flutter/pull/557#discussion_r1518445075

lib/widgets/app.dart
Gosplex: I had used this on my branch
`git restore -p`
##  Output
`No changes.`
What else do you suggest so that I could revert the changes if made

lib/widgets/app.dart
Gosplex: Sorry to as @gnprice  Sir
Could you please provide a reference to this.
I think it would do me a lot good

lib/widgets/app.dart
Gosplex: Please Sir
Throw my more light on this.
What exactly do we stand to achieve and at what point could there be a `duplicate part`
This would enable me know what I am looking for and also finish the task completely.

chrisbobbe: I think Greg is saying that we want to avoid repeating a lot of the same details over and over.

When I read this PR, I see this kind of thing repeated many times:

```dart
        // Apply letter spacing to the text style for […] buttons
        style: ButtonStyle(
          textStyle: MaterialStateProperty.all<TextStyle>(
            const TextStyle(
              letterSpacing: 0.01,
            ),
          ),
        ),
```

This repetition takes many lines of code, so it makes it more difficult to find and read more interesting lines of code.

It also has a maintenance cost. For example, if our designer said he wanted 1.1% spacing instead of 1%, there would be many changes to make instead of just one, and it would be easier to accidentally miss some changes that needed to happen as part of that.

So, for example, can we extract the `ButtonStyle` value out to a new variable and reuse that variable?

Gosplex: Okay 
Quite understandable
Thanks for the clarification.

lib/widgets/app.dart
gnprice: Flutter's API reference docs are at: https://api.flutter.dev/

The search box there should be able to take you to the `TextStyle` class and its `letterSpacing` property. That's the property that's being set by this line of code.

What does the documentation for that property say about what the property's value means?

lib/widgets/app.dart
gnprice: For getting help with tools, Zulip itself is a better medium than GitHub — you'll get faster answers that way, because more people will see the question. See this doc:
https://zulip.com/development-community/

For questions about Git in particular, the `#git help` stream is best.

(You'll need to pass additional arguments to `git restore` in addition to the `-p` flag. See `git help restore`, as I mentioned; if you have questions after looking at that, then ask in `#git help` and we can help you out there.)

gnprice:
Because this PR hasn't been ready for review, I'm closing this PR for now in order to get it out of our queue. Please feel free to re-open it in the future after addressing the feedback above.

Issue #108, opened 2023-05-23
login: Gracefully handle attempts to log into an account that's already logged in


chrisbobbe:
Currently, if you log into an account then go back through "Add an account" with the _same_ account, pressing "Log in" will cause a SQL exception saying that the duplicate record was not inserted. We ignore that exception, except to remove a progress indicator.

We should at least give feedback on that SQL exception. Possibly we could also recognize what's happening before sending off the `fetchApiKey` request and then the database request.

chrisbobbe:
Ah, and to highlight something: we shouldn't misread other SQL errors as being about a duplicate account insert, and we shouldn't forget to handle those other SQL exceptions appropriately. (E.g., disk-full exceptions.)

PR #558, opened 2024-03-08
login: display message when account already logged in

chrisirhc:
Handles #108 , though doesn't address other kinds of SQL errors, I set up some scaffold code to start handling errors that come from the store.

(This is my first PR (in Dart/Flutter/Zulip), so bear with me.)

My initial instinct was to use the Snackbar to signal this issue. (Screenshot in below comment)
However, I saw after that there's a showErrorDialog, so I'm updating it to use it in a separate commit.

The Snackbar is likely a better UX for this scenario since there's little the user can do in a dialog, and there's no input needed from the user. We can also add a action in the snackbar to navigate to the logged in account.

Happy to squash changes once I hear back on some feedback.

Screenshot (using macOS build to debug, blurred out credentials):
Frame 3 (1)

Fixes: #108

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -63,6 +63,18 @@
   "@actionSheetOptionUnstarMessage":  {
     "description": "Label for unstar button on action sheet."
   },
+  "errorAccountLoggedInTitle": "Account already logged in",
+  "@errorAccountLoggedInTitle": {
+    "description": "Error title on attempting to log into an account that's already logged in."
+  },
+  "errorAccountLoggedIn": "The account {email} at {server} is already in your list of accounts.",
+  "@errorAccountLoggedIn": {
+    "description": "Error message on attempting to log into an account that's already logged in.",
+    "placeholders": {
+      "email": {"type": "String", "example": "user@example.com"},
+      "server": {"type": "String", "example": "https://example.com"}
+    }
+  },
   "errorCouldNotFetchMessageSource": "Could not fetch message source",
   "@errorCouldNotFetchMessageSource": {
     "description": "Error message when the source of a message could not be fetched."
--- a/lib/model/database.dart
+++ b/lib/model/database.dart
@@ -2,8 +2,10 @@ import 'dart:io';
 
 import 'package:drift/drift.dart';
 import 'package:drift/native.dart';
+import 'package:drift/remote.dart';
 import 'package:path/path.dart' as path;
 import 'package:path_provider/path_provider.dart';
+import 'package:sqlite3/common.dart';
 
 part 'database.g.dart';
 
@@ -103,7 +105,21 @@ class AppDatabase extends _$AppDatabase {
     );
   }
 
-  Future<int> createAccount(AccountsCompanion values) {
-    return into(accounts).insert(values);
+  Future<int> createAccount(AccountsCompanion values) async {
+    try {
+      return await into(accounts).insert(values);
+    } catch (e) {
+      // Unwrap cause if it's a remote Drift call. On the app, it's running
+      // via a remote, but on local tests, it's running natively so
+      // unwrapping is not required.
+      final cause = (e is DriftRemoteException) ? e.remoteCause : e;
+      if (cause case SqliteException(
+              extendedResultCode: SqlExtendedError.SQLITE_CONSTRAINT_UNIQUE)) {
+        throw AccountAlreadyExistsException();
+      }
+      rethrow;
+    }
   }
 }
+
+class AccountAlreadyExistsException implements Exception {}
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -25,7 +25,7 @@ import 'stream.dart';
 import 'unreads.dart';
 
 export 'package:drift/drift.dart' show Value;
-export 'database.dart' show Account, AccountsCompanion;
+export 'database.dart' show Account, AccountsCompanion, AccountAlreadyExistsException;
 
 /// Store for all the user's data.
 ///
--- a/lib/widgets/login.dart
+++ b/lib/widgets/login.dart
@@ -258,16 +258,31 @@ class _PasswordLoginPageState extends State<PasswordLoginPage> {
     required int userId,
   }) async {
     final globalStore = GlobalStoreWidget.of(context);
-    // TODO(#108): give feedback to user on SQL exception, like dupe realm+user
-    final accountId = await globalStore.insertAccount(AccountsCompanion.insert(
-      realmUrl: widget.serverSettings.realmUrl,
-      email: email,
-      apiKey: apiKey,
-      userId: userId,
-      zulipFeatureLevel: widget.serverSettings.zulipFeatureLevel,
-      zulipVersion: widget.serverSettings.zulipVersion,
-      zulipMergeBase: Value(widget.serverSettings.zulipMergeBase),
-    ));
+    final realmUrl = widget.serverSettings.realmUrl;
+    final int accountId;
+    try {
+      accountId = await globalStore.insertAccount(AccountsCompanion.insert(
+        realmUrl: realmUrl,
+        email: email,
+        apiKey: apiKey,
+        userId: userId,
+        zulipFeatureLevel: widget.serverSettings.zulipFeatureLevel,
+        zulipVersion: widget.serverSettings.zulipVersion,
+        zulipMergeBase: Value(widget.serverSettings.zulipMergeBase),
+      ));
+      // TODO give feedback to user on other SQL exceptions
+    } on AccountAlreadyExistsException {
+      if (!mounted) {
+        return;
+      }
+      final zulipLocalizations = ZulipLocalizations.of(context);
+      showErrorDialog(
+        context: context,
+        title: zulipLocalizations.errorAccountLoggedInTitle,
+        message: zulipLocalizations.errorAccountLoggedIn(
+          email, realmUrl.toString()));
+      return;
+    }
 
     if (!mounted) {
       return;
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -2,11 +2,11 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:checks/checks.dart';
-import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:http/http.dart' as http;
 import 'package:test/scaffolding.dart';
 import 'package:zulip/api/core.dart';
 import 'package:zulip/api/exception.dart';
+import 'package:zulip/model/localizations.dart';
 
 import '../stdlib_checks.dart';
 import 'exception_checks.dart';
@@ -161,7 +161,7 @@ void main() {
           ..which(condition));
     }
 
-    final zulipLocalizations = lookupZulipLocalizations(ZulipLocalizations.supportedLocales.first);
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
     checkRequest(http.ClientException('Oops'), (it) => it.message.equals('Oops'));
     checkRequest(const TlsException('Oops'), (it) => it.message.equals('Oops'));
     checkRequest((foo: 'bar'), (it) => it
--- a/test/model/database_test.dart
+++ b/test/model/database_test.dart
@@ -41,6 +41,54 @@ void main() {
         'acked_push_token': null,
       });
     });
+
+    test('create account with same realm and userId ', () async {
+      final accountData = AccountsCompanion.insert(
+        realmUrl: Uri.parse('https://chat.example/'),
+        userId: 1,
+        email: 'asdf@example.org',
+        apiKey: '1234',
+        zulipVersion: '6.0',
+        zulipMergeBase: const Value('6.0'),
+        zulipFeatureLevel: 42,
+      );
+      final accountDataWithSameUserId = AccountsCompanion.insert(
+        realmUrl: Uri.parse('https://chat.example/'),
+        userId: 1,
+        email: 'otheremail@example.org',
+        apiKey: '12345',
+        zulipVersion: '6.0',
+        zulipMergeBase: const Value('6.0'),
+        zulipFeatureLevel: 42,
+      );
+      await database.createAccount(accountData);
+      await check(database.createAccount(accountDataWithSameUserId))
+        .throws<AccountAlreadyExistsException>();
+    });
+
+    test('create account with same realm and email', () async {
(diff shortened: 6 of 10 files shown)


lib/model/store.dart
```diff
@@ -570,7 +571,21 @@ class LiveGlobalStore extends GlobalStore {
...
   Future<Account> doInsertAccount(AccountsCompanion data) async {
-    final accountId = await _db.createAccount(data); // TODO(log): db errors
+    final int accountId;
+    try {
+      accountId = await _db.createAccount(data); // TODO(log): db errors
+    } catch (e) {
+      switch (e) {
+        case DriftRemoteException(
+            remoteCause: SqliteException(
+              extendedResultCode: 2067 // SQLITE_CONSTRAINT_UNIQUE
```
gnprice, line 581: Interesting. What reference did you use to determine that this was the numeric value to expect for this error?

I think I'd want this to be a named constant somewhere, and then that somewhere can have a comment linking to such a reference.

Ideally `package:sqlite3` would provide all these codes as named constants. Failing that, we can put this one in our `lib/model/database.dart`, and add more codes next to it as we need them. It can look like
```dart
abstract class SqliteResultCode {
  /// The SQLite result code `SQLITE_CONSTRAINT_UNIQUE`.
  const ConstraintUnique = 2067;
}
```
with some dartdoc on the class that includes a link to the relevant reference.

lib/api/exception.dart
```diff
@@ -131,3 +131,5 @@ class MalformedServerResponseException extends ServerException {
             .errorMalformedResponseWithCause(
               httpStatus, causeException.toString()));
 }
+
+class AccountAlreadyExistsException implements Exception {}
```
gnprice, line 135: This `lib/api/` tree is for bindings to the Zulip server API. So this exception should go somewhere else.

Probably `lib/model/store.dart` (where it gets thrown from), at the bottom of the file, is good.

gnprice (commented):
Thanks @chrisirhc for the contribution! This code generally looks good.

Can you post a screenshot for the snackbar approach? That'd help in thinking about which would be best.

Before we can review this change fully, it will need tests. See the section in our README about writing tests.

Meanwhile, a couple of comments below from a first look.

chrisirhc:
> Thanks @chrisirhc for the contribution! This code generally looks good.
> 
> Can you post a screenshot for the snackbar approach? That'd help in thinking about which would be best.
Yep, here:
image

> 
> Before we can review this change fully, it will need tests. See the section in our README about writing tests.
Yep done ✅ 
> 
> Meanwhile, a couple of comments below from a first look.
Yep addressed ✅

lib/model/store.dart
chrisirhc: Hm good point, I imported sqlite3 and used the exported error code constant instead.

test/widgets/login_test.dart
```diff
@@ -113,6 +113,27 @@ void main() {
           id: testBinding.globalStore.accounts.single.id));
     });
 
+    testWidgets('account already exists', (tester) async {
+      final zulipLocalizations = lookupZulipLocalizations(ZulipLocalizations.supportedLocales.first);
```
gnprice, line 117: We have a somewhat more concise pattern we use for this:
```suggestion
      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
```

From a grep, I see there's one existing place in the tests where we do the more verbose thing, in `test/api/core_test.dart`. From `git log -p -S Localizations test/` I see that's the very first time we wrote a test that compared to a localized string. A separate small, [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc), commit to just update that one spot would be welcome to add to this PR.

gnprice, line 117: Separately, let's put this line just before where we use it. It logically fits as part of checking the end state, vs. preparing the state.

assets/l10n/app_en.arb
```diff
@@ -63,6 +63,18 @@
   "@actionSheetOptionUnstarMessage":  {
     "description": "Label for unstar button on action sheet."
   },
+  "errorAccountLoggedInTitle": "Account already logged in",
+  "@errorAccountLoggedInTitle": {
+    "description": "Error message when the account is already logged in."
+  },
+  "errorAccountLoggedIn": "Account logged in.\nServer: {server}\nUser: {email}",
+  "@errorAccountLoggedIn": {
+    "description": "Error message when the account is already logged in.",
```
gnprice, line 72:
```suggestion
    "description": "Error message on attempting to log into an account that's already logged in.",
```

assets/l10n/app_en.arb
```diff
@@ -63,6 +63,18 @@
   "@actionSheetOptionUnstarMessage":  {
     "description": "Label for unstar button on action sheet."
   },
+  "errorAccountLoggedInTitle": "Account already logged in",
+  "@errorAccountLoggedInTitle": {
+    "description": "Error message when the account is already logged in."
```
gnprice, line 68: Same rewording as for the other description, but also "Error title" rather than "Error message".

assets/l10n/app_en.arb
```diff
@@ -63,6 +63,18 @@
...
+  },
+  "errorAccountLoggedIn": "Account logged in.\nServer: {server}\nUser: {email}",
```
gnprice, line 70: Here's a rewording:
```suggestion
  "errorAccountLoggedIn": "The account {email} at {url} is already in your list of accounts.",
```

One thing is that the terse headlinese version is already covered by the title; here in the body of the dialog box, we have a bit more room and can use complete sentences.

gnprice (commented):
Thanks @chrisirhc for the revision! Generally this looks great; various small comments below.

Thanks for posting the snackbar version. I think in this setting the modal dialog box is actually the more appropriate UX: the thing the user tried to do did not work and isn't going to work, and a modal error screen seems like a fitting reflection of that. A snackbar (which disappears after a few seconds) feels more appropriate for an FYI sort of situation where the user doesn't necessarily need to pay attention to the error, perhaps because it might go away on its own.

One thing this will need before we can merge it, apart from the various small changes described below, is to organize the branch as a series of clear and coherent commits following [Zulip's commit style guide](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html). For this PR, I think that looks like:
* a small commit adding that direct dependency;
* a commit defining and throwing `AccountAlreadyExistsException`;
* a commit adding the UI to react to that exception, along with its localizations and tests.

chrisirhc:
Thanks for the detailed review and comments.
Apologies, I missed the commit guide earlier and some of the conventions. Rebased the commits, edited the commit messages, as well as the nit comments.

lib/model/store.dart
```diff
@@ -570,7 +572,18 @@ class LiveGlobalStore extends GlobalStore {
...
   Future<Account> doInsertAccount(AccountsCompanion data) async {
-    final accountId = await _db.createAccount(data); // TODO(log): db errors
+    final int accountId;
+    try {
+      accountId = await _db.createAccount(data); // TODO(log): db errors
+    } catch (e) {
+      if (e case DriftRemoteException(
+            remoteCause: SqliteException(
+              extendedResultCode: SqlExtendedError.SQLITE_CONSTRAINT_UNIQUE))) {
+          throw AccountAlreadyExistsException();
```
gnprice, line 582: Rereading this, there are enough details here that I'm realizing we should definitely have a test for it. Or I guess two tests: one for duplicate realmUrl and email, and one for duplicate realmUrl and userId.

That also causes me to notice that the best home for this logic would be in our `database.dart`, inside this `AppDatabase.createAccount` method. That way these rather database-specific details go in our database layer.

So then the tests can be: create one account, create another that should collide with it, check that the second one throws an AccountAlreadyExistsException.

test/api/core_test.dart
```diff
@@ -161,7 +161,7 @@ void main() {
           ..which(condition));
     }
 
-    final zulipLocalizations = lookupZulipLocalizations(ZulipLocalizations.supportedLocales.first);
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
```
gnprice, line 164: nit in commit message:

    store [nfc]: Refactor remaining inconsistent localization use in test

Prefix should be "test [nfc]:" — this commit isn't related to the store, but it is in tests. Then the "in test" at the end isn't needed, either.

(Alternatively it could be "api test [nfc]:" to indicate which area's tests it's about. But this particular change is really about the tests in general, and it just happens that there's only one example of this pattern.)

lib/model/store.dart
gnprice, line 582: nit in commit message:

    store: Throw exception when account already logged in
    
    When account is already logged in, throw AccountAlreadyExists to allow caller
    to catch and handle exception.

Line too long in the prose; see https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#formatting-guidelines .

gnprice (commented):
Thanks @chrisirhc for the revision! Generally this all looks great, including the new commit structure; a few nits below.

There's also one substantive point I didn't think of before, which is that this logic that inspects the database exception should have a test.

No worries re not seeing the commit style guide earlier. There's a lot to absorb (in this and really any significant codebase), so it's normal that one's first PR or two has some things you've missed; what matters more is your trajectory as you learn the codebase and pick up our style.

test/model/test_store.dart
```diff
@@ -76,6 +77,16 @@ class TestGlobalStore extends GlobalStore {
 
   @override
   Future<Account> doInsertAccount(AccountsCompanion data) async {
+    // Check for duplication is typically handled by the database but since
+    // we're not using a real database, this needs to be handled here.
+    // See [AppDatabase.createAccount].
```
chrisirhc, line 82: There's now a little bit of asymmetry in that the `TestGlobalStore` throws `AccountAlreadyExistsException` whereas its thrown by the database in real usage, so I added a comment to call that out.

lib/model/database.dart
```diff
@@ -103,7 +104,17 @@ class AppDatabase extends _$AppDatabase {
...
   }
 
-  Future<int> createAccount(AccountsCompanion values) {
-    return into(accounts).insert(values);
+  Future<int> createAccount(AccountsCompanion values) async {
+    try {
+      return await into(accounts).insert(values);
+    } catch (e) {
+      if (e case SqliteException(
+          extendedResultCode: SqlExtendedError.SQLITE_CONSTRAINT_UNIQUE)) {
```
gnprice, line 112: Hmm — this differs from the previous revision in that it looks for a `SqliteException` directly, rather than a `DriftRemoteException` that contains a `SqliteException`. Does this version work when you try it manually in the actual app?

Trying it just now, indeed it doesn't — the exception just propagates uncaught, the same behavior as before the PR.

OTOH this version is what we get in the test environment. It's awkward that the test environment differs from the live app in that way, but I think we can still make it work so that the tests have most of the value they should have:
```suggestion
    } catch (error) {
      final e = (error is DriftRemoteException) ? error.remoteCause : error;
      if (e case SqliteException(
          extendedResultCode: SqlExtendedError.SQLITE_CONSTRAINT_UNIQUE)) {
```

That way just that one line behaves differently in tests vs. the live app, and the rest of the error handling should behave the same.

gnprice, line 112: nit:
```suggestion
      if (e case SqliteException(
              extendedResultCode: SqlExtendedError.SQLITE_CONSTRAINT_UNIQUE)) {
```

test/model/database_test.dart
```diff
@@ -41,6 +41,60 @@ void main() {
...
+        zulipVersion: '6.0',
+        zulipMergeBase: const Value('6.0'),
+        zulipFeatureLevel: 42,
+      );
+      await database.createAccount(accountData);
+      try  {
+        await database.createAccount(accountDataWithSameUserId);
+      } catch (e) {
+        check(e).isA<AccountAlreadyExistsException>();
+      }
```
gnprice, line 69: Handy `check` shortcut for this:
```suggestion
      await check(database.createAccount(accountDataWithSameUserId))
        .throws<AccountAlreadyExistsException>();
```

lib/widgets/login.dart
```diff
@@ -5,6 +5,7 @@ import '../api/exception.dart';
 import '../api/route/account.dart';
 import '../api/route/realm.dart';
 import '../api/route/users.dart';
+import '../model/database.dart';
 import '../model/store.dart';
```
gnprice, line 9: It feels a bit unsatisfying to have to say that this file imports our database library, as that should be a layer it doesn't interact with directly.

And there is a handy Dart feature we can use to avoid that, which in fact we already use for a couple of related types. We can say:
```diff
--- lib/model/store.dart
+++ lib/model/store.dart
@@ -29,2 +27,2 @@ import 'stream.dart';
 export 'package:drift/drift.dart' show Value;
-export 'database.dart' show Account, AccountsCompanion;
+export 'database.dart' show Account, AccountsCompanion, AccountAlreadyExistsException;
```

Then this `model/database.dart` import becomes happily redundant.

gnprice (commented):
Thanks @chrisirhc for the revision! This code generally looks good, but this version no longer actually works :slightly_smiling_face: — see below.

Also a few small comments. If it weren't for the part where this version doesn't work, I'd probably go ahead and merge and fix those up myself; but since you're doing another revision anyway, please fix those too.

One commit-message nit: in this message:

    login: Display message when account already logged in
    
    This provides some user feedback since the previous behavior looked as if
    nothing happened.

the prose is too wide, like in https://github.com/zulip/zulip-flutter/pull/558#discussion_r1520298368 . The max width should be around 68 columns, and no more than 70.

gnprice:
Oh, one other thought: please go ahead and mark this (both in the PR description, and the commit message of the main commit) as fixing #108. (See [style guide here](https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#mention-fixed-issue) for details.) The case where the user is already logged into the account is the primary thing #108 is about; we can file a separate followup issue for handling other SQL errors like a full disk.

(We'll eventually have other SQL queries that modify things too, not just this one, and errors like a full disk will be cross-cutting across all of those. So the best time to write that handling might in any case be after we have a couple more such queries and a couple more call sites for them, the better to think about what the API should look like at those call sites.)

lib/model/database.dart
chrisirhc: Ack, added a comment to call this out.

chrisirhc:
Yep done. Rebased and verified.

chrisbobbe:
Thanks for your work on this! 🙂 Looks good; merging. (Greg is on vacation through Monday and asked me to re-review.)

gnprice:
Looks great. Thanks all!

PR #559, opened 2024-03-09
[Draft] login: ability to remove logged in account

chrisirhc:


gnprice:
Thanks @chrisirhc for sending this PR (back in March)!

This feature turns out to be a fairly tricky one, when it comes to shutting down any PerAccountStore instances that we may have for the now-logged-account (or that we might have been in the process of setting up). See #948 for how we've ended up handling those.

Closing this PR in favor of that later one.

Issue #391, opened 2023-11-18
inbox: On collapsing a section from its sticky header, auto-scroll as needed to put the collapsed header in view


chrisbobbe:
In the inbox, if you've scrolled down several screenfuls through a section (either the all-DMs section or a stream section), you can still collapse the section by tapping the collapse icon on the stream header.

But when you do that, something confusing happens: everything that was below the last item in the section (which may or may not be onscreen) gets shifted upward by the whole vertical space that used to be occupied by the section. Since you'd been scrolling so much, the shift ends up putting some content offscreen in the upward direction. That's confusing because you might have been interested in that content, and skipping over it wasn't part of your intention when you collapsed a section that was unrelated to the content.

In this screen recording, the scenario happens multiple times in a row, with `#issues`, `#design`, and then `#development help` taking its turn to be the section that's partly scrolled through, and getting its sticky header tapped to collapse it.

![Nov-17-2023 23-52-59](https://github.com/zulip/zulip-flutter/assets/22248748/1e573dc8-6cc3-40a5-9e9d-60aaeea585c3)

I think the confusion would disappear if, when you press the sticky-header's collapse icon, we ensure that the collapsed header (the one that always moves, not the sticky one) is entirely visible on the screen, by changing the scroll position. We can try using something like [`Scrollable.ensureVisible`](https://api.flutter.dev/flutter/widgets/Scrollable/ensureVisible.html) for that. That way you can see that the section was collapsed, as intended, and the start of the next section is immediately below it and easy to start reading.

sm-sayedi:
I would like to work on this issue. 

So far, I have partially solved it as below:

<img src="https://github.com/zulip/zulip-flutter/assets/59946442/c09b17cf-8311-4a0a-9bc4-47cdc9d901d7" width="300" />

gnprice:
@sm-sayedi Sure — please send along your code for this as a PR and we'll be glad to take a look. From what I see in your screencast, the behavior looks like what we want.

PR #560, opened 2024-03-11
inbox: Fix collapsing a section from its sticky header

sm-sayedi:
Collapsing a stream/all-DMs section through a sticky header behaves as expected without shifting other sections off the screen.

Fixes: zulip#391.

### DIFF
--- a/lib/widgets/inbox.dart
+++ b/lib/widgets/inbox.dart
@@ -209,11 +209,19 @@ abstract class _HeaderItem extends StatelessWidget {
   final int count;
   final bool hasMention;
 
+  /// A build context within the [_StreamSection] or [_AllDmsSection].
+  ///
+  /// Used to ensure the [_StreamSection] or [_AllDmsSection] that encloses the
+  /// current [_HeaderItem] is visible after being collapsed through this
+  /// [_HeaderItem].
+  final BuildContext sectionContext;
+
   const _HeaderItem({
     required this.collapsed,
     required this.pageState,
     required this.count,
     required this.hasMention,
+    required this.sectionContext,
   });
 
   String get title;
@@ -223,7 +231,15 @@ abstract class _HeaderItem extends StatelessWidget {
   Color get uncollapsedBackgroundColor;
   Color? get unreadCountBadgeBackgroundColor;
 
-  void Function() get onCollapseButtonTap;
+  Future<void> Function() get onCollapseButtonTap => () async {
+    if (!collapsed) {
+      await Scrollable.ensureVisible(
+        sectionContext,
+        alignmentPolicy: ScrollPositionAlignmentPolicy.keepVisibleAtStart,
+      );
+    }
+  };
+
   void Function() get onRowTap;
 
   @override
@@ -271,6 +287,7 @@ class _AllDmsHeaderItem extends _HeaderItem {
     required super.pageState,
     required super.count,
     required super.hasMention,
+    required super.sectionContext,
   });
 
   @override get title => 'Direct messages'; // TODO(i18n)
@@ -280,7 +297,8 @@ class _AllDmsHeaderItem extends _HeaderItem {
   @override get uncollapsedBackgroundColor => const Color(0xFFF3F0E7);
   @override get unreadCountBadgeBackgroundColor => null;
 
-  @override get onCollapseButtonTap => () {
+  @override get onCollapseButtonTap => () async {
+    await super.onCollapseButtonTap();
     pageState.allDmsCollapsed = !collapsed;
   };
   @override get onRowTap => onCollapseButtonTap; // TODO open all-DMs narrow?
@@ -304,6 +322,7 @@ class _AllDmsSection extends StatelessWidget {
       hasMention: data.hasMention,
       collapsed: collapsed,
       pageState: pageState,
+      sectionContext: context,
     );
     return StickyHeaderItem(
       header: header,
@@ -386,6 +405,7 @@ class _StreamHeaderItem extends _HeaderItem {
     required super.pageState,
     required super.count,
     required super.hasMention,
+    required super.sectionContext,
   });
 
   @override get title => subscription.name;
@@ -397,7 +417,8 @@ class _StreamHeaderItem extends _HeaderItem {
   @override get unreadCountBadgeBackgroundColor =>
     subscription.colorSwatch().unreadCountBadgeBackground;
 
-  @override get onCollapseButtonTap => () {
+  @override get onCollapseButtonTap => () async {
+    await super.onCollapseButtonTap();
     if (collapsed) {
       pageState.uncollapseStream(subscription.streamId);
     } else {
@@ -427,6 +448,7 @@ class _StreamSection extends StatelessWidget {
       hasMention: data.hasMention,
       collapsed: collapsed,
       pageState: pageState,
+      sectionContext: context,
     );
     return StickyHeaderItem(
       header: header,
--- a/test/widgets/inbox_test.dart
+++ b/test/widgets/inbox_test.dart
@@ -13,6 +13,38 @@ import '../flutter_checks.dart';
 import '../model/binding.dart';
 import '../model/test_store.dart';
 
+/// Repeatedly drags `view` by `moveStep` until `finder` is invisible.
+///
+/// Between each drag, advances the clock by `duration`.
+///
+/// Throws a [StateError] if `finder` is still visible after `maxIteration`
+/// drags.
+///
+/// See also:
+///  * [WidgetController.dragUntilVisible], which does the inverse.
+Future<void> dragUntilInvisible(
+  WidgetTester tester,
+  FinderBase<Element> finder,
+  FinderBase<Element> view,
+  Offset moveStep, {
+  int maxIteration = 50,
+  Duration duration = const Duration(milliseconds: 50),
+}) {
+  return TestAsyncUtils.guard<void>(() async {
+    final iteration = maxIteration;
+    while (maxIteration > 0 && finder.evaluate().isNotEmpty) {
+      await tester.drag(view, moveStep);
+      await tester.pump(duration);
+      maxIteration -= 1;
+    }
+    if (maxIteration <= 0 && finder.evaluate().isNotEmpty) {
+      throw StateError(
+        'Finder is still visible after $iteration iterations.'
+        ' Consider increasing the number of iterations.');
+    }
+  });
+}
+
 void main() {
   TestZulipBinding.ensureInitialized();
 
@@ -51,6 +83,15 @@ void main() {
     await tester.pumpAndSettle();
   }
 
+  List<StreamMessage> generateStreamMessages({
+    required ZulipStream stream,
+    required int count,
+    required List<MessageFlag> flags,
+  }) {
+    return List.generate(count, (index) => eg.streamMessage(
+      stream: stream, topic: '${stream.name} topic $index', flags: flags));
+  }
+
   /// Set up an inbox view with lots of interesting content.
   Future<void> setupVarious(WidgetTester tester) async {
     final stream1 = eg.stream(streamId: 1, name: 'stream 1');
@@ -61,12 +102,16 @@ void main() {
     await setupPage(tester,
       streams: [stream1, stream2],
       subscriptions: [sub1, sub2],
-      users: [eg.selfUser, eg.otherUser, eg.thirdUser],
+      users: [eg.selfUser, eg.otherUser, eg.thirdUser, eg.fourthUser],
       unreadMessages: [
         eg.streamMessage(stream: stream1, topic: 'specific topic', flags: []),
+        ...generateStreamMessages(stream: stream1, count: 10, flags: []),
         eg.streamMessage(stream: stream2, flags: []),
+        ...generateStreamMessages(stream: stream2, count: 40, flags: []),
         eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], flags: []),
         eg.dmMessage(from: eg.otherUser, to: [eg.selfUser, eg.thirdUser], flags: []),
+        eg.dmMessage(from: eg.thirdUser, to: [eg.selfUser], flags: []),
+        eg.dmMessage(from: eg.fourthUser, to: [eg.selfUser], flags: []),
       ]);
   }
 
@@ -310,6 +355,28 @@ void main() {
           checkAppearsUncollapsed(tester, findSectionContent);
         });
 
+        testWidgets('collapse all-DMs section when partially offscreen: '
+          'header remains sticky at top', (tester) async {
+          await setupVarious(tester);
+
+          final listFinder = find.byType(Scrollable);
+          final dmFinder = find.text(eg.otherUser.fullName).hitTestable();
+
+          // Scroll part of [_AllDmsSection] offscreen.
+          await dragUntilInvisible(
+            tester, dmFinder, listFinder, const Offset(0, -50));
+
+          // Check that the header is present (which must therefore
+          // be as a sticky header).
+          check(findAllDmsHeaderRow(tester)).isNotNull();
+
+          await tapCollapseIcon(tester);
+
+          // Check that the header is still visible even after
+          // collapsing the section.
+          check(findAllDmsHeaderRow(tester)).isNotNull();
+        });
+
         // TODO check it remains collapsed even if you scroll far away and back
 
         // TODO check that it's always uncollapsed when it appears after being
@@ -383,6 +450,52 @@ void main() {
           checkAppearsCollapsed(tester, 1, findSectionContent);
           await tapCollapseIcon(tester, 1);
           checkAppearsUncollapsed(tester, 1, findSectionContent);
+        });
+
+        testWidgets('collapse stream section when partially offscreen: '
+          'header remains sticky at top', (tester) async {
+          await setupVarious(tester);
+
+          final topicFinder = find.text('stream 1 topic 4').hitTestable();
+          final listFinder = find.byType(Scrollable);
+
+          // Scroll part of [_StreamSection] offscreen.
+          await dragUntilInvisible(
+            tester, topicFinder, listFinder, const Offset(0, -50));
+
+          // Check that the header is present (which must therefore
+          // be as a sticky header).
+          check(findStreamHeaderRow(tester, 1)).isNotNull();
+
+          await tapCollapseIcon(tester, 1);
+
+          // Check that the header is still visible even after
+          // collapsing the section.
+          check(findStreamHeaderRow(tester, 1)).isNotNull();
+        });
+
+        testWidgets('collapse stream section in middle of screen: '
+          'header stays fixed', (tester) async {
+          await setupVarious(tester);
+
+          final headerRow = findStreamHeaderRow(tester, 1);
+          // Check that the header is present.
+          check(headerRow).isNotNull();
+
+          final rectBeforeTap = tester.getRect(find.byWidget(headerRow!));
+          final scrollableTop = tester.getRect(find.byType(Scrollable)).top;
+          // Check that the header is somewhere in the middle of the screen.
+          check(rectBeforeTap.top).isGreaterThan(scrollableTop);
+
+          await tapCollapseIcon(tester, 1);
+
+          final headerRowAfterTap = findStreamHeaderRow(tester, 1);
+          final rectAfterTap =
+            tester.getRect(find.byWidget(headerRowAfterTap!));
+
+          // Check that the position of the header before and after
+          // collapsing is the same.
+          check(rectAfterTap).equals(rectBeforeTap);
         });
 
         // TODO check it remains collapsed even if you scroll far away and back
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -119,6 +119,8 @@ final Account otherAccount = account(
 
 final User thirdUser = user(fullName: 'Third User', email: 'third@example');
 
+final User fourthUser  = user(fullName: 'Fourth User', email: 'fourth@example');
+
 ////////////////////////////////////////////////////////////////
 // Streams and subscriptions.
 //


lib/widgets/inbox.dart
```diff
@@ -377,17 +380,44 @@ class _DmItem extends StatelessWidget {
...
+  /// through this [StreamHeaderItem] being used as a sticky header.
+  ///
+  /// A non-null value should be passed if this [StreamHeaderItem]
+  /// is used as a sticky header.
+  final BuildContext? parentContext;
+
+  const StreamHeaderItem({
     required this.subscription,
     required super.collapsed,
     required super.pageState,
```
sm-sayedi, line 397: The preceding change has led to showing a lint warning for line no. 397 saying: `Invalid use of a private type in a public API.
Try making the private type public, or making the API that uses the private type also be private`.
I haven't touched it anymore so to get your feedback about it.

lib/widgets/inbox.dart
```diff
@@ -377,17 +380,44 @@ class _DmItem extends StatelessWidget {
...
   final Subscription subscription;
 
-  const _StreamHeaderItem({
+  /// This context is used to ensure the [_StreamSection] that encloses
+  /// the current [StreamHeaderItem] is visible after being collpased
+  /// through this [StreamHeaderItem] being used as a sticky header.
+  ///
+  /// A non-null value should be passed if this [StreamHeaderItem]
+  /// is used as a sticky header.
+  final BuildContext? parentContext;
```
gnprice, line 392: This is a concern that's common to both kinds of sections and their headers, so let's put it on the base class `_HeaderItem`.

Let's also call it `sectionContext`, rather than "parent". The exact sequence of ancestors is rather an implementation detail that can change over time… and in fact I think already in this version the context this points to isn't actually the direct "parent" of this widget in any sense. Better to have the name reflect the thing we actually need to be true of this context, which is that it's the one for the whole section this header corresponds to.

test/widgets/inbox_test.dart
```diff
@@ -51,25 +51,46 @@ void main() {
     await tester.pumpAndSettle();
   }
 
+  List<StreamMessage> generateStreamMessages(
+    ZulipStream stream,
+    int count,
+  ) =>
+      List.generate(
```
gnprice, line 58: nit: much too long for `=>`  syntax — use braces:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#consider-using-for-short-functions-and-methods

test/widgets/inbox_test.dart
```diff
@@ -383,6 +452,13 @@ void main() {
           checkAppearsCollapsed(tester, 1, findSectionContent);
           await tapCollapseIcon(tester, 1);
           checkAppearsUncollapsed(tester, 1, findSectionContent);
+
+          // check if collapsing stream through sticky header behaves
+          // as expected after scroll
+          await checkCollapseAfterScroll(tester, 2, true);
+          // check if collapsing stream through normal header behaves
+          // as expected after scroll
+          await checkCollapseAfterScroll(tester, 3, false);
```
gnprice, line 461: This should go in a new test, not as additional checks at the end of an existing test.

That's important for two reasons:
* It means that if some change breaks the other functionality that this test checks, or breaks this functionality, or breaks both of them, the test results clearly indicate which of those scenarios it is.
* It makes it possible to read and understand this test without first thinking through all the unrelated manipulation above it which is there for the sake of the existing test.

test/widgets/inbox_test.dart
```diff
@@ -51,25 +51,46 @@ void main() {
...
   /// Set up an inbox view with lots of interesting content.
   Future<void> setupVarious(WidgetTester tester) async {
     final stream1 = eg.stream(streamId: 1, name: 'stream 1');
     final sub1 = eg.subscription(stream1);
     final stream2 = eg.stream(streamId: 2, name: 'stream 2');
     final sub2 = eg.subscription(stream2);
+    final stream3 = eg.stream(streamId: 3, name: 'stream 3');
+    final sub3 = eg.subscription(stream3);
+    final stream4 = eg.stream(streamId: 4, name: 'stream 4');
+    final sub4 = eg.subscription(stream4);
```
gnprice, line 73: Similarly, all this should go into setup for the one new test, not be added implicitly to all the existing tests. That way we avoid making the existing tests more complicated to understand.

(Also the new test probably doesn't need 4 different streams, so it can probably be simpler than this too.)

Quite possibly `setupVarious` is a mistake already for this reason, but we can avoid making it more so.

test/widgets/inbox_test.dart
```diff
@@ -104,6 +125,24 @@ void main() {
...
+    required bool isSticky,
+  }) {
+    final stream = store.streams[streamId]!;
+    return find.ancestor(
+      of: find.text(stream.name),
+      matching: find.byWidgetPredicate((widget) =>
+          widget is StreamHeaderItem &&
+          (isSticky
+              ? widget.parentContext != null
+              : widget.parentContext == null)),
```
gnprice, line 142: From the user's perspective, there aren't two kinds of header "sticky" and "not sticky" — there's just one header, which has certain behavior about sticking at the top of the viewport when its section is scrolling past.

For these UI tests, in general we aim to write them in a way that reflects the UI as the user sees it, keeping to a minimum the references to our own UI classes and their properties etc. One major reason for that is to avoid making the tests brittle — we want to be able to rearrange and refactor the internals of the code, and if that doesn't change any behavior the user sees then the tests shouldn't care. If we do have to go update a bunch of tests in order to make such a change, that's a drag.

I think it should be possible to write the tests for this new functionality the same way as we've written the other tests in this file, using `findStreamHeaderRow`. If you're running into obstacles doing that, please start a thread in `#mobile-team` on chat.zulip.org and we'll be glad to discuss.

lib/widgets/inbox.dart
```diff
@@ -377,17 +380,44 @@ class _DmItem extends StatelessWidget {
   }
 }
 
-class _StreamHeaderItem extends _HeaderItem {
+class StreamHeaderItem extends _HeaderItem {
```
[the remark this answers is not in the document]
gnprice: Cool, thanks for flagging it. As described in another comment, these tests ideally wouldn't care about the details of this class, and I think we can write them in a way that doesn't.

If we did make this public, we'd need to give it a more specific name: like `InboxStreamHeaderItem`. (In particular when it's public it'll appear in IDE auto-completion across the whole codebase.)

gnprice (commented):
Very interesting, thanks @sm-sayedi!

The behavior here looks exactly right for stream sections. As #391 says, we'll also want the same behavior for the DMs section.

See other comments below about ways of organizing this logic and of writing the tests.

lib/widgets/inbox.dart
```diff
@@ -421,15 +454,16 @@ class _StreamSection extends StatelessWidget {
...
     final subscription = PerAccountStoreWidget.of(context).subscriptions[data.streamId]!;
-    final header = _StreamHeaderItem(
+    final header = StreamHeaderItem(
       subscription: subscription,
       count: data.count,
       hasMention: data.hasMention,
       collapsed: collapsed,
       pageState: pageState,
     );
+    final stickyHeader = header.copyWith(parentContext: context);
```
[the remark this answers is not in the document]
sm-sayedi: Yeah, sure. It seems redundant to have two headers, but I couldn't find a better way. If we use one header as follows:

```dart
final header = StreamHeaderItem(
      subscription: subscription,
      count: data.count,
      hasMention: data.hasMention,
      collapsed: collapsed,
      pageState: pageState,
      sectionContext: context,
    );
    return StickyHeaderItem(
      header: header,
      child: Column(children: [
        header,
        if (!collapsed) ...data.items.map((item) {
        
```
Then it will work fine only on sticky headers. Collapsing a section with its header in the middle of the screen will also jump to the top of the screen as both are the same instances of the header.



So to auto-scroll to the top by only by tapping the sticky header, there should be two instances of the header, with the `context` only passed to the sticky one.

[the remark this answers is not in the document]
gnprice: That's the behavior that should be controlled by the `alignmentPolicy` I mentioned. What do its docs say?

lib/widgets/inbox.dart
```diff
@@ -38,7 +38,10 @@ class _InboxPageState extends State<InboxPage> with PerAccountStoreAwareStateMix
 
   get collapsedStreamIds => _collapsedStreamIds;
   final Set<int> _collapsedStreamIds = {};
-  void collapseStream(int streamId) {
+  void collapseStream(int streamId, {BuildContext? streamContext}) async {
+    if (streamContext != null) {
+      await Scrollable.ensureVisible(streamContext);
```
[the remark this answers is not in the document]
sm-sayedi: Done.

lib/widgets/inbox.dart
```diff
@@ -401,7 +431,10 @@ class _StreamHeaderItem extends _HeaderItem {
     if (collapsed) {
       pageState.uncollapseStream(subscription.streamId);
     } else {
-      pageState.collapseStream(subscription.streamId);
+          pageState.collapseStream(
```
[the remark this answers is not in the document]
sm-sayedi: Done.

lib/widgets/inbox.dart
sm-sayedi: Done.

test/widgets/inbox_test.dart
sm-sayedi: Done.

lib/widgets/inbox.dart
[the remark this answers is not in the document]
sm-sayedi: Ok thanks, got it!

test/widgets/inbox_test.dart
sm-sayedi: I am having a hard time writing tests for the new implementation, as I cannot differentiate between a header being sticky or normal. So I will start a thread in `#mobile-team`.

sm-sayedi: With the new implementation, I am having a hard time writing tests for this feature. I am not even sure what to test.
In the previous implementation with the `BuildContext` only passed to the sticky headers, I could check for the value of the `sectionContext` for nullability, so a sticky header before collapsing would have a non-null `context` and after collapsing I would check for the presence of the same header with a non-null value. The same for a normal header having a null `context` and present on the screen. Now the `sectionContext` is a non-null value all the time, so it's hard to differentiate between the two states (types) of headers.

Or as you said before, should I share this on a thread in `#mobile-team`?!

gnprice: Yeah, a chat thread would be the best medium for this.

sm-sayedi:
I have updated my PR as per your review. I have also updated the tests for it. Please have a look. Thanks!

test/widgets/inbox_test.dart
```diff
@@ -64,7 +79,9 @@ void main() {
       users: [eg.selfUser, eg.otherUser, eg.thirdUser],
       unreadMessages: [
         eg.streamMessage(stream: stream1, topic: 'specific topic', flags: []),
+        ...generateStreamMessages(stream1, 10),
         eg.streamMessage(stream: stream2, flags: []),
+        ...generateStreamMessages(stream2, 40),
         eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], flags: []),
         eg.dmMessage(from: eg.otherUser, to: [eg.selfUser, eg.thirdUser], flags: []),
       ]);
```
sm-sayedi, line 87: I still used the `setupVarious` method for the new tests as I figured out there was no need for two additional streams. I just added additional messages to the existing streams. If you still believe it has made it complicated, please let me know so that I set them up separately for the new tests!

test/widgets/inbox_test.dart
```diff
@@ -310,6 +327,30 @@ void main() {
...
+          final findSectionContent = find.text(eg.otherUser.fullName);
+
+          checkAppearsUncollapsed(tester, findSectionContent);
+          await tapCollapseIcon(tester);
+          // Check that the header is still visible even after
+          // collapsing the section
+          check(headerRow).isNotNull();
+          checkAppearsCollapsed(tester, findSectionContent);
+          await tapCollapseIcon(tester);
+          checkAppearsUncollapsed(tester, findSectionContent);
```
gnprice, line 351: This test logic seems not to function as intended. As a result I'm not confident that this version would successfully accomplish the test's purpose, which is to fail if at some point in the future we make a change that accidentally reverts this feature.

Specifically, a test I like to do on any test for a new feature or bugfix is to run the test against the old code, from before the change. Here, that looks like this:
```
$ git checkout origin lib/
$ flutter test test/widgets/inbox_test.dart
00:03 +11: InboxPage collapsing all-DMs section collapse all-DMs section after scroll                 
The following assertion was thrown running a test:
Finder specifies a widget that would not receive pointer events.
A call to tap() with finder "Found 1 widget with the given widget (Icon(IconData(U+0F102), size:
20.0, color: Color(0x7f1d2e48))): [
  Icon(IconData(U+0F102), size: 20.0, color: Color(0x7f1d2e48), dependencies: [Directionality,
IconTheme]),
]" derived an Offset (Offset(20.0, 46.0)) that would not hit test on the specified widget.
Maybe the widget is actually off-screen, or another widget is obscuring it, or the widget cannot
receive pointer events.
The finder corresponds to this RenderBox: RenderSemanticsAnnotations#e51b6 relayoutBoundary=up11(creator: Semantics ← Icon ← Padding ← Row ← Listener ← RawGestureDetector ← GestureDetector ← Semantics ← DefaultSelectionStyle ← Builder ← MouseRegion ← Semantics ← ⋯, parentData: offset=Offset(10.0, 10.0) (can use size), constraints: BoxConstraints(unconstrained), size: Size(20.0, 20.0))
The hit test result at that offset is: HitTestResult(_RenderInkFeatures#10262@Offset(20.0, 46.0),
RenderPhysicalModel#5f7e5@Offset(20.0, 46.0),
RenderAnnotatedRegion<SystemUiOverlayStyle>#7e1d9@Offset(20.0, 46.0),
RenderSemanticsAnnotations#f8204@Offset(20.0, 46.0), RenderConstrainedBox#d3925@Offset(20.0, 46.0),
RenderCustomMultiChildLayoutBox#c4b2b@Offset(20.0, 46.0), _RenderInkFeatures#7f830@Offset(20.0,
46.0), RenderPhysicalModel#a28d4@Offset(20.0, 46.0), RenderSemanticsAnnotations#1e887@Offset(20.0,
46.0), RenderRepaintBoundary#f1955@Offset(20.0, 46.0), RenderIgnorePointer#7d6b4@Offset(20.0, 46.0),
_RenderSnapshotWidget#8b907@Offset(20.0, 46.0), _RenderSnapshotWidget#ed2f3@Offset(20.0, 46.0),
_RenderSnapshotWidget#68834@Offset(20.0, 46.0), _RenderSnapshotWidget#0b54b@Offset(20.0, 46.0),
RenderRepaintBoundary#20409@Offset(20.0, 46.0), RenderSemanticsAnnotations#ae43e@Offset(20.0, 46.0),
RenderOffstage#141a0@Offset(20.0, 46.0), RenderSemanticsAnnotations#7000f@Offset(20.0, 46.0),
_RenderTheater#5ef0d@Offset(20.0, 46.0), RenderAbsorbPointer#a8ea2@Offset(20.0, 46.0),
RenderPointerListener#9d330@Offset(20.0, 46.0), RenderSemanticsAnnotations#a580a@Offset(20.0, 46.0),
RenderCustomPaint#e8d4b@Offset(20.0, 46.0), RenderSemanticsAnnotations#39568@Offset(20.0, 46.0),
RenderSemanticsAnnotations#8755e@Offset(20.0, 46.0), RenderTapRegionSurface#bfdb7@Offset(20.0,
46.0), RenderSemanticsAnnotations#a6989@Offset(20.0, 46.0),
RenderSemanticsAnnotations#3a436@Offset(20.0, 46.0), RenderSemanticsAnnotations#b7c7b@Offset(20.0,
46.0), HitTestEntry<HitTestTarget>#36c09(_ReusableRenderView#f920c),
HitTestEntry<HitTestTarget>#e0311(<AutomatedTestWidgetsFlutterBinding>))
If you expected this target not to be able to receive pointer events, pass "warnIfMissed: false" to
"tap()".
To make this error into a non-fatal warning, set WidgetController.hitTestWarningShouldBeFatal to
false.

#0      WidgetController._getElementPoint (package:flutter_test/src/controller.dart:1916:11)
#1      WidgetController.getCenter (package:flutter_test/src/controller.dart:1786:12)
#2      WidgetController.tap (package:flutter_test/src/controller.dart:1034:18)
#3      main.<anonymous closure>.<anonymous closure>.<anonymous closure>.tapCollapseIcon (file:///home/greg/z/flutterz/test/widgets/inbox_test.dart:277:24)
#4      main.<anonymous closure>.<anonymous closure>.<anonymous closure>.<anonymous closure> (file:///home/greg/z/flutterz/test/widgets/inbox_test.dart:350:17)
<asynchronous suspension>
#5      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:183:15)
<asynchronous suspension>
#6      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1018:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

The test description was:
  collapse all-DMs section after scroll
════════════════════════════════════════════════════════════════════════════════════════════════════
00:03 +11 -1: InboxPage collapsing all-DMs section collapse all-DMs section after scroll [E]          
  Test failed. See exception logs above.
  The test description was: collapse all-DMs section after scroll
[…]
```

So:
* The test does fail on the old code. That's good!
* But it fails at line 350, in `tapCollapseIcon`. It does *not* fail at this check on line 348:

```dart
          // Check that the header is still visible even after
          // collapsing the section
          check(headerRow).isNotNull();
```

This test run was on the old code, from before your change, where #391 still exists and it is possible for the header to disappear when collapsing it. The test tried to check for that behavior, but it didn't find anything amiss when it did so. That means that either the test isn't successfully reproducing the behavior, or else the check it's making isn't successfully detecting the behavior.

So, please revise this test so that not only does it pass when paired with your changes, but it fails in a crisp way when run on the code from before your changes.

test/widgets/inbox_test.dart
```diff
@@ -310,6 +327,30 @@ void main() {
...
+          await setupVarious(tester);
+
+           final listFinder = find.byType(Scrollable);
+          // Scroll the [StickyHeaderListView] enough so that
+          // the [_AllDmsSection] shows a sticky header
+          await tester.drag(listFinder, const Offset(0, -50));
+
+          final headerRow = findAllDmsHeaderRow(tester);
+          // Check that the sticky header is present
+          check(headerRow).isNotNull();
```
gnprice, line 340: One useful tactic for making the test robust (as requested in my previous comment, a few lines below here) would be to insert clearer checks on the test setup here.

This check has a comment saying "Check that the sticky header is present". But it seems to check just that there is *a* header for the DMs section — not that it's behaving as a sticky header.

Relatedly: why is 50px an appropriate amount to scroll by? If something about the setup changed so that 50px was no longer an appropriate amount (maybe the layout got taller, or shorter, or the data in `setupVarious` was edited, or something else), can we have something in the test that would detect that?

test/widgets/inbox_test.dart
```diff
@@ -385,6 +426,30 @@ void main() {
...
+
+          final findSectionContent = find.text('specific topic');
+
+          checkAppearsUncollapsed(tester, 1, findSectionContent);
+          await tapCollapseIcon(tester, 1);
+          // Check that the header is still visible even after
+          // collapsing the section
+          check(headerRow).isNotNull();
+          checkAppearsCollapsed(tester, 1, findSectionContent);
```
gnprice, line 448: With the same cross-check as in my other comment — running the new test on the old code — this test also fails to cleanly fail, but a bit differently. Here's the output:
```
$ git checkout origin lib/
$ flutter test test/widgets/inbox_test.dart
[…]
00:03 +12 -1: InboxPage collapsing stream section collapse stream section after scroll                
Expected: a Widget? that:
  is not null
Actual: <null>

#0      check.<anonymous closure> (package:checks/src/checks.dart:85:9)
#1      _TestContext.nest (package:checks/src/checks.dart:751:12)
#2      NullableChecks.isNotNull (package:checks/src/extensions/core.dart:128:20)
#3      main.<anonymous closure>.<anonymous closure>.<anonymous closure>.checkAppearsCollapsed (file:///home/greg/z/flutterz/test/widgets/inbox_test.dart:405:28)
#4      main.<anonymous closure>.<anonymous closure>.<anonymous closure>.<anonymous closure> (file:///home/greg/z/flutterz/test/widgets/inbox_test.dart:448:11)
<asynchronous suspension>
#5      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:183:15)
<asynchronous suspension>
#6      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1018:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

The test description was:
  collapse stream section after scroll
════════════════════════════════════════════════════════════════════════════════════════════════════
00:03 +12 -2: InboxPage collapsing stream section collapse stream section after scroll [E]            
  Test failed. See exception logs above.
  The test description was: collapse stream section after scroll
```

So it does fail, as one would hope — but on line 448, at `checkAppearsCollapsed`. This part:
```dart
          // Check that the header is still visible even after
          // collapsing the section
          check(headerRow).isNotNull();
```
again fails to detect that anything's wrong.

test/widgets/inbox_test.dart
```diff
@@ -385,6 +426,30 @@ void main() {
           checkAppearsUncollapsed(tester, 1, findSectionContent);
         });
 
+        testWidgets('collapse stream section after scroll', (tester) async {
```
gnprice, line 429: Let's also have a test that checks that when you collapse a section while the header is in the middle of the screen, we *don't* scroll to put it at the top. It was nontrivial to get that right:
https://github.com/zulip/zulip-flutter/pull/560#discussion_r1521885523
which makes it an especially good target for a test.

lib/widgets/inbox.dart
```diff
@@ -397,10 +411,14 @@ class _StreamHeaderItem extends _HeaderItem {
   @override get unreadCountBadgeBackgroundColor =>
     subscription.colorSwatch().unreadCountBadgeBackground;
 
-  @override get onCollapseButtonTap => () {
+  @override get onCollapseButtonTap => () async {
     if (collapsed) {
       pageState.uncollapseStream(subscription.streamId);
     } else {
+      await Scrollable.ensureVisible(
```
gnprice, line 418: nit in commit message:

    Fixes: zulip#391.

See our style guide on formatting this line: https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#fixes-format

test/widgets/inbox_test.dart
gnprice: Cool. I think this version of `setupVarious` is fine.

gnprice (commented):
Thanks @sm-sayedi for the revision! The implementation now all looks good, modulo a few small comments. There's some more work still needed on the tests.

sm-sayedi:
Thanks for the review! Revision pushed, with no changes for the tests yet. Following are some of the comments!

test/widgets/inbox_test.dart
```diff
@@ -51,6 +51,11 @@ void main() {
     await tester.pumpAndSettle();
   }
 
+  List<StreamMessage> generateStreamMessages(ZulipStream stream, int count) {
+    return List.generate(count, (index) => eg.streamMessage(
+      stream: stream, topic: '${stream.name} topic $index', flags: []));
```
sm-sayedi, line 56: Added the `flags` just in here, not in the `generateStreamMessages` parms. Should that be added in the `params`? Because it would still not give the full control to specify the `flags` for each individual `message`!

test/widgets/inbox_test.dart
sm-sayedi: Yeah, the `50px` is surely not a robust way of scrolling. A more appropriate way would be to scroll until a specific widget is off-screen. I hope there was an opposite method to `WidgetTester().scrollUntilVisible`, something like `scrollUntilInvisible`, in that case, it would be so much easy and error-free when using it. 

For example, if we wanted to test the `_AllDmsSection`, in the `setupVarious` method, we would add DM messages from two different users, like from `otherUser` and `thirdUser` to `selfUser`.  Now in our test, we would write something like this:

```dart
await setupVarious(tester);

final listFinder = find.byType(Scrollable);

final findSectionContent = find.text(eg.otherUser.fullName);
await scrollUntilInvisible(findSectionContent, 50,
              scrollable: listFinder); // The method I wish was available
              
// Now we're sure that the [_AllDmsSection] has a sticky header, as we scrolled by one message.
              
final headerRow = findAllDmsHeaderRow(tester);
check(headerRow).isNotNull();

// Following, we check if the headerRow is still visible after tapping on it.
```

I even tried to replicate the `scrollUntilInvisible` method from the `WidgetTester().scrollUntilVisible` implementation. Internally it calls the `dragUntilVisible` method, so in fact, I created the `dragUntilInvisible` method like the following:

The `WidgetTester().dragUntilVisible` implementation
```dart
Future<void> dragUntilVisible(
    finders.FinderBase<Element> finder,
    finders.FinderBase<Element> view,
    Offset moveStep, {
      int maxIteration = 50,
      Duration duration = const Duration(milliseconds: 50),
  }) {
    return TestAsyncUtils.guard<void>(() async {
      while (maxIteration > 0 && finder.evaluate().isEmpty) {
        await drag(view, moveStep);
        await pump(duration);
        maxIteration -= 1;
      }
      await Scrollable.ensureVisible(element(finder));
    });
  }
```

My implementation for the `dragUntilInvisible` method
```dart
Future<void> dragUntilInvisible(
    finders.FinderBase<Element> finder,
    finders.FinderBase<Element> view,
    Offset moveStep, {
      int maxIteration = 50,
      Duration duration = const Duration(milliseconds: 50),
  }) {
    return TestAsyncUtils.guard<void>(() async {
      while (maxIteration > 0 && finder.evaluate().isNotEmpty) {
        await drag(view, moveStep);
        await pump(duration);
        maxIteration -= 1;
      }
    });
  }
```

I hoped it would work, but it didn't, and the reason is that the `finder.evaluate.isNotEmpty` is always `true`, which means that even if the widget from the `finder` is off-screen, the `finder` still finds it. I tested this behavior in the `test` for the `stream section` too. 

If I write a simple test like this:

```dart
await setupVarious(tester);

final topicFinder = find.text('stream 2 topic 39');

print(topicFinder.evaluate().length);
```
This will print 1 to the console, even though it should print 0, because a topic with the title `stream 2 topic 39` is surely off-screen in the bottom.

Sorry for the long comment, I hope I was clear.

test/widgets/inbox_test.dart
gnprice: Yeah, let's have `generateStreamMessages` take this as a parameter, so that `prepareVarious` can say `flags: []` explicitly (the same way it does for its direct calls to `eg.streamMessage`).

It's fine that the flags will be the same for all the messages in a given `generateStreamMessages` call. We don't currently have a use case for something more complex; if in the future we do, we can come up with an appropriate API for that then.

test/widgets/inbox_test.dart
gnprice: > This will print 1 to the console, even though it should print 0, because a topic with the title `stream 2 topic 39` is surely off-screen in the bottom.

Ah. What's happening here is that the whole section is currently a single item in the list: #389

As a result, because part of the `stream2` section is in view, the whole section gets built and laid out, so all the widgets that belong to it are findable by `find.text(…)`.

You can fix that by using `Finder.hitTestable`:
```dart
  testWidgets('tmp', (tester) async {
    await setupVarious(tester);
    final topicFinder = find.text('stream 2 topic 39').hitTestable();
    check(topicFinder.evaluate()).isEmpty();
  });
```

That test passes, while the version without `hitTestable` doesn't.

sm-sayedi: I see. Thanks for the comment.

What're your thoughts about the idea and implementation of `dragUntilInvisible`?

gnprice: Sure, seems reasonable.

Do make sure that if it runs out of iterations without succeeding, that it throws instead of returning. I believe `dragUntilVisible` is getting that effect from the `element(finder)` call.

sm-sayedi:
I have pushed the latest revision. Please have a look!

test/widgets/inbox_test.dart
```diff
@@ -252,6 +265,28 @@ void main() {
                 || widget.icon == ZulipIcons.arrow_right))));
       }
 
+      Future<void> dragUntilInvisible(
```
gnprice, line 268: Let's pull this out to the library toplevel, above `main`. It's a generic test helper that we might want to move to a more central file in order to reuse, and that would also make sense to go upstream; so that helps highlight it for such reuse.

gnprice: For similar reasons, let's give this method a doc comment. It can be largely cribbed from the one on `dragUntilVisible` upstream.

test/widgets/inbox_test.dart
```diff
@@ -252,6 +265,28 @@ void main() {
...
+        return TestAsyncUtils.guard<void>(() async {
+          final iteration = maxIteration;
+          while (maxIteration > 0 && finder.evaluate().isNotEmpty) {
+            await tester.drag(view, moveStep);
+            await tester.pump(duration);
+            maxIteration -= 1;
+          }
+          if (maxIteration <= 0 && finder.evaluate().isNotEmpty) {
+            throw StateError(
+              'Finder is still visible after $iteration iterations. Consider increasing the number of iterations.');
```
gnprice, line 285: nit: line too long:
```suggestion
            throw StateError(
              'Finder is still visible after $iteration iterations.'
              ' Consider increasing the number of iterations.');
```

test/widgets/inbox_test.dart
```diff
@@ -385,6 +444,45 @@ void main() {
...
+        testWidgets('collapse stream section after scroll', (tester) async {
+          await setupVarious(tester);
+
+          final topicFinder = find.text('stream 1 topic 2' ).hitTestable();
+          final listFinder = find.byType(Scrollable);
+
+          // Scroll the [StickyHeaderListView] enough so that
+          // the [_StreamSection] shows a sticky header
+          await dragUntilInvisible(
+              tester, topicFinder, listFinder, const Offset(0, -100));
```
gnprice, line 456: Is there a reason for the move step to be different here from in the other test?

If not — well, the difference makes the reader wonder if there's some subtle reason it needs to be that way. So best to avoid raising that question if there isn't. :slightly_smiling_face:

test/widgets/inbox_test.dart
```diff
@@ -310,6 +345,30 @@ void main() {
           checkAppearsUncollapsed(tester, findSectionContent);
         });
 
+        testWidgets('collapse all-DMs section after scroll', (tester) async {
+          await setupVarious(tester);
+
+          final listFinder = find.byType(Scrollable);
+          final findSectionContent =
+              find.text(eg.otherUser.fullName).hitTestable();
```
gnprice, line 353: nit: two-space indents:
```suggestion
          final findSectionContent =
            find.text(eg.otherUser.fullName).hitTestable();
```

(Better yet, give a shorter name like `findContent` and then this fits on one line. But in any case there are some other spots that should be two-space indents too.)

test/widgets/inbox_test.dart
```diff
@@ -310,6 +345,30 @@ void main() {
...
+          Widget? headerRow = findAllDmsHeaderRow(tester);
+          // Check that the header is present (which in this case
+          // is a sticky one as we've scrolled enough)
+          check(headerRow).isNotNull();
+
+          await tapCollapseIcon(tester);
+          // Check that the header is still visible even after
+          // collapsing the section
+          headerRow = findAllDmsHeaderRow(tester);
+          check(headerRow).isNotNull();
```
gnprice, line 369: Rather than mutate a variable, just inline this:
```suggestion
          check(findAllDmsHeaderRow(tester)).isNotNull();
```

If the expression were longer so that this didn't make sense to inline, then it'd be better to define a new `final` variable rather than mutate the old variable. That keeps it easy to see which initializing expression each variable use is referring to.

test/widgets/inbox_test.dart
```diff
@@ -310,6 +345,30 @@ void main() {
...
+          await setupVarious(tester);
+
+          final listFinder = find.byType(Scrollable);
+          final findSectionContent =
+              find.text(eg.otherUser.fullName).hitTestable();
+
+          // Scroll the [StickyHeaderListView] enough so that
+          // the [_AllDmsSection] shows a sticky header
+          await dragUntilInvisible(
+              tester, findSectionContent, listFinder, const Offset(0, -50));
```
gnprice, line 358: This code on its own doesn't ensure that there's a sticky header from that section — it could be that when this completes, that section is out of view entirely.

This logic does work (because of the check that comes next), but I'd tell the story a bit differently:
```dart
          // Scroll part of [_AllDmsSection] offscreen…
          await dragUntilInvisible(
              tester, findSectionContent, listFinder, const Offset(0, -50));
          // … and check its header is still present (which must therefore
          // be as a sticky header).
          check(findAllDmsHeaderRow(tester)).isNotNull();
```

gnprice: In fact better yet, making it a bit more self-contained:
```dart
          // Scroll part of [_AllDmsSection] offscreen…
          await dragUntilInvisible(tester,
            find.text(eg.otherUser.fullName).hitTestable(),
            find.byType(Scrollable),
            const Offset(0, -50),
          );
          // … and check its header is still present (which must therefore
          // be as a sticky header).
          check(findAllDmsHeaderRow(tester)).isNotNull();
```

gnprice (commented):
Thanks @sm-sayedi for the revision! These tests now generally look good too. Some small comments below.

We'll also want another kind of test as mentioned at: https://github.com/zulip/zulip-flutter/pull/560#discussion_r1532787448

test/widgets/inbox_test.dart
sm-sayedi: I just increased it in here so that there is less probability of exhausting the `maxIteration`, as we're scrolling more than the other test. But I will make it compatible with the other one, as it will definitely confuse future readers.

sm-sayedi:
Thanks for the review! Pushed the latest revision. Please have a look.

test/widgets/inbox_test.dart
```diff
@@ -310,6 +352,29 @@ void main() {
...
 
+        testWidgets('collapse all-DMs section after scroll', (tester) async {
+          await setupVarious(tester);
+
+          final listFinder = find.byType(Scrollable);
+          final dmFinder = find.text(eg.otherUser.fullName).hitTestable();
+
+          // Scroll part of [_AllDmsSection] offscreen.
+          await dragUntilInvisible(
+            tester, dmFinder, listFinder, const Offset(0, -50));
```
sm-sayedi, line 363: Still used the variables instead of directly populating the `finder`s for better readability, but please let me know if you think the other way is better!

test/widgets/inbox_test.dart
sm-sayedi: Sorry for this inconvenience each time. As the project doesn't use the `dart auto-formatter`, sometimes it stays out of sight.

test/widgets/inbox_test.dart
```diff
@@ -385,6 +450,50 @@ void main() {
...
+          final headerRowAfterTap = findStreamHeaderRow(tester, 1);
+          // Check that the header is still visible even after
+          // collapsing the section.
+          check(headerRowAfterTap).isNotNull();
+        });
+
+        testWidgets('collapse stream section in the middle of screen', (tester) async {
+          await setupVarious(tester);
+
+          final headerRow = findStreamHeaderRow(tester, 1);
```
sm-sayedi, line 479: The test for [#560 comment](https://github.com/zulip/zulip-flutter/pull/560#discussion_r1532787448) was included the the previous revision, just added a few comments in this one.

gnprice: Ah, I see, thanks.

test/widgets/inbox_test.dart
```diff
@@ -385,6 +450,50 @@ void main() {
...
+          final rectBeforeTap = tester.getRect(find.byWidget(headerRow!));
+
+          // Check that the header is present (which must therefore
+          // be somewhere in the middle of screen)
+          check(headerRow).isNotNull();
```
gnprice, line 484: The `isNotNull` check should come before the `!` — otherwise it's impossible for the check to fail, because if `headerRow` were null we'd have already thrown at `headerRow!`.

test/widgets/inbox_test.dart
```diff
@@ -385,6 +450,50 @@ void main() {
...
+        });
+
+        testWidgets('collapse stream section in the middle of screen', (tester) async {
+          await setupVarious(tester);
+
+          final headerRow = findStreamHeaderRow(tester, 1);
+          final rectBeforeTap = tester.getRect(find.byWidget(headerRow!));
+
+          // Check that the header is present (which must therefore
+          // be somewhere in the middle of screen)
```
gnprice, line 483: Hmm, but why does this mean it's in the middle of the screen? As far as one can see from reading this test's own code, it seems like it could just as well be the sticky header at the top of the viewport.

The reason it's not at the top is because there are unread DMs too, and so there's a DMs section above it. But that's a detail of `setupVarious`, and it seems quite possible that someone in the future could change that, causing this test to just always pass even if the functionality it's meant to test subsequently gets broken. After all, when they do make that change, nothing in this test will break.

There are two basic ways to solve that sort of problem:
* Add a check within this test that verifies that the setup meets the conditions that this test specifically needs from it.
* Specify the relevant details directly in the test, rather than far away in a place where it's not obvious that there are tests that rely on them.

For example here:
* For the first approach, check that the top of `rectBeforeTap` is somewhere below the top of the scrollable viewport.
* Or for the second approach, instead of using `setupVarious`, call `setupPage` directly. That setup can then be simpler than what `setupVarious` does, focusing only on the details this test needs. For example: one DM conversation, then a stream with just one conversation, then a second stream with lots of conversations (enough so that it's clearly more than a screenful, so that scrolling is possible).

Here I think either approach would work fine. (Only one is needed, not both.)

test/widgets/inbox_test.dart
```diff
@@ -385,6 +450,50 @@ void main() {
...
+
+          await tapCollapseIcon(tester, 1);
+
+          final headerRowAfterTap = findStreamHeaderRow(tester, 1);
+          // Check that the header is still visible even after
+          // collapsing the section.
+          check(headerRowAfterTap).isNotNull();
+        });
+
+        testWidgets('collapse stream section in the middle of screen', (tester) async {
```
gnprice, line 476: I think mentioning here the intended outcome (though very concisely) would help the reader understand the test. I suspect it'd have helped me when I overlooked this test previously: https://github.com/zulip/zulip-flutter/pull/560#discussion_r1536588587

So:
```suggestion
        testWidgets('collapse stream section in middle of screen: header stays fixed', (tester) async {
```

test/widgets/inbox_test.dart
```diff
@@ -385,6 +450,50 @@ void main() {
           checkAppearsUncollapsed(tester, 1, findSectionContent);
         });
 
+        testWidgets('collapse stream section after scroll', (tester) async {
```
gnprice, line 453: Similarly to the other test description, let's mention briefly the expected outcome:
```suggestion
        testWidgets('collapse stream section when partially offscreen: header remains sticky at top', (tester) async {
```

That version also gets a bit more precise than "after scroll", which could mean a lot of different things — for example if we had some bug that was triggered by scrolling down, then scrolling back up to the same spot, then doing something, then a test for that bug's fix could equally well be described with "after scroll".

test/widgets/inbox_test.dart
```diff
@@ -385,6 +450,50 @@ void main() {
...
+          await setupVarious(tester);
+
+          final topicFinder = find.text('stream 1 topic 4' ).hitTestable();
```
gnprice, line 456: nit:
```suggestion
          final topicFinder = find.text('stream 1 topic 4').hitTestable();
```

test/widgets/inbox_test.dart
gnprice: For this `dragUntilInvisible` call, it can go either way — there's some value in that variable name `dmFinder` in particular. So arranging it this way is fine.

For `headerRow` below, though, definitely better to inline. When the definition reads:
```dart
          final headerRow = findAllDmsHeaderRow(tester);
```
there's not really anything the variable name is expressing that wouldn't also be clear from seeing the expression `findAllDmsHeaderRow(tester)` directly. Similarly `headerRowAfterTap`.

gnprice (commented):
Thanks for the revision! This is very close now. A couple of comments on that one test I didn't spot in the last revision, and some stylistic comments elsewhere.

sm-sayedi:
Thanks for the review. On it!

Just opened a [draft PR](https://github.com/zulip/zulip-flutter/pull/590) as the tests are remaining. Would love to get a review on it.

sm-sayedi:
Revisions pushed! Ready for another review!

gnprice:
Looks good ­— merging! Thanks @sm-sayedi for all your work on this.

sm-sayedi:
Thank you so much @gnprice !! I appreciate your patience and excellent guidance throughout!!

chrisbobbe:
Yay! I remember feeling disappointed that this fix wasn't in my original implementation of the inbox page. I'm glad it's done now.

sm-sayedi:
Happy we got it done together!!

PR #561, opened 2024-03-12
store: Retry registerQueue on failure; start testing UpdateMachine.load

### LINKED ISSUE #556: Retry registerQueue
This is essential when we're re-registering (as we do since #184) after an event queue expired. Currently, if that re-register fails, the whole event-polling loop stops so that #184 and #185 are effectively defeated.

This appears to be the cause (on at least one occasion) of the polling failure mentioned in #555. See chat thread where we [debugged it with the help of nginx logs](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Only.20.22all.20messages.22.20updating/near/1754227).

gnprice:
Fixes #556.

In addition to the test this adds, I wanted to manually test it
end-to-end, to help be sure this covered the scenario where this
retry is known to be needed in the wild:
 * The app is offline for a while, perhaps because the device
   is asleep.
 * The app comes online, tries polling, and finds the event queue
   has expired, so it attempts a re-register.
 * Before that completes (which after all takes several seconds
   if the realm is a large one), the app goes offline again.
 * That request's response therefore never reaches the app,
   and so when it eventually comes back online it needs to retry.

Step 1 is annoying to carry out literally, because it means
waiting 10 minutes for the event queue to expire.  To work around
that, I sabotaged the getEvents binding function to use a wrong
`queue_id` value:

    'queue_id': RawParameter('wrong' + queueId),

so that the server would always respond with a BAD_EVENT_QUEUE_ID
error, just the same as if the queue had expired.

Then to take the app offline and online again, I just turned
airplane mode on and off on my device.  Because I used a
physical device connected to my computer over USB, that caused
no interference to my watching the logs on the console.

In my manual testing, the retries worked perfectly: no matter
how many times I turned airplane mode on and off, or with what
timing, the app always returned to getting a fresh queue and
polling it for events.

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -704,6 +704,26 @@ class UpdateMachine {
     }
   }
 
+  /// In debug mode, controls whether [registerNotificationToken] should
+  /// have its normal effect.
+  ///
+  /// Outside of debug mode, this is always true and the setter has no effect.
+  static bool get debugEnableRegisterNotificationToken {
+    bool result = true;
+    assert(() {
+      result = _debugEnableRegisterNotificationToken;
+      return true;
+    }());
+    return result;
+  }
+  static bool _debugEnableRegisterNotificationToken = true;
+  static set debugEnableRegisterNotificationToken(bool value) {
+    assert(() {
+      _debugEnableRegisterNotificationToken = value;
+      return true;
+    }());
+  }
+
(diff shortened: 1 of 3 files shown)


chrisbobbe (approved):
Thanks! LGTM; please merge at will after seeing a small comment on a comment.

gnprice:
Thanks for the review! Merged, with that change to the comment.

Issue #554, opened 2024-03-08
Reset backoff interval after successful connection


gnprice:
Since #184, we retry long-poll failures, with exponential backoff up to an interval of 10 seconds.

Even after the device is again able to reach the server, though, the backoff interval on next failure remains 10 seconds, for as long as the app keeps running. That seems unhelpful. Instead, when we find we can again consistently reach the server, the backoff intervals should reset to what they would be if the app were restarted.

The simplest thing to do here would be: as soon as there's one more success, just reset the backoff to its initial state. I'm not sure if a more complex strategy would be best in general. But I think for these long-poll requests the simple thing is fine, for the reasons below.

## Detailed reasoning

The possible risk in completely resetting backoff on a single success is that if the server is having trouble and handling some but not all its requests, that plan would mean we'd do a lot more retries, partway toward what would happen if we weren't backing off at all. I'm not sure if there's a standard recommendation for how to strike a balance there.

However, these particular requests are a long-poll — which means that a *successful* request most often takes a long time to come back: up to 60 seconds. Probably most of the time it's longer than our max backoff interval of 10 seconds. So if after that, our next request randomly fails (because the server is having trouble) and we retry in 100ms, that's not that different from if we retry after 10 seconds.

PR #564, opened 2024-03-13
backoff: reset on successful event poll

chrisirhc:
Fixes: #554

### DIFF
--- a/test/api/backoff_test.dart
+++ b/test/api/backoff_test.dart
@@ -51,4 +51,17 @@ void main() {
       check(maxFromAllTrials).isGreaterThan(expectedMax * 0.75);
     }
   });
+
+  test('BackoffMachine resets duration', () async {
+    final backoffMachine = BackoffMachine();
+    await measureWait(backoffMachine.wait());
+    final duration2 = await measureWait(backoffMachine.wait());
+    check(backoffMachine.waitsCompleted).equals(2);
+
+    backoffMachine.reset();
+    check(backoffMachine.waitsCompleted).equals(0);
+    final durationReset1 = await measureWait(backoffMachine.wait());
+    check(durationReset1).isLessThan(duration2);
+    check(backoffMachine.waitsCompleted).equals(1);
+  });
 }
--- a/lib/api/backoff.dart
+++ b/lib/api/backoff.dart
@@ -49,4 +49,10 @@ class BackoffMachine {
 
     _waitsCompleted++;
   }
+
+  /// Reset the machine to its initial state with minimal wait.
+  void reset() {
+    _startTime = null;
+    _waitsCompleted = 0;
+  }
 }
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -694,7 +694,6 @@ class UpdateMachine {
             assert(debugLog('Transient error polling event queue for $store: $e\n'
                 'Backing off, then will retry…'));
             // TODO tell user if transient polling errors persist
-            // TODO reset to short backoff eventually
             await backoffMachine.wait();
             assert(debugLog('… Backoff wait complete, retrying poll.'));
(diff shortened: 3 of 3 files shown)


chrisbobbe:
I've just edited the issue description to reference the issue it looks like this is about. Please correct it if it's wrong, but if it's right, please add a "Fixes" line to the relevant commit message too, once it's ready to be a not-draft. 🙂

chrisirhc:
Ah yes, I created this but forgot about it. 
I saw a different way of approaching this by reseting the reference to backoffMachine as null instead of maintaining this reset state over at:
https://github.com/zulip/zulip-flutter/blob/79a7384b5c8ad616f9bbd3874f8828736c279538/lib/model/store.dart#L633-L644

I could've gone down that approach as well, to keep the codebase consistent. Let me know if there's an opinion on this.

test/api/backoff_test.dart
```diff
@@ -51,4 +51,17 @@ void main() {
...
+  test('BackoffMachine resets duration', () async {
+    final backoffMachine = BackoffMachine();
+    await measureWait(backoffMachine.wait());
+    final duration2 = await measureWait(backoffMachine.wait());
+    check(backoffMachine.waitsCompleted).equals(2);
+
+    backoffMachine.reset();
+    check(backoffMachine.waitsCompleted).equals(0);
+    final durationReset1 = await measureWait(backoffMachine.wait());
+    check(durationReset1).isLessThan(duration2);
```
gnprice, line 64: This test will flake — it'll fail about 25% of the time. For example:
```
$ time flutter test test/api/backoff_test.dart 
00:01 +1 -1: BackoffMachine.reset [E]                                            
  Expected: a Duration that:
    is less than <0:00:00.023591>
  Actual: <0:00:00.049613>
  Which: is not less than <0:00:00.023591>
  package:checks/src/checks.dart 85:9             check.<fn>
  package:checks/src/checks.dart 708:12           _TestContext.expect
  package:checks/src/extensions/core.dart 167:13  ComparableChecks.isLessThan
  test/api/backoff_test.dart 64:27                main.<fn>
```

The trouble is that the behavior of `wait` is randomized, and the range of possible durations for the second wait (or even the Nth wait for large N) overlaps with the range for the first wait.

That randomization is why the existing test for that method, above, runs 100 trials and then makes a fairly loose check on the results; the parameters are chosen, using some math that's in the comments, to ensure the flake rate is below one in a billion.

(I picked that threshold so that even in a large codebase, even with lots of randomized test cases using that threshold, the total flake rate due to randomization would be negligible compared to other sources of flakiness like races, other bugs, and infra issues.)

We totally could write a working test for this `reset` method by using the same strategy. But maybe the complexity of that is a good reason to skip the `reset` method and go with the other approach you mentioned at https://github.com/zulip/zulip-flutter/pull/564#issuecomment-2021780297 , of just resetting a `BackoffMachine?` local to null.

gnprice (commented):
Thanks @chrisirhc for this PR! I'm back from vacation this week (and the rush of GSoC applicants is over), and I'm starting to catch up now.

One substantive comment below.

Also a commit-message nit: the summary line after the colon should be in sentence case. So e.g. "store: Reset backoff on success", rather than "store: reset …".

chrisbobbe:
Gentle bump to @chrisirhc to see if you're still interested in working on this. 🙂

chrisirhc:
I'll likely get back to this in a few weeks. 
Based on the review comment, the change may end up being smaller/straightforward. So if anyone else wants to go ahead and do it before then, I won't be opposed. 🙂

gnprice:
We ended up implementing the approach from https://github.com/zulip/zulip-flutter/pull/564#issuecomment-2021780297, as #871. Thanks again @chrisirhc!

Issue #562, opened 2024-03-12
msglist: Leave blank space for "mark as read" button so using it doesn't cause messages to shift


gnprice:
A request from @terpimost. See [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flutter.20feedback/near/1753662).

This is a sort of followup to #400.

abelaba:
Hello,

I have been going over the codebase for a couple of days now to understand the project more and read about Zulips best git practices. I would love the chance to work on this project and work on this issue.
 
I have been able to locate which file I need to work on to fix this issue and I have started working on it.

abelaba:
I have been able to fix this issue by updating the MarkAsReadWidget. Currently, I am trying to find a way to update the existing tests. The tests were previously written to check if the size of the widget > 0, but that won't work anymore since the widget will always have a size.

gnprice:
Hi @abelaba, welcome. I've assigned you the issue.

For questions about details of the code, the best place would be a chat thread in `#mobile-team` on chat.zulip.org.

In general, one thing we try to do when writing tests is to write them in a way that reflects the UI as the user sees it. So: as far as the user is concerned, what's different when the button is shown vs. when it's not shown? After your changes, the size of that widget will be the same in both cases; but there will be other things that are different. So then try to write a test that checks for one of those differences.

If you have follow-up questions after spending some time trying to do that, please go ahead and start a thread in `#mobile-team` with details on what you've thought or tried so far, and we'll be glad to help.

PR #565, opened 2024-03-14
msglist: Leave blank space for "mark as read" button so using it doesn't cause messages to shift

abelaba:
This pull request addresses the issue of the message list shifting downwards when the "mark as read" button is pressed. The previous implementation would return an empty widget when the button was pressed and cause the message list to shift down. Now it takes a widget with a definite height, which is equal to the size of the button, and won't cause the message list to shift down.

Fixes: #562

| Before | After |
| --- | --- |
|  https://github.com/zulip/zulip-flutter/assets/61546411/e13f6a05-0226-4a14-9654-8312271281d3 | https://github.com/zulip/zulip-flutter/assets/61546411/b8a66ad3-394f-499f-9079-51591e4d0766 |

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -461,34 +461,39 @@ class MarkAsReadWidget extends StatelessWidget {
     final zulipLocalizations = ZulipLocalizations.of(context);
     final store = PerAccountStoreWidget.of(context);
     final unreadCount = store.unreads.countInNarrow(narrow);
-    return AnimatedCrossFade(
-      duration: const Duration(milliseconds: 300),
-      crossFadeState: (unreadCount > 0) ? CrossFadeState.showSecond : CrossFadeState.showFirst,
-      firstChild: const SizedBox.shrink(),
-      secondChild: SizedBox(width: double.infinity,
-        // Design referenced from:
-        //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=132-9684&mode=design&t=jJwHzloKJ0TMOG4M-0
-        child: Padding(
-          // vertical padding adjusted for tap target height (48px) of button
-          padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 10 - ((48 - 38) / 2)),
-          child: FilledButton.icon(
-            style: FilledButton.styleFrom(
-              backgroundColor: _UnreadMarker.color,
-              minimumSize: const Size.fromHeight(38),
-              textStyle:
-                // Restate [FilledButton]'s default, which inherits from
-                // [zulipTypography]…
-                Theme.of(context).textTheme.labelLarge!
-                // …then clobber some attributes to follow Figma:
-                .merge(const TextStyle(
-                  fontSize: 18,
-                  height: (23 / 18))
-                .merge(weightVariableTextStyle(context, wght: 400))),
-              shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
-            ),
-            onPressed: () => _handlePress(context),
-            icon: const Icon(Icons.playlist_add_check),
-            label: Text(zulipLocalizations.markAllAsReadLabel)))));
+    final areMessagesRead = unreadCount == 0;
+
+    return IgnorePointer(
+      ignoring: areMessagesRead,
+      child: AnimatedOpacity(
+        opacity: areMessagesRead ? 0 : 1,
+        duration: Duration(milliseconds: areMessagesRead ? 2000 : 300),
+        curve: Curves.easeOut,
+        child: SizedBox(width: double.infinity,
+          // Design referenced from:
+          //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=132-9684&mode=design&t=jJwHzloKJ0TMOG4M-0
+          child: Padding(
+            // vertical padding adjusted for tap target height (48px) of button
+            padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 10 - ((48 - 38) / 2)),
+            child: FilledButton.icon(
+              style: FilledButton.styleFrom(
+                backgroundColor: _UnreadMarker.color,
+                minimumSize: const Size.fromHeight(38),
+                textStyle:
+                  // Restate [FilledButton]'s default, which inherits from
+                  // [zulipTypography]…
+                  Theme.of(context).textTheme.labelLarge!
+                  // …then clobber some attributes to follow Figma:
+                  .merge(const TextStyle(
+                    fontSize: 18,
+                    height: (23 / 18))
+                  .merge(weightVariableTextStyle(context, wght: 400))),
+                shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
+              ),
+              onPressed: () => _handlePress(context),
+              icon: const Icon(Icons.playlist_add_check),
+              label: Text(zulipLocalizations.markAllAsReadLabel))))),
+    );
   }
 }
 
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -610,13 +610,10 @@ void main() {
 
   group('MarkAsReadWidget', () {
     bool isMarkAsReadButtonVisible(WidgetTester tester) {
-      // Zero height elements on the edge of a scrolling viewport
-      // are treated as invisible for hit-testing, see
-      // [SliverMultiBoxAdaptorElement.debugVisitOnstageChildren].
-      // Set `skipOffstage: false` here to safely target the
-      // [MarkAsReadWidget] even when it is inactive.
-      return tester.getSize(
-        find.byType(MarkAsReadWidget, skipOffstage: false)).height > 0;
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+      final finder = find.text(
+        zulipLocalizations.markAllAsReadLabel).hitTestable();
+      return finder.evaluate().isNotEmpty;
     }
 
     testWidgets('from read to unread', (WidgetTester tester) async {
@@ -646,6 +643,31 @@ void main() {
       ));
       await tester.pumpAndSettle();
       check(isMarkAsReadButtonVisible(tester)).isFalse();
+    });
+
+    testWidgets("messages don't shift position", (WidgetTester tester) async {
+      final message = eg.streamMessage(flags: []);
+      final unreadMsgs = eg.unreadMsgs(streams:[
+        UnreadStreamSnapshot(topic: message.subject, streamId: message.streamId,
+          unreadMessageIds: [message.id])
+      ]);
+      await setupMessageListPage(tester,
+        messages: [message], unreadMsgs: unreadMsgs);
+      check(isMarkAsReadButtonVisible(tester)).isTrue();
+      check(tester.widgetList(find.byType(MessageItem))).length.equals(1);
+      final before = tester.getTopLeft(find.byType(MessageItem)).dy;
+
+      store.handleEvent(UpdateMessageFlagsAddEvent(
+        id: 1,
+        flag: MessageFlag.read,
+        messages: [message.id],
+        all: false,
+      ));
+      await tester.pumpAndSettle();
+      check(isMarkAsReadButtonVisible(tester)).isFalse();
+      check(tester.widgetList(find.byType(MessageItem))).length.equals(1);
+      final after = tester.getTopLeft(find.byType(MessageItem)).dy;
+      check(after).equals(before);
     });
 
     group('onPressed behavior', () {


chrisbobbe:
Thanks for the PR!

What happens if you go to your device's settings and increase the font size? My guess is that the button's height will be more than 58, and the empty space that replaces the button will be too small. When the button appears or disappears, the layout will seem like it's "jumping".

Instead of replacing the button with a blank placeholder with an explicit height, can you try an approach that still lays out the button, but just doesn't show it? Maybe with a combination of [`AnimatedOpacity`](https://api.flutter.dev/flutter/widgets/AnimatedOpacity-class.html) and [`IgnorePointer`](https://api.flutter.dev/flutter/widgets/IgnorePointer-class.html)?

abelaba:
Thank you for your comment. I hadn't thought about the font size. You are right with my first implementation when I increased my device font size the layout would shift, now I have updated the widget with AnimatedOpacity and IgnorePointer and even if I increase my device font size the layout won't shift.

I don't know why the CI is failing. It says it has something to do with the intl package, but I haven't modified that.

test/widgets/message_list_test.dart
```diff
@@ -610,13 +610,28 @@ void main() {
...
+        matching: find.byType(AnimatedOpacity),
+      );
+
+      final IgnorePointer ignorePointer = tester.widget(ignorePointerFinder);
+      final AnimatedOpacity animatedOpacity = tester.widget(animatedOpacityFinder);
+
+      return !ignorePointer.ignoring && animatedOpacity.opacity == 1.0;
     }
 
     testWidgets('from read to unread', (WidgetTester tester) async {
```
gnprice, line 637: In addition to updating the existing tests, this change should add a new test: we want to check that the messages don't shift position when marking as read.

gnprice (commented):
Thanks @abelaba for the contribution!

Trying this out (and from your screen captures above), the behavior generally looks good but there's one aspect we'll definitely want to fix: when marking as read, the button fades out quickly, but the unread markers on the individual messages fade out more slowly. The juxtaposition is kind of jarring. The two animations represent the same action happening (the messages getting marked as read), so let's have them happen on the same schedule.

One other small comment on the implementation code, and there's also more work needed in order to have solid tests for this; details below.

lib/widgets/message_list.dart
```diff
@@ -461,34 +461,36 @@ class MarkAsReadWidget extends StatelessWidget {
...
-                .merge(weightVariableTextStyle(context, wght: 400))),
-              shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
-            ),
-            onPressed: () => _handlePress(context),
-            icon: const Icon(Icons.playlist_add_check),
-            label: Text(zulipLocalizations.markAllAsReadLabel)))));
+    return IgnorePointer(
+      ignoring: unreadCount == 0,
+      child: AnimatedOpacity(
+        opacity: (unreadCount > 0) ? 1 : 0,
```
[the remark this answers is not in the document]
abelaba: I have fixed this and pulled it out as a local boolean variable.

test/widgets/message_list_test.dart
abelaba: I have added a new test that addresses this. It checks if the vertical position of the first message item is the same before and after the "Mark all messages as read" button is pressed.

test/widgets/message_list_test.dart
```diff
@@ -610,13 +610,28 @@ void main() {
...
+
+      final animatedOpacityFinder = find.descendant(
+        of: markAsReadWidget,
+        matching: find.byType(AnimatedOpacity),
+      );
+
+      final IgnorePointer ignorePointer = tester.widget(ignorePointerFinder);
+      final AnimatedOpacity animatedOpacity = tester.widget(animatedOpacityFinder);
+
+      return !ignorePointer.ignoring && animatedOpacity.opacity == 1.0;
```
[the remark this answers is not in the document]
gnprice: Cool — yeah, `Finder.hitTestable` is a good solution.

test/widgets/message_list_test.dart
```diff
@@ -909,6 +921,35 @@ void main() {
...
+          messages: [message.id],
+          all: false,
+        ));
+
+        await tester.pumpAndSettle();
+        check(isMarkAsReadButtonVisible(tester)).isFalse();
+
+        final firstMessageItemTopPositionAfter = tester.getTopLeft(find.byType(MessageItem).first).dy;
+
+        check(firstMessageItemTopPositionBefore).equals(firstMessageItemTopPositionAfter);
```
gnprice, line 950: These variable names are extremely long. That makes for rather a mouthful when reading through the code; and it also means that a line like this one, with just two of these variables and otherwise very simple, is already overly long.

Better names would be `messagePositionBefore` / `messagePositionAfter`.

Or just: `before` and `after`. These are local variables, which only exist within this fairly short function, so the names only have to say enough detail to clearly distinguish them from the other things going on within the function. And it's clear from the initializer expressions, like `tester.getTopLeft(find.byType(MessageItem).first).dy`, that this is a position of the top of the first message item.

gnprice (commented):

test/widgets/message_list_test.dart
```diff
@@ -610,13 +610,25 @@ void main() {
...
+      // Locate the [Text] widget with the markAllAsReadLabel
+      // to determine if the [MarkAsReadWidget] is visible.
+      // This is done by finding a descendant of the [MarkAsReadWidget]
+      // that is a [Text] widget with the markAllAsReadLabel.
+      // The hitTestable() method is used to make the descendant widget hit-testable,
+      // allowing it to be found and evaluated in the widget tree.
+      final markAsReadWidgetButtonDescendant = find.descendant(
+        of: markAsReadWidget,
+        matching: find.text(zulipLocalizations.markAllAsReadLabel),
+      ).hitTestable();
```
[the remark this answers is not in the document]
abelaba: I have fixed this.

test/widgets/message_list_test.dart
```diff
@@ -610,13 +610,25 @@ void main() {
...
+      // Locate the [MarkAsReadWidget] and check its state
+      // to determine if it is visible.
+      var markAsReadWidget = find.byType(MarkAsReadWidget);
+
+      // Locate the [Text] widget with the markAllAsReadLabel
+      // to determine if the [MarkAsReadWidget] is visible.
+      // This is done by finding a descendant of the [MarkAsReadWidget]
+      // that is a [Text] widget with the markAllAsReadLabel.
+      // The hitTestable() method is used to make the descendant widget hit-testable,
+      // allowing it to be found and evaluated in the widget tree.
```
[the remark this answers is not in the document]
abelaba: Thank you for the explanation. I hadn't thought about it that way. I have removed the comments.

test/widgets/message_list_test.dart
```diff
@@ -909,6 +921,35 @@ void main() {
...
+        final firstMessageItemTopPositionBefore = tester.getTopLeft(find.byType(MessageItem).first).dy;
+
+        store.handleEvent(UpdateMessageFlagsAddEvent(
+          id: 1,
+          flag: MessageFlag.read,
+          messages: [message.id],
+          all: false,
+        ));
+
+        await tester.pumpAndSettle();
```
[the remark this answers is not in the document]
abelaba: I have moved the test to just above the "onPressed behavior" group. I have also fixed the long lines.

test/widgets/message_list_test.dart
abelaba: I have renamed the variables to `before` and `after`.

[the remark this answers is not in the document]
abelaba: I tried to use the` .single `method, but that returns an `Element` rather than a `Finder` and the `tester.getTopLeft` method takes a `Finder`. So in order to explicitly show that there is only one MessageItem on this test, I have added a check that checks if the number of MessageItem widgets found is equal to 1. If there is a better way to handle this test I am open to suggestions.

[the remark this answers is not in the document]
gnprice: Hmm, I see. Sure, this works.

test/widgets/message_list_test.dart
```diff
@@ -648,6 +645,31 @@ void main() {
...
+    testWidgets('messages don\'t shift position', (WidgetTester tester) async {
+      final message = eg.streamMessage(flags: []);
+      final unreadMsgs = eg.unreadMsgs(streams:[
+        UnreadStreamSnapshot(topic: message.subject, streamId: message.streamId,
+          unreadMessageIds: [message.id])
+      ]);
+      await setupMessageListPage(tester,
+        messages: [message], unreadMsgs: unreadMsgs);
+      check(isMarkAsReadButtonVisible(tester)).isTrue();
+      check(tester.widgetList(find.byType(MessageItem)).length).equals(1);
```
gnprice, line 657: nit: do more processing on the result of `check`, instead of its argument:
```suggestion
      check(tester.widgetList(find.byType(MessageItem))).length.equals(1);
```

This allows `package:checks` to give more informative output if the check fails.

test/widgets/message_list_test.dart
```diff
@@ -648,6 +645,31 @@ void main() {
...
+        id: 1,
+        flag: MessageFlag.read,
+        messages: [message.id],
+        all: false,
+      ));
+      await tester.pumpAndSettle();
+      check(isMarkAsReadButtonVisible(tester)).isFalse();
+      check(tester.widgetList(find.byType(MessageItem)).length).equals(1);
+      final after = tester.getTopLeft(find.byType(MessageItem)).dy;
+      check(before).equals(after);
```
gnprice, line 670: Check actual equals expected:
```suggestion
      check(after).equals(before);
```

That way `check` is able to give accurate output if the check fails. (To see the failure, use `git checkout @~ lib/` and then `flutter test test/widgets/message_list_test.dart`, to run your new test on the old code.)

test/widgets/message_list_test.dart
```diff
@@ -648,6 +645,31 @@ void main() {
       check(isMarkAsReadButtonVisible(tester)).isFalse();
     });
 
+    testWidgets('messages don\'t shift position', (WidgetTester tester) async {
```
gnprice, line 648: nit: make this a double-quoted string, to look a bit cleaner with the embedded apostrophe

gnprice (commented):
Thanks for the revision! Just small comments remain.

(I also haven't yet tested this or the previous revision live, but it's late in the evening now here.)

test/widgets/message_list_test.dart
```diff
@@ -610,13 +610,10 @@ void main() {
...
     bool isMarkAsReadButtonVisible(WidgetTester tester) {
-      // Zero height elements on the edge of a scrolling viewport
-      // are treated as invisible for hit-testing, see
-      // [SliverMultiBoxAdaptorElement.debugVisitOnstageChildren].
-      // Set `skipOffstage: false` here to safely target the
-      // [MarkAsReadWidget] even when it is inactive.
-      return tester.getSize(
-        find.byType(MarkAsReadWidget, skipOffstage: false)).height > 0;
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+      final markAsReadWidgetButton = find.text(
```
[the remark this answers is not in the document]
abelaba: Fixed.

test/widgets/message_list_test.dart
abelaba: Okay. Fixed.

abelaba: Okay. Fixed.

test/widgets/message_list_test.dart
abelaba: Fixed.

test/widgets/message_list_test.dart
abelaba: Thanks for the explanation. Fixed.

gnprice:
Thanks for the revision! This looks good. Merging, with one fix in the commit message:

    msglist: Leave blank space for "mark as read" button so using it doesn't cause messages to shift
    
    This commit addresses the issue of the message list shifting downwards when the "mark as read" button is pressed. The previous implementation would return an empty widget when the button was pressed and cause the message list to shift down. Now it takes a widget which will display the button based on the opacity level and will disable the region the button holds once the button is pressed, thus won't cause the message list to shift down.
    
    Fixes: #562

Prose should be wrapped to about 68 columns, and at most 70:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#formatting-guidelines

And then the summary line can go a bit longer than that, because it can be hard to squeeze it down while conveying all the information; but I did some tightening of that too. The result:

    msglist: Reserve space for "mark as read" button so messages don't shift up/down
    
    This commit addresses the issue of the message list shifting downwards
    when the "mark as read" button is pressed. The previous implementation
    would return an empty widget when the button was pressed and cause the
    message list to shift down. Now it takes a widget which will display
    the button based on the opacity level and will disable the region the
    button holds once the button is pressed, thus won't cause the message
    list to shift down.
    
    Fixes: #562

Issue #388, opened 2023-11-18
Support adding an arbitrary reaction


gnprice:
This will be a followup to:
* #125 

After that issue (and for Beta 1), we'll have the ability to add a thumbs-up from the message action sheet, and to tap on an existing reaction to join in or to remove it. This issue is to add the ability to choose an arbitrary reaction to add.

The zulip-mobile RN app has this functionality but the UI is kind of clumsy. With Flutter we have a lot more control to make the UI work however we like. So this is an opportunity to reimagine this UI to be much nicer. /cc @alya @terpimost

Some previous threads with ideas for improving this UI in the context of the zulip-mobile RN app:
* https://github.com/zulip/zulip-mobile/issues/4565 (and [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Quick.20reactions/near/1156419))
* https://github.com/zulip/zulip-mobile/issues/4567
* https://github.com/zulip/zulip-mobile/issues/4578

I think all three of those are good ideas, all together or in any combination, or we can go in another direction if we think it'd be good. For the third one (a button next to existing reactions) we can borrow the design from web, but the other two don't correspond to anything in web and so we'd need fresh designs for them.

chrisbobbe:
A user reports in [CZO](https://chat.zulip.org/#narrow/stream/48-mobile/topic/order.20of.20emojis.20for.20reaction/near/1706954) that zulip-mobile's choice of putting `:heart:`, `:love:`, and `:love_you:` at the top of the emoji picker doesn't make sense for a work context: "I’m always a bit scared I’ll tap by mistake on the love-you emoji which is pretty close." Which makes sense; I've had a similar feeling myself! 😅

chrisbobbe:
We've had another report by email, of wanting a less clumsy add-reaction UI than we currently have in zulip-mobile. They mention (if I understand correctly) having some frequently-used emoji directly in the message-long-press sheet, as discussed here.

They also had another interesting idea:

> If I can configure a swipe left action to be a certain emoji I would love that I think

alainjr10:
Hey. I started doing some prelimary contributions to this issue, however, i can't seem to be able to pull all the available emojis. According to #669 , there's still an open issue about tracking all supported emojis. However, on Zulip web client, we can access a list of all emojis. so i was wondering how i can access those emojis to use on mobile.

gnprice:
[...]

PR #566, opened 2024-03-15
add emoji picker for reactions

chrisirhc:
TODO:
- [x] Tests
- [x] Localization
- [ ] Add support for all emoji that aren't available by default in emoji_picker_flutter (e.g. symbols like *️⃣ , which are available in iamcal but not in emoji_picker_flutter) (Related: https://github.com/Fintasys/emoji_picker_flutter/issues/153 ) 
- [x] Filter away emoji that were already added by current user (similar to previous ThumbsUp behavior)
- [ ] Show 3-4 most recent emojis in the menu for reuse via `final recentEmojis = await EmojiPickerUtils().getRecentEmojis();`
- [ ] Clean up commit messages to fit style guide

Demo:
[Screen_recording_20240315_110944 Zulip Emoji.webm](https://github.com/zulip/zulip-flutter/assets/132584/03e59d98-9380-4ea3-a147-c56dc5694193)

Partially addresses #388


gnprice:
Thanks @chrisirhc for sending this! Sorry I didn't manage to look at it sooner.

There's a couple of key pieces of functionality this is missing that are part of Zulip's model of emoji:
* the list of known emoji from Unicode varies as the server gets upgraded;
* more critically, there are custom emoji that people have added in each Zulip realm, beyond those in Unicode.
* (Also there's one "Zulip extra emoji", which is a sort of built-in global emoji: `:zulip:`, whose glyph is the Zulip logo.)

Those are covered in #669, which I've recently largely completed.

The `emoji_picker_flutter` package which this pulls in is interesting, but AFAICT from the readme it looks like it wants to handle only Unicode emoji — not custom emoji.

So I'm closing this PR because I don't think the direction it's taking is one that will meet our needs.

Issue #15, opened 2023-02-28
Pin a precise version of Flutter


gnprice:
So far, in the prototype phase, we've been able to just always use the latest Flutter version in a given channel (either beta or main). At some point we will want more control, specifying a particular version and being sure everyone gets that version even if it's not the very newest. That way everyone's seeing the same behavior, and a change happening outside our own repo can't break our tests.

In particular, if we use versions from the Flutter main branch, then that will be a constantly moving target. Using the moving target will still be OK as long as we don't frequently see anything change that breaks us, but at some point we'll want to pin the exact version to be sure of that.

### Implementation notes

The `pubspec.yaml` file lets us specify a lower bound for the Flutter version, but not an upper bound. (The syntax permits an upper bound, but it's ignored.) In particular it doesn't permit specifying an exact version. Upstream discussions:
* https://github.com/flutter/flutter/issues/95472#issuecomment-1119312969
* https://github.com/flutter/flutter/pull/26577#issuecomment-454698844
* https://github.com/flutter/flutter/issues/40428
* https://github.com/flutter/flutter/issues/95807#issuecomment-1006724141

Relatedly, while the `flutter` tool provides `flutter upgrade` as a handy way to get the latest Flutter version within any given channel (stable, beta, or main), there's no such way to get a specific version. So if Flutter did respect the `pubspec.yaml` file as far as complaining when the version was newer than specified, it still wouldn't provide a convenient way to actually get the specified version.

From those upstream threads, it seems like the usual solution in the Flutter community is to use one of several tools people have built to manage Flutter versions. I haven't looked into these in detail yet at all, but listing the ones I've seen mentioned:
* https://fvm.app/
* https://github.com/passsy/flutter_wrapper

So when pursuing this, we'll want to look at those and also look to see if there are other implementations to consider.

gnprice:
This caused a bit of confusion in a recent PR:
* https://github.com/zulip/zulip-flutter/pull/223#discussion_r1267409152

We also had an outright build-break a couple of months ago:
* https://github.com/zulip/zulip-flutter/commit/2f0f4692f2d55748d055ac7c140ab7414307a36d (that fix was from #103)

And there's the recurring small nuisance of:
* #85 

So while our README says that our lack of pinning so far hasn't been a problem, that seems to no longer be true. It's not an overwhelming problem, but it is an actual problem.

---

[...]

PR #567, opened 2024-03-18
deps: Upgrade `intl` constraint, from ^0.18.0 to ^0.19.0

chrisbobbe:
No potentially breaking changes identified in the changelog:
  https://github.com/dart-lang/i18n/blob/main/pkgs/intl/CHANGELOG.md#0190

This addresses the following error:

  Note: intl is pinned to version 0.19.0 by flutter_localizations
    from the flutter SDK.
  See https://dart.dev/go/sdk-version-pinning for details.

  Because zulip depends on flutter_localizations from sdk which
    depends on intl 0.19.0, intl 0.19.0 is required.
  So, because zulip depends on intl ^0.18.0, version solving failed.

  You can try the following suggestion to make the pubspec resolve:
  * Try upgrading your constraint on intl: flutter pub add intl:^0.19.0

Fixes: #575

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -764,7 +764,7 @@ class _SendButtonState extends State<_SendButton> {
         tooltip: zulipLocalizations.composeBoxSendTooltip,
         style: const ButtonStyle(
           // Match the height of the content input.
-          minimumSize: MaterialStatePropertyAll(Size.square(_sendButtonSize)),
+          minimumSize: WidgetStatePropertyAll(Size.square(_sendButtonSize)),
           // With the default of [MaterialTapTargetSize.padded], not just the
           // tap target but the visual button would get padded to 48px square.
           // It would be nice if the tap target extended invisibly out from the
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -616,10 +616,10 @@ packages:
     dependency: "direct main"
     description:
       name: intl
-      sha256: "3bc132a9dbce73a7e4a21a17d06e1878839ffbf975568bc875c60537824b0c4d"
+      sha256: d6f56758b7d3014a48af9701c085700aac781a92a87a62b1333b46d8879661cf
       url: "https://pub.dev"
     source: hosted
-    version: "0.18.1"
+    version: "0.19.0"
   io:
     dependency: transitive
     description:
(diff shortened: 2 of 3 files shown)


gnprice:
Thanks!

Let's ~~follow this, as another commit in the same PR,~~ accompany this, squashed into the same commit, with bumping the minimum Flutter version in pubspec.yaml. As is, this change fixes things when you have a new Flutter version, but breaks them if you have an older Flutter version:
```
$ tools/check 
Running analyze...
Resolving dependencies... 
Note: intl is pinned to version 0.18.1 by flutter_localizations from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.


Because zulip depends on flutter_localizations from sdk which depends on intl
  0.18.1, intl 0.18.1 is required.
So, because zulip depends on intl ^0.19.0, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Consider downgrading your constraint on intl: flutter pub add intl:^0.18.1
Running test...
Resolving dependencies... 
Note: intl is pinned to version 0.18.1 by flutter_localizations from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.


Because zulip depends on flutter_localizations from sdk which depends on intl
  0.18.1, intl 0.18.1 is required.
So, because zulip depends on intl ^0.19.0, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Consider downgrading your constraint on intl: flutter pub add intl:^0.18.1
Running build_runner...
Running drift...
Running icons...

FAILED: analyze test

To rerun the suites that failed, run:
  $ tools/check analyze test
```

Fundamentally the root issue here is #15 — we should at some point get around to pinning the Flutter version, and then this sort of thing would just be another version constraint we update at the same time as updating that pin. But pending that, when we make a change that causes things to stop working on an older Flutter version, we should reflect that in the Flutter version constraint we do express.

In particular I think it'd be good to maintain an invariant that if you take the Flutter version lower bound in the pubspec, and you *do* interpret it as a pin of an exact version, then things work. That way there's always a straightforward, unambiguous way to get a working setup — and what's missing (for #15) is just the automation to update to that exact version.

gnprice:
(I also just filed #575 for the issue this fixes; so please add a "fixes" line for that in the commit message.)

chrisbobbe:
Thanks for the review! Revision pushed, and did a quick manual smoke test on my iPhone. I saw in particular that the send button on the compose box looks right.

gnprice:

The changes all LGTM. I'd like the commit message to foreground the Flutter version bump more; I'll revise that and merge.

Issue #280, opened 2023-08-17
Open links externally by default on iOS


gnprice:
When opening an external link, we have a choice whether to do so within the app or in an external browser.

Currently we do so following the upstream Flutter defaults [documented at launchUrl](https://pub.dev/documentation/url_launcher/latest/url_launcher/launchUrl.html) (except on Android as described at #279). On iOS, that means opening an in-app browser.

But what we chose on zulip-mobile was for the default on iOS to be an external browser, for reasons discussed [here in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1169118), and I think all the same reasoning still applies. So we should do that instead.

Separately:
* We should offer an option (again just as we do in zulip-mobile), so that this is only about the default. That's part of #97.

PR #574, opened 2024-03-20
content: Open links externally on iOS

rajveermalviya:
On iOS we prefer LaunchMode.externalApplication because LaunchMode.platformDefault uses SFSafariViewController which provides a weird UX for reasons mentioned in [this]( https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1169118) discussion.

| Before | After |
| - | - |
| before |  after |

Fixes: #280

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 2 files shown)


lib/widgets/content.dart
```diff
@@ -895,7 +896,14 @@ void _launchUrl(BuildContext context, String urlString) async {
   bool launched = false;
   String? errorMessage;
   try {
-    launched = await ZulipBinding.instance.launchUrl(url);
+    launched = await ZulipBinding.instance.launchUrl(url, mode: switch (defaultTargetPlatform) {
```
gnprice, line 899: nit: line too long:
```suggestion
    launched = await ZulipBinding.instance.launchUrl(url,
      mode: switch (defaultTargetPlatform) {
```

lib/widgets/content.dart
```diff
@@ -895,7 +896,14 @@ void _launchUrl(BuildContext context, String urlString) async {
...
+      // On iOS we prefer LaunchMode.externalApplication because LaunchMode.platformDefault
+      // uses SFSafariViewController which provides a weird UX for reasons mentioned
```
gnprice, line 901: nit: lines too long

(and ditto for the same text in the commit message)

gnprice (commented):
Thanks @rajveermalviya! This looks good except a couple of formatting nits below. I'll fix those and merge.

(The CI failure is #575. But `tools/check` passes for me locally, on an appropriate version of Flutter, so all's well.)

Issue #311, opened 2023-09-22
Overall app navigation for beta


gnprice:
In the current implementation, the way you get to the various screens (like the "All messages" narrow and the list of DM conversations) is from a "home page" that just has some buttons. It's a consciously prototype-level UI.

One reason we've stuck with that so far is that we don't yet have all the major screens that would naturally belong in an overall navigation for the app. The two still open now are:
* #187 
* #117

For the beta, we should have some more presentable answer. Probably the home screen should be the inbox view #117, rather than a bunch of nav buttons; the navigation then has to go someplace else.

A first answer would be to copy the navigation we have in zulip-mobile, with bottom tabs. We also have some designs for an overhaul of the app's navigation. Given the beta timeline, we may not be able to iterate on those and integrate them for the beta, but we can revisit them for the next stage.

gnprice:
The discussion above (in the issue description) is focused on the navigation between different screens for dealing with a given account. For probably most users, there's only one account they use and so this is the whole story.

But because for some users it is essential to use multiple accounts — not least, for us developers as we use test accounts and test servers as well as our own main accounts — we'll always need some reasonable way to switch between accounts and to add a new account. So that should be part of our navigation thinking too.

The current solution, which we've had since very early in the prototype, is that the app opens showing the choose-account screen, and you can tap an account from there to start loading data for that account. This works, but it's a bit of an extra step when starting the app, and for users who have just one account it may feel unnecessary.

It'd be good to instead go straight to the account you were last using. But in order to do that, we'll need some other place in the UI for you to get to other accounts. See [chat discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/zulip.20flutter.20login.20screen/near/1714661).

c-o-d-e-y-o-d-a:
Hello there, has the issue  "option to choose home view.." been assigned to someone yet? I would like to work on this.

gnprice:
@c-o-d-e-y-o-d-a That issue has its own thread, #517. Discussion of it should go there, or in Zulip.

If you'd like to work on that issue (or another issue), please see the Zulip project's guide to "where to look for an issue", picking an issue, and claiming an issue:
https://zulip.readthedocs.io/en/latest/contributing/contributing.html#your-first-codebase-contribution

gnprice:
[...]

Issue #76, opened 2023-04-20
Implement deep links


gnprice:
This feature means setting things up so that when you open an appropriate Zulip link from arbitrary apps on your device — like your mail client when reading a Zulip email notification, or your browser when looking at a GitHub thread where someone refers to a Zulip conversation — it opens in the Zulip app, rather than a browser, and goes directly to an appropriate screen in the app.

This is a feature we don't currently have in the zulip-mobile RN app (nor for that matter in zulip-desktop). It'd be a nice feature to have.

For Zulip, there are two ways the feature can work, and we probably ultimately want both of them:
* Handle a URL like `https://chat.zulip.org/` or `https://example.zulipchat.com/#narrow/stream/37-foo/topic/bar` — the same URLs that the web app puts in the location bar, and provides at "Copy link to message".

  This is the most convenient, when it applies. It requires us to tell the OS at app install time (or upgrade time) what URL patterns we can open, so we'll only be able to do it for URLs like those of Zulip Cloud and chat.zulip.org, not for arbitrary Zulip servers.

* Make up a way of encoding an arbitrary Zulip realm URL, plus that other information like `#narrow/stream/37-foo/topic/bar` that the web app puts in a fragment, into a `zulip://…` URL. Then handle those.

  This one can work in full generality, for all servers.

  In order to be useful, it requires that we start providing these URLs in some contexts where people might want to follow them. For example, Zulip email notifications have a "view it in Zulip" link, and we could start providing two different links there: one for "on the web" and one for "in the Zulip app". (Ideally this would be combined with zulip-desktop gaining support for the same URLs around the same time.)

Chat threads with further technical discussion:
* [#mobile > Deep linking from outside the app, 2021-02](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Deep.20linking.20from.20outside.20the.20app/near/1111895)
* [#mobile > deep links?, 2021-03](https://chat.zulip.org/#narrow/stream/48-mobile/topic/deep.20links.3F/near/1148057)

Related issues in zulip-mobile and zulip-desktop, for different aspects of this:
* https://github.com/zulip/zulip-mobile/issues/4577
* https://github.com/zulip/zulip-desktop/issues/470
* https://github.com/zulip/zulip-mobile/issues/4751
* https://github.com/zulip/zulip-mobile/issues/4776

Draft implementation for zulip-mobile on Android, of the `zulip://` version:
* https://github.com/zulip/zulip-mobile/pull/4775

[...]

PR #578, opened 2024-03-22
Add bottom bar for navigation, tabs for inbox

chrisirhc:
I'm currently just fiddling around and looking and what maybe needs to be refactored around to get a similar navigation as the current Zulip app.

At this stage, the PR is a set of minimal changes to make it look/feel closer to a slightly more polished navigation. 

So far some findings:
- A more sophisticated router like go-router might be useful so that we can create deep links that follow the exact routing that the Zulip web client uses. Except with 1 main difference, specifying the server needs to be in the url.
  - Not sure what this would look like, perhaps: `https://open.zulip.app/<serverIdentifier>/<path on zulip server>`

Screenshot_20240326_215308

[Screen_recording_20240326_215102.webm](https://github.com/zulip/zulip-flutter/assets/132584/6988aca5-2bf9-4216-8a14-0828e368c8e0)

Note that the animations between the different navigation routes for the bottom navigation bar is slightly incorrect, since it uses a *push* animation that applies to the entire page including the navigation bar, rather than only animating/transitioning the main body area that's being changed.

Partially addresses #311

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -255,6 +255,52 @@ class HomePage extends StatelessWidget {
 
     return Scaffold(
       appBar: AppBar(title: const Text("Home")),
+      bottomNavigationBar: NavigationBar(
+        selectedIndex: 0,
+        destinations: [
+          const NavigationDestination(
+            selectedIcon: Icon(Icons.inbox),
+            icon: Icon(Icons.inbox_outlined),
+            label: 'Inbox'),
+          const NavigationDestination(
+            selectedIcon: Icon(Icons.tag),
+            icon: Icon(Icons.tag_outlined),
+            label: 'Streams'),
+          const NavigationDestination(
+            selectedIcon: Icon(Icons.group_outlined),
+            icon: Icon(Icons.group),
+            label: 'Direct Messages'),
+          // TODO enable this when it's available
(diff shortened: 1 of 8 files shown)


Syrineladeb02:
Great work on this pull request, Chrisirhc!
I've thoroughly tested the changes you've made, and everything appears to be functioning smoothly. 
I particularly appreciate the enhancements made to the navigation bar and the streamlined user experience it provides.
One suggestion for improvement could be to consider implementing a more sophisticated router, such as go-router, to enable deep links that closely mimic the routing behavior of the Zulip web client, while ensuring server identification in the URL.

gnprice:
Thanks @chrisirhc for exploring this and sending the PR!

We've more recently gotten around to thinking through what we want the navigation to look like, and I just wrote down a proper spec for it:
* #1035 

It overlaps with what this PR did, so I've added a link there saying this may be a useful starting point.


> A more sophisticated router like go-router might be useful so that we can create deep links that follow the exact routing that the Zulip web client uses. Except with 1 main difference, specifying the server needs to be in the url.
> * Not sure what this would look like, perhaps: `https://open.zulip.app/<serverIdentifier>/<path on zulip server>`

Yeah, deep links would be a useful feature but are tricky. Previous discussion here:
* #76

PR #579, opened 2024-03-23
pin flutter SDK version via git submodule

### LINKED ISSUE #15: Pin a precise version of Flutter

chrisirhc:
* Introduce a git submodule to pin the version of flutter SDK under
  `vendor/flutter`.
* Use direnv to add `vendor/flutter/bin` to the current PATH so that
  subsequent calls to `flutter` on the shell uses the our pinned
  version of flutter.
* Update instructions to use pinned version of flutter.
* Pin to 18340ea16c of flutter which matches the current lowerbound
version in pubsec.yaml (3.21.0-12.0.pre.26).

In addition, this PR also includes in a separate commit:
* ci-flutter-main job which runs on pushes, to validate the build with the latest main commit. (see https://github.com/zulip/zulip-flutter/issues/15#issuecomment-2016195583 for context)

NOTE: Users can still choose to opt-out and use their own version of
flutter. This just provides a recommended and tested version on our
CI.

Remaining TODOs:
- [x] Should envrc / direnv do a check to see if the flutter version matches expectations?
- [x] Add Android Studio setup instructions

Closes #15 

---

### Pros
* One fewer dependency and installation since it just depends on git
  * One less thing to learn/break.
* Does not depend on fvm which seems a bit clunky
* Easy to make contributions to flutter with the submodule setup once you find your way around submodules
* Few changes needed to CI and current setup process. 
* `flutter upgrade` just works™️ , just need to commit the change of what the submodule commit is pointing to, which is simple.
* git tells us if the submodule is in the clean state

### Cons
* Users need to run additional commands to setup submodule (This ends up being similar to commands to fvm, but we're not dealing with fvm's issues)
* Somewhat annoying that if you make any changes in the submodule, your work tree will be deemed "dirty". Perhaps a non-issue if there's some git command workaround.
  * Related https://stackoverflow.com/questions/31871888/what-goes-wrong-when-using-git-worktree-with-git-submodules indicates that there are some incompatibilities with `git worktree`.

### DIFF
--- a/.github/workflows/ci-flutter-main.yml
+++ b/.github/workflows/ci-flutter-main.yml
@@ -1,0 +1,33 @@
+# The `check-flutter-main` job serves 2 purposes:
+# 1. It checks that the code works with the flutter's latest main
+#    channel.
+# 2. It checks that tools/check and our related scripts don't embed
+#    any hidden assumptions that the Flutter SDK lives at vendor/flutter/.
+name: CI
+
+on: push
+
+jobs:
+  check-flutter-main:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Clone Flutter SDK
+      # We can't do a depth-1 clone, because we need the most recent tag
+      # so that Flutter knows its version and sees the constraint in our
+      # pubspec is satisfied.  It's uncommon for flutter/flutter to go
+      # more than 100 commits between tags.  Fetch 1000 for good measure.
+      run: |
+        git clone --depth=1000 https://github.com/flutter/flutter ~/flutter
+        TZ=UTC git --git-dir ~/flutter/.git log -1 --format='%h | %ci | %s' --date=iso8601-local
+        echo ~/flutter/bin >> "$GITHUB_PATH"
+
+    - name: Download Flutter SDK artifacts (flutter precache)
+      run: flutter precache --universal
+
+    - name: Download our dependencies (flutter pub get)
+      run: flutter pub get
+
+    - name: Run tools/check
+      run: TERM=dumb tools/check --all --verbose
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,17 +6,25 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - name: Check out repository
+      uses: actions/checkout@v3
+      # We're not using `with: submodules` here because we can't do a
+      # depth=1 clone. See below for more details.
 
-    - name: Clone Flutter SDK
+    - name: Checkout flutter submodule
       # We can't do a depth-1 clone, because we need the most recent tag
       # so that Flutter knows its version and sees the constraint in our
       # pubspec is satisfied.  It's uncommon for flutter/flutter to go
-      # more than 100 commits between tags.  Fetch 1000 for good measure.
+      # more than 100 commits between tags. Fetch 1000 for good measure.
       run: |
-        git clone --depth=1000 https://github.com/flutter/flutter ~/flutter
-        TZ=UTC git --git-dir ~/flutter/.git log -1 --format='%h | %ci | %s' --date=iso8601-local
-        echo ~/flutter/bin >> "$GITHUB_PATH"
+        git submodule update --init --depth 1000
+
+    - name: Add vendored flutter to PATH
+      run: |
+        echo vendor/flutter/bin >> "$GITHUB_PATH"
+
+    - name: Run setup
+      run: tools/setup-vendor-flutter
 
     - name: Download Flutter SDK artifacts (flutter precache)
       run: flutter precache --universal
--- a/README.md
+++ b/README.md
@@ -97,7 +97,6 @@ Two specific points to expand on:
     1.1. Initialize the pinned Flutter tree:
 
     ```sh
-    git submodule update --init
     tools/setup-vendor-flutter
     ```
 
--- a/README.md
+++ b/README.md
@@ -94,12 +94,10 @@ Two specific points to expand on:
 
 1. Setup the pinned version of flutter.
 
-    1.1. Make sure you have initialized submodule by doing:
+    1.1. Initialize the pinned Flutter tree:
 
     ```sh
-    # Update and initialize submodule
     git submodule update --init
-    # Run setup-vendor-flutter
     tools/setup-vendor-flutter
     ```
 
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Two specific points to expand on:
     1.2. Install [direnv][direnv], and set up its hook:
 
     ```sh
-    # On Debian or Ubuntu (bash):
+    # On Debian or Ubuntu (with Bash):
     sudo apt install direnv && echo 'eval "$(direnv hook bash)"' >>~/.bashrc
 
     # Or using Homebrew:
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Two specific points to expand on:
 
 ### Setting up
 
-1. Setup the pinned version of flutter.
+1. Set up the pinned version of Flutter.
 
     1.1. Initialize the pinned Flutter tree:
 
--- a/README.md
+++ b/README.md
@@ -92,26 +92,75 @@ Two specific points to expand on:
 
 ### Setting up
 
-1. Follow the [Flutter installation guide](https://docs.flutter.dev/get-started/install)
-   for your platform of choice.
-2. Switch to the latest version of Flutter by running `flutter channel main`
-   and `flutter upgrade` (see [Flutter version](#flutter-version) below).
-3. Ensure Flutter is correctly configured by running `flutter doctor`.
-4. Start the app with `flutter run`, or from your IDE.
+1. Setup the pinned version of flutter.
+
+    1.1. Make sure you have initialized submodule by doing:
+
+    ```sh
+    # Update and initialize submodule
+    git submodule update --init
+    # Run setup-vendor-flutter
+    tools/setup-vendor-flutter
+    ```
+
+    1.2. Install [direnv][direnv], and set up its hook:
+
+    ```sh
+    # On Debian or Ubuntu (bash):
+    sudo apt install direnv && echo 'eval "$(direnv hook bash)"' >>~/.bashrc
+
+    # Or using Homebrew:
+    brew install direnv
+
+    # Or see: https://direnv.net/docs/installation.html
+    # Then: https://direnv.net/docs/hook.html
+
+    # After installing and setting up the hook, run this in the zulip-flutter
+    # directory to allow the .envrc file.
+    direnv allow
+    ```
+
+   1.3. Follow remaining instructions in [Flutter installation guide](https://docs.flutter.dev/get-started/install)
+   for your platform of choice. You can skip the Flutter SDK installation as that is now done.
+
+2. Ensure Flutter is correctly configured by running `flutter doctor`.
+3. Start the app with `flutter run`, or from your IDE.
+   If you're using VSCode, you're set.
+   If you're using Android Studio, please read the next step.
+
+   3.1 For Android Studio users only (unless you are using your own
+       custom version of flutter):
+
+   After you first open, the project, go to:
+     `Settings -> Languages & Frameworks -> Flutter`
+   Under `Flutter SDK Path`, enter `<REPO_DIR>/vendor/flutter` (where
+   `<REPO_DIR>` is the location of the checkout of the repo).
 
 
 ### Flutter version
 
-While in the beta phase, we use the latest Flutter from Flutter's
-main branch.  Use `flutter channel main` and `flutter upgrade`.
+We pin to a particular version of flutter SDK via git submodules in
+[vendor/flutter/](vendor/flutter/).
+If your local checkout of this repository does not have this submodule
+checked out at the same commit, the build may fail.
 
-We don't pin a specific version, because Flutter itself doesn't offer
-a way to do so.  So far that hasn't been a problem.  When it becomes one,
-we'll figure it out; there are several tools for this in the Flutter
-community.  See [issue #15][].
+However, if you want to manage your own flutter SDK version you can
+opt out of this behavior by either of:
+- Skip installing direnv or don't do `direnv allow`.
+- Comment out the lines in `.envrc`.
 
-[issue #15]: https://github.com/zulip/zulip-flutter/issues/15
+Do note that if you do this, you need to manually make sure that the
+flutter SDK version matches or is compatible with the version
+indicated by the submodule commit SHA.
 
+Otherwise, the build can fail as we have not tested the current code
+with that particular flutter SDK version.
+
+If you want manage your own flutter SDK version, follow the [Flutter
+installation guide](https://docs.flutter.dev/get-started/install) for
+your platform of choice.
+
+[direnv]: https://direnv.net/
 
 ### Tests
 
@@ -242,13 +291,16 @@ that's a good prompt to do this.  We also do this when there's a
 new PR merged that we particularly want to take.
 
 To update the version bounds:
+* First, make sure you're on the `main` channel, run:
+  `flutter channel main`.
 * Use `flutter upgrade` to upgrade your local Flutter and Dart.
 * Update the lower bounds at `environment` in `pubspec.yaml`
   to the new versions, as seen in `flutter --version`.
 * Run `flutter pub get`, which will update `pubspec.lock`.
 * Make a quick check that things work: `tools/check`,
   and do a quick smoke-test of the app.
-* Commit and push the changes in `pubspec.yaml` and `pubspec.lock`.
+* Commit and push the changes in the submodule `vendor/flutter`,
+  `pubspec.yaml` and `pubspec.lock`.
 
 
 ### Upgrading dependencies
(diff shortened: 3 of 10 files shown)


chrisirhc:
IMO this ends up being cleaner and still meets the goals of #15 .

gnprice:
Thanks! I agree ­— I think this plan looks good. The main thing it needs is some more instructions, in particular on that Android Studio / IntelliJ setup we discussed around https://github.com/zulip/zulip-flutter/pull/577#issuecomment-2016431596 .

The instructions for upgrading Flutter will also need some playtesting — I suspect the `flutter upgrade` step will want some kind of adjustment.

---

One peculiarity I notice, from playing with this, is that `git submodule` checks out the submodule at a detached HEAD, rather than at a named branch like `main`, and as a result `flutter --version` can't tell what channel it's coming from:
```
$ vendor/flutter/bin/flutter --version
Flutter 3.21.0-12.0.pre.26 • channel [user-branch] • unknown source
[…]
```

That contrasts with, for my existing main Flutter tree at a slightly older version:
```
$ ~/n/flutter/flutter/bin/flutter --version
Flutter 3.21.0-12.0.pre.29 • channel main • https://github.com/flutter/flutter.git
[…]
```

Not sure what-all downstream effects that might have. If it sufficiently confuses things, we might need to find a way to coax `git submodule` to keep things on a branch called `main`, rather than on detached HEADs. That or have a little script in `tools/` that runs `git submodule update --init` and then fixes things up (e.g. with `git checkout -B main` in the submodule), or something like that.

Hmm, in fact, here's one downstream effect:
```
$ vendor/flutter/bin/flutter upgrade
Unable to upgrade Flutter: Your Flutter checkout is currently not on a release branch.
Use "flutter channel" to switch to an official channel, and retry. Alternatively, re-install Flutter by going to https://flutter.dev/docs/get-started/install.
```

I think taking `flutter upgrade` out of the instructions would be a fine solution to that — instead we can say something like `git -C vendor/flutter pull --ff-only`. We'll just need some kind of instructions that have been playtested to work.

chrisirhc:
By adding a manual setup step to do `git -C vendor/flutter checkout -B main` so that the submodule is using a local `main` branch, we can let flutter it's on the main channel. This only needs to be done once. Subsequently, everything runs as expected with `flutter upgrade`. I've updated this PR to address that.

I'm now following up on the Android Studio investigation but it does look like manual instructions are needed.
Will updated the description of this PR shortly.

README.md
```diff
@@ -92,26 +92,74 @@ Two specific points to expand on:
 
 ### Setting up
 
-1. Follow the [Flutter installation guide](https://docs.flutter.dev/get-started/install)
-   for your platform of choice.
```
gnprice, line 96: I think this still belongs in the instructions, because it sets up other pieces like the IDE and the Android or iOS build stack. But it can go after setting up the pinned Flutter, so that it can use that version.

README.md
```diff
@@ -92,26 +92,74 @@ Two specific points to expand on:
...
-2. Switch to the latest version of Flutter by running `flutter channel main`
-   and `flutter upgrade` (see [Flutter version](#flutter-version) below).
-3. Ensure Flutter is correctly configured by running `flutter doctor`.
-4. Start the app with `flutter run`, or from your IDE.
+1. Setup the pinned version of flutter.
+   (See [Flutter version](#flutter-version) for details on what this
+    step does and read it if you want to use your own version of
+    flutter.)
```
gnprice, line 98: Let's leave this part out — it's basically only ever a good idea if you've already been working in this repo (and want to upgrade the pinned version, or develop an upstream change while testing against zulip-flutter), so no need to mention it up front.

README.md
```diff
@@ -92,26 +92,74 @@ Two specific points to expand on:
...
+
+    1.1. Make sure you have initialized submodule by doing (do
+  not do `--depth=1` as flutter relies on tags to know its version):
```
gnprice, line 101: [[[WP:BEANS]]](https://en.wikipedia.org/wiki/WP:BEANS) — simpler just not to mention `--depth` :slightly_smiling_face:

.github/workflows/ci-flutter-main.yml
```diff
@@ -0,0 +1,33 @@
+# The `check-flutter-main` job serves 2 purposes:
+# 1. It checks that the code works with the flutter's latest main
+#    channel.
+# 2. It checks that our code environment setup works with a typical
+#    Flutter SDK installation setup (via ~/flutter).
```
gnprice, line 5: I started writing a comment saying I wasn't convinced this second point was really testing anything, but then decided it is after all. I don't think "our code environment setup" quite captures it, though — that sounds to me like things like `tools/setup-vendor-flutter` that are precisely what isn't getting exercised here. So here's a rewording:

```suggestion
# 2. It checks that tools/check and our related scripts don't embed
#    any hidden assumptions that the Flutter SDK lives at vendor/flutter/.
```

tools/setup-vendor-flutter
```diff
@@ -0,0 +1,19 @@
...
+# One-time setup to let git submodule + flutter SDK know that we're
+# using the main channel (main branch).
+# We should do a quick check that there's no changes.
+if ! submodule_is_clean; then
+    echo >&2 "There are changes in the submodule or it is not initialized."
+    echo >&2 "Please run 'git submodule update --init' then run this script again."
+    exit 1
+fi
+
+git -C vendor/flutter checkout -B main HEAD
```
gnprice, line 19: Have you found any behavior that's affected if we leave this out, other than `flutter upgrade`?

If not, it'd be good to leave it out of the initial setup instructions — one less step to take, and in particular brings it back down to zero steps that are entirely our own custom scripts instead of things that are widely used and may be familiar. Then it can just be part of the instructions around `flutter upgrade` instead, which is something a much smaller number of our contributors need to think about.

gnprice (commented):
Thanks! Generally this looks great. Various comments below.

tools/setup-vendor-flutter
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
chrisirhc: Whoops, bad rebase.

.github/workflows/ci.yml
```diff
@@ -6,17 +6,29 @@ jobs:
...
+    - name: Add vendored flutter to PATH
+      run: |
+        echo vendor/flutter/bin >> "$GITHUB_PATH"
+
+    - name: Run setup
+      run: tools/setup-vendor-flutter
+
+    - name: Display version (Temporary)
+      run: |
+        flutter --version
```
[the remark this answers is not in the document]
chrisirhc: Ah forgot to remove it.

tools/setup-vendor-flutter
chrisirhc: Yes, I found that `flutter doctor` complains that the flutter SDK isn't set up correctly. And since `flutter doctor` would be used by newcomers a lot, it would be confusing to those users.
For reference, the output from `flutter doctor`:
```
[!] Flutter (Channel [user-branch], 3.21.0-12.0.pre.25, on macOS 14.4 23E214 darwin-x64, locale en-SG)
    ! Flutter version 3.21.0-12.0.pre.25 on channel [user-branch] at /Users/chris/code/zulip-flutter/vendor/flutter
      Currently on an unknown channel. Run `flutter channel` to switch to an official channel.
      If that doesn't fix the issue, reinstall Flutter by following instructions at
      https://flutter.dev/docs/get-started/install.
    ! Upstream repository unknown source is not a standard remote.
      Set environment variable "FLUTTER_GIT_URL" to unknown source to dismiss this error.
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 15.3)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2023.2)
[✓] IntelliJ IDEA Community Edition (version EAP IC-241.14494.17)
[✓] IntelliJ IDEA Ultimate Edition (version EAP IU-241.14494.17)
[✓] VS Code (version 1.87.2)
[✓] VS Code (version 1.81.0-insider)
[✓] Connected device (4 available)
[✓] Network resources
! Doctor found issues in 1 category.
```

chrisirhc:
Updated the PR and addressed the comments.

gnprice:
Thanks again @chrisirhc for all your work on this! Sorry I haven't managed to return for another review for a bit — we've just this week gotten past the peak PR volume from GSoC application season.

I'll be going [on vacation](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Greg.20vacation/near/1768574) shortly for the next couple of weeks. I'll aim to pick this back up the week I'm back, as I catch up with things. Based on where it was at my last review, I imagine we'll be pretty close to merge. This will be a very nice improvement to have!

tools/lib/git.sh
```diff
@@ -90,3 +90,13 @@ git_base_commit() {
 git_changed_files() {
     git diff --name-only --diff-filter=d "$@"
 }
+
+# Note that this assumes ONE submodule. If we ever have more than one,
+# we'll need to generalize this.
```
gnprice, line 95: This comment no longer applies, right? (After the revision corresponding to https://github.com/zulip/zulip-flutter/pull/579#discussion_r1540345664 .)

tools/setup-vendor-flutter
```diff
@@ -0,0 +1,19 @@
...
+
+# shellcheck source=tools/lib/git.sh
+. "${this_dir}"/lib/git.sh
+
+
+# One-time setup to let git submodule + flutter SDK know that we're
+# using the main channel (main branch).
+# We should do a quick check that there's no changes.
+if ! submodule_is_clean "vendor/flutter" ; then
+    echo >&2 "There are changes in 'vendor/flutter' tree or it is not initialized."
```
gnprice, line 14: grammar nit:
```suggestion
    echo >&2 "There are changes in the 'vendor/flutter' tree or it is not initialized."
```

(or some other rewording)

README.md
```diff
@@ -92,26 +92,75 @@ Two specific points to expand on:
...
+    # Update and initialize submodule
+    git submodule update --init
+    # Run setup-vendor-flutter
+    tools/setup-vendor-flutter
+    ```
+
+    1.2. Install [direnv][direnv], and set up its hook:
+
+    ```sh
+    # On Debian or Ubuntu (bash):
```
gnprice, line 109:
```suggestion
    # On Debian or Ubuntu (with Bash):
```

With the terser version, I'm not sure quite how to parse the relationship between "Debian", "Ubuntu", and "bash".

README.md
```diff
@@ -92,26 +92,75 @@ Two specific points to expand on:
 
 ### Setting up
 
-1. Follow the [Flutter installation guide](https://docs.flutter.dev/get-started/install)
-   for your platform of choice.
-2. Switch to the latest version of Flutter by running `flutter channel main`
-   and `flutter upgrade` (see [Flutter version](#flutter-version) below).
-3. Ensure Flutter is correctly configured by running `flutter doctor`.
-4. Start the app with `flutter run`, or from your IDE.
+1. Setup the pinned version of flutter.
```
gnprice, line 95:
```suggestion
1. Set up the pinned version of Flutter.
```

(similar to https://github.com/zulip/zulip-flutter/pull/579#discussion_r1540337383)

There's also an occurrence of this in `.envrc`.

README.md
```diff
@@ -92,26 +92,75 @@ Two specific points to expand on:
...
+
+    1.1. Make sure you have initialized submodule by doing:
+
+    ```sh
+    # Update and initialize submodule
+    git submodule update --init
+    # Run setup-vendor-flutter
+    tools/setup-vendor-flutter
+    ```
```
gnprice, line 104: Similar to https://github.com/zulip/zulip-flutter/pull/579#discussion_r1540348954:
> Let's try to write this message with the audience in mind of someone who's just cloned this repo and is starting to go through the README, and has never heard of Git submodules (and even if they have, doesn't know our particular layout or how we're using submodules).

So here:
```suggestion
    1.1. Initialize the pinned Flutter tree:

    ```sh
    git submodule update --init
    tools/setup-vendor-flutter
    ```
```

gnprice: Relatedly: can we move the `git submodule update --init` inside the script, so that this step is just a single command?

Hmm, though on the other hand, and zooming out a bit: what happens when as a contributor you rebase (or reset to a new `origin/main`), and the version that's pinned has changed? You'll need to run some `git submodule` command to update your local clone, right?

If so, that seems like something we should mention in this doc too. But it also points to keeping an explicit `git submodule` step here in the setup after all, to avoid having something that's initially magic but where the developer later needs to look behind the magic anyway.

tools/setup-vendor-flutter
gnprice: Cool, that reasoning makes sense to me.

gnprice (commented):
Thanks @chrisirhc! Various small comments below. There's also one substantive question (about the resulting developer workflow) which I'll copy here for visibility:

What happens when as a contributor you rebase (or reset to a new `origin/main`), and the version that's pinned has changed? You'll need to run some `git submodule` command to update your local clone, right?

In fact, when I look at `git help submodule`, I'm not sure what an appropriate command would be. One obvious candidate is `git submodule update`… but I think that will take the submodule onto a detached HEAD, same as it does on first setup (as discussed above starting at https://github.com/zulip/zulip-flutter/pull/579#issuecomment-2017334691).

In that case we probably end up recommending the developer run a script like `tools/setup-vendor-flutter` any time `git status` tells them `vendor/flutter` is out of sync, or some sort of instruction like that.

---

Once we have all the phases of the main contributor workflow (so upgrade/rebase as well as setup) shaken out, I'll make one more editing pass over all the docs; and then I think it'll be time to play-test the instructions.

I'll handle that by setting up a branch with this PR plus an ad-hoc Flutter upgrade, in order to simulate that; then asking people to switch to the PR branch, follow the README instructions, then to the upgrade branch, and to test things work and give feedback on anything confusing.

gnprice:
I've also rebased the PR branch, resolving a small merge conflict (foreshadowed at https://github.com/zulip/zulip-flutter/pull/579#discussion_r1540336928 :slightly_smiling_face: ) and updating the pinned version.

README.md
chrisirhc: Ah,yes, in my latest changes I've explored moving `git submodule update --init` into the `setup-vendor-flutter` script and making this script runnable any/all the time. If it's already set up, it'll be a noop.

chrisirhc:
Updated the PR and addressed the comments though I haven't squashed the commits yet.

One other thing to consider is to add a `post-checkout` git hook that runs `setup-vendor-flutter` whenever user does a checkout. 
I don't want to add too much magic or over-scope this work, so leaving that out for now, but it might be something to consider based on feedback/play-testing.

A quick preview of the message if the setup-vendor-flutter fails:
```
chris@Chris-Slate ~/c/zulip-flutter (chua/test-change-branch) [1]> ./tools/setup-vendor-flutter                                             (base) 
Initializing 'vendor/flutter' submodule
running 'git submodule update --init' to initialize it:
----------------------------------------
error: Your local changes to the following files would be overwritten by checkout:
        bin/internal/engine.version
Please commit your changes or stash them before you switch branches.
Aborting
fatal: Unable to checkout 'a46a7fdc3d622908c24c509ae48c874368c7eb12' in submodule path 'vendor/flutter'
----------------------------------------
Failed to initialize the 'vendor/flutter' submodule, please resolve the issues indicated above.
```

gnprice:
Thanks again @chrisirhc for all your work exploring this design space with me. Sorry I haven't found time to return to this and do the instructions-editing and play-testing I described at https://github.com/zulip/zulip-flutter/pull/579#pullrequestreview-2032654008 , or to otherwise think through completing it.

I'm closing this PR to reflect that reality, and the fact that I don't expect to find time for this for at least the next few months (as we [focus on getting the app launched](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556)). The changes in the PR will remain available in Git, and I'll make a note over on the issue that this remains the place to start from when we pick this up again in the future.

PR #581, opened 2024-03-24
subscriptions: Fix stream name sorting case insensitively

### LINKED ISSUE #568: Sort stream names case-insensitively in "Subscribed streams"
Bug report [in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Case-sensitive.20stream.20sorting/near/1757094) (thanks @alexmv):
> In the Flutter app, it looks like streams are sorted case-sensitively (i.e. `GSOC` above `announce`) in the "Subscribed streams" list -- but not the "Inbox".  The React mobile app and webapp also sort case-insensitively.

We should fix that.

Khader-1:
Stream names should be sorted regardless of the case and this applies for pinned and unpinned subscriptions:
**A comes before b and B comes after a**
    
So this adds simple toLowerCase transformation before sorting
Fixes #568

### DIFF
--- a/test/widgets/subscription_list_test.dart
+++ b/test/widgets/subscription_list_test.dart
@@ -120,6 +120,18 @@ void main() {
       ]);
       check(listedStreamIds(tester)).deepEquals([2, 3, 1]);
     });
+
+    testWidgets('subscriptions sorting is case insensitive', (tester) async {
+      await setupStreamListPage(tester, subscriptions: [
+        eg.subscription(eg.stream(streamId: 1, name: 'a'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 2, name: 'B'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 3, name: 'c'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 4, name: 'D'), pinToTop: false),
+        eg.subscription(eg.stream(streamId: 5, name: 'e'), pinToTop: false),
+        eg.subscription(eg.stream(streamId: 6, name: 'F'), pinToTop: false),
+      ]);
+      check(listedStreamIds(tester)).deepEquals([1, 2, 3, 4, 5, 6]);
+    });
   });
 
   testWidgets('unread badge shows with unreads', (tester) async {
--- a/test/widgets/subscription_list_test.dart
+++ b/test/widgets/subscription_list_test.dart
@@ -89,18 +89,37 @@ void main() {
     check(isUnpinnedHeaderInTree()).isTrue();
   });
 
-  testWidgets('subscription sort', (tester) async {
-    await setupStreamListPage(tester, subscriptions: [
-      eg.subscription(eg.stream(streamId: 1, name: 'd'), pinToTop: true),
-      eg.subscription(eg.stream(streamId: 2, name: 'c'), pinToTop: false),
-      eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: true),
-      eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: false),
-    ]);
-    check(isPinnedHeaderInTree()).isTrue();
-    check(isUnpinnedHeaderInTree()).isTrue();
+  group('subscription sorting', () {
+    Iterable<int> listedStreamIds(WidgetTester tester) => tester
+      .widgetList<SubscriptionItem>(find.byType(SubscriptionItem))
+      .map((e) => e.subscription.streamId);
 
-    final streamListItems = tester.widgetList<SubscriptionItem>(find.byType(SubscriptionItem)).toList();
-    check(streamListItems.map((e) => e.subscription.streamId)).deepEquals([3, 1, 4, 2]);
+    testWidgets('pinned are shown on the top', (tester) async {
+      await setupStreamListPage(tester, subscriptions: [
+        eg.subscription(eg.stream(streamId: 1, name: 'a'), pinToTop: false),
+        eg.subscription(eg.stream(streamId: 2, name: 'b'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 3, name: 'c'), pinToTop: false),
+      ]);
+      check(listedStreamIds(tester)).deepEquals([2, 1, 3]);
+    });
+
+    testWidgets('pinned subscriptions are sorted', (tester) async {
+      await setupStreamListPage(tester, subscriptions: [
+        eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 1, name: 'c'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 2, name: 'a'), pinToTop: true),
+      ]);
+      check(listedStreamIds(tester)).deepEquals([2, 3, 1]);
+    });
+
+    testWidgets('unpinned subscriptions are sorted', (tester) async {
+      await setupStreamListPage(tester, subscriptions: [
+        eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: false),
+        eg.subscription(eg.stream(streamId: 1, name: 'c'), pinToTop: false),
+        eg.subscription(eg.stream(streamId: 2, name: 'a'), pinToTop: false),
+      ]);
+      check(listedStreamIds(tester)).deepEquals([2, 3, 1]);
+    });
   });
 
   testWidgets('unread badge shows with unreads', (tester) async {
--- a/lib/widgets/subscription_list.dart
+++ b/lib/widgets/subscription_list.dart
@@ -77,8 +77,8 @@ class _SubscriptionListPageState extends State<SubscriptionListPage> with PerAcc
       }
     }
     // TODO(i18n): add locale-aware sorting
-    pinned.sortBy((subscription) => subscription.name);
-    unpinned.sortBy((subscription) => subscription.name);
+    pinned.sortBy((subscription) => subscription.name.toLowerCase());
+    unpinned.sortBy((subscription) => subscription.name.toLowerCase());
 
     return Scaffold(
       appBar: AppBar(title: const Text("Streams")),


Khader-1:
Commit restructured
Thanks @sm-sayedi

Khader-1:
Commit restructured
Thanks @sm-sayedi

test/widgets/subscription_list_test.dart
```diff
@@ -92,8 +92,8 @@ void main() {
   testWidgets('subscription sort', (tester) async {
     await setupStreamListPage(tester, subscriptions: [
       eg.subscription(eg.stream(streamId: 1, name: 'd'), pinToTop: true),
-      eg.subscription(eg.stream(streamId: 2, name: 'c'), pinToTop: false),
-      eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: true),
+      eg.subscription(eg.stream(streamId: 2, name: 'C'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 3, name: 'B'), pinToTop: true),
       eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: false),
```
gnprice, line 97: The main purpose of a test is to make sure we don't accidentally revert your change in the future.

So when writing tests, it's important to check that if you undo your changes, the test would actually fail (and therefore succeed at its purpose). Similarly if you try breaking your changes in a partial way.

Here, there are two `toLowerCase` calls you add. Right now they're on consecutive lines and obviously parallel, and it's unlikely that in that form we'd ever remove one of them without the other — but in the future as things get more complex, they may become separated by many lines of code, and then it becomes entirely likely that'd they'd evolve out of sync with each other. So the test should successfully catch either one of the two `toLowerCase` calls going missing.

As written, I believe this test would catch only one, and not the other.

gnprice: Probably the best way to do this robustly is to add a new test case, rather than edit this one. When one test is testing several different aspects of things, it becomes difficult to be sure what they all are, and therefore difficult to update the test correctly.

For example, it's already hard to tell by looking at this change whether it defeats one of the existing checks this test was intended to make. That's a sign that this test is already checking too many things.

So let's take this in two steps:
* First, one commit splits up this existing test case: one test case to check that the pinned and unpinned sections each get sorted, and one to check that pinned goes before unpinned.
* Then, one commit makes your change to the lib code, and adds a new test case that checks that lower-case vs. upper-case letters are treated the same.

gnprice (commented):
Thanks @Khader-1 for the contribution!

We'll need some more work to properly test this; details below.

Please see also the Zulip style guide for commit messages:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-messages
and make a go at revising your commit message into the Zulip style.

Khader-1:
I didn't expect to learn so many things in such a small PR!
Thanks indeed, I appreciate the informative guiding feedback😊
Still there somethings that I am not so sure about in my commits:
- regarding the commit message summary Is there a document or a diagram that I can refer to regarding the ubiquitous language used in Zulip?

gnprice:
> * regarding the commit message summary Is there a document or a diagram that I can refer to regarding the ubiquitous language used in Zulip?

The doc I linked above is the writeup we have about how to write a commit message:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-messages

But I'm not sure I understand your question. If you have questions about commit messages which that doc doesn't answer, a thread on chat.zulip.org would be the best place to ask — can always ask in `#general`, and depending on the nature of the question there might be a better stream someone will move it to.

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,71 @@ void main() {
...
+    await setupStreamListPage(tester, subscriptions: [
+      eg.subscription(eg.stream(streamId: 1, name: 'd'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 2, name: 'c'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: false),
       eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: false),
     ]);
+    check(isUnpinnedHeaderInTree()).isTrue();
+
+    final streamListItems = tester.widgetList<SubscriptionItem>(find.byType(SubscriptionItem)).toList();
+    check(streamListItems.map((e) => e.subscription.streamId)).deepEquals([4, 3, 2, 1]);
```
gnprice, line 128: When making a larger number of test cases, one aspect that's key for keeping it easy to understand the tests and maintain them is to keep each test case short — focused and concise, with only the details that are important to making that test case work.

So in particular, now that each of these test cases is only testing one thing, they don't need four items in the list — I think two items will make the point just as well.

Or three: like b, a, c -> a, b, c, to ensure that the test doesn't pass if we end up with a bug where we're forgetting to sort but happen to show things in reverse order relative to the list in the input data.

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,71 @@ void main() {
...
+  });
+
+  testWidgets('unpinned subscriptions are sorted', (tester) async {
+    await setupStreamListPage(tester, subscriptions: [
+      eg.subscription(eg.stream(streamId: 1, name: 'd'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 2, name: 'c'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: false),
       eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: false),
     ]);
+    check(isUnpinnedHeaderInTree()).isTrue();
```
gnprice, line 125: Also as part of keeping each test focused, there's no need to add this line here. The headers are checked in a different pair of test cases above. (And they're checked more thoroughly there, checking both absence and presence.)

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,71 @@ void main() {
...
+    check(streamListItems.map((e) => e.subscription.streamId)).deepEquals([4, 3, 2, 1]);
+  });
+
+  testWidgets('pinned subscriptions are sorted case insensitively', (tester) async {
+    await setupStreamListPage(tester, subscriptions: [
+      eg.subscription(eg.stream(streamId: 1, name: 'D'), pinToTop: true),
+      eg.subscription(eg.stream(streamId: 2, name: 'c'), pinToTop: true),
+      eg.subscription(eg.stream(streamId: 3, name: 'B'), pinToTop: true),
+      eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: true),
+    ]);
```
gnprice, line 111: The case-insensitive part might be easiest to read if it's consolidated as one test case with four items: two pinned, and two unpinned, in order to exercise both.

(Or I guess six items, given the other point I mentioned.)

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,71 @@ void main() {
...
+    await setupStreamListPage(tester, subscriptions: [
+      eg.subscription(eg.stream(streamId: 1, name: 'D'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 2, name: 'c'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 3, name: 'B'), pinToTop: false),
+      eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: false),
+    ]);
+    check(isUnpinnedHeaderInTree()).isTrue();
+
+    final streamListItems = tester.widgetList<SubscriptionItem>(find.byType(SubscriptionItem)).toList();
+    check(streamListItems.map((e) => e.subscription.streamId)).deepEquals([4, 3, 2, 1]);
```
gnprice, line 141: Another part of keeping each test case focused and concise would be to factor out the definition of `streamListItems`, plus the map to stream IDs, as a little helper function. Then each of these test cases can just say something like:
```dart
    check(listedStreamIds(tester)).deepEquals([4, 3, 2, 1]);
```

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,71 @@ void main() {
     check(isUnpinnedHeaderInTree()).isTrue();
   });
 
-  testWidgets('subscription sort', (tester) async {
+  testWidgets('pinned subscriptions are sorted', (tester) async {
```
gnprice, line 92: It'd probably be helpful for the organization of this code (for the reader to see the structure of it) to group all these sorting test cases into a `group` call — like `group('sorting', …`.

gnprice (commented):

test/widgets/subscription_list_test.dart
Khader-1: Yes sure, additionally if I am getting it right, as we tested before that:
* pinned comes before unpinned streams
* both pinned and unpinned streams are sorted
I believe there will be no need to shuffle the streams this time as that would not be relevant to what we test (**that letter case does not matter in stream name sorting**) and that's why we refactored the tests in the first place. 
so something like this:
```dart
    testWidgets('subscriptions sorting is case insensitive', (tester) async {
      await setupStreamListPage(tester, subscriptions: [
        eg.subscription(eg.stream(streamId: 1, name: 'a'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 2, name: 'B'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 3, name: 'c'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 4, name: 'D'), pinToTop: false),
        eg.subscription(eg.stream(streamId: 5, name: 'e'), pinToTop: false),
        eg.subscription(eg.stream(streamId: 6, name: 'F'), pinToTop: false),
      ]);

      check(streamListIds(tester)).deepEquals([1, 2, 3, 4, 5, 6]);
    });
```
Should be sufficient.

test/widgets/subscription_list_test.dart
Khader-1: I feel like the hardest thing about test code refactoring is to consider when and where to refactor. but I am improving already, Thank you.

test/widgets/subscription_list_test.dart
Khader-1: Yes sure, additionally if I am getting it right, as we tested before that:
* pinned comes before unpinned streams
* both pinned and unpinned streams are sorted
I believe there will be no need to shuffle the streams this time as that would not be relevant to what we test (**that letter case does not matter in stream name sorting**) and that's why we refactored the tests in the first place. 
so something like this:
```dart
    testWidgets('subscriptions sorting is case insensitive', (tester) async {
      await setupStreamListPage(tester, subscriptions: [
        eg.subscription(eg.stream(streamId: 1, name: 'a'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 2, name: 'B'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 3, name: 'c'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 4, name: 'D'), pinToTop: false),
        eg.subscription(eg.stream(streamId: 5, name: 'e'), pinToTop: false),
        eg.subscription(eg.stream(streamId: 6, name: 'F'), pinToTop: false),
      ]);

      check(streamListIds(tester)).deepEquals([1, 2, 3, 4, 5, 6]);
    });
```
Should be sufficient.

Khader-1: Yes sure, additionally if I am getting it right, as we tested before that:
* pinned comes before unpinned streams
* both pinned and unpinned streams are sorted
I believe there will be no need to shuffle the streams this time as that would not be relevant to what we test (**that letter case does not matter in stream name sorting**) and that's why we refactored the tests in the first place. 
so something like this:
```dart
    testWidgets('subscriptions sorting is case insensitive', (tester) async {
      await setupStreamListPage(tester, subscriptions: [
        eg.subscription(eg.stream(streamId: 1, name: 'a'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 2, name: 'B'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 3, name: 'c'), pinToTop: true),
        eg.subscription(eg.stream(streamId: 4, name: 'D'), pinToTop: false),
        eg.subscription(eg.stream(streamId: 5, name: 'e'), pinToTop: false),
        eg.subscription(eg.stream(streamId: 6, name: 'F'), pinToTop: false),
      ]);

      check(streamListIds(tester)).deepEquals([1, 2, 3, 4, 5, 6]);
    });
```
Should be sufficient.

test/widgets/subscription_list_test.dart
Khader-1: I feel like the hardest thing about test code refactoring is to consider when and where to refactor. but I am improving already, Thank you.

Khader-1: I feel like the hardest thing about test code refactoring is to consider when and where to refactor. but I am improving already, Thank you.

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,49 @@ void main() {
...
-      eg.subscription(eg.stream(streamId: 3, name: 'b'), pinToTop: true),
-      eg.subscription(eg.stream(streamId: 4, name: 'a'), pinToTop: false),
-    ]);
-    check(isPinnedHeaderInTree()).isTrue();
-    check(isUnpinnedHeaderInTree()).isTrue();
-
-    final streamListItems = tester.widgetList<SubscriptionItem>(find.byType(SubscriptionItem)).toList();
-    check(streamListItems.map((e) => e.subscription.streamId)).deepEquals([3, 1, 4, 2]);
+  group('subscription sorting', () {
+    streamListIds(WidgetTester tester) => tester
```
gnprice, line 93: nit: Write an explicit return type on function definitions:
```suggestion
    Iterable<int> streamListIds(WidgetTester tester) => tester
```

Otherwise the start of the line looks a lot like a function call.

test/widgets/subscription_list_test.dart
```diff
@@ -89,18 +89,49 @@ void main() {
...
+      .widgetList<SubscriptionItem>(find.byType(SubscriptionItem))
+      .map((e) => e.subscription.streamId);
+
+    testWidgets('pinned are shown on the top', (tester) async {
+      await setupStreamListPage(tester, subscriptions: [
+        eg.subscription(eg.stream(streamId: 1, name: 'a'), pinToTop: false),
+        eg.subscription(eg.stream(streamId: 2, name: 'b'), pinToTop: true),
+        eg.subscription(eg.stream(streamId: 3, name: 'c'), pinToTop: false),
+      ]);
+      check(streamListIds(tester)).deepEquals([2, 1, 3]);
```
gnprice, line 103: This name `streamListIds`, or "stream list IDs", sounds like it'd describe things that were IDs of "stream lists", or "list IDs" that were related to streams.

But in fact these are IDs of streams, namely of the streams shown in the list.

In the old code `streamListItems` worked as a name because each element was an item (a `SubscriptionItem`, in fact) in the stream list (the `StreamListPage`). So, a "stream list item".

For this code, one name that would work is `listedStreamIds` as I suggested at https://github.com/zulip/zulip-flutter/pull/581#discussion_r1540332942. (Each element of this return value is a stream ID, specifically the ID of a stream that's listed in this stream list.) So let's use that.

gnprice (commented):
Thanks @Khader-1 for the revision! This is quite close to merge now. Two small comments below, and one on the commit messages:

23e1ed508 subscriptions_list test [nfc]: Split subscriptions sort test case
caf47b29d subscription_list: Fix stream name sorting case insensitively

The two commits should agree on how to write the summary prefix :slightly_smiling_face: . They'll differ in that one of them says "test [nfc]" and the other doesn't, but the `subscription_list` vs. `subscriptions_list` part should match.

The names of the affected files say `subscription_list`, so let's match that spelling.

gnprice:

PR #582, opened 2024-03-24
autocomplete: Exclude deactivated users from @-mention autocomplete popup

### LINKED ISSUE #451: autocomplete: Exclude deactivated users from @-mention autocomplete popup, like in web
This does not involve changing the logic about when the `|user_id` syntax is included in the generated @-mention Markdown. In particular, if there are multiple users with the same `full_name`, but only one that's not deactivated, we should still include the disambiguating `|user_id` syntax when you tap that user in the autocomplete.

Khader-1:
### autocomplete: Exclude deactivated users from @-mention autocomplete popup
Deactivated users can not be mentioned, so this ensures that deactivated users are excluded from the list that gets rendered in autocomplete popup.

### compose test: Check ID still included if same-name users deactivated
While autocomplete does not include deactivated users, there still is a possibility that same-name users with one or more of them deactivated gets mentioned, which will make the mentioned user ambiguous without the id, this is explained in #451.

Fixes #451

### DIFF
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -287,6 +287,11 @@ class MentionAutocompleteQuery {
 
   bool testUser(User user, AutocompleteDataCache cache) {
     // TODO(#236) test email too, not just name
+
+    return _testName(user, cache);
+  }
+
+  bool _testName(User user, AutocompleteDataCache cache) {
     // TODO(#237) test with diacritics stripped, where appropriate
 
     final List<String> nameWords = cache.nameWordsForUser(user);
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -288,6 +288,8 @@ class MentionAutocompleteQuery {
   bool testUser(User user, AutocompleteDataCache cache) {
     // TODO(#236) test email too, not just name
 
+    if (!user.isActive) return false;
+
     return _testName(user, cache);
   }
 
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -59,6 +59,7 @@ User user({
   int? userId,
   String? email,
   String? fullName,
+  bool? isActive,
   String? avatarUrl,
   Map<int, ProfileFieldUserData>? profileData,
 }) {
@@ -68,7 +69,7 @@ User user({
     email: email ?? 'name@example.com', // TODO generate example emails
     fullName: fullName ?? 'A user', // TODO generate example names
     dateJoined: '2023-04-28',
-    isActive: true,
+    isActive: isActive ?? true,
     isOwner: false,
     isAdmin: false,
     isGuest: false,
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -312,6 +312,12 @@ void main() {
       expected ? check(result).isTrue() : check(result).isFalse();
     }
 
+    test('user is always excluded when not active regardless of other criteria', () {
+      doCheck('Full Name', eg.user(fullName: 'Full Name', isActive: false), false);
+      // When active then other criteria will be checked
+      doCheck('Full Name', eg.user(fullName: 'Full Name', isActive: true), true);
+    });
+
     test('user is included if fullname words match the query', () {
       doCheck('', eg.user(fullName: 'Full Name'), true);
       doCheck('', eg.user(fullName: ''), true); // Unlikely case, but should not crash
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -305,40 +305,42 @@ void main() {
     }
   });
 
-  test('MentionAutocompleteQuery.testUser', () {
+  group('MentionAutocompleteQuery.testUser', () {
     doCheck(String rawQuery, User user, bool expected) {
       final result = MentionAutocompleteQuery(rawQuery)
         .testUser(user, AutocompleteDataCache());
       expected ? check(result).isTrue() : check(result).isFalse();
     }
 
-    doCheck('', eg.user(fullName: 'Full Name'), true);
-    doCheck('', eg.user(fullName: ''), true); // Unlikely case, but should not crash
-    doCheck('Full Name', eg.user(fullName: 'Full Name'), true);
-    doCheck('full name', eg.user(fullName: 'Full Name'), true);
-    doCheck('Full Name', eg.user(fullName: 'full name'), true);
-    doCheck('Full', eg.user(fullName: 'Full Name'), true);
-    doCheck('Name', eg.user(fullName: 'Full Name'), true);
-    doCheck('Full Name', eg.user(fullName: 'Fully Named'), true);
-    doCheck('Full Four', eg.user(fullName: 'Full Name Four Words'), true);
-    doCheck('Name Words', eg.user(fullName: 'Full Name Four Words'), true);
-    doCheck('Full F', eg.user(fullName: 'Full Name Four Words'), true);
-    doCheck('F Four', eg.user(fullName: 'Full Name Four Words'), true);
-    doCheck('full full', eg.user(fullName: 'Full Full Name'), true);
-    doCheck('full full', eg.user(fullName: 'Full Name Full'), true);
+    test('user is included if fullname words match the query', () {
+      doCheck('', eg.user(fullName: 'Full Name'), true);
+      doCheck('', eg.user(fullName: ''), true); // Unlikely case, but should not crash
+      doCheck('Full Name', eg.user(fullName: 'Full Name'), true);
+      doCheck('full name', eg.user(fullName: 'Full Name'), true);
+      doCheck('Full Name', eg.user(fullName: 'full name'), true);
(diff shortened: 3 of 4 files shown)


sm-sayedi:
Better than before, but still needs some improvements.

> It should pass tests (so test updates needed by a change should be in the same commit as the original change, not a separate “fix the tests that were broken by the last commit” commit).

As stated in the [Zulip Commit Discipline](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html) page, the tests for a change should be in the same commit.

Also, instead of closing a PR and then opening a new one just for the sake of some changes, there is an easy way of doing it explained in [this](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#) page.

Khader-1:
> > It should pass tests (so test updates needed by a change should be in the same commit as the original change, not a separate “fix the tests that were broken by the last commit” commit).

@sm-sayedi  You're right although the tests in the second commit do not test the same change that the first one introduces. they depend on some of the changes.

Regarding the rebase.. I learned it hard way!

Thanks for you helpful guidance☺️

test/example_data.dart
```diff
@@ -61,14 +61,15 @@ User user({
   String? fullName,
   String? avatarUrl,
   Map<int, ProfileFieldUserData>? profileData,
+  bool? isActive,
```
gnprice, line 64: nit: keep these parameters in the same order as the order they're used in below

lib/model/autocomplete.dart
```diff
@@ -289,6 +289,8 @@ class MentionAutocompleteQuery {
     // TODO(#236) test email too, not just name
     // TODO(#237) test with diacritics stripped, where appropriate
 
+    if (!user.isActive) return false;
+
     final List<String> nameWords = cache.nameWordsForUser(user);
 
     int nameWordsIndex = 0;
```
gnprice, line 296: This function will get confusing if we start adding more and more features to it and include all the details in the same function.

Let's have a prep NFC commit that factors out a helper method `_testName`, with the same signature as `testUser`. Then the main commit can add the `user.isActive` check to `testUser`.

gnprice (commented):
Thanks @Khader-1 for the contribution, and @sm-sayedi for the initial reviews! Comments below on this version.

test/model/compose_test.dart
```diff
@@ -97,7 +97,7 @@ world
     test('whitespace around info string', () {
       checkFenceWrap('''
 ````
-``` javascript 
+``` javascript
```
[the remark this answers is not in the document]
Khader-1: Thanks for the feedback! I couldn't understand how this got updated in the first place! it was actually auto-formatted.  So yes I think we can put a comment here to make it explicit
image

Khader-1 (commented):
This has gone through many iterations! hopefully each iteration is a step in the right direction.
I think it is ready to be reviewed now

test/model/compose_test.dart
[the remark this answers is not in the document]
gnprice: I think we can make it a bit more explicit still, as well as more concise:
`````dart
      final infoString = ' javascript ';
      checkFenceWrap('''
````
```$infoString
`````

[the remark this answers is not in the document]
Comment (Khader-1) [reply test/model/compose_test.dart]: Thanks! This reads better and is actually explicit.
Applied now👍

File: lib/model/autocomplete.dart [hunk 4]
```diff
@@ -289,6 +289,12 @@ class MentionAutocompleteQuery {
     // TODO(#236) test email too, not just name
     // TODO(#237) test with diacritics stripped, where appropriate
```
gnprice, line 290: This TODO about diacritics belongs inside `_testName`, since it's about a change we would make within that function.

test/model/autocomplete_test.dart
```diff
@@ -305,40 +305,49 @@ void main() {
...
+      doCheck('Full ame', eg.user(fullName: 'Full Name'), false);
+      doCheck('Full Full', eg.user(fullName: 'Full Name'), false);
+      doCheck('Name Name', eg.user(fullName: 'Full Name'), false);
+      doCheck('Name Full', eg.user(fullName: 'Full Name'), false);
+      doCheck('Name Four Full Words', eg.user(fullName: 'Full Name Four Words'), false);
+      doCheck('F Full', eg.user(fullName: 'Full Name Four Words'), false);
+      doCheck('Four F', eg.user(fullName: 'Full Name Four Words'), false);
+    });
+
+    test('user is always excluded when not active regardless of other criteria', () {
```
gnprice, line 347: nit: put this test case above the other one, to match the order of the code being tested

(That helps make it easy to find the test corresponding to a given piece of application code, and vice versa.)

test/model/autocomplete_test.dart
```diff
@@ -305,40 +305,49 @@ void main() {
...
+      doCheck('Full Name', eg.user(fullName: 'Full Name', isActive: false), false);
+    });
+
   });
```
gnprice, line 351: nit: no blank line at end of block:
```suggestion
    });
  });
```

test/model/autocomplete_test.dart
```diff
@@ -305,40 +305,49 @@ void main() {
...
+      doCheck('Full Full', eg.user(fullName: 'Full Name'), false);
+      doCheck('Name Name', eg.user(fullName: 'Full Name'), false);
+      doCheck('Name Full', eg.user(fullName: 'Full Name'), false);
+      doCheck('Name Four Full Words', eg.user(fullName: 'Full Name Four Words'), false);
+      doCheck('F Full', eg.user(fullName: 'Full Name Four Words'), false);
+      doCheck('Four F', eg.user(fullName: 'Full Name Four Words'), false);
+    });
+
+    test('user is always excluded when not active regardless of other criteria', () {
+      doCheck('Full Name', eg.user(fullName: 'Full Name', isActive: false), false);
```
gnprice, line 348: Just to make this test case fully self-contained, let's give it a second line that looks exactly the same except the user is active, and does match.

That line will be identical to one of the lines in the existing test case, so in a sense it's redundant. But it makes a useful baseline for the reader to compare this other check to. It means that the reader can look at this test case and be confident that it really is testing what it's meant to test — that there isn't some accident where the user stopped matching because something changed about the name matching, or whatever — without having to look around at other tests, or to look back at the implementation that's being tested.

gnprice (commented):
Thanks @Khader-1! This is quite close to merge.

A few small comments below. The one other area that needs a bit of revision before we merge this is the formatting of the commit messages.
* For the formatting around the "[nfc]" marker, see examples at https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc .
* For the "fixes" line, see https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#fixes-format .
* See Zulip's Git style guide: https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#formatting-guidelines
  on line-wrapping for the prose in the body of your commit message.
* The commit summary can go slightly longer than 70 columns when there's a lot to express, but this one is 95 columns and longer than it needs to be:
  `autocomplete test[nfc]: Use a test group to handle MentionAutocompleteQuery.testUser test cases`
  * Saying just `testUser` instead of `MentionAutocompleteQuery.testUser` is the main part of a fix.
  * You can also make the wording more concise; summary lines can be in a terse "headlinese" style. To see examples, try `git log --oneline --author Greg` or `git log --oneline --author cbobbe`.

To see example commit messages in general, here's a slightly broader version of that last pair of commands:
```
$ git log --stat --author '(greg|cbobbe)@zulip' -P
```

Khader-1 (commented):
This is not tagged as `[nfc]` because it does actually add functionality or at least it tests  it's existence. Besides the new test case is mentioned as a part of the [issue](https://github.com/zulip/zulip-flutter/issues/451).

gnprice:
> This is not tagged as `[nfc]` because it does actually add functionality or at least it tests it's existence. Besides the new test case is mentioned as a part of the [issue](https://github.com/zulip/zulip-flutter/issues/451).

Sounds good, yeah.

(I assume you're referring to the second commit in the series:
62015a2b6 compose test [nfc]: Make trailing whitespace explicit in test cases
9672a0af4 compose test: Check ID still included if same-name users deactivated
a88222a7b autocomplete [nfc]: factor out _testName filtering helper
f9f476f65 autocomplete test [nfc]: Use test group to handle testUser cases
854a18c97 autocomplete: Exclude user from autocomplete popup if deactivated

because that's the one that was marked NFC in the previous revision, and isn't in this one.)

gnprice:
Thanks @Khader-1 for the revision! This all looks good except two small nits in commit messages. I'll fix those up and merge.

    autocomplete [nfc]: factor out _testName filtering helper

The summary line should be in sentence case after the prefix: so "Factor out …".

    autocomplete: Exclude user from autocomplete popup if deactivated
    
    Deactivated users can not be mentioned, so this commit ensures that
    deactivated users are excluded from the list that gets rendered in
    autocomplete popup.
    
    Fixes #451

The "fixes" line should have a colon: https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#fixes-format

PR #586, opened 2024-03-24
#465_connecting_banner

satyam372:
video link:-

https://github.com/zulip/zulip-flutter/assets/112092492/19f0710a-1eef-4ccf-a43a-60520153aece

To implement this feature i have used connectivity-plus package "connectivity_plus: ^5.0.2" . This package is used to see if the device has an active internet connection or not . 
I have written code for snackbar in "snackbar.dart" file and then included it in app.dart file and message_list.dart file.
The snackbart.dart file is created inside the widget folder
Through this feature we can give the status of connectivity to the user.

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -1,5 +1,5 @@
 import 'dart:async';
-
+import 'snackbar.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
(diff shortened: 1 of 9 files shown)


lib/widgets/app.dart
abelaba, line 1: I think some of your code has been auto-formatted and this project doesn't use auto-formatting, so I think you should check that. If your editor or IDE automatically reformats the existing code, you'll want to turn that off.

abelaba:
You need to also add tests for the new snack bar widget you have added. 

https://github.com/zulip/zulip-flutter?tab=readme-ov-file#writing-tests

abelaba:
Your commit messages should also follow this https://github.com/zulip/zulip-flutter?tab=readme-ov-file#writing-tests

gnprice:
Thanks @abelaba for the initial review! Those are all good comments.

@satyam372 In fact those are exactly the things that will be needed before we can review this PR in detail. (Plus one more thing: the tests will need to be passing.) For background, see: https://github.com/zulip/zulip-flutter#submitting-a-pull-request

satyam372:
Thankyou @gnprice and @abelaba for your feedback . I have started working on it and will make a good PR.

Issue #356, opened 2023-11-01
Handle video previews


gnprice:
We should handle video preview elements in message content. Here's an example message that uses one:
https://chat.zulip.org/#narrow/stream/9-issues/topic/.E2.9C.94.20inbox.20unread.20dot.20only.20present.20in.20view/near/1668976

To get the server's HTML for that message, see comments in `test/model/content_test.dart`. (Don't try to get the HTML from the web app via browser devtools; this is one kind of content where the web app is known to show HTML that's been adjusted from what the server gave it.)

Handling these means parsing them in `lib/model/content.dart`, and then displaying them appropriately in `lib/widgets/content.dart`.

---

In an ideal version of this feature,
* we'd start by showing a still thumbnail image, probably with a "this is a video" indication similar to what web has;
* then if the user taps on the video, we'd play it, probably after going to the lightbox;
* and when playing, we'd have appropriate controls to pause, to resume, to start over, etc.

If we can readily get all of that functionality off the shelf, then great. If not, then for an initial version it'd be good enough to:
* start by showing a still thumbnail image, with a "this is a video" indication as above;
* then if the user taps on the video, send them to a web browser with a link that will let them view it.

In that case we'd file a followup issue for the remaining functionality.

PR #587, opened 2024-03-24
content: Handle video previews

rajveermalviya:
Implement thumbnail image based video previews for Youtube & Vimeo
video links, and inline video player preview using the video_player
package for user uploaded videos.

For user uploaded video, current implementation will fetch the metadata
when the message containing the video comes into view. This metadata
is used by the player to determine if video is supported on the device.
If it isn't then there will be no preview, and tapping on the play
button will open the video externally. If the video is supported, then
the first frame of the video will be presented as a preview in the
message container while tapping on the play button will start buffering
and playing the video in the lightbox.

There are still some quirks with the current implementation:

- On iOS/macOS, there is a bug where whole video is downloaded before
  playing: https://github.com/flutter/flutter/issues/126760

- On iOS/macOS, unlike on Android the first frame is not shown after
  initialization: https://github.com/flutter/flutter/issues/139107

- Current implementation uses url_launcher for fallback in case video
  is not supported by video_player, we should switch to webview
  instead to correctly handle auth headers for private videos.

Fixes: #356

https://github.com/zulip/zulip-flutter/assets/36819268/f7a0b8b0-e8ce-48af-b934-0ab1a47e4e26


pubspec.yaml
```diff
@@ -65,6 +65,7 @@ dependencies:
   url_launcher: ^6.1.11
   url_launcher_android: ">=6.1.0"
   sqlite3: ^2.4.0
+  video_player: ^2.8.3
```
gnprice, line 68: The dependency should be added in its own commit. See `git log --stat -p --full-diff pubspec.*` for examples.

macos/Podfile.lock
```diff
@@ -163,7 +169,8 @@ SPEC CHECKSUMS:
   sqlite3: 73b7fc691fdc43277614250e04d183740cb15078
   sqlite3_flutter_libs: 06a05802529659a272beac4ee1350bfec294f386
   url_launcher_macos: d2691c7dd33ed713bf3544850a623080ec693d95
+  video_player_avfoundation: 02011213dab73ae3687df27ce441fbbcc82b5579
 
 PODFILE CHECKSUM: 353c8bcc5d5b0994e508d035b5431cfe18c1dea7
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.15.0
```
gnprice, line 176: This CocoaPods bump is fine, but should be its own commit because it's logically independent of the other things happening. See `git log --stat -p -G COCOAPODS` for past examples.

I suspect that on your machine if you just do `flutter run` for iOS and macOS targets, starting from main, it'll already make these two changes.

lib/model/content.dart
```diff
@@ -1024,6 +1083,10 @@ class _ZulipContentParser {
       return parseImageNode(element);
     }
 
+    if (localName == 'div' && classes.containsAll(['message_inline_image', 'message_inline_video'])) {
```
gnprice, line 1086: We'll want to use `className` rather than `classes`; we made that switch earlier this year. See `git log -p -S classes lib/model/content.dart` for examples and motivation.

This one will be a bit nontrivial (because there's more than one class), but see the latest couple of commits matched by that filter for examples of similar complexity.

gnprice (commented):
Interesting! Looking forward to this feature.

This will need tests, naturally. I'm guessing that's the main reason you've marked it as draft.

I'll wait to do a detailed review until you've said the PR is ready for one. But from a quick skim just now, here's a few high-level points.

rajveermalviya:
Thanks for the initial comments @gnprice! It should be ready for the first round of review.

For the two iOS bugs, we could just move the initialization into the lightbox on iOS and don't care about first frame preview, but then we loose the ability to check if video is supported or not when user taps on play, if that happens when already in the lightbox then we will have to show a dialog that video's unsupported with a button to open it in webview, pretty weird UX.

Or just always open video in webview on iOS.

Or ofcourse get these bugs fixed, & live with them temporarily.

test/model/content_test.dart
```diff
@@ -521,6 +521,118 @@ class ContentExample {
...
+  ]);
+
+  static const videoEmbedVimeoPreviewEnabled = ContentExample(
+    'video preview for vimeo embed with realm link previews enabled',
+    null,
+    '<p>'
+      '<a href="https://vimeo.com/1084537">Vimeo - Big Buck Bunny</a>'
+    '</p>\n'
+    '<div class="embed-video message_inline_image">'
+      '<a data-id="&lt;iframe src=&quot;https://player.vimeo.com/video/1084537?app_id=122963&quot; width=&quot;640&quot; height=&quot;360&quot; frameborder=&quot;0&quot; allow=&quot;autoplay; fullscreen; picture-in-picture; clipboard-write&quot; title=&quot;Big Buck Bunny&quot;&gt;&lt;/iframe&gt;" href="https://vimeo.com/1084537" title="Big Buck Bunny">'
```
gnprice, line 578: The value of the `data-id` attribute is a blob of HTML? For an iframe element? Wacky. That's not much of an "ID".

If the Zulip server produces this, or has produced it in the past, then it's good to have in a test case. But it smells rather like a bug. It'd be good to report it in a chat thread in `#issues` and see what people say.

This has enough resemblance to a potential HTML injection that I waffled on whether to treat it as a potential security issue (which should be sent to security@zulip.com rather than discussed in public). So I poked around in the server implementation, and the logic for this is pretty explicit in `zerver/lib/markdown/__init__.py`:
```python
        elif extracted_data.type == "video":
            self.add_a(
                # …
                class_attr="embed-video message_inline_image",
                data_id=extracted_data.html,
```

That makes it look intentional enough that I'm OK with keeping the discussion public. But I'd still be interested in what the story is with calling this "id" when it's actually some HTML for an iframe.

test/model/content_test.dart
```diff
@@ -521,6 +521,118 @@ class ContentExample {
...
+    'video preview for vimeo embed with realm link previews disabled',
+    null,
+    '<p>'
+      '<a href="https://vimeo.com/1084537">https://vimeo.com/1084537</a></p>', [
+    ParagraphNode(links: null, nodes: [
+      LinkNode(url: 'https://vimeo.com/1084537', nodes: [TextNode('https://vimeo.com/1084537')]),
+    ]),
+  ]);
+
+  static const videoEmbedVimeoPreviewEnabled = ContentExample(
```
gnprice, line 571:
```suggestion
  static const videoEmbedVimeo = ContentExample(
```

Given that the preview-disabled case isn't really a "video embed" at all, we can treat this as the main Vimeo example and give it a shorter name.

gnprice (commented):
Thanks @rajveermalviya! I've read through the parsing commit in detail, and it generally looks good; comments below. The two deps commits also look good. I haven't yet read through the widgets/playback commit.

Those two upstream issues are interesting. Are you observing their symptoms?

In issue
* https://github.com/flutter/flutter/issues/126760

it kind of sounds like https://github.com/flutter/packages/pull/3826 may have fixed the issue last July. The original reporter (who had a Vimeo example) didn't reply after that. There's one person who showed up in January reporting similar symptoms, but it's not clear that their issue wasn't some specific interaction with the CDN they were using.

In issue
* https://github.com/flutter/flutter/issues/139107

there are several reports that things work fine in a Flutter iOS app and that the only trouble is in Flutter web on iOS browsers, and the issue is tagged that way. The thread is a bit scattered, as there are also reports of trouble on Flutter iOS.

test/model/content_test.dart
rajveermalviya: [Reported](https://chat.zulip.org/#narrow/stream/9-issues/topic/Vimeo.20link.20previews.20HTML.20.22data-id.22.20isn't.20a.20.20Vimeo.20video.20ID/near/1767563)

rajveermalviya:
Thanks @gnprice for the review! Rebased with comments addressed.

> Are you observing their symptoms?

Yes I am able to reproduce them atleast on macOS and iOS Simulator, but haven't tested if this could be a CDN issue. I will test it using bare AVPlayer to see if it really is something to fix in Zulip's CDN config.

rajveermalviya:
(rebased to main)

gnprice (commented):
Thanks @rajveermalviya for the revision! Here's a fresh round of review on the parsing commit.

I'll aim to get to the widgets/playback commit soon, too; but there's an unusually large rate of PRs coming in right now, so I'll cycle through some of the others first.

rajveermalviya:
@gnprice thanks for the review! Pushed a new revision.

gnprice (commented):
Thanks! Just one comment on the parsing commit in this revision.

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+          // TODO clean up this padding by imitating web less precisely;
+          //   in particular, avoid adding loose whitespace at end of message.
+          padding: const EdgeInsets.only(right: 5, bottom: 5),
+          child: LightboxHero(
+            message: message,
+            src: widget.src,
+            child: Container(
+              height: 100,
+              width: 150,
+              color: Colors.black,
```
gnprice, line 548: There's a lot of details here that seem like they get duplicated into three places, namely MessageImage and the two video types:
* a GestureDetector
* contains an UnconstrainedBox aligned centerLeft
* contains certain 5px padding, with a TODO comment
* then there's a certain 150x100 shape.

It'd be great to abstract some of those so that they can be shared. I think it's not a coincidence these are similar — they're designed to look similar ­— so it's good to arrange the code in a way that will help them naturally remain similar as we make changes.

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+              child: Stack(
+                alignment: Alignment.center,
+                children: [
+                  if (_initialized && !_controller!.value.hasError) ...[
+                    AspectRatio(
+                      aspectRatio: _controller!.value.aspectRatio,
+                      child: VideoPlayer(_controller!)),
+                    Container(color: const Color.fromRGBO(0, 0, 0, 0.30)),
```
gnprice, line 556: Huh, if I'm reading this right, it puts a 30%-opacity black layer in front of the video (by putting it later in a `Stack`'s children).

Is that intended? It doesn't seem like what we'd want.

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+                  ],
+                  if (_initialized)
+                    const Icon(
+                      Icons.play_arrow_rounded,
+                      color: Colors.white,
+                      size: 25)
```
gnprice, line 562: How did you choose 25px for the size?

It appears to be 32px on web and in zulip-mobile. (For the latter, see `src/webview/static/base.css`.)

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+
+    // For YouTube and Vimeo links, display a widget with a thumbnail.
+    // When the thumbnail is tapped, open the video link in an external browser or
+    // a supported external app.
+    if (node.previewImageUrl != null) {
+      return MessageEmbedVideoPreview(
```
gnprice, line 403: The information in this comment would be good to turn into a short dartdoc for the `MessageEmbedVideoPreview` class.

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+          //   in particular, avoid adding loose whitespace at end of message.
+          padding: const EdgeInsets.only(right: 5, bottom: 5),
+          child: Container(
+            height: 100,
+            width: 150,
+            color: Colors.black,
+            child: Stack(
+              alignment: Alignment.center,
+              children: [
+                if (previewImageUrl != null) ...[
```
gnprice, line 446: Similar comment as on the `Container` fallback for the `node.srcUrl` case above.

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+      ? MessageInlineVideoPreview(src: resolvedSrc)
+      : Container();
+  }
+}
+
+class MessageEmbedVideoPreview extends StatelessWidget {
+  const MessageEmbedVideoPreview({
+    super.key,
+    required this.previewImage,
+    required this.src,
```
gnprice, line 418: Looking at these two widget classes for the two kinds of videos, and at `MessageVideo` which dispatches to them, I think it'd probably be best to have two different ContentNode subclasses for them.

In particular there's a very important difference in semantics between these two, if I'm correctly understanding how these work:
* For "inline" videos, `node.srcUrl` is a URL on the Zulip server, and so it's fine to load it when the user just scrolls passively past that message (and doing so is helpful for the reasons you mention in the PR description).
* For "embed" videos, `node.srcUrl` is a URL on some *other* external service. For privacy reasons it's therefore important that we not go and talk to that service unless the user actually chooses to interact with the video. OTOH `node.previewImageUrl` is a URL back on the user's Zulip server (provided by Camo), and safe to eagerly fetch from.

I believe this PR already handles that distinction correctly. But in this version it's pretty subtle in the code, and so it'd be easy for some future change to break that privacy property without realizing there was something important there. We should therefore arrange the code so that the distinction is clear. As part of that, we should avoid having these two kinds of URLs with very different privacy properties both live as values of the same field on the same class.

Conversely, `MessageVideo` is doing very little, really just dispatching to these two related classes that have pretty different logic from each other. So even apart from the privacy issue, it feels like having two different `ContentNode` subclasses would better match up with how the logic actually works.

If there's anyplace where it would be helpful to do so, those two `ContentNode` subclasses could always have a common base class with a name like `VideoNode`.

lib/widgets/content.dart
```diff
@@ -382,6 +387,188 @@ class MessageImage extends StatelessWidget {
...
+        : () { // TODO(log)
+            if (_controller!.value.hasError) {
+              // TODO use webview instead, to support auth headers
+              _launchUrl(context, widget.src.toString());
+            } else {
+              Navigator.of(context).push(getLightboxRoute(
+                context: context,
+                message: message,
+                src: widget.src,
+                videoController: _controller,
```
gnprice, line 532: It looks like this means that playback in the lightbox picks up from right where it was. Neat!

rajveermalviya:
@gnprice, thank you for your review. I've just pushed a new revision. Apologies for the delay

gnprice:
Thanks for the revision!

I'm just about to head out [on vacation](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Greg.20vacation/near/1768574), so it'll be a couple of weeks before I'm able to give this PR another review. But I'll look forward to picking it up when I'm back.

@chrisbobbe if you find some time to help move this forward with a review while I'm away, the key thing I'd highlight from the above review threads is [this point](https://github.com/zulip/zulip-flutter/pull/587#discussion_r1546968146) about the important privacy-related difference between inline vs. embedded videos.

rajveermalviya:
(rebase to main)

rajveermalviya:
> There are still some quirks with the current implementation:
>
> - On iOS/macOS, there is a bug where whole video is downloaded before
>   playing: https://github.com/flutter/flutter/issues/126760
>
> - On iOS/macOS, unlike on Android the first frame is not shown after
>   initialization: https://github.com/flutter/flutter/issues/139107
>
> - Current implementation uses url_launcher for fallback in case video
>   is not supported by video_player, we should switch to webview
>   instead to correctly handle auth headers for private videos.

Took a look at these issues again yesterday -

#### On iOS/macOS video_player ends up downloading the whole video after init

With a [custom SwiftUI impl of AVPlayer](https://github.com/rajveermalviya/VideoPlayerTest) I am getting same behaviour as video_player plugin for the small video files hosted by Zulip and [my personal S3 bucket](https://github.com/rajveermalviya/VideoPlayerTest/blob/85effbbee5017f6f0b780c298b6f243c84d2867c/VideoPlayerTest/ContentView.swift#L42) (hosted on Backblaze B2). But with a larger video file (~280MB) the behaviour is different on both impls (custom & video_player), they download about ~1min of video after init and then stops, after playing both impls will start buffering on-demand. So, its safe to say that is just that the AVPlayer's resource loading is a bit more aggressive than Android ExoPlayer (what video_player uses on Android). But I only tested this on macOS and iOS Simulator on Wifi, so it may be that AVPlayer would not be so aggressive on Cellular network on an actual iPhone.

Note: The problem is **not that** we can't start playing video unless it's downloaded completely, it does start playing immediately after init. The problem is when user hasn't even clicked on the video and it is downloaded as whole, unlike on Android where just the metadata & first few frames is loaded after init and playing the video starts downloading more of it.

#### On iOS/macOS video_player doesn't show first frames as the preview after init

Turns out I lied, it does show up, it was just that the specific video I tested had empty black frames for first few seconds. And another video that had first few frames was just taking up little longer to init than on Android.

#### Fallback behaviour when video is unsupported

The current implementation of launching the video url to external app or browser doesn't help much, because of two reasons:

- Video url will most probably require Auth headers and we can't expose them to external app obvious security reasons, and API doesn't allow it anyway.
- If we open external browsers or webview then it doesn't help there too, because on iOS if AVPlayer doesn't support it then its safe to say Safari won't support it too, similarly on Android if ExoPlayer doesn't support a video (more unlikely) then it's safe to say other browsers won't support it.

So, one way to do this correctly would be to provide an option to download the video file, and then tell user to open it a different video player app, or we do the launching ourselves. This would be the similar behaviour as on Zulip Mobile.

Another way is to rely on other video player libraries like [gstreamer](https://gitlab.freedesktop.org/gstreamer/gstreamer) or [libmpv](https://github.com/mpv-player/mpv), where they can use hardware accelerated codecs provided by the platform ([VideoToolbox](https://developer.apple.com/documentation/videotoolbox) on iOS/macOS & [MediaCodec](https://developer.android.com/reference/android/media/MediaCodec) on Android) and fallback to the software codecs (list of codecs is configurable). This would allow supporting the WebM vp8/vp9 videos (common with Linux screen recordings) to be playable on iOS, which AVPlayer doesn't support currently.

There is one such (popular) plugin that uses `libmpv`: [media_kit](https://github.com/media-kit/media-kit).

rajveermalviya:
With these issues in mind I have updated the implementation to not initialize the video player when the message containing the video comes into view, instead it will all happen now when the video lightbox opens.

This make the UX a bit boring than before, no video previews, no auto resume from same position in video when video is reopened. But avoids the unnecesary aggressive downloading behaviour.


https://github.com/zulip/zulip-flutter/assets/36819268/c4abcb7e-e58d-476c-a114-0395f5ea2977

lib/widgets/lightbox.dart
```diff
@@ -208,11 +213,244 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+      }
+    }
+  }
+
+  @override
+  void dispose() {
+    _controller?.removeListener(_handleVideoControllerUpdates);
+    _controller?.dispose();
+    widget.routeEntranceAnimation.removeStatusListener(_handleRouteEntranceAnimationStatusChange);
+    super.dispose();
```
gnprice, line 287: Cool, good to see the order of these updated to match (in reverse order) changes to the order of the corresponding initializations.

lib/widgets/lightbox.dart
```diff
@@ -208,11 +213,244 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+                        width: 32,
+                        height: 32,
+                        child: CircularProgressIndicator(color: Colors.white))
+                  ]),
+              ))))));
+  }
+}
+
+extension DurationFormatting on Duration {
+  String formatHHMMSS() {
```
gnprice, line 434: This name is a bit misleading, because the method's behavior is that the presence of HH is conditional.

Rather than try to give this a name that accurately conveys what it does in a way that applies in a general `Duration` context, let's just embrace that it's a private helper for video playback, and make it a private static method on `_VideoLightboxPageState`. Then it can have a generic name like `_formatDuration`.

lib/widgets/lightbox.dart
```diff
@@ -208,11 +213,244 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+    final hoursString = inHours.toString().padLeft(2, '0');
+    final minutesString = inMinutes.remainder(60).toString().padLeft(2, '0');
+    final secondsString = inSeconds.remainder(60).toString().padLeft(2, '0');
+
+    return '${hoursString == '00' ? '' : '$hoursString:'}$minutesString:$secondsString';
```
gnprice, line 439: nit: let's try leaving off "String" in these names — there's no other kind of "hours" etc. value around to confuse them with, and I think it'll make the combining logic at the end a bit easier to read.

lib/widgets/lightbox.dart
```diff
@@ -208,11 +213,244 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+          // Clobber the MediaQueryData prepared by Scaffold with one that's not
+          // affected by the app bars. On this screen, the app bars are
+          // translucent, dismissible overlays above the pan-zoom layer in the
+          // Z direction, so the pan-zoom layer doesn't need avoid them in the Y
+          // direction.
+          data: MediaQuery.of(context),
+
+          child: GestureDetector(
+            behavior: HitTestBehavior.translucent,
+            onTap: _handleTap,
```
gnprice, line 413: There's a lot of structure here — basically everything from `return` through this `onTap` line, and some other code that supports that — that's duplicated between the two kinds of lightbox. I think that's not a coincidence, in that it's basically the code for making "a lightbox" independent of exactly what's shown in the lightbox.

So let's factor things out to deduplicate that code. Two reasonable strategies are:
* Make a common base class, or I guess pair of base classes, like `abstract class _LightboxPage extends StatefulWidget` and `abstract class _LightboxPageState extends State<_LightboxPage>`. Then the subclasses don't override `build`, but rather some subroutines that `build` calls, like `buildBottomAppBar` and `buildChild`. (Plus `initState` etc.)
  This is the approach taken at `MessageActionSheetMenuItemButton` and some other places.
* Make a common helper widget, like `class _LightboxPageLayout extends StatefulWidget`. Then the specific lightbox widgets pass constructor arguments like `buildBottomAppBar` and `child`. The state lifetime methods like `initState` etc. are independent on each widget and don't have to interact via `super` calls.

  This is the approach taken at `_ComposeBoxLayout` and I think some other places.

lib/widgets/lightbox.dart
```diff
@@ -208,11 +213,244 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+
+    try {
+      await _controller!.initialize();
+      await _controller!.play();
+    } catch (error) { // TODO(log)
+      assert(debugLog("VideoPlayerController.initialize failed: $error"));
+      if (mounted) {
+        await showDialog(
+          context: context,
+          barrierDismissible: false,
```
gnprice, line 270: I'd also be interested in reunifying this with the methods in `dialog.dart`. Those are helpful for keeping UI consistency across our dialog interactions.

What's the reasoning for setting `barrierDismissible: false` here? It looks like that, and the logic to pop the lightbox in addition to the dialog, are the two differences from `showErrorDialog`.

We could add an option there to control these differences, given a clear story for what they're about. (Perhaps just a boolean to choose between the existing behavior and this behavior? Or perhaps the two changes shouldn't be coupled. It depends on what the reasoning is for this `barrierDismissible` change.)

gnprice (commented):
OK, back last week from vacation, and catching up on PRs. Thanks for the revision! Here's a fresh review, including everything covered before (in particular parsing) plus now the lightbox code too.

I believe at this point I've reviewed everything, except the test code from the last commit.

> With these issues in mind I have updated the implementation to not initialize the video player when the message containing the video comes into view, instead it will all happen now when the video lightbox opens.
>
> This make the UX a bit boring than before, […]

OK, sounds good. That also means less code and complexity, which helps us get closer to merging this. :slightly_smiling_face:

I do think being able to preview the video right in the message list, and having the playback continue smoothly from where it left off when you move to the lightbox, are pretty neat. So let's file a followup issue for adding those, and link to your comment https://github.com/zulip/zulip-flutter/pull/587#issuecomment-2053570181 — that will help us pick that work up later. (→ #656)

lib/widgets/lightbox.dart
```diff
@@ -208,11 +213,244 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+  VideoPlayerController? _controller;
+
+  @override
+  void initState() {
+    super.initState();
+    widget.routeEntranceAnimation.addStatusListener(_handleRouteEntranceAnimationStatusChange);
+    // We delay initialization by a single frame to make sure the
+    // BuildContext is a valid context, as its needed to retrieve
+    // the PerAccountStore & ZulipLocalizations during initialization.
+    SchedulerBinding.instance.addPostFrameCallback((_) => _initialize());
```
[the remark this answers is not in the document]
rajveermalviya: This was done to ensure that the dependencies (PerAccountStore & ZulipLocalizations) would be registered on BuildContext after the `build` method is invoked, thereby delaying initialization until the next frame.

Anyways changed it now to instead do the initialization in `didChangeDependencies`, which get called after `initState` but with dependencies registered to the State object.

lib/widgets/lightbox.dart
rajveermalviya: Added a comment on the reasoning for `barrierDismissible: false`.

```dart
// To avoid showing the disabled video lightbox for the unnsupported
// video, we make sure user doesn't reach there by dismissing the dialog
// by clicking around it, user must press the 'OK' button, which will
// take user back to content message list.
barrierDismissible: false,
```

rajveermalviya:
Thanks for the review @gnprice, pushed a new revision!

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+      message: widget.message,
+      buildBottomAppBar: (context, color, elevation) =>
+        _controller == null
+          ? null
+          : BottomAppBar(
+              height: 150,
+              color: color,
+              elevation: elevation,
+              child: Column(mainAxisAlignment: MainAxisAlignment.end, children: [
+              Row(children: [
```
gnprice, line 342: child should be indented:
```suggestion
              child: Column(mainAxisAlignment: MainAxisAlignment.end, children: [
                Row(children: [
```

(as should a list element in general)

I'd usually say "nit" for formatting issues; but this is a big widget expression with a lot of nesting and multiple lists, and so between this and a couple of other formatting issues it actually gets noticeably hard to read the structure of it.

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+                  value: _controller!.value.position.inSeconds.toDouble(),
+                  max: _controller!.value.duration.inSeconds.toDouble(),
+                  activeColor: Colors.white,
+                  onChanged: (value) {
+                    _controller!.seekTo(Duration(seconds: value.toInt()));
+                  })),
+                Text(_formatDuration(_controller!.value.duration),
+                  style: const TextStyle(color: Colors.white)),
+                ]),
+              IconButton(
```
gnprice, line 355: A line starting with close-bracket (or close-paren, etc.) should be indented the same as the line with the corresponding open-bracket (or open-parent, etc.).

So this `]),` should match the indentation of `Row(children: [`, rather than be indented an extra level. Combined with the comment above (the `Row` should be indented one more level), that comes to:

```suggestion
                  Text(_formatDuration(_controller!.value.duration),
                    style: const TextStyle(color: Colors.white)),
                ]),
                IconButton(
```

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+              child: Column(mainAxisAlignment: MainAxisAlignment.end, children: [
+              Row(children: [
+                Text(_formatDuration(_controller!.value.position),
+                  style: const TextStyle(color: Colors.white)),
+                Expanded(child: Slider(
+                  value: _controller!.value.position.inSeconds.toDouble(),
+                  max: _controller!.value.duration.inSeconds.toDouble(),
+                  activeColor: Colors.white,
+                  onChanged: (value) {
+                    _controller!.seekTo(Duration(seconds: value.toInt()));
```
gnprice, line 350: Interesting, this puts everything into whole seconds. What's the reasoning for that?

This behavior seems probably fine, except for one case where it seems like it could be annoying: if you try to seek to the end using the slider, it'll be a little bit before the end. If the video happens to be 2.9 seconds long, or 1234.9 seconds, then the slider will only get you to 900ms away from the end, no matter how hard you drag it to the end.

What happens if we just use the times with as much precision as we get them with?

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+                child: VideoPlayer(_controller!)),
+            if (_controller == null || !_controller!.value.isInitialized || _controller!.value.isBuffering)
+              const SizedBox(
+                width: 32,
+                height: 32,
+                child: CircularProgressIndicator(color: Colors.white)),
+            ]))));
+  }
+
+  String _formatDuration(Duration value) {
```
gnprice, line 384: nit: other methods come before `build`:

Separately: this is naturally `static` — it doesn't care about any data from the class instance — so let's declare it that way.

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return _LightboxPageLayout(
+      routeEntranceAnimation: widget.routeEntranceAnimation,
+      message: widget.message,
+      buildBottomAppBar: (context, color, elevation) =>
+        _controller == null
+          ? null
+          : BottomAppBar(
```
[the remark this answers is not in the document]
gnprice: Oh in fact better yet: make this a method `_buildBottomAppBar`, defined just above `build`, and then make this a tearoff:
```suggestion
      buildBottomAppBar: _buildBottomAppBar,
```

The block body reduces the indentation by a couple of levels relative to the expression body and `?:`, which is welcome. The method will reduce it by a couple more, and also help the code organization a bit by getting these long details out of the main widget expression at the end of `build`.

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+}
+
+class _ImageLightboxPageState extends State<_ImageLightboxPage> {
+  @override
+  Widget build(BuildContext context) {
+    return _LightboxPageLayout(
+      routeEntranceAnimation: widget.routeEntranceAnimation,
+      message: widget.message,
+      buildBottomAppBar: (context, color, elevation) => BottomAppBar(
+        color: color,
```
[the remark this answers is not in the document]
gnprice: This might as well be a method tearoff just like in `VideoLightboxPage`, for maximum parallelism, though it's short enough that it's also fine as a function expression with block body.

lib/model/content.dart
```diff
@@ -374,6 +374,58 @@ class ImageNode extends BlockContentNode {
   }
 }
 
+class InlineVideoNode extends BlockContentNode {
+  const InlineVideoNode({
+    super.debugHtmlNode,
+    required this.srcUrl,
+  });
+
+  final String srcUrl;
```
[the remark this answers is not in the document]
gnprice: Thanks! I started from the text in your revised version, and made some edits. I'll push those in an added commit on top; please squash that commit's changes into the respective commits in the branch.

lib/widgets/content.dart
```diff
@@ -373,11 +377,93 @@ class MessageImage extends StatelessWidget {
...
+    return MessageMediaContainer(
+      onTap: () => _launchUrl(context, node.hrefUrl),
+      child: Stack(
+        alignment: Alignment.center,
+        children: [
+          if (previewImageSrcUrl != null)
+            RealmContentNetworkImage(
+              previewImageSrcUrl,
+              filterQuality: FilterQuality.medium),
+          const Icon(
```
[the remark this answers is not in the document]
gnprice: The comment added on the other case is fine, but this case is the one that feels more surprising to me, so it's where I think a comment would be most helpful. Here's a version:

```suggestion
          // Show the "play" icon even when previewImageSrcUrl didn't resolve;
          // the action uses hrefUrl, which might still work.
          const Icon(
```

lib/widgets/lightbox.dart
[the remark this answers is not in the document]
gnprice: Cool. Yeah, `didChangeDependencies` is a good solution for that.

(Though see my other comment in this new batch about `PerAccountStoreAwareStateMixin` and its `onNewStore`, which is a specialized version of `didChangeDependencies` that's especially helpful when depending on `PerAccountStore` in particular.)

lib/widgets/lightbox.dart
```diff
@@ -197,22 +199,207 @@ class _LightboxPageState extends State<_LightboxPage> {
...
+  VideoPlayerController? _controller;
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    assert(debugLog('_VideoLightboxPageState($hashCode).didChangeDependencies'));
+
+    // We are pretty sure that the dependencies (PerAccountStore & ZulipLocalizations)
+    // won't change while user is on the lightbox page but we still
+    // handle the reinitialization for correctness.
```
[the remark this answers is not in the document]
rajveermalviya: > `PerAccountStoreAwareStateMixin`

Ah, that is indeed what I was looking for, not sure how I missed it when looking for existing examples in the code base.

> `ZulipLocalizations` can also change.

That is indeed a fun possibility, anyways added comments regardless.

lib/widgets/lightbox.dart
rajveermalviya: Initial reasoning was that interactivity of seeking would be much laggy if we did `seekTo` on every 'onChange' for higher precisions.

Now the slider uses millisecond precision, which seems to be what we are getting from the controller (by logging the controller.position). Also the call to `controller.seekTo` now happens in `onChangeEnd` which now seems like what should have been done initially anyway.

rajveermalviya:
Thanks for the review! Pushed a new revision.

rajveermalviya:
(rebased to main)

test/test_video_player.dart
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @override
+  Future<void> setLooping(int textureId, bool looping) async {
+    assert(textureId == _textureId);
+  }
+
+  @override
+  Future<void> play(int textureId) async {
+    assert(textureId == _textureId);
```
gnprice, line 62: These asserts are very helpful! They help significantly reduce the reader's worry that perhaps these methods end up getting called in a way that breaks their assumptions.

Let's add a few more assertions in that spirit:
* `create` check not already initialized
* `dispose` check it is initialized
* couple of other methods add this `textureId` check, where they don't have it already

and I'll add others inline.

test/test_video_player.dart
```diff
@@ -0,0 +1,109 @@
...
+  @override
+  Stream<VideoEvent> videoEventsFor(int textureId) {
+    assert(textureId == _textureId);
+    return _streamController.stream;
+  }
+
+  @override
+  Future<void> setLooping(int textureId, bool looping) async {
+    assert(textureId == _textureId);
+  }
```
gnprice, line 58: Since we're not storing the `looping` value, let's check it has the value we expect, namely false.

test/test_video_player.dart
```diff
@@ -0,0 +1,109 @@
...
+  // Future<void> seekTo(int textureId, Duration position) {
+  //   throw UnimplementedError('seekTo() has not been implemented.');
+  // }
+
+  @override
+  Future<void> setPlaybackSpeed(int textureId, double speed) async {
+  }
+
+  // Future<Duration> getPosition(int textureId) {
+  //   throw UnimplementedError('getPosition() has not been implemented.');
```
gnprice, line 94: We can leave out this commented-out code. If/when we write a test where the method does get called, it'll be easy to add a similar stub then; if not, we don't need it.

test/test_video_player.dart
```diff
@@ -0,0 +1,109 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+import 'package:video_player_platform_interface/video_player_platform_interface.dart';
+
+void setupMockVideoPlayerPlatform() {
+  VideoPlayerPlatform.instance = MockVideoPlayerPlatform();
+}
+
+class MockVideoPlayerPlatform extends Fake
```
gnprice, line 12:
```suggestion
class FakeVideoPlayerPlatform extends Fake
```

That is, let's call this a "fake" instead of a "mock".

Broadly, a mock's method implementations do nothing except record they were called, in a log the test will then inspect. A fake's method implementations do some facsimile of what a real implementation would do, just with all the expensive parts replaced by shortcuts.

This is already more of a fake than a mock, because e.g. the test looks at `isPlaying` rather than counting calls to `play` and `pause`. And that's a good thing — generally fakes are preferable to mocks. They make the tests less coupled to accidental implementation details of the code under test, and better aligned to the code's actual interface and spec.

test/widgets/content_test.dart
```diff
@@ -328,6 +328,95 @@ void main() {
...
+    }
+
+    testWidgets('video preview for youtube embed', (tester) async {
+      const example = ContentExample.videoEmbedYoutube;
+      await prepareContent(tester, example.html);
+
+      final expectedVideo = example.expectedNodes[1] as EmbedVideoNode;
+      final expectedResolvedPreviewUrl = eg
+        .store()
+        .tryResolveUrl(expectedVideo.previewImageSrcUrl)!;
```
gnprice, line 354: nit:
```suggestion
      final expectedResolvedPreviewUrl = eg.store()
        .tryResolveUrl(expectedVideo.previewImageSrcUrl)!;
```

("eg" alone means very little, so keep it together with the name of what we're pulling off it)

test/widgets/content_test.dart
```diff
@@ -328,6 +328,95 @@ void main() {
...
+      await tester.tap(find.byIcon(Icons.play_arrow_rounded));
+      check(testBinding.takeLaunchUrlCalls())
+        .single.equals((url: Uri.parse(expectedLaunchUrl), mode: LaunchMode.platformDefault));
+    });
+
+    testWidgets('video preview for vimeo embed', (tester) async {
+      const example = ContentExample.videoEmbedVimeo;
+      await prepareContent(tester, example.html);
+
+      final expectedTitle = (((example.expectedNodes[0] as ParagraphNode).nodes[0] as LinkNode).nodes[0] as TextNode).text;
```
gnprice, line 370: nit: line very long
```suggestion
      final expectedTitle = (((example.expectedNodes[0] as ParagraphNode)
        .nodes.single as LinkNode).nodes.single as TextNode).text;
```

Also `.single` is good when applicable, as it makes more of the expectations clear.

test/widgets/content_test.dart
```diff
@@ -328,6 +328,95 @@ void main() {
...
+      await tester.ensureVisible(find.text(expectedTitle));
```
gnprice, line 371: Huh interesting — why is this needed in this test but not in the previous one?

gnprice: More generally, if these test cases work exactly the same way, it'd be nice to express that — basically factor out most of their guts, after the `await prepareContent` line, into a local helper like `checkEmbedVideo`. Which maybe takes the title string and the `EmbedVideoNode` as arguments.

test/widgets/content_test.dart
```diff
@@ -328,6 +328,95 @@ void main() {
...
+        .equals(expectedResolvedUrl);
+
+      final expectedLaunchUrl = expectedVideo.hrefUrl;
+      await tester.tap(find.byIcon(Icons.play_arrow_rounded));
+      check(testBinding.takeLaunchUrlCalls())
+        .single.equals((url: Uri.parse(expectedLaunchUrl), mode: LaunchMode.platformDefault));
+    });
+  });
+
+  group("MessageInlineVideo", () {
```
gnprice, line 387: nit: put these tests in same order as the code they're testing

gnprice (commented):
Thanks for the revision!

With this review I've gotten through the whole thing now, including those widget tests toward the end. Various comments below; outside those newly-reviewed tests, we're down to just a couple of them. So this is quite close.

As mentioned below, I'll also push an added commit on top, meant to be squashed in.

rajveermalviya:
Thanks for the review @gnprice, new revision pushed!

gnprice (commented):
Thanks @rajveermalviya for all your work on this feature! This all looks great — merging :tada: 

There's one comment I remember writing yesterday but that somehow got lost on GitHub; not sure how. The comment was to ask for a few more tests, specifically some tests of the time slider. I'll write up a new version of that comment, but there's no need for it to block merging — please write those tests as a follow-up PR instead.

test/widgets/lightbox_test.dart
```diff
@@ -0,0 +1,156 @@
...
+    testWidgets('shows a VideoPlayer, and video is playing', (tester) async {
+      await setupPage(tester, videoSrc: Uri.parse('https://a/b.mp4'));
+
+      check(FakeVideoPlayerPlatform.initialized).isTrue();
+      check(FakeVideoPlayerPlatform.isPlaying).isTrue();
+
+      await tester.ensureVisible(find.byType(VideoPlayer));
+    });
+
+    testWidgets('toggles between play and pause', (tester) async {
```
gnprice, line 142: Let's have a couple of tests of the slider mechanism, too. In particular:

* Slider starts at zero; let half the time pass, slider is at halfway point; let more than the full time pass, slider is at end. (E.g. it didn't jump back to zero.)

  To manipulate time in a test, use our `awaitFakeAsync` helper; see existing examples. Then call `FakeAsync.elapse`.

* Start dragging slider; drag it back and forth a few times; finish dragging. At each step, check its displayed position is the place you just dragged it to. Then at the end, check `seekTo` was only called once.
  I.e., a regression test for the lag-causing issue you described avoiding at https://github.com/zulip/zulip-flutter/pull/587#discussion_r1595126667 .

* Drag slider from one point to another. Check the slider appears at the place you dragged it to now, and at every frame until it starts advancing again because the video is playing. I.e., a regression test for the bug that the `_isSliderDragging` and `_sliderValue` logic is avoiding as explained in your comment quoted at https://github.com/zulip/zulip-flutter/pull/587#discussion_r1596190776 .

As described above at https://github.com/zulip/zulip-flutter/pull/587#pullrequestreview-2050737086 , please do those as a follow-up PR. Thanks!

gnprice:
> So let's file a followup issue for adding those, and link to your comment [#587 (comment)](https://github.com/zulip/zulip-flutter/pull/587#issuecomment-2053570181) — that will help us pick that work up later.

Filed this as #656.

PR #588, opened 2024-03-25
autocomplete: In @-mention results, match insensitively to diacritics

### LINKED ISSUE #237: autocomplete: In user-mention results, match insensitively to diacritics
If there's a user with `Étienne` in their name, we should still include it in the results for the query `etienne`.

The matching is already case-insensitive, but it should be insensitive to diacritics too.

Khader-1:
Fixes #237
A new dependency was introduced: [diacritic](https://pub.dev/packages/diacritic): ^0.1.5

### DIFF
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -327,6 +327,11 @@ void main() {
     doCheck('full full', eg.user(fullName: 'Full Full Name'), true);
     doCheck('full full', eg.user(fullName: 'Full Name Full'), true);
 
+    // assert that name checking is diacritic insensitive
+    doCheck('fűll ñámé', eg.user(fullName: 'Full Name'), true);
+    doCheck('fűll', eg.user(fullName: 'Full Name'), true);
+    doCheck('ñámé', eg.user(fullName: 'Full Name'), true);
+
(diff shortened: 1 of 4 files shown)


test/model/autocomplete_test.dart
```diff
@@ -327,6 +327,11 @@ void main() {
     doCheck('full full', eg.user(fullName: 'Full Full Name'), true);
     doCheck('full full', eg.user(fullName: 'Full Name Full'), true);
 
+    // assert that name checking is diacritic insensitive
```
Khader-1, line 330: This could be enhanced with test groups but that will be against the pattern in this file, I imagine some little tweaks would be required.
I would prefer updating it but I try to follow the patterns I find in the codebase.
So.. what do you suggest?

chrisbobbe:
To update this thread: it looks like we've identified a dependency to pull in 🙂 https://chat.zulip.org/#narrow/stream/48-mobile/topic/deps.3A.20Add.20new.20package.20to.20handle.20diacritics/near/1770345 so I/we will be happy to see a revision that does that, and review this again when it's un-marked as a draft.

Khader-1:
I believe this one was fixed by #1806 , closing..

PR #589, opened 2024-03-25
content: Handle strikethrough, the <del> element

### LINKED ISSUE #357: Handle strikethrough, the <del> element
We should handle the `<del>` element. This means "strikethrough" formatting; it's what you get with Markdown syntax like `~~this~~`, enclosing a span in `~~`.

Handling these means parsing them in `lib/model/content.dart`, and then styling them appropriately in `lib/widgets/content.dart`. For appropriate styling, see the Zulip web app's `web/styles/rendered_markdown.css`.

Khader-1:
This handles the `<del>` element. This means "strikethrough" formatting; it's what you get with Markdown syntax like `~~this~~`, enclosing a span in`~~`.

Handling these means:
- Parsing into `DeletedNode`
- Styling nodes that are of type `DeletedNode` with a line-through

Fixes #357

Final Result: 
Simulator Screenshot - iPhone 15 Pro Max - 2024-03-25 at 06 28 41

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -450,6 +450,10 @@ class StrongNode extends InlineContainerNode {
   const StrongNode({super.debugHtmlNode, required super.nodes});
 }
 
+class DeletedNode extends InlineContainerNode {
+  const DeletedNode({super.debugHtmlNode, required super.nodes});
+}
+
 class EmphasisNode extends InlineContainerNode {
   const EmphasisNode({super.debugHtmlNode, required super.nodes});
 }
@@ -716,6 +720,9 @@ class _ZulipContentParser {
     }
     if (localName == 'strong' && className.isEmpty) {
       return StrongNode(nodes: nodes(), debugHtmlNode: debugHtmlNode);
+    }
+    if (localName == 'del' && className.isEmpty) {
+      return DeletedNode(nodes: nodes(), debugHtmlNode: debugHtmlNode);
     }
     if (localName == 'em' && className.isEmpty) {
       return EmphasisNode(nodes: nodes(), debugHtmlNode: debugHtmlNode);
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -622,6 +622,8 @@ class _InlineContentBuilder {
       return const TextSpan(text: "");
     } else if (node is StrongNode) {
       return _buildStrong(node);
+    } else if (node is DeletedNode) {
+      return _buildDeleted(node);
     } else if (node is EmphasisNode) {
       return _buildEmphasis(node);
     } else if (node is LinkNode) {
@@ -652,6 +654,9 @@ class _InlineContentBuilder {
 
   InlineSpan _buildStrong(StrongNode node) => _buildNodes(node.nodes,
     style: weightVariableTextStyle(_context, wght: 600, wghtIfPlatformRequestsBold: 900));
+
+  InlineSpan _buildDeleted(DeletedNode node) => _buildNodes(node.nodes,
+    style: const TextStyle(decoration: TextDecoration.lineThrough));
 
   InlineSpan _buildEmphasis(EmphasisNode node) => _buildNodes(node.nodes,
     style: const TextStyle(fontStyle: FontStyle.italic));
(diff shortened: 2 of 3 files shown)


lib/model/content.dart
```diff
@@ -450,6 +450,10 @@ class StrongNode extends InlineContainerNode {
   const StrongNode({super.debugHtmlNode, required super.nodes});
 }
 
+class DeletedNode extends InlineContainerNode {
```
Khader-1, line 453: I am not sure about the name here it seems ambiguous. Any Suggestions will be very welcomed!

test/model/content_test.dart
```diff
@@ -590,11 +596,12 @@ void main() {
...
-  testParseInline('parse nested strong, em, code',
+  testParseInline('parse nested del, strong, em, code',
     // "***`word`***"
-    '<p><strong><em><code>word</code></em></strong></p>',
-    const StrongNode(nodes: [EmphasisNode(nodes: [InlineCodeNode(nodes: [
-      TextNode('word')])])]));
+    '<p><del><strong><em><code>word</code></em></strong></del></p>',
+    const DeletedNode(
+      nodes: [StrongNode(nodes: [EmphasisNode(nodes: [InlineCodeNode(nodes: [
+        TextNode('word')])])])]));
```
chrisbobbe, line 604: The code-commented Markdown—

```dart
// "***`word`***"
```

should also be updated. (Here and in multiple places below.)

chrisbobbe (commented):
Thanks, this is looking good! 🙂 One small comment below, and Greg will likely have more.

test/model/content_test.dart
Khader-1: That's right!

Khader-1: @chrisbobbe Thanks for the feedback☺️
I've opened this before getting reviews on the previous PRs by Greg. 
I've learned a lot of things that can refactor this already! I will take a second go at it then mark it for review

lib/widgets/content.dart
```diff
@@ -653,6 +655,10 @@ class _InlineContentBuilder {
   InlineSpan _buildStrong(StrongNode node) => _buildNodes(node.nodes,
     style: weightVariableTextStyle(_context, wght: 600, wghtIfPlatformRequestsBold: 900));
 
+
+  InlineSpan _buildDeleted(DeletedNode node) => _buildNodes(node.nodes,
```
gnprice, line 659: nit: just one blank line to separate:
```suggestion
    style: weightVariableTextStyle(_context, wght: 600, wghtIfPlatformRequestsBold: 900));

  InlineSpan _buildDeleted(DeletedNode node) => _buildNodes(node.nodes,
```

lib/model/content.dart
gnprice: I think I see what you mean — it sounds like it might mean a node that's been deleted somehow, like it's no longer part of the tree.

Alternative names could be `DelNode`, or `StrikethroughNode`.

I don't like the name `DelNode`, though — it feels pretty opaque. We don't name other node types based on abbreviations like that; e.g. we say `EmphasisNode`, not `EmNode`.

And although `StrikethroughNode` could be reasonable if we made a pattern of naming these based on how we're going to render them, we don't — we instead name them based on their HTML meaning, so e.g. `StrongNode` rather than `BoldNode`. It'd be awkward to have a mixture of those patterns.

So let's go with `DeletedNode`, as it seems the most canonical name here to parallel `EmphasisNode` and others. I think the confusion shouldn't be too much of a problem, because we don't actually have a concept like "deleting" a node to confuse it with. If we did, we might have to rename to avoid that confusion.

gnprice (commented):
Thanks @Khader-1 for the contribution, and thanks @chrisbobbe for the initial review!

This all looks good to me except one nit below. I'll fix that and merge.

test/model/content_test.dart
```diff
@@ -580,6 +580,12 @@ void main() {
     '<p><strong>bold</strong></p>',
     const StrongNode(nodes: [TextNode('bold')]));
 
+
+  testParseInline('parse deleted/strike-through',
```
gnprice, line 584: Oh and similarly:

```suggestion
    const StrongNode(nodes: [TextNode('bold')]));

  testParseInline('parse deleted/strike-through',
```

lib/model/content.dart
Khader-1: Thanks for you thoughts, I totally agree.

Issue #227, opened 2023-07-19
autocomplete: Add user avatars to user-mention autocompletes


chrisbobbe:

sm-sayedi:
I would like to work on this issue!

So far, my work looks like this:

<img src="https://github.com/zulip/zulip-flutter/assets/59946442/b89191b9-471e-4ebf-8dc5-784f2176b52d" width="300"/>

gnprice:
@sm-sayedi Thanks! Please go ahead.

That demo looks good, except that I'd put less vertical spacing between the items in the results. If you can find an example of this UI in @terpimost's [Figma design for mobile](https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=0-1), though, then follow what that says. @terpimost might also add a comment in this thread.

(As usual, please prioritize revisions on existing PRs over starting new ones — so that means revising #560 based on the feedback in the review I just left there. But once that's all ready for another review, this would be a good next issue.)

Issue #100, opened 2023-05-19
accounts ui: List of accounts doesn't scroll when more than a screenful


chrisbobbe:

sm-sayedi:
I would like to work on this issue!

My work so far:

<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
    </tr>
    <tr>
        <td><img src="https://github.com/zulip/zulip-flutter/assets/59946442/36ebfeb9-76b3-485f-a59e-1054bab8448a" width="300"/></td>
        <td><img src="https://github.com/zulip/zulip-flutter/assets/59946442/c8ddbe32-96e1-477d-b4ed-58e0e903f912" width="300"/></td>
    </tr>
</table>

gnprice:
Sure — please go ahead. That behavior looks good.

PR #590, opened 2024-03-25
autocomplete: Add user avatars to user-mention autocompletes

sm-sayedi:
When mentioning users, their avatars are shown beside their names in the autocomplete popup.

Fixes: #227

### DIFF
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 
+import 'content.dart';
 import 'store.dart';
 import '../model/autocomplete.dart';
 import '../model/compose.dart';
@@ -119,10 +120,11 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
 
   Widget _buildItem(BuildContext _, int index) {
     final option = _resultsToDisplay[index];
+    Widget avatar;
     String label;
     switch (option) {
       case UserMentionAutocompleteResult(:var userId):
-        // TODO(#227) avatar
+        avatar = Avatar(userId: userId, size: 32, borderRadius: 3);
         label = PerAccountStoreWidget.of(context).users[userId]!.fullName;
     }
     return InkWell(
@@ -130,8 +132,13 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
         _onTapOption(option);
       },
       child: Padding(
-        padding: const EdgeInsets.all(16.0),
-        child: Text(label)));
+        padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
+        child: Row(
+          children: [
+            avatar,
+            const SizedBox(width: 8),
+            Text(label),
+          ])));
   }
 
   @override
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -107,10 +107,6 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
         // TODO(i18n) language-appropriate space character; check active keyboard?
         //   (maybe handle centrally in `widget.controller`)
         replacementString = '${mention(store.users[userId]!, silent: intent.query.silent, users: store.users)} ';
-      case WildcardMentionAutocompleteResult():
-        replacementString = '[unimplemented]'; // TODO(#234)
-      case UserGroupMentionAutocompleteResult():
-        replacementString = '[unimplemented]'; // TODO(#233)
     }
 
     widget.controller.value = intent.textEditingValue.replaced(
@@ -128,10 +124,6 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
       case UserMentionAutocompleteResult(:var userId):
         // TODO(#227) avatar
         label = PerAccountStoreWidget.of(context).users[userId]!.fullName;
-      case WildcardMentionAutocompleteResult():
-        label = '[unimplemented]'; // TODO(#234)
-      case UserGroupMentionAutocompleteResult():
-        label = '[unimplemented]'; // TODO(#233)
     }
     return InkWell(
       onTap: () {
--- a/test/widgets/autocomplete_test.dart
+++ b/test/widgets/autocomplete_test.dart
@@ -6,6 +6,7 @@ import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/route/messages.dart';
 import 'package:zulip/model/compose.dart';
 import 'package:zulip/model/narrow.dart';
+import 'package:zulip/model/store.dart';
 import 'package:zulip/widgets/message_list.dart';
 import 'package:zulip/widgets/store.dart';
 
@@ -13,11 +14,17 @@ import '../api/fake_api.dart';
 import '../example_data.dart' as eg;
 import '../model/binding.dart';
 import '../model/test_store.dart';
+import 'content_test.dart';
 
 /// Simulates loading a [MessageListPage] and tapping to focus the compose input.
 ///
 /// Also adds [users] to the [PerAccountStore],
 /// so they can show up in autocomplete.
+///
+/// Also sets [debugNetworkImageHttpClientProvider] to return a constant image.
+///
+/// The caller must set [debugNetworkImageHttpClientProvider] back to null
+/// before the end of the test.
 Future<Finder> setupToComposeInput(WidgetTester tester, {
   required List<User> users,
 }) async {
@@ -38,6 +45,8 @@ Future<Finder> setupToComposeInput(WidgetTester tester, {
     historyLimited: false,
     messages: [message],
   ).toJson());
+
+  prepareBoringImageHttpClient();
 
   await tester.pumpWidget(
     MaterialApp(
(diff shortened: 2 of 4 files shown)


lib/widgets/autocomplete.dart
```diff
@@ -123,23 +124,31 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
 
   Widget _buildItem(BuildContext _, int index) {
     final option = _resultsToDisplay[index];
+    Widget avatar;
```
sm-sayedi, line 127: Added this as a non-nullable `Widget` and assigned it a `SizedBox.shrink` for the parts not yet implemented. Not sure if this approach is better, or making it a `Widget?` and not assigning anything for parts not yet implemented!!

gnprice: Hmm. Probably the right answer is actually that we should delete those two other result subclasses that we currently never use :-). That'd be good to do as an [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc) prep commit.

The details of them will always be available in the Git history. Could also include like one line for each of them, commented out, with their respective TODO comments moved there.

gnprice: If these instead represented some situation that is already possible and that we just didn't have avatars for (or didn't yet), then one thing we'd want to be sure to do is to get the layout right — there should be a blank space with the right size for an avatar. So that might look like using `AvatarShape` unconditionally, with a child that's either `AvatarImage` or some placeholder, potentially `SizedBox.shrink()`.

If we used just `SizedBox.shrink`, it looks like that'd get the layout wrong, because it looks below like this is being used in a context that doesn't dictate its size and expects it to provide its own size. It wouldn't currently matter, because those cases are unreachable… but better not to rely on that, hence the prep commit mentioned above.

test/widgets/autocomplete_test.dart
```diff
@@ -77,34 +89,59 @@ void main() {
       await tester.enterText(composeInputFinder, 'hello @user ');
       await tester.enterText(composeInputFinder, 'hello @user t');
       await tester.pumpAndSettle(); // async computation; options appear
+
       // "User Two" and "User Three" appear, but not "User One"
       check(tester.widgetList(find.text('User One'))).isEmpty();
+      final user1AvatarFinder =
```
sm-sayedi, line 95: Added the tests for this feature in the previous test block with the motivation that the user name and avatar are closely related to each other.

sm-sayedi:
Thanks for the review. Revision pushed with tests included.

test/widgets/autocomplete_test.dart
```diff
@@ -77,34 +89,59 @@ void main() {
...
+          findNetworkImage(store.tryResolveUrl('user1.png').toString());
+      check(tester.widgetList(user1AvatarFinder)).isEmpty();
+
       tester.widget(find.text('User Two'));
+      final user2AvatarFinder =
+          findNetworkImage(store.tryResolveUrl('user2.png').toString());
+      tester.widget(user2AvatarFinder);
```
gnprice, line 102: If we do include this new test logic as part of this same test case, though, it's important to keep the original test narrative readable. Part of keeping it readable is keeping it concise, without too many details in between each of the comments that are telling each step of the story.

Relatedly, this comment `// "User Two" and "User Three" appear, but not "User One"` is a heading visually attached to the three lines below it. When those become 14 lines including two blank lines, it becomes a lot harder to visually parse the comment as describing that whole group.

The basic strategy for keeping it concise is to abstract details into helpers. You already have these helpers `user1AvatarFinder`, etc., and so a first step would be to just pull those up into the setup part of this function (above the first narrative comment) so that this can read:
```dart
      // "User Two" and "User Three" appear, but not "User One"
      check(tester.widgetList(find.text('User One'))).isEmpty();
      check(tester.widgetList(user1AvatarFinder)).isEmpty();
      tester.widget(find.text('User Two'));
      tester.widget(user2AvatarFinder);
      tester.widget(find.text('User Three'));
      tester.widget(user3AvatarFinder);
```

That works, but I think we can do better by bringing it back down to one line per user. Try writing a helper so that this can look like:
```dart
      // "User Two" and "User Three" appear, but not "User One"
      checkUserShown(user1, expected: false);
      checkUserShown(user2, expected: true);
      checkUserShown(user3, expected: true);
```

test/widgets/autocomplete_test.dart
```diff
@@ -39,6 +41,8 @@ Future<Finder> setupToComposeInput(WidgetTester tester, {
     messages: [message],
   ).toJson());
 
+  prepareBoringImageHttpClient();
```
gnprice, line 44: With this call, it becomes part of the interface of this `setupToComposeInput` function that its caller needs to set `debugNetworkImageHttpClientProvider` back to null before the end of the test. So this function should gain a note to that effect in its dartdoc, just like `prepareBoringImageHttpClient` has.

(It's regrettable that that reset can't be done using `addTearDown`, so that callers and callers' callers wouldn't have to care like this. There's an open issue upstream about that, relating to `debugNetworkImageHttpClientProvider` and a bunch of other Flutter test knobs of a similar flavor. I don't have the link offhand, but it'd be good to track it down again and then put a comment on `prepareBoringImageHttpClient` pointing to that.)

test/widgets/content_test.dart
```diff
@@ -27,6 +27,21 @@ import 'dialog_checks.dart';
 import 'message_list_checks.dart';
 import 'page_checks.dart';
 
+/// Set [debugNetworkImageHttpClientProvider] to return a constant image.
+///
+/// Returns the [FakeImageHttpClient] that handles the requests.
+///
+/// The caller must set [debugNetworkImageHttpClientProvider] back to null
+/// before the end of the test.
+FakeImageHttpClient prepareBoringImageHttpClient() {
```
gnprice, line 36: Cool, pulling this out so it can be imported elsewhere makes sense.

It's fine to have this squashed into the main commit as you currently do, because neither this nor the rest of the commit is so long as to start getting harder to read (and because this isn't touching any of the same code, so doesn't interfere with reading the diff for the main changes). But you'd also be welcome to split it out as an NFC prep commit.

gnprice (commented):
Thanks @sm-sayedi! Generally this looks good. Various comments below, all fairly small.

sm-sayedi:
Thanks for the review. Revision pushed.

Also requested to work on [issue 100](https://github.com/zulip/zulip-flutter/issues/100#issuecomment-2020072515) simultaneously!

test/widgets/autocomplete_test.dart
```diff
@@ -65,10 +69,18 @@ void main() {
   TestZulipBinding.ensureInitialized();
 
   group('ComposeAutocomplete', () {
+
+    Finder findNetworkImage(String url) {
+      return find.byWidgetPredicate((widget) =>
+        widget is Image &&
+        widget.image is NetworkImage &&
+        (widget.image as NetworkImage).url == url);
```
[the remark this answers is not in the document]
sm-sayedi: Went with the pattern-based version! Love the concise syntax!

test/widgets/autocomplete_test.dart
sm-sayedi: I was also thinking about the same helper method!! As there will be more bits added to it in the future, specifically with the user pronouns on the way soon!

test/widgets/autocomplete_test.dart
sm-sayedi: I found a [closed issue](https://github.com/flutter/flutter/issues/36455) upstream. [This one](https://github.com/flutter/flutter/issues/36455) also seemed closer to this, but wasn't sure to add it to the `prepareBoringImageHttpClient` comment.

test/widgets/autocomplete_test.dart
```diff
@@ -65,10 +74,25 @@ void main() {
   TestZulipBinding.ensureInitialized();
 
   group('ComposeAutocomplete', () {
+
+    Finder findNetworkImage(String url) {
+      return find.byWidgetPredicate((widget) => switch(widget) {
+        Image(image: NetworkImage(url: var imageUrl)) when imageUrl == url => true,
+        _ => false,
```
gnprice, line 81: nit: because the first pattern is so long, the result "true" gets a bit hard to spot when it's off at the end of that line. So:
```suggestion
        Image(image: NetworkImage(url: var imageUrl)) when imageUrl == url
          => true,
        _ => false,
```

test/widgets/autocomplete_test.dart
```diff
@@ -77,34 +101,37 @@ void main() {
       await tester.enterText(composeInputFinder, 'hello @user ');
       await tester.enterText(composeInputFinder, 'hello @user t');
       await tester.pumpAndSettle(); // async computation; options appear
+
       // "User Two" and "User Three" appear, but not "User One"
-      check(tester.widgetList(find.text('User One'))).isEmpty();
-      tester.widget(find.text('User Two'));
-      tester.widget(find.text('User Three'));
+      checkUserShown(user1, store, expected: false);
```
gnprice, line 106: This version is fine. But for other situations where the shared context is more verbose than just `store, ` and so it's more of a pain to have to pass it each time, there are two solutions we often use:
* The helper function can be local within the test function, so that it sees the same `store` as a local.
* The context like `store` can be a `late` variable that's declared up at the top of a test group, or of the whole test `main`, and that gets initialized afresh by a helper function that each test case calls. Those helpers are often named `prepare` or of the form `prepareFoo`.

test/widgets/content_test.dart
gnprice, line 36: nit: prefix for this commit:

f3aebc9fd autocomplete [nfc]: Pull `prepareBoringImageHttpClient` out to toplevel

should be `autocomplete test [nfc]:`, to indicate that it's only touching the test code

(And then the "nfc" part means that on top of that, it has only an NFC effect on even the test code.)

gnprice (commented):
Thanks! This looks great — just two nits below.

I'll fix those up and merge.

test/widgets/content_test.dart
gnprice: Oh and looking again, it's not really about the autocomplete tests in particular, so `autocomplete` isn't quite right.

I'll make it `content test [nfc]`, since it is in that file. Could also have made it just `test [nfc]`, since the point is it's now exporting this for other tests.

test/widgets/autocomplete_test.dart
gnprice: Thanks for looking! Found the one I was thinking of (which it turns out is one I filed :slightly_smiling_face:), and added a comment: 68116a247beb744abaf0d8fde32e792152e0ec18

Looks like there was a backlink to it in the second issue you found:
image
as that issue is basically a different specific case of the issue I had in mind.

sm-sayedi:
Thanks @gnprice for all this and for the instructive comments!!

Issue #351, opened 2023-10-31
Replace flutter_local_notifications with a thin API binding using Pigeon


gnprice:
For our initial implementation of notification UI #122 (introduced in #344), we use the package `flutter_local_notifications`. That's because it's what's currently [recommended by Flutter upstream](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/notification.20libraries/near/1670987) for showing a notification, and because I looked around a bit a few months ago and concluded that it was the best existing option for that.

As discussed at https://github.com/zulip/zulip-flutter/pull/344#discussion_r1375012273 and [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/notification.20libraries/near/1670982), though, that package has quite a bit of baggage for functionality we'll never need, like timers. It also goes for an unfortunate pattern — reminiscent of React Native and its ecosystem, in fact — of providing a bunch of logic of its own, so that the API it presents is different from that of the underlying platform and adds its own quirks, bugs, and lack of documentation. For example, the issue I filed which is discussed at https://github.com/zulip/zulip-flutter/pull/344#discussion_r1375041646 .

So I don't find this package very satisfying, and I think it's likely that we'll ultimately want to rip it out in favor of something we write ourselves. The plan would be:
* Expose the platform APIs as thinly and faithfully as possible. Let the logic of deciding how to use those APIs stay in Dart.
* Use Pigeon to make the writing of those API bindings as easy and automated as possible.

gnprice:
As described at https://github.com/zulip/zulip-flutter/issues/341#issuecomment-2014183497 , I started trying Pigeon today and wrote a draft branch that does this for the method to show a notification, as well as a couple of other methods we don't yet use in main. So I'll plan to clean that up and send it.

I might go ahead and cover the rest of the library (since what I already have is the most complex of the methods we use), or might leave that for later.

rajveermalviya:
(Just another data point of why it would be better to have our custom platform integration for some APIs)

Yesterday I experimented with #128 and my [draft branch](https://github.com/zulip/zulip-flutter/compare/main...rajveermalviya:zulip-flutter:pr-android-inbox-style-notif) uses `flutter_local_notification` because it provides all the functionality that is needed to implement #128.

[...]

PR #592, opened 2024-03-26
notif: Switch to our own Pigeon-based plugin for showing a notification

gnprice:
This delivers the first major tranche of #351, switching from
package:flutter_local_notifications to our own thin API binding
based on Pigeon. In particular we convert over one of
the handful of different methods we use from that package,
and it's the most complex of them by a substantial margin.

Getting this to work involves a bit of tricky, but fortunately
one-time, infrastructure: a shim extra Flutter plugin in order
to get our new Flutter plugin properly wired up.

Fixes-partly: #351

/cc @rajveermalviya


chrisbobbe:
```
FAILED: analyze build_runner
```

gnprice:
Ah thanks.

The analyze failure is this issue I reported upstream in Pigeon:
* https://github.com/flutter/flutter/issues/145633

but that I then forgot about when sending this. I'll work around probably by just disabling the lint globally; I doubt many cases of it will accumulate before someone (possibly us) gets around to fixing that Pigeon issue.

The build_runner failure looks like I need to tweak that check to know that not all `*.g.dart` files come from build_runner.

I'll fix those, but meanwhile I think all the existing code in the PR is ready for review.

gnprice:
Fixed those issues — CI is now passing.

chrisbobbe:
Cool! It's exciting to start using Pigeon.

In manual testing on the office Android device (Samsung Galaxy S9 running Android 9), I don't seem to be getting notifications when the app is closed. They do appear when the app is open and foregrounded, or open in the background.

I'm not sure if this behavior is new with these commits, though. I'll check and see what's happening on `main`.

chrisbobbe:
Ah yeah, I see the same thing happening on `main`. So I guess there's an existing bug. Is it possible this is related to #342?

gnprice:
Yeah, I actually saw similar behavior when I was testing this yesterday, and similarly determined it was also present on main :-/ I have a local note to myself but hadn't gotten to investigating further.

It's probably yet another distinct cause from #342, but definitely a similar symptom.

One potential mitigating factor is that I think I've only seen that in debug builds, not when using `flutter run --release`. If that pattern holds, then it doesn't affect normal users. But I don't know of a good reason why it'd have that pattern.

chrisbobbe:
Oh, FWIW the issue seems to have stopped reproducing for me. I'm still on `main`. Here is my sequence:

- Freshly install a debug build from `main`, open the app, log in
- Cause a notification via DM; it appears (reported above)
- Put the app in the background
- Cause a notification via DM; it appears (reported above)
- Close the app
- Cause a notification via DM; it does not appear (reported above)
- Open the app again
- Cause a notification via DM; it appears
- Put the app in the background
- Cause a notification via DM; it appears
- Close the app
- Cause a notification via DM; it appears! (unexpected)

Did you happen to notice something similar?

chrisbobbe:
Anyway, the issue is present in `main` so shouldn't block merging this. Please merge at will. 🙂

gnprice:
> Oh, FWIW the issue seems to have stopped reproducing for me. I'm still on `main`.

Yeah, the issue has been flaky in my observation — sometimes it reproduces, sometimes not. That's the main reason I'm not confident in the "only debug builds" pattern I've observed.


> Here is my sequence:
> 
> * Freshly install a debug build from `main`, open the app, log in
> * Cause a notification via DM; it appears (reported above)
> * Put the app in the background
> * Cause a notification via DM; it appears (reported above)
> * Close the app
> * Cause a notification via DM; it does not appear (reported above)

Yeah, this is exactly the sequence that I did many times in the course of developing these changes (repeating it after various changes to the branch, to check that things were still working).

I suspect a trigger for it might be that closing the app is taken as a bad sign by the system — the user force-closed the app! maybe it was misbehaving for them — and so because I was doing it again and again, the system eventually decided the app shouldn't be trusted with the privilege of starting up in the background. If so, then that'd be another mitigating factor in that it's not likely to happen with users who aren't developing the app.

gnprice:
Merged! Thanks for the review.

gnprice:
And filed #597 for that issue where notifications sometimes don't appear.

PR #593, opened 2024-03-26
accounts-ui: Make accounts list scrollable

sm-sayedi:
Before this fix, the list of accounts did not scroll when they were more than a screenful and would show an overflow error on the screen.

After this fix, the list of accounts scrolls properly with no overflow error and without shifting other content offscreen.

<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

Fixes: #100

### DIFF
--- a/test/flutter_checks.dart
+++ b/test/flutter_checks.dart
@@ -5,6 +5,13 @@ import 'package:checks/checks.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
+extension RectChecks on Subject<Rect> {
+  Subject<double> get top => has((d) => d.top, 'top');
+  Subject<double> get bottom => has((d) => d.bottom, 'bottom');
+
+  // TODO others
+}
+
 extension AnimationChecks<T> on Subject<Animation<T>> {
   Subject<AnimationStatus> get status => has((d) => d.status, 'status');
   Subject<T> get value => has((d) => d.value, 'value');
--- a/test/widgets/app_test.dart
+++ b/test/widgets/app_test.dart
@@ -1,11 +1,15 @@
 import 'package:checks/checks.dart';
-import 'package:flutter/widgets.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:zulip/model/database.dart';
 import 'package:zulip/widgets/app.dart';
 import 'package:zulip/widgets/inbox.dart';
 import 'package:zulip/widgets/page.dart';
+import 'package:zulip/widgets/store.dart';
 
 import '../example_data.dart' as eg;
+import '../flutter_checks.dart';
 import '../model/binding.dart';
 import '../test_navigation.dart';
 import 'page_checks.dart';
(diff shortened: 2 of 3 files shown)


test/widgets/app_test.dart
```diff
@@ -50,5 +53,60 @@ void main() {
...
+          localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+          supportedLocales: ZulipLocalizations.supportedLocales,
+          home: GlobalStoreWidget(
+            child: ChooseAccountPage())));
+
+      // global store gets loaded
+      await tester.pumpAndSettle();
+    }
+
+    Iterable<Account> dummyAccounts(int count) sync* {
```
sm-sayedi, line 78: Is this better here or moved to the `example_data.dart` file?

test/widgets/app_test.dart
```diff
@@ -51,4 +54,56 @@ void main() {
...
+          home: GlobalStoreWidget(
+            child: ChooseAccountPage())));
+
+      // global store gets loaded
+      await tester.pumpAndSettle();
+    }
+
+    List<Account> dummyAccounts(int count) {
+      return List.generate(count, (i) => eg.account(
+        id: i,
```
sm-sayedi, line 82: Is this better here or moved to the `example_data.dart` file?

gnprice: Here is good. The `example_data.dart` file is good for things that are likely to be reused across a broad range of tests; but I think it'll be fairly uncommon that we want to test having a lot of accounts.

test/widgets/app_test.dart
```diff
@@ -51,4 +54,56 @@ void main() {
...
+          localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+          supportedLocales: ZulipLocalizations.supportedLocales,
+          home: GlobalStoreWidget(
+            child: ChooseAccountPage())));
+
+      // global store gets loaded
+      await tester.pumpAndSettle();
+    }
+
+    List<Account> dummyAccounts(int count) {
```
gnprice, line 80:
```suggestion
    List<Account> generateAccounts(int count) {
```

Better to avoid "dummy"; it has another meaning as an insult (literally "person who can't talk", commonly used for "stupid person").

test/widgets/app_test.dart
```diff
@@ -51,4 +54,56 @@ void main() {
       ]);
     });
   });
+
+  group('choose account page', () {
```
gnprice, line 58: nit:
```suggestion
  group('ChooseAccountPage', () {
```

Using the identifier that's also found in the code under test helps make it easy to find the test from the code tested and vice versa.

test/widgets/app_test.dart
```diff
@@ -51,4 +54,56 @@ void main() {
...
+        id: i,
+        user: eg.user(fullName: 'User $i', email: 'user$i@example'),
+        apiKey: 'user${i}apikey',
+      ));
+    }
+
+    Finder findAccount(Account account) => find.text(account.email).hitTestable();
+
+    void checkAccountShown(Account account, {required bool expected}) {
+      check(findAccount(account).evaluate().length).equals(expected ? 1 : 0);
```
gnprice, line 91: nit: do more processing on the result of `check`, instead of its argument:
```suggestion
      check(findAccount(account).evaluate()).length.equals(expected ? 1 : 0);
```

This allows `package:checks` to give more informative output if the check fails.

test/widgets/app_test.dart
```diff
@@ -51,4 +54,56 @@ void main() {
...
+    }
+
+    testWidgets('accounts list is scrollable when more than a screenful', (tester) async {
```
gnprice, line 94: This test looks good.

Let's also add a test for the case of just one account, checking that the layout is centered. For that, can just check that both the one account item and the add-account button are located, say, at least 1/3 of the screen's height away from both the top and bottom of the screen.

That makes a good regression test for something that an alternate implementation of this feature could easily have broken if it were buggy.

Similarly let's have one for the case with zero accounts, which just has the add-account button.

gnprice (commented):
Thanks @sm-sayedi! Generally this looks great; several small comments on the tests.

sm-sayedi:
Thanks for the review! Revision pushed. PTAL!

test/widgets/app_test.dart
```diff
@@ -51,4 +54,113 @@ void main() {
       ]);
     });
   });
+
+  group('ChooseAccountPage', () {
+    Future<void> setupChooseAccountPage(WidgetTester tester, {
+      required List<Account>? accounts,
```
gnprice, line 60: nit: requiring the caller to explicitly pass this makes sense, but then it may as well just take a list:
```suggestion
      required List<Account> accounts,
```

so the caller explicitly says `[]`, instead of saying `null` and having that mean `[]`.

(In fact, now that I think of that, even in the version where this parameter was optional it probably should have been `List<Account> accounts = [],` rather than being nullable.)

test/widgets/app_test.dart
```diff
@@ -51,4 +54,113 @@ void main() {
...
+    void checkButtonShown<T extends ButtonStyleButton>({
+      required String withText,
+      required bool expected,
+    }) {
+      check(findButton<T>(withText: withText).evaluate())
+        .length.equals(expected ? 1 : 0);
+    }
+
+    testWidgets('accounts list is scrollable when more than a screenful',
+      (tester) async {
```
gnprice, line 109: nit: we just leave the `(tester) async {` part on the same line even if that makes the line long:

```suggestion
    testWidgets('accounts list is scrollable when more than a screenful', (tester) async {
```

Because that part is totally boring and predictable, it's OK if it gets scrolled off screen for the reader.

test/widgets/app_test.dart
```diff
@@ -51,4 +54,113 @@ void main() {
...
+      final accounts = generateAccounts(15);
```
gnprice, line 110: (If OTOH we had a situation where it wasn't so boring and predictable and did need to go on the next line, we'd want to avoid having the `(tester) async {` aligned with the body, because it makes it look like part of the body. For a function, the standard way to do that is to indent the body:
```dart
    testWidgets('accounts list is scrollable when more than a screenful',
      (tester) async {
        final accounts = generateAccounts(15);
```
)

test/widgets/app_test.dart
```diff
@@ -51,4 +54,113 @@ void main() {
...
+      checkAccountShown(account, expected: true);
+      checkButtonShown(withText: buttonText, expected: true);
+
+      final screenHeight =
+        (tester.view.physicalSize / tester.view.devicePixelRatio).height;
+      final thirdOfScreenHeight = screenHeight / 3;
+
+      final accountRect = tester.getRect(findAccount(account));
+      check(accountRect.top).isGreaterThan(thirdOfScreenHeight);
+      check(accountRect.bottom).isLessThan(screenHeight - thirdOfScreenHeight);
```
gnprice, line 143: Let's write this in more thoroughly `package:checks` style:
```suggestion
      check(tester.getRect(findAccount(account)))
        ..top.isGreaterThan(thirdOfScreenHeight)
        ..bottom.isLessThan(screenHeight - thirdOfScreenHeight);
```

You'll need to write those accessors `top` and `bottom` , since we haven't yet. They'll go in `test/flutter_checks.dart`; see existing examples there.

gnprice, line 143: Separately, for the numbers, let's just write:
```suggestion
      check(accountRect.top).isGreaterThan(1/3 * screenHeight);
      check(accountRect.bottom).isLessThan(2/3 * screenHeight);
```

For me at least, that's clearer to read than the more word-based version.

gnprice (commented):
Thanks for the revision! Generally this looks great; a few small comments below.

sm-sayedi:
Thanks for the review! Pushed the revision. Please have a look!

test/flutter_checks.dart
```diff
@@ -5,6 +5,12 @@ import 'package:checks/checks.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
+extension RectChecks on Subject<Rect> {
+  Subject<double> get top => has((d) => d.top, 'top');
+  Subject<double> get bottom => has((d) => d.bottom, 'bottom');
+
+  // TODO others
+}
 extension AnimationChecks<T> on Subject<Animation<T>> {
```
gnprice, line 14: nit:
```suggestion
}

extension AnimationChecks<T> on Subject<Animation<T>> {
```

gnprice (commented):
One nit below; otherwise all looks good. I'll fix that nit and merge. Thanks @sm-sayedi for adding this!

Issue #573, opened 2024-03-19
notif: Translate "to you and N others", handling number properly


gnprice:
See [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications.20differences/near/1757473).

The report is only about fixing the singular/plural handling in English: we currently show "and 1 others". But to fix it, we should just internationalize this string, and then the number handling will get fixed automatically as part of that.

There's an existing TODO comment in the code for this:
```dart
        '${data.senderFullName} to you and ${allRecipientIds.length - 2} others', // TODO(i18n), also plural; […]
```

abelaba:
@gnprice. I can get started on this.

PR #594, opened 2024-03-27
notif: Internationalize group DM notification message and

abelaba:
handle number properly

This commit addresses the issue of handling plural and singular forms of the string when a group DM notification message is received. The previous implementation used a hard-coded string. Now it uses an internationalized string.

I have also added a test so that both the singular and plural forms of the string are tested.

Fixes: #573

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -440,5 +440,13 @@
   "allMessagesPageTitle": "All messages",
   "@allMessagesPageTitle": {
     "description": "Title for the page of all messages"
+  },
+  "notifGroupDmConversationLabel": "{senderFullName} to you and {numOthers, plural, =1{1 other} other{{numOthers} others}}",
+  "@notifGroupDmConversationLabel": {
+    "description": "Label for a group DM conversation notification.",
+    "placeholders": {
+      "senderFullName": {"type": "String", "example": "Alice"},
+      "numOthers": {"type": "int", "example": "4"}
+    }
   }
 }
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -10,6 +10,7 @@ import '../api/notifications.dart';
 import '../host/android_notifications.dart';
 import '../log.dart';
 import '../model/binding.dart';
+import '../model/localizations.dart';
 import '../model/narrow.dart';
 import '../widgets/app.dart';
 import '../widgets/message_list.dart';
@@ -89,13 +90,15 @@ class NotificationDisplayManager {
 
   static void _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) {
     assert(debugLog('notif message content: ${data.content}'));
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
     final title = switch (data.recipient) {
       FcmMessageStreamRecipient(:var streamName?, :var topic) =>
         '$streamName > $topic',
       FcmMessageStreamRecipient(:var topic) =>
         '(unknown stream) > $topic', // TODO get stream name from data
       FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 =>
-        '${data.senderFullName} to you and ${allRecipientIds.length - 2} others', // TODO(i18n), also plural; TODO use others' names, from data
+        zulipLocalizations.notifGroupDmConversationLabel(
+          data.senderFullName, allRecipientIds.length - 2), // TODO use others' names, from data
       FcmMessageDmRecipient() =>
         data.senderFullName,
     };
--- a/test/notifications/display_test.dart
+++ b/test/notifications/display_test.dart
@@ -169,11 +169,20 @@ void main() {
         expectedTagComponent: 'stream:${message.streamId}:${message.subject}');
     }));
 
-    test('group DM', () => awaitFakeAsync((async) async {
+    test('group DM: 3 users', () => awaitFakeAsync((async) async {
       await init();
       final message = eg.dmMessage(from: eg.thirdUser, to: [eg.otherUser, eg.selfUser]);
       await checkNotifications(async, messageFcmMessage(message),
-        expectedTitle: "${eg.thirdUser.fullName} to you and 1 others",
+        expectedTitle: "${eg.thirdUser.fullName} to you and 1 other",
+        expectedTagComponent: 'dm:${message.allRecipientIds.join(",")}');
+    }));
+
+    test('group DM: more than 3 users', () => awaitFakeAsync((async) async {
+      await init();
+      final message = eg.dmMessage(from: eg.thirdUser,
+        to: [eg.otherUser, eg.selfUser, eg.fourthUser]);
+      await checkNotifications(async, messageFcmMessage(message),
+        expectedTitle: "${eg.thirdUser.fullName} to you and 2 others",
         expectedTagComponent: 'dm:${message.allRecipientIds.join(",")}');
     }));
 


lib/notifications.dart
```diff
@@ -283,14 +284,16 @@ class NotificationDisplayManager {
   }
 
   static void _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
```
abelaba, line 287: On the translations.md documentation it says to use `ZulipLocalizations.of(context)` to access the localization object, but I haven't been able to get the `BuildContext` here. Is there a better way to handle this @gnprice?

gnprice: Yeah. This is a context where there's no associated widget in Flutter's trees, so there's no relevant BuildContext and we should just use this global value.

assets/l10n/app_en.arb
```diff
@@ -440,5 +440,13 @@
   "allMessagesPageTitle": "All messages",
   "@allMessagesPageTitle": {
     "description": "Title for the page of all messages"
+  },
+  "groupDMNotificationMessage": "{senderFullName} to you and {num, plural, =1{1 other} other{{num} others}}.",
```
gnprice, line 444:
```suggestion
  "notifGroupDmConversationLabel": "{senderFullName} to you and {num, plural, =1{1 other} other{{num} others}}.",
```

* In camel-cased identifier names, write acronyms like `…Dm…`, not `…DM…`.
* For this file which has a big list of translatable strings from across the whole app, giving things a prefix that relates to what part of the app it's about is helpful for organization.
* "Message" is confusing here, because it sounds like it's about the Zulip message that someone sent and caused the notification.

test/notifications_test.dart
```diff
@@ -166,11 +167,24 @@ void main() {
         expectedTagComponent: 'stream:${message.streamId}:${message.subject}');
     }));
 
-    test('group DM', () => awaitFakeAsync((async) async {
+    test('group DM: 2 users', () => awaitFakeAsync((async) async {
       await init();
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
       final message = eg.dmMessage(from: eg.thirdUser, to: [eg.otherUser, eg.selfUser]);
```
gnprice, line 173: This is a good separate test case to have.

But there are 3 users here, not 2 :slightly_smiling_face: — so the test name should reflect that.

assets/l10n/app_en.arb
gnprice: Separately on this line:

This added a period at the end of the string. That isn't in the previous version, and it doesn't make sense in context where this is used in the UI.

test/notifications_test.dart
```diff
@@ -166,11 +167,24 @@ void main() {
...
       await checkNotifications(async, messageFcmMessage(message),
-        expectedTitle: "${eg.thirdUser.fullName} to you and 1 others",
+        expectedTitle: zulipLocalizations.groupDMNotificationMessage(
+          eg.thirdUser.fullName, message.allRecipientIds.length - 2),
```
gnprice, line 176: Given that the point of this test case is to check that we get the singular right, it's better to have the expected text be explicit here. It means that if the text changes, we'll need to update the test, but that's worth it when the text is the focus of the test case.

assets/l10n/app_en.arb
```diff
@@ -440,5 +440,13 @@
...
+  "@groupDMNotificationMessage": {
+    "description": "Notification message for a group DM.",
+    "placeholders": {
+      "senderFullName": {"type": "String", "example": "Alice"},
+      "num": {"type": "int", "example": "4"}
```
gnprice, line 449: The name "num" isn't very helpful here as to which number the caller should pass. After all, the right answer involves subtracting 2 from the number it first lays hands on (which is the total number of users in the conversation).

Instead, `numOthers` would be a good name.

gnprice (commented):
Thanks @abelaba for taking care of this! Comments below.

test/notifications_test.dart
abelaba: Fixed

test/notifications_test.dart
abelaba: Okay. Fixed.

assets/l10n/app_en.arb
abelaba: Fixed. I have updated it with `numOthers`.

assets/l10n/app_en.arb
abelaba: Yes, I was going to ask you if I should add a period at the end of the sentence when I got the PR out of draft. Thanks for the comment.

test/notifications/display_test.dart
```diff
@@ -169,11 +169,20 @@ void main() {
...
+    test('group DM: 3 users', () => awaitFakeAsync((async) async {
       await init();
       final message = eg.dmMessage(from: eg.thirdUser, to: [eg.otherUser, eg.selfUser]);
       await checkNotifications(async, messageFcmMessage(message),
-        expectedTitle: "${eg.thirdUser.fullName} to you and 1 others",
+        expectedTitle: "${eg.thirdUser.fullName} to you and 1 other",
+        expectedTagComponent: 'dm:${message.allRecipientIds.join(",")}');
+    }));
+
+    test('group DM: more than 3 users', () => awaitFakeAsync((async) async {
```
abelaba, line 180: Should I label this test as '4 users' rather than 'more than 3 users' or use another label for the test? My idea for this test is to test the plural form of the string when there are more than 3 users in the conversation that is why I labeled it as such.

gnprice: Yeah, that reasoning sounds good to me. The important thing for the test to do its job is that there are more than 3 users; having exactly 4 is just a detail of how the test happens to be implemented.

lib/notifications/display.dart
```diff
@@ -89,13 +90,15 @@ class NotificationDisplayManager {
...
     assert(debugLog('notif message content: ${data.content}'));
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
     final title = switch (data.recipient) {
       FcmMessageStreamRecipient(:var streamName?, :var topic) =>
         '$streamName > $topic',
       FcmMessageStreamRecipient(:var topic) =>
         '(unknown stream) > $topic', // TODO get stream name from data
       FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 =>
-        '${data.senderFullName} to you and ${allRecipientIds.length - 2} others', // TODO(i18n), also plural; TODO use others' names, from data
+        zulipLocalizations.notifGroupDmConversationLabel(
```
gnprice, line 100: nit in commit message: summary line should be kept on one line:

    notif: Internationalize group DM notification conversation
    label
    
    This commit addresses the issue of handling plural and
    […]

See https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-messages .

Just wrapping the same text onto a single line does it:
`notif: Internationalize group DM notification conversation label`
as that's only 64 columns wide. Commit summary lines can go up to a bit over 70 columns.

But we can improve the summary line further by tightening some: "notification" is covered by the prefix, and "internationalize" is a long word that has… well, the related term "internationalization" has a standard abbreviation. So:
`notif: Add i18n for conversation label on group DMs`

gnprice (commented):
Thanks @abelaba! One nit below.

Then before I actually merge, I want to try it out myself in a manual end-to-end test, because notifications are an area that's been subject to unexpected failures and so I want to be a bit paranoid about changes there. It's late in the evening here now, so I won't get to that at the moment but I'll aim to do it tomorrow.

lib/notifications/display.dart
abelaba: Fixed.

gnprice:
OK, tested and all looks good — merged. Thanks again @abelaba!

Issue #109, opened 2023-05-23
login: Link to doc to help users understand what a "server URL" is and how to find theirs


chrisbobbe:
Like we did in zulip/zulip-mobile@52288facc.

VatsalBhesaniya:
Hi @chrisbobbe,
I'm interested in working on this issue.

I have reviewed how this is implemented in [zulip/zulip-mobile@`52288facc`](https://github.com/zulip/zulip-mobile/commit/52288facc).

I noticed that in Flutter `onTap` event of the `InputDecoration` `label` does not work due to the onClick behavior focusing the text field.

To address this, I propose two possible solutions:
Solution 1(Separate Text Widget)
We can add a separate `Text` widget positioned above the `TextField`, allowing for a tappable link, and do not use the default label from InputDecoration.

https://github.com/zulip/zulip-flutter/assets/38466275/f8e1b616-9afc-4241-808c-899a56acdabe

Solution 2(TextField Suffix)
Alternatively, we could leverage a TextField suffix widget containing a link. However, this approach has the limitation of the link being visible only when the field is focused.
Alternatively, we can add a `suffix` widget in `TextField` containing a link. However, this approach has the limitation of the link being visible only when the `TextField` is focused.

https://github.com/zulip/zulip-flutter/assets/38466275/82a78d2b-74b5-4a0b-975b-6a516b78b944

I'm open to any improvements to these options and exploring any other suggestions you might have.

chrisbobbe:
Interesting; thanks for your thoughts on this!

I wonder about using [`InputDecoration.helperText`](https://api.flutter.dev/flutter/material/InputDecoration/helperText.html). The doc says:

> Text that provides context about the [InputDecorator.child](https://api.flutter.dev/flutter/material/InputDecorator/child.html)'s value, such as how the value will be used.

which seems like a pretty good match for what we want here.

…Hmm, but it looks like that asks for a string, and there isn't a similar parameter to specify a Widget. So it looks like `InputDecoration` doesn't currently support putting a clickable link in that position. It would be convenient if it had a param with a name like `helper`, on the pattern of:

- [`InputDecoration.errorText`](https://api.flutter.dev/flutter/material/InputDecoration/errorText.html) / [`InputDecoration.error`](https://api.flutter.dev/flutter/material/InputDecoration/error.html) (`error` was added in flutter/flutter@8d837062a)
[...]

PR #595, opened 2024-03-27
Add helper text for "server URL" with link to documentation

VatsalBhesaniya:
Add helper text below the "server URL" field that, when tapped, opens relevant Zulip documentation explaining what a server URL is and how to find theirs. It will help users understand the "server URL" field during login.

Fixes: [#109](https://github.com/zulip/zulip-flutter/issues/109)

https://github.com/zulip/zulip-flutter/assets/38466275/2258cbfa-43d6-4339-b50c-c7967ff1472e

Simulator Screenshot - iPhone 15 Pro Max - 2024-03-29 at 10 55 02

https://github.com/zulip/zulip-flutter/assets/38466275/bb9f5144-7b5c-4ea0-bbb1-6a8f59da1cfc

Screenshot_1711685963

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -308,6 +308,10 @@
   "@loginServerUrlInputLabel": {
     "description": "Input label in login page for Zulip server URL entry."
   },
+  "serverUrlDocLinkLabel": "What's this?",
+  "@serverUrlDocLinkLabel": {
+    "description": "Link to doc to help users understand what a server URL is and how to find theirs."
+  },
   "errorUnableToOpenLinkTitle": "Unable to open link",
   "@errorUnableToOpenLinkTitle": {
     "description": "Error title when a link fails to open."
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -308,6 +308,17 @@
   "@loginServerUrlInputLabel": {
     "description": "Input label in login page for Zulip server URL entry."
   },
+  "errorUnableToOpenLinkTitle": "Unable to open link",
+  "@errorUnableToOpenLinkTitle": {
+    "description": "Error title when a link fails to open."
+  },
+  "errorUnableToOpenLinkMessage": "Link could not be opened: {url}",
+  "@errorUnableToOpenLinkMessage": {
+    "description": "Error message when a link fails to open.",
+    "placeholders": {
+      "url": {"type": "String", "example": "http://example.com/"}
+    }
+  },
   "loginHidePassword": "Hide password",
   "@loginHidePassword": {
     "description": "Icon label for button to hide password in input form."
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -1,7 +1,5 @@
-import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 import 'package:html/dom.dart' as dom;
 import 'package:intl/intl.dart';
 import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
@@ -9,14 +7,11 @@ import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import '../api/core.dart';
 import '../api/model/model.dart';
 import '../model/avatar_url.dart';
-import '../model/binding.dart';
 import '../model/content.dart';
-import '../model/internal_link.dart';
 import 'code_block.dart';
-import 'dialog.dart';
 import 'icons.dart';
+import 'launch_url.dart';
 import 'lightbox.dart';
-import 'message_list.dart';
 import 'store.dart';
 import 'text.dart';
 
@@ -437,7 +432,7 @@ class MessageEmbedVideo extends StatelessWidget {
     final previewImageSrcUrl = store.tryResolveUrl(node.previewImageSrcUrl);
 
     return MessageMediaContainer(
-      onTap: () => _launchUrl(context, node.hrefUrl),
+      onTap: () => launchUrlWithRealm(context, node.hrefUrl),
       child: Stack(
         alignment: Alignment.center,
         children: [
@@ -609,7 +604,7 @@ class _BlockInlineContainerState extends State<_BlockInlineContainer> {
 
   void _prepareRecognizers() {
     _recognizers.addEntries(widget.links.map((node) => MapEntry(node,
-      TapGestureRecognizer()..onTap = () => _launchUrl(context, node.url))));
+      TapGestureRecognizer()..onTap = () => launchUrlWithRealm(context, node.url))));
   }
 
   void _disposeRecognizers() {
@@ -998,52 +993,6 @@ class GlobalTime extends StatelessWidget {
               const SizedBox(width: 1),
               Text(text, style: surroundingTextStyle),
             ]))));
-  }
-}
-
-void _launchUrl(BuildContext context, String urlString) async {
-  Future<void> showError(BuildContext context, String? message) {
-    return showErrorDialog(context: context,
-      title: 'Unable to open link',
-      message: [
-        'Link could not be opened: $urlString',
-        if (message != null) message,
-      ].join("\n\n"));
-  }
-
-  final store = PerAccountStoreWidget.of(context);
-  final url = store.tryResolveUrl(urlString);
-  if (url == null) { // TODO(log)
-    await showError(context, null);
-    return;
-  }
-
-  final internalNarrow = parseInternalLink(url, store);
-  if (internalNarrow != null) {
-    Navigator.push(context,
-      MessageListPage.buildRoute(context: context,
-        narrow: internalNarrow));
-    return;
-  }
-
-  bool launched = false;
-  String? errorMessage;
-  try {
-    launched = await ZulipBinding.instance.launchUrl(url,
-      mode: switch (defaultTargetPlatform) {
-        // On iOS we prefer LaunchMode.externalApplication because (for
-        // HTTP URLs) LaunchMode.platformDefault uses SFSafariViewController,
-        // which gives an awkward UX as described here:
-        //  https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1169118
-        TargetPlatform.iOS => UrlLaunchMode.externalApplication,
-        _ => UrlLaunchMode.platformDefault,
-      });
-  } on PlatformException catch (e) {
-    errorMessage = e.message;
-  }
-  if (!launched) { // TODO(log)
-    if (!context.mounted) return;
-    await showError(context, errorMessage);
   }
 }
 
--- a/lib/widgets/launch_url.dart
+++ b/lib/widgets/launch_url.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 
 import '../model/binding.dart';
 import '../model/internal_link.dart';
@@ -10,11 +11,12 @@ import 'store.dart';
 
 /// Handles showing an error dialog with a customizable message.
 Future<void> _showError(BuildContext context, String? message, String urlString) {
+  final zulipLocalizations = ZulipLocalizations.of(context);
   return showErrorDialog(
     context: context,
-    title: 'Unable to open link',
+    title: zulipLocalizations.errorUnableToOpenLinkTitle,
     message: [
-      'Link could not be opened: $urlString',
+      zulipLocalizations.errorUnableToOpenLinkMessage(urlString),
       if (message != null) message,
     ].join("\n\n"));
 }
--- a/lib/widgets/launch_url.dart
+++ b/lib/widgets/launch_url.dart
@@ -1,0 +1,62 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+import '../model/binding.dart';
+import '../model/internal_link.dart';
+import 'dialog.dart';
+import 'message_list.dart';
+import 'store.dart';
+
+/// Handles showing an error dialog with a customizable message.
+Future<void> _showError(BuildContext context, String? message, String urlString) {
+  return showErrorDialog(
+    context: context,
+    title: 'Unable to open link',
+    message: [
+      'Link could not be opened: $urlString',
+      if (message != null) message,
+    ].join("\n\n"));
+}
+
+/// Launches a URL without considering a realm base URL.
+void launchUrlWithoutRealm(BuildContext context, Uri url) async {
+  bool launched = false;
+  String? errorMessage;
+  try {
+    launched = await ZulipBinding.instance.launchUrl(url,
+      mode: switch (defaultTargetPlatform) {
+        // On iOS we prefer LaunchMode.externalApplication because (for
+        // HTTP URLs) LaunchMode.platformDefault uses SFSafariViewController,
+        // which gives an awkward UX as described here:
+        //  https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1169118
+        TargetPlatform.iOS => UrlLaunchMode.externalApplication,
+        _ => UrlLaunchMode.platformDefault,
+      });
+  } on PlatformException catch (e) {
+    errorMessage = e.message;
+  }
+  if (!launched) {
+    if (!context.mounted) return;
+    await _showError(context, errorMessage, url.toString());
+  }
+}
+
+/// Launches a URL considering a realm base URL.
+void launchUrlWithRealm(BuildContext context, String urlString) async {
+  final store = PerAccountStoreWidget.of(context);
+  final url = store.tryResolveUrl(urlString);
+  if (url == null) { // TODO(log)
+    await _showError(context, null, urlString);
+    return;
+  }
+
+  final internalNarrow = parseInternalLink(url, store);
+  if (internalNarrow != null) {
+    Navigator.push(context,
+      MessageListPage.buildRoute(context: context, narrow: internalNarrow));
+    return;
+  }
+
+  launchUrlWithoutRealm(context, url);
+}
--- a/lib/widgets/login.dart
+++ b/lib/widgets/login.dart
@@ -15,6 +15,7 @@ import '../model/store.dart';
 import 'app.dart';
 import 'dialog.dart';
 import 'input.dart';
+import 'launch_url.dart';
 import 'page.dart';
 import 'store.dart';
 import 'text.dart';
@@ -107,6 +108,9 @@ class ServerUrlTextEditingController extends TextEditingController {
 
 class AddAccountPage extends StatefulWidget {
   const AddAccountPage({super.key});
+
+  static final Uri serverUrlHelpUrl =
+    Uri.parse('https://zulip.com/help/logging-in#find-the-zulip-log-in-url');
 
   static Route<void> buildRoute() {
     return _LoginSequenceRoute(page: const AddAccountPage());
@@ -213,7 +217,6 @@ class _AddAccountPageState extends State<AddAccountPage> {
           child: ConstrainedBox(
             constraints: const BoxConstraints(maxWidth: 400),
             child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
-              // TODO(#109) Link to doc about what a "server URL" is and how to find it
               // TODO(#111) Perhaps give tappable realm URL suggestions based on text typed so far
               TextField(
                 controller: _controller,
@@ -229,7 +232,17 @@ class _AddAccountPageState extends State<AddAccountPage> {
                 decoration: InputDecoration(
                   labelText: zulipLocalizations.loginServerUrlInputLabel,
                   errorText: errorText,
-                  helperText: kLayoutPinningHelperText,
+                  helper: GestureDetector(
+                    onTap: () {
+                      launchUrlWithoutRealm(context, AddAccountPage.serverUrlHelpUrl);
+                    },
+                    child: Text(
+                      // Restate Material default
+                      // (`_InputDecoratorDefaultsM3.helperText` upstream)…
+                      style: Theme.of(context).textTheme.bodySmall!
+                        // …but use blue because this is a link.
+                        .apply(color: const HSLColor.fromAHSL(1, 200, 1, 0.4).toColor()),
+                      zulipLocalizations.serverUrlDocLinkLabel)),
                   hintText: 'your-org.zulipchat.com')),
               const SizedBox(height: 8),
               ElevatedButton(
--- a/test/widgets/login_test.dart
+++ b/test/widgets/login_test.dart
@@ -1,8 +1,11 @@
 import 'package:checks/checks.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:http/http.dart' as http;
+import 'package:url_launcher/url_launcher.dart';
+
 import 'package:zulip/api/model/web_auth.dart';
 import 'package:zulip/api/route/account.dart';
 import 'package:zulip/api/route/realm.dart';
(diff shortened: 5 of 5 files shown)


abelaba:
Hey, so on thing I would say is that your commit message should have a description. As stated here

abelaba:
You must also add new tests for your code.
https://github.com/zulip/zulip-flutter?tab=readme-ov-file#writing-tests

lib/widgets/login.dart
```diff
@@ -223,7 +225,13 @@ class _AddAccountPageState extends State<AddAccountPage> {
                 decoration: InputDecoration(
                   labelText: zulipLocalizations.loginServerUrlInputLabel,
                   errorText: errorText,
-                  helperText: kLayoutPinningHelperText,
+                  helper: InkWell(
+                    onTap: () {
+                      _launchUrl(context);
+                    },
+                    child: Text(
```
chrisbobbe, line 232: Hmm, it looks like our choice to use `helper` instead of `helperText` means that the default Material-defined text style is getting thrown out. When I put "What's this?" in `helperText` and I remove `helper`, the style changes noticeably. The smaller text size, in particular, looks nicer to me.

Reading code, here's the default Material-defined text style that gets applied when the caller uses `helperText`. It's `_InputDecoratorDefaultsM3.helperStyle` in packages/flutter/lib/src/material/input_decorator.dart in the Flutter tree:

```dart
  @override
  TextStyle? get helperStyle => MaterialStateTextStyle.resolveWith((Set<MaterialState> states) {
    final TextStyle textStyle = _textTheme.bodySmall ?? const TextStyle();
    if (states.contains(MaterialState.disabled)) {
      return textStyle.copyWith(color: _colors.onSurface.withOpacity(0.38));
    }
    return textStyle.copyWith(color: _colors.onSurfaceVariant);
  });
```

It would be annoying to have to copy all of that, and since `_InputDecoratorDefaultsM3` is private, I don't think we'll be able to reference it directly. But maybe we can simplify a bit:

- This input doesn't need a disabled state, so we can drop that part.
- We want to set a bluish color because this is a link, so we can forget about `onSurfaceVariant`.

Which conveniently just leaves one small thing, the text size. So for example:

```dart
                    child: Text(
                      // Restate Material default
                      // (`_InputDecoratorDefaultsM3.helperText` upstream)…
                      style: Theme.of(context).textTheme.bodySmall!
                        // …but use blue because this is a link.
                        .apply(color: const HSLColor.fromAHSL(1, 200, 1, 0.4).toColor()),
```

(For the `HSLColor.fromAHSL` part, see a previous comment in my review.)

lib/widgets/login.dart
```diff
@@ -223,7 +225,13 @@ class _AddAccountPageState extends State<AddAccountPage> {
                 decoration: InputDecoration(
                   labelText: zulipLocalizations.loginServerUrlInputLabel,
                   errorText: errorText,
-                  helperText: kLayoutPinningHelperText,
+                  helper: InkWell(
```
chrisbobbe, line 228: Hmm, awkward that the ink area is really rigidly rectangular:

image

For comparison, here's something Google does on a similar text input. This is on web, and I'm hovering on the link:

image

Notice the rounded corners, and that the start of the text is still aligned with the left edge of the text field. I think that alignment would unfortunately be more trouble than it's worth for us to do here.

Probably what I would do here is just not bother with a touch-response effect, and just use a plain `GestureDetector` instead of an `InkWell`. Like we do for links in Zulip message content.

chrisbobbe (changes requested):
Thanks @VatsalBhesaniya, I'm looking forward to this! This will help people get what they need to start using the app.

Small comments below. Also, this will need tests. Here are the things I would test:

- Basic happy case: When the text is tapped, the URL is opened. For this, see our uses of `testBinding.takeLaunchUrlCalls()`.
- Error case: When the URL fails to be opened, we show an error dialog. For this, see our uses of `testBinding.launchUrlResult` and `checkErrorDialog`.

Also, please mark the commit with a `Fixes` line. I see you've already done this in the GitHub PR description. 🙂

assets/l10n/app_en.arb
```diff
@@ -289,6 +289,10 @@
   "@loginServerUrlInputLabel": {
     "description": "Input label in login page for Zulip server URL entry."
   },
+  "serverURLDocLinkLabel": "(What's this?)",
```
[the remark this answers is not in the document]
VatsalBhesaniya: Makes sense

lib/widgets/login.dart
VatsalBhesaniya: Thank you for the insightful analysis of the `_InputDecoratorDefaultsM3.helperStyle`. I appreciate you walking me through the breakdown of the style properties.

I've updated the code as you suggested.

lib/widgets/login.dart
VatsalBhesaniya: Thanks for pointing this out! I agree the current ink rectangle feels a bit rigid.
Following your suggestion, I've replaced `InkWell` with a plain `GestureDetector`.

I have updated the screen recordings in the PR description.

VatsalBhesaniya:
> Hey, so one thing I would say is that your commit message should have a description. As stated here https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-description

@abelaba Thank you for your feedback. 🙂

VatsalBhesaniya:
> Thanks @VatsalBhesaniya, I'm looking forward to this! This will help people get what they need to start using the app.
> 
> Small comments below. Also, this will need tests. Here are the things I would test:
> 
> * Basic happy case: When the text is tapped, the URL is opened. For this, see our uses of `testBinding.takeLaunchUrlCalls()`.
> * Error case: When the URL fails to be opened, we show an error dialog. For this, see our uses of `testBinding.launchUrlResult` and `checkErrorDialog`.
> 
> Also, please mark the commit with a `Fixes` line. I see you've already done this in the GitHub PR description. 🙂

Thanks for the positive feedback!
I appreciate your suggestions on tests. I've added tests covering the success and the error cases.

I've also updated the commit message to include the `Fixes` line.

Please let me know if you have any further feedback or require any additional changes to the tests.

chrisbobbe:
Hi! One small comment I meant to leave last time, and I think Greg intends to do the next full review. 🙂

There's still a `TODO(#109)` in the code that should be removed in the commit that's marked as fixing #109. I've proposed an automated way to catch things like this:

- #599

test/widgets/login_test.dart
```diff
@@ -142,4 +144,46 @@ void main() {
...
+    Future<void> prepareAddAccountPage(WidgetTester tester) async {
+      await tester.pumpWidget(const MaterialApp(
+        localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+        supportedLocales: ZulipLocalizations.supportedLocales,
+        home: AddAccountPage(),
+      ));
+    }
+
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+    const url = 'https://zulip.com/help/logging-in#find-the-zulip-log-in-url';
```
gnprice, line 158: Let's give this constant an appropriate name in the code under test, export it, and then the test can just refer to that. The exact URL seems like the sort of thing that may change.

So e.g.: `AddAccountPage.serverUrlHelpUrl`.

test/widgets/login_test.dart
```diff
@@ -142,4 +144,46 @@ void main() {
...
+
+    Future<Finder> findHelperText(WidgetTester tester) async {
+      return find.byWidgetPredicate((widget) =>
+          widget is GestureDetector &&
+          widget.child is Text &&
+          (widget.child as Text).data == zulipLocalizations.serverURLDocLinkLabel);
```
gnprice, line 164: How about just `find.text(zulipLocalizations.serverURLDocLinkLabel)`?

test/widgets/login_test.dart
```diff
@@ -142,4 +144,46 @@ void main() {
...
+    }
+
+    testWidgets('launches URL when helper text is tapped', (WidgetTester tester) async {
+      await prepareAddAccountPage(tester);
+      final helper = await findHelperText(tester);
+      await tester.tap(helper);
+
+      check(testBinding.takeLaunchUrlCalls())
+          .single
+          .equals((url: Uri.parse(url), mode: LaunchMode.platformDefault));
```
gnprice, line 174: nit: two-space indents; and `.single` is so simple that it's helpful to make the code more compact by combining it with another line:
```suggestion
      check(testBinding.takeLaunchUrlCalls())
        .single.equals((url: Uri.parse(url), mode: LaunchMode.platformDefault));
```

test/widgets/login_test.dart
```diff
@@ -142,4 +144,46 @@ void main() {
...
+    const url = 'https://zulip.com/help/logging-in#find-the-zulip-log-in-url';
+
+    Future<Finder> findHelperText(WidgetTester tester) async {
+      return find.byWidgetPredicate((widget) =>
+          widget is GestureDetector &&
+          widget.child is Text &&
+          (widget.child as Text).data == zulipLocalizations.serverURLDocLinkLabel);
+    }
+
+    testWidgets('launches URL when helper text is tapped', (WidgetTester tester) async {
```
gnprice, line 167: nit: follow style of existing test cases, leaving off "WidgetTester " on test-body callback:
```suggestion
    testWidgets('launches URL when helper text is tapped', (tester) async {
```

assets/l10n/app_en.arb
```diff
@@ -289,6 +289,21 @@
   "@loginServerUrlInputLabel": {
     "description": "Input label in login page for Zulip server URL entry."
   },
+  "serverURLDocLinkLabel": "What's this?",
```
gnprice, line 292: nit: In camel case, capitalize acronyms like "URL" the same way as words:
```suggestion
  "serverUrlDocLinkLabel": "What's this?",
```

See the Dart style guide for discussion: https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

gnprice (commented):
Thanks @VatsalBhesaniya for adding this (and for the upstream PR https://github.com/flutter/flutter/pull/145157 that added this `helper` field!) See comments below.

test/widgets/login_test.dart
VatsalBhesaniya: Agreed! Added a new static const variable `serverUrlHelpUrl` in `AddAccountPage`.

test/widgets/login_test.dart
VatsalBhesaniya: Sure! It will be more concise.

lib/widgets/login.dart
```diff
@@ -235,6 +244,31 @@ class _AddAccountPageState extends State<AddAccountPage> {
   }
 }
 
+void _launchUrl(BuildContext context, ZulipLocalizations zulipLocalizations) async {
```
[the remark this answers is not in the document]
VatsalBhesaniya: Thank you very much for your guidance! I was thinking about refactoring it, but I wasn't sure of the best approach with the realm URL differences. 
I've implemented the separation you suggested in a new file, `launch_url.dart`.

I have created a new commit for these changes as they involve refactoring and impact files outside the original PR scope (specifically `content.dart`). 

I have updated the PR description with the following:
- A new screen recording demonstrating the updated iOS launch mode logic.
- Screenshots of the error case (which I should have added earlier).

lib/widgets/launch_url.dart
```diff
@@ -0,0 +1,62 @@
...
+    return;
+  }
+
+  final internalNarrow = parseInternalLink(url, store);
+  if (internalNarrow != null) {
+    Navigator.push(context, MessageListPage.buildRoute(context: context, narrow: internalNarrow));
+    return;
+  }
+
+  launchUrlWithoutRealm(context, url.toString());
```
gnprice, line 61: At this point we've already parsed the URL. So instead of turning it back to a string and then parsing it again, it'd be better to pass the parsed `Uri` object through.

lib/widgets/launch_url.dart
```diff
@@ -0,0 +1,62 @@
...
+    errorMessage = e.message;
+  }
+  if (!launched) {
+    if (!context.mounted) return;
+    await _showError(context, errorMessage, urlString);
+  }
+}
+
+/// Launches a URL considering a realm base URL (if available).
+void launchUrlWithRealm(BuildContext context, String urlString) async {
```
gnprice, line 47: What does "if available" mean?

Then the next question someone might ask themself if they're trying to use this function and looking at this documentation is: What happens if it isn't available?

gnprice (commented):
Thanks @VatsalBhesaniya for the revision!

In this version, the first commit introduces a duplicate `_launchUrl` function which is missing the `mode` logic, much like in the previous revision as discussed at https://github.com/zulip/zulip-flutter/pull/595#discussion_r1543818447 . Then the second commit deduplicates the two functions into one combined place.

As part of writing [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html), we want to avoid that intermediate step where we're adding a `_launchUrl` function that has the wrong behavior. So instead a good commit sequence would be:
* Move the existing `_launchUrl` to the new file, renaming it, perhaps splitting into pieces, but maintaining the exact same behavior. This will be an [NFC commit](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc).
* Add i18n for that function.
* Add the UI that resolves #109, calling that function.

After that's done it'll again be possible to do a fine-grained review. In the meantime, here's a couple of comments from a high-level look.

lib/widgets/launch_url.dart
VatsalBhesaniya: Thanks for pointing this out!

Thanks for pointing that out! I initially misunderstood the behavior of `launchUrlWithRealm`. I thought it would fall back to launching without a realm base URL if unavailable. Therefore, I added "if available" to the doc comment.

I've removed the "if available" from the comment to clarify this behavior. Please let me know if you have any suggestions to improve this doc comment.

VatsalBhesaniya:
> Thanks @VatsalBhesaniya for the revision!
> 
> In this version, the first commit introduces a duplicate `_launchUrl` function which is missing the `mode` logic, much like in the previous revision as discussed at [#595 (comment)](https://github.com/zulip/zulip-flutter/pull/595#discussion_r1543818447) . Then the second commit deduplicates the two functions into one combined place.
> 
> As part of writing [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html), we want to avoid that intermediate step where we're adding a `_launchUrl` function that has the wrong behavior. So instead a good commit sequence would be:
> 
> * Move the existing `_launchUrl` to the new file, renaming it, perhaps splitting into pieces, but maintaining the exact same behavior. This will be an [NFC commit](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc).
> * Add i18n for that function.
> * Add the UI that resolves [login: Link to doc to help users understand what a "server URL" is and how to find theirs #109](https://github.com/zulip/zulip-flutter/issues/109), calling that function.
> 
> After that's done it'll again be possible to do a fine-grained review. In the meantime, here's a couple of comments from a high-level look.

Thank you for your detailed feedback and guidance!

I have followed your suggested approach and updated the commits accordingly. Let me know if you have any further suggestions.

test/widgets/login_test.dart
```diff
@@ -218,4 +221,41 @@ void main() {
...
+    testWidgets('shows error dialog when URL fails to open', (tester) async {
+      await prepareAddAccountPage(tester);
+      testBinding.launchUrlResult = false;
+      final helper = await findHelperText(tester);
+      await tester.tap(helper);
+      await tester.pump();
+
+      checkErrorDialog(tester,
+          expectedTitle: zulipLocalizations.errorUnableToOpenLinkTitle,
+          expectedMessage: zulipLocalizations.errorLinkCouldNotBeOpened(AddAccountPage.serverUrlHelpUrl));
```
chrisbobbe, line 258: To exercise just a bit more of the UI, we can wrap this `checkErrorDialog(…)` in `await tester.tap(find.byWidget(…))`, so that it simulates tapping the dialog's OK button.

lib/widgets/content.dart
```diff
@@ -1,22 +1,17 @@
-import 'package:flutter/foundation.dart';
```
chrisbobbe, line 1:
```
launch_url[nfc]: Move launchUrl logic to a new file
```

Commit message nit: add a space between "launch_url" and "[nfc]"

assets/l10n/app_en.arb
```diff
@@ -308,6 +308,17 @@
   "@loginServerUrlInputLabel": {
     "description": "Input label in login page for Zulip server URL entry."
   },
+  "errorUnableToOpenLinkTitle": "Unable to open link",
```
chrisbobbe, line 315:
```
launch_url: Add i18n for launchUrl error messages

- Utilize i18n for consistent error messaging.
```

Commit message nit:

I think "consistency" isn't exactly the right explanation for this change. The goal of internationalization is to make the app more usable in other languages, with help from our kind volunteer translators.

I would just drop the body and leave the summary line, in this case. If you [view the Git history](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/git.md#2-the-secret-to-using-git-log--p) you can see many examples of commit messages that don't have a body, because the summary line is sufficient.

assets/l10n/app_en.arb
```diff
@@ -308,6 +308,10 @@
   "@loginServerUrlInputLabel": {
     "description": "Input label in login page for Zulip server URL entry."
   },
+  "serverUrlDocLinkLabel": "What's this?",
```
chrisbobbe, line 311:
```
login: Link to doc for what "server URL" is and how to find it

- Add informative helper text below the "server URL" field in the login screen.
- When tapped, the helper text opens Zulip documentation explaining server URLs and how to find them.
- Improves user experience during login by providing clear guidance.
- Add i18n for "What's this?" helper text.
- Fixes: #109
```

Commit message nit:

Separate the "Fixes" line from the bulleted list. In fact, I think the "Fixes" line adequately explains what's being done and why, so I would delete the rest of the body:

```
login: Link to doc for what "server URL" is and how to find it

Fixes: #109
```

chrisbobbe (commented):
Thanks @VatsalBhesaniya! Small comments below.

Also, please wrap the body text of each commit message; some lines have gotten too long: https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#formatting-guidelines

chrisbobbe:
Hi @VatsalBhesaniya, it looks like you've updated the PR branch. Is it ready for another review? It's helpful if you post a comment saying it's ready for review, for explicitness; that way I can be sure. 🙂

VatsalBhesaniya:
> Hi @VatsalBhesaniya, it looks like you've updated the PR branch. Is it ready for another review? When you push updates and it's ready for another review, it's helpful if you post a comment saying so, for explicitness; that way reviewers don't have to guess. 🙂

Hi @chrisbobbe, Yes, the PR branch is ready for another review! I apologize for the confusion. I made updates locally, but for some reason, they weren't reflecting on GitHub. This hasn't happened to me before, so I hesitated to request another review until I was certain everything was pushed correctly.

test/widgets/login_test.dart
```diff
@@ -61,7 +64,48 @@ void main() {
...
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+
+    Future<Finder> findHelperText(WidgetTester tester) async {
+      return find.text(zulipLocalizations.serverUrlDocLinkLabel);
+    }
+
+    testWidgets('launches URL when helper text is tapped', (tester) async {
+      await prepareAddAccountPage(tester);
+      final helper = await findHelperText(tester);
+      await tester.tap(helper);
```
chrisbobbe, line 85:
```suggestion
      await tester.tap(await findHelperText(tester));
```

nit: could save a line by inlining this expression (here and in the other test)

test/widgets/login_test.dart
```diff
@@ -61,7 +64,48 @@ void main() {
...
+      await tester.pump();
+
+      await tester.tap(find.byWidget(checkErrorDialog(
+        tester,
+        expectedTitle: zulipLocalizations.errorUnableToOpenLinkTitle,
+        expectedMessage: zulipLocalizations.errorUnableToOpenLinkMessage(
+          AddAccountPage.serverUrlHelpUrl.toString(),
+        ),
+      )));
```
chrisbobbe, line 106:
```suggestion
      await tester.tap(find.byWidget(checkErrorDialog(tester,
        expectedTitle: zulipLocalizations.errorUnableToOpenLinkTitle,
        expectedMessage: zulipLocalizations.errorUnableToOpenLinkMessage(
          AddAccountPage.serverUrlHelpUrl.toString(),
        ))));
```

nit: could be denser here, using fewer lines

chrisbobbe (commented):
Thanks @VatsalBhesaniya! This is very close. Small comments below.

chrisbobbe:
Hi @VatsalBhesaniya, are you planning to return to this?

lib/widgets/login.dart
VatsalBhesaniya: Added comments.

VatsalBhesaniya:
> Hi @VatsalBhesaniya, are you planning to return to this?

Sorry for the delay. I have made all the changes requested.

chrisbobbe:
Great, thank you! This revision LGTM; I'll mark this for Greg's review now.

alya:
I don't understand the "Unable to open link" screenshots in the PR description. Is that something users will see?

assets/l10n/app_en.arb
```diff
@@ -308,6 +308,21 @@
...
+  },
+  "errorUnableToOpenLinkTitle": "Unable to open link",
+  "@errorUnableToOpenLinkTitle": {
+    "description": "Error title when a link fails to open."
+  },
+  "errorUnableToOpenLinkMessage": "Link could not be opened: {url}",
+  "@errorUnableToOpenLinkMessage": {
+    "description": "Error message when a link fails to open.",
+    "placeholders": {
+      "url": {"type": "String", "example": "http://example.com/"}
```
gnprice, line 323: nit: even in examples, let's stick to HTTPS everywhere:
```suggestion
      "url": {"type": "String", "example": "https://example.com/"}
```

test/widgets/login_test.dart
```diff
@@ -61,7 +64,44 @@ void main() {
     expectErrorFromText('email@example.com', ServerUrlValidationError.noUseEmail);
   });
 
-  // TODO test AddAccountPage
```
gnprice, line 64: There's lots more to test on this page :slightly_smiling_face: — so this TODO is still relevant and let's keep it.

test/widgets/login_test.dart
```diff
@@ -61,7 +64,44 @@ void main() {
...
+  group('AddAccountPage server URL helper text', () {
+    Future<void> prepareAddAccountPage(WidgetTester tester) async {
+      await tester.pumpWidget(const MaterialApp(
+        localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+        supportedLocales: ZulipLocalizations.supportedLocales,
+        home: AddAccountPage(),
+      ));
+    }
+
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
```
gnprice, line 76: This `GlobalLocalizations.zulipLocalizations` is application state that can change. So reading it should happen within individual test cases (including helpers they call), i.e. individual `testWidgets` callbacks, not outside them.

test/widgets/login_test.dart
```diff
@@ -61,7 +64,44 @@ void main() {
...
+
+      check(testBinding.takeLaunchUrlCalls()).single.equals((
+        url: AddAccountPage.serverUrlHelpUrl,
+        mode: LaunchMode.platformDefault,
+      ));
+    });
+
+    testWidgets('shows error dialog when URL fails to open', (tester) async {
+      await prepareAddAccountPage(tester);
+      testBinding.launchUrlResult = false;
```
gnprice, line 94: When a test manipulates `testBinding`, it needs to call `addTearDown(testBinding.reset)` somewhere so that the state gets reset before the next test.

(In the "error dialog if invalid link" in widgets/content_test.dart which this one is presumably based on, that gets taken care of by the `prepare` helper, because the latter does other things with `testBinding` that also need a reset.)

You can reproduce the problem with the command:
```
$ flutter test test/widgets/login_test.dart --test-randomize-ordering-seed=random
```
At your current revision, that often fails because `launchUrlResult` remains set to false when the other new test tries to run. After you fix the problem with a reset, that command should pass every time.

gnprice (commented):
Thanks @VatsalBhesaniya, and thanks @chrisbobbe for the previous revisions! Code comments below.

Echoing Alya's question: what are the circumstances where you'd expect a user to see that error dialog?

In addition to the two screen recordings, which are helpful for the interaction of opening the link: I think a very helpful thing to include as a screenshot would be just this add-account screen itself. That'd help for being able to look at the new helper text and gauge how it looks. (A video is less good for that because it's hard to get it in full resolution.)

PIG208:
Hi @VatsalBhesaniya! Thanks for the previous revisions. Any plan to continue working on this?

gnprice:
Closing this PR to get it out of our queue. Thanks @VatsalBhesaniya for your work on it so far.

The existing PR branch will remain available in Git, for the next person to refer to when picking up this issue #109, and it has the "completion candidate" label which may make easier to find for that purpose. I'll also make a note on the issue.

Issue #465, opened 2023-12-27
Show "Connecting…" banner (or equivalent) when server data is stale


gnprice:
When we lose our real-time stream of events from the server, either because of network failures or because the server has expired the event queue and we need to request a new one:
* #184
* #185

we continue to show the latest server data we had, but it may now be stale. We should communicate to the user that that's what's happening.

In the existing zulip-mobile app, we do this with a banner that reads "Connecting…" and appears just under the app bar.

In this app, we could do the same thing, or we could find a different design. One downside of the "Connecting…" banner is that it occupies some vertical space, which means the rest of the content needs to move down to make room (or else let its top portion be obscured by the banner). Nevertheless it might be the best available option.

Even if we use the same design concept, we'll want to pin down the details: the text style, the layout, the banner's background, the specifics of how it animates in and out.

(In zulip-mobile the details are in `src/common/LoadingBanner.js`; there's no animation. But the conceptually similar "No Internet connection" banner, found in `src/boot/OfflineNoticeProvider.js`, does have an animation… on iOS only, because React Native's implementation breaks on Android. Here on Flutter, we should have no such trouble implementing whatever animation we choose.)

gnprice:
Bumping this in priority because even after #185 and #184, we're still sometimes showing the symptom of lacking an event queue, even when the device has a fine internet connection:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Only.20.22all.20messages.22.20updating/near/1754151
(I guess Alex didn't explicitly say the last part, but I definitely had a fine connection when I observed similar symptoms the other day.)

And as Alex said there:
> silently not having updates means I cannot trust the app. Having some user feedback seems very important, here.

chrisbobbe:
Maybe something like a toast?

Or a thin, linear indeterminate progress indicator at the bottom of the app bar (`LinearProgressIndicator`), like we do when logging you in? (It disappears quickly in this video, but it's there.)

https://github.com/zulip/zulip-flutter/assets/22248748/0b63a678-37ab-49c3-8365-ca80579cb1ab

gnprice:
> a thin, linear indeterminate progress indicator at the bottom of the app bar (`LinearProgressIndicator`), like we do when logging you in?

Yeah, that sounds like the easiest thing to do that remains pretty clean-looking. We can always revise it to another design later.

[...]

PR #584, opened 2024-03-24
added connecting snackbar and snackbar.dart file

satyam372:
Added 'snackbar.dart' file to show the snackbar. I have created this file inside widget folder and then imported it in app.dart and message_list.dart file . 

working : I have used "connectivity_plus: ^5.0.2" package to implement this.   This package allows developers to see the connectivity of device . This helps us to determine  whether our device has active internet connection or not .

Video link:-
https://github.com/zulip/zulip-flutter/assets/112092492/911a49a4-568c-4b29-82bc-1c8d0bf5872b

PR #598, opened 2024-03-28
snackbar: Handle connectivity changes and show appropriate SnackBar#465

satyam372:
PR for issue #465 
This commit introduces a SnackBarPage widget that manages changes in connectivity status and displays corresponding SnackBars. The SnackBarPage widget listens to connectivity changes using the Connectivity package and updates the UI to show a SnackBar indicating either "No Internet Connection" or "Connecting" based on the current connectivity status and the value of the isStale property.

The initState method initializes the connectivity status when the widget is first created, while the didChangeDependencies method subscribes to connectivity changes to update the UI accordingly. To prevent memory leaks, the dispose method cancels the subscription when the widget is disposed.

The showSnackBar method is responsible for displaying the appropriate SnackBar based on the current connectivity status and the value of the isStale property. It utilizes ScaffoldMessenger to manage SnackBar display and dismissal.

### DIFF
--- a/lib/widgets/snackbar.dart
+++ b/lib/widgets/snackbar.dart
@@ -1,0 +1,224 @@
+//
+// import 'dart:async';
+// import 'package:flutter/material.dart';
+// import 'package:connectivity_plus/connectivity_plus.dart';
+//
+// class SnackBarPage extends StatefulWidget {
+//   const SnackBarPage({super.key});
+//
+//   @override
+//   SnackBarPageState createState() => SnackBarPageState();
+// }
+//
+// class SnackBarPageState extends State<SnackBarPage> {
+//   late ConnectivityResult _connectivityResult; // ignore: unused_field
+//   late StreamSubscription<ConnectivityResult> _connectivitySubscription;
+//
+//   @override
+//   void initState() {
+//     super.initState();
+//     // Initialize connectivity status
+//     _initConnectivity();
+//   }
+//
+//   @override
+//   void didChangeDependencies() {
+//     super.didChangeDependencies();
+//     // Subscribe to connectivity changes
+//     _connectivitySubscription = Connectivity().onConnectivityChanged.listen((ConnectivityResult result) {
+//       setState(() {
+//         _connectivityResult = result;
+//       });
+//       // Show or dismiss snackbar based on connectivity changes
+//       showSnackBar(result);
+//     });
+//   }
+//
+//   @override
+//   void dispose() {
+//     _connectivitySubscription.cancel(); // Cancel subscription to avoid memory leaks
+//     super.dispose();
+//   }
+//
+//   Future<void> _initConnectivity() async {
+//     final ConnectivityResult connectivityResult = await Connectivity().checkConnectivity();
+//     setState(() {
+//       _connectivityResult = connectivityResult;
+//     });
+//     //  initial snackbar based on connectivity status
+//     showSnackBar(connectivityResult);
+//   }
+//
+//   void showSnackBar(ConnectivityResult connectivityResult) {
+//     final bool isConnected = connectivityResult != ConnectivityResult.none;
+//     ScaffoldMessenger.of(context).hideCurrentSnackBar();
+//
+//     if (isConnected) {
+//       ScaffoldMessenger.of(context).showSnackBar(
+//         const SnackBar(
+//           content: Row(
+//             children: [
+//               Icon(
+//                 Icons.sync,
+//                 color: Colors.white,
+//               ),
+//               SizedBox(width: 8),
+//               Text(
+//                 'Connecting',
+//                 style: TextStyle(color: Colors.white),
+//               ),
+//             ],
+//           ),
+//           duration:  Duration(seconds: 3),
+//         ),
+//       );
+//     } else {
+//       ScaffoldMessenger.of(context).showSnackBar(
+//         const SnackBar(
+//           content:  Row(
+//             children: [
+//               Icon(
+//                 Icons.error_outline,
+//                 color: Colors.white,
+//               ),
+//               SizedBox(width: 2),
+//               Text(
+//                 'No Internet Connection',
+//                 style:  TextStyle(color: Colors.white),
+//               ),
+//             ],
+//           ),
+//           duration:  Duration(seconds: 365),
+//         ),
+//       );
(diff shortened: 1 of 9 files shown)


gnprice:
When a PR is meant to fix an issue, please refer to the issue in both the commit and the PR description. This is mentioned [in the Zulip Git style guide](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-description), and in more detail here:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#mention-fixed-issue

Similarly, when there are previous closely related PRs, mention those too. For example (and I'm mentioning these now in order to get the cross-references): previous iterations of this PR were #586, #584, #583.

lib/widgets/snackbar.dart
```diff
@@ -0,0 +1,224 @@
...
+    super.initState();
+    // Initialize connectivity status
+    _initConnectivity();
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    // Subscribe to connectivity changes
+    _connectivitySubscription = Connectivity().onConnectivityChanged.listen((ConnectivityResult result) {
```
gnprice, line 140: I believe this PR is meant to be for #465. The title of that issue says:
> Show "Connecting…" banner (or equivalent) when server data is stale

But I don't see any logic here that relates to whether the server data is stale. Instead, it seems to be all about whether the device believes it has an internet connection.

Knowing whether the device thinks it has an internet connection can be helpful for choosing the exact warning message to show. But it doesn't in itself do anything to resolve the original issue: in the main situation we're interested in, there is a perfectly good internet connection.

gnprice (commented):
@satyam372 It looks like this PR doesn't yet have any logic to address the issue #465, which is the issue where we want a banner or snack bar resembling this one. See in particular my last comment below.

I'm marking this PR as draft; feel free to mark as ready again when it has logic that addresses the issue.

The other comments will also be relevant for a revised version of this PR (and for other PRs in general).

satyam372:
New Approach :-

1)The function registers an event queue with the Zulip server.
2)It sets a flag to indicate registration is in progress.
3)Sends a POST request to the server.
4)If successful (status code 200), it extracts the queueId and lastEventId, hides the "Connecting…" banner, and shows a success message.
5)If not successful, it prints an error message and shows a  connecting message.

satyam372:
`   Future<void> registerEventQueue() async {
    setState(() {
      _isRegisteringEventQueue = true;
    });
final String apiUrl = 'https://yourZulipDomain.zulipchat.com/api/v1/events';     



    try {
      final response = await http.post(
        Uri.parse(apiUrl),
        headers: {
          'Content-Type': 'application/x-www-form-urlencoded',
          //  required authentication headers 
        },
      );

      if (response.statusCode == 200) {
        final responseData = json.decode(response.body);
        final queueId = responseData['queue_id'];
        final lastEventId = responseData['last_event_id'];

        //  queueId and lastEventId to start listening for events
        // Hide the "Connecting…" banner and update the UI with refreshed data
        setState(() {
          _isRegisteringEventQueue = false;
        });

        ScaffoldMessenger.of(context).showSnackBar(
          const SnackBar(
            content: Text(
              'Event queue registered successfully',
              style: TextStyle(color: Colors.white),
            ),
            duration: Duration(seconds: 3),
          ),
        );
      } else {
        print('Failed to register event queue: ${response.statusCode}');
        showSnackBar(ConnectivityResult.none); //  "No Internet Connection" message
      }
    } catch (e) {
      print('Error registering event queue: $e');
      showSnackBar(ConnectivityResult.none); // "No Internet Connection" message
    }
  }`


we can use "https://yourZulipDomain.zulipchat.com/api/v1/events" this endpoint to get the data about the the event queue. which i read about this url in zulip docs.          

@gnprice whar are your views on the new approach .

lib/widgets/snackbar.dart
satyam372: Thankyou for your feedback

satyam372: Approach :- 1) Checking Internet Connection:  connectivity_plus package to determine if the device has 
                        an active internet connection. This ensures that your app can communicate with the server.

                    2) State Management: state management using libraries like Provider or Riverpod. Store 
                        timestamps or flags with your data to track when it was last updated.

                    3) getting real-time events from GET https://yourZulipDomain.zulipchat.com/api/v1/events get
                          updates about changes in your data. 

                    4) Timestamps: Attaching  time stamp to  data when  fetching it from the server and compare . 

@gnprice this was my approach to the issue #465 . Please correct me if I am going wrong.

satyam372: Sample Code for my logic 


`    void _refreshData() async {
    try {
      var response = await http.get(Uri.parse('https://jsonplaceholder.typicode.com/posts'));
// Replace the url with https://yourZulipDomain.zulipchat.com/api/v1/events

      if (response.statusCode == 200) {
        setState(() {
          _isDataStale = false;
        });
        _showSnackbar('Data refreshed successfully');
      } else {
        setState(() {
          _isDataStale = true;
        });
        _showSnackbar('Failed to refresh data. Status code: ${response.statusCode}');
      }
    } catch (error) {
      setState(() {
        _isDataStale = true;
      });
      _showSnackbar('An error occurred: $error');
    }
  }

  void _showSnackbar(String message) {
    ScaffoldMessenger.of(context).showSnackBar(
      SnackBar(
        content: Text(message),
        duration: Duration(seconds: 3),
      ),
    );
  }
}`

gnprice: For cross-reference: we discussed this in chat at https://chat.zulip.org/#narrow/stream/48-mobile/topic/.23F465.20Show.20.22Connecting.E2.80.A6.22.20banner.20.28or.20equivalent.29.20when.20serv.2E.2E.2E/near/1770549 .

chrisbobbe:
Please disable your autoformatter and undo the unrelated formatting changes, so that we can do a deeper code review without distractions. The project's approach to formatting is covered in the README: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#code-formatting

chrisbobbe:
Also, please organize your commits: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request

PR #281, opened 2023-08-21
login: Implement browser login

rajveermalviya:
Fixes #36

Implement authentication using browser auth methods, currently only tested and implemented on Android, but I doubt getting it to work on iOS would be an extensive - in theory should only require [enabling universal link support in Flutter](https://docs.flutter.dev/cookbook/navigation/set-up-universal-links).

Testing this locally will require the zulip-mobile app uninstalled, because the redirect url from the browser 'zulip://login' will also be handled by zulip-mobile and if that happens the behavior will UB.

Also only some `external_methods` are tested and only those are enabled currently (google, github, gitlab).

| Preview |
| - |
| [output.webm](https://github.com/zulip/zulip-flutter/assets/36819268/a6011869-265d-4a4c-936d-6fdb6a6c75a9) |

PR #600, opened 2024-03-28
login: Support web-based auth methods

chrisbobbe:
https://github.com/zulip/zulip-flutter/assets/22248748/a97f0696-404b-4063-8958-6fd5b225f636

Fixes: #36

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -25,6 +25,13 @@
                 <action android:name="android.intent.action.MAIN"/>
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
+            <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="zulip" android:host="login" />
+            </intent-filter>
         </activity>
         <!-- Don't delete the meta-data below.
              This is used by the Flutter tool to generate GeneratedPluginRegistrant.java -->
--- a/lib/api/model/web_auth.dart
+++ b/lib/api/model/web_auth.dart
@@ -1,0 +1,87 @@
+import 'dart:math';
+
+import 'package:convert/convert.dart';
+import 'package:flutter/foundation.dart';
+
+/// The authentication information contained in the zulip:// redirect URL.
+class WebAuthPayload {
+  final Uri realm;
+  final String email;
+  final int? userId; // TODO(server-5) new in FL 108
+  final String otpEncryptedApiKey;
+
+  WebAuthPayload._({
+    required this.realm,
+    required this.email,
+    required this.userId,
+    required this.otpEncryptedApiKey,
+  });
+
+  factory WebAuthPayload.parse(Uri url) {
+    if (
+      url case Uri(
+        scheme: 'zulip',
+        host: 'login',
+        queryParameters: {
+          'realm': String realmStr,
+          'email': String email,
+          // 'user_id' handled below
+          'otp_encrypted_api_key': String otpEncryptedApiKey,
+        },
+      )
+    ) {
+      final Uri? realm = Uri.tryParse(realmStr);
+      if (realm == null) throw const FormatException();
+
+      // TODO(server-5) require in queryParameters (new in FL 108)
+      final userIdStr = url.queryParameters['user_id'];
+      int? userId;
+      if (userIdStr != null) {
+        userId = int.tryParse(userIdStr, radix: 10);
+        if (userId == null) throw const FormatException();
+      }
+
+      if (!RegExp(r'^[0-9a-fA-F]{64}$').hasMatch(otpEncryptedApiKey)) {
+        throw const FormatException();
+      }
+
+      return WebAuthPayload._(
+        otpEncryptedApiKey: otpEncryptedApiKey,
+        email: email,
+        userId: userId,
+        realm: realm,
+      );
+    } else {
+      // TODO(dart): simplify after https://github.com/dart-lang/language/issues/2537
+      throw const FormatException();
+    }
+  }
+
+  String decodeApiKey(String otp) {
+    final otpBytes = hex.decode(otp);
+    final otpEncryptedApiKeyBytes = hex.decode(otpEncryptedApiKey);
+    if (otpBytes.length != otpEncryptedApiKeyBytes.length) {
+      throw const FormatException();
+    }
+    return String.fromCharCodes(Iterable.generate(otpBytes.length,
+      (i) => otpBytes[i] ^ otpEncryptedApiKeyBytes[i]));
+  }
+}
+
+String generateOtp() {
+  final rand = Random.secure();
+  final Uint8List bytes = Uint8List.fromList(
+    List.generate(32, (_) => rand.nextInt(256)));
+  return hex.encode(bytes);
+}
+
+/// For tests, create an OTP-encrypted API key.
+@visibleForTesting
+String debugEncodeApiKey(String apiKey, String otp) {
+  final apiKeyBytes = apiKey.codeUnits;
+  assert(apiKeyBytes.every((byte) => byte <= 0xff));
+  final otpBytes = hex.decode(otp);
+  assert(apiKeyBytes.length == otpBytes.length);
+  return hex.encode(List.generate(otpBytes.length,
+    (i) => apiKeyBytes[i] ^ otpBytes[i]));
+}
--- a/lib/widgets/login.dart
+++ b/lib/widgets/login.dart
@@ -178,7 +178,7 @@ class _AddAccountPageState extends State<AddAccountPage> {
 
       // TODO(#36): support login methods beyond username/password
       Navigator.push(context,
-        PasswordLoginPage.buildRoute(serverSettings: serverSettings));
+        LoginPage.buildRoute(serverSettings: serverSettings));
     } finally {
       setState(() {
         _inProgress = false;
@@ -235,21 +235,21 @@ class _AddAccountPageState extends State<AddAccountPage> {
   }
 }
 
-class PasswordLoginPage extends StatefulWidget {
-  const PasswordLoginPage({super.key, required this.serverSettings});
+class LoginPage extends StatefulWidget {
+  const LoginPage({super.key, required this.serverSettings});
 
   final GetServerSettingsResult serverSettings;
 
   static Route<void> buildRoute({required GetServerSettingsResult serverSettings}) {
     return _LoginSequenceRoute(
-      page: PasswordLoginPage(serverSettings: serverSettings));
+      page: LoginPage(serverSettings: serverSettings));
   }
 
   @override
-  State<PasswordLoginPage> createState() => _PasswordLoginPageState();
+  State<LoginPage> createState() => _LoginPageState();
 }
 
-class _PasswordLoginPageState extends State<PasswordLoginPage> {
+class _LoginPageState extends State<LoginPage> {
   bool _inProgress = false;
 
   Future<void> _tryInsertAccountAndNavigate({
@@ -330,7 +330,7 @@ class _PasswordLoginPageState extends State<PasswordLoginPage> {
 class _UsernamePasswordForm extends StatefulWidget {
   const _UsernamePasswordForm({required this.loginPageState});
 
-  final _PasswordLoginPageState loginPageState;
+  final _LoginPageState loginPageState;
 
   @override
   State<_UsernamePasswordForm> createState() => _UsernamePasswordFormState();
--- a/lib/widgets/login.dart
+++ b/lib/widgets/login.dart
@@ -436,12 +436,12 @@ class _LoginPageState extends State<LoginPage> {
           : null),
       body: SafeArea(
         minimum: const EdgeInsets.symmetric(horizontal: 8),
-        top: false,
         bottom: false,
         child: Center(
           child: SingleChildScrollView(
+            padding: const EdgeInsets.only(top: 8),
             child: SafeArea(
-              minimum: const EdgeInsets.symmetric(vertical: 8),
+              minimum: const EdgeInsets.only(bottom: 8),
               // TODO also detect vertical scroll gestures that start on the
               //   left or the right of this box
               child: ConstrainedBox(
--- a/lib/widgets/login.dart
+++ b/lib/widgets/login.dart
@@ -238,12 +238,12 @@ class _AddAccountPageState extends State<AddAccountPage> {
 class LoginPage extends StatefulWidget {
   const LoginPage({super.key, required this.serverSettings});
 
-  final GetServerSettingsResult serverSettings;
-
   static Route<void> buildRoute({required GetServerSettingsResult serverSettings}) {
     return _LoginSequenceRoute(
       page: LoginPage(serverSettings: serverSettings));
   }
+
+  final GetServerSettingsResult serverSettings;
 
   @override
   State<LoginPage> createState() => _LoginPageState();
--- a/lib/widgets/login.dart
+++ b/lib/widgets/login.dart
@@ -1,16 +1,23 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
+import 'package:url_launcher/url_launcher.dart';
 
 import '../api/exception.dart';
+import '../api/model/web_auth.dart';
 import '../api/route/account.dart';
 import '../api/route/realm.dart';
 import '../api/route/users.dart';
+import '../log.dart';
+import '../model/binding.dart';
 import '../model/store.dart';
 import 'app.dart';
 import 'dialog.dart';
 import 'input.dart';
 import 'page.dart';
 import 'store.dart';
+import 'text.dart';
 
 class _LoginSequenceRoute extends MaterialWidgetRoute<void> {
   _LoginSequenceRoute({
@@ -176,7 +183,6 @@ class _AddAccountPageState extends State<AddAccountPage> {
         return;
       }
 
-      // TODO(#36): support login methods beyond username/password
       Navigator.push(context,
         LoginPage.buildRoute(serverSettings: serverSettings));
     } finally {
@@ -240,10 +246,22 @@ class LoginPage extends StatefulWidget {
 
   static Route<void> buildRoute({required GetServerSettingsResult serverSettings}) {
     return _LoginSequenceRoute(
-      page: LoginPage(serverSettings: serverSettings));
+      page: LoginPage(serverSettings: serverSettings, key: _lastBuiltKey));
   }
 
   final GetServerSettingsResult serverSettings;
+
+  /// Log in using the payload of a web-auth URL like zulip://login?…
+  static Future<void> handleWebAuthUrl(Uri url) async {
+    return _lastBuiltKey.currentState?.handleWebAuthUrl(url);
+  }
+
+  /// A key for the page from the last [buildRoute] call.
+  static final _lastBuiltKey = GlobalKey<_LoginPageState>();
+
+  /// The OTP to use, instead of an app-generated one, for testing.
+  @visibleForTesting
+  static String? debugOtpOverride;
 
   @override
   State<LoginPage> createState() => _LoginPageState();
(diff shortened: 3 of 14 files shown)


gnprice (commented):
Thanks @chrisbobbe! Looking forward to having this in.

Generally the code all looks good. Comments below, mostly small.

As you mention, we'll want some tests on `WebAuthPayload`. We have a few tests of that kind in zulip-mobile in `webAuth-test.js`.

Let's also have a screenshot of the new login screen, and a video of the end-to-end auth flow. Those will be helpful for seeing what the UX is like.

lib/api/model/web_auth.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  factory WebAuthPayload.parse(Uri url) {
+    if (
+      url case Uri(
+        scheme: 'zulip',
+        host: 'login',
+        queryParameters: {
+          'email': String(isEmpty: false) && var email,
+          'realm': String(isEmpty: false) && var realmStr,
+          'otp_encrypted_api_key': String(isEmpty: false) && var otpEncryptedApiKey,
```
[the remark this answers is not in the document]
chrisbobbe: Ah, yeah. I think it was partly from just carrying over empty-string checks in #281.

lib/api/model/web_auth.dart
```diff
@@ -0,0 +1,88 @@
...
+        }
+        userId = maybeParsed;
+      }
+
+      final Uri realm;
+      final maybeParsedRealm = Uri.tryParse(realmStr);
+      if (maybeParsedRealm == null) {
+        throw const FormatException();
+      }
+      realm = maybeParsedRealm;
```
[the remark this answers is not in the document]
chrisbobbe: Indeed. Thanks!

lib/widgets/login.dart
```diff
@@ -235,23 +240,103 @@ class _AddAccountPageState extends State<AddAccountPage> {
...
+
+  /// Log in using the payload of a web-auth URL like zulip://login?…
+  Future<void> handleWebAuthUrl(Uri url) async {
+    setState(() {
+      _inProgress = true;
+    });
+    try {
+      assert (await ZulipBinding.instance.supportsCloseForLaunchMode(_webAuthLaunchMode));
+      await ZulipBinding.instance.closeInAppWebView();
+      if ((debugOtpOverride ?? _otp) == null) {
```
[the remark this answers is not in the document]
chrisbobbe: I'm a little curious how the Flutter framework does this, but in this case I'm not sure how exactly to do it. `assert` only does anything in debug mode, right, so I'm not sure what change we'd want to make to affect things outside of debug mode.

lib/widgets/login.dart
```diff
@@ -235,23 +240,103 @@ class _AddAccountPageState extends State<AddAccountPage> {
...
+      await ZulipBinding.instance.launchUrl(url, mode: _webAuthLaunchMode);
+    } catch (e) {
+      assert(debugLog(e.toString()));
+
+      if (e is PlatformException && e.message != null && e.message!.startsWith('Error while launching')) {
+        // Ignore; I've seen this on my iPhone even when auth succeeds.
+        // Specifically, Apple web auth…which on iOS should be replaced by
+        // Apple native auth; that's #462.
+        // Possibly related:
+        //   https://github.com/flutter/flutter/issues/91660
```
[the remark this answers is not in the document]
chrisbobbe: I'm not sure. In that comment, the person says "if anything fails browser side". But when I watch what goes on in the browser, nothing stands out to me as matching that description.

lib/widgets/login.dart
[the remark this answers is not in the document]
gnprice: An example of what I have in mind is `UpdateMachine.debugEnableRegisterNotificationToken`. The key is that the `assert` callback sets a local variable that belongs to the outer function.

Then for authentic upstream examples, browse through `framework.dart` with that pattern in mind. The first one I spot from just starting to read through the `Element` implementation is `Element.debugIsDefunct`.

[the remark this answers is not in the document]
chrisbobbe: Interesting; thanks! I think I've got it:

```dart
  String? _otp;
  String? _getOtp() {
    String? result = _otp;
    assert(() {
      result ??= LoginPage.debugOtpOverride;
      return true;
    }());
    return result;
  }
```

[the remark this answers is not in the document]
gnprice: Yeah, exactly.

I think you can also simplify slightly, deduplicating the use of `_otp`, by initializing to `null`, having the assert override that, and then using `?? _otp` after that.

chrisbobbe:
Hmm, a test failure in CI:

```
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/store_test.dart: UpdateMachine.load retries registerQueue on NetworkError (failed)
  Bad state: Exceeded timeout 1:00:00.000000 while flushing timers
  package:fake_async/fake_async.dart 217:9   FakeAsync.flushTimers.<fn>
  package:fake_async/fake_async.dart 241:21  FakeAsync._fireTimersWhile
  package:fake_async/fake_async.dart 214:5   FakeAsync.flushTimers
  test/fake_async.dart 29:7                  awaitFakeAsync
  test/model/store_test.dart 176:57          main.<fn>.<fn>
```

A flake, do you think?

chrisbobbe:
Thanks for the review! Revision pushed, and I've posted a screenshot and video in the PR description.

gnprice:
> Hmm, a test failure in CI:

Hmm. Does seem unlikely to be caused by this change — so it may be a pre-existing flake.

If so, then it's a bug in either the test or the code under test, so it'd be good to fix. Would you file a bug to track that flake?

chrisbobbe:
> Would you file a bug to track that flake?

Sure: #602

lib/widgets/login.dart
```diff
@@ -235,23 +241,119 @@ class _AddAccountPageState extends State<AddAccountPage> {
   }
 }
 
-class PasswordLoginPage extends StatefulWidget {
-  const PasswordLoginPage({super.key, required this.serverSettings});
+class LoginPage extends StatefulWidget {
+  const LoginPage({super.key, required this.serverSettings});
+
+  /// A key for the page from the last [buildRoute] call.
+  static final _lastBuiltKey = GlobalKey<_LoginPageState>();
```
gnprice, line 248: Let's order all these members in the way suggested in the Flutter style guide:

I think that'll make it somewhat easier to look at this class and understand the API.

(It looks like the existing code wasn't quite in that order, and perhaps it'd be better if it were. But that matters a lot less when it's so simple; as it gains more complexity, the structure becomes more important for the reader.)

lib/widgets/login.dart
```diff
@@ -235,23 +241,119 @@ class _AddAccountPageState extends State<AddAccountPage> {
...
-  State<PasswordLoginPage> createState() => _PasswordLoginPageState();
+  State<LoginPage> createState() => _LoginPageState();
 }
 
-class _PasswordLoginPageState extends State<PasswordLoginPage> {
+class _LoginPageState extends State<LoginPage> {
   bool _inProgress = false;
 
+  String? _otp;
+  String? _getOtp() {
```
gnprice, line 274: This is functionally a getter backed by a mutable field, so let's follow the Flutter style guide's order for those:

gnprice: Also could make this an actual getter/field pair. E.g. `String? get _otp` and `String? __otp;`.

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:zulip/api/model/web_auth.dart';
+
+import '../../example_data.dart' as eg;
+
+void main() {
+  group('WebAuthPayload', () {
+    const otp = '186f6d085a5621ebaf1ccfc05033e8acba57dae03f061705ac1e58c402c30a31';
+    final encryptedApiKey = debugEncodeApiKey(eg.selfAccount.apiKey, otp);
+    final wellFormed = Uri.parse(
+      'zulip://login?otp_encrypted_api_key=$encryptedApiKey&email=self%40example&user_id=1&realm=https%3A%2F%2Fchat.example%2F');
```
gnprice, line 12: nit: break line for readability
```suggestion
    final wellFormed = Uri.parse(
      'zulip://login?otp_encrypted_api_key=$encryptedApiKey'
        '&email=self%40example&user_id=1&realm=https%3A%2F%2Fchat.example%2F');
```

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    test('parse fails when an expected field is missing', () {
+      final queryParams = {...wellFormed.queryParameters}..remove('email');
+      final input = wellFormed.replace(queryParameters: queryParams);
+      check(() => WebAuthPayload.parse(input)).throws<FormatException>();
+    });
+
+    test('parse fails when otp_encrypted_api_key is nonsense', () {
+      final queryParams = {...wellFormed.queryParameters}..['otp_encrypted_api_key'] = 'asdf';
```
gnprice, line 44: nit: key detail is off past 80 columns, so break line:
```suggestion
      final queryParams = {...wellFormed.queryParameters}
        ..['otp_encrypted_api_key'] = 'asdf';
```

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,71 @@
...
+      final input = wellFormed.replace(host: 'foo');
+      check(() => WebAuthPayload.parse(input)).throws<FormatException>();
+    });
+
+    test('parse fails when scheme is not "zulip"', () {
+      final input = wellFormed.replace(scheme: 'https');
+      check(() => WebAuthPayload.parse(input)).throws<FormatException>();
+    });
+
+    test('decodeApiKey fails when otp is nonsense', () {
```
gnprice, line 59: nit:
```suggestion
    test('decodeApiKey fails when otp is wrong length', () {
```

Seems more precise about what's really being checked here. (And an OTP is random noise when it *is* properly generated, so the reader may wonder what it means for it to be "nonsense".)

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,71 @@
...
+
+void main() {
+  group('WebAuthPayload', () {
+    const otp = '186f6d085a5621ebaf1ccfc05033e8acba57dae03f061705ac1e58c402c30a31';
+    final encryptedApiKey = debugEncodeApiKey(eg.selfAccount.apiKey, otp);
+    final wellFormed = Uri.parse(
+      'zulip://login?otp_encrypted_api_key=$encryptedApiKey&email=self%40example&user_id=1&realm=https%3A%2F%2Fchat.example%2F');
+
+    test('basic happy case', () {
+      final result = WebAuthPayload.parse(wellFormed);
```
gnprice, line 15: nit: I like the name "payload" for similar locals elsewhere in this commit — I think that'd be clearer here than "result".

lib/api/model/web_auth.dart
```diff
@@ -0,0 +1,87 @@
...
+    final otpEncryptedApiKeyBytes = hex.decode(otpEncryptedApiKey);
+    if (otpBytes.length != otpEncryptedApiKeyBytes.length) {
+      throw const FormatException();
+    }
+    return String.fromCharCodes(Iterable.generate(otpBytes.length,
+      (i) => otpBytes[i] ^ otpEncryptedApiKeyBytes[i]));
+  }
+}
+
+String generateOtp() {
```
gnprice, line 71: Let's give this a unit test, too. For a smoke test: it returns a value (rather than throw), and the value is a hex string of the right length.

And then given that a previous draft of this function elsewhere actually [had a bug](https://github.com/zulip/zulip-flutter/pull/281#discussion_r1384405726) where the randomness was off — it had `rand.nextInt(255)` instead of 256 — let's have a test that would have caught that, so necessarily a probabilistic test.

We can follow the lead of our BackoffMachine test and say the probability of a false failure just needs to be under 1e-9. So…
* generate N = 216 of these
* take the set of bytes that ever appear (after decoding them all from hex)
* check that all 256 possible byte values do appear somewhere.

That should be plenty quick to run, I hope. And each possible byte value gets N * 32 opportunities to show up, each with probability 1/256; so its probability of missing all of those is exp(- N * 32 / 256) < 2e-12, and there are 256 such possible byte values so the probability that any of them gets missed is < 1e-9.

gnprice: Then since we'll be generating all those sample results, may as well use those to subsume the smoke test: just check each of them has the right length (and we'll already be decoding them as hex so implicitly checking they're hex strings).

One possible failure mode of a buggy implementation could be to drop leading zero bytes, for example, and this check would catch that.

gnprice (commented):
Thanks! Generally this all looks great. There's one more test I'd like, and then various small style things.

I also want to try this end-to-end myself, and look closer at those settings in the manifest and Info.plist. But posting these code comments without blocking on those.

lib/widgets/login.dart
chrisbobbe: > It looks like the existing code wasn't quite in that order

Hmm, really? Here's the class before the commit that adds to it:

```dart
class LoginPage extends StatefulWidget {
  const LoginPage({super.key, required this.serverSettings});

  final GetServerSettingsResult serverSettings;

  static Route<void> buildRoute({required GetServerSettingsResult serverSettings}) {
    return _LoginSequenceRoute(
      page: LoginPage(serverSettings: serverSettings));
  }

  @override
  State<LoginPage> createState() => _LoginPageState();
}
```

I see the following, matching the order in the style guide:

> 1. Constructors […]

> […]

> 4. Final fields that are set from the constructor.
> 5. Other static methods [i.e. that don't return the same type as the class].

> […]

> 10. Methods […]

chrisbobbe: I see that the updated class, with web-auth things added to it, doesn't match the numbered list; I'll fix that.

lib/widgets/login.dart
chrisbobbe: Interesting; yeah, I guess it's fine to have a thing that starts with two underscores. 😛

lib/widgets/login.dart
gnprice: > I see the following, matching the order in the style guide:

Oh, I guess that's right. I was reading `buildRoute` as returning the class's type, but it doesn't.

I think it's likely helpful to treat it for the ordering as if it does return the class's type, though. It plays pretty much the role of a factory constructor — it's the thing that application code, at least, should always use instead of calling the actual constructor directly. (Possibly we should even make the constructor private when we have a `buildRoute` method like this; but that might be annoying in tests.)

chrisbobbe:

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,97 @@
...
+      final manyOtps = List.generate(n, (_) => generateOtp());
+
+      final bytesThatAppear = <int>{};
+      for (final otp in manyOtps) {
+        final bytes = hex.decode(otp);
+        check(bytes.length).equals(32);
+        bytesThatAppear.addAll(bytes);
+      }
+
+      final byteValues = List.generate(256, (index) => index);
```
gnprice, line 80: nit: can use Iterable.generate instead of allocating a List, and then in fact can leave out the callback argument:
```suggestion
      final byteValues = Iterable.generate(256);
```

Then probably clearest to just inline that.

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,97 @@
...
+
+    test('decodeApiKey fails when otp is wrong length', () {
+      final payload = WebAuthPayload.parse(wellFormed);
+      check(() => payload.decodeApiKey('asdf')).throws<FormatException>();
+    });
+  });
+
+  group('generateOtp', () {
+    test('smoke, and check all 256 byte values are used', () {
+      const n = 216;
```
gnprice, line 70: Let's make the principle here a bit more explicit — helpful for anyone thinking of adding another probabilistic test and finding this as an example:
```suggestion
    test('smoke, and check all 256 byte values are used', () {
      // This is a probabilistic test.  We've chosen `n` so that when the test
      // should pass, the probability it fails is < 1e-9.  See analysis below.
      const n = 216;
```

test/api/model/web_auth_test.dart
```diff
@@ -0,0 +1,97 @@
...
+    });
+
+    test('decodeApiKey fails when otp is wrong length', () {
+      final payload = WebAuthPayload.parse(wellFormed);
+      check(() => payload.decodeApiKey('asdf')).throws<FormatException>();
+    });
+  });
+
+  group('generateOtp', () {
+    test('smoke, and check all 256 byte values are used', () {
```
gnprice, line 69: I did a quick test to confirm this test runs quickly, and it does. Running this test file with or without this test case included, 3 runs each, the difference in runtime is less than the noise between runs:
```
$ time flutter test test/api/model/web_auth_test.dart 
00:09 +8: All tests passed!                                                         
time: 10.660s wall (13.258s u, 0.864s s)

$ time flutter test test/api/model/web_auth_test.dart 
00:01 +8: All tests passed!                                                         
time: 2.808s wall (3.231s u, 0.613s s)

$ time flutter test test/api/model/web_auth_test.dart 
00:01 +8: All tests passed!                                                         
time: 2.724s wall (3.155s u, 0.596s s)

$ time flutter test test/api/model/web_auth_test.dart 
00:01 +8: All tests passed!                                                         
time: 2.702s wall (3.165s u, 0.581s s)

$ time flutter test test/api/model/web_auth_test.dart --name WebA
00:01 +7: All tests passed!                                                         
time: 2.764s wall (3.275s u, 0.580s s)

$ time flutter test test/api/model/web_auth_test.dart --name WebA
00:01 +7: All tests passed!                                                         
time: 2.759s wall (3.110s u, 0.642s s)

$ time flutter test test/api/model/web_auth_test.dart --name WebA
00:01 +7: All tests passed!                                                         
time: 2.710s wall (3.188s u, 0.568s s)
```

(I made one warmup run before the runs I counted, though I was surprised how big the warmup effect was.)

In particular I'm pretty sure the runtime is <50ms, which is fine.

gnprice (commented):
Thanks for the revision! Small comments on the new test; otherwise all looks good.

There's also those other review steps (https://github.com/zulip/zulip-flutter/pull/600#pullrequestreview-1971921342) that I want to make before merging.

gnprice:
Hmm, well — this doesn't work for me, when I try it on my device.

I tried rust-lang.zulipchat.com, and hit the GitHub button. In the UI, I got the "Something went wrong" dialog box. In logcat, there was this:
```
I/UrlLauncher(11715): component name for https://rust-lang.zulipchat.com/accounts/login/social/github?mobile_flow_otp=affc188a02b5d5fb2da77fc4215fab209dc466a71e279dd9589e51f65748f904 is null
I/UrlLauncher(11715): component name for https://flutter.dev is null
I/flutter (11715): Instance of 'Error'
```

That "component name for" message appears to come from `canLaunchUrl`. The `UrlLauncher` tag on the log line is a hint to look in the `url_launcher` package, and indeed:
```
greg@dark-matter:~/n/flutter/packages/packages/url_launcher (main u=) 
$ gr -C7 'component name'
url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
65-
66-  @Override
67-  public @NonNull Boolean canLaunchUrl(@NonNull String url) {
68-    Intent launchIntent = new Intent(Intent.ACTION_VIEW);
69-    launchIntent.setData(Uri.parse(url));
70-    String componentName = intentResolver.getHandlerComponentName(launchIntent);
71-    if (BuildConfig.DEBUG) {
72:      Log.i(TAG, "component name for " + url + " is " + componentName);
73-    }
74-    if (componentName == null) {
75-      return false;
```

If I just comment out our `canLaunchUrl` call, though:
```diff
@@ -321,9 +321,9 @@ class _LoginPageState extends State<LoginPage> {
     try {
       final url = widget.serverSettings.realmUrl.resolve(method.loginUrl)
         .replace(queryParameters: {'mobile_flow_otp': _otp!});
-      if (!(await ZulipBinding.instance.canLaunchUrl(url))) {
-        throw Error();
-      }
+      // if (!(await ZulipBinding.instance.canLaunchUrl(url))) {
+      //   throw Error();
+      // }
```
then…

* it does successfully launch the URL — I get the Zulip server's "choose account" web page, to choose from the different email addresses I have on my GitHub account
* when I choose one, I get a bit of system UI asking what app to open with, and I choose "Zulip beta" (this is something normal users, who only have one Zulip app installed at a time, should never see)
* then I'm back in the app's own UI, and:
  * there's that error dialog again
  * but also the progress indicator on the app bar
* and when I dismiss the error dialog, it's back on LoginPage as if nothing had happened.

And meanwhile in logcat there's this:
```
W/CustomTabsClient(11715): Unable to find any Custom Tabs packages, you may need to add a <queries> element to your manifest. See the docs for CustomTabsClient#getPackageName.
I/flutter (11715): 'package:zulip/widgets/login.dart': Failed assertion: line 290 pos 15: 'await ZulipBinding.instance.supportsCloseForLaunchMode(_webAuthLaunchMode)': is not true.
```


Then I commented out that other check:
```diff
-      assert (await ZulipBinding.instance.supportsCloseForLaunchMode(_webAuthLaunchMode));
+      // assert (await ZulipBinding.instance.supportsCloseForLaunchMode(_webAuthLaunchMode));
```

and tried again, and it works!

gnprice:
So in short it does work if we remove those two checks.

Both checks seem to be trying to confirm that the thing the code is about to do next is expected to work. So it's probably best to just try the thing, and if it doesn't work then that's an error we have to deal with.

gnprice:
This also probably indicates it's a good idea to make a minimal effort to report the actual error message when there's an error, because it's likely there will be errors some users run into and we'll want that information for debugging.

In particular if `e is Exception`, we can use `e.message`. The error dialog's title can remain short and generic like "Something went wrong".

android/app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,13 @@
                 <action android:name="android.intent.action.MAIN"/>
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
+            <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
+            <intent-filter>
```
gnprice, line 29: OK, the changes here and in Info.plist LGTM.

(The intent filter and `CFBundleUrlTypes` match what we have in zulip-mobile; the other bits are what's prescribed in the docs under https://docs.flutter.dev/ui/navigation/deep-linking .)

So what remains to do on this PR is those small comments at https://github.com/zulip/zulip-flutter/pull/600#pullrequestreview-1972204614 on the new test, and then the bigger point from my end-to-end testing after that.

test/api/model/web_auth_test.dart
chrisbobbe: Thanks!

chrisbobbe:
> In particular if `e is Exception`, we can use `e.message`.

Hmm, I'm seeing that `Exception` doesn't have a `message`.

chrisbobbe:
Thanks for the review! Revision pushed, and see my comment just before this: https://github.com/zulip/zulip-flutter/pull/600#issuecomment-2030901366

gnprice:
> Hmm, I'm seeing that `Exception` doesn't have a `message`.

Ah, yeah, I read too quickly:
```dart
abstract interface class Exception {
  factory Exception([var message]) => _Exception(message);
}
```

So it's an interface, but with no members — just a marker interface. And its "message" is only an argument to the factory constructor.

Anyway, we can do so for `PlatformException`, which is likely to include anything thrown by `launchUrl` or `closeInAppWebView`.

gnprice:
OK, retested and this version works out of the box. The changes look good, so I'll go ahead and merge. Glad to have this feature in!

Then I'll file a follow-up issue for putting more information in the error dialog, for the sake of debugging whatever errors people turn out to run into. Probably be good to do that soon after, while it's fresh in mind (and also so we have it soon for that debugging.)

Issue #353, opened 2023-11-01
Handle `<hr>` horizontal lines


gnprice:
We should handle the `<hr>` element.

This element means a "thematic break", shown as a dividing line; the name stands for  "horizontal rule". It's the result of a `---` line in Markdown.

Handling these means parsing them in `lib/model/content.dart`, and then styling them appropriately in `lib/widgets/content.dart`. For appropriate styling, see the Zulip web app's `web/styles/rendered_markdown.css`.

jassirr:
@gnprice @chrisbobbe can I take this one up?

jassirr:
@gnprice @chrisbobbe any update?

Khader-1:
Instructive, but I imagine having the link to [web/styles/rendered_markdown.css](https://github.com/zulip/zulip/blob/main/web/styles/rendered_markdown.css) would give more context for readers.
Anyways, I am working on it!

gnprice:
Seen by a beta user: [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1773998).

PR #603, opened 2024-03-29
content: Handle <hr> horizontal lines

Khader-1:
Final Result:

image

Fixes: #353

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -74,6 +74,8 @@ class BlockContentList extends StatelessWidget {
           // This goes in a Column.  So to get the effect of a newline,
           // just use an empty Text.
           return const Text('');
+        } else if (node is ThematicBreakNode) {
+          return const ThematicBreak();
         } else if (node is ParagraphNode) {
           return Paragraph(node: node);
         } else if (node is HeadingNode) {
@@ -104,6 +106,22 @@ class BlockContentList extends StatelessWidget {
         }
       }),
     ]);
+  }
+}
+
+class ThematicBreak extends StatelessWidget {
+  const ThematicBreak({super.key});
+
+  static const htmlHeight = 2.0;
+  static const htmlMarginY = 20.0;
+
+  @override
+  Widget build(BuildContext context) {
+    return Divider(
+      color: const HSLColor.fromAHSL(1, 0, 0, .87).toColor(),
+      thickness: htmlHeight,
+      height: 2 * htmlMarginY + htmlHeight,
+    );
   }
 }
 
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -521,6 +521,17 @@ class ContentExample {
       blockUnimplemented('more text'),
     ]]),
   ]);
+
+  static const thematicBreak =  ContentExample(
+    'parse thematic break (<hr>)',
+    // https://chat.zulip.org/#narrow/stream/7-test-here/near/1774718
+    'a\n---\nb',
+    '<p>a</p>\n<hr>\n<p>b</p>',
+    [
+      ParagraphNode(links: null, nodes: [TextNode('a')]),
+      ThematicBreakNode(),
+      ParagraphNode(links: null, nodes: [TextNode('b')]),
+    ]);
 }
 
 UnimplementedBlockContentNode blockUnimplemented(String html) {
@@ -720,6 +731,8 @@ void main() {
       ParagraphNode(links: null, nodes: [TextNode('a')]),
       LineBreakNode(),
     ]);
+
+  testParseExample(ContentExample.thematicBreak);
 
   testParse('parse two plain-text paragraphs',
     // "hello\n\nworld"
--- a/test/widgets/content_test.dart
+++ b/test/widgets/content_test.dart
@@ -62,6 +62,13 @@ void main() {
     });
   }
 
+  group('ThematicBreak', () {
+    testWidgets('smoke ThematicBreak', (tester) async {
+      await prepareContentBare(tester, ContentExample.thematicBreak.html);
+      tester.widget(find.byType(ThematicBreak));
+    });
+  });
+
   group('Heading', () {
     testWidgets('plain h6', (tester) async {
       await prepareContentBare(tester,
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -178,6 +178,19 @@ class LineBreakNode extends BlockContentNode {
   int get hashCode => 'LineBreakNode'.hashCode;
 }
 
+/// A `hr` element
+class ThematicBreakNode extends BlockContentNode {
+  const ThematicBreakNode({super.debugHtmlNode});
+
+  @override
+  bool operator ==(Object other) {
+    return other is ThematicBreakNode;
+  }
+
+  @override
+  int get hashCode => 'ThematicBreakNode'.hashCode;
+}
+
 /// A `p` element, or a place where the DOM tree logically wanted one.
 ///
 /// We synthesize these in the absence of an actual `p` element in cases where
@@ -967,6 +980,10 @@ class _ZulipContentParser {
 
     if (localName == 'br' && className.isEmpty) {
       return LineBreakNode(debugHtmlNode: debugHtmlNode);
+    }
+
+    if (localName == 'hr' && className.isEmpty) {
+      return ThematicBreakNode(debugHtmlNode: debugHtmlNode);
     }
 
     if (localName == 'p' && className.isEmpty) {


lib/widgets/content.dart
```diff
@@ -47,7 +47,7 @@ class InheritedMessage extends InheritedWidget {
 
   @override
   bool updateShouldNotify(covariant InheritedMessage oldWidget) =>
-    !identical(oldWidget.message, message);
+      !identical(oldWidget.message, message);
```
Khader-1: Irrelevant update.

lib/widgets/content.dart
```diff
@@ -107,6 +109,27 @@ class BlockContentList extends StatelessWidget {
...
 }
 
+/// A Divider matching the style found in:
+/// [web/styles/rendered_markdown.css](https://github.com/zulip/zulip/blob/main/web/styles/rendered_markdown.css)
+class ThematicBreak extends StatelessWidget {
+  const ThematicBreak({
+    super.key,
+    this.thickness = 2,
+    this.color,
+  });
```
chrisbobbe, line 119: Let's remove these params and just hard-code the intended values. This element doesn't need to appear with different thicknesses or colors in different places.

lib/widgets/content.dart
```diff
@@ -107,6 +109,27 @@ class BlockContentList extends StatelessWidget {
   }
 }
 
+/// A Divider matching the style found in:
+/// [web/styles/rendered_markdown.css](https://github.com/zulip/zulip/blob/main/web/styles/rendered_markdown.css)
+class ThematicBreak extends StatelessWidget {
```
chrisbobbe, line 114: I think we can write this widget without a dartdoc, like the other content widgets. A person reading the code will assume that our UI elements correspond with the web app's UI elements. It can be useful to refer to specific web-app code in places where we intentionally deviate from web (for example to adapt to the mobile environment, such as in the absence of hover interactions). But that's not what we're doing here.

lib/widgets/content.dart
```diff
@@ -107,6 +109,27 @@ class BlockContentList extends StatelessWidget {
...
+
+  final double thickness;
+  final Color? color;
+
+  @override
+  Widget build(BuildContext context) {
+    return Divider(
+      color: color ?? const HSLColor.fromAHSL(1, 0, 0, .87).toColor(),
+      thickness: 2,
+    );
```
chrisbobbe, line 129: Looking at the web app, I see that it puts more space above and below the line than we get with this. Let's match web, or else write down why we aren't doing so.

test/model/content_test.dart
```diff
@@ -875,6 +893,8 @@ void main() {
       ]]),
     ]);
 
+  testParseExample(ContentExample.thematicBreaks);
```
chrisbobbe, line 896: I think a better place to put this is just below the `'parse <br> in block context'` test. That seems consistent with the choice to define `ThematicBreakNode` just after `LineBreakNode`.

chrisbobbe (commented):
Thanks, this looks mostly good! Some small comments, below.

Also, let's add a some test for the new `ThematicBreak` widget, in test/widgets/content_test.dart. One shortcut to doing that is `testContentSmoke`, but unfortunately it won't help us here—it requires `ContentExample.expectedText` to be non-null, and these "thematic break" lines don't come with any text to check. 🙂 But there are some tests in that file in the same situation, and they end up being pretty simple. For example, the smoke tests for `MessageImageEmoji`.

Khader-1:
Thanks @chrisbobbe! I have restructured the code accordingly.

test/model/content_test.dart
```diff
@@ -521,6 +521,24 @@ class ContentExample {
...
+
+  static const thematicBreaks =  ContentExample(
+    'parse thematic break (<hr>) in block context',
+    '---'
+    '---'
+    'a'
+    '---'
+    'b'
+    '---',
+    '<hr><hr><p>a</p><hr><p>b</p><hr>',
```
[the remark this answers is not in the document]
gnprice: Then as a style nit: prefer to put these strings on relatively few lines here in the test source code, writing `\n` where necessary inside the string literals. When there's a lot going on in the string, it can use more lines in order to keep it readable; but this one is fairly simple, so `markdown` and `html` should use one line of source code each.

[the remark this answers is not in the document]
gnprice: A higher-level point of testing style: generally we try to keep each test case tightly focused, containing only the details that are needed in order to make the point of the test.

This test case feels like it's gesturing at several different things; it can be improved by being made shorter, to focus on just the minimum for demonstrating an `<hr>` element. Those other things (like two `<hr>` in a row?) can be exercised with their own focused tests, though I'm not sure we need tests for them.

The absolute minimum would be just `---`. Because it'd be kind of odd to send a message in normal use that's just a bare `<hr>`, it'd be reasonable to make it slightly longer to resemble normal use: like `a\n---\nb`. But then that's plenty.

test/widgets/content_test.dart
```diff
@@ -95,6 +95,15 @@ void main() {
     });
   });
 
+  group('ThematicBreak', () {
+    testWidgets('smoke ThematicBreak', (tester) async {
+      await prepareContentBare(tester,
+        // "---"
+        '<hr>');
```
gnprice, line 102: Let's reuse here the handy value you've defined on `ContentExample`.

lib/widgets/content.dart
```diff
@@ -167,6 +169,22 @@ class Heading extends StatelessWidget {
   }
 }
 
+class ThematicBreak extends StatelessWidget {
```
gnprice, line 172: nit: let's order these classes the same way as their invocations appear in `BlockContentList`  above.

(Then similarly for the test code.)

gnprice (commented):
Thanks @Khader-1, and thanks @chrisbobbe for the review!

A few more comments below, mainly on the tests.

Please also update the screenshot at the top of the PR for the current revision (since I believe you've adjusted the layout). That'll help for getting feedback on the visual result.

test/model/content_test.dart
```diff
@@ -521,6 +521,16 @@ class ContentExample {
       blockUnimplemented('more text'),
     ]]),
   ]);
+
+  static const thematicBreaks =  ContentExample(
+    'parse thematic break (<hr>) in block context',
+    'a\n---\nb',
+    '<p>a</p><hr><p>b</p>',
```
gnprice, line 528: Huh, is this the HTML the server outputs for this input? I'd have guessed there'd be some newlines in there, like between `</p>` and `<hr>`, based on how some other examples in this file look.

Would you link in the PR description to a test message on chat.zulip.org that demonstrates this example? (That's probably something we should start making a practice of, for better reproducibility. Or perhaps even put the link into the source code.)

lib/widgets/content.dart
```diff
@@ -56,6 +56,22 @@ class InheritedMessage extends InheritedWidget {
...
```
gnprice, line 59: This should come after `BlockContentList`, like the other widgets that are used there.

(Just in the same order as those, as mentioned at https://github.com/zulip/zulip-flutter/pull/603#discussion_r1545946957 .)

gnprice (commented):

test/model/content_test.dart
Khader-1: Apologies for the oversight. I adjusted the output for readability purposes, which resulted in a cleaner HTML string. However, I understand the importance of maintaining accuracy for testing purposes. Moving forward, I'll ensure the test output aligns precisely with the server's actual output.

gnprice: Cool, thanks. Yeah, readability is important, but for these tests their job is to make sure we're successfully parsing whatever the server sends us, with all its quirks and sometimes messes — so the top priority is for the data to represent those authentically.

For now let's just add the link in a comment, like this:
```dart
  static const thematicBreaks =  ContentExample(
    'parse thematic break (<hr>) in block context',
    // https://chat.zulip.org/#narrow/stream/7-test-here/near/1774718
    'a\n---\nb',
    '<p>a</p>\n<hr>\n<p>b</p>',
```

Then I'll file a separate issue for filling that in for the existing examples, adding it as a field, and also adding some automated validation of it.

gnprice: Filed #617.

test/model/content_test.dart
```diff
@@ -521,6 +521,16 @@ class ContentExample {
       blockUnimplemented('more text'),
     ]]),
   ]);
+
+  static const thematicBreaks =  ContentExample(
```
gnprice, line 525: nit: there's only one of these, so singular:
```suggestion
  static const thematicBreak =  ContentExample(
```

test/model/content_test.dart
```diff
@@ -521,6 +521,16 @@ class ContentExample {
...
+    'parse thematic break (<hr>) in block context',
```
gnprice, line 526: nit: "in block context" is redundant because that's the only way `<hr>` can appear:
```suggestion
    'parse thematic break (<hr>)',
```

(there's a phrase like that for the test that produces `LineBreakNode`, but that's to contrast it with the "inside a paragraph" version that produces `LineBreakInlineNode`)

gnprice (commented):
Thanks for the revision! Two wording nits below, and I replied above about the URL. I'll make those three tweaks and merge.

Issue #355, opened 2023-11-01
Blank text nodes after code blocks are "unimplemented"


gnprice:
In a few messages, the current prototype reports as "unimplemented" a text node that appears to consist of two newlines. Specifically I see this in two recent messages on chat.zulip.org:
https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/OpenAPI.20requestBody/near/1672619
https://chat.zulip.org/#narrow/stream/378-api-design/topic/unreads.3A.20access.20to.20messages/near/1668410

Both are from me; I don't know if there's some pattern in how I format things that happens to tickle this, or if that's a coincidence. In both cases, the unimplemented node appears just after a code block.

We should pin down what's happening in these messages and fix it so that we handle that HTML correctly with no complaints. For how to get the server's HTML for those messages, see comments in `test/model/content_test.dart`.

gnprice:
> We should pin down what's happening in these messages and fix it so that we handle that HTML correctly with no complaints.

FTR, we determined (https://github.com/zulip/zulip-flutter/pull/604#discussion_r1550614333) that the pattern that triggers this quirk in the server is for the message's Markdown source to express the code block with four-space-indent syntax, rather than triple-backticks syntax.

(And that does indeed cover the two original examples above, as well as all other examples we've found.)

PR #604, opened 2024-03-29
content: Handle blank text nodes after code blocks.

Khader-1:
|before|after|
|-----|----|
|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-30 at 02 18 02|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-30 at 02 18 45|

Fixes: #355

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -1124,6 +1124,8 @@ class _ZulipContentParser {
     return result;
   }
 
+  static final _redundantLineBreaksRegexp = RegExp(r'^\n+$');
+
   List<BlockContentNode> parseBlockContentList(dom.NodeList nodes) {
     assert(_debugParserContext == _ParserContext.block);
     final List<BlockContentNode> result = [];
@@ -1131,7 +1133,9 @@ class _ZulipContentParser {
     for (final node in nodes) {
       // We get a bunch of newline Text nodes between paragraphs.
       // A browser seems to ignore these; let's do the same.
-      if (node is dom.Text && (node.text == '\n')) continue;
+      if (node is dom.Text && _redundantLineBreaksRegexp.hasMatch(node.text)) {
+        continue;
+      }
 
       final block = parseBlockContent(node);
       if (block is ImageNode) {
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -301,6 +301,17 @@ class ContentExample {
         '\n</code></pre></div>'),
     ]);
 
+  static const codeBlockFollowedByMultipleLineBreaks = ContentExample(
+    'blank text nodes after code blocks',
+    '    code block.\n\nsome content',
+    // https://chat.zulip.org/#narrow/stream/7-test-here/near/1774823
+    '<div class="codehilite">'
+      '<pre><span></span><code>code block.\n</code></pre></div>\n\n'
+    '<p>some content</p>', [
+      CodeBlockNode([CodeBlockSpanNode(text: "code block.", type: CodeBlockSpanType.text)]),
+      ParagraphNode(links: null, nodes: [TextNode("some content")]),
+    ]);
+
   static final mathInline = ContentExample.inline(
     'inline math',
     r"$$ \lambda $$",
@@ -865,6 +876,7 @@ void main() {
   testParseExample(ContentExample.codeBlockHighlightedMultiline);
   testParseExample(ContentExample.codeBlockWithHighlightedLines);
   testParseExample(ContentExample.codeBlockWithUnknownSpanType);
+  testParseExample(ContentExample.codeBlockFollowedByMultipleLineBreaks);
 
   testParseExample(ContentExample.mathBlock);
   testParseExample(ContentExample.mathBlockInQuote);


test/model/content_test.dart
```diff
@@ -875,6 +890,8 @@ void main() {
       ]]),
     ]);
 
+    testParseExample(ContentExample.lineBreaksAfterCodeBlocks);
```
Khader-1, line 893: NIT: fix indentation!

test/model/content_test.dart
```diff
@@ -521,6 +521,21 @@ class ContentExample {
       blockUnimplemented('more text'),
     ]]),
   ]);
+
+  static const lineBreaksAfterCodeBlocks = ContentExample(
+    'blank text nodes after code blocks',
+    '    code block''\n\n''some content',
+    '<div class="codehilite">'
+      '<pre><code>'
```
chrisbobbe, line 529: Hmm, when I do it, I get something different here:

```suggestion
      '<pre><span></span><code>'
```

Here's my message: https://chat.zulip.org/#narrow/stream/7-test-here/topic/content/near/1770526

test/model/content_test.dart
```diff
@@ -521,6 +521,21 @@ class ContentExample {
       blockUnimplemented('more text'),
     ]]),
   ]);
+
+  static const lineBreaksAfterCodeBlocks = ContentExample(
+    'blank text nodes after code blocks',
+    '    code block''\n\n''some content',
```
chrisbobbe, line 527:
```suggestion
    '    code block.\n\nsome content',
```

chrisbobbe (commented):
Thanks! Small comments below.

Khader-1 (commented):
Thanks for the review @chrisbobbe! I believe this is ready now.

test/model/content_test.dart
```diff
@@ -301,6 +301,21 @@ class ContentExample {
...
+    'blank text nodes after code blocks',
+    '    code block.\n\nsome content',
+    '<div class="codehilite">'
+      '<pre><span></span><code>'
+        'code block.\n'
+      '</code></pre>'
+    '</div>\n\n'
+    '<p>'
+      'some content'
+    '</p>', [
```
gnprice, line 314: nit: let's make this test data a bit more compact: the whole `<div>` element on two lines, and the `<p>` element on one line. See the existing `codeBlock…` cases for examples.

test/model/content_test.dart
```diff
@@ -301,6 +301,21 @@ class ContentExample {
         '\n</code></pre></div>'),
     ]);
 
+  static const lineBreaksAfterCodeBlocks = ContentExample(
```
gnprice, line 304: nit: let's name this starting with `codeBlock`, to help group it with the other examples about code blocks

lib/model/content.dart
```diff
@@ -1107,7 +1107,7 @@ class _ZulipContentParser {
     for (final node in nodes) {
       // We get a bunch of newline Text nodes between paragraphs.
       // A browser seems to ignore these; let's do the same.
-      if (node is dom.Text && (node.text == '\n')) continue;
+      if (node is dom.Text && RegExp(r'^\n+$').hasMatch(node.text)) continue;
```
gnprice, line 1110: This parsing code is a hot path, so I try to stay fairly paranoid about performance here.

(A bit more discussion here — I should probably add this point to the README or a style guide:
https://github.com/zulip/zulip-flutter/pull/587#discussion_r1539898705
https://github.com/zulip/zulip-flutter/pull/587#discussion_r1546882850
)

As part of that, instead of making a RegExp object inside a parsing function, when we need one it's a `static final` member on the class. For examples, search for "RegExp" in this file.

test/model/content_test.dart
```diff
@@ -301,6 +301,21 @@ class ContentExample {
...
+    'blank text nodes after code blocks',
+    '    code block.\n\nsome content',
```
gnprice, line 306: Interesting! I wonder if this is the explanation for all the times we've seen this symptom. It hasn't felt super common in my experience — which would be consistent with that, because I think this four-space-indent syntax is not super commonly used in Zulip messages (because the triple-backticks syntax is usually more convenient).

gnprice (commented):
Thanks! Comments below.

test/model/content_test.dart
Khader-1: I attempted to replicate the issue by copying the messages mentioned in the report. I can confirm that this occurs exclusively when utilizing the four-space-indent syntax. As my attempts to reproduce it using triple-backticks were unsuccessful.

Additionally, it is noteworthy that I couldn't reproduce the issue when typing on the mobile app. It was only after sending the message from the web client that the double line breaks appeared.

test/model/content_test.dart
```diff
@@ -301,6 +301,16 @@ class ContentExample {
         '\n</code></pre></div>'),
     ]);
 
+  static const codeBlockFollowedByMultipleLineBreaks = ContentExample(
+    'blank text nodes after code blocks',
+    '    code block.\n\nsome content',
+    '<div class="codehilite">'
```
gnprice, line 307: Oh and: let's add a link here to an example message, just like we did in #603.

test/model/content_test.dart
Khader-1: Upon further investigation, I've determined that this issue is associated with the 'trim' function utilized prior to dispatching the message in the mobile application:

```dart
// lib/widgets/compose_box.dart:69
@override
String _computeTextNormalized() {
  String trimmed = text.trim();
  return trimmed.isEmpty ? kNoTopicTopic : trimmed;
}
```

I'm uncertain whether this behavior should be altered to align with the web version or if it's acceptable to maintain its current state. I'd recommend creating a new issue to handle this. What are your thoughts, @chrisbobbe?

Khader-1:
Appreciate the feedback, @gnprice. I believe it's good to go now, @chrisbobbe; would love to hear your insights on the [trim behavior](https://github.com/zulip/zulip-flutter/pull/604#discussion_r1555878637).

test/model/content_test.dart
chrisbobbe: Thanks for the investigation, @Khader-1! It sounds like you've found that the web client doesn't trim leading and trailing whitespace from the raw message content sent in the send-message request. Is that right?

(I think you mean the same method of `ComposeContentController`:

```dart
  @override
  String _computeTextNormalized() {
    return text.trim();
  }
```

The one you quoted belongs to `ComposeTopicController`.)

I've always assumed it's a good idea for the client to trim extra whitespace from the start and end of raw message content. It just seems neater that way, to me. If it's true that trimming is preferable (and I might be wrong), I would wonder why the web client doesn't do so. 🙂 I wonder if there's some Markdown syntax that requires leading or trailing whitespace and is defeated when the whitespace is removed.

If you want, you could raise the question of why the web app doesn't do trimming, and whether clients should trim or not, on [CZO](https://zulip.com/development-community/). Perhaps in the `#api design` stream, because it's about the expectations of the [send-message endpoint](https://zulip.com/api/send-message). But it seems like the answer doesn't need to block progress on the current issue.

test/model/content_test.dart
```diff
@@ -301,6 +301,17 @@ class ContentExample {
         '\n</code></pre></div>'),
     ]);
 
+  static const codeBlockFollowedByMultipleLineBreaks = ContentExample(
+    'blank text nodes after code blocks',
+    '    code block.\n\nsome content',
+    // https://chat.zulip.org/#narrow/stream/7-test-here/topic/2.20new.20lines/near/1774823
```
chrisbobbe, line 307: nit: leave out topic to reduce visual clutter, as we did in #603 and as specified in #617, so:

```dart
    // https://chat.zulip.org/#narrow/stream/7-test-here/near/1774823
```

chrisbobbe (approved):
Thanks, @Khader-1, looks great!! Just one nit below, which I'll fix and then merge.

Issue #156, opened 2023-06-07
ui: Distinguish `isBot: true` users with a bot marker


chrisbobbe:
As Alex [points out](https://chat.zulip.org/#narrow/stream/48-mobile/topic/no.20marker.20for.20bots/near/1587460) about the zulip-mobile app:
> The mobile UI doesn't differentiate between bots and real humans in any way -- except clicking into the user profile and noticing the `-bot@chat.zulip.org` email address (though that page will also show them as a "member" which seems further confusing).  With the profusion of LLM bots running around, this can be moderately misleading to users.

We should implement this in the Flutter app.

The web app currently adds a little gray robot icon next to the user's name:

<img width="153" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/ed42e347-4037-442c-b2f0-a08ed056314f">

gnprice:
Once we take care of #200, we'll have a straightforward way to pull in our own custom icons.

If we pick up this issue before then, we can instead just use one of the Material Design icons that Flutter makes convenient through [`Icons`](https://api.flutter.dev/flutter/material/Icons-class.html). Then this icon will just be one of the icons that we sweep through and replace as part of  #200 or a followup.

sm-sayedi:
I would like to take this one!

<img src="https://github.com/zulip/zulip-flutter/assets/59946442/7a99bd7e-3d36-4b53-859c-6c86020987ad" width="300" />

PR #605, opened 2024-03-29
msglist: Distinguish `isBot: true` message senders with a bot marker

sm-sayedi:
A bot icon is shown with the name of the bot users.

<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

Fixes: #156

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -896,7 +896,10 @@ class MessageWithPossibleSender extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+
     final message = item.message;
+    final sender = store.users[message.senderId];
 
     Widget? senderRow;
     if (item.showSender) {
@@ -926,6 +929,14 @@ class MessageWithPossibleSender extends StatelessWidget {
                       ).merge(weightVariableTextStyle(context, wght: 600,
                                 wghtIfPlatformRequestsBold: 900)),
                       overflow: TextOverflow.ellipsis)),
+                  if (sender?.isBot ?? false) ...[
+                    const SizedBox(width: 5),
+                    const Icon(
+                      ZulipIcons.bot,
+                      size: 15,
+                      color: Color.fromARGB(255, 159, 173, 173),
+                    ),
+                  ],
                 ]))),
           const SizedBox(width: 4),
           Text(time,
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -43,6 +43,7 @@ void main() {
     int? messageCount,
     List<Message>? messages,
     List<ZulipStream>? streams,
+    List<User>? users,
     List<Subscription>? subscriptions,
     UnreadMessagesSnapshot? unreadMsgs,
   }) async {
@@ -55,6 +56,7 @@ void main() {
 
     // prepare message list data
     store.addUser(eg.selfUser);
+    store.addUsers(users ?? []);
     assert((messageCount == null) != (messages == null));
     messages ??= List.generate(messageCount!, (index) {
       return eg.streamMessage(sender: eg.selfUser);
@@ -476,6 +478,49 @@ void main() {
 
       await handleNewAvatarEventAndPump(tester, '/bar.jpg');
       checkResultForSender('/bar.jpg');
+
+      debugNetworkImageHttpClientProvider = null;
+    });
+
+    testWidgets('Bot user is distinguished by showing an icon', (tester) async {
+      // When using this function, provide only one bot user
+      // to [PerAccountStore] through [setupMessageListPage] function.
+      void checkUser(User user, {required bool isBot}) {
+        final nameFinder = find.text(user.fullName);
+        final botFinder = find.byIcon(ZulipIcons.bot);
+
+        check(nameFinder.evaluate().singleOrNull).isNotNull();
+        check(botFinder.evaluate().singleOrNull).isNotNull();
+
+        final userFinder = find.ancestor(
+          of: nameFinder,
+          matching: find.ancestor(
+            of: botFinder,
+            matching: find.byType(Row),
+          ));
+
+        isBot
+          ? check(userFinder.evaluate()).isNotEmpty()
+          : check(userFinder.evaluate()).isEmpty();
+      }
+
+      prepareBoringImageHttpClient();
+
+      final users = [
+        eg.user(fullName: 'User 1', isBot: true),
+        eg.user(fullName: 'User 2', isBot: false),
+        eg.user(fullName: 'User 3', isBot: false),
+      ];
+
+      await setupMessageListPage(
+        tester,
+        messages: users.map((user) => eg.streamMessage(sender: user)).toList(),
+        users: users,
+      );
+
+      checkUser(users[0], isBot: true);
+      checkUser(users[1], isBot: false);
+      checkUser(users[2], isBot: false);
 
       debugNetworkImageHttpClientProvider = null;
     });
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -1,5 +1,4 @@
 import 'dart:convert';
-import 'dart:io';
 
 import 'package:checks/checks.dart';
 import 'package:flutter/material.dart';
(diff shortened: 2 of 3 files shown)


test/widgets/message_list_test.dart
```diff
@@ -484,6 +487,22 @@ void main() {
 
       debugNetworkImageHttpClientProvider = null;
     });
+
+    testWidgets('Bot user is distinguished by showing an icon', (tester) async {
+      prepareBoringImageHttpClient();
```
sm-sayedi, line 492: In the test before this test, I noticed that there was a fake `HttpClient` set, that could be replaced by calling this method that was factored out in [#590](https://github.com/zulip/zulip-flutter/pull/590/commits/4ab1a82ebf5f71ba27413fded606af075c871b50).

```diff
testWidgets('Updates avatar on RealmUserUpdateEvent', (tester) async {

-  final httpClient = FakeImageHttpClient();
-    debugNetworkImageHttpClientProvider = () => httpClient;
-    httpClient.request.response
-      ..statusCode = HttpStatus.ok
-      ..content = kSolidBlueAvatar;

+  prepareBoringImageHttpClient();
  
}
```

Is it ok if I add an additional [NFC] commit to this PR with the requested change, or should we just ignore it for now?

lib/widgets/message_list.dart
```diff
@@ -926,6 +928,10 @@ class MessageWithPossibleSender extends StatelessWidget {
                       ).merge(weightVariableTextStyle(context, wght: 600,
                                 wghtIfPlatformRequestsBold: 900)),
                       overflow: TextOverflow.ellipsis)),
+                  if (messageSender?.isBot ?? false) ...[
+                    const SizedBox(width: 8),
+                    const Icon(ZulipIcons.bot, size: 18, color: Colors.grey),
```
chrisbobbe, line 933: Let's match the color to what the web app uses.

I'm also curious how the 18px size is chosen, and the 8px distance from the user's name. To my eye I think those values might both want to be a bit smaller. What does the web app do?

test/widgets/message_list_test.dart
chrisbobbe: Seems reasonable to me! 🙂

test/widgets/message_list_test.dart
```diff
@@ -484,6 +487,22 @@ void main() {
...
+
+      final users = List.generate(3, (index) => eg.user(isBot: index == 0));
+
+      await setupMessageListPage(
+        tester,
+        messages: users.map((user) => eg.streamMessage(sender: user)).toList(),
+        users: users,
+      );
+
+      tester.widget(find.byIcon(ZulipIcons.bot));
```
chrisbobbe, line 502: Can you check specifically that the icon appears next to the sender that is a bot, and also that an icon _doesn't_ appear on the senders that are not bots?

chrisbobbe (commented):
Thanks, this looks mostly good! Small comments below.

sm-sayedi:
Thanks for the review! A few comments below.

lib/widgets/message_list.dart
sm-sayedi: > This should be fixed. Even on the webapp the bot icon isn't super prominent, and I have no idea why it's blurred.

I saw [this](https://chat.zulip.org/#narrow/stream/48-mobile/topic/no.20marker.20for.20bots/near/1587470) comment on CZO, so I decided to go with a more prominent icon. Any recommendations from your side are most welcome!

test/widgets/message_list_test.dart
sm-sayedi: Sure thing!

lib/widgets/message_list.dart
sm-sayedi: On the web, the icon size is 12px, the color is rgb(159, 173, 173), and the distance to the user's name is 5px. 
With the same properties applied, it looks like this:



Should I go with this?

chrisbobbe: Hmm. To me, that all looks good except the size; I think it looks too small now. The font size in the web app is smaller than in the mobile app, so maybe this icon should be scaled up according to that. Maybe try 15px?

test/widgets/message_list_test.dart
```diff
@@ -484,6 +487,22 @@ void main() {
 
       debugNetworkImageHttpClientProvider = null;
     });
+
+    testWidgets('Bot user is distinguished by showing an icon', (tester) async {
+      prepareBoringImageHttpClient();
+
+      final users = List.generate(3, (index) => eg.user(isBot: index == 0));
```
gnprice, line 494: This is probably clearer as a list literal with three explicit `eg.user` calls: `[eg.user(isBot: true), eg.user(isBot: false),` etc. Not much longer that way, and less logic for the reader to unpack.

lib/widgets/message_list.dart
```diff
@@ -897,6 +897,8 @@ class MessageWithPossibleSender extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final message = item.message;
+    final messageSender =
+      PerAccountStoreWidget.of(context).users[message.senderId];
```
gnprice, line 901: nit: (two nits)
```suggestion
    final store = PerAccountStoreWidget.of(context);
    final message = item.message;
    final sender = store.users[message.senderId];
```

That way the result of `PerAccountStoreWidget.of` goes into a `store` local, like in most of our other build methods where it appears. The uniformity is mildly helpful; and this way also makes it maximally easy to add another bit of logic using the store later, without either duplicating the lookup or rewriting these lines.

Separately, most everything within this method is about the particular message `message`, so a name "sender" communicates basically the same meaning as "message sender". Similarly the existing other locals include `senderRow` and `time`, without the added qualifier "message".

gnprice (commented):
Thanks @sm-sayedi for handling this, and @chrisbobbe for the review!

Other than the things Chris remarked on above, this all looks good to me, except a few nits below.

gnprice:
Oh and one other nit, in the commit message:

    ui: Distinguish `isBot: true` users with a bot marker
    
    Fixes: #156

Let's make the prefix `msglist:` — that's more specific than "ui", and lets the reader know this is about how we show users in specifically the message list.

Similarly, s/users/message senders/ would further help pin down where in the UI we're talking about (and the resulting summary line is 68 columns, so there's room; summary lines can go a bit over 70).

lib/widgets/message_list.dart
sm-sayedi: Applied the changes. Updated the description for the new images. Please have a look!

sm-sayedi:
Thanks for the review! Pushed the latest revision. PTAL!

sm-sayedi:
> Similarly, s/users/message senders/ would further help pin down where in the UI we're talking about (and the resulting summary line is 68 columns, so there's room; summary lines can go a bit over 70).
Sorry, I am not sure where to add this in the commit message!

gnprice:
> > Similarly, s/users/message senders/ would further help pin down where in the UI we're talking about (and the resulting summary line is 68 columns, so there's room; summary lines can go a bit over 70).
> 
> Sorry, I am not sure where to add this in the commit message!

Ah, that was perhaps a bit cryptic of a way to put it. What I meant by "s/users/message senders/" was "replace 'users' with 'message senders'".

(The `s///` syntax is found in various venerable tools from the Unix world. Here's [Perl's documentation for it](https://perldoc.perl.org/perlretut#Search-and-replace).)

gnprice:
Thanks for the revision!

This all looks good to me except one more commit-message nit:
23b1fe1d0 msglist-test [nfc]: Refactor some code to use `prepareBoringImageHttpClient`
f2cf926db msglist: Distinguish `isBot: true` message senders with a bot marker

For that first commit, the prefix should be `msglist test [nfc]:`, with a space, rather than `msglist-test [nfc]:`. In general if you look at `git log --oneline`, you'll see that the "test" modifier is always separated by a space.

Other than that I'll leave this for @chrisbobbe to re-review for the icon size and placement, and anything else he notices, and then merge.

sm-sayedi:

Now waiting for a review from your side @chrisbobbe!

chrisbobbe:
Thanks, looks good to me! Merging.

After this, there are still some places in the UI where we show the name of a bot user without making it clear it's a bot. For example, it would probably be helpful to show the bot icon next to a bot user's name on the profile page (`ProfilePage`). I'm not sure if we need to show the icon in _all_ the places where names appear, though. Would you be willing to file a new issue with a checklist of all the places where bots' names might be displayed in the app, indicating which ones should start marking bots with the bot icon, which ones don't need to do that, and which ones you're not sure about?

sm-sayedi:
> Would you be willing to file a new issue with a checklist of all the places where...

Sure, will do so.

PR #608, opened 2024-04-02
autocomplete: Sort user-mention autocomplete results

sm-sayedi:
In @-mention autocomplete, users are suggested based on:

1. Recent activity in the current topic/stream. 
2. Recent DM conversations.
3. Human vs. Bot users.
4. Alphabetical order.

If one option exhausts, the other is considered for the suggestions in the preceding order.

_**Note:** This PR will no longer receive any new code pushes as it is divided into other PRs. See #693, #692, #828._

### DIFF
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -125,23 +125,26 @@ class AutocompleteViewManager {
     assert(removed);
   }
 
-  void handleRealmUserAddEvent(RealmUserAddEvent event) {
+  /// Recomputes the autocomplete results for users.
+  ///
+  /// Calls [MentionAutocompleteView.refreshStaleUserResults] for all that are registered.
+  void _refreshStaleUserResults() {
     for (final view in _mentionAutocompleteViews) {
       view.refreshStaleUserResults();
     }
   }
 
+  void handleRealmUserAddEvent(RealmUserAddEvent event) {
+    _refreshStaleUserResults();
+  }
+
   void handleRealmUserRemoveEvent(RealmUserRemoveEvent event) {
-    for (final view in _mentionAutocompleteViews) {
-      view.refreshStaleUserResults();
-    }
+    _refreshStaleUserResults();
     autocompleteDataCache.invalidateUser(event.userId);
   }
 
   void handleRealmUserUpdateEvent(RealmUserUpdateEvent event) {
-    for (final view in _mentionAutocompleteViews) {
-      view.refreshStaleUserResults();
-    }
+    _refreshStaleUserResults();
     autocompleteDataCache.invalidateUser(event.userId);
   }
 
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -267,13 +267,34 @@ class MentionAutocompleteView extends ChangeNotifier {
 
   List<User>? _sortedUsers;
 
-  List<User> sortByRelevance({required List<User> users}) {
+  int compareByRelevance({
+    required User userA,
+    required User userB,
+  }) {
+    final dmPrecedence = store.recentDmConversationsView.compareByDms(userA, userB);
+    return dmPrecedence;
+  }
+
+  List<User> sortByRelevance({
+    required List<User> users,
+    required Narrow narrow,
+  }) {
+    switch (narrow) {
+      case StreamNarrow():
+      case TopicNarrow():
+      case DmNarrow():
+        users.sort((userA, userB) => compareByRelevance(
+          userA: userA,
+          userB: userB));
+      case AllMessagesNarrow():
+        // do nothing in this case for now
+    }
     return users;
   }
 
   void _sortUsers() {
     final users = store.users.values.toList();
-    _sortedUsers = sortByRelevance(users: users);
+    _sortedUsers = sortByRelevance(users: users, narrow: narrow);
   }
 
   Future<List<MentionAutocompleteResult>?> _computeResults(MentionAutocompleteQuery query) async {
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -293,7 +293,12 @@ class MentionAutocompleteView extends ChangeNotifier {
     } else if (userA.isBot && !userB.isBot) {
       return 1;
     }
-    return 0;
+
+    final userAName = store.autocompleteViewManager.autocompleteDataCache
+      .normalizedNameForUser(userA);
+    final userBName = store.autocompleteViewManager.autocompleteDataCache
+      .normalizedNameForUser(userB);
+    return userAName.compareTo(userBName);
   }
 
   List<User> sortByRelevance({
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -270,7 +270,19 @@ class MentionAutocompleteView extends ChangeNotifier {
   int compareByRelevance({
     required User userA,
     required User userB,
+    required int? streamId,
+    required String? topic,
   }) {
+    if (streamId != null) {
+      final conversationPrecedence = store.recentSenders.compareByRecency(
+          userA,
+          userB,
+          streamId: streamId,
+          topic: topic);
+      if (conversationPrecedence != 0) {
+        return conversationPrecedence;
+      }
+    }
     final dmPrecedence = store.recentDmConversationsView.compareByDms(userA, userB);
     return dmPrecedence;
   }
@@ -281,11 +293,23 @@ class MentionAutocompleteView extends ChangeNotifier {
   }) {
     switch (narrow) {
       case StreamNarrow():
+        users.sort((userA, userB) => compareByRelevance(
+          userA: userA,
+          userB: userB,
+          streamId: narrow.streamId,
+          topic: null));
       case TopicNarrow():
+        users.sort((userA, userB) => compareByRelevance(
+          userA: userA,
+          userB: userB,
+          streamId: narrow.streamId,
+          topic: narrow.topic));
       case DmNarrow():
         users.sort((userA, userB) => compareByRelevance(
           userA: userA,
-          userB: userB));
+          userB: userB,
+          streamId: null,
+          topic: null));
       case AllMessagesNarrow():
         // do nothing in this case for now
     }
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -148,6 +148,14 @@ class AutocompleteViewManager {
     autocompleteDataCache.invalidateUser(event.userId);
   }
 
+  void handleMessageEvent(MessageEvent event) {
+    _refreshStaleUserResults();
+  }
+
+  void handleOlderMessages() {
+    _refreshStaleUserResults();
+  }
+
   /// Called when the app is reassembled during debugging, e.g. for hot reload.
   ///
   /// Calls [MentionAutocompleteView.reassemble] for all that are registered.
@@ -193,6 +201,7 @@ class MentionAutocompleteView extends ChangeNotifier {
   @override
   void dispose() {
     store.autocompleteViewManager.unregisterMentionAutocomplete(this);
+    _sortedUsers = null;
     // We cancel in-progress computations by checking [hasListeners] between tasks.
     // After [super.dispose] is called, [hasListeners] returns false.
     // TODO test that logic (may involve detecting an unhandled Future rejection; how?)
@@ -216,6 +225,7 @@ class MentionAutocompleteView extends ChangeNotifier {
   /// Called in particular when we get a [RealmUserEvent].
   void refreshStaleUserResults() {
     if (_query != null) {
+      _sortedUsers = null;
       _startSearch(_query!);
     }
   }
@@ -225,6 +235,7 @@ class MentionAutocompleteView extends ChangeNotifier {
   /// This will redo the search from scratch for the current query, if any.
   void reassemble() {
     if (_query != null) {
+      _sortedUsers = null;
       _startSearch(_query!);
     }
   }
@@ -254,22 +265,43 @@ class MentionAutocompleteView extends ChangeNotifier {
     notifyListeners();
   }
 
+  List<User>? _sortedUsers;
+
+  List<User> sortByRelevance({required List<User> users}) {
+    return users;
+  }
+
+  void _sortUsers() {
+    final users = store.users.values.toList();
+    _sortedUsers = sortByRelevance(users: users);
+  }
+
   Future<List<MentionAutocompleteResult>?> _computeResults(MentionAutocompleteQuery query) async {
     final List<MentionAutocompleteResult> results = [];
-    final Iterable<User> users = store.users.values;
 
-    final iterator = users.iterator;
+    if (_sortedUsers == null) {
+      _sortUsers();
+    }
+
+    final sortedUsers = _sortedUsers!;
+    final iterator = sortedUsers.iterator;
     bool isDone = false;
     while (!isDone) {
       // CPU perf: End this task; enqueue a new one for resuming this work
       await Future(() {});
+
+      if (_sortedUsers != sortedUsers) {
+        // The list of users this loop has been working from has become stale.
+        // Abort so _startSearch can retry with the new list.
+        throw ConcurrentModificationError();
+      }
 
       if (query != _query || !hasListeners) { // false if [dispose] has been called.
         return null;
       }
 
       for (int i = 0; i < 1000; i++) {
-        if (!iterator.moveNext()) { // Can throw ConcurrentModificationError
+        if (!iterator.moveNext()) {
           isDone = true;
           break;
         }
@@ -280,7 +312,7 @@ class MentionAutocompleteView extends ChangeNotifier {
         }
       }
     }
-    return results; // TODO(#228) sort for most relevant first
+    return results;
   }
 }
 
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -284,7 +284,16 @@ class MentionAutocompleteView extends ChangeNotifier {
       }
     }
     final dmPrecedence = store.recentDmConversationsView.compareByDms(userA, userB);
-    return dmPrecedence;
+    if (dmPrecedence != 0) {
+      return dmPrecedence;
+    }
+
+    if (!userA.isBot && userB.isBot) {
+      return -1;
+    } else if (userA.isBot && !userB.isBot) {
+      return 1;
+    }
+    return 0;
   }
 
   List<User> sortByRelevance({
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -426,13 +426,21 @@ class MentionAutocompleteQuery {
 }
 
 class AutocompleteDataCache {
+  final Map<int, String> _normalizedNamesByUser = {};
+
+  /// The lowercase `fullName` of [user].
+  String normalizedNameForUser(User user) {
+    return _normalizedNamesByUser[user.userId] ??= user.fullName.toLowerCase();
+  }
+
   final Map<int, List<String>> _nameWordsByUser = {};
 
   List<String> nameWordsForUser(User user) {
-    return _nameWordsByUser[user.userId] ??= user.fullName.toLowerCase().split(' ');
+    return _nameWordsByUser[user.userId] ??= normalizedNameForUser(user).split(' ');
   }
 
   void invalidateUser(int userId) {
+    _normalizedNamesByUser.remove(userId);
     _nameWordsByUser.remove(userId);
   }
 }
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -273,6 +273,10 @@ class MentionAutocompleteView extends ChangeNotifier {
     required int? streamId,
     required String? topic,
   }) {
+    // TODO(#234): give preference to "all", "everyone" or "stream".
+
+    // TODO(#618): give preference to subscribed users first.
+
     if (streamId != null) {
       final conversationPrecedence = store.recentSenders.compareByRecency(
           userA,
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -381,6 +381,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     for (final message in result.messages) {
       if (_messageVisible(message)) {
         _addMessage(message);
+        store.recentSenders.handleMessage(message);
       }
     }
     _fetched = true;
@@ -417,6 +418,9 @@ class MessageListView with ChangeNotifier, _MessageSequence {
         ? result.messages // Avoid unnecessarily copying the list.
         : result.messages.where(_messageVisible);
 
+      for (final message in fetchedMessages) {
+        store.recentSenders.handleMessage(message);
+      }
       store.autocompleteViewManager.handleOlderMessages();
 
       _insertAllMessages(0, fetchedMessages);
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -417,6 +417,8 @@ class MessageListView with ChangeNotifier, _MessageSequence {
         ? result.messages // Avoid unnecessarily copying the list.
         : result.messages.where(_messageVisible);
 
+      store.autocompleteViewManager.handleOlderMessages();
+
       _insertAllMessages(0, fetchedMessages);
       _haveOldest = result.foundOldest;
     } finally {
--- a/lib/model/recent_dm_conversations.dart
+++ b/lib/model/recent_dm_conversations.dart
@@ -1,3 +1,5 @@
+import 'dart:math';
+
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 
@@ -19,9 +21,22 @@ class RecentDmConversationsView extends ChangeNotifier {
         DmNarrow.ofRecentDmConversation(conversation, selfUserId: selfUserId),
         conversation.maxMessageId,
       )).toList()..sort((a, b) => -a.value.compareTo(b.value));
+    final map = Map.fromEntries(entries);
+    final sorted = QueueList.from(entries.map((e) => e.key));
+
+    final msgsByUser = <int, int>{};
+    for (final entry in entries) {
+      final dmNarrow = entry.key;
+      final msg = entry.value;
+      for (final userId in dmNarrow.otherRecipientIds) {
+        // only take the latest message of a user among all the conversations.
+        msgsByUser.putIfAbsent(userId, () => msg);
+      }
+    }
     return RecentDmConversationsView._(
-      map: Map.fromEntries(entries),
-      sorted: QueueList.from(entries.map((e) => e.key)),
+      map: map,
+      sorted: sorted,
+      latestMessagesByRecipient: msgsByUser,
       selfUserId: selfUserId,
     );
   }
@@ -29,6 +44,7 @@ class RecentDmConversationsView extends ChangeNotifier {
   RecentDmConversationsView._({
     required this.map,
     required this.sorted,
+    required this.latestMessagesByRecipient,
     required this.selfUserId,
   });
 
@@ -38,7 +54,23 @@ class RecentDmConversationsView extends ChangeNotifier {
   /// The [DmNarrow] keys of [map], sorted by latest message descending.
   final QueueList<DmNarrow> sorted;
 
+  /// Map from user ID to the latest message ID in any conversation with the user.
+  ///
+  /// Both 1:1 and group DM conversations are considered.
+  /// The self-user ID is excluded even if there is a self-DM conversation.
+  ///
+  /// (The identified message was not necessarily sent by the identified user;
+  /// it might have been sent by anyone in its conversation.)
+  final Map<int, int> latestMessagesByRecipient;
+
   final int selfUserId;
+
+  int compareByDms(User userA, User userB) {
+    final aLatestMessageId = latestMessagesByRecipient[userA.userId] ?? -1;
+    final bLatestMessageId = latestMessagesByRecipient[userB.userId] ?? -1;
+
+    return bLatestMessageId.compareTo(aLatestMessageId);
+  }
 
   /// Insert the key at the proper place in [sorted].
   ///
@@ -58,7 +90,7 @@ class RecentDmConversationsView extends ChangeNotifier {
     }
   }
 
-  /// Handle [MessageEvent], updating [map] and [sorted].
+  /// Handle [MessageEvent], updating [map], [sorted], and [latestMessagesByRecipient].
   ///
   /// Can take linear time in general.  That sounds inefficient...
   /// but it's what the webapp does, so must not be catastrophic. 🤷
@@ -116,6 +148,13 @@ class RecentDmConversationsView extends ChangeNotifier {
         sorted.removeAt(i); // linear time, ouch
         _insertSorted(key, message.id);
       }
+    }
+    for (final recipient in key.otherRecipientIds) {
+      latestMessagesByRecipient.update(
+        recipient,
+        (latestMessageId) => max(message.id, latestMessageId),
+        ifAbsent: () => message.id,
+      );
     }
     notifyListeners();
   }
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -1,0 +1,156 @@
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+
+import '../api/model/model.dart';
+
+class MessageIdTracker {
+  final List<int> _ids = [];
+
+  void add(int id) {
+    if (_ids.isEmpty) {
+      _ids.add(id);
+    } else {
+      int i = lowerBound(_ids, id);
+      if (i < _ids.length && _ids[i] == id) return; // the [id] already exists, so do not add it.
+      _ids.insert(i, id);
+    }
+  }
+
+  // TODO: remove
+
+  /// The maximum id in the tracker list.
+  ///
+  /// Returns -1 if the tracker list is empty.
+  int get maxId => _ids.isNotEmpty ? _ids.last : -1;
+
+  /// Getter for the tracked message IDs.
+  ///
+  /// This is intended for testing purposes only.
+  List<int> get idsForTesting => _ids;
+
+  @override
+  bool operator == (covariant MessageIdTracker other) {
+    if (identical(this, other)) return true;
+
+    return _ids.equals(other._ids);
+  }
+
+  @override
+  int get hashCode => Object.hashAll(_ids);
+}
+
+/// A data structure to keep track of stream and topic messages.
+///
+/// The owner should call [clear] in order to free resources.
+class RecentSenders {
+  // streamSenders[streamId][senderId] = IdTracker
+  final Map<int, Map<int, MessageIdTracker>> _streamSenders = {};
+
+  // topicSenders[streamId][topic][senderId] = IdTracker
+  final Map<int, Map<String, Map<int, MessageIdTracker>>> _topicSenders = {};
+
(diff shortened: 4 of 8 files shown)


sm-sayedi:
The code still may have some bugs, and it is unclean, but it would be great to know if I am on the right path for this problem!

And I will be more than happy to have feedback on my [draft proposal](https://chat.zulip.org/#narrow/stream/14-GSoC/topic/Sayed.20Mahmood.20-.20GSoC'24.20Proposal/near/1771583).

lib/model/store.dart
```diff
@@ -287,6 +287,12 @@ class PerAccountStore extends ChangeNotifier with StreamStore {
 
   final Map<int, User> users;
 
+  Map<int, Set<int>> get streamUserIds => {..._streamUserIds};
```
gnprice, line 290: This getter copies the whole Map. That's an O(N) operation (in the number of streams found in the map), so it shouldn't be a getter. See discussion here:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#getters-feel-faster-than-methods

In this case, though, the caller only needs a single entry from the map. So we can instead make it a method like (untested) `Iterable<int> knownSendersInStream(int streamId) => _knownSendersInStream[streamId];`

lib/model/store.dart
```diff
@@ -353,6 +359,22 @@ class PerAccountStore extends ChangeNotifier with StreamStore {
     super.dispose();
   }
 
+  void getMessageUsers(Message message) {
+    if (message is StreamMessage) {
+      Set<int> userIds = _streamUserIds[message.streamId] ?? {};
+      if (message.senderId == selfUserId) return;
+      userIds = {message.senderId}..addAll(userIds);
```
gnprice, line 366: This copies the Set. Better to mutate it instead — that has cost O(1), rather than O(N) in the size of the set.

lib/model/store.dart
```diff
@@ -287,6 +287,12 @@ class PerAccountStore extends ChangeNotifier with StreamStore {
 
   final Map<int, User> users;
 
+  Map<int, Set<int>> get streamUserIds => {..._streamUserIds};
+  final Map<int, Set<int>> _streamUserIds = {};
+
+  Set<int> get dmUserIds => {..._dmUserIds};
+  Set<int> _dmUserIds = {};
```
gnprice, line 294: These fields should get dartdoc explaining what their intended meaning is. That's helpful for the reader looking either at the code that initializes and maintains them, or at the code that consumes them.

In particular it separates those two problems — one can look only at some code that consumes the field and work out whether the code is correct, given the field's documented meaning, and separately one can look only at some code that maintains the field and work out whether that code is correct.

Without documentation of the intended meaning, one has to look at both sides together, guess what the meaning might be, then try to validate that guess against both sides, and potentially repeat with an adjusted guess.

lib/model/autocomplete.dart
```diff
@@ -246,7 +246,23 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+
+    switch (narrow) {
+      case AllMessagesNarrow():
+        userIds = {};
+      case StreamNarrow():
+        userIds = store.dmUserIds;
+        userIds.addAll(store.streamUserIds[(narrow as StreamNarrow).streamId] ?? {});
+      case TopicNarrow():
+        userIds = store.dmUserIds;
+        userIds.addAll(store.streamUserIds[(narrow as TopicNarrow).streamId] ?? {});
```
gnprice, line 259: Have you looked into the web client's logic for this? I believe its top priority when composing to a topic narrow is the participants in that individual topic, rather than the whole stream.

gnprice (commented):
Sure. Here's some high-level comments.

> And I will be more than happy to have feedback on my [draft proposal](https://chat.zulip.org/#narrow/stream/14-GSoC/topic/Sayed.20Mahmood.20-.20GSoC'24.20Proposal/near/1771583).

Ah indeed. :slightly_smiling_face: Replied there.

sm-sayedi:
Thanks for the review! Revision pushed, but this time changed the whole implementation to almost match it with the Web app.

lib/model/autocomplete.dart
```diff
@@ -244,11 +246,108 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+  }
+
   Future<List<MentionAutocompleteResult>?> _computeResults(MentionAutocompleteQuery query) async {
     final List<MentionAutocompleteResult> results = [];
-    final Iterable<User> users = store.users.values;
 
-    final iterator = users.iterator;
+    if (_sortedUsers == null) {
+      _sortUsers();
+    }
```
sm-sayedi, line 348: Two options here:
1. Sort all users one time at the start of the query, then filter them??
2. Filter all users, then sort them each time according to the query??

In both cases, filtering is done on the same full-length list,
but sorting is done in case:
  1) one time on a full-length list
  2) each time on a smaller-length filtered list

The web is using the second approach! Which one should we choose?

lib/model/store.dart
```diff
@@ -767,3 +773,145 @@ class UpdateMachine {
...
+  }
+
+  @override
+  String toString() {
+    return _ids.toString();
+  }
+}
+
+/// A data structure to keep track of stream and topic messages.
+class RecentSenders {
```
gnprice, line 814: For adding new data structures on PerAccountStore — anything more complicated than a single Map, anything that calls for its own class to encapsulate operations on it — we put them in separate files to help keep down the size of `store.dart`. (Particularly because `store.dart` has a lot of stuff with a different flavor to it, like `GlobalStore`, `LiveGlobalStore`, and `UpdateMachine`.)

For this, `recent_senders.dart` would be a good name.

gnprice (commented):
Thanks! Comments below.

sm-sayedi:
Thanks for the review! Pushed the changes. 

When you think it's ready, please let me know so I proceed with writing tests.

lib/model/autocomplete.dart
```diff
@@ -244,11 +246,116 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+    final bRecencyIndex = store.recentDmConversationsView.getRecencyIndex(userB.userId);
+
+    if (aRecencyIndex > bRecencyIndex) {
+      return -1;
+    } else if (bRecencyIndex > aRecencyIndex) {
+      return 1;
+    }
+
+    final aIsPartner = store.recentDmConversationsView.isPartner(userA.userId);
+    final bIsPartner = store.recentDmConversationsView.isPartner(userB.userId);
```
sm-sayedi, line 262: I wonder if this has any effect in our case! In the `recent_dm_conversations.dart` file, the `final Map<int, int> _dmRecencyData = {};` line in fact keeps track of all the DM users besides the latest message ID. So theoretically, if there is no recent message exchanged, it means we never had a conversation with that user!

lib/model/autocomplete.dart
```diff
@@ -244,11 +246,108 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+
+    if (aIsPartner && !bIsPartner) {
+      return -1;
+    } else if (bIsPartner && !aIsPartner) {
+      return 1;
+    }
+
+    return tertiaryCompare(userA, userB);
```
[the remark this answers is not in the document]
sm-sayedi: Matched it in the current revision!

gnprice (commented):
Thanks!

This strategy looks good; see comments below. I didn't cover all the points that I might comment on if doing a full review, but I tried to cover all the data structures and algorithms — I think all the data structures and algorithms here are basically what we want modulo these comments.

I'll be [on vacation](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Greg.20vacation/near/1768574) after today, so @chrisbobbe will review the next few revisions. There's a significant amount of complexity here, so he'll probably have plenty of comments to make about code style, dartdoc, tests, potentially also some of the data structures and algorithms, and likely also some correctness details I'd missed. He might merge the PR while I'm away, if he feels it's ready, and might flag some questions he'd like my input on. (Potentially both — we can always revise the code later.)

In any case I'll look forward to seeing when I'm back a version of this PR that's been polished up and is either merged, or nearly ready to merge :slightly_smiling_face:

chrisbobbe:
> I'll be [on vacation](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Greg.20vacation/near/1768574) after today, so @chrisbobbe will review the next few revisions.

Yep, happy to review! 🙂 I see this PR is currently marked as a draft, so I'll plan to review it once it's marked as ready for review.

sm-sayedi:
Thanks @gnprice for the review!

Revision pushed with the feedback addressed @chrisbobbe! Ready for your review now!

lib/model/recent_dm_conversations.dart
```diff
@@ -40,6 +42,24 @@ class RecentDmConversationsView extends ChangeNotifier {
...
 
+  // The ID of the latest messages exchanged with other users.
+  final Map<int, int> _dmRecencyData = {};
+
+  /// The ID of the latest message exchanged with this user.
+  ///
+  /// Returns -1 if there has been no DM message exchanged ever.
+  int getRecencyIndex(final int userId) => _dmRecencyData[userId] ?? -1;
+
+  void _computeDmRecencyData() {
```
sm-sayedi, line 53: I hope this version is better than before!

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+
+/// A data structure to keep track of message ids.
+class IdTracker {
+  final List<int> _ids = [];
+
+  void add(int id) {
+    if (_ids.isEmpty) {
+      _ids.add(id);
+    } else {
+      int i = _ids.possibleIndexOf(id);
```
sm-sayedi, line 59: Instead of using `List.indexOf` or `List.indexWhere` which uses the linear search, we could take advantage of the fact this list is sorted and use binary search to find the index.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+      if (i >= 0) { // the [id] already exists, so do not add it.
```
sm-sayedi, line 60: I preferred this way of checking for element's presence, instead of using `List.contains` which again uses the linear search.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+      int mid = (low + high) ~/ 2;
+      if (element == this[mid]) {
+        return mid;
+      } else if (element < this[mid]) {
+        high = mid - 1;
+      } else {
+        low = mid + 1;
+      }
+    }
+    return -low - 1;
```
sm-sayedi, line 47: Using `-low - 1` instead of just `-low` will differentiate between the case where the element is found at _**index 0**_ and the case where the element is not found, but should be placed at the very start of the list.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+  /// If [element] is not found, returns the possible index (1-based)
+  /// with a negative sign, where if the [element] is inserted,
+  /// the order of the list is maintained.
+  ///
+  /// ```dart
+  /// final somePrimes = [11, 13, 17, 23, 27, 29];
+  /// int index = somePrimes.possibleIndexOf(7); // -1
+  /// index = somePrimes.possibleIndexOf(19); // -4
+  /// ```
+  int possibleIndexOf(T element) {
```
chrisbobbe, line 33: lib/model/algorithms.dart would probably be a better place for a function like this. The functions there aren't written as `extension`s, but I'm not sure if that's deliberate and worth emulating for this one.

lib/model/recent_senders.dart
chrisbobbe: Let's explain this in a code comment. 🙂 It's exactly the kind of thing that any reader will be interested in, and most readers will just be looking at the code, without also reading through this PR thread.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+      } else {
+        low = mid + 1;
+      }
+    }
+    return -low - 1;
+  }
+}
+
+/// A data structure to keep track of message ids.
+class IdTracker {
```
chrisbobbe, line 52: I think `MessageIdTracker` would be a more helpful name for this, since a lot of the nearby code is dealing with user IDs. 🙂

With that name, the dartdoc as written would stand out as redundant/useless, though, wouldn't it…but that could be dealt with by removing it. But if there's anything else that's interesting about this class (other than that it's a data structure that keeps track of message IDs), that could be good to put in the dartdoc.

lib/model/autocomplete.dart
```diff
@@ -244,11 +246,102 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+      return -1;
+    } else if (bRecencyIndex > aRecencyIndex) {
+      return 1;
+    }
+
+    if (!userA.isBot && userB.isBot) {
+      return -1;
+    } else if (userA.isBot && !userB.isBot) {
+      return 1;
+    }
```
chrisbobbe, line 265: This part doesn't seem like it belongs in a method named `compareByDms`, as Greg pointed out above: https://github.com/zulip/zulip-flutter/pull/608#discussion_r1554283946

lib/model/autocomplete.dart
```diff
@@ -339,6 +432,12 @@ class AutocompleteDataCache {
   void invalidateUser(int userId) {
     _nameWordsByUser.remove(userId);
   }
+
+  final Map<String, String> _namesLowercased = {};
+
+  String nameLowercased(String name) {
+    return _namesLowercased[name] ??= name.toLowerCase();
+  }
```
chrisbobbe, line 440: Is there a reason not to key this by user ID, like `_nameWordsByUser` is? (If doing that, we should make sure `invalidateUser` acts appropriately on the lowercased-name data.)

chrisbobbe: How about `_lowercasedNameByUser` and `lowercasedNameForUser`, on the pattern of `_nameWordsByUser` and `nameWordsForUser`?

lib/model/message_list.dart
```diff
@@ -381,6 +381,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     for (final message in result.messages) {
       if (_messageVisible(message)) {
         _addMessage(message);
+        store.recentSenders.processStreamMessage(message);
```
chrisbobbe, line 384: Can you talk a bit about the performance considerations of this?

chrisbobbe (commented):
Thanks, @sm-sayedi! Comments below, and I see there's a test failing in CI. Does that failure reproduce when you run `tools/check` locally?

The new code will also need new tests.

sm-sayedi:
Thanks for the review! Revision pushed. Please have a look.

The tests are failing locally for me too. I was waiting for the overall approach of this feature to be steady to start writing tests for it. I will do it after now, but at the same time, I would appreciate your feedback on the recent revision!

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+        messageId: messageId);
+  }
+
+  // TODO: removeTopicMessage
+
+  int compareByRecency({
+    required User userA,
+    required User userB,
+    required int streamId,
+    // TODO(#493): may turn this into a non-nullable string.
```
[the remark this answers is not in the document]
sm-sayedi: Removed it this time! At first I thought, if the compose box is populated with a topic narrow, we may show the user results based on that topic narrow, even if we're in the stream view. But now looking at the web, it's not the case.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
+import 'dart:math' as math;
+
+import '../api/model/model.dart';
+
+extension Max<T extends num> on Iterable<T> {
+  /// Finds the maximum number in an [Iterable].
+  ///
+  /// Returns null if the [Iterable] is empty.
+  T? max() => isNotEmpty ? reduce(math.max) : null;
+}
```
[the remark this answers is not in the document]
sm-sayedi: Thanks for spotting this. It was used in the previous implementation, and I forgot to remove it.

lib/model/recent_dm_conversations.dart
```diff
@@ -40,6 +42,24 @@ class RecentDmConversationsView extends ChangeNotifier {
 
   final int selfUserId;
 
+  // The ID of the latest messages exchanged with other users.
+  final Map<int, int> _dmRecencyData = {};
```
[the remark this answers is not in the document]
sm-sayedi: Thank you! This is a well-descriptive dartdoc. Added it.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+    final topicMap = _topicSenders[streamId] ?? {};
+    final senderMap = topicMap[topic] ?? {};
+    final idTracker = senderMap[senderId] ?? IdTracker();
+    _topicSenders[streamId] = topicMap;
+    topicMap[topic] = senderMap;
+    senderMap[senderId] = idTracker;
+    idTracker.add(messageId);
+  }
+
+  void processStreamMessage(Message message) {
```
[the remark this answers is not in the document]
sm-sayedi: > Could fix by renaming to processMessage, or by narrowing the param's type to StreamMessage instead of Message.

Went with `processAsStreamMessage`. I thought when using `processMessage`, one would think that EACH message is processed; and when narrowing the param's type to `StreamMessage`, each caller should always make sure they're passing the correct message type by using a check, for example:
```dart
if (message is StreamMessage) {
  recentSenders.processStreamMessage(message);
}
```
I think it would be cleaner to add this check in the method itself and use a somewhat descriptive method name.
What do you think?

> Maybe we can also find a more specific verb than "process".

I think the verb "process" fits good in here as it extracts some data from the message and keeps track of it in a data structure. But I would appreciate a better verb from your side!

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+  @override
+  String toString() {
+    return _ids.toString();
+  }
+}
+
+/// A data structure to keep track of stream and topic messages.
+///
+/// The owner should call [clear] in order to free resources.
+class RecentSenders {
```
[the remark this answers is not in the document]
sm-sayedi: Changed them to private.

lib/model/recent_dm_conversations.dart
```diff
@@ -40,6 +42,24 @@ class RecentDmConversationsView extends ChangeNotifier {
...
+
+  void _computeDmRecencyData() {
+    for (final dmNarrow in sorted) {
+      for (final userId in dmNarrow.otherRecipientIds) {
+        if (!_dmRecencyData.containsKey(userId)) {
+          _dmRecencyData[userId] = map[dmNarrow]!;
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
sm-sayedi: As for my understanding of `sorted`, I think it is the right direction. If a user is in multiple conversations, we only pick the latest message ID of the first-occurring conversation in `sorted` as it is the most recent one, all of the upcoming conversations for that user in `sorted` are later than that.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,189 @@
...
+  int maxIdForStreamSender({required int streamId, required int senderId}) {
+    return _streamSenders[streamId]?[senderId]?.maxId ?? -1;
+  }
+
+  int maxIdForStreamTopicSender({
+    required int streamId,
+    required String topic,
+    required int senderId,
+  }) {
+    topic = topic.toLowerCase();
```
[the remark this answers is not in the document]
sm-sayedi: Removed those lines. I thought topics were case-insensitive. 🙂

lib/model/recent_senders.dart
sm-sayedi: Moved it into there. Still went with the `extension` approach as it feels relatively smooth when using it. But if you think it's inconsistent with its surrounding code, please let me know and I will change it to a top-level function.

lib/model/autocomplete.dart
sm-sayedi: > Is there a reason not to key this by user ID, like `_nameWordsByUser` is?

Two reasons: a) to not pass the whole `User` object while we can do the work with just one of its properties. b) there can be multiple users with the same name so we can save the space a little bit by having just one key-value pair for all those user names.
However, this will bring some complexity when adding the appropriate logic in `invalidateUser`. Let's assume we somehow pass `fullName` to 'invalidateUser' besides `userId` to remove the lowercase name. It will remove the user name for all the users that share the same name. 
So I am not sure whether to key it by `userId` or `fullName`?

lib/model/message_list.dart
sm-sayedi: I am not hundred percent sure if it's an expensive task or not. I think it does relatively simple task of adding `IdTracker` to a map two times.

lib/model/recent_dm_conversations.dart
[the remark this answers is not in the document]
gnprice: Yeah, I believe this code's behavior is the right way around.

But this is a perfect example of something that would be good to cover with a test :slightly_smiling_face: — there should be a test where it's easy for the reader of the test to see that the test would fail if the implementation had this bug.

lib/model/autocomplete.dart
gnprice: Yeah, I think best to key it by user ID, the same way as `_nameWordsByUser`.

It's uncommon for users to have the same name, so there's minimal space savings to be had from that, which makes it not worth the complexity of worrying about interactions there.

Relatedly: in `nameWordsForUser` we're calling `toLowerCase` on `user.fullName`, just like this does. So let's have that step share this same cache.

gnprice:
Thanks @chrisbobbe for the round of review, and @sm-sayedi for the revision!

I posted a few comments above, mainly replying on some of the comment threads from Chris's review.

The main next step here will be to write some tests, and to fix the existing test that's failing. I'll have more detailed comments on the code too, but the overall structure is definitely close enough that it'll be helpful to start writing tests. And when a test is failing, even in a draft PR it's generally never too soon to either fix it, or at least diagnose a general idea of why it's failing :slightly_smiling_face:

lib/model/recent_senders.dart
sm-sayedi: For this specific case, the `List.lowerBound` method provides a solution to some extent as it returns the `index` where the `element` should be, but it does not give any clue if the element is in the list or not. But this method comes with that additional info as stated in its docs.

lib/model/algorithms.dart
```diff
@@ -116,3 +116,46 @@ QueueList<int> setUnion(Iterable<int> xs, Iterable<int> ys) {
...
+  ///
+  /// Note: Using a negative 1-based index will differentiate between the case where
+  /// the element is found at index 0 and the case where the element is not
+  /// found, but should be placed at the very start of the list (index -1).
+  int possibleIndexOf(T element) {
+    int low = 0;
+    int high = length - 1;
+
+    while (high >= low) {
+      int mid = (low + high) ~/ 2;
```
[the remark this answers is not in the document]
sm-sayedi: Thank you. Just Learned an interesting fact.

lib/model/autocomplete.dart
```diff
@@ -337,13 +429,25 @@ class MentionAutocompleteQuery {
 }
 
 class AutocompleteDataCache {
+  final Map<int, String> _namesByUser = {};
+
+  /// The lowercase `fullName` of [user].
+  String nameForUser(User user) {
```
gnprice, line 435:
```suggestion
  /// The lowercase `fullName` of [user].
  String normalizedNameForUser(User user) {
```

It's not quite the name, but rather a version of the name that's been processed into a normalized form useful for autocomplete. So we can have the method's name distance itself a bit from saying it's simply the user's name.

lib/model/autocomplete.dart
```diff
@@ -337,13 +429,25 @@ class MentionAutocompleteQuery {
...
+  }
+
   final Map<int, List<String>> _nameWordsByUser = {};
 
   List<String> nameWordsForUser(User user) {
-    return _nameWordsByUser[user.userId] ??= user.fullName.toLowerCase().split(' ');
+    final nameWords = _nameWordsByUser[user.userId];
+    if (nameWords != null) return nameWords;
+
+    final name = _namesByUser[user.userId] ??= user.fullName.toLowerCase();
```
gnprice, line 445: Let's avoid duplicating this logic:
```suggestion
    final name = nameForUser(user);
```

Particularly as we might in future put more logic there, like normalizing out diacritics.

gnprice: Then I think this method's body can be simplified back down into the form `return cache[key] ??= …initializer…;`.

lib/model/store.dart
```diff
@@ -171,6 +172,26 @@ abstract class GlobalStore extends ChangeNotifier {
   String toString() => '${objectRuntimeType(this, 'GlobalStore')}#${shortHash(this)}';
 }
 
+/// A wrapper class for `Map<K, V>`.
+///
+/// This can be listened to for any modifications done to `Map<K, V>`.
+class MapNotifier<K, V> extends ValueNotifier<Map<K, V>> {
+  MapNotifier(super._value);
```
sm-sayedi, line 179: The reason for this change is that `users` used in `model/autocomplete.dart` are now sorted by relevance before being filtered. Before this PR, an `iterator` was made from `users.values` which kept the relationship with `users` so any modification to `users` would be reflected in the iterator too, for example, `ConcurrentModificationError` being thrown. For this change, `users` should be changed to a list using `users.values.toList` in order to be sorted, and this breaks the relationship with `users`. By using this wrapper, we can listen to modifications to the original `users` and act upon that.

lib/model/autocomplete.dart
```diff
@@ -251,11 +253,95 @@ class MentionAutocompleteView extends ChangeNotifier {
...
     final List<MentionAutocompleteResult> results = [];
-    final Iterable<User> users = store.users.values;
 
-    final iterator = users.iterator;
+    if (_sortedUsers == null) {
+      _sortUsers();
+    }
+
+    final iterator = _sortedUsers!.iterator;
+    store.users.addListener(_usersModified);
```
gnprice, line 344: Hmm interesting. Can you elaborate on what this is about?

This would also be good to pull out as its own prep commit — definitely the `store.dart` changes to support it, and perhaps this `addListener` and the associated bits of logic in this file would also go in that commit.

One reason a separate prep commit is helpful is that its commit message provides a natural venue to give the explanation about why it's useful. :slightly_smiling_face:

gnprice: … ah I see, you made a comment just now about this. Reading now…

gnprice (commented):
(Submitting this partial review to resolve a race)

lib/model/store.dart
gnprice: I see, that makes sense.

Definitely belongs in its own prep commit, as discussed in https://github.com/zulip/zulip-flutter/pull/608#discussion_r1592958809. :slightly_smiling_face: It's fine that as of that commit, the new functionality won't really be *needed*; the commit message can explain why we're about to start needing it.

test/model/recent_dm_conversations_test.dart
```diff
@@ -35,7 +36,8 @@ void main() {
             key([]):     200,
             key([1]):    100,
           })
-          ..sorted.deepEquals([key([1, 2]), key([]), key([1])]);
+          ..sorted.deepEquals([key([1, 2]), key([]), key([1])])
+          ..latestMessagesByRecipientForTesting.deepEquals({1: 300, 2: 300});
```
gnprice, line 40: It's good to update the existing test cases for this code to check the new data structure.

But are there any new test cases that become useful based on thinking through edge cases of the new data structure? I'd guess there would be. It'd then be good to add those too.

gnprice (commented):
OK, here's some more review comments before I move on to other things. Thanks for the revision!

A couple of high-level comments:
* Let's break this up into a series of commits: adding the recent-senders data, adding the new data structure to recent-DMs, adding a helper like `possibleIndexIn` or `MapNotifier` (though those two in particular I think we'll just leave out), and so on.

  Then adding the changes to autocomplete behavior can come at the end. That can also be a series of commits, adding the several different new criteria it uses.

* In fact let's also break it up into several PRs, because that will help us be able to merge parts of this sooner. Let's start with recent DMs, since that's the simpler of the two new data structures. So a PR that leaves out recent senders, and can leave out is-bot and name too, but has a series of commits that first add the new recent-DMs data (after doing any prep needed for that), then do any needed prep for the concept of `_sortedUsers`, then starts comparing by recent DMs.

  Then the other sort criteria can be in followup PRs.

lib/model/message_list.dart
```diff
@@ -381,6 +381,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     for (final message in result.messages) {
       if (_messageVisible(message)) {
         _addMessage(message);
+        store.recentSenders.handleMessage(message);
```
gnprice, line 384: The ideal version of this would go on top of a central message store, #648, so that `MessageListView` wouldn't have to know individually about all the miscellaneous data structures that care about a summary of messages.

This PR doesn't need to block on that one; for now, it's fine to add these calls in a more ad-hoc way like this.

The current version of this PR doesn't call this `RecentSenders.handleMessage` method in enough cases, though. Compare which messages get passed to `reconcileMessages` in #648.

sm-sayedi:
Thanks @gnprice for the detailed review! Pushed the revision with some comments here and there. PTAL. 

**Note:** I divided the giant commit into multiple commits instead of opening several PRs for related features to avoid becoming blocked by initial PRs as the latter would rely on them. Again, if you prefer the multiple PRs approach, please let me know, and I will do that.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,174 @@
...
+  void clear() {
+    _streamSenders.clear();
+    _topicSenders.clear();
+  }
+
+  int _latestMessageIdOfSenderInStream({required int streamId, required int senderId}) {
+    return _streamSenders[streamId]?[senderId]?.maxId ?? -1;
+  }
+
+  int _latestMessageIdOfSenderInTopic({
```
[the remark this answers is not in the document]
sm-sayedi: Haven't touched this part yet. Should I move `compareByRecency` to its caller and thus change these two to public? Right now they're not used outside of this class.

lib/model/store.dart
sm-sayedi: Looking at the code, adding, removing, or updating a user will start the search again. But I added this specific `notifier` to pass a failing test in `autocomplete_test.dart` ("MentionAutocompleteView mutating store.users while in progress causes retry").

test/model/recent_dm_conversations_test.dart
sm-sayedi: I thought a lot about it, but frankly I couldn't come up with other edge cases!🙂 I would appreciate any clue from your side!

lib/model/autocomplete.dart
```diff
@@ -145,6 +145,18 @@ class AutocompleteViewManager {
     autocompleteDataCache.invalidateUser(event.userId);
   }
 
+  void handleMessageEvent(MessageEvent event) {
```
sm-sayedi, line 148: I am not sure if it is necessary to pass the `event` parameter as it is not used for now. I just followed the code for `handleRealmUserAddEvent(RealmUserAddEvent event)` method.

lib/model/autocomplete.dart
```diff
@@ -145,6 +145,18 @@ class AutocompleteViewManager {
...
+    for (final view in _mentionAutocompleteViews) {
+      view.refreshStaleUserResults();
+    }
```
sm-sayedi, line 151: Wouldn't it be good to extract this bit to a (private) method as it is used in several other neighboring methods?! A name like `refreshAutocompleteResults` may be a proper name for that method as then we can replace it with all the method calls that just have the body of this method in their bodies.

lib/model/message_list.dart
sm-sayedi, line 384: I am not sure whether to add it inside the if block or outside!!

lib/model/autocomplete.dart
gnprice: Yeah, following that pattern is good. This is also the pattern you'll find at the call site in `PerAccountStore.handleEvent`.

lib/model/autocomplete.dart
gnprice: Yeah, a prep commit to factor this out would be good. A good name would be `_refreshStaleUserResults`, to keep it more specific. (Right now the only kind of autocomplete we have is to return users, but we'll also have autocomplete for streams and topics and other things, and those won't necessarily need to get invalidated when the list of users or their ranking changes.)

lib/model/message_list.dart
gnprice: Good question.

What does the dartdoc on that `_messageVisible` method say?

Based on that, what would be the consequence of having that condition control whether to make this call? And what would be the consequence of ignoring that condition when making this call?

gnprice (commented):
Thanks for the revision! Some comments above. Here, replies to your inline questions.

gnprice (commented):
OK, and here's the rest of a review. See also the couple of groups of comments just above.

Splitting the commit up into 10 smaller commits was very helpful, thanks!

> **Note:** I divided the giant commit into multiple commits instead of opening several PRs for related features to avoid becoming blocked by initial PRs as the latter would rely on them. Again, if you prefer the multiple PRs approach, please let me know, and I will do that.

Yeah, splitting into several PRs is helpful for two reasons:
* GitHub makes things increasingly unwieldy with even a moderately-long PR thread. This one is currently 126 comments long (according to https://github.com/zulip/zulip-flutter/pulls), so it's already past the point where when I open the PR page, the first thing I have to do in order to make it usable is Ctrl+F "hidden", click "Load more…", Ctrl+G, click again, Ctrl+G in a loop until there are no matches. The longer the thread gets, the more iterations of that loop are needed.

  The "hidden" thing is a bad and dysfunctional choice by GitHub that I can only explain in terms of messed-up internal incentives in their organization. But it's the way it works, so we have to adapt. (And even without that, a really long thread would still get unwieldy.)

* I think we'll be ready to merge the "recent DMs" part before the "recent senders" part. So focusing one PR on the "recent DMs" part, and the other pieces needed for that (like `_sortedUsers` and the logic around it), makes it possible to get that in sooner.

The later PR will block on the earlier PR before it can be merged, but that's not really a change from what would happen with a single PR: the later changes would still have to wait for the earlier changes to be ready before they can be merged.

One workflow bit to make explicit: GitHub doesn't know anything about the relationship between PRs, but that's fine. Just stack the second branch atop the first one in Git, and send it as another PR; the first N commits in the second PR will be the commits in the first PR. Be sure to mention clearly in the second PR's description that it's based atop the first — that helps reviewers focus on the commits that are new in the second PR.

lib/model/recent_senders.dart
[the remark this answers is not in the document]
gnprice: Oh, and missed this question in my sweep just above. Yeah, let's move that function to near where it's called (and/or inline it into its caller), and make these two public.

That'll be part of the second PR that gets stacked atop this one, since this is in the recent-senders code.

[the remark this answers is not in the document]
gnprice: (Adding another comment now, as part of "a review" this time, just to make this show up in the main PR page's chronological list.)

gnprice:
See also a reply on one more question: https://github.com/zulip/zulip-flutter/pull/608#discussion_r1600538137

gnprice:
> Thanks for the review! Pushed the recent DMs revision here although I wasn't sure to create a new PR for this too.

Thanks!

Yeah, now that I think about it: please close this PR and open a new one with the same contents. In particular that will be helpful because the last round of review above covers changes that will be part of the several different PRs in the series; by letting this PR thread end here, that review will remain accessible when looking at each of those PRs. If we continued in this same thread as the PR for the first part of the changes, then that review would get pushed up into history and potentially into the "hidden" area.

sm-sayedi:
I have opened #693 as the first part of this PR. Please have a look.

gnprice:
Closing this in favor of its successor PRs #693 and #692. Thanks @sm-sayedi for all your work on this so far!

Issue #609, opened 2024-04-02
Show user more info on error in web auth


gnprice:
This is a followup to:
* #36 

In the version we merged as #600, if anything goes wrong, the error dialog is always the same generic message. But this flow is inherently a bit finicky — it involves sending the user out to a browser, then receiving a callback by having the browser redirect to a special URL — so it's likely things will sometimes go wrong. And indeed a previous draft had a couple of different things go wrong (https://github.com/zulip/zulip-flutter/pull/600#issuecomment-2030685985). I debugged those pretty quickly because I was running a debug build connected to my desktop and watching a developer console… but when beta users run into issues, we won't have those advantages. It'll therefore really help us out if the error dialog shows more detail about the error, so that a screenshot from the user can start narrowing down where the problem is.

As a first step, when we get a `PlatformException` we should display its `message` field.

PR #610, opened 2024-04-02
login: If web-auth error is PlatformException, show error.message to user

chrisbobbe:
Fixes: #609


gnprice:
Thanks! Looks good — merging.

PR #621, opened 2024-04-05
Upgrade Flutter and all packages to latest

### LINKED ISSUE #612: "Missing API declaration" warning from Apple
After I uploaded today's v0.0.13 release to TestFlight for our iOS beta users, we got an email from Apple with the following warning:

> Although submission for TestFlight review was successful, you may want to correct the following issues in your next submission for TestFlight review. Once you've corrected the issues, upload a new binary to App Store Connect.
>
> **ITMS-91053: Missing API declaration** - Your app’s code in the “Runner” file references one or more APIs that require reasons, including the following API categories: NSPrivacyAccessedAPICategoryUserDefaults. While no action is required at this time, starting May 1, 2024, when you upload a new app or app update, you must include a NSPrivacyAccessedAPITypes array in your app’s privacy manifest to provide approved reasons for these APIs used by your app’s code. For more details about this policy, including a list of required reason APIs and approved reasons for usage, visit: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api.
>
> Apple Developer Relations

So we'll need to resolve that in the next few weeks, by 2024-05-01, in order to continue publishing betas for iOS whenever we like.

Whatever "one or more APIs" they're referring to must be something in one of our dependencies, because we have zero Swift or Objective-C code of our own so far. Following links a bit: the relevant section is [here](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401), and the only API in that category is [`UserDefaults`](https://developer.apple.com/documentation/foundation/userdefaults). I don't know offhand which of our dependencies might use that; it might be in code we never invoke.

It seems like Apple has just recently started sending these warnings. We didn't get one for the v0.0.12 release, just three weeks ago, and we don't have any new dependencies since then. Threads like https://github.com/flutter/flutter/issues/145269 also suggest that many people just started getting these circa 2024-03-16, a little over two weeks ago.

To handle this:

* The first step is to upgrade all our dependencies. That might resolve the issue — there's a way for dependencies to provide these privacy manifests (to be aggregated into the app's manifest), and it sounds like a lot of plugins have just recently added them. Which makes sense given that these warnings just started going out to the developers using the plugins.

* If that doesn't resolve it: the plugins may nevertheless have the privacy manifests, as they don't always get automatically aggregated. Detailed steps to take are at https://github.com/flutter/flutter/issues/131940#issuecomment-2006965339 ; see also update at https://github.com/flutter/flutter/issues/131940#issuecomment-2016573862 .

* If *that* doesn't resolve it, that'll mean some plugin is missing a privacy manifest it needs. At that point it's a debugging problem. The general approach needed is described [here](https://github.com/flutter/flutter/issues/145269#issuecomment-2026343247); some examples inspecting binary dependencies with `nm | grep` are [here](https://github.com/flutter/flutter/issues/145269#issuecomment-2029855632) and in the comments after it.

Useful threads in the Flutter tracker include:
* https://github.com/flutter/flutter/issues/131940
* https://github.com/flutter/flutter/issues/145269

chrisbobbe:
Done because it's the first step in our plan for #612, the "Missing API declaration" warning from Apple.

Changelogs for deps that are major-version upgraded:
  https://pub.dev/packages/device_info_plus/changelog
  https://pub.dev/packages/file_picker/changelog
  https://pub.dev/packages/flutter_local_notifications/changelog
  https://pub.dev/packages/package_info_plus/changelog
  https://pub.dev/packages/share_plus/changelog
  https://pub.dev/packages/pigeon/changelog

Manual tests done:

- My iPhone 13 Pro running iOS 17.3.1:
  - App starts up; pages load (including Inbox and message list)
- Office Android device (Samsung Galaxy S9 running Android 9):
  - App starts up; pages load (including Inbox and message list)
  - DM notification received; navigates to conversation when tapped
  - `device_info_plus` exercised and seems to work correctly ([discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20Java.20versions.3F/near/1775374))

Related: #612


gnprice:
Thanks!

All these changes look good. I won't have time to read those changelogs before heading out on vacation; but if you've read them and nothing jumps out at you, and given that it passes your manual testing, please go ahead and merge. I looked through all these changes other than reading the changelogs.

    All of these changelogs except `pigeon` mention adding "privacy
    info" or "privacy manifest" in the version we're taking here --
    exactly the kind of thing we want to see, with #612 in mind.

Great!

    I would
    guess that `pigeon` doesn't need one, as it's mainly just a code
    generator tool.

Yeah. And more concretely: it's a dev dependency only, so it's not part of the actual app build.

gnprice:
Ah, and CI is failing — looks like you just need to update the Pigeon-generated output.

(And then I guess rerun the manual testing, unless the changes are trivial.)

gnprice:
Huh odd.

I guess one thing you can do to debug is to edit `tools/check` to print the full diff in that case, rather than the status.

gnprice:
Alternatively, try making a meaningless edit to the Pigeon input file and rerunning Pigeon — wonder if this is a caching issue, like the one we've seen with other code generators (using build_runner).

chrisbobbe:
> I guess one thing you can do to debug is to edit `tools/check` to print the full diff in that case, rather than the status.

Aha, this seems like a useful clue from that output:

```diff
-// Autogenerated from Pigeon (v17.2.0), do not edit directly.
+// Autogenerated from Pigeon (v18.0.0), do not edit directly.
```

chrisbobbe:
The issue seems to be that the old Pigeon 17 is running locally despite the upgrade in the headline commit taking it to Pigeon 18. The command `tools/check pigeon --all-files` kept succeeding.

Then I did something different:

```
$ dart run pigeon
Resolving dependencies in `/Users/chrisbobbe/dev/zulip-flutter`... 
Downloading packages... 
Got dependencies in `/Users/chrisbobbe/dev/zulip-flutter`.
Building package executable... (5.2s)
Built pigeon:pigeon.
Pigeon is a tool for generating type-safe communication code between Flutter
and the host platform.

usage: pigeon --input <pigeon path> --dart_out <dart path> [option]*
[…]
```

and thereafter, `tools/check pigeon --all-files` is failing, which is the expected behavior from Pigeon 18. I think the `Building package executable...` step is what got it working. Is there a nice way to make sure that gets run automatically when our Pigeon version changes?

chrisbobbe:
Anyway, revision pushed; `tools/check pigeon` should be happy now.

chrisbobbe:
And I've just redone manual tests for Android notifications. (The code did look NFC to me, but just for good measure.)

> All these changes look good. I won't have time to read those changelogs before heading out on vacation; but if you've read them and nothing jumps out at you, and given that it passes your manual testing, please go ahead and merge.

Cool; yep. Merging.

chrisbobbe:
Filed #622 for the Pigeon version issue.

PR #423, opened 2023-11-29
Implement stream and topic muting

gnprice:
Fixes: #346 
Fixes: #395 

This series also contains some refactoring of PerAccountStore, creating a new class StreamStore and new file `lib/model/stream.dart` as a home for the section of the data that's about streams and topics. There's probably a couple of other swathes of data that could benefit from a similar treatment.

Also, in order to let the Unreads instance have access to StreamStore:

---

store [nfc]: The event-poll loop has-a store, not is-a store

This seems to make more sense conceptually.

For its most immediate practical effect, this means
UpdateMachine.fromInitialSnapshot doesn't care whether
PerAccountStore.fromInitialSnapshot is a generative or a
factory constructor, which will give us the flexibility to
make it the latter.

This change will also probably be helpful for making other
decouplings of the polling and registerNotificationToken logic
from the data structures of PerAccountStore.

---

store [nfc]: Make fromInitialSnapshot a factory constructor

This will let us first construct the StreamStore, as a local of
the constructor, and then pass that to the Unreads constructor.

PR #624, opened 2024-04-17
unreads [nfc]: Update dartdoc following stream/topic-muting logic

chrisbobbe:
In #423 "Implement stream and topic muting", the Unreads class did start doing some filtering based in stream/topic muting, specifically in these methods:
- countInStream
- countInAllMessagesNarrow

So, update the dartdoc so it doesn't claim that this filtering is an entirely separate concern.


gnprice:
Sure, looks good; merging.

Issue #310, opened 2023-09-22
Autocomplete for topic to send to


gnprice:
We should offer autocompletion in the topic input of the compose box. (When that topic input exists, i.e. for interleaved narrows.)

This is a step on the way to:
* #149

and an element we'll almost surely want in any UX solution we choose there.

Not to be confused with autocomplete for mentioning a topic inside the content of a message:
* #124

Khader-1:
I've reviewed this task, and I believe it's one step up compared to my previous contributions. I am confident I can tackle it effectively.
To address this, I propose the following steps:
- Implementing a new API call to retrieve data from https://chat.zulip.org/json/users/me/:stream_id/topics
- Modifying the widget structure to incorporate autocomplete functionality, similar to the message compose input
- Developing a view model and a controller similar to those used for the compose input

I am keen to take on this task and intend to start by creating an initial prototype before requesting formal assignment.

gnprice:
@Khader-1 Sounds good! Please go ahead.

timabbott:
This data structure is likely a thing that will end up being used for other things. `stream_topic_history.ts` in the Zulip web app may be a useful reference. There are some messy/subtle problems with how it tries to live-update, detailed in https://github.com/zulip/zulip/pull/29015, so a lot of the current complexity is wrong. But specifically the logic in `stream_topic_history_util.ts` might be a helpful reference for how the web app does fetch-on-demand these data, and caches it.

(It can take like 0.5s to get a response in a stream with thousands of topics).

gnprice:
Thanks Tim!

For a solution for this issue, I think I'd be happy with a version where the list of topics for a given stream is only cached for the lifetime of a given autocomplete interaction — that is, for the lifetime of a given topic input field.

In particular that avoids most of the complexity of maintaining the data structure, because it's OK if it doesn't stay accurately in sync in subtle edge cases (and really it's 95% OK if it doesn't get updated at all), because not many changes will happen during its lifetime.

The main cost is that 0.5s lag to get the list of topics. But we'd have that anyway the first time a user goes to compose in a given stream, so I'm not sure there's a huge difference between that and every time.

Hmm, though I guess one benefit of keeping it around longer is that we could be more eager about fetching it — like when you *view* a given stream narrow, rather than when you start composing. Then you might rarely actually see the 0.5s latency.

---

[...]

Issue #228, opened 2023-07-19
autocomplete: Sort user-mention autocomplete results


chrisbobbe:
Perhaps by name, to start with, but really:

https://github.com/zulip/zulip-flutter/pull/207#pullrequestreview-1520406847

> (Or better yet something more relevant than name: some kind of relevance to search plus relevance to the context, like people subscribed to the stream or that have talked recently in the narrow. But sorting by user ID seems like it's just going to feel weird and arbitrary to people.)

sm-sayedi:
I would like to work on this one!

So far I've made it sorted by the name.

<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
    </tr>
    <tr>
        <td><img src="https://github.com/zulip/zulip-flutter/assets/59946442/476ab511-654f-4458-8297-589a4f6ae43d" width="300"/></td>
        <td><img src="https://github.com/zulip/zulip-flutter/assets/59946442/538672fd-8c90-4663-a742-6e492d2a429c" width="300"/></td>
    </tr>
</table>

gnprice:
@sm-sayedi Sure, please do — assigning to you.

I think a good behavior to shoot for here would be to sort it the same way Zulip web does. There might be things that can be improved there, but if so then they should probably be changed there too. Two particular things I believe it uses that would be helpful:
* people who've recently participated in the conversation you're looking at get shown sooner
* people who you've had DM conversations with get shown sooner

In general we keep less message history in the mobile client than the web client, because of the different context, so that might affect what information we have for implementing some of the same features.

But if there are other kinds of data that would help implement the same features as web, and that we don't currently track in the store but are part of the [/api/register-queue](https://zulip.com/api/register-queue) response, then we can take this as a prompt to add those to our API bindings and start tracking them in the store, as part of implementing this issue.

And as usual, if anything seems uncertain or you have any questions as you're going about that, please ask in `#mobile-team` and we'll be glad to discuss.

chrisbobbe:
@sm-sayedi with #828 merged, what remains to be done here? 🙂

sm-sayedi:
As mentioned in #608 description:
> 3. Human vs. Bot users.
> 4. Alphabetical order.

So one other PR is needed to close this issue. 😃

gnprice:
Great!

[...]

PR #627, opened 2024-04-23
autocomplete: Introduce topic input autocomplete

Khader-1:
Fixes: #310

https://github.com/zulip/zulip-flutter/assets/63330019/d47b388d-f879-4ee5-b290-62c0342f61d3

### DIFF
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -3,6 +3,7 @@ import 'package:flutter/services.dart';
 
 import '../api/model/events.dart';
 import '../api/model/model.dart';
+import '../api/route/channels.dart';
 import '../widgets/compose_box.dart';
 import 'narrow.dart';
 import 'store.dart';
@@ -40,6 +41,15 @@ extension ComposeContentAutocomplete on ComposeContentController {
         textEditingValue: value);
     }
     return null;
+  }
+}
+
+extension ComposeTopicAutocomplete on ComposeTopicController {
+  AutocompleteIntent<TopicAutocompleteQuery>? autocompleteIntent() {
+    return AutocompleteIntent(
+      syntaxStart: 0,
+      query: TopicAutocompleteQuery(value.text),
+      textEditingValue: value);
   }
 }
 
@@ -112,6 +122,7 @@ class AutocompleteIntent<QueryT extends AutocompleteQuery> {
 /// On reassemble, call [reassemble].
 class AutocompleteViewManager {
   final Set<MentionAutocompleteView> _mentionAutocompleteViews = {};
+  final Set<TopicAutocompleteView> _topicAutocompleteViews = {};
 
   AutocompleteDataCache autocompleteDataCache = AutocompleteDataCache();
 
@@ -125,6 +136,16 @@ class AutocompleteViewManager {
     assert(removed);
   }
 
+  void registerTopicAutocomplete(TopicAutocompleteView view) {
+    final added = _topicAutocompleteViews.add(view);
+    assert(added);
+  }
+
+  void unregisterTopicAutocomplete(TopicAutocompleteView view) {
+    final removed = _topicAutocompleteViews.remove(view);
+    assert(removed);
+  }
+
   void handleRealmUserRemoveEvent(RealmUserRemoveEvent event) {
     autocompleteDataCache.invalidateUser(event.userId);
   }
@@ -135,10 +156,13 @@ class AutocompleteViewManager {
 
   /// Called when the app is reassembled during debugging, e.g. for hot reload.
   ///
-  /// Calls [MentionAutocompleteView.reassemble] for all that are registered.
+  /// Calls [AutocompleteView.reassemble] for all that are registered.
   ///
   void reassemble() {
     for (final view in _mentionAutocompleteViews) {
+      view.reassemble();
+    }
+    for (final view in _topicAutocompleteViews) {
       view.reassemble();
     }
   }
@@ -531,3 +555,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
 // TODO(#233): // class UserGroupMentionAutocompleteResult extends MentionAutocompleteResult {
 
 // TODO(#234): // class WildcardMentionAutocompleteResult extends MentionAutocompleteResult {
+
+class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, TopicAutocompleteResult, String> {
+  TopicAutocompleteView._({required super.store, required this.streamId});
+
+  factory TopicAutocompleteView.init({required PerAccountStore store, required int streamId}) {
+    final view = TopicAutocompleteView._(store: store, streamId: streamId);
+    store.autocompleteViewManager.registerTopicAutocomplete(view);
+    view._fetch();
+    return view;
+  }
+
+  final int streamId;
+  Iterable<String> _topics = [];
+  bool _isFetching = false;
+
+  /// Fetches topics of the current stream narrow, expected to fetch
+  /// only once per lifecycle.
+  ///
+  /// Starts fetching once the stream narrow is active, then when results
+  /// are fetched it restarts search to refresh UI showing the newly
+  /// fetched topics.
+  Future<void> _fetch() async {
+     assert(!_isFetching);
+    _isFetching = true;
+    final result = await getStreamTopics(store.connection, streamId: streamId);
+    _topics = result.topics.map((e) => e.name);
+    _isFetching = false;
+    if (_query != null) _startSearch(_query!);
+  }
+
+  @override
+  Iterable<String> getSortedItemsToTest(TopicAutocompleteQuery query) => _topics;
+
+  @override
+  TopicAutocompleteResult? testItem(TopicAutocompleteQuery query, String item) {
+    if (query.testTopic(item)) {
+      return TopicAutocompleteResult(topic: item);
+    }
+    return null;
+  }
+
+  @override
+  void dispose() {
+    store.autocompleteViewManager.unregisterTopicAutocomplete(this);
+    super.dispose();
+  }
+}
+
+class TopicAutocompleteQuery extends AutocompleteQuery {
+  TopicAutocompleteQuery(super.raw);
+
+  bool testTopic(String topic) {
+    // TODO(#881): Sort by match relevance, like web does.
+    return topic != raw && topic.toLowerCase().contains(raw.toLowerCase());
+  }
+
+  @override
+  String toString() {
+    return '${objectRuntimeType(this, 'TopicAutocompleteQuery')}(raw: $raw)';
+  }
+
+  @override
+  bool operator ==(Object other) {
+    return other is TopicAutocompleteQuery && other.raw == raw;
+  }
+
+  @override
+  int get hashCode => Object.hash('TopicAutocompleteQuery', raw);
+}
+
+class TopicAutocompleteResult extends AutocompleteResult {
+  final String topic;
+
+  TopicAutocompleteResult({required this.topic});
+}
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -7,7 +7,7 @@ import '../widgets/compose_box.dart';
 import 'narrow.dart';
 import 'store.dart';
 
-extension Autocomplete on ComposeContentController {
+extension ComposeContentAutocomplete on ComposeContentController {
   AutocompleteIntent? autocompleteIntent() {
     if (!selection.isValid || !selection.isNormalized) {
       // We don't require [isCollapsed] to be true because we've seen that
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -8,7 +8,7 @@ import 'narrow.dart';
 import 'store.dart';
 
 extension ComposeContentAutocomplete on ComposeContentController {
-  AutocompleteIntent? autocompleteIntent() {
+  AutocompleteIntent<MentionAutocompleteQuery>? autocompleteIntent() {
     if (!selection.isValid || !selection.isNormalized) {
       // We don't require [isCollapsed] to be true because we've seen that
       // autocorrect and even backspace involve programmatically expanding the
@@ -68,8 +68,8 @@ final RegExp mentionAutocompleteMarkerRegex = (() {
     unicode: true);
 })();
 
-/// The content controller's recognition that the user might want autocomplete UI.
-class AutocompleteIntent {
+/// The text controller's recognition that the user might want autocomplete UI.
+class AutocompleteIntent<QueryT extends AutocompleteQuery> {
   AutocompleteIntent({
     required this.syntaxStart,
     required this.query,
@@ -91,7 +91,7 @@ class AutocompleteIntent {
   // that use a custom/subclassed [TextEditingValue], so that's not convenient.
   final int syntaxStart;
 
-  final MentionAutocompleteQuery query; // TODO other autocomplete query types
+  final QueryT query;
 
   /// The [TextEditingValue] whose text [syntaxStart] refers to.
   final TextEditingValue textEditingValue;
@@ -151,21 +151,90 @@ class AutocompleteViewManager {
   // void dispose() { … }
 }
 
-/// A view-model for a mention-autocomplete interaction.
+/// A view-model for an autocomplete interaction.
 ///
 /// The owner of one of these objects must call [dispose] when the object
 /// will no longer be used, in order to free resources on the [PerAccountStore].
 ///
 /// Lifecycle:
-///  * Create with [init].
+///  * Create an instance of a concrete subtype.
 ///  * Add listeners with [addListener].
 ///  * Use the [query] setter to start a search for a query.
 ///  * On reassemble, call [reassemble].
 ///  * When the object will no longer be used, call [dispose] to free
 ///    resources on the [PerAccountStore].
-class MentionAutocompleteView extends ChangeNotifier {
+abstract class AutocompleteView<QueryT extends AutocompleteQuery, ResultT extends AutocompleteResult, CandidateT> extends ChangeNotifier {
+  AutocompleteView({required this.store});
+
+  final PerAccountStore store;
+
+  Iterable<CandidateT> getSortedItemsToTest(QueryT query);
+
+  ResultT? testItem(QueryT query, CandidateT item);
+
+  QueryT? get query => _query;
+  QueryT? _query;
+  set query(QueryT? query) {
+    _query = query;
+    if (query != null) {
+      _startSearch(query);
+    }
+  }
+
+  /// Called when the app is reassembled during debugging, e.g. for hot reload.
+  ///
+  /// This will redo the search from scratch for the current query, if any.
+  void reassemble() {
+    if (_query != null) {
+      _startSearch(_query!);
+    }
+  }
+
+  Iterable<ResultT> get results => _results;
+  List<ResultT> _results = [];
+
+  Future<void> _startSearch(QueryT query) async {
+    final newResults = await _computeResults(query);
+    if (newResults == null) {
+      // Query was old; new search is in progress. Or, no listeners to notify.
+      return;
+    }
+
+    _results = newResults;
+    notifyListeners();
+  }
+
+  Future<List<ResultT>?> _computeResults(QueryT query) async {
+    final List<ResultT> results = [];
+    final Iterable<CandidateT> data = getSortedItemsToTest(query);
+
+    final iterator = data.iterator;
+    bool isDone = false;
+    while (!isDone) {
+      // CPU perf: End this task; enqueue a new one for resuming this work
+      await Future(() {});
+
+      if (query != _query || !hasListeners) { // false if [dispose] has been called.
+        return null;
+      }
+
+      for (int i = 0; i < 1000; i++) {
+        if (!iterator.moveNext()) {
+          isDone = true;
+          break;
+        }
+        final CandidateT item = iterator.current;
+        final result = testItem(query, item);
+        if (result != null) results.add(result);
+      }
+    }
+    return results;
+  }
+}
+
+class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery, MentionAutocompleteResult, User> {
   MentionAutocompleteView._({
-    required this.store,
+    required super.store,
     required this.narrow,
     required this.sortedUsers,
   });
@@ -182,6 +251,9 @@ class MentionAutocompleteView extends ChangeNotifier {
     store.autocompleteViewManager.registerMentionAutocomplete(view);
     return view;
   }
+
+  final Narrow narrow;
+  final List<User> sortedUsers;
 
   static List<User> _usersByRelevance({
     required PerAccountStore store,
@@ -289,6 +361,19 @@ class MentionAutocompleteView extends ChangeNotifier {
         streamId: streamId, senderId: userB.userId));
   }
 
+  @override
+  Iterable<User> getSortedItemsToTest(MentionAutocompleteQuery query) {
+    return sortedUsers;
+  }
+
+  @override
+  MentionAutocompleteResult? testItem(MentionAutocompleteQuery query, User item) {
+    if (query.testUser(item, store.autocompleteViewManager.autocompleteDataCache)) {
+      return UserMentionAutocompleteResult(userId: item.userId);
+    }
+    return null;
+  }
+
   /// Determines which of the two users is more recent in DM conversations.
   ///
   /// Returns a negative number if [userA] is more recent than [userB],
@@ -349,81 +434,43 @@ class MentionAutocompleteView extends ChangeNotifier {
     // TODO test that logic (may involve detecting an unhandled Future rejection; how?)
     super.dispose();
   }
-
-  final PerAccountStore store;
-  final Narrow narrow;
-  final List<User> sortedUsers;
-
-  MentionAutocompleteQuery? get query => _query;
-  MentionAutocompleteQuery? _query;
-  set query(MentionAutocompleteQuery? query) {
-    _query = query;
-    if (query != null) {
-      _startSearch(query);
-    }
-  }
-
-  /// Called when the app is reassembled during debugging, e.g. for hot reload.
-  ///
-  /// This will redo the search from scratch for the current query, if any.
-  void reassemble() {
-    if (_query != null) {
-      _startSearch(_query!);
-    }
-  }
-
-  Iterable<MentionAutocompleteResult> get results => _results;
-  List<MentionAutocompleteResult> _results = [];
-
-  Future<void> _startSearch(MentionAutocompleteQuery query) async {
-    final newResults = await _computeResults(query);
-    if (newResults == null) {
-      // Query was old; new search is in progress. Or, no listeners to notify.
-      return;
-    }
-
-    _results = newResults;
-    notifyListeners();
-  }
-
-  Future<List<MentionAutocompleteResult>?> _computeResults(MentionAutocompleteQuery query) async {
-    final List<MentionAutocompleteResult> results = [];
-    final iterator = sortedUsers.iterator;
-    bool isDone = false;
-    while (!isDone) {
-      // CPU perf: End this task; enqueue a new one for resuming this work
-      await Future(() {});
-
-      if (query != _query || !hasListeners) { // false if [dispose] has been called.
-        return null;
-      }
-
-      for (int i = 0; i < 1000; i++) {
-        if (!iterator.moveNext()) {
-          isDone = true;
-          break;
-        }
-
-        final User user = iterator.current;
-        if (query.testUser(user, store.autocompleteViewManager.autocompleteDataCache)) {
-          results.add(UserMentionAutocompleteResult(userId: user.userId));
-        }
-      }
-    }
-    return results;
-  }
 }
 
-class MentionAutocompleteQuery {
-  MentionAutocompleteQuery(this.raw, {this.silent = false})
+abstract class AutocompleteQuery {
+  AutocompleteQuery(this.raw)
     : _lowercaseWords = raw.toLowerCase().split(' ');
 
   final String raw;
+  final List<String> _lowercaseWords;
+
+  /// Whether all of this query's words have matches in [words] that appear in order.
+  ///
+  /// A "match" means the word in [words] starts with the query word.
+  bool _testContainsQueryWords(List<String> words) {
+    // TODO(#237) test with diacritics stripped, where appropriate
+    int wordsIndex = 0;
+    int queryWordsIndex = 0;
+    while (true) {
+      if (queryWordsIndex == _lowercaseWords.length) {
+        return true;
+      }
+      if (wordsIndex == words.length) {
+        return false;
+      }
+
+      if (words[wordsIndex].startsWith(_lowercaseWords[queryWordsIndex])) {
+        queryWordsIndex++;
+      }
+      wordsIndex++;
+    }
+  }
+}
+
+class MentionAutocompleteQuery extends AutocompleteQuery {
+  MentionAutocompleteQuery(super.raw, {this.silent = false});
 
   /// Whether the user wants a silent mention (@_query, vs. @query).
   final bool silent;
-
-  final List<String> _lowercaseWords;
 
   bool testUser(User user, AutocompleteDataCache cache) {
     // TODO(#236) test email too, not just name
@@ -434,25 +481,7 @@ class MentionAutocompleteQuery {
   }
 
   bool _testName(User user, AutocompleteDataCache cache) {
-    // TODO(#237) test with diacritics stripped, where appropriate
-
-    final List<String> nameWords = cache.nameWordsForUser(user);
-
-    int nameWordsIndex = 0;
-    int queryWordsIndex = 0;
-    while (true) {
-      if (queryWordsIndex == _lowercaseWords.length) {
-        return true;
-      }
-      if (nameWordsIndex == nameWords.length) {
-        return false;
-      }
-
-      if (nameWords[nameWordsIndex].startsWith(_lowercaseWords[queryWordsIndex])) {
-        queryWordsIndex++;
-      }
-      nameWordsIndex++;
-    }
+    return _testContainsQueryWords(cache.nameWordsForUser(user));
   }
 
   @override
@@ -489,7 +518,9 @@ class AutocompleteDataCache {
   }
 }
 
-sealed class MentionAutocompleteResult {}
+class AutocompleteResult {}
+
+sealed class MentionAutocompleteResult extends AutocompleteResult {}
 
 class UserMentionAutocompleteResult extends MentionAutocompleteResult {
   UserMentionAutocompleteResult({required this.userId});
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -217,3 +217,54 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
           ])));
   }
 }
+
+class TopicAutocomplete extends AutocompleteField<TopicAutocompleteQuery, TopicAutocompleteResult, String> {
+  const TopicAutocomplete({
+    super.key,
+    required this.streamId,
+    required ComposeTopicController super.controller,
+    required super.focusNode,
+    required this.contentFocusNode,
+    required super.fieldViewBuilder,
+  });
+
+  final FocusNode contentFocusNode;
+
+  final int streamId;
+
+  @override
+  ComposeTopicController get controller => super.controller as ComposeTopicController;
+
+  @override
+  AutocompleteIntent<TopicAutocompleteQuery>? autocompleteIntent() => controller.autocompleteIntent();
+
+  @override
+  TopicAutocompleteView initViewModel(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    return TopicAutocompleteView.init(store: store, streamId: streamId);
+  }
+
+  void _onTapOption(BuildContext context, TopicAutocompleteResult option) {
+    final intent = autocompleteIntent();
+    if (intent == null) return;
+    final replacementString = option.topic;
+    controller.value = intent.textEditingValue.replaced(
+      TextRange(
+        start: intent.syntaxStart,
+        end: intent.textEditingValue.text.length),
+      replacementString,
+    );
+    contentFocusNode.requestFocus();
+  }
+
+  @override
+  Widget buildItem(BuildContext context, int index, TopicAutocompleteResult option) {
+    return InkWell(
+      onTap: () {
+        _onTapOption(context, option);
+      },
+      child: Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
+        child: Text(option.topic)));
+  }
+}
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 
+import '../api/model/model.dart';
 import 'content.dart';
 import 'store.dart';
 import '../model/autocomplete.dart';
@@ -7,37 +8,38 @@ import '../model/compose.dart';
 import '../model/narrow.dart';
 import 'compose_box.dart';
 
-class ComposeAutocomplete extends StatefulWidget {
-  const ComposeAutocomplete({
+abstract class AutocompleteField<QueryT extends AutocompleteQuery, ResultT extends AutocompleteResult, CandidateT> extends StatefulWidget {
+  const AutocompleteField({
     super.key,
-    required this.narrow,
     required this.controller,
     required this.focusNode,
     required this.fieldViewBuilder,
   });
 
-  /// The message list's narrow.
-  final Narrow narrow;
-
-  final ComposeContentController controller;
+  final TextEditingController controller;
   final FocusNode focusNode;
   final WidgetBuilder fieldViewBuilder;
 
+  AutocompleteIntent<QueryT>? autocompleteIntent();
+
+  Widget buildItem(BuildContext context, int index, ResultT option);
+
+  AutocompleteView<QueryT, ResultT, CandidateT> initViewModel(BuildContext context);
+
   @override
-  State<ComposeAutocomplete> createState() => _ComposeAutocompleteState();
+  State<AutocompleteField<QueryT, ResultT, CandidateT>> createState() => _AutocompleteFieldState<QueryT, ResultT, CandidateT>();
 }
 
-class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccountStoreAwareStateMixin<ComposeAutocomplete> {
-  MentionAutocompleteView? _viewModel; // TODO different autocomplete view types
+class _AutocompleteFieldState<QueryT extends AutocompleteQuery, ResultT extends AutocompleteResult, CandidateT> extends State<AutocompleteField<QueryT, ResultT, CandidateT>> with PerAccountStoreAwareStateMixin<AutocompleteField<QueryT, ResultT, CandidateT>> {
+  AutocompleteView<QueryT, ResultT, CandidateT>? _viewModel;
 
   void _initViewModel() {
-    final store = PerAccountStoreWidget.of(context);
-    _viewModel = MentionAutocompleteView.init(store: store, narrow: widget.narrow)
+    _viewModel = widget.initViewModel(context)
       ..addListener(_viewModelChanged);
   }
 
-  void _composeContentChanged() {
-    final newAutocompleteIntent = widget.controller.autocompleteIntent();
+  void _handleControllerChange() {
+    final newAutocompleteIntent = widget.autocompleteIntent();
     if (newAutocompleteIntent != null) {
       if (_viewModel == null) {
         _initViewModel();
@@ -55,7 +57,7 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
   @override
   void initState() {
     super.initState();
-    widget.controller.addListener(_composeContentChanged);
+    widget.controller.addListener(_handleControllerChange);
   }
 
   @override
@@ -69,22 +71,22 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
   }
 
   @override
-  void didUpdateWidget(covariant ComposeAutocomplete oldWidget) {
+  void didUpdateWidget(covariant AutocompleteField<QueryT, ResultT, CandidateT> oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (widget.controller != oldWidget.controller) {
-      oldWidget.controller.removeListener(_composeContentChanged);
-      widget.controller.addListener(_composeContentChanged);
+      oldWidget.controller.removeListener(_handleControllerChange);
+      widget.controller.addListener(_handleControllerChange);
     }
   }
 
   @override
   void dispose() {
-    widget.controller.removeListener(_composeContentChanged);
+    widget.controller.removeListener(_handleControllerChange);
     _viewModel?.dispose(); // removes our listener
     super.dispose();
   }
 
-  List<MentionAutocompleteResult> _resultsToDisplay = [];
+  List<ResultT> _resultsToDisplay = [];
 
   void _viewModelChanged() {
     setState(() {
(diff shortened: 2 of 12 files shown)


lib/model/stream_topics.dart
```diff
@@ -0,0 +1,220 @@
...
+
+/// A per-account manager for the view-models of autocomplete interactions.
+///
+/// There should be exactly one of these per PerAccountStore.
+///
+/// Since this manages a cache of user data, the handleRealmUser…Event functions
+/// must be called as appropriate.
+///
+/// On reassemble, call [reassemble].
+class TopicAutocompleteViewManager {
```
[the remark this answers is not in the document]
gnprice: Similarly let's also put `TopicAutocompleteView` and friends in the existing `model/autocomplete.dart` file.

lib/widgets/compose_box.dart
```diff
@@ -887,10 +910,26 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> implements Compose
...
       contentFocusNode: _contentFocusNode,
-      topicInput: TextField(
-        controller: _topicController,
-        style: TextStyle(color: colorScheme.onSurface),
-        decoration: InputDecoration(hintText: zulipLocalizations.composeBoxTopicHintText),
+      topicInput: Autocomplete<Topic>(
+        optionsViewOpenDirection: OptionsViewOpenDirection.up,
+        optionsBuilder: (textEditingValue) => model?.results ?? [],
+        initialValue: _topicController.value,
+        displayStringForOption: (option) {
```
gnprice, line 917: Yeah, all this stuff in the build method is the other major reason it'll be helpful (as [discussed in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Autocomplete.20for.20topic.20to.20send.20to.20.23F310/near/1785736)) to pull out a separate widget that's specifically responsible for the autocomplete. That way `_StreamComposeBoxState` gets to avoid having to know about all these details of autocomplete, and stay focused on putting together the compose box as a whole.

lib/model/stream_topics.dart
```diff
@@ -0,0 +1,220 @@
...
+    if (newResults == null) {
+      // Query was old; new search is in progress. Or, no listeners to notify.
+      return;
+    }
+
+    _results = newResults;
+    notifyListeners();
+  }
+
+  Future<List<Topic>?> _computeResults(TopicAutocompleteQuery query) async {
```
gnprice, line 129: This method, and its caller `_startSearch`, and the whole mechanism of `query` and `results`, is fairly tricky. It's also quite generic to the idea of autocomplete in general, not particularly specific to doing autocomplete on @-mentions vs. topics vs. anything else.

So it'd be good to share this code. To do that, a good commit structure would be:
* First, a prep commit factors it out into something generic which `MentionAutocompleteView` uses.
* Then, the PR's main commit adds `TopicAutocompleteView` as a new class that also uses the same generic mechanism.

Probably the way to structure that code-sharing is as a shared base class: something like
```dart
abstract class AutocompleteView<Q, R> extends ChangeNotifier {
  // …
  Q? get query => _query;
  // …
  Iterable<R> get results => _results;
  // …
```

I guess abstracting out `_computeResults` is slightly trickier. But even if that part stays duplicated and it's only the other parts that become shared, that'll be helpful.

lib/api/route/streams.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:json_annotation/json_annotation.dart';
+
+import '../core.dart';
+part 'streams.g.dart';
+
+/// https://zulip.com/api/v1/user/me/:stream_id/topics
+Future<GetTopicsResult> getStreamTopics(
```
gnprice, line 7: This should link to the corresponding API docs page — that's the page the reader will want to refer to in order to understand the semantics of this function.

gnprice (commented):
Thanks @Khader-1! Generally this strategy looks good, modulo the comment already made in [the chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Autocomplete.20for.20topic.20to.20send.20to.20.23F310/near/1785736). A few other high-level comments below.

lib/widgets/compose_box.dart
```diff
@@ -372,6 +374,79 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+    super.dispose();
+  }
+
+
+  @override
+  void onNewStore() {
+    _initModel(PerAccountStoreWidget.of(context));
+  }
+
+  void _initModel(PerAccountStore store)async  {
```
Khader-1, line 408: I am not sure if it is right for the api call to be in the widget directly.

lib/widgets/compose_box.dart
```diff
@@ -7,6 +7,8 @@ import 'package:image_picker/image_picker.dart';
 
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
+import '../api/route/streams.dart';
+import '../model/autocomplete.dart' as autocomplete;
```
Khader-1, line 11: This is here to avoid the ambiguity resulting from the presence of the `Autocomplete` extension in `../model/autocomplete.dart`.

lib/model/autocomplete.dart
```diff
@@ -244,6 +234,40 @@ class MentionAutocompleteView extends ChangeNotifier {
     notifyListeners();
   }
 
+  Future<List<R>?> _computeResults(Q query);
```
Khader-1, line 237: This could be refactored to avoid duplications by using a sort of strategy pattern.

Khader-1 (commented):
Thanks @gnprice! I have made some changes. Yet this still needs some tests. Some comments about things I am not so sure about are below. Any other notes or suggestions are welcome.

lib/model/autocomplete.dart
```diff
@@ -352,3 +376,133 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+
+  Future<void> fetch() async {
+    final result = await getStreamTopics(store.connection, streamId: narrow.streamId);
+    topics = result.topics;
+    notifyListeners();
+  }
+
+  @override
+  Future<List<Topic>?> _computeResults(TopicAutocompleteQuery query) async {
+    if (topics == null) return [];
```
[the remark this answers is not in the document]
gnprice: If in order to handle that scenario you find you want to modify `_startSearch` and/or the `query` setter, that's fine: just override them here in this class. After we have a working version, we can look for ways to factor things back out.

(Also: this is a scenario we'll definitely want a test for.)

lib/widgets/compose_box.dart
```diff
@@ -372,6 +374,79 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+        optionsBuilder: (textEditingValue) => model?.results ?? [],
+        initialValue: widget.controller.value,
+        displayStringForOption: (option) {
+          return option.name;
+        },
+        onSelected: (option) => widget.controller.text = option.name,
+        fieldViewBuilder: (context, textEditingController, focusNode, onFieldSubmitted) => TextField(
+          controller: textEditingController,
+          onChanged: (value) {
+            widget.controller.text = value;
```
gnprice, line 437: It looks like this will end up with having two separate controllers for the topic input, one of which syncs changes to the other via this `onChanged` callback.

That seems like a recipe for having annoying bugs in the future where the controllers get out of sync, or where we add some behavior to `ComposeTopicController` and that behavior doesn't quite work correctly because it's not on the actual controller that's attached directly to the text input.

Instead, let's use the controller we already have (`widget.controller`) and make that the actual controller for the text input itself. For an example, see how we handle autocomplete in the content input, as discussed earlier [in chat](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Autocomplete.20for.20topic.20to.20send.20to.20.23F310/near/1785736).

lib/widgets/autocomplete.dart
```diff
@@ -171,7 +224,7 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
                 padding: EdgeInsets.zero,
                 shrinkWrap: true,
                 itemCount: _resultsToDisplay.length,
-                itemBuilder: _buildItem))));
+                itemBuilder: (context, index) => widget.itemBuilder(context, index, _resultsToDisplay[index])))));
```
sm-sayedi, line 227: nit: Too long for one line.

lib/widgets/compose_box.dart
```diff
@@ -372,6 +383,43 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+        contentFocusNode: contentFocusNode,
+        fieldViewBuilder: (context) {
+          return TextField(
+            controller: controller,
+            focusNode: focusNode,
+            textInputAction: TextInputAction.next,
+            style: TextStyle(color: colorScheme.onSurface),
+            decoration: InputDecoration.collapsed(hintText: zulipLocalizations.composeBoxTopicHintText),
+          );
+        },));
```
sm-sayedi, line 419:
```suggestion
        fieldViewBuilder: (context) => TextField(
          controller: controller,
          focusNode: focusNode,
          textInputAction: TextInputAction.next,
          style: TextStyle(color: colorScheme.onSurface),
          decoration: InputDecoration.collapsed(hintText: zulipLocalizations.composeBoxTopicHintText))));
```

sm-sayedi (commented):
Thank you @Khader-1 for all the work on this. The generic approach seems promising. Added a few comments, mostly small. And, it would be better to see this PR rebased on the main for the next revision as some changes happened in `MentionAutocompleteView` in #693 being merged.

Also, the following commits fail the tests (need to pass):

- [e955ef9](https://github.com/zulip/zulip-flutter/pull/627/commits/e955ef951cf898306bb88648eda6bed1986a3aec): autocomplete [nfc]: Remove unused narrow in MentionAutocompleteView
- [3f55c7b](https://github.com/zulip/zulip-flutter/pull/627/commits/3f55c7b26ec86b71ada5174721bc979bad529d47): autocomplete[nfc]: Use generic fields in AutocompleteViewManager

lib/api/route/streams.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:json_annotation/json_annotation.dart';
+
```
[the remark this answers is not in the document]
Khader-1: I see, but I believe it would be more aligned with the code base to have the topics' functionality nested within streams. for example:


```dart
// lib/model/stream.dart

  /// Whether this topic should appear when not specifically focusing
  /// on this stream.
  ///
  /// This takes into account the user's visibility policy for the stream
  /// overall, as well as their policy for this topic.
  ///
  /// For UI contexts that are specific to a particular stream, see
  /// [isTopicVisibleInStream].
  bool isTopicVisible(int streamId, String topic) {
    switch (topicVisibilityPolicy(streamId, topic)) {
      case UserTopicVisibilityPolicy.none:
        switch (subscriptions[streamId]?.isMuted) {
          case false: return true;
          case true:  return false;
          case null:  return false; // not subscribed; treat like muted
        }
      case UserTopicVisibilityPolicy.muted:
        return false;
      case UserTopicVisibilityPolicy.unmuted:
      case UserTopicVisibilityPolicy.followed:
        return true;
      case UserTopicVisibilityPolicy.unknown:
        assert(false);
        return true;
    }
  }
}
```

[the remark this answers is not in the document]
Khader-1: Also topic it self is absolutely coupled with a stream and I can't think of any functionality for a topic without having to deal with it's stream.

Khader-1:
Thanks @sm-sayedi! you're comments are really helpful!
I've pushed a new revision PTAL and please refer to this [discussion regarding naming (streams / topics)](https://github.com/zulip/zulip-flutter/pull/627#discussion_r1636304597).

lib/model/autocomplete.dart
```diff
@@ -363,3 +467,77 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
 // TODO(#233): // class UserGroupMentionAutocompleteResult extends MentionAutocompleteResult {
 
 // TODO(#234): // class WildcardMentionAutocompleteResult extends MentionAutocompleteResult {
+
+class TopicAutocompleteDataProvider extends AutocompleteDataProvider<Topic, TopicAutocompleteQuery, TopicAutocompleteResult> {
+  final PerAccountStore store;
+  final int streamId;
+  Iterable<Topic>? topics;
+  bool isFetching = false;
```
sm-sayedi, line 475:
```suggestion
  Iterable<Topic>? _topics;
  bool _isFetching = false;
```
I think it is safe to change these to private fields.

lib/model/autocomplete.dart
```diff
@@ -243,11 +250,22 @@ class MentionAutocompleteView extends ChangeNotifier {
     }
   }
 
-  Iterable<MentionAutocompleteResult> get results => _results;
-  List<MentionAutocompleteResult> _results = [];
+  Iterable<R> get results => _results;
+  List<R> _results = [];
+
+  Future<void> _startSearch(Q query) async {
```
sm-sayedi, line 256: The implementation of this method is changed in #693, specifically, there is no retry now. PTAL.

sm-sayedi (commented):
Thanks for the revision. Two comments below.

lib/model/autocomplete.dart
Khader-1: Oh sorry! I must have missed this one while rebasing.

lib/model/autocomplete.dart
Khader-1: Yes, thanks!

Khader-1:
Thanks @sm-sayedi! I've pushed a new revision PTAL.

sm-sayedi (approved):
Thanks @Khader-1 for all the work on this. LGTM.

Let's move on to the mentor review with @kenclary!

kenclary (approved):
good stuff

alya:
@Khader-1 Could you please add screenshots that show your changes to the PR description?

test/example_data.dart
```diff
@@ -180,6 +181,13 @@ ZulipStream stream({
 }
 const _stream = stream;
 
+Topic topic({
+  int? maxId,
+  String? name,
+}) => Topic(
+  maxId: maxId ?? 123,
+  name: name ?? 'Test Topic #$maxId');
```
chrisbobbe, line 189: If `maxId` and `name` are both omitted, won't this say 'Test Topic #null' or something?

test/model/autocomplete_test.dart
```diff
@@ -474,4 +477,112 @@ void main() {
...
+  });
+
+  group('TopicAutocompleteQuery.testTopic', () {
+    doCheck(String rawQuery, Topic topic, bool expected) {
+      final result = TopicAutocompleteQuery(rawQuery)
+        .testTopic(topic, AutocompleteDataCache());
+      expected ? check(result).isTrue() : check(result).isFalse();
+    }
+
+    test('topic is included if name matches the query', () {
```
chrisbobbe, line 557: How about 'topic is included if it matches the query'?

test/widgets/autocomplete_test.dart
```diff
@@ -72,6 +74,52 @@ Future<Finder> setupToComposeInput(WidgetTester tester, {
...
+  await tester.pumpWidget(Builder(builder: (context) =>
+    MaterialApp(
+      theme: zulipThemeData(context),
+      localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+      supportedLocales: ZulipLocalizations.supportedLocales,
+      home: GlobalStoreWidget(
+        child: PerAccountStoreWidget(
+          accountId: eg.selfAccount.id,
+          child: MessageListPage(
+            narrow: StreamNarrow(stream.streamId)))))));
```
chrisbobbe, line 111: A bit shorter, and also a bit more representative:

```suggestion
  await tester.pumpWidget(const ZulipApp());
  await tester.pump();
  final navigator = await ZulipApp.navigator;
  navigator.push(MessageListPage.buildRoute(accountId: eg.selfAccount.id,
    narrow: StreamNarrow(stream.streamId)));
```

I'd like to do this with `setupToComposeInput` too but haven't gotten around to it yet.

lib/model/autocomplete.dart
```diff
@@ -151,6 +151,20 @@ class AutocompleteViewManager {
...
 }
 
+/// Provides the full list of possible items and filter predicate for
+/// `AutocompleteView` without having to deal with state management or
+/// result computation process.
+///
+/// Having this factored out of `AutocompleteView` helps in limiting
+/// it's responsibility to managing the autocomplete state and result
+/// computation process.
+abstract class AutocompleteDataProvider<T, Q extends AutocompleteQuery, R extends AutocompleteResult> {
```
chrisbobbe, line 195: Reading @gnprice's review, I'm not sure if this was part of the factoring he suggested; Greg, what are your thoughts on it?

lib/widgets/compose_box.dart
```diff
@@ -886,16 +922,14 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> implements Compose
...
   @override
   Widget build(BuildContext context) {
-    final colorScheme = Theme.of(context).colorScheme;
-    final zulipLocalizations = ZulipLocalizations.of(context);
-
     return _ComposeBoxLayout(
       contentController: _contentController,
       contentFocusNode: _contentFocusNode,
-      topicInput: TextField(
+      topicInput: _StreamTopicInput(
```
chrisbobbe, line 928: `_StreamTopicInput` uses `AutocompleteInputWrapper`, which (in this revision) is purely about the layout/appearance, as mentioned in a previous comment. It adds an `InputDecorator` and a `ConstrainedBox`. What's the effect of adding those things here too?

lib/widgets/compose_box.dart
```diff
@@ -265,6 +265,22 @@ class ComposeContentController extends ComposeController<ContentValidationError>
...
+
+  @override
+  Widget build(BuildContext context) => InputDecorator(
+    decoration: const InputDecoration(),
+    child: ConstrainedBox(
+      constraints: const BoxConstraints(
+        minHeight: _sendButtonSize - 2 * _inputVerticalPadding,
+        // TODO constrain this adaptively (i.e. not hard-coded 200)
+        maxHeight: 200,
+      ), child:child));
```
chrisbobbe, line 281:
```suggestion
      ),
      child: child));
```

chrisbobbe (changes requested):
Thanks, @Khader-1! Comments below.

lib/model/autocomplete.dart
```diff
@@ -111,34 +122,57 @@ class AutocompleteIntent {
 ///
 /// On reassemble, call [reassemble].
 class AutocompleteViewManager {
-  final Set<MentionAutocompleteView> _mentionAutocompleteViews = {};
+  final Map<Type, Set<AutocompleteView>> _views = {};
```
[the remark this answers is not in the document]
Khader-1: The motivation behind using this collection is to provide a generic way for registering and unregistering the vm without having to repeat the some code blocks and instead having that logic implemented in the parent `AutocompleteView` making it a little bit cleaner and less complicated for subclasses implementing it.

Having in mind that we don't expect to have a big number of kinds of `AutocompleteView` this is completely optional.

lib/model/autocomplete.dart
```diff
@@ -151,83 +151,34 @@ class AutocompleteViewManager {
...
-  /// and returns `0` if both [userA] and [userB] are equally recent
-  /// or there is no DM exchanged with them whatsoever.
-  @visibleForTesting
-  static int compareByDms(User userA, User userB, {required PerAccountStore store}) {
-    final recentDms = store.recentDmConversationsView;
-    final aLatestMessageId = recentDms.latestMessagesByRecipient[userA.userId];
-    final bLatestMessageId = recentDms.latestMessagesByRecipient[userB.userId];
+  /// This could be used to transform results after they've been
+  /// computed for sorting, filtering etc.
+  final List<R> Function(List<R> results)? resultsFilter;
```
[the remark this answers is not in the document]
Khader-1: I expected to use this specifically when merging changes of #228. Now that it is already merged it seems to be unnecessary. Will not include it in the next revision..

lib/model/autocomplete.dart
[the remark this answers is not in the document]
gnprice: Yeah, I think this will be simplest if it's a separate variable for each type of view, rather than a `Map<Type, …>`.

It means a few similar-looking registerFoo / unregisterFoo methods. But each of them will be quite short, so there's not much gained by deduplicating them.

lib/model/autocomplete.dart
gnprice: Yeah, I think this can be simplified by making the two `AutocompleteDataProvider` methods be just abstract methods on `AutocompleteView` (like they are in the commit before this), and the concrete views implement those methods.

In the current revision, the split between AutocompleteDataProvider and AutocompleteView isn't really doing much to separate complexity because both of the AutocompleteView subclasses end up being pretty trivial, mostly just identifying how to construct the AutocompleteDataProvider. That is, virtually all the complexity lives on one side of the division, so the division isn't dividing much.

lib/widgets/compose_box.dart
```diff
@@ -265,6 +265,22 @@ class ComposeContentController extends ComposeController<ContentValidationError>
   }
 }
 
+class AutocompleteInputWrapper extends StatelessWidget {
```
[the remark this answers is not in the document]
Khader-1: No nothing specific to autocomplete.. it's just that I am using the same decoration and box constraints for both `TopicAutocomplete` and `ComposeContentAutocomplete`.

lib/widgets/compose_box.dart
Khader-1: It's purely for layout and appearance. When I try to drop it, the decoration of the topic-autocomplete text field gets messed up.
image

Khader-1:
Thanks for the review @chrisbobbe! And Thanks @alya!
Please check the latest revision and the updated video in PR description.

alya:
I noticed that in the video, the cursor is initially in the message area. Where do we initially put the cursor when the compose box is opened?

alya:
What's the topic ordering? Is it the same as in the web app, or a different algorithm?

Khader-1:
Please check the updated video now, @alya. It shows a clearer life cycle. Regarding the order, there is currently no sorting algorithm. I'll look into the web repo to copy its algorithm.

lib/model/autocomplete.dart
```diff
@@ -350,9 +406,19 @@ class AutocompleteDataCache {
...
     _nameWordsByUser.remove(userId);
   }
+
+  List<String> nameWordsForTopic(Topic topic) {
+    return _nameWordsByTopic[topic.value] ??= topic.value.toLowerCase().split(' ');
+  }
+
+  void invalidateTopic(String value) {
+    _nameWordsByTopic.remove(value);
+  }
```
chrisbobbe, line 416: Bump on https://github.com/zulip/zulip-flutter/pull/627#discussion_r1663327014 ; "name words" is confusing when talking about topics.

lib/model/autocomplete.dart
```diff
@@ -133,6 +155,12 @@ class AutocompleteViewManager {
     autocompleteDataCache.invalidateUser(event.userId);
   }
 
+  void handleTopicsFetchCompleted() {
+    for (final view in _topicAutocompleteViews) {
+      view.reassemble();
```
chrisbobbe, line 160: bump on https://github.com/zulip/zulip-flutter/pull/627#discussion_r1663479516

What's the goal with this method? The UI should update as long as the updated view's `notifyListeners()` is called in the right places, right?

…aha, I see that `TopicAutocompleteView.fetch` doesn't call `notifyListeners`. If you add that, then I think we can just drop this `handleTopicsFetchCompleted`.

test/example_data.dart
```diff
@@ -180,6 +181,16 @@ ZulipStream stream({
...
 const _stream = stream;
 
+Topic topic({
+  int? maxId,
+  String? value,
+}) {
+  maxId ??= 123;
+  return Topic(
+  maxId: maxId,
+  value: value ?? 'Test Topic #$maxId');
```
chrisbobbe, line 191:
```suggestion
  return Topic(maxId: maxId, value: value ?? 'Test Topic #$maxId');
```

lib/widgets/compose_box.dart
```diff
@@ -372,6 +372,38 @@ class _StreamContentInputState extends State<_StreamContentInput> {
   }
 }
 
+class _StreamTopicInput extends StatelessWidget {
+  const _StreamTopicInput({
```
chrisbobbe, line 376: Would `_TopicInput` be a better name? There's nothing in it about letting the user choose a stream.

lib/widgets/autocomplete.dart
```diff
@@ -7,37 +7,131 @@ import '../model/compose.dart';
...
+      return TopicAutocompleteView.init(store: store, streamId: streamId);
+    },
+    itemBuilder: (context, index, option) => InkWell(
+      onTap: () {
+        final intent = controller.autocompleteIntent();
+        if (intent == null) return;
+        final label = option.topic.value;
+        controller.value = intent.textEditingValue.replaced(TextRange(
+          start: intent.syntaxStart,
+          end: intent.textEditingValue.selection.end), label);
```
chrisbobbe, line 95:
```suggestion
        controller.value = intent.textEditingValue.replaced(TextRange(
          start: intent.syntaxStart,
          end: intent.textEditingValue.text.length), label);
```

Unlike with the content input, I think the user will always want the entire text in the input replaced, even when the cursor is somewhere in the middle of the input for some reason. Let's add a test to confirm this behavior, too.

chrisbobbe (commented):
Thanks, @Khader-1! Comments below.

lib/model/autocomplete.dart
```diff
@@ -274,58 +245,134 @@ class MentionAutocompleteView extends ChangeNotifier {
...
       }
     }
     return results;
   }
 }
 
-class MentionAutocompleteQuery {
-  MentionAutocompleteQuery(this.raw, {this.silent = false})
-    : _lowercaseWords = raw.toLowerCase().split(' ');
+class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery, MentionAutocompleteResult, User> {
```
Khader-1, line 257: The data type here will be changed eventually when adding other ***mentionable types*** but I think leaving it as `User` for now is okay.

Khader-1 (commented):
Thanks @chrisbobbe for the review! PTAL!

test/widgets/autocomplete_test.dart
```diff
@@ -72,6 +67,49 @@ Future<Finder> setupToComposeInput(WidgetTester tester, {
...
 }
 
+/// Simulates loading a [MessageListPage] with a stream narrow
+/// and tapping to focus the topic input.
+///
+/// Also prepares test-topics to be sent to topics api request,
+/// so they can show up in autocomplete.
+///
+/// Returns the found `TextField` element with the hintText matching
+/// topic's input hint
```
chrisbobbe, line 77:
```suggestion
/// Returns a [Finder] for the topic input's [TextField].
```

lib/model/autocomplete.dart
```diff
@@ -363,3 +421,76 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+
+class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, TopicAutocompleteResult, Topic> {
+  final int streamId;
+  Iterable<Topic> _topics = [];
+  bool _isFetching = false;
+
+  TopicAutocompleteView.init({
+      required super.store,
+      required this.streamId,
+  }) {
```
chrisbobbe, line 433:
```suggestion
  TopicAutocompleteView.init({required super.store, required this.streamId}) {
```

lib/widgets/autocomplete.dart
```diff
@@ -1,43 +1,139 @@
...
+      return TopicAutocompleteView.init(store: store, streamId: streamId);
+    },
+    itemBuilder: (context, index, option) => InkWell(
+      onTap: () {
+        final intent = controller.autocompleteIntent();
+        if (intent == null) return;
+        final label = option.topic.value;
+        controller.value = intent.textEditingValue.replaced(TextRange(
+          start: intent.syntaxStart,
+          end: intent.textEditingValue.text.length), label);
```
chrisbobbe, line 97: Bump on getting test coverage for this behavior: https://github.com/zulip/zulip-flutter/pull/627#discussion_r1685049138

chrisbobbe (commented):
Thanks! Small comments below.

Khader-1:
Thanks @chrisbobbe! I've pushed a new revision PTAL.

Khader-1:
Thanks @chrisbobbe! I've pushed a new revision PTAL.

Khader-1:
Thanks @chrisbobbe! I've pushed a new revision PTAL.

chrisbobbe:
Thanks! Looks like there's a conflict from the just-merged #805. I think you can resolve by skipping this commit:

```
autocomplete test [nfc]: Refactor widget setup in `setupToComposeInput`
```

and using `TestZulipApp` instead of `ZulipApp` in test/widgets/autocomplete_test.dart in the main commit.

Also, if you get some test failures about `TextDirection.ltr` and `TextDirection.rtl`, run `flutter run` and rerun the tests, and they should succeed. (There's an autogenerated file that should get updated; let me know if that doesn't work.)

Khader-1:
Thanks @chrisbobbe! I've resolved with a new revision. PTAL

chrisbobbe:
Thanks! Marking as ready for Greg's review.

lib/model/autocomplete.dart
```diff
@@ -317,110 +413,61 @@ class MentionAutocompleteView extends ChangeNotifier {
...
-    : _lowercaseWords = raw.toLowerCase().split(' ');
+  @override
+  String toString() {
+    return '${objectRuntimeType(this, 'AutocompleteQuery')}(raw: $raw})';
+  }
 
-  final String raw;
+  @override
+  bool operator ==(Object other) {
+    return other is AutocompleteQuery && other.raw == raw;
```
gnprice, line 453: This is wrong — for example it would let queries of two different concrete subclasses be equal just because `raw` was equal.

Probably this override just shouldn't be here. It can't ever get invoked, right? Both concrete subclasses override it.

Similarly the toString and hashCode seem like they can't get invoked and so should just be dropped.

test/model/autocomplete_checks.dart
```diff
@@ -3,14 +3,28 @@ import 'package:zulip/model/autocomplete.dart';
...
-  Subject<AutocompleteIntent?> get autocompleteIntent => has((c) => c.autocompleteIntent(), 'autocompleteIntent');
+  Subject<AutocompleteIntent<MentionAutocompleteQuery>?> get autocompleteIntent => has((c) => c.autocompleteIntent(), 'autocompleteIntent');
 }
 
-extension AutocompleteIntentChecks on Subject<AutocompleteIntent> {
+extension ComposeTopicControllerChecks on Subject<ComposeTopicController> {
+  Subject<AutocompleteIntent<TopicAutocompleteQuery>?> get autocompleteIntent => has((c) => c.autocompleteIntent(), 'autocompleteIntent');
+}
+
+extension AutocompleteIntentChecks on Subject<AutocompleteIntent<MentionAutocompleteQuery>> {
```
gnprice, line 13: This doesn't look right — the extension is named as if it's on AutocompleteIntent in general, but it's only on a specific subtype.

```suggestion
extension AutocompleteIntentChecks on Subject<AutocompleteIntent> {
```

You'll have to update a getter to match the changes to the class.

Similarly in `test/widgets/compose_box_checks.dart`.

gnprice (commented):
Thanks @Khader-1, and thanks @sm-sayedi and @chrisbobbe for the previous reviews!

For this round, I've read the first two commits:
c6d3ae103 autocomplete [nfc]: Rename Autocomplete to ComposeContentAutocomplete
28c2694b7 autocomplete [nfc]: Factor out generic AutocompleteView

Comments on those below. I think some of the comments are likely to apply also to the third commit:
b2f9e5c2a autocomplete [nfc]: Factor out generic AutocompleteField

so please try to revise that as well where these comments relate to it.

lib/model/autocomplete.dart
```diff
@@ -151,21 +173,90 @@ class AutocompleteViewManager {
...
 /// Lifecycle:
-///  * Create with [init].
+///  * Create an instance of a concrete subtype.
 ///  * Add listeners with [addListener].
 ///  * Use the [query] setter to start a search for a query.
 ///  * On reassemble, call [reassemble].
 ///  * When the object will no longer be used, call [dispose] to free
 ///    resources on the [PerAccountStore].
-class MentionAutocompleteView extends ChangeNotifier {
+abstract class AutocompleteView<QueryT extends AutocompleteQuery, ResultT extends AutocompleteResult, CandidateT> extends ChangeNotifier {
```
gnprice, line 188: nit in commit message: indented summary

     autocomplete [nfc]: Factor out generic AutocompleteView
    
    The mechanism of query and results is generic to the idea of
    autocomplete in general, it's not specific to autocomplete on
    @-mentions vs. topics vs. anything else.

lib/widgets/autocomplete.dart
```diff
@@ -1,43 +1,139 @@
...
 import 'content.dart';
 import 'store.dart';
 import '../model/autocomplete.dart';
 import '../model/compose.dart';
 import '../model/narrow.dart';
 import 'compose_box.dart';
 
-class ComposeAutocomplete extends StatefulWidget {
-  const ComposeAutocomplete({
+class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, MentionAutocompleteResult, User> {
```
gnprice, line 12: nit: define base class first, then define subclass

lib/widgets/autocomplete.dart
```diff
@@ -1,43 +1,139 @@
...
+    viewModelBuilder: (context) {
+      final store = PerAccountStoreWidget.of(context);
+      return MentionAutocompleteView.init(store: store, narrow: narrow);
+    },
+    itemBuilder: (context, index, option) {
+      Widget avatar;
+      String label;
+      switch (option) {
+        case UserMentionAutocompleteResult(:var userId):
+          avatar = Avatar(userId: userId, size: 32, borderRadius: 3);
```
gnprice, line 31: This gets fairly messy to read, as a long function expression in a constructor's initializer list.

Let's define `itemBuilder` and friends on the base class as methods instead of fields, so that the subclasses can just override those methods. See our `MessageActionSheetMenuItemButton` for an example.

lib/api/route/streams.dart
```diff
@@ -0,0 +1,43 @@
...
+import '../core.dart';
+part 'streams.g.dart';
```
gnprice, line 4: Let's name this file `channels.dart` instead of `streams.dart`, so you don't have to rename it later as part of #631 :slightly_smiling_face:

lib/api/route/streams.dart
```diff
@@ -0,0 +1,43 @@
...
+
+/// https://zulip.com/api/get-stream-topics
+Future<GetTopicsResult> getStreamTopics(
```
gnprice, line 7: OTOH I guess keep the method name the same as the operation is called in the API docs, like this.

lib/api/route/streams.dart
```diff
@@ -0,0 +1,43 @@
...
+  ApiConnection connection, {
```
gnprice, line 8: nit: format the same as other endpoint methods:
```suggestion
Future<GetTopicsResult> getStreamTopics(ApiConnection connection, {
```

lib/api/route/streams.dart
```diff
@@ -0,0 +1,43 @@
...
+  required int streamId,
```
gnprice, line 9: (also keep this as `streamId`; renaming all the "stream ID" variables and parameters and fields will be a later phase of #631)

lib/api/route/streams.dart
gnprice, line 7: Name the result class to match the endpoint method, following the pattern of other endpoints:

```suggestion
Future<GetStreamTopicsResult> getStreamTopics(
```

lib/api/route/streams.dart
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  factory GetTopicsResult.fromJson(Map<String, dynamic> json) =>
+    _$GetTopicsResultFromJson(json);
+
+  Map<String, dynamic> toJson() => _$GetTopicsResultToJson(this);
+}
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class Topic {
```
gnprice, line 30: Class names effectively live in one global namespace, not local to a file. (After all, this is a type we end up using in some other files.) So this name is much too general for this class, which has a fairly specialized purpose.

In particular, we have tons of variables and parameters etc. that mean a topic and are called "topic" — they all have type String, because that's how Zulip identifies topics.

One good name would be:

```suggestion
class GetStreamTopicsEntry {
```

Bit long, but it does the job of appropriately placing this class in context.

lib/api/route/streams.dart
```diff
@@ -0,0 +1,43 @@
...
+  final int maxId;
+  @JsonKey(name: 'name')
+  final String value;
```
gnprice, line 33: `name` is a perfectly reasonable name for this — so let's stay aligned with the server API.

lib/model/autocomplete.dart
```diff
@@ -43,6 +44,16 @@ extension Autocomplete on ComposeContentController {
   }
 }
 
+extension ComposeTopicAutocomplete on ComposeTopicController {
+  AutocompleteIntent<TopicAutocompleteQuery>? autocompleteIntent() {
+    if (!selection.isValid || !selection.isNormalized) return null;
```
gnprice, line 49: Is there a reason we need these checks? It doesn't seem like we're using the selection in the remaining logic.

lib/model/autocomplete.dart
```diff
@@ -447,9 +499,19 @@ class AutocompleteDataCache {
   void invalidateUser(int userId) {
     _nameWordsByUser.remove(userId);
   }
+
+  List<String> wordsOfTopic(Topic topic) {
+    return _wordsOfTopics[topic.value] ??= topic.value.toLowerCase().split(' ');
+  }
+
+  void invalidateTopic(String value) {
```
gnprice, line 507: Is this needed? It doesn't seem to get called.

lib/model/autocomplete.dart
```diff
@@ -460,3 +522,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+  bool _isFetching = false;
+
+
+  /// Fetches topics of the current stream narrow, expected to fetch
+  /// only once per lifecycle.
+  ///
+  /// Starts fetching once the stream narrow is active, then when results
+  /// are fetched it notifies `autocompleteViewManager` to refresh UI
+  /// showing the newly fetched topics.
+  Future<void> fetch() async {
```
gnprice, line 547: Is this meant to be called from elsewhere? If not, it can be private.

lib/model/autocomplete.dart
```diff
@@ -460,3 +522,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+  Iterable<Topic> _topics = [];
+  bool _isFetching = false;
+
+
+  /// Fetches topics of the current stream narrow, expected to fetch
+  /// only once per lifecycle.
+  ///
+  /// Starts fetching once the stream narrow is active, then when results
+  /// are fetched it notifies `autocompleteViewManager` to refresh UI
+  /// showing the newly fetched topics.
```
gnprice, line 546: What does this text about "notifies `autocompleteViewManager`" correspond to? I don't see it in the implementation.

lib/model/autocomplete.dart
```diff
@@ -460,3 +522,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+    store.autocompleteViewManager.unregisterTopicAutocomplete(this);
+    super.dispose();
+  }
+}
+
+class TopicAutocompleteQuery extends AutocompleteQuery {
+  TopicAutocompleteQuery(super.raw);
+
+  bool testTopic(Topic topic, AutocompleteDataCache cache) {
+    return _testContainsQueryWords(cache.wordsOfTopic(topic));
```
gnprice, line 578: Is this the desired semantics for the filtering in the topic autocomplete? What did you look at to confirm that?

lib/model/autocomplete.dart
```diff
@@ -460,3 +522,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+  bool operator ==(Object other) {
+    return other is TopicAutocompleteQuery && other.raw == raw;
+  }
+
+  @override
+  int get hashCode => Object.hash('TopicAutocompleteQuery', raw);
+}
+
+class TopicAutocompleteResult extends AutocompleteResult {
+  final Topic topic;
```
gnprice, line 596: Does this need to use this Topic class, or can it just be a String?

Same question for [TopicAutocompleteView._topics], too.

gnprice (commented):

For this round, I've read through these three commits:
265ad97d5 autocomplete [nfc]: Rename Autocomplete to ComposeContentAutocomplete
7d9cc850f  autocomplete [nfc]: Factor out generic AutocompleteView
beaf4e24f autocomplete: Setup stream topics API binding

and part of these two:
7c24173d7 autocomplete [nfc]: Factor out generic AutocompleteField
b34f5c42a autocomplete: Implement targeted topic autocomplete

I think those latter two will be easier to read after acting on a couple of my comments below.

Khader-1:

gnprice (commented):
Thanks for the revision! Comments below. In this round I got through the whole PR.

lib/model/autocomplete.dart
```diff
@@ -43,6 +44,16 @@ extension Autocomplete on ComposeContentController {
   }
 }
 
+extension ComposeTopicAutocomplete on ComposeTopicController {
+  AutocompleteIntent<TopicAutocompleteQuery>? autocompleteIntent() {
+    // if (!selection.isValid || !selection.isNormalized) return null;
```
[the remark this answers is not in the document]
Khader-1: Oh I am Sorry!
This shouldn't have been committed nor checked in.. 
I will make sure to always self-review next time.

lib/model/autocomplete.dart
```diff
@@ -461,3 +514,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+    super.dispose();
+  }
+}
+
+class TopicAutocompleteQuery extends AutocompleteQuery {
+  TopicAutocompleteQuery(super.raw);
+
+  bool testTopic(String topic) => topic.isNotEmpty
+    && topic != raw
+    && topic.toLowerCase().contains(raw.toLowerCase());
```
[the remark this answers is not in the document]
Khader-1: Yes, this is a followup to [#627 (comment)](https://github.com/zulip/zulip-flutter/pull/627#discussion_r1697748918), unfortunately I could not load the comment when I worked on this revision due to the large number of comments which caused page to timeout without loading all comments..

And yes, I believe this copies what [zulip-mobile does](https://github.com/zulip/zulip-mobile/blob/main/src/autocomplete/TopicAutocomplete.js#L59)  and it also matches what [web does](https://github.com/zulip/zulip/blob/main/web/src/composebox_typeahead.ts#L1216)

lib/model/autocomplete.dart
```diff
@@ -461,3 +514,78 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+    if (_isFetching) return;
+    _isFetching = true;
+    final result = await getStreamTopics(store.connection, streamId: streamId);
+    _topics = result.topics.map((e) => e.name);
+    _isFetching = false;
+    if (_query != null) _startSearch(_query!);
+  }
+
+  @override
+  Iterable<String> getSortedItemsToTest(TopicAutocompleteQuery query) => _topics;
```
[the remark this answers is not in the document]
Khader-1: Thanks! I've [started a thread](https://chat.zulip.org#narrow/stream/412-api-documentation/topic/Does.20get-stream-topics.20returns.20a.20sorted.20list.20of.20topics.3F/near/1916151)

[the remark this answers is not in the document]
Khader-1: Anyways, I believe neither zulip-web nor zulip-mobile sorts the returned result.

lib/model/autocomplete.dart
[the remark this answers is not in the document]
gnprice: No need to apologize :slightly_smiling_face: — just good to fix it.

lib/model/autocomplete.dart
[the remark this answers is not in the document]
gnprice: Great, thanks for those references.

It looks like this does match what zulip-mobile does.

Let's drop the `isNotEmpty` condition — it doesn't make sense to me why that should be there. Looking in the Git history (`git log -L 59,61:src/autocomplete/TopicAutocomplete.js`), it was added in https://github.com/zulip/zulip-mobile/pull/1478 with no explanation. I suspect the point there was actually to check `x` wasn't `null` or `undefined`, rather than that it wasn't the empty string; back in 2017 that codebase didn't really have type-checking, so a null or undefined value could easily slip in where it didn't belong. (Even then, the right fix would have been to stop the unwanted value from entering in the first place, rather than paper over it like that commit did.)

Let's also drop the `topic != raw` condition. It doesn't seem to match the behavior of web.

Looking at web's code, it also has some reordering among the matches. Let's leave that for a future, post-launch issue, though. … OK, I've filed #881 for that — so we can leave here a comment like `// TODO(#881): sort by match relevance`.

lib/widgets/autocomplete.dart
```diff
@@ -140,50 +216,52 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
...
+    return TopicAutocompleteView.init(store: store, streamId: streamId);
+  }
+
+  void _onTapOption(BuildContext context, TopicAutocompleteResult option) {
+    final intent = autocompleteIntent();
+    if (intent == null) return;
+    final replacementString = option.topic;
+    controller.value = intent.textEditingValue.replaced(TextRange(
+      start: intent.syntaxStart,
+      end: intent.textEditingValue.text.length), replacementString);
```
gnprice, line 253: nit: `replacementString` gets kind of visually lost here, off after the end of a multi-line other expression

The formatting used in the similar existing method seems fine:
```suggestion
    controller.value = intent.textEditingValue.replaced(
      TextRange(
        start: intent.syntaxStart,
        end: intent.textEditingValue.text.length),
      replacementString,
    )
```

lib/model/autocomplete.dart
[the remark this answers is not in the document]
gnprice: (I think you already know this is still open, but bumping just so this remains visible in the new review)

lib/widgets/autocomplete.dart
```diff
@@ -140,50 +216,52 @@ class _ComposeAutocompleteState extends State<ComposeAutocomplete> with PerAccou
             Text(label),
           ])));
   }
+}
+
+class TopicAutocomplete extends AutocompleteField<TopicAutocompleteQuery, TopicAutocompleteResult, String> {
```
gnprice, line 221: nit in commit message:

    autocomplete: Implement targeted topic autocomplete
    
    Fixes: zulip#310

just `#310`, not `zulip#310`

gnprice (commented):
Thanks for the revision! This is quite close now. Just small comments below, and a couple of subthreads remain open from above — https://github.com/zulip/zulip-flutter/pull/627#discussion_r1712250710 as noted below, and https://github.com/zulip/zulip-flutter/pull/627#discussion_r1713064079 where I replied above.

lib/model/autocomplete.dart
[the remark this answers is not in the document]
Khader-1: I've [tried to verify this](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/Does.20get-stream-topics.20returns.20a.20sorted.20list.20of.20topics.3F/near/1917883). I think we can now use `_topics` assuming it is already sorted.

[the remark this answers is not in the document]
gnprice: Cool. Verifying the existing server's behavior isn't enough in itself, because that doesn't guarantee the server won't change behavior later… but that thread now also has Tim's agreement that we intend that as part of the API, and an intention to write it down in the docs. So that means we can rely on it for the future too.

lib/model/autocomplete.dart
```diff
@@ -500,3 +555,77 @@ class UserMentionAutocompleteResult extends MentionAutocompleteResult {
...
+    store.autocompleteViewManager.unregisterTopicAutocomplete(this);
+    super.dispose();
+  }
+}
+
+class TopicAutocompleteQuery extends AutocompleteQuery {
+  TopicAutocompleteQuery(super.raw);
+
+  bool testTopic(String topic) {
+    return topic != raw && topic.toLowerCase().contains(raw.toLowerCase());
```
gnprice, line 610: bump on one part of https://github.com/zulip/zulip-flutter/pull/627#discussion_r1713064079 : adding a `TODO(#881)` comment

gnprice: Also I added a few words to the commit message saying we picked web's behavior over zulip-mobile's, and linking to this subthread.

gnprice (commented):
Thanks for the revision! Everything looks good now, with a couple of nits below; I'll fix those and merge.

I also tweaked the main commit's summary line:

    -    autocomplete: Implement targeted topic autocomplete
    +    autocomplete: Add autocomplete for topic to send to

because "targeted" sounded to me like it was a kind of autocomplete that was somehow "targeted". I think what you intended was instead that the topic was "targeted", as in being the topic the draft message is aimed toward, so reworded to hopefully make that clearer.

PR #628, opened 2024-04-23
Upgrade Flutter and all packages to latest

chrisbobbe:
I'm actually optimistic that this will deal with

- #612

; for details see this commit:

```
deps: Upgrade packages within constraints (tools/upgrade pub)
```

Fixes: #612

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -3,14 +3,14 @@ PODS:
     - Flutter
   - device_info_plus (0.0.1):
     - Flutter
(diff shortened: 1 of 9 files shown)


gnprice:
Thanks @chrisbobbe for taking care of this!

All LGTM — merging.

It does sound like this probably fixes #612. We won't know for sure until we try uploading a build with it to TestFlight; when we do, if it turns out the issue isn't fixed, we can always reopen.

PR #629, opened 2024-04-24
text: Add and use proportionalLetterSpacing

chrisbobbe:
This could be helpful for #548, but I'm sending it now so we can see if we like the logic and want to reuse it for that. It does make a small adjustment to some letter spacing, to better follow the user's device-level text size setting, so I'll post some screenshots.

### DIFF
--- a/test/widgets/text_test.dart
+++ b/test/widgets/text_test.dart
@@ -6,6 +6,14 @@ import 'package:zulip/widgets/text.dart';
 
 import '../flutter_checks.dart';
 
+// From trying the options on an iPhone 13 Pro running iOS 16.6.1:
+const kTextScaleFactors = <double>[
+  0.8235, // smallest
+  1,
+  1.3529, // largest without using the "Larger Accessibility Sizes" setting
+  3.1176, // largest
+];
+
 void main() {
   group('zulipTypography', () {
     Future<Typography> getZulipTypography(WidgetTester tester, {
--- a/test/widgets/text_test.dart
+++ b/test/widgets/text_test.dart
@@ -1,4 +1,3 @@
-
 import 'package:checks/checks.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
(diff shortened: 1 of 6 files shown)


chrisbobbe:
| Before | After |
| --- | --- |
| EF238594-937F-4703-8D0B-8049472DF677 | 76FCBDD0-9BD3-41CC-9385-2A5D587D0457 |
| 8CFED067-CF29-43CC-B9DF-8FD558B252A7 | A0FA20FC-9A54-4854-9A2B-B820AF1F5F2A |
| 107AA3B7-BA63-49AE-9DBF-F61E9527305D | 3D8222EF-7172-4981-9600-A02203FA6E98 |
| 3F3FBCB9-6B86-40DC-A65C-1805E8C09B50 | F56C0324-9AB3-4298-8AC1-36C3C5487913 |

test/widgets/text_test.dart
```diff
@@ -207,4 +206,65 @@ void main() {
...
+              style: TextStyle(letterSpacing: getValue(context))))));
+
+        final TextStyle? style = tester.widget<Text>(find.byType(Text)).style;
+        final actualLetterSpacing = style!.letterSpacing!;
+        check((actualLetterSpacing - expected).abs()).isLessThan(0.0001);
+      });
+    }
+
+    // From trying the options on an iPhone 13 Pro running iOS 16.6.1:
+    const textScaleFactors = <double>[
```
gnprice, line 234: In the tests, the `textScaleFactors` list is copied from
    emoji_reaction_test. Perhaps there's a convenient place to define it
    centrally.

I think probably most convenient would be to put it at the top level of this file `text_test.dart`, and then `emoji_reaction_test.dart` can import from there.

It feels slightly messy to have one file be importing from another file that has its own `main` function; but it works fine, and it makes a low-fuss solution here.

gnprice (commented):
Thanks! LGTM; one suggestion below. Please go ahead and merge when ready.

chrisbobbe:
Thanks! Done.

PR #635, opened 2024-04-25
ui: Set letter spacing to 1% in buttons

chrisbobbe:
Stacked atop #629. Screenshots coming soon.

Fixes: #548

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -15,7 +15,7 @@ import 'page.dart';
 import 'recent_dm_conversations.dart';
 import 'store.dart';
 import 'subscription_list.dart';
-import 'text.dart';
+import 'theme.dart';
 
 class ZulipApp extends StatefulWidget {
   const ZulipApp({super.key, this.navigatorObservers});
@@ -109,34 +109,7 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
 
   @override
   Widget build(BuildContext context) {
-    final theme = ThemeData(
-      typography: zulipTypography(context),
-      appBarTheme: const AppBarTheme(
-        // Set these two fields to prevent a color change in [AppBar]s when
-        // there is something scrolled under it. If an app bar hasn't been
-        // given a backgroundColor directly or by theme, it uses
-        // ColorScheme.surfaceContainer for the scrolled-under state and
-        // ColorScheme.surface otherwise, and those are different colors.
-        scrolledUnderElevation: 0,
-        backgroundColor: Color(0xfff5f5f5),
-
-        shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
-      ),
-      // This applies Material 3's color system to produce a palette of
-      // appropriately matching and contrasting colors for use in a UI.
-      // The Zulip brand color is a starting point, but doesn't end up as
-      // one that's directly used.  (After all, we didn't design it for that
-      // purpose; we designed a logo.)  See docs:
-      //   https://api.flutter.dev/flutter/material/ColorScheme/ColorScheme.fromSeed.html
-      // Or try this tool to see the whole palette:
-      //   https://m3.material.io/theme-builder#/custom
-      colorScheme: ColorScheme.fromSeed(
-        seedColor: kZulipBrandColor,
-      ),
-      scaffoldBackgroundColor: const Color(0xfff6f6f6),
-      tooltipTheme: const TooltipThemeData(preferBelow: false),
-    );
-
+    final themeData = zulipThemeData(context);
     return GlobalStoreWidget(
       child: Builder(builder: (context) {
         final globalStore = GlobalStoreWidget.of(context);
@@ -146,7 +119,7 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
           title: 'Zulip',
           localizationsDelegates: ZulipLocalizations.localizationsDelegates,
           supportedLocales: ZulipLocalizations.supportedLocales,
-          theme: theme,
+          theme: themeData,
 
           navigatorKey: ZulipApp.navigatorKey,
           navigatorObservers: widget.navigatorObservers ?? const [],
@@ -180,12 +153,6 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
         }));
   }
 }
-
-/// The Zulip "brand color", a purplish blue.
-///
-/// This is chosen as the sRGB midpoint of the Zulip logo's gradient.
-// As computed by Anders: https://github.com/zulip/zulip-mobile/pull/4467
-const kZulipBrandColor = Color.fromRGBO(0x64, 0x92, 0xfe, 1);
 
 class ChooseAccountPage extends StatelessWidget {
   const ChooseAccountPage({super.key});
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -484,8 +484,10 @@ class MarkAsReadWidget extends StatelessWidget {
                   // [zulipTypography]…
                   Theme.of(context).textTheme.labelLarge!
                   // …then clobber some attributes to follow Figma:
-                  .merge(const TextStyle(
+                  .merge(TextStyle(
                     fontSize: 18,
+                    letterSpacing: proportionalLetterSpacing(context,
+                      kButtonTextLetterSpacingProportion, baseFontSize: 18),
                     height: (23 / 18))
                   .merge(weightVariableTextStyle(context, wght: 400))),
                 shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -1,0 +1,39 @@
+import 'package:flutter/material.dart';
+
+import 'text.dart';
+
+ThemeData zulipThemeData(BuildContext context) {
+  return ThemeData(
+    typography: zulipTypography(context),
+    appBarTheme: const AppBarTheme(
+      // Set these two fields to prevent a color change in [AppBar]s when
+      // there is something scrolled under it. If an app bar hasn't been
+      // given a backgroundColor directly or by theme, it uses
+      // ColorScheme.surfaceContainer for the scrolled-under state and
+      // ColorScheme.surface otherwise, and those are different colors.
+      scrolledUnderElevation: 0,
+      backgroundColor: Color(0xfff5f5f5),
+
+      shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
+    ),
+    // This applies Material 3's color system to produce a palette of
+    // appropriately matching and contrasting colors for use in a UI.
+    // The Zulip brand color is a starting point, but doesn't end up as
+    // one that's directly used.  (After all, we didn't design it for that
+    // purpose; we designed a logo.)  See docs:
+    //   https://api.flutter.dev/flutter/material/ColorScheme/ColorScheme.fromSeed.html
+    // Or try this tool to see the whole palette:
+    //   https://m3.material.io/theme-builder#/custom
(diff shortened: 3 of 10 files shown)


chrisbobbe:
Seen here with the "Mark all messages as read" button:

| Before | After |
| --- | --- |
| 91EBA500-2E48-4903-B4A8-EFCDD165211C | AC8EF83E-5881-419C-B936-019E39C0C007 |

chrisbobbe:
(There were some tests failing, and I also caught a bug when resolving those; pushed a fix.)

chrisbobbe:
Do reaction chips count as buttons? We can give 1% letter spacing to the label text (which shows either a number or the reacting users' names); that wouldn't be hard.

<details>
<summary>code</summary>

```diff
--- lib/widgets/emoji_reaction.dart
+++ lib/widgets/emoji_reaction.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/foundation.dart';
+++ lib/widgets/emoji_reaction.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:zulip/widgets/theme.dart';
 
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
@@ -153,6 +154,7 @@ class ReactionChip extends StatelessWidget {
                 //   which we learn about especially late).
                 final maxLabelWidth = (maxRowWidth - 6) * 0.75; // 6 is padding
 
+                final labelScaler = _labelTextScalerClamped(context);
                 return Row(
                   mainAxisSize: MainAxisSize.min,
                   crossAxisAlignment: CrossAxisAlignment.center,
@@ -168,9 +170,13 @@ class ReactionChip extends StatelessWidget {
                         constraints: BoxConstraints(maxWidth: maxLabelWidth),
                         child: Text(
                           textWidthBasis: TextWidthBasis.longestLine,
-                          textScaler: _labelTextScalerClamped(context),
+                          textScaler: labelScaler,
                           style: TextStyle(
                             fontSize: (14 * 0.90),
+                            letterSpacing: proportionalLetterSpacing(context,
+                              kButtonTextLetterSpacingProportion,
+                              baseFontSize: (14 * 0.90),
+                              textScaler: labelScaler),
                             height: 13 / (14 * 0.90),
                             color: labelColor,
                           ).merge(weightVariableTextStyle(context,
```

<details>

lib/widgets/app.dart
```diff
@@ -149,7 +118,7 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
           title: 'Zulip',
           localizationsDelegates: ZulipLocalizations.localizationsDelegates,
           supportedLocales: ZulipLocalizations.supportedLocales,
-          theme: theme,
+          theme: outerThemeData(context),
```
gnprice, line 121: The commit message says:

    theme [nfc]: Pull `outerThemeData` out from app.dart to new theme.dart file

but this makes another change too: it computes this ThemeData from within the builder callback for `GlobalStoreWidget.child`. Is that intended?

If the method computing this theme data isn't going to be trying to use the store, then it's probably cleanest to keep the computation outside there, with e.g. `final theme = outerThemeData(context);` at the top of the build method. That way it's clear we compute it once eagerly.

lib/widgets/theme.dart
```diff
@@ -0,0 +1,80 @@
...
+//   Then callers could call a hypothetical ThemeData.merge method with it.
+//   But such a method doesn't exist yet:
+//     https://github.com/flutter/flutter/issues/43823
+ThemeData innerThemeData(BuildContext context) {
+  final theme = Theme.of(context);
+  assert(theme.textTheme.labelLarge!.debugLabel!.contains('zulipTypography'));
+  return theme.copyWith(
+    elevatedButtonTheme:  ElevatedButtonThemeData( style: _commonButtonStyle(context)),
+    filledButtonTheme:    FilledButtonThemeData(   style: _commonButtonStyle(context)),
+    iconButtonTheme:      IconButtonThemeData(     style: _commonButtonStyle(context)),
```
gnprice, line 57: Seems cleanest to factor this common style out as a local — that way we only compute it once.

lib/widgets/theme.dart
```diff
@@ -0,0 +1,80 @@
...
+    menuButtonTheme:      MenuButtonThemeData(     style: _commonButtonStyle(context)),
+    outlinedButtonTheme:  OutlinedButtonThemeData( style: _commonButtonStyle(context)),
+    segmentedButtonTheme: SegmentedButtonThemeData(style: _commonButtonStyle(context)),
+    textButtonTheme:      TextButtonThemeData(     style: _commonButtonStyle(context)),
+  );
+}
+
+double kButtonTextLetterSpacingProportion = 0.01;
```
gnprice, line 65:
```suggestion
const kButtonTextLetterSpacingProportion = 0.01;
```

lib/widgets/theme.dart
```diff
@@ -0,0 +1,80 @@
+import 'package:flutter/material.dart';
+
+import 'text.dart';
+
+ThemeData outerThemeData(BuildContext context) {
```
gnprice, line 5: Is there a coherent description we can give for what the meaning of this "outer ThemeData" is?

It seems like it's basically an intermediate step in the calculation of the "inner ThemeData", which is the theme data we actually want to use in the app.

In that case, can we have one function that just computes the desired ThemeData start to finish, from a single BuildContext? I think that'd make it easier to reason about how the caller should use the function. That function could even have this intermediate step as a ThemeData local variable, if that's helpful.

test/widgets/autocomplete_test.dart
```diff
@@ -48,18 +47,14 @@ Future<Finder> setupToComposeInput(WidgetTester tester, {
...
-          child: MessageListPage(
-            narrow: DmNarrow(
-              allRecipientIds: [eg.selfUser.userId, eg.otherUser.userId],
-              selfUserId: eg.selfUser.userId,
-            ))))));
+  await tester.pumpWidget(const ZulipApp());
+  await tester.pump();
+  final navigator = await ZulipApp.navigator;
+  navigator.push(MessageListPage.buildRoute(accountId: eg.selfAccount.id,
+    narrow: DmNarrow(
```
gnprice, line 54: These rearrangements of existing tests are probably clearest as a separate prep commit.

test/widgets/autocomplete_test.dart
```diff
@@ -48,18 +47,14 @@ Future<Finder> setupToComposeInput(WidgetTester tester, {
...
+      allRecipientIds: [eg.selfUser.userId, eg.otherUser.userId],
+      selfUserId: eg.selfUser.userId,
+    )));
 
   // global store, per-account store, and message list get loaded
   await tester.pumpAndSettle();
```
gnprice, line 60: This comment is no longer accurate. Perhaps just cut it:
```suggestion
    )));
  await tester.pumpAndSettle();
```

And perhaps it can be reduced to just a `pump`?

lib/widgets/message_list.dart
```diff
@@ -480,12 +486,13 @@ class MarkAsReadWidget extends StatelessWidget {
                 backgroundColor: _UnreadMarker.color,
                 minimumSize: const Size.fromHeight(38),
                 textStyle:
-                  // Restate [FilledButton]'s default, which inherits from
-                  // [zulipTypography]…
-                  Theme.of(context).textTheme.labelLarge!
+                  // Restate [FilledButton]'s default…
+                  themeData.textTheme.labelLarge!.merge(filledButtonThemeTextStyle)
```
gnprice, line 490: Huh.

Does this mean we could just set `.textTheme.labelLarge` on our main ThemeData, and skip reaching into all the individual button styles? That seems like it'd be cleaner.

In principle that has a slightly broader effect than specified in #548, in that it affects any other Material widgets that use the `labelLarge` text style. But fundamentally those are widgets that the Material designers intended to have a similar text style to that of buttons (even though they're not called "buttons"), so it's likely that that'll be the right answer anyway.

If we end up with such a widget where Vlad wants a different spacing, we can address that then… and I think it's fairly likely that in that case he'd want something that wasn't the Material widget in any event.

lib/widgets/app.dart
```diff
@@ -109,37 +109,6 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
 
   @override
   Widget build(BuildContext context) {
-    final theme = ThemeData(
```
gnprice, line 112: Cool, good to get this pulled out of `ZulipApp`.

gnprice (commented):
Thanks @chrisbobbe! Comments below.

lib/widgets/app.dart
chrisbobbe: Indeed; thanks for noticing this!

lib/widgets/message_list.dart
chrisbobbe: That's a good idea! Done.

chrisbobbe:
Thanks for the review! Revision pushed, with the simpler approach that you outlined in https://github.com/zulip/zulip-flutter/pull/635#discussion_r1581591462 and that we talked about in the office.

test/widgets/theme_test.dart
```diff
@@ -0,0 +1,78 @@
...
+        double? expectedLetterSpacing;
+        await tester.pumpWidget(
+          Builder(builder: (context) => MaterialApp(
+            theme: zulipThemeData(context),
+            home: Builder(builder: (context) {
+              expectedFontSize = Theme.of(context).textTheme.labelLarge!.fontSize!;
+              expectedLetterSpacing = proportionalLetterSpacing(context,
+                0.01, baseFontSize: expectedFontSize!);
+              return Builder(builder: (context) =>
+                buttonBuilder(context, buttonText));
```
gnprice, line 33:
```suggestion
              return buttonBuilder(context, buttonText);
```

This would have the same effect, right? It doesn't seem like this build context is effectively any different from its parent.

test/widgets/theme_test.dart
```diff
@@ -0,0 +1,78 @@
...
+      double? ambientTextScaleFactor,
+    }) async {
+      testWidgets(description, (WidgetTester tester) async {
+        if (ambientTextScaleFactor != null) {
+          tester.platformDispatcher.textScaleFactorTestValue = ambientTextScaleFactor;
+          addTearDown(tester.platformDispatcher.clearTextScaleFactorTestValue);
+        }
+        const buttonText = 'Zulip';
+        double? expectedFontSize;
+        double? expectedLetterSpacing;
```
gnprice, line 24: Can tighten up expectations here slightly:
```suggestion
        late final double expectedFontSize;
        late final double expectedLetterSpacing;
```

That way it's clear (and enforced) that these only get set once, and that they always get set before they're read.

test/widgets/theme_test.dart
```diff
@@ -0,0 +1,78 @@
...
+        check(text.style!)
+          ..fontSize.equals(expectedFontSize)
+          ..letterSpacing.equals(expectedLetterSpacing);
+      });
+    }
+
+    doCheck('with device text size adjusted',
+      ambientTextScaleFactor: 2.0,
+      buttonBuilder: (context, text) => ElevatedButton(onPressed: () {},
+        child: Text(text)));
```
gnprice, line 46: It seems like none of these builders actually use the context. So they can be simplified to just widgets (by pulling up the `buttonText` constant too):
```suggestion
      button: ElevatedButton(onPressed: () {}, child: const Text(buttonText)));
```

gnprice (commented):
Thanks! This all looks good; just small comments below.

(I'll review #629 separately, but it also looks good modulo small comments.)

test/widgets/theme_test.dart
chrisbobbe: Ah right, cool!

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! LGTM; merging.

PR #637, opened 2024-04-26
ide: Include .idea/zulip.iml

chrisbobbe:
This has some useful `excludeFolder` items that should help Android Studio see that certain files are not part of our project and should be excluded in searches that are supposed to be limited to "project files".

### DIFF
--- a/.idea/zulip.iml
+++ b/.idea/zulip.iml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/.pub" />
+      <excludeFolder url="file://$MODULE_DIR$/build" />
+      <excludeFolder url="file://$MODULE_DIR$/ios/.symlinks" />
+      <excludeFolder url="file://$MODULE_DIR$/linux/flutter/ephemeral" />
+      <excludeFolder url="file://$MODULE_DIR$/macos/Flutter/ephemeral" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/.pub" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/build" />
+      <excludeFolder url="file://$MODULE_DIR$/tools/icons/node_modules" />
+      <excludeFolder url="file://$MODULE_DIR$/windows/flutter/ephemeral" />
+    </content>
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="Dart SDK" level="project" />
+    <orderEntry type="library" name="Flutter Plugins" level="project" />
+    <orderEntry type="library" name="Dart Packages" level="project" />
+  </component>
+</module>
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ migrate_working_dir/
 /.idea/*
 /android/.idea/*
 !/.idea/inspectionProfiles/
+!/.idea/modules.xml
+!/.idea/zulip.iml
 !/android/.idea/inspectionProfiles/
 
 # The .vscode folder contains launch configuration and tasks you configure in
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/zulip.iml" filepath="$PROJECT_DIR$/.idea/zulip.iml" />
+    </modules>
+  </component>
+</project>
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,3 +1,7 @@
+<!-- To view the "merged manifest" in Android Studio,
+     open the android/ subtree as its own project
+     and follow these instructions:
+       https://developer.android.com/build/manage-manifests#inspect_the_merged_manifest_and_find_conflicts -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
    <uses-permission android:name="android.permission.INTERNET"/>
    <application


.idea/zulip-flutter.iml
```diff
@@ -0,0 +1,23 @@
...
+      <excludeFolder url="file://$MODULE_DIR$/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/.pub" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/build" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/macos/Flutter/ephemeral" />
+      <excludeFolder url="file://$MODULE_DIR$/ios/.symlinks" />
+      <excludeFolder url="file://$MODULE_DIR$/linux/flutter/ephemeral" />
+      <excludeFolder url="file://$MODULE_DIR$/windows/flutter/ephemeral" />
+    </content>
+    <orderEntry type="inheritedJdk" />
```
gnprice, line 17: This element is absent in the file on my home desktop. I'm not sure what it means.

gnprice (commented):
Thanks for sending this!

One complication in trying to merge something like this for general use is that Android Studio seems to pick a variety of names for this `.iml` file.
* For you, it's `.idea/zulip-flutter.iml`.
* For me on my desktop in the office, it's `.idea/flutter.iml`.

From those two, it seemed like the pattern was that it uses the basename of the project's parent directory — your clone of this repo is in a directory whose name ends in `…/zulip-flutter`, and mine is at `~/z/flutter`. (Generally I leave the `zulip-` prefix off of these directories' names, to keep down the amount of typing or tab-completion to get to them. Similarly I have my zulip-mobile tree at `~/z/mobile`.)

But then the wrinkles continue:

* For me on my main desktop, at home, the file is at `zulip.iml` — outside the `.idea/` directory, at the root of the tree.

  (Meanwhile this tree is at `~/z/flutterz` — same idea as above, but I also wanted it to have a different basename from my checkout of Flutter itself, because parts of the Android Studio UI identify a project by just the basename of its directory.)

I don't know why that one's in a different place.

… Aha — it's because `flutter create` put it there, back when I began the prototype. See the template file `packages/flutter_tools/templates/app_shared/projectName.iml.tmpl` in the Flutter upstream tree.

So, maybe we can put the file there and it'll work? The experiment to try would be:
* Exit Android Studio.
* Move your `.idea/zulip-flutter.iml` to just `zulip.iml`.
* Re-open Android Studio.
* Confirm that it's using the file from its new location: that it hasn't made any new `.iml` file, and that the excluded folders are excluded.

chrisbobbe:
> So, maybe we can put the file [at the root of the tree] and it'll work? The experiment to try would be:
> 
> * Exit Android Studio.
> * Move your `.idea/zulip-flutter.iml` to just `zulip.iml`.
> * Re-open Android Studio.
> * Confirm that it's using the file from its new location: that it hasn't made any new `.iml` file, and that the excluded folders are excluded.

I *think* I've gotten this to work, but I think an `.idea/modules.xml` file is involved in making it work at the new location. Here is the content of that file when it works:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="ProjectModuleManager">
    <modules>
      <module fileurl="file://$PROJECT_DIR$/zulip.iml" />
    </modules>
  </component>
</project>

```

Should we commit that file too, and if so, is it OK for it to stay in `.idea/` or do we think it's better to move it out to the root of the tree?

Is there a reason to prefer moving the `.iml` file out to the root? I think it's working for me even if the `.iml` file is `.idea/zulip.iml`. And when it's there—even with that name `zulip.iml` instead of `zulip-flutter.iml`—it seems to work even in the absence of an `.idea/modules.xml` file.

.idea/zulip-flutter.iml
chrisbobbe: Yeah, I'm not sure either. Maybe leave it out, and if we need it we'll eventually find out?

gnprice:
> I think it's working for me even if the `.iml` file is `.idea/zulip.iml`. And when it's there—even with that name `zulip.iml` instead of `zulip-flutter.iml`—it seems to work even in the absence of an `.idea/modules.xml` file.

Sure, let's try that, then.


> an `.idea/modules.xml` file

FWIW I have such a file too, and it looks like this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="ProjectModuleManager">
    <modules>
      <module fileurl="file://$PROJECT_DIR$/zulip.iml" filepath="$PROJECT_DIR$/zulip.iml" />
      <module fileurl="file://$PROJECT_DIR$/android/zulip_android.iml" filepath="$PROJECT_DIR$/android/zulip_android.iml" />
    </modules>
  </component>
</project>
```

The timestamp looks like when I started the prototype. There's a corresponding template in the Flutter tree, which fits.

The `android/zulip_android.iml` file is another one I have locally, that `flutter create` created. After we have this `.iml` file for the main project merged, I may follow up by dealing with that one; it's a lower priority because it's much less common for us to open the `android/` subtree as its own project (and indeed I'm not sure anyone other than me ever has).

chrisbobbe:
> it's much less common for us to open the `android/` subtree as its own project (and indeed I'm not sure anyone other than me ever has).

I've found this very helpful when I want to use Android Studio's ["merged manifest" feature](https://developer.android.com/build/manage-manifests#inspect_the_merged_manifest_and_find_conflicts). The "Merged Manifest" tab isn't offered when you open the project from the root, but it _is_ offered when you open the `android/` subtree as its own project. I remember searching online for why the "Merged Manifest" tab wasn't there and not finding anything helpful. Maybe worth pointing out this solution in a comment at the top of `android/app/src/main/AndroidManifest.xml`? I could do that.

chrisbobbe:

gnprice:
Thanks! Pushed to the branch a couple of changes, based on testing on my office desktop just now and our in-person discussion.

gnprice:
Also tried that updated branch on my laptop, and it seems to work smoothly there too.

chrisbobbe:
Thanks for the review, testing, and updates to the branch! Revision pushed, incorporating those proposed changes.

.idea/zulip.iml
```diff
@@ -0,0 +1,22 @@
...
+      <excludeFolder url="file://$MODULE_DIR$/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/.pub" />
+      <excludeFolder url="file://$MODULE_DIR$/build" />
+      <excludeFolder url="file://$MODULE_DIR$/ios/.symlinks" />
+      <excludeFolder url="file://$MODULE_DIR$/linux/flutter/ephemeral" />
+      <excludeFolder url="file://$MODULE_DIR$/macos/Flutter/ephemeral" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/.pub" />
+      <excludeFolder url="file://$MODULE_DIR$/packages/zulip_plugin/build" />
+      <excludeFolder url="file://$MODULE_DIR$/scripts/icons/node_modules" />
```
gnprice, line 15: Ha, what this should actually be (contra https://github.com/zulip/zulip-flutter/pull/637#discussion_r1583753412 ) is:

```suggestion
      <excludeFolder url="file://$MODULE_DIR$/tools/icons/node_modules" />
```

There is no `scripts/`; that must have been… aha, it's what `tools/` was called before e5d1e3bf6f5b8c8cfc931282ac8f15ab4289670e.

If we'd had this file in those versions of the tree, we'd want it to say `scripts/`; but we didn't, and at this point it should say `tools/`.

---

My local `zulip.iml` has both; it was just hard to spot with all the noise from `.dart_tool`, `.pub`, and `build` in every plugin package. To make sure there's nothing else hiding there, here it is with all the noise stripped out:
```
$ grep -vFe '/.dart_tool"' -e '/.pub"' -e '/build"' zulip.iml 
<?xml version="1.0" encoding="UTF-8"?>
<module type="JAVA_MODULE" version="4">
  <component name="NewModuleRootManager" inherit-compiler-output="true">
    <exclude-output />
    <content url="file://$MODULE_DIR$">
      <excludeFolder url="file://$MODULE_DIR$/scripts/icons/node_modules" />
      <excludeFolder url="file://$MODULE_DIR$/linux/flutter/ephemeral" />
      <excludeFolder url="file://$MODULE_DIR$/tools/icons/node_modules" />
      <excludeFolder url="file://$MODULE_DIR$/windows/flutter/ephemeral" />
    </content>
    <orderEntry type="sourceFolder" forTests="false" />
    <orderEntry type="library" name="Dart SDK" level="project" />
    <orderEntry type="library" name="Flutter Plugins" level="project" />
    <orderEntry type="library" name="Dart Packages" level="project" />
  </component>
</module>
```

So, looks like everything else there is covered in this revision. There is this line:
```xml
    <orderEntry type="library" name="Dart Packages" level="project" />
```
But I can't find anything missing as a consequence of leaving that line out. For example…

… hmm no. I tried a search and found it did get results from "Dart Packages":
image

But then I found that that entry had been added back in somehow:
```diff
--- .idea/zulip.iml
+++ .idea/zulip.iml
@@ -18,5 +18,6 @@
     <orderEntry type="sourceFolder" forTests="false" />
     <orderEntry type="library" name="Dart SDK" level="project" />
     <orderEntry type="library" name="Flutter Plugins" level="project" />
+    <orderEntry type="library" name="Dart Packages" level="project" />
   </component>
 </module>
\ No newline at end of file
```

I discarded that with `git reset --hard`, tried again, and now I'm only seeing the results that can also be found under "Flutter Plugins":
image

Similarly try searching for "Table" (a class from `package:drift`); it doesn't appear.

Or in fact something from the Flutter framework, like "Widget".

So yeah, that's a line we want to include. :slightly_smiling_face:

gnprice (commented):
Thanks! Comments below. One minor line to change… and then one boring-looking line that's missing that turns out to be important.

I think this is very close, though — I'm hoping that after those two tweaks, it'll be ready.

chrisbobbe:
Cool, thanks! Revision pushed.

Android Studio wanted to add back a newline at the end of the file; do we want to keep that? I think we would usually include the newline.

gnprice:
> Android Studio wanted to add back a newline at the end of the file; do we want to keep that? I think we would usually include the newline.

Great, yeah. This is a file it's inevitably going to keep modifying at its whim, so whatever it wants to automatically do we'll follow (because it'd be a big pain to try to do otherwise); glad the answer happens to be to include the final newline.

gnprice:
Thanks for all the revisions on this! Looks good; merging, with a small commit-message tweak mainly to mention the `modules.xml` in the summary line as well as the `zulip.iml`.

Issue #620, opened 2024-04-05
Web auth sometimes fails


gnprice:
See detailed report [in chat](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1775625).

Some potentially key details include:
* This was on Android.
* The browser which powered the webview was Firefox for Android.
* After redirecting to the app:
  > When I am sent back to the app, it just shows the main "Choose account"//"Add an account" page. There is no error message

chrisbobbe:
Responded in the linked chat thread with some thoughts, including a potential solution that I want to check with Greg when he's back from vacation.

PR #639, opened 2024-04-30
android: Set launchMode="singleTask" instead of "singleTop"

chrisbobbe:
This addresses #620 by making it so the whole app state isn't reset when coming back from Firefox in the web-auth flow. Discussion:
  https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1776828

Later in that discussion, Greg found that this mode is the most appropriate for our app, and that setting taskAffinity="" is probably a good idea as a security-hardening measure.

Fixes: #620

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -7,7 +7,7 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
             android:taskAffinity=""
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -8,6 +8,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:launchMode="singleTop"
+            android:taskAffinity=""
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
             android:networkSecurityConfig="@xml/network_security_config"


gnprice:
Thanks for debugging and fixing this!

Let's make this two commits: first `taskAffinity`, which is an independent hardening measure, then `launchMode`. Otherwise all LGTM.

chrisbobbe:
Cool; done, PTAL.

gnprice:
Thanks! Looks good; merging, with two commit-message tweaks. One is a minor formatting change to make a paragraph wrap better, and the other a small correction or clarification:

```
$ git range-diff --creation-factor=90 origin pr/639 @
1:  7588b5a47 ! 1:  138b1e28b android: Set taskAffinity=""
    @@ Commit message
         android: Set taskAffinity=""
     
         This looks like a security-hardening measure that would be good to
    -    do, independently of the upcoming change to
    -    `launchMode="singleTask"`. Discussion:
    +    do, independently of the upcoming change to launchMode="singleTask".
    +    Discussion:
           https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1790071
     
      ## android/app/src/main/AndroidManifest.xml ##
2:  1e9a27d1f ! 2:  1251a0b06 android: Set launchMode="singleTask" instead of "singleTop"
    @@ Commit message
         when coming back from Firefox in the web-auth flow. Discussion:
           https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1776828
         Later in that discussion, Greg found that this mode is the most
    -    appropriate for our app.
    +    appropriate for the main activity of our app:
    +      https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1790019

         Fixes: #620
```

In particular the key about `singleTask` is that it's appropriate for the "main" or "root" activity of an app. We have some ancillary activities in our merged manifest, like this one (which I guess would get used if we passed [LaunchMode.inAppWebView] to `launchUrl`):
```xml
        <activity
            android:theme="@ref/0x1030007"
            android:name="io.flutter.plugins.urllauncher.WebViewActivity"
            android:exported="false" />
```

and those wouldn't be appropriate for `android:launchMode="singleTask"`.

chrisbobbe:
I see; interesting. Thanks!

PR #642, opened 2024-05-03
model [nfc]: Remove StreamColorSwatch; just use ColorSwatch<StreamColorVariant>

chrisbobbe:
This indirection was mildly helpful; it let us consume these colors
with slightly more concise code, and without needing `!` to satisfy
the analyzer.

But we get the same correct functionality without the indirection,
and a bit more transparently. The more substantive benefit of using
ColorSwatch directly is that we now have a ready-made lerp function
-- `ColorSwatch.lerp` -- that can operate on our swatch instances.
That should be helpful later when we want stream colors to animate
smoothly when dark mode is toggled, along with various other style
attributes that will naturally animate too.

-----

Quoting that [`ColorSwatch.lerp`](https://api.flutter.dev/flutter/painting/ColorSwatch/lerp.html):

```dart
  /// Linearly interpolate between two [ColorSwatch]es.
  ///
  /// It delegates to [Color.lerp] to interpolate the different colors of the
  /// swatch.
  ///
  /// If either color is null, this function linearly interpolates from a
  /// transparent instance of the other color.
  ///
  /// The `t` argument represents position on the timeline, with 0.0 meaning
  /// that the interpolation has not started, returning `a` (or something
  /// equivalent to `a`), 1.0 meaning that the interpolation has finished,
  /// returning `b` (or something equivalent to `b`), and values in between
  /// meaning that the interpolation is at the relevant point on the timeline
  /// between `a` and `b`. The interpolation can be extrapolated beyond 0.0 and
  /// 1.0, so negative values and values greater than 1.0 are valid (and can
  /// easily be generated by curves such as [Curves.elasticInOut]). Each channel
  /// will be clamped to the range 0 to 255.
  ///
  /// Values for `t` are usually obtained from an [Animation<double>], such as
  /// an [AnimationController].
  static ColorSwatch<T>? lerp<T>(ColorSwatch<T>? a, ColorSwatch<T>? b, double t) {
    if (identical(a, b)) {
      return a;
    }
    final Map<T, Color> swatch;
    if (b == null) {
      swatch = a!._swatch.map((T key, Color color) => MapEntry<T, Color>(key, Color.lerp(color, null, t)!));
    } else {
      if (a == null) {
        swatch = b._swatch.map((T key, Color color) => MapEntry<T, Color>(key, Color.lerp(null, color, t)!));
      } else {
        swatch = a._swatch.map((T key, Color color) => MapEntry<T, Color>(key, Color.lerp(color, b[key], t)!));
      }
    }
    return ColorSwatch<T>(Color.lerp(a, b, t)!.value, swatch);
  }
```

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -469,22 +469,10 @@ class StreamColorSwatch extends ColorSwatch<_StreamColorVariant> {
 
   Color get unreadCountBadgeBackground => this[_StreamColorVariant.unreadCountBadgeBackground]!;
 
-  /// The stream icon on a plain-colored surface, such as white.
-  ///
-  /// For the icon on a [barBackground]-colored surface,
-  /// use [iconOnBarBackground] instead.
   Color get iconOnPlainBackground => this[_StreamColorVariant.iconOnPlainBackground]!;
 
-  /// The stream icon on a [barBackground]-colored surface.
-  ///
-  /// For the icon on a plain surface, use [iconOnPlainBackground] instead.
-  /// This color is chosen to enhance contrast with [barBackground]:
-  ///   <https://github.com/zulip/zulip/pull/27485>
   Color get iconOnBarBackground => this[_StreamColorVariant.iconOnBarBackground]!;
 
-  /// The background color of a bar representing a stream, like a recipient bar.
-  ///
-  /// Use this in the message list, the "Inbox" view, and the "Streams" view.
   Color get barBackground => this[_StreamColorVariant.barBackground]!;
 
   static Map<_StreamColorVariant, Color> _compute(int base) {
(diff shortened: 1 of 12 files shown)


gnprice:
Hmm, it does feel like the code that's using this class generally all gets worse this way.

For the lerp, how about we just copy that method's implementation? It doesn't seem like it's doing anything all that complicated.

In order to have the map lying around, we could save it on our own private field in addition to passing it to the superclass. The redundancy is mildly annoying, but it's only duplicating one pointer, so it's not a material cost at runtime.

chrisbobbe:
Sure, makes sense. Since that approach will be pretty different from what's in this PR, I'll close this PR and open a new one.

Issue #95, opened 2023-05-18
ui: Implement dark theme


chrisbobbe:
The user should be able to choose from these three options for the app's appearance:

- Light
- Dark
- Follow OS

Flutter's Material Design implementation paves the way to a pretty good 1.0 version of a dark theme, for the boring, simple bits of UI we paint with `Theme.of(context).colorScheme`. For example, here's the result of just this simple change:

```diff
diff --git lib/widgets/app.dart lib/widgets/app.dart
index 533d3df8d..2eec13d40 100644
--- lib/widgets/app.dart
+++ lib/widgets/app.dart
@@ -20,7 +20,7 @@ class ZulipApp extends StatelessWidget {
       //   https://api.flutter.dev/flutter/material/ColorScheme/ColorScheme.fromSeed.html
       // Or try this tool to see the whole palette:
       //   https://m3.material.io/theme-builder#/custom
-      colorScheme: ColorScheme.fromSeed(seedColor: kZulipBrandColor));
+      colorScheme: ColorScheme.fromSeed(seedColor: kZulipBrandColor, brightness: Brightness.dark));
     return GlobalStoreWidget(
       child: MaterialApp(
         title: 'Zulip',
```

| Before | After |
| --- | --- |
| ![image](https://github.com/zulip/zulip-flutter/assets/22248748/b9bd52c3-b490-43ba-907a-a91cb5f8d9e0) | ![image](https://github.com/zulip/zulip-flutter/assets/22248748/89ef5717-6343-4786-86ac-ca9f169f0057) |

For the message list, though, we don't draw from `Theme.of(context).colorScheme`, so it doesn't respond to setting `Brightness.dark` there:

| Before | After |
| --- | --- |
| ![image](https://github.com/zulip/zulip-flutter/assets/22248748/5519929e-cc4c-4e81-aaca-5509920bea69) | ![image](https://github.com/zulip/zulip-flutter/assets/22248748/528760b8-dba2-4680-b00e-6893fba64dcb) |

Probably we should bundle all the message list's light-theme colors into a palette called `messageListColors` or something, then create a version of that palette for dark theme.

gnprice:
If we do this before we have local settings #97, we can have it just follow the system theme — that's what should be the default anyway, so it may as well become the only setting in the alpha. If we already have #97 when we implement this, though, might as well offer the setting from the beginning.

neeldoshii:
Can I work on this issue @gnprice

gnprice:
[...]

PR #643, opened 2024-05-04
model: Implement dark-mode stream color swatches

chrisbobbe:
Stacked atop ~~#642~~ #654.

-----

We added the light-mode color computations in #381.

We're not ready to use these yet, but the color computations are
unlikely to change before that time comes, and finding the right
ones is a chore that's good to get done.

Related: #95

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -468,12 +468,6 @@ class StreamColorSwatch extends ColorSwatch<StreamColorVariant> {
 
   const StreamColorSwatch._(int base, this._swatch) : super(base, _swatch);
 
-  /// A [StreamColorSwatch], from a [Map<_StreamColorVariant, Color>]
-  /// written manually.
-  @visibleForTesting
-  const StreamColorSwatch.debugFromBaseAndSwatch(int base, swatch)
-    : this._(base, swatch);
-
   final Map<StreamColorVariant, Color> _swatch;
 
   /// The [Subscription.color] int that the swatch is based on.
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -463,7 +463,8 @@ class Subscription extends ZulipStream {
 /// Use this in UI code for colors related to [Subscription.color],
 /// such as the background of an unread count badge.
 class StreamColorSwatch extends ColorSwatch<StreamColorVariant> {
-  StreamColorSwatch(int base) : this._(base, _compute(base));
+  StreamColorSwatch(int base) : this._(base, _computeLight(base));
+  StreamColorSwatch.dark(int base) : this._(base, _computeDark(base));
 
   const StreamColorSwatch._(int base, this._swatch) : super(base, _swatch);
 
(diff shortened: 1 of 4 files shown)


chrisbobbe:
Here is how I generated the "expected" values to put in the tests. (This comment is adapted from https://github.com/zulip/zulip-flutter/pull/381#discussion_r1399319296 ; this time for dark mode.)

Here's the code I added in the [replit](https://replit.com/@VladKorobov/zulip-topic-feed-colors#script.js), just under `ZULIP_ASSIGNMENT_COLORS`. It duplicates some of the logic, where it wasn't exposed with a convenient helper function. But it can be checked against the implementation. → Actually, one color follows the web app instead of the replit; I've noted that in the code and also in the snippet below.

<details>

```dart
const dartColorIntFromHexString = (hexString) => {
  const lowercaseWithoutHash = hexString.substring(1).toLowerCase();
  switch (lowercaseWithoutHash.length) {
    case 6: // opaque (alpha omitted)
      return `0xff${lowercaseWithoutHash}`;
    case 8: // not opaque
      // convert rrggbbaa to aarrggbb
      return '0x'
        + lowercaseWithoutHash.substring(6, 8)
        + lowercaseWithoutHash.substring(0, 2);
    default:
      throw new Error();
  }
};

const dartCheckLine = (baseHex, expectedHex) => {
  return `runCheck(${dartColorIntFromHexString(baseHex)}, const Color(${dartColorIntFromHexString(expectedHex)}));\n`;
};

let iconOnPlainBackgroundResult = '';
let iconOnBarBackgroundResult = '';
let barBackgroundResult = '';
ZULIP_ASSIGNMENT_COLORS.forEach(color => {
  const correctedColor = correctStreamColor({ color });
  const barBackground = getRecipientBarColor({ color: correctStreamColor({ color }), darkMode: true });
  const iconOnPlainBackground = correctedColor;
  const iconOnBarBackground = correctedColor; // follows web, not replit

  iconOnPlainBackgroundResult += dartCheckLine(color, iconOnPlainBackground);
  iconOnBarBackgroundResult += dartCheckLine(color, iconOnBarBackground);
  barBackgroundResult += dartCheckLine(color, barBackground);
});

console.log(iconOnPlainBackgroundResult);
console.log(iconOnBarBackgroundResult);
console.log(barBackgroundResult);
```

</details>

lib/api/model/model.dart
```diff
@@ -462,86 +463,142 @@ class Subscription extends ZulipStream {
...
+
+    // Follows `.recepient` in Vlad's replit:
+    //   <https://replit.com/@VladKorobov/zulip-topic-feed-colors#script.js>
+    //
+    // TODO I think [LabColor.interpolate] doesn't actually do LAB mixing;
+    //   it just calls up to the superclass method [ColorModel.interpolate]:
+    //     <https://pub.dev/documentation/flutter_color_models/latest/flutter_color_models/ColorModel/interpolate.html>
+    //   which does ordinary RGB mixing. Investigate and send a PR?
+    // TODO fix bug where our results differ from the replit's (see unit tests)
+    StreamColor.barBackground:
```
gnprice, line 571: This comment is all the same as in the light case, right?

Let's try to deduplicate comments that don't differ. It makes it easier to find the spots where there is new information in these comments (after reading the other version's comments), like for `iconOnBarBackground`.

gnprice (commented):
Thanks! Just small comments below. (Modulo the review I just left on #642.)

gnprice:
Thanks for the revision! LGTM — merging.

Issue #634, opened 2024-04-24
Rename "All messages" to "Combined feed"


gnprice:
This is the zulip-flutter counterpart of:
* https://github.com/zulip/zulip-mobile/issues/5832

We should do the same thing here.

Corresponding web/server issue:
* https://github.com/zulip/zulip/issues/27802

PR #644, opened 2024-05-05
ui: Rename "All messages" to "Combined feed".

Lalit3716:
This closes: #634.

This is my first try to contribute here, so wanted to ask making this change didn't required any corresponding test changes?
Are we not testing like these buttons should be shown on home screen or something?

Screenshot:
Screenshot 2024-05-05 at 8 21 33 PM

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -460,9 +460,9 @@
   "@recentDmConversationsPageTitle": {
     "description": "Title for the page of recent DM conversations"
   },
-  "allMessagesPageTitle": "All messages",
-  "@allMessagesPageTitle": {
-    "description": "Title for the page of all messages"
+  "combinedFeedPageTitle": "Combined feed",
+  "@combinedFeedPageTitle": {
+    "description": "Title for the page of combined feed."
   },
   "notifGroupDmConversationLabel": "{senderFullName} to you and {numOthers, plural, =1{1 other} other{{numOthers} others}}",
   "@notifGroupDmConversationLabel": {
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -272,7 +272,7 @@ class HomePage extends StatelessWidget {
             onPressed: () => Navigator.push(context,
               MessageListPage.buildRoute(context: context,
                 narrow: const AllMessagesNarrow())),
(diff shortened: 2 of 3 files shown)


assets/l10n/app_en.arb
```diff
@@ -456,9 +456,9 @@
   "@recentDmConversationsPageTitle": {
     "description": "Title for the page of recent DM conversations"
   },
-  "allMessagesPageTitle": "All messages",
+  "allMessagesPageTitle": "Combined feed",
   "@allMessagesPageTitle": {
-    "description": "Title for the page of all messages"
+    "description": "Title for the page of combined feed."
   },
```
chrisbobbe, line 462: nit: let's call it `combinedFeedPageTitle` instead of `allMessagesPageTitle`

chrisbobbe (commented):
Thanks @Lalit3716!

I think it's OK not to have UI tests for the current home page, which is a temporary, prototype-level UI. (#311 describes how we plan to replace it.)

```
ui: Rename "All messages" to "Combined feed".
```

nit: this commit message should have a `Fixes` line, like other commits that fix issues in the tracker. I recommend looking at the project's history using Greg's ["secret"](https://github.com/zulip/zulip-mobile/blob/6d5d56d175644cd0cdf47f3cd30ffadf6756bbdc/docs/howto/git.md#2-the-secret-to-using-git-log--p) for doing so efficiently, to get a sense of our commit-message style. Another small thing you'll notice is that we tend not to end the summary line with a period, so `ui: Rename "All messages" to "Combined feed"` is fine and a little bit shorter.

assets/l10n/app_en.arb
Lalit3716: Done 👍

Lalit3716:
Thanks @chrisbobbe for pointing out those nits and yeah it makes sense to not have tests that we will end up removing later 👍

gnprice:
Thanks @Lalit3716, and thanks @chrisbobbe for the initial review! Looks good — merging.

gnprice:
I've also just filed a followup issue #676 for completing this renaming in our codebase as well as the UI. @Lalit3716 if you'd like to pick up another issue, that would be a great next one to try.

Lalit3716:
Thank you, I will pick this follow-up issue next sounds straight forward to do :)

PR #645, opened 2024-05-06
model [nfc]: Comment on lack of [AutocompleteViewManager.dispose]

chrisbobbe:


gnprice:
Thanks. I prefer a different way of putting it, so I'll push a version that does that. LMK whether the revised version seems clear to you, and we can iterate from there.

chrisbobbe:
Sure, LGTM!

PR #654, opened 2024-05-09
model: Implement StreamColorSwatch.lerp, copying ColorSwatch.lerp

chrisbobbe:
Done as an alternative to #642.

Related: #95


gnprice:
Cool, with this approach it LGTM — merging.

PR #657, opened 2024-05-10
content/msglist: Update some elements closer to new, 2023+ design, to prepare for light/dark

chrisbobbe:
It will be helpful to consult the web app when we define the dark-theme styles of our content elements. This PR prepares for that. I went through all the content elements as we've implemented them, and for all the things that should differ between light and dark theme, ensured that our light-theme style matches the web app.

I also brought the default content text color in line with the web app. (I mention this separately because this color is an attribute of the message list, not content itself, I guess; at least that's how it seems in the code.)

| Before | After |
| --- | --- |
| Screenshot 2024-05-10 at 3 16 07 PM | Screenshot 2024-05-10 at 3 16 25 PM |

Related: #157 
Related: #95

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -773,9 +773,7 @@ final _kInlineCodeStyle = kMonospaceTextStyle
   .merge(const TextStyle(backgroundColor: Color(0xffeeeeee)));
 
 final _kCodeBlockStyle = kMonospaceTextStyle
-  .merge(const TextStyle(
-    backgroundColor: Color.fromRGBO(255, 255, 255, 1),
-    fontSize: 0.825 * kBaseFontSize))
+  .merge(const TextStyle(fontSize: 0.825 * kBaseFontSize))
   .merge(
     // TODO(a11y) pass a BuildContext, to handle platform request for bold text.
     //   To get one, the result of this whole computation (to the TextStyle
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -915,7 +915,10 @@ class GlobalTime extends StatelessWidget {
           child: Row(
             mainAxisSize: MainAxisSize.min,
             children: [
-              Icon(ZulipIcons.clock, size: surroundingFontSize),
+              Icon(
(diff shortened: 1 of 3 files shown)


lib/widgets/content.dart
```diff
@@ -406,12 +406,13 @@ class CodeBlock extends StatelessWidget {
...
-  static final _borderColor = const HSLColor.fromAHSL(0.15, 0, 0, 0).toColor();
-
   @override
   Widget build(BuildContext context) {
     return _CodeBlockContainer(
-      borderColor: _borderColor,
+      // (The other caller, MathBlock, passes a nontransparent color.)
+      // TODO(#46) remove comment about MathBlock when TeX is implemented;
+      //   presumably it won't apply anymore.
+      borderColor: Colors.transparent,
```
gnprice, line 415: Does this even need a comment? Presumably we'll remove the `borderColor` parameter itself when that happens.

lib/widgets/content.dart
```diff
@@ -813,27 +812,26 @@ class UserMention extends StatelessWidget {
...
       child: InlineContent(
         // If an @-mention is inside a link, let the @-mention override it.
         recognizer: null,  // TODO make @-mentions tappable, for info on user
         // One hopes an @-mention can't contain an embedded link.
         // (The parser on creating a UserMentionNode has a TODO to check that.)
         linkRecognizers: null,
+
+        // TODO(#647) when self-user is non-silently mentioned, make bold, and:
+        // TODO(#646) when self-user is non-silently mentioned,
+        //   differ font color between direct and wildcard mentions
```
gnprice, line 829: nit:
```suggestion
        // TODO(#646) when self-user is non-silently mentioned,
        //   distinguish font color between direct and wildcard mentions
```

At least for me, "differ" is only an intransitive verb.

gnprice (commented):
Thanks! LGTM, with two nits below. Otherwise please go ahead and merge.

chrisbobbe:
Thanks! Done.

PR #680, opened 2024-05-15
reactions: Increase prominence of own reactions, following web

chrisbobbe:
This follows zulip/zulip@793020992, which landed after we implemented reaction chips here. (That commit makes some padding adjustments, which we don't need because changing a chip's border width doesn't affect the chip's size.)

| Before | After |
| --- | --- |
| Screenshot 2024-05-15 at 2 46 13 PM | Screenshot 2024-05-15 at 2 46 37 PM |


chrisbobbe:
Like PR #657, this is meant to fix a small deviation from web in our light-theme implementation, toward #95.

gnprice:
LGTM, thanks! Merging.

PR #682, opened 2024-05-15
ui: Slightly adjust app bar and scaffold-background, to align with web

chrisbobbe:
These colors are known in the web app as (in order of appearance):

  --color-background-navbar
  --color-navbar-bottom-border
  --color-background

(The nav-bar border color is unchanged, or changed within a rounding error.)

The colors we had before this were reasonable, since they were taken from the Figma. But the Figma is now over a year old, and it seems at least as reasonable to follow the web app, especially since they're only slightly different. Moreover, we want to implement dark theme soon (#95), and it'll be best to take the light and dark styles from the same source. That currently has to be the web app, since the Figma doesn't have a dark theme.

-----

Screenshots coming soon.

### DIFF
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -5,16 +5,17 @@ import 'text.dart';
 ThemeData zulipThemeData(BuildContext context) {
   return ThemeData(
     typography: zulipTypography(context),
-    appBarTheme: const AppBarTheme(
+    appBarTheme: AppBarTheme(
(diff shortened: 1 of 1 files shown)


chrisbobbe:
| Before | After |
| --- | --- |
| image | image |
| image | image |

chrisbobbe:
Closing as superseded by #687; see discussion linked from there.

PR #687, opened 2024-05-18
ui: Adjust app bar bg and scaffold bg colors toward new Figma design

chrisbobbe:
(This replaces #682.)

From discussion:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/design.3A.20colors/near/1801938

and the Figma linked there:
  https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2074%3A26325&t=rjuqcQaHMiGBUReF-1

Now, the app-bar bottom border is colored by the new Figma's `border-bar`, and the main background color is `bg-main`. The app bar background is unchanged because it already matches `bg-top-bar`.

And all these color variables have dark-theme variants! (See discussion for what those are.) That'll help when we implement dark theme, #95.

### DIFF
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -12,9 +12,12 @@ ThemeData zulipThemeData(BuildContext context) {
       // ColorScheme.surfaceContainer for the scrolled-under state and
       // ColorScheme.surface otherwise, and those are different colors.
       scrolledUnderElevation: 0,
-      backgroundColor: Color(0xfff5f5f5),
+      backgroundColor: Color(0xfff5f5f5), // `bg-top-bar` in Figma
 
-      shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
+      shape: Border(bottom: BorderSide(
+        color: Color(0x33000000), // `border-bar` in Figma
+        strokeAlign: BorderSide.strokeAlignInside, // (default restated for explicitness)
+      )),
     ),
     // This applies Material 3's color system to produce a palette of
     // appropriately matching and contrasting colors for use in a UI.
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Screenshots:

| Before | After |
| --- | --- |
| 842F41EF-8DA5-423A-9986-F3282EB795E5 | 9F8C4207-A41E-4257-9471-835F87F996B5 |
| 7936C278-1006-4701-9C52-B69B74287CE7 | 9444D3FD-E11B-4B5F-AB28-B522E8C1114C |

lib/widgets/theme.dart
```diff
@@ -14,7 +14,10 @@ ThemeData zulipThemeData(BuildContext context) {
       scrolledUnderElevation: 0,
       backgroundColor: Color(0xfff5f5f5),
 
-      shape: Border(bottom: BorderSide(color: Color(0xffcccccc))),
+      shape: Border(bottom: BorderSide(
+        color: Color(0x33000000),
```
gnprice, line 18: Now, the app-bar bottom border is colored by the new Figma's
    `border-bar`, and the main background color is `bg-main`. The app
    bar background is unchanged because it already matches `bg-top-bar`.

This feels like the sort of information that's helpful to have in the code itself, so as a comment something like:

```suggestion
        color: Color(0x33000000), // `border-bar` in Figma
```

As part of #95 I think you'll be turning these into some sort of named reference, so that this would be `foo.borderBarColor` for some `foo`. That'll be another way of putting this information into the code, which is better than a comment. But until then a comment seems helpful.

gnprice (commented):
Thanks! Small comment below.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merging.

Issue #676, opened 2024-05-14
Rename "All messages" to "Combined feed" in our codebase


gnprice:
This is a followup to:
* #634 

That issue was for updating the UI. We should also update all the references to this Zulip feature within our code.

A good search to use for this is:
`git grep -iP 'all.?messages'`
That will find the phrase "all messages" with any capitalization, and with or without a space (or other character) in between the words. Currently it finds 73 results.

When this issue is complete, that search's results won't necessarily be empty — there may be a few matches that aren't referring to the former "All messages" narrow, now called "Combined feed". But it should be possible to look at each of the remaining results and determine how that's not what it's referring to.

For this issue, the most important part of the work is making a [highly reviewable PR](https://zulip.readthedocs.io/en/latest/contributing/reviewable-prs.html). Be sure to review that page of the Zulip docs, and the page on [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html).

PR #688, opened 2024-05-19
Rename "All messages" to "Combined feed" in the codebase.

Lalit3716:
Fixes: #676

### DIFF
--- a/lib/model/narrow.dart
+++ b/lib/model/narrow.dart
@@ -39,13 +39,12 @@ sealed class SendableNarrow extends Narrow {
   MessageDestination get destination;
 }
 
-/// The narrow called "All messages" in the UI.
+/// The narrow called "Combined feed" in the UI.
 ///
-/// This does not literally mean all messages, or even all messages
-/// that the user has access to: in particular it excludes muted streams
-/// and topics.
-class AllMessagesNarrow extends Narrow {
-  const AllMessagesNarrow();
+/// All messages the user has access to, excluding unsubscribed streams
+/// and muted streams and topics. See [PerAccountStore.isTopicVisible].
+class CombinedFeedNarrow extends Narrow {
+  const CombinedFeedNarrow();
 
   @override
   bool containsMessage(Message message) {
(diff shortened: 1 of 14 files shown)


lib/model/narrow.dart
```diff
@@ -39,13 +39,9 @@ sealed class SendableNarrow extends Narrow {
   MessageDestination get destination;
 }
 
-/// The narrow called "All messages" in the UI.
-///
-/// This does not literally mean all messages, or even all messages
-/// that the user has access to: in particular it excludes muted streams
-/// and topics.
```
Lalit3716, line 46: I am not sure whether I should remove this comment now or not?

chrisbobbe: I think we can replace it with something like:

```dart
/// All messages the user has access to, excluding unsubscribed streams
/// and muted streams and topics. See [PerAccountStore.isTopicVisible].
```

chrisbobbe (commented):
Thanks! This is close to being ready.

I noticed that you included `Fixes: #676` in both commits. Please only include the `Fixes: #676` line in the one commit that completely fixes the issue; that is, the commit at which the "all messages" name no longer appears in the codebase. (If you want, you could include a `Fixes-partly: #676` in the first commit, but I think we don't have a consistent pattern of doing that.)

Also, I find one remaining reference to `AllMessagesNarrow`, in test/widgets/message_list_test.dart.

chrisbobbe:
It looks like some conflicts have appeared. Please resolve those and comment when this is ready for another review.

lib/model/narrow.dart
Lalit3716: Sounds good done!

Lalit3716:
Resolved the conflicts and removed the `Fixes` from the first commit. This is ready for another look.

lib/model/narrow.dart
```diff
@@ -39,13 +39,11 @@ sealed class SendableNarrow extends Narrow {
...
-class AllMessagesNarrow extends Narrow {
-  const AllMessagesNarrow();
+/// The narrow called "Combined feed" in the UI.
+/// All messages the user has access to, excluding unsubscribed streams
+/// and muted streams and topics. See [PerAccountStore.isTopicVisible].
```
chrisbobbe, line 44: nit: empty line between dartdoc's summary line and body

chrisbobbe (commented):
Thanks! LGTM except one nit below, but I'll go ahead and label this for Greg to review.

gnprice (commented):
Thanks @Lalit3716 and @chrisbobbe!

Apart from the comment Chris made above, I reran the search from the issue, and it looks like there's one match remaining that should be updated:
```
$ git grep -i 'all.?messages' 
…
lib/widgets/message_list.dart:            // widgets. As a simple test, flinging through All Messages in
```

Lalit3716:
Updated ready for another look now!

gnprice:

Issue #690, opened 2024-05-20
content: Code blocks don't respond to system bold-text setting


chrisbobbe:

chrisbobbe:
(This is something I spotted a while ago, but it's a minor issue and I wanted to write my fix after #F544 to avoid a conflict.)

PR #691, opened 2024-05-20
content: Handle bold-text setting in code blocks and error-unimplementeds

chrisbobbe:
Screenshots show the app with the system bold text setting on.

Fixes: #690

| Before | After |
| --- | --- |
| 728B9036-3F4C-4E00-AAA5-4ED34CDD88CE | B6815335-22C0-4B2F-9A21-D2190A67018B |

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -847,18 +847,10 @@ const _kInlineCodeFontSizeFactor = 0.825;
 /// Inline code should use this and also apply [_kInlineCodeFontSizeFactor]
 /// to the font size of the surrounding text
 /// (which might be a Paragraph, a Heading, etc.).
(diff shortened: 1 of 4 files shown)


gnprice:
Thanks! Looks good; merging, with one nit:
```
    -    text [nfc]: Remove some params that equal defaults
    +    text [nfc]: Remove some arguments that equal defaults
```

because I'd read "remove some parameters" to mean changing the *signature* of a function, rather than some of its call sites. See:
https://developer.mozilla.org/en-US/docs/Glossary/Parameter#parameters_versus_arguments

(It's common to use the terms interchangeably and context usually makes the meaning clear, but here in the terseness of a summary line there's not much context.)

Issue #150, opened 2023-06-02
Handle moved messages (update_message events, 2/2)


gnprice:
https://zulip.com/api/get-events#update_message

This issue is for the case where the messages were moved to a different stream or topic. (The case where a message's content changed is #118.)

In particular these should be handled in the MessageListView view-model, so that the message disappears from the view if it no longer belongs in that particular narrow, or appears if it newly does belong.

gnprice:
This will be mostly covered by #787. But after that lands, we should follow up with a quick scan to identify anything else that needs updating on message moves — there are several new data structures in the store that we've merged in recent weeks.

PR #692, opened 2024-05-20
model: Introduce data structures for "recent senders criterion" of user-mention autocomplete

sm-sayedi:
This is the second PR in the series of PRs #608 is divided into. This solely focuses on bringing the data structures for the upcoming "recent senders" criterion of @-mention autocomplete. #828 is the next PR in the series.

Fixes part of: #228

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -403,6 +403,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       numAfter: 0,
     );
     store.reconcileMessages(result.messages);
+    store.recentSenders.handleMessages(result.messages); // TODO(#824)
     for (final message in result.messages) {
       if (_messageVisible(message)) {
         _addMessage(message);
@@ -439,6 +440,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       }
 
       store.reconcileMessages(result.messages);
+      store.recentSenders.handleMessages(result.messages); // TODO(#824)
 
       final fetchedMessages = _allMessagesVisible
         ? result.messages // Avoid unnecessarily copying the list.
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -101,7 +101,7 @@ class RecentSenders {
 
 @visibleForTesting
 class MessageIdTracker {
-  /// A list of distinct message IDs, sorted ascendingly.
+  /// A list of distinct message IDs, sorted ascending.
   @visibleForTesting
   QueueList<int> ids = QueueList();
 
@@ -128,7 +128,7 @@ class MessageIdTracker {
 
   /// Add the messages IDs to the tracker list at the proper place, if not present.
   ///
-  /// [newIds] should be sorted ascendingly.
+  /// [newIds] should be sorted ascending.
   void addAll(QueueList<int> newIds) {
     if (ids.isEmpty) {
       ids = newIds;
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -78,23 +78,23 @@ class RecentSenders {
     }
 
     final DeleteMessageEvent(:streamId!, :topic!) = event;
-    final sendersByStream = streamSenders[streamId];
-    final topicsByStream = topicSenders[streamId];
-    final sendersByTopic = topicsByStream?[topic];
+    final sendersInStream = streamSenders[streamId];
+    final topicsInStream = topicSenders[streamId];
+    final sendersInTopic = topicsInStream?[topic];
     for (final entry in messagesByUser.entries) {
       final MapEntry(key: senderId, value: messages) = entry;
 
-      final messagesBySenderInStream = sendersByStream?[senderId];
-      messagesBySenderInStream?.removeAll(messages);
-      if (messagesBySenderInStream?.maxId == null) sendersByStream?.remove(senderId);
+      final streamTracker = sendersInStream?[senderId];
+      streamTracker?.removeAll(messages);
+      if (streamTracker?.maxId == null) sendersInStream?.remove(senderId);
 
-      final messagesBySenderInTopic = sendersByTopic?[senderId];
-      messagesBySenderInTopic?.removeAll(messages);
-      if (messagesBySenderInTopic?.maxId == null) sendersByTopic?.remove(senderId);
+      final topicTracker = sendersInTopic?[senderId];
+      topicTracker?.removeAll(messages);
+      if (topicTracker?.maxId == null) sendersInTopic?.remove(senderId);
     }
-    if (sendersByStream?.isEmpty ?? false) streamSenders.remove(streamId);
-    if (sendersByTopic?.isEmpty ?? false) topicsByStream?.remove(topic);
-    if (topicsByStream?.isEmpty ?? false) topicSenders.remove(streamId);
+    if (sendersInStream?.isEmpty ?? false) streamSenders.remove(streamId);
+    if (sendersInTopic?.isEmpty ?? false) topicsInStream?.remove(topic);
+    if (topicsInStream?.isEmpty ?? false) topicSenders.remove(streamId);
   }
 }
 
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -34,13 +34,13 @@ class RecentSenders {
   /// [messages] should be sorted by [id] ascendingly, which are, the way app
   /// receives and handles messages.
   void handleMessages(List<Message> messages) {
-    final messagesByUserInStream = <(int, int), List<int>>{};
-    final messagesByUserInTopic = <(int, String, int), List<int>>{};
+    final messagesByUserInStream = <(int, int), QueueList<int>>{};
+    final messagesByUserInTopic = <(int, String, int), QueueList<int>>{};
     for (final message in messages) {
       if (message is! StreamMessage) continue;
       final StreamMessage(:streamId, :topic, :senderId, id: int messageId) = message;
-      (messagesByUserInStream[(streamId, senderId)] ??= []).add(messageId);
-      (messagesByUserInTopic[(streamId, topic, senderId)] ??= []).add(messageId);
+      (messagesByUserInStream[(streamId, senderId)] ??= QueueList()).add(messageId);
+      (messagesByUserInTopic[(streamId, topic, senderId)] ??= QueueList()).add(messageId);
     }
 
     for (final entry in messagesByUserInStream.entries) {
@@ -128,9 +128,9 @@ class MessageIdTracker {
   /// Add the messages IDs to the tracker list at the proper place, if not present.
   ///
   /// [newIds] should be sorted ascendingly.
-  void addAll(List<int> newIds) {
+  void addAll(QueueList<int> newIds) {
     if (ids.isEmpty) {
-      ids = QueueList.from(newIds);
+      ids = newIds;
       return;
     }
     ids = setUnion(ids, newIds);
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -5,10 +5,10 @@ import '../api/model/events.dart';
 import '../api/model/model.dart';
 import 'algorithms.dart';
 
-/// A data structure to keep track of stream and topic messages of users (senders).
+/// Tracks the latest messages sent by each user, in each stream and topic.
 ///
 /// Use [latestMessageIdOfSenderInStream] and [latestMessageIdOfSenderInTopic]
-/// to get the relevant data.
+/// for queries.
 class RecentSenders {
   // streamSenders[streamId][senderId] = MessageIdTracker
   @visibleForTesting
@@ -18,21 +18,24 @@ class RecentSenders {
   @visibleForTesting
   final Map<int, Map<String, Map<int, MessageIdTracker>>> topicSenders = {};
 
+  /// The latest message the given user sent to the given stream,
+  /// or null if no such message is known.
   int? latestMessageIdOfSenderInStream({
     required int streamId,
     required int senderId,
   }) => streamSenders[streamId]?[senderId]?.maxId;
 
+  /// The latest message the given user sent to the given topic,
+  /// or null if no such message is known.
   int? latestMessageIdOfSenderInTopic({
     required int streamId,
     required String topic,
     required int senderId,
   }) => topicSenders[streamId]?[topic]?[senderId]?.maxId;
 
-  /// Records the necessary data from each message if it is a [StreamMessage].
+  /// Records the necessary data from a batch of just-fetched messages.
   ///
-  /// [messages] should be sorted by [id] ascendingly, which are, the way app
-  /// receives and handles messages.
+  /// The messages must be sorted by [Message.id] ascending.
   void handleMessages(List<Message> messages) {
     final messagesByUserInStream = <(int, int), QueueList<int>>{};
     final messagesByUserInTopic = <(int, String, int), QueueList<int>>{};
@@ -55,9 +58,7 @@ class RecentSenders {
     }
   }
 
-  /// Records the necessary data from [message] if it is a [StreamMessage].
-  ///
-  /// If [message] is not a [StreamMessage], this is a no-op.
+  /// Records the necessary data from a new message.
   void handleMessage(Message message) {
     if (message is! StreamMessage) return;
     final StreamMessage(:streamId, :topic, :senderId, id: int messageId) = message;
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -1,0 +1,148 @@
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+
+import '../api/model/events.dart';
+import '../api/model/model.dart';
+import 'algorithms.dart';
+
+/// A data structure to keep track of stream and topic messages of users (senders).
+///
+/// Use [latestMessageIdOfSenderInStream] and [latestMessageIdOfSenderInTopic]
+/// to get the relevant data.
+class RecentSenders {
+  // streamSenders[streamId][senderId] = MessageIdTracker
+  @visibleForTesting
+  final Map<int, Map<int, MessageIdTracker>> streamSenders = {};
+
+  // topicSenders[streamId][topic][senderId] = MessageIdTracker
+  @visibleForTesting
+  final Map<int, Map<String, Map<int, MessageIdTracker>>> topicSenders = {};
+
+  int? latestMessageIdOfSenderInStream({
+    required int streamId,
+    required int senderId,
+  }) => streamSenders[streamId]?[senderId]?.maxId;
+
+  int? latestMessageIdOfSenderInTopic({
+    required int streamId,
+    required String topic,
+    required int senderId,
+  }) => topicSenders[streamId]?[topic]?[senderId]?.maxId;
+
+  /// Records the necessary data from each message if it is a [StreamMessage].
+  ///
+  /// [messages] should be sorted by [id] ascendingly, which are, the way app
+  /// receives and handles messages.
+  void handleMessages(List<Message> messages) {
+    final messagesByUserInStream = <(int, int), List<int>>{};
+    final messagesByUserInTopic = <(int, String, int), List<int>>{};
+    for (final message in messages) {
+      if (message is! StreamMessage) continue;
+      final StreamMessage(:streamId, :topic, :senderId, id: int messageId) = message;
+      (messagesByUserInStream[(streamId, senderId)] ??= []).add(messageId);
+      (messagesByUserInTopic[(streamId, topic, senderId)] ??= []).add(messageId);
+    }
+
+    for (final entry in messagesByUserInStream.entries) {
+      final (streamId, senderId) = entry.key;
+      ((streamSenders[streamId] ??= {})
+        [senderId] ??= MessageIdTracker()).addAll(entry.value);
+    }
+    for (final entry in messagesByUserInTopic.entries) {
+      final (streamId, topic, senderId) = entry.key;
+      (((topicSenders[streamId] ??= {})[topic] ??= {})
+        [senderId] ??= MessageIdTracker()).addAll(entry.value);
+    }
+  }
+
+  /// Records the necessary data from [message] if it is a [StreamMessage].
+  ///
+  /// If [message] is not a [StreamMessage], this is a no-op.
+  void handleMessage(Message message) {
+    if (message is! StreamMessage) return;
+    final StreamMessage(:streamId, :topic, :senderId, id: int messageId) = message;
+    ((streamSenders[streamId] ??= {})
+      [senderId] ??= MessageIdTracker()).add(messageId);
+    (((topicSenders[streamId] ??= {})[topic] ??= {})
+      [senderId] ??= MessageIdTracker()).add(messageId);
+  }
+
+  void handleDeleteMessageEvent(DeleteMessageEvent event, Map<int, Message> cachedMessages) {
+    if (event.messageType != MessageType.stream) return;
+
+    final messagesByUser = <int, List<int>>{};
+    for (final id in event.messageIds) {
+      final message = cachedMessages[id] as StreamMessage?;
+      if (message == null) continue;
+      (messagesByUser[message.senderId] ??= []).add(id);
+    }
+
+    final DeleteMessageEvent(:streamId!, :topic!) = event;
+    final sendersByStream = streamSenders[streamId];
+    final topicsByStream = topicSenders[streamId];
+    final sendersByTopic = topicsByStream?[topic];
+    for (final entry in messagesByUser.entries) {
+      final MapEntry(key: senderId, value: messages) = entry;
+
+      final messagesBySenderInStream = sendersByStream?[senderId];
+      messagesBySenderInStream?.removeAll(messages);
+      if (messagesBySenderInStream?.maxId == null) sendersByStream?.remove(senderId);
+
+      final messagesBySenderInTopic = sendersByTopic?[senderId];
+      messagesBySenderInTopic?.removeAll(messages);
+      if (messagesBySenderInTopic?.maxId == null) sendersByTopic?.remove(senderId);
+    }
+    if (sendersByStream?.isEmpty ?? false) streamSenders.remove(streamId);
+    if (sendersByTopic?.isEmpty ?? false) topicsByStream?.remove(topic);
+    if (topicsByStream?.isEmpty ?? false) topicSenders.remove(streamId);
+  }
+}
+
+@visibleForTesting
+class MessageIdTracker {
+  /// A list of distinct message IDs, sorted ascendingly.
+  @visibleForTesting
+  QueueList<int> ids = QueueList();
+
+  /// The maximum id in the tracker list, or `null` if the list is empty.
+  int? get maxId => ids.lastOrNull;
+
+  /// Add the message ID to the tracker list at the proper place, if not present.
+  ///
+  /// Optimized, taking O(1) time for the case where that place is the end,
+  /// because that's the common case for a message that is received through
+  /// [PerAccountStore.handleEvent]. May take O(n) time in some rare cases.
+  void add(int id) {
+    if (ids.isEmpty || id > ids.last) {
+      ids.addLast(id);
+      return;
+    }
+    final i = lowerBound(ids, id);
+    if (i < ids.length && ids[i] == id) {
+      // The ID is already present. Nothing to do.
+      return;
+    }
+    ids.insert(i, id);
+  }
+
+  /// Add the messages IDs to the tracker list at the proper place, if not present.
+  ///
+  /// [newIds] should be sorted ascendingly.
+  void addAll(List<int> newIds) {
+    if (ids.isEmpty) {
+      ids = QueueList.from(newIds);
+      return;
+    }
+    ids = setUnion(ids, newIds);
+  }
+
+  void removeAll(List<int> idsToRemove) {
+    ids.removeWhere((id) {
+      final i = lowerBound(idsToRemove, id);
+      return i < idsToRemove.length && idsToRemove[i] == id;
+    });
+  }
+
+  @override
+  String toString() => ids.toString();
+}
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -23,6 +23,7 @@ import 'database.dart';
 import 'message.dart';
 import 'message_list.dart';
 import 'recent_dm_conversations.dart';
+import 'recent_senders.dart';
 import 'channel.dart';
 import 'typing_status.dart';
 import 'unreads.dart';
@@ -256,6 +257,7 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
       ),
       recentDmConversationsView: RecentDmConversationsView(
         initial: initialSnapshot.recentPrivateConversations, selfUserId: account.userId),
+      recentSenders: RecentSenders(),
     );
   }
 
@@ -276,6 +278,7 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
     required MessageStoreImpl messages,
     required this.unreads,
     required this.recentDmConversationsView,
+    required this.recentSenders,
   }) : assert(selfUserId == globalStore.getAccount(accountId)!.userId),
        assert(realmUrl == globalStore.getAccount(accountId)!.realmUrl),
        assert(realmUrl == connection.realmUrl),
@@ -368,6 +371,8 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
   final Unreads unreads;
 
   final RecentDmConversationsView recentDmConversationsView;
+
+  final RecentSenders recentSenders;
 
   ////////////////////////////////
   // Other digests of data.
@@ -492,6 +497,7 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
         _messages.handleMessageEvent(event);
         unreads.handleMessageEvent(event);
         recentDmConversationsView.handleMessageEvent(event);
+        recentSenders.handleMessage(event.message); // TODO(#824)
         // When adding anything here (to handle [MessageEvent]),
         // it probably belongs in [reconcileMessages] too.
 
@@ -502,6 +508,11 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
 
       case DeleteMessageEvent():
         assert(debugLog("server event: delete_message ${event.messageIds}"));
+        // This should be called before [_messages.handleDeleteMessageEvent(event)],
+        // as we need to know about each message for [event.messageIds],
+        // specifically, their `senderId`s. By calling this after the
+        // aforementioned line, we'll lose reference to those messages.
+        recentSenders.handleDeleteMessageEvent(event, messages);
         _messages.handleDeleteMessageEvent(event);
         unreads.handleDeleteMessageEvent(event);
 
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -296,6 +296,7 @@ void main() {
     check(done).isFalse();
     for (int i = 0; i < 3; i++) {
       await Future(() {});
+      if (done) break;
     }
     check(done).isTrue();
     final results = view.results
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -18,6 +18,7 @@ import '../api/model/model_checks.dart';
 import '../example_data.dart' as eg;
 import '../stdlib_checks.dart';
 import 'content_checks.dart';
+import 'recent_senders_test.dart' as recent_senders_test;
 import 'test_store.dart';
 
 void main() {
@@ -141,6 +142,25 @@ void main() {
       ..haveOldest.isTrue();
   });
 
+  // TODO(#824): move this test
+  test('fetchInitial, recent senders track all the messages', () async {
+    const narrow = CombinedFeedNarrow();
+    await prepare(narrow: narrow);
+    final messages = [
+      eg.streamMessage(),
+      // Not subscribed to the stream with id 10.
+      eg.streamMessage(stream: eg.stream(streamId: 10)),
+    ];
+    connection.prepare(json: newestResult(
+      foundOldest: false,
+      messages: messages,
+    ).toJson());
+    await model.fetchInitial();
+
+    check(model).messages.length.equals(1);
+    recent_senders_test.checkMatchesMessages(store.recentSenders, messages);
+  });
+
   test('fetchOlder', () async {
     const narrow = CombinedFeedNarrow();
     await prepare(narrow: narrow);
@@ -231,6 +251,27 @@ void main() {
     check(model)
       ..fetchingOlder.isFalse()
       ..messages.length.equals(200);
+  });
+
+  // TODO(#824): move this test
+  test('fetchOlder, recent senders track all the messages', () async {
+    const narrow = CombinedFeedNarrow();
+    await prepare(narrow: narrow);
+    final initialMessages = List.generate(10, (i) => eg.streamMessage(id: 100 + i));
+    await prepareMessages(foundOldest: false, messages: initialMessages);
+
+    final oldMessages = List.generate(10, (i) => eg.streamMessage(id: 89 + i))
+      // Not subscribed to the stream with id 10.
+      ..add(eg.streamMessage(id: 99, stream: eg.stream(streamId: 10)));
+    connection.prepare(json: olderResult(
+      anchor: 100, foundOldest: false,
+      messages: oldMessages,
+    ).toJson());
+    await model.fetchOlder();
+
+    check(model).messages.length.equals(20);
+    recent_senders_test.checkMatchesMessages(store.recentSenders,
+      [...initialMessages, ...oldMessages]);
   });
 
   test('MessageEvent', () async {
--- a/test/model/recent_senders_test.dart
+++ b/test/model/recent_senders_test.dart
@@ -71,8 +71,7 @@ void main() {
       setupModel(oldMessages);
       model.handleMessages(newMessages);
       final expectedMessages = [...oldMessages, ...newMessages]
-        ..removeWhere((m) => m is! StreamMessage)
-        ..sort((m1, m2) => m1.id.compareTo(m2.id));
+        ..removeWhere((m) => m is! StreamMessage);
       checkMatchesMessages(model, expectedMessages);
     }
 
--- a/test/model/recent_senders_test.dart
+++ b/test/model/recent_senders_test.dart
@@ -4,7 +4,7 @@ import 'package:zulip/api/model/model.dart';
 import 'package:zulip/model/recent_senders.dart';
 import '../example_data.dart' as eg;
 
-/// [messages] should be sorted by [id] ascendingly.
+/// [messages] should be sorted by [id] ascending.
 void checkMatchesMessages(RecentSenders model, List<Message> messages) {
   final Map<int, Map<int, Set<int>>> messagesByUserInStream = {};
   final Map<int, Map<String, Map<int, Set<int>>>> messagesByUserInTopic = {};
--- a/test/model/recent_senders_test.dart
+++ b/test/model/recent_senders_test.dart
@@ -77,60 +77,42 @@ void main() {
     }
 
     group('single tracker', () {
-      test('batch goes before the existing messages', () {
+      void checkHandleMessagesSingle(List<int> oldIds, List<int> newIds) {
         checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
+          for (final id in oldIds)
+            eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: id),
         ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
+          for (final id in newIds)
+            eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: id),
         ]);
+      }
+
+      test('batch goes before the existing messages', () {
+        checkHandleMessagesSingle([300, 400], [100, 200]);
       });
 
       test('batch goes after the existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 600),
-        ]);
+        checkHandleMessagesSingle([300, 400], [500, 600]);
       });
 
       test('batch is interspersed among the existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
-        ]);
+        checkHandleMessagesSingle([200, 400], [100, 300, 500]);
       });
 
       test('batch contains some of already-existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
-        ]);
+        checkHandleMessagesSingle([200, 300, 400], [100, 200, 400, 500]);
       });
+    });
 
-      test('batch with both DM and stream messages -> ignores DM, processes stream messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], id: 400),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
-        ]);
-      });
+    test('batch with both DM and stream messages -> ignores DM, processes stream messages', () {
+      checkHandleMessages([
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
+      ], [
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
+        eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], id: 400),
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
+      ]);
     });
 
     group('multiple trackers', () {
--- a/test/model/recent_senders_test.dart
+++ b/test/model/recent_senders_test.dart
@@ -58,7 +58,9 @@ void main() {
   group('RecentSenders.handleMessages', () {
     late RecentSenders model;
     final stream1 = eg.stream(streamId: 1);
+    final stream2 = eg.stream(streamId: 2);
     final user10 = eg.user(userId: 10);
+    final user20 = eg.user(userId: 20);
 
     void setupModel(List<Message> messages) {
       model = RecentSenders();
@@ -104,75 +106,39 @@ void main() {
     });
 
     test('batch with both DM and stream messages -> ignores DM, processes stream messages', () {
-      checkHandleMessages([
-        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-      ], [
-        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-        eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], id: 400),
-        eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
+      checkHandleMessages([], [
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10),
+        eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]),
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10),
       ]);
     });
 
-    group('multiple trackers', () {
-      test('batch goes before the existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 500),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 600),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 400),
-        ]);
-      });
+    test('add new sender', () {
+      checkHandleMessages(
+        [eg.streamMessage(stream: stream1, topic: 'a', sender: user10)],
+        [eg.streamMessage(stream: stream1, topic: 'a', sender: user20)]);
+    });
 
-      test('batch goes after the existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 600),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 500),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 700),
-        ]);
-      });
+    test('add new topic', () {
+      checkHandleMessages(
+        [eg.streamMessage(stream: stream1, topic: 'a', sender: user10)],
+        [eg.streamMessage(stream: stream1, topic: 'b', sender: user10)]);
+    });
 
-      test('batch is interspersed among the existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 500),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 700),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 400),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 600),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 800),
-        ]);
-      });
+    test('add new stream', () {
+      checkHandleMessages(
+        [eg.streamMessage(stream: stream1, topic: 'a', sender: user10)],
+        [eg.streamMessage(stream: stream2, topic: 'a', sender: user10)]);
+    });
 
-      test('batch contains some of already-existing messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 300),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 400),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 400),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 500),
-        ]);
-      });
-
-      test('batch with both DM and stream messages -> ignores DM, processes stream messages', () {
-        checkHandleMessages([
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 200),
-        ], [
-          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
-          eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], id: 200),
-          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 400),
-        ]);
-      });
+    test('multiple conversations and senders interspersed', () {
+      checkHandleMessages([], [
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10),
+        eg.streamMessage(stream: stream1, topic: 'b', sender: user10),
+        eg.streamMessage(stream: stream2, topic: 'a', sender: user10),
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user20),
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user10),
+      ]);
     });
   });
 
--- a/test/model/recent_senders_test.dart
+++ b/test/model/recent_senders_test.dart
@@ -57,10 +57,10 @@ void main() {
 
   group('RecentSenders.handleMessages', () {
     late RecentSenders model;
-    final stream1 = eg.stream(streamId: 1);
-    final stream2 = eg.stream(streamId: 2);
-    final user10 = eg.user(userId: 10);
-    final user20 = eg.user(userId: 20);
+    final streamA = eg.stream();
+    final streamB = eg.stream();
+    final userX = eg.user();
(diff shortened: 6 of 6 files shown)


lib/model/message_list.dart
```diff
@@ -381,6 +381,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     for (final message in result.messages) {
       if (_messageVisible(message)) {
         _addMessage(message);
+        store.recentSenders.handleMessage(message);
       }
```
sm-sayedi, line 385: Reply to your [608 comment](https://github.com/zulip/zulip-flutter/pull/608#discussion_r1600478354):
> Good question.
>
> What does the dartdoc on that _messageVisible method say?
>
> Based on that, what would be the consequence of having that condition control whether to make this call? And what would be the consequence of ignoring that condition when making this call?

I concluded that the call should be inside the if condition. We track the messages that are shown in the feed, and if they're not shown, then there's no need to show the users who sent these messages, in the autocomplete results. Thanks for the clue!!

gnprice:
Let's start applying the ["buddy review" system](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/GitHub.20workflows/near/1800080) on this PR, now that the start of the official GSoC period is imminent and it seems like everyone's actively engaged again.

@rajveermalviya, would you do the first review on this PR? Some quick bits of context to note:
* The first few commits are part of #693, as mentioned in the description, and are getting reviewed there. So please focus on the commits that come after those.
* The last round of review on the predecessor PR started here (and came as several separate comments): https://github.com/zulip/zulip-flutter/pull/608#pullrequestreview-2056029568

  Some of that is about what's now in #693; some of it is about what's added in this PR.

Meanwhile we'll proceed with maintainer review for the first PR in this series, #693, because after the previous rounds of review (on #608) that one is further along.

As a reminder since we're just starting to use this system: when you get a buddy review request, [please prioritize it](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/GitHub.20workflows/near/1800082) right after important bugs and your regressions, ahead of most of your own PRs. (In particular I don't think anyone is currently working on an issue that would come ahead of doing buddy reviews.)

lib/model/message_list.dart
rajveermalviya: If we look at web impl, [`process_stream_message`](https://github.com/zulip/zulip/blob/62dfd93a83b8604c53fcd7c6b00bfab90fa55d99/web/src/message_helper.js#L54) in [`process_new_message`](https://github.com/zulip/zulip/blob/62dfd93a83b8604c53fcd7c6b00bfab90fa55d99/web/src/message_helper.js#L11-L84) doesn't seem to care about topic visibility policy. Not sure if that's intended or a bug.

rajveermalviya (commented):
Some comments, otherwise looks great!
Thanks!

lib/model/message_list.dart
gnprice: Yeah, good comparison.

Here, `_messageVisible` says whether the message should appear in this particular message list:
```dart
  /// Whether [message] should actually appear in this message list,
  /// given that it does belong to the narrow.
  ///
  /// This depends in particular on whether the message is muted in
  /// one way or another.
  ///
  /// See also [_allMessagesVisible].
  bool _messageVisible(Message message) {
```

But if the message doesn't appear in this message list, it might still appear in a different one later. A message hidden from the combined feed because the stream is muted will appear if you navigate to that stream; a message hidden because the stream and/or topic is muted will appear if you navigate to the topic.

The recent-senders data structure isn't specific to a particular message list; it's shared across them all (within an account). So let's have the data it gets not be controlled by the message list's `_messageVisible`.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,138 @@
...
+  }) {
+    final topicsMap = _topicSenders[streamId] ??= {};
+    final sendersMap = topicsMap[topic] ??= {};
+    final idTracker = sendersMap[senderId] ??= MessageIdTracker();
+    idTracker.add(messageId);
+  }
+
+  /// Extracts and keeps track of the necessary data from a [message] only
+  /// if it is a stream message.
+  void handleMessage(Message message) {
```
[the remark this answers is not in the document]
sm-sayedi: This change would match it with other somewhat similar parts of the codebase, but there are places this method is called where there is no `MessageEvent` available, such as in `model/message_list.dart` file. In contrast, `RecentDmConversationsView.handleMessageEvent` is only called in `PerAccountStore.handleEvent` where the `MessageEvent` is available.

sm-sayedi:
Thanks, @rajveermalviya for the review! Pushed the revision. Please have a look.

test/model/autocomplete_test.dart
```diff
@@ -274,35 +276,183 @@ void main() {
...
+    });
+
+    // test('fetching older messages', () async {
+    //   final users = generateUsers(count: 1500);
+    //   final store = eg.store();
+    //   final stream = eg.stream();
+    //   final narrow = StreamNarrow(stream.streamId);
+    //   final subscription = eg.subscription(stream);
+    //   await store.addStream(stream);
+    //   await store.addSubscription(subscription);
```
sm-sayedi, line 415: @gnprice I have commented out this test for now as it wouldn't pass. I spent a significant amount of time to diagnose what was the problem, but unfortunately, I wasn't able to figure it out. I would appreciate it if you can have a look at it.

gnprice: Yeah. I think fundamentally the issue here is that these tests using `await Future(() {});` are fragile.

They're trying to make several different things happen in a certain order, but those things are asynchronous in the code. The `await Future(() {})` basically allows each thread of async control flow to advance from one `await` to the next `await`… but what that means in terms of actual behavior of the models depends in a very brittle way on the exact details of how the code being tested is written. Small differences in that code will cause things in the test to happen in a different order and break the test.

That's also why the existing tests of this flavor have so many checks like `check(done).isFalse()` — those are there to make sure that the test's assumptions about what order things happen in are satisfied. That way, if those assumptions are broken by a refactor or by a change to the test, the test fails instead of silently ceasing to successfully test what it's meant to test.

---

It looks like this test is failing at the `check(done).isFalse()` here:
```dart
      await act(store);

      await Future(() {});
      check(done).isFalse();
      await Future(() {});
      check(done).isTrue();
      expect(view.results);
```

so IOW the search is done sooner than the test expects. Presumably it's due to some difference between the `fetchOlder` in this test, and the variations on `handleEvent` found in the neighboring tests.

---

The test passes if you make this change:
```diff
@@ -293,7 +296,6 @@ void main() {
       view.addListener(() { done = true; });
       view.query = MentionAutocompleteQuery(rawQuery);
 
-      await Future(() {});
       check(done).isFalse();
 
       await act(store);
```

But in that case is the test still successfully testing what it's supposed to test? I'm not sure what the assumptions are of your `doCheck` method.

(Probably a chat thread would be a good place to go into further detail.)

lib/model/recent_senders.dart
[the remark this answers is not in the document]
rajveermalviya: That makes sense.

gnprice:
@hackerkid Putting this back in your queue after we discussed today how mentor review is meant to work :slightly_smiling_face: 

(#713 would be a good PR to review first, though — it's smaller, and also this one is stacked atop #693 which is still in progress, so there's no rush for this one to get through review before #693 is ready.)

sm-sayedi:
@hackerkid Now that #693 is merged, this PR is now ready for the mentor review!

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,133 @@
...
+      if (i < _ids.length && _ids[i] == id) return; // the [id] already exists, so do not add it.
+      _ids.insert(i, id);
+    }
+  }
+
+  // TODO: remove
+
+  /// The maximum id in the tracker list.
+  ///
+  /// Returns -1 if the tracker list is empty.
```
hackerkid, line 30: Are these comments still relevant?

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,133 @@
...
+  int get maxId => _ids.isNotEmpty ? _ids.last : -1;
```
hackerkid, line 31: Have we considered returning null instead of -1?

hackerkid:
@sm-sayedi > These data structures are used to keep track of user messages in topics
and streams.

Can you also please expand the commit message to explicitly mention what MessageTracker and RecentSender does?

lib/model/recent_senders.dart
sm-sayedi: Yeah, I think they're relevant, specifically the **"Returns `null` if the tracker list is empty"** part which will help the reader what to expect.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  final List<int> _ids = [];
+
+  void add(int id) {
+    if (_ids.isEmpty) {
+      _ids.add(id);
+    } else {
+      int i = lowerBound(_ids, id);
+      if (i < _ids.length && _ids[i] == id) return; // the [id] already exists, so do not add it.
```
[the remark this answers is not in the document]
sm-sayedi: I think that refactor would provide a more descriptive tone to the code, but I am not sure if it is necessary for this stage as we would end up using it just in one place (in `add()` method). We have a similar method `RecentDmConversationsView._insertSorted`, but that's used in two places. Maybe @gnprice would like to provide their suggestion here.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  // TODO: remove
+
+  /// The maximum id in the tracker list.
+  ///
+  /// Returns -1 if the tracker list is empty.
+  int get maxId => _ids.isNotEmpty ? _ids.last : -1;
+
+  /// Getter for the tracked message IDs.
```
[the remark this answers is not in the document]
sm-sayedi: Removed.

lib/model/recent_senders.dart
sm-sayedi: Yeah, I think returning `null` is better as we preferred it in the previous PR (#693) too.

sm-sayedi:
Thanks @hackerkid for the review! Revision pushed. Also added a few comments in the previous review thread. PTAL.

lib/model/recent_senders.dart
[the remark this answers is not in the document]
gnprice: I agree it'd be good to make that information more explicit. I think a good main way to do that would be by writing down how the data structure `_ids` is supposed to work — the fact that it's meant to be sorted, and have only distinct elements. Like so:
```dart
  /// A list of distinct message IDs, sorted.
  final List<int> _ids = [];
```

Once the reader knows that, it's mostly clear already what `add` should do; the only remaining nuance is that it's fine if the message ID is already there, and it just won't do anything in that case. So:
```dart
  /// Add the given message ID to the list, if not already present.
  void add(int id) {
```

In general, the meanings of the data structures are often the highest-impact facts to write down about how an area of code is intended to work. There's an often-quoted line about this [by Fred Brooks](https://en.wikiquote.org/wiki/Fred_Brooks), one of the earliest software-engineering managers:
> Show me your flowcharts and conceal your tables, and I shall continue to be mystified. Show me your tables, and I won’t usually need your flowcharts; they’ll be obvious.

The reference to "flowcharts" shows how old the observation is (published 1975), but it remains accurate today.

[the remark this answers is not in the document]
gnprice: BTW I'd encourage asking questions directed at me in #mobile-team, if it's a GitHub thread I'm not already active in. (Could even ask here first, and then just quote it in a message there.) I try to periodically catch up on all threads in the repo, but I skim and might miss things (almost missed this one), and I'm also likely to see it much sooner in Zulip.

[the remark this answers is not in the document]
hackerkid: > Show me your flowcharts and conceal your tables, and I shall continue to be mystified. Show me your tables, and I won’t usually need your flowcharts; they’ll be obvious
Thanks for sharing this. I had never came across this before!!

What do you think of making the variable name more descriptive instead of using comment? Comments tend to get outdated.
```
final List<int> _distinctSortedMessageIDs = [];
```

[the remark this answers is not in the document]
gnprice: > making the variable name more descriptive instead of using comment? Comments tend to get outdated.

I definitely sympathize with that direction in general. But this doesn't feel to me like the right situation for it.

I think a key part of what's giving me that feeling is that this is the central data structure that all the code surrounding it is structured around. If those aspects (the values are distinct, they're sorted) were to change in some future commit, it'd look more like scrapping and replacing this whole class than like making small tweaks to it. So that makes the info that goes into a piece of docs on this field very stable.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,134 @@
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+
+import '../api/model/model.dart';
+
+class MessageIdTracker {
```
chrisbobbe, line 6: I kind of wonder if there's something off-the-shelf for what this is doing. No need to block merging on it, though, unless you or Greg see something that gives much better performance.

(I looked at [`HeapPriorityQueue`](https://api.flutter.dev/flutter/package-collection_collection/HeapPriorityQueue-class.html) but it doesn't have an efficient way to prevent duplicate entries.)

lib/model/message_list.dart
```diff
@@ -383,6 +383,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       if (_messageVisible(message)) {
         _addMessage(message);
       }
+      store.recentSenders.handleMessage(message);
```
chrisbobbe, line 386:
```
autocomplete: Add "recent activity in current topic/stream" criterion
```

The changes in this file in this commit look like they belong in an earlier commit:

```
store: Add `RecentSenders` data structure
```

That earlier commit calls `recentSenders.handleMessage` in one place it needs to, but these are places where it needs to be called too.

chrisbobbe (commented):
Thanks! Comments below.

lib/model/autocomplete.dart
```diff
@@ -188,8 +188,92 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+  /// First checks for the activity in [topic] if provided.
+  ///
+  /// If no [topic] is provided, or the activity in the topic is the same (which
+  /// is extremely rare) or there is no activity in the topic at all, then
+  /// checks for the activity in the stream with [streamId].
+  ///
+  /// Returns a negative number if [userA] has more recent activity than [userB],
+  /// returns a positive number if [userB] has more recent activity than [userA],
+  /// and returns `0` if both [userA] and [userB] have the same recent activity
+  /// (which is extremely rare) or has no activity at all.
```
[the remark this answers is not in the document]
sm-sayedi: Thank you for bringing this up. I tried to cover the two cases where comparing two message `id`s can result in `0`. One is that they're both `0` which means these two users sent no messages. Another is that they're both non-zero, but equal. At first, I thought that messages `id`s are based on the timestamp they're sent, so there's IS a very low probability they're sent at the exact same time. But now after looking at the Zulip API docs carefully, it seems that messages `id`s are unique. So maybe instead of "extremely rare", I can use the term "impossible". 😀

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,134 @@
...
+  }) {
+    final topicsMap = _topicSenders[streamId] ??= {};
+    final sendersMap = topicsMap[topic] ??= {};
+    final idTracker = sendersMap[senderId] ??= MessageIdTracker();
+    idTracker.add(messageId);
+  }
+
+  /// Extracts and keeps track of the necessary data from a [message] only
+  /// if it is a stream message.
+  void handleMessage(Message message) {
```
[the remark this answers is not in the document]
sm-sayedi: Wrote the last line as: **If [message] is not a [StreamMessage], this is a no-op.** The reason I changed it to this is that maybe in the future we will have other types of messages besides `StreamMessage` and `DmMessage`, so in that case, there is no need to update this dartdoc.

sm-sayedi:
Thanks @chrisbobbe for the helpful comments. Revision pushed. Added [#692 comment](https://github.com/zulip/zulip-flutter/pull/692#discussion_r1660916372) and [#692 comment](https://github.com/zulip/zulip-flutter/pull/692#discussion_r1660922011). PTAL.

Note: This PR is rebased on top of #777 because of the build errors mentioned in #771, so the first two commits are from #777.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  @visibleForTesting
+  List<int> get debugIds => _ids;
+
+  @override
+  bool operator == (covariant MessageIdTracker other) {
+    if (identical(this, other)) return true;
+
+    return _ids.equals(other._ids);
+  }
```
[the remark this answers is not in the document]
gnprice: Yeah, the comparison should be done in test code. One way to do that is to keep using `deepEquals`, and take advantage of its feature where an element of a collection can be a condition callback; see its docs, or `git grep 'deepEquals.<'` for some examples in our codebase.

In particular, this doesn't have the right semantics for an `==` override. A `MessageIdTracker` is mutable — indeed it's right there in the name, it tracks a set of message IDs as they change — so two tracker objects shouldn't be thought of as the same even if they happen to have the same set of IDs in them at the moment.

lib/model/recent_senders.dart
gnprice: I think this data structure is fine but there's a couple of things the algorithms should do to be more efficient.

Basically the two kinds of updates we regularly make here are:
* Adding one message (from a `MessageEvent`), which is at least nearly the latest and usually the very latest. (Might be only nearly latest because fetching messages can race with getting events.)
* Adding a whole bunch of messages (from a message fetch).

So we should make those efficient. That means:
* When the new message is the very latest, make that take $O(1)$ time instead of $\Theta(\log n)$.
* When a message fetch gets $n$ messages and there are already $m$ messages, make that not be quadratic — make it take $O((n + m) \log (n + m))$ time, instead of $\Theta((n + m)^2)$. (In fact you can make it $O(n \log n + m)$ time so long as the newly-arrived messages are all older than all the existing messages in the given tracker, which should be the common case here.)

  The key there is, first, thread things through so the `RecentSenders` gets a single method call with the whole batch of messages, rather than one per message. (There'll still be a `handleMessageEvent` for the `MessageEvent` case.) Then, similarly, for each `MessageIdTracker` that's affected, make a single method call to that `MessageIdTracker` with all the relevant messages.

chrisbobbe (commented):
Thanks! One small comment below, and please go ahead and implement Greg's suggestion above about the data structure. Since I think Greg will do the next review, I'll add the integration review label.

chrisbobbe:
Ah and there's this from Greg to be resolved: https://github.com/zulip/zulip-flutter/pull/692#discussion_r1662911244

lib/model/recent_senders.dart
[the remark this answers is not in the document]
sm-sayedi: In the tests, the `==` override is not used directly, but instead, it is used indirectly by `Map` equality in the "Recent Senders" group tests. I tried to replace a small equality check with the suggested feature:
```diff
- final expectedStreamSenders = {
-   1: {
-     10: MessageIdTracker.fromIds([100])
-   }
- };
- check(recentSenders.debugStreamSenders).deepEquals(expectedStreamSenders);
+ check(recentSenders.debugStreamSenders).deepEquals(<Condition<Object?>, Condition<Object?>>{
+   (it) => it.equals(1): (it) => it
+       .isA<Map<int, MessageIdTracker>>()
+       .deepEquals(<Condition<Object?>, Condition<Object?>>{
+         (it) => it.equals(10): (it) => it.isA<MessageIdTracker>(),
+       }),
+ });
```
Yet, for comparing the `ids` of `MessageIdTracker` we need to have an extension on `Subject`. I think this second approach is less readable. We have more complex maps in this test group, such as:
```
final expectedTopicSenders = {
  1: {
    'topic1': {
      10: MessageIdTracker.fromIds([300]),
      20: MessageIdTracker.fromIds([200]),
    },
    'topic2': {
      10: MessageIdTracker.fromIds([100]),
      20: MessageIdTracker.fromIds([400]),
    },
  },
  2: {
    'topic3': {
      20: MessageIdTracker.fromIds([500]),
    }
  }
};
```
The new check for this map would be so complex and hard to follow, I think. 😀 What're your thoughts?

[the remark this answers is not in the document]
gnprice: You can still write `{1: { 10: … } }`, and use a condition callback only for the `MessageIdTracker` value itself. See the `deepEquals` docs:
https://pub.dev/documentation/checks/0.3.0/checks/MapChecks/deepEquals.html

I don't think there's an example quite like this in our tree yet, but you can find some in #783:
https://github.com/zulip/zulip-flutter/pull/783/files#diff-d35f15226b8e53f6c22e599061576c4e84923b365f2fb88666e9344a6eb6deb9

lib/model/autocomplete.dart
```diff
@@ -188,8 +188,82 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+  /// First checks for the activity in [topic] if provided.
+  ///
+  /// If no [topic] is provided, or the activity in the topic is the same (which
+  /// is impossible as message `id` is unique) or there is no activity in the
+  /// topic at all, then checks for the activity in the stream with [streamId].
+  ///
+  /// Returns a negative number if [userA] has more recent activity than [userB],
+  /// returns a positive number if [userB] has more recent activity than [userA],
+  /// and returns `0` if both [userA] and [userB] have the same recent activity
+  /// (which is impossible as message `id` is unique) or have no activity at all.
```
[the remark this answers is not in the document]
gnprice: Yeah, no need to mention the impossible case :-)

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,136 @@
...
+
+  /// A list of distinct message IDs, sorted ascendingly.
+  final List<int> _ids = [];
+
+  /// Add the given message ID to the list, if not already present.
+  void add(int id) {
+    if (_ids.isEmpty) {
+      _ids.add(id);
+    } else {
+      int i = lowerBound(_ids, id);
```
gnprice, line 108: nit: use final when the variable doesn't need to be mutated:
```suggestion
      final i = lowerBound(_ids, id);
```

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,136 @@
...
+
+  // TODO: remove
+
+  /// The maximum id in the tracker list.
+  ///
+  /// Returns `null` if the tracker list is empty.
+  int? get maxId => _ids.lastOrNull;
+
+  @visibleForTesting
+  List<int> get debugIds => _ids;
```
gnprice, line 122: nit: put these above add/remove; see https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense

lib/model/store.dart
```diff
@@ -319,6 +320,8 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
 
   final Map<int, User> users;
 
+  final RecentSenders recentSenders = RecentSenders();
```
gnprice, line 323: It's true that this can be seen as "data about" users, as the section heading here says — but it's also "summaries of messages". Let's put it in that latter section, because the things it most resembles are Unreads and especially RecentConversationsView.

lib/model/store.dart
```diff
@@ -470,6 +474,7 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
       assert(debugLog("server event: message ${jsonEncode(event.message.toJson())}"));
       _messages.handleMessageEvent(event);
       unreads.handleMessageEvent(event);
+      recentSenders.handleMessage(event.message);
       recentDmConversationsView.handleMessageEvent(event);
```
gnprice, line 478: nit: put these calls in the same order the fields are declared

(unless there's some dependency between them that forces an ordering — but if there were then that should be explicitly commented)

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,153 @@
+import 'package:checks/checks.dart';
+import 'package:test/scaffolding.dart';
+import 'package:zulip/model/recent_senders.dart';
+import '../example_data.dart' as eg;
+
+void main() {
+  group('MessageIdTracker', () {
```
gnprice, line 7: nit: order tests to match the code under tests — so this group after the RecentSenders group, given that the MessageIdTracker class comes after the RecentSenders class

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,153 @@
...
+    late MessageIdTracker idTracker;
+
+    void prepare() {
+      idTracker = MessageIdTracker();
+    }
```
gnprice, line 12: This can just be inlined in each test case:
```dart
      final idTracker = MessageIdTracker();
```

Still just one line, and makes each test more explicit and self-contained.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,136 @@
...
+    if (_ids.isEmpty) {
+      _ids.add(id);
+    } else {
+      int i = lowerBound(_ids, id);
+      if (i < _ids.length && _ids[i] == id) return; // the [id] already exists, so do not add it.
+      _ids.insert(i, id);
+    }
+  }
+
+  // TODO: remove
```
gnprice, line 114: If there's no remove, this class can be a lot simpler — it can have just an `int?` for the max ID and still implement the exact same behavior as far as calls to `add` and `maxId` can see.

Then once we're doing that, we might as well drop the class and have `RecentSenders` put ints in its maps, instead of these objects.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,153 @@
...
+
+    test('ids are sorted ascendingly, with maxId pointing to the last element', () {
+      prepare();
+      check(idTracker.debugIds).isEmpty();
+      idTracker.add(1);
+      idTracker.add(9);
+      idTracker.add(0);
+      idTracker.add(5);
+      check(idTracker.debugIds).deepEquals([0, 1, 5, 9]);
+      check(idTracker.maxId).equals(idTracker.debugIds.last);
```
gnprice, line 35: I'd rather write this test in terms of the class's interface, the same interface that the app code needs from it. So that means checking `maxId` instead of peeking at the full list.

This is actually what prompted me to write the comment just above about the absence of `remove` — from the class's actual interface, there's no way to tell that 0, 1, and 5 are still in the list at the end, or that they're in order. Because there's really no way to tell, we don't actually *want* a test that tries to check that — it's not the test's business, or the business of any code but the implementation of the class itself.

For background on that approach, see this section of Zulip's "testing philosophy" doc:

(OTOH after adding `remove`, then the full content of this test case can be had by removing 9 and checking that maxId is now 5, etc.)

gnprice: A further step in that direction is that this whole `MessageIdTracker` class can be taken as an internal implementation detail of `RecentSenders`, and all these tests can be written in terms of `RecentSenders`. Essentially they're all just examples of test cases one can write for `RecentSenders` that involve only a single sender in a single conversation.

That doesn't require them to get any less simple than they are in this version. I'll make separate comments below on the RecentSenders test cases about ways to simplify those.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    test('adding multiple messages', () {
+      prepare();
+      final message1 = eg.streamMessage(
+        stream: eg.stream(streamId: 1),
+        topic: 'topic1',
+        sender: eg.user(userId: 10),
+        id: 300,
+      );
```
gnprice, line 89:
```suggestion
      final stream1 = eg.stream(streamId: 1);
      final user10 = eg.user(userId: 10);
      final message1 = eg.streamMessage(
        stream: stream1, topic: 'topic1', sender: user10, id: 300);
```

Then stream1 and user10 can be reused in the other messages.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,153 @@
...
+            10: MessageIdTracker.fromIds([100]),
+            20: MessageIdTracker.fromIds([400]),
+          },
+        },
+        2: {
+          'topic3': {
+            20: MessageIdTracker.fromIds([500]),
+          }
+        }
+      };
```
gnprice, line 147: This is kind of a lot of data to read, even though this is one test that isn't super complicated.

We'll want more tests than this — especially once we add support for adding messages in bulk rather than one by one (the algorithm discussed at https://github.com/zulip/zulip-flutter/pull/692#discussion_r1662930537), and once we add support for removing messages, both of which will make this data structure's logic more complex. To do that, it'll be crucial to make each test case less of a burden to write.

Fortunately, all this expected data is also totally mechanically determined by the input data above. In this version, you as the test author have to write this all down, and then a reviewer has to compare it item by item against the input data in order to confirm it's correct (so that it's checking the implementation is right, instead of checking it's wrong). But instead you can have the computer do that work.

For an example, see `model/unreads_test.dart`, and in particular the `checkMatchesMessages` helper.

Following that example, both of these expected-data variables, and the `check` calls below, can look instead like:
```dart
  checkMatchesMessages([message1, message2, message3, message4, message5]);
```

Or given the other comment above about how `prepare` and the late variable aren't necessary here, it'd be:
```dart
  checkMatchesMessages(model, [message1, message2, message3, message4, message5]);
```
with a `final model = RecentSenders();` at the top of the test case.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,153 @@
...
+      idTracker.add(1);
+      idTracker.add(9);
+      idTracker.add(0);
+      idTracker.add(5);
+      check(idTracker.debugIds).deepEquals([0, 1, 5, 9]);
+      check(idTracker.maxId).equals(idTracker.debugIds.last);
+    });
+  });
+
+  group('RecentSenders', () {
```
gnprice, line 39: This has no tests for the methods the app relies on for reading data from this class :slightly_smiling_face:. They won't be complicated, but there should be tests for `latestMessageIdOfSenderInStream` and its topic counterpart.

In particular those tests should exercise when there are no such messages — it should detect the difference between returning null for that case, and returning -1 as the interface was in a previous revision of this PR.

Also should exercise when the stream is present but not the topic, and when the stream is absent.

gnprice (commented):
Thanks @sm-sayedi, and thanks @rajveermalviya and @chrisbobbe for the previous revisions!

Comments below. I've read the first three commits:

a20e95f63 autocomplete test [nfc]: Break the loop when event-queue wait is not necessary
1cdaeeaf5 recent_senders: Add the new `MessageIdTracker` and `RecentSenders` data structures
ea22e86ba store/msglist: Add and use `RecentSenders` data structure

and I'll leave the remaining two for a future round:
e9c3b657b autocomplete [nfc]: Factor out `compareNullable`
d6b16c386 autocomplete: Add "recent activity in current topic/stream" criterion

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,167 @@
...
+  /// [MessageIdTracker.add] method. It should be `false` (default) when
+  /// [messages] are the initial messages in a narrow (fetched through
+  /// [MessageListView.fetchInitial]), and `true` when they are the older
+  /// messages in a narrow (fetched through [MessageListView.fetchOlder]).
+  void handleMessages(List<Message> messages, {bool inReverse = false}) {
+    for (int i = inReverse ? messages.length - 1 : 0;
+                 inReverse ? i >= 0              : i < messages.length;
+                 inReverse ? i--                 : i++) {
+      handleMessage(messages[i]);
+    }
```
sm-sayedi, line 43: In the case of `MessageListView.fetchOlder`, by traversing the list of messages in reverse order, message IDs go into the start or the end of the tracker list. But in the normal order, they'll also go in the middle of the tracker list. Now that we use `QueueList` (which takes $O(1)$ for both `addFirst` and `addLast`) as the type of ID tracker list, traversing in reverse will result in optimization.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,167 @@
...
+    idTracker.add(messageId);
+
+    // Track message in topic.
+    final topicsMap = topicSenders[streamId] ??= {};
+    sendersMap = topicsMap[topic] ??= {};
+    idTracker = sendersMap[senderId] ??= MessageIdTracker();
+    idTracker.add(messageId);
+  }
+
+  void handleDeleteMessageEvent(DeleteMessageEvent event, Map<int, Message> cachedMessages) {
```
sm-sayedi, line 66: Zulip Web seems not to delete the recently deleted message id from the tracker list, but I am unsure whether to follow them exactly!

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,167 @@
...
+    idTracker.remove(messageId);
+    if (idTracker.maxId == null) sendersMap.remove(senderId);
+    if (sendersMap.isEmpty) topicsMap.remove(topic);
+    if (topicsMap.isEmpty) topicSenders.remove(streamId);
+  }
+}
+
+class MessageIdTracker {
+  /// A list of distinct message IDs, sorted ascendingly.
+  final QueueList<int> _ids = QueueList.from([]);
```
sm-sayedi, line 126: I think using `QueueList` is the best data structure for this scenario. It has $O(1)$ time complexity for both `addFirst` and `addLast`, as well as when accessing elements by index.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,167 @@
...
+  /// The maximum id in the tracker list, or `null` if the list is empty.
+  int? get maxId => _ids.lastOrNull;
+
+  /// Add the message ID to the tracker list at the proper place, if not already present.
+  ///
+  /// Optimized, taking O(1) time, for the cases where that place is the start
+  /// (message fetched through [MessageListView.fetchOlder]) or the end (message
+  /// fetched through [MessageListView.fetchInitial] or
+  /// [PerAccountStore.handleEvent]), because those are the common cases for a
+  /// message the app receives. May take O(n) time in some rare cases.
```
sm-sayedi, line 137:
```
...May take O(n) time in some rare cases.
```
Looking at the codebase in the current stage, I think this will never happen as the only way the app receives a message is through the methods mentioned above.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,167 @@
...
+      _ids.addLast(id);
+    } else {
+      _ids.insert(i, id);
+    }
+  }
+
+  void remove(int id) => _ids.remove(id);
+
+  @override
+  bool operator ==(covariant MessageIdTracker other) {
```
sm-sayedi, line 156: I left `==` operator in this revision too. By using `checkMatchesMessages` in `recent_senders_test.dart`, I wasn't able to avoid `==` and instead user `deepEquals` with a `Condition` callback.

sm-sayedi (commented):
Thank you @gnprice for the review! Just pushed the revision. PTAL.

test/model/recent_senders_checks.dart
```diff
@@ -0,0 +1,29 @@
...
+  Subject<Map<int, Map<String, Map<int, MessageIdTracker>>>> get topicSenders =>
+    has((rs) => rs.topicSenders, 'topicSenders');
+
+  Subject<int?> latestMessageIdOfSenderInStream({
+    required int streamId,
+    required int senderId,
+  }) => has(
+          (rs) => rs.latestMessageIdOfSenderInStream(streamId: streamId,
+            senderId: senderId),
+          'latestMessageIdOfSenderInStream',
```
gnprice, line 17: Generally we use these "checks" extensions only for getters, not method calls. One reason is that that "name" argument to `has` becomes complicated: just the name of the method is no longer adequate to describe where the problem is, in the error output when the test fails.

Instead, just call the method before passing to `check`, as in `check(model.foo(1, 2)).equals(3)`.

test/model/recent_senders_checks.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:checks/checks.dart';
+import 'package:zulip/model/recent_senders.dart';
+
+extension RecentSendersChecks on Subject<RecentSenders> {
+  Subject<Map<int, Map<int, MessageIdTracker>>> get streamSenders =>
+    has((rs) => rs.streamSenders, 'streamSenders');
+
+  Subject<Map<int, Map<String, Map<int, MessageIdTracker>>>> get topicSenders =>
+    has((rs) => rs.topicSenders, 'topicSenders');
```
gnprice, line 9: nit: follow maximally-boring, uniform pattern found on other checks-extensions:
```suggestion
  Subject<Map<int, Map<int, MessageIdTracker>>> get streamSenders => has((rs) => rs.streamSenders, 'streamSenders');
  Subject<Map<int, Map<String, Map<int, MessageIdTracker>>>> get topicSenders => has((rs) => rs.topicSenders, 'topicSenders');
```

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+
+      model.handleMessages(messages);
+
+      check(model)
+        ..latestMessageIdOfSenderInStream(
+            streamId: stream1.streamId, senderId: user1.userId).equals(300)
+        ..latestMessageIdOfSenderInStream(
+            streamId: stream1.streamId, senderId: user2.userId).equals(null)
+        ..latestMessageIdOfSenderInStream(
+            streamId: stream2.streamId, senderId: user1.userId).equals(null);
```
gnprice, line 116: Let's write these as three separate test cases (three `test` calls). That way they each get their own name that indicates the key point they're testing.

The setup can be simplified for the latter two cases, just adding a single message.

gnprice: Alternatively the latter cases could get a short one-line comment each, saying what they're testing: "unknown sender", "unknown stream". That may be better, particularly when I look at the similar `…InTopic` case below.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+
+      model.handleMessages(messages);
+
+      check(model)
+        ..latestMessageIdOfSenderInTopic(streamId: stream1.streamId,
+            topic: 'a', senderId: user1.userId).equals(300)
+        ..latestMessageIdOfSenderInTopic(streamId: stream1.streamId,
+            topic: 'a', senderId: user2.userId).equals(null)
+        ..latestMessageIdOfSenderInTopic(streamId: stream1.streamId,
+            topic: 'a', senderId: user2.userId).equals(null)
```
gnprice, line 139: This is the same check twice, right?

Let's take that as another indicator that these test cases (this and the `…InStream` case above) can be improved in their clarity. :slightly_smiling_face: When the test is really clear, it should be easy for you to scan what it's saying and hard to miss that two checks are actually the same.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+
+    test('just process a stream message', () {
+      final model = RecentSenders();
+      final dmMessage = eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]);
+      final streamMessage = eg.streamMessage(id: 100, sender: eg.user(userId: 10),
+        stream: eg.stream(streamId: 1), topic: 'topic');
+
+      model.handleMessage(dmMessage);
+      model.handleMessage(streamMessage);
+      checkMatchesMessages([streamMessage], model);
```
gnprice, line 48: actual first, then expected:
```suggestion
      checkMatchesMessages(model, [streamMessage]);
```

That's the same order used by `check` itself. The consistency is helpful for not getting them swapped.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+    test('starts with empty stream and topic senders', () {
+      final model = RecentSenders();
+      checkMatchesMessages([], model);
+    });
+
+    test('just process a stream message', () {
+      final model = RecentSenders();
+      final dmMessage = eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]);
+      final streamMessage = eg.streamMessage(id: 100, sender: eg.user(userId: 10),
+        stream: eg.stream(streamId: 1), topic: 'topic');
```
gnprice, line 44:
```suggestion
      final streamMessage = eg.streamMessage();
```

None of these details of the stream message are relevant to what the test is saying, right? It just needs them to have *some* plausible values.

That's then the job of `eg` aka `example_data.dart`. That file supplies boring but plausible data, so that tests can stay focused on expressing only the details that are relevant to the rest of the test.

(In the case of `eg.dmMessage` the `from` and `to` are required only because it's harder to come up with any truly boring canonical values for them to take — most tests do care whether it's a 1:1 or group DM, or whether it's to or from the self-user. And it's plausible only if the self-user is among the recipients, and the self-user in a given test case might not be `eg.selfUser`.)

test/model/recent_senders_test.dart
gnprice: One other way to make these more compact and give them less visual repetition: since you're already specifying the stream IDs and user IDs explicitly above, just use the numbers here.

While you're at it, though: give the user IDs distinct values from the stream IDs. That'll avoid confusion for the reader, and also mean this would catch if the code under test somehow swapped the roles of stream ID and user ID. (That's why in my suggestion at https://github.com/zulip/zulip-flutter/pull/692#discussion_r1664653053 I had `streamId: 1` but `userId: 10`.)

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+
+      model.handleMessage(dmMessage);
+      model.handleMessage(streamMessage);
+      checkMatchesMessages([streamMessage], model);
+    });
+
+    test('add a batch of messages', () {
```
gnprice, line 51: This should test adding a batch when the data structure isn't empty.

In particular let's exercise all the different cases that appear in the logic of the code being tested: the new messages for a given ID tracker are all at the start, or they're all at the end, or they're interspersed among the existing messages. Also cover where some of the new messages are already present.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+}
+
+void main() {
+  group('RecentSenders', () {
+    test('starts with empty stream and topic senders', () {
+      final model = RecentSenders();
+      checkMatchesMessages([], model);
+    });
+
+    test('just process a stream message', () {
```
gnprice, line 40: This is clearest as two cases:
* stream message gets included
* DM gets ignored

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+import 'package:zulip/api/model/model.dart';
+import 'package:zulip/model/recent_senders.dart';
+import '../example_data.dart' as eg;
+import 'recent_senders_checks.dart';
+
+void checkMatchesMessages(List<Message> messages, RecentSenders model) {
+  final Map<int, Map<int, MessageIdTracker>> expectedStreamSenders = {};
+  final Map<int, Map<String, Map<int, MessageIdTracker>>> expectedTopicSenders = {};
+  for (final message in messages) {
+    if (message is! StreamMessage) continue;
```
gnprice, line 13: Clearer if this throws on a DM, rather than ignore it. The test cases shouldn't be passing DMs among the expected messages — if they do, the test case will be confusing because it may look to the reader as if the DM is expected to be there in the data structure.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+        eg.streamMessage(stream: stream1, topic: 'b', sender: user1, id: 100),
+        eg.streamMessage(stream: stream1, topic: 'a', sender: user2, id: 200),
+      ];
+
+      model.handleMessages(messages);
+      checkMatchesMessages(messages, model);
+
+      model.handleDeleteMessageEvent(
+        DeleteMessageEvent(id: 1, messageIds: [300, 200],
+          messageType: MessageType.stream, streamId: stream1.streamId, topic: 'a'),
```
gnprice, line 87: See `eg.deleteMessageEvent` to simplify this.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,201 @@
...
+            topic: 'a', senderId: user2.userId).equals(null)
+        ..latestMessageIdOfSenderInTopic(streamId: stream1.streamId,
+            topic: 'a', senderId: user2.userId).equals(null)
+        ..latestMessageIdOfSenderInTopic(streamId: stream1.streamId,
+            topic: 'b', senderId: user1.userId).equals(null)
+        ..latestMessageIdOfSenderInTopic(streamId: stream2.streamId,
+            topic: 'a', senderId: user1.userId).equals(null);
+    });
+
+    group('MessageIdTracker', () {
```
gnprice, line 146: Do these three tests cover anything that isn't already covered by the tests above? It looks like they don't — I think each of them is very similar to some test above and is basically just testing the same thing.

So let's write each of those tests just once. I think the contents of these tests are successfully using the approach I described at https://github.com/zulip/zulip-flutter/pull/692#discussion_r1664646094 — but then at that point they can just be treated as more tests of `RecentSenders` :slightly_smiling_face:, and they can also be deduplicated with other `RecentSenders` tests that do the same thing.

Relatedly, `MessageIdTracker` can be made private. Or `@visibleForTesting` if the tests have a need for it, but I think after my comment above on `checkMatchesMessages` they won't.

gnprice (commented):
Thanks @sm-sayedi for the revision! Comments below. This is definitely getting closer.

This round is again based on just the first half of the branch:
034678ada autocomplete test [nfc]: Break the loop when event-queue wait is not necessary
028906174 recent_senders: Add the new `MessageIdTracker` and `RecentSenders` data structures

and I haven't yet read closely the second half:
34570707a autocomplete [nfc]: Factor out `compareNullable`
9e3d4ec3f autocomplete: Add "recent activity in current topic/stream" criterion

lib/model/recent_senders.dart
sm-sayedi: I am having a slightly hard time coming up with an efficient solution for this. 😃 
Here's what I have thought of the `handleMessages` method so far, especially, determining all the affected `MessageIdTracker`s:
```dart
    final streamTracker = <(int, int), List<int>>{};
    final topicTracker = <(int, String, int), List<int>>{};

    for (final message in messages) {
      if (message is! StreamMessage) return;
      final StreamMessage(:streamId, :topic, :senderId, id: int messageId) = message;
      (streamTracker[(streamId, senderId)] ??= []).add(messageId);
      (topicTracker[(streamId, topic, senderId)] ??= []).add(messageId);
    }

    for (final entry in streamTracker.entries) {
      final (streamId, senderId) = entry.key;
      final tracker = (streamSenders[streamId] ??= {})[senderId] ??= MessageIdTracker();
      tracker.addAll(entry.value);
    }
    for (final entry in topicTracker.entries) {
      final (streamId, topic, senderId) = entry.key;
      final tracker = ((topicSenders[streamId] ??= {})[topic] ??= {})[senderId] ??= MessageIdTracker();
      tracker.addAll(entry.value);
    }
```
But as you can see, it already has 3 for loops. Then when I want to have a single method for adding the relevant message IDs to a `MessageIdTracker`, there's the need to check that we don't add the already-existing IDs from the incoming `id`s list. So I think it again needs to call `add` method for each message.

gnprice: Having 3 loops is fine - they're happening in sequence, not nested, so the total cost is only the sum of the three loops' respective cost. And for example three times $O(f(n))$ is still $O(f(n))$.

Then for the implementation of addAll, it doesn't need to be $O(n+m)$ (for $n$ new messages and $m$ existing messages) — the goal is only for it to be $O((n+m) \log (n+m))$. In the implementation of `add`, the expensive part is the actual insertion, moving elements around, which can take $m$ steps. The search is a binary search, which is much cheaper at $O(\log m)$ steps. So addAll can do $n$ searches just fine — the thing it has to avoid is $n$ individual insertions.

sm-sayedi: Here's the revised implementation for `addAll`:
```dart
  void addAll(List<int> ids) {
    final idsToBeAdded = <int>[];

    if (_ids.isEmpty) {
      _ids.addAll(ids);
      return;
    }

    for (final id in ids) {
      final i = lowerBound(_ids, id);
      if (i < _ids.length && _ids[i] == id) {
        // The ID is already present. Nothing to do.
        continue;
      }
      idsToBeAdded.add(id);
    }

    _ids.insertAll(0, idsToBeAdded);
  }
```
Am I on the right track?

gnprice: Yeah, that's a good direction. What's a bound (in big-O notation) on how long that will take?

Then after that insertAll, the IDs in the tracker may be out of order. So that will need to be fixed.

sm-sayedi: For `addAll`, I think it takes $O(nlogm) + O(n + m)$ which is $O(nlogm)$ at the end of the day, if I am not wrong.

But thinking about the current implementation in this PR where there's `inReverse` in `handleMessages`, I think it takes less time possible, just $O(n)$. For each message id, it just adds it to the start or the end of a `MessageIdTracker` list (thanks to `inReverse`). Now that we use `QueueList` (`Deque`), it just takes $O(1)$ and eventually making the whole adding operation take $O(n)$!

gnprice: > But thinking about the current implementation in this PR where there's `inReverse` in `handleMessages`, I think it takes less time possible, just O(n). For each message id, it just adds it to the start or the end of a `MessageIdTracker` list (thanks to `inReverse`). Now that we use `QueueList` (`Deque`), it just takes O(1) and eventually making the whole adding operation take O(n)!

The thing is that the messages might not go at the very start or end. For example, consider what happens in a stream's tracker if the user first visits a topic narrow within that stream (and maybe scrolls back through a bunch of history), then visits the stream narrow itself.

We're also getting close to properly supporting moved messages, #150 — see #787. So after both this PR and that one are merged (or perhaps as part of a later revision of #787 after this PR is merged), we'll be handling message moves on this RecentSenders data structure too. Then possible scenarios include:
* Some older messages get moved into a topic. They go before lots of more recent history in the topic; but our client hasn't yet loaded any of that history. The topic ID tracker ends up with just those older messages.

  Then later the user visits that topic narrow. The first fetched messages are after, not before, all the existing messages. They scroll up further… and the messages from `fetchOlder` go in between a bunch of older messages and a bunch of newer messages.

* Some messages get moved into a topic, after the user has already scrolled through a lot of history in the topic so we have a long list of messages there. The move won't go through this `handleMessages` code path exactly, so that's a problem we could wait to solve until we implement moves; but we'll need to solve it then, and the same solution that handles the other scenarios will handle this one.

gnprice: > For `addAll`, I think it takes O(nlogm)+O(n+m) which is O(nlogm) at the end of the day, if I am not wrong.

Yep, the `addAll` you quoted above takes time bounded by $O(n \log m) + O(n + m)$ ­— the first term for the loop that calls `lowerBound`, then the second for the `insertAll` call.

That can be simplified to
$O(n \log m + n + m)$;
and then to
$O(n \log m + m)$
because $n$ is bounded by $n \log m$.

It's not true that it's $O(n \log m)$, though. Either $m$ could be bigger than $n \log m$, or vice versa, so neither term can be simplified away.

But the goal I mentioned above was a bound of $O((n + m) \log (n + m))$, so it's still under that bound and we're still in good shape.

The `addAll` above still needs another bit of logic at the end to deal with the fact that `insertAll` may leave the IDs in the tracker out of order, though.

(Formatting tip: write `\log`, with a backslash, to get it formatted as the log operator, with an upright font and spacing around it.)

sm-sayedi: Thanks for the explanation, it was really helpful; also thanks for the formation tip!

Here's the new revision for `addAll`, still with $O(n\log m + m)$:

```dart
  void addAll(List<int> ids) {
    if (_ids.isEmpty) {
      _ids.addAll(ids);
      return;
    }

    final idsToBeAdded = <int>[];
    for (final id in ids) {
      final i = lowerBound(_ids, id);
      if (i < _ids.length && _ids[i] == id) {
        // The ID is already present. Nothing to do.
        continue;
      }
      idsToBeAdded.add(id);
    }

    _ids = setUnion(_ids, idsToBeAdded);
  }
```

gnprice: Great. Looking forward to seeing the revision!

sm-sayedi:
Thanks @gnprice for the last review. Revision pushed. PTAL.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,155 @@
...
+      if (topicsMap?.isEmpty ?? false) topicSenders.remove(streamId);
+    }
+  }
+}
+
+@visibleForTesting
+class MessageIdTracker {
+  /// A list of distinct message IDs, sorted ascendingly.
+  @visibleForTesting
+  QueueList<int> ids = QueueList.from([]);
```
sm-sayedi, line 114: Is it fine to add `@visibleForTesting` to the member itself, instead of making it private and introducing another getter for testing?

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,155 @@
...
+  /// Add the messages IDs to the tracker list at the proper place, if not present.
+  ///
+  /// [newIds] should be sorted ascendingly.
+  void addAll(List<int> newIds) {
+    if (ids.isEmpty) {
+      ids.addAll(newIds);
+      return;
+    }
+    ids = setUnion(ids, newIds);
+  }
```
sm-sayedi, line 146: This now takes $O(n + m)$ as `setUnion` gets rid of the shared IDs too in addition to keeping the resulting list in order.

Also, would it be better to replace `ids.addAll(newIds)` with `ids = newIds`? It should align it with the last line of the method.

lib/model/message_list.dart
```diff
@@ -403,6 +403,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       numAfter: 0,
     );
     store.reconcileMessages(result.messages);
+    store.recentSenders.handleMessages(result.messages);
```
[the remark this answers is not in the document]
gnprice: (Just to cross-link from this subthread: that's #824. Thanks!)

test/model/message_list_test.dart
```diff
@@ -233,6 +252,27 @@ void main() {
...
+  test('fetchOlder, recent senders track all the messages', () async {
+    const narrow = CombinedFeedNarrow();
+    await prepare(narrow: narrow);
+    final initialMessages = List.generate(10, (i) => eg.streamMessage(id: 100 + i));
+    await prepareMessages(foundOldest: false, messages: initialMessages);
+
+    final oldMessages = List.generate(10,
+      (i) => eg.streamMessage(id: 90 + i))
+        // Not subscribed to the stream with id 10.
+        ..add(eg.streamMessage(stream: eg.stream(streamId: 10)));
```
gnprice, line 265: This last message will have a larger message ID than any of the list's existing messages (the ones in `initialMessages`.

That apparently doesn't break this test, but it is unrealistic and could be confusing. So just specify an ID to avoid that. (And adjust the other IDs to make room; as is, it belongs between 99 and 100.)

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,229 @@
...
+      final model = RecentSenders();
+      final streamMessage = eg.streamMessage();
+      final dmMessage = eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]);
+
+      model.handleMessage(streamMessage);
+      model.handleMessage(dmMessage);
+      checkMatchesMessages(model, [streamMessage]);
+    });
+
+    group('add a batch of messages', () {
```
gnprice, line 75: nit: name tests after the code's identifiers for what they're testing (here and a few other places):
```suggestion
    group('handleMessages', () {
```

That helps make it easy to navigate from a piece of the app's code to its tests and back.

gnprice: Better yet: call this `RecentSenders.handleMessages`, and remove the outer `group` layer.

There's no real need for a `RecentSenders` group around the outside — this whole file is for testing `RecentSenders`, and given it's called `recent_senders_test.dart` that can be expected to continue :slightly_smiling_face:

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,229 @@
...
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
+        ];
+        setupModel(existingMessages);
+
+        final messages = [
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
+          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 300),
+          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 500),
+        ];
```
gnprice, line 132: Does this do what the test name says it does? It looks like it doesn't:

stream 1, topic 'a' was [200, 400], and gets 100, which goes at the beginning
stream 1, topic 'b' was absent and gets some messages

Interspersed would look like: some tracker had [100, 300, 500], and gets [200, 400].

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,229 @@
...
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
+        ];
+        setupModel(existingMessages);
+
+        final messages = [
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
+          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 200),
+          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 400),
+          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 500),
```
gnprice, line 151: Similarly here. Though this one has another problem, which is that the data is unrealistic: some of these are the same message IDs again but in a different place.

(That isn't actually completely impossible, given the fetch-vs-event race: the messages could have moved. But if we were going to write a test that exercised that, it'd be a test that was focused on that scenario. And before #150 it won't make sense to try to test that anyway.)

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,229 @@
...
+        final existingMessages = [
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 300),
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 400),
+        ];
+        setupModel(existingMessages);
+
+        final messages = [
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
+          eg.streamMessage(stream: stream1, topic: 'b', sender: user10, id: 200),
+        ];
```
gnprice, line 97: One thing going on in several of these test cases is that the thing they're about (as described by their names) is something that's all about a single tracker at a time; but then they also have some messages that go in a different tracker.

That's two pretty much separate areas of the code being tested, and so the tests are cleanest and easiest to think about if those ideas are covered in separate tests. So these test cases would all be focused on a single tracker; but then there should also be a couple of test cases that show that `handleMessages` correctly handles taking messages that belong in several different trackers, and distributing them across those.

Also that it accepts and ignores DMs — that they don't stop the stream/channel messages in the same batch from getting processed normally.

test/model/recent_senders_test.dart
gnprice: Oh, I guess they are interspersed on the stream tracker.

Still, I think the test is easiest to read and think about if there's just one tracker involved of each kind. So changing the two `topic: 'b'` to `topic: 'a'` would do it.

gnprice (commented):
Thanks for the revision! This data structure is getting close to merge now.

Comments below; most of them small, a couple of them more significant.

Then let's split up the PR here, since this thread has gotten long:
* update the PR to contain only those first two commits;
* edit the PR description so it doesn't close the issue;
* open a new PR for the full branch, i.e. effectively for the latter two commits but stacked atop this PR.

That way we'll be able to merge the data structure soon!

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,155 @@
...
+    final StreamMessage(:streamId, :topic, :senderId, id: int messageId) = message;
+
+    ((streamSenders[streamId] ??= {})
+      [senderId] ??= MessageIdTracker()).add(messageId);
+    (((topicSenders[streamId] ??= {})[topic] ??= {})
+      [senderId] ??= MessageIdTracker()).add(messageId);
+  }
+
+  void handleDeleteMessageEvent(DeleteMessageEvent event, Map<int, Message> cachedMessages) {
+    if (event.messageType != MessageType.stream) return;
```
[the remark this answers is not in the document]
gnprice: Also started an `#api documentation` thread to confirm and clarify the fact I stated above:
https://chat.zulip.org/#narrow/stream/19-documentation/topic/delete_message.20events/near/1895069

sm-sayedi:
Thanks @gnprice for the review. Updates pushed. Please have a look.

test/model/recent_senders_test.dart
sm-sayedi: My bad. Those messages were supposed to be in `topic: 'a'`, but after copy-pasting it from the previous revision, I forgot to rename it.
And the new messages are supposed to scatter around and between the old ones.

test/model/recent_senders_test.dart
sm-sayedi: As before, these messages were also meant to be in `topic: 'a'`. 🙂 Thanks for catching.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,154 @@
...
+
+      idTracker?.removeAll(messages);
+      if (idTracker?.maxId == null) sendersMap?.remove(senderId);
+      if (sendersMap?.isEmpty ?? false) topicsMap?.remove(topic);
+      if (topicsMap?.isEmpty ?? false) topicSenders.remove(streamId);
+    }
+  }
+}
+
+class MessageIdTracker {
```
sm-sayedi, line 108: Making this class private would cause the following lint message, thus failing the CI:
```
Invalid use of a private type in a public API.
Try making the private type public, or making the API that uses the private type also be private.
```

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,329 @@
...
+      });
+
+      test('batch with both DM and stream messages -> ignores DM, processes stream messages', () {
+        final existingMessages = [
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
+        ];
+        setupModel(existingMessages);
+
+        final dmMessage = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser], id: 200);
```
gnprice, line 150: bad test data: message ID collision

(and can't be the same message seen twice, because stream/channel messages can't turn into DMs or vice versa)

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,329 @@
...
+        setupModel(existingMessages);
+
+        final messages = [
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 100),
+          eg.streamMessage(stream: stream1, topic: 'a', sender: user10, id: 200),
+        ];
+        model.handleMessages(messages);
+
+        checkMatchesMessages(model,
+          [...existingMessages, ...messages]..sort((m1, m2) => m1.id.compareTo(m2.id)));
```
gnprice, line 85: These handleMessages tests are pretty verbose, with a lot of stuff in them that's the same from test to test and isn't what the test is about. I'll push a couple of added commits that factor that out, so that it's a lot easier for the reader to scan these tests and see what they're saying.

test/model/recent_senders_test.dart
```diff
@@ -0,0 +1,329 @@
...
+        ];
+        model.handleMessages(messages);
+
+        checkMatchesMessages(model,
+          [...existingMessages, ...messages..remove(dmMessage)]..sort((m1, m2) => m1.id.compareTo(m2.id)));
+      });
+    });
+
+    group('multiple trackers', () {
+      test('batch goes before the existing messages', () {
```
gnprice, line 164: I don't think it's all that helpful to take the single-tracker tests and make a copy of each of them that distributes across a couple of different topics. What I asked for at https://github.com/zulip/zulip-flutter/pull/692#discussion_r1683760587 was:
> there should also be a couple of test cases that show that `handleMessages` correctly handles taking messages that belong in several different trackers, and distributing them across those.

These test cases don't really do that. They don't test that `handleMessages` correctly handles:
* having messages in several streams
* or messages from several senders
* or creating a new tracker where there wasn't one already;

really the one thing they add is testing that it handles getting messages in multiple topics of one stream.

Conversely there isn't a need to repeat the testing of all the different ways the old and new messages in a given tracker can be related; that logic all operates within each tracker separately, so the single-tracker tests cover it.

These tests do add a lot of lines of code. One effect of that is that it takes some work to scan through and see all the things the tests are saying, in order to get the picture of what scenarios are covered and what scenarios aren't. One reason making tests concise and focused (as discussed in comments just above this one) is important is that it helps with detecting missing scenarios, like here.

I'll push another commit replacing these with tests that cover those.

gnprice (commented):
Thanks for the revision! This is very close now. Comments below.

For the remaining comments on the tests, I decided it'd be clearest to demonstrate rather than describe, so I'll push to this branch some added commits that make those changes. Then please leave those in the branch, after your main commits, while acting on the several other comments below. For the two comments that request followup commits, those can go after my test commits.

sm-sayedi:
Thank you @gnprice for the review and the added commits; they're really helpful. New revision pushed. PTAL.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,148 @@
...
+  /// [newIds] should be sorted ascendingly.
+  void addAll(QueueList<int> newIds) {
+    if (ids.isEmpty) {
+      ids = newIds;
+      return;
+    }
+    ids = setUnion(ids, newIds);
+  }
+
+  void removeAll(QueueList<int> idsToRemove) {
```
gnprice, line 139: nit: this doesn't use that it's a QueueList, so might as well take a general List (and the caller might as well make plain default lists with `[]`, rather than QueueLists)

gnprice (commented):
Thanks for the revision! No remaining comments on the behavior; just three small comments below, and then this will be ready for merge.

lib/model/recent_senders.dart
```diff
@@ -0,0 +1,148 @@
...
+
+    final DeleteMessageEvent(:streamId!, :topic!) = event;
+    for (final entry in messagesByUser.entries) {
+      final MapEntry(key: senderId, value: messages) = entry;
+
+      final sendersByStream = streamSenders[streamId];
+      final messagesBySenderInStream = sendersByStream?[senderId];
+      messagesBySenderInStream?.removeAll(messages);
+      if (messagesBySenderInStream?.maxId == null) sendersByStream?.remove(senderId);
+      if (sendersByStream?.isEmpty ?? false) streamSenders.remove(streamId);
```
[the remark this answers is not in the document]
sm-sayedi: I also moved the lines related to the topic outside of the loop as they were not dependent on `senderId` or `messages` either.

lib/model/recent_senders.dart
sm-sayedi: The motivation was that we use `QueueList` internally as well as in `addAll`, so I thought passing a `QueueList` in `removeAll` would align with them. Anyway, changed it back to `List`. 🙂

sm-sayedi:
Thanks @gnprice for the review. Revision pushed.

gnprice:
Looks good — merging! Thanks again for all your work on this. On to #828, the last stage, so that we can get this feature out to users :slightly_smiling_face: 

I added some small NFC commits on top; please take a look through those:

73c827cd8 recent_senders [nfc]: Rename some foosByBar that mean foosInBar

A map of "foos by bar" is a map where each key is a "bar", and
each value is a "foo" -- the phrase can be read as an abbreviation
of "foos indexed by bar" or "foos, where each foo is indexed by
its respective bar".

So in this function `messagesByUser` is accurately named because
a key represents a user and a value represents some messages.

But these other variables should be e.g. `topicsInStream`, not
`topicsByStream`.  That one isn't a map where the key is a stream
and the value is a topic or topics; rather it's a collection of
topics (and more data about them) which is the collection
specifically of topics in the stream we're acting on.


eca8af9dd recent_senders [nfc]: Add and revise some docs


dbb2dcfbf recent_senders [nfc]: Write "sorted ascending"

PR #694, opened 2024-05-21
 lightbox test: Add video player regression tests 

rajveermalviya:
Adds tests mentioned in https://github.com/zulip/zulip-flutter/pull/587#discussion_r1597195082

### DIFF
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -277,20 +277,7 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
   }
 
   void _handleVideoControllerUpdate() {
-    setState(() {
-      // After 'controller.seekTo' is called in 'Slider.onChangeEnd' the
-      // position indicator switches back to the actual controller's position
-      // but since the call 'seekTo' completes before the actual controller
-      // updates are notified, the position indicator that switches to controller's
-      // position can show the older position before the call to 'seekTo' for a
-      // single frame, resulting in a glichty UX.
-      //
-      // To avoid that, we delay the position indicator switch from '_sliderValue' to
-      // happen when we are notified of the controller update.
-      if (_isSliderDragging && _sliderValue == widget.controller.value.position) {
-        _isSliderDragging = false;
-      }
-    });
+    setState(() {});
   }
 
   static String _formatDuration(Duration value) {
@@ -325,15 +312,15 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
               _sliderValue = Duration(milliseconds: value.toInt());
             });
           },
-          onChangeEnd: (value) {
+          onChangeEnd: (value) async {
             final durationValue = Duration(milliseconds: value.toInt());
-            setState(() {
-              _sliderValue = durationValue;
-            });
-            widget.controller.seekTo(durationValue);
-
-            // The toggling back of '_isSliderDragging' is omitted here intentionally,
-            // see '_handleVideoControllerUpdates'.
+            await widget.controller.seekTo(durationValue);
+            if (mounted) {
+              setState(() {
+                _sliderValue = durationValue;
+                _isSliderDragging = false;
+              });
+            }
           },
         ),
       ),
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -248,6 +248,31 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
   }
 }
 
+class VideoDurationLabel extends StatelessWidget {
+  const VideoDurationLabel(this.duration, {
+    super.key,
+    this.semanticsLabel,
+  });
+
+  final Duration duration;
+  final String? semanticsLabel;
+
+  @visibleForTesting
+  static String formatDuration(Duration value) {
+    final hours = value.inHours.toString().padLeft(2, '0');
+    final minutes = value.inMinutes.remainder(60).toString().padLeft(2, '0');
+    final seconds = value.inSeconds.remainder(60).toString().padLeft(2, '0');
+    return '${hours == '00' ? '' : '$hours:'}$minutes:$seconds';
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Text(formatDuration(duration),
+      semanticsLabel: semanticsLabel,
+      style: const TextStyle(color: Colors.white));
+  }
+}
+
 class _VideoPositionSliderControl extends StatefulWidget {
   final VideoPlayerController controller;
 
@@ -280,13 +305,6 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
     setState(() {});
   }
 
-  static String _formatDuration(Duration value) {
-    final hours = value.inHours.toString().padLeft(2, '0');
-    final minutes = value.inMinutes.remainder(60).toString().padLeft(2, '0');
-    final seconds = value.inSeconds.remainder(60).toString().padLeft(2, '0');
-    return '${hours == '00' ? '' : '$hours:'}$minutes:$seconds';
-  }
-
   @override
   Widget build(BuildContext context) {
     final currentPosition = _isSliderDragging
@@ -294,8 +312,8 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
       : widget.controller.value.position;
 
     return Row(children: [
-      Text(_formatDuration(currentPosition),
-        style: const TextStyle(color: Colors.white)),
+      VideoDurationLabel(currentPosition,
+        semanticsLabel: "Current position"),
       Expanded(
         child: Slider(
           value: currentPosition.inMilliseconds.toDouble(),
@@ -324,8 +342,8 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
           },
         ),
       ),
-      Text(_formatDuration(widget.controller.value.duration),
-        style: const TextStyle(color: Colors.white)),
+      VideoDurationLabel(widget.controller.value.duration,
+        semanticsLabel: "Video duration"),
     ]);
   }
 }
--- a/test/widgets/lightbox_test.dart
+++ b/test/widgets/lightbox_test.dart
@@ -1,17 +1,25 @@
 import 'dart:async';
 
 import 'package:checks/checks.dart';
+import 'package:clock/clock.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter/material.dart';
 import 'package:plugin_platform_interface/plugin_platform_interface.dart';
 import 'package:video_player_platform_interface/video_player_platform_interface.dart';
 import 'package:video_player/video_player.dart';
+import 'package:zulip/model/localizations.dart';
 import 'package:zulip/widgets/lightbox.dart';
 import 'package:zulip/widgets/store.dart';
 
 import '../example_data.dart' as eg;
 import '../model/binding.dart';
+import 'dialog_checks.dart';
+
+const kTestVideoUrl = "https://a/video.mp4";
+const kTestUnsupportedVideoUrl = "https://a/unsupported.mp4";
+const kTestVideoDuration = Duration(seconds: 10);
 
 class FakeVideoPlayerPlatform extends Fake
     with MockPlatformInterfaceMixin
@@ -22,21 +30,63 @@ class FakeVideoPlayerPlatform extends Fake
     VideoPlayerPlatform.instance = instance;
   }
 
-  static const int _textureId = 0xffffffff;
+  static const int _kTextureId = 0xffffffff;
 
   StreamController<VideoEvent> _streamController = StreamController<VideoEvent>();
+  bool _hasError = false;
+  Duration _lastSetPosition = Duration.zero;
+  Stopwatch? _stopwatch;
+
+  List<String> get callLog => _callLog;
+  final List<String> _callLog = [];
 
   bool get initialized => _initialized;
   bool _initialized = false;
 
-  bool get isPlaying => _isPlaying;
-  bool _isPlaying = false;
+  bool get isCompleted => _isCompleted;
+  bool _isCompleted = false;
+
+  bool get isPlaying => _stopwatch?.isRunning ?? false;
+
+  Duration get position {
+    assert(_stopwatch != null);
+    final pos = _stopwatch!.elapsed + _lastSetPosition;
+    return pos >= kTestVideoDuration ? kTestVideoDuration : pos;
+  }
 
   void reset() {
     _streamController.close();
     _streamController = StreamController<VideoEvent>();
+    _hasError = false;
+    _lastSetPosition = Duration.zero;
+    _stopwatch?.stop();
+    _stopwatch?.reset();
+    _callLog.clear();
     _initialized = false;
-    _isPlaying = false;
+    _isCompleted = false;
+  }
+
+  // This helper function explicitly dispatches events that are
+  // automatically dispatched by the platform video player in
+  // a real implementation:
+  //  https://github.com/flutter/packages/blob/260102b64c0fac9c66b7574035421fa6c09f5f89/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java#L189
+  void pumpEvents() {
+    if (position >= kTestVideoDuration) {
+      if (!_isCompleted) {
+        _isCompleted = true;
+        _streamController.add(VideoEvent(
+          eventType: VideoEventType.completed,
+        ));
+      }
+
+      if (isPlaying) {
+        _stopwatch?.stop();
+        _streamController.add(VideoEvent(
+          eventType: VideoEventType.isPlayingStateUpdate,
+          isPlaying: false,
+        ));
+      }
+    }
   }
 
   @override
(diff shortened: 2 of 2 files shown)


gnprice:
Let's start using the ["buddy review" system](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/GitHub.20workflows/near/1800080) here, now that the start of the official GSoC period is imminent and it seems like everyone's actively engaged again.

@sm-sayedi, would you do the first review on this PR?

As a reminder since we're just starting to use this system: when you get a buddy review request, [please prioritize it](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/GitHub.20workflows/near/1800082) right after important bugs and your regressions, ahead of most of your own PRs. (In particular I don't think anyone is currently working on an issue that would come ahead of doing buddy reviews.)

sm-sayedi:
Seems pretty decent to me. The tests are written cleverly. Good job! 

Just one small comment below.

gnprice:
@sm-sayedi Thanks for that review!

@kenclary I think this is ready for [your review](https://zulip.readthedocs.io/en/latest/outreach/mentoring.html#supporting-your-mentee) as @rajveermalviya's GSoC mentor, now that it's had a first round of buddy review and there's only one small comment outstanding. Please take a look!

Then @sm-sayedi after the thread above is resolved and buddy review is complete, please remove the "buddy review" label and add "maintainer review", and request a review from @chrisbobbe.

test/widgets/lightbox_test.dart
```diff
@@ -127,20 +228,24 @@ void main() {
...
     testWidgets('shows a VideoPlayer, and video is playing', (tester) async {
-      await setupPage(tester, videoSrc: Uri.parse('https://a/b.mp4'));
+      await setupPage(tester, videoSrc: Uri.parse(FakeVideoPlayerPlatform.kTestVideoUrl));
 
       check(FakeVideoPlayerPlatform.initialized).isTrue();
       check(FakeVideoPlayerPlatform.isPlaying).isTrue();
 
       await tester.ensureVisible(find.byType(VideoPlayer));
+
+      FakeVideoPlayerPlatform.cancelTimer();
```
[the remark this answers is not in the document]
rajveermalviya: Yeah generally one would think that tearDown callback should run before the timer leak checks, but unfortunately that isn't the case, which is understandable - see https://github.com/flutter/flutter/issues/24166

So, in this test case, the timer is created when the page is initialized and generally should cancel out when page is disposed, but since we don't have that setup (for simplicity of the test) the `dispose` isn't called on the VideoPlayerController, leading to a "timer leak".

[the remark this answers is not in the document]
rajveermalviya: Anyway, I think pushing the lightbox page in a route and pop-ing it before the end of each test (just like how `cancelTimer` is called currently), would avoid this "hacky" looking `cancelTimer` function, so I'll change the tests behavior to that.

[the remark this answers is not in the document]
rajveermalviya: (Also would like to get to know a better way of doing this, if I am missing something here)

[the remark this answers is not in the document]
rajveermalviya: So, turns out dispose is being called, and my tentative workaround (of temporary page route) was completely unnecessary, good to know.

Also after debugging the tests, the timer leak checking actually is triggered in the dispose call, it happens [exactly here](https://github.com/flutter/packages/blob/7b423f58218025f770efac80485802c31039e4a6/packages/video_player/video_player/lib/video_player.dart#L513) which is surprising and seems like a magic behaviour (since it's implementation is internal to dart vm).

Anyway after leading to nowhere, I replaced the timer implementation to instead use stopwatch. And everything seems to be correct now.

rajveermalviya:
@sm-sayedi Thanks for the review, I pushed a new revision which as mentioned in https://github.com/zulip/zulip-flutter/pull/694#discussion_r1613450430 uses `Stopwatch` to track play/pause time instead of a `Timer` which entirely removes the concern of manually cancelling the timers in each test.

test/widgets/lightbox_test.dart
```diff
@@ -16,11 +19,30 @@ import '../model/binding.dart';
...
+  static bool _hasError = false;
+  static bool _isPlaying = false;
+  static Duration _position = Duration.zero;
+  static Duration _lastSetPosition = Duration.zero;
+  static Stopwatch? _stopwatch;
+
+  static List<String> get callLogs => _callLogs;
+  static bool get initialized => _initialized;
+  static bool get hasError => _hasError;
+  static bool get isPlaying => _isPlaying;
```
sm-sayedi, line 40:
```suggestion
  static bool get isPlaying => _stopwatch?.isRunning ?? false;
```

sm-sayedi (commented):
Thanks, @rajveermalviya for the review! LGTM.

Just a small change below.

rajveermalviya:
Thanks for the review @sm-sayedi, new revision pushed!

sm-sayedi (commented):
Thanks @rajveermalviya ! I am satisfied with this revision for the buddy review. Let's move on to the maintainer review with @chrisbobbe .

test/widgets/lightbox_test.dart
```diff
@@ -109,6 +189,16 @@ void main() {
...
 
   group("VideoLightboxPage", () {
+    void verifySliderPosition(WidgetTester tester, Duration duration) {
+      final slider = tester.widget(find.byType(Slider)) as Slider;
+      check(slider.value)
+        .equals(duration.inMilliseconds.toDouble());
+      final positionLabel = tester.widget(
+        find.byKey(VideoPositionSliderControl.kCurrentPositionLabelKey)) as VideoDurationLabel;
+      check(positionLabel.duration)
+        .equals(duration);
```
chrisbobbe, line 199: This checks that the `VideoDurationLabel` widget gets the intended `Duration` value, but we don't yet check that `VideoDurationLabel` correctly renders that `Duration` into the `hh:mm:ss` format. That would be good to do, either here or in some separate unit tests.

chrisbobbe (commented):
Thanks @rajveermalviya! Small comments below.

rajveermalviya:
Thanks for the review @chrisbobbe, pushed a new revision.

chrisbobbe:
Thanks, LGTM! Labeling for Greg's review.

test/widgets/lightbox_test.dart
```diff
@@ -152,5 +267,148 @@ void main() {
...
+      final gesture = await tester.startGesture(trackStartPos);
+      FakeVideoPlayerPlatform.pumpEvents();
+      await tester.pump();
+      checkPosition(tester, Duration.zero);
+      check(FakeVideoPlayerPlatform.position).equals(Duration.zero);
+
+      await gesture.moveBy(Offset(twentyPercent, 0.0));
+      FakeVideoPlayerPlatform.pumpEvents();
+      await tester.pump();
+      checkPosition(tester, FakeVideoPlayerPlatform.kTestVideoDuration * 0.2);
```
gnprice, line 338: nit:
```suggestion
      checkPosition(tester, 0.2 * FakeVideoPlayerPlatform.kTestVideoDuration);
```

That's the standard convention for mathematicians and physicists — one would write $0.2 \cdot T$ or $0.2 \\, T$, and rarely to never $T \cdot 0.2$ or $T \\, 0.2$.

Here, it's helpful also just by foregrounding the part that's more interesting, so it isn't inconspicuously off at the end of a longish line.

test/widgets/lightbox_test.dart
```diff
@@ -152,5 +267,148 @@ void main() {
...
+      check(FakeVideoPlayerPlatform.position).equals(Duration.zero);
+
+      await gesture.moveBy(Offset(twentyPercent, 0.0));
+      FakeVideoPlayerPlatform.pumpEvents();
+      await tester.pump();
+      checkPosition(tester, FakeVideoPlayerPlatform.kTestVideoDuration * 0.4);
+      check(FakeVideoPlayerPlatform.position).equals(Duration.zero);
+
+      await gesture.moveBy(Offset(twentyPercent, 0.0));
```
gnprice, line 347: This test can be made a bit stronger by dragging the slider backward, as well as forward, before finally letting go. One could imagine bugs where either switching directions, or moving backward, triggers some different and wrong behavior.

(I'm getting this idea from just rereading my list at https://github.com/zulip/zulip-flutter/pull/587#discussion_r1597195082 .)

test/widgets/lightbox_test.dart
```diff
@@ -152,5 +267,148 @@ void main() {
...
+      // Verify that after dragging ends, video position is at the
+      // halfway point, and after that it starts advancing as the time
+      // passes.
+      check(FakeVideoPlayerPlatform.position).equals(halfTime);
+
+      const waitTime = Duration(seconds: 1);
+      await tester.pump(waitTime);
+      FakeVideoPlayerPlatform.pumpEvents();
+      checkPosition(tester, halfTime + (waitTime * 1));
+      check(FakeVideoPlayerPlatform.position).equals(halfTime + (waitTime * 1));
```
gnprice, line 401: Would this be effective as a regression test for the potential bug it's meant to catch?

I.e., quoting my [original request](https://github.com/zulip/zulip-flutter/pull/587#discussion_r1597195082):
> a regression test for the bug that the `_isSliderDragging` and `_sliderValue` logic is avoiding as explained in your comment quoted at https://github.com/zulip/zulip-flutter/pull/587#discussion_r1596190776 .

The regression test I suggested there was:
> Drag slider from one point to another. Check the slider appears at the place you dragged it to now, and at every frame until it starts advancing again because the video is playing.

Where "every frame" is key, because the bug you describe in that comment is a single-frame glitch. It seems like skipping over a whole second could bypass such a glitch.

gnprice (commented):
Thanks @rajveermalviya, and thanks @sm-sayedi and @chrisbobbe for the helpful previous reviews! Comments below.

test/widgets/lightbox_test.dart
```diff
@@ -152,5 +267,148 @@ void main() {
...
+    testWidgets('video advances overtime & stops playing when it ends', (tester) async {
+      await setupPage(tester, videoSrc: Uri.parse(FakeVideoPlayerPlatform.kTestVideoUrl));
+      check(FakeVideoPlayerPlatform.isPlaying).isTrue();
+      check(FakeVideoPlayerPlatform.position).equals(Duration.zero);
+
+      final halfTime = FakeVideoPlayerPlatform.kTestVideoDuration * 0.5;
+
+      await tester.pump(halfTime);
+      FakeVideoPlayerPlatform.pumpEvents();
+      checkPosition(tester, halfTime);
```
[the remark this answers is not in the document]
rajveermalviya: In a real app, these events are sent from platform bindings, for example [here's android](https://github.com/flutter/packages/blob/260102b64c0fac9c66b7574035421fa6c09f5f89/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java#L189-L196).

The initial revisions of this PR used `Timer.periodic` instead of `Stopwatch`, and in that implementation these events were dispatched from the `Timer.periodic` callback, but with `Stopwatch` there's no such hook thus the need for an explicit `pumpEvents` function.

(In this revision; Removed the usage from tests which don't rely on it.)

test/widgets/lightbox_test.dart
```diff
@@ -152,5 +267,148 @@ void main() {
...
+      check(FakeVideoPlayerPlatform.position)
+        .equals(FakeVideoPlayerPlatform.kTestVideoDuration - const Duration(milliseconds: 500));
+      check(FakeVideoPlayerPlatform.isPlaying).isTrue();
+
+      // At exactly the end of the video.
+      await tester.pump(const Duration(milliseconds: 500));
+      FakeVideoPlayerPlatform.pumpEvents();
+      checkPosition(tester, FakeVideoPlayerPlatform.kTestVideoDuration);
+      check(FakeVideoPlayerPlatform.position).equals(FakeVideoPlayerPlatform.kTestVideoDuration);
+      check(FakeVideoPlayerPlatform.isPlaying).isFalse(); // stopped playing
```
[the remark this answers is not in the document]
rajveermalviya: replied in https://github.com/zulip/zulip-flutter/pull/694#discussion_r1635482992

test/widgets/lightbox_test.dart
rajveermalviya: Unfortunately, I was unable to get this working with operator overloading extensions, local operator overloads on `num`/`double` won't take precedence over the upstream ones.
https://github.com/dart-lang/language/issues/966

gnprice: Oh I see, I forgot that this is a Duration and not a number. So be it, then.

rajveermalviya: In any case, for now I've reduced the character count:
```diff
-       checkPosition(tester, FakeVideoPlayerPlatform.kTestVideoDuration * 0.4);
+       checkPosition(tester, kTestVideoDuration * 0.4);
```

test/widgets/lightbox_test.dart
rajveermalviya: While debugging this, I found that it was possible to [correctly mitigate](https://github.com/zulip/zulip-flutter/pull/694/commits/819118f0b57afc024a7de3ad1da641b6d90be0b0) flickering of slider rather than the previous workaround of delaying the redraw when listeners were called.

The first commit applies the correct fix: https://github.com/zulip/zulip-flutter/pull/694/commits/819118f0b57afc024a7de3ad1da641b6d90be0b0

rajveermalviya: Also, making the wait interval lower than 500ms wasn't working because `video_player` plugin runs a `Timer.periodic` with 500ms interval for dispatching position updates via the `ValueNotifier`:
https://github.com/flutter/packages/blob/260102b64c0fac9c66b7574035421fa6c09f5f89/packages/video_player/video_player/lib/video_player.dart#L566

rajveermalviya:
Thanks for the review @gnprice, pushed a new revision, PTAL :)

test/widgets/lightbox_test.dart
gnprice: > `video_player` plugin runs a `Timer.periodic` with 500ms interval for dispatching position updates

Huh indeed — wacky. That'll work fine when the video is like 10 minutes long so that 500ms is no more than a couple of pixels, but seems annoying for videos that are like 5 seconds long.

And yeah, that behavior is very conspicuous when I look for it. E.g. here on this 12-second video:
https://chat.zulip.org/#narrow/stream/9-issues/topic/Spoiler.20collapse.20.2F.20uncollapse.20is.20unreliable/near/1823188

but really even on this 3-minute video it's noticeable:
https://chat.zulip.org/#narrow/stream/7-test-here/topic/video.20test/near/1776292

Certainly out of scope for this PR, though.

lib/widgets/lightbox.dart
```diff
@@ -325,20 +329,20 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
...
+          onChangeEnd: (value) async {
             final durationValue = Duration(milliseconds: value.toInt());
-            setState(() {
-              _sliderValue = durationValue;
-            });
-            widget.controller.seekTo(durationValue);
-
-            // The toggling back of '_isSliderDragging' is omitted here intentionally,
-            // see '_handleVideoControllerUpdates'.
+            await widget.controller.seekTo(durationValue);
```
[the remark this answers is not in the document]
rajveermalviya: It refers to `seekTo`'s implementation which [calls into platforms bindings](https://github.com/flutter/packages/blob/7805455bf62787d948593fc780e0313fc8fb2bec/packages/video_player/video_player/lib/video_player.dart#L637), waits for it's Future to complete and then calls `_updatePosition` which dispatches position update via [`ValueNotifier.value` setter](https://github.com/flutter/packages/blob/7805455bf62787d948593fc780e0313fc8fb2bec/packages/video_player/video_player/lib/video_player.dart#L745-L757) – resulting in our `_handleVideoControllerUpdate` to be called.

rajveermalviya:

gnprice (commented):
OK, and here's the rest of a round of review!

I haven't yet read your new revision of a couple of hours ago ­— I was pulled AFK after writing most of this and just sat back down a few minutes ago. But I think this is mostly on different parts of the code than I commented on above.

test/widgets/lightbox_test.dart
```diff
@@ -1,77 +1,148 @@
...
+  bool get isPlaying => _stopwatch?.isRunning ?? false;
+
+  Duration get position {
+    assert(_stopwatch != null);
+    final pos = _stopwatch!.elapsed + _lastSetPosition;
+    if (pos.compareTo(kTestVideoDuration) >= 0) {
+      return kTestVideoDuration;
+    } else {
+      return pos;
+    }
```
gnprice, line 58:
```suggestion
    return pos <= kTestVideoDuration ? pos : kTestVideoDuration;
```

Or is there a subtle difference between that and `compareTo` which I'm missing?

test/widgets/lightbox_test.dart
```diff
@@ -1,77 +1,148 @@
...
     _streamController = StreamController<VideoEvent>();
-    initialized = false;
-    isPlaying = false;
+    _hasError = false;
+    _lastSetPosition = Duration.zero;
+    _stopwatch?.stop();
+    _stopwatch?.reset();
+    _callLog.clear();
+    _initialized = false;
+    _isCompleted = false;
```
gnprice, line 70: I see putting these in order was helpful :slightly_smiling_face: (as the `_hasError` and `_isCompleted` lines are new in this revision)

lib/widgets/lightbox.dart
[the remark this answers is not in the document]
gnprice: OK, so the ValueNotifier this is referring to is the controller? In that case I think a helpful rewording would be to replace "invoking the `ValueNotifier.value` setter" with "setting the controller's `value`".

gnprice (commented):
Thanks for the revision! Comments below on the changes there. Then I think my previous round above (https://github.com/zulip/zulip-flutter/pull/694#pullrequestreview-2114236513), on other parts of the code, is still open for you to act on.

rajveermalviya:
Thanks for the review @gnprice, pushed a new revision!

gnprice (commented):
Thanks @rajveermalviya for all your work on this! Merging now.

There were a couple of remaining items where I figured it'd be better at this point to demonstrate what I have in mind rather than try to explain it. So I've added a few commits on top:
1e2143c5b lightbox test [nfc]: Tighten up checkPositions doc
97109a002 lightbox test [nfc]: Tighten test cases using a checkPositionsRelative
b9065e863 lightbox test [nfc]: Extract findSliderPosition
ff029afb8 lightbox test: Revise no-flicker-upon-drag test, and add explanatory links

The second commit completes the changes I was going for in https://github.com/zulip/zulip-flutter/pull/694#discussion_r1637307665 ; the version in your latest revision is helpful, but doesn't get all of the benefits of concision that I had in mind.

The last commit follows up on https://github.com/zulip/zulip-flutter/pull/694#discussion_r1630430553, and the next-to-last is just a small refactor to enable that one.

The first commit is a stylistic revision to the handy new doc comment on `checkPositions`. For background on the changes, see:
https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph
https://dart.dev/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

In general all the advice on that page (about documentation style) is worth reading, as is the corresponding section of the Flutter style guide:
https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#documentation-dartdocs-javadocs-etc

Issue #239, opened 2023-07-24
Register our tests in Flutter's "customer tests" suite


gnprice:
Flutter upstream has a scheme for Flutter users' test suites to be integrated as part of Flutter's own CI, in the "customer tests" suite:
https://github.com/flutter/tests

This has the effect that a Flutter change which breaks those tests is treated as a "breaking change", the same category as a change which breaks Google's own internal tests:
https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
That in turn sets a higher bar for being worth the breakage and calls for more effort from the upstream author of the change to help developers migrate across it.

We should register for this scheme, now that we're out of the experimental phase and are committed to Flutter. The main prerequisite is that we'll need to have some appropriate CI scripts in the first place:
* #60 

(The two major barriers that I suspect are the reasons most Flutter developers don't participate are (a) if your code isn't open-source, then it gets logistically complicated to do so, and (b) it necessarily requires you to be keeping up with Flutter `main`/`master`, rather than only upgrading on stable releases. But we're open-source and are already keeping up with Flutter main, which we value because it lets us make upstream changes and promptly benefit from them.)

I believe there are two occasions so far where this would have flagged an upstream change as breaking:
* https://github.com/flutter/flutter/pull/126108 (see our commit https://github.com/zulip/zulip-flutter/commit/2f0f4692f2d55748d055ac7c140ab7414307a36d )
* https://github.com/flutter/flutter/pull/130536 (see [chat thread about breakage](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20Failing.20tests/near/1615254))

gnprice:
This is done! The following PR in flutter/tests is now merged:
* https://github.com/flutter/tests/pull/368

PR #696, opened 2024-05-21
deps: Unpin intl, to follow Flutter's pinning; trim comments

gnprice:
Unpinning `package:intl` is required for adding our tests to Flutter's "customer tests", #239:
https://github.com/flutter/tests/commit/a047b7a5e6184969d943f88887bbe32658ed99c1
See comment added in the last commit here for background.

Also cut some unnecessary (or no-longer-necessary) verbose comments that we've had in this file, mostly left by the `flutter create` template.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -26,6 +26,17 @@ dependencies:
   flutter_localizations:
     sdk: flutter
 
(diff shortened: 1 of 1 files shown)


chrisbobbe:
LGTM! Merging.

Issue #697, opened 2024-05-22
content: Markers on list items are smaller than list-item text


chrisbobbe:
In web, the markers "1.", "2.", and "3." in

1. one
2. two
3. three

are the same size as the list-item text ("one", "two", "three"):

<img width="136" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/1bedbc66-13b7-4eb4-805a-0635f6734abc">

Not so in zulip-flutter, though:

<img width=400 src="https://github.com/zulip/zulip-flutter/assets/22248748/1b56bcb0-86b8-430f-81e6-56c4d50b9d3c">

Issue #641, opened 2024-05-01
Patch up spurious vertical space after quoted code blocks


gnprice:
If you send a Zulip message with a code block inside a quotation, the code block gets followed by an extra blank line, for no good reason.

The root issue here is actually a server bug, ultimately due to the Zulip server's use of the Python-Markdown library. Its current symptom in zulip-flutter is the same symptom that Zulip web has shown for many years. But it's likely Zulip web will start patching this up at some point so that the bug no longer has user-facing symptoms, and we should do the same in zulip-flutter. (Even in a future where the server bug itself is someday fixed, it'll still likely apply to all existing messages sent before that point, so it'll remain useful for clients to patch up the issue.)

Here's a [sample message](https://chat.zulip.org/#narrow/stream/3-backend/topic/extra.20markup.20on.20quoted.20fences/near/1784081), and screenshots of how it appears:
| web | flutter (Linux) |
| --- | --- |
| ![image](https://github.com/zulip/zulip-flutter/assets/28173/d07eb322-c461-4c1e-955f-64bb2cadb5f3) | ![image](https://github.com/zulip/zulip-flutter/assets/28173/4a2d4567-dd4d-47f7-90ce-cef6ca1f1509) |

Note the quotation marker at the left extending well below the code block, by about the height of a line.

The HTML looks like this:
```html
<blockquote>
<p><div class="codehilite"><pre>[…]</pre></div><br>
</p>
</blockquote>
```

Note the surrounding `<p>` element, and the `<br>`. Those wouldn't normally be there.

Thanks to @karlstolley and @andersk for [diagnosing](https://chat.zulip.org/#narrow/stream/3-backend/topic/extra.20markup.20on.20quoted.20fences/near/1784077) this issue.

An issue that appears related at first sight but I think ultimately isn't (except that they're both about quirky HTML emitted for code blocks) is:
* #355

PR #698, opened 2024-05-22
content: Consolidate message-content base style and apply it more accurately

chrisbobbe:
I have four main goals for this PR:

- Stop applying the Zulip-message text color to things that aren't Zulip-message text (and refactor to make these errors easy to avoid)
- Don't regress on performance in the message list (for example by making a `DefaultTextStyle` for every message)
- Define and use a consolidated "base message text style" for content widgets to inherit from (which I do by lumping together the text color and the `Paragraph` styles)
- Add a `ThemeExtension` to our app's `ThemeData`, as a home for certain content styles (that we don't want to recompute for each message, or that differ between light and dark theme and will need to be `lerp`ed for the fade animation)

Along the way, I make some small changes/fixes:

- Message sender name was `hsl(0deg 0% 15%)`; it now matches web at `hsl(0deg 0% 20%)`
- #697 (which I filed just now)
- `<br>` in a block context, a.k.a. `LineBreakNode`, has the proper/expected font size and line height, instead of values from a Material default. (Inconspicuous and not worth filing an issue for; I don't know how to make one of these except with help from the server bug that causes #641.)
- Code blocks, and our basic math-block implementation, now have 1.4x line height instead of 1.43x. (This is to match web instead of a Material default, and is inconspicuous.)

(edit: Scroll down for up-to-date screenshots.)

Fixes: #697
Related: #95

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -39,7 +39,6 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
       color: const HSLColor.fromAHSL(1, 0, 0, 0.15).toColor(),
       fontSize: kBaseFontSize,
       letterSpacing: 0,
-      textBaseline: TextBaseline.alphabetic,
       height: (22 / kBaseFontSize),
       leadingDistribution: TextLeadingDistribution.even,
       decoration: TextDecoration.none,
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -34,6 +34,8 @@ import 'text.dart';
 class ContentTheme extends ThemeExtension<ContentTheme> {
   ContentTheme() :
     textStylePlainParagraph = TextStyle(
+      debugLabel: 'ContentTheme.textStylePlainParagraph',
+
       color: const HSLColor.fromAHSL(1, 0, 0, 0.15).toColor(),
       fontSize: kBaseFontSize,
       height: (22 / kBaseFontSize),
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -32,14 +32,22 @@ import 'text.dart';
 /// background. For what this is in the message list, see
 /// widgets/message_list.dart.
 class ContentTheme extends ThemeExtension<ContentTheme> {
-  ContentTheme() :
+  ContentTheme(BuildContext context) :
     textStylePlainParagraph = TextStyle(
-      debugLabel: 'ContentTheme.textStylePlainParagraph',
+      inherit: false,
 
       color: const HSLColor.fromAHSL(1, 0, 0, 0.15).toColor(),
       fontSize: kBaseFontSize,
+      letterSpacing: 0,
+      textBaseline: TextBaseline.alphabetic,
       height: (22 / kBaseFontSize),
-    );
+      leadingDistribution: TextLeadingDistribution.even,
+      decoration: TextDecoration.none,
+      fontFamily: kDefaultFontFamily,
+      fontFamilyFallback: defaultFontFamilyFallback,
+    )
+      .merge(weightVariableTextStyle(context))
+      .copyWith(debugLabel: 'ContentTheme.textStylePlainParagraph');
 
   ContentTheme._({
     required this.textStylePlainParagraph,
@@ -55,9 +63,12 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
     return extension!;
   }
 
-  /// The [TextStyle] we use for plain, unstyled paragraphs.
+  /// The complete [TextStyle] we use for plain, unstyled paragraphs.
   ///
   /// Also the base style that all other text content should inherit from.
+  ///
+  /// This is the complete style for plain paragraphs. Plain-paragraph content
+  /// should not need styles from other sources, such as Material defaults.
   final TextStyle textStylePlainParagraph;
 
   @override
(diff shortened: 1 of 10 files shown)


gnprice:
Let's skip the refactor that passes `ambientTextStyle` everywhere:
ec772121a content [nfc]: Apply default color by passing down ambientTextStyle

because it seems like it makes a lot of the code worse by having to pass around a new parameter that mostly just goes through untouched. Doing that through a widget subtree is exactly the sort of thing that InheritedWidget is designed to replace. And I'm dubious this way is necessarily any faster — it's doing a little more work in all those widget constructors within each message, in place of doing a little more work at the root widget of the message.

I think having one `DefaultTextStyle` at the top of each `MessageContent` is fine. Widgets are cheap, as long as they're simple widgets.

Avoiding `DefaultTextStyle.merge`, in favor of giving some already-computed style directly, will be good for explicitness as well as for potentially a small performance gain. `ContentTheme` looks like a good strategy for doing that.

chrisbobbe:
Sure, makes sense. Revision pushed.

lib/widgets/content.dart
```diff
@@ -36,7 +105,9 @@ class MessageContent extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return InheritedMessage(message: message,
-      child: BlockContentList(nodes: content.nodes));
+      child: DefaultTextStyle(
```
gnprice, line 108: The content elements that had been getting these attributes from
    TextTheme.bodyMedium (
    
      fontSize: 14.0
      height: 1.43
    
    ), and are now getting them from here (
    
      fontSize: 17
      height: 22 / 17 which is ~1.29
    
    ), are:
    
    - line breaks (for `LineBreakNode`),
    - code blocks (and our simplified math blocks), and
    - list-item markers.
    
    So, those all look different after this change.

In your screenshots, I see the difference in the list-item markers. I don't see a difference in the code blocks; did their font size end up changing, or was that change suppressed by something else?

lib/widgets/message_list.dart
```diff
@@ -915,10 +913,11 @@ class MessageWithPossibleSender extends StatelessWidget {
                   const SizedBox(width: 8),
                   Flexible(
                     child: Text(message.senderFullName, // TODO get from user data
-                      style: const TextStyle(
+                      style: TextStyle(
                         fontFamily: 'Source Sans 3',
                         fontSize: 18,
                         height: (22 / 18),
+                        color: const HSLColor.fromAHSL(1, 0, 0, 0.2).toColor(),
```
gnprice, line 920: msglist: Set message-sender-name color to match web

    …
    That color:
      const HSLColor.fromAHSL(1, 0, 0, 0.15).toColor()
    doesn't match web, it turns out. (Or the old Figma, which uses
    true black.)

It'd be good to have this change in screenshots, too.

lib/widgets/content.dart
```diff
@@ -20,6 +20,75 @@ import 'message_list.dart';
...
+    }
+    return ContentTheme._(
+      textStylePlainParagraph: TextStyle.lerp(textStylePlainParagraph, other?.textStylePlainParagraph, t)!,
+    );
+  }
+
+  /// The [ContentTheme] from the context's active theme.
+  ///
+  /// The [ThemeData] must include [ContentTheme] in [ThemeData.extensions].
+  static ContentTheme of(BuildContext context) {
```
gnprice, line 84: nit: let's follow the Flutter conventional order for class members:

lib/widgets/content.dart
```diff
@@ -20,6 +20,75 @@ import 'message_list.dart';
...
+/// Styles belong here if they differ between light and dark themes,
+/// so that they will animate smoothly (with help from [lerp]) on theme changes.
+/// Styles also belong here if we want to centralize computing them,
+/// for performance. (The message list is particularly performance-sensitive.)
+///
+/// Content elements are assumed to be painted on a theme-appropriate
+/// background. For what this is in the message list, see
+/// widgets/message_list.dart.
+class ContentTheme extends ThemeExtension<ContentTheme> {
+  /// The complete [TextStyle] we use for plain, unstyled paragraphs.
```
gnprice, line 34: - [Typography.blackCupertino] or [TextTheme.blackMountainView]
      (depending on platform)

(Typography.blackMountainView?)

gnprice (commented):
Thanks for the revision! Small comments; otherwise LGTM.

lib/widgets/content.dart
chrisbobbe: Oh, oops: yeah, code blocks set their own font size (0.825 * `kBaseFontSize`), so that doesn't change in this commit. Code blocks don't set their line height, so that does change, from 1.43x to ~1.29x. But then a subsequent commit sets it to 1.4x, which is very close to the original 1.43x.

chrisbobbe:
Thanks for the review! Revision pushed, and here's a fresh batch of screenshots from it. One interesting finding is that the difference between 1.43x and 1.4x line height (in code blocks) doesn't seem to be visible on my iPhone. Maybe there's a step that rounds to the nearest physical pixel, and maybe it would be observable on a device with higher pixel density than mine.

| Before | After |
| --- | --- |
| F26A8DA6-DCA5-431E-9497-4FCB8A6DBE14 | 1ED35D30-7F0E-4ABC-B9DD-120105ECEC07 |
| B0656C6B-E5BE-422A-8404-3B2FECD17DDE | F1226D39-9362-4375-BBE4-9D333BB5456D |
| F3110C2D-B511-4F6A-86F0-1A74B5841E6E | 5E4EF34F-1FD1-49F2-A69F-184BBD51836B |
| 0D5D4234-0B91-4B1D-A7D0-A76DF441282C | 9F69079E-CDC9-4620-912C-386340D5A9DF |

gnprice:
Cool, thanks for the revision and the screenshots!

I see that sender-name-color difference, though it's awfully subtle — I had to look several times before I spotted it.

In the last screenshot, the difference in unordered list-item markers is pretty noticeable. The new, larger size definitely looks better.

All looks good; merging.

PR #701, opened 2024-05-22
Make handleEvent an async method

gnprice:
With #458, we'll want this method to write changes to the database
on certain events, to update the metadata we keep about the user's
Zulip servers.

When it does that, we'll want it to wait for that write to finish
before it moves on to make other changes, and for its caller
[UpdateMachine.poll] to similarly wait.  That saves us from having
to worry about potential races if some later event also wants to
make database changes while the first changes haven't finished yet.
It also gives us a good foundation for generally not showing the
user a confirmation that some change has been made until we've
actually saved the change.

So to do that, we'll need the method to be async, and to return a
Future which its call sites should wait for.

This PR does that, after a series of commits donig the same for its
callers, recursively.

### DIFF
--- a/integration_test/unreadmarker_test.dart
+++ b/integration_test/unreadmarker_test.dart
@@ -51,11 +51,11 @@ void main() {
 
     final messages = await setupMessageListPage(tester, 500);
     await binding.traceAction(() async {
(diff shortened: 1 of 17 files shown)


chrisbobbe:
Interesting; yeah, I guess the need for this makes sense. LGTM, merging.

Issue #572, opened 2024-03-19
notif: Use `#` with stream names


gnprice:
See [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications.20differences/near/1757473).

Issue #571, opened 2024-03-19
notif: Show time in collapsed notification


gnprice:
See [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications.20differences/near/1757473).

To implement this, I believe what's needed is to make a "summary notification" and put the information there. See the notifications code in zulip-mobile (in Kotlin, under `android/`) for comparison.

We already show the time on individual notifications — just not on the collapsed notification.

This might get done at the same time as the following related issue:
* #569

Issue #569, opened 2024-03-19
notif: Show organization URL in collapsed notification


gnprice:
See [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications.20differences/near/1757473).

To implement this, I believe what's needed is to make a "summary notification" and put the information there. See the notifications code in zulip-mobile (in Kotlin, under `android/`) for comparison.

Better yet would be the Zulip organization's name, rather than URL. That's tracked as a followup issue:
* #570

Issue #341, opened 2023-10-26
Remove notifications when messages are read, on Android


gnprice:
We have this feature in zulip-mobile, and should match it here.

We should also eventually add this feature on iOS. We don't yet have it there in zulip-mobile (https://github.com/zulip/zulip-mobile/issues/3119), so that will be tracked separately and happen later.

gnprice:
Mentioned in feedback here:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/notifications.20from.20cold.20start/near/1749654

(Which was the first day after Android notifications started working generally.)

SharmaDhruv2511:
I would like to contribute to it @gnprice

SharmaDhruv2511:
I would like to contribute to it @gnprice

gnprice:
@SharmaDhruv2511 Please see our guide here: https://github.com/zulip/zulip-flutter#picking-an-issue-to-work-on

gnprice:
I picked this up today, and I've gotten the feature working successfully in a draft branch. There's still a fair amount of cleanup needed to turn that into something mergeable, though.

The way this feature works in zulip-mobile (which has worked quite well) is that on each notification we include in the "extras" a key `lastZulipMessageId`. Then when an `RemoveFcmMessage` comes in from the server, bearing a list of Zulip message IDs that have been read (and previously caused a notification), we look through the existing notifications in the notification UI to see which ones correspond to those Zulip message IDs and so should be removed.

To date, the way we've been interacting with the Android SDK for controlling notifications is with `package:flutter_local_notifications`, which has a Flutter plugin for this purpose. As discussed at #351, though, that package isn't entirely satisfactory. And implementing this feature turns up another case of that: the plugin provides a version of the `getActiveNotifications` method which tells us about the existing notifications in the notification UI… but it doesn't give us a way to see the extras on the notifications. So our choices are:
* Find some other way of implementing the feature which works with the existing `package:flutter_local_notifications` API. There may not exist a clean way to do this, though we can likely find some kind of hack.
* Patch `package:flutter_local_notifications` to provide those extras. Do so locally at first, and try to upstream it.
* #351.

I decided this was a good prompt for trying out Pigeon and doing #351. It worked. Pigeon has some rough edges, but it definitely feels a lot more pleasant than trying to write by hand all the serialization, deserialization, and dispatch logic for talking between Dart and Kotlin.

[...]

PR #703, opened 2024-05-23
notif: Create summary notification

rajveermalviya:
https://github.com/zulip/zulip-flutter/assets/36819268/95429096-c9c4-46ee-907b-d5cc424ea0ea

Fixes: #572
Fixes: #569
Fixes: #571

### DIFF
--- a/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
+++ b/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
@@ -23,15 +23,20 @@ private class AndroidNotificationHost(val context: Context)
     override fun notify(
         tag: String?,
         id: Long,
+        autoCancel: Boolean?,
         channelId: String,
         color: Long?,
         contentIntent: PendingIntent?,
         contentText: String?,
         contentTitle: String?,
         extras: Map<String?, String?>?,
+        groupKey: String?,
+        inboxStyle: InboxStyle?,
+        isGroupSummary: Boolean?,
         smallIconResourceName: String?
     ) {
         val notification = NotificationCompat.Builder(context, channelId).apply {
+            autoCancel?.let { setAutoCancel(it) }
             color?.let { setColor(it.toInt()) }
             contentIntent?.let { setContentIntent(
                 android.app.PendingIntent.getActivity(context,
@@ -49,6 +54,12 @@ private class AndroidNotificationHost(val context: Context)
             contentTitle?.let { setContentTitle(it) }
             extras?.let { setExtras(
                 Bundle().apply { it.forEach { (k, v) -> putString(k, v) } } ) }
+            groupKey?.let { setGroup(it) }
+            inboxStyle?.let { setStyle(
+                NotificationCompat.InboxStyle()
+                    .setSummaryText(it.summaryText)
+            ) }
+            isGroupSummary?.let { setGroupSummary(it) }
             smallIconResourceName?.let { setSmallIcon(context.resources.getIdentifier(
                 it, "drawable", context.packageName)) }
         }.build()
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -142,6 +142,7 @@ class NotificationDisplayManager {
         // TODO this doesn't set the Intent flags we set in zulip-mobile; is that OK?
         //   (This is a legacy of `flutter_local_notifications`.)
         ),
+      autoCancel: true,
     );
 
     await ZulipBinding.instance.androidNotificationHost.notify(
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -94,9 +94,9 @@ class NotificationDisplayManager {
     final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
     final title = switch (data.recipient) {
       FcmMessageStreamRecipient(:var streamName?, :var topic) =>
-        '$streamName > $topic',
+        '#$streamName > $topic',
       FcmMessageStreamRecipient(:var topic) =>
-        '(unknown channel) > $topic', // TODO get stream name from data
+        '#(unknown channel) > $topic', // TODO get stream name from data
       FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 =>
         zulipLocalizations.notifGroupDmConversationLabel(
           data.senderFullName, allRecipientIds.length - 2), // TODO use others' names, from data
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -89,7 +89,7 @@ class NotificationDisplayManager {
     }
   }
 
-  static void _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) {
+  static Future<void> _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) async {
     assert(debugLog('notif message content: ${data.content}'));
     final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
     final title = switch (data.recipient) {
@@ -103,13 +103,16 @@ class NotificationDisplayManager {
       FcmMessageDmRecipient() =>
         data.senderFullName,
     };
-    final conversationKey = _conversationKey(data);
-    ZulipBinding.instance.androidNotificationHost.notify(
+    final groupKey = _groupKey(data);
+    final conversationKey = _conversationKey(data, groupKey);
+
+    await ZulipBinding.instance.androidNotificationHost.notify(
       // TODO the notification ID can be constant, instead of matching requestCode
       //   (This is a legacy of `flutter_local_notifications`.)
       id: notificationIdAsHashOf(conversationKey),
       tag: conversationKey,
       channelId: NotificationChannelManager.kChannelId,
+      groupKey: groupKey,
 
       contentTitle: title,
       contentText: data.content,
@@ -140,6 +143,21 @@ class NotificationDisplayManager {
         //   (This is a legacy of `flutter_local_notifications`.)
         ),
     );
+
+    await ZulipBinding.instance.androidNotificationHost.notify(
+      id: notificationIdAsHashOf(groupKey),
+      tag: groupKey,
+      channelId: NotificationChannelManager.kChannelId,
+      groupKey: groupKey,
+      isGroupSummary: true,
+
+      color: kZulipBrandColor.value,
+      // TODO vary notification icon for debug
+      smallIconResourceName: 'zulip_notification', // This name must appear in keep.xml too: https://github.com/zulip/zulip-flutter/issues/528
+      inboxStyle: InboxStyle(
+        // TODO(#570) Show organization name, not URL
+        summaryText: data.realmUri.toString()),
+    );
   }
 
   /// A notification ID, derived as a hash of the given string key.
@@ -157,8 +175,7 @@ class NotificationDisplayManager {
       | ((bytes[3] & 0x7f) << 24);
   }
 
-  static String _conversationKey(MessageFcmMessage data) {
-    final groupKey = _groupKey(data);
+  static String _conversationKey(MessageFcmMessage data, String groupKey) {
     final conversation = switch (data.recipient) {
       FcmMessageStreamRecipient(:var streamId, :var topic) => 'stream:$streamId:$topic',
       FcmMessageDmRecipient(:var allRecipientIds) => 'dm:${allRecipientIds.join(',')}',
--- a/test/notifications/display_test.dart
+++ b/test/notifications/display_test.dart
@@ -132,7 +132,7 @@ void main() {
             ..groupKey.equals(expectedGroupKey)
             ..isGroupSummary.isNull()
             ..inboxStyle.isNull()
-            ..autoCancel.isNull()
+            ..autoCancel.equals(true)
             ..contentIntent.which((it) => it.isNotNull()
               ..requestCode.equals(expectedId)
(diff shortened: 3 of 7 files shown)


Khader-1 (commented):
Thanks! This looks good.

test/notifications/display_test.dart
```diff
@@ -166,7 +188,7 @@ void main() {
       final stream = eg.stream();
       final message = eg.streamMessage(stream: stream);
       await checkNotifications(async, messageFcmMessage(message, streamName: null),
-        expectedTitle: '(unknown stream) > ${message.subject}',
+        expectedTitle: '#(unknown stream) > ${message.subject}',
```
sumanthvrao, line 191: Do you want to extract this out to the other commit on stream renaming? I'm guessing it isn't necessary for this commit to work?

sumanthvrao (commented):
Feel free to add a small description for each of your commits. It's much easier for a review to get a gist of what's happening in each commit that way, and it's a nice to have feature of any code base.

test/notifications/display_test.dart
rajveermalviya: @sumanthvrao Sorry but I am not sure what's requested here, this change is already in a separate commit and it is the first one in the sequence. Would you want this commit to be after the main commit?

Also this change is fixing #572 which is just adding a `#` before stream names, the `streams` -> `channel` renaming is being done in https://github.com/zulip/zulip-flutter/pull/707 which includes changes for these notification strings.

sumanthvrao: Oh, apologies if I am misunderstanding here, but how are the two changes in your `summary notif commit` that add a # before stream name different from your first commit that does the same (for 2 other strings in different part of the codebase)? I was suggesting you merge the former into the first commit?

rajveermalviya: Ah, sorry for the misunderstanding, this should definitely belong in the first commit.

rajveermalviya:
Thanks for the review @sumanthvrao!
Pushed a new revision, PTAL :)

chrisbobbe:
Thanks @rajveermalviya! Could you please rebase this past #711, which upgraded Pigeon?

rajveermalviya:
@chrisbobbe rebased to main, PTAL :)

lib/notifications/display.dart
```diff
@@ -139,6 +142,22 @@ class NotificationDisplayManager {
         // TODO this doesn't set the Intent flags we set in zulip-mobile; is that OK?
         //   (This is a legacy of `flutter_local_notifications`.)
         ),
+      autoCancel: true,
```
chrisbobbe, line 145:
```
notif: Create summary notification
```

I see that zulip-mobile also does this (`setAutoCancel(true)`). Is this change coherent with the rest of what this commit is doing to create a summary notification? Or could it stand separately as its own commit? The answer will help me understand the user-facing changes that we're making, and how they're related.

chrisbobbe (commented):
Thanks! Small comments below. Greg may have more detailed feedback on the code; he has more experience with it (and also with the UX of Android notifications).

I don't see the implementation for #571 by reading the code, but it does look like that issue is fixed. I noticed that your video shows the time ("1m") in the summary notification. I guess the platform just takes care of adding the time, so we don't have to do anything special to make it appear?

test/notifications/display_test.dart
```diff
@@ -376,10 +376,18 @@ extension on Subject<AndroidNotificationHostApiNotifyCall> {
...
 
 extension on Subject<PendingIntent> {
   Subject<int> get requestCode => has((x) => x.requestCode, 'requestCode');
   Subject<String> get intentPayload => has((x) => x.intentPayload, 'intentPayload');
   Subject<int> get flags => has((x) => x.flags, 'flags');
 }
+
+extension on Subject<InboxStyle> {
+  Subject<String> get summaryText => has((x) => x.summaryText, 'summaryText');
+}
```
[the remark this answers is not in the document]
rajveermalviya: Moved their introduction to the later (main) commit: https://github.com/zulip/zulip-flutter/pull/703/commits/14d52e4f385050f5d96acc6bcafdb1bf5baeda92

lib/notifications/display.dart
```diff
@@ -139,6 +142,22 @@ class NotificationDisplayManager {
...
+      channelId: NotificationChannelManager.kChannelId,
+      tag: groupKey,
+      groupKey: groupKey,
+      isGroupSummary: true,
+      color: kZulipBrandColor.value,
+      // TODO vary notification icon for debug
+      smallIconResourceName: 'zulip_notification', // This name must appear in keep.xml too: https://github.com/zulip/zulip-flutter/issues/528
+      inboxStyle: InboxStyle(
+        // TODO(#570) Show organization name, not URL
+        summaryText: data.realmUri.toString()),
```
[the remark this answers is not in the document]
rajveermalviya: Also updated (main) commit description with the doc link, even though it was already present in the previous commit.

lib/notifications/display.dart
rajveermalviya: Moved this to a separate commit — it's not really needed for Group summary notifications, but was added just to make the UX better, and match the zulip-mobile implementation.

rajveermalviya:
Thanks for the review @chrisbobbe, addressed your comments in the latest revision.

rajveermalviya:
> I don't see the implementation for https://github.com/zulip/zulip-flutter/issues/571 by reading the code, but it does look like that issue is fixed. I noticed that your video shows the time ("1m") in the summary notification. I guess the platform just takes care of adding the time, so we don't have to do anything special to make it appear?

Yes, and the time it shows is actually when the notification was (received via FCM and) created, not the actual time of the message — these may defer if FCM delays the message delivery for various reasons. There is an API to set a custom timestamp using [`setWhen`](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setWhen(long)) & [`setShowWhen`](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setShowWhen(boolean)) but they're not used currently by zulip-mobile. On the contrary the `MessagingStyle` notifications which zulip-mobile uses, does [require providing a custom timestamp](https://developer.android.com/reference/androidx/core/app/NotificationCompat.MessagingStyle.Message).

chrisbobbe:
Cool, thanks! This revision LGTM; over to Greg for his review.

gnprice (commented):
Thanks @rajveermalviya, and thanks @sumanthvrao and @chrisbobbe for the reviews!

Generally this all looks great — just a few small comments below.

lib/notifications/display.dart
```diff
@@ -139,6 +142,22 @@ class NotificationDisplayManager {
...
+      autoCancel: true,
```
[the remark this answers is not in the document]
rajveermalviya: There's seems to be no difference in behavior with it added or not, things verified in both conditions:
- Tapping on summary header, which does nothing; unless clicked on the collapse icon — which toggles between collapsed state.
- Tapping on child notif, opens the app and routes to the correct message, also clears that specific notif from the panel/statusbar (only if child notif had `setAutoCancel(true)`).
- Keep opening (which also clears them) child notifs, and made sure summary notif automatically dismisses when there are no child notifs anymore.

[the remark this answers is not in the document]
rajveermalviya: From [Stackoverflow](https://stackoverflow.com/questions/45590860/group-summary-notification-not-canceled-when-last-containing-notification-is-can#comment107044464_53043333), it seems like the summary notification may linger if child notifs are removed programmatically and `setAutoCancel(true)` on summary notif is not set. So, it may be a problem when #341 is implemented, but this is just a speculation though, will have to test and see there's really a difference while implementing #341.

rajveermalviya:
Thanks for the review @gnprice! Pushed a new revision.

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
```diff
@@ -51,6 +55,13 @@ private class AndroidNotificationHost(val context: Context)
                 Bundle().apply { it.forEach { (k, v) -> putString(k, v) } } ) }
             smallIconResourceName?.let { setSmallIcon(context.resources.getIdentifier(
                 it, "drawable", context.packageName)) }
+            groupKey?.let { setGroup(it) }
```
gnprice, line 58: nit: match here the ordering of the parameters above (following up on https://github.com/zulip/zulip-flutter/pull/703#discussion_r1643507237 )

test/notifications/display_test.dart
```diff
@@ -111,24 +111,47 @@ void main() {
...
-        ..extras.isNull()
-        ..contentIntent.which((it) => it.isNotNull()
-          ..requestCode.equals(expectedId)
-          ..flags.equals(expectedIntentFlags)
-          ..intentPayload.equals(jsonEncode(data.toJson()))
-        );
+
+      check(testBinding.androidNotificationHost.takeNotifyCalls())
+        .deepEquals(<Condition<Object?>>[
+          (it) => it.isA<AndroidNotificationHostApiNotifyCall>()
```
gnprice, line 122: Hmm, this `isA` check seems unhelpful. It's statically a fact that each of these items is an `AndroidNotificationHostApiNotifyCall`, so we shouldn't have to repeat that.

How about the version suggested at https://github.com/zulip/zulip-flutter/pull/703#discussion_r1643500420 ? Or the version that was in the previous revision — as that comment said, either of those is fine.

lib/notifications/display.dart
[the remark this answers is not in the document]
gnprice: Interesting, thanks for the investigation.

That issue where the summary lingers reminds me of a bug we briefly had in zulip-mobile:
https://github.com/zulip/zulip-mobile/issues/5119
Looks like that didn't involve a lack of `setAutoCancel(true)`, though ­— we had that on the summary notification from the beginning (in zulip/zulip-mobile@4229325b4e3421b4c04c6939907ba19945151521).

Given that this logically shouldn't do anything (based on the reasoning in that TODO comment), and that we've also tested empirically and can't find it doing anything, let's leave it out. We can always add it back if we later discover after #341 that it's needed as a hack.

In any case, we definitely should not include the line while leaving out the TODO comment that was attached to it. That comment is useful information, which we shouldn't discard.

gnprice (commented):
Thanks for the revision! Just a couple of small comments.

test/notifications/display_test.dart
rajveermalviya: The issue is if it's `deepEquals(<Condition<AndroidNotificationHostApiNotifyCall>>[])` it [doesn't recognize](https://github.com/dart-lang/test/blob/329c6dff4dfcb178ba6c7009cf1c1b9215b317aa/pkgs/checks/lib/src/collection_equality.dart#L109) that the child is a `Condition`, instead it goes on with comparing `AndroidNotificationHostApiNotifyCall` with `Condition<AndroidNotificationHostApiNotifyCall>` which obviously fails. This is unlike `containsInOrder` which [correctly checks](https://github.com/dart-lang/test/blob/329c6dff4dfcb178ba6c7009cf1c1b9215b317aa/pkgs/checks/lib/src/extensions/iterable.dart#L102-L106) for the generic type `Condition<T>`.

So, that's why the use of `<Condition<Object?>>` here and then asserting `AndroidNotificationHostApiNotifyCall` in the test.

rajveermalviya:
Thanks @gnprice for the review! Just pushed the latest revision.

test/notifications/display_test.dart
gnprice: I see. OK, in that case let's stick with the `..length`/`..containsInOrder` version. It's one line longer, but that's worth it in order to make the types make sense, given that the `isA` here doesn't logically belong here as part of what the test is checking (because it'll statically always be true).

I did wonder if there was a subtlety I was missing :slightly_smiling_face: in why that version wasn't equivalent to `deepEquals`: https://github.com/zulip/zulip-flutter/pull/703#discussion_r1643500909 — this definitely counts as such a subtlety.

gnprice: In general if there's something that comes up when you're acting on a review comment and it seems like there's an unexpected obstacle or things go in a different direction, please flag that proactively in the thread (either in reply to the previous comment, or as a new GitHub code-review comment), whether or not you end up going ahead and implementing the new direction.

That's one of the things that helps keep code reviews moving forward smoothly, because it saves the reviewer from having to guess about why your revision of that code came out the way it did.

rajveermalviya: For sure, for this one specifically though (`isA`), I'd looked for any existing use of `deepEquals` in the repo after your initial feedback in https://github.com/zulip/zulip-flutter/pull/703#discussion_r1643500909 and had found [this](https://github.com/zulip/zulip-flutter/blob/6ecdbdc94013b77aea7898638395fa1f3f3fa698/test/widgets/app_test.dart#L34-L36). That gave me the impression that it was an OK change to introduce. :)

gnprice: Cool. Yeah, that's a place where `isA` is needed anyway, because statically it's a list of `Route<dynamic>` but we want this element to specifically be a `WidgetRoute`.

The previous comment that this really goes in a different direction from is the one about where in these checks to put the `AndroidNotificationHostApiNotifyCall` type: https://github.com/zulip/zulip-flutter/pull/703#discussion_r1643500420
So it turned out that the `deepEquals` suggestion was incompatible with that comment. But a contradiction between different review comments I make is the kind of thing I'd very much like to learn about explicitly.

gnprice:
OK, looks great — merging! Thanks @rajveermalviya for building this, and thanks again @sumanthvrao and @chrisbobbe for the reviews. Looking forward to the improved notifications :smile:

Issue #632, opened 2024-04-24
Recognize "channel" operator in narrow links


gnprice:
This is the zulip-flutter counterpart to the following zulip-mobile issue:
* https://github.com/zulip/zulip-mobile/issues/5860

We should do the same thing here.

The umbrella server/web issue is:
* https://github.com/zulip/zulip/issues/28468

Issue #631, opened 2024-04-24
Rename "stream" within codebase to "channel" 


gnprice:
This is a follow-up to:
* #630 

After renaming "stream" to "channel" in the UI, we should do the same throughout our code. The changes for this issue should be purely [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc) — user-visible changes are in #630 or other issues.

There are some places in the Zulip server API where the word "stream" appears. The API isn't immediately changing — we'll probably rename all of those eventually, but that'll be a longer migration over time — but even for those, we should still update the names within our code to use the new terminology. For an example, see `UnreadMessagesSnapshot.dms`:
```dart
  @JsonKey(name: 'pms')
  final List<UnreadDmSnapshot> dms;
```

Related issues:
* https://github.com/zulip/zulip/issues/28468

gnprice:
I've added some notes in the issue description, from discussion on #801. (Not putting them here in an issue comment, because the way GitHub handles backlinks from commits means the thread is already quite long and beginning to get hidden, so a comment may not remain visible for long.)

gnprice:
Thanks @Khader-1 for all your work renaming a large swath of our "stream" references!

In order to focus on getting this new app ready to roll out to our users as soon as we can, I'm pushing back the priority of the remainder of this issue.

For the interim, when writing new code, we'll use "channel" names whenever they don't feel like they'd clash too much with existing names that are still "stream". We might also opportunistically convert a few more names where they feel like they'll substantially improve the coherence of that interim state of things. But we'll put a pause on pushing forward to rename all the "stream" references across the codebase.

I just closed #847 and dropped one of the commits from #846, which would have renamed a couple of classes that represent outbound parts of the API (ApiNarrowStream and StreamDestination). For those I want to keep the class names in sync with what we speak to the server, and so pair those renames with starting to use the "channel" names in the API, similar to what we did with ApiNarrowDm.

Then the one open PR for this issue at the moment is #848. I'll close that since we're not going to pursue this right now; we'll return to it later.

rohitarer:
Hello, I would like to work on this

PR #704, opened 2024-05-24
internal_link: Recognize channel operator in narrow links

Khader-1:
Fixes: #632

### DIFF
--- a/test/model/internal_link_test.dart
+++ b/test/model/internal_link_test.dart
@@ -9,6 +9,13 @@ import 'package:zulip/model/store.dart';
 import '../example_data.dart' as eg;
 import 'test_store.dart';
 
+// Using Set instead of List in to avoid any duplicated test urls.
+Set<String> getUrlSyntaxVariants(String urlString) {
+  final urlWithChannelSyntax = urlString.replaceFirst('#narrow/stream', '#narrow/channel');
+  final urlWithStreamSyntax = urlString.replaceFirst('#narrow/channel', '#narrow/stream');
+  return {urlWithStreamSyntax, urlWithChannelSyntax};
+}
+
 Future<PerAccountStore> setupStore({
   required Uri realmUrl,
   List<ZulipStream>? streams,
@@ -36,12 +43,14 @@ void main() {
     assert(streams != null || users != null);
     for (final testCase in testCases) {
       final String urlString = testCase.$1;
-      final Narrow? expected = testCase.$2;
-      test(urlString, () async {
-        final store = await setupStore(realmUrl: realmUrl, streams: streams, users: users);
-        final url = store.tryResolveUrl(urlString)!;
-        check(parseInternalLink(url, store)).equals(expected);
-      });
+      for (final urlString in getUrlSyntaxVariants(urlString)) {
+        final Narrow? expected = testCase.$2;
+        test(urlString, () async {
+          final store = await setupStore(realmUrl: realmUrl, streams: streams, users: users);
+          final url = store.tryResolveUrl(urlString)!;
+          check(parseInternalLink(url, store)).equals(expected);
+        });
+      }
     }
   }
 
@@ -131,12 +140,14 @@ void main() {
       final String description = testCase.$2;
       final String urlString = testCase.$3;
       final Uri realmUrl = testCase.$4;
-      test('${expected ? 'accepts': 'rejects'} $description: $urlString', () async {
-        final store = await setupStore(realmUrl: realmUrl, streams: streams);
-        final url = store.tryResolveUrl(urlString)!;
-        final result = parseInternalLink(url, store);
-        check(result != null).equals(expected);
-      });
+      for (final urlString in getUrlSyntaxVariants(urlString)) {
+        test('${expected ? 'accepts': 'rejects'} $description: $urlString', () async {
+          final store = await setupStore(realmUrl: realmUrl, streams: streams);
+          final url = store.tryResolveUrl(urlString)!;
+          final result = parseInternalLink(url, store);
+          check(result != null).equals(expected);
+        });
+      }
     }
   });
 
@@ -165,6 +176,18 @@ void main() {
         ('/#narrow/stream/stream/topic/topic/near/1',        TopicNarrow(5, 'topic')),
         ('/#narrow/stream/stream/subject/topic/near/1',      TopicNarrow(5, 'topic')),
         ('/#narrow/stream/stream/subject/topic',             TopicNarrow(5, 'topic')),
+      ];
+      testExpectedNarrows(testCases, streams: streams);
+    });
+
+    group('Both `stream` and `channel` can be used interchangeably', () {
+      const testCases = [
+        ('/#narrow/stream/check',                         StreamNarrow(1)),
+        ('/#narrow/channel/check',                        StreamNarrow(1)),
+        ('/#narrow/stream/check/topic/test',              TopicNarrow(1, 'test')),
+        ('/#narrow/channel/check/topic/test',             TopicNarrow(1, 'test')),
+        ('/#narrow/stream/check/topic/test/near/378333',  TopicNarrow(1, 'test')),
+        ('/#narrow/channel/check/topic/test/near/378333', TopicNarrow(1, 'test')),
       ];
       testExpectedNarrows(testCases, streams: streams);
     });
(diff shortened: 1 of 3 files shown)


test/model/internal_link_test.dart
```diff
@@ -130,13 +130,16 @@ void main() {
       final bool expected = testCase.$1;
       final String description = testCase.$2;
       final String urlString = testCase.$3;
+      final String urlWithChannelSyntax = urlString.replaceFirst('#narrow/stream/', '#narrow/channel/');
```
rajveermalviya, line 133: nit: Feels like this can go unnoticed when skimming through tests, instead I'd just add those extra cases in the `testCases` list.

test/model/internal_link_test.dart
```diff
@@ -195,6 +219,8 @@ void main() {
       final testCases = [
         ('/#narrow/stream/name/topic/',           null), // missing operand
         ('/#narrow/stream/name/unknown/operand/', null), // unknown operator
+        ('/#narrow/channel/name/topic/',           null), // missing operand
```
rajveermalviya, line 222: nit:
I'd prefer a newline here to better segregate `stream` vs `channel`.

test/model/internal_link_test.dart
```diff
@@ -258,21 +301,28 @@ void main() {
...
 
     group('if stream not found, use stream ID anyway', () {
       testExpectedStreamNarrow('#narrow/stream/123-topic',         123);
+      testExpectedStreamNarrow('#narrow/channel/123-topic',         123);
     });
 
     group('on stream link with wrong name, ID wins', () {
       testExpectedStreamNarrow('#narrow/stream/1-nonsense',        1);
       testExpectedStreamNarrow('#narrow/stream/1-',                1);
+      testExpectedStreamNarrow('#narrow/channel/1-nonsense',        1);
```
rajveermalviya, line 315: nit: newline, here and other similar cases below

test/model/internal_link_test.dart
```diff
@@ -336,27 +402,36 @@ void main() {
...
       group('basic', () {
-        String mkUrlString(operand) {
-          return '#narrow/stream/${stream.streamId}-${stream.name}/topic/$operand';
+        String mkUrlString(operand, String streamSyntax) {
+          return '#narrow/$streamSyntax/${stream.streamId}-${stream.name}/topic/$operand';
         }
         final testCases = [
-          (mkUrlString('(no.20topic)'), TopicNarrow(stream.streamId, '(no topic)')),
-          (mkUrlString('lunch'),        TopicNarrow(stream.streamId, 'lunch')),
+          (mkUrlString('(no.20topic)', 'stream'), TopicNarrow(stream.streamId, '(no topic)')),
```
rajveermalviya, line 409: nit: Add whitespace to preserve formatting, here and similar cases below

```dart
          (mkUrlString('(no.20topic)', 'stream'),  TopicNarrow(stream.streamId, '(no topic)')),
          (mkUrlString('lunch',        'stream'),  TopicNarrow(stream.streamId, 'lunch')),
          (mkUrlString('(no.20topic)', 'channel'), TopicNarrow(stream.streamId, '(no topic)')),
          (mkUrlString('lunch',        'channel'), TopicNarrow(stream.streamId, 'lunch')),
```

rajveermalviya (commented):
Thanks!
Looks great, just some comments related to style nits.

Khader-1:
Thanks for the review @rajveermalviya! and Thanks for the reply @gnprice!

How about this, keeping in mind that I'm already working on #631. We can take this approach:
- Augment `testExpectedNarrows` to test both `channel` and `stream` syntax as mentioned by @gnprice.
- Duplicate other test cases that do not use `testExpectedNarrows` to handle both cases.
- Finally while working on #631 we'll reference `stream` as `channel` in test-cases that use `testExpectedNarrows`

I've already pushed a revision with the first two steps, please take a look.

test/model/internal_link_test.dart
```diff
@@ -36,12 +36,17 @@ void main() {
...
-      final Narrow? expected = testCase.$2;
-      test(urlString, () async {
-        final store = await setupStore(realmUrl: realmUrl, streams: streams, users: users);
-        final url = store.tryResolveUrl(urlString)!;
-        check(parseInternalLink(url, store)).equals(expected);
-      });
+      // This is here to avoid duplicating each test case that contains `#narrow/stream` or `#narrow/channel`
+      final urlWithChannelSyntax = urlString.replaceFirst('#narrow/stream', '#narrow/channel');
+      final urlWithStreamSyntax = urlString.replaceFirst('#narrow/channel', '#narrow/stream');
+      for (final urlString in {urlWithStreamSyntax, urlWithChannelSyntax}) {
```
rajveermalviya, line 42:
```suggestion
      for (final urlString in [urlWithStreamSyntax, urlWithChannelSyntax]) {
```

`{'hello'}` creates a `Set<String>`, if that's unintentional use `[` `]` instead, which creates `List<String>`

rajveermalviya (approved):
Thanks for the revision! Apart from one comment, it looks great :)

Now over to @kenclary for mentor review.

test/model/internal_link_test.dart
Khader-1: No it is actually intentional. Specifically, when the url being tested does not contain `#narrow/stream` nor `#narrow/channel` then both `urlWithStreamSyntax`, `urlWithChannelSyntax` would be the same and using the list would cause unnecessary duplications.

kenclary: that sounds correct. it's obscure, though. can it get a comment explaining how the object is better than an array?

Khader-1: Sure, a new revision is there now.

gnprice:
> How about this, keeping in mind that I'm already working on #631. We can take this approach:

Sure, it's OK if the followup commit I mentioned at the end of https://github.com/zulip/zulip-flutter/pull/704#issuecomment-2141171679 happens in a later PR as part of #631, instead of this PR.

I liked what the previous revision was doing to deduplicate the test cases that come from the big list of tuples, though, too. So what I had in mind was to deduplicate callers of `testExpectedNarrows` in addition to those.

For explicitly duplicated test cases, I really do mean just a handful ­— for example one boring StreamNarrow and one boring TopicNarrow, plus a TopicNarrow with `/near/`.

Khader-1:
Thanks @gnprice! Please check the new revision

test/model/internal_link_test.dart
```diff
@@ -9,6 +9,13 @@ import 'package:zulip/model/store.dart';
 import '../example_data.dart' as eg;
 import 'test_store.dart';
 
+// Using Set instead of List in to avoid any duplicated test urls.
+Set<String> getUrlSyntaxVariants(String urlString) {
+  final urlWithChannelSyntax = urlString.replaceFirst('#narrow/stream', '#narrow/channel');
+  final urlWithStreamSyntax = urlString.replaceFirst('#narrow/channel', '#narrow/stream');
```
chrisbobbe, line 15: Let's relax this so it handles the `stream`/`channel` operator even if it's at some position other than the first. It's usually the first, but it doesn't have to be. For example, this is a valid narrow link that the web app knows how to handle: https://chat.zulip.org/#narrow/is/unread/stream/14-GSoC

I think this could be done like this:
- in `urlString`, skip past the `#narrow/` at the beginning
- split the rest by '/'
- in that list of segments, look at the even-indexed ones (i.e., `index % 2 == 0`)
- for each of those that's equal to "stream", replace it with "channel" and vice versa

(We wouldn't want to just replace all occurrences of "stream" or "channel" in the URL; we want to avoid doing things like destroying a topic operand that happens to have one of those strings in it.)

chrisbobbe (commented):
Thanks! One small comment below.

test/model/internal_link_test.dart
gnprice: It's true those are valid links, but it looks like we don't currently have any examples of them in this test file.

(Likely we should! Our implementation looks like it accepts them. But we don't right now. These tests were basically ported from zulip-mobile, because the tests there for this area were quite thorough, but that implementation doesn't accept those links.)

So for this PR, I'm happy to keep it simple in this respect.

Adding that logic would be a good followup, along with adding some test cases where /channel/ comes after /topic/, and where things come in other surprising orders.

chrisbobbe: Cool, thanks for that suggestion, Greg. Since this was my only comment, I'll mark this for integration review. 🙂

chrisbobbe:
Marking for Greg's review following https://github.com/zulip/zulip-flutter/pull/704#discussion_r1636979513 ; @Khader-1 please take a look at that comment about a followup. 🙂

gnprice:
Thanks @Khader-1, and thanks @rajveermalviya and everyone for the reviews!

This looks good — merging. (See comment above about a useful followup.)

Issue #647, opened 2024-05-07
content: Highlight non-silent mentions of the self-user


chrisbobbe:
We should highlight non-silent mentions of the self-user, following web, so that users can easily see when a mention is intended to get their attention.

For some examples (and here I'm logged in as the "Chris Bobbe" account):

<img width="224" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/8568761a-821b-415f-8983-e136d40cfec5">

<img width="150" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/f4ca8c4f-ede0-4efc-9169-08e672dedc48">

Notice the highlighting is done in two ways:

- The text of the mention itself is styled
- The whole message gets an interesting background color

chrisbobbe:
(Should we add this to #309?)

gnprice:
Thanks. Yeah, like #646, this is part of #157.

I think rather than add to #309, we can just rely on the `a-content` label and the milestones.

alya:
In case it's helpful, note the unsubscribed stream edge case: #27765.

chrisbobbe:
The web app also colors the background of the message's corresponding date separator (if any):

<img width="746" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/3e33b0b4-7ae6-41b4-836f-be3864e8ca88">

We should probably follow web.

PR #706, opened 2024-05-24
content: Make **bold** bolder even in spoiler headers and h1/h2/etc.

### LINKED ISSUE #705: content: Bold spans aren't bolder than surrounding text in spoiler headers or h1/h2/etc.
Bold spans should be bolder than surrounding text, even when that surrounding text is bold because it's in a spoiler header or an h1/h2/etc. (or column headings in tables, with #360).

| Current | Better |
| --- | --- |
| 75927CB2-41BB-415B-9530-CE418DBB9368 | 445F7D01-3E38-4A19-B5CD-7746474655D5 |

In fact, in the spoiler header, the bold text is _less_ bold. That's because, following web, we hard-code bold spans to weight 600 but spoiler headers to weight 700.

chrisbobbe:
This fixes #705 by making boldness context-dependent, much like we did for some font sizes in #544.

That user-facing change is nice but fairly subtle. I'm also hoping this will be helpful by adding content tests for boldness in the first place; we didn't have any before. And I've set a pattern in the `UserMention` tests that should help reduce confusion when we write boldness tests for #647. (Confusion mainly related to the fact that `UserMention` is in a `WidgetSpan`.)

The need for more helpers in lib/widgets/text.dart is another reminder that it'll be great to have https://github.com/flutter/flutter/issues/148026 fixed upstream. But (as with the existing helpers there) I'm hoping this handles things in a way that makes it straightforward to migrate once that's done.

Fixes: #705

| Before | After |
| --- | --- |
| 75927CB2-41BB-415B-9530-CE418DBB9368 | 445F7D01-3E38-4A19-B5CD-7746474655D5 |

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -723,6 +723,12 @@ class InlineContent extends StatelessWidget {
     required this.nodes,
   }) {
     assert(style.fontSize != null);
+    assert(
+      style.debugLabel!.contains('weightVariableTextStyle')
+      // ([ContentTheme.textStylePlainParagraph] applies [weightVariableTextStyle])
+      || style.debugLabel!.contains('ContentTheme.textStylePlainParagraph')
+      || style.debugLabel!.contains('bolderWghtTextStyle')
+    );
     _builder = _InlineContentBuilder(this);
   }
 
@@ -733,6 +739,7 @@ class InlineContent extends StatelessWidget {
   ///
   /// Must set [TextStyle.fontSize]. Some descendant spans will consume it,
   /// e.g., to make their content slightly smaller than surrounding text.
+  /// Similarly must set a font weight using [weightVariableTextStyle].
   final TextStyle style;
 
   final List<InlineContentNode> nodes;
@@ -832,7 +839,7 @@ class _InlineContentBuilder {
   }
 
   InlineSpan _buildStrong(StrongNode node) => _buildNodes(node.nodes,
-    style: weightVariableTextStyle(_context!, wght: 600));
+    style: bolderWghtTextStyle(widget.style, by: 200));
 
   InlineSpan _buildDeleted(DeletedNode node) => _buildNodes(node.nodes,
     style: const TextStyle(decoration: TextDecoration.lineThrough));
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -252,6 +252,11 @@ class Heading extends StatelessWidget {
           fontSize: kBaseFontSize * emHeight,
           height: 1.4,
         )
+          // Could set boldness relative to ambient text style, which itself
+          // might be bolder than normal (e.g. in spoiler headers).
+          // But this didn't seem like a clear improvement and would make inline
+          // **bold** spans less distinct; discussion:
+          //   https://github.com/zulip/zulip-flutter/pull/706#issuecomment-2141326257
           .merge(weightVariableTextStyle(context, wght: 600)),
         node: node));
   }
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -209,7 +209,7 @@ TextStyle weightVariableTextStyle(BuildContext context, {
   double? wght,
   double? wghtIfPlatformRequestsBold,
 }) {
-  double value = wght ?? FontWeight.normal.value.toDouble();
+  double value = wght ?? wghtFromFontWeight(FontWeight.normal);
   if (MediaQuery.boldTextOf(context)) {
     // The framework has a condition on [MediaQueryData.boldText]
     // in the [Text] widget, but that only affects `fontWeight`.
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -1,4 +1,5 @@
 import 'dart:io';
+import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
@@ -284,6 +285,23 @@ FontWeight clampVariableFontWeight(double wght) {
       else                 return FontWeight.w900;
     }
   }
+}
+
+/// A "wght" value extracted from a [TextStyle].
+///
+/// Returns the value in [TextStyle.fontVariations] if present.
+/// If that's absent but a [TextStyle.fontWeight] is present,
+/// returns [wghtFromFontWeight] for that value.
+///
+/// The returned value already reflects any response to the system bold-text
+/// setting, so if the [TextStyle] was built using [weightVariableTextStyle],
+/// this value might be larger than the `wght` that was passed to that function.
+double? wghtFromTextStyle(TextStyle style) {
+  double? result = style.fontVariations?.firstWhereOrNull((v) => v.axis == 'wght')?.value;
+  if (result == null && style.fontWeight != null) {
+    result = wghtFromFontWeight(style.fontWeight!);
+  }
+  return result;
 }
 
 /// A good guess at a font's "wght" value to match a given [FontWeight].
--- a/lib/widgets/text.dart
+++ b/lib/widgets/text.dart
@@ -264,6 +264,44 @@ double bolderWght(double baseWght, {double by = 300}) {
   return clampDouble(baseWght + by, kWghtMin, kWghtMax);
 }
 
+/// A [TextStyle] whose [FontVariation] "wght" and [TextStyle.fontWeight]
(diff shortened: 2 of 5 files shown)


chrisbobbe:
Screenshots for the last commit:

| Before | After |
| --- | --- |
| B2CBF31C-9538-4C4F-B3EF-E0FCA1B0215D | AAE42E6F-D39D-4CEF-A671-D8AC4D8A5DCA |

chrisbobbe:
Oh, and here are versions of all the above screenshots with the system "bold text" setting on:

| Before | After |
| --- | --- |
| 349819CE-5D35-41E6-8C26-4F5C79401034 | 5A4FD1BB-8AC8-4824-B8ED-D8F59AA7A699 |
| 87C61FD1-EE31-4393-8D22-9EDB0C5D0D3F | FA2E9B62-0E17-41C4-AD0E-52FCBADF22DF |

lib/widgets/text.dart
```diff
@@ -286,6 +327,23 @@ FontWeight clampVariableFontWeight(double wght) {
...
+///
+/// Returns the value in [TextStyle.fontVariations] if present.
+/// If that's absent but a [TextStyle.fontWeight] is present,
+/// returns [wghtFromFontWeight] for that value.
+///
+/// The returned value already reflects any response to the system bold-text
+/// setting, so if the [TextStyle] was built using [weightVariableTextStyle],
+/// this value might be larger than the `wght` that was passed to that function.
+double? wghtFromTextStyle(TextStyle style) {
+  double? result = style.fontVariations?.singleWhereOrNull((v) => v.axis == 'wght')?.value;
```
gnprice, line 340:
```suggestion
  double? result = style.fontVariations?.firstWhereOrNull((v) => v.axis == 'wght')?.value;
```

The "single" behavior means that if for some reason there are multiple `wght` variations, this would ignore them all and fall back to `fontWeight`. That doesn't seem intended.

I don't feel a strong need to have this code check that there aren't multiple of them. (In fact it looks like its one caller already has an assert to that effect.) If we're not going to check, then it should just do the simplest thing, which is to break the loop and return as soon as it finds one. That's what the "first" version does.

test/widgets/content_test.dart
```diff
@@ -154,6 +155,43 @@ void main() {
     });
   }
 
+  /// To check font weight, set up tests with and without bold-text setting.
```
gnprice, line 158: The "To check" here reads like it's an instruction: "If you want to check font weight, set up tests thus-and-so."

A direct fix for that would be:
> Set up tests with and without bold-text setting, to check font weight.

But I feel like what I want in the summary line is more of a description of what the said tests will test. Here's a version:

```suggestion
  /// Test the font weight found by [styleFinder] in the rendering of [content].
  ///
  /// The weight will be expected to be [expectedWght] when the system
  /// bold-text setting is not set, and to vary appropriately when it is set.
```

test/widgets/content_test.dart
```diff
@@ -154,6 +155,43 @@ void main() {
...
+  ///
+  /// For [expectedWght], pass the value that is expected when the system
+  /// bold-text setting is disabled. This will set up a test with the setting is
+  /// disabled, expecting that value, and also a test with the setting enabled,
+  /// expecting an appropriate larger value.
+  ///
+  /// [styleFinder] must return the [TextStyle] containing the "wght"
+  /// (in [TextStyle.fontVariations]) and the [TextStyle.fontWeight]
+  /// to be checked.
+  testFontWeight(String description, {
```
gnprice, line 168: nit: explicit return type (so it doesn't look like a function call)

test/widgets/content_test.dart
```diff
@@ -469,7 +526,84 @@ void main() {
     check((ratioInHeader - ratioInParagraph).abs()).isLessThan(0.001);
   }
 
-  testContentSmoke(ContentExample.strong);
+  group('strong (bold)', () {
+    testContentSmoke(ContentExample.strong);
+
+    TextStyle findWordBold (tester) {
```
gnprice, line 532: nit: space

lib/widgets/content.dart
```diff
@@ -722,7 +723,15 @@ class InlineContent extends StatelessWidget {
     required this.style,
     required this.nodes,
   }) {
-    assert(style.fontSize != null);
+    assert(
+      style.fontSize != null
+      && (
+        style.debugLabel!.contains('weightVariableTextStyle')
```
gnprice, line 729: nit: just use two asserts; formatting gets simpler/shorter, and the two conditions aren't closely related anyway

lib/widgets/content.dart
```diff
@@ -797,7 +807,7 @@ class _InlineContentBuilder {
       // and our parser doesn't.  So don't do anything here.
       return const TextSpan(text: "");
     } else if (node is StrongNode) {
-      return _buildStrong(node);
+      return _buildStrong(ambientTextStyle: widget.style, node);
```
gnprice, line 810: No need to pass this to one of this class's own methods — it can look at `widget.style` directly, just like this `_buildNode` method can.

test/widgets/content_test.dart
```diff
@@ -469,7 +526,84 @@ void main() {
...
+      styleFinder: findWordBold,
+    );
+
+    for (final level in HeadingLevel.values) {
+      final name = level.name;
+      assert(RegExp(r'^h[1-6]$').hasMatch(name));
+      testFontWeight('in $name',
+        expectedWght: 800,
+        // **bold**
+        content: plainContent('<$name><strong>bold</strong></$name>'),
```
gnprice, line 550: This comment is inaccurate because there's the heading element.

Fine to have this automatic generation of six test cases, just need to adjust the comment. E.g.:
```suggestion
        // # **bold**, ## **bold**, ### **bold**, etc.
        content: plainContent('<$name><strong>bold</strong></$name>'),
```

gnprice (commented):
Thanks! Various nits below.

> Screenshots for the last commit. ("h1: regular" has weight 900, and "bold" has weight 1000…I don't see much of a difference there ~, but that's fine~. This might be a reason to drop the last commit; distinguishing a bold span from surrounding text seems important and worth preserving as completely as possible.)

Hmm, yeah. What does a screenshot look like without that last commit? (That last commit being:
a4604c7fc content: Make h1 font weight combine with spoiler-header font weight
)

chrisbobbe:
> Hmm, yeah. What does a screenshot look like without that last commit? (That last commit being:
> https://github.com/zulip/zulip-flutter/commit/a4604c7fc2f0fbc56ed064e051836e4c9ff161e1 content: Make h1 font weight combine with spoiler-header font weight
> )

Sure, here it is:

| Without that commit | With |
| --- | --- |
| 8AB20836-0A67-4291-B76C-625C323A72E7 | C45AAB03-31B3-46CA-85EB-EBCFA64B769B |
| Bold text setting on: 3C1902DA-BF77-443B-A0E1-A9D209ECB775 | Bold text setting on: A6E8AD32-E2AF-4AC1-B261-644F12442BD4 |

lib/widgets/text.dart
chrisbobbe: Makes sense; I agree.

chrisbobbe:
Thanks for the review! Revision pushed, and see the screenshots I posted just before this. I'd be happy to remove that last commit, this keeping clearer emphasis on bold spans when they appear in h1/h2/etc. in spoiler headers.

lib/widgets/content.dart
```diff
@@ -831,8 +839,9 @@ class _InlineContentBuilder {
     }
   }
 
-  InlineSpan _buildStrong(StrongNode node) => _buildNodes(node.nodes,
-    style: weightVariableTextStyle(_context!, wght: 600));
+  InlineSpan _buildStrong(StrongNode node) =>
+    _buildNodes(style: bolderWghtTextStyle(widget.style, by: 200),
+      node.nodes);
```
gnprice, line 844: nit: unnecessary churn here, and makes it look different from the similar neighboring methods; can keep `node.nodes` first and the existing formatting

gnprice (commented):
Thanks!

Looks like the last commit doesn't affect the bold-text-setting-on case (the screenshots look identical to me except for the signal bars at top right), and for the default case I think it's probably better without. So let's drop that commit.

Other than that, one nit below, and then please go ahead and merge!

chrisbobbe:
Thanks! Done.

Issue #630, opened 2024-04-24
Rename "stream" to "channel" in the UI


gnprice:
This is the zulip-flutter counterpart of:
* https://github.com/zulip/zulip-mobile/issues/5827
* https://github.com/zulip/zulip/issues/28468

Unlike in zulip-mobile, in zulip-flutter let's do this unconditionally: we'll refer to a channel as a "channel" even on older servers where the web UI says "stream". By the time we roll out zulip-flutter to general users some months from now, it'll have been "channel" on Zulip Cloud and in the latest server release for at least a few months, so probably a relatively small fraction of users will be on older servers that use the old terminology, and it's OK if they see the new terminology in their mobile app.

(When we add wildcard mentions #234, those will need to condition on the server version, because it's a correctness issue: sending `@**channel**` to an older server just won't work. (I'll update #234 to note that.) But that's a lot more localized.)

This issue is purely about the user-facing strings: everywhere we say "stream" to the user, we should say "channel" instead. I'll file a follow-up issue for renaming identifiers in our codebase.

Issue #277, opened 2023-08-17
Sweep through UI strings to internationalize


gnprice:
This is a followup to #275. After we have that basic framework for using translated strings in the UI, we'll want to sweep through our existing code and convert it all to use that framework.

Initially this won't do much of anything from a user's perspective, because we won't have more than a handful of strings translated until we wire things up to a service where people can contribute such translations, #276.

Nevertheless it's useful to do this early, because it sets the pattern so that we naturally follow it in new code, rather than having more code that needs to change in an eventual sweep later. We may also come across places that need a `context` object and didn't before, or otherwise call for refactoring, and it's convenient to do that sooner so that other changes to the same code are based on the new structure from the beginning.

Other notes:
* Many of these spots are marked with `TODO(i18n)`, though many others are not.
* Some spots marked `TODO(i18n)` are for different needs like formatting dates and times. Those are out of scope of this issue.
* If some strings are especially difficult to wire up for translation for some reason, we can mark them with `TODO(i18n)` and also file followup issues, and then leave them out for this issue.

Detective-Khalifah:
A refactor in essence

gnprice:
The bulk of this is done now, after #1148. (We've also largely been internationalizing new code as we go, over the past year and a half since this issue was filed.)

There's one known remaining spot to be handled, and then some more sweeping to do to confirm that that's all:
https://github.com/zulip/zulip-flutter/pull/1148#issuecomment-2635404118

PIG208:
Opened #1325 to comment on the remaining case.

Follow-up issues for more complex translations:

#1285 for `_errorUnimplemented`

```dart
InlineSpan _errorUnimplemented(UnimplementedNode node, {required BuildContext context}) {
  final contentTheme = ContentTheme.of(context);
  final errorStyle = contentTheme.textStyleError;
  final errorCodeStyle = contentTheme.textStyleErrorCode;
  // For now this shows error-styled HTML code even in release mode,
  // because release mode isn't yet about general users but developer demos,
  // and we want to keep the demos honest.
  // TODO(#194) think through UX for general release
  // TODO(#1285) translate this
  final htmlNode = node.htmlNode;
  if (htmlNode is dom.Element) {
    return TextSpan(children: [
      TextSpan(text: "(unimplemented:", style: errorStyle),
      TextSpan(text: htmlNode.outerHtml, style: errorCodeStyle),
      TextSpan(text: ")", style: errorStyle),
    ]);
[...]

Issue #276, opened 2023-08-17
Wire up to a platform for contributing translations


gnprice:
This is a followup to #275. After we have that basic framework for using translated strings in the UI, we'll want to set things up so that Zulip's volunteer translators can contribute translations.

The obvious direction to go for this _would_ be to wire things up to Transifex, which is what we use [for zulip-mobile](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/translations.md) and [for Zulip web, desktop, and server](https://zulip.readthedocs.io/en/latest/translating/translating.html#translators-workflow). But… the format the Dart ecosystem supports for internationalized messages is ARB, and [Transifex does not support ARB](https://chat.zulip.org/#narrow/stream/58-translation/topic/alternatives.20to.20Transifex.3F/near/1588639).

If we were determined to stick with Transifex, we could build tooling to work around that — ARB files appear to be just JSON with some reasonable schema, so it shouldn't be hard to translate ARB to some other format and back. But we've independently [been wanting to move off Transifex](https://chat.zulip.org/#narrow/stream/58-translation/topic/alternatives.20to.20Transifex.3F/near/1431909), to a more open platform like Weblate. So probably this will involve moving to a new service.

TBD how to handle that from a project-management perspective. Some thoughts:

* The end state should be that the whole Zulip project is using a single translation service, be that Weblate or something else. We should avoid dragging on for a long period (like many months) on two services, or migrating and then finding ourselves migrating again — those are likely to be a burden on our translation contributors.

* Translating zulip-flutter may not need to block on migrating the whole Zulip project, though. It might instead serve as the pilot, followed by migrating the rest of the project. We'll just want, when we commit to a service for zulip-flutter, to be confident that we will then be able to migrate the rest of the project there without too much further fuss.

* So in particular before we broadly ask our translators to go start translating on a new service, we'll want to have poked around both to understand exactly how we want it set up for zulip-flutter, and also to understand the setup we'll want for other parts of the project at least well enough to convince ourselves that there won't be any dealbreakers later.

gnprice:
One quirk in Dart's `package:intl` which we might have to handle when we go to implement this issue: for number/plural forms, the package expects forms named `=1` and `=0` when it actually means the behaviors whose standard names are `one` and `zero`. See:
* https://github.com/dart-lang/i18n/issues/31
* https://github.com/dart-lang/i18n/issues/21

[...]

PR #707, opened 2024-05-25
ui: Rename streams to channels in user-facing strings

Khader-1:
Fixes: #630


rajveermalviya (approved):
Thanks, This looks good to me :)

Updating `l10n` data labels & template variables can be done later, as mentioned in https://github.com/zulip/zulip-flutter/issues/630#issuecomment-2130145993

gnprice:
Thanks @rajveermalviya for the first review!

When advancing a review to the next stage, let's be sure to ping on the thread the next reviewer, as well as update the labels, to help make sure everyone's on the same page. So @kenclary over to you :slightly_smiling_face: 

(And please also take a look in your inbox for the invite I've sent for this repo; that'll let you update labels and reviewers in turn, but as a side effect it'll also make it possible to name you in the GitHub "request review" feature, because it'll make GitHub aware that you're connected to this org/repo.)

kenclary (approved):
I love a straightforward PR.

chrisbobbe (commented):
Thanks @Khader-1! Some small comments on the commit sequence:

Here are the commits in this revision:

3437549f1 intl: Use channel instead of stream
122739037 ui: Rename stream to channel in user-facing strings
b407423d6 notif: Rename stream to channel in displayed notifications

The first and second commits are both about changing strings that appear in the in-app UI, which at this point means all the UI except notifications. So I think it would be fine to squash them into one commit, with a summary line like:

```
ui: Rename stream to channel in user-facing strings, except notifications 
```

(If you don't want to squash them together, then each commit should have some more words about exactly which strings are being treated. Basically the second commit is about the in-app strings that are still TODOs for #277.)

I think it would also be fine to squash the notifications-strings commit into that as well (i.e., squashing your revision's _three_ commits into one), and have the commit message say that all user-facing strings are treated in the commit. That's still in line with our convention that a commit represents one coherent idea.

Khader-1:
Thanks @chrisbobbe! A new revision with a single commit PTAL.

gnprice:
Thanks all! Looks good — merging.

I'll add a second commit on top, to update the Japanese translation. :slightly_smiling_face: (That's a small partial translation, which we have just for the sake of end-to-end testing that our i18n setup within the app works, pending #276 which will let us take translation contributions broadly like we do for other Zulip codebases.)

FTR, the searches I did to confirm the rename was complete were:
```
$ git grep -i '".*stream'
$ git grep -i \''.*stream'
```
Most of the results were in `test/` or in generated `*.g.dart` files for deserializing the Zulip server API; I looked through the rest and confirmed they were also either in the API, in a comment, or otherwise not user-facing.

Looking forward to seeing most of those — and matches for `git grep -i stream` more generally — go away as part of #631! That will be a bigger task: definitely many commits, and probably best as several PRs, to keep each PR down to a convenient size for review.

Issue #699, opened 2024-05-22
is the vision for this to be the zulip client for mobile and desktop?


naderm11:
Is the vision for this to be the Zulip client for mobile and desktop?
If so, I would like to help translate the whole thing into Arabic, but will it support the RTL layout in the UI?

gnprice:
This app will become the official Zulip mobile client. About desktop, see:
https://github.com/zulip/zulip-flutter#desktop-support

On RTL layout, see:
* https://github.com/zulip/zulip-flutter/issues/77

We'd be glad to have contributions to the Arabic translation. For contributing to Zulip's translations, see:
https://zulip.readthedocs.io/en/latest/translating/translating.html
Currently zulip-flutter itself isn't set up for translations (that's #276); but when it is, it'll use the same platform as the rest of the Zulip project, documented on that page.

As described at #77, none of the existing Zulip clients (neither the existing/legacy zulip-mobile client, nor the web or desktop clients) currently support an RTL layout for the UI. But we nevertheless have an Arabic translation for them, as well as a near-complete Persian translation, thanks to contributors who've found those worthwhile. It would be great to have another person helping out on the Arabic translation.

naderm11:
Hi @gnprice, Thanks for your reply.
Following the provided translation guide seems to present many logistical challenges, and it lacks specific instructions for Arabic. To explore how the app would handle translation, I translated the app and generated the `app_ar.arb` file based on the keys in `app_en.arb.`

The app works well with RTL, and no issues were found during testing in the simulator. Some keys are missing, but these can be added over time. The default font could have looked better for Arabic. I attempted to add [font](https://fonts.google.com/share?selection.family=Tajawal:wght@200;300;400;500;700;800;900) to improve the user experience, but I had some trouble incorporating it into Flutter.

The app needs some configuration to recognize the new language, and if the translation is not there, the app should display the English text; right now, it shows nothing.
I'm unsure how strictly we need to follow the language guide, but making the capital effort and translating the available keys now won't hurt :)
The file contains the Arabic translation, and it's not machine-generated, so the terminologies follow the standard webapp terms used in Arabic; some terminologies are modified to fit the Zulip context, like streams and other technologies that specific to the app; it can be edited it later if it does not fit the context
[Zulip_flutter_ar_language_file .zip](https://github.com/zulip/zulip-flutter/files/15419521/Zulip_flutter_ar_language_file.zip)

gnprice:
Cool, thanks for doing that work.

[...]

PR #708, opened 2024-05-26
Added app_ar.arb to support Arabic translation ...

naderm11:
and modified app.dart to support RTL layout. Not very familiar with Flutter code, so please review :)

Fixes: #803


naderm11:
I am not sure why the checks are failing, please let me know to do the needed changes

naderm11:
fix #699

lib/widgets/app.dart
```diff
@@ -117,7 +118,12 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
         final initialAccountId = globalStore.accounts.firstOrNull?.id;
         return MaterialApp(
           title: 'Zulip',
-          localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+          localizationsDelegates: const [
+            ...ZulipLocalizations.localizationsDelegates,
+            GlobalMaterialLocalizations.delegate,
```
gnprice, line 123: Interesting, yeah, this looks like probably a needed change.

Just to confirm my understanding: what happens if you leave this change out?

Let's put this change in a commit of its own, perhaps at the start of the branch. (And the commit message can have the answer to my question about why it's needed.)

lib/widgets/app.dart
```diff
@@ -117,7 +118,12 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+            GlobalWidgetsLocalizations.delegate,
+            GlobalCupertinoLocalizations.delegate,
```
gnprice, line 125: nit: place in logical order:
```suggestion
            ...ZulipLocalizations.localizationsDelegates,
            GlobalMaterialLocalizations.delegate,
            GlobalCupertinoLocalizations.delegate,
            GlobalWidgetsLocalizations.delegate,
```

If our app-specific delegates go first, then the next-most-specific should go next, and so on to the most general delegate going at the end. The `flutter/material` package depends on `flutter/cupertino` which depends on `flutter/widgets`, so that gives the ordering.

lib/widgets/app.dart
```diff
@@ -129,7 +135,11 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
                 (_) => widget._declareReady());
             }
             GlobalLocalizations.zulipLocalizations = ZulipLocalizations.of(context);
-            return child!;
+            final locale = Localizations.localeOf(context); // Get locale from context
```
gnprice, line 138: nit: this comment says only information that's equally readable from the code itself, so just leave it out:
```suggestion
            final locale = Localizations.localeOf(context);
```

lib/widgets/app.dart
```diff
@@ -129,7 +135,11 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+            return Directionality(
+              textDirection: locale.languageCode == 'ar' ? TextDirection.rtl : TextDirection.ltr,
```
gnprice, line 140: Huh, is this needed? I would have thought that Flutter would supply this automatically, in `MaterialApp` (and `WidgetsApp` which it's built around).

Looking in the Flutter source, it sure does look that way. It's also mentioned in the docs of [`Localization`](https://main-api.flutter.dev/flutter/widgets/Localizations-class.html), which is one of the things [`WidgetsApp`](https://main-api.flutter.dev/flutter/widgets/WidgetsApp-class.html) provides.

Suppose you leave this change out; how does the behavior change?

(If we do need a version of this change, it'll be another example of a change that should be its own separate commit.)

assets/l10n/app_en.arb
```diff
@@ -1,4 +1,5 @@
 {
+  "@@locale": "en",
```
gnprice, line 2: Hmm interesting. What's the effect of this / what led you to add it?

gnprice (commented):
Thanks @naderm11 for the contribution!

The main thing we'll want adjusted in these changes before we can merge them is to take each of the different ideas that are involved here and separate them into their own commit, with their own commit message explaining what they're for. You can read all about that [here](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html), and more briefly [here](https://zulip.readthedocs.io/en/latest/git/overview.html); but I've also left specific comments below about what that means for this PR.

gnprice:
@naderm11 Do you think you'll be able to return to this PR and answer the questions above? It'd be great to have this completed and merged.

chrisbobbe:
This would be helpful indeed. 🙂

Issue #613, opened 2024-04-03
Try faster fade-out on mark-as-read button


gnprice:
This is a followup to:
* #562 
* #565

See [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/messages.20shift.20on.20mark-as-read/near/1773710).

alya:
I've been noticing this one when I use the app.

gnprice:
Specifically, per [Vlad's suggestion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read/near/1800824) yesterday:
* Let's have it fade out in 100ms, rather than 2000ms.
* Let's also try having the animation scale the button down, as well as reducing opacity.

In the future it would be good to have the button respond optimistically, without waiting for an ack from the server. (I'll file an issue for that.) Then we can make the animation longer again, because there won't be the built-in delay from waiting for the server.

Issue #417, opened 2023-11-27
Change the style of touch feedback 


gnprice:
Currently when you touch a button, or an item in a list like the inbox screen or streams screen, we use Material Design's "ink splash" effect to give feedback showing you what you touched.

@terpimost writes at https://github.com/zulip/zulip-flutter/issues/339#issuecomment-1827028273 :
> I don't assume we have material UI components and their tap animations... I suggest to not bring it to iOS version at least.
> I think it is ok to darken the hit area or a button on tap. We can also scale down the button a bit as a default option. It is also ok to not bother for the smaller elements which user might not even see under the finger.

So we can try one of those instead.

Whatever effect we choose (or suite of different effects for different situations), we should test to see we're happy with how it behaves not only on buttons, but on those list items like a stream on the streams screen or a conversation in the inbox.

chrisbobbe:
Many `IconButton`s could probably be replaced with `ZulipIconButton` added in ec6715bef63f5b29928db7ede5149de7f52b6638, which doesn't have this splash effect. See Figma: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=7728-10468&m=dev

PR #710, opened 2024-05-30
ui: Use faster animation for mark as read

Khader-1:
Fixes: #613

Final Result:

https://github.com/user-attachments/assets/95912caf-2b7d-4676-b07e-a74c551aace0

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -516,6 +516,7 @@ class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
             padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 10 - ((48 - 38) / 2)),
             child: FilledButton.icon(
               style: FilledButton.styleFrom(
+                splashFactory: NoSplash.splashFactory,
                 minimumSize: const Size.fromHeight(38),
                 textStyle:
                   // Restate [FilledButton]'s default, which inherits from
@@ -543,7 +544,7 @@ class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
   }
 }
 
-class MarkAsReadAnimation extends StatelessWidget {
+class MarkAsReadAnimation extends StatefulWidget {
   final bool loading;
   final bool hidden;
   final Widget child;
@@ -556,12 +557,33 @@ class MarkAsReadAnimation extends StatelessWidget {
   });
 
   @override
+  State<MarkAsReadAnimation> createState() => _MarkAsReadAnimationState();
+}
+
+class _MarkAsReadAnimationState extends State<MarkAsReadAnimation> {
+  bool _isPressed = false;
+
+  void _setIsPressed(bool isPressed) {
+    setState(() {
+      _isPressed = isPressed;
+    });
+  }
+
+  @override
   Widget build(BuildContext context) {
-    return AnimatedOpacity(
-      opacity: hidden ? 0 : loading ? 0.5 : 1,
-      duration: const Duration(milliseconds: 500),
-      curve: Curves.easeOut,
-      child: child);
+    return GestureDetector(
+      onTapDown: (_) => _setIsPressed(true),
+      onTapUp: (_) => _setIsPressed(false),
+      onTapCancel: () => _setIsPressed(false),
+      child: AnimatedScale(
+        scale: _isPressed ? 0.95 : 1,
+        duration: const Duration(milliseconds: 100),
+        curve: Curves.easeOut,
+        child: AnimatedOpacity(
+          opacity: widget.hidden ? 0 : widget.loading ? 0.5 : 1,
+          duration: const Duration(milliseconds: 500),
+          curve: Curves.easeOut,
+          child: widget.child)));
   }
 }
 
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -559,7 +559,7 @@ class MarkAsReadAnimation extends StatelessWidget {
   Widget build(BuildContext context) {
     return AnimatedOpacity(
       opacity: hidden ? 0 : loading ? 0.5 : 1,
-      duration: Duration(milliseconds: hidden ? 2000 : 300),
+      duration: const Duration(milliseconds: 500),
       curve: Curves.easeOut,
       child: child);
   }
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -505,10 +505,9 @@ class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
 
     return IgnorePointer(
       ignoring: areMessagesRead,
-      child: AnimatedOpacity(
-        opacity: areMessagesRead ? 0 : _loading ? 0.5 : 1,
-        duration: Duration(milliseconds: areMessagesRead ? 2000 : 300),
-        curve: Curves.easeOut,
+      child: MarkAsReadAnimation(
+        loading: _loading,
+        hidden: areMessagesRead,
         child: SizedBox(width: double.infinity,
           // Design referenced from:
           //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=132-9684&mode=design&t=jJwHzloKJ0TMOG4M-0
@@ -541,6 +540,28 @@ class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
               onPressed: _loading ? null : () => _handlePress(context),
               icon: const Icon(Icons.playlist_add_check),
               label: Text(zulipLocalizations.markAllAsReadLabel))))));
+  }
+}
+
+class MarkAsReadAnimation extends StatelessWidget {
+  final bool loading;
+  final bool hidden;
+  final Widget child;
+
+  const MarkAsReadAnimation({
+    super.key,
+    required this.loading,
+    required this.hidden,
+    required this.child
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return AnimatedOpacity(
+      opacity: hidden ? 0 : loading ? 0.5 : 1,
+      duration: Duration(milliseconds: hidden ? 2000 : 300),
+      curve: Curves.easeOut,
+      child: child);
   }
 }
 
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -458,10 +458,17 @@ class ScrollToBottomButton extends StatelessWidget {
   }
 }
 
-class MarkAsReadWidget extends StatelessWidget {
+class MarkAsReadWidget extends StatefulWidget {
   const MarkAsReadWidget({super.key, required this.narrow});
 
   final Narrow narrow;
+
+  @override
+  State<MarkAsReadWidget> createState() => _MarkAsReadWidgetState();
+}
+
+class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
+  bool _loading = false;
 
   void _handlePress(BuildContext context) async {
     if (!context.mounted) return;
@@ -469,19 +476,22 @@ class MarkAsReadWidget extends StatelessWidget {
     final store = PerAccountStoreWidget.of(context);
     final connection = store.connection;
     final useLegacy = connection.zulipFeatureLevel! < 155;
+    setState(() => _loading = true);
 
     try {
-      await markNarrowAsRead(context, narrow, useLegacy);
+      await markNarrowAsRead(context, widget.narrow, useLegacy);
     } catch (e) {
       if (!context.mounted) return;
       final zulipLocalizations = ZulipLocalizations.of(context);
-      await showErrorDialog(context: context,
+      showErrorDialog(context: context,
         title: zulipLocalizations.errorMarkAsReadFailedTitle,
         message: e.toString()); // TODO(#741): extract user-facing message better
       return;
+    } finally {
+      setState(() => _loading = false);
     }
     if (!context.mounted) return;
-    if (narrow is CombinedFeedNarrow && !useLegacy) {
+    if (widget.narrow is CombinedFeedNarrow && !useLegacy) {
       PerAccountStoreWidget.of(context).unreads.handleAllMessagesReadSuccess();
     }
   }
@@ -490,13 +500,13 @@ class MarkAsReadWidget extends StatelessWidget {
   Widget build(BuildContext context) {
     final zulipLocalizations = ZulipLocalizations.of(context);
     final store = PerAccountStoreWidget.of(context);
-    final unreadCount = store.unreads.countInNarrow(narrow);
+    final unreadCount = store.unreads.countInNarrow(widget.narrow);
     final areMessagesRead = unreadCount == 0;
 
     return IgnorePointer(
       ignoring: areMessagesRead,
       child: AnimatedOpacity(
-        opacity: areMessagesRead ? 0 : 1,
+        opacity: areMessagesRead ? 0 : _loading ? 0.5 : 1,
         duration: Duration(milliseconds: areMessagesRead ? 2000 : 300),
         curve: Curves.easeOut,
         child: SizedBox(width: double.infinity,
@@ -507,8 +517,6 @@ class MarkAsReadWidget extends StatelessWidget {
             padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 10 - ((48 - 38) / 2)),
             child: FilledButton.icon(
               style: FilledButton.styleFrom(
-                // TODO(#95) need dark-theme colors (foreground and background)
-                backgroundColor: _UnreadMarker.color,
                 minimumSize: const Size.fromHeight(38),
                 textStyle:
                   // Restate [FilledButton]'s default, which inherits from
@@ -522,11 +530,17 @@ class MarkAsReadWidget extends StatelessWidget {
                     height: (23 / 18))
                   .merge(weightVariableTextStyle(context, wght: 400))),
                 shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
+              ).copyWith(
+                // Give the buttons a constant color regardless of whether their
+                // state is disabled, pressed, etc.  We handle those states
+                // separately, via MarkAsReadAnimation.
+                // TODO(#95) need dark-theme colors (foreground and background)
+                foregroundColor: WidgetStateColor.resolveWith((_) => Colors.white),
+                backgroundColor: WidgetStateColor.resolveWith((_) => _UnreadMarker.color),
               ),
-              onPressed: () => _handlePress(context),
+              onPressed: _loading ? null : () => _handlePress(context),
               icon: const Icon(Icons.playlist_add_check),
-              label: Text(zulipLocalizations.markAllAsReadLabel))))),
-    );
+              label: Text(zulipLocalizations.markAllAsReadLabel))))));
   }
 }
 
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -883,6 +883,65 @@ void main() {
           unreadMessageIds: [message.id]),
       ]);
 
+      group('MarkAsReadAnimation', () {
+        void checkAppearsLoading(WidgetTester tester, bool expected) {
+          final semantics = tester.firstWidget<Semantics>(find.descendant(
+            of: find.byType(MarkAsReadWidget),
+            matching: find.byType(Semantics)));
+          check(semantics.properties.enabled).equals(!expected);
+
+          final opacity = tester.widget<AnimatedOpacity>(find.descendant(
+            of: find.byType(MarkAsReadWidget),
+            matching: find.byType(AnimatedOpacity)));
+          check(opacity.opacity).equals(expected ? 0.5 : 1.0);
+        }
+
+        testWidgets('loading is changed correctly', (WidgetTester tester) async {
+          final narrow = TopicNarrow.ofMessage(message);
+          await setupMessageListPage(tester,
+            narrow: narrow, messages: [message], unreadMsgs: unreadMsgs);
+          check(isMarkAsReadButtonVisible(tester)).isTrue();
+
+          connection.prepare(
+            delay: const Duration(milliseconds: 2000),
+            json: UpdateMessageFlagsForNarrowResult(
+              processedCount: 11, updatedCount: 3,
+              firstProcessedId: null, lastProcessedId: null,
+              foundOldest: true, foundNewest: true).toJson());
+
+          checkAppearsLoading(tester, false);
+
+          await tester.tap(find.byType(MarkAsReadWidget));
+          await tester.pump();
+          checkAppearsLoading(tester, true);
+
+          await tester.pump(const Duration(milliseconds: 2000));
+          checkAppearsLoading(tester, false);
+        });
+
+        testWidgets('loading is changed correctly if request fails', (WidgetTester tester) async {
+          final narrow = TopicNarrow.ofMessage(message);
+          await setupMessageListPage(tester,
+            narrow: narrow, messages: [message], unreadMsgs: unreadMsgs);
+          check(isMarkAsReadButtonVisible(tester)).isTrue();
+
+          connection.prepare(httpStatus: 400, json: {
+            'code': 'BAD_REQUEST',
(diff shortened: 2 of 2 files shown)


lib/widgets/message_list.dart
```diff
@@ -629,6 +623,24 @@ class _UnreadMarker extends StatelessWidget {
   }
 }
 
+class _AnimatedUnread extends StatelessWidget {
+  const _AnimatedUnread({required this.isRead, required this.child});
```
rajveermalviya, line 627:
```suggestion
  const _AnimatedVisibilityEaseOut({required this.visible, required this.child});
```

maybe a better name would be `_AnimatedVisibilityEaseOut`, since it takes a `Widget child` and that seems much more general interface than anything related to "unread".

lib/widgets/message_list.dart
```diff
@@ -629,6 +623,24 @@ class _UnreadMarker extends StatelessWidget {
...
+
+  final bool isRead;
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return AnimatedOpacity(
+              opacity: isRead ? 0 : 1,
```
rajveermalviya:
```suggestion
        opacity: isRead ? 0 : 1,
```
nit: indentation

lib/widgets/message_list.dart
```diff
@@ -631,15 +631,21 @@ class _AnimatedUnread extends StatelessWidget {
...
-              // Web uses 2s and 0.3s durations, and a CSS ease-out curve.
-              // See zulip:web/styles/message_row.css .
-              // As of now we wait for the ack from the server
-              // so we're going with 0.1s as soon as the ack is received.
-              duration: Duration(milliseconds: isRead ? 100 : 300),
-              curve: Curves.easeOut,
-              child: child);
+    final duration = Duration(milliseconds: isRead ? 100 : 300);
+    final visibility = isRead ? 0.0 : 1.0;
+    return AnimatedScale(
```
rajveermalviya, line 636: needs test for `AnimatedScale`.

rajveermalviya (commented):
Thanks @Khader-1, some comments :)

Khader-1:
Oh, sorry this one had to be marked as a draft as [I am still waiting for confirmation for some decisions](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Autocomplete.20for.20topic.20to.20send.20to.20.23310/near/1811767).
But thanks for the review!

gnprice:
@Khader-1 Are there still open questions in the chat discussion? If so it'd be good to bump the discussion to clarify the question you currently have; it looks to me like the questions there have been answered and we have a direction to go with.

(I think the link above is a different thread than you meant; the one for this PR looks to be [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read/near/1809629).)

Khader-1:
You're right, thank for the catch @gnprice! 
This is ready for review now @rajveermalviya.

lib/widgets/message_list.dart
```diff
@@ -643,6 +651,35 @@ class _UnreadMarker extends StatelessWidget {
...
+    required this.child});
+
+  final bool visible;
+  final bool loading;
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    final opacity = loading ? 0.5 : visible ? 1.0 : 0.0;
+    final scale = loading ? 0.95 : 1.0;
```
rajveermalviya, line 668: After loading is finished the button will scale up instead of scaling down, unlike mentioned [here](https://github.com/zulip/zulip-flutter/issues/613#issuecomment-2116318789) & [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read/near/1811157).

lib/widgets/message_list.dart
```diff
@@ -505,8 +514,7 @@ class MarkAsReadWidget extends StatelessWidget {
               ),
               onPressed: () => _handlePress(context),
```
rajveermalviya, line 515: It is possible to click on it again while previous `markNarrowAsRead` is still in progress, click should be disabled while it's in `loading` state, not sure if this should be different issue+PR or just a commit in this PR.

rajveermalviya (commented):
Thanks @Khader-1! Apart from some comments this LGTM.

lib/widgets/message_list.dart
rajveermalviya: cc @gnprice

lib/widgets/message_list.dart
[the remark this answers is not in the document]
Khader-1: Thanks! Good Idea.

lib/widgets/message_list.dart
```diff
@@ -503,10 +513,9 @@ class MarkAsReadWidget extends StatelessWidget {
                   .merge(weightVariableTextStyle(context, wght: 400))),
                 shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
               ),
-              onPressed: () => _handlePress(context),
+              onPressed: () => isLoading ? null : _handlePress(context),
```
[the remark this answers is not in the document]
kenclary: Why would it force that?

kenclary (approved):
This looks good to me.

chrisbobbe (changes requested):
Thanks, @Khader-1!

Is [this discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read/near/1813212) the latest for the design? I think there are some differences between that design and what's implemented here:

- The 95% scaling is meant for the button's pressed state, not its loading state. So when the finger goes down, it should animate to 95% size. When the finger is lifted, it should animate back to 100%. There isn't a scaling animation assigned to the loading state.
- When loading is finished, it should just fade to transparency for 500ms. I don't see Vlad describing a scaling animation for the button's exit.

Does that seem right to you, and is there a discussion I'm missing?

Khader-1:
Thanks @chrisbobbe! I misunderstood what Vlad meant with **i pressed state**. I think the new implementation matches the one you described. PTAL

lib/widgets/message_list.dart
```diff
@@ -503,10 +513,52 @@ class MarkAsReadWidget extends StatelessWidget {
...
+  void _setScale(bool isPressed) {
+    setState(() {
+      _isPressed = isPressed;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      onTapDown: (_) => _setScale(true),
```
Khader-1, line 550: We can create a custom button widget if that's expected to be used in other cases.

lib/widgets/message_list.dart
```diff
@@ -503,10 +513,52 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      onTapUp: (_) => _setScale(false),
+      onTapCancel: () => _setScale(false),
+      child: AnimatedScale(
+        scale: _isPressed ? 0.95 : 1,
+        duration: const Duration(milliseconds: 100),
```
Khader-1, line 555: I've set this to `100ms` as I found the button press being too fast to get any feed back with higher durations.

lib/widgets/message_list.dart
```diff
@@ -503,10 +513,52 @@ class MarkAsReadWidget extends StatelessWidget {
                   .merge(weightVariableTextStyle(context, wght: 400))),
                 shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
               ),
-              onPressed: () => _handlePress(context),
+              onPressed: () => loading ? null : _handlePress(context),
```
chrisbobbe, line 516:
```suggestion
              onPressed: loading ? null : () => _handlePress(context),
```

See dartdoc:

```dart
  /// Called when the button is tapped or otherwise activated.
  ///
  /// If this callback and [onLongPress] are null, then the button will be disabled.
  ///
  /// See also:
  ///
  ///  * [enabled], which is true if the button is enabled.
  final VoidCallback? onPressed;
```

This should let the `FilledButton` signal to screen-reader software that it's disabled, which is helpful for accessibility.

…I notice that this causes `FilledButton` to apply its own chosen "disabled" style to the button, making it gray, which is very different from Vlad's design. We can tune that style, with [`style.backgroundColor`](https://api.flutter.dev/flutter/material/ButtonStyle/backgroundColor.html), but unfortunately that API doesn't seem to offer a way to animate the transition. So I think we just keep the animation you've added (and keep the null `onPressed` for the disabled state) but add some lines to clobber that default gray style:

```dart
              style: FilledButton.styleFrom(
                // …
                // (remove `backgroundColor: _UnreadMarker.color` and its comment from here)
              ).copyWith(
                // clobber `FilledButton`'s default disabled state
                // TODO(#95) need dark-theme colors (foreground and background)
                foregroundColor: WidgetStateColor.resolveWith((_) => Colors.white),
                backgroundColor: WidgetStateColor.resolveWith((_) => _UnreadMarker.color),
              ),
```

lib/widgets/message_list.dart
```diff
@@ -503,10 +513,52 @@ class MarkAsReadWidget extends StatelessWidget {
...
+    return GestureDetector(
+      onTapDown: (_) => _setScale(true),
+      onTapUp: (_) => _setScale(false),
+      onTapCancel: () => _setScale(false),
+      child: AnimatedScale(
+        scale: _isPressed ? 0.95 : 1,
+        duration: const Duration(milliseconds: 100),
+        curve: Curves.easeOut,
+        child: AnimatedOpacity(
+          opacity: widget.hidden ? 0 : widget.loading ? 0.55 : 1,
```
chrisbobbe, line 558: This changes the appearance of the button depending on the loading state. Let's make that change in the `Handle loading state` commit, since otherwise the loading state isn't fully handled in that commit.

Also, is there a reason to choose 0.55 instead of 0.5, as Vlad mentioned in the [discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read/near/1813212)?

chrisbobbe: Also, let's separate the on-pressed animation and the fade-to-hidden animation into separate commits, too.

In the commit for the on-pressed animation, let's remove the existing press animation, which is Material Design's "ink splash" effect; see #417. That can be done by adding `splashFactory: NoSplash.splashFactory` to the `FilledButton.styleFrom` call.

chrisbobbe (commented):
Thanks, @Khader-1! Small comments below.

Ah also: let's use `msglist` for the prefix, like in previous commits, instead of `msg_list`.

Khader-1:
Thanks @chrisbobbe! I've made a new revision, PTAL.

test/widgets/message_list_test.dart
```diff
@@ -753,6 +753,96 @@ void main() {
...
+          check(state.loading).isFalse();
+
+          await tester.tap(find.byType(MarkAsReadWidget));
+          await tester.pump();
+          check(state.loading).isTrue();
+
+          await tester.idle();
+          await tester.pump();
+          check(state.loading).isFalse();
+        });
```
chrisbobbe, line 778: It feels like we should be able to test the user's experience of the loading state more directly. The `loading` field does power the user-visible signs that the button is loading, but it's also kind of an implementation detail. How about something like the following:

```diff
--- test/widgets/message_list_test.dart
+++ test/widgets/message_list_test.dart
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'dart:ui';
 
 import 'package:checks/checks.dart';
 import 'package:collection/collection.dart';
@@ -888,6 +889,20 @@ void main() {
           unreadMessageIds: [message.id]),
       ]);
 
+      checkAppearsLoading(WidgetTester tester, bool expected) {
+        final semantics = tester.firstWidget<Semantics>(find.descendant(
+          of: find.byType(MarkAsReadWidget),
+          matching: find.byType(Semantics)));
+        check(semantics.properties.enabled).equals(!expected);
+
+        final opacity = tester.widget<AnimatedOpacity>(find.descendant(
+          of: find.byType(MarkAsReadWidget),
+          matching: find.byType(AnimatedOpacity)));
+        check(opacity.opacity).equals(expected ? 0.5 : 1.0);
+      }
+
       group('MarkAsReadAnimation', () {
         testWidgets('loading is changed correctly', (WidgetTester tester) async {
           final narrow = TopicNarrow.ofMessage(message);
@@ -895,21 +910,21 @@ void main() {
             narrow: narrow, messages: [message], unreadMsgs: unreadMsgs);
           check(isMarkAsReadButtonVisible(tester)).isTrue();
 
-          connection.prepare(json: UpdateMessageFlagsForNarrowResult(
-            processedCount: 11, updatedCount: 3,
-            firstProcessedId: null, lastProcessedId: null,
-            foundOldest: true, foundNewest: true).toJson());
+          connection.prepare(
+            delay: const Duration(milliseconds: 2000),
+            json: UpdateMessageFlagsForNarrowResult(
+              processedCount: 11, updatedCount: 3,
+              firstProcessedId: null, lastProcessedId: null,
+              foundOldest: true, foundNewest: true).toJson());
 
-          final state = tester.state<MarkAsReadWidgetState>(find.byType(MarkAsReadWidget));
-          check(state.loading).isFalse();
+          checkAppearsLoading(tester, false);
 
           await tester.tap(find.byType(MarkAsReadWidget));
           await tester.pump();
-          check(state.loading).isTrue();
+          checkAppearsLoading(tester, true);
 
-          await tester.idle();
-          await tester.pump();
-          check(state.loading).isFalse();
+          await tester.pump(const Duration(milliseconds: 2000));
+          checkAppearsLoading(tester, false);
         });
 
         testWidgets('loading is changed correctly if request fails', (WidgetTester tester) async {
```

lib/widgets/message_list.dart
```diff
@@ -438,30 +438,41 @@ class ScrollToBottomButton extends StatelessWidget {
   }
 }
 
-class MarkAsReadWidget extends StatelessWidget {
+class MarkAsReadWidget extends StatefulWidget {
```
chrisbobbe, line 441:
```
msglist: Use scale animation for mark-as-read button i pressed state
```

When Vlad [said](https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read/near/1813212) "i pressed state", I think he probably meant "in pressed state". 😅

test/widgets/message_list_test.dart
```diff
@@ -753,6 +753,96 @@ void main() {
...
+          await tester.pump();
+
+          check(state.loading).isFalse();
+        });
+
+        testWidgets('in idle state', (WidgetTester tester) async {
+          final child = Container();
+
+          await tester.pumpWidget(MaterialApp(
+	            home: MarkAsReadAnimation(hidden: false, loading: false, child: child)));
```
chrisbobbe, line 809: nit: two-space indent, here and below

chrisbobbe (commented):
Thanks! Small comments below.

lib/widgets/message_list.dart
Khader-1: That explains everything!😅

test/widgets/message_list_test.dart
Khader-1: Thanks! I've created [a new NFC](https://github.com/zulip/zulip-flutter/pull/710/commits/55d3f861e48ee42b076e7c124ddcda22cf65d58a) for simulating delayed connections PTAL.

Khader-1: oh.. I see there is an implementation for that merged already! rebasing..

Khader-1:
Thanks @chrisbobbe! A new revision is here PTAL.

lib/widgets/message_list.dart
```diff
@@ -459,30 +459,41 @@ class ScrollToBottomButton extends StatelessWidget {
...
   const MarkAsReadWidget({super.key, required this.narrow});
 
   final Narrow narrow;
 
+  @override
+  State<MarkAsReadWidget> createState() => MarkAsReadWidgetState();
+}
+
+class MarkAsReadWidgetState extends State<MarkAsReadWidget> {
```
chrisbobbe, line 471:
```
msglist: Handle loading state in MarkAsReadWidget

MarkAsReadWidget needs to be disabled during loading state to prevent
multiple requests to the server.

Additionally it is needed for styling/animation convenience.
```

That last sentence ("Additionally…") doesn't really make sense; let's drop it. (What is more "convenient" about disabling the button?) I think it was from a previous revision when this commit was focused more on an implementation detail rather than the changes that users could see.

chrisbobbe (commented):
Thanks! One commit-message nit, and I'll mark this for Greg's review.

lib/widgets/message_list.dart
```diff
@@ -523,11 +532,58 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      _isPressed = isPressed;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      onTapDown: (_) => _setScale(true),
+      onTapUp: (_) => _setScale(false),
+      onTapCancel: () => _setScale(false),
```
gnprice, line 577: Hmm, I don't love that we're effectively winding up here with a button that has two separate gesture detectors watching for a press — this one tracking whether it's currently being pressed, and the one the Material library supplies which we're using for responding to having been pressed.

In particular this feels like it's repeating the work that the Material button is already doing, and we could get a similar effect here by tracking the button's WidgetState, looking for WidgetState.pressed.

That's not a blocker for merging this, but it's something I'll want to clean up later if we end up sticking with this UI design.

lib/widgets/message_list.dart
```diff
@@ -523,11 +532,58 @@ class MarkAsReadWidget extends StatelessWidget {
...
+  @override
+  State<MarkAsReadAnimation> createState() => _MarkAsReadAnimationState();
+}
+
+class _MarkAsReadAnimationState extends State<MarkAsReadAnimation> {
+  bool _isPressed = false;
+
+  void _setScale(bool isPressed) {
+    setState(() {
+      _isPressed = isPressed;
```
gnprice, line 568: nit: this method's name doesn't reflect what it does; it also makes it hard to guess what the meaning of an argument "true" vs. "false" is. A good name would be `_setIsPressed`.

gnprice (commented):
Thanks @Khader-1, and thanks @rajveermalviya and @chrisbobbe for the previous reviews! Comments below.

(And just so it doesn't get buried / so everything's mentioned in one place: there's one comment from Chris open above.)

lib/widgets/message_list.dart
```diff
@@ -523,11 +532,58 @@ class MarkAsReadWidget extends StatelessWidget {
                     height: (23 / 18))
                   .merge(weightVariableTextStyle(context, wght: 400))),
                 shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
+              ).copyWith(
+                // clobber `FilledButton`'s default disabled state
+                // TODO(#95) need dark-theme colors (foreground and background)
+                foregroundColor: WidgetStateColor.resolveWith((_) => Colors.white),
+                backgroundColor: WidgetStateColor.resolveWith((_) => _UnreadMarker.color),
```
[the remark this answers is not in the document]
chrisbobbe: Ah:

Wo the function you pass to [`WidgetStateColor.resolveWith`](https://api.flutter.dev/flutter/widgets/WidgetStateColor/resolveWith.html) receives a set of [`WidgetState`](https://api.flutter.dev/flutter/widgets/WidgetState.html)s, which are hovered, focused, pressed, etc., and disabled. Normally, you'd pass a function that checks that set of states and returns different colors based on it.

Here, we don't want different colors for the different states. We don't even want a different color for the disabled state; that's because we have other code for that. That other code uses an animation, which is especially helpful at the end of the loading state, when otherwise the button would jump to full opacity before the mark-as-read event arrives and the exit animation starts. And it doesn't seem possible to animate between disabled/enabled here using `ButtonStyle`.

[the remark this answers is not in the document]
chrisbobbe: (Oh and the other key piece of information: without passing these arguments, where we set the colors unconditionally, the widget would choose its own colors for the disabled state.)

lib/widgets/message_list.dart
Khader-1: That's right. Thanks!

lib/widgets/message_list.dart
Khader-1: I agree! I've tried various approaches before, but I haven't found a straightforward solution that works as a replacement.

I'd love to hear your suggestions and I'd love to implement them.

Khader-1:
Thanks @chrisbobbe and @gnprice for the reviews! I've pushed a new revision PTAL.

lib/widgets/message_list.dart
```diff
@@ -459,30 +459,41 @@ class ScrollToBottomButton extends StatelessWidget {
...
 
   final Narrow narrow;
 
+  @override
+  State<MarkAsReadWidget> createState() => _MarkAsReadWidgetState();
+}
+
+class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
+  @visibleForTesting
+  bool loading = false;
```
gnprice, line 473:
```suggestion
class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
  bool loading = false;
```

Pretty sure this annotation has no effect when the class itself is private.

test/widgets/message_list_test.dart
```diff
@@ -888,6 +888,106 @@ void main() {
...
+        });
+
+        testWidgets('in loading state', (WidgetTester tester) async {
+          final child = Container();
+
+          await tester.pumpWidget(MaterialApp(
+            home: MarkAsReadAnimation(hidden: false, loading: true, child: child)));
+
+          checkHasOpacityValue(tester, child, 0.5);
+        });
```
gnprice, line 970: Following up on https://github.com/zulip/zulip-flutter/pull/710#discussion_r1692348196 : as written, this is a test that doesn't give us a lot of value, because it's operating at the layer of an implementation detail of how the UI is organized and as a result it largely just repeats what's in the code of MarkAsReadAnimation.

In the context of this section on testing philosophy that I linked earlier:
these are "unit tests" of the kind that doesn't serve us well.

Instead, the tests should be written in a way that's more in terms of the UI as the user experiences it. That means that instead of these four test cases that use MarkAsReadAnimation directly, the relevant behavior should be tested in the style of the two test cases earlier in this group: use `setupMessageListPage` to create a message-list page where the button appears, then tap on the button and check how the button looks.

(Some of these test cases may be redundant with what those other two test cases already do, in which case they can just be left out.)

gnprice, line 970: Following up on https://github.com/zulip/zulip-flutter/pull/710#discussion_r1692348196 : as written, this is a test that doesn't give us a lot of value, because it's operating at the layer of an implementation detail of how the UI is organized and as a result it largely just repeats what's in the code of MarkAsReadAnimation.

In the context of this section on testing philosophy that I linked earlier:
these are "unit tests" of the kind that doesn't serve us well.

Instead, the tests should be written in a way that's more in terms of the UI as the user experiences it. That means that instead of these four test cases that use MarkAsReadAnimation directly, the relevant behavior should be tested in the style of the two test cases earlier in this group: use `setupMessageListPage` to create a message-list page where the button appears, then tap on the button and check how the button looks.

(Some of these test cases may be redundant with what those other two test cases already do, in which case they can just be left out.)

lib/widgets/message_list.dart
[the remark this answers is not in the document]
gnprice: Cool, thanks. @Khader-1 here's a way to explain that in the comment:
```suggestion
              ).copyWith(
                // Give the buttons a constant color regardless of whether their
                // state is disabled, pressed, etc.  We handle those states
                // separately, via MarkAsReadAnimation.
                // TODO(#95) need dark-theme colors (foreground and background)
                foregroundColor: WidgetStateColor.resolveWith((_) => Colors.white),
                backgroundColor: WidgetStateColor.resolveWith((_) => _UnreadMarker.color),
```

gnprice (commented):
Thanks for the revision! Just a few comments now.

test/widgets/message_list_test.dart
Khader-1: I

Khader-1:

gnprice:
Thanks for the revision! Merging.

I squashed this commit:
a24104ba9 msglist [nfc]: Remove unnecessary whitespace

into the one after it, since that one is already touching neighboring code and this two-line change didn't seem like it made that any harder to read.

Issue #424, opened 2023-11-29
Show muted streams as faded in streams list


gnprice:
This is a followup to #346. We have all the relevant data handy, so this is just a matter of applying the desired style.

@terpimost do you have a design for this already in Figma? I see the streams screen at https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=112%3A9610&mode=dev , but it doesn't look like any of those are shown as muted.

terpimost:
I don't have the design yet. but I suggest to make them 50-60% transparency:
<img width="414" alt="image" src="https://github.com/zulip/zulip-flutter/assets/1903309/dbb8e35e-0be5-4d55-9b2d-4d9a565940a6">
But we do want the background to be white, otherwise it would break the list

dhruvpatidar359:
may i work on this issue , if the design is decided  ?

gnprice:
Yeah, the draft design above is enough that we can go ahead and implement this. We can always adjust the design later.

@dhruvpatidar359 do take a look though at our README section on picking an issue to work on: https://github.com/zulip/zulip-flutter#picking-an-issue-to-work-on

PR #711, opened 2024-05-30
subscription_list: Show muted streams as faded in streams list

Khader-1:
Fixes: #424

| After | Before|
| -------- | ------- |
| image | image |

### DIFF
--- a/lib/widgets/subscription_list.dart
+++ b/lib/widgets/subscription_list.dart
@@ -209,6 +209,7 @@ class SubscriptionItem extends StatelessWidget {
   Widget build(BuildContext context) {
     final swatch = colorSwatchFor(context, subscription);
     final hasUnreads = (unreadCount > 0);
+    final opacity = subscription.isMuted ? 0.55 : 1.0;
     return Material(
       // TODO(#95) need dark-theme color
       color: Colors.white,
@@ -222,8 +223,10 @@ class SubscriptionItem extends StatelessWidget {
           const SizedBox(width: 16),
           Padding(
             padding: const EdgeInsets.symmetric(vertical: 11),
-            child: Icon(size: 18, color: swatch.iconOnPlainBackground,
-              iconDataForStream(subscription))),
+            child: Opacity(
+              opacity: opacity,
+              child: Icon(size: 18, color: swatch.iconOnPlainBackground,
+                iconDataForStream(subscription)))),
           const SizedBox(width: 5),
           Expanded(
             child: Padding(
@@ -231,21 +234,28 @@ class SubscriptionItem extends StatelessWidget {
               // TODO(design): unclear whether bold text is applied to all subscriptions
               //   or only those with unreads:
               //   https://github.com/zulip/zulip-flutter/pull/397#pullrequestreview-1742524205
-              child: Text(
-                style: const TextStyle(
-                  fontSize: 18,
-                  height: (20 / 18),
-                  // TODO(#95) need dark-theme color
-                  color: Color(0xFF262626),
-                ).merge(weightVariableTextStyle(context,
-                    wght: hasUnreads ? 600 : null)),
-                maxLines: 1,
-                overflow: TextOverflow.ellipsis,
-                subscription.name))),
+              child: Opacity(
+                opacity: opacity,
+                child: Text(
+                  style: const TextStyle(
+                    fontSize: 18,
+                    height: (20 / 18),
+                    // TODO(#95) need dark-theme color
+                    color: Color(0xFF262626),
+                  ).merge(weightVariableTextStyle(context,
+                      wght: hasUnreads ? 600 : null)),
+                  maxLines: 1,
+                  overflow: TextOverflow.ellipsis,
+                  subscription.name)))),
           if (unreadCount > 0) ...[
             const SizedBox(width: 12),
             // TODO(#747) show @-mention indicator when it applies
-            UnreadCountBadge(count: unreadCount, backgroundColor: swatch, bold: true),
+            Opacity(
+              opacity: opacity,
+              child: UnreadCountBadge(
+                count: unreadCount,
+                backgroundColor: swatch,
+                bold: true)),
           ],
           const SizedBox(width: 16),
         ])));
--- a/lib/widgets/subscription_list.dart
+++ b/lib/widgets/subscription_list.dart
@@ -1,4 +1,3 @@
-import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 
 import '../api/model/model.dart';
@@ -48,6 +47,15 @@ class _SubscriptionListPageState extends State<SubscriptionListPage> with PerAcc
     });
   }
 
+  void _sortSubs(List<Subscription> list) {
+    list.sort((a, b) {
+      if (a.isMuted && !b.isMuted) return 1;
+      if (!a.isMuted && b.isMuted) return -1;
+      // TODO(i18n): add locale-aware sorting
+      return a.name.toLowerCase().compareTo(b.name.toLowerCase());
+    });
+  }
+
   @override
   Widget build(BuildContext context) {
     // Design referenced from:
@@ -77,9 +85,8 @@ class _SubscriptionListPageState extends State<SubscriptionListPage> with PerAcc
         unpinned.add(subscription);
       }
     }
-    // TODO(i18n): add locale-aware sorting
-    pinned.sortBy((subscription) => subscription.name.toLowerCase());
-    unpinned.sortBy((subscription) => subscription.name.toLowerCase());
+    _sortSubs(pinned);
+    _sortSubs(unpinned);
 
     return Scaffold(
       appBar: AppBar(title: const Text("Channels")),
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -233,6 +233,16 @@ Subscription subscription(
   );
 }
 
+UserTopicItem userTopicItem(
+    ZulipStream stream, String topic, UserTopicVisibilityPolicy policy) {
+  return UserTopicItem(
+    streamId: stream.streamId,
+    topicName: topic,
+    lastUpdated: 1234567890,
+    visibilityPolicy: policy,
+  );
+}
+
 ////////////////////////////////////////////////////////////////
 // Messages, and pieces of messages.
 //
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -149,6 +149,17 @@ final User fourthUser  = user(fullName: 'Fourth User', email: 'fourth@example');
 // Streams and subscriptions.
 //
 
+/// A fresh stream ID, from a random but always strictly increasing sequence.
+int _nextStreamId() => (_lastStreamId += 1 + Random().nextInt(10));
+int _lastStreamId = 200;
+
+/// Construct an example stream.
+///
+/// If stream ID `streamId` is not given, it will be generated from a random
+/// but increasing sequence.
+/// Use an explicit `streamId` only if the ID needs to correspond to some
+/// other data in the test, or if the IDs need to increase in a different order
+/// from the calls to [stream].
 ZulipStream stream({
   int? streamId,
   String? name,
@@ -165,7 +176,7 @@ ZulipStream stream({
   int? streamWeeklyTraffic,
 }) {
   return ZulipStream(
-    streamId: streamId ?? 123, // TODO generate example IDs
+    streamId: streamId ?? _nextStreamId(),
     name: name ?? 'A stream', // TODO generate example names
     description: description ?? 'A description', // TODO generate example descriptions
     renderedDescription: renderedDescription ?? '<p>A description</p>', // TODO generate random
@@ -281,6 +292,8 @@ Map<String, dynamic> _messagePropertiesFromContent(String? content, String? cont
 int _nextMessageId() => (_lastMessageId += 1 + Random().nextInt(100));
 int _lastMessageId = 1000;
 
+const defaultStreamMessageStreamId = 123;
+
 /// Construct an example stream message.
 ///
 /// If the message ID `id` is not given, it will be generated from a random
@@ -288,6 +301,9 @@ int _lastMessageId = 1000;
 /// Use an explicit `id` only if the ID needs to correspond to some other data
 /// in the test, or if the IDs need to increase in a different order from the
 /// calls to [streamMessage] and [dmMessage].
+///
+/// The message will be in `stream` if given.  Otherwise,
+/// an example stream with ID `defaultStreamMessageStreamId` will be used.
 ///
 /// See also:
 ///  * [dmMessage], to construct an example direct message.
@@ -303,7 +319,7 @@ StreamMessage streamMessage({
   int? timestamp,
   List<MessageFlag>? flags,
 }) {
-  final effectiveStream = stream ?? _stream();
+  final effectiveStream = stream ?? _stream(streamId: defaultStreamMessageStreamId);
   // The use of JSON here is convenient in order to delegate parts of the data
   // to helper functions.  The main downside is that it loses static typing
   // of the properties as we're constructing the data.  That's probably OK
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -52,7 +52,7 @@ void main() {
     UnreadMessagesSnapshot? unreadMsgs,
   }) async {
     addTearDown(testBinding.reset);
-    streams ??= subscriptions ??= [eg.subscription(eg.stream())];
+    streams ??= subscriptions ??= [eg.subscription(eg.stream(streamId: eg.defaultStreamMessageStreamId))];
     await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot(
       streams: streams, subscriptions: subscriptions, unreadMsgs: unreadMsgs));
     store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
--- a/test/widgets/subscription_list_test.dart
+++ b/test/widgets/subscription_list_test.dart
@@ -132,6 +132,18 @@ void main() {
       ]);
       check(listedStreamIds(tester)).deepEquals([1, 2, 3, 4, 5, 6]);
     });
+
+    testWidgets('muted subscriptions come last among pinned streams and among unpinned streams', (tester) async {
+      await setupStreamListPage(tester, subscriptions: [
+        eg.subscription(eg.stream(streamId: 1, name: 'a'), isMuted: true, pinToTop: true),
+        eg.subscription(eg.stream(streamId: 2, name: 'b'), isMuted: false, pinToTop: true),
+        eg.subscription(eg.stream(streamId: 3, name: 'c'), isMuted: true, pinToTop: true),
+        eg.subscription(eg.stream(streamId: 4, name: 'd'), isMuted: false, pinToTop: false),
+        eg.subscription(eg.stream(streamId: 5, name: 'e'), isMuted: true, pinToTop: false),
+        eg.subscription(eg.stream(streamId: 6, name: 'f'), isMuted: false, pinToTop: false),
+      ]);
+      check(listedStreamIds(tester)).deepEquals([2, 1, 3, 4, 6, 5]);
+    });
   });
 
   testWidgets('unread badge shows with unreads', (tester) async {
--- a/test/widgets/subscription_list_test.dart
+++ b/test/widgets/subscription_list_test.dart
@@ -202,4 +202,37 @@ void main() {
     check(tester.widget<UnreadCountBadge>(find.byType(UnreadCountBadge)).backgroundColor)
       .equals(swatch);
   });
+
+  testWidgets('muted streams are displayed as faded', (tester) async {
+    void checkOpacityForStreamAndBadge(String streamName, int unreadCount, double opacity) {
+      final streamFinder = find.text(streamName);
+      final streamOpacity = tester.widget<Opacity>(
+        find.ancestor(of: streamFinder, matching: find.byType(Opacity)));
+      final badgeFinder = find.text('$unreadCount');
+      final badgeOpacity = tester.widget<Opacity>(
+        find.ancestor(of: badgeFinder, matching: find.byType(Opacity)));
+      check(streamOpacity.opacity).equals(opacity);
+      check(badgeOpacity.opacity).equals(opacity);
+    }
+
+    final stream1 = eg.stream(name: 'Stream 1');
+    final stream2 = eg.stream(name: 'Stream 2');
+    await setupStreamListPage(tester,
+      subscriptions: [
+        eg.subscription(stream1, isMuted: true),
+        eg.subscription(stream2, isMuted: false),
+      ],
+      userTopics: [
+        eg.userTopicItem(stream1, 'a', UserTopicVisibilityPolicy.unmuted),
+        eg.userTopicItem(stream2, 'b', UserTopicVisibilityPolicy.unmuted),
+      ],
+      unreadMsgs: eg.unreadMsgs(streams: [
+        UnreadStreamSnapshot(streamId: stream1.streamId, topic: 'a', unreadMessageIds: [1, 2]),
+        UnreadStreamSnapshot(streamId: stream2.streamId, topic: 'b', unreadMessageIds: [3]),
+      ]),
+    );
+
+    checkOpacityForStreamAndBadge('Stream 1', 2, 0.55);
+    checkOpacityForStreamAndBadge('Stream 2', 1, 1.0);
+  });
 }
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -41,7 +41,7 @@ void main() {
 
   /// Initialize [model] and the rest of the test state.
   Future<void> prepare({Narrow narrow = const CombinedFeedNarrow()}) async {
-    final stream = eg.stream();
+    final stream = eg.stream(streamId: eg.defaultStreamMessageStreamId);
     subscription = eg.subscription(stream);
     store = eg.store();
     await store.addStream(stream);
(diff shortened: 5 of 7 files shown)


test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +190,41 @@ void main() {
     check(tester.widget<UnreadCountBadge>(find.byType(UnreadCountBadge)).backgroundColor)
       .equals(swatch);
   });
+
+  testWidgets('muted streams are displayed as faded', (tester) async {
+    final stream = eg.stream(name: 'Channel 1');
```
sm-sayedi, line 195:
```suggestion
    final stream1 = eg.stream(streamId: 1, name: 'Channel 1');
```
(nit): Let's numerise this stream name as well as add an explicit `streamId` to make it match `stream2`.

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +190,41 @@ void main() {
...
+        UserTopicItem(
+          streamId: stream2.streamId,
+          topicName: 'b',
+          lastUpdated: 1234567890,
+          visibilityPolicy: UserTopicVisibilityPolicy.unmuted,
+        ),
+      ],
+      unreadMsgs: unreadMsgs);
+
+    final channel1Finder = find.text('Channel 1');
```
sm-sayedi, line 222:
```suggestion
    final stream1Finder = find.text(stream1.name);
```
I think it would be good to stick with the name "stream" instead of "channel" for this PR as all of the codebase still uses the "stream" terminology till this point in time. This may be changed to "channel" through #631.
Also using `stream1.name` is less error-prone and requires less refactoring if the name of the stream is changed.

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +190,41 @@ void main() {
...
+    final channel1Opacity = tester.widget<Opacity>(find.ancestor(of: channel1Finder, matching: find.byType(Opacity))).opacity;
```
sm-sayedi, line 223:
```suggestion
    final stream1Opacity = tester.widget<Opacity>(
      find.ancestor(of: stream1Finder, matching: find.byType(Opacity))).opacity;
```
(nit): The line is too long.

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +190,41 @@ void main() {
...
+    check(channel1Opacity).isLessThan(1.0);
```
sm-sayedi, line 224:
```suggestion
    check(stream1Opacity).equals(0.55);
```
Let's use the exact opacity value used in the code to make sure this test fails if there is an unintentional change to the opacity.

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +190,41 @@ void main() {
...
+
+    final channel2Finder = find.text('Channel 2');
+    final channel2Opacity = tester.widget<Opacity>(find.ancestor(of: channel2Finder, matching: find.byType(Opacity))).opacity;
```
sm-sayedi, line 227:
```suggestion
    final stream2Finder = find.text(stream2.name);
    final stream2Opacity = tester.widget<Opacity>(
      find.ancestor(of: stream2Finder, matching: find.byType(Opacity))).opacity;
```

sm-sayedi (commented):
Thanks @Khader-1. LGTM! Just a few small comments below.

Khader-1:
Thanks @sm-sayedi! I've pushed a revision PTAL.

sm-sayedi (approved):
Thanks for the revision, @Khader-1. Looks great. ~You can now move on the the mentor review.~

Upon reviewing #714, I feel the need for another small change in the code. See below.

lib/widgets/subscription_list.dart
```diff
@@ -210,36 +210,39 @@ class SubscriptionItem extends StatelessWidget {
...
-                overflow: TextOverflow.ellipsis,
-                subscription.name))),
-          if (unreadCount > 0) ...[
-            const SizedBox(width: 12),
-            // TODO(#384) show @-mention indicator when it applies
-            UnreadCountBadge(count: unreadCount, backgroundColor: swatch, bold: true),
-          ],
-          const SizedBox(width: 16),
-        ])));
+        child: Opacity(
```
sm-sayedi, line 213: Instead of `Opacity` wrapping the whole `Row`, let's make it wrap only the icon and the stream name. This will be helpful when there is a dot marker added through #714.

Khader-1 (commented):
Thanks for the review @sm-sayedi! A new revision is pushed PTAL.

sm-sayedi (approved):
Thanks for the revision @Khader-1! LGTM. Let's move on to the mentor review from @kenclary!

kenclary (approved):
looks good to me.

alya:
I think we should show muted channels below unmuted ones, as we do in the left sidebar in the web app, unless we specifically have reason to do otherwise. (I know that's not how it works in the RN app.)

alya:
Could you please update the screenshots in the PR description to include more muted streams, including ones both with and without unread messages?

Khader-1:
Thanks @alya! I agree. I've updated the sorting logic and the screenshot. 
When it comes to notifications, the current behavior is to show nothing if the stream is muted, but this will be replaced with #714

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,43 @@ void main() {
     check(tester.widget<UnreadCountBadge>(find.byType(UnreadCountBadge)).backgroundColor)
       .equals(swatch);
   });
+
+  testWidgets('muted streams are displayed as faded', (tester) async {
+    final stream1 = eg.stream(name: 'Stream 1');
+    final stream2 = eg.stream(streamId: 2, name: 'Stream 2');
```
chrisbobbe, line 208: Why pass `streamId` when making `stream2` but not `stream1`? Does this test need a specific stream ID for `stream2` but not `stream1`?

…Ah, I think I see: this test doesn't particularly care about the specific stream-ID values, but it does need them to be different from each other. That's accomplished by letting `stream1` be the default value 123, and giving `stream2` some different value.

Slightly nicer would be to just pass `streamId: 1` and `streamId: 2`. That would remove this surprising asymmetry that I noticed.

Better than that, though, would be to make a prep commit that takes care of this TODO in `eg.stream`:

```dart
    streamId: streamId ?? 123, // TODO generate example IDs
```

and then not pass `streamId` when making `stream1` or `stream2`, because they'll be automatically generated and different from each other. Like we did for `eg.user` in bb78e6ce, and `eg.streamMessage`/`eg.dmMessage` in 13e54f60b.

chrisbobbe (changes requested):
Thanks @Khader-1! Small comments below.

Then also, please separate the main commit into two commits, for the two user-facing changes we're making:
- adding new sorting logic
- applying opacity

For screenshots that include a muted channel with unreads, this should be possible if the muted channel has an unmuted topic with unreads. I'd like to see the unread-counter marker in this case, to compare how it looks with and without being faded.

chrisbobbe:
@Khader-1 I've noticed you pushed some updates to the branch. Is it ready for another review?

Khader-1:
Not yet... I was actually testing the updated version on Zulip's cloud-dev VM.
Anyways, I think it will be ready today, will ping here when that's the case.

Khader-1 (commented):
Thanks for the review @chrisbobbe! I believe this ready now.

test/example_data.dart
```diff
@@ -163,7 +174,7 @@ ZulipStream stream({
   int? streamWeeklyTraffic,
 }) {
   return ZulipStream(
-    streamId: streamId ?? 123, // TODO generate example IDs
+    streamId: streamId ?? _nextStreamId(), // TODO generate example IDs
```
chrisbobbe, line 177: We can remove the TODO now :)

test/widgets/subscription_list_test.dart
```diff
@@ -132,6 +132,18 @@ void main() {
       ]);
       check(listedStreamIds(tester)).deepEquals([1, 2, 3, 4, 5, 6]);
     });
+
+    testWidgets('muted subscriptions come at last', (tester) async {
```
chrisbobbe, line 136: At first I had trouble connecting this test description to the test's behavior: there are three muted subscriptions, for stream IDs 1, 3, and 5, but those IDs don't all come last in the list of expected stream IDs.

I think we mean something like: 'muted subscriptions come last among pinned streams and among unpinned streams'. There might be a more concise way to say that.

chrisbobbe (commented):
Thanks! And thanks for posting updated screenshots. I notice that the unread marker isn't faded, like it is in Vlad's draft design in the issue: https://github.com/zulip/zulip-flutter/issues/424#issuecomment-1831121034

Let's make it so the unread marker is also faded.

chrisbobbe:
I see you've pushed some changes; please comment here when this is ready for another review.

Khader-1:
This is now ready; I had to update the Flutter SDK, which unfortunately took quite some time.

lib/widgets/subscription_list.dart
```diff
@@ -213,32 +220,42 @@ class SubscriptionItem extends StatelessWidget {
...
           if (unreadCount > 0) ...[
             const SizedBox(width: 12),
             // TODO(#747) show @-mention indicator when it applies
-            UnreadCountBadge(count: unreadCount, backgroundColor: swatch, bold: true),
+            Opacity(
+              opacity: subscription.isMuted ? 0.55 : 1,
+              child: UnreadCountBadge(
+                count: unreadCount,
+                backgroundColor: swatch,
+                bold: true)),
```
chrisbobbe, line 258: Can we simplify by putting a single `Opacity` higher in the tree, instead of two different `Opacity`s?

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,47 @@ void main() {
     check(tester.widget<UnreadCountBadge>(find.byType(UnreadCountBadge)).backgroundColor)
       .equals(swatch);
   });
+
+  testWidgets('muted streams are displayed as faded', (tester) async {
+
+    testOpacityForStreamAndBadge(String streamName, double opacity) {
```
chrisbobbe, line 208: nit: remove blank line

chrisbobbe (commented):
Thanks @Khader-1! Small comments below, and I'll go ahead and mark for Greg's review.

test/example_data.dart
```diff
@@ -147,6 +147,17 @@ final User fourthUser  = user(fullName: 'Fourth User', email: 'fourth@example');
 // Streams and subscriptions.
 //
 
+/// A fresh stream ID, from a random but always strictly increasing sequence.
+int _nextStreamId() => (_lastStreamId += 1 + Random().nextInt(100));
+int _lastStreamId = 1000;
```
gnprice, line 152: nit:
```suggestion
int _nextStreamId() => (_lastStreamId += 1 + Random().nextInt(10));
int _lastStreamId = 100;
```

That way they're less likely to look the same as a random user ID or message ID; and they're smaller than those, which is realistic.

test/widgets/message_list_test.dart
```diff
@@ -52,7 +52,7 @@ void main() {
     UnreadMessagesSnapshot? unreadMsgs,
   }) async {
     addTearDown(testBinding.reset);
-    streams ??= subscriptions ??= [eg.subscription(eg.stream())];
+    streams ??= subscriptions ??= [eg.subscription(eg.stream(streamId: 123))];
```
gnprice, line 55: So it was kind of a latent bug already, or at least a wart, that these tests depended on this constant value. I think concretely they're depending on `eg.streamMessage` producing messages in the same stream that `eg.stream()` gives.

But I think if we make this a constant that this test file supplies, that actually makes that situation somewhat worse. It's now quite obscure where this value "123" is coming from or what it connects to — why it needs to be this instead of 234 or 1 or 1000. In the version in main, you can click through to `eg.stream` and see the 123 there, and at least what the test is depending on is only that you get that same value (from that same one line of code) repeatedly, rather than that you get 123 here and there's some other piece of code somewhere else (namely in `eg.streamMessage`) that also says 123 and the important thing is that this line matches that one.

A minimal fix for that problem would be for there to be some named value in `example_data.dart`, and then all the places where you're adding `streamId: 123` here can refer to that value by name. That recovers the property that it's clear what other pieces of code any given piece of code is counting on agreeing with about the stream ID.

lib/widgets/subscription_list.dart
```diff
@@ -213,32 +220,42 @@ class SubscriptionItem extends StatelessWidget {
...
-                  color: Color(0xFF262626),
-                ).merge(weightVariableTextStyle(context,
-                    wght: hasUnreads ? 600 : null)),
-                maxLines: 1,
-                overflow: TextOverflow.ellipsis,
-                subscription.name))),
+            child: Opacity(
+              opacity: subscription.isMuted ? 0.55 : 1,
+              child: Row(
+                children: [
```
gnprice, line 227: nit:
```suggestion
              child: Row(children: [
```

Then that saves a level of indentation for all the children.

lib/widgets/subscription_list.dart
```diff
@@ -213,32 +220,42 @@ class SubscriptionItem extends StatelessWidget {
...
+                            wght: hasUnreads ? 600 : null)),
+                        maxLines: 1,
+                        overflow: TextOverflow.ellipsis,
+                        subscription.name)))]))),
           if (unreadCount > 0) ...[
             const SizedBox(width: 12),
             // TODO(#747) show @-mention indicator when it applies
-            UnreadCountBadge(count: unreadCount, backgroundColor: swatch, bold: true),
+            Opacity(
+              opacity: subscription.isMuted ? 0.55 : 1,
```
gnprice, line 254: Can this have just one Opacity around the whole thing? Or is there something I'm missing that isn't getting the Opacity applied?

If so, then that may also enable going back to a single Row instead of a pair of them nested.

lib/widgets/subscription_list.dart
```diff
@@ -213,32 +220,42 @@ class SubscriptionItem extends StatelessWidget {
...
+                        style: const TextStyle(
+                          fontSize: 18,
+                          height: (20 / 18),
+                          // TODO(#95) need dark-theme color
+                          color: Color(0xFF262626),
+                        ).merge(weightVariableTextStyle(context,
+                            wght: hasUnreads ? 600 : null)),
+                        maxLines: 1,
+                        overflow: TextOverflow.ellipsis,
+                        subscription.name)))]))),
```
gnprice, line 249: nit: when there's a children list and the open-bracket is followed by a newline, the close-bracket starts a line:
```suggestion
                        subscription.name))),
                      ]))),
```

(except I wrote that here in the GitHub comment UI so didn't count the indentation level — should match appropriately)

test/widgets/subscription_list_test.dart
gnprice, line 208: nit: function definitions should have explicit return type — otherwise they look a lot like function calls

gnprice, line 208: nit: names starting with "test" are reserved for helpers that substitute for `test`, so that create a new test case (by calling `test`, directly or indirectly).

For a helper like this, say "check".

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,47 @@ void main() {
...
+        UserTopicItem(
+          streamId: stream2.streamId,
+          topicName: 'b',
+          lastUpdated: 1234567890,
+          visibilityPolicy: UserTopicVisibilityPolicy.unmuted,
+        ),
+      ], unreadMsgs: unreadMsgs);
+
+      testOpacityForStreamAndBadge('Stream 1', 0.55);
+      testOpacityForStreamAndBadge('Stream 2', 1.0);
```
gnprice, line 246: nit: overindented

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,47 @@ void main() {
...
+
+  testWidgets('muted streams are displayed as faded', (tester) async {
+
+    testOpacityForStreamAndBadge(String streamName, double opacity) {
+      final streamFinder = find.text(streamName);
+      final streamOpacity = tester.widget<Opacity>(
+        find.ancestor(of: streamFinder, matching: find.byType(Opacity))).opacity;
+      final badgeOpacity = tester.widget<Opacity>(find.ancestor(of: find.descendant(
+        of: find.ancestor(of: streamFinder, matching: find.byType(Row)),
+        matching: find.byType(UnreadCountBadge)), matching: find.byType(Opacity))).opacity;
```
gnprice, line 214: This is a bit hard to parse visually. It's worth spending a few more lines in order to make the structure more regular:
```suggestion
      final badgeOpacity = tester.widget<Opacity>(find.ancestor(
        of: find.descendant(
          of: find.ancestor(
            of: streamFinder,
            matching: find.byType(Row)),
          matching: find.byType(UnreadCountBadge)),
        matching: find.byType(Opacity)));
```

Or better:
```suggestion
      final badgeOpacity = tester.widget<Opacity>(
        find.ancestor(matching: find.byType(Opacity),
          of: find.descendant(matching: find.byType(UnreadCountBadge),
            of: find.ancestor(matching: find.byType(Row),
              of: streamFinder))));
```

gnprice: (Oh and then `check(badgeOpacity.opacity).equals(opacity)` below.)

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,47 @@ void main() {
...
+      subscriptions: [
+        eg.subscription(stream1, isMuted: true),
+        eg.subscription(stream2, isMuted: false)
+      ],
+      userTopics: [
+        UserTopicItem(
+          streamId: stream1.streamId,
+          topicName: 'a',
+          lastUpdated: 1234567890,
+          visibilityPolicy: UserTopicVisibilityPolicy.unmuted,
```
gnprice, line 235: This can be made a bit more compact (and with less noise from boring parts like `lastUpdated`) with an appropriate helper.

There's one already in `model/stream_test.dart`. So that can be moved to `example_data.dart` in a prep commit.

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,47 @@ void main() {
...
+        ),
+        UserTopicItem(
+          streamId: stream2.streamId,
+          topicName: 'b',
+          lastUpdated: 1234567890,
+          visibilityPolicy: UserTopicVisibilityPolicy.unmuted,
+        ),
+      ], unreadMsgs: unreadMsgs);
+
```
gnprice, line 244:
```suggestion
      ],
      unreadMsgs: unreadMsgs,
    );
```

The `unreadMsgs` parameter is a peer of `userTopics` and `subscriptions` — all of them are in the same position as different pieces of the data. So best to format them as peers.

gnprice: Separately: `unreadMsgs` can be inlined here.

test/widgets/subscription_list_test.dart
gnprice: But also: this ancestor/descendant/ancestor dance feels like it's getting into the implementation details of how this UI is laid out.

How about giving these streams different numbers of unreads? And then this can just be finding the Opacity ancestors of `find.text("2")` and `find.text("1")` respectively.

gnprice (commented):
Thanks @Khader-1, and thanks @sm-sayedi and @chrisbobbe for the previous reviews! Comments below.

lib/widgets/subscription_list.dart
Khader-1: It could be done as such and it was actually but we had to change it as it will eventually be changed when working on #714; Check [Sayed's comment](https://github.com/zulip/zulip-flutter/pull/711#discussion_r1622478616)

lib/widgets/subscription_list.dart
Khader-1: It could be done as such and it was actually but we had to change it as it will eventually be changed when working on https://github.com/zulip/zulip-flutter/pull/714; Check https://github.com/zulip/zulip-flutter/pull/711#discussion_r1622478616

test/example_data.dart
```diff
@@ -220,6 +233,16 @@ Subscription subscription(
   );
 }
 
+UserTopicItem userTopicItem(
```
gnprice, line 236: test: Factor out `userTopicItem` helper

Is this commit intended to be NFC?
https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc

If so, it's helpful to the reader to mark it as such.

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,47 @@ void main() {
...
+      check(badgeOpacity.opacity).equals(opacity);
+    }
+
+    final stream1 = eg.stream(name: 'Stream 1');
+    final stream2 = eg.stream(name: 'Stream 2');
+    await setupStreamListPage(tester,
+      subscriptions: [
+        eg.subscription(stream1, isMuted: true),
+        eg.subscription(stream2, isMuted: false)
+      ],
```
gnprice, line 224: nit:
```suggestion
        eg.subscription(stream2, isMuted: false),
      ],
```

test/example_data.dart
```diff
@@ -147,6 +147,19 @@ final User fourthUser  = user(fullName: 'Fourth User', email: 'fourth@example');
 // Streams and subscriptions.
 //
 
+/// A fresh stream ID, from a random but always strictly increasing sequence.
+int _nextStreamId() => (_lastStreamId += 1 + Random().nextInt(10));
+int _lastStreamId = 100;
+
+const staticStreamId = 123;
```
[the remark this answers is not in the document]
gnprice: Or I guess that name doesn't capture it either — the meaning is actually a little complicated, in that this is the default for `eg.streamMessage` but not for `eg.stream` (which defaults to the random sequence).

Let's make it:
```suggestion
const defaultStreamMessageStreamId = 123;
```

and also put the definition right above `streamMessage`.

That name is kind of wordy, but that may actually be a good thing here — the way this works is a bit surprising and we should probably reorganize so that tests aren't referring to a shared constant stream ID like this.

The `streamMessage` doc should also mention that there's this default.

test/widgets/subscription_list_test.dart
gnprice: bump — you added that useful prep commit but then didn't actually use the new shared helper here :slightly_smiling_face:

lib/widgets/subscription_list.dart
gnprice: I see.

In that case let's split the Opacity up further, so that it applies separately to the icon and name. That will save the extra layer of Expanded and Row, which I think makes this code a bit easier to read and understand.

gnprice (commented):
Thanks for the revision! This is close to merge now. Small comments below.

Khader-1:
Thanks Greg! Sorry about the lack of focus.. I believe This is ready now.

test/widgets/subscription_list_test.dart
gnprice: bump :slightly_smiling_face:

test/widgets/subscription_list_test.dart
```diff
@@ -190,4 +202,37 @@ void main() {
...
+      userTopics: [
+        eg.userTopicItem(stream1, 'a', UserTopicVisibilityPolicy.unmuted),
+        eg.userTopicItem(stream2, 'b', UserTopicVisibilityPolicy.unmuted),
+      ],
+      unreadMsgs: eg.unreadMsgs(streams: [
+        UnreadStreamSnapshot(streamId: stream1.streamId, topic: 'a', unreadMessageIds: [1, 2]),
+        UnreadStreamSnapshot(streamId: stream2.streamId, topic: 'b', unreadMessageIds: [3])
```
gnprice, line 231: nit: similarly, needs trailing comma (it got deleted in this revision)

lib/widgets/subscription_list.dart
```diff
@@ -215,30 +223,42 @@ class SubscriptionItem extends StatelessWidget {
...
           Padding(
             padding: const EdgeInsets.symmetric(vertical: 11),
-            child: Icon(size: 18, color: swatch.iconOnPlainBackground,
-              iconDataForStream(subscription))),
+            child: Opacity(
+              opacity: opacity,
+              child: Icon(size: 18, color: swatch.iconOnPlainBackground,
+                iconDataForStream(subscription)),
+            ),
+          ),
```
gnprice, line 231: nit: follow the existing style for formatting these close-parens
```suggestion
                iconDataForStream(subscription)))),
```

lib/widgets/subscription_list.dart
```diff
@@ -215,30 +223,42 @@ class SubscriptionItem extends StatelessWidget {
...
+                    fontSize: 18,
+                    height: (20 / 18),
+                    // TODO(#95) need dark-theme color
+                    color: Color(0xFF262626),
+                  ).merge(weightVariableTextStyle(context,
+                      wght: hasUnreads ? 600 : null)),
+                  maxLines: 1,
+                  overflow: TextOverflow.ellipsis,
+                  subscription.name),
+              ))),
```
gnprice, line 252: nit: similarly here

test/example_data.dart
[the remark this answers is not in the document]
gnprice: bump on these two aspects:

> and also put the definition right above `streamMessage`.

> The `streamMessage` doc should also mention that there's this default.

gnprice:
Thanks @Khader-1 for the revision! Just a couple of small things still open, and a few nits new in this revision; details above.

Khader-1:
Thanks @gnprice for the review! I've made a new revision PTAL.

test/example_data.dart
```diff
@@ -278,6 +299,7 @@ Map<String, dynamic> _messagePropertiesFromContent(String? content, String? cont
 /// A fresh message ID, from a random but always strictly increasing sequence.
 int _nextMessageId() => (_lastMessageId += 1 + Random().nextInt(100));
 int _lastMessageId = 1000;
+const defaultStreamMessageStreamId = 123;
```
gnprice, line 302: nit:
```suggestion
int _nextMessageId() => (_lastMessageId += 1 + Random().nextInt(100));
int _lastMessageId = 1000;

const defaultStreamMessageStreamId = 123;
```

The first two items are closely related to each other, but not to the third. So the visual grouping can reflect that.

gnprice (commented):
Thanks @Khader-1 for all your work on this! Looks good — merging.

I fixed the one nit below, and also tweaked the `eg.streamMessage` doc slightly:
```diff
-/// If the message stream `stream` is not given, an example stream with
-/// `defaultStreamMessageStreamId` will be used.
+/// The message will be in `stream` if given.  Otherwise,
+/// an example stream with ID `defaultStreamMessageStreamId` will be used.
```

Issue #673, opened 2024-05-14
Copy link to message


gnprice:
We should make it easy to get a link to a particular message you're looking at.

This is particularly useful for pointing to the message in a message you send in a different conversation, or in something you write in some other system outside of Zulip.

In the legacy zulip-mobile app we don't have this, and instead have a "Share" option that isn't super useful. We've talked there occasionally for years about replacing it with "copy link":
* https://github.com/zulip/zulip-mobile/issues/2623

but never quite put it on the agenda to actually do. Here we should just do it.

For an example for the UI and copying to the clipboard, see the existing "Copy message text" item in the message action sheet. For constructing the URL for the message, see the implementation of quote-and-reply.

PR #713, opened 2024-05-30
msglist: Add "Copy message link" button to action sheet

sm-sayedi:
A new option is added to the long press action sheet of a message to copy the message link to the clipboard.

Here's an illustration:
<table>
  <tr>
    <td><b>Photo</b></td>
    <td><b>Video</b></td>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

Fixes: #673
Fixes: #732

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -43,8 +43,8 @@
   "@permissionsDeniedReadExternalStorage": {
     "description": "Message for dialog asking the user to grant permissions for external storage read access."
   },
-  "actionSheetOptionCopy": "Copy message text",
-  "@actionSheetOptionCopy":  {
+  "actionSheetOptionCopyMessageText": "Copy message text",
+  "@actionSheetOptionCopyMessageText":  {
     "description": "Label for copy message text button on action sheet."
   },
   "actionSheetOptionShare": "Share",
@@ -168,8 +168,8 @@
   "@successLinkCopied": {
     "description": "Success message after copy link action completed."
   },
-  "successMessageCopied": "Message Copied",
-  "@successMessageCopied": {
+  "successMessageTextCopied": "Message text copied",
+  "@successMessageTextCopied": {
     "description": "Message when content of a message was copied to the user's system clipboard."
   },
   "composeBoxAttachFilesTooltip": "Attach files",
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -47,6 +47,10 @@
   "@actionSheetOptionCopyMessageText":  {
     "description": "Label for copy message text button on action sheet."
   },
+  "actionSheetOptionCopyMessageLink": "Copy link to message",
+  "@actionSheetOptionCopyMessageLink": {
+    "description": "Label for copy message link button on action sheet."
+  },
   "actionSheetOptionShare": "Share",
   "@actionSheetOptionShare":  {
     "description": "Label for share button on action sheet."
@@ -171,6 +175,10 @@
   "successMessageTextCopied": "Message text copied",
   "@successMessageTextCopied": {
     "description": "Message when content of a message was copied to the user's system clipboard."
+  },
+  "successMessageLinkCopied": "Message link copied",
+  "@successMessageLinkCopied": {
+    "description": "Message when link of a message was copied to the user's system clipboard."
   },
   "composeBoxAttachFilesTooltip": "Attach files",
   "@composeBoxAttachFilesTooltip": {
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -6,6 +6,8 @@ import 'package:share_plus/share_plus.dart';
 import '../api/exception.dart';
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
+import '../model/internal_link.dart';
+import '../model/narrow.dart';
 import 'clipboard.dart';
 import 'compose_box.dart';
 import 'dialog.dart';
@@ -44,6 +46,7 @@ void showMessageActionSheet({required BuildContext context, required Message mes
           messageListContext: context,
         ),
         CopyMessageTextButton(message: message, messageListContext: context),
+        CopyMessageLinkButton(message: message, messageListContext: context),
         ShareButton(message: message, messageListContext: context),
       ]);
     });
@@ -307,6 +310,37 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
     copyWithPopup(context: messageListContext,
       successContent: Text(zulipLocalizations.successMessageTextCopied),
       data: ClipboardData(text: rawContent));
+  }
+}
+
+class CopyMessageLinkButton extends MessageActionSheetMenuItemButton {
+  CopyMessageLinkButton({
+    super.key,
+    required super.message,
+    required super.messageListContext,
+  });
+
+  @override IconData get icon => Icons.link;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionCopyMessageLink;
+  }
+
+  @override void onPressed(BuildContext context) {
+    Navigator.of(context).pop();
+    final zulipLocalizations = ZulipLocalizations.of(messageListContext);
+
+    final store = PerAccountStoreWidget.of(messageListContext);
+    final messageLink = narrowLink(
+      store,
+      SendableNarrow.ofMessage(message, selfUserId: store.selfUserId),
+      nearMessageId: message.id,
+    );
+
+    copyWithPopup(context: messageListContext,
+      successContent: Text(zulipLocalizations.successMessageLinkCopied),
+      data: ClipboardData(text: messageLink.toString()));
   }
 }
 
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -39,12 +39,12 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       return Column(children: [
         if (!hasThumbsUpReactionVote) AddThumbsUpButton(message: message, messageListContext: context),
         StarButton(message: message, messageListContext: context),
-        ShareButton(message: message, messageListContext: context),
         if (isComposeBoxOffered) QuoteAndReplyButton(
           message: message,
           messageListContext: context,
         ),
         CopyMessageTextButton(message: message, messageListContext: context),
+        ShareButton(message: message, messageListContext: context),
       ]);
     });
 }
@@ -160,63 +160,6 @@ class StarButton extends MessageActionSheetMenuItemButton {
           UpdateMessageFlagsOp.remove => zulipLocalizations.errorUnstarMessageFailedTitle,
           UpdateMessageFlagsOp.add    => zulipLocalizations.errorStarMessageFailedTitle,
         }, message: errorMessage);
-    }
-  }
-}
-
-class ShareButton extends MessageActionSheetMenuItemButton {
-  ShareButton({
-    super.key,
-    required super.message,
-    required super.messageListContext,
-  });
-
-  @override IconData get icon => Icons.adaptive.share;
-
-  @override
-  String label(ZulipLocalizations zulipLocalizations) {
-    return zulipLocalizations.actionSheetOptionShare;
-  }
-
-  @override void onPressed(BuildContext context) async {
-    // Close the message action sheet; we're about to show the share
-    // sheet. (We could do this after the sharing Future settles
-    // with [ShareResultStatus.success], but on iOS I get impatient with
-    // how slowly our action sheet dismisses in that case.)
-    // TODO(#24): Fix iOS bug where this call causes the keyboard to
-    //   reopen (if it was open at the time of this
-    //   `showMessageActionSheet` call) and cover a large part of the
-    //   share sheet.
-    Navigator.of(context).pop();
-    final zulipLocalizations = ZulipLocalizations.of(messageListContext);
-
-    final rawContent = await fetchRawContentWithFeedback(
-      context: messageListContext,
-      messageId: message.id,
-      errorDialogTitle: zulipLocalizations.errorSharingFailed,
-    );
-
-    if (rawContent == null) return;
-
-    if (!messageListContext.mounted) return;
-
-    // TODO: to support iPads, we're asked to give a
-    //   `sharePositionOrigin` param, or risk crashing / hanging:
-    //     https://pub.dev/packages/share_plus#ipad
-    //   Perhaps a wart in the API; discussion:
-    //     https://github.com/zulip/zulip-flutter/pull/12#discussion_r1130146231
-    final result = await Share.share(rawContent);
-
-    switch (result.status) {
-      // The plugin isn't very helpful: "The status can not be determined".
-      // Until we learn otherwise, assume something wrong happened.
-      case ShareResultStatus.unavailable:
-        if (!messageListContext.mounted) return;
-        await showErrorDialog(context: messageListContext,
-          title: zulipLocalizations.errorSharingFailed);
-      case ShareResultStatus.success:
-      case ShareResultStatus.dismissed:
-        // nothing to do
     }
   }
 }
@@ -364,5 +307,62 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
     copyWithPopup(context: messageListContext,
       successContent: Text(zulipLocalizations.successMessageTextCopied),
       data: ClipboardData(text: rawContent));
+  }
+}
+
+class ShareButton extends MessageActionSheetMenuItemButton {
+  ShareButton({
+    super.key,
+    required super.message,
+    required super.messageListContext,
+  });
+
+  @override IconData get icon => Icons.adaptive.share;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionShare;
+  }
+
+  @override void onPressed(BuildContext context) async {
+    // Close the message action sheet; we're about to show the share
+    // sheet. (We could do this after the sharing Future settles
+    // with [ShareResultStatus.success], but on iOS I get impatient with
+    // how slowly our action sheet dismisses in that case.)
+    // TODO(#24): Fix iOS bug where this call causes the keyboard to
+    //   reopen (if it was open at the time of this
+    //   `showMessageActionSheet` call) and cover a large part of the
+    //   share sheet.
+    Navigator.of(context).pop();
+    final zulipLocalizations = ZulipLocalizations.of(messageListContext);
+
+    final rawContent = await fetchRawContentWithFeedback(
+      context: messageListContext,
+      messageId: message.id,
+      errorDialogTitle: zulipLocalizations.errorSharingFailed,
+    );
+
+    if (rawContent == null) return;
+
+    if (!messageListContext.mounted) return;
+
+    // TODO: to support iPads, we're asked to give a
+    //   `sharePositionOrigin` param, or risk crashing / hanging:
+    //     https://pub.dev/packages/share_plus#ipad
+    //   Perhaps a wart in the API; discussion:
+    //     https://github.com/zulip/zulip-flutter/pull/12#discussion_r1130146231
+    final result = await Share.share(rawContent);
+
+    switch (result.status) {
+      // The plugin isn't very helpful: "The status can not be determined".
+      // Until we learn otherwise, assume something wrong happened.
+      case ShareResultStatus.unavailable:
+        if (!messageListContext.mounted) return;
+        await showErrorDialog(context: messageListContext,
+          title: zulipLocalizations.errorSharingFailed);
+      case ShareResultStatus.success:
+      case ShareResultStatus.dismissed:
+        // nothing to do
+    }
   }
 }
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -361,7 +361,7 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
 
     if (!messageListContext.mounted) return;
 
-    copyWithPopup(context: context,
+    copyWithPopup(context: messageListContext,
       successContent: Text(zulipLocalizations.successMessageTextCopied),
       data: ClipboardData(text: rawContent));
   }
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -44,7 +44,7 @@ void showMessageActionSheet({required BuildContext context, required Message mes
           message: message,
           messageListContext: context,
         ),
-        CopyButton(message: message, messageListContext: context),
+        CopyMessageTextButton(message: message, messageListContext: context),
       ]);
     });
 }
@@ -330,8 +330,8 @@ class QuoteAndReplyButton extends MessageActionSheetMenuItemButton {
   }
 }
 
-class CopyButton extends MessageActionSheetMenuItemButton {
-  CopyButton({
+class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
+  CopyMessageTextButton({
     super.key,
     required super.message,
     required super.messageListContext,
@@ -341,7 +341,7 @@ class CopyButton extends MessageActionSheetMenuItemButton {
 
   @override
   String label(ZulipLocalizations zulipLocalizations) {
-    return zulipLocalizations.actionSheetOptionCopy;
+    return zulipLocalizations.actionSheetOptionCopyMessageText;
   }
 
   @override void onPressed(BuildContext context) async {
@@ -362,7 +362,7 @@ class CopyButton extends MessageActionSheetMenuItemButton {
     if (!messageListContext.mounted) return;
 
     copyWithPopup(context: context,
-      successContent: Text(zulipLocalizations.successMessageCopied),
+      successContent: Text(zulipLocalizations.successMessageTextCopied),
       data: ClipboardData(text: rawContent));
   }
 }
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -449,7 +449,7 @@ void main() {
     });
   });
 
-  group('CopyButton', () {
+  group('CopyMessageTextButton', () {
     setUp(() async {
       TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger.setMockMethodCallHandler(
         SystemChannels.platform,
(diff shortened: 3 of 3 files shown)


test/widgets/action_sheet_test.dart
```diff
@@ -490,4 +491,32 @@ void main() {
       check(await Clipboard.getData('text/plain')).isNull();
     });
   });
+
+  group('CopyMessageLinkButton', () {
+    setUp(() async {
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger.setMockMethodCallHandler(
```
Khader-1, line 497: nit: I'd rather not add the leading whitespace
```suggestion
TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger.setMockMethodCallHandler(
```

test/widgets/action_sheet_test.dart
```diff
@@ -490,4 +491,32 @@ void main() {
...
+      await tester.tap(find.byIcon(Icons.link));
+      await tester.pump(); // [MenuItemButton.onPressed] called in a post-frame callback: flutter/flutter@e4a39fa2e
+    }
+
+    testWidgets('copies message link to clipboard', (tester) async {
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
```
Khader-1, line 514: nit: Same here, I think both blocks are handling the arrangement steps for the test case so that extra white space feels as an extra
```suggestion
      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
```

Khader-1 (commented):
Thank you! Everything looks good, except for a couple of nits in the comments below.

sm-sayedi:
Thanks @Khader-1 for the review. Revision pushed. PTAL.

Khader-1 (approved):
Thanks! LGTM.

alya:
Please be sure to add a still screenshot of visual changes alongside any videos. It is very difficult to review some things on a video (like how the options in the menu look all together, in this case).

sm-sayedi:
> Please be sure to add a still screenshot of visual changes alongside any videos. It is very difficult to review some things on a video (like how the options in the menu look all together, in this case).

@alya An image has been added. Please have a look. Next time I should not increase the GIF speed.

alya:
Thanks! Increased speed may be fine; I'd like to have a still screenshot alongside it regardless, so that it's easy to think about it.

alya:
I would label the option "Copy link to message" to match the web app, unless we feel like that's too long.

alya:
It also seems like it should be next to the "Share" option, so maybe we should move "Share" down to be last? We'll probably want to think about the organization more (referring to the web app and the RN app) as we add more options.

sm-sayedi:
@alya Revision pushed with the changes you suggested. Visuals in the PR comment are also updated. PTAL.

alya:
Works for me, thanks!

hackerkid:
@sm-sayedi Thanks for working on this. This looks great :)

2 minor non blocking comments. 

1. It was not super obvious what Share did without going through the issue and reading the context. A more explicit version like "Share message text" could be more useful since now it is placed below "Copy link to message" option. I think it's possible a lot of users might confuse Share option to "Share link to message".

2. The icon for Copy message text uses the popular copy icon but the copy link Icon is that of a URL. There is a mismatch in how we uses the icons. In one case we use the icon to highlight the action(copy) but in the second case it is used to highlight the entity(link) to which action is applied.

hackerkid (commented):
@sm-sayedi Thanks for working on this. This looks great :)

2 minor non blocking comments. 

1. It was not super obvious what Share did without going through the issue and reading the context. A more explicit version like "Share message text" could be more useful since now it is placed below "Copy link to message" option. I think it's possible a lot of users might confuse Share option to "Share link to message".

2. The icon for Copy message text uses the popular copy icon but the copy link Icon is that of a URL. There is a mismatch in how we uses the icons. In one case we use the icon to highlight the action(copy) but in the second case it is used to highlight the entity(link) to which action is applied.

sm-sayedi:
Thanks @hackerkid for the review! 

You're right, there is room for further adjustments. I think @gnprice first suggested replacing "Share" with the "Copy link to message" option in [zulip-mobile #2623](https://github.com/zulip/zulip-mobile/issues/2623), but I am not sure if it's still the plan. Regarding the icons, the copy icon was already used for the "Copy message text" option, and for the "Copy link to message" option, I just followed the web in using the link icon. Maybe @terpimost has a suggestion about this!

alya:
Probably #design is a good place to discuss the icons, for broader visibility. I wouldn't block merging the PR on that, though.

assets/l10n/app_en.arb
```diff
@@ -43,8 +43,8 @@
   "@permissionsDeniedReadExternalStorage": {
     "description": "Message for dialog asking the user to grant permissions for external storage read access."
   },
-  "actionSheetOptionCopy": "Copy message text",
-  "@actionSheetOptionCopy":  {
+  "actionSheetOptionCopyMessageText": "Copy message text",
+  "@actionSheetOptionCopyMessageText":  {
```
chrisbobbe, line 47: Cool. I think I'd like to rename `successMessageCopied` to `successMessageTextCopied`, too.

I think it would also be good to make that UI string more parallel to this existing "Copy message text" button. Instead of "Message Copied" (which has the wrong capitalization anyway), it could say "Message text copied".

Changing the user-facing text won't be NFC, so if it goes in this commit, the commit should be unmarked `[nfc]`. I don't think I have a preference if that change goes here or in a different commit.

lib/widgets/action_sheet.dart
```diff
@@ -361,7 +361,7 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
 
     if (!messageListContext.mounted) return;
 
-    copyWithPopup(context: context,
+    copyWithPopup(context: messageListContext,
```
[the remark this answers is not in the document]
chrisbobbe: It would be good to add a regression test for this bug. Reading through test/widgets/clipboard_test.dart, I think this might look something like the following:

```diff
--- test/widgets/action_sheet_test.dart
+++ test/widgets/action_sheet_test.dart
@@ -8,6 +8,7 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:http/http.dart' as http;
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/route/messages.dart';
+import 'package:zulip/model/binding.dart';
 import 'package:zulip/model/compose.dart';
 import 'package:zulip/model/internal_link.dart';
 import 'package:zulip/model/localizations.dart';
@@ -465,6 +466,9 @@ void main() {
     }
 
     testWidgets('success', (WidgetTester tester) async {
+      // for #732 regression check below
+      testBinding.deviceInfoResult = IosDeviceInfo(systemVersion: '16.0');
+
       final message = eg.streamMessage();
       await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
       final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
@@ -473,6 +477,14 @@ void main() {
       await tapCopyMessageTextButton(tester);
       await tester.pump(Duration.zero);
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals('Hello world');
+
+      // Regression check for #732
+      final snackBar = tester.widget<SnackBar>(find.byType(SnackBar));
+      check(snackBar.behavior).equals(SnackBarBehavior.floating);
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+      tester.widget(find.descendant(matchRoot: true,
+        of: find.byWidget(snackBar.content),
+        matching: find.text(zulipLocalizations.successMessageCopied)));
     });
 
     testWidgets('request has an error', (WidgetTester tester) async {
```

lib/widgets/action_sheet.dart
```diff
@@ -39,12 +39,12 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       return Column(children: [
         if (!hasThumbsUpReactionVote) AddThumbsUpButton(message: message, messageListContext: context),
         StarButton(message: message, messageListContext: context),
-        ShareButton(message: message, messageListContext: context),
         if (isComposeBoxOffered) QuoteAndReplyButton(
           message: message,
           messageListContext: context,
         ),
         CopyMessageTextButton(message: message, messageListContext: context),
+        ShareButton(message: message, messageListContext: context),
```
chrisbobbe:
```
action-sheet: Move the "Share" option to the bottom of options
```

Let's also move the 'ShareButton' group in the tests, so the tests still match the order of the code they're testing.

lib/widgets/action_sheet.dart
```diff
@@ -310,6 +313,37 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
...
+
+    final store = PerAccountStoreWidget.of(messageListContext);
+    final messageLink = narrowLink(
+      store,
+      SendableNarrow.ofMessage(message, selfUserId: store.selfUserId),
+      nearMessageId: message.id,
+    );
+
+    copyWithPopup(context: messageListContext,
+        successContent: Text(zulipLocalizations.successLinkCopied),
```
chrisbobbe, line 342: Let's make a new string in `zulipLocalizations`, probably with the name `successMessageLinkCopied`, saying "Message link copied", similar to the "Message text copied" from my previous comment. The `successLinkCopied` string is meant for a different interaction (long-pressing a link that appears in rendered message content).

test/widgets/action_sheet_test.dart
```diff
@@ -490,4 +491,31 @@ void main() {
...
+    testWidgets('copies message link to clipboard', (tester) async {
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
```
chrisbobbe, line 518: Is there a reason to compute `expectedLink` before simulating the button tap? If `narrowLink` had any side effects on the `store` it's passed, then this wouldn't be quite representative of the interaction we're trying to test. I think `narrowLink` doesn't have such side effects (and is unlikely to grow any). But if we call it after the test has finished processing the button tap, then it's just something we don't have to think about.

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
chrisbobbe: Then in the later commit adding "Copy message link", we can have a similar check, but it doesn't need "regression check" comments because it doesn't have a history of being affected by #732. (It's just a nice check to have, since it's easy to get the code wrong, like with the other button.)

chrisbobbe (commented):
Thanks, @sm-sayedi! Small comments below.

test/widgets/action_sheet_test.dart
```diff
@@ -472,6 +476,14 @@ void main() {
       await tapCopyMessageTextButton(tester);
       await tester.pump(Duration.zero);
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals('Hello world');
+
+      // regression check for #732
+      final snackbar = tester.widget<SnackBar>(find.byType(SnackBar));
```
sm-sayedi, line 418: This test (and the next test for CopyMessageLinkButtion) still passes with the old code where the wrong `BuildContext` is passed! In the test environment, that context seems to be mounted even `Navigator.of(context).pop()` is called before.

sm-sayedi (commented):
Thanks @chrisbobbe for the review! Revision pushed! Added one comment below. PTAL.

test/widgets/action_sheet_test.dart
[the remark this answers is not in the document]
chrisbobbe: Oh interesting; thanks for noticing this. I'm not sure what could be going on there. I'm pretty sure I would have tried this when I suggested the regression test in https://github.com/zulip/zulip-flutter/pull/713#discussion_r1637100823 . One thing I'm curious about: does #732 still reproduce with the old code when you run the app?

I guess it's possible that some other change happened that ended up fixing it.

…aha! On version 0.0.16 on iOS, #732 actually doesn't reproduce anymore! So I guess that's what happened.

We still want to pass the correct `BuildContext`, though, since it's reasonable for it to be unmounted after the action sheet is popped, and it would be a bug to do things with it after it's unmounted. But this should go in its own commit, without the `Fixes` line. An earlier commit can add the regression check, and also leave out the `Fixes` line, and just say that the bug seems to have been fixed at some point after we discovered it.

chrisbobbe:
Thanks, LGTM! I replied above (https://github.com/zulip/zulip-flutter/pull/713#discussion_r1653351260 ) with thoughts on #732.

test/widgets/action_sheet_test.dart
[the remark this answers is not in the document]
sm-sayedi: > One thing I'm curious about: does https://github.com/zulip/zulip-flutter/issues/732 still reproduce with the old code when you run the app?

Yeah, the bug is still there when we run the app on a real device (checked on my Android 12).

> On version 0.0.16 on iOS, https://github.com/zulip/zulip-flutter/issues/732 actually doesn't reproduce anymore!

It is reproduced on Android devices with `sdkInt <= 32`. If we put some logging when testing with a real device, the `BuildContext` is unmounted, but with tests, it is still mounted. So the bug still exists, it seems that the `BuildContext` popping mechanism is kind of weird in the testing environment.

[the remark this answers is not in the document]
chrisbobbe: Ah OK; thanks for checking that.

Are you seeing the bug with both the "Copy message text" button and the "Copy link to message" button? One difference is that the former waits for a network request between the `.pop()` and the `copyWithPopup`. I wonder if the context is unmounted asynchronously, some duration after the `.pop()`, and that duration is racing with the network request. (Sometimes one finishes first, sometimes the other.)

I think this might be what's happening. On my iPhone, if I edit the "Copy message text" code to remove the `await fetchRawContentWithFeedback` (hard-coding `rawContent` to some string instead), then I replace it with `await Future.delayed(Duration(milliseconds: n));`, the bug reproduces when `n` is large (like 1000) but not when `n` is small (like 10). When it reproduces, `context.mounted` is false, as you've seen too; when it doesn't reproduce, it's true.

If this seems right to you, then I think the fix is to make the test simulate the network request taking much longer than the unmounting that it's racing against. I'm having trouble getting this to work properly, but it should be possible, and Greg may be able to help if you have trouble with it too.

chrisbobbe:
Marking for Greg's review, as I think there may be a way forward regarding #732, see: https://github.com/zulip/zulip-flutter/pull/713#discussion_r1657899344

test/widgets/action_sheet_test.dart
```diff
@@ -449,29 +387,40 @@ void main() {
...
       await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
       final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
 
       prepareRawContentResponseSuccess(store, message: message, rawContent: 'Hello world');
-      await tapCopyButton(tester);
+      await tapCopyMessageTextButton(tester);
       await tester.pump(Duration.zero);
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals('Hello world');
+
+      // regression check for #732
```
gnprice, line 417: Let's split this into its own test case (its own separate `testWidgets` call).

That will duplicate a few lines from this test case, but that's OK. It'll help make both test cases easier to read and to understand what they're saying, by telling one story at a time instead of telling two stories at once.

gnprice (commented):
Thanks @sm-sayedi, and thanks everyone for the previous reviews! This code all looks great modulo one small comment below.

We'll also want to resolve the subthread above, to get a working test for that bug fix. The hypothesis at https://github.com/zulip/zulip-flutter/pull/713/files#r1657899344 makes sense to me. I'll see if I can quick whip up a way to make a FakeApiConnection request take a long (simulated) time to complete.

test/widgets/action_sheet_test.dart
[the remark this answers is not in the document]
gnprice: Yep, here's a (hacky) diff that succeeds in reproducing the issue after `git checkout @~~~ -- lib` to get the code from before the bugfix, but passes after the bugfix:
```diff
diff --git test/api/fake_api.dart test/api/fake_api.dart
index 1cbbffcdc..5e4411133 100644
--- test/api/fake_api.dart
+++ test/api/fake_api.dart
@@ -94,7 +94,7 @@ class FakeHttpClient extends http.BaseClient {
         return Future(() => throw exception);
       case _PreparedSuccess(:var bytes, :var httpStatus):
         final byteStream = http.ByteStream.fromBytes(bytes);
-        return Future(() => http.StreamedResponse(
+        return Future.delayed(Duration(milliseconds: 1000), () => http.StreamedResponse(
           byteStream, httpStatus, request: request));
     }
   }
diff --git test/widgets/action_sheet_test.dart test/widgets/action_sheet_test.dart
index db7852627..b54fa947d 100644
--- test/widgets/action_sheet_test.dart
+++ test/widgets/action_sheet_test.dart
@@ -411,7 +411,9 @@ void main() {
 
       prepareRawContentResponseSuccess(store, message: message, rawContent: 'Hello world');
       await tapCopyMessageTextButton(tester);
-      await tester.pump(Duration.zero);
+      for (int i = 0; i < 11; i++) {
+        await tester.pump(const Duration(milliseconds: 100));
+      }
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals('Hello world');
 
       // regression check for #732
```

I'll see about wrapping the `fake_api.dart` side of that into an option you can pass to `FakeApiConnection.prepare`.

Then @sm-sayedi please experiment and try to simplify those magic numbers of 1000ms, 100ms, 11 iterations — don't need to spend a ton of time doing so, but if we can it'd be nice to reduce it to something that sheds a bit more light on just how this gets triggered.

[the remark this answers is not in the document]
gnprice: OK, I pushed a draft commit to the tip of this PR:
571df4b5e wip fake_api [nfc]: Add `delay` option to connection.prepare; TODO test

I'll need to write a test for that test code, and then can send it as a PR. In the meantime the only thing it's missing is such a test, so @sm-sayedi you can build atop that and plan to rebase atop the finished version once it lands.

With that commit in place, a somewhat less hacky diff that reproduces the issue is:
```diff
diff --git test/widgets/action_sheet_test.dart test/widgets/action_sheet_test.dart
index db7852627..ea91f632e 100644
--- test/widgets/action_sheet_test.dart
+++ test/widgets/action_sheet_test.dart
@@ -90,7 +90,7 @@ void main() {
     // Prepare fetch-raw-Markdown response
     // TODO: Message should really only differ from `message`
     //   in its content / content_type, not in `id` or anything else.
-    (store.connection as FakeApiConnection).prepare(json:
+    (store.connection as FakeApiConnection).prepare(delay: Duration(seconds: 1), json:
       GetMessageResult(message: eg.streamMessage(contentMarkdown: rawContent)).toJson());
   }
 
@@ -411,7 +411,9 @@ void main() {
 
       prepareRawContentResponseSuccess(store, message: message, rawContent: 'Hello world');
       await tapCopyMessageTextButton(tester);
-      await tester.pump(Duration.zero);
+      for (int i = 0; i < 11; i++) {
+        await tester.pump(const Duration(milliseconds: 100));
+      }
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals('Hello world');
 
       // regression check for #732
```

Then the `prepareRawContentResponseSuccess` should take an optional duration to configure that behavior; I'll leave that for you to do as part of writing the test for the bugfix.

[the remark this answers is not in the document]
gnprice: Sent that `delay` feature as #770.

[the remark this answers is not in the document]
sm-sayedi: > Are you seeing the bug with both the "Copy message text" button and the "Copy link to message" button?

Yeah, for the "Copy link to message" button the bug is present both in manual and widget testing. As you explained in [#713 comment](https://github.com/zulip/zulip-flutter/pull/713#discussion_r1657899344), `.pop()` takes a small amount of time to finish. When testing on my phone, it took around 20 milliseconds for the context to be unmounted.
For the "Copy message text" button the bug is only present in widget testing (which is resolved in the latest revision). In manual testing, `await fetchRawContentWithFeedback` gives the `BuildContext` enough time to be unmounted.

test/widgets/action_sheet_test.dart
```diff
@@ -490,4 +456,116 @@ void main() {
...
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
+      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
+    });
+
+    testWidgets('copies message link to clipboard with a snackbar', (tester) async {
```
sm-sayedi, line 486: I wasn't able to make this test not pass when the wrong context was passed. The approach that you suggested for the "Copy message text" button doesn't quite apply here as we don't await for any operation as we do in "Copy message text" (`await fetchRawContentWithFeedback`).

sm-sayedi (commented):
Thanks @chrisbobbe and @gnprice for the reviews. Revision pushed. PTAL.

One [comment](https://github.com/zulip/zulip-flutter/pull/713#discussion_r1659164403) in the sub-thread above and one below.

test/widgets/action_sheet_test.dart
```diff
@@ -469,18 +408,45 @@ void main() {
...
+
+      prepareRawContentResponseSuccess(store,
+        message: message,
+        rawContent: 'Hello world',
+        delay: const Duration(milliseconds: 500));
+
+      await tapCopyMessageTextButton(tester);
+      for (int i = 0; i < 5; i++) {
+        await tester.pump(const Duration(milliseconds: 100));
+      }
```
sm-sayedi, line 432:
```suggestion
      await tester.pump(const Duration(milliseconds: 500));
```
I wonder why this line doesn't behave the same!!

gnprice: The first version will advance the simulated clock by 100ms, then render a frame; then advance another 100ms, then render another frame; and so on 5 times.

The second version will advance the simulated clock by 500ms, and then render a frame.

So the first version renders four frames before the 500ms mark, in addition to one frame at it. Each of those frames will rebuild widgets that need rebuilding, which may add and remove things from the tree. If one of those frames causes the action sheet's context to get removed from the tree, then that'd do it.

test/widgets/action_sheet_test.dart
gnprice: Yeah, I think the lack of such an `await` means that if this "Copy message link" button were to pass the wrong context to `copyWithPopup`, it would be only a latent bug and not a live one.

That is, the outward behavior would still be completely correct, and it'd be a bug only in the sense that the code would be misleading and would be susceptible to starting to do the wrong thing in the future as a result of otherwise-innocent changes. Like introducing a step with an `await`.

It'd still be good to have a test that would catch such a breakage. After all, the main purpose of a test is precisely to catch when an innocent-seeming change accidentally breaks something.

But I don't see a way to write such a test that I'd actually expect to successfully break in that case. It'd involve somehow causing any futures that the button's `onPress` method awaited to take a long (simulated) time to complete, while the navigation animation pressed on ahead… and I don't see a way to target those delays like that, given that there isn't currently anything it actually does await on and there's no way to predict what (if anything) it might someday in the future grow a need to await on.

So it's fine to not have that form of test for this new button.

sm-sayedi:
Just rebased this PR on top of #777 because of the build errors mentioned in #771, so the first two commits are from #777.

test/widgets/action_sheet_test.dart
```diff
@@ -490,4 +456,116 @@ void main() {
...
+
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
+      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
```
gnprice: This check is redundant with the same check in the test case above it. Best to leave it out, so that each test is focused on one question and it's therefore easier to understand what the test is about.

test/widgets/action_sheet_test.dart
```diff
@@ -490,4 +456,116 @@ void main() {
...
+
+      final snackbar = tester.widget<SnackBar>(find.byType(SnackBar));
+      check(snackbar.behavior).equals(SnackBarBehavior.floating);
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+      tester.widget(find.descendant(matchRoot: true,
+        of: find.byWidget(snackbar.content),
+        matching: find.text(zulipLocalizations.successMessageLinkCopied)));
```
gnprice: Given the discussion above at https://github.com/zulip/zulip-flutter/pull/713#discussion_r1659170195 , I think there isn't really anything this test case is testing that isn't covered by the tests for `copyWithPopup`.

In the regression test above, we're testing that the button calls `copyWithPopup` in a *right* way, rather than a wrong way. But this test currently doesn't really have a way to distinguish a wrong vs. a right way of calling that function. So at most it's testing that we do call `copyWithPopup`. I think that isn't worth the cost of having this test here, and having to reread and try to understand it when changing things. (Possibly it could be worth the cost if we found a really concise, simple way of expressing this test.) So let's just leave this test case out.

gnprice (commented):
Thanks @sm-sayedi for the revision! Looks good; merging, with two adjustments:

As described below, one of these test cases can be left out. I revised the commit to do that.

I also added a new commit with some small clarifications:
aeb6a94ff action-sheet test: Simplify and explain a bit more the #732 repro test

including clearly labeling that whole test case as a regression test for #732, and taking material from https://github.com/zulip/zulip-flutter/pull/713#discussion_r1659398767 . Please take a look at those changes.

Issue #712, opened 2024-05-30
Show a dot for unreads if channel is muted


Khader-1:
Just as in the web app we need to show a dot indicator on the subscription item if it is muted and has unreads.
<img width="282" alt="image" src="https://github.com/zulip/zulip-flutter/assets/63330019/ca9f5162-3ba6-485e-ae4b-215caf7c1961">
## Related issues
- #346

PR #714, opened 2024-05-30
subscription_list: Show a dot for unreads if channel is muted

Khader-1:
Fixes: #712

Final Results
image

### DIFF
--- a/lib/widgets/subscription_list.dart
+++ b/lib/widgets/subscription_list.dart
@@ -246,7 +246,7 @@ class SubscriptionItem extends StatelessWidget {
                     // TODO(design) check if this is the right variable
                     color: designVariables.labelMenuButton,
                   ).merge(weightVariableTextStyle(context,
-                      wght: hasUnreads ? 600 : null)),
+                      wght: hasUnreads && !subscription.isMuted ? 600 : null)),
                   maxLines: 1,
                   overflow: TextOverflow.ellipsis,
                   subscription.name)))),
--- a/lib/widgets/subscription_list.dart
+++ b/lib/widgets/subscription_list.dart
@@ -189,8 +189,11 @@ class _SubscriptionList extends StatelessWidget {
       itemBuilder: (BuildContext context, int index) {
         final subscription = subscriptions[index];
         final unreadCount = unreadsModel!.countInChannel(subscription.streamId);
-        // TODO(#712): if stream muted, show a dot for unreads
-        return SubscriptionItem(subscription: subscription, unreadCount: unreadCount);
+        final showMutedUnreadBadge = unreadCount == 0
+          && unreadsModel!.countInChannelNarrow(subscription.streamId) > 0;
+        return SubscriptionItem(subscription: subscription,
+          unreadCount: unreadCount,
+          showMutedUnreadBadge: showMutedUnreadBadge);
     });
   }
 }
@@ -201,10 +204,12 @@ class SubscriptionItem extends StatelessWidget {
     super.key,
     required this.subscription,
     required this.unreadCount,
+    required this.showMutedUnreadBadge,
   });
 
   final Subscription subscription;
   final int unreadCount;
+  final bool showMutedUnreadBadge;
 
   @override
   Widget build(BuildContext context) {
@@ -250,7 +255,7 @@ class SubscriptionItem extends StatelessWidget {
                   maxLines: 1,
                   overflow: TextOverflow.ellipsis,
                   subscription.name)))),
-          if (unreadCount > 0) ...[
+          if (hasUnreads) ...[
             const SizedBox(width: 12),
             // TODO(#747) show @-mention indicator when it applies
             Opacity(
@@ -259,6 +264,10 @@ class SubscriptionItem extends StatelessWidget {
                 count: unreadCount,
                 backgroundColor: swatch,
                 bold: true)),
+          ] else if (showMutedUnreadBadge) ...[
+            const SizedBox(width: 12),
+            // TODO(#747) show @-mention indicator when it applies
+            const MutedUnreadBadge(),
           ],
           const SizedBox(width: 16),
         ])));
--- a/lib/widgets/unread_count_badge.dart
+++ b/lib/widgets/unread_count_badge.dart
@@ -57,3 +57,18 @@ class UnreadCountBadge extends StatelessWidget {
           count.toString())));
   }
 }
+
+class MutedUnreadBadge extends StatelessWidget {
+  const MutedUnreadBadge({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      width: 8,
+      height: 8,
+      margin: const EdgeInsetsDirectional.only(end: 3),
+      decoration: BoxDecoration(
+        color: const HSLColor.fromAHSL(0.5, 0, 0, 0.8).toColor(),
+        shape: BoxShape.circle));
+  }
+}
--- a/test/widgets/subscription_list_test.dart
+++ b/test/widgets/subscription_list_test.dart
@@ -154,6 +154,7 @@ void main() {
       eg.subscription(stream),
     ], unreadMsgs: unreadMsgs);
     check(find.byType(UnreadCountBadge).evaluate()).length.equals(1);
+    check(find.byType(MutedUnreadBadge).evaluate().length).equals(0);
   });
 
   testWidgets('unread badge counts unmuted only', (tester) async {
@@ -174,6 +175,7 @@ void main() {
     check(tester.widget<Text>(find.descendant(
         of: find.byType(UnreadCountBadge), matching: find.byType(Text))))
       .data.equals('1');
+    check(find.byType(MutedUnreadBadge).evaluate().length).equals(0);
   });
 
   testWidgets('unread badge does not show with no unreads', (tester) async {
@@ -183,6 +185,46 @@ void main() {
       eg.subscription(stream),
     ], unreadMsgs: unreadMsgs);
     check(find.byType(UnreadCountBadge).evaluate()).length.equals(0);
+    check(find.byType(MutedUnreadBadge).evaluate().length).equals(0);
+  });
+
+  testWidgets('muted unread badge shows when unreads are visible in channel but not inbox', (tester) async {
+    final stream = eg.stream();
+    final unreadMsgs = eg.unreadMsgs(channels: [
+      UnreadChannelSnapshot(streamId: stream.streamId, topic: 'b', unreadMessageIds: [3]),
+    ]);
+    await setupStreamListPage(tester,
+      subscriptions: [eg.subscription(stream, isMuted: true)],
+      userTopics: [eg.userTopicItem(stream, 'b', UserTopicVisibilityPolicy.none)],
+      unreadMsgs: unreadMsgs);
+
+    check(find.byType(MutedUnreadBadge).evaluate().length).equals(1);
+  });
+
+  testWidgets('muted unread badge does not show when unreads are visible in both channel & inbox', (tester) async {
+    final stream = eg.stream();
+    final unreadMsgs = eg.unreadMsgs(channels: [
+      UnreadChannelSnapshot(streamId: stream.streamId, topic: 'b', unreadMessageIds: [3]),
+    ]);
+    await setupStreamListPage(tester,
+      subscriptions: [eg.subscription(stream, isMuted: false)],
+      userTopics: [eg.userTopicItem(stream, 'b', UserTopicVisibilityPolicy.none)],
+      unreadMsgs: unreadMsgs);
+
+    check(find.byType(MutedUnreadBadge).evaluate().length).equals(0);
+  });
+
+  testWidgets('muted unread badge does not show when unreads are not visible in channel nor inbox', (tester) async {
+    final stream = eg.stream();
+    final unreadMsgs = eg.unreadMsgs(channels: [
+      UnreadChannelSnapshot(streamId: stream.streamId, topic: 'b', unreadMessageIds: [3]),
+    ]);
+    await setupStreamListPage(tester,
+      subscriptions: [eg.subscription(stream, isMuted: true)],
+      userTopics: [eg.userTopicItem(stream, 'b', UserTopicVisibilityPolicy.muted)],
+      unreadMsgs: unreadMsgs);
+
+    check(find.byType(MutedUnreadBadge).evaluate().length).equals(0);
   });
 
   testWidgets('color propagates to icon and badge', (tester) async {
--- a/test/widgets/subscription_list_test.dart
+++ b/test/widgets/subscription_list_test.dart
@@ -6,6 +6,7 @@ import 'package:zulip/api/model/model.dart';
 import 'package:zulip/widgets/icons.dart';
 import 'package:zulip/widgets/channel_colors.dart';
 import 'package:zulip/widgets/subscription_list.dart';
+import 'package:zulip/widgets/text.dart';
 import 'package:zulip/widgets/unread_count_badge.dart';
 
 import '../flutter_checks.dart';
@@ -232,5 +233,44 @@ void main() {
 
     checkOpacityForStreamAndBadge('Stream 1', 2, 0.55);
     checkOpacityForStreamAndBadge('Stream 2', 1, 1.0);
+  });
+
+  testWidgets('stream name of unmuted streams with unmuted unreads is bold', (tester) async {
+    void checkStreamNameWght(String streamName, double? expectedWght) {
+      final streamFinder = find.text(streamName);
+      final wght = wghtFromTextStyle(tester.widget<Text>(streamFinder).style!);
+      check(wght).equals(expectedWght);
+    }
+
+    final unmutedStreamWithUnmutedUnreads =   eg.stream(name: 'Unmuted stream with unmuted unreads');
+    final unmutedStreamWithNoUnmutedUnreads = eg.stream(name: 'Unmuted stream with no unmuted unreads');
+    final mutedStreamWithUnmutedUnreads =     eg.stream(name: 'Muted stream with unmuted unreads');
+    final mutedStreamWithNoUnmutedUnreads =   eg.stream(name: 'Muted stream with no unmuted unreads');
+
+    await setupStreamListPage(tester,
+      subscriptions: [
+        eg.subscription(unmutedStreamWithUnmutedUnreads,   isMuted: false),
+        eg.subscription(unmutedStreamWithNoUnmutedUnreads, isMuted: false),
+        eg.subscription(mutedStreamWithUnmutedUnreads,     isMuted: true),
+        eg.subscription(mutedStreamWithNoUnmutedUnreads,   isMuted: true),
+      ],
+      userTopics: [
+        eg.userTopicItem(unmutedStreamWithUnmutedUnreads,   'a', UserTopicVisibilityPolicy.unmuted),
+        eg.userTopicItem(unmutedStreamWithNoUnmutedUnreads, 'b', UserTopicVisibilityPolicy.muted),
+        eg.userTopicItem(mutedStreamWithUnmutedUnreads,     'c', UserTopicVisibilityPolicy.unmuted),
+        eg.userTopicItem(mutedStreamWithNoUnmutedUnreads,   'd', UserTopicVisibilityPolicy.muted),
+      ],
+      unreadMsgs: eg.unreadMsgs(channels: [
+        UnreadChannelSnapshot(streamId: unmutedStreamWithUnmutedUnreads.streamId,   topic: 'a', unreadMessageIds: [1]),
+        UnreadChannelSnapshot(streamId: unmutedStreamWithNoUnmutedUnreads.streamId, topic: 'b', unreadMessageIds: [2]),
+        UnreadChannelSnapshot(streamId: mutedStreamWithUnmutedUnreads.streamId,     topic: 'c', unreadMessageIds: [3]),
+        UnreadChannelSnapshot(streamId: mutedStreamWithNoUnmutedUnreads.streamId,   topic: 'd', unreadMessageIds: [4]),
+      ]),
+    );
+
+    checkStreamNameWght(unmutedStreamWithUnmutedUnreads.name,   600);
+    checkStreamNameWght(unmutedStreamWithNoUnmutedUnreads.name, 400);
+    checkStreamNameWght(mutedStreamWithUnmutedUnreads.name,     400);
+    checkStreamNameWght(mutedStreamWithNoUnmutedUnreads.name,   400);
   });
 }


Khader-1 (commented):
< Wrong tab > 😄

lib/widgets/subscription_list.dart
```diff
@@ -239,6 +243,10 @@ class SubscriptionItem extends StatelessWidget {
             // TODO(#384) show @-mention indicator when it applies
             UnreadCountBadge(count: unreadCount, backgroundColor: swatch, bold: true),
           ],
+          if (hasMutedUnreads) ...[
```
sm-sayedi, line 246:
```suggestion
          ] else ...[
```
Can't we write something like this? 
And I wonder how is `hasMutedUnreads` different from `unreadCount <= 0`?!

Also upon interacting with the web, I noticed three interesting cases that control how the `SubscriptionItem` is shown:

1. The stream is unmuted, so everything looks normal with the unread count badge present.
2. The stream is muted and all its topics are in the **Default** state, so the icon and stream title have an opacity and the unread count badge is replaced by a gray color dot.
3. The stream is muted but one or more of its topics are in the **Unmute** state, so the icon and stream title have an opacity and there is again the unread count badge showing the sum of the **Unmute** topics' unread count. (This is the reason I suggested making the `Opacity` control only the icon and the stream title in [711 comment](https://github.com/zulip/zulip-flutter/pull/711#discussion_r1622478616).)

So to match it with the web, I think we have to consider these three cases.

lib/widgets/unread_count_badge.dart
```diff
@@ -65,3 +66,22 @@ class UnreadCountBadge extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    return Opacity(
+      opacity: 0.5,
+      child: Container(
+        width: 8,
+        height: 8,
+        margin: const EdgeInsetsDirectional.only(end: 3),
+        decoration: const BoxDecoration(
+          color: HslColor(0, 0, 0.8),
+          shape: BoxShape.circle,),),);
```
sm-sayedi, line 85:
```suggestion
          color: HslColor(0, 0, 80),
          shape: BoxShape.circle)));
```
According to the web, `lightness = 80%` and in Flutter it's `80` too.
Also, those commas are not necessary if they're on the same line.

sm-sayedi (commented):
Thanks @Khader-1. A few comments below.

lib/widgets/unread_count_badge.dart
Khader-1: Thanks! new revision pushed PTAL

lib/widgets/subscription_list.dart
Khader-1: - `unreadCount > 0`: there are unreads that are not unmuted.
- `hasMutedUnreads`: there are unreads that are muted.
The three states that we have are:

| unreadCount > 0 | hasMutedUnreads | badge |
|---|---|---|
| true | any | UnreadCountBadge |
| false | true | MutedUnreadBadge |
| false | false | None | 

So you can see that we don't need to calculate `hasMutedUnreads` in case of `unreadCount > 0` being true, which could save some heavy computing in some cases. So I try to avoid any unnecessary computations by having the `hasMutedUnreads` being conditioned on `unreadCount == 0`.
```dart
        final hasMutedUnreads = unreadCount == 0 && unreadsModel!.hasMutedInStream(subscription.streamId);
```
which would cause as a side effect that guarantees that:
```dart
 unreadCount > 0 && hasMutedUnreads = false // this is guaranteed always
 ```
 So finally I'd update the code for good measure as:
 ```dart
          if (unreadCount > 0) ...[
            const SizedBox(width: 12),
            // TODO(#384) show @-mention indicator when it applies
            UnreadCountBadge(count: unreadCount, backgroundColor: swatch, bold: true),
          ] else if (hasMutedUnreads) ...[ // else here is not required but it is best to have it just to make sure
            const SizedBox(width: 12),
            const MutedUnreadBadge(),
          ],
```

alya:
We'll want to combine releasing this with a change that makes muted channel names look faded, as otherwise it will be confusing why some channels have unread counters and some have just dots.

gnprice:
> We'll want to combine releasing this with a change that makes muted channel names look faded, as otherwise it will be confusing why some channels have unread counters and some have just dots.

Yeah, good thought. This should therefore come after:
* #424
* #711

lib/widgets/subscription_list.dart
sm-sayedi: I am snapping my head around to understand the difference, but unfortunately, I am not able to 😀. Specifically, as you said:
> unreadCount > 0: there are unreads that are not unmuted.

vs.
> hasMutedUnreads: there are unreads that are muted.

Isn't **_not unmuted_** = **_muted_**?

Also while interacting with the UI, I couldn't notice any difference between these changes and when I replace `else if (hasMutedUnreads)` with `else` and completely ignore the `hasMutedUnreads`!

sm-sayedi (commented):
Thanks for the revision @Khader-1!

Added a [comment ](https://github.com/zulip/zulip-flutter/pull/714/#discussion_r1627603658) to the previous thread, please check. 

Also, as Ayla stated, these changes need to be combined with the ones that make muted channel names look faded. So I suggest placing this PR on top of #711, that way it will be easy to check all the three cases that I mentioned in [#714 review comment](https://github.com/zulip/zulip-flutter/pull/714/#discussion_r1622522229). You can check this [GitHub answer](https://stackoverflow.com/a/29916361/14186586) for this kind of rebasing, specifically `git rebase --onto` with 2 arguments.

lib/widgets/subscription_list.dart
Khader-1: Oh sorry... Actually there is a mistake here
`unreadCount > 0`: there are unreads that are ~~not~~ unmuted.

gnprice:
It looks like this PR was paused while waiting for work on #711. Since #711 is now very close to merge, @Khader-1 I think it'd be a good time to rebase this atop the current version of that, and address the comments above if there are any you haven't already.

Then comment here when this PR is ready for the next round of buddy review from @sm-sayedi.

Khader-1:
Thanks Greg! I've just pushed a revision PTAL @sm-sayedi.

sm-sayedi (commented):
Thanks @Khader-1 for the revision. This shows almost the expected behavior.

One case still doesn't behave as expected:
For example, if we have a stream that is not muted and has two topics, with one topic being muted and having unreads and another being default but having no unreads, the mobile app will show a dot, although the web app doesn't show anything. The dot should only show for muted streams.

https://github.com/user-attachments/assets/b4e7d84f-7e8b-475c-9f75-a0bf4e73f909


Also, I am not sure if we want to show the muted stream name in bold when it has unreads for unmuted topics!!

Khader-1:
Thanks @sm-sayedi! Those two cases are handled in the new revision PTAL!

sm-sayedi (approved):
Great work @Khader-1! Looks nice to me. Moving on to the mentor review from @kenclary!

alya:
> Also, I am not sure if we want to show the muted stream name in bold when it has unreads for unmuted topics!!

Hm, yeah, that sounds right to me.

lib/widgets/unread_count_badge.dart
```diff
@@ -65,3 +66,22 @@ class UnreadCountBadge extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    return Opacity(
+      opacity: 0.5,
+      child: Container(
+        width: 8,
+        height: 8,
+        margin: const EdgeInsetsDirectional.only(end: 3),
+        decoration: const BoxDecoration(
+          color: HslColor(0, 0, 80),
+          shape: BoxShape.circle)));
```
chrisbobbe, line 85: How about putting the 0.5 opacity in the `BoxDecoration.color`, with `HSLColor.fromHSLA(…).toColor()`?

chrisbobbe (commented):
Thanks! Small comments below.

Khader-1 (commented):
Thanks for the review @chrisbobbe! I've pushed a new revision PTAL.

chrisbobbe:
Looks like CI is failing; could you fix that?

Khader-1:
Sorry about that! I've fixed it now. PTAL @chrisbobbe.

lib/widgets/subscription_list.dart
```diff
@@ -256,6 +263,9 @@ class SubscriptionItem extends StatelessWidget {
                 count: unreadCount,
                 backgroundColor: swatch,
                 bold: true)),
+          ] else if (hasAllUnreadsMuted && subscription.isMuted) ...[
+            const SizedBox(width: 12),
+            const MutedUnreadBadge(),
```
chrisbobbe, line 268: I _think_ TODO above this also applies here:

```dart
// TODO(#747) show @-mention indicator when it applies
```

In the web app, if I have an unread mention in a muted stream, where the topic is not unmuted, I see a dot and an @-mention indicator:

image

So let's copy that TODO into this case too.

lib/widgets/unread_count_badge.dart
```diff
@@ -65,3 +65,20 @@ class UnreadCountBadge extends StatelessWidget {
           count.toString())));
   }
 }
+
+class MutedUnreadBadge extends StatelessWidget {
+  const MutedUnreadBadge({
+    super.key,
+  });
```
chrisbobbe, line 72:
```suggestion
  const MutedUnreadBadge({super.key});
```

chrisbobbe (commented):
Thanks! Small comments below, and I'll mark this for Greg's review.

Khader-1:
Thanks @chrisbobbe!

lib/model/unreads.dart
```diff
@@ -206,6 +206,15 @@ class Unreads extends ChangeNotifier {
     }
   }
 
+  /// Checks if stream contains strictly muted unreads,
+  /// using [ChannelStore.isTopicVisible].
+  bool hasMutedInStream(int streamId) {
```
gnprice, line 211: This will include unreads that are in muted topics. That doesn't sound like the desired behavior — and it's not what web does, based on checking just now with a channel where I have some such unreads.

I think the desired behavior is: we want to see if there are any unreads here where `isTopicVisibleInStream` is true but the stricter `isTopicVisible` isn't.

In other words:
* We show an indicator on the channel/stream when there are some unreads that we would show if you narrow to the channel. (That's what `countInStreamNarrow` / `isTopicVisibleInStream` would measure.)
* When moreover some of those unreads are messages we'd show in the combined feed, then the indicator we show is a count of those unreads. (That's `countInStream` / `isTopicVisible`.)
* So we show a dot when the first condition is true but not the second.

Can you find the code in web that controls this aspect? That'd be good for confirming what web's behavior is to verify that my description above matches it.

Then for implementing that behavior, probably cleanest is to use the existing method `countInStreamNarrow`.

test/widgets/subscription_list_test.dart
```diff
@@ -186,6 +187,49 @@ void main() {
     check(find.byType(UnreadCountBadge).evaluate()).length.equals(0);
   });
 
+  testWidgets('muted unread badge shows with muted unreads', (tester) async {
```
gnprice, line 190: The structure of these test cases looks fine.

This case should have the opposite expectation, though — the topic (and not only the stream/channel) is muted, so there should be no badge.

Then let's also have test cases exercising each of the possible scenarios:
* all the possible combinations of isTopicVisible true/false and isTopicVisibleInStream true/false;
* for each of those, all the possible values for whether the stream is muted.

This muted-unreads badge is closely related to the unread-count badge. So some of those scenarios may be covered by the existing test cases above, and those can be handled by just adding another check to those cases.

test/widgets/subscription_list_test.dart
```diff
@@ -186,6 +187,49 @@ void main() {
...
+    final stream = eg.stream();
+    final unreadMsgs = eg.unreadMsgs(streams: [
+      UnreadStreamSnapshot(streamId: stream.streamId, topic: 'b', unreadMessageIds: [3]),
+    ]);
+    await setupStreamListPage(tester,
+      subscriptions: [eg.subscription(stream, isMuted: true)],
+      userTopics: [eg.userTopicItem(stream, 'b', UserTopicVisibilityPolicy.muted)],
+      unreadMsgs: unreadMsgs);
+    final finder = find.byWidgetPredicate((widget) => widget is MutedUnreadBadge);
```
gnprice, line 199: Can simplify with `find.byType`:
```suggestion
    final finder = find.byType(MutedUnreadBadge);
```

Then just inline into the check below.

For examples, see the similar test cases above about UnreadCountBadge.

gnprice (commented):
Thanks @Khader-1, and thanks @sm-sayedi @alya @chrisbobbe for the previous reviews!

The behavior of this appears to be wrong, as I wrote just above.

The first commit:
b3456dfda subscription_list: Ensure bold name applies for unmuted streams only
looks good. Comments below on the remainder of the second/main commit. I skipped commenting on that commit's changes in `subscription_list.dart`, because I think those will look different after revising the behavior.

lib/model/unreads.dart
Khader-1: Thanks @gnprice! 

Yes, I can verify you're description:

> I think the desired behavior is: we want to see if there are any unreads here where isTopicVisibleInStream is true but the stricter isTopicVisible isn't.

I've taken a look into the web code and found that the only case that applies is when stream isMuted (or `stream.isMuted` is null) and topic policy is none. and that Equals to `isTopicVisibleInStream && !isTopicVisible`.

| topic policy | stream isMuted | isTopicVisible | isTopicVisibleInStream | isTopicVisibleInStream && !isTopicVisible | web counts as unmuted unread |
| --- | --- | --- | --- | --- | --- |
| none | true | false | true | true | true |
| muted | true | false | false | false | false |
| unmuted | true | true | true | false | false |
| followed | true | true | true | false | false |
| none | null | true | true | false | true |
| muted | null | false | false | false | false |
| unmuted | null | true | true | false | false |
| followed | null | true | true | false | false |
| none | false | true | true | false | false |
| muted | false | false | false | false | false |
| unmuted | false | true | true | false | false |
| followed | false | true | true | false | false |

You can find that in [web/src/unread.ts@UnreadTopicCounter.get_stream_count_info](https://github.com/zulip/zulip/blob/main/web/src/unread.ts#L393)

```ts
get_stream_count_info(
    stream_id: number,
    update_first_unmuted_message_id = false,
): StreamCountInfo {
    const per_stream_bucketer = this.bucketer.get(stream_id);

    if (!per_stream_bucketer) {
        return {
            unmuted_count: 0,
            muted_count: 0,
            followed_count: 0,
            stream_is_muted: false,
        };
    }

    const sub = sub_store.get(stream_id);
    let unmuted_count = 0;
    let muted_count = 0;
    let followed_count = 0;
    for (const [topic, msgs] of per_stream_bucketer) {
        const topic_count = msgs.size;

        if (user_topics.is_topic_followed(stream_id, topic)) {
            followed_count += topic_count;
        }

        if (user_topics.is_topic_unmuted_or_followed(stream_id, topic)) {
            unmuted_count += topic_count;
            if (update_first_unmuted_message_id) {
                first_unread_unmuted_message_id = Math.min(
                    first_unread_unmuted_message_id,
                    ...msgs,
                );
            }
        } else if (user_topics.is_topic_muted(stream_id, topic)) {
            muted_count += topic_count;
        } else if (sub?.is_muted) {
            muted_count += topic_count;
        } else {
            unmuted_count += topic_count;
            if (update_first_unmuted_message_id) {
                first_unread_unmuted_message_id = Math.min(
                    first_unread_unmuted_message_id,
                    ...msgs,
                );
            }
        }
    }
    const stream_count = {
        unmuted_count,
        muted_count,
        followed_count,
        stream_is_muted: sub?.is_muted ?? false,
    };
    return stream_count;
}
```

And in [web/src/stream_list.ts@update_count_in_dom](https://github.com/zulip/zulip/blob/main/web/src/stream_list.ts#L100)

```ts
export function update_count_in_dom(
    $stream_li: JQuery,
    stream_counts: StreamCountInfo,
    stream_has_any_unread_mention_messages: boolean,
    stream_has_any_unmuted_unread_mention: boolean,
    stream_has_only_muted_unread_mention: boolean,
): void {
    // The subscription_block properly excludes the topic list,
    // and it also has sensitive margins related to whether the
    // count is there or not.
    const $subscription_block = $stream_li.find(".subscription_block");

    ui_util.update_unread_mention_info_in_dom(
        $subscription_block,
        stream_has_any_unread_mention_messages,
    );

    if (stream_has_any_unmuted_unread_mention) {
        $subscription_block.addClass("has-unmuted-mentions");
        $subscription_block.removeClass("has-only-muted-mentions");
    } else {
        $subscription_block.removeClass("has-unmuted-mentions");
        if (!stream_counts.stream_is_muted && stream_has_only_muted_unread_mention) {
            $subscription_block.addClass("has-only-muted-mentions");
        } else {
            $subscription_block.removeClass("has-only-muted-mentions");
        }
    }

    // Here we set the count and compute the values of two classes:
    // .stream-with-count is used for the layout CSS to know whether
    // to leave space for the unread count, and has-unmuted-unreads is
    // used in muted streams to set the fading correctly to indicate
    // those are unread
    if (stream_counts.unmuted_count > 0 && !stream_counts.stream_is_muted) {
        // Normal stream, has unmuted unreads; display normally.
        ui_util.update_unread_count_in_dom($subscription_block, stream_counts.unmuted_count);
        $subscription_block.addClass("stream-with-count");
        $subscription_block.removeClass("has-unmuted-unreads");
        $subscription_block.removeClass("has-only-muted-unreads");
    } else if (stream_counts.unmuted_count > 0 && stream_counts.stream_is_muted) {
        // Muted stream, has unmuted unreads.
        ui_util.update_unread_count_in_dom($subscription_block, stream_counts.unmuted_count);
        $subscription_block.addClass("stream-with-count");
        $subscription_block.addClass("has-unmuted-unreads");
        $subscription_block.removeClass("has-only-muted-unreads");
    } else if (stream_counts.muted_count > 0 && stream_counts.stream_is_muted) {
        // Muted stream, only muted unreads.
        ui_util.update_unread_count_in_dom($subscription_block, stream_counts.muted_count);
        $subscription_block.addClass("stream-with-count");
        $subscription_block.removeClass("has-unmuted-unreads");
        $subscription_block.removeClass("has-only-muted-unreads");
    } else if (
        stream_counts.muted_count > 0 &&
        !stream_counts.stream_is_muted &&
        stream_has_only_muted_unread_mention
    ) {
        // Normal stream, only muted unreads, including a mention:
        // Display the mention, faded, and a faded unread count too,
        // so that we don't weirdly show the mention indication
        // without an unread count.
        ui_util.update_unread_count_in_dom($subscription_block, stream_counts.muted_count);
        $subscription_block.removeClass("has-unmuted-unreads");
        $subscription_block.addClass("stream-with-count");
        $subscription_block.addClass("has-only-muted-unreads");
    } else if (stream_counts.muted_count > 0 && !stream_counts.stream_is_muted) {
        // Normal stream, only muted unreads: display nothing. The
        // current thinking is displaying those counts with muted
        // styling would be more distracting than helpful.
        ui_util.update_unread_count_in_dom($subscription_block, 0);
        $subscription_block.removeClass("has-unmuted-unreads");
        $subscription_block.removeClass("stream-with-count");
    } else {
        // No unreads: display nothing.
        ui_util.update_unread_count_in_dom($subscription_block, 0);
        $subscription_block.removeClass("has-unmuted-unreads");
        $subscription_block.removeClass("has-only-muted-unreads");
        $subscription_block.removeClass("stream-with-count");
    }
}
```

gnprice: Cool, thanks for investigating that.

In your table, in the row for none/null, isTopicVisible should be false, not true, right? We have:
```dart
  bool isTopicVisible(int streamId, String topic) {
    switch (topicVisibilityPolicy(streamId, topic)) {
      case UserTopicVisibilityPolicy.none:
        switch (subscriptions[streamId]?.isMuted) {
          case false: return true;
          case true:  return false;
          case null:  return false; // not subscribed; treat like muted
```

so none/null behaves the same as none/true.

That's good, though, because I think that's what makes this conclusion true:

> the only case that applies is when stream isMuted (or `stream.isMuted` is null) and topic policy is none. and that Equals to `isTopicVisibleInStream && !isTopicVisible`.

Khader-1:
Thank @gnprice! I've made a revision PTAL

lib/widgets/subscription_list.dart
```diff
@@ -200,15 +204,18 @@ class SubscriptionItem extends StatelessWidget {
     super.key,
     required this.subscription,
     required this.unreadCount,
+    required this.hasUnmutedUnreads,
```
gnprice, line 207: This field is redundant with `unreadCount`. The code in this widget's build method, and the code at the call site constructing one of these widgets, will both be clearer if there's just the one field — otherwise the reader has to wonder about the case where `hasUnmutedUnreads` doesn't agree with `unreadCount > 0`.

lib/widgets/subscription_list.dart
```diff
@@ -188,8 +188,12 @@ class _SubscriptionList extends StatelessWidget {
       itemBuilder: (BuildContext context, int index) {
         final subscription = subscriptions[index];
         final unreadCount = unreadsModel!.countInChannel(subscription.streamId);
-        // TODO(#712): if stream muted, show a dot for unreads
-        return SubscriptionItem(subscription: subscription, unreadCount: unreadCount);
+        final hasUnmutedUnreads = unreadCount > 0;
+        final showMutedUnreadBadge = !hasUnmutedUnreads && unreadsModel!.hasMutedInChannel(subscription.streamId);
```
gnprice, line 192: As I suggested at https://github.com/zulip/zulip-flutter/pull/714#discussion_r1694343677, I think this logic will be cleaner if it uses the existing method `countInChannelNarrow` (previously `countInStreamNarrow`).

gnprice (commented):
Thanks! Comments below.

Khader-1:
Thanks @gnprice! Please check the latest revision.

lib/widgets/subscription_list.dart
```diff
@@ -189,8 +189,10 @@ class _SubscriptionList extends StatelessWidget {
       itemBuilder: (BuildContext context, int index) {
         final subscription = subscriptions[index];
         final unreadCount = unreadsModel!.countInChannel(subscription.streamId);
-        // TODO(#712): if stream muted, show a dot for unreads
-        return SubscriptionItem(subscription: subscription, unreadCount: unreadCount);
+        final showMutedUnreadBadge = unreadCount == 0 && unreadsModel!.countInChannelNarrow(subscription.streamId) > 0;
```
gnprice, line 192: nit: line too long, and important information after column 80
```suggestion
        final showMutedUnreadBadge = unreadCount == 0
          && unreadsModel!.countInChannelNarrow(subscription.streamId) > 0;
```

gnprice (commented):
Thanks for the revision! Looks good, with one nit below. Fixing that and merging.

Issue #340, opened 2023-10-26
Play Zulip's distinctive notification sound on Android


gnprice:
In zulip-mobile we have a distinctive notification sound we created a couple of years ago, which we use on Android by default. We also have a suite of variant sounds which we provide for the user to choose from if they use the system notification settings.

We should do the same here. My initial implementation of #320 / #122 won't do so, in order to manage complexity.

We should also eventually do so on iOS. That will be tracked separately, because it's a feature we don't yet have in zulip-mobile either: https://github.com/zulip/zulip-mobile/issues/3101 .

PR #717, opened 2024-06-02
notif: Use Zulip's notification sound on Android 

rajveermalviya:
Updates: #340

For now, only includes the default sound, i.e. it doesn't show multiple sound options in notification settings - this will be done in later PRs.


lib/notifications/display.dart
```diff
@@ -53,14 +56,37 @@ class NotificationChannelManager {
...
-      ?.createNotificationChannel(AndroidNotificationChannel(
+    final androidPlugin = plugin.resolvePlatformSpecificImplementation<AndroidFlutterLocalNotificationsPlugin>();
+    if (androidPlugin == null) {
+      // TODO(log)
+      return;
+    }
+
+    // Delete any older channels.
+    var found = false;
+    final channels = await androidPlugin.getNotificationChannels();
```
gnprice, line 67: Rather than add more parts of the `package:flutter_local_notifications` API that we use, let's use Pigeon for this. That way in order to be confident of understanding what the code is doing, we only have to read our own code and the Android docs, and not also the `flutter_local_notifications` code :slightly_smiling_face: (which often introduces a lot of logic of its own).

As a concrete strategy, let's put this issue after #351: so we convert the existing call sites first (the remaining ones, after #592 covered the most complex of them), and then we go on to add Pigeon bindings for more of the notifications API.

sm-sayedi (approved):
Thanks @rajveermalviya for this. A few small nits below, otherwise LGTM! I have also tested it on my phone, and the notification sound works fine.

As PR was opened during the official GSoC period, let's move on to the mentor review with @sumanthvrao.

pigeon/notifications.dart
```diff
@@ -33,6 +34,15 @@ class NotificationChannel {
 
   final String? name;
   final bool? lightsEnabled;
+
+  /// The name of the sound file resource.
+  ///
+  /// The resource ID is retrieved using
+  /// `android.content.res.Resources.getIdentifier`, which is then used to
+  /// generate an `android.resource://` URI. This URI is then passed
+  /// to the `NotificationChannelCompat` builder.
```
PIG208, line 43: We could rephrase this to something similar to https://github.com/zulip/zulip-flutter/blob/cdbe14122eac145afaa86d19f152b5451e3c457b/pigeon/notifications.dart#L172-L175, mentioning how `soundResourceName` is expected to be known in keep.xml.

lib/notifications/display.dart
```diff
@@ -52,11 +52,27 @@ class NotificationChannelManager {
...
+      }
+    }
+
+    if (found) {
+      // The channel already exists; nothing to do.
+      return;
+    }
+
+    // The channel doesn't exist.  Create it.
+
```
PIG208, line 78: nit: remove empty line

lib/notifications/display.dart
```diff
@@ -52,11 +52,27 @@ class NotificationChannelManager {
...
-  //    If we do this, we should also look for any channel with the old
-  //    channel ID and delete it.  See zulip-mobile's `createNotificationChannel`
-  //    in android/app/src/main/java/com/zulipmobile/notifications/NotificationChannelManager.kt .
   static Future<void> _ensureChannel() async {
+    // See if our current-version channel already exists; delete any obsolete
+    // previous channels.
+    var found = false;
+    final channels = await _androidHost.getNotificationChannels();
+    for (final channel in channels) {
+      assert(channel != null); // TODO(flutter#97848)
```
PIG208, line 64: nit: we can refer to #942 here

PIG208 (commented):
Thanks for the PR! Tested manually and it works well for me. Left some comments.

Issue #128, opened 2023-05-26
Inbox-style notifications, like on Android in `zulip-mobile`


chrisbobbe:
This comes after #122 / #320.

This means:
* Each conversation gets a separate notification.
* That notification is an Android "inbox-style" notification: it shows up to a handful of the most recent messages, and for a 1:1 conversation the avatar of the other person.

The initial implementation of #320 will have separate notifications for separate conversations but, in order to break down the complexity into separate stages, each notification will only show the one latest message in that conversation.

PR #718, opened 2024-06-03
notif: Support messaging-style notifications

rajveermalviya:
https://github.com/zulip/zulip-flutter/assets/36819268/ddcaecd1-b489-4564-bf96-1888e71acd4c

Fixes: #128

### DIFF
--- a/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
+++ b/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
@@ -8,9 +8,37 @@ import android.util.Log
 import androidx.annotation.Keep
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
+import androidx.core.graphics.drawable.IconCompat
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 
 private const val TAG = "ZulipPlugin"
+
+fun toAndroidPerson(person: Person): androidx.core.app.Person {
+    return androidx.core.app.Person.Builder().apply {
+        person.iconBitmap?.let { setIcon(IconCompat.createWithData(it, 0, it.size)) }
+        setKey(person.key)
+        setName(person.name)
+    }.build()
+}
+
+fun toPigeonPerson(person: androidx.core.app.Person): Person {
+    return Person(
+        // The API doesn't provide a way to retrieve the icon data,
+        // so we set this to null.
+        //
+        // Notably, Android retains a limited number [1] of messages
+        // in the messaging style, and it also retains the icon data
+        // for persons within those messages. Therefore, there's no
+        // need to include the person's icon data in each message.
+        // Only one icon data instance is needed for each unique
+        // person's key in the retained messages.
+        //
+        // [1]: https://developer.android.com/reference/androidx/core/app/NotificationCompat.MessagingStyle#MAXIMUM_RETAINED_MESSAGES()
+        null,
+        person.key!!,
+        person.name!!.toString(),
+    )
+}
 
 private class AndroidNotificationHost(val context: Context)
         : AndroidNotificationHostApi {
@@ -33,6 +61,8 @@ private class AndroidNotificationHost(val context: Context)
         groupKey: String?,
         inboxStyle: InboxStyle?,
         isGroupSummary: Boolean?,
+        messagingStyle: MessagingStyle?,
+        number: Long?,
         smallIconResourceName: String?
     ) {
         val notification = NotificationCompat.Builder(context, channelId).apply {
@@ -60,10 +90,47 @@ private class AndroidNotificationHost(val context: Context)
                     .setSummaryText(it.summaryText)
             ) }
             isGroupSummary?.let { setGroupSummary(it) }
+            messagingStyle?.let { messagingStyle ->
+                val style = NotificationCompat.MessagingStyle(toAndroidPerson(messagingStyle.user))
+                    .setConversationTitle(messagingStyle.conversationTitle)
+                    .setGroupConversation(messagingStyle.isGroupConversation)
+                messagingStyle.messages.forEach { it?.let {
+                    style.addMessage(NotificationCompat.MessagingStyle.Message(
+                        it.text,
+                        it.timestampMs,
+                        toAndroidPerson(it.person),
+                    ))
+                } }
+                setStyle(style)
+            }
+            number?.let { setNumber(it.toInt()) }
             smallIconResourceName?.let { setSmallIcon(context.resources.getIdentifier(
                 it, "drawable", context.packageName)) }
         }.build()
         NotificationManagerCompat.from(context).notify(tag, id.toInt(), notification)
+    }
+
+    override fun getActiveNotificationMessagingStyleByTag(tag: String): MessagingStyle? {
+        val activeNotification = NotificationManagerCompat.from(context)
+            .activeNotifications
+            .find { it.tag == tag }
+        activeNotification?.notification?.let { notification ->
+            NotificationCompat.MessagingStyle
+                .extractMessagingStyleFromNotification(notification)
+                ?.let { style ->
+                    return MessagingStyle(
+                        toPigeonPerson(style.user),
+                        style.conversationTitle!!.toString(),
+                        style.messages.map { MessagingStyleMessage(
+                            it.text!!.toString(),
+                            it.timestamp,
+                            toPigeonPerson(it.person!!)
+                        ) },
+                        style.isGroupConversation,
+                    )
+                }
+        }
+        return null
     }
 }
 
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -1,10 +1,11 @@
 import 'dart:convert';
 
+import 'package:http/http.dart' as http;
 import 'package:collection/collection.dart';
 import 'package:crypto/crypto.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
-import 'package:flutter_local_notifications/flutter_local_notifications.dart';
+import 'package:flutter_local_notifications/flutter_local_notifications.dart' hide Person;
 
 import '../api/notifications.dart';
 import '../host/android_notifications.dart';
@@ -92,7 +93,36 @@ class NotificationDisplayManager {
   static Future<void> _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) async {
     assert(debugLog('notif message content: ${data.content}'));
     final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
-    final title = switch (data.recipient) {
+    final groupKey = _groupKey(data);
+    final conversationKey = _conversationKey(data, groupKey);
+
+    final oldMessagingStyle = await ZulipBinding.instance.androidNotificationHost
+      .getActiveNotificationMessagingStyleByTag(conversationKey);
+
+    final MessagingStyle messagingStyle;
+    if (oldMessagingStyle != null) {
+      messagingStyle = oldMessagingStyle;
+      messagingStyle.messages =
+        oldMessagingStyle.messages.toList(); // Clone fixed-length list to growable.
+    } else {
+      messagingStyle = MessagingStyle(
+        user: Person(
+          key: _personKey(data.realmUri, data.userId),
+          name: 'You'), // TODO(i18n)
+        messages: [],
+        isGroupConversation: switch (data.recipient) {
+          FcmMessageStreamRecipient() => true,
+          FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 => true,
+          FcmMessageDmRecipient() => false,
+        });
+    }
+
+    // The title typically won't change between messages in a conversation, but we
+    // update it anyway. This means a DM sender's display name gets updated if it's
+    // changed, which is a rare edge case but probably good. The main effect is that
+    // group-DM threads (pending #794) get titled with the latest sender, rather than
+    // the first.
+    messagingStyle.conversationTitle = switch (data.recipient) {
       FcmMessageStreamRecipient(:var streamName?, :var topic) =>
         '#$streamName > $topic',
       FcmMessageStreamRecipient(:var topic) =>
@@ -103,8 +133,14 @@ class NotificationDisplayManager {
       FcmMessageDmRecipient() =>
         data.senderFullName,
     };
-    final groupKey = _groupKey(data);
-    final conversationKey = _conversationKey(data, groupKey);
+
+    messagingStyle.messages.add(MessagingStyleMessage(
+      text: data.content,
+      timestampMs: data.time * 1000,
+      person: Person(
+        key: _personKey(data.realmUri, data.senderId),
+        name: data.senderFullName,
+        iconBitmap: await _fetchBitmap(data.senderAvatarUrl))));
 
     await ZulipBinding.instance.androidNotificationHost.notify(
       // TODO the notification ID can be constant, instead of matching requestCode
@@ -114,12 +150,12 @@ class NotificationDisplayManager {
       channelId: NotificationChannelManager.kChannelId,
       groupKey: groupKey,
 
-      contentTitle: title,
-      contentText: data.content,
       color: kZulipBrandColor.value,
       // TODO vary notification icon for debug
       smallIconResourceName: 'zulip_notification', // This name must appear in keep.xml too: https://github.com/zulip/zulip-flutter/issues/528
-      // TODO(#128) inbox-style
+
+      messagingStyle: messagingStyle,
+      number: messagingStyle.messages.length,
 
       contentIntent: PendingIntent(
         // TODO make intent URLs distinct, instead of requestCode
@@ -196,6 +232,8 @@ class NotificationDisplayManager {
     return "${data.realmUri}|${data.userId}";
   }
 
+  static String _personKey(Uri realmUri, int userId) => "$realmUri|$userId";
+
   static void _onNotificationOpened(NotificationResponse response) async {
     final payload = jsonDecode(response.payload!) as Map<String, dynamic>;
     final data = MessageFcmMessage.fromJson(payload);
@@ -237,5 +275,16 @@ class NotificationDisplayManager {
       // TODO(#82): Open at specific message, not just conversation
       page: MessageListPage(narrow: narrow)));
     return;
+  }
+
+  static Future<Uint8List?> _fetchBitmap(Uri url) async {
+    try {
+      // TODO timeout to prevent waiting indefinitely
+      final resp = await http.get(url);
+      return resp.bodyBytes;
+    } catch (e) {
+      // TODO(log)
+      return null;
+    }
   }
 }
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -108,7 +108,7 @@ class NotificationDisplayManager {
       messagingStyle = MessagingStyle(
         user: Person(
           key: _personKey(data.realmUri, data.userId),
-          name: 'You'), // TODO(i18n)
+          name: zulipLocalizations.notifSelfUser),
         messages: [],
         isGroupConversation: switch (data.recipient) {
           FcmMessageStreamRecipient() => true,
--- a/pigeon/notifications.dart
+++ b/pigeon/notifications.dart
@@ -36,6 +36,64 @@ class InboxStyle {
   final String summaryText;
 }
 
+/// Corresponds to `androidx.core.app.Person`
+///
+/// See: https://developer.android.com/reference/androidx/core/app/Person
+class Person {
+  Person({
+    required this.iconBitmap,
+    required this.key,
+    required this.name,
+  });
+
+  /// An icon for this person.
+  ///
+  /// This should be compressed image data, in a format to be passed
+  /// to `androidx.core.graphics.drawable.IconCompat.createWithData`.
+  /// Supported formats include JPEG, PNG, and WEBP.
+  ///
+  /// See:
+  ///  https://developer.android.com/reference/androidx/core/graphics/drawable/IconCompat#createWithData(byte[],int,int)
+  final Uint8List? iconBitmap;
+
+  final String key;
+  final String name;
+}
+
+/// Corresponds to `androidx.core.app.NotificationCompat.MessagingStyle.Message`
+///
+/// See: https://developer.android.com/reference/androidx/core/app/NotificationCompat.MessagingStyle.Message
+class MessagingStyleMessage {
+  MessagingStyleMessage({
+    required this.text,
+    required this.timestampMs,
+    required this.person,
+  });
+
+  final String text;
+  final int timestampMs;
+  final Person person;
+}
+
+/// Corresponds to `androidx.core.app.NotificationCompat.MessagingStyle`
+///
+/// See: https://developer.android.com/reference/androidx/core/app/NotificationCompat.MessagingStyle
+class MessagingStyle {
+  MessagingStyle({
+    required this.user,
+    required this.conversationTitle,
+    required this.isGroupConversation,
+    required this.messages,
+  });
+
+  final Person user;
+  final String? conversationTitle;
+  // TODO(pigeon): Make list item non-nullable, once pigeon supports non-nullable type arguments.
+  //   https://github.com/flutter/flutter/issues/97848
+  final List<MessagingStyleMessage?> messages;
+  final bool isGroupConversation;
+}
+
 @HostApi()
 abstract class AndroidNotificationHostApi {
   /// Corresponds to `android.app.NotificationManager.notify`,
@@ -73,8 +131,23 @@ abstract class AndroidNotificationHostApi {
     String? groupKey,
     InboxStyle? inboxStyle,
     bool? isGroupSummary,
+    MessagingStyle? messagingStyle,
+    int? number,
     String? smallIconResourceName,
     // NotificationCompat.Builder has lots more methods; add as needed.
     // Keep them alphabetized, for easy comparison with that class's docs.
   });
+
+  /// Wraps `androidx.core.app.NotificationManagerCompat.getActiveNotifications`,
+  /// combined with `androidx.core.app.NotificationCompat.MessagingStyle.extractMessagingStyleFromNotification`.
+  ///
+  /// Returns the messaging style, if any, of an active notification
+  /// that has tag `tag`.  If there are several such notifications,
+  /// an arbitrary one of them is used.
+  /// Returns null if there are no such notifications.
+  ///
+  /// See:
+  ///   https://developer.android.com/reference/kotlin/androidx/core/app/NotificationManagerCompat#getActiveNotifications()
+  ///   https://developer.android.com/reference/kotlin/androidx/core/app/NotificationCompat.MessagingStyle#extractMessagingStyleFromNotification(android.app.Notification)
+  MessagingStyle? getActiveNotificationMessagingStyleByTag(String tag);
 }
--- a/test/model/binding.dart
+++ b/test/model/binding.dart
@@ -3,7 +3,7 @@ import 'dart:async';
 import 'package:firebase_core/firebase_core.dart';
 import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:flutter/foundation.dart';
-import 'package:flutter_local_notifications/flutter_local_notifications.dart';
+import 'package:flutter_local_notifications/flutter_local_notifications.dart' hide Person;
 import 'package:flutter_local_notifications_platform_interface/flutter_local_notifications_platform_interface.dart';
 import 'package:test/fake.dart';
 import 'package:url_launcher/url_launcher.dart' as url_launcher;
@@ -506,6 +506,13 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
   }
   List<AndroidNotificationHostApiNotifyCall> _notifyCalls = [];
 
+  final Map<String, MessagingStyle?> _activeNotificationsMessagingStyle = {};
+
+  /// Clears all active notifications that have been created via [notify].
+  void clearActiveNotifications() {
+    _activeNotificationsMessagingStyle.clear();
+  }
+
   @override
   Future<void> notify({
     String? tag,
@@ -520,6 +527,8 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
     String? groupKey,
     InboxStyle? inboxStyle,
     bool? isGroupSummary,
+    MessagingStyle? messagingStyle,
+    int? number,
     String? smallIconResourceName,
   }) async {
     _notifyCalls.add((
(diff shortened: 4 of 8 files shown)


android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
```diff
@@ -8,10 +8,27 @@ import android.util.Log
...
+    return androidx.core.app.Person.Builder().apply {
+        person.iconData?.let { setIcon(IconCompat.createWithData(it, 0, it.size)) }
+        setKey(person.key)
+        setName(person.name)
+    }.build()
+}
+
+fun toPigeonPerson(person: androidx.core.app.Person): Person {
+    return Person(
+        null, // TODO: Serialize icon maybe using content URI, mData is not exposed.
```
rajveermalviya, line 26: Currently, we don't return Person's IconData back when returning active notification (via `getActiveNotificationMessagingStyleByTag`) because we don't seem to have an API to get data bytes back.

Fortunately though it doesn't seem to be necessary because in my local testing, the avatar is correctly visibile even when same sender has multiple messages, which makes sense because here's what ends up happening:
```dart
MessagingStyle(
  user: Person('You')
  messages: [
    MessagingStyleMessage(content: '...',
      person: Person(key: id1, name: 'John', iconData: null)), // Old, got from active notif
    MessagingStyleMessage(content: '...',
      person: Person(key: id1, name: 'John', iconData: null)),  // Old, got from active notif
    MessagingStyleMessage(content: '...',
      person: Person(key: id1, name: 'John', iconData: [...])), // Latest message, newly added
  ],
  ...
)
```

So, Android seems to be okay with older messages _**not**_ having `iconData` for a specific Person.

Khader-1 (approved):
Thank you, @rajveermalviya! The code looks good to me. The manual test as well as the attached video demonstrate that the behavior described in the issue has been correctly addressed.

lib/notifications/display.dart
```diff
@@ -92,7 +93,29 @@ class NotificationDisplayManager {
...
+    final conversationKey = _conversationKey(data, groupKey);
+
+    final oldMessagingStyle = await ZulipBinding.instance.androidNotificationHost
+      .getActiveNotificationMessagingStyleByTag(conversationKey);
+
+    final MessagingStyle messagingStyle = oldMessagingStyle != null
+      ? MessagingStyle(
+          user: oldMessagingStyle.user,
+          messages: oldMessagingStyle.messages?.toList() ?? [], // Clone a fixed-length list
+          isGroupConversation: oldMessagingStyle.isGroupConversation)
```
chrisbobbe, line 106: Is there a reason to create a new `MessagingStyle` based on `oldMessagingStyle` (but with `conversationTitle` omitted)? What if we just used `oldMessagingStyle`?

I did get a test failure when I tried this, but it's pretty opaque to me:

<details>

```
00:18 +770 ~6 -1: /Users/chrisbobbe/dev/zulip-flutter/test/notifications/display_test.dart: NotificationDisplayManager show multiple stream messages [E]
  Expected: a List<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})> that:
    has length that:
      equals <6>
    contains, in order: [Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void,
    Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void,
    Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void,
    Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void,
    Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void,
    Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void]
  Actual: [(autoCancel: true, channelId: messages-1, color: 4284781310, contentIntent: Instance of 'PendingIntent', contentText: null, contentTitle: null, extras: null, groupKey: https://chat.example/|1084, id: 1266374773, inboxStyle: null, isGroupSummary: null, messagingStyle: Instance of 'MessagingStyle', number: 1, smallIconResourceName: zulip_notification, tag: https://chat.example/|1084|stream:1:example topic),
  (autoCancel: true, channelId: messages-1, color: 4284781310, contentIntent: null, contentText: null, contentTitle: null, extras: null, groupKey: https://chat.example/|1084, id: 712478168, inboxStyle: Instance of 'InboxStyle', isGroupSummary: true, messagingStyle: null, number: null, smallIconResourceName: zulip_notification, tag: https://chat.example/|1084),
  (autoCancel: true, channelId: messages-1, color: 4284781310, contentIntent: Instance of 'PendingIntent', contentText: null, contentTitle: null, extras: null, groupKey: https://chat.example/|1084, id: 1266374773, inboxStyle: null, isGroupSummary: null, messagingStyle: Instance of 'MessagingStyle', number: 2, smallIconResourceName: zulip_notification, tag: https://chat.example/|1084|stream:1:example topic),
  (autoCancel: true, channelId: messages-1, color: 4284781310, contentIntent: null, contentText: null, contentTitle: null, extras: null, groupKey: https://chat.example/|1084, id: 712478168, inboxStyle: Instance of 'InboxStyle', isGroupSummary: true, messagingStyle: null, number: null, smallIconResourceName: zulip_notification, tag: https://chat.example/|1084),
  (autoCancel: true, channelId: messages-1, color: 4284781310, contentIntent: Instance of 'PendingIntent', contentText: null, contentTitle: null, extras: null, groupKey: https://chat.example/|1084, id: 1266374773, inboxStyle: null, isGroupSummary: null, messagingStyle: Instance of 'MessagingStyle', number: 3, smallIconResourceName: zulip_notification, tag: https://chat.example/|1084|stream:1:example topic),
  (autoCancel: true, channelId: messages-1, color: 4284781310, contentIntent: null, contentText: null, contentTitle: null, extras: null, groupKey: https://chat.example/|1084, id: 712478168, inboxStyle: Instance of 'InboxStyle', isGroupSummary: true, messagingStyle: null, number: null, smallIconResourceName: zulip_notification, tag: https://chat.example/|1084)]
  Which: did not have an element matching the expectation at index 0 <Closure: (Subject<({bool? autoCancel, String channelId, int? color, PendingIntent? contentIntent, String? contentText, String? contentTitle, Map<String?, String?>? extras, String? groupKey, int id, InboxStyle? inboxStyle, bool? isGroupSummary, MessagingStyle? messagingStyle, int? number, String? smallIconResourceName, String? tag})>) => void>
  package:checks/src/checks.dart 85:9                check.<fn>
  package:checks/src/checks.dart 708:12              _TestContext.expect
  package:checks/src/extensions/iterable.dart 93:13  IterableChecks.containsInOrder
  test/notifications/display_test.dart 188:11        main.<fn>.checkNotification
  test/notifications/display_test.dart 205:7         main.<fn>.checkNotifications
  test/notifications/display_test.dart 241:13        main.<fn>.<fn>.<fn>
  

To run this test again: /Users/chrisbobbe/.local/lib/flutter/bin/cache/dart-sdk/bin/dart test /Users/chrisbobbe/dev/zulip-flutter/test/notifications/display_test.dart -p vm --plain-name 'NotificationDisplayManager show multiple stream messages'
00:43 +1192 ~168 -1: Some tests failed.     
```

</details>

lib/notifications/display.dart
```diff
@@ -92,7 +93,29 @@ class NotificationDisplayManager {
...
+      : MessagingStyle(
+          user: Person(
+            key: data.userId.toString(),
+            name: 'You'), // TODO(i18n)
+          messages: [],
+          isGroupConversation: switch (data.recipient) {
+            FcmMessageStreamRecipient() => true,
+            FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 => true,
+            FcmMessageDmRecipient() => false,
+        });
```
chrisbobbe, line 116:
```suggestion
          });
```

test/notifications/display_test.dart
```diff
@@ -151,53 +179,86 @@ void main() {
...
 
+    test('multiple stream messages', () => awaitFakeAsync((async) async {
+      await init();
+      final stream = eg.stream(streamId: 1, name: 'stream 1');
+      final message1 = eg.streamMessage(id: 101, stream: stream);
+      final messageData1 = messageFcmMessage(message1, streamName: stream.name);
+      final message2 = eg.streamMessage(id: 102, stream: stream);
+      final messageData2 = messageFcmMessage(message2, streamName: stream.name);
+      final message3 = eg.streamMessage(id: 102, stream: stream);
+      final messageData3 = messageFcmMessage(message3, streamName: stream.name);
```
chrisbobbe, line 239: Are `message2` and `message3` intended to have the same ID?

lib/notifications/display.dart
```diff
@@ -238,4 +263,14 @@ class NotificationDisplayManager {
...
+
+  static Future<Uint8List?> _fetchBitmap(Uri url) async {
+    try {
+      final resp = await http.get(url);
+      return resp.bodyBytes;
+    } catch (e) {
+      // TODO(log)
+      return null;
+    }
+  }
```
chrisbobbe, line 275: I wonder how long this request could sit unresolved, and if we want to time out after some short-ish amount of time. Users will want to see their notifications as soon as we can reasonably show them, even if they're missing an avatar image, I think.

No need to block, this could be a TODO; we haven't implemented this in zulip-mobile.

lib/notifications/display.dart
```diff
@@ -103,8 +126,15 @@ class NotificationDisplayManager {
...
         data.senderFullName,
     };
-    final groupKey = _groupKey(data);
-    final conversationKey = _conversationKey(data, groupKey);
+
+    messagingStyle.messages?.add(MessagingStyleMessage(
+      text: data.content,
+      timestampMs: data.time * 1000,
+      person: Person(
+        key: data.senderId.toString(),
```
chrisbobbe, line 134: Should we add the org URL to the key? There could be users on different realms with the same ID.

chrisbobbe (commented):
Thanks! Small comments below, and marking for Greg's review. He can take a closer look than me because he's more familiar with Android notifications logic.

lib/notifications/display.dart
```diff
@@ -92,7 +93,29 @@ class NotificationDisplayManager {
...
+    final groupKey = _groupKey(data);
+    final conversationKey = _conversationKey(data, groupKey);
+
+    final oldMessagingStyle = await ZulipBinding.instance.androidNotificationHost
+      .getActiveNotificationMessagingStyleByTag(conversationKey);
+
+    final MessagingStyle messagingStyle = oldMessagingStyle != null
+      ? MessagingStyle(
+          user: oldMessagingStyle.user,
+          messages: oldMessagingStyle.messages?.toList() ?? [], // Clone a fixed-length list
```
gnprice, line 105: Instead of copying `oldMessagingStyle`, we can mutate it.

(I guess this overlaps with Chris's question https://github.com/zulip/zulip-flutter/pull/718#discussion_r1663743097 .)

lib/notifications/display.dart
```diff
@@ -92,7 +93,29 @@ class NotificationDisplayManager {
...
+          isGroupConversation: oldMessagingStyle.isGroupConversation)
+      : MessagingStyle(
+          user: Person(
+            key: data.userId.toString(),
```
gnprice, line 109: Hmm, I wonder what the scope of these Person keys is. Do they apply just within a conversation, or across all the app's notifications?

Unless we find clear confirmation of the former, let's conservatively assume the latter. That means we should include the realm URL in this key, as well as the user ID.

… Ha and I see Chris already asked the same question :smile:: https://github.com/zulip/zulip-flutter/pull/718#discussion_r1663781640

lib/notifications/display.dart
```diff
@@ -92,7 +93,29 @@ class NotificationDisplayManager {
...
+            name: 'You'), // TODO(i18n)
+          messages: [],
+          isGroupConversation: switch (data.recipient) {
+            FcmMessageStreamRecipient() => true,
+            FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 => true,
+            FcmMessageDmRecipient() => false,
+        });
+
+    messagingStyle.conversationTitle = switch (data.recipient) {
```
gnprice, line 118: There's a comment above this in the corresponding zulip-mobile code, about the title changing between messages in a conversation, which I think is still relevant.

lib/notifications/display.dart
```diff
@@ -103,8 +126,15 @@ class NotificationDisplayManager {
       FcmMessageDmRecipient() =>
         data.senderFullName,
     };
-    final groupKey = _groupKey(data);
-    final conversationKey = _conversationKey(data, groupKey);
+
+    messagingStyle.messages?.add(MessagingStyleMessage(
```
gnprice, line 130:
```suggestion
    messagingStyle.messages!.add(MessagingStyleMessage(
```

If `messages` here were somehow null, it'd be wrong for us to skip adding this new message.

So if it actually could be null, this should be like `(messagingStyle.messages ??= []).add(…`. But I think in fact it can't.

That also might be a sign that the wrapper API should just take a list here, rather than a nullable list. There's no distinction being expressed by null vs. an empty list, right?

lib/notifications/display.dart
```diff
@@ -114,8 +144,8 @@ class NotificationDisplayManager {
       channelId: NotificationChannelManager.kChannelId,
       groupKey: groupKey,
 
-      contentTitle: title,
-      contentText: data.content,
+      messagingStyle: messagingStyle,
+      number: messagingStyle.messages?.length,
```
gnprice, line 148: nit: put these after the color and icon, like in the corresponding zulip-mobile code

The icon and its color are the first thing the user sees, before unfurling the notification. So they logically come before the contents.

(That would have been clever in the old code too, oh well.)

gnprice: Also there's a TODO comment down there which this commit can delete :slightly_smiling_face:

gnprice (commented):
Thanks @rajveermalviya, and thanks @Khader-1 and @chrisbobbe for the previous reviews! Excited to see this feature.

Comments below; I've read through all but the tests of the main commit. There are also Chris's comments above.

(And in a future round of review I'll also want to do some manual testing myself, to supplement what you and @Khader-1 have done already. I didn't yet in this review.)

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
rajveermalviya: > In your example above, it's the same sender each time. What happens if some of the old senders are different from the one that's latest and has the icon data?

It works for old senders too, (also visible in the PR description video):
Screenshot 2024-07-08 at 00 48 38

pigeon/notifications.dart
```diff
@@ -36,6 +36,55 @@ class InboxStyle {
...
+  MessagingStyle({
+    required this.user,
+    required this.conversationTitle,
+    required this.isGroupConversation,
+    required this.messages,
+  });
+
+  final Person user;
+  final String? conversationTitle;
+  final List<MessagingStyleMessage?>? messages;
```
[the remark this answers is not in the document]
rajveermalviya: Currently pigeon doesn't support non-null type arguments, and throws this error:
```
Error: pigeon/notifications.dart:98: Generic type parameters must be nullable in field "messages" in class "MessagingStyle".
```
https://github.com/flutter/flutter/issues/97848

Added a `TODO(pigeon)` comment explaining it.

lib/notifications/display.dart
rajveermalviya: I wasn't able to find much info if a Person is scoped to a certain groupKey. So for now, updated it to include realm URL in the person key, just in case.

rajveermalviya:
Thanks for the reviews @chrisbobbe and @gnprice, pushed a new revision PTAL.

pigeon/notifications.dart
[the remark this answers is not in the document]
gnprice: Ah great, thanks. And in fact following that link, I see I'd already upvoted that Pigeon issue :slightly_smiling_face:

The nullability here doesn't cause much of a problem directly, as long as the reader knows about it — it's mostly just confusing, when the reader isn't aware of that Pigeon issue and so expects that the nullability would be a choice by the author of this API and therefore should mean something. So a comment giving that context is perfect.

test/model/binding.dart
```diff
@@ -521,9 +528,24 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
       groupKey: groupKey,
       inboxStyle: inboxStyle,
       isGroupSummary: isGroupSummary,
+      messagingStyle: messagingStyle,
+      number: number,
       smallIconResourceName: smallIconResourceName,
     ));
+
+    if (tag != null && messagingStyle != null) {
+      _activeNotificationsMessagingStyle[tag] = MessagingStyle(
```
gnprice, line 537: If `messagingStyle` is null, then this should probably still update the map, right? The current notification with this tag will become a notification that has no messaging style.

lib/notifications/display.dart
```diff
@@ -92,7 +93,35 @@ class NotificationDisplayManager {
...
+          name: zulipLocalizations.notifSelfUser),
+        messages: [],
+        isGroupConversation: switch (data.recipient) {
+          FcmMessageStreamRecipient() => true,
+          FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 => true,
+          FcmMessageDmRecipient() => false,
+        });
+    }
+
+    // For stream messages, the title remains constant. For DMs, it may
```
gnprice, line 120: Hmm, why do you say it remains constant? It can change if the channel/stream gets renamed, right?

I think the comment in zulip-mobile is accurate here too. It just needs "PM" modernized to read "DM", and the reference to https://github.com/zulip/zulip-mobile/issues/5116 can refer to the existing TODO below… or better yet, file a zulip-flutter issue for better titling group DM threads in notifications, referring back to https://github.com/zulip/zulip-mobile/issues/5116 .

lib/notifications/display.dart
```diff
@@ -92,7 +93,35 @@ class NotificationDisplayManager {
...
+    // change if the DM sender's display name updates. For group DMs,
```
gnprice, line 121: (And I'm not sure what "if the DM sender's display name updates" means. A conversation may have several DM senders.)

lib/notifications/display.dart
```diff
@@ -114,12 +149,11 @@ class NotificationDisplayManager {
       channelId: NotificationChannelManager.kChannelId,
       groupKey: groupKey,
 
-      contentTitle: title,
-      contentText: data.content,
       color: kZulipBrandColor.value,
       // TODO vary notification icon for debug
       smallIconResourceName: 'zulip_notification', // This name must appear in keep.xml too: https://github.com/zulip/zulip-flutter/issues/528
-      // TODO(#128) inbox-style
+      messagingStyle: messagingStyle,
```
gnprice, line 155: nit: blank line to separate these

gnprice (commented):
Thanks for the revision! Comments below. For this round I focused only on the areas I reviewed last time.

rajveermalviya:
Thanks for the review @gnprice, pushed a new revision — PTAL.

test/notifications/display_test.dart
```diff
@@ -106,26 +108,54 @@ void main() {
...
-      final expectedId =
-        NotificationDisplayManager.notificationIdAsHashOf(expectedTag);
       const expectedIntentFlags =
         PendingIntentFlag.immutable | PendingIntentFlag.updateCurrent;
 
-      check(testBinding.androidNotificationHost.takeNotifyCalls())
-        ..length.equals(2)
-        ..containsInOrder(<Condition<AndroidNotificationHostApiNotifyCall>>[
-          (it) => it
+      final notifyCallsChecks = <Condition<AndroidNotificationHostApiNotifyCall>>[];
```
gnprice, line 119: nit:
```suggestion
      final notifyCallsChecks = <Condition<Object?>>[];
```

This doesn't directly affect any behavior, because the individual items already have type `Condition<Object?>` thanks to the annotation on the list literal that introduces them. But it's easier to think about when the different parts are consistent.

gnprice (commented):
Thanks for the revision! From reading only the changes and my previous review, just one nit below.

@chrisbobbe might have follow-ups on his review; I haven't checked the state of each of those subthreads. I'll also want to review the remaining parts I haven't yet (per https://github.com/zulip/zulip-flutter/pull/718#pullrequestreview-2157272887 ); just leaving this reply first as a checkpoint.

chrisbobbe:
> @chrisbobbe might have follow-ups on his review; I haven't checked the state of each of those subthreads.

Just checked these; no follow-ups from me. 🙂

rajveermalviya:
Thanks for the reviews, just pushed a new revision addressing the one nit.

test/notifications/display_test.dart
```diff
@@ -106,26 +108,54 @@ void main() {
...
+        // List of all the checks for messages on each notify calls.
+        final messagesChecks = <Condition<Object?>>[];
+        for (int j = 0; j < (i + 1); j++) {
+          final data = messages[j];
+          final expectedSenderKey = '${data.realmUri}|${data.senderId}';
+          messagesChecks.add((it) => it.isA<MessagingStyleMessage>()
+            ..text.equals(data.content)
+            ..timestampMs.equals(data.time * 1000)
+            ..person.which((it) => it.isNotNull()
+              ..iconBitmap.isNotNull()
```
gnprice, line 138: Huh, what does the sender icon look like? I'd think we'd have to set up the test HTTP client for that fetch to not give an error.

(Is it returning an error response or something, and we're getting the bytes of that body? If so then that seems like a bug in the code under test, which we should fix — we only actually want the response body bytes if the HTTP status was 200.)

It'd be good to, for at least one test case, set up the test HTTP client to return a known value at the expected avatar URL, and then check that we get that value.

test/notifications/display_test.dart
```diff
@@ -151,53 +181,87 @@ void main() {
...
         expectedTitle: '#${stream.name} > ${message.topic}',
         expectedTagComponent: 'stream:${message.streamId}:${message.topic}');
     }));
 
+    test('multiple stream messages', () => awaitFakeAsync((async) async {
+      await init();
+      final stream = eg.stream(streamId: 1, name: 'stream 1');
+      final message1 = eg.streamMessage(id: 101, stream: stream, timestamp: 1);
+      final message2 = eg.streamMessage(id: 102, stream: stream, timestamp: 2);
+      final message3 = eg.streamMessage(id: 103, stream: stream, timestamp: 3);
```
gnprice, line 238: This test is relying on all three of these being in the same conversation. So it should make that explicit in the data, by specifying the topic.

OTOH I don't think there's a need to specify the message IDs. See the `eg.streamMessage` doc. If we don't need to specify them, then best to leave them out, because they're totally boring as far as this test is concerned.

gnprice: Similarly the stream ID and name to `eg.stream`.

test/model/binding.dart
```diff
@@ -521,9 +530,25 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
...
+      number: number,
       smallIconResourceName: smallIconResourceName,
     ));
+
+    if (tag != null) {
+      _activeNotificationsMessagingStyle[tag] = messagingStyle != null
+        ? MessagingStyle(
+            user: messagingStyle.user,
+            conversationTitle: messagingStyle.conversationTitle,
+            messages: messagingStyle.messages.toList(), // clone to avoid shared-references
```
gnprice, line 543: Oh, this should remove icons from the senders. That way it accurately simulates the behavior of our Kotlin code.

gnprice (commented):
Thanks again @rajveermalviya! I've now read through the whole thing.

Comments below, almost all of them on the test code. One other high-level comment is that this needs more tests.

As is, this updates the existing tests, and it adds one important test case which is getting several messages in a row in a single conversation. But it doesn't have tests to cover some important other areas of the changes.

Other test cases that should be added include:
* Fetching a sender avatar, successfully. (This is touched on in one of the comments below.) This might be made into another thing the existing tests check, instead of its own separate test case.
* Fetching a sender avatar but failing.
* Getting a message in conversation A, then conversation B, then A again — the history should include the previous A message but not the B message.
* Getting several messages in a group DM thread, from different senders — the title should reflect the latest sender.
* Getting several messages in a channel thread, where the channel's name changes midway. They should remain one conversation, and the title should update.
* Getting several messages in a 1:1 DM thread, where the sender's name changes midway. The title should update.
* Getting several messages in a 1:1 DM thread, where the sender's *email* changes midway. They should remain one conversation. (This confirms we're threading by sender ID, not sender email.)

---

In order to write those cases in a way where it's easy for the reader to follow the narrative and see what's being tested, I think it'll be important to make the test cases' code look like: receive a MessageFcmMessage; check that `takeNotifyCalls()` looks as expected; receive another MessageFcmMessage; check `takeNotifyCalls()` again; etc.

That structure isn't really compatible with the outer `checkNotifications` helper; but I think it's quite compatible with using the inner `checkNotification` helper for the checks, in something like its original form where it expects just two `notify` calls. (Pass it a `List<MessageFcmMessage>` that's just the ones that should be reflected in the newly-updated notification.)

I think a good solution is that the existing test cases can continue using `checkNotifications` — and then that's enough for making the point that `checkNotifications` exists to make, which is just that both foreground and background messages are wired up. The new test cases, the ones that involve receiving several notification messages, can use the inner `checkNotification` helper directly. Perhaps also a small new test helper for receiving a notification message, so that the test cases look like:
```dart
  receiveFcmMessage(fcmMessage3);
  checkNotifications(async,
    [fcmMessage1, fcmMessage2, fcmMessage3],
    expectedFoo: …);
```

Then the outer `checkNotifications` helper goes back to taking just one message, like in main, and can call `checkNotification([data], expectedFoo: …)`.

---

Thinking about sequencing: those tests are important, but this is already by nature a complex PR, and it'd be good to get it merged soon. So let's do this:
* Please handle the comments below, except one or two of them that are about adding more test cases.
  * Also do the `checkNotification`/`checkNotifications` refactor described just above, since that will simplify the diff in this PR. So the one new test case will use `checkNotification` directly.
* Then please start work on those additional test cases, but as a followup PR to go atop this one.
* I'll also give this PR some manual testing, which is the one remaining area of review I haven't done.
* When those other comments are resolved, and provided manual testing looks good, I'll go ahead and merge.
* Please treat that followup PR adding tests as a priority. Generally we want tests in the same commit as the code they're testing, so in splitting up the PR like this I'm counting on the tests getting completed soon after.

How's that sound?

test/notifications/display_test.dart
rajveermalviya: The fake http client seems to return an empty list for `Response.bodyBytes` here, so this turns out to be a non-null empty list.

----

I'll add the fake http client infra in the followup PR with the above mentioned tests.

rajveermalviya:
Thanks for the review, @gnprice. I've pushed a revision that addresses your comments and includes the `checkNotification`/`checkNotifications` refactor.

Next, I'll send a follow-up PR that will include all the test cases mentioned, as well as the HTTP client test mentioned in one of the comments.

test/notifications/display_test.dart
```diff
@@ -166,30 +197,79 @@ void main() {
...
+        expectedIsGroupConversation: true,
+        expectedTitle: '#${stream.name} > ${message2.topic}',
+        expectedTagComponent: 'stream:${message2.streamId}:${message2.topic}');
+
+      await receiveFcmMessage(async, data3);
+      checkNotification(data3,
+        messageStyleMessages: [data1, data2, data3],
+        expectedIsGroupConversation: true,
+        expectedTitle: '#${stream.name} > ${message3.topic}',
+        expectedTagComponent: 'stream:${message3.streamId}:${message3.topic}');
```
gnprice, line 263: Similar to https://github.com/zulip/zulip-flutter/pull/718#discussion_r1673223547 : this test is relying on all three of these messages being in the same stream and topic. (If they weren't, then the `messageStyleMessages: [data1, data2, data3]` above wouldn't be the right expectation.) So let's make that explicit in the structure of the test:
```suggestion
        expectedTitle: '#${stream.name} > $topic',
        expectedTagComponent: 'stream:${stream.streamId}:$topic');
```

by expressing the topic and stream ID the same way each time.

Or probably better yet: pull out locals for `expectedTitle` and `expectedTagComponent` and compute them just once.

test/notifications/display_test.dart
```diff
@@ -106,26 +108,54 @@ void main() {
...
-        ..containsInOrder(<Condition<AndroidNotificationHostApiNotifyCall>>[
-          (it) => it
+      final notifyCallsChecks = <Condition<Object?>>[];
+      for (int i = 0; i < messages.length; i++) {
+        final data = messages[i];
+        final expectedTag =
+          '${data.realmUri}|${data.userId}|$expectedTagComponent';
+        final expectedGroupKey = '${data.realmUri}|${data.userId}';
+        final expectedId =
+          NotificationDisplayManager.notificationIdAsHashOf(expectedTag);
```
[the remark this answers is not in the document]
gnprice: bump on this part:

> Then above that it should have an assert that verifies the assumption.

(I.e., the assert should verify that `realmUri` and `userId` are the same in all the given messages.)

gnprice:
Thanks @rajveermalviya for the revision! This is very close to merge now. Comments above, and looking forward to that follow-up PR with more tests too.

rajveermalviya:

rajveermalviya:
Had a rebase conflict; fixed now.

gnprice (commented):
Thanks @rajveermalviya for building this and for all the revisions! Looks great — merging. And I see you've just sent #813 as a draft of that followup :smiley: 

(I did end up testing this end-to-end, mainly by having this PR in the builds I put on my phone the last couple of weeks. Glad to have the feature.)

test/notifications/display_test.dart
```diff
@@ -107,25 +110,53 @@ void main() {
...
       final expectedId =
         NotificationDisplayManager.notificationIdAsHashOf(expectedTag);
       const expectedIntentFlags =
         PendingIntentFlag.immutable | PendingIntentFlag.updateCurrent;
+      final expectedSelfUserKey = '${data.realmUri}|${data.userId}';
+
+      final messageStyleMessagesChecks =
+        messageStyleMessages.mapIndexed((i, messageData) {
+          assert(messageData.realmUri == data.realmUri);
+          assert(messageData.userId == data.userId);
```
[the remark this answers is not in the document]
rajveermalviya: Added this as an `nfc` in #813

Issue #719, opened 2024-06-04
Ban implicit dynamic


gnprice:
The Dart analyzer has options for preventing `dynamic` from appearing in the code accidentally:
https://dart.dev/tools/analysis#enabling-additional-type-checks
They look like this in `analysis_options.yaml`:
```yaml
analyzer:
  language:
    strict-casts: true
    strict-inference: true
    strict-raw-types: true
```

The Flutter tree enables those options. We should too — we've often had little surprises where something turns out to have been `dynamic` because we were a bit sloppy in writing down types. (No live bugs yet, as far as I can recall, but clearly the potential for some.)

PR #721, opened 2024-06-04
Enable `strict-inference`

gnprice:
This prevents many situations where a type would otherwise implicitly be `dynamic`.  See docs:
  https://dart.dev/tools/analysis#enabling-additional-type-checks
  https://github.com/dart-lang/language/blob/dd11c4eac/resources/type-system/strict-inference.md

Fixes-partly: #719


chrisbobbe:
Cool! LGTM; merged.

PR #722, opened 2024-06-05
lint: Enable strict-raw-types

gnprice:
This complements `strict-inference` to prevent many more situations
where a type would otherwise implicitly be `dynamic`.  See docs:
  https://dart.dev/tools/analysis#enabling-additional-type-checks
  https://github.com/dart-lang/language/blob/dd11c4eac/resources/type-system/strict-raw-types.md

This PR is stacked atop #721.

Fixes-partly: #719


chrisbobbe:
Thanks, LGTM! Merged.

PR #723, opened 2024-06-05
lint: Enable strict-casts

gnprice:
This prevents implicitly casting from `dynamic` to another type. Where the use of `dynamic` was intended, that helps us make the cast explicit; it also turns out to find a few places where it wasn't intended.

A couple of those places where it wasn't intended seem like they should have been flagged by one of the analyzer options previously enabled in this series: `strict-inference` and `strict-raw-types`. So there may be some remaining implicit `dynamic` types that also slipped through those, and don't happen to have triggered this. Still, these three together should greatly reduce our number of implicit `dynamic` types.

(Where using `dynamic` is intended, that's often only due to the constraints of an external API, like in JSON decoding; I think a preferable version of that API would use `Object?` instead. I'm not sure there's anywhere in our codebase where we'd actually want `dynamic` absent constraints like that one.)

This PR is stacked atop #722, which is atop #721.

Fixes: #719

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,5 +1,5 @@
 # For docs on this file, see:
-#   https://dart.dev/guides/language/analysis-options
+#   https://dart.dev/tools/analysis
 include: package:flutter_lints/flutter.yaml
 
 analyzer:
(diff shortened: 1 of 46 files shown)


chrisbobbe:
Great; thanks for enabling all these rules for #719! LGTM; merged.

PR #700, opened 2024-05-22
content test: Skip source-file-reading test on Windows

gnprice:
This fails there, or at least does in the `flutter/tests` CI environment, with an error like:

    ::group::❌ C:/Users/RUNNER~1/AppData/Local/Temp/flutter_customer_testing.zulip.4a7c3de6/tests/test/model/content_test.dart: all content examples are tested (failed)
    FileSystemException: Cannot open file, path = '/C:/Users/RUNNER~1/AppData/Local/Temp/flutter_customer_testing.zulip.4a7c3de6/tests/test/model/content_test.dart' (OS Error: The filename, directory name, or volume label syntax is incorrect.
    , errno = 123)
    dart:io                               _File.readAsStringSync
    test\model\content_test.dart 1082:39  main.<fn>
    ::endgroup::

I'm not sure what's wrong, but don't really feel like finding out (or working out how to fix it) -- and, conveniently, there's really no need to do so.  See comment.

---

This is needed for #239. See https://github.com/flutter/tests/pull/368 and its CI result at https://github.com/flutter/tests/actions/runs/9197725826/job/25298839228?pr=368#step:3:9762 .

PR #724, opened 2024-06-05
api: Embed platform and app info in user-agent

### LINKED ISSUE #467: In user-agent, include platform, platform version, and app version

rajveermalviya:
Fixes #467

### DIFF
--- a/lib/api/core.dart
+++ b/lib/api/core.dart
@@ -1,9 +1,11 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:http/http.dart' as http;
 
 import '../log.dart';
+import '../model/binding.dart';
 import '../model/localizations.dart';
 import 'exception.dart';
 
@@ -37,6 +39,7 @@ class ApiConnection {
     String? email,
     String? apiKey,
     required http.Client client,
+    required this.useBinding,
   }) : assert((email != null) == (apiKey != null)),
        _authValue = (email != null && apiKey != null)
          ? _authHeaderValue(email: email, apiKey: apiKey)
@@ -51,7 +54,7 @@ class ApiConnection {
     String? apiKey,
   }) : this(client: http.Client(),
             realmUrl: realmUrl, zulipFeatureLevel: zulipFeatureLevel,
-            email: email, apiKey: apiKey);
+            email: email, apiKey: apiKey, useBinding: true);
 
   final Uri realmUrl;
 
@@ -68,6 +71,36 @@ class ApiConnection {
   /// See:
   ///  * API docs at <https://zulip.com/api/changelog>.
   int? zulipFeatureLevel;
+
+  /// Toggles the use of a user-agent generated via [ZulipBinding].
+  ///
+  /// When set to true, the user-agent will be generated using
+  /// [ZulipBinding.deviceInfo] and [ZulipBinding.packageInfo].
+  /// Otherwise, a fallback user-agent [kFallbackUserAgentHeader] will be used.
+  final bool useBinding;
+
+  Map<String, String>? _cachedUserAgentHeader;
+
+  void addUserAgent(http.BaseRequest request) {
+    if (!useBinding) {
+      request.headers.addAll(kFallbackUserAgentHeader);
+      return;
+    }
+
+    if (_cachedUserAgentHeader != null) {
+      request.headers.addAll(_cachedUserAgentHeader!);
+      return;
+    }
+
+    final deviceInfo = ZulipBinding.instance.syncDeviceInfo;
+    final packageInfo = ZulipBinding.instance.syncPackageInfo;
+    if (deviceInfo == null || packageInfo == null) {
+      request.headers.addAll(kFallbackUserAgentHeader);
+      return;
+    }
+    _cachedUserAgentHeader = _buildUserAgentHeader(deviceInfo, packageInfo);
+    request.headers.addAll(_cachedUserAgentHeader!);
+  }
 
   final String? _authValue;
 
@@ -88,9 +121,10 @@ class ApiConnection {
     assert(debugLog("${request.method} ${request.url}"));
 
     addAuth(request);
-    request.headers.addAll(userAgentHeader());
     if (overrideUserAgent != null) {
       request.headers['User-Agent'] = overrideUserAgent;
+    } else {
+      addUserAgent(request);
     }
 
     final http.StreamedResponse response;
@@ -213,10 +247,47 @@ Map<String, String> authHeader({required String email, required String apiKey}) 
   };
 }
 
+/// Fallback user-agent header.
+///
+/// See documentation on [ApiConnection.useBinding].
+@visibleForTesting
+const kFallbackUserAgentHeader = {'User-Agent': 'ZulipFlutter'};
+
 Map<String, String> userAgentHeader() {
+  final deviceInfo = ZulipBinding.instance.syncDeviceInfo;
+  final packageInfo = ZulipBinding.instance.syncPackageInfo;
+  if (deviceInfo == null || packageInfo == null) {
+    return kFallbackUserAgentHeader;
+  }
+  return _buildUserAgentHeader(deviceInfo, packageInfo);
+}
+
+Map<String, String> _buildUserAgentHeader(BaseDeviceInfo deviceInfo, PackageInfo packageInfo) {
+  final osInfo = switch (deviceInfo) {
+    AndroidDeviceInfo(
+      :var release)       => 'Android $release', // "Android 14"
+    IosDeviceInfo(
+      :var systemVersion) => 'iOS $systemVersion', // "iOS 17.4"
+    MacOsDeviceInfo(
+      :var majorVersion,
+      :var minorVersion,
+      :var patchVersion)  => 'macOS $majorVersion.$minorVersion.$patchVersion', // "macOS 14.5.0"
+    WindowsDeviceInfo()   => 'Windows', // "Windows"
+    LinuxDeviceInfo(
+      :var name,
+      :var versionId)     => 'Linux; $name${versionId != null ? ' $versionId' : ''}', // "Linux; Fedora Linux 40" or "Linux; Fedora Linux"
+    _                     => throw UnimplementedError(),
+  };
+  final PackageInfo(:version, :buildNumber) = packageInfo;
+
+  // Possible examples:
+  //  'ZulipFlutter/0.0.15+15 (Android 14)'
+  //  'ZulipFlutter/0.0.15+15 (iOS 17.4)'
+  //  'ZulipFlutter/0.0.15+15 (macOS 14.5.0)'
+  //  'ZulipFlutter/0.0.15+15 (Windows)'
+  //  'ZulipFlutter/0.0.15+15 (Linux; Fedora Linux 40)'
   return {
-    // TODO(#467) include platform, platform version, and app version
-    'User-Agent': 'ZulipFlutter',
+    'User-Agent': 'ZulipFlutter/$version+$buildNumber ($osInfo)',
   };
 }
 
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -13,8 +13,8 @@ void main() {
     return true;
   }());
   LicenseRegistry.addLicense(additionalLicenses);
-  LiveZulipBinding.ensureInitialized();
   WidgetsFlutterBinding.ensureInitialized();
+  LiveZulipBinding.ensureInitialized();
   NotificationService.instance.start();
   runApp(const ZulipApp());
 }
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -156,7 +156,7 @@ abstract class ZulipBinding {
 
 /// Like [device_info_plus.BaseDeviceInfo], but without things we don't use.
 abstract class BaseDeviceInfo {
-  BaseDeviceInfo();
+  const BaseDeviceInfo();
 }
 
 /// Like [device_info_plus.AndroidDeviceInfo], but without things we don't use.
@@ -175,7 +175,7 @@ class AndroidDeviceInfo extends BaseDeviceInfo {
   ///   https://developer.android.com/reference/android/os/Build.VERSION_CODES.html
   final int sdkInt;
 
-  AndroidDeviceInfo({required this.release, required this.sdkInt});
+  const AndroidDeviceInfo({required this.release, required this.sdkInt});
 }
 
 /// Like [device_info_plus.IosDeviceInfo], but without things we don't use.
@@ -185,7 +185,7 @@ class IosDeviceInfo extends BaseDeviceInfo {
   /// See: https://developer.apple.com/documentation/uikit/uidevice/1620043-systemversion
   final String systemVersion;
 
-  IosDeviceInfo({required this.systemVersion});
+  const IosDeviceInfo({required this.systemVersion});
 }
 
 /// Like [device_info_plus.MacOsDeviceInfo], but without things we don't use.
@@ -199,7 +199,7 @@ class MacOsDeviceInfo extends BaseDeviceInfo {
   /// See: https://developer.apple.com/documentation/foundation/operatingsystemversion/1415564-patchversion
   final int patchVersion;
 
-  MacOsDeviceInfo({
+  const MacOsDeviceInfo({
     required this.majorVersion,
     required this.minorVersion,
     required this.patchVersion,
@@ -228,7 +228,9 @@ class MacOsDeviceInfo extends BaseDeviceInfo {
 //      makes it less ideal.
 //  [1]: https://gist.github.com/rajveermalviya/58b3add437280cc7f8356f3697099b7c
 //  [2]: https://github.com/zulip/zulip-flutter/pull/724#discussion_r1628318991
-class WindowsDeviceInfo implements BaseDeviceInfo {}
+class WindowsDeviceInfo implements BaseDeviceInfo {
+  const WindowsDeviceInfo();
+}
 
 /// Like [device_info_plus.LinuxDeviceInfo], but without things we don't use.
 class LinuxDeviceInfo implements BaseDeviceInfo {
@@ -249,7 +251,7 @@ class LinuxDeviceInfo implements BaseDeviceInfo {
   /// See: https://www.freedesktop.org/software/systemd/man/latest/os-release.html#VERSION_ID=
   final String? versionId;
 
-  LinuxDeviceInfo({required this.name, required this.versionId});
+  const LinuxDeviceInfo({required this.name, required this.versionId});
 }
 
 /// Like [package_info_plus.PackageInfo], but without things we don't use.
@@ -257,7 +259,7 @@ class PackageInfo {
   final String version;
   final String buildNumber;
 
-  PackageInfo({
+  const PackageInfo({
     required this.version,
     required this.buildNumber,
   });
@@ -329,7 +331,7 @@ class LiveZulipBinding extends ZulipBinding {
         device_info_plus.MacOsDeviceInfo()   => MacOsDeviceInfo(majorVersion: info.majorVersion,
                                                                 minorVersion: info.minorVersion,
                                                                 patchVersion: info.patchVersion),
-        device_info_plus.WindowsDeviceInfo() => WindowsDeviceInfo(),
+        device_info_plus.WindowsDeviceInfo() => const WindowsDeviceInfo(),
         device_info_plus.LinuxDeviceInfo()   => LinuxDeviceInfo(name: info.name,
                                                                 versionId: info.versionId),
         _                                    => throw UnimplementedError(),
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -6,6 +6,7 @@ import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 import 'package:url_launcher/url_launcher.dart' as url_launcher;
 
 import '../host/android_notifications.dart';
+import '../log.dart';
 import '../widgets/store.dart';
 import 'store.dart';
 
@@ -101,8 +102,18 @@ abstract class ZulipBinding {
   /// Provides device and operating system information,
   /// via package:device_info_plus.
   ///
+  /// The returned Future resolves to null if an error is
+  /// encountered while fetching the data.
+  ///
   /// This wraps [device_info_plus.DeviceInfoPlugin.deviceInfo].
-  Future<BaseDeviceInfo> deviceInfo();
+  Future<BaseDeviceInfo?> get deviceInfo;
+
+  /// Provides device and operating system information,
+  /// via package:device_info_plus.
+  ///
+  /// This is the value [deviceInfo] resolved to,
+  /// or null if that hasn't resolved yet.
+  BaseDeviceInfo? get syncDeviceInfo;
 
   /// Initialize Firebase, to use for notifications.
   ///
@@ -161,6 +172,10 @@ class IosDeviceInfo extends BaseDeviceInfo {
 /// Methods wrapping a plugin, like [launchUrl], invoke the actual
 /// underlying plugin method.
 class LiveZulipBinding extends ZulipBinding {
+  LiveZulipBinding() {
+    _deviceInfo = _prefetchDeviceInfo();
+  }
+
   /// Initialize the binding if necessary, and ensure it is a [LiveZulipBinding].
   static LiveZulipBinding ensureInitialized() {
     if (ZulipBinding._instance == null) {
@@ -196,13 +211,25 @@ class LiveZulipBinding extends ZulipBinding {
   }
 
   @override
-  Future<BaseDeviceInfo> deviceInfo() async {
-    final deviceInfo = await device_info_plus.DeviceInfoPlugin().deviceInfo;
-    return switch (deviceInfo) {
-      device_info_plus.AndroidDeviceInfo(:var version)   => AndroidDeviceInfo(sdkInt: version.sdkInt),
-      device_info_plus.IosDeviceInfo(:var systemVersion) => IosDeviceInfo(systemVersion: systemVersion),
-      _                                                  => throw UnimplementedError(),
-    };
+  Future<BaseDeviceInfo?> get deviceInfo => _deviceInfo;
+  late Future<BaseDeviceInfo?> _deviceInfo;
+
+  @override
+  BaseDeviceInfo? get syncDeviceInfo => _syncDeviceInfo;
+  BaseDeviceInfo? _syncDeviceInfo;
+
+  Future<BaseDeviceInfo?> _prefetchDeviceInfo() async {
+    try {
+      final info = await device_info_plus.DeviceInfoPlugin().deviceInfo;
+      _syncDeviceInfo = switch (info) {
+        device_info_plus.AndroidDeviceInfo(:var version)   => AndroidDeviceInfo(sdkInt: version.sdkInt),
+        device_info_plus.IosDeviceInfo(:var systemVersion) => IosDeviceInfo(systemVersion: systemVersion),
+        _                                                  => throw UnimplementedError(),
+      };
+    } catch (e, st) {
+      assert(debugLog('Failed to prefetch device info: $e\n$st')); // TODO(log)
+    }
+    return _syncDeviceInfo;
   }
 
   @override
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -161,13 +161,21 @@ abstract class BaseDeviceInfo {
 
 /// Like [device_info_plus.AndroidDeviceInfo], but without things we don't use.
 class AndroidDeviceInfo extends BaseDeviceInfo {
+  /// The Android version string, Build.VERSION.RELEASE, e.g. "14".
+  ///
+  /// Upstream documents this as an opaque string with no particular structure,
+  /// but e.g. on stock Android 14 it's "14".
+  ///
+  /// See: https://developer.android.com/reference/android/os/Build.VERSION#RELEASE
+  final String release;
+
   /// The Android SDK version.
   ///
   /// Possible values are defined in:
   ///   https://developer.android.com/reference/android/os/Build.VERSION_CODES.html
   final int sdkInt;
 
-  AndroidDeviceInfo({required this.sdkInt});
+  AndroidDeviceInfo({required this.release, required this.sdkInt});
 }
 
 /// Like [device_info_plus.IosDeviceInfo], but without things we don't use.
@@ -178,6 +186,70 @@ class IosDeviceInfo extends BaseDeviceInfo {
   final String systemVersion;
 
   IosDeviceInfo({required this.systemVersion});
+}
+
+/// Like [device_info_plus.MacOsDeviceInfo], but without things we don't use.
+class MacOsDeviceInfo extends BaseDeviceInfo {
+  /// See: https://developer.apple.com/documentation/foundation/operatingsystemversion/1414662-majorversion
+  final int majorVersion;
+
+  /// See: https://developer.apple.com/documentation/foundation/operatingsystemversion/1413801-minorversion
+  final int minorVersion;
+
+  /// See: https://developer.apple.com/documentation/foundation/operatingsystemversion/1415564-patchversion
+  final int patchVersion;
+
+  MacOsDeviceInfo({
+    required this.majorVersion,
+    required this.minorVersion,
+    required this.patchVersion,
+  });
+}
+
+/// Like [device_info_plus.WindowsDeviceInfo], currently only used to
+/// determine if we're on Windows.
+// TODO Determine a method to identify the Windows version.
+//  Currently, we do not include Windows version information because
+//  Windows OS does not provide a straightforward way to obtain
+//  recognizable version information.
+//  Here's an example of `WindowsDeviceInfo` data[1]. Based on that
+//  data, there are two possible approaches to identify the Windows
+//  version:
+//    - One approach is to use a combination of the majorVersion,
+//      minorVersion, and buildNumber fields. However, this data does
+//      not directly correspond to recognizable Windows versions
+//      (for example major=10, minor=0, build=22631 actually represents
+//      "Windows 11, 23H2"). Refer to the link in this comment[2] for
+//      Chromium's implementation of parsing Windows version numbers.
+//    - Another approach is to use the productName field. While this
+//      field contains the Windows version, it also includes extraneous
+//      information. For example, some productName strings are:
+//      "Windows 11 Pro" and "Windows 10 Home Single Language", which
+//      makes it less ideal.
+//  [1]: https://gist.github.com/rajveermalviya/58b3add437280cc7f8356f3697099b7c
+//  [2]: https://github.com/zulip/zulip-flutter/pull/724#discussion_r1628318991
+class WindowsDeviceInfo implements BaseDeviceInfo {}
+
+/// Like [device_info_plus.LinuxDeviceInfo], but without things we don't use.
+class LinuxDeviceInfo implements BaseDeviceInfo {
+  /// The operating system name, 'NAME' field in /etc/os-release.
+  ///
+  /// Examples: 'Fedora', 'Debian GNU/Linux', or just 'Linux'.
+  ///
+  /// See: https://www.freedesktop.org/software/systemd/man/latest/os-release.html#NAME=
+  final String name;
+
+  /// The operating system version, 'VERSION_ID' field in /etc/os-release.
+  ///
+  /// This string contains only the version number and excludes the
+  /// OS name and version codenames.
+  ///
+  /// Examples: '17', '11.04'.
+  ///
+  /// See: https://www.freedesktop.org/software/systemd/man/latest/os-release.html#VERSION_ID=
+  final String? versionId;
+
+  LinuxDeviceInfo({required this.name, required this.versionId});
 }
 
 /// Like [package_info_plus.PackageInfo], but without things we don't use.
@@ -251,9 +323,16 @@ class LiveZulipBinding extends ZulipBinding {
     try {
       final info = await device_info_plus.DeviceInfoPlugin().deviceInfo;
       _syncDeviceInfo = switch (info) {
-        device_info_plus.AndroidDeviceInfo(:var version)   => AndroidDeviceInfo(sdkInt: version.sdkInt),
-        device_info_plus.IosDeviceInfo(:var systemVersion) => IosDeviceInfo(systemVersion: systemVersion),
-        _                                                  => throw UnimplementedError(),
+        device_info_plus.AndroidDeviceInfo() => AndroidDeviceInfo(release: info.version.release,
+                                                                  sdkInt: info.version.sdkInt),
+        device_info_plus.IosDeviceInfo()     => IosDeviceInfo(systemVersion: info.systemVersion),
+        device_info_plus.MacOsDeviceInfo()   => MacOsDeviceInfo(majorVersion: info.majorVersion,
+                                                                minorVersion: info.minorVersion,
+                                                                patchVersion: info.patchVersion),
+        device_info_plus.WindowsDeviceInfo() => WindowsDeviceInfo(),
+        device_info_plus.LinuxDeviceInfo()   => LinuxDeviceInfo(name: info.name,
+                                                                versionId: info.versionId),
+        _                                    => throw UnimplementedError(),
       };
     } catch (e, st) {
       assert(debugLog('Failed to prefetch device info: $e\n$st')); // TODO(log)
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -3,6 +3,7 @@ import 'package:firebase_core/firebase_core.dart' as firebase_core;
 import 'package:firebase_messaging/firebase_messaging.dart' as firebase_messaging;
 import 'package:flutter/foundation.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
+import 'package:package_info_plus/package_info_plus.dart' as package_info_plus;
 import 'package:url_launcher/url_launcher.dart' as url_launcher;
 
 import '../host/android_notifications.dart';
@@ -115,6 +116,22 @@ abstract class ZulipBinding {
   /// or null if that hasn't resolved yet.
   BaseDeviceInfo? get syncDeviceInfo;
 
+  /// Provides application package information,
+  /// via package:package_info_plus.
+  ///
+  /// The returned Future resolves to null if an error is
+  /// encountered while fetching the data.
+  ///
+  /// This wraps [package_info_plus.PackageInfo.fromPlatform].
+  Future<PackageInfo?> get packageInfo;
+
+  /// Provides application package information,
+  /// via package:package_info_plus.
+  ///
+  /// This is the value [packageInfo] resolved to,
+  /// or null if that hasn't resolved yet.
+  PackageInfo? get syncPackageInfo;
+
   /// Initialize Firebase, to use for notifications.
   ///
   /// This wraps [firebase_core.Firebase.initializeApp].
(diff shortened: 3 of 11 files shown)


lib/api/core.dart
```diff
@@ -214,9 +215,39 @@ Map<String, String> authHeader({required String email, required String apiKey})
...
+    IosDeviceInfo(
+      :var systemVersion) => ('iOS', systemVersion), // "17.4"
+    MacOsDeviceInfo(
+      :var majorVersion,
+      :var minorVersion,
+      :var patchVersion)  => ('macOS', '$majorVersion.$minorVersion.$patchVersion'), // "14.5.0"
+    WindowsDeviceInfo(
+      :var majorVersion,
+      :var minorVersion,
+      :var buildNumber)   => ('Windows', '$majorVersion.$minorVersion $buildNumber'), // "10.0 22631" (actually means Windows 11, 23H2)
```
rajveermalviya, line 235: Unlike other OSes - Windows doesn't seem to provide a way to get a recognizable result from it's provided values, we have to parse the number ourselves. See https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_osversioninfoexw#remarks
For example, [here's how chromium parses it](https://source.chromium.org/chromium/chromium/src/+/main:base/win/windows_version.cc;l=371-470;drc=93b31639a1e56424686adc90b1de30aaec981dfc) (BSD 3-clause licensed).

I felt like rather than doing that guessing work/parsing in the client, we should report concrete numbers, what are you thoughts?

lib/api/core.dart
```diff
@@ -214,9 +215,39 @@ Map<String, String> authHeader({required String email, required String apiKey})
...
   return {
-    // TODO(#467) include platform, platform version, and app version
-    'User-Agent': 'ZulipFlutter',
+    // Possible examples:
+    //  'ZulipFlutter/0.0.15+15 (Android, 14)'
+    //  'ZulipFlutter/0.0.15+15 (iOS, 17.4)'
+    //  'ZulipFlutter/0.0.15+15 (macOS, 14.5.0)'
+    //  'ZulipFlutter/0.0.15+15 (Windows, 10.0 22631)'
+    //  'ZulipFlutter/0.0.15+15 (Linux, Fedora Linux 40)'
+    'User-Agent': 'ZulipFlutter/$appVersion+$appBuildNumber ($osInfo)',
```
rajveermalviya, line 250: Is this format okay?

rajveermalviya (commented):
@gnprice, some questions about the user-agent format, PR as a whole is still WIP.

lib/api/core.dart
gnprice: For Windows I'd prioritize keeping to a minimum the effort and code we spend. So I'd be perfectly happy with reporting no OS version at all, just saying "Windows" with no further details, or with reporting some version string that's easy to get at.

That goes for all three desktop platforms, really — see https://github.com/zulip/zulip-flutter#desktop-support — but especially for Windows, because I'm not sure anyone has even ever run the app there.

As far as I'm aware, the first time any of our code ran on Windows was our test suite when I sent a PR to add us to flutter/tests; see #700, then `flutter test` failed with a mysterious other error, so before merge I skipped Windows in our flutter/tests entry. That means even our tests have been run on Windows a grand total of twice that I'm aware of.

lib/api/core.dart
gnprice: Good question. Let's discuss that on `#api design`, on this thread:
https://chat.zulip.org/#narrow/stream/378-api-design/topic/zulip-flutter.20user.20agent/near/1712676

So please post your example values there and we can get feedback on them.

test/api/fake_api_test.dart
```diff
@@ -2,10 +2,13 @@ import 'package:checks/checks.dart';
 import 'package:test/scaffolding.dart';
 import 'package:zulip/api/exception.dart';
 
+import '../model/binding.dart';
 import 'exception_checks.dart';
 import 'fake_api.dart';
 
 void main() {
+  TestZulipBinding.ensureInitialized();
```
rajveermalviya, line 10: @gnprice Thoughts on testing setup for this?

Because `ApiConnection` depends on `userAgent` function, and it depends on `ZulipBinding` to get package & device info — now every test that uses `ApiConnection` will need to also first call `TestZulipBinding.ensureInitialized()`.

Alternative implementation is to make `ApiConnection` take `userAgent` in it's constructor, which would just move this reposnibility to the upper layer (`LiveGlobalStore` for the real impl), but that would remove the need for `TestZulipBinding.ensureInitialized()` in each test.

gnprice: Hmm yeah, having to do that in all the tests would be annoying.

It looks like there's only a small number of call sites of the ApiConnection constructors. So that's probably a good solution.

rajveermalviya: Hmm after trying the alternative I suggested, ended up reverting it, because it led to a bit of constructor argument poisoning up the chain of `ApiConnection` — making it harder to read the code where some constructors calls take the userAgent value, and some leave it out (null).

Instead ended up refactoring first implementation by moving the `userAgent` function to `class ZulipBinding` to make it explicit that it now depends on members in `ZulipBinding` and thus the binding needs to be initialized to get user-agent.

lib/main.dart
```diff
@@ -7,14 +7,14 @@ import 'model/binding.dart';
...
-void main() {
+Future<void> main() async {
   assert(() {
     debugLogEnabled = true;
     return true;
   }());
   LicenseRegistry.addLicense(additionalLicenses);
-  LiveZulipBinding.ensureInitialized();
   WidgetsFlutterBinding.ensureInitialized();
+  await LiveZulipBinding.ensureInitialized();
```
rajveermalviya, line 17: The goal for this change is to pre-fetch the `deviceInfo` and `packageInfo` while ZulipBinding is being initialized and then later they can be used from [non-async functions](https://github.com/zulip/zulip-flutter/blob/8458b725ae0a9f6c7fe1cfea2ec71b4ae183b3ff/lib/widgets/content.dart#L1208-L1244) via the getters.

test/api/fake_api_test.dart
gnprice: What were the callers that that propagated up the call chain to? I was imagining it'd be just that `ApiConnection.live` requires a real user-agent, and `FakeApiConnection` defaults to a boring fake one, and then very little other code has to get involved.

lib/main.dart
gnprice: That make sense from an ergonomics perspective; but it makes me nervous to be introducing this wait into the startup path. Probably on any reasonable system the queries will return quickly and it's fine, but
* there may be bugs in these third-party plugins that cause them to sometimes hang before supplying an answer;
* there may be devices out there where some bug in a vendor's low-quality modifications to the Android platform software cause the queries to hang, or just to be slow, even though the plugins are doing only perfectly reasonable things that totally ought to be reliably fast.

Then the thing about putting awaits like these into the startup path is that it'd escalate any bugs of those types into a serious problem in the Zulip app for affected users.

So let's arrange things instead so that if for some reason those queries are slow, or even hang forever and never return, everything fails gracefully — just the places that actually wanted to use the answers have to make do with not knowing the answers.

One good way to do that, API-wise, would be to indeed make those into nice non-async getters, but returning nullable. Then when the binding is initialized, it kicks off the queries, and if the queries have finished by the time we try to make a request then we use the results; if not, we fall back to not knowing this information.

gnprice: I guess most of the time we'll be firing off one request promptly after startup, the register-events request. So if we find that that often happens before the platform queries finish, then we can add an extra wrinkle where we wait for the queries to finish… but with a timeout, like 10ms or something, so we prioritize the user's experience over being sure we include this information.

Let's keep that as a separate followup commit, though. Or it'd be great if you try a little manual experimenting to see if that's a problem in practice; if it seems like it isn't, we can skip that wrinkle entirely, and keep it in our back pocket as a possible followup for later if it turns out to be a problem sometimes in wider deployment.

test/api/fake_api_test.dart
rajveermalviya: The implementation I tried was to keep `deviceInfo` and `packageInfo` as async function instead of the current sync getters on `ZulipBinding` and then trying to populate the ApiConnection args (deviceInfo & packageInfo) from the async functions it's being called. But there was [one call](https://github.com/zulip/zulip-flutter/blob/99f6a8b944248f01009e1959ff0f64e5c48789e5/lib/model/store.dart#L225) which was not async, so had to add the args for it too. And doing so led to many other changes in the codebase which were moot, thus reverted it.

But if we stick to sync function `userAgentHeader` and let `ApiConnection.live` inline it in it's constructor like:
```dart
ApiConnection.live({
  ...
}) : this(client: http.Client(),
          ...
          userAgentHeader: ZulipBinding.instance.userAgentHeader());
```
then it does work without the need to poison the callers, I'll send a revision with this now.

rajveermalviya: Okay just tried it, and seems like there isn't much difference in line count, because now some tests that needs to check userAgent needs to pass the `testBinding.userAgent` to `FakeApiConnection.with_` which are quite a few. 

But on the plus side, the tests that don't check `userAgent` don't need to add ZulipBinding initialization lines:
```diff
-  TestZulipBinding.ensureInitialized();
-  tearDown(testBinding.reset);
```

gnprice: > Okay just tried it, and seems like there isn't much difference in line count, because now some tests that needs to check userAgent needs to pass the `testBinding.userAgent` to `FakeApiConnection.with_` which are quite a few.

Cool. Yeah, even if the total line count is the same but the impact is all concentrated in the one group of tests that's about the user-agent value itself, that sounds a lot nicer.

We can probably also improve that further with a local helper or two in those tests.

lib/main.dart
rajveermalviya: Thanks, yeah we definitely should have a fail safe if plugins fails for some reason or are slow to retrieve the data.

The latest revision now doesn't `await` on pre-fetchs for both (`deviceInfo` and `packageInfo`), and where the getters are called there's now some kind of relevant fallback just-in-case.

> if you try a little manual experimenting to see if that's a problem in practice

I tried that, by [`throw`-ing in case either of infos are null](https://github.com/zulip/zulip-flutter/blob/a9cf8d3527e5a27b2fe983f6709176620edad90b/lib/model/binding.dart#L353) but didn't see any exceptions in both release and debug builds for Android & macOS (can't try release on iOS Simulator) during startup.

sm-sayedi (commented):
Thanks @rajveermalviya for this. LGTM! Just a few (small) comments below!

rajveermalviya:
Thanks for the review @sm-sayedi, pushed a new revision!

lib/model/binding.dart
```diff
@@ -140,13 +140,23 @@ abstract class BaseDeviceInfo {
...
   ///
   /// Possible values are defined in:
   ///   https://developer.android.com/reference/android/os/Build.VERSION_CODES.html
   final int sdkInt;
 
-  AndroidDeviceInfo({required this.sdkInt});
+  AndroidDeviceInfo({
+    required this.release,
+    required this.sdkInt,
+  });
```
sm-sayedi, line 166:
```suggestion
  AndroidDeviceInfo({required this.release, required this.sdkInt});
```
nit: Can be inlined.

lib/model/binding.dart
```diff
@@ -159,6 +169,59 @@ class IosDeviceInfo extends BaseDeviceInfo {
...
+  ///
+  /// Examples: '17', '11.04'.
+  ///
+  /// This field is optional and may be null on some systems.
+  final String? versionId;
+
+  LinuxDeviceInfo({
+    required this.name,
+    required this.versionId,
+  });
```
sm-sayedi, line 229:
```suggestion
  LinuxDeviceInfo({required this.name, required this.versionId});
```

sm-sayedi (approved):
Thanks for the revision. Just added two comments (very small) below, otherwise looks good to me.

Let's move on to the mentor review with @sumanthvrao.

rajveermalviya:
Thanks for the review @sm-sayedi, updated.

lib/widgets/about_zulip.dart
```diff
@@ -41,9 +34,10 @@ class _AboutZulipPageState extends State<AboutZulipPage> {
             child: ConstrainedBox(
               constraints: const BoxConstraints(maxWidth: 400),
               child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
-                ListTile(
-                  title: Text(zulipLocalizations.aboutPageAppVersion),
-                  subtitle: Text(_packageInfo?.version ?? '(…)')),
+                if (appVersion != null)
+                  ListTile(
+                    title: Text(zulipLocalizations.aboutPageAppVersion),
+                    subtitle: Text(appVersion)),
```
chrisbobbe, line 40: This is a behavior change, so let's leave it out of a pure-refactor `[nfc]` commit. Before, when the app version is undetermined, we'd show a `ListTile` with subtitle "(…)". Now, in that case, the whole `ListTile` is omitted.

test/api/core_test.dart
```diff
@@ -8,25 +8,31 @@ import 'package:zulip/api/core.dart';
...
+          connection.prepare(json: {});
+          await connection.get(kExampleRouteName, (json) => json, 'example/route', params);
+          check(connection.lastRequest!).isA<http.Request>()
+            ..method.equals('GET')
+            ..url.asString.equals('${eg.realmUrl.origin}$expectedRelativeUrl')
+            ..headers.deepEquals({
+              ...authHeader(email: eg.selfAccount.email, apiKey: eg.selfAccount.apiKey),
+              ...testBinding.userAgentHeader(),
+            })
+            ..body.equals('');
```
chrisbobbe, line 36: I think this is a one-line change plus a reformatting with different whitespace; is that right? The whitespace change is kind of distracting to me as I review looking for meaningful changes. (Here and in several places below.)

test/api/core_test.dart
```diff
@@ -320,6 +321,41 @@ void main() {
...
+      ('ZulipFlutter/0.0.1+1 (Linux; Fedora Linux 40)', LinuxDeviceInfo(name: 'Fedora Linux', versionId: '40'),              packageInfo),
+      ('ZulipFlutter/0.0.1+1 (Linux; Fedora Linux)',    LinuxDeviceInfo(name: 'Fedora Linux', versionId: null),              packageInfo),
+    ];
+
+    for (final (userAgent, deviceInfo, pacakgeInfo) in testCases) {
+      test('matches $userAgent', () async {
+        testBinding.deviceInfoResult = deviceInfo;
+        testBinding.packageInfoResult = pacakgeInfo;
+        await checkUserAgent(userAgent);
+      });
```
chrisbobbe, line 356: It would be cleanest/most principled to add a line

```dart
```

at the top of the `test`, like we do in other tests. (But I don't think the lack of the line causes a problem with this test as written.)

test/api/core_test.dart
```diff
@@ -320,6 +321,41 @@ void main() {
...
+    final packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+
+    final testCases = [
+      ('ZulipFlutter/0.0.1+1 (Android 14)',             AndroidDeviceInfo(release: '14', sdkInt: 34),                        packageInfo),
+      ('ZulipFlutter/0.0.1+1 (iOS 17.4)',               IosDeviceInfo(systemVersion: '17.4'),                                packageInfo),
+      ('ZulipFlutter/0.0.1+1 (macOS 14.5.0)',           MacOsDeviceInfo(majorVersion: 14, minorVersion: 5, patchVersion: 0), packageInfo),
+      ('ZulipFlutter/0.0.1+1 (Windows)',                WindowsDeviceInfo(),                                                 packageInfo),
+      ('ZulipFlutter/0.0.1+1 (Linux; Fedora Linux 40)', LinuxDeviceInfo(name: 'Fedora Linux', versionId: '40'),              packageInfo),
+      ('ZulipFlutter/0.0.1+1 (Linux; Fedora Linux)',    LinuxDeviceInfo(name: 'Fedora Linux', versionId: null),              packageInfo),
+    ];
```
chrisbobbe, line 349: Cool! The way these cases is presented is nice and compact and easy to read, without lots of extra boilerplate for each case.

chrisbobbe (commented):
Thanks! Small comments below.

test/api/core_test.dart
gnprice: It looks like the whitespace changes in question are all an indentation change. It's natural for those to happen together with substantive changes sometimes — for example they'll happen whenever adding or removing a wrapper widget around a child subtree.

A handy tool for reading such diffs is `git log -b`, which ignores indentation changes. With that, this diff hunk looks like:
```diff
-      return FakeApiConnection.with_(account: eg.selfAccount, (connection) async {
+      return FakeApiConnection.with_(account: eg.selfAccount,
+        userAgentHeader: testBinding.userAgentHeader(),
+        (connection) async {
           connection.prepare(json: {});
           await connection.get(kExampleRouteName, (json) => json, 'example/route', params);
           check(connection.lastRequest!).isA<http.Request>()
             ..method.equals('GET')
             ..url.asString.equals('${eg.realmUrl.origin}$expectedRelativeUrl')
             ..headers.deepEquals({
               ...authHeader(email: eg.selfAccount.email, apiKey: eg.selfAccount.apiKey),
-            ...userAgentHeader(),
+              ...testBinding.userAgentHeader(),
             })
             ..body.equals('');
```

gnprice: (With `git config diff.colorMovedWS`, you can also get your default `git log -p` output to treat indentation changes the same way as moved code and give them a different background color. That's what I have [in my gitconfig](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3#file-gitconfig-L91-L119). I still use `-b` to get a clearer view in extreme cases.)

chrisbobbe: Cool. Yeah; I used `git log -b` to help verify the change, but I think I could have looked closer at the regular `git log -p` output—it turns out I already have that git config option set the way you do, and here's what it looks like:

image

test/api/core_test.dart
chrisbobbe: (Ah I did just notice there's a typo "pacakgeInfo" a few lines below this.)

test/api/core_test.dart
gnprice: Yeah, exactly. And yeah, those colors are adjustable - they're the diff.oldMoved and friends found in my gitconfig. Definitely worth trying some quick changes to them to see what works best for you.

From the screenshot, it looks like your terminal's default background color is already a faint gray rather than black, which ends up being pretty close to the background colors set there. I have my terminal background set to pure black, for maximum contrast, and then the colors in my config are chosen to be distinct from that.

chrisbobbe: This is better—I've changed the Android Studio color scheme to "high contrast":

image

gnprice: Yeah, here's what that looks like for me:
image


The striping in your screenshot looks like it could get a bit annoying. You might be able to adjust the line spacing to fix that — generally I'd expect a terminal to use 1.0x line spacing anyway.

chrisbobbe: > The striping in your screenshot looks like it could get a bit annoying. You might be able to adjust the line spacing to fix that — generally I'd expect a terminal to use 1.0x line spacing anyway.

Yep, got it! I adjusted a font setting in the Android Studio settings. It was a little confusing because things outside the terminal were responding to the setting, but the terminal wasn't—but then it did after I restarted Android Studio.

lib/main.dart
```diff
@@ -13,8 +13,8 @@ void main() {
     return true;
   }());
   LicenseRegistry.addLicense(additionalLicenses);
-  LiveZulipBinding.ensureInitialized();
   WidgetsFlutterBinding.ensureInitialized();
+  LiveZulipBinding.ensureInitialized();
   NotificationService.instance.start();
```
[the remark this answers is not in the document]
rajveermalviya: Yes they need to be reordered, since `deviceInfo` is being prefetched in `LiveZulipBinding.ensureInitialized()` and `PlatformChannel`s require `WidgetsFlutterBinding` to be initialized first.

lib/widgets/about_zulip.dart
rajveermalviya: removed this behavior change

test/api/core_test.dart
rajveermalviya: There is a `addTearDown(testBinding.reset)` at the start of the main, which should work for all tests in that file.

rajveermalviya:
Thanks for the review @chrisbobbe, pushed a new revision PTAL.

test/api/core_test.dart
chrisbobbe: Oh! Indeed; thanks for pointing this out.

lib/model/binding.dart
```diff
@@ -195,16 +234,6 @@ class LiveZulipBinding extends ZulipBinding {
...
-  @override
-  Future<BaseDeviceInfo> deviceInfo() async {
-    final deviceInfo = await device_info_plus.DeviceInfoPlugin().deviceInfo;
-    return switch (deviceInfo) {
-      device_info_plus.AndroidDeviceInfo(:var version)   => AndroidDeviceInfo(sdkInt: version.sdkInt),
-      device_info_plus.IosDeviceInfo(:var systemVersion) => IosDeviceInfo(systemVersion: systemVersion),
-      _                                                  => throw UnimplementedError(),
-    };
-  }
-
```
chrisbobbe, line 207: Ah, and I didn't notice last time: is there a reason to move this `device_info_plus`-related code before other things in the class? I think `ZulipBinding`, `LiveZulipBinding`, and `TestZulipBinding` should declare their members in the same order; the consistency should help reading and editing them together.

chrisbobbe (commented):
Thanks! Two small comments below, and I'll go ahead and mark this for Greg's review.

lib/model/binding.dart
```diff
@@ -66,6 +68,46 @@ abstract class ZulipBinding {
     _instance = this;
   }
 
+  /// Provides device and operating system information,
+  /// via package:device_info_plus.
+  ///
+  /// Returns a Future resolving to null if an error was
+  /// encountered while fetching, otherwise a Future
+  /// resolving to the fetched device info.
```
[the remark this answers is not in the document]
rajveermalviya: Addressed, though removed `This wraps ...` because the getter don't really wrap those functions anymore.

chrisbobbe:
LGTM, thanks! Over to Greg, then.

lib/main.dart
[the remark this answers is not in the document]
gnprice: What does the error look like if the order of these gets swapped back?

If it doesn't make clear that the problem is that `WidgetsFlutterBinding` is needed first, then ideally we can do something to make the error clear.

If that's hard or annoying, then just a comment here explaining this ordering constraint would also be an acceptable solution.

lib/model/binding.dart
```diff
@@ -164,11 +262,68 @@ class LiveZulipBinding extends ZulipBinding {
   /// Initialize the binding if necessary, and ensure it is a [LiveZulipBinding].
   static LiveZulipBinding ensureInitialized() {
     if (ZulipBinding._instance == null) {
-      LiveZulipBinding();
+      final binding = LiveZulipBinding();
+      binding._deviceInfo = binding._prefetchDeviceInfo();
+      binding._packageInfo = binding._prefetchPackageInfo();
```
gnprice, line 267: nit: this logic is probably cleaner to read as a constructor body

lib/model/binding.dart
gnprice: Same question :slightly_smiling_face:

lib/model/binding.dart
```diff
@@ -164,11 +262,68 @@ class LiveZulipBinding extends ZulipBinding {
...
+
+  @override
+  Future<PackageInfo?> get packageInfo => _packageInfo;
+  late Future<PackageInfo?> _packageInfo;
+
+  @override
+  PackageInfo? get maybePackageInfo => _maybePackageInfo;
+  PackageInfo? _maybePackageInfo;
+
+  Future<BaseDeviceInfo?> _prefetchDeviceInfo() async {
```
gnprice, line 291: nit: keep "device info" members next to "device info" members, and "package info" members next to "package info" members

More generally, [as the Flutter style guide puts it](https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense):
> Fields should come before the methods that manipulate them, if they are specific to a particular group of methods.

where the key part of "before" is that they're *next to* those methods, though in particular they should come just before them.

lib/model/binding.dart
```diff
@@ -66,6 +68,34 @@ abstract class ZulipBinding {
...
+  /// The returned Future resolves to null if an error is
+  /// encountered while fetching the data.
+  Future<BaseDeviceInfo?> get deviceInfo;
+
+  /// Provides device and operating system information,
+  /// via package:device_info_plus.
+  ///
+  /// May return null if prefetching hasn't completed yet,
+  /// or an error occured while fetching the data.
+  BaseDeviceInfo? get maybeDeviceInfo;
```
gnprice, line 83:
```suggestion
  /// This is the value [deviceInfo] resolved to,
  /// or null if that hasn't resolved yet.
  BaseDeviceInfo? get maybeDeviceInfo;
```

lib/model/binding.dart
```diff
@@ -66,6 +68,34 @@ abstract class ZulipBinding {
     _instance = this;
   }
 
+  /// Provides device and operating system information,
+  /// via package:device_info_plus.
+  ///
+  /// The returned Future resolves to null if an error is
+  /// encountered while fetching the data.
+  Future<BaseDeviceInfo?> get deviceInfo;
```
gnprice, line 76: I think the "This wraps" line in the existing dartdoc is helpful — it helps tie this member to a specific part of the plugin's API.

lib/model/binding.dart
gnprice, line 83: I feel like "maybe" doesn't quite capture the relationship between this and `deviceInfo`. We often say "foo" and "maybeFoo" when the former throws in the absence of a result and the latter returns null, but this is something else.

How about:
```suggestion
  BaseDeviceInfo? get syncDeviceInfo;
```

I think that fairly well captures why you might want this one. And then it's natural that the trade-off is that the sync version might not have data yet.

lib/model/binding.dart
```diff
@@ -164,11 +262,68 @@ class LiveZulipBinding extends ZulipBinding {
...
+                                                                minorVersion: info.minorVersion,
+                                                                patchVersion: info.patchVersion),
+        device_info_plus.WindowsDeviceInfo() => WindowsDeviceInfo(),
+        device_info_plus.LinuxDeviceInfo()   => LinuxDeviceInfo(name: info.name,
+                                                                versionId: info.versionId),
+        _                                    => throw UnimplementedError(),
+      };
+    } catch (e) {
+      assert(debugLog('Failed to prefetch device info: $e'));
+      // TODO(log)
```
gnprice, line 308: nit:
```suggestion
      assert(debugLog('Failed to prefetch device info: $e')); // TODO(log)
```

This style makes a grep for "TODO(log)" comments more informative, as the very line it's on is self-explanatory.

lib/model/binding.dart
```diff
@@ -133,13 +163,20 @@ abstract class BaseDeviceInfo {
 
 /// Like [device_info_plus.AndroidDeviceInfo], but without things we don't use.
 class AndroidDeviceInfo extends BaseDeviceInfo {
+  /// The user-visible version string.
+  ///
+  /// E.g., "1.0" or "3.4b5" or "bananas". This field is an opaque string.
+  /// Do not assume that its value has any particular structure or that
+  /// values of RELEASE from different releases can be somehow ordered.
+  final String release;
```
gnprice, line 171: Does this admonition come from some external docs? If so let's link to that.

lib/model/binding.dart
```diff
@@ -152,6 +189,67 @@ class IosDeviceInfo extends BaseDeviceInfo {
   IosDeviceInfo({required this.systemVersion});
 }
 
+/// Like [device_info_plus.MacOsDeviceInfo], but without things we don't use.
+class MacOsDeviceInfo extends BaseDeviceInfo {
+  /// The major release number, such as 10 in version 10.9.3.
+  final int majorVersion;
```
gnprice, line 195: Cool, these are good descriptions.

Is there some Apple documentation we can link to for the underlying platform fields they correspond to?

lib/model/binding.dart
```diff
@@ -152,6 +189,67 @@ class IosDeviceInfo extends BaseDeviceInfo {
...
+  MacOsDeviceInfo({
+    required this.majorVersion,
+    required this.minorVersion,
+    required this.patchVersion,
+  });
+}
+
+/// Like [device_info_plus.WindowsDeviceInfo], currently only used to
+/// determine if we're on Windows.
+class WindowsDeviceInfo implements BaseDeviceInfo {}
```
gnprice, line 212: I think I remember us having some discussion of why it's inconvenient to have more detail on Windows versions, but I've forgotten what the reason was. A short comment here (perhaps with a link to that previous thread) would be helpful for the next person who wonders why this one subclass has less information in it.

lib/model/binding.dart
```diff
@@ -152,6 +189,67 @@ class IosDeviceInfo extends BaseDeviceInfo {
...
+/// See:
+///   https://www.freedesktop.org/software/systemd/man/os-release.html
+class LinuxDeviceInfo implements BaseDeviceInfo {
+  /// A string identifying the operating system, without a version component,
+  /// and suitable for presentation to the user.
+  ///
+  /// Examples: 'Fedora', 'Debian GNU/Linux'.
+  ///
+  /// If not set, defaults to 'Linux'.
+  final String name;
```
gnprice, line 225:
```suggestion
  /// Examples: 'Fedora', 'Debian GNU/Linux', or just 'Linux'.
  final String name;
```

As far as we're concerned, there's no question of setting it to something or not doing so — it just is. But the mention of the string "Linux" is useful in that given it's the default for the `/etc/os-release` file, it's a plausible value worth mentioning as an example.

lib/model/binding.dart
```diff
@@ -152,6 +189,67 @@ class IosDeviceInfo extends BaseDeviceInfo {
...
+  /// A lower-case string identifying the operating system version, excluding
+  /// any OS name information or release code name, and suitable for processing
+  /// by scripts or usage in generated filenames.
+  ///
+  /// The version is mostly numeric, and contains no spaces or other characters
+  /// outside of 0–9, a–z, '.', '_' and '-'.
+  ///
+  /// Examples: '17', '11.04'.
+  ///
+  /// This field is optional and may be null on some systems.
```
gnprice, line 236:
```suggestion
```

The type already expresses that it might be null.

lib/model/binding.dart
```diff
@@ -152,6 +189,67 @@ class IosDeviceInfo extends BaseDeviceInfo {
...
+  final String? versionId;
```
gnprice, line 237: Let's make explicit exactly what this corresponds to from the platform side:
```suggestion
  /// Examples: '17', '11.04'.
  ///
  /// Docs: https://www.freedesktop.org/software/systemd/man/latest/os-release.html#VERSION_ID=
  final String? versionId;
```

test/api/fake_api.dart
```diff
@@ -171,12 +176,16 @@ class FakeApiConnection extends ApiConnection {
     Uri? realmUrl,
     int? zulipFeatureLevel = eg.futureZulipFeatureLevel,
     Account? account,
+    Map<String, String> Function()? userAgentHeader,
```
gnprice, line 179: It looks like this is always the same expression, when not null: `testBinding.userAgentHeader`.

How about making it a boolean instead? Like `useBinding: true`. That'd reduce the amount of noise at the call sites.

Then I think it can actually be a boolean up at the base class, too.

After that change, I think `buildUserAgent` can move to `ApiConnection` (e.g. as a private static), which is where it most logically belongs. We'd lose memoization between different `ApiConnection` instances, but that's fine — we aren't constantly making new connection instances.

gnprice (commented):

Comments below. I've read fully through the first three commits:
7be37dd7c binding: Prefetch deviceInfo during initialization
f9ce9e56c binding: Add packageInfo getter
662346ebb about_zulip [nfc]: Refactor to use packageInfo from ZulipBinding

and I've skimmed through the remaining three commits:

6a0fa8724 binding: Support more device/os variants for deviceInfo getter
4544709bc binding [nfc]: Move userAgentHeader function to ZulipBinding
702310960 api: Embed platform and app info in user-agent

and left high-level comments that will lead to some refactors, so I'll read those in more detail in a later round of review.

lib/main.dart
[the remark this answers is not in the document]
rajveermalviya: Here's how the errors look like:
```
flutter: Failed to prefetch device info: Binding has not yet been initialized.
The "instance" getter on the ServicesBinding binding mixin is only available once that binding has been initialized.
Typically, this is done by calling "WidgetsFlutterBinding.ensureInitialized()" or "runApp()" (the latter calls the former). Typically this call is done in the "void main()" method. The "ensureInitialized" method is idempotent; calling it multiple times is not harmful. After calling that method, the "instance" getter will return the binding.
In a test, one can call "TestWidgetsFlutterBinding.ensureInitialized()" as the first line in the test's "main()" method to initialize the binding.
If ServicesBinding is a custom binding mixin, there must also be a custom binding class, like WidgetsFlutterBinding, but that mixes in the selected binding, and that is the class that must be constructed before using the "instance" getter.
#0      BindingBase.checkInstance.<anonymous closure> (package:flutter/src/foundation/binding.dart:309:9)
#1      BindingBase.checkInstance (package:flutter/src/foundation/binding.dart:390:6)
#2      ServicesBinding.instance (package:flutter/src/services/binding.dart:68:54)
#3      _findBinaryMessenger (package:flutter/src/services/platform_channel.dart:158:25)
#4      MethodChannel.binaryMessenger (package:flutter/src/services/platform_channel.dart:293:56)
#5      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:327:15)
#6      MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:507:12)
#7      MethodChannelDeviceInfo.deviceInfo (package:device_info_plus_platform_interface/method_channel/method_channel_device_info.dart:19:24)
#8      DeviceInfoPlugin.macOsInfo (package:device_info_plus/device_info_plus.dart:86:48)
#9      DeviceInfoPlugin.deviceInfo (package:device_info_plus/device_info_plus.dart:107:16)
#10     LiveZulipBinding._prefetchDeviceInfo (package:zulip/model/binding.dart:338:62)
#11     new LiveZulipBinding (package:zulip/model/binding.dart:290:19)
#12     LiveZulipBinding.ensureInitialized (package:zulip/model/binding.dart:297:7)
#13     main (package:zulip/main.dart:16:20)
#14     _runMain.<anonymous closure> (dart:ui/hooks.dart:301:23)
#15     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:297:19)
#16     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)

flutter: Failed to prefetch package info: Binding has not yet been initialized.
The "instance" getter on the ServicesBinding binding mixin is only available once that binding has been initialized.
Typically, this is done by calling "WidgetsFlutterBinding.ensureInitialized()" or "runApp()" (the latter calls the former). Typically this call is done in the "void main()" method. The "ensureInitialized" method is idempotent; calling it multiple times is not harmful. After calling that method, the "instance" getter will return the binding.
In a test, one can call "TestWidgetsFlutterBinding.ensureInitialized()" as the first line in the test's "main()" method to initialize the binding.
If ServicesBinding is a custom binding mixin, there must also be a custom binding class, like WidgetsFlutterBinding, but that mixes in the selected binding, and that is the class that must be constructed before using the "instance" getter.
#0      BindingBase.checkInstance.<anonymous closure> (package:flutter/src/foundation/binding.dart:309:9)
#1      BindingBase.checkInstance (package:flutter/src/foundation/binding.dart:390:6)
#2      ServicesBinding.instance (package:flutter/src/services/binding.dart:68:54)
#3      _findBinaryMessenger (package:flutter/src/services/platform_channel.dart:158:25)
#4      MethodChannel.binaryMessenger (package:flutter/src/services/platform_channel.dart:293:56)
#5      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:327:15)
#6      MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:507:12)
#7      MethodChannel.invokeMapMethod (package:flutter/src/services/platform_channel.dart:534:49)
#8      MethodChannelPackageInfo.getAll (package:package_info_plus_platform_interface/method_channel_package_info.dart:13:32)
#9      PackageInfo.fromPlatform (package:package_info_plus/package_info_plus.dart:80:61)
#10     LiveZulipBinding._prefetchPackageInfo (package:zulip/model/binding.dart:367:56)
#11     new LiveZulipBinding (package:zulip/model/binding.dart:291:20)
#12     LiveZulipBinding.ensureInitialized (package:zulip/model/binding.dart:297:7)
#13     main (package:zulip/main.dart:16:20)
#14     _runMain.<anonymous closure> (dart:ui/hooks.dart:301:23)
#15     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:297:19)
#16     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
```

Basically both of the prefetch fail because of same reason — MethodChannel requiring `WidgetsFlutterBinding` initialized first.

rajveermalviya:
Thanks for the reviews @chrisbobbe and @gnprice, pushed a new revision PTAL.

lib/main.dart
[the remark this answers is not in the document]
gnprice: Cool, that seems clear enough — thanks.

test/api/core_test.dart
```diff
@@ -6,27 +6,34 @@ import 'package:http/http.dart' as http;
...
-          ..url.asString.equals('${eg.realmUrl.origin}$expectedRelativeUrl')
-          ..headers.deepEquals({
-            ...authHeader(email: eg.selfAccount.email, apiKey: eg.selfAccount.apiKey),
-            ...userAgentHeader(),
-          })
-          ..body.equals('');
+      return FakeApiConnection.with_(account: eg.selfAccount,
+          useBinding: true,
+          (connection) async {
+          connection.prepare(json: {});
```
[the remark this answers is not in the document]
gnprice: Alternatively: do these test cases need `useBinding: true`? It feels like it's not part of what this test case (and the ones for `post`, etc. below) are about.

I think the tests for these methods can use the fallback default user-agent, same as most of our other API-using tests do. Then of course we have the separate test cases you've written below for testing specifically the User-Agent logic.

test/api/route/messages_test.dart
```diff
@@ -320,96 +324,108 @@ void main() {
...
-            'to': streamId.toString(),
-            'topic': topic,
-            'content': content,
-            'queue_id': '"abc:123"',
-            'local_id': '"456"',
-            'read_by_sender': 'true',
-          });
+      return FakeApiConnection.with_(
+        useBinding: true,
+        (connection) async {
```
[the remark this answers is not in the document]
gnprice: Or similarly to what I wrote above about tests like `ApiConnection.get` — do these test cases need `useBinding: true` anyway?

test/api/route/messages_test.dart
```diff
@@ -10,11 +10,15 @@ import 'package:zulip/api/route/messages.dart';
...
 
 import '../../example_data.dart' as eg;
+import '../../model/binding.dart';
 import '../../stdlib_checks.dart';
 import '../fake_api.dart';
 import 'route_checks.dart';
 
 void main() {
+  TestZulipBinding.ensureInitialized();
+  tearDown(testBinding.reset);
```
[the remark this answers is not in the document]
gnprice: Also, does even `useBinding: true` cause any need for this line? It doesn't actually change any state on the binding, does it?

So if not, we can leave this line out.

test/api/core_test.dart
```diff
@@ -308,6 +321,41 @@ void main() {
...
+      ('ZulipFlutter/0.0.1+1 (Windows)',                WindowsDeviceInfo(),                                                 packageInfo),
+      ('ZulipFlutter/0.0.1+1 (Linux; Fedora Linux 40)', LinuxDeviceInfo(name: 'Fedora Linux', versionId: '40'),              packageInfo),
+      ('ZulipFlutter/0.0.1+1 (Linux; Fedora Linux)',    LinuxDeviceInfo(name: 'Fedora Linux', versionId: null),              packageInfo),
+    ];
+
+    for (final (userAgent, deviceInfo, packageInfo) in testCases) {
+      test('matches $userAgent', () async {
+        testBinding.deviceInfoResult = deviceInfo;
+        testBinding.packageInfoResult = packageInfo;
+        await checkUserAgent(userAgent);
```
gnprice, line 355: Similar to my other comment about `testBinding.reset` and teardowns:
```suggestion
        testBinding.deviceInfoResult = deviceInfo;
        testBinding.packageInfoResult = packageInfo;
        addTearDown(testBinding.reset);
        await checkUserAgent(userAgent);
```

I think that suffices for taking the place of the global `tearDown` call.

test/api/core_test.dart
```diff
@@ -308,6 +321,41 @@ void main() {
...
+              ...authHeader(email: eg.selfAccount.email, apiKey: eg.selfAccount.apiKey),
+              ...{'User-Agent': expectedUserAgent},
+            });
+      });
+    }
+
+    final packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+
+    final testCases = [
+      ('ZulipFlutter/0.0.1+1 (Android 14)',             AndroidDeviceInfo(release: '14', sdkInt: 34),                        packageInfo),
```
gnprice, line 343: It looks like `packageInfo` doesn't actually vary between these test cases. In that case, let's simplify by taking it out of the data list.

test/api/core_test.dart
```diff
@@ -308,6 +321,41 @@ void main() {
...
+      return FakeApiConnection.with_(account: eg.selfAccount,
+        useBinding: true,
+        (connection) async {
+          connection.prepare(json: {});
+          await connection.get(kExampleRouteName, (json) => json, 'example/route', null);
+          check(connection.lastRequest!).isA<http.Request>()
+            .headers.deepEquals({
+              ...authHeader(email: eg.selfAccount.email, apiKey: eg.selfAccount.apiKey),
+              ...{'User-Agent': expectedUserAgent},
+            });
```
gnprice, line 336: Since this test is all about the User-Agent (and other tests are checking the auth headers), let's simplify:
```suggestion
            .headers['User-Agent'].equals(expectedUserAgent);
```

test/api/core_test.dart
```diff
@@ -308,6 +321,41 @@ void main() {
...
     }
   });
+
+  group('ApiConnection user-agent', () {
+    Future<void> checkUserAgent(String expectedUserAgent) async {
+      return FakeApiConnection.with_(account: eg.selfAccount,
+        useBinding: true,
+        (connection) async {
+          connection.prepare(json: {});
+          await connection.get(kExampleRouteName, (json) => json, 'example/route', null);
```
gnprice, line 331: Let's also pick any one of the test cases below, and do one check with each of the different request methods: `get`, `post`, `postFileFromStream`, `delete`.

gnprice (commented):

With this round, I've now read through the whole PR.

rajveermalviya:

test/api/route/messages_test.dart
```diff
@@ -316,7 +316,7 @@ void main() {
         ..method.equals('POST')
         ..url.path.equals('/api/v1/messages')
         ..bodyFields.deepEquals(expectedBodyFields)
-        ..headers['User-Agent'].equals(expectedUserAgent ?? userAgentHeader()['User-Agent']!);
+        ..headers['User-Agent'].equals(expectedUserAgent ?? kFallbackUserAgentHeader['User-Agent']!);
```
gnprice, line 319: I like how much smaller the impact of this on unrelated tests is now!

gnprice (commented):
Thanks @rajveermalviya for the revision! This all looks great — merging.

Issue #602, opened 2024-03-29
test: CI flake in "UpdateMachine.load retries registerQueue on NetworkError"


chrisbobbe:
Noticed when working on code that's almost definitely unrelated (implementing the web-auth feature):

https://github.com/zulip/zulip-flutter/pull/600#issuecomment-2026363037
```
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/store_test.dart: UpdateMachine.load retries registerQueue on NetworkError (failed)
  Bad state: Exceeded timeout 1:00:00.000000 while flushing timers
  package:fake_async/fake_async.dart 217:9   FakeAsync.flushTimers.<fn>
  package:fake_async/fake_async.dart 241:21  FakeAsync._fireTimersWhile
  package:fake_async/fake_async.dart 214:5   FakeAsync.flushTimers
  test/fake_async.dart 29:7                  awaitFakeAsync
  test/model/store_test.dart 176:57          main.<fn>.<fn>
```

gnprice:
Thanks for spotting this. If that indeed also happens in main, then it's a bug in either the test code, or the code under test, so it'd be good to debug and fix.

chrisbobbe:
Noticed this again, this time locally in `tools/check`.

gnprice:
This flake happened today in a Flutter "customer tests" run:
* https://github.com/flutter/flutter/issues/149230

So it's past time we should resolve it. I'll investigate, and either fix the test or disable it.

gnprice:
Aha. The problem is here in `BackoffMachine.wait`:
```dart
    await Future<void>.delayed(Duration(milliseconds: duration.round()));
```

If `duration` is less than 0.5 (which happens with probability 0.5% on the first backoff, since it's uniform from 0 to 100), then it rounds to zero, and we wait for a zero duration.

That then confuses the sequence of happenings that the test is counting on, which is fairly brittle.

I'll try to tighten down the self-checks inside the test, so that if its fairly-brittle sequencing assumptions fail then the test detects that up front. And then I'll fix backoff so that it always waits a nonzero interval — `Duration` accepts microseconds, so applying a floor of one microsecond should have no material effect on the jitter but should solve this kind of problem for tests.

gnprice:
Oho, and one of the contributing factors here was this limitation called out in `awaitFakeAsync`:
```
  // TODO: if the future returned by [callback] completes with an error,
  //   it would be good to throw that error immediately rather than finish
[...]

PR #726, opened 2024-06-05
backoff: Always wait a positive duration

gnprice:
Fixes: #602


chrisbobbe:
LGTM, thanks for figuring this out! Merged.

PR #727, opened 2024-06-06
content: Put code-block styles in `ContentTheme`; use weightVariableTextStyle for bold spans

chrisbobbe:
| Before | After |
| --- | --- |
| 34AB2CF1-E279-4702-A855-80722BE9056A | 17039B63-B706-4F85-9253-CC5DF863DFB3 |

Related: #95

### DIFF
--- a/lib/widgets/code_block.dart
+++ b/lib/widgets/code_block.dart
@@ -15,7 +15,8 @@ class CodeBlockTextStyles {
       plain: kMonospaceTextStyle
         .merge(const TextStyle(
           fontSize: 0.825 * kBaseFontSize,
-          height: 1.4)),
+          height: 1.4))
+        .merge(weightVariableTextStyle(context)),
 
       // .hll { background-color: hsl(60deg 100% 90%); }
       hll: TextStyle(backgroundColor: const HSLColor.fromAHSL(1, 60, 1, 0.90).toColor()),
--- a/lib/widgets/code_block.dart
+++ b/lib/widgets/code_block.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 
 import '../model/code_block.dart';
+import 'content.dart';
 import 'text.dart';
 
 /// [TextStyle]s used to render code blocks.
@@ -11,6 +12,11 @@ class CodeBlockTextStyles {
   factory CodeBlockTextStyles(BuildContext context) {
     final bold = weightVariableTextStyle(context, wght: 700);
     return CodeBlockTextStyles._(
+      plain: kMonospaceTextStyle
+        .merge(const TextStyle(
+          fontSize: 0.825 * kBaseFontSize,
+          height: 1.4)),
+
       // .hll { background-color: hsl(60deg 100% 90%); }
       hll: TextStyle(backgroundColor: const HSLColor.fromAHSL(1, 60, 1, 0.90).toColor()),
 
@@ -204,6 +210,7 @@ class CodeBlockTextStyles {
   }
 
   CodeBlockTextStyles._({
+    required this.plain,
     required TextStyle hll,
     required TextStyle c,
     required TextStyle err,
@@ -329,6 +336,9 @@ class CodeBlockTextStyles {
     _vg = vg,
     _vi = vi,
     _il = il;
+
+  /// The baseline style that the [forSpan] styles get applied on top of.
+  final TextStyle plain;
 
   final TextStyle _hll;
   final TextStyle _c;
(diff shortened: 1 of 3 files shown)


lib/widgets/code_block.dart
```diff
@@ -1,267 +1,549 @@
...
+    required TextStyle s1,
+    required TextStyle ss,
+    required TextStyle bp,
+    required TextStyle vc,
+    required TextStyle vg,
+    required TextStyle vi,
+    required TextStyle il,
+  }) :
+    _hll = hll,
+    _c = c,
```
gnprice, line 279: Oof, lotta boilerplate.

But I think this is the structure we need in order to support dark mode (and particularly to support nicely lerping in and out of it), so that's just how it'll be. Definitely glad we have this off in its own file.

lib/widgets/code_block.dart
```diff
@@ -1,267 +1,549 @@
...
 
-// .gr { color: hsl(0deg 100% 50%); }
-final _kCodeBlockStyleGr = TextStyle(color: const HSLColor.fromAHSL(1, 0, 1, 0.50).toColor());
+      // .gr { color: hsl(0deg 100% 50%); }
+      gr: TextStyle(color: const HSLColor.fromAHSL(1, 0, 1, 0.50).toColor()),
 
-// .gh { color: hsl(240deg 100% 25%); font-weight: bold; }
-final _kCodeBlockStyleGh = TextStyle(color: const HSLColor.fromAHSL(1, 240, 1, 0.25).toColor(), fontWeight: FontWeight.bold);
+      // .gh { color: hsl(240deg 100% 25%); font-weight: bold; }
+      gh: TextStyle(color: const HSLColor.fromAHSL(1, 240, 1, 0.25).toColor()).merge(bold),
```
gnprice, line 62: Hmm I see. The `FontWeight.bold` was getting ignored because of the `wght` variation, which leaked through from the ambient TextStyle (which in turn came from the ContentTheme). Is that right?

It'd be good to add a few words to the commit message about the visible effect of this. I wasn't sure what to expect before I opened up the screenshots at full size and compared them.

gnprice: In particular:

    And while we're add it, add some tests that spot-check the weight of
    a non-bold span and a bold span.

Those tests aren't just a by-the-way — they're regression tests for the bug this fixes :slightly_smiling_face:

lib/widgets/code_block.dart
```diff
@@ -1,267 +1,549 @@
...
+  final TextStyle _vg;
+  final TextStyle _vi;
+  final TextStyle _il;
+
+  /// The [TextStyle] for a [CodeBlockSpanType], if there is one.
+  ///
+  /// Used to render syntax highlighting.
+  // Span styles adapted from:
+  //   https://github.com/zulip/zulip/blob/213387249e7ba7772084411b22d8cef64b135dd0/web/styles/pygments.css
+  TextStyle? span(CodeBlockSpanType type) {
```
gnprice, line 411: How about:
```suggestion
  TextStyle? forSpan(CodeBlockSpanType type) {
```

The name just "span" reads to me like it's meant to return a span.

lib/widgets/code_block.dart
```diff
@@ -1,267 +1,549 @@
...
+    _ss = ss,
+    _bp = bp,
+    _vc = vc,
+    _vg = vg,
+    _vi = vi,
+    _il = il;
+
+  final TextStyle plain;
+
+  final TextStyle _hll;
```
gnprice, line 343: `plain` here has a different role from all the other styles, in that it's meant to be a baseline that each of them gets applied on top of.

So that asymmetry would be good to make explicit, perhaps in dartdoc on `plain`. In particular that helps explain why in the last commit, an explicit weight gets added to `plain` but there's no need for it on the rest.

gnprice (commented):
Thanks! This looks good — various small comments below.

lib/widgets/code_block.dart
chrisbobbe: > Hmm I see. The `FontWeight.bold` was getting ignored because of the `wght` variation, which leaked through from the ambient TextStyle (which in turn came from the ContentTheme). Is that right?

No, I don't think that's quite right. The "Before" screenshot already shows the text being bolder than surrounding text, just not in the way we want it to be. I think this is because of:

- https://github.com/flutter/flutter/issues/136779

For glyphs that render in a variable-weight font, we want to specify a `wght` value, so that they appear as the font designer intended. Just setting `TextStyle.fontWeight` by itself doesn't do that—not until https://github.com/flutter/flutter/issues/148026 anyway—and our way of setting `wght` is to use `weightVariableTextStyle`. These glyphs did _have_ a `wght` value before this change, but it was wrong. It should be 700, but it was 400, from the ambient `TextStyle` from `ContentTheme`.

chrisbobbe:

lib/widgets/code_block.dart
gnprice: Yeah, I guess "ignored" isn't right. But as you say, it (the `FontWeight.bold`) wasn't having the desired effect.

gnprice:
Thanks for the revision! This all looks good, but it seems to have a conflict with #728, which I just merged. So please go ahead and rebase past that, and then merge.

chrisbobbe:
Done; thanks for the review!

PR #728, opened 2024-06-07
text: Convert the last uses of `FontWeight.bold` to `weightVariableTextStyle`

chrisbobbe:
(That is, the last uses except the ones in code-block styles, which are treated in #727.)

| Before | After |
| --- | --- |
| 6EEB5903-E8FC-46D4-84CE-083ACE1C1E59 | 2CF00F52-126B-4950-88A8-69E65A531044 |
| 2F7D6B2C-E328-4AE5-AC0A-2CE5AEB23A7A | C8BDEB92-A391-4316-AB36-6EED5586DCFE |
| 56A0F77C-578E-4D22-8A30-7D4AEFD26450 | C0EF8351-7DFD-4CAC-B7AE-3049A039E5A7 |

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -46,10 +46,16 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
       fontFamilyFallback: defaultFontFamilyFallback,
     )
       .merge(weightVariableTextStyle(context))
-      .copyWith(debugLabel: 'ContentTheme.textStylePlainParagraph');
+      .copyWith(debugLabel: 'ContentTheme.textStylePlainParagraph'),
+    textStyleError = const TextStyle(
+      fontSize: kBaseFontSize, fontWeight: FontWeight.bold, color: Colors.red),
+    textStyleErrorCode = kMonospaceTextStyle
+      .merge(const TextStyle(fontSize: kBaseFontSize, color: Colors.red));
 
   ContentTheme._({
     required this.textStylePlainParagraph,
+    required this.textStyleError,
+    required this.textStyleErrorCode,
   });
 
   /// The [ContentTheme] from the context's active theme.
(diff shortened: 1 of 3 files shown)


alya:
Can we give more space to the labels in the profile screen? It seems unnecessary to make so many of them spill onto two lines.

lib/widgets/profile.dart
```diff
@@ -10,10 +10,13 @@ import 'content.dart';
...
 import 'store.dart';
+import 'text.dart';
 
 class _TextStyles {
   static const primaryFieldText = TextStyle(fontSize: 20);
-  static const customProfileFieldLabel = TextStyle(fontSize: 15, fontWeight: FontWeight.bold);
+  static TextStyle customProfileFieldLabel(BuildContext context) =>
+    const TextStyle(fontSize: 15)
+      .merge(weightVariableTextStyle(context, wght: 700));
   static const customProfileFieldText = TextStyle(fontSize: 15);
```
gnprice, line 20: nit: blank lines to separate, since these are no longer one line each

lib/widgets/content.dart
```diff
@@ -1351,8 +1360,5 @@ InlineSpan _errorUnimplemented(UnimplementedNode node) {
   }
 }
 
-const errorStyle = TextStyle(
```
gnprice, line 1354: content [nfc]: Move codeStyle to ContentTheme

Do you mean s/codeStyle/errorStyle/?

lib/widgets/content.dart
```diff
@@ -1351,8 +1360,5 @@ InlineSpan _errorUnimplemented(UnimplementedNode node) {
...
-  fontSize: kBaseFontSize, fontWeight: FontWeight.bold, color: Colors.red);
-
 final errorCodeStyle = kMonospaceTextStyle
   .merge(const TextStyle(fontSize: kBaseFontSize, color: Colors.red));
```
gnprice, line 1364: Maybe move this at the same time? Seems like we'll need to move it too for dark mode; and in any case it's closely related to `errorStyle`, so it seems like both of them are easier to understand if they're defined next to each other.

gnprice (commented):
Thanks! This answers exactly the question that came to mind when I was reviewing #727. :slightly_smiling_face: 

Code all looks good, with small comments below.

gnprice:
> Should we give more space to the labels in the profile screen? It seems unnecessary to make not particularly long labels spill onto two lines.

Yeah, that'd likely be a good change.

We might also try turning the weight down a little on the labels, like perhaps to 600 instead of 700.

In general this is a screen that's not yet based on any design from Vlad; I think there wasn't one in the Figma document when Shu was building this screen, so we just made something similar to what zulip-mobile has. So we can make tweaks where we think there are easy improvements, but we don't need to spend time trying to really dial it in. At some point in the future, Vlad will design this screen as a whole, and then we'll revise it to implement that.

chrisbobbe:
| `main` | after `wghtVariableTextStyle` | after increasing horizontal space |
| --- | --- | --- |
| 6EEB5903-E8FC-46D4-84CE-083ACE1C1E59 | 2CF00F52-126B-4950-88A8-69E65A531044 | C245D4CF-32A2-42DF-A3BA-6B59EC01913D |

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good — merging.

If we want to make further tweaks to the profile screen (like increasing the label's width further, or reducing their font weight), that'll be easy to do as a followup.

Issue #684, opened 2024-05-16
Parse topic permalinks


gnprice:
When parsing an internal link, we should handle the upcoming "topic permalink" feature:
* https://github.com/zulip/zulip/issues/21505

In the current draft, these look like `/with/NNNN`, with a message ID.

Currently the app doesn't interpret `/near/` links at all (that's #82 and #683). So we should handle these new `/with/` links the same way for now — accept them and ignore them.

In the future when we implement #683, these new links should be included in that behavior, but unlike `/near/` links they won't have the behavior of #82. (They'll have the behavior #80 instead, like other narrow links.)

gnprice:
(Assigned and unassigned because I was looking at the wrong column in [the project board](https://chat.zulip.org/#narrow/stream/378-api-design/topic/topic.20permalinks/near/1810605); what I meant to do was to set the "owner", to place it in Sayed's section of the board. Did that now, and @sm-sayedi please go ahead and self-assign the issue whenever you start working on it.)

Issue #683, opened 2024-05-16
Follow /near/ links through message moves


gnprice:
When a Zulip narrow link has a `/near/` operator in it, this means two things:

(A) Start the view around the given message, rather than at the user's first unread.

(B) If the message has moved — for example if the topic was renamed or resolved or split up — then find where the message is now, and go to that conversation instead of the one specified by any `channel`/`stream` or `topic` operators in the link.

Part A is #82. This issue is for part B

The corresponding zulip-mobile issue, back in 2022 (when /near/ links gained that meaning in the first place), was https://github.com/zulip/zulip-mobile/issues/5306 .

This issue is independent of #82, and we can do them in either order. (Web had part A without B for many years; zulip-mobile has part B but still lacks part A.)

gnprice:
User report in chat: [#mobile > Links to moved topics are broken on Flutter @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Links.20to.20moved.20topics.20are.20broken.20on.20Flutter/near/2310684)

PR #729, opened 2024-06-11
internal_link: Parse topic permalinks

sm-sayedi:
These links (`/with/NNNN`) are just accepted and ignored in this PR, they will be interpreted in #683.

Fixes: #684

### DIFF
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -179,8 +179,9 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
         if (dmIds == null) return null;
         dmElement = ApiNarrowDm(dmIds, negated: negated);
 
-      case _NarrowOperator.near:
-        continue; // TODO(#82): support for near
+      case _NarrowOperator.near: // TODO(#82): support for near
+      case _NarrowOperator.with_: // TODO(#683): support for with
+        continue;
 
       case _NarrowOperator.unknown:
         return null;
@@ -206,6 +207,9 @@ enum _NarrowOperator {
   // 'dm' is new in server-7.0; means the same as 'pm-with'
   dm,
   near,
+  // cannot use `with` as it is a reserved keyword in Dart
+  @JsonValue('with')
+  with_,
   pmWith,
   stream,
   channel,
--- a/lib/model/internal_link.g.dart
+++ b/lib/model/internal_link.g.dart
@@ -11,6 +11,7 @@ part of 'internal_link.dart';
 const _$_NarrowOperatorEnumMap = {
   _NarrowOperator.dm: 'dm',
   _NarrowOperator.near: 'near',
+  _NarrowOperator.with_: 'with',
   _NarrowOperator.pmWith: 'pm-with',
   _NarrowOperator.stream: 'stream',
   _NarrowOperator.channel: 'channel',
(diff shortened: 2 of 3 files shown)


lib/model/internal_link.dart
```diff
@@ -205,6 +206,9 @@ enum _NarrowOperator {
   // 'dm' is new in server-7.0; means the same as 'pm-with'
   dm,
   near,
+  // cannot use lowerCamelCase `with` as it is a reserved keyword in Dart
+  // ignore: constant_identifier_names
+  With,
```
gnprice, line 211: Fun.

Let's call it `with_`. The leading capital reads to me like it's a different kind of thing — like not an enum value at all, but a class or something like a class.

sm-sayedi: It will generate the same "with_" string value in the enum map which is not compatible with the API!
with enum

gnprice: Ah, that can be adjusted, though. The JSON serialization is all nicely customizable with more annotations — for a value of an enum, it's `@JsonValue`:
https://pub.dev/documentation/json_annotation/4.9.0/json_annotation/JsonValue-class.html

sm-sayedi: Done! Thank you.

Also, as this is my first time using the `json_annotation` package, I have a hard time working with it:
When I run the `dart run build_runner watch --delete-conflicting-outputs` command, all of the other auto-generated files are also modified, such as `lib/api/model/events.g.dart`, `lib/api/route/account.g.dart`, and more — and some are deleted such as `lib/host/android_notifications.g.dart`. So far, I have used `git restore file` to reject changes to those files, but I am not sure if I am doing things correctly from the start. Also, the files that these modified files are used in, gives an error like this:
```
Target of URI hasn't been generated: './android_notifications.g.dart'.
Try running the generator that will generate the file referenced by the URI.dart[uri_has_not_been_generated]
```
Even though, I run the previous command.

rajveermalviya (approved):
Looks and works great!
Moving on to mentor review, cc @hackerkid.

lib/model/internal_link.dart
gnprice: (For cross-reference: this [turned out](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.20json_annotation.20unexpected.20behavior/near/1824402) to be a newlines issue, of CRLF vs. LF.)

hackerkid (approved):
Thanks for working on this @sm-sayedi. Changes looks good to me :)

chrisbobbe (commented):
Thanks, @sm-sayedi!

One nit in the commit message, otherwise LGTM.

    internal_link: Parse topic permalinks
    
    These links (`/with/NNNN`) are just accepted and ignored in this PR,
    they will be interpreted in #683.
    
    Fixes: #684

The links themselves aren't accepted and ignored; it's only the `with` operator that's accepted and ignored. So for example (from the added test cases), `/#narrow/pm-with/1,2-group/with/2` is interpreted as a DM narrow with users 1 and 2, just like `/#narrow/pm-with/1,2-group` would be.

I noticed some pre-existing disorganization and duplication in this file; I might try to clean things up as a followup once this is merged.

gnprice:
Thanks @sm-sayedi and reviewers!

This looks good, apart from the point Chris made above about the commit message. I'll fix that up and merge, so I can include this in the upcoming v0.0.16 release to get us all ready for starting to use this new API feature on the server.

PR #730, opened 2024-06-12
Upgrade Flutter and packages (2024-06)

gnprice:
This includes a new feature in `tools/upgrade` to partially automate the Flutter upgrade.

~~There are also a couple of packages that have major-version upgrades to take: `pigeon`, `firebase_core`, `firebase_messaging`. I'll look at those separately as a followup.~~ (**edit:** done, and added to PR.)


chrisbobbe:
Thanks! LGTM except I think a few of these commits have an unintended author field? Please merge at will after looking at that.

gnprice:
Hmm good catch, thanks! Some of them have my personal email instead of my zulip.com email. I don't normally notice that, because we have a `.mailmap` file in the repo.

Looking with `git log --no-mailmap`, there's a number of those, even fairly recently. I think the pattern is that it's commits I made from the Mac on my desk. … And indeed the config there was wrong. Fixed now.

I'll fix those in these commits too, and merge.

PR #733, opened 2024-06-14
tools/upgrade: Avoid system regexps (via Bash); just use Perl

gnprice:
This regexp didn't work on macOS, as discovered by Chris:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.60tools.2Fupgrade.20flutter-local.60/near/1825388

Specifically it looks like the `\s` and `\S` patterns, which mean "whitespace character" and "not-whitespace character" respectively, don't work there.  If I write things like `[^ ]` instead, then this code works.

From `man 7 re_format` on a Mac, it seems the issue is that those features exist only in "enhanced extended REs", not in mere "extended REs".  (Bash relies for its regexps on the system regexp implementation, which is why its behavior can vary like this.)

On GNU/Linux there's no such distinction... but also the man pages (starting from `man 3 regex`) never actually say just what the regexp language is, so that's hardly better.  The closest is a "see also" to `man 1 grep`, which appears to document a regexp language but is evidently incomplete: it doesn't mention `\s` or `\S`, and empirically `grep -E` accepts both.

(... Ah of course, it's in the Info page: `info grep`.  Still, nothing I can find actually says that that's the language libc's `regcomp` and `regexec` implement.)

Anyway.  Evidently the system regexp implementations Bash relies on come with this mess of subtly different, sometimes half-documented, degrees of ancientness in the languages they accept.  In order to avoid having to care about all that, do the matching in Perl, which brings its own full-featured regexp implementation on all platforms. The cost is a bit of mess in getting the data from Bash to Perl and back; so be it.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -21,7 +21,7 @@ packages:
     dependency: transitive
     description: dart
     source: sdk
-    version: "0.3.0"
+    version: "0.3.1"
   analyzer:
     dependency: transitive
     description:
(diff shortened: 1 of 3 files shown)


chrisbobbe:
🤯

chrisbobbe:
Yep, seems to have worked! Thanks! I've pushed the result here, and also a small typo fix. (There are a few ways that typo might have been fixed—please feel free to pick your favorite.)

PR #734, opened 2024-06-14
theme: Add `ThemeExtension` for Figma design variables; add/use two more variables in `AppBar`

chrisbobbe:
It seems helpful to have a central place to put the design variables we export from the Figma design, to help pave the way to implementing it. This PR adds such a place, as a `ThemeExtension`.

It also takes two more color variables from the Figma (`title` and `icon`), to add to the three we've already been using, and applies them to the app bar, since that's something we see marked "ready for dev": https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2957-50895&m=dev

While we're at it, this also applies the Figma's specified font weight, size, and line height to the app-bar title. (Those aren't expressed as variables in the Figma.)

There's still more to do to align the app bar with what we see in the Figma, and I add TODOs about some of this.

| Before | After |
| --- | --- |
| B688D1CA-8C88-44E2-AC57-35EF75F32F47 | 519ECC53-DCB1-453D-A141-D1F5F8E7BCC2 |

### DIFF
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -16,6 +16,10 @@ ThemeData zulipThemeData(BuildContext context) {
       scrolledUnderElevation: 0,
       backgroundColor: const Color(0xfff5f5f5), // `bg-top-bar` in Figma
 
+      actionsIconTheme: const IconThemeData(
+        color: Color(0xff666699), // `icon` in Figma
+      ),
+
       titleTextStyle: TextStyle(
         inherit: false,
         color: const Color(0xff1a1a1a), // `title` in Figma
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -4,9 +4,10 @@ import 'content.dart';
 import 'text.dart';
 
 ThemeData zulipThemeData(BuildContext context) {
+  final designVariables = DesignVariables();
   return ThemeData(
     typography: zulipTypography(context),
-    extensions: [ContentTheme(context)],
+    extensions: [ContentTheme(context), designVariables],
     appBarTheme: AppBarTheme(
       // Set these two fields to prevent a color change in [AppBar]s when
       // there is something scrolled under it. If an app bar hasn't been
@@ -14,16 +15,16 @@ ThemeData zulipThemeData(BuildContext context) {
       // ColorScheme.surfaceContainer for the scrolled-under state and
       // ColorScheme.surface otherwise, and those are different colors.
       scrolledUnderElevation: 0,
-      backgroundColor: const Color(0xfff5f5f5), // `bg-top-bar` in Figma
+      backgroundColor: designVariables.bgTopBar,
 
       // TODO match layout to Figma
-      actionsIconTheme: const IconThemeData(
-        color: Color(0xff666699), // `icon` in Figma
+      actionsIconTheme: IconThemeData(
+        color: designVariables.icon,
       ),
 
       titleTextStyle: TextStyle(
         inherit: false,
-        color: const Color(0xff1a1a1a), // `title` in Figma
+        color: designVariables.title,
         fontSize: 20,
         letterSpacing: 0.0,
         height: (30 / 20),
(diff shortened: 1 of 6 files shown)


chrisbobbe:
In this commit:

```
ui: Follow new Figma in app-bar title style
```

the `titleSpacing: 4` affects the app-bar layout in the lightbox. I think the change is probably OK, and possibly an improvement (in fact that particular app bar might do better with `titleSpacing: 0`)? Here are the screenshots:

| Before | After |
| --- | --- |
| 89D837A8-08F3-4852-8298-0442E2F516EE | BC86B64F-BA40-44E0-BC87-B4B74864B1AA |

chrisbobbe:
Ah hmm, I wonder if it's worth polishing the message list app bar, which I realized is also affected by this change. I think it looks a bit worse now, but I don't think there's a finalized Figma design for it.

| Before | After |
| --- | --- |
| 723EFA50-B392-43EF-AEBB-C813A9D3370F | 4E4DE33F-A843-4568-AB19-A5353DCBCD95 |

chrisbobbe:
> Ah hmm, I wonder if it's worth polishing the message list app bar

I've added a commit that does this:

```
msglist: Tweak app-bar title to look better after recent text-size change
```

It does change the vertical alignment from baseline to center. One concern that came up in the discussion linked from the code comment (https://github.com/zulip/zulip-flutter/pull/219#discussion_r1281120269 ) was that choosing "center" would make the icon look too low. But I don't think it looks too low now (see screenshots below); maybe one factor is our change of font since that older discussion?

Screenshots shown with different values of the system text-size setting:

| Text scaling <1x | Text scaling 1x | Text scaling >1x |
| --- | --- | --- |
| B9676499-0FD9-46E1-BBA6-65D1C27021CD | CCD4E617-4B08-4EAA-9665-E582CEF1034D | 3CA95394-D384-4E37-8C0B-87BAF053823E |
| C0A0BBE3-00AB-4C5C-9A3B-10BE9A0293F0 | 1F80E38D-7906-438B-936A-4D630F2E5C32 | 4D6F6525-64A1-4138-8FB0-50D2E7CBBC30 |
| 7BA7A7F4-E445-4AD9-B70E-9A8C085412AB | 95D12F41-5473-4348-B63A-503C4D7D0989 | 0E9F0FE7-3882-4474-A8A7-A0984E86BBFE |

alya:
I guess the icon starts looking too small in the screenshots on the right; I don't know how fixable that is.

chrisbobbe:
> I guess the icon starts looking too small in the screenshots on the right; I don't know how fixable that is.

Hmm, yeah; the icon is closely associated with the text, so probably it should scale with it, instead of always being the same size? There's a setting for that; it should be easy: https://api.flutter.dev/flutter/widgets/Icon/applyTextScaling.html

lib/widgets/theme.dart
```diff
@@ -42,3 +68,71 @@ ThemeData zulipThemeData(BuildContext context) {
 /// This is chosen as the sRGB midpoint of the Zulip logo's gradient.
 // As computed by Anders: https://github.com/zulip/zulip-mobile/pull/4467
 const kZulipBrandColor = Color.fromRGBO(0x64, 0x92, 0xfe, 1);
+
+/// Variables from the Figma design.
+///
+/// For how to export these from the Figma, see:
+///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2945-49492&t=MEb4vtp7S26nntxm-0
```
gnprice, line 75: Handy link!

gnprice (approved):
Thanks @chrisbobbe! This all LGTM; one nit below, and please go ahead and merge.

I think the behavior shown in the screenshots where the icon doesn't grow when the text is big is actually fine — it's normal and helpful that increasing text size doesn't cause the rest of the UI to expand, other than text. See discussion here:
https://github.com/zulip/zulip-mobile/blob/e352f563ecf2fa9b09b688d5a65b6bc89b0358bc/docs/background/webview.md#why-mostly-px-and-not-rem

Arguably these particular icons are more text-like (like a kind of emoji) and so should scale with the text size after all. But I don't think we need to worry about it.

chrisbobbe:
Thanks! Done.

chrisbobbe:
> I think the behavior shown in the screenshots where the icon doesn't grow when the text is big is actually fine — it's normal and helpful that increasing text size doesn't cause the rest of the UI to expand, other than text.

Yeah; true. Maybe worth mentioning that we can apply custom scalers where we want to: for example, if we wanted to let icons respond to the text-size setting only when it's set smaller than the default 1x.

Issue #736, opened 2024-06-14
content/msglist: Horizontal scrollbar in code block sometimes has wrong vertical position


chrisbobbe:
In the code block in this screenshot, the scroll bar should be positioned lower than it is:

<img width=400 src="https://github.com/zulip/zulip-flutter/assets/22248748/7af836e0-11aa-4566-a903-0c38adc86093">

I think it might have to do with the bottom inset. I seem to be seeing it in the "Combined feed" but not any other message-list views. "Combined feed" behaves differently about padding the bottom inset because it doesn't show the compose box.

chrisbobbe:
Looks like this broke in this commit:

6a8cf5cf3 msglist [nfc]: Unpack to a CustomScrollView of a single sliver

From `CustomScrollView`'s dartdoc:

```dart
/// [CustomScrollView]s don't automatically avoid obstructions from [MediaQuery]
/// like [ListView]s do. To reproduce the behavior, wrap the slivers in
/// [SliverSafeArea]s.
````

Here's before and after that commit (which was in January, so the appearance won't be current in either screenshot):

| Before | After |
| --- | --- |
| ![C6CEB09A-7695-4087-87AA-D1089732AC8A](https://github.com/zulip/zulip-flutter/assets/22248748/d7f9e8ee-5cce-414e-94e8-3cfffe325c36) | ![2D3A22ED-DA45-4920-9E2D-0BCDDB59DAF5](https://github.com/zulip/zulip-flutter/assets/22248748/1b760fa3-4fbd-4801-8a51-d6799fd16053) |

chrisbobbe:
We could fix it in `main` by wrapping our `SliverStickyHeaderList` sliver in a `SliverSafeArea`, when the narrow is "Combined feed". (And add a `TODO(#311)` like the ones we have already.)

@gnprice, would this approach disrupt your work toward #82?

Then for a regression test, maybe we render a `CodeBlock` in the message list on the "Combined feed" page, and check that it receives a `MediaQuery.padding.bottom` of zero.

gnprice:
Hmm interesting, good catch.

Wrapping in a `SliverSafeArea` would mean that the actual list stays out of the bottom inset area, right? That doesn't seem like the desired behavior — I think it's expected that the messages scroll visibly through that bottom area.

A `MediaQuery.removePadding` might be a good fix. That would mean that the list forgets that there is a bottom inset.

chrisbobbe:
[...]

PR #738, opened 2024-06-17
content/msglist: Fix vertical position of horizontal scrollbar in codeblock

chrisbobbe:
In the issue, Greg said:

https://github.com/zulip/zulip-flutter/issues/736#issuecomment-2169118690
> Wrapping in a `SliverSafeArea` would mean that the actual list stays out of the bottom inset area, right? That doesn't seem like the desired behavior — I think it's expected that the messages scroll visibly through that bottom area.

But that's not what I'm seeing. There is a visible change: the blank space that you can scroll into view, under the messages, is now taller by the height of the bottom inset. (Before, the height was just 36px hard-coded.) But you can still scroll this blank space out of view, and when you do, the messages scroll visibly through that bottom area.

| Before | After |
| --- | --- |
| image | image |

GIF from after:

Fixes: #736

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -301,7 +301,7 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
     final length = model!.items.length;
     const centerSliverKey = ValueKey('center sliver');
 
-    final sliver = SliverStickyHeaderList(
+    Widget sliver = SliverStickyHeaderList(
       headerPlacement: HeaderPlacement.scrollingStart,
       delegate: SliverChildBuilderDelegate(
         // To preserve state across rebuilds for individual [MessageItem]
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Ah and of course screenshots of the bugfix itself:

| Before | After |
| --- | --- |
| image | Screenshot 2024-06-17 at 4 14 07 PM |

gnprice:
I see. I guess it makes sense that that's what happens once I think about it again, and particularly after looking at the implementation of `SliverSafeArea` — it turns out to be quite simple, just a `SliverPadding` and a `MediaQuery.removePadding`:
```dart
    assert(debugCheckHasMediaQuery(context));
    final EdgeInsets padding = MediaQuery.paddingOf(context);
    return SliverPadding(
      padding: EdgeInsets.only(
        left: math.max(left ? padding.left : 0.0, minimum.left),
        top: math.max(top ? padding.top : 0.0, minimum.top),
        right: math.max(right ? padding.right : 0.0, minimum.right),
        bottom: math.max(bottom ? padding.bottom : 0.0, minimum.bottom),
      ),
      sliver: MediaQuery.removePadding(
        context: context,
        removeLeft: left,
        removeTop: top,
        removeRight: right,
        removeBottom: bottom,
        child: sliver,
      ),
    );
  }
```

Seems fine; merging. Thanks for the fix!

I don't love the extra space this causes below the mark-read button, but I think that's mostly that I don't love that extra space in general; this can reasonably be seen as just bringing the combined-feed narrow in line with the others in showing that extra space.

Issue #393, opened 2023-11-18
model: Track/memoize stream-color variants somewhere other than `/api/model`


chrisbobbe:
Currently (well, after #381), the `Subscription` class has a `colorSwatch` method to return a memoized `StreamColorSwatch`. That code is in `/api/model`.

Probably there's a better, more UI-focused place to do those computations (adjusting lightness, opacity, etc.) and provide them to consumers: https://github.com/zulip/zulip-flutter/pull/381#discussion_r1396528348

chrisbobbe:
What if we made one simple global cache, as a `Map<int, StreamColorSwatch>` keyed by `Subscription.color` values? This might make sense, because a `StreamColorSwatch` isn't stateful and it's computed purely from a `Subscription.color` value.

When UI code is processing a subscription with `.color` `0xff76ce90`, it can call a function that adds an entry for `0xff76ce90` if it's missing, and returns the corresponding `StreamColorSwatch`.

It's possible for that entry to fall into disuse at some point during the app session; for example, if
- the ~~stream's~~ channel's color is changed
- the user leaves the channel
- the user logs out of the org with the channel

But I don't see really significant downsides to that.

gnprice:
Those leaks should be basically fine in practice, since those changes aren't common and since this cache wouldn't outlast the current process. Still, it'd be nicer to have something that doesn't leak.

How about just indexing on the stream ID? (As part of the `PerAccountStore`.) Then there'd be no leaks.

If the user has lots of subscriptions sharing the same color, then that'd give up opportunities to reuse the results. But that's probably not common, and in any event the worst case is just one `StreamColorSwatch` per subscription.

chrisbobbe:
It would be nice for UI elements to consume these color swatches via [`AnimatedTheme`](https://api.flutter.dev/flutter/material/AnimatedTheme-class.html) / [`ThemeExtension`](https://api.flutter.dev/flutter/material/ThemeExtension-class.html), so they would animate between their light- and dark-mode colors when you toggle dark mode, just like other style attributes that are consumed via `AnimatedTheme`. If the cache is global instead of per-account, the natural way to do that is through `MaterialApp.theme`, which configures an `AnimatedTheme` built by `MaterialApp`.

[...]

PR #746, opened 2024-06-18
ui: Provide [StreamColorSwatch]es via DesignVariables, not api/model/

chrisbobbe:
(Instead of putting them in Subscription model objects.)

I wonder if there's a better approach than this. One thing is that having separate subclasses `StreamColorSwatchesLight` and `StreamColorSwatchesDark` means some duplicated boilerplate.

Related: #95
Fixes: #393


test/widgets/stream_colors_test.dart
```diff
@@ -0,0 +1,510 @@
...
+        final swatch2 = instance.forBaseColor(base2);
+        check(swatch2).equals(StreamColorSwatch.lerp(
+          StreamColorSwatch.light(base2), StreamColorSwatch.dark(base2), 0.4)!);
+        check(instance.forBaseColor(base2)).identicalTo(swatch2);
+        check(instance.forBaseColor(base1)).identicalTo(swatch1);
+      });
+
+      test('from dark to light', () {
+        final instance = StreamColorSwatchesDark.instance
+          .lerp(StreamColorSwatchesLight.instance, 0.4);
```
gnprice, line 72: I think this test is redundant with the one above it — there isn't any logic in the lerping that works differently between dark and light, so these are just two distinct groups of swatches whichever way around they are.

test/widgets/theme_test.dart
```diff
@@ -68,4 +75,40 @@ void main() {
     doCheck('TextButton',
       button: TextButton(onPressed: () {}, child: const Text(buttonText)));
   });
+
+  group('colorSwatchFor', () {
+    void doTest(int baseColor, Brightness brightness) {
+      final description = '${Color(baseColor).toString()}, $brightness';
```
gnprice, line 81: nit:
```suggestion
      final description = '${Color(baseColor)}, $brightness';
```

test/widgets/theme_test.dart
```diff
@@ -68,4 +75,40 @@ void main() {
...
+        navigator.push(MaterialWidgetRoute(page: Builder(builder: (context) {
+          actualSwatch = colorSwatchFor(context, subscription);
+          return const Placeholder();
+        })));
+        await tester.pumpAndSettle();
+
+        final expectedSwatch = switch (brightness) {
+          Brightness.light => StreamColorSwatch.light(baseColor),
+          Brightness.dark  => StreamColorSwatch.dark(baseColor),
+        };
```
gnprice, line 104: Do you want to pass an `expected` parameter instead? I'm not sure how this `switch (brightness)` will handle the lerping case in the TODO below. (And the `dark` case is relevant only as part of implementing that TODO.)

gnprice (commented):
Thanks! Glad to get this moved out of `lib/api/`. Comments below.

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,252 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_color_models/flutter_color_models.dart';
+
+import '../api/model/model.dart';
+import 'color.dart';
+
+/// A caching factory for [StreamColorSwatch]es, from [subscription.color] bases.
```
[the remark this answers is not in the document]
chrisbobbe: You give it a `[subscription.color]`, and it returns a `[StreamColorSwatch]` for it. If there's one cached, it returns that, otherwise it creates one, caches it, and returns it.

(I guess "creates one" is just what I was trying to get at with "factory".)

chrisbobbe:

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,223 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_color_models/flutter_color_models.dart';
+
+import '../api/model/model.dart';
+import 'color.dart';
+
+/// A provider for [StreamColorSwatch]es that computes and caches them.
+abstract class StreamColorSwatches {
```
gnprice, line 10: I think "provider", like "factory", is a word I'd like to avoid here in favor of unpacking what it means :slightly_smiling_face:

Working from your explanation at https://github.com/zulip/zulip-flutter/pull/746#discussion_r1645230417 , here's a version I think would be clear:
```suggestion
/// Computes a [StreamColorSwatch] for any given color, with caching.
abstract class StreamColorSwatches {
```

Perhaps better yet, here's a slightly different way of looking at it:

```suggestion
/// A lazily-computed map from a stream's base color to a
/// corresponding [StreamColorSwatch].
abstract class StreamColorSwatches {
```

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,223 @@
...
+  ///
+  /// If [this] and [other] are [identical], returns [this].
+  ///
+  /// Else returns an instance whose [forBaseColor] will call
+  /// [this.forBaseColor] and [other.forBaseColor]
+  /// and return [StreamColorSwatch.lerp]'s result on those.
+  /// This computation is cached on the instance
+  /// in order to save work building [t]'s animation frame when there are
+  /// multiple UI elements using the same [subscription.color].
+  StreamColorSwatches lerp(StreamColorSwatches? other, double t) {
```
gnprice, line 32: Perhaps make this static? That seems to be the convention elsewhere, or at least on `Color`.

(I had this thought when looking at the top of the class declaration, where there's `light` and `dark` and what one basically wants to say is that we have three kinds of instances of this class: light, dark, and lerps between them.)

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,223 @@
...
+
+/// A provider for [StreamColorSwatch]es that computes and caches them.
+abstract class StreamColorSwatches {
+  static final StreamColorSwatches light = _StreamColorSwatchesLight._();
+  static final StreamColorSwatches dark = _StreamColorSwatchesDark._();
+
+  final Map<int, StreamColorSwatch> _cache = {};
+
+  /// Gives the cached [StreamColorSwatch] for a [subscription.color].
+  StreamColorSwatch forBaseColor(int base) =>
```
gnprice, line 17:
```suggestion
  /// Gives the [StreamColorSwatch] for a [subscription.color].
  StreamColorSwatch forBaseColor(int base) =>
```

I wouldn't say this gives "the cached swatch", because there may not have been a cached swatch yet when it was called.

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,223 @@
...
+    _cache[base] ??= computeForBaseColor(base);
+
+  StreamColorSwatch computeForBaseColor(int base);
```
gnprice, line 20: Could also make this private, as nothing should be calling it directly other than the implementation of `forBaseColor`.

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,223 @@
...
+      return this;
+    }
+
+    return _StreamColorSwatchesLerped(this, other, t);
+  }
+}
+
+/// The [StreamColorSwatches] for the light theme.
+class _StreamColorSwatchesLight extends StreamColorSwatches {
+  _StreamColorSwatchesLight._();
```
gnprice, line 48: nit:
```suggestion
  _StreamColorSwatchesLight();
```

Has the same effect since the class is already private, and makes the syntax a bit cleaner at the call site.

lib/widgets/stream_colors.dart
```diff
@@ -0,0 +1,223 @@
...
+    if (identical(this, other)) {
+      return this;
+    }
+
+    return _StreamColorSwatchesLerped(this, other, t);
+  }
+}
+
+/// The [StreamColorSwatches] for the light theme.
+class _StreamColorSwatchesLight extends StreamColorSwatches {
```
gnprice, line 47: nit: move this doc to the static instance, which is public

gnprice (commented):
Thanks for the revision! Small comments, largely nits.

lib/widgets/stream_colors.dart
chrisbobbe: Cool, yeah; I like the second one.

lib/widgets/stream_colors.dart
chrisbobbe: Sure, I can try that. I think I did non-static because that's what `ThemeExtension` does. (`CodeBlockTextStyles` came out non-static for that reason as well, I think.)

chrisbobbe: `TextStyle`'s lerp is also static.

…Hmm, although I guess maybe `CodeBlockTextStyles` and `StreamColorSwatches` are more like `ThemeExtension` than like `Color` and `TextStyle`?

One thing about `CodeBlockTextStyles` and `StreamColorSwatches` is that their lerps are only ever called by a `ThemeExtension`'s lerp. In that lerp, the `a` instance (i.e., `this`) is non-nullable, and that's convenient for `_StreamColorSwatchesLerped.computeForBaseColor` because it makes it easy to reason locally that its non-nullable `StreamColorSwatch` return type is correct.

chrisbobbe:
Thanks for the review! Revision pushed, and I replied above: https://github.com/zulip/zulip-flutter/pull/746#discussion_r1649326017

lib/widgets/stream_colors.dart
gnprice: > I think I did non-static because that's what `ThemeExtension` does.

Hmm I see.

I just did a quick survey of the Flutter tree. It looks like the pattern is very consistent, and `ThemeExtension` is an outlier. So let's follow the typical pattern and make it static.

Specifically the search I did was, in the Flutter repo:
```
$ git grep ' lerp[<(]' packages/flutter/lib/
```

That finds definitions of a method named `lerp`, static or otherwise.

There are a number of non-static `lerp` methods… on `Tween` subclasses and the like, that look like this:
```
packages/flutter/lib/src/rendering/tweens.dart
27:  FractionalOffset? lerp(double t) => FractionalOffset.lerp(begin, end, t);
50:  Alignment lerp(double t) => Alignment.lerp(begin, end, t)!;
75:  AlignmentGeometry? lerp(double t) => AlignmentGeometry.lerp(begin, end, t);
```
So e.g. `AlignmentTween.lerp` is non-static; but it's not a lerp between `AlignmentTween`s, rather a lerp between `Alignment`s, and in fact it forwards to the static `Alignment.lerp`.

There are something like 90 static `lerp` methods that take both endpoints as arguments, and like 40 non-static `lerp` methods that take neither as arguments, getting them both from the receiver (as with those Tween subclasses). And I think `ThemeExtension` is the sole example of its pattern, where it gets one endpoint from the receiver and the other as an argument.

gnprice: (This also means that that API choice on `ThemeExtension` was basically a mistake.)

gnprice:
Thanks! All LGTM except that one open thread above.

I'll go ahead and merge this, since that's something straightforward to fix up separately. Then as a followup let's fix both this `lerp` method and the existing other `lerp` methods to be static, as discussed in that thread https://github.com/zulip/zulip-flutter/pull/746#discussion_r1649474626 .

lib/widgets/stream_colors.dart
chrisbobbe: Thanks for that investigation!

chrisbobbe:
Cool; sure, I'll follow up with a fix to make them static.

PR #748, opened 2024-06-18
compose_box [nfc]: Update some stale issue references

chrisbobbe:
I'm pretty sure this change is right, but I'm sending it as a PR instead of pushing directly, in case there's some subtlety I'm missing.


gnprice:
Thanks! Yep, looks right — merging.

PR #750, opened 2024-06-18
msglist: Support data model for edited/moved marker

PIG208:
#171 

TODOs:
- [x] Handle message update events
- [x] <del>Support in UI (deferred to a separate PR)</del>

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -464,6 +464,9 @@ sealed class Message {
   final String contentType;
 
   // final List<MessageEditHistory> editHistory; // TODO handle
+  @JsonKey(readValue: MessageEditState._readFromMessage, fromJson: Message._messageEditStateFromJson)
+  MessageEditState editState;
+
   final int id;
   bool isMeMessage;
   int? lastEditTimestamp;
@@ -490,6 +493,12 @@ sealed class Message {
   @JsonKey(name: 'match_subject')
   final String? matchTopic;
 
+  static MessageEditState _messageEditStateFromJson(dynamic json) {
+    // This is a no-op so that [MessageEditState._readFromMessage]
+    // can return the enum value directly.
+    return json as MessageEditState;
+  }
+
   static Reactions? _reactionsFromJson(dynamic json) {
     final list = (json as List<dynamic>);
     return list.isNotEmpty ? Reactions.fromJson(list) : null;
@@ -508,6 +517,7 @@ sealed class Message {
     required this.client,
     required this.content,
     required this.contentType,
+    required this.editState,
     required this.id,
     required this.isMeMessage,
     required this.lastEditTimestamp,
@@ -573,6 +583,7 @@ class StreamMessage extends Message {
     required super.client,
     required super.content,
     required super.contentType,
+    required super.editState,
     required super.id,
     required super.isMeMessage,
     required super.lastEditTimestamp,
@@ -675,6 +686,7 @@ class DmMessage extends Message {
     required super.client,
     required super.content,
     required super.contentType,
+    required super.editState,
     required super.id,
     required super.isMeMessage,
     required super.lastEditTimestamp,
@@ -697,4 +709,79 @@ class DmMessage extends Message {
 
   @override
   Map<String, dynamic> toJson() => _$DmMessageToJson(this);
+}
+
+enum MessageEditState {
+  none,
+  edited,
+  moved;
+
+  // Adapted from the shared code:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
+  // The canonical resolved-topic prefix.
+  static const String _resolvedTopicPrefix = '✔ ';
+
+  /// Whether the given topic move reflected either a "resolve topic"
+  /// or "unresolve topic" operation.
+  ///
+  /// The Zulip "resolved topics" feature is implemented by renaming the topic;
+  /// but for purposes of [Message.editState], we want to ignore such renames.
+  /// This method identifies topic moves that should be ignored in that context.
+  static bool topicMoveWasResolveOrUnresolve(String topic, String prevTopic) {
+    if (topic.startsWith(_resolvedTopicPrefix)
+        && topic.substring(_resolvedTopicPrefix.length) == prevTopic) {
+      return true;
+    }
+
+    if (prevTopic.startsWith(_resolvedTopicPrefix)
+        && prevTopic.substring(_resolvedTopicPrefix.length) == topic) {
+      return true;
+    }
+
+    return false;
+  }
+
+  static MessageEditState _readFromMessage(Map<dynamic, dynamic> json, String key) {
+    // Adapted from `analyze_edit_history` in the web app:
+    //   https://github.com/zulip/zulip/blob/c31cebbf68a93927d41e9947427c2dd4d46503e3/web/src/message_list_view.js#L68-L118
+    final editHistory = json['edit_history'] as List<dynamic>?;
+    final lastEditTimestamp = json['last_edit_timestamp'] as int?;
+    if (editHistory == null) {
+      return (lastEditTimestamp != null)
+        ? MessageEditState.edited
+        : MessageEditState.none;
+    }
+
+    // Edit history should never be empty whenever it is present
+    assert(editHistory.isNotEmpty);
+
+    bool hasMoved = false;
+    for (final entry in editHistory) {
+      if (entry['prev_content'] != null) {
+        return MessageEditState.edited;
+      }
+
+      if (entry['prev_stream'] != null) {
+        hasMoved = true;
+        continue;
+      }
+
+      // TODO(server-5) prev_subject was the old name of prev_topic on pre-5.0 servers
+      final prevTopic = (entry['prev_topic'] ?? entry['prev_subject']) as String?;
+      final topic = entry['topic'] as String?;
+      if (prevTopic != null) {
+        // TODO(server-5) pre-5.0 servers do not have the 'topic' field
+        if (topic == null) {
+          hasMoved = true;
+        } else {
+          hasMoved |= !topicMoveWasResolveOrUnresolve(topic, prevTopic);
+        }
+      }
+    }
+
+    if (hasMoved) return MessageEditState.moved;
+
+    // This can happen when a topic is resolved but nothing else has been edited
+    return MessageEditState.none;
+  }
 }
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -125,6 +125,11 @@ class MessageStoreImpl with MessageStore {
     if (message == null) return;
 
     message.flags = event.flags;
+    if (event.origContent != null) {
+      // The message is guaranteed to be edited.
+      // See also: https://zulip.com/api/get-events#update_message
+      message.editState = MessageEditState.edited;
+    }
     if (event.renderedContent != null) {
       assert(message.contentType == 'text/html',
         "Message contentType was ${message.contentType}; expected text/html.");
@@ -168,10 +173,26 @@ class MessageStoreImpl with MessageStore {
       return;
     }
 
-    // final newStreamId = event.newStreamId; // null if topic-only move
-    // final newTopic = event.newTopic!;
+    final newTopic = event.newTopic!;
+    final newChannelId = event.newStreamId; // null if topic-only move
+
+    if (newChannelId == null
+        && MessageEditState.topicMoveWasResolveOrUnresolve(event.origTopic!, newTopic)) {
+      // The topic was only resolved/unresolved.
+      // No change to the messages' editState.
+      return;
+    }
+
     // TODO(#150): Handle message moves.  The views' recipient headers
     //   may need updating, and consequently showSender too.
+    //   Currently only editState gets updated.
+    for (final messageId in event.messageIds) {
+      final message = messages[messageId];
+      if (message == null) continue;
+      // Do not override the edited marker if the message has also been moved.
+      if (message.editState == MessageEditState.edited) continue;
+      message.editState = MessageEditState.moved;
+    }
   }
 
   void handleDeleteMessageEvent(DeleteMessageEvent event) {
--- a/test/api/model/model_test.dart
+++ b/test/api/model/model_test.dart
@@ -148,6 +148,9 @@ void main() {
       );
       check(m2).flags.deepEquals([MessageFlag.read, MessageFlag.unknown]);
     });
+
+    // Code relevant to messageEditState is tested separately in the
+    // MessageEditState group.
   });
 
   group('DmMessage', () {
@@ -210,6 +213,112 @@ void main() {
         .deepEquals([2, 3, 11]);
       check(parse(withRecipients([user11, user2, user3])).allRecipientIds)
         .deepEquals([2, 3, 11]);
+    });
+  });
+
+  group('MessageEditState', () {
+    Map<String, dynamic> baseJson() => deepToJson(eg.streamMessage()) as Map<String, dynamic>;
+
+    group('Edit history is absent', () {
+      test('Message with no evidence of an edit history -> none', () {
+        check(Message.fromJson(baseJson()..['edit_history'] = null))
+          .editState.equals(MessageEditState.none);
+      });
+
+      test('Message without edit history has last edit timestamp -> edited', () {
+        check(Message.fromJson(baseJson()
+            ..['edit_history'] = null
+            ..['last_edit_timestamp'] = 1678139636))
+          .editState.equals(MessageEditState.edited);
+      });
+    });
+
+    void checkEditState(MessageEditState editState, List<Map<String, dynamic>> editHistory){
+      check(Message.fromJson(baseJson()..['edit_history'] = editHistory))
+        .editState.equals(editState);
+    }
+
+    group('edit history exists', () {
+      test('Moved message has last edit timestamp but no actual edits -> moved', () {
+        check(Message.fromJson(baseJson()
+            ..['edit_history'] = [{'prev_stream': 5, 'stream': 7}]
+            ..['last_edit_timestamp'] = 1678139636))
+          .editState.equals(MessageEditState.moved);
+      });
+
+      test('Channel change only -> moved', () {
+        checkEditState(MessageEditState.moved,
+          [{'prev_stream': 5, 'stream': 7}]);
+      });
+
+      test('Topic name change only -> moved', () {
+        checkEditState(MessageEditState.moved,
+          [{'prev_topic': 'old_topic', 'topic': 'new_topic'}]);
+      });
+
+      test('Both topic and content changed -> edited', () {
+        checkEditState(MessageEditState.edited, [
+          {'prev_topic': 'old_topic', 'topic': 'new_topic'},
+          {'prev_content': 'old_content'},
+        ]);
+        checkEditState(MessageEditState.edited, [
+          {'prev_content': 'old_content'},
+          {'prev_topic': 'old_topic', 'topic': 'new_topic'},
+        ]);
+      });
+
+      test('Both topic and content changed in a single edit -> edited', () {
+        checkEditState(MessageEditState.edited,
+          [{'prev_topic': 'old_topic', 'topic': 'new_topic', 'prev_content': 'old_content'}]);
+      });
+
+      test('Content change only -> edited', () {
+        checkEditState(MessageEditState.edited,
+          [{'prev_content': 'old_content'}]);
+      });
+
+      test("'prev_topic' present without the 'topic' field -> moved", () {
+        checkEditState(MessageEditState.moved,
+          [{'prev_topic': 'old_topic'}]);
+      });
+
+      test("'prev_subject' present from a pre-5.0 server -> moved", () {
+        checkEditState(MessageEditState.moved,
+          [{'prev_subject': 'old_topic'}]);
+      });
+    });
+
+    group('topic resolved in edit history', () {
+      test('Topic was only resolved -> none', () {
+        checkEditState(MessageEditState.none,
+          [{'prev_topic': 'old_topic', 'topic': '✔ old_topic'}]);
+      });
+
+      test('Topic was resolved but the content changed in the history -> edited', () {
+        checkEditState(MessageEditState.edited, [
+          {'prev_topic': 'old_topic', 'topic': '✔ old_topic'},
+          {'prev_content': 'old_content'},
+        ]);
+      });
+
+      test('Topic was resolved but it also moved in the history -> moved', () {
+        checkEditState(MessageEditState.moved, [
+          {'prev_topic': 'old_topic', 'topic': 'new_topic'},
+          {'prev_topic': '✔ old_topic', 'topic': 'old_topic'},
+        ]);
+      });
+
+      test('Topic was moved but it also was resolved in the history -> moved', () {
+        checkEditState(MessageEditState.moved, [
+          {'prev_topic': '✔ old_topic', 'topic': 'old_topic'},
+          {'prev_topic': 'old_topic', 'topic': 'new_topic'},
+        ]);
+      });
+
+      test('Unresolving topic with a weird prefix -> moved', () {
+          checkEditState(MessageEditState.moved,
+            [{'prev_topic': '✔ ✔old_topic', 'topic': 'old_topic'}]);
+      });
     });
   });
 }
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -437,6 +437,38 @@ UpdateMessageEvent updateMessageEditEvent(
   );
 }
 
+UpdateMessageEvent updateMessageMoveEvent(
+  List<Message> messages, {
+  int? newStreamId,
+  String? origTopic,
+  String? newTopic,
+  String? origContent,
+  String? newContent,
+}) {
+  assert(messages.isNotEmpty);
+  final origMessage = messages[0];
+  final messageId = origMessage.id;
+  return UpdateMessageEvent(
+    id: 0,
+    userId: selfUser.userId,
+    renderingOnly: false,
+    messageId: messageId,
+    messageIds: messages.map((message) => message.id).toList(),
+    flags: origMessage.flags,
+    editTimestamp: 1234567890, // TODO generate timestamp
+    origStreamId: origMessage is StreamMessage ? origMessage.streamId : null,
+    newStreamId: newStreamId,
+    propagateMode: null,
+    origTopic: origTopic,
+    newTopic: newTopic,
+    origContent: origContent,
+    origRenderedContent: origContent,
+    content: newContent,
+    renderedContent: newContent,
+    isMeMessage: false,
+  );
+}
+
 UpdateMessageFlagsRemoveEvent updateMessageFlagsRemoveEvent(
   MessageFlag flag,
   Iterable<Message> messages, {
--- a/test/model/message_test.dart
+++ b/test/model/message_test.dart
@@ -207,7 +207,8 @@ void main() {
         ..content.not((it) => it.equals(updateEvent.renderedContent!))
         ..lastEditTimestamp.isNull()
         ..flags.not((it) => it.deepEquals(updateEvent.flags))
-        ..isMeMessage.not((it) => it.equals(updateEvent.isMeMessage!));
+        ..isMeMessage.not((it) => it.equals(updateEvent.isMeMessage!))
+        ..editState.equals(MessageEditState.none);
 
       await store.handleEvent(updateEvent);
       checkNotifiedOnce();
@@ -216,7 +217,8 @@ void main() {
         ..content.equals(updateEvent.renderedContent!)
         ..lastEditTimestamp.equals(updateEvent.editTimestamp)
         ..flags.equals(updateEvent.flags)
-        ..isMeMessage.equals(updateEvent.isMeMessage!);
+        ..isMeMessage.equals(updateEvent.isMeMessage!)
+        ..editState.equals(MessageEditState.edited);
     });
 
     test('ignore when message unknown', () async {
@@ -268,6 +270,79 @@ void main() {
 
     test('rendering-only update does not change timestamp (for old server versions)', () async {
       await checkRenderingOnly(legacy: true);
+    });
+
+    group('Handle message edit state update', () {
+      final message = eg.streamMessage();
+      final otherMessage = eg.streamMessage();
+
+      Future<void> sendEvent(Message message, UpdateMessageEvent event) async {
+        await prepare();
+        await prepareMessages([message, otherMessage]);
+
+        await store.handleEvent(event);
+        checkNotifiedOnce();
+      }
+
+      test('message not moved update', () async {
+        await sendEvent(message, eg.updateMessageEditEvent(message));
+        check(store).messages[message.id].editState.equals(MessageEditState.edited);
+        check(store).messages[otherMessage.id].editState.equals(MessageEditState.none);
+      });
+
+      test('message topic moved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'old topic',
+          newTopic:  'new topic'));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.moved)));
+      });
+
+      test('message topic resolved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'new topic',
+          newTopic:  '✔ new topic'));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.none)));
+      });
+
+      test('message topic unresolved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: '✔ new topic',
+          newTopic:  'new topic'));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.none)));
+      });
+
+      test('message topic both resolved and edited update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'new topic',
+          newTopic:  '✔ new topic 2'));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.moved)));
+      });
+
+      test('message topic both unresolved and edited update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: '✔ new topic',
+          newTopic:  'new topic 2'));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.moved)));
+      });
+
+      test('message stream moved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'topic',
+          newTopic: 'topic',
+          newStreamId:  20));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.moved)));
+      });
+
+      test('message is both moved and updated', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'topic',
+          newTopic: 'topic',
+          newStreamId:  20,
+          origContent: 'old content',
+          newContent: 'new content'));
+        check(store).messages[message.id].editState.equals(MessageEditState.edited);
+        check(store).messages[otherMessage.id].editState.equals(MessageEditState.moved);
+      });
     });
   });
 
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,4 +17,7 @@
 
   // This much more focused automatic fix is helpful, though.
   "files.trimTrailingWhitespace": true,
+
+  // Suppress the warnings that appear for TODOs throughout the codebase.
+  "dart.showTodos": false,
 }
--- a/lib/api/model/model.g.dart
+++ b/lib/api/model/model.g.dart
@@ -265,6 +265,8 @@ StreamMessage _$StreamMessageFromJson(Map<String, dynamic> json) =>
       client: json['client'] as String,
       content: json['content'] as String,
       contentType: json['content_type'] as String,
+      editState: Message._messageEditStateFromJson(
+          MessageEditState._readFromMessage(json, 'edit_state')),
       id: (json['id'] as num).toInt(),
       isMeMessage: json['is_me_message'] as bool,
       lastEditTimestamp: (json['last_edit_timestamp'] as num?)?.toInt(),
@@ -288,6 +290,7 @@ Map<String, dynamic> _$StreamMessageToJson(StreamMessage instance) =>
       'client': instance.client,
       'content': instance.content,
       'content_type': instance.contentType,
+      'edit_state': _$MessageEditStateEnumMap[instance.editState]!,
       'id': instance.id,
       'is_me_message': instance.isMeMessage,
       'last_edit_timestamp': instance.lastEditTimestamp,
@@ -307,6 +310,12 @@ Map<String, dynamic> _$StreamMessageToJson(StreamMessage instance) =>
       'stream_id': instance.streamId,
     };
 
+const _$MessageEditStateEnumMap = {
+  MessageEditState.none: 'none',
+  MessageEditState.edited: 'edited',
+  MessageEditState.moved: 'moved',
+};
+
 DmRecipient _$DmRecipientFromJson(Map<String, dynamic> json) => DmRecipient(
       id: (json['id'] as num).toInt(),
       email: json['email'] as String,
@@ -324,6 +333,8 @@ DmMessage _$DmMessageFromJson(Map<String, dynamic> json) => DmMessage(
       client: json['client'] as String,
       content: json['content'] as String,
       contentType: json['content_type'] as String,
+      editState: Message._messageEditStateFromJson(
+          MessageEditState._readFromMessage(json, 'edit_state')),
       id: (json['id'] as num).toInt(),
       isMeMessage: json['is_me_message'] as bool,
       lastEditTimestamp: (json['last_edit_timestamp'] as num?)?.toInt(),
@@ -346,6 +357,7 @@ Map<String, dynamic> _$DmMessageToJson(DmMessage instance) => <String, dynamic>{
       'client': instance.client,
       'content': instance.content,
       'content_type': instance.contentType,
+      'edit_state': _$MessageEditStateEnumMap[instance.editState]!,
       'id': instance.id,
       'is_me_message': instance.isMeMessage,
       'last_edit_timestamp': instance.lastEditTimestamp,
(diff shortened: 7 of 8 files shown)


test/api/model/model_test.dart
```diff
@@ -628,4 +631,100 @@ void main() {
...
+
+      check(Message.fromJson(baseJson()
+          ..['edit_history'] = null
+          ..['last_edit_timestamp'] = 1678139636).messageEditState
+      ).equals(MessageEditState.edited);
+    });
+
+    test('edit history exists', () {
+      // Only channel name changed: moved
+      check(Message.fromJson(baseJson()
```
gnprice, line 651: These would make good individual test cases, i.e. individual `test` calls:
```suggestion
    group('edit history exists', () {
      test('Only channel name changed: moved', () {
        check(Message.fromJson(baseJson()
```

That finer-grained organization can become extremely helpful when making a change that breaks some of the test cases, because it makes it possible to see a full list all in one go of all the cases the change breaks, as well as the ones it doesn't.

The main cost is in vertical space occupied by the test code: it adds two lines per test case (or three lines, if they weren't already separated by blank lines). That's generally worth it when each case is already 5 lines or so or longer, as these are.

test/api/model/model_test.dart
```diff
@@ -628,4 +631,100 @@ void main() {
...
+        }]
+      ).messageEditState).equals(MessageEditState.moved);
+
+      // Only topic name changed: moved
+      check(Message.fromJson(baseJson()
+        ..['edit_history'] = [{
+          'prev_topic': 'old_topic',
+          'topic': 'new_topic',
+        }]
+      ).messageEditState).equals(MessageEditState.moved);
```
gnprice, line 664: These tests can be made more compact with a local helper like so:
```suggestion
      checkEditState(expected: MessageEditState.moved, [{
        'prev_topic': 'old_topic',
        'topic': 'new_topic',
      }]);
```

Making the test cases more compact like that is helpful in two ways:
* It makes them shorter vertically, so that one can see more of them at once.
* It cuts out elements that are repetitive from test case to test case, so that it's easier to focus on the parts that are interesting and different in each test case.

Both of those are helpful for scanning the list and thinking about what the tests are saying, and trying to think of other cases that aren't yet covered and would be interesting to test.

gnprice (commented):
Thanks!

Quick comments on test style. Both of these are key points for my general approach to writing tests, so I figure they're good to comment on up front.

I haven't read this in detail beyond those two high-level points (and beyond what we talked about in our pairing call today); LMK when it's ready for review or if there are other aspects you'd like feedback on early.

test/api/model/model_test.dart
```diff
@@ -581,6 +581,9 @@ void main() {
       );
       check(m2).flags.deepEquals([MessageFlag.read, MessageFlag.unknown]);
     });
+
+    // Code relevant to messageEditState is tested separately in the
+    // MessageEditState group.
```
PIG208, line 586: Alternatively, we could also nest those tests inside this group. That's basically what I did for the `UpdateMessageTest` groups in the next commit.

PIG208:
This PR is rebased on top of #753 with a commit from there edited (to apply these changes: https://github.com/zulip/zulip-flutter/pull/753#discussion_r1647902456)

test/api/model/model_test.dart
gnprice: I think having its own group at the end of the test file (like it is in this revision) is good. The code being tested is mainly in `MessageEditState`, which is at the end of its file, so having the tests at the end keeps the tests arranged in the same order as the code they're testing. That's helpful when navigating the tests.

gnprice:
Cool. If this is ready for review (or once it's ready), please mark it in GitHub as "ready for review" i.e. no longer a draft PR. Then @chrisbobbe would you do the first round of review on this PR?

test/model/message_test.dart
```diff
@@ -269,6 +271,48 @@ void main() {
...
+      });
+
+      test('message stream moved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent(message,
+          origTopic: 'topic',
+          newTopic: 'topic',
+          newStreamId:  20,
+        ));
+        check(store).messages.values.single.editState.equals(MessageEditState.moved);
+      });
```
chrisbobbe, line 314: Hmm, this test is failing in the current revision. I think it's because the event-handling code unconditionally sets the message's state to `MessageEditState.edited`, in `_handleUpdateMessageEventContent`—that doesn't seem desirable 🙂—and then `_handleUpdateMessageEventMove` leaves that in place; here's from `_handleUpdateMessageEventMove`:

```dart
      // Do not override the edited marker if the message has also been moved.
      if (message.editState == MessageEditState.edited) continue;
```

test/model/message_test.dart
```diff
@@ -269,6 +271,48 @@ void main() {
     test('rendering-only update does not change timestamp (for old server versions)', () async {
       await checkRenderingOnly(legacy: true);
     });
+
+    group('Handle message edit state update', () {
```
chrisbobbe, line 275: For all the kinds of moves, let's also test where the message is in `event.messageIds` but it isn't `event.messageId`.

test/model/message_test.dart
```diff
@@ -269,6 +271,48 @@ void main() {
...
+        check(store).messages.values.single.editState.equals(MessageEditState.moved);
+      });
+
+      test('message topic resolved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent(message,
+          origTopic: 'new topic',
+          newTopic:  '✔ new topic',
+        ));
+        check(store).messages.values.single.editState.equals(MessageEditState.none);
+      });
```
chrisbobbe, line 305: Would also be good to test a topic being unresolved, too.

test/model/message_test.dart
chrisbobbe: Separately, there's a different interesting kind of move that would be good to cover in these tests: a move where _both_ the stream and the topic are changed.

chrisbobbe (changes requested):
Thanks, @PIG208! Comments below.

gnprice:
@PIG208 Useful background reading on some of the comments Chris made above about doc comments:
https://dart.dev/effective-dart/documentation#doc-comments
https://github.com/flutter/flutter/blob/d4cffa25402/docs/contributing/Style-guide-for-Flutter-repo.md#documentation-dartdocs-javadocs-etc

Generally we follow the Flutter style guide for docs, with the exception that because the audience for the docs is smaller (because we're writing an app and not a library), we skip the really high-effort things like sample code and diagrams, and don't hold ourselves to documenting everything. We do generally document things that carry significant non-obvious semantics in them — so `MessageEditState.areSameTopic` in this PR is a good example of where it's quite helpful to be writing docs.

chrisbobbe:
#753 is now merged, so Zixuan please rebase onto current `main`.

lib/api/model/model.dart
```diff
@@ -868,3 +880,84 @@ class DmMessage extends Message {
...
+    // TODO(#744) Extract this to its own home to fully support mark as resolve
+
+    // Code adapted from the shared code: web/shared/src/resolve_topic.ts
+
+    /**
+     * Pattern for an arbitrary resolved-topic prefix.
+     *
+     * These always begin with the canonical prefix, but can go on longer.
+     */
+    // The class has the same characters as RESOLVED_TOPIC_PREFIX.
```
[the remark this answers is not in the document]
PIG208: Those are from the shared code. I assume that "the class" refers the the regex character class `[]` here. Updating this to clarify the comments.

lib/api/model/model.dart
```diff
@@ -868,3 +880,84 @@ class DmMessage extends Message {
...
+    prevTopic = prevTopic.replaceFirst(resolvedTopicPrefixRe, '');
+
+    return topic != prevTopic;
+  }
+
+  static MessageEditState readFromMessage(Map<dynamic, dynamic> json, String key) {
+    return isMoved(json['edit_history'] as List<dynamic>?, json['last_edit_timestamp'] as int?);
+  }
+
+  static MessageEditState isMoved(List<dynamic>? editHistory, int? lastEditTimeStamp) {
```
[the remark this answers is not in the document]
PIG208: The intention was to make it easier for us to parse these information from `Message` instead of `Map<String, dynamic>` in the future. But yeah it can be inlined.

test/model/message_test.dart
PIG208: > Separately, there's a different interesting kind of move that would be good to cover in these tests: a move where _both_ the stream and the topic are changed.

This particular case was covered in "message stream moved update". But I do need to cover the case where a message is edited among multiple moved message.

test/model/message_test.dart
PIG208: Modified `_handleUpdateMessageEventContent` to check `origContent` so that we know if there has been an edit. This should make the message of id `messageId` "edited", but the remaining messages of ids from `messageIds` remain as "moved".

test/model/message_test.dart
PIG208: Done adding the new test. See "message is both moved and updated".

PIG208:
I have rebased and addressed the comments. Thanks for the reviews and the link on dartdoc!

test/api/model/model_test.dart
```diff
@@ -212,4 +215,99 @@ void main() {
...
     });
   });
+
+  group('MessageEditState', () {
+    Map<String, dynamic> baseJson() => deepToJson(eg.streamMessage()) as Map<String, dynamic>;
+
+    test('edit history is absent', () {
+      check(Message.fromJson(baseJson()
+          ..['edit_history'] = null
+      ).editState).equals(MessageEditState.none);
```
chrisbobbe, line 225: Can do this with two lines:

```suggestion
      check(Message.fromJson(baseJson()..['edit_history'] = null))
        .editState.equals(MessageEditState.none);
```

This suggestion also includes changing `.editState).equals` to `).editState.equals`, so using the `editState` getter on `Subject<Message>` (thanks to our `MessageChecks` extension) instead of the one on `Message`. We like to do this because it makes the test's output more expressive on failure.

test/model/message_test.dart
```diff
@@ -269,6 +271,80 @@ void main() {
...
+        await sendEvent(message, eg.updateMessageEditEvent(message));
+        check(store).messages[message.id].editState.equals(MessageEditState.edited);
+        check(store).messages[otherMessage.id].editState.equals(MessageEditState.none);
+      });
+
+      test('message topic moved update', () async {
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'old topic',
+          newTopic:  'new topic',
+        ));
```
chrisbobbe, line 297: Can save a line:
```suggestion
        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
          origTopic: 'old topic',
          newTopic:  'new topic'));
```
(here and below)

test/model/message_test.dart
```diff
@@ -269,6 +271,80 @@ void main() {
...
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'new topic',
+          newTopic:  '✔ new topic',
+        ));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.none)));
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: '✔ new topic',
+          newTopic:  'new topic 2',
+        ));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.moved)));
```
chrisbobbe, line 311: Interesting; so it's like, 'checkmark prefix removed, and another change made', right?

Because of 'checkmark prefix removed', maybe this check makes a better pairing with the first check in the mark-_unresolved_ test?

Maybe better in any case to make a separate test case for this check. The event it's testing is for neither resolving nor unresolving a topic.

test/model/message_test.dart
```diff
@@ -269,6 +271,80 @@ void main() {
...
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: '✔ new topic',
+          newTopic:  'new topic',
+        ));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.none)));
+        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
+          origTopic: 'new topic',
+          newTopic:  '✔ new topic 2',
+        ));
+        check(store).messages.values.every(((message) => message.editState.equals(MessageEditState.moved)));
```
chrisbobbe, line 324: (See previous comment that applies similiarly here.)

chrisbobbe (commented):
Thanks! Comments below, including one or two things I seem to have missed last time. 🙂

lib/model/message.dart
```diff
@@ -125,6 +125,11 @@ class MessageStoreImpl with MessageStore {
...
 
     message.flags = event.flags;
+    // According to the API, the content of the message is edited if origContent
+    // is not null.
+    if (event.origContent != null) {
+      message.editState = MessageEditState.edited;
+    }
     if (event.renderedContent != null) {
       assert(message.contentType == 'text/html',
         "Message contentType was ${message.contentType}; expected text/html.");
```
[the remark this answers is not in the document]
PIG208: Will the `event.userId != null` result in false positives in server 5.0+? It is documented to be always present, so there can be a combination of `renderingOnly == true` and `userId != null` that should not be considered an edit.

lib/model/message.dart
```diff
@@ -143,7 +148,10 @@ class MessageStoreImpl with MessageStore {
     // The interaction between the fields of these events are a bit tricky.
     // For reference, see: https://zulip.com/api/get-events#update_message
 
-    if (event.origTopic == null) {
+    final bool isChannelMoved = event.newStreamId != null;
+    bool isTopicMoved = event.origTopic != null;
+
+    if (!isTopicMoved && !isChannelMoved) {
```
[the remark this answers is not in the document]
PIG208: Yeah, that should be the case.

lib/api/model/model.dart
```diff
@@ -698,3 +710,81 @@ class DmMessage extends Message {
...
+
+    // Pattern for an arbitrary resolved-topic prefix.
+    // These always begin with the canonical prefix, but can go on longer.
+    // It's designed to remove a weird "✔ ✔✔ " prefix, if present.
+    final RegExp resolvedTopicPrefixRe = RegExp('^✔ [ ✔]*');
+
+    topic = topic.replaceFirst(resolvedTopicPrefixRe, '');
+    prevTopic = prevTopic.replaceFirst(resolvedTopicPrefixRe, '');
+
+    return topic != prevTopic;
```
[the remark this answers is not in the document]
PIG208: Oh, good catch!

test/example_data.dart
```diff
@@ -420,6 +420,38 @@ UpdateMessageEvent updateMessageEditEvent(
...
+    editTimestamp: 1234567890, // TODO generate timestamp
+    origStreamId: origMessage is StreamMessage ? origMessage.streamId : null,
+    newStreamId: newStreamId,
+    propagateMode: null,
+    origTopic: origTopic,
+    newTopic: newTopic,
+    origContent: origContent,
+    origRenderedContent: origRenderedContent,
+    content: 'some probably-mismatched new Markdown',
+    renderedContent: origMessage.content,
```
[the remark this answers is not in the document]
PIG208: We do have a test case for the situation where both an update and a move occur in a single event. Should we keep the test? Updated the PR to just add params that default to `null`.

lib/model/message.dart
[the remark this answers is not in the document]
chrisbobbe: Oh, good catch; yeah, that seems like a good solution for that.

test/example_data.dart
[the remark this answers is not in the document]
chrisbobbe: Yeah, that's right; sounds good.

test/api/model/model_test.dart
```diff
@@ -212,4 +215,97 @@ void main() {
...
+    void checkEditState(MessageEditState editState, List<Map<String, dynamic>> editHistory){
+      check(Message.fromJson(baseJson()..['edit_history'] = editHistory))
+        .editState.equals(editState);
+    }
+
+    group('edit history exists', () {
+      test('Channel name change only -> moved', () {
+        checkEditState(MessageEditState.moved, [{
+          'prev_stream': 'old_stream',
+          'stream': 'new_stream',
```
chrisbobbe, line 241: These are meant to be channel IDs, right? Let's use integers, for a more realistic representation.

test/api/model/model_test.dart
```diff
@@ -212,4 +215,97 @@ void main() {
...
+          ..['last_edit_timestamp'] = 1678139636))
+        .editState.equals(MessageEditState.edited);
+    });
+
+    void checkEditState(MessageEditState editState, List<Map<String, dynamic>> editHistory){
+      check(Message.fromJson(baseJson()..['edit_history'] = editHistory))
+        .editState.equals(editState);
+    }
+
+    group('edit history exists', () {
```
chrisbobbe, line 237: Oh, I just noticed: it seems that a single item in the edit-history list could represent _both_ a content change and a move. From the [doc](https://zulip.com/api/get-messages):

> The other fields are optional, and will be present or not depending on whether the channel, topic, and/or message content were modified in the edit event.

So let's add a test that includes both a content change and a move in the same edit-history item. The move could be a topic change or a stream change, or we could have separate tests for each.

chrisbobbe (commented):
Thanks! Comments below, all small, again including some I didn't notice before.

test/api/model/model_test.dart
PIG208: Added the test "Both topic and content changed in a single edit -> edited".

PIG208:
Thanks for the comments! I have updated the PR accordingly.

test/api/model/model_test.dart
```diff
@@ -212,4 +215,85 @@ void main() {
...
+        ]);
+      });
+
+
+      test('Unresolving topic with a weird prefix -> none', () {
+          checkEditState(MessageEditState.none,
+            [{'prev_topic': '✔ ✔old_topic', 'topic': 'old_topic'}]);
+      });
+
+    });
```
chrisbobbe, line 297: nit: remove one empty line above this test, and one below it

chrisbobbe (commented):
Thanks! A small whitespace nit below, otherwise LGTM. Over to @gnprice for his review.

test/api/model/model_test.dart
PIG208: Fixed. Thanks!

lib/api/model/model.dart
```diff
@@ -698,3 +710,78 @@ class DmMessage extends Message {
...
+    bool hasEditedContent = false;
+    bool hasMoved = false;
+    for (final entry in editHistory) {
+      if (entry['prev_content'] != null) {
+        hasEditedContent = true;
+      }
+
+      if (entry['prev_stream'] != null) {
+        hasMoved = true;
+      }
```
gnprice, line 762:
```suggestion
      if (entry['prev_stream'] != null) {
        hasMoved = true;
        continue;
      }
```

That way if it was moved between channels, we skip having to examine any topic change in the same entry.

lib/model/message.dart
```diff
@@ -125,10 +125,21 @@ class MessageStoreImpl with MessageStore {
     if (message == null) return;
 
     message.flags = event.flags;
+    // According to the API, the content of the message is edited if origContent
+    // is not null.
+    if (event.origContent != null) {
+      message.editState = MessageEditState.edited;
```
gnprice, line 131: This comment doesn't really say anything that isn't contained in the code just below it; so best to cut the comment.

Concretely, what the code says is: "if origContent is not null, then the message was edited (and not merely moved)". Then "According to the API" doesn't add any information because this function's job is already to implement this part of the API; if this logic is correct, it's because it corresponds to what the API says.

One version of this comment that would add information would be a link to the relevant spot in the API docs. That's helpful because it gives the reader a specific way to verify that the code matches what the API says. If this function is getting to where it feels like it's using some non-obvious details of the API, then probably it'd be good to put such a link at the top. … In fact I see I did that already with the sibling function `_handleUpdateMessageEventMove`; feel free to do something similar for this function.

lib/model/message.dart
```diff
@@ -125,10 +125,21 @@ class MessageStoreImpl with MessageStore {
...
+    }
     if (event.renderedContent != null) {
       assert(message.contentType == 'text/html',
         "Message contentType was ${message.contentType}; expected text/html.");
       message.content = event.renderedContent!;
+      if (
+        !(event.renderingOnly == true
+        || event.userId == null) // TODO(server-5) can remove userId check
+      ) {
+        message.editState = MessageEditState.edited;
```
gnprice, line 141: Hmm I see.

How does this interact with the `origContent` check above? Do we need both checks?

In light of this detail about `renderingOnly`, is it correct to have that `origContent` logic? It seems like it might cause us to record the state as "edited" when in fact it was a rendering-only change.

lib/model/message.dart
```diff
@@ -125,10 +125,21 @@ class MessageStoreImpl with MessageStore {
...
+    if (event.origContent != null) {
+      message.editState = MessageEditState.edited;
+    }
     if (event.renderedContent != null) {
       assert(message.contentType == 'text/html',
         "Message contentType was ${message.contentType}; expected text/html.");
       message.content = event.renderedContent!;
+      if (
+        !(event.renderingOnly == true
+        || event.userId == null) // TODO(server-5) can remove userId check
```
gnprice, line 139: There's a similar fallback on `renderingOnly` above, in `_handleUpdateMessageEventTimestamp`. This is meant to do the exact same thing, so let's make it look as similar as possible.

Ideally it'd live in just one place, but that requires some refactoring that probably isn't worth it and would make the code uglier in a different direction (passing `isRenderingOnly` around? giving up on having these organized as separate methods?).

If the stakes were higher, then I guess the polished solution would be to use `JsonKey.readValue` so that `event.renderingOnly` had the right value from the beginning. But I think that isn't worth it; these should be the only two places we care about this value, and they're almost right next to each other.

lib/model/message.dart
gnprice: More generally: I think it's probably best for the updates to `editState` to live in one place. In other words, to organize the updates in this function by what they're updating, rather than by what data they're consulting. I think that gives the easiest setup for thinking through each update's logic to check whether it's right.

lib/model/message.dart
```diff
@@ -143,7 +154,9 @@ class MessageStoreImpl with MessageStore {
     // The interaction between the fields of these events are a bit tricky.
     // For reference, see: https://zulip.com/api/get-events#update_message
 
-    if (event.origTopic == null) {
+    bool isTopicUpdated = event.origTopic != null;
+
+    if (!isTopicUpdated) {
       // There was no move.
```
gnprice, line 160: This seems less clear to me.

In fact I think this name is actively misleading: as the comment says, if `origTopic` is null then there was no move — not a change of topic, and not a change of channel either. That's not obvious (which is why there's a comment about it), but it's a fact about the API as documented at the link in the comment above.

lib/model/message.dart
```diff
@@ -168,10 +181,26 @@ class MessageStoreImpl with MessageStore {
...
 
-    // final newStreamId = event.newStreamId; // null if topic-only move
+    final newChannelId = event.newStreamId; // null if topic-only move
+
+    // Eliminate the case when the topic is updated only to be resolved/unresolved.
+    // return early as the messages' editState need no update.
+    if (
+      newChannelId == null
+      && MessageEditState.areSameTopic(event.origTopic!, event.newTopic!)
+    ) return;
```
gnprice, line 191: nit:
```suggestion
    if (newChannelId == null
        && MessageEditState.areSameTopic(event.origTopic!, event.newTopic!)) {
      return;
    }
```

lib/model/message.dart
```diff
@@ -168,10 +181,26 @@ class MessageStoreImpl with MessageStore {
       return;
     }
 
-    // final newStreamId = event.newStreamId; // null if topic-only move
+    final newChannelId = event.newStreamId; // null if topic-only move
+
+    // Eliminate the case when the topic is updated only to be resolved/unresolved.
+    // return early as the messages' editState need no update.
```
gnprice, line 187: I find it's often helpful to put this sort of comment inside the `if` body, rather than above the `if`. That helps it get straight to the point with less need to repeat what the `if` condition says, because the comment's location means it's talking about the case where that condition is true.

So here:
```dart
    if (newChannelId == null
        && MessageEditState.areSameTopic(event.origTopic!, event.newTopic!)) {
      // The topic was only resolved or unresolved.
      // No change to the messages' editState.
      return;
    }
```

I guess that also illustrates another way it helps make the comment more focused: it has less need to repeat what the body of the `if` says either, because the body is right there. Between the two, the comment gets to avoid repeating the "what" (that the code already expresses) and can focus only on the "why".

lib/model/message.dart
```diff
@@ -168,10 +181,26 @@ class MessageStoreImpl with MessageStore {
...
+    if (
+      newChannelId == null
+      && MessageEditState.areSameTopic(event.origTopic!, event.newTopic!)
+    ) return;
+
     // final newTopic = event.newTopic!;
```
gnprice, line 193: can use this commented line too

(or could delete it, that's also fine — but definitely no need to have it here commented out when the code is going and referring to `event.newTopic!` anyway)

gnprice (commented):

Comments below. I've read through all but the last commit, and all but its tests.

Oh, and a commit-message comment:

    message: Handle UpdateMessageEvent for moving/editing.

That's broader than what this commit does — that description would cover actually moving the messages too, i.e. #150. Instead, could say something like:
`message: Update editState on events`

lib/api/model/model.dart
```diff
@@ -490,6 +493,12 @@ sealed class Message {
   @JsonKey(name: 'match_subject')
   final String? matchTopic;
 
+  static MessageEditState _messageEditStateFromJson(dynamic json) {
+    // The value passed here must be a MessageEditState already due to
+    // processing work done in [MessageEditState.readFromMessage].
+    return json as MessageEditState;
```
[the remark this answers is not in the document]
PIG208: Then we will need to have `readValue` return a string. It should work but then we don't have the type checking benefits of using the enum.

lib/model/message.dart
PIG208: Maybe we can just extract the fallback to `UpdateMessageEvent` as a getter?

lib/model/message.dart
PIG208: Checking the server code, `origContent` only appears when there is an actual update. So I think it is safe to assume that the second is not needed.

lib/api/model/model.dart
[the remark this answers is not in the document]
PIG208: Updated the PR so we return strings for the enums.

lib/model/message.dart
PIG208: After removing the edit state update nested in the `if (event.renderedContent != null)` block, it seems sufficient to just explain the `origContent` check and link the API doc there. The other ones seem to be straightforward to understand as-is.

lib/api/model/model.dart
```diff
@@ -698,3 +710,78 @@ class DmMessage extends Message {
...
+
+  static MessageEditState readFromMessage(Map<dynamic, dynamic> json, String key) {
+    // TODO refactor this into a helper that computes this from the serialized
+    // MessageEditHistory.
+    final editHistory = json['edit_history'] as List<dynamic>?;
+    final lastEditTimestamp = json['last_edit_timestamp'] as int?;
+    if (editHistory == null) {
+      return (lastEditTimestamp != null)
+        ? MessageEditState.edited
+        : MessageEditState.none;
```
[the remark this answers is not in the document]
PIG208: Copied a github "perma link" there.

test/api/model/model_test.dart
```diff
@@ -212,4 +215,83 @@ void main() {
...
+        checkEditState(MessageEditState.edited,
+          [{'prev_topic': 'old_topic', 'topic': 'new_topic', 'prev_content': 'old_content'}]);
+      });
+
+      test('Content change only -> edited', () {
+        checkEditState(MessageEditState.edited,
+          [{'prev_content': 'old_content'}]);
+      });
+
+      test("'prev_topic' present without the 'topic' field -> moved", () {
```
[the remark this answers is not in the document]
PIG208: Added a test for the `prev_subject` case.

lib/model/message.dart
PIG208: Now that we don't rely on the logic to set edit state, there is no need to need the fallback any more.

lib/api/model/model.dart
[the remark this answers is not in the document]
gnprice: Oh, I see. Looking at the [`JsonKey.readValue` docs](https://pub.dev/documentation/json_annotation/4.9.0/json_annotation/JsonKey/readValue.html), it says:
> Note: using this feature does not change any of the subsequent decoding logic for the field. For instance, if the field is of type [DateTime](https://api.dart.dev/stable/3.4.4/dart-core/DateTime-class.html) we expect the function provided here to return a [String](https://api.dart.dev/stable/3.4.4/dart-core/String-class.html).

I do like having the function return the enum values, rather than strings. (Especially as these are strings we made up — they're not even in the API.) So let's switch that back.

I think what confused me is that this `_messageEditStateFromJson` method wasn't doing any conversion, so it seemed redundant. Now I think I understand: the point is that by default the generated `fromJson` expects to do a conversion at this step, and this is here in order to avoid doing any conversion… effectively so that the `readValue` can include that conversion as part of its job.

I can try to revise the comment to something that would have successfully conveyed this for me, after you revert to add this back.

lib/model/message.dart
gnprice: Hmm interesting. That's not how I would have interpreted the docs:
> orig_content: string
> Only present if this event changed the message content.

But I guess it can be read the other way too, using "changed the message content" to mean that it wasn't a rendering-only update either.

… Oh, and comparing with what it says at `content`:
> Only present if this event changed the message content or updated the message content for an [inline URL preview](https://zulip.readthedocs.io/en/latest/subsystems/sending-messages.html#inline-url-previews).

I guess the docs are making a contrast there.

So in short I agree this interpretation seems sound. I'll start a thread in `#api documentation` to see about clarifying the docs.

test/api/model/model_test.dart
```diff
@@ -212,4 +215,111 @@ void main() {
...
+        checkEditState(MessageEditState.moved,
+          [{'prev_topic': 'old_topic'}]);
+      });
+
+      // TODO(server-5) prev_subject was the old name of prev_topic on pre-5.0 servers
```
gnprice, line 285: The purpose of a TODO-server comment is to make sure we don't miss a spot where we can delete or simplify code when we drop support for older server versions.

As a result, it's uncommon for tests to need TODO-server comments. The functionality being tested will have its own such comment… and then when we remove that code, normally the point of the test is that it should detect we've done that :slightly_smiling_face:

lib/api/model/model.dart
```diff
@@ -698,3 +710,74 @@ class DmMessage extends Message {
...
+  /// When a topic is resolved, the clients agree on adding a ✔ prefix to the
+  /// topic string. Topics whose only difference is the ✔ prefix are considered
+  /// the same. This helper can be helpful when checking if a message has been
+  /// moved.
+  static bool areSameTopic(String topic, String prevTopic) {
+    // TODO(#744) Extract this to its own home to support "mark as resolve".
+    topic = topic.replaceFirst(_resolvedTopicPrefixRe, '');
+    prevTopic = prevTopic.replaceFirst(_resolvedTopicPrefixRe, '');
+
+    return topic == prevTopic;
```
gnprice, line 737: This seems to be different from the logic that's used in `analyze_edit_history` at that very helpful link you added:
https://github.com/zulip/zulip/blob/c31cebbf68a93927d41e9947427c2dd4d46503e3/web/src/message_list_view.js#L68-L118

Let's adjust this code to have the same behavior as what's in web — the question of "was this message moved or not" is basically part of the Zulip model, so it's good for different clients to agree on that question even in edge cases.

lib/api/model/model.dart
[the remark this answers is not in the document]
PIG208: Got it. Reverted.

lib/api/model/model.dart
```diff
@@ -698,3 +710,74 @@ class DmMessage extends Message {
...
+  }
+
+  static MessageEditState _readFromMessage(Map<dynamic, dynamic> json, String key) {
+    // Adapted from the web app:
+    //   https://github.com/zulip/zulip/blob/c31cebbf68a93927d41e9947427c2dd4d46503e3/web/src/message_list_view.js#L68-L118
```
gnprice, line 742:
```suggestion
    // Adapted from `analyze_edit_history` in the web app:
    //   https://github.com/zulip/zulip/blob/c31cebbf68a93927d41e9947427c2dd4d46503e3/web/src/message_list_view.js#L68-L118
```

The function name isn't strictly needed — the permalink, complete with line numbers, carries all the information and is the most important thing — but is a helpful signpost for a reader who hasn't followed the link.

lib/api/model/model.dart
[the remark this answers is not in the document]
gnprice:
```suggestion
  static MessageEditState _messageEditStateFromJson(dynamic json) {
    // This is a no-op so that [MessageEditState._readFromMessage]
    // can return the enum value directly.
    return json as MessageEditState;
```

gnprice (commented):
Thanks for the revision! (And the revert on that enum-vs-string question.) A few comments.

lib/api/model/model.dart
PIG208: Interestingly, the test

```
      test('Unresolving topic with a weird prefix -> none', () {
          checkEditState(MessageEditState.none,
            [{'prev_topic': '✔ ✔old_topic', 'topic': 'old_topic'}]);
      });
```

fails with the implementation change. This definitely aligns a previously different behavior [with the web app](https://chat.zulip.org/#narrow/stream/7-test-here/topic/test.20move.20weird.20prefix). I will change the expected value to `MessageEditState.moved`.

PIG208:
The PR has been updated to address the reviews. Thanks!

lib/api/model/model.dart
```diff
@@ -698,3 +710,83 @@ class DmMessage extends Message {
...
+  moved;
+
+  // Adapted from the shared code:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
+  // Pattern for an arbitrary resolved-topic prefix.
+  // These always begin with the canonical prefix, but can go on longer.
+  // It's designed to remove a weird "✔ ✔✔ " prefix, if present.
+  static RegExp _resolvedTopicPrefixRe = RegExp('^✔ [ ✔]*');
+
+  static bool _isResovled(String topic) {
```
gnprice, line 726:
```suggestion
  static bool _isResolved(String topic) {
```

lib/api/model/model.dart
```diff
@@ -698,3 +710,83 @@ class DmMessage extends Message {
...
+    return _resolvedTopicPrefixRe.matchAsPrefix(topic) != null;
```
gnprice, line 727: This has the right behavior, I think, but can be simplified:
```suggestion
    return topic.startsWith('✔ ');
```

but probably that prefix should actually be pulled out as a `static const` member on the class so that it gets a name.

That simplification makes it a bit easier to reason through — no thinking about regexps — and should also be more efficient.

lib/api/model/model.dart
```diff
@@ -698,3 +710,83 @@ class DmMessage extends Message {
...
+  }
+
+  /// Whether two topics are equal, ignoring any resolved-topic prefix.
+  ///
+  /// When a topic is resolved, the clients agree on adding a ✔ prefix to the
+  /// topic string. Topics whose only difference is the ✔ prefix are considered
+  /// the same. This helper can be helpful when checking if a message has been
+  /// moved.
+  static bool areSameTopic(String topic, String prevTopic) {
+    // TODO(#744) Extract this to its own home to support "mark as resolve".
```
gnprice, line 737: I'm not sure what the "this" is that this comment refers to.

lib/api/model/model.dart
```diff
@@ -698,3 +710,83 @@ class DmMessage extends Message {
...
+    if (_isResovled(topic) && topic.substring(2) == prevTopic) {
```
gnprice, line 738: The 2 here is kind of a magic number. I think what it really means is `_resolvedTopicPrefix.length`.

lib/api/model/model.dart
```diff
@@ -698,3 +710,83 @@ class DmMessage extends Message {
...
+    return _resolvedTopicPrefixRe.matchAsPrefix(topic) != null;
+  }
+
+  /// Whether two topics are equal, ignoring any resolved-topic prefix.
+  ///
+  /// When a topic is resolved, the clients agree on adding a ✔ prefix to the
+  /// topic string. Topics whose only difference is the ✔ prefix are considered
+  /// the same. This helper can be helpful when checking if a message has been
+  /// moved.
+  static bool areSameTopic(String topic, String prevTopic) {
```
gnprice, line 736:
```suggestion
  static bool sameTopicModuloResolve(String topic, String prevTopic) {
```

That makes it a bit clearer what this method's peculiarity is. For most purposes, if we have the resolved and unresolved version of a topic coexisting, they're distinct topics just the same as if the names were unrelated.

gnprice (commented):
Thanks for the revision! I think all the behavior looks correct now. Various small comments.

lib/api/model/model.dart
PIG208: Moving this to the dartdoc of the method to make it clearer

PIG208:
Thanks for the review @gnprice ! The PR has been updated to address the comments.

lib/api/model/model.dart
```diff
@@ -698,3 +710,82 @@ class DmMessage extends Message {
...
+
+  /// Whether two topics are equal, ignoring any resolved-topic prefix.
+  ///
+  /// When a topic is resolved, the clients agree on adding a ✔ prefix to the
+  /// topic string. Topics whose only difference is the ✔ prefix are considered
+  /// the same. This helper can be helpful when checking if a message has been
+  /// moved.
+  ///
+  /// TODO(#744) Extract to its own home when supporting "mark as resolve".
+  static bool sameTopicModuloResolve(String topic, String prevTopic) {
```
gnprice, line 736: nit: not something relevant for trying to use the function, so belongs as an implementation comment rather than dartdoc:
```suggestion
  /// moved.
  //
  // TODO(#744) Extract to its own home when supporting "mark as resolve".
  static bool sameTopicModuloResolve(String topic, String prevTopic) {
```

gnprice, line 736: Hmm, but is this specific method something we'd want as part of "mark as resolve"? I'm not seeing why it would be.

I think we can probably just leave this TODO comment out; when we go to implement that feature, I think we'll remember that this area of code is here and we'll naturally take care of that.

lib/api/model/model.dart
```diff
@@ -698,3 +710,82 @@ class DmMessage extends Message {
...
+  // Adapted from the shared code:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
+  // The canonical resolved-topic prefix.
+  static const String _resolvedTopicPrefix = '✔ ';
+
+  static bool _isResolved(String topic) {
+    return topic.startsWith(_resolvedTopicPrefix);
+  }
+
+  /// Whether two topics are equal, ignoring any resolved-topic prefix.
```
gnprice, line 728: Rereading this summary line (and the new name I suggested at https://github.com/zulip/zulip-flutter/pull/750#discussion_r1657943540 to match it), and then the implementation, I realize that this isn't actually accurate!

If we'd implemented a method that met this description, then it would be reflexive — `sameTopicModuloResolve(topic, topic)` would be true for any `topic`. But it isn't; in fact that's always `false`.

This is actually good because it also helps resolve the concern that I was otherwise thinking of here (and was going to say should be a TODO comment): in general Zulip topics are case-insensitive, so if this really meant a version of "these topics are equal" or "same topic", then it should be case-insensitive too.

I think what this method actually is about is: does this topic move reflect a resolve or unresolve operation. That explains the lack of case insensitivity: a resolve or unresolve wouldn't have changed case. And it explains the irreflexivity: a resolve or unresolve always changes the topic.

lib/api/model/model.dart
```diff
@@ -698,3 +710,82 @@ class DmMessage extends Message {
...
+  ///
+  /// When a topic is resolved, the clients agree on adding a ✔ prefix to the
+  /// topic string. Topics whose only difference is the ✔ prefix are considered
+  /// the same. This helper can be helpful when checking if a message has been
+  /// moved.
+  ///
+  /// TODO(#744) Extract to its own home when supporting "mark as resolve".
+  static bool sameTopicModuloResolve(String topic, String prevTopic) {
+    if (_isResolved(topic) && topic.substring(_resolvedTopicPrefix.length) == prevTopic) {
```
gnprice, line 737: nit: As is, this logic only really makes sense if you go look at the implementation of `_isResolved` and see that it's checking the topic starts with `_resolvedTopicPrefix` — the second condition doesn't make much sense to ask until one knows that starts-with condition is true.

IOW, for the reader the code only really makes much sense once they mentally inline the definition of `_isResolved`. So it's clearer if we just do that in the first place:
```suggestion
    if (topic.startsWith(_resolvedTopicPrefix)
      && topic.substring(_resolvedTopicPrefix.length) == prevTopic) {
```

gnprice (commented):
Thanks! Small comments below.

One is actually a latent bug ­— the `sameTopicModuloResolve` (or previously `areSameTopic`) method doesn't do what its name and dartdoc say it does.

lib/api/model/model.dart
PIG208: Hmm, this difference is quite subtle. We can rewrite this as "Whether the two topics differs and only differs by whether one of them is resolved or not."

PIG208:
The previous push addresses the review comments and updates the name of the resolved topic helper.

lib/api/model/model.dart
```diff
@@ -698,3 +710,79 @@ class DmMessage extends Message {
...
+  static const String _resolvedTopicPrefix = '✔ ';
+
+  /// Whether the two topics differs and only differs by whether one of them is
+  /// resolved or not.
+  ///
+  /// When a topic is resolved, the clients agree on adding a ✔ prefix to the
+  /// topic string. Topics whose only difference is the ✔ prefix are considered
+  /// the same. This helper can be helpful when checking if a message has been
+  /// moved.
+  static bool topicsDifferByResolvednessOnly(String topic, String prevTopic) {
```
gnprice, line 731: This name and documentation describe how the function is implemented; but the way I'd like to think of what it *means* is a bit higher-level. From https://github.com/zulip/zulip-flutter/pull/750#discussion_r1661612507 :
> I think what this method actually is about is: does this topic move reflect a resolve or unresolve operation.

So here's a version based on that:
```suggestion
  /// Whether the given topic move reflected either a "resolve topic"
  /// or "unresolve topic" operation.
  ///
  /// The Zulip "resolved topics" feature is implemented by renaming the topic;
  /// but for purposes of [Message.editState], we want to ignore such renames.
  /// This method identifies topic moves that should be ignored in that context.
  static bool topicMoveWasResolveOrUnresolve(String topic, String prevTopic) {
```

The name is kind of wordy but I couldn't think of a shorter one that was accurate. And `topicsDifferByResolvednessOnly` happens to be the exact same number of characters :slightly_smiling_face:

gnprice (commented):
Thanks for the revision! This looks good except that the method's name and docs didn't quite get to what I was thinking of with https://github.com/zulip/zulip-flutter/pull/750#discussion_r1661612507 above; details below.

I'll apply that change, and merge.

PR #753, opened 2024-06-18
Add skeleton of handling a move UpdateMessageEvent

gnprice:
This PR is stacked atop #751.

Main commit:

The way the Zulip server API expresses whether there was a move,
and if so what kind of move, is a bit quirky.  This change adds
logic to interpret that information into a more canonical form.

The logic here is translated from what we have in zulip-mobile:
  https://github.com/zulip/zulip-mobile/blob/e352f563e/src/api/misc.js#L26

This doesn't yet do anything to actually update our data structures,
and it's an NFC change for release builds.  The only behavior change
is new asserts that fire when the event from the server is malformed.

---

This is based on part of my draft branch for handling moved messages, #150. This covers the portion of that branch that's involved in just interpreting what an `UpdateMessageEvent` is saying about a message move, which is logic that will be needed in #171 / #750 too — so @PIG208 you'll want to rebase atop this.

The remainder of the draft branch is about actually updating `Message.topic` and `Message.streamId` to reflect a message move, and starting to update the data structures in `MessageListView` accordingly. So those changes will be needed for #150 but aren't involved in #171, and I'll leave them be for now and return to them sometime later.

### DIFF
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -1,5 +1,6 @@
 import '../api/model/events.dart';
 import '../api/model/model.dart';
+import '../log.dart';
 import 'message_list.dart';
 
 /// The portion of [PerAccountStore] for messages and message lists.
@@ -96,8 +97,7 @@ class MessageStoreImpl with MessageStore {
     assert(event.messageIds.contains(event.messageId), "See https://github.com/zulip/zulip-flutter/pull/753#discussion_r1649463633");
     _handleUpdateMessageEventTimestamp(event);
     _handleUpdateMessageEventContent(event);
-    // TODO(#150): Handle message moves.  The views' recipient headers
-    //   may need updating, and consequently showSender too.
+    _handleUpdateMessageEventMove(event);
     for (final view in _messageListViews) {
       view.notifyListenersIfAnyMessagePresent(event.messageIds);
     }
@@ -137,6 +137,41 @@ class MessageStoreImpl with MessageStore {
     for (final view in _messageListViews) {
       view.messageContentChanged(event.messageId);
     }
+  }
+
+  void _handleUpdateMessageEventMove(UpdateMessageEvent event) {
+    // The interaction between the fields of these events are a bit tricky.
+    // For reference, see: https://zulip.com/api/get-events#update_message
+
+    if (event.origTopic == null) {
+      // There was no move.
+      assert(() {
+        if (event.newStreamId != null && event.origStreamId != null
+            && event.newStreamId != event.origStreamId) {
+          // This should be impossible; `orig_subject` (aka origTopic) is
+          // documented to be present when either the stream or topic changed.
+          debugLog('Malformed UpdateMessageEvent: stream move but no origTopic'); // TODO(log)
+        }
+        return true;
+      }());
+      return;
+    }
+
+    if (event.newTopic == null) {
+      // The `subject` field (aka newTopic) is documented to be present on moves.
+      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic')); // TODO(log)
+      return;
+    }
+    if (event.origStreamId == null) {
+      // The `stream_id` field (aka origStreamId) is documented to be present on moves.
+      assert(debugLog('Malformed UpdateMessageEvent: move but no origStreamId')); // TODO(log)
+      return;
+    }
+
+    // final newStreamId = event.newStreamId; // null if topic-only move
+    // final newTopic = event.newTopic!;
+    // TODO(#150): Handle message moves.  The views' recipient headers
+    //   may need updating, and consequently showSender too.
   }
 
   void handleDeleteMessageEvent(DeleteMessageEvent event) {
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -93,32 +93,43 @@ class MessageStoreImpl with MessageStore {
   }
 
   void handleUpdateMessageEvent(UpdateMessageEvent event) {
+    assert(event.messageIds.contains(event.messageId), "See https://github.com/zulip/zulip-flutter/pull/753#discussion_r1649463633");
+    _handleUpdateMessageEventTimestamp(event);
     _handleUpdateMessageEventContent(event);
     // TODO(#150): Handle message moves.  The views' recipient headers
     //   may need updating, and consequently showSender too.
+    for (final view in _messageListViews) {
+      view.notifyListenersIfAnyMessagePresent(event.messageIds);
+    }
+  }
+
+  void _handleUpdateMessageEventTimestamp(UpdateMessageEvent event) {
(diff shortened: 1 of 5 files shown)


test/model/message_test.dart
```diff
@@ -162,6 +162,34 @@ void main() {
...
   group('handleUpdateMessageEvent', () {
+    test('update timestamps on all messages', () async {
+      const t1 = 1718748879;
+      const t2 = t1 + 60;
+      final message1 = eg.streamMessage(lastEditTimestamp: null);
+      final message2 = eg.streamMessage(lastEditTimestamp: t1);
+      // This event is a bit artificial, but convenient.
+      // TODO use a realistic move-messages event here
+      final updateEvent = Event.fromJson({
+        ...eg.updateMessageEditEvent(message1).toJson(),
```
gnprice, line 173: @PIG208 as part of #171 you'll be writing a helper for conveniently making `UpdateMessageEvent` values representing a move (an `eg.updateMessageMoveEvent` similar to the existing `eg.updateMessageEditEvent`). So then you can follow that with a small commit fixing this TODO.

lib/model/message.dart
```diff
@@ -138,6 +138,36 @@ class MessageStoreImpl with MessageStore {
...
+          // This should be impossible; `orig_subject` (aka origTopic) is
+          // documented to be present when either the stream or topic changed.
+          debugLog('Malformed UpdateMessageEvent: stream move but no origTopic'); // TODO(log)
+        }
+        return true;
+      }());
+      return;
+    }
+
+    if (event.newTopic == null) {
```
PIG208, line 156: I found this check a bit confusing (as well as the following one on `origStreamId`) before looking at the API doc. It could be helpful to spell out what we are expecting these checks (something that I can add in my PRs stacked atop this).

gnprice: Hmm yeah, these are a bit terse.

Would this version explain it?
```diff
@@ -139,6 +139,9 @@ class MessageStoreImpl with MessageStore {
   }
 
   void _handleUpdateMessageEventMove(UpdateMessageEvent event) {
+    // The interaction between the fields of these events are a bit tricky.
+    // For reference, see: https://zulip.com/api/get-events#update_message
+
     if (event.origTopic == null) {
       // There was no move.
       assert(() {
@@ -154,10 +157,12 @@ class MessageStoreImpl with MessageStore {
     }
 
     if (event.newTopic == null) {
-      assert(debugLog('Malformed UpdateMessageEvent: move but no topic')); // TODO(log)
+      // The `subject` field (aka newTopic) is documented to be present on moves.
+      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic')); // TODO(log)
       return;
     }
     if (event.origStreamId == null) {
+      // The `stream_id` field (aka origStreamId) is documented to be present on moves.
       assert(debugLog('Malformed UpdateMessageEvent: move but no origStreamId')); // TODO(log)
       return;
     }
```

(That also fixes "no topic" to "no newTopic" — that's from a draft where I'd renamed that field only to "topic" rather than "newTopic".)

PIG208: Yeah this works.

chrisbobbe:
GitHub is showing that this has a conflict; Greg, could you resolve that please?

gnprice:
Sure — rebased.

lib/model/message.dart
```diff
@@ -93,32 +93,42 @@ class MessageStoreImpl with MessageStore {
...
 
   void handleUpdateMessageEvent(UpdateMessageEvent event) {
+    _handleUpdateMessageEventTimestamp(event);
     _handleUpdateMessageEventContent(event);
     // TODO(#150): Handle message moves.  The views' recipient headers
     //   may need updating, and consequently showSender too.
+    for (final view in _messageListViews) {
+      view.notifyListenersIfAnyMessagePresent(event.messageIds);
+    }
   }
```
chrisbobbe:
```
message: Update all edit timestamps on UpdateMessageEvent
```

So in this commit, message lists check `event.messageIds` to see if they need to notify listeners, and they don't check `event.messageId` for that purpose. I think the API doc isn't quite explicit that `message_ids` will include the message ID for a pure content edit: https://zulip.com/api/get-events#update_message

I might be splitting hairs, though. On the `message_id` field, the doc does have this:

> […] the separate `message_ids` field, which is guaranteed to include `message_id`.

…but then again, here's the full sentence that quote is from:

> If the channel or topic was changed, the set of moved messages is encoded in the separate `message_ids` field, which is guaranteed to include `message_id`.

I did just test a pure content edit, though, and the message ID did appear in `message_ids`.

chrisbobbe (commented):
Cool; thanks! One small comment below; if you think it's not an issue, or it's just an API-doc issue, please merge at will.

lib/model/message.dart
PIG208: It does appear to be the case that `message_ids` [always show up](https://github.com/zulip/zulip/blob/be9369541af01557f7ccd35d655324320b30e6be/zerver/actions/message_edit.py#L388).

However, this feels like one of the cases where it will be more convenient if we know that the field should never be missing or nullable. Having our OpenAPI spec [properly defined](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/Outdated.20API.20response.20doc.20in.20get_streams/near/1390715) so that a "only present if ..." field is marked as nullable/required explicitly will definitely help.

chrisbobbe: Ah, so I'm not actually looking at whether the `message_ids` field is present or not; I'm assuming it is present. I'm looking for an API guarantee that, in a pure content-edit event, `message_ids` will include the ID of the edited message, instead of being an empty array. If it can be an empty array, then I expect a message list would fail to show the new content immediately, because we wouldn't be calling `MessageListView.notifyListeners`).

gnprice: Hmm, I agree it'd be good to have that better nailed down. We have this in our API types in zulip-mobile:
```js
  // Any stream/topic changes apply to all of message_ids, which is
  //   guaranteed to include message_id.
  // TODO(server-future): It'd be nice for these to be sorted; NB they may
  //   not be.  As of server 5.0-dev-3868-gca17a452f (2022-02), there's no
  //   guarantee of that in the API nor, apparently, in the implementation.
  message_ids: $ReadOnlyArray<number>,
```

Looks like that dates to zulip/zulip-mobile@4ce9e20b5ba1eba0f560506c1cc835fce1834419 . That's actually after the era where we'd started having good API docs outside the zulip-mobile type definitions. But I don't see in that PR thread https://github.com/zulip/zulip-mobile/pull/5155 , or the [chat thread](chat thread) it links to, any discussion of this particular question.

Anyway, I do think we intend to make that guarantee. I'll ask in the thread and see about making it explicit in the API docs.

gnprice:
Updated to add an assert checking the condition discussed above.

chrisbobbe:
Thanks, LGTM! Merging.

PR #755, opened 2024-06-19
notif: Enable autoCancel flag for summary notification

rajveermalviya:
While trying out the release build from main on a relatively older (Android 11) device, I found that it results in lingering summary notif:

| Android 14 | Android 11 |
| - | - |
| <video src="https://github.com/zulip/zulip-flutter/assets/36819268/17c44e0e-0317-47b9-b379-68817bc32e4b" /> | <video src="https://github.com/zulip/zulip-flutter/assets/36819268/0a4a0850-7630-410a-8dc8-7b8ab774c52f" /> |

Specifying `autoCancel: true` fixes this behavior on the older device.

### DIFF
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -158,6 +158,10 @@ class NotificationDisplayManager {
       inboxStyle: InboxStyle(
         // TODO(#570) Show organization name, not URL
         summaryText: data.realmUri.toString()),
+      // On Android 11 and lower, if autoCancel is not specified,
+      // the summary notification may linger even after all child
+      // notifications have been opened and cleared.
+      autoCancel: true,
     );
   }
 
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -158,9 +158,11 @@ class NotificationDisplayManager {
       inboxStyle: InboxStyle(
         // TODO(#570) Show organization name, not URL
         summaryText: data.realmUri.toString()),
+
       // On Android 11 and lower, if autoCancel is not specified,
       // the summary notification may linger even after all child
       // notifications have been opened and cleared.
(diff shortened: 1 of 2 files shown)


rajveermalviya:
So, I tried the emulators (from Android 12 - Android 9) and turns out it's specifically the behavior on Android 11 (API 30) and lower:
| Android 11 | Android 10 | Android 9 |
| - | - | - |
| api-30 | api-29 | api-28 |

It would have been great if this was documented somewhere ^^. Also the need for an integration test that would automate this and would have caught it — if it was set-up to run on multiple emulators with different Android versions. But definitely not great for each CI run, but could be good to run before each release in CI or locally.

gnprice:
Excellent, thanks for following up and for the thorough investigation!

(Previous thread for cross-reference: https://github.com/zulip/zulip-flutter/pull/703#discussion_r1643498883 .)

Merging this. I'll also add a line in the comment, to use our pattern for marking things we can simplify away when we drop support for an old platform. (We most commonly use this for versions of the Zulip server, the `TODO(server-N)` comments mentioned in the README, but it applies equally well for versions of Android or iOS.)


> It would have been great if this was documented somewhere ^^.

Yep. Android unfortunately isn't real consistent about documenting this sort of thing.


> Also the need for an integration test that would automate this and would have caught it

Yeah, agreed. I'll go file an issue or two about doing that. We're not going to invest in it soon; but I want to record this experience, because it's a great motivating example to have in mind when the launch is complete and we're thinking about what to prioritize next.

gnprice:
> I'll go file an issue or two about doing that.

Done:
* #757 
* #758

rajveermalviya:
> Done

Great, Thanks @gnprice!

Issue #754, opened 2024-06-18
content: Update code-block syntax highlighting styles for light mode


chrisbobbe:
In 2023-12, web changed the set of colors it uses for syntax highlighting in code blocks in light mode. That was after we first implemented light-mode syntax highlighting here, so we're out of sync with web.

However, as Greg points out in https://github.com/zulip/zulip-flutter/pull/689#discussion_r1607195685 , the [discussion](https://chat.zulip.org/#narrow/stream/431-redesign-project/topic/code.20span.20colors/near/1695039) on web's change petered out with some open feedback, and no firm conclusion on what we'd do next.

So, let's first await a color scheme in web that's reasonably stable/happy, then translate it to zulip-flutter.

PR #760, opened 2024-06-22
code_block [nfc]: Update outdated comment

chrisbobbe:
This referred to the dark theme that web was using before it switched to Pygments Monokai, in zulip/zulip#26898. No particular need to explain the choice of background color, especially as we intend to replace these styles anyway; that's #754.


gnprice:
Thanks! LGTM; merging.

Issue #735, opened 2024-06-14
content: @-mentions and global times are scaled too much by system text-size setting


chrisbobbe:
Expected: With the system text size increased, @-mention text and global-time text are scaled appropriately, like surrounding text

Actual: Text in these elements is too large; larger than surrounding text

Discussion on https://github.com/flutter/flutter/issues/126962 about how this could be fixed upstream.

<img width=400 src="https://github.com/zulip/zulip-flutter/assets/22248748/07485333-9c3b-4842-bcf7-aaaac72aa97a">

chrisbobbe:
Tim seems to have noticed this today, after switching to a new device: [#mobile-team > mention styling @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/mention.20styling/near/2169277)

PR #762, opened 2024-06-24
Show marked/edited markers, without swipe gesture

### LINKED ISSUE #616: Star icon too high on messages
In the message list, when a message is starred, the star icon appears too high relative to the message text:
image

See [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/star.20location/near/1769933) for discussion.

Instead, the "shoulders" of the star (the near-horizontal top edges of the left and right wedges of the star) should be at about the x-height of the first line of text, i.e. about the same height as the top of letters like "x" and "e". Like this (from [this PR subthread](https://github.com/zulip/zulip-flutter/pull/449#discussion_r1452645175)):
image

PIG208:
<del>This is still a WIP.</del>

The swipe gesture is implemented such that the marker is
- expanded when there is horizontal movement to the right
- stretched following the touch movement
- retracted when the touch is released

Notably, the content box still re-wraps as the marker pushes it to the right when it expands.To fix this, the content box might need to have a fixed size so the text can go off-screen and remain unchanged. This turned out to be tricky to accomplish.

This is stacked on top of #750. It fixes #616 and fixes part of #171.

https://github.com/zulip/zulip-flutter/assets/39874143/1d57d454-e393-4e2a-a9b8-170d6d328c35


gnprice:
Thanks! Looking forward to having this.

@terpimost @alya for feedback on how the swipe gesture behaves from a user perspective.

> Notably, the content box still re-wraps as the marker pushes it to the right when it expands.To fix this, the content box might need to have a fixed size so the text can go off-screen and remain unchanged. This turned out to be tricky to accomplish.

For this I think the key is, instead of a `Row`, to use something like a [`SlideTransition`](https://main-api.flutter.dev/flutter/widgets/SlideTransition-class.html) (which is what `Dismissible` uses). If the coordinate system that one uses feels inconvenient, then [`PositionedTransition`](https://main-api.flutter.dev/flutter/widgets/PositionedTransition-class.html) might be cleaner. (And we can discuss in more detail in a chat thread.)

alya:
Pushing the context so that part of it gets hidden, rather than rewrapping, makes sense to me.

PIG208:
The text re-wrapping issue has been fixed. I have also added a responsive fix to the star alignment issue. It's a draft because I still need to add tests for it.

gnprice:
Great.

I was using the app this afternoon and happened to have a build from this PR on my phone, and found I really appreciated having the new information about which messages had been edited.

That caused me to think of an idea for how we can split this into a part that delivers the key functionality and that we can merge soon, and a more complex part that can come after it as a follow-up:
* For a first version, we could show just the icons, without the swipe-to-reveal feature.

  That cuts out the gesture handling, the animation, and all the tricky bits about coordinating how the marker's layout and colors change in synchrony with the motion.

  The part that remains will be exactly how we intend the final feature to look in steady state, whenever the user isn't actively pulling on one of these messages to reveal the marker. So it'll also be a good way for us to focus on polishing that design to exactly how we want it, before introducing the complexity of the swipe animation.

  From a user perspective, the one thing that's missing in this version is that it may not be super discoverable what the icons mean. But once you figure it out (perhaps by seeing them on a thread where you have other clues), you know; there's only two of them to learn.

* Then a subsequent PR introduces swipe-to-reveal and the colored markers.

I see the UI logic in this revision looks a lot cleaner, so that follow-up might happen soon after. Even then, though, I think the split will be helpful for reducing the per-PR complexity for review. So let's try making that split.

PIG208:
I extracted a diff that implements the animation to #765. This PR should be ready for review now.

PIG208:
Rebased and added some test cases for the UI change.

chrisbobbe:
GitHub is telling me that some conflicts have appeared; would you resolve those please?

PIG208:
Just rebased the PR.

lib/widgets/edit_state_marker.dart
```diff
@@ -25,11 +25,11 @@ class _SwipableMessageRowState extends State<SwipableMessageRow> {
...
-    // TODO(#157): fix how star marker aligns with message content
-    // Design from Figma at:
-    //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=813%3A28817&mode=dev .
-    var star = Padding(padding: const EdgeInsets.only(top: 5.5),
-      child: Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor));
+    var star = Padding(
+      padding: const EdgeInsets.only(top: 5.5),
+      child: Row(children: [
+          const Text("", style: TextStyle(fontSize: 15)),
+          Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor)]));
```
chrisbobbe, line 32: So normally, to align text-like elements by their baselines, we would use a `Row` with certain arguments passed:

```dart
Row(
  crossAxisAlignment: CrossAxisAlignment.baseline,
  textBaseline: localizedTextBaseline(context),
```

I don't know of a way to do that with a `Stack` instead of a `Row`, though. Do you think you could make it work—including with the animation—using a `Row`, where the first child is the edited/moved marker, the second is the message, and the third is the star? You may need to give the message area a fixed size, but I think that could be done with a `LayoutBuilder`, as in the diff below.

I haven't thoroughly tested this, and I don't know how it might interact with the animation requirement—but an advantage is that we'll keep the baseline alignment even as the message's first line varies in height. Screenshots below, showing a message starting with the line `# heading 1`, which makes a tall h1 element:

<details>
<summary>diff atop the PR branch</summary>

```diff
diff --git lib/widgets/edit_state_marker.dart lib/widgets/edit_state_marker.dart
index fdd40a43b..37fc93c7c 100644
--- lib/widgets/edit_state_marker.dart
+++ lib/widgets/edit_state_marker.dart
@@ -4,6 +4,7 @@ import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 
 import '../api/model/model.dart';
 import 'icons.dart';
+import 'text.dart';
 import 'theme.dart';
 
 class SwipableMessageRow extends StatefulWidget {
@@ -25,26 +26,23 @@ class _SwipableMessageRowState extends State<SwipableMessageRow> {
   Widget build(BuildContext context) {
     final theme = Theme.of(context).extension<DesignVariables>()!;
 
-    var star = Padding(
-      padding: const EdgeInsets.only(top: 5.5),
-      child: Row(children: [
-          const Text("", style: TextStyle(fontSize: 15)),
-          Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor)]));
     final hasMarker = widget.message.editState != MessageEditState.none;
 
-    return Stack(
-      children: [
-        if (hasMarker) Positioned(
 
-    var star = Padding(
-      padding: const EdgeInsets.only(top: 5.5),
-      child: Row(children: [
-          const Text("", style: TextStyle(fontSize: 15)),
-          Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor)]));
     final hasMarker = widget.message.editState != MessageEditState.none;
 
-    return Stack(
-      children: [
-        if (hasMarker) Positioned(
-          left: 0,
-          child: EditStateMarker(editState: widget.message.editState)),
-        Padding(
-          padding: const EdgeInsets.only(left: 16, right: 16),
-          child: widget.child),
-        if (widget.message.flags.contains(MessageFlag.starred))
-          Positioned(
-            right: 0,
-            child: star),
-      ],
+    return LayoutBuilder(
+      builder: (context, constraints) => Row(
+        crossAxisAlignment: CrossAxisAlignment.baseline,
+        textBaseline: localizedTextBaseline(context),
+        children: [
+          hasMarker
+            ? EditStateMarker(editState: widget.message.editState)
+            : const SizedBox(width: 16),
+          SizedBox(width: constraints.maxWidth - 32, child: widget.child),
+          SizedBox(width: 16,
+            child: widget.message.flags.contains(MessageFlag.starred)
+              ? Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor)
+              : null),
+        ],
+      ),
     );
   }
```
</details>

| This PR | with `Row` |
| --- | --- |
| E1B3BDED-0E58-425A-ABC3-7B646322F6BA | CB5DDBAC-8715-44B2-92D0-1C26351110FE |

lib/widgets/edit_state_marker.dart
```diff
@@ -0,0 +1,127 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context).extension<DesignVariables>()!;
+
+    // TODO(#157): fix how star marker aligns with message content
+    // Design from Figma at:
+    //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=813%3A28817&mode=dev .
+    var star = Padding(padding: const EdgeInsets.only(top: 5.5),
+      child: Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor));
+    final hasMarker = widget.message.editState != MessageEditState.none;
```
chrisbobbe: Hmm, this file is called edit_state_marker.dart, so I'm wondering if the message-star UI code quite belongs in it. Maybe we want a different name for the file? @gnprice might have thoughts on the organization here.

test/widgets/message_list_test.dart
```diff
@@ -576,6 +576,39 @@ void main() {
...
+      final message2 = eg.streamMessage();
+      await setupMessageListPage(tester, messages: [message, message2]);
+      checkMarkersCount(edited: 0, moved: 0);
+
+      await store.handleEvent(eg.updateMessageEditEvent(message, renderedContent: "edited"));
+      await tester.pump();
+      checkMarkersCount(edited: 1, moved: 0);
+
+      await store.handleEvent(eg.updateMessageMoveEvent
+        ([message, message2], origTopic: 'old', newTopic: 'new'));
```
chrisbobbe, line 602: I don't think I've seen a method call split between its name and its opening `(`; I don't think we want that. :)

chrisbobbe (commented):
Thanks, this will be a useful feature! Comments below, including a curiosity about improving the star's (and the edit/move marker's) alignment with the message content.

lib/widgets/theme.dart
```diff
@@ -81,7 +81,11 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     borderBar = const Color(0x33000000),
     icon = const Color(0xff666699),
     title = const Color(0xff1a1a1a),
-    streamColorSwatches = StreamColorSwatches.light;
+    streamColorSwatches = StreamColorSwatches.light,
+    starColor = const HSLColor.fromAHSL(0.5, 47, 1, 0.41).toColor(),
+    bgMarker = const Color(0xffddecf6),
+    textMarker = const Color(0xff26516e),
+    textMarkerLight = const Color(0xff92a7b6);
```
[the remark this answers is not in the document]
PIG208: Not sure if edit permission is required for me to see the local variables. Sent an edit request.

lib/widgets/edit_state_marker.dart
PIG208: Yeah "edit_state_marker" is not a great fit. We could name it as with related to `SwipableMessageRow`, which is the only widget supposed to be exposed.

lib/widgets/edit_state_marker.dart
PIG208: https://github.com/zulip/zulip-flutter/assets/39874143/e5a74684-e444-4f80-bb70-6e48f44a3368

Wrapping the child with Transform.translate seems to work. We just need to deal with it overflowing the Row.

lib/widgets/theme.dart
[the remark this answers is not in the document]
PIG208: Pushed a revision without the theme changes for star and other colors used for the marker.

PIG208:
Thanks for the review! I pushed some updates addressing some reviews, but there are still some open discussions on matters like how we organize the theme variables and the edit marker code.

lib/widgets/theme.dart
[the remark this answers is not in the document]
gnprice: I think it's fine to use the same class to house variables that aren't named in that list in Figma.

Let's just keep it organized in a way that makes that clear. So in particular have such variables' definitions in their own stanza, with a comment. Like this:
```dart
  // Not exactly from the Figma design, but from Vlad anyway.
  final StreamColorSwatches streamColorSwatches;

  final Color starColor;
  final Color bgMarker;
  final Color textMarker;
  final Color textMarkerLight;
```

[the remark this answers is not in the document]
gnprice: > Not sure if edit permission is required for me to see the local variables. Sent an edit request.

I think I just approved that request. If you're still having trouble or have any questions, let's discuss in `#mobile-team` — I can try poking at other settings as an admin, and Chris might have other tips or Vlad might chime in with advice.

lib/widgets/edit_state_marker.dart
gnprice: How about having the caller stay responsible for the star? Then the child it passes here can include the star as well as the message content and reactions.

lib/widgets/theme.dart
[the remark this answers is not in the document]
PIG208: Reverted the theme-less change. With the permission the local variables approach now works for me. Thanks!

PIG208:
I have refactored the star out of `SwipableMessageRow` and re-implemented the thing with the `Row`-based approach drafted by @chrisbobbe. It works fine without animation.

While adding animation, it turned out that the overflow issue was pretty tricky, but solvable by wrapping the entire widget with an `OverflowBox` within the `LayoutBuilder`, and constraining the size of the message content from the parent (see #765).

gnprice:
> While adding animation, it turned out that the overflow issue was pretty tricky, but solvable by wrapping the entire widget with an `OverflowBox` within the `LayoutBuilder`, and constraining the size of the message content from the parent (see #765).

Interesting. I hope we can avoid needing a LayoutBuilder; but we'll cross that bridge when we get to it.

For this PR I'm happy if we merge something that's clean and has the right static behavior, even if it seems like it makes a LayoutBuilder or other tricks necessary for #765. After this is done and we focus on #765, maybe we'll find a way to avoid LayoutBuilder within the same architecture we merged in this PR; or maybe we'll do so but have to rework some of that architecture; or maybe we'll decide we need the LayoutBuilder after all.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,115 @@
...
+    return ConstrainedBox(
+      constraints: const BoxConstraints(maxWidth: widthCollapsed),
+      child: Container(
+        clipBehavior: Clip.hardEdge,
+        decoration: BoxDecoration(
+          borderRadius: BorderRadius.circular(2),
+          color: theme.bgMarker.withAlpha(0)),
+        child: Padding(
+          padding: const EdgeInsets.all(1.0),
+          child: marker),
```
chrisbobbe, line 111: I don't see vertical padding in the Figma. Am I missing it?

The height of the rounded-rectangle gets set by the line of text (e.g. "Edited"), it looks like. That's 22, because the line-height is set to that, and I don't see anything creating more vertical whitespace, in or out of the rounded-rectangle.

I do see some horizontal padding, but I wonder if this is the right way to express it. Here's a screenshot of the Figma:

image

In collapsed state, I see a 3px horizontal gap between the icon and the message content.

In uncollapsed state, I see:

a) 4px between the rounded-rectangle and the message content
b) 2px between the icon and the right edge of the rounded-rectangle
c) 5px between "Edited" and the icon
d) 3px between the left edge of the rounded-rectangle and "Edited"

To my eye, (a) should maybe just become 3px, to match the 3px in the collapsed state. (Maybe helps keep the implementation simple too?) And it seems to me like (b) and (d) should equal each other, to be symmetrical; not sure which value is better. Then since the uncollapse action adds space between the icon and the message content, perhaps (b) should animate from 0 to 2 as part of the animation (and then we animate (d), too, to be symmetrical).

Also, let's do (c) explicitly with a 5px gap, instead of a space character as in this revision. That might be something to animate from zero too, I'm not sure.

chrisbobbe (commented):
Thanks! Comments below.

lib/widgets/message_list.dart
```diff
@@ -962,25 +963,22 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-                ? Padding(padding: const EdgeInsets.only(top: 4),
-                    child: Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor))
-                : null),
-          ]),
+              SizedBox(width: 16,
+                child: message.flags.contains(MessageFlag.starred)
+                  ? Icon(ZulipIcons.star_filled, size: 16, color: theme.starColor)
+                  : null),
+            ],
+          ),
```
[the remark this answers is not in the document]
gnprice: Yeah, separating those would definitely be helpful for review. They seem like they should be naturally pretty independent changes, too.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,115 @@
...
+            overflow: TextOverflow.clip,
+            softWrap: false,
+            textAlign: TextAlign.center,
+            style: TextStyle(fontSize: 15, color: theme.textMarker.withAlpha(0)))),
+        // To match the Figma design, we cannot make the collapsed width of the
+        // marker larger. We need to explicitly allow the icon to overflow.
+        OverflowBox(
+          fit: OverflowBoxFit.deferToChild,
+          maxWidth: 10,
+          child: Icon(icon, size: iconSize, color: theme.textMarkerLight),
```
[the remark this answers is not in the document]
chrisbobbe: (Also, I don't think we have a source for the edited-moved marker colors in dark theme yet. So let's put a `// TODO(#95) need dark-theme color` on them.)

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,115 @@
...
+
+    var marker = Row(
+      mainAxisAlignment: MainAxisAlignment.end,
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        Flexible(
+          fit: FlexFit.loose,
+          // For now, [markerText] is not displayed because it is transparent and
+          // there is not enough space in the parent ConstrainedBox.
+          child: Text('$markerText ',
```
gnprice, line 87: Let's leave this out of this static version, since it's not yet being used — that will make the logic easier to think about. Then we can introduce it (modulo other edits, like Chris's comment about the space character vs. an explicit number of pixels of spacing) in the followup PR #765 that adds the swipe-to-reveal feature.

lib/widgets/swipable_message_row.dart
PIG208: Thanks for the breakdown on the design. Really helpful! I'm moving the relevant updates for this to #765.

lib/widgets/swipable_message_row.dart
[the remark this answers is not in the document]
PIG208: Yeah, I think with 50% it matches the design better.
<details>

<summary>Color comparsion</summary>

| Before | After (50% opacity) |
| --- | --- |
| 93b790eb-912a-47d7-a2c1-1d2970356c5e | 6e6c3d4d-9395-4201-b967-191ffa736d4c

</details>

PIG208:
Thanks for the reviews! This has been updated and is ready for another round now.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,86 @@
...
+      case MessageEditState.moved:
+        icon = ZulipIcons.message_moved;
+        iconSize = 8;
+        break;
+    }
+
+    return ConstrainedBox(
+      constraints: const BoxConstraints(maxWidth: widthCollapsed),
+      child: Padding(
+        padding: const EdgeInsets.only(left: 5, right: 3),
```
chrisbobbe, line 77: Since left and right are unequal, this is a place to use `EdgeInsetsDirectional` instead of `EdgeInsets`.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,86 @@
...
+      ],
+    );
+  }
+}
+
+class _EditStateMarker extends StatelessWidget {
+  /// The minimum width of the marker.
+  //
+  // Currently, only the collapsed state of the marker has been implemented,
+  // where only the marker icon, not the marker text, is visible.
```
chrisbobbe, line 45: nit: empty line unnecessary

```suggestion
  /// The minimum width of the marker.
  // Currently, only the collapsed state of the marker has been implemented,
  // where only the marker icon, not the marker text, is visible.
```

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,86 @@
...
+      constraints: const BoxConstraints(maxWidth: widthCollapsed),
+      child: Padding(
+        padding: const EdgeInsets.only(left: 5, right: 3),
+        child: OverflowBox(
+          fit: OverflowBoxFit.deferToChild,
+          maxWidth: 8,
+          child: Icon(icon, size: iconSize,
+            color: designVariables.editedMovedMarkerCollapsed)),
+        ),
+    );
```
chrisbobbe, line 84: nit:

```suggestion
        child: OverflowBox(
          fit: OverflowBoxFit.deferToChild,
          maxWidth: 8,
          child: Icon(icon, size: iconSize,
            color: designVariables.editedMovedMarkerCollapsed))));
```

chrisbobbe (commented):
Thanks! LGTM except a few small comments below, and I'll mark this as ready for Greg's review.

chrisbobbe:
I see you've pushed changes; is this ready for Greg's review? I took a look at what you pushed, and it resolves my previous review. Thanks!

PIG208:
Yes, this is ready for review.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,83 @@
...
+    switch (_editState) {
+      case MessageEditState.none:
+        return const SizedBox(width: widthCollapsed);
+      case MessageEditState.edited:
+        icon = ZulipIcons.edited;
+        iconSize = 14;
+        break;
+      case MessageEditState.moved:
+        icon = ZulipIcons.message_moved;
+        iconSize = 8;
```
gnprice, line 69: It looks like the underlying icon SVG has a height of 8 but a width of 9. Will this cause it to be scaled down to fit into an 8x8 box?

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,83 @@
...
+        break;
+    }
+
+    return ConstrainedBox(
+      constraints: const BoxConstraints(maxWidth: widthCollapsed),
+      child: Padding(
+        padding: const EdgeInsetsDirectional.only(start: 5, end: 3),
+        child: OverflowBox(
+          fit: OverflowBoxFit.deferToChild,
+          maxWidth: 8,
```
gnprice, line 79: Can we do this without OverflowBox?

There's no real overflow happening in the end, right? The wider icon, for moved, is only width 9 — so it should fit just fine in the 16px width that this Padding widget has. Can we adjust the padding to match that instead?

lib/widgets/message_list.dart
```diff
@@ -976,13 +975,9 @@ class MessageWithPossibleSender extends StatelessWidget {
                 ])),
             SizedBox(width: 16,
```
gnprice, line 976: nit: should follow indentation of other child

lib/widgets/message_list.dart
```diff
@@ -964,10 +963,10 @@ class MessageWithPossibleSender extends StatelessWidget {
           if (senderRow != null)
             Padding(padding: const EdgeInsets.fromLTRB(16, 2, 16, 0),
               child: senderRow),
-          Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
```
gnprice, line 967: msglist: Fix star icon alignment issue.
    
    By aligning the row using `textBaseline`, we no longer need to apply
    the padding to the star.

Neat!

gnprice (commented):
Thanks @PIG208, and thanks @chrisbobbe for the previous reviews! Looking forward to having this feature.

I'm only partway through the last commit and it's time for me to go home, so I'll leave the rest of that commit for a later review. Comments below for so far.

lib/widgets/swipable_message_row.dart
PIG208: Yeah, that should be the case.

| | moved (8x8) | edited (8x14) |
| - | - | - |
| Flutter | Screenshot from 2024-07-12 21-02-36 | Screenshot from 2024-07-12 21-01-17 |
| Figma | Screenshot from 2024-07-12 21-07-57 | Screenshot from 2024-07-12 21-08-10 |

`iconSize` scales both the width and height of the icon, until it exceeds the constraints of the parent.

<details>

image

</details>

lib/widgets/swipable_message_row.dart
PIG208: That's doable. I think `OverflowBo`x is something that we might need for the animated version.

gnprice: OK. Even if we end up needing it for animated, I'd like to try starting without it as a baseline - I think that'll make this easier to reason about.

lib/widgets/swipable_message_row.dart
gnprice: Hmm I see - so the icon file has a width of 9 but the design gives it a width of 8. Confusing. Worth clarifying in a comment, at the least.

Probably also good to ask Vlad about it, in a chat thread in... #mobile, I guess, since it's design (even though it's a fine detail of design).

lib/widgets/message_list.dart
```diff
@@ -964,25 +963,20 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-                // Design from Figma at:
-                //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=813%3A28817&mode=dev .
-                ? Padding(padding: const EdgeInsets.only(top: 4),
-                    child: Icon(ZulipIcons.star_filled, size: 16, color: _starColor))
-                : null),
-          ]),
+              SizedBox(width: 16,
+                child: message.flags.contains(MessageFlag.starred)
+                  ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
+                  : null)]),
```
gnprice, line 979: nit:
```suggestion
                  : null),
            ]),
```

(The pattern is that when there's a list, so the items are peers or siblings of each other, the last item gets a trailing comma just like the other items and the closing delimiter like `]` goes on the next line.

When there's one item that's the main payload and the other items are configuration around that — so in particular, for the `child` parameter vs. the other parameters of some widget constructor — then the main payload goes at the end and piles up the closing `)` delimiters without spilling onto more lines.)

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+
+class SwipableMessageRow extends StatefulWidget {
+  const SwipableMessageRow({
+    super.key,
+    required this.children,
+    required this.message,
+  });
+
+  final List<Widget> children;
+  final Message message;
```
gnprice, line 17: nit: "child" or "children" fields on widgets come after other fields (here and in constructor parameters… and in constructor calls, as mentioned on another comment just now)

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+
+  @override
+  State<StatefulWidget> createState() => _SwipableMessageRowState();
+}
+
+class _SwipableMessageRowState extends State<SwipableMessageRow> {
+  @override
+  Widget build(BuildContext context) {
```
gnprice, line 25: Seems like this can just as well be a stateless widget:
```suggestion
  @override
  Widget build(BuildContext context) {
```

(We'll need state here for animation, in #765, but it's easy to add then. And the state might go on a child, rather than this widget itself — one way or another I'd like to arrange it so we don't need a state in the case of MessageEditState.none, at least once the state starts holding other resources like an AnimationController.)

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+      ],
+    );
+  }
+}
+
+class _EditStateMarker extends StatelessWidget {
+  /// The minimum width of the marker.
+  // Currently, only the collapsed state of the marker has been implemented,
+  // where only the marker icon, not the marker text, is visible.
+  static const double widthCollapsed = 16;
```
gnprice, line 45: nit: this goes after the constructor and field:
https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+
+  const _EditStateMarker({
+    required MessageEditState editState,
+  }) : _editState = editState;
```
gnprice, line 49: nit: can simplify this by making the field public

The class is private anyway, so the visibility of the field doesn't matter.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+  final MessageEditState _editState;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    final IconData icon;
+    final double iconSize;
+
+    switch (_editState) {
```
gnprice, line 60: nit: these two locals are tightly related to this `switch` (it's what sets them), so keep them adjacent in the same stanza:
```suggestion
    final IconData icon;
    final double iconSize;
    switch (_editState) {
```

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+      case MessageEditState.none:
+        return const SizedBox(width: widthCollapsed);
+      case MessageEditState.edited:
+        icon = ZulipIcons.edited;
+        iconSize = 14;
+        break;
+      case MessageEditState.moved:
```
gnprice, line 67: nit:
```suggestion
      case MessageEditState.moved:
```

Dart has removed the need for `break` as a routine part of a switch statement, which is a nice language improvement. See [spec discussion](https://github.com/dart-lang/language/blob/2b3d78182c226aa26d46df341fc2cdb48a88cbaf/accepted/3.0/patterns/feature-specification.md#implicit-break).

It looks like there is a lint rule for this. Likely we should enable it:
https://dart.dev/tools/linter-rules/unnecessary_breaks

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+  Widget build(BuildContext context) {
+    final hasMarker = widget.message.editState != MessageEditState.none;
+
+    return Row(
+      crossAxisAlignment: CrossAxisAlignment.baseline,
+      textBaseline: localizedTextBaseline(context),
+      children: [
+        hasMarker
+          ? _EditStateMarker(editState: widget.message.editState)
+          : const SizedBox(width: 16),
```
gnprice, line 34:
```suggestion
          : const SizedBox(width: _EditStateMarker.widthCollapsed),
```

It's important that this 16 is the same as the other 16, right? So let's make that explicit by referring to it by a common name.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    final IconData icon;
+    final double iconSize;
+
+    switch (_editState) {
+      case MessageEditState.none:
+        return const SizedBox(width: widthCollapsed);
```
gnprice, line 62: This case isn't intended to be possible, right? Let's make that explicit with an assert:
```suggestion
      case MessageEditState.none:
        assert(false);
        return const SizedBox(width: widthCollapsed);
```

lib/widgets/theme.dart
```diff
@@ -111,6 +123,12 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
   // Not exactly from the Figma design, but from Vlad anyway.
   final StreamColorSwatches streamColorSwatches;
 
+  // Not named variables in Figma; taken from older Figma drafts, or elsewhere.
+  final Color star;
+  final Color editedMovedMarkerBg;
+  final Color editedMovedMarkerExpanded;
```
gnprice, line 129: These two aren't used yet in this commit (or PR), right? Let's save them for the commit where they do get used:
```suggestion
```

test/widgets/message_list_test.dart
```diff
@@ -576,6 +576,39 @@ void main() {
...
+      check(find.byIcon(ZulipIcons.message_moved).evaluate()).length.equals(moved);
+    }
+
+    testWidgets('no edited or moved messages', (tester) async {
+      final message = eg.streamMessage();
+      await setupMessageListPage(tester, messages: [message]);
+      checkMarkersCount(edited: 0, moved: 0);
+    });
+
+    testWidgets('edited and moved messages from events', (tester) async {
```
gnprice, line 591: Let's have another test case that checks that the markers don't affect the layout of anything else. That seems like a class of bug that could easily be introduced in some future change to the message UI.

So that can look like:
* put three messages on the screen; the one we'll change the state of, and one each before and after it
* note the Rect position of everything relevant:
  * each of the three messages
  * the target message's star (make it starred so this appears)
  * some text in the target message
  * maybe other things I'm not thinking of
  * … oh a reaction, that's a good one (and there might still be more I'm not thinking of)
* then make the message be moved; and look at all the same locations and confirm none have changed
* then make the message be edited; and ditto

gnprice (commented):
OK, and here's the rest of a review (based on your latest revision, so including a couple of comments new with this revision).

Most are small. Among the more interesting: the first comment is one where I'm uncertain whether this gives the right layout. The last comment asks for a layout-related test.

lib/widgets/swipable_message_row.dart
PIG208: I checked again. So it looks like a later edition has an icon of size 8x8 when downloaded. But I'm starting to think if we should use the frame (8x14) containing that icon, not just the icon itself (8x8) for the design. That way the size also matches the "edited" marker icon.

lib/widgets/swipable_message_row.dart
```diff
@@ -0,0 +1,80 @@
...
+        icon = ZulipIcons.message_moved;
+        iconSize = 8;
+        break;
+    }
+
+    return ConstrainedBox(
+      constraints: const BoxConstraints(maxWidth: widthCollapsed),
+      child: Padding(
+        padding: const EdgeInsetsDirectional.only(start: 5, end: 3),
+        child: Icon(icon, size: iconSize,
```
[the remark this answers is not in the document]
PIG208: I'm switching to use the 8x14 Figma frame that contains the move icon, with `EdgeInsetsDirectional.only(start: 2)` for the padding. The look appears to be the same when we were using `EdgeInsetsDirectional.only(start: 5, end: 3)`, but that clips the icon from 14 in width to 8 in width.

test/widgets/message_list_test.dart
PIG208: Added the "edit state updates do not affect layout" test case. It uses a heavy helper function to record the items you mentioned, as well as `ImageNode`s and `ParagraphNode`s from the target message.

PIG208:
Thanks for the review @gnprice ! The PR has been updated.

gnprice (commented):
Thanks @PIG208 for the revision! Comments below.

Would you also post an updated screenshot of how this looks with each of the two kinds of markers? I'm not sure if the images above are the same as in the current revision.

In particular when I look at an edited message, the bottom of the icon is aligned with the baseline of the (first line of the) message text… but then that leaves the top of the icon sticking up well above the ascender line of the text. The result is that it feels like the icon is sticking up above the message, so that it's not visually clear if it's meant to be attached to the message or what.

test/widgets/message_list_test.dart
```diff
@@ -576,6 +577,107 @@ void main() {
     });
   });
 
+  group('EditStateMarker', () {
+    void checkMarkersCount({required int edited, required int moved}) {
+      check(find.byIcon(ZulipIcons.edited).evaluate()).length.equals(edited);
+      check(find.byIcon(ZulipIcons.message_moved).evaluate()).length.equals(moved);
+    }
+
+    testWidgets('no edited or moved messages', (WidgetTester tester) async {
```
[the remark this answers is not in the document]
PIG208: I see. I added those in the previous revision because I saw we doing that in the surrounding test cases.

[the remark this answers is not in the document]
gnprice: Yeah, matching the surrounding code is a good instinct.

Like I said, having the explicit type here is pretty harmless. But maybe I'll do a quick sweep to remove them for `testWidgets` calls throughout our tests, just so it's uniformly a bit simpler. (From a quick scan, about 1/3 of our `testWidgets` calls have them.)

test/widgets/message_list_test.dart
```diff
@@ -576,6 +577,107 @@ void main() {
...
+      }
+      return result;
+    }
+
+    testWidgets('edit state updates do not affect layout', (WidgetTester tester) async {
+      prepareBoringImageHttpClient();
+      final messages = [
+        eg.streamMessage(),
+        eg.streamMessage(
+          content: ContentExample.imageClusterThenContent.html,
```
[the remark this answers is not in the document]
PIG208: I wanted to test that because I discovered the part with `MessageImageList` do not respond to swipe on the animated branch. It is not relevant now but might be later.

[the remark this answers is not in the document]
PIG208: This is something I should note in the commit message, apparently.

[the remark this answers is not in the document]
gnprice: Interesting, that makes sense.

I think that still won't be relevant for this layout test, though — having the image interfere with a gesture (because the image has its own gesture recognizers) is a lot more natural than having it interfere with an ancestor's layout. So it's still probably best left out of this test, but then you might use this example in a new test in the swipe-to-reveal PR.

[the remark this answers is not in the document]
gnprice: (Or even in that later test, it sounds like `imageSingle` would likely be enough to make the point — it has an image and even a paragraph too, just one of each instead of two of each.)

test/widgets/message_list_test.dart
```diff
@@ -576,6 +577,107 @@ void main() {
...
+            of: baseFinder, matching: matching)..tryEvaluate();
+          check(finder.found, because: '${message.content}: $matching')
+            .isNotEmpty();
+          return List.generate(
+            finder.found.length,
+            (i) => tester.getRect(finder.at(i)));
+        }
+
+        result.add({
+          'MessageWithPossibleSender': [tester.getRect(baseFinder)],
```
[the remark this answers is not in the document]
gnprice: Oh, and I forgot to post what the error output looks like after that change. Here it is:
```
$ flutter test test/widgets/message_list_test.dart --name EditState
00:03 +2: EditStateMarker edit state updates do not affect layout                                        
Expected: a List<List<(String, Rect)>> that:
  is deeply equal to [[(MessageWithPossibleSender, Rect.fromLTRB(20.0, 187.0, 780.0, 259.0)),
  (MessageContent, Rect.fromLTRB(36.0, 225.0, 764.0, 255.0))],
  [(MessageWithPossibleSender, Rect.fromLTRB(20.0, 259.0, 780.0, 457.0)),
  (MessageContent, Rect.fromLTRB(36.0, 263.0, 764.0, 430.0)),
  (Star, Rect.fromLTRB(764.0, 270.3, 780.0, 286.3)),
  (ReactionChipsList, Rect.fromLTRB(36.0, 430.0, 764.0, 453.0))],
  [(MessageWithPossibleSender, Rect.fromLTRB(20.0, 457.0, 780.0, 495.0)),
  (MessageContent, Rect.fromLTRB(36.0, 461.0, 764.0, 491.0))]]
Actual: [[(MessageWithPossibleSender, Rect.fromLTRB(20.0, 187.0, 780.0, 259.0)),
(MessageContent, Rect.fromLTRB(36.0, 225.0, 764.0, 255.0))],
[(MessageWithPossibleSender, Rect.fromLTRB(20.0, 259.0, 780.0, 457.0)),
(MessageContent, Rect.fromLTRB(34.0, 263.0, 764.0, 430.0)),
(Star, Rect.fromLTRB(764.0, 270.3, 780.0, 286.3)),
(ReactionChipsList, Rect.fromLTRB(34.0, 430.0, 764.0, 453.0))],
[(MessageWithPossibleSender, Rect.fromLTRB(20.0, 457.0, 780.0, 495.0)),
(MessageContent, Rect.fromLTRB(36.0, 461.0, 764.0, 491.0))]]
Which: at [<1>][<1>] is <(MessageContent, Rect.fromLTRB(34.0, 263.0, 764.0, 430.0))>
which does not equal <(MessageContent, Rect.fromLTRB(36.0, 263.0, 764.0, 430.0))>
```

Much clearer! The developer can immediately see that the discrepancy is that the left edge of MessageContent is 2px too far to the left. (Which indeed it is, with the artificial error I induced as described above which makes the marker icon 2px too small.)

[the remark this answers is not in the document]
PIG208: Agreed with the change. When I was testing Flutter uglyprinted the maps. This made me I wonder if they can be displayed better. I remember vaguely that pytest highlights the differences between two dicts in a diff-like fashion, making it easier to spot the differences.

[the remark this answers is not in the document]
gnprice: Yeah, I think some other nice improvements could be:
* better pretty-print the expected and actual data — above you can see both maps and lists lack indentation to show the structure, in particular
* fancier diff comparison at an algorithmic level — detecting things like "an element was inserted" in the middle of a list, for example
* printing actual vs. expected in diff-style formatting, with `-`/`+` lines and/or with red and green colors

The last one is a matter of preference which people might differ on. The other two are probably uncontroversial improvements… given a solid implementation of them. Both can be complex to get right — pretty-printing in particular can sound simple but turns out to be tricky when you try to really pin it down.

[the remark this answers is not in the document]
gnprice: (And all three of those are found in some popular other test frameworks — pytest is a good comparison, and the Jest framework for JS is another. Pretty sure each of the three is found in one or both of those.)

PIG208:
<details>
<summary>Latest screenshots</summary>

| Moved | Edited |
| - | - |
| image | Screenshot from 2024-07-16 14-09-08 |
| image | Screenshot from 2024-07-16 14-11-43 |
| Screenshot from 2024-07-16 14-15-49 | image |
| image | Screenshot from 2024-07-16 14-16-04 |


</details>

PIG208:
Both icons are 14x14. `message_moved` floats from the baseline while `edited` aligns its bottom to it. It differs from the design:

| Moved | Edited |
| - | - |
| Screenshot from 2024-07-16 14-24-52 | Screenshot from 2024-07-16 14-24-38 |

I'm going to apply an offset to shift them from the baseline to the desired position.

PIG208:
<details>
<summary>Latest screenshots (with the offset)</summary>

| Moved | Edited |
| - | - |
| Screenshot from 2024-07-16 14-31-57 | Screenshot from 2024-07-16 14-31-42 |
| Screenshot from 2024-07-16 14-32-20 | Screenshot from 2024-07-16 14-32-25 |
| Screenshot from 2024-07-16 14-34-04 | Screenshot from 2024-07-16 14-34-41 |
| Screenshot from 2024-07-16 14-35-32 | Screenshot from 2024-07-16 14-35-11 |




</details>

PIG208:
<details>
<summary>
Latest screenshots (with the 16x16 icons):
</summary>

| Moved | Edited |
| - | - |
| Screenshot from 2024-07-16 19-34-12 | Screenshot from 2024-07-16 19-34-07 |
| Screenshot from 2024-07-16 19-34-44 | Screenshot from 2024-07-16 19-34-30 |
| Screenshot from 2024-07-16 19-35-54 | Screenshot from 2024-07-16 19-35-46 |

</details>

PIG208:
With the new icons, not sure if the offsets are needed anymore.
This should be ready for review.

lib/widgets/message_list.dart
```diff
@@ -964,25 +963,21 @@ class MessageWithPossibleSender extends StatelessWidget {
           if (senderRow != null)
             Padding(padding: const EdgeInsets.fromLTRB(16, 2, 16, 0),
               child: senderRow),
-          Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
-            const SizedBox(width: 16),
-            Expanded(
-              child: Column(
+          EditStateMarker(
+            message: message,
+            children: [
```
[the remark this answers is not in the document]
gnprice: bump :slightly_smiling_face:

gnprice:
Hmm, here's a screenshot of what I see:
image

So visually the bottom of the icon is higher than the baseline of the text, and the top of the icon is much higher than the ascender line of the text.

That's on my usual display settings, which have the font size set to small (as well as the "display size", which effectively just means my screen has more logical pixels). On default settings, I see:
image

So the same thing, but a little less so, because the icons aren't as much taller than the text as they are at my usual settings.

gnprice:
---

On our screen-sharing call today, when you were running the app locally on Linux, it looked like you were seeing the symptom of #735. That suggests that your system has the text scaling set to >1x, i.e. the fonts are scaled up. If I crank the text size on my phone a couple of notches above default, I reproduce a layout more like your screenshots above.

When making screenshots for a piece of layout that involves text, it'd be good to do so with a 1x text scale. It can be helpful to try smaller or larger scaling too, but 1x should be the baseline.

I'm not sure what platform setting might be governing the text scaling Flutter picks up from the platform on Linux. If you can't readily find that setting, you can always add a hack locally in `lib/app.dart` that forces the scaling to 1x — for example a `MediaQuery` around the `ZulipApp`.

gnprice:
For the vertical alignment for these icons, there are probably fancy things we can do to get it just right for a range of situations — but let's skip that for the present, while we move on to other issues from our tracker.

For now, let's just try to get it so that at 1x text scaling, when the message starts with a plain paragraph, the top of the edit icon doesn't stick out above the ascender line of the text, or anyway not by much more than the bottom of the icon sticks out below the baseline of the text. If we do that by a fixed few px of vertical adjustment, that'll be fine.

gnprice:
Other than that layout issue, the code here all looks good modulo this small comment above: https://github.com/zulip/zulip-flutter/pull/762#discussion_r1680182894

Thanks for the revision!

lib/widgets/message_list.dart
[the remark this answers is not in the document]
PIG208: Oops. Fixed!

PIG208:
Updated the PR to add the vertical adjustments and replace the outdated icon font. Thanks for noticing that scaling issue!

<details>

<summary>With the offset, default text scaling (fd38fb86f418b8243df77f900a3e4df4b090660d)</summary>

| Moved | Edited |
| - | - |
| Screenshot from 2024-07-16 20-23-35 | Screenshot from 2024-07-16 20-23-27 |
| Screenshot from 2024-07-16 20-23-42 | Screenshot from 2024-07-16 20-23-47 |
| Screenshot from 2024-07-16 20-24-05 | Screenshot from 2024-07-16 20-24-00 |

</details>

gnprice:
Great, thanks! Looks good — merging.

I added a comment about those vertical offsets:
```diff
+        // These offsets are chosen ad hoc, but give a good vertical alignment
+        // of the icons with the first line of the message, when the message
+        // begins with a paragraph, at default text scaling.  See:
+        //   https://github.com/zulip/zulip-flutter/pull/762#issuecomment-2232041922
         offset = const Offset(0, 2);
```

Also edited the PR description so that #171 doesn't get accidentally closed.

Issue #636, opened 2024-04-26
ui: Distinguish `isBot: true` users in more places, not just message senders in the message list


chrisbobbe:
Quoting myself in https://github.com/zulip/zulip-flutter/pull/605#issuecomment-2045964365 , which added bot markers on message senders in the message list:

> After this, there are still some places in the UI where we show the name of a bot user without making it clear it's a bot. For example, it would probably be helpful to show the bot icon next to a bot user's name on the profile page (`ProfilePage`). I'm not sure if we need to show the icon in _all_ the places where names appear, though. [Let's make] a checklist of all the places where bots' names might be displayed in the app, indicating which ones should start marking bots with the bot icon, which ones don't need to do that, and which ones [we're] not sure about[.]

chrisbobbe:
@sm-sayedi, are you still up for drafting a checklist like this? 🙂

sm-sayedi:
> @sm-sayedi, are you still up for drafting a checklist like this? 🙂

Unfortunately, I haven't completed the list yet as I got busy with #608! After I finish the first round of tests for that, I will return to this one and finalize the list. Should I then post it as a comment under this issue?

chrisbobbe:
Sure, sounds good! Thank you!

PR #764, opened 2024-06-26
ui: Distinguish bot users in more places

sm-sayedi:
Now that #605 is merged already, bot icons are shown with the name of bot users in `MessageListPage`. Additional places where users appear are covered in this PR:

- `ProfilePage`
- `RecentDmConversationsPage`

### ProfilePage
<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
        <td><b>After (with long name)</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
        <td></td>
    </tr>
</table>

### RecentDmConversationsPage
<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
        <td><b>After (with long name)</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
        <td></td>
    </tr>
</table>

~**Note:** In `RecentDmConversationsPage` I wasn't sure whether to add (and where to add) the bot icon to a group dm where there may be one or multiple bot recipients.~

Fixes: #636

### DIFF
--- a/lib/widgets/profile.dart
+++ b/lib/widgets/profile.dart
@@ -7,6 +7,7 @@ import '../api/model/model.dart';
 import '../model/content.dart';
 import '../model/narrow.dart';
 import 'content.dart';
+import 'icons.dart';
 import 'message_list.dart';
 import 'page.dart';
 import 'store.dart';
@@ -46,10 +47,23 @@ class ProfilePage extends StatelessWidget {
       Center(
         child: Avatar(userId: userId, size: 200, borderRadius: 200 / 8)),
       const SizedBox(height: 16),
-      Text(user.fullName,
-        textAlign: TextAlign.center,
-        style: _TextStyles.primaryFieldText
-          .merge(weightVariableTextStyle(context, wght: 700))),
+      Row(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: [
+          if (user.isBot) ...[
+            const Icon(
+              ZulipIcons.bot,
+              size: 22,
+              color: Color.fromARGB(255, 159, 173, 173),
+            ),
+            const SizedBox(width: 10),
+          ],
+          Flexible(child: Text(user.fullName,
+            textAlign: TextAlign.center,
+            style: _TextStyles.primaryFieldText
+              .merge(weightVariableTextStyle(context, wght: 700)))),
+        ],
+      ),
       // TODO(#291) render email field
       Text(roleToLabel(user.role, zulipLocalizations),
         textAlign: TextAlign.center,
--- a/lib/widgets/recent_dm_conversations.dart
+++ b/lib/widgets/recent_dm_conversations.dart
@@ -91,24 +91,29 @@ class RecentDmConversationsItem extends StatelessWidget {
     final store = PerAccountStoreWidget.of(context);
     final selfUser = store.users[store.selfUserId]!;
 
-    final String title;
+    final List<({String name, bool isBot})> users;
     final Widget avatar;
     switch (narrow.otherRecipientIds) { // TODO dedupe with DM items in [InboxPage]
       case []:
-        title = selfUser.fullName;
+        users = [(name: selfUser.fullName, isBot: selfUser.isBot)];
         avatar = AvatarImage(userId: selfUser.userId, size: _avatarSize);
       case [var otherUserId]:
         // TODO(#296) actually don't show this row if the user is muted?
         //   (should we offer a "spam folder" style summary screen of recent
         //   1:1 DM conversations from muted users?)
         final otherUser = store.users[otherUserId];
-        title = otherUser?.fullName ?? '(unknown user)';
+        users = [(name: otherUser?.fullName ?? '(unknown user)',
+          isBot: otherUser?.isBot ?? false)];
         avatar = AvatarImage(userId: otherUserId, size: _avatarSize);
       default:
         // TODO(i18n): List formatting, like you can do in JavaScript:
         //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya'])
         //   // 'Chris、Greg、Alya'
-        title = narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', ');
+        users = narrow.otherRecipientIds.map((id) {
+          final user = store.users[id];
+          return (name: user?.fullName ?? '(unknown user)',
+            isBot: user?.isBot ?? false);
+        }).toList();
         // TODO(#95) need dark-theme color
         avatar = ColoredBox(color: const Color(0x33808080),
           child: Center(
@@ -131,7 +136,7 @@ class RecentDmConversationsItem extends StatelessWidget {
             const SizedBox(width: 8),
             Expanded(child: Padding(
               padding: const EdgeInsets.symmetric(vertical: 4),
-              child: Text(
+              child: Text.rich(
                 style: const TextStyle(
                   fontSize: 17,
                   height: (20 / 17),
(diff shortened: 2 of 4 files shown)


rajveermalviya (approved):
Thanks! Changes LGTM and tests well :)
Moving on to mentor review, cc @hackerkid.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -114,6 +117,7 @@ class RecentDmConversationsItem extends StatelessWidget {
           child: Center(
             // TODO(#95) need dark-theme color
             child: Icon(ZulipIcons.group_dm, color: Colors.black.withOpacity(0.5))));
+        isBot = false;
```
hackerkid, line 120: Why is this always false? Isn't it possible to create DMs with multiple bots?

sm-sayedi: > In `RecentDmConversationsPage`, I wasn't sure whether to add (and where to add) the bot icon to a group dm where there may be one or multiple bot recipients.

As I mentioned in the PR description above, I wasn't quite sure about this. There are two cases in group DMs with bot users: 

- A group DM with both real and bot users; in which I think it is not reasonable to add a bot icon at the end of the label as we do it in `1:1` DM with bot user. One approach we can take is to add a bot icon after the name of each bot user, something like this: **Welcome Bot 🤖, Mahmood, S Muhmood**. Or we can ignore the bot icon at all. I am not sure.
- A group DM with just bot users; in which I think we can add a bot icon at the end of the label, but again, it will be confused with the first case, is it that the group DM is with multiple real users and one last bot user or the group DM is with all bot users?! If we ignore the bot icon in the first case, then this cannot be confusing!

Let's also hear from @gnprice about this!

gnprice: This is a question about how the UI should behave for users, so it's a good one to get feedback on from @alya .

alya: The bot icon should be associated with a particular user, not with a conversation as a whole. So I would expect it to appear after each bot's name, regardless of how many bot and non-bot users there are in the conversation.

alya: By the way, https://github.com/zulip/zulip/issues/26831 is the corresponding web app issue.

gnprice:
> The bot icon should be associated with a particular user, not with a conversation as a whole. So I would expect it to appear after each bot's name

@sm-sayedi For how to implement this behavior, the key Flutter feature I think you'll want is [WidgetSpan](https://main-api.flutter.dev/flutter/widgets/WidgetSpan-class.html). For example in message content that's how we handle custom emoji, user mentions, and global times.

lib/widgets/recent_dm_conversations.dart
hackerkid: That makes a lot of sense. Thanks for the clarification @alya!

sm-sayedi:
@hackerkid, @alya Revision pushed with images also updated in the PR description. PTAL.

hackerkid (approved):

alya:
We would ideally show the icon even if the name gets abbreviated.

gnprice:
Advancing to maintainer review, since mentor review is complete.

In general reviewers should advance the PR to the next stage when they're done reviewing (or done except a few small comments). It's also a good idea for PR authors to check in to see if that hasn't happened, and advance it themselves.

For cross-reference, there's a chat thread started from Alya's comment above:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/bot.20icon.20in.20recent-dms.20items.20-.20.23F764/near/1846941

That doesn't affect the first commit in this PR, though. Not sure if it'll affect the later commits enough to hold off reviewing them until it's been revised for that.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -140,7 +145,19 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+                    if (user.isBot) ...[
+                      const WidgetSpan(child: SizedBox(width: 5)),
+                      const WidgetSpan(
+                        child: Icon(ZulipIcons.bot, size: 15,
+                          color: Color.fromARGB(255, 159, 173, 173)),
+                        alignment: PlaceholderAlignment.middle),
+                    ],
+                    if (index < users.length - 1) const TextSpan(text: ', '),
+                  ]);
+                }).toList())))),
```
chrisbobbe, line 160: I see a comment on the `switch`:

```dart
 // TODO dedupe with DM items in [InboxPage]
```

and that reminds me that we also show users' names in the Inbox page, when there are unread DMs. Would you add a prep commit that makes a helper function that can be used in this page and the Inbox page too? Perhaps it can return a `TextSpan`, and it can live in either file or perhaps a new file.

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -206,39 +217,95 @@ void main() {
...
 
-          checkAvatar(tester, DmNarrow.ofMessage(message, selfUserId: eg.selfUser.userId));
-          checkTitle(tester, '(unknown user)');
+          testWidgets('bot recipient -> shows no bot icon', (tester) async {
+            await checkRecipient(tester, isBot: true, looksBot: false);
+          });
+
+          testWidgets('non-bot recipient -> shows no bot icon', (tester) async {
+            await checkRecipient(tester, isBot: false, looksBot: false);
+          });
```
chrisbobbe, line 266:
```dart
        group('no error when user somehow missing from store.users', () {
          // […]
          testWidgets('bot recipient -> shows no bot icon', (tester) async {
            await checkRecipient(tester, isBot: true, looksBot: false);
          });

          testWidgets('non-bot recipient -> shows no bot icon', (tester) async {
            await checkRecipient(tester, isBot: false, looksBot: false);
          });
```

The "bot recipient" test doesn't actually simulate a bot recipient, right, because these tests are checking what happens when the user data is missing. I see that the `User` passed to `eg.dmMessage` has `isBot: true`, but the `isBot` doesn't actually get written into the `DmMessage` object, so the app code can't treat this case differently than the "non-bot recipient" case.

Simplest I think to leave this test is it was, but with an added `checkBotIcon(count: 0);`.

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -206,39 +217,95 @@ void main() {
...
+            final user = eg.user(userId: 1, isBot: isBot, fullName: 'Long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name long name');
+            final message = eg.dmMessage(from: eg.selfUser, to: [user]);
+            await setupPage(tester, users: [user], dmMessages: [message]);
+            checkTitle(tester, user.fullName, 2);
+            checkBotIcon(count: looksBot ? 1 : 0);
+          }
+
+          testWidgets('bot recipient -> shows no bot icon', (WidgetTester tester) async {
+            await checkRecipient(tester, isBot: true, looksBot: false);
+          });
```
chrisbobbe, line 304: We would like the bot icon to appear even when the name is long; see Alya's comment: https://github.com/zulip/zulip-flutter/pull/764#issuecomment-2207386553

If that's hard to do and has to be left as a TODO, that might be OK, but we shouldn't have a test that confirms the undesired behavior. 🙂

lib/widgets/profile.dart
```diff
@@ -46,10 +47,23 @@ class ProfilePage extends StatelessWidget {
...
+              color: Color.fromARGB(255, 159, 173, 173),
+            ),
+            const SizedBox(width: 10),
+          ],
+          Flexible(child: Text(user.fullName,
+            textAlign: TextAlign.center,
+            style: _TextStyles.primaryFieldText
+              .merge(weightVariableTextStyle(context, wght: 700)))),
+        ],
+      ),
```
chrisbobbe, line 66: I think it would look better to style this like we do in the other places: first the name, then the bot icon, as an inline `WidgetSpan`. Is the style in this revision based on something you'd like me to look at?

lib/widgets/profile.dart
```diff
@@ -46,10 +47,23 @@ class ProfilePage extends StatelessWidget {
...
-        style: _TextStyles.primaryFieldText
-          .merge(weightVariableTextStyle(context, wght: 700))),
+      Row(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: [
+          if (user.isBot) ...[
+            const Icon(
+              ZulipIcons.bot,
+              size: 22,
+              color: Color.fromARGB(255, 159, 173, 173),
```
chrisbobbe, line 57: Let's put this in the new `DesignVariables`, and use the new variable here and in the message list (where we added the bot icon for senders in #605).

chrisbobbe (commented):
Thanks @sm-sayedi! Comments below.

gnprice:
Thanks @sm-sayedi for your work on this, and @hackerkid @chrisbobbe for the reviews!

After reviewing our priorities over the past week, this issue #636 is one I'm content to leave open past the launch of the app. So let's leave this PR for it aside; it'll serve as a basis to start from when someone comes back later, after launch, to pick up the issue and finish it. I'll close the PR to get it out of our dashboards since we won't actively be working on it, but it'll remain linked from the issue so easy to find when the time comes.

In the meantime, we have plenty of [other issues](https://github.com/orgs/zulip/projects/5/views/1) to focus our energy on :slightly_smiling_face:

Issue #421, opened 2023-11-28
Update message list on muting/unmuting a stream or topic


gnprice:
In my first implementation of stream and topic muting #346, I intend to leave in place one missing form of event update: if you have a message list you're already looking at (or already on the navigation stack), and then you mute or unmute a stream or topic so that it should cause some messages to disappear or appear in that message list, it won't notice that it should update.

This isn't fundamentally hard — just requires a bit of wiring up of the information — but it'll be a little bit of refactoring, and there's already enough refactoring going on in my draft for #346 that I'd like to keep the complexity down for now.

We don't yet offer UI within the app for muting or unmuting streams and topics (that's #347 and #348), so it won't currently be easy to actually run into this issue.

chrisbobbe:
Unsubscribing from a stream means clearing out its mute state (`isMuted` is stored on the subscription). I wonder if that means message lists should also sometimes update on `SubscriptionRemoveEvent`s.

I don't think it's possible for a stream to be muted at the time you subscribe to it; i.e., the subscription has `isMuted: true` in the `SubscriptionAddEvent`. But if it is, we should handle that too.

gnprice:
> I wonder if that means message lists should also sometimes update on `SubscriptionRemoveEvent`s.

Yeah, seems right. Filed that as #821.

chrisbobbe:
#822 didn't handle updating the message list on muting/unmuting a channel, and it looks like that's still unhandled. 🙂 I've opened #1255 to revive the channel part of this issue.

PR #766, opened 2024-06-26
message: Handle message deletion event.

PIG208:
When handling a message deletion event, we just simply remove the messages from the per account store and wipe them from the message list view if they are present.

Because message deletion can affect what items are to be seen (date separator, recipient header, etc.), if the view has any message deleted, a `_reprocessAll` call is triggered. It would be ideal to just removed the affected items for better performance.

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -155,6 +155,30 @@ mixin _MessageSequence {
     _processMessage(messages.length - 1);
   }
 
+  /// Removes the given messages, if present.
+  ///
+  /// Returns true if at least one message was present, false otherwise.
+  /// If none of [messageIds] are found, this is a no-op.
+  bool _removeMessagesById(Iterable<int> messageIds) {
+    final messagesToRemoveById = <int>{};
+    final contentToRemove = Set<ZulipContent>.identity();
+    for (final messageId in messageIds) {
+      final index = _findMessageWithId(messageId);
+      if (index == -1) continue;
+      messagesToRemoveById.add(messageId);
+      contentToRemove.add(contents[index]);
+    }
+    if (messagesToRemoveById.isEmpty) return false;
+
+    assert(contents.length == messages.length);
+    messages.removeWhere((message) => messagesToRemoveById.contains(message.id));
+    contents.removeWhere((content) => contentToRemove.contains(content));
+    assert(contents.length == messages.length);
+    _reprocessAll();
+
+    return true;
+  }
+
   void _insertAllMessages(int index, Iterable<Message> toInsert) {
     // TODO parse/process messages in smaller batches, to not drop frames.
     //   On a Pixel 5, a batch of 100 messages takes ~15-20ms in _insertAllMessages.
@@ -425,6 +449,12 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     } finally {
       _fetchingOlder = false;
       _updateEndMarkers();
+      notifyListeners();
+    }
+  }
+
+  void handleDeleteMessageEvent(DeleteMessageEvent event) {
+    if (_removeMessagesById(event.messageIds)) {
       notifyListeners();
     }
   }
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -388,6 +388,21 @@ const _unreadMsgs = unreadMsgs;
 // Events.
 //
 
+DeleteMessageEvent deleteMessageEvent(List<StreamMessage> messages) {
+  assert(messages.isNotEmpty);
+  final streamId = messages.first.streamId;
+  final topic = messages.first.topic;
+  assert(messages.every((m) => m.streamId == streamId));
+  assert(messages.every((m) => m.topic == topic));
+  return DeleteMessageEvent(
+    id: 0,
+    messageIds: messages.map((message) => message.id).toList(),
+    messageType: MessageType.stream,
+    streamId: messages[0].streamId,
+    topic: messages[0].topic,
+  );
+}
+
 UpdateMessageEvent updateMessageEditEvent(
   Message origMessage, {
   int? userId = -1, // null means null; default is [selfUser.userId]
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -269,6 +269,67 @@ void main() {
     check(model).fetched.isFalse();
   });
 
+  group('DeleteMessageEvent', () {
+    final stream = eg.stream();
+    final messages = List.generate(30, (i) => eg.streamMessage(stream: stream));
+
+    test('in narrow', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages);
+
+      check(model).messages.length.equals(30);
+      await store.handleEvent(eg.deleteMessageEvent(messages.sublist(0, 10)));
+      checkNotifiedOnce();
+      check(model).messages.length.equals(20);
+    });
+
+    test('not all in narrow', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages.sublist(5));
+
+      check(model).messages.length.equals(25);
+      await store.handleEvent(eg.deleteMessageEvent(messages.sublist(0, 10)));
+      checkNotifiedOnce();
+      check(model).messages.length.equals(20);
+    });
+
+    test('not in narrow', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages.sublist(5));
+
+      check(model).messages.length.equals(25);
+      await store.handleEvent(eg.deleteMessageEvent(messages.sublist(0, 5)));
+      checkNotNotified();
+      check(model).messages.length.equals(25);
+    });
+
+    test('complete message deletion', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages.sublist(0, 25));
+
+      check(model).messages.length.equals(25);
+      await store.handleEvent(eg.deleteMessageEvent(messages));
+      checkNotifiedOnce();
+      check(model).messages.length.equals(0);
+    });
+
+    test('non-consecutive message deletion', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages);
+      final messagesToDelete = messages.sublist(2, 5) + messages.sublist(10, 15);
+
+      check(model).messages.length.equals(30);
+      await store.handleEvent(eg.deleteMessageEvent(messagesToDelete));
+      checkNotifiedOnce();
+      final expected = [
+        ...messages.sublist(0, 2),
+        ...messages.sublist(5, 10),
+        ...messages.sublist(15)].map((message) => message.id);
+      check(model).messages.length.equals(expected.length);
+      check(model.messages.map((message) => message.id)).deepEquals(expected);
+    });
+  });
+
   group('notifyListenersIfMessagePresent', () {
     test('message present', () async {
       await prepare(narrow: const CombinedFeedNarrow());
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -321,12 +321,11 @@ void main() {
       check(model).messages.length.equals(30);
       await store.handleEvent(eg.deleteMessageEvent(messagesToDelete));
       checkNotifiedOnce();
-      final expected = [
+      check(model.messages.map((message) => message.id)).deepEquals([
         ...messages.sublist(0, 2),
         ...messages.sublist(5, 10),
-        ...messages.sublist(15)].map((message) => message.id);
-      check(model).messages.length.equals(expected.length);
-      check(model.messages.map((message) => message.id)).deepEquals(expected);
+        ...messages.sublist(15),
+      ].map((message) => message.id));
     });
   });
 
--- a/test/model/message_test.dart
+++ b/test/model/message_test.dart
@@ -375,6 +375,39 @@ void main() {
     });
   });
 
+  group('handleDeleteMessageEvent', () {
+    test('delete an unknown message', () async {
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      await prepare();
+      await prepareMessages([message1]);
+      await store.handleEvent(eg.deleteMessageEvent([message2]));
+      checkNotNotified();
+      check(store).messages.values.single.id.equals(message1.id);
+    });
+
+    test('delete messages', () async {
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      await prepare();
+      await prepareMessages([message1, message2]);
+      await store.handleEvent(eg.deleteMessageEvent([message1, message2]));
+      checkNotifiedOnce();
+      check(store).messages.isEmpty();
+    });
+
+    test('delete an unknown message with a known message', () async {
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      final message3 = eg.streamMessage();
+      await prepare();
+      await prepareMessages([message1, message2]);
+      await store.handleEvent(eg.deleteMessageEvent([message2, message3]));
+      checkNotifiedOnce();
+      check(store).messages.values.single.id.equals(message1.id);
+    });
+  });
+
   group('handleReactionEvent', () {
     test('add reaction', () async {
       final originalMessage = eg.streamMessage(reactions: []);
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -175,7 +175,12 @@ class MessageStoreImpl with MessageStore {
   }
 
   void handleDeleteMessageEvent(DeleteMessageEvent event) {
-    // TODO handle DeleteMessageEvent, particularly in MessageListView
+    for (final messageId in event.messageIds) {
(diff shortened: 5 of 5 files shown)


test/model/message_test.dart
```diff
@@ -375,6 +375,40 @@ void main() {
     });
   });
 
+  group('handleDeleteMessageEvent', () {
+
+    test('delete unknown message', () async {
```
chrisbobbe, line 380: nit: no blank line between `group` and first `test`

test/example_data.dart
```diff
@@ -388,6 +388,16 @@ const _unreadMsgs = unreadMsgs;
...
 
+DeleteMessageEvent deleteMessageEvent(List<StreamMessage> messages) {
+  return DeleteMessageEvent(
+    id: 0,
+    messageIds: messages.map((message) => message.id).toList(),
+    messageType: MessageType.stream,
+    streamId: messages[0].streamId,
+    topic: messages[0].topic,
+  );
+}
```
chrisbobbe, line 399: Could be good to assert that all the messages have the same stream ID and topic. The [API doc](https://zulip.com/api/get-events#delete_message) isn't totally clear, but it seems like an imagined event with messages in different streams/topics wouldn't be representative. Maybe like this:

```suggestion
DeleteMessageEvent deleteMessageEvent(List<StreamMessage> messages) {
  assert(messages.isNotEmpty);
  final streamId = messages.first.streamId;
  final topic = messages.first.topic;
  assert(messages.every((m) => m.streamId == streamId));
  assert(messages.every((m) => m.topic == topic));
  return DeleteMessageEvent(
    id: 0,
    messageIds: messages.map((message) => message.id).toList(),
    messageType: MessageType.stream,
    streamId: streamId,
    topic: topic,
  );
}
```

lib/model/message_list.dart
```diff
@@ -155,6 +155,27 @@ mixin _MessageSequence {
     _processMessage(messages.length - 1);
   }
 
+  /// Given message ids, remove the corresponding messages from the view, and
+  /// return true if at least one message has been removed, or false otherwise.
+  ///
+  /// If none of the message ids provided can be found, this is essentially
+  /// a no-op.
```
chrisbobbe, line 162: Here's one idea of how this could be made a little more concise:

```dart
  /// Removes the given messages, if present.
  ///
  /// Returns true if at least one message was present, false otherwise.
  /// If none of [messageIds] are found, this is a no-op.
```

I think all the other relevant facts are implied.

test/model/message_test.dart
```diff
@@ -375,6 +375,40 @@ void main() {
...
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      await prepare();
+      await prepareMessages([message1, message2]);
+      await store.handleEvent(eg.deleteMessageEvent([message1, message2]));
+      checkNotifiedOnce();
+      check(store).messages.isEmpty();
+    });
+
+    test('delete an unknown messages with a known messages', () async {
```
chrisbobbe, line 400:
```suggestion
    test('delete an unknown message and a known message', () async {
```
I think?

lib/model/message_list.dart
```diff
@@ -155,6 +155,27 @@ mixin _MessageSequence {
...
+
+      assert(contents.length == messages.length);
+      messages.removeAt(messageIndex);
+      contents.removeAt(messageIndex);
+      assert(contents.length == messages.length);
+      hasRemovedMessage = true;
+    }
+    if (hasRemovedMessage) _reprocessAll();
+    return hasRemovedMessage;
+  }
```
chrisbobbe, line 177: One thing I notice in this implementation is that the `List.removeAt` calls can be expensive, if `messageIndex` is low and the list's length is large. From the [doc](https://api.flutter.dev/flutter/dart-core/List/removeAt.html); this part in particular:

> […] and moves all later objects down by one position.

There are ways we could be more efficient in that scenario (low `messageIndex`, large length), but I think they'd all come with tradeoffs in other scenarios. Also, bulk-delete events probably aren't very common anyway. But mentioning this in case you or Greg see something that's obviously an improvement.

gnprice: Yeah, the cost of this loop is quadratic, which is enough to make me nervous — with $n$ messages in the list, if the event has $k$ messages to delete and $d$ of them are in this message list, it's $\Theta(k \log n + d n)$. (The first term is from the `_findMessageWithId` lookups.)

In particular if you're looking at a conversation of $n$ messages, and an admin decides to delete the whole conversation so $d = k = n$, that's $\Theta(n^2)$.

An algorithm that would fix that is:
* First do the lookups, to determine if there are any affected messages in this list at all. That costs $\Theta(k \log n)$. (And saves us from spending $\Theta(n)$ on the next step if $d = 0$. If $d > 1$, we're resigned to spending that much time anyway.)
* Then, if there are any affected messages, use `List.removeWhere` to remove them from each of the lists all at once. (Build a `Set` out of `messageIds` first, to make that efficient.) This costs $\Theta(k + n)$. (The first term is for building the `Set`; the second for the `removeWhere` calls.)

That makes the total cost $O(k \log n + n)$ — no more quadratics. And it keeps it $O(k \log n)$ for the case where no messages were affected.

test/model/message_test.dart
```diff
@@ -375,6 +375,40 @@ void main() {
     });
   });
 
+  group('handleDeleteMessageEvent', () {
```
gnprice, line 378: This will also need tests for the behavior of `MessageListView`, as well as that of `MessageStoreImpl`. After all, the former is where most of the code this change introduces is. :slightly_smiling_face:

Those tests can go in `model/message_list_test.dart`; search for MessageEvent there for examples to follow. I think, given that we're using `_reprocessAll` rather than trying to do something more clever (as `handleMessageEvent` does), the only tests needed here are analogous to the tests with "MessageEvent" in their names.

Note one key thing those tests do is they call that `checkInvariants` function, in particular on every `notifyListeners`. It conducts a rather fierce set of consistency checks on the model, in particular recomputing everything that derives from `messages` and confirming it all has the right values.

gnprice (commented):
Thanks @PIG208, and thanks @chrisbobbe for the initial review!

Comments below.

    Because message deletion can affect what items are to be seen (date
    separator, recipient header, etc.), if the view has any message deleted, a
    `_reprocessAll` call is triggered. It would be ideal to just removed the
    affected items for better performance.

Yeah, I'm content to have deletion take linear time in the length of the existing list, which is what `_reprocessAll` does. We're already doomed to linear time just for updating `messages` itself, so long as it's a plain list.

(And as Chris pointed to above and I go into more detail about below, this current revision can actually spend *quadratic* time updating `messages` — so `_reprocessAll` is cheap by comparison with that.)

lib/model/message_list.dart
PIG208: Thanks for the time complexity analysis! So [`List.removeWhere`](https://api.dart.dev/stable/2.9.3/dart-core/List/removeWhere.html) performs in linear time? I couldn't find reference for how it is implemented. My guess is that they just make a copy of the list with the items removed.

PIG208: Looks like removing content with `List.removeWhere` wouldn't be quite straightforward because it is a list parallel to messages that don't have the message ID.

test/model/message_test.dart
PIG208: Oh right, there used to be more messages.

test/model/message_test.dart
PIG208: > Note one key thing those tests do is they call that checkInvariants function, in particular on every notifyListeners. It conducts a rather fierce set of consistency checks on the model, in particular recomputing everything that derives from messages and confirming it all has the right values.

I see that in the 'MessageEvent, before fetch' test, but not the other 'MessageEvent' tests. Was that intended?

lib/model/message_list.dart
gnprice: > So [`List.removeWhere`](https://api.dart.dev/stable/2.9.3/dart-core/List/removeWhere.html) performs in linear time? I couldn't find reference for how it is implemented. My guess is that they just make a copy of the list with the items removed.

My guess is it's a bit more efficient than that in that it reuses the existing array, and just moves items down to lower indexes to cover gaps. But yeah, should definitely be linear time.

The docs on List are a bit lacking in spelling out the performance characteristics, I think in part because there isn't a great place to put that information; `List` itself is actually an interface, and implementations can have different performance characteristics. But on `List` it does say:
> The default growable list, as created by [], keeps an internal buffer, and grows that buffer when necessary. This guarantees that a sequence of [add](https://api.dart.dev/stable/3.4.4/dart-core/List/add.html) operations will each execute in amortized constant time. Setting the length directly may take time proportional to the new length, and may change the internal capacity so that a following add operation will need to immediately increase the buffer capacity. Other list implementations may have different performance behavior.

So basically for each method you should think of it as doing whatever is the natural most efficient thing it could be doing for that method, given that data structure.

That data structure is essentially the same as a Python list or a JS array, if that helps in thinking about it.

(btw that's an old Dart version in that docs link; if you delete the "2.9.3/" for the version so it just says "stable", it'll give you the current version.)

PIG208:
The PR has been updated to address the reviews.
- Used the proposed new algorithm to delete messages from the message list
- Add tests to `test/model/message_list_test.dart`

lib/model/message_list.dart
gnprice: > Looks like removing content with `List.removeWhere` wouldn't be quite straightforward because it is a list parallel to messages that don't have the message ID.

Hmm, I see.

In that case probably the best thing is to write the loop ourselves. Essentially do this:

> reuses the existing array, and just moves items down to lower indexes to cover gaps.

but do the testing only on `messages`, and the moving on both `messages` and `content`.

test/model/message_test.dart
gnprice: It's there :slightly_smiling_face: Via this part:
> in particular on every `notifyListeners`.

See `prepare` for that call site.

gnprice: (Quite possibly we should have `checkNotNotified` call `checkInvariants` too, and then the individual test cases will be more uniform in not needing to explicitly do so)

PIG208: Oh, I see! So the call is added in a listener. Then the post event check of 'MessageEvent, not in narrow' seems to be missing because it isn't expecting a notification. Yeah it makes sense to add it to `checkNotNotified` too.

lib/model/message_list.dart
```diff
@@ -478,6 +502,12 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     }
   }
 
+  void messagesRemoved(List<int> messageIds) {
```
gnprice, line 505: nit: let's style this as another event-handling method, so:
* name and signature reflect handling an event — so the same as the method on `MessageStoreImpl`
* ordered in the class next to the other event-handling methods, and in the same order the event types are defined — so just above `handleMessageEvent`

The body will be the same, so the guts of it will still be that `_removeMessagesById` defined on `_MessageSequence`. But that way basically it's this class's responsibility to determine that that's what needs to be done.

That `_removeMessagesById` method is also the part that will get reused, possibly with some adaptation, for other operations like muting a subset of messages or moving some messages out of the narrow.

lib/model/message_list.dart
```diff
@@ -155,6 +155,30 @@ mixin _MessageSequence {
     _processMessage(messages.length - 1);
   }
 
+  /// Removes the given messages, if present.
+  ///
+  /// Returns true if at least one message was present, false otherwise.
+  /// If none of [messageIds] are found, this is a no-op.
+  bool _removeMessagesById(Iterable<int> messageIds) {
+    final messagesToRemoveById = <int>{};
+    final contentToRemove = <ZulipContent>{};
```
gnprice, line 164:
```suggestion
    final contentToRemove = Set<ZulipContent>.identity();
```

That has no direct effect, because `ZulipContent` doesn't override `==` or `hashCode`. But it's not totally obvious that it wouldn't — indeed some of the other `ContentNode` subclasses do — so best to be explicit.

And comparing for identity, rather than general equality, is crucial to this algorithm's behavior both for performance and, in edge cases, for correctness.

lib/model/message_list.dart
```diff
@@ -155,6 +155,30 @@ mixin _MessageSequence {
...
+      final index = _findMessageWithId(messageId);
+      if (index == -1) continue;
+      messagesToRemoveById.add(messageId);
+      contentToRemove.add(contents[index]);
+    }
+    if (messagesToRemoveById.isEmpty) return false;
+
+    assert(contents.length == messages.length);
+    messages.removeWhere((message) => messagesToRemoveById.contains(message.id));
+    contents.removeWhere((content) => contentToRemove.contains(content));
```
gnprice, line 175: Neat solution. I think this algorithm works fine.

I might replace it with a direct loop if I find that convenient in the course of building the other features (like #421) that will reuse a version of this method, but I'll cross that bridge if I come to it.

test/model/message_list_test.dart
```diff
@@ -271,6 +271,42 @@ void main() {
     checkInvariants(model);
   });
 
+  group('DeleteMessageEvent', () {
+    final stream = eg.stream();
+    final messages = List.generate(30, (i) => eg.streamMessage(stream: stream));
+
+    test('in narrow', () async {
```
gnprice, line 278: Let's also have a test case where the deleted messages are non-consecutive, and don't include the first or last message. That's one of the things that makes the logic complicated.

Also where all the messages get deleted. That seems like the sort of thing that a buggy implementation could get something wrong on.

gnprice (commented):
Thanks for the revision! All looks good modulo a couple of small comments.

lib/model/message_list.dart
PIG208: Ah that's interesting. So an identity set uses [identical](https://api.dart.dev/stable/3.4.4/dart-core/identical.html) to distinguish elements, which compares the object references instead.

PIG208:
The PR has been updated. Thanks for the review @gnprice !

lib/model/message_list.dart
PIG208: Got it. Leaving it as-is for this revision.

test/model/message_list_test.dart
```diff
@@ -271,6 +271,68 @@ void main() {
...
+      check(model).messages.length.equals(25);
+      await store.handleEvent(eg.deleteMessageEvent(messages));
+      checkNotifiedOnce();
+      check(model).messages.length.equals(0);
+    });
+
+    test('non-consecutive message deletion', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages);
+      final messageToDelete = messages.sublist(2, 5) + messages.sublist(10, 15);
```
gnprice, line 322: nit:
```suggestion
      final messagesToDelete = messages.sublist(2, 5) + messages.sublist(10, 15);
```

test/model/message_list_test.dart
```diff
@@ -271,6 +271,68 @@ void main() {
...
+
+      check(model).messages.length.equals(30);
+      await store.handleEvent(eg.deleteMessageEvent(messageToDelete));
+      checkNotifiedOnce();
+      final remainingMessageIds = [
+        ...messages.sublist(0, 2),
+        ...messages.sublist(5, 10),
+        ...messages.sublist(15)].map((message) => message.id).toSet();
+      check(model).messages.length.equals(22);
```
gnprice, line 331: nit:
```suggestion
      check(model).messages.length.equals(remainingMessageIds.length);
```

to make it less of a magic number

test/model/message_list_test.dart
```diff
@@ -271,6 +271,68 @@ void main() {
...
+      check(model.messages.map((message) => message.id)).containsInOrder(remainingMessageIds);
```
gnprice, line 332: Hmm, this doesn't seem like the right check. It checks that all the elements from the expected iterable are present in the actual iterable, and in the same order… but the expected iterable is a Set, which doesn't necessarily iterate through its elements in any predictable order.

And in fact the `Iterable.toSet` method's doc explicitly says:
> The order of the elements in the set is not guaranteed to be the same as for the iterable.

So this test could flake if that set's iteration order changes.

How about leaving off the `toSet`, and using `.deepEquals(remainingMessageIds)`?

test/model/message_list_test.dart
```diff
@@ -271,6 +271,68 @@ void main() {
...
+
+    test('non-consecutive message deletion', () async {
+      await prepare(narrow: StreamNarrow(stream.streamId));
+      await prepareMessages(foundOldest: true, messages: messages);
+      final messageToDelete = messages.sublist(2, 5) + messages.sublist(10, 15);
+
+      check(model).messages.length.equals(30);
+      await store.handleEvent(eg.deleteMessageEvent(messageToDelete));
+      checkNotifiedOnce();
+      final remainingMessageIds = [
```
gnprice, line 327: Oh and a nit: "remaining message IDs" sounds like a description of which message IDs are *actually* remaining, but this variable is computed from our expectations of which ones *should* be remaining. So better to give it a name with "expected" in it. For example:
```suggestion
      final expectedRemainingMessageIds = [
```

but that's a bit wordy. Can just say:
```suggestion
      final expected = [
```
since this is the only "expected" thing being checked against in the remainder of the test.

gnprice (commented):
Thanks for the revision! Remaining comments are all in the tests; one flaky test (at least in principle), and a few nits.

PIG208:
Thanks for the review! I have updated the PR and also removed an outdated `checkInvariants` call.

gnprice:
Thanks! All looks great; merging, with one added commit for a couple of nits:
1f9756379 message test [nfc]: Tighten message-deletion test slightly

PR #756, opened 2024-06-20
content: Implement Pygments Monokai, for upcoming dark theme

chrisbobbe:
For dark theme, web currently uses a mix of Pygments Monokai and Pygments Default. Mixing in Pygments Default seems accidental, so here we just use pure Pygments Monokai. [Discussion](https://chat.zulip.org/#narrow/stream/431-redesign-project/topic/code.20span.20colors/near/1832283).

I've included screenshots, below, of just Pygments Monokai, and of Pygments Monokai with web's adjustments. (Except for the adjustment of putting a border around `err` text; that's hard in Flutter.)

Compare with Pygments's own demonstration of Pygments Monokai here: https://pygments.org/styles/

(To make these screenshots, I plugged in dark-theme colors for the code-block background and message background. So just focus on the code blocks and disregard the jarring appearance of light-theme-styled elements elsewhere. When we implement dark theme #95, we'll do it coherently.)

| This PR (pure Pygments Monokai) | Following web (having some Pygments Default mixed in) |
| --- | --- |
| 8F5F5901-6293-4CF6-9A16-13516A8C2813 | D497DCF5-8115-469B-BBCF-C6827DF33DEE |

Fixes: #749

PR #767, opened 2024-06-26
content [nfc]: Finish writing down dark-theme color variants for all content

chrisbobbe:
Following #756 (dark-theme variant for code-block syntax highlighting), this completes the sweep through message-content colors for #95 dark theme. By the end of the branch, all colors are defined either in `ContentTheme` with light/dark variants, or still inline but with a comment explaining that they don't need light/dark variants.

A dev-only commit at the end makes it handy to see the results, but we shouldn't merge it because it just hard-codes dark theme in message content and nowhere else. 🙂 I recommend scrolling through "Combined feed" to see if anything stands out.

There's a tiny color change in light theme—

```
content: Adjust inline-code background color to match web
```

—as it seems either f452053dc was wrong or the web app changed since then. I can make screenshots if that would be helpful.

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -34,6 +34,7 @@ import 'text.dart';
 class ContentTheme extends ThemeExtension<ContentTheme> {
   factory ContentTheme.light(BuildContext context) {
     return ContentTheme._(
+      colorCodeBlockBackground: const HSLColor.fromAHSL(0.04, 0, 0, 0).toColor(),
       colorMessageMediaContainerBackground: const Color.fromRGBO(0, 0, 0, 0.03),
       colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor(),
       textStylePlainParagraph: _plainParagraphCommon(context).copyWith(
@@ -49,6 +50,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
 
   factory ContentTheme.dark(BuildContext context) {
     return ContentTheme._(
+      colorCodeBlockBackground: const HSLColor.fromAHSL(0.04, 0, 0, 1).toColor(),
       colorMessageMediaContainerBackground: const HSLColor.fromAHSL(0.03, 0, 0, 1).toColor(),
       colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor().withOpacity(0.2),
       textStylePlainParagraph: _plainParagraphCommon(context).copyWith(
@@ -63,6 +65,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
   }
 
   ContentTheme._({
+    required this.colorCodeBlockBackground,
     required this.colorMessageMediaContainerBackground,
     required this.colorThematicBreak,
     required this.textStylePlainParagraph,
@@ -81,6 +84,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
     return extension!;
   }
 
+  final Color colorCodeBlockBackground;
   final Color colorMessageMediaContainerBackground;
   final Color colorThematicBreak;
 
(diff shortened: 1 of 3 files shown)


chrisbobbe:
(Making this a draft because of that NOMERGE commit.)

lib/widgets/content.dart
```diff
@@ -32,33 +32,72 @@ import 'text.dart';
...
+      colorCodeBlockBackground: const HSLColor.fromAHSL(0.04, 0, 0, 0).toColor(),
+      colorDirectMentionBackground: const HSLColor.fromAHSL(0.2, 240, 0.7, 0.7).toColor(),
+      colorGlobalTimeBackground: const HSLColor.fromAHSL(1, 0, 0, 0.93).toColor(),
+      colorGlobalTimeBorder: const HSLColor.fromAHSL(1, 0, 0, 0.8).toColor(),
+      colorMathBlockBorder: const HSLColor.fromAHSL(0.15, 240, 0.8, 0.5).toColor(),
+      colorMessageMediaContainerBackground: const Color.fromRGBO(0, 0, 0, 0.03),
+      colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor(),
+      textStylePlainParagraph: _plainParagraphCommon(context)
+        .apply(color: const HSLColor.fromAHSL(1, 0, 0, 0.15).toColor())
+        .copyWith(debugLabel: 'ContentTheme.textStylePlainParagraph'),
```
gnprice, line 46: Is there something subtle `apply` is doing here that's different from another argument to `copyWith`?

lib/widgets/content.dart
```diff
@@ -32,33 +32,72 @@ import 'text.dart';
...
+    return ContentTheme._(
+      colorCodeBlockBackground: const HSLColor.fromAHSL(0.04, 0, 0, 1).toColor(),
+      colorDirectMentionBackground: const HSLColor.fromAHSL(0.25, 240, 0.52, 0.6).toColor(),
+      colorGlobalTimeBackground: const HSLColor.fromAHSL(0.2, 0, 0, 0).toColor(),
+      colorGlobalTimeBorder: const HSLColor.fromAHSL(0.4, 0, 0, 0).toColor(),
+      colorMathBlockBorder: const HSLColor.fromAHSL(1, 240, 0.4, 0.4).toColor(),
+      colorMessageMediaContainerBackground:
+        const HSLColor.fromAHSL(0.03, 0, 0, 1).toColor(),
+      colorThematicBreak:
+        const HSLColor.fromAHSL(1, 0, 0, .87).toColor().withOpacity(0.2),
```
gnprice, line 70: nit:
```suggestion
      colorMessageMediaContainerBackground: const HSLColor.fromAHSL(0.03, 0, 0, 1).toColor(),
      colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor().withOpacity(0.2),
```

to keep it parallel with the light constructor, for comparison.

(It's a little regrettable that the 0.2 opacity gets pushed off past 80 columns that way. But better that than breaking parallelism.)

lib/widgets/content.dart
```diff
@@ -1082,6 +1155,8 @@ class GlobalTime extends StatelessWidget {
             children: [
               Icon(
                 size: ambientTextStyle.fontSize!,
+                // (When GlobalTime appears in a link, it should be blue
+                // like the text.)
                 color: DefaultTextStyle.of(context).style.color!,
```
gnprice, line 1160: Huh interesting. I'm not sure that's entirely intentional! But I reproduce on web, so :shrug:

gnprice (commented):
Great to see! Two nits but otherwise please go ahead and merge (minus the NOMERGE commit).

I scrolled through the feed at tip of branch, and there was a lot that looked wrong but I think it was all outside the content, so expected.

I didn't attempt to verify all the styles taken from web. I'm assuming you did those accurately :slightly_smiling_face:. (And if one of them is somehow wrong, it's easy to fix later.)

lib/widgets/content.dart
chrisbobbe: Oh, no, I don't think so. I think I was just thinking of the `.copyWith(debugLabel` as its own separate step at the end, because we don't want anything to clobber the `debugLabel` we're giving it. But this change wouldn't cause it to be clobbered, so sounds good.

chrisbobbe:
Thanks for the review!

> Two nits but otherwise please go ahead and merge (minus the NOMERGE commit).

Done.

PR #769, opened 2024-06-27
msglist test: Call checkInvariants on checkNotNotified.

PIG208:
Stemmed from https://github.com/zulip/zulip-flutter/pull/766#discussion_r1657648710.

We already always check the invariants through a listener on the MessageListView instance. Now we do it more consistently when not notified.


gnprice (commented):

One comment below. There's also another call site on L270 that can be cut.

gnprice:
Thanks! Looks good; merging.

PR #770, opened 2024-06-28
fake_api [nfc]: Add `delay` option to connection.prepare

gnprice:
This allows simulating a request that takes some time to return and consequently races with something else. Useful for situations like at https://github.com/zulip/zulip-flutter/pull/713#discussion_r1657899344 .

This is NFC because `Future.delayed(Duration.zero, f)` is exactly equivalent to `Future(f)`.  (The docs aren't real clear on this; but reading the implementations confirms they boil down to the very same `Timer` constructor call, and then the docs do seem to be trying to say that when read in light of that information.)

### DIFF
--- a/test/api/fake_api.dart
+++ b/test/api/fake_api.dart
@@ -8,19 +8,22 @@ import 'package:zulip/model/store.dart';
 import '../example_data.dart' as eg;
 
 sealed class _PreparedResponse {
+  final Duration delay;
+
+  _PreparedResponse({this.delay = Duration.zero});
 }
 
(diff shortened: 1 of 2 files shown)


chrisbobbe:
LGTM, thanks! I'll hold off on merging so @sm-sayedi can take a look; I see you've requested his review.

sm-sayedi:
Looks good to me too. Thanks for this @gnprice.

gnprice:
Merged. Thanks for the reviews!

PR #778, opened 2024-06-30
Intl: Rename stream to channel in translation labels

Khader-1:
This is the first one of a series of PRs that target #631.

Fixes parts of #631

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -211,11 +211,11 @@
   "@composeBoxSelfDmContentHint": {
     "description": "Hint text for content input when sending a message to yourself."
   },
-  "composeBoxStreamContentHint": "Message #{stream} > {topic}",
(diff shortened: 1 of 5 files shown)


rajveermalviya (approved):
Thanks @Khader-1, LGTM :)

Over to @kenclary for mentor review.

kenclary (approved):
lgtm

chrisbobbe:
Thanks! LGTM except let's squash the second commit into the first, so that we don't have a commit where there's a reference that uses the wrong (outdated) name of something.

Marking for Greg's review.

gnprice:
Thanks all! This LGTM except the point Chris mentioned above, and one nit:

041479d0e intl: Rename string labels `stream` to `channel`

The commit-summary prefix we've mainly used for this area is `i18n:`.

I'll fix both of those and merge.

Issue #131, opened 2023-05-26
Offer mark-as-unread


gnprice:
This is the feature to mark the messages in the current narrow as *un*read, starting from a given message. The UI for this will be in the long-press menu on the message.

Prerequisite:
* #79 

Related but different:
* #81 
* #130

PR #779, opened 2024-06-30
action_sheet: Add "Mark as unread from here" button

Khader-1:
Fixes: #131
This adds a new button "**Mark as unread from here**" to the bottom sheet activated for read messages, when clicked all messages starting from selected one down are marked as unread.
| Before | After|
| --- | --- |
| image| image |

https://github.com/zulip/zulip-flutter/assets/63330019/b800f93f-6b9d-47d6-a41d-07daf3da1773

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -51,6 +51,10 @@
   "@actionSheetOptionCopyMessageLink": {
     "description": "Label for copy message link button on action sheet."
   },
+  "actionSheetOptionMarkAsUnread": "Mark as unread from here",
+  "@actionSheetOptionMarkAsUnread":  {
+    "description": "Label for mark as unread button on action sheet."
+  },
   "actionSheetOptionShare": "Share",
   "@actionSheetOptionShare":  {
     "description": "Label for share button on action sheet."
@@ -435,6 +439,21 @@
   "errorMarkAsReadFailedTitle": "Mark as read failed",
   "@errorMarkAsReadFailedTitle": {
     "description": "Error title when mark as read action failed."
+  },
+  "markAsUnreadComplete": "Marked {num, plural, =1{1 message} other{{num} messages}} as unread.",
+  "@markAsUnreadComplete": {
+    "description": "Message when marking messages as unread has completed.",
+    "placeholders": {
+      "num": {"type": "int", "example": "4"}
+    }
+  },
+  "markAsUnreadInProgress": "Marking messages as unread...",
+  "@markAsUnreadInProgress": {
+    "description": "Progress message when marking messages as unread."
+  },
+  "errorMarkAsUnreadFailedTitle": "Mark as unread failed",
+  "@errorMarkAsUnreadFailedTitle": {
+    "description": "Error title when mark as unread action failed."
   },
   "today": "Today",
   "@today": {
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -8,6 +8,7 @@ import '../api/model/model.dart';
 import '../api/route/messages.dart';
 import '../model/internal_link.dart';
 import '../model/narrow.dart';
+import 'actions.dart';
 import 'clipboard.dart';
 import 'compose_box.dart';
 import 'dialog.dart';
@@ -28,6 +29,10 @@ void showMessageActionSheet({required BuildContext context, required Message mes
   // any message list, so that's fine.
   final messageListPage = MessageListPage.ancestorOf(context);
   final isComposeBoxOffered = messageListPage.composeBoxController != null;
+  final narrow = messageListPage.narrow;
+  final isMessageRead = message.flags.contains(MessageFlag.read);
+  final markAsUnreadSupported = store.connection.zulipFeatureLevel! >= 155; // TODO(server-6)
+  final showMarkAsUnreadButton = markAsUnreadSupported && isMessageRead;
 
   final hasThumbsUpReactionVote = message.reactions
     ?.aggregated.any((reactionWithVotes) =>
@@ -45,6 +50,11 @@ void showMessageActionSheet({required BuildContext context, required Message mes
         if (isComposeBoxOffered) QuoteAndReplyButton(
           message: message,
           messageListContext: context,
+        ),
+        if (showMarkAsUnreadButton) MarkAsUnreadButton(
+          message: message,
+          messageListContext: context,
+          narrow: narrow,
         ),
         CopyMessageTextButton(message: message, messageListContext: context),
         CopyMessageLinkButton(message: message, messageListContext: context),
@@ -275,6 +285,29 @@ class QuoteAndReplyButton extends MessageActionSheetMenuItemButton {
     if (!composeBoxController.contentFocusNode.hasFocus) {
       composeBoxController.contentFocusNode.requestFocus();
     }
+  }
+}
+
+class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
+  MarkAsUnreadButton({
+    super.key,
+    required super.message,
+    required super.messageListContext,
+    required this.narrow,
+  });
+
+  final Narrow narrow;
+
+  @override IconData get icon => Icons.mark_chat_unread_outlined;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionMarkAsUnread;
+  }
+
+  @override void onPressed(BuildContext context) async {
+    Navigator.of(context).pop();
+    markNarrowAsUnreadFromMessage(messageListContext, message, narrow);
   }
 }
 
--- a/lib/widgets/actions.dart
+++ b/lib/widgets/actions.dart
@@ -16,14 +16,11 @@ import '../model/narrow.dart';
 import 'dialog.dart';
 import 'store.dart';
 
-Future<void> markNarrowAsRead(
-  BuildContext context,
-  Narrow narrow,
-  bool useLegacy, // TODO(server-6)
-) async {
+Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
   try {
     final store = PerAccountStoreWidget.of(context);
     final connection = store.connection;
+    final useLegacy = connection.zulipFeatureLevel! < 155; // TODO(server-6)
     if (useLegacy) {
       await _legacyMarkNarrowAsRead(context, narrow);
       return;
--- a/lib/widgets/actions.dart
+++ b/lib/widgets/actions.dart
@@ -16,7 +16,8 @@ import '../model/narrow.dart';
 import 'dialog.dart';
 import 'store.dart';
 
-Future<void> markNarrowAsRead(
+/// Returns true if mark as read process is completed successfully.
+Future<bool> markNarrowAsRead(
   BuildContext context,
   Narrow narrow,
   bool useLegacy, // TODO(server-6)
@@ -24,7 +25,8 @@ Future<void> markNarrowAsRead(
   final store = PerAccountStoreWidget.of(context);
   final connection = store.connection;
   if (useLegacy) {
-    return await _legacyMarkNarrowAsRead(context, narrow);
+    await _legacyMarkNarrowAsRead(context, narrow);
+    return true;
   }
 
   // Compare web's `mark_all_as_read` in web/src/unread_ops.js
@@ -66,7 +68,7 @@ Future<void> markNarrowAsRead(
       flag: MessageFlag.read);
     if (!context.mounted) {
       scaffoldMessenger.clearSnackBars();
-      return;
+      return false;
     }
     responseCount++;
     updatedCount += result.updatedCount;
@@ -81,7 +83,7 @@ Future<void> markNarrowAsRead(
           ..showSnackBar(SnackBar(behavior: SnackBarBehavior.floating,
               content: Text(zulipLocalizations.markAsReadComplete(updatedCount))));
       }
-      return;
+      return true;
     }
 
     if (result.lastProcessedId == null) {
@@ -91,7 +93,7 @@ Future<void> markNarrowAsRead(
       await showErrorDialog(context: context,
         title: zulipLocalizations.errorMarkAsReadFailedTitle,
         message: zulipLocalizations.errorInvalidResponse);
-      return;
+      return false;
     }
     anchor = NumericAnchor(result.lastProcessedId!);
 
--- a/lib/widgets/actions.dart
+++ b/lib/widgets/actions.dart
@@ -62,6 +62,26 @@ Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
   }
 }
 
+Future<void> markNarrowAsUnreadFromMessage(
+  BuildContext context,
+  Message message,
+  Narrow narrow,
+) async {
+  final connection = PerAccountStoreWidget.of(context).connection;
+  assert(connection.zulipFeatureLevel! >= 155); // TODO(server-6)
+  final zulipLocalizations = ZulipLocalizations.of(context);
+  await updateMessageFlagsStartingFromAnchor(
+    context: context,
+    apiNarrow: narrow.apiEncode(),
+    startingAnchor: NumericAnchor(message.id),
+    includeAnchor: true,
+    op: UpdateMessageFlagsOp.remove,
+    flag: MessageFlag.read,
+    onCompletedMessage: zulipLocalizations.markAsUnreadComplete,
+    progressMessage: zulipLocalizations.markAsUnreadInProgress,
+    onFailedTitle: zulipLocalizations.errorMarkAsUnreadFailedTitle);
+}
+
 /// Updates message flags by applying given operation `op` using given `flag`
 /// the update happens on given `apiNarrow` starting from given `startingAnchor`
 ///
--- a/lib/widgets/actions.dart
+++ b/lib/widgets/actions.dart
@@ -22,97 +22,106 @@ Future<bool> markNarrowAsRead(
   Narrow narrow,
   bool useLegacy, // TODO(server-6)
 ) async {
-  final store = PerAccountStoreWidget.of(context);
-  final connection = store.connection;
-  if (useLegacy) {
-    await _legacyMarkNarrowAsRead(context, narrow);
-    return true;
-  }
-
-  // Compare web's `mark_all_as_read` in web/src/unread_ops.js
-  // and zulip-mobile's `markAsUnreadFromMessage` in src/action-sheets/index.js .
-  final zulipLocalizations = ZulipLocalizations.of(context);
-  final scaffoldMessenger = ScaffoldMessenger.of(context);
-  // Use [AnchorCode.oldest], because [AnchorCode.firstUnread]
-  // will be the oldest non-muted unread message, which would
-  // result in muted unreads older than the first unread not
-  // being processed.
-  Anchor anchor = AnchorCode.oldest;
-  int responseCount = 0;
-  int updatedCount = 0;
-
-  // Include `is:unread` in the narrow.  That has a database index, so
-  // this can be an important optimization in narrows with a lot of history.
-  // The server applies the same optimization within the (deprecated)
-  // specialized endpoints for marking messages as read; see
-  // `do_mark_stream_messages_as_read` in `zulip:zerver/actions/message_flags.py`.
-  final apiNarrow = narrow.apiEncode()..add(ApiNarrowIsUnread());
-
-  while (true) {
-    final result = await updateMessageFlagsForNarrow(connection,
-      anchor: anchor,
-      // [AnchorCode.oldest] is an anchor ID lower than any valid
-      // message ID; and follow-up requests will have already
-      // processed the anchor ID, so we just want this to be
-      // unconditionally false.
-      includeAnchor: false,
-      // There is an upper limit of 5000 messages per batch
-      // (numBefore + numAfter <= 5000) enforced on the server.
-      // See `update_message_flags_in_narrow` in zerver/views/message_flags.py .
-      // zulip-mobile uses `numAfter` of 5000, but web uses 1000
-      // for more responsive feedback. See zulip@f0d87fcf6.
-      numBefore: 0,
-      numAfter: 1000,
-      narrow: apiNarrow,
-      op: UpdateMessageFlagsOp.add,
-      flag: MessageFlag.read);
-    if (!context.mounted) {
-      scaffoldMessenger.clearSnackBars();
-      return false;
-    }
-    responseCount++;
-    updatedCount += result.updatedCount;
-
-    if (result.foundNewest) {
-      if (responseCount > 1) {
-        // We previously showed an in-progress [SnackBar], so say we're done.
-        // There may be a backlog of [SnackBar]s accumulated in the queue
-        // so be sure to clear them out here.
-        scaffoldMessenger
-          ..clearSnackBars()
-          ..showSnackBar(SnackBar(behavior: SnackBarBehavior.floating,
-              content: Text(zulipLocalizations.markAsReadComplete(updatedCount))));
-      }
+  try {
+    final store = PerAccountStoreWidget.of(context);
+    final connection = store.connection;
+    if (useLegacy) {
+      await _legacyMarkNarrowAsRead(context, narrow);
       return true;
     }
 
-    if (result.lastProcessedId == null) {
-      // No messages were in the range of the request.
-      // This should be impossible given that `foundNewest` was false
-      // (and that our `numAfter` was positive.)
-      await showErrorDialog(context: context,
-        title: zulipLocalizations.errorMarkAsReadFailedTitle,
-        message: zulipLocalizations.errorInvalidResponse);
-      return false;
-    }
-    anchor = NumericAnchor(result.lastProcessedId!);
+    // Compare web's `mark_all_as_read` in web/src/unread_ops.js
+    // and zulip-mobile's `markAsUnreadFromMessage` in src/action-sheets/index.js .
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final scaffoldMessenger = ScaffoldMessenger.of(context);
+    // Use [AnchorCode.oldest], because [AnchorCode.firstUnread]
+    // will be the oldest non-muted unread message, which would
+    // result in muted unreads older than the first unread not
+    // being processed.
+    Anchor anchor = AnchorCode.oldest;
+    int responseCount = 0;
+    int updatedCount = 0;
 
-    // The task is taking a while, so tell the user we're working on it.
-    // No need to say how many messages, as the [MarkAsUnread] widget
-    // should follow along.
-    // TODO: Ideally we'd have a progress widget here that showed up based
-    //   on actual time elapsed -- so it could appear before the first
-    //   batch returns, if that takes a while -- and that then stuck
-    //   around continuously until the task ends. For now we use a
-    //   series of [SnackBar]s, which may feel a bit janky.
-    //   There is complexity in tracking the status of each [SnackBar],
-    //   due to having no way to determine which is currently active,
-    //   or if there is an active one at all.  Resetting the [SnackBar] here
-    //   results in the same message popping in and out and the user experience
-    //   is better for now if we allow them to run their timer through
-    //   and clear the backlog later.
-    scaffoldMessenger.showSnackBar(SnackBar(behavior: SnackBarBehavior.floating,
-      content: Text(zulipLocalizations.markAsReadInProgress)));
+    // Include `is:unread` in the narrow.  That has a database index, so
+    // this can be an important optimization in narrows with a lot of history.
+    // The server applies the same optimization within the (deprecated)
+    // specialized endpoints for marking messages as read; see
+    // `do_mark_stream_messages_as_read` in `zulip:zerver/actions/message_flags.py`.
+    final apiNarrow = narrow.apiEncode()..add(ApiNarrowIsUnread());
+
+    while (true) {
+      final result = await updateMessageFlagsForNarrow(connection,
+        anchor: anchor,
+        // [AnchorCode.oldest] is an anchor ID lower than any valid
+        // message ID; and follow-up requests will have already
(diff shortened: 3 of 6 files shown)


lib/widgets/action_sheet.dart
```diff
@@ -366,3 +368,34 @@ class CopyButton extends MessageActionSheetMenuItemButton {
...
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetMarkAsUnread;
+  }
+
+  @override void onPressed(BuildContext context) async {
+    Navigator.of(context).pop();
+    final ids = [
+      message.id,
+      ...model.getNarrowMessagesAfter(message).map((e) => e.id)];
```
Khader-1, line 393: Is there a standard (cleaner) way to get messages that are in the same narrow and come after a specific message?

Khader-1 (commented):
This is still in progress, I think there could be an easier way to do things..

lib/widgets/action_sheet.dart
```diff
@@ -400,3 +403,69 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+    return zulipLocalizations.actionSheetMarkAsUnread;
+  }
+
+  @override void onPressed(BuildContext context) async {
+    Navigator.of(context).pop();
+    final zulipLocalizations = ZulipLocalizations.of(messageListContext);
+
+    try {
+      final store = PerAccountStoreWidget.of(messageListContext);
+      final narrow = SendableNarrow.ofMessage(message, selfUserId: store.selfUserId);
```
sm-sayedi, line 427: This will always return a `TopicNarrow` no matter which narrow the message is shown on. This results in marking the messages unread just in a single topic narrow even if we are in a stream narrow.

sm-sayedi (commented):
Thanks @Khader-1 for this. A few comments below.

Also, I would write the commit message as such:

action_sheet: Add "Mark as unread from here" button

Using `` is usually used for distinguishing code-related words from the rest of the commit message.

lib/widgets/action_sheet.dart
Khader-1: Right! I wonder though is there another alternative other that receiving the narrow in showMessageActionSheet arguments?

sm-sayedi: I think passing that Narrow down is a proper option for now.

gnprice: Passing down a Narrow would work but would be a bit annoying because it'd have to be passed down through to the MessageWithPossibleSender that calls showMessageActionSheet.

I've just written up and sent PR #811, which provides a way to do this similar to what these action sheets already do to get the compose box controller (for quote-and-reply etc.). So @Khader-1 please go ahead and stack this PR atop that one, and then you can use that.

Khader-1: Thanks @gnprice! I've just done that in the new revision. PTAL @sm-sayedi.

kenclary (approved):
lgtm

alya:
@Khader-1 , as usual, please add screenshots of all UI changes to the PR description, in addition to any videos.

Khader-1:
Thanks @alya, I've updated the description.
Additionally rebased to the latest revision from #811.

alya:
Thanks! Putting the new option just between "Quote and reply" and "Copy message text" would be most consistent with the web app UI, so let's do that.

When an issue doesn't specify where to put a new option, you can go ahead and flag it for discussion.

alya:
We might need to figure out how to organize that menu soon -- it's getting long already, and there's more to come. :)

alya:
> Maybe consider filing an issue to track this? I'm happy to work on any reordering or restructuring that may be needed.

I'm not sure if we have a design for this, and what options are available for visually structuring the menu in Flutter. We'll probably need a #mobile discussion first.

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +405,70 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
+  MarkAsUnreadButton({
+    super.key,
+    required super.message,
+    required super.messageListContext,
+  }) : narrow = MessageListPage.ancestorOf(messageListContext).narrow;
+
+  final Narrow narrow;
+
+  @override IconData get icon => Icons.mark_chat_unread_outlined;
```
PIG208, line 418: Looks like we got this from Material Design. Since we don't have a design for this, maybe we can confirm that with Vlad?

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +405,70 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+      final connection = store.connection;
+      Anchor anchor = NumericAnchor(message.id);
+      final apiNarrow = narrow.apiEncode();
+
+      while (true) {
+        final result = await updateMessageFlagsForNarrow(connection,
+          anchor: anchor,
+          includeAnchor: true,
+          numBefore: 0,
+          numAfter: 1000,
```
PIG208, line 440: Seems like we have [5000](https://github.com/zulip/zulip-mobile/blob/e352f563ecf2fa9b09b688d5a65b6bc89b0358bc/src/action-sheets/index.js#L203-L280) from zulip-mobile, and [1000-1](https://github.com/zulip/zulip/blob/43e1a643c0321b351d1f091f02eacc95caafb796/web/src/unread_ops.ts#L250) from web. How did you pick this number?

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +405,70 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+          narrow: apiNarrow,
+          op: UpdateMessageFlagsOp.remove,
+          flag: MessageFlag.read);
+
+        if (!context.mounted) return;
```
PIG208, line 445: Looks like this is misused here. A similar issue from before: https://github.com/zulip/zulip-flutter/pull/779#discussion_r1676347611.

PIG208 (changes requested):
Left some comments, but I haven't gone through the tests.

Because of some similarity of the `onPressed` handler and `markNarrowAsRead`, we may extract the code in common in a prep commit. This can be reused as we implement "mark as unread".

gnprice:
Following up on @PIG208's review:

It looks like a good bit of the code in this button is copied from that `markNarrowAsRead` function over in `lib/widgets/message_list.dart`.

Copying code can be a fine choice, but it's definitely something to flag in the PR. That can save a lot of time for the reviewer because it lets them start with that existing code as a baseline and focus on the changes from it, rather than spending time reviewing the copied code from scratch.

gnprice:
OK, and I've just pushed some additional commits to this branch that set things up to be able to refactor in order to share that code:
* One commit moves `markNarrowAsRead` to a new file lib/widgets/actions.dart .
* Another takes the existing unit tests for that functionality, which were at the layer of the UI presented by MarkAsReadWidget, and ports most of them to operate on the function directly.

Then @Khader-1 for this PR, the steps in that direction that you'll need to do are:
* Separate the logic in `markNarrowAsRead` that is specific to marking as read from the logic that's also desirable for marking as unread. Probably there'll still be a `markNarrowAsRead` and it will call a more general function that does most of the work. This should be an NFC commit.
* Then as a second commit, convert the `markNarrowAsRead` tests to be in terms of the new more-general function, for the tests that are about that function's job. (Other tests will be about the things that remain specific to `markNarrowAsRead`, so they'll remain for that function.)
* Then rebase your original commit from this PR on top of those changes, and have it use the new more-general function instead of having a copy of the code.

Using the existing code will give us several things that aren't present in this copy:
* progress feedback to the user if there are lots of messages to go through
* tests for the somewhat tricky logic here that handles doing multiple requests for multiple batches

gnprice:
> OK, and I've just pushed some additional commits to this branch that set things up to be able to refactor in order to share that code

Sent those as their own PR #827, in the interest of getting them merged separately so as to reduce the quantity of changes we try to review all in one batch. So @Khader-1 you can rebase this PR atop that one and then proceed with the steps in my previous comment.

Khader-1:
Thank you, @PIG208 and @gnprice, for the reviews! I've just pushed a new revision that addresses all the points raised except for the icon selection decision, which will be discussed further [in this Zulip thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/.23F779.20Mark.20as.20unread.20from.20here).

test/widgets/actions_test.dart
```diff
@@ -101,18 +101,129 @@ void main() {
           });
     });
 
+    testWidgets('CombinedFeedNarrow on legacy server', (WidgetTester tester) async {
```
PIG208, line 104: When moving code around while refactoring, while there could be some cases where a split is not possible, it is generally preferred to have separate commits for the move and for the refactor to make it easier to review.

lib/widgets/actions.dart
```diff
@@ -21,41 +21,63 @@ Future<bool> markNarrowAsRead(
...
+  required MessageFlag flag,
+  required String Function(int num) onCompletedMessage,
+  required String onFailedTitle,
+  required String onProgressTitle,
+}) async {
+  try {
+    final store = PerAccountStoreWidget.of(context);
+    final connection = store.connection;
+
+    final zulipLocalizations = ZulipLocalizations.of(context);
```
PIG208, line 90: nit: At first I was confused as of why we still need this after having the parameters for the error messages, but then I saw it used for the `errorInvalidResponse` below. We can move `zulipLocalizations` to closer to where it is used, or even just inline it because it is only used once.

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +406,25 @@ class ShareButton extends MessageActionSheetMenuItemButton {
     }
   }
 }
+
+class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
+  MarkAsUnreadButton({
+    super.key,
+    required super.message,
+    required super.messageListContext,
+  }) : narrow = MessageListPage.ancestorOf(messageListContext).narrow;
```
PIG208, line 415: We want to find the narrow without calling this in the constructor or implicitly through the build method for performance reasons.

See: https://github.com/zulip/zulip-flutter/blob/dbb2dcfbf4e8e2239b2dca0142bbc3fab3cc0640/lib/widgets/message_list.dart#L48-L53

lib/widgets/actions.dart
```diff
@@ -16,101 +16,111 @@ import '../model/narrow.dart';
 import 'dialog.dart';
 import 'store.dart';
 
-Future<void> markNarrowAsRead(
+Future<bool> markNarrowAsRead(
```
PIG208, line 19: This might worth some dart doc explaining what it does and the meaning of the return value.

lib/widgets/actions.dart
```diff
@@ -21,41 +21,63 @@ Future<bool> markNarrowAsRead(
...
 ) async {
-  try {
-    final store = PerAccountStoreWidget.of(context);
-    final connection = store.connection;
-    if (useLegacy) {
-      await _legacyMarkNarrowAsRead(context, narrow);
-      return true;
-    }
+  if (useLegacy) {
+    await _legacyMarkNarrowAsRead(context, narrow);
```
PIG208, line 25: Now that the the try-catch has been moved inside the helper, making the call here can still throw. This is no longer an nfc. Let's check if there's some more handling needed here.

lib/widgets/actions.dart
```diff
@@ -21,41 +21,63 @@ Future<bool> markNarrowAsRead(
...
+    op: UpdateMessageFlagsOp.add,
+    flag: MessageFlag.read,
+    onCompletedMessage: zulipLocalizations.markAsReadComplete,
+    onFailedTitle: zulipLocalizations.errorMarkAsReadFailedTitle,
+    // No need to say how many messages, as the [MarkAsUnread] widget
+    // should follow along.
+    onProgressTitle: zulipLocalizations.markAsReadInProgress);
+}
+
+Future<bool> updateMessageFlagsStartingFromAnchor({
```
PIG208, line 75: Having dart doc here will be helpful too. Along with the other public helpers that has this return type. Something simple should suffice.

test/widgets/actions_test.dart
```diff
@@ -188,95 +302,23 @@ void main() {
...
+        context: context, apiNarrow: apiNarrow, op: UpdateMessageFlagsOp.add, flag: MessageFlag.read,
+        includeAnchor: false, startAnchor: AnchorCode.oldest, onCompletedMessage: onCompletedMessage,
+        onFailedTitle: onFailedTitle, onProgressTitle: onProgressTitle);
       await tester.pump(Duration.zero);
       await tester.pumpAndSettle();
       checkErrorDialog(tester,
-        expectedTitle: zulipLocalizations.errorMarkAsReadFailedTitle,
+        expectedTitle: onFailedTitle,
         expectedMessage: 'NetworkException: Oops (ClientException: Oops)');
     }, skip: true, // TODO move this functionality inside markNarrowAsRead
```
PIG208, line 323: In the first commit, we refactored `markNarrowAsRead` to handle this TODO. We should update this test in that commit and possibly not skip it.

PIG208 (commented):
Left some comments. Thanks for the updates @Khader-1. I haven't looked into the details of the actions_test change. Overall this seems cleaner now that we have extracted the helper.

kenclary:
What's the status of this? have all the requests been dealt with?

Khader-1:
Oh! Yes it is.
I should have left a comment here, Thanks @kenclary for the heads up!
I've rebased now PTAL @PIG208.

test/widgets/actions_test.dart
```diff
@@ -269,4 +270,37 @@ void main() {
...
+    const progressMessage = 'progressMessage';
+    const onFailedTitle = 'onFailedTitle';
+
+    testWidgets('smoke test', (tester) async {
+      final narrow = TopicNarrow.ofMessage(eg.streamMessage());
+      final apiNarrow = narrow.apiEncode()..add(ApiNarrowIsUnread());
+      await prepare(tester);
+      connection.prepare(json: UpdateMessageFlagsForNarrowResult(
+        processedCount: 11, updatedCount: 3,
+        firstProcessedId: null, lastProcessedId: null,
```
PIG208, line 197: Let's try to use realistic values for `firstProcessId` and `lastProcessedId`, to be consistent with what the server might return.

test/widgets/actions_test.dart
```diff
@@ -269,4 +270,37 @@ void main() {
...
+        ..url.path.equals('/api/v1/messages/flags/narrow')
+        ..bodyFields.deepEquals({
+            'anchor': 'oldest',
+            'include_anchor': 'false',
+            'num_before': '0',
+            'num_after': '1000',
+            'narrow': jsonEncode(apiNarrow),
+            'op': 'add',
+            'flag': 'read',
+          });
```
PIG208, line 215: It will be helpful to await the future returned by the previous `updateMessageFlagsStartingFromAnchor` call, and ensure that it returns the expected value, just so we know that it "did pass" as expected.

test/widgets/actions_test.dart
```diff
@@ -150,8 +150,7 @@ void main() {
       await tester.pump(Duration.zero);
       await tester.pumpAndSettle();
       check(store.unreads.oldUnreadsMissing).isFalse();
-    }, skip: true, // TODO move this functionality inside markNarrowAsRead
-    );
+    }, skip: true);
```
PIG208, line 108: Looks like we can stop skipping this test too

test/widgets/action_sheet_test.dart
```diff
@@ -536,4 +536,20 @@ void main() {
...
+      await setupToMessageActionSheet(tester, message: unreadMessage, narrow: TopicNarrow.ofMessage(unreadMessage));
+
+      check(find.byIcon(Icons.mark_chat_unread_outlined).evaluate().length).equals(0);
+    });
+
+    testWidgets('visible if message is read', (WidgetTester tester) async {
+      final message = eg.streamMessage(flags: [MessageFlag.read]);
+      await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
+
+      check(find.byIcon(Icons.mark_chat_unread_outlined).evaluate().length).equals(1);
```
PIG208, line 552: nit:
```suggestion
      check(find.byIcon(Icons.mark_chat_unread_outlined).evaluate()).single;
```

Similarly, the test above can use `isEmpty()`.

test/widgets/action_sheet_test.dart
```diff
@@ -536,4 +536,20 @@ void main() {
...
+    });
+  });
```
PIG208, line 554: Like the tests for other buttons, we can add more end-to-end tests where we tap the `MarkAsUnread` button and check the happy path or errors.

lib/widgets/actions.dart
```diff
@@ -67,6 +67,24 @@ Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
...
 
+Future<void> markNarrowAsUnreadFromMessage(
+  BuildContext context,
+  Message message,
+  Narrow narrow,
+) async {
+  final zulipLocalizations = ZulipLocalizations.of(context);
+  await updateMessageFlagsStartingFromAnchor(
+    context: context,
+    startingAnchor: NumericAnchor(message.id),
```
PIG208, line 78: nit: Let's order the arguments consistently. `startingAnchor` as a parameter comes after `apiNarrow`; `onFailedTitle` comes after `progressMessage`.

PIG208:
@Khader-1 Thanks for the revision! This looks great. Left some comments mostly about the tests.

Khader-1:
Thanks @PIG208! I've pushed a new revision PTAL!

test/widgets/actions_test.dart
```diff
@@ -256,39 +181,6 @@ void main() {
             'flag': 'read',
           });
     });
-
-    testWidgets('MentionsNarrow on legacy server', (WidgetTester tester) async {
```
PIG208, line 262: Looks like this test got removed but was not re-added.

PIG208 (commented):
Looks good other than one comment. Thanks for the revision @Khader-1!

Khader-1:
Thanks for the catch @PIG208! I've just pushed a new revision PTAL.
Marking this for integration review..

lib/widgets/message_list.dart
```diff
@@ -689,28 +688,9 @@ class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
...
-      showErrorDialog(context: context,
-        title: zulipLocalizations.errorMarkAsReadFailedTitle,
-        message: e.toString()); // TODO(#741): extract user-facing message better
-      return;
-    } finally {
-      setState(() => _loading = false);
-    }
-    if (!context.mounted) return;
-    if (widget.narrow is CombinedFeedNarrow && !useLegacy) {
-      PerAccountStoreWidget.of(context).unreads.handleAllMessagesReadSuccess();
```
gnprice, line 712: In the first commit:
f2dec7f84 actions [nfc]: Port generic error handling to `markNarrowAsRead`

there looks to be a change in when these lines get run. Previously we'd run them whenever there wasn't an exception; now we do so only when `markNarrowAsRead` returns true. And there are a couple of cases where it didn't throw an exception, but now returns false. That makes the commit not NFC.

I think those behavior changes are probably good, but let's separate them from the NFC change which affects more lines of code. A first commit can make the NFC change of moving that try/catch inside the function — to make it NFC, it should return false just when the old version would have thrown. Then a second commit can start returning false in more cases.

(Then continue with the next commit:
d4d26155b actions [nfc]: Port `handleAllMessagesReadSuccess` to `markNarrowAsRead`

That commit being NFC relies on having already made this non-NFC change.)

lib/widgets/actions.dart
```diff
@@ -16,101 +16,171 @@ import '../model/narrow.dart';
...
-    if (!context.mounted) {
-      scaffoldMessenger.clearSnackBars();
-      return;
+      flag: MessageFlag.read,
+      onCompletedMessage: zulipLocalizations.markAsReadComplete,
+      progressMessage: zulipLocalizations.markAsReadInProgress,
+      onFailedTitle: zulipLocalizations.errorMarkAsReadFailedTitle);
+
+    if (!didPass || !context.mounted) return;
+    if (narrow is CombinedFeedNarrow && !useLegacy) {
```
gnprice, line 57: Here `useLegacy` is always false, right?

gnprice (commented):
Thanks @Khader-1, and thanks @sm-sayedi @alya @kenclary @PIG208 for the previous reviews!

I've read through the first few commits:
f2dec7f84 actions [nfc]: Port generic error handling to `markNarrowAsRead`
d4d26155b actions [nfc]: Port `handleAllMessagesReadSuccess` to `markNarrowAsRead`
45a570255 actions [nfc]: Move legacy-server check inside markNarrowAsRead

Generally they look good — thanks for completing those refactors. Going AFK for a bit, so here's the comments I have on those.

Later I'll return and review the remaining commits:
4793beb6a actions [nfc]: Factor out `updateMessageFlagsStartingFromAnchor`
759453666 actions_test [nfc]: Remove unnecessary whitespace
337451a58 actions_test: Create `updateMessageFlagsStartingFromAnchor` test group
680d49105 actions_test [nfc]: Move generic cases out of markNarrowAsRead
4a8640802 actions_test [nfc]: Use updateMessageFlagsStartingFromAnchor in it's cases
a5525e71f action_sheet: Add "Mark as unread from here" button

lib/widgets/actions.dart
```diff
@@ -16,101 +16,171 @@ import '../model/narrow.dart';
...
+      PerAccountStoreWidget.of(context).unreads.handleAllMessagesReadSuccess();
```
gnprice, line 58: This gets moved inside the generic try/catch. It should stay outside — keeping each `try` block narrow, and focused on the place where we expect exceptions to be thrown without it being a bug in our code, helps make it possible to reason about what their effects are.

See discussion here:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#catch-specific

(This actually happens in the second commit, but it's something I noticed only while reading a later commit.)

lib/widgets/actions.dart
```diff
@@ -16,101 +16,171 @@ import '../model/narrow.dart';
...
     }
-    responseCount++;
-    updatedCount += result.updatedCount;
+  } catch (e) {
+    if (!context.mounted) return;
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    await showErrorDialog(context: context,
+      title: zulipLocalizations.errorMarkAsReadFailedTitle,
```
gnprice, line 64: Hmm — it doesn't feel right that we have two of these generic try/catch blocks nested one inside the other, and both reporting the same error. That feels like a sign that one of them is too broad.

This is what caused me to notice the point mentioned above, about the handleAllMessagesReadSuccess call getting moved inside the try/catch.

Then after that's fixed, I think the remaining thing this try/catch is covering is basically the _legacyMarkNarrowAsRead call. (There's no way updateMessageFlagsStartingFromAnchor can throw, short of a bug in its `catch` block; and there's little else here, and none of it can throw except through some surprising bug in our code.) So the remaining thing to do to fix this redundancy is to move this second try/catch to be inside that function.

lib/widgets/actions.dart
```diff
@@ -16,101 +16,171 @@ import '../model/narrow.dart';
...
+    if (useLegacy) {
+      await _legacyMarkNarrowAsRead(context, narrow);
+      return;
+    }
 
-  while (true) {
-    final result = await updateMessageFlagsForNarrow(connection,
-      anchor: anchor,
+    // Compare web's `mark_all_as_read` in web/src/unread_ops.js
+    // and zulip-mobile's `markAsUnreadFromMessage` in src/action-sheets/index.js .
```
gnprice, line 30: This comment belongs in updateMessageFlagsStartingFromAnchor, rather than one of its callers; it's really about the logic that now lives in that function.

test/widgets/actions_test.dart
```diff
@@ -278,19 +201,139 @@ void main() {
...
+      final apiNarrow = narrow.apiEncode()..add(ApiNarrowIsUnread());
+      await prepare(tester);
+      connection.prepare(json: UpdateMessageFlagsForNarrowResult(
+        processedCount: 11, updatedCount: 3,
+        firstProcessedId: 1, lastProcessedId: 1980,
+        foundOldest: true, foundNewest: true).toJson());
+      final didPass = updateMessageFlagsStartingFromAnchor(
+        context: context, apiNarrow: apiNarrow, op: UpdateMessageFlagsOp.add, flag: MessageFlag.read,
+        includeAnchor: false, startingAnchor: AnchorCode.oldest, onCompletedMessage: onCompletedMessage,
+        onFailedTitle: onFailedTitle, progressMessage: progressMessage);
```
gnprice, line 222: nit: lines too long; and the arguments grouped on each line don't have anything thematically in common, so the wrapping together onto the same line makes it harder to spot any given parameter or see what's going on.

Instead, expanding out onto a line per argument is fine.

To reduce repetition between these test cases, can make a local helper that calls this function and provides all the values that don't change from one case to another.

assets/l10n/app_en.arb
```diff
@@ -51,6 +51,10 @@
   "@actionSheetOptionCopyMessageLink": {
     "description": "Label for copy message link button on action sheet."
   },
+  "actionSheetMarkAsUnread": "Mark as unread from here",
```
gnprice, line 54:
```suggestion
  "actionSheetOptionMarkAsUnread": "Mark as unread from here",
```

i.e. match the pattern of the other strings around it

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +410,27 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+
+  @override IconData get icon => Icons.mark_chat_unread_outlined;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetMarkAsUnread;
+  }
+
+  @override void onPressed(BuildContext context) async {
+    if (!context.mounted) return;
```
gnprice, line 432: What's the motivation for this line? It doesn't appear in the surrounding other button implementations.

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +410,27 @@ class ShareButton extends MessageActionSheetMenuItemButton {
     }
   }
 }
+
+class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
```
gnprice, line 414: nit: order classes to match how they appear in the menu, as listed at the top of the file — that makes it a bit easier to navigate around the file

lib/widgets/actions.dart
```diff
@@ -16,101 +16,171 @@ import '../model/narrow.dart';
...
+      // Use [AnchorCode.oldest], because [AnchorCode.firstUnread]
+      // will be the oldest non-muted unread message, which would
+      // result in muted unreads older than the first unread not
+      // being processed.
+      startingAnchor: AnchorCode.oldest,
       // [AnchorCode.oldest] is an anchor ID lower than any valid
       // message ID; and follow-up requests will have already
       // processed the anchor ID, so we just want this to be
       // unconditionally false.
       includeAnchor: false,
```
gnprice, line 49: The reasoning in the second half of this comment applies just as well on markNarrowAsUnreadFromMessage, or any other caller of updateMessageFlagsStartingFromAnchor, as it does on markNarrowAsRead. It's a fact about how the loop works, rather than about the value of `startingAnchor`.

So in the loop we should make `includeAnchor` be false after the first iteration, though the first iteration will depend on the caller.

lib/widgets/actions.dart
```diff
@@ -16,101 +16,171 @@ import '../model/narrow.dart';
...
+  }
+}
+
+Future<void> markNarrowAsUnreadFromMessage(
+  BuildContext context,
+  Message message,
+  Narrow narrow,
+) async {
+  final zulipLocalizations = ZulipLocalizations.of(context);
+  await updateMessageFlagsStartingFromAnchor(
```
gnprice, line 76: This won't work if the server's feature level is below 155. That's the point of the check in `markNarrowAsRead`, right?

To handle that:
* This function can just have an assert confirming the server is new enough (by checking `store.connection.zulipFeatureLevel`).
* The action sheet shouldn't offer the button if the server isn't new enough.
* At each of those places, leave a `// TODO(server-6)` comment so we can simplify the logic away when we drop support for affected versions. For discussion, see: https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#todo-server

gnprice (commented):
OK, finished reading the branch — comments on the remaining commits below.

I'll also want to make a pass over the commit structure and comments; but that'll be relatively quick once these items are addressed.

lib/widgets/actions.dart
Khader-1: Oh.. yes that's right!

gnprice (commented):
Thanks for the revision! Just nits below, and I've also pushed a few added commits on the end:
463e27010 actions [nfc]: Simplify includeAnchor logic slightly in update-flags loop
4f6283ee4 actions [nfc]: Simplify name of anchor parameter to update-flags
fab52eff8 actions [nfc]: Revise doc comment for updateMessageFlagsStartingFromAnchor

After fixing those nits, this will be ready to merge.

Khader-1:

Khader-1:

test/widgets/actions_test.dart
```diff
@@ -278,19 +201,133 @@ void main() {
...
+            'num_before': '0',
+            'num_after': '1000',
+            'narrow': jsonEncode(apiNarrow),
+            'op': 'add',
+            'flag': 'read',
+          });
+      check(await didPass).isTrue();
+    });
+
+    testWidgets('pagination', (WidgetTester tester) async {
```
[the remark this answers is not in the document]
gnprice: bump on this part:
> Similarly this previous commit can be squashed in:
> https://github.com/zulip/zulip-flutter/commit/c4fee15cf7373b140e3c13abec138b0746a9f330 actions_test [nfc]: Create updateMessageFlagsStartingFromAnchor test group

gnprice (commented):
Thanks @Khader-1 for the revision! Two comments below.

I think these will both be fastest for me to just fix up, so I'll do that and merge. But please do take a look through the material about handling Git merge/rebase conflicts and about `git range-diff`.

Issue #633, opened 2024-04-24
Generate narrow links with "channel", not "stream", where supported


gnprice:
This is the zulip-flutter counterpart to the following zulip-mobile issue:
* https://github.com/zulip/zulip-mobile/issues/5861

We should do the same thing here.

Like in zulip-mobile, it won't yet be time to do this until a zulip-mobile release with https://github.com/zulip/zulip-mobile/issues/5860 (to recognize the same "channel" operator) has been out to users for some weeks.

Related server/web issues:
* Umbrella issue: https://github.com/zulip/zulip/issues/28468
* Specific corresponding issue: https://github.com/zulip/zulip/issues/30385

gnprice:
In order for this to really work properly, it'll need:
* #458 

Without that, for a user who's been using this app with e.g. chat.zulip.org since before the relevant feature level, and hasn't reinstalled the app since then, the app will continue to think the server has an older version and so it won't activate this feature.

gnprice:
The server and web app won't be generating these links for several more months, for Server 10 rather than Server 9, in order to leave time for mobile users who don't get prompt updates to update to versions that support the new form of URL. See [chat thread](https://chat.zulip.org/#narrow/stream/378-api-design/topic/URLs.20for.20stream.20to.20channel.20rename/near/1848218).

PR #781, opened 2024-07-01
internal_link: Generate narrow links with "channel", not "stream"

sm-sayedi:
Fixes: #633

### DIFF
--- a/lib/api/model/narrow.dart
+++ b/lib/api/model/narrow.dart
@@ -35,7 +35,7 @@ sealed class ApiNarrowElement {
 }
 
 class ApiNarrowStream extends ApiNarrowElement {
-  @override String get operator => 'stream';
+  @override String get operator => 'channel';
 
   @override final int operand;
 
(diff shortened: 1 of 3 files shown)


Khader-1 (approved):
Thanks! LGTM.
Let's move on to the mentor review with @hackerkid

chrisbobbe:
Thanks @sm-sayedi! This needs a feature-level condition, as mentioned in #633 and the zulip-mobile issue linked from there.

sm-sayedi:
> This needs a feature-level condition, as mentioned in #633 and the zulip-mobile issue linked from there.

Does it mean we should wait for some time or a condition should be used in the code?

gnprice:
> Does it mean we should wait for some time or a condition should be used in the code?

Both :slightly_smiling_face: It'll need a feature-level condition so we generate these only when the server is new enough that its web app understands these links; and we'll also want to wait for some time, so that mobile users who don't get prompt updates (the majority do, but many don't) reach a version that understands these links.

That wait just became clearer a few days ago, after [an `#api design` thread](https://chat.zulip.org/#narrow/stream/378-api-design/topic/URLs.20for.20stream.20to.20channel.20rename/near/1846142) prompted by the question of when to start generating these on web. I've updated #633 accordingly, and pushed it to a later milestone.

Because it'll be a few months before we can deploy an implementation of this, let's pause work on it for now, and I'll close this PR. The PR will remain linked from the issue, so we can always draw on it when we do pick up the issue in the future.

PR #782, opened 2024-07-01
content: Parse code block `data-code-language` attribute

chrisbobbe:
We don't currently need this -- I wrote it when I thought proper styling depended on knowing whether the code block was declared with a language or not -- but it took some thinking-through to get just right, so might as well do it now in case we want it later (e.g., to include the language attribute in Flutter Semantics for code blocks).

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -287,9 +287,20 @@ class SpoilerNode extends BlockContentNode {
 }
 
 class CodeBlockNode extends BlockContentNode {
-  const CodeBlockNode(this.spans, {super.debugHtmlNode});
+  const CodeBlockNode(
+    this.spans, {
+    this.codeLanguage,
+    super.debugHtmlNode,
+  });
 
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks for posting this!

What were the pitfalls that came up before you thought it through more, in order to get it just right?

Because content parsing is a performance-sensitive area, I'd like to wait to add any given field or additional parsing logic until we have an immediate use case for it. So I'm closing this PR, but then the closed PR will remain as a demonstration of how we could add this field if that does come up in the future.

Then please feel free to add a comment line like:
```dart
  // final String? codeLanguage; // for an implementation, see https://github.com/zulip/zulip-flutter/pull/782
```
so that this PR will be readily findable in that future.

chrisbobbe:
> What were the pitfalls that came up before you thought it through more, in order to get it just right?

No pitfalls, just the process of realizing/remembering how little extra code was required to test the parsing. :) That I didn't even have to write additional tests to get coverage.

Issue #665, opened 2024-05-13
Show "typing" status


gnprice:
I.e., half of this feature:
https://zulip.com/help/typing-notifications
along with #666.

In the API, the relevant data appears as events:
https://zulip.com/api/get-events#typing-start
https://zulip.com/api/get-events#typing-stop

In the code, let's organize the data under a class `TypingStatus`, similar to `Unreads` or `RecentDmConversationsView`.

For logic for comparison, see what's in zulip-mobile or Zulip web. The implementation in zulip-mobile isn't great in how it handles expiry, though (in `typingActions.js`) — in particular we shouldn't need a polling loop. I haven't studied the web implementation.

PR #783, opened 2024-07-01
model: Add data models for typing events.

PIG208:
Still mostly a work in progress.

This should partially address #665

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -707,6 +707,9 @@ class DeleteMessageEvent extends Event {
 
   final List<int> messageIds;
   // final int messageId; // Not present; we support the bulk_message_deletion capability
+  // The server never actually sends "direct" here yet (it's "private" instead),
+  // but we accept both forms for forward-compatibility.
+  @MessageTypeConverter()
   final MessageType messageType;
   final int? streamId;
   final String? topic;
@@ -735,10 +738,25 @@ class DeleteMessageEvent extends Event {
 
 /// As in [DeleteMessageEvent.messageType]
 /// or [UpdateMessageFlagsMessageDetail.type].
-@JsonEnum(fieldRename: FieldRename.snake)
+@JsonEnum(alwaysCreate: true)
 enum MessageType {
   stream,
-  private;
+  direct;
+}
+
+class MessageTypeConverter extends JsonConverter<MessageType, String> {
+  const MessageTypeConverter();
+
+  @override
+  MessageType fromJson(String json) {
+    if (json == 'private') json = 'direct'; // TODO(server-future)
+    return $enumDecode(_$MessageTypeEnumMap, json);
+  }
+
+  @override
+  String toJson(MessageType object) {
+    return _$MessageTypeEnumMap[object]!;
+  }
 }
 
 /// A Zulip event of type `update_message_flags`.
@@ -820,6 +838,9 @@ class UpdateMessageFlagsRemoveEvent extends UpdateMessageFlagsEvent {
 /// As in [UpdateMessageFlagsRemoveEvent.messageDetails].
 @JsonSerializable(fieldRename: FieldRename.snake)
 class UpdateMessageFlagsMessageDetail {
+  // The server never actually sends "direct" here yet (it's "private" instead),
+  // but we accept both forms for forward-compatibility.
+  @MessageTypeConverter()
   final MessageType type;
   final bool? mentioned;
   final List<int>? userIds;
@@ -841,7 +862,7 @@ class UpdateMessageFlagsMessageDetail {
       case MessageType.stream:
         result.streamId as int;
         result.topic as String;
-      case MessageType.private:
+      case MessageType.direct:
         result.userIds as List<int>;
     }
     return result;
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -62,6 +62,7 @@ sealed class Event {
           case 'remove': return UpdateMessageFlagsRemoveEvent.fromJson(json);
           default: return UnexpectedEvent.fromJson(json);
         }
+      case 'typing': return TypingEvent.fromJson(json);
       case 'reaction': return ReactionEvent.fromJson(json);
       case 'heartbeat': return HeartbeatEvent.fromJson(json);
       // TODO add many more event types
@@ -736,8 +737,9 @@ class DeleteMessageEvent extends Event {
   Map<String, dynamic> toJson() => _$DeleteMessageEventToJson(this);
 }
 
-/// As in [DeleteMessageEvent.messageType]
-/// or [UpdateMessageFlagsMessageDetail.type].
+/// As in [DeleteMessageEvent.messageType],
+/// [UpdateMessageFlagsMessageDetail.type],
+/// or [TypingEvent.messageType].
 @JsonEnum(alwaysCreate: true)
 enum MessageType {
   stream,
@@ -869,6 +871,69 @@ class UpdateMessageFlagsMessageDetail {
   }
 
   Map<String, dynamic> toJson() => _$UpdateMessageFlagsMessageDetailToJson(this);
+}
+
+/// A Zulip event of type `typing`:
+///   https://zulip.com/api/get-events#typing-start
+///   https://zulip.com/api/get-events#typing-stop
+@JsonSerializable(fieldRename: FieldRename.snake)
+class TypingEvent extends Event {
+  @override
+  @JsonKey(includeToJson: true)
+  String get type => 'typing';
+
+  final TypingOp op;
+  @MessageTypeConverter()
+  final MessageType messageType;
+  @JsonKey(readValue: _readSenderId)
+  final int senderId;
+  @JsonKey(name: 'recipients', fromJson: _recipientIdsFromJson)
+  final List<int>? recipientIds;
+  final int? streamId;
+  final String? topic;
+
+  TypingEvent({
+    required super.id,
+    required this.op,
+    required this.messageType,
+    required this.senderId,
+    required this.recipientIds,
+    required this.streamId,
+    required this.topic,
+  });
+
+  static Object? _readSenderId(Map<Object?, Object?> json, String key) {
+    return (json['sender'] as Map<String, dynamic>)['user_id'];
+  }
+
+  static List<int>? _recipientIdsFromJson(Object? json) {
+    if (json == null) return null;
+    return (json as List<Object?>).map(
+      (item) => (item as Map<String, Object?>)['user_id'] as int).toList();
+  }
+
+  factory TypingEvent.fromJson(Map<String, dynamic> json) {
+    final result = _$TypingEventFromJson(json);
+    // Crunchy-shell validation
+    switch (result.messageType) {
+      case MessageType.stream:
+        result.streamId as int;
+        result.topic as String;
+      case MessageType.direct:
+        result.recipientIds as List<int>;
+    }
+    return result;
+  }
+
+  @override
+  Map<String, dynamic> toJson() => _$TypingEventToJson(this);
+}
+
+/// As in [TypingEvent.op].
+@JsonEnum(fieldRename: FieldRename.snake)
+enum TypingOp {
+  start,
+  stop
 }
 
 /// A Zulip event of type `reaction`, with op `add` or `remove`.
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -24,6 +24,14 @@ class InitialSnapshot {
 
   final List<CustomProfileField> customProfileFields;
 
+  // TODO(server-8): Remove the default values.
+  @JsonKey(defaultValue: 15000)
+  final int serverTypingStartedExpiryPeriodMilliseconds;
+  @JsonKey(defaultValue: 5000)
+  final int serverTypingStoppedWaitPeriodMilliseconds;
+  @JsonKey(defaultValue: 10000)
+  final int serverTypingStartedWaitPeriodMilliseconds;
+
   // final List<…> mutedTopics; // TODO(#422) we ignore this feature on older servers
 
   final Map<String, RealmEmojiItem> realmEmoji;
@@ -86,6 +94,9 @@ class InitialSnapshot {
     required this.zulipMergeBase,
     required this.alertWords,
     required this.customProfileFields,
+    required this.serverTypingStartedExpiryPeriodMilliseconds,
+    required this.serverTypingStoppedWaitPeriodMilliseconds,
+    required this.serverTypingStartedWaitPeriodMilliseconds,
     required this.realmEmoji,
     required this.recentPrivateConversations,
     required this.subscriptions,
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -24,6 +24,7 @@ import 'message.dart';
 import 'message_list.dart';
 import 'recent_dm_conversations.dart';
 import 'stream.dart';
+import 'typing_status.dart';
 import 'unreads.dart';
 
 export 'package:drift/drift.dart' show Value;
@@ -242,6 +243,10 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
         .followedBy(initialSnapshot.realmNonActiveUsers)
         .followedBy(initialSnapshot.crossRealmBots)
         .map((user) => MapEntry(user.userId, user))),
+      typingStatus: TypingStatus(
+        selfUserId: account.userId,
+        typingStartedExpiryPeriod: Duration(milliseconds: initialSnapshot.serverTypingStartedExpiryPeriodMilliseconds),
+      ),
       streams: streams,
       messages: MessageStoreImpl(),
       unreads: Unreads(
@@ -266,6 +271,7 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
     required this.selfUserId,
     required this.userSettings,
     required this.users,
+    required this.typingStatus,
     required StreamStoreImpl streams,
     required MessageStoreImpl messages,
     required this.unreads,
@@ -318,6 +324,8 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
   // Users and data about them.
 
   final Map<int, User> users;
+
+  final TypingStatus typingStatus;
 
   ////////////////////////////////
   // Streams, topics, and stuff about them.
@@ -383,6 +391,7 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
     recentDmConversationsView.dispose();
     unreads.dispose();
     _messages.dispose();
+    typingStatus.dispose();
     super.dispose();
   }
 
@@ -485,6 +494,9 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
       assert(debugLog("server event: update_message_flags/${event.op} ${event.flag.toJson()}"));
       _messages.handleUpdateMessageFlagsEvent(event);
       unreads.handleUpdateMessageFlagsEvent(event);
+    } else if (event is TypingEvent) {
+      assert(debugLog("server event: typing/${event.op} ${event.messageType}"));
+      typingStatus.handleTypingEvent(event);
     } else if (event is ReactionEvent) {
       assert(debugLog("server event: reaction/${event.op}"));
       _messages.handleReactionEvent(event);
--- a/lib/model/typing_status.dart
+++ b/lib/model/typing_status.dart
@@ -1,0 +1,83 @@
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+
+import '../api/model/events.dart';
+import 'narrow.dart';
+
+/// The model for tracking the typing status organized by narrows.
+///
+/// Listeners are notified when a typist is added or removed from any narrow.
+class TypingStatus extends ChangeNotifier {
+  TypingStatus({
+    required this.selfUserId,
+    required this.typingStartedExpiryPeriod,
+  });
+
+  final int selfUserId;
+  final Duration typingStartedExpiryPeriod;
+
+  Iterable<SendableNarrow> get debugActiveNarrows => _timerMapsByNarrow.keys;
+
+  Iterable<int> typistIdsInNarrow(SendableNarrow narrow) =>
+    _timerMapsByNarrow[narrow]?.keys ?? [];
+
+  // Using SendableNarrow as the key covers the narrows
+  // where typing notifications are supported (topics and DMs).
+  final Map<SendableNarrow, Map<int, Timer>> _timerMapsByNarrow = {};
+
+  @override
+  void dispose() {
+    for (final timersByTypistId in _timerMapsByNarrow.values) {
+      for (final timer in timersByTypistId.values) {
+        timer.cancel();
+      }
+    }
+    _timerMapsByNarrow.clear();
+    super.dispose();
+  }
+
+  bool _addTypist(SendableNarrow narrow, int typistUserId) {
+    final narrowTimerMap = _timerMapsByNarrow[narrow] ??= {};
+    final typistTimer = narrowTimerMap[typistUserId];
+    final isNewTypist = typistTimer == null;
+    typistTimer?.cancel();
+    narrowTimerMap[typistUserId] = Timer(typingStartedExpiryPeriod, () {
+      if (_removeTypist(narrow, typistUserId)) {
+        notifyListeners();
+      }
+    });
+    return isNewTypist;
+  }
+
+  bool _removeTypist(SendableNarrow narrow, int typistUserId) {
+    final narrowTimerMap = _timerMapsByNarrow[narrow];
+    final typistTimer = narrowTimerMap?.remove(typistUserId);
+    if (typistTimer == null) {
+      return false;
+    }
+    typistTimer.cancel();
+    if (narrowTimerMap!.isEmpty) _timerMapsByNarrow.remove(narrow);
+    return true;
+  }
+
+  void handleTypingEvent(TypingEvent event) {
+    SendableNarrow narrow = switch (event.messageType) {
+      MessageType.direct => DmNarrow(
+        allRecipientIds: event.recipientIds!..sort(), selfUserId: selfUserId),
+      MessageType.stream => TopicNarrow(event.streamId!, event.topic!),
+    };
+
+    bool hasUpdate = false;
+    switch (event.op) {
+      case TypingOp.start:
+        hasUpdate = _addTypist(narrow, event.senderId);
+      case TypingOp.stop:
+        hasUpdate = _removeTypist(narrow, event.senderId);
+    }
+
+    if (hasUpdate) {
+      notifyListeners();
+    }
+  }
+}
--- a/test/api/model/events_test.dart
+++ b/test/api/model/events_test.dart
@@ -153,7 +153,16 @@ void main() {
     })).throws<void>();
   });
 
-  test('update_message_flags/remove: require messageDetails in mark-as-unread', () {
+  test('delete_message: private -> direct', () {
+    check(DeleteMessageEvent.fromJson({
+      'id': 1,
+      'type': 'delete_message',
+      'message_ids': [1, 2, 3],
+      'message_type': 'private',
+    })).messageType.equals(MessageType.direct);
+  });
+
+  group('update_message_flags/remove', () {
     final baseJson = {
       'id': 1,
       'type': 'update_message_flags',
@@ -162,14 +171,30 @@ void main() {
       'messages': [123],
       'all': false,
     };
-    check(() => UpdateMessageFlagsRemoveEvent.fromJson(baseJson)).returnsNormally();
-    check(() => UpdateMessageFlagsRemoveEvent.fromJson({
-      ...baseJson, 'flag': 'read',
-    })).throws<void>();
-    check(() => UpdateMessageFlagsRemoveEvent.fromJson({
-      ...baseJson,
-      'flag': 'read',
-      'message_details': {'123': {'type': 'private', 'mentioned': false, 'user_ids': [2]}},
-    })).returnsNormally();
+    final messageDetail = {'type': 'direct', 'mentioned': false, 'user_ids': [2]};
+
+    test('require messageDetails in mark-as-unread', () {
+      check(() => UpdateMessageFlagsRemoveEvent.fromJson(baseJson)).returnsNormally();
+      check(() => UpdateMessageFlagsRemoveEvent.fromJson({
+        ...baseJson, 'flag': 'read',
+      })).throws<void>();
+      check(() => UpdateMessageFlagsRemoveEvent.fromJson({
+        ...baseJson,
+        'flag': 'read',
+        'message_details': {'123': messageDetail},
+      })).returnsNormally();
+    });
+
+    test('private -> direct', () {
+      check(UpdateMessageFlagsRemoveEvent.fromJson({
+        ...baseJson,
+        'flag': 'read',
+        'message_details': {
+          '123': {
+            ...messageDetail,
+            'type': 'private',
+          }}})).messageDetails.isNotNull()
+               .values.single.type.equals(MessageType.direct);
+    });
   });
 }
--- a/test/api/model/events_test.dart
+++ b/test/api/model/events_test.dart
@@ -197,4 +197,49 @@ void main() {
                .values.single.type.equals(MessageType.direct);
     });
   });
+
+  group('typing status event', () {
+    final baseJson = {
+      'id': 1,
+      'type': 'typing',
+      'op': 'start',
+      'sender': {'user_id': 123, 'email': '123@example.com'},
+    };
+
+    final directMessageJson = {
+      ...baseJson,
+      'message_type': 'direct',
+      'recipients': [1, 2, 3].map((e) => {'user_id': e, 'email': '$e@example.com'}).toList(),
+    };
+
+    test('direct message typing events', () {
+      check(TypingEvent.fromJson(directMessageJson))
+        ..recipientIds.isNotNull().deepEquals([1, 2, 3])
+        ..senderId.equals(123);
+    });
+
+    test('private type missing recipient', () {
+      check(() => TypingEvent.fromJson({
+        ...baseJson, 'message_type': 'private'})).throws<void>();
+    });
+
+    test('private -> direct', () {
+      check(TypingEvent.fromJson({
+        ...directMessageJson,
+        'message_type': 'private',
+      })).messageType.equals(MessageType.direct);
+    });
+
+    test('stream type missing streamId/topic', () {
+      check(() => TypingEvent.fromJson({
+        ...baseJson, 'message_type': 'stream', 'stream_id': 123, 'topic': 'foo'}))
+        .returnsNormally();
+      check(() => TypingEvent.fromJson({
+        ...baseJson, 'message_type': 'stream'})).throws<void>();
+      check(() => TypingEvent.fromJson({
+        ...baseJson, 'message_type': 'stream', 'topic': 'foo'})).throws<void>();
+      check(() => TypingEvent.fromJson({
+        ...baseJson, 'message_type': 'stream', 'stream_id': 123})).throws<void>();
+    });
+  });
 }
--- a/test/model/typing_status_test.dart
+++ b/test/model/typing_status_test.dart
@@ -1,0 +1,221 @@
+import 'package:checks/checks.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:zulip/api/model/events.dart';
+import 'package:zulip/api/model/model.dart';
+import 'package:zulip/model/narrow.dart';
+import 'package:zulip/model/typing_status.dart';
+
+import '../example_data.dart' as eg;
+import '../fake_async.dart';
+
+void main() {
+  late TypingStatus model;
+  late int notifiedCount;
+
+  void checkNotNotified() {
+    check(notifiedCount).equals(0);
+  }
+
+  void checkNotifiedOnce() {
+    check(notifiedCount).equals(1);
+    notifiedCount = 0;
+  }
+
+  void handleTypingEvent(SendableNarrow narrow, TypingOp op, User sender) {
+    assert(sender != eg.selfUser);
+    model.handleTypingEvent(eg.typingEvent(narrow, op, sender.userId));
+  }
+
+  void checkTypists(Map<SendableNarrow, List<User>> typistsByNarrow) {
+    final actualTypistsByNarrow = <SendableNarrow, Iterable<int>>{};
+    for (final narrow in model.debugActiveNarrows) {
+      actualTypistsByNarrow[narrow] = model.typistIdsInNarrow(narrow);
+    }
+    check(actualTypistsByNarrow).deepEquals(
+      typistsByNarrow.map((k, v) => MapEntry(k, v.map((e) => e.userId))));
+  }
+
+  void prepareModel({
+    int? selfUserId,
+    Map<SendableNarrow, List<User>> typistsByNarrow = const {},
+  }) {
+    model = TypingStatus(
+      selfUserId: selfUserId ?? eg.selfUser.userId,
+      typingStartedExpiryPeriod: const Duration(milliseconds: 15000));
+    check(model.debugActiveNarrows).isEmpty();
+    notifiedCount = 0;
+    model.addListener(() => notifiedCount += 1);
+
+    typistsByNarrow.forEach((narrow, typists) {
+      for (final typist in typists) {
+        handleTypingEvent(narrow, TypingOp.start, typist);
+        checkNotifiedOnce();
+      }
+    });
+    checkTypists(typistsByNarrow);
+  }
+
+  final stream = eg.stream();
+  final topicNarrow = TopicNarrow(stream.streamId, 'foo');
+
+  final dmNarrow = DmNarrow.withUser(eg.otherUser.userId, selfUserId: eg.selfUser.userId);
+  final groupNarrow = DmNarrow.withOtherUsers(
+    [eg.otherUser.userId, eg.thirdUser.userId], selfUserId: eg.selfUser.userId);
+
+  test('dispose', () => awaitFakeAsync((async) async {
+    prepareModel(typistsByNarrow: {dmNarrow: [eg.otherUser]});
+
+    model.dispose();
+    checkTypists({});
+    check(async.pendingTimers).isEmpty();
+  }));
+
+  group('handle typing start events', () {
+    test('add typists in separate narrows', () {
+      prepareModel();
+
+      handleTypingEvent(dmNarrow, TypingOp.start, eg.otherUser);
+      checkTypists({dmNarrow: [eg.otherUser]});
+      checkNotifiedOnce();
+
+      handleTypingEvent(groupNarrow, TypingOp.start, eg.thirdUser);
+      checkTypists({dmNarrow: [eg.otherUser], groupNarrow: [eg.thirdUser]});
+      checkNotifiedOnce();
+
+      handleTypingEvent(topicNarrow, TypingOp.start, eg.fourthUser);
+      checkTypists({
+        dmNarrow: [eg.otherUser],
+        groupNarrow: [eg.thirdUser],
+        topicNarrow: [eg.fourthUser]});
+      checkNotifiedOnce();
+    });
+
+    test('add a typist in the same narrow', () {
+      prepareModel(typistsByNarrow: {groupNarrow: [eg.otherUser]});
+
+      handleTypingEvent(groupNarrow, TypingOp.start, eg.thirdUser);
+      checkTypists({groupNarrow: [eg.otherUser, eg.thirdUser]});
+      checkNotifiedOnce();
+    });
+
+    test('sort dm recipients', () {
+      prepareModel(selfUserId: 5);
+      final recipientIds = [5, 4, 10, 8, 2, 1];
+
+      final eventUnsorted = TypingEvent(id: 1, op: TypingOp.start,
+        senderId: 1,
+        messageType: MessageType.direct,
+        recipientIds: recipientIds,
+        streamId: null, topic: null);
+      // DmNarrow's constructor expects the recipient IDs to be sorted,
+      // and [model.handleTypingEvent] should handle that.
+      model.handleTypingEvent(eventUnsorted);
+      check(model.typistIdsInNarrow(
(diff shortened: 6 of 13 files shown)


lib/api/model/initial_snapshot.dart
```diff
@@ -86,6 +90,10 @@ class InitialSnapshot {
     required this.zulipMergeBase,
     required this.alertWords,
     required this.customProfileFields,
+    // TODO(server-8): Remove the default values.
+    this.serverTypingStartedExpiryPeriodMilliseconds = 15000,
```
gnprice, line 94: Good thought.

Let's use `JsonKey.defaultValue` to express this instead, though:
https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/defaultValue.html

That way these fallbacks are a bit more clearly demarcated as part of the processing of the JSON data, rather than being part of the constructor for the class.

lib/api/model/events.dart
```diff
@@ -832,6 +834,71 @@ class UpdateMessageFlagsMessageDetail {
   Map<String, dynamic> toJson() => _$UpdateMessageFlagsMessageDetailToJson(this);
 }
 
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class TypingStatusEvent extends Event {
```
gnprice, line 839: Let's keep the event class names totally boring and predictable just from the API's `type` value: so `TypingEvent`.

lib/api/model/events.dart
```diff
@@ -832,6 +834,71 @@ class UpdateMessageFlagsMessageDetail {
...
+/// As in [TypingStatusEvent.op].
+@JsonEnum(fieldRename: FieldRename.snake)
+enum TypingStatusOp {
+  start,
+  stop
+}
+
+/// As in [TypingStatusEvent.sender] and [TypingStatusEvent.recipients].
+@JsonSerializable(fieldRename: FieldRename.snake)
+class Typist {
```
gnprice, line 887: This name "Typist" fits for the sender, but not for the recipients.

Even if it were just the sender, the disadvantage of this kind of name is that it doesn't clearly tie the class back to these events. This is basically a single-use-site helper for this event class, so it's good for the name to reflect that so it doesn't sound like something potentially more general.

A good boring name would then be `TypingEventUser`.

lib/api/model/events.dart
```diff
@@ -832,6 +834,71 @@ class UpdateMessageFlagsMessageDetail {
...
+@JsonSerializable(fieldRename: FieldRename.snake)
+class TypingStatusEvent extends Event {
+  @override
+  @JsonKey(includeToJson: true)
+  String get type => 'typing';
+
+  final TypingStatusOp op;
+  final MessageType messageType;
+  final Typist sender;
+  final List<Typist>? recipients;
```
gnprice, line 847: Actually, though, instead of having this `Typist` (or `TypingEventUser`) class at all, I think we can simplify by just dropping the emails and making these an `int` and a `List<int>`. Any time you see a "Zulip API email address" mentioned in the docs, that's basically an obsolescent relic of the days when Zulip users were primarily identified by their email address, rather than their immutable numeric user ID. There might still be a few corners of the API where we need to use them; but our first choice should always be to ignore them and see if we can do everything with just the user IDs.

For an example of a similar simplification, see `SubscriptionRemoveEvent.streamIds`.

gnprice (commented):
Neat! Here's a couple of high-level comments on modeling the API.

lib/api/model/initial_snapshot.dart
PIG208: This originally used `defaultValue`, but that would add repetition to the example data. In that sense we might want to define the fallbacks as `static const`s for re-use purposes.

lib/api/model/events.dart
PIG208: Good thought! This appears to be a good case to use `readValue` for.

lib/api/model/initial_snapshot.dart
gnprice: I think it's fine for the example data to say the value again — in fact it's preferable for it to do so.

One angle on that: after all, once we discharge that `TODO(server-8)` comment, the example data will need to specific the value for itself because there won't be a fallback value in the live code.

Abstracting a bit: although it's not entirely a coincidence that the value we fall back to for old servers, and the boring value we want to use in tests where this property isn't intended to be interesting, are equal… it didn't have to be the case that they're equal, and they are distinct concepts. So the example data should say the value it wants as the boring example value, while the deserialization in the live code says the data it wants to fall back to for old servers.

PIG208:
Updated the PR to support the view-model for typing status. This should be ready for review now. I will follow up with the UI change in a separate PR.

PIG208:
Because `MessageType` expects either "stream" or "private", modern servers that send "direct" in the typing events are not supported unless we fix #785.

lib/api/model/events.dart
```diff
@@ -850,6 +852,68 @@ class UpdateMessageFlagsMessageDetail {
...
+@JsonSerializable(fieldRename: FieldRename.snake)
+class TypingEvent extends Event {
+  @override
+  @JsonKey(includeToJson: true)
+  String get type => 'typing';
+
+  final TypingOp op;
+  final MessageType messageType;
+  @JsonKey(name: 'sender', readValue: _readSenderId)
+  final int senderId;
```
chrisbobbe, line 868: Let's have `_readSenderId` do all of the work of reading the value, instead of sharing it with `name: 'sender'`. That way the reader doesn't have to go back and forth between the two places to see how the work gets done.

Like how in `SubscriptionRemoveEvent`, even though `streamIds` is computed from information at a `subscriptions` key, it doesn't use `name: 'subscriptions'`, and instead the string 'subscriptions' is used in the `_readStreamIds` implementation:

<details>

```dart
/// A [SubscriptionEvent] with op `remove`: https://zulip.com/api/get-events#subscription-remove
class SubscriptionRemoveEvent extends SubscriptionEvent {
  @override
  @JsonKey(includeToJson: true)
  String get op => 'remove';

  @JsonKey(readValue: _readStreamIds)
  final List<int> streamIds;

  static List<int> _readStreamIds(Map<dynamic, dynamic> json, String key) {
    return (json['subscriptions'] as List<dynamic>)
      .map((e) => (e as Map<String, dynamic>)['stream_id'] as int)
      .toList();
  }

  SubscriptionRemoveEvent({required super.id, required this.streamIds});

  factory SubscriptionRemoveEvent.fromJson(Map<String, dynamic> json) =>
    _$SubscriptionRemoveEventFromJson(json);

  @override
  Map<String, dynamic> toJson() => _$SubscriptionRemoveEventToJson(this);
}
```
</details>

test/api/model/events_test.dart
```diff
@@ -172,4 +172,53 @@ void main() {
...
+    };
+
+    test('direct message typing events', () {
+      final directMessageJson = {
+        ...baseJson,
+        'message_type': 'private',
+        'recipients': [1, 2, 3].map((e) => {'user_id': e, 'email': '$e@example.com'}).toList(),
+      };
+      check(TypingEvent.fromJson(directMessageJson))
+        ..recipientIds.isNotNull().deepEquals([1,2,3])
```
chrisbobbe, line 191: nit: `[1, 2, 3]`

lib/model/store.dart
```diff
@@ -361,6 +369,8 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
 
   final RecentDmConversationsView recentDmConversationsView;
 
+  final TypingStatus typingStatus;
```
chrisbobbe, line 372: Hmm. I'm not sure this fits in the category "Messages, and summaries of messages". Maybe it's a better fit for "Users and data about them" and could go there?

lib/model/typing_status.dart
```diff
@@ -0,0 +1,83 @@
...
+  TypingStatus({
+    required this.selfUserId,
+    required this.typingStartedExpiryPeriod,
+    required this.typingStoppedWaitPeriod,
+    required this.typingStartedWaitPeriod,
+  });
+
+  @override
+  void dispose() {
+    for (final typistTimer in typistIdsByNarrow.values) {
```
chrisbobbe, line 27: `typistTimer` doesn't quite fit as a name for a `Map<int, Timer>`. Maybe `timersForNarrow`?

lib/model/typing_status.dart
```diff
@@ -0,0 +1,83 @@
...
+
+/// The view-model for tracking the typing status organized by narrows.
+class TypingStatus extends ChangeNotifier {
+  @visibleForTesting
+  final Map<SendableNarrow, Map<int, Timer>> typistIdsByNarrow = {};
+
+  final int selfUserId;
+  final Duration typingStartedExpiryPeriod;
+  final Duration typingStoppedWaitPeriod;
+  final Duration typingStartedWaitPeriod;
```
chrisbobbe, line 16: I wonder if `typingStoppedWaitPeriod` and `typingStartedWaitPeriod` belong in this class, which is about reading and tracking users' typing status. I see that they're unused in this PR. These two values are about _sending_ the self-user's typing status, which seems maybe different enough that we'll end up using them in a separate subsystem in a different class.

lib/model/typing_status.dart
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  bool _addTypist(SendableNarrow narrow, int typistUserId, void Function() callback) {
+    final narrowTimerMap = typistIdsByNarrow[narrow] ??= {};
+    final typistTimer = narrowTimerMap[typistUserId];
+    final isNewTypist = typistTimer == null;
+    typistTimer?.cancel();
+    narrowTimerMap[typistUserId] = Timer(typingStartedExpiryPeriod, callback);
+    return isNewTypist;
+  }
```
chrisbobbe, line 42: The `callback` can just be inlined in here, can't it? It doesn't close over anything that we can't access here, right:

```suggestion
  bool _addTypist(SendableNarrow narrow, int typistUserId) {
    final narrowTimerMap = typistIdsByNarrow[narrow] ??= {};
    final typistTimer = narrowTimerMap[typistUserId];
    final isNewTypist = typistTimer == null;
    typistTimer?.cancel();
    narrowTimerMap[typistUserId] = Timer(typingStartedExpiryPeriod, () {
      if (_removeTypist(narrow, typistUserId)) {
        notifyListeners();
      }
    });
    return isNewTypist;
  }
```

chrisbobbe: Oh, and since this `notifyListeners` happens after a delay, let's put it behind a check that the instance hasn't been `dispose`d yet. I'm not sure if there's a better way to check that than just `hasListeners`.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+extension TypingStatusChecks on Subject<TypingStatus> {
+  Subject<Map<SendableNarrow, Map<int, Timer>>> get typistsByLocation => has((x) => x.typistIdsByNarrow, 'typistsByLocation');
+}
+
+extension TimerChecks on Subject<Timer> {
+  Subject<bool> get isActive => has((x) => x.isActive, 'isActive');
+}
+
+
+void main() {
```
chrisbobbe, line 21: nit: extra empty line

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+  late TypingStatus model;
+  late int notifiedCount;
+
+  void prepareModel() {
+    model = TypingStatus(
+        selfUserId: eg.selfUser.userId,
+        typingStartedExpiryPeriod: const Duration(milliseconds: 15000),
+        typingStoppedWaitPeriod: const Duration(milliseconds: 5000),
+        typingStartedWaitPeriod: const Duration(milliseconds: 10000));
```
chrisbobbe, line 30: nit: two-space indent

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:zulip/api/model/events.dart';
+import 'package:zulip/model/narrow.dart';
+import 'package:zulip/model/typing_status.dart';
+
+import '../example_data.dart' as eg;
+import '../fake_async.dart';
+
+extension TypingStatusChecks on Subject<TypingStatus> {
+  Subject<Map<SendableNarrow, Map<int, Timer>>> get typistsByLocation => has((x) => x.typistIdsByNarrow, 'typistsByLocation');
```
chrisbobbe, line 13: There's a naming mismatch here.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+          streamId: null,
+          topic: null);
+    }
+  }
+
+  final stream = eg.stream();
+  final topicNarrow = TopicNarrow(stream.streamId, 'foo');
+
+  final otherUser = eg.otherUser;
+  final thirdUser = eg.thirdUser;
```
chrisbobbe, line 66: nit: Not much is saved by having these variables, right; `otherUser` is only three characters shorter than `eg.otherUser`.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+        groupNarrow: {otherUser.userId: anyTimer}});
+
+      model.handleTypingEvent(
+        mkTypingEvent(dmNarrow, TypingOp.stop, otherUser.userId));
+      check(model).typistsByLocation.deepEquals({
+        groupNarrow: {otherUser.userId: anyTimer}});
+      checkNotNotified();
+    }));
+
+    test('remove another typist in a narrow', () => awaitFakeAsync((async) async {
```
chrisbobbe, line 147: I'm not sure I understand; would the meaning change if this test were called 'remove a typist in a narrow' instead?

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+        mkTypingEvent(groupNarrow, TypingOp.start, otherUser.userId));
+      var oldTimer = check(model).typistsByLocation[groupNarrow][otherUser.userId]
+        ..isActive.isTrue();
+      checkNotifiedOnce();
+      model.handleTypingEvent(
+        mkTypingEvent(groupNarrow, TypingOp.start, otherUser.userId));
+      oldTimer.isActive.isFalse();
+      checkNotNotified();
+    }));
+  });
```
chrisbobbe, line 234: This checks that the old timer was canceled. It seems like it would be helpful to also check that a new timer was started.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,278 @@
...
+    [otherUser.userId, thirdUser.userId], selfUserId: eg.selfUser.userId);
+
+  anyTimer(Subject<Object?> it) => it.isA<Timer>();
+
+  group('handle typing start events', () {
+    test('add typists in separate narrows', () {
+      prepareModel();
+
+      model.handleTypingEvent(
+        mkTypingEvent(dmNarrow, TypingOp.start, otherUser.userId));
```
chrisbobbe, line 78: There's a lot of repetition of

```dart
      model.handleTypingEvent(
        mkTypingEvent(
```

, which always takes at least two lines. Perhaps worth making a helper that lets us do the same thing in one line?

chrisbobbe (commented):
Exciting, thanks! Small comments below.

lib/model/store.dart
gnprice: Yeah, neither is a perfect fit.

I like the "users" area a bit better, because that's definitely where presence data will naturally go, and this is a lot like presence.

test/model/typing_status_test.dart
PIG208: This is specifically for testing the case when there are multiple typists but we only want to remove one. Renaming this to "remove one of two typists in the same narrow" might be better.

lib/model/typing_status.dart
PIG208: Hmm, good point. Let's leave them out then.

lib/model/typing_status.dart
PIG208: We can also track it with a `bool`, but `hasListeners` should be enough.

lib/api/model/events.dart
PIG208: I wonder if the `as int` part is necessary for `_readSenderId` because the code generator handles that. Otherwise this refactor should be good.

chrisbobbe: I agree it looks like it's not necessary.

chrisbobbe:
Thanks! LGTM; marking for Greg's review now.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+
+      checkTyping(dmNarrow, TypingOp.start, eg.otherUser)
+        .typistsByNarrow.deepEquals(
+          {dmNarrow: {eg.otherUser.userId: anyTimer}});
+      checkNotifiedOnce();
+
+      checkTyping(groupNarrow, TypingOp.start, eg.thirdUser)
+        .typistsByNarrow.deepEquals({
+          dmNarrow: {eg.otherUser.userId: anyTimer},
+          groupNarrow: {eg.thirdUser.userId: anyTimer}});
```
gnprice, line 69: This is a bit confusing to read because the `checkTyping` function turns out to have a side effect, mutating the model, so e.g. the behavior at this point would be different if the other `checkTyping` call above were left out.

Better to make the side effects explicit, by separating the `model.handleTypingEvent` call from the `check(model)`:

```suggestion
      model.handleTypingEvent(eg.typingEvent(groupNarrow, TypingOp.start, eg.thirdUser.userId));
      check(model).typistsByNarrow.deepEquals({
        dmNarrow: {eg.otherUser.userId: anyTimer},
        groupNarrow: {eg.thirdUser.userId: anyTimer}});
```

Then that first line could still benefit from a helper function to reduce the boilerplate; but it can be focused just on that step, and as a result also it can have a name to match, without "check".

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  final stream = eg.stream();
+  final topicNarrow = TopicNarrow(stream.streamId, 'foo');
+
+  final dmNarrow = DmNarrow.withUser(eg.otherUser.userId, selfUserId: eg.selfUser.userId);
+  final groupNarrow = DmNarrow.withOtherUsers(
+    [eg.otherUser.userId, eg.thirdUser.userId], selfUserId: eg.selfUser.userId);
+
+  anyTimer(Subject<Object?> it) => it.isA<Timer>();
```
gnprice, line 55: Function definitions should have explicit return types, so that they don't look like function calls:
```suggestion
  void anyTimer(Subject<Object?> it) => it.isA<Timer>();
```

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+          dmNarrow: {eg.otherUser.userId: anyTimer},
+          groupNarrow: {eg.thirdUser.userId: anyTimer},
+          topicNarrow: {eg.selfUser.userId: anyTimer}});
+      checkNotifiedOnce();
+    });
+
+    test('add a typist in the same narrow', () {
+      prepareModel();
+
+      checkTyping(groupNarrow, TypingOp.start, eg.selfUser)
```
gnprice, line 83: Hmm, I don't think we expect to ever see a typing event for the self-user — I think that'd probably represent a bug in the server. So let's avoid that case in test data.

(Unless we were to have a test where that edge case is the point of the test, though I don't think there's a reason we'd need that. The point is really just that test data should be boring except in the aspects that are the point of the test; and having a typing event for the self-user isn't boring, because it seems like it shouldn't happen.)

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+      prepareModel();
+
+      checkTyping(dmNarrow, TypingOp.start, eg.otherUser);
+      checkNotifiedOnce();
+      checkTyping(groupNarrow, TypingOp.start, eg.thirdUser);
+      checkNotifiedOnce();
+      checkTyping(topicNarrow, TypingOp.start, eg.selfUser);
+      checkNotifiedOnce();
+
+      async.elapse(const Duration(seconds: 5));
```
gnprice, line 147: Does this have a purpose in the story this test is telling? It doesn't seem like it does.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+      checkTyping(dmNarrow, TypingOp.start, eg.otherUser);
+      final oldTimer = check(model).typistsByNarrow[dmNarrow][eg.otherUser.userId]
+        ..isActive.isTrue();
+      checkNotifiedOnce();
+
+      checkTyping(dmNarrow, TypingOp.stop, eg.otherUser)
+        .typistsByNarrow.isEmpty();
+      oldTimer.isActive.isFalse();
+      checkNotifiedOnce();
+    }));
```
gnprice, line 182: Here's a bit more comprehensive of a check:
```suggestion
      checkNotifiedOnce();
      check(async.pendingTimers).isEmpty();
    }));
```

In fact I probably should adjust `awaitFakeAsync` so that it automatically checks that `async.pendingTimers` is empty when the callback (i.e. the test body) completes. If a test accidentally leaks timers, that's a fairly likely way of learning about a bug in the code under test.

That also avoids the need to keep a `Subject<Timer>` around, in that variable `oldTimer`. Doing that (as opposed to immediately invoking a series of methods on the `Subject` and then discarding it) is something I don't think is conventional, and I'm not entirely sure how it behaves — `Subject` is a bit opaque that way.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+      async.elapse(const Duration(seconds: 5));
+      check(model).typistsByNarrow.deepEquals({
+        dmNarrow: {eg.otherUser.userId: anyTimer},
+        groupNarrow: {eg.thirdUser.userId: anyTimer},
+        topicNarrow: {eg.selfUser.userId: anyTimer}});
+
+      checkTyping(groupNarrow, TypingOp.stop, eg.thirdUser)
+        .typistsByNarrow.deepEquals({
+          dmNarrow: {eg.otherUser.userId: anyTimer},
+          topicNarrow: {eg.selfUser.userId: anyTimer}});
```
gnprice, line 156: One thing about this style of writing the checks is that it feels like it's making these tests a bit too closely tied to the exact data structure that `TypingStatus` uses internally. In particular, to the fact that there are all these timers there.

Relatedly, the fact that these are timers isn't actually relevant to these tests — they never do anything with the timers. (There are other tests lower in the file that test the timer behavior.) So that detail is only a distraction in these tests.

I think one thing that would clean this up further is to let the main Map data structure be private, and expose something like `Iterable<SendableNarrow> get debugActiveNarrows => typistIdsByNarrow.keys`. Then the test code can combine that debug getter with `getTypistIdsInNarrow` to get a map from narrow to iterable of sender IDs. And these checks in the individual test cases can look like:
```suggestion
      checkTypists({
          dmNarrow: [eg.otherUser],
          topicNarrow: [eg.selfUser]});
```

Because there'd be a test helper involved, it can also take care of mapping user objects to user IDs, further simplifying the individual test cases.

gnprice (commented):

Comments below. For this round I haven't read all of the last commit's test cases in detail, because there are some higher-level comments on the tests that I think will change how many of them look.

lib/model/typing_status.dart
```diff
@@ -0,0 +1,81 @@
...
+
+  bool _addTypist(SendableNarrow narrow, int typistUserId) {
+    final narrowTimerMap = typistIdsByNarrow[narrow] ??= {};
+    final typistTimer = narrowTimerMap[typistUserId];
+    final isNewTypist = typistTimer == null;
+    typistTimer?.cancel();
+    narrowTimerMap[typistUserId] = Timer(typingStartedExpiryPeriod, () {
+      if (_removeTypist(narrow, typistUserId)) {
+        if (hasListeners) {
+          notifyListeners();
```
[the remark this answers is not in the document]
chrisbobbe: Ah, yeah, I guess that makes sense. My comment https://github.com/zulip/zulip-flutter/pull/783#discussion_r1667206325 didn't come from noticing anything actually going wrong, I just thought about the pattern where we check `context.mounted` in UI code.

lib/api/model/events.dart
```diff
@@ -850,6 +852,68 @@ class UpdateMessageFlagsMessageDetail {
...
+///   https://zulip.com/api/get-events#typing-start
+///   https://zulip.com/api/get-events#typing-stop
+@JsonSerializable(fieldRename: FieldRename.snake)
+class TypingEvent extends Event {
+  @override
+  @JsonKey(includeToJson: true)
+  String get type => 'typing';
+
+  final TypingOp op;
+  final MessageType messageType;
```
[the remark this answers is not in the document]
PIG208: > OTOH if the most convenient way to add the logic winds up being separate for each place the enum is used, then this PR might as well stick to this one spot where it's actually needed, and we can handle the others later.

I put the change that affected other events in a prep commit: https://github.com/zulip/zulip-flutter/pull/783/commits/9410969ed75edd1154007811e4278d32e2d2db6b
This might not be the cleanest way to do it, but all we need is annotating the field with the converter.

lib/api/model/events.dart
```diff
@@ -850,6 +852,68 @@ class UpdateMessageFlagsMessageDetail {
...
+    required super.id,
+    required this.op,
+    required this.messageType,
+    required this.senderId,
+    required this.recipientIds,
+    required this.streamId,
+    required this.topic,
+  });
+
+  static dynamic _readSenderId(Map<dynamic, dynamic> json, String key) {
```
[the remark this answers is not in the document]
PIG208: I didn't go with the `int` alternative so the generator take care of that.

lib/model/typing_status.dart
```diff
@@ -0,0 +1,81 @@
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+
+import '../api/model/events.dart';
+import 'narrow.dart';
+
+/// The view-model for tracking the typing status organized by narrows.
+class TypingStatus extends ChangeNotifier {
```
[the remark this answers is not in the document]
PIG208: Thanks. That makes sense.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,240 @@
...
+
+import '../example_data.dart' as eg;
+import '../fake_async.dart';
+
+extension TypingStatusChecks on Subject<TypingStatus> {
+  Subject<Map<SendableNarrow, Map<int, Timer>>> get typistsByNarrow => has((x) => x.typistIdsByNarrow, 'typistsByNarrow');
+}
+
+extension TimerChecks on Subject<Timer> {
+  Subject<bool> get isActive => has((x) => x.isActive, 'isActive');
```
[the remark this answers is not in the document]
PIG208: Removed it as we don't interact with timers directly in the tests anymore.

PIG208:
Thanks for the review @gnprice ! This is ready for another one now.

lib/model/typing_status.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  final int selfUserId;
+  final Duration typingStartedExpiryPeriod;
+
+  Iterable<SendableNarrow> get debugActiveNarrows => _timerMapsByNarrow.keys;
+
+  Iterable<int> typistIdsInNarrow(SendableNarrow narrow) =>
+    _timerMapsByNarrow[narrow]?.keys ?? [];
+
+  final Map<SendableNarrow, Map<int, Timer>> _timerMapsByNarrow = {};
```
gnprice, line 23: Using SendableNarrow as the key covers the narrows where typing
    notifications are supported (topics and dms).

This is good information for a comment in the tree (and then it can be omitted from the commit message). Rationale here:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#commit-messages-vs-comments

So e.g.:
```suggestion
  // Using SendableNarrow as the key covers the narrows
  // where typing notifications are supported (topics and DMs).
  final Map<SendableNarrow, Map<int, Timer>> _timerMapsByNarrow = {};
```

(Note double-slash, not triple-slash, because that doesn't have the content of a doc comment.)

lib/model/typing_status.dart
```diff
@@ -0,0 +1,78 @@
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+
+import '../api/model/events.dart';
+import 'narrow.dart';
+
+/// The model for tracking the typing status organized by narrows.
+class TypingStatus extends ChangeNotifier {
```
gnprice, line 9: Consumers of the typing
    status will only be notified if a typist has been added or removed from
    any of the narrows.

Similarly to my comment just above (or a bit below if reading this in the Files tab) about:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#commit-messages-vs-comments

In this case the information is something one should know when consuming this class, so it's good to put in its doc:

```suggestion
/// The model for tracking the typing status organized by narrows.
///
/// Listeners are notified when a typist is added or removed from any narrow.
class TypingStatus extends ChangeNotifier {
```

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+
+  final dmNarrow = DmNarrow.withUser(eg.otherUser.userId, selfUserId: eg.selfUser.userId);
+  final groupNarrow = DmNarrow.withOtherUsers(
+    [eg.otherUser.userId, eg.thirdUser.userId], selfUserId: eg.selfUser.userId);
+
+  group('handle typing start events', () {
+    test('add typists in separate narrows', () {
+      prepareModel();
+
+      model.handleTypingEvent(eg.typingEvent(dmNarrow, TypingOp.start, eg.otherUser.userId));
```
gnprice, line 53: A lot of these `model.handleTypingEvent` lines have key information — the identity of the typist — beyond column 80.

So let's shorten with a little helper, as I alluded to at the end of https://github.com/zulip/zulip-flutter/pull/783#discussion_r1671621238 . Can look like:
```suggestion
      handleTypingEvent(dmNarrow, TypingOp.start, eg.otherUser);
```

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+      checkTypists({groupNarrow: [eg.otherUser, eg.thirdUser]});
+      checkNotifiedOnce();
+    });
+
+    test('sort dm recipients', () {
+      prepareModel();
+
+      final eventUnsorted = TypingEvent(id: 1, op: TypingOp.start, senderId: eg.otherUser.userId,
+        messageType: MessageType.direct,
+        recipientIds: [5, 4, 10, 8, 2, 1, eg.selfUser.userId],
```
gnprice, line 79: This doesn't satisfy the promises in the API:
https://zulip.com/api/get-events#typing-start
because the sender isn't among the recipients.

Probably most convenient to just use a number instead of `eg.otherUser.userId`.

You could also make it all explicit numbers, by giving `prepareModel` an optional parameter `selfUserId` and using that for this test case.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+        streamId: null,
+        topic: null);
+      // DmNarrow's constructor expects the recipient IDs to be sorted.
+      check(() => model.handleTypingEvent(eventUnsorted)).returnsNormally();
```
gnprice, line 83: Let's add a check that it has a typist at the expected DmNarrow. That's always a bit more reassuring than just "didn't throw" — if the code under test didn't throw, then maybe that's only because it hit an error but decided that meant it should do nothing, and it might not have even reached the bit of code the test was intended to exercise.

(These `returnsNormally` checks appear in some existing test cases you saw in the events tests, as baselines for comparison with `throws` checks. They work OK there, because their job is only to provide a contrast with the `throws` checks.)

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+      model.handleTypingEvent(eg.typingEvent(dmNarrow, TypingOp.start, eg.otherUser.userId));
+      checkNotifiedOnce();
+      model.handleTypingEvent(eg.typingEvent(groupNarrow, TypingOp.start, eg.thirdUser.userId));
+      checkNotifiedOnce();
+      model.handleTypingEvent(eg.typingEvent(topicNarrow, TypingOp.start, eg.fourthUser.userId));
+      checkNotifiedOnce();
+      checkTypists({
+          dmNarrow: [eg.otherUser],
+          groupNarrow: [eg.thirdUser],
+          topicNarrow: [eg.fourthUser]});
```
gnprice, line 127: nit: two-space indent
```suggestion
      checkTypists({
        dmNarrow: [eg.otherUser],
        groupNarrow: [eg.thirdUser],
        topicNarrow: [eg.fourthUser]});
```

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+    });
+
+    test('remove one of two typists in the same narrow', () async {
+      prepareModel();
+
+      model.handleTypingEvent(eg.typingEvent(groupNarrow, TypingOp.start, eg.otherUser.userId));
+      checkNotifiedOnce();
+      model.handleTypingEvent(eg.typingEvent(groupNarrow, TypingOp.start, eg.thirdUser.userId));
+      checkNotifiedOnce();
+      checkTypists({groupNarrow: [eg.otherUser, eg.thirdUser]});
```
gnprice, line 108: Most of these test cases have to start by adding some typists, when that part is effectively part of the setup of the test and isn't the thing the test is about.

So one helpful step for making the tests more focused on the main story they're trying to tell would be to have a shared helper take care of that, so that the test cases just specify what state they want. Here, that can probably just be an optional parameter on `prepareModel`. So:
```suggestion
      prepareModel(typists: {groupNarrow: [eg.otherUser, eg.thirdUser]});
```

and then the implementation of `prepareModel` can make the `handleTypingEvent` calls, check notified, and even use `checkTypists` to double-check that the state came out as intended.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+      model.handleTypingEvent(eg.typingEvent(groupNarrow, TypingOp.start, eg.otherUser.userId));
+      checkTypists({groupNarrow: [eg.otherUser]});
+      checkNotifiedOnce();
+
+      model.handleTypingEvent(eg.typingEvent(dmNarrow, TypingOp.stop, eg.otherUser.userId));
+      checkTypists({groupNarrow: [eg.otherUser]});
+      checkNotNotified();
+    });
+
+    test('remove one of two typists in the same narrow', () async {
```
gnprice, line 101: Let's also have a test for removing the last typist in a narrow, checking that we don't leak an empty list for that narrow into our map.

… On rereading, I see you do cover that case at the end of "remove typists from different narrows". But it'd be good to have a separate test case just about that case, with a description to match. Otherwise it's easy to miss that that's there, as I did. Also if some future change were to accidentally break this behavior, if all that did was cause that last step of a larger test case to fail, it'd be easy to think it's just cleanup code and so miss that it's checking something meaningful, defeating the purpose of the test.

This is also one of the benefits of tightening the setup further, with something like `prepareModel(typists: …)` like in my immediate previous comment on the code just below here. When setup is concise, it's cheap to make more separate test cases and there isn't much reason to be tempted to combine conceptually different questions into a single test case.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+      checkTypists({dmNarrow: [eg.otherUser]});
+      checkNotNotified();
+
+      async.elapse(const Duration(seconds: 5));
+      checkTypists({});
+      checkNotifiedOnce();
+    }));
+
+
+    test('typist is removed when the expiry period ends', () => awaitFakeAsync((async) async {
```
gnprice, line 197: nit: put this before the other expiry test case — the narrative makes most sense when this one is read first, because the other is basically this plus an extra wrinkle

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,207 @@
...
+      final eventUnsorted = TypingEvent(id: 1, op: TypingOp.start, senderId: eg.otherUser.userId,
+        messageType: MessageType.direct,
+        recipientIds: [5, 4, 10, 8, 2, 1, eg.selfUser.userId],
+        streamId: null,
+        topic: null);
+      // DmNarrow's constructor expects the recipient IDs to be sorted.
+      check(() => model.handleTypingEvent(eventUnsorted)).returnsNormally();
+    });
+
+  });
```
gnprice, line 86: nit: stray blank line at end of block

There's also a double blank line later in the file.

gnprice (commented):

I've now read all the test cases. Some of the comments in this round should make the test cases more compact and so easier to scan through as a whole, so in a future round I'll reread them with an eye to thinking of any other cases that should be added.

lib/model/typing_status.dart
PIG208: Also updated the commit message explaining that this design was derived from the web implementation, which probably counts as some "why the new version is better than, or not worse than, the old version" information.

test/model/typing_status_test.dart
PIG208: Modified the helper to get that work. Thanks!

test/model/typing_status_test.dart
PIG208: Yeah. I also found having to initialize the `DmNarrow` is informative in terms of where the sorted requirement comes from.

PIG208:
The PR has been updated to address the comments. Thanks for the review!

lib/model/typing_status.dart
gnprice: Agreed, yeah that's helpful.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      checkTypists({dmNarrow: [eg.otherUser], groupNarrow: [eg.thirdUser]});
+      checkNotifiedOnce();
+    });
+
+    test('add a typist in the same narrow', () {
+      prepareModel();
+
+      handleTypingEvent(groupNarrow, TypingOp.start, eg.otherUser);
+      checkTypists({groupNarrow: [eg.otherUser]});
+      checkNotifiedOnce();
```
gnprice, line 83: This can also benefit from `prepareModel(typistsByNarrow: …)`.

(previous discussion: https://github.com/zulip/zulip-flutter/pull/783#discussion_r1673060182 )

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+    });
+  });
+
+  group('handle typing stop events', () {
+    test('remove a typist from an unknown narrow', () {
+      prepareModel();
+
+      handleTypingEvent(groupNarrow, TypingOp.start, eg.otherUser);
+      checkTypists({groupNarrow: [eg.otherUser]});
+      checkNotifiedOnce();
```
gnprice, line 114: and this

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+    });
+  });
+
+  group('cancelling old timer', () {
+    test('when typing stopped early', () => awaitFakeAsync((async) async {
+      prepareModel();
+
+      handleTypingEvent(dmNarrow, TypingOp.start, eg.otherUser);
+      checkTypists({dmNarrow: [eg.otherUser]});
+      checkNotifiedOnce();
```
gnprice, line 166: and this

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      check(async.pendingTimers).isEmpty();
+    }));
+
+    test('when typing repeatedly started', () => awaitFakeAsync((async) async {
+      prepareModel();
+
+      handleTypingEvent(groupNarrow, TypingOp.start, eg.otherUser);
+      checkTypists({groupNarrow: [eg.otherUser]});
+      checkNotifiedOnce();
```
gnprice, line 180: and this

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      check(async.pendingTimers).length.equals(1);
+      checkNotNotified();
+    }));
+  });
+
+  group('typing start expiry period', () {
+    test('typist is removed when the expiry period ends', () => awaitFakeAsync((async) async {
+      prepareModel();
+
+      handleTypingEvent(dmNarrow, TypingOp.start, eg.otherUser);
```
gnprice, line 195: and this

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      checkTypists({dmNarrow: [eg.otherUser]});
+      async.elapse(const Duration(seconds: 10));
+      checkTypists({});
+    }));
+
+    test('repeated typing start event resets the timer', () => awaitFakeAsync((async) async {
+      prepareModel();
+
+      handleTypingEvent(dmNarrow, TypingOp.start, eg.otherUser);
+      checkNotifiedOnce();
```
gnprice, line 206: and this

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      // and [model.handleTypingEvent] should handle that.
+      model.handleTypingEvent(eventUnsorted);
+      check(model.typistIdsInNarrow(
+        DmNarrow(allRecipientIds: recipientIds..sort(), selfUserId: 5),
+      )).single.equals(1);
+    });
+  });
+
+  group('handle typing stop events', () {
+    test('remove a typist from an unknown narrow', () {
```
gnprice, line 109: Another case to exercise is: remove an unknown typist from a known narrow

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  final stream = eg.stream();
+  final topicNarrow = TopicNarrow(stream.streamId, 'foo');
+
+  final dmNarrow = DmNarrow.withUser(eg.otherUser.userId, selfUserId: eg.selfUser.userId);
+  final groupNarrow = DmNarrow.withOtherUsers(
+    [eg.otherUser.userId, eg.thirdUser.userId], selfUserId: eg.selfUser.userId);
+
+  group('handle typing start events', () {
```
gnprice, line 65: Another case to have an explicit test case for: a start event in a topic narrow.

(This gets covered by the "remove typists from different narrows" case; but because it's not the point of the test, it'd be easy for that to go away in a future revision of the tests and then that coverage would get dropped.)

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      checkNotifiedOnce();
+
+      handleTypingEvent(topicNarrow, TypingOp.stop, eg.fourthUser);
+      checkTypists({});
+      checkNotifiedOnce();
+    });
+  });
+
+  group('cancelling old timer', () {
+    test('when typing stopped early', () => awaitFakeAsync((async) async {
```
gnprice, line 161: This group doesn't feel to me like the logical way to organize these two tests. Both test cases are good ones to have; but this one, let's put under the "stop events" group, with name "cancel timer".

The other case below, let's put under the "start events" group, with name "cancel existing timer".

In particular, these don't feel to me like two cases of one idea — it's more like two separate pieces of logic, that are part of what's expected for a stop event and a start event respectively.

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+  group('typing start expiry period', () {
+    test('typist is removed when the expiry period ends', () => awaitFakeAsync((async) async {
+      prepareModel();
+
+      handleTypingEvent(dmNarrow, TypingOp.start, eg.otherUser);
+      async.elapse(const Duration(seconds: 5));
+      checkTypists({dmNarrow: [eg.otherUser]});
+      async.elapse(const Duration(seconds: 10));
+      checkTypists({});
+    }));
```
gnprice, line 200: should also `checkNotifiedOnce` here

and for good measure, `checkNotNotified` at the 5-second mark

test/model/typing_status_test.dart
gnprice: Or perhaps better: these two could be organized together with the cases currently under "expiry period" below, as a group that's about the timers.

Then the narrative, test case by test case, could be:
* start sets a timer — look, just wait 15 seconds and the typist gets removed (the current "typist is removed when the expiry period ends")
* stop cancels the timer (the current "when typing stopped early")
* repeated start resets the timer — look, there's just one timer pending, and you have to wait 15 seconds before the typist gets removed, even if it'd already been 10 seconds since the original start (the current "repeated typing start event resets the timer", plus the `pendingTimers` checks from the current "when typing repeatedly started")

I think that makes a pretty clean story.

test/model/typing_status_test.dart
gnprice: Also at the end check pendingTimers is empty. (No need for checking in the middle that it's nonempty ­— the fact that `async.elapse` causes a spontaneous change is enough to make that point.)

test/model/typing_status_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      async.elapse(const Duration(seconds: 10));
+      checkTypists({dmNarrow: [eg.otherUser]});
+      checkNotNotified();
+
+      async.elapse(const Duration(seconds: 5));
+      checkTypists({});
+      checkNotifiedOnce();
+    }));
+  });
+}
```
gnprice, line 223: One more test case to add: test `dispose`.

lib/model/typing_status.dart
```diff
@@ -0,0 +1,82 @@
...
+  final Map<SendableNarrow, Map<int, Timer>> _timerMapsByNarrow = {};
+
+  @override
+  void dispose() {
+    for (final timersByTypistId in _timerMapsByNarrow.values) {
+      for (final timer in timersByTypistId.values) {
+        timer.cancel();
+      }
+    }
+    super.dispose();
```
gnprice, line 36:
```suggestion
  void dispose() {
    for (final timersByTypistId in _timerMapsByNarrow.values) {
      for (final timer in timersByTypistId.values) {
        timer.cancel();
      }
    }
    _timerMapsByNarrow.clear();
    super.dispose();
```

In general there isn't a need for `dispose` methods to clear maps and such just for the sake of clearing them; it's the garbage collector's job to discard them when the object is truly discarded.

But in this case, since we're disposing the timers, let's also clear the map, because that way this maintains the basic invariant of this data structure: all the timers in the map are live, and all the live timers this class created are in the map.

gnprice (commented):
Thanks for the revision! The changes look good. Comments below, including thinking through other test cases to make the tests complete.

test/model/typing_status_test.dart
PIG208: A related update: for the revised "repeated typing start event resets the timer"  test, I kept some `pendingTimers` checks. It probably makes more sense there because we don't want duplicated timers.

PIG208:
Thanks for the review! I have updated the PR to address the comments.

test/model/typing_status_test.dart
PIG208: Updated all tests other than "add typists in separate narrows" and "sort dm recipients" to use this.

gnprice:
Thanks for the revision! This all looks great — merging.

Issue #291, opened 2023-08-31
profile: Display email


sirpengi:
In the profile screen (#195 ) we need to show the user's email address if it is accessible, but not the fake email that Zulip creates for the api.

Also need to handle the user update event for `null` values of `delivery_email` appropriately.

Helpful links:
- Search for "delivery_email:" in https://zulip.com/api/get-events#realm_user-update
- `getDisplayEmailForUser` in zulip-mobile: https://github.com/zulip/zulip-mobile/blob/8133bf2b8/src/users/userSelectors.js#L318

MrunalrajRedij:
Hey, If it is not already fixed. Can I work on this issue?

gnprice:
@MrunalrajRedij Please take a look at our guide for picking an issue to work on:
https://github.com/zulip/zulip-flutter#picking-an-issue-to-work-on

gnprice:
> Also need to handle the user update event for `null` values of `delivery_email` appropriately.

This is accomplished by #F784. So once that's merged, `User.deliveryEmail` can be used normally.

PR #784, opened 2024-07-02
api: JsonNullable; handle absent-vs-null RealmUserUpdateEvent.deliveryEmail

gnprice:
This introduces `JsonNullable` to handle cases where the Zulip API expresses a distinction by having some property be null vs. absent entirely. It's based on the draft I wrote in a chat conversation last year:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20omitted.20vs.2E.20null.20in.20JSON/near/1551714

Then we use that for RealmUserUpdateEvent.deliveryEmail; which causes the former User.deliveryEmailStaleDoNotUse to no longer be stale, and we can call it simply `User.deliveryEmail`. That will be helpful for #291 (fyi @sm-sayedi).

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 
+import 'json.dart';
 import 'model.dart';
 
 part 'events.g.dart';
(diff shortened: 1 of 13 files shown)


chrisbobbe:
Awesome, thanks for doing this! I'd been wondering how this was going to look. LGTM, merged.

PR #787, opened 2024-07-04
Handle moved messages in MessageStore and message lists

gnprice:
Fixes most of #150. (After this lands we should do a quick scan to identify anything else that needs updating on message moves — there are several new data structures in the store that we've merged in recent weeks.)

A couple of months ago I spent a few hours writing part of an implementation of #150. It's mostly complete, apart from the lack of tests… and any bugs it has that will be found by writing those tests.

@PIG208 since you're picking up #150 shortly, I spent a small amount of time just now cleaning up that draft branch and rebasing atop the changes from #750, and writing down what remains to be done to finish it. So please start from this, which will let you avoid duplicating that work.

I believe there are three remaining things this needs:

* Tests. As usual, these should go in the same commit that makes the changes.

  There's a variety of kinds of test cases this needs,
  but here's one in particular that may not be apparent:

  If the `generation += 1` line is commented out, the message list
  has a race bug where a fetchOlder starts; we reset (because
  messages were moved into the narrow); and then the fetch returns
  and appends in the wrong spot.  Also a related race bug where it's
  fetchInitial, and it lacks the moved messages.  So there should be
  test cases exercising each of those situations, which would fail if
  that line gets commented out.

* Update StreamMessage.displayRecipient.  This goes in a followup commit.

    If we wanted, we could update it by looking up the new stream's
    name.  To enable MessageStoreImpl to do that, we'd pass it a
    StreamStore, the same way we do with Unreads.

    But in fact the only time displayRecipient is useful is when we
    don't have data on that stream.  So there's not much point in
    looking it up here.  Instead, make the field nullable, and just
    set it to null when the message is moved between streams.

* Handle UpdateMessageEvent.propagateMode.  This also goes in a followup commit. If it gets complex (or if the rest of the branch before it gets complex), let's split it out as a followup PR.

    This probably means
    _MessageListPageState having its own notion of the current narrow,
    which is initialized with `widget.narrow` but can change.
    Then `_MessageListState._modelChanged` can check if the model's
    narrow has changed, and if so tell the page state to update.

    For comparison, see zulip-mobile's `src/events/doEventActionSideEffects.js`,
    or web.

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -576,7 +576,11 @@ class StreamMessage extends Message {
   @JsonKey(includeToJson: true)
   String get type => 'stream';
 
-  final String displayRecipient;
+  // This is not nullable API-wise, but if the message moves across channels,
+  // [displayRecipient] still refers to the original channel and it has to be
+  // invalidated.
+  @JsonKey(required: true, disallowNullValue: true)
+  String? displayRecipient;
   final int streamId;
 
   StreamMessage({
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -480,7 +480,7 @@ sealed class Message {
   final int senderId;
   final String senderRealmStr;
   @JsonKey(name: 'subject')
-  final String topic;
+  String topic;
   // final List<string> submessages; // TODO handle
   final int timestamp;
   String get type;
@@ -581,7 +581,7 @@ class StreamMessage extends Message {
   // invalidated.
   @JsonKey(required: true, disallowNullValue: true)
   String? displayRecipient;
-  final int streamId;
+  int streamId;
 
   StreamMessage({
     required super.client,
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -179,22 +179,43 @@ class MessageStoreImpl with MessageStore {
       return;
     }
 
-    if (newStreamId == null
-        && MessageEditState.topicMoveWasResolveOrUnresolve(origTopic, newTopic!)) {
-      // The topic was only resolved/unresolved.
-      // No change to the messages' editState.
-      return;
-    }
+    final wasResolveOrUnresolve = (newStreamId == null
+      && MessageEditState.topicMoveWasResolveOrUnresolve(origTopic, newTopic!));
 
-    // TODO(#150): Handle message moves.  The views' recipient headers
-    //   may need updating, and consequently showSender too.
-    //   Currently only editState gets updated.
     for (final messageId in event.messageIds) {
       final message = messages[messageId];
       if (message == null) continue;
-      // Do not override the edited marker if the message has also been moved.
-      if (message.editState == MessageEditState.edited) continue;
-      message.editState = MessageEditState.moved;
+
+      if (message is! StreamMessage) {
+        assert(debugLog('Bad UpdateMessageEvent: stream/topic move on a DM')); // TODO(log)
+        continue;
+      }
+
+      if (newStreamId != null) {
+        message.streamId = newStreamId;
+        // See [StreamMessage.displayRecipient] on why the invalidation is
+        // needed.
+        message.displayRecipient = null;
+      }
+
+      if (newTopic != null) {
+        message.topic = newTopic;
+      }
+
+      if (!wasResolveOrUnresolve
+          && message.editState == MessageEditState.none) {
+        message.editState = MessageEditState.moved;
+      }
+    }
+
+    for (final view in _messageListViews) {
+      view.messagesMoved(
+        origStreamId: origStreamId,
+        newStreamId: newStreamId ?? origStreamId,
+        origTopic: origTopic,
+        newTopic: newTopic ?? origTopic,
+        messageIds: event.messageIds,
+      );
     }
   }
 
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -152,6 +152,7 @@ class MessageStoreImpl with MessageStore {
     final newStreamId = event.newStreamId; // null if topic-only move
     final origTopic = event.origTopic;
     final newTopic = event.newTopic;
+    final propagateMode = event.propagateMode;
 
     if (origTopic == null) {
       // There was no move.
@@ -176,6 +177,11 @@ class MessageStoreImpl with MessageStore {
     if (origStreamId == null) {
       // The `stream_id` field (aka origStreamId) is documented to be present on moves.
       assert(debugLog('Malformed UpdateMessageEvent: move but no origStreamId')); // TODO(log)
+      return;
+    }
+    if (propagateMode == null) {
+      // The `propagate_mode` field (aka propagateMode) is documented to be present on moves.
+      assert(debugLog('Malformed UpdateMessageEvent: move but no propagateMode')); // TODO(log)
       return;
     }
 
@@ -215,6 +221,7 @@ class MessageStoreImpl with MessageStore {
         origTopic: origTopic,
         newTopic: newTopic ?? origTopic,
         messageIds: event.messageIds,
+        propagateMode: propagateMode,
       );
     }
   }
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -167,9 +167,10 @@ class MessageStoreImpl with MessageStore {
       return;
     }
 
-    if (newTopic == null) {
-      // The `subject` field (aka newTopic) is documented to be present on moves.
-      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic')); // TODO(log)
+    if (newStreamId == null && newTopic == null) {
+      // If neither the channel nor topic name changed, nothing moved.
+      // In that case `orig_subject` (aka origTopic) should have been null.
+      assert(debugLog('Malformed UpdateMessageEvent: move but no newStreamId or newTopic')); // TODO(log)
       return;
     }
     if (origStreamId == null) {
@@ -179,7 +180,7 @@ class MessageStoreImpl with MessageStore {
     }
 
     if (newStreamId == null
-        && MessageEditState.topicMoveWasResolveOrUnresolve(origTopic, newTopic)) {
+        && MessageEditState.topicMoveWasResolveOrUnresolve(origTopic, newTopic!)) {
       // The topic was only resolved/unresolved.
       // No change to the messages' editState.
       return;
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -359,7 +359,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   }
 
   final PerAccountStore store;
-  final Narrow narrow;
+  Narrow narrow;
 
   /// Whether [message] should actually appear in this message list,
   /// given that it does belong to the narrow.
@@ -535,12 +535,24 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     }
   }
 
+  void _handlePropagateMode(PropagateMode propagateMode, Narrow newNarrow) {
+    switch (propagateMode) {
+      case PropagateMode.changeAll:
+      case PropagateMode.changeLater:
+        narrow = newNarrow;
+        _reset();
+        fetchInitial();
+      case PropagateMode.changeOne:
+    }
+  }
+
   void messagesMoved({
     required int origStreamId,
     required int newStreamId,
     required String origTopic,
     required String newTopic,
     required List<int> messageIds,
+    required PropagateMode propagateMode,
   }) {
     switch (narrow) {
       case DmNarrow():
@@ -571,7 +583,9 @@ class MessageListView with ChangeNotifier, _MessageSequence {
           case (false, false): return;
           case (true,  true ): return; // TODO(log) no-op move
           case (false, true ): _messagesMovedIntoNarrow();
-          case (true,  false): _messagesMovedFromNarrow(messageIds); // TODO handle propagateMode
+          case (true,  false):
+            _messagesMovedFromNarrow(messageIds);
+            _handlePropagateMode(propagateMode, TopicNarrow(newStreamId, newTopic));
         }
     }
   }
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -65,6 +65,9 @@ class MessageListHistoryStartItem extends MessageListItem {
 ///
 /// This comprises much of the guts of [MessageListView].
 mixin _MessageSequence {
+  /// A sequence number for invalidating stale fetches.
+  int generation = 0;
+
   /// The messages.
   ///
   /// See also [contents] and [items].
@@ -190,6 +193,17 @@ mixin _MessageSequence {
       (message) => parseContent(message.content)));
     assert(contents.length == messages.length);
     _reprocessAll();
+  }
+
+  /// Reset all [_MessageSequence] data, and cancel any active fetches.
+  void _reset() {
+    generation += 1;
+    messages.clear();
+    _fetched = false;
+    _haveOldest = false;
+    _fetchingOlder = false;
+    contents.clear();
+    items.clear();
   }
 
   /// Redo all computations from scratch, based on [messages].
@@ -398,12 +412,14 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     assert(!fetched && !haveOldest && !fetchingOlder);
     assert(messages.isEmpty && contents.isEmpty);
     // TODO schedule all this in another isolate
+    final generation = this.generation;
     final result = await getMessages(store.connection,
       narrow: narrow.apiEncode(),
       anchor: AnchorCode.newest,
       numBefore: kMessageListFetchBatchSize,
       numAfter: 0,
     );
+    if (this.generation > generation) return;
     store.reconcileMessages(result.messages);
     store.recentSenders.handleMessages(result.messages); // TODO(#824)
     for (final message in result.messages) {
@@ -426,6 +442,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     _fetchingOlder = true;
     _updateEndMarkers();
     notifyListeners();
+    final generation = this.generation;
     try {
       final result = await getMessages(store.connection,
         narrow: narrow.apiEncode(),
@@ -434,6 +451,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
         numBefore: kMessageListFetchBatchSize,
         numAfter: 0,
       );
+      if (this.generation > generation) return;
 
       if (result.messages.isNotEmpty
           && result.messages.last.id == messages[0].id) {
@@ -451,9 +469,11 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       _insertAllMessages(0, fetchedMessages);
       _haveOldest = result.foundOldest;
     } finally {
-      _fetchingOlder = false;
-      _updateEndMarkers();
-      notifyListeners();
+      if (this.generation == generation) {
+        _fetchingOlder = false;
+        _updateEndMarkers();
+        notifyListeners();
+      }
     }
   }
 
@@ -486,6 +506,73 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     final index = _findMessageWithId(messageId);
     if (index != -1) {
       _reparseContent(index);
+    }
+  }
+
+  void _messagesMovedInternally(List<int> messageIds) {
+    for (final messageId in messageIds) {
+      if (_findMessageWithId(messageId) != -1) {
+        _reprocessAll();
+        notifyListeners();
+        return;
+      }
+    }
+  }
+
+  void _messagesMovedIntoNarrow() {
+    // If there are some messages we don't have in [MessageStore], and they
+    // occur later than the messages we have here, then we just have to
+    // re-fetch from scratch.  That's always valid, so just do that always.
+    // TODO in cases where we do have data to do better, do better.
+    _reset();
+    notifyListeners();
+    fetchInitial();
+  }
+
+  void _messagesMovedFromNarrow(List<int> messageIds) {
+    if (_removeMessagesById(messageIds)) {
+      notifyListeners();
+    }
+  }
+
+  void messagesMoved({
+    required int origStreamId,
+    required int newStreamId,
+    required String origTopic,
+    required String newTopic,
+    required List<int> messageIds,
+  }) {
+    switch (narrow) {
+      case DmNarrow():
+        // DMs can't be moved (nor created by moves),
+        // so the messages weren't in this narrow and still aren't.
+        return;
+
+      case CombinedFeedNarrow():
+      case MentionsNarrow():
+        // The messages were and remain in this narrow.
+        // TODO(#421): … except they may have become muted or not.
+        //   We'll handle that at the same time as we handle muting itself changing.
+        // Recipient headers, and downstream of those, may change, though.
+        _messagesMovedInternally(messageIds);
+
+      case ChannelNarrow(:final streamId):
+        switch ((origStreamId == streamId, newStreamId == streamId)) {
+          case (false, false): return;
+          case (true,  true ): _messagesMovedInternally(messageIds);
+          case (false, true ): _messagesMovedIntoNarrow();
+          case (true,  false): _messagesMovedFromNarrow(messageIds);
+        }
+
+      case TopicNarrow(:final streamId, :final topic):
+        final oldMatch = (origStreamId == streamId && origTopic == topic);
+        final newMatch = (newStreamId == streamId && newTopic == topic);
+        switch ((oldMatch, newMatch)) {
+          case (false, false): return;
+          case (true,  true ): return; // TODO(log) no-op move
+          case (false, true ): _messagesMovedIntoNarrow();
+          case (true,  false): _messagesMovedFromNarrow(messageIds); // TODO handle propagateMode
+        }
     }
   }
 
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -463,26 +463,28 @@ UpdateMessageEvent updateMessageEditEvent(
   );
 }
 
-UpdateMessageEvent updateMessageMoveEvent(
-  List<Message> messages, {
+UpdateMessageEvent _updateMessageMoveEvent(
+  List<int> messageIds, {
+  required int origStreamId,
   int? newStreamId,
-  String? origTopic,
+  required String origTopic,
   String? newTopic,
   String? origContent,
   String? newContent,
+  required List<MessageFlag> flags,
 }) {
-  assert(messages.isNotEmpty);
-  final origMessage = messages[0];
-  final messageId = origMessage.id;
+  assert(newTopic != origTopic
+         || (newStreamId != null && newStreamId != origStreamId));
+  assert(messageIds.isNotEmpty);
   return UpdateMessageEvent(
     id: 0,
     userId: selfUser.userId,
     renderingOnly: false,
-    messageId: messageId,
-    messageIds: messages.map((message) => message.id).toList(),
-    flags: origMessage.flags,
+    messageId: messageIds.first,
+    messageIds: messageIds,
+    flags: flags,
     editTimestamp: 1234567890, // TODO generate timestamp
-    origStreamId: origMessage is StreamMessage ? origMessage.streamId : null,
+    origStreamId: origStreamId,
     newStreamId: newStreamId,
     propagateMode: null,
     origTopic: origTopic,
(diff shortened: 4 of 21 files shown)


PIG208:
Thanks for the notes. I think I will just push to your branch and continue using PR.

PIG208:
Added the test cases.
This is still failing because we always notifies the listener if there are affected message present, and message moves might also notify the listener. I think we want to let the caller handle notifications instead.

lib/model/message_list.dart
```diff
@@ -424,13 +440,15 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
     try {
+      final generation = this.generation;
       final result = await getMessages(store.connection,
         narrow: narrow.apiEncode(),
         anchor: NumericAnchor(messages[0].id),
         includeAnchor: false,
         numBefore: kMessageListFetchBatchSize,
         numAfter: 0,
       );
+      if (this.generation > generation) return;
```
PIG208, line 451: Not sure if we want to skip the finally block as well. At least we shouldn't need to notify the listeners again.
```
    } finally {
      _fetchingOlder = false;
      _fetchingOlder = false;
      _updateEndMarkers();
      _updateEndMarkers();
      notifyListeners();
      notifyListeners();
    }
```

gnprice:
> This is still failing because we always notifies the listener if there are affected message present, and message moves might also notify the listener.

Hmm I see, so the listeners end up getting notified twice?

I think this is probably a case where we should just be OK with the duplication. Here's what I wrote on a different PR a few weeks ago (https://github.com/zulip/zulip-flutter/pull/648#discussion_r1623018764):
> Generally I think it's OK to have a little duplication of those calls — for example if some event can cause it to be called two or three times, and it'd take a lot of refactoring or bookkeeping to prevent that and call just once, then that's OK. If some listener callback is expensive enough to make that a problem, then it should have appropriate memoization, or defer the real work until the next frame starts, or something, in order to mitigate that.
>
> But this [in that other PR] is potentially hundreds of times, so that gets me a bit concerned.

(Possibly I should put that in docs somewhere. The way I found it just now is by a Zulip search within `#mobile-github` — that's been a side benefit I've quite enjoyed from having that GitHub firehose appear in Zulip.)

lib/model/message_list.dart
gnprice: In the current revision:
```diff
     } finally {
       _fetchingOlder = false;
-      _updateEndMarkers();
-      notifyListeners();
+      if (this.generation == generation) {
+        _updateEndMarkers();
+        notifyListeners();
+      }
     }
```

The thing is that the reason there's a `notifyListeners` there, and an `_updateEndMarkers`, is as a followup to the `_fetchingOlder` update. Changing `_fetchingOlder` changes what set of end markers should be present; and when the end markers change, listeners should be notified (so that e.g. the UI can update to reflect the change).

We could get an effect like this, because `_fetchingOlder` may already be false. But the way to do that would be with an explicit check on `_fetchingOlder`: no _updateEndMarkers etc. if we're not actually changing the value.

In fact, this actually points to a bug (that was introduced in my draft): if we abort because the generation advanced, then we should *not* set `_fetchingOlder` to false. It might happen to be true by the time we reach this point, because the user might have scrolled up again after the new `fetchInitial` completed, causing a new `fetchOlder`. If it is true after the generation advanced, then the fact that it's true doesn't belong to this `fetchOlder` call, but instead to that newer `fetchOlder` call, and so this `fetchOlder` call shouldn't be interfering.

So please fix that bug from my draft :slightly_smiling_face:, and also add a test that would reproduce the bug.

PIG208: Thanks for the elaboration. I think encapsulating `_fetchingOlder` update with the generation check should be sufficient as a fix.

PIG208:
> This probably means
_MessageListPageState having its own notion of the current narrow,
which is initialized with widget.narrow but can change.
Then _MessageListState._modelChanged can check if the model's
narrow has changed, and if so tell the page state to update.

Relying the usual updates through dependencies does not work here because `MessageListPage` is a parent of `MessageList`. At first I was thinking if we want to push a route for `MessageListPage` to `Navigator` (similar to https://github.com/zulip/zulip-mobile/issues/5251), as a way to notify `_MessageListPageState` from `_MessageListState`.

But it might not be necessary. An alternative would be having `MessageListPage` pass a callback function to `MessageList`, and call it when `_modelChanged` happens to detect change in narrow.

gnprice:
> having `MessageListPage` pass a callback function to `MessageList`, and call it when `_modelChanged` happens to detect change in narrow.

Yeah, exactly.

We wouldn't want to push a new route — that'd leave the old one still on the nav stack, and make a new _MessageListPage State with a new _MessageListState and MessageListView. What zulip-mobile does is to mutate the route parameters on the existing route. If there's a way to do that with Flutter's navigator API then that could also be a good solution; I haven't looked into that.

test/model/message_list_test.dart
```diff
@@ -424,6 +425,355 @@ void main() {
...
+      final narrow = TopicNarrow(stream.streamId, 'topic');
+      final initialMessages = List.generate(5, (i) => eg.streamMessage(stream: stream, topic: 'topic'));
+      final movedMessages = List.generate(5, (i) => eg.streamMessage(stream: stream, topic: 'topic'));
+
+      test('internal move', () async {
+        await prepareNarrow(narrow, initialMessages + movedMessages);
+
+        await store.handleEvent(eg.updateMessageMoveEvent(movedMessages,
+          newTopic: movedMessages[0].topic,
+          origTopic: movedMessages[0].topic
```
PIG208, line 459: I wonder how this is possibly a move. Does internal move within a topic narrow actually happen?

lib/api/model/model.dart
```diff
@@ -576,7 +576,10 @@ class StreamMessage extends Message {
   @JsonKey(includeToJson: true)
   String get type => 'stream';
 
-  final String displayRecipient;
+  // This is not nullable API-wise, but if the message move across streams we
+  // might set [displayRecipient] to null to invalidate it.
```
PIG208, line 580: Not sure if this comment warrants dart doc given how it is mainly explaining the motivation for the `@JsonKey` here.

test/model/message_list_test.dart
gnprice: Hmm. Yeah, this case seems like it should be a no-op and shouldn't happen. So maybe it should just look like this (in the implementation):
```dart
          case (true,  true ): return; // TODO(log) no-op move
```

lib/api/model/model.dart
gnprice: Yeah, I think a plain implementation comment (like here) is fine.

PIG208:
Pushed to fix the CI error.

chrisbobbe:
Ah it looks like there's a rebase conflict. Could one of you please resolve that; maybe Zixuan?

PIG208:
Conflict resolved. (It was related to 7a2c9d2d36f9b0523cf1a34d4dbbb68b7f181aca).

chrisbobbe (commented):
Thanks, this will be helpful to have! Comments below.

test/example_data.dart
```diff
@@ -465,13 +465,17 @@ UpdateMessageEvent updateMessageEditEvent(
...
+  int? origStreamId,
   int? newStreamId,
   String? origTopic,
   String? newTopic,
   String? origContent,
   String? newContent,
 }) {
   assert(messages.isNotEmpty);
+  assert(origTopic != null, 'origTopic required for a message move');
+  assert(newTopic != null, 'newTopic required for a message move');
```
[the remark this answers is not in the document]
gnprice: Relatedly, if we want these asserts then they should probably just be expressed in the function's signature.

This function could also use some documentation to say how `messages` relates to the other parameters. In `updateMessageEditEvent` the Message parameter is meant to be the original, pre-edit version of the message, as reflected in its name `origMessage`. Here there's an echo of that (it's used as the default for `origStreamId`) but it doesn't seem to be applied consistently. It'd be good to have a consistent story one way or another — for example maybe it's more convenient to pass a list of the new versions of the messages, but then the function should apply that story throughout.

lib/model/message_list.dart
```diff
@@ -485,6 +505,84 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+  }
+
+  void _messagesMovedFromNarrow(
+    List<int> messageIds,
+    {TopicNarrow? newNarrow, PropagateMode? propagateMode}
+  ) {
+    if (_removeMessagesById(messageIds)) {
+      if (propagateMode == PropagateMode.changeAll
+          || propagateMode == PropagateMode.changeLater) {
+        narrow = newNarrow!;
```
gnprice, line 535: Separately: I think the overall behavior is going to be pretty broken as of this commit — the `MessageListView` changes what narrow it's tracking, but the associated `MessageListPage` is still about the old narrow. So the app bar still shows the old narrow, and the compose box will still send to the old narrow, even though the messages you're reading (including even any new messages that come in) are in the new narrow.

I think that can be fixed by just squashing the two commits together. They're both pretty small changes outside their tests.

test/example_data.dart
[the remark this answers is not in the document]
PIG208: > but then the function should apply that story throughout

Agreed. Will make it so that the helper always assume `messages` are the post-move messages, and express that through documentation and the removal of `newTopic` and `newStream` parameters that can then be consistently derived.

lib/model/message_list.dart
```diff
@@ -485,6 +505,84 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
   }
 
+  void _messagesMovedInternally(List<int> messageIds) {
+    for (final messageId in messageIds) {
+      if (_findMessageWithId(messageId) != -1) {
+        _reprocessAll();
+        notifyListeners();
+        return;
+      }
+    }
```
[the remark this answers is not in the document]
PIG208: Since this is returning right after an ID is found, the current implementation should be sufficient.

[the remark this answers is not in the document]
chrisbobbe: Oh I see; yeah, I guess my eyes just skipped right past the `return`.

PIG208:
Updated the PR. Thanks for the reviews @chrisbobbe!

chrisbobbe (commented):
Thanks! Small comments below.

I found a bug in manual testing, but it might really be a bug in the API documentation, or really the API itself, and it was latent before this PR. So maybe the fix should be a followup? Anyway, the issue is the some events are getting caught by this check in `_handleUpdateMessageEventMove`:

```dart
    if (newTopic == null) {
      // The `subject` field (aka newTopic) is documented to be present on moves.
      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic')); // TODO(log)
      return;
    }
```

It seems that when the message is moved to a new stream, but the _topic_ is the same as the topic in the original stream, the `subject` field (aka `newTopic`) is missing.

Here's the [API doc](https://zulip.com/api/get-events#update_message):

> `subject`: string
> 
> Only present if this event moved messages to a different topic.
> 
> The post-edit topic for all of the messages with IDs in `message_ids`.

gnprice:
> I found a bug in manual testing, but it might really be a bug in the API documentation, or really the API itself,

Good find, thanks. Started an `#api documentation` thread:
https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/update_message.20events/near/1902777

test/example_data.dart
```diff
@@ -495,6 +496,50 @@ UpdateMessageEvent updateMessageMoveEvent(
   );
 }
 
+/// [UpdateMessageEvent] where [origMessages] are moved to somewhere else.
+UpdateMessageEvent updateMessageMoveFromEvent({
```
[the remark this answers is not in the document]
gnprice: Yeah, those names SGTM.

PIG208:
Updated the PR addressing the comments. Thanks for the review! I was able to reproduce that API issue, and tweaked the checks we have to match the server behavior (also with some changes to both updateMessageEventMove helpers). Hopefully the correct behavior will documented soon.

test/example_data.dart
```diff
@@ -495,6 +497,52 @@ UpdateMessageEvent updateMessageMoveEvent(
   );
 }
 
+/// [UpdateMessageEvent] where [origMessages] are moved to somewhere else.
+UpdateMessageEvent updateMessageEventMoveFrom({
```
chrisbobbe, line 502:
```
test [nfc]: Split eg.updateMessageMoveEvent helper.
```

Is there more we can do to focus this commit on the described change? I see that `eg.updateMessageMoveEvent` gets split into two functions, so I know that that function's callers will have to adapt. But I see other changes around the callers that don't seem necessary for that adaptation, and I'm finding it difficult to verify the refactor itself. For example, it looks like a `prepareOrigMessages` helper is introduced. That seems like it should be possible to do in a separate commit.

lib/model/message.dart
```diff
@@ -167,9 +167,10 @@ class MessageStoreImpl with MessageStore {
       return;
     }
 
-    if (newTopic == null) {
-      // The `subject` field (aka newTopic) is documented to be present on moves.
-      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic')); // TODO(log)
+    if (newTopic == null && newStreamId == null) {
+      // At least one of `subject` (aka newTopic) or `new_stream_id` (aka
+      // newStreamId) is present documented to be present on moves.
```
chrisbobbe, line 173: typo in "present documented to be present"

test/model/message_list_test.dart
```diff
@@ -465,6 +466,337 @@ void main() {
...
+        await store.handleEvent(eg.updateMessageEventMoveFrom(
+          origMessages: movedMessages,
+          newTopic: 'new',
+        ));
+        checkHasMessages(initialMessages + movedMessages);
+        checkNotified(count: 2);
+      });
+    });
+
+    group('fetch races', (){
```
chrisbobbe, line 654: nit: space before `{`

lib/model/message.dart
```diff
@@ -167,33 +168,61 @@ class MessageStoreImpl with MessageStore {
...
+      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic or newStreamId')); // TODO(log)
```
chrisbobbe, line 174:
```
message: Handle moved messages from UpdateMessageEvent
```

Let's put this change in a separate commit before this one. It looks like it fixes a small but observable bug: in the case where you move a message to a different stream but keep the topic the same, we'll now update `message.editState` as appropriate for a move. 🙂

chrisbobbe (commented):
Thanks! Comments below.

test/example_data.dart
PIG208: Thanks for the advice. Extracted this part into `test [nfc]: Extract prepareOrigMessages`.

PIG208:
Updated the PR. Thanks!

chrisbobbe:

PIG208:
(pushed to resolve merge conflicts)

lib/model/message.dart
```diff
@@ -167,33 +168,61 @@ class MessageStoreImpl with MessageStore {
...
+      // The `propagate_mode` field (aka propagateMode) is documented to be present on moves.
+      assert(debugLog('Malformed UpdateMessageEvent: move but no propagateMode')); // TODO(log)
       return;
     }
 
-    // TODO(#150): Handle message moves.  The views' recipient headers
-    //   may need updating, and consequently showSender too.
-    //   Currently only editState gets updated.
+    final wasResolveOrUnresolve = (newStreamId == null && newTopic != null
+      && MessageEditState.topicMoveWasResolveOrUnresolve(origTopic, newTopic));
```
gnprice, line 189:
```suggestion
    final wasResolveOrUnresolve = (newStreamId == null
      && MessageEditState.topicMoveWasResolveOrUnresolve(origTopic, newTopic!));
```

The case where that `newTopic != null` condition would evaluate to false should be impossible, right? That's covered by the conditional above which leads to an early return.

When a condition is meant to be impossible, it's best to make that clear rather than just not do anything for that condition. Otherwise the reader has to wonder whether there's some handling that's missing for that condition.

test/model/message_test.dart
```diff
@@ -273,75 +273,122 @@ void main() {
...
+          newStreamId: 20,
           newContent: 'new content'));
-        check(store).messages[message.id].editState.equals(MessageEditState.edited);
-        check(store).messages[otherMessage.id].editState.equals(MessageEditState.moved);
+        checkNotified(count: 2);
+        check(store).messages[origMessages[0].id].editState.equals(MessageEditState.edited);
+        check(store).messages[origMessages[1].id].editState.equals(MessageEditState.moved);
+      });
+
+      test('message stream moved without topic change', () async {
```
gnprice, line 384: Looks like there's two tests with this name. Is it accurate for both? If so it should still get some extra words to distinguish what each test is about.

test/model/message_test.dart
```diff
@@ -273,75 +273,122 @@ void main() {
...
       test('message is both moved and updated', () async {
-        await sendEvent(message, eg.updateMessageMoveEvent([message, otherMessage],
-          origTopic: 'topic',
-          newTopic: 'topic',
-          newStreamId:  20,
-          origContent: 'old content',
+        await prepareOrigMessages(origTopic: 'topic', origContent: 'old content');
+        await store.handleEvent(eg.updateMessageEventMoveFrom(
+          origMessages: origMessages,
+          newTopic:  'new topic',
```
gnprice, line 376: This `newTopic` line changes in the commit:

    model: Expect the absence of newTopic for certain moves.

Are those changes connected? Does the new topic here need to change at all — could it just remain the way the test was, where the topic doesn't get renamed but the stream ID changes?

lib/model/message.dart
```diff
@@ -167,33 +168,61 @@ class MessageStoreImpl with MessageStore {
...
+      }
+
+      if (!wasResolveOrUnresolve
+          && message.editState == MessageEditState.none) {
+        message.editState = MessageEditState.moved;
+      }
+    }
+
+    for (final view in _messageListViews) {
+      view.messagesMoved(
```
gnprice, line 218: nit in commit message:

    Co-authored-by: Zixuan James Li <zixuan@zulip.com>

If we're going to have "Signed-off-by:" lines, let's follow the conventions of how they're used. It's a trailer that comes after the material you're signing off on. Are you signing off on the parts of this commit which you co-authored, or only on other parts? If the former, your "Signed-off-by:" should come after your "Co-authored-by:". (And if the latter, it's not clear what the line could really mean and it's probably best to just leave it out.)

For example usage, go into the Linux kernel repo and try a command like `git log --grep Co-author`.

gnprice (commented):
Thanks @PIG208 and @chrisbobbe! Comments below.

I got through the first four commits:
c7edec418 test [nfc]: Extract prepareOrigMessages.
6d5cd3a67 test [nfc]: Split eg.updateMessageMoveEvent helper.
57a3f2099 model [nfc]: Make displayRecipient nullable.
908b7c238 model: Expect the absence of newTopic for certain moves.

and part of the fifth. Left for a future round are the remainder of the fifth and the to commits after that:
4522cfa09 message: Handle moved messages from UpdateMessageEvent.
8afbde38b msglist [nfc]: Make narrow mutable for _MessageListPageState.
b84540fda msglist: Update narrow when propageMode is changeAll or changeLater.

test/widgets/message_list_test.dart
```diff
@@ -687,8 +688,8 @@ void main() {
       final rectsBefore = captureMessageRects(tester, messages, messageWithMarker);
       checkMarkersCount(edited: 0, moved: 0);
 
-      await store.handleEvent(eg.updateMessageMoveEvent(
-        [messageWithMarker], origTopic: 'old', newTopic: messageWithMarker.topic));
+      await store.handleEvent(eg.updateMessageEventMoveTo(
+        origTopic: 'old', newMessages: [messageWithMarker]));
```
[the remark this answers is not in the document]
PIG208: Hmm yeah I agree. This test needs to be refactored such that messageWithMarker is initialized in another topic, before being moved into the shared one.

lib/model/message.dart
```diff
@@ -167,33 +168,61 @@ class MessageStoreImpl with MessageStore {
       return;
     }
 
-    if (newTopic == null) {
-      // The `subject` field (aka newTopic) is documented to be present on moves.
-      assert(debugLog('Malformed UpdateMessageEvent: move but no newTopic')); // TODO(log)
+    if (newTopic == null && newStreamId == null) {
+      // At least one of `subject` (aka newTopic) or `new_stream_id` (aka
+      // newStreamId) is documented to be present on moves.
```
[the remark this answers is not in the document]
PIG208: Thanks! To summarize, prior the comment merely was an isolated interpretation of the API docs, but after the edit, it explains the assertion with context.

lib/model/message.dart
PIG208: That's a good catch. Thanks!

test/model/message_test.dart
PIG208: Looks like one of them is an artifact of a misapplied fixup commit. Neither is needed, so both are removed. Renaming an old test that covers this same story.

test/model/message_test.dart
PIG208: Right, the tests changes shouldn't be necessary.

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
PIG208: The refactor actually requires moving the message back and forth, and it has to be implemented after `message: Handle moved messages from UpdateMessageEvent.`. It wouldn't work otherwise because the move might violate the assertions (since neither topic/streamId can be updated at the time the example data helper is split).

test/example_data.dart
```diff
@@ -463,15 +463,18 @@ UpdateMessageEvent updateMessageEditEvent(
...
   int? newStreamId,
-  String? origTopic,
   String? newTopic,
   String? origContent,
   String? newContent,
+  PropagateMode propagateMode = PropagateMode.changeOne,
 }) {
   assert(messages.isNotEmpty);
+  assert(newTopic != origTopic || newStreamId != null);
   final origMessage = messages[0];
```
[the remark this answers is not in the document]
PIG208: Make sense. Removing the variable entirely now that we have the `flags` parameter.

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
gnprice: I see. OK, in that case then the earlier commit can just leave a comment saying yes, this data looks wrong, it'll be fixed soon.

The blocker for fixing it is basically #150, right? In that case the comment can look like `// TODO(#150): …`.

test/example_data.dart
```diff
@@ -463,15 +463,18 @@ UpdateMessageEvent updateMessageEditEvent(
...
-UpdateMessageEvent updateMessageMoveEvent(
+UpdateMessageEvent _updateMessageMoveEvent(
   List<Message> messages, {
+  required int origStreamId,
+  required String origTopic,
   int? newStreamId,
-  String? origTopic,
   String? newTopic,
   String? origContent,
   String? newContent,
```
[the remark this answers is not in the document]
PIG208: While `UpdateMessageEvent` orders `flags` and `propagateMode` differently, I put these two parameters to the last for this helper to have all the orig/new pairs grouped together.

test/example_data.dart
```diff
@@ -495,6 +498,54 @@ UpdateMessageEvent updateMessageMoveEvent(
   );
 }
 
+/// [UpdateMessageEvent] where [origMessages] are moved to somewhere else.
+UpdateMessageEvent updateMessageEventMoveFrom({
+  required List<StreamMessage> origMessages,
+  String? newTopic,
+  int? newStreamId,
```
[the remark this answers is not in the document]
PIG208: Speaking of https://github.com/zulip/zulip-flutter/pull/787#discussion_r1696096924, do you think we should insert `propagateMode` and `flags` in between? Currently they are added to the end of the named parameters.

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
PIG208: Yeah. This is fixed in `message_list_test: Move marker message back and forth`, after move handling is implemented. I will include the TODO comment in the refactor that touches but not fixes this test.

test/example_data.dart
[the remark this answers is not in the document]
gnprice: I think at the end is fine — it seems like a logical place for those two. (`flags` is closely tied to `content`, which is at the end.)

Putting them in the middle, in the same order as on the event class's declaration, would also be fine. There's a bit of a tension because, ultimately, the order in the docs doesn't match a logical order; on the class we choose to match the docs.

PIG208:
This is ready for review now. Thanks for the review!

gnprice (commented):
Thanks! Those first five commits look good now:
e5576ba83 model: Expect the absence of newTopic for certain moves.
f3d00a0fc message test [nfc]: Extract prepareOrigMessages.
11f3b564d message_test [nfc]: Rename message move test for accuracy.
c3c69bce6 test [nfc]: Split eg.updateMessageMoveEvent helper.
bcd6f9688 model [nfc]: Make displayRecipient nullable.

Some comments below on the tests in the main commit:
0b54bb7e1 message: Handle moved messages from UpdateMessageEvent.

Heading into a meeting now but I might return later today to review more of that commit. Either way, there's more tests to write here, so I think you can productively start on revisions before my next review.

PIG208:
Addressed the comments other than the new tests one. Added some additional wip tests in the last commit, and will continue working on it when I get back.

PIG208:
Updated the tests to be more comprehensive (especially for topic narrow).

test/widgets/message_list_test.dart
```diff
@@ -605,6 +606,86 @@ void main() {
           expectedMessage: 'NetworkException: Oops (ClientException: Oops)');
       });
     });
+
+    group('Update Narrow on message move', () {
+      final channel = eg.stream(name: 'move test stream');
+      final message = eg.streamMessage(stream: channel, content: 'Message to move');
```
gnprice, line 612: The message gets mutated. So each test should make it fresh — else it gets the version mutated by the last test.

test/widgets/message_list_test.dart
```diff
@@ -605,6 +606,86 @@ void main() {
...
+
+        connection.prepare(json: SendMessageResult(id: 1).toJson());
+        await tester.tap(find.byIcon(Icons.send));
+        await tester.pump();
+        check(connection.lastRequest).isA<http.Request>()
+          ..method.equals('POST')
+          ..url.path.equals('/api/v1/messages')
+          ..bodyFields.deepEquals({
+            'type': 'stream',
+            'to': '${message.streamId}',
```
gnprice, line 650: Similarly, because the message gets mutated this should refer to the stream/channel object and not the message.

test/widgets/message_list_test.dart
```diff
@@ -605,6 +606,86 @@ void main() {
...
+      }
+
+      void handleMessageMoveEvent(List<StreamMessage> messages, String newTopic) {
+        store.handleEvent(eg.updateMessageEventMoveFrom(
+          origMessages: messages,
+          newTopic: newTopic,
+          propagateMode: PropagateMode.changeAll));
+      }
+
+      testWidgets('compose box send message after move', (WidgetTester tester) async {
```
gnprice, line 627: Let's have this test move the message to a new channel as well as topic. That way we more fully exercise things changing.

gnprice (commented):
Thanks for the revision! Here's a review of the whole rest of the PR.

Most comments are small; there's one more-significant point about the behavior of propagateMode.

lib/model/message_list.dart
```diff
@@ -487,6 +507,92 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+  }
+
+  void _messagesMovedFromNarrow(
+    List<int> messageIds,
+    PropagateMode propagateMode,
+    {TopicNarrow? newNarrow}
+  ) {
+    if (_removeMessagesById(messageIds)) {
+      if (newNarrow != null) {
+        _handlePropagateMode(propagateMode, newNarrow);
```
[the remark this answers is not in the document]
PIG208: > Fixing that would be complex and is out of scope at present, even if we were to decide it's desirable.

Starting a [conversation](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Accurately.20handle.20propagateMode.2EchangeLater) so that we have a place for future discussion on this.

PIG208:
This should be ready for review now. Thanks!

PIG208:
(To keep track of the status) this PR is ready for review.

test/widgets/message_list_test.dart
```diff
@@ -620,6 +621,93 @@ void main() {
     });
   });
 
+  group('Update Narrow on message move', () {
+    final channel = eg.stream(name: 'move test stream');
+    final otherChannel = eg.stream(name: 'other move test stream');
+    final narrow = TopicNarrow(channel.streamId, 'example topic');
```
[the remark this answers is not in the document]
gnprice: > A straightforward way to do that is to make a local variable `topic` and use it in all the places that need to be talking about the same topic.

The other easy way that comes to mind is to not specify the topic name at all, but get it from the message with `message.topic` (and/or indirect versions of the same, like `TopicNarrow.ofMessage`).

test/widgets/message_list_test.dart
```diff
@@ -620,6 +621,93 @@ void main() {
...
+      await setupMessageListPage(tester, narrow: narrow, messages: [message], streams: [channel, otherChannel]);
+
+      final channelContentInputFinder = find.descendant(
+        of: find.byType(ComposeAutocomplete),
+        matching: find.byType(TextField));
+
+      await tester.enterText(channelContentInputFinder, 'Some text');
+      check(tester.widget<TextField>(channelContentInputFinder))
+        ..decoration.isNotNull().hintText.equals('Message #${channel.name} > example topic')
+        ..controller.isNotNull().text.equals('Some text');
```
gnprice, line 653: Cool, good idea adding this check — it makes a baseline for the reader to compare the check below to.

test/widgets/message_list_test.dart
```diff
@@ -620,6 +621,93 @@ void main() {
...
+      prepareGetMessageResponse([message]);
```
gnprice, line 654: nit:
```suggestion
        ..controller.isNotNull().text.equals('Some text');

      prepareGetMessageResponse([message]);
```

Generally in a test case when there's a sequence like
* make something happen or change
* check the state of the world
* make something else happen or change
* check the new state of the world
* etc.

I find it helpful, for the reader, to group those change/check pairs into stanzas using blank lines.

lib/model/message_list.dart
```diff
@@ -489,6 +509,87 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+      case PropagateMode.changeAll:
+      case PropagateMode.changeLater:
+        narrow = newNarrow;
+        _reset();
+        fetchInitial();
+      case PropagateMode.changeOne:
+    }
+  }
+
+  void _messagesMovedFromNarrow(List<int> messageIds) {
```
gnprice, line 543: nit: group this with the other `_messagesMoved…` methods, above `_handlePropagateMode`

gnprice (commented):
Thanks @PIG208 for the revision! Just a handful of small comments now; then this will be all ready to merge.

PIG208:
Thanks. Updated the PR.

gnprice:

I made a few commit-message tweaks, all in the summary prefixes:

    -    model: Expect the absence of newTopic for certain moves.
    +    message: Expect the absence of newTopic for certain moves.

    -    message_test [nfc]: Rename message move test for accuracy.
    +    message test [nfc]: Rename message move test for accuracy.

    -    model [nfc]: Make displayRecipient nullable.
    +    message [nfc]: Make displayRecipient nullable.

The "test" modifier in a summary prefix is always separated by a space, not underscore.

Rather than "model", these can be more specific by saying "message", referring to the MessageStore part of the model. (The first commit is all about that code directly. The later displayRecipient commit doesn't touch that code itself… but the change it's making to Message can be thought of as being entirely for the purpose of Message's role in MessageStore, where those objects get mutated.)

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
gnprice: > Probably we should make `eg.streamMessage` generate the topic name at random, to help prevent this pattern, the same way we do stream/channel IDs and message IDs and so on.

Sent #932 for this.

> and none of those assume _what_ the default topic name is, only that it's some constant value, which is also an assumption to avoid making implicit but is a different one.

I ended up deciding this assumption was fine, though; it's convenient in a variety of tests. So I made the randomness happen only once per run, and documented that.

PR #789, opened 2024-07-05
code_block: Fix null-check `!` errors during hot reloads

chrisbobbe:
On hot-reload, `lerp` is triggered even when the theme has not changed. Discussion:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Exceptions.20during.20hot.20reloads/near/1846095

I considered just going ahead with placing the web app's current light-theme styles for code blocks (#754), but pretty soon I realized there would still be some nulls in there after doing so, and those would need to be treated like we're treating the nulls here.


gnprice:
Thanks! Looks good; merging.

Issue #716, opened 2024-06-01
Handle unknown users everywhere


gnprice:
Our local store has a map with all the users we know about, in `PerAccountStore.users` aka `store.users`. But this might not be all the users in the realm. In particular, if the self-user is a guest user, there may be users we don't have permission to know about.

For the most part the users we actually encounter will be users we do have data on: for example, a guest user still has permission to know about users that are subscribed to any of the same streams they are, which means that most of the messages they see there will be sent by users they know about. But there could be older messages by users who've since unsubscribed, and other scenarios. So we will sometimes encounter a user ID that doesn't exist in `store.users`.

We should therefore make sure we never crash when we encounter a user that's not in `store.users`. This means:

* When we look up a user with an expression like `store.users[userId]`, we should explicitly handle the possibility that the result is null.

  This means no `!` and no `assert` saying it's non-null. Depending on the context, the code might use `?.` and `??` to fall back to a placeholder like `ZulipLocalizations.unknownUserName`, or might return without doing anything, or something else as appropriate.

  * The exception is that it's fine to assume the self-user is present. We have a few places that do this; they could benefit from being factored out into a small getter like `store.selfUser`.

* When we have a context like a `Message` that provides its own details on a user (like the message sender), we should fall back to those if the user is unknown. That way we not only don't crash, but provide helpful information, in cases like a guest user looking at older messages sent by someone no longer subscribed to their streams.

We already do these correctly in many places, but there are several places we don't. So this task is about sweeping through those, probably in one commit per call site, and fixing them.

A key part of doing this is to find an exhaustive list of places that are potentially affected. You can do this in an IDE by going to the definition of `PerAccountStore.users` and then using the "find references" feature.

mofirojean:
> Our local store has a map with all the users we know about, in `PerAccountStore.users` aka `store.users`. But this might not be all the users in the realm. In particular, if the self-user is a guest user, there may be users we don't have permission to know about.
> 
> For the most part the users we actually encounter will be users we do have data on: for example, a guest user still has permission to know about users that are subscribed to any of the same streams they are, which means that most of the messages they see there will be sent by users they know about. But there could be older messages by users who've since unsubscribed, and other scenarios. So we will sometimes encounter a user ID that doesn't exist in `store.users`.
> 
> We should therefore make sure we never crash when we encounter a user that's not in `store.users`. This means:
> 
> * When we look up a user with an expression like `store.users[userId]`, we should explicitly handle the possibility that the result is null.
[...]

PR #790, opened 2024-07-05
msglist: Display typing indicators on typing.

### LINKED ISSUE #665: Show "typing" status

PIG208:
Display a typing indicator when there are people typing in a `SendableNarrow` such as `DmNarrow` and `TopicNarrow`, in a fashion similar to the web app.

This is stacked on top of #783 and fixes #665.

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -483,5 +483,24 @@
   "notifSelfUser": "You",
   "@notifSelfUser": {
     "description": "Display name for the user themself, to show after replying in an Android notification"
+  },
+  "onePersonTyping": "{typist} is typing…",
+  "@onePersonTyping": {
+    "description": "Text to display when there is one user typing.",
+    "placeholders": {
+      "typist": {"type": "String", "example": "Alice"}
+    }
+  },
+  "twoPeopleTyping": "{typist} and {otherTypist} are typing…",
+  "@twoPeopleTyping": {
+    "description": "Text to display when there are two users typing.",
+    "placeholders": {
+      "typist": {"type": "String", "example": "Alice"},
+      "otherTypist": {"type": "String", "example": "Bob"}
+    }
+  },
+  "manyPeopleTyping": "Several people are typing…",
+  "@manyPeopleTyping": {
+    "description": "Text to display when there are multiple users typing."
   }
 }
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 
+import '../../model/algorithms.dart';
 import 'json.dart';
 import 'model.dart';
 
@@ -900,7 +901,7 @@ class TypingEvent extends Event {
     required this.recipientIds,
     required this.streamId,
     required this.topic,
-  });
+  }) : assert(isSortedWithoutDuplicates(recipientIds ?? []));
 
   static Object? _readSenderId(Map<Object?, Object?> json, String key) {
     return (json['sender'] as Map<String, dynamic>)['user_id'];
@@ -909,7 +910,7 @@ class TypingEvent extends Event {
   static List<int>? _recipientIdsFromJson(Object? json) {
     if (json == null) return null;
     return (json as List<Object?>).map(
-      (item) => (item as Map<String, Object?>)['user_id'] as int).toList();
+      (item) => (item as Map<String, Object?>)['user_id'] as int).toList()..sort();
   }
 
   factory TypingEvent.fromJson(Map<String, dynamic> json) {
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -901,7 +901,7 @@ class TypingEvent extends Event {
     required this.recipientIds,
     required this.streamId,
     required this.topic,
-  }) : assert(isSortedWithoutDuplicates(recipientIds ?? []));
+  }) : assert(recipientIds == null || isSortedWithoutDuplicates(recipientIds));
 
   static Object? _readSenderId(Map<Object?, Object?> json, String key) {
     return (json['sender'] as Map<String, dynamic>)['user_id'];
--- a/lib/api/route/events.dart
+++ b/lib/api/route/events.dart
@@ -16,7 +16,7 @@ Future<InitialSnapshot> registerQueue(ApiConnection connection) {
       'notification_settings_null': true,
       'bulk_message_deletion': true,
       'user_avatar_url_field_optional': false, // TODO(#254): turn on
-      'stream_typing_notifications': false, // TODO implement
+      'stream_typing_notifications': true,
       'user_settings_object': true,
     },
   });
--- a/lib/model/typing_status.dart
+++ b/lib/model/typing_status.dart
@@ -64,7 +64,7 @@ class TypingStatus extends ChangeNotifier {
   void handleTypingEvent(TypingEvent event) {
     SendableNarrow narrow = switch (event.messageType) {
       MessageType.direct => DmNarrow(
-        allRecipientIds: event.recipientIds!..sort(), selfUserId: selfUserId),
+        allRecipientIds: event.recipientIds!, selfUserId: selfUserId),
       MessageType.stream => TopicNarrow(event.streamId!, event.topic!),
     };
 
--- a/lib/model/typing_status.dart
+++ b/lib/model/typing_status.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:flutter/foundation.dart';
 
 import '../api/model/events.dart';
+import '../log.dart';
 import 'narrow.dart';
 
 /// The model for tracking the typing status organized by narrows.
@@ -38,6 +39,10 @@ class TypingStatus extends ChangeNotifier {
   }
 
   bool _addTypist(SendableNarrow narrow, int typistUserId) {
+    if (typistUserId == selfUserId) {
+      assert(debugLog('typing status: adding self as typist'));
+      return false;
+    }
     final narrowTimerMap = _timerMapsByNarrow[narrow] ??= {};
     final typistTimer = narrowTimerMap[typistUserId];
     final isNewTypist = typistTimer == null;
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -2,6 +2,7 @@ import 'dart:math';
 
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_color_models/flutter_color_models.dart';
 import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:intl/intl.dart';
 
@@ -9,6 +10,7 @@ import '../api/model/model.dart';
 import '../model/message_list.dart';
 import '../model/narrow.dart';
 import '../model/store.dart';
+import '../model/typing_status.dart';
 import 'action_sheet.dart';
 import 'actions.dart';
 import 'compose_box.dart';
@@ -496,9 +498,9 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
           final valueKey = key as ValueKey<int>;
           final index = model!.findItemWithMessageId(valueKey.value);
           if (index == -1) return null;
-          return length - 1 - (index - 2);
+          return length - 1 - (index - 3);
         },
-        childCount: length + 2,
+        childCount: length + 3,
         (context, i) {
           // To reinforce that the end of the feed has been reached:
           //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20Mark-as-read/near/1680603
@@ -506,7 +508,9 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
 
           if (i == 1) return MarkAsReadWidget(narrow: widget.narrow);
 
-          final data = model!.items[length - 1 - (i - 2)];
+          if (i == 2) return TypingStatusWidget(narrow: widget.narrow);
+
+          final data = model!.items[length - 1 - (i - 3)];
           return _buildItem(data, i);
         }));
 
@@ -606,6 +610,64 @@ class ScrollToBottomButton extends StatelessWidget {
         // Web has the same color in light and dark mode.
         color: const HSLColor.fromAHSL(0.5, 240, 0.96, 0.68).toColor(),
         onPressed: _navigateToBottom));
+  }
+}
+
+class TypingStatusWidget extends StatefulWidget {
+  const TypingStatusWidget({super.key, required this.narrow});
+
+  final Narrow narrow;
+
+  @override
+  State<StatefulWidget> createState() => _TypingStatusWidgetState();
+}
+
+class _TypingStatusWidgetState extends State<TypingStatusWidget> with PerAccountStoreAwareStateMixin<TypingStatusWidget> {
+  TypingStatus? model;
+
+  @override
+  void onNewStore() {
+    model?.removeListener(_modelChanged);
+    model = PerAccountStoreWidget.of(context).typingStatus
+      ..addListener(_modelChanged);
+  }
+
+  @override
+  void dispose() {
+    model?.removeListener(_modelChanged);
+    super.dispose();
+  }
+
+  void _modelChanged() {
+    setState(() {
+      // The actual state lives in [model].
+      // This method was called because that just changed.
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final narrow = widget.narrow;
+    if (narrow is! SendableNarrow) return const SizedBox();
+
+    final store = PerAccountStoreWidget.of(context);
+    final localizations = ZulipLocalizations.of(context);
+    final typistIds = model!.typistIdsInNarrow(narrow);
+    if (typistIds.isEmpty) return const SizedBox();
+    final text = switch (typistIds.length) {
+      1 => localizations.onePersonTyping(
+        store.users[typistIds.first]?.fullName ?? localizations.unknownUserName),
+      2 => localizations.twoPeopleTyping(
+        store.users[typistIds.first]?.fullName ?? localizations.unknownUserName,
+        store.users[typistIds.last]?.fullName  ?? localizations.unknownUserName),
+      _ => localizations.manyPeopleTyping,
+    };
+
+    return Padding(
+      padding: const EdgeInsetsDirectional.only(start: 16, top: 2),
+      child: Text(text,
+        style: const TextStyle(
+          color: HslColor(0, 0, 53), fontStyle: FontStyle.italic)));
   }
 }
 
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -320,6 +320,70 @@ void main() {
     });
   });
 
+  group('TypingStatusWidget', () {
+    final users = [eg.selfUser, eg.otherUser, eg.thirdUser, eg.fourthUser];
+    final finder = find.descendant(
+      of: find.byType(TypingStatusWidget),
+      matching: find.byType(Text)
+    );
+
+    Future<void> checkTyping(WidgetTester tester, TypingEvent event, {required String expected}) async {
+      await store.handleEvent(event);
+      await tester.pump();
+      check(tester.widget<Text>(finder)).data.equals(expected);
+    }
+
+    final dmMessage = eg.dmMessage(
+      from: eg.selfUser, to: [eg.otherUser, eg.thirdUser, eg.fourthUser]);
+    final dmNarrow = DmNarrow.ofMessage(dmMessage, selfUserId: eg.selfUser.userId);
+
+    final streamMessage = eg.streamMessage();
+    final topicNarrow = TopicNarrow.ofMessage(streamMessage);
+
+    for (final (description, message, narrow) in [
+      ('typing in dm',    dmMessage,      dmNarrow),
+      ('typing in topic', streamMessage,  topicNarrow),
+    ]) {
+      testWidgets(description, (tester) async {
+        await setupMessageListPage(tester,
+          narrow: narrow, users: users, messages: [message]);
+        await tester.pump();
+        check(finder.evaluate()).isEmpty();
+        await checkTyping(tester,
+          eg.typingEvent(narrow, TypingOp.start, eg.otherUser.userId),
+          expected: 'Other User is typing…');
+        await checkTyping(tester,
+          eg.typingEvent(narrow, TypingOp.start, eg.selfUser.userId),
+          expected: 'Other User is typing…');
+        await checkTyping(tester,
+          eg.typingEvent(narrow, TypingOp.start, eg.thirdUser.userId),
+          expected: 'Other User and Third User are typing…');
+        await checkTyping(tester,
+          eg.typingEvent(narrow, TypingOp.start, eg.fourthUser.userId),
+          expected: 'Several people are typing…');
+        await checkTyping(tester,
+          eg.typingEvent(narrow, TypingOp.stop, eg.otherUser.userId),
+          expected: 'Third User and Fourth User are typing…');
+        // Verify that typing indicators expire after a set duration.
+        await tester.pump(const Duration(seconds: 15));
+        check(finder.evaluate()).isEmpty();
+      });
+    }
+
+    testWidgets('unknown user typing', (tester) async {
+      final streamMessage = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(streamMessage);
+      await setupMessageListPage(tester,
+        narrow: narrow, users: [], messages: [streamMessage]);
+      await checkTyping(tester,
+        eg.typingEvent(narrow, TypingOp.start, 1000),
+        expected: '(unknown user) is typing…',
+      );
+      // Wait for the pending timers to end.
+      await tester.pump(const Duration(seconds: 15));
+    });
+  });
+
   group('MarkAsReadWidget', () {
     bool isMarkAsReadButtonVisible(WidgetTester tester) {
       final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -320,6 +320,229 @@ void main() {
     });
   });
 
+  group('MarkAsReadWidget', () {
+    bool isMarkAsReadButtonVisible(WidgetTester tester) {
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+      final finder = find.text(
+        zulipLocalizations.markAllAsReadLabel).hitTestable();
+      return finder.evaluate().isNotEmpty;
+    }
+
+    testWidgets('from read to unread', (WidgetTester tester) async {
+      final message = eg.streamMessage(flags: [MessageFlag.read]);
+      await setupMessageListPage(tester, messages: [message]);
+      check(isMarkAsReadButtonVisible(tester)).isFalse();
+
+      await store.handleEvent(eg.updateMessageFlagsRemoveEvent(
+        MessageFlag.read, [message]));
+      await tester.pumpAndSettle();
+      check(isMarkAsReadButtonVisible(tester)).isTrue();
+    });
+
+    testWidgets('from unread to read', (WidgetTester tester) async {
+      final message = eg.streamMessage(flags: []);
+      final unreadMsgs = eg.unreadMsgs(channels:[
+        UnreadChannelSnapshot(topic: message.topic, streamId: message.streamId, unreadMessageIds: [message.id])
+      ]);
+      await setupMessageListPage(tester, messages: [message], unreadMsgs: unreadMsgs);
+      check(isMarkAsReadButtonVisible(tester)).isTrue();
+
+      await store.handleEvent(UpdateMessageFlagsAddEvent(
+        id: 1,
+        flag: MessageFlag.read,
+        messages: [message.id],
+        all: false,
+      ));
+      await tester.pumpAndSettle();
+      check(isMarkAsReadButtonVisible(tester)).isFalse();
+    });
+
+    testWidgets("messages don't shift position", (WidgetTester tester) async {
+      final message = eg.streamMessage(flags: []);
+      final unreadMsgs = eg.unreadMsgs(channels:[
+        UnreadChannelSnapshot(topic: message.topic, streamId: message.streamId,
+          unreadMessageIds: [message.id])
+      ]);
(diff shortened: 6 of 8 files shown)


PIG208:
https://github.com/zulip/zulip-flutter/assets/39874143/acee3a3f-b567-4f07-b0dd-dcc9dcfc3d9a

test/widgets/message_list_test.dart
```diff
@@ -1009,4 +1009,64 @@ void main() {
...
+      matching: find.byType(Text)
+    );
+
+    checkTyping(WidgetTester tester, TypingEvent event, {required List<User> expected}) async {
+      await store.handleEvent(event);
+      await tester.pump();
+      final otherTypist = (expected.length > 1) ? expected[1] : null;
+      check(finder.evaluate()).single.has((x) => x.widget, 'widget').isA<Text>()
+        .data.equals(zulipLocalizations.typingIndicator(
+          expected.length, expected[0].fullName, otherTypist?.fullName ?? ''));
```
gnprice, line 1027: In light of my other comment, let's have these tests use concrete strings for expectations, instead of `zulipLocalizations` results :slightly_smiling_face: Try changing the tests first, and please confirm whether my prediction above is correct (at least about English).

I think just testing in English will suffice to make the point and confirm that the logic is happening outside of the i18n plural rules.

gnprice (commented):
Thanks! We'll get to a full review after #783.

On a quick skim, though, here's a pair of initial comments. And one more:
* Please add a screenshot of what the typing indicators look like when visible. … It looks like we don't currently mention that [in our guide to reviewable PRs](https://zulip.readthedocs.io/en/stable/contributing/reviewable-prs.html#tips-and-best-practices), but I've seen @alya mention in a couple of threads in the past that that's helpful for reviewing the UI, even when there's also a video, because it makes it easier to look closely at the key moments in the UI's behavior.

  (That guide does have a couple of other good tips on posting screenshots and videos, though.)

alya:
Looking at the video, do we specifically have a mobile design for this? If not, I'd be inclined to match the web UI styling. Or if the differences are intentional, we should have a #design thread as to why it's different, and whether to update the web UI.

Screenshot 2024-07-05 at 15 14 06@2x

alya:
> Thanks! We'll get to a full review after #783.
> 
> On a quick skim, though, here's a pair of initial comments. And one more:
> 
> * Please add a screenshot of what the typing indicators look like when visible. … It looks like we don't currently mention that [in our guide to reviewable PRs](https://zulip.readthedocs.io/en/stable/contributing/reviewable-prs.html#tips-and-best-practices), but I've seen @alya mention in a couple of threads in the past that that's helpful for reviewing the UI, even when there's also a video, because it makes it easier to look closely at the key moments in the UI's behavior.
>   (That guide does have a couple of other good tips on posting screenshots and videos, though.)

That's an old version of the documentation. I think it's more clear in the [current version](https://zulip.readthedocs.io/en/latest/contributing/reviewable-prs.html#explain-your-changes).

gnprice:
> That's an old version of the documentation. I think it's more clear in the [current version](https://zulip.readthedocs.io/en/latest/contributing/reviewable-prs.html#explain-your-changes).

Ah indeed, thanks — sorry for the confusion. I thought I remembered seeing such a writeup, so I was a bit puzzled not to find it; I didn't notice I'd landed on the "stable" version of the docs :slightly_smiling_face:

assets/l10n/app_en.arb
```diff
@@ -479,5 +479,14 @@
       "senderFullName": {"type": "String", "example": "Alice"},
       "numOthers": {"type": "int", "example": "4"}
     }
+  },
+  "typingIndicator": "{numPeople, plural, =1{{typist} is typing} =2{{typist} and {otherTypist} are typing} other{Several people are typing}}",
```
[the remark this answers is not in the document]
PIG208: Thanks! Those are some interesting distinctions. `=2` seems to be used for English after modifying the test case.

We can split this into 3 separate strings (or just two strings equivalent to the "=1" and "other" here, respectively) without the "plural" feature.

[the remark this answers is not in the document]
gnprice: > `=2` seems to be used for English after modifying the test case.

Huh, surprising.

In that case I'd guess that in Chinese it'd use the `=1` and `=2` forms too, if present… but a translator UI, like that in Transifex, generally wouldn't ask the Chinese translators for such forms, because in the normal grammatical behavior of the language (more precisely, in its behavior as recorded in the CLDR database, shown in that giant table at my third link) there's no distinction that would call for such forms. So we'd end up with only an "other" form in our translations.

And I'd expect that in Russian it'd still behave the way I predicted above, using the `=1` form more broadly than this bit of program logic intends.

PIG208:
Updated the PR to use separate translation strings and a design that looks more similar to the one used in the web app.

<details>

<summary>Design comparison</summary>

| web | Flutter |
| ----- | --- |
| Screenshot from 2024-07-05 18-57-53 | 
| Screenshot from 2024-07-05 18-58-03 | 

</details>

assets/l10n/app_en.arb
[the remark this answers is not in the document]
gnprice: Cross-posted a version of the above to #mobile-team, and then added further findings:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/i18n.20plurals.20.2F.20grammatical.20number/near/1848780

alya:
I think we should include the `...` on mobile, too...

PIG208:
<details>

<summary> Updated Design: </summary>

Screenshot from 2024-07-08 15-40-48
Screenshot from 2024-07-08 15-40-40
Screenshot from 2024-07-08 15-40-33

</details>

alya:
Works for me!

lib/widgets/message_list.dart
```diff
@@ -510,6 +514,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+
+    final localization = ZulipLocalizations.of(context);
+    final String text = switch (typistNames.length) {
+      1 => localization.onePersonTyping(typistNames[0]),
+      2 => localization.twoPeopleTyping(typistNames[0], typistNames[1]),
+      _ => localization.manyPeopleTyping,
+    };
+
+    return Padding(
+      padding: const EdgeInsets.only(left: 16, top: 2),
```
chrisbobbe, line 561: Since left and right aren't equal, this looks like a place to use `EdgeInsetsDirectional`.

lib/widgets/message_list.dart
```diff
@@ -510,6 +514,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      child: Text(text,
+        textAlign: TextAlign.left,
```
chrisbobbe, line 563: `TextAlign.start`, I think.

test/widgets/message_list_test.dart
```diff
@@ -1009,4 +1009,61 @@ void main() {
...
+    );
+
+    checkTyping(WidgetTester tester, TypingEvent event, {required String expected}) async {
+      await store.handleEvent(event);
+      await tester.pump();
+      check(finder.evaluate()).single.has((x) => x.widget, 'widget').isA<Text>()
+        .data.equals(expected);
+    }
+
+    testTyping(WidgetTester tester, SendableNarrow narrow) async {
```
chrisbobbe, line 1027: I think we only normally use the name `testFoo` when the function itself calls `test`/`testWidgets`. I see that `checkTyping` is already taken, though, by the helper just above this. Hmm.

Could find a different name, or maybe not use a helper and just cover the dm and topic cases with something like this (untested):

```dart
for (final (description, narrow, message) in [
  ('dm', …,  …),
  ('topic', …,  …),
]) {
  testWidgets(description, …
```

lib/widgets/message_list.dart
```diff
@@ -510,6 +514,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final narrow = widget.narrow;
+    const placeholder = SizedBox(height: 8);
+    if (narrow is! SendableNarrow) return placeholder;
+
+    final typistNames = model!.typistIdsInNarrow(narrow)
+      .where((id) => id != store.selfUserId)
+      .map((id) => store.users[id]!.fullName)
```
chrisbobbe, line 549: I think technically this should have an "(unknown user)" fallback, for #716? I acknowledge that it would be rare and silly for a nonexistent user to show up as typing, though.

assets/l10n/app_en.arb
```diff
@@ -483,5 +483,24 @@
   "notifSelfUser": "You",
   "@notifSelfUser": {
     "description": "Display name for the user themself, to show after replying in an Android notification"
+  },
+  "onePersonTyping": "{typist} is typing...",
```
chrisbobbe, line 487: We probably want the "…" character instead of "...", right?

chrisbobbe (commented):
Thanks! Small comments below.

PIG208:
Updated the PR. Thanks for the review @chrisbobbe!

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
   }
 }
 
+class _TypingStatusState extends State<TypingStatusWidget> with PerAccountStoreAwareStateMixin<TypingStatusWidget> {
```
chrisbobbe, line 537: Normally we organize stateful-widget code by putting the widget first and then the state.

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      _ => localization.manyPeopleTyping,
+    };
+
+    return Padding(
+      padding: const EdgeInsetsDirectional.only(start: 16, top: 2),
+      child: Text(text,
+        textAlign: TextAlign.start,
+        style: const TextStyle(
+          color: HslColor(0, 0, 53), fontStyle: FontStyle.italic)),
+    );
```
chrisbobbe, line 586:
```suggestion
          color: HslColor(0, 0, 53), fontStyle: FontStyle.italic)));
```

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      // The actual state lives in [model].
+      // This method was called because that just changed.
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final narrow = widget.narrow;
+    const placeholder = SizedBox(height: 8);
```
chrisbobbe, line 564: When the typing status is not shown, this puts an extra 8px between the last message and the mark-as-read button. If that's not part of the design, let's avoid it.

chrisbobbe (commented):
Thanks! Small comments below, and I'll mark this for Greg's review.

lib/api/route/events.dart
```diff
@@ -16,7 +16,7 @@ Future<InitialSnapshot> registerQueue(ApiConnection connection) {
       'notification_settings_null': true,
       'bulk_message_deletion': true,
       'user_avatar_url_field_optional': false, // TODO(#254): turn on
-      'stream_typing_notifications': false, // TODO implement
+      'stream_typing_notifications': true,
```
gnprice, line 19: This should get its own commit and own commit message, since it's changing what we get from the server (whereas the rest of the commit has a very different character, adding UI).

I think the effect is that it's just causing us to get more `TypingEvent`s, and our code already handles the new kind?

lib/widgets/message_list.dart
gnprice: Relatedly, follow the normal pattern for the relationship between widget and state classes' names: widget Foo has state _FooState.

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+    if (narrow is! SendableNarrow) return placeholder;
+
+    final localization = ZulipLocalizations.of(context);
```
gnprice, line 567: nit: plural, not singular — this is an object that contains a whole set of localizations

assets/l10n/app_en.arb
```diff
@@ -483,5 +483,24 @@
   "notifSelfUser": "You",
   "@notifSelfUser": {
     "description": "Display name for the user themself, to show after replying in an Android notification"
+  },
+  "onePersonTyping": "{typist} is typing…",
+  "@onePersonTyping": {
+    "description": "Text to display when there is a user typing.",
```
gnprice, line 489: nit:
```suggestion
    "description": "Text to display when there is one user typing.",
```

In the two other cases below, I'd also say "there is a user typing".

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+    final typistNames = model!.typistIdsInNarrow(narrow)
+      .where((id) => id != store.selfUserId)
```
gnprice, line 569: Can we do this filtering on the model instead? That will simplify this build method's code, I think, as well as mean less work for it to do (by helping it skip allocating a new list).

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      .map((id) => store.users[id]?.fullName ?? localization.unknownUserName)
+      .toList();
+    if (typistNames.isEmpty) return placeholder;
+
+    final String text = switch (typistNames.length) {
```
gnprice, line 574: Specifically, once there's no `where` here, these two length checks can happily use the original Iterable instead of the result of the `map`.

Then we turn out not to need the List at all; and if there are 3+ typists in the list, we don't need to go and look up each of their names, which we're not going to use.

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+      1 => localization.onePersonTyping(typistNames[0]),
+      2 => localization.twoPeopleTyping(typistNames[0], typistNames[1]),
+      _ => localization.manyPeopleTyping,
+    };
+
+    return Padding(
+      padding: const EdgeInsetsDirectional.only(start: 16, top: 2),
+      child: Text(text,
+        textAlign: TextAlign.start,
```
gnprice, line 583: Huh, is this different from the default?

test/widgets/message_list_test.dart
```diff
@@ -981,4 +981,70 @@ void main() {
       });
     });
   });
+
+  group('TypingStatus', () {
```
gnprice, line 985: nit: this name looks like an identifier rather than normal English (being in PascalCase), so it should be the actual identifier
```suggestion
  group('TypingStatusWidget', () {
```

gnprice, line 985: separate nit: where possible we try to keep tests in the same order as the code they're testing, to help with navigating between tests and code under test

Here MarkAsReadWidget is already out of order, so moving that would be a good small prep commit. Then this can go next to that.

lib/widgets/message_list.dart
```diff
@@ -530,6 +534,68 @@ class MarkAsReadWidget extends StatelessWidget {
...
+        style: const TextStyle(
+          color: HslColor(0, 0, 53), fontStyle: FontStyle.italic)),
+    );
+  }
+}
+
+class TypingStatusWidget extends StatefulWidget {
```
gnprice, line 590: nit (a thought prompted by looking at the order of the tests): it'd be most logical for this to come just above MarkAsReadWidget, rather than just below — since that matches how it appears in the UI

(That would be the opposite order from their call sites, but that's because the code at the call sites happens to be written in a context where it's counting from the bottom.)

test/widgets/message_list_test.dart
```diff
@@ -981,4 +981,70 @@ void main() {
...
+      await store.handleEvent(event);
+      await tester.pump();
+      check(finder.evaluate()).single.has((x) => x.widget, 'widget').isA<Text>()
+        .data.equals(expected);
+    }
+
+    final dmMessage = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
+    final dmNarrow = DmNarrow.withUsers(
+      users.map((u) => u.userId).toList(),
+      selfUserId: eg.selfUser.userId);
```
gnprice, line 1002: This message doesn't appear in this narrow. For that, the `to` argument should have the other users. (The code below assumes the message is in the narrow, so this affects the realism of the test data.)

Then once you do that, there's a handy constructor `DmNarrow.ofMessage`.

test/widgets/message_list_test.dart
```diff
@@ -981,4 +981,70 @@ void main() {
...
   });
+
+  group('TypingStatus', () {
+    final users = [eg.selfUser, eg.otherUser, eg.thirdUser, eg.fourthUser];
+    final finder = find.descendant(
+      of: find.byType(TypingStatusWidget),
+      matching: find.byType(Text)
+    );
+
+    checkTyping(WidgetTester tester, TypingEvent event, {required String expected}) async {
```
gnprice, line 992: nit: function definition should have explicit return type, so it doesn't look like a function call

(hmm, we should really look to see if there's a lint rule for that, and turn it on — seems eminently lintable)

test/widgets/message_list_test.dart
```diff
@@ -981,4 +981,70 @@ void main() {
...
+      await store.handleEvent(event);
+      await tester.pump();
+      check(finder.evaluate()).single.has((x) => x.widget, 'widget').isA<Text>()
+        .data.equals(expected);
```
gnprice, line 996: `tester.widget` can simplify this:
```suggestion
      check(tester.widget<Text>(finder)).data.equals(expected);
```

gnprice (commented):
Thanks @PIG208 for implementing this, and @chrisbobbe for the previous review!

Generally this looks good. Comments below, in addition to those from Chris above.

assets/l10n/app_en.arb
PIG208: Do you mean "there is one user typing"? Or "there are two users typing" and etc?

lib/widgets/message_list.dart
PIG208: I'm not sure if I understand "these two length checks can happily use the original Iterable instead of the result of the map." here. Do you mean something like this?

```dart
    final String text = switch (typistIds) {
      [final firstTypist] =>
        localizations.onePersonTyping(store.users[firstTypist]?.fullName ?? localizations.unknownUserName),
      [final firstTypist, final secondTypist] =>
        localizations.twoPeopleTyping(
          store.users[firstTypist]?.fullName ?? localizations.unknownUserName,
          store.users[secondTypist]?.fullName ?? localizations.unknownUserName,
        ),
      _ => localizations.manyPeopleTyping,
    };
```

assets/l10n/app_en.arb
gnprice: Ah sorry, that was unclear: the text below is fine for those other cases. What I mean is, I would say all three of these situations are accurately described with "there is a user typing". So the text for this case should be more specific in order to distinguish it.

lib/widgets/message_list.dart
gnprice: Like that except with no `toList`.

Which I think means the `switch` will be on `typistIds.length`, rather than `typistIds` itself.

lib/api/route/events.dart
PIG208: Yes, that's correct.

PIG208:
Thanks for the reviews! Updated the PR and left some questions.

lib/model/typing_status.dart
```diff
@@ -20,7 +20,7 @@ class TypingStatus extends ChangeNotifier {
   Iterable<SendableNarrow> get debugActiveNarrows => _timerMapsByNarrow.keys;
 
   Iterable<int> typistIdsInNarrow(SendableNarrow narrow) =>
-    _timerMapsByNarrow[narrow]?.keys ?? [];
+    _timerMapsByNarrow[narrow]?.keys.where((id) => id != selfUserId) ?? [];
```
gnprice, line 23: Ah, in https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692310354 I didn't just mean doing the filtering in the model code rather than the UI code — I meant doing so in the model data structure itself. That is, just never have an entry here where the key is the self-user.

That way we get to avoid the `where`. Which in turn gets us out of scanning through the whole list if there are more than two.

lib/model/typing_status.dart
```diff
@@ -64,7 +64,7 @@ class TypingStatus extends ChangeNotifier {
   void handleTypingEvent(TypingEvent event) {
     SendableNarrow narrow = switch (event.messageType) {
       MessageType.direct => DmNarrow(
-        allRecipientIds: event.recipientIds!..sort(), selfUserId: selfUserId),
+        allRecipientIds: event.recipientIds!.toList()..sort(), selfUserId: selfUserId),
```
gnprice, line 67: No need for the copy (i.e. this `toList` call). In principle it's not ideal to be mutating the list because in principle something else could be looking at it… but (a) nothing else actually does look at this, because it's a specific feature that this model is entirely responsible for, and (b) really it ought to be sorted anyway, so this is just doing a public service for anything else that might be using the data.

(The ideal version of this would be that we sort at deserialization time.)

lib/api/route/events.dart
gnprice: commit-structure nit: put this commit before the main commit, so that the main commit can be accurate in claiming to fix the issue #665

(otherwise, #665 isn't actually fully working as of that commit)

lib/widgets/message_list.dart
```diff
@@ -458,6 +462,67 @@ class ScrollToBottomButton extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final narrow = widget.narrow;
+    if (narrow is! SendableNarrow) return const SizedBox();
+
+    final localizations = ZulipLocalizations.of(context);
+    final typistIds = model!.typistIdsInNarrow(narrow).toList();
+    if (typistIds.isEmpty) return const SizedBox();
+    final String text = switch (typistIds) {
+      [final firstTypist] =>
```
gnprice, line 507: https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692365513 (a reply I made yesterday to a question after your revision — echoing here just so it remains visible in this review)

test/widgets/message_list_test.dart
```diff
@@ -294,6 +294,233 @@ void main() {
...
+    );
+
+    Future<void> checkTyping(WidgetTester tester, TypingEvent event, {required String expected}) async {
+      await store.handleEvent(event);
+      await tester.pump();
+      check(tester.widget<Text>(finder)).data.equals(expected);
+    }
+
+    final dmMessage = eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]);
+    final dmNarrow = DmNarrow.ofMessage(dmMessage, selfUserId: eg.selfUser.userId);
```
gnprice, line 311: Let's keep test data realistic — so if we're going to have eg.thirdUser and eg.fourthUser typing in this DM narrow, they should be members of it. (following up on https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692326183 )

gnprice (commented):

lib/model/typing_status.dart
PIG208: The DmNarrow.ofMessage factory creates a unmodifiable list, which motivates the toList call here, because we can't mutate the list.

PIG208: I think we might not be using DmNarrow.ofMessage in the application code, so tweaking the tests can be a solution. But it would be helpful if there is a way to statically discover issues like this.

PIG208: Resolved by only sorting them when deserializing TypingEvent, and skipping the sort call from TypingStatus.

lib/widgets/message_list.dart
PIG208: Rewritten that part of code. It ended up looking a bit verbose, but it does save the `.toList` call or unnecessary `.map`'s.

PIG208:
Updated the PR. Thanks!

PIG208:
Rebased to resolve a conflict with 0d577a0376ab534df4ff2ce72cf39665132d128e

PIG208:
pushed to rebase and address https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692364361

assets/l10n/app_en.arb
```diff
@@ -483,5 +483,24 @@
...
+  "onePersonTyping": "{typist} is typing…",
+  "@onePersonTyping": {
+    "description": "Text to display when there is one user typing.",
+    "placeholders": {
+      "typist": {"type": "String", "example": "Alice"}
+    }
+  },
+  "twoPeopleTyping": "{typist} and {otherTypist} are typing…",
+  "@twoPeopleTyping": {
+    "description": "Text to display when there is one user typing.",
```
gnprice, line 496: This description doesn't match the text; it's also the same as the description of a different string, which makes it unhelpful for translators.

I guess https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692364361 left things still confusing? The previous revision was fine on this file.

gnprice (commented):
Hmm, in the current revision it doesn't look like my comments at https://github.com/zulip/zulip-flutter/pull/790#pullrequestreview-2200637627 have been addressed. Perhaps you rebased a different version of the branch than you intended?

assets/l10n/app_en.arb
PIG208: Right, I thought https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692364361 meant to have all three use this same description string.

PIG208:
I think a previous update was lost to a hard reset. Digging my reflog right now.

PIG208:
This should be fixed now.

lib/api/model/events.dart
```diff
@@ -900,16 +901,16 @@ class TypingEvent extends Event {
     required this.recipientIds,
     required this.streamId,
     required this.topic,
-  });
+  }) : assert(isSortedWithoutDuplicates(recipientIds ?? []));
```
gnprice, line 904: nit:
```suggestion
  }) : assert(recipientIds == null || isSortedWithoutDuplicates(recipientIds));
```

I think that's a bit clearer, because in the situation where `recipientIds` is null it's not that that's a way of expressing that the list of recipients is empty — rather it means it doesn't make sense to talk about recipients (because it's a stream/channel message).

lib/api/model/events.dart
```diff
@@ -900,16 +901,16 @@ class TypingEvent extends Event {
...
 
   static Object? _readSenderId(Map<Object?, Object?> json, String key) {
     return (json['sender'] as Map<String, dynamic>)['user_id'];
   }
 
   static List<int>? _recipientIdsFromJson(Object? json) {
     if (json == null) return null;
-    return (json as List<Object?>).map(
-      (item) => (item as Map<String, Object?>)['user_id'] as int).toList();
+    return (json as Iterable<Object?>).map(
```
gnprice, line 912: nit:
```suggestion
    return (json as List<Object?>).map(
```

That keeps this code resembling the generated deserialization code when it comes to boring parts like casting pieces of the data to their expected types. There isn't some expected situation where this would be a non-list iterable, right?

lib/model/typing_status.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:flutter/foundation.dart';
 
 import '../api/model/events.dart';
+import '../log.dart';
```
gnprice, line 6: nit: this happens a commit sooner than it should

(so the linter would warn at that commit)

lib/api/model/events.dart
```diff
@@ -900,16 +901,16 @@ class TypingEvent extends Event {
...
+      (item) => (item as Map<String, Object?>)['user_id'] as int).toList()..sort();
```
gnprice, line 913: commit-message nit:

    events: Sort recipientIds for TypingEvent.

The prefix we've used for changes here is `api:`, so let's stay consistent with that.

A command like `git log --oneline lib/api/model/events.dart` is very handy for looking to see what precedents are for this kind of question.

lib/model/typing_status.dart
```diff
@@ -38,6 +39,10 @@ class TypingStatus extends ChangeNotifier {
   }
 
   bool _addTypist(SendableNarrow narrow, int typistUserId) {
+    if (typistUserId == selfUserId) {
```
gnprice, line 42: commit-message nit:

    model: Do not allow adding self as typist.

For files in lib/model/, we normally make the prefix more specific than this — here, `typing_status:` would be good.

(I'd be tempted to shorten it to `typing:`, but that's too confusing with the concept of the type system and type checker.)

test/widgets/message_list_test.dart
```diff
@@ -843,227 +1130,4 @@ void main() {
         ..status.equals(AnimationStatus.dismissed);
     });
   });
-
-  group('MarkAsReadWidget', () {
```
gnprice, line 847: commit-message nit:

    test [nfc]: Reorder MarkAsReadWidget test group.

Prefix should be more specific: "msglist test [nfc]:".

(Same story as at https://github.com/zulip/zulip-flutter/pull/787#discussion_r1695802626 .)

lib/widgets/message_list.dart
gnprice: Great. It may look a little verbose… but I think in actual line count (for this build method) it's the shortest version yet :slightly_smiling_face: Sometimes just writing things very concretely pays off.

gnprice (commented):
Thanks for the revision! This all looks great — just nits in this round.

PIG208:

gnprice:
Thanks! Looks good — merging.

lib/widgets/message_list.dart
PIG208: Came across https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20mark-as-read, and I think the typing indicators have the same issue of causing a message shift. I wonder if we want to do something similar to #562.

PIG208: The follow-up discussion is here: https://chat.zulip.org/#narrow/stream/48-mobile/topic/messages.20shift.20on.20typing.20indicator/near/1925986

PR #795, opened 2024-07-08
notif: Replace flutter_local_notification `createNotificationChannel` with pigeon

rajveermalviya:
Work towards #351

### DIFF
--- a/lib/host/android_notifications.dart
+++ b/lib/host/android_notifications.dart
@@ -1,5 +1,40 @@
 export './android_notifications.g.dart';
 
+/// For use in [NotificationChannel.importance].
+///
+/// See:
+///   https://developer.android.com/reference/android/app/NotificationChannel#setImportance(int)
+///   https://developer.android.com/reference/android/app/NotificationChannel#getImportance()
+abstract class NotificationImportance {
+  /// Corresponds to `IMPORTANCE_UNSPECIFIED`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_UNSPECIFIED()
+  static const unspecified = -1000;
+
+  /// Corresponds to `IMPORTANCE_NONE`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_NONE()
+  static const none = 0;
+
+  /// Corresponds to `IMPORTANCE_MIN`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_MIN()
+  static const min = 1;
+
+  /// Corresponds to `IMPORTANCE_LOW`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_LOW()
+  static const low = 2;
+
+  /// Corresponds to `IMPORTANCE_DEFAULT`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_DEFAULT()
+  static const default_ = 3;
+
+  /// Corresponds to `IMPORTANCE_HIGH`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_HIGH()
+  static const high = 4;
+
+  /// Corresponds to `IMPORTANCE_MAX`:
+  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_MAX()
+  static const max = 5;
+}
+
 /// For use in [PendingIntent.flags].
 ///
 /// See: https://developer.android.com/reference/android/app/PendingIntent#constants_1
--- a/pigeon/notifications.dart
+++ b/pigeon/notifications.dart
@@ -8,6 +8,34 @@ import 'package:pigeon/pigeon.dart';
   kotlinOptions: KotlinOptions(package: 'com.zulip.flutter'),
 ))
 
+/// Corresponds to `androidx.core.app.NotificationChannelCompat`
+///
+/// See: https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat
+class NotificationChannel {
+  /// Corresponds to `androidx.core.app.NotificationChannelCompat.Builder`
+  ///
+  /// See: https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat.Builder
+  NotificationChannel({
+    required this.id,
+    required this.importance,
+    this.name,
+    this.lightsEnabled,
+    this.vibrationPattern,
+  });
+
+  final String id;
+
+  /// Specifies the importance level of notifications
+  /// to be posted on this channel.
+  ///
+  /// Must be a valid constant from [NotificationImportance].
+  final int importance;
+
+  final String? name;
+  final bool? lightsEnabled;
+  final Int64List? vibrationPattern;
+}
+
 /// Corresponds to `android.app.PendingIntent`.
 ///
 /// See: https://developer.android.com/reference/android/app/PendingIntent
@@ -96,6 +124,11 @@ class MessagingStyle {
 
 @HostApi()
 abstract class AndroidNotificationHostApi {
+  /// Corresponds to `androidx.core.app.NotificationManagerCompat.createNotificationChannel`.
+  ///
+  /// See: https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#createNotificationChannel(androidx.core.app.NotificationChannelCompat)
+  void createNotificationChannel(NotificationChannel channel);
+
   /// Corresponds to `android.app.NotificationManager.notify`,
   /// combined with `androidx.core.app.NotificationCompat.Builder`.
   ///
--- a/test/model/binding.dart
+++ b/test/model/binding.dart
@@ -578,10 +578,20 @@ class FakeIOSFlutterLocalNotificationsPlugin extends Fake implements IOSFlutterL
 }
 
 class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
+  /// Consume the log of calls made to [createNotificationChannel].
+  ///
+  /// This returns a list of the arguments to all calls made
+  /// to [createNotificationChannel] since the last call to this method.
+  List<NotificationChannel> takeCreatedChannels() {
+    final result = _createdChannels;
+    _createdChannels = [];
+    return result;
+  }
+  List<NotificationChannel> _createdChannels = [];
+
   @override
-  Future<void> createNotificationChannel(NotificationChannel channel) {
-    // TODO: implement createNotificationChannel
-    throw UnimplementedError();
+  Future<void> createNotificationChannel(NotificationChannel channel) async {
+    _createdChannels.add(channel);
   }
 
   /// Consume the log of calls made to [notify].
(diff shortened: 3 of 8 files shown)


pigeon/notifications.dart
```diff
@@ -77,4 +116,9 @@ abstract class AndroidNotificationHostApi {
     // NotificationCompat.Builder has lots more methods; add as needed.
     // Keep them alphabetized, for easy comparison with that class's docs.
   });
+
+  /// Corresponds to `androidx.core.app.NotificationManagerCompat.createNotificationChannel`.
+  ///
+  /// See: https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#createNotificationChannel(androidx.core.app.NotificationChannelCompat)
+  void createNotificationChannel(NotificationChannel channel);
```
sm-sayedi, line 123: I am not sure if this could return `Future<void>` instead of `void` in here too, as the autogenerated concrete class has a return type of `Future<void>`. Looking at the docs from the link provided, it seems that this method has a return type of `void`, but that's in Java, and it doesn't have a type `Future`; at least that's how I know so far about Java. 😀

sm-sayedi (commented):
Thanks @rajveermalviya for this! Just a small comment below, otherwise LGTM!

Let's move forward to the mentor review with @sumanthvrao!

pigeon/notifications.dart
rajveermalviya: This file is a Pigeon file, not a standard Dart file. It contains types and methods that define the native plugin API interface. See [Pigeon's example](https://github.com/flutter/packages/blob/main/packages/pigeon/example/README.md#dart-input).

Adding a `Future<void>` here will result in an error:
```
Error: pigeon/notifications.dart:120: Unknown type: Future
```

sm-sayedi: Ok thanks, got it!

kenclary (approved):
this looks fine to me, but I have basically no familiarity with pigeon.

lib/host/android_notifications.dart
```diff
@@ -25,3 +25,29 @@ abstract class PendingIntentFlag {
   /// Corresponds to `FLAG_ALLOW_UNSAFE_IMPLICIT_INTENT`.
   static const allowUnsafeImplicitIntent = 1 << 24;
 }
+
+/// For use in [NotificationChannel.importance].
+///
+/// See: https://developer.android.com/reference/android/app/NotificationChannel#setImportance(int)
+abstract class NotificationImportance {
+  /// Corresponds to [`IMPORTANCE_UNSPECIFIED`](https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_UNSPECIFIED())
```
chrisbobbe, line 33: The `[text](example.com)` syntax makes it a bit harder to work out what the intended URL is and select it, especially here where the intended URL ends with ")".

I did notice Android Studio helpfully trying to linkify this in a hover interaction:

image

but then it similarly gets confused about what the intended URL is; it takes me here: https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_UNSPECIFIED(

when it should have taken me here: https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_UNSPECIFIED()

so the browser just leaves me at the top of the page, instead of scrolling me down to the `#IMPORTANCE_UNSPECIFIED()` element.

So I think probably most helpful to give up that syntax and just write some thing like:

```dart
  /// Corresponds to `IMPORTANCE_UNSPECIFIED`:
  ///   https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_UNSPECIFIED()
```

(here and elsewhere in the PR)

chrisbobbe (commented):
Thanks! Small comments below, and now over to Greg for his review.

rajveermalviya:
Thanks for the review @chrisbobbe, pushed a new revision PTAL :)

pigeon/notifications.dart
```diff
@@ -94,8 +94,39 @@ class MessagingStyle {
   final bool isGroupConversation;
 }
 
+/// Corresponds to `androidx.core.app.NotificationChannelCompat`
+///
+/// See: https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat
+class NotificationChannel {
+  /// Corresponds to [`NotificationChannelCompat.Builder`](https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat.Builder#Builder(java.lang.String,int))
```
gnprice, line 101: nit: one remaining instance of the pattern @chrisbobbe pointed out above

gnprice: Separately, the reference to the builder class is good but I think it's most accurate to link to the class as a whole rather than that constructor. Three of the five parameters here don't go to the builder's constructor, but rather to method calls on the builder.

lib/host/android_notifications.dart
```diff
@@ -25,3 +25,36 @@ abstract class PendingIntentFlag {
   /// Corresponds to `FLAG_ALLOW_UNSAFE_IMPLICIT_INTENT`.
   static const allowUnsafeImplicitIntent = 1 << 24;
 }
+
+/// For use in [NotificationChannel.importance].
+///
+/// See: https://developer.android.com/reference/android/app/NotificationChannel#setImportance(int)
```
gnprice, line 31:
```suggestion
/// See:
///   https://developer.android.com/reference/android/app/NotificationChannel#setImportance(int)
///   https://developer.android.com/reference/android/app/NotificationChannel#getImportance()
```

The getter's doc has some additional information that isn't on the setter, and they don't link to each other, so best to provide both.

pigeon/notifications.dart
```diff
@@ -94,8 +94,39 @@ class MessagingStyle {
   final bool isGroupConversation;
 }
 
+/// Corresponds to `androidx.core.app.NotificationChannelCompat`
+///
+/// See: https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat
+class NotificationChannel {
```
gnprice, line 100: nit: put this at the top of the file, before PendingIntent — that makes a nice logical lifecycle-oriented order

The methods on AndroidNotificationHostApi are already good this way.

The enum class NotificationImportance should get the same treatment, coming before PendingIntentFlag.

test/model/binding.dart
```diff
@@ -474,27 +474,23 @@ class FakeFlutterLocalNotificationsPlugin extends Fake implements FlutterLocalNo
 }
 
 class FakeAndroidFlutterLocalNotificationsPlugin extends Fake implements AndroidFlutterLocalNotificationsPlugin {
+}
+
+class FakeIOSFlutterLocalNotificationsPlugin extends Fake implements IOSFlutterLocalNotificationsPlugin {
+}
```
gnprice, line 480: Neat!

Does this mean we can also cut out the test binding's `resolvePlatformSpecificImplementation`? That'd be a nice further cleanup commit to come after this one.

test/model/binding.dart
```diff
@@ -513,6 +509,11 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
     _activeNotificationsMessagingStyle.clear();
   }
 
+  @override
+  Future<void> createNotificationChannel(NotificationChannel channel) async {
```
gnprice, line 513: nit: keep this next to the data it acts on (so `_createdChannels`)

gnprice: Relatedly, in the first commit this should go above `takeNotifyCalls`, so as not to separate that from `notify`.

gnprice (commented):

Generally this looks great — a few nits below.

rajveermalviya:
Thanks for the review @gnprice, pushed a new revision.

pigeon/notifications.dart
```diff
@@ -8,6 +8,33 @@ import 'package:pigeon/pigeon.dart';
   kotlinOptions: KotlinOptions(package: 'com.zulip.flutter'),
 ))
 
+/// Corresponds to `androidx.core.app.NotificationChannelCompat`
+///
+/// See:
+///   https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat
+///   https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat.Builder#Builder(java.lang.String,int)
+class NotificationChannel {
+  NotificationChannel({
```
gnprice, line 17: Ah, just to clarify what I meant at https://github.com/zulip/zulip-flutter/pull/795#discussion_r1684946174 because it might have been confusing:

I think having a doc comment on this Pigeon/Dart constructor is fine and potentially helpful. What I meant was that the doc on this constructor should link to the doc of the upstream builder class as a whole, and not to the specific point on that page for the builder's constructor.

So like this:

```suggestion
/// See:
///   https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat
///   https://developer.android.com/reference/androidx/core/app/NotificationChannelCompat.Builder
class NotificationChannel {
  NotificationChannel({
```

but also that second link would be fine to move back to being a doc comment on this NotificationChannel constructor rather than on the NotificationChannel class.

test/model/binding.dart
```diff
@@ -513,34 +512,6 @@ class FakeFlutterLocalNotificationsPlugin extends Fake implements FlutterLocalNo
     return true;
   }
 
-  FlutterLocalNotificationsPlatform? _platform;
-
-  @override
-  T? resolvePlatformSpecificImplementation<T extends FlutterLocalNotificationsPlatform>() {
```
gnprice, line 519: commit-structure nit: this should come as a separate commit after the main commit, as per https://github.com/zulip/zulip-flutter/pull/795#discussion_r1684959410

gnprice (commented):
Thanks for the revision! Two nits, and otherwise this looks great.

rajveermalviya:

test/model/binding.dart
```diff
@@ -557,27 +528,28 @@ class FakeFlutterLocalNotificationsPlugin extends Fake implements FlutterLocalNo
 }
 
 class FakeAndroidFlutterLocalNotificationsPlugin extends Fake implements AndroidFlutterLocalNotificationsPlugin {
+}
+
+class FakeIOSFlutterLocalNotificationsPlugin extends Fake implements IOSFlutterLocalNotificationsPlugin {
```
gnprice, line 533: nit: these can be cut too, since they're effectively part of the implementation of `resolvePlatformSpecificImplementation`

(the previous revision did cut them, so I'm guessing their coming back was a rebase error)

test/model/binding.dart
gnprice, line 519: nit in commit message:

    notif [nfc]: Remove unused test harness for flutter_local_notification

This sounds like it might mean our whole test harness for that package. Also the package's name has an "s" at the end. So:

    notif [nfc]: Remove unused test harness for parts of flutter_local_notifications

gnprice (commented):
Thanks for the revision! Looks good — merging, after fixing two quick nits mentioned below.

This is great progress on #351; looking forward to seeing the remaining pieces.

gnprice:
Oh and a different commit-message nit — the commit series looks like this:
1f641b7c7 notif: Add createNotificationChannel method to pigeon bindings
2af238ba4 notif: Add createNotificationChannel method to pigeon bindings
b35b73c28 notif [nfc]: Remove unused test harness for parts of flutter_local_notifications

So the first commit message is duplicated.

I think you intended the second commit to have the message it had in the last revision:

    notif: Replace flutter_local_notification `createNotificationChannel` with pigeon
    
    Updates #351

So I'll put that message back.

This is probably also a good moment for a reminder that I highly recommend setting up and using [a graphical Git client](https://github.com/zulip/zulip-mobile/blob/2217c858e207f9f092651dd853051843c3f04422/docs/howto/git.md#graphical-client), like `gitk`. Especially when doing a rebase, or operations like splitting up commits, a graphical client can be extremely helpful for getting a clear picture of what the old and new branch look like. I think if you'd used `gitk` or similar when self-reviewing your revision, you'd probably have immediately spotted that the commit message on the second commit had gotten clobbered.

PR #796, opened 2024-07-08
Do Android build in CI

### LINKED ISSUE #772: Run Android build in CI
This would help us more quickly catch issues where the build gets broken, like:
* #771 

It'd also be helpful for when we're making changes to our Android-side code (including our Kotlin source code and our Android build setup).

Once we have this in our own CI, we should also (and I'll file a separate issue for this) see about putting it into Flutter's customer tests (like we've already done with `flutter analyze` and `flutter test`, as #239). That way, an issue like #771 would be caught at an earlier stage — the PR that made that change in the Flutter repo would have already seen a test failure. At that point, running it in our own CI on Flutter upgrades should be a bit redundant — but it'll still be valuable for upgrading other dependencies, and for our own changes.

gnprice:
Fixes #772.

(Started as a draft PR, because I wasn't sure if more setup was needed in ci.yml in order for this build to work. Indeed some was.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4.1.7
 
+    - name: Set up JDK
+      uses: actions/setup-java@v4.2.1
+      with:
+        java-version: 17
+        distribution: temurin
+
     - name: Clone Flutter SDK
       # We can't do a depth-1 clone, because we need the most recent tag
       # so that Flutter knows its version and sees the constraint in our
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4.1.7
 
     - name: Clone Flutter SDK
       # We can't do a depth-1 clone, because we need the most recent tag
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Thanks, LGTM! I'll hold off merging until @rajveermalviya has a chance to look.

gnprice:
The other day I remarked:
https://github.com/zulip/zulip-flutter/pull/718#discussion_r1667976403
that when we handle #772, we should include running the linter. It turns out that the type of issue that we were discussing there was a warning from the Kotlin compiler — so the last commit in this PR, telling the Kotlin compiler to treat those warnings as errors, does the job.

There's also an Android linter, but it has a different job. I'll send a separate PR for running that.

rajveermalviya (approved):
Thanks for working on this @gnprice, LGTM!

I tried this with macOS M1 runner (`macos-latest`), but surprisingly it didn't make much difference: https://github.com/rajveermalviya/zulip-flutter/actions/runs/9852479999

gnprice:
> I tried this with macOS M1 runner (`macos-latest`), but surprisingly it didn't make much difference: https://github.com/rajveermalviya/zulip-flutter/actions/runs/9852479999

Thanks for trying that. Yeah, it's a shame that this adds so much to the time CI requires — about 4 minutes. It'd be good to find a way to reduce that, but I'm not sure there will be a good way; I think building an Android app is just slow compared to what we do in `flutter test`, and the worker machines we get from GitHub Actions are low-powered, and so on net it's slow.

PR #802, opened 2024-07-11
docs: Create setup guide, Add guide to setup on a cloud server

Khader-1:


gnprice (commented):
Thanks!

It looks like most of this document covers information that's in the README, or in upstream Flutter docs that the README links to, or in the Zulip project's general developer docs (like on using Git and creating PRs).

Let's not spend time trying to write up new comprehensive docs covering that material. I think what we already have covers that pretty well; if there are specific gaps we notice, we can fill those in individually.

The key thing that would be useful for you to write down is some notes just jotting down the steps you took for setting up a dev environment in a cloud server. And then as I wrote [in your check-ins thread](https://chat.zulip.org/#narrow/stream/14-GSoC/topic/Khader.20-.20Check-ins/near/1839803):
> Don't worry about spending time polishing the text for that guide; some rough notes that just list the steps you took will be great. In the future if/when other people have the same need, we can point them at those notes, and we can spend time polishing it then based on their experience using the notes.

In particular I think the DM you sent me a few weeks ago already contains basically all the information that's needed. So this `docs/setup.md` can look like:

```markdown
# Dev setup

For general instructions on setting up to do development on this Flutter-based Zulip mobile app, see the README.

This file covers specific topics in more detail.

## On a cloud server

[… then your instructions here, based on that DM …]
```

except make "see the README" into a link.

rajveermalviya:
@Khader-1 is this ready for review?
I've been assigned for the buddy review but PR is still in Draft.

Khader-1:
It's indeed @rajveermalviya. I remember marking it as ready but I see it's not reflected for some reason!

rajveermalviya (changes requested):
Thanks for creating this guide @Khader-1.

Assuming the goal is (since it's not really clear currently) for running `flutter run` on a non-graphical remote vm and emulator on the host machine, I tried to do this setup with Ubuntu Server image in a VM and found that there are many details missing in the current proposed doc, especially for connecting the emulator.

Here's a draft document of all the steps I carried out: https://gist.github.com/rajveermalviya/5b16d0cd332f1f17abea326284cb0c07

Also I'd expect there to be a (interactive) script which possibly only works with specific distro and specifc version, to ease the setup considerably, and then this doc could just point to that script. This should probably be done later though.

gnprice:
Thanks @Khader-1 for writing this down, and @rajveermalviya for the thorough review!

> Also I'd expect there to be a (interactive) script […] and then this doc could just point to that script. This should probably be done later though.

Yeah, agreed on both counts — that's something we'd want to have if it were a common setup, and it's something I wouldn't want to invest in now. As I alluded to above, the idea with this PR was just to note down the things that Khader already had to work out in the course of setting up a cloud server so that he could do development there (to get better network connectivity than he has locally — particularly for Gradle, which downloads a ton of different dependencies).

I've taken the first half of @rajveermalviya's instructions from that gist, and combined them with some of Khader's text and with text of my own framing it all. The result is instructions for setting up for doing Android builds on a non-graphical remote machine — basically for doing so without Android Studio.

Then I'm not sure what the right answer is for actually running the app. One thing one can definitely do is `flutter build apk`, and download the APK to install and run it; the main thing you lose that way is hot reload, as well as other debugging features like the inspector.

@Khader-1 is `flutter build apk` the approach you've been taking? Or have you found a way to use `flutter run` with the cloud VM? If so, are you using something like the steps in the second half of Rajesh's instructions?

README.md
```diff
@@ -92,13 +92,23 @@ Two specific points to expand on:
...
 1. Follow the [Flutter installation guide](https://docs.flutter.dev/get-started/install)
    for your platform of choice.
 2. Switch to the latest version of Flutter by running `flutter channel main`
    and `flutter upgrade` (see [Flutter version](#flutter-version) below).
 3. Ensure Flutter is correctly configured by running `flutter doctor`.
 4. Start the app with `flutter run`, or from your IDE.
 
+Parts of our test suite require an additional dependency:
+
+5. Install SQLite, for example by running `sudo apt install libsqlite3-dev`.
```
gnprice, line 107: @Khader-1 The previous draft called for installing `sqlite3` too. I *think* that shouldn't be needed, though I'm not certain of that — that package is mainly just the `sqlite3` command-line binary:
```
$ dpkg -L sqlite3
/.
/usr
/usr/bin
/usr/bin/sqldiff
/usr/bin/sqlite3
/usr/share
/usr/share/doc
/usr/share/doc/sqlite3
/usr/share/doc/sqlite3/changelog.Debian.gz
/usr/share/doc/sqlite3/changelog.gz
/usr/share/doc/sqlite3/changelog.html.gz
/usr/share/doc/sqlite3/copyright
/usr/share/man
/usr/share/man/man1
/usr/share/man/man1/sqlite3.1.gz
```

(So specifically it's just that, and the `sqldiff` command-line tool which I hadn't heard of until just now, and some documentation. At least that's what it is on my system.)

gnprice:
I'm going to go ahead and merge this revised version, because it seems like clearly an improvement and we can always edit it further later.

If there is more to say about the workflow that @Khader-1's already worked out, then I'd be glad to see that added to the doc. But we don't need to invest in figuring out workflows for the sake of writing these instructions — there are plenty of other things for us to do :slightly_smiling_face: and we can always return to this area if we find more need for it in the future.

PR #804, opened 2024-07-11
theme: Implement DesignVariables.dark

chrisbobbe:

PR #805, opened 2024-07-12
emoji_reaction: Add EmojiReactionTheme, including dark variant

chrisbobbe:
Screenshot (also showing some unrelated tweaks toward dark-theme that will land separately):

image

I thought it might be helpful to have a new `ThemeExtension` for this part of the UI instead of adding all these values to `DesignVariables`. This way at least we don't have to alphabetize these items along with a lot of attributes from unrelated parts of the UI. I see that we're taking the `DesignVariables` approach in #762, though, and it's possible that that'll end up being preferred for some reason.

Related: #95


gnprice:
Huh, when I run tests locally at the tip of this, they fail — I think on exactly the RTL half of the cases:
```
$ flutter test test/widgets/emoji_reaction_test.dart 
00:02 +0: ReactionChipsList smoke (same reaction, different users, with one unknown user): show names when few / text / rtl / text scale: 0.8235
Expected: a TextDirection that:
  equals <TextDirection.rtl>
Actual: <TextDirection.ltr>
Which: are not equal

#0      check.<anonymous closure> (package:checks/src/checks.dart:85:9)
#1      _TestContext.expect (package:checks/src/checks.dart:708:12)
#2      CoreChecks.equals (package:checks/src/extensions/core.dart:90:13)
#3      main.<anonymous closure>.setupChipsInBox (file:///home/greg/z/flutterz/test/widgets/emoji_reaction_test.dart:74:51)
<asynchronous suspension>
#4      main.<anonymous closure>.runSmokeTest.<anonymous closure> (file:///home/greg/z/flutterz/test/widgets/emoji_reaction_test.dart:132:17)
[…]
```

That begins at the "Set up whole ZulipApp" commit. It persists even after I run `flutter pub get` (which I believe should be redundant, I believe `flutter test` does that).

Clearly it works for you and in CI. So presumably there's something that's getting cached for me locally — like the list of locales, perhaps, so that it doesn't know about `ar`.

Persists also if I run `flutter build apk --config-only`, which I would have expected to take care of generating any files it was going to generate.

chrisbobbe:
What does `ZulipLocalizations.supportedLocales` look like for you?

gnprice:
`[en, ja]`

chrisbobbe:
Hmm, mine was auto-updating to include `ar` with `flutter run`.

gnprice:
Cool, yeah, `flutter run` does it.

I think that probably means `flutter gen-l10n` will do it. Ah indeed, and our docs say this:
> Then run the app (with `flutter run` or in your IDE),
> or perform a hot reload,
> to cause the Dart bindings to be updated based on your
> changes to the ARB file.
> (You can also trigger an update directly, with `flutter gen-l10n`.)

I guess we'll have to warn everyone to do that after this gets merged.

gnprice:
(Probably this should be thought of as a bug in the `flutter` tool.)

test/widgets/emoji_reaction_test.dart
```diff
@@ -39,30 +42,37 @@ void main() {
...
+      final locale = switch (textDirection) {
+        TextDirection.ltr => const Locale('en'),
+        TextDirection.rtl => const Locale('ar'),
+      };
+      tester.platformDispatcher.localeTestValue = locale;
+      tester.platformDispatcher.localesTestValue = [locale];
+      addTearDown(tester.platformDispatcher.clearLocaleTestValue);
+      addTearDown(tester.platformDispatcher.clearLocalesTestValue);
+
+      await tester.pumpWidget(const ZulipApp());
```
gnprice, line 59: This commit turns out to make these tests quite a bit slower:
```
  # before:
$ time flutter test test/widgets/emoji_reaction_test.dart
00:05 +160: All tests passed!                                                            

time: 6.078s wall (7.313s u, 0.940s s)


  # after:
$ time flutter test test/widgets/emoji_reaction_test.dart
00:09 +160: All tests passed!                                                            

time: 10.528s wall (12.424s u, 1.321s s)
```

The slowdown happens with this commit:
fa3f012e9 emoji_reaction test: Set up whole ZulipApp, not just MaterialApp

It'd be good to try to pin down where that time is going. The strategy of reusing ZulipApp in tests makes a lot of sense in principle, so we're likely to try using it in more and more tests… and I'd like to figure out how to do so without slowing all our tests down by 75%.

The strategy I'd try is to just start commenting out parts of ZulipApp's functionality, and rerunning this one test file, and see which pieces are contributing the time.

Then once we know which pieces are at issue, we can see about finding hacks for tests to cut those out, or strip them down to a minimum.

gnprice (commented):
(Haven't yet read the main commit at the end.)

gnprice (commented):
Thanks @chrisbobbe! Generally this looks good.
* The same two small comments as at #804 apply here.
* Let's also investigate that test slowdown above, and see if we can find an easy workaround for it.

And then repeating here just so the action items are in one place:
* On merging this, we'll need to make a PSA in #mobile-team that people need to run `flutter gen-l10n` (or just `flutter run`) after rebasing in order to get tests working again.
  * And as a bonus followup it'd be nice to find or file an upstream flutter-tool issue about the need for that.

gnprice:
> I thought it might be helpful to have a new `ThemeExtension` for this part of the UI instead of adding all these values to `DesignVariables`. This way at least we don't have to alphabetize these items along with a lot of attributes from unrelated parts of the UI. I see that we're taking the `DesignVariables` approach in #762, though, and it's possible that that'll end up being preferred for some reason.

Yeah, I'm not sure. Happy to have both approaches going, and we can see which we like better and optionally refactor later to a single approach.

chrisbobbe:

lib/widgets/emoji_reaction.dart
```diff
@@ -8,6 +8,104 @@ import 'content.dart';
 import 'store.dart';
 import 'text.dart';
 
+/// Emoji-reaction styles that differ between light and dark themes.
+class EmojiReactionTheme extends ThemeExtension<EmojiReactionTheme> {
+  factory EmojiReactionTheme.light() {
+    return EmojiReactionTheme._(
```
gnprice, line 14: bump on redirecting constructor (one of those two #804 comments I mentioned)

test/widgets/test_app.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:zulip/widgets/store.dart';
+import 'package:zulip/widgets/theme.dart';
+
+/// A lightweight mock of [ZulipApp], suitable for most widget tests.
+class TestZulipApp extends StatelessWidget {
+  const TestZulipApp({
+    super.key,
+    this.accountId,
+    this.navigatorObservers,
+    this.child = const Placeholder(),
```
gnprice, line 13: nit: this final ordering is good but the reordering should get squashed into the commit that introduces the widget

test/widgets/text_test.dart
```diff
@@ -397,13 +399,9 @@ void main() {
...
 
-        await tester.pumpWidget(const ZulipApp());
-        await tester.pump();
-
-        final navigator = await ZulipApp.navigator;
-        navigator.push(MaterialWidgetRoute(page: Builder(builder: (context) =>
+        await tester.pumpWidget(TestZulipApp(child: Builder(builder: (context) =>
           Text('123', style: TextStyle(textBaseline: localizedTextBaseline(context))))));
-        await tester.pumpAndSettle();
+        await tester.pump();
```
gnprice, line 404: Nice that a lot of `pumpAndSettle` calls seem to become unnecessary in favor of just `pump`.

gnprice (commented):
Thanks! Here's a partial review before I switch to another task.

Small comments below, and I tried out the performance effects of these two commits:
0a2b459b4 test: Add TestZulipApp; replace uses of ZulipApp
c026fce27 test: Use TestZulipApp everywhere else it makes sense

on `flutter test test/widgets` on my machine, measuring the fastest of three runs.

Before:
time: 15.837s wall (68.737s u, 9.182s s)

After replacing ZulipApp uses:
time: 15.389s wall (64.851s u, 8.641s s)

After using TestZulipApp more widely:
time: 15.684s wall (64.010s u, 9.070s s)

So replacing ZulipApp made a small speedup, about 7% less total CPU time. Then using TestZulipApp more widely had roughly zero impact, as hoped.

I haven't yet read the last four commits (or revisions to them, at least) in detail:
c026fce27 test: Use TestZulipApp everywhere else it makes sense
5967ff01c emoji_reaction test [nfc]: Move some setup code out next to similar code
309cff541 emoji_reaction test: Add another convenient test-setup check
7dd90d185 emoji_reaction: Add EmojiReactionTheme, including dark variant

gnprice:
Noting here to help us remember it: #827 introduces another spot that could benefit from TestZulipApp:
https://github.com/zulip/zulip-flutter/pull/827#discussion_r1684912782

so whichever of these PRs lands first, we can update that spot in the other PR.

test/widgets/emoji_reaction_test.dart
```diff
@@ -39,30 +42,24 @@ void main() {
...
-      double? width,
-      TextDirection? textDirection,
+      double width = 245.0, // (seen in context on an iPhone 13 Pro)
     }) async {
       final message = eg.streamMessage(reactions: reactions);
 
-      await tester.pumpWidget(
-        MaterialApp(
-          home: Directionality(
-            textDirection: textDirection ?? TextDirection.ltr,
```
gnprice, line 50: nit: currently these get turned into the `tester.platformDispatcher` logic in the big conversion commit:
c026fce27 test: Use TestZulipApp everywhere else it makes sense

Probably that'd be better as a separate commit coming after that one, since it's different from the otherwise-repetitive other changes in the commit.

test/widgets/text_test.dart
```diff
@@ -96,11 +95,13 @@ void main() {
...
       testWidgets(description, (WidgetTester tester) async {
-        await tester.pumpWidget(
-          MaterialApp(
-            home: MediaQuery(
-              data: MediaQueryData(boldText: platformRequestsBold),
-              child: Builder(builder: (context) => Text('', style: styleBuilder(context))))));
+        addTearDown(testBinding.reset);
+        tester.platformDispatcher.accessibilityFeaturesTestValue =
+          FakeAccessibilityFeatures(boldText: platformRequestsBold);
+        addTearDown(tester.platformDispatcher.clearAccessibilityFeaturesTestValue);
```
gnprice, line 101: This seems fine to include in the same commit, though. It's fairly straightforwardly just another way of expressing the effect of that `MediaQuery`.

test/widgets/theme_test.dart
```diff
@@ -24,23 +23,19 @@ void main() {
...
-                0.01, baseFontSize: expectedFontSize);
-              return button;
-            }))));
-
-        final text = tester.renderObject<RenderParagraph>(find.text(buttonText)).text;
+        await tester.pumpWidget(TestZulipApp(child: button));
+        await tester.pump();
+        final buttonTextFinder = find.text(buttonText);
+        final context = tester.element(buttonTextFinder);
+        final text = tester.renderObject<RenderParagraph>(buttonTextFinder).text;
```
gnprice, line 35: small simplification:
```suggestion
        final text = (context.renderObject as RenderParagraph).text;
```

Then also `buttonTextFinder` is used just once and can be inlined.

gnprice: also nit: reorder this line to just before the `check`, after computing the expected values; or perhaps even just inline it into the `check` call

test/widgets/emoji_reaction_test.dart
```diff
@@ -204,6 +215,45 @@ void main() {
         }
       }
     }
+
+    testWidgets('Smoke test for light/dark/lerped', (tester) async {
```
gnprice, line 219: nit: nested in unintended group

test/widgets/emoji_reaction_test.dart
```diff
@@ -204,6 +215,45 @@ void main() {
...
+      await prepare();
+      await store.addUsers([eg.selfUser, eg.otherUser]);
+
+      debugFollowPlatformBrightness = true;
```
gnprice, line 223: needs teardown

gnprice (commented):
OK, and here's the rest of a review on this revision. Generally looks great — several nits, and two other comments on tests which should be quick to fix.

test/widgets/emoji_reaction_test.dart
chrisbobbe: The 'ReactionChipsList' group? It uses some helpers that are in that group. Those could be moved out of the group, though.

gnprice: Yeah, this test didn't seem to be about the ReactionChipsList widget.

I guess that is the widget that test helper is setting up, though, so maybe the grouping is appropriate.

chrisbobbe:

gnprice:
Thanks! Looks good — merging.

PR #807, opened 2024-07-12
msglist: Support @-mentions narrow

### LINKED ISSUE #250: msglist: Offer @-mentions narrow
Like the "Mentions" narrow the web app offers at the top of the left sidebar.

We should support this narrow in the message list, and then provide a way to navigate to it.

If we do this before building a helpful overall navigation for the app, the way to navigate to it can be just another button on HomePage.

This is a sort of followup to #72.

PIG208:
Early prototype to support @-mentions narrow.

TODO:

- [x] Organize the commits
- [x] <del>Evaluate ways to support `wildcardMetnioned`, `channelWildcardMentioned` and `topicWildcardMentioned` with backwards/forward compatibility.</del> Deferred to #817
- [x] Test cases

Fixes #250

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -472,6 +472,10 @@
   "@combinedFeedPageTitle": {
     "description": "Title for the page of combined feed."
   },
+  "mentionsPageTitle": "Mentions",
+  "@mentionsPageTitle": {
+    "description": "Title for the page of @-mentions."
+  },
   "notifGroupDmConversationLabel": "{senderFullName} to you and {numOthers, plural, =1{1 other} other{{numOthers} others}}",
   "@notifGroupDmConversationLabel": {
     "description": "Label for a group DM conversation notification.",
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -222,6 +222,7 @@ class MentionAutocompleteView extends ChangeNotifier {
       case DmNarrow():
         break;
       case CombinedFeedNarrow():
+      case MentionsNarrow():
         assert(false, 'No compose box, thus no autocomplete is available in ${narrow.runtimeType}.');
     }
     return (userA, userB) => _compareByRelevance(userA, userB,
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -154,6 +154,7 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
   ApiNarrowStream? streamElement;
   ApiNarrowTopic? topicElement;
   ApiNarrowDm? dmElement;
+  ApiNarrowIsMentioned? isMentionedElement;
 
   for (var i = 0; i < segments.length; i += 2) {
     final (operator, negated) = _parseOperator(segments[i]);
@@ -181,6 +182,10 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
         if (dmIds == null) return null;
         dmElement = ApiNarrowDm(dmIds, negated: negated);
 
+      case _NarrowOperator.is_:
+        if (isMentionedElement != null) return null;
+        if (operand == 'mentioned') isMentionedElement = ApiNarrowIsMentioned();
+
       case _NarrowOperator.near: // TODO(#82): support for near
       case _NarrowOperator.with_: // TODO(#683): support for with
         continue;
@@ -190,7 +195,10 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
     }
   }
 
-  if (dmElement != null) {
+  if (isMentionedElement != null) {
+    if (streamElement != null || topicElement != null || dmElement != null) return null;
+    return const MentionsNarrow();
+  } else if (dmElement != null) {
     if (streamElement != null || topicElement != null) return null;
     return DmNarrow.withUsers(dmElement.operand, selfUserId: store.selfUserId);
   } else if (streamElement != null) {
@@ -212,6 +220,9 @@ enum _NarrowOperator {
   // cannot use `with` as it is a reserved keyword in Dart
   @JsonValue('with')
   with_,
+  // cannot use `is` as it is a reserved keyword in Dart
+  @JsonValue('is')
+  is_,
   pmWith,
   stream,
   channel,
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -85,6 +85,8 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
         fragment.write('${element.operand.join(',')}-$suffix');
       case ApiNarrowDm():
         assert(false, 'ApiNarrowDm should have been resolved');
+      case ApiNarrowIsMentioned():
+        fragment.write(element.operand.toString());
       case ApiNarrowIsUnread():
         fragment.write(element.operand.toString());
       case ApiNarrowMessageId():
--- a/lib/model/narrow.dart
+++ b/lib/model/narrow.dart
@@ -291,17 +291,21 @@ class MentionsNarrow extends Narrow {
 
   @override
   bool containsMessage(Message message) {
-    return message.flags.any((flag) =>
+    return message.flags.any((flag) {
       switch (flag) {
-        MessageFlag.mentioned => true,
-        MessageFlag.wildcardMentioned => true,
-        MessageFlag.read => false,
-        MessageFlag.starred => false,
-        MessageFlag.collapsed => false,
-        MessageFlag.hasAlertWord => false,
-        MessageFlag.historical => false,
-        MessageFlag.unknown => false,
-      });
+        case MessageFlag.mentioned:
+        case MessageFlag.wildcardMentioned:
+          return true;
+
+        case MessageFlag.read:
+        case MessageFlag.starred:
+        case MessageFlag.collapsed:
+        case MessageFlag.hasAlertWord:
+        case MessageFlag.historical:
+        case MessageFlag.unknown:
+          return false;
+      }
+    });
   }
 
   @override
--- a/lib/model/narrow.dart
+++ b/lib/model/narrow.dart
@@ -286,4 +286,34 @@ class DmNarrow extends Narrow implements SendableNarrow {
   int get hashCode => Object.hash('DmNarrow', _key);
 }
 
-// TODO other narrow types: starred, mentioned; searches; arbitrary
+class MentionsNarrow extends Narrow {
+  const MentionsNarrow();
+
+  @override
+  bool containsMessage(Message message) {
+    return message.flags.any((flag) =>
+      switch (flag) {
+        MessageFlag.mentioned => true,
+        MessageFlag.wildcardMentioned => true,
+        MessageFlag.read => false,
+        MessageFlag.starred => false,
+        MessageFlag.collapsed => false,
+        MessageFlag.hasAlertWord => false,
+        MessageFlag.historical => false,
+        MessageFlag.unknown => false,
+      });
+  }
+
+  @override
+  ApiNarrow apiEncode() => [ApiNarrowIsMentioned()];
+
+  @override
+  bool operator ==(Object other) {
+    if (other is! MentionsNarrow) return false;
+    // Conceptually there's only one value of this type.
+    return true;
+  }
+
+  @override
+  int get hashCode => 'MentionsNarrow'.hashCode;
+}
--- a/lib/widgets/actions.dart
+++ b/lib/widgets/actions.dart
@@ -134,5 +134,12 @@ Future<void> _legacyMarkNarrowAsRead(BuildContext context, Narrow narrow) async 
         messages: unreadDms,
         op: UpdateMessageFlagsOp.add,
         flag: MessageFlag.read);
+    case MentionsNarrow():
+      final unreadMentions = store.unreads.mentions.toList();
+      if (unreadMentions.isEmpty) return;
+      await updateMessageFlags(connection,
+        messages: unreadMentions,
+        op: UpdateMessageFlagsOp.add,
+        flag: MessageFlag.read);
   }
 }
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1025,13 +1025,14 @@ class ComposeBox extends StatelessWidget {
 
   static bool hasComposeBox(Narrow narrow) {
     switch (narrow) {
-      case CombinedFeedNarrow():
-        return false;
-
       case ChannelNarrow():
       case TopicNarrow():
       case DmNarrow():
         return true;
+
+      case CombinedFeedNarrow():
+      case MentionsNarrow():
+        return false;
     }
   }
 
@@ -1046,6 +1047,7 @@ class ComposeBox extends StatelessWidget {
       case DmNarrow():
         return _FixedDestinationComposeBox(key: controllerKey, narrow: narrow);
       case CombinedFeedNarrow():
+      case MentionsNarrow():
         return const SizedBox.shrink();
     }
   }
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1023,6 +1023,18 @@ class ComposeBox extends StatelessWidget {
   final GlobalKey<ComposeBoxController>? controllerKey;
   final Narrow narrow;
 
+  static bool hasComposeBox(Narrow narrow) {
+    switch (narrow) {
+      case CombinedFeedNarrow():
+        return false;
+
+      case ChannelNarrow():
+      case TopicNarrow():
+      case DmNarrow():
+        return true;
+    }
+  }
+
   @override
   Widget build(BuildContext context) {
     final narrow = this.narrow;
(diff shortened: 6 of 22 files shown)


lib/model/autocomplete.dart
```diff
@@ -187,7 +187,7 @@ class MentionAutocompleteView extends ChangeNotifier {
     required PerAccountStore store,
     required Narrow narrow,
   }) {
-    assert(narrow is! CombinedFeedNarrow);
+    assert(narrow is SendableNarrow);
```
gnprice, line 190: This is too strong — this class should work on a stream narrow, too. The point is it should work on any narrow where we offer a compose box.

(Do we have an existing test that fails with this change? If not, a good small side fix would be to add one.)

For purposes of this PR, we can just add one more case to this assert, and probably a one-line comment explaining why it's here. There might be a refactor to do if we start prolifierating to three or four such cases.

PIG208: We do have test cases that fail this, which pass `StreamNarrow`s.

PIG208:
Reading this [chat thread](https://chat.zulip.org/#narrow/stream/378-api-design/topic/highlight.20topic-wildcard.20mention/near/1621613) gives me some background on why we need the `topic_wildcard_mentioned` flag (which is to provide necessary information to determine if the @topic mention needs to be highlighted for the user).

Looks like we don't support rendering topic mentions yet, but I think we can skip that for now as @-mentions narrow doesn't rely on this feature.

PIG208:
Regarding whether we should support both `stream_wildcard_mention` and the potential future `channel_wildcard_mention` at the same time, I think the answer is no if the server is going with this [migration strategy](https://chat.zulip.org/#narrow/stream/378-api-design/topic/renaming.20to.20.22channel.22.20across.20API/near/1796687). Likely at that point we will add a client capability, but we shouldn't worry about it now.

gnprice:
> Looks like we don't support rendering topic mentions yet, but I think we can skip that for now as @-mentions narrow doesn't rely on this feature.

Yeah, it's fine to leave that flag out if it isn't needed for this feature.

If it's easy to include it, though, then it's probably less total work if we include it now than if we add the other flag now and that flag separately later.

gnprice:
Yeah agreed, let's not try to include support for a future `channel_wildcard_mention` flag at this point.

PIG208:
When working on this, I found exhaustiveness checks useful to covering places that need to be updated when adding a narrow. For addings tests, having each file corresponds to a test file helped, but I also needed to do a sweep by searching for `CombinedFeedNarrow`.

Bookkeeping comments like `"TODO other narrow types: starred, mentioned; searches; arbitrary"` will be a lot more helpful if we use them more consistently at places. However, it is understandable that the cases for different narrows can be pretty different. We can't know in advance all the time. We likely won't be adding a lot of narrows, so it should be fine without other actions taken.

PIG208:
I read the [comment](https://github.com/zulip/zulip-flutter/blob/af2e17862560a886a500aa00f26606d3b7283611/lib/model/unreads.dart#L95-L120) on the weird corner-case when messages might be absent in `mentions`, and it appears to me that it wouldn't interfere with this feature change.

gnprice:
> For addings tests, having each file corresponds to a test file helped, but I also needed to do a sweep by searching for `CombinedFeedNarrow`.
> 
> Bookkeeping comments like `"TODO other narrow types: starred, mentioned; searches; arbitrary"` will be a lot more helpful if we use them more consistently at places. However, it is understandable that the cases for different narrows can be pretty different. We can't know in advance all the time.

Yeah, I think it's OK to have to search for existing similar cases (like CombinedFeedNarrow) when adding a new type like this.

That TODO line wasn't meant for the purpose of finding all the different places that different types of narrows will need to be handled. It'd be pretty tough to have such comments comprehensively enough to rely much on them — as you saw, there are a lot of places that need updating. If you run the command `git log -p -S 'TODO other' lib/model/narrow.dart` (and use [my "secret" for reading `git log -p` output](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/git.md#git-log-secret)), you can see the history of that line; it was added as part of a very early sketch, and at this point should probably just be deleted in favor of the issues we have for adding more narrow types.

It would be good to make it so that those exhaustiveness checks were a more comprehensive guide. I think that mostly means factoring out all the conditions that look like this:
`narrow is CombinedFeedNarrow || narrow is MentionsNarrow`
into a method somewhere that says whether we offer a compose box for a narrow, and that is exhaustive by construction.

Probably the most natural home for that is as a static on ComposeBox. That'll mean the one spot in `lib/model/` (on autocomplete) can't use it; but that one spot isn't really used right now anyway, and I think once it is it'll naturally involve an exhaustive `switch` too.

PIG208:
Rebased this atop #808, and extracted the `is CombinedNarrow` checks to `ComposeBox.hasComposeBox`.

lib/widgets/compose_box.dart
```diff
@@ -958,6 +958,18 @@ class ComposeBox extends StatelessWidget {
   final GlobalKey<ComposeBoxController>? controllerKey;
   final Narrow narrow;
 
+  static bool hasComposeBox(Narrow narrow) {
+    switch (narrow) {
+      case CombinedFeedNarrow():
+        return true;
```
gnprice, line 965: This is inverted, right?

PIG208: Right. Pushed a fix.

PIG208:
Rebased to resolve the merge conflicts from the [`markNarrowAsUnread`](https://github.com/zulip/zulip-flutter/pull/827) refactor.

chrisbobbe (commented):
Thanks! Small comments below.

Then one question I had was if we want to update `parseInternalLink` so that it can return a `MentionsNarrow` (I guess just when the link is `/#narrow/is/mentioned`). I don't think it's common to want to post a link to the mentions narrow, but it crossed my mind as something that would now be possible to support.

lib/widgets/message_list.dart
```diff
@@ -541,7 +541,7 @@ class RecipientHeader extends StatelessWidget {
     final message = this.message;
     return switch (message) {
       StreamMessage() => StreamMessageRecipientHeader(message: message,
-        showStream: narrow is CombinedFeedNarrow),
+        showStream: !ComposeBox.hasComposeBox(narrow)),
       DmMessage() => DmRecipientHeader(message: message),
     };
```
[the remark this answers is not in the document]
PIG208: Yeah. I prefer adding a helper with switch cases in this case.

PIG208:
Thanks for the reminder! Added support for parsing the internal links and updated the PR.

test/model/internal_link_test.dart
```diff
@@ -221,6 +221,16 @@ void main() {
       testExpectedNarrows(testCases, streams: streams);
     });
 
+    group('"/#narrow/is/mentioned returns expected MentionsNarrow', () {
+      final testCases = [
+        ('/#narrow/is/mentioned',        const MentionsNarrow()),
+        ('/#narrow/is/mentioned/near/1', const MentionsNarrow()),
+        ('/#narrow/is/mentioned/with/2', const MentionsNarrow()),
+        ('/#narrow/-is/mentioned',       null),
+      ];
```
chrisbobbe, line 230: Let's add cases like:

```dart
('/#narrow/stream/7-test-here/is/mentioned', null),
('/#narrow/dm/17327-Chris-Bobbe-(Test-Account)/is/mentioned', null),
```

Those are valid narrows to want to see in a client, but we don't support them yet, and it seems good that we're taking the user out to the browser for those, rather than showing the whole `/is/mentioned` narrow in the app.

chrisbobbe (commented):
Thanks! One small comment below, and I'll mark this for Greg's review.

test/model/internal_link_test.dart
PIG208: Also added one with the topic segment. Fixed a bug with these new cases!

lib/widgets/compose_box.dart
```diff
@@ -1023,6 +1023,19 @@ class ComposeBox extends StatelessWidget {
   final GlobalKey<ComposeBoxController>? controllerKey;
   final Narrow narrow;
 
+  static bool hasComposeBox(Narrow narrow) {
+    switch (narrow) {
+      case CombinedFeedNarrow():
+      case MentionsNarrow():
+        return false;
+
+      case StreamNarrow():
```
gnprice, line 1032: nit: order this switch the same way as the one in `build` just below — this is effectively a summary of that one and needs to be kept in sync, so keeping the same ordering helps with verifying that at a glance

lib/widgets/message_list.dart
```diff
@@ -190,6 +189,9 @@ class MessageListAppBarTitle extends StatelessWidget {
           final names = otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)');
           return Text("DMs with ${names.join(", ")}"); // TODO show avatars
         }
+
+      case MentionsNarrow():
+        return Text(zulipLocalizations.mentionsPageTitle);
```
gnprice, line 194: nit: put this case next to CombinedFeedNarrow, as that's the most similar

test/model/message_list_test.dart
```diff
@@ -734,6 +734,52 @@ void main() {
...
+      await store.addSubscription(eg.subscription(stream2, isMuted: true));
+      await store.addUserTopic(stream2, 'C', UserTopicVisibilityPolicy.unmuted);
+
+      List<Message> getMessages(int startingId) => [
+        eg.streamMessage(id: startingId,
+          stream: stream1, topic: "A", flags: [MessageFlag.wildcardMentioned]),
+        eg.streamMessage(id: startingId + 1,
+          stream: stream1, topic: "B", flags: [MessageFlag.wildcardMentioned]),
+        eg.streamMessage(id: startingId + 2,
+          stream: stream2, topic: "C", flags: [MessageFlag.wildcardMentioned]),
```
gnprice, line 754: When writing a test, a question to ask is: what story does this test need to tell? What is it checking, and what scenarios should it be exercising?

Here, the story that's needed is: messages get included in the view, even messages you might think wouldn't be. Specifically that means: including even a message in a muted topic and muted channel, including even DMs, and including even messages that have flag `wildcardMentioned` but not `mentioned`.

(The other half of the story would be: messages get excluded from the view, even messages you might think would be included. But for this narrow there are no messages that get excluded, so there's nothing to say in that direction.)

So then the test data should be written to focus on that story. In particular, the messages here that are to an unmuted channel or topic aren't needed — if messages are included even when muted, surely they are when unmuted too.

(It looks like this set of data is based on the one for CombinedFeedNarrow. But there the answer is different for muted vs. unmuted, so it's important to cover both sides.)

test/model/narrow_test.dart
```diff
@@ -149,4 +149,17 @@ void main() {
...
+  group('MentionsNarrow', () {
+    test('containsMessage', () {
+      const narrow = MentionsNarrow();
+
+      check(narrow.containsMessage(eg.streamMessage(
+        flags: []))).isFalse();
+      check(narrow.containsMessage(eg.streamMessage(
+        flags:[MessageFlag.mentioned]))).isTrue();
+      check(narrow.containsMessage(eg.streamMessage(
+        flags: [MessageFlag.wildcardMentioned]))).isTrue();
```
gnprice, line 162: nit: this is a bit hard to read because the `isFalse` and `isTrue`, which are the most critical bits of information, get visually buried.

This reformatting helps make them stick out:

```suggestion
      check(narrow.containsMessage(
        eg.streamMessage(flags: []))).isFalse();
      check(narrow.containsMessage(
        eg.streamMessage(flags: [MessageFlag.mentioned]))).isTrue();
      check(narrow.containsMessage(
        eg.streamMessage(flags: [MessageFlag.wildcardMentioned]))).isTrue();
```

gnprice: Stepping back, though, the main purpose of a test is to keep the code from getting accidentally broken by future changes. Here, I think the biggest risk for accidental breakage is when we add another mentions-related flag, that we might forget to handle it in MentionsNarrow.containsMessage. This test as written won't help in that case — it won't get broken by that.

Probably the most effective thing to do is to make the method's implementation use an exhaustive switch rather than a pair of equality checks. Then it's fine to still have this test, but there's no longer so much relying on it.

test/widgets/actions_test.dart
```diff
@@ -267,6 +268,27 @@ void main() {
...
+      await tester.pump(Duration.zero);
+      check(connection.lastRequest).isA<http.Request>()
+        ..method.equals('POST')
+        ..url.path.equals('/api/v1/messages/flags')
+        ..bodyFields.deepEquals({
+            'messages': jsonEncode([message.id]),
+            'op': 'add',
+            'flag': 'read',
+          });
+        await tester.pumpAndSettle();
```
gnprice, line 289: nit: overindented; but also do we need this line? the similar test just above doesn't seem to have it

test/model/internal_link_test.dart
```diff
@@ -221,6 +221,19 @@ void main() {
       testExpectedNarrows(testCases, streams: streams);
     });
 
+    group('"/#narrow/is/mentioned returns expected MentionsNarrow', () {
+      final testCases = [
+        ('/#narrow/is/mentioned',                                     const MentionsNarrow()),
+        ('/#narrow/is/mentioned/near/1',                              const MentionsNarrow()),
+        ('/#narrow/is/mentioned/with/2',                              const MentionsNarrow()),
+        ('/#narrow/stream/7-test-here/is/mentioned',                  null),
+        ('/#narrow/stream/check/topic/test/is/mentioned',             null),
```
gnprice, line 230: One more test case to add:
```suggestion
        ('/#narrow/stream/7-test-here/is/mentioned',                  null),
        ('/#narrow/stream/check/topic/test/is/mentioned',             null),
        ('/#narrow/topic/test/is/mentioned',                          null),
```

This is again on the principle of exercising the cases that are closest to the line between one answer and another, most at risk of going the wrong way. If for example the `topicElement != null` check were omitted in the lines this commit adds to the implementation, then I don't think any other test cases would fail, but this one would.

lib/model/internal_link.dart
```diff
@@ -188,7 +195,10 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
     }
   }
 
-  if (dmElement != null) {
+  if (isMentionedElement != null) {
+    if (streamElement != null || topicElement != null ||  dmElement != null) return null;
```
gnprice, line 199: nit:
```suggestion
    if (streamElement != null || topicElement != null || dmElement != null) return null;
```

gnprice (commented):
Thanks @PIG208 for building this, and thanks @chrisbobbe for the previous review! Generally this looks great. Comments below.

Also, would you add screenshots of the UI changes? The button on the home page, and the new message-list page.

lib/widgets/actions.dart
```diff
@@ -134,5 +134,12 @@ Future<void> _legacyMarkNarrowAsRead(BuildContext context, Narrow narrow) async
         messages: unreadDms,
         op: UpdateMessageFlagsOp.add,
         flag: MessageFlag.read);
+    case MentionsNarrow():
+      final unreadMentions = store.unreads.mentions.toList();
```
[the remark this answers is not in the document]
PIG208: `updateMessageFlags` (which calls `connection.post`) expects a List<int> of message ids.

PIG208:
Will get back to this next week and complete the update.

PIG208:
| Home menu | Mentions page |
| - | - |
| image | image |

assets/l10n/app_en.arb
```diff
@@ -344,6 +344,10 @@
   "@loginErrorMissingUsername": {
     "description": "Error message when an empty username was provided."
   },
+  "mentionsPageTitle": "Mentions",
```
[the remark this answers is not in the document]
PIG208: Ordering them to be consistent with `MessageListAppBarTitle`.

PIG208:
Updated the PR to

- resolve the review comments
- take care of some places where "channel" can be used, depending on the context
- make the last commit to fix the issue instead

lib/model/narrow.dart
```diff
@@ -286,4 +286,29 @@ class DmNarrow extends Narrow implements SendableNarrow {
...
+class MentionsNarrow extends Narrow {
+  const MentionsNarrow();
+
+  @override
+  bool containsMessage(Message message) {
+    return message.flags.any((flag) =>
+      switch (flag) {
+        MessageFlag.mentioned => true,
+        MessageFlag.wildcardMentioned => true,
+        _ => false
```
gnprice, line 298: So in https://github.com/zulip/zulip-flutter/pull/807#discussion_r1693650623 the point was to make sure that when we add a new mentions-related flag, we don't forget to handle it here. This version doesn't really help with that. :slightly_smiling_face:

Instead, list all the flags. That incurs a small cost when we add a non-mentions-related flag, but it's worth it for preventing the class of bug where we forget this for a mentions-related flag.

lib/widgets/actions.dart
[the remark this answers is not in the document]
gnprice: Ah I see, and `Unreads.mentions` is a Set. Sounds good, then.

gnprice (commented):
Thanks for the revision! Just one comment below — otherwise all looks good.

lib/model/narrow.dart
PIG208: Good thought, thanks!

PIG208:
Pushed to address the switch case change. Should be ready once it passes the tests.

gnprice:
Thanks! Looks good — merging, with a couple of tweaks. See the added commit; and I fixed a couple of summary prefixes from "internal" and "internal_links" to "internal_link". ("Internal" sounds like some much more general meaning — like anything internal to the app.)

PR #808, opened 2024-07-12
dart [nfc]: Replace else-if's with switch.

PIG208:
The first commit preps for later changes to add more narrows (#807). Using `switch` gives us the benefit of exhaustive match checks.

narrow is refactored to become private because otherwise `switch` is not able to promote the type of it.

https://dart.dev/tools/non-promotion-reasons#private

The remaining commits are follow-ups to sweep the remaining TODO(dart-3) items throughout the code base.

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -99,7 +99,7 @@ class ZulipContent extends ContentNode {
 ///   <https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flow_layout/Block_and_inline_layout_in_normal_flow>
 ///
 /// Almost all nodes are either a [BlockContentNode] or an [InlineContentNode].
-abstract class BlockContentNode extends ContentNode {
+sealed class BlockContentNode extends ContentNode {
   const BlockContentNode({super.debugHtmlNode});
 }
 
@@ -134,7 +134,7 @@ class _BlockContentListNode extends DiagnosticableTree {
 /// but provides an inline layout context for its children.
 ///
 /// See also [InlineContainerNode].
-class BlockInlineContainerNode extends BlockContentNode {
+sealed class BlockInlineContainerNode extends BlockContentNode {
   const BlockInlineContainerNode({
     super.debugHtmlNode,
     required this.links,
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -487,7 +487,7 @@ class EmbedVideoNode extends BlockContentNode {
 ///   https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flow_layout/Block_and_inline_layout_in_normal_flow#elements_participating_in_an_inline_formatting_context
 ///
 /// Almost all nodes are either an [InlineContentNode] or a [BlockContentNode].
-abstract class InlineContentNode extends ContentNode {
+sealed class InlineContentNode extends ContentNode {
   const InlineContentNode({super.debugHtmlNode});
 }
 
@@ -548,7 +548,7 @@ class LineBreakInlineNode extends InlineContentNode {
 /// itself does.
 ///
 /// See also [BlockInlineContainerNode].
-abstract class InlineContainerNode extends InlineContentNode {
+sealed class InlineContainerNode extends InlineContentNode {
   const InlineContainerNode({super.debugHtmlNode, required this.nodes});
 
   final List<InlineContentNode> nodes;
@@ -614,7 +614,7 @@ class UserMentionNode extends InlineContainerNode {
   //   final bool isSilent;
 }
 
-abstract class EmojiNode extends InlineContentNode {
+sealed class EmojiNode extends InlineContentNode {
   const EmojiNode({super.debugHtmlNode});
 }
 
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -297,7 +297,7 @@ class CodeBlockNode extends BlockContentNode {
   }
 }
 
-class CodeBlockSpanNode extends InlineContentNode {
+class CodeBlockSpanNode extends ContentNode {
   const CodeBlockSpanNode({super.debugHtmlNode, required this.text, required this.type});
 
   final String text;
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -396,115 +396,130 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
   }
 
   Future<void> handleEvent(Event event) async {
-    if (event is HeartbeatEvent) {
-      assert(debugLog("server event: heartbeat"));
-    } else if (event is RealmEmojiUpdateEvent) {
-      assert(debugLog("server event: realm_emoji/update"));
-      realmEmoji = event.realmEmoji;
-      notifyListeners();
-    } else if (event is AlertWordsEvent) {
-      assert(debugLog("server event: alert_words"));
-      // We don't yet store this data, so there's nothing to update.
-    } else if (event is UserSettingsUpdateEvent) {
-      assert(debugLog("server event: user_settings/update ${event.property?.name ?? '[unrecognized]'}"));
-      if (event.property == null) {
-        // unrecognized setting; do nothing
-        return;
-      }
-      switch (event.property!) {
-        case UserSettingName.twentyFourHourTime:
-          userSettings?.twentyFourHourTime        = event.value as bool;
-        case UserSettingName.displayEmojiReactionUsers:
-          userSettings?.displayEmojiReactionUsers = event.value as bool;
-        case UserSettingName.emojiset:
-          userSettings?.emojiset                  = event.value as Emojiset;
-      }
-      notifyListeners();
-    } else if (event is CustomProfileFieldsEvent) {
-      assert(debugLog("server event: custom_profile_fields"));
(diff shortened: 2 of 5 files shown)


PIG208:
Hmm, not sure why dart doesn't promote the type when pattern matching a record. This seems like a bug.

https://github.com/zulip/zulip-flutter/blob/51e01f416a06be746f03292f41a34bb109a8b914/lib/model/message_list.dart#L281-L288

lib/widgets/compose_box.dart
```diff
@@ -953,24 +953,18 @@ class _FixedDestinationComposeBoxState extends State<_FixedDestinationComposeBox
...
-      return _FixedDestinationComposeBox(key: controllerKey, narrow: narrow);
-    } else if (narrow is CombinedFeedNarrow) {
-      return const SizedBox.shrink();
-    } else {
-      throw Exception("impossible narrow"); // TODO(dart-3): show this statically
-    }
+    return switch (_narrow) {
+      StreamNarrow() => _StreamComposeBox(key: controllerKey, narrow: _narrow),
+      TopicNarrow() => _FixedDestinationComposeBox(key: controllerKey, narrow: _narrow),
+      DmNarrow() => _FixedDestinationComposeBox(key: controllerKey, narrow: _narrow),
```
gnprice, line 966: I think this one will be a bit easier to read as a switch statement (rather than expression).

That way the returned expressions remain aligned with each other, and also less indented — these lines get long.

lib/widgets/compose_box.dart
```diff
@@ -953,24 +953,18 @@ class _FixedDestinationComposeBoxState extends State<_FixedDestinationComposeBox
...
+  const ComposeBox({super.key, this.controllerKey, required Narrow narrow}) : _narrow = narrow;
 
+  final Narrow _narrow;
   final GlobalKey<ComposeBoxController>? controllerKey;
-  final Narrow narrow;
 
   @override
   Widget build(BuildContext context) {
-    final narrow = this.narrow;
-    if (narrow is StreamNarrow) {
```
gnprice, line 964: narrow is refactored to become private because otherwise `switch` is not
    able to promote the type of it.

No need to change the class's definition; this local variable was solving the same problem already.

lib/model/content.dart
```diff
@@ -297,7 +297,7 @@ class CodeBlockNode extends BlockContentNode {
   }
 }
 
-class CodeBlockSpanNode extends InlineContentNode {
+class CodeBlockSpanNode extends ContentNode {
```
gnprice, line 300: This should go in its own prep commit. It's rearranging this class hierarchy, which is a potentially more significant kind of change from replacing else-if with switch, so it should get its own commit message to highlight that change and the explanation of why it's good and OK.

lib/model/store.dart
```diff
@@ -387,112 +387,110 @@ class PerAccountStore extends ChangeNotifier with StreamStore, MessageStore {
...
-        final profileData = (user.profileData ??= {});
-        final update = event.customProfileField!;
-        if (update.value != null) {
-          profileData[update.id] = ProfileFieldUserData(value: update.value!, renderedValue: update.renderedValue);
-        } else {
-          profileData.remove(update.id);
+    switch(event) {
+      case HeartbeatEvent():
+        assert(debugLog("server event: heartbeat"));
+      case RealmEmojiUpdateEvent():
```
gnprice, line 393: nit: let's separate these cases visually from each other with blank lines:
```suggestion
        assert(debugLog("server event: heartbeat"));

      case RealmEmojiUpdateEvent():
```

I think that'll help a bit in visually scanning to find where a given event is handled.

(With the `if` form, the `is FooEvent` stuck out more to the right, and the close-brace also made a more distinctive marker at the left, so there was less need for extra visual emphasis.)

gnprice (commented):
Thanks @PIG208! This will be a good cleanup. Here's small comments on a quick initial skim.

gnprice:
> Hmm, not sure why dart doesn't promote the type when pattern matching a record. This seems like a bug.

Yeah, not sure. Did you run into that in a new context, or only in the context illustrated by that comment?

PIG208:
A minimal, reproducible example:

```dart
class Foo {}

class FooA extends Foo {}

class FooB extends Foo {}

void funA(FooA first, FooA second) {}
void funB(FooB first, FooB second) {}

void demo(Foo first, Foo second) {
  switch ((first, second)) {
    case ((FooA(), FooA())):
      // The argument type 'Foo' can't be assigned to the parameter type 'FooA'
      funA(first, second);
    case ((FooB(), FooB())):
      // The argument type 'Foo' can't be assigned to the parameter type 'FooB'
      funB(first, second);
  }
}

void demoB(Foo first, Foo second) {
  final record = (first, second);
  switch (record) {
    case ((FooA(), FooA())):
      funA(record.$1, record.$2); // OK
    case ((FooB(), FooB())):
      funB(record.$1, record.$2); // OK
  }
}
```

I will check dart-lang/sdk's issues and see if I need to report this to them.

PIG208:
Updated the PR.

lib/model/content.dart
chrisbobbe, line 300:
```
content [nfc]: Make CodeBlockSpanNode inherit ContentNode.

We only expect CodeBlockSpanNode within a CodeBlockNode, and the
InlineContent widget should not be responsible for building it, as the
widget does for all other InlineContentNode.

```

Commit message nit: "Make CodeBlockSpanNode inherit ContentNode" seems accidentally misleading, because it already was inheriting from `ContentNode`, just more distantly. (`InlineContentNode` extends `ContentNode`.)

How about something like "Make CodeBlockSpanNode extend ContentNode directly"?

chrisbobbe (commented):
Thanks! LGTM, with one commit-message nit.

lib/model/content.dart
PIG208: Thanks! Fixed in the latest push.

lib/widgets/content.dart
```diff
@@ -930,45 +917,30 @@ class _InlineContentBuilder {
...
-      return WidgetSpan(alignment: PlaceholderAlignment.middle,
-        child: UserMention(ambientTextStyle: widget.style, node: node));
-    } else if (node is UnicodeEmojiNode) {
-      return TextSpan(text: node.emojiUnicode, recognizer: _recognizer);
-    } else if (node is ImageEmojiNode) {
-      return WidgetSpan(alignment: PlaceholderAlignment.middle,
-        child: MessageImageEmoji(node: node));
-    } else if (node is MathInlineNode) {
-      return TextSpan(
+      LineBreakInlineNode() => const TextSpan(text: ""),
```
gnprice, line 924: The way this comment is arranged doesn't make it entirely clear where "here" is ­— whether it means the case above, or the case below, or what. Relatedly, it has the look to me of a comment that's at risk of getting separated from the code it's about by some future change.

In general I prefer a comment about how we handle some condition to be inside the body of the conditional, rather than above the condition, largely for this reason.

lib/widgets/content.dart
```diff
@@ -930,45 +917,30 @@ class _InlineContentBuilder {
...
+      StrongNode() => _buildStrong(node),
+      DeletedNode() => _buildDeleted(node),
+      EmphasisNode() => _buildEmphasis(node),
+      LinkNode() => _buildLink(node),
+      InlineCodeNode() => _buildInlineCode(node),
+      UserMentionNode() => WidgetSpan(alignment: PlaceholderAlignment.middle,
+        child: UserMention(ambientTextStyle: widget.style, node: node)),
```
gnprice, line 931: Looking at these and thinking about the comment above, I think the arrangement I want is a switch statement with blank lines separating cases. Much like you did for `handleEvent` at my request above: https://github.com/zulip/zulip-flutter/pull/808#discussion_r1677246698

I'll just fix that up before merging. Then after that, I think the tiny helper methods here like `_buildEmphasis` make sense to inline, so I'll do that too.

lib/widgets/content.dart
```diff
@@ -266,44 +266,31 @@ class BlockContentList extends StatelessWidget {
   Widget build(BuildContext context) {
     return Column(crossAxisAlignment: CrossAxisAlignment.stretch, children: [
       ...nodes.map((node) {
-        if (node is LineBreakNode) {
+        return switch (node) {
           // This goes in a Column.  So to get the effect of a newline,
           // just use an empty Text.
```
gnprice, line 271: Similarly on this comment, though in this case I think the expression form of `switch` works fine.

lib/widgets/content.dart
```diff
@@ -266,44 +266,31 @@ class BlockContentList extends StatelessWidget {
...
+              "It should be wrapped in [ImageNodeList]."
+            );
+            return MessageImage(node: node);
+          }(),
+          InlineVideoNode() => MessageInlineVideo(node: node),
+          EmbedVideoNode() => MessageEmbedVideo(node: node),
+          UnimplementedBlockContentNode() => Text.rich(_errorUnimplemented(node, context: context)),
+        };
+
       }),
```
gnprice, line 294: nit: stray blank line (blank line at end of block)

gnprice (commented):
Thanks for cleaning these up!

A few formatting comments below, particularly about placement of comments. I'll make the changes I'd like and then merge.

Issue #675, opened 2024-05-14
msglist/dm: Provide proper feedback when sending DMs to a deactivated user


sm-sayedi:
When trying to send a DM message to a deactivated user, the message goes into void and nothing happens so the user knows they're exchanging messages with a deactivated user.

Here's how it looks right now:

<img src="https://github.com/zulip/zulip-flutter/assets/59946442/eb99074c-b1c4-4c1c-883c-282adf8a3011" width="300" />
<br/>
<br/>
This is how the web handles it:
<img src="https://github.com/zulip/zulip-flutter/assets/59946442/f939dcb9-4921-4598-b3ed-746993db744f" />
<br />
<br />
The react-native mobile app just shows an infinite loading indicator on the right side of the message.

gnprice:
Thanks! Agreed, we should fix this.

Issue #674, opened 2024-05-14
Hide compose box based on stream posting policy


gnprice:
When this Zulip feature is in effect:
https://zulip.com/help/stream-sending-policy
it may mean that the self-user wouldn't be able to send to a given stream. When that's the case, we should replace the compose box with a banner explaining this is the case.

In the API, this is `ZulipStream.streamPostPolicy`.

In zulip-mobile we have a version of this, but it only detects "announcement-only" streams, i.e. where the policy is `StreamPostPolicy.administrators`. (Because long ago that was the only version that existed.) Just recently I happened to see a user get affected by this on a stream where the policy was `StreamPostPolicy.moderators` and they weren't a moderator — the symptom (on zulip-mobile) was that the app tries to send and then the message just sits there with an outbox spinner forever. So that underlines that this feature is a meaningful one which we should be sure to implement (and do so completely).

PR #816, opened 2024-07-16
compose_box: Replace compose box with a banner in DMs with deactivated users

sm-sayedi:
When DMs have one or multiple deactivated users, all parts of the compose box are replaced with a banner, saying: _You cannot send messages to deactivated users_.

<table>
    <tr>
        <td><b>DMs with active user</b></td>
        <td><b>DMs with deactivated user</b></td>
        <td><b>DMs with deactivated user - Dark</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
<td></td>
    </tr>
</table>

https://github.com/user-attachments/assets/f219ef6f-3f65-44d8-a5d8-93b7883eb348

**Note:** This work is the result of a productive pair programming session with @rajveermalviya.

Fixes: #675

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -286,6 +286,7 @@ class RealmUserUpdateEvent extends RealmUserEvent {
 
   @JsonKey(readValue: _readFromPerson) final RealmUserUpdateCustomProfileField? customProfileField;
   @JsonKey(readValue: _readFromPerson) final String? newEmail;
+  @JsonKey(readValue: _readFromPerson) final bool? isActive;
 
   static Object? _readFromPerson(Map<dynamic, dynamic> json, String key) {
     return (json['person'] as Map<String, dynamic>)[key];
@@ -319,6 +320,7 @@ class RealmUserUpdateEvent extends RealmUserEvent {
     this.deliveryEmail,
     this.customProfileField,
     this.newEmail,
+    this.isActive,
   });
 
   factory RealmUserUpdateEvent.fromJson(Map<String, dynamic> json) =>
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -16,6 +16,7 @@ import '../model/store.dart';
 import 'autocomplete.dart';
 import 'dialog.dart';
 import 'store.dart';
+import 'theme.dart';
 
 const double _inputVerticalPadding = 8;
 const double _sendButtonSize = 36;
@@ -850,11 +851,13 @@ class _ComposeBoxLayout extends StatelessWidget {
     required this.sendButton,
     required this.contentController,
     required this.contentFocusNode,
+    this.blockingErrorBanner,
   });
 
   final Widget? topicInput;
   final Widget contentInput;
   final Widget sendButton;
+  final Widget? blockingErrorBanner;
   final ComposeContentController contentController;
   final FocusNode contentFocusNode;
 
@@ -883,28 +886,30 @@ class _ComposeBoxLayout extends StatelessWidget {
         minimum: const EdgeInsets.fromLTRB(8, 0, 8, 8),
         child: Padding(
           padding: const EdgeInsets.only(top: 8.0),
-          child: Column(children: [
-            Row(crossAxisAlignment: CrossAxisAlignment.end, children: [
-              Expanded(
-                child: Theme(
-                  data: inputThemeData,
-                  child: Column(children: [
-                    if (topicInput != null) topicInput!,
-                    if (topicInput != null) const SizedBox(height: 8),
-                    contentInput,
-                  ]))),
-              const SizedBox(width: 8),
-              sendButton,
-            ]),
-            Theme(
-              data: themeData.copyWith(
-                iconTheme: themeData.iconTheme.copyWith(color: colorScheme.onSurfaceVariant)),
-              child: Row(children: [
-                _AttachFileButton(contentController: contentController, contentFocusNode: contentFocusNode),
-                _AttachMediaButton(contentController: contentController, contentFocusNode: contentFocusNode),
-                _AttachFromCameraButton(contentController: contentController, contentFocusNode: contentFocusNode),
-              ])),
-          ]))));  }
+          child: blockingErrorBanner != null
+            ? SizedBox(width: double.infinity, child: blockingErrorBanner)
+            : Column(children: [
+                Row(crossAxisAlignment: CrossAxisAlignment.end, children: [
+                  Expanded(
+                    child: Theme(
+                      data: inputThemeData,
+                      child: Column(children: [
+                        if (topicInput != null) topicInput!,
+                        if (topicInput != null) const SizedBox(height: 8),
+                        contentInput,
+                      ]))),
+                  const SizedBox(width: 8),
+                  sendButton,
+                ]),
+                Theme(
+                  data: themeData.copyWith(
+                    iconTheme: themeData.iconTheme.copyWith(color: colorScheme.onSurfaceVariant)),
+                  child: Row(children: [
+                    _AttachFileButton(contentController: contentController, contentFocusNode: contentFocusNode),
+                    _AttachMediaButton(contentController: contentController, contentFocusNode: contentFocusNode),
+                    _AttachFromCameraButton(contentController: contentController, contentFocusNode: contentFocusNode),
+                  ])),
+            ]))));  }
 }
 
 abstract class ComposeBoxController<T extends StatefulWidget> extends State<T> {
@@ -973,6 +978,27 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> implements Compose
   }
 }
 
+class _ErrorBanner extends StatelessWidget {
+  const _ErrorBanner({required this.label});
+
+  final String label;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    return Container(
+      padding: const EdgeInsets.all(8),
+      decoration: BoxDecoration(
+        color: designVariables.errorBannerBackground,
+        border: Border.all(color: designVariables.errorBannerBorder),
+        borderRadius: BorderRadius.circular(5)),
+      child: Text(label,
+        style: TextStyle(fontSize: 18, color: designVariables.errorBannerLabel),
+      ),
+    );
+  }
+}
+
 class _FixedDestinationComposeBox extends StatefulWidget {
   const _FixedDestinationComposeBox({super.key, required this.narrow});
 
@@ -998,6 +1024,19 @@ class _FixedDestinationComposeBoxState extends State<_FixedDestinationComposeBox
     super.dispose();
   }
 
+  Widget? _errorBanner(BuildContext context) {
+    if (widget.narrow case DmNarrow(:final otherRecipientIds)) {
+      final store = PerAccountStoreWidget.of(context);
+      final hasDeactivatedUser = otherRecipientIds.any((id) =>
+        !(store.users[id]?.isActive ?? true));
+      if (hasDeactivatedUser) {
+        return _ErrorBanner(label: ZulipLocalizations.of(context)
+          .errorBannerDeactivatedDmLabel);
+      }
+    }
+    return null;
+  }
+
   @override
   Widget build(BuildContext context) {
     return _ComposeBoxLayout(
@@ -1013,7 +1052,8 @@ class _FixedDestinationComposeBoxState extends State<_FixedDestinationComposeBox
         topicController: null,
         contentController: _contentController,
         getDestination: () => widget.narrow.destination,
-      ));
+      ),
+      blockingErrorBanner: _errorBanner(context));
   }
 }
 
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -141,6 +141,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       channelColorSwatches: ChannelColorSwatches.light,
       atMentionMarker: const HSLColor.fromAHSL(0.5, 0, 0, 0.2).toColor(),
       dmHeaderBg: const HSLColor.fromAHSL(1, 46, 0.35, 0.93).toColor(),
+      errorBannerBackground: const HSLColor.fromAHSL(1, 4, 0.33, 0.90).toColor(),
+      errorBannerBorder: const HSLColor.fromAHSL(0.4, 3, 0.57, 0.33).toColor(),
+      errorBannerLabel: const HSLColor.fromAHSL(1, 4, 0.58, 0.33).toColor(),
       loginOrDivider: const Color(0xffdedede),
       loginOrDividerText: const Color(0xff575757),
       sectionCollapseIcon: const Color(0x7f1e2e48),
@@ -163,6 +166,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       // TODO(#95) need proper dark-theme color (this is ad hoc)
       atMentionMarker: const HSLColor.fromAHSL(0.4, 0, 0, 1).toColor(),
       dmHeaderBg: const HSLColor.fromAHSL(1, 46, 0.15, 0.2).toColor(),
+      errorBannerBackground: const HSLColor.fromAHSL(1, 0, 0.61, 0.19).toColor(),
+      errorBannerBorder: const HSLColor.fromAHSL(0.4, 3, 0.73, 0.74).toColor(),
+      errorBannerLabel: const HSLColor.fromAHSL(1, 2, 0.73, 0.80).toColor(),
       loginOrDivider: const Color(0xff424242),
       loginOrDividerText: const Color(0xffa8a8a8),
       // TODO(#95) need proper dark-theme color (this is ad hoc)
@@ -185,6 +191,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     required this.channelColorSwatches,
     required this.atMentionMarker,
     required this.dmHeaderBg,
+    required this.errorBannerBackground,
+    required this.errorBannerBorder,
+    required this.errorBannerLabel,
     required this.loginOrDivider,
     required this.loginOrDividerText,
     required this.sectionCollapseIcon,
@@ -218,6 +227,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
   // Not named variables in Figma; taken from older Figma drafts, or elsewhere.
   final Color atMentionMarker;
   final Color dmHeaderBg;
+  final Color errorBannerBackground;
+  final Color errorBannerBorder;
+  final Color errorBannerLabel;
   final Color loginOrDivider; // TODO(#95) need proper dark-theme color (this is ad hoc)
   final Color loginOrDividerText; // TODO(#95) need proper dark-theme color (this is ad hoc)
   final Color sectionCollapseIcon;
@@ -238,6 +250,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     ChannelColorSwatches? channelColorSwatches,
     Color? atMentionMarker,
     Color? dmHeaderBg,
+    Color? errorBannerBackground,
+    Color? errorBannerBorder,
+    Color? errorBannerLabel,
     Color? loginOrDivider,
     Color? loginOrDividerText,
     Color? sectionCollapseIcon,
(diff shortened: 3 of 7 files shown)


lib/widgets/compose_box.dart
```diff
@@ -271,16 +271,21 @@ class _ContentInput extends StatelessWidget {
...
   final FocusNode focusNode;
   final String hintText;
+  final bool enabled;
 
   @override
   Widget build(BuildContext context) {
     ColorScheme colorScheme = Theme.of(context).colorScheme;
+    if (!enabled) {
+      controller.clear();
+    }
```
Khader-1, line 288: optional nit 
```suggestion
    if (!enabled) controller.clear();
```

Khader-1 (approved):
Thanks @sm-sayedi! All looks good, I have one suggestion below PTAL. Over to mentor review with @hackerkid now.

alya:
The text in the screenshot is slightly different from the web app:

Screenshot 2024-07-17 at 15 00 11@2x

I don't have strong feelings on what's better, but we might as well keep it consistent. I would default to just using the web app text, unless others think this is an improvement. If it turns out that folks like this better, we'd want to tweak it in the web app.

lib/widgets/compose_box.dart
```diff
@@ -667,20 +694,24 @@ class _SendButton extends StatefulWidget {
...
 
   @override
   State<_SendButton> createState() => _SendButtonState();
 }
 
 class _SendButtonState extends State<_SendButton> {
   void _hasErrorsChanged() {
-    setState(() {
-      // Update disabled/non-disabled state
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
gnprice, line 711: I spotted this and wondered about it, and then happened to see something similar in another PR today and looked into it. See thread here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/addPostFrameCallback/near/1893536

sm-sayedi:
> The text in the screenshot is slightly different from the web app:

The image provided shows the text message for a 1:1 conversation; for a group conversation, it is the same as the web app. We thought it would be slightly better to distinguish between 1 vs. many deactivated users.

alya:
I see. I would just use the plural everywhere. It's stated as a rule, and doesn't imply that there are specifically many users in the current conversation.

sm-sayedi:
Revision pushed with PR description images updated.

alya:
Looks good to me, thanks!

lib/widgets/compose_box.dart
```diff
@@ -378,32 +380,37 @@ class _FixedDestinationContentInput extends StatelessWidget {
...
         final fullName = store.users[otherUserId]?.fullName;
         if (fullName == null) return zulipLocalizations.composeBoxGenericContentHint;
         return zulipLocalizations.composeBoxDmContentHint(fullName);
 
-      case DmNarrow(): // A group DM thread.
+      case (DmNarrow(), true): // A group DM thread.
         return zulipLocalizations.composeBoxGroupDmContentHint;
+
+      case (DmNarrow(), false):
+        return zulipLocalizations.composeBoxDeactivatedDmContentHint;
```
gnprice, line 413: Here we're relying on an assumption that if `enabled` is false, it's because there's a deactivated user.

That's how the code currently works in this version, as of the tip of this PR. But it's not expressed in the name `enabled`, nor in any dartdoc on that field. So now that the field is added, it'd be very natural for someone in the future to add some other condition that causes `enabled` to be false; but that would produce a misleading message here.

One adequate fix would be to give the field a more specific name that matches the semantics this widget is attaching to it. For example, `disabledBecauseDmRecipientDeactivated`. Then this code can look for that flag, and return this specific error message when it's set, with a clear conscience.

A more complex fix would be to make an enum in the same spirit as `TopicValidationError` and `ContentValidationError`, with a name like `DestinationError`. It'd have just one value at first, named like `DestinationError.dmRecipientDeactivated`. Then the flag here would be declared like
```dart
  final DestinationError? destinationError;
```
and null would mean there was no error.

The advantage of that more complex fix is that it'd be straightforward to add another value to the enum in the future, for a different error condition (like running afoul of the stream post policy, #674).

test/widgets/message_list_test.dart
```diff
@@ -981,4 +983,222 @@ void main() {
       });
     });
   });
+
+  group('message list page for DMs with deactivated users', () {
```
gnprice, line 987: Does the new logic interact with the message list? I don't think it does — it seems like it's all contained in the compose box.

In that case these tests should go in `test/widgets/compose_box_test.dart` instead, and should be able to use `prepareComposeBox` there (or something like it) without setting up a MessageListPage.

In general when adding code in a file `lib/foo/bar.dart`, we try to put the tests for that code in the corresponding file `test/foo/bar_test.dart`.

gnprice (commented):
Thanks @sm-sayedi and @rajveermalviya! Comments below.

Taking a step back, though, and this may make some of my detailed code comments moot: the current UI in this PR is that if you're in this error condition where you can't actually send a message, we show some error text saying so… but we do so in the form of a "hint" placeholder in the content input.

That doesn't seem like a conventional use of that piece of UI. For example the other analogous hints we put there are like "Message #general > ideas", not errors or warnings.

It also doesn't match what we do on web, where there's a red banner like Alya showed above.

How about instead:
* When there's this error condition, we don't show the content input at all, or the send button, or the attach-files/etc. buttons.
* Instead, there's a red banner. It'll look similar to the one on web, though unlike the one on web it will take the place of those other pieces of the compose box.
* There'll still be the gray background for the compose box as a whole. The banner will sit inside that, similar to how the content input and the buttons etc. normally sit inside it.

A bonus is that I think that will simplify a lot of the code in this PR: fewer layers to thread the enabled/disabled information down through, and no need for the `_contentController.clear` call that made the data flow trickier as discussed at https://github.com/zulip/zulip-flutter/pull/816#discussion_r1682109265 .

sm-sayedi:
Thanks @gnprice for the review. The suggested design looks better and more intuitive. Implemented that. PTAL.

lib/api/model/events.dart
```diff
@@ -285,6 +285,7 @@ class RealmUserUpdateEvent extends RealmUserEvent {
 
   @JsonKey(readValue: _readFromPerson) final RealmUserUpdateCustomProfileField? customProfileField;
   @JsonKey(readValue: _readFromPerson) final String? newEmail;
+  @JsonKey(readValue: _readFromPerson) final bool? isActive;
```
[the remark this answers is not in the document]
sm-sayedi: > It looks like this PR is missing code to actually apply the update in handleEvent, though.

Thanks for spotting. I was wondering why the tests for the previous revision were failing, and now I know why. When rebasing ` store [nfc]: Replace else-if's with switch for handleEvent`, I forgot to include this part when resolving the conflicts. 😀

lib/widgets/compose_box.dart
```diff
@@ -982,6 +984,29 @@ class _FixedDestinationComposeBox extends StatefulWidget {
...
+
+  final String label;
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      padding: const EdgeInsets.all(8),
+      decoration: BoxDecoration(
+        color: const Color.fromRGBO(238, 222, 221, 1),
+        border: Border.all(color: const Color.fromRGBO(132, 41, 36, 0.4)),
```
gnprice, line 998: Rather than hard-code colors here, let's put them on DesignVariables so that they each have one value for light mode and another for dark mode. For background context, see #95.

In the class, put them in the group of fields at the end, for variables that didn't come from Vlad.

To test dark mode locally, you can change the `brightness` local in `zulipThemeData` and then hot reload.

lib/widgets/compose_box.dart
```diff
@@ -982,6 +984,29 @@ class _FixedDestinationComposeBox extends StatefulWidget {
   State<_FixedDestinationComposeBox> createState() => _FixedDestinationComposeBoxState();
 }
 
+class _ErrorBanner extends StatelessWidget {
```
gnprice, line 987: This is separating a StatefulWidget from its corresponding State; move it above to not separate them.

lib/widgets/compose_box.dart
```diff
@@ -998,6 +1023,19 @@ class _FixedDestinationComposeBoxState extends State<_FixedDestinationComposeBox
     super.dispose();
   }
 
+  Widget? _placeholder(BuildContext context) {
+    if (widget.narrow case DmNarrow(:final otherRecipientIds)) {
+      final store = PerAccountStoreWidget.of(context);
+      final showPlaceholder = otherRecipientIds.any((id) =>
+        !(store.users[id]?.isActive ?? true));
+      if (showPlaceholder) {
```
gnprice, line 1031: A more descriptive name for this local would be like `hasDeactivatedUser`.

lib/widgets/compose_box.dart
```diff
@@ -998,6 +1023,19 @@ class _FixedDestinationComposeBoxState extends State<_FixedDestinationComposeBox
     super.dispose();
   }
 
+  Widget? _placeholder(BuildContext context) {
+    if (widget.narrow case DmNarrow(:final otherRecipientIds)) {
+      final store = PerAccountStoreWidget.of(context);
+      final showPlaceholder = otherRecipientIds.any((id) =>
+        !(store.users[id]?.isActive ?? true));
```
gnprice, line 1030: This `?? true` doesn't seem right — it means we'll show this error banner if there's a user we don't have access to see. In that situation it isn't particularly likely the user is deactivated — it's much more likely that our user is a guest and they've somehow navigated to a narrow containing a user they can't see. So the error message would be misleading.

Simplest fix is to just say `?? false`. If we can't see the user, then the server probably won't ultimately let us send them a message… but then we'll get an error message at that point, so the effect is only annoying and not catastrophic.

A more complete fix would be to look first for any unknown users in the list, and in that case show a different message. Could be "You cannot send messages to unknown users." — "unknown user" is the term we use for these users in the UI and in the Help Center:
https://zulip.com/help/guest-users#configure-whether-guests-can-see-all-other-users

lib/widgets/compose_box.dart
```diff
@@ -850,11 +850,13 @@ class _ComposeBoxLayout extends StatelessWidget {
     required this.sendButton,
     required this.contentController,
     required this.contentFocusNode,
+    this.placeholder,
   });
 
   final Widget? topicInput;
   final Widget contentInput;
   final Widget sendButton;
+  final Widget? placeholder;
```
gnprice, line 859: This seems like more than a placeholder — it's not occupying a space where we lack something else to put there, but rather is displacing the things we would otherwise put here.

Could call it… `blockingErrorBanner`. That name disambiguates from an alternative UI (which is what web does for this very condition, and we might do in the future for some other conditions) where there's an error banner but the rest of the compose-box UI is still there too.

lib/widgets/message_list.dart
```diff
@@ -117,24 +117,25 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
...
-              //   if those details get complicated, refactor to avoid copying.
-              // TODO(#311) If we have a bottom nav, it will pad the bottom
-              //   inset, and this should always be true.
-              removeBottom: widget.narrow is! CombinedFeedNarrow,
-
-              child: Expanded(
-                child: MessageList(narrow: widget.narrow))),
-            ComposeBox(controllerKey: _composeBoxKey, narrow: widget.narrow),
-          ]))));
+          child: Column(crossAxisAlignment: CrossAxisAlignment.stretch,
```
gnprice, line 120: Hmm interesting — why is this change needed?

test/widgets/compose_box_test.dart
```diff
@@ -368,4 +374,112 @@ void main() {
...
+        final deactivatedUsers = [eg.user(isActive: false), eg.user(isActive: false)];
+        await prepareComposeBox(tester, narrow: groupDmNarrowWith(deactivatedUsers),
+          users: deactivatedUsers);
+        checkComposeBox(isShown: false);
+
+        await changeUserStatus(tester, user: deactivatedUsers[0], isActive: true);
+        checkComposeBox(isShown: false);
+
+        await changeUserStatus(tester, user: deactivatedUsers[1], isActive: true);
+        checkComposeBox(isShown: true);
```
gnprice, line 481: Cool, these test cases are nice and compact.

gnprice (commented):
Thanks @sm-sayedi for the revision! I agree — this design looks good. Small comments on the code.

lib/widgets/message_list.dart
sm-sayedi: Without `.stretch`, the error banner would not fill the available width. Another alternative could be to give `width: double.infinity` to the `Container` in `_ErrorBanner`, but I think that is not a good option, as this widget may be reused in other places where it doesn't necessarily need to occupy all the available width; so I thought let the parent decide its width.

sm-sayedi:
Thanks @gnprice for the review. Changes pushed.

chrisbobbe:
Looks like CI is failing; could you fix that?

sm-sayedi:
> Looks like CI is failing; could you fix that?

Done!

lib/widgets/compose_box.dart
sm-sayedi: I think it's the opposite. 🙂 If we don't know about a user, we assume it as an active one. `showPlaceholder` will be true if there is a deactivated user, so in this case, we don't show the error banner.

test/widgets/compose_box_test.dart
```diff
@@ -40,14 +42,15 @@ void main() {
...
+    await tester.pumpWidget(Builder(builder: (context) =>
       MaterialApp(
+        theme: zulipThemeData(context),
         localizationsDelegates: ZulipLocalizations.localizationsDelegates,
         supportedLocales: ZulipLocalizations.supportedLocales,
         home: GlobalStoreWidget(
           child: PerAccountStoreWidget(
             accountId: eg.selfAccount.id,
-            child: ComposeBox(controllerKey: controllerKey, narrow: narrow)))));
+            child: ComposeBox(controllerKey: controllerKey, narrow: narrow))))));
```
chrisbobbe, line 53: To resolve the conflict with the just-merged #805, you should be able to just drop these changes. `TestZulipApp` (used here in #805) takes care of `theme: zulipThemeData(context)` for you.

chrisbobbe (commented):
I haven't read the commits in detail, but wanted to make a note to help you resolve conflicts with my just-merged PR #805.

sm-sayedi:
Thanks @chrisbobbe for the tip! Conflict resolved! PTAL!

lib/widgets/message_list.dart
gnprice: Hmm, I see. This still seems like the wrong level in the tree to be handling that, though: in particular, this forces the MessageList widget itself to be wide, which isn't desirable. This also represents the message-list page telling the compose box it must be wide; I think it'd make more sense for the compose box to decide how wide it wants to be, within the width available from its parent.

So that means `_ComposeBoxLayout` would handle making the banner take all the width available to it. That's the widget that already handles that for the compose box's text inputs.

lib/widgets/compose_box.dart
gnprice: Ahhh, I miscounted the `!` operator around this. Carry on, then.

chrisbobbe:
Looks like this has a conflict with the recently merged 7bd5befed2fbd6ec607a76bf43b07714101bbfaf; could you resolve that please?

sm-sayedi:
Thanks @chrisbobbe. Conflicts resolved!

lib/widgets/theme.dart
```diff
@@ -136,6 +136,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       title: const Color(0xff1a1a1a),
       streamColorSwatches: StreamColorSwatches.light,
       star: const HSLColor.fromAHSL(0.5, 47, 1, 0.41).toColor(),
+      errorBannerLabel: const Color.fromRGBO(133, 42, 35, 1),
+      errorBannerBackground: const Color.fromRGBO(238, 222, 221, 1),
+      errorBannerBorder: const Color.fromRGBO(132, 41, 36, 0.4),
```
chrisbobbe, line 141: Since the web css uses HSL, let's use `HSLColor.fromAHSL(…).toColor()`, so it's easier to check against web. (Here and in `.dark` below.)

lib/widgets/theme.dart
```diff
@@ -181,6 +190,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
 
   // Not named variables in Figma; taken from older Figma drafts, or elsewhere.
   final Color star;
+  final Color errorBannerLabel;
+  final Color errorBannerBackground;
+  final Color errorBannerBorder;
```
chrisbobbe, line 195: I think we'd like the items in this section to be in alphabetical order. When making that change, please also update the other boilerplate where these names appear in the class.

lib/widgets/compose_box.dart
```diff
@@ -973,6 +978,29 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> implements Compose
...
+      padding: const EdgeInsets.all(8),
+      decoration: BoxDecoration(
+        color: designVariables.errorBannerBackground,
+        border: Border.all(color: designVariables.errorBannerBorder),
+        borderRadius: BorderRadius.circular(5)),
+      child: Text(label,
+        maxLines: 2,
+        overflow: TextOverflow.ellipsis,
+        style: TextStyle(fontSize: 18, color: designVariables.errorBannerLabel),
+      ),
```
chrisbobbe, line 999: Can we remove the `maxLines` and `overflow`, and just let the banner be as tall as it needs to be? It seems unhelpful to cut off some of the message without giving a way to see the whole thing. As long as we control `label`, we should be able to keep it short enough that it doesn't take up way too much vertical space.

test/widgets/compose_box_test.dart
```diff
@@ -361,4 +367,112 @@ void main() {
...
+    }
+
+    Future<void> changeUserStatus(WidgetTester tester,
+        {required User user, required bool isActive}) async {
+      await store.handleEvent(RealmUserUpdateEvent(id: 1,
+        userId: user.userId, isActive: isActive));
+      await tester.pump();
+    }
+
+    final selfUser = eg.selfUser;
```
chrisbobbe, line 405: Let's just inline this; three character's isn't much savings. 🙂

chrisbobbe (commented):
Thanks! Small comments below.

sm-sayedi:
Thanks @chrisbobbe for the review. Revision pushed. PTAL.

lib/widgets/theme.dart
```diff
@@ -147,6 +150,9 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       title: const Color(0xffffffff),
       channelColorSwatches: ChannelColorSwatches.dark,
       // TODO(#95) unchanged in dark theme?
+      errorBannerBackground: const HSLColor.fromAHSL(1, 0, 0.61, 0.19).toColor(),
+      errorBannerBorder: const HSLColor.fromAHSL(0.4, 3, 0.73, 0.74).toColor(),
+      errorBannerLabel: const HSLColor.fromAHSL(1, 2, 0.73, 0.80).toColor(),
       star: const HSLColor.fromAHSL(0.5, 47, 1, 0.41).toColor(),
```
chrisbobbe, line 156: The `TODO(#95)` has been separated from the line that it's about.

chrisbobbe (changes requested):
Thanks! LGTM except one thing below, and I'll mark this for Greg's review.

chrisbobbe:

chrisbobbe (approved):

gnprice:
Thanks again @sm-sayedi and @rajveermalviya, and @chrisbobbe for the previous review!

This looks good — merging, with one commit-message tweak:
```
$ git range-diff origin pr/816 @
1:  8fe389dfc ! 1:  8c5053791 api: Add `isActive` to `RealmUserUpdateEvent`
    @@ Metadata
      ## Commit message ##
         api: Add `isActive` to `RealmUserUpdateEvent`
     
    +    This case was new in feature level 222 (server-8).
    +
         This will become handy in the next commit(s) where we want to change the UI
         based on `User.isActive` when having a DM conversation.
     
2:  89c8c95ea = 2:  cc451157a compose_box test [nfc]: Make `prepareComposeBox` take a list of users
3:  6e18c99ac = 3:  052f2036e compose_box: Replace compose box with a banner in DMs with deactivated users
```

to give a bit more context on why that `isActive` case is being added now and how it happened that it wasn't already there.

I feel like `_ComposeBoxLayout` is doing two somewhat different jobs now, and could be improved with some refactoring. But rather than block merging this on doing that refactor, I'll try to whip that up and send it as a follow-up PR.

Issue #815, opened 2024-07-16
On failed message send, show error dialog


gnprice:
When you hit the send button in the compose box, and the send request fails for whatever reason, we currently drop that error on the floor; we don't show the user any feedback about the failure. We should fix that, by showing an error dialog box.

For examples of how to do this, see the existing call sites of `showErrorDialog`.

(Historical note: I think the only reason we haven't already done this is that this feature, being so critical to using Zulip, was added early in the prototype before we'd worked out dialog boxes. That's a natural thing to happen, but a bit ironically it means that we've had error handling for a number of less-critical actions for a while without yet having it here.)

This is effectively a simplified version of #720: for this issue, we won't yet make any changes to the compose box and send button, like to give them a "disabled" appearance or keep the text available. Those will be good improvements to make (and #720 tracks making them), but even just showing the error dialog will be an improvement over the status quo.

One effect of this is that until the permission check #791 is implemented, we'll be in much the same position for that feature as zulip-mobile now is (since zulip/zulip-mobile#5878).

Other reports of this issue, indirectly, appear in:
* #675 
* #720

Issue #791, opened 2024-07-06
Support new settings for restricting direct messages


gnprice:
This is an upcoming Zulip feature that's expected to appear in Zulip Server 9.0. We'll want to support it properly in zulip-flutter, and we might as well do so now while it's fresh in mind.

The corresponding zulip/zulip and zulip-mobile issues are:
* https://github.com/zulip/zulip/issues/24467
* https://github.com/zulip/zulip-mobile/issues/5870

The UX for this should be similar to #675, which is a different feature that also prevents sending DMs to certain users.

Issue #720, opened 2024-06-04
Await send request; handle failure


gnprice:
If you're using the app, then step into an elevator (or otherwise lose your network connection), and hit send on a message, the current experience is pretty bad: the message just vanishes, there's no error message, and you don't even get back the text you wrote so you can try again.

The UX we'd ultimately like to have here is local echo, with an "outbox" system that can retry the send when you're back on the network:
* #576 
* #133 

But those are complex to get right, and I think we won't get to them in the next few months. There's a much simpler thing we can do that would greatly mitigate this, so we should do the simpler thing first.

Specifically, when you hit the send button:
* We'll put the compose box and send button into a disabled, "working on it" state — for example the text inputs might be grayed out, and the send button showing a spinner instead of the "send" icon.
* In that state, we'll let you select the text from the input box in order to copy it, in case you decide it's taking too long and want to give up.
* We'll await the result of the send request. If it succeeds, the send is done and we can proceed just like the send button does now: clear the input and take it back to a ready state. If it fails, we'll show an error message and make the input ready again, but still holding the text the user entered.

(To keep this simple to implement, we'll only be waiting for the *send request* to complete — not for the [`message` event](https://zulip.com/api/get-events#message), which is what causes the message to appear full-fledged in the message list. This means that the message might appear in the list before or after the compose area goes back to ready. I'm hopeful that in practice this won't often be bothersome, and will be better than the status quo. If we find we need to wait for the `message` event in order for the interaction to feel right, then that gets a lot closer to a full version of #133, and we'll re-evaluate whether there's still a simple version that makes sense.)

@terpimost will be preparing a more concrete design. Until that's ready, this issue isn't ready for implementation.

Thanks also to @terpimost for the feedback pointing out the need for this, in particular the elevator story :-)

gnprice:
I've filed the first steps of this feature, which will already be useful in themselves, as a separate issue:
* #815

terpimost:
Does this reflect the logic?
<img width="752" alt="image" src="https://github.com/user-attachments/assets/0ebddd9b-0172-42db-9b42-f9a641ef0eba">
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4031-13405&t=hE9wEtLE59jG5rUn-1

gnprice:
That looks good, thanks!

[...]

PR #819, opened 2024-07-17
compose box: On failed message send, show error dialog

chrisbobbe:
Fixes: #815

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -5,6 +5,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import 'package:image_picker/image_picker.dart';
 
+import '../api/exception.dart';
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
 import '../model/compose.dart';
@@ -716,7 +717,7 @@ class _SendButtonState extends State<_SendButton> {
       || widget.contentController.hasValidationErrors.value;
   }
 
-  void _send() {
+  void _send() async {
     if (_hasValidationErrors) {
       final zulipLocalizations = ZulipLocalizations.of(context);
       List<String> validationErrorMessages = [
@@ -735,9 +736,26 @@ class _SendButtonState extends State<_SendButton> {
 
     final store = PerAccountStoreWidget.of(context);
     final content = widget.contentController.textNormalized;
-    store.sendMessage(destination: widget.getDestination(), content: content);
 
     widget.contentController.clear();
+
+    try {
+      // TODO(#720) clear content input only on success response;
+      //   while waiting, put input(s) and send button into a disabled
+      //   "working on it" state (letting input text be selected for copying).
+      await store.sendMessage(destination: widget.getDestination(), content: content);
+    } on ApiRequestException catch (e) {
+      if (!mounted) return;
+      final zulipLocalizations = ZulipLocalizations.of(context);
+      final message = switch (e) {
+        ZulipApiException() => zulipLocalizations.errorServerMessage(e.message),
+        _ => e.message,
+      };
+      showErrorDialog(context: context,
+        title: zulipLocalizations.errorMessageNotSent,
+        message: message);
+      return;
+    }
   }
 
   @override
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -15,6 +15,7 @@ import '../example_data.dart' as eg;
 import '../flutter_checks.dart';
 import '../model/binding.dart';
 import '../stdlib_checks.dart';
+import 'dialog_checks.dart';
 
 void main() {
   TestZulipBinding.ensureInitialized();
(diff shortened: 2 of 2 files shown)


lib/widgets/compose_box.dart
```diff
@@ -735,7 +736,23 @@ class _SendButtonState extends State<_SendButton> {
 
     final store = PerAccountStoreWidget.of(context);
     final content = widget.contentController.textNormalized;
-    store.sendMessage(destination: widget.getDestination(), content: content);
+
+    try {
+      // TODO(#720) put input(s) and send button into a disabled "working on it"
+      //   state (allowing input text to be selected for copying).
+      await store.sendMessage(destination: widget.getDestination(), content: content);
```
gnprice, line 743: This changes more than the error behavior — it also means we wait to clear the content input until the request completes.

Our promptly clearing the input isn't entirely a good thing; that's what #720 is about. But if we take that out without doing the other work called for in #720, I think the effect is that there's a lack of feedback so that it feels like the send button didn't do anything, or is sluggish.

Trying this, I notice that even on my fast home network connection. Then to demonstrate how it behaves when the request is especially slow, try this patch:
```diff
--- lib/api/core.dart
+++ lib/api/core.dart
@@ -188,6 +188,7 @@ class ApiConnection {
     if (params != null) {
       request.bodyFields = encodeParameters(params)!;
     }
+    await Future<void>.delayed(const Duration(seconds: 2));
     return send(routeName, fromJson, request, overrideUserAgent: overrideUserAgent);
   }
 
```

The way to fix that looks something like:
```dart
final sendFuture = store.sendMessage(…);
widget.contentController.clear();

try {
  await sendFuture;
} on …
```

Or I guess more straightforwardly, we can just call `clear` before `sendMessage`.

gnprice (commented):
Thanks for taking care of this quickly! Comments below.

test/widgets/compose_box_test.dart
```diff
@@ -155,13 +172,77 @@ void main() {
...
+      required void Function(Message) prepareResponse,
+    }) async {
+      final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+      final stream = eg.stream(streamId: 123);
+      final message = eg.streamMessage(
+        stream: stream,
+        topic: 'some topic',
+        content: 'hello world',
+      );
+      await prepareComposeBox(tester, TopicNarrow.ofMessage(message));
```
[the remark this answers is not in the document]
chrisbobbe: Ah indeed!

lib/widgets/compose_box.dart
chrisbobbe: Makes sense.

chrisbobbe:

lib/widgets/compose_box.dart
```diff
@@ -735,9 +736,26 @@ class _SendButtonState extends State<_SendButton> {
...
     final content = widget.contentController.textNormalized;
-    store.sendMessage(destination: widget.getDestination(), content: content);
 
     widget.contentController.clear();
+
+    try {
+      // TODO(#720) await send request, putting input(s) and send button into a
+      //   disabled "working on it" state (allowing input text to be selected
+      //   for copying).
+      await store.sendMessage(destination: widget.getDestination(), content: content);
```
gnprice, line 746: nit: first clause of the TODO sounds like a description of what the code is already doing

test/widgets/compose_box_test.dart
```diff
@@ -155,13 +172,71 @@ void main() {
...
+      final key = await prepareComposeBox(tester,
+        TopicNarrow.ofMessage(eg.streamMessage()));
+      checkComposeBoxTextFields(tester, controllerKey: key,
+        expectTopicTextField: false);
+    });
+  });
+
+  group('message-send request response', () {
+    Future<void> setupAndTapSend(WidgetTester tester, {
+      required void Function(int) prepareResponse,
```
gnprice, line 191: nit: can give a hint what the callback parameter means:
```suggestion
      required void Function(int messageId) prepareResponse,
```

(wasn't so necessary when the type was Message)

gnprice (commented):
Thanks for the revision! Two nits above, and otherwise LGTM.

alya:
Maybe we can drop "The server said:" part? It doesn't seem necessary to me from an end user perspective.

gnprice:
In general this error dialog is handling an error that the server gave us but we didn't anticipate. (Anticipating this particular error is #791, which is behind a couple of larger prerequisite issues.)

So that's meant to be a situation that's rare; and because we didn't anticipate the error, it's also possible that the text is somewhat obscure or out of context. That's why I think it's potentially helpful to disambiguate that this is text from the server and not some other error message from within the app.

We use the same pattern in handling some other errors too, so it'd be easy to change them all together.

alya:
Ok. Might be worth discussing in #mobile, but definitely not a blocker for merging this PR, given that context.

chrisbobbe:

gnprice:
Thanks! Looks good — merging.

Issue #799, opened 2024-07-10
Make use of thumbnails


alya:
In https://github.com/zulip/zulip/issues/16210, we're implementing thumbnailing for uploaded images.

We should use thumbnails in the mobile app. In particular, the message view should show thumbnails, while the lightbox should download the full-scale images.

Blockers:
- https://github.com/zulip/zulip/issues/16210

gnprice:
For implementing this feature, the pieces we'll need are:
* There's a [new bit of HTML for a placeholder](https://chat.zulip.org/#narrow/stream/378-api-design/topic/thumbnail.20placeholder/near/1833796) where the server is preparing a thumbnail.

  We'll need to parse that in `lib/model/content.dart`, and then build an appropriate widget to show the desired UI: a spinner similar to what Zulip web shows.

* In images in message content, the HTML will look very similar but the two image URLs that appear (one at the `href` of an enclosing `a`, one at the `src` of an `img`) will start being different from each other: one is the thumbnail and one the original image.

  We'll need to start tracking both those URLs in the content parser. Then we'll use one for the image in the message list, and the other for the lightbox.

  The "Copy link" feature in the lightbox should also get the original full-size image. I think this will happen automatically given a natural implementation of the rest of the feature, but it should be included as one of the test cases.

Issue #190, opened 2023-06-14
Systematically survey message content for unimplemented features


gnprice:
Our parsing of Zulip message content HTML is designed to be precise about what it expects, and explicit about anything it doesn't understand. This means that when we encounter some content that does something we don't have support for, our code generally knows that, rather than silently plow ahead with a wrong interpretation.

This is helpful because, among other things, it means that if we take a corpus of Zulip message content, we can run our parser on it to learn about constructs that exist in the wild that we haven't yet implemented. (This includes constructs that a current Zulip server will generate, and constructs that older servers would generate and consequently still exist in old messages.)

So we should do that, as an iterative process:
* Run on some set of messages; find things that are unimplemented; file issues; fix them, or at least those that are most common, until the number of unsupported messages is small.
* Then run on a larger set of messages, or one with more very old messages, to find more unimplemented things.
* Repeat until it's difficult to find a message with something we don't know about, and we're comfortable with whatever set of known unimplemented features remain.

Some specific likely steps in that process:
* Write a script that can fetch a bunch of messages in a loop, feed them through `parseContent`, and report any `UnimplementedNode` results (as well as any crashes, which should be fixed immediately).
* Run that script on some recent public messages (i.e., messages to public streams) on chat.zulip.org.
* Run it on all public messages on chat.zulip.org.
* Run it on all public messages in realms that [publicly list themselves as open communities](https://zulip.com/communities/). For example, have the script sign up a test user in each realm and log in as that test user.

gnprice:
We'll also want to keep these scripts in reusable form. It'll be useful to rerun them periodically, both as we make changes to the parser and to validate that there aren't new patterns we haven't implemented and are unaware of.

In particular when there's an invariant we believe applies and want to validate that empirically, we can add asserts for that invariant to the parser and then rerun the survey script, and see if it finds failures.

(We'll basically always want to run the script with assertions enabled, since the point of it is to find situations we didn't expect.)

gnprice:
(Separately, when we run into an unimplemented feature in a message we're trying to show in the message list, our current UI is pretty loudly explicit about that. That's helpful for development but we'll naturally want to handle it differently for normal use beyond the beta: #194.)

PR #820, opened 2024-07-17
content: Handle thumbnails

rajveermalviya:
| Thumbnails | Placeholders |
| - | - |
| <video src="https://github.com/user-attachments/assets/19413ccd-cc06-42c0-8894-a1058b6c8bac"> | <video src="https://github.com/user-attachments/assets/3c7d1f77-65b3-4851-a9f1-630e497dfbe3"> |

Fixes: #799

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -351,26 +351,51 @@ class ImageNodeList extends BlockContentNode {
 }
 
 class ImageNode extends BlockContentNode {
-  const ImageNode({super.debugHtmlNode, required this.srcUrl});
+  const ImageNode({
+    super.debugHtmlNode,
+    required this.srcUrl,
+    required this.thumbnailUrl,
+    required this.loading,
+  });
 
-  /// The unmodified `src` attribute for the image.
+  /// The canonical source URL of the image.
   ///
-  /// This may be a relative URL string.  It also may not work without adding
+  /// This may be a relative URL string. It also may not work without adding
   /// authentication credentials to the request.
   final String srcUrl;
 
+  /// The thumbnail URL of the image.
+  ///
+  /// This may be a relative URL string. It also may not work without adding
+  /// authentication credentials to the request.
+  ///
+  /// This will be null if the server hasn't yet generated a thumbnail,
+  /// or is a version that doesn't offer thumbnails.
+  /// It will also be null when [loading] is true.
+  final String? thumbnailUrl;
+
+  /// A flag to indicate whether to show the placeholder.
+  ///
+  /// Typically it will be `true` while Server is generating thumbnails.
+  final bool loading;
+
   @override
   bool operator ==(Object other) {
-    return other is ImageNode && other.srcUrl == srcUrl;
+    return other is ImageNode
+      && other.srcUrl == srcUrl
+      && other.thumbnailUrl == thumbnailUrl
+      && other.loading == loading;
   }
 
   @override
-  int get hashCode => Object.hash('ImageNode', srcUrl);
+  int get hashCode => Object.hash('ImageNode', srcUrl, thumbnailUrl, loading);
 
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
     properties.add(StringProperty('srcUrl', srcUrl));
+    properties.add(StringProperty('thumbnailUrl', thumbnailUrl));
+    properties.add(FlagProperty('loading', value: loading, ifTrue: "is loading"));
   }
 }
 
@@ -1014,7 +1039,7 @@ class _ZulipContentParser {
 
   BlockContentNode parseImageNode(dom.Element divElement) {
     assert(_debugParserContext == _ParserContext.block);
-    final imgElement = () {
+    final elements = () {
       assert(divElement.localName == 'div'
           && divElement.className == 'message_inline_image');
 
@@ -1028,21 +1053,51 @@ class _ZulipContentParser {
       final grandchild = child.nodes[0];
       if (grandchild is! dom.Element) return null;
       if (grandchild.localName != 'img') return null;
-      if (grandchild.className.isNotEmpty) return null;
-      return grandchild;
+      return (child, grandchild);
     }();
 
     final debugHtmlNode = kDebugMode ? divElement : null;
-    if (imgElement == null) {
+    if (elements == null) {
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
 
+    final (linkElement, imgElement) = elements;
+    final href = linkElement.attributes['href'];
+    if (href == null) {
+      return UnimplementedBlockContentNode(htmlNode: divElement);
+    }
+    if (imgElement.className == 'image-loading-placeholder') {
+      return ImageNode(
+        srcUrl: href,
+        thumbnailUrl: null,
+        loading: true,
+        debugHtmlNode: debugHtmlNode);
+    }
     final src = imgElement.attributes['src'];
     if (src == null) {
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
 
-    return ImageNode(srcUrl: src, debugHtmlNode: debugHtmlNode);
+    final String srcUrl;
+    final String? thumbnailUrl;
+    if (src.startsWith('/user_uploads/thumbnail/')) {
+      srcUrl = href;
+      thumbnailUrl = src;
+    } else if (src.startsWith('/external_content/')
+        || src.startsWith('https://uploads.zulipusercontent.net/')) {
+      srcUrl = src;
+      thumbnailUrl = null;
+    } else if (href == src)  {
+      srcUrl = src;
+      thumbnailUrl = null;
+    } else {
+      return UnimplementedBlockContentNode(htmlNode: divElement);
+    }
+    return ImageNode(
+      srcUrl: srcUrl,
+      thumbnailUrl: thumbnailUrl,
+      loading: false,
+      debugHtmlNode: debugHtmlNode);
   }
 
   static final _videoClassNameRegexp = () {
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
@@ -571,26 +572,31 @@ class MessageImage extends StatelessWidget {
     final message = InheritedMessage.of(context);
 
     // TODO image hover animation
-    final src = node.srcUrl;
-
+    final srcUrl = node.srcUrl;
+    final thumbnailUrl = node.thumbnailUrl;
     final store = PerAccountStoreWidget.of(context);
-    final resolvedSrc = store.tryResolveUrl(src);
+    final resolvedSrcUrl = store.tryResolveUrl(srcUrl);
+    final resolvedThumbnailUrl = thumbnailUrl == null
+      ? null : store.tryResolveUrl(thumbnailUrl);
+
     // TODO if src fails to parse, show an explicit "broken image"
 
     return MessageMediaContainer(
-      onTap: resolvedSrc == null ? null : () { // TODO(log)
+      onTap: resolvedSrcUrl == null ? null : () { // TODO(log)
         Navigator.of(context).push(getLightboxRoute(
           context: context,
           message: message,
-          src: resolvedSrc,
+          src: resolvedSrcUrl,
           mediaType: MediaType.image));
       },
-      child: resolvedSrc == null ? null : LightboxHero(
-        message: message,
-        src: resolvedSrc,
-        child: RealmContentNetworkImage(
-          resolvedSrc,
-          filterQuality: FilterQuality.medium)));
+      child: node.loading
+        ? const CupertinoActivityIndicator()
+        : resolvedSrcUrl == null ? null : LightboxHero(
+            message: message,
+            src: resolvedSrcUrl,
+            child: RealmContentNetworkImage(
+              resolvedThumbnailUrl ?? resolvedSrcUrl,
+              filterQuality: FilterQuality.medium)));
   }
 }
 
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -587,6 +587,7 @@ class MessageImage extends StatelessWidget {
           context: context,
           message: message,
           src: resolvedSrcUrl,
+          thumbnailUrl: resolvedThumbnailUrl,
           mediaType: MediaType.image));
       },
       child: node.loading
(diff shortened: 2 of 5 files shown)


lib/model/content.dart
```diff
@@ -351,26 +351,51 @@ class ImageNodeList extends BlockContentNode {
 }
 
 class ImageNode extends BlockContentNode {
-  const ImageNode({super.debugHtmlNode, required this.srcUrl});
+  const ImageNode({
+    super.debugHtmlNode,
+    required this.srcUrl,
+    required this.thumbnailUrl,
+    required this.loading,
+  });
```
gnprice, line 359: Let's split this into two constructors:
```suggestion
  const ImageNode({
    super.debugHtmlNode,
    required String this.srcUrl,
    required String this.thumbnailUrl,
  }) : loading = false;
```

and then an `ImageNode.loading` that makes those other fields null.

That way the constructors enforce that there are these two mutually exclusive configurations.

lib/model/content.dart
```diff
@@ -1028,21 +1053,46 @@ class _ZulipContentParser {
...
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
 
+    final (linkElement, imgElement) = elements;
+    if (imgElement.className == 'image-loading-placeholder') {
+      return ImageNode(
+        srcUrl: null,
+        thumbnailUrl: null,
+        loading: true,
+        debugHtmlNode: debugHtmlNode);
```
gnprice, line 1070: ~~so in particular:~~
```suggestion
      return ImageNode.loading(debugHtmlNode: debugHtmlNode);
```

lib/model/content.dart
gnprice: … Actually, though, scratch that — because I think we want to model this a bit differently.

Even if the thumbnail is still pending, that's no reason to stop the user from opening the lightbox. So let's always have the original URL, which is the one from the `a` element.

lib/model/content.dart
```diff
@@ -1028,21 +1053,46 @@ class _ZulipContentParser {
...
+    }
     final src = imgElement.attributes['src'];
     if (src == null) {
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
 
-    return ImageNode(srcUrl: src, debugHtmlNode: debugHtmlNode);
+    final String srcUrl;
+    final String? thumbnailUrl;
+    if (src.startsWith('/user_uploads/thumbnail/')) {
```
gnprice, line 1084: When this condition is false, let's have a check that `src == href`, and return unimplemented if they're unequal.

I believe those were always the same when thumbnailing isn't involved; if there are other cases that exist, I'd like to learn about them, either in manual testing as we use the app or as part of #190.

gnprice (commented):
Thanks @rajveermalviya! Taking this straight to integration review because it interacts with the server's 9.0 release timeline.

Here's a partial review before I go AFK for dinner. I've read just the model code so far (lib/model/content.dart).

test/model/content_test.dart
```diff
@@ -459,8 +485,38 @@ class ContentExample {
...
-      ImageNode(srcUrl: 'https://uploads.zulipusercontent.net/8f63bc2632a0e41be3f457d86c077e61b4a03e7e/68747470733a2f2f636861742e7a756c69702e6f72672f757365725f617661746172732f322f7265616c6d2f69636f6e2e706e673f76657273696f6e3d34'),
+      ImageNode(srcUrl: 'https://uploads.zulipusercontent.net/f535ba07f95b99a83aa48e44fd62bbb6c6cf6615/68747470733a2f2f636861742e7a756c69702e6f72672f757365725f617661746172732f322f7265616c6d2f69636f6e2e706e673f76657273696f6e3d33',
+        thumbnailUrl: null, loading: false),
+      ImageNode(srcUrl: 'https://uploads.zulipusercontent.net/8f63bc2632a0e41be3f457d86c077e61b4a03e7e/68747470733a2f2f636861742e7a756c69702e6f72672f757365725f617661746172732f322f7265616c6d2f69636f6e2e706e673f76657273696f6e3d34',
+        thumbnailUrl: null, loading: false),
+    ]),
+  ]);
+
+  static const imageClusterWithThumbnails = ContentExample(
+    'multiple images with thumbnails',
```
gnprice, line 496: This can get the same "NoThumbnails" treatment as the single-image case above.

gnprice (commented):
Thanks again @rajveermalviya for building this! Done with a review — comments below. Generally it looks great.

The comments are mostly small, but there are a couple that I think might mean glitches a user could spot:
* on _buildAppBarFooter and the height of the app bar
* on addPostFrameCallback vs. scheduleFrameCallback

lib/model/content.dart
rajveermalviya: There seems to be multiple cases where `src != href`, forexample src urls starting with `/external_content/` and `https://uploads.zulipusercontent.net/`.
Should they be added one by explicitly (using `startsWith`)?

lib/widgets/lightbox.dart
```diff
@@ -234,17 +252,53 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
 
   @override
   Widget build(BuildContext context) {
+    final thumbnailUrl = widget.thumbnailUrl;
     return _LightboxPageLayout(
       routeEntranceAnimation: widget.routeEntranceAnimation,
       message: widget.message,
+      buildAppBarFooter: _loadingProgress == null ? null : _buildAppBarFooter,
```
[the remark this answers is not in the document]
rajveermalviya: It doesn't cause the layout shift, because `_LightboxPageLayout` sets [`Scaffold.extendBodyBehindAppBar`](https://api.flutter.dev/flutter/material/Scaffold/extendBodyBehindAppBar.html) to `true` [here](https://github.com/zulip/zulip-flutter/blob/7285325a3068b40e1feed9db4909aef04d31946f/lib/widgets/lightbox.dart#L189).

lib/widgets/content.dart
```diff
@@ -571,26 +572,32 @@ class MessageImage extends StatelessWidget {
...
           mediaType: MediaType.image));
       },
-      child: resolvedSrc == null ? null : LightboxHero(
-        message: message,
-        src: resolvedSrc,
-        child: RealmContentNetworkImage(
-          resolvedSrc,
-          filterQuality: FilterQuality.medium)));
+      child: node.loading
+        ? const CupertinoActivityIndicator()
```
rajveermalviya, line 594: Also forgot to flag this earlier, the reason for not using the `loader-{black,white}.svg` asset here is that, currently animated icons can't be included in ZulipIcons font (don't know if animated fonts are even possible), I tried and it just displayed a single (first) frame.

So, for now used the `CupertinoActivityIndicator` widget which conveniently is similar to the svg.

rajveermalviya:
Thanks for the review @gnprice, pushed a new revision PTAL :)

lib/model/content.dart
rajveermalviya: > they be added one by one explicitly
Did this for now.

lib/widgets/lightbox.dart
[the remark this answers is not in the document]
gnprice: Ah right, that makes sense, thanks.

lib/widgets/content.dart
gnprice: Handy that that's there!

Yeah, I suspect animated fonts are not possible. If we were determined to use the icons from those SVGs, we could find a way (if nothing else, we could rasterize them to high-enough-resolution PNGs). But using this widget instead is a lot simpler and SGTM.

Would you post a screenshot and a video showing what this placeholder looks like? That doesn't require any fiddling on the server — the quick hack I used just now is:
```diff
--- lib/model/content.dart
+++ lib/model/content.dart
@@ -1064,7 +1064,7 @@ class _ZulipContentParser {
     if (href == null) {
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
-    if (imgElement.className == 'image-loading-placeholder') {
+    if (true || imgElement.className == 'image-loading-placeholder') {
       return ImageNode(
         srcUrl: href,
         thumbnailUrl: null,
```

lib/model/content.dart
```diff
@@ -351,26 +351,49 @@ class ImageNodeList extends BlockContentNode {
...
+  /// This may be a relative URL string. It also may not work without adding
   /// authentication credentials to the request.
   final String srcUrl;
 
+  /// The thumbnail URL of the image.
+  ///
+  /// This may be a relative URL string. It also may not work without adding
+  /// authentication credentials to the request.
+  /// It may be null if Server didn't generate a thumbnail, or doesn't support
+  /// it yet. It will also be null when [loading] is true.
```
gnprice, line 372:
```suggestion
  ///
  /// This will be null if the server hasn't yet generated a thumbnail,
  /// or is a version that doesn't offer thumbnails.
  /// It will also be null when [loading] is true.
```

(in particular I found the meaning of "yet" a bit hard to parse in the current revision)

lib/widgets/lightbox.dart
```diff
@@ -209,17 +216,30 @@ class _ImageLightboxPage extends StatefulWidget {
...
   @override
   State<_ImageLightboxPage> createState() => _ImageLightboxPageState();
 }
 
 class _ImageLightboxPageState extends State<_ImageLightboxPage> {
+  double? _loadingProgress;
+
+  PreferredSizeWidget? _buildAppBarBottom(BuildContext context) {
+    if (_loadingProgress == null) {
+      return null;
```
gnprice, line 236: Cool, moving this here (from the call site) is a nice simplification too.

lib/widgets/lightbox.dart
```diff
@@ -232,19 +252,54 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
...
+
+    // `loadingProgress` becomes null when Image has finished downloading.
+    final double? progress = loadingProgress?.expectedTotalBytes == null ? null
+      : loadingProgress!.cumulativeBytesLoaded / loadingProgress.expectedTotalBytes!;
+
+    if (progress != _loadingProgress) {
+      _loadingProgress = progress;
+      // This function is called in a build method and setState
+      // can't be called in a build method, so delay it.
+      SchedulerBinding.instance.scheduleFrameCallback((_) { if (mounted) setState(() {}); });
```
gnprice, line 277: nit: lots going on on this line (lots of control flow, with two function literals and a conditional); split it up

In fact I think it benefits from being expanded slightly more, to turn the if-mounted check into an early return to match the pattern it usually appears as:

```suggestion
      SchedulerBinding.instance.scheduleFrameCallback((_) {
        if (!mounted) return;
        setState(() {});
      });
```

lib/widgets/lightbox.dart
```diff
@@ -232,19 +252,54 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
...
+    if (widget.thumbnailUrl == null) return child;
+
+    // `loadingProgress` becomes null when Image has finished downloading.
+    final double? progress = loadingProgress?.expectedTotalBytes == null ? null
+      : loadingProgress!.cumulativeBytesLoaded / loadingProgress.expectedTotalBytes!;
+
+    if (progress != _loadingProgress) {
+      _loadingProgress = progress;
+      // This function is called in a build method and setState
+      // can't be called in a build method, so delay it.
```
gnprice, line 276: Given the discussion in #mobile-team about how these delayed callbacks are generally to be avoided, let's include in the comment an indication of why in this case it's the best available option. That'll help the reader not wonder if we just didn't think it through; if they want to try to find a solution we missed, it'll let them see the information we'd found so they can start from there.

So:
```suggestion
      // The [Image.network] API lets us learn progress information only at
      // its build time.  That's too late for updating the progress indicator,
      // so delay that update to the next frame.  For discussion, see:
      //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/addPostFrameCallback/near/1893539
      //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/addPostFrameCallback/near/1894124
```

gnprice (commented):
Thanks @rajveermalviya for the revision! Just a few nits in this round, below; plus please post a screenshot and video as per https://github.com/zulip/zulip-flutter/pull/820#discussion_r1683626383 above.

lib/model/content.dart
```diff
@@ -1028,21 +1051,51 @@ class _ZulipContentParser {
...
 
-    return ImageNode(srcUrl: src, debugHtmlNode: debugHtmlNode);
+    final String srcUrl;
+    final String? thumbnailUrl;
+    if (src.startsWith('/user_uploads/thumbnail/')) {
+      srcUrl = href;
+      thumbnailUrl = src;
+    } else if (src.startsWith('/external_content/')
+      || src.startsWith('https://uploads.zulipusercontent.net/')) {
+      srcUrl = src;
```
gnprice, line 1086: nit: indent `if` condition's continuation to not look like part of the body:
```suggestion
    } else if (src.startsWith('/external_content/')
        || src.startsWith('https://uploads.zulipusercontent.net/')) {
      srcUrl = src;
```

lib/model/content.dart
gnprice: Yep, that sounds good.

Oh, it looks like we don't have a test case that exercises the `/external_content/` case, though. Do you have an example? That'd be good to add.

gnprice (commented):
Oh and two more comments — I'd missed some of the changes in my reading a minute ago. One is a still-smaller nit; the other asks for one more test case.

rajveermalviya:

lib/widgets/content.dart
rajveermalviya: Added a video showing placeholders in the PR description.

test/model/content_test.dart
```diff
@@ -420,12 +421,52 @@ class ContentExample {
...
+        '<img class="image-loading-placeholder" src="/static/images/loading/loader-black.svg"></a></div>', [
+    ImageNodeList([
+      ImageNode(srcUrl: '/user_uploads/2/c3/wb9FXk8Ej6qIc28aWKcqUogD/image.jpg',
+        thumbnailUrl: null, loading: true),
+    ]),
+  ]);
+
+  static const imageSingleExternal = ContentExample(
+    'single image external',
+    "https://upload.wikimedia.org/wikipedia/commons/7/78/Verregende_bloem_van_een_Helenium_%27El_Dorado%27._22-07-2023._%28d.j.b%29.jpg",
```
gnprice, line 463: Is there an example message for this on chat.zulip.org which you can link to?

gnprice (commented):
Thanks for the revision! Everything looks good, modulo one question below.

@alya see the placeholders demonstrated above (and any frame of that video works as a screenshot). They're slightly different from the design we have in web, because this is easiest for reasons discussed at https://github.com/zulip/zulip-flutter/pull/820#discussion_r1682277072 . Given the release timeline (and because this version seems fine) I won't block merging this to wait for feedback on the design, but we can always adjust it later if desired.

test/model/content_test.dart
rajveermalviya: Ha, yeah I'd just noticed I forgot to add that and pushed a new revision with that.

gnprice:
Merged! Thanks again @rajveermalviya for the fast work building this and the swift revisions.

PR #822, opened 2024-07-17
msglist: Handle UserTopicEvent, hiding/showing messages as needed

gnprice:
In particular this will allow us to start offering UI for muting
and unmuting topics, and have the message list the user is looking at
update appropriately when they do so.

Fixes: #421

---

The first two commits are taken from #787 (/cc @PIG208):
aef0ab1eb test [nfc]: Allow overriding original stream and add checks.
9c50ba57e message: Handle moved messages from UpdateMessageEvent.

So they can be ignored in reviewing this PR, and we'll plan to merge #787 before this one.

### DIFF
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -311,6 +311,179 @@ void main() {
     check(model).fetched.isFalse();
   });
 
+  group('UserTopicEvent', () {
+    // The ChannelStore.willChangeIfTopicVisible/InStream methods have their own
+    // thorough unit tests.  So these tests focus on the rest of the logic.
+
+    final stream = eg.stream();
+    const String topic = 'foo';
+
+    Future<void> setVisibility(UserTopicVisibilityPolicy policy) async {
+      await store.handleEvent(eg.userTopicEvent(stream.streamId, topic, policy));
+    }
+
+    /// (Should run after `prepare`.)
+    Future<void> prepareMutes([
+      bool streamMuted = false,
+      UserTopicVisibilityPolicy policy = UserTopicVisibilityPolicy.none,
+    ]) async {
+      await store.addStream(stream);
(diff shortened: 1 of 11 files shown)


chrisbobbe:
This has gathered some conflicts; could you resolve those please?

gnprice:
Sure — done.

chrisbobbe:
Thanks! LGTM, but I think we intend to wait for #787 before merging this.

test/model/message_list_test.dart
```diff
@@ -311,6 +311,179 @@ void main() {
     check(model).fetched.isFalse();
   });
 
+  group('UserTopicEvent', () {
+    // The ChannelStore.willChangeIfTopicVisible/InStream methods have their own
+    // thorough unit tests.  So these tests focus on the rest of the logic.
+
+    final stream = eg.stream();
+    const String topic = 'foo';
```
PIG208, line 319: nit: Is it intentional that `topic` is only used in `setVisibility`? It might be helpful to reuse it to indicate elsewhere in the tests that this particular topic is affected by the helper.

PIG208 (commented):
Went through the PR. Looks good to me too! Just one nit question.

test/model/message_list_test.dart
gnprice: Yeah, thanks — that would be clearer.

gnprice:
Thanks for the reviews!

Now that #787 is merged, I've rebased this and addressed the comment above; PTAL.

chrisbobbe:
Thanks, LGTM! Just awaiting @PIG208, then.

PIG208:
Other than the example_data conflict, this looks good to me!

gnprice:
Thanks! Rebased; merging.

PR #827, opened 2024-07-19
Move markNarrowAsRead to new "actions" file; port most of its tests

gnprice:
This is a refactoring we'll want for #131 / #779, as discussed at https://github.com/zulip/zulip-flutter/pull/779#issuecomment-2237737547 .

This doesn't cover actually refactoring `markNarrowAsRead` to be reusable for marking as unread — @Khader-1 that'll be for you to do as part of #779. But this sets up the framework for doing that refactor.

The two main commits' messages:

---

actions [nfc]: Move markNarrowAsRead to new file

This function isn't quite part of the UI -- it's mostly a driver
around some API requests.  But it doesn't belong in the API code
either, because it interacts with UI widgets in order to give
progress and failure feedback to the user.

Because there's nontrivial logic in here and it isn't really specific
to this particular piece of UI, we'd like to reuse much of this code,
in particular for #131 marking as unread (as discussed at #779).
This new file provides an appropriate home for that reusable code.

Before we can actually reuse this for marking as unread, it'll need
some refactoring.  Before we undertake refactoring it, it should have
tests that operate at the layer of this function itself, so that they
can easily be adapted to cover the mark-unread case too.  There are
existing tests that cover this pretty thoroughly, but they act on the
UI in the message list that drives it.  So we'll port those tests over
in the next couple of commits.

---

actions test: Port most MarkAsReadWidget tests to markNarrowAsRead

This sets us up to be able to refactor this function for other
use cases that aren't exercised by MarkAsReadWidget, and have good
tests for those cases too.

There are a couple of pieces of functionality of the widget's
_handlePress method that probably should be within the function
markNarrowAsRead instead, so that they can be reused together.
For the tests for those, we port them too but mark them as
skipped, with TODO comments.  After refactoring to move that
functionality inside, we can enable the tests.

The new markNarrowAsRead unit tests do most of the work of tests
for MarkAsReadWidget.  We leave behind a couple of smoke tests
that mostly just show it is indeed using markNarrowAsRead, plus
the tests whose ported versions are skipped.

### DIFF
--- a/test/widgets/actions_test.dart
+++ b/test/widgets/actions_test.dart
@@ -1,0 +1,284 @@
+import 'dart:convert';
+
+import 'package:checks/checks.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:http/http.dart' as http;
+import 'package:zulip/api/model/initial_snapshot.dart';
+import 'package:zulip/api/model/narrow.dart';
+import 'package:zulip/api/route/messages.dart';
+import 'package:zulip/model/localizations.dart';
+import 'package:zulip/model/narrow.dart';
+import 'package:zulip/model/store.dart';
+import 'package:zulip/widgets/actions.dart';
+import 'package:zulip/widgets/store.dart';
+import 'package:zulip/widgets/theme.dart';
+
+import '../api/fake_api.dart';
+import '../example_data.dart' as eg;
+import '../model/binding.dart';
+import '../model/unreads_checks.dart';
+import '../stdlib_checks.dart';
+import 'dialog_checks.dart';
+
+void main() {
+  TestZulipBinding.ensureInitialized();
+
+  group('markNarrowAsRead', () {
+    late PerAccountStore store;
+    late FakeApiConnection connection;
+    late BuildContext context;
+
+    Future<void> prepare(WidgetTester tester, {
+      UnreadMessagesSnapshot? unreadMsgs,
+    }) async {
+      addTearDown(testBinding.reset);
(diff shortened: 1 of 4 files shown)


test/widgets/actions_test.dart
```diff
@@ -0,0 +1,284 @@
...
+        unreadMsgs: unreadMsgs));
+      store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+      connection = store.connection as FakeApiConnection;
+
+      await tester.pumpWidget(Builder(builder: (context) =>
+        MaterialApp(
+          theme: zulipThemeData(context),
+          localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+          supportedLocales: ZulipLocalizations.supportedLocales,
+          home: GlobalStoreWidget(
```
gnprice, line 47: This makes another spot that could benefit from TestZulipApp, introduced in #805. So whichever of these PRs lands first, the other can update this spot after rebasing.

test/widgets/message_list_test.dart
```diff
@@ -1006,128 +966,6 @@ void main() {
...
-      });
-
-      testWidgets('CombinedFeedNarrow on legacy server', (WidgetTester tester) async {
-        const narrow = CombinedFeedNarrow();
-        await setupMessageListPage(tester,
-          narrow: narrow, messages: [message], unreadMsgs: unreadMsgs);
-        check(isMarkAsReadButtonVisible(tester)).isTrue();
-
-        // Might as well test with oldUnreadsMissing: true.
-        store.unreads.oldUnreadsMissing = true;
```
chrisbobbe, line 1048: The analyzer is flagging an unused import '../model/unreads_checks.dart'.

chrisbobbe (commented):
Thanks! LGTM except for one nit; please merge at will after fixing that.

gnprice:
Ah indeed, thanks. Fixed and merged.

Issue #188, opened 2023-06-14
Offer a list of all streams, and a way to subscribe


gnprice:
Sometimes one wants to navigate to a stream when not already subscribed, or to subscribe to a stream.

In zulip-mobile this is offered on the "All streams" screen, reached from the bottom of the "Streams" tab.

We should offer equivalent functionality, whether through a similar design or something different.

gnprice:
Yes indeed — this screen should show all streams. (Well, all streams that the user is able to see.)

Cross-referencing, we have a more detailed discussion in this chat thread:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/all-streams.20screen.20.23F188/near/1761243

gnprice:
This feature now has a design in Figma (thanks Vlad!). I've updated the description to reflect that.

chrisbobbe:
When this is done we should update the Help Center doc. We removed instructions for the legacy app when preparing to launch the new app: zulip/zulip@977c3943d198d1f9482f9c7f695e1f356fd94c98

alya:
Requested [in the FHIR community](https://chat.fhir.org/#narrow/channel/179238-Zulip/topic/search.20in.20stream.3F/near/526273818).

gnprice:
We had a couple of requests for this feature today:
* [#mobile > ✔ flutter: browse channels](https://chat.zulip.org/#narrow/channel/48-mobile/topic/.E2.9C.94.20flutter.3A.20browse.20channels/with/2226839)
* #1733

alya:
Another request in the Rust community: [#zulip > new mobile app feedback @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/122653-zulip/topic/new.20mobile.20app.20feedback/near/531274313)

chrisbobbe:
Another request for this on CZO: [#mobile-design > Mobile app channel list @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Mobile.20app.20channel.20list/near/2230749)

PR #832, opened 2024-07-23
[WIP] All streams screen

Khader-1:
Fixes: #188


lib/widgets/channel_list.dart
```diff
@@ -34,12 +34,32 @@ class _ChannelListPageState extends State<ChannelListPage> with PerAccountStoreA
     return Scaffold(
       appBar: AppBar(title: Text(zulipLocalizations.channelListPageTitle)),
       body: SafeArea(
-        child: ListView.builder(
+        child: streams.isEmpty ? const _NoStreamsItem() : ListView.builder(
```
rajveermalviya, line 42: This page is currently only opened through `SubscriptionListPage`, So if that page just doesn't show the button when there are no non-subscribed streams, this would only occur if `ChannelListPage` was already opened and new events on the store removed all the streams. So this is a good case to have.

lib/widgets/channel_list.dart
```diff
@@ -101,6 +104,55 @@ class ChannelItem extends StatelessWidget {
...
+    final scaffoldMessenger = ScaffoldMessenger.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    try {
+      final res = await subscribeToChannels(connection, [stream]);
+      if (!context.mounted) return;
+      scaffoldMessenger.clearSnackBars();
+      if (_emailSubscriptionsContains(store, res.subscribed, stream.name)) {
+        scaffoldMessenger.showSnackBar(SnackBar(behavior: SnackBarBehavior.floating,
+          content: Text(zulipLocalizations.messageSubscribedToChannel(stream.name))));
+      } else if (_emailSubscriptionsContains(store, res.alreadySubscribed, stream.name)) {
```
rajveermalviya, line 185: Not sure if this case is needed

rajveermalviya (changes requested):
Thanks for working on this @Khader-1!

Overall looks great, just some comments.

lib/widgets/subscription_list.dart
```diff
@@ -106,7 +108,7 @@ class _SubscriptionListPageState extends State<SubscriptionListPage> with PerAcc
               _SubscriptionList(unreadsModel: unreadsModel, subscriptions: unpinned),
             ],
 
-            // TODO(#188): add button leading to "All Streams" page with ability to subscribe
+            const _ChannelListLinkItem(),
```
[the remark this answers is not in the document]
gnprice: Good question. Chat thread: https://chat.zulip.org/#narrow/stream/48-mobile/topic/All.20channels.20screen.20-.20.23F832/near/1906519

Khader-1:
Thanks @rajveermalviya! I've pushed a new revision PTAL.

assets/l10n/app_en.arb
```diff
@@ -363,6 +363,13 @@
       "num": {"type": "int", "example": "4"}
     }
   },
+  "browseMoreNChannels": "Browse {num, plural, =0{all channels} =1{1 more channel} other{{num} more channels}}",
```
rajveermalviya, line 366: Neat!

rajveermalviya (approved):
Thanks for the revision!
This looks good to me, moving on to the mentor review from @kenclary.

kenclary (approved):
lgtm

alya:
Let's keep the empty view message consistent with the web app, unless there's a specific reason to do otherwise. So, it would be:

> There are no channels you can view in this organization.

Screenshot 2024-08-07 at 14 23 51@2x

alya:
We use sentence case throughout the Zulip app. So, "All channels", rather than "All Channels".

alya:
Can you please include screenshots with the subscribed/unsubscribed messages? It's difficult to review text on a video.

Khader-1:
Thanks @rajveermalviya, @kenclary and @alya for the reviews! Moving on for maintainer review now.
I've pushed a new revision and updated appended screenshots, PTAL @alya.

gnprice:
-    channel_list: Setup subscribeToChannels api binding
    +    api: Add route subscribeToChannels

(Thanks for making this update! I definitely appreciate when someone takes a comment from one PR and proactively applies it to other open PRs where it'd be relevant.)

chrisbobbe:
@alya some of the snackbar messages have exclamation marks in them; do we want to keep those? Not sure what our pattern is for ending punctuation. Maybe they should even all have a period.

lib/widgets/channel_list.dart
```diff
@@ -0,0 +1,74 @@
...
+                  // TODO(#95) need dark-theme color
+                  color: Color(0xFF262626)),
+                maxLines: 1,
+                overflow: TextOverflow.ellipsis),
+                if (stream.description.isNotEmpty) Text(
+                  stream.description,
+                  style: const TextStyle(
+                    fontSize: 12,
+                    // TODO(#95) need dark-theme color
+                    color: Color(0xCC262626)),
```
PIG208, line 93: Do we want these colors go to one of the `ThemeExtension` classes we have? Maybe @chrisbobbe has an idea on that.

lib/widgets/subscription_list.dart
```diff
@@ -199,6 +201,41 @@ class _SubscriptionList extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final notShownStreams = store.streams.length - store.subscriptions.length;
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return SliverToBoxAdapter(
+      child: Material(
+        // TODO(#95) need dark-theme color
+        color: Colors.white,
```
PIG208, line 216: Similar question to the other comment on the color here.

test/widgets/channel_list_test.dart
```diff
@@ -0,0 +1,42 @@
...
+
+void main() {
+  TestZulipBinding.ensureInitialized();
+
+  Future<void> setupChannelListPage(WidgetTester tester, {
+    required List<ZulipStream> streams, required List<Subscription> subscriptions}) async {
+    addTearDown(testBinding.reset);
+    final initialSnapshot = eg.initialSnapshot(
+      subscriptions: subscriptions,
+      streams: streams.toList(),
```
PIG208, line 31: nit: Perhaps we can skip the `toList` call here too.

test/widgets/subscription_list_test.dart
```diff
@@ -20,11 +21,12 @@ void main() {
     required List<Subscription> subscriptions,
     List<UserTopicItem> userTopics = const [],
     UnreadMessagesSnapshot? unreadMsgs,
+    List<ZulipStream>? streams,
   }) async {
     addTearDown(testBinding.reset);
     final initialSnapshot = eg.initialSnapshot(
       subscriptions: subscriptions,
-      streams: subscriptions,
+      streams: streams ?? subscriptions,
```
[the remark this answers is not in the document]
PIG208: It also looks like a reasonable choice, because in most cases when `streams` is not given, the test probably does not care about its value.

chrisbobbe (changes requested):
Thanks! Comments below.

Also, since we now have a dark theme (since #867 🎉), please update this to provide a dark-theme variant. Since there isn't something we can really follow in the Figma or the web app, I expect each color will need one of these TODOs:

```dart
// TODO(design-dark) need proper dark-theme color (this is ad hoc)
```

(if you've chosen the dark-theme color yourself)

or

```dart
// TODO(design) check if this is the right variable
```

(if you've used a Figma design variable)

For examples, see c6abaf923.

test/widgets/subscription_list_test.dart
```diff
@@ -57,6 +59,23 @@ void main() {
...
   });
 
+  testWidgets('link to other channels is shown', (tester) async {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+    final streams = List.generate(5, (index) => eg.stream());
+    await setupStreamListPage(tester,
+      streams: streams,
+      subscriptions: [eg.subscription(streams[1])]);
+
+    check(find.text(zulipLocalizations.browseMoreNChannels(4)).evaluate()).isNotEmpty();
```
PIG208, line 69: We can use the actual localized strings we expect here, because it will also be a bug if the pluralization does not work as expected.

test/widgets/subscription_list_test.dart
```diff
@@ -57,6 +59,23 @@ void main() {
...
+  });
+
+  testWidgets('link to other channels is not shown if server has no visible channels', (tester) async {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+    await setupStreamListPage(tester, streams: [], subscriptions: []);
+
+    check(find.text(zulipLocalizations.browseMoreNChannels(0)).evaluate()).isEmpty();
```
PIG208, line 76: Since this string is different from the one in the other test, we could have another positive case in comparison.

test/widgets/channel_list_test.dart
```diff
@@ -0,0 +1,42 @@
...
+    await tester.pumpAndSettle();
+  }
+
+  int getItemCount() {
+    return find.byType(ChannelItem).evaluate().length;
+  }
+
+  testWidgets('smoke', (tester) async {
+    await setupChannelListPage(tester, streams: [], subscriptions: []);
+    check(getItemCount()).equals(0);
```
PIG208, line 50: We can further extract the `.equals` part into the helper, turn it into `checkItemCount`. We should also probably do something like `check(...).length.equals` instead of better error message.

lib/widgets/channel_list.dart
```diff
@@ -24,12 +24,32 @@ class ChannelListPage extends StatelessWidget {
...
       appBar: AppBar(title: Text(zulipLocalizations.channelListPageTitle)),
       body: SafeArea(
-        child: ListView.builder(
+        child: streams.isEmpty ? const _NoStreamsItem() : ListView.builder(
           itemCount: streams.length,
           itemBuilder: (context, index) => ChannelItem(stream: streams[index]))));
   }
 }
 
+class _NoStreamsItem extends StatelessWidget {
```
PIG208, line 38: This could be `_NoChannelsItem`.

lib/api/route/channel.dart
```diff
@@ -29,3 +29,23 @@ class SubscribeToChannelsResult {
   Map<String, dynamic> toJson() => _$SubscribeToChannelsResultToJson(this);
 }
 
+Future<UnsubscribeFromChannelsResult> unsubscribeFromChannels(ApiConnection connection, List<ZulipStream> streams) {
```
PIG208, line 32: Let's link to the relevant API documentation here.

lib/api/route/channel.dart
```diff
@@ -29,3 +29,23 @@ class SubscribeToChannelsResult {
...
+  return connection.delete('unsubscribe', UnsubscribeFromChannelsResult.fromJson,
+    'users/me/subscriptions', {'subscriptions': streams.map((e) => e.name).toList()});
+}
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class UnsubscribeFromChannelsResult {
+  final List<String>? removed;
+  final List<String>? notRemoved;
```
PIG208, line 40: Could you double-check if these can be `null`?

lib/model/channel.dart
```diff
@@ -168,6 +168,11 @@ class ChannelStoreImpl with ChannelStore {
 
       case SubscriptionRemoveEvent():
         for (final streamId in event.streamIds) {
+          assert(streams.containsKey(streamId)
+            && streams[streamId] is Subscription);
```
PIG208, line 172: We could break the `&&`'s into their own lines, so it is clearer when only one of them fails.

PIG208 (commented):
Read the commits up to d16b58f703681a98c4072f863be884358c5f6bab, and left some comments.

lib/widgets/channel_list.dart
gnprice: Yeah, now that #95 is done, all new pieces of UI should support dark theme from the beginning. That means UI code gets colors from `DesignVariables` or another `ThemeExtension`, and colors for new UI get added there, with both light-theme and dark-theme values.

test/widgets/subscription_list_test.dart
gnprice: Yeah — when there's nontrivial logic around what the string should look like, that's a good case for having the test (or at least one test) write the actual content it expects from the string, rather than calling the same ZulipLocalizations method that the code under test will be calling.

alya:
For the confirmation notices:

1. We have a pattern of always putting either a privacy icon (this is best if it's not too tricky), or at least a `#`, before the name of a channel.
2. Where did the strings for these notices come from? I would probably tweak some of them, but would like to keep them consistent if there's a source for them. At a minimum, we should consistently put punctuation at the end -- the unsuccessful ones are missing a `.`.

alya:
For the exception ones, it seems like we're showing too much to the user. If it's not going to cause us significant difficulties investigating, I'd leave off the everything in the parentheses.

lib/model/channel.dart
```diff
@@ -173,6 +173,8 @@ class ChannelStoreImpl with ChannelStore {
           assert(streamsByName.containsKey(streams[streamId]!.name)
             && streamsByName[streams[streamId]!.name] is Subscription);
           assert(subscriptions.containsKey(streamId));
+          streams[streamId] = subscriptions[streamId]!.toStream();
+          streamsByName[streams[streamId]!.name] = subscriptions[streamId]!.toStream();
```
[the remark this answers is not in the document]
Khader-1: Well, when you unsubscribe from a channel and then try to resubscribe to it again, the assertion that you can't subscribe to a stream that is of type `Subscription` fails.

assets/l10n/app_en.arb
```diff
@@ -363,6 +363,13 @@
...
+  "@browseMoreNChannels": {
+    "description": "Label showing the number of other channels that user can subscribe to",
```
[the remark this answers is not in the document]
PIG208: Because now we have a separate rule, the `=0` variant can be dropped.

chrisbobbe (commented):
Thanks for the review! Some small nits from me, then I think @PIG208 will want to check this new revision against his review. Also see Alya's open comments at https://github.com/zulip/zulip-flutter/pull/832#issuecomment-2303594964 and right after.

lib/api/route/channel.dart
PIG208: We should use dart-doc for the API documentation link like we do for other routes.

test/widgets/channel_list_test.dart
```diff
@@ -0,0 +1,70 @@
...
+    await testBinding.globalStore.add(eg.selfAccount, initialSnapshot);
+
+    await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id, child: const ChannelListPage()));
+
+    // global store, per-account store
+    await tester.pumpAndSettle();
+  }
+
+  void checkItemCount(int expectedCount) {
+    check(find.byType(ChannelItem).evaluate().length).equals(expectedCount);
```
PIG208, line 46: nit:
```suggestion
    check(find.byType(ChannelItem).evaluate()).length.equals(expectedCount);
```

PIG208 (commented):
Thanks for the revision! Just took a quick look again. Reopened two previous comments and left another small one.

Khader-1:
Thanks @alya, @chrisbobbe and @PIG208 for the reviews!

> We have a pattern of always putting either a privacy icon (this is best if it's not too tricky), or at least a #, before the name of a channel.

I tried to use privacy icon but I could not find a direct way to inline the icon within the message text with proper wrapping. I'd love to hear any suggestion from Chris or Zixuan on how to do that.

> Where did the strings for these notices come from?

I wrote them down. I tried to get the same as strings as mobile but I believe mobile shows a snackbar only if it fails to subscribe.

alya:
> I wrote them down. I tried to get the same as strings as mobile but I believe mobile shows a snackbar only if it fails to subscribe.

Ok, please start a discussion in #mobile regarding what these strings should be. You can post a list of situations where messages are shown and your suggested strings (with `.` added, since we definitely want that).

chrisbobbe:
> I tried to use privacy icon but I could not find a direct way to inline the icon within the message text with proper wrapping. I'd love to hear any suggestion from Chris or Zixuan on how to do that.

Try using a `Text.rich` instead of a `Text`, and put the privacy icon in a `WidgetSpan`. We have a few uses of `WidgetSpan` in message content: user mentions, image emojis, and global times.

Khader-1:
Thanks @alya, @chrisbobbe and @PIG208! I'm finally back with another revision, PTAL.

lib/widgets/channel_list.dart
```diff
@@ -100,6 +103,82 @@ class ChannelItem extends StatelessWidget {
...
+      SnackBar(
+        behavior: SnackBarBehavior.floating,
+        content: Text.rich(TextSpan(text: message, children: [
+          WidgetSpan(child: Padding(
+            padding: const EdgeInsets.only(bottom: 2),
+            child: Icon(
+              size: 12,
+              iconDataForStream(stream),
+              color: Theme.of(context).scaffoldBackgroundColor))),
+          TextSpan(text: ' ${stream.name}.', style: weightVariableTextStyle(context, wght: 700)),
```
PIG208, line 205: Oh, I see. Regarding the earlier comment on the translated string, this wouldn't work for all languages because the order might be different. We can just keep it simple and style the message the same way.

test/widgets/channel_list_test.dart
```diff
@@ -67,4 +82,159 @@ void main() {
...
     });
   });
+
+  group('subscription toggle', () {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+
+    Future<ZulipStream> prepareSingleStream(WidgetTester tester) async {
+        final stream = eg.stream();
+        await setupChannelListPage(tester, streams: [stream], subscriptions: []);
+        return stream;
```
PIG208, line 92: nit: indentation

test/widgets/channel_list_test.dart
```diff
@@ -67,4 +82,159 @@ void main() {
...
+      final richTextFinder = find.byWidgetPredicate(
+        (widget) => widget is RichText && widget.text.toPlainText().contains(message));
+      check(richTextFinder.evaluate()).single;
+      final richTextWidget = tester.widget<RichText>(richTextFinder);
+      check(richTextWidget.text.toPlainText()).contains(message);
+      check(richTextWidget.text.toPlainText()).contains(channelName);
+      check(find.descendant(
+        of: richTextFinder,
+        matching: find.byIcon(ZulipIcons.hash_sign),
+      ).evaluate()).single;
```
PIG208, line 121: nit: use `findsOne` as mentioned earlier

test/widgets/channel_list_test.dart
```diff
@@ -67,4 +82,159 @@ void main() {
...
+
+      check(find.byIcon(Icons.add).evaluate()).isNotEmpty();
+    });
+
+    testWidgets('is disabled while loading', (WidgetTester tester) async {
+      final stream = eg.stream();
+      await setupChannelListPage(tester, streams: [stream], subscriptions: []);
+      connection.prepare(json: SubscribeToChannelsResult(
+        subscribed: {eg.selfUser.email: [stream.name]},
+        alreadySubscribed: {}).toJson());
```
PIG208, line 152: We could add a delay to make this more realistic.

lib/widgets/channel_list.dart
```diff
@@ -0,0 +1,105 @@
...
+      child: Padding(
+        padding: const EdgeInsets.all(10),
+        child: Text(zulipLocalizations.noChannelsFound,
+          textAlign: TextAlign.center,
+          style: TextStyle(
+            // TODO(design) check if this is the right variable
+            color: designVariables.subscriptionListHeaderText,
+            fontSize: 18,
+            height: (20 / 18),
+          ))));
```
PIG208, line 60:
```suggestion
            height: (20 / 18)))));
```

lib/widgets/subscription_list.dart
```diff
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
```
PIG208, line 2: We will need to regenerate the translation strings using `tools/check l10n --fix`, and update this import to use `import '../generated/l10n/zulip_localizations.dart';` instead.

PIG208 (commented):
Glad to have you back @Khader-1! Thanks for the update! I have read through the commits and left some comments.

I think the last two commits can be squashed together:
* f6045303 channel_list: Implement unsubscribe from channel
* 793e9bf4 channel_list: Implement subscribe to channel

And similarly, these two can be squashed together:
* cc676ac9 api: Add route unsubscribeFromChannels
* 96836fc8 api: Add route subscribeToChannels

lib/widgets/channel_list.dart
```diff
@@ -0,0 +1,105 @@
...
+    return MaterialAccountWidgetRoute(accountId: accountId, context: context,
+      page: const ChannelListPage());
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final streams = store.streams.values.toList()..sort((a, b) {
+      return a.name.toLowerCase().compareTo(b.name.toLowerCase());
```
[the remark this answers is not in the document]
gnprice: Yeah, this is definitely fine for now. We have a much bigger computation of the same kind in InboxPage, and I happened to do some measurements the other day and found that it's a minor issue: #1065 

So this much smaller computation is at most a minor issue, and potentially not a problem at all.

What would be good to add is just a one-line TODO(perf) comment, like:
```dart
    // TODO(perf) memoize sorting the channels, outside of build phase
```

Then in the future we might come back and measure, and either (a) remove the TODO and leave a comment saying we decided it's OK, or (b) remove the TODO and memoize the computation.

lib/widgets/channel_list.dart
gnprice: Yeah, to make this work properly we'd want to end up writing something like:
```dart
Text.rich(zulipLocalizations.errorFailedToSubscribeToChannel(
  channel: TextSpan(children: [
    WidgetSpan(/* … */),
    TextSpan(text: ' ${channel.name}', /* … bold … */),
  ]))
```

Then the string for translation would look like
"Failed to subscribe to {channel}."

and the generated implementation for `errorFailedToSubscribeToChannel`, for English, would look like
```dart
  TextSpan errorFailedToSubscribeToChannel({required InlineSpan channel}) {
    return TextSpan(children: [
      const TextSpan(text: 'Failed to subscribe to '),
      channel,
      const TextSpan(text: '.'),
    ]);
  }
```

Flutter's tooling for l10n doesn't currently have that feature, though. There's an open issue for it:
* https://github.com/flutter/flutter/issues/147974

gnprice: One workaround would be to make the string for translation be:
"Failed to subscribe to channel: "

with a colon. Then we can show the translated version of that, followed by the fancy formatted spans, and no period.

Basically by making the message be less in the format of normal prose, and more in the format of a form like something printed out by a computer system, we ensure the translator can make something where the channel name (and attached icon) naturally goes at the end. In different languages the place that the object of a verb like "subscribe" goes, or that the whole relative clause "to subscribe …" goes in a sentence with "failed", can vary and may not come at the end… but every language (that anyone would translate our UI into) is used to dealing with forms and bureaucracies and computers, and has ways to express a line in a computery bureaucratic form.

gnprice:
Thanks again @Khader-1 for all your work on this PR!

Do you think you'll have time to make a revision based on Zixuan's feedback above?

gnprice:
Marking this PR as draft for now to get it out of our review queue. @Khader-1 if in the future you're able to pick this up again and revise for the feedback above, feel free to mark again for review. Hope you're well.

PR #833, opened 2024-07-23
 lightbox: Display thumbnail with original image sizing 

### LINKED ISSUE #830: Show thumbnail at main image's size in lightbox
With #820 (implementing #799), when you open an image in the lightbox we immediately show the thumbnail (which we'll typically already have loaded, from when you saw the preview in the message list) while we're loading the full image.

One wrinkle is that for some images this means that the thumbnail appears smaller than the full image will, so there's a pop where it expands when the fetch finishes. This is still better than not showing the image at all (at least in cases where the fetch is more than perhaps a few hundred milliseconds), which is why we do it.

But with an extra feature that's now being added in the server, we can do better: the image HTML will encode the original image's dimensions, so we can use those to make sure we show the thumbnail at the same size we'll eventually show the full image. Let's use that.

* `#api design` thread: https://chat.zulip.org/#narrow/stream/378-api-design/topic/thumbnailing.20dimensions/near/1898379
* Server PR: https://github.com/zulip/zulip/pull/31041

rajveermalviya:
https://github.com/user-attachments/assets/0a007765-0879-4124-9f1b-2cf597518dac

(Initial image opens are older messages missing `data-original-dimensions` attribute)

Fixes #830

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -356,6 +356,8 @@ class ImageNode extends BlockContentNode {
     required this.srcUrl,
     required this.thumbnailUrl,
     required this.loading,
+    required this.originalWidth,
+    required this.originalHeight,
   });
 
   /// The canonical source URL of the image.
@@ -379,16 +381,25 @@ class ImageNode extends BlockContentNode {
   /// Typically it will be `true` while Server is generating thumbnails.
   final bool loading;
 
+  /// The width of the canonical image.
+  final double? originalWidth;
+
+  /// The height of the canonical image.
+  final double? originalHeight;
+
   @override
   bool operator ==(Object other) {
     return other is ImageNode
       && other.srcUrl == srcUrl
       && other.thumbnailUrl == thumbnailUrl
-      && other.loading == loading;
+      && other.loading == loading
+      && other.originalWidth == originalWidth
+      && other.originalHeight == originalHeight;
   }
 
   @override
-  int get hashCode => Object.hash('ImageNode', srcUrl, thumbnailUrl, loading);
+  int get hashCode => Object.hash('ImageNode',
+    srcUrl, thumbnailUrl, loading, originalWidth, originalHeight);
 
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
@@ -396,6 +407,8 @@ class ImageNode extends BlockContentNode {
     properties.add(StringProperty('srcUrl', srcUrl));
     properties.add(StringProperty('thumbnailUrl', thumbnailUrl));
     properties.add(FlagProperty('loading', value: loading, ifTrue: "is loading"));
+    properties.add(DoubleProperty('originalWidth', originalWidth));
+    properties.add(DoubleProperty('originalHeight', originalHeight));
   }
 }
 
@@ -1037,6 +1050,8 @@ class _ZulipContentParser {
     return CodeBlockNode(spans, debugHtmlNode: debugHtmlNode);
   }
 
+  static final _imageDimensionsRegExp = RegExp(r'^(\d+)x(\d+)$');
+
   BlockContentNode parseImageNode(dom.Element divElement) {
     assert(_debugParserContext == _ParserContext.block);
     final elements = () {
@@ -1071,6 +1086,8 @@ class _ZulipContentParser {
         srcUrl: href,
         thumbnailUrl: null,
         loading: true,
+        originalWidth: null,
+        originalHeight: null,
         debugHtmlNode: debugHtmlNode);
     }
     final src = imgElement.attributes['src'];
@@ -1093,10 +1110,32 @@ class _ZulipContentParser {
     } else {
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
+
+    double? originalWidth, originalHeight;
+    final originalDimensions = imgElement.attributes['data-original-dimensions'];
+    if (originalDimensions != null) {
+      // Server encodes this string as "{width}x{height}" (eg. "300x400")
+      final match = _imageDimensionsRegExp.firstMatch(originalDimensions);
+      if (match != null) {
+        final width = int.tryParse(match.group(1)!, radix: 10);
+        final height = int.tryParse(match.group(2)!, radix: 10);
+        if (width != null && height != null) {
+          originalWidth = width.toDouble();
+          originalHeight = height.toDouble();
+        }
+      }
+
+      if (originalWidth == null || originalHeight == null) {
+        return UnimplementedBlockContentNode(htmlNode: divElement);
+      }
+    }
+
     return ImageNode(
       srcUrl: srcUrl,
       thumbnailUrl: thumbnailUrl,
       loading: false,
+      originalWidth: originalWidth,
+      originalHeight: originalHeight,
       debugHtmlNode: debugHtmlNode);
   }
 
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -576,7 +576,9 @@ class MessageImage extends StatelessWidget {
           context: context,
           message: message,
           src: resolvedSrcUrl,
-          thumbnailUrl: resolvedThumbnailUrl));
+          thumbnailUrl: resolvedThumbnailUrl,
+          originalWidth: node.originalWidth,
+          originalHeight: node.originalHeight));
       },
       child: node.loading
         ? const CupertinoActivityIndicator()
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -572,12 +572,11 @@ class MessageImage extends StatelessWidget {
 
     return MessageMediaContainer(
       onTap: resolvedSrcUrl == null ? null : () { // TODO(log)
-        Navigator.of(context).push(getLightboxRoute(
+        Navigator.of(context).push(getImageLightboxRoute(
           context: context,
           message: message,
           src: resolvedSrcUrl,
-          thumbnailUrl: resolvedThumbnailUrl,
-          mediaType: MediaType.image));
+          thumbnailUrl: resolvedThumbnailUrl));
       },
       child: node.loading
         ? const CupertinoActivityIndicator()
@@ -603,12 +602,10 @@ class MessageInlineVideo extends StatelessWidget {
 
     return MessageMediaContainer(
       onTap: resolvedSrc == null ? null : () { // TODO(log)
-        Navigator.of(context).push(getLightboxRoute(
+        Navigator.of(context).push(getVideoLightboxRoute(
           context: context,
           message: message,
-          src: resolvedSrc,
-          thumbnailUrl: null,
-          mediaType: MediaType.video));
+          src: resolvedSrc));
       },
       child: Container(
         color: Colors.black, // Web has the same color in light and dark mode.
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -217,12 +217,16 @@ class _ImageLightboxPage extends StatefulWidget {
     required this.message,
     required this.src,
     required this.thumbnailUrl,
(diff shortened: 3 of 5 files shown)


lib/model/content.dart
```diff
@@ -1093,10 +1109,32 @@ class _ZulipContentParser {
     } else {
       return UnimplementedBlockContentNode(htmlNode: divElement);
     }
+
+    double? originalWidth, originalHeight;
+    final originalDimensions = imgElement.attributes['data-original-dimensions'];
+    if (originalDimensions != null) {
+      // Server encodes this string as "{width}x{height}" (eg. "300x400")
+      final cutsets = originalDimensions.split('x');
```
gnprice, line 1117: As part of my usual performance paranoia for the content parser, let's avoid `split` and allocating a list here: instead can find the index of the first 'x' and parse the substrings before and after (if there's a second `x`, it'll cause the int parsing to fail), or can use a regexp like `^(\d+)x(\d+)$`.

gnprice (commented):
Great, thanks @rajveermalviya for building this!

About to go into a meeting, but here's a couple of comments from beginning to read the PR.

lib/model/content.dart
```diff
@@ -379,23 +381,35 @@ class ImageNode extends BlockContentNode {
   /// Typically it will be `true` while Server is generating thumbnails.
   final bool loading;
 
+
+  /// The width of the canonical image.
```
gnprice, line 385: nit: extra blank line

lib/widgets/lightbox.dart
```diff
@@ -583,3 +586,43 @@ Route<void> getLightboxRoute({
     },
   );
 }
+
+Route<void> getImageLightboxRoute({
+  int? accountId,
+  BuildContext? context,
+  required Message message,
+  required Uri src,
+  required Uri? thumbnailUrl,
```
gnprice, line 595: nit: include trailing comma in intermediate commit

(that helps reduce the diff both for that commit and the one after it)

lib/widgets/content.dart
```diff
@@ -603,12 +604,10 @@ class MessageInlineVideo extends StatelessWidget {
...
     return MessageMediaContainer(
       onTap: resolvedSrc == null ? null : () { // TODO(log)
-        Navigator.of(context).push(getLightboxRoute(
+        Navigator.of(context).push(getVideoLightboxRoute(
           context: context,
           message: message,
-          src: resolvedSrc,
-          thumbnailUrl: null,
-          mediaType: MediaType.video));
+          src: resolvedSrc));
```
gnprice, line 610: Cool, I like this refactor splitting into two different entry-point functions. The resulting API feels more thoroughly well-typed.

lib/widgets/lightbox.dart
```diff
@@ -258,9 +262,28 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
     // The full image is available, so display it.
     if (frame != null) return child;
 
+    // Use `BoxFit.contain` if the original image size is larger than
+    // the device screen size, otherwise `null` (default).
```
gnprice, line 266: What's the behavior of the default? If we can it's probably best to pass an explicit value that specifies the behavior we want — if nothing else, that should make it easier to reason about this logic where we're choosing between different values for this option.

lib/widgets/lightbox.dart
```diff
@@ -258,9 +262,28 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
...
+    // This allows the thumbnail image to mimic the same behaviour as
+    // when the original image is loaded, which is, keep small images
+    // small (and in the center) and large images contained in the
+    // viewport.
+    BoxFit? fit;
+    final originalWidth = widget.originalWidth;
+    final originalHeight = widget.originalHeight;
+    if (originalWidth != null && originalHeight != null) {
+      final physicalSize =
+        MediaQuery.sizeOf(context) * MediaQuery.devicePixelRatioOf(context);
```
gnprice, line 277: Hmm, the meaning of `MediaQuery.sizeOf` is that it gives the size of the whole screen.

That's often the same as the size available to this widget, but it may not be — in particular we have a `SafeArea` around this, so I believe if you use a device with a notched display then this will give a larger size than what's actually available to this widget. That seems like it could lead the below logic astray.

Ideally we'd specify a single value for this option that tells `Image` to do the same thing that you explain in the comment above. Is this `BoxFit.scaleDown`?
https://main-api.flutter.dev/flutter/painting/BoxFit.html

gnprice: Ah I see, but then it's not clear there's a good way to tell `Image` about `originalWidth` and `originalHeight`.

In that case, how about we handle the sizing of the image directly? I think putting it in a SizedBox inside a FittedBox, or something along those lines, should be able to control the size it's displayed at.

lib/widgets/lightbox.dart
```diff
@@ -258,9 +262,28 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
     // The full image is available, so display it.
     if (frame != null) return child;
 
+    // Use `BoxFit.contain` if the original image size is larger than
+    // the device screen size, otherwise `null` (default).
+    //
+    // This allows the thumbnail image to mimic the same behaviour as
+    // when the original image is loaded, which is, keep small images
+    // small (and in the center) and large images contained in the
+    // viewport.
```
gnprice, line 271: Let's add a test case or two for this behavior, since it seems like nontrivial logic.

I see we don't have any existing tests for the image lightbox. We don't have to add tests for the rest of its behavior now; but it'd be good to test this, and in particular to have a test exercising the situation I mentioned in another comment with a notched display.

If I find time today I can try to add a couple of basic tests for the existing layout of the image lightbox, to help set the framework for the new tests.

gnprice (commented):
OK, finished a review — comments below.

lib/widgets/lightbox.dart
gnprice: Sent that as #835. So you can rebase this PR atop that one, and then add the new couple of test cases.

lib/model/content.dart
```diff
@@ -1093,10 +1110,32 @@ class _ZulipContentParser {
...
     }
+
+    double? originalWidth, originalHeight;
+    final originalDimensions = imgElement.attributes['data-original-dimensions'];
+    if (originalDimensions != null) {
+      // Server encodes this string as "{width}x{height}" (eg. "300x400")
+      final match = _imageDimensionsRegExp.firstMatch(originalDimensions);
+      if (match != null) {
+        final width = int.tryParse(match.group(1)!);
+        final height = int.tryParse(match.group(2)!);
```
gnprice, line 1121: There's a subtle and unfortunate gotcha in Dart's `int.parse` and `int.tryParse`, which it's best to explicitly disable:
```suggestion
        final width = int.tryParse(match.group(1)!, radix: 10);
        final height = int.tryParse(match.group(2)!, radix: 10);
```

For discussion, see f5477717aec34590bee82298db3a64800bb14877.

In this case I believe it doesn't have any effect on the behavior, because we already know from the regexp that each group matches exactly `\d+` and so can't trigger this gotcha. But that's a pretty subtle and nonlocal property, so it's best to avoid the gotcha more directly too: that way (a) the gotcha won't easily be reintroduced by an otherwise-innocent change elsewhere in this logic, and (b) it maintains the pattern for the benefit of anyone copy-pasting or adapting this code for somewhere else in the codebase.

gnprice: (Separately I was tempted to say `tryParse` can be just `parse`, because we already know they match exactly `\d+`. But if we got an image with something like `data-original-dimensions="1234123412341234123412341234x5"`, the parse would indeed fail because the number is too big for an int. So `tryParse` is more correct — better to return unimplemented than to throw — and it's good we have it.)

lib/widgets/lightbox.dart
```diff
@@ -259,8 +263,14 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
...
 
     // Display the thumbnail image while original image is downloading.
-    return RealmContentNetworkImage(widget.thumbnailUrl!,
-      filterQuality: FilterQuality.medium);
+    return FittedBox(
+      fit: BoxFit.scaleDown,
+      child: SizedBox(
+        width: widget.originalWidth,
+        height: widget.originalHeight,
+        child: RealmContentNetworkImage(widget.thumbnailUrl!,
```
gnprice, line 271: Cool, this seems nice and clean.

gnprice (commented):
Thanks for the revision!

I see you hadn't marked this as ready for the next round of review, but took a look at it anyway just to help keep it moving along given its connection to the imminent server-9.0 release.

One small comment below, and otherwise everything here looks good — all it needs is those couple of new test cases as discussed above.

test/widgets/lightbox_test.dart
```diff
@@ -271,9 +298,97 @@ void main() {
...
+    //   tester.widgetList(find.descendant(
+    //     of: find.byType(RealmContentNetworkImage),
+    //     matching: find.byType(RawImage)),
+    //   ).single;
+
+    //   // Image is 100x100, it should stay that size and shouldn't scale up.
+    //   final renderImage = tester.renderObject<RenderImage>(find.descendant(
+    //     of: find.byType(RealmContentNetworkImage),
+    //     matching: find.byType(RawImage)));
+    //   check(renderImage.size).equals(const Size(100, 100));
```
rajveermalviya, line 386: I can't get this test to pass because seems like flutter test enviroment doesn't report correct image sizes?

test/widgets/lightbox_test.dart
```diff
@@ -271,9 +298,97 @@ void main() {
...
+    //   // Both visible
+    //   tester.widget(find.byWidgetPredicate(
+    //     (widget) => widget is RealmContentNetworkImage && widget.src == thumbnail));
+    //   tester.widget(find.byWidgetPredicate(
+    //     (widget) => widget is RealmContentNetworkImage && widget.src == src));
+
+    //   print('pump');
+    //   await tester.pump();
+    //   print('done pump');
+    //   tester.widgetList<RealmContentNetworkImage>(find.byType(RealmContentNetworkImage)).forEach((w) => print(w.src));
```
rajveermalviya, line 332: Same for this test, the frame arg in `_frameBuilder` never changes from `null`, even though `_loadingBuilder`'s `progress` become 1.0.

rajveermalviya (commented):
Thanks for the review @gnprice, I've addressed your comments in the latest revision.

But haven't gotten tests to work correctly yet.

gnprice:
Thanks. Chat thread for debugging those tests: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23F833.20lightbox.20tests/near/1902807

gnprice:
OK, merging this without those tests, in order to get it into a v0.0.18 beta today since the server 9.0 release just came out. Thanks again @rajveermalviya!

I made a couple of commit-message tweaks:
```
$ git range-diff origin pr/833 @
…
      ## Commit message ##
    -    model: Parse original image dimensions
    +    content: Parse original image dimensions
…
      ## Commit message ##
    -    lightbox [nfc]: Refactor `getLightboxRoute`
    +    lightbox [nfc]: Split getLightboxRoute for images vs. videos
```

The second commit can be a bit more specific about what changed, since "[nfc]" already means it was a refactor; and for the first, the summary prefix can be more specific in that we usually mark a change to lib/model/foo.dart as "foo:" rather than "model:".

I definitely appreciate the work you've done already to write those draft tests, and would be glad to get tests along those lines into the tree. So please go ahead and send that as a follow-up PR, using what we found in that chat thread linked above.

(And maybe we still won't have a test for the situation where the frame argument is never changing from null, since that remains mysterious. In that case we can leave a TODO comment for it, perhaps including a draft of the test just with `skip: true` so it doesn't run. The other tests will still be good to have.)

Issue #829, opened 2024-07-22
Specify content-type in uploading images


gnprice:
Currently when the user uploads a file from the compose box, we always set the content-type to `application/octet-stream`.

Instead, we should set the content-type to an appropriate specific value, based on what we know about the file.

Before Zulip Server 9.0 (and before Zulip Server in main until last week), this didn't matter because the server would look at the extension in the filename to decide how to handle the file. As part of the thumbnailing work, though, the server now relies on the content-type set by the client. The symptom is that images don't get displayed in the message at all.

Reported by @alexmv. [Chat thread here](https://chat.zulip.org/#narrow/stream/9-issues/topic/thumbnailing/near/1899037).

gnprice:
Implementation notes:

It looks like adding the content-type in the HTTP request will be straightforward. We'll need our `ApiConnection.postFileFromStream` to add a parameter for it, as well as its caller the `uploadFile` endpoint binding.

Then it's a matter of finding the right content-type to set. If the platform gives us a content-type, we should use that. If it doesn't, we can look at the filename's extension ourselves and make an inference; for comparison, see `getMimeTypeFromFileExtension` in `src/utils/url.js` in zulip-mobile. I'd guess that when getting an image or video from the camera, the platform tells us the content-type, but that when picking an existing file it may not — after all, it doesn't really know any more than the file's name and contents either.

PR #834, opened 2024-07-24
compose: Include content-type in image-upload requests

chrisbobbe:
While working on this I realized we didn't have any widget tests for the image-upload flows! So I added some. There's still more to test there, and there might be some helpful deduplication in the tests.

But anyway, manual testing on my iPhone showed that the user-visible symptom of #829 on CZO is resolved:

> The symptom is that images don't get displayed in the message at all.

Fixes: #829

### DIFF
--- a/lib/api/core.dart
+++ b/lib/api/core.dart
@@ -3,6 +3,7 @@ import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:http/http.dart' as http;
+import 'package:http_parser/http_parser.dart';
 
 import '../log.dart';
 import '../model/binding.dart';
@@ -192,10 +193,20 @@ class ApiConnection {
   }
 
   Future<T> postFileFromStream<T>(String routeName, T Function(Map<String, dynamic>) fromJson,
-      String path, Stream<List<int>> content, int length, {String? filename}) async {
+      String path, Stream<List<int>> content, int length,
+      {String? filename, String? contentType}) async {
     final url = realmUrl.replace(path: "/api/v1/$path");
+    MediaType? parsedContentType;
+    if (contentType != null) {
+      try {
+        parsedContentType = MediaType.parse(contentType);
+      } on FormatException {
+        // TODO log
+      }
+    }
     final request = http.MultipartRequest('POST', url)
-      ..files.add(http.MultipartFile('file', content, length, filename: filename));
+      ..files.add(http.MultipartFile('file', content, length,
+        filename: filename, contentType: parsedContentType));
     return send(routeName, fromJson, request);
   }
 
--- a/lib/api/route/messages.dart
+++ b/lib/api/route/messages.dart
@@ -262,9 +262,10 @@ Future<UploadFileResult> uploadFile(
   required Stream<List<int>> content,
   required int length,
   required String filename,
+  required String? contentType,
 }) {
   return connection.postFileFromStream('uploadFile', UploadFileResult.fromJson, 'user_uploads',
-    content, length, filename: filename);
+    content, length, filename: filename, contentType: contentType);
 }
 
 @JsonSerializable(fieldRename: FieldRename.snake)
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -1,4 +1,5 @@
 import 'package:device_info_plus/device_info_plus.dart' as device_info_plus;
+import 'package:file_picker/file_picker.dart' as file_picker;
 import 'package:firebase_core/firebase_core.dart' as firebase_core;
 import 'package:firebase_messaging/firebase_messaging.dart' as firebase_messaging;
 import 'package:flutter/foundation.dart';
@@ -10,6 +11,8 @@ import '../host/android_notifications.dart';
 import '../log.dart';
 import '../widgets/store.dart';
 import 'store.dart';
+
+export 'package:file_picker/file_picker.dart' show FilePickerResult, FileType, PlatformFile;
 
 /// Alias for [url_launcher.LaunchMode].
 typedef UrlLaunchMode = url_launcher.LaunchMode;
@@ -152,6 +155,15 @@ abstract class ZulipBinding {
 
   /// Wraps the [AndroidNotificationHostApi] constructor.
   AndroidNotificationHostApi get androidNotificationHost;
+
+  /// Pick files from the media library, via package:file_picker.
+  ///
+  /// This wraps [file_picker.pickFiles].
+  Future<file_picker.FilePickerResult?> pickFiles({
+    bool allowMultiple,
+    bool withReadStream,
+    file_picker.FileType type,
+  });
 }
 
 /// Like [device_info_plus.BaseDeviceInfo], but without things we don't use.
@@ -389,4 +401,17 @@ class LiveZulipBinding extends ZulipBinding {
 
   @override
   AndroidNotificationHostApi get androidNotificationHost => AndroidNotificationHostApi();
+
+  @override
+  Future<file_picker.FilePickerResult?> pickFiles({
+    bool allowMultiple = false,
+    bool withReadStream = false,
+    file_picker.FileType type = file_picker.FileType.any,
+  }) async {
+    return file_picker.FilePicker.platform.pickFiles(
+      allowMultiple: allowMultiple,
+      withReadStream: withReadStream,
+      type: type,
+    );
+  }
 }
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -4,6 +4,7 @@ import 'package:firebase_core/firebase_core.dart' as firebase_core;
 import 'package:firebase_messaging/firebase_messaging.dart' as firebase_messaging;
 import 'package:flutter/foundation.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
+import 'package:image_picker/image_picker.dart' as image_picker;
 import 'package:package_info_plus/package_info_plus.dart' as package_info_plus;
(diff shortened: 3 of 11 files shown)


test/widgets/compose_box_test.dart
```diff
@@ -238,4 +242,134 @@ void main() {
...
+        // (the request is checked more thoroughly in API tests)
+        check(connection.lastRequest!).isA<http.MultipartRequest>()
+          ..method.equals('POST')
+          ..files.single.which((it) => it
+            ..field.equals('file')
+            ..length.equals(12345)
+            ..filename.equals('image.jpg')
+            ..contentType.asString.equals('image/jpeg')
+            ..has<Future<List<int>>>((f) => f.finalize().toBytes(), 'contents')
+              .completes((it) => it.deepEquals(['asdf'.codeUnits].expand((l) => l)))
```
gnprice, line 311:
```suggestion
              .completes((it) => it.deepEquals('asdf'.codeUnits))
```

This is equivalent, right?

gnprice: Ah I see, but this makes it parallel to what the more general helpers in tests like of ApiConnection.postFileFromStream are doing. Sure.

test/widgets/compose_box_test.dart
```diff
@@ -238,4 +242,134 @@ void main() {
...
+          );
+        checkAppearsLoading(tester, true);
+
+        await tester.pump(const Duration(seconds: 1));
+        const expectedContent = '[image.jpg](/user_uploads/1/4e/m2A3MSqFnWRLUf9SaPzQ0Up_/image.jpg)\n\n';
+        checkAppearsLoading(tester, false);
+        check(composeBoxController.contentController.text).equals(expectedContent);
```
gnprice, line 318: nit: bit simpler if expected gets inlined:
```suggestion
        check(composeBoxController.contentController.text)
          .equals('[image.jpg](/user_uploads/1/4e/m2A3MSqFnWRLUf9SaPzQ0Up_/image.jpg)\n\n');
        checkAppearsLoading(tester, false);
```

plus it's less indented so a bit easier to read.

lib/api/core.dart
```diff
@@ -192,10 +193,20 @@ class ApiConnection {
   }
 
   Future<T> postFileFromStream<T>(String routeName, T Function(Map<String, dynamic>) fromJson,
-      String path, Stream<List<int>> content, int length, {String? filename}) async {
+      String path, Stream<List<int>> content, int length, {String? filename,
+      String? contentType}) async {
```
gnprice, line 197: nit:
```suggestion
      String path, Stream<List<int>> content, int length,
      {String? filename, String? contentType}) async {
```

otherwise `path` and `contentType` look like the same kind of parameter, and it's a bit subtle to see that one is positional and the other named

lib/api/core.dart
```diff
@@ -192,10 +193,20 @@ class ApiConnection {
...
     final url = realmUrl.replace(path: "/api/v1/$path");
+    MediaType? effectiveContentType;
+    if (contentType != null) {
+      try {
+        effectiveContentType = MediaType.parse(contentType);
```
gnprice, line 202: I think of the "effectiveFoo" naming seen in Flutter code as referring to a fallback cascade: e.g. it's the value passed on this widget, or else its default from the ambient theme, or else the value from the default theme.

This is a different sort of thing: it's the value given, but parsed. A good name could be `parsedContentType`.

lib/api/route/messages.dart
```diff
@@ -262,9 +262,10 @@ Future<UploadFileResult> uploadFile(
   required Stream<List<int>> content,
   required int length,
   required String filename,
+  required String? mimeType,
```
gnprice, line 265: Probably cleanest to call the parameter `contentType`.

In general we try to make our names match the names in the API. Here this doesn't correspond to anything mentioned in the API docs:
https://zulip.com/api/upload-file

but it does correspond to a `content-type:` header in the request. So that seems like the canonical name to use.

lib/widgets/compose_box.dart
```diff
@@ -574,7 +586,17 @@ Future<Iterable<_File>> _getFilePickerFiles(BuildContext context, FileType type)
 
   return result.files.map((f) {
     assert(f.readStream != null);  // We passed `withReadStream: true` to pickFiles.
-    return _File(content: f.readStream!, length: f.size, filename: f.name);
+    String? mimeType;
+    if (f.path != null) {
+      mimeType = lookupMimeType(f.path!,
+        headerBytes: f.bytes?.take(defaultMagicNumbersMaxLength).toList());
```
gnprice, line 592: Hmm, this upstream API seems a bit quirky: it requires the path, even though it won't actually even look at the path if it finds a magic number in `headerBytes`.

I suspect this is a consequence of someone moving a bit too quickly through this when updating the API for NNBD — that they should have made it explicitly accept `String? path`.

Anyway, if we do find ourselves with no path, we can get better behavior here if we ask this function to look at the magic numbers anyway. To do that, we can say something like `f.path ?? ''` — so the extension is `''`, which should be a value that doesn't map to any particular type so we still get `null` if no magic numbers match.

lib/widgets/compose_box.dart
```diff
@@ -659,7 +681,14 @@ class _AttachFromCameraButton extends _AttachUploadsButton {
     }
     final length = await result.length();
 
-    return [_File(content: result.openRead(), length: length, filename: result.name)];
+    return [_File(
+      content: result.openRead(),
+      length: length,
+      filename: result.name,
+      mimeType: result.mimeType
+        // TODO pass headerBytes (what's a good way to get that?)
```
gnprice, line 689: Looking at what we have on `result`, I think `result.openRead(0, defaultMagicNumbersMaxLength)` should get the data.

And then that's a Stream… poking around that API a bit, one solution is:
```dart
  final headerBytes = await result.openRead(0, defaultMagicNumbersMaxLength).expand((l) => l).toList();
```
(suitably formatted)

gnprice (commented):
Thanks for building this! Comments below, mostly small.

I also have a small follow-up commit I'll push to add to the branch:

    binding [nfc]: Fully wrap package:file_picker by re-exporting types
    
    This allows all our other files to avoid importing that package
    directly.

PTAL, and we can do the same for package:image_picker if that seems easy too. (The other approach is typedefs like our `UrlLaunchMode`, for when the upstream name seems too generic to put into our namespace unadorned.)

lib/widgets/compose_box.dart
gnprice: https://github.com/dart-lang/mime/issues/11

Looks like my guess about the NNBD migration being involved was wrong — that issue thread predates NNBD. Anyway, we have a workaround.

test/widgets/compose_box_test.dart
```diff
@@ -238,4 +242,134 @@ void main() {
...
+      final sendButtonElement = tester.element(sendButtonFinder);
+      final colorScheme = Theme.of(sendButtonElement).colorScheme;
+      final expectedForegroundColor = expected
+        ? colorScheme.onSurface.withOpacity(0.38)
+        : colorScheme.onPrimary;
+      check(sendButtonWidget.color).equals(expectedForegroundColor);
+    }
+
+    group('attach from media library', () {
+      void checkAppearsLoading(WidgetTester tester, bool expected) {
```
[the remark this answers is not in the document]
chrisbobbe: Oh oops, yeah, this is the same as in the parent group. An incomplete refactor; will fix.

test/widgets/compose_box_test.dart
chrisbobbe: I don't have a preference between the two ways; I think I just copy-pasted from the `ApiConnection.postFileFromStream` tests.

lib/widgets/compose_box.dart
chrisbobbe: Thanks for spotting this!

chrisbobbe:

lib/widgets/compose_box.dart
```diff
@@ -659,7 +686,27 @@ class _AttachFromCameraButton extends _AttachUploadsButton {
...
+    List<int>? headerBytes;
+    try {
+      headerBytes = await result.openRead(
+        0,
+        // Despite its dartdoc, [XFile.openRead] can throw if `end` is greater
+        // than the file's length. We can *probably* trust our `length` to be
+        // accurate, but it's nontrivial to verify. If it's inaccurate, we'd
+        // rather sacrifice this part of the MIME lookup than throw the whole
+        // upload. So, the try/catch.
+        min(defaultMagicNumbersMaxLength, length)
```
gnprice, line 698: Hmm, annoying. This workaround makes sense.

lib/model/binding.dart
```diff
@@ -152,6 +157,23 @@ abstract class ZulipBinding {
...
+  Future<file_picker.FilePickerResult?> pickFiles({
+    bool allowMultiple,
+    bool withReadStream,
+    file_picker.FileType type,
+  });
+
+  /// Pick files from the camera or media library, via package:image_picker.
+  ///
+  /// This wraps [image_picker.pickImage].
+  Future<image_picker.XFile?> pickImage({
```
gnprice, line 173: Handy that this also lets us avoid depending directly on `package:cross_file`.

gnprice (commented):
Thanks for the revision! All looks good — I'll give this some manual testing on Android, and then merge.

gnprice:
OK, on my Pixel 8 running Android 14 I reproduced the issue without the last commit here, and confirmed that with this PR it's fixed both for the camera and for picking an image. Merged.

PR #836, opened 2024-07-25
content [nfc]: Use Divider instead of DecoratedBox.

PIG208:
This should offer the same appearance for the spoiler divider despite being shorter.

Discovered while working on #823.

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -522,12 +522,8 @@ class _SpoilerState extends State<Spoiler> with TickerProviderStateMixin {
                   ]))),
               FadeTransition(
                 opacity: _animation,
-                child: const SizedBox(height: 0, width: double.infinity,
-                  child: DecoratedBox(
-                    decoration: BoxDecoration(
-                      border: Border(
-                        // Web has the same color in light and dark mode.
-                        bottom: BorderSide(width: 1, color: Color(0xff808080))))))),
+                // Web has the same color in light and dark mode.
+                child: const Divider(color: Color(0xff808080), height: 0)),
               SizeTransition(
                 sizeFactor: _animation,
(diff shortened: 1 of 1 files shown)


PIG208:
| Before | After |
| --- | --- |
| Screenshot from 2024-07-24 20-04-26 | Screenshot from 2024-07-24 20-04-16 |

<details>
<summary> side by side </summary>

image

</details>

chrisbobbe:
When I open the "Before" and "After" images in separate Chrome tabs, then switch between the tabs, I do see a difference in the appearance. I see more whitespace above the divider in the "After" image.

PIG208:
Made a side-by-side version after zooming in on browser tabs:

image

PIG208:
Oh I do see the difference visible when switching between tabs now. Let me investigate

PIG208:
Changed the height to 0. It should look identical to the old version now. An easy way to check in dev is through hot reloading with and without the commit.

| Before | After |
| - | - |
| image | Screenshot from 2024-07-25 16-48-31 |

chrisbobbe:
Sure, seems fine to me. Over to Greg for his review.

gnprice:
I think this refactor isn't helpful. Divider is a Material widget — its main job is that it pulls in a bunch of defaults from Material Design.

This overrides three of the five defaults (color, height, thickness — the thickness defaults to 1 but is effectively forced to 0 by the zero height); for the other two, we want indent of zero not because that's the Material 3 default but because it's our design. We often use Material widgets for things like buttons and app navigation, but I think for the message content we generally have a pretty firm idea of what we want the design to look like and don't want it to depend on the choices in Material.

PIG208:
Thanks for the explanation. Closing the PR.

PR #837, opened 2024-07-26
narrow: Rename StreamNarrow to ChannelNarrow

Khader-1:
Fixes parts of #631

### DIFF
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -168,7 +168,7 @@ void main() {
   });
 
   test('MentionAutocompleteView misc', () async {
-    const narrow = StreamNarrow(1);
+    const narrow = ChannelNarrow(1);
     final store = eg.store();
     await store.addUsers([eg.selfUser, eg.otherUser, eg.thirdUser]);
     final view = MentionAutocompleteView.init(store: store, narrow: narrow);
@@ -185,7 +185,7 @@ void main() {
 
   test('MentionAutocompleteView not starve timers', () {
     fakeAsync((binding) async {
-      const narrow = StreamNarrow(1);
+      const narrow = ChannelNarrow(1);
       final store = eg.store();
       await store.addUsers([eg.selfUser, eg.otherUser, eg.thirdUser]);
       final view = MentionAutocompleteView.init(store: store, narrow: narrow);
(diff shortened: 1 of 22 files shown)


test/model/autocomplete_test.dart
```diff
@@ -525,7 +525,7 @@ void main() {
         checkRankEqual(narrow, [users[3], users[4]]);
       });
 
-      test('StreamNarrow: stream recency > DM recency', () async {
+      test('ChannelNarrow: stream recency > DM recency', () async {
```
sm-sayedi, line 528: Based on the changes in the first commit, I think this belongs to that commit.

sm-sayedi (approved):
Thanks @Khader-1! One small comment below, otherwise LGTM.

Moving to integration review.

Khader-1:
Thanks @sm-sayedi!

test/model/unreads_test.dart
```diff
@@ -190,13 +190,13 @@ void main() {
         eg.streamMessage(stream: stream, topic: 'c', flags: []),
       ]);
       check(model.countInStream      (stream.streamId)).equals(5);
```
gnprice, line 192: nit: preserve alignment (here and below)
```suggestion
      check(model.countInStream       (stream.streamId)).equals(5);
```

gnprice (commented):
Thanks @Khader-1, and thanks @sm-sayedi for the previous review!

These changes look good, with one formatting nit below. They're not quite complete, though — there are a few other references that should be updated.

A useful way to search for such references is like this:
```
$ git grep -iP 'stream.?narrow'
```
(You can skip the `-P` if you set `git config --global grep.patternType perl`.)
At the tip of this branch, that finds three references to "stream narrow" that remain to update.

Khader-1:
Thanks @gnprice! I've pushed a new revision PTAL

test/model/unreads_test.dart
```diff
@@ -174,7 +174,7 @@ void main() {
       check(model.countInCombinedFeedNarrow()).equals(5);
     });
 
-    test('countInStream/Narrow', () async {
+    test('countInChannel/Narrow', () async {
```
gnprice, line 177: This change doesn't seem right — it's referring to a pair of identifiers, and…

oh I see, one of them changed and the other didn't.

These two method names are meant to parallel each other, because their meanings are so closely related. I'll add a followup commit renaming the other method.

gnprice (commented):
Thanks! Merging, with one added commit as mentioned below.

PR #838, opened 2024-07-26
channel_colors: Rename Stream to Channel

Khader-1:
Fixes parts of #631


test/widgets/channel_colors_checks.dart
```diff
@@ -3,7 +3,7 @@ import 'dart:ui';
 import 'package:checks/checks.dart';
 import 'package:zulip/widgets/channel_colors.dart';
 
-extension StreamColorSwatchChecks on Subject<ChannelColorSwatch> {
+extension ChannelColorSwatchChecks on Subject<ChannelColorSwatch> {
```
sm-sayedi, line 6: I think this will be better included in [`88dd4b1`](88dd4b18a46b9a82a195b17d03813d8945cd83a9): channel_colors [nfc]: Rename StreamColorSwatch to ChannelColorSwatch.

lib/widgets/channel_colors.dart
```diff
@@ -41,26 +41,26 @@ abstract class ChannelColorSwatches {
...
     if (identical(a, b)) return a;
 
-    return _StreamColorSwatchesLerped(a, b, t);
+    return _ChannelColorSwatchesLerped(a, b, t);
   }
 }
 
-class _StreamColorSwatchesLight extends ChannelColorSwatches {
-  _StreamColorSwatchesLight();
+class _ChannelColorSwatchesLight extends ChannelColorSwatches {
```
sm-sayedi, line 48: I am not sure, but I think this and all other subclasses of `ChannelColorSwatches` should also go in [`bdf6fe4`](bdf6fe406fcaf7404c360b7370ae65b0b399a2ef): channel_colors [nfc]: Rename StreamColorSwatches to ChannelColorSwatches.

sm-sayedi (commented):
Thanks @Khader-1. Two small comments below.

lib/widgets/channel_colors.dart
```diff
@@ -6,137 +6,137 @@ import 'package:flutter_color_models/flutter_color_models.dart';
...
   ///   <https://github.com/zulip/zulip/pull/27485>
-  Color get iconOnBarBackground => this[StreamColorVariant.iconOnBarBackground]!;
+  Color get iconOnBarBackground => this[ChannelColorVariant.iconOnBarBackground]!;
 
-  /// The background color of a bar representing a stream, like a recipient bar.
+  /// The background color of a bar representing a channel, like a recipient bar.
   ///
-  /// Use this in the message list, the "Inbox" view, and the "Streams" view.
-  Color get barBackground => this[StreamColorVariant.barBackground]!;
+  /// Use this in the message list, the "Inbox" view, and the "Channels" view.
```
gnprice, line 107: Hmm, this line ideally belonged in the commit that renamed that view in the UI. Well, good to catch it now.

test/widgets/channel_colors_test.dart
```diff
@@ -137,12 +137,12 @@ void main() {
...
       test('iconOnPlainBackground', () {
         void runCheck(int base, Color expected) {
-          check(StreamColorSwatch.light(base)).iconOnPlainBackground.equals(expected);
+          check(ChannelColorSwatch.light(base)).iconOnPlainBackground.equals(expected);
         }
 
         // Check against everything in ZULIP_ASSIGNMENT_COLORS
         // in <https://replit.com/@VladKorobov/zulip-topic-feed-colors#script.js>.
-        // (Skipping `streamColors` because there are 100+ of them.)
+        // (Skipping `channelColors` because there are 100+ of them.)
```
gnprice, line 145: This is referring to an identifier, so should be updated at the same time the identifier is.

In this case the identifier is in the linked piece of code on replit.com, which hasn't been updated (and presumably won't be, as it's a historical piece of showing-one's-work). So this should continue to refer to the name that actually appears there.

gnprice (commented):
Thanks @Khader-1, and thanks @sm-sayedi for the previous review!

I agree with Sayed's comments above. One other comment below.

lib/widgets/channel_colors.dart
Khader-1: Would it be better to have it in it's own NFC?

Khader-1:
Thanks @sm-sayedi and @gnprice for the reviews! I've pushed a new revision PTAL

sm-sayedi (approved):
Thanks for the revision @Khader-1. One nit below, otherwise LGTM.

Rename 6fd6246422b72ad06bea34fa6da061299808aa9c commit message to use `StreamColorSwatches` and `ChannelColorSwatches` instead of `StreamColorSwatch` and `ChannelColorSwatch`.
Also, I suggest to use the code markdown for the class names in the commit messages.

gnprice:
Thanks!

Just commit-message nits at this round. Here's the summary lines:
```
$ git uvr
9ef291dbe channel_colors [nfc]: Rename files to channel_colors
7722b5aaa channel_colors [nfc]: Rename StreamColorSwatch to ChannelColorSwatch
6fd624642 channel_colors [nfc]: Rename `StreamColorSwatch` to `ChannelColorSwatch`
86f09db15 channel_colors [nfc]: Rename References of ChannelColorSwatches to match new name
178a48132  channel_colors [nfc]: Rename StreamColorVariant to ChannelColorVariant
f865db109 channel_colors [nfc]: Rename stream to channel in comments/docs
```

(that's using an alias from [my gitconfig](https://gist.github.com/gnprice/5f71b63464a76bd7e0480f207b1bbff3); you'll get a similar view from any variation on `git log --oneline`, or from using [a graphical Git viewer](https://github.com/zulip/zulip-mobile/blob/2217c858e207f9f092651dd853051843c3f04422/docs/howto/git.md#graphical-client) as I definitely recommend doing.)

* As Sayed pointed out, one of those got singular names that should be plural.
* There's also a leading space now in a later commit message.
* Writing backticks around these names would be OK, but I like it a bit better without them. Generally in a Git summary line, and often in a commit message's body too, I'll use backticks on an identifier only if it might otherwise look confusingly like a normal word.

  Looking for examples using the command:

      $ git log --author greg --grep '`.*`'

  the last time I used backticks in a summary line looks to have been:

  f367a1f94cd1a75458ef0570240abdd64b1c4993 api: Fix missing `type` in some events' toJson

  where "type" without backticks could look like I was using the word "type", which would parse differently.

Khader-1:
Thanks @gnprice! this is very helpful!
I've pushed a new revision, PTAL.

gnprice:
Thanks! Looks good — merging.

PR #842, opened 2024-07-26
msglist: Follow web instead of Figma for message timestamp color

chrisbobbe:
Just because web has a dark-theme variant for this color, and the Figma doesn't yet.

(Web actually expresses this with separate `color` and `opacity` attributes. We just collapse those into a Color with the right alpha.)

| Before | After |
| --- | --- |
| 8EDAAFA3-80E8-4573-A90F-07EF730FFD7A | 3661485E-4E12-49CD-A90E-97EEB93720ED |


gnprice:
Thanks! It looks like this got subsumed into #843, which I've just merged.

Issue #668, opened 2024-05-13
Track realm settings


gnprice:
Zulip has a long list of realm settings. Many of them are relevant to the mobile app. This issue is for beginning to track a few of them; once we do that, it'll be easy to add individual settings when they come up.

The settings are documented as part of the main event that updates them:
https://zulip.com/api/get-events#realm-update_dict
Note there's a different event that can also update them:
https://zulip.com/api/get-events#realm-update
and so we need to support both.

The settings are also present on the initial snapshot, generally with a name `realm_foo` for a realm setting named `foo`.

For a model for how this can work, see our handling of user settings, in particular `UserSettingsName`. The goal is to set up a system where adding any one new setting involves adding only a small number of lines of code in a small number of places. The PR for this issue should end with a commit that demonstrates that, by adding one more setting, just like #261 did for user settings with its last commit [`4bcbffd52`](https://github.com/zulip/zulip-flutter/pull/261/commits/4bcbffd529663737c6e3480dfd477bfcd70aa7c8).

gnprice:
Pushing this back to the post-launch milestone — we can pull it forward again when we identify a specific feature that depends on it.

MritunjayTiwari14:
Can I please take up this issue? My approach is to first create a sealed class, RealmEvent, and then implement RealmUpdateDictEvent, which inherits from RealmEvent—similar to the Update events in `event.dart`. In RealmUpdateEvent, we would create an enum RealmProperty to identify the type of value received—similar to UserSettings.

To handle the `data` under RealmUpdateDictEvent, we would introduce the class RealmUpdateDictData in the model.

To process the realm/update event, as [suggested](https://chat.zulip.org/api/get-events#realm-update:~:text=A%20correct%20client%20implementation%20should%20convert%20these%20events%20into%20the%20corresponding%20realm/update_dict%20event%20and%20then%20process%20that.) by the server, by directly converting it into realm/update_dict on the client side. All of this would have been implemented under HandleRealmEvent. Which Handle realm event for all op values?

chrisbobbe:
I actually already have a draft for this issue which I hope to send soon, so this isn't a great issue to pick up.

Issue #268, opened 2023-08-14
Require Zulip Server 5


gnprice:
Prerequisite for this issue:
* #267

Currently the app works with any Zulip Server version 4.x or newer. This is documented in our README and reflects what's documented at https://zulip.readthedocs.io/en/latest/overview/release-lifecycle.html . With #267, we'll start giving an explicit error if the user tries to connect to an older server.

At some point we'll want to increase that minimum to 5.x. As recorded at https://blog.zulip.com/tag/major-releases/ , 4.0 was released in 2021-05, over 2 years ago, and 5.0 in 2022-03, approaching 1.5 years ago.

When we do, we can follow up by sweeping the codebase for `TODO(server-5)` comments and acting on all of those. That will be in the same spirit as https://github.com/zulip/zulip-mobile/pull/5100 and https://github.com/zulip/zulip-mobile/pull/5192 in the old zulip-mobile RN app. (And https://github.com/zulip/zulip-mobile/pull/5708, though that's a draft that was made moot by our migration to Flutter.)

---

There is one spot where we already don't fully support older than server 5.x: currently we ignore user settings on such servers, falling back to defaults. As I wrote at https://github.com/zulip/zulip-flutter/pull/261#discussion_r1289384485 (reviewing @chrisbobbe's PR which introduced support for user settings and skipped it for those old servers):
> Cool, sounds good. This does seem like a spot where properly supporting both old and new versions would be kind of a pain. (Would have been totally doable, if the timing were such that we had to — we'd probably use a `readValue` that would just be one more spot that needs another line or two of code per setting — but enough of a pain to not be worth it given that we don't have to.)

The assumption that relies on is that we will desupport server 4.x (or if not then go back and deal with the old settings API after all) before launch.

### References
* Reduced, pre-launch version of this issue: #1456

gnprice:
> There is one spot where we already don't fully support older than server 5.x

(But in general we should avoid introducing more such places for now. As I wrote at https://github.com/zulip/zulip-flutter/pull/238#discussion_r1275361091 :
> We'll probably bump that version higher sometime soon. But until we do that explicitly, I'd like to keep the code uniformly supporting the version we say we're supporting. That helps keep it manageable to be able to look at a piece of code and then at the API docs and confirm the code is behaving as it should, and to avoid accidentally leaving out something we need on newer versions.

)

gnprice:
[...]

PR #844, opened 2024-07-27
profile: Display user email

sm-sayedi:
<table>
  <tr>
    <td><image src="https://github.com/user-attachments/assets/58580678-4b63-4820-b4c4-f76371b66a9a" width="300"/></td>
    <td><image src="https://github.com/user-attachments/assets/3b6bf959-b1e3-4769-9ec8-59b433125b32" width="300"/></td>      
  </tr>
</table>

Fixes: #291

### DIFF
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -24,6 +24,16 @@ class InitialSnapshot {
 
   final List<CustomProfileField> customProfileFields;
 
+  /// The realm-level policy, on pre-FL 163 servers, for visibility of real email addresses.
+  ///
+  /// Search for "email_address_visibility" in https://zulip.com/api/register-queue.
+  ///
+  /// This field is removed in Zulip 7.0 (FL 163) and replaced with a user-level
+  /// setting:
+  ///   * https://zulip.com/api/update-settings#parameter-email_address_visibility
+  ///   * https://zulip.com/api/update-realm-user-settings-defaults#parameter-email_address_visibility
+  final EmailAddressVisibility? emailAddressVisibility; // TODO(server-7): remove
+
   // TODO(server-8): Remove the default values.
   @JsonKey(defaultValue: 15000)
   final int serverTypingStartedExpiryPeriodMilliseconds;
@@ -94,6 +104,7 @@ class InitialSnapshot {
     required this.zulipMergeBase,
     required this.alertWords,
     required this.customProfileFields,
+    required this.emailAddressVisibility,
     required this.serverTypingStartedExpiryPeriodMilliseconds,
     required this.serverTypingStoppedWaitPeriodMilliseconds,
     required this.serverTypingStartedWaitPeriodMilliseconds,
@@ -115,6 +126,14 @@ class InitialSnapshot {
     _$InitialSnapshotFromJson(json);
 
   Map<String, dynamic> toJson() => _$InitialSnapshotToJson(this);
+}
+
+enum EmailAddressVisibility {
+  @JsonValue(1) everyone,
+  @JsonValue(2) members,
+  @JsonValue(3) admins,
+  @JsonValue(4) nobody,
+  @JsonValue(5) moderators,
 }
 
 /// An item in `realm_default_external_accounts`.
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -236,6 +236,7 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
       realmDefaultExternalAccounts: initialSnapshot.realmDefaultExternalAccounts,
       realmEmoji: initialSnapshot.realmEmoji,
       customProfileFields: _sortCustomProfileFields(initialSnapshot.customProfileFields),
+      emailAddressVisibility: initialSnapshot.emailAddressVisibility,
       accountId: accountId,
       selfUserId: account.userId,
       userSettings: initialSnapshot.userSettings,
@@ -269,6 +270,7 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
     required this.realmDefaultExternalAccounts,
     required this.realmEmoji,
     required this.customProfileFields,
+    required this.emailAddressVisibility,
     required this.accountId,
     required this.selfUserId,
     required this.userSettings,
@@ -311,6 +313,8 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
   final Map<String, RealmDefaultExternalAccount> realmDefaultExternalAccounts;
   Map<String, RealmEmojiItem> realmEmoji;
   List<CustomProfileField> customProfileFields;
+  /// For docs, please see [InitialSnapshot.emailAddressVisibility].
+  final EmailAddressVisibility? emailAddressVisibility; // TODO(#668): update this realm setting
 
   ////////////////////////////////
   // Data attached to the self-account on the realm.
--- a/lib/widgets/profile.dart
+++ b/lib/widgets/profile.dart
@@ -3,9 +3,11 @@ import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 
+import '../api/model/initial_snapshot.dart';
 import '../api/model/model.dart';
 import '../model/content.dart';
 import '../model/narrow.dart';
+import '../model/store.dart';
 import 'content.dart';
 import 'message_list.dart';
 import 'page.dart';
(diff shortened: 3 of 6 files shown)


Khader-1 (approved):
Thanks @sm-sayedi! LGTM! 
Now moving on to mentor review with @hackerkid.

chrisbobbe:
Thanks! The issue links to zulip-mobile's implementation: https://github.com/zulip/zulip-mobile/blob/8133bf2b8/src/users/userSelectors.js#L318

Here's that function:

```js
/**
 * The given user's real email address, if known, for displaying in the UI.
 *
 * Null if our user isn't able to see this user's real email address.
 */
export function getDisplayEmailForUser(realm: RealmState, user: UserOrBot): string | null {
  if (user.delivery_email !== undefined) {
    return user.delivery_email;
  } else if (realm.emailAddressVisibility === EmailAddressVisibility.Everyone) {
    // On future servers, we expect this case will never happen: we'll always include
    // a delivery_email when you have access, including when the visibility is Everyone
    // https://github.com/zulip/zulip-mobile/pull/5515#discussion_r997731727
    return user.email;
  } else {
    return null;
  }
}
```

What's the user-facing consequence of not including logic like the check on `realm.emailAddressVisibility`?

sm-sayedi:
Quoting the documentation for `delivery_email` from https://zulip.com/api/register-queue:
> `delivery_email`: string | null
> 
> The user's real email address. This value will be `null` if you cannot access user's real email address. For bot users, this field is always set to the real email of the bot, because bot users always have `email_address_visibility` set to everyone.
> 
> **Changes:** Prior to Zulip 7.0 (feature level 163), this field was present only when `email_address_visibility` was restricted and you had access to the user's real email. As of this feature level, this field is always present, including the case when `email_address_visibility` is set to everyone (and therefore not restricted).

So I think that check was only necessary before feature level 163.

chrisbobbe:
Looks like FL 163 was released in Zulip Server 7: https://zulip.com/api/changelog#changes-in-zulip-70

We aim to support all the way back to Zulip Server 4: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#editing-api-types

but we may bump that to 5 soon: #268

Even when we bump our lower bound to 5, though, we'll still have users running 5 and 6. So the FL 163 threshold is one we should pay attention to.

sm-sayedi:
Thanks for the review @chrisbobbe. Changes pushed. PTAL.

lib/widgets/profile.dart
```diff
@@ -33,6 +35,36 @@ class ProfilePage extends StatelessWidget {
...
 
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if [selfUser] isn't able to see [user]'s real email address.
+  ///
+  /// **Note:** Starting from Zulip version 7.0 (FL 163), there's a change in
+  /// the API about how [selfUser] can access [user]'s real email address.
+  /// Search for "delivery_email" in https://zulip.com/api/register-queue.
+  String? _getDisplayEmailFor(User user, {required PerAccountStore store}) {
+    if (store.account.zulipFeatureLevel >= 163) {
```
chrisbobbe, line 46: Let's include a `TODO(server-7)` for removing the <163 case. That way, when we decide to bump our minimum-supported threshold to 7, this spot will naturally show up in a quick search for backwards-compatibility code that can be removed at that time.

lib/api/model/initial_snapshot.dart
```diff
@@ -24,6 +24,15 @@ class InitialSnapshot {
...
   final List<CustomProfileField> customProfileFields;
 
+  /// The realm-level policy for which a user can see other user's real email address.
+  ///
+  /// Search for "email_address_visibility" in https://zulip.com/api/register-queue.
+  ///
+  /// **Note:** This field is removed in Zulip 7.0 (FL 163) and replaced with:
+  ///   * https://zulip.com/api/update-settings#parameter-email_address_visibility
+  ///   * https://zulip.com/api/update-realm-user-settings-defaults#parameter-email_address_visibility
+  final EmailAddressVisibility? emailAddressVisibility;
```
chrisbobbe, line 34: Let's add a `// TODO(server-7): remove` on this.

lib/api/model/initial_snapshot.dart
```diff
@@ -24,6 +24,15 @@ class InitialSnapshot {
 
   final List<CustomProfileField> customProfileFields;
 
+  /// The realm-level policy for which a user can see other user's real email address.
```
chrisbobbe, line 27: I'm having trouble parsing this summary line. How about:

```suggestion
  /// The realm-level policy, on pre-FL 163 servers, for visibility of real email addresses.
```

lib/api/model/initial_snapshot.dart
```diff
@@ -24,6 +24,15 @@ class InitialSnapshot {
...
+  ///
+  /// Search for "email_address_visibility" in https://zulip.com/api/register-queue.
+  ///
+  /// **Note:** This field is removed in Zulip 7.0 (FL 163) and replaced with:
+  ///   * https://zulip.com/api/update-settings#parameter-email_address_visibility
+  ///   * https://zulip.com/api/update-realm-user-settings-defaults#parameter-email_address_visibility
```
chrisbobbe, line 33:
```suggestion
  /// This field is removed in Zulip 7.0 (FL 163) and replaced with a user-level
  /// setting:
  ///   * https://zulip.com/api/update-settings#parameter-email_address_visibility
  ///   * https://zulip.com/api/update-realm-user-settings-defaults#parameter-email_address_visibility
```

lib/widgets/profile.dart
```diff
@@ -33,6 +35,36 @@ class ProfilePage extends StatelessWidget {
       page: ProfilePage(userId: userId));
   }
 
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if [selfUser] isn't able to see [user]'s real email address.
```
chrisbobbe, line 40: `selfUser` is formatted here as though it's an identifier in the function's code, but it isn't actually. So let's just say "the self-user".

(Here and below.)

chrisbobbe (commented):
Thanks @sm-sayedi!

Small comments below, and also let's isolate the API-types change into the `api: …` commit, and add `PerAccountStore.emailAddressVisibility` in a separate commit.

sm-sayedi:
Thanks @chrisbobbe for the review! Comments addressed. Please have a look.

lib/widgets/profile.dart
```diff
@@ -33,6 +35,37 @@ class ProfilePage extends StatelessWidget {
...
   }
 
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if self-user isn't able to see [user]'s real email address.
+  ///
+  /// **Note:** Starting from Zulip version 7.0 (FL 163), there's a change in
+  /// the API about how self-user can access [user]'s real email address.
+  /// Search for "delivery_email" in https://zulip.com/api/register-queue.
+  // TODO(server-7, FL >= 163): remove
```
chrisbobbe, line 45:
```suggestion
  // TODO(server-7): remove; use user-level setting
```

(The form we've settled on is `TODO(server-x)`; it doesn't need the feature level. All servers that are 7+ will have FL >= 163.)

chrisbobbe (commented):
Thanks! One nit below, and I'll mark this for Greg's review.

lib/model/store.dart
```diff
@@ -311,6 +313,8 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
   final Map<String, RealmDefaultExternalAccount> realmDefaultExternalAccounts;
   Map<String, RealmEmojiItem> realmEmoji;
   List<CustomProfileField> customProfileFields;
+  /// For docs, please see [InitialSnapshot.emailAddressVisibility].
+  final EmailAddressVisibility? emailAddressVisibility; // TODO(server-7): remove
```
gnprice, line 317: This isn't quite right because in this revision it doesn't get updated when the setting changes. This value is an example of a realm setting:
* https://zulip.com/api/get-events#realm-update_dict
* #668 

But I don't want dealing with that to block this feature, or indeed to block launching the app. It's rare this setting will change, and the consequences of having an out-of-date value for a while are small. So let's just mark it with a TODO comment:
```suggestion
  final EmailAddressVisibility? emailAddressVisibility; // TODO(#668) update this realm setting
```

(The TODO-server comment can be omitted because it's effectively covered by the one on InitialSnapshot.emailAddressVisibility — when we go to remove that, we'll necessarily notice we need to remove this at the same time.)

lib/widgets/profile.dart
```diff
@@ -33,6 +35,37 @@ class ProfilePage extends StatelessWidget {
...
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if self-user isn't able to see [user]'s real email address.
+  ///
+  /// **Note:** Starting from Zulip version 7.0 (FL 163), there's a change in
+  /// the API about how self-user can access [user]'s real email address.
+  /// Search for "delivery_email" in https://zulip.com/api/register-queue.
+  // TODO(server-7): remove; use user-level setting
+  String? _getDisplayEmailFor(User user, {required PerAccountStore store}) {
+    if (store.account.zulipFeatureLevel >= 163) {
```
gnprice, line 47:
```suggestion
  String? _getDisplayEmailFor(User user, {required PerAccountStore store}) {
    if (store.account.zulipFeatureLevel >= 163) { // TODO(server-7)
```

Having the comment right at the condition is helpful for seeing that the condition is covered by a TODO-server comment.

The instruction "use user-level setting" isn't right for this spot — we won't be looking at the user's individual `emailAddressVisibility` setting, but rather just looking at their `deliveryEmail`. After all, that's what this code already does for modern servers. If there were something else we should be doing in the future, then for modern servers we should be doing that in the first place.

lib/widgets/profile.dart
```diff
@@ -33,6 +35,37 @@ class ProfilePage extends StatelessWidget {
       page: ProfilePage(userId: userId));
   }
 
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if self-user isn't able to see [user]'s real email address.
+  ///
+  /// **Note:** Starting from Zulip version 7.0 (FL 163), there's a change in
+  /// the API about how self-user can access [user]'s real email address.
+  /// Search for "delivery_email" in https://zulip.com/api/register-queue.
```
gnprice, line 44:
```suggestion
  /// The given user's real email address, if known, for displaying in the UI.
  ///
  /// Returns null if self-user isn't able to see [user]'s real email address.
```

The caller of this method doesn't need to know about that API change, right? It's something this method itself takes care of.

In that case it doesn't belong in a doc comment.

The reader for whom this information is relevant is one trying to understand the details of the method's implementation. That means the place this information could belong is in an implementation comment — `//` instead of `///`, and inside the method's body. I think the code already covers the fact of the API change; but the comment in the modern-server case could benefit from the 'Search for "delivery_email" …' line.

gnprice (commented):
Thanks @sm-sayedi, and thanks @chrisbobbe for the previous reviews!

Generally this looks good. A few small comments below, and then this will be ready to merge.

sm-sayedi:
Thanks @gnprice for the review! Changes pushed!

gnprice:

PR #846, opened 2024-07-28
api: Rename to ChannelDestination and ChannelPostPolicy from "stream"

Khader-1:
Fixes parts of #631.

This focuses on two models:
- `ChannelDestination`
- `ChannelPostPolicy`

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -320,9 +320,9 @@ class ZulipStream {
   final bool isWebPublic; // present since 2.1, according to /api/changelog
   final bool historyPublicToSubscribers;
   final int? messageRetentionDays;
-
-  final ChannelPostPolicy streamPostPolicy;
-  // final bool isAnnouncementOnly; // deprecated for `streamPostPolicy`; ignore
+  @JsonKey(name: 'stream_post_policy')
+  final ChannelPostPolicy channelPostPolicy;
+  // final bool isAnnouncementOnly; // deprecated for `channelPostPolicy`; ignore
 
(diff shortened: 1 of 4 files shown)


sm-sayedi (approved):
Thanks @Khader-1! LGTM! On to the integration review.

lib/api/route/messages.dart
```diff
@@ -183,7 +183,7 @@ Future<SendMessageResult> sendMessage(
   final supportsReadBySender = connection.zulipFeatureLevel! >= 236; // TODO(server-8)
   return connection.post('sendMessage', SendMessageResult.fromJson, 'messages', {
     ...(switch (destination) {
-      StreamDestination() => {
+      ChannelDestination() => {
         'type': RawParameter('stream'),
```
gnprice, line 187: This is a similar story to #847 — StreamDestination describes part of our outgoing API where we say "stream", and so I'd like to package the rename together with starting to say "channel" in what we send. The server does accept "channel" here:
https://zulip.com/api/send-message#parameter-type

gnprice (commented):
Thanks @Khader-1! Merging the ChannelPostPolicy / StreamPostPolicy part of this.

On ChannelDestination, see below.

(The PR branch also contained #847. It's fine to stack one PR atop another one whenever that's helpful, but do be sure to mention it in the PR description because it's important context for reviewers. Since I closed #847 for reasons described there, I'm dropping those commits here too.)

gnprice:
FTR since GitHub makes dropped commits hard to find: that StreamDestination / ChannelDestination commit was:
9a5eea09b api [nfc]: Rename `StreamDestination` to `ChannelDestination`

PR #847, opened 2024-07-28
api: Rename `ApiNarrowStream` to `ApiNarrowChannel`

Khader-1:
Fixes parts of #631

### DIFF
--- a/lib/api/model/narrow.dart
+++ b/lib/api/model/narrow.dart
@@ -34,14 +34,14 @@ sealed class ApiNarrowElement {
   };
 }
 
-class ApiNarrowStream extends ApiNarrowElement {
+class ApiNarrowChannel extends ApiNarrowElement {
   @override String get operator => 'stream';
 
   @override final int operand;
 
-  ApiNarrowStream(this.operand, {super.negated});
+  ApiNarrowChannel(this.operand, {super.negated});
 
-  factory ApiNarrowStream.fromJson(Map<String, dynamic> json) => ApiNarrowStream(
+  factory ApiNarrowChannel.fromJson(Map<String, dynamic> json) => ApiNarrowChannel(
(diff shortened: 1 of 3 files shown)


rajveermalviya (approved):
Thanks @Khader-1!
This looks good to me, moving on to integration review.

lib/api/model/narrow.dart
```diff
@@ -34,14 +34,14 @@ sealed class ApiNarrowElement {
   };
 }
 
-class ApiNarrowStream extends ApiNarrowElement {
+class ApiNarrowChannel extends ApiNarrowElement {
   @override String get operator => 'stream';
```
gnprice, line 38: An instance of an ApiNarrowElement subclass represents a narrow filter:
https://zulip.com/api/construct-narrow
and generally the subclass ApiNarrowFoo represents narrow filters with the operator "foo".

Here, we're going to want to start representing the operator "channel" and sending that to the server instead of "stream". That's a bit tricky because we can't say "channel" to a server that doesn't understand it, i.e. one older than feature level 250. We've successfully handled the same thing for the `pm-with` to `dm` operator; see `ApiNarrowDm` and its subclasses.

I think I don't want to rename this class without doing that handling, though — I feel like that will make it harder for us to notice that that's something we still need to do. So I'll close this PR for now.

(This is different from the situation with parts of the API that are about data we receive. For those, generally the server is still sending the old name and there's nothing for us to do yet toward speaking the new name with the server; the next steps are on the server side. Here, the server accepts both the old and the new name, so it's done what it can and the next steps are entirely on the client side, for us to start sending the new name instead of the old.)

gnprice (commented):
Thanks @Khader-1! See comment below — let's leave this class ApiNarrowStream out of the current round of work on #631.

PR #850, opened 2024-07-29
widgets [nfc]: Strength-reduce flutter/material and flutter/cupertino imports

gnprice:
I noticed this in lib/main.dart, which is short and clearly doesn't
have anything that should require the Material library.  Then just
did a sweep with the command:

  $ perl -i -0pe 's,package:flutter/\Kmaterial,widgets,' \
      -- $(git grep -l package:flutter/material)

followed by `git checkout @` on all the files where the analyzer
complained.

Then similarly for flutter/cupertino.


chrisbobbe:
Thanks, LGTM! Merged.

PR #813, opened 2024-07-15
notif: Add more tests for messaging style notif and fix for fetchBitmap

rajveermalviya:
Add more tests for Android messaging style notif implementation, listed here: https://github.com/zulip/zulip-flutter/pull/718#pullrequestreview-2170569390

PR #305, opened 2023-09-15
Handle Zulip internal links

sirpengi:
Adds support to recognize internal UI links, such as links to other streams or topics.

PR #851, opened 2024-07-30
lint: Enable always_declare_return_types; fix existing places they're missing

gnprice:
When the return type is omitted, the syntax for a function declaration looks a lot like that for a function call.

Some of these also get clearer because the type is nontrivial and a typedef for it is helpful for describing another function's signature too.

Prompted by
https://github.com/zulip/zulip-flutter/pull/813#discussion_r1696064086
and previously
https://github.com/zulip/zulip-flutter/pull/790#discussion_r1692330545
… and previously [several other occasions](https://chat.zulip.org/#narrow/stream/494-mobile-github/search/explicit.20return.20type), going back to:
https://github.com/zulip/zulip-flutter/pull/305#discussion_r1355788136

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -21,6 +21,7 @@ linter:
   # For a list of all available lints, with docs, see:
   #   https://dart-lang.github.io/linter/lints/index.html.
   rules:
+    always_declare_return_types: true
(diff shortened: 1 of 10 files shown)


chrisbobbe:
LGTM, thanks! Merged.

PR #852, opened 2024-07-30
Add a loading indicator for PerAccountStore

### LINKED ISSUE #465: Show "Connecting…" banner (or equivalent) when server data is stale

PIG208:
<details>

<summary>Preview</summary>

| isLoading=false | isLoading=true |
| - | - |
| image | image |
| image | image |

</details>

Fixes #465.

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -297,6 +297,15 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
   final GlobalStore _globalStore;
   final ApiConnection connection; // TODO(#135): update zulipFeatureLevel with events
 
+  bool get isLoading => _isLoading;
+  bool _isLoading = false;
+  @visibleForTesting
+  set isLoading(bool value) {
+    if (_isLoading == value) return;
+    _isLoading = value;
+    notifyListeners();
+  }
+
   ////////////////////////////////
   // Data attached to the realm or the server.
 
@@ -770,6 +779,7 @@ class UpdateMachine {
         result = await getEvents(store.connection,
           queueId: queueId, lastEventId: lastEventId);
       } catch (e) {
+        store.isLoading = true;
         switch (e) {
           case ZulipApiException(code: 'BAD_EVENT_QUEUE_ID'):
             assert(debugLog('Lost event queue for $store.  Replacing…'));
@@ -797,6 +807,7 @@ class UpdateMachine {
         }
       }
 
+      store.isLoading = false;
       final events = result.events;
       for (final event in events) {
         await store.handleEvent(event);
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -8,6 +8,7 @@ import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import '../model/localizations.dart';
 import '../model/narrow.dart';
 import 'about_zulip.dart';
+import 'app_bar.dart';
 import 'inbox.dart';
 import 'login.dart';
 import 'message_list.dart';
@@ -252,7 +253,9 @@ class HomePage extends StatelessWidget {
     }
 
     return Scaffold(
-      appBar: AppBar(title: const Text("Home")),
+      appBar: ZulipAppBar(
+        title: const Text("Home"),
+        isLoading: store.isLoading),
       body: Center(
         child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
           DefaultTextStyle.merge(
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -253,9 +253,7 @@ class HomePage extends StatelessWidget {
     }
 
     return Scaffold(
-      appBar: ZulipAppBar(
-        title: const Text("Home"),
-        isLoading: store.isLoading),
+      appBar: ZulipAppBar(title: const Text("Home")),
       body: Center(
         child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
           DefaultTextStyle.merge(
--- a/lib/widgets/inbox.dart
+++ b/lib/widgets/inbox.dart
@@ -4,6 +4,7 @@ import '../api/model/model.dart';
 import '../model/narrow.dart';
 import '../model/recent_dm_conversations.dart';
 import '../model/unreads.dart';
+import 'app_bar.dart';
 import 'icons.dart';
 import 'message_list.dart';
 import 'page.dart';
@@ -160,7 +161,9 @@ class _InboxPageState extends State<InboxPage> with PerAccountStoreAwareStateMix
     }
 
     return Scaffold(
-      appBar: AppBar(title: const Text('Inbox')),
+      appBar: ZulipAppBar(
+        title: const Text('Inbox'),
+        isLoading: store.isLoading),
       body: SafeArea(
         // Don't pad the bottom here; we want the list content to do that.
         bottom: false,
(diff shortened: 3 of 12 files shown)


lib/model/store.dart
```diff
@@ -295,6 +295,15 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
   final GlobalStore _globalStore;
   final ApiConnection connection; // TODO(#135): update zulipFeatureLevel with events
 
+  bool get isLoading => _isLoading;
+  bool _isLoading = false;
+  @visibleForTesting
```
PIG208, line 300: Useful for a widget test added later.

PIG208:
Oops, just pushed again to fix that.

lib/widgets/message_list.dart
```diff
@@ -255,12 +256,13 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
...
-      appBar: AppBar(title: MessageListAppBarTitle(narrow: widget.narrow),
+      appBar: ZulipAppBar(
+        title: MessageListAppBarTitle(narrow: widget.narrow),
+        isLoading: store.isLoading,
         backgroundColor: appBarBackgroundColor,
         shape: removeAppBarBottomBorder
           ? const Border()
-          : null, // i.e., inherit
-      ),
+          : null),
```
chrisbobbe, line 265: Let's keep this comment. A reader might see `null` and think it means there's no bottom border. That would be wrong; what `null` actually does is make the app bar inherit its bottom border from the `AppBarTheme`.

chrisbobbe (commented):
Thanks!

I'm curious if you explored an approach with a toast: https://github.com/zulip/zulip-flutter/issues/465#issuecomment-1984871814

Would it be easy to put some code in one central place, ensuring that a toast showing the org's name is visible for each account in the loading state? (E.g., for CZO: "Connecting to **Zulip Community**…")

Or, a refinement (that could be annoying to implement): filter that list of toasts down to the account corresponding to the foregrounded screen, and no need to show the org's name.

In the screenshots, I'm noticing the color isn't a great match for the app bar, and the split between the app bar and the sticky recipient header is looking kind of awkward.

If we're going with this linear-indicator approach, let's definitely track that it's missing on the lightbox screen, which needs it because it's a per-account screen too.

PIG208:
Thanks for the review! I didn't experiment with `Toast`. It sounds like another viable option.

Could you elaborate on this?

> Or, a refinement (that could be annoying to implement): filter that list of toasts down to the account corresponding to the foregrounded screen, and no need to show the org's name.

Visually, matching the background color should be fairly simple, because `LinearProgressIndicator` is shipped with the app bar.
```diff
diff --git lib/widgets/app_bar.dart lib/widgets/app_bar.dart
index da49a639..a1c4645b 100644
--- lib/widgets/app_bar.dart
+++ lib/widgets/app_bar.dart
@@ -12,6 +12,6 @@ class ZulipAppBar extends AppBar {
     bottom: PreferredSize(
       preferredSize: const Size.fromHeight(4.0),
       child: (isLoading)
-        ? const LinearProgressIndicator(minHeight: 4.0)
+        ? LinearProgressIndicator(backgroundColor: backgroundColor, minHeight: 4.0)
         : const SizedBox.shrink()));
 }
```

PIG208:
Do we have a toast library in mind to use? Or are we writing one from scratch?

gnprice:
Chat thread about what type of indicator to use:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/banner.2Fsnack.20bar.2Fprogress.20indicator.20for.20connection.20error/near/1909308

gnprice:
> If we're going with this linear-indicator approach, let's definitely track that it's missing on the lightbox screen, which needs it because it's a per-account screen too.

I think it's actually fine to not have the indicator on the lightbox screen. Critically, there's very little there to get live-updated — even if you're looking at stale data there and you think it's up to date, it's hard to get misled about much as a result. In particular it never gets updated by a new message arriving.

It should definitely get a comment saying we've chosen not to put the indicator there, though, and why.

chrisbobbe:
Sure, I guess that all makes sense.

chrisbobbe:
So the current revision is close to what we want, right? Is this the only thing missing:

> It should definitely get a comment saying we've chosen not to put the indicator there, though, and why.

PIG208:
Right, will update this later.

PIG208:
Updated the PR to mention the ZulipAppBar exceptions in comments after its dart doc.

PIG208:
Updated the PR to handle `isLoading` when there are transient connection errors.

chrisbobbe:
Thanks! LGTM, over to Greg for his review.

PIG208:
(Rebased the PR)

lib/widgets/app.dart
```diff
@@ -179,9 +180,10 @@ class ChooseAccountPage extends StatelessWidget {
     assert(!PerAccountStoreWidget.debugExistsOf(context));
     final globalStore = GlobalStoreWidget.of(context);
     return Scaffold(
-      appBar: AppBar(
+      appBar: ZulipAppBar(
         title: Text(zulipLocalizations.chooseAccountPageTitle),
-        actions: const [ChooseAccountPageOverflowButton()]),
+        actions: const [ChooseAccountPageOverflowButton()],
+        isLoading: false),
```
gnprice: How about just `AppBar`, if we're going to hard-code `isLoading` here to false?

lib/widgets/inbox.dart
```diff
@@ -160,7 +161,9 @@ class _InboxPageState extends State<InboxPage> with PerAccountStoreAwareStateMix
     }
 
     return Scaffold(
-      appBar: AppBar(title: const Text('Inbox')),
+      appBar: ZulipAppBar(
+        title: const Text('Inbox'),
+        isLoading: store.isLoading),
```
gnprice: I think this pattern can be simplified. I'll push a separate commit on top to do so.

gnprice (commented):

Generally this looks good. I also put a build with this PR (among others) onto my phone last week before going on vacation, and the behavior was very helpful.

A couple of small comments below. I've also pushed some additional commits on top:
c384847c1 wip squash? plain AppBar where isLoading always false
667adf856 app_bar [nfc]: Move ZulipAppBar.isLoading to within widget
a6965aad5 app_bar [nfc]: Explain why the lightbox skips ZulipAppBar

The first one reflects one comment, and can be squashed; I added it just because that change is needed for the next commit.

PIG208:
Thanks! Rebased the PR and squashed the commit.

gnprice:
Looks good! Thanks for the revision — merging.

Issue #90, opened 2023-05-12
Bottom sheet redesign


gnprice:
When you long-press a message, we show a bottom sheet with actions you can take on it, which we also call an "action sheet". (Currently in this prototype there's just one action, "share". In zulip-mobile it's a much longer list, and we also show a similar widget when you long-press on a topic name and some other places.)

Currently it's a pretty plain bottom sheet. We should adjust the styling to match the following design prepared by @terpimost:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-27050&m=dev
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-27168&m=dev

Here's a screenshot taken from there. For details of the layout and behavior, see the Figma design.
![image](https://github.com/user-attachments/assets/e838e050-d103-46cc-958d-df5e6239ec1a)

To help manage the work, I've split implementing this design up into two GitHub issues. For this issue, we'll only implement the screen shown above, not the three others you'll find to the right of it in Figma showing variations. Then there's a followup issue here:
* #217

gnprice:
I've just split out part of this issue to a new one:
* #217

So for this issue, let's leave out the message text at the top of the bottom sheet; we'll pursue that in the new separate issue. (At the moment our bottom sheet has many fewer items than this, which I think also reduces somewhat the need for that feature because there's room on the screen to still see the message while the bottom sheet is open.)

Let's also leave out of scope for this issue any changes to what icons we use for the options. Replacing the icons will be done as part of #200 or a followup, and can happen either before or after we complete this issue.

Puneet2504:
@gnprice instead of openeing the bottom sheet we can have something that draws up when we long press on the message field having message field placed where it is.

gnprice:
@Puneet2504 This issue thread is about implementing the design that our designer @terpimost has already developed. We could discuss potential other designs, but the place for that would be in a new thread in #mobile on chat.zulip.org. (You'll also need to be more specific about what the proposed alternate design is — I don't quite understand what you're describing — and why you think it'd be better.)

SharmaDhruv2511:
is this issue still open? @gnprice I would like to contribute to it

Sahilpervez:
@gnprice I would like to work on this issue.

Sahilpervez:
@gnprice Here is the Idea I would like to propose for this issue.

[...]

Issue #217, opened 2023-07-12
Show message text in bottom sheet


gnprice:
In the redesigned message bottom sheet #90 from @terpimost, the top of the message sheet shows the message you're acting on.

We've left that part out of the scope of #90; this is a follow-up issue for adding it. See Figma design:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-29796&m=dev
![image](https://github.com/user-attachments/assets/1909713a-337f-4be5-8fa4-6bb04c3f52e2)

and quoting a note from there:
> The message is exactly the same as in the feed. It shouldn’t be just a picture, because we want to allow TEXT SELECTION HERE.

See also the notes to the right of there in the Figma design, about scrolling behavior.

#### Implementation

I expect this will involve reusing most of the implementation of MessageWithPossibleSender. That will probably require some refactoring, in order to make the relevant parts reusable without other aspects that aren't desired here.

Also if the message in its original context in the message list didn't show the sender, we'll want to still show the sender here.

<details><summary>Old ideas</summary>

In the redesigned message bottom sheet #90 from @terpimost, the top of the message sheet shows the text of the message you're acting on (cutting it off at two lines). This is helpful when the message itself isn't visible, which becomes more and more likely as we add more actions to the bottom sheet.

I'm splitting this out from the rest of #90, because the concept of "the text of the message" is actually a bit slippery and raises some product/design questions.

* We could use the message's Markdown source. But that isn't great for this UI context: it may have a bunch of markup that obscures what the message actually says.
  * Consider in particular the case where the message starts with a quote-reply.
  * Consider also a message that starts with one short paragraph like "Yes.", and then has more details in a second paragraph. The first two lines of the Markdown source will be just "Yes." followed by a blank line. That's not very informative, and we might want to show more text than that.

  If we decide we're OK with that, there's a second pitfall which we'll need to pick a response to: we don't normally keep around the Markdown source for all the messages we have, only the rendered HTML. On the occasions where we need Markdown source (like for generating a quote-and-reply), we make a separate request to the server to get it. We can do that here… but then we need to decide what to show when that request hasn't completed yet, which will be at least a frame or two and often more.
  * We could delay the whole bottom sheet. That seems bad.
  * We could show nothing; when the text arrives, the top edge of the bottom sheet jumps up to make room.
[...]

Issue #200, opened 2023-06-17
Build an asset font for custom icons


gnprice:
Much like we do in zulip-mobile and Zulip web. We'll want a font file containing the icons, and then a generated Dart file that maps icon names to glyph IDs in the font.

For references, see:
* `tools/build-icon-font` in zulip-mobile, which is a Node script doing a similar job using the NPM package `@vusion/webfonts-generator`
* [the Dart script](https://github.com/lucide-icons/lucide/blob/34155d48e7927afe80512a4ba4994085e799e52c/packages/lucide-flutter/tool/generate_fonts.dart) that Lucide uses to generate the mapping file when building their Flutter package (but that script consumes a file that something else generated, which provides the data just in a different format)

It would be nice if we can write the script purely in Dart. If we can't readily lay hands on appropriate tools to make that easy, we could settle for having a Node script so that we can use NPM packages.

The generated output — both the font file and the generated Dart file — should be committed to the repo, just like we do for other generated files and just like we do for the corresponding files in zulip-mobile. That way the script only has to be run when changing the set of icons, and there's no complication added to a normal development cycle.

gnprice:
The custom icons we'll want are those [in `@zulip/shared`](https://github.com/zulip/zulip/tree/92c83c1df489975f8fe887f190e5d0af7c9534a2/web/shared/icons), used by Zulip web and zulip-mobile.

Potentially we might also want to include other icons not found there.

One known place where we'll want to use a custom icon is in #119, in the spot that replaces an avatar for group DM threads; see [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/design.3A.20DM-conversation.20list/near/1594671). If we write this script after #119 is already done, then we'll have some other icon there initially and we'll want to replace it with the custom icon.

PR #12, opened 2023-02-23
msglist: Add message action sheet with "Share" button

chrisbobbe:

PR #853, opened 2024-07-31
action_sheet: Redesign bottom sheet

sm-sayedi:
Matches the design of the bottom sheet with the Figma design https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-27050&m=dev.

Made the bottom sheet take as much space as needed and when it takes more than a screenful, the option button list is scrollable.

<table>
    <tr>
        <td><b>Before</b></td>
        <td><b>After</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
    <tr>
        <td><b>After (more than a screenfull)</b></td>
        <td><b>After - Dark</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
</table>

#### Screen recording (screenful with top and bottom shadows)

https://github.com/user-attachments/assets/71880a57-3270-4278-925f-8c79c2f8d3f2

Fixes: #90


lib/widgets/action_sheet.dart
```diff
@@ -36,20 +36,46 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       && reactionWithVotes.userIds.contains(store.selfUserId))
     ?? false;
 
-  showDraggableScrollableModalBottomSheet<void>(
+  final designVariables = DesignVariables.of(context);
+  showModalBottomSheet<void>(
```
sm-sayedi, line 40: For the current behavior of the action sheet (mentioned in [#90 comment](https://github.com/zulip/zulip-flutter/issues/90#issuecomment-2016666599)), there is no need to use `showDraggableScrollableModalBottomSheet`.

As a side note, `_DragHandleLayer` is no longer necessary in `showDraggableScrollableModalBottomSheet` as there is `showDragHandle` property available in `showModalBottomSheet`. Also, `showDraggableScrollableModalBottomSheet` is no longer used in the codebase, so I am not sure if it is meant to stay there for future usage.

lib/widgets/action_sheet.dart
```diff
@@ -69,11 +95,16 @@ abstract class MessageActionSheetMenuItemButton extends StatelessWidget {
...
   Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
     return MenuItemButton(
-      leadingIcon: Icon(icon),
+      trailingIcon: Icon(icon, color: designVariables.actionSheetMenuButtonForeground),
+        style: MenuItemButton.styleFrom(
+          backgroundColor: designVariables.actionSheetMenuButtonBackground,
+          foregroundColor: designVariables.actionSheetMenuButtonForeground,
+        ),
```
rajveermalviya, line 105: nit:

```suggestion
      style: MenuItemButton.styleFrom(
        backgroundColor: designVariables.actionSheetMenuButtonBackground,
        foregroundColor: designVariables.actionSheetMenuButtonForeground,
      ),
```

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +433,24 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+      style: ElevatedButton.styleFrom(
+        elevation: 0,
+        backgroundColor: designVariables.actionSheetCancelButtonBackground,
+        foregroundColor: designVariables.actionSheetCancelButtonForeground,
+        shadowColor: Colors.transparent,
+        shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
+      ),
+      onPressed: () => Navigator.pop(context),
+      child: Text(ZulipLocalizations.of(context).dialogCancel,
+        style: const TextStyle(fontSize: 16)),
```
rajveermalviya, line 453: Same for this, text style needs to match Figma design.

lib/widgets/action_sheet.dart
```diff
@@ -402,3 +433,24 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    return ElevatedButton(
+      style: ElevatedButton.styleFrom(
+        elevation: 0,
+        backgroundColor: designVariables.actionSheetCancelButtonBackground,
+        foregroundColor: designVariables.actionSheetCancelButtonForeground,
+        shadowColor: Colors.transparent,
+        shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
+      ),
```
rajveermalviya, line 450: Can use `TextButton` here:

```suggestion
    return TextButton(
      style: TextButton.styleFrom(
        backgroundColor: designVariables.actionSheetCancelButtonBackground,
        foregroundColor: designVariables.actionSheetCancelButtonForeground,
        shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
      ),
```

rajveermalviya: Also add explicit padding of `EdgeInsets.all(10)` as per Figma design:

Screenshot 2024-08-06 at 21 50 27

lib/widgets/action_sheet.dart
```diff
@@ -36,20 +36,46 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+                    CopyMessageTextButton(message: message, messageListContext: context),
+                    CopyMessageLinkButton(message: message, messageListContext: context),
+                    ShareButton(message: message, messageListContext: context),
+                  // TODO: The following line could be replaced by the [spacing]
+                  // property when https://github.com/flutter/flutter/issues/55378 is fixed.
+                  ].expand((item) => [const SizedBox(height: 1), item]).skip(1).toList()),
+                ),
+              ),
+            ),
+            const MessageActionSheetCancelButton(),
```
rajveermalviya, line 75: There needs to be 8px padding between Menu buttons and Cancel button as per design:

Screenshot 2024-08-06 at 21 52 34

lib/widgets/action_sheet.dart
```diff
@@ -36,20 +36,46 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+                  child: Column(children: [
+                    if (!hasThumbsUpReactionVote) AddThumbsUpButton(message: message, messageListContext: context),
+                    StarButton(message: message, messageListContext: context),
+                    if (isComposeBoxOffered) QuoteAndReplyButton(
+                      message: message,
+                      messageListContext: context,
+                    ),
+                    CopyMessageTextButton(message: message, messageListContext: context),
+                    CopyMessageLinkButton(message: message, messageListContext: context),
+                    ShareButton(message: message, messageListContext: context),
```
rajveermalviya, line 68: The order doesn't match the Figma design, it should be:

```dart
AddThumbsUpButton(),
CopyMessageTextButton(),
CopyMessageLinkButton(),
ShareButton(),
StarButton(),
QuoteAndReplyButton(),
```

lib/widgets/action_sheet.dart
```diff
@@ -36,20 +36,46 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+                child: ClipRRect(
+                  borderRadius: BorderRadius.circular(7),
+                  child: Column(children: [
+                    if (!hasThumbsUpReactionVote) AddThumbsUpButton(message: message, messageListContext: context),
+                    StarButton(message: message, messageListContext: context),
+                    if (isComposeBoxOffered) QuoteAndReplyButton(
+                      message: message,
+                      messageListContext: context,
+                    ),
+                    CopyMessageTextButton(message: message, messageListContext: context),
```
rajveermalviya, line 66: `CopyMessageTextButton`'s text needs to be changed to "Copy to clipboard".

lib/widgets/action_sheet.dart
rajveermalviya, line 68: `ShareButton`'s text should be changed to "Share link"

rajveermalviya (changes requested):
Thanks for working on this @sm-sayedi.

Added comments for some things I've noticed to be different from the [linked](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-27050&m=dev) Figma design.

In addition to those, the Icons also need to be updated to match Figma design. https://github.com/zulip/zulip-flutter/issues/90#issuecomment-1633352421 mentions to do it after #200 is done, and now it's done, so I think they should be updated and any new ones should be added in `ZulipIcons`.

lib/widgets/action_sheet.dart
sm-sayedi: Yeah, it is different. But [Alya decided](https://chat.zulip.org/#narrow/stream/48-mobile/topic/bottom.20sheet.20redesign.20-.20.23F853/near/1908715) to keep it as it is. 🙂

lib/widgets/action_sheet.dart
```diff
@@ -69,11 +98,21 @@ abstract class MessageActionSheetMenuItemButton extends StatelessWidget {
...
+        padding: const EdgeInsets.symmetric(vertical: 12, horizontal: 16),
+        tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+        minimumSize: Size.zero,
+        backgroundColor: designVariables.actionSheetMenuButtonBackground,
+        foregroundColor: designVariables.actionSheetMenuButtonForeground,
+      ),
       onPressed: () => onPressed(context),
-      child: Text(label(zulipLocalizations)));
+      child: Text(label(zulipLocalizations),
+        style: const TextStyle(fontSize: 20, fontWeight: FontWeight.w600),
```
rajveermalviya, line 114: I think fontWeight is not being applied correctly, you need to use `weightVariableTextStyle`:
```suggestion
        style: const TextStyle(fontSize: 20)
          .merge(weightVariableTextStyle(context, wght: 600)),
```

rajveermalviya (approved):
Thanks for the revision @sm-sayedi, this looks good to me, just one comment.

Moving on to the mentor review from @hackerkid.

lib/widgets/action_sheet.dart
rajveermalviya: Similarly for the Cancel button.

lib/widgets/action_sheet.dart
```diff
@@ -36,20 +38,47 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+                    if (isComposeBoxOffered) QuoteAndReplyButton(
+                      message: message,
+                      messageListContext: context,
+                    ),
+                    CopyMessageTextButton(message: message, messageListContext: context),
+                    CopyMessageLinkButton(message: message, messageListContext: context),
+                    ShareButton(message: message, messageListContext: context),
+                  // TODO: The following line could be replaced by the [spacing]
+                  // property when https://github.com/flutter/flutter/issues/55378 is fixed.
+                  ].expand((item) => [const SizedBox(height: 1), item]).skip(1).toList()),
```
sm-sayedi, line 73: Now that the mentioned upstream issue is fixed, we can simply add `spacing: 1` to `Column`. I will do that after #872 is merged.

lib/widgets/action_sheet.dart
gnprice: Yeah, a follow-up commit in this PR can remove that function — that'd be a nice cleanup, thanks. I expect we'll want all our bottom sheets to work the same way as this one.

And if we do want to go back and use that implementation for something, it's always there in the Git history.


> As a side note, `_DragHandleLayer` is no longer necessary in `showDraggableScrollableModalBottomSheet` as there is `showDragHandle` property available in `showModalBottomSheet`.

True. It looks like that was added in https://github.com/flutter/flutter/pull/122445, back in 2023-03… and both @chrisbobbe and I actually reviewed it (mainly on the predecessor thread https://github.com/flutter/flutter/pull/120855). That landed after we merged Chris's implementation here, in #12.

I don't recall if that upstream implementation had behavior differences that would have caused us to stick with ours. I suspect we just didn't bother to switch to it, having already moved on to other tasks by the time it landed upstream.

chrisbobbe: > I don't recall if that upstream implementation had behavior differences that would have caused us to stick with ours. I suspect we just didn't bother to switch to it, having already moved on to other tasks by the time it landed upstream.

Yes, I investigated some months ago and there was a behavior difference that caused us to stick with ours. Here's the context around where we use `_DragHandleLayer`:

```dart
    builder: (BuildContext context) {
      // Make the content start below the drag handle in the y-direction, but
      // when the content is scrollable, let it scroll under the drag handle in
      // the z-direction.
      return Stack(
        children: [
          _DraggableScrollableLayer(builder: builder),
          _DragHandleLayer(),
        ]);
    });
```

If we migrated to `showDragHandle: true`, then we'd have to give up the part after "but" in that comment. You can simulate this by making there be more than a screenful of options in the sheet:

| Current | Migrated to `showDragHandle: true` |
| --- | --- |
| image | image |

The culprit is a `Padding` in upstream that always forces the caller's content to begin below the drag-handle area in the y-direction:

<details>
<summary>code</summary>

https://github.com/flutter/flutter/blob/ba31bf8c0/packages/flutter/lib/src/material/bottom_sheet.dart#L402
```dart
    Widget bottomSheet = Material(
      key: _childKey,
      color: color,
      elevation: elevation,
      surfaceTintColor: surfaceTintColor,
      shadowColor: shadowColor,
      shape: shape,
      clipBehavior: clipBehavior,
      child: NotificationListener<DraggableScrollableNotification>(
        onNotification: extentChanged,
        child: !showDragHandle
          ? widget.builder(context)
          : Stack(
              alignment: Alignment.topCenter,
              children: <Widget>[
                dragHandle!,
                Padding(
                  padding: const EdgeInsets.only(top: kMinInteractiveDimension),
                  child: widget.builder(context),
                ),
              ],
            ),
      ),
    );
```
</details>

I did [mention](https://github.com/flutter/flutter/pull/120855#discussion_r1117789100) our desired behavior in my review of that upstream work that Greg mentioned, but didn't follow up when it turned out not to be supported.

lib/widgets/action_sheet.dart
```diff
@@ -41,25 +44,50 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+                      narrow: narrow,
+                    ),
+                    CopyMessageTextButton(message: message, messageListContext: context),
+                    CopyMessageLinkButton(message: message, messageListContext: context),
+                    ShareButton(message: message, messageListContext: context),
+                  ]),
+                ),
+              ),
+            ),
+            const SizedBox(height: 8),
```
chrisbobbe, line 86: When the buttons area can scroll (when there are more buttons than fit on the user's screen), the content, as it scrolls, should be allowed to show through this 8px gap, but masked with a gradient from transparent to `DesignVariables.bgContextMenu`: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-42553&m=dev

(Same with an area of height 8px above the buttons.)

If this turns out to be difficult, we can file an issue with a TODO pointing to it, but let's not lose track of it. It seems independent of #217. #217 is one factor that will make this scroll effect more noticeable (by taking up more vertical space), but it's not the only factor that can; other factors are the number of buttons, the user's screen size, and the user's system text-size setting.

lib/widgets/action_sheet.dart
```diff
@@ -79,11 +107,22 @@ abstract class MessageActionSheetMenuItemButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
     return MenuItemButton(
-      leadingIcon: Icon(icon),
+      trailingIcon: Icon(icon, color: designVariables.actionSheetMenuButtonForeground),
+      style: MenuItemButton.styleFrom(
+        padding: const EdgeInsets.symmetric(vertical: 12, horizontal: 16),
```
chrisbobbe, line 115: I remember Vlad saying that he doesn't want the material "splash" animation on button presses, so let's specify that:

```dart
      style: MenuItemButton.styleFrom(
```

Then to differentiate the background color between pressed and not-pressed, it looks like we're supposed to apply 12% opacity to the `contextMenuItemBg` variable for not-pressed, and 20% opacity for pressed:
- https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3302-20443&t=3eSOW3rCFnEwnweA-0
- https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3302-20443&t=3eSOW3rCFnEwnweA-0

(Those links are for light mode, but it turns out the same values 12% and 20% are used in dark mode.)

`MenuItemButton.styleFrom` doesn't seem to support specifying different background colors based on pressed and not-pressed, but it returns a `ButtonStyle`, and `ButtonStyle` does support that. So, instead of passing `backgroundColor` to `.styleFrom`, could just add after the `styleFrom(…)`:

```dart
      )
        .copyWith(
          backgroundColor: WidgetStateColor.resolveWith((states) =>
            designVariables.contextMenuItemBg.withOpacity(
              states.contains(WidgetState.pressed) ? 0.2 : 0.12))),
```

chrisbobbe (changes requested):
Thanks, this will make the message action sheet much more polished! Excited to have this. Comments below.

lib/widgets/action_sheet.dart
chrisbobbe: (This does turn out to be moot for us, though, since the redesigned sheet doesn't have a drag handle.)

test/widgets/action_sheet_test.dart
```diff
@@ -610,4 +610,39 @@ void main() {
...
+      check(find.text(zulipLocalizations.actionSheetOptionQuoteAndReply)
+        .evaluate().length).equals(isShown ? 1 : 0);
+      check(find.text(zulipLocalizations.actionSheetOptionMarkAsUnread)
+        .evaluate().length).equals(isShown ? 1 : 0);
+      check(find.text(zulipLocalizations.actionSheetOptionCopyMessageText)
+        .evaluate().length).equals(isShown ? 1 : 0);
+      check(find.text(zulipLocalizations.actionSheetOptionCopyMessageLink)
+        .evaluate().length).equals(isShown ? 1 : 0);
+      check(find.text(zulipLocalizations.actionSheetOptionShare)
+        .evaluate().length).equals(isShown ? 1 : 0);
```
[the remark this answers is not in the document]
sm-sayedi: What about checking for the number of buttons? If more buttons are added/removed, this test will fail and will make us update the number. I think this will have a good representation of actions sheet being shown!

lib/widgets/action_sheet.dart
```diff
@@ -79,11 +107,22 @@ abstract class MessageActionSheetMenuItemButton extends StatelessWidget {
...
+        tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+        minimumSize: Size.zero,
```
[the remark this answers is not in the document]
sm-sayedi: For these specific settings, leaving them out doesn't make any difference, but if we wanted to give less than 12px vertical padding to the buttons, then we need these two to see that effect! Should I remove them for now?

sm-sayedi:
Thanks @chrisbobbe for the review! Changes pushed. PTAL.

test/widgets/action_sheet_test.dart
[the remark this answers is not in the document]
gnprice: > If more buttons are added/removed, this test will fail and will make us update the number.

I think that's the sort of update that is likely to feel like just busywork when we have to do it. See the paragraphs around the mention of "busywork" here:

If there were a test that had the purpose of testing the list of options, that could be a useful test. But I think that purpose is basically covered by the individual options' tests. And the list varies based on a number of conditions, so really there can't usefully be just one test for the whole list.



> I think this will have a good representation of actions sheet being shown!

I think the best representation of the list of options in the action sheet is the one in the source code, in `showMessageActionSheet`. With a bit of cleanup to have a more uniform format, it looks like this:
```dart
      return Column(children: [
        if (!hasThumbsUpReactionVote)
          AddThumbsUpButton(message: message, messageListContext: context),
        StarButton(message: message, messageListContext: context),
        if (isComposeBoxOffered)
          QuoteAndReplyButton(message: message, messageListContext: context),
        if (showMarkAsUnreadButton)
          MarkAsUnreadButton(message: message, messageListContext: context, narrow: narrow),
        CopyMessageTextButton(message: message, messageListContext: context),
        CopyMessageLinkButton(message: message, messageListContext: context),
        ShareButton(message: message, messageListContext: context),
      ]);
```

which is hard to beat; really the only way to be more concise would be to reduce those argument lists passed to the constructors.

In general it's very often the case that the best representation of what's in the UI, or of how something should behave, is in the source code. Tests are valuable for other purposes (like preventing regressions, and thereby enabling making changes confidently) but IME for this purpose they're rarely better than the code under test.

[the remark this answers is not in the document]
gnprice: > With a bit of cleanup to have a more uniform format

Sent #902 for that.

Relatedly, in this PR it would be helpful to pull out the list of columns as a local variable in that build method. That will keep it from getting real deeply indented (which is helpful for not making the formatting cramped), and will also help make the list easy to spot by separating it visually from all the layout details.

sm-sayedi:
Thanks for the comments @gnprice. Pushed the new revision. This time only checked for the "Star message" and "Cancel" buttons to test if the action sheet is shown.

chrisbobbe (commented):

lib/widgets/theme.dart
```diff
@@ -219,6 +244,11 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
 
   // Not named variables in Figma; taken from older Figma drafts, or elsewhere.
   final Color atMentionMarker;
+  final Color bgContextMenu;
+  final Color contextMenuCancelBg;
+  final Color contextMenuCancelText;
+  final Color contextMenuItemBg;
+  final Color contextMenuItemText;
```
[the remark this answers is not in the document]
sm-sayedi: Thanks. I forgot that these became named variables from Figma. Also, still kept `contextMenuCancelBg` in the file because of a [previous discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/bottom.20sheet.20redesign.20-.20.23F853/near/1909523), but left the necessary comments.

lib/widgets/action_sheet.dart
```diff
@@ -435,3 +616,30 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+      style: TextButton.styleFrom(
+        padding: const EdgeInsets.all(10),
+        tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+        minimumSize: Size.zero,
+        foregroundColor: designVariables.contextMenuCancelText,
+        shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
+        splashFactory: NoSplash.splashFactory,
+      ).copyWith(backgroundColor: WidgetStateColor.resolveWith((states) =>
+          designVariables.contextMenuCancelBg.withOpacity(
+            states.contains(WidgetState.pressed) ? 0.20 : 0.15))),
```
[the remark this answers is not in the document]
sm-sayedi: It's actually taken from Figma 🙂: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=421-14113&t=gfnSxRpfQnio7zUP-0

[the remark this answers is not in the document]
chrisbobbe: I see, thanks for pointing me to that! I was on a different page ("Main menu" instead of "Context menu"), but that page is also marked as ready for dev, and I agree it's a close enough match that it makes sense to use it here.

lib/widgets/action_sheet.dart
```diff
@@ -12,51 +15,210 @@ import 'actions.dart';
...
+      future: Future.microtask(() => true),
+      builder: (context, snapshot) {
+        if (!snapshot.hasData) {
+          return const SizedBox.shrink();
+        }
+        return ListenableBuilder(
+          listenable: scrollController,
+          builder: (context, __) => builder(context, scrollController),
+        );
+      });
```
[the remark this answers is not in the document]
sm-sayedi: The top and bottom shadows appear only when there is content above or under the available height of the `ScrollView` so we listen to `ScrollController` and act upon its `position.extentBefore` and `position.extentAfter`. `FutureBuilder` is there to wait for `ScrollController` to be attached to the `SingleChildScrollView` so we can use those properties. `ListenableBuilder` is necessary for efficiently listening to `ScrollController` and redrawing a portion of the UI instead of all of the UI in the `build` method. So the junk will only happen to those small portions of the UI.

> In a search just now, I found something called [`ShaderMask`](https://api.flutter.dev/flutter/widgets/ShaderMask-class.html); could that be helpful?

I don't think it's going to be helpful in this specific case as we have to wrap the whole `SingleChildScrollView` with `ShaderMask` and now that we want to show those shadows based on the conditions mentioned above, the whole `SingleChildScrollView` has to be redrawn each time, which is inefficient.

sm-sayedi:
Thanks @chrisbobbe for the review! Changes pushed.

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
sm-sayedi: Yeah, that (cancel) button is actually the "main component" of the button used in the "Context menu".
temp figma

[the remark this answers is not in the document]
chrisbobbe: 💯

lib/widgets/action_sheet.dart
```diff
@@ -435,3 +476,26 @@ class ShareButton extends MessageActionSheetMenuItemButton {
...
+  const MessageActionSheetCancelButton({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    return TextButton(
+      style: TextButton.styleFrom(
+        padding: const EdgeInsets.all(10),
+        tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+        minimumSize: Size.zero,
```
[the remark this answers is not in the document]
chrisbobbe: Bump on removing these if they don't do anything: https://github.com/zulip/zulip-flutter/pull/853#discussion_r1727692665

lib/widgets/action_sheet.dart
```diff
@@ -12,51 +15,201 @@ import 'actions.dart';
...
+                  Column(
+                    mainAxisSize: MainAxisSize.min,
+                    children: [
+                      // Serves as the top dynamic padding, which changes when
+                      // the action sheet is scrolled in the upper direction.
+                      _ScrollControllerBuilder(
+                        scrollController: scrollController,
+                        builder: (_, scrollController) => SizedBox(
+                          height: math.max(
+                            16 - scrollController.position.extentBefore,
```
chrisbobbe, line 117: Bump on https://github.com/zulip/zulip-flutter/pull/853#discussion_r1725975882 (using 8 here instead of 16).

chrisbobbe (commented):
Thanks for the revision!

I see your explanation at https://github.com/zulip/zulip-flutter/pull/853#discussion_r1727992678 about the `FutureBuilder`, etc., and I think I've found some working code that seems much simpler, and also follows the Figma more closely.

In particular, I think the "dynamic padding" that you mention in some code comments isn't something we want. Here's the description from https://github.com/zulip/zulip-flutter/pull/853#discussion_r1720301682 of what I saw in the Figma, which I still think is accurate:

> When the buttons area can scroll (when there are more buttons than fit on the user's screen), the content, as it scrolls, should be allowed to show through this 8px gap, but masked with a gradient from transparent to DesignVariables.bgContextMenu: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-42553&m=dev
> 
> (Same with an area of height 8px above the buttons.)

and some working code for that:

<details>

```dart
    // Pad the bottom inset. The left/top/right insets are already handled by
    // `showModalBottomSheet.useSafeArea: true` above, which keeps the sheet
    // out of those insets.
    return SafeArea(
      minimum: const EdgeInsets.only(bottom: 16),
      child: Padding(
        padding: const EdgeInsets.fromLTRB(16, 8, 16, 0),
        child: Column(
          children: [
            // TODO(#217): show message text
            Flexible(
              child: Stack(
                children: [
                  SingleChildScrollView(
                    padding: const EdgeInsets.symmetric(vertical: 8),
                    child: ClipRRect(
                      borderRadius: BorderRadius.circular(7),
                      child: Column(spacing: 1, children: optionButtons))),
                  Positioned(
                    top: 0, right: 0, left: 0,
                    child: Container(
                      height: 8,
                      decoration: BoxDecoration(
                        gradient: LinearGradient(
                          begin: Alignment.topCenter,
                          end: Alignment.bottomCenter,
                          colors: [
                            designVariables.bgContextMenu,
                            designVariables.bgContextMenu.withOpacity(0),
                          ])))),
                  Positioned(
                    bottom: 0, right: 0, left: 0,
                    child: Container(
                      height: 8,
                      decoration: BoxDecoration(
                        gradient: LinearGradient(
                          begin: Alignment.topCenter,
                          end: Alignment.bottomCenter,
                          colors: [
                            designVariables.bgContextMenu.withOpacity(0),
                            designVariables.bgContextMenu,
                          ])))),
                ])),
            const MessageActionSheetCancelButton(),
          ])));
```

</details>

i.e.:

- letting the bottom of the scroll's viewport extend all the way to the top edge of the cancel button, but with 8px bottom padding on the content and an 8px gradient overlay stacked on top of that in the z-direction
- letting the top of the scroll's viewport extend as far as 8px from the top of the sheet, but with 8px top padding on the content and an 8px gradient overlay stocked on top of that in the z-direction

With the `SizedBox` whose height changes on the scroll extent, the vertical displacement doesn't seem natural. If the scroll extent is small but nonzero, then when I drag my finger, the content doesn't track exactly with my finger position.

My proposed code could maybe be further refactored to deduplicate some code, but I think it's an improvement over the solution here, which is still complicated enough that I'm not sure I've really understood it. 🙂 If you'd still like to show your proposal to Greg (after his vacation), that's fine; in that case you can just include a version of my proposal in a followup commit without squashing it into the main commit.

sm-sayedi:
Thanks @chrisbobbe for the review and the proposed approach! I prefer yours and there is no need for the "dynamic padding".
Changes pushed. Please have a look!

test/widgets/action_sheet_test.dart
```diff
@@ -616,4 +616,27 @@ void main() {
       check(mockSharePlus.sharedString).isNull();
     });
   });
+
+  group('MessageActionSheetCancelButton', () {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+
+    Finder cancelButtonFinder() => find.text(zulipLocalizations.dialogCancel);
```
chrisbobbe, line 623: Can this just be a `Finder`, instead of a function that returns a `Finder`?

lib/widgets/action_sheet.dart
```diff
@@ -41,21 +44,76 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+                              designVariables.bgContextMenu,
+                            ])))),
+                  ],
+                ),
+              ),
+              const MessageActionSheetCancelButton(),
+            ],
+          ),
+        ),
+      );
```
chrisbobbe, line 116: nit: indentation

chrisbobbe (commented):
Thanks! Small comments below.

Then about the shadow effect, I noticed that @PIG208 has implemented basically the same thing—take a look at `_InsetShadowBox` in the current revision of #928. If that seems similar enough, it could be a nice opportunity to pull out a helper widget. (I wouldn't block on doing that, though; we can always do it as a followup to both PRs.)

Marking for @gnprice's review.

PIG208:
`_InsetShadowBox` was [previously](https://github.com/zulip/zulip-flutter/commit/e16251febac9b000c50969845025bc4a1b54e67a) implemented with a `SingleChildScrollView` wrapped by a `NotificationListener`.

The [current implementation](https://github.com/PIG208/zulip-flutter/commit/8c2469a8276f0e6338a320f85e6d3cbf490f13a0) got a lot simpler because I realized that we always have some paddings such that the content does not get covered by the gradient (and thus the shadow is not visible without scrolling). It seems that we also have the paddings here, it should be a straightforward refactor later to reuse the widget.

That PR will probably land later than this one, but you can cherry-pick ccf2cf76bac8e012f60458be5d7fab64d2e81cb3 to use `InsetShadowBox` here.

lib/widgets/draggable_scrollable_modal_bottom_sheet.dart
```diff
@@ -1,118 +0,0 @@
...
-  required BuildContext context,
-  required WidgetBuilder builder,
-}) {
-  return showModalBottomSheet<T>(
-    context: context,
-
-    // Clip.hardEdge looks bad; Clip.antiAliasWithSaveLayer looks pixel-perfect
-    // on my iPhone 13 Pro but is marked as "much slower":
-    //   https://api.flutter.dev/flutter/dart-ui/Clip.html
-    clipBehavior: Clip.antiAlias,
```
gnprice, line 94: This comment seems like it's just as relevant after this change as before. So it should get moved along with the `clipBehavior: Clip.antiAlias,` line.

lib/widgets/action_sheet.dart
```diff
@@ -431,3 +505,28 @@ class ShareButton extends MessageActionSheetMenuItemButton {
     }
   }
 }
+
+class MessageActionSheetCancelButton extends StatelessWidget {
```
gnprice, line 509: nit: Let's put this right after `MessageActionSheetMenuItemButton` — there's a lot of similarities in their code, so it's helpful to see them at the same time.

test/widgets/action_sheet_test.dart
gnprice: nit: also a good name would be `findCancelButton` — that way it's parallel to the `find.foo` names from `flutter_test` upstream

gnprice (commented):
Thanks @sm-sayedi, and thanks @rajveermalviya and @chrisbobbe for the previous reviews!

Generally this all looks great. For the shadows, I think #990 will merge soon; then you can rebase atop that, which I think will simplify this code somewhat. (I'll also do a bit of manual testing of that aspect of the UI at that stage.)

Other than that, small comments below, and Chris has a couple of small comments open at https://github.com/zulip/zulip-flutter/pull/853#pullrequestreview-2353061398 above.

Let's also squash the "delete the old way" commit into the main commit here. Because the new code is replacing the old, the comparison is helpful for understanding some aspects of what the new code is doing.

test/widgets/action_sheet_test.dart
sm-sayedi: > Can this just be a `Finder`, instead of a function that returns a `Finder`?

It could be a direct `Finder` but I think making it a function with a relevant name makes the code a little bit more readable. 🙂

sm-sayedi:
Thanks @chrisbobbe and @gnprice for the reviews. Changes pushed.

test/widgets/action_sheet_test.dart
gnprice: I think Chris's suggestion was to make it a local variable so it still has a relevant name — just not a function:
```dart
    final findCancelButton = find.text(zulipLocalizations.dialogCancel);
```

lib/widgets/action_sheet.dart
```diff
@@ -41,21 +44,48 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+              // TODO(#217): show message text
+              Flexible(child: InsetShadowBox(
+                top: 8, bottom: 8,
+                color: DesignVariables.of(context).bgContextMenu,
+                child: SingleChildScrollView(
+                  padding: const EdgeInsets.only(top: 16, bottom: 8),
+                  child: ClipRRect(
+                    borderRadius: BorderRadius.circular(7),
+                    child: Column(spacing: 1, children: optionButtons),
+                  )))),
```
gnprice, line 86: nit:
```suggestion
                    child: Column(spacing: 1,
                      children: optionButtons))))),
```

* our standard close-paren formatting
* and perhaps more importantly: keeps `children: optionButtons`, which is the ultimate payload that contains the meaningful content which everything else here is wrapping, visible on its own line

gnprice (commented):
Thanks for the revision! One nit below, and one open above at https://github.com/zulip/zulip-flutter/pull/853#discussion_r1790967468 .

Otherwise the code all looks good. I said above I'd also do a final bit of manual testing; I won't get to that today but will put it on my queue for early next week.

(I have had earlier versions of this PR deployed on my phone off and on over the last few weeks, and the UI has looked great.)

sm-sayedi:
Thanks for the review @gnprice. Changes applied, PTAL.

gnprice:
All looks good; and I did that manual testing of the scrolling behavior, and that looks good too. Merging.

Thanks again @sm-sayedi for all your work on this!

PR #854, opened 2024-07-31
unread_count_badge: Clear out all `TODO(#95)`s

chrisbobbe:
For the user-facing change in this commit:

```
unread_count_badge: Clear out a TODO by following Vlad's design
```

The difference (in the text color of the unread-count badges on the streams/topics) really is tiny. I used Mac's Digital Color Meter to see that there was in fact a difference, though.

| Before | After |
| --- | --- |
| BDA772E1-FEA1-4AE7-80E0-4A26BB09866B | 2DF3752E-B708-4676-848E-B2932EEB6119 |


chrisbobbe:
Resolved the test failure.

gnprice:
Wow, indeed, that sure is a subtle color difference.

All looks good, thanks — merging.

PR #856, opened 2024-08-01
Replace `flutter_local_notifications` with pigeon bindings

rajveermalviya:
Fixes: #351

### DIFF
--- a/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
+++ b/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
@@ -3,6 +3,7 @@ package com.zulip.flutter
 import android.annotation.SuppressLint
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
 import android.util.Log
 import androidx.annotation.Keep
@@ -97,8 +98,12 @@ private class AndroidNotificationHost(val context: Context)
             contentIntent?.let { setContentIntent(
                 android.app.PendingIntent.getActivity(context,
                     it.requestCode.toInt(),
-                    it.intent.let { intent -> Intent(context, MainActivity::class.java).apply {
-                        action = intent.action
+                    it.intent.let { intent -> Intent(
+                        intent.action,
+                        Uri.parse(intent.dataUrl),
+                        context,
+                        MainActivity::class.java
+                    ).apply {
                         intent.extras.forEach { (k, v) -> putExtra(k!!, v!!) }
                     } },
                     it.flags.toInt())
--- a/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
+++ b/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
@@ -105,6 +105,7 @@ private class AndroidNotificationHost(val context: Context)
                         MainActivity::class.java
                     ).apply {
                         intent.extras.forEach { (k, v) -> putExtra(k!!, v!!) }
+                        flags = intent.flags.toInt()
                     } },
                     it.flags.toInt())
             ) }
--- a/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
+++ b/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
@@ -97,13 +97,10 @@ private class AndroidNotificationHost(val context: Context)
             contentIntent?.let { setContentIntent(
                 android.app.PendingIntent.getActivity(context,
                     it.requestCode.toInt(),
-                    Intent(context, MainActivity::class.java).apply {
-                        // This action name and extra name are special to
-                        // FlutterLocalNotificationsPlugin, which handles receiving the Intent.
-                        // TODO take care of receiving the notification-opened Intent ourselves
-                        action = "SELECT_NOTIFICATION"
-                        putExtra("payload", it.intentPayload)
-                    },
+                    it.intent.let { intent -> Intent(context, MainActivity::class.java).apply {
+                        action = intent.action
+                        intent.extras.forEach { (k, v) -> putExtra(k!!, v!!) }
+                    } },
                     it.flags.toInt())
             ) }
             contentText?.let { setContentText(it) }
--- a/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
+++ b/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
@@ -104,7 +104,6 @@ private class AndroidNotificationHost(val context: Context)
                         context,
                         MainActivity::class.java
                     ).apply {
-                        intent.extras.forEach { (k, v) -> putExtra(k!!, v!!) }
                         flags = intent.flags.toInt()
                     } },
                     it.flags.toInt())
--- a/lib/host/android_notifications.dart
+++ b/lib/host/android_notifications.dart
@@ -60,3 +60,11 @@ abstract class PendingIntentFlag {
   /// Corresponds to `FLAG_ALLOW_UNSAFE_IMPLICIT_INTENT`.
   static const allowUnsafeImplicitIntent = 1 << 24;
 }
+
+/// For use in [AndroidIntent.action].
+///
+/// See: https://developer.android.com/reference/android/content/Intent#constants_1
+abstract class IntentAction {
+  /// Corresponds to `ACTION_VIEW`.
+  static const view = 'android.intent.action.VIEW';
+}
--- a/lib/host/android_notifications.dart
+++ b/lib/host/android_notifications.dart
@@ -68,3 +68,14 @@ abstract class IntentAction {
   /// Corresponds to `ACTION_VIEW`.
   static const view = 'android.intent.action.VIEW';
 }
+
+/// For use in [AndroidIntent.flags].
+///
+/// See: https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_BROUGHT_TO_FRONT
+abstract class IntentFlag {
+  /// Corresponds to `FLAG_ACTIVITY_CLEAR_TOP`.
+  static const activityClearTop = 1 << 26;
+
+  /// Corresponds to `FLAG_ACTIVITY_NEW_TASK`.
+  static const activityNewTask = 1 << 28;
+}
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -196,10 +196,22 @@ class NotificationDisplayManager {
         intent: AndroidIntent(
           action: IntentAction.view,
           dataUrl: intentDataUrl.toString(),
-          extras: {}),
-        // TODO this doesn't set the Intent flags we set in zulip-mobile; is that OK?
-        //   (This is a legacy of `flutter_local_notifications`.)
-        ),
+          extras: {},
+          // See these sections in the Android docs:
+          //   https://developer.android.com/guide/components/activities/tasks-and-back-stack#TaskLaunchModes
+          //   https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_CLEAR_TOP
+          //
+          // * From the doc on `PendingIntent.getActivity` at
+          //     https://developer.android.com/reference/android/app/PendingIntent#getActivity(android.content.Context,%20int,%20android.content.Intent,%20int)
+          //   > Note that the activity will be started outside of the context of an
+          //   > existing activity, so you must use the Intent.FLAG_ACTIVITY_NEW_TASK
+          //   > launch flag in the Intent.
+          //
+          // * The flag FLAG_ACTIVITY_CLEAR_TOP is mentioned as being what the
+          //   notification manager does; so use that.  It has no effect as long
+          //   as we only have one activity; but if we add more, it will destroy
+          //   all the activities on top of the target one.
+          flags: IntentFlag.activityClearTop | IntentFlag.activityNewTask)),
       autoCancel: true,
     );
 
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -204,7 +204,14 @@ class NotificationDisplayManager {
         // TODO is setting PendingIntentFlag.updateCurrent OK?
         //   (That's a legacy of `flutter_local_notifications`.)
         flags: PendingIntentFlag.immutable | PendingIntentFlag.updateCurrent,
-        intentPayload: jsonEncode(dataJson),
+        intent: AndroidIntent(
+          // This action name and extra name are special to
+          // FlutterLocalNotificationsPlugin, which handles receiving the Intent.
+          // TODO take care of receiving the notification-opened Intent ourselves
+          action: 'SELECT_NOTIFICATION',
+          extras: <String, String>{
+            'payload': jsonEncode(dataJson),
+          }),
         // TODO this doesn't set the Intent flags we set in zulip-mobile; is that OK?
         //   (This is a legacy of `flutter_local_notifications`.)
         ),
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -192,10 +192,7 @@ class NotificationDisplayManager {
         // The intent data URL is distinct for each conversation, so this value
         // doesn't matter.
         requestCode: 0,
-
-        // TODO is setting PendingIntentFlag.updateCurrent OK?
-        //   (That's a legacy of `flutter_local_notifications`.)
-        flags: PendingIntentFlag.immutable | PendingIntentFlag.updateCurrent,
+        flags: PendingIntentFlag.immutable,
         intent: AndroidIntent(
           action: IntentAction.view,
           dataUrl: intentDataUrl.toString(),
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -196,7 +196,6 @@ class NotificationDisplayManager {
         intent: AndroidIntent(
           action: IntentAction.view,
           dataUrl: intentDataUrl.toString(),
-          extras: {},
           // See these sections in the Android docs:
           //   https://developer.android.com/guide/components/activities/tasks-and-back-stack#TaskLaunchModes
           //   https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_CLEAR_TOP
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -133,7 +133,7 @@ class NotificationDisplayManager {
     } else {
       messagingStyle = MessagingStyle(
         user: Person(
-          key: _personKey(data.realmUri, data.userId),
+          key: _personKey(data.realmUrl, data.userId),
           name: zulipLocalizations.notifSelfUser),
         messages: [],
         isGroupConversation: switch (data.recipient) {
@@ -164,7 +164,7 @@ class NotificationDisplayManager {
       text: data.content,
       timestampMs: data.time * 1000,
       person: Person(
-        key: _personKey(data.realmUri, data.senderId),
+        key: _personKey(data.realmUrl, data.senderId),
         name: data.senderFullName,
         iconBitmap: await _fetchBitmap(data.senderAvatarUrl))));
 
@@ -223,7 +223,7 @@ class NotificationDisplayManager {
       smallIconResourceName: 'zulip_notification', // This name must appear in keep.xml too: https://github.com/zulip/zulip-flutter/issues/528
       inboxStyle: InboxStyle(
         // TODO(#570) Show organization name, not URL
-        summaryText: data.realmUri.toString()),
+        summaryText: data.realmUrl.toString()),
 
       // On Android 11 and lower, if autoCancel is not specified,
       // the summary notification may linger even after all child
@@ -336,10 +336,10 @@ class NotificationDisplayManager {
   static String _groupKey(FcmMessageWithIdentity data) {
     // The realm URL can't contain a `|`, because `|` is not a URL code point:
     //   https://url.spec.whatwg.org/#url-code-points
-    return "${data.realmUri}|${data.userId}";
+    return "${data.realmUrl}|${data.userId}";
   }
 
-  static String _personKey(Uri realmUri, int userId) => "$realmUri|$userId";
+  static String _personKey(Uri realmUrl, int userId) => "$realmUrl|$userId";
 
   static void _onNotificationOpened(NotificationResponse response) async {
     final payload = jsonDecode(response.payload!) as Map<String, dynamic>;
@@ -366,7 +366,7 @@ class NotificationDisplayManager {
 
     final globalStore = GlobalStoreWidget.of(context);
     final account = globalStore.accounts.firstWhereOrNull((account) =>
-      account.realmUrl == data.realmUri && account.userId == data.userId);
+      account.realmUrl == data.realmUrl && account.userId == data.userId);
     if (account == null) return; // TODO(log)
 
     final narrow = switch (data.recipient) {
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -189,18 +189,9 @@ class NotificationDisplayManager {
       },
 
       contentIntent: PendingIntent(
-        // TODO make intent URLs distinct, instead of requestCode
-        //   (This way is a legacy of flutter_local_notifications.)
-        //   The Intent objects we make for different conversations look the same.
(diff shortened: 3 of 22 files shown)


gnprice:
Exciting!

LMK if there are questions you have for me at this stage, or particular spots in the code you'd like an early high-level review of.

rajveermalviya:
@gnprice, I have a question about whether this implementation is desirable. Currently it creates a pending intent using the [ACTION_VIEW](https://developer.android.com/reference/android/content/Intent#ACTION_VIEW) action with a [URL payload](https://github.com/zulip/zulip-flutter/blob/07475510fcd0310f5765a6ec98edc6ab1d5a9c15/lib/api/model/notification.dart#L70-L100) for each notification. When the user opens the notification, the intent functions as a deeplink and is handled by the existing URL handling in Flutter. Then this url is parsed and handled here for incoming intents [while the app is open](https://github.com/zulip/zulip-flutter/blob/07475510fcd0310f5765a6ec98edc6ab1d5a9c15/lib/widgets/app.dart#L90-L99) and [while the app is closed](https://github.com/zulip/zulip-flutter/blob/07475510fcd0310f5765a6ec98edc6ab1d5a9c15/lib/widgets/app.dart#L148-L165).

Also, when the intent is created we set explicity [set the reciever component](https://github.com/zulip/zulip-flutter/blob/07475510fcd0310f5765a6ec98edc6ab1d5a9c15/android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt#L90) so that ensures that this instance of `zulip://notificiation_open` url intent is only handled by our app.

This way this implementation avoids creating a custom intent handler and reuses Flutter existing URL route handler.

rajveermalviya:
Moved the discussion to chat thread here — https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20Drop.20.60flutter_local_notifications.60.20.23F856/near/1910352

Khader-1 (approved):
Thanks @rajveermalviya!

kenclary (approved):
lgtm

lib/widgets/app.dart
```diff
@@ -143,16 +146,31 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
               MaterialWidgetRoute(page: const ChooseAccountPage()),
-              if (initialAccountId != null) ...[
+              if (notificationRoute != null && accountId != null)...[
+                HomePage.buildRoute(accountId: accountId),
+                InboxPage.buildRoute(accountId: accountId),
+                notificationRoute,
+              ] else if (initialAccountId != null) ...[
                 HomePage.buildRoute(accountId: initialAccountId),
                 InboxPage.buildRoute(accountId: initialAccountId),
               ],
```
chrisbobbe, line 170:
```
notif: Introduce URL based notif open implementation
```

This looks like maybe a behavior change, right? Before, when you opened the app from a notification, the home page and inbox page would be for the first account in the list (the "initial account" in `initialAccountId`), even if that was different from the notification's account. Now, those pages are chosen to match the notification.

I agree the new behavior seems better. Let's mention the change, though, and any other user-visible behavior changes, in the commit message.

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
```diff
@@ -87,7 +87,7 @@ private class AndroidNotificationHost(val context: Context)
                         // FlutterLocalNotificationsPlugin, which handles receiving the Intent.
                         // TODO take care of receiving the notification-opened Intent ourselves
```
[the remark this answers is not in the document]
chrisbobbe: Also, I see that this special name "payload" remains after removing `flutter_local_notifications`. Is that because we're happy/neutral to keep it around, or is it a sign that there's still helpful cleanup of `flutter_local_notifications` to do?

chrisbobbe (commented):
Thanks! Comments below, all small. Also, I haven't done any manual testing of this; did you?

chrisbobbe:
It looks like there's been some further discussion on these changes: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/narrow.20links.20with.20userid.3F/near/1941174

Please post here when that's resolved and this is ready for another review. 🙂

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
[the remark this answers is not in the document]
rajveermalviya: Updated with main commit (`notif: Handle opening of notification via URI intents`) removing dangling references for `payload`.

lib/widgets/app.dart
rajveermalviya: Removed this behaviour (for now) to keep the diff smaller, this would probably be better done in a separate PR.

rajveermalviya:
Apologies for the delay in the revision. Thanks for the previous review @chrisbobbe, pushed a new revision, PTAL.

chrisbobbe:
Hmm, looks like some tests are failing, could you take a look?

rajveermalviya:
Weird, I can't repro locally. Error [in CI](https://github.com/zulip/zulip-flutter/actions/runs/11355816054/job/31585881246#step:7:1065) is:
```
```

It seems unrelated to this PR, because I am getting the [same failure on `main`](https://github.com/rajveermalviya/zulip-flutter/actions/runs/11303397558/job/31586654377#step:7:1065) in my fork.
I'll investigate it later in the day, unless someone else does it first.

chrisbobbe:
Rerunning the CI checks; this should fix it: https://github.com/zulip/zulip-flutter/pull/999#issuecomment-2417351945

chrisbobbe (commented):
Thanks! Small comments below from a very light review; Greg knows this system better than I do, and I'll go ahead and mark for his review to save some time. 🙂

lib/notifications/display.dart
```diff
@@ -94,34 +95,48 @@ class NotificationChannelManager {
...
+    if (isBackground) {
+      _backgroundIsolateGlobalStore ??= await ZulipBinding.instance.loadGlobalStore();
+      globalStore = _backgroundIsolateGlobalStore!;
+    } else {
+      NavigatorState navigator = await ZulipApp.navigator;
+      final context = navigator.context;
+      assert(context.mounted);
+      if (!context.mounted) return; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
+      globalStore = GlobalStoreWidget.of(context);
+    }
```
gnprice, line 132: Hmm, it feels like the wrong structure for this NotificationDisplayManager class to be maintaining its own GlobalStore instance (in _backgroundIsolateGlobalStore).

Having this method, and its caller chain, need to know about whether it's running in a background isolate or in the foreground seems messy too.

I've just sent #1005 to solve this a different way. Then this becomes:
```suggestion
    final globalStore = await ZulipBinding.instance.getGlobalStore();
```

lib/notifications/display.dart
```diff
@@ -395,4 +391,59 @@ class NotificationDisplayManager {
...
+  ///
+  /// This uri is used to route to the relevant conversation's [MessageListPage]
+  /// when a notification with this intent uri is opened.
+  @visibleForTesting
+  static Future<Uri?> notificationIntentUriFromFcmMessage({
+    required GlobalStore globalStore,
+    required MessageFcmMessage data,
+  }) async {
+    final account = globalStore.accounts.firstWhereOrNull((account) =>
+      account.realmUrl.origin == data.realmUri.origin && account.userId == data.userId);
```
gnprice, line 406: notif: Match realmUri using only the origin part of the uri

Can you elaborate on the motivation for this change? What's the situation where it makes a difference, and why is the new behavior better than the old behavior in that situation?

gnprice (commented):
Thanks @rajveermalviya for taking care of this, and thanks @chrisbobbe for the previous reviews!

For tonight I've just skimmed this; a couple of comments above.

test/notifications/display_test.dart
```diff
@@ -291,39 +290,30 @@ void main() {
...
-      testBinding.firebaseMessaging.onMessage.add(
-        RemoteMessage(data: data.toJson()));
-      async.flushMicrotasks();
-      checkNotification(data,
-        messageStyleMessages: [data],
-        expectedIsGroupConversation: expectedIsGroupConversation,
-        expectedTitle: expectedTitle,
-        expectedTagComponent: expectedTagComponent);
-      testBinding.androidNotificationHost.clearActiveNotifications();
+    void testNotification(String description, Future<void> Function(Future<void> Function(FcmMessage data) receiveFcmMessage) body, {http.Client Function()? httpClientFactory}) {
```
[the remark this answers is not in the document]
rajveermalviya: After rebasing past #1005, creation of foreground notifications no longer rely on `GlobalStoreWidget` inherited widget, thus not needing `testWidget`.  So now removed this refactor commit.

lib/notifications/display.dart
rajveermalviya: Added some reasoning in the commit message, but mainly it's because when re-creating the realmUri from narrowUri (using `Uri.parse(narrowUri.origin)`), we don't get the trailing `/` which is present in `account.realmUri`, resulting in matching to fail.

rajveermalviya:
Thanks for the review @chrisbobbe and @gnprice!
@gnprice I've now rebased this PR on top of #1005 with suggested changes, PTAL.

chrisbobbe:
I've unassigned myself since it's Greg's turn to review. 🙂

One thing was weird though. Here's that update as GitHub shows it here:

image

But the GitHub bot on CZO rendered it this way:

https://chat.zulip.org/#narrow/channel/494-mobile-github/topic/zulip-flutter/near/1964999
> chrisbobbe unassigned gnprice to [PR #856 Replace `flutter_local_notifications` with pigeon bindings](https://github.com/zulip/zulip-flutter/pull/856).

gnprice:
Seems like a bug in our GitHub bot. The #integrations channel would be a good place to report it; I think we have a contributor currently who'd be likely to pick up an issue like that.

chrisbobbe:
Sure; reported there: https://chat.zulip.org/#narrow/channel/127-integrations/topic/GitHub.3A.20Wrong.20user.20identified.20as.20unassigned/near/1965136

lib/notifications/display.dart
```diff
@@ -341,46 +338,25 @@ class NotificationDisplayManager {
...
     final context = navigator.context;
     assert(context.mounted);
-    if (!context.mounted) return; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
+    if (!context.mounted) return false; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
 
     final globalStore = GlobalStoreWidget.of(context);
-    final account = globalStore.accounts.firstWhereOrNull((account) =>
-      account.realmUrl == data.realmUri && account.userId == data.userId);
-    if (account == null) return; // TODO(log)
+    final result = await _narrowFromNotificationIntentUri(globalStore: globalStore, uri: uri);
```
gnprice, line 351: Hmm — I think this gets a bit sticky for the UX in the case where Zulip wasn't already running and you open a notification. (Or where it was already running but hadn't loaded data for the account the notification is for.) The trouble is that this adds an `await` for getting the per-account store, before we do any navigation.

In main, `data` contains all the information we need for constructing a `Narrow`; so this `_navigateForNotification` method synchronously constructs a `Narrow` and immediately calls `navigator.push` for the appropriate `MessageListPage`, with no waiting. If the data isn't already loaded, the user may then see a loading page for a while… but at least they do see a loading page.

In the current version of this PR, I believe what happens in that situation is that we'll get to this line, and to the `await globalStore.perAccount(…)` line inside `_narrowFromNotificationIntentUri`, and then it'll sit there for however long it takes for the data to load — and during that time, there'll be no feedback to the user indicating that the notification did anything different from just launching the app.

(Well, I guess if the notification came from the first account in their list, then launching the app would have started that account's data loading and shown a corresponding loading page anyway. So in that case, which is the most common case because (presumably) most users have only one account, it'll work out fine. But if the notification is for an account that's later in the list, and if the app was already open or if the first account in the list is faster to load than the one the notification is for, the problem arises: we'll be showing a non-loading page that has nothing to do with the notification, with no visible indication that the notification had any effect, and then some time later the notification suddenly belatedly causes a navigation.)

---

It'd be possible to solve this problem head-on: basically we'd add a page that just takes the narrow URL as an argument, instead of a `Narrow`, and navigate there. Then that would show a loading page until the data arrives, at which point it'd replace itself with a normal MessageListPage. We'd need to take some care to avoid unnecessary page-transition animations, as well as an unnecessary flicker in the case where the data is already present.

We'll probably end up wanting some form of that head-on solution eventually, as part of handling Zulip links from outside the app, which is a post-launch feature.

But I think a simpler solution is that we can change the form of these `zulip://notification/…` URLs so that they directly encode the data we need: it's either a channel/stream ID and a topic, or a list of recipient IDs. Let's go with that solution for now, to keep down the complexity of this transition.

lib/notifications/display.dart
```diff
@@ -395,4 +371,60 @@ class NotificationDisplayManager {
...
+    }
+
+    final narrow = switch (data.recipient) {
+      FcmMessageChannelRecipient(:var streamId, :var topic) =>
+        TopicNarrow(streamId, topic),
+      FcmMessageDmRecipient(:var allRecipientIds) =>
+        DmNarrow(allRecipientIds: allRecipientIds, selfUserId: account.userId),
+    };
+    final perAccountStore = await globalStore.perAccount(account.id);
+    final narrowUri = narrowLink(perAccountStore, narrow, nearMessageId: data.zulipMessageId);
```
gnprice, line 399: nit: use just `store` for the name of a PerAccountStore local variable
```suggestion
    final store = await globalStore.perAccount(account.id);
    final narrowUri = narrowLink(store, narrow, nearMessageId: data.zulipMessageId);
```

That's the convention we consistently use across the codebase. Then when there is a GlobalStore, it's distinguished explicitly as `globalStore`, even if there doesn't happen to be a PerAccountStore in scope at the same spot.

lib/notifications/display.dart
```diff
@@ -118,10 +109,15 @@ class NotificationDisplayManager {
 
   static Future<void> _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) async {
     assert(debugLog('notif message content: ${data.content}'));
+
+    final globalStore = await ZulipBinding.instance.getGlobalStore();
     final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
     final groupKey = _groupKey(data);
     final conversationKey = _conversationKey(data, groupKey);
 
+    final intentUri = await notificationIntentUriFromFcmMessage(globalStore: globalStore, data: data);
```
gnprice, line 118: This has a similar latency issue to the one in my previous comment on opening the notification.

Here, I think the symptom would be: if a notification arrives and the app isn't already running (or doesn't yet have data for the account the notification is for), we'd end up having to sit here for however long it takes to load the per-account data. That could be a while — even a few seconds (like users of chat.zulip.org would see when on a good network connection on the same continent as the server) is enough to make notifications feel laggy in the context of a messaging application.

For users on a flaky or slow network connection, especially in a big realm like chat.zulip.org, the delay could be more like a minute, and could be enough to make the OS time out and stop the app, so the notification doesn't appear at all. In particular I think if the device is asleep it might (though I'm not sure) remain in a lower-power state even while handling a notification, which would further slow down the download of a couple of MB of server data.

Unlike the situation when opening a notification, this would affect even users who have only one account.

So I think we definitely want to avoid pulling in the per-account store on the path to displaying a notification. We can do that with the same solution discussed at the other end: just encode the notification's narrow information directly, rather than converting to a full Zulip narrow URL.

I guess this means the `getGlobalStore` method from #1005 probably isn't needed right now after all; oh well.

At some point in the future (solidly post-launch), we'll want to use per-account server data in displaying a notification, to do things like showing the names of other recipients in a DM thread. A prerequisite for that will be to store server data in the app's local database, so that we can load a version for it without waiting to load everything from the server from scratch.

lib/notifications/display.dart
```diff
@@ -169,8 +165,6 @@ class NotificationDisplayManager {
         iconBitmap: await _fetchBitmap(data.senderAvatarUrl))));
 
     await _androidHost.notify(
-      // TODO the notification ID can be constant, instead of matching requestCode
-      //   (This is a legacy of `flutter_local_notifications`.)
       id: notificationIdAsHashOf(conversationKey),
```
gnprice, line 168: This TODO suggests we can make the notification ID a constant, instead of this hash. Can we do that now, or is there still an obstacle to it?

gnprice (commented):
OK, and I've now finished reading up through these commits:

1f3c69c86 notif: Match realmUri using only the origin part of the URI
5bf5ad831 notif [nfc]: Introduce `AndroidIntent`; replace `PendingIntent.intentPayload`
b3b60413c internal_link [nfc]: Make `encodeHashComponent` public
0330389d0 notif test: Fix incorrect test accounts with same accountIds
22a969172 notif: Handle opening of notification via URI intents

Comments below. I haven't yet read the test code in the last commit, because I think those changes will get simplified significantly following some of these comments.

lib/notifications/display.dart
rajveermalviya: Removed this change because `narrowUri` doesn't exist anymore, thus this change is not required anymore. Will send a separate PR to fix `example_data.dart`, with corresponding tests.

rajveermalviya:

test/notifications/display_test.dart
```diff
@@ -89,6 +89,28 @@ RemoveFcmMessage removeFcmMessage(List<Message> zulipMessages, {Account? account
   }) as RemoveFcmMessage;
 }
 
+Uri messageIntentDataUrl(MessageFcmMessage data) {
```
gnprice, line 92: I think this can be replaced with just calling `NotificationOpenPayload.buildUrl`. We'll want unit tests for that class as I mentioned above; but those will give us test coverage on that method more thoroughly anyway, I think, than we get from reproducing the URLs here in the context of test cases that are focused on other logic (namely `_onMessageFcmMessage`).

gnprice (commented):
Thanks for the revision! This approach looks good.

Comments below, mainly small. I've (re)read through what are now the first three commits:
11d8c1394 notif [nfc]: Introduce AndroidIntent; Replace PendingIntent.intentPayload
847c96a1e notif test: Fix incorrect test accounts with same account-ids
2e7a3b229 notif: Handle opening of notification on Android using data URL intents

including now the test code in the last of those commits. (Which is the main commit of the branch, so I think we've covered at this point most of the complexity in the PR.)

rajveermalviya:
Thanks for the review @gnprice! New revision pushed, PTAL.

gnprice (commented):
Thanks for the revision! I've re-reviewed the branch up through the main commit, which is now the first four commits:
ace76d744 notif: Use newer FcmMessageWithIdentity.realmUrl; with compatibility fallback
1cc270d01 notif [nfc]: Introduce AndroidIntent; replace PendingIntent.intentPayload
b506d3088 notif test: Fix incorrect test accounts with same account-ids
8faca5a14 notif: Handle opening of notification on Android using data URL intents

and just a few small comments below.

Trying to get my review queue back to empty today, so I'll leave this one here for now. Then I'll go through the rest of the branch in a subsequent round.

rajveermalviya:

pubspec.yaml
```diff
@@ -45,8 +45,6 @@ dependencies:
   firebase_core: ^3.1.0
   firebase_messaging: ^15.0.1
   flutter_color_models: ^1.3.3+2
-  flutter_local_notifications: ^17.2.1
```
gnprice, line 48: Searching for remaining references to this package, I find just one:
```
$ git grep -i 'flutter.?local.?notif'
android/app/src/main/res/raw/keep.xml:     The issue is that the package:flutter_local_notifications API has us
```

So that should be updated too.

lib/host/android_notifications.dart
```diff
@@ -60,3 +60,22 @@ abstract class PendingIntentFlag {
...
+/// See: https://developer.android.com/reference/android/content/Intent#constants_1
+abstract class IntentAction {
+  /// Corresponds to `ACTION_VIEW`.
+  static const view = 'android.intent.action.VIEW';
+}
+
+/// For use in [AndroidIntent.flags].
+///
+/// See: https://developer.android.com/reference/android/content/Intent#constants_1
+abstract class IntentFlag {
```
gnprice, line 75: nit:
```suggestion
/// See: https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_BROUGHT_TO_FRONT
abstract class IntentFlag {
```

There's a ton of constants on that page, so going straight to one of the flags (the one that's currently first in the list) saves some scrolling.

pigeon/notifications.dart
```diff
@@ -36,19 +36,30 @@ class NotificationChannel {
   final Int64List? vibrationPattern;
 }
 
+/// Corresponds to `android.content.Intent`
+///
+/// See:
+///   https://developer.android.com/reference/android/content/Intent
+///   https://developer.android.com/reference/android/content/Intent#Intent(java.lang.String,%20android.net.Uri,%20android.content.Context,%20java.lang.Class%3C?%3E)
+class AndroidIntent {
+  AndroidIntent({required this.action, required this.dataUrl, required this.flags});
```
gnprice, line 45: nit:
```suggestion
  AndroidIntent({required this.action, required this.dataUrl, this.flags = 0});
```

This isn't a parameter of the underlying constructor, so effectively it's optional in the underlying API. (One has to set it later after construction.) So making it optional here seems the most faithful to that API.

gnprice (commented):
Thanks! Finished reviewing the whole branch, and generally this all looks great — just small comments below.

Would you do some manual testing of these changes, if you haven't already? In particular, see my PRs that implemented notifications and made previous changes to them; I describe there some test protocols we should repeat for these changes.

In general, notifications are one of the small number of areas where manual testing is needed to confirm things still work correctly after any significant change. That's because a key part of its behavior depends on interacting correctly with external APIs that (a) are supplied from the platform, so aren't present in a `flutter test` environment, and (b) have a lot of subtle details of their semantics, not all of which are necessarily even documented, and which can be easy to miss even if they are.

rajveermalviya:

Although I was unable to find the mentioned test protocol, I did some fresh manual testing of following scenarios:

- Recieved muliple messages from two different users in two different streams and thus different topics, notifications are correctly displayed and routed to correct conversation when opened.
- Recieved muliple messages from two different users in same stream but two different topics, notifications are correctly displayed and routed to correct conversation when opened.
- Recieved a message in one-one DM, notification is correctly displayed and routed to the DM page when opened.
- Recieved messages in group DM with two users (total 3 users), notification is correctly displayed and routed to the group DM page when opened.

(Tested the above cases with both foreground and background notifications)

- Recieved messages in a stream+topic conversation, notification is correctly displayed, marked as read the conversation from web client, notifications for that specific conversation were removed.
- Recieved messages in a one-one DM,  notification is correctly displayed, marked as read the conversation from web client, notifications for that specific conversation were removed.

rajveermalviya:
In addition to that, I also had the mobile notifications enabled for all channels in multiple (busy) servers to test the realm group summary notifications.

test/api/notifications_test.dart
```diff
@@ -210,18 +224,29 @@ void main() {
       check(parse({ ...baseJson, 'awesome_feature': 'enabled' })).jsonEquals(baseline);
     });
 
+    test('uses deprecated fields when newer fields are missing', () {
+      final baseline = parse(baseJson);
+
+      // FL 257 deprecated 'realm_uri' in favor of 'realm_url'.
+      final jsonSansRealm =
+        { ...baseJson }..remove('realm_url')..remove('realm_uri');
+      check(parse({ ...jsonSansRealm, 'realm_url': 'https://zulip.example.com/' })).jsonEquals(baseline);
```
gnprice, line 233: bump; and I think in fact this revision took this point in the opposite direction :slightly_smiling_face: 

I believe the point of this test is to check that when the old name is present, and not the new name, we get the data from the old name. So the input this test constructs to pass to `parse` should use the old name.

gnprice (commented):
Cool, thanks for that testing. That set of scenarios sounds solid.

The revision also looks good. Just one comment below, and then I think this is ready to merge.

test/api/notifications_test.dart
rajveermalviya: Right, sorry about that! Updated to use the older field.

rajveermalviya:

gnprice:
Looks good — merging. Thanks for all your work getting this transition finished!

PR #861, opened 2024-08-02
Refactor compose-box layout with error banner

gnprice:
This is a followup to #816, as foreshadowed at https://github.com/zulip/zulip-flutter/pull/816#issuecomment-2266030983 .

This lets _ComposeBoxLayout stay focused on the complexity of how the text inputs, the editing buttons, and the send button relate to each other. In the case where an error banner is present, this also lets us skip in the build method some logic to compute things that won't end up getting used.

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -870,13 +870,11 @@ class _ComposeBoxLayout extends StatelessWidget {
     required this.sendButton,
     required this.contentController,
     required this.contentFocusNode,
-    this.blockingErrorBanner,
   });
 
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks! LGTM, and I see this is also marked for @sm-sayedi's review.

sm-sayedi (commented):
Thanks @gnprice for the refactor. `_ComposeBoxLayout` was really getting kind of cluttered! 😀 LGTM now!

gnprice:
Thanks for the reviews! Merging.

PR #864, opened 2024-08-02
Sketch removing notification when message read

### LINKED ISSUE #341: Remove notifications when messages are read, on Android

gnprice:
Fixes #341.

This is a draft that isn't meant to be merged. But in my manual testing, it does work!

I originally wrote this in March, as described at https://github.com/zulip/zulip-flutter/issues/341#issuecomment-2014183497 , and it became the impetus to start work on #351 converting to use Pigeon. So the bulk of that draft branch turned into #592.

This draft PR is the result of me taking the rest of the material from that draft branch and rebasing it. Some of it may be a bit rough because it was part of our first experiment with Pigeon. The commit structure is rough because I left it mostly as it was in the draft branch, only squashing a couple of fixup commits.

Things this needs before we can ship it:
* Use the binding instead of the host API directly, so tests are possible.
* Write a few tests.
* Clean up the commits. A good structure might look like: add one Pigeon method, add the other, then make the `lib/notifications/display.dart` changes.

  To do that, probably just squash everything together and then split into new commits.
* Maybe handle the TODO on `getActiveNotifications` to accept a list of extras to care about, so that we don't take a whole bunch of irrelevant extras and try to serialize them back to Dart.

So @rajveermalviya please take it from here for #341.

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -3,6 +3,7 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/.idea/zulip.iml" filepath="$PROJECT_DIR$/.idea/zulip.iml" />
+      <module fileurl="file://$PROJECT_DIR$/android/zulip_android.iml" filepath="$PROJECT_DIR$/android/zulip_android.iml" />
     </modules>
   </component>
 </project>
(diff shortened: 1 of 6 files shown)


gnprice:
This became #878, which is now merged. Thanks @rajveermalviya!

Issue #186, opened 2023-06-14
Handle unrecoverable long-poll failures


gnprice:
As described in #184, when a long-poll for events fails, the current prototype just stops polling. That issue is about transient failures, and #185 is about a non-transient but recoverable error. That leaves failures that we can't expect to fix, neither by just retrying nor by some other request: e.g., the user's auth credentials were rejected.

At a minimum, in this case we should give the user some indication that the data they're seeing is no longer being updated: e.g. set the `isStale` flag mentioned in #185.

If specifically what happened is that authentication failed, then we should probably ask the user to log in again to that account.

If we do those two things, and also make sure the user always has the option to navigate away so they can add or choose another account to use, then I think that's pretty satisfactory. This situation shouldn't be common, so it's OK if it isn't super smooth as long as there's some reasonable way for the user to recover.

gnprice:
Now that the following two issues are complete:
* #465 
* #555 

the current behavior is:


> At a minimum, in this case we should give the user some indication that the data they're seeing is no longer being updated

This is done (as #465).



> If specifically what happened is that authentication failed, then we should probably ask the user to log in again to that account.

The user will get a "snack bar"-style popup saying generically "Error connecting to Zulip. Retrying…". That has a "Details" button; if they hit that, they get a modal saying something like:
> ### Error
>
> Error connecting to Zulip at https://chat.zulip.org. Will retry:
>
> ZulipApiException: 401 UNAUTHORIZED getEvents: Invalid API key

(The exact message at the end, from "Invalid" on, might vary depending on the nature of the auth failure.)

So with that, it's not as clear or convenient as it could be, but the information is there. And the user can indeed always navigate away so they can add or choose another account. Moreover after we have #463 (coming soon), the user can forget the account that had the problem, in order to add it anew.

---

In short I think this issue is done according to its title: we do handle these failures. I'll file a fresh issue for the remaining task of noticing auth failures specifically and flagging those for the user.

gnprice:
> I'll file a fresh issue for the remaining task of noticing auth failures specifically and flagging those for the user.

Done:
* #1054

While writing it up I realized that registerQueue is actually the more important case than getEvents (i.e. long-poll), for reasons explained there.

PR #868, opened 2024-08-06
Show poll-failure details

### LINKED ISSUE #555: Show detailed poll-failure feedback, in beta
This is a followup to:
* #465 

We seem to be sometimes failing to poll the event queue, even after #185 and #184, for reasons that aren't yet clear; see [chat thread](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Only.20.22all.20messages.22.20updating/near/1754151). To help us pin that down, it'd be good to show some more details to the user while it's happening. So when we hit any kind of error in polling the event queue, let's show that error to the user.

A straightforward and discoverable way to do that would be with a toast. So let's start with that.

We could also subsequently build a bit more elaborate of a system where we record the message in some kind of in-app log, and then have a screen the user can navigate to that shows the error log. We don't yet have a settings screen at all, though, so there'd be some UI work to do to even have a place to put that screen (in addition to building the log UI itself).

PIG208:
Ready for review.

This fixes #555.

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -160,6 +160,18 @@
       "message": {"type": "String", "example": "Invalid format"}
     }
   },
+  "errorConnectingToServerShort": "Error connecting to Zulip. Retrying…",
+  "@errorConnectingToServerShort": {
+    "description": "Short error message for a generic unknown error connecting to the server."
+  },
+  "errorConnectingToServerDetails": "Error connecting to Zulip at {serverUrl}. Will retry:\n\n{error}",
+  "@errorConnectingToServerDetails": {
+    "description": "Dialog error message for a generic unknown error connecting to the server with details.",
+    "placeholders": {
+      "serverUrl": {"type": "String", "example": "http://example.com/"},
+      "error": {"type": "String", "example": "Invalid format"}
+    }
+  },
   "errorSharingFailed": "Sharing failed",
   "@errorSharingFailed": {
     "description": "Error message when sharing a message failed."
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -964,13 +964,28 @@ class UpdateMachine {
             debugLog('… Event queue replaced.');
             return;
 
-          case Server5xxException() || NetworkException():
+          case Server5xxException():
             assert(debugLog('Transient error polling event queue for $store: $e\n'
                 'Backing off, then will retry…'));
             maybeReportTransientError(
               localizations.errorConnectingToServerShort,
               details: localizations.errorConnectingToServerDetails(
                 serverUrl, e.toString()));
+            await (backoffMachine ??= BackoffMachine()).wait();
+            assert(debugLog('… Backoff wait complete, retrying poll.'));
+            continue;
+
+          case NetworkException():
+            assert(debugLog('Transient error polling event queue for $store: $e\n'
+                'Backing off, then will retry…'));
+            if (e.cause is! SocketException) {
+              // Heuristic check to only report interesting errors to the user.
+              // A [SocketException] is common when the app returns from sleep.
+              maybeReportTransientError(
+                localizations.errorConnectingToServerShort,
+                details: localizations.errorConnectingToServerDetails(
+                  serverUrl, e.toString()));
+            }
             await (backoffMachine ??= BackoffMachine()).wait();
             assert(debugLog('… Backoff wait complete, retrying poll.'));
             continue;
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -22,6 +22,7 @@ import '../notifications/receive.dart';
 import 'autocomplete.dart';
 import 'database.dart';
 import 'emoji.dart';
+import 'localizations.dart';
 import 'message.dart';
 import 'message_list.dart';
 import 'recent_dm_conversations.dart';
@@ -912,10 +913,28 @@ class UpdateMachine {
     }());
   }
 
+  /// This controls when we start to report transient errors to the user when
+  /// polling.
+  ///
+  /// At the 6th failure, the expected time elapsed since the first failure
+  /// will be 1.55 seocnds.
+  static const transientFailureCountNotifyThreshold = 5;
+
   void poll() async {
     assert(!_disposed);
 
     BackoffMachine? backoffMachine;
+    int accumulatedTransientFailureCount = 0;
+
+    /// This only reports transient errors after reaching
+    /// a pre-defined threshold of retries.
+    void maybeReportTransientError(String? message, {String? details}) {
+      accumulatedTransientFailureCount++;
+      if (accumulatedTransientFailureCount > transientFailureCountNotifyThreshold) {
+        reportErrorToUserBriefly(message, details: details);
+      }
+    }
+
     while (true) {
       if (_debugLoopSignal != null) {
         await _debugLoopSignal!.future;
@@ -935,6 +954,8 @@ class UpdateMachine {
         if (_disposed) return;
 
         store.isLoading = true;
+        final localizations = GlobalLocalizations.zulipLocalizations;
+        final serverUrl = store.connection.realmUrl.origin;
         switch (e) {
           case ZulipApiException(code: 'BAD_EVENT_QUEUE_ID'):
             assert(debugLog('Lost event queue for $store.  Replacing…'));
@@ -946,7 +967,10 @@ class UpdateMachine {
           case Server5xxException() || NetworkException():
             assert(debugLog('Transient error polling event queue for $store: $e\n'
                 'Backing off, then will retry…'));
-            // TODO tell user if transient polling errors persist
+            maybeReportTransientError(
+              localizations.errorConnectingToServerShort,
+              details: localizations.errorConnectingToServerDetails(
+                serverUrl, e.toString()));
             await (backoffMachine ??= BackoffMachine()).wait();
             assert(debugLog('… Backoff wait complete, retrying poll.'));
             continue;
@@ -954,7 +978,11 @@ class UpdateMachine {
           default:
             assert(debugLog('Error polling event queue for $store: $e\n'
                 'Backing off and retrying even though may be hopeless…'));
-            // TODO tell user on non-transient error in polling
+            // TODO(#186): Handle unrecoverable failures
+            reportErrorToUserBriefly(
+              localizations.errorConnectingToServerShort,
+              details: localizations.errorConnectingToServerDetails(
+                serverUrl, e.toString()));
             await (backoffMachine ??= BackoffMachine()).wait();
             assert(debugLog('… Backoff wait complete, retrying poll.'));
             continue;
@@ -978,6 +1006,9 @@ class UpdateMachine {
       // and failures, the successes themselves should space out the requests.
       backoffMachine = null;
       store.isLoading = false;
+      // Dismiss existing errors, if any.
+      reportErrorToUserBriefly(null);
+      accumulatedTransientFailureCount = 0;
 
       final events = result.events;
       for (final event in events) {
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:checks/checks.dart';
 import 'package:fake_async/fake_async.dart';
@@ -716,6 +717,24 @@ void main() {
         check(takeLastReportedError()).isNotNull().startsWith(
           "Error connecting to Zulip. Retrying…\n"
           "Error connecting to Zulip at");
+      }));
+
+      test('ignore boring errors', () => awaitFakeAsync((async) async {
+        await prepare();
+
+        for (int i = 0; i < UpdateMachine.transientFailureCountNotifyThreshold; i++) {
+          connection.prepare(exception: const SocketException('failed'));
+          pollAndFail(async);
+          check(takeLastReportedError()).isNull();
+          // This skips the pending polling backoff.
+          async.flushTimers();
+        }
+
+        connection.prepare(exception: const SocketException('failed'));
+        pollAndFail(async);
+        // Normally we start showing user visible error messages for transient
+        // errors after enough number of retries.
+        check(takeLastReportedError()).isNull();
       }));
     });
   });
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:checks/checks.dart';
+import 'package:fake_async/fake_async.dart';
 import 'package:flutter/foundation.dart';
 import 'package:http/http.dart' as http;
 import 'package:test/scaffolding.dart';
@@ -12,6 +13,7 @@ import 'package:zulip/api/route/messages.dart';
 import 'package:zulip/api/route/realm.dart';
 import 'package:zulip/model/message_list.dart';
 import 'package:zulip/model/narrow.dart';
+import 'package:zulip/log.dart';
 import 'package:zulip/model/store.dart';
 import 'package:zulip/notifications/receive.dart';
 
@@ -651,6 +653,70 @@ void main() {
 
     test('retries on MalformedServerResponseException', () {
       checkRetry(() => connection.prepare(httpStatus: 200, body: 'nonsense'));
+    });
+
+    group('report error', () {
+      String? lastReportedError;
+      String? takeLastReportedError() {
+        final result = lastReportedError;
+        lastReportedError = null;
+        return result;
+      }
+
+      /// This is an alternative to [ZulipApp]'s implementation of
+      /// [reportErrorToUserBriefly] for testing.
+      Future<void> logAndReportErrorToUserBriefly(String? message, {
+        String? details,
+      }) async {
+        if (message == null) return;
+        lastReportedError = '$message\n$details';
+      }
+
+      Future<void> prepare() async {
+        reportErrorToUserBriefly = logAndReportErrorToUserBriefly;
+        addTearDown(() => reportErrorToUserBriefly = defaultReportErrorToUserBriefly);
+
+        await prepareStore(lastEventId: 1);
+        updateMachine.debugPauseLoop();
+        updateMachine.poll();
+      }
+
+      void pollAndFail(FakeAsync async) {
+        updateMachine.debugAdvanceLoop();
+        async.elapse(Duration.zero);
+        checkLastRequest(lastEventId: 1);
+        check(store).isLoading.isTrue();
+      }
+
+      test('report non-transient errors', () => awaitFakeAsync((async) async {
+        await prepare();
+
+        connection.prepare(httpStatus: 400, json: {
+          'result': 'error', 'code': 'BAD_REQUEST', 'msg': 'Bad request'});
+        pollAndFail(async);
+        check(takeLastReportedError()).isNotNull().startsWith(
+          "Error connecting to Zulip. Retrying…\n"
+          "Error connecting to Zulip at");
+      }));
+
+      test('report transient errors', () => awaitFakeAsync((async) async {
+        await prepare();
+
+        // There should be no user visible error messages during these retries.
+        for (int i = 0; i < UpdateMachine.transientFailureCountNotifyThreshold; i++) {
+          connection.prepare(httpStatus: 500, body: 'splat');
+          pollAndFail(async);
+          check(takeLastReportedError()).isNull();
+          // This skips the pending polling backoff.
+          async.flushTimers();
+        }
+
+        connection.prepare(httpStatus: 500, body: 'splat');
+        pollAndFail(async);
+        check(takeLastReportedError()).isNotNull().startsWith(
+          "Error connecting to Zulip. Retrying…\n"
+          "Error connecting to Zulip at");
+      }));
     });
   });
 


PIG208:
A question to be answered: how do we determine if an error is transient?

Currently we handle 3 classes of errors in `UpdateMachine`:
1. `ZulipApiException`: The event queue is lost. This is commonly caused by inactivity.
1. `Server5xxException`, `NetworkException`: A server-side bug, or hiccups due to unstable Internet connection.
1. default: Could be anything that happens in the client code during the `getEvents` call.

I think it is usually safe to assume that 1 is recoverable, while we are less certain about that for 2 and 3. An idea is to define a number of retries for the client to start displaying error messages. And then in a follow-up PR the client can try to do something different (which is #186).

PIG208:
For non-transient errors, we will probably show a modal instead.

gnprice:
If there's a `ZulipApiException` *other* than the event queue being gone, then that's likely non-transient — it's either a bug in the client, or some kind of unexpected incompatibility between server and client. A retry is likely to just hit the same error again.

(Those fall into the `default` case.)

gnprice:
Other than that point, the description above all looks right.

lib/model/store.dart
```diff
@@ -777,22 +790,27 @@ class UpdateMachine {
           case Server5xxException() || NetworkException():
             assert(debugLog('Transient error polling event queue for $store: $e\n'
                 'Backing off, then will retry…'));
-            // TODO tell user if transient polling errors persist
             // TODO reset to short backoff eventually
+            accumlatedTransientFailureCount++;
+            if (accumlatedTransientFailureCount > transientFailureCountNotifyThreshold) {
+              reportErrorToUserInDialog('Failed to reach server. Will retry: $e');
```
gnprice, line 796: Let's leave this case as a toast. It's a server 5xx or a network exception, so it's unlikely there's anything the client should be doing differently about it — no need to get in the user's way and potentially annoy them.

gnprice: Then let's start showing the toasts once it's been a few seconds since we started retrying. The simple way to do that is to make the count threshold about 5, rather than 10 — at the 6th failure, it's been an expected 1.55s since the first failure, and we'll be backing off for an expected 1.6s before the next retry.

gnprice (commented):
(Quick comments on the high-level behavior.)

lib/log.dart
```diff
@@ -30,3 +33,45 @@ bool debugLog(String message) {
   }());
   return true;
 }
+
+/// Display an error message.
```
chrisbobbe, line 37: How about "Display an error message in a [SnackBar]." I like how the summary line on `reportErrorToUserInDialog`'s dartdoc says how the error is displayed.

lib/log.dart
```diff
@@ -30,3 +33,45 @@ bool debugLog(String message) {
...
+///
+/// This shows a [SnackBar] containing the message after app startup,
+/// otherwise logs it to the console.
```
chrisbobbe, line 40: Can we be more specific about "after app startup"? I can imagine confused readers wondering how the snackbar could be shown at any time other than after the app starts up.

lib/log.dart
```diff
@@ -30,3 +33,45 @@ bool debugLog(String message) {
...
+///
+/// See also: [ZulipApp._reportErrorToUserBriefly]
```
chrisbobbe, line 42: This is a private method, and also kind of feels like an implementation detail. Is this the kind of information that callers need, to understand if `reportErrorToUserBriefly` is right for them, and what its expectations/guarantees are like? (Similarly with the dartdoc on `reportErrorToUserInDialog`.)

lib/log.dart
```diff
@@ -30,3 +33,45 @@ bool debugLog(String message) {
...
+void Function(String message) get reportErrorToUserBriefly => (message) {
```
chrisbobbe, line 43: How about `reportErrorToUserInSnackbar`, to be more informative, and more parallel with `reportErrorToUserInDialog`?

lib/log.dart
```diff
@@ -30,3 +33,45 @@ bool debugLog(String message) {
...
+void Function(String message) get reportErrorToUserInDialog => (message) {
+  _debugLastReportedError = message;
+  _reportErrorToUserInDialog(message);
+};
+void Function(String message) _reportErrorToUserInDialog = _defaultReportErrorToUser;
+set reportErrorToUserInDialog(void Function(String message) value) => _reportErrorToUserInDialog = value;
+
+String? get debugLastReportedError => _debugLastReportedError;
+String? _debugLastReportedError;
+String? takeLastReportedError() {
```
chrisbobbe, line 65: Should this be `debugTakeLastReportedError`?

test/widgets/app_test.dart
```diff
@@ -166,5 +167,49 @@ void main() {
...
+      check(ZulipApp.ready).value.isFalse();
+      await tester.pump();
+      check(takeLastReportedError()).equals('test error message');
+      check(finder.evaluate()).isEmpty();
+
+      check(ZulipApp.ready).value.isTrue();
+      // After app startup, reportErrorToUserInDialog displays a dialog.
+      reportErrorToUserInDialog('test error message');
+      await tester.pump();
+      check(finder.evaluate()).single;
```
chrisbobbe, line 212: Could maybe be strengthened with `checkErrorDialog`; just a thought.

test/model/store_test.dart
```diff
@@ -427,53 +427,53 @@ void main() {
...
+          check(updateMachine.lastEventId).equals(2);
+          check(store).isLoading.isFalse();
+        });
       });
     }
 
-    test('retries on Server5xxException', () {
-      checkRetry(() => connection.prepare(httpStatus: 500, body: 'splat'));
-    });
+    group('retries', () {
```
chrisbobbe, line 482:
```
store test [nfc]: Make a group for retry tests.
```

Let's leave off the `[nfc]` in the commit message: https://github.com/zulip/zulip-flutter/pull/870#discussion_r1717639672

lib/model/store.dart
```diff
@@ -801,6 +801,7 @@ class UpdateMachine {
             assert(debugLog('Error polling event queue for $store: $e\n'
                 'Backing off and retrying even though may be hopeless…'));
             // TODO tell user on non-transient error in polling
+            reportErrorToUserInDialog('Error loading server data. Will retry: $e');
```
chrisbobbe, line 810:
```
store: Reoprt non-transient polling errors.
```

commit-message nit: spelling

chrisbobbe: Also this TODO can be removed:

```dart
            // TODO tell user on non-transient error in polling
```

chrisbobbe: Also, can this be a translated string? If that's not straightforward, let's include a `TODO(i18n)`.

chrisbobbe: Hmm. And for non-transient errors, we expect that retrying won't help, right? It seems like we should have a TODO to not retry. I'm also wondering if retrying _and_ showing a dialog, for a task that's likely to be hopeless, will spam the user with the same dialog repetitively.

chrisbobbe (commented):
Thanks! Comments below.

Then one other comment is, for all the new poll-failure feedback we're giving the user, let's identify which server it's about. This is particularly relevant when the issue is a network problem, but applies whenever the feedback can be significantly delayed. The user might have given up waiting for this account, like a minute or more ago, and is doing other things—maybe even in a different account—when the feedback appears.

test/widgets/app_test.dart
PIG208: Could be a bit repetitive, but it should be good to also check the title.

lib/log.dart
PIG208: Yeah that sounds like a good choice.

lib/log.dart
PIG208: The primary goal of including this line is to have a back link to where the setters get used, since it is not obvious. I think it would be more appropriate to exclude that bit from the dart doct, though, considering the point that it's some implementation detail.

lib/model/store.dart
PIG208: > Hmm. And for non-transient errors, we expect that retrying won't help, right? It seems like we should have a TODO to not retry. I'm also wondering if retrying and showing a dialog, for a task that's likely to be hopeless, will spam the user with the same dialog repetitively.

Yes, that's #186. Added a line mentioning that.

PIG208:
The PR has been updated. Thanks for the review!

I'm not entirely sure about the quality of the translation strings. These messages will appear more technical at the current stage for the ease of debugging. However, polishing user facing errors might be a post-launch task.

PIG208:
Updated the PR to avoid wrapping `test` calls in `checkRetry`.

lib/log.dart
```diff
@@ -1,3 +1,6 @@
+import 'package:flutter/material.dart';
+
+import 'widgets/app.dart';
```
gnprice, line 3: This `lib/log.dart` file is one that virtually the whole app imports, including parts of `lib/api/`. So I'd like to keep it from depending on the app's widgets.

That's the motivation for the indirection found in my draft branch that IIRC you started from:
https://github.com/zulip/zulip-flutter/compare/main...gnprice:zulip-flutter:dev-retry#diff-055b57342eabea1b640e04ca24eb50bc75d0c61383b928ad7211e37159a6c57bR34-R42
https://github.com/zulip/zulip-flutter/commit/68682403587804ab6dd392491fd7700596972ea3#diff-4c278608930a95a1cba05bce5a0f7d8f43bae364ed1e722eca5d8c1c4b29b57aR94-R101

where this file just has a mutable callback that defaults to something simple, and then ZulipApp can set the callback to something more complex after it starts up.

(In fact this file doesn't end up using these imports, right? It still has that indirection. I suspect the reason the linter hasn't complained about them is just the references in dartdoc.)

PIG208: > In fact this file doesn't end up using these imports, right?

Right, these imports were included only for the dart doc references. Removing them and updating the commit message to reflect the motivation:

```diff
-    The imports in `lib/log.dart` are for the code references in
-    the reportErrorToUser* helpers' dart doc.
-
-    The motivation of having these set of helpers, rather than using
-    showErrorDialog and showSnackBar directly, is that we make displaying
-    errors more accessible in non-widget code.
+    The motivation of having this indirection, rather than using
+    `showErrorDialog` and `showSnackBar` directly, is to keep dependencies
+    of `lib/log.dart` from relying on widget code.    
```

PIG208:
Updated the PR to adjust comments on `lib/log.dart` and made dialog title a parameter. I also tweaked the formatting and made some other minor improvements.

PIG208:
This update modifies the translation strings to include the realm URL in error messages.

gnprice:
That fix LGTM on a skim, thanks. I'll leave @chrisbobbe to do the next round of full review.

lib/log.dart
```diff
@@ -30,3 +30,48 @@ bool debugLog(String message) {
   }());
   return true;
 }
+
+/// Display an error message in a [SnackBar].
+///
+/// This shows a [SnackBar] containing the message after [ZulipApp] becomes ready,
+/// otherwise logs it to the console.
```
chrisbobbe, line 37:
```suggestion
/// This shows a [SnackBar] containing the message if [ZulipApp] is ready,
/// otherwise logs it to the console.
```

(similarly with `reportErrorToUserInDialog` below)

lib/log.dart
```diff
@@ -30,3 +30,48 @@ bool debugLog(String message) {
...
+// This and `reportErrorToUserInDiaglog` get set in [ZulipApp].  We need this
```
chrisbobbe, line 38:
```suggestion
// This and `reportErrorToUserInDialog` get set in [ZulipApp].  We need this
```

assets/l10n/app_en.arb
```diff
@@ -160,6 +160,31 @@
...
+  },
+  "errorConnectingToServer": "Failed to reach server at {url}. Will retry:\n\n{error}",
+  "@errorConnectingToServer": {
+    "description": "Dialog message when the the server cannot be reached",
+    "placeholders": {
+      "url": {"type": "String", "example": "http://example.com/"},
+      "error": {"type": "String", "example": "Network request failed: HTTP status 500"}
+    }
+  },
+  "errorReconnectingToServer": "Reconnecting to server at {url}...",
```
chrisbobbe, line 183: How about "Reconnecting to {url}…"? It's a bit shorter, and also I think we prefer "…" over "...".

lib/widgets/app.dart
```diff
@@ -63,22 +65,52 @@ class ZulipApp extends StatefulWidget {
...
   /// Useful in tests.
   final List<NavigatorObserver>? navigatorObservers;
 
+  static void _reportErrorToUserInSnackBar(String message) {
+    scaffoldMessenger?.showSnackBar(SnackBar(content: Text(message)));
+  }
+
+  static void _reportErrorToUserInDialog(String title, String message) {
+    final context = navigatorKey.currentContext;
+    if (context == null) return;
```
chrisbobbe, line 105: I wonder when this might happen, and whether we want to log to the console in this case, in case there's a developer watching. (Like we do for the case where `ZulipApp` isn't ready.)

assets/l10n/app_en.arb
```diff
@@ -160,6 +160,31 @@
...
+  "@errorLoadingServerData": {
+    "description": "Dialog message for a generic unknown error from the server.",
+    "placeholders": {
+      "url": {"type": "String", "example": "http://example.com/"},
+      "error": {"type": "String", "example": "Invalid format"}
+    }
+  },
+  "errorConnectingToServer": "Failed to reach server at {url}. Will retry:\n\n{error}",
+  "@errorConnectingToServer": {
+    "description": "Dialog message when the the server cannot be reached",
```
chrisbobbe, line 177:
```suggestion
    "description": "Dialog message when the server cannot be reached",
```

chrisbobbe (commented):

assets/l10n/app_en.arb
PIG208: Yeah, we can do a tiny sweep to replace "..." with "…"  in our translation strings. I saw the character used in tests and comments, but not sure why we didn't do that in `app_en.arb`.

lib/widgets/app.dart
PIG208: Hmm, good question. The guarantee we enforce is that when `_ready` is `true`, `currentContext` is always non-null (see `ZulipApp._declareReady`), so it would be a bug if `context` is `null` here. We should use a `!` here and in the other method instead. I also added an assertion `assert(_ready.value)` just to make the invariant explicit.

assets/l10n/app_en.arb
PIG208: Opened #911.

PIG208:
Thanks for the comments. This should be ready for review again!

gnprice:
@PIG208 @chrisbobbe I recommend installing this PR version on your main devices for a while, with `flutter run --release`. I just installed it yesterday, and the output is interesting but can definitely use some tweaking to be less noisy (particularly with dialogs).

PIG208:
Regarding the noisiness of the error dialog, I think a better UX is to present a snackbar with a brief error summary and a "Show more" button that brings up the dialog page with the full error message.

PIG208:
Implemented the proposed change in a squash! commit. Before moving on to the next review, I would like to gather some feedback on this new UX ([discussion](https://chat.zulip.org/#narrow/stream/48-mobile/topic/error.20reporting/near/1931665))

lib/model/store.dart
```diff
@@ -786,9 +789,12 @@ class UpdateMachine {
...
       } catch (e) {
         store.isLoading = true;
+        final localizations = GlobalLocalizations.zulipLocalizations;
+        final serverUrl = store.connection.realmUrl.origin;
         switch (e) {
           case ZulipApiException(code: 'BAD_EVENT_QUEUE_ID'):
             assert(debugLog('Lost event queue for $store.  Replacing…'));
+            reportErrorToUser(localizations.errorReconnectingToServer(serverUrl));
             await store._globalStore._reloadPerAccount(store.accountId);
             dispose();
```
chrisbobbe, line 799: How about hiding the snackbar as soon as we have a new event queue?

```suggestion
            reportErrorToUser(localizations.errorReconnectingToServer(serverUrl));
            await store._globalStore._reloadPerAccount(store.accountId);
            ZulipApp.scaffoldMessenger?.hideCurrentSnackBar();
            dispose();
```

(Or maybe `reportErrorToUser` can return a callback, or return the `ScaffoldFeatureController` returned by `ScaffoldMessengerState.showSnackBar`, that we can use to hide the snackbar if one was shown.)

In my manual testing, the reconnection to CZO is finished before the snackbar disappears, but the snackbar is still present and holding my attention for some time, making it seem (I think) like the reconnection is taking longer than it actually takes. Also, it looks like maybe `ScaffoldMessengerState.showSnackBar` lets us make the exit animation duration shorter than the entrance animation, with its `snackBarAnimationStyle` param, and that might also be something we could do to help make the app feel snappy and responsive.

chrisbobbe (commented):
Thanks! Small comments below, and I'm continuing to test this out manually on my device.

lib/model/store.dart
PIG208: These all sound like great ideas, thanks.

chrisbobbe:
Thanks! This revision LGTM. One thought is that it would be nice if the snackbar API gave us a way to identify a specific snackbar to remove, so we could be sure we're removing the one we intend to. But it looks like our choices are the "current" snackbar and all snackbars. The "current" snackbar I guess means the latest one from the queue, if it hasn't expired, and that could be right or wrong depending on the durations of the snackbars and how much time has passed.

PIG208:
Yeah, I figured that tracking if the last snack bar has expired would reduce collateral dismissal of unrelared snackbars in most cases.

PIG208:
Updated the PR to add a comment on the https://github.com/zulip/zulip-flutter/pull/868#discussion_r1733560511 implementation.

PIG208:
Pushed to rebase.

chrisbobbe:
Thanks, LGTM! Over to Greg for his review.

PIG208:
Added a commit removing a stale TODO, and squashed the two of the error reporting commits (for transient and non-transient errors) together.

gnprice:
> The public message history from CZO can be downloaded from [this protected dropbox link](https://www.dropbox.com/scl/fi/5ugemicgz6lwrsxscqvwk/corpuses.zip?rlkey=xdje3piu5scypt3glc0n4lbrz&st=muzlmp6t&dl=0). We probably will think of a better way of reusing this in the future.

(Was this comment intended for #917?)

lib/log.dart
```diff
@@ -30,3 +32,51 @@ bool debugLog(String message) {
...
+ReportErrorCallback get reportErrorToUser => (message, {details}) {
+  assert(() {
+    _debugLastReportedError = details != null ? '$message\n$details' : message;
+    return true;
+  }());
+  (_reportErrorToUser ??= _defaultReportErrorToUser)(message, details: details);
+};
+ReportErrorCallback? _reportErrorToUser;
+set reportErrorToUser(ReportErrorCallback? value) =>
+  _reportErrorToUser = value;
```
gnprice, line 59: nit: use block body

Generally a setter always gets a block body.

See https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#consider-using--for-short-functions-and-methods

lib/log.dart
```diff
@@ -30,3 +32,51 @@ bool debugLog(String message) {
...
+  if (details != null) assert(debugLog(details));
+}
+
+/// A record containing the error message and error details from the last
+/// [reportErrorToUser] call.
+@visibleForTesting
+String? get debugLastReportedError => _debugLastReportedError;
+String? _debugLastReportedError;
+@visibleForTesting
+String? debugTakeLastReportedError() {
```
gnprice, line 78: Just the "take" version should be enough.

lib/log.dart
```diff
@@ -30,3 +32,51 @@ bool debugLog(String message) {
...
+// This gets set in [ZulipApp].  We need this indirection to keep `lib/log.dart`
+// from importing widget code, because the file is a dependency for the rest of
+// the app.
+ReportErrorCallback get reportErrorToUser => (message, {details}) {
+  assert(() {
+    _debugLastReportedError = details != null ? '$message\n$details' : message;
+    return true;
+  }());
+  (_reportErrorToUser ??= _defaultReportErrorToUser)(message, details: details);
+};
```
gnprice, line 56: This structure with a getter, that returns a fixed function expression, that mostly delegates to a function-typed field, that gets lazily initialized to a default value, feels a bit hard to understand.

It seems like the reason the getter is here, vs. just a field, is to inject this recording of `_debugLastReportedError`, to support `debugTakeLastReportedError`. Can we have the tests just look for the snack bar instead?

That'd make the tests more end-to-end, testing the actual user-facing behavior, which is good:

And it seems like it'd simplify this code, too.

assets/l10n/app_en.arb
```diff
@@ -160,6 +160,41 @@
       "message": {"type": "String", "example": "Invalid format"}
     }
   },
+  "errorLoadingServerDataShort": "Error loading server data from {url}",
```
gnprice, line 163: Let's try this wording:
```suggestion
  "errorLoadingServerDataShort": "Error connecting to Zulip. Retrying…",
```

and for the details:

`Error connecting to Zulip at {serverUrl}. Will retry:\n\n{error}`

Then use that for both the `Server5xxException() || NetworkException()` case and the default case. Both cases are about "an error loading server data", or "an error connecting to Zulip", anyway.

test/model/store_test.dart
```diff
@@ -427,53 +431,80 @@ void main() {
...
-        async.flushTimers();
-        checkLastRequest(lastEventId: 1);
-        check(updateMachine.lastEventId).equals(2);
-        check(store).isLoading.isFalse();
-      });
-    }
+    group('retries on errors', () {
+      void checkRetry(void Function() prepareError, {
+        required String errorMessage,
+        int expectedFailureCountNofityThreshold = 0,
```
gnprice, line 437:
```suggestion
        int expectedFailureCountNotifyThreshold = 0,
```

gnprice (commented):

Some comments below. I've also just installed this PR branch on my main device, so I can see how this version of the UX works out over time.

lib/log.dart
PIG208: It will indeed be a helpful refactor to simplify the entire thing to a single line.

This is mostly used for `test/model/store_test.dart` only, and we test it end-to-end in `test/widgets/app_test.dart`. I think we don't instantiate `TestZulipApp` for tests in `test/model`, but this can be easier resolved by assigning a custom `reportErrorToUserBriefly` that updates a variable like `checkNotified` does.

lib/log.dart
```diff
@@ -30,3 +32,51 @@ bool debugLog(String message) {
...
+///
+/// If `message` is null, this will clear the existing [SnackBar]s if there
+/// are any.  Useful for promptly dismissing errors.
+///
+/// If `details` is non-null, the [SnackBar] will contain a button that would
+/// open a dialog containing the error details.
+// This gets set in [ZulipApp].  We need this indirection to keep `lib/log.dart`
+// from importing widget code, because the file is a dependency for the rest of
+// the app.
+ReportErrorCallback get reportErrorToUser => (message, {details}) {
```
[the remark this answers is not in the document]
PIG208: Yeah, that makes sense. I previously created two functions `reportErrorToUserBriefly` (later renamed to `reportErrorToUserInSnackbar`), `reportErrorToUserInDialog`.

In this revision, these two functions were merged into one, losing the suffix, but technically, since it displays a snackbar either way, and only shows a dialog if you click the button on the snackbar, it should be still considered "brief".

PIG208:

assets/l10n/app_en.arb
```diff
@@ -160,6 +160,23 @@
       "message": {"type": "String", "example": "Invalid format"}
     }
   },
+  "errorLoadingServerDataShort": "Error connecting to Zulip. Retrying…",
```
gnprice, line 163:
```suggestion
  "errorConnectingToServerShort": "Error connecting to Zulip. Retrying…",
```

(didn't update that part in my previous suggestion, oops)

test/model/store_test.dart
```diff
@@ -391,6 +392,22 @@ void main() {
...
+    }
+
+    void logAndReportErrorToUserBriefly(String? message, {String? details}) {
+      if (message == null) return;
+      lastReportedError = '$message\n$details';
+    }
+    // This overrides the function globally, but we don't need to worry about
+    // leaking the state because that is only relevant to widget tests.
+    // See `lib/log.dart ` for details.
+    reportErrorToUserBriefly = logAndReportErrorToUserBriefly;
```
gnprice, line 409: Better to just systematically avoid leaking state between tests.

In general the way to do that is to (a) set it up from within a test case, and then (b) use `addTearDown` to ensure it gets reset when the test case is done. Given the function definition above, I think code like
```dart
  test('some test', () {
    reportErrorToUserBriefly = logAndReportErrorToUserBriefly;
    addTearDown(reportErrorToUserBriefly = defaultReportErrorToUserBriefly);
    // … rest of test …
```

should do it.

When the setup is more complicated than that, that's the sort of thing that we write helper functions named like "prepareFoo" for.

gnprice (commented):
Thanks!

The changes in the revision look good, modulo a couple of comments below.

I've also updated the build on my device, so I'll see how the UX works out.

Would you take the reportErrorToUserBriefly commit and the one it depends on, and send those as a separate PR (which this one can then stack on top of)? That way we can get those changes merged very soon, so they can be used in other features.

I'll take a pass over the commit messages but otherwise I think those two commits are basically ready. For the later commits that actually start giving the user feedback, I'll want a bit of bake time seeing the UX.

gnprice:
Quick observation on the UX:

I reopened the app after a while. It showed the loading indicator at top for a while, then successfully reconnected.

But then for some time after that, it showed a chain of several rounds of snack bars saying there was an error, followed by one saying "reconnecting". Each one lasts several seconds, and there were about three of them that started after the loading indicator finished.

I hit the "details" link on a couple of the errors. Both were NetworkException, saying the host name chat.zulip.org couldn't be resolved.

PIG208:
Opened #935 for the two commits implementing the report error feature,
and #936 for the other nfc commit.

PIG208:
Updated the PR to ignore `NetworkException` where the underlying cause is a `SocketException`, added some more documentation and cleaned up some diffs to make them easier to read. Feel free to install a build of this PR to test it out!

gnprice:
Thanks for the update! Installing on my phone now.

test/model/store_test.dart
```diff
@@ -455,53 +479,110 @@ void main() {
...
+              // result in an error message.
+              check(takeLastReportedError()).isNull();
+              continue;
+            }
+            if (i < numFailedRequests - 1) {
+              // The error message should not appear until the `updateMachine`
+              // has retried the given number of times.
+              check(takeLastReportedError()).isNull();
+              continue;
+            }
```
gnprice, line 535: The logic in this test (or set of tests) gets fairly tangly to follow.

I think the main thing is that it'd be helpful to test two aspects separately: there's the existing test cases that just check that we retry, and let's have separate new test cases that just check the error-reporting.

Then I think both sets of tests can be a lot simpler than this — most of the logic is only relevant for one aspect or the other. It's OK that some other fragments of code will be duplicated between them.

gnprice: Separately and more tactically, I think it'd help to pull the last iteration here outside of the loop — have the loop only concern itself with the first N-1 requests. That would avoid these multiple conditionals about whether it's the last iteration.

lib/model/store.dart
```diff
@@ -811,26 +829,57 @@ class UpdateMachine {
...
-            // TODO tell user if transient polling errors persist
+            if (e.cause is! SocketException) {
+              // Heuristic check to only report interesting errors to the user.
+              // This currently ignores [SocketException], which typically
+              // occurs when the client goes back from sleep.
+              // TODO: Investigate if there are other cases of [SocketException]
+              //   that might turn out to be interesting.
+              //
+              // See also:
+              //  * [NetworkException.cause], which is the underlying exception.
```
gnprice, line 865:
```suggestion
            if (e.cause is! SocketException) {
              // Heuristic check to only report interesting errors to the user.
              // A [SocketException] is common when the app returns from sleep.
```

I think the TODO isn't actionable.

The "See also" isn't needed — `.cause` is just above, and the reader can follow that to its definition using an IDE.

Then while at it I tightened the sentence about SocketException. No need to say what the code does, as the code here speaks for itself; only to explain why we chose to make the code do that.

gnprice (commented):
OK, I've observed this on my phone for a while; and it's been pretty quiet, which is good.

After seeing it in action, I think we can leave out the last commit, with the "Reconnecting to {serverUrl}…" message. It feels redundant with the loading indicator at the top; and the fact that the toast disappears on its own schedule while the loading indicator can be still going seems incongruous, and likely to be confusing.

Otherwise this generally all looks good; just a few comments below.

(I also feel like the `UpdateMachine.poll` method has gotten kind of unwieldy; maybe I'll take a stab at reorganizing that after merging this PR. It's already kind of unwieldy in main.)

PIG208:
Thanks for the feedback! Dropped the last commit and previous changes relevant to the tests. A new group of tests is now responsible for testing the error reporting feature.

test/model/store_test.dart
```diff
@@ -652,6 +655,90 @@ void main() {
     test('retries on MalformedServerResponseException', () {
       checkRetry(() => connection.prepare(httpStatus: 200, body: 'nonsense'));
     });
+
+    group('reportErrorToUserBriefly', () {
```
gnprice, line 659: nit:
```suggestion
    group('error reporting', () {
```

Otherwise it sounds like these are tests of `reportErrorToUserBriefly`.

test/model/store_test.dart
```diff
@@ -652,6 +655,90 @@ void main() {
...
+        await prepareStore(lastEventId: 1);
+        updateMachine.debugPauseLoop();
+        updateMachine.poll();
+      }
+
+      late void Function() prepareError;
+
+      void pollAndFail(FakeAsync async) {
+        prepareError();
+        updateMachine.debugAdvanceLoop();
```
gnprice, line 689:
```suggestion
      void pollAndFail(FakeAsync async) {
        updateMachine.debugAdvanceLoop();
```

Instead the callers can say `connection.prepare(…)` (i.e., the body of what they would set `prepareError` to) just before each call to `pollAndFail`. That isn't really any longer — it makes a very slight duplication of the content of those `connection.prepare` calls — and it makes the control flow substantially more straightforward.

test/model/store_test.dart
```diff
@@ -652,6 +655,90 @@ void main() {
...
+      test('report transient errors', () => awaitFakeAsync((async) async {
+        prepareError = () => connection.prepare(httpStatus: 500, body: 'splat');
+        await prepare();
+
+        // There should be no user visible error messages during these retries.
+        for (int i = 0; i < UpdateMachine.transientFailureCountNotifyThreshold; i++) {
+          pollAndFail(async);
+          check(takeLastReportedError()).isNull();
+          // This would skip the pending polling backoff.
+          async.flushTimers();
```
gnprice, line 715:
```suggestion
          // This skips the pending polling backoff.
          async.flushTimers();
```

I.e. this does skip that backoff, right?

Or am I missing a subtlety where it doesn't actually do so, but only *would* do so if something were different?

gnprice (commented):
Thanks for the revision! Just small comments below, all in the tests.

And as a note to myself, there's a followup I mentioned at https://github.com/zulip/zulip-flutter/pull/868#pullrequestreview-2376436250 I might do after merge.

PIG208:
Thanks! Ready for review now.

gnprice:
Looks good — merging.

Issue #251, opened 2023-08-03
msglist: Offer starred-messages narrow


gnprice:
Like the "Starred messages" narrow the web app offers at the top of the left sidebar.

We should support this narrow in the message list, and then provide a way to navigate to it.

If we do this before building a helpful overall navigation for the app, the way to navigate to it can be just another button on HomePage.

This is a sort of followup to #72.

PIG208:
Completed via #870.

PR #869, opened 2024-08-06
Internal_link: Add ApiIsNarrow.

PIG208:
Prep work for #251.

### DIFF
--- a/lib/api/model/narrow.dart
+++ b/lib/api/model/narrow.dart
@@ -1,3 +1,7 @@
+import 'package:json_annotation/json_annotation.dart';
+
+part 'narrow.g.dart';
+
 typedef ApiNarrow = List<ApiNarrowElement>;
 
 /// Resolve any [ApiNarrowDm] elements appropriately.
@@ -114,19 +118,44 @@ class ApiNarrowPmWith extends ApiNarrowDm {
   ApiNarrowPmWith._(super.operand, {super.negated});
 }
 
-// TODO: generalize into ApiNarrowIs
-class ApiNarrowIsMentioned extends ApiNarrowElement {
+class ApiNarrowIs extends ApiNarrowElement {
   @override String get operator => 'is';
-  @override String get operand => 'mentioned';
 
-  ApiNarrowIsMentioned({super.negated});
+  @override final IsOperand operand;
+
+  ApiNarrowIs(this.operand, {super.negated});
+
+  factory ApiNarrowIs.fromJson(Map<String, dynamic> json) => ApiNarrowIs(
+    IsOperand.fromRawString(json['operand'] as String),
+    negated: json['negated'] as bool? ?? false,
+  );
 }
 
-class ApiNarrowIsUnread extends ApiNarrowElement {
-  @override String get operator => 'is';
-  @override String get operand => 'unread';
+/// An operand value of "is" operator.
+///
+/// See also:
+///   - https://zulip.com/api/construct-narrow
+///   - https://zulip.com/help/search-for-messages#search-your-important-messages
+///   - https://zulip.com/help/search-for-messages#search-by-message-status
+@JsonEnum(alwaysCreate: true)
+enum IsOperand {
+  dm,        // TODO(server-7) new in FL 177
+  private,   // TODO(server-7) deprecated in FL 177, equivalent to [dm].
+  alerted,
+  mentioned,
+  starred,
+  followed,  // TODO(server-9) new in FL 265
+  resolved,
+  unread,
+  unknown;
 
-  ApiNarrowIsUnread({super.negated});
+  static IsOperand fromRawString(String raw) => $enumDecode(
+    _$IsOperandEnumMap, raw, unknownValue: unknown);
+
+  @override
+  String toString() => _$IsOperandEnumMap[this]!;
+
+  String toJson() => toString();
 }
 
 class ApiNarrowMessageId extends ApiNarrowElement {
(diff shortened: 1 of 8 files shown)


PIG208:
Should be ready for review now.

chrisbobbe:
Thanks, LGTM! Over to Greg for his review.

PIG208:
Extracted a bug fix to make `Add generic support for is operands` nfc.

lib/api/model/narrow.dart
```diff
@@ -1,3 +1,5 @@
+import '../../model/internal_link.dart';
```
gnprice, line 1: Generally let's avoid importing in lib/api/ from other parts of the code — I'd like to think of that subtree as a Zulip API binding layer, such that if we wanted to we could easily split it out as its own Dart package.

This is for IsOperand, right? That's part of this API (it's part of https://zulip.com/api/construct-narrow ), so let's just put its definition here in this file.

lib/api/model/narrow.dart
```diff
@@ -114,19 +116,17 @@ class ApiNarrowPmWith extends ApiNarrowDm {
...
-  @override String get operand => 'mentioned';
 
-  ApiNarrowIsMentioned({super.negated});
-}
+  @override final String operand;
 
-class ApiNarrowIsUnread extends ApiNarrowElement {
-  @override String get operator => 'is';
-  @override String get operand => 'unread';
+  ApiNarrowIs(IsOperand operand, {super.negated}) : operand = operand.toString();
```
gnprice, line 124: I don't really understand this paragraph:

    We convert the IsOperand value into a String when constructing
    ApiNarrowIs, so that we don't need to mingle with
    ApiNarrowElement.toJson to take care of the enum.

What would happen if we had this class store the enum?

lib/model/internal_link.dart
```diff
@@ -292,3 +303,28 @@ List<int>? _parseDmOperand(String operand) {
...
+/// An operand value of "is" operator.
+///
+/// See also:
+///   - https://zulip.com/api/construct-narrow
+///   - https://zulip.com/help/search-for-messages#search-your-important-messages
+///   - https://zulip.com/help/search-for-messages#search-by-message-status
+@JsonEnum(alwaysCreate: true)
+enum IsOperand {
+  dm,
+  private, // This is deprecated in FL 177 and equivalent to [dm].
```
gnprice, line 316:
```suggestion
  dm, // TODO(server-7) new in FL 177
  private, // TODO(server-7) deprecated in FL 177, equivalent to [dm]
```

The `TODO(server-N)` format will help us spot in the future that we can drop `private`.

For the newly-introduced value, having a comment puts us on guard to not try to use it with old servers. Then we need a TODO-server comment to remind us to delete that warning comment… so it may as well be one line that does both.

lib/model/internal_link.dart
```diff
@@ -292,3 +303,28 @@ List<int>? _parseDmOperand(String operand) {
...
+  alerted,
+  mentioned,
+  starred,
+  followed,
```
gnprice, line 320: needs a TODO-server comment, similar to `dm` above

lib/model/internal_link.dart
```diff
@@ -292,3 +303,28 @@ List<int>? _parseDmOperand(String operand) {
     return null;
   }
 }
+
+/// An operand value of "is" operator.
+///
+/// See also:
+///   - https://zulip.com/api/construct-narrow
+///   - https://zulip.com/help/search-for-messages#search-your-important-messages
+///   - https://zulip.com/help/search-for-messages#search-by-message-status
```
gnprice, line 312: Cool, thanks for the thorough collection of links — this made it easy for me to verify the list of values.

lib/model/internal_link.dart
```diff
@@ -195,9 +193,22 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
     }
   }
 
-  if (isMentionedElement != null) {
+  if (isElementOperands.isNotEmpty) {
     if (streamElement != null || topicElement != null || dmElement != null) return null;
-    return const MentionsNarrow();
+    switch (isElementOperands.singleOrNull) {
```
gnprice, line 198: Let's make this more explicit:
```suggestion
    if (isElementOperands.length > 1) return null;
    switch (isElementOperands.single) {
```

I actually started writing a comment asking if the null case was impossible, because this is inside an `isNotEmpty` conditional. Then I realized there are two different cases that `singleOrNull` turns to null.

gnprice (commented):
Thanks for doing this refactor! Small comments.

lib/api/model/narrow.dart
PIG208: Testing with that again, I think I was mistaken that we need special handling for `ApiNarrowElement.toJson` to support the use of the enum field, but actually the fix is to just provide a `toJson` method on `IsOperand`.

PIG208:
Thanks for the review @gnprice. Updated the PR.

lib/model/internal_link.dart
PIG208: Somewhat related: how do we deal with deprecated API values? Do we still need to support them post server-7, until they are fully removed in a future version?

gnprice: When it comes to links, we'll want to accept the old forms forever (as will web). People will have put links to Zulip in all sorts of external places we can't update — for example in GitHub comments, for Zulip users who use GitHub like we do — and so we should keep those links working.

But once we stop supporting the old server versions that lack the new forms, we can stop ever generating the old forms. So for example ApiNarrowDm and its two subclasses will simplify back down to just ApiNarrowDm, and the only wrinkle remaining for the long term will be that we'll accept `pm-with` in links and translate that to ApiNarrowDm. We'll handle IsOperand.private similarly.

PIG208:
Thanks. This should be fixed now.

gnprice:
Thanks, looks good — merging.

Issue #818, opened 2024-07-16
Handle live updates to whether messages should appear in the @-mentions narrow


PIG208:
We should handle `UpdateMessageEvent` so that when the message becomes visible to the @-mentions narrow, it is added to the `MessageListView`.

This is a follow-up to #250.

Currently the web app does not handle this, so we are not pressed to do so.

PR #870, opened 2024-08-07
narrow: Add starred messages narrow

PIG208:
Unreads can probably be handled in a separate PR, because it will probably grow in complexity.

This is stacked on top of #869

### DIFF
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -199,10 +199,11 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
     switch (isElementOperands.single) {
       case IsOperand.mentioned:
         return const MentionsNarrow();
+      case IsOperand.starred:
+        return const StarredMessagesNarrow();
       case IsOperand.dm:
       case IsOperand.private:
       case IsOperand.alerted:
-      case IsOperand.starred:
       case IsOperand.followed:
       case IsOperand.resolved:
       case IsOperand.unread:
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -272,6 +272,12 @@ class MessageStoreImpl with MessageStore {
           view.notifyListenersIfAnyMessagePresent(event.messages);
           // TODO(#818): Support MentionsNarrow live-updates when handling
           //   @-mention flags.
+
+          // To make it easier to re-star a message, we opt-out from supporting
+          // live-updates when starred flag is removed.
+          //
+          // TODO: Support StarredMessagesNarrow live-updates when starred flag
+          //   is added.
         }
       }
     }
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -270,6 +270,8 @@ class MessageStoreImpl with MessageStore {
       if (anyMessageFound) {
         for (final view in _messageListViews) {
           view.notifyListenersIfAnyMessagePresent(event.messages);
+          // TODO(#818): Support MentionsNarrow live-updates when handling
+          //   @-mention flags.
         }
       }
     }
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -392,6 +392,12 @@ void main() {
       await setupToMessageActionSheet(tester, message: message, narrow: const MentionsNarrow());
       check(findQuoteAndReplyButton(tester)).isNull();
     });
+
+    testWidgets('not offered in StarredMessagesNarrow (composing to reply is not yet supported)', (tester) async {
+      final message = eg.streamMessage(flags: [MessageFlag.starred]);
+      await setupToMessageActionSheet(tester, message: message, narrow: const StarredMessagesNarrow());
+      check(findQuoteAndReplyButton(tester)).isNull();
+    });
   });
 
   group('MarkAsUnread', () {
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -388,7 +388,7 @@ void main() {
     });
 
     testWidgets('not offered in MentionsNarrow (composing to reply is not yet supported)', (tester) async {
-      final message = eg.streamMessage();
+      final message = eg.streamMessage(flags: [MessageFlag.mentioned]);
       await setupToMessageActionSheet(tester, message: message, narrow: const MentionsNarrow());
       check(findQuoteAndReplyButton(tester)).isNull();
     });
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -122,14 +122,19 @@ void main() {
   });
 
   group('presents message content appropriately', () {
-    // regression test for https://github.com/zulip/zulip-flutter/issues/736
-    testWidgets('content in "Combined feed" not asked to consume insets (including bottom)', (tester) async {
+    testWidgets('content not asked to consume insets (including bottom), even without compose box', (tester) async {
+      // Regression test for: https://github.com/zulip/zulip-flutter/issues/736
       const fakePadding = FakeViewPadding(left: 10, top: 10, right: 10, bottom: 10);
       tester.view.viewInsets = fakePadding;
       tester.view.padding = fakePadding;
 
       await setupMessageListPage(tester, narrow: const CombinedFeedNarrow(),
         messages: [eg.streamMessage(content: ContentExample.codeBlockPlain.html)]);
+
+      // Verify this message list lacks a compose box.
+      // (The original bug wouldn't reproduce with a compose box present.)
+      final state = MessageListPage.ancestorOf(tester.element(find.text("verb\natim")));
+      check(state.composeBoxController).isNull();
 
       final element = tester.element(find.byType(CodeBlock));
       final padding = MediaQuery.of(element).padding;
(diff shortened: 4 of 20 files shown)


PIG208:
This does not implement live updates for newly starred/unstarred messages. We should able to handle #818 together with this.

PIG208 (commented):
We will leave unreads for starred messages for now, as it is pretty uncommon to have unreads for them.

PIG208:
Skipping the first 3 commits, this should be reviewable independently from #869.

chrisbobbe:
Looks like CI is failing.

PIG208:
Should be fixed now. It was an error emerged from moved messages support across narrows.

chrisbobbe:
Thanks! Ah, but it looks like some conflicts appeared after you pushed that revision; could you resolve those please? :)

PIG208:
Thanks! Just rebased it.

lib/model/internal_link.dart
```diff
@@ -203,6 +203,7 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
       case IsOperand.private:
       case IsOperand.alerted:
       case IsOperand.starred:
+        return const StarredMessagesNarrow();
```
chrisbobbe, line 206:
```
internal_link: Parse "/is/starred" internal links.
```

This is a bug. :) Since cases before `IsOperand.starred` are empty, they'll also give `const StarredMessagesNarrow()`, just like the `IsOperand.starred` case.

https://dart.dev/language/branches#switch-statements

> Non-empty `case` clauses jump to the end of the switch after completion. They do not require a `break` statement.

It's a common pattern to want multiple cases to give the same thing, and to express that with empty `case` clauses, as here.

To fix (here and in the test, which has the same bug), I would just reorder:

```dart
    switch (isElementOperands.single) {
      case IsOperand.mentioned:
        return const MentionsNarrow();
      case IsOperand.starred:
        return const StarredMessagesNarrow();
      case IsOperand.dm:
      case IsOperand.private:
      case IsOperand.alerted:
      case IsOperand.followed:
      case IsOperand.resolved:
      case IsOperand.unread:
      case IsOperand.unknown:
        return null;
    }
```

chrisbobbe (changes requested):
Thanks! Small comments below, including one about a bug I noticed.

In manual testing, I noticed that when I'm viewing the starred-messages narrow, when I unstar a message, it remains in the list. At first I thought this was a bug, but Greg pointed out that it's actually a product decision; it lets you easily undo the unstar action if you want. So that seems fine.

—But let's put a comment in the code, clarifying that it's intentional (to offer undo), so readers don't have to wonder. This point already applies for messages with @-mentions in the mentions narrow, I think, so it would be good to mention that narrow too, in a separate, earlier commit. I think the right place to comment on these things would be in `MessageStoreImpl.handleUpdateMessageFlagsEvent`, near where it operates on the message-list views (by having them notify listeners but not remove messages).

…Hmm. Thinking more, actually…The current behavior for remove-star events does seem helpful, but wouldn't it also be helpful if `MessageStoreImpl.handleUpdateMessageFlagsEvent` acted on _add_-star events by making the message appear in the starred-narrow message list? Similarly for mentioned messages in the mentions narrow.

PIG208:
Regarding

> …Hmm. Thinking more, actually…The current behavior for remove-star events does seem helpful, but wouldn't it also be helpful if MessageStoreImpl.handleUpdateMessageFlagsEvent acted on add-star events by making the message appear in the starred-narrow message list? Similarly for mentioned messages in the mentions narrow.

Live-updates for starred messages and @-mentions are post-launch features, so we leave them out for now:

> This does not implement live updates for newly starred/unstarred messages. We should able to handle #818 together with this as a post-launch follow-up.

PIG208:
Having comments for now should be suitable.

PIG208:

chrisbobbe:
Thanks, LGTM! Over to Greg, then.

gnprice (commented):

Generally this looks great. Comments below on the tests, mostly just on their structure.

Separately, I think the last three commits make most sense squashed together:
8d8841ee9 narrow: Add starred messages narrow.
3b3b4c389 internal_link: Parse "/is/starred" internal links.
bf0759dea nav: Add a button for starred messages page.

The latter two are small changes adding pieces that really should be there in order to say we support the starred-messages narrow. It seems like the main difference between those and most of the changes in the main commit is just that the main commit has the changes that are prompted by exhaustive-switch checking once the StarredMessagesNarrow type is added. That's a very useful tool but doesn't need to decide our commit structure.

(Even in the other direction: if the changes required to satisfy exhaustive `switch`es added up to a big change that we didn't want to put in a single commit, we could split it up by making some of the new cases just `throw UnimplementedError()` or the like, with TODO comments pointing to the relevant issue.)

PIG208:

Made separate functions for one test each in action_sheet_test.dart and message_list_test.dart, and removed the extra regression test. I also squashed the later commits together as suggested.

test/widgets/message_list_test.dart
```diff
@@ -123,31 +123,25 @@ void main() {
...
-      const fakePadding = FakeViewPadding(left: 10, top: 10, right: 10, bottom: 10);
-      tester.view.viewInsets = fakePadding;
-      tester.view.padding = fakePadding;
-
-      await setupMessageListPage(tester, narrow: const CombinedFeedNarrow(),
-        messages: [eg.streamMessage(content: ContentExample.codeBlockPlain.html)]);
-
-      final element = tester.element(find.byType(CodeBlock));
-      final padding = MediaQuery.of(element).padding;
-      check(padding).equals(EdgeInsets.zero);
```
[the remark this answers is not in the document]
PIG208: Yeah, this interpretation makes sense me. I decided to remove the `MentionsNarrow` test, because I think they will be largely verifying the same code path. It is reasonable to check for the absence of compose box in different narrows, but we are more interested in testing the inset fix here.

test/widgets/message_list_test.dart
```diff
@@ -717,40 +711,34 @@ void main() {
...
       }
 
-      testWidgets('show stream name in CombinedFeedNarrow', (tester) async {
-        await setupMessageListPage(tester,
-          narrow: const CombinedFeedNarrow(),
-          messages: [message], subscriptions: [eg.subscription(stream)]);
-        await tester.pump();
-        check(findInMessageList('stream name')).length.equals(1);
-        check(findInMessageList('topic name')).length.equals(1);
-      });
```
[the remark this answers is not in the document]
PIG208: Thanks. `sticky_header_test.dart` is an interesting example. Dropped the parameterization commit.

test/widgets/action_sheet_test.dart
```diff
@@ -388,10 +388,16 @@ void main() {
     });
 
     testWidgets('not offered in MentionsNarrow (composing to reply is not yet supported)', (tester) async {
-      final message = eg.streamMessage();
+      final message = eg.streamMessage(flags: [MessageFlag.mentioned]);
       await setupToMessageActionSheet(tester, message: message, narrow: const MentionsNarrow());
```
gnprice, line 392: action_sheet test: Create message with the appropriate flag.
    
    The message wouldn't otherwise appear in MentionsNarrow.

Hmm indeed. But I'm a bit puzzled then: why wasn't this test already failing?

Ah it's because the code actually *will* show the message. So this is just a realism fix, making the simulated server response something a non-buggy server might give on this request.

lib/model/message.dart
```diff
@@ -270,6 +270,14 @@ class MessageStoreImpl with MessageStore {
       if (anyMessageFound) {
         for (final view in _messageListViews) {
           view.notifyListenersIfAnyMessagePresent(event.messages);
+          // TODO(#818): Support MentionsNarrow live-updates when handling
+          // @-mentioned flags.
```
gnprice, line 274: nit: indent any continuation lines on a TODO comment:
```suggestion
          // TODO(#818): Support MentionsNarrow live-updates when handling
          //   @-mentioned flags.
```

gnprice: (also "@-mentioned flags" reads to me as about flags that were @-mentioned; could say "mention flags" or "@-mention flags")

gnprice (commented):
Thanks for the revision!

A few new nits below — otherwise all looks good. I'll fix those up and merge.

test/widgets/message_list_test.dart
```diff
@@ -135,19 +135,6 @@ void main() {
       final padding = MediaQuery.of(element).padding;
       check(padding).equals(EdgeInsets.zero);
     });
-
-    testWidgets('content in MentionsNarrow not asked to consume insets (including bottom)', (tester) async {
```
gnprice, line 139: Oh one other commit-message comment:

    A test on CombinedFeedNarrow is sufficient for preventing this particular
    regression. We could go further and test this on other narrows, but
    there is little to gain other than verifying that those other narrows do
    not have a compose box, which is covered statically with the exhaustive
    check from hasComposeBox.

The last sentence here doesn't sound right, as this test didn't check that the narrow doesn't have a compose box. Rather, if the narrow had a compose box, the test would pass even if the bug were to return on other narrows that don't have the compose box. The original bug reproduced only on CombinedFeedNarrow (the only narrow lacking a compose box at the time), which is why the repro test used that narrow.

gnprice:
> I'll fix those up and merge.

Done. I also added a commit to better pin down that existing regression test's assumptions:
5bcb5d4ce msglist test [nfc]: Pin down lack of compose box in no-insets regression test

PR #871, opened 2024-08-08
Reset backoff duration after successful connection.

### LINKED ISSUE #554: Reset backoff interval after successful connection

PIG208:
This fixes #554, and supersedes #564.

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -807,8 +807,24 @@ class UpdateMachine {
         }
       }
 
+      // After one successful request, we reset backoff to its initial state.
+      // That way if the user is off the network and comes back on, the app
+      // doesn't wind up in a state where it's slow to recover the next time
+      // one request fails.
+      //
+      // This does mean that if the server is having trouble and handling some
+      // but not all of its requests, we'll end up doing a lot more retries than
+      // if we stayed at the max backoff interval; partway toward what would
+      // happen if we weren't backing off at all.
+      //
+      // But at least for [getEvents] requests, as here, it should be OK,
+      // because this is a long-poll.  That means a typical successful request
+      // takes a long time to come back; in fact longer than our max backoff
+      // duration (which is 10 seconds).  So if we're getting a mix of successes
+      // and failures, the successes themselves should space out the requests.
       backoffMachine = null;
       store.isLoading = false;
+
       final events = result.events;
       for (final event in events) {
         await store.handleEvent(event);
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -763,7 +763,7 @@ class UpdateMachine {
   }
 
   void poll() async {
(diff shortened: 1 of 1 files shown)


test/api/backoff_test.dart
```diff
@@ -14,27 +14,28 @@ Future<Duration> measureWait(Future<void> future) async {
...
+void verifyRandomizedDurations({
+  required List<int> expectedMaxDurationsInMs,
+  required List<Duration> Function(List<Duration>) computeActualDurations,
+}) {
     // This is a randomized test.  [numTrials] is chosen so that the failure
-    // probability < 1e-9.  There are 2 * 11 assertions, and each one has a
-    // failure probability < 1e-12; see below.
+    // probability < 1e-9 for up to 1000 assertions.
+    // There are 2 * [expectedMaxDurationsInMs.length] assertions,
+    // and each one has a failure probability < 1e-12; see below.
```
chrisbobbe, line 29: The part "for up to 1000 assertions" is new in this commit. What does it mean, and is there a logic change it corresponds to?

chrisbobbe (commented):
Thanks! One comment below, and I'll also mark this for Greg's review.

test/api/backoff_test.dart
PIG208: This commit allows an arbitrary number of expected durations to be passed, so the `1000` is here to convey how many expected values can be passed for the upper bound of 1e-9 to still hold.

PIG208:
After having a discussion with Greg, we will drop the testing part of this because this change is easy to verify, but testing is complicated.

gnprice:

Merging, with an additional commit adding a long comment:
cca6f244a store [nfc]: Explain why we reset backoff, and the trade-offs

Issue #825, opened 2024-07-19
Plugin "default implementation" warnings on latest Flutter


gnprice:
Building the app, or running `flutter test` or even just `flutter pub get`, produces the following warnings when using the latest Flutter from main — often several times over:
```
Package file_picker:linux references file_picker:linux as the default plugin, but it does not provide an
inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: linux:
default_package: file_picker` or add an inline implementation to file_picker via `platforms: linux:`
`pluginClass` or `dartPluginClass`.


Package flutter_local_notifications:linux references flutter_local_notifications_linux:linux as the default
plugin, but the package does not exist.
Ask the maintainers of flutter_local_notifications to either avoid referencing a default implementation via
`platforms: linux: default_package: flutter_local_notifications_linux` or create a plugin named
flutter_local_notifications_linux.


Package file_picker:macos references file_picker:macos as the default plugin, but it does not provide an
inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: macos:
default_package: file_picker` or add an inline implementation to file_picker via `platforms: macos:`
`pluginClass` or `dartPluginClass`.


Package file_picker:windows references file_picker:windows as the default plugin, but it does not provide an
inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms:
windows: default_package: file_picker` or add an inline implementation to file_picker via `platforms:
windows:` `pluginClass` or `dartPluginClass`.
```

It looks like the warnings were introduced in:
* https://github.com/flutter/flutter/pull/137040

Upstream issues in those plugins are:
* Already fixed, so we just need to upgrade: https://github.com/MaikuB/flutter_local_notifications/issues/2368
* Filed by me just now: https://github.com/miguelpruivo/flutter_file_picker/issues/1555

These are warnings, not errors, so don't seem to be blocking us. And they only apply to desktop platforms, not our real targets, so there should be no impact in production. The warning spew is annoying, though, so it'll be good to clear them up.

PR #872, opened 2024-08-08
deps: Upgrade Flutter to 3.24.0-1.0.pre.518

sm-sayedi:

We haven't had a Flutter upgrade for over a month now; specifically, there is the [`spacing` property](https://github.com/flutter/flutter/pull/152472) available to `Column` and `Row` now, which is [helpful](https://github.com/zulip/zulip-flutter/pull/853/files#r1710247901) in #853. So I think it's a good time to have an upgrade!

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -194,10 +194,10 @@ packages:
     dependency: "direct main"
     description:
       name: collection
-      sha256: ee67cb0715911d28db6bf4af1026078bd6f0128b07a5f66fb2ed94ec6783c09a
+      sha256: a1ace0a119f20aabc852d165077c036cd864315bd99b7eaa10a60100341941bf
       url: "https://pub.dev"
     source: hosted
-    version: "1.18.0"
+    version: "1.19.0"
   color_models:
     dependency: transitive
     description:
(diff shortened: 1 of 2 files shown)


rajveermalviya:
Another big highlight this update is — the [Impeller renderer](https://docs.flutter.dev/perf/impeller) replacing Skia renderer finally being enabled by default on Android (on supported devices, it'll still fallback to Skia at runtime where Vulkan is unsupported or doesn't provide the required capabilities).

gnprice:
Thanks for taking care of this! Looks good — merging.

It's annoying that this will start forcing the use of Flutter versions where #825 appears. That was a reason I'd been procrastinating doing this upgrade. But we can't just hang back on an old Flutter version indefinitely, so given there are concrete reasons we want the upgrade it's time to do it. We'll just tolerate the annoying warning messages until either someone else fixes the affected plugins, or we get tired of the warnings and do so ourselves.

… Oh hey, and digging a bit, it looks like they may indeed have been fixed — the upstream issue on flutter_file_picker just didn't get closed to say so. So I'll see if a quick upgrade handles that.

Issue #845, opened 2024-07-28
Include `/` after hostname in generated links


gnprice:
The "Copy link to message" option in the message action sheet produces links that look like this:
https://chat.zulip.org#narrow/stream/48-mobile/topic/Copied.20message.20links.20do.20not.20linkify/near/1904591

Note there's no `/` after the hostname, before the `#`. This is still a valid URL, I believe — but it looks funny, and more concretely it means that if you post it in a Zulip message it doesn't get linkified. We should fix it so that the link instead looks like, for example:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Copied.20message.20links.20do.20not.20linkify/near/1904591

It looks like none of our existing tests for this functionality spell out the full URL that gets produced — they're all in terms of pieces of the parsed URL, or compute the expected URL with things like `realmUrl.resolve(…)`. Those are fine sorts of tests to have in order to keep those tests focused on the details they're meant to be about; but as part of a PR fixing this issue, we should also add a few tests that include, verbatim, the full URL they expect as the output.

### Implementation

A PR to fix this issue should do two things:
* Add some tests of the `narrowLink` function, alongside its existing tests, that specify the verbatim full URL they expect.
* Adjust that function so that its results look like `https://chat.example/#narrow/…` instead of `https://chat.example#narrow/…`.

gnprice:
This issue is open to be worked on. There was a previous PR attempting it:
* #874 

but that PR didn't work out and I've closed it.

That PR thread is long, and most of what's there probably isn't useful to read before sending a fresh PR. After rereading the thread just now, I've edited the issue description above to lay out more explicitly what should go into an implementation.

This was in the "Beta 3: Summer 2024" milestone because it was actively being worked on at the end of the summer. Since the thread of work on that PR is over, I've moved it to the current milestone, "Beta 4: Fall 2024".

PR #874, opened 2024-08-09
internal_links: Add "/" before fragment identifier

satyam372:
Fixes: #845 

This adds a **'/ '**  after hostname. when this Url is posted in Zulip message it gets linkified.
Screenshot_20240809-213108 1

### DIFF
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -96,7 +96,11 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
     fragment.write('/near/$nearMessageId');
   }
 
-  return store.realmUrl.replace(fragment: fragment.toString());
+  final realmUrlWithSlash = store.realmUrl.path.isNotEmpty
+      ? store.realmUrl
+      : store.realmUrl.replace(path: '${store.realmUrl.path}/');
+
+  return realmUrlWithSlash.replace(fragment: fragment.toString());
 }
 
 /// A [Narrow] from a given URL, on `store`'s realm.
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -38,11 +38,14 @@ late FakeApiConnection connection;
 Future<void> setupToMessageActionSheet(WidgetTester tester, {
   required Message message,
   required Narrow narrow,
+  Account? account,
 }) async {
   addTearDown(testBinding.reset);
 
-  await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
-  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  account ??= eg.selfAccount;
+  final initialSnapshot = eg.initialSnapshot();
+  await testBinding.globalStore.add(account, initialSnapshot);
+  final store = await testBinding.globalStore.perAccount(account.id);
   await store.addUser(eg.user(userId: message.senderId));
   if (message is StreamMessage) {
     final stream = eg.stream(streamId: message.streamId);
@@ -61,7 +64,7 @@ Future<void> setupToMessageActionSheet(WidgetTester tester, {
     messages: [message],
   ).toJson());
 
-  await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
+  await tester.pumpWidget(TestZulipApp(accountId: account.id,
     child: MessageListPage(initNarrow: narrow)));
 
   // global store, per-account store, and message list get loaded
@@ -550,6 +553,82 @@ void main() {
       await tester.pump(Duration.zero);
       final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
+    });
+  });
+
+  group('Message link verbatim URL tests', () {
+    Future<void> tapCopyMessageLinkButton(WidgetTester tester) async {
+      await tester.ensureVisible(find.byIcon(Icons.link, skipOffstage: false));
+      await tester.tap(find.byIcon(Icons.link));
+      await tester.pump(); // [MenuItemButton.onPressed] called in a post-frame callback: flutter/flutter@e4a39fa2e
+    }
+
+    Future<void> testMessageLink({
+      required WidgetTester tester,
+      required String realmUrl,
+      required String topic,
+      required String expectedLink,
+    }) async {
+      final stream = eg.stream(streamId: 42, name: 'stream 42');
+      final message = eg.streamMessage(id: 12345, stream: stream, topic: topic);
+      final narrow = TopicNarrow.ofMessage(message);
+      final account = eg.selfAccount.copyWith(realmUrl: Uri.parse(realmUrl));
+      await setupToMessageActionSheet(
+      tester, message: message, narrow: narrow, account: account);
+
+      final store = await testBinding.globalStore.perAccount(account.id);
+      if (!store.streams.containsKey(stream.streamId)) {
+       await store.addStream(stream);
+      }
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
+    }
+
+    testWidgets('Copies link with custom realm path', (tester) async {
+      await testMessageLink(
+      tester: tester,
+      realmUrl: 'https://chat.example/foo',
+      topic: 'test topic',
+      expectedLink: 'https://chat.example/foo#narrow/stream/42-stream-42/topic/test.20topic/near/12345',
+      );
+    });
+
+    testWidgets('Copies link with realm URL without path', (tester) async {
+      await testMessageLink(
+      tester: tester,
+      realmUrl: 'https://chat.zulip.org',
+      topic: 'test topic',
+      expectedLink: 'https://chat.zulip.org/#narrow/stream/42-stream-42/topic/test.20topic/near/12345',
+      );
+    });
+
+    testWidgets('Copies link with special characters in topic', (tester) async {
+      await testMessageLink(
+      tester: tester,
+      realmUrl: 'https://chat.zulip.org',
+      topic: 'What\'s up? 123!',
+      expectedLink: 'https://chat.zulip.org/#narrow/stream/42-stream-42/topic/What\'s.20up.3F.20123!/near/12345',
+      );
+    });
+
+    testWidgets('Copies link with realm URL ending in path', (tester) async {
+      await testMessageLink(
+      tester: tester,
+      realmUrl: 'https://chat.zulip.org/path',
+      topic: 'Numbers & Symbols: 100%',
+      expectedLink: 'https://chat.zulip.org/path#narrow/stream/42-stream-42/topic/Numbers.20.26.20Symbols.3A.20100.25/near/12345',
+      );
+    });
+
+    testWidgets('Copies link with realm URL ending in slash', (tester) async {
+      await testMessageLink(
+      tester: tester,
+      realmUrl: 'https://chat.zulip.org/',
+      topic: 'Regular topic',
+      expectedLink: 'https://chat.zulip.org/#narrow/stream/42-stream-42/topic/Regular.20topic/near/12345',
+      );
     });
   });
 


test/model/internal_link_test.dart
```diff
@@ -54,6 +54,37 @@ void main() {
...
   }
 
+  void testExpectedUrls(List<(String, String)> testCases, {
+    List<ZulipStream>? streams,
+    List<User>? users,
+  }) {
+    assert(streams != null || users != null);
+    for (final testCase in testCases) {
+      final String inputUrl = testCase.$1;
+      final String expectedUrl = testCase.$2;
```
PIG208, line 64: nit:
```suggestion
   for (final (inputUrl, expectedUrl) in testCases) {
```
I can see that this helper parallels `testExpectedNarrows`, where the same pattern is also applicable.

lib/model/internal_link.dart
```diff
@@ -98,7 +98,7 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
     fragment.write('/near/$nearMessageId');
   }
 
-  return store.realmUrl.replace(fragment: fragment.toString());
```
PIG208, line 101: 96593ba `internal_links [nit]: Add path: '/' to store.realmUrl.replace for proper Url handling`
I think we don't ever use `[nit]` in commit summaries, so we can just drop that.

test/model/internal_link_test.dart
```diff
@@ -54,6 +54,37 @@ void main() {
     }
```
PIG208, line 54: 3010732 `internal_link_test: Create 'parseInternalLink full url cases' test`

We usually use `internal_link test` for changes like this. `git log --oneline -- path/to/file` is a way to look for predecessors.

lib/model/internal_link.dart
PIG208: We don't have tests in this commit, and the tests added in the other commit do not fail if we remove this change. We need to do some testing here.

test/model/internal_link_test.dart
```diff
@@ -54,6 +54,37 @@ void main() {
...
+    final streams = [
+      eg.stream(streamId: 1, name: 'check'),
+      eg.stream(streamId: 3, name: 'mobile'),
+    ];
+
+    final fullUrlTestCases = [
+      ('https://example.com/#narrow/stream/check', 'https://example.com/#narrow/stream/check'),
+      ('https://example.com/#narrow/stream/3-mobile/topic/topic1', 'https://example.com/#narrow/stream/3-mobile/topic/topic1'),
+      ('https://another.com/#narrow/stream/mobile', 'https://another.com/#narrow/stream/mobile'),
+      ('https://example.com/#narrow/stream/check/topic/test', 'https://example.com/#narrow/stream/check/topic/test'),
```
PIG208, line 83: I'm not sure if I understand the purpose of these test cases, as the input and the expected output are the same.

PIG208:
Thanks for the PR! Left some comments.

lib/model/internal_link.dart
```diff
@@ -59,7 +59,7 @@ String? decodeHashComponent(String str) {
 Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
   // TODO(server-7)
   final apiNarrow = resolveDmElements(
-    narrow.apiEncode(), store.connection.zulipFeatureLevel!);
+  narrow.apiEncode(), store.connection.zulipFeatureLevel!);
```
PIG208, line 62: Looks like the indentation change here is unintentional.

lib/model/internal_link.dart
```diff
@@ -95,8 +95,11 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
   if (nearMessageId != null) {
     fragment.write('/near/$nearMessageId');
   }
+  final realmUrlWithSlash = store.realmUrl.path.endsWith('/')
+      ? store.realmUrl
+      : store.realmUrl.replace(path: '${store.realmUrl.path}/');
```
PIG208, line 100: nit:
```suggestion
  final realmUrlWithSlash = store.realmUrl.path.endsWith('/')
    ? store.realmUrl
    : store.realmUrl.replace(path: '${store.realmUrl.path}/');
```

test/model/internal_link_test.dart
```diff
@@ -451,4 +452,48 @@ void main() {
...
+    ];
+
+    test('narrowLink should include a trailing slash in the path and before #', () async {
+      final store = await setupStore(realmUrl: realmUrl, streams: streams);
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      final messageLink = narrowLink(store, narrow, nearMessageId: 12345);
+      final urlString = messageLink.toString();
+      final fragmentIndex = urlString.indexOf('#');
+      expect(fragmentIndex, greaterThan(0));
```
PIG208, line 468: We usually use [checks](https://pub.dev/packages/checks) for our tests. So this can be `check(fragmentIndex).isGreaterThan(0);` instead. This applies to the other checks here too.

test/model/internal_link_test.dart
```diff
@@ -451,4 +452,48 @@ void main() {
...
+      }
+    });
+  });
+
+  group('narrowLink Full URL generation', () {
+    final streams = [
+      eg.stream(streamId: 123, name: 'check'),
+    ];
+    test('narrowLink should generate the full URL correctly', () async {
+      final store = await setupStore(realmUrl: realmUrl, streams: streams);
```
PIG208, line 481: This `realmUrl` is `https://example.com/`, which comes with a trailing `/`. So we might want to override this for testing.

test/model/internal_link_test.dart
```diff
@@ -451,4 +452,48 @@ void main() {
...
+
+      expect(uri.scheme, equals('https'));
+      expect(uri.host, equals('example.com'));
+      expect(uri.path.endsWith('/'), isTrue);
+      const expectedFragmentStart = 'narrow/stream/123-check/topic';
+      expect(uri.fragment.startsWith(expectedFragmentStart), isTrue);
+      expect(
+          urlString.startsWith('https://example.com/#narrow/stream/123-check/topic'),
+          isTrue
+      );
```
PIG208, line 496: I think this test and the one before it can be combined into one. However, looking at this more, it might be simpler to just check if `urlString` equals the URL we expect (i.e.: `'https://example.com/#narrow/stream/123-check/topic'`).

Or, even better, we can add the test to `test/widgets/action_sheet_test.dart`, as a test similar to "copies message link to clipboard". This way we get to test it end-to-end. In this case, no change is needed for `test/model/internal_link_test.dart`.

lib/model/internal_link.dart
```diff
@@ -59,7 +59,7 @@ String? decodeHashComponent(String str) {
 Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
```
PIG208, line 59: Some nits on commit message:
- `updates` -> `Updates`
- `...fragment identifier This...` -> `...fragment identifier. This...`
- `likified` -> `linkified`

and an empty space between the paragraph and `Fixes: #845`.

PIG208 (commented):
Thanks for the update! Went over this and left some comments. One other suggestion is to add the tests in the same commit that makes the change.

test/widgets/action_sheet_test.dart
```diff
@@ -553,6 +553,21 @@ void main() {
     });
   });
 
+
+  group('narrowLink URL generation tests', () {
+
```
PIG208, line 558: Extra empty line here and above the group.

test/widgets/action_sheet_test.dart
```diff
@@ -553,6 +553,21 @@ void main() {
...
+    testWidgets('Full Url Testing', (tester) async {
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
```
PIG208, line 562: For this test, it matters for us that the `store.realmUrl` used does not end with a slash. However, the default value is "https://chat.example/" (see `eg.account`). We should find a way to override that when setting up the store.

test/widgets/action_sheet_test.dart
```diff
@@ -553,6 +553,21 @@ void main() {
...
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+      final generatedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
```
PIG208, line 564: To make this end-to-end, let's generate the link through `tapCopyMessageLinkButton`. We should move this test to the "CopyMessageLinkButton" group.

test/widgets/action_sheet_test.dart
```diff
@@ -553,6 +553,21 @@ void main() {
...
+      const expectedPrefix = 'https://chat.example/#narrow/stream/123-stream-123/topic';
+      final truncatedLink = generatedLink.split('/example').first;
+      check(truncatedLink).equals(expectedPrefix);
```
PIG208, line 567: nit:
```suggestion
      check(generatedLink).startsWith('https://chat.example/#narrow/stream/123-stream-123/topic');
```

lib/model/internal_link.dart
```diff
@@ -96,7 +96,11 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
     fragment.write('/near/$nearMessageId');
```
PIG208, line 96: Looks like I missed this in the last round. "Variable" in the commit summary is capitalized, but we should use "variable" instead.

PIG208 (commented):
Thanks for the update! The implementation now looks good to me. Added some comments on the new test.

test/widgets/action_sheet_test.dart
```diff
@@ -24,6 +24,7 @@ import '../api/fake_api.dart';
 import '../example_data.dart' as eg;
 import '../flutter_checks.dart';
 import '../model/binding.dart';
+import '../model/internal_link_test.dart';
 import '../model/test_store.dart';
```
satyam372, line 28: nit: Import internal_link_test.dart to use setupstore to generate link without an '/' before '#'.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +552,19 @@ void main() {
...
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+      final store = await setupStore(realmUrl: Uri.parse('https://chat.example'));
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
+      final generatedLink = (await Clipboard.getData('text/plain'))?.text ?? '';
+      final normalizedExpectedLink = expectedLink.replaceAll('123-unknown', '123-stream-123');
+      check(generatedLink).equals(normalizedExpectedLink);
```
satyam372, line 566: image

before normalization, the test was failing due to mismatch.

satyam372:
PTAL @PIG208

test/widgets/action_sheet_test.dart
PIG208: Hmm, interesting. Looks like that could be a result of [not having the stream in store](https://github.com/zulip/zulip-flutter/blob/ea8c76584534e4cd32452f0257c1f5aaec00550b/lib/model/internal_link.dart#L75). This is because `expectedLink` and `generatedLink` are generated from separate instances of `PerAccountStore`:
`expectedLink` uses the store created with `setupStore` above;
`generatedLink` uses the store set up from `setupToMessageActionSheet`.

How about adding an optional parameter `Account? account` to `setupToMessageActionSheet`, which can be used to override the `realmUrl`?

PIG208 (commented):
Thanks for the revision! Left a comment regarding the output mismatch issue.

test/widgets/action_sheet_test.dart
```diff
@@ -38,11 +38,12 @@ late FakeApiConnection connection;
 Future<void> setupToMessageActionSheet(WidgetTester tester, {
   required Message message,
   required Narrow narrow,
+  Account? account,
 }) async {
```
satyam372, line 42: Add parameter Account so that we can override realmUrl to produce Url without '/' before '#'.

satyam372:
PTAL @PIG208 And thank you for your suggestion.

test/widgets/action_sheet_test.dart
```diff
@@ -38,11 +38,12 @@ late FakeApiConnection connection;
...
   addTearDown(testBinding.reset);
-
-  await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
-  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  final effectiveAccount = account ?? eg.selfAccount;
+  await testBinding.globalStore.add(effectiveAccount, eg.initialSnapshot());
+  final store = await testBinding.globalStore.perAccount(effectiveAccount.id);
   await store.addUser(eg.user(userId: message.senderId));
   if (message is StreamMessage) {
     final stream = eg.stream(streamId: message.streamId);
```
PIG208, line 49: Thanks! There is one extra spot below where the `account` can be used.

```
await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id
```

test/widgets/action_sheet_test.dart
```diff
@@ -38,11 +38,12 @@ late FakeApiConnection connection;
 Future<void> setupToMessageActionSheet(WidgetTester tester, {
   required Message message,
   required Narrow narrow,
+  Account? account,
 }) async {
   addTearDown(testBinding.reset);
-
-  await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
-  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  final effectiveAccount = account ?? eg.selfAccount;
```
PIG208, line 44: nit:
```suggestion
  account ??= eg.selfAccount;
```

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +552,26 @@ void main() {
...
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
+
+      final customAccount = eg.selfAccount.copyWith(id: 2,realmUrl: Uri.parse('https://chat.example'));
```
PIG208, line 566: nit:
```suggestion
      final customAccount = eg.selfAccount.copyWith(id: 2, realmUrl: Uri.parse('https://chat.example'));
```

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +552,26 @@ void main() {
...
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow, account: customAccount);
+      final customAccountStore = await testBinding.globalStore.perAccount(customAccount.id);
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final generatedLink = narrowLink(customAccountStore, narrow, nearMessageId: message.id).toString();
+      check(expectedLink).equals(generatedLink);
```
PIG208, line 573: nit (simplify):
```suggestion
      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
```

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +552,26 @@ void main() {
...
+
+    testWidgets('Full Url Testing', (tester) async {
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
```
PIG208, line 564: Looks like we can remove these lines after updating the last two lines.

PIG208 (commented):
Thank you for the revision! The refactor looks good overall. I think these will be my last couple of comments before we move on to integration review.

test/widgets/action_sheet_test.dart
```diff
@@ -38,11 +38,12 @@ late FakeApiConnection connection;
...
   required Message message,
   required Narrow narrow,
+  Account? account,
 }) async {
   addTearDown(testBinding.reset);
-
-  await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
-  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  final effectiveAccount = account ?? eg.selfAccount;
+  await testBinding.globalStore.add(effectiveAccount, eg.initialSnapshot());
```
PIG208, line 45: We probably want to override `zulipFeatureLevel` to `account.zulipFeatureLevel` for `eg.initialSnapshot` too.

test/widgets/action_sheet_test.dart
```diff
@@ -616,4 +634,4 @@ void main() {
       check(mockSharePlus.sharedString).isNull();
     });
   });
-}
+}
```
PIG208, line 637: nit: looks like this is unintended?

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,21 @@ void main() {
       final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
     });
+
+    testWidgets('Full Url Testing', (tester) async {
+      final message = eg.streamMessage();
+      final narrow = TopicNarrow.ofMessage(message);
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow);
```
PIG208, line 561: We shouldn't need to setup the action sheet again. We did that at line 566.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,21 @@ void main() {
...
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
```
PIG208, line 563: nit: Move this to just before the line where we define `expectedLink`, so the reader gets the contextual information about where the clipboard data comes from

PIG208 (commented):
LGTM overall! Some small comments, and I going to mark this for integration review.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,21 @@ void main() {
...
+
+      final customAccount = eg.selfAccount.copyWith(id: 2, realmUrl: Uri.parse('https://chat.example'));
+      await setupToMessageActionSheet(tester, message: message, narrow: narrow, account: customAccount);
+      final customAccountStore = await testBinding.globalStore.perAccount(customAccount.id);
+
+      final expectedLink = narrowLink(customAccountStore, narrow, nearMessageId: message.id).toString();
```
PIG208, line 569: These lines are too long. Try to wrap them so that they are under 80 columns wide.

For example:
```diff
-     await setupToMessageActionSheet(tester, message: message, narrow: narrow, account: customAccount);
+     await setupToMessageActionSheet(
+       tester, message: message, narrow: narrow, account: customAccount);

```

test/widgets/action_sheet_test.dart
```diff
@@ -60,8 +63,8 @@ Future<void> setupToMessageActionSheet(WidgetTester tester, {
     historyLimited: false,
     messages: [message],
   ).toJson());
-
-  await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
+  await tester.pumpAndSettle();
```
PIG208, line 66: Looks like this can be removed, unless I missed something?

test/widgets/action_sheet_test.dart
PIG208: Bump. We should add an empty line to the end of the file.

lib/model/internal_link.dart
```diff
@@ -96,7 +96,11 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
...
   }
 
-  return store.realmUrl.replace(fragment: fragment.toString());
+  final realmUrlWithSlash = store.realmUrl.path.endsWith('/')
+      ? store.realmUrl
+      : store.realmUrl.replace(path: '${store.realmUrl.path}/');
```
gnprice, line 101: This logic seems too broad. If the realm URL looks like `https://chat.example/foo`, we don't have a reason to turn that to `https://chat.example/foo/`.

Unless `https://chat.example/foo#narrow/etc` fails to linkify in a Zulip message, that is. Does it?

test/widgets/action_sheet_test.dart
```diff
@@ -38,11 +38,14 @@ late FakeApiConnection connection;
...
   required Message message,
   required Narrow narrow,
+  Account? account,
 }) async {
   addTearDown(testBinding.reset);
 
-  await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
-  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  account ??= eg.selfAccount;
+  final initialSnapshot = eg.initialSnapshot(zulipFeatureLevel: account.zulipFeatureLevel);
```
gnprice, line 46: Does this need to start mentioning zulipFeatureLevel? If not, keep the story simpler by leaving it out.

test/widgets/action_sheet_test.dart
gnprice: (Not an empty line — the problem is that every line should end with a newline, and this commit is editing the file's last line by removing its terminating newline. Previous discussion: https://github.com/zulip/zulip-flutter/pull/534#discussion_r1515353889 )

gnprice (commented):
Thanks @satyam372 for the contribution, and @PIG208 for the previous reviews!

Initial comments from me below, in addition to @PIG208's remaining comments above.

This PR also lacks the kind of test that was specified in the issue:
> as part of a PR fixing this issue, we should also add a few tests that include, verbatim, the full URL they expect as the output.

It will need to add those.

satyam372:
Thanks @gnprice and @PIG208  for the review. just a small question.  
In the current tests, we are dynamically generating and verifying the URL.
Should we be comparing the generated URL directly against a hardcoded, literal full URL string within the test (e.g., including specific realm URLs and message IDs), rather than dynamically building it?

gnprice:
Yes. As the issue says:
> we should also add a few tests that include, verbatim, the full URL they expect as the output.

For further questions about how to complete the PR, I recommend a chat thread in `#mobile-dev-help`. That may get a faster response.

lib/model/internal_link.dart
satyam372: Yes, I tested some URLs and Here are the results:

In a browser: The browser automatically adds a / before the #.

In Zulip Flutter application: When I paste the URL, it does not linkify and is not clickable.
Based on this behavior, I have added a / before the #.

gnprice: The question is about what Zulip does, though, when you put such a URL into a message. That's what this issue #845 is about:
> This is still a valid URL, I believe — but it looks funny, and more concretely it means that if you post it in a Zulip message it doesn't get linkified.

So I tested that just now. The answer is that it does linkify:
https://chat.zulip.org/#narrow/stream/7-test-here/topic/linkification/near/1957680

That is, if you send a message with this text:
https://chat.example/foo#narrow/etc
it becomes a link.

That's the same as what happens if you send a message with this text:
https://chat.example/#narrow/etc

By contrast if you send a message with this text:
https://chat.example#narrow/etc
it doesn't become a link. That's the reason #845 is an issue.

gnprice: > In Zulip Flutter application: When I paste the URL, it does not linkify and is not clickable.

This is puzzling because it's different from what I observed in my test just now. (I used the web app to send that test message — but the message content gets turned to HTML on the server, so it doesn't matter what client is used.)

Can you post a link to a test message where you found the URL did not linkify? That would make it possible to pin down what variable is controlling the difference in behavior.

satyam372:
PTAL @PIG208

lib/model/internal_link.dart
PIG208: The link `https://chat.example/foo#narrow/etc` does linkify for me. So we probably don't need the slash in this case.

I think the fix for this will just be using `startswith` instead of `endsWith`, so that we don't process `https://chat.example/foo`, but still handle `https://chat.example#anything`.

These are something that are good to have tests for.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,93 @@ void main() {
...
+          tester, message: message, narrow: narrow, account: account);
+
+      final store = await testBinding.globalStore.perAccount(account.id);
+      if (!store.streams.containsKey(stream.streamId)) {
+        await store.addStream(stream);
+      }
+
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      const expectedLink = 'https://chat.zulip.org/#narrow/stream/123-stream-123/topic/example.20topic.20341/near/13579';
```
PIG208, line 601: These `expectedLink`s look good to me.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,93 @@ void main() {
...
+    Future<void> tapCopyMessageLinkButton(WidgetTester tester) async {
+      await tester.ensureVisible(find.byIcon(Icons.link, skipOffstage: false));
+      await tester.tap(find.byIcon(Icons.link));
+      await tester.pump(); // [MenuItemButton.onPressed] called in a post-frame callback: flutter/flutter@e4a39fa2e
+    }
+
+    testWidgets('Copies message link with verbatim, full URL output', (tester) async {
+      final stream = eg.stream(streamId: 123, name: 'stream 123');
+      final message = eg.streamMessage(
+          id: 13579, stream: stream, topic: 'example topic 341');
```
PIG208, line 587: nit: indentation, and some other places in this file.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,93 @@ void main() {
...
+      await tapCopyMessageLinkButton(tester);
+      await tester.pump(Duration.zero);
+      final expectedLink = narrowLink(
+          customAccountStore, narrow, nearMessageId: message.id).toString();
+      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
+    });
+  });
+
+  group('Message link verbatim URL tests', () {
+
```
PIG208, line 577: nit: we can remove the line here.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,93 @@ void main() {
...
+      check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
+    });
+
+    testWidgets('Copies link with special characters in stream name and topic', (tester) async {
+      final stream = eg.stream(streamId: 42, name: 'weird & wonderful');
+      final message = eg.streamMessage(
+          id: 98765, stream: stream, topic: 'Whats up? 123!');
+      final narrow = TopicNarrow.ofMessage(message);
+      final account = eg.selfAccount.copyWith(
+          realmUrl: Uri.parse('https://chat.zulip.org'));
```
PIG208, line 611: It should be fine to reuse the same set of stream and topic, because we are more interested in the variants of `realmUrl`. In addition to this one, let's also try something like `https://chat.zulip.org/path` and `https://chat.zulip.org/`. Since most of these tests will be similar, extracting a helper will be simplify the code.

test/widgets/action_sheet_test.dart
```diff
@@ -551,6 +554,93 @@ void main() {
       final expectedLink = narrowLink(store, narrow, nearMessageId: message.id).toString();
       check(await Clipboard.getData('text/plain')).isNotNull().text.equals(expectedLink);
     });
+
+    testWidgets('Full Url Testing', (tester) async {
```
PIG208, line 558: Now that we have the new tests, we should be able to remove this one.

PIG208 (commented):
Thanks for the revision! I know there has been many review comments, but we are making some progress. I think the structure of the test looks fine. You can go to #mobile-dev-help anytime if you run into issues.

lib/model/internal_link.dart
gnprice: Or even simpler than `startsWith` — check `path.isNotEmpty`. I believe the path will always either be empty or start with `/`.

satyam372:
PTAL @PIG208

satyam372:
Please review @gnprice.

Issue #747, opened 2024-06-18
SubscriptionListPage: Show @-mention indicator beside channel unread-count badge


chrisbobbe:
In the "Subscribed channels" screen, we should show an @-mention marker, if it applies, beside the channel unread-count badge. Like we already do in the "Inbox" screen; that was #384.

We have a TODO in the code for this, but it's a `TODO(#384)`. I'll update that TODO to point to this issue.

chrisbobbe:
We had a request for this in beta feedback; CZO discussion: https://chat.zulip.org/#narrow/channel/48-mobile/topic/beta.20feedback.3A.20.22.40.22.20indicator.20beside.20unread.20badge.20in.20Channels/near/2067831

chrisbobbe:
#1065 would give us an efficient data structure we can use for this. Let's plan to revisit once that's done.

PR #333, opened 2023-10-24
notif: Parse notification messages, on Android

gnprice:
This PR is stacked atop #325, and makes the next step toward #320/#122.

Issue #332, opened 2023-10-23
Optimize unreads model on mark-as-read


gnprice:
In #304 we implemented the `Unreads` model, and measured that its performance was quite fast for most operations, but not for marking messages as read. The timings are in the series of messages starting at https://github.com/zulip/zulip-flutter/pull/304#issuecomment-1775980154 .

In particular, in those experiments it took 0ms (to the nearest ms) to construct the whole model in the first place, as well as most other operations. But it took about 14ms for a mark-as-read event, whether for marking one or many individual messages as read.

As I said at https://github.com/zulip/zulip-flutter/pull/304#issuecomment-1776034014 :
> Given how very fast the constructor is in particular, I feel like it should be possible to get mark-as-read to be much faster — in particular I think it fundamentally shouldn't be any slower than the constructor. And taking 14ms is enough that it'll often cause a dropped frame, so that's worth avoiding.

The first step here is probably to profile to see specifically where the time is being spent, e.g. to get a flame chart:
https://docs.flutter.dev/tools/devtools/cpu-profiler#flame-chart

(My original suggestion at https://github.com/zulip/zulip-flutter/pull/304#discussion_r1339386508 didn't seem to help, or not much. That was without the benefit of profiling.)

Given that the existing code is already much faster at constructing the whole model from scratch than it is at handling mark-as-read, I'm expecting that there'll turn out to be a way to make this much faster without needing the deeper data-structure changes described at zulip/zulip-mobile#4684 . If we do need the latter (or do eventually pursue those too, for the sake of going even faster), then that will be a post-launch task.

PR #875, opened 2024-08-09
subscription_list: Show @-mention indicator when that applies

Khader-1:
Fixes: #747

unreads changes here are reflections of what web [has at unread.ts](https://github.com/zulip/zulip/blob/main/web/src/unread.ts#L499)

image

### DIFF
--- a/lib/model/unreads.dart
+++ b/lib/model/unreads.dart
@@ -223,6 +223,28 @@ class Unreads extends ChangeNotifier {
     }
   }
 
+  Set<int> get channelsWithUnreadMentions {
+    final channels = <int>{};
+    for (var messageId in mentions) {
+      final streamId = _streamIdsAndTopicsByMessageId[messageId]?.streamId;
+      if (streamId != null) {
+        channels.add(streamId);
+      }
+    }
+    return channels;
+  }
+
+  Set<int> get channelsWithUnmutedMentions {
+    final channels = <int>{};
+    for (var messageId in mentions) {
+      final (:streamId, :topic) = _streamIdsAndTopicsByMessageId[messageId]!;
+      if (channelStore.isTopicVisible(streamId, topic)) {
+        channels.add(streamId);
+      }
+    }
+    return channels;
+  }
+
   void reconcileMessages(List<Message> messages) {
     for (final message in messages) {
       if (message.flags.contains(MessageFlag.read)) continue;
--- a/lib/model/unreads.dart
+++ b/lib/model/unreads.dart
@@ -39,6 +39,7 @@ class Unreads extends ChangeNotifier {
     required ChannelStore channelStore,
   }) {
     final streams = <int, Map<String, QueueList<int>>>{};
+    final streamIdsAndTopicsByMessageId = <int, ({int streamId, String topic})>{};
     final dms = <DmNarrow, QueueList<int>>{};
     final mentions = Set.of(initial.mentions);
 
@@ -46,6 +47,9 @@ class Unreads extends ChangeNotifier {
       final streamId = unreadChannelSnapshot.streamId;
       final topic = unreadChannelSnapshot.topic;
       (streams[streamId] ??= {})[topic] = QueueList.from(unreadChannelSnapshot.unreadMessageIds);
+      final messageInfo = (streamId: streamId, topic: topic);
+      streamIdsAndTopicsByMessageId.addEntries(
+        unreadChannelSnapshot.unreadMessageIds.map((messageId) => MapEntry(messageId, messageInfo)));
     }
 
     for (final unreadDmSnapshot in initial.dms) {
@@ -62,6 +66,7 @@ class Unreads extends ChangeNotifier {
     return Unreads._(
       channelStore: channelStore,
       streams: streams,
+      streamIdsAndTopicsByMessageId: streamIdsAndTopicsByMessageId,
       dms: dms,
       mentions: mentions,
       oldUnreadsMissing: initial.oldUnreadsMissing,
@@ -72,11 +77,12 @@ class Unreads extends ChangeNotifier {
   Unreads._({
     required this.channelStore,
     required this.streams,
+    required Map<int, ({int streamId, String topic})> streamIdsAndTopicsByMessageId,
     required this.dms,
     required this.mentions,
     required this.oldUnreadsMissing,
     required this.selfUserId,
-  });
+  }) : _streamIdsAndTopicsByMessageId = streamIdsAndTopicsByMessageId;
 
   final ChannelStore channelStore;
 
@@ -85,6 +91,11 @@ class Unreads extends ChangeNotifier {
 
   /// Unread stream messages, as: stream ID → topic → message IDs (sorted).
   final Map<int, Map<String, QueueList<int>>> streams;
+
+  /// Stream IDs and topics for unread stream messages, as: message ID → (stream ID, topic).
+  ///
+  /// This is the same data as [streams], but in this message ID-keyed form.
+  final Map<int, ({int streamId, String topic})> _streamIdsAndTopicsByMessageId;
 
   /// Unread DM messages, as: DM narrow → message IDs (sorted).
   final Map<DmNarrow, QueueList<int>> dms;
@@ -350,6 +361,7 @@ class Unreads extends ChangeNotifier {
           case UpdateMessageFlagsAddEvent():
             if (event.all) {
               streams.clear();
+              _streamIdsAndTopicsByMessageId.clear();
               dms.clear();
               mentions.clear();
               oldUnreadsMissing = false;
@@ -446,6 +458,7 @@ class Unreads extends ChangeNotifier {
 
   void _addLastInStreamTopic(int messageId, int streamId, String topic) {
     ((streams[streamId] ??= {})[topic] ??= QueueList()).addLast(messageId);
+    _streamIdsAndTopicsByMessageId[messageId] = (streamId: streamId, topic: topic);
   }
 
   // [messageIds] must be sorted ascending and without duplicates.
@@ -458,6 +471,9 @@ class Unreads extends ChangeNotifier {
       // TODO(server-6) remove 6.0 comment
       (existing) => setUnion(existing, messageIds),
     );
+    final messageInfo = (streamId: streamId, topic: topic);
+    _streamIdsAndTopicsByMessageId.addEntries(
+      messageIds.map((messageId) => MapEntry(messageId, messageInfo)));
   }
 
   // TODO use efficient model lookups
@@ -481,6 +497,9 @@ class Unreads extends ChangeNotifier {
     for (final streamId in newlyEmptyStreams) {
       streams.remove(streamId);
     }
+    for (final messageId in idsToRemove) {
+      _streamIdsAndTopicsByMessageId.remove(messageId);
+    }
   }
 
   void _removeAllInStreamTopic(Set<int> incomingMessageIds, int streamId, String topic) {
@@ -496,6 +515,9 @@ class Unreads extends ChangeNotifier {
       if (topics.isEmpty) {
         streams.remove(streamId);
       }
+    }
+    for (final messageId in incomingMessageIds) {
+      _streamIdsAndTopicsByMessageId.remove(messageId);
     }
   }
 
--- a/lib/model/unreads.dart
+++ b/lib/model/unreads.dart
@@ -32,8 +32,6 @@ import 'channel.dart';
 /// unsubscribed streams and messages sent by muted users.
 // TODO When [oldUnreadsMissing], if you load a message list with very old unreads,
 //   sync to those unreads, because the user has shown an interest in them.
-// TODO When loading a message list with stream messages, check all the stream
-//   messages and refresh [mentions] (see [mentions] dartdoc).
 class Unreads extends ChangeNotifier {
   factory Unreads({
     required UnreadMessagesSnapshot initial,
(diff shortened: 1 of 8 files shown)


sm-sayedi (commented):
Thanks @Khader-1 for this. LGTM! Moving on to the mentor review! cc @kenclary

kenclary (approved):
lgtm

Khader-1:
Thanks @sm-sayedi and @kenclary for the reviews!
I figured that there are more cases that still need handling..
Specifically, the unread count / muted unread badge are related have more complex cases when there are mentions, I'll make more investigation on zulip/zulip and update this comment.

And currently, we don't have a record for muted mentions which should be fixed in this PR.
So marking this as a draft for now until it is ready.

chrisbobbe:
GitHub is showing there's a conflict in lib/widgets/theme.dart, could you resolve that please?

Khader-1:
Sure! just fixed PTAL @chrisbobbe.

lib/widgets/theme.dart
```diff
@@ -114,13 +114,14 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       mainBackground: const Color(0xfff0f0f0),
       title: const Color(0xff1a1a1a),
       channelColorSwatches: ChannelColorSwatches.light,
-      atMentionMarker: const HSLColor.fromAHSL(0.5, 0, 0, 0.2).toColor(),
+      atMentionMarker: const HSLColor.fromAHSL(0.7, 0, 0, 0.2).toColor(),
+      mutedAtMentionMarker: const HSLColor.fromAHSL(0.5, 0, 0, 0.2).toColor(),
```
chrisbobbe, line 118: Could you explain how you chose both of these colors? I see `atMentionMarker was changed from its previous value, but I don't understand why.

chrisbobbe (commented):
Thanks! I've read through the first six commits:

69abbbc82 msglist: Add message to mentions if mentioned
200129e8b unreads [nfc]: Introduce _reverseStreamsLookup data structure
674c4ad30 unreads [nfc]: Enhance performance in _isPresentInStreams
6e9afd5a4 unreads [nfc]: Enhance performance in _removeAllInStreams
195c1dec5 inbox [nfc]: Move AtMentionMarker to unread_count_badge
2e18afda0 inbox [nfc]: Introduce muted property to AtMentionMarker

with the last two left for next time:

0c5200976 subscription_list: Show @-mention marker when that applies
64ac89edc subscription_list: Show unread count when hasOnlyMutedMentions

because it looks like there's something to be straightened out in that sixth commit, as noted below. (All of `AtMentionMarker`'s callers passing `muted: true`, which I think doesn't make sense intuitively.)

lib/model/unreads.dart
```diff
@@ -353,7 +353,7 @@ class Unreads extends ChangeNotifier {
             } else {
               final messageIdsSet = Set.of(event.messages);
               mentions.removeAll(messageIdsSet);
-              _slowRemoveAllInStreams(messageIdsSet);
+              _removeAllInStreams(messageIdsSet);
```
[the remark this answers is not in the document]
gnprice: This version will make it slower if marking a long conversation as read — this loop is quadratic:
```dart
    for (var messageId in idsToRemove) {
      final info = _reverseStreamsLookup.remove(messageId);
      if (info != null && streams[info.streamId]?[info.topic] != null) {
        streams[info.streamId]![info.topic]!.remove(messageId);
```

because that `List.remove` call on a message in the middle of the list takes time linear in the number of remaining elements after it.

Probably the most efficient thing (in terms of our effort) would be to go ahead and add the new data structure, but leave the two "Enhance performance" commits aside and let #332 track the task of doing something like them.

lib/widgets/theme.dart
```diff
@@ -114,13 +114,14 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
...
       dmHeaderBg: const HSLColor.fromAHSL(1, 46, 0.35, 0.93).toColor(),
       errorBannerBackground: const HSLColor.fromAHSL(1, 4, 0.33, 0.90).toColor(),
       errorBannerBorder: const HSLColor.fromAHSL(0.4, 3, 0.57, 0.33).toColor(),
       errorBannerLabel: const HSLColor.fromAHSL(1, 4, 0.58, 0.33).toColor(),
       groupDmConversationIcon: Colors.black.withOpacity(0.5),
-      groupDmConversationIconBg: const Color(0x33808080),
+      groupDmConversationIconBg: const Color.fromARGB(51, 6, 0, 0),
```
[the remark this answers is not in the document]
Khader-1: Yes! Fixed now!

lib/widgets/unread_count_badge.dart
```diff
@@ -75,7 +75,9 @@ class MutedUnreadBadge extends StatelessWidget {
 }
 
 class AtMentionMarker extends StatelessWidget {
-  const AtMentionMarker({super.key});
+  const AtMentionMarker({super.key, required this.muted});
+
+  final bool muted;
```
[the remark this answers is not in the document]
Khader-1: In fact I've checked web again and turns out to me that mentions in the inbox screen are always not muted.

lib/widgets/theme.dart
Khader-1: Well I've tried to copy the values from the web app.. here is my inspector values:
for unmuted mentions and mentions in inbox screen
`hsl(0, 0, 0.2)` or `#333` with an opacity `0.7`


And for muted mentions:

`hsla(0, 0%, 20%, .5)` with an opacity `0.7`

Which could be simplified as: `hsla(0, 0%, 20%, 0.35)`

Please let me know if I am missing something.

PIG208: It will probably be helpful to link (in the commit message) the lines containing the CSS styles in effect.

lib/model/unreads.dart
```diff
@@ -39,13 +39,17 @@ class Unreads extends ChangeNotifier {
     required ChannelStore channelStore,
```
PIG208, line 39: `unreads [nfc]: Introduce _reverseStreamsLookup data structure`

The commit summary line needs an update to match the new name.

test/model/unreads_test.dart
```diff
@@ -326,6 +326,48 @@ void main() {
...
+        final stream2 = eg.stream();
+
+        prepare();
+        fillWithMessages([
+          eg.streamMessage(stream: stream1, flags: [MessageFlag.mentioned]),
+          eg.streamMessage(stream: stream1, flags: []),
+          eg.streamMessage(stream: stream2, flags: []),
+        ]);
+
+        check(model.channelsWithUnreadMentions.single).equals(stream1.streamId);
```
PIG208, line 341: nit:
```suggestion
        check(model.channelsWithUnreadMentions).single.equals(stream1.streamId);
```
So that we get a better error message showing what's in `model.channelsWithUnreadMentions` if this fails.

test/widgets/subscription_list_test.dart
```diff
@@ -316,4 +316,58 @@ void main() {
...
+      final streamWithNoMentions = eg.stream();
+
+      await setupStreamListPage(tester,
+        subscriptions: [
+          eg.subscription(streamWithMentions),
+          eg.subscription(streamWithNoMentions),
+        ],
+        userTopics: [
+          eg.userTopicItem(streamWithMentions,   'a', UserTopicVisibilityPolicy.none),
+          eg.userTopicItem(streamWithNoMentions, 'b', UserTopicVisibilityPolicy.muted),
```
PIG208, line 359: We probably don't want to have this topic muted for this test, so that these channels only differ by the single factor we want to test with.

test/model/unreads_test.dart
```diff
@@ -326,6 +326,48 @@ void main() {
...
+        await channelStore.addSubscriptions([
+          eg.subscription(stream1),
+          eg.subscription(stream2),
+          eg.subscription(stream3, isMuted: true)]);
+
+        await channelStore.addUserTopic(stream1, 'a normal', UserTopicVisibilityPolicy.none);
+        await channelStore.addUserTopic(stream2, 'b muted', UserTopicVisibilityPolicy.muted);
+        await channelStore.addUserTopic(stream3, 'c normal', UserTopicVisibilityPolicy.none);
+
+
```
PIG208, line 362: nit: extra empty line

test/model/unreads_test.dart
```diff
@@ -326,6 +326,48 @@ void main() {
...
+        fillWithMessages([
+          eg.streamMessage(stream: stream1, flags: [MessageFlag.mentioned], topic: 'a normal'),
+          eg.streamMessage(stream: stream2, flags: [MessageFlag.mentioned], topic: 'b muted'),
+          eg.streamMessage(stream: stream3, flags: [MessageFlag.mentioned], topic: 'c normal'),
```
PIG208, line 366: It will be helpful to also test have a channel containing unread messages with no mentions.

lib/model/unreads.dart
```diff
@@ -223,6 +223,28 @@ class Unreads extends ChangeNotifier {
...
+        channels.add(streamId);
+      }
+    }
+    return channels;
+  }
+
+  Set<int> get channelsWithUnmutedMentions {
+    final channels = <int>{};
+    for (var messageId in mentions) {
+      final info = _streamIdsAndTopicsByMessageId[messageId]!;
```
PIG208, line 259: nit:
```suggestion
      final (:streamId, :topic) = _streamIdsAndTopicsByMessageId[messageId]!;
```
This way we can avoid the `info` variable.

lib/widgets/subscription_list.dart
```diff
@@ -230,7 +235,8 @@ class SubscriptionItem extends StatelessWidget {
 
     final swatch = colorSwatchFor(context, subscription);
     final hasUnreads = (unreadCount > 0);
-    final opacity = subscription.isMuted ? 0.55 : 1.0;
+    const mutedOpacity = 0.55;
```
PIG208, line 238: It would be helpful to link the source of the opacity.

test/widgets/subscription_list_test.dart
```diff
@@ -189,6 +189,29 @@ void main() {
     check(find.byType(MutedUnreadBadge).evaluate().length).equals(0);
   });
 
+  testWidgets('unread badge shows as faded when non-muted subscription has only muted mentions', (tester) async {
```
PIG208, line 192: It's good to have a test for this, but it currently doesn't check the opacity.

lib/widgets/subscription_list.dart
Khader-1: I am not sure if I got it correctly; I just used the same one that we've used before for the channel name and unread count badge.

Are you suggesting to link it to zulip mobile or web design?

PIG208: Oh I see. Yeah `0.55` is not new here; it was added in b0b8a50703d2633fbc629a680f4f58ac4a6e8251. The design came from Vlad [here](https://github.com/zulip/zulip-flutter/issues/424#issuecomment-1831121034). Perhaps we can link to that GitHub comment in an NFC commit.

lib/widgets/subscription_list.dart
```diff
@@ -185,16 +185,26 @@ class _SubscriptionList extends StatelessWidget {
...
       itemBuilder: (BuildContext context, int index) {
         final subscription = subscriptions[index];
         final unreadCount = unreadsModel!.countInChannel(subscription.streamId);
         final showMutedUnreadBadge = unreadCount == 0
           && unreadsModel!.countInChannelNarrow(subscription.streamId) > 0;
+        final hasMentions = channelsWithMentions.contains(subscription.streamId);
+        final hasOnlyMutedMentions = !subscription.isMuted && hasMentions
+          && !channelsWithUnmutedMentions.contains(subscription.streamId);
+        final mutedUnreadCount = hasOnlyMutedMentions && unreadCount == 0 ?
+          unreadsModel!.countAll(subscription.streamId) : 0;
```
gnprice, line 201: It feels like there's a lot of different conditions that are interacting here, which means there are a lot of different cases for how this can end up behaving. I can't easily tell from reading this code what all those cases are and what its behavior is in all of them, which makes it hard to tell if the behavior is what we want.

Can you try laying out (just in text, in this thread) what you believe the different cases are for a given channel, and how you believe this screen should behave in each case? Then we can discuss at that level what behavior we want, and then go back to the code and make sure the code reflects that behavior.

(Ultimately I suspect we can express the desired behavior with somewhat simpler logic than this.)

PIG208: I agree with Greg that laying this out will be helpful.

Adding the @-markers to the `SubscriptionItem`, in addition to the unreads seem to complicate the widget more. Perhaps we should be cautious about adding more flags to that widget and simplify these conditions.

lib/widgets/subscription_list.dart
```diff
@@ -276,6 +282,15 @@ class SubscriptionItem extends StatelessWidget {
...
                 bold: true)),
+          ] else if (hasOnlyMutedMentions && !subscription.isMuted) ...[
+            const SizedBox(width: 12),
+            const AtMentionMarker(muted: true),
+            Opacity(
+              opacity: mutedOpacity,
+              child: UnreadCountBadge(
+                count: mutedUnreadCount,
+                backgroundColor: swatch,
+                bold: true)),
```
PIG208, line 293: Looks like the later part is mostly identical to the previous branch except for the opacity/unread count. I think it would be beneficial to find a way that do not couple `AtMentionMarker` with `UnreadCountBadge`, so that each has individual logic, making the change easier to verify.

PIG208: For example, we can have something like (not sure if the conditions are right):
```dart
          if (hasMentions) ...[
            AtMentionMarker(muted: hasOnlyMutedMentions || subscription.isMuted)
          ]
```
that is adjancent to but independent from the `UnreadCountBadge` branches

gnprice:
Marking this PR as draft for now to get it out of our review queue. @Khader-1 if in the future you're able to pick this up again and revise for the feedback above, feel free to mark again for review. Hope you're well.

gnprice:
Closing this PR for now to get it out of our list. @Khader-1 feel free to reopen if you're able to return to this in the future; in the meantime, it'll remain here on GitHub for someone to pick up from when taking up the issue #747. Thanks again for your work on this!

PR #876, opened 2024-08-10
Drop three User fields we aren't keeping up to date; verify that's all

gnprice:
This was prompted by noticing recently that we weren't updating
[User.isActive]:
  https://github.com/zulip/zulip-flutter/pull/816#discussion_r1689098946
(/cc @sm-sayedi)

After we fixed that, I looked and found there were actually three
other fields on User which we weren't updating -- a latent bug,
since we never looked at those fields, but a bug.  Fix that.

Then add a comment showing my work on verifying that that's it,
and giving us a head start on re-verifying that in the future
with whatever's changed in the API between now and then.


lib/model/store.dart
```diff
@@ -465,6 +465,10 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
         if (user == null) {
           return; // TODO log
         }
+        // Fields on [User] not updated here:
+        //   userId, dateJoined, isBot, botType, isSystemBot
```
PIG208, line 469: Do we want to have a reference on `User`, so that we remember to update this comment when there is another immutable field?

sm-sayedi, line 469: What about also making these immutable fields `final`?!

gnprice: Good idea, thanks. Switched to that.

gnprice: Well, the more important case is when adding a mutable field, so that we add the code to update it.

Anyway, switched to a comment on `User`, in conjunction with making the immutable fields `final`.

chrisbobbe:
Thanks, LGTM! Merged.

PR #878, opened 2024-08-11
notif: Handle remove notification message

rajveermalviya:
Fixes: #341

### DIFF
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -159,7 +159,7 @@ class NotificationDisplayManager {
       number: messagingStyle.messages.length,
       extras: {
         // Used to decide when a `RemoveFcmMessage` event should clear this notification.
-        kExtraZulipMessageId: data.zulipMessageId.toString(),
+        kExtraLastZulipMessageId: data.zulipMessageId.toString(),
       },
 
       contentIntent: PendingIntent(
@@ -227,7 +227,7 @@ class NotificationDisplayManager {
     //   https://github.com/zulip/zulip-mobile/pull/4842#pullrequestreview-725817909
     var haveRemaining = false;
     final activeNotifications = await _androidHost.getActiveNotifications(
-      desiredExtras: [kExtraZulipMessageId]);
+      desiredExtras: [kExtraLastZulipMessageId]);
     for (final statusBarNotification in activeNotifications) {
       if (statusBarNotification == null) continue; // TODO(pigeon) eliminate this case
 
@@ -248,7 +248,7 @@ class NotificationDisplayManager {
       // Don't act on the summary notification for the group.
       if (statusBarNotification.tag == groupKey) continue;
 
-      final lastMessageIdStr = notification.extras[kExtraZulipMessageId];
+      final lastMessageIdStr = notification.extras[kExtraLastZulipMessageId];
       assert(lastMessageIdStr != null);
       if (lastMessageIdStr == null) continue; // TODO(log)
       final lastMessageId = int.parse(lastMessageIdStr, radix: 10);
@@ -277,13 +277,13 @@ class NotificationDisplayManager {
     }
   }
 
-  /// The key for the message-id entry in [Notification.extras] metadata.
+  /// A key we use in [Notification.extras] for the [Message.id] of the
+  /// latest Zulip message in the notification's conversation.
   ///
-  /// Currently, it is used to store the message-id in the respective
-  /// notification which is later fetched to determine if a [RemoveFcmMessage]
-  /// event should clear that specific notification.
+  /// We use this to determine if a [RemoveFcmMessage] event should
+  /// clear that specific notification.
   @visibleForTesting
-  static const kExtraZulipMessageId = 'zulipMessageId';
+  static const kExtraLastZulipMessageId = 'lastZulipMessageId';
 
   /// A notification ID, derived as a hash of the given string key.
   ///
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -209,15 +209,32 @@ class NotificationDisplayManager {
   }
 
   static void _onRemoveFcmMessage(RemoveFcmMessage data) async {
+    // We have an FCM message telling us that some Zulip messages were read
+    // and should no longer appear as notifications.  We'll remove their
+    // conversations' notifications, if appropriate, and then the whole
+    // notification group if it's now empty.
     assert(debugLog('notif remove zulipMessageIds: ${data.zulipMessageIds}'));
 
+    // There may be a lot of messages mentioned here, across a lot of
+    // conversations.  But they'll all be for one account, so they'll
+    // fall under one notification group.
     final groupKey = _groupKey(data);
+
+    // Find any conversations we can cancel the notification for.
+    // The API doesn't lend itself to removing individual messages as
+    // they're read, so we wait until we're ready to remove the whole
+    // conversation's notification.  For background discussion, see:
+    //   https://github.com/zulip/zulip-mobile/pull/4842#pullrequestreview-725817909
+    var haveRemaining = false;
     final activeNotifications = await _androidHost.getActiveNotifications(
       desiredExtras: [kExtraZulipMessageId]);
-
-    var haveRemaining = false;
     for (final statusBarNotification in activeNotifications) {
       if (statusBarNotification == null) continue; // TODO(pigeon) eliminate this case
+
+      // The StatusBarNotification object describes an active notification in the UI.
+      // Its `.tag`, `.id`, and `.notification` are the same values as we passed to
+      // [AndroidNotificationHostApi.notify] (and so to `NotificationManager#notify`
+      // in the underlying Android APIs).  So these are good to match on and inspect.
       final notification = statusBarNotification.notification;
 
       // Sadly we don't get toString on Pigeon data classes: flutter#59027
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -5,7 +5,7 @@ import 'package:http/http.dart' as http;
 import 'package:collection/collection.dart';
 import 'package:crypto/crypto.dart';
 import 'package:flutter/foundation.dart';
-import 'package:flutter/widgets.dart';
+import 'package:flutter/widgets.dart' hide Notification;
 import 'package:flutter_local_notifications/flutter_local_notifications.dart' hide Person;
 
 import '../api/notifications.dart';
@@ -84,7 +84,7 @@ class NotificationDisplayManager {
   static void onFcmMessage(FcmMessage data, Map<String, dynamic> dataJson) {
     switch (data) {
       case MessageFcmMessage(): _onMessageFcmMessage(data, dataJson);
-      case RemoveFcmMessage(): break; // TODO(#341) handle
+      case RemoveFcmMessage(): _onRemoveFcmMessage(data);
       case UnexpectedFcmMessage(): break; // TODO(log)
     }
   }
@@ -155,6 +155,10 @@ class NotificationDisplayManager {
 
       messagingStyle: messagingStyle,
       number: messagingStyle.messages.length,
+      extras: {
+        // Used to decide when a `RemoveFcmMessage` event should clear this notification.
+        kExtraZulipMessageId: data.zulipMessageId.toString(),
+      },
 
       contentIntent: PendingIntent(
         // TODO make intent URLs distinct, instead of requestCode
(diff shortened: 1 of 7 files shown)


sm-sayedi (approved):
Thanks @rajveermalviya for working on this nice feature. Did a manual testing of the feature which works fine. LGTM. Moving on to the mentor review! cc @kenclary

kenclary (approved):
lgtm too :)

test/model/binding.dart
```diff
@@ -560,8 +560,8 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
 
   final Map<String, MessagingStyle?> _activeNotificationsMessagingStyle = {};
 
-  /// Clears all active notifications that have been created via [notify].
-  void clearActiveNotifications() {
+  /// Clears all messaginging style of active notifications that have been created via [notify].
```
chrisbobbe, line 563: nit: spelling; also, instead of "messaging style", maybe this should say `[MessagingStyle]`? It's not intuitively clear what a "messaging style" is, and maybe Android could have chosen a better name for it, but using the identifier should at least shorten the reader's path to looking up what it means in the Android documentation.

lib/notifications/display.dart
```diff
@@ -202,6 +205,55 @@ class NotificationDisplayManager {
...
+        ' id: ${statusBarNotification.id}, tag: ${statusBarNotification.tag},'
+        ' notification: (group: ${notification.group}, extras: ${notification.extras}))'));
+
+      // Don't act on notifications that are for other Zulip accounts/identities.
+      if (notification.group != groupKey) continue;
+
+      // Don't act on the summary notification for the group.
+      if (statusBarNotification.tag == groupKey) continue;
+
+      final lastMessageIdStr = notification.extras[kExtraZulipMessageId];
```
chrisbobbe, line 231: As an independent improvement, could you write a dartdoc for `kExtraZulipMessageId`? I found myself wondering what it was, while reading this.

chrisbobbe (commented):
Thanks! Here's a light review, and I expect Greg may have more to say since he knows the Android-notifications system better than I do.

rajveermalviya:
Thanks for the reviews @kenclary and @chrisbobbe!
Pushed a new revision @chrisbobbe, PTAL.

lib/notifications/display.dart
```diff
@@ -202,6 +205,61 @@ class NotificationDisplayManager {
...
+      // child notifications are canceled programatically as done above.
+      await ZulipBinding.instance.androidNotificationHost
+        .cancel(tag: groupKey, id: notificationIdAsHashOf(groupKey));
+    }
+  }
+
+  /// The identifier key for the message-id of embedded in the
+  /// notification's extras bundle.
+  @visibleForTesting
+  static const kExtraZulipMessageId = 'zulipMessageId';
```
chrisbobbe, line 261: Thanks for adding this dartdoc. Does it have a missing word or something? I'm not parsing "[…] the message-id of embedded […]".

Also, "the notification" makes it sound like I should already know what kind of thing—and also which specific thing of that kind—is being referred to. Is it an instance of `StatusBarNotification`, or of `Notification`? Looking through the code, I think this string is used as a key in the `Notification.extras` map, so it seems helpful to include a reference to `[Notification.extras]`.

I'm not as familiar with the Android-notifications code as you and Greg are, and one question I had was whether this constant's value is part of the Zulip API, or if it's purely internal to our code. It seems like the latter, but I think if the dartdoc were fleshed out a bit more to answer that question, I think it would be helpful.

test/model/binding.dart
```diff
@@ -555,10 +555,13 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
...
 
+  Iterable<StatusBarNotification> get activeNotifications => _activeNotifications.values;
+  final Map<(int, String?), StatusBarNotification> _activeNotifications = {};
+
   final Map<String, MessagingStyle?> _activeNotificationsMessagingStyle = {};
 
-  /// Clears all active notifications that have been created via [notify].
-  void clearActiveNotifications() {
+  /// Clears all [MessagingStyle] of active notifications that have been created via [notify].
+  void clearActiveNotificationsMessagingStyle() {
```
chrisbobbe, line 564: `_activeNotificationsMessagingStyle` is a collection, so do we want to say `[MessagingStyle]s`, plural?

chrisbobbe (commented):
Thanks! Comments below.

lib/notifications/display.dart
```diff
@@ -202,6 +205,61 @@ class NotificationDisplayManager {
...
+        ' notification: (group: ${notification.group}, extras: ${notification.extras}))'));
+
+      // Don't act on notifications that are for other Zulip accounts/identities.
+      if (notification.group != groupKey) continue;
+
+      // Don't act on the summary notification for the group.
+      if (statusBarNotification.tag == groupKey) continue;
+
+      final lastMessageIdStr = notification.extras[kExtraZulipMessageId];
+      if (lastMessageIdStr == null) continue;
```
gnprice, line 232: This case shouldn't happen, right? It'd be a bug within this file, or at least in our understanding of how the Android SDK features we're using are supposed to work.

So let's make that explicit:

```suggestion
      assert(lastMessageIdStr != null);
      if (lastMessageIdStr == null) continue; // TODO(log)
```

lib/notifications/display.dart
```diff
@@ -202,6 +205,61 @@ class NotificationDisplayManager {
...
+      final lastMessageId = int.parse(lastMessageIdStr, radix: 10);
```
gnprice, line 233: Cool, good use of `radix: 10`

lib/notifications/display.dart
```diff
@@ -202,6 +205,61 @@ class NotificationDisplayManager {
...
+      if (data.zulipMessageIds.contains(lastMessageId)) {
+        // The latest Zulip message in this conversation was read.
+        // That's our cue to cancel the notification for the conversation.
+        await ZulipBinding.instance.androidNotificationHost
+          .cancel(tag: statusBarNotification.tag, id: statusBarNotification.id);
+      } else {
```
gnprice, line 239: The draft had a `debugLog` line here. Let's keep it; I think it's not any noisier than other debugLog lines I'm glad to have, like on Zulip events.

(And we definitely want it given we have the other debugLog lines above in this method — without this line, those just leave the reader in suspense without knowing what actually happened :slightly_smiling_face:)

test/model/binding.dart
```diff
@@ -555,10 +555,13 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
...
     _activeNotificationsMessagingStyle.clear();
   }
```
gnprice, line 566: Why clear only `_activeNotificationsMessagingStyle` and not `_activeNotifications`? It seems like that'll leave the two inconsistent with each other in a way that isn't realistic for how the real Android system behaves.

So I think the existing (in main) name and doc for this method are good, and it just needs to clear the new data structure as well as the old.

test/notifications/display_test.dart
```diff
@@ -495,6 +534,112 @@ void main() {
...
+
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+      await receiveFcmMessage(async, data);
+      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
+        conditionActiveNotif(data, 'stream:${message.streamId}:${message.topic}'),
+        conditionSummaryActiveNotif(expectedGroupKey),
+      ]);
+      testBinding.firebaseMessaging.onMessage.add(
+        RemoteMessage(data: removeFcmMessage([message]).toJson()));
+      async.flushMicrotasks();
```
gnprice, line 552: nit: this can use `receiveFcmMessage`, right?

(and a few more places below)

test/notifications/display_test.dart
```diff
@@ -495,6 +534,112 @@ void main() {
...
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
```
gnprice, line 555: nit:
```suggestion
      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();

```

these are close enough together the reader can be expected to remember the previous one :slightly_smiling_face:

test/notifications/display_test.dart
```diff
@@ -495,6 +534,112 @@ void main() {
...
+      await receiveFcmMessage(async, data);
+      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
+        conditionActiveNotif(data, 'stream:${message.streamId}:${message.topic}'),
+        conditionSummaryActiveNotif(expectedGroupKey),
+      ]);
+      testBinding.firebaseMessaging.onBackgroundMessage.add(
+        RemoteMessage(data: removeFcmMessage([message]).toJson()));
+      async.flushMicrotasks();
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
```
gnprice, line 564: hmm but maybe just delete this whole stanza? seems like it's a copy of the one above. I don't think it's expected to get a MessageFcmMessage for the same Zulip message twice. If that's intentional and you think it's a scenario which should be tested, it needs at least a comment explaining the narrative of it.

test/notifications/display_test.dart
```diff
@@ -495,6 +534,112 @@ void main() {
...
+      final data3 = messageFcmMessage(message3, streamName: stream.name);
+      final expectedGroupKey = '${data1.realmUri}|${data1.userId}';
+
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+
+      await receiveFcmMessage(async, data1);
+      await receiveFcmMessage(async, data2);
+      await receiveFcmMessage(async, data3);
+
+      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
```
gnprice, line 585: nit:

```suggestion
      await receiveFcmMessage(async, data1);
      await receiveFcmMessage(async, data2);
      await receiveFcmMessage(async, data3);
      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
```

Generally when a test has code like "make this change, this change, this change; then check the state of the world", I like to group those into one stanza. Then a blank line separates the checks from the next round of making changes.

That way the stanzas, or the blank lines separating them, serve to group together those change/check pairs.

test/notifications/display_test.dart
```diff
@@ -495,6 +534,112 @@ void main() {
...
+        conditionActiveNotif(data3, 'stream:${stream.streamId}:$topicA'),
+        conditionSummaryActiveNotif(expectedGroupKey),
+      ]);
+
+      testBinding.firebaseMessaging.onMessage.add(
+        RemoteMessage(data: removeFcmMessage([message1, message2]).toJson()));
+      async.flushMicrotasks();
+
+      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
```
gnprice, line 594: This could benefit from a brief comment telling the story of the test:
```suggestion
      // A RemoveFcmMessage for the first two messages; the notification stays.
      testBinding.firebaseMessaging.onMessage.add(
        RemoteMessage(data: removeFcmMessage([message1, message2]).toJson()));
      async.flushMicrotasks();

      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
```

(and then a couple of nits above apply to other aspects of these lines)

Similarly later in this test case:
```dart
      // Then a RemoveFcmMessage for the last message; clear the notification.
```

The test name above already gives a good summary of the plot. But I think these are helpful to make sure the reader sees the point, and how the details connect to the idea that's being tested.

gnprice (commented):

Generally this looks great. I like these test cases. Various small comments below, and I see there are a couple open from Chris's review today above.

Let's add one other test case: the RemoveFcmMessage shouldn't affect any notifications for other accounts. To make the test as sharp as possible, let's exercise the case where by sheer bad luck the other account's notification has the same message ID, and is to the same user ID — the only difference is the realm URL. (This isn't even *totally* improbable: if the accounts are in two different very small or new servers, our user could easily have the same very small user ID in both and then the two servers could even hit the same small message ID around the same time.)

I guess to be thorough let's make that two test cases (or perhaps two steps in one test case): different realm URL but same user ID, and same realm URL but different user ID. And in both cases the same message ID. (For the same realm and different user IDs, that's even less improbable: our user has two accounts in the realm, both got notified by some message, and now they've read the message from just one account. I guess for that scenario it would actually be a defensible product choice to remove both notifications, since it's truly the same message and not two unrelated messages that happen to have the same ID… but I like the behavior we have, so let's test it.)

I'll also want to look closer at the implementation of `_onRemoveFcmMessage` in comparison with the zulip-mobile version — it looks right, and most of the comments even look familiar, but I'll want to do a careful line-by-line comparison to be sure we understand any differences. (Skipping that in this round only because it's late evening here.)

rajveermalviya:
Thanks for the reviews @chrisbobbe and @gnprice! Pushed a new revision, PTAL.

chrisbobbe:
Cool, thanks! I see my review has been addressed, so I'll remove the maintainer-review label. 🙂

gnprice (commented):
Thanks for the revision! This looks great.

One nit below, and then I've pushed a few added commits on top:
b9addf2d9 notif [nfc]: Make shorter local alias for ZulipBinding.instance.androidNotificationHost
f2c8e405d notif [nfc]: Keep more comments in remove-notif from zulip-mobile implementation
aa5c31d48 notif [nfc]: Clarify meaning of last-message-ID extra

I did that comparison to the zulip-mobile code which I mentioned above, and all the logic looks like a match but there were more comments I felt should come over too; that's the second commit above.

I also fixed the one nit. @chrisbobbe would you take a look at the new commits? In particular see if they answer the questions you had at https://github.com/zulip/zulip-flutter/pull/878#discussion_r1718915953 above. Then I think this is ready to merge.

chrisbobbe:
> I also fixed the one nit. @chrisbobbe would you take a look at the new commits? In particular see if they answer the questions you had at [#878 (comment)](https://github.com/zulip/zulip-flutter/pull/878#discussion_r1718915953) above. Then I think this is ready to merge.

Yep, it's clear now, thanks! Please merge at will.

gnprice:
Thanks! Merged.

Issue #124, opened 2023-05-26
stream/topic autocomplete


chrisbobbe:
After #49, which is likely to inform the architecture here.

ryanwwest:
I'm excited for this feature!

gnprice:
For anyone picking up this issue: the PR #884 has previous work that might be helpful to borrow from. See Zulip's guide to continuing unfinished work:
https://zulip.readthedocs.io/en/latest/contributing/continuing-unfinished-work.html

alya:
https://github.com/zulip/zulip/issues/36370 has a plan for updating channel ordering in the web app typeahead. We can have refining the ordering be a follow-up in the mobile app, but could maybe pull in some basic ideas, like handling deactivated and unsubscribed channels, if we aren't doing that already.

sm-sayedi:
Yeah, that would be a good follow-up. Right now in #1902, we do follow the criteria in [`typeahead_helper.sort_streams`](https://github.com/zulip/zulip/blob/9fe24279d7208233d003d1313abc95f654decf4d/web/src/typeahead_helper.ts#L998-L1008) for sorting in autocomplete.

PR #884, opened 2024-08-12
autocomplete: Introduce channel/topic #-mention

Khader-1:
- **autocomplete [nfc]: Rename Autocomplete to ComposeContentAutocomplete**
- **autocomplete [nfc]: Factor out generic AutocompleteView**
- **autocomplete [nfc]: Factor out generic AutocompleteField**
- **api: Add route getStreamTopics**
- **autocomplete: Implement targeted topic autocomplete**
- **autocomplete [nfc]: Factor out generic AutocompleteField**
- **autocomplete: Implement targeted topic autocomplete**
- **autocomplete [nfc]: Use MentionAutocompleteQuery as parent class**
- **autocomplete [nfc]: Convert MentionAutocomplete CandidateT to Object**
- **autocomplete [nfc]: Update intent logic to enable more mention types**
- **autocomplete [nfc]: Introduce `_triage` function similar to web**
- **autocomplete: Introduce channel compose autocomplete**


lib/model/autocomplete.dart
```diff
@@ -289,6 +538,38 @@ class MentionAutocompleteView extends ChangeNotifier {
         streamId: streamId, senderId: userB.userId));
   }
 
+  @override
+  Iterable<Object> getSortedItemsToTest(MentionAutocompleteQuery query) {
+    switch (query) {
+      case UserMentionAutocompleteQuery():
+        return sortedUsers;
+      case ChannelMentionAutocompleteQuery(:var raw):
```
gnprice, line 546: Instead of putting conditional logic within `MentionAutocompleteView` to handle both @-mentions and channels/topics, let's leave that class focused on @-mentions and make a new class for channels/topics.

I guess the existing "mention" naming made that design ambiguous, because I do say "#-mention" for mentioning a channel or topic. But the intent of MentionAutocompleteView and MentionAutocompleteQuery is that they're about @-mentions.

Probably a good prep commit is to rename all the MentionAutocomplete… classes to say AtMentionAutocomplete… instead, to make that clearer. ("User mention" isn't quite right, because the same autocomplete interaction can produce a mention of a group or of a wildcard like `@channel`.)

Similarly the new classes can say HashMentionAutocomplete… — "channel mention" isn't quite right because these also include mentioning a topic.

lib/model/autocomplete.dart
```diff
@@ -289,6 +538,38 @@ class MentionAutocompleteView extends ChangeNotifier {
...
+          sortingComparator: _channelComparator);
+        return [...nameMatches.matches, ...descriptionMatches.matches, ...descriptionMatches.rest];
+    }
+  }
+
+  @override
+  MentionAutocompleteResult? testItem(MentionAutocompleteQuery query, Object item) {
+    switch (query) {
+      case UserMentionAutocompleteQuery():
+        item as User;
```
gnprice, line 560: The need for this cast here (and the corresponding fact that `CandidateT` is just `Object` when the truth is that the candidate type is determined by the query type and is always either User or ZulipStream) is also a sign that this isn't the right arrangement of the types.

lib/model/autocomplete.dart
```diff
@@ -23,26 +24,126 @@ extension Autocomplete on ComposeContentController {
...
+    }
+  }
+  return _TriageRawData(exactMatches: exactMatches,
+    beginsWithCaseSensitiveMatches: beginsWithCaseSensitiveMatches,
+    beginsWithCaseInsensitiveMatches: beginsWithCaseInsensitiveMatches,
+    wordBoundaryMatches: wordBoundaryMatches,
+    noMatches: noMatches);
+}
+
+({List<T> matches, List<T> rest}) _triage<T> ({
```
gnprice, line 121: Because this function (and its helper `_triageRaw`) are designed to closely mimic particular pieces of web's logic, they should get comments linking to the corresponding web code.

That helps a reviewer compare to the web code to see what differences there might be. It also helps in the future — particularly if that link is a permalink, which it should be — for a reader to compare against what's changed in web since then, and see if changes should be copied over to here.

gnprice (commented):
Thanks! Here's a couple of high-level comments on a brief skim.

lib/model/autocomplete.dart
Khader-1: Thanks @gnprice! I'd like to get you're suggestion in that case as we'd have two view models for the same autocomplete field, which is not implemented in the current design.

So do you think it'd be OK to update `AutocompleteField` so it has a list of `AutocompleteViewModel` and then, based on the query select and use one of them to search for results? Or would you take another approach?

gnprice: The autocomplete view model's lifetime corresponds to the interaction the user has to attempt an autocomplete. There should be only one of those going on at a time — we won't want to try to show @-mention and #-mention results at the same time, for example — so there should only need to be one autocomplete view model for the field at a time.

The `AutocompleteIntent` will specify which kind of autocomplete is desired. So that can be used to construct the right type of `AutocompleteView`. When there isn't (yet) an `AutocompleteIntent`, we don't make an `AutocompleteView`.

gnprice:
Closing this PR to get it out of our list, since it hasn't been revised after the initial round last August. The branch will remain here on GitHub for someone to draw from when taking up the issue #124. Thanks again @Khader-1 for your work on this!

Issue #165, opened 2023-06-08
polls: Read-only support


gnprice:
As a first step of poll support (#164), we should render poll messages.

For product context, see https://zulip.com/help/create-a-poll .

For details of this part of the API… unfortunately this is one of the few remaining parts of the Zulip API that is effectively undocumented, and one has to resort to reverse-engineering. Two useful resources for reference are:
* The implementation in zulip-mobile, which comes from reverse-engineering and embodies notes about the format.
  * This lives mainly in `src/webview/html/message.js`, in particular `widgetBody`. (Here "widget" is a term this part of Zulip sometimes uses for an abstraction whose primary example is polls.)
  * See also types `Submessage`, `SubmessageData`, `WidgetEventData`, and `WidgetData` in `src/api/modelTypes.js`, and their jsdoc.
* The implementation in Zulip web.
  * Part of this is in JS code that's shared with zulip-mobile: see `web/shared/src/poll_data.js` and its neighbor `poll_data.js.flow`. This is used by the zulip-mobile implementation mentioned above.
  * The web-only code that consumes that `poll_data` module may also be informative.

gnprice:
Ideally the information we show here would match what's in Zulip web: in particular for each option it'd show the number of votes and also the list of people who voted for it.

Currently zulip-mobile actually shows less than that:
* zulip/zulip-mobile#4882

I think it's probably just about as easy to go ahead and show that information, when we're working on this issue, as to not do so. In that case we should go ahead and show it.

(I think when we added poll support in zulip-mobile I may have been concerned that the list of voters would be too big to comfortably view in the message, the way web does it, on a mobile-size screen. A more mobile-optimized alternative would be to have the list on a separate screen, similar to the list of people who reacted on a message. But given that we never followed up by building that second screen and the interaction to get to it, it would have been better in retrospect to go ahead with the perhaps-unwieldy inline list of voters.)

gnprice:
A small extra implementation note:

There are some ancient messages (2019-01 and earlier) on at least chat.zulip.org which use a different variation of the polls API, which was changed while the feature was still experimental before it was released. When we see such a message, we should just reject it without trying to parse it further.

Specifically that variation involves `new_comment` rather than `new_option`. Details at:
* https://github.com/zulip/zulip/issues/30148

PR #885, opened 2024-08-12
Support polls (read-only)

PIG208:
Stacked on top of #823.

Fixes: #165

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -3,6 +3,7 @@ import 'package:json_annotation/json_annotation.dart';
 import 'events.dart';
 import 'initial_snapshot.dart';
 import 'reaction.dart';
+import 'submessage.dart';
 
 export 'json.dart' show JsonNullable;
 export 'reaction.dart';
@@ -533,6 +534,9 @@ sealed class Message {
   final String senderRealmStr;
   @JsonKey(name: 'subject')
   String topic;
+  /// Poll data if "submessages" describe a poll, `null` otherwise.
+  @JsonKey(name: 'submessages', readValue: _readPoll, fromJson: Poll.fromJson, toJson: Poll.toJson)
+  Poll? poll;
   final int timestamp;
   String get type;
 
@@ -562,6 +566,13 @@ sealed class Message {
   static List<MessageFlag> _flagsFromJson(Object? json) {
     final list = json as List<Object?>;
     return list.map((raw) => MessageFlag.fromRawString(raw as String)).toList();
+  }
+
+  static Poll? _readPoll(Map<Object?, Object?> json, String key) {
+    return Submessage.parseSubmessagesJson(
+      json['submessages'] as List<Object?>? ?? [],
+      messageSenderId: (json['sender_id'] as num).toInt(),
+    );
   }
 
   Message({
--- a/lib/api/model/model.g.dart
+++ b/lib/api/model/model.g.dart
@@ -282,7 +282,7 @@ StreamMessage _$StreamMessageFromJson(Map<String, dynamic> json) {
     matchTopic: json['match_subject'] as String?,
     displayRecipient: json['display_recipient'] as String?,
     streamId: (json['stream_id'] as num).toInt(),
-  );
+  )..poll = Poll.fromJson(Message._readPoll(json, 'submessages'));
 }
 
 Map<String, dynamic> _$StreamMessageToJson(StreamMessage instance) {
@@ -301,6 +301,7 @@ Map<String, dynamic> _$StreamMessageToJson(StreamMessage instance) {
     'sender_id': instance.senderId,
     'sender_realm_str': instance.senderRealmStr,
     'subject': instance.topic,
+    'submessages': Poll.toJson(instance.poll),
     'timestamp': instance.timestamp,
     'flags': instance.flags,
     'match_content': instance.matchContent,
@@ -360,7 +361,7 @@ DmMessage _$DmMessageFromJson(Map<String, dynamic> json) => DmMessage(
       matchTopic: json['match_subject'] as String?,
       displayRecipient: const DmRecipientListConverter()
           .fromJson(json['display_recipient'] as List),
-    );
+    )..poll = Poll.fromJson(Message._readPoll(json, 'submessages'));
 
 Map<String, dynamic> _$DmMessageToJson(DmMessage instance) => <String, dynamic>{
       'client': instance.client,
@@ -377,6 +378,7 @@ Map<String, dynamic> _$DmMessageToJson(DmMessage instance) => <String, dynamic>{
       'sender_id': instance.senderId,
       'sender_realm_str': instance.senderRealmStr,
       'subject': instance.topic,
+      'submessages': Poll.toJson(instance.poll),
       'timestamp': instance.timestamp,
       'flags': instance.flags,
       'match_content': instance.matchContent,
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -65,7 +65,9 @@ enum SubmessageType {
 ///  * Zero or more submessages with content [PollEventSubmessage] if the
 ///    message is a poll (i.e. if the first submessage was a [PollWidgetData]),
 ///    and similarly for other types of widgets.
-sealed class SubmessageData {}
+sealed class SubmessageData {
+  Object? toJson();
+}
 
 /// The data encoded in a submessage to make the message a Zulip widget.
 ///
@@ -87,6 +89,7 @@ sealed class WidgetData extends SubmessageData {
     };
   }
 
+  @override
   Object? toJson();
 }
 
@@ -188,6 +191,7 @@ sealed class PollEventSubmessage extends SubmessageData {
     }
   }
 
+  @override
   Map<String, Object?> toJson();
 }
 
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -173,7 +173,7 @@ sealed class PollEventSubmessage extends SubmessageData {
   ///
   /// For options that are a part of the initial [PollWidgetData], the
   /// [senderId] should be `null`.
-  static String optionKey({required int? senderId, required int idx}) =>
+  static PollOptionKey optionKey({required int? senderId, required int idx}) =>
     // "canned" is a canonical constant coined by the web client:
     //   https://github.com/zulip/zulip/blob/40f59a05c/web/shared/src/poll_data.ts#L238
     '${senderId ?? 'canned'},$idx';
@@ -204,6 +204,8 @@ enum PollEventSubmessageType {
   static final _byRawString = _$PollEventSubmessageTypeEnumMap
     .map((key, value) => MapEntry(value, key));
 }
+
+typedef PollOptionKey = String;
 
 /// A poll event when an option is added.
 ///
@@ -264,7 +266,7 @@ class PollVoteEventSubmessage extends PollEventSubmessage {
   /// The key of the affected option.
   ///
   /// See [PollEventSubmessage.optionKey].
-  final String key;
+  final PollOptionKey key;
   @JsonKey(name: 'vote', unknownEnumValue: PollVoteOp.unknown)
   final PollVoteOp op;
 
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -38,7 +38,7 @@ class Submessage {
   // We cannot parse the String into one of the [SubmessageData] classes because
   // information from other submessages are required. Specifically, we need:
   //   * the index of this submessage in [Message.submessages];
-  //   * the [WidgetType] of the first [Message.submessages].
+  //   * the parsed [WidgetType] from the first [Message.submessages].
   final String content;
 
   factory Submessage.fromJson(Map<String, Object?> json) =>
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'package:json_annotation/json_annotation.dart';
 
 import '../../log.dart';
+import 'events.dart';
 
 part 'submessage.g.dart';
 
@@ -401,6 +402,17 @@ class Poll {
   /// Contains the text of all options from [_options].
   final Set<String> _existingOptionTexts = {};
   final Map<PollOptionKey, PollOption> _options = {};
+
+  void handleSubmessageEvent(SubmessageEvent event) {
+    final PollEventSubmessage? pollEventSubmessage;
+    try {
+      pollEventSubmessage = PollEventSubmessage.fromJson(jsonDecode(event.content) as Map<String, Object?>);
+    } catch (e) {
+      assert(debugLog('Malformed submessage event data for poll: $e\n${jsonEncode(event)}')); // TODO(log)
+      return;
+    }
+    _applyEvent(event.senderId, pollEventSubmessage);
+  }
 
   void _applyEvent(int senderId, PollEventSubmessage event) {
     switch (event) {
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -1,4 +1,8 @@
+import 'dart:convert';
+
 import 'package:json_annotation/json_annotation.dart';
+
+import '../../log.dart';
 
 part 'submessage.g.dart';
 
@@ -40,6 +44,30 @@ class Submessage {
   //   * the index of this submessage in [Message.submessages];
   //   * the parsed [WidgetType] from the first [Message.submessages].
   final String content;
+
+  /// Parse a JSON list into a [Poll].
+  // TODO: Use a generalized return type when supporting other Zulip widgets.
+  static Poll? parseSubmessagesJson(List<Object?> json, {
+    required int messageSenderId,
+  }) {
+    final submessages = json.map((e) => Submessage.fromJson(e as Map<String, Object?>)).toList();
+    if (submessages.isEmpty) return null;
+
+    assert(submessages.first.senderId == messageSenderId);
+
+    final widgetData = WidgetData.fromJson(jsonDecode(submessages.first.content));
+    switch (widgetData) {
+      case PollWidgetData():
+        return Poll.fromSubmessages(
+          widgetData: widgetData,
+          pollEventSubmessages: submessages.skip(1),
+          messageSenderId: messageSenderId,
+        );
+      case UnsupportedWidgetData():
+        assert(debugLog('Unsupported widgetData: ${widgetData.json}'));
+        return null;
+    }
+  }
 
   factory Submessage.fromJson(Map<String, Object?> json) =>
     _$SubmessageFromJson(json);
@@ -318,4 +346,127 @@ class UnknownPollEventSubmessage extends PollEventSubmessage {
 
   @override
   Map<String, Object?> toJson() => json;
+}
+
+/// States of a poll Zulip widget.
+///
+/// See also:
+/// - https://zulip.com/help/create-a-poll
+/// - https://github.com/zulip/zulip/blob/304d948416465c1a085122af5d752f03d6797003/web/shared/src/poll_data.ts
+class Poll {
+  /// Construct a poll from submessages.
+  ///
+  /// For a poll Zulip widget, the first submessage's content contains a
+  /// [PollWidgetData], and all the following submessages' content each contains
+  /// a [PollEventSubmessage].
+  factory Poll.fromSubmessages({
+    required PollWidgetData widgetData,
+    required Iterable<Submessage> pollEventSubmessages,
+    required int messageSenderId,
+  }) {
+    final poll = Poll._(
+      messageSenderId: messageSenderId,
+      question: widgetData.extraData.question,
+      options: widgetData.extraData.options,
+    );
+
+    for (final submessage in pollEventSubmessages) {
+      final event = PollEventSubmessage.fromJson(jsonDecode(submessage.content) as Map<String, Object?>);
+      poll._applyEvent(submessage.senderId, event);
+    }
+    return poll;
+  }
+
+  Poll._({
+    required this.messageSenderId,
+    required this.question,
+    required List<String> options,
+  }) {
+    for (int index = 0; index < options.length; index += 1) {
+      // Initial poll options use a placeholder senderId.
+      // See [PollEventSubmessage.optionKey] for details.
+      _addOption(senderId: null, idx: index, option: options[index]);
+    }
+  }
+
+  final int messageSenderId;
+  String question;
+
+  /// The limit of options any single user can add to a poll.
+  ///
+  /// See https://github.com/zulip/zulip/blob/304d948416465c1a085122af5d752f03d6797003/web/shared/src/poll_data.ts#L69-L71
+  static const _maxIdx = 1000;
+
+  Iterable<PollOption> get options => _options.values;
+  /// Contains the text of all options from [_options].
+  final Set<String> _existingOptionTexts = {};
+  final Map<PollOptionKey, PollOption> _options = {};
+
+  void _applyEvent(int senderId, PollEventSubmessage event) {
+    switch (event) {
+      case PollNewOptionEventSubmessage():
+        _addOption(senderId: senderId, idx: event.idx, option: event.option);
+
+      case PollQuestionEventSubmessage():
+        if (senderId != messageSenderId) {
+          // Only the message owner can edit the question.
+          assert(debugLog('unexpected poll data: user $senderId is not allowed to edit the question')); // TODO(log)
+          return;
+        }
+
+        question = event.question;
+
+      case PollVoteEventSubmessage():
+        final option = _options[event.key];
+        if (option == null) {
+          assert(debugLog('vote for unknown key ${event.key}')); // TODO(log)
+          return;
+        }
+
+        switch (event.op) {
+          case PollVoteOp.add:
+            option.voters.add(senderId);
+          case PollVoteOp.remove:
+            option.voters.remove(senderId);
+          case PollVoteOp.unknown:
+            assert(debugLog('unknown vote op ${event.op}')); // TODO(log)
+        }
+
+      case UnknownPollEventSubmessage():
+    }
+  }
+
+  void _addOption({required int? senderId, required int idx, required String option}) {
+    if (idx > _maxIdx || idx < 0) return;
+
+    // The web client suppresses duplicate options, which can be created through
+    // the /poll command as there is no server-side validation.
+    if (_existingOptionTexts.contains(option)) return;
+
+    final key = PollEventSubmessage.optionKey(senderId: senderId, idx: idx);
+    assert(!_options.containsKey(key));
+    _options[key] = PollOption(text: option);
+    _existingOptionTexts.add(option);
+  }
+
+  static Poll? fromJson(Object? json) {
+    // [Submessage.parseSubmessagesJson] does all the heavy lifting for parsing.
+    return json as Poll?;
+  }
+
+  static List<Submessage> toJson(Poll? poll) {
+    // Rather than maintaining a up-to-date submessages list, return as if it is
+    // empty, because we are not sending the submessages to the server anyway.
+    return [];
+  }
+}
+
+class PollOption {
+  PollOption({required this.text});
+
+  final String text;
+  final Set<int> voters = {};
+
+  @override
+  String toString() => 'PollOption(text: $text, voters: {${voters.join(', ')}})';
 }
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -1,3 +1,5 @@
+import 'dart:convert';
+
 import '../api/model/events.dart';
 import '../api/model/model.dart';
 import '../log.dart';
@@ -305,6 +307,23 @@ class MessageStoreImpl with MessageStore {
           userId: event.userId,
         );
     }
+
+    for (final view in _messageListViews) {
+      view.notifyListenersIfMessagePresent(event.messageId);
+    }
+  }
+
+  void handleSubmessageEvent(SubmessageEvent event) {
+    final message = messages[event.messageId];
+    if (message == null) return;
+
+    final poll = message.poll;
+    if (poll == null) {
+      assert(debugLog('Missing poll for submessage event:\n${jsonEncode(event)}')); // TODO(log)
+      return;
+    }
+
+    poll.handleSubmessageEvent(event);
 
     for (final view in _messageListViews) {
       view.notifyListenersIfMessagePresent(event.messageId);
--- a/test/api/model/submessage_test.dart
+++ b/test/api/model/submessage_test.dart
@@ -3,6 +3,7 @@ import 'package:test/scaffolding.dart';
 import 'package:zulip/api/model/submessage.dart';
 
 import '../../example_data.dart' as eg;
+import '../../stdlib_checks.dart';
 import 'submessage_checks.dart';
 
 void main() {
@@ -27,23 +28,25 @@ void main() {
     });
   });
 
-  test('smoke WidgetData',  () {
-    check(WidgetData.fromJson(eg.pollWidgetDataFavoriteLetter)).isA<PollWidgetData>()
+  test('invalid widget_type -> UnsupportedWidgetData/throw', () {
+    final pollWidgetData = deepToJson(eg.pollWidgetData(
+      question: 'example question',
+      options: ['A', 'B', 'C'],
+    )) as Map<String, Object?>;
+
+    check(WidgetData.fromJson(pollWidgetData)).isA<PollWidgetData>()
       ..widgetType.equals(WidgetType.poll)
       ..extraData.which((x) => x
-          ..question.equals('favorite letter')
+          ..question.equals('example question')
           ..options.deepEquals(['A', 'B', 'C'])
         );
-  });
-
-  test('invalid widget_type -> UnsupportedWidgetData/throw', () {
     check(WidgetData.fromJson({
-      ...eg.pollWidgetDataFavoriteLetter,
+      ...pollWidgetData,
       'widget_type': 'unknown_foo',
     })).isA<UnsupportedWidgetData>();
 
     check(() => WidgetData.fromJson({
-      ...eg.pollWidgetDataFavoriteLetter,
+      ...pollWidgetData,
       'widget_type': 123,
     })).throws<TypeError>();
   });
--- a/test/api/model/submessage_test.dart
+++ b/test/api/model/submessage_test.dart
@@ -114,4 +114,7 @@ void main() {
       'key': PollEventSubmessage.optionKey(senderId: null, idx: 0)
     })).isA<PollVoteEventSubmessage>().op.equals(PollVoteOp.unknown);
   });
+
+  // Parsing polls with PollEventSubmessages are tested in
+  // `test/model/message_test.dart` in the "handleSubmessageEvent" test.
 }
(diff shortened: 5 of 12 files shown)


gnprice:
For ease of cross-reference: the previous PR #823 originally contained these changes too, and had a round of UX review starting with these screenshots:
https://github.com/zulip/zulip-flutter/pull/823#issuecomment-2266101540

PIG208:
Moving this to integration review because these changes went through maintainer review in the original PR.

lib/model/message.dart
```diff
@@ -302,4 +305,24 @@ class MessageStoreImpl with MessageStore {
       view.notifyListenersIfMessagePresent(event.messageId);
     }
   }
+
+  void handleSubmessageEvent(SubmessageEvent event) {
+    final poll = messages[event.messageId]?.poll;
+    if (poll == null) return;
```
gnprice, line 311: If the message is missing, that's normal. But if the message is present and not a poll, something's wrong. Let's have separate `if … return;` lines for those, with `// TODO(log)` on the latter.

lib/model/message.dart
```diff
@@ -302,4 +305,24 @@ class MessageStoreImpl with MessageStore {
...
+    PollEventSubmessage? pollEventSubmessage;
+    try {
+      pollEventSubmessage = PollEventSubmessage.fromJson(jsonDecode(event.content) as Map<String, Object?>);
+    } on TypeError catch (e) {
+      assert(debugLog('Malformed submessage event data for poll: $e\n${jsonEncode(event)}')); // TODO(log)
+      return;
+    }
```
gnprice, line 318: Let's move these details to live on `Poll` — that seems like the natural home for them.

Then the signature of that method can be `void handleSubmessageEvent(SubmessageEvent event)`, just like its caller here, and following our usual convention for methods that handle events.

test/example_data.dart
```diff
@@ -658,6 +675,21 @@ ReactionEvent reactionEvent(Reaction reaction, ReactionOp op, int messageId) {
   );
 }
 
+SubmessageEvent submessageEvent(
+  int messageId,
+  int senderId,
+  {required String content}
+) {
```
gnprice, line 682: nit: follow our standard formatting for signatures that have both positional and named parameters:
```suggestion
  int senderId, {
  required String content,
}) {
```

(here and a few other places)

test/example_data.dart
```diff
@@ -658,6 +675,21 @@ ReactionEvent reactionEvent(Reaction reaction, ReactionOp op, int messageId) {
...
+  return SubmessageEvent(
+    id: 1,
```
gnprice, line 684: id 0 — this is basically a subtle rebase error :slightly_smiling_face:

test/example_data.dart
```diff
@@ -658,6 +675,21 @@ ReactionEvent reactionEvent(Reaction reaction, ReactionOp op, int messageId) {
   );
 }
 
+SubmessageEvent submessageEvent(
```
gnprice, line 678: nit: order here should match order the classes are declared

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
         .reactions.isNotNull().jsonEquals([eg.unicodeEmojiReaction]);
     });
   });
+
+  group('handleSubmessageEvent', () {
```
gnprice, line 607: nit: order of tests should match order of code under test

(which in this case should, and does, match the order the event classes are defined)

test/model/message_test.dart
```diff
@@ -52,16 +57,38 @@ void main() {
...
     await messageList.fetchInitial();
     checkNotifiedOnce();
   }
 
+  /// Perform a single-message initial message fetch for [messageList] with
+  /// submessages.
+  ///
+  /// The test case must have already called [prepare] to initialize the state.
+  Future<void> prepareMessageWithSubmessage(
+    Message message, List<Submessage> submessages, {
```
gnprice, line 79: nit:
```suggestion
  Future<void> prepareMessageWithSubmessages(
    Message message, List<Submessage> submessages, {
```

since it takes a list of submessages, not just one

gnprice, line 79: nit:
```suggestion
  Future<void> prepareMessageWithSubmessages(
    Message message, List<Submessage> submessages, {
```

since it takes a list of submessages, not just one

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+    late StreamMessage streamMessage;
+
+    Future<int> preparePollMessage({List<PollOption>? options}) async {
+      List<PollOption> effectiveOptions = options
+        ?? [PollOption(text: 'foo'), PollOption(text: 'bar')];
+      const messageId = 123;
+      streamMessage = eg.streamMessage(
+        id: messageId,
+        sender: eg.selfUser,
+      );
```
gnprice, line 617: nit: fold onto one line

In general, since I'm seeing several cases of this in this PR: using four lines for something that can naturally fit on one is expensive in terms of how much can fit on the screen, and therefore of how far one can look around at surrounding code without scrolling back and forth. So it's good to keep things on fewer lines when they fit easily without cramping.

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+          senderId: eg.selfUser.userId,
+          content: jsonEncode(PollWidgetData(extraData: PollWidgetExtraData(
+            question: 'Favorite word',
+            options: effectiveOptions.map((e) => e.text).toList(),
+          )),
+        )),
+        for (int i = 0; i < effectiveOptions.length; i++)
+          ...[
+            for (final voter in effectiveOptions[i].voters)
+            eg.submessage(
```
gnprice, line 629: nit: missing indentation

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+              senderId: voter,
+              content: jsonEncode(PollVoteEventSubmessage(
+                key: PollEventSubmessage.optionKey(senderId: null, idx: i),
+                op: PollVoteOp.add,
+              )),
+            )
+          ]
```
gnprice, line 636: nit: missing trailing comma; see other comment above

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+        ?? [PollOption(text: 'foo'), PollOption(text: 'bar')];
+      const messageId = 123;
+      streamMessage = eg.streamMessage(
+        id: messageId,
+        sender: eg.selfUser,
+      );
+      final submessages = [
+        eg.submessage(
+          senderId: eg.selfUser.userId,
+          content: jsonEncode(PollWidgetData(extraData: PollWidgetExtraData(
```
gnprice, line 621: This can also use `eg.pollWidgetData`, once that's added (per a comment above)

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+
+  group('handleSubmessageEvent', () {
+    late StreamMessage streamMessage;
+
+    Future<int> preparePollMessage({List<PollOption>? options}) async {
+      List<PollOption> effectiveOptions = options
+        ?? [PollOption(text: 'foo'), PollOption(text: 'bar')];
+      const messageId = 123;
+      streamMessage = eg.streamMessage(
+        id: messageId,
```
gnprice, line 615: Similar to a comment above: is this needed? See the `eg.streamMessage` doc.

gnprice, line 615: Similar to a comment above: is this needed? See the `eg.streamMessage` doc.

gnprice, line 615: Similar to a comment above: is this needed? See the `eg.streamMessage` doc.

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+    test('message is unknown', () async {
+      await prepare();
+      await store.handleEvent(
+        eg.submessageEvent(1000, eg.otherUser.userId, content: '{}')
+      );
+      checkNotNotified();
+    });
+
+    test('message has no submessages', () async {
+      final streamMessage = eg.streamMessage(id: 123);
```
gnprice, line 652: nit:
```suggestion
      final message = eg.streamMessage(id: 123);
```

The fact it's a stream message isn't important, right? So can leave that out of the story this tells to the reader. (It's still there in the `eg.streamMessage` call, but gets less emphasis.)

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
...
+      check(store.messages[messageId]).isNotNull().poll.isNotNull();
+    });
+
+    test('update question', () async {
+      final messageId = await preparePollMessage();
+      await store.handleEvent(
+        eg.submessageEvent(
+          streamMessage.id,
+          eg.selfUser.userId,
+          content: jsonEncode(PollQuestionEventSubmessage(question: 'New question'))));
```
gnprice, line 688:
```suggestion
      final messageId = await preparePollMessage(question: 'Old question');
      await store.handleEvent(
        eg.submessageEvent(
          streamMessage.id,
          eg.selfUser.userId,
          content: jsonEncode(PollQuestionEventSubmessage(question: 'New question'))));
```

This way the test case is self-contained for the reader — the contrast between the original question value and the new question value is explicit within the test case's own source code.

(Similar to my comment above about `eg.pollWidgetDataFavoriteLetter`.)

test/model/message_test.dart
```diff
@@ -576,4 +603,197 @@ void main() {
         .reactions.isNotNull().jsonEquals([eg.unicodeEmojiReaction]);
     });
   });
+
+  group('handleSubmessageEvent', () {
+    late StreamMessage streamMessage;
```
gnprice, line 608:
```suggestion
    late Message message;
```

Similar to my preceding comment (on code a bit below this) — the detail that it's a stream message doesn't matter to the narrative, so can be left out.

test/test_log.dart
```diff
@@ -0,0 +1,15 @@
...
+
+import 'package:checks/checks.dart';
+import 'package:zulip/log.dart';
+
+Future<Subject<List<String>>> checkLogs(FutureOr<void> Function() callback) async {
+  assert(logHistory == null);
+  logHistory = [];
+  try {
+    await callback();
+    return check(logHistory).isA<List<String>>();
```
gnprice, line 11:
```suggestion
    return check(logHistory!);
```

If `logHistory` is back to null at this point, that seems like not just a normal failure of the calling test case — it means something that shouldn't be messing with that variable has a surprising bug where it's doing so anyway. So we can avoid cluttering test failure messages with a line about how the log history was expected to be (and is) a `List<String>`.

lib/log.dart
```diff
@@ -1,9 +1,14 @@
+import 'package:flutter/foundation.dart';
 
 /// Whether [debugLog] should do anything.
 ///
 /// This has an effect only in a debug build.
 bool debugLogEnabled = false;
 
+/// Used for log inspection in tests.
+@visibleForTesting
+List<String>? logHistory;
```
gnprice, line 10:
```suggestion
List<String>? debugLogHistory;
```

This will live in our global namespace (e.g. it'll appear in autocomplete from all over the app) so the "debug" prefix is important context.

gnprice (commented):
Thanks! Here's a review focused on the first two commits, which set up the model:
144d5afea api: Construct polls data store from submessages.
838893683 event: Handle submessage event for polls.

I didn't read through all of the tests, because I left some high-level comments there that I think will make some of them easier to read.

Left for a future round are the remaining commits:
6cc7d0d44 (optional) test: Add a test helper for inspecting logs.
3a0f853b1 poll: Support read-only poll widget UI.

lib/api/model/submessage.dart
```diff
@@ -313,3 +318,129 @@ class UnknownPollEventSubmessage extends PollEventSubmessage {
...
+      // Initial poll options use a placeholder senderId.
+      // See [PollEventSubmessage.optionKey] for details.
+      _addOption(null, options[index], idx: index);
+    }
+  }
+
+  factory Poll.fromSubmessages(
+    List<Submessage> submessages,
+    {required int senderId, required PollWidgetData widgetData}
+  ) {
```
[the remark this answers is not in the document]
PIG208: Because finding out which widget the message contains requires the caller to parse `WidgetData` first, they are asked to include `widgetData` just to avoid parsing it.

`senderId` is the sender ID of the enclosing message.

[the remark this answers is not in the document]
PIG208: I think a way to make it clearer is to ask the caller to include only the supposed PollEventSubmessage, by skipping the first submessage on the message, and make them all named:

```suggestion
  factory Poll.fromSubmessages({
    required Iterable<Submessage> pollEventSubmessages,
    required int messageSenderId,
    required PollWidgetData widgetData
  }) {
```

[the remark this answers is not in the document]
gnprice: Sounds good.

Let's order them logically, too: message sender ID first since the message encloses the whole thing, and widget data before event submessages.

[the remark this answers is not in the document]
gnprice: > it seems like it's assumed that it includes the [PollWidgetData]. Probably cleaner to have it not include that.

(And because I see this was ambiguous: what I intended here was to have the `submessages` argument not include the [PollWidgetData].)

lib/api/model/submessage.dart
```diff
@@ -313,3 +318,129 @@ class UnknownPollEventSubmessage extends PollEventSubmessage {
...
+
+  Iterable<String> getVoterNames(PerAccountStore store, String fallback) =>
+    voters.map((userId) => store.users[userId]?.fullName ?? fallback);
+
+  @override
+  bool operator ==(Object other) {
+    if (other is! PollOption) return false;
+
+    return other.hashCode == hashCode;
+  }
```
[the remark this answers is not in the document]
PIG208: This override is used in tests. (e.g. places where `defaultOptions` is used)

test/api/model/submessage_test.dart
```diff
@@ -111,4 +116,179 @@ void main() {
...
+    Message messageWithSubmessageContents(
+      Map<String, Object?> content,
+      {List<(int senderId, PollEventSubmessage event)>? events}
+    ) {
+      return eg.streamMessage(
+        id: 123,
+        sender: eg.otherUser,
+        submessages: [
+          eg.submessage(
+            content: jsonEncode(content),
```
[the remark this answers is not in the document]
PIG208: This point applies to `eg.submessageEvent` too I think.

test/api/model/submessage_test.dart
```diff
@@ -111,4 +116,179 @@ void main() {
...
+                senderId: senderId,
+              )
+        ]);
+    }
+
+    test('parse poll from message', () {
+      check(messageWithSubmessageContents(eg.pollWidgetDataFavoriteLetter))
+        .poll.isNotNull()
+          ..question.equals('favorite letter')
+          ..options.deepEquals(defaultOptions);
```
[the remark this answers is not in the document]
PIG208: Those are some great insights. Thanks!

lib/model/message.dart
```diff
@@ -302,4 +305,24 @@ class MessageStoreImpl with MessageStore {
...
     }
   }
+
+  void handleSubmessageEvent(SubmessageEvent event) {
+    final poll = messages[event.messageId]?.poll;
+    if (poll == null) return;
+    PollEventSubmessage? pollEventSubmessage;
+    try {
+      pollEventSubmessage = PollEventSubmessage.fromJson(jsonDecode(event.content) as Map<String, Object?>);
+    } on TypeError catch (e) {
```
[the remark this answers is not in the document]
PIG208: That sounds very reasonable. Spliting the `jsonDecode` call out to its own `try` block, and logging each of them.

test/model/message_test.dart
PIG208: I think `handleDeleteMessageEvent` is out of order, will move that in a prep commit.

test/model/message_test.dart
PIG208: That's a great catch. I find it to be quite a good demonstration.

lib/api/model/submessage.dart
```diff
@@ -313,3 +318,129 @@ class UnknownPollEventSubmessage extends PollEventSubmessage {
...
+  String question;
+
+  /// The limit of options any single user can add to a poll.
+  ///
+  /// See https://github.com/zulip/zulip/blob/304d948416465c1a085122af5d752f03d6797003/web/shared/src/poll_data.ts#L69-L71
+  static const maxIdx = 1000; // TODO validate
+
+  Iterable<PollOption> get options => _options.values;
+  final Set<String> _optionNames = {};
+  final Map<String, PollOption> _options = {};
```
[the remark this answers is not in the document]
PIG208: Extracted a `OptionKey` type alias to make it clearer that `_options`'s keys are keys.
Also renamed `_optionNames` to `_existingOptionTexts` with a new comment.

lib/api/model/model.dart
```diff
@@ -512,6 +519,45 @@ sealed class Message {
...
   }
 
+  static List<Submessage>? _submessagesFromJson(Object? json) {
+    final list = json as List<Object?>;
+    return list.isNotEmpty
+      ? list.map((e) => Submessage.fromJson(e as Map<String, Object?>)).toList()
+      : null;
+  }
+
+  static Poll? _readPoll(Map<Object?, Object?> json, String key) {
```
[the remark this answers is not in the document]
PIG208: Moved them to `[Submessage.parseSubmessagesJson]`, where we can also parse other types of Zulip widget in the future.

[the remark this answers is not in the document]
PIG208: Moved fromJson and toJson to `Poll`. `toJson` remains as a static method, and `_readPoll` is kept at where it was (now it just calls `Submessage.parseSubmessagesJson`).

PIG208:
Thanks for the review @gnprice! I have addressed them and updated the PR.

lib/api/model/submessage.dart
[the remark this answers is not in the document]
gnprice: I see. This override is still incorrect, though, for the reason I gave in the first comment (even in this revision which has fixed the secondary issue in the third comment). See the doc on `Object.==`.

It may work in tests because the right-hand side of the comparison is a throwaway that won't be mutated later… but an override here isn't restricted to use in tests, so having an incorrect `==` implementation is dangerous as a source of future bugs.

Looking at those tests, you can express those checks without a `==` override by using  `.deepEquals(<Condition<Object?>>[ … ])`. See `test/notifications/display_test.dart` for an example, in the `checkNotification` helper. See also #878 (not yet merged) for further examples in a slightly different style, with the `conditionActiveNotif` and `conditionSummaryActiveNotif` helpers.

lib/api/model/model.dart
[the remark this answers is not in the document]
gnprice: Cool. Yeah, the new nicely-short `_readPoll` makes sense to keep on `Message`, as it's expressing the details that are about property names on message objects.

lib/api/model/model.g.dart
```diff
@@ -288,7 +288,9 @@ StreamMessage _$StreamMessageFromJson(Map<String, dynamic> json) {
     matchTopic: json['match_subject'] as String?,
     displayRecipient: json['display_recipient'] as String?,
     streamId: (json['stream_id'] as num).toInt(),
-  );
+  )..poll = Message._readPoll(json, 'submessages') == null
+      ? null
+      : Poll.fromJson(Message._readPoll(json, 'submessages'));
```
gnprice, line 293: Ow, this is inefficient. (Since `_readPoll` includes all the work of actually parsing the JSON.)

If there's not a quick and clean way to rearrange things to avoid this double parsing, that's fine and it shouldn't block merging this — polls aren't *that* common so it's not a major performance issue. But in that case let's leave a TODO comment about it at `Message.poll`.

test/api/model/submessage_test.dart
```diff
@@ -111,4 +116,176 @@ void main() {
...
+        submessages: [
+          eg.submessage(
+            content: content,
+            senderId: eg.otherUser.userId,
+          ),
+          for (final (senderId, event) in events)
+            eg.submessage(
+              content: event,
+              senderId: senderId,
+            ),
```
gnprice, line 146: nit: fold onto one line

test/api/model/submessage_test.dart
```diff
@@ -111,4 +116,176 @@ void main() {
...
+    });
+
+    test('parse poll with vote events on post-creation options', () {
+      check(messageWithSubmessageContents(
+        pollWidgetData,
+        events: [
+          (eg.otherUser.userId, PollNewOptionEventSubmessage(idx: 0, option: 'D')),
+          (eg.otherUser.userId, PollVoteEventSubmessage(key: '${eg.otherUser.userId},0', op: PollVoteOp.add)),
+          (eg.otherUser.userId, PollVoteEventSubmessage(key: '${eg.selfUser.userId},0', op: PollVoteOp.add)),
+          (eg.otherUser.userId, PollVoteEventSubmessage(key: '${eg.selfUser.userId},0', op: PollVoteOp.remove)),
```
gnprice, line 212: Does this latter key correspond to an option? It looks like there's the initial options, then one created by eg.otherUser, and none by eg.selfUser.

test/api/model/submessage_test.dart
```diff
@@ -111,4 +116,176 @@ void main() {
...
+            eg.pollOption(text: 'D', voters: [eg.otherUser.userId]),
+          ]);
+    });
+
+    test('applyEvent: adding repeated option', () {
+      final poll = messageWithSubmessageContents(pollWidgetData).poll!;
+      check(poll.options).deepEquals(defaultOptions);
+      poll.applyEvent(eg.otherUser.userId, PollNewOptionEventSubmessage(
+        option: defaultOptions[0].text, idx: 0));
+      check(poll.options).deepEquals(defaultOptions);
```
gnprice, line 228: In either the name or a one-line comment within the test, say explicitly how the story ends: IIUC, the answer is that the event gets ignored.

gnprice: This would do it:
```suggestion
    test('applyEvent: option with duplicate text ignored', () {
      final poll = messageWithSubmessageContents(pollWidgetData).poll!;
      check(poll.options).deepEquals(defaultOptions);
      poll.applyEvent(eg.otherUser.userId, PollNewOptionEventSubmessage(
        option: defaultOptions[0].text, idx: 0));
      check(poll.options).deepEquals(defaultOptions);
```

test/api/model/submessage_test.dart
```diff
@@ -111,4 +116,176 @@ void main() {
...
+      ))
+        .poll.isNotNull()
+          ..question.equals('favorite letter')
+          ..options.deepEquals([
+            ...defaultOptions,
+            eg.pollOption(text: 'D', voters: [eg.otherUser.userId]),
+          ]);
+    });
+
+    test('applyEvent: adding repeated option', () {
```
gnprice, line 223: Having these test cases for error paths is great. Let's also have test cases for the normal happy paths, though. As a bonus, those can make the error tests clearer because they serve as a baseline for the reader to compare to.

As is, if I'm reading right it looks like these tests would all pass if `applyEvent` just always did nothing :slightly_smiling_face:

gnprice: Hmm I guess the tests in `lib/model/message_test.dart` end up covering those happy paths.

I think it's clearest if the success cases and error cases have their tests next to each other — as I said above, the success tests make a helpful baseline for reading the error tests, and also the reader of the success tests may get nervous if it looks like there are no tests for the error cases.

I like the layer that the `lib/model/message_test.dart` tests are expressed at: calling `store.handleEvent` with a Zulip server event, just as if the event had come in from the server in the polling loop. That's how we try to write most of our model tests, following this principle:

So let's try to move these error tests to that file and convert them to that file's style.

In fact let's go a step farther: let's move even the "parse poll …" test cases the same way. Then this whole "Message.poll" test group gets replaced by a comment saying the tests for deserializing [Message.poll] can be found over there.

Really what's going on here is that for this data structure, we've fused the job of deserializing it with the job of setting up a nontrivial data structure, including applying some significant model logic to apply all those poll-event submessages. So the deserialization code has a lot of the nature of model code, and it's fitting if its tests therefore wind up in the model code's test files. And then the tests for later events coming in definitely belong on the model side (as discussed above), and they're exercising most of the same code as the deserialization… so the deserialization tests belong next to the later-events tests, in `test/model/`.

lib/api/model/submessage.dart
```diff
@@ -313,3 +346,158 @@ class UnknownPollEventSubmessage extends PollEventSubmessage {
...
+      return;
+    }
+
+    final PollEventSubmessage? pollEventSubmessage;
+    try {
+      pollEventSubmessage = PollEventSubmessage.fromJson(decodedContent as Map<String, Object?>);
+    } catch (e) {
+      assert(debugLog('Malformed submessage event data for poll: $e\n${jsonEncode(event)}')); // TODO(log)
+      return;
+    }
```
gnprice, line 419: Oh one more comment: we can simplify this a bit by making this just one try/catch that doesn't distinguish between whether it was the `jsonDecode` or the `fromJson` call that threw. Conceptually those are two phases of a single operation anyway: deserialize a string into a PollEventSubmessage.

Similarly if you walk through the logic in ApiConnection.send, I'm pretty sure the behavior there is that both kinds of failures throw very similar exceptions (instances of MalformedServerResponseException). So it's not worth getting more fine-grained here than we are there, the main place we deserialize JSON from the server.

(following up on https://github.com/zulip/zulip-flutter/pull/885#discussion_r1716404535 )

lib/api/model/submessage.dart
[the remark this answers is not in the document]
PIG208: Thanks for the pointer. It makes sense to skip the override because they are only used in test. https://github.com/zulip/zulip-flutter/pull/885#discussion_r1716261591.

Also started some further [discussion](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Object.2E.3D.3D.20behavior) on `Object.==` in general.

PIG208:
Partially updated. The tests reorganization is WIP.

lib/api/model/submessage.dart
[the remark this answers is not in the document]
gnprice: > It makes sense to skip the override because they are only used in test. [#885 (comment)](https://github.com/zulip/zulip-flutter/pull/885#discussion_r1716261591).

Yep, that's the subthread I was replying to :slightly_smiling_face: — this is a bit of a quirky aspect of GitHub's interface, so I just wrote up some tips here:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/github.20subthread.20structure/near/1923632

PIG208:
The tests in `Message.poll` up ended rewritten in two destinations. Here's a broad overview:

- "new message with initial submessages" under "handleMessageEvent":

  Contains all the "parse poll" tests. Rewritten so that we handle new `MessageEvent`s for polls. Because these tests are about processing the initial state of the message, there is no `SubmessageEvent`s to handle.

- "handleSubmessageEvent":

  Contains all the "applyEvent" tests. Each of them got moved to the corresponding `PollEventSubmessageType` group.

Extra note: dropped the "no poll if submessages is empty" test because it is a duplicate of the existing "message has no submessages" test.

test/model/message_test.dart
```diff
@@ -473,36 +634,220 @@ void main() {
...
+        await checkLogs(() => store.handleEvent(SubmessageEvent(
+          id: 0, msgType: SubmessageType.widget, submessageId: 123,
+          messageId: message.id,
+          senderId: eg.selfUser.userId,
+          content: jsonEncode({
+            'type': 'question',
+            // Invalid type for question
+            'question': 100,
+          }))))..length.equals(2)
+               ..last.contains('Malformed submessage event data for poll');
```
gnprice, line 715: So, thinking about this `checkLogs` sort of test check now that I've reached this commit:

This feels to me like the sort of thing that's going to be quite brittle. (E.g. why length 2 — I guess the first is the one in `PerAccountStore.handleEvent`?) If we start widely adopting this pattern, then any time we add or remove a `debugLog` call there's going to be a bunch of random tests all over that need to be updated for it.

Also when we change the content of a `debugLog` call, we'll have to update the corresponding test. That's OK for a lot of things (if it's really the test aimed at the thing one is changing), but it makes changes more heavyweight. And these log messages are internal details, in the sense that they're aimed only at us the developers — they're not something our users, or any external system, ever see.

If we find a place where there's a `debugLog` call that we feel is really critical to ensure doesn't accidentally get broken, then that could be a good use case for something like this despite those costs. But I think this `debugLog` call is typical in that it isn't critical to that degree. Let's therefore leave this check out.

gnprice: (As for the server tests:

    This is inspired by `TestCase.assertLogs`, which is widely adapted
    in the Zulip server tests.

I have the sense that a lot of those uses are motivated mostly by the need to *suppress* the logs so they don't spew into the output of running the tests. (That's especially clear when they don't end up asserting anything about the logs emitted; and I suspect it's the only important motivation in many tests that do assert something about them too.) In this codebase we just handle that systematically using `debugLogEnabled`.

The other motivation for `assertLogs` checks in the server is that in the server, those log messages go into the production server log, which is an important operational tool for Zulip server operators including ourselves. That has somewhat the same shape as our `debugLog` calls for debugging… with the crucial difference that when we see `debugLog` output, the app is a debug build, and so we're presumably sitting there debugging it and can easily add more log lines as we wish and then hot-reload and repeat whatever we did. That's very different from the scenario where something went wrong an hour ago on one's live production server and you're frantically trying to pin down what happened.)

gnprice (commented):
Thanks for the revision! Here's a next round, covering all the first N-1 commits:

2bfbff0e3 message test [nfc]: Move handleDeleteMessageEvent
ed5f68ec4 submessage: Extract OptionKey.
2d8f11e9c submessage [nfc]: Minor wording change to a comment.
ae9f87cef submessage test [nfc]: Extract pollWidgetDataJson.
a85dbe4d6 api: Construct polls data store from submessages.
727b9d82c event: Handle submessage event for polls.
b00a64f44 (optional) test: Add a test helper for inspecting logs.

and leaving only the last commit for a future round:
bf599f2f2 poll: Support read-only poll widget UI.

lib/api/model/submessage.dart
```diff
@@ -313,3 +349,144 @@ class UnknownPollEventSubmessage extends PollEventSubmessage {
   @override
   Map<String, Object?> toJson() => json;
 }
+
+/// States of a poll Zulip widget.
+///
+/// See also:
+/// - https://zulip.com/help/create-a-poll
+/// - https://github.com/zulip/zulip/blob/304d948416465c1a085122af5d752f03d6797003/web/shared/src/poll_data.ts
+class Poll {
```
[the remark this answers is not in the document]
PIG208: I have been inclined to squash the `handleSubmessageEvent` commit with this commit since last update, because all the tests here would be moved to the later commit.

An alternative would be retrofitting the updated tests in the earlier commit with as little change as necessary (to make the diffs look better), then moving it. This way we have atomicity in both commits.

I think it makes sense to leave the isolated exercises of `_applyEvent` in the later commit that introduces submessage events, because that's when we have a public interface to access `_applyEvent`.

lib/api/model/submessage.dart
```diff
@@ -38,9 +44,33 @@ class Submessage {
   // We cannot parse the String into one of the [SubmessageData] classes because
   // information from other submessages are required. Specifically, we need:
   //   * the index of this submessage in [Message.submessages];
-  //   * the [WidgetType] of the first [Message.submessages].
+  //   * the parsed [WidgetType] from the first [Message.submessages].
   final String content;
 
+  /// Parse a JSON list into a [Poll].
+  // It will start returning `Poll` as we support more Zulip widgets.
+  static Poll? parseSubmessagesJson(List<Object?> json, {
```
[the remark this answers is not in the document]
PIG208: The `Poll` in the comment was unintended. Made it a TODO instead:
```
// TODO: Use a generalized return type when supporting other Zulip widgets.
```

test/model/message_test.dart
PIG208: Thanks for the explanation! Dropping the commit.

Another motivation for adding `checkLogs` was to have a convenient way of verifying we expect to happen in our code -- too convenient that it relies heavily on implementation details. I agree that testing externally visible interfaces are preferred.

PIG208:
I have updated the PR. In this update:
- added some tests back to the commit that introduces `Poll`, before moving them later;
- rearrange test grouping.

lib/widgets/poll.dart
```diff
@@ -0,0 +1,90 @@
...
+    final store = PerAccountStoreWidget.of(context);
+    final theme = ContentTheme.of(context);
+
+    final textStylePollPrimary = const TextStyle(fontSize: 16)
+      .merge(weightVariableTextStyle(context, wght: 600));
+    final textStylePollSecondary = TextStyle(
+      fontSize: 16, color: theme.colorPollNames);
+
+    String getNameByUserId(int userId) =>
+        store.users[userId]?.fullName ?? zulipLocalizations.unknownUserName;
```
gnprice, line 26: Ah, when at https://github.com/zulip/zulip-flutter/pull/885#discussion_r1724086159 I mentioned a local variable instead of a method:
> It can probably just be a local variable in that build method, not a method or anything.

I was thinking of the result of `map` and `join` as a local variable.

I think that'd be a good further tweak — the widget expression below has a lot of details of layout, so it'd be good to move the more interesting computations out of it to happen above.

lib/api/model/submessage.dart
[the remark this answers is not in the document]
gnprice: I see, in the new revision this earlier commit adds tests in `test/api/model/submessage_test.dart`, and then the later commit moves them and adapts them to `test/model/message_test.dart`.

I don't really understand why the tests can't be in the latter form already in the earlier commit. The polls are there on the messages in the message store, right? So the tests that find them in the message store should work fine.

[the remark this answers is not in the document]
gnprice: If we really would have to move around and rewrite the tests like that, then it'd be better to squash the commits.

gnprice (commented):
Thanks for the revision! I haven't yet read the final commit closely:
42c9c1086 poll: Support read-only poll widget UI.

but the rest all generally look good. A few comments below.

I'll also push a revision that I think demonstrates the main comment below, about the tests — please take a look and see if I'm missing something.

test/model/message_test.dart
```diff
@@ -576,4 +584,346 @@ void main() {
         .reactions.isNotNull().jsonEquals([eg.unicodeEmojiReaction]);
     });
   });
+
+  group('handleSubmessageEvent', () {
```
gnprice, line 588: Oh, one other nit:

This group isn't specific to `handleSubmessageEvent` — it's about interpreting Message objects that come in the first place with submessages, too, where `SubmessageEvent` may never be involved.

lib/api/model/submessage.dart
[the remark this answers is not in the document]
PIG208: Yeah, I think I overlooked the fact that those `MessageEvent`-based tests are not blocked by `SubmessageEvent`. My plan in an earlier version of this revision was to reintroduce all tests, including the `applyEvent` ones, in their rewritten form without event handling. I agree that leaving `submessage_test.dart` out from the very beginning is a better decision.

test/model/message_test.dart
PIG208: Gave the outer group a different name: "handle Poll related events". The name `handleSubmessageEvent` should be applicable to the subgroup of tests that do take care of `SubmessageEvent`. To parallel that, I renamed the other subgroup to "handleMessageEvent with initial submessages"

PIG208:
Pushed a minor formatting update.

gnprice:
Thanks for the revision! Those changes all look good. All the commits before the last one are now ready, so I'm going to go ahead and merge them.

I won't make it to a review of the last commit today, though, before I go on vacation:
798f1b36f poll: Support read-only poll widget UI.

So please send that as a fresh PR. Let's also have @chrisbobbe do the first review on that — I know a similar commit was in #823 which he reviewed before, but it's been a bit of a marathon series of changes :slightly_smiling_face: (which is natural because this is a complex feature), so he may have new comments to make when looking at it with fresh eyes.

PR #886, opened 2024-08-12
compose_box: Replace compose box with a banner when cannot post in a channel

sm-sayedi:
When a user cannot post in a channel based on `ZulipStream.channelPostPolicy`, all parts of the compose box (in both channel and topic narrow) are replaced with a banner, saying: _You do not have permission to post in this channel_.

### Screen recording

https://github.com/user-attachments/assets/c599db3c-755f-41b6-b119-0d6fc931f701

Fixes: #674


rajveermalviya (approved):
Thanks for working on this @sm-sayedi, this looks great!

Moving on to the mentor review from @hackerkid.

assets/l10n/app_en.arb
```diff
@@ -188,6 +188,10 @@
   "@errorBannerDeactivatedDmLabel": {
     "description": "Label text for error banner when sending a message to one or multiple deactivated users."
   },
+  "errorBannerCannotPostInChannelLabel": "You do not have permission to post in this channel.",
+  "@errorBannerCannotPostInChannelLabel": {
+    "description": "Label text for error banner when sending a message in a channel with no posting permission."
```
chrisbobbe, line 193:
```suggestion
    "description": "Error-banner text replacing the compose box when you do not have permission to send a message to the channel."
```

lib/widgets/compose_box.dart
```diff
@@ -999,8 +999,21 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> implements Compose
     super.dispose();
   }
 
+  Widget? _errorBanner(BuildContext context) {
```
chrisbobbe, line 1002: There's already an `_errorBanner` method, in `_FixedDestinationComposeBoxState`, and it duplicates a lot of this logic. Can we centralize the computation, perhaps in an early-return style in the build method of `ComposeBox`?

test/widgets/action_sheet_test.dart
```diff
@@ -43,7 +43,7 @@ Future<void> setupToMessageActionSheet(WidgetTester tester, {
 
   await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
   final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
-  await store.addUser(eg.user(userId: message.senderId));
+  await store.addUsers([eg.selfUser, eg.user(userId: message.senderId)]);
```
chrisbobbe, line 46:
```
compose_box: Replace compose box with a banner when cannot post in a channel
```

This seems like a reasonable change to the action-sheet tests, but why is it needed in this commit, which is about the compose box?

…Ah, I think I understand: it's a boring but necessary bit of setup so that the simulated action sheet doesn't crash in the code that decides whether to show the error banner. Is that right?

There are quite a few other changes in this commit that look like they're made for the same reason. Would you move those to a prep commit before this commit? That should make it easier to focus on the interesting changes here. 🙂

chrisbobbe (commented):
Thanks! Comments below, covering everything except the tests added in the main, last commit:

compose_box: Replace compose box with a banner when cannot post in a channel

because I've suggested a cleanup that should help me review those tests more efficiently; see below. 🙂

sm-sayedi:
Thanks @chrisbobbe for the review! Revision pushed. PTAL!

chrisbobbe:
Thanks! Ah, it looks like this has gathered some conflicts—would you mind rebasing and resolving those please?

sm-sayedi:
Conflicts resolved @chrisbobbe! Please have a look!

chrisbobbe (commented):

lib/api/model/model.dart
```diff
@@ -302,6 +314,10 @@ enum UserRole{
   final int? apiValue;
 
   int? toJson() => apiValue;
+
+  bool isAtLeast(UserRole role) {
+    return (apiValue ?? 0) <= (role.apiValue ?? 0);
+  }
```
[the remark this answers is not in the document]
sm-sayedi: > when servers give an API value we don't recognize—say, 350, which would be between 300 "moderator" and 400 "member"—we could still store that value, and use it in the new isAtLeast method.

I think in the current code for `UserRole` we get `UserRole.unknown` for all the unrecognized API values. With that being said, we cannot get that new unrecognized value to compare it with the threshold apiValue. 🙂

sm-sayedi:
Thanks @chrisbobbe for the review. Pushed the new changes. Also added [#886 comment](https://github.com/zulip/zulip-flutter/pull/886#discussion_r1789424071). PTAL.

cc @PIG208

lib/api/model/model.dart
```diff
@@ -315,6 +315,11 @@ enum UserRole{
   final int? apiValue;
 
   int? toJson() => apiValue;
+
+  bool isAtLeast(UserRole threshold) {
+    // Roles with more privilege have lower [apiValue].
+    return (apiValue ?? 0) <= (threshold.apiValue ?? 0);
```
PIG208, line 321: Looking at this, I was thinking about why `0` is chosen as the default value for both operands. Considering the context this helper is used, I suppose that we want to fallback to the behavior that, when the role has an unknown value, we allow the user to post.

This consideration is specific to the posting permission feature we are implementing here. I think the best thing to do is to assert that both api values are non-null; then we handle the case when either of them is null in `hasPostingPermission`.

lib/widgets/compose_box.dart
```diff
@@ -1128,8 +1110,37 @@ class ComposeBox extends StatelessWidget {
...
+    final selfUser = store.users[store.selfUserId]!;
+    switch (narrow) {
+      case ChannelNarrow narrow:
+        final channel = store.streams[narrow.streamId]!;
+        return channel.hasPostingPermission(selfUser, byDate: DateTime.now(), realmWaitingPeriodThreshold: store.realmWaitingPeriodThreshold)
+          ? null : _ErrorBanner(label: ZulipLocalizations.of(context).errorBannerCannotPostInChannelLabel);
+      case TopicNarrow narrow:
+        final channel = store.streams[narrow.streamId]!;
+        return channel.hasPostingPermission(selfUser, byDate: DateTime.now(), realmWaitingPeriodThreshold: store.realmWaitingPeriodThreshold)
+          ? null : _ErrorBanner(label: ZulipLocalizations.of(context).errorBannerCannotPostInChannelLabel);
```
PIG208, line 1124: nit: let's extract `ZulipLocalizations.of(context)` as a variable

PIG208, line 1124: Then we can simplify this to:
```suggestion
      case ChannelNarrow(:final streamId):
      case TopicNarrow(:final streamId):
        final channel = store.streams[streamId]!;
        if (channel.hasPostingPermission(selfUser, byDate: DateTime.now(),
          realmWaitingPeriodThreshold: store.realmWaitingPeriodThreshold)) {
          return _ErrorBanner(
            label: localizations.errorBannerCannotPostInChannelLabel);
        }
```

lib/widgets/compose_box.dart
```diff
@@ -1128,8 +1110,37 @@ class ComposeBox extends StatelessWidget {
...
+      case DmNarrow(:final otherRecipientIds):
+        final hasDeactivatedUser = otherRecipientIds.any((id) =>
+          !(store.users[id]?.isActive ?? true));
+        return hasDeactivatedUser ? _ErrorBanner(label: ZulipLocalizations.of(context)
+          .errorBannerDeactivatedDmLabel) : null;
```
PIG208, line 1129: nit: to make the lines shorter:
```suggestion
        if (hasDeactivatedUser) {
          return _ErrorBanner(
            label: localizations.errorBannerDeactivatedDmLabel);
        }
```

test/widgets/compose_box_test.dart
```diff
@@ -405,97 +407,249 @@ void main() {
...
+          userId: user.userId, isActive: isActive));
+        await tester.pump();
+      }
 
-    DmNarrow groupDmNarrowWith(List<User> otherUsers) => DmNarrow.withOtherUsers(
-      otherUsers.map((u) => u.userId), selfUserId: eg.selfUser.userId);
+      DmNarrow dmNarrowWith(User otherUser) => DmNarrow.withUser(otherUser.userId,
+        selfUserId: eg.selfUser.userId);
 
-    group('1:1 DMs', () {
```
PIG208, line 418: Looks like we are both moving tests to groups and adding the new tests.
The grouping part can be done in a prep commit so the diffs will be easier to read.

test/widgets/compose_box_test.dart
```diff
@@ -405,97 +407,249 @@ void main() {
...
+        (ChannelPostPolicy.administrators, UserRole.unknown,       true),
+        (ChannelPostPolicy.administrators, UserRole.guest,         false),
+        (ChannelPostPolicy.administrators, UserRole.member,        false),
+        (ChannelPostPolicy.administrators, UserRole.moderator,     false),
+        (ChannelPostPolicy.administrators, UserRole.administrator, true),
+        (ChannelPostPolicy.administrators, UserRole.owner,         true),
+      ];
+
+      for (final testCase in testCases) {
+        final (ChannelPostPolicy policy, UserRole role, bool canPost) = testCase;
```
PIG208, line 547: nit:
```suggestion
      for (final (ChannelPostPolicy policy, UserRole role, bool canPost) in testCases) {
```

PIG208, line 547: nit:
```suggestion
      for (final (ChannelPostPolicy policy, UserRole role, bool canPost) in testCases) {
```

test/widgets/compose_box_test.dart
```diff
@@ -405,97 +407,249 @@ void main() {
...
+          selfUser: selfUser,
+          streams: [eg.stream(streamId: 1, channelPostPolicy: ChannelPostPolicy.administrators)]);
         checkComposeBox(isShown: true);
 
-        await changeUserStatus(tester, user: activeUsers[0], isActive: false);
+        await changeUserRole(tester, user: selfUser, role: UserRole.moderator);
         checkComposeBox(isShown: false);
       });
 
-      testWidgets('all deactivated users become active -> '
```
PIG208, line 468: Hmm, the deletion is a bit far apart from where it was re-added.

A goal of restructuring the change would be that we have one commit that moves the old tests, and one that only adds the new ones.

lib/api/model/model.dart
```diff
@@ -382,6 +387,19 @@ class ZulipStream {
     _$ZulipStreamFromJson(json);
 
   Map<String, dynamic> toJson() => _$ZulipStreamToJson(this);
+
+  bool hasPostingPermission(User user, {required DateTime byDate, required int realmWaitingPeriodThreshold}) {
```
PIG208, line 391: nit: let's break this into multiple lines

lib/api/model/model.dart
```diff
@@ -382,6 +387,19 @@ class ZulipStream {
...
+    final role = user.role;
+    return switch (channelPostPolicy) {
+      ChannelPostPolicy.any            => true,
+      ChannelPostPolicy.fullMembers    => role.isAtLeast(UserRole.member) && (role == UserRole.member
```
PIG208, line 395: By turning this into `||` we can replace the `... ? ... : true` clause with `... && ...`.

lib/api/model/model.dart
```diff
@@ -263,6 +263,18 @@ class User {
...
+
+  /// Whether the user has passed the realm's waiting period to be a full member.
+  ///
+  /// See:
+  ///   https://zulip.com/api/roles-and-permissions#determining-if-a-user-is-a-full-member
+  ///
+  /// To determine if a user is a full member, callers must also check that the
+  /// user's role is at least Role.Member.
+  bool hasPassedWaitingPeriod(DateTime byDate, int realmWaitingPeriodThreshold) {
+    final dateJoined = DateTime.parse(this.dateJoined);
```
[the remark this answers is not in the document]
gnprice: We could, but I'd prefer to keep that parsing deferred: converting user objects from JSON is a hot spot performance-wise, because there can be so many of them, and parsing a DateTime is the sort of thing that I feel could easily be slow.

And conversely I think there's only a few places like here where we care about the value of `dateJoined`, so it's not much burden on code complexity for these to have to handle the parsing; and they're all for just one user at a time, so there's no performance concern.

lib/api/model/model.dart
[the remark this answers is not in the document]
chrisbobbe: The proposal is to store the unrecognized API value so that `isAtLeast` can get its hands on it :). Looking closer, I think that would need some preparation, though, such as converting `UserRole` from an `enum` to a plain `class`.

Greg reminds me that we don't actually expect any more API values to be added, though, because the "role" concept is on track to be replaced by "user groups". So the work I've described for handling unrecognized values won't be worth it.

lib/api/model/model.dart
sm-sayedi: Right. If we're going to allow the user to post when `threshold.apiValue == null`, then it should be:
```dart
return (apiValue ?? 0) <= (threshold.apiValue ?? double.maxFinite);
```

Also, as Chris mentioned in [#886 comment](https://github.com/zulip/zulip-flutter/pull/886#discussion_r1796280370), we're not going to expect any new values thus none of these will become null.

lib/model/store.dart
```diff
@@ -235,6 +235,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
       connection: connection,
       realmUrl: realmUrl,
       maxFileUploadSizeMib: initialSnapshot.maxFileUploadSizeMib,
+      realmWaitingPeriodThreshold: initialSnapshot.realmWaitingPeriodThreshold,
```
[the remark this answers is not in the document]
sm-sayedi: Moved it before the line for `maxFileUploadSizeMib`.

Also, looking at these "realm and server" section properties, I'm unsure if they follow a specific order. If they're meant to follow the order of `InitialSnapshot`, then they'll have an order like this:
```dart
customProfileFields
realmWaitingPeriodThreshold
realmUrl
realmDefaultExternalAccounts
maxFileUploadSizeMib
emailAddressVisiblity
```

lib/api/model/model.dart
sm-sayedi: I don't think I have understood it correctly. Do you mean the following?
```dart
ChannelPostPolicy.fullMembers    => role.isAtLeast(UserRole.member) || (role == UserRole.member
                                      && user.hasPassedWaitingPeriod(byDate, realmWaitingPeriodThreshold)),
```
I don't think it results in the same value. If this is not what you meant, could you please write the piece of code you suggest?

sm-sayedi:
Thanks @PIG208 for the review. Changes pushed. PTAL.

lib/api/model/model.dart
```diff
@@ -303,6 +315,11 @@ enum UserRole{
   final int? apiValue;
 
   int? toJson() => apiValue;
+
+  bool isAtLeast(UserRole threshold) {
+    // Roles with more privilege have lower [apiValue].
+    return (apiValue ?? 0) <= (threshold.apiValue ?? double.maxFinite);
```
gnprice, line 321: It doesn't make sense semantically for an "is at least" function to make up an answer like this does when one of the roles is unknown. When one role is unknown, the truth is that we don't know whether `this` is at least `threshold`.

(It could be done given the additional structure on unknown roles that Chris described at https://github.com/zulip/zulip-flutter/pull/886#discussion_r1720487156 and https://github.com/zulip/zulip-flutter/pull/886#discussion_r1731912511 . But as discussed there, it's not worth implementing that structure.)

Because this method can't determine the answer in that case, it should refuse to guess. The simplest way to do that is to have this method throw if either role is unknown, and make it the caller's job to check for that first and decide how to act on the uncertainty.

lib/model/store.dart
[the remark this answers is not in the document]
gnprice: The constructor arguments at this call site should match the order of the parameters in the constructor's declaration, which in turn should match the order the fields are declared in.

For the field declarations, generally we've tried to keep them in a logical order (which is our default approach for ordering things). So `realmUrl` comes before all the miscellaneous settings, because it's the value that identifies the whole realm in the first place. I don't think there's a strong reason the other five you listed are ordered in the way they currently are, though.

That logical ordering may differ from what's in `InitialSnapshot`, because:
```dart
class InitialSnapshot {
  // Keep these fields in the order they appear in the API docs.
  // (For many API types we choose a more logical order than the docs.
  // But this one is so long that that'd make it become impossible to
  // compare the lists by hand.)
```

sm-sayedi:
Thanks @gnprice. Changes pushed.

lib/api/model/model.dart
```diff
@@ -370,6 +388,24 @@ class ZulipStream {
     _$ZulipStreamFromJson(json);
 
   Map<String, dynamic> toJson() => _$ZulipStreamToJson(this);
+
+  bool hasPostingPermission(User user,
+      {required DateTime byDate, required int realmWaitingPeriodThreshold}) {
+    final role = user.role;
+    // We let the users with [unknown] role to send the message, then the server
+    // will decide to accept it or not based on its actual role.
+    if (role == UserRole.unknown) return true;
```
gnprice, line 397: Cool, this solution looks good. Thanks for the revision!

The PR is still in maintainer review, so I'll leave @PIG208 to do the next round of review on the PR as a whole.

test/widgets/compose_box_test.dart
```diff
@@ -505,5 +505,151 @@ void main() {
...
+            narrow: const ChannelNarrow(1),
+            selfUser: selfUser,
+            realmWaitingPeriodThreshold: 30,
+            streams: [eg.stream(streamId: 1, channelPostPolicy: ChannelPostPolicy.fullMembers)]);
+          checkComposeBox(isShown: false);
+        });
+      });
+
+      Future<void> changeUserRole(WidgetTester tester,
+          {required User user, required UserRole role}) async {
```
PIG208, line 591: nit: Formatting
```suggestion
      Future<void> changeUserRole(WidgetTester tester, {
        required User user,
        required UserRole role,
      }) async {
```
We usually do this for named parameters when wrapping them.

test/widgets/compose_box_test.dart
```diff
@@ -505,5 +505,151 @@ void main() {
...
+        await store.handleEvent(RealmUserUpdateEvent(id: 1,
+          userId: user.userId, role: role));
+        await tester.pump();
+      }
+
+      Future<void> changeChannelPolicy(WidgetTester tester,
+          {required ZulipStream channel, required ChannelPostPolicy policy}) async {
```
PIG208, line 598: Similar formatting nit here

lib/api/model/model.dart
PIG208: Hmm I think you are right! I probably mistook the original expression as something else.

test/widgets/compose_box_test.dart
```diff
@@ -405,98 +407,102 @@ void main() {
...
-      check(bannerTextFinder.evaluate().length).equals(isShown ? 1 : 0);
+    void checkBannerWithLabel(String label, {required bool isShown}) {
+      check(find.text(label).evaluate().length).equals(isShown ? 1 : 0);
     }
 
-    void checkComposeBox({required bool isShown}) {
+    void checkComposeBoxIsShown(bool isShown, {required String bannerLabel}) {
       checkComposeBoxParts(areShown: isShown);
-      checkBanner(isShown: !isShown);
+      checkBannerWithLabel(bannerLabel, isShown: !isShown);
```
PIG208, line 423: This refactor (`checkComposeBox` -> `checkComposeBoxIsShown`) should be fine in commit `compose_box: Replace compose box with a banner when cannot post in a channel`, because that makes it obvious what these changes are used for.

test/widgets/compose_box_test.dart
```diff
@@ -405,98 +407,102 @@ void main() {
       matching: find.widgetWithIcon(IconButton, icon));
 
     void checkComposeBoxParts({required bool areShown}) {
-      check(contentFieldFinder().evaluate().length).equals(areShown ? 1 : 0);
+      final inputFieldCount = inputFieldFinder().evaluate().length;
+      areShown ? check(inputFieldCount).isGreaterThan(0) : check(inputFieldCount).equals(0);
```
PIG208, line 411: This looks like an interesting change (`contentFieldFinder` -> `inputFieldFinder`), because this checks if the length is `> 0` now. Looks like this is for handling the presence of the topic input. We may move this to the next commit where it becomes necessary.

PIG208:
Read through the changes and left some comments mostly for style and formatting. Thanks for the revision! The test changes look clean now; I noticed some changes that probably belongs to `compose_box: Replace compose box with a banner when cannot post in a channel`.

sm-sayedi:
Thanks @PIG208 for the review. Changes pushed.

PIG208:
Looks good. Thanks! Moving this to Greg's review.

test/widgets/message_list_test.dart
```diff
@@ -101,15 +101,17 @@ void main() {
     testWidgets('MessageListPageState.narrow', (tester) async {
       final stream = eg.stream();
       await setupMessageListPage(tester, narrow: ChannelNarrow(stream.streamId),
-        messages: [eg.streamMessage(stream: stream, content: "<p>a message</p>")]);
+        messages: [eg.streamMessage(stream: stream, content: "<p>a message</p>")],
+        streams: [stream]);
```
gnprice, line 105: nit:
```suggestion
        streams: [stream],
        messages: [eg.streamMessage(stream: stream, content: "<p>a message</p>")]);
```

The message lives within the stream, so this is the logical order. Also `messages` is the main "payload" parameter, while the others are more like metadata — much like `child` or `children` vs. other parameters in a typical Flutter widget constructor.

test/widgets/compose_box_test.dart
```diff
@@ -31,14 +31,27 @@ void main() {
...
-      {required Narrow narrow, List<User> users = const []}) async {
+  Future<GlobalKey<ComposeBoxController>> prepareComposeBox(WidgetTester tester, {
+    required Narrow narrow,
+    User? selfUser,
+    int? realmWaitingPeriodThreshold,
+    List<User> users = const [],
+    List<ZulipStream> streams = const [],
+  }) async {
+    if (narrow is ChannelNarrow || narrow is TopicNarrow) {
+      final channelId = narrow is ChannelNarrow ? narrow.streamId : (narrow as TopicNarrow).streamId;
```
gnprice, line 42: This can be tightened with a fun use of pattern-matching:
```suggestion
    if (narrow case ChannelNarrow(:var streamId) || TopicNarrow(:var streamId)) {
```

test/widgets/compose_box_test.dart
```diff
@@ -31,14 +31,27 @@ void main() {
...
+      assert(streams.any((stream) => stream.streamId == channelId),
+        'Add a channel with "streamId" the same as of $narrow.streamId to the store.');
+    }
     addTearDown(testBinding.reset);
-    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    final account = eg.account(user: selfUser ?? eg.selfUser);
```
gnprice, line 47: Because the same self-user object is needed below, best to do the `?? eg.selfUser` fallback just once, to avoid accidental mismatch on future changes to this code. So:
```suggestion
    selfUser ??= eg.selfUser;
    final account = eg.account(user: selfUser);
```

test/widgets/compose_box_test.dart
```diff
@@ -31,14 +31,27 @@ void main() {
...
   late FakeApiConnection connection;
 
-  Future<GlobalKey<ComposeBoxController>> prepareComposeBox(WidgetTester tester,
-      {required Narrow narrow, List<User> users = const []}) async {
+  Future<GlobalKey<ComposeBoxController>> prepareComposeBox(WidgetTester tester, {
+    required Narrow narrow,
+    User? selfUser,
+    int? realmWaitingPeriodThreshold,
+    List<User> users = const [],
+    List<ZulipStream> streams = const [],
```
gnprice, line 39: compose_box test [nfc]: Have `prepareComposeBox` accept additional params
    
    The additional params are: selfUser, streams, and realmWaitingPeriodThreshold.

Let's squash the `streams` part into the commit that needs that (which is the next one), and the other two into the commit that needs those (which is the main commit at the end).

That way they're each appearing in the context of the changes that show how and why they're used.

The changes are also basically independent of each other but are happening here in closely neighboring lines of code; so having them in separate commits from each other will I think make both changes easier to read. (The commits they'll be getting squashed into don't have any other changes in this particular spot of the code.)

lib/widgets/compose_box.dart
```diff
@@ -1128,8 +1110,40 @@ class ComposeBox extends StatelessWidget {
...
   }
 
+  Widget? _errorBanner(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final selfUser = store.users[store.selfUserId]!;
+    final localizations = ZulipLocalizations.of(context);
+    switch (narrow) {
+      case ChannelNarrow(:final streamId):
+      case TopicNarrow(:final streamId):
+        final channel = store.streams[streamId]!;
```
gnprice, line 1120: This will crash if we find ourselves at a narrow for a stream/channel we don't know about.

We should avoid crashing in that case. For example it can happen if the user tries to follow a link that goes to a channel they don't have access to.

The principle is the same as the one described for users in #716. If you look at the other references to `store.streams` in this file, you'll see they follow that principle.

test/widgets/compose_box_test.dart
```diff
@@ -385,97 +407,251 @@ void main() {
...
-          users: deactivatedUsers);
-        checkComposeBox(isShown: false);
+    group('in topic/channel narrow according to channel post policy', () {
+      void checkComposeBox({required bool isShown}) => checkComposeBoxIsShown(isShown,
+        bannerLabel: zulipLocalizations.errorBannerCannotPostInChannelLabel);
+
+      final testCases = [
+        (ChannelPostPolicy.unknown,        UserRole.unknown,       true),
+        (ChannelPostPolicy.unknown,        UserRole.guest,         true),
+        (ChannelPostPolicy.unknown,        UserRole.member,        true),
```
gnprice, line 516: These can be adapted to produce those unit tests that `hasPostingPermission` should get.

Then once that has unit tests, the tests for this UI code can be simpler: they only need to exercise the cases that are different as far as the UI code sees things, not cases that differ only within the implementation of the model method. So e.g. cases for a channel narrow and a topic narrow, and needing a banner and not needing one; but not for all the nuances of when one is needed and when it isn't.

gnprice (commented):
Thanks @sm-sayedi for building this, and thanks @chrisbobbe and @PIG208 for the previous reviews!

Generally this looks good; comments below. I've read the first 8 of the 9 commits in full, and all but the tests of the final main commit.

lib/api/model/model.dart
```diff
@@ -263,6 +263,18 @@ class User {
...
+
+  /// Whether the user has passed the realm's waiting period to be a full member.
+  ///
+  /// See:
+  ///   https://zulip.com/api/roles-and-permissions#determining-if-a-user-is-a-full-member
+  ///
+  /// To determine if a user is a full member, callers must also check that the
+  /// user's role is at least `UserRole.member`.
+  bool hasPassedWaitingPeriod(DateTime byDate, int realmWaitingPeriodThreshold) {
+    final dateJoined = DateTime.parse(this.dateJoined);
```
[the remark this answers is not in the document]
sm-sayedi: > What do these `date_joined` strings look like from the server?

It looks like they're in the UTC format. For my account it is `2024-02-24T11:18+00:00`, although my time zone offset is +04:30. Even if it has an offset other than `+00:00`, `DateTime.parse` will still convert it to its equivalent UTC time. This is the relevant part of its documentation:

```
The result is always in either local time or UTC. If a time zone offset other than UTC is specified, the time is converted to the equivalent UTC time.
```

test/widgets/compose_box_test.dart
```diff
@@ -573,97 +600,199 @@ void main() {
...
+      Future<void> changeChannelPolicy(WidgetTester tester, {
+        required ZulipStream channel,
+        required ChannelPostPolicy policy,
+      }) async {
+        await store.handleEvent(eg.channelUpdateEvent(channel,
+          property: ChannelPropertyName.channelPostPolicy, value: policy));
+        await tester.pump();
+      }
+
+      testWidgets('user role decreases -> compose box is replaced with the banner', (tester) async {
```
sm-sayedi, line 749: I am not sure if the verbs "decrease" and "increase" are the proper ones to use here! 😃 I thought a lot about them, but couldn't come up with better ones. Any suggestions would be appreciated!

sm-sayedi:
Thanks @gnprice for the previous review. New changes pushed. PTAL.

Also added a few comments just above and [one](https://github.com/zulip/zulip-flutter/pull/886/#discussion_r1837600528) in the previous review sub-thread.

lib/api/model/model.dart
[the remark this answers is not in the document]
gnprice: Great, that works, then. Would you start a thread in `#api documentation` about making that format explicit in the docs?

> Even if it has an offset other than `+00:00`, `DateTime.parse` will still convert it to its equivalent UTC time.

Yeah, which would be fine. But if it didn't have an explicit offset at all, I believe `DateTime.parse` would interpret it as the client's local time, which would be the wrong result (unless the client's time zone happened to coincide with UTC).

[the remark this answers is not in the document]
sm-sayedi: Yeah sure.

> ...I believe DateTime.parse would interpret it as the client's local time...
Yeah, looking at its implementation and some logging, it is in the client's local time.

test/widgets/compose_box_test.dart
```diff
@@ -573,97 +600,199 @@ void main() {
...
+        final selfUser = eg.user(role: UserRole.guest);
+        final channel = eg.stream(streamId: 1, channelPostPolicy: ChannelPostPolicy.any);
+        await prepareComposeBox(tester,
+          narrow: const ChannelNarrow(1),
+          selfUser: selfUser,
+          streams: [channel]);
         checkComposeBox(isShown: true);
 
-        await changeUserStatus(tester, user: activeUsers[0], isActive: false);
+        await changeChannelPolicy(tester, channel: channel, policy: ChannelPostPolicy.fullMembers);
```
gnprice, line 782: nit: line too long — the key information is past 80 columns

(similarly the initial channelPostPolicy value above)

test/widgets/compose_box_test.dart
```diff
@@ -573,97 +600,199 @@ void main() {
...
+        checkComposeBox(isShown: false);
       });
-    });
 
-    group('group DMs', () {
-      testWidgets('compose box replaced with a banner', (tester) async {
-        final deactivatedUsers = [eg.user(isActive: false), eg.user(isActive: false)];
-        await prepareComposeBox(tester, narrow: groupDmNarrowWith(deactivatedUsers),
-          users: deactivatedUsers);
+      testWidgets('user role increases -> banner is replaced with the compose box', (tester) async {
```
gnprice, line 761: You've already taken the trouble to write these four test cases so we may as well include them here, but for future reference: I typically wouldn't ask for tests to check that updating the data in the store causes the widgets to properly update for the new values. In general we can just rely on the assumption that Flutter's rebuild mechanisms work correctly :slightly_smiling_face: 

The situation where I might want tests for widgets getting updated is when we have a stateful widget with some nontrivial logic of its own for arranging for updates to happen, in which case that logic needs tests. But here it's just `PerAccountStoreWidget.of` that causes rebuilds.

(We also rely on our `handleEvent` implementation calling `notifyListeners` in all the right situations. That isn't currently as thoroughly tested as it ideally would be — instead, for these two event types in particular and some others, we rely on the fact that the implementation straightforwardly says `notifyListeners();` at the end. But if or when at some point we do test that point, the tests in `test/model/` that exercise handling events are the right layer for that.)

test/model/store_test.dart
```diff
@@ -247,6 +247,101 @@ void main() {
...
+    final store = eg.store(initialSnapshot:
+      eg.initialSnapshot(realmWaitingPeriodThreshold: 2));
+
+    final testCases = [
+      ('2024-11-25T10:00+00:00', DateTime.utc(2024, 11, 25 + 0, 10, 00), false),
+      ('2024-11-25T10:00+00:00', DateTime.utc(2024, 11, 25 + 1, 10, 00), false),
+      ('2024-11-25T10:00+00:00', DateTime.utc(2024, 11, 25 + 2, 09, 59), false),
+      ('2024-11-25T10:00+00:00', DateTime.utc(2024, 11, 25 + 2, 10, 00), true),
+      ('2024-11-25T10:00+00:00', DateTime.utc(2024, 11, 25 + 1000, 07, 00), true),
+    ];
```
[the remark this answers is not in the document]
gnprice: Yeah, since we're doing the computations all in terms of UTC (which is good news), that's the right way for the tests to work too.



> I couldn't find a way to force the test environment to have a different time zone other than (my) local or UTC.

Unfortunately there doesn't seem to be such a way. We have a comment on one test where that came up, in `test/widgets/message_list_test.dart` (testing the dates we show in recipient headers):
```dart
      // We show the dates in the user's timezone.  Dart's standard library
      // doesn't give us a way to control which timezone is used — only to
      // choose between UTC and the user's timezone, whatever that may be.
      // So we do the latter, and that means these dates come out differently
      // depending on the timezone in the environment running the tests.
      // Related upstream issues:
      //   https://github.com/dart-lang/sdk/issues/28985 (about DateTime.now, not timezone)
      //   https://github.com/dart-lang/sdk/issues/44928 (about the Dart implementation's own internal tests)
      // For this test, just accept outputs corresponding to any possible timezone.
      tester.widget(find.textContaining(RegExp("Dec 1[89], 2022")));
      tester.widget(find.textContaining(RegExp("Aug 2[23], 2022")));
```

lib/model/store.dart
```diff
@@ -430,6 +434,27 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
...
+  ///
+  /// To determine if a user is a full member, callers must also check that the
+  /// user's role is at least `UserRole.member`.
+  bool hasPassedWaitingPeriod(User user, {required DateTime byDate}) {
+    // [User.dateJoined] is in UTC. For logged-in users, the format is:
+    // YYYY-MM-DDTHH:mm+00:00, which includes the timezone offset for UTC.
+    // For logged-out spectators, the format is: YYYY-MM-DD, which doesn't
+    // include the timezone offset. In the later case, [DateTime.parse] will
+    // interpret it as the client's local timezone, which could lead to
+    // incorrect results.
```
gnprice, line 450:
```suggestion
    // incorrect results; but that's acceptable for now because the app
    // doesn't support viewing as a spectator.
```

That way the reader isn't left hanging wondering what we do about those incorrect results.

gnprice (commented):
Thanks for the revision! Generally this all looks good; various small comments below.

sm-sayedi:
Thanks @gnprice for the review. New changes pushed. Please have a look.

gnprice:
Thanks! Looks good — merging.

One commit-message tweak: I noticed the commits had two variants of what the prefix should be for compose-box changes:
388b83b2d compose test [nfc]: Add a parent test group
f14e32dac compose_box: Replace compose box with a banner when cannot post in a channel

namely "compose" vs "compose_box". So I looked at past examples; the shorter "compose" form is what we used originally, and have still used many more times than "compose_box". So edited to "compose":
157418c26 compose: Replace compose box with a banner when cannot post in a channel

---

FTR here's the command I ended up using to see past usage:
`git log --format='%>(20)%an %cs %h %<(26,trunc)%s' --grep compose origin | grep -P '\bcompose(.box)?|'`

It's a bit more sophisticated than the quick `git log --oneline --grep foo` commands I've used many times.

Issue #879, opened 2024-08-11
Crash on Android when tapping on camera icon in compose box, in release mode


rajveermalviya:
Logs: 
```shell-session
$ flutter run --release
...
...
...
I/flutter (14099): [IMPORTANT:flutter/shell/platform/android/android_context_vk_impeller.cc(59)] Using the Impeller rendering backend (Vulkan).
I/flutter (14099): [IMPORTANT:flutter/shell/platform/android/android_context_vk_impeller.cc(59)] Using the Impeller rendering backend (Vulkan).
E/AndroidRuntime(14099): FATAL EXCEPTION: flutter-worker-1
E/AndroidRuntime(14099): Process: com.zulip.flutter, PID: 14099
E/AndroidRuntime(14099): java.lang.IncompatibleClassChangeError: Class 'android.content.res.XmlBlock$Parser' does not implement interface 'w7.a' in call to 'int w7.a.next()' (declaration of 'androidx.core.content.b' appears in /data/app/~~SJEWO15N5DtF7C3Opydr8A==/com.zulip.flutter-7Osxw0qoFXmxClFDo-wbPg==/base.apk)
E/AndroidRuntime(14099):        at androidx.core.content.b.j(Unknown Source:19)
E/AndroidRuntime(14099):        at androidx.core.content.b.g(Unknown Source:11)
E/AndroidRuntime(14099):        at androidx.core.content.b.h(Unknown Source:1)
E/AndroidRuntime(14099):        at p6.l$b.b(Unknown Source:2)
E/AndroidRuntime(14099):        at p6.l.P(Unknown Source:51)
E/AndroidRuntime(14099):        at p6.l.X(Unknown Source:35)
E/AndroidRuntime(14099):        at p6.n.j(Unknown Source:68)
E/AndroidRuntime(14099):        at p6.s$f.n(Unknown Source:33)
E/AndroidRuntime(14099):        at p6.s$f.f(Unknown Source:0)
E/AndroidRuntime(14099):        at p6.t.a(Unknown Source:2)
E/AndroidRuntime(14099):        at j6.a$b.a(Unknown Source:17)
E/AndroidRuntime(14099):        at a6.c.l(Unknown Source:18)
E/AndroidRuntime(14099):        at a6.c.m(Unknown Source:41)
E/AndroidRuntime(14099):        at a6.c.i(Unknown Source:0)
E/AndroidRuntime(14099):        at a6.b.run(Unknown Source:12)
[...]

PR #887, opened 2024-08-12
deps android: Upgrade Android Gradle Plugin

rajveermalviya:
Fixes: #879

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -28,6 +28,13 @@ android {
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
+
+        // The use of desugar_jdk_libs causes warning noise at build time:
+        //   https://github.com/zulip/zulip-flutter/pull/887#issuecomment-2287653388
+        //   https://issuetracker.google.com/issues/294273986
+        // TODO(#351): Try removing core-library desugaring once we've
+        //   removed flutter_local_notifications.
+        coreLibraryDesugaringEnabled true
     }
 
     kotlinOptions {
@@ -97,4 +104,5 @@ flutter {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:2.0.4"
 }
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -23,7 +23,6 @@ android {
     namespace "com.zulip.flutter"
 
     compileSdkVersion flutter.compileSdkVersion
-    ndkVersion flutter.ndkVersion
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -13,4 +13,4 @@ agpVersion=8.1.4
 #   https://kotlinlang.org/docs/releases.html#release-details
 # A helpful discussion is at:
 #   https://stackoverflow.com/a/74425347
-kotlinVersion=1.9.10
+kotlinVersion=2.0.10
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -6,7 +6,7 @@ android.enableJetifier=true
 # Defining them here makes them available both in
 # settings.gradle and in the build.gradle files.
 
-agpVersion=8.1.4
+agpVersion=8.5.2
 
 # Generally update this to the version found in recent releases
 # of Android Studio, as listed in this table:
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -6,7 +6,7 @@
 #  the wrapper is the one from the new Gradle too.)
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-all.zip
 networkTimeout=10000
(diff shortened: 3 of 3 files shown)


android/app/build.gradle
```diff
@@ -23,9 +23,9 @@ android {
     namespace "com.zulip.flutter"
 
     compileSdkVersion flutter.compileSdkVersion
-    ndkVersion flutter.ndkVersion
```
gnprice, line 26: android build: Avoid specifying ndkVersion explicitly

This commit message needs a bit more explanation, so that our future selves and/or successors reading this a couple of years from now when the question of the NDK version comes up again don't have to wonder what we were thinking or sleuth it out.

I'll write a version.

gnprice (commented):

I read through all the linked release notes (well, skimming over the parts that seem irrelevant to us like Kotlin/Native), and these all make sense to me. I took some additional notes for a couple of the commits, which I'll push as a revision.

It looks like I'll have to upgrade my Android Studio before I can test this locally (I'm currently on Hedgehog). Is the last commit, the AGP upgrade, the one that fixes #879? Let's mention that in the commit message.

rajveermalviya:

android/gradle.properties
```diff
@@ -6,11 +6,11 @@ android.enableJetifier=true
 # Defining them here makes them available both in
 # settings.gradle and in the build.gradle files.
 
-agpVersion=8.1.4
+agpVersion=8.5.0
```
gnprice, line 9: Is there a reason to go for 8.5.0 instead of 8.5.2? It looks like that's the latest.

(My usual approach would be to go for the latest patch release within a given minor release, even when holding back at an older minor or major release for whatever reason.)

gnprice:
OK, I upgraded my Android Studio so I could build this, and confirmed that it fixes #879 (and that I can reproduce that before this PR).

I'm seeing also this warning on `flutter run`:
```
$ flutter run --release -d p
Launching lib/main.dart on Pixel 8 in release mode...
Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.concurrent.ConcurrentHashMap$TreeBin {
  int lockState;
}`
Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.concurrent.ConcurrentHashMap {
  int sizeCtl;
  int transferIndex;
  long baseCount;
  int cellsBusy;
}`
Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.concurrent.ConcurrentHashMap$CounterCell {
  long value;
}`
Info: Proguard configuration rule does not match anything: `-keepclassmembers enum * {
  public static **[] values();
  public static ** valueOf(java.lang.String);
  public static final synthetic <fields>;
}`
Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.IntSummaryStatistics {
  long count;
  long sum;
  int min;
  int max;
}`
Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.LongSummaryStatistics {
  long count;
  long sum;
  long min;
  long max;
}`
Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.DoubleSummaryStatistics {
  long count;
  double sum;
  double min;
  double max;
}`
Running Gradle task 'assembleRelease'...                           58.0s
```

Not catastrophic if we have to live with those warnings for a bit, but it is somewhat annoying. Would you spend a bit of time investigating that and trying to find a fix?

gnprice:
Also confirmed that #879 reproduces just before the last commit, so it is that commit (the AGP upgrade) that fixes it.

(As a side effect I learned that that next-to-last commit already has the warning above, so whatever introduces that comes before the last commit.)

rajveermalviya:
> seeing warnings on flutter run
> ```
> ...
> Info: Proguard configuration rule does not match anything: `-keepclassmembers class j$.util.concurrent.ConcurrentHashMap$TreeBin {
> ...
>```

I only got that after a `flutter clean`, doesn't get be printed in subsequent runs of `flutter run`, guess it's because of caching. Also it seems it has been reported upstream to be a `com.android.tools:desugar_jdk_libs` issue — https://issuetracker.google.com/issues/294273986

android/gradle.properties
rajveermalviya: > Is there a reason to go for 8.5.0 instead of 8.5.2?

Not really, hadn't realized there was a newer one, I just took that version from release notes.
Updated now to 8.5.2.

rajveermalviya:

gnprice:
> I only got that after a `flutter clean`, doesn't seem to printed in subsequent runs of `flutter run`, guess it's because of caching.

Yeah, that seems to be how warnings in Android builds often behave. It can be annoying when trying to debug and eliminate the warnings, because it makes it tricky to confidently reproduce them… but it does also mitigate the impact of the warning noise somewhat.


> Also it seems it has been reported upstream to be a `com.android.tools:desugar_jdk_libs` issue — https://issuetracker.google.com/issues/294273986

Cool, thanks for tracking that down. I upvoted the issue and subscribed. Looks like it's been there a year, which is less promising than if it were just recently filed; but it also lacks a repro, so after merging this I'll comment there offering zulip-flutter as a repro, and potentially that'll help get it fixed.

That also means the `coreLibraryDesugaring` commit must be the one that introduces the warning. I'll add a comment at that line, and a note in #351, that once we remove flutter_local_notifications we can try removing that step.

gnprice:
> a `com.android.tools:desugar_jdk_libs` issue — https://issuetracker.google.com/issues/294273986

Following up on this: my comment prompted that thread to get closed as a duplicate, and it turns out that the bug had been fixed (in that library's main branch) just a couple of days earlier:
https://issuetracker.google.com/issues/358166486#comment8

So eventually that'll filter into AGP, and then into an AGP release that we upgrade to.

Meanwhile we're about to complete #351 removing `package:flutter_local_notifications`, so that PR #856 is removing this desugaring step anyway as hoped.

PR #888, opened 2024-08-14
Prep for supporting set typing status

PIG208:


lib/model/typing_status.dart
```diff
@@ -86,3 +89,73 @@ class TypingStatus extends ChangeNotifier {
...
+  }
+
+  Future<void> _maybeNotifyStart(PerAccountStore store) async {
+    if (_nextStartTime == null || DateTime.now().isAfter(_nextStartTime!)) {
+      await _notifyStart(store);
+    }
+  }
+
+  Future<void> _notifyStart(PerAccountStore store) async {
+    _nextStartTime = DateTime.now().add(typingStartedWaitPeriod);
```
gnprice, line 138: Quick high-level comment: let's write this so that we definitely don't have to think about time zones to reason about it. :slightly_smiling_face: 

I think an OK version could be made with `DateTime.timestamp()`, which uses UTC. Probably a cleaner approach is to use `Stopwatch` — then we're really not thinking about calendar dates or anything like that, just elapsed durations, which is exactly right for this subsystem.

PIG208:
Moved the main commits to #897, so that we can get some earlier independent changes reviewed separately.

test/api/fake_api.dart
```diff
@@ -29,11 +29,11 @@ class _PreparedSuccess extends _PreparedResponse {
 /// An [http.Client] that accepts and replays canned responses, for testing.
 class FakeHttpClient extends http.BaseClient {
 
-  http.BaseRequest? lastRequest;
+  List<http.BaseRequest> requestHistory = [];
```
chrisbobbe, line 33: It might be helpful to make this a getter that returns an `Iterable`, which is backed by a private field that's a List? Then consumers of the request history can't mess with things by modifying the list.

test/api/fake_api.dart
```diff
@@ -49,40 +49,40 @@ class FakeHttpClient extends http.BaseClient {
...
     Object? exception,
     int? httpStatus,
     Map<String, dynamic>? json,
     String? body,
     Duration delay = Duration.zero,
   }) {
-    assert(_nextResponse == null,
-      'FakeApiConnection.prepare was called while already expecting a request');
+    // TODO: Prevent a source of bugs by ensuring that there is no outstanding
+    // prepared responses when the test ends.
```
chrisbobbe, line 62: nit: "that there _are_ no outstanding prepared responses"

lib/api/route/typing.dart
```diff
@@ -0,0 +1,26 @@
...
+        'type': RawParameter((supportsTypeChannel) ? 'channel' : 'stream'),
+        'op': RawParameter(op.toJson()),
+        'stream_id': destination.streamId,
+        'topic': RawParameter(destination.topic),
+      });
+    case DmDestination():
+      return connection.post('setTypingStatus', (_) {}, 'typing', {
+        'op': RawParameter(op.toJson()),
+        'to': destination.userIds,
+      });
```
chrisbobbe, line 24: What's the effect of not passing 'type' in the DM case? From the doc, it looks like the preferred client behavior (assuming Zulip Server 4+) is to pass 'private' when FL is <174, else 'direct'. I'm not sure I've read closely enough to know what happens across the range of supported servers if we just don't pass 'type' at all.

lib/api/route/typing.dart
```diff
@@ -0,0 +1,26 @@
...
+  required TypingOp op,
+  required MessageDestination destination,
+}) {
+  final supportsTypeChannel = connection.zulipFeatureLevel! >= 248; // TODO(server-9)
+  switch (destination) {
+    case StreamDestination():
+      return connection.post('setTypingStatus', (_) {}, 'typing', {
+        'type': RawParameter((supportsTypeChannel) ? 'channel' : 'stream'),
+        'op': RawParameter(op.toJson()),
+        'stream_id': destination.streamId,
```
chrisbobbe, line 17: The API doc says 'stream_id' is new in FL 215, and there's a different way we need to express the destination for <215.

test/api/route/typing_test.dart
```diff
@@ -0,0 +1,81 @@
...
+  Future<void> checkSetTypingStatus(FakeApiConnection connection, {
+    required TypingOp op,
+    required MessageDestination destination,
+    required Map<String, String> expectedBodyFields,
+  }) async {
+    connection.prepare(json: {});
+    await setTypingStatus(connection,
+      op: op,
+      destination: destination,
+    );
```
chrisbobbe, line 27:
```suggestion
      destination: destination);
```

test/api/route/typing_test.dart
```diff
@@ -0,0 +1,81 @@
...
+      checkSetTypingStatus(connection,
+        op: TypingOp.start,
+        destination: const StreamDestination(streamId, topic),
+        expectedBodyFields: {
+          'type': 'stream',
+          'op': 'start',
+          'stream_id': streamId.toString(),
+          'topic': topic,
+        },
+      );
```
chrisbobbe, line 78:
```suggestion
        });
```

chrisbobbe (commented):
Thanks! Comments below.

lib/api/route/typing.dart
PIG208: Interesting, thanks for catching that!

lib/api/route/typing.dart
PIG208: This was intended to match the mobile apps' implementation, where we omit the type field for dm typing notfications I think it is not clear that the default was "private" back then. Skipping `type` this way allows us to not worry about the "private" -> "direct" transition of this API.

See: https://chat.zulip.org/#narrow/stream/378-api-design/topic/stream.20typing.20indicator

chrisbobbe: I see. I guess since zulip-mobile does that not-preferred thing and presumably still works, probably makes it harmless if we also do the not-preferred thing. 🙂 But also, the burden of handling the "private" -> "direct" transition isn't heavy—

> pass 'private' when FL is <174, else 'direct'

—so I'd be in favor of just doing that. If not, though, let's include a comment to make it clear why we're doing something that's different from the thing that seems natural/expected from a reading of the API doc.

PIG208:
Thanks for the review @chrisbobbe! I have updated the PR.

- Dropped `fromSendableNarrow` to keep the api code from depending on lib/model.
- Reorganized the tests and made them shorter.
- Handle the legacy case for FL 215.

PIG208:
Pushed to rebase.

test/api/fake_api.dart
```diff
@@ -29,18 +30,18 @@ class _PreparedSuccess extends _PreparedResponse {
...
 class FakeHttpClient extends http.BaseClient {
 
-  http.BaseRequest? lastRequest;
+  Iterable<http.BaseRequest> get requestHistory => _requestHistory;
+  List<http.BaseRequest> _requestHistory = [];
 
-  http.BaseRequest? takeLastRequest() {
-    final result = lastRequest;
-    lastRequest = null;
+  List<http.BaseRequest> takeRequests() {
```
chrisbobbe, line 36: How about also making the return type `Iterable` here?

chrisbobbe (commented):
Thanks for the revision! This looks good except one small comment below, and there's still one open comment from last time ( https://github.com/zulip/zulip-flutter/pull/888#discussion_r1727988511 ) but I'll go ahead and mark this for Greg's review, and perhaps he'll add his thoughts on that too.

PIG208:
Thanks! I should push an update later addressing these comments.

test/api/fake_api.dart
PIG208: I didn't apply the same refactor there because `takeRequests` do not need to protect the list from modification. So `List` would be an accurate return type.

PIG208:
OK, this has been updated :)

PIG208:
Fixed indentation in a test.

PIG208:
Pushed to rebase.

PIG208:
Dropped the `checkDmExpectedOp` helper because the `TypingOp.toJson` is already tested with the typing status for topic tests. Also revised the wording of some tests.

lib/api/route/typing.dart
```diff
@@ -0,0 +1,30 @@
...
+Future<void> setTypingStatus(ApiConnection connection, {
+  required TypingOp op,
+  required MessageDestination destination,
+}) {
+  switch (destination) {
+    case StreamDestination():
+      final supportsTypeChannel = connection.zulipFeatureLevel! >= 248; // TODO(server-9)
+      final supportsStreamId = connection.zulipFeatureLevel! >= 215; // TODO(server-8)
+      return connection.post('setTypingStatus', (_) {}, 'typing', {
+        'type':  RawParameter((supportsTypeChannel) ? 'channel' : 'stream'),
```
gnprice, line 16: nit: extra parens

lib/api/route/typing.dart
```diff
@@ -0,0 +1,30 @@
...
+        'op':    RawParameter(op.toJson()),
+        'topic': RawParameter(destination.topic),
+        ...(supportsStreamId) ? {'stream_id': destination.streamId}
+                              : {'to': [destination.streamId]}
```
gnprice, line 20: nit: put in logical order: channel/stream ID before topic, because the topic lives within the channel

Similarly, put `type` just above the channel ID (or list of DM recipients), because it's the next layer of namespacing above those.

gnprice, line 20: A fun Dart 3 way to write this:
```suggestion
        if (supportsStreamId) 'stream_id': destination.streamId
        else                  'to': [destination.streamId],
```

test/api/route/typing_test.dart
```diff
@@ -0,0 +1,88 @@
...
+import '../../stdlib_checks.dart';
+import '../fake_api.dart';
+
+void main() {
+  const streamId = 123;
+  const topic = 'topic';
+  const userIds = [101, 102, 103];
+
+  Future<void> checkSetTypingStatus(FakeApiConnection connection, TypingOp op, {
+    required MessageDestination destination,
```
gnprice, line 19: nit:
```suggestion
  Future<void> checkSetTypingStatus(FakeApiConnection connection,
    TypingOp op, {
    required MessageDestination destination,
```

else "op" gets a bit lost visually

test/api/route/typing_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    await setTypingStatus(connection, op: op, destination: destination);
+    check(connection.lastRequest).isA<http.Request>()
+      ..method.equals('POST')
+      ..url.path.equals('/api/v1/typing')
+      ..bodyFields.deepEquals(expectedBodyFields);
+  }
+
+  Future<void> checkSetTypingStatusForTopic(TypingOp op, String expectedOp) =>
+    FakeApiConnection.with_((connection) =>
+      checkSetTypingStatus(connection, op,
```
gnprice, line 32: nit: block bodies

see that Flutter style guide section I've linked a few times before :slightly_smiling_face:

test/api/route/typing_test.dart
```diff
@@ -0,0 +1,88 @@
...
+        destination: const StreamDestination(streamId, topic),
+        expectedBodyFields: {
+          'type': 'channel',
+          'op': expectedOp,
+          'stream_id': streamId.toString(),
+          'topic': topic,
+        }));
+
+  test('send typing status start for topic', () =>
+    checkSetTypingStatusForTopic(TypingOp.start, 'start'));
```
gnprice, line 42: And definitely block bodies for test cases.

lib/api/route/typing.dart
gnprice: Yeah, cleanest to just do the thing that's preferred in the current API.

(And then adapt for old servers as necessary — but "type" has been accepted since server-4, so effectively forever now.)

gnprice:

Generally this all looks good. Several comments above, all small.

PIG208:
Thanks! Pushed an update to the last commit.

gnprice:
Thanks! Looks good; merging.

Issue #234, opened 2023-07-21
autocomplete: Support wildcard @-mentions


chrisbobbe:
With #207 being merged, we now have an @-mention autocomplete UI for user mentions! 🎉

We should extend it to support autocompleting wildcard mentions, like `@**all**` and `@**channel**`.

The exact set of wildcard-mention options will depend on the server version:
- "channel" being an option for new servers, and "stream" for old servers — due to https://github.com/zulip/zulip/issues/28468. See https://github.com/zulip/zulip-mobile/pull/5830 for how we handle that in zulip-mobile.
- "topic" being an option for servers 8+ but not older servers; see https://github.com/zulip/zulip-mobile/pull/5792

Issue #222, opened 2023-07-18
Add Font Awesome icons


gnprice:
In Zulip web we use Font Awesome for many of the icons in the UI. We should pull in an appropriate dependency so we can start using the same icons in zulip-flutter. (This is a sort of followup to #200.)

This package seems appropriate and popular:
https://pub.dev/packages/font_awesome_flutter

One concrete place we could use such an icon is in the title of the "All messages" narrow:
https://github.com/zulip/zulip-flutter/pull/219#discussion_r1267326238

terpimost:
I don't suggest to adopt any font awesome icons at all

gnprice:
Cool, thanks. In that case I guess we can close this issue. It'll always be here as a reference in case we later decide for some reason we do want some such icons after all.

PR #889, opened 2024-08-16
autocomplete: Support @-wildcard in user-mention autocomplete

sm-sayedi:
### Screen recording

https://github.com/user-attachments/assets/f5a17bb3-2862-46c9-a83e-33b939e0538c

Fixes: #234


assets/icons/bullhorn.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M12,8H4A2,2 0 0,0 2,10V14A2,2 0 0,0 4,16H5V20A1,1 0 0,0 6,21H8A1,1 0 0,0 9,20V16H12L17,20V4L12,8M15,15.6L13,14H4V10H13L15,8.4V15.6M21.5,12C21.5,13.71 20.54,15.26 19,16V8C20.53,8.75 21.5,10.3 21.5,12Z" /></svg>
```
gnprice, line 1: What was the origin of this icon — is it from a particular place in Figma?

(I don't see a `bullhorn.svg` in Zulip web, which would be the other usual source.)

See `git log --stat assets/icons/` for example previous commit messages.

lib/api/model/model.dart
```diff
@@ -178,6 +178,43 @@ enum Emojiset {
...
+
+  /// The value to be put at the compose box after choosing an option from autocomplete.
+  ///
+  /// The same as [name], except for "stream" it is "channel" in FL >= 247 (server-9).
+  final String value; // TODO(sever-9): remove, instead use [name]
+
+  /// The full name of the wildcard to be shown in autocomplete suggestions.
+  ///
+  /// Ex: "all (Notify channel)" or "everyone (Notify recipients)".
+  final String fullDisplayName;
```
gnprice, line 205: This seems more about our UI than about the Zulip server API. So let's find a home for it outside of lib/api/.

I think a good solution would look similar to `MentionAutocompleteResult` and its subclasses. Instead of a `User` being an object that knows how to behave as an @-mention autocomplete candidate, we'd have an object that describes an @-mention autocomplete candidate and refers to a `User` to do so.

gnprice: In fact probably a good solution would be for these "autocomplete candidate" objects to be the same objects as the result objects.

I'll try doing a quick refactor in that direction.

lib/model/store.dart
```diff
@@ -341,6 +342,54 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
 
   final Map<int, User> users;
 
+  Map<String, Wildcard> wildcardsForNarrow(Narrow narrow) => Map.fromEntries(
+    _wildcardsForNarrow(narrow).map((w) => MapEntry(w.name, w)));
+
+  List<Wildcard> _wildcardsForNarrow(Narrow narrow) {
```
gnprice, line 348: This should live in an autocomplete-specific file — `store.dart`, and the `PerAccountStore` class definition, are fairly crowded places, so anything that can naturally go somewhere more specific should do so.

lib/model/store.dart
```diff
@@ -341,6 +342,54 @@ class PerAccountStore extends ChangeNotifier with ChannelStore, MessageStore {
...
+    final isDmNarrow = narrow is DmNarrow;
+    final isChannelWildcardAvailable = account.zulipFeatureLevel >= 247; // TODO(server-9)
+    final isTopicWildcardAvailable = account.zulipFeatureLevel >= 188; // TODO(sever-8)
+    return [
+      Wildcard(
```
gnprice, line 353: Does this mirror the corresponding logic in zulip-mobile? (I haven't read this closely.) A link for comparison would be handy.

gnprice (commented):
Thanks @sm-sayedi! A few high-level comments below.

gnprice:
> In fact probably a good solution would be for these "autocomplete candidate" objects to be the same objects as the result objects.
> 
> I'll try doing a quick refactor in that direction.

OK, I started down that road (generating #895 along the way), and then realized it would mean we'd allocate one of those objects for every user, in `MentionAutocompleteView._usersByRelevance` and so `MentionAutocompleteView.init`. That doesn't sound great performance-wise when there are like 30k users in a realm.

Instead, let's handle the wildcard-vs-user distinction in the control flow. Take a look at #896, and build your changes atop that. Then the existing results for an @-mention are computed here:
```dart
  Future<List<MentionAutocompleteResult>?> computeResults() async {
    final results = <MentionAutocompleteResult>[];
    if (await filterCandidates(filter: _testUser,
          candidates: sortedUsers, results: results)) {
      return null;
    }
    return results;
  }
```

where that `filterCandidates` call does the main loop and accumulates results onto that `results` local. To add wildcard candidates, you can just add more code next to that, calling `List.add` on the same `results` local.

lib/api/model/model.dart
gnprice: (→ https://github.com/zulip/zulip-flutter/pull/889#issuecomment-2297696566 below)

assets/icons/bullhorn.svg
sm-sayedi: That icon was from materialdesignicons.com. Updated the icon in this revision to match it with the web. The web uses a Font Awesome icon.

lib/api/model/model.dart
sm-sayedi: > This seems more about our UI than about the Zulip server API. So let's find a home for it outside of lib/api/.

Now that the model class is moved outside of lib/api, is it okay for `fullDisplayName` to still live in the class?

lib/model/store.dart
sm-sayedi: It doesn't follow the zulip-mobile in the exact same way. Here's the web implementation: https://github.com/zulip/zulip-mobile/blob/main/src/autocomplete/WildcardMentionItem.js

sm-sayedi:
Thanks @gnprice for the previous review. It was really helpful. Pushed a new revision (still draft). PTAL.

assets/icons/bullhorn.svg
gnprice: Cool. That sounds like a case of this issue, then:
* #222 

Given Vlad's reply there, we'll probably want some other icon for this. I think the most efficient way to resolve that is to start a thread in `#mobile` and ask Vlad there what icon he'd like us to use.

He did recently finish a design that's related (I still need to update the relevant issue to reflect that):
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3859-2936&t=z2va3CAInaEg1eWy-0
but looking there, it doesn't have an example for a wildcard, only for a group or an individual user. So go ahead and ask in chat, and include a link to that Figma node for context.

lib/api/model/model.dart
gnprice: Yeah. Try to find the arrangement that seems cleanest (and once the PR is ready, reviewers may have suggestions), but between lib/model/ and lib/widgets/ either way is basically fine for where that lives.

lib/model/store.dart
gnprice: OK. Let's try to make sure it gets the same results, even if the code ends up organized differently. (There should be test cases in zulip-mobile that will be helpful to translate over, to verify that.)

Then also include a link to that zulip-mobile implementation in either a code comment or commit message. For the link, make it a GitHub permalink:
https://github.com/zulip/zulip-flutter/pull/692#discussion_r1674802255
https://github.com/zulip/zulip-flutter/pull/884#discussion_r1714682330
That'll be helpful for reviewers, as a point of comparison.

alya:
We're currently [thinking](https://chat.zulip.org/#narrow/stream/101-design/topic/wildcard.20mentions.20icon/near/1911443) we should use the three-people icon for wildcard mentions, rather than a loudspeaker. Assuming we move forward with it for the web app, we should make the same change here.

lib/model/autocomplete.dart
```diff
@@ -483,6 +468,42 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
...
   }
 
+  bool _isChannelWildcardIncluded = false;
+
+  @override
+  Future<List<MentionAutocompleteResult>?> computeResults() async {
+    _isChannelWildcardIncluded = false;
+    final results = <MentionAutocompleteResult>[];
+    // give priority to wildcard mentions
+    if (await filterCandidates(filter: _testWildcard,
```
gnprice, line 478: The fancy `filterCandidates` machinery isn't needed for the wildcards, because there's only a handful of them. (It's useful when there are potentially thousands of candidates.) So this can be simplified to a direct loop through the wildcard candidates.

That also lets `_inChannelWildcardIncluded` be a local variable, which simplifies reasoning about it.

… Then I started writing that the logic might be further clarified by unrolling the loop, akin to the logic in zulip-mobile:
https://github.com/zulip/zulip-mobile/blob/715d60a5e87fe37032bce58bd72edb99208e15be/src/autocomplete/WildcardMentionItem.js#L113-L138

But I see we discussed this previously above:
https://github.com/zulip/zulip-flutter/pull/889#discussion_r1727788259
and that you have the corresponding logic living instead in the widgets code, in `ComposeAutocomplete._wildcards`. That structure may be fine too. Should have a link to the zulip-mobile version, though, as discussed there.

lib/model/autocomplete.dart
```diff
@@ -284,46 +284,33 @@ abstract class AutocompleteView<QueryT extends AutocompleteQuery, ResultT extend
...
     store.autocompleteViewManager.registerMentionAutocomplete(view);
     return view;
   }
 
   final Narrow narrow;
+  final List<Wildcard> wildcards;
   final List<User> sortedUsers;
 
-  @override
-  Future<List<MentionAutocompleteResult>?> computeResults() async {
```
gnprice, line 311: Moving this method to lower in the file is perfectly reasonable, but should happen as its own separate NFC prep commit. That way the substantive change stands out better.

lib/widgets/autocomplete.dart
```diff
@@ -165,7 +167,54 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
+    final isTopicWildcardAvailable = store.account.zulipFeatureLevel >= 188; // TODO(sever-8)
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    return [
+      Wildcard(
+        name: zulipLocalizations.all,
+        value: 'all',
+        fullDisplayName: 'all (${isDmNarrow
+          ? zulipLocalizations.notifyRecipients
+          : zulipLocalizations.notifyChannel(isChannelWildcardAvailable
+              ? "channel" : "stream")})',
```
gnprice, line 185: I think it will be cleaner to move this `fullDisplayName` logic to where the result is actually being displayed, at build time.

lib/widgets/autocomplete.dart
```diff
@@ -196,12 +247,17 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
     switch (option) {
       case UserMentionAutocompleteResult(:var userId):
-        avatar = Avatar(userId: userId, size: 32, borderRadius: 3);
-        label = PerAccountStoreWidget.of(context).users[userId]!.fullName;
+        avatar = Avatar(userId: userId, size: 32, borderRadius: 3); // web uses 21px
+        label = store.users[userId]!.fullName;
+      case WildcardMentionAutocompleteResult(:var wildcardName):
+        avatar = const Icon(ZulipIcons.bullhorn, size: 29); // web uses 19px
+        print('wildcard name: $wildcardName');
+        label = _wildcards(context, store).singleWhere((w) => w.name == wildcardName).fullDisplayName;
```
gnprice, line 260: Let's avoid doing a search like this, and instead have the MentionAutocompleteResult value include whatever information this needs.

(If we did do a search, then like with `store.users[userId]` above, we should make it so that we're just looking it up in an existing data structure rather than recomputing the whole list again.)

In fact I think I'd like to refactor UserMentionAutocompleteResult so that it carries the actual `User` object, rather than just a user ID. That change is orthogonal to this PR, though.

lib/widgets/autocomplete.dart
gnprice: Also avoid string manipulation on translated strings. See the "general background" at the end of this section:
https://github.com/zulip/zulip-mobile/blob/main/docs/howto/translations.md#all-code-contributors-wiring-our-code-for-translations

Instead, put the two parts of this into separate widgets or `TextSpan`s. That way they can also get different text styles.

gnprice (commented):
Hmm, I missed that this had a request for my review, sorry ­@sm-sayedi — being a draft and with neither of the review labels, it looked like it was awaiting another revision first.

In general any time there's a PR where you're awaiting review for more than a week or two, please feel free to ping on the PR thread or in chat. We generally try to turn around reviews within a day or two, so a wait of over a week is likely a sign that something fell through the cracks.

Comments below. Since this PR is still a draft, I looked only at high-level aspects of the structure.

gnprice:
Also just pushed a rebase of the branch. (I did the rebase in order to better review it, since some things had changed in main; having done that, I figure I might as well share the result.)

The main change is the rebase across e91694f509dd22356bf82c6c50aa1ae331e2f3f4 (#908), which causes some of the changes here to disappear as having already happened.

lib/widgets/autocomplete.dart
```diff
@@ -223,28 +229,62 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
...
+            : localizations.notifyStream,
+      Wildcard.channel => localizations.notifyChannel,
+      Wildcard.stream => isChannelWildcardAvailable
+        ? localizations.notifyChannel
+        : localizations.notifyStream,
+      Wildcard.topic => localizations.notifyTopic,
+    };
+    return Text.rich(TextSpan(text: '${wildcard.name} ', children: [
+      TextSpan(text: description, style: TextStyle(fontSize: 12,
+        color: Colors.black.withValues(alpha: 0.8)))]));
```
sm-sayedi, line 286: Web uses 85% of the name (i.e., **all**) font size for the description (i.e., **Notify channel**). For us, 85% or 14 is almost 12.

sm-sayedi:
The CI is failing and it's related to icons. I am not sure why it is happening!

gnprice:
The error output is:
```
Running icons...

added 162 packages in 29s
Error: there were icon updates:
 M assets/icons/ZulipIcons.ttf
```

Seems like that font file isn't updated to match your other changes. The icons.dart file has instructions for updating for icons.

sm-sayedi:
Just transferred the discussion to CZO: https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/CI.20failure.20related.20to.20Zulip.20icons.20-.20.23F889/near/1994794.

test/model/autocomplete_test.dart
```diff
@@ -438,7 +438,7 @@ void main() {
     });
   });
 
-  group('MentionAutocompleteView sorting users results', () {
+  group('MentionAutocompleteView sorting user mention results', () {
```
chrisbobbe, line 441:
```suggestion
  group('MentionAutocompleteView sorting mention results', () {
```

(since wildcard mentions aren't user mentions)

chrisbobbe (commented):
Thanks, @sm-sayedi! Comments below.

> Note: The first two commits are actually necessary for displaying wildcard mentions based on a [permission](https://sm-sayedi.zulipchat.com/help/restrict-wildcard-mentions). I first created those commits and then started the [CZO discussion ](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/.40-wildcard.20in.20user-mention.20autocomplete.20-.20.23F889), in which it was decided to leave the permission part for a future issue. I am not sure if we want to keep these two commits for now!

I think it's fine to leave those commits in. Checking that the new fields are added in the proper order isn't super trivial; we might as well keep the work we did for that. 🙂

lib/model/autocomplete.dart
```diff
@@ -615,6 +596,59 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
...
+      wildcardMentions.add(WildcardMentionAutocompleteResult(
+        wildcard: Wildcard.everyone));
+    } else if (isChannelOrTopicNarrow) {
+      if (query.testWildcard(Wildcard.channel) && isChannelWildcardAvailable) {
+        wildcardMentions.add(WildcardMentionAutocompleteResult(
+          wildcard: Wildcard.channel));
+      } else if (query.testWildcard(Wildcard.stream)) {
+        wildcardMentions.add(WildcardMentionAutocompleteResult(
+          wildcard: Wildcard.stream));
+      }
```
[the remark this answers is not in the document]
sm-sayedi: Yeah, they both differ, but that's actually what the web does. From my point of view, when a user types a query, they want to see the matched option in the autocomplete. So for the query "stream", they'll naturally look for an option with a (sub)string "stream", but if an option with the label "channel" appears at the top, I think it will be a little bit confusing. 🙂

sm-sayedi:
Thank you @chrisbobbe for the thorough review. All comments addressed. Please have a look.

lib/model/autocomplete.dart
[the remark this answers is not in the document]
alya: I'm not sure what this PR currently does, but web looks like this:

Screenshot 2025-01-21 at 15 45 30

And selecting that option inserts an @ channel metion.

I'm pretty happy with that, as I think it's pretty clear for the user, but we're still working towards training folks to use the current channel terminology.

chrisbobbe:
CI is failing, could you take a look please? 🙂

lib/model/autocomplete.dart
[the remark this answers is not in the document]
sm-sayedi: This PR does exactly the same thing as the web. 🙂

sm-sayedi:
Thanks for pointing it out. The CI failure was related to #1290; just rebased on top of the main. PTAL.

chrisbobbe:
There's an analyzer failure at this commit, could you take a look?

```
autocomplete test: Make `setupToComposeInput` accept `narrow` param
```

```
Running analyze...
Analyzing zulip-flutter...                                              

  error • The argument type 'TopicName' can't be assigned to the parameter type 'String?'.  •
         test/widgets/autocomplete_test.dart:64:57 • argument_type_not_assignable
  error • The argument type 'TopicName' can't be assigned to the parameter type 'String'.  •
         test/widgets/autocomplete_test.dart:102:69 • argument_type_not_assignable

2 issues found. (ran in 1.6s)
```

lib/model/autocomplete.dart
```diff
@@ -684,19 +728,26 @@ abstract class ComposeAutocompleteQuery extends AutocompleteQuery {
...
   @override
   MentionAutocompleteView initViewModel(PerAccountStore store, Narrow narrow) {
     return MentionAutocompleteView.init(store: store, narrow: narrow, query: this);
   }
 
+  bool testWildcardOption(WildcardMentionOption wildcardOption) {
+    // TODO(#237): match insensitively to diacritics
+    return wildcardOption.localizedCanonicalString(localizations)
+      .contains(_lowercase);
+  }
```
chrisbobbe, line 747: This should also check for a match in `wildcardOption.canonicalString`, like in zulip-mobile; not only `localizedCanonicalString`. I had this in mind with https://github.com/zulip/zulip-flutter/pull/889#discussion_r1896021336 but didn't make it super explicit. 🙂 Because `canonicalString` is the string that'll actually be inserted into the message.

lib/model/compose.dart
```diff
@@ -5,6 +5,31 @@ import 'internal_link.dart';
...
+  channel(canonicalString: 'channel'),
+  // Deprecated in FL 247. Empirically, current servers (FL 330)
+  // still parse "@**stream**" in messages though.
+  stream(canonicalString: 'stream'),
+  topic(canonicalString: 'topic'); // New in FL 224.
+
+  /// The string identifying this option (e.g. "all" as in "@**all**").
+  // If servers want to localize this, we can compute it
+  // dynamically from server data instead of hard-coding it.
+  const WildcardMentionOption({required this.canonicalString});
```
chrisbobbe, line 26: I think this dartdoc is meant to go on the `canonicalString` field rather than the constructor?

…Ah I see this came from some code I proposed 😅

lib/model/compose.dart
```diff
@@ -101,18 +126,33 @@ String wrapWithBacktickFence({required String content, String? infoString}) {
   return resultBuffer.toString();
 }
 
-/// An @-mention, like @**Chris Bobbe|13313**.
+/// An @user-mention, like @**Chris Bobbe|13313**.
```
chrisbobbe, line 129: nit: "A @user-mention"

lib/model/compose.dart
```diff
@@ -101,18 +126,33 @@ String wrapWithBacktickFence({required String content, String? infoString}) {
...
 /// through all users; avoid it in performance-sensitive codepaths.
-String mention(User user, {bool silent = false, Map<int, User>? users}) {
+String userMention(User user, {bool silent = false, Map<int, User>? users}) {
   bool includeUserId = users == null
     || users.values.where((u) => u.fullName == user.fullName).take(2).length == 2;
 
   return '@${silent ? '_' : ''}**${user.fullName}${includeUserId ? '|${user.userId}' : ''}**';
 }
 
+/// An @wildcard-mention, like @**channel**.
```
chrisbobbe, line 141: nit: "A @wildcard-mention"

lib/widgets/autocomplete.dart
```diff
@@ -223,28 +230,57 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
...
+            : localizations.wildcardMentionStreamDescription,
+      WildcardMentionOption.channel => localizations.wildcardMentionChannelDescription,
+      WildcardMentionOption.stream => isChannelWildcardAvailable
+        ? localizations.wildcardMentionChannelDescription
+        : localizations.wildcardMentionStreamDescription,
+      WildcardMentionOption.topic => localizations.wildcardMentionTopicDescription,
+    };
+    return Text.rich(TextSpan(text: '${wildcard.canonicalString} ', children: [
+      TextSpan(text: description, style: TextStyle(fontSize: 12,
+        color: Colors.black.withValues(alpha: 0.8)))]));
```
chrisbobbe, line 282: I assume we won't want to use a space character to separate the wildcard text and its description:

image

but that change can wait until a later pass where we align this with the Figma (pending #995), like https://github.com/zulip/zulip-flutter/pull/889/files#r1896061825

lib/model/autocomplete.dart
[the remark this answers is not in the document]
chrisbobbe: Sure, yeah, that reasoning makes sense :)

chrisbobbe (commented):
Thanks, I think this is quite close! Small comments below.

lib/model/compose.dart
gnprice: I'd actually pronounce this as "an at-user mention" so would use "an".

(But I think it could be clearer with a few more words: "An @-mention of an individual user", similar to the doc on UserMentionAutocompleteResult.)

sm-sayedi:
Thanks @chrisbobbe for the review. Changes pushed.

chrisbobbe:

chrisbobbe:
While reviewing, I noticed a new bug, but it's caused by an upstream change, not the changes here; filed as #1310.

lib/model/compose.dart
```diff
@@ -5,6 +5,31 @@ import 'internal_link.dart';
...
+///   https://zulip.com/help/mention-a-user-or-group
+///   https://zulip.com/help/dm-mention-alert-notifications
+enum WildcardMentionOption {
+  all(canonicalString: 'all'),
+  everyone(canonicalString: 'everyone'),
+  channel(canonicalString: 'channel'),
+  // Deprecated in FL 247. Empirically, current servers (FL 330)
+  // still parse "@**stream**" in messages though.
+  stream(canonicalString: 'stream'),
+  topic(canonicalString: 'topic'); // New in FL 224.
```
gnprice, line 21:
```suggestion
  topic(canonicalString: 'topic'); // TODO(server-8): New in FL 224.
```

The TODO-server comment is how we make sure we can systematically clean out these historical comments in the future when they're no longer relevant. Basically all references to a Zulip feature level (as well as anything else that will become obsolete when we drop support for some old Zulip Server version) should have a TODO-server comment that will make sure we find it.

lib/model/compose.dart
```diff
@@ -5,6 +5,31 @@ import 'internal_link.dart';
...
+
+  const WildcardMentionOption({required this.canonicalString});
+
+  /// The string identifying this option (e.g. "all" as in "@**all**").
+  // If servers want to localize this, we can compute it
+  // dynamically from server data instead of hard-coding it.
+  final String canonicalString;
```
gnprice, line 28: I don't 100% understand this comment (the `//` part), but I think it's speculating about a possible future change to the API. We can leave that out — if the API changes, we'll deal with it then.

lib/model/compose.dart
```diff
@@ -101,18 +126,33 @@ String wrapWithBacktickFence({required String content, String? infoString}) {
...
+  required PerAccountStore store,
+}) {
+  final isChannelWildcardAvailable = store.account.zulipFeatureLevel >= 247; // TODO(server-9)
+  assert(isChannelWildcardAvailable || wildcardOption != WildcardMentionOption.channel);
+  final isTopicWildcardAvailable = store.account.zulipFeatureLevel >= 224; // TODO(sever-8)
+  assert(isTopicWildcardAvailable || wildcardOption != WildcardMentionOption.topic);
+
+  final name = wildcardOption == WildcardMentionOption.stream && isChannelWildcardAvailable
+    ? WildcardMentionOption.channel.canonicalString
+    : wildcardOption.canonicalString;
```
gnprice, line 152: All these equality comparisons to different enum values feel like they'd be clearer as one `switch` statement that lists all the values.

Could start with `String name = wildcardOption.canonicalString;` and then have the `switch` mutate it in one case, to avoid repeating the common version and have the exception stand out.

lib/widgets/autocomplete.dart
```diff
@@ -223,28 +230,57 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
...
       padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
       child: Row(children: [
         avatar,
         const SizedBox(width: 8),
-        Text(label),
+        label,
       ]));
   }
+
+  Widget wildcardLabel(WildcardMentionOption wildcard, {
```
gnprice, line 261: nit: put this method before `build`; and mark it private

lib/widgets/autocomplete.dart
gnprice, line 282: Does this color work equally well in dark theme?

Let's move it to DesignVariables so the theme behavior is explicit.

gnprice (commented):
Thanks @sm-sayedi for carrying this through! And thanks @chrisbobbe for the previous reviews.

Generally this looks good. Some comments below. I haven't yet read the test changes, because I think a couple of these comments will lead to some simplifications there.

lib/widgets/autocomplete.dart
sm-sayedi: Thanks for pointing it out. It didn't look good in the dark theme. Replaced it with the following, and now it looks good:
```dart
color: DefaultTextStyle.of(context).style.color?.withValues(alpha: 0.8)
```
Does it still need to be defined in `DesignVariables`?

lib/model/autocomplete.dart
```diff
@@ -6,13 +6,15 @@ import 'package:flutter/services.dart';
...
+import '../generated/l10n/zulip_localizations.dart';
 import '../widgets/compose_box.dart';
+import 'compose.dart';
 import 'emoji.dart';
 import 'narrow.dart';
 import 'store.dart';
 
 extension ComposeContentAutocomplete on ComposeContentController {
-  AutocompleteIntent<ComposeAutocompleteQuery>? autocompleteIntent() {
+  AutocompleteIntent<ComposeAutocompleteQuery>? autocompleteIntent(ZulipLocalizations localizations) {
```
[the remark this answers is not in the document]
sm-sayedi: In the current revision, the base `AutocompleteView` class holds a reference to `ZulipLocalizations`. With that, all the subclasses (should) retrieve the instance through `FooAutocompleteView.init`. What if we only have the `ZulipLocalizations` property in `MentionAutocompleteView`, not in the base class? I think for now, having `localizations` only makes sense in `MentionAutocompleteView` not the other two subclasses. Then if in the future, it makes sense for the others too, we can refactor the code and move `ZulipLocalizations localizations` to the base class, just like in this revision?!

sm-sayedi:
Thanks @gnprice for the review. Changes pushed, with two comments, one just above this comment and another in [here](https://github.com/zulip/zulip-flutter/pull/889#discussion_r1934082762). PTAL.

lib/model/autocomplete.dart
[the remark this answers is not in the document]
gnprice: Yeah, I think having the localizations live either on the base class or the subclass would be fine — whichever seems easier and/or produces cleaner code right now. I'm not sure other types of autocomplete will ever need localizations; if they do, then add you say, we can refactor then.

lib/widgets/autocomplete.dart
gnprice: Yeah, that's fine.

In principle an expression like that should still go into DesignVariables, because it still might vary by theme — for example, if in the future we add "high contrast" themes, they might want to keep the alpha at 1, or set it to 90% or something.

But that isn't a feature we have on the horizon. And these design details will get replaced soon anyway with #995 / #913. So with this solution which works for the existing themes, it's not worth the overhead to add to DesignVariables.

lib/model/compose.dart
```diff
@@ -5,6 +5,28 @@ import 'internal_link.dart';
...
+/// The available user wildcard mention options,
+/// known to the server as [canonicalString].
+///
+/// See API docs:
+///   https://zulip.com/api/message-formatting#mentions-and-silent-mentions
+enum WildcardMentionOption {
+  all(canonicalString: 'all'),
+  everyone(canonicalString: 'everyone'),
+  channel(canonicalString: 'channel'),
+  // Deprecated in FL 247. Empirically, current servers (FL 330)
```
gnprice, line 17: needs a TODO-server comment (similar to https://github.com/zulip/zulip-flutter/pull/889#discussion_r1931389011 )

lib/model/autocomplete.dart
```diff
@@ -615,6 +605,59 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
...
+      if (tryOption(WildcardMentionOption.all)) break all;
+      if (tryOption(WildcardMentionOption.everyone)) break all;
+      if (isComposingChannelMessage) {
+        final isChannelWildcardAvailable = store.account.zulipFeatureLevel >= 247; // TODO(server-9)
+        if (isChannelWildcardAvailable && tryOption(WildcardMentionOption.channel)) break all;
+        if (tryOption(WildcardMentionOption.stream)) break all;
+      }
+    }
+
+    final isTopicWildcardAvailable = store.account.zulipFeatureLevel >= 224; // TODO(sever-8)
```
gnprice, line 634:
```suggestion
    final isTopicWildcardAvailable = store.account.zulipFeatureLevel >= 224; // TODO(server-8)
```

The spelling is important for these to serve their purpose of being findable via grep :slightly_smiling_face: 

(there's one other like this)

lib/widgets/autocomplete.dart
```diff
@@ -168,19 +171,22 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
   ComposeContentController get controller => super.controller as ComposeContentController;
 
   @override
-  AutocompleteIntent<ComposeAutocompleteQuery>? autocompleteIntent() => controller.autocompleteIntent();
+  AutocompleteIntent<ComposeAutocompleteQuery>? autocompleteIntent(
+    BuildContext context) => controller.autocompleteIntent();
```
gnprice, line 175: These changes are no longer needed, since the query and intent no longer carry the localizations.

lib/model/autocomplete.dart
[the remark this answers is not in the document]
gnprice: OK, and now having read this revision: I agree, moving it to the subclass `MentionAutocompleteView` will make things cleaner.

gnprice:
Thanks for the revision! I've read through the whole `lib/` changes again, and just a few comments above. I'll take a look at the tests next.

test/model/autocomplete_test.dart
```diff
@@ -752,43 +756,129 @@ void main() {
...
+
+      final testCases = [
+        ('',          channelNarrow, [WildcardMentionOption.all, WildcardMentionOption.topic]),
+        ('',          topicNarrow,   [WildcardMentionOption.all, WildcardMentionOption.topic]),
+        ('',          dmNarrow,      [WildcardMentionOption.all]),
+
+        ('c',         channelNarrow, [WildcardMentionOption.channel, WildcardMentionOption.topic]),
+        ('ch',        topicNarrow,   [WildcardMentionOption.channel]),
+        ('str',       channelNarrow, [WildcardMentionOption.stream]),
+        ('e',         topicNarrow,   [WildcardMentionOption.everyone]),
```
gnprice, line 812: Hmm, I guess with the English names there's not actually a way to test the fact that .channel is preferred over .stream — the only letters they have in common are "a" and "e", and those queries also match the more-preferred .all and .everyone respectively.

test/model/autocomplete_test.dart
```diff
@@ -752,43 +756,129 @@ void main() {
...
+      ];
+
+      for (final (String query, Narrow narrow, List<WildcardMentionOption> wildcardOptions) in testCases) {
+        test('for "$query" query in ${narrow.runtimeType} narrow wildcards are $wildcardOptions', () async {
+          check(getWildcardOptionsFor(query, narrow: narrow)).deepEquals(wildcardOptions);
+        });
+      }
+
+      test('no wildcards for a silent mention', () {
+        check(getWildcardOptionsFor('', isSilent: true, narrow: channelNarrow)).isEmpty();
```
gnprice, line 832: nit: line too long: key information (the `.isEmpty`) past 80 columns:
```suggestion
        check(getWildcardOptionsFor('', isSilent: true, narrow: channelNarrow))
          .isEmpty();
```

test/model/autocomplete_test.dart
```diff
@@ -752,43 +756,129 @@ void main() {
...
         const narrow = StarredMessagesNarrow();
-        check(() => MentionAutocompleteView.init(store: store, narrow: narrow,
-                      query: MentionAutocompleteQuery('')))
+        check(() => MentionAutocompleteView.init(store: store, localizations: localizations,
+                      narrow: narrow, query: MentionAutocompleteQuery('')))
           .throws<AssertionError>();
       });
     });
 
+    group('computeWildcardMentionResults', () {
```
gnprice, line 781: Let's also have a couple of test cases exercising the fact that queries match against the localized names of "all", "channel", etc. After all, you've done significant work for the plumbing to make that possible — would a shame to let it get accidentally broken in the future because we lacked a test for it :slightly_smiling_face: 

Since you've included Arabic translations for the new strings, you could use those for those test cases.

test/model/autocomplete_test.dart
```diff
@@ -812,20 +902,27 @@ void main() {
...
 
-      // Check the ranking of the full list of users.
+      // Check the ranking of the full list of mentions.
       // The order should be:
-      // 1. Users most recent in the current topic/stream.
-      // 2. Users most recent in the DM conversations.
-      // 3. Human vs. Bot users (human users come first).
-      // 4. Alphabetical order by name.
-      check(await getResults(topicNarrow, MentionAutocompleteQuery('')))
+      // 1. Wildcards.
```
gnprice, line 907:
```suggestion
      // 1. Wildcards before individual users.
```

This numbered list isn't really a list of categories from first to last in the results, but more a list of ranking criteria from highest to lowest priority in their effect on the order of results. (E.g. compare the "human vs. bot users" item.)

test/model/autocomplete_test.dart
```diff
@@ -812,20 +902,27 @@ void main() {
...
+      // 2. Users most recent in the current topic/stream.
+      // 3. Users most recent in the DM conversations.
+      // 4. Human vs. Bot users (human users come first).
+      // 5. Users by name alphabetical order.
+      final results1 = await getResults(topicNarrow, MentionAutocompleteQuery(''));
+      check(getWildcardOptionsFromResults(results1.take(2)))
+        .deepEquals([WildcardMentionOption.all, WildcardMentionOption.topic]);
+      check(getUsersFromResults(results1.skip(2)))
         .deepEquals([1, 5, 4, 2, 7, 3, 6]);
```
gnprice, line 916: Cool, this structure works.

FYI the more general solution for this sort of situation — the one we'd need to go for if this got a bit more complicated, for example if the two types of results were interleaved ­— is to use `deepEquals` on the original list and pass an expected list of `Condition` values, instead of values to do equality checks on. You can find some examples in `test/notifications/display_test.dart`, like this:
```dart
      check(testBinding.androidNotificationHost.activeNotifications).deepEquals(<Condition<Object?>>[
        conditionActiveNotif(data1, conversationKey1),
        conditionSummaryActiveNotif(expectedGroupKey),
        conditionActiveNotif(data2, conversationKey2),
      ]);
```

where `conditionSummaryActiveNotif` and `conditionActiveNotif` are helpers local to that test file.

test/widgets/autocomplete_test.dart
```diff
@@ -178,6 +196,69 @@ void main() {
...
+      checkWildcardShown(WildcardMentionOption.channel, store, expected: false);
+      checkWildcardShown(WildcardMentionOption.topic, store, expected: false);
+      checkWildcardShown(WildcardMentionOption.all, store, expected: false);
+      checkWildcardShown(WildcardMentionOption.everyone, store, expected: false);
+      checkWildcardShown(WildcardMentionOption.stream, store, expected: false);
+
+      // Then a new autocomplete intent brings up options again
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(composeInputFinder, 'hello @chan');
+      await tester.enterText(composeInputFinder, 'hello @channel');
```
gnprice, line 242: I think this and the remaining steps of the test can be left out here — they're basically redundant with the test case above, since the logic controlling this doesn't depend on the particular type of result being shown.

test/widgets/autocomplete_test.dart
```diff
@@ -178,6 +196,69 @@ void main() {
...
+
+      // Options are filtered correctly for query
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(composeInputFinder, 'hello @');
+      await tester.enterText(composeInputFinder, 'hello @c');
+      await tester.pumpAndSettle(); // async computation; options appear
+
+      checkWildcardShown(WildcardMentionOption.channel, store, expected: true);
+      checkWildcardShown(WildcardMentionOption.topic, store, expected: true);
+      checkWildcardShown(WildcardMentionOption.all, store, expected: false);
```
gnprice, line 223: (By contrast this step of the test is important, because it lets us confirm that the options do get shown — and have the pieces that `checkWildcardShown` checks for — and that no errors get thrown when doing so.

IOW the `enterText` and pump above exercise the wildcard-specific logic that's been added to the `_MentionAutocompleteItem` widget, and these checks confirm that that logic successfully got run and produced the right result.)

gnprice (commented):
OK, and read through the tests too. Comments below.

test/model/autocomplete_test.dart
sm-sayedi: Yeah, that's right! But in the new revision, it is exercised in the localized test cases at least!

lib/widgets/autocomplete.dart
```diff
@@ -223,26 +231,55 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
...
+        : isChannelWildcardAvailable
+            ? localizations.wildcardMentionChannelDescription
+            : localizations.wildcardMentionStreamDescription,
+      WildcardMentionOption.channel => localizations.wildcardMentionChannelDescription,
+      WildcardMentionOption.stream => isChannelWildcardAvailable
+        ? localizations.wildcardMentionChannelDescription
+        : localizations.wildcardMentionStreamDescription,
+      WildcardMentionOption.topic => localizations.wildcardMentionTopicDescription,
+    };
+    return Text.rich(TextSpan(text: '${wildcardOption.canonicalString} ', children: [
```
sm-sayedi, line 258: What about replacing `wildcardOption.canonicalString` with `wildcardOption.localizedCanonicalString`?! I think `canonicalString` would cuase some confusions when the app is used in other languages. For example, when the app is used in arabic and someone tries to choose `@all` or `@everyone` wildcards by writing it's translation in the compose box, it will not be straightforward which word to write down in their language to match "all" or "everyone" becuase there may be different synonyms of these words! And the translated description of these two wildcards will also not provide a hint for that word, becuase it always says: "Notify channel/stream".

sm-sayedi:
Thanks @gnprice for the review. Pushed the changes. Please have a look.

lib/widgets/autocomplete.dart
gnprice: Yeah, could be good. What does web do?

sm-sayedi: Web shows the English version of wildcards and only matches for the English strings!

lib/model/localizations.dart
```diff
@@ -3,4 +3,8 @@ import '../generated/l10n/zulip_localizations.dart';
 abstract final class GlobalLocalizations {
   static ZulipLocalizations zulipLocalizations =
     lookupZulipLocalizations(ZulipLocalizations.supportedLocales.first);
+
+  static ZulipLocalizations zulipLocalizationsArabic =
+    lookupZulipLocalizations(ZulipLocalizations.supportedLocales
+      .firstWhere((locale) => locale.languageCode == 'ar'));
```
gnprice, line 9: This is for use only in tests, so it should go in test code instead. (Probably just in the specific test file that's using it — other tests that use a particular translation might use other languages.)

gnprice: (The existing field next to this one refers to whatever language the user has chosen — it's initialized here to English, but gets updated by ZulipApp as soon as the app starts up far enough to know the user's preference.)

lib/model/autocomplete.dart
```diff
@@ -615,6 +602,59 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
     return userAName.compareTo(userBName); // TODO(i18n): add locale-aware sorting
   }
 
+  void computeWildcardMentionResults({
```
gnprice, line 605: nit in commit message:

    autocomplete: Support @-wildcard in user-mention autocomplete
    
    The implementation logic is similar to the zulip-mobile implementation:
      https://github.com/zulip/zulip-mobile/blob/a115df1f71c9dc31e9b41060a8d57b51c
017d786/src/autocomplete/WildcardMentionItem.js

There's a few other things going on in that file, so this link can be made more precise with a line-number range:
https://github.com/zulip/zulip-mobile/blob/a115df1f71c9dc31e9b41060a8d57b51c017d786/src/autocomplete/WildcardMentionItem.js#L92-L140

lib/model/autocomplete.dart
```diff
@@ -419,42 +421,29 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
...
     required this.sortedUsers,
+    required this.localizations,
   });
 
   factory MentionAutocompleteView.init({
     required PerAccountStore store,
-    required Narrow narrow,
     required MentionAutocompleteQuery query,
+    required Narrow narrow,
+    required ZulipLocalizations localizations,
```
gnprice, line 431: nit: logically the store and the localizations are a lot alike — they're more or less global context that the whole app, or a large swath of it, is operating within — so let's keep them together. Then the narrow is somewhat more-specific context, describing the whole current page, and the query is more specific than that, so should follow in that order.

lib/model/autocomplete.dart
```diff
@@ -419,42 +421,29 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
     required super.query,
     required this.narrow,
     required this.sortedUsers,
+    required this.localizations,
```
gnprice, line 424: nit: similarly, put this above the narrow and sorted users

lib/model/autocomplete.dart
gnprice: Then there's a few ways we've made this differ from zulip-mobile, so let's briefly mention those:
* suppress wildcards on silent mentions
* match "stream" queries even on new servers (but emit `@**channel**`)
* possibly something I'm forgetting

gnprice (commented):
Thanks! Looks good, with just a few small comments below. There's also a UX question you asked above at https://github.com/zulip/zulip-flutter/pull/889#discussion_r1936195759 (but it won't block merge).

lib/widgets/autocomplete.dart
gnprice: Ah I see, so it doesn't even let you search for them by the names in your language? I guess that must have been something we thought of for zulip-mobile.

This UX will be fine to merge — it's the same as in the existing mobile app in this respect, and that one difference from web (the ability to match your language's term) seems like clearly an improvement. I agree it seems suboptimal that there's no direct hint of what term in your language will match an option.

gnprice: If you want to pursue making that UX change, the best place to discuss would be in a chat thread so more people can see it — either in #mobile, or in #feedback if you think it should apply to web too (and maybe web should gain the feature mobile has where you can search by your language's term in the first place).

lib/model/localizations.dart
gnprice: I also noticed there were no docs in this file, so I just added some:
173138598 i18n [nfc]: Add doc on GlobalLocalizations.zulipLocalizations

Hopefully that makes the story clearer.

lib/widgets/autocomplete.dart
chrisbobbe: > I guess that must have been something we thought of for zulip-mobile.

Correct

sm-sayedi:
Thanks @gnprice. Pushed a new revision.

gnprice:
Thanks again for all your work on this! All looks great now — merging.

PR #895, opened 2024-08-19
test [nfc]: Cut redundant checks-extensions from compose_box_checks

gnprice:
(A tiny cleanup I happened across this while writing another change.)

These types live in lib/model/autocomplete.dart, so the corresponding checks-extensions live in test/model/autocomplete_checks.dart (which indeed has them).


chrisbobbe:
Thanks! Merged.

PR #896, opened 2024-08-19
autocomplete [nfc]: Give subclasses control of computeResults

gnprice:
This gives AutocompleteView subclasses the flexibility to have an implementation that's more than a single loop through candidates.

That will be useful for #234 / #889, /cc @sm-sayedi.

To keep the tricky aspects of the main search loop encapsulated, we move them into a helper function `filterCandidates` which those `computeResults` implementations can call.


chrisbobbe:
Thanks, LGTM! Merging.

PR #899, opened 2024-08-20
deps: Upgrade Flutter to 3.25.0-1.0.pre.41

PIG208:


rajveermalviya:
I just tested with the latest Flutter main, and I'm happy to report that (on my device) animations and transitions are smooth again, with no rendering artifacts on the Impeller rendering backend. It seems those issues have been fixed in this update, though not sure which commits addressed them.

gnprice:
Thanks @PIG208, and thanks @rajveermalviya for that testing! Glad to hear they've fixed those Impeller issues.

Merging. I revised the README change and added an update to the pubspec.yaml comment.

PR #765, opened 2024-06-26
ui: Add animation support for the edited/moved marker.

PIG208:
This is a sequel to #762 that animates the swipe gesture.

PR #900, opened 2024-08-20
msglist: Implement new edited/moved label design.

PIG208:
This replaces the gutter edit state marker design. Since it leads to message displacement, a previous test case is no longer applicable.

We also inline part of what was previously in edit_state_marker.dart, because the new design simplifies the implementation.

Figma design: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3038-56393&t=WMT80mwuFaruNbVf-1


PIG208:
Wasn't able to reopen #765 after accidentally closing it:

image

Here's the new PR for the new design.

PIG208:
<details>

<summary>Screenshots
</summary>

| light | dark | 
| - | - |
| Screenshot from 2024-08-20 17-43-19 | Screenshot from 2024-08-20 17-43-06 |


</details>

PIG208:
Rebased and simplify the implementation a bit.

gnprice:
> Wasn't able to reopen #765 after accidentally closing it:

FWIW for next time: I think the solution there is to push the commits you want back to the PR's branch, and then re-open the PR.

(In particular the error message there says "There are no new commits on the … branch", i.e. the branch is empty compared to main. You can fix that by pushing to fix what's in the branch.)

gnprice:
Though in this case, since the implementation is radically different from #765, a fresh PR is probably cleanest anyway.

PIG208:
> FWIW for next time: I think the solution there is to push the commits you want back to the PR's branch, and then re-open the PR.

I do recall running into the same issue a few years back. Couldn't find the link right now.

My attempted fix this time was amending the latest commit, and adding an empty commit on top of it. It's interesting that it didn't work. Will look into this next time (without mistakenly pushing to upstream!)

lib/widgets/message_list.dart
```diff
@@ -86,7 +84,7 @@ class MessageListTheme extends ThemeExtension<MessageListTheme> {
     required this.dateSeparator,
     required this.dateSeparatorText,
     required this.dmRecipientHeaderBg,
-    required this.editedMovedMarkerCollapsed,
+    required this.editStateLabel,
```
chrisbobbe, line 87: It looks like there's a Figma design variable for this, `labelEdited`, so let's use that and put it in `DesignVariables`, sorted alphabetically in the default section (not marked `// Not exactly from the Figma design […]` or `// Not named variables in Figma […]`).

lib/widgets/message_list.dart
```diff
@@ -1262,15 +1270,24 @@ class MessageWithPossibleSender extends StatelessWidget {
...
                   if ((message.reactions?.total ?? 0) > 0)
-                    ReactionChipsList(messageId: message.id, reactions: message.reactions!)
+                    ReactionChipsList(messageId: message.id, reactions: message.reactions!),
+                  if (editStateText != null)
+                    Text(editStateText, textAlign: TextAlign.end,
+                      style: TextStyle(
+                        color: messageListTheme.editStateLabel,
+                        fontFamily: 'Source Sans 3',
+                        fontSize: 12,
+                        height: (12 / 12))),
```
chrisbobbe, line 1290: I guess the Figma isn't clear whether the reaction chips or the marker should come first, vertically. How about putting the marker first? In the case where there are dozens of reaction chips that wrap onto many lines, I think it would be neater and clearer if the EDITED or MOVED marker were above them instead of below them.

lib/widgets/message_list.dart
```diff
@@ -1262,15 +1270,24 @@ class MessageWithPossibleSender extends StatelessWidget {
...
                 children: [
                   MessageContent(message: message, content: item.content),
                   if ((message.reactions?.total ?? 0) > 0)
-                    ReactionChipsList(messageId: message.id, reactions: message.reactions!)
+                    ReactionChipsList(messageId: message.id, reactions: message.reactions!),
+                  if (editStateText != null)
+                    Text(editStateText, textAlign: TextAlign.end,
+                      style: TextStyle(
+                        color: messageListTheme.editStateLabel,
+                        fontFamily: 'Source Sans 3',
```
chrisbobbe, line 1288: I think we'll get 'Source Sans 3' automatically here, like in most other places, because this is rendered on a `Material` (via `Scaffold`, I think), and `Material` [applies `TextTheme.bodyMedium`](https://github.com/flutter/flutter/blob/a09941dcf1bae5196c8107dfced5d9c02c121905/packages/flutter/lib/src/material/material.dart#L467) to its children via `AnimatedDefaultTextStyle`, and we've set `TextTheme.bodyMedium` to include 'Source Sans 3' (see `zulipTypography` in lib/widgets/text.dart).

chrisbobbe: Looks like we're also redundantly passing 'Source Sans 3' where we render the sender name in `MessageWithPossibleSender`. Removing that (perhaps with a link to my explanation here, or quoting/paraphrasing it) would be a helpful cleanup in a separate NFC commit.

lib/widgets/message_list.dart
chrisbobbe, line 1290:
```suggestion
                        height: (12 / 12),
                        letterSpacing: proportionalLetterSpacing(context, 0.05, baseFontSize: 12),
```

I notice the Figma has 0.6px letter spacing, but if we use `proportionalLetterSpacing`, it'll grow and shrink with the system font-size setting, and that seems like a nice refinement.

assets/l10n/app_en.arb
```diff
@@ -529,5 +529,13 @@
   "manyPeopleTyping": "Several people are typing…",
   "@manyPeopleTyping": {
     "description": "Text to display when there are multiple users typing."
+  },
+  "messageIsEditedLabel": "EDITED",
+  "@messageIsEditedLabel": {
+    "description": "Label for an edited message."
```
chrisbobbe, line 535:
```suggestion
    "description": "Label for an edited message. (Use ALL CAPS for cased alphabets: Latin, Greek, Cyrillic, etc.)"
```

(similarly below)

chrisbobbe (commented):
Thanks! Small comments below.

assets/l10n/app_en.arb
PIG208: One thing I was wondering about when implementing this, is that whether we should programmatically upper case the letters and translate the label normally, or just do it the current way.

We do have a precedent: "loginMethodDivider": "OR". But I can also see the appeal to using `i18n.toUpperCase` for something purely for styling purposes (similar to the `text-transform: uppercase;` approach as one could do in web), because in this case we aren't using uppercase letters to convey emphasis/exaggeration in the writing.

PIG208:
<details>

| light | dart |
| -     |  -   |
| Screenshot from 2024-08-26 14-23-01 | image |

</details>

lib/widgets/message_list.dart
PIG208: Yeah that sounds good to me.

PIG208: Just to be sure, started a [disccusion on CZO](https://chat.zulip.org/#narrow/stream/48-mobile/topic/check-x.20icons.20in.20left.20margin/near/1931552).

assets/l10n/app_en.arb
PIG208: Started a discussion [here](https://chat.zulip.org/#narrow/stream/58-translation/topic/Requiring.20translation.20in.20uppercase.20for.20applicable.20languages/near/1931579). My guess is that it might be preferred to just use uppercase letters, but I'm curious if there are better ways.

PIG208:
Implemented the suggested changes. Thanks for the review @chrisbobbe! Started some CZO discussions on things that I have questions about.

lib/widgets/message_list.dart
chrisbobbe: Oh interesting, so yeah, it looks like the 5% is expressed in the Figma too :) thanks for noticing that.

chrisbobbe:
Thanks! This LGTM except the CZO discussion seems like it's settling on having the marker come after the reactions list instead of before. Marking for Greg's review once he's back from vacation.

PIG208:
Pushed to move the marker after the reactions list.

PIG208:
Updated the PR to share the final row between reaction chips and the edit state label mentioned [here](https://chat.zulip.org/#narrow/stream/48-mobile/topic/check-x.20icons.20in.20left.20margin/near/1936724).

If necessary, we could add an RTL test case later for the new commit if we like this approach.

PIG208:
<details>
<summary>Details</summary>

| Barely enough space | Not enough space |
| - | - |
| Screenshot from 2024-09-04 23-35-25 | Screenshot from 2024-09-04 23-35-29 |


</details>

lib/widgets/message_list.dart
```diff
@@ -1268,15 +1275,27 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                      if ((message.reactions?.total ?? 0) > 0)
+                        ReactionChipsList(
+                          messageId: message.id,
+                          reactions: message.reactions!,
+                          // This copy of `editStateLabel` is invisible.
+                          // See docs for more details.
+                          trailingPlaceholder: editStateLabel),
+                    ]),
+                  if (editStateLabel != null)
+                    PositionedDirectional(bottom: 0, end: 0, child: editStateLabel),
```
gnprice, line 1298: What happens if there are no reactions, but the message's own text ends with a full line? I think in this version that will end up having an overlap of the "EDITED" marker and the message text.

lib/widgets/message_list.dart
```diff
@@ -1251,14 +1241,31 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+    switch (message.editState) {
+      case MessageEditState.edited:
+        editStateText = localizations.messageIsEditedLabel;
+      case MessageEditState.moved:
+        editStateText = localizations.messageIsMovedLabel;
+      case MessageEditState.none:
+    }
+
+    final editStateLabel = editStateText == null ? null : Text(editStateText,
+      textAlign: TextAlign.end,
```
gnprice, line 1262: nit:
```suggestion
    final editStateLabel = editStateText == null ? null
      : Text(editStateText,
          textAlign: TextAlign.end,
```

That way the `Text` constructor name, and `editStateText` which holds the actual text content, stay more visible and next to the other arguments to `Text`.

gnprice (commented):

Let's skip for now the complexity of trying to squash the edited/moved marker onto the same line as the end of the message. I think this version doesn't yet quite work — see below — and I'm also a little concerned about making the layout complicated, if for example the user is reading a long thread that got moved and so all the messages have markers. If people really find the extra vertical space annoying, we can revisit that.

Otherwise this all looks good, with just one nit below.

lib/widgets/message_list.dart
PIG208: Start `textAlign` in a new line for the now inlined `editStateLabel` because we don't need to render another copy now.

lib/widgets/message_list.dart
PIG208: Hmm, yeah, that could totally happen. It does look like that furthering this implementation to fix that as well can make it more complicated, unless we get a different solution.

PIG208:
Dropped the commit 783b5a2bd41e3bb9254cde64961936d6e41405db, and pushed an update. Thanks!

gnprice:
Thanks! Looks good; merging.

Issue #267, opened 2023-08-14
Disallow connecting to unsupported ancient servers


gnprice:
Much like the feature in the zulip-mobile RN app we added as zulip/zulip-mobile#5633 / zulip/zulip-mobile#5102.

Currently this means a minimum server version of 4.0, as documented at https://zulip.readthedocs.io/en/latest/overview/release-lifecycle.html and our README, and as reflected in our API-facing code which ignores all pre-server-4 complications. At some point we'll increase that to 5.0, then in the future to 6.0 and 7.0 and so on.

(A handy link for a record of Zulip Server release dates, for planning when we can bump that minimum: https://blog.zulip.com/tag/major-releases/ .)

gnprice:
In the UI, this will show up in two contexts: when you try to log into a new account, or when you pick an already logged-in account from the list on the initial screen of the app. In either case, if the server is too old, we'll need to show you an error and take you back.

To picture these contexts, see the first and the fourth of the handy screen captures in the PR that added this feature in zulip-mobile:
https://github.com/zulip/zulip-mobile/pull/5633
(The other two are for cases that don't exist in zulip-flutter.)

chrisbobbe:
Greg proposes ([discussion](https://chat.zulip.org/#narrow/stream/378-api-design/topic/.60zulip_version.60.20in.20server.20settings/near/1728362)) that we use the feature level to decide if the server is ancient, rather than trying to parse the version string:

> It occurs to me now that we should probably switch from making that check [in zulip-mobile] based on `zulip_version` to doing so based on `zulip_feature_level`. That's dead simple to parse (it's just an integer); and it's more foolproof for the server to generate too, as it's just quoting a number that's written directly in the server's source tree:
> ```py
> API_FEATURE_LEVEL = 237
> ```

> Originally we used `zulip_version` because `zulip_feature_level` wasn't an option — it was introduced in the 3.0 cycle, and we wanted to tell the difference between 2.1 and 2.0 and 1.9 and 1.8 and so on.
> 
> But now our minimum version is one where `zulip_feature_level` already exists, so that constraint no longer applies.
>

gnprice:
A few more details on that:
* https://github.com/zulip/zulip-flutter/issues/268#issuecomment-1915741004

gnprice:
This was blocked for a while on #458, which is why I assigned it to myself. Now that #702 is merged, that blocker is resolved.

[...]

PR #904, opened 2024-08-22
Start Requiring Zulip Server 5

PIG208:
Blocked by #267, but these are migrations that can happen in parallel. I imagine that there won't be much overhead rebasing this kind of changes.

Fixes: #268


PIG208:
The PR itself is ready, but we won't be ready to merge it until we refuse to connect to older servers. Going through some reviews will still be helpful, though.

lib/api/model/model.dart
```diff
@@ -205,7 +205,7 @@ class User {
   // bool isOwner; // obsoleted by [role]; ignore
   // bool isAdmin; // obsoleted by [role]; ignore
   // bool isGuest; // obsoleted by [role]; ignore
-  bool? isBillingAdmin; // TODO(server-5)
+  bool isBillingAdmin;
```
chrisbobbe, line 208:
```
api: Mark isBillingAdmin as requried, relying on server 5+, FL 73+

While `realm_user` update events also have this field added at the same
feature level, it remains optional.

See "Feature level 73" from Zulip API changelog:

   https://zulip.com/api/changelog

```

I would just leave out the sentence about how `isBillingAdmin` remains optional in `RealmUserUpdateEvent`. That's the normal, expected thing to happen; it matches all the other `person` fields, in that they might be null if they didn't change in the event.

Also nit in summary line: 'required"

lib/api/model/model.dart
```diff
@@ -233,14 +233,6 @@ class User {
     return (value != null && value.isNotEmpty) ? value : null;
   }
 
-  static bool _readIsSystemBot(Map<dynamic, dynamic> json, String key) {
-    // This field is absent in `realm_users` and `realm_non_active_users`,
-    // which contain no system bots; it's present in `cross_realm_bots`.
```
chrisbobbe, line 238:
```
api: Drop _readIsSystemBot, relying on server 5+, FL 83+
```

This comment still applies, right? The change made in FL 83 is just that `is_cross_realm_bot` stops being a name we have to care about. The comment can go above the `defaultValue: false`, and it seems like we should keep the test that confirms that `false` is chosen if `is_system_bot` is absent.

lib/api/model/model.dart
```diff
@@ -824,16 +824,12 @@ enum MessageEditState {
...
-      if (prevTopic != null) {
-        // TODO(server-5) pre-5.0 servers do not have the 'topic' field
-        if (topic == null) {
-          hasMoved = true;
-        } else {
-          hasMoved |= !topicMoveWasResolveOrUnresolve(topic, prevTopic);
-        }
+      if (topic != null || prevTopic != null) {
+        // Both are present if the topic was edited
+        assert(topic != null && prevTopic != null);
```
chrisbobbe, line 831: The condition in this `assert` might happen because of unexpected data from the server. To validate data from the server, we'll want code that runs in production, and reserve `assert`s to check our own invariants that we maintain within the bounds of the app. Could replace the assert and comment with:

```dart
        // Crunch-shell validation: both are present if the topic was edited
        topic as String;
        prevTopic as String;
```

chrisbobbe (commented):
Thanks for going ahead with this! This will be good to have ready to merge once #267 is done (that's on my plate right now). Small comments below.

PIG208:
Thanks! Updated the PR.

chrisbobbe:
Thanks! LGTM except bump on this part of https://github.com/zulip/zulip-flutter/pull/904#discussion_r1732067432 :

> it seems like we should keep the test that confirms that false is chosen if `is_system_bot` is absent.

PIG208:
Got that test back. Thanks for the review!

chrisbobbe:
Great! Marking for Greg's review once he's back from vacation.

gnprice:
BTW there's generally no need to push a revision when it's a pure rebase, with no merge conflicts or other changes. The update causes notifications, and doesn't really save effort later — there'll typically be another rebase anyway just before merge, and `git rebase` is a very easy step to do at merge time when there's no conflict to resolve.

(As it happens I'm just about to take a look at this PR, but it wasn't because of the notification from the revision just now; it was from scanning the list of [PRs marked for my review](https://github.com/zulip/zulip-flutter/pulls?q=is%3Apr+is%3Aopen+label%3A%22integration+review%22).)

lib/api/route/messages.dart
```diff
@@ -1,64 +1,16 @@
...
 
 part 'messages.g.dart';
 
-/// Convenience function to get a single message from any server.
-///
-/// This encapsulates a server-feature check.
-///
-/// Gives null if the server reports that the message doesn't exist.
-// TODO(server-5) Simplify this away; just use getMessage.
-Future<Message?> getMessageCompat(ApiConnection connection, {
```
gnprice, line 16: nit in commit message:

    See also: 631f4d68c58a84dff66a6a9a80322584f0d60152

Instead say:

    See also commit 631f4d68c58a84dff66a6a9a80322584f0d60152.

That way it's clear this hex blob refers to a commit (implicitly one in this same repo).

gnprice (commented):
Thanks @PIG208 for taking care of these, and thanks @chrisbobbe for the previous reviews!

Generally this looks great. Small comments below.

lib/api/model/model.dart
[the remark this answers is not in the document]
PIG208: Yeah, I think the context info is pretty helpful. While the commit summaries get longer, it is justified for this kind of migration changes.

PIG208:
Thanks for the review! Added some TODOs commits for `GetServerSettingsResult` changes, making them easy to drop, and addressed the issues from the review. This revision should be fairly readable with range-diff.

lib/api/route/realm.g.dart
```diff
@@ -20,7 +20,7 @@ GetServerSettingsResult _$GetServerSettingsResultFromJson(
           .toList(),
       zulipFeatureLevel: (json['zulip_feature_level'] as num).toInt(),
       zulipVersion: json['zulip_version'] as String,
-      zulipMergeBase: json['zulip_merge_base'] as String?,
+      zulipMergeBase: json['zulip_merge_base'] as String,
```
gnprice, line 23: This doesn't seem to have a corresponding change in realm.dart.

Presumably this came about as you made some edits and missed rerunning `build_runner` for one of them, or something like that. I am puzzled that our CI didn't complain, though — `tools/check build_runner` should have failed because this doesn't match what a freshly-generated file should say. So that would be worth trying to debug.

lib/api/route/realm.dart
```diff
@@ -34,7 +34,8 @@ class GetServerSettingsResult {
 
   final int zulipFeatureLevel;
   final String zulipVersion;
-  final String? zulipMergeBase; // TODO(server-5)
+  // TODO: Modernize this once we get to #267
+  final String? zulipMergeBase;
```
gnprice, line 38: This should keep a `TODO(server-5)` comment until we do actually modernize it. That way it naturally shows up in any future round of grepping for opportunities to modernize our API usage.

lib/api/model/web_auth.dart
```diff
@@ -7,7 +7,7 @@ import 'package:flutter/foundation.dart';
 class WebAuthPayload {
   final Uri realm;
   final String email;
-  final int? userId; // TODO(server-5) new in FL 108
+  final int userId;
```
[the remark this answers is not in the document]
gnprice: bump / nit: use `api:` for prefix, like for other API changes

(in this revision it's:

    login: Mark WebAuthPayload.userId as required, relying on server 5+, FL 108+.

)

lib/api/model/events.dart
```diff
@@ -704,13 +704,13 @@ class UpdateMessageEvent extends Event {
...
-  final int? userId; // TODO(server-5)
-  final bool? renderingOnly; // TODO(server-5)
+  final int userId;
+  final bool renderingOnly;
   final int messageId;
   final List<int> messageIds;
 
   final List<MessageFlag> flags;
-  final int? editTimestamp; // TODO(server-5)
+  final int editTimestamp;
```
gnprice, line 713: nit: this doesn't seem accurate:

    Fields including `userId` and `editTimestamp` are optional and act
    as alternatives to `renderingOnly` for older server versions.

since those fields are no longer optional.

I think the commit message can skip explaining the fallback that previously existed for renderingOnly. It's directly relevant only when looking at the past; and the details are in this commit's diff.

lib/api/model/initial_snapshot.dart
```diff
@@ -275,15 +269,9 @@ class UnreadMessagesSnapshot {
 /// An item in [UnreadMessagesSnapshot.dms].
 @JsonSerializable(fieldRename: FieldRename.snake)
 class UnreadDmSnapshot {
-  @JsonKey(readValue: _readOtherUserId)
   final int otherUserId;
```
gnprice, line 272: nit:

    api: Remove sender_id fallback for UnreadMessagesSnapshot.otherUserId, relying on server 5+, FL 119+

that's not the class this field is on :slightly_smiling_face:

gnprice (commented):
Thanks for the revisions! Just a few comments below.

gnprice:
Also marked as draft to reflect that we won't be ready to merge this until we have #267.

Marking as draft helps avoid me forgetting that and going ahead and merging after a next round of review, if the thread otherwise looks ready :slightly_smiling_face:

lib/api/route/realm.g.dart
PIG208: For future reference, this was fixed in #947.

lib/api/model/events.dart
PIG208: Yeah, "were" would be more accurate here. Leaving this out makes sense.

lib/api/model/initial_snapshot.dart
PIG208: Good find! I should learn to not trust the diffs heading and be more careful from now on :sweat: 

```diff
diff --git a/lib/api/model/initial_snapshot.dart b/lib/api/model/initial_snapshot.dart
index 565d89fe..6be29e78 100644
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -269,15 +269,9 @@ class UnreadMessagesSnapshot {
 /// An item in [UnreadMessagesSnapshot.dms].
 @JsonSerializable(fieldRename: FieldRename.snake)
 class UnreadDmSnapshot {
-  @JsonKey(readValue: _readOtherUserId)
   final int otherUserId;
   final List<int> unreadMessageIds;
 
-  // TODO(server-5): Simplify away.
-  static dynamic _readOtherUserId(Map<dynamic, dynamic> json, String key) {
-    return json[key] ?? json['sender_id'];
-  }
-
```

gnprice: Ah, yeah, the diff-hunk headers can be misleading like that sometimes :slightly_smiling_face:

gnprice:
Removed the label to get this out of my queue since it's on hold (as described above). We'll bring it back when we pick up the underlying issue again.

gnprice:
We have #267 now, so this is something we could potentially merge when its code is ready. I've just rebased it atop current main.

gnprice:
The comments in my previous round https://github.com/zulip/zulip-flutter/pull/904#pullrequestreview-2307995616 above were all addressed.

Skimming now, there are a few commits whose summary lines indicate they're not ready:
9f15154d9 TODO api: Add TODO to modernize GetServerSettingsResult.zulipMergeBase
c514b59f6 FIXME LINK api/notif [nfc]: Cut comment on FcmMessageChannelRecipient.streamId, relying on server 5+, FL 115+.
662ccaeae TODO api: Mark GetServerSettingsResult.realmWebPublicAccessEnabled as required, relying on server 5+, FL 116+

Other than that, it seems like these changes are probably ready. I guess I'll want to reread, though, at least briefly, since it's been most of a year since we put it on hold (awaiting #267).

lib/api/route/realm.dart
```diff
@@ -36,7 +36,8 @@ class GetServerSettingsResult {
 
   final int zulipFeatureLevel;
   final String zulipVersion;
-  final String? zulipMergeBase; // TODO(server-5)
+  // TODO(server-5): Modernize this once we get to #267.
+  final String? zulipMergeBase;
```
chrisbobbe:
```
TODO api: Add TODO to modernize GetServerSettingsResult.zulipMergeBase

Marked as TODO as this might change in the future.
See https://github.com/zulip/zulip-flutter/pull/904#discussion_r1747943940.

See "Feature level 88" from Zulip API changelog:

   https://zulip.com/api/changelog

```

I read the linked discussion, and this change can go ahead; it doesn't interfere with what we've implemented for disallow-connecting-to-ancient-servers.

(Ditto the corresponding change in the initial snapshot, which would have been more likely to have a conflict but also doesn't.)

lib/api/model/events.dart
```diff
@@ -797,13 +797,13 @@ class UpdateMessageEvent extends Event {
   @JsonKey(includeToJson: true)
   String get type => 'update_message';
 
-  final int? userId; // TODO(server-5)
-  final bool? renderingOnly; // TODO(server-5)
+  final int userId;
```
chrisbobbe: Discussion on this `userId` change: [#api documentation > &#96;user_id&#96; in &#96;update_message&#96; event @ 💬](https://chat.zulip.org/#narrow/channel/412-api-documentation/topic/.60user_id.60.20in.20.60update_message.60.20event/near/2215174)

lib/api/route/realm.dart
```diff
@@ -51,7 +51,7 @@ class GetServerSettingsResult {
   final String realmName;
   final String realmIcon;
   final String realmDescription;
-  final bool? realmWebPublicAccessEnabled; // TODO(server-5)
+  final bool realmWebPublicAccessEnabled;
```
chrisbobbe, line 53:
```
TODO api: Mark GetServerSettingsResult.realmWebPublicAccessEnabled as required, relying on server 5+, FL 116+

See https://github.com/zulip/zulip-flutter/pull/904#discussion_r1747943940.

See "Feature level 116" from Zulip API changelog:

   https://zulip.com/api/changelog

```

as in a previous comment of mine: this is fine; no interaction with disallow-ancient-servers

chrisbobbe (commented):
Thanks @PIG208!

Here's a review, since this is marked for maintainer review—but probably Greg or I will end up making new revisions?

There are three `TODO(server-5)` comments left at the tip of the branch; I've commented on one in my review, and the other two are also small but snuck in after this PR was first written almost a year ago.

lib/api/model/events.dart
chrisbobbe: From that discussion (thanks @PIG208!) it sounds like we don't want to make `userId` required: we can expect the `user_id` key to always be _present_ in the JSON, but it'll be null for rendering-only events.

So I think the change here is just to remove the `TODO(server-5)`, in a separate commit that explains why we're not making it required.

PIG208:
Yeah. I think this mainly needs an update resolving the TODOs/new comments and a reread. I can get back to this tomorrow night (don't have my laptop with me today). Feel free to beat me to it if needed.

PIG208:
Pushed an update addressing the new TODOs and the comments above.

lib/api/model/model.dart
```diff
@@ -261,7 +261,7 @@ class User {
   // bool isOwner; // obsoleted by [role]; ignore
   // bool isAdmin; // obsoleted by [role]; ignore
   // bool isGuest; // obsoleted by [role]; ignore
-  bool? isBillingAdmin; // TODO(server-5)
+  bool? isBillingAdmin; // TODO(server-10)
```
gnprice: Since we didn't have any logic using this now-obsolete field, we don't have to wait before removing our code for it; we can go ahead and stop looking for the field now.

I'll add a commit to that effect.

lib/api/model/web_auth.dart
```diff
@@ -25,21 +25,16 @@ class WebAuthPayload {
...
       if (realm == null) throw const FormatException();
 
-      // TODO(server-5) require in queryParameters (new in FL 108)
-      final userIdStr = url.queryParameters['user_id'];
-      int? userId;
-      if (userIdStr != null) {
-        userId = int.tryParse(userIdStr, radix: 10);
-        if (userId == null) throw const FormatException();
-      }
+      final int? userId = int.tryParse(userIdStr, radix: 10);
```
gnprice, line 36: nit: omit boring types
```suggestion
      final userId = int.tryParse(userIdStr, radix: 10);
```

lib/api/model/model.dart
```diff
@@ -1165,18 +1159,15 @@ enum MessageEditState {
...
-        if (topic == null) {
-          hasMoved = true;
-        } else {
-          hasMoved |= !topicMoveWasResolveOrUnresolve(topic, prevTopic);
-        }
+      if (topic != null || prevTopic != null) {
+        // Crunchy-shell validation: Both are present if the topic was edited
+        topic as TopicName;
+        prevTopic as TopicName;
+        hasMoved |= !topicMoveWasResolveOrUnresolve(topic, prevTopic);
```
gnprice, line 1270: This can be simpler:
```dart
      final prevTopicStr = entry['prev_topic'] as String?;
      if (prevTopicStr != null) {
        final prevTopic = TopicName.fromJson(prevTopicStr);
        final topic = TopicName.fromJson(entry['topic'] as String);
        hasMoved |= !topicMoveWasResolveOrUnresolve(topic, prevTopic);
```

The one difference in behavior that way is that with this version, if a history entry has `topic` but not `prev_topic`, this code won't notice the mismatch; the other version would throw.

Which is fine. In general our API bindings don't aim to be a validator for correct server behavior; if the server includes something extraneous that they weren't looking for, they won't notice or complain.

The "crunchy shell" concept would come in if this layer of our code were going to pass this data on, to be consumed by other code deeper in the app (the "soft center"). When that's happening, it's important that the data that gets passed on be internally coherent. But here `topic` and `prevTopic` don't get passed on anywhere, except to this `topicMoveWasResolveOrUnresolve` call; and we don't make that call anyway unless they're both present.

lib/api/core.dart
```diff
@@ -14,7 +14,7 @@ import 'exception.dart';
 ///
 /// When updating this, also update [kMinSupportedZulipFeatureLevel]
 /// and the README.
-// TODO(#268) address all TODO(server-5), TODO(server-6), and TODO(server-7)
+// TODO(#992) address all TODO(server-6), and TODO(server-7)
```
gnprice: nit:
```suggestion
// TODO(#992) address all TODO(server-6) and TODO(server-7)
```

lib/api/route/messages.dart
```diff
@@ -1,69 +1,18 @@
...
```
gnprice, line 16: nit in commit message: use 9 hex digits for commit IDs (https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#mentioning-commits)

    See also commit 631f4d68 .

lib/api/model/web_auth.dart
gnprice: … I guess the existing code just above for Uri.tryParse does the same thing, so this matches. Whatever, then; we can clean up them both if we're touching this code in the future.

gnprice:
OK, revised and updated. @chrisbobbe would you take a look at the revision?

lib/api/route/realm.dart
```diff
@@ -36,7 +36,7 @@ class GetServerSettingsResult {
 
   final int zulipFeatureLevel;
   final String zulipVersion;
-  final String? zulipMergeBase; // TODO(server-5)
+  final String zulipMergeBase;
```
chrisbobbe, line 39: Indeed, I've thought more about this, and while we don't rely on zulipMergeBase to decide if a server is ancient, it would break things to require it without also checking to see if the server is too old. Sent #1783 for that, which we should merge before coming back to this.

chrisbobbe:
```
api: Mark GetServerSettingsResult.zulipMergeBase as required

See "Feature level 88" from Zulip API changelog:

   https://zulip.com/api/changelog

This does not interfere with our handling of
disallow-connecting-to-ancient-servers, which does not rely on
GetServerSettingsResult.
```

In this commit message, let's just remove that last paragraph, or reword to explain that we do use `GetServerSettingsResult`, but it's OK because of #1783.

lib/api/model/initial_snapshot.dart
```diff
@@ -18,7 +18,7 @@ class InitialSnapshot {
   final int lastEventId;
   final int zulipFeatureLevel;
   final String zulipVersion;
-  final String? zulipMergeBase; // TODO(server-5)
+  final String zulipMergeBase;
```
chrisbobbe, line 21:
```
api: Mark InitialSnapshot.zulipMergeBase as required, relying on server 5+, FL 88+

See "Feature level 88" from Zulip API changelog:

   https://zulip.com/api/changelog

This does not interfere with our handling of
disallow-connecting-to-ancient-servers, which does not rely on
InitialSnapshot.
```

Similarly remove misleading last paragraph at the end; disallow-connecting-to-ancient-servers does rely on `InitialSnapshot`, in the sense that that's one of the places we learn if the server is ancient or not.

gnprice:
Thanks for the review (and for #1783)!

Made those commit-message edits and rebased. I also added a related comment in a small new commit:
506b87a74 server_support [nfc]: Explain why zulipMergeBase should remain nullable

I'll let CI run on this, and then I plan to merge.

PR #873, opened 2024-08-08
deps: Upgrade packages within constraints (tools/upgrade pub), mostly

gnprice:
Following up on https://github.com/zulip/zulip-flutter/pull/872#issuecomment-2276823629 .

This fixes #825 because of the updates to file_picker 8.0.7 and
flutter_local_notifications_linux 4.0.1.

Holding back the Drift-related packages (drift, drift_dev, sqlparser)
just because there appear to be some changes there that require small
updates on our side, and today I'm catching up on things and so don't
have time to dig into those myself.

There are also some updates that `tools/upgrade pub-major` would
make beyond these, which I'm similarly leaving aside for the moment.

Fixes: #825

PR #907, opened 2024-08-22
gitignore: Add a few lines following an upstream change

chrisbobbe:
This change has been getting applied by `flutter run`, and Greg identified this upstream commit as the cause:
  https://github.com/flutter/flutter/commit/29f332c82

Discussion:
  https://github.com/zulip/zulip-flutter/pull/873#issuecomment-2285235234

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,11 @@
 *.swp
 .DS_Store
 .atom/
+.build/
 .buildlog/
 .history
 .svn/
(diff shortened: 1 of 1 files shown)


gnprice:
LGTM, thanks — please go ahead and merge.

Based on that upstream commit, I expect this change gets applied any time you build or run the app for iOS or macOS.

PR #908, opened 2024-08-22
autocomplete [nfc]: Cut disused CandidateT type parameters

gnprice:
This is a small followup to #896 — after making these type parameters no longer needed, I forgot to also go back and remove them.

(/cc @sm-sayedi as you'll want to rebase #889 atop this)


chrisbobbe:
Thanks, LGTM! Merging.

Issue #810, opened 2024-07-13
Unhandled exception in dispose of message store


gnprice:
I saw the following exception on my device just now:
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Concurrent modification during iteration: _Set len:0.
#0      _CompactIterator.moveNext (dart:collection-patch/compact_hash.dart:714)
#1      MessageStoreImpl.dispose (package:zulip/model/message.dart:57)
#2      PerAccountStore.dispose (package:zulip/model/store.dart:385)
#3      GlobalStore._setPerAccount (package:zulip/model/store.dart:135)
#4      GlobalStore._reloadPerAccount (package:zulip/model/store.dart:128)
<asynchronous suspension>
#5      UpdateMachine.poll (package:zulip/model/store.dart:733)
<asynchronous suspension>
```

The context was:
* I'd had the app open previously, and was looking at a message list
* Then I connected the device to my computer, and was looking at the logcat view in Android Studio
* Then I opened up the app again, from the background
* The error appeared a few seconds later.

From the stack trace, it looks like what happened is that the polling loop found the event queue had expired; it went to replace the store with a new one from a new initial fetch; and then it hit this exception while trying to dispose of the old store.

---

Looking at the call stack, I'm not sure there are user-visible symptoms from this error. It looks like there are only a few more steps that hadn't yet been taken, all of them various `dispose` calls.

On the other hand earlier this evening, I had observed a puzzling failure to replace the event queue:
* #809

So if this could cause a symptom like that after all, then this could be the diagnosis for that issue #809. (That issue is why I was looking at logcat in the first place. But no luck; it'd been a couple of hours since the issue had happened, and the logs from that time were long since gone.)

In any case, we should fix the issue so that `MessageStoreImpl.dispose` completes successfully.

---

A PR fixing this issue should include two kinds of tests:
* A test focused on that `dispose` method: make a message store, register a message list or two on it, then call `dispose`. At a minimum this test can verify no exception is thrown; preferably it should also verify that `dispose` did its job.
* A direct regression test for this issue, which will be a bit more of an integration test: make a PerAccountStore and UpdateMachine; register a message list or two on the store; then induce an expired-queue exception in the polling loop so that it replaces the store, and expect no unhandled exceptions. (I believe the test framework will automatically fail the test if there's an unhandled exception.)

  For this test, take as a model the existing tests of the expired-queue recovery feature.

[...]

PR #909, opened 2024-08-22
Handle disposal of message store correctly.

PIG208:
Fixes: #810

### DIFF
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -10,6 +10,8 @@ mixin MessageStore {
   /// All known messages, indexed by [Message.id].
   Map<int, Message> get messages;
 
+  Set<MessageListView> get debugMessageListViews;
+
   void registerMessageList(MessageListView view);
   void unregisterMessageList(MessageListView view);
 
@@ -38,6 +40,9 @@ class MessageStoreImpl with MessageStore {
   final Set<MessageListView> _messageListViews = {};
 
   @override
+  Set<MessageListView> get debugMessageListViews => _messageListViews;
+
+  @override
   void registerMessageList(MessageListView view) {
     final added = _messageListViews.add(view);
     assert(added);
@@ -56,7 +61,10 @@ class MessageStoreImpl with MessageStore {
   }
 
   void dispose() {
-    for (final view in _messageListViews) {
+    // When a MessageListView is disposed, it removes itself from the Set
+    // `MessageStoreImpl._messageListViews`. Instead of iterating on that Set,
+    // iterate on a copy, to avoid concurrent modifications.
+    for (final view in _messageListViews.toList()) {
       view.dispose();
     }
   }
--- a/test/model/message_test.dart
+++ b/test/model/message_test.dart
@@ -77,6 +77,18 @@ void main() {
     checkNotified(count: messageList.fetched ? messages.length : 0);
   }
 
+  test('disposing multiple registered MessageListView instances', () async {
+    // Regression test for: https://github.com/zulip/zulip-flutter/issues/810
+    await prepare(narrow: const MentionsNarrow());
+    MessageListView.init(store: store, narrow: const StarredMessagesNarrow());
+    check(store.debugMessageListViews).length.equals(2);
+
+    // When disposing, the [MessageListView]s are expected to unregister
+    // themselves from the message store.
+    store.dispose();
+    check(store.debugMessageListViews).isEmpty();
+  });
+
   group('reconcileMessages', () {
     test('from empty', () async {
       await prepare();
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -8,6 +8,8 @@ import 'package:zulip/api/model/events.dart';
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/route/events.dart';
 import 'package:zulip/api/route/messages.dart';
+import 'package:zulip/model/message_list.dart';
+import 'package:zulip/model/narrow.dart';
 import 'package:zulip/model/store.dart';
 import 'package:zulip/notifications/receive.dart';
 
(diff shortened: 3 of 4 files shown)


PIG208:
The CI failures reproduce the error mentioned in the issue: https://github.com/zulip/zulip-flutter/actions/runs/10517036561/job/29140404556.

PIG208:
Pushed again to resolve merge conflicts.

chrisbobbe:
Thanks!

```
[…]

This fixes a bug caused by MessageListView instances unregistering themselves
from the collection containing these instances, while we are iterating the exact
same set when disposing the message store.

[…]
```

For the part before the comma, `MessageListView.dispose` is in that call stack, right? Could you point to what's calling that? I don't think I understand yet what all the concurrent modifications are.

PIG208:
> For the part before the comma, MessageListView.dispose is in that call stack, right?

Yes, this is called from `PerAccountStore.dispose`, which recursively disposes the message store, and the message store disposes the registered message list views.

chrisbobbe:
> `PerAccountStore.dispose`, which recursively disposes the message store

I don't understand this yet. Since you said "recursively", I'm looking for a function that, when called, causes itself to be called. Is that what you have in mind, and could you point more specifically to which function this is and where the recursive call is?

I see that the message store disposes the registered message list views, but that shouldn't cause a concurrent modification error by itself. I'm looking for what else is trying to modify the collection while that's happening, and I haven't found it yet.

PIG208:
I guess the correct word would be propagatively. The for loop modified in the fixup commit iterates over the registered message lists and calls dispose on them, those message lists then unregistered themselves to cause modification concurrent to the for loop.

chrisbobbe:
Ohh, I see, yeah. Part of the work of disposing a `MessageListView` is to remove it from a collection of `MessageListView`s, and it's a problem if we're iterating over that same collection in order to dispose all its members one after another.

chrisbobbe:
One commit-message clarification might be to say "removing" instead of "unregistering" in "MessageListView instances unregistering themselves from the collection". It's true that there's code involved with the name `unregisterMessageList`, but "remove" is the canonical term for the operation on a collection, and that operation is contributing to the concurrent modification bug.

chrisbobbe:
How about, to replace the paragraph I quoted in https://github.com/zulip/zulip-flutter/pull/909#issuecomment-2311480310 :

> Part of the work of disposing a `MessageListView` is to remove it from a collection of `MessageListView`s, and we've been doing this in a loop over the exact same collection (the set `MessageStoreImpl._messageListViews`), and this has been causing concurrent modification errors. This fix avoids those errors by having the loop iterate over a clone of that collection instead.

PIG208:
Updated the description and tweaked the comment. Thanks for the feedback!

lib/model/message.dart
```diff
@@ -56,7 +61,11 @@ class MessageStoreImpl with MessageStore {
   }
 
   void dispose() {
-    for (final view in _messageListViews) {
+    // When a MessageListView is disposed, it removes itself from the set
+    // `MessageStoreImpl._messageListViews`, which is the exact collection this
+    // iterates on. We need to make a copy of the collection to avoid concurrent
+    // modifications.
+    for (final view in _messageListViews.toList()) {
```
chrisbobbe, line 68: nit: Well, it isn't the exact collection this iterates on :) because we're avoiding doing that, for the reason this mentions. How about:

```suggestion
    // When a MessageListView is disposed, it removes itself from the Set
    // `MessageStoreImpl._messageListViews`. Instead of iterating on that Set,
    // iterate on a copy, to avoid concurrent modifications.
```

chrisbobbe (commented):
Thanks! LGTM except for one nit below.

PIG208:
Updated the PR. Thanks!

chrisbobbe:
Thanks for the revision! In the changed comment, let's be consistent in saying "set" or "Set", and there's a comma that should be a period.

test/model/store_test.dart
```diff
@@ -427,6 +429,36 @@ void main() {
...
+      updateMachine.debugAdvanceLoop();
+      async.flushMicrotasks();
+      await Future<void>.delayed(Duration.zero);
+      check(store).isLoading.isTrue();
+      check(store.debugMessageStore.debugMessageLists).isEmpty();
+
+      // The global store has a new store.
+      check(globalStore.perAccountSync(store.accountId)).not((it) => it.identicalTo(store));
+      updateFromGlobalStore();
+      check(store).isLoading.isFalse();
```
gnprice, line 459: This can be more focused on the aspect this particular test is about:
```suggestion
      await Future<void>.delayed(Duration.zero);

      // The old store's [MessageListView]s have been disposed.
      // (And no exception was thrown; that was #810.)
      check(store.debugMessageStore.debugMessageListViews).isEmpty();
```

The other checks are effectively covered by the preceding test case.

Meanwhile the explicit comment about no exception is helpful because in this test case that's the main point of the test. Every test case implicitly checks that no miscellaneous exceptions were thrown from the code under test; but usually that's not the main thing the test is about, so when it's not mentioned the reader might overlook it.

test/model/message_test.dart
```diff
@@ -919,4 +919,23 @@ void main() {
...
+    await prepare(narrow: const MentionsNarrow());
+    MessageListView.init(store: store, narrow: const StarredMessagesNarrow());
+
+    Condition<Object?> conditionMessageListView(Narrow narrow) =>
+      (it) => it.isA<MessageListView>().narrow.equals(narrow);
+
+    check(store.debugMessageStore.debugMessageLists).deepEquals([
+      conditionMessageListView(const MentionsNarrow()),
+      conditionMessageListView(const StarredMessagesNarrow()),
+    ]);
```
gnprice, line 934: My first thought here is that this is probably clearer if `conditionMessageListView` just gets inlined at its two call sites. Less abstraction to work through, and not really any longer.

But I think this can actually go a step further and just check the length of the list of [MessageListView]s, as you do in the other test. The fact that it's these two particular narrows is sort of inevitable given the setup of the test, and it's not what this test is intended to check on.

(Also conceivably it could iterate in the other order. The current implementation doesn't do that, because the set comes from a set literal `{}` and that's a `LinkedHashSet`, which preserves the insertion order; but that's not a guarantee that `MessageStoreImpl` otherwise has any reason to make, and if it didn't then this test would sometimes break.)

lib/model/message.dart
```diff
@@ -56,7 +61,10 @@ class MessageStoreImpl with MessageStore {
   }
 
   void dispose() {
-    for (final view in _messageListViews) {
+    // When a MessageListView is disposed, it removes itself from the set
+    // `MessageStoreImpl._messageListViews`. Instead of iterating on that set,
+    // iterate on a copy, to avoid concurrent modifications.
+    for (final view in _messageListViews.toList()) {
```
gnprice, line 67: nit in commit message:

    This fixes a bug caused by MessageListView instances unregistering themselves
    from the collection containing these instances, while we are iterating the exact
    same set when disposing the message store.
    
lines too long for prose; from https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#formatting-guidelines :
> Your commit message should be line-wrapped to about 68 characters per line, but no more than 70, so that your commit message will be easy to read in `git log` in a normal terminal.

These lines are as long as 80 columns.

See also the tip there about configuring Git and your editor.

gnprice (commented):
Thanks @PIG208, and thanks @chrisbobbe for the reviews!

Generally this looks good. Small comments below.

PIG208:
Updated the PR, thanks for the review! Also removed the `visibleForTesting` annotation in this revision.

test/model/message_test.dart
```diff
@@ -919,4 +919,23 @@ void main() {
...
+
+  test('disposing multiple registered MessageListView instances', () async {
+    // Regression test for: https://github.com/zulip/zulip-flutter/issues/810
+    await prepare(narrow: const MentionsNarrow());
+    MessageListView.init(store: store, narrow: const StarredMessagesNarrow());
+
+    Condition<Object?> conditionMessageListView(Narrow narrow) =>
+      (it) => it.isA<MessageListView>().narrow.equals(narrow);
+
+    check(store.debugMessageStore.debugMessageLists).deepEquals([
```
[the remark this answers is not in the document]
PIG208: That's pretty neat. Thanks!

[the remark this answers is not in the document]
gnprice: (This also caused me to notice that I'd introduced a deviation from that pattern over on ChannelStore; sent a fix for that as #931, along with adding some docs.)

PIG208:
Updated test comments and the commit message.

gnprice:
Thanks for all the revisions! Looks good — merging.

PR #910, opened 2024-08-23
deps: Bump json_annotation range to resolve a warning in build_runner check

chrisbobbe:
Just to get the output of `tools/check build_runner` clear of warnings, so future warnings stand out.

Discussion:
  https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/build_runner.20unexpected.20output/near/1929816


gnprice:
Thanks! And indeed I see the output is now free of these lines:
```
Warning:  json_serializable on lib/api/notifications.dart:
The version constraint "^4.8.1" on json_annotation allows versions before 4.9.0 which is not allowed.
```

Looks good; merging.

PR #911, opened 2024-08-23
i18n: Consistently use … in favor of ...

PIG208:
We generally prefer … elsewhere in our codebase.

Motivated by https://github.com/zulip/zulip-flutter/pull/868#discussion_r1728101455

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -239,7 +239,7 @@
   "@composeBoxTopicHintText":  {
     "description": "Hint text for topic input widget in compose box."
   },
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Thanks, LGTM! Over to Greg when he's back from vacation.

gnprice:
Thanks for the cleanup. Looks good — merging, with some clarifications to the commit message:

    -    i18n: Consistently use … in favor of ...
    +    i18n: Consistently use "…" rather than "..."
     
    -    We generally prefer … elsewhere in our codebase.
    +    We generally prefer the ellipsis character … in user-facing strings.

PR #912, opened 2024-08-23
poll: Support read-only poll widget UI.

### LINKED ISSUE #165: polls: Read-only support

PIG208:
The UI follows the webapp until we get a new design.

The dark theme colors were tentatively picked. The `TextStyle`s are the same for both light and dark theme. All the styling are based on values taken from the webapp.

References:

  - light theme:

    https://github.com/zulip/zulip/blob/2011e0df760cea52c31914e7b77d9b4e38e9ee74/web/styles/widgets.css#L138-L185 https://github.com/zulip/zulip/blob/2011e0df760cea52c31914e7b77d9b4e38e9ee74/web/styles/dark_theme.css#L358

  - dark theme:

    https://github.com/zulip/zulip/blob/2011e0df760cea52c31914e7b77d9b4e38e9ee74/web/styles/dark_theme.css#L966-L987

Fixes #165.

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -4,6 +4,7 @@ import 'package:html/dom.dart' as dom;
 import 'package:html/parser.dart';
 
 import '../api/model/model.dart';
+import '../api/model/submessage.dart';
 import 'code_block.dart';
 
 /// A node in a parse tree for Zulip message-style content.
@@ -75,6 +76,16 @@ mixin UnimplementedNode on ContentNode {
 
 /// A parsed, ready-to-render representation of Zulip message content.
 sealed class ZulipMessageContent {}
+
+/// A wrapper around a mutable representation of a Zulip poll message.
+///
+/// Consumers are expected to listen for [Poll]'s changes to receive
+/// live-updates.
+class PollContent implements ZulipMessageContent {
+  const PollContent(this.poll);
+
+  final Poll poll;
+}
 
 /// A complete parse tree for a Zulip message's content,
 /// or other complete piece of Zulip HTML content.
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -73,13 +73,16 @@ mixin UnimplementedNode on ContentNode {
   }
 }
 
+/// A parsed, ready-to-render representation of Zulip message content.
+sealed class ZulipMessageContent {}
+
 /// A complete parse tree for a Zulip message's content,
 /// or other complete piece of Zulip HTML content.
 ///
 /// This is a parsed representation for an entire value of [Message.content],
 /// [Stream.renderedDescription], or other text from a Zulip server that comes
 /// in the same Zulip HTML format.
-class ZulipContent extends ContentNode {
+class ZulipContent extends ContentNode implements ZulipMessageContent {
   const ZulipContent({super.debugHtmlNode, required this.nodes});
 
   final List<BlockContentNode> nodes;
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -18,6 +18,7 @@ import 'dialog.dart';
 import 'icons.dart';
 import 'lightbox.dart';
 import 'message_list.dart';
+import 'poll.dart';
 import 'store.dart';
 import 'text.dart';
 
@@ -263,6 +264,7 @@ class MessageContent extends StatelessWidget {
         style: ContentTheme.of(context).textStylePlainParagraph,
         child: switch (content) {
           ZulipContent() => BlockContentList(nodes: content.nodes),
+          PollContent()  => PollWidget(poll: content.poll),
         }));
   }
 }
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -41,6 +41,10 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
       colorGlobalTimeBorder: const HSLColor.fromAHSL(1, 0, 0, 0.8).toColor(),
       colorMathBlockBorder: const HSLColor.fromAHSL(0.15, 240, 0.8, 0.5).toColor(),
       colorMessageMediaContainerBackground: const Color.fromRGBO(0, 0, 0, 0.03),
+      colorPollNames: const HSLColor.fromAHSL(1, 0, 0, .45).toColor(),
+      colorPollVoteCountBackground: const HSLColor.fromAHSL(1, 0, 0, 1).toColor(),
+      colorPollVoteCountBorder: const HSLColor.fromAHSL(1, 156, 0.28, 0.7).toColor(),
+      colorPollVoteCountText: const HSLColor.fromAHSL(1, 156, 0.41, 0.4).toColor(),
       colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor(),
       textStylePlainParagraph: _plainParagraphCommon(context).copyWith(
         color: const HSLColor.fromAHSL(1, 0, 0, 0.15).toColor(),
@@ -66,6 +70,10 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
       colorGlobalTimeBorder: const HSLColor.fromAHSL(0.4, 0, 0, 0).toColor(),
       colorMathBlockBorder: const HSLColor.fromAHSL(1, 240, 0.4, 0.4).toColor(),
       colorMessageMediaContainerBackground: const HSLColor.fromAHSL(0.03, 0, 0, 1).toColor(),
+      colorPollNames: const HSLColor.fromAHSL(1, 236, .15, .7).toColor(),
+      colorPollVoteCountBackground: const HSLColor.fromAHSL(0.2, 0, 0, 0).toColor(),
+      colorPollVoteCountBorder: const HSLColor.fromAHSL(1, 185, 0.35, 0.35).toColor(),
+      colorPollVoteCountText: const HSLColor.fromAHSL(1, 185, 0.35, 0.65).toColor(),
       colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor().withValues(alpha: 0.2),
       textStylePlainParagraph: _plainParagraphCommon(context).copyWith(
         color: const HSLColor.fromAHSL(1, 0, 0, 0.85).toColor(),
@@ -90,6 +98,10 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
     required this.colorGlobalTimeBorder,
     required this.colorMathBlockBorder,
     required this.colorMessageMediaContainerBackground,
+    required this.colorPollNames,
+    required this.colorPollVoteCountBackground,
+    required this.colorPollVoteCountBorder,
+    required this.colorPollVoteCountText,
     required this.colorThematicBreak,
     required this.textStylePlainParagraph,
     required this.codeBlockTextStyles,
@@ -115,6 +127,10 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
   final Color colorGlobalTimeBorder;
   final Color colorMathBlockBorder; // TODO(#46) this won't be needed
   final Color colorMessageMediaContainerBackground;
+  final Color colorPollNames;
+  final Color colorPollVoteCountBackground;
+  final Color colorPollVoteCountBorder;
+  final Color colorPollVoteCountText;
   final Color colorThematicBreak;
 
   /// The complete [TextStyle] we use for plain, unstyled paragraphs.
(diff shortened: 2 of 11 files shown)


PIG208:
| light | dark|
| - | - |
| Screenshot from 2024-08-22 22-30-32 | Screenshot from 2024-08-22 22-31-01 |

lib/widgets/poll.dart
```diff
@@ -0,0 +1,77 @@
...
+              Expanded(
+                child: Padding(
+                  padding: const EdgeInsetsDirectional.only(start: 5.0),
+                  child: Wrap(children: [
+                    Padding(
+                      padding: const EdgeInsetsDirectional.only(end: 5.0),
+                      child: Text(option.text,
+                        style: textStylePollBold.copyWith(fontSize: 16))),
+                    if (option.voters.isNotEmpty)
+                      Text('(${getOptionVoterNames(option)})', style: textStylePollNames),
```
chrisbobbe, line 57: I think this might need a TODO for localizing the parenthesis characters (like to use full-width parens in Chinese).

lib/widgets/poll.dart
```diff
@@ -0,0 +1,77 @@
...
+    final theme = ContentTheme.of(context);
+
+    final textStylePollBold = const TextStyle(fontSize: 18)
+      .merge(weightVariableTextStyle(context, wght: 600));
+    final textStylePollNames = TextStyle(
+      fontSize: 16, color: theme.colorPollNames);
+
+    String getOptionVoterNames(PollOption option) => option.voters.map((userId) =>
+      store.users[userId]?.fullName
+        ?? zulipLocalizations.unknownUserName).join(', ');
```
chrisbobbe, line 27: Let's add a `TODO(i18n)` like this:

```dart
        // TODO(i18n): List formatting, like you can do in JavaScript:
        //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Zixuan'])
        //   // 'Chris、Greg、Alya、Zixuan'
```

(we have this TODO in a few places already)

lib/widgets/poll.dart
```diff
@@ -0,0 +1,77 @@
...
+                    if (option.voters.isNotEmpty)
+                      Text('(${getOptionVoterNames(option)})', style: textStylePollNames),
+                  ]))),
+            ])),
+    ];
+
+    Text question = (poll.question.isNotEmpty)
+      ? Text(poll.question, style: textStylePollBold)
+      : Text(zulipLocalizations.pollWidgetQuestionMissing,
+        style: textStylePollBold.copyWith(fontStyle: FontStyle.italic));
```
chrisbobbe, line 65:
```suggestion
      : Text(zulipLocalizations.pollWidgetQuestionMissing,
          style: textStylePollBold.copyWith(fontStyle: FontStyle.italic));
```

chrisbobbe (commented):
Thanks! Small comments below.

PIG208:
Thanks! Updated the PR.

chrisbobbe:
Thanks! LGTM except bump on https://github.com/zulip/zulip-flutter/pull/912#discussion_r1737264933 , and I'll mark this for Greg's review.

lib/model/content.dart
```diff
@@ -74,6 +75,12 @@ mixin UnimplementedNode on ContentNode {
 
 sealed class ZulipMessageContent {}
 
+class PollContent implements ZulipMessageContent {
```
PIG208, line 78: While this is pretty short, I'm not sure if we need a better way to organize `ZulipMessageContent`, `PollContent`, and `MessageContent`. The file `lib/model/content.dart` is predominantly about our AST for Zulip HTML content.

PIG208:
Rebased the PR.

PIG208:
Pushed some simplifications to `test/widgets/poll_test.dart`.

PIG208:
Update: use `spacing` when applicable and fix an issue with large vote numbers:

| before | after |
| - | - |
| Screenshot from 2024-09-13 22-28-26 | Screenshot from 2024-09-13 22-28-56 |

lib/widgets/poll.dart
```diff
@@ -0,0 +1,85 @@
...
+      fontSize: 16, color: theme.colorPollNames);
+
+    String getOptionVoterNames(PollOption option) {
+      // TODO(i18n): List formatting, like you can do in JavaScript:
+      //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Zixuan'])
+      //   // 'Chris、Greg、Alya、Zixuan'
+      return option.voters
+        .map((userId) =>
+          store.users[userId]?.fullName ?? zulipLocalizations.unknownUserName)
+        .join(', ');
```
PIG208, line 32: Indented this in a style taken from `ListNode.debugDescribeChildren`, to avoid going over 80 characters.
```dart
  @override
  List<DiagnosticsNode> debugDescribeChildren() {
    return items
      .mapIndexed((i, nodes) =>
        _BlockContentListNode(nodes).toDiagnosticsNode(name: 'item $i'))
      .toList();
  }
```

PIG208:
Update: resolve a merge conflict and make minor formatting change.

gnprice (commented):

Comments below.

Also one higher-level comment: the PollWidget depends on the mutable data in Poll, which isn't how Flutter expects things to work: see docs on [StatelessWidget.build]. Instead let's make Poll into a ChangeNotifier, and PollWidget into a stateful widget that listens to it.

That will also let us skip notifying the entire MessageListView's listeners when a poll gets updated, which will be good for performance in some scenarios: if you're looking at a thread with a poll that's actively getting votes, it'll let us avoid rebuilding all the other messages every time there's a vote on the poll.

(I think this version ends up behaving correctly because the PollWidget gets rebuilt as a descendant of the MessageList, but best to align with the architecture anyway. That aligning with the architecture is good for performance is a bonus.)

lib/widgets/content.dart
```diff
@@ -222,17 +251,20 @@ const double kBaseFontSize = 17;
 /// This does not include metadata like the sender's name and avatar, the time,
 /// or the message's status as starred or edited.
 class MessageContent extends StatelessWidget {
-  const MessageContent({super.key, required this.message, required this.content});
+  const MessageContent({super.key, required this.message, required ZulipMessageContent content}) : _content = content;
```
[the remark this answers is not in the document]
PIG208: This is for type promotion, similar to [this other instance](https://github.com/zulip/zulip-flutter/pull/808#issue-2406389310).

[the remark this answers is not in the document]
gnprice: > narrow is refactored to become private because otherwise `switch` is not able to promote the type of it.

I see. In that case:
https://github.com/zulip/zulip-flutter/pull/808#discussion_r1677245004

:slightly_smiling_face: 

In general, there's a local fix for that problem at any place where it arises, whereas changing a field from public to private changes the class's API for everywhere. So it's better to not let this affect whether the field is public or private.

PIG208:
Updated the PR. Now we are using `localizedTextBaseline` to align the text. This makes a slight difference:

| before | after |
| - | - |
| Screenshot from 2024-09-18 21-46-28 | Screenshot from 2024-09-18 21-45-32 |

Made `Poll` a `ChangeNotifier`, and let `Poll` listen to `Poll` instead. The second to last commit summary is intended to be discarded as the final commit squashes into it.

test/model/message_list_test.dart
```diff
@@ -544,65 +509,76 @@ void main() {
...
+      check(model).messages.single.poll.isNull();
+      check(model).contents.single.isA<ZulipContent>();
     });
 
-    test('some messages present', () async {
-      await prepare(narrow: const CombinedFeedNarrow());
-      await prepareMessages(foundOldest: false,
-        messages: messages.where((m) => m.id != 151).toList());
-      model.notifyListenersIfAnyMessagePresent([150, 151, 152]);
+    test('parse PollContent', () async {
```
gnprice, line 564: These two new cases don't belong in a group that's about `handleMessageEvent`, because they're not really about that method — in particular they're not any more specific to that method than they are to `fetchInitial` or `fetchOlder`.

In general this test file is one where our usual principle of organizing the tests in the same order as the code they test often breaks down: `model/message_list.dart` is one of our most intricate areas of model code, and might be the one file with the greatest complexity of interrelationships between different parts of its implementation.

You can see that organizing principle already broken down in roughly the latter half of the file: the "regression tests for …" and "stream/topic muting" groups, the tests "showSender is maintained correctly" and "recipient headers are maintained consistently", and the `checkInvariants` function which underlies most of the tests and describes invariants that cut across the implementation code.

So these can go in a new small group of their own, perhaps just above "recipient headers are maintained consistently" since they're both about logic within _MessageSequence.

Then I think we don't need these commits:
1919c1ba2 msglist test [nfc]: Make a group for MessageEvent tests
99a62b71a msglist test [nfc]: Name test groups after the event handler method
b69f0bfef msglist test [nfc]: Move notifyListeners tests to match implementation

test/model/message_list_test.dart
```diff
@@ -1780,32 +1824,6 @@ void checkInvariants(MessageListView model) {
...
-}
-
-extension MessageListMessageItemChecks on Subject<MessageListMessageItem> {
-  Subject<Message> get message => has((x) => x.message, 'message');
-  Subject<ZulipContent> get content => has((x) => x.content, 'content');
-  Subject<bool> get showSender => has((x) => x.showSender, 'showSender');
-  Subject<bool> get isLastInBlock => has((x) => x.isLastInBlock, 'isLastInBlock');
-}
-
-extension MessageListViewChecks on Subject<MessageListView> {
```
gnprice, line 1798: So, the downside of moving these checks-extensions to their own file is that its name `message_list_checks.dart` starts the same way as the test file `message_list_test.dart`. That makes it get in the way of tab-completion, when writing the test file's name for a `git log` or `flutter test` command.

For that reason I've been keeping the extensions within the test file when they're pretty specific to the one file, or to that and one or two others. I think these are best kept here for that reason.

gnprice (commented):
Thanks for the revision! Generally this looks good — a few comments below.

I haven't read closely all the details of the build method, but I'll skip doing so; this seems good enough.

Though it looks like @alya hasn't been consulted on this thread yet. Alya, any feedback on the UI in the screenshots above?

> The second to last commit summary is intended to be discarded as the final commit squashes into it.

I recommend making the commit messages into a draft of what you'd like to see merged. :slightly_smiling_face: This sort of remark would be easy for me to miss.

In general if there are details you want to say only at review time and don't think belong in the Git history, the PR thread (or a chat thread) is a good venue for those.

In that commit message I think the details are good for the Git history, though:

    The UI follows the webapp until we get a new design.
    
    The dark theme colors were tentatively picked. The `TextStyle`s are
    the same for both light and dark theme. All the styling are based on
    values taken from the webapp.
    
    References:
    
      - light theme:
    
        https://github.com/zulip/zulip/blob/2011e0df760cea52c31914e7b77d9b4e38e9ee74/web/styles/widgets.css#L138-L185
        https://github.com/zulip/zulip/blob/2011e0df760cea52c31914e7b77d9b4e38e9ee74/web/styles/dark_theme.css#L358
    
      - dark theme:
    
        https://github.com/zulip/zulip/blob/2011e0df760cea52c31914e7b77d9b4e38e9ee74/web/styles/dark_theme.css#L966-L987
    
They give good context for what we were thinking when we made the particular decisions that appear in that code. That's the sort of thing that's helpful to a later reader who's, for example, thinking about changing the code and wants to know what background they might be missing.

alya:
Looks reasonable to me! 

I remember making some comments about polls earlier (which I believe have been addressed), but I guess that was on a different PR.

PIG208:
The second to last commit is to be kept now, because previous those two were in the reverse order because of the way the tests were organized. My (stale) comment on that did make it more confusing!

Pushed an update addressing the review comments.

gnprice:
Thanks for the revision! This all looks good — merging. Happy to get this feature in.

I took out one new paragraph the latest revision added to a commit message:

    -    _reparseContent currently happens when there is content edit, which is
    -    forbidden for poll messages, (and you can't turn an existing message
    -    into a poll).  In most cases the same `PollContent` instance will be
    -    used for each poll message, unlike `ZulipContent`, and a poll message
    -    never needs to get reparsed.

It's true that over the lifetime of a given MessageListView, in a release build, a given poll message will be represented by one PollContent instance. But it can get a new PollContent if `MessageListView.reassemble` is called, which happens routinely in development. And in all builds, if the same message appears in a new message list, it'll get a new PollContent instance there. So there's no guarantee of uniqueness.

From a performance perspective, the main point here is that allocating one of these trivial `PollContent` wrapper objects is a much cheaper operation than parsing a message's HTML, which is what we do for normal messages in all the same situations where we make a `PollContent` for polls.

Issue #763, opened 2024-06-25
video lightbox: Screen doesn't stay on while video is playing


rajveermalviya:
https://github.com/zulip/zulip-flutter/assets/36819268/b4ce73d6-ab78-4f22-bf59-f813ec1ac07f

We can use [`package:wakelock_plus`](https://pub.dev/packages/wakelock_plus) which supports multiple platforms to implement this.

PR #916, opened 2024-08-23
 lightbox: Toggle wakelock while video is playing

rajveermalviya:
Fixes: #763

### DIFF
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -8,6 +8,7 @@ import 'package:video_player/video_player.dart';
 import '../api/core.dart';
 import '../api/model/model.dart';
 import '../log.dart';
+import '../model/binding.dart';
 import 'content.dart';
 import 'dialog.dart';
 import 'page.dart';
@@ -492,11 +493,24 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
     _controller?.removeListener(_handleVideoControllerUpdate);
     _controller?.dispose();
     _controller = null;
+    // The VideoController doesn't emit a pause event
+    // while disposing, so disable the wakelock here
+    // explicitly.
+    ZulipBinding.instance.toggleWakelock(enable: false);
     super.dispose();
   }
 
   void _handleVideoControllerUpdate() {
     setState(() {});
+    _updateWakelock();
+  }
+
+  Future<void> _updateWakelock() async {
+    if (_controller!.value.isPlaying) {
+      await ZulipBinding.instance.toggleWakelock(enable: true);
+    } else {
+      await ZulipBinding.instance.toggleWakelock(enable: false);
+    }
   }
 
   Widget? _buildBottomAppBar(BuildContext context, Color color, double elevation) {
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -145,6 +145,8 @@ PODS:
   - video_player_avfoundation (0.0.1):
     - Flutter
     - FlutterMacOS
+  - wakelock_plus (0.0.1):
+    - Flutter
 
 DEPENDENCIES:
   - app_settings (from `.symlinks/plugins/app_settings/ios`)
@@ -162,6 +164,7 @@ DEPENDENCIES:
   - sqlite3_flutter_libs (from `.symlinks/plugins/sqlite3_flutter_libs/ios`)
   - url_launcher_ios (from `.symlinks/plugins/url_launcher_ios/ios`)
   - video_player_avfoundation (from `.symlinks/plugins/video_player_avfoundation/darwin`)
+  - wakelock_plus (from `.symlinks/plugins/wakelock_plus/ios`)
 
 SPEC REPOS:
   trunk:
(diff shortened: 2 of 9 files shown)


Khader-1 (approved):
Thanks LGTM! And with manual testing I confirm that both iOS and Android versions does not sleep while a video is playing.

kenclary (approved):
also looks good to me

lib/widgets/lightbox.dart
```diff
@@ -492,11 +493,25 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
...
     _controller?.dispose();
     _controller = null;
+    WakelockPlus.disable();
     super.dispose();
   }
 
-  void _handleVideoControllerUpdate() {
-    setState(() {});
+  Future<void> _handleVideoControllerUpdate() async {
+    if (mounted) setState(() {});
```
chrisbobbe, line 501:
```
lightbox: Toggle wakelock while video is playing
```

The `if (mounted)` condition is new here. It looks reasonable, I think, but is it part of "Toggle wakelock while video is playing"? It looks orthogonal and like the kind of thing that belongs in a prep commit.

lib/widgets/lightbox.dart
```diff
@@ -492,11 +493,25 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
...
+      if (!wakelockEnabled) {
+        assert(debugLog('Wakelock enabled'));
+        await WakelockPlus.enable();
+      }
+    } else {
+      if (wakelockEnabled) {
+        assert(debugLog('Wakelock disabled'));
+        await WakelockPlus.disable();
+      }
+    }
```
chrisbobbe, line 514: It seems unfortunate to have an async gap here, with `await WakelockPlus.enabled;`. Then we have to think about bugs that could come from race conditions.

I see we're using the resolved value in order to avoid making redundant calls to `WakelockPlus.{enable,disable}`. But in the library's docs, it says that's not necessary:

https://pub.dev/documentation/wakelock_plus/latest/wakelock_plus/WakelockPlus-class.html
> You do not need to worry about making redundant calls, e.g. calling [WakelockPlus.enable](https://pub.dev/documentation/wakelock_plus/latest/wakelock_plus/WakelockPlus/enable.html) when the wakelock is already enabled as the plugin handles this for you, i.e. it checks the status to determine if the wakelock is already enabled or disabled.

So could the implementation really just be:

```dart
  Future<void> _handleVideoControllerUpdate() async {
    setState(() {});
    if (_controller == null) return;
    
    _controller!.value.isPlaying
      ? WakelockPlus.enable()
      : WakelockPlus.disable();
  }
```

?

lib/widgets/lightbox.dart
```diff
@@ -492,11 +493,25 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
...
     super.dispose();
   }
 
-  void _handleVideoControllerUpdate() {
-    setState(() {});
+  Future<void> _handleVideoControllerUpdate() async {
+    if (mounted) setState(() {});
+
+    final wakelockEnabled = await WakelockPlus.enabled;
+    if (_controller!.value.isPlaying) {
```
chrisbobbe, line 504: Could `_controller` be null here? I think maybe so, if the widget were disposed during the async gap. So we probably want an early return if `_controller == null`.

chrisbobbe (commented):
Thanks! Comments below.

Let's also add a TODO in the lightbox widget tests for testing the wakelock behavior, which will likely be done with help from `ZulipBinding`, by adding things for the wakelock package in the style of `ZulipBinding.packageInfo` etc.

rajveermalviya:
Thanks for the review @chrisbobbe, pushed a new revision that includes tests using a helper function via `ZulipBinding`.

lib/widgets/lightbox.dart
```diff
@@ -492,11 +493,25 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
...
   }
 
   void _handleVideoControllerUpdate() {
     setState(() {});
+    _updateWakelock();
+  }
+
+  Future<void> _updateWakelock() async {
+    if (_controller == null) return;
+    if ( _controller!.value.isPlaying) {
```
chrisbobbe, line 510: nit: extra space

chrisbobbe (approved):
Thanks! One nit below, and I'll mark for Greg's review.

lib/widgets/lightbox.dart
```diff
@@ -492,11 +493,25 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
     _controller?.removeListener(_handleVideoControllerUpdate);
     _controller?.dispose();
     _controller = null;
+    // The VideoController doesn't emit a pause event
+    // while disposing, so disable the wakelock here
+    // explicitly.
+    ZulipBinding.instance.toggleWakelock(enable: false);
```
gnprice, line 499: Good to have caught this — this API seems a bit fragile.

It would also get quite buggy if we ever had several widgets asking for this at the same time; for example, the first one that went away would turn it off even though the others were still around.

A good, fully Flutter-style API for this would probably involve a widget whose job is to keep the screen on for whatever window it appears in, and that behind the scenes deals with talking to the plugin. But that is a problem for an ambitious maintainer of `package:wakelock_plus` (or author of an alternative package) to solve, not us :slightly_smiling_face:. I think we'll only have one of these active at a time.

gnprice (commented):
Thanks @rajveermalviya for noticing and fixing this! And thanks @Khader-1 for the testing and @chrisbobbe for the previous reviews.

Generally these changes look great. One small comment below, and there's the nit Chris pointed out above.

gnprice:
Also just pushed an added docs commit:
3c782cfdb binding [nfc]: Expand toggleWakelock doc to clarify what it does

with information I found from digging into the implementation of that `package:wakelock_plus` plugin.

PR #917, opened 2024-08-27
Systematically survey message content for unimplemented features.

PIG208:
Currently a WIP. The code needs a bit of cleanup.

### DIFF
--- a/tools/content/check-features
+++ b/tools/content/check-features
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+default_steps=(fetch check)
+
+usage() {
+    cat <<EOF
+usage: tools/content/check-features [OPTION]... [STEP]... <CORPUS_DIR>
+
+Fetch messages from a Zulip server and check the content parser for
+unimplemented features.
+
+By default, run the following steps:
+  ${default_steps[*]}
+
+CORPUS_DIR is required.  It is the directory to store or read corpus files.
+This directory will be created if it does not exist already.
+
+The steps are:
+
+  fetch     Fetch the corpus needed from the server specified via the config
+            file into \`CORPUS_DIR\` incrementally.  This step can take a long
+            time on servers with a lot of public messages when starting from
+            scratch.
+            This wraps around tools/content/fetch_messages.dart.
+
+  check     Check for unimplemented content parser features.  This requires
+            the corpus directory \`CORPUS_DIR\` to contain at least one corpus
+            file.
+            This wraps around tools/content/unimplemented_features_test.dart.
+
+Options:
+
+  --config <FILE>
+            A zuliprc file with identity information including email, API key
+            and the Zulip server URL to fetch the messages from.
+            Mandatory if running step \`fetch\`.  To get the file, see
+            https://zulip.com/api/configuring-python-bindings#download-a-zuliprc-file.
+
+  --verbose Print more details about everything, especially when checking for
+            unsupported features.
+
+  --help    Show this help message.
+EOF
+}
+
+opt_corpus_dir=
+opt_zuliprc=
+opt_verbose=
+opt_steps=()
+while (( $# )); do
+    case "$1" in
+        fetch|check) opt_steps+=("$1"); shift;;
+        --config) shift; opt_zuliprc="$1"; shift;;
+        --verbose) opt_verbose=1; shift;;
+        --help) usage; exit 0;;
+        *)
+            if [ -n "$opt_corpus_dir" ]; then
+                # Forbid passing multiple corpus directories.
+                usage >&2; exit 2
+            fi
+            opt_corpus_dir="$1"; shift;;
+    esac
+done
+
+if [ -z "$opt_corpus_dir" ]; then
+    echo  >&2 "Error: Positional argument CORPUS_DIR is required."
+    echo  >&2
+    usage >&2; exit 2
+fi
+
+if (( ! "${#opt_steps[@]}" )); then
+    opt_steps=( "${default_steps[@]}" )
+fi
+
+run_fetch() {
+    if [ -z "$opt_zuliprc" ]; then
+        echo >&2 "Error: Option \`--config\` is required for step \`fetch\`."
+        echo >&2
+        usage >&2; exit 2
+    fi
+
+    if [ -n "$opt_verbose" ]; then
+        echo "Fetching all public messages using API config \"$opt_zuliprc\"." \
+             " This can take a long time."
+    fi
+    # This may have a side effect of creating or modifying the corpus
+    # file named after the Zulip server's host name.
+    dart tools/content/fetch_messages.dart --config-file "$opt_zuliprc" \
+        --corpus-dir "$opt_corpus_dir" \
+    || return 1
+}
+
+run_check() {
+    flutter test tools/content/unimplemented_features_test.dart \
+        --dart-define=corpusDir="$opt_corpus_dir" \
+        --dart-define=verbose="$opt_verbose" \
+    || return 1
+}
+
+for step in "${opt_steps[@]}"; do
+    echo "Running ${step}"
+    case "${step}" in
+        fetch) run_fetch ;;
+        check) run_check ;;
+        *)     echo >&2 "Internal error: unknown step ${step}" ;;
+    esac
+done
--- a/tools/content/fetch_messages.dart
+++ b/tools/content/fetch_messages.dart
@@ -1,0 +1,233 @@
+import 'dart:convert';
+import 'dart:io';
+import 'dart:math';
+
+// Avoid any Flutter-related dependencies so this can be run as a CLI program.
+import 'package:args/args.dart';
+import 'package:http/http.dart' as http;
+import 'package:ini/ini.dart' as ini;
+import 'package:zulip/api/backoff.dart';
+
+import 'model.dart';
+
+/// Fetch all public message contents from a Zulip server in bulk.
+///
+/// It outputs JSON entries of the message IDs and the rendered HTML contents in
+/// JSON Lines (https://jsonlines.org) format. The output can be used later to
+/// perform checks for discovering unimplemented features.
+///
+/// Because message IDs are only unique within a single server, the script
+/// names corpora by the server host names.
+///
+/// This script is meant to be run via `tools/content/check-features`.
+///
+/// For more help, run `tools/content/check-features --help`.
+///
+/// See also:
+/// * tools/content/unimplemented_features_test.dart, which runs checks against
+///   the fetched corpora.
+void main(List<String> args) async {
+  final argParser = ArgParser();
+  argParser.addOption(
+    'config-file',
+    help: 'A zuliprc file with identity information including email, API key\n'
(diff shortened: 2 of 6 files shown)


PIG208:
Should be ready for review once the CI passes

chrisbobbe:
Exciting!!

Would it be easy to also include the number of messages with each feature?

PIG208:
Yeah, it should be straightforward.

PIG208:
TODO: Add this to `tools/check` making it easier to run.

PIG208:
In this revision, I moved to using `zuliprc` to make it easier to run it with `tools/check parser` without specifying additional arguments. The new suite is not included in `tools/check --all`. Just entering `tools/check parser` should give you sufficient guidance on how to use it.

PIG208:
For reviewing this, it might be helpful to test `tools/check parser` and see if you can get it working without needing to look at the commit message. Because it is meant to be re-run in the future. It might be easier to test with an API key from a low-traffic public Zulip server.

PIG208:
The public message history from CZO can be downloaded from [this protected dropbox link](https://www.dropbox.com/scl/fi/5ugemicgz6lwrsxscqvwk/corpuses.zip?rlkey=xdje3piu5scypt3glc0n4lbrz&st=muzlmp6t&dl=0). We probably will think of a better way of reusing this in the future.

chrisbobbe:
Nice! This worked for me locally :) and nothing stood out to me from a quick skim of the code. Marking for @gnprice's review.

tools/content/unimplemented_features_test.dart
```diff
@@ -0,0 +1,147 @@
...
+/// standard output, followed by the details of each feature.  To look for live
+/// examples, you can search on the Zulip community by message ID from all
+/// public channels.
+///
+/// For example, a search query like "near: 12345 channels: public" would work.
+///
+/// See also:
+/// * lib/model/content.dart, which implements of the content parser.
+/// * tools/content/fetch_messages.dart, which produces the corpuses.
+void main() async {
```
gnprice, line 41: The test can be run manually via:
        `flutter test --dart-define=corpusDir=path/to/corpusDir tools/content`

This is a pretty funky command line.

There's a good reason it's done this way, and we smooth it over in the end by having a nice wrapper script. But the wrapper script should appear in the same commit as this file does, so that the "test" file offers a reasonable way to run it from the beginning.

If you want to split the changes into two commits, probably a good way to do so would be one commit for fetching and another for parsing.

tools/content/unimplemented_features_test.dart
```diff
@@ -0,0 +1,147 @@
+// Override `fluter test`'s default timeout
```
gnprice, line 1: There's no `#!` line ("shebang line") here, so trying to execute it directly won't work. It therefore shouldn't have the executable flag set (unlike `check-features`).

tools/content/fetch_messages.dart
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env dart
```
gnprice, line 1: Probably cleanest to leave the shebang line (and the executable flag) off of this file, too. Then it's clear there's one intended way to invoke this functionality, via the `check-features` wrapper script.

tools/content/check-features
```diff
@@ -0,0 +1,108 @@
...
+By default, run the following steps:
+  ${default_steps[*]}
+
+CORPUS_DIR is required.  It is the directory to store or read corpus files
+which will be created if it does not exist already.
+
+The steps are:
+
+  fetch     Fetch the corpuses needed from the server specified \`--config\`
+            file into \`CORPUS_DIR\` incrementally.  This step can take a long
```
gnprice, line 22: I think I'm not succesfully parsing this sentence.

tools/content/check-features
```diff
@@ -0,0 +1,108 @@
...
+    esac
+done
+
+if [ -z "$opt_corpus_dir" ]; then
+    echo  >&2 "Error: Positional argument CORPUS_DIR is required."
+    echo  >&2
+    usage >&2; exit 2
+fi
+
+if (( ! "${#opt_steps[@]} " )); then
```
gnprice, line 72: Is the space before the closing `"` intended?

tools/content/fetch_messages.dart
```diff
@@ -0,0 +1,223 @@
...
+    valueHelp: 'path/to/zuliprc',
+  );
+  argParser.addOption(
+    'corpus-dir',
+    help: 'The directory to look for/store the corpus file.  If not given,\n'
+          'the script will write output to stdout.  Otherwise, this will\n'
+          'first read from the existing corpus file (assumed to be named\n'
+          'as "host name of the Zulip server.jsonl") to avoid duplicates\n'
+          'before fetching more messages',
+    valueHelp: 'path/to/czo.jsonl',
```
gnprice, line 48: This example value looks like it doesn't match the meaning — it looks like it's meant as the name of the file itself, rather than a directory.

tools/content/fetch_messages.dart
```diff
@@ -0,0 +1,223 @@
...
+    help: 'A zuliprc file with identity information including email, API key\n'
+          'and the Zulip server URL to fetch the messages from (required).\n\n'
+          'To get the file, see\n'
+          'https://zulip.com/api/configuring-python-bindings#download-a-zuliprc-file.',
+    valueHelp: 'path/to/zuliprc',
+  );
+  argParser.addOption(
+    'corpus-dir',
+    help: 'The directory to look for/store the corpus file.  If not given,\n'
+          'the script will write output to stdout.  Otherwise, this will\n'
```
gnprice, line 44: Writing to stdout seems like a bit of a pitfall for the unwary — it's going to send a lot of spew to one's terminal, right?

I imagine it was helpful when writing and debugging the script, and you were editing the code to control how much output it would send.

Fine to keep the option of stdout, but let's make it not the default. It could be a `--stdout` flag, incompatible with `--corpus-dir`.

tools/content/fetch_messages.dart
```diff
@@ -0,0 +1,223 @@
...
+
+    // This I/O operation could fail, but crashing is fine here.
+    final flushFuture = output.flush();
+    // Make sure the delay happens concurrently to the flush.
+    await Future<void>.delayed(fetchInterval);
+    await flushFuture;
+    backoff = null;
+  }
+  exit(0);
+}
```
gnprice, line 173: Does this `exit` call differ from simply falling off the end of `main` here and returning?

tools/content/fetch_messages.dart
```diff
@@ -0,0 +1,223 @@
...
+  // See also:
+  //   https://zulip.com/api/configuring-python-bindings#configuration-keys-and-environment-variables
+  final parsedConfig = ini.Config.fromString(configFile.readAsStringSync());
+  final email = parsedConfig.get('api', 'email') as String;
+  final apiKey = parsedConfig.get('api', 'key') as String;
+  final site = Uri.parse(parsedConfig.get('api', 'site') as String);
+
+  final outputDirStr = parsedArguments['corpus-dir'] as String?;
+  final fetchNewer = parsedArguments['fetch-newer'] as bool;
+  int? anchorMessageId;
```
gnprice, line 103: In a CLI program I think it's generally helpful to split the parsing of the command line (and of config that's closely tied to the command line) from the rest of the logic.

So between these two lines is basically where that transition happens — we're done inspecting `parsedArguments` (and `parsedConfig`), and we're starting to do the work and to have side effects. It'd therefore be good to take the bottom half of the function, below this point, and move it to its own function; `main` would end by calling that function, passing a bunch of arguments like `email` and `outputDirStr`.

tools/content/model.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:io';
+import 'dart:convert';
+
+import 'package:json_annotation/json_annotation.dart';
+
+/// A data structure representing a message.
+@JsonSerializable()
```
gnprice, line 7: It looks like this `JsonSerializable` (and so the `json_annotation` import) doesn't end up getting used.

tools/content/unimplemented_features_test.dart
```diff
@@ -0,0 +1,147 @@
...
+/// examples, you can search on the Zulip community by message ID from all
+/// public channels.
+///
+/// For example, a search query like "near: 12345 channels: public" would work.
+///
+/// See also:
+/// * lib/model/content.dart, which implements of the content parser.
+/// * tools/content/fetch_messages.dart, which produces the corpuses.
+void main() async {
+  Future<void> checkForUnimplementedFeatureInFile(File file) async {
```
gnprice, line 42: nit:
```suggestion
  Future<void> checkForUnimplementedFeaturesInFile(File file) async {
```

It's checking for any possible unimplemented features, right? Vs. some particular unimplemented feature.

tools/content/unimplemented_features_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+  // `featureName` is a prettified identifier used for categorizing
+  // unimplemented features that are likely closely related.
+  final String featureName;
+  final htmlNode = value.debugHtmlNode;
+  if (htmlNode is dom.Element) {
+    if (htmlNode.className.isEmpty) {
+      featureName = '<${htmlNode.localName!}>';
+    } else {
+      featureName = '<${htmlNode.localName!} class="${htmlNode.classes.join(" ")}">';
```
gnprice, line 140:
```suggestion
      featureName = '<${htmlNode.localName!} class="${htmlNode.className}">';
```

Or is the difference useful?

(`className` is the original data; `classes` is a view on it.)

gnprice (commented):
Thanks @PIG208 for building this, and @chrisbobbe for the previous review!

I'm not yet done reading this, but initial comments below. Most are small.

tools/content/unimplemented_features_test.dart
PIG208: The `classes` view just trims each individual name, I think it is not necessary here.

tools/content/fetch_messages.dart
PIG208: The thing about writing to stdout and keeping the script quiet was mainly for usability with pipes. I think we can eliminate this option and just make `--corpus-dir` mandatory.

tools/content/unimplemented_features_test.dart
PIG208: Putting them in the same commit makes sense. There is not much complexity that can be resolved by a sequence of separate ones.

PIG208:

gnprice (commented):
OK, I've read through the whole thing — and I've also now run it myself, and browsed through the output. Thanks again @PIG208 for building this!

Below are a few small comments. Just one (the last) affects behavior; it's a papercut I ran into when using the script.

Then after these small things are fixed, I'd like to go ahead and merge this. It works and has served its purpose; any deeper changes can wait until a possible future where we're using this script more extensively.

(I discovered a few new tidbits from reading through the output myself: https://github.com/zulip/zulip-flutter/issues/921#issuecomment-2420926837 . Definitely glad the script makes nice detailed output to look at.)

PIG208:
Thanks for the review! I have updated the PR skipping the `_walk` refactor, which conceptually sounds like a good practice to be aware of.

gnprice:

PR #919, opened 2024-08-28
Handle the case when a code block has no grandchildren

PIG208:
For some very old messages (e.g.: [this](https://chat.zulip.org/#narrow/stream/2-general/topic/Error.20in.20dev.20server/near/18765)), a code block might contain no grandchildren, or equivalently, the `<pre></pre>` block within the `<div class="codehilite"></div>` element is empty.

While these messages might be present on CZO only due to some ancient server bugs, the client should not crash when it does encounter them.

```
Crashes:
- RangeError (length): Invalid value: Valid value range is empty: -1
  Oldest message: 5020; newest message: 29855
```

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -990,7 +990,7 @@ class _ZulipContentParser {
       if (child is! dom.Element) return null;
       if (child.localName != 'pre') return null;
 
-      if (child.nodes.length > 2) return null;
+      if (child.nodes.length > 2 || child.nodes.isEmpty) return null;
       if (child.nodes.length == 2) {
         final first = child.nodes[0];
         if (first is! dom.Element
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -986,7 +986,7 @@ class _ZulipContentParser {
           && divElement.className == "codehilite");
 
       if (divElement.nodes.length != 1) return null;
-      final child = divElement.nodes[0];
+      final child = divElement.nodes.single;
       if (child is! dom.Element) return null;
       if (child.localName != 'pre') return null;
 
@@ -997,7 +997,7 @@ class _ZulipContentParser {
             || first.localName != 'span'
             || first.nodes.isNotEmpty) return null;
       }
-      final grandchild = child.nodes[child.nodes.length - 1];
+      final grandchild = child.nodes.last;
       if (grandchild is! dom.Element) return null;
       if (grandchild.localName != 'code') return null;
 
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -344,6 +344,17 @@ class ContentExample {
       ]),
     ]);
 
+  // Current servers no longer produce this, but it can be found in ancient
+  // messages.  For example:
+  //   https://chat.zulip.org/#narrow/stream/2-general/topic/Error.20in.20dev.20server/near/18765
+  static final codeBlockWithEmptyBody = ContentExample(
+      'code block, with an empty body',
+      '```',
+      '<div class="codehilite"><pre></pre></div>', [
+      blockUnimplemented(
+          '<div class="codehilite"><pre></pre></div>'),
+  ]);
+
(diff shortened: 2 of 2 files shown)


PIG208:
#917 is stacked on top of this.

test/model/content_test.dart
```diff
@@ -344,6 +344,14 @@ class ContentExample {
       ]),
     ]);
 
+  static final codeBlockWithEmptyBody = ContentExample(
```
chrisbobbe, line 347: Let's have a comment that makes it clear we don't expect this in messages from current servers. The link to the ancient message you put in the PR description is helpful; we could include that here too.

lib/model/content.dart
```diff
@@ -990,14 +990,14 @@ class _ZulipContentParser {
...
-      if (child.nodes.length > 2) return null;
+      if (child.nodes.length > 2 || child.nodes.isEmpty) return null;
       if (child.nodes.length == 2) {
         final first = child.nodes[0];
         if (first is! dom.Element
             || first.localName != 'span'
             || first.nodes.isNotEmpty) return null;
       }
-      final grandchild = child.nodes[child.nodes.length - 1];
+      final grandchild = child.nodes.last;
```
chrisbobbe, line 1000: Is this a pure refactor, or does this change behavior somehow? I noticed there's a line `final first = child.nodes[0];` above this; should that also be changed to say `child.nodes.first`?

chrisbobbe (commented):
Thanks! Some small comments below.

lib/model/content.dart
PIG208: This refactoring is purely for readability. Let's get an nfc commit for both.

chrisbobbe:
Thanks! I happened to check in on this PR and noticed that the current revision (ending in fb5ea67afd48ff9f4c931c38ab02a38a1e780087) responds to my review, so I'm comfortable marking this as ready for Greg's review.

PIG208:
I think I didn't post an update message due to the CI failure, but those are the deprecation warnings. So it should be fine once we have them fixed.

chrisbobbe:
Ahh makes sense; yeah. This is #923 and I'm hoping to hear Greg's thoughts on it :)

lib/model/content.dart
```diff
@@ -986,18 +986,18 @@ class _ZulipContentParser {
           && divElement.className == "codehilite");
 
       if (divElement.nodes.length != 1) return null;
-      final child = divElement.nodes[0];
+      final child = divElement.nodes.first;
```
gnprice, line 989: If we use a fancy getter here, it should be `single` — that way we correctly express the expectation that this is the only node (not the first of potentially several nodes).

lib/model/content.dart
```diff
@@ -986,18 +986,18 @@ class _ZulipContentParser {
...
       if (child is! dom.Element) return null;
       if (child.localName != 'pre') return null;
 
-      if (child.nodes.length > 2) return null;
+      if (child.nodes.length > 2 || child.nodes.isEmpty) return null;
```
gnprice, line 993: nit: include the key "why" information in the commit message, rather than rely on the PR description. We aim for the Git history to be able to serve as our primary record of what we did.

In particular key points include
* this was a crash bug;
* the affected messages in public chat.zulip.org history ranged from message ID 5020 to 29855, all in 2016.

Details that can be read off from the commit's added test case don't need to be repeated in the commit message, though.

gnprice (commented):
Thanks @PIG208 for the fix, and @chrisbobbe for the reviews!

Generally this looks good. Two small comments below.

PIG208:
Thanks for the review. Updated the PR!

gnprice:
Thanks! Looks good; merging, with a tweak to clarify the commit message a bit:

```
         A code block containing no grandchildren (i.e. a
         <div class="codehilite"> element containing an empty <pre> block)
    -    for some very dated messages can crash the parser if not handled.
    +    appears in some very old messages.  The parser was crashing on them.
```

Issue #915, opened 2024-08-23
Compose-box layout redesign


gnprice:
From @terpimost we have a spiffy new design for the compose box:
![image](https://github.com/user-attachments/assets/bf431527-e870-4d19-8747-f57a7934e8ec)

Details are in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3954-13395&m=dev

This issue is for updating the compose box to match that design. That means:
* the white background, the border from the message list, where applicable the separator between topic and content
* the text styling for topic and for content
* the styling of the send button and the compose buttons (attach file, take photo, etc.)
* the height and scrolling behavior of the content input
* other related details I'm overlooking at the moment

Out of scope:
* the set of buttons that exist — adding buttons corresponds to new features, which will happen in their own issues
* the "..." overflow button, scrolling, and fading for excess compose buttons — for now we have just three of them, not enough to make those come into effect
* the autocomplete results — those are #913 and #914
* editing a message — that's #126
* loading state and error feedback for message send — those are tracked in #720

PR #928, opened 2024-09-04
Compose box redesign

PIG208:
Fixes: #915


PIG208:
(getting the screenshots ready)

PIG208:
The design specifies a 8px high padding before the message content:

image

The content can be scrolled past the padding, then clipped at the top border of the compose box:

image

This turned out to be tricky to implement, because `InputDecorator`'s `contentPadding` also pads the scrollable view, and the content cannot be scrolled past the padding, leaving a gap:
| with padding | without padding |
| - | - |
| image | Screenshot from 2024-09-07 00-50-07 |

A previous workaround was to fully vertically expand `TextField` by wrapping it within a `SingleChildScrollView` (implemented in e16251febac9b000c50969845025bc4a1b54e67a), which does not handle the case that the cursor gets scrolled out of view when editing:

image

PIG208:
I think I have found a true simple solution to this, by setting `TextField.clipBehavior` to `Clip.none`, set `InputDecoration.contentPadding` and wrap the `TextField` with a `ClipRect` under our control. This seems to work perfectly without the cursor issue.

PIG208:
<details>
<summary>Preview</summary>


| light | dark |
| -     |   -  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |


</details>

PIG208:
Updated the PR with the screenshots. Ready for @chrisbobbe's review!

alya:
Thanks! Could you please add a link to the design in Figma?

alya:
I haven't compared carefully vs. the designs in Figma, but the screenshots look nice!

lib/widgets/compose_box.dart
```diff
@@ -903,45 +904,47 @@ class _ComposeBoxLayout extends StatelessWidget {
...
-          borderRadius: BorderRadius.all(Radius.circular(4.0)),
-          borderSide: BorderSide.none),
-        filled: true,
-        fillColor: colorScheme.surface,
-      ),
-    );
+        // Leaving `contentPadding` out would still result in the same layout.
+        // For better predictability, set it explicitly instead of relying on
+        // the default value.
+        contentPadding: EdgeInsets.zero,
```
chrisbobbe, line 974: This didn't turn out to be true in my testing just now:

| with `EdgeInsets.zero` | left out |
| --- | --- |
| image | image |

I see the layout shifting if I open these images in separate browser tabs and switch between them.

chrisbobbe (commented):
Exciting! It feels nice to be moving forward with our new, finalized designs. 🙂

Comments below from a light review; I may have more, later, when I have time to do a deeper review.

lib/widgets/compose_box.dart
```diff
@@ -867,18 +867,16 @@ class _ComposeBoxContainer extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    ColorScheme colorScheme = Theme.of(context).colorScheme;
+    final designVariables = DesignVariables.of(context);
```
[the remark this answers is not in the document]
PIG208: I have been wondering about that too. I had 3 commits in an earlier revision, where 0e99b76f508a9759323c59f05a1477673a04938b implements the entire redesign, which would be the alternative to the now more granular commit sequence. I guess it might help to use that structure with a checklist of redesign details.

Or maybe we can try to strike a balance between organizing these changes in a single a big commit or multiple small ones. There are some arguably distinctive parts of the redesign like the icon changes that might be extracted.

lib/widgets/compose_box.dart
```diff
@@ -841,37 +844,17 @@ class _SendButtonState extends State<_SendButton> {
...
-          minimumSize: WidgetStatePropertyAll(Size.square(_sendButtonSize)),
-          // With the default of [MaterialTapTargetSize.padded], not just the
-          // tap target but the visual button would get padded to 48px square.
-          // It would be nice if the tap target extended invisibly out from the
-          // button, to make a 48px square, but that's not the behavior we get.
-          tapTargetSize: MaterialTapTargetSize.shrinkWrap,
-        ),
-        color: foregroundColor,
+        color: _hasValidationErrors
+          // TODO(design): need send button color when disabled
```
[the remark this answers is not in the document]
PIG208: Hmm, thinking about this, I believe that we can just skip disabling the send button when the content body is empty. This is the only case when we have the send button but not its friends disabled. Once we get to #720, we can implement something that overrides the opacity of all these UI elements.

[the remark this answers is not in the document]
chrisbobbe: Sure, that sounds reasonable to me.

lib/widgets/compose_box.dart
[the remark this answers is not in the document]
chrisbobbe: Happy to defer to @gnprice's judgment here. 🙂

lib/widgets/compose_box.dart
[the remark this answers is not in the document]
PIG208: Just double-checked and found that we visually disable this button when uploading too. In both cases only the send button is disabled. So I plan to leave it as-is for now to keep the current behavior.

PIG208:
Thanks for the previous comments @chrisbobbe! This is ready for review now.

PIG208:
This is ready for review now (forgot to comment on that).

PIG208:
Extracted `InsetShadowBox` so that #853 can reuse this.

lib/widgets/compose_box.dart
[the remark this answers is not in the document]
gnprice: I haven't yet read this PR in detail, but I'm hopeful that this redesign can be done as a series of commits that are usefully broken up from each other, so that a reader is able to focus on one aspect at a time.

If there are commits where it doesn't seem possible to write a clean commit message, though, that'd be a sign that those commits should be squashed with related changes (if that enables a clean commit message).

PIG208:
Will push an update here soon.

chrisbobbe:
Cool, reviewing now.

chrisbobbe (changes requested):
Thanks! Comments below, this time from a thorough review. I noticed two more items to add to the "out of scope" section in the issue (#915), and I've just added them:

> * editing a message — that's #126
> * loading state and error feedback for message send — those are tracked in #720

lib/widgets/compose_box.dart
```diff
@@ -286,30 +286,47 @@ class _ContentInput extends StatelessWidget {
...
-          fieldViewBuilder: (context) {
-            return TextField(
+    const verticalPadding = 8.0;
+    const contentLineHeight = 22.0;
+
+    final designVariables = DesignVariables.of(context);
+
+    return ConstrainedBox(
+      constraints: const BoxConstraints(
+        // The minimum height fits a little more than 2 lines to match the spec
```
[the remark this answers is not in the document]
PIG208: It would be pretty hard to adapt to multiple font sizes. We could use `minLines` on `TextField`, but not `maxLines`. Maybe we should bend the design and make the content input a bit taller.

lib/widgets/compose_box.dart
```diff
@@ -19,8 +19,8 @@ import 'icons.dart';
 import 'store.dart';
 import 'theme.dart';
 
-const double _inputVerticalPadding = 8;
-const double _sendButtonSize = 36;
+const double _composeButtonWidth = 44;
+const double _composeButtonHeight = 42;
```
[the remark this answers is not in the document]
PIG208: Renamed `_composeButtonHeight` to `_composeButtonsRowHeight` so that's a bit more accurate.

lib/widgets/compose_box.dart
```diff
@@ -19,8 +19,8 @@ import 'icons.dart';
 import 'store.dart';
 import 'theme.dart';
 
-const double _inputVerticalPadding = 8;
```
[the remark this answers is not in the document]
PIG208: Hmm, that would affect other `IconButton`s including the back button on the AppBar. Perhaps we should do that in a cleanup PR. Refactored the theme setup so that the splash effect is properly disabled.

PIG208:
Thanks. The PR is ready for review again!

chrisbobbe (commented):
Thanks! Comments below.

PIG208:

The last three commits are for the proof-of-concept tests for the content input scaling feature; they are optional:

* 3e282e4b wip; compose_box content input text scaling test
* 81b27f55 compose_box test: Set topic and content through a helper (cherry-picked from #1033)
* 93d5f999 compose_box test: Use a layout similar to the message list page (#1034)

Not sure if this aspect of the UI requires its own set of tests verifying the maximum number of lines to display. It basically verifies that the max height sticks to a certain number, and that the maximum number of visible lines matches where we expected from the figma design.

test/widgets/compose_box_test.dart
```diff
@@ -55,13 +56,15 @@ void main() {
...
     await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
-      child: Column(
-        // This positions the compose box at the bottom of the screen,
-        // simulating the layout of the message list page.
-        children: [
-          const Expanded(child: SizedBox.expand()),
-          ComposeBox(controllerKey: controllerKey, narrow: narrow),
-        ])));
+      child: MediaQuery(
+        data: MediaQueryData(textScaler: textScaler ?? TextScaler.noScaling),
```
chrisbobbe, line 60: If we use the global `textScaleFactorTestValue` (see some other tests that use this), we can skip the `MediaQuery` widget, making the test's widget tree more representative.

chrisbobbe (commented):
Thanks, this is getting close! Comments below, all small.

PIG208:

There is [another revision](https://github.com/zulip/zulip-flutter/compare/main...PIG208:zulip-flutter:60f153165bcb883b2a7f2b513da7586420cd3bcb) on top of this that I didn't push, because it reorganizes the commits (even though the diff with all commits combined is identical to this one), which might make verifying the update a bit more difficult.

chrisbobbe:
`tools/check` is failing at this commit:

```
compose_box: Use new icons from the redesign
```

```
00:19 +1525 ~166: /Users/chrisbobbe/dev/zulip-flutter/test/widgets/compose_box_test.dart: ComposeBox typing notices hitting send button sends a "typing stopped" notice
The following assertion was thrown running a test:
The finder "Found 0 widgets with icon "IconData(U+0E571)": []" (used in a call to "tap()") could not
find any matching widgets.

#0      WidgetController._getElementPoint (package:flutter_test/src/controller.dart:1895:7)
#1      WidgetController.getCenter (package:flutter_test/src/controller.dart:1795:12)
#2      WidgetController.tap (package:flutter_test/src/controller.dart:1043:18)
#3      main.<anonymous closure>.<anonymous closure> (file:///Users/chrisbobbe/dev/zulip-flutter/test/widgets/compose_box_test.dart:281:20)
<asynchronous suspension>
#4      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:189:15)
<asynchronous suspension>
#5      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1027:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

The test description was:
  hitting send button sends a "typing stopped" notice
════════════════════════════════════════════════════════════════════════════════════════════════════
00:19 +1525 ~166 -1: /Users/chrisbobbe/dev/zulip-flutter/test/widgets/compose_box_test.dart: ComposeBox typing notices hitting send button sends a "typing stopped" notice [E]
  Test failed. See exception logs above.
  The test description was: hitting send button sends a "typing stopped" notice
```

PIG208:
Pushed to fix this. Thanks!

chrisbobbe (commented):
Thanks! Small comment below, and bump on https://github.com/zulip/zulip-flutter/pull/928#discussion_r1833431291; I still see a few `.withValues(alpha:`s in compose_box.dart.

In fact, would it make sense to add `withFadedAlpha` in its own prep commit, and basically do this (from https://github.com/zulip/zulip-flutter/pull/928#discussion_r1815661095) through the whole codebase:

> If we have that multiply-alpha method on `ColorExtension`, it seems wise to prefer that over `.withValues(alpha:` whenever the receiver is a color specified in the Figma, even if the color is currently opaque, because it could easily change to non-opaque in the future.

We should at least do so where we're newly implementing the Figma.

PIG208:
Added:

* eaed8fba emoji: Use withFadedAlpha for highlightColor
* b60f13c6 color: Add ColorExtension.withFadedAlpha

There are some remaining callers of `withValues`:

<details>
<summary>Details</summary>

```
$ git grep withValues
lib/widgets/channel_colors.dart:          .withValues(alpha: 0.3),
lib/widgets/channel_colors.dart:          .withValues(alpha: 0.3),
lib/widgets/color.dart:    return withValues(alpha: a * factor);
lib/widgets/content.dart:      colorThematicBreak: const HSLColor.fromAHSL(1, 0, 0, .87).toColor().withValues(alpha: 0.2),
lib/widgets/emoji_reaction.dart:      borderSelected: Colors.black.withValues(alpha: 0.45),
lib/widgets/emoji_reaction.dart:      borderUnselected: Colors.black.withValues(alpha: 0.05),
lib/widgets/emoji_reaction.dart:      bgSelected: Colors.black.withValues(alpha: 0.8),
lib/widgets/emoji_reaction.dart:      bgUnselected: Colors.black.withValues(alpha: 0.3),
lib/widgets/emoji_reaction.dart:      borderSelected: Colors.white.withValues(alpha: 0.75),
lib/widgets/emoji_reaction.dart:      borderUnselected: Colors.white.withValues(alpha: 0.15),
lib/widgets/emoji_reaction.dart:      textSelected: Colors.white.withValues(alpha: 0.85),
lib/widgets/emoji_reaction.dart:      textUnselected: Colors.white.withValues(alpha: 0.75),
lib/widgets/inset_shadow.dart:      colors: [color, color.withValues(alpha: 0)]));
lib/widgets/lightbox.dart:    final appBarBackgroundColor = Colors.grey.shade900.withValues(alpha: 0.87);
lib/widgets/message_list.dart:      streamRecipientHeaderChevronRight: Colors.black.withValues(alpha: 0.3),
lib/widgets/message_list.dart:      unreadMarkerGap: Colors.white.withValues(alpha: 0.6),
lib/widgets/message_list.dart:      streamRecipientHeaderChevronRight: Colors.white.withValues(alpha: 0.3),
lib/widgets/theme.dart:      bgCounterUnread: const Color(0xff666699).withValues(alpha: 0.15),
lib/widgets/theme.dart:      borderBar: Colors.black.withValues(alpha: 0.2),
lib/widgets/theme.dart:      editorButtonPressedBg: Colors.black.withValues(alpha: 0.06),
lib/widgets/theme.dart:      groupDmConversationIcon: Colors.black.withValues(alpha: 0.5),
lib/widgets/theme.dart:      modalBarrierColor: const Color(0xff000000).withValues(alpha: 0.3),
lib/widgets/theme.dart:      unreadCountBadgeTextForChannel: Colors.black.withValues(alpha: 0.9),
lib/widgets/theme.dart:      bgCounterUnread: const Color(0xff666699).withValues(alpha: 0.37),
lib/widgets/theme.dart:      borderBar: Colors.black.withValues(alpha: 0.5),
lib/widgets/theme.dart:      contextMenuCancelText: const Color(0xffffffff).withValues(alpha: 0.75),
lib/widgets/theme.dart:      editorButtonPressedBg: Colors.white.withValues(alpha: 0.06),
lib/widgets/theme.dart:      labelCounterUnread: const Color(0xffffffff).withValues(alpha: 0.7),
lib/widgets/theme.dart:      labelMenuButton: const Color(0xffffffff).withValues(alpha: 0.85),
lib/widgets/theme.dart:      textInput: const Color(0xffffffff).withValues(alpha: 0.9),
lib/widgets/theme.dart:      groupDmConversationIcon: Colors.white.withValues(alpha: 0.5),
lib/widgets/theme.dart:      modalBarrierColor: const Color(0xff000000).withValues(alpha: 0.5),
lib/widgets/theme.dart:      unreadCountBadgeTextForChannel: Colors.white.withValues(alpha: 0.9),
test/widgets/color_test.dart:        .isSameColorAs(color.withValues(alpha: 0.25));
test/widgets/color_test.dart:        .isSameColorAs(color.withValues(alpha: 0.5));
```

</details>

Most of them, other than the ones in `lib/widgets/channel_colors.dart` are from themes defining the colors. The obvious ones are handled in the new commits, and we can do a full sweep in a follow-up PR.

I also noticed that the text scale tests stop being reliable when the text scale factor is lower than or equal to 0.5, reporting that we get less than 8 fully or partially visible lines. However, that doesn't happen when I run `flutter run test/widgets/compose_box_test.dart`, which is puzzling.

chrisbobbe:
> I also noticed that the text scale tests stop being reliable when the text scale factor is lower than or equal to 0.5

Odd. But iOS at least doesn't let you choose a factor that small. 🤷‍♂️ We've written down some of iOS's options in test/widgets/text_test.dart:

```dart
// From trying the options on an iPhone 13 Pro running iOS 16.6.1:
const kTextScaleFactors = <double>[
  0.8235, // smallest
  1,
  1.3529, // largest without using the "Larger Accessibility Sizes" setting
  3.1176, // largest
];
```

chrisbobbe (commented):
Thanks for the revision! Comments below, all small.

PIG208:
Thanks! Updated the PR.

lib/widgets/color.dart
```diff
@@ -38,4 +38,21 @@ extension ColorExtension on Color {
...
+  /// `factor` must not be less than 0 or greater than 1.
+  ///
+  /// To fade a color variable from [DesignVariables], [ContentTheme], etc.,
+  /// use this instead of calling [withValues] with `factor` passed as `alpha`,
+  /// which simply replaces the color's [a] instead of multiplying by it.
+  /// Using [withValues] gives the same result for an opaque color,
+  /// but a wrong result for a semi-transparent color,
+  /// and we want our color variables to be free to change
+  /// without breaking things.
+  Color withFadedAlpha(double factor) {
```
chrisbobbe, line 53:
```
color [nfc]: Add ColorExtension.withFadedAlpha

We use this instead at most places where we preivously have
`.withValues(alpha:` on color variables, except the one in
`lib/widgets/emoji_reaction.dart` that is semi-transparent,
to keep this change NFC.

This also skips the ones in `lib/widgets/channel_colors.dart`,
and the one in `lib/widgets/lightbox.dart`
(`Colors.grey.shade900.withValues(alpha: 0.87)`), as determining if
those can be swapped out with `withFadedAlpha` seems out of scope as of
now.

```

Commit message nits:

In the first paragraph, I'm not sure what "instead" refers to. How about: "And use it everywhere we were using `.withValues(alpha:` to fade a color variable, except for one place in lib/widgets/emoji_reaction.dart where the variable is semitransparent to begin with. We'll do that as a non-NFC change in a later commit.

About the second paragraph:
- I think the `withValues` calls in lib/widgets/channel_colors.dart should stay. The linked design for the computation (a replit by Vlad) doesn't specify multiplication; it says to set the alpha channel to 0.3, which we're doing with `withValues`.
- I think the `withValues` in lightbox.dart can stay. It's setting an alpha channel on a constant from Flutter (`Colors.grey.shade900`) that realistically won't ever change to be semi-transparent. Even if it did, it's not clear that 0.87 would be the wrong value to end up with.
- Like with the many `withValues` calls in lib/widgets/theme.dart, I think these don't need to be specifically mentioned in the commit message. They're using `withValues(alpha:` as intended, to create a color with its alpha channel set to a value. They might have done that same thing a different way, like in the `Color` constructor. That's different from "fad[ing] a color variable from [DesignVariables] […]", which we often do e.g. to translate from the Figma an `opacity` attribute on an element that's colored with a variable.

chrisbobbe (commented):
Thanks! One small commit-message comment, and I'll mark this for Greg's review.

PIG208:
I have pushed a change reorganizing the commits for better readability. This is effectively an NFC change overall — no functionality change after reordering. 

**Before**
66d55db6 theme: Update borderBar to match Figma design
6dd42643 compose_box: Use new icons from the redesign
f52104a3 compose_box: Use new send button color, shape and icon
b13840c3 compose_box [nfc]: Refactor away shared iconTheme override
d18ec182 fix link; compose_box: Support new compose box layout
d5fa7f93 compose_box: Support redesigned button feedback
23ef514f color [nfc]: Add ColorExtension.withFadedAlpha
a29e69b5 emoji: Use withFadedAlpha for highlightColor
00386935 compose_box: Implement topic input redesign
6df7931e compose_box: Implement content input redesign
f8f97dd7 optional; compose_box content input text scaling test

**After**
_prep changes_
13ab190c color [nfc]: Add ColorExtension.withFadedAlpha
ef6cf91b emoji: Use withFadedAlpha for highlightColor
ce761fdf theme: Update borderBar to match Figma design
a3da9918 compose_box [nfc]: Refactor away shared iconTheme override
2780d25f compose_box: Use new icons from the redesign
_main redesign changes_
17372299 fix link; compose_box: Support new compose box layout
455a28ec compose_box: Use new send button color, shape and icon
70376681 compose_box: Support redesigned button feedback
61904987 compose_box: Implement input redesign
cbac604e optional; compose_box content input text scaling test

**Changes**

- Moved
23ef514f color [nfc]: Add ColorExtension.withFadedAlpha
and
a29e69b5 emoji: Use withFadedAlpha for highlightColor
to the very front of the sequence. They no longer touch the colors that use `withValues` in `lib/widgets/compose_box.dart`, leaving them to the later commits that implement the UI changes.

- Grouped the lighter commits together as prep changes:
17372299 fix link; compose_box: Support new compose box layout
455a28ec compose_box: Use new send button color, shape and icon
70376681 compose_box: Support redesigned button feedback
61904987 compose_box: Implement input redesign
cbac604e (HEAD -> main, me/dev-compose) optional; compose_box content input text scaling test
These commits barely requires checking the Figma redesign, other than confirming the icons

- Moved
f52104a3 compose_box: Use new send button color, shape and icon
after
d18ec182 fix link; compose_box: Support new compose box layout
so that the order is more natural, simplifying the diffs for changes relevant to the send button.

- Combined
00386935 compose_box: Implement topic input redesign
and
6df7931e compose_box: Implement content input redesign
into
61904987 compose_box: Implement input redesign

lib/widgets/color.dart
PIG208: Thanks! Changed the message to
```
And use it everywhere we were using `.withValues(alpha:`
to fade a color variable, except for one place in
`lib/widgets/emoji_reaction.dart` where the variable is
semi-transparent in dark mode to begin with.
We'll do that as a non-NFC change in a later commit.
```
with a minor tweak mentioning that, in dark mode, the color is semi-transparent.

chrisbobbe:
Cool, thanks for the update! Reviewing again now.

lib/widgets/compose_box.dart
```diff
@@ -930,30 +930,22 @@ class _SendButtonState extends State<_SendButton> {
 
   @override
   Widget build(BuildContext context) {
-    final disabled = _hasValidationErrors;
-    final colorScheme = Theme.of(context).colorScheme;
+    final designVariables = DesignVariables.of(context);
```
chrisbobbe, line 998:
```
compose_box: Use new send button color, shape and icon
```

nit: this commit doesn't change the send button's icon; it was changed (to `ZulipIcons.send`) in an earlier commit.

lib/widgets/compose_box.dart
```diff
@@ -473,20 +519,39 @@ class _TopicInput extends StatelessWidget {
...
+                hintText: zulipLocalizations.composeBoxTopicHintText,
+                hintStyle: topicTextStyle.copyWith(
+                  color: designVariables.textInput.withFadedAlpha(0.5))))),
+          SizedBox(height: 0, width: double.infinity,
+            child: DecoratedBox(decoration: BoxDecoration(
+              border: Border(
+                bottom: BorderSide(
+                  width: 1,
+                  color: designVariables.foreground.withFadedAlpha(0.2)))))),
+        ]));
```
chrisbobbe, line 554: The `Column` with `SizedBox(height: 0`, `DecoratedBox`, etc., seems like more complexity than we should need just for a horizontal line. Would this work?

```dart
    return TopicAutocomplete(
      streamId: streamId,
      controller: controller,
      focusNode: focusNode,
      contentFocusNode: contentFocusNode,
      fieldViewBuilder: (context) => Container(
        padding: const EdgeInsets.only(top: 10, bottom: 9),
        decoration: BoxDecoration(border: Border(bottom: BorderSide(
          width: 1,
          color: designVariables.foreground.withFadedAlpha(0.2)))),
        child: TextField(
          controller: controller,
          focusNode: focusNode,
          textInputAction: TextInputAction.next,
          style: topicTextStyle,
          decoration: InputDecoration(
            hintText: zulipLocalizations.composeBoxTopicHintText,
            hintStyle: topicTextStyle.copyWith(
              color: designVariables.textInput.withFadedAlpha(0.5))))));
```

lib/widgets/compose_box.dart
```diff
@@ -972,18 +965,16 @@ class _ComposeBoxContainer extends StatelessWidget {
...
+        border: Border(top: BorderSide(color: designVariables.borderBar))),
       child: Material(
-        color: colorScheme.surfaceContainerHighest,
-        child: SafeArea(
-          minimum: const EdgeInsets.fromLTRB(8, 0, 8, 8),
-          child: Padding(
-            padding: const EdgeInsets.only(top: 8.0),
-            child: child))));
+        color: designVariables.composeBoxBg,
+        child: SafeArea(child: child)));
```
chrisbobbe, line 1034: Could be a followup, but there's something in the current logic that would be nice not to regress on: when the device has left and right insets (as on an iPhone in landscape mode), we don't add more left and right padding on top of the padding that consumes the inset.

Compare:

| Current (branch tip) | Better (less wasted space) |
| --- | --- |
| image | image |

Here's the `git diff -b` to get that "better" version atop your branch:

```diff
$ git diff -b
diff --git lib/widgets/compose_box.dart lib/widgets/compose_box.dart
index f58b865d5..5e9beb87a 100644
--- lib/widgets/compose_box.dart
+++ lib/widgets/compose_box.dart
@@ -1031,7 +1031,8 @@ class _ComposeBoxContainer extends StatelessWidget {
         border: Border(top: BorderSide(color: designVariables.borderBar))),
       child: Material(
         color: designVariables.composeBoxBg,
-        child: SafeArea(child: child)));
+        child: SafeArea(left: false, right: false,
+          child: child)));
   }
 }
 
@@ -1076,16 +1077,17 @@ class _ComposeBoxLayout extends StatelessWidget {
 
     return _ComposeBoxContainer(
       child: Column(children: [
-        Padding(
-          padding: const EdgeInsets.symmetric(horizontal: 16),
+        SafeArea(
+          minimum: const EdgeInsets.symmetric(horizontal: 16),
           child: Theme(
             data: inputThemeData,
             child: Column(children: [
               if (topicInput != null) topicInput!,
               contentInput,
             ]))),
-        Container(
-          padding: const EdgeInsets.symmetric(horizontal: 8),
+        SafeArea(
+          minimum: const EdgeInsets.symmetric(horizontal: 8),
+          child: SizedBox(
             height: _composeButtonsRowHeight,
             child: IconButtonTheme(
               data: iconButtonThemeData,
@@ -1105,6 +1107,7 @@ class _ComposeBoxLayout extends StatelessWidget {
                   ]),
                   sendButton,
                 ]))),
+        ),
       ]));
   }
 }
```

chrisbobbe (commented):
Thanks! The reordering was helpful; I ended up going back through all the commits, instead of mostly just comparing the range-diff, and I found some things I'd missed before, including a branch-structure issue that I should have prioritized earlier. As I say in the comment, I'm happy to wait for Greg's input before asking you to do another revision. 🙂

lib/widgets/compose_box.dart
```diff
@@ -14,13 +14,14 @@ import '../model/compose.dart';
 import '../model/narrow.dart';
 import '../model/store.dart';
 import 'autocomplete.dart';
+import 'color.dart';
```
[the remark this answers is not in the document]
chrisbobbe: I also think "layout" is not a super well-defined term; maybe that's one source of difficulty understanding this commit's focus. If I were implementing this, I might do one of these things:

- Lump basically all the substantive changes into one commit that's simple to describe ('implement redesign'), or:
- If doing it incrementally, instead of one "update the layout" commit, do specific things in separate commits, like:
  - Move all the buttons into one row
  - Update font size
  - Change 8px padding to 16px padding
  - (etc.)

chrisbobbe:
Moving this to integration review since I'm about to be on vacation.

lib/widgets/compose_box.dart
```diff
@@ -927,38 +988,22 @@ class _SendButtonState extends State<_SendButton> {
...
-          // button, to make a 48px square, but that's not the behavior we get.
-          tapTargetSize: MaterialTapTargetSize.shrinkWrap,
-        ),
-        color: foregroundColor,
-        icon: const Icon(Icons.send),
+        icon: Icon(ZulipIcons.send,
+          // We set the color on `Icon` instead of `IconButton` because the
+          // Figma design does not apply it on the background of the button
+          // for states like highlighted, etc.
+          color: iconColor),
```
gnprice, line 1006: I don't follow this comment. This `color` property is about the foreground color, not the background, right?

lib/widgets/compose_box.dart
```diff
@@ -363,34 +367,76 @@ class _ContentInputState extends State<_ContentInput> with WidgetsBindingObserve
...
+    // system text-size setting. We clamp to a max scale factor to limit
+    // how tall the content input can get; that's to save room for the message
+    // list. The user can still scroll the input to see everything.
+    return verticalPadding + 7.727 * scaledLineHeight;
+  }
+
+  static const verticalPadding = 8.0;
+  static const fontSize = 17.0;
+  static const lineHeight = 22.0;
+  static const lineHeightRatio = lineHeight / fontSize;
```
gnprice, line 393: nit: these are all internal, right? Let's make that clear by making them private

lib/widgets/compose_box.dart
```diff
@@ -363,34 +367,76 @@ class _ContentInputState extends State<_ContentInput> with WidgetsBindingObserve
...
               maxLines: null,
               textCapitalization: TextCapitalization.sentences,
-            );
-          }),
-        ));
+              decoration: InputDecoration(
+                // This padding ensures that the user can always scroll long
+                // content entirely out of the top or bottom shadow if desired.
+                // With this and the `minLines: 2` above, an empty content input
+                // gets 60px vertical distance between the top of the top shadow
```
gnprice, line 430: That's at default text scaling, right?

On my phone at my usual settings, I see 54px for this. If I switch my system text-size setting to default, I reproduce 60px.

lib/widgets/compose_box.dart
```diff
@@ -363,34 +367,76 @@ class _ContentInputState extends State<_ContentInput> with WidgetsBindingObserve
...
+                // and the bottom of the bottom shadow (with no text-size
+                // scaling). That's a bit more than the 54px given in the Figma,
```
gnprice, line 432: Ah I see, rereading, there's this reference at the end of the sentence. Clearer if it's moved next to the mention of 60x vertical distance, I think; it wasn't clear to me the first time what this parenthetical was about.

gnprice (commented):
Thanks @PIG208 for building this, and @chrisbobbe for the previous rounds of review!

Various comments below. I've read most of the branch, but not the last commit:
ea7d8b34a compose_box: Support redesigned button feedback

or the tests of the next-to-last, main commit:
1f49f5184 compose_box: Implement most of compose box redesign

lib/widgets/icons.dart
```diff
@@ -33,65 +33,77 @@ abstract final class ZulipIcons {
...
   /// The Zulip custom icon "group_dm".
-  static const IconData group_dm = IconData(0xf10a, fontFamily: "Zulip Icons");
+  static const IconData group_dm = IconData(0xf10c, fontFamily: "Zulip Icons");
 
   /// The Zulip custom icon "hash_sign".
-  static const IconData hash_sign = IconData(0xf10b, fontFamily: "Zulip Icons");
+  static const IconData hash_sign = IconData(0xf10d, fontFamily: "Zulip Icons");
+
+  /// The Zulip custom icon "image".
+  static const IconData image = IconData(0xf10e, fontFamily: "Zulip Icons");
```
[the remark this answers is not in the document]
PIG208: Because the icon wasn't rendering correctly, I resorted to the original source. However, processing the exported icon with the [svg fixer](https://www.npmjs.com/package/oslllo-svg-fixer) seems to fix this, as I just tested.

lib/widgets/compose_box.dart
PIG208: We do want to set the foreground color to `iconColor`; either `IconButton.color` or `Icon.color` will work.

We do want to avoid the default background colors (i.e. the color of the button: `IconButton.hoverColor`, `IconButton.highlightColor`); these are derived from `IconButton.color` if it is set (it is the fallback when the ambient `ButtonStyle.overlayColor` is not set, see [`IconButton.styleFrom`](https://github.com/flutter/flutter/blob/eaa99f28b1cebf39be5fe92633d79af0b64fe329/packages/flutter/lib/src/material/icon_button.dart#L646-L659), which is called by `IconButton.build`).

This exact behavior of falling back to the foreground color (`IconButton.color`) doesn't seem to be mentioned on the colors' dartdoc. When we set `IconButton.highlightColor` through an `IconButtonTheme`, it gets overridden by the color derived from `IconButton.color`.

we can clarify this bit by mentioning the specific button styles involved.

lib/widgets/compose_box.dart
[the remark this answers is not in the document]
PIG208: I agree. I ended up combining the changes without extracting the changes to the colors, layout etc., while leaving the button feedback, icon changes separate because they feel isolated from the main change enough. Making new revisions and verifying the implementation with the design seem to be easier this way.

PIG208:

lib/widgets/compose_box.dart
```diff
@@ -927,38 +987,23 @@ class _SendButtonState extends State<_SendButton> {
...
-          tapTargetSize: MaterialTapTargetSize.shrinkWrap,
-        ),
-        color: foregroundColor,
-        icon: const Icon(Icons.send),
+        icon: Icon(ZulipIcons.send,
+          // We set [Icon.color] instead of [IconButton.color] because the
+          // latter implicitly uses colors derived from it to override the
+          // ambient [ButtonStyle.overlayColor], where we set the color for
+          // the highlight state to match the Figma design.
+          color: iconColor),
```
gnprice, line 1006: Great, thanks — this comment explains it.

test/widgets/compose_box_test.dart
```diff
@@ -49,6 +56,11 @@ void main() {
     await store.addUsers([eg.selfUser, ...users]);
     connection = store.connection as FakeApiConnection;
 
+    if (textScaleFactor != null) {
+      tester.platformDispatcher.textScaleFactorTestValue = textScaleFactor;
+      addTearDown(tester.platformDispatcher.clearTextScaleFactorTestValue);
```
gnprice, line 61: Can this step just go in the test cases that need it? That way we avoid making this shared helper more complex with a growing number of optional features, plus it makes it transparent in the test cases exactly what's happening.

If this step were a bit more code than this (or if we were writing lots of very short tests that used it), another tactic would be to make a helper function just for this, which those tests could call.

test/widgets/compose_box_test.dart
```diff
@@ -668,4 +681,81 @@ void main() {
...
+      }
+      check(height).isNotNull().isCloseTo(maxHeight, 0.5);
+      // The last line added did not stretch the content input,
+      // so only the lines before it are at least partially visible.
+      check(content.split('\n').length - 1).equals(maxVisibleLines);
+    }
+
+    testWidgets('normal text scale factor', (tester) async {
+      TypingNotifier.debugEnable = false;
+      addTearDown(TypingNotifier.debugReset);
```
gnprice, line 723: The reason this step is needed is the `enterText` calls inside `checkContentInputMaxHeight`, right?

In that case, a good home for this is inside that function. That makes clearer the relationship to why it's needed, and as a bonus makes each test case simpler.

test/widgets/compose_box_test.dart
```diff
@@ -668,4 +681,81 @@ void main() {
...
+      // The content input contains two lines at minimum.
+      String content = 'foo\nfoo';
+
+      // On top of the existing two lines, add exactly the number of lines
+      // needed such that the last line will not visible in the content input
+      // after reaching the max height.
+      for (int i = 0; i < maxVisibleLines - 2 + 1; i++) {
+        // Add one line at a time,
+        // until the content input reaches its max height.
+        await tester.enterText(finder, content);
```
gnprice, line 706: nit: just inline `finder`; the name "finder" isn't very informative about what this line is doing

Or use `contentInputFinder`? That'd match the `getRect` call just below, and its name is clear.

test/widgets/compose_box_test.dart
```diff
@@ -668,4 +681,81 @@ void main() {
...
+      final finder = find.byType(ComposeBox);
+
+      double? height;
+      // The content input contains two lines at minimum.
+      String content = 'foo\nfoo';
+
+      // On top of the existing two lines, add exactly the number of lines
+      // needed such that the last line will not visible in the content input
+      // after reaching the max height.
+      for (int i = 0; i < maxVisibleLines - 2 + 1; i++) {
```
gnprice, line 703: This loop is kind of confusing to follow. A very local fix would be to make `i` represent the number of lines in `content`:
```suggestion
      // On top of the existing two lines, add exactly the number of lines
      // needed such that the last line will not visible in the content input
      // after reaching the max height.
      for (int i = 2; i < maxVisibleLines + 1; i++) {
```

I think a further clarification is then to have just one loop variable, the number of lines, instead of effectively two in parallel (`i` and `content`). That can look like this:
```dart
      // Add one line at a time, until the content input reaches its max height.
      int numLines;
      double? height;
      for (numLines = 2; numLines <= 1000; numLines++) {
        final content = List.generate(numLines, (_) => 'foo').join('\n');
        await tester.enterText(finder, content);
        await tester.pump();
        final newHeight = tester.getRect(contentInputFinder).height;
        if (newHeight == height) {
          break;
        }
        height = newHeight;
      }
      check(height).isNotNull().isCloseTo(maxHeight, 0.5);
      // The last line added did not stretch the content input,
      // so only the lines before it are at least partially visible.
      check(numLines - 1).equals(maxVisibleLines);
```

gnprice (commented):
Thanks! The changes in the revision all look great.

I've now read the whole thing; comments below on the remaining parts.

One commit-message nit: use "compose" for the prefix rather than "compose_box", to save a bit of space in our often-crowded summary lines. The shorter version is what we originally used and have used most; see https://github.com/zulip/zulip-flutter/pull/886#issuecomment-2479910865 where I looked at previous commits.

gnprice:
> One commit-message nit: use "compose" for the prefix rather than "compose_box"

Ah, I see we raced and you've already made that change :slightly_smiling_face:

lib/widgets/compose_box.dart
```diff
@@ -1002,45 +1046,54 @@ class _ComposeBoxLayout extends StatelessWidget {
...
-        contentPadding: const EdgeInsets.symmetric(
-          horizontal: 12.0, vertical: _inputVerticalPadding),
-        border: const OutlineInputBorder(
-          borderRadius: BorderRadius.all(Radius.circular(4.0)),
-          borderSide: BorderSide.none),
-        filled: true,
-        fillColor: colorScheme.surface,
-      ),
-    );
+        contentPadding: EdgeInsets.zero,
```
gnprice, line 1056: commit-message nit:

    - For `InputDecorationTheme` on `_ComposeBoxLayout`, we eliminate
      `contentPadding`, while keeping `isDense` as `true` to explicitly
      remove paddings on the input widgets.

I reread this paragraph just now and thought "wait, didn't we set contentPadding to zero? I sure thought this code still set both isDense and contentPadding". (Then I went back to the code to check.) IOW, I read this as saying we stopped passing that argument.

s/eliminate/zero out/ would fix it.

gnprice: Also move the commas: "we eliminate contentPadding while keeping isDense true, to explicitly …".

As is, it looks like it's saying that the way we remove padding is by keeping isDense true, all at the same time as we eliminate contentPadding. In reality (IIUC) both parameters are part of removing padding.

test/widgets/compose_box_test.dart
PIG208: Inlined these without a helper, because they are short enough.

gnprice (commented):
More commit-message nits:

      `ButtonStyle` for the send button that was added in # 399, which is

should complete the link

---

    See also:
    
      - https://github.com/zulip/zulip-flutter/pull/928#discussion_r1815661095
      - https://github.com/zulip/zulip-flutter/pull/928#issuecomment-2335042432,
        which elaborates on the issue we intend to solve with the
        `ClipRect` setup.
      - https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3954-13395
      - https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3862-14350
      - https://github.com/zulip/zulip-flutter/pull/928#discussion_r1842988359

These should each get a few words about what they're about, like the second one has. In particular that's needed so that the reader can tell whether they're relevant for whatever the reader is currently trying to understand. (Otherwise they have to go open all of them one by one… which is extra slow because both Figma, and GitHub in long threads, take a long time to load.)

The second link sounds (from the description here) like it's about the same thing as the second paragraph above. So probably it should just move into that paragraph, like:

      The shadow is always there, but it is only visible when the
      `TextField` is long enough to be scrollable. Discussion here:
        https://github.com/zulip/zulip-flutter/pull/928#issuecomment-2335042432

and then (a) its context is clear and (b) someone reading the paragraph sees there's a relevant link.

    See also:
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3707-41711&n
ode-type=frame&t=sSYomsJzGCt34D8N-0

---

Similarly this should describe how the link relates to what the commit message is saying. If I understand right, that can be:

    Design spec here:
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3707-41711&n
ode-type=frame&t=sSYomsJzGCt34D8N-0

PIG208:
Thanks! Updated the PR.

PIG208:
Moved the links to the paragraph they each correspond to, and added the comment on `ClipRect` (the related paragraph is still kept in the commit, in case the reader is looking back at this for a more verbose explanation on this design's origin).

gnprice:
Looks good — merging. Thanks again for all your work on this!

PR #931, opened 2024-09-07
channel [nfc]: Document methods on ChannelStore; expose debugTopicVisibility directly

gnprice:
These members didn't have docs. They have some nontrivial invariants, so can especially benefit from docs.

Then this arrangement of debugTopicVisibility
follows the pattern we generally use for other members of
ChannelStore and MessageStore, and lets us eliminate the getter
returning a ChannelStoreImpl.

This pattern -- exposing all the data as members directly on the
overall PerAccountStore, delegating to the underlying channel store
and message store as needed -- is convenient because it means that
code consuming the store doesn't have to worry about which sub-area
of the store we've organized a given piece of data into: it's all
just methods on the store.

Noticed this was here while reviewing a PR that included a new
similar deviation from this pattern (/cc @PIG208):
  https://github.com/zulip/zulip-flutter/pull/909#discussion_r1747875935

### DIFF
--- a/lib/model/channel.dart
+++ b/lib/model/channel.dart
@@ -1,3 +1,5 @@
+import 'package:flutter/foundation.dart';
+
 import '../api/model/events.dart';
 import '../api/model/initial_snapshot.dart';
 import '../api/model/model.dart';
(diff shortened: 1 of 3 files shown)


chrisbobbe:
Thanks, LGTM! Merging.

PR #932, opened 2024-09-09
test: Use random topic in eg.streamMessage

gnprice:
This avoids a test accidentally coming to depend on the particular topic name that `eg.streamMessage` happens to choose by default.

Discussed previously at:
  https://github.com/zulip/zulip-flutter/pull/787#discussion_r1710743918

For convenience, we preserve the fact that successive calls to `eg.streamMessage` without a `topic` argument will get the same topic each time.  (This saves various test cases from having to specify the topic just to ensure all their messages have some common topic.)

### DIFF
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -320,6 +320,15 @@ int _lastMessageId = 1000;
 
 const defaultStreamMessageStreamId = 123;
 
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks, LGTM! Merging.

Issue #731, opened 2024-06-12
Lint on unawaited futures


gnprice:
There's a pair of lint rules in the Dart analyzer that detect when we have a Future and neglect to await it:
https://dart.dev/tools/linter-rules/unawaited_futures
https://dart.dev/tools/linter-rules/discarded_futures

(Despite the name, `unawaited_futures` doesn't cover one major class of unawaited futures: those where the enclosing function isn't even `async`. Those are covered by `discarded_futures`.)

It'd be good to enable those. The bugs they catch can be pretty subtle otherwise.

PIG208:
Related to discarded_futures: https://github.com/dart-lang/linter/issues/3739

gnprice:
We resolved a large swath of this issue with #934, enabling the `unawaited_futures` linter rule. That covers the case where the unawaited future appears in an `async` function body.

The remaining task that's open here is to cover the cases where the unawaited future's enclosing function body isn't already marked with `async`. As discussed above, there is a linter rule `discarded_futures` that attempts something in that direction, but it has a lot of false positives which make it not something we can happily adopt.

So in order to make progress on the remaining part of this issue, what's needed is a new lint rule (or changes to `discarded_futures`) upstream. I posted a comment on that upstream issue a little while ago, at the time of the discussion above:
https://github.com/dart-lang/linter/issues/3739#issuecomment-2390079683

With that, I think there's nothing further for us to do on this but wait and hope somebody takes that on. (In principle we could dig into the Dart linter codebase and attempt to contribute a rule ourselves. Maybe at some point in the future we'll do that. But I think this particular desired rule isn't important enough to us to justify that level of investment, and certainly not to do so in the near future.)

So, moving this to our furthest-off milestone, "post-launch".

chrisbobbe:
> So in order to make progress on the remaining part of this issue, what's needed is a new lint rule (or changes to `discarded_futures`) upstream. I posted a comment on that upstream issue a little while ago, at the time of the discussion above:
> [dart-lang/sdk#58889](https://github.com/dart-lang/sdk/issues/58889)

That upstream issue has been marked as fixed! So we can try the `discarded_futures` lint again and see if it still gives false positives.

PR #934, opened 2024-09-10
Enable unawaited_futures

PIG208:
Fixes part of #731

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -25,3 +25,4 @@ linter:
     no_literal_bool_comparisons: true
     prefer_relative_imports: true
     unnecessary_statements: true
+    unawaited_futures: true
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
@@ -377,10 +378,9 @@ class NotificationDisplayManager {
 
     assert(debugLog('  account: $account, narrow: $narrow'));
     // TODO(nav): Better interact with existing nav stack on notif open
-    navigator.push(MaterialAccountWidgetRoute<void>(accountId: account.id,
+    unawaited(navigator.push(MaterialAccountWidgetRoute<void>(accountId: account.id,
       // TODO(#82): Open at specific message, not just conversation
-      page: MessageListPage(initNarrow: narrow)));
-    return;
+      page: MessageListPage(initNarrow: narrow))));
   }
 
   static Future<Uint8List?> _fetchBitmap(Uri url) async {
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
@@ -65,18 +66,16 @@ void main() {
 
     test('send rejects off-realm URL (with default useAuth)', () async {
       void checkAllow(String realmUrl, String requestUrl) {
-        finish(() async {
-          check(await makeRequest(realmUrl, requestUrl))
-            .isA<http.Request>()
-            .url.asString.equals(requestUrl);
-        }());
+        // No need to await directly; `check` ensures the future completes
+        // before the enclosing test is considered complete.
+        unawaited(check(makeRequest(realmUrl, requestUrl))
+          .completes((it) => it.isA<http.Request>()
+            .url.asString.equals(requestUrl)));
       }
 
-      void checkDeny(String realmUrl, String requestUrl) async {
-        finish(() async {
-          await check(makeRequest(realmUrl, requestUrl))
-            .throws<StateError>();
-        }());
+      void checkDeny(String realmUrl, String requestUrl) {
+        unawaited(check(makeRequest(realmUrl, requestUrl))
+          .throws<StateError>());
       }
 
       // Baseline: normal requests are allowed.
@@ -246,7 +245,7 @@ void main() {
   test('API network errors', () async {
     void checkRequest<T extends Object>(
         T exception, Condition<NetworkException> condition) {
-      finish(check(tryRequest(exception: exception))
+      unawaited(check(tryRequest(exception: exception))
         .throws<NetworkException>((it) => it
           ..routeName.equals(kExampleRouteName)
           ..cause.equals(exception)
@@ -300,7 +299,7 @@ void main() {
     void checkMalformed({
         int httpStatus = 400, Map<String, dynamic>? json, String? body}) {
       assert((json == null) != (body == null));
-      finish(check(tryRequest(httpStatus: httpStatus, json: json, body: body))
+      unawaited(check(tryRequest(httpStatus: httpStatus, json: json, body: body))
         .throws<MalformedServerResponseException>((it) => it
           ..routeName.equals(kExampleRouteName)
           ..httpStatus.equals(httpStatus)
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -64,15 +64,19 @@ void main() {
     });
 
     test('send rejects off-realm URL (with default useAuth)', () async {
-      Future<void> checkAllow(String realmUrl, String requestUrl) async {
-        check(await makeRequest(realmUrl, requestUrl))
-          .isA<http.Request>()
-          .url.asString.equals(requestUrl);
+      void checkAllow(String realmUrl, String requestUrl) {
+        finish(() async {
+          check(await makeRequest(realmUrl, requestUrl))
+            .isA<http.Request>()
+            .url.asString.equals(requestUrl);
+        }());
       }
 
-      Future<void> checkDeny(String realmUrl, String requestUrl) async {
-        await check(makeRequest(realmUrl, requestUrl))
-          .throws<StateError>();
+      void checkDeny(String realmUrl, String requestUrl) async {
+        finish(() async {
+          await check(makeRequest(realmUrl, requestUrl))
+            .throws<StateError>();
+        }());
       }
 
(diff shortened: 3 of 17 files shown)


PIG208:
TODO: `TestGlobalStore`'s `doInsertAccount` implementation has a race condition for checking the invariants. Concurrently adding accounts allows you to bypass the check.

chrisbobbe:
This sounds helpful!! It looks like we have an issue for it: #731 🙂

PIG208:
This is stacked on top of #937.

lib/widgets/actions.dart
```diff
@@ -155,9 +157,9 @@ Future<bool> updateMessageFlagsStartingFromAnchor({
         // No messages were in the range of the request.
         // This should be impossible given that `foundNewest` was false
         // (and that our `numAfter` was positive.)
-        showErrorDialog(context: context,
+        unawaited(showErrorDialog(context: context,
```
gnprice, line 160: I just wrote down some thoughts on `showErrorDialog`'s API on a related PR:
https://github.com/zulip/zulip-flutter/pull/937#discussion_r1774377096

so adding a link to that here.

I think the many `unawaited` calls aroung `showErrorDialog` here get fairly noisy. That makes the diff bigger, which is undesirable but could be OK… but then I think it also leaves enough noise in the tree that it might start to undermine the effect of this lint. If we find ourselves scattering `unawaited` calls all over because there are some functions that just need it most times you call them, then the other more interesting `unawaited` calls don't stand out so much.

Probably a good fix for that is to have our own pair of `showFooDialog` wrappers provide an API that's more like the one I describe in that comment.

That would also be a better API for `showDialog` upstream to have. But it's a tough change to make, because the change in return type would break tons of downstream callers; I don't see a clear way to get to a great API given that constraint. So I wouldn't invest in trying to make the change upstream.

PIG208: Yeah, I like the `showSnackBar` API. `Future` is more primitive in the sense that it doesn't carry much contextual information. The wrapper class is a step up from simply using `typedef`, and offers good documentation discoverability.

PIG208:
This is stacked on top of #970 ([`dialog [nfc]: Wrap showErrorDialog's return value`](https://github.com/zulip/zulip-flutter/pull/970/files)).

test/api/core_test.dart
```diff
@@ -34,18 +34,18 @@ void main() {
...
-    checkRequest({},               '/api/v1/example/route?');
-    checkRequest({'x': 3},         '/api/v1/example/route?x=3');
-    checkRequest({'x': 3, 'y': 4}, '/api/v1/example/route?x=3&y=4');
-    checkRequest({'x': null},      '/api/v1/example/route?x=null');
-    checkRequest({'x': true},      '/api/v1/example/route?x=true');
-    checkRequest({'x': 'foo'},     '/api/v1/example/route?x=%22foo%22');
-    checkRequest({'x': [1, 2]},    '/api/v1/example/route?x=%5B1%2C2%5D');
-    checkRequest({'x': {'y': 1}},  '/api/v1/example/route?x=%7B%22y%22%3A1%7D');
-    checkRequest({'x': RawParameter('foo')},
+    await checkRequest(null,             '/api/v1/example/route');
```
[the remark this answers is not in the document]
PIG208: Skimming on mobile, this should also apply to c5160663231e1d627ac1ac98f4a853a79faa0573, right? I recall that `check(SomethingFuture)` has some special handling.

gnprice:
Other than `api/core_test.dart` as in the comment above (for which see #985), these changes all generally look like improvements to the code. And this is a PR that I expect would gather frequent conflicts as we merge other changes.

So I'd like to get this completed, even though as discussed on #731 we won't currently be able to enable the companion lint rule `discarded_futures`.

test/api/core_test.dart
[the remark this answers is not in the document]
gnprice: Yeah, the new `finish` helper in #985 can probably apply to some other places as well. No need to go sweep through using it in all the existing tests, but worth using to simplify new tests or tests changed in this PR.

chrisbobbe:
> So I'd like to get this completed

To help this along, I've just reviewed #970; over to you, Greg, for integration review on that.

PIG208:
Updated the PR by cherry picking #970 and #985 and cleaning up some commit messages.

chrisbobbe:
Thanks! LGTM, except the commit that enables `unawaited_futures` should mention #731. As discussed on that issue, it sounds like we won't be enabling that other rule `discarded_futures`, so maybe the commit should say `Fixes: `?

PIG208:
I think we should keep #731 open because there is still some chance that we can use that rule, but not in the near future. Mentioning it in the commit message seems fine.

lib/notifications/display.dart
```diff
@@ -354,10 +355,9 @@ class NotificationDisplayManager {
 
     assert(debugLog('  account: $account, narrow: $narrow'));
     // TODO(nav): Better interact with existing nav stack on notif open
-    navigator.push(MaterialAccountWidgetRoute<void>(accountId: account.id,
+    unawaited(navigator.push(MaterialAccountWidgetRoute<void>(accountId: account.id,
```
gnprice, line 358: nit in commit message: duplicate text

    The `Future` that `Navigator.push` returns resolves when a
    corresponding `Navigator.pop` is called.
    
    The future Navgiator.push returns completes when the there is a
    corresponding Navigator.pop call.  We don't use that as of now, so we
    use unawaited to explicitly discard it.

test/widgets/actions_test.dart
```diff
@@ -52,7 +53,7 @@ void main() {
         processedCount: 11, updatedCount: 3,
         firstProcessedId: null, lastProcessedId: null,
         foundOldest: true, foundNewest: true).toJson());
-      markNarrowAsRead(context, narrow);
+      unawaited(markNarrowAsRead(context, narrow));
       await tester.pump(Duration.zero);
```
gnprice, line 57: Huh, odd. I wonder why this test is using `tester.pump` — is that just a substitute for awaiting the future?

What if we replaced both these lines with this?
```dart
      await markNarrowAsRead(context, narrow);
```

(Similarly for the other `markNarrowAsRead` calls in this file.)

test/model/store_test.dart
```diff
@@ -153,12 +154,12 @@ void main() {
...
     test('reject changing id, realmUrl, or userId', () async {
       final globalStore = eg.globalStore(accounts: [eg.selfAccount]);
-      check(globalStore.updateAccount(eg.selfAccount.id, const AccountsCompanion(
-        id: Value(1234)))).throws();
-      check(globalStore.updateAccount(eg.selfAccount.id, AccountsCompanion(
-        realmUrl: Value(Uri.parse('https://other.example'))))).throws();
-      check(globalStore.updateAccount(eg.selfAccount.id, const AccountsCompanion(
-        userId: Value(1234)))).throws();
+      finish(check(globalStore.updateAccount(eg.selfAccount.id, const AccountsCompanion(
+        id: Value(1234)))).throws());
```
gnprice, line 158: For `throws`, I think the normal thing is to await the future:
```suggestion
      await check(globalStore.updateAccount(eg.selfAccount.id, const AccountsCompanion(
        id: Value(1234)))).throws();
```

From the [docs](https://github.com/dart-lang/test/tree/8e8a83607d90a7a6813fa378b2d1962a2fc0d44b/pkgs/checks#checking-asynchronous-expectations) linked in the commit message:
> Expectation extension methods checking asynchronous behavior return a `Future`. The future should typically be awaited within the test body, […]

analysis_options.yaml
```diff
@@ -25,3 +25,4 @@ linter:
     no_literal_bool_comparisons: true
     prefer_relative_imports: true
     unnecessary_statements: true
+    unawaited_futures: true
```
gnprice, line 28: The commit message steps into a classic GitHub pitfall :slightly_smiling_face: :

    Partially fixes: #731

GitHub doesn't understand the "partially", so that will close the issue.

If you don't mean "fixes", then the trick is to put some other word between it and the issue number. For example:
```
Fixes-partly: #731
```

gnprice: nit in commit message: the convention for email-header-style lines at the end of a commit message is that the words before the colon are joined by hyphens `-`, not spaces ` `

gnprice (commented):
Thanks @PIG208 for cleaning these up, and thanks @chrisbobbe for the previous review! Generally this all looks good; small comments below.

test/widgets/actions_test.dart
PIG208: It's stuck forever, right before the `connection.post` call to "messages/flags/narrow". It's puzzling that it happens to a `FakeApiConnection` though.

PIG208: I guess this is because we call inside `testWidgets`. This issue is not specific to `markNarrowAsRead`. For example, a `store.sendMessage` call stucks as well. The `FakeApiClient` returns a `Future.delayed(response.delay, computation);` that does not complete.

PIG208:
Thanks for the review! Updated the PR and fixed some more errors that appeared after rebasing.

test/widgets/actions_test.dart
gnprice: Ah, I see. So basically it's running into the problem that `awaitFakeAsync` solves.

It feels odd for these tests to not be awaiting that future — it means they're not ensuring that the requests have actually completed, and that the function has done the work it's going to do when it gets the responses, before they complete.

I'll send a quick separate PR that fixes these.

gnprice: #993

gnprice:
Thanks for the revision! All looks good; merging, with one NFC commit on top:
1aef86814 api test [nfc]: Simplify some unawaited checks

Issue #166, opened 2023-06-08
polls: Vote and unvote on options


gnprice:
As part of poll support (#164), we should support voting for options, and removing one's vote for an option.

This will come after read-only support, #165.

PR #939, opened 2024-09-12
Support poll vote/unvote

PIG208:
~~This is stacked on top of #912.~~

Fixes: #166

### DIFF
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -82,9 +82,12 @@ class Submessage {
 /// The only type of submessage that actually exists in Zulip (as of 2024,
 /// and since this "submessages" subsystem was created in 2017–2018)
 /// is [SubmessageType.widget].
+@JsonEnum(fieldRename: FieldRename.snake, alwaysCreate: true)
 enum SubmessageType {
   widget,
-  unknown,
+  unknown;
+
+  String toJson() => _$SubmessageTypeEnumMap[this]!;
 }
 
 /// The data encoded in a submessage at [Submessage.content].
--- a/lib/api/model/submessage.dart
+++ b/lib/api/model/submessage.dart
@@ -462,7 +462,7 @@ class Poll extends ChangeNotifier {
 
     final key = PollEventSubmessage.optionKey(senderId: senderId, idx: idx);
     assert(!_options.containsKey(key));
-    _options[key] = PollOption(text: option);
+    _options[key] = PollOption(key: key, text: option);
     _existingOptionTexts.add(option);
   }
 
@@ -479,8 +479,9 @@ class Poll extends ChangeNotifier {
 }
 
 class PollOption {
-  PollOption({required this.text});
+  PollOption({required this.key, required this.text});
 
+  final PollOptionKey key;
   final String text;
   final Set<int> voters = {};
 
--- a/lib/api/route/submessage.dart
+++ b/lib/api/route/submessage.dart
@@ -1,0 +1,16 @@
+import '../core.dart';
+import '../model/submessage.dart';
+
+/// https://zulip.readthedocs.io/en/latest/subsystems/widgets.html#polls-todo-lists-and-games
+/// https://github.com/zulip/zulip-mobile/blob/fc23edd67a5ec7f32c7c5f6cd81893b94dc043a2/src/api/submessages/sendSubmessage.js
+Future<void> sendSubmessage(ApiConnection connection, {
+  required int messageId,
+  required SubmessageType submessageType,
+  required SubmessageData content,
+}) {
+  return connection.post('sendSubmessage', (_) {}, 'submessage', {
+    'message_id': messageId,
+    'msg_type': RawParameter(submessageType.toJson()),
+    'content': content,
+  });
+}
--- a/lib/widgets/poll.dart
+++ b/lib/widgets/poll.dart
@@ -68,15 +68,15 @@ class _PollWidgetState extends State<PollWidget> {
           store.users[userId]?.fullName ?? zulipLocalizations.unknownUserName)
         .join(', ');
 
-      return Padding(
-        padding: const EdgeInsets.only(bottom: 5),
-        child: Row(
-          spacing: 5,
-          crossAxisAlignment: CrossAxisAlignment.baseline,
-          textBaseline: localizedTextBaseline(context),
-          children: [
-            ConstrainedBox(
-              constraints: const BoxConstraints(minWidth: 25, minHeight: 25),
+      return Row(
+        crossAxisAlignment: CrossAxisAlignment.baseline,
+        textBaseline: localizedTextBaseline(context),
+        children: [
+          ConstrainedBox(
+            constraints: const BoxConstraints(
+              minWidth: 25 + 5, minHeight: 25 + 5),
+            child: Padding(
+              padding: const EdgeInsetsDirectional.only(bottom: 5, end: 5),
               child: Container(
                 // Inner padding preserves whitespace even when the text's
                 // width approaches the button's min-width (e.g. because
@@ -89,8 +89,14 @@ class _PollWidgetState extends State<PollWidget> {
                 child: Center(
                   child: Text(option.voters.length.toString(),
                     style: textStyleBold.copyWith(
-                      color: theme.colorPollVoteCountText, fontSize: 13))))),
-            Expanded(
+                      color: theme.colorPollVoteCountText, fontSize: 13)))))),
+          Expanded(
+            child: Padding(
+              // This and the bottom padding on the vote count box both extend
+              // the row by the same extent. This ensures that there still will
+              // be consistent spacing between rows when the text takes more
+              // vertical space than the vote count box.
+              padding: const EdgeInsets.only(bottom: 5),
               child: Wrap(
                 spacing: 5,
                 children: [
@@ -98,8 +104,8 @@ class _PollWidgetState extends State<PollWidget> {
                   if (option.voters.isNotEmpty)
                     // TODO(i18n): Localize parenthesis characters.
                     Text('($voterNames)', style: textStyleVoterNames),
-                ])),
-          ]));
+                ]))),
+        ]);
     }
 
     return Column(
--- a/lib/widgets/poll.dart
+++ b/lib/widgets/poll.dart
@@ -46,6 +46,8 @@ class _PollWidgetState extends State<PollWidget> {
 
   @override
   Widget build(BuildContext context) {
+    const verticalPadding = 2.5;
+
     final zulipLocalizations = ZulipLocalizations.of(context);
     final theme = ContentTheme.of(context);
     final store = PerAccountStoreWidget.of(context);
@@ -73,10 +75,10 @@ class _PollWidgetState extends State<PollWidget> {
         textBaseline: localizedTextBaseline(context),
         children: [
           ConstrainedBox(
-            constraints: const BoxConstraints(
-              minWidth: 25 + 5, minHeight: 25 + 5),
+            constraints: const BoxConstraints(minWidth: 44, minHeight: 44),
             child: Padding(
-              padding: const EdgeInsetsDirectional.only(bottom: 5, end: 5),
+              padding: const EdgeInsetsDirectional.only(
+                end: 5, top: verticalPadding, bottom: verticalPadding),
               child: Container(
                 // Inner padding preserves whitespace even when the text's
                 // width approaches the button's min-width (e.g. because
@@ -89,14 +91,14 @@ class _PollWidgetState extends State<PollWidget> {
                 child: Center(
                   child: Text(option.voters.length.toString(),
                     style: textStyleBold.copyWith(
-                      color: theme.colorPollVoteCountText, fontSize: 13)))))),
+                      color: theme.colorPollVoteCountText, fontSize: 20)))))),
           Expanded(
             child: Padding(
-              // This and the bottom padding on the vote count box both extend
-              // the row by the same extent. This ensures that there still will
-              // be consistent spacing between rows when the text takes more
+              // This and the padding on the vote count box both extend the row
+              // by the same extent. This ensures that there still will be
+              // consistent spacing between rows when the text takes more
               // vertical space than the vote count box.
-              padding: const EdgeInsets.only(bottom: 5),
+              padding: const EdgeInsets.symmetric(vertical: verticalPadding),
               child: Wrap(
                 spacing: 5,
                 children: [
@@ -111,15 +113,19 @@ class _PollWidgetState extends State<PollWidget> {
     return Column(
       crossAxisAlignment: CrossAxisAlignment.start,
       children: [
-        Padding(padding: const EdgeInsets.only(bottom: 6), child: question),
+        question,
+        // `verticalPadding` out of 6px comes from the first option row.
+        const SizedBox(height: 6 - verticalPadding),
         if (widget.poll.options.isEmpty)
           Padding(
-            // This is consistent with the option rows' bottom padding.
-            padding: const EdgeInsets.only(bottom: 5),
+            // This is consistent with the option rows' padding.
+            padding: const EdgeInsets.symmetric(vertical: verticalPadding),
             child: Text(zulipLocalizations.pollWidgetOptionsMissing,
               style: textStyleVoterNames.copyWith(fontStyle: FontStyle.italic))),
         for (final option in widget.poll.options)
           buildOptionItem(option),
+        // `verticalPadding` out of 5px comes from the last option row.
+        const SizedBox(height: 5 - verticalPadding),
       ]);
   }
 }
--- a/lib/widgets/poll.dart
+++ b/lib/widgets/poll.dart
@@ -85,7 +85,6 @@ class _PollWidgetState extends State<PollWidget> {
                   borderRadius: BorderRadius.circular(3)),
                 child: Center(
                   child: Text(option.voters.length.toString(),
-                    textAlign: TextAlign.center,
                     style: textStyleBold.copyWith(
                       color: theme.colorPollVoteCountText, fontSize: 13))))),
             Expanded(
--- a/lib/widgets/poll.dart
+++ b/lib/widgets/poll.dart
@@ -78,6 +78,9 @@ class _PollWidgetState extends State<PollWidget> {
             ConstrainedBox(
               constraints: const BoxConstraints(minWidth: 25, minHeight: 25),
               child: Container(
+                // Inner padding preserves whitespace even when the text's
+                // width approaches the button's min-width (e.g. because
+                // there are more than three digits).
                 padding: const EdgeInsets.symmetric(horizontal: 4),
                 decoration: BoxDecoration(
                   color: theme.colorPollVoteCountBackground,
--- a/lib/widgets/poll.dart
+++ b/lib/widgets/poll.dart
@@ -113,8 +113,11 @@ class _PollWidgetState extends State<PollWidget> {
       children: [
         Padding(padding: const EdgeInsets.only(bottom: 6), child: question),
         if (widget.poll.options.isEmpty)
-          Text(zulipLocalizations.pollWidgetOptionsMissing,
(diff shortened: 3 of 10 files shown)


PIG208:
<details>
<summary>Preview</summary>





</details>

PIG208:
<details>

<summary>Preview</summary>
| | |
| - | - |
| Screenshot from 2024-10-07 19-55-57 | Screenshot from 2024-10-07 19-57-28 |

</details>

lib/api/route/submessage.dart
```diff
@@ -0,0 +1,19 @@
...
+  required int messageId,
+  required SubmessageData content,
+}) {
+  return connection.post('sendSubmessage', (_) {}, 'submessage', {
+    'message_id': messageId,
+    // 'widget' applies to submessages for polls and todos.
+    // There is also 'zform' for trivia quiz bot, but we currently do not have
+    // plan to support that.
+    'msg_type': RawParameter('widget'),
+    'content': RawParameter(jsonEncode(content)),
```
chrisbobbe, line 17: Would we get the same result from `'content': content` as we do from this, and if so, would it be preferable because it looks simpler?

test/model/message_test.dart
```diff
@@ -628,13 +628,13 @@ void main() {
           eg.submessage(senderId: effectiveMessageSender.userId,
             content: eg.pollWidgetData(
               question: question ?? 'example question',
-              options: (options != null)
-                ? options.map((e) => e.text).toList()
+              options: (optionVoterIdsPairs != null)
+                ? optionVoterIdsPairs.map((e) => e.$1).toList()
```
chrisbobbe, line 632: Oh, odd, I guess `$1` refers to the first element of the tuple? So it's not zero-indexed.

lib/widgets/poll.dart
```diff
@@ -68,15 +68,14 @@ class _PollWidgetState extends State<PollWidget> {
...
-      return Padding(
-        padding: const EdgeInsets.only(bottom: 5),
-        child: Row(
-          spacing: 5,
-          crossAxisAlignment: CrossAxisAlignment.baseline,
-          textBaseline: localizedTextBaseline(context),
-          children: [
-            ConstrainedBox(
-              constraints: const BoxConstraints(minWidth: 25).tighten(height: 25),
+      return Row(
```
chrisbobbe, line 88:
```
poll [nfc]: Move vote count paddings to a box
```

> This does not visually change the paddings.

This doesn't seem to be true. Here's before/after this change, with message https://chat.zulip.org#narrow/stream/7-test-here/topic/new.20poll.20test/near/1956979 :

| Before | After |
| --- | --- |
| image | image |

When I open the two images in different browser tabs and switch between them, I see a layout change. Do you reproduce?

lib/widgets/poll.dart
```diff
@@ -73,10 +74,14 @@ class _PollWidgetState extends State<PollWidget> {
...
-            constraints: const BoxConstraints(minWidth: 25 + 5).tighten(height: 25 + 5),
+            constraints: const BoxConstraints(
+              minWidth: 39 + 5).tighten(height: 39 + verticalPadding * 2),
             child: Padding(
-              padding: const EdgeInsetsDirectional.only(bottom: 5, end: 5),
+              padding: const EdgeInsetsDirectional.only(
+                end: 5, top: verticalPadding, bottom: verticalPadding),
               child: Container(
+                // This is only in effect
+                // when the vote count has more than 2 digits.
```
chrisbobbe: I'm a little suspicious of a comment like this :) because I don't know if it's true as written for all values of the user's system bold-text and text-size settings. If this needs a comment, maybe it could be something like:

```dart
                  // Inner padding preserves whitespace even when the text's
                  // width approaches the button's min-width (e.g. because 
                  // there are three digits).
```

chrisbobbe (commented):
Thanks! Comments below.

Also, I noticed that the web app has a "pressed" style for the button. Do we want to either do that, or make an issue and TODO for it?

I guess my cursor isn't captured in this screenshot :) but you can imagine it there:

image

alya:
Screenshots and CZO discussion look good to me!

lib/api/route/submessage.dart
PIG208: The case of `content` is a bit peculiar, because the API expects a plain string instead of an object, while the string is expected to be some JSON encoded value.

I think it would be good to mention this in comments.

lib/widgets/poll.dart
PIG208: Hmm, I can see that in the screenshots, but not on my Android device.

PIG208:
The "pressed" UX might fit better in a separate PR. I guess we need to have a way to mark some data as "dirty" until an event comes and update it (but there could also be cases where we expect multiple events).

I also noticed that my previous screenshots have a non-default sized text. The 22px font size for the vote count text is actually inadequate. Updated that to 20px with new screenshots:

<details>
<summary>Updated screenshots</summary>


| | | |
| - | - | - |
|  |  |  |


</details>

chrisbobbe:
> The "pressed" UX might fit better in a separate PR. I guess we need to have a way to mark some data as "dirty" until an event comes and update it (but there could also be cases where we expect multiple events).

I was just thinking about highlighting the button when the user's finger is on it; I wasn't thinking about the state between the tap and the event arriving.

lib/widgets/poll.dart
chrisbobbe: When the `bottom: 5` padding wraps the whole row (before this commit), instead of just the vote-count rectangle, there will be 5px whitespace below the text (option name + voters' names) even when the text takes more vertical space than the vote-count rectangle.

I think you can reproduce if you make an option with a name so long that it takes two or three lines on your device.

gnprice:
> I think it might be the more complicated thing where it highlights the button between the tap and the event arrival.

That more complicated thing sounds like when it has an effect, it'd be liable to make it feel like the button was stuck. (Though I haven't tried to play with this aspect of web's UX.)

I'd be glad to have immediate feedback while the user's finger is there, and perhaps animating back to the resting state over a short fixed duration thereafter like 100ms, while not trying to have any longer-lived "dirty" state. This would make it similar to the existing behavior on the mark-as-read button.

chrisbobbe:
> That more complicated thing sounds like when it has an effect, it'd be liable to make it feel like the button was stuck. (Though I haven't tried to play with this aspect of web's UX.)

I agree; let's not do this. I think Vlad likes a scale animation for the "pressed" state, like we did for the mark-as-read button in 791b703c.

Then I think https://github.com/zulip/zulip-flutter/pull/939#discussion_r1794340316 is still pending, about a layout-change commit that should be unmarked NFC. One consequence of the change is that it removes any spacing between two options when the first one's text runs onto multiple lines:

| Before | After |
| --- | --- |
| image | image |

I don't love that consequence, but it's helped in a later commit that puts 2.5px spacing above the button.

chrisbobbe:
Then also, as discussed in the office just now: in the place where we use a fixed height, we'll want a min-height instead, so that it doesn't break with a large system text size setting:

| Current revision | Changed to min-height |
| --- | --- |
| image | image |

PIG208:
Updated the PR. The button feedback will be done in a separate PR. The current plan is to extract a common animated button widget.

This should handle larger font sizes now, as well as consistent spacing; the padding box commit should be truly NFC now.

chrisbobbe:
Great! LGTM, marking for Greg's review.

lib/widgets/poll.dart
```diff
@@ -69,46 +86,70 @@ class _PollWidgetState extends State<PollWidget> {
...
+            child: ConstrainedBox(
+              constraints: const BoxConstraints(
+                minWidth: 39 + 5, minHeight: 39 + verticalPadding * 2),
+              child: Padding(
+                // For accessibility, the touch target is padded to be larger
+                // than the vote count box.  Still, we avoid padding at the
+                // start because we want to align all the poll options to the
+                // surrounding messages.
+                padding: const EdgeInsetsDirectional.only(
+                  end: 5, top: verticalPadding, bottom: verticalPadding),
```
gnprice, line 106: nit:

    poll [nfc]: Move vote count paddings to a box

I'm not sure what "to a box" means — `Padding` is always a box, right?

Could say:

> poll [nfc]: Move vote count paddings inside children of Row

and I think that accurately conveys what this commit is doing.

gnprice: other nit in same commit message:

    specified on the outer (now removed) Padding.  The Container is
    expanded because it now also contains the paddings.

I think you mean the ConstrainedBox, not Container.

lib/widgets/poll.dart
```diff
@@ -69,46 +86,70 @@ class _PollWidgetState extends State<PollWidget> {
...
+                      textAlign: TextAlign.center,
+                      style: textStyleBold.copyWith(
+                        color: theme.colorPollVoteCountText, fontSize: 20))))))),
+          Expanded(
+            child: Padding(
+              // When the bottom of the text reaches farther than the vote count
+              // box's padded bottom edge, this padding helps ensure that we
+              // still maintain a consistent spacing of `verticalPadding * 2`
+              // logical pixels between option rows.
+              padding: const EdgeInsets.only(bottom: verticalPadding),
```
gnprice, line 127: The comment and code don't seem to agree here. Do you want `EdgeInsets.symmetric(vertical: verticalPadding)`?

lib/api/model/submessage.dart
```diff
@@ -404,6 +404,9 @@ class Poll extends ChangeNotifier {
   final Set<String> _existingOptionTexts = {};
   final Map<PollOptionKey, PollOption> _options = {};
 
+  bool hasUserVotedFor({required int userId, required PollOptionKey key}) =>
+    _options.containsKey(key) && _options[key]!.voters.contains(userId);
```
gnprice, line 408:
```suggestion
    _options[key]?.voters.contains(userId) ?? false;
```

That avoids doing the lookup twice.

lib/widgets/poll.dart
```diff
@@ -69,46 +86,70 @@ class _PollWidgetState extends State<PollWidget> {
...
-                    style: textStyleBold.copyWith(
-                      color: theme.colorPollVoteCountText, fontSize: 13))))),
-            Expanded(
+      return Row(
+        crossAxisAlignment: CrossAxisAlignment.baseline,
+        textBaseline: localizedTextBaseline(context),
+        children: [
+          GestureDetector(
+            // TODO: Implement feedback when the user taps the button
+            onTap: () => _toggleVote(option),
```
gnprice, line 95: nit: squash the TODO-comment commit into the commit that adds the code it's commenting on

lib/widgets/poll.dart
```diff
@@ -44,21 +48,34 @@ class _PollWidgetState extends State<PollWidget> {
     });
   }
 
+  void _toggleVote(PollOption option) async {
+    final store = PerAccountStoreWidget.of(context);
+    // The poll data in store might be obselete before we get the event
+    // that updates it.  This is fine because the result will be consistent
+    // eventually, regardless of the possible duplicate requests.
+    final op = widget.poll.hasUserVotedFor(userId: store.selfUserId, key: option.key)
+      ? PollVoteOp.remove
```
gnprice, line 57: I don't quite follow what this comment is saying. What's the scenario where it's obsolete, and when?

If this is intended as reasoning for why we're determining `op` based on the data in the store, rather than something else like telling the server to toggle from whatever state the server has:

I think the real point there is that when the UI shows the user a state X and the user says to change it to state Y, the thing we want to tell the server is to please change in the direction X → Y. We basically never want to translate that into something like "please flip between state X or Y, whichever one you currently have it in". (The only exception would be if the nature of "state X" and "state Y" is somehow such that that translation would be a natural interpretation of what the user actually intends.)

So, here, the UI we're showing either has the user as having already voted for the option, or as not having done so. If the UI says the user has already voted and the user hits the button, that means to remove their vote, unconditionally. Even if the server currently thinks they haven't voted, what the user asked for is to remove their vote, not add it — after all, they were looking at a UI that said they had voted.

One consequence of that general principle is that the canonical thing to do here wouldn't be to go look up `option.key` in the store again and find what it says. I think in this revision that will end up having the same behavior we want… but the canonical way to get that behavior is to look directly at the data structure that was used for rendering the UI. Here, `option.voters` is what controlled the list of names we showed. So how about we just inspect `option.voters` directly?

gnprice (commented):
Thanks @PIG208 for building this, and @chrisbobbe for the previous reviews!

Generally this looks good. Comments below.

lib/widgets/poll.dart
PIG208: The other `verticalPadding` comes from the top of the next vote count box (in the next row); combined we get `verticalPadding * 2`. This was added to fix https://github.com/zulip/zulip-flutter/pull/939#issuecomment-2406034465, but I think the comment can articulate that purpose better.

lib/widgets/poll.dart
PIG208: > I think the real point there is that when the UI shows the user a state X and the user says to change it to state Y, the thing we want to tell the server is to please change in the direction X → Y. We basically never want to translate that into something like "please flip between state X or Y, whichever one you currently have it in". 

Yeah exactly that. The user couldn't instantly toggle the vote because it takes time to send an update to the server and then receive the event for it — that's not really a bad thing.

In the legacy mobile app, with locally echoing, you can have a sequence of states like this:
- User taps the button to unvote;
- User taps the button again to vote;
- The submessage event for the unvote arrives;
- The submessage event for the vote arrives;

which result in visually four toggles. This is probably because we don't have a way to deduplicate the updates. The current behavior of the this PR revision doesn't have that problem.

lib/widgets/poll.dart
```diff
@@ -69,46 +86,70 @@ class _PollWidgetState extends State<PollWidget> {
...
+                  // width approaches the button's min-width (e.g. because
+                  // there are more than three digits).
+                  padding: const EdgeInsets.symmetric(horizontal: 4),
+                  decoration: BoxDecoration(
+                    color: theme.colorPollVoteCountBackground,
+                    border: Border.all(color: theme.colorPollVoteCountBorder),
+                    borderRadius: BorderRadius.circular(3)),
+                  child: Center(
+                    child: Text(option.voters.length.toString(),
+                      textAlign: TextAlign.center,
```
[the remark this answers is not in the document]
PIG208: Tested that I could see none. I checked my reflog and found that they have coexisted in some earliest draft of the read-only `PollWidget`. It's likely that they were just redundant before and lived.

PIG208:
This PR has been updated. Thanks for the review!

lib/widgets/poll.dart
```diff
@@ -69,46 +84,62 @@ class _PollWidgetState extends State<PollWidget> {
           store.users[userId]?.fullName ?? zulipLocalizations.unknownUserName)
         .join(', ');
 
-      return Padding(
-        padding: const EdgeInsets.only(bottom: 5),
```
gnprice, line 73: poll: Move vote count paddings inside children of Row

It looks like in this revision, this commit actually moves the padding only inside the first child, and drops the padding from the second one.

    This preserves most of the visuals. Except that when the vote count
    box is shorter than the voter names text, the padding between rows
    is gone.

That seems undesirable. What if we include padding on both children, like in the previous revision?

I think this change in this revision may have been in response to this thread:
https://github.com/zulip/zulip-flutter/pull/939#discussion_r1809681899
but I don't really understand the motivation for removing the padding from the voter names.

gnprice (commented):
Thanks for the revision! Generally this looks good; one question below.

Since this revision changed the layout, please also post updated screenshots.

lib/widgets/poll.dart
gnprice: (Thread continues below: https://github.com/zulip/zulip-flutter/pull/939#discussion_r1815952172 )

gnprice (commented):
Great, thanks for changing that back.

Small comments below — most of them about a follow-up in that same area.

lib/widgets/poll.dart
```diff
@@ -69,46 +84,68 @@ class _PollWidgetState extends State<PollWidget> {
...
     }
 
     return Column(
       crossAxisAlignment: CrossAxisAlignment.start,
       children: [
-        Padding(padding: const EdgeInsets.only(bottom: 6), child: question),
+        Padding(
+          // `verticalPadding` out of 6 logical pixels come from the first
+          // option row.
+          padding: const EdgeInsets.only(bottom: 6 - verticalPadding),
```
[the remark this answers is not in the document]
PIG208: > Ah, I guess if the text is long, then there's still top padding because CrossAxisAlignment.baseline means the option text starts at the same level as the count and expands down from there, and so the button still extends further upward from the baseline than the text does.

Yeah, exactly. I've made them symmetric in the new revision.

PIG208:
Pushed a new revision. Thanks for the review!

- Now we use vertically symmetric padding for all the option items, and limit the one-sided paddings.
- This also fixed an issue with the "This poll has no options yet" text where the bottom padding is missing.
- I added `5 - verticalPadding` to the entire poll widget so that there is still 5px space after it.

| before | after |
| - | - |
|  |  |

lib/widgets/poll.dart
```diff
@@ -69,48 +84,74 @@ class _PollWidgetState extends State<PollWidget> {
...
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Padding(
+            // `verticalPadding` out of 6 logical pixels
+            // come from the first option row.
+            padding: const EdgeInsets.only(bottom: 6 - verticalPadding),
+            child: question),
+          if (widget.poll.options.isEmpty)
+            Padding(
+              // This is consistent with the option rows' padding.
```
gnprice, line 149: This change makes this commit no longer NFC, right?
ceca4cec7 poll [nfc]: Move vote count paddings inside children of Row

It's also not part of what that summary describes. So I think it should just become a separate commit, fixing that layout bug you noted.

lib/widgets/poll.dart
```diff
@@ -69,48 +84,74 @@ class _PollWidgetState extends State<PollWidget> {
...
+      // come from the last option row.
+      padding: const EdgeInsets.only(bottom: 5 - verticalPadding),
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Padding(
+            // `verticalPadding` out of 6 logical pixels
+            // come from the first option row.
+            padding: const EdgeInsets.only(bottom: 6 - verticalPadding),
+            child: question),
```
gnprice, line 146: nit: this is a bit simpler to express with a SizedBox spacer:
```suggestion
          question,
          // `verticalPadding` out of 6px come from the first option row.
          const SizedBox(height: 6 - verticalPadding),
```

lib/widgets/poll.dart
```diff
@@ -69,48 +84,74 @@ class _PollWidgetState extends State<PollWidget> {
...
-        if (widget.poll.options.isEmpty)
-          Text(zulipLocalizations.pollWidgetOptionsMissing,
-            style: textStyleVoterNames.copyWith(fontStyle: FontStyle.italic)),
-        for (final option in widget.poll.options)
-          buildOptionItem(option),
-      ]);
+    return Padding(
+      // `verticalPadding` out of 5 logical pixels
+      // come from the last option row.
+      padding: const EdgeInsets.only(bottom: 5 - verticalPadding),
```
gnprice, line 138: Then similarly this padding can be expressed as a spacer at the end of the column's children. I think that makes it significantly clearer what's going on, because it keeps things in a consistent order.

gnprice (commented):
Thanks for the revision! Glad to have that layout issue fixed — and I think the layout logic all makes sense to me now. Just a few style comments below.

PIG208:
Thanks! The PR has been updated.

gnprice:
Thanks! Looks good now — merging, with two tweaks:

    -    poll: Enlarge and vetically center vote count box
    +    poll: Enlarge and vertically center vote count box


    -+        // `verticalPadding` out of 6px come from the first option row.
    ++        // `verticalPadding` out of 6px comes from the first option row.

(I know the latter comes from my suggestion above; rereading, I'd incompletely edited it from the previous version.)

PR #940, opened 2024-09-13
test: Use random email in eg.user

PIG208:
This specifically fixes a test data realism issue of the test "find account among several" in
`test/notifications/display_test.dart` due to duplicate email addresses within a realm.  This also useful in general in cases when the email addresses do not matter and should be kept boring.

The generation uses an increasing sequence to make collision unlikely (still possible with caller specified values).

This is taken from #934 and similar to #932.

### DIFF
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -70,6 +70,10 @@ GetServerSettingsResult serverSettings({
 int _nextUserId() => (_lastUserId += 1 + Random().nextInt(100));
 int _lastUserId = 1000;
 
+/// A random email address, different from previously generated ones.
+String _nextEmail() => 'mail${_lastEmailSuffix += Random().nextInt(1000)}@example.com';
+int _lastEmailSuffix = 1000;
+
 /// Construct an example user.
 ///
 /// If user ID `userId` is not given, it will be generated from a random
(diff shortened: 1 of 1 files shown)


test/example_data.dart
```diff
@@ -70,13 +70,21 @@ GetServerSettingsResult serverSettings({
 int _nextUserId() => (_lastUserId += 1 + Random().nextInt(100));
 int _lastUserId = 1000;
 
+/// A random email address, unique from previously generated ones.
```
chrisbobbe, line 73:
```suggestion
/// A random email address, different from previously generated ones.
```

chrisbobbe (approved):
Thanks! LGTM with one suggestion below; marking for Greg's review.

gnprice:
Thanks @PIG208 for the fix, and @chrisbobbe for the previous review!

Looks good; merging, with one docs tweak:
```diff
-/// If `deliveryEmail` is not given, it will be generated from a random
-/// but unique sequence of strings.
+/// If `deliveryEmail` is not given, it will be generated from a
+/// random sequence of distinct strings.
```

For me saying the *sequence* is unique would mean that it's distinct from other sequences; instead we mean that all its elements are unique, i.e. distinct from each other.

PR #947, opened 2024-09-17
check: Use --text when potentially diffing generated files

PIG208:
The diffs generated from running `tools/check build_runner` were not detected.  See comment: https://github.com/zulip/zulip-flutter/pull/904#discussion_r1762083210

We normally exclude diffs from generated files through our `.gitattributes` configurations.  This happens to `git --quiet -- [<path>...]` too.  -a/--text allows us to include those files regardless.

### DIFF
--- a/tools/check
+++ b/tools/check
@@ -189,7 +189,8 @@ files_check() {
 check_no_changes() {
     local change_description="$1"; shift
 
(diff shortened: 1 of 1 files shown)


chrisbobbe:
LGTM, marking for Greg's review.

Issue #463, opened 2023-12-26
Support logging out / forgetting about an account


gnprice:
We should add a bit of UI to let the user tell the client to forget about one of the accounts it's logged into.

This can be helpful for decluttering the list of accounts, after logging into some test account or some other account one doesn't expect to use again.

In zulip-mobile we have this in the form of a red trash-can icon on each of the accounts in the list. I don't love that UI and we can probably do something better.

gnprice:
Requested in chat: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20app.20.2B.20SAML.20auth/near/1775672

PR #948, opened 2024-09-17
login: Support logging out of an account

chrisbobbe:
Screenshots coming soon. :)

Fixes: #463

### DIFF
--- a/lib/api/route/notifications.dart
+++ b/lib/api/route/notifications.dart
@@ -1,4 +1,3 @@
-
 import '../core.dart';
 
 /// https://zulip.com/api/add-fcm-token
@@ -6,6 +5,15 @@ Future<void> addFcmToken(ApiConnection connection, {
   required String token,
 }) {
   return connection.post('addFcmToken', (_) {}, 'users/me/android_gcm_reg_id', {
+    'token': RawParameter(token),
+  });
+}
+
+/// https://zulip.com/api/remove-fcm-token
+Future<void> removeFcmToken(ApiConnection connection, {
+  required String token,
+}) {
+  return connection.delete('removeFcmToken', (_) {}, 'users/me/android_gcm_reg_id', {
     'token': RawParameter(token),
   });
 }
@@ -18,5 +26,14 @@ Future<void> addApnsToken(ApiConnection connection, {
   return connection.post('addApnsToken', (_) {}, 'users/me/apns_device_token', {
     'token': RawParameter(token),
     'appid': RawParameter(appid),
+  });
+}
+
+/// https://zulip.com/api/remove-apns-token
+Future<void> removeApnsToken(ApiConnection connection, {
+  required String token,
+}) {
+  return connection.delete('removeApnsToken', (_) {}, 'users/me/apns_device_token', {
+    'token': RawParameter(token),
   });
 }
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -144,7 +144,14 @@ abstract class GlobalStore extends ChangeNotifier {
   /// This method should be called only by the implementation of [perAccount].
   /// Other callers interested in per-account data should use [perAccount]
   /// and/or [perAccountSync].
-  Future<PerAccountStore> loadPerAccount(int accountId);
+  Future<PerAccountStore> loadPerAccount(int accountId) {
+    return doLoadPerAccount(accountId);
+  }
+
+  /// Load per-account data for the given account, unconditionally.
+  ///
+  /// This method should be called only by [loadPerAccount].
+  Future<PerAccountStore> doLoadPerAccount(int accountId);
 
   // Just the Iterables, not the actual Map, to avoid clients mutating the map.
   // Mutations should go through the setters/mutators below.
@@ -682,7 +689,7 @@ class LiveGlobalStore extends GlobalStore {
   final AppDatabase _db;
 
   @override
-  Future<PerAccountStore> loadPerAccount(int accountId) async {
+  Future<PerAccountStore> doLoadPerAccount(int accountId) async {
     final updateMachine = await UpdateMachine.load(this, accountId);
     return updateMachine.store;
   }
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -310,6 +310,14 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
   final GlobalStore _globalStore;
   final ApiConnection connection; // TODO(#135): update zulipFeatureLevel with events
 
+  UpdateMachine? get updateMachine => _updateMachine;
+  UpdateMachine? _updateMachine;
+  set updateMachine(UpdateMachine? value) {
+    assert(_updateMachine == null);
+    assert(value != null);
+    _updateMachine = value;
+  }
+
   bool get isLoading => _isLoading;
   bool _isLoading = false;
   @visibleForTesting
@@ -452,6 +460,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
     unreads.dispose();
     _messages.dispose();
     typingStatus.dispose();
+    updateMachine?.dispose();
     super.dispose();
   }
 
@@ -729,7 +738,9 @@ class UpdateMachine {
          // case of unauthenticated access to a web-public realm.  We authenticated.
          throw Exception("bad initial snapshot: missing queueId");
        })(),
-       lastEventId = initialSnapshot.lastEventId;
+       lastEventId = initialSnapshot.lastEventId {
+    store.updateMachine = this;
+  }
 
   /// Load the user's data from the server, and start an event queue going.
   ///
@@ -874,8 +885,8 @@ class UpdateMachine {
         switch (e) {
           case ZulipApiException(code: 'BAD_EVENT_QUEUE_ID'):
             assert(debugLog('Lost event queue for $store.  Replacing…'));
+            // This disposes the store, which disposes this update machine.
             await store._globalStore._reloadPerAccount(store.accountId);
-            dispose();
             debugLog('… Event queue replaced.');
             return;
 
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -454,17 +454,23 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
     autocompleteViewManager.reassemble();
   }
 
+  bool _disposed = false;
+
   @override
   void dispose() {
+    assert(!_disposed);
     recentDmConversationsView.dispose();
     unreads.dispose();
     _messages.dispose();
     typingStatus.dispose();
     updateMachine?.dispose();
+    _disposed = true;
     super.dispose();
   }
 
   Future<void> handleEvent(Event event) async {
+    assert(!_disposed);
+
     switch (event) {
       case HeartbeatEvent():
         assert(debugLog("server event: heartbeat"));
@@ -606,6 +612,8 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
   }
 
   Future<void> sendMessage({required MessageDestination destination, required String content}) {
+    assert(!_disposed);
+
     // TODO implement outbox; see design at
     //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3881.20Sending.20outbox.20messages.20is.20fraught.20with.20issues/near/1405739
     return _apiSendMessage(connection,
@@ -790,6 +798,8 @@ class UpdateMachine {
   final String queueId;
   int lastEventId;
 
+  bool _disposed = false;
+
   static Future<InitialSnapshot> _registerQueueWithRetry(
       ApiConnection connection) async {
     BackoffMachine? backoffMachine;
@@ -868,8 +878,11 @@ class UpdateMachine {
     BackoffMachine? backoffMachine;
 
     while (true) {
+      if (_disposed) return;
+
       if (_debugLoopSignal != null) {
         await _debugLoopSignal!.future;
+        if (_disposed) return;
         assert(() {
           _debugLoopSignal = Completer();
           return true;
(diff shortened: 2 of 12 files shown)


chrisbobbe:
imageimage

chrisbobbe:
This implements the logic where per-account routes on the nav are removed on logout, but it doesn't animate their exits; they just disappear suddenly. I think that's probably fine, and I didn't find a way to add the animation in.

chrisbobbe:
Marked as a draft because in particular I think this may be lacking some test coverage that we'll want; for example, of the various `if (maybeAccount == null) return;`s added in `PerAccountStore`.

lib/model/store.dart
```diff
@@ -723,6 +783,7 @@ class UpdateMachine {
     );
     final updateMachine = UpdateMachine.fromInitialSnapshot(
       store: store, initialSnapshot: initialSnapshot);
+    store.updateMachine = updateMachine;
```
gnprice, line 786: Let's put this assignment inside the `UpdateMachine` constructor. That way it's more crisply an invariant that the UpdateMachine's store will always point back to it.

Then relatedly (also about making invariants more explicit), `PerAccountStore.updateMachine` can get a setter that asserts the old value was null, so we're not redirecting the same store to start pointing to a different update machine.

lib/model/store.dart
```diff
@@ -190,10 +208,27 @@ abstract class GlobalStore extends ChangeNotifier {
...
 
+  /// Remove an account from the store.
+  Future<void> removeAccount(int accountId) async {
+    await doRemoveAccount(accountId);
+    assert(_accounts.containsKey(accountId));
+    _accounts.remove(accountId);
+    _perAccountStores.remove(accountId)
+      // TODO close connection inside `.dispose` instead (once tests can adapt)
+      ?..dispose()..connection.close();
+    _perAccountStoresLoading.remove(accountId);
```
gnprice, line 219: Even though this is logic I largely suggested a week or two ago, I found it kind of confusing to read. :slightly_smiling_face: I think the logic is good but would benefit from some comments.

I can take a pass at writing those comments in the next round.

test/model/store_test.dart
```diff
@@ -164,6 +164,88 @@ void main() {
...
+    }) {
+      expectAccount
+        ? check(store.getAccount(accountId)).isNotNull()
+        : check(store.getAccount(accountId)).isNull();
+      expectStore
+        ? check(store.debugPerAccountStores[accountId]).isNotNull()
+        : check(store.debugPerAccountStores[accountId]).isNull();
+      expectStoreLoading
+        ? check(store.debugPerAccountStoresLoading[accountId]).isNotNull()
+        : check(store.debugPerAccountStoresLoading[accountId]).isNull();
```
gnprice, line 181: Ideally these tests would be written without this inspection of the internal state of GlobalStore, and instead in terms of its outward-facing behavior seen by its callers. Ultimately all the behavior we care about should be observable that way.

test/model/store_test.dart
```diff
@@ -164,6 +164,88 @@ void main() {
...
+
+      checkGlobalStore(globalStore, eg.selfAccount.id,
+        expectAccount: true, expectStore: false, expectStoreLoading: true);
+      int notifyCount = 0;
+      globalStore.addListener(() => notifyCount++);
+
+      await globalStore.removeAccount(eg.selfAccount.id);
+
+      checkGlobalStore(globalStore, eg.selfAccount.id,
+        expectAccount: false, expectStore: false, expectStoreLoading: false);
```
gnprice, line 234: For example here, I think the fact that we remove the account from `_perAccountStoresLoading` can be observed by checking:
* A fresh `perAccount` call at this point throws directly, rather than returning a Future that throws. (Though this fact isn't that important — maybe shouldn't even get a test, since it'd be fine if it changed.)
* A fresh `perAccount` call *after* the old `loadingFuture` has completed will throw directly, rather than returning a Future that throws. This basically expresses the fact that we don't have an entry in `_perAccountStoresLoading` sticking around indefinitely after the whole interaction is done.

test/model/store_test.dart
```diff
@@ -164,6 +164,88 @@ void main() {
...
+      required bool expectAccount,
+      required bool expectStore,
+      bool expectStoreLoading = false,
+    }) {
+      expectAccount
+        ? check(store.getAccount(accountId)).isNotNull()
+        : check(store.getAccount(accountId)).isNull();
+      expectStore
+        ? check(store.debugPerAccountStores[accountId]).isNotNull()
+        : check(store.debugPerAccountStores[accountId]).isNull();
```
gnprice, line 178: These can just use `perAccountSync`, right?

test/model/store_test.dart
gnprice: Or thinking again, maybe the only fact we really want here about `_perAccountStoresLoading` is:
* At the end of the whole story, `_perAccountStoresLoading` is empty. (Could have a debug getter that just reports its length.)

In other words that means we aren't leaking a reference to an old Future that's no longer relevant.

The difference between throwing immediately and returning a Future that throws is maybe nice but isn't really fundamental — it's not like we need to have an extra fast answer when looking for an account that no longer exists.

gnprice (commented):
Thanks! Looking forward to having this.

Comments below on reading mainly the logic asked about in your last comment above.

lib/model/store.dart
```diff
@@ -75,7 +75,10 @@ abstract class GlobalStore extends ChangeNotifier {
       email: account.email, apiKey: account.apiKey);
   }
 
+  Map<int, PerAccountStore> get debugPerAccountStores => _perAccountStores;
```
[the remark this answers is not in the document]
chrisbobbe: Oh, interesting. Sure.

lib/model/store.dart
chrisbobbe: Cool; thanks.

test/model/store_test.dart
chrisbobbe: > * At the end of the whole story, `_perAccountStoresLoading` is empty. (Could have a debug getter that just reports its length.)

Cool; sure, I'll try this for the next revision.

chrisbobbe:

lib/api/route/notifications.dart
```diff
@@ -1,11 +1,7 @@
 
 import '../core.dart';
 
-// This endpoint is undocumented.  Compare zulip-mobile:
-//   https://github.com/zulip/zulip-mobile/blob/86d94fa89/src/api/notifications/savePushToken.js
-// and see the server implementation:
-//   https://github.com/zulip/zulip/blob/34ceafadd/zproject/urls.py#L383
-//   https://github.com/zulip/zulip/blob/34ceafadd/zerver/views/push_notifications.py#L47
+/// https://zulip.com/api/add-fcm-token
 Future<void> registerFcmToken(ApiConnection connection, {
```
gnprice, line 5: After reading these API docs against our code, I just sent #998 to more fully update to what the docs say. I think that replaces part of this commit from this PR:
e20fa5e99 api: Add unregister{Apns,Fcm}Token; link to new docs for register-token fns

lib/widgets/store.dart
```diff
@@ -195,18 +205,42 @@ class _PerAccountStoreWidgetState extends State<PerAccountStoreWidget> {
...
+          // `store != null` case above.
+          await globalStore.perAccount(widget.accountId);
+        } on AccountNotFoundException {
+          // The account was logged out while its store was loading.
+          // This widget will be showing [placeholder] perpetually,
+          // but that's OK as long as other code will be removing it from the UI
+          // (usually by using [routeToRemoveOnLogout]).
+        } catch (e) {
+          // TODO(log)
+        }
```
gnprice, line 238:
```suggestion
        }
```

If we don't catch the exception, it'll become an unhandled exception (because nothing awaits the future returned by this IIFE, and the exception will cause that future to complete with an error). That'll automatically cause it to get logged, once we have any kind of generic logging in place.

Could also say something like
```dart
          rethrow; // TODO(log)
```

but, apart from playing host to the comment, that has exactly the same effect as not having the `catch` block in the first place. Because I think we don't need the TODO, that makes it redundant.

lib/widgets/app.dart
```diff
@@ -221,15 +223,63 @@ class ChooseAccountPage extends StatelessWidget {
...
   }
 
+  Future<void> _logOutAccount(BuildContext context, int accountId) async {
+    final globalStore = GlobalStoreWidget.of(context);
+
+    final account = globalStore.getAccount(accountId);
+    if (account == null) return; // TODO(log)
+
+    // Async IIFE to not block removing the account on the unregister-token request.
+    unawaited(() async {
```
gnprice, line 261: nit: move this anonymous function to be its own private method with a name; it's 17 lines long, so I think that'll help make the overall structure easier to see

lib/widgets/app.dart
```diff
@@ -221,15 +223,63 @@ class ChooseAccountPage extends StatelessWidget {
...
+        await NotificationService.unregisterToken(connection, token: token);
+      } catch (e) {
+        // TODO retry? handle failures?
+      } finally {
+        connection.close();
+      }
+    }());
+
+    // TODO error handling? disable logout button during this?
+    await globalStore.removeAccount(accountId);
```
gnprice, line 280: Eh — this is only talking to the local database, not the network, so it should pretty consistently be fast and reliable. I'd guess p50 of <1ms, p99 of <50ms.

(That's jargon/shorthand for "50th percentile" and "99th percentile" respectively.)

And this is an uncommon operation. So it isn't worth building extra UI for the small fraction of users where it takes more than a frame for the operation to complete.

gnprice (commented):
Thanks! Generally this looks good. Comments below, having read completely through the first 7 commits:
114784517 test: Assert `isOpen` on FakeApiConnection.prepare
e20fa5e99 api: Add unregister{Apns,Fcm}Token; link to new docs for register-token fns
de0213e8f store [nfc]: Add indirection doLoadPerAccount
a73e9297d store [nfc]: Comment in _PerAccountStoreWidgetState.didChangeDependencies
5e7d2f5e9 store: Add nullable [PerAccountStore.updateMachine]; dispose in [dispose]
a2718637a store: Implement removeAccount
3e11686f5 notif: Implement NotificationService.unregisterToken

and skimmed the last one:
054da3120 login: Support logging out of an account

I also sent #998 as mentioned above; and I have a draft locally of the comments I mentioned at https://github.com/zulip/zulip-flutter/pull/948#discussion_r1769366315 last time.

chrisbobbe:

lib/model/store.dart
```diff
@@ -144,7 +146,22 @@ abstract class GlobalStore extends ChangeNotifier {
...
   /// Other callers interested in per-account data should use [perAccount]
   /// and/or [perAccountSync].
-  Future<PerAccountStore> loadPerAccount(int accountId);
+  Future<PerAccountStore> loadPerAccount(int accountId) async {
+    assert(_accounts.containsKey(accountId));
+    final store = await doLoadPerAccount(accountId);
+    if (!_accounts.containsKey(accountId)) {
+      // [removeAccount] was called during [doLoadPerAccount].
+      // TODO close connection inside `.dispose` instead (once tests can adapt)
+      store..dispose()..connection.close();
```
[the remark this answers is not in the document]
chrisbobbe: Thanks for going back over this with me in the office! :)

lib/model/store.dart
```diff
@@ -441,14 +487,20 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
 
   @override
   void dispose() {
+    assert(!_disposed);
     recentDmConversationsView.dispose();
     unreads.dispose();
     _messages.dispose();
     typingStatus.dispose();
+    updateMachine?.dispose();
+    connection.close();
```
gnprice, line 496: nit:

    This seems hygienic. It makes sense to explicitly clean up HTTP
    client resources when we know we'll be setting up a new client
    instance to replace the old one. (The dead-queue reload is currently
    the only path where PerAccountStore.dispose is called.)

In this situation it's not that we will be setting up a new HTTP client instance; rather we already have, which was passed to the new PerAccountStore (and even before that, was used for fetching the new store's initial snapshot). The new store gets constructed before the old one is disposed.

gnprice (commented):
Thanks for the revision! Generally this looks great; comments below.

I haven't yet read fully through the final commit:
6bc8bcc60 login: Support logging out of an account

In the interest of getting things merged efficiently, maybe let's split this PR up — let's have this first PR cover the commits leading up to `removeAccount`:
ed1adbea9 test: Assert `isOpen` on FakeApiConnection.prepare
18914a44c store [nfc]: Add indirection doLoadPerAccount
6cfe90030 store [nfc]: Comment in _PerAccountStoreWidgetState.didChangeDependencies
260991021 store: Add nullable [PerAccountStore.updateMachine]; dispose in [dispose]
b2cedd568 store: Add dartdoc on UpdateMachine.dispose
3e0d49920 store: Add and check _disposed on PerAccountStore and UpdateMachine
7aff4d1bf store test: Expose TestGlobalStore.useCachedApiConnections; default false
f3a336109 store: Call connection.close in PerAccountStore.dispose
ab9ce6d1e store: Implement removeAccount

and then a follow-up PR can cover the remaining commits:
325c663ae api: Add remove{Apns,Fcm}Token
b4b956974 notif: Implement NotificationService.unregisterToken
6bc8bcc60 login: Support logging out of an account

alya:
Do the screenshots need to be updated for this one?

chrisbobbe:
> Do the screenshots need to be updated for this one?

Ah right! Actually the UI hasn't changed from what's shown in the screenshots in the PR description. I remember us talking in the office about making it look better, but I don't remember if we settled on something different. If there are changes to make, how about doing them as a followup?

chrisbobbe:
> In the interest of getting things merged efficiently, maybe let's split this PR up

Sounds good. I just sent #1007 for those first several commits, and I'll send a followup with the remaining ones (implementing logout) as a followup. Then as mentioned in my previous comment, maybe any UI changes can be a followup to that.

lib/model/store.dart
```diff
@@ -890,6 +959,8 @@ class UpdateMachine {
         }
       }
 
+      if (_disposed) return;
```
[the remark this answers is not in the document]
chrisbobbe: Things do feel more consistent and better with this pattern! :)

PR #949, opened 2024-09-17
test: Use color checks that tolerate floating-point errors

chrisbobbe:
Done to follow a draft migration guide:
  https://github.com/flutter/website/blob/997c893ccec5fb15dd085a4d23fe8c43870c631b/src/content/release/breaking-changes/wide-gamut-framework.md
for the proposed breaking change that raised
  https://github.com/flutter/flutter/issues/155113

Some other work was needed outside what's described in that draft;
see:
  https://github.com/flutter/flutter/issues/155113#issuecomment-2347374468
  https://github.com/flutter/flutter/issues/155113#issuecomment-2356356254

### DIFF
--- a/test/widgets/colors_checks.dart
+++ b/test/widgets/colors_checks.dart
@@ -1,0 +1,11 @@
+import 'package:checks/checks.dart';
+import 'package:flutter_test/flutter_test.dart' as flutter_matcher;
+import 'package:flutter/painting.dart';
+import 'package:legacy_checks/legacy_checks.dart';
+
+extension ColorSwatchChecks<T> on Subject<ColorSwatch<T>> {
+  /// package:checks-style wrapper for [flutter_matcher.isSameColorSwatchAs].
+  void isSameColorSwatchAs(ColorSwatch<T> colorSwatch) {
+    legacyMatcher(flutter_matcher.isSameColorSwatchAs(colorSwatch));
+  }
+}
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1360,5 +1360,5 @@ packages:
     source: path
     version: "0.0.1"
 sdks:
-  dart: ">=3.6.0-217.0.dev <4.0.0"
-  flutter: ">=3.25.0-1.0.pre.296"
+  dart: ">=3.6.0-263.0.dev <4.0.0"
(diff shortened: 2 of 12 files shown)


gnprice (commented):

The changes all look good, except that I think a Flutter deps bump is needed before adding the references to `isSameColorSwatchAs`. (The analyzer is failing for me locally because I have a Flutter tree from a few days ago.)

    Some other work was needed outside what's described in that draft;
    see:
      https://github.com/flutter/flutter/issues/155113#issuecomment-2347374468
      https://github.com/flutter/flutter/issues/155113#issuecomment-2356356254

I think this work just meant (a) using `package:flutter_checks`, and (b) those few lines adding a wrapper for isSameColorSwatchAs. Is that right or is there another piece I'm missing?

chrisbobbe:
Yep! Revision pushed, PTAL.

gnprice:
Thanks! Looks good — merging.

test/widgets/colors_checks.dart
```diff
@@ -0,0 +1,11 @@
...
+import 'package:flutter_test/flutter_test.dart' as flutter_matcher;
+import 'package:flutter/painting.dart';
+import 'package:legacy_checks/legacy_checks.dart';
+
+extension ColorSwatchChecks<T> on Subject<ColorSwatch<T>> {
+  /// package:checks-style wrapper for [flutter_matcher.isSameColorSwatchAs].
+  void isSameColorSwatchAs(ColorSwatch<T> colorSwatch) {
+    legacyMatcher(flutter_matcher.isSameColorSwatchAs(colorSwatch));
+  }
+}
```
lishaduck, line 11: Cool, I'm glad `legacy_checks` was useful!
I was worried I'd missed one for a second before I realized it was new. I know y'all follow the master channel, so, as I don't think I wrote down a flutter version support policy anywhere, I guess I'll just note here that I don't want to publish anything that unnecessarily requires a "nightly". I'm happy to accept PRs once they land in stable, but you'll probably still need to maintain a few of these.

gnprice: Sure, that approach makes sense.

Thanks to `legacyMatcher`, any individual wrappers like this one should be pretty cheap to write, like this one was.

PR #951, opened 2024-09-18
deps: Upgrade video_player_android to 2.7.3

rajveermalviya:
Changelog:
  https://pub.dev/packages/video_player_android/changelog

This update migrates the plugin to use newer API recommended
to support Impeller:
  https://docs.flutter.dev/release/breaking-changes/android-surface-plugins

Without this migration, the videos do play but they look
pixelated, resulting in a bad UX.

| before | after |
| - | - |
| before | after |

Additionally, it updates the underlying Media3-ExoPlayer
library to version 1.4.1:
  https://developer.android.com/jetpack/androidx/releases/media3#1.4.1

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1244,10 +1244,10 @@ packages:
     dependency: transitive
     description:
       name: video_player_android
-      sha256: "4de50df9ee786f5891d3281e1e633d7b142ef1acf47392592eb91cba5d355849"
+      sha256: "38d8fe136c427abdce68b5e8c3c08ea29d7a794b453c7a51b12ecfad4aad9437"
       url: "https://pub.dev"
     source: hosted
-    version: "2.6.0"
+    version: "2.7.3"
   video_player_avfoundation:
     dependency: transitive
     description:


chrisbobbe:
Sure, LGTM, thanks! Marking for Greg's review.

gnprice:
Thanks! Looks good — merging.

> Additionally, it updates the underlying Media3-ExoPlayer
> library to version 1.4.1:
> https://developer.android.com/jetpack/androidx/releases/media3#1.4.1

Was the previous version 1.4.0, or something else? That's always important information for being able to interpret a changelog like that :slightly_smiling_face: 

(Not blocking the merge on that, because knowing those details *probably* won't end up being relevant, and if it does then we can always do the archeology then.)

rajveermalviya:
> Was the previous version 1.4.0

Correct, previously it was 1.4.0 (https://github.com/flutter/packages/commit/a6db33c25dc33191c187035eea53623cfa1d16f8; Side note: The squash merge of the Dependabot PR uses the PR description as the commit message, which results in an ugly commit message. The actual commit message can be found [here](https://github.com/flutter/packages/pull/7564/commits/a62ef8d453a1470d9443b866f42731b1f8ccafac).)

> That's always important information for being able to interpret a changelog like that 🙂
> …
> (Not blocking the merge on that,

Thanks, I'll keep that in mind!

gnprice:
> The squash merge of the Dependabot PR uses the PR description as the commit message, which results in an ugly commit message.

Wow, yep. Good that Dependabot provides something more readable in its original commit message.

Issue #953, opened 2024-09-19
Emoji in messages not fully opaque in dark mode


chrisbobbe:
@gnprice reported:

https://chat.zulip.org/#narrow/stream/48-mobile/topic/dark.20theme/near/1924392
> Perhaps related: emoji in message content look dim, as if they're partly transparent and have the background showing through.

Filing this issue to track that bug; I'll send a fix soon.

PR #954, opened 2024-09-19
msglist: Follow web's change making dark mode's --color-text-default opaque

chrisbobbe:
Fixes: #953

| Before | After |
| --- | --- |
| image | image |


gnprice:
Thanks for the fix! Looks good; merging.

I think this had been awaiting my review, but it didn't have the label so I skipped past looking at it. Fixed now.

chrisbobbe:
Ah indeed. Thanks for catching it!

PR #955, opened 2024-09-19
ui: Make some dark-theme adjustments from Greg's feedback

chrisbobbe:
This addresses all the remaining items in the CZO thread at https://chat.zulip.org/#narrow/stream/48-mobile/topic/dark.20theme/near/1936853 .

Screenshots coming soon.

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -299,6 +299,8 @@ class HomePage extends StatelessWidget {
     final store = PerAccountStoreWidget.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
 
+    final colorScheme = ColorScheme.of(context);
+
     InlineSpan bold(String text) => TextSpan(
       style: const TextStyle().merge(weightVariableTextStyle(context, wght: 700)),
       text: text);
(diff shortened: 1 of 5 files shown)


chrisbobbe:
Home page (light theme affected too because it was convenient to just swap a light/dark pair that comes from Material):

| Before | After |
| --- | --- |
| image | image |
| image | image |

Error style (light theme unchanged):

| Before | After |
| --- | --- |
| image | image |

Dot for muted unreads (light theme unchanged):

The way this looks, I think it might not be an improvement, actually. The color matches what gets used on web, but it ends up with pretty low contrast because the background color differs between mobile and web. But also, like many UI elements that currently follow web, this is something we'll revisit once we have a design from Vlad.

| Before | After |
| --- | --- |
| image | image |

chrisbobbe:
cc @alya in case you'd like to take a look.

alya:
Works for me, thanks!

lib/widgets/app.dart
```diff
@@ -254,63 +256,68 @@ class HomePage extends StatelessWidget {
...
+            if (testStreamId != null) ...[
+              const SizedBox(height: 16),
+              ElevatedButton(
+                onPressed: () => Navigator.push(context,
+                  MessageListPage.buildRoute(context: context,
+                    narrow: ChannelNarrow(testStreamId!))),
+                child: const Text("#test here")), // scaffolding hack, see above
+            ],
+          ])),
+      ));
```
gnprice, line 321: nit:
```suggestion
          ]))));
```

lib/widgets/theme.dart
```diff
@@ -158,6 +159,7 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       groupDmConversationIconBg: const Color(0x33cccccc),
       loginOrDivider: const Color(0xff424242),
       loginOrDividerText: const Color(0xffa8a8a8),
+      mutedUnreadBadge: const HSLColor.fromAHSL(0.5, 0, 0, 0.3).toColor(),
```
gnprice, line 162: This is pretty low-contrast in dark mode, but as with many of our UI
    elements, we'll adjust as necessary when we have a design from Vlad.

From the screenshot, this gets pretty hard to see. It's good to have this properly wired up into DesignVariables, but I think the existing value that's the same as in light theme actually works out better.

Let's maybe just make this the same value (with 0.8 lightness) as for the light theme, and leave a `// TODO(design-dark)` for it. Or spend a minute or two trying values in between like 0.6, and pick something that looks clear.

gnprice (commented):
Thanks for taking care of these! Two small comments.

chrisbobbe:
Thanks! Revision pushed.

Here's the before-and-after for `MutedUnreadBadge` in dark theme (`main` vs. the tip of the branch), with 0.6 for the lightness:

| Before | After |
| --- | --- |
| image | image |

gnprice:
Thanks! Looks good; merging.

PR #958, opened 2024-09-19
ios: Add workaround for package:firebase_messaging when building with Xcode 16

rajveermalviya:
Without this workaround iOS build currently fails with the following error:
```
Lexical or Preprocessor Issue (Xcode): Include of non-modular header inside framework module 'firebase_messaging FLTFirebaseMessagingPlugin':
'/Users/rajveer/Projects/zulip-flutter/ios/Pods/Headers/Public/Firebase/Firebase.h'
/Users/rajveer/.pub-cache/hosted/pub.dev/firebase_messaging-15.0.4/ios/Classes/FLTFirebaseMessagingPlugin.h:11:8
```
Upstream issue and the workaround comment:
  https://github.com/firebase/flutterfire/issues/13323
  https://github.com/firebase/flutterfire/issues/13323#issuecomment-2355013360

Docs: https://developer.apple.com/documentation/xcode/build-settings-reference#Allow-Non-modular-Includes-In-Framework-Modules

The documentation states that enabling this setting may result in issues later on, which is why this workaround should only be temporary until firebase_messaging is patched with the correct fix.

### DIFF
--- a/ios/Flutter/Zulip.xcconfig
+++ b/ios/Flutter/Zulip.xcconfig
@@ -1,0 +1,10 @@
+// Configuration settings file format documentation can be found at:
+// https://help.apple.com/xcode/#/dev745c5c974
+
+// TODO(firebase/flutterfire#13323): remove this flag
+//
+// This flag is added to work around the iOS build failing
+// on Xcode 16.  Remove it when `package:firebase_messaging`
+// is patched with a fix to build successfully on Xcode 16:
+//   https://github.com/firebase/flutterfire/issues/13323
+CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES=YES
--- a/ios/Flutter/Debug.xcconfig
+++ b/ios/Flutter/Debug.xcconfig
@@ -1,2 +1,3 @@
 #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
 #include "Generated.xcconfig"
+#include "Zulip.xcconfig"
--- a/ios/Flutter/Release.xcconfig
+++ b/ios/Flutter/Release.xcconfig
@@ -1,2 +1,3 @@
 #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.release.xcconfig"
 #include "Generated.xcconfig"
+#include "Zulip.xcconfig"
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -48,6 +48,7 @@
 		97C146FD1CF9000F007C117D /* Assets.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = Assets.xcassets; sourceTree = "<group>"; };
 		97C147001CF9000F007C117D /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/LaunchScreen.storyboard; sourceTree = "<group>"; };
 		97C147021CF9000F007C117D /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
+		B3AF53A72CA20BD10039801D /* Zulip.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; name = Zulip.xcconfig; path = Flutter/Zulip.xcconfig; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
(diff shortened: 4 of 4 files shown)


rajveermalviya:
Chat thread:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Can't.20compile.20for.20iOS.20on.20macOS.2015.2E0.20.28latest.29.20and.20Xcode.2016.2E0/near/1944045

chrisbobbe:
LGTM, and I've tested on Xcode `Version 15.4 (15F31d)` and it works there too.

An alternative that should also work, and worked in my testing, is to add a line

```
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES
```

in ios/Flutter/Debug.xcconfig and ios/Flutter/Release.xcconfig, and leave `project.pbxproj` untouched. This might be nicer, because the `project.pbxproj` is hard to read, and there's a lot in there. Putting this setting in an uncrowded file, with a helpful comment saying it's a temporary workaround for a specific issue, seems helpful.

The Xcode GUI lets you see how the final value for this setting is resolved.

Here's with the PR as-is, with the additions in `project.pbxproj`:

image

And with my proposal:

image

In my proposal, the prevailing "YES" value comes from the config-file column, instead of the target ("Runner") column.

I wasn't sure which of `Debug.xcconfig` and `Release.xcconfig` would control the "Profile" configuration, or if neither of them would, and I found experimentally that it's `Release.xcconfig`. I'm not finding the place in the Xcode GUI that tells me that, but this looks like the spot in `project.pbxproj`:

```
		249021D4217E4FDB00AE95B9 /* Profile */ = {
			isa = XCBuildConfiguration;
			baseConfigurationReference = 7AFA3C8E1D35360C0083082E /* Release.xcconfig */;
```

gnprice:
> ios/Flutter/Debug.xcconfig and ios/Flutter/Release.xcconfig, and leave `project.pbxproj` untouched.

Neat, I like that option. @rajveermalviya would you try that and confirm that it works for you?

It looks like those files also accept comments, with `//`. (See `Generated.xcconfig` in the same directory for an example.) That'd be good for mentioning that it's a workaround which should be temporary, and linking to the chat thread and that upstream issue https://github.com/firebase/flutterfire/issues/13323 .

rajveermalviya:
Thanks for the reviews @chrisbobbe and @gnprice!

Yes, adding the flag to `ios/Flutter/{Debug.xcconfig, Release.xcconfig}` does result in a successfull build as well.
I've pushed a new revision which does that via `Zulip.xcconfig` (new file) and `#include`-ing that in both `Debug.xcconfig` and `Release.xcconfig`.

ios/Flutter/Zulip.xcconfig
```diff
@@ -0,0 +1,7 @@
+// TODO(firebase/flutterfire#13323): remove this flag
+//
+// This flag is added to workaround the iOS build failing
+// on Xcode 16, remove it when `package:firebase_messaging`
+// is patched with a fix to build successfully on Xcode 16.
+//
```
PIG208, line 6: nit: extra empty line

rajveermalviya:
Thanks for the review @PIG208, pushed a new revision.

Additionally, added references for `Zulip.xcconfig` so that it becomes visible in XCode (without them it's not).

ios/Flutter/Zulip.xcconfig
```diff
@@ -0,0 +1,9 @@
+// Configuration settings file format documentation can be found at:
+// https://help.apple.com/xcode/#/dev745c5c974
```
gnprice, line 2: Helpful, thanks for including this link.

ios/Flutter/Zulip.xcconfig
```diff
@@ -0,0 +1,9 @@
...
+
+// TODO(firebase/flutterfire#13323): remove this flag
```
gnprice, line 4: I guess we've talked about having a tool to scan for TODO comments that point to closed issues, and when we do it could potentially understand this syntax for external issues. So it's fine to include it.

But let's also include an actual link. That will make it nice and easy for someone who happens across this a month from now or a year from now to quickly check on the state of the upstream issue, even if we haven't built that potential tool by then.

ios/Flutter/Zulip.xcconfig
```diff
@@ -0,0 +1,9 @@
...
+//
+// This flag is added to workaround the iOS build failing
```
gnprice, line 6: nit: the verb is "work around", and "workaround" is its noun form:
```suggestion
// This flag is added to work around the iOS build failing
```

Just like other verb phrases made of a verb plus preposition, where they're run together to make a noun:
"to log in" / "this login"
"to set up" / "this setup"
"to clean up" / "this cleanup"

(In each case the two versions are pronounced differently, too — the stress moves from the end in the verb form to the beginning in the noun form. That probably contributes to making the difference salient for me.)

ios/Flutter/Zulip.xcconfig
```diff
@@ -0,0 +1,9 @@
...
+// on Xcode 16, remove it when `package:firebase_messaging`
+// is patched with a fix to build successfully on Xcode 16.
```
gnprice, line 8:
```suggestion
// This flag is added to work around the iOS build failing
// on Xcode 16.  Remove it when `package:firebase_messaging`
// is patched with a fix to build successfully on Xcode 16:
//   https://github.com/firebase/flutterfire/issues/13323
```

gnprice (commented):
Thanks! I like this structure. Small comment tweaks below.

Then this is otherwise ready for merge, so please update the commit message to a final/mergeable version.

rajveermalviya:

gnprice:
Thanks! Looks good — merging.

PR #961, opened 2024-09-24
Prepare for curly_braces lint rule change

srawlins:
To fix https://github.com/dart-lang/linter/issues/4870, the analyzer is now reporting if-statement bodies without curly braces when the if-statement condition spans multiple lines.

This change prepares zulip-flutter for this lint rule change.


gnprice:
Thanks! Merged.

Issue #670, opened 2024-05-13
Autocomplete for emoji in message compose


gnprice:
This is the third top-level kind of autocomplete in Zulip messages, after @-mentions #49 and #-mentions #124. It triggers on a colon `:`.

Help Center doc is here (see the "Via Markdown" tab under "Use an emoji in your message"):
https://zulip.com/help/emoji-and-emoticons

The strings this autocomplete acts on are Zulip "emoji names". Zulip has a particular data model for identifying emoji, which is an underdocumented part of the Zulip API. As a prerequisite for this issue, we'll implement that model:
* #669

Issue #669, opened 2024-05-13
Track all valid Zulip emoji


gnprice:
This is a prerequisite for:
* #388 
* #670

In order to offer the user a UI to choose an emoji from the emoji Zulip knows about, we'll need to know what emoji those are and what Zulip calls them.

Zulip has a particular data model for identifying emoji, which is an underdocumented part of the Zulip API. For an introduction, see `reaction_type` here:
https://zulip.com/api/add-reaction#parameter-reaction_type
and the two companion fields `emoji_code` and `emoji_name`. (The same emoji model is used for reactions as for emoji in messages.)

Then see zulip-mobile's `src/emoji/data.js`, which tracks this model.

A good implementation of this issue will:
* Cover all three of Zulip's types of emoji.
* Support efficiently looking up an emoji by name, or by type and code.
* Support iterating through the list of all Zulip emoji.

gnprice:
With #976 (which I've just sent) we'll have all the data.

The remaining data-model work I want to do before declaring victory here is to handle searching through the list of emoji to produce the data we'd use for an emoji-picker UI (for #388 or #670). Between that and the logic we already have (and in #967 refactored onto EmojiStore) for looking up an individual emoji to decide how to display it, we'll be covering all the ways the app needs to use emoji data, which is the key criterion for being confident that our model has all the features it needs.

PR #967, opened 2024-09-27
Introduce EmojiStore for some existing emoji logic

gnprice:
This is a refactoring that prepares us for #669, as well as for #388 and #670.

### DIFF
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -3,7 +3,6 @@ import 'package:flutter/material.dart';
 
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
-import '../model/content.dart';
 import 'content.dart';
 import 'store.dart';
 import 'text.dart';
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -3,6 +3,7 @@ import 'package:flutter/material.dart';
 
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
+import '../model/emoji.dart';
 import 'content.dart';
 import 'store.dart';
 import 'text.dart';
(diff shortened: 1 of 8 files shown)


test/model/emoji_test.dart
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      checkDisplay(emojiCode: '100', emojiName: 'logo').isA<ImageEmojiDisplay>()
+        ..resolvedUrl.equals(eg.realmUrl.resolve('/emoji/100.png'))
+        ..resolvedStillUrl.isNull();
+
+      // Code matches against code, not against name.
+      checkDisplay(emojiCode: '123', emojiName: '100').isA<ImageEmojiDisplay>()
+        ..resolvedUrl.equals(eg.realmUrl.resolve('/emoji/123.png'))
+        ..resolvedStillUrl.isNull();
```
chrisbobbe, line 49: I'm not sure I follow the comment. "Code" (the first word) means "emoji code", right, or does it mean the app code? What bug is this test written against?

chrisbobbe (commented):
Thanks! LGTM, small comments below.

chrisbobbe:
(CI should be resolved by rebasing past #968.)

test/model/emoji_test.dart
gnprice: Yeah, it means "emoji code" as opposed to "emoji name". I'll make the comment less elliptical:
```dart
      // Emoji code matches against emoji code, not against emoji name.
```

The hypothetical bug here isn't a super likely one (because if we had it, then probably most uses would just fail and so we'd notice promptly in manual testing): it's that because both emoji codes and emoji names are just strings, we could take the given emoji name and look it up someplace where the keys are actually supposed to be emoji codes.

lib/widgets/emoji_reaction.dart
```diff
@@ -391,43 +372,38 @@ class _ImageEmoji extends StatelessWidget {
...
-        src = doNotAnimate && item.stillUrl != null ? item.stillUrl! : item.sourceUrl;
-    }
-    final parsedSrc = Uri.tryParse(src);
-    if (parsedSrc == null) { // TODO(log)
-      return _textFallback;
-    }
-    final resolved = store.realmUrl.resolveUri(parsedSrc);
+    final resolvedUrl =
+      doNotAnimate ? (emojiDisplay.resolvedStillUrl ?? emojiDisplay.resolvedUrl)
+        : emojiDisplay.resolvedUrl;
```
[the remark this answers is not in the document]
gnprice: Sure, I'll do the second of those.

gnprice:
Thanks for the review! Merged after those changes.

Issue #774, opened 2024-06-28
Add our Android build to Flutter's customer tests


gnprice:
This will be a followup to:
* #772 

After we have the Android build running in our own CI, we should register it in Flutter's "customer tests" (like we did with our `flutter analyze` and `flutter test` as #239) so that it runs as part of Flutter's CI too. In particular this would have caught #772 at an earlier stage — the PR that made that change in the Flutter repo would have already seen a test failure.

I think there aren't any existing tests of this kind in the flutter/tests registry, and it's possible it'll call for some augmenting of that system's API. In particular it runs each test suite for 10 iterations by default, to catch flakes; but an Android build is a lot slower than a `flutter test` run on a comparable size of app, enough to make that expensive. Plus when it's purely a build, it's hard to imagine it flaking. So we may want to run that test for only 1 iteration, while still running our `flutter test` for 10 iterations. Anyway, that'll be a subject to raise in `#hackers-tests` in Discord if, when working on this issue, we decide something like that is indeed wanted.

gnprice:
When implementing this, one thing to take care of is that we should have it use an adjusted configuration that doesn't error out on deprecation warnings, something akin to `flutter analyze --no-fatal-infos`.

We saw on #968 an example of a deprecation warning in the Android code (affecting one of our plugin dependencies). As discussed at https://github.com/zulip/zulip-flutter/pull/968#issuecomment-2379876091 , tests in the Flutter "customer tests" are expected to ignore those when deciding pass/fail.

---

I believe we already error out (in the Android build) only on *warnings*, not info:
```groovy
    lint {
        // For docs on options available here:
        //   https://developer.android.com/reference/tools/gradle-api/8.5/com/android/build/api/dsl/Lint
        checkAllWarnings = true
        warningsAsErrors = true
    }
```

so I guess our normal config already complies with the policy as [actually written](https://github.com/flutter/tests). But I believe the motivation for the policy is to avoid erroring on deprecations. So if we can, it'd be good to find a way to get deprecations treated as "info" level, so that they don't cause errors.

From the doc linked in that comment:
https://developer.android.com/reference/tools/gradle-api/8.5/com/android/build/api/dsl/Lint
there's the `informational` option. That should do it, if there's an appropriate name we can pass to that option.

gnprice:
[...]

PR #968, opened 2024-09-27
deps: Upgrade Flutter SDK and `video_player_android`

rajveermalviya:
Fixes the CI failure: https://github.com/zulip/zulip-flutter/actions/runs/11063896720/job/30740794793#step:7:1899

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1276,10 +1276,10 @@ packages:
     dependency: transitive
     description:
       name: vm_service
-      sha256: "5c5f338a667b4c644744b661f309fb8080bb94b18a7e91ef1dbd343bed00ed6d"
+      sha256: f6be3ed8bd01289b34d679c2b62226f63c0e69f9fd2e50a6b3c1c729a961041b
       url: "https://pub.dev"
     source: hosted
-    version: "14.2.5"
+    version: "14.3.0"
   wakelock_plus:
     dependency: "direct main"
     description:
@@ -1392,5 +1392,5 @@ packages:
     source: path
     version: "0.0.1"
 sdks:
-  dart: ">=3.6.0-263.0.dev <4.0.0"
-  flutter: ">=3.26.0-1.0.pre.155"
+  dart: ">=3.6.0-279.0.dev <4.0.0"
+  flutter: ">=3.26.0-1.0.pre.249"
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1244,10 +1244,10 @@ packages:
     dependency: transitive
     description:
       name: video_player_android
-      sha256: "38d8fe136c427abdce68b5e8c3c08ea29d7a794b453c7a51b12ecfad4aad9437"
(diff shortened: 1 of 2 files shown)


rajveermalviya:
(If Android builds were part of Flutter's Customer test suite (https://github.com/zulip/zulip-flutter/issues/774), it would have added more friction on https://github.com/flutter/engine/pull/55463.)

gnprice:
Thanks for digging into this! I noticed CI fail on #967 last night, but it was right at the end of my day so I didn't investigate.

If I understand the commit messages right, `tools/check --all` would always fail at the middle commit, right? That commit makes us start requiring a Flutter version that will always cause that warning.

In that case, can the two commits go in the opposite order instead? I expect that way things work at every step (if you use the minimum required Flutter version), which is our usual approach because it makes things a bit cleaner to understand.

rajveermalviya:
Thanks for the review @gnprice, pushed a revision with reordered commits and updated commit message.

gnprice:
Thanks! Looks good — merging.

I tweaked the commit message a bit:

    -    Without this update, Android builds will fail with Flutter
    -    SDK revisions after this commit[0], due to our strict
    -    warnings as errors config in `app/build.gradle`.
    +    Without this update, Android builds will fail with Flutter SDK
    +    versions after the following change that landed yesterday:
    +      https://github.com/flutter/engine/pull/55463
    +      https://github.com/flutter/flutter/commit/7bb9352911d0471568a77fc57b94a646fe9571af
    +    due to our strict warnings as errors config in `app/build.gradle`.
         The build error would look like this:
     
           /…/video_player_android-2.7.3/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java:88:
           warning: [removal] onSurfaceCreated() in Callback has been deprecated and marked for removal
             public void onSurfaceCreated() {
                         ^
           error: warnings found and -Werror specified
     
    -    [0]: https://github.com/flutter/engine/commit/c2c242d50cf8d78057f347f600460f17c102e927
    -

mainly because the reference to "this commit" sounded like it meant *this* commit, i.e. the one the commit message is on. (I think just putting the link immediately after the reference, instead of down in a footnote, would have been enough to fix that problem. But then once I was editing it, I changed "this" to "the following" to further disambiguate, and added a bit more information too.)

gnprice:
> (If Android builds were part of Flutter's Customer test suite (#774), it would have added more friction on [flutter/engine#55463](https://github.com/flutter/engine/pull/55463).)

This is a good reminder that #774 would be good to do… but I think having our Android build there wouldn't actually have affected this. That's because [that repo's policy](https://github.com/flutter/tests) is that deprecation warnings are to be ignored. It says:
> The tests must pass even if there are analysis 'info' level issues in the code. Generally, this means that if the test performs static analysis, it does so ignoring info level items (i.e., `flutter analyze --no-fatal-infos`).

and the point of that is precisely about deprecation warnings — they want to be able to add deprecations and have those not be breaking changes. (Later when they actually *remove* the deprecated item, I guess that's a breaking change if people haven't migrated.)

So that's something I'll want to watch for when implementing #774. I'll make a note there.

PR #970, opened 2024-09-28
dialog [nfc]: Wrap showErrorDialog's return value

PIG208:
As a result, showErrorDialog no longer returns a Future that can be mistakenly awaited.  The wrapped return value offers clarity on how the Future is supposed to be used.

There is no user visible change because existing callers (other than lightbox's) that wait for its return value are all asynchronous handlers for UI elements like buttons, and waits unnecessarily.

See also discussion
  https://github.com/zulip/zulip-flutter/pull/937#discussion_r1774377096.

### DIFF
--- a/lib/widgets/dialog.dart
+++ b/lib/widgets/dialog.dart
@@ -15,16 +15,33 @@ Widget _dialogActionText(String text) {
   );
 }
 
+/// A wrapper providing access to the status of an [AlertDialog].
+///
+/// See also:
+///  * [showDialog], whose return value this class is intended to wrap.
+class DialogStatusController {
+  const DialogStatusController(this._closed);
+
+  /// Resolves when the dialog is closed.
+  Future<void> get closed => _closed;
+  final Future<void> _closed;
+}
+
 /// Displays an [AlertDialog] with a dismiss button.
 ///
-/// Returns a [Future] that resolves when the dialog is closed.
-Future<void> showErrorDialog({
+/// Returns a [DialogStatusController]. Useful for checking if the dialog has
+/// been closed.
+// This API is inspired by [ScaffoldManager.showSnackBar].  We wrap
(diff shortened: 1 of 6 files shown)


lib/widgets/dialog.dart
```diff
@@ -15,16 +15,33 @@ Widget _dialogActionText(String text) {
...
-/// Returns a [Future] that resolves when the dialog is closed.
-Future<void> showErrorDialog({
+/// See also:
+///  * [showDialog], whose return value this class is intended to wrap.
+class DialogStatusController {
+  const DialogStatusController(this._closed);
+
+  final Future<void> _closed;
+  /// Resolves when the dialog is closed.
+  Future<void> get closed => _closed;
```
chrisbobbe, line 27: nit: getter before private field: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense

lib/widgets/lightbox.dart
```diff
@@ -476,11 +476,12 @@ class _VideoLightboxPageState extends State<VideoLightboxPage> with PerAccountSt
...
       if (!mounted) return;
       final zulipLocalizations = ZulipLocalizations.of(context);
-      // Wait until the dialog is closed
-      await showErrorDialog(
+      final dialog = showErrorDialog(
         context: context,
         title: zulipLocalizations.errorDialogTitle,
         message: zulipLocalizations.errorVideoPlayerFailed);
+      // Wait until the dialog is closed
+      await dialog.closed;
```
chrisbobbe, line 484: Happily, I think the comment is now redundant and can be swept away :)

chrisbobbe (commented):
Small comments below, otherwise LGTM. Thanks!

PIG208:

chrisbobbe:
LGTM, thanks! Over to Greg.

PR #975, opened 2024-10-01
backoff: Make duration bounds customizable

gnprice:
This will let us have some backoff loops that back off more patiently than others, e.g. because their work is happening in the background and isn't expected to be something the user is already waiting for.

In particular I plan to use this for backing off the fetch of server emoji data, for #669.

### DIFF
--- a/lib/api/backoff.dart
+++ b/lib/api/backoff.dart
@@ -5,7 +5,10 @@ import 'dart:math';
 /// Call the constructor before a loop starts, and call [wait] in each iteration
 /// of the loop.  Do not re-use the instance after exiting the loop.
 class BackoffMachine {
-  BackoffMachine();
+  BackoffMachine({
+    this.firstBound = const Duration(milliseconds: 100),
+    this.maxBound = const Duration(seconds: 10),
+  }) : assert(firstBound <= maxBound);
 
   /// How many waits have completed so far.
   ///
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Ah I reviewed this before I noticed it was assigned to @PIG208. LGTM, but I'd be happy to wait for @PIG208's review if that's desired or @PIG208 if you've started yours already. :)

lib/api/backoff.dart
```diff
@@ -7,23 +7,38 @@ import 'dart:math';
...
   int get waitsCompleted => _waitsCompleted;
   int _waitsCompleted = 0;
 
-  /// A future that resolves after the appropriate duration.
+  /// The bound on the duration of the first wait.
+  ///
+  /// The actual duration will vary randomly up to this value; see [wait].
+  static const firstBound = Duration(milliseconds: 100);
+
+  /// The maximum bound on the duration of each wait, even after many waits.
```
PIG208, line 25: nit: Should we mention explicitly that this, and `firstBound` are upper bounds on the first line, respectively? I feel that these are the first reads for one when looking at the `BackoffMachine` interface.

PIG208:
Sorry for the delay! Posted a comment. This LGTM overall.

gnprice:
Thanks for the reviews! Merging, with that tweak.

PR #981, opened 2024-10-04
notif: Support migration of Android notification channels

rajveermalviya:
(Split from #717)

Needed for #340, when updating notification channels to use a custom notification sound.

### DIFF
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -25,6 +25,10 @@ AndroidNotificationHostApi get _androidHost => ZulipBinding.instance.androidNoti
 
 /// Service for configuring our Android "notification channel".
 class NotificationChannelManager {
+  /// The channel ID we use for our one notification channel, which we use for
+  /// all notifications.
+  // TODO(launch) check this doesn't match zulip-mobile's current or previous
+  //   channel IDs
   @visibleForTesting
   static const kChannelId = 'messages-1';
 
@@ -36,6 +40,8 @@ class NotificationChannelManager {
   static final kVibrationPattern = Int64List.fromList([0, 125, 100, 450]);
 
   /// Create our notification channel, if it doesn't already exist.
+  ///
+  /// Deletes obsolete channels, if present, from old versions of the app.
   //
   // NOTE when changing anything here: the changes will not take effect
   // for existing installs of the app!  That's because we'll have already
@@ -52,11 +58,28 @@ class NotificationChannelManager {
   //    settings for the channel -- like "override Do Not Disturb", or "use
   //    a different sound", or "don't pop on screen" -- their changes get
   //    reset.  So this has to be done sparingly.
-  //
-  //    If we do this, we should also look for any channel with the old
-  //    channel ID and delete it.  See zulip-mobile's `createNotificationChannel`
-  //    in android/app/src/main/java/com/zulipmobile/notifications/NotificationChannelManager.kt .
-  static Future<void> _ensureChannel() async {
+  @visibleForTesting
+  static Future<void> ensureChannel() async {
+    // See if our current-version channel already exists; delete any obsolete
+    // previous channels.
+    var found = false;
+    final channels = await _androidHost.getNotificationChannels();
+    for (final channel in channels) {
+      assert(channel != null); // TODO(#942)
+      if (channel!.id == kChannelId) {
+        found = true;
+      } else {
+        await _androidHost.deleteNotificationChannel(channel.id);
+      }
+    }
+
+    if (found) {
+      // The channel already exists; nothing to do.
+      return;
+    }
+
+    // The channel doesn't exist. Create it.
+
     await _androidHost.createNotificationChannel(NotificationChannel(
       id: kChannelId,
       name: 'Messages', // TODO(i18n)
@@ -81,7 +104,7 @@ class NotificationDisplayManager {
     if (launchDetails?.didNotificationLaunchApp ?? false) {
       _handleNotificationAppLaunch(launchDetails!.notificationResponse);
     }
-    await NotificationChannelManager._ensureChannel();
+    await NotificationChannelManager.ensureChannel();
   }
 
   static void onFcmMessage(FcmMessage data, Map<String, dynamic> dataJson) {
--- a/test/model/binding.dart
+++ b/test/model/binding.dart
@@ -559,6 +559,11 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
     _createdChannels.add(channel);
   }
 
+  @override
+  Future<List<NotificationChannel?>> getNotificationChannels() async {
+    return _createdChannels.toList(growable: false);
+  }
+
   /// Consume the log of calls made to [notify].
   ///
   /// This returns a list of the arguments to all calls made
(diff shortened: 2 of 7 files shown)


test/model/binding.dart
```diff
@@ -554,20 +554,32 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
   }
   List<NotificationChannel> _createdChannels = [];
 
+  /// Consumes the log of sequence of calls made to [getNotificationChannels],
+  /// [deleteNotificationChannel] and [createNotificationChannel].
```
chrisbobbe, line 558: "log of sequence of calls" -> "log of calls"?

lib/notifications/display.dart
```diff
@@ -52,11 +56,28 @@ class NotificationChannelManager {
...
+    // See if our current-version channel already exists; delete any obsolete
+    // previous channels.
+    var found = false;
+    final channels = await _androidHost.getNotificationChannels();
+    for (final channel in channels) {
+      assert(channel != null); // TODO(#942)
+      if (channel!.id == kChannelId) {
+        found = true;
+      } else {
+        await _androidHost.deleteNotificationChannel(channel.id);
```
chrisbobbe, line 70: It sounds like it would be helpful to mention this in the dartdoc; something like:

```dart
  /// Create our notification channel, if it doesn't already exist.
  /// 
  /// Deletes obsolete channels, if present, from old versions of the app.
```

lib/notifications/display.dart
```diff
@@ -25,6 +25,10 @@ AndroidNotificationHostApi get _androidHost => ZulipBinding.instance.androidNoti
 
 /// Service for configuring our Android "notification channel".
 class NotificationChannelManager {
+  /// The channel ID we use for our one notification channel, which we use for
+  /// all notifications.
+  // TODO(?) at launch make sure this channel-id doesn't collide with the one
+  // present in zulip-mobile.
```
chrisbobbe, line 31: More accurately, I think, the collision to avoid is with a channel ID that's active in the installed app the user will be upgrading from. In theory, that could be different from the one zulip-mobile has in `main`, right, because the user could be upgrading from a very old zulip-mobile straight to zulip-flutter?

I see in `main` zulip-mobile has "messages-3". There's a handy list of previous values in a code comment on it:

```kotlin
/** The channel ID we use for our one notification channel, which we use for all notifications. */
// Previous values: "default", "messages-1", (alpha-only: "messages-2")
val CHANNEL_ID = "messages-3"
```

which…hmm, that list does include "messages-1". I'm not sure how old that one is, and how many of our users will still have that as their active channel.

Rather than tracking that down, maybe simplest to be conservative and just say something like

```dart
  // TODO(launch) check this doesn't match zulip-mobile's current or previous channel IDs
```

chrisbobbe (commented):
Thanks! Small comment below, and marking for Greg's review.

rajveermalviya:
Thanks for the review @chrisbobbe! Pushed a new revision PTAL.

chrisbobbe:
Thanks! LGTM, over to @gnprice.

lib/notifications/display.dart
```diff
@@ -25,6 +25,10 @@ AndroidNotificationHostApi get _androidHost => ZulipBinding.instance.androidNoti
 
 /// Service for configuring our Android "notification channel".
 class NotificationChannelManager {
+  /// The channel ID we use for our one notification channel, which we use for
+  /// all notifications.
+  // TODO(launch) check this doesn't match zulip-mobile's current or previous
+  // channel IDs
```
gnprice, line 31: nit: in a TODO comment (or other labeled comment), indent any continuation lines to keep them grouped under the heading:
```suggestion
  // TODO(launch) check this doesn't match zulip-mobile's current or previous
  //   channel IDs
```

test/notifications/display_test.dart
```diff
@@ -131,6 +131,59 @@ void main() {
             NotificationChannelManager.kVibrationPattern)
       ;
     });
+
+    test('channel is not recreated if one with same id already exists', () async {
+      await NotificationChannelManager.ensureChannel();
```
gnprice, line 136: It looks like these tests affect the state in the binding, and then leak that state. To fix that, they should start with `addTearDown(testBinding.reset)` (like the existing smoke test does via `init`).

test/notifications/display_test.dart
```diff
@@ -131,6 +131,59 @@ void main() {
...
+      check(testBinding.androidNotificationHost.takeChannelMethodCallLogs())
+        .deepEquals(['getNotificationChannels', 'createNotificationChannel']);
+
+      await NotificationChannelManager.ensureChannel();
```
gnprice, line 140: It looks like the first `ensureChannel` here is to simulate a channel having gotten created in the past, and then the second is the call that we're testing the behavior of.

I think it'd be cleaner to do the setup more directly, without `ensureChannel`. Using `ensureChannel` will only produce a channel that's exactly the way the current code sets it up; this breaks down in the next test case, where we want to simulate some previous version of the app having set it up, and it also breaks down with the possibility that the user may have altered the settings on the channel. (If they have, we should leave the existing channel in place; only the ID counts for saying it's the same channel.)

test/model/binding.dart
```diff
@@ -554,8 +554,32 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
...
+    final result = _channelMethodCallLogs;
+    _channelMethodCallLogs = [];
+    return result;
+  }
+  List<String> _channelMethodCallLogs = [];
+
+  @override
+  Future<List<NotificationChannel?>> getNotificationChannels() async {
+    _channelMethodCallLogs.add('getNotificationChannels');
+    return _createdChannels.toList(growable: false);
```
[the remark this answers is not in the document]
gnprice: I see, what the tests effectively do is call `createNotificationChannel` in order to simulate a channel having been created in the distant past.

That works fine. But this approach changes the semantics of `_createdChannels` — particularly because `deleteNotificationChannel` now removes channels from that list. Instead of a log of what was created, it's really serving now as a list of the channels that currently exist. The name, and the "take"-based API which is suited for a log, are no longer a fit.

So let's have a prep commit that replaces `takeCreatedChannels` with a getter like:
```dart
  Iterable<NotificationChannel> get createdChannels => _createdChannels;
```

Then that's well adapted for the direction it gets extended in the rest of this PR.

test/notifications/display_test.dart
```diff
@@ -131,6 +131,59 @@ void main() {
...
+      check(testBinding.androidNotificationHost.takeChannelMethodCallLogs())
+        .deepEquals(['getNotificationChannels']);
```
gnprice, line 142: I'm generally not a fan of having "get" calls in a log that a test checks like this — it feels too far into testing the implementation details (cf https://zulip.readthedocs.io/en/stable/testing/philosophy.html#integration-testing-or-unit-testing ).

(We do routinely include the equivalent in our API tests — we check what HTTP requests the app made to the simulated server. But because these queries are local to the device, I'd like to try to treat them more as internal.)

Can we leave these out of the log, and have the tests just look at what the code does with the information? I.e., the `create` and `delete` calls.

gnprice (commented):
Thanks @rajveermalviya, and thanks @chrisbobbe for the previous review!

Generally this looks great. Various comments below.

rajveermalviya:
Thanks for the reviews @chrisbobbe and @gnprice!
@gnprice pushed a new revision with update test bindings. PTAL.

gnprice:

PR #982, opened 2024-10-04
notif: Use Zulip's custom notification sound on Android

rajveermalviya:
(Stacked on top of #981)
(Split from #717)
Fixes: #340


PIG208 (commented):
Tested this on my device and this works well. Thanks! Left some comments on the use of "URL".

rajveermalviya:
Thanks for the review @PIG208! Addressed them in the new revision.

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
```diff
@@ -69,6 +85,87 @@ private class AndroidNotificationHost(val context: Context)
...
+
+        val sounds = mutableListOf<StoredNotificationsSound>()
+        val query = context.contentResolver.query(
+            collection,
+            projection,
+            selection,
+            selectionArgs,
+            sortOrder,
+        )
+        query?.use { cursor ->
```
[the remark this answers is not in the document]
gnprice: > inserting duplicates, which could then pile up.

(This shouldn't happen, because once we successfully create the channel we won't be entering this whole sound-file-management area of code on future launches… but once things are going wrong in our interaction with the platform, maybe there's also a failure that we hit before the point of successfully creating the channel.)

lib/notifications/display.dart
```diff
@@ -39,6 +57,104 @@ class NotificationChannelManager {
...
+
+    final soundsToAdd = NotificationSound.values.toList();
+    final storedSounds = await _androidHost.listStoredSoundsInNotificationsDirectory();
+    for (final storedSound in storedSounds) {
+      assert(storedSound != null); // TODO(#942)
+
+      // If the file is one we put there, and has the name we give to our
+      // default sound, then use it as the default sound.
+      if (storedSound!.fileName == kDefaultNotificationSound.fileDisplayName
+          && storedSound.isOwner) {
```
gnprice, line 117:
```suggestion
          && storedSound.isOwned) {
```

Or could call it `ownedByThisPackage`.

Just "is owner", as a field on a sound, reads as if the sound is the owner of something, or an owner of something. (Even though it's not clear what that would mean.)

lib/notifications/display.dart
```diff
@@ -39,6 +57,104 @@ class NotificationChannelManager {
...
+      // that sound.  This applies even if we didn't put it there: the
+      // name is taken, so if we tried adding it anyway it'd get some
+      // other name (like "Zulip - Chime (1).m4a", with " (1)" added).
+      // Which means the *next* launch would try to add it again ad infinitum.
+      // We could avoid this given some other way to uniquely identify the
+      // file, but haven't found an obvious one.
+      //
+      // This does mean it's possible the file isn't the one we would have
+      // put there... but it probably is, just from a debug vs. release build
+      // of the app (because those have different package names).  And anyway,
```
gnprice, line 131:
```suggestion
      // of the app (because those may have different package names).  And anyway,
```

Currently zulip-flutter differs from zulip-mobile in this respect. We might change that in the future, though, so the remark can remain, just needs this hedge.

lib/notifications/display.dart
```diff
@@ -39,6 +57,104 @@ class NotificationChannelManager {
...
+      // this is a file we're supplying for the user in case they want it, not
+      // something where the app depends on it having specific content.
+      soundsToAdd.removeWhere((v) => v.fileDisplayName == storedSound.fileName);
```
gnprice, line 134:
```suggestion
      soundsToAdd.removeWhere((v) => v.fileDisplayName == storedSound.fileName);
```

Cool. The zulip-mobile version uses a hash map instead of a list… but given that this data structure has only a handful of elements in it anyway, there's no performance advantage to a map, so a list is fine (and is certainly simpler).

pigeon/notifications.dart
```diff
@@ -181,6 +208,32 @@ abstract class AndroidNotificationHostApi {
   /// See: https://developer.android.com/reference/kotlin/androidx/core/app/NotificationManagerCompat#deleteNotificationChannel(java.lang.String)
   void deleteNotificationChannel(String channelId);
 
+  /// Corresponds to `android.content.ContentResolver.query`.
+  ///
+  /// Returns the list of notification sounds present under
+  /// `Notifications/Zulip/` directory in device's shared media storage.
```
gnprice, line 214:
```suggestion
  /// The list of notification sound files present under `Notifications/Zulip/`
  /// in the device's shared media storage,
  /// found with `android.content.ContentResolver.query`.
  ///
  /// This is a complex ad-hoc method.
  /// For detailed behavior, see its implementation.
```

Unlike most of our Pigeon bindings, it's not really accurate to say this "corresponds" to any particular method or class in the underlying platform APIs. Rather it bundles together our own logic atop a bunch of underlying methods.

Generally that's something we try to avoid in these Pigeon bindings — our usual preferred approach is to keep them a thin layer that just exposes the underlying APIs, so that the real logic happens in our Dart code.

Here, for pragmatic reasons, it makes a lot of sense to instead have an ad-hoc bundle like this, just the way you did. I think the key point is that there's a lot of different underlying methods that would need bindings if we took our usual approach, compared to a much smaller interface that describes what we actually need. Or maybe even more important: the cursor needs to be a persistent object through all the successive `moveToNext` and `getString` etc. calls, and maintaining that across the Dart/Kotlin boundary would be annoying.

(If we were going to be dealing with this ContentResolver query stuff again and again, then it'd start making sense to do the work to provide thin bindings to it after all, like in our usual approach.)

So this structure is good but the doc just needs to match it.

gnprice (commented):
Thanks @rajveermalviya for building this, and @PIG208 for the previous review!

Generally this looks great. I appreciate how it closely follows the structure of the zulip-mobile implementation, right down to comments — all of that really helps in simplifying the review.

Comments below. Because there's a lot to read here, I haven't yet gone through the tests in the final/main commit; but I've read everything else.

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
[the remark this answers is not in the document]
rajveermalviya: > or perhaps throw an exception. Then the caller on the Dart side can see that and abort with `return defaultSoundUrl`.

Updated with this.

rajveermalviya:
Thanks for the review @gnprice! I've pushed the revision — PTAL.

lib/notifications/display.dart
```diff
@@ -39,6 +57,110 @@ class NotificationChannelManager {
...
+  static Future<String> _ensureInitNotificationSounds() async {
+    String defaultSoundUrl = _resourceUrlFromName(
+      resourceTypeName: 'raw',
+      resourceEntryName: kDefaultNotificationSound.resourceName).toString();
+
+    final shouldUseResourceFile = switch (await ZulipBinding.instance.deviceInfo) {
+      // Before Android 10 Q, we don't attempt to put the sounds in shared media storage.
+      // Just use the resource file directly.
+      // TODO(android-sdk-29): Simplify this away.
+      AndroidDeviceInfo(:var sdkInt) => sdkInt <= 28,
```
gnprice, line 101: nit:
```suggestion
      // Before Android 10 Q, we don't attempt to put the sounds in shared media storage.
      // Just use the resource file directly.
      // TODO(android-sdk-29): Simplify this away.
      AndroidDeviceInfo(:var sdkInt) => sdkInt < 29,
```

In general I think it's cleanest to express version conditions as a single threshold, which is the first version where the change appeared: so conditions look like "actual >= threshold" or "actual < threshold", but never "actual <= one-less-than-threshold".

That's because if things are instead expressed with two neighboring numbers (like 28 and 29 here), it starts getting harder to keep straight in one's head. I noticed this in the previous round, but was prompted to comment on it by my comment just before this one :slightly_smiling_face:

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
             NotificationChannelManager.kVibrationPattern);
     });
+
+    test('on Android 28 (and lower) resource file is used for notification sound', () async {
+      addTearDown(testBinding.reset);
+      final androidNotificationHost = testBinding.androidNotificationHost;
+
+      // Override android version
+      testBinding.deviceInfoResult =
+        const AndroidDeviceInfo(sdkInt: 28, release: '10');
```
gnprice, line 221: nit:
```suggestion
      testBinding.deviceInfoResult =
        const AndroidDeviceInfo(sdkInt: 28, release: '10');
```

The content of this comment is already expressed clearly enough by the code just below it.

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+      final soundUrl =
+        'android.resource://com.zulip.flutter/raw/$defaultSoundResourceName';
+      check(androidNotificationHost.takeCreatedChannels()).single
+        ..id.equals(NotificationChannelManager.kChannelId)
+        ..name.equals('Messages')
+        ..importance.equals(NotificationImportance.high)
+        ..lightsEnabled.equals(true)
+        ..soundUrl.equals(soundUrl)
+        ..vibrationPattern.isNotNull().deepEquals(
+            NotificationChannelManager.kVibrationPattern);
```
gnprice, line 239:
```suggestion
      check(androidNotificationHost.takeCreatedChannels()).single
        ..soundUrl.equals(soundUrl);
```

This is the only part of this check that differs between these new test cases, or between them and the corresponding check in the smoke test, right?

The rest of the details of this call aren't what these test cases are meant to focus on. So leaving them out helps the reader focus their attention on the part that the test is about.

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+      androidNotificationHost.setupStoredNotificationSounds(
+        NotificationSound.values.map((e) => StoredNotificationSound(
+          fileName: e.fileDisplayName,
+          isOwned: true,
+          contentUrl: androidNotificationHost.fakeStoredNotificationSoundUrl(e.resourceName)),
+        ).toList(),
+      );
+
+      await NotificationChannelManager.ensureChannel();
+      check(androidNotificationHost.takeCopySoundResourceToMediaStoreCalls()).length.equals(0);
```
gnprice, line 283: nit: line too long, with key information past 80 columns; also can simplify to `isEmpty`:

```suggestion
      check(androidNotificationHost.takeCopySoundResourceToMediaStoreCalls())
        .isEmpty();
```

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+        ..id.equals(NotificationChannelManager.kChannelId)
+        ..name.equals('Messages')
+        ..importance.equals(NotificationImportance.high)
+        ..lightsEnabled.equals(true)
+        ..soundUrl.equals(soundUrl)
+        ..vibrationPattern.isNotNull().deepEquals(
+            NotificationChannelManager.kVibrationPattern);
+    });
+
+    test('new notifications sounds are copied to media store', () async {
```
gnprice, line 299: nit: use standardized term:
```suggestion
    test('new notification sounds are copied to media store', () async {
```

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+      final androidNotificationHost = testBinding.androidNotificationHost;
+
+      // Emulate that except one sound, all other sounds are already in
+      // media store.
+      androidNotificationHost.setupStoredNotificationSounds(
+        NotificationSound.values.map((e) => StoredNotificationSound(
+          fileName: e.fileDisplayName,
+          isOwned: true,
+          contentUrl: androidNotificationHost.fakeStoredNotificationSoundUrl(e.resourceName)),
+        ).skip(1).toList()
```
gnprice, line 310: nit:

```suggestion
        NotificationSound.values.skip(1).map((e) => StoredNotificationSound(
          fileName: e.fileDisplayName,
          isOwned: true,
          contentUrl: androidNotificationHost.fakeStoredNotificationSoundUrl(e.resourceName)),
        ).toList()
```

otherwise the `skip(1)` gets a bit buried at the end

gnprice (commented):
Thanks! Those changes all look good. I've now read through the tests too; just small comments below.

Going AFK for dinner, so posting these now. The one remaining question I haven't thought through is to see if there are other cases I'd want tests for that aren't yet covered here.

test/model/binding.dart
```diff
@@ -549,6 +549,51 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
     _activeChannels.remove(channelId);
   }
 
+  /// Generates a fake URL for a notification sound present in media store.
+  String fakeStoredNotificationSoundUrl(String resourceName) {
```
gnprice, line 553:
```suggestion
  /// A URL that the fake [copySoundResourceToMediaStore] would produce
  /// for a resource with the given name.
  String fakeStoredNotificationSoundUrl(String resourceName) {
```

On my previous read I'd missed the connection of this to the `copySoundResourceToMediaStore` implementation (and in fact wondered if this might better go into `example_data.dart` instead). On rereading these tests I see the connection, and it makes perfect sense that this is the home of this method — and this connection is important context for a reader making sense of the tests that use this method.

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+
+      await NotificationChannelManager.ensureChannel();
+      final firstSound = NotificationSound.values.first;
+      check(androidNotificationHost.takeCopySoundResourceToMediaStoreCalls())
+        .single
+        ..sourceResourceName.equals(firstSound.resourceName)
+        ..targetFileDisplayName.equals(firstSound.fileDisplayName);
+
+      final defaultSoundResourceName =
+        NotificationChannelManager.kDefaultNotificationSound.resourceName;
```
gnprice, line 321: nit: pull this out as a local for a test group, so the individual cases don't have to repeat the definition

For organizing the tests, we can follow the pattern of the existing groups in this test file, like "NotificationDisplayManager show" and "NotificationDisplayManager open":
* rename the existing "NotificationChannelManager" group to a name like "NotificationChannelManager create channel"
* add a group "NotificationChannelManager sounds" as a home for the new tests

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+      final soundUrl =
+        androidNotificationHost.fakeStoredNotificationSoundUrl(defaultSoundResourceName);
```
gnprice, line 323: nit: then also pull out a helper for this:
```suggestion
      final soundUrl = fakeStoredUrl(defaultSoundResourceName);
```

just so it gets a much shorter name

test/notifications/display_test.dart
```diff
@@ -208,6 +211,158 @@ void main() {
...
+      check(androidNotificationHost.takeCreatedChannels()).single
+        ..id.equals(NotificationChannelManager.kChannelId)
+        ..name.equals('Messages')
+        ..importance.equals(NotificationImportance.high)
+        ..lightsEnabled.equals(true)
+        ..soundUrl.equals(soundUrl)
```
gnprice, line 329: Then in fact with that shorter name, we can comfortably inline the expression for the expected value into this check:
```suggestion
        ..soundUrl.equals(fakeStoredUrl(defaultSoundResourceName))
```

gnprice (commented):
OK, I've thought through that one remaining question from above and don't have anything to add — this already covers all the relevant cases I see.

(In principle I guess we might add test cases for situations like the `query` method on the resolver failing. But I don't know if that ever happens in practice; and conversely, testing it would require some added complication in the test binding. So I'm content to leave that untested, just relying on our reading the code and seeing that it looks right, and focus our energy on the remaining issues in our queue.)

A couple more small comments below, though, just as a result of having reread the tests more closely.

test/notifications/display_test.dart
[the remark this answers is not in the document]
rajveermalviya: Correct, also verified on the emulator that the reported `release` string is '9' (not 'Pie').

rajveermalviya:

PR #985, opened 2024-10-05
test [nfc]: Introduce "finish" helper for async checks

gnprice:
This brings the file test/api/core_test.dart up to being clean against the `unawaited_futures` lint rule, toward #731.

It does so with a smaller diff, and simpler resulting code, than by adding `await` on each of the futures.  For comparison:
  https://github.com/zulip/zulip-flutter/pull/934#discussion_r1788454470

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1113,7 +1113,7 @@ packages:
     source: hosted
     version: "1.25.8"
   test_api:
-    dependency: transitive
+    dependency: "direct dev"
     description:
       name: test_api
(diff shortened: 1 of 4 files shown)


PIG208:
Looks like there is an issue with the dependency:

   info • The imported package 'test_api' isn't a dependency of the importing package • test/test_async.dart:1:8 • depend_on_referenced_packages

gnprice:
Oops, thanks. I had a commit for that but it was missing from the PR branch — I think I must have accidentally dropped it when moving the changes to an independent branch, after developing them on top of #934.

Should be fixed now.

PIG208:
LGTM! Should be ready to merge then.

Issue #978, opened 2024-10-02
Update "bot" icon


gnprice:
When a message was sent by a bot, we put an icon by the sender's name to indicate it's a bot.

In the web app the icon we use was recently redesigned:
* https://github.com/zulip/zulip/pull/31673
* https://chat.zulip.org/#narrow/stream/101-design/topic/notification.20bot.20icon/near/1918735

to a friendlier-looking "bot" face with one antenna instead of two antennas. We should update to the new one-antenna icon.

Reported by @alexmv in chat: https://chat.zulip.org/#narrow/stream/48-mobile/topic/Bot.20icon.20asset.20needs.20update/near/1953361

fombalang:
Hi @gnprice 
I would like to work on this issue, please.

gnprice:
Sure, please go ahead.

fombalang:
Alright!

fombalang:
Hello
I have updated the bot icon.

![Screenshot_20241004-085226](https://github.com/user-attachments/assets/782db678-9e6e-412e-bebf-8ca7a2c52c68)

![Screenshot_20241004-085217](https://github.com/user-attachments/assets/6ce5a689-6a13-4eac-8157-1dd90f69fa97)

Before I send a PR for this, I would like to verify if a change like this would require that tests be made or updated? Seeing as I only updated the icon SVG file and ran the `tools/icons/build-icon-font` command without changing any Dart code?

Thank you

gnprice:
Cool ­— please go ahead and send the PR. This change will be fine without tests.

In general if you have a change that you think may not need tests, I recommend sending the PR without tests but then calling that question out clearly in the PR description. That helps speed things along if reviewers agree that tests aren't needed. It also means reviewers can see the rest of your changes, which can help answer the question of whether tests are needed.

fombalang:
Okay, Understood

PR #986, opened 2024-10-05
icons: Updated bot icon with single antenna version.

fombalang:
This PR updates the previous bot icon, replacing it with a new single antenna bot icon. 

No tests have been written or updated for this PR as the changes made were minimal and did not involve changing any dart code, just updating the SVG asset file and rebuilding the icon font.

Fixes: #978 

Thank you.


chrisbobbe:
Please write down specifically where you got the new icon from. The commit message is a good place for that, and the first place someone in the future will look when they want to know where it came from.

pubspec.lock
```diff
@@ -1060,10 +1060,10 @@ packages:
     dependency: "direct dev"
```
PIG208, line 1060: Looks like the update to this file is unintended.

fombalang:
> Please write down specifically where you got the new icon from. The commit message is a good place for that, and the first place someone in the future will look when they want to know where it came from.

Yes, the icon was gotten from the `zulip` repository from this PR about updating the icon. https://github.com/zulip/zulip/pull/31673
I downloaded the SVG file from that repository and used it.

pubspec.lock
fombalang: Yes, it was unintended. It was changed when I ran a `flutter pub get`

fombalang:
> > Please write down specifically where you got the new icon from. The commit message is a good place for that, and the first place someone in the future will look when they want to know where it came from.
> 
> Yes, the icon was gotten from the `zulip` repository from this PR about updating the icon. [zulip/zulip#31673](https://github.com/zulip/zulip/pull/31673) I downloaded the SVG file from that repository and used it. Specifically from https://github.com/zulip/zulip/blob/main/web/shared/icons/bot.svg

Should I send an updated commit with a description of where I got the icon from or adding a comment here would be fine? @chrisbobbe

chrisbobbe:
Yes, please update the commit. As I said above:

> […] The commit message is a good place for that, and the first place someone in the future will look when they want to know where it came from.

fombalang:
> Yes, please update the commit. As I said above:
> 
> > […] The commit message is a good place for that, and the first place someone in the future will look when they want to know where it came from.

Okay, understood, thank you.

fombalang:
I have updated the commit @chrisbobbe

assets/icons/bot.svg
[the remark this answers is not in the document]
fombalang: Okay, noted, thank you

chrisbobbe:
Ah, I see you've also wrapped the URL onto two lines :) but we do like to keep URLs intact instead of breaking them. So that can be an exception to the wrap-to-68-columns rule.

fombalang:
> Ah, I see you've also wrapped the URL onto two lines :) but we do like to keep URLs intact instead of breaking them. So that can be an exception to the wrap-to-68-columns rule.

Yeah, I was actually confused about what to do with URLs, let me update that then.

fombalang:
> > Ah, I see you've also wrapped the URL onto two lines :) but we do like to keep URLs intact instead of breaking them. So that can be an exception to the wrap-to-68-columns rule.
> 
> Yeah, I was actually confused about what to do with URLs, let me update that then.

Updated

chrisbobbe:
Ah, and one thing I forgot in my previous reviews: since this makes a visible change, please post before/after screenshots. The code and commit message look good to me though, and I'll move this along to Greg for his review (but please do post those screenshots).

chrisbobbe:
Oh, I see you did post screenshots, just on the issue thread instead of here: https://github.com/zulip/zulip-flutter/issues/978#issuecomment-2393209398

In future, please post the screenshots on the PR thread. But I do think those screenshots look good.

gnprice:
Thanks @fombalang for taking care of this, and @chrisbobbe @PIG208 for the previous reviews!

This looks good; merging. I fixed one nit in the commit message:

    -    icons: Update bot icon with single antenna version
    +    icons: Update bot icon with single-antenna version

PR #990, opened 2024-10-08
compose_box [nfc]: Add InsetShadowBox

PIG208:
This casts a static shadow on top of a child widget from its top edge and bottom edge.

To review its effect, see #928. This is intended to prep for #853.

### DIFF
--- a/lib/widgets/inset_shadow.dart
+++ b/lib/widgets/inset_shadow.dart
@@ -1,0 +1,59 @@
+import 'package:flutter/widgets.dart';
+
+/// A widget that overlays rectangular inset shadows on a child.
+///
+/// The use case of this is casting shadows on scrollable UI elements.
+/// For example, when there is a list of items, the shadows could be
+/// visual indicators for over scrolled areas.
+///
+/// Note that this is a bit different from the CSS `box-shadow: inset`,
+/// because it only supports rectangular shadows.
+///
+/// See also:
+///  * https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3860-11890&node-type=frame&t=oOVTdwGZgtvKv9i8-0
+///  * https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#inset
+class InsetShadowBox extends StatelessWidget {
+  const InsetShadowBox({
+    super.key,
+    this.top = 0,
+    this.bottom = 0,
+    required this.color,
+    required this.child,
+  });
+
+  /// The distance that the shadow from the child's top edge grows downwards.
+  ///
+  /// This does not pad the child widget.
+  final double top;
+
+  /// The distance that the shadow from the child's bottom edge grows upwards.
+  ///
+  /// This does not pad the child widget.
+  final double bottom;
+
+  /// The shadow color to fade into transparency from the top and bottom borders.
+  final Color color;
+
+  final Widget child;
+
+  BoxDecoration _shadowFrom(AlignmentGeometry begin) {
+    return BoxDecoration(gradient: LinearGradient(
+      begin: begin, end: -begin,
+      colors: [color, color.withValues(alpha: 0)]));
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      // This is necessary to pass the constraints as-is,
+      // so that the [Stack] is transparent during layout.
+      fit: StackFit.passthrough,
+      children: [
+        child,
+        Positioned(top: 0, height: top, left: 0, right: 0,
+          child: DecoratedBox(decoration: _shadowFrom(Alignment.topCenter))),
+        Positioned(bottom: 0, height: bottom, left: 0, right: 0,
+          child: DecoratedBox(decoration: _shadowFrom(Alignment.bottomCenter))),
+      ]);
+  }
+}
--- a/test/flutter_checks.dart
+++ b/test/flutter_checks.dart
@@ -6,6 +6,10 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
+extension PaintChecks on Subject<Paint> {
+  Subject<Shader?> get shader => has((x) => x.shader, 'shader');
+}
+
 extension RectChecks on Subject<Rect> {
   Subject<double> get top => has((d) => d.top, 'top');
   Subject<double> get bottom => has((d) => d.bottom, 'bottom');
--- a/test/widgets/inset_shadow_test.dart
+++ b/test/widgets/inset_shadow_test.dart
@@ -1,0 +1,64 @@
+import 'dart:ui' as ui;
+import 'package:checks/checks.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:legacy_checks/legacy_checks.dart';
+import 'package:zulip/widgets/inset_shadow.dart';
+
+import '../flutter_checks.dart';
+
+void main() {
+  testWidgets('constraints from the parent are not modified', (tester) async {
+    await tester.pumpWidget(const Directionality(
+      textDirection: TextDirection.ltr,
+      child: Align(
+        // Position child at the top-left corner of the box at (0, 0)
+        // to ease the check on [Rect] later.
+        alignment: Alignment.topLeft,
+        child: SizedBox(width: 20, height: 20,
+          child: InsetShadowBox(top: 7, bottom: 3,
+            color: Colors.red,
+            child: SizedBox.shrink())))));
+
+    // We expect that the child of [InsetShadowBox] gets the constraints
+    // from [InsetShadowBox]'s parent unmodified, so that the only effect of
+    // the widget is adding shadows.
+    final parentRect = tester.getRect(find.byType(SizedBox).at(0));
+    final childRect = tester.getRect(find.byType(SizedBox).at(1));
+    check(parentRect).equals(const Rect.fromLTRB(0, 0, 20, 20));
+    check(childRect).equals(parentRect);
+  });
+
+  testWidgets('render shadow correctly', (tester) async {
+    PaintPatternPredicate paintGradient({required Rect rect}) {
+      // This is inspired by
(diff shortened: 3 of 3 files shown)


lib/widgets/inset_shadow_box.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:flutter/material.dart';
+
+/// A widget that overlays inset shadows on a child.
```
chrisbobbe, line 3: Could you expand on this a little bit, maybe mentioning the scroll-through-shadow use case?

There's a similar-sounding effect that this looks like it _might_ describe, but it turns out it doesn't; that's `box-shadow: inset`. It would be nice to have a full implementation of `box-shadow: inset`—we have a TODO in the reaction-chip UI:

```dart
      // TODO shadow effect, following web, which uses `box-shadow: inset`:
      //   https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#inset
      //   Needs Flutter support for something like that:
      //     https://github.com/flutter/flutter/issues/18636
      //     https://github.com/flutter/flutter/issues/52999
      //   Until then use a solid color; a much-lightened version of the shadow color.
      //   Also adapt by making [borderUnselected] more transparent, so we'll
      //   want to check that against web when implementing the shadow.
      bgUnselected: const HSLColor.fromAHSL(0.08, 210, 0.50, 0.875).toColor(),
```

—but that's not what `InsetShadowBox` is. 🙂 (`InsetShadowBox` wouldn't handle the "stadium" shape of the reaction chips, for example.)

lib/widgets/inset_shadow_box.dart
```diff
@@ -0,0 +1,45 @@
...
+  Widget build(BuildContext context) {
+    return Stack(
+        children: [
+          child,
+          Positioned(top: 0, left: 0, right: 0,
+            child: Container(height: top, decoration: _shadowFrom(Alignment.topCenter))),
+          Positioned(bottom: 0, left: 0, right: 0,
+            child: Container(height: bottom, decoration: _shadowFrom(Alignment.bottomCenter))),
+        ]);
+  }
```
chrisbobbe, line 44:
```suggestion
  Widget build(BuildContext context) {
    return Stack(children: [
      child,
      Positioned(top: 0, left: 0, right: 0,
        child: Container(height: top, decoration: _shadowFrom(Alignment.topCenter))),
      Positioned(bottom: 0, left: 0, right: 0,
        child: Container(height: bottom, decoration: _shadowFrom(Alignment.bottomCenter))),
    ]);
  }
```

lib/widgets/inset_shadow_box.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:flutter/material.dart';
```
chrisbobbe, line 1: nit:

```suggestion
import 'package:flutter/widgets.dart';
```

(not sure how much we care about being agnostic about Material though.)

chrisbobbe (commented):
Thanks! Small comments below, and I'll mark for Greg's review.

lib/widgets/inset_shadow_box.dart
PIG208: Makes sense. I think this means that we should rename this widget to `RectangularInsetShadowBox`, to make it clearer.

lib/widgets/rectangular_inset_shadow_box.dart
```diff
@@ -0,0 +1,51 @@
...
+      begin: begin, end: -begin,
+      colors: [color, color.withValues(alpha: 0)]));
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(children: [
+      child,
+      Positioned(top: 0, left: 0, right: 0,
+        child: Container(height: top, decoration: _shadowFrom(Alignment.topCenter))),
```
gnprice, line 46: nit: Positioned can control the height, right?
```suggestion
      Positioned(top: 0, height: top, left: 0, right: 0,
        child: Container(decoration: _shadowFrom(Alignment.topCenter))),
```

Then also `Container` can I think be reduced to `DecoratedBox`.

lib/widgets/rectangular_inset_shadow_box.dart
```diff
@@ -0,0 +1,51 @@
...
+
+/// A widget that overlays rectangular inset shadows on a child.
+///
+/// The use case of this is casting shadow on scrollable UI elements.
+/// For example, when there is a list of items, the shadow could be
+/// a visual indicator for over scrolled areas.
+///
+/// See also:
+///  * https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3860-11890&node-type=frame&t=oOVTdwGZgtvKv9i8-0
+class RectangularInsetShadowBox extends StatelessWidget {
```
gnprice, line 11: "rectangular box" seems redundant — a box is necessarily rectangular. I think InsetShadowBox would be a fine name.

Or I guess this is a box with rectangular inset shadows (the shadows are rectangular). But I think an "inset shadow" on a box, which is a rectangle, is also going to be rectangular.

It's true this doesn't cover as wide a range of cases as CSS `box-shadow: inset`, as discussed at https://github.com/zulip/zulip-flutter/pull/990#discussion_r1792373624 . I think that reflects the CSS feature growing to be wider than its name fits. To clarify that this doesn't cover that CSS feature, the clearest thing would be to just mention it in the doc.

lib/widgets/rectangular_inset_shadow_box.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:flutter/widgets.dart';
```
gnprice, line 1: nit: generally we've followed the Flutter upstream tree in grouping related functionality together into source files, rather than atomizing each class into its own tiny file.

One consequence is that the file names are shorter and more general than the class names.

Here there might not be any existing other code for this to share a file with, but we can still give it a bit of a broader, shorter name. Say `inset_shadow.dart`.

gnprice (commented):

Generally this looks great. A few nits below.

Let's also have a test or two.
* One thing to test is that the shadows don't add padding — so e.g. a child with height 40 and shadows of height 8 produce height 40, not 56.
* Ideally I'd like to have a test for the widget's actual main functionality too. I'm not sure there's a good way to do that without more work than it's worth, though.
  * One way would be with golden images. That seems like definitely more work than it's worth here.
  * The other possible way would be a painting test. That'd be worth spending like 15 minutes investigating to try to get to work — it might not be hard.

PIG208:
Updated the PR with tests and fixes. Ready for review!

test/widgets/inset_shadow.dart
```diff
@@ -0,0 +1,57 @@
+import 'dart:ui' as ui;
+import 'package:checks/checks.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:legacy_checks/legacy_checks.dart';
+import 'package:zulip/widgets/inset_shadow.dart';
+
+import '../flutter_checks.dart';
+
+void main() {
```
gnprice, line 10: This test doesn't actually run in `flutter test` (or `tools/check`) because the name doesn't end with `_test.dart` :slightly_smiling_face:

test/widgets/inset_shadow.dart
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  testWidgets('render shadow correctly', (tester) async {
+    PaintPatternPredicate paintGradient({required Rect rect}) {
+      // This is inspired by
+      //   https://github.com/flutter/flutter/blob/7b5462cc34af903e2f2de4be7540ff858685cdfc/packages/flutter/test/cupertino/route_test.dart#L1449-L1475
+      return (Symbol methodName, List<dynamic> arguments) {
+        check(methodName).equals(#drawRect);
+        check(arguments[0]).isA<Rect>().equals(rect);
+        check(arguments[1]).isA<Paint>().shader.isA<ui.Gradient>();
```
gnprice, line 34: Worth a one-line comment saying that Gradient is opaque and so this is the most we can check about it.

test/widgets/inset_shadow.dart
```diff
@@ -0,0 +1,57 @@
...
+        // Position child at the top-left corner of the box at (0, 0)
+        // to ease the check on [Rect] later.
+        alignment: Alignment.topLeft,
+        child: SizedBox(width: 20, height: 20,
+          child: InsetShadowBox(top: 7, bottom: 3,
+            color: Colors.red,
+            child: Placeholder())))));
+
+    final childRect = tester.getRect(find.byType(Placeholder));
+    check(childRect).equals(const Rect.fromLTRB(0, 0, 20, 20));
```
gnprice, line 24: I believe this check would pass no matter what `InsetShadowBox` attempted to do for its layout. Its parent (the `SizedBox`) has decreed exactly what its size is, so it's not able to be any other size. See:
https://docs.flutter.dev/get-started/fundamentals/layout#constraints
https://api.flutter.dev/flutter/rendering/RenderBox-class.html#layout

It's worth trying to verify that — edit InsetShadowBox so that it does use `top` and `bottom` as padding, and see if this test still passes. I believe it will.

To make the test effective, I think what's needed is to have a `SizedBox` as the child, not the parent. Then do confirm that that test fails if you make that change to InsetShadowBox.

test/widgets/inset_shadow.dart
```diff
@@ -0,0 +1,57 @@
...
+    await tester.pumpWidget(const Directionality(
+      textDirection: TextDirection.ltr,
+      child: Align(
+        // Position child at the top-left corner of the box at (0, 0)
+        // to ease the check on [Rect] later.
+        alignment: Alignment.topLeft,
+        child: SizedBox(width: 20, height: 20,
+          child: InsetShadowBox(top: 7, bottom: 3,
+            color: Colors.red,
+            child: Placeholder())))));
```
gnprice, line 21: In tests, typically preferable to say something like `SizedBox.shrink()` instead of `Placeholder`. The Placeholder widget has a more complicated behavior that's useful for making it visible in development, but in a test makes the behavior more complex to reason through.

gnprice (commented):

test/widgets/inset_shadow.dart
PIG208: We actually caught a bug with the implementation that allows `InsetShadowBox` to modify the constraints from the parent. The test is updated to catch layout changes.

gnprice:

PR #993, opened 2024-10-11
actions test: Await all responses; cut pumpAndSettle

gnprice:
This ensures that if one of these markNarrowAsRead calls were to change behavior so that it no longer completed within the test (and then possibly threw an error, or failed to ever complete at all), the test would catch that.

Prompted by this thread:
https://github.com/zulip/zulip-flutter/pull/934#discussion_r1796319431

### DIFF
--- a/test/widgets/actions_test.dart
+++ b/test/widgets/actions_test.dart
@@ -39,8 +39,7 @@ void main() {
 
     await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
       child: const Scaffold(body: Placeholder())));
-    // global store, per-account store get loaded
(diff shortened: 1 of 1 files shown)


test/widgets/actions_test.dart
```diff
@@ -52,8 +52,9 @@ void main() {
         processedCount: 11, updatedCount: 3,
```
PIG208, line 52: There is one more call-site of `pumpAndSettle` from `prepare` that can be cut too.

PIG208:
Looks good! Just one small comment.

gnprice:
Thanks for the review! Merging with that tweak.

PR #994, opened 2024-10-11
deps: Make flutter_checks and legacy_checks direct dev dependencies

PIG208:
They are only used in tests.

I was reminded of this by https://github.com/zulip/zulip-flutter/pull/917#discussion_r1796280646

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -421,7 +421,7 @@ packages:
     source: sdk
     version: "0.0.0"
   flutter_checks:
-    dependency: "direct main"
+    dependency: "direct dev"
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks for cleaning that up! Merged with a tweak to make the commit message more specific:

    -    deps: Make flutter_checks and legacy_checks direct dev dependencies
    +    deps: Reduce flutter_checks and legacy_checks to dev dependencies

so that it's clear which direction the change is happening in — they used to be (direct) full dependencies, vs. used to be indirect (dev) dependencies.

Issue #913, opened 2024-08-23
Redesign for @-mention autocomplete items


gnprice:
From @terpimost we have a spiffy new design for how the result items in @-mention autocomplete should look:
![image](https://github.com/user-attachments/assets/0663e07a-8858-4c0b-bd49-e21e841eee58)

Detailed specs are in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3859-3131&m=dev

This issue is only about the design for each item. Out of scope is how the whole list of autocomplete results is presented; I'll file a separate related issue for that.

fombalang:
Hello everyone.
I started working on this issue

**Before**
![before](https://github.com/user-attachments/assets/3cd448c4-4d8b-4180-b09b-32921e5c99d8) 

**After**
![after](https://github.com/user-attachments/assets/128eb401-66f8-47da-a1dd-e5b601641bee)

I had some worries I would like to have clarified please.
The design file says the color token for the metaData (email) field is called `contextMenuItemMeta`, but there is no `contextMenuItemMeta` token in the `DesignVariables` so I used the `designVariables.labelEdited` color token instead.

Also, according to the design, the metaData can be constructed from various user information, please I would like to find out what variables exactly will be used for the metaData field? At the moment I am using the `users[userId]!.email` variable, which seems to be the user's Zulip specific email, not sure if that works?

Any feedback would be much appreciated, thank you.
@gnprice @chrisbobbe

gnprice:
I made a chat thread to answer those questions:
https://chat.zulip.org/#narrow/stream/516-mobile-dev-help/topic/redesign.20for.20.40-mention.20autocomplete.20items/near/1958826

For future questions like these — questions about how a feature should work, or about how to develop a change — I recommend asking in a public channel in chat.zulip.org. That'll typically get you a faster response, and the discussion can be helpful for others reading it too. #mobile-dev-help is a good choice any time you're not sure which channel to use.

fombalang:
> I made a chat thread to answer those questions: https://chat.zulip.org/#narrow/stream/516-mobile-dev-help/topic/redesign.20for.20.40-mention.20autocomplete.20items/near/1958826
> 
[...]

Issue #914, opened 2024-08-23
Redesign for autocomplete result lists


gnprice:
From @terpimost we have a new design for how the autocomplete result lists should be presented:
![image](https://github.com/user-attachments/assets/23a5e945-0b8b-4910-9609-6da6b87e421c)

Details are in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3954-13395&m=dev

The design within each item in the results is a separate related issue:
* #913 

This issue is about how the result list as a whole is presented:
* the fading effect at the top and/or bottom when scrolled
* the list's height, including not going off the top of the screen (I think we may already successfully avoid this)
* its rounded corners, border, and shadow
* its placement (dipping vertically a few pixels into the top of the text field, and aligning horizontally with the width of messages in the message list)
* any other details I'm missing right now

This should apply to autocomplete both for the content input, and for the topic input.

chrisbobbe:
Hi, welcome. This issue has the M6 Post-launch milestone, which means the core team [isn't going to be able](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556) to spend time in the next couple of months to review or help with work on it. If you'd like to contribute to Zulip, see the guide in our README and the resources linked from there: https://github.com/zulip/zulip-flutter#picking-an-issue-to-work-on

PR #995, opened 2024-10-11
autocomplete: Implement new design for @-mention autocomplete items

fombalang:
Hello everyone,

This PR implements the updated design for the @-mention autocomplete items as detailed in this Figma File
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3859-3131&m=dev

I added new `contextMenuItemLabel` and `contextMenuItemMeta` color variables to the `designVariables` class from the design.
I also had to manually set the line height to match the design exactly.

As Greg described [here](https://chat.zulip.org/#narrow/stream/516-mobile-dev-help/topic/redesign.20for.20.40-mention.20autocomplete.20items/near/1958830), the autocomplete item only shows the user's real email address when the user has allowed their email to be visible, if not it doesn't show anything. I used the `_getDisplayEmailFor()` method from the profile page to get the user's real email. I am not sure if it is a good idea to have a copy of the same method here, maybe we should find a way to not have multiple copies of the same method.

Screenshot_20241011-114402

I updated autocomplete tests to also check if the email(metaData) is also shown as the updated design now shows extra metadata.

I attempted to add another test to verify if the email is not shown when the user's `emailAddressVisibility != emailAddressVisibility.everyone`, while doing so I learned a bit more about how the app manages users and data, but to be honest I couldn't get it to work properly, and I did not want to massively change up the current test file. Still trying to wrap my head around that.
This also leads me to a bug I think, which I discovered in the `example_data.dart`  `initialSnapshot` method where some of the arguments from the `intialSnapshot` method aren't being passed down to the method, I think someone missed that when setting up the method.

Please review when you can, thank you

Fixes: #913


lib/widgets/autocomplete.dart
```diff
@@ -193,14 +198,47 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
     );
   }
 
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if self-user isn't able to see [user]'s real email address.
+  String? _getDisplayEmailFor(User user, {required PerAccountStore store}) {
```
chrisbobbe, line 204: Yes, as you suggested in the PR description, we should avoid copy-pasting this complicated logic from somewhere else. If we have some complicated logic written twice, and it needs to be updated, there's a risk that we'll update one place and forget the other.

Maybe a good place for a reusable function is lib/api/model/model.dart, below the `User` class? That refactor should be in an [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#user-content-nfc) prep commit.

lib/widgets/autocomplete.dart
```diff
@@ -193,14 +198,47 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
+    String metaData;
+
     switch (option) {
       case UserMentionAutocompleteResult(:var userId):
+        final store = PerAccountStoreWidget.of(context);
+        final user = store.users[userId];
         avatar = Avatar(userId: userId, size: 32, borderRadius: 3);
-        label = PerAccountStoreWidget.of(context).users[userId]!.fullName;
+        label = user!.fullName;
+        metaData = _getDisplayEmailFor(user, store: store) ?? '';
```
chrisbobbe, line 241: nit: Since "metadata" is one word, the variable should be called `metadata`.

lib/widgets/autocomplete.dart
```diff
@@ -193,14 +198,47 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
   Widget buildItem(BuildContext context, int index, MentionAutocompleteResult option) {
+    final designVariables = DesignVariables.of(context);
     Widget avatar;
     String label;
+    String metaData;
+
     switch (option) {
       case UserMentionAutocompleteResult(:var userId):
+        final store = PerAccountStoreWidget.of(context);
+        final user = store.users[userId];
```
chrisbobbe, line 238:
```suggestion
        final user = store.users[userId]!;
```

instead of putting the `!` later, when we try to access `.fullName`.

lib/widgets/autocomplete.dart
chrisbobbe: Also, let's make it nullable, so `String? metadata;`, so its empty value is `null` instead of the empty string.

lib/widgets/autocomplete.dart
```diff
@@ -212,8 +250,16 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
+            Column(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+              Text(label, style: TextStyle(fontSize: 18,  height: 1.1, color: designVariables.contextMenuItemLabel) //Creates a line height equavalent to ~20
+                  .merge(weightVariableTextStyle(context, wght: 600))),
+              Visibility(
+                visible: metaData.isNotEmpty,
+                child: Text(metaData, style: TextStyle(height: 1.14, color: designVariables.contextMenuItemMeta)), //Creates a line height equavalent to ~16
+              )])])));
```
chrisbobbe, line 262: indentation and line lengths:

```suggestion
            Column(
              mainAxisSize: MainAxisSize.min,
              crossAxisAlignment: CrossAxisAlignment.start,
              children: [
                Text(
                  style: TextStyle(
                    fontSize: 18,
                    height: 1.1, // Creates a line height equavalent to ~20
                    color: designVariables.contextMenuItemLabel,
                  )
                    .merge(weightVariableTextStyle(context, wght: 600)),
                  label),
                Visibility(
                  visible: metaData.isNotEmpty,
                  child: Text(
                    style: TextStyle(
                      height: 1.14, // Creates a line height equavalent to ~16
                      color: designVariables.contextMenuItemMeta,
                    ),
                    metaData)),
              ]),
          ])));
```

lib/widgets/autocomplete.dart
```diff
@@ -212,8 +250,16 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
           children: [
             avatar,
             const SizedBox(width: 8),
-            Text(label),
-          ])));
+            Column(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+              Text(label, style: TextStyle(fontSize: 18,  height: 1.1, color: designVariables.contextMenuItemLabel) //Creates a line height equavalent to ~20
```
chrisbobbe, line 257: See other places where we set `height`; here, we would express it as `20 / 18`.

lib/widgets/autocomplete.dart
```diff
@@ -212,8 +250,16 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
+                  .merge(weightVariableTextStyle(context, wght: 600))),
+              Visibility(
+                visible: metaData.isNotEmpty,
+                child: Text(metaData, style: TextStyle(height: 1.14, color: designVariables.contextMenuItemMeta)), //Creates a line height equavalent to ~16
```
chrisbobbe, line 261: `height: 16 / 14`

chrisbobbe: Also, this should have an explicit `fontSize: 14`. Even if that matches the default, let's be explicit here.

lib/widgets/autocomplete.dart
```diff
@@ -212,8 +250,16 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
             const SizedBox(width: 8),
-            Text(label),
-          ])));
+            Column(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+              Text(label, style: TextStyle(fontSize: 18,  height: 1.1, color: designVariables.contextMenuItemLabel) //Creates a line height equavalent to ~20
+                  .merge(weightVariableTextStyle(context, wght: 600))),
+              Visibility(
```
chrisbobbe, line 259: Instead of `Visibility`, how about using Dart's ["collection if" feature](https://dart.dev/language/collections#control-flow-operators):

```dart
if (metaData.isNotEmpty) Text(/* … */),
```

, but combined with my earlier suggestions it would be

```dart
if (metadata != null) Text(/* … */),
```

chrisbobbe (dismissed):
Thanks! Comments below.

In the Figma:

- The gap between the avatar photo and the text is 6px, not 8px
- The outer padding is top: 4, right: 8, bottom: 4, left: 4 (not top: 8, right: 16, bottom: 8, left: 16) Since left and right are unequal, we should use `EdgeInsetsDirectional` for this.
- Please post screenshots showing the "…" effect when the name or metadata text are very long (not sure this is implemented in this revision)
- The avatar photo is a 36px rounded square with 4px border radius (not a 32px square with 3px border radius)
- Our designer doesn't want the "ink splash" effect on buttons. We should remove the `InkWell` widget and implement the on-pressed appearance specified here:
    https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3732-28939&node-type=symbol&m=dev
  image

gnprice:
> This also leads me to a bug I think, which I discovered in the `example_data.dart` `initialSnapshot` method where some of the arguments from the `intialSnapshot` method aren't being passed down to the method, I think someone missed that when setting up the method.

Sure, this would be good to fix — please go ahead and add a commit with that fix if you think you already see how to do it, or start a thread in #mobile-team and we can discuss in more detail.

fombalang:
Okay, thank you for the review, I will make the changes.
For the changes such as the bug and the refactor of the email logic, should I add those into this PR, or should it be in a different PR?

gnprice:
If it's needed for the other work you're doing in this PR, then a commit in this PR is a good way to do it.

If it's not related to what's in this PR, then a separate PR is best.

fombalang:
> If it's needed for the other work you're doing in this PR, then a commit in this PR is a good way to do it.
> 
> If it's not related to what's in this PR, then a separate PR is best.

Okay

fombalang:
> * Our designer doesn't want the "ink splash" effect on buttons. We should remove the `InkWell` widget and implement the on-pressed appearance specified here:

About the splash, I kept the inkwell widget, because a gesture detector doesn't show splashes, but I removed the sparkle effect, and left just the color change on tap. I also added a border radius in the splash as shown on the design, it's a bit difficult to see here but it is there. Would this work according to the design?

https://github.com/user-attachments/assets/9c190f0c-c373-4cbb-85f9-7def3b56611a

fombalang:
I also did the updates for the case where the user has a really long name or email.

**Before**
Screenshot_20241014-191455

**After**
Screenshot_20241014-191948

chrisbobbe:
> Would this work according to the design?

If your code causes any noticeable differences from the design, please identify them and say why you think they are OK. This will shorten the path to deciding if the code will work. :)

It looks like you haven't pushed your latest revision to GitHub. When it's ready for another review, please do that, and comment saying that it's ready for review.

fombalang:
> It looks like you haven't pushed your latest revision to GitHub. When it's ready for another review, please do that, and comment saying that it's ready for review.

No, I haven't yet.

That is because I have been facing some problems with the `ComposeAutocomplete`  test, I found that changing the dimensions of the avatar widget from 32 to 36 causes the test to fail. Not only that I found that there was a threshold of sizes for which the test passed, 32 -33 are good, and from 34 upward the test fails.

And I have looked at the test, and tried to do some searching as to why it might be failing, and I haven't succeeded. I did not see any obvious place where the avatar widget size was checked in the test. I also thought it might be because of an outdated golden test snapshot, but I could not find any references to a snapshot file. So I have been confused and stumped.

I was just about to comment on that, wondering if anyone can help point me in the right direction.

gnprice:
If you push the code here, and then start a conversation in #mobile-dev-help with the test failure you're seeing, that will probably be the most efficient way to get help investigating it.

fombalang:
> If you push the code here, and then start a conversation in #mobile-dev-help with the test failure you're seeing, that will probably be the most efficient way to get help investigating it.

Okay, I will do that

fombalang:
> It looks like you haven't pushed your latest revision to GitHub. When it's ready for another review, please do that, and comment saying that it's ready for review.

I have pushed my changes, please review

fombalang:
I noticed that some other tests failed on the pipeline, some database tests to be specific.

```
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/database_test.dart: non-migration tests create account (failed)
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/database_test.dart: non-migration tests create account with same realm and userId  (failed)
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/database_test.dart: non-migration tests create account with same realm and email (failed)
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/database_test.dart: migrations upgrade to v2, empty (failed)
❌ /home/runner/work/zulip-flutter/zulip-flutter/test/model/database_test.dart: migrations upgrade to v2, with data (failed)
``` 

These tests didn't fail before, and they also passed locally.

I think it is related to the migration to `Ubuntu 24.04` which apparently seems to have been causing a lot of issues and is now being rolled back.
https://github.com/actions/runner-images/issues/10636

chrisbobbe:
Thanks for pointing out those unrelated test failures. I've run CI again and those failures are gone; see https://github.com/zulip/zulip-flutter/pull/999#issuecomment-2417351945

Some failures remain and are related to the changes in this PR; I see a discussion in #mobile-dev-help about that.

chrisbobbe:
> If you push the code here, and then start a conversation in #mobile-dev-help with the test failure you're seeing, that will probably be the most efficient way to get help investigating it.

Thanks for starting that conversation. Here's a link, so people reading this will know the conversation is happening and where to find it: https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/ComposeAutocomplete.20test.20failure.20help/near/1963058

fombalang:
> If it's needed for the other work you're doing in this PR, then a commit in this PR is a good way to do it.
> 
> If it's not related to what's in this PR, then a separate PR is best.

Hello
I have created a PR for this bug, please review when you can.
https://github.com/zulip/zulip-flutter/pull/1011

chrisbobbe:
> #1011

Thanks; I'll add this to my review queue. 🙂

I see Greg has replied in that [discussion](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/ComposeAutocomplete.20test.20failure.20help/near/1963852) about the test failures. Please comment here when those failures are resolved and this PR is ready for another review. 🙂

gnprice:
It's puzzling that GitHub didn't run the CI checks on the latest revision.

It's not because the commits are the same; they're different. Even though the last commit has the same changes in it as the last commit of the previous revision, it's still a different commit as far as Git is concerned because it has a different history. You can see here:
image

> fombalang force-pushed the autocomplete-results-list branch from e107e03 to 5b85e89 1 hour ago

that the latest revision's tip commit is 5b85e89 and the revision before that was e107e03.

Anyway, we can run the tests locally, and as long as they pass when run locally then that confirms that they pass.

gnprice:
I tried the "dismiss review" button on Chris's previous "request changes" review, just in case that was related to why CI wasn't running. Doesn't seem like it had an effect — still not running.

I don't even have a button to tell checks to run. Instead the "Checks" tab just looks like this:
image

> Workflow runs completed with no jobs

It's as if there weren't any files under `.github/workflows/` at all. But in reality there is; your branch doesn't touch that directory, so it has the same workflow file in it as it has in main.

:shrug: We'll just rely on running checks locally, then. If this odd GitHub behavior starts happening repeatedly, we'll have to investigate further.

fombalang:
> that the latest revision's tip commit is [5b85e89](https://github.com/zulip/zulip-flutter/commit/5b85e89a75bd9e62125ef7604ea2e3d4049864e8) and the revision before that was [e107e03](https://github.com/zulip/zulip-flutter/commit/e107e03dead4c69abdfa6aaaa7a5577c2450f853).

Okay, I see. which means it should have triggered again.

fombalang:
> 🤷 We'll just rely on running checks locally, then. If this odd GitHub behavior starts happening repeatedly, we'll have to investigate further.

Okay then.
Everything passes locally on my end.

image

chrisbobbe:
The GitHub UI is showing that there are some rebase conflicts:

image

Could you please resolve those and comment here when that's done? If you need help, please ask in `#git help` in the [development community](https://zulip.com/development-community/).

fombalang:
> development community

I can just resolve using the GitHub UI right? or would that add an unwanted commit?

lib/widgets/autocomplete.dart
```diff
@@ -193,27 +196,61 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
     );
   }
 
+
```
chrisbobbe, line 199: nit: don't add blank line

lib/widgets/autocomplete.dart
```diff
@@ -193,27 +196,61 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
       onTap: () {
         _onTapOption(context, option);
       },
+      highlightColor: designVariables.editorButtonPressedBg,
+      splashFactory: NoSplash.splashFactory,
+      splashColor: Colors.transparent,
+      borderRadius: BorderRadius.circular(8),
       child: Padding(
-        padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
+        padding: const EdgeInsets.only(left: 4, right: 8, top: 4, bottom: 4),
```
chrisbobbe, line 224: If left and right are unequal, use `EdgeInsetsDirectional` instead of `EdgeInsets`.

lib/widgets/autocomplete.dart
```diff
@@ -193,27 +196,61 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
+                ),
+                if(metadata != null) Text(
+                    style: TextStyle(
+                        fontSize: 14,
+                        height: 16/14,
+                        color: designVariables.contextMenuItemMeta),
+                    overflow: TextOverflow.ellipsis,
+                    maxLines: 1,
+                    metadata,
+                )]))])));
```
chrisbobbe, line 253: Clean up indentation and whitespace to make this easy to read; look around the codebase to see how it should look.

test/widgets/autocomplete_test.dart
```diff
@@ -108,37 +108,39 @@ void main() {
   group('ComposeAutocomplete', () {
 
     Finder findNetworkImage(String url) {
+      String trimmedUrl  = url.substring(0, url.length -4); //Remove the extension (.png) from the URL
```
chrisbobbe, line 111: Let's do this with a [`RegExp`](https://api.flutter.dev/flutter/dart-core/RegExp-class.html) instead of just chopping characters off the end, which seems more confusing than necessary if `url` ever ends up ending in something other than ".png".

test/widgets/autocomplete_test.dart
```diff
@@ -108,37 +108,39 @@ void main() {
...
+        // Options are filtered correctly for query
+        // TODO(#226): Remove this extra edit when this bug is fixed.
+        await tester.enterText(composeInputFinder, 'hello @user ');
+        await tester.enterText(composeInputFinder, 'hello @user t');
+        await tester.pumpAndSettle(); // async computation; options appear
+
+        // "User Two" and "User Three" appear, but not "User One"
+        checkUserShown(user1, store, expected: false);
+        checkUserShown(user2, store, expected: true);
+        checkUserShown(user3, store, expected: true);
```
chrisbobbe, line 143: Keep two-space indent here; don't make it four-space.

chrisbobbe (changes requested):
Thanks! A few small comments from a light review, but more substantively:

The first commit—

```
autocomplete: Implement new design for @-mention autocomplete items
```

—doesn't pass `tools/check`. Please fix that for the next round, so that [all commits pass all tests](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#each-commit-must-be-coherent).

fombalang:
> Thanks! A few small comments from a light review, but more substantively:
> 
> The first commit—
> 
> ```
> autocomplete: Implement new design for @-mention autocomplete items
> ```
> 
> —doesn't pass `tools/check`. Please fix the branch so that [all commits pass all tests](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#each-commit-must-be-coherent).

Okay.
I am confused though, the check runs against the latest commit, right?
How do I then ensure this commit would pass the check independently?
Also, I squashed the fix that I did for the test into this commit.

fombalang:
I have made the changes, please review them when available.

chrisbobbe:
> How do I then ensure this commit would pass the check independently?

By running `tools/check` at that commit. 🙂 See the README on tests: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#tests

test/widgets/autocomplete_test.dart
```diff
@@ -108,24 +108,26 @@ void main() {
   group('ComposeAutocomplete', () {
 
     Finder findNetworkImage(String url) {
+      String trimmedUrl = url.replaceAll(RegExp(r'\.(png|jpg|jpeg|gif|bmp|webp)$',caseSensitive: false), ''); //Remove the extension from the URL
       return find.byWidgetPredicate((widget) => switch(widget) {
-        Image(image: NetworkImage(url: var imageUrl)) when imageUrl == url
-          => true,
+        Image(image: NetworkImage(url: var imageUrl)) when imageUrl.startsWith(trimmedUrl)
+        => true,
```
chrisbobbe, line 114: nit: the indentation on the `=> true,` line helps with readability; don't remove it:

```suggestion
        Image(image: NetworkImage(url: var imageUrl)) when imageUrl.startsWith(trimmedUrl)
          => true,
```

test/widgets/autocomplete_test.dart
```diff
@@ -108,24 +108,26 @@ void main() {
   group('ComposeAutocomplete', () {
 
     Finder findNetworkImage(String url) {
+      String trimmedUrl = url.replaceAll(RegExp(r'\.(png|jpg|jpeg|gif|bmp|webp)$',caseSensitive: false), ''); //Remove the extension from the URL
```
chrisbobbe, line 111: Ah, one thing I missed last time: we're changing `findNetworkImage` so it's specifically helpful for finding an _avatar_ image, not just any image, so let's change its name accordingly. How about `findAvatarImage`?

I don't think the implementation needs to recite those other extensions ("jpg", etc.). Our code for producing a "-medium" URL only handles `.png`s; see `UploadedAvatarUrl.get` in lib/model/avatar_url.dart.

I also think we can organize the code to be a bit clearer about what it's doing; maybe like this:

```dart
    Finder findAvatarImage(String url) {
      /// Whether an Image widget's URL is for the avatar we're looking for.
      ///
      /// Accepts the -medium.png variant; see [UploadedAvatarUrl.get].
      bool urlMatches(String candidateUrl) =>
        candidateUrl.startsWith(url.replaceAll(RegExp(r'\.png$'), ''));

      return find.byWidgetPredicate((widget) => switch(widget) {
        Image(image: NetworkImage(url: var imageUrl)) when urlMatches(imageUrl)
          => true,
        _ => false,
      });
    }
```

But also, I think there's another way to write `findAvatarImage` without even having to think about the `.png` / `-medium.png` detail. All we want from this `Finder` is to know whether an avatar is shown or not, right? How about this:

```dart
    Finder findAvatarImage(int userId) =>
      find.byWidgetPredicate((widget) => widget is AvatarImage && widget.userId == userId);
```

How about an NFC prep commit that replaces `findNetworkImage` with that?

chrisbobbe (commented):

test/widgets/autocomplete_test.dart
fombalang: Alright, I'll get started on the changes

fombalang:
I have made the latest changes. Please review when you can.

lib/widgets/autocomplete.dart
```diff
@@ -195,24 +198,56 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
+                  style: TextStyle(
+                    fontSize: 18,
+                    height: 20 / 18,
+                    color: designVariables.contextMenuItemLabel,
+                  )
+                    .merge(weightVariableTextStyle(context, wght: 600)),
+                  overflow: TextOverflow.ellipsis,
+                  maxLines: 1,
+                  label),
+                if(metadata != null) Text(
```
chrisbobbe, line 241: nit:

```suggestion
                if (metadata != null) Text(
```

as in my suggestion in https://github.com/zulip/zulip-flutter/pull/995#discussion_r1826382501.

lib/widgets/theme.dart
```diff
@@ -316,7 +331,10 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       borderBar: borderBar ?? this.borderBar,
       contextMenuCancelText: contextMenuCancelText ?? this.contextMenuCancelText,
       contextMenuItemBg: contextMenuItemBg ?? this.contextMenuItemBg,
+      contextMenuItemLabel: contextMenuItemLabel ?? this.contextMenuItemLabel,
+      contextMenuItemMeta: contextMenuItemMeta ?? this.contextMenuItemMeta,
       contextMenuItemText: contextMenuItemText ?? this.contextMenuItemBg,
```
chrisbobbe, line 336: Oh wow, something's wrong in the code that was already here before your changes:

```suggestion
      contextMenuItemText: contextMenuItemText ?? this.contextMenuItemText,
```

It would be helpful if you could add a small commit fixing that and a similar instance in `lerp` below.

chrisbobbe (commented):
Thanks, this is getting closer!

In https://github.com/zulip/zulip-flutter/pull/995#discussion_r1826404745, by "prep" commit I meant a commit that comes before the commit that needs it, to prepare for it. 🙂 In the branch as it is now, the helper function gets rewritten one way, but then that way is thrown out and replaced with the one we want. Instead of changing it twice, we should just change it once. That change should happen in a commit that comes before the commit that redesigns the UI, since that's the commit that needs the change.

lib/widgets/theme.dart
fombalang: Oh, didn't even notice this

fombalang:
> Thanks, this is getting closer!
> 
> In [#995 (comment)](https://github.com/zulip/zulip-flutter/pull/995#discussion_r1826404745), by "prep" commit I meant a commit that comes before the commit that needs it, to prepare for it. 🙂 In the branch as it is now, the helper function gets rewritten one way, but then that way is thrown out and replaced with the one we want. Instead of changing it twice, we should just change it once. That change should happen in a commit that comes before the commit that redesigns the UI, since that's the commit that needs the change.

Oh, I see now. 

Also, do I have to include "prep" in the commit message?

chrisbobbe:
No, we don't have a rule about that. :)

fombalang:
I have made the updates and fixed the conflicts and reordered the prep commit.
Also added a fix for the `contextMenuItemText` color variables as an NFC commit. I am a bit confused if a change like this would be considered an NFC commit? 🤔

chrisbobbe:
`tools/check` fails at this commit:

```
autocomplete_test [nfc]: Rename findNetworkImage to findAvatarImage
```

It looks like that commit is doing more than it says in the commit message. It's also changing some tests so that they expect an email address to appear in an autocomplete item. That's something we want to do, but not at this commit, before the app code has actually been changed to show the email address. We should add that expectation in the same commit that satisfies it:

```
autocomplete: Implement new design for @-mention autocomplete items
```

chrisbobbe:
> Also added a fix for the `contextMenuItemText` color variables as an NFC commit. I am a bit confused if a change like this would be considered an NFC commit? 🤔

Thank you!

Looks like it does make a functional change, although a small one. The `lerp` method is used to fade the colors smoothly over 200 milliseconds when transitioning between light mode and dark mode. So I imagine the bug you've fixed here could be reproduced on `main` this way:

1. Open the app [on Android](https://github.com/zulip/zulip-flutter/pull/867#issuecomment-2266293400)
2. In system settings, switch to light mode and schedule the app to switch to dark mode a minute or so after the current time (or vice versa)
3. Open the message list
4. Long-press a message to bring up the action sheet
5. While watching the text in the action sheet, wait for the scheduled switch from light to dark (or vice versa) that you set up in step 2
6. Notice that the text's color goes in the wrong direction over the 200ms animation, then suddenly snaps to the correct color when the animation is finished

I haven't tried this myself. This is quite a small bug, but one we're glad to fix! No need to spend time reproducing it; all you need to do is change your commit message to remove `[nfc]`. 🙂

fombalang:
> That's something we want to do, but not at this commit, before the app code has actually been changed to show the email address. We should add that expectation in the same commit that satisfies it:

Yeah. I think during my rebase I must have messed it up somehow, when I was reordering the commit structure.

fombalang:
I have updated the commits, removed the NFC label from the color fix, and adjusted the other commits.

I ran tools check on each commit using 
`git checkout <commit-hash> -- . && tools/check && git checkout -- .`
Not sure if this is the right command to use, but all the commits passed.

chrisbobbe:
The command I use, from a feature branch that tracks `main`, is:

```
git rebase -i --exec 'tools/check --diff @~'
```

Are you familiar with interactive rebase? This sets up an interactive rebase with an "exec" step that runs the command `tools/check --diff @~` at each commit. (The `--diff @~` part makes `tools/check` run only on files that differ from the previous commit.)

test/widgets/autocomplete_test.dart
```diff
@@ -108,25 +109,21 @@ void main() {
...
-    }
+    Finder findAvatarImage(int userId) =>
+      find.byWidgetPredicate((widget) => widget is AvatarImage && widget.userId == userId);
 
     void checkUserShown(User user, PerAccountStore store, {required bool expected}) {
       check(find.text(user.fullName).evaluate().length).equals(expected ? 1 : 0);
+      check(find.text(user.deliveryEmail!).evaluate().length).equals(expected ? 1 : 0);
       final avatarFinder =
-        findNetworkImage(store.tryResolveUrl(user.avatarUrl!).toString());
+        findAvatarImage(user.userId);
```
chrisbobbe, line 119: This can now fit on one line, like in your previous revision:

```dart
final avatarFinder = findAvatarImage(user.userId);
```

chrisbobbe (commented):
Thanks! LGTM except for one tiny nit below, and I'll mark this for Greg's review.

fombalang:
> The command I use, from a feature branch that tracks `main`, is:
> 
> ```
> git rebase -i --exec 'tools/check --diff @~'
> ```
> 
> Are you familiar with interactive rebase? This sets up an interactive rebase with an "exec" step that runs the command `tools/check --diff @~` at each commit. (The `--diff @~` part makes `tools/check` run only on files that differ from the previous commit.)

Okay, I will try this and see.
Yes, I am familiar with interactive rebase, although I am mainly used to working in the Android Studio UI and not the terminal.

test/widgets/autocomplete_test.dart
fombalang: Okay. Let me quickly do that

fombalang:
> Thanks! LGTM except for one tiny nit below, and I'll mark this for Greg's review.

Alright, great!
Thank you very much for your help reviewing!

lib/widgets/autocomplete.dart
```diff
@@ -195,24 +198,56 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
 
   @override
   Widget buildItem(BuildContext context, int index, MentionAutocompleteResult option) {
+    final designVariables = DesignVariables.of(context);
     Widget avatar;
     String label;
+    String? metadata;
+
     switch (option) {
```
gnprice, line 206: nit: use blank lines to separate stanzas logically:
```suggestion
    final designVariables = DesignVariables.of(context);

    Widget avatar;
    String label;
    String? metadata;
    switch (option) {
```

In particular `avatar`, `label`, and `metadata` should have their declarations remain together with the logic that initializes them.

lib/widgets/autocomplete.dart
```diff
@@ -195,24 +198,56 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
       case UserMentionAutocompleteResult(:var userId):
-        avatar = Avatar(userId: userId, size: 32, borderRadius: 3);
-        label = PerAccountStoreWidget.of(context).users[userId]!.fullName;
+        final store = PerAccountStoreWidget.of(context);
+        final user = store.users[userId]!;
+        avatar = Avatar(userId: userId, size: 36, borderRadius: 4);
+        label = user.fullName;
+        metadata = getDisplayEmailFor(user, store: store);
```
gnprice, line 212: "metadata" isn't really a good term for this (though I see it appears in the Figma design). It means "data about data"; but a user's email address is data about the user, just as much as their name is.

Ultimately what this variable is for — currently it's just email address but the design suggests including pronouns, and probably it'll include group descriptions and so on for other types of results — is just that it's a string that appears alongside the label, with less emphasis than the label.

So a good term would be "sublabel". Possible other directions would include "extra", "extraLabel", "secondLine".

lib/widgets/autocomplete.dart
```diff
@@ -195,24 +198,56 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
     return InkWell(
       onTap: () {
         _onTapOption(context, option);
       },
+      highlightColor: designVariables.editorButtonPressedBg,
+      splashFactory: NoSplash.splashFactory,
+      borderRadius: BorderRadius.circular(5),
       child: Padding(
-        padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
+        padding: const EdgeInsetsDirectional.only(start: 4, end: 8, top: 4, bottom: 4),
```
gnprice, line 222: nit:
```suggestion
        padding: const EdgeInsetsDirectional.fromSTEB(4, 4, 8, 4),
```

When all four sides (or even three sides) are present, `only` is no longer helpful and just listing all four like this is more concise.

lib/widgets/autocomplete.dart
```diff
@@ -195,24 +198,56 @@ class ComposeAutocomplete extends AutocompleteField<MentionAutocompleteQuery, Me
...
       case UserMentionAutocompleteResult(:var userId):
-        avatar = Avatar(userId: userId, size: 32, borderRadius: 3);
-        label = PerAccountStoreWidget.of(context).users[userId]!.fullName;
+        final store = PerAccountStoreWidget.of(context);
+        final user = store.users[userId]!;
+        avatar = Avatar(userId: userId, size: 36, borderRadius: 4);
+        label = user.fullName;
+        metadata = getDisplayEmailFor(user, store: store);
     }
     return InkWell(
```
gnprice, line 214: nit: similarly, separate these two stanzas:
```suggestion
    }

    return InkWell(
```

(That'd ideally be the case already in the existing code, but it's more so after both parts grow in this change. Can just add this blank line as part of the same main commit.)

gnprice (commented):
Thanks @fombalang for implementing this, and @chrisbobbe for the previous reviews!

Generally this all looks good. Various small comments below.

This should also get tests. No need to check things like that the font size or the colors etc. match the design — that sort of test wouldn't really add value beyond duplicating what's in the implementation. But there is a bit of logic here that would be good to test:
* When the user has a known delivery email, that email appears.
* When they don't, their fake email (`user.email`) doesn't appear. Plus we don't crash, and we still display the user's avatar and name.

To do this, let's add a couple of new test cases (new `testWidgets` calls).
* One takes a user with a non-null `deliveryEmail`, and confirms that it appears. Also the avatar and name appear.
* One takes a user with null `deliveryEmail`, and confirms that the avatar and name still appear, and the `user.email` doesn't.

Before this change which shows emails, that kind of testing was effectively covered as a side effect of the "appear, disappear, and change" test, because it checks the avatar and name are shown. But now that there's this more complex per-user logic on what to show, it should get its own explicit, separate, tests.

fombalang:
> Generally this all looks good. Various small comments below.

Alright, thank you for the corrections. I will implement these changes.

fombalang:
> This should also get tests. No need to check things like that the font size or the colors etc. match the design — that sort of test wouldn't really add value beyond duplicating what's in the implementation. But there is a bit of logic here that would be good to test:

Yeah, I originally intended to add some test cases similar to the ones you describe, which is what initially led me to find the small bug with the `emailAdressVisbility`. But, I couldn't figure out exactly how the account stores work to properly override the emailAdressVisbility and create the test cases. But I will get back to it and see how to add the test cases.

gnprice:
Cool, sounds good.



> But, I couldn't figure out exactly how the account stores work to properly override the emailAdressVisbility and create the test cases. But I will get back to it and see how to add the test cases.

Yeah, if you browse through the other tests in `test/widgets/`, you'll see examples of using that `eg.initialSnapshot` function you sent a fix for, and that should help you get going. If you have questions after giving that another try, please go ahead and ask in `#mobile-dev-help` on chat.zulip.org.

gnprice:
I've just merged #1069, which touched some of this same code. @fombalang when you push your next revision, don't worry about rebasing it forward past those changes — just push it without rebasing, and I can take care of rebasing it on top of those changes and resolving the conflicts.

(Ordinarily we expect contributors to handle resolving rebase conflicts; it's an important Git skill. But these conflicts are unusually complex, and have more to do with my #1069 than with your PR or with the routine need for resolving conflicts.)

If you use `git rebase -i` for revising your branch (which you probably will want to do), you can say `git rebase -i --keep-base` to tell it not to rebase forward, and to instead focus on what `rebase -i` specifically is good at.

Separately, even if you're still working on the tests, I'd encourage you to go ahead and push the revision you have that addresses all the other review comments. That way I can re-review those changes while you're still working on the tests.

fombalang:
> Separately, even if you're still working on the tests, I'd encourage you to go ahead and push the revision you have that addresses all the other review comments. That way I can re-review those changes while you're still working on the tests.

Okay, understood. Yes, I am still working on the test. but the other changes were done. I will push the updates then.
I was holding off on pushing because I thought the test should be included in the main commit according to the commit guides. Would it be okay if I add the tests as a separate commit later? I already started facing some challenges rebasing my changes as main started to diverge more and more.

fombalang:
I have sent the latest updates. Please review when you can.

fombalang:
I have added a commit for the delivery email visibility test case.
Do you think I should squash it into the main autocomplete commit, or is it fine as is?

gnprice:
Thanks for the revision!

I haven't yet reviewed the new changes, but first I've gone ahead and rebased them onto main, like I offered above; and I've pushed the result back to the PR branch.

So from here you should be able to rebase atop #1087 as [discussed in chat](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/ComposeAutocomplete.20test.20emailAddressVisbility.20help/near/1988702) today, or atop future other changes merged to main. I don't expect to have other changes landing in the near future that would have complex conflicts like the kind #1069 caused.

gnprice:
Also just added one more commit on top, to update the style for emoji autocomplete to match what this PR does for @-mention autocomplete. We'll leave that commit in the PR branch, and can merge it along with the rest of the PR when ready. (@chrisbobbe PTAL too since that's new material.)

test/widgets/autocomplete_test.dart
```diff
@@ -147,33 +153,54 @@ void main() {
...
+
+      // Options are filtered correctly for query
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(composeInputFinder, 'hello @user ');
+      await tester.enterText(composeInputFinder, 'hello @user t');
+      await tester.pumpAndSettle(); // async computation; options appear
+
+      // "User Two"'s delivery email is not visible and "User Three"'s delivery email is visible
+      checkUserShown(user2, expected: true, deliveryEmailExpected: false);
+      checkUserShown(user3, expected: true, deliveryEmailExpected: true);
```
gnprice, line 203: Let's have two separate test cases for this (i.e. two `testWidgets` calls), as I said above:
https://github.com/zulip/zulip-flutter/pull/995#pullrequestreview-2437163329

That's useful because it means if some change someone's working on causes these tests to fail in the future, they'll see specifically whether it was one case, the other, or both of them that failed.

It's also useful because it lets each of the test cases be simpler, which makes them easier to read, understand, and update when needed.

test/example_data.dart
```diff
@@ -125,7 +125,7 @@ User user({
   _checkPositive(userId, 'user ID');
   return User(
     userId: userId ?? _nextUserId(),
-    deliveryEmail: effectiveDeliveryEmail,
+    deliveryEmail: deliveryEmail,
```
gnprice, line 128: In general this sort of change should be in a separate commit before the main commit. That way it can get its own commit message clearly explaining why it's a good change. (Here if we were making this change we'd want some other changes in the surrounding code to match: it doesn't make sense to still have a variable called "effective delivery email" if it's not what actually gets used as the delivery email.)

In this case, though, the problem this is solving should be resolved by #1087, so you can rebase your work on top of that PR and drop this change.

gnprice (commented):
Thanks @fombalang for the revision!

I'll be out from tomorrow through Monday for the Thanksgiving holiday here in the US, so I won't get to a detailed review before I'm back. But I took a quick look at the new test; high-level comments below. So please go ahead and update the PR based on this round of feedback.

> Do you think I should squash it into the main autocomplete commit, or is it fine as is?

Yeah, the new test should go in the same commit as the code it tests.

See generally our Git style guide:

test/widgets/autocomplete_test.dart
fombalang: Okay, understood

test/example_data.dart
fombalang: Alright.

lib/widgets/autocomplete.dart
```diff
@@ -311,16 +311,26 @@ class _EmojiAutocompleteItem extends StatelessWidget {
...
       : [candidate.emojiName, ...candidate.aliases].join(", "); // TODO(#1080)
 
+    // TODO(design): emoji autocomplete results
+    // There's no design in Figma for emoji autocomplete results.
+    // Instead we adapt the design for the emoji picker to the
+    // context of autocomplete results as exemplified by _MentionAutocompleteItem.
+    // That means: emoji size, text size, text line-height from emoji picker;
+    // text color (for contrast with background) and outer padding
+    // from _MentionAutocompleteItem; padding around emoji glyph
+    // to bring it to same size as avatar in _MentionAutocompleteItem.
```
chrisbobbe, line 321: The text line-height in the Figma emoji picker is 24, not 18, so the code or the comment should change:

image

Also worth mentioning that the font weight comes from the emoji picker too (normal weight).

chrisbobbe (commented):
Thanks! Small comments from a review of Greg's new commit:

> Also just added one more commit on top, to update the style for emoji autocomplete to match what this PR does for @-mention autocomplete. We'll leave that commit in the PR branch, and can merge it along with the rest of the PR when ready. (@chrisbobbe PTAL too since that's new material.)

lib/widgets/autocomplete.dart
gnprice: > The text line-height in the Figma emoji picker

Hmm, good catch! It's 18 in this other box:
image

That one has the text wrapping onto two lines. It's the same 18px in the other example box where that happens, and 24px in those that fit on one line.

I don't think it's practical to have the line-height vary based on whether the text wraps. That's something the text shaping engine decides, and so isn't yet known at the build stage for a given frame, which is when we decide the line-height.

The line-height mostly ends up mattering when there's more than one line, I think — when there's just one line, it controls the height of the Text widget, but the emoji is taller so that doesn't much affect the layout. (It might slighly affect the vertical alignment as the text is centered.) So let's go with the 18px that's specified for the two-line case.

gnprice:
Thanks @chrisbobbe for the review! Pushed a revision adding a comment with a digest of that subthread.

@fombalang For the next step, please go ahead and revise the PR based on the feedback above at https://github.com/zulip/zulip-flutter/pull/995#pullrequestreview-2463104713 .

lib/widgets/autocomplete.dart
gnprice: Also started a chat thread to let Vlad know: https://chat.zulip.org/#narrow/channel/48-mobile/topic/variable.20line-height/near/1994856

chrisbobbe: Sure, sounds good.

chrisbobbe: Ah did you see this part:

> Also worth mentioning that the font weight comes from the emoji picker too (normal weight).

fombalang:
> Thanks @chrisbobbe for the review! Pushed a revision adding a comment with a digest of that subthread.
> 
> @fombalang For the next step, please go ahead and revise the PR based on the feedback above at [#995 (review)](https://github.com/zulip/zulip-flutter/pull/995#pullrequestreview-2463104713) .

Alright, will do so.
Apologies for the delay with the revisions, I will be sending them soon.

PIG208:
Hi @fombalang! Do you plan to go back to this and revise the PR?

fombalang:
> Hi @fombalang! Do you plan to go back to this and revise the PR?

Hello @PIG208 , yes I am working on the revisions now.

fombalang:
Hi all, and Happy New Year! 🎉

I've rebased my changes on top of the latest main, faced some difficulties cause they had diverged quite a bit but I was able to get it done.
I also separated the test cases for the delivery email into two as @gnprice suggested, one for when the email is available and one where it is not.

One thing to note, so a while ago we did the update about changing the implementation for the `findNetworkImage` to `findAvatarImage` to make it less brittle to changes like the avatar size, since we hadn't merged yet, the new tests for the emojis still use the `findNetworkImage` method.

I did not want to touch that, so I added the new `findAvatarImage` we did as a separate method which is used for the `autocomplete` (`@mentions`) group of tests.
Is that okay, or should we go back and merge both methods? I figured it'd be fine to separate it since the `findAvatarImage` is more specific for avatars only, and the `findNetworkImage` can still be used in other areas (the emojis for example.)

gnprice:
Thanks @fombalang for the revision! Happy New Year.

> since we hadn't merged yet, the new tests for the emojis still use the `findNetworkImage` method.

That's fine — the emoji image URLs don't have the same tricky dependency on external facts (the device pixel ratio) that the avatar URLs do. Each image emoji has just a single size of image and a single URL to find it at. (Well, animated emoji have a second URL for the non-animated version; but the setting that controls whether to use that is a user setting entirely internal to the app, so is well controlled in tests.) So `findNetworkImage` is fine for those.

Looking at the rest of the changes now.

test/widgets/autocomplete_test.dart
```diff
@@ -147,33 +153,70 @@ void main() {
...
+
+      debugNetworkImageHttpClientProvider = null;
+    });
+
+    testWidgets('delivery email not visible when unavailable', (tester) async {
+      final user1 = eg.user(userId: 1, fullName: 'User One', avatarUrl: 'user1.png',);
+      final composeInputFinder = await setupToComposeInput(tester, users: [user1]);
+
+      TypingNotifier.debugEnable = false;
+      addTearDown(TypingNotifier.debugReset);
```
gnprice, line 192: Huh, why do these two new tests need this and the existing test doesn't?

I suspect these don't need it either — it looks like `setupToComposeInput` handles it.

test/widgets/autocomplete_test.dart
```diff
@@ -147,33 +153,70 @@ void main() {
...
+      await tester.enterText(composeInputFinder, 'hello @user o');
+      await tester.pumpAndSettle(); // async computation; options appear
+
+      // Check "User One"'s delivery email is not visible
+      checkUserShown(user1, expected: true, deliveryEmailExpected: false);
+
+      debugNetworkImageHttpClientProvider = null;
+    });
+    testWidgets('delivery email visible when available', (tester) async {
+      final user1 = eg.user(userId: 1, fullName: 'User One', avatarUrl: 'user1.png', deliveryEmail: 'email1@email.com');
```
gnprice, line 206: line too long — important information is past 80 columns

Concretely, here's what that line looks like for me right now:
image

As you can see, the delivery email is invisible.

gnprice: (I have a big monitor, but I use that horizontal space to show multiple things side by side, especially when reviewing. Often that's:
* the code, in my IDE
* another IDE pane with a related file, e.g. to read tests side by side with the code they're testing
* a terminal with `git log --stat -p` output, showing the content of the PR
* a terminal with `git range-diff` output, to compare the current revision of a PR with a previous one
* the browser window with the GitHub page I'm typing the review into

)

test/widgets/autocomplete_test.dart
```diff
@@ -147,33 +153,70 @@ void main() {
...
+      addTearDown(TypingNotifier.debugReset);
+
+      // Options are filtered correctly for query
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(composeInputFinder, 'hello @user ');
+      await tester.enterText(composeInputFinder, 'hello @user o');
+      await tester.pumpAndSettle(); // async computation; options appear
+
+      // Check "User One"'s delivery email is not visible
+      checkUserShown(user1, expected: true, deliveryEmailExpected: false);
```
gnprice, line 201: This comment doesn't make a lot of sense, since User One has no delivery email in the first place :slightly_smiling_face: 

OTOH as I wrote at https://github.com/zulip/zulip-flutter/pull/995#pullrequestreview-2437163329 , this test case should check that `user.email` doesn't appear. It's easy to imagine having a bug where when there's no `user.deliveryEmail`, we show `user.email` instead — which would be wrong because `user.email` in that case is a meaningless fake email made up by the Zulip server.

gnprice: It may be cleanest to do that check by inlining the contents of `checkUserShown` into this test case (and the contrasting test case below), and adapting the details, rather than giving `checkUserShown` more parameters and conditions.

lib/widgets/autocomplete.dart
gnprice: Ah thanks. Fixed that in the version I pushed just now, as well as deleting the comment obviated by Vlad's reply on that chat thread:
```diff
@@ -318,25 +318,20 @@ class _EmojiAutocompleteItem extends StatelessWidget {
     // There's no design in Figma for emoji autocomplete results.
     // Instead we adapt the design for the emoji picker to the
     // context of autocomplete results as exemplified by _MentionAutocompleteItem.
-    // That means: emoji size, text size, text line-height from emoji picker;
-    // text color (for contrast with background) and outer padding
-    // from _MentionAutocompleteItem; padding around emoji glyph
+    // That means: emoji size, text size, text line-height, and font weight
+    // from emoji picker; text color (for contrast with background) and
+    // outer padding from _MentionAutocompleteItem; padding around emoji glyph
     // to bring it to same size as avatar in _MentionAutocompleteItem.
     return Padding(
       padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
       child: Row(children: [
         if (glyph != null) ...[
           Padding(padding: const EdgeInsets.all(6),
             child: glyph),
           const SizedBox(width: 6),
         ],
         Expanded(
           child: Text(
-            // The Figma design for the emoji picker actually calls for the
-            // line-height to be 24px when the label fits on one line,
-            // and 18px when it wraps.  But whether it'll wrap is something we
-            // don't know at build time, so make it always 18px.  Discussion:
-            //   https://github.com/zulip/zulip-flutter/pull/995#discussion_r1868352275
             style: TextStyle(fontSize: 17, height: 18 / 17,
               color: designVariables.contextMenuItemLabel),
```

PIG208:
Hi @fombalang. Thanks for working on this! Do you think you will have time to continue working on this addressing the review comments?

gnprice:
(The issue this addresses is an M6 Post-launch issue, but I'm still eager to get this PR finished and merged, as an exception to our [stricter focus on launch issues](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556) — a lot of effort has gone into this already, and it's close to merge and will be great to have.)

PIG208:
In a call, we found that this PR is actually for an M5-a issue.

I think the mislabeling is probably due to #913 (M5-a) and #914 (M6) being similar issues for different parts of the design. #913 is more feasible for completion near term because it only focuses on the redesign for autocomplete items.

I will pick up this PR (reusing this branch) from here and address the comments. Thanks for all the work @fombalang, and the prior reviews from Chris and Greg!

PIG208:
The PR has been updated. It preserves most of the design changes, and on top of that:
- reuse the sub label style, applying it to wildcard autocomplete result items (that previously followed web);
- update to match the design's lighter font weight for items where the sub label is not visible.

Updated screenshot:

mixed options

gnprice:
Thanks @PIG208 for picking this up! Taking a look now.

> I think the mislabeling is probably due to #913 (M5-a) and #914 (M6) being similar issues for different parts of the design.

Aha — yeah, that is a likely diagnosis of the mixup.

lib/widgets/autocomplete.dart
```diff
@@ -256,32 +262,60 @@ class _MentionAutocompleteItem extends StatelessWidget {
...
     }
 
+    final labelWidget = Text(
+      label,
+      style: TextStyle(
+        fontSize: 18,
+        height: 20 / 18,
+        color: designVariables.contextMenuItemLabel,
+      ).merge(weightVariableTextStyle(context,
+          wght: sublabel == null ? 500 : 600)),
```
gnprice, line 294: Huh interesting, subtle.

lib/widgets/autocomplete.dart
```diff
@@ -256,32 +262,60 @@ class _MentionAutocompleteItem extends StatelessWidget {
...
+      maxLines: 1);
+
     return Padding(
-      padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
+      padding: const EdgeInsetsDirectional.fromSTEB(4, 4, 8, 4),
       child: Row(children: [
-        avatar,
-        const SizedBox(width: 8),
-        label,
+        SizedBox.square(dimension: 36, child: avatar),
```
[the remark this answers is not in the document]
gnprice: Anyway, in the user/avatar case, this SizedBox is redundant. Let's simplify a bit by moving it to within the wildcard case.

So effectively it becomes part of the `avatar` local's job to control its size, to avoid redundancy since when `Avatar` is used it will be controlling the size anyway.

gnprice (commented):
Thanks @PIG208 for picking this up! Just small comments below.

lib/model/store.dart
```diff
@@ -498,6 +498,33 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
     return byDate.difference(dateJoined).inDays >= realmWaitingPeriodThreshold;
   }
 
+  /// The given user's real email address, if known, for displaying in the UI.
+  ///
+  /// Returns null if self-user isn't able to see [user]'s real email address.
+  String? userDisplayEmail(User user) {
```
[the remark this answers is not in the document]
PIG208: Perhaps like `Unreads` that rely on a reference to a `ChannelStoreImpl`, we can split out realm settings for `UserStore`. Probably a good refactor project as a follow-up.

lib/widgets/autocomplete.dart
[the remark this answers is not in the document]
PIG208: Yeah, from `Icon.build`:

```dart
    return Semantics(
      label: semanticLabel,
      child: ExcludeSemantics(
        child: SizedBox(width: iconSize, height: iconSize, child: Center(child: iconWidget)),
      ),
    );
```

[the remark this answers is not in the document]
PIG208: The `SizedBox` is ignored because the `SizedBox` in our code, and `iconWidget` is a `RichText` whose `fontSize` is `iconSize`.

lib/model/store.dart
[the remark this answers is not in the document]
gnprice: Yeah, or I'd organize it slightly differently to match the existing way UserStore works: see https://github.com/zulip/zulip-flutter/pull/1327#issuecomment-2635897862 .

(Small correction: `Unreads` takes a `ChannelStore`, not a `ChannelStoreImpl`. Generally the only references to the "impl" classes are in PerAccountStore.)

gnprice:
Thanks! Looks good; merging.

PR #998, opened 2024-10-15
api [nfc]: Update notif-token routes based on API docs

gnprice:
When we originally wrote the bindings for these routes, they weren't
yet documented, so we invented our own names for them.  Now they are,
so the API docs give canonical names.  Switch to those,
and mark the `appid` parameter required as the docs say.

PR #999, opened 2024-10-16
check: Add step to install `libsqlite3-dev` system dependency

rajveermalviya:
The `package:sqlite3` requires the system-installed sqlite3 shared library (`libsqlite3.so`) when running directly via Dart, on Linux. Whereas, when running under Flutter, it uses bundled libraries provided by `package:sqlite3_flutter_libs`.

Recently, the `ubuntu-latest` runner image has switched to defaulting to `ubuntu-24.04`:
  https://github.com/actions/runner-images/issues/10636
Previously, it used `ubuntu-22.04`, which had the `libsqlite3-dev` package pre-installed:
  https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#installed-apt-packages
However, the `ubuntu-24.04` image no longer includes this package by default:
  https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#installed-apt-packages

As a result, five unit tests now fail with the following error:
```
```
To resolve these test failures, add a step to manually install the `libsqlite3-dev` package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,12 @@ on: [push, pull_request]
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4.1.7
+
+    - name: Install system dependencies
+      run: TERM=dumb sudo apt install libsqlite3-dev -y
 
     - name: Set up JDK
       uses: actions/setup-java@v4.2.1


rajveermalviya:
Well, just as I was putting this PR up, it looks like the `ubuntu-latest == ubuntu-24.04` rollout is being rolled back:
https://github.com/actions/runner-images/issues/10636#issuecomment-2417303444

fun 😄

fombalang:
> Well, just as I was putting this PR up, it looks like the `ubuntu-latest == ubuntu-24.04` rollout is being rolled back: [actions/runner-images#10636 (comment)](https://github.com/actions/runner-images/issues/10636#issuecomment-2417303444)
> 
> fun 😄

Oh, I was wondering why some test suddenly failed on the pipeline when I sent in my PR.

gnprice:
Thanks @rajveermalviya for sorting this out! I'd run into the same issue yesterday at https://github.com/zulip/zulip-flutter/pull/998#issuecomment-2415545081 but hadn't yet spent time to investigate.

This seems like a good change even if for the moment it's no longer needed. How about combining it with a change to use `ubuntu-24.04` specifically? That way CI would validate for us that the change fully solves the problem.

rajveermalviya:
Thanks for the review @gnprice! I've updated the commit to also change the checks' runner image to use `ubuntu-24.04` explicitly. PTAL.

gnprice:
Thanks! Looks good; merging.

I tweaked the commit message, mainly here:

    -    As a result, five unit tests now fail with the following error:
    +    Without that package, five unit tests would fail with the following
    +    error:

because (since GitHub rolled back the change of `ubuntu-latest` to 24.04) there's no longer a "now" when the unit tests actually do fail; rather it's a hypothetical, that if we just bumped the version without adding this setup step, those tests would fail.

PR #1005, opened 2024-10-18
store [nfc]: Offer singleton global store on binding

gnprice:
This will be useful in the context of showing a notification, where there may not exist an element tree at all (because we may be in a background isolate), and even if there happens to be one (because we happen to be in the foreground) it doesn't otherwise relate to anything the code is doing.

Prompted by #856 (/cc @rajveermalviya).


chrisbobbe:
Great! LGTM, merged.

PR #1007, opened 2024-10-18
store: Implement removeAccount

chrisbobbe:
This PR was split out of #948, following https://github.com/zulip/zulip-flutter/pull/948#pullrequestreview-2376668035 , toward getting things merged more efficiently.

Related: #948

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -144,7 +144,14 @@ abstract class GlobalStore extends ChangeNotifier {
   /// This method should be called only by the implementation of [perAccount].
   /// Other callers interested in per-account data should use [perAccount]
   /// and/or [perAccountSync].
-  Future<PerAccountStore> loadPerAccount(int accountId);
+  Future<PerAccountStore> loadPerAccount(int accountId) {
+    return doLoadPerAccount(accountId);
+  }
+
+  /// Load per-account data for the given account, unconditionally.
+  ///
+  /// This method should be called only by [loadPerAccount].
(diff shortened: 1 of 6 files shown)


lib/model/store.dart
```diff
@@ -850,8 +916,11 @@ class UpdateMachine {
     BackoffMachine? backoffMachine;
 
     while (true) {
+      if (_disposed) return;
+
       if (_debugLoopSignal != null) {
```
gnprice, line 921: (followup on https://github.com/zulip/zulip-flutter/pull/948#discussion_r1805674256 )

```suggestion
    assert(!_disposed);

    BackoffMachine? backoffMachine;
    while (true) {
      if (_debugLoopSignal != null) {
```

I think a check at the top of the loop shouldn't be needed — it's not in itself right after an await, and we already have checks after each await.

Conversely an assert at the top of the whole method is good for the same reason as on other methods: it helps us catch if we're attempting to call the method when the object is already disposed, which would be a bug.

gnprice (commented):
Thanks! One comment below; otherwise all looks good.

lib/model/store.dart
chrisbobbe: Oh indeed! Reading more closely, I see you did suggest putting it at the top of the method, not the top of the loop.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

chrisbobbe:
Great! Just sent #1010 for the rest of #948.

PR #1010, opened 2024-10-19
login: Support logging out of an account (take 2)

chrisbobbe:
These are the three last commits from #948, following up on the just-merged #1007 to replace #948: https://github.com/zulip/zulip-flutter/pull/948#pullrequestreview-2376668035

The screenshots in #948's description still match this code. As I said on #948, there might be some improvements to make to the UI, perhaps as a followup: https://github.com/zulip/zulip-flutter/pull/948#issuecomment-2423253936

Fixes: #463

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -275,24 +275,32 @@ class ChooseAccountPage extends StatelessWidget {
   }
 }
 
-enum ChooseAccountPageOverflowMenuItem { aboutZulip }
-
 class ChooseAccountPageOverflowButton extends StatelessWidget {
   const ChooseAccountPageOverflowButton({super.key});
 
   @override
   Widget build(BuildContext context) {
-    return PopupMenuButton<ChooseAccountPageOverflowMenuItem>(
-      itemBuilder: (BuildContext context) => const [
-        PopupMenuItem(
-          value: ChooseAccountPageOverflowMenuItem.aboutZulip,
-          child: Text('About Zulip')),
-      ],
-      onSelected: (item) {
-        switch (item) {
-          case ChooseAccountPageOverflowMenuItem.aboutZulip:
+    final designVariables = DesignVariables.of(context);
+    final materialLocalizations = MaterialLocalizations.of(context);
+    return MenuAnchor(
+      menuChildren: [
+        MenuItemButton(
+          onPressed: () {
             Navigator.push(context, AboutZulipPage.buildRoute(context));
-        }
+          },
+          child: const Text('About Zulip')), // TODO(i18n)
+      ],
+      builder: (BuildContext context, MenuController controller, Widget? child) {
+        return IconButton(
+          tooltip: materialLocalizations.showMenuTooltip, // "Show menu"
+          onPressed: () {
+            if (controller.isOpen) {
+              controller.close();
+            } else {
+              controller.open();
+            }
+          },
+          icon: Icon(Icons.adaptive.more, color: designVariables.icon));
       });
   }
 }
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -8,8 +8,10 @@ import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 import '../log.dart';
 import '../model/localizations.dart';
 import '../model/narrow.dart';
+import '../model/store.dart';
 import '../notifications/display.dart';
 import 'about_zulip.dart';
+import 'actions.dart';
 import 'app_bar.dart';
 import 'dialog.dart';
 import 'inbox.dart';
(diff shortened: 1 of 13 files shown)


lib/widgets/app.dart
```diff
@@ -221,15 +223,64 @@ class ChooseAccountPage extends StatelessWidget {
...
     Widget? subtitle,
   }) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
     return Card(
       clipBehavior: Clip.hardEdge,
       child: ListTile(
         title: title,
         subtitle: subtitle,
+        trailing: PopupMenuButton<AccountItemOverflowMenuItem>(
```
gnprice, line 233: It looks like the PopupMenuButton doc recommends using the newer MenuAnchor instead. What if we used that — how would it differ?

chrisbobbe: | `PopupMenuButton` | `MenuAnchor` |
| --- | --- |
| image | image |

chrisbobbe: It's not much different, but I'm happy to use the more modern thing. 🤷‍♂️

gnprice: Yeah, SGTM

chrisbobbe:
```
app: Update three-dots button to use new M3 MenuAnchor widget
```

gnprice (commented):
Thanks! Generally this looks good. Comments below from a full review.

@alya see question below about wording for a confirmation modal.

Then one other thing this should have before we merge it is some tests, specifically for the last commit. This is kind of tricky code, and it's the sort of feature that we infrequently exercise but that is important to people when they do use it — so that makes tests extra useful.

lib/widgets/app.dart
```diff
@@ -221,15 +223,69 @@ class ChooseAccountPage extends StatelessWidget {
...
+    final account = globalStore.getAccount(accountId);
+    if (account == null) return; // TODO(log)
+
+    // Unawaited, to not block removing the account on this request.
+    unawaited(_unregisterToken(globalStore, account));
+
+    await globalStore.removeAccount(accountId);
+  }
+
+  Future<void> _unregisterToken(GlobalStore globalStore, Account account) async {
```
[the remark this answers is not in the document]
chrisbobbe: > So I think `actions.dart` is a fine home for them at least until then.

Cool, yeah, that does look like a good place for it.

lib/widgets/app.dart
```diff
@@ -221,15 +223,69 @@ class ChooseAccountPage extends StatelessWidget {
...
     return Card(
       clipBehavior: Clip.hardEdge,
       child: ListTile(
         title: title,
         subtitle: subtitle,
+        trailing: MenuAnchor(
+          menuChildren: [
+            MenuItemButton(
+              onPressed: () => unawaited(_logOutAccount(context, accountId)),
+              child: Text(zulipLocalizations.chooseAccountPageLogOutButton)),
```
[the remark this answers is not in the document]
chrisbobbe: > red "destructive action" styling on the "Log out" button.

Sure. I've just filed #1032 for this; how about I leave this part as a `TODO(#1032)` in my next revision?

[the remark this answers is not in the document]
gnprice: Sure, SGTM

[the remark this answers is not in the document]
alya: Hm, maybe we should try to explain the impact on the user a bit more? Something like this, assuming it's accurate?

> Log out?
> To use this account in the future, you will have to re-enter the URL for your organization and your account information.
> (buttons:) Log out / Cancel

chrisbobbe:
Thanks for the review! Revision pushed, this time with a bunch of tests. 🙂

test/widgets/actions_test.dart
```diff
@@ -30,19 +41,247 @@ void main() {
...
+      await prepare(tester);
+      check(testBinding.globalStore).accountIds.single.equals(eg.selfAccount.id);
+      const unregisterDelay = Duration(seconds: 5);
+      assert(unregisterDelay > TestGlobalStore.removeAccountDuration);
+      final newConnection = separateConnection()
+        ..prepare(delay: unregisterDelay, json: {'msg': '', 'result': 'success'});
+
+      final future = logOutAccount(context, eg.selfAccount.id);
+      // Unregister-token request and account removal dispatched together
+      assertSingleUnregisterRequest(newConnection);
```
gnprice, line 111: nit: this function wraps a `check` call, so use the verb "check" to name it:
```suggestion
      checkSingleUnregisterRequest(newConnection);
```

test/widgets/actions_test.dart
```diff
@@ -30,19 +41,247 @@ void main() {
...
+      check(testBinding.globalStore).accountIds.isEmpty();
+      check(connection.isOpen).isFalse();
+      check(newConnection.isOpen).isTrue(); // still busy with unregister-token
+
+      await tester.pump(unregisterDelay - TestGlobalStore.removeAccountDuration);
+      check(newConnection.isOpen).isFalse();
+
+      // No unexpected requests made on either connection
+      check(connection.takeRequests()).isEmpty();
+      check(newConnection.takeRequests()).isEmpty();
```
gnprice, line 127: I think these are redundant with the check in `send` — requests that are unexpected won't have a prepared response, so they'll throw, and the test framework will catch the unhandled exception and report.

test/widgets/actions_test.dart
```diff
@@ -30,19 +41,247 @@ void main() {
...
+      await prepare(tester, ackedPushToken: '123');
+
+      final newConnection = separateConnection()
+        ..prepare(json: {'msg': '', 'result': 'success'});
+      final future = unregisterToken(testBinding.globalStore, eg.selfAccount.id);
+      await tester.pump(Duration.zero);
+      await future;
+      assertSingleUnregisterRequest(newConnection, expectedToken: '123');
+      check(newConnection.isOpen).isFalse();
+    });
```
gnprice, line 240:
```suggestion
    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
```

That way we exercise both the Android and iOS cases in unregisterToken.

gnprice: (I think adding this to just this test is enough.)

test/widgets/actions_test.dart
```diff
@@ -30,19 +41,247 @@ void main() {
 
   Future<void> prepare(WidgetTester tester, {
     UnreadMessagesSnapshot? unreadMsgs,
+    String? ackedPushToken = '123',
```
gnprice, line 44: This change can be squashed into the commit after it, which starts using this feature.

(It doesn't touch the same code as the later commit; and it doesn't really need its own commit message, in that it's self-explanatory given the context of that other commit.)

gnprice (commented):
Thanks for the revision! Generally this all looks good; just small comments below.

Would you also post a screenshot of the new confirmation dialog?

chrisbobbe:
Thanks! Revision pushed, and here's that screenshot of the new confirmation dialog:

image

gnprice:
Thanks! Looks good; merging.

PR #1012, opened 2024-10-19
handle unknown users everywhere

mofirojean:
fixes: #716

### DIFF
--- a/lib/widgets/inbox.dart
+++ b/lib/widgets/inbox.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 
 import '../api/model/model.dart';
 import '../model/narrow.dart';
@@ -368,18 +369,19 @@ class _DmItem extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final store = PerAccountStoreWidget.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
     final selfUser = store.users[store.selfUserId]!;
 
     final designVariables = DesignVariables.of(context);
 
     final title = switch (narrow.otherRecipientIds) { // TODO dedupe with [RecentDmConversationsItem]
       [] => selfUser.fullName,
-      [var otherUserId] => store.users[otherUserId]?.fullName ?? '(unknown user)',
+      [var otherUserId] => store.users[otherUserId]?.fullName ?? zulipLocalizations.unknownUserName,
 
       // TODO(i18n): List formatting, like you can do in JavaScript:
       //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Shu'])
       //   // 'Chris、Greg、Alya、Shu'
-      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', '),
+      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? zulipLocalizations.unknownUserName).join(', '),
     };
 
     return Material(
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -322,6 +322,8 @@ class MessageListAppBarTitle extends StatelessWidget {
   Widget build(BuildContext context) {
     final zulipLocalizations = ZulipLocalizations.of(context);
 
+    // TODO(i18n): provide tranlations just as 'zulipLocalizations.unknownUserName'
+    // for 'unknown channel'
     switch (narrow) {
       case CombinedFeedNarrow():
         return Text(zulipLocalizations.combinedFeedPageTitle);
@@ -349,7 +351,7 @@ class MessageListAppBarTitle extends StatelessWidget {
         if (otherRecipientIds.isEmpty) {
           return const Text("DMs with yourself");
         } else {
-          final names = otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)');
+          final names = otherRecipientIds.map((id) => store.users[id]?.fullName ?? zulipLocalizations.unknownUserName);
           return Text("DMs with ${names.join(", ")}"); // TODO show avatars
         }
     }
--- a/lib/widgets/recent_dm_conversations.dart
+++ b/lib/widgets/recent_dm_conversations.dart
@@ -93,6 +93,7 @@ class RecentDmConversationsItem extends StatelessWidget {
   Widget build(BuildContext context) {
     final store = PerAccountStoreWidget.of(context);
     final selfUser = store.users[store.selfUserId]!;
+    final zulipLocalizations = ZulipLocalizations.of(context);
 
     final designVariables = DesignVariables.of(context);
 
@@ -107,13 +108,13 @@ class RecentDmConversationsItem extends StatelessWidget {
         //   (should we offer a "spam folder" style summary screen of recent
         //   1:1 DM conversations from muted users?)
         final otherUser = store.users[otherUserId];
-        title = otherUser?.fullName ?? '(unknown user)';
+        title = otherUser?.fullName ?? zulipLocalizations.unknownUserName;
         avatar = AvatarImage(userId: otherUserId, size: _avatarSize);
       default:
         // TODO(i18n): List formatting, like you can do in JavaScript:
         //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya'])
         //   // 'Chris、Greg、Alya'
-        title = narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', ');
+        title = narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? zulipLocalizations.unknownUserName).join(', ');
         avatar = ColoredBox(color: designVariables.groupDmConversationIconBg,
           child: Center(
             child: Icon(color: designVariables.groupDmConversationIcon,
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
 
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
@@ -143,6 +144,7 @@ class ReactionChip extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final store = PerAccountStoreWidget.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
 
     final reactionType = reactionWithVotes.reactionType;
     final emojiCode = reactionWithVotes.emojiCode;
@@ -157,7 +159,7 @@ class ReactionChip extends StatelessWidget {
       ? userIds.map((id) {
           return id == store.selfUserId
             ? 'You'
-            : store.users[id]?.fullName ?? '(unknown user)'; // TODO(i18n)
+            : store.users[id]?.fullName ?? zulipLocalizations.unknownUserName;
         }).join(', ')
       : userIds.length.toString();
 


chrisbobbe:
@PIG208 could you take this one? :)

PIG208:
Hi! Please take a look at the [commit-discipline](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html) to fix the commit messages. Per [this suggestion](https://github.com/zulip/zulip-flutter/issues/716#issuecomment-2420223777), I think our goal for this PR is to fix *part of* #716, before going over all the places that need to be updated.

lib/widgets/inbox.dart
```diff
@@ -368,18 +369,19 @@ class _DmItem extends StatelessWidget {
...
     final title = switch (narrow.otherRecipientIds) { // TODO dedupe with [RecentDmConversationsItem]
       [] => selfUser.fullName,
-      [var otherUserId] => store.users[otherUserId]?.fullName ?? '(unknown user)',
+      [var otherUserId] => store.users[otherUserId]?.fullName ?? zulipLocalizations.unknownUserName,
 
       // TODO(i18n): List formatting, like you can do in JavaScript:
       //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Shu'])
       //   // 'Chris、Greg、Alya、Shu'
-      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', '),
+      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? zulipLocalizations.unknownUserName).join(', '),
```
PIG208, line 384: This line was already quite long:
```suggestion
      _ => narrow.otherRecipientIds.map((id) =>
             store.users[id]?.fullName ?? zulipLocalizations.unknownUserName,
           ).join(', '),
```

PIG208 (commented):
I think the second commit can be squashed into the first one.

One other thing to note is that you can use the [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#nfc) tag for your commits if they don't affect the behavior.

This is a change that we would like to have but it is not addressing #716 by finding places where we don't have unknown users. Nevertheless, this should be a good checkpoint before we start working on a PR.

lib/widgets/inbox.dart
mofirojean: I have updated the code with the suggestions you gave, and also joined the two commits to one.

Moving next to implement the issue at hand.

lib/widgets/inbox.dart
```diff
@@ -368,18 +369,21 @@ class _DmItem extends StatelessWidget {
...
-      [var otherUserId] => store.users[otherUserId]?.fullName ?? '(unknown user)',
+      [var otherUserId] => store.users[otherUserId]?.fullName ?? zulipLocalizations.unknownUserName,
 
       // TODO(i18n): List formatting, like you can do in JavaScript:
       //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Shu'])
       //   // 'Chris、Greg、Alya、Shu'
-      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', '),
+      _ => narrow.otherRecipientIds.map((id) =>
+        store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
+      ).join(', '),
```
PIG208, line 386: Looks like the updated code is not properly indented. Could you check again?

lib/widgets/message_list.dart
```diff
@@ -322,6 +322,8 @@ class MessageListAppBarTitle extends StatelessWidget {
   Widget build(BuildContext context) {
     final zulipLocalizations = ZulipLocalizations.of(context);
 
+    // TODO(i18n): provide tranlations just as 'zulipLocalizations.unknownUserName'
+    // for 'unknown channel'
```
PIG208, line 326: nit: `tranlations` -> `translations`

PIG208: It would be better to move this comment to right before any of the lines containing `(unknown channel)`, so it is easier to spot where the translation is applicable.

lib/widgets/message_list.dart
```diff
@@ -349,7 +351,7 @@ class MessageListAppBarTitle extends StatelessWidget {
         if (otherRecipientIds.isEmpty) {
           return const Text("DMs with yourself");
         } else {
-          final names = otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)');
+          final names = otherRecipientIds.map((id) => store.users[id]?.fullName ?? zulipLocalizations.unknownUserName);
```
PIG208, line 354: nit:
```suggestion
          final names = otherRecipientIds.map((id) =>
            store.users[id]?.fullName ?? zulipLocalizations.unknownUserName);
```

lib/widgets/recent_dm_conversations.dart
```diff
@@ -107,13 +108,13 @@ class RecentDmConversationsItem extends StatelessWidget {
...
         final otherUser = store.users[otherUserId];
-        title = otherUser?.fullName ?? '(unknown user)';
+        title = otherUser?.fullName ?? zulipLocalizations.unknownUserName;
         avatar = AvatarImage(userId: otherUserId, size: _avatarSize);
       default:
         // TODO(i18n): List formatting, like you can do in JavaScript:
         //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya'])
         //   // 'Chris、Greg、Alya'
-        title = narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', ');
+        title = narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? zulipLocalizations.unknownUserName).join(', ');
```
PIG208, line 117: nit:
```suggestion
        title = narrow.otherRecipientIds.map((id) =>
            store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
          ).join(', ');
```

PIG208 (commented):
The `[nfc]` tag is usually used as a suffix. Here are some examples:

* 402f998e color [nfc]: Fix a straggling reference to Color.value, in a pigeon's doc
* 2737d4ce store [nfc]: Comment in _PerAccountStoreWidgetState.didChangeDependencies
* 26069bd7 store [nfc]: Add indirection doLoadPerAccount
* fc80be1a store [nfc]: Offer singleton global store on binding

So we should update `[nfc]: handle unknown users everywhere` to maybe `i18n [nfc]: Handle unknown users everywhere` (also note the capitalization change).

lib/widgets/inbox.dart
mofirojean: Okay, i will do that right away

lib/widgets/message_list.dart
mofirojean: Okay :+1: , just noticed i didn't spell that right

mofirojean:
Hi @PIG208 , I updated the changes you requested. I have some questions with regards to the issue. As I understand it, the app has a map called `PerAccountStore.users` (also referred to as `store.users`), which holds information about all known users, and sometimes this list may not include all the users in the realm. Also, in some cases, when you try to look up a user by their userId in `store.users`, the result might be null (i.e., that user doesn’t exist in the map) and this may result in a crash.

Amongst the solutions, it was said not to use `!` or `assertions` and use the dart's null-safety operator and then provide fallbacks for missing users, along with other solutions. I checked the codebase with the reference `store.users` and identified a few areas that match the description of the issue.

for the first case i worked on adding the placeholders and changing some translations as well.

I then saw this area with an assertion, `models/compose.dart`

```dart
String quoteAndReplyPlaceholder(PerAccountStore store, {
  required Message message,
}) {
  final sender = store.users[message.senderId];
  // Is this the area i should resolve
  assert(sender != null);

  final url = narrowLink(store,
    SendableNarrow.ofMessage(message, selfUserId: store.selfUserId),
    nearMessageId: message.id);
  ....
}
```
Is this an area, that requires the changes that address our issue?

lib/widgets/inbox.dart
```diff
@@ -368,18 +369,21 @@ class _DmItem extends StatelessWidget {
...
-      [var otherUserId] => store.users[otherUserId]?.fullName ?? '(unknown user)',
+      [var otherUserId] => store.users[otherUserId]?.fullName ?? zulipLocalizations.unknownUserName,
 
       // TODO(i18n): List formatting, like you can do in JavaScript:
       //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Shu'])
       //   // 'Chris、Greg、Alya、Shu'
-      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', '),
+      _ => narrow.otherRecipientIds.map((id) =>
+              store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
+           ).join(', '),
```
PIG208, line 386: nit: use 2-space indentation
```suggestion
      _ => narrow.otherRecipientIds.map((id) =>
             store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
           ).join(', '),
```

PIG208 (commented):
Thanks for the revision! Just one more comment from me.

Regarding the question on `quoteAndReplyPlaceholder`, I think that one in particular is one of the places we have access to the message that contains the sender information. #716 mentioned that we should fallback to that if the user is unknown.

>  When we have a context like a Message that provides its own details on a user (like the message sender), we should fall back to those if the user is unknown. That way we not only don't crash, but provide helpful information, in cases like a guest user looking at older messages sent by someone no longer subscribed to their streams.

lib/widgets/inbox.dart
mofirojean: Okay, thanks for pointing that out 
Let me make the changes

PIG208:
We can also update the commit summary, because we haven' handled unknown users everywhere yet.

mofirojean:
> We should also update the commit summary, because we haven' handled unknown users everywhere yet.

About this, I checked other areas that didn't handle the null cases and it seems to me that they have been handled to resolve unknown users. Unless i might be confuse with what is required of me.

PIG208:
In this change we only start to use the translation string, but there are some other cases where we do not handle unknown users yet.

lib/widgets/inbox.dart
```diff
@@ -368,18 +369,21 @@ class _DmItem extends StatelessWidget {
...
       [] => selfUser.fullName,
-      [var otherUserId] => store.users[otherUserId]?.fullName ?? '(unknown user)',
+      [var otherUserId] => store.users[otherUserId]?.fullName ?? zulipLocalizations.unknownUserName,
 
       // TODO(i18n): List formatting, like you can do in JavaScript:
       //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Shu'])
       //   // 'Chris、Greg、Alya、Shu'
-      _ => narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', '),
+      _ => narrow.otherRecipientIds.map((id) =>
+            store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
```
PIG208, line 385: bump: we use 2 spaces for indentation

mofirojean: image

I am using android studio as my IDE and i configured it to use 2 spaces, and the changes I provided used this 2 spaces for indentation

mofirojean:
> In this change we only start to use the translation string, but there are some other cases where we do not handle unknown users yet.

Okay, I get it now. Please can you look this code from the `widget/content.dart,` which i referenced using `store.users[userId]`,

```dart
class AvatarImage extends StatelessWidget {
  ...
  @override
    final user = store.users[userId];
   
   /// In this case we have a condition to check if the user is null, does this mean
   ///  i need to add more logic to this or it handles the case as the issue describes ?
    if (user == null) { // TODO(log)
      return const SizedBox.shrink();
    }
  ...
  }
}

```

lib/widgets/inbox.dart
PIG208: I see. That should work for most cases automatically, but looks like at the line (started with `store.users[id]?.fullName`) I commented, there is only one space indented from the text it aligns to (started with `narrow.otherRecipientIds.map`) in the previous line.

You can count the number of spaces at the column where the `n` from `narrow.otherRecipientIds.map` is at, at line `store.users[id]?.fullName` to see what I mean.


I would suggest the following change:

before:
```dart
      _ => narrow.otherRecipientIds.map((id) =>
            store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
```

after:
```dart
      _ => narrow.otherRecipientIds.map((id) =>
             store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
```

PIG208:
Thanks for the update! For the question on `widget/content.dart`, I think in that case the unknown users have been taken care of.

lib/widgets/inbox.dart
mofirojean: I was able to add the changes as requested and i updated the PR description.

PIG208:
Looks good overall! This message `i18n [nfc]: Handle unknown users everywhere` doesn't seem to reflect on what this is doing though, because this commit is mostly about switching to localized strings.

How about something like `i18n: Use translation for "(unknown users)"`?
(Looking at the change, I realized that this is not an `nfc` because this is a behavior change to support languages other than English)

mofirojean:
> Looks good overall! This message `i18n [nfc]: Handle unknown users everywhere` doesn't seem to reflect on what this is doing, though, because this commit is mostly about switching to localized strings.
> 
> How about something like `i18n: Use translation for "(unknown users)"`? 

Yes that true, let me make the changes

gnprice (commented):
Thanks @mofirojean for taking care of this, and @PIG208 for the previous reviews!

Generally this looks good. A few small comments below.

The PR description also needs to be fixed (just like the commit message was earlier) so it doesn't say this will fix #716.

PIG208:
Hi @mofirojean! Thanks for working on this. Could you rebase and address the latest review comments?

mofirojean:
Okay thanks @PIG208 I will submit a PR with the changes

lib/widgets/recent_dm_conversations.dart
```diff
@@ -107,13 +108,15 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+        title = otherUser?.fullName ?? zulipLocalizations.unknownUserName;
         avatar = AvatarImage(userId: otherUserId, size: _avatarSize);
       default:
         // TODO(i18n): List formatting, like you can do in JavaScript:
         //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya'])
         //   // 'Chris、Greg、Alya'
-        title = narrow.otherRecipientIds.map((id) => store.users[id]?.fullName ?? '(unknown user)').join(', ');
+        title = narrow.otherRecipientIds.map((id) =>
+            store.users[id]?.fullName ?? zulipLocalizations.unknownUserName
+          ).join(', ');
```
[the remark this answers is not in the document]
mofirojean: @gnprice I just made the changes you requested and i also rebased the code as you instructed

gnprice:
CI is failing — you'll need to resolve that.

This revision also now makes a giant reformatting change to one file:
```
$ git log --stat -1
commit 4ccd41ca864e02946b3896e46e5aba45759d4cc1 (HEAD -> main, pr/1012)
Author: mofirojean <mofirojean@gmail.com>
Date:   Wed Oct 23 22:21:09 2024 +0100

    i18n: Use translation for "(unknown users)"

 lib/widgets/emoji_reaction.dart          |    4 +-
 lib/widgets/inbox.dart                   |    8 +-
 lib/widgets/message_list.dart            | 1063 ++++++++++++++++----------------
 lib/widgets/profile.dart                 |    3 +-
 lib/widgets/recent_dm_conversations.dart |    7 +-
 5 files changed, 549 insertions(+), 536 deletions(-)
```

That makes it very hard to read what the actual substantive change is. You'll need to fix it so it only makes the changes that are relevant to this PR.

mofirojean:
Hi @gnprice 
I am a little confused on the errors that arise with the CI failing, this happened after i rebased to keep the branch up to date.

Please if you have any suggestions on how to go about it, it will be much of a help to me.

gnprice:
The first errors are:
```
Running analyze...
Analyzing zulip-flutter...                                      

  error • Target of URI doesn't exist: 'package:flutter_gen/gen_l10n/zulip_localizations.dart' • lib/widgets/emoji_reaction.dart:2:8 • uri_does_not_exist
  error • Target of URI doesn't exist: 'package:flutter_gen/gen_l10n/zulip_localizations.dart' • lib/widgets/inbox.dart:2:8 • uri_does_not_exist
  error • Undefined name 'ZulipLocalizations' • lib/widgets/inbox.dart:364:32 • undefined_identifier
  error • Undefined name 'ZulipLocalizations' • lib/widgets/recent_dm_conversations.dart:83:32 • undefined_identifier

4 issues found. (ran in 24.1s)
```

That's because `package:flutter_gen` no longer exists. See https://github.com/zulip/zulip-flutter/commit/661adfc7b86df7fb46da7baac344ec021e6224d0 and https://github.com/zulip/zulip-flutter/pull/1058, which also demonstrate how to update code that was using it.

---

In general I recommend asking questions like this one in the #mobile-dev-help channel on chat.zulip.org. That way more people see it and are able to answer, so you're likely to get an answer sooner.

I also see and reply to messages in that channel (and the other mobile channels) sooner than I'm able to do for comments in GitHub, because Zulip makes it easier to keep up with relevant conversations when there's a lot happening than GitHub or its email notifications do.

mofirojean:
> The first errors are:
> 
> ```
> Running analyze...
> Analyzing zulip-flutter...                                      
> 
>   error • Target of URI doesn't exist: 'package:flutter_gen/gen_l10n/zulip_localizations.dart' • lib/widgets/emoji_reaction.dart:2:8 • uri_does_not_exist
>   error • Target of URI doesn't exist: 'package:flutter_gen/gen_l10n/zulip_localizations.dart' • lib/widgets/inbox.dart:2:8 • uri_does_not_exist
>   error • Undefined name 'ZulipLocalizations' • lib/widgets/inbox.dart:364:32 • undefined_identifier
>   error • Undefined name 'ZulipLocalizations' • lib/widgets/recent_dm_conversations.dart:83:32 • undefined_identifier

Okay thanks @gnprice , i will look into it

gnprice:
The use of "(unknown user)" everywhere has now been hand as part of #1148, namely in commit 76425ed6a1d0426b97421dc0b703ad83199199ea. I'll close this PR, because that's what this PR was working toward.

@mofirojean if you'd like to make another try at contributing to Zulip, you're welcome to either try #716 again, or pick up a different issue.

mofirojean:
Okay, thank you for the opportunity

Issue #979, opened 2024-10-03
Loading indicator does not go away when the event queue appears to be active


PIG208:
We added a loading indicator in #852 that appears when even queue polling fails with any error.

However, it seems that the loading indicator can stay on even the event queue is doing fine, and does not go away until you get a new event.

I don't have a consistent way to reproduce this, but what I did was basically:

1. Leave the app on and lock the phone
2. After a while, unlock the phone
3. Observe if the loading indicator stays there longer than usual
4. Try to do something (like starring a message), and the loading indicator should go away promptly

My guess is that when the event queue goes stale, `isLoading` gets set to true. When we do get the new event queue, we do not set `isLoading` back to false.

gnprice:
Thanks for the report!

I don't think I've seen this symptom myself yet. It sounds reminiscent of this other event-queue-related symptom I saw once back in July:
* #979

but I haven't seen that one since then either.

I think this won't be actionable until we hear more reports of it happening, so I'll push it out to the post-launch milestone. If we do see or hear of it again, this thread will help us aggregate those reports.

gnprice:
I happened to see this symptom yesterday during development, and thanks to seeing the debug log I think I have a diagnosis. This was on my Pixel 8 running Android 15, while connected by USB to my desktop for debugging.

What happens is:
* The device goes to sleep, perhaps while the app was in the foreground.
* The app actually keeps running — one can tell because it keeps printing log messages to the console.
* But it isn't able to make network requests.
  * The symptom of this is that the requests fail with SocketException, generally with DNS errors that it can't find `chat.zulip.org` (or whatever the host is).
    * We'd previously seen that those errors are common; that's why they're the one type of transient failure we silenced from reporting to the user, in #868.
  * In retrospect this makes sense as a compromise for the OS to choose: let the foreground app keep running, at least for a little while, in case the user wakes the device right back up again; but disable the network, because the wifi radio and especially the cell radio are expensive for the battery.
* So our event-poll loop keeps getting failures, and backs off farther and farther. So far, so good.
* Now the user wakes the device, and the app is able to use the network again. Two things go wrong that both contribute to this symptom.
* First, we may be in the middle of a long backoff. So it could be up to 10 seconds (typically more like 3 seconds) before we make the next request. During the remaining backoff time, the loading indicator will continue; and we won't learn of new events, even events that already happened while we were offline.
  * This first issue, I was basically already aware of. It'd be good to fix but isn't the main cause of the symptom this thread is about.
[...]

PR #1013, opened 2024-10-20
store: Make isLoading false after getting a event queue

satyam372:
Fixes: zulip#979

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -504,6 +504,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
     switch (event) {
       case HeartbeatEvent():
         assert(debugLog("server event: heartbeat"));
+        isLoading = false;
 
       case RealmEmojiUpdateEvent():
         assert(debugLog("server event: realm_emoji/update"));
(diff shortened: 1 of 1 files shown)


gnprice:
Is there a particular question you'd like feedback on related to this draft PR?

satyam372:
Yeah, I just want feedback on the changes. Before attempting to solve the issue, I followed the steps mentioned by @PIG208 in the description of issue #979:

**I don't have a consistent way to reproduce this, but what I did was basically:

1.Leave the app on and lock the phone.
2.After a while, unlock the phone.
3.Observe if the loading indicator stays visible longer than usual.
4.Try to do something (like starring a message), and the loading indicator should go away promptly.**

I tried this, but it worked fine for me; there was no visible loading indicator (or maybe I did it incorrectly).
So, I'm not sure if these changes will solve the issue or not.

This PR doesn't include tests yet. I will include them and push the revision soon.

gnprice:
Yes, as that description says, we don't know a way to consistently reproduce the issue.

The main work that's needed for making progress on this issue will be to debug it: to figure out how and why it happens, or at least reach some sort of theory for why it might happen. That requires either experimenting to find a reproducer, or carefully studying the code and thinking through possible scenarios.

This PR adds `isLoading = false` lines in a couple of arbitrary places in the code. Without a theory of why those would affect the issue, that isn't helpful.

Closing this PR because it isn't going in a useful direction.

PR #1017, opened 2024-10-22
dialog: Use Cupertino-flavored alert dialogs on iOS

### LINKED ISSUE #996: dialog: Use Cupertino-flavored alert dialogs on iOS
We should experiment with the [`AlertDialog.adaptive`](https://api.flutter.dev/flutter/material/AlertDialog/AlertDialog.adaptive.html) constructor, instead of [`AlertDialog`](https://api.flutter.dev/flutter/material/AlertDialog/AlertDialog.html), which forces a Material-style dialog on iOS:

| Current | Should be more like |
| --- | --- |
| image | image |

The "after" screenshot was made by simply changing `AlertDialog(` to `AlertDialog.adaptive(`. A complete fix will be a little more complicated than that, though. See the code sample on the [`AlertDialog.adaptive` doc](https://api.flutter.dev/flutter/material/AlertDialog/AlertDialog.adaptive.html), which shows a platform switch for the dialog's action buttons, to style those appropriately (including the buttons' on-press appearance):

```dart
  Widget adaptiveAction(
      {required BuildContext context,
      required VoidCallback onPressed,
      required Widget child}) {
    final ThemeData theme = Theme.of(context);
    switch (theme.platform) {
      case TargetPlatform.android:
      case TargetPlatform.fuchsia:
      case TargetPlatform.linux:
      case TargetPlatform.windows:
        return TextButton(onPressed: onPressed, child: child);
      case TargetPlatform.iOS:
      case TargetPlatform.macOS:
        return CupertinoDialogAction(onPressed: onPressed, child: child);
    }
  }
```

Relatedly, I think the `TextAlign.end` in our `_dialogActionText` helper should not be applied on iOS because on iOS the button text is meant to be center-aligned.

I also notice that our "Source Sans 3" font isn't being applied in the Cupertino-style dialog in the screenshot. We'll want to debug and fix that.

u7088495:
This pull request closes #996.<br>

In `dialog.dart`:<br>
Switched `alertDialog(` to `alertDialog.adaptive(`, as per #996.<br>
Defined new private widget `_adaptiveAction` which displays a  `CupertinoDialogAction` for  IOS and a `TextButton` otherwise.  `_adaptiveAction(` is used in place of `TextButton(` when stating dialog actions. <br><br>

The new result for IOS:
Screenshot 2024-10-22 at 3 31 13 pm
<br>

`dialog-checks.dart` has been updated to check the text content of the respective dialog types depending on the platform


lib/widgets/dialog.dart
```diff
@@ -15,6 +16,30 @@ Widget _dialogActionText(String text) {
...
+      case TargetPlatform.linux:
+        return TextButton(onPressed: onPressed, child: child);
+      case TargetPlatform.windows:
+        return TextButton(onPressed: onPressed, child: child);
+      case TargetPlatform.iOS:
+        return CupertinoDialogAction(onPressed: onPressed, child: child);
+      case TargetPlatform.macOS:
+        return CupertinoDialogAction(onPressed: onPressed, child: child);
+    }
+  }
```
chrisbobbe, line 41: This can be tightened up in a few ways:

- Use empty `case`s to fall through. From the [doc on Dart `switch` statements](https://dart.dev/language/branches#switch-statements):
  
  > https://dart.dev/language/branches#switch-statements
- Use `defaultTargetPlatform` instead of passing through `context`. When the app is run on a device, `defaultTargetPlatform` will match the platform (iOS or Android). When we need to simulate a specific platform in tests, we set `defaultTargetPlatform`; search for `debugDefaultTargetPlatformOverride` for how we do this.
- More concise dartdoc

Also, because the logic in `_dialogActionText` fits and is recommended for the Material-style dialog, let's only apply it on Android. This helper is a fine place for that conditional logic; how about changing its interface so it takes a String instead of a Widget, and applies `_dialogActionText` in the Android branch. In my proposal below, I've also given `_dialogActionText` an appropriately specific name, `_materialDialogActionText`. That helper could even be inlined, perhaps in a followup [NFC](https://github.com/zulip/zulip-mobile/blob/main/docs/glossary.md#user-content-nfc) commit.

So, putting all that together:

```suggestion
/// A platform-appropriate action for [AlertDialog.adaptive]'s [actions] param.
Widget _adaptiveAction({required VoidCallback onPressed, required String text}) {
  switch (defaultTargetPlatform) {
    case TargetPlatform.android:
    case TargetPlatform.fuchsia:
    case TargetPlatform.linux:
    case TargetPlatform.windows:
      return TextButton(onPressed: onPressed, child: _materialDialogActionText(text));
    case TargetPlatform.iOS:
    case TargetPlatform.macOS:
      return CupertinoDialogAction(onPressed: onPressed, child: Text(text));
  }
}
```

test/widgets/dialog_checks.dart
```diff
@@ -6,21 +8,41 @@ import 'package:flutter_test/flutter_test.dart';
 /// Checks for an error dialog matching an expected title
 /// and, optionally, matching an expected message. Fails if none is found.
 ///
-/// On success, returns the widget's "OK" button.
+/// On success, returns the widget's "OK" button
+/// (which is a [CupertinoDialogAction] for OS platforms).
```
chrisbobbe, line 12: We can leave this dartdoc unchanged. It doesn't matter what specific widget the button is, as long as it responds to taps.

test/widgets/dialog_checks.dart
```diff
@@ -6,21 +8,41 @@ import 'package:flutter_test/flutter_test.dart';
...
+  else {
+    final dialog = tester.widget<Dialog>(find.byType(Dialog));
+    tester.widget(find.widgetWithText(Dialog, expectedTitle));
+    if (expectedMessage != null) {
+      tester.widget(find.widgetWithText(Dialog, expectedMessage));
+    }
+    return tester.widget(
+      find.descendant(of: find.byWidget(dialog),
+        matching: find.widgetWithText(TextButton, 'OK')));
+  }
```
chrisbobbe, line 46: Let's use an exhaustive `switch` on `defaultTargetPlatform`, like we do elsewhere. Also, there are several formatting nits that makes this code harder to read than it needs to be.

Proposal:

```dart
  switch (defaultTargetPlatform) {
    case TargetPlatform.android:
    case TargetPlatform.fuchsia:
    case TargetPlatform.linux:
    case TargetPlatform.windows: {
      final dialog = tester.widget<Dialog>(find.byType(Dialog));
      tester.widget(find.widgetWithText(Dialog, expectedTitle));
      if (expectedMessage != null) {
        tester.widget(find.widgetWithText(Dialog, expectedMessage));
      }
      return tester.widget(
        find.descendant(of: find.byWidget(dialog),
          matching: find.widgetWithText(TextButton, 'OK')));
    }
    case TargetPlatform.iOS:
    case TargetPlatform.macOS: {
      final dialog = tester.widget<CupertinoAlertDialog>(
        find.byType(CupertinoAlertDialog));
      tester.widget(find.descendant(matchRoot: true,
        of: find.byWidget(dialog.title!), matching: find.text(expectedTitle)));
      if (expectedMessage != null) {
        tester.widget(find.descendant(matchRoot: true,
          of: find.byWidget(dialog.content!), matching: find.text(expectedMessage)));
      }
      return tester.widget(find.descendant(of: find.byWidget(dialog),
        matching: find.widgetWithText(CupertinoDialogAction, 'OK')));
    }
  }
```

pubspec.lock
```diff
@@ -315,10 +315,10 @@ packages:
     dependency: transitive
     description:
       name: file
-      sha256: "5fc22d7c25582e38ad9a8515372cd9a93834027aacf1801cf01164dac0ffa08c"
+      sha256: a3b4f84adafef897088c160faf7dfffb7696046cb13ae90b508c2cbc95d3b8d4
```
chrisbobbe, line 318: The changes to this file don't look related; please remove them.

chrisbobbe (commented):
Thanks! Comments below.

Also, please [tidy up](https://zulip.readthedocs.io/en/latest/git/using.html#examine-and-tidy-your-commit-history) the branch's commit history for clear and coherent commits. Each commit should be clean and pass all tests (you can run our tests with `tools/check`).

pubspec.lock
u7088495: Ok thankyou for the feedback, will be sure to onboard these points when working on issues in the future. I'll fix these things up in a new commit.

gnprice:
(Misclick, sorry)

lib/widgets/dialog.dart
u7088495: Is this the idea for inlining the `_materialDialogActionText`?
```dart
return TextButton(onPressed: onPressed, child: Text(text, textAlign: TextAlign.end));
```
will update the commit message of the latest commit as well as per the guidelines

lib/widgets/dialog.dart
```diff
@@ -3,28 +3,14 @@ import 'package:flutter/material.dart';
...
-Widget _materialDialogActionText(String text) {
-  return Text(
-    text,
-
-    // As suggested by
-    //   https://api.flutter.dev/flutter/material/AlertDialog/actions.html :
-    // > It is recommended to set the Text.textAlign to TextAlign.end
-    // > for the Text within the TextButton, so that buttons whose
-    // > labels wrap to an extra line align with the overall
-    // > OverflowBar's alignment within the dialog.
```
chrisbobbe, line 13: This comment is still important because it explains why the code is the way it is. Please move this comment along with the rest of the code you're moving, instead of deleting it.

pubspec.lock
```diff
@@ -318,7 +318,7 @@ packages:
       sha256: "5fc22d7c25582e38ad9a8515372cd9a93834027aacf1801cf01164dac0ffa08c"
       url: "https://pub.dev"
     source: hosted
-    version: "7.0.0"
+    version: "7.0.1"
```
chrisbobbe, line 321: This unrelated change still hasn't been removed.

chrisbobbe (commented):
Thanks. You'll need to tidy up your branch, as I mentioned [above](https://github.com/zulip/zulip-flutter/pull/1017#pullrequestreview-2390773805), before we can review this again. If you need help, please ask in `#git help` in the [development community](https://zulip.com/development-community/).

u7088495:
> Thanks. You'll need to tidy up your branch, as I mentioned [above](https://github.com/zulip/zulip-flutter/pull/1017#pullrequestreview-2390773805), before we can review this again. If you need help, please ask in `#git help` in the [development community](https://zulip.com/development-community/).

yep my bad, will do so. Thanks for your patients with me on this.

test/widgets/dialog_checks.dart
```diff
@@ -6,21 +8,40 @@ import 'package:flutter_test/flutter_test.dart';
 /// Checks for an error dialog matching an expected title
 /// and, optionally, matching an expected message. Fails if none is found.
 ///
-/// On success, returns the widget's "OK" button.
+/// On success, returns the widget's "OK" button
+///
```
chrisbobbe, line 12: nit: keep ".", and don't add a blank line

test/widgets/dialog_checks.dart
```diff
@@ -6,21 +8,40 @@ import 'package:flutter_test/flutter_test.dart';
...
+      }
+      return tester.widget(find.descendant(of: find.byWidget(dialog),
+        matching: find.widgetWithText(CupertinoDialogAction, 'OK')));
+    }
   }
-
-  return tester.widget(
-    find.descendant(of: find.byWidget(dialog),
-      matching: find.widgetWithText(TextButton, 'OK')));
+
```
chrisbobbe, line 46: nit: remove blank line

lib/widgets/dialog.dart
```diff
@@ -3,28 +3,20 @@ import 'package:flutter/material.dart';
...
     case TargetPlatform.linux:
     case TargetPlatform.windows:
-      return TextButton(onPressed: onPressed, child: _materialDialogActionText(text));
+      // As suggested by
+      //   https://api.flutter.dev/flutter/material/AlertDialog/actions.html :
+      // > It is recommended to set the Text.textAlign to TextAlign.end
+      // > for the Text within the TextButton, so that buttons whose
+      // > labels wrap to an extra line align with the overall
+      // > OverflowBar's alignment within the dialog.
+      return TextButton(onPressed: onPressed, child: Text(text, textAlign: TextAlign.end));
```
chrisbobbe, line 19: nit: There's a lot happening on this line of code. Let's arrange it so the comment is on just the specific thing that it's about, like it was before:

```dart
    // As suggested by
    //   https://api.flutter.dev/flutter/material/AlertDialog/actions.html :
    // > It is recommended to set the Text.textAlign to TextAlign.end
    // > for the Text within the TextButton, so that buttons whose
    // > labels wrap to an extra line align with the overall
    // > OverflowBar's alignment within the dialog.
    textAlign: TextAlign.end,
```

chrisbobbe (changes requested):
Thanks, this is much closer! Small comments below.

Also, a few commit message nits:

```
dialog: display adaptive dialogs and action buttons based on the target platform

AlertDialog was changed to AlertDialog.adaptive to the effect described in #996.
_adaptiveAction was implemented to display a platform appropriate action for
AlertDialog.adaptive's actions param, as was also discussed in #996.
tests in dialog_test were updated to perform platform appropriate tests.
```

- This commit fixes an issue (🎉), so let's put a `Fixes: #996` line at the end of it.
- Also, I think the paragraph ("AlertDialog was changed…") doesn't add anything that's not already obvious from reading the code changes and the linked issue, so let's just delete it.

```
dialog [nfc]: inline _materialDialogActionTest in _adaptiveAction
    
As was suggested in a comment of the pull request https://github.com/zulip/zulip-flutter/pull/1017#discussion_r1813819656.
```

- The URL should be on a new line; we try to wrap to 68 columns except where doing so would make things more confusing. How about:

      As suggested at:
        https://github.com/zulip/zulip-flutter/pull/1017#discussion_r1813819656 

Then for both commit messages, use initial caps for the part after the prefix, so:

- `dialog: Display adaptive dialogs and action buttons based on the target platform`
- `dialog [nfc]: inline _materialDialogActionTest in _adaptiveAction`

For examples of commit messages in the project's style, see the project's Git history; I recommend Greg's excellent [tip](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/git.md#2-the-secret-to-using-git-log--p) about how to do that.

lib/widgets/dialog.dart
u7088495: alrighty I think I've fixed up all those nits

u7332137 (approved):

lib/widgets/dialog.dart
```diff
@@ -1,18 +1,32 @@
...
+          // > It is recommended to set the Text.textAlign to TextAlign.end
+          // > for the Text within the TextButton, so that buttons whose
+          // > labels wrap to an extra line align with the overall
+          // > OverflowBar's alignment within the dialog.
+          textAlign: TextAlign.end));
+    case TargetPlatform.iOS:
+    case TargetPlatform.macOS:
+      return CupertinoDialogAction(
+        onPressed: onPressed, 
+        child: Text(text));
```
chrisbobbe, line 28:
```
dialog [nfc]: Inline _materialDialogActionTest in _adaptiveAction
```

nit: remove space character at end of line:

```suggestion
      return CupertinoDialogAction(
        onPressed: onPressed,
        child: Text(text));
```

or, perhaps better, just leave out this change; this code was fitting fine on just one line.

lib/widgets/dialog.dart
```diff
@@ -1,18 +1,32 @@
...
+Widget _adaptiveAction({required VoidCallback onPressed, required String text}) {
+  switch (defaultTargetPlatform) {
+    case TargetPlatform.android:
+    case TargetPlatform.fuchsia:
+    case TargetPlatform.linux:
+    case TargetPlatform.windows:
+      return TextButton(
+        onPressed: onPressed, 
+        child: Text(
+          text,
```
chrisbobbe, line 16: nit: no space character at end of line:

```suggestion
      return TextButton(
        onPressed: onPressed,
        child: Text(
          text,
```

chrisbobbe (commented):
Thanks!

Some new nits below, and also some more commit-message nits that I should have caught last time 🙂:

-----

```
dialog: Display adaptive dialogs and action buttons based on the target platform
```

This summary line is too long, at 80 characters. How about:

```
dialog: Use Cupertino-flavored alert dialogs on iOS
```

This also has a bit more information: the fact that iOS was the platform getting the wrong-style dialog. (The fact that the dialog's buttons match the rest of the dialog isn't surprising enough to need a mention. 🙂)

For what the length limit actually is, see [discussion](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/max.20commit-message.20summary.20line.20length/near/1971493), which I've just started 🙂. This was a helpful opportunity for me to spot a change in the zulip/zulip documentation that I'd missed!

-----

```
dialog [nfc]: Inline _materialDialogActionTest in _adaptiveAction
```

The function's name is `_materialDialogActionText`, not `_materialDialogActionTest`.

u7088495:
> Thanks!
> 
> Some new nits below, and also some more commit-message nits that I should have caught last time 🙂:
> 
> ```
> dialog: Display adaptive dialogs and action buttons based on the target platform
> ```
> 
> This summary line is too long, at 80 characters. How about:
> 
> ```
> dialog: Use Cupertino-flavored alert dialogs on iOS
> ```
> 
> This also has a bit more information: the fact that iOS was the platform getting the wrong-style dialog. (The fact that the dialog's buttons match the rest of the dialog isn't surprising enough to need a mention. 🙂)
> 
> For what the length limit actually is, see [discussion](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/max.20commit-message.20summary.20line.20length/near/1971493), which I've just started 🙂. This was a helpful opportunity for me to spot a change in the zulip/zulip documentation that I'd missed!
> 
> ```
> dialog [nfc]: Inline _materialDialogActionTest in _adaptiveAction
> ```
> 
> The function's name is `_materialDialogActionText`, not `_materialDialogActionTest`.

all sorted :)

chrisbobbe:

u7088495:
> Thanks, LGTM! Marking for Greg's review.

Thanks so much for your help through this process, It's been hugely helpful to me!

test/widgets/dialog_checks.dart
```diff
@@ -12,15 +14,32 @@ Widget checkErrorDialog(WidgetTester tester, {
...
+      tester.widget(find.widgetWithText(Dialog, expectedTitle));
+      if (expectedMessage != null) {
+        tester.widget(find.widgetWithText(Dialog, expectedMessage));
+      }
+      return tester.widget(
+        find.descendant(of: find.byWidget(dialog),
+          matching: find.widgetWithText(TextButton, 'OK')));
+    }
+    case TargetPlatform.iOS:
+    case TargetPlatform.macOS: {
```
gnprice, line 32: nits: no need for braces; and use a blank line to visually separate the two case blocks, since they're both several lines long

```suggestion
          matching: find.widgetWithText(TextButton, 'OK')));

    case TargetPlatform.iOS:
    case TargetPlatform.macOS:
```

test/widgets/dialog_checks.dart
```diff
@@ -12,15 +14,32 @@ Widget checkErrorDialog(WidgetTester tester, {
...
-      of: find.byWidget(dialog.content!), matching: find.text(expectedMessage)));
+  switch (defaultTargetPlatform) {
+    case TargetPlatform.android:
+    case TargetPlatform.fuchsia:
+    case TargetPlatform.linux:
+    case TargetPlatform.windows: {
+      final dialog = tester.widget<Dialog>(find.byType(Dialog));
+      tester.widget(find.widgetWithText(Dialog, expectedTitle));
+      if (expectedMessage != null) {
+        tester.widget(find.widgetWithText(Dialog, expectedMessage));
```
gnprice, line 25: This test logic in the non-Cupertino case differs from the old test logic. But the behavior in the non-Cupertino case is supposed to be the same as it was before, right? What's the motivation for changing the test logic?

gnprice: Ah I see, the old version fails because there's no longer a widget whose exact type is `AlertDialog` — the widget is an instance of a subclass instead — and `find.byType` searches by exact type.

A clean solution to that is to search by type in a way that includes subclasses. With the original tests plus just this tweak, all the tests pass:
```diff
-  final dialog = tester.widget<AlertDialog>(find.byType(AlertDialog));
+  final dialog = tester.widget<AlertDialog>(find.bySubtype<AlertDialog>());
```

I found `bySubtype` by looking at the doc for `byType`, and in particular the "See also" section at the end:
```dart
  /// See also:
  /// * [bySubtype], which allows subtype tests.
```

gnprice (commented):
Thanks @u7088495, and thanks @chrisbobbe for the previous reviews!

Generally these changes look good. Comments below.

Then there's one key thing this is missing before we can merge it: it needs tests. The current revision updates existing tests so that they continue to pass, but we also need tests that confirm the new logic is working correctly. That's how we ensure that a future change doesn't accidentally break this functionality.

A quick way to confirm that this version does not test the new functionality is to make the new code path artificially fail, like this:
```diff
     case TargetPlatform.iOS:
     case TargetPlatform.macOS:
+      throw 1;
       return CupertinoDialogAction(onPressed: onPressed, child: Text(text));
```

With that, the tests still pass. This PR needs to add a test or two that would fail with that change.

The tests should also fail if the changes get reverted. Try `git checkout origin lib/widgets/dialog.dart` to temporarily undo all your changes except the tests, then run `flutter test` to confirm that your new tests fail on the old code.

gnprice:
@u7088495 Do you think you'll have time to come back to this PR and write a couple of tests (and handle the small other comments above)?

It seems like a fair amount of work has gone into this so far; it'd be great to get it finished and merged.

(The issue this is for is in the M7 Future milestone, and we're now [focusing primarily on launch issues](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556) until the app's launch is ready. But I'm happy to make an exception for finishing this PR, if you're up for doing the remaining work needed to complete it.)

BrynMtchll:
Hey @gnprice yep! I've got some time at the moment to finish it off - thanks for making the exception.

using `debugDefaultTargetPlatformOverride` to test the new logic works, though if one of the test fails, `debugDefaultTargetPlatformOverride` isn't reset to null, causing a cascade of `debugAssertAllFoundationVarsUnset` in subsequent tests.
So it seems like using `TargetPlatformVariant`, like you suggested in a Zulip channel [here](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/tests.20for.20specific.20platforms/near/1983768), is a better way to go, but that would require adding the variant to all encasulating `testWidgets()`, is that right?

u7088495:
Hey @gnprice yep! I've got some time at the moment to finish it off - thanks for making the exception.

using debugDefaultTargetPlatformOverride to test the new logic works, though if one of the test fails, debugDefaultTargetPlatformOverride isn't reset to null, causing a cascade of debugAssertAllFoundationVarsUnset in subsequent tests.
So it seems like using TargetPlatformVariant, like you suggested in a Zulip channel [here](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/tests.20for.20specific.20platforms/near/1983768), is a better way to go, but that would require adding the variant to all encapsulating testWidgets(), is that right?

gnprice:
It isn't necessary to add variants to all the tests that call `checkErrorDialog`. For test cases that are about parts of the code other than the logic for platform variation in alert dialogs, it's best for them to remain focused on the things they're about.

The key for testing this change will be to add a few new test cases that are focused on this new logic.

(For similar questions in the future, I recommend asking in the `#mobile-dev-help` channel in Zulip — I'm more likely to see that sooner, and also other people will see it who might answer.)

BrynMtchll:
Right, yep I will do so, thanks for the clarification.

test/widgets/dialog_checks.dart
```diff
@@ -47,19 +77,39 @@ void checkNoErrorDialog(WidgetTester tester) {
...
+      tester.widget(find.descendant(matchRoot: true,
+        of: find.byWidget(dialog.content!), matching: find.text(expectedMessage)));
+
+      final actionButton = tester.widget(find.descendant(of: find.byWidget(dialog),
+          matching: find.widgetWithText(CupertinoDialogAction, expectedActionButtonText ?? 'Continue')));
+
+      final cancelButton = tester.widget(find.descendant(of: find.byWidget(dialog),
+          matching: find.widgetWithText(CupertinoDialogAction, 'Cancel')));
+
+      return (actionButton, cancelButton);
```
gnprice, line 113: nit: group these as one stanza:
```suggestion
      final actionButton = tester.widget(find.descendant(of: find.byWidget(dialog),
          matching: find.widgetWithText(CupertinoDialogAction, expectedActionButtonText ?? 'Continue')));
      final cancelButton = tester.widget(find.descendant(of: find.byWidget(dialog),
          matching: find.widgetWithText(CupertinoDialogAction, 'Cancel')));
      return (actionButton, cancelButton);
```

That helps in seeing the overall organization, since there's now this extra layer of structure around these that is the `switch` statement.

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,53 @@
...
+    await tester.pump();
+    context = tester.element(find.byType(Placeholder));
+  }
+
+  group('showErrorDialog', () {
+    testWidgets('show error dialog', (tester) async {
+      await prepare(tester);
+      
+      String title = "Dialog Title";
+      String message = "Dialog message.";
```
gnprice, line 28: nit: where possible, declare final (or in fact const)

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,53 @@
...
+    addTearDown(testBinding.reset);
+
+    await tester.pumpWidget(const TestZulipApp(
+      child: Scaffold(body: Placeholder())));
+    await tester.pump();
+    context = tester.element(find.byType(Placeholder));
+  }
+
+  group('showErrorDialog', () {
+    testWidgets('show error dialog', (tester) async {
```
gnprice, line 24: nit: the test name says the same thing as the group name — no need for the group

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,53 @@
+import 'package:flutter/material.dart';
+import 'package:zulip/widgets/dialog.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../model/binding.dart';
+import 'dialog_checks.dart';
+import 'test_app.dart';
+
+void main() {
```
gnprice, line 9: Cool, these tests look broadly right.

Let's have them check one other important aspect: each of the buttons in the dialog does the intended thing if you tap it.

gnprice (commented):

lib/widgets/dialog.dart
```diff
@@ -64,18 +76,15 @@ void showSuggestedActionDialog({
...
       content: SingleChildScrollView(child: Text(message)),
       actions: [
-        TextButton(
+        _adaptiveAction(
           onPressed: () => Navigator.pop(context),
-          child: _dialogActionText(zulipLocalizations.dialogCancel)),
-        TextButton(
-          onPressed: () {
-            onActionButtonPress();
-            Navigator.pop(context);
```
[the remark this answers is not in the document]
gnprice: It looks like this was probably an error in resolving rebase conflicts. Mistakes there happen, but that's one of the reasons it's important to reread your changes when sending them for review:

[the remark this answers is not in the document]
gnprice: There are also some important tips that can make it a lot easier to resolve rebase/merge conflicts correctly, including a Git config setting (`merge.conflictstyle diff3`) that is tragically not the default but you should definitely set. Details here:
https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/rebase.20conflicts/near/1136429
https://chat.zulip.org/#narrow/stream/2-general/topic/navigating.20code.20review.3A.20Rebasing.20work/near/1737371

Also in the latter thread are tips about self-review specifically for the result of rebasing, starting at this message:
https://chat.zulip.org/#narrow/channel/2-general/topic/navigating.20code.20review.3A.20Rebasing.20work/near/1737381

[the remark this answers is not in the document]
u7088495: Yep that's my mistake, thanks for the materials - I'll make an effort to be more thorough about it

test/widgets/dialog_checks.dart
```diff
@@ -15,22 +17,50 @@ Widget checkErrorDialog(WidgetTester tester, {
...
-  check(find.byType(AlertDialog)).findsNothing();
+/// Checks that there is no dialog.
+/// Fails if one is found.
+void checkNoDialog(WidgetTester tester) {
+  switch (defaultTargetPlatform) {
+    case TargetPlatform.android:
+    case TargetPlatform.fuchsia:
+    case TargetPlatform.linux:
+    case TargetPlatform.windows:
+      check(find.byType(Dialog)).findsNothing();
```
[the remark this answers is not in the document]
u7088495: I'm fairly sure this works correctly - a `Dialog` widget is generated in the widget tree by `AlertDialog.adaptive`, and  the check failed when I moved it to `checkErrorDialog`.

Screenshot 2025-02-17 at 10 47 36 am

Would it still be better to use `find.byType`?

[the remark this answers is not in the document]
gnprice: Hmm, right, `AlertDialog` builds a `Dialog` child rather than subclassing it. And there are no subclasses of `Dialog`.

`find.byType` is fine, then.

Still, we can make this check stronger:
* No need for the switch — just unconditionally check there's no Dialog, and also no CupertinoAlertDialog. After all, if we did somehow end up with a Dialog on iOS or a CupertinoAlertDialog on Android, that would probably be a bug; and it definitely wouldn't be a case of there being "no dialog", so this `checkNoDialog` function should fail.
* Reading this code and comparing to `checkErrorDialog` and `checkSuggestedActionDialog` still makes me nervous that they don't seem to line up — it sure looks as if there might be cases where one of those passes and yet this also passes. (The only reason that's impossible is a fact about the details of how `AlertDialog` works, which is that it uses `Dialog` as a child.)

  To fix that, this should use have a `find.bySubtype<AlertDialog>` check too. The `find.byType(Dialog)` is a good additional check to make sure there isn't some unexpected other type of dialog, but it doesn't substitute for directly checking that the scenario in the other checks can't be happening.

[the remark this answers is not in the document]
u7088495: sure thing that all makes a lot of sense.
Should I also add in a `find.byType(AlertDialog)` check for the sake of thoroughness as well? To ensure that no dialog has been created using the `AlertDialog` constructor directly.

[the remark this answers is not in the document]
gnprice: The `find.bySubtype<AlertDialog>` check should cover that, right?

gnprice:
@u7088495 I see you've pushed a revision to this PR, but just to make sure we're on the same page: generally I'll assume the new version isn't yet meant to be ready for a re-review until you post a comment saying it is.

(In this case there's a test failure, too, so I think you are indeed still working on the revision.)

u7088495:
Yep! It's just about there, just checking over everything now.

u7088495:
Alright @gnprice I think it's ready for review now, though I'm not entirely sure about the way I've laid out the tests so there may well be a thing or two to improve there.

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+  TestZulipBinding.ensureInitialized();
+
+  late BuildContext context;
+
+  const String title = "Dialog Title";
+  const String message = "Dialog message.";
```
gnprice, line 16: nit: omit types when they're obvious:
```suggestion
  const title = "Dialog Title";
  const message = "Dialog message.";
```

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+
+  group('showErrorDialog', () {
+    testWidgets('show error dialog', (tester) async {
+      await prepare(tester);
+
+      showErrorDialog(context: context, title: title, message: message);
+      await tester.pump();
+      checkErrorDialog(tester, expectedTitle: title, expectedMessage: message);
+
+    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
```
gnprice, line 35: nit: no blank line at end of block
```suggestion
      checkErrorDialog(tester, expectedTitle: title, expectedMessage: message);
    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
```

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+
+    testWidgets('user closes error dialog', (tester) async {
+      await prepare(tester);
+
+      showErrorDialog(context: context, title: title, message: message);
+      await tester.pump();
+
+      Widget button = checkErrorDialog(tester, expectedTitle: title);
```
gnprice, line 43: same point as https://github.com/zulip/zulip-flutter/pull/1017#discussion_r1955562905

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+
+      checkSuggestedActionDialog(tester, expectedTitle: title, expectedMessage: message,
+        expectedActionButtonText: actionButtonText);
+    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
+
+    testWidgets('user presses action button', (tester) async {
+      await prepare(tester);
+
+      late String actionText;
+      const String expectedActionText = "Action performed!";
```
gnprice, line 68: This can be simplified, because the string isn't doing any work here:
```suggestion
      bool wasPressed = false;
```

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+      void onActionButtonPress() {
+        actionText = expectedActionText;
+      }
+      showSuggestedActionDialog(context: context, title: title, message: message,
+        actionButtonText: actionButtonText, onActionButtonPress: onActionButtonPress);
+      await tester.pump();
+
+      final (Widget actionButton, _) = checkSuggestedActionDialog(tester, expectedTitle: title, expectedMessage: message,
```
gnprice, line 76: nit: omit obvious types 
```suggestion
      final (actionButton, _) = checkSuggestedActionDialog(tester, expectedTitle: title, expectedMessage: message,
```

gnprice, line 76: nit: line too long — keep all the relevant information within 80 columns

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+        expectedActionButtonText: actionButtonText);
+      await tester.tap(find.byWidget(actionButton));
+      await tester.pump();
+      checkNoDialog(tester);
+      check(actionText).equals(expectedActionText);
+    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
+
+    testWidgets('user cancels', (tester) async {
```
gnprice, line 84: As written, I believe this test would pass even if we accidentally made the "cancel" button do the same thing as the other button. Let's have the test detect the difference there.

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+    testWidgets('show error dialog', (tester) async {
+      await prepare(tester);
+
+      showErrorDialog(context: context, title: title, message: message);
+      await tester.pump();
+      checkErrorDialog(tester, expectedTitle: title, expectedMessage: message);
+
+    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
+
+    testWidgets('user closes error dialog', (tester) async {
```
gnprice, line 37: This test should cover both platforms too.

test/widgets/dialog_checks.dart
```diff
@@ -15,22 +17,44 @@ Widget checkErrorDialog(WidgetTester tester, {
...
 }
 
-// TODO(#996) update this to check for per-platform flavors of alert dialog
-void checkNoErrorDialog(WidgetTester tester) {
+/// Checks that there is no dialog.
+/// Fails if one is found.
+void checkNoDialog(WidgetTester tester) {
+  check(find.byType(Dialog)).findsNothing();
   check(find.byType(AlertDialog)).findsNothing();
+  check(find.bySubtype<AlertDialog>()).findsNothing();
```
gnprice, line 56: These two are redundant.

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,98 @@
...
+      await prepare(tester);
+
+      showErrorDialog(context: context, title: title, message: message);
+      await tester.pump();
+
+      Widget button = checkErrorDialog(tester, expectedTitle: title);
+      await tester.tap(find.byWidget(button));
+      await tester.pump();
+      checkNoDialog(tester);
```
gnprice, line 46: In the current version, the analyzer fails here after the first or second commit. A PR should pass all tests at each commit of the PR, in order to keep the commits fully coherent:
https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#each-commit-must-be-coherent

Probably cleanest is to have the rename of `checkNoDialog`, and accompanying changes, happen as the first commit.

gnprice (commented):
Thanks! Comments below.

test/widgets/dialog_test.dart
u7088495: Do things like this also count as obvious types?
`const String title = "Dialog Title";`

u7088495: and so should var be used for 
`bool wasPressed = false;`?

gnprice: > `const String title = "Dialog Title";`

Yeah, just `const title = …`.



> and so should var be used for
> `bool wasPressed = false;`?

Ah good question. No, when the variable is going to be mutated we use the type instead of `var`.

One reason for that is that it's very common for the type a mutable variable needs to be more general than the one it's initialized with — for example it gets initialized to `null` but later mutated to some other value. Dart's type inference would use only the initializer to decide the value, so would give it the narrower type (and then a type error when later trying to store a value that doesn't fit that narrow type). By contrast for final or const variables, the type chosen by type inference is almost always the type we want.

u7088495: ok interesting, good to know, thanks

u7088495:
@gnprice ok I think I've made all those changes and cleaned up the commit history now

test/widgets/dialog_checks.dart
```diff
@@ -15,22 +17,42 @@ Widget checkErrorDialog(WidgetTester tester, {
...
-    find.descendant(of: find.byWidget(dialog),
-      matching: find.widgetWithText(TextButton, 'OK')));
+      return tester.widget(find.descendant(of: find.byWidget(dialog),
+        matching: find.widgetWithText(CupertinoDialogAction, 'OK')));
+  }
 }
 
-// TODO(#996) update this to check for per-platform flavors of alert dialog
-void checkNoErrorDialog(WidgetTester tester) {
-  check(find.byType(AlertDialog)).findsNothing();
```
gnprice, line 33: Ah, in https://github.com/zulip/zulip-flutter/pull/1017#discussion_r1978521358 I meant those two checks were redundant with each other. If there's no widget that's of a subtype of AlertDialog, there's certainly none that's of the exact type AlertDialog.

gnprice (commented):
Thanks. This looks good except one comment below.

One other question: see this message about how we'll be crediting commits in the upcoming Zulip release blog post:
[#general > Zulip 10.0 release credits: code contributions @ 💬](https://chat.zulip.org/#narrow/channel/2-general/topic/Zulip.2010.2E0.20release.20credits.3A.20code.20contributions/near/2115455)

As is, this PR would cause you to be credited like so (either in the upcoming release blog post, or in the one after that if it's merged after that release):
```
     3  Brynly Mitchell - u7088495
```

If that's how you'd like your name to appear, there's nothing more for you to do. If you'd like it written differently, you can use `git commit --amend --author=…` to change it in a rebase. Then you can [set your name in your Git config](https://chat.zulip.org/#narrow/channel/2-general/topic/Zulip.208.2E0.20release.20credits/near/1702524) to control the name that future commits get automatically.

gnprice:
I see you've pushed a revision since my last review. Is this intended to be ready again for merge/review?

(It's useful to make that clear with a comment, because people often push to a branch with intermediate revisions before pushing one they consider ready.)

BrynMtchll:
Ah yep It's just about ready but I just hadn't gotten around to doing a once-over to check everything looks alright.

BrynMtchll:
ok should be all good to go now :)

gnprice:
@BrynMtchll I'm guessing you're the same person as @u7088495? Please pick a single account to stick to, at least for interacting with any given project (like Zulip) — it makes things less confusing :slightly_smiling_face: 

It looks like this has some merge/rebase conflicts after 009497806f66c6580ddd79f5026c5a967b93dda0 / #1410. Would you rebase and resolve those? Then I think this will be all ready for merge.

u7088495:
Alright I think it's ready; I refactored the learn more test logic into `checkErrorDialog` as well.

lib/widgets/dialog.dart
```diff
@@ -46,17 +58,18 @@ DialogStatus showErrorDialog({
...
       content: message != null ? SingleChildScrollView(child: Text(message)) : null,
       actions: [
         if (learnMoreButtonUrl != null)
-          TextButton(
+          _adaptiveAction(
             onPressed: () => PlatformActions.launchUrl(context, learnMoreButtonUrl),
-            child: _dialogActionText(zulipLocalizations.errorDialogLearnMore)),
-        TextButton(
+            text: zulipLocalizations.errorDialogLearnMore,
+          ),
```
gnprice, line 69: nit: preserve formatting
```suggestion
            text: zulipLocalizations.errorDialogLearnMore),
```

lib/widgets/dialog.dart
```diff
@@ -1,20 +1,32 @@
...
+        child: Text(
+          text,
+          // As suggested by
+          //   https://api.flutter.dev/flutter/material/AlertDialog/actions.html :
+          // > It is recommended to set the Text.textAlign to TextAlign.end
+          // > for the Text within the TextButton, so that buttons whose
+          // > labels wrap to an extra line align with the overall
+          // > OverflowBar's alignment within the dialog.
+          textAlign: TextAlign.end));
+    case TargetPlatform.iOS:
```
gnprice, line 26: nit: a blank line to separate these cases helps make the structure easier to see:
```suggestion
          textAlign: TextAlign.end));

    case TargetPlatform.iOS:
```

test/widgets/dialog_checks.dart
```diff
@@ -14,25 +19,55 @@ import 'package:zulip/widgets/dialog.dart';
...
+        of: find.byWidget(dialog.title!), matching: find.text(expectedTitle)));
+      if (expectedMessage != null) {
+        tester.widget(find.descendant(matchRoot: true,
+          of: find.byWidget(dialog.content!), matching: find.text(expectedMessage)));
+      }
+      if (expectedLearnMoreButtonUrl != null) {
+        check(testBinding.takeLaunchUrlCalls()).single.equals((
+          url: expectedLearnMoreButtonUrl,
+          mode: LaunchMode.inAppBrowserView));
+      }
```
gnprice, line 40: Checking for this button is useful, but this check doesn't work in this context. The check will work only if the caller has already gone and tapped on the button. There's nothing in this function's dartdoc that says the caller should do that, and it's not something a reader would naturally assume they need to do.

If a caller is going and tapping on the button themself, then the caller can also easily go and do this check — the check isn't using any specific knowledge about how the error dialogs work.

It looks like there's one call site that passes this option. So let's move this check to that call site.

gnprice (commented):
Thanks for the revision! There's one substantive change needed in the new test code, and a couple of easy nits.

u7088495:
hey @gnprice, sorry I've stepped away from this - I've come back to it and think it's ready for another review now :).

test/widgets/dialog_test.dart
```diff
@@ -1,30 +1,77 @@
...
+        case TargetPlatform.linux:
+        case TargetPlatform.windows:
+          check(testBinding.takeLaunchUrlCalls()).single.equals((
+            url: learnMoreButtonUrl,
+            mode: LaunchMode.inAppBrowserView));
+        case TargetPlatform.iOS:
+        case TargetPlatform.macOS:
+          check(testBinding.takeLaunchUrlCalls()).single.equals((
+            url: learnMoreButtonUrl,
+            mode: LaunchMode.externalApplication));
```
gnprice: This is fairly verbose, in a way that obscures the fact that the only difference between the cases is the value of `mode`.

Instead, let's first determine teh mode we expect; then have just one, more focused, check that calls `takeLaunchUrlCalls`.

For examples, search the existing tests in test/ for "LaunchMode.externalApplication".

test/widgets/dialog_test.dart
```diff
@@ -1,30 +1,77 @@
...
       await tester.pump();
       await tester.tap(find.text('Learn more'));
-      check(testBinding.takeLaunchUrlCalls()).single.equals((
-        url: Uri.parse('https://foo.example'),
-        mode: LaunchMode.inAppBrowserView));
-    });
+
+      checkErrorDialog(tester, expectedTitle: title);
+
+      switch (defaultTargetPlatform) {
```
gnprice: These steps are out of their logical order. The dialog must be there before the `tester.tap` on one of its buttons can work; so let's check for the dialog before trying to tap one of its buttons.

u7088495:
alright should be ready for another revision :)

test/widgets/dialog_test.dart
```diff
@@ -1,30 +1,71 @@
...
+      final expectedMode = switch (defaultTargetPlatform) {
+        TargetPlatform.android => LaunchMode.inAppBrowserView,
+        TargetPlatform.iOS =>     LaunchMode.externalApplication,
+        _ => throw StateError('attempted to test with $defaultTargetPlatform'),
+      };
+
+      check(testBinding.takeLaunchUrlCalls()).single
+        .equals((url: learnMoreButtonUrl, mode: expectedMode));
+
+    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
```
gnprice: nit: no blank line at end of block
```suggestion
        .equals((url: learnMoreButtonUrl, mode: expectedMode));
    }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
```

test/widgets/dialog_test.dart
```diff
@@ -1,30 +1,71 @@
...
-        mode: LaunchMode.inAppBrowserView));
-    });
+
+      final expectedMode = switch (defaultTargetPlatform) {
+        TargetPlatform.android => LaunchMode.inAppBrowserView,
+        TargetPlatform.iOS =>     LaunchMode.externalApplication,
+        _ => throw StateError('attempted to test with $defaultTargetPlatform'),
+      };
+
+      check(testBinding.takeLaunchUrlCalls()).single
```
gnprice: nit: use blank lines to group test steps into stanzas of set up, then check.

See https://github.com/zulip/zulip-flutter/pull/1317#discussion_r1980467256 and the previous comments linked from there.

gnprice: Here's the adjusted version of this test which I just merged:
```dart
    testWidgets('tap "Learn more" button', (tester) async {
      await prepare(tester);

      final learnMoreButtonUrl = Uri.parse('https://foo.example');
      showErrorDialog(context: context, title: title, learnMoreButtonUrl: learnMoreButtonUrl);
      await tester.pump();
      checkErrorDialog(tester, expectedTitle: title);

      await tester.tap(find.text('Learn more'));
      final expectedMode = switch (defaultTargetPlatform) {
        TargetPlatform.android => LaunchMode.inAppBrowserView,
        TargetPlatform.iOS =>     LaunchMode.externalApplication,
        _ => throw StateError('attempted to test with $defaultTargetPlatform'),
      };
      check(testBinding.takeLaunchUrlCalls()).single
        .equals((url: learnMoreButtonUrl, mode: expectedMode));
```

Note how each stanza (each group separated by blank lines) first sets up a situation, then checks that it's as expected.

u7088495:
> Thanks @u7088495 for all your work on this! All looks good now except a couple of nits below. I'll fix those and merge.

Amazing, thank you so much for all of your help through this @gnprice, I really appreciate it, it's been a great experience for me.

PR #1018, opened 2024-10-22
deps: Upgrade Flutter to 3.27.0-1.0.pre.166

PIG208:

Built successfully on my Android device; still need to confirm that it works on IOS.
This contains a major version bump to the dart SDK ([changelog](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#370)).

3.7.0 adds [Wildcard Variables](https://github.com/dart-lang/language/issues/3712), that we currently doesn't seem to use.

There is also changes to the Dart to Javascript Compiler (dart2js) and Dart Development Compiler (dartdevc), both are "switched to use an AOT snapshot instead of a JIT snapshot". Neither of them applies to us because we do not use the web builds.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -315,10 +315,10 @@ packages:
     dependency: transitive
     description:
       name: file
-      sha256: "5fc22d7c25582e38ad9a8515372cd9a93834027aacf1801cf01164dac0ffa08c"
+      sha256: a3b4f84adafef897088c160faf7dfffb7696046cb13ae90b508c2cbc95d3b8d4
       url: "https://pub.dev"
     source: hosted
-    version: "7.0.0"
+    version: "7.0.1"
   file_picker:
     dependency: "direct main"
     description:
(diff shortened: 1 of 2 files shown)


chrisbobbe:
> still need to confirm that it works on IOS.

Yup, just tested on an iOS simulator and it works there too. Thanks; merging!

gnprice:
Thanks for taking care of this! Looks good; merging.

The Dart version bump is from 3.6.0-etc to 3.7.0-etc, so I'd call that a minor-version bump; the major version is just 3.

We definitely don't yet use wildcard variables, because as the changelog says:
> Dart 3.7 adds [wildcard variables](https://github.com/dart-lang/language/issues/3712) to the language. To use them, set your package's [SDK constraint](https://dart.dev/guides/language/evolution) lower bound to 3.7 or greater (`sdk: '^3.7.0'`).

they only become allowed once you bump the Dart version bound in your pubspec to at least 3.7.0, as this PR does :slightly_smiling_face:

PR #1019, opened 2024-10-22
lint: Re-enable analyzer on Pigeon-generated code; fix one revealed issue

gnprice:
We'd been suppressing the analyzer on Pigeon-generated code, because of an upstream issue. That issue has been fixed:
  https://github.com/flutter/flutter/issues/145633

In fact it turns out it was fixed in pigeon 18.0.0, which we took in 8df306412 back in April; we just didn't notice at the time.

Meanwhile this file grew a different analyzer issue during the time it wasn't getting checked — and that one was an actual actionable one for us to fix, where a doc comment still referred to a deprecated bit of API, namely [Color.value].  So fix that too.  Glad to have the analyzer re-enabled here.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -3,15 +3,6 @@
 include: package:flutter_lints/flutter.yaml
 
 analyzer:
-  exclude:
-    # Skip analysis on Pigeon-generated code, because it currently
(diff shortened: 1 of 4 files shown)


chrisbobbe:
LGTM, thanks! Merging.

Issue #933, opened 2024-09-09
Handle codehilite elements with multiple classes


PIG208:
It is totally possible for Pygments to [generate HTML elements with more than one classes](https://github.com/pygments/pygments/blob/07b69d42ca431249250195e7f7fbd21e1ccdfb2c/pygments/formatters/html.py#L853-L855).

This is considered quite an edge case and only happens to some languages, such as PostgreSQL (see https://github.com/pygments/pygments/issues/2780). It currently renders as an unimplemented feature whenever a code block contains tokens with more than one classes.

See a minimal example [here](https://chat.zulip.org/#narrow/stream/7-test-here/topic/tap/near/1939061).

HTML of the unsupported message:

```html
<pre>
  <span></span>
  <code>
    <span class="s s-Name">"</span>
  </code>
</pre>
```

PIG208:
Listed under #921.

gnprice:
Other than PostgreSQL, this apparently happens with YAML. Here's [a test message](https://chat.zulip.org/#narrow/stream/7-test-here/topic/Greg/near/1949014).

PIG208:
For these kind of messages, we should be able to safely apply the style of the most specific known class from the class list. These non-standard token types usually aren't likely to be handled by styles. We could also maintain a list of unexpected token types, but that does not seem necessary.

PR #1021, opened 2024-10-23
content: Handle code block spans with multiple CSS classes

rajveermalviya:
Fixes: #933

| Flutter | Web |
| - | - |
| Screenshot 2024-10-23 at 20 35 38 | Screenshot 2024-10-23 at 20 35 54 |

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -1023,9 +1023,17 @@ class _ZulipContentParser {
           span = CodeBlockSpanNode(text: text, type: CodeBlockSpanType.text);
 
         case dom.Element(localName: 'span', :final text, :final className):
-          final CodeBlockSpanType type = codeBlockSpanTypeFromClassName(className);
-          switch (type) {
-            case CodeBlockSpanType.unknown:
+          // Empirically, when a Pygments node has multiple classes, the first
+          // class names a standard token type and the rest are for non-standard
+          // token types specific to the language.  Zulip web only styles the
+          // standard token classes and ignores the others, so we do the same.
+          // See: https://github.com/zulip/zulip-flutter/issues/933
+          final spanType = className.split(' ')
+            .map(codeBlockSpanTypeFromClassName)
+            .firstWhereOrNull((e) => e != CodeBlockSpanType.unknown);
+
+          switch (spanType) {
+            case null:
               // TODO(#194): Show these as un-syntax-highlighted code, in production.
               return UnimplementedBlockContentNode(htmlNode: divElement);
             case CodeBlockSpanType.highlightedLines:
@@ -1033,7 +1041,7 @@ class _ZulipContentParser {
               //       inherited styles for `span.hll` nodes.
               return UnimplementedBlockContentNode(htmlNode: divElement);
             default:
-              span = CodeBlockSpanNode(text: text, type: type);
+              span = CodeBlockSpanNode(text: text, type: spanType);
           }
 
         default:
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -344,6 +344,23 @@ class ContentExample {
       ]),
     ]);
 
+  static const codeBlockSpansWithMultipleClasses = ContentExample(
+    'code block spans with multiple CSS classes',
+    '```yaml\n- item\n```',
+    expectedText: '- item',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Greg/near/1949014
+    '<div class="codehilite" data-code-language="YAML">'
+        '<pre><span></span><code><span class="p p-Indicator">-</span>'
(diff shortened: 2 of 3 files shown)


lib/model/content.dart
```diff
@@ -1023,17 +1023,28 @@ class _ZulipContentParser {
           span = CodeBlockSpanNode(text: text, type: CodeBlockSpanType.text);
 
         case dom.Element(localName: 'span', :final text, :final className):
```
PIG208, line 1025: nit: can use `dom.Element.classes` here

rajveermalviya:
Thanks for the review @PIG208! Pushed a new revision, PTAL :)

PIG208:
LGTM!

lib/model/content.dart
```diff
@@ -1022,18 +1022,28 @@ class _ZulipContentParser {
...
-        case dom.Element(localName: 'span', :final text, :final className):
-          final CodeBlockSpanType type = codeBlockSpanTypeFromClassName(className);
-          switch (type) {
-            case CodeBlockSpanType.unknown:
+        case dom.Element(localName: 'span', :final text, :final classes):
+          // The general form for Pygments nodes with multiple classes observed
+          // are, that the first class is the standard token class and trailing
+          // tokens are non-standard tokens. Zulip web client only styles the
+          // standard tokens classes. Same is done here, only standard token
+          // class names are emitted.
```
gnprice, line 1030:
```suggestion
          // Empirically, when a Pygments node has multiple classes, the first
          // class names a standard token type and the rest are for non-standard
          // token types specific to the language.  Zulip web only styles the
          // standard token classes and ignores the others, so we do the same.
```

The "are emitted" confused me: we don't emit any class names here, right? We consume class names, and may emit a CodeBlockSpanType value.

test/model/content_test.dart
```diff
@@ -395,6 +395,23 @@ class ContentExample {
...
+    'code block nodes with multiple class names',
+    '```yaml\n- item\n```',
+    expectedText: '- item',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Greg/near/1949014
+    '<div class="codehilite" data-code-language="YAML">'
+      '<pre><span></span><code><span class="p p-Indicator">-</span>'
+      '<span class="w"> </span>'
+      '<span class="l l-Scalar l-Scalar-Plain">item</span>\n'
+      '</code></pre></div>', [
+      CodeBlockNode([
```
gnprice, line 408: super nit: indent the string continuations deeper so they don't look like part of the same series as the list element below

```suggestion
    '<div class="codehilite" data-code-language="YAML">'
        '<pre><span></span><code><span class="p p-Indicator">-</span>'
        '<span class="w"> </span>'
        '<span class="l l-Scalar l-Scalar-Plain">item</span>\n'
        '</code></pre></div>', [
      CodeBlockNode([
```

This is similar to what we do with `if` conditions. You can also see it in action on the similar examples codeBlockHighlightedShort and codeBlockHighlightedMultiline above.

test/model/content_test.dart
```diff
@@ -395,6 +395,23 @@ class ContentExample {
       ParagraphNode(links: null, nodes: [TextNode("some content")]),
     ]);
 
+  static const codeBlockNodesWithMultipleClasses = ContentExample(
```
gnprice, line 398: nit: put this just after the two `codeBlockHighlighted…` examples, so it's not separated from them by all the unimplemented code-block examples

gnprice (commented):
Thanks @rajveermalviya for taking care of this, and @PIG208 for the previous review!

Generally this looks great. Small comments below, most of them nits.

rajveermalviya:
Thanks for the reviews @PIG208, @gnprice!
@gnprice Pushed a new revision, PTAL.

test/model/content_test.dart
```diff
@@ -344,6 +344,23 @@ class ContentExample {
       ]),
     ]);
 
+  static const codeBlockNodesWithMultipleClasses = ContentExample(
+    'code block nodes with multiple class names',
```
gnprice, line 348: Rereading, I realized this name is misleading: "code block node" should correspond `CodeBlockNode`, since we have such an identifier; but the HTML element that corresponds to one of those is the whole `div.codehilite`. The things that have multiple classes are the `span` great-grandchildren of that, which correspond to `CodeBlockSpanNode`.

I think "code block span" is a fine term for these.

So that applies to this example's name and description, and to the commit message.

gnprice (commented):
Thanks for the revision! Looks good, with just one comment below on something I missed earlier.

rajveermalviya:
Thanks for the review @gnprice. Revision pushed.

gnprice:
Thanks! All looks good — merging.

PR #1024, opened 2024-10-25
linux build: Disable a warning that url_launcher_linux can trigger

gnprice:
This works around the following issue in url_launcher_linux:
  https://github.com/flutter/flutter/issues/153083

The issue only appears on some hosts, namely those with "older versions of glib".


chrisbobbe:
Sure, SGTM; merging.

Issue #1026, opened 2024-10-25
test flake: Consecutive autogenerated email addresses are identical 0.1% of the time


chrisbobbe:
I just got this test failure when working on unrelated code (#683):

```
Running test...
00:10 +1123 ~6: /Users/chrisbobbe/dev/zulip-flutter/test/notifications/display_test.dart: NotificationDisplayManager open find account among several
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following AccountAlreadyExistsException was thrown running a test:
Instance of 'AccountAlreadyExistsException'

When the exception was thrown, this was the stack:
#0      TestGlobalStore.doInsertAccount (file:///Users/chrisbobbe/dev/zulip-flutter/test/model/test_store.dart:101:7)
#1      GlobalStore.insertAccount (package:zulip/model/store.dart:180:27)
#2      TestGlobalStore.add (file:///Users/chrisbobbe/dev/zulip-flutter/test/model/test_store.dart:84:11)
#3      main.<anonymous closure>.<anonymous closure> (file:///Users/chrisbobbe/dev/zulip-flutter/test/notifications/display_test.dart:900:39)
<asynchronous suspension>
#4      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:189:15)
<asynchronous suspension>
#5      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1027:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

The test description was:
  find account among several
════════════════════════════════════════════════════════════════════════════════════════════════════
00:10 +1123 ~6 -1: /Users/chrisbobbe/dev/zulip-flutter/test/notifications/display_test.dart: NotificationDisplayManager open find account among several [E]
  Test failed. See exception logs above.
  The test description was: find account among several
  

To run this test again: /Users/chrisbobbe/.local/lib/flutter/bin/cache/dart-sdk/bin/dart test /Users/chrisbobbe/dev/zulip-flutter/test/notifications/display_test.dart -p vm --plain-name 'NotificationDisplayManager open find account among several'
```

The bug is in the way we auto-generate email addresses for example `User` objects:

```dart
/// A random email address, different from previously generated ones.
String _nextEmail() => 'mail${_lastEmailSuffix += Random().nextInt(1000)}@example.com';
[...]

PR #1027, opened 2024-10-25
test: Fix flake with identical consecutively autogenerated email addresses

chrisbobbe:
Fixes: #1026


gnprice:
Thanks for fixing this! Looks good — merging.

@PIG208 FYI (re #940). Detailed diagnosis is in the issue #1026.

Issue #1028, opened 2024-10-26
Follow /with/ links through message moves


chrisbobbe:
Narrow links with `/with/<id>` are the upcoming "topic permalink" feature:
* https://github.com/zulip/zulip/issues/21505

These links look like:
  https://chat.zulip.org/#narrow/channel/7-test-here/topic/Chris/with/1970220

Since #684, we accept and ignore the `/with/<id>` element. This issue is for using it as intended, i.e., opening the `<id>` message's current conversation even if that's different from the stream/topic in the link because it was moved.

We can include this element when requesting messages from the server. That's helpful because the work of finding the message's current conversation is all done on the server in a single round-trip. The server's get-messages result is from the `<id>` message's current conversation, if we have access to it.

From the [API doc](https://zulip.com/api/construct-narrow#narrows-that-use-ids) on constructing a narrow (what we call `ApiNarrow` in our code):

> The `with` operator is designed to be used for permanent links to topics, which means they should continue to work when the topic is [moved](https://zulip.com/help/move-content-to-another-topic) or [resolved](https://zulip.com/help/resolve-a-topic). If the message with the specified ID exists, and can be accessed by the user, then it will return messages with the `channel`/`topic`/`dm` operators corresponding to the current conversation containing that message, and replacing any such filters included in the narrow.

So, when the user taps a topic permalink, we should:

- Open the message list, showing the link's stream/topic in the app bar (even if it might be stale)
- When making the initial message-fetch request, include the "with" element
- If the fetch result is from a different stream/topic than the one in the link, update the message list's narrow to the new stream/topic (which causes the app bar and other things to update)
- In any case, update the message list's narrow so the "with" element isn't sent in fetch-older requests

PR #1029, opened 2024-10-26
msglist: Follow /with/ links through message moves

chrisbobbe:
Fixes: #1028

### DIFF
--- a/lib/api/model/narrow.dart
+++ b/lib/api/model/narrow.dart
@@ -6,14 +6,24 @@ part 'narrow.g.dart';
 
 typedef ApiNarrow = List<ApiNarrowElement>;
 
-/// Resolve any [ApiNarrowDm] elements appropriately.
+/// Adapt the given narrow to be sent to the given Zulip server version.
 ///
-/// This encapsulates a server-feature check.
-ApiNarrow resolveDmElements(ApiNarrow narrow, int zulipFeatureLevel) {
-  if (!narrow.any((element) => element is ApiNarrowDm)) {
-    return narrow;
-  }
+/// Any elements that take a different name on old vs. new servers
+/// will be resolved to the specific name to use.
+/// Any elements that are unknown to old servers and can
+/// reasonably be omitted will be omitted.
+ApiNarrow resolveApiNarrowForServer(ApiNarrow narrow, int zulipFeatureLevel) {
   final supportsOperatorDm = zulipFeatureLevel >= 177; // TODO(server-7)
+
+  bool hasDmElement = false;
+  for (final element in narrow) {
+    switch (element) {
+      case ApiNarrowDm(): hasDmElement = true;
+      default:
+    }
+  }
+  if (!hasDmElement) return narrow;
+
   return narrow.map((element) => switch (element) {
     ApiNarrowDm() => element.resolve(legacy: !supportsOperatorDm),
     _             => element,
@@ -102,7 +112,7 @@ class ApiNarrowTopic extends ApiNarrowElement {
 /// and more generally its [operator] getter must not be called.
 /// Instead, call [resolve] and use the object it returns.
 ///
-/// If part of [ApiNarrow] use [resolveDmElements].
+/// If part of [ApiNarrow] use [resolveApiNarrowForServer].
 class ApiNarrowDm extends ApiNarrowElement {
   @override String get operator {
     assert(false,
--- a/lib/api/model/narrow.dart
+++ b/lib/api/model/narrow.dart
@@ -14,20 +14,33 @@ typedef ApiNarrow = List<ApiNarrowElement>;
 /// reasonably be omitted will be omitted.
 ApiNarrow resolveApiNarrowForServer(ApiNarrow narrow, int zulipFeatureLevel) {
   final supportsOperatorDm = zulipFeatureLevel >= 177; // TODO(server-7)
+  final supportsOperatorWith = zulipFeatureLevel >= 271; // TODO(server-9)
 
   bool hasDmElement = false;
+  bool hasWithElement = false;
   for (final element in narrow) {
     switch (element) {
-      case ApiNarrowDm(): hasDmElement = true;
+      case ApiNarrowDm():   hasDmElement = true;
+      case ApiNarrowWith(): hasWithElement = true;
       default:
     }
   }
-  if (!hasDmElement) return narrow;
+  if (!(hasDmElement || (hasWithElement && !supportsOperatorWith))) {
+    return narrow;
+  }
 
-  return narrow.map((element) => switch (element) {
-    ApiNarrowDm() => element.resolve(legacy: !supportsOperatorDm),
-    _             => element,
-  }).toList();
+  final result = <ApiNarrowElement>[];
+  for (final element in narrow) {
+    switch (element) {
+      case ApiNarrowDm():
+        result.add(element.resolve(legacy: !supportsOperatorDm));
+      case ApiNarrowWith() when !supportsOperatorWith:
+        break; // drop unsupported element
+      default:
+        result.add(element);
+    }
+  }
+  return result;
 }
 
 /// An element in the list representing a narrow in the Zulip API.
@@ -158,6 +171,22 @@ class ApiNarrowPmWith extends ApiNarrowDm {
   @override String get operator => 'pm-with';
 
   ApiNarrowPmWith._(super.operand, {super.negated});
+}
+
+/// An [ApiNarrowElement] with the 'with' operator.
+///
+/// If part of [ApiNarrow] use [resolveApiNarrowForServer].
+class ApiNarrowWith extends ApiNarrowElement {
+  @override String get operator => 'with';
+
+  @override final int operand;
+
+  ApiNarrowWith(this.operand, {super.negated});
+
+  factory ApiNarrowWith.fromJson(Map<String, dynamic> json) => ApiNarrowWith(
+    json['operand'] as int,
+    negated: json['negated'] as bool? ?? false,
+  );
 }
 
 class ApiNarrowIs extends ApiNarrowElement {
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -86,6 +86,8 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
         fragment.write('${element.operand.join(',')}-$suffix');
       case ApiNarrowDm():
         assert(false, 'ApiNarrowDm should have been resolved');
+      case ApiNarrowWith():
+        fragment.write(element.operand.toString());
       case ApiNarrowIs():
         fragment.write(element.operand.toString());
       case ApiNarrowMessageId():
@@ -160,6 +162,7 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
   ApiNarrowStream? streamElement;
   ApiNarrowTopic? topicElement;
   ApiNarrowDm? dmElement;
+  ApiNarrowWith? withElement;
   Set<IsOperand> isElementOperands = {};
 
   for (var i = 0; i < segments.length; i += 2) {
@@ -188,12 +191,17 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
         if (dmIds == null) return null;
         dmElement = ApiNarrowDm(dmIds, negated: negated);
 
+      case _NarrowOperator.with_:
+        if (withElement != null) return null;
+        final messageId = int.tryParse(operand, radix: 10);
+        if (messageId == null) return null;
+        withElement = ApiNarrowWith(messageId);
+
       case _NarrowOperator.is_:
         // It is fine to have duplicates of the same [IsOperand].
         isElementOperands.add(IsOperand.fromRawString(operand));
 
       case _NarrowOperator.near: // TODO(#82): support for near
-      case _NarrowOperator.with_: // TODO(#683): support for with
         continue;
 
       case _NarrowOperator.unknown:
@@ -202,7 +210,9 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
   }
 
   if (isElementOperands.isNotEmpty) {
-    if (streamElement != null || topicElement != null || dmElement != null) return null;
+    if (streamElement != null || topicElement != null || dmElement != null || withElement != null) {
+      return null;
+    }
     if (isElementOperands.length > 1) return null;
     switch (isElementOperands.single) {
       case IsOperand.mentioned:
@@ -219,13 +229,14 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
         return null;
     }
   } else if (dmElement != null) {
-    if (streamElement != null || topicElement != null) return null;
+    if (streamElement != null || topicElement != null || withElement != null) return null;
     return DmNarrow.withUsers(dmElement.operand, selfUserId: store.selfUserId);
   } else if (streamElement != null) {
     final streamId = streamElement.operand;
     if (topicElement != null) {
-      return TopicNarrow(streamId, topicElement.operand);
+      return TopicNarrow(streamId, topicElement.operand, with_: withElement?.operand);
     } else {
+      if (withElement != null) return null;
       return ChannelNarrow(streamId);
     }
   }
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -59,7 +59,7 @@ String? decodeHashComponent(String str) {
 // you do so by passing the `anchor` param.
 Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
   // TODO(server-7)
-  final apiNarrow = resolveDmElements(
+  final apiNarrow = resolveApiNarrowForServer(
     narrow.apiEncode(), store.connection.zulipFeatureLevel!);
   final fragment = StringBuffer('narrow');
   for (ApiNarrowElement element in apiNarrow) {
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -511,6 +511,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       numAfter: 0,
     );
     if (this.generation > generation) return;
+    _adjustNarrowForTopicPermalink(result.messages.firstOrNull);
     store.reconcileMessages(result.messages);
     store.recentSenders.handleMessages(result.messages); // TODO(#824)
     for (final message in result.messages) {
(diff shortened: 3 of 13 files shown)


lib/model/message_list.dart
```diff
@@ -493,11 +494,45 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+  /// because the message was moved.
+  ///
+  /// If such a "redirect" happened, this helper updates the stream and topic
+  /// in [narrow] to match the message's current conversation.
+  /// It also removes the "with" component from [narrow]
+  /// whether or not a redirect happened.
+  ///
+  /// See API doc:
+  ///   https://zulip.com/api/construct-narrow#message-ids
+  void _adjustTopicPermalinkNarrow(Narrow narrow_, Message? someFetchedMessageOrNull) {
```
PIG208, line 512: Looks like the caller passes the same `narrow` as the `narrow` that lives on the message list. Should we remove `narrow_` and just use `narrow` instead?

test/widgets/message_list_test.dart
```diff
@@ -51,13 +52,18 @@ void main() {
...
+    bool skipPumpAndSettle = false,
   }) async {
     addTearDown(testBinding.reset);
     streams ??= subscriptions ??= [eg.subscription(eg.stream(streamId: eg.defaultStreamMessageStreamId))];
     await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot(
       streams: streams, subscriptions: subscriptions, unreadMsgs: unreadMsgs));
     store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
     connection = store.connection as FakeApiConnection;
+    if (zulipFeatureLevel != null) {
+      connection.zulipFeatureLevel = zulipFeatureLevel;
```
PIG208, line 65: Hmm, I think `eg.selfAccount` carries a different `zulipFeatureLevel` if we only override the feature level on the `connection`.

PIG208 (commented):
Looks good overall! Left some comments.

lib/model/message_list.dart
chrisbobbe: I agree it's odd for this function to have a `narrow_` param; let's remove that.

It's convenient in the function's body to not keep using `narrow` each time we need it, because then we'd have to repeat things like `narrow as TopicNarrow` even in places where it's clear to us that the value can't be anything other than a `TopicNarrow`. I think what's happening there is: when we read `narrow`, the analyzer sees that as calling a getter, not accessing a variable, and it doesn't assume the getter will return the same value / `Narrow` subtype from one call to the next. And while in theory the analyzer could conclude that this class's `narrow` getter doesn't have that surprising behavior—it's just a field with the getter defined implicitly—it doesn't want to go and check for _subclasses_ that might override the getter with that surprising behavior.

How about a `narrow_` variable at the top of the function body, with a comment.

PIG208: Yeah, that explanation makes sense. I remember making the setter of `narrow` private might fix the issue, but I'm not sure. The proposed change sounds good to me too.

PIG208: Hmm nevermind, I think the local variable would be a better solution here.

See https://github.com/zulip/zulip-flutter/pull/808#discussion_r1677245004

chrisbobbe:

PIG208:
Thanks! This revision looks good to me.

lib/api/model/narrow.dart
```diff
@@ -4,18 +4,36 @@ part 'narrow.g.dart';
...
 
-/// Resolve any [ApiNarrowDm] elements appropriately.
+/// Resolve any [ApiNarrowElement]s, such as [ApiNarrowDm]s, appropriately.
 ///
 /// This encapsulates a server-feature check.
-ApiNarrow resolveDmElements(ApiNarrow narrow, int zulipFeatureLevel) {
-  if (!narrow.any((element) => element is ApiNarrowDm)) {
-    return narrow;
+// TODO(server-7) remove [ApiNarrowDm] reference in dartdoc
+ApiNarrow resolveApiNarrowElements(ApiNarrow narrow, int zulipFeatureLevel) {
```
gnprice, line 11: This name doesn't sound right to me for the expanded scope of this function. (In particular "elements" in the name feels redundant; it doesn't seem to convey any more information than `resolveApiNarrow` would.)

Here's a version:
```dart
/// Adapt the given narrow to be sent to the given Zulip server version.
///
/// Any elements that take a different name on old vs. new servers
/// will be resolved to the specific name to use.
/// Any elements that are unknown to old servers and can
/// reasonably be omitted will be omitted.
ApiNarrow resolveApiNarrowForServer(ApiNarrow narrow, int zulipFeatureLevel) {
```

gnprice (commented):
Thanks @chrisbobbe, and thanks @PIG208 for the previous reviews!

Generally this looks great. Comments below, most of them small.

test/widgets/message_list_test.dart
```diff
@@ -51,12 +52,17 @@ void main() {
     List<User>? users,
     List<Subscription>? subscriptions,
     UnreadMessagesSnapshot? unreadMsgs,
+    int? zulipFeatureLevel,
+    bool skipPumpAndSettle = false,
   }) async {
     addTearDown(testBinding.reset);
     streams ??= subscriptions ??= [eg.subscription(eg.stream(streamId: eg.defaultStreamMessageStreamId))];
-    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot(
+    final effectiveFeatureLevel = zulipFeatureLevel ?? eg.recentZulipFeatureLevel;
```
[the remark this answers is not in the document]
chrisbobbe: Interesting; I'll keep this pattern in mind. I'd always thought it was a code smell to reassign a function param.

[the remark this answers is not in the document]
gnprice: Yeah — just like mutating a local variable, it requires a bit of care to avoid making things confusing. But I think doing so right at the top of the function body, where it's effectively still parsing its arguments, is one pattern where it's generally pretty clean.

In this case it's also what this function is already doing with two other parameters on the previous line, so this might as well follow the same pattern.

chrisbobbe:
Thanks for the review and patience! Revision pushed. 🎉

lib/api/model/narrow.dart
```diff
@@ -6,18 +6,43 @@ part 'narrow.g.dart';
...
-    _             => element,
-  }).toList();
+
+  final result = <ApiNarrowElement>[];
+  for (final element in narrow) {
+    switch (element) {
+      case ApiNarrowDm():
+        result.add(element.resolve(legacy: !supportsOperatorDm));
+      case ApiNarrowWith():
+        assert(!supportsOperatorWith);
```
gnprice, line 38: What if there's both an ApiNarrowWith and an ApiNarrowDm? Then I think this assertion can fail.

lib/api/model/narrow.dart
```diff
@@ -6,18 +6,43 @@ part 'narrow.g.dart';
...
+  bool hasWithElement = false;
+  for (final element in narrow) {
+    switch (element) {
+      case ApiNarrowDm():   hasDmElement = true;
+      case ApiNarrowWith(): hasWithElement = true;
+      default:
+    }
+  }
+  if (!hasDmElement && (!hasWithElement || supportsOperatorWith)) {
     return narrow;
```
gnprice, line 29: Related to my previous comment just below: this condition ought to be equivalent to "the loop below will produce a result different from `narrow`" — that's the basis for the optimization this represents.

I think this condition is correct but the logic below doesn't match it.

lib/model/internal_link.dart
```diff
@@ -159,6 +161,7 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
 
   ApiNarrowStream? streamElement;
   ApiNarrowTopic? topicElement;
+  ApiNarrowWith? withElement;
   ApiNarrowDm? dmElement;
```
gnprice, line 165: nit: put these, and the cases below, in the same order as each other and as the respective ApiNarrowElement subclasses

lib/model/internal_link.dart
```diff
@@ -193,16 +196,21 @@ Narrow? _interpretNarrowSegments(List<String> segments, PerAccountStore store) {
         isElementOperands.add(IsOperand.fromRawString(operand));
 
       case _NarrowOperator.near: // TODO(#82): support for near
-      case _NarrowOperator.with_: // TODO(#683): support for with
         continue;
 
+      case _NarrowOperator.with_:
+        if (withElement != null) return null;
+        withElement = ApiNarrowWith(int.parse(operand, radix: 10));
```
gnprice, line 203: on parse failure should return null, not throw

lib/model/narrow.dart
```diff
@@ -108,22 +111,33 @@ class TopicNarrow extends Narrow implements SendableNarrow {
...
   @override
   StreamDestination get destination => StreamDestination(streamId, topic);
 
   @override
-  String toString() => 'TopicNarrow($streamId, ${topic.displayName})';
+  String toString() {
+    final fields = [
+      streamId.toString(),
+      topic.displayName,
+      if (with_ != null) with_!,
```
gnprice, line 128: should include "with: " for clarity

lib/model/message_list.dart
```diff
@@ -524,12 +525,47 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+    if (narrow is! TopicNarrow || narrow.with_ == null) return;
+
+    switch (someFetchedMessageOrNull) {
+      case null:
+        // This can't be a redirect; a redirect can't produce an empty result.
+        // (The server only redirects if the message is accessible to the user,
+        // and if it is, it'll appear in the result, making it non-empty.)
+        this.narrow = narrow.sansWith();
+      case StreamMessage():
+        this.narrow = TopicNarrow.ofMessage(someFetchedMessageOrNull);
```
gnprice, line 554: Over on the widget state there's a comment which should get updated:
```dart
  void _modelChanged() {
    if (model!.narrow != widget.narrow) {
      // A message move event occurred, where propagate mode is
      // [PropagateMode.changeAll] or [PropagateMode.changeLater].
      widget.onNarrowChanged(model!.narrow);

File: test/model/message_list_test.dart [hunk 11]
```diff
@@ -95,222 +95,286 @@ void main() {
...
-      ..messages.length.equals(kMessageListFetchBatchSize)
-      ..haveOldest.isFalse();
-    checkLastRequest(
-      narrow: narrow.apiEncode(),
-      anchor: 'newest',
-      numBefore: kMessageListFetchBatchSize,
-      numAfter: 0,
-    );
-  });
+    final otherStream = eg.stream();
```
gnprice, line 102: nit: otherChannel, to match the neighboring someChannel

test/model/message_list_test.dart
```diff
@@ -95,222 +95,286 @@ void main() {
...
+      });
 
-    connection.prepare(json: olderResult(
-      anchor: 1000, foundOldest: false,
-      messages: List.generate(100, (i) => eg.streamMessage(id: 900 + i)),
-    ).toJson());
-    final fetchFuture = model.fetchOlder();
-    checkNotifiedOnce();
-    check(model).fetchingOlder.isTrue();
+      test('topic permalink, message was moved', () async {
```
gnprice, line 149: I think this test case and the one above it can probably be left out — it doesn't seem like they add anything that isn't covered either by the neighboring tests, or the two test cases added below.

test/widgets/message_list_test.dart
```diff
@@ -222,6 +238,87 @@ void main() {
...
+        // We only made one fetch.
+        check(connection.takeRequests()).single.isA<http.Request>()
+          ..method.equals('GET')
+          ..url.path.equals('/api/v1/messages')
+          ..url.queryParameters.deepEquals({
+            'narrow': jsonEncode([
+              ApiNarrowStream(someStream.streamId),
+              ApiNarrowTopic(eg.t(someTopic)),
+              ApiNarrowWith(1),
+            ]),
```
gnprice, line 313: This can use `narrow.apiEncode()` like the other test case does, right? Or is there a difference I'm missing?

test/widgets/message_list_test.dart
```diff
@@ -222,6 +238,87 @@ void main() {
...
+      const otherTopic = 'other topic';
+
+      testWidgets('with message move', (tester) async {
+        final narrow = TopicNarrow(someStream.streamId, eg.t(someTopic), with_: 1);
+        await setupMessageListPage(tester,
+          narrow: narrow,
+          // server sends the /with/<id> message in its current, different location
+          messages: [eg.streamMessage(id: 1, stream: otherStream, topic: otherTopic)],
+          streams: [someStream, otherStream],
+          subscriptions: [eg.subscription(someStream), eg.subscription(otherStream)],
```
gnprice, line 256: nit: I think the subscriptions aren't needed here

(some other tests need them because they use CombinedFeedNarrow)

gnprice (commented):
Thanks for the revision! I think this is close; comments below.

lib/api/model/narrow.dart
chrisbobbe: Ah right—thanks for the catch.

lib/model/message_list.dart
chrisbobbe: Ooh thanks for the catch.

test/widgets/message_list_test.dart
chrisbobbe: Oh right! That makes sense.

chrisbobbe:
Thanks for the review! Revision pushed, this time with tests for `resolveApiNarrowForServer`, since some bugs snuck in (thank you for catching them in https://github.com/zulip/zulip-flutter/pull/1029#discussion_r1956920148 !).

test/api/model/narrow_test.dart
```diff
@@ -0,0 +1,69 @@
...
+      // Unlikely to occur in the wild but should still be handled correctly
+      doTest('dm + with', [dm, with_], [dm.resolve(legacy: false), with_]);
+    });
+
+    // TODO(server-7)
+    group('FL 176', () {
+      final zulipFeatureLevel = 176;
+      doTest('dm (legacy)', zulipFeatureLevel: zulipFeatureLevel,
+        [dm], [dm.resolve(legacy: true)]);
+      doTest('topic, not permalink', [stream, topic], [stream, topic]);
```
gnprice, line 46: Is this meant to pass zulipFeatureLevel too?

lib/api/model/narrow.dart
```diff
@@ -150,6 +173,22 @@ class ApiNarrowPmWith extends ApiNarrowDm {
   ApiNarrowPmWith._(super.operand, {super.negated});
 }
 
+/// An [ApiNarrowElement] with the 'with' operator.
+///
+/// If part of [ApiNarrow] use [resolveApiNarrowElements].
```
gnprice, line 178: nit: update name for revision

lib/model/internal_link.dart
```diff
@@ -78,6 +78,8 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
         fragment.write('$streamId-$slugifiedName');
       case ApiNarrowTopic():
         fragment.write(_encodeHashComponent(element.operand.apiName));
+      case ApiNarrowWith():
```
gnprice, line 81: nit: match order of cases

test/api/model/narrow_test.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:zulip/api/model/model.dart';
+import 'package:zulip/api/model/narrow.dart';
+
+import '../../example_data.dart' as eg;
+import '../../stdlib_checks.dart';
+import 'events_checks.dart';
+import 'model_checks.dart';
+
+void main() {
+  group('resolveApiNarrowForServer', () {
```
gnprice, line 14: This should get a comment pointing to the existing set of tests that cover this method, which are in `test/api/route/messages_test.dart`, particularly the `Narrow.toJson` test.

gnprice (commented):
Thanks! Looks good; just a few comments.

gnprice:
(Oh and there are some unused imports the analyzer is complaining about.)

test/api/model/narrow_test.dart
chrisbobbe: Oops, yes.

test/api/model/narrow_test.dart
chrisbobbe: Ah indeed! Thanks for pointing out that existing coverage—which I see I even added to, since an earlier revision, but forgot about that because I wrote that revision a while ago. 🙂

chrisbobbe:

test/api/route/messages_test.dart
```diff
@@ -184,21 +191,51 @@ void main() {
...
       checkNarrow([ApiNarrowDm([123, 234])], jsonEncode([
         {'operator': 'pm-with', 'operand': [123, 234]},
       ]));
+
+      // Unlikely to occur in the wild but should still be handled correctly
+      checkNarrow([ApiNarrowDm([123, 234]), ApiNarrowWith(1)], jsonEncode([
+        {'operator': 'pm-with', 'operand': [123, 234]},
+      ]));
+
+      connection.zulipFeatureLevel = 270;
```
gnprice, line 224: nit: put this section above the FL 170 section — that way there's more of a monotonic range here from newest to oldish to older

test/api/route/messages_test.dart
```diff
@@ -184,21 +191,51 @@ void main() {
...
+      checkNarrow(eg.topicNarrow(12, 'stuff', with_: 1).apiEncode(), jsonEncode([
+        {'operator': 'stream', 'operand': 12},
+        {'operator': 'topic', 'operand': 'stuff'},
+      ]));
       checkNarrow([ApiNarrowDm([123, 234])], jsonEncode([
         {'operator': 'pm-with', 'operand': [123, 234]},
       ]));
+
+      // Unlikely to occur in the wild but should still be handled correctly
+      checkNarrow([ApiNarrowDm([123, 234]), ApiNarrowWith(1)], jsonEncode([
```
gnprice, line 220:
```suggestion
      ]));
      checkNarrow([ApiNarrowDm([123, 234]), ApiNarrowWith(1)], jsonEncode([
```

I think the apology isn't needed :slightly_smiling_face:

test/api/route/messages_test.dart
```diff
@@ -184,21 +191,51 @@ void main() {
...
-      checkNarrow([ApiNarrowDm([123, 234])], jsonEncode([
+      // Unlikely to occur in the wild but should still be handled correctly
+      checkNarrow([ApiNarrowDm([123, 234]), ApiNarrowWith(1)], jsonEncode([
         {'operator': 'dm', 'operand': [123, 234]},
+        {'operator': 'with', 'operand': 1},
       ]));
 
       connection.zulipFeatureLevel = 176;
+
+      checkNarrow(eg.topicNarrow(12, 'stuff', with_: 1).apiEncode(), jsonEncode([
```
gnprice, line 211: nit: and then without those other comments, the change to the ZFL can be comfortably kept in the same stanza with the checks that rely on it
```suggestion
      ]));

      connection.zulipFeatureLevel = 176;
      checkNarrow(eg.topicNarrow(12, 'stuff', with_: 1).apiEncode(), jsonEncode([
```

I think that substantially helps to mitigate the risk of confusion from this use of mutable state, where one otherwise might see one of those later `checkNarrow` calls in isolation and wonder why the expected behavior is what it is

lib/api/model/narrow.dart
```diff
@@ -6,18 +6,41 @@ part 'narrow.g.dart';
...
+  bool hasDmElement = false;
+  bool hasWithElement = false;
+  for (final element in narrow) {
+    switch (element) {
+      case ApiNarrowDm():   hasDmElement = true;
+      case ApiNarrowWith(): hasWithElement = true;
+      default:
+    }
+  }
+  if (!hasDmElement && (!hasWithElement || supportsOperatorWith)) {
```
gnprice, line 28: nit: use a De Morgan transform of this:
```suggestion
  if (!hasDmElement && !(hasWithElement && !supportsOperatorWith)) {
```

That way it more directly corresponds to the condition that's written in the switch cases below.

(You could even then apply De Morgan to the outer operator too — write "not (A or B)" instead of "not-A and not-B" — and that might be clearer still.)

gnprice (commented):
Thanks for the revision! Looks good; a few small comments.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merging.

PR #1030, opened 2024-10-27
test: Bump eg.recentZulipFeatureLevel and recentZulipVersion to 9.0

gnprice:
This keeps eg.account and eg.initialSnapshot using recent versions by default.


chrisbobbe:
LGTM, thanks! Merged.

Issue #360, opened 2023-11-01
Handle <table> elements


gnprice:
We should handle HTML tables in message content, i.e. this Zulip feature: https://zulip.com/help/tables

Handling these means parsing them in `lib/model/content.dart`, and then displaying them appropriately in `lib/widgets/content.dart`. For appropriate styling, see the Zulip web app's `web/styles/rendered_markdown.css`.

PR #1031, opened 2024-10-28
content: Handle `<table>` elements

rajveermalviya:
Fixes: #360 

| Flutter | Web |
| - | - |
| Screenshot 2024-10-29 at 01 50 02 | Screenshot 2024-10-29 at 01 37 36 |

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -543,12 +543,42 @@ class TableRowNode extends BlockContentNode {
   }
 }
 
+// The text-alignment setting that applies to a cell's column, from the delimiter row.
+//
+// See GitHub-flavored Markdown:
+//   https://github.github.com/gfm/#tables-extension-
+enum TableColumnTextAlignment {
+  /// All cells' text left-aligned, represented in Markdown as `|: --- |`.
+  left, // TODO(i18n) RTL issues? https://github.com/zulip/zulip/issues/32265
+  /// All cells' text center-aligned, represented in Markdown as `|: --- :|`.
+  center,
+  /// All cells' text right-aligned, represented in Markdown as `| --- :|`.
+  right, // TODO(i18n) RTL issues? https://github.com/zulip/zulip/issues/32265
+  /// Cells' text aligned the default way, represented in Markdown as `| --- |`.
+  defaults
+}
+
 class TableCellNode extends BlockInlineContainerNode {
   const TableCellNode({
     super.debugHtmlNode,
     required super.nodes,
     required super.links,
+    required this.textAlignment,
   });
+
+  /// The table column text-alignment to be used for this cell.
+  // In Markdown, alignment is defined per column using the delimiter row.
+  // However, the generated HTML specifies alignment for each cell in a row
+  // individually, that matches the UI widget implementation which is also
+  // row based and needs alignment information to be per cell.
+  final TableColumnTextAlignment textAlignment;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(EnumProperty('textAlignment', textAlignment,
+      defaultValue: TableColumnTextAlignment.defaults));
+  }
 }
 
 /// A content node that expects an inline layout context from its parent.
@@ -1278,10 +1308,25 @@ class _ZulipContentParser {
       assert(node.localName == (isHeader ? 'th' : 'td'));
       assert(node.className.isEmpty);
 
+      final cellStyle = node.attributes['style'];
+      final TableColumnTextAlignment? textAlignment;
+      switch (cellStyle) {
+        case null:
+          textAlignment = TableColumnTextAlignment.defaults;
+        case 'text-align: left;':
+          textAlignment = TableColumnTextAlignment.left;
+        case 'text-align: center;':
+          textAlignment = TableColumnTextAlignment.center;
+        case 'text-align: right;':
+          textAlignment = TableColumnTextAlignment.right;
+        default:
+          return null;
+      }
       final parsed = parseBlockInline(node.nodes);
       return TableCellNode(
         nodes: parsed.nodes,
-        links: parsed.links);
+        links: parsed.links,
+        textAlignment: textAlignment);
     }
 
     List<TableCellNode>? parseTableCells(dom.NodeList cellNodes, bool isHeader) {
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -504,6 +504,53 @@ class EmbedVideoNode extends BlockContentNode {
   }
 }
 
+class TableNode extends BlockContentNode {
+  const TableNode({super.debugHtmlNode, required this.rows});
+
+  final List<TableRowNode> rows;
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return rows
+      .mapIndexed((i, row) => row.toDiagnosticsNode(name: 'row $i'))
+      .toList();
+  }
+}
+
+class TableRowNode extends BlockContentNode {
+  const TableRowNode({
+    super.debugHtmlNode,
+    required this.cells,
+    required this.isHeader,
+  });
+
+  final List<TableCellNode> cells;
+
+  /// Indicates whether this row is the header row.
+  final bool isHeader;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(FlagProperty('isHeader', value: isHeader, ifTrue: "is header"));
+  }
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return cells
+      .mapIndexed((i, cell) => cell.toDiagnosticsNode(name: 'cell $i'))
+      .toList();
+  }
+}
+
+class TableCellNode extends BlockInlineContainerNode {
+  const TableCellNode({
+    super.debugHtmlNode,
+    required super.nodes,
+    required super.links,
+  });
+}
+
 /// A content node that expects an inline layout context from its parent.
 ///
 /// When rendered into a Flutter widget tree, an inline content node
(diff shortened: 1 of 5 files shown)


lib/widgets/content.dart
```diff
@@ -1196,6 +1203,61 @@ class GlobalTime extends StatelessWidget {
...
+        ? Padding(
+            padding: const EdgeInsets.all(5),
+            child: _buildBlockInlineContainer(
+              node: node,
+              style: isHeading
+                ? DefaultTextStyle.of(context).style
+                    .merge(weightVariableTextStyle(context, wght: 700))
+                : DefaultTextStyle.of(context).style,
+              textAlign: node.textAlign))
+        : const SizedBox.shrink());
```
chrisbobbe, line 1257: Should we include the `Padding` in this case too? This looks like it could make a cell that doesn't take up any space at all, and I don't think that's consistent with the web app. Here's what the web app does with a table that's nothing but empty cells:

https://chat.zulip.org/#narrow/channel/7-test-here/topic/Chris/near/1971419
image

and each one does take up some space, and from the inspector it looks like that's coming from some padding (though 4px in all directions instead of 5px).

chrisbobbe (commented):
Great to see this! I might be jumping on this draft PR too early, but I had a few thoughts from a quick skim 🙂. One comment below, and:

In the 'strong (bold)' tests, let's test that a `**bold**` span in a header cell gets a font weight of 900, even when that span is nested inside another span. (900 is 700 from the header-cell styling plus 200 from the bold-span code.) So, another pair of tests like these:

<details>

```dart
    testFontWeight('in spoiler header',
      expectedWght: 900,
      // ```spoiler regular **bold**
      // content
      // ```
      content: plainContent(
        '<div class="spoiler-block"><div class="spoiler-header">\n'
          '<p>regular <strong>bold</strong></p>\n'
          '</div><div class="spoiler-content" aria-hidden="true">\n'
          '<p>content</p>\n'
          '</div></div>'
      ),
      styleFinder: findWordBold,
    );

    testFontWeight('in different kind of span in spoiler header',
      expectedWght: 900,
      // ```spoiler *italic **bold***
      // content
      // ```
      content: plainContent(
        '<div class="spoiler-block"><div class="spoiler-header">\n'
          '<p><em>italic <strong>bold</strong></em></p>\n'
          '</div><div class="spoiler-content" aria-hidden="true">\n'
          '<p>content</p>\n'
          '</div></div>'
      ),
      styleFinder: findWordBold,
    );
```

</details>

rajveermalviya:
Thanks for the initial comments @chrisbobbe! This is now ready for review, PTAL.

lib/widgets/content.dart
```diff
@@ -1196,6 +1235,58 @@ class GlobalTime extends StatelessWidget {
...
+
+  final TableCellNode node;
+  final bool isHeader;
+
+  @override
+  Widget build(BuildContext context) {
+    return TableCell(
+      verticalAlignment: TableCellVerticalAlignment.middle,
+      child: Padding(
+        padding: const EdgeInsets.all(4),
```
chrisbobbe, line 1276:
```suggestion
        // Web has 4px padding and 1px border on all sides.
        // In web, the 1px border grows each cell by 0.5px in all directions.
        // Our border doesn't affect the layout, it's just painted on,
        // so we add 0.5px on all sides to match web.
        padding: const EdgeInsets.all(4 + 0.5),
```

chrisbobbe (commented):
Thanks! Comments below.

I wonder if `tools/content/check-features` shows any surprises after implementing table support; do you have time to run that?

lib/widgets/content.dart
rajveermalviya: Interesting, found an upstream issue for this: https://github.com/flutter/flutter/issues/78691.
Included the link for reference.

rajveermalviya:
Thanks for the review @chrisbobbe! Pushed a new revision.
I also ran the `tools/content/check-features` script and found that with this PR there are zero failures for the `<table>` content 🎉. This is hopefully accurate, as the current parser impl is overly strict about missing or extra classes, attributes and nodes.

lib/model/content.dart
```diff
@@ -504,6 +504,88 @@ class EmbedVideoNode extends BlockContentNode {
...
+  // Represented in markdown as `|: --- :|` for the specific column in the delimeter row.
+  center,
+  /// Text inside the table cell is right-aligned.
+  // Represented in markdown as `| --- :|` for the specific column in the delimeter row.
+  // TODO(i18n) RTL issues? https://github.com/zulip/zulip/issues/32265
+  right,
+  /// Text inside the table cell is aligned the default way.
+  // Represented in markdown as `| --- |` for the specific column in the delimeter row.
+  defaults
+}
```
chrisbobbe, line 569: In how we name and document this enum, I'd like to be more explicit that what's being represented is a per-column setting, in the string from the server and in our model. That's how users experience the feature: we tell them they can set the alignment per-column, but not per individual cell (or for a row, or for the whole table, etc.).

It's true that the enum value is repeated on all of a column's cells, and our parsing logic has to understand that fact in order to recognize what the setting is for a given column. I also think it's fine for our model to repeat something on all of a column's cells too, because that's convenient for our parsing code and our UI code. But when doing those things, I'd still like the per-column abstraction to be clear somewhere, and I think the best place for that is in this enum's name and documentation. As in my proposal in https://github.com/zulip/zulip-flutter/pull/1031#discussion_r1831950371 :

```dart
// The text-alignment setting that applies to a cell's column, from the delimiter row.
//
// See GitHub-flavored Markdown:
//   https://github.github.com/gfm/#tables-extension-
enum TableColumnTextAlignment {
  /// All cells' text left-aligned, represented in Markdown as `|: --- |`.
  left, // TODO(i18n) RTL issues? https://github.com/zulip/zulip/issues/32265
  /// All cells' text center-aligned, represented in Markdown as `|: --- :|`.
  center,
  /// All cells' text right-aligned, represented in Markdown as `| --- :|`.
  right, // TODO(i18n) RTL issues? https://github.com/zulip/zulip/issues/32265
  /// Cells' text aligned the default way, represented in Markdown as `| --- |`.
  defaults,
}
```

In theory, our parsing code could grow a validation step to make sure the setting is expressed coherently, i.e., scan through a column's cells to check that their `style` attribute is consistent. I don't think we have to do that (could be expensive), but I do think it would be consistent with what's supposed to be represented.

I think the explicit per-column abstraction will be helpful when we come back to address the RTL issues. As part of that work, we'll want to get clear ideas of what we expect to happen vs. what's actually happening, at all of the steps from the compose-box input to how the message appears in the message list.

chrisbobbe (commented):
Thanks! Small comments below.

lib/model/content.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:ui';
```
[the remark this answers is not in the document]
gnprice: Yeah, that adjustment SGTM.

We already import `package:flutter/foundation.dart` in a number of places in our model code, I think mostly for ChangeNotifier, and that transitively pulls in `dart:ui`. So if something would be messy to do without `dart:ui`, it'd be fine to import it.

But referring to something UI-specific like `TextAlign` is definitely a good prompt to think about if the logic would best live in our widgets code. In this example, I think this resulting change in the later revision, where this bit of logic moved from the parser to the widget:
```diff
       switch (cellStyle) {
         case null:
-          // Default text alignment specified in for `tr > th` element
-          // in `web/styles/rendered_markdown.css`:
-          //  https://github.com/zulip/zulip/blob/d556c0e0a5e9e8ba0ff548354bf2ba6ef2c97d4b/web/styles/rendered_markdown.css#L140
-          textAlign = isHeader ? TextAlign.left : null;
+          textAlignment = TableCellTextAlignment.defaults;
```

```diff
+      // The web client sets `text-align: left;` for the header cells,
+      // overriding the default browser alignment (which is `center` for header
+      // and `start` for body). By default, the [Table] widget uses `start` for
+      // text alignment, a saner choice that supports RTL text. So, defer to that.
+      // See discussion:
+      //  https://github.com/zulip/zulip-flutter/pull/1031#discussion_r1831950371
+      TableCellTextAlignment.defaults => null,
```

was a improvement in clarity.

rajveermalviya:
Thanks for the review @chrisbobbe! Pushed a new revision, PTAL.

chrisbobbe:

gnprice (commented):

Generally this looks great. Various comments below, mainly small.

rajveermalviya:

lib/model/content.dart
```diff
@@ -504,6 +504,84 @@ class EmbedVideoNode extends BlockContentNode {
...
+  // However, the generated HTML specifies alignment for each cell in a row
+  // individually, that matches the UI widget implementation which is also
+  // row based and needs alignment information to be per cell.
+  final TableColumnTextAlignment textAlignment;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(EnumProperty<TableColumnTextAlignment>(
+      'textAlignment', textAlignment,
```
gnprice, line 580: nit: type can be inferred
```suggestion
    properties.add(EnumProperty('textAlignment', textAlignment,
```

lib/model/content.dart
```diff
@@ -1222,6 +1298,136 @@ class _ZulipContentParser {
...
+
+    bool parseTableBodyRows(dom.NodeList tbodyNodes, int headerColumnCount, List<TableRowNode> parsedRows) {
+      for (final node in tbodyNodes) {
+        if (node is dom.Text && node.text == '\n') continue;
+
+        if (node is! dom.Element) return false;
+        if (node.localName != 'tr') return false;
+        if (node.className.isNotEmpty) return false;
+        if (node.attributes.isNotEmpty) return false;
+        if (node.nodes.isEmpty) return false;
```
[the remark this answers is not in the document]
gnprice: > Hmm, looks like for this case the generated HTML adds the empty `<td>` node:

I see — and a `<tr>` around that. I guess that also suffices for causing this code (with the rejection on `node.nodes.isEmpty`) to work correctly.

I was looking at the GFM spec:
https://github.github.com/gfm/#example-205
but hadn't checked what Zulip actually produces. Good to have checked!


> Something similar happens when markdown misses a column in the body row,

Yeah. That part matches the spec:
https://github.github.com/gfm/#example-204

lib/widgets/content.dart
```diff
@@ -866,7 +901,7 @@ class _BlockInlineContainerState extends State<_BlockInlineContainer> {
   @override
   Widget build(BuildContext context) {
     return InlineContent(recognizer: null, linkRecognizers: _recognizers,
-      style: widget.style, nodes: widget.nodes);
+      style: widget.style, nodes: widget.nodes, textAlign: widget.textAlign);
```
gnprice, line 904: Cool, I think this fixes https://github.com/zulip/zulip-flutter/pull/1031#discussion_r1855858138 .

Let's have a test case that would have caught that bug. Doesn't need to exercise all the possible alignments (that's covered by another test case you already have), just one of them.

gnprice (commented):
Thanks for the revision! Comments below — one more test case and otherwise all nits.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

PR #1033, opened 2024-10-31
Support setting typing status in more situations

PIG208:
A follow-up to #897. This will cover more types of interactions described [here](https://github.com/zulip/zulip-flutter/pull/897#discussion_r1818182796).

Specifically, this will track interactions with compose buttons and autocomplete scroll view, and send typing notices.

PR #1034, opened 2024-11-01
compose_box test: Use a layout similar to the message list page

PIG208:
This gives more vertical space above the compose box.  This will provide enough space for the autocomplete to appear without overflowing when we add tests for it.

This should be useful for both #928 and #1033. It doesn't aim to reproduce the exact things we do with `MessageListPage`.

| before | after |
| - | - |
| image | image |


chrisbobbe:
LGTM, thanks! (Sorry for the delay; I meant to post LGTM earlier but I forgot.) Marking for Greg's review.

gnprice:
Thanks! LGTM; merging.

Issue #348, opened 2023-10-31
UI for topic muting/unmuting/following


gnprice:
This will be a followup from #346, and a complement to #347.

In zulip-mobile, we offer this only for muting and unmuting — we haven't implemented the new "follow topic" feature. At a minimum, for launch, we should offer the same level of functionality as zulip-mobile; but then probably one of the first features we'd want to add after launch would be full "follow topic" support, so we might build the full version in the first place.

In zulip-mobile, we offer this in one place, the "topic action sheet", which you get by long-pressing a topic:
* as the app bar of a topic narrow;
* as the recipient header of a stream message in the message list;
* in the screen with the list of topics for a stream;
* in the inbox view.

We should do something broadly comparable, unless we do leap ahead to include "follow topic" support. We might in any case vary the details; for example if you long-press the stream name within a recipient header, that should perhaps give the stream action sheet instead of the topic action sheet.

PIG208:
[Topic following in zulip-mobile](https://github.com/zulip/zulip-mobile/issues/5771) was added on Nov 22, 2023. To offer the same level of functionality, we should probably implement following as well.

PIG208:
See [discussion](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Adding.20user.20topic.20visibility.20markers.20.28muted.2C.20default.2C.20etc.2E.29/near/1969293) here on the design of this feature.

PR #1041, opened 2024-11-01
Support topic muting/unmuting/following

PIG208:
<details>

| channel | topic | screenshot |
| - | - | - |
| unmuted | inherit |   |
| unmuted | muted |  |
| muted | inherit |  |
| muted | followed |  |
| muted | unmuted |  |

</details>

Reference zulip-mobile implementation: https://github.com/zulip/zulip-mobile/blob/715d60a5e87fe37032bce58bd72edb99208e15be/src/action-sheets/index.js#L656-L753

Fixes: #348


lib/widgets/action_sheet.dart
```diff
@@ -22,6 +23,297 @@ import 'store.dart';
...
+  final int streamId;
+  final String topic;
+  final BuildContext topicParentContext;
+
+  UserTopicVisibilityPolicy get visibilityPolicy;
+
+  @override void onPressed(BuildContext context) async {
+    Navigator.of(context).pop();
+    try {
+      await updateUserTopic(
```
gnprice, line 199: This seems more specific than the name `TopicActionSheetButton` reflects. We'll have other options in the topic action sheet that aren't about visibility policy, like resolve/unresolve.

It makes sense to group all these options under a common base class; it should just get a more specific name.

lib/widgets/action_sheet.dart
```diff
@@ -22,6 +23,297 @@ import 'store.dart';
...
+  const TopicActionSheetButton({
+    super.key,
+    required this.streamId,
+    required this.topic,
+    required this.topicParentContext,
+  });
+
+  final int streamId;
+  final String topic;
+  final BuildContext topicParentContext;
```
gnprice, line 192: This name doesn't feel right conceptually. I'll see about sending a quick PR refactoring the existing message-action-sheet options that might help clarify what this field is about.

gnprice (commented):
Thanks! Quick initial comments from skimming this draft.

lib/widgets/action_sheet.dart
gnprice: Sent #1044.

chrisbobbe (commented):
Thanks! Comments below, and I agree with your comment at https://github.com/zulip/zulip-flutter/issues/348#issuecomment-2435903346 🙂

> I think a design requirement of this feature is to also display mute/following states for each topic. At this point we can borrow that from the legacy mobile app.

lib/widgets/action_sheet.dart
```diff
@@ -141,6 +142,199 @@ class ActionSheetCancelButton extends StatelessWidget {
...
+            optionButtons.add(follow);
+          }
+        case UserTopicVisibilityPolicy.followed:
+          optionButtons.add(mute);
+          if (supportsFollowingTopics) {
+            optionButtons.add(unfollow);
+          }
+        case UserTopicVisibilityPolicy.unknown:
+          assert(false);
+          return;
```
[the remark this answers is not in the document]
gnprice: Yeah. By having this `unknown` value in the enum we're already paying most of the cost of gracefully handling the situation where the server introduces a new policy value, so we should handle it here too.

test/widgets/action_sheet_test.dart
```diff
@@ -93,6 +97,286 @@ void main() {
...
+      });
+    }
+
+    final mute =                 findButton(to:   UserTopicVisibilityPolicy.muted);
+    final unmute =               findButton(from: UserTopicVisibilityPolicy.muted,
+                                            to:   UserTopicVisibilityPolicy.none);
+    final unmuteInMutedChannel = findButton(to:   UserTopicVisibilityPolicy.unmuted);
+    final follow =               findButton(to:   UserTopicVisibilityPolicy.followed);
+    final unfollow =             findButton(from: UserTopicVisibilityPolicy.followed,
+                                            to:   UserTopicVisibilityPolicy.none);
```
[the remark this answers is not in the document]
PIG208: This reasoning makes sense. A remaining issue is that `unmuteInMutedChannel` and `unmute` have the same text label, but behave differently. So I guess in addition to verifying the text, we also want to confirm that the behavior matches what we expect.

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
PIG208: We do not expect it here because we already explicitly keep it out of our data structure, so it would be a bug to get `unknown` here.

https://github.com/zulip/zulip-flutter/blob/9e42f2694afffd3a40c7ca4786e8641622513350/lib/model/channel.dart#L216-L222

lib/api/route/channels.dart
```diff
@@ -38,3 +39,33 @@ class GetStreamTopicsEntry {
...
+  if (connection.zulipFeatureLevel! < 170) {
+    // TODO(server-7): deprecated in FL 170
+    assert(visibilityPolicy == UserTopicVisibilityPolicy.none
+           || visibilityPolicy == UserTopicVisibilityPolicy.muted);
+    final op = visibilityPolicy == UserTopicVisibilityPolicy.none ? 'remove'
+                                                                  : 'add';
+    // There can be an error when muting a topic that is already muted
+    // or unmuting one that is already unmuted.  Let it throw because we can't
+    // reliably filter out the error, which doesn't have a specific "code".
+    return connection.post('muteTopic', (_) {}, 'users/me/subscriptions/muted_topics', {
```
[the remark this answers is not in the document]
PIG208: Oh good catch! Looks like we don't even have the `connection.patch` method; will add it in a prep commit.

test/widgets/action_sheet_test.dart
```diff
@@ -93,6 +97,286 @@ void main() {
...
+
+    Future<void> setupToTopicActionSheet(WidgetTester tester, {
+      required bool? isChannelMuted,
```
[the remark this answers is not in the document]
PIG208: > Let's be more explicit about what null means for isChannelMuted. I think it's supposed to mean that the self-user isn't subscribed to the channel (right?), but that isn't clear from test-failure output—

Yeah. Updated the test description and leave the topic names as-is, because the description is what shows up when a test fails.

PIG208:

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
gnprice: Hmm, indeed. In that case it doesn't matter what this line does — though it's still probably cleaner to have it skip this type of button rather than abort the whole function. And it should have a comment saying why it's impossible.

I've noticed in a few other places that it'd be good to have a type for "visibility policy, but only the known/valid ones". Probably that points to removing `unknown` from the enum, and using a nullable type where we actually want that value. Out of scope for this PR, though.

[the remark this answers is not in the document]
PIG208: Yeah, this will allow us to remove the cases [here](https://github.com/search?q=repo%3Azulip%2Fzulip-flutter+case+UserTopicVisibilityPolicy.unknown%3A+assert%28false%29&type=code) (where this new `assert(false)` originally comes from). It is good to have the type confirm that we process the unknown values at the edge.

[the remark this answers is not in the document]
PIG208: Opened #1074 and include TODO comments linked in it.

lib/widgets/action_sheet.dart
```diff
@@ -181,6 +142,264 @@ class MessageActionSheetCancelButton extends StatelessWidget {
...
+      pageContext: context);
+  }
+
+  final mute =                 button(to:   UserTopicVisibilityPolicy.muted);
+  final unmute =               button(from: UserTopicVisibilityPolicy.muted,
+                                      to:   UserTopicVisibilityPolicy.none);
+  final unmuteInMutedChannel = button(to:   UserTopicVisibilityPolicy.unmuted);
+  final follow =               button(to:   UserTopicVisibilityPolicy.followed);
+  final unfollow =             button(from: UserTopicVisibilityPolicy.followed,
+                                      to:   UserTopicVisibilityPolicy.none);
```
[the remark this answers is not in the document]
chrisbobbe: Web's radio-buttons approach seems like a fine solution to me, but again we might postpone that; we should make an issue if it's planned though. @gnprice, what do you think?

assets/icons/ZulipIcons.ttf
chrisbobbe, line 1:
```
icons: Take mute/unmute/following icons from the web app
```

Hmm yeah, looks like the Figma doesn't have icons for these yet. Following web seems fine, but let's note in the commit message that we checked https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=544-22131&node-type=canvas&m=dev and didn't find corresponding icons there.

assets/l10n/app_en.arb
```diff
@@ -59,6 +59,22 @@
   "@permissionsDeniedReadExternalStorage": {
     "description": "Message for dialog asking the user to grant permissions for external storage read access."
   },
+  "actionSheetOptionMuteTopic": "Mute topic",
```
chrisbobbe, line 62:
```
action_sheet: Support muting/unmuting/following topics
```

Bump on https://github.com/zulip/zulip-flutter/pull/1041#pullrequestreview-2453170881 :

> Thanks! Comments below, and I agree with your comment at [#348 (comment)](https://github.com/zulip/zulip-flutter/issues/348#issuecomment-2435903346) 🙂
> 
> > I think a design requirement of this feature is to also display mute/following states for each topic. At this point we can borrow that from the legacy mobile app.

The `Fixes:` line doesn't belong on this commit until we do that 🙂 or make it not part of #348.

chrisbobbe (commented):
Thanks! Comments below, including some things I missed last time (oops).

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
PIG208: Yeah, this PR follows the state transitions in the mobile app very closely, including the UX. If we are going with a newer radio-button design, it will probably save some time to get this right with a Figma redesign.

See [discussion here](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Adding.20user.20topic.20visibility.20markers.20.28muted.2C.20default.2C.20etc.2E.29/near/1969293).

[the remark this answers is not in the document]
chrisbobbe: Oh I see, that discussion is helpful context! I missed it in my review because I didn't see it linked from this PR or the issue. 🙂 I agree with the outcome there:

> Yeah, some version of that would probably be good to put in mobile's action sheet too.
> 
> Definitely post-launch, though.

Would you make an issue for that? It can link to this discussion, to memoize the work of figuring out what state transitions we're missing. (I used a lot of post-it notes, haha.)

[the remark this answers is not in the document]
PIG208: Sure! Opened #1078.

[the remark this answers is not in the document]
chrisbobbe: Thanks!

lib/widgets/message_list.dart
```diff
@@ -305,7 +305,7 @@ class MessageListAppBarTitle extends StatelessWidget {
   }) {
     // A null [Icon.icon] makes a blank space.
     final icon = (stream != null) ? iconDataForStream(stream) : null;
-    return Row(
+    final appBar = Row(
```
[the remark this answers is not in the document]
PIG208: While updating this PR to support message list app bar with two rows, I found a better place where this `GestureDetector` belongs to:

```dart
      case TopicNarrow(:var streamId, :var topic):
        final stream = store.streams[streamId];

        return SizedBox(
          width: double.infinity,
          child: GestureDetector(
            onLongPress: () => showTopicActionSheet(context,
              channelId: streamId, topic: topic),
            child: Column(
              children: [
                _buildStreamRow(context, stream: stream),
                _buildTopicRow(context, stream: stream, topic: topic),
              ])));
```

lib/api/route/channels.dart
```diff
@@ -38,3 +39,30 @@ class GetStreamTopicsEntry {
...
+  required UserTopicVisibilityPolicy visibilityPolicy,
+}) {
+  assert(visibilityPolicy != UserTopicVisibilityPolicy.unknown);
+  if (connection.zulipFeatureLevel! < 170) {
+    // TODO(server-7): deprecated in FL 170
+    assert(visibilityPolicy == UserTopicVisibilityPolicy.none
+           || visibilityPolicy == UserTopicVisibilityPolicy.muted);
+    final op = visibilityPolicy == UserTopicVisibilityPolicy.none ? 'remove'
+                                                                  : 'add';
+    return connection.patch('muteTopic', (_) {}, 'users/me/subscriptions/muted_topics', {
```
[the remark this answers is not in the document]
PIG208: Yeah. The story is that the caller of this is not expected to handle the error, mostly because it isn't helpful to the user. If the topic is already muted/unmuted, and that the user wants it to be muted/unmuted, respectively, there really isn't anything that requires actions from the user:

```diff
     api: Add route updateUserTopic

     For the legacy case, there can be an error when muting a topic that
-    is already muted or unmuting one that is already unmuted.  Let it
-    throw because we can't reliably filter out the error, which doesn't
-    have a specific "code".
+    is already muted or unmuting one that is already unmuted.
+
+    The callers are not expected to handle such errors because they aren't
+    really actionable.
```

lib/widgets/action_sheet.dart
```diff
@@ -181,6 +142,264 @@ class MessageActionSheetCancelButton extends StatelessWidget {
...
+
+      case (_, UserTopicVisibilityPolicy.muted):
+        return zulipLocalizations.actionSheetOptionMuteTopic;
+      case (_, UserTopicVisibilityPolicy.unmuted):
+        return zulipLocalizations.actionSheetOptionUnmuteTopic;
+      case (_, UserTopicVisibilityPolicy.followed):
+        return zulipLocalizations.actionSheetOptionFollowTopic;
+
+      case (_, UserTopicVisibilityPolicy.none):
+        return '';
```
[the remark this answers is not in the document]
PIG208: It turned out that this would be better as a separate case and keep the `assert(false)`. Added an explanation below:

```dart
      case (_, UserTopicVisibilityPolicy.none):
        // This is unexpected because `UserTopicVisibilityPolicy.muted` and
        // `UserTopicVisibilityPolicy.followed` (handled in separate `case`'s)
        // are the only expected `currentVisibilityPolicy`
        // when `newVisibilityPolicy` is `UserTopicVisibilityPolicy.none`.
        assert(false);
        return '';
```

PIG208:
I have updated the PR to implement `MessageListAppBarTitle` and to address the review feedback. Thanks @chrisbobbe for the review! The UI change is mostly separate from the topic action sheet sheet, so neither needs to block each other if we decide to implement it as a follow-up.

<details>
<summary>Compare to the legacy app</summary>

| | old | new | 
| - | - | - |
| ChannelNarrow |  |  |
| followed | 0ddde80f-fd24-4898-afd6-612853fe2d4b(1) |  |
| muted | |  |
| none |  |  |

</details>


<details>
<summary>Compare to the legacy app: font sizes</summary>

| | old | new |
| - | - | - |
| large, channel |  |  |
| large, topic |   |  |
| small, channel  |  |  |
| small, topic |  |  |

</details>

<details>
<summary>When the channel is unknown</summary>

| | |
| - | - |
| before |  |
| after |  |

</details>

lib/api/route/channels.dart
```diff
@@ -38,3 +39,30 @@ class GetStreamTopicsEntry {
 
   Map<String, dynamic> toJson() => _$GetStreamTopicsEntryToJson(this);
 }
+
+/// https://zulip.com/api/update-user-topic
+/// https://zulip.com/api/mute-topic (legacy)
+Future<void> updateUserTopic(ApiConnection connection, {
```
gnprice, line 45: Let's keep the API binding thin: the function named `updateUserTopic` just invokes the endpoint `update-user-topic`, not another endpoint.

To handle this compatibility fallback, we can use the same structure as `getMessage` and its wrapper `getMessageCompat`.

chrisbobbe:
> The UI change (app bar with two rows) is mostly separate from the topic action sheet sheet, so neither needs to block each other if we decide to implement it as a follow-up.

The two-row app bar is in the last commit:

```
msglist: Show channel name and topic name on two rows
```

Sure, fine to put the channel name and topic on separate lines. I see that makes it convenient to implement part of [this](https://github.com/zulip/zulip-flutter/pull/1041#pullrequestreview-2453170881):

> Thanks! Comments below, and I agree with your comment at [#348 (comment)](https://github.com/zulip/zulip-flutter/issues/348#issuecomment-2435903346) 🙂
> 
> > I think a design requirement of this feature is to also display mute/following states for each topic. At this point we can borrow that from the legacy mobile app.

As I mentioned there, and also [in my previous review](https://github.com/zulip/zulip-flutter/pull/1041#pullrequestreview-2453170881), we need to show the topic's mute/follow state, where it's needed, before we count #348 complete. I absolutely would block on that part. I would also not lump it into a commit that says it's about splitting the app bar into two lines.

In fact, as a branch-level comment, it should really get done before the commit that offers UI to change the state. Look what happens if I use the new UI in this revision:

https://github.com/user-attachments/assets/b7385ce2-b5b2-4059-9f5c-bf15dc033249

I long-press the recipient bar for the 6020 topic and tap "Follow topic". There's no visible change to the topic, so the user will think the button does nothing. The recipient bar needs to respond by showing the "follow" icon, like it does in zulip-mobile. If a piece of UI says it does something, it should clearly do that thing. We should hold that invariant at every commit, but especially at a commit that has the feature's `Fixes: ` mark.

…Actually, in a quick test on zulip-mobile, the recipient bar there shows the "follow" icon but it doesn't live-update on the event; you have to leave and re-enter the message list. Oops, that's not good! 🙂 In zulip-flutter it should live-update.

In zulip-mobile we show the icon in the Inbox page, and it live-updates there. We should do that too, because that's also a place we offer "Follow topic" etc.

gnprice:
> …Actually, in a quick test on zulip-mobile, the recipient bar there shows the "follow" icon but it doesn't live-update on the event; you have to leave and re-enter the message list. Oops, that's not good! 🙂

Yeah, I think that's a consequence of the "follow topic" feature being new in Zulip since after zulip-mobile went into maintenance mode. When I implemented that feature I probably just decided not to handle all the plumbing that'd be needed in zulip-mobile's webview-based message list to cause it to update on that change.

In zulip-flutter I think such an icon will naturally live-update if implemented in the most obvious way. (To be fair to RN: the same would be true in zulip-mobile for anything within the world of React components, so the bulk of the app except what's inside a message list. For example, the "follow" icon where it appears on the Inbox page, as noted.)

PIG208:
Oh I see! Yeah, that last commit only adds the icon to the app bar, but not the recipient headers. We need them for the inbox topic items as well. Will add those in my next update.

PIG208:
Haven't finished working on this yet (still need to self-review the added commits). Will post screenshots tomorrow.

PIG208:
Thanks for the previous reviews @chrisbobbe! This is ready for review now.

We now show the topic visibility polices at three different places:

<details>

| | light | dark |
| - | - | - |
| recipient header | recipient header light | recipient header dark |
| app bar | app bar light | app bar dark |
| inbox | inbox light | inbox dark |

</details>

lib/widgets/action_sheet.dart
```diff
@@ -146,9 +107,10 @@ abstract class MessageActionSheetMenuItemButton extends StatelessWidget {
         padding: const EdgeInsets.symmetric(vertical: 12, horizontal: 16),
         foregroundColor: designVariables.contextMenuItemText,
         splashFactory: NoSplash.splashFactory,
-      ).copyWith(backgroundColor: WidgetStateColor.resolveWith((states) =>
-          designVariables.contextMenuItemBg.withFadedAlpha(
-            states.contains(WidgetState.pressed) ? 0.20 : 0.12))),
+      ).copyWith(backgroundColor: WidgetStateColor.fromMap({
+        WidgetState.pressed: designVariables.contextMenuCancelPressedBg,
+        ~WidgetState.pressed: designVariables.contextMenuCancelBg,
```
gnprice, line 112: This seems wrong — it's using the cancel button's colors for the normal buttons.

(Noticed this in debugging with @rajveermalviya why the rebase of #1103 atop these changes had a color mismatch between these buttons and that PR's new ones.)

lib/widgets/action_sheet.dart
```diff
@@ -169,10 +131,9 @@ class MessageActionSheetCancelButton extends StatelessWidget {
         foregroundColor: designVariables.contextMenuCancelText,
         shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(7)),
         splashFactory: NoSplash.splashFactory,
-      ).copyWith(backgroundColor: WidgetStateColor.fromMap({
-        WidgetState.pressed: designVariables.contextMenuCancelPressedBg,
-        ~WidgetState.pressed: designVariables.contextMenuCancelBg,
-      })),
+      ).copyWith(backgroundColor: WidgetStateColor.resolveWith((states) =>
+          designVariables.contextMenuCancelBg.withFadedAlpha(
+            states.contains(WidgetState.pressed) ? 0.20 : 0.15))),
```
gnprice, line 136: Seems like the colors got swapped between the cancel button and the others.

Meanwhile, these colors are slightly changed from the previous colors for the normal buttons: was alpha 0.12, now 0.15. That happens in a commit marked NFC, so it's a bug of one kind or another.

PIG208: Updated the PR to fix that. This came from the push ended with 9c8316f8a02e383ec86d15239dfb6dd6aa6ad334. I think this is an error in resolving conflicts with the [commit](63146d378692635194cc733675d2d964b29f117c) that introduced the `contextMenuCancelBg` theme variable.

lib/widgets/inbox.dart
```diff
@@ -558,6 +560,6 @@ class _AtMentionMarker extends StatelessWidget {
     //   https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=224-16386&mode=design&t=JsNndFQ8fKFH0SjS-0
     return Padding(
       padding: const EdgeInsetsDirectional.only(end: 4),
-      child: Icon(ZulipIcons.at_sign, size: 14, color: designVariables.atMentionMarker));
+      child: Icon(icon, size: 14, color: designVariables.atMentionMarker));
```
chrisbobbe, line 569:
```
inbox [nfc]: Generalize _AtMentionMarker to take different icons
```

The code comment no longer fits this widget's more generalized role.

chrisbobbe (commented):
Thanks! Comments on the new code for showing the topic visibility icons.

lib/widgets/action_sheet.dart
gnprice: Cool. Yeah, always good to reread one's branch after a rebase that had conflicts. If you have your Git diff-viewer set up to show moves, then this pair of changes stands out because they're non-moves in the midst of a big move.

gnprice:
@PIG208 would you send those first five commits (the action-sheet refactor) as a quick separate PR? Since https://github.com/zulip/zulip-flutter/pull/1041#discussion_r1876958469 I think those will be ready for merge.

Then that will help us simplify managing merge conflicts with #1103 and #1076, apropos of [a discussion in chat today](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/conflict.20resolving.20for.20a.20preview.20release/near/1997645).

PIG208:
Opened #1125

lib/widgets/message_list.dart
```diff
@@ -325,12 +324,24 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+        Padding(padding: const EdgeInsetsDirectional.only(end: 8.0),
+          child: Icon(size: 20, icon)),
+        Flexible(child: Text(stream?.name ?? '(unknown stream)',
+          style: const TextStyle(
+            fontSize: 20,
+          ).merge(weightVariableTextStyle(context)))),
       ]);
   }
 
+  Widget _buildTopicRow(BuildContext context, {
```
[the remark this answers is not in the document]
PIG208: Yeah, if we want to support platform specific behaviors, we can't have `crossAxisAlignment: CrossAxisAlignment.start` on the `Column`. But in that case it will use the default `CrossAxisAlignment.center`, which makes the app bar look weird on an Android device:

<details>
<summary>On Android, without `CrossAxisAlignment.start`</summary>



</details>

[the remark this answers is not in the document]
gnprice: Yeah, I think a good behavior would be:
* on Android etc., both rows are start-aligned
* on iOS etc., both rows are center-aligned

following the same thing AppBar itself does by default for placing the `title` widget.

lib/widgets/message_list.dart
```diff
@@ -1064,11 +1065,23 @@ class StreamMessageRecipientHeader extends StatelessWidget {
...
+                        overflow: TextOverflow.ellipsis,
+                        style: recipientHeaderTextStyle(context))),
+                    Container(
+                      width: 20,
+                      padding: const EdgeInsetsDirectional.only(start: 4),
+                      // TODO(design): Choose an color for the icon
+                      child: Icon(size: 14, color: designVariables.atMentionMarker,
+                        // A null [Icon.icon] makes a blank space.
+                        iconDataForTopicVisibilityPolicy(
+                          store.topicVisibilityPolicy(message.streamId, topic)))),
```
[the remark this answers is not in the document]
PIG208: Yeah, this is an example of following the legacy app implementation too closely. I'm not sure what the intention was there.

https://github.com/zulip/zulip-mobile/blob/a115df1f71c9dc31e9b41060a8d57b51c017d786/src/streams/TopicItem.js#L47-L51

https://github.com/zulip/zulip-mobile/blob/a115df1f71c9dc31e9b41060a8d57b51c017d786/src/streams/TopicItem.js#L128

Using a `SizedBox` sounds good to me.

lib/widgets/message_list.dart
```diff
@@ -325,9 +324,32 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+          store.topicVisibilityPolicy(stream.streamId, topic));
+    return Row(
+      children: [
+        Flexible(child: Text(topic, style: const TextStyle(
+          fontSize: 13,
+        ).merge(weightVariableTextStyle(context)))),
+        if (icon != null)
+          Padding(
+            padding: const EdgeInsetsDirectional.only(start: 4),
+            child: Opacity(opacity: 0.4, child: Icon(icon, size: 14))),
```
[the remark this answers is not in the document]
PIG208: For inbox, I think it makes sense to reuse the "@" color and its variable right now. We can always add a new variable in the future, but for now having them look consistent seems fine.

lib/widgets/message_list.dart
```diff
@@ -327,10 +328,57 @@ class MessageListAppBarTitle extends StatelessWidget {
       children: [
         Icon(size: 16, icon),
         const SizedBox(width: 4),
-        Flexible(child: Text(text)),
+        Flexible(child: Text(stream?.name ?? '(unknown stream)')),
```
gnprice, line 331: should say "channel" (like it does in main)

lib/widgets/message_list.dart
```diff
@@ -327,10 +328,57 @@ class MessageListAppBarTitle extends StatelessWidget {
...
       ]);
   }
 
+  Widget _buildTopicRow(BuildContext context, {
+    required ZulipStream? stream,
+    required String topic,
+  }) {
+    final store = PerAccountStoreWidget.of(context);
+    final designVariables = DesignVariables.of(context);
+    final icon = (stream == null) ? null
```
gnprice, line 341: nit: extra parens

gnprice (commented):
Here's the start of a review from me — I've read these commits so far:
92f47a57c api: Add ApiConnection.patch
8c05e1cee api: Add route updateUserTopic and its compat helper
6db70dbeb icons: Take mute/unmute/following icons from the web app
78cf9e5ec msglist: Show channel name and topic name on two rows

lib/widgets/action_sheet.dart
```diff
@@ -143,6 +144,225 @@ class ActionSheetCancelButton extends StatelessWidget {
...
+  final visibilityPolicy = store.topicVisibilityPolicy(channelId, topic);
+
+  // TODO(server-7): simplify this condition away
+  final supportsUnmutingTopics = store.connection.zulipFeatureLevel! >= 170;
+  // TODO(server-8): simplify this condition away
+  final supportsFollowingTopics = store.connection.zulipFeatureLevel! >= 219;
+
+  final optionButtons = <ActionSheetMenuItemButton>[];
+  if (channelMuted != null && !channelMuted) {
+    switch (visibilityPolicy) {
```
gnprice, line 183: nit:
```suggestion
  if (channelMuted != null && !channelMuted) {
    // Channel is subscribed and not muted.
    switch (visibilityPolicy) {
```

And then for the next case:
```dart
    // Channel is muted.
```

I had the thought that these conditionals were a bit opaque and required some unpacking by the reader. Then I looked at the handy linked zulip-mobile code and saw it has comments for exactly that need :slightly_smiling_face: — so let's carry those over.

test/widgets/action_sheet_test.dart
```diff
@@ -99,6 +103,302 @@ void main() {
...
+          && widget.currentVisibilityPolicy == from
+          && widget.newVisibilityPolicy == to;
+      }), matching: find.text(label));
+    }
+
+    final mute =                 findButton(to:    UserTopicVisibilityPolicy.muted,
+                                            label: 'Mute topic');
+    final unmute =               findButton(from:  UserTopicVisibilityPolicy.muted,
+                                            to:    UserTopicVisibilityPolicy.none,
+                                            label: 'Unmute topic');
```
gnprice, line 190: Hmm, this is very specific about the details of how the implementation is arranged.

From the user's perspective it's just a button labeled "Unmute topic", right? And they tap it, and it should produce the right request that unmutes the topic.

So… (next comment below)

test/widgets/action_sheet_test.dart
```diff
@@ -99,6 +103,302 @@ void main() {
...
+        });
+    }
+
+    testWidgets('unmuteInMutedChannel', (tester) async {
+      await setupToTopicActionSheet(tester,
+        isChannelMuted: true,
+        visibilityPolicy: UserTopicVisibilityPolicy.none);
+      await tester.tap(unmuteInMutedChannel);
+      await tester.pump();
+      checkUpdateUserTopicRequest(UserTopicVisibilityPolicy.unmuted);
```
gnprice, line 269: … how about just:
```suggestion
      await tester.tap(find.text('Unmute topic'));
      await tester.pump();
      checkUpdateUserTopicRequest(UserTopicVisibilityPolicy.unmuted);
```

I think that covers all the facts that matter to the app's behavior as far as the user is concerned. And it both makes the test simpler, and less brittle to changes in the implementation.

gnprice: (Well, I guess there's one more fact this version doesn't check: the icon. But the icon isn't really being checked in the current revision either — just some inputs toward computing the icon.

It'd be a nice bonus to also check what icon is shown for each button, but I'll be happy merging this without that check.)

test/widgets/action_sheet_test.dart
```diff
@@ -99,6 +103,302 @@ void main() {
...
+      await prepare();
+      await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
+        child: const InboxPage()));
+      await tester.pump();
+
+      await tester.longPress(find.text(topic));
+      // sheet appears onscreen; default duration of bottom-sheet enter animation
+      await tester.pump(const Duration(milliseconds: 250));
+      check(find.byType(BottomSheet)).findsOne();
+    });
```
gnprice, line 135: We can be a bit more specific in pinning down that the bottom sheet that comes up is the right bottom sheet:
```suggestion
      check(find.byType(BottomSheet)).findsOne();
      check(find.text('Follow topic')).findsOne();
    });
```

test/widgets/action_sheet_test.dart
```diff
@@ -99,6 +103,302 @@ void main() {
...
+        'result': 'error', 'code': 'BAD_REQUEST', 'msg': ''});
+      await tester.tap(unfollow);
+      await tester.pumpAndSettle();
+
+      checkErrorDialog(tester, expectedTitle: 'Failed to unfollow topic');
+    });
+
+    group('check expected buttons', () {
+      final testCases = {
+        (false, UserTopicVisibilityPolicy.muted,    [unmute, follow]),
```
gnprice, line 323: nit: conceptually a list, not set
```suggestion
      final testCases = [
        (false, UserTopicVisibilityPolicy.muted,    [unmute, follow]),
```

(here and below)

test/widgets/action_sheet_test.dart
```diff
@@ -99,6 +103,302 @@ void main() {
...
+        (null,  UserTopicVisibilityPolicy.none,     <Finder>[]),
+      };
+
+      for (final (isChannelMuted, visibilityPolicy, buttons) in testCases) {
+        final description = 'isChannelMuted: ${isChannelMuted ?? "(not subscribed)"}, $visibilityPolicy';
+        testWidgets(description, (tester) async {
+          await setupToTopicActionSheet(tester,
+            isChannelMuted: isChannelMuted,
+            visibilityPolicy: visibilityPolicy,
+            featureLevel: 218);
```
gnprice, line 368: nit: use consistent standard name for a given concept
```suggestion
            zulipFeatureLevel: 218);
```

test/widgets/action_sheet_test.dart
```diff
@@ -99,6 +103,302 @@ void main() {
...
+        .findsExactly(expectedButtonFinders.length);
+    }
+
+    void checkUpdateUserTopicRequest(UserTopicVisibilityPolicy expectedPolicy) async {
+      check(connection.lastRequest).isA<http.Request>()
+        ..url.path.equals('/api/v1/user_topics')
+        ..bodyFields.deepEquals({
+          'stream_id': '${channel.streamId}',
+          'topic': topic,
+          'visibility_policy': '${expectedPolicy.toJson()}',
```
gnprice, line 259:
```suggestion
          'visibility_policy': jsonConvert(expectedPolicy),
```

Just calling `toString` (or doing string interpolation, which calls `toString`) on the `toJson` result happens to work here, because the toJson result is an int. But it doesn't get a JSON encoding in general. Better to just use the always-correct thing, which isn't any more complicated to write.

gnprice (commented):
OK, and here's the rest of a full review. Comments below; mostly nits, and a few substantive items in the tests.

test/widgets/action_sheet_test.dart
PIG208: Maybe something like:
```dart
    final mute =     find.text('Mute topic');
    final unmute =   find.text('Unmute topic');
    final follow =   find.text('Follow topic');
    final unfollow = find.text('Unfollow topic');
```

so that the later tests like this one:
```dart
      final testCases = [
        (false, UserTopicVisibilityPolicy.muted,    [unmute, follow]),
        (false, UserTopicVisibilityPolicy.none,     [mute, follow]),
        (false, UserTopicVisibilityPolicy.unmuted,  [mute, follow]),
        (false, UserTopicVisibilityPolicy.followed, [mute, unfollow]),

        (true,  UserTopicVisibilityPolicy.muted,    [unmute, follow]),
        (true,  UserTopicVisibilityPolicy.none,     [unmute, follow]),
        (true,  UserTopicVisibilityPolicy.unmuted,  [mute, follow]),
        (true,  UserTopicVisibilityPolicy.followed, [mute, unfollow]),

        (null,  UserTopicVisibilityPolicy.none,     <Finder>[]),
      ];
```
can still be formatted compactly.

gnprice: Sure, that works.

gnprice:
Well, score one for running `tools/check` after I rebase and before I push! There was a non-local merge conflict (i.e. semantically a merge conflict but not detected by Git):
```
Analyzing flutterz...                                                   

  error • Missing concrete implementations of 'getter
         ZulipLocalizations.actionSheetOptionFollowTopic', 'getter
         ZulipLocalizations.actionSheetOptionMuteTopic', 'getter
         ZulipLocalizations.actionSheetOptionUnfollowTopic', 'getter
         ZulipLocalizations.actionSheetOptionUnmuteTopic', and 4 more •
         lib/generated/l10n/zulip_localizations_fr.dart:8:7 •
         non_abstract_class_inherits_abstract_member
  error • Missing concrete implementations of 'getter
         ZulipLocalizations.actionSheetOptionFollowTopic', 'getter
         ZulipLocalizations.actionSheetOptionMuteTopic', 'getter
         ZulipLocalizations.actionSheetOptionUnfollowTopic', 'getter
         ZulipLocalizations.actionSheetOptionUnmuteTopic', and 4 more •
         lib/generated/l10n/zulip_localizations_pl.dart:8:7 •
         non_abstract_class_inherits_abstract_member
  error • Missing concrete implementations of 'getter
         ZulipLocalizations.actionSheetOptionFollowTopic', 'getter
         ZulipLocalizations.actionSheetOptionMuteTopic', 'getter
         ZulipLocalizations.actionSheetOptionUnfollowTopic', 'getter
         ZulipLocalizations.actionSheetOptionUnmuteTopic', and 4 more •
         lib/generated/l10n/zulip_localizations_ru.dart:8:7 •
         non_abstract_class_inherits_abstract_member

3 issues found. (ran in 5.5s)
```

Fixed it with:
```
$ git rebase -x 'tools/check l10n --fix'
```

gnprice:
(The issue was that in main we'd added those translations since this PR's current revision; so when adding strings, those generated files now needed to be updated just like the others.)

PR #1044, opened 2024-11-05
Clarify action sheet's "page context"; fix small bugs

gnprice:
This was prompted by my comment at https://github.com/zulip/zulip-flutter/pull/1041#discussion_r1828562960 (/cc @PIG208). But then as I got into this code I found some other small cleanups to make, and also spotted a couple of bugs to fix.

## Selected commit messages

action_sheet [nfc]: Rename pageContext field on button widgets

Most of the ways we use this field (specifically, all the ways we
use it other than in findMessageListPage) aren't related to it
being the message list in particular.

When we add action sheets other than the message action sheet,
they'll still need a page context.  So prepare the way for those
by giving this a name that can generalize.

---

action_sheet [nfc]: Avoid passing footgun context to onPressed methods

The usual Flutter idiom, which we use across the rest of the app,
would call for using the parameter called `context` wherever a
BuildContext is required: for getting the store or localizations,
or passing to a method like showErrorDialog, and so on.

That idiom doesn't work here, because that context belongs to the
action sheet itself, which gets dismissed at the start of the method.
Worse, it will sometimes seem to work, nondeterministically: the
context only gets unmounted after the animation to dismiss the action
sheet is complete, which takes something like 200ms.  So the
temptation to use it has been a recurring source of small bugs.

Fix the problem by not passing the action sheet's context to these
methods at all.

---

action_sheet: Mark as unread in current narrow, vs. from when action sheet opened

The narrow of a given MessageListPage can change over time,
in particular if viewing a topic that gets moved.

If it does, the mark-unread action should follow along.  Otherwise
we'd have a frustrating race where mark-unread just didn't work
if, for example, someone happened to resolve the topic while you
were in the middle of trying to mark as unread.

Fixes: #1045

### DIFF
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -484,10 +484,9 @@ class ShareButton extends MessageActionSheetMenuItemButton {
     // sheet. (We could do this after the sharing Future settles
     // with [ShareResultStatus.success], but on iOS I get impatient with
     // how slowly our action sheet dismisses in that case.)
-    // TODO(#24): Fix iOS bug where this call causes the keyboard to
-    //   reopen (if it was open at the time of this
-    //   `showMessageActionSheet` call) and cover a large part of the
-    //   share sheet.
+    // TODO(#591): Fix iOS bug where if the keyboard was open before the call
+    //   to `showMessageActionSheet`, it reappears briefly between
+    //   the `pop` of the action sheet and the appearance of the share sheet.
     Navigator.of(context).pop();
     final zulipLocalizations = ZulipLocalizations.of(pageContext);
 
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -183,11 +183,7 @@ class MessageActionSheetCancelButton extends StatelessWidget {
 // This button is very temporary, to complete #125 before we have a way to
 // choose an arbitrary reaction (#388). So, skipping i18n.
 class AddThumbsUpButton extends MessageActionSheetMenuItemButton {
-  AddThumbsUpButton({
-    super.key,
-    required super.message,
-    required super.pageContext,
-  });
+  AddThumbsUpButton({super.key, required super.message, required super.pageContext});
 
   @override IconData get icon => ZulipIcons.smile;
 
@@ -223,11 +219,7 @@ class AddThumbsUpButton extends MessageActionSheetMenuItemButton {
 }
 
 class StarButton extends MessageActionSheetMenuItemButton {
-  StarButton({
-    super.key,
-    required super.message,
-    required super.pageContext,
-  });
+  StarButton({super.key, required super.message, required super.pageContext});
 
   @override IconData get icon => _isStarred ? ZulipIcons.star_filled : ZulipIcons.star;
 
@@ -320,11 +312,7 @@ Future<String?> fetchRawContentWithFeedback({
 }
 
 class QuoteAndReplyButton extends MessageActionSheetMenuItemButton {
-  QuoteAndReplyButton({
-    super.key,
-    required super.message,
-    required super.pageContext,
-  });
+  QuoteAndReplyButton({super.key, required super.message, required super.pageContext});
 
   @override IconData get icon => ZulipIcons.format_quote;
 
@@ -380,11 +368,7 @@ class QuoteAndReplyButton extends MessageActionSheetMenuItemButton {
 }
 
 class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
-  MarkAsUnreadButton({
-    super.key,
-    required super.message,
-    required super.pageContext,
-  });
+  MarkAsUnreadButton({super.key, required super.message, required super.pageContext});
 
   @override IconData get icon => Icons.mark_chat_unread_outlined;
 
@@ -400,11 +384,7 @@ class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
 }
 
 class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
-  CopyMessageTextButton({
-    super.key,
-    required super.message,
-    required super.pageContext,
-  });
+  CopyMessageTextButton({super.key, required super.message, required super.pageContext});
 
   @override IconData get icon => ZulipIcons.copy;
 
@@ -437,11 +417,7 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
 }
 
 class CopyMessageLinkButton extends MessageActionSheetMenuItemButton {
-  CopyMessageLinkButton({
-    super.key,
-    required super.message,
-    required super.pageContext,
-  });
+  CopyMessageLinkButton({super.key, required super.message, required super.pageContext});
 
   @override IconData get icon => Icons.link;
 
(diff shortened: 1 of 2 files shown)


lib/widgets/action_sheet.dart
```diff
@@ -315,8 +324,7 @@ class QuoteAndReplyButton extends MessageActionSheetMenuItemButton {
     // This will be null only if the compose box disappeared after the
     // message action sheet opened, and before "Quote and reply" was pressed.
     // Currently a compose box can't ever disappear, so this is impossible.
-    ComposeBoxController composeBoxController =
-      MessageListPage.ancestorOf(messageListContext).composeBoxController!;
+    var composeBoxController = findMessageListPage().composeBoxController!;
```
PIG208, line 330: Is `var` used for fitting this in one line?

lib/widgets/action_sheet.dart
```diff
@@ -105,6 +104,16 @@ abstract class MessageActionSheetMenuItemButton extends StatelessWidget {
   final Message message;
   final BuildContext messageListContext;
 
+  /// The [MessageListPageState] this action sheet was triggered from.
+  ///
+  /// Uses the inefficient [BuildContext.findAncestorStateOfType];
+  /// don't call this in a build method.
+  MessageListPageState findMessageListPage() {
+    assert(messageListContext.mounted,
+      "findMessageListPage should be called only when messageListContext is known to still be mounted");
```
PIG208: nit: let's use `''` for consistency

test/widgets/action_sheet_test.dart
```diff
@@ -32,6 +33,17 @@ import 'compose_box_checks.dart';
 import 'dialog_checks.dart';
 import 'test_app.dart';
 
+GetMessagesResult _getMessagesResult(Message message) {
+  return GetMessagesResult(
+    anchor: message.id,
+    foundNewest: true,
+    foundOldest: true,
+    foundAnchor: true,
```
PIG208, line 41: Does the test rely on `foundAnchor: true` or `anchor: message.id`? If not, `eg.newestResult` suffices.

test/widgets/action_sheet_test.dart
```diff
@@ -442,6 +445,44 @@ void main() {
...
+          processedCount: 11, updatedCount: 3,
+          firstProcessedId: 1, lastProcessedId: 1980,
+          foundOldest: true, foundNewest: true).toJson());
+        await tester.tap(find.byIcon(Icons.mark_chat_unread_outlined, skipOffstage: false));
+        await tester.pumpAndSettle();
+        check(connection.lastRequest).isA<http.Request>()
+          ..method.equals('POST')
+          ..url.path.equals('/api/v1/messages/flags/narrow')
+          ..bodyFields['narrow'].equals(
+              jsonEncode(TopicNarrow(newStream.streamId, newTopic).apiEncode()));
```
PIG208, line 464: nit: indentation

test/widgets/action_sheet_test.dart
```diff
@@ -442,6 +445,44 @@ void main() {
             });
       });
 
+      testWidgets('on topic move, acts on new topic', (tester) async {
+        final stream = eg.stream();
+        const topic = 'old topic';
+        final message = eg.streamMessage(flags: [MessageFlag.read],
+          stream: stream, topic: topic);
+        await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
```
PIG208, line 435: nit: it might be good to wrap this line because `narrow` is particularly interesting to this test

test/widgets/action_sheet_test.dart
```diff
@@ -442,6 +445,44 @@ void main() {
...
+        // Get the action sheet fully deployed while the old narrow applies.
+        await tester.pumpAndSettle();
```
PIG208, line 437: Hmm, the test passes and would fail as expected without the fix, if I remove this. I see that `setupToMessageActionSheet` calls `tester.pump` with a 250ms delay. Maybe that's sufficient?

test/widgets/action_sheet_test.dart
```diff
@@ -442,6 +445,44 @@ void main() {
...
+        final newStream = eg.stream();
+        const newTopic = 'other topic';
+        // This result isn't quite realistic for this request: it should get
+        // the updated channel/stream ID and topic, because we don't even
+        // start the request until after we get the move event.
+        // But constructing the right result is annoying at the moment, and
+        // it doesn't matter anyway: [MessageStoreImpl.reconcileMessages] will
+        // keep the version updated by the event.  If that somehow changes in
+        // some future refactor, it'll cause this test to fail.
+        connection.prepare(json: _getMessagesResult(message).toJson());
```
PIG208, line 448: For tests where we need to prepare the new messages, I [typically used](https://github.com/zulip/zulip-flutter/blob/ffa41be876743ee9b57982c7a315e6f97fe65491/test/model/message_list_test.dart#L926-L938) `eg.updateMessageEventMoveTo` with the message that has the new channel and topic, but I can see that we are in favor of `eg.updateMessageEventMoveFrom` here because in this context — moving away from the current narrow — the "move from" variant feels a bit more natural to use.

PIG208 (commented):
Thanks for the PR! Left some small comments. Otherwise this looks good to me.

lib/widgets/action_sheet.dart
gnprice: More that specifying the type feels redundant. It'd look like this:

Well, the alternative is:
```dart
    ComposeBoxController composeBoxController =
      findMessageListPage().composeBoxController!;
```

(or the same with `findMessageListPage()` moved before the newline).

Our usual practice is to say `final` and let the type be inferred. Here it's not `final` because it gets updated below; so `var` is the choice that corresponds to that.

test/widgets/action_sheet_test.dart
gnprice: Yeah, `newestResult` would be an improvement. (Though it doesn't currently live in `eg`; maybe it should, with a longer name in that case.)

Adding a separate prep commit to switch to that.

test/widgets/action_sheet_test.dart
gnprice: This looks right to me. Do you mean indenting it like this?
```dart
          ..bodyFields['narrow'].equals(
            jsonEncode(TopicNarrow(newStream.streamId, newTopic).apiEncode()));
```

That'd make `bodyFields` and `jsonEncode` appear to be at the same level, which isn't desired.

FWIW dartfmt agrees with my indentation (though it moves `jsonEncode(` to before the newline). I checked with a quick Ctrl+Alt+L after selecting the surrounding few lines.

test/widgets/action_sheet_test.dart
gnprice: I'll expand the comment. Does this version explain why this call is here?
```dart
        // Get the action sheet fully deployed while the old narrow applies.
        // (This way we maximize the range of potential bugs this test can catch,
        // by giving the code maximum opportunity to latch onto the old topic.)
        await tester.pumpAndSettle();
```

gnprice:
Thanks @PIG208 for the detailed review! Pushed a revision, and replied to some subthreads above; PTAL.

PIG208:
LGTM. Thanks for the update!

Issue #1046, opened 2024-11-05
Handle message links (a.message-link)


gnprice:
The server now emits a new shape of content when the user links to a message:
```html
<a class="message-link"
  href="/#narrow/channel/9-announce/topic/Zulip.20updates/near/214">
 #announce &gt; Zulip updates @ 💬
</a>
```

This is suddenly pretty common, too, because the web client automatically converts to the syntax for this when the user pastes such a URL.

This is newly merged today. Docs will appear here:
https://zulip.com/api/message-formatting
but for the moment can be read instead here:
https://github.com/zulip/zulip/commit/77064a17260ce888de73a9b7c01e33550e4d05be

We should handle these. Currently they're showing as "unimplemented".

For this issue, to get it merged quickly, we just want basic support: treat the link like any other link. A fancier feature is to follow renames of the stream/channel; that will be a separate issue and is out of scope for this one.

* Server/web issue: https://github.com/zulip/zulip/issues/31920
* Server/web PR: https://github.com/zulip/zulip/pull/31965
* `#api design` thread: https://chat.zulip.org/#narrow/channel/378-api-design/topic/notation.20for.20near.20links/near/1972281

PR #1048, opened 2024-11-06
content: Handle message links

rajveermalviya:
Fixes: #1046

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -855,7 +855,9 @@ class _ZulipContentParser {
 
     if (localName == 'a'
         && (className.isEmpty
-            || (className == 'stream-topic' || className == 'stream'))) {
+            || className == 'stream-topic'
+            || className == 'stream'
+            || className == 'message-link')) {
       final href = element.attributes['href'];
       if (href == null) return unimplemented();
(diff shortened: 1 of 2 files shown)


lib/model/content.dart
```diff
@@ -855,7 +855,9 @@ class _ZulipContentParser {
 
     if (localName == 'a'
         && (className.isEmpty
-            || (className == 'stream-topic' || className == 'stream'))) {
+            || (className == 'stream-topic'
+                || className == 'stream'
+                || className == 'message-link'))) {
```
gnprice, line 860: nit:
```suggestion
            || className == 'stream-topic'
            || className == 'stream'
            || className == 'message-link')) {
```

(I think the extra parens are a legacy of how the logic used to be more complicated with `classes`; introduced in 3cd84bb97a3661ecc464b6227040aa0e9a10e0e0, simplified in 4c945477badbbad050eacea0322da2c326a2a92d but could have been simplified a bit further.)

gnprice (commented):
Thanks @rajveermalviya for the rapid implementation! Comments below.

rajveermalviya:
Thanks for the review @gnprice! Pushed a new revision — PTAL.

gnprice:
Thanks! Looks good; merging.

Issue #1016, opened 2024-10-21
Handle `message_embed` website previews


gnprice:
I.e., this Zulip feature:
https://zulip.com/help/image-video-and-website-previews#configure-whether-website-previews-are-shown
which we've also called "inline URL preview".

The HTML elements for these look like `<div class="message_embed">`. Example message here:
https://chat.zulip.org/#narrow/stream/101-design/topic/auto-linkification.20of.20issues/near/193181

On chat.zulip.org these are scarce because it's a realm-level setting which we turned off after a brief period enabled, so it's easy for us to forget about this feature. But on realms that have the setting turned on, they'll naturally be much more common.

These appeared in the results of the survey #917 / #190 of unimplemented content features. Thanks to @andersk for pointing out (https://github.com/zulip/zulip-flutter/issues/965#issuecomment-2427591885) that this is a current feature.

Issue #681, opened 2024-05-15
msglist: Use special background color for DM messages


chrisbobbe:
This is part of #157.

The web app uses a special background color for DM messages:

<img width="743" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/f2e8de96-f0f4-4d89-988b-ca9b74a27c88">

This color is currently identified as `--color-background-private-message-content` in web/styles/app_variables.css in zulip/zulip.

We should use this color.

We should follow the web app in also giving this background to a DM message's corresponding date separator, if any:

<img width="746" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/a6cfef92-63dd-4d0d-8c35-046427ca661b">

pranay-sankar-muppanenei:
is this issue is still open,if it is please assign it to me?

i have some doubts about this issue.

in a web app there is a different background color for a message but it is not the case with the mobile apps.this is what i understood.

can you please give some clarity?

E-m-i-n-e-n-c-e:
Fixed it. Only had to change like 8-10 lines of code. Please assign it to me

![Image](https://github.com/user-attachments/assets/28848946-c605-455b-bc9b-0b412d2ef51f)

![Image](https://github.com/user-attachments/assets/221ddbbd-6383-455b-9ff1-98ec463f386e)

![Image](https://github.com/user-attachments/assets/dcf519b2-81f4-4fe4-94b2-5d3ce11ce996)

![Image](https://github.com/user-attachments/assets/7f6d8787-fcc7-4b5b-95d3-a41f562caec6)

PR #1049, opened 2024-11-06
content: Handle link previews

rajveermalviya:
Fixes: #1016

<details>
<summary> Screenshots </summary>

| Web | Flutter |
| - | - |
| Screenshot 2025-02-24 at 12 27 55 | Screenshot 2025-02-24 at 12 26 56 |
| Screenshot 2025-02-24 at 12 28 55 | Screenshot 2025-02-24 at 12 28 33 |

| Web (Small) | Flutter (Small) |
| - | - |
| Screenshot 2025-02-24 at 12 37 54 | Screenshot 2025-02-24 at 12 38 23 |
| Screenshot 2025-02-24 at 12 38 13 | Screenshot 2025-02-24 at 12 37 19 |

</details>

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -1391,7 +1391,7 @@ class _ZulipContentParser {
     return EmbedVideoNode(hrefUrl: href, previewImageSrcUrl: imgSrc, debugHtmlNode: debugHtmlNode);
   }
 
-  static final _websitePreviewImageSrcRegexp = RegExp(r'background-image: url\("(.+)"\)');
+  static final _websitePreviewImageSrcRegexp = RegExp(r'background-image: url\(("?)(.+?)\1\)');
 
   BlockContentNode parseWebsitePreviewNode(dom.Element divElement) {
     assert(divElement.localName == 'div'
@@ -1415,7 +1415,7 @@ class _ZulipContentParser {
       ]) {
         final match = _websitePreviewImageSrcRegexp.firstMatch(imageStyleAttr);
         if (match == null) return null;
-        final imageSrcUrl = match.group(1);
+        final imageSrcUrl = match.group(2);
         if (imageSrcUrl == null) return null;
 
         String? parseTitle(dom.Element element) {
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -504,6 +504,58 @@ class EmbedVideoNode extends BlockContentNode {
   }
 }
 
+// See:
+//  https://ogp.me/
+//  https://oembed.com/
+//  https://zulip.com/help/image-video-and-website-previews#configure-whether-website-previews-are-shown
+class WebsitePreviewNode extends BlockContentNode {
+  const WebsitePreviewNode({
+    super.debugHtmlNode,
+    required this.hrefUrl,
+    required this.imageSrcUrl,
+    required this.title,
+    required this.description,
+  });
+
+  /// The URL from which this preview data was retrieved.
+  final String hrefUrl;
+
+  /// The image URL representing the webpage, content value
+  /// of `og:image` HTML meta property.
+  final String imageSrcUrl;
+
+  /// Represents the webpage title, derived from either
+  /// the content of the `og:title` HTML meta property or
+  /// the <title> HTML element.
+  final String? title;
+
+  /// Description about the webpage, content value of
+  /// `og:description` HTML meta property.
+  final String? description;
+
+  @override
+  bool operator ==(Object other) {
+    return other is WebsitePreviewNode
+      && other.hrefUrl == hrefUrl
+      && other.imageSrcUrl == imageSrcUrl
+      && other.title == title
+      && other.description == description;
+  }
+
+  @override
+  int get hashCode =>
+    Object.hash('WebsitePreviewNode', hrefUrl, imageSrcUrl, title, description);
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(StringProperty('hrefUrl', hrefUrl));
+    properties.add(StringProperty('imageSrcUrl', imageSrcUrl));
+    properties.add(StringProperty('title', title));
+    properties.add(StringProperty('description', description));
+  }
+}
+
 class TableNode extends BlockContentNode {
   const TableNode({super.debugHtmlNode, required this.rows});
 
@@ -1339,6 +1391,113 @@ class _ZulipContentParser {
     return EmbedVideoNode(hrefUrl: href, previewImageSrcUrl: imgSrc, debugHtmlNode: debugHtmlNode);
   }
 
+  static final _websitePreviewImageSrcRegexp = RegExp(r'background-image: url\("(.+)"\)');
+
+  BlockContentNode parseWebsitePreviewNode(dom.Element divElement) {
+    assert(divElement.localName == 'div'
+      && divElement.className == 'message_embed');
+
+    final debugHtmlNode = kDebugMode ? divElement : null;
+    final result = () {
+      if (divElement.nodes case [
+        dom.Element(
+          localName: 'a',
+          className: 'message_embed_image',
+          attributes: {
+            'href': final String imageHref,
+            'style': final String imageStyleAttr,
+          },
+          nodes: []),
+        dom.Element(
+          localName: 'div',
+          className: 'data-container',
+          nodes: [...]) && final dataContainer,
+      ]) {
+        final match = _websitePreviewImageSrcRegexp.firstMatch(imageStyleAttr);
+        if (match == null) return null;
+        final imageSrcUrl = match.group(1);
+        if (imageSrcUrl == null) return null;
+
+        String? parseTitle(dom.Element element) {
+          assert(element.localName == 'div' &&
+            element.className == 'message_embed_title');
+          if (element.nodes case [
+            dom.Element(localName: 'a', className: '') && final child,
+          ]) {
+            final titleHref = child.attributes['href'];
+            // Make sure both image hyperlink and title hyperlink are same.
+            if (imageHref != titleHref) return null;
+
+            if (child.nodes case [dom.Text(text: final title)]) {
+              return title;
+            }
+          }
+          return null;
+        }
+
+        String? parseDescription(dom.Element element) {
+          assert(element.localName == 'div' &&
+            element.className == 'message_embed_description');
+          if (element.nodes case [dom.Text(text: final description)]) {
+            return description;
+          }
+          return null;
+        }
+
+        String? title, description;
+        switch (dataContainer.nodes) {
+          case [
+            dom.Element(
+              localName: 'div',
+              className: 'message_embed_title') && final first,
+            dom.Element(
+              localName: 'div',
+              className: 'message_embed_description') && final second,
+          ]:
+            title = parseTitle(first);
+            if (title == null) return null;
+            description = parseDescription(second);
+            if (description == null) return null;
+
+          case [dom.Element(localName: 'div') && final single]:
+            switch (single.className) {
+              case 'message_embed_title':
+                title = parseTitle(single);
+                if (title == null) return null;
+
+              case 'message_embed_description':
+                description = parseDescription(single);
+                if (description == null) return null;
+
+              default:
+                return null;
+            }
+
+          case []:
+            // Server generates an empty `<div class="data-container"></div>`
+            // if website HTML has neither title (derived from
+            // `og:title` or `<title>…</title>`) nor description (derived from
+            // `og:description`).
+            break;
+
+          default:
+            return null;
+        }
+
+        return WebsitePreviewNode(
+          hrefUrl: imageHref,
+          imageSrcUrl: imageSrcUrl,
+          title: title,
+          description: description,
+          debugHtmlNode: debugHtmlNode);
+      } else {
+        return null;
+      }
+    }();
+
+    return result ?? UnimplementedBlockContentNode(htmlNode: divElement);
+  }
+
   BlockContentNode parseTableContent(dom.Element tableElement) {
     assert(tableElement.localName == 'table'
         && tableElement.className.isEmpty);
@@ -1581,6 +1740,10 @@ class _ZulipContentParser {
             return parseEmbedVideoNode(element);
         }
       }
+    }
+
+    if (localName == 'div' && className == 'message_embed') {
+      return parseWebsitePreviewNode(element);
     }
 
     // TODO more types of node
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -42,6 +42,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
       colorDirectMentionBackground: const HSLColor.fromAHSL(0.2, 240, 0.7, 0.7).toColor(),
       colorGlobalTimeBackground: const HSLColor.fromAHSL(1, 0, 0, 0.93).toColor(),
       colorGlobalTimeBorder: const HSLColor.fromAHSL(1, 0, 0, 0.8).toColor(),
+      colorLink: const HSLColor.fromAHSL(1, 200, 1, 0.4).toColor(),
       colorMathBlockBorder: const HSLColor.fromAHSL(0.15, 240, 0.8, 0.5).toColor(),
       colorMessageMediaContainerBackground: const Color.fromRGBO(0, 0, 0, 0.03),
       colorPollNames: const HSLColor.fromAHSL(1, 0, 0, .45).toColor(),
@@ -75,6 +76,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
       colorDirectMentionBackground: const HSLColor.fromAHSL(0.25, 240, 0.52, 0.6).toColor(),
       colorGlobalTimeBackground: const HSLColor.fromAHSL(0.2, 0, 0, 0).toColor(),
       colorGlobalTimeBorder: const HSLColor.fromAHSL(0.4, 0, 0, 0).toColor(),
+      colorLink: const HSLColor.fromAHSL(1, 200, 1, 0.4).toColor(), // the same as light in Web
       colorMathBlockBorder: const HSLColor.fromAHSL(1, 240, 0.4, 0.4).toColor(),
       colorMessageMediaContainerBackground: const HSLColor.fromAHSL(0.03, 0, 0, 1).toColor(),
       colorPollNames: const HSLColor.fromAHSL(1, 236, .15, .7).toColor(),
@@ -107,6 +109,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
     required this.colorDirectMentionBackground,
     required this.colorGlobalTimeBackground,
     required this.colorGlobalTimeBorder,
+    required this.colorLink,
     required this.colorMathBlockBorder,
     required this.colorMessageMediaContainerBackground,
     required this.colorPollNames,
@@ -139,6 +142,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
   final Color colorDirectMentionBackground;
   final Color colorGlobalTimeBackground;
   final Color colorGlobalTimeBorder;
+  final Color colorLink;
   final Color colorMathBlockBorder; // TODO(#46) this won't be needed
   final Color colorMessageMediaContainerBackground;
   final Color colorPollNames;
@@ -199,6 +203,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
     Color? colorDirectMentionBackground,
     Color? colorGlobalTimeBackground,
     Color? colorGlobalTimeBorder,
+    Color? colorLink,
     Color? colorMathBlockBorder,
     Color? colorMessageMediaContainerBackground,
     Color? colorPollNames,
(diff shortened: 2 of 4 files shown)


lib/widgets/content.dart
```diff
@@ -799,6 +801,90 @@ class MathBlock extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    final messageListTheme = MessageListTheme.of(context);
+    final isSmallWidth = MediaQuery.sizeOf(context).width <= 576;
+
+    final dataContainer = Container(
+      constraints: const BoxConstraints(maxHeight: 80),
+      padding: const EdgeInsets.symmetric(horizontal: 5),
+      child: InsetShadowBox(
+        bottom: 8,
+        color: messageListTheme.streamMessageBgDefault,
```
chrisbobbe, line 819: This looks like it'll be wrong for DMs, and (in future) for messages where we highlight the background because of @-mentions in the message (#647).

lib/widgets/content.dart
```diff
@@ -799,6 +801,90 @@ class MathBlock extends StatelessWidget {
...
+            crossAxisAlignment: CrossAxisAlignment.start,
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              if (node.title != null)
+                GestureDetector(
+                  onTap: () => _launchUrl(context, node.hrefUrl),
+                  child: Text(node.title!,
+                    style: TextStyle(
+                      fontSize: 1.2 * kBaseFontSize,
+                      color: const HSLColor.fromAHSL(1, 200, 1, 0.4).toColor()))),
```
chrisbobbe, line 835: I see a hard-coded color; does it follow web? It needs either a variable in `ContentTheme` or this comment:

```dart
// Web has the same color in light and dark mode.
```

(Same for any other hard-coded colors.)

Please also post screenshots in light mode; I see screenshots for dark mode already.

lib/widgets/content.dart
```diff
@@ -799,6 +801,90 @@ class MathBlock extends StatelessWidget {
...
+                  constraints: const BoxConstraints(maxWidth: 500),
+                  child: Text(node.description!)),
+              ],
+              const SizedBox(height: 8),
+            ]))));
+
+    if (isSmallWidth) {
+      return Container(
+        decoration: const BoxDecoration(border:
+          Border(left: BorderSide(color: Color(0xFFEDEDED), width: 3))),
```
chrisbobbe, line 848: Same comment about hard-coded colors; also, I think we more often use lowercase (so `0xffededed` instead of `0xFFEDEDED`); here and below.

lib/widgets/content.dart
```diff
@@ -799,6 +801,90 @@ class MathBlock extends StatelessWidget {
...
+  const MessageLinkPreview({super.key, required this.node});
+
+  final LinkPreviewNode node;
+
+  @override
+  Widget build(BuildContext context) {
+    final messageListTheme = MessageListTheme.of(context);
+    final isSmallWidth = MediaQuery.sizeOf(context).width <= 576;
+
+    final dataContainer = Container(
```
chrisbobbe, line 814: Is `dataContainer` the best name? I see the `Container` widget being used…what's the "data" and how does that widget "contain" it?

How about building this method's return value with help from a mutable `Widget result` variable? So here:

`Widget result = Container(/* etc. */);`

then below,

```dart
result = isSmallWidth
  ? Column(/* etc. */, children: [/* etc. */, result])
  : Row(/* etc. */, children: [/* etc. */, result]);
```

then `result = Container(decoration: /* etc. */, child: result);`

and finally `return result;`.

chrisbobbe (commented):
Thanks! Looks like this needs a rebase, so I'll do a more thorough review after that. But here are some comments from a quick skim (in particular I haven't read the parsing code or checked the UI code against web).

rajveermalviya:
Thanks for the initial comments @chrisbobbe! Pushed a new revision, PTAL.

lib/widgets/content.dart
rajveermalviya: Updated the screenshots.

chrisbobbe:
What's a good way for me to test this; do I need to set up a dev server? 🙂 I see link previews are disabled on CZO; I've asked on CZO if there's a reason for that: https://chat.zulip.org/#narrow/channel/9-issues/topic/Link.20previews.20for.20Zulip.20URLs/near/2013846

gnprice:
Dev server, or make a test realm on Zulip Cloud.

lib/widgets/content.dart
```diff
@@ -839,6 +840,94 @@ class MathBlock extends StatelessWidget {
...
+      ]);
+
+    final clippedTitleAndDescription = Container(
+      constraints: const BoxConstraints(maxHeight: 80),
+      padding: const EdgeInsets.symmetric(horizontal: 5),
+      child: InsetShadowBox(
+        bottom: 8,
+        // TODO(#647) use different color for highlighted messages
+        // TODO(#681) use different color for DM messages
+        color: messageListTheme.streamMessageBgDefault,
```
chrisbobbe, line 881: Ah right; yeah, I forgot we haven't done #681 yet.

I guess this needs one more TODO I hadn't thought of before:

```dart
// TODO(#488) use different color for non-message contexts
```

Probably the desired effect of that TODO will be to guide the implementation toward a color param rather than a param that's about the aspects of a Zulip message.

lib/widgets/content.dart
```diff
@@ -839,6 +840,94 @@ class MathBlock extends StatelessWidget {
...
+          alignment: Alignment.topLeft,
+          constrainedAxis: Axis.horizontal,
+          clipBehavior: Clip.antiAlias,
+          child: Padding(
+            padding: const EdgeInsets.only(bottom: 8.0),
+            child: isSmallWidth
+              ? titleAndDescription
+              : LayoutBuilder(
+                  builder: (context, constraints) => ConstrainedBox(
+                    constraints: BoxConstraints(
```
chrisbobbe, line 892: There's a lot about constraints in this code: a `Container.constraints`, an `UnconstrainedBox`, a `LayoutBuilder`, a `ConstrainedBox.constraints`. I'm not really following it yet; do you think there might be a simpler way to write it?

lib/widgets/content.dart
```diff
@@ -839,6 +840,94 @@ class MathBlock extends StatelessWidget {
...
+            clippedTitleAndDescription,
+          ])
+      : Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
+          GestureDetector(
+            onTap: () => _launchUrl(context, node.hrefUrl),
+            child: RealmContentNetworkImage(Uri.parse(node.imageSrcUrl),
+              fit: BoxFit.cover,
+              width: 80,
+              height: 80,
+              alignment: Alignment.center)),
```
chrisbobbe, line 917: Isn't `Alignment.center` the default; can we leave out this argument?

lib/widgets/content.dart
```diff
@@ -839,6 +840,94 @@ class MathBlock extends StatelessWidget {
...
+          constrainedAxis: Axis.horizontal,
+          clipBehavior: Clip.antiAlias,
+          child: Padding(
+            padding: const EdgeInsets.only(bottom: 8.0),
+            child: isSmallWidth
+              ? titleAndDescription
+              : LayoutBuilder(
+                  builder: (context, constraints) => ConstrainedBox(
+                    constraints: BoxConstraints(
+                      maxWidth: constraints.maxWidth - 115),
```
chrisbobbe, line 893: I wonder about dropping this `maxWidth - 115` detail, for a few benefits:

- More of the text can show before it gets clipped
- Removes the need for `LayoutBuilder` which isn't great for
  - performance
  - code complexity (e.g. my difficulty in a previous comment)
- We allow more horizontal space for other paragraph content without issues

Could leave a code comment saying we're not following web in this way.

lib/widgets/content.dart
```diff
@@ -839,6 +840,94 @@ class MathBlock extends StatelessWidget {
...
+              alignment: Alignment.center)),
+          Flexible(child: clippedTitleAndDescription),
+        ]);
+
+    return Container(
+      decoration: const BoxDecoration(
+        border: Border(left: BorderSide(
+          // Web has the same color in light and dark mode.
+          color: Color(0xffededed), width: 3))),
+      padding: const EdgeInsets.all(5),
```
chrisbobbe, line 926: Web also puts 5px bottom margin, a.k.a. `--markdown-interelement-space-px`, in addition to this. In zulip-flutter do we have something systematic for vertical spacing between block elements, or is each element responsible for adding its own space at the bottom and/or top?

chrisbobbe (commented):
Thanks! Comments below.

lib/widgets/content.dart
rajveermalviya: Removed this `LayoutBuilder` here.

lib/widgets/content.dart
rajveermalviya: From a [quick test](https://chat.zulip.org/#narrow/channel/7-test-here/topic/.E2.9C.94.20Rajesh/near/2052366), looks like the spacing between different widgets is not consistent (or non-existent in some cases). On Web `--markdown-interelement-space-px` is calculated from line-height (which for me becomes 6.72px instead of 5px), so we'd need something similar for zulip-flutter too. However, it'll need a sweep across all the content widgets therefore making it out-of-scope for this PR.

Should 5px bottom margin as a quick fix, suffice for now?

rajveermalviya:

lib/widgets/content.dart
chrisbobbe: > Should 5px bottom margin as a quick fix, suffice for now?

Sure, sounds good

lib/widgets/content.dart
```diff
@@ -148,6 +151,7 @@ class ContentTheme extends ThemeExtension<ContentTheme> {
   final Color colorTableCellBorder;
   final Color colorTableHeaderBackground;
   final Color colorThematicBreak;
+  final Color colorLink;
```
chrisbobbe, line 155: nit: keep in alphabetical order (here and elsewhere in class definition)

lib/widgets/content.dart
```diff
@@ -1030,7 +1037,7 @@ class _InlineContentBuilder {
         _pushRecognizer(recognizer);
         final result = _buildNodes(node.nodes,
           // Web has the same color in light and dark mode.
```
chrisbobbe, line 1128: Let's remove this comment; it would be easy for it to become wrong if the light/dark variants start to be different. Also you've already left a comment with the same meaning on the dark variant.

chrisbobbe (commented):
Looks good, thanks! Just nits below, and I think https://github.com/zulip/zulip-flutter/pull/1049#discussion_r1936507952 is still open. Otherwise LGTM; marking for Greg's review.

test/model/content_test.dart
```diff
@@ -1155,6 +1155,67 @@ class ContentExample {
       ], isHeader: false),
     ]),
   ]);
+
+  static const linkPreviewSmoke = ContentExample(
+    'link preview smoke',
+    'https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title+description.html',
+    '<p><a href="https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title+description.html">https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title+description.html</a></p>\n'
+      '<div class="message_embed">'
```
gnprice, line 1163: nit: indentation
```suggestion
    '<p><a href="https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title+description.html">https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title+description.html</a></p>\n'
    '<div class="message_embed">'
```

The `div` is a sibling of the `p`, not a child, so the indentation should express that.

lib/model/content.dart
```diff
@@ -1453,6 +1504,101 @@ class _ZulipContentParser {
...
+            switch (single.className) {
+              case 'message_embed_title':
+                title = parseTitle(single);
+                if (title == null) return null;
+              case 'message_embed_description':
+                description = parseDescription(single);
+                if (description == null) return null;
+            }
+
+          default:
```
gnprice, line 1584: It looks like this would accept the children being a single `div` with some other unexpected class. I think that's unintended.

lib/model/content.dart
```diff
@@ -1453,6 +1504,101 @@ class _ZulipContentParser {
...
+            return null;
+        }
+
+        return LinkPreviewNode(
```
gnprice, line 1588: It looks like this parser code has (modulo my comment just above) an invariant that it always has at least one of `title` or `description` non-null when it decides to construct a LinkPreviewNode.

Is that invariant something you believe holds? It seems like the sort of thing that'd be helpful for reasoning about (the avoidance of) edge cases when displaying these as widgets.

If so, we can add an `assert` at the constructor to confirm that it holds. That'd make it something the widgets code can count on.

lib/model/content.dart
```diff
@@ -1453,6 +1504,101 @@ class _ZulipContentParser {
...
+          hrefUrl: imageHref,
+          imageSrcUrl: imageSrcUrl,
+          title: title,
+          description: description,
+        );
+      } else {
+        return null;
+      }
```
gnprice, line 1596: I sure am hoping for Dart to at some point gain an analogue of Rust's let-else, or Swift's guard-let :slightly_smiling_face:

gnprice (commented):

Here's an initial round, just on `model/content.dart`. Posting this now because about to switch to another task.

rajveermalviya:

lib/model/content.dart
```diff
@@ -1339,6 +1391,112 @@ class _ZulipContentParser {
...
+        }
+
+        String? title, description;
+        switch (dataContainer.nodes) {
+          case []:
+            // Server generates an empty `<div class="data-container"></div>`
+            // if website HTML doesn't have both title (derived from
+            // `og:title` or `<title>…</title>`) and description (derived from
+            // `og:description`).
+            break;
```
gnprice, line 1453:
```suggestion
          case []:
            // Server generates an empty `<div class="data-container"></div>`
            // if website HTML has neither title (derived from
            // `og:title` or `<title>…</title>`) nor description (derived from
            // `og:description`).
            break;
```

I'd read "if doesn't have both A and B" as meaning "if not (has A and has B)". But I think that's not what you meant — if that were the case, the single-child case below shouldn't be possible.

lib/model/content.dart
```diff
@@ -1339,6 +1391,112 @@ class _ZulipContentParser {
...
+            element.className == 'message_embed_description');
+          if (element.nodes case [dom.Text(text: final description)]) {
+            return description;
+          }
+          return null;
+        }
+
+        String? title, description;
+        switch (dataContainer.nodes) {
+          case []:
```
gnprice, line 1448: nit: either put this case after the other two, or swap the order of the other two — that way the cases of 0, 1, and 2 children go in consecutive order (in one direction or the other)

gnprice (commented):
Thanks! The revisions look good; just two nits below.

Next I'll look at the remaining parts of the PR.

lib/widgets/content.dart
```diff
@@ -839,6 +848,101 @@ class MathBlock extends StatelessWidget {
...
+          ])
+      : Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
+          if (resolvedImageSrcUrl != null)
+            GestureDetector(
+              onTap: () => _launchUrl(context, node.hrefUrl),
+              child: RealmContentNetworkImage(
+                resolvedImageSrcUrl,
+                fit: BoxFit.cover,
+                width: 80,
+                height: 80)),
```
gnprice, line 925: As far as I can tell (I just spent a few minutes digging through the implementation of `Image` to investigate this), the `width` and `height` parameters here don't do anything beyond what wrapping this in a `Size` widget would do.

It looks like all our other `RealmContentNetworkImage` widgets take the latter approach — skipping these parameters, and relying on their parents for size — and that seems to work fine.

So let's do that. Then the rest of this widget's details, and its GestureDetector parent, can be deduplicated with the small-width case above.

lib/widgets/content.dart
```diff
@@ -839,6 +848,101 @@ class MathBlock extends StatelessWidget {
...
+class WebsitePreview extends StatelessWidget {
+  const WebsitePreview({super.key, required this.node});
+
+  final WebsitePreviewNode node;
+
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final resolvedImageSrcUrl = store.tryResolveUrl(node.imageSrcUrl);
+    final isSmallWidth = MediaQuery.sizeOf(context).width <= 576;
```
gnprice, line 860: I don't love using MediaQuery.sizeOf here — it feels like it's inevitably asking the wrong question. (Instead of the width of *the entire app's viewport*, it'd be much more to the point to specify this design in terms of *this widget*'s own width, which will have been dictated by its parent.)

But this is fine, because I think implementing this design, with the way it flips between wide and tall forms, in that cleaner way would require either (a) LayoutBuilder, which is significantly less clean in other ways, or (b) significantly more work. And the exact design of this feature isn't something I'd want us to spend a lot of time on.

test/model/content_test.dart
```diff
@@ -1029,6 +1029,86 @@ class ContentExample {
...
+        url: 'https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title.html'),
+    ]),
+    WebsitePreviewNode(
+      hrefUrl: 'https://pub-14f7b5e1308d42b69c4a46608442a50c.r2.dev/image+title.html',
+      imageSrcUrl: 'https://uploads.zulipusercontent.net/98fe2fe57d1ac641d4d84b6de2c520ff48fcf498/68747470733a2f2f7374617469632e7a756c6970636861742e636f6d2f7374617469632f696d616765732f6c6f676f2f7a756c69702d69636f6e2d313238783132382e706e67',
+      title: 'Zulip — organized team chat',
+      description: null),
+  ]);
+
+  static const websitePreviewWithoutTitleAndDescription = ContentExample(
```
gnprice, line 1093:
```suggestion
  static const websitePreviewWithoutTitleOrDescription = ContentExample(
```

test/widgets/content_test.dart
```diff
@@ -1005,6 +1005,65 @@ void main() {
...
+      await prepareContent(tester, plainContent(html),
+        wrapWithPerAccountStoreWidget: true);
+    }
+
+    testWidgets('smoke', (tester) async {
+      final url = Uri.parse(ContentExample.websitePreviewSmoke.markdown!);
+      await prepare(tester, ContentExample.websitePreviewSmoke.html);
+      tester.widget(find.byType(WebsitePreview));
+
+      await tester.tap(find.text('Zulip — organized team chat'));
```
gnprice, line 1019: One more quick check we can add to the two of these that have a description: check `find.text` finds something with the description's text.

(Otherwise e.g. this test case and the `websitePreviewWithoutDescription` one below are making exactly the same set of checks, I think, which seems incomplete since there are expected differences in behavior.)

test/widgets/content_test.dart
```diff
@@ -1005,6 +1005,65 @@ void main() {
...
+    });
+
+    testWidgets('smoke: without title and description', (tester) async {
+      final url = Uri.parse(ContentExample.websitePreviewWithoutTitleAndDescription.markdown!);
+      await prepare(tester, ContentExample.websitePreviewWithoutTitleAndDescription.html);
+      tester.widget(find.byType(WebsitePreview));
+
+      await tester.tap(find.byType(RealmContentNetworkImage));
+      check(testBinding.takeLaunchUrlCalls())
+        .single.equals((url: url, mode: LaunchMode.platformDefault));
```
gnprice, line 1062: The `find.byType(WebsitePreview)` check can be left out of these, I think — we're checking that there's an image which if you tap it opens the example's URL, and that seems like good confirmation in itself that we didn't just not manage to render the UI for this feature.

The advantage of leaving it out is that it's one fewer thing that has to be changed if we make a refactor to this code (for example even just renaming the widget class), one that isn't meant to affect the UI the user sees.

gnprice (commented):
OK, and finished reading the rest of the PR — comments below.

lib/widgets/content.dart
```diff
@@ -839,6 +848,101 @@ class MathBlock extends StatelessWidget {
...
+      crossAxisAlignment: CrossAxisAlignment.start,
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        if (node.title != null)
+          GestureDetector(
+            onTap: () => _launchUrl(context, node.hrefUrl),
+            child: Text(node.title!,
+              style: TextStyle(
+                fontSize: 1.2 * kBaseFontSize,
+                height: kTextHeightNone,
```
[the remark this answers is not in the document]
rajveermalviya: Web uses `line-height: normal` for this. [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal) for it says that it can be roughly `1.2`. Which seems correct, atleast on Firefox desktop and Firefox mobile for Android:

https://github.com/user-attachments/assets/f61b0eaf-ea02-4724-a10f-80554d23d7f3

So, updated to use that value.

rajveermalviya:

lib/model/content.dart
```diff
@@ -1339,6 +1391,112 @@ class _ZulipContentParser {
...
+            // if website HTML has neither title (derived from
+            // `og:title` or `<title>…</title>`) nor description (derived from
+            // `og:description`).
+            break;
+
+          default:
+            return null;
+        }
+
+        return WebsitePreviewNode(
```
gnprice, line 1486: This needs to pass `debugHtmlNode` — I noticed that while doing some debugging to follow up on https://github.com/zulip/zulip-flutter/pull/1049#discussion_r1964440200 :slightly_smiling_face: 

(The issue turned out to be that I'd done only half of the edits I suggested in the backreference comment above. Should make sure we support debugging, though.)

lib/widgets/content.dart
```diff
@@ -839,6 +849,103 @@ class MathBlock extends StatelessWidget {
...
+          crossAxisAlignment: CrossAxisAlignment.start,
+          spacing: 15,
+          children: [
+            if (image != null)
+              SizedBox(height: 110, width: double.infinity, child: image),
+            clippedTitleAndDescription,
+          ])
+      : Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
+          if (image != null)
+            AspectRatio(aspectRatio: 1, child: image),
```
gnprice, line 926: How about an 80x80 SizedBox instead? That's simpler to think about, and I think it ends up doing the same thing here.

(If it didn't do the same thing, that would probably be an issue with the AspectRatio version — I think we want the fixed size.)

gnprice (commented):
Thanks! Generally this all looks good — a few small comments.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

Issue #945, opened 2024-09-17
Retry storm on fetchOlder in message list


gnprice:
@alexmv [reported](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20retry.20storm/near/1942930) the app on his device was sending a rapid series of [getMessages](https://chat.zulip.org/api/get-messages) requests to the server — around 6 such requests a second.

Because the server rate-limits requests (by default to 200 per minute for each user, total across all types of requests), the user-facing effect was that the requests all failed, with the symptom that scrolling up further in the message list didn't work. The requests also put load on the server, though mitigated by the rate-limiting.

We should fix this.

### Diagnosis

From looking at the code, what's happening is:
* There's no overt retry logic for that fetch.
* But: the thing that triggers that fetch is a listener for the scroll state. It can get called frequently by Flutter's scrolling logic; it's supposed to be lightweight to call repeatedly.

  So naturally if we already have one of these fetches in flight, that listener promptly returns. Otherwise if you're near the top of the data we have, and we haven't heard from the server that that's the beginning of history, we make a new [/api/get-messages](https://chat.zulip.org/api/get-messages) request to fetch more.

  And when the fetches are already failing, that amounts to a retry with no backoff.

### Implementation

To fix the issue, we'll add backoff to `MessageListView.fetchOlder`, which is the method that invokes the API request.

When `fetchOlder` gets an error from the server, it should start a backoff timer (using our `BackoffMachine` so the intervals escalate) and ignore any further calls until the backoff expires, just like it ignores them when `fetchingOlder` is true.

We won't add any new retry logic. The `fetchOlder` call site, in `_MessageListState._handleScrollMetrics`, already provides plenty of retrying — all the user has to do is touch the screen to scroll slightly up or down, and `fetchOlder` will get called again. As Alex saw, it can also get called even with no interaction when nothing else seems to be happening.

### Out of scope

A further bonus refinement would be for us to understand 429 responses, and wait for the time that that response says:
* #946

But to really do that right probably belongs at a different layer, in `ApiConnection`, so that the information is shared across all the different requests we make on behalf of a given account.

And conversely I think if we fix the rapid retries, that will largely solve the problem. I suspect the reason we ended up with 429s here in the first place was likely because of too-rapid retries after some other source of error. Even if not, exponential backoff should let us get out of the situation.

PR #1050, opened 2024-11-06
msglist: Throttle fetchOlder retries

PIG208:
This approach is different from how a BackoffMachine is typically used,
because the message list doesn't send and retry requests in a loop; its
caller retries rapidly on scroll changes, and we want to ignore the
excessive requests.

The test drops irrelevant requests with `connection.takeRequests`
without checking, as we are only interested in verifying that no request
was sent.

Fixes: #945

### DIFF
--- a/lib/api/backoff.dart
+++ b/lib/api/backoff.dart
@@ -36,6 +36,25 @@ class BackoffMachine {
   /// not the (random) previous wait duration itself.
   static const double base = 2;
 
+  /// In debug mode, overrides the duration of the backoff wait.
+  ///
+  /// Outside of debug mode, this is always `null` and the setter has no effect.
+  static Duration? get debugDuration {
+    Duration? result;
+    assert(() {
+      result = _debugDuration;
+      return true;
+    }());
+    return result;
+  }
+  static Duration? _debugDuration;
+  static set debugDuration(Duration? newValue) {
+    assert(() {
+      _debugDuration = newValue;
+      return true;
+    }());
+  }
+
   /// A future that resolves after an appropriate backoff time,
   /// with jitter applied to capped exponential growth.
   ///
@@ -66,8 +85,8 @@ class BackoffMachine {
   Future<void> wait() async {
     final bound = _minDuration(maxBound,
                                firstBound * pow(base, _waitsCompleted));
-    final duration = _maxDuration(const Duration(microseconds: 1),
-                                  bound * Random().nextDouble());
+    final duration = debugDuration ?? _maxDuration(const Duration(microseconds: 1),
+                                                   bound * Random().nextDouble());
     await Future<void>.delayed(duration);
     _waitsCompleted++;
   }
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -288,6 +288,7 @@ mixin _MessageSequence {
 
   /// Update [items] to include markers at start and end as appropriate.
   void _updateEndMarkers() {
+    assert(fetched);
     assert(!(haveOldest && fetchingOlder));
     final startMarker = switch ((fetchingOlder, haveOldest)) {
       (true, _) => const MessageListLoadingItem(MessageListDirection.older),
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -1,6 +1,9 @@
+import 'dart:async';
+
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 
+import '../api/backoff.dart';
 import '../api/model/events.dart';
 import '../api/model/model.dart';
 import '../api/route/messages.dart';
@@ -89,8 +92,31 @@ mixin _MessageSequence {
   bool _haveOldest = false;
 
   /// Whether we are currently fetching the next batch of older messages.
+  ///
+  /// When this is true, [fetchOlder] is a no-op.
+  /// That method is called frequently by Flutter's scrolling logic,
+  /// and this field helps us avoid spamming the same request just to get
+  /// the same response each time.
+  ///
+  /// See also [fetchOlderCoolingDown].
   bool get fetchingOlder => _fetchingOlder;
   bool _fetchingOlder = false;
+
+  /// Whether [fetchOlder] had a request error recently.
+  ///
+  /// When this is true, [fetchOlder] is a no-op.
+  /// That method is called frequently by Flutter's scrolling logic,
+  /// and this field mitigates spamming the same request and getting
+  /// the same error each time.
+  ///
+  /// "Recently" is decided by a [BackoffMachine] that resets
+  /// when a [fetchOlder] request succeeds.
+  ///
+  /// See also [fetchingOlder].
+  bool get fetchOlderCoolingDown => _fetchOlderCoolingDown;
+  bool _fetchOlderCoolingDown = false;
+
+  BackoffMachine? _fetchOlderCooldownBackoffMachine;
 
   /// The parsed message contents, as a list parallel to [messages].
   ///
@@ -107,7 +133,7 @@ mixin _MessageSequence {
   /// before, between, or after the messages.
   ///
   /// This information is completely derived from [messages] and
-  /// the flags [haveOldest] and [fetchingOlder].
+  /// the flags [haveOldest], [fetchingOlder] and [fetchOlderCoolingDown].
   /// It exists as an optimization, to memoize that computation.
   final QueueList<MessageListItem> items = QueueList();
 
@@ -241,6 +267,8 @@ mixin _MessageSequence {
     _fetched = false;
     _haveOldest = false;
     _fetchingOlder = false;
+    _fetchOlderCoolingDown = false;
+    _fetchOlderCooldownBackoffMachine = null;
     contents.clear();
     items.clear();
   }
@@ -289,8 +317,10 @@ mixin _MessageSequence {
   /// Update [items] to include markers at start and end as appropriate.
   void _updateEndMarkers() {
     assert(fetched);
-    assert(!(haveOldest && fetchingOlder));
-    final startMarker = switch ((fetchingOlder, haveOldest)) {
+    assert(!(fetchingOlder && fetchOlderCoolingDown));
+    final effectiveFetchingOlder = fetchingOlder || fetchOlderCoolingDown;
+    assert(!(effectiveFetchingOlder && haveOldest));
+    final startMarker = switch ((effectiveFetchingOlder, haveOldest)) {
       (true, _) => const MessageListLoadingItem(MessageListDirection.older),
       (_, true) => const MessageListHistoryStartItem(),
       (_,    _) => null,
@@ -470,7 +500,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Future<void> fetchInitial() async {
     // TODO(#80): fetch from anchor firstUnread, instead of newest
     // TODO(#82): fetch from a given message ID as anchor
-    assert(!fetched && !haveOldest && !fetchingOlder);
+    assert(!fetched && !haveOldest && !fetchingOlder && !fetchOlderCoolingDown);
     assert(messages.isEmpty && contents.isEmpty);
     // TODO schedule all this in another isolate
     final generation = this.generation;
@@ -498,20 +528,28 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Future<void> fetchOlder() async {
     if (haveOldest) return;
     if (fetchingOlder) return;
+    if (fetchOlderCoolingDown) return;
     assert(fetched);
     assert(messages.isNotEmpty);
     _fetchingOlder = true;
     _updateEndMarkers();
     notifyListeners();
     final generation = this.generation;
+    bool hasFetchError = false;
     try {
-      final result = await getMessages(store.connection,
-        narrow: narrow.apiEncode(),
-        anchor: NumericAnchor(messages[0].id),
-        includeAnchor: false,
-        numBefore: kMessageListFetchBatchSize,
-        numAfter: 0,
-      );
+      final GetMessagesResult result;
+      try {
+        result = await getMessages(store.connection,
+          narrow: narrow.apiEncode(),
+          anchor: NumericAnchor(messages[0].id),
+          includeAnchor: false,
+          numBefore: kMessageListFetchBatchSize,
+          numAfter: 0,
+        );
+      } catch (e) {
+        hasFetchError = true;
+        rethrow;
+      }
       if (this.generation > generation) return;
 
       if (result.messages.isNotEmpty
@@ -532,6 +570,19 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     } finally {
       if (this.generation == generation) {
         _fetchingOlder = false;
+        if (hasFetchError) {
+          assert(!fetchOlderCoolingDown);
+          _fetchOlderCoolingDown = true;
+          unawaited((_fetchOlderCooldownBackoffMachine ??= BackoffMachine())
+            .wait().then((_) {
+              if (this.generation != generation) return;
+              _fetchOlderCoolingDown = false;
+              _updateEndMarkers();
+              notifyListeners();
+            }));
+        } else {
+          _fetchOlderCooldownBackoffMachine = null;
+        }
         _updateEndMarkers();
         notifyListeners();
       }
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -3,6 +3,8 @@ import 'dart:convert';
 import 'package:checks/checks.dart';
 import 'package:http/http.dart' as http;
 import 'package:test/scaffolding.dart';
+import 'package:zulip/api/backoff.dart';
+import 'package:zulip/api/exception.dart';
 import 'package:zulip/api/model/events.dart';
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/model/narrow.dart';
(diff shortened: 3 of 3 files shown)


chrisbobbe:
(Rerunning CI.)

chrisbobbe (commented):
Thanks! Good to get this bug fixed. Comments below.

Also, this change breaks an invariant we had before: whenever the top of the message list is scrolled into view, we would show a "start marker", which was either a loading indicator or some text saying there aren't any older messages to load.

We currently only show the loading indicator when `fetchingOlder` is true, and that's false when we're in the new "cooldown" period after a failed fetch-older request. How about also showing the loading indicator during the cooldown period? When a request is failing over and over, the cooldown period will quickly become multiple seconds long, with `fetchingOlder` flickering to `true` in between for perhaps milliseconds at a time; possibly not long enough for a frame.

Then, either here or as a followup, we could give the UI an "error"/"problem" state that's distinct from the loading state. If the request has failed, and especially if it's failed several times, that's an important sign that maybe the user should stop expecting it to succeed. Maybe `MessageListLoadingItem` could get another param like `bool problem`, for which we pass `true` when the `BackoffMachine.waitsCompleted` exceeds a certain value, like 4. And in that case make the loading indicator look like

image

or something, instead of just

image


.

PIG208:
(Working on another update to address a potential race; don't review yet)

PIG208:
The PR has been updated with the proposed changes, and additionally, a check for `this.generation == generation` to avoid potential races after adding `_updateEndMarkers` and `notifyListeners` to the backoff callback.

I have also moved the backoff machine variable, renamed to `_fetchOlderCooldownBackoffMachine` with `Cooldown` as a noun, to `_MessageSequence` right next to `fetchOlderCoolingDown`, because they are relevant to each other. The backoff machine does not have a public getter and we keep it private.

The loading indicator change might be out of scope for this PR, and we should work on that as a follow-up.

chrisbobbe:
Looks like CI is failing, could you take a look?

PIG208:
Updated the PR. Thanks!

lib/model/message_list.dart
```diff
@@ -530,11 +530,13 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       _insertAllMessages(0, fetchedMessages);
       _haveOldest = result.foundOldest;
     } finally {
-      if (this.generation == generation) {
-        _fetchingOlder = false;
-        _updateEndMarkers();
-        notifyListeners();
+      if (this.generation != generation) {
+        // ignore: control_flow_in_finally
+        return;
```
chrisbobbe, line 574: When disabling a lint rule (especially one that says it's meant to prevent bugs or unexpected behavior), please explain in a code comment why it's OK.

test/model/message_list_test.dart
```diff
@@ -236,6 +237,36 @@ void main() {
...
+    await prepareMessages(foundOldest: false, messages: initialMessages);
+
+    connection.prepare(httpStatus: 400, json: {
+      'result': 'error', 'code': 'BAD_REQUEST', 'msg': 'Bad request'});
+    check(async.pendingTimers).isEmpty();
+    await check(model.fetchOlder()).throws<ZulipApiException>();
+    checkNotified(count: 2);
+    check(model).fetchOlderCoolingDown.isTrue();
+
+    connection.takeRequests();
```
chrisbobbe, line 253:
```
msglist: Throttle fetchOlder retries

This approach is different from how a BackoffMachine is typically used,
because the message list doesn't send and retry requests in a loop; its
caller retries rapidly on scroll changes, and we want to ignore the
excessive requests.

The test drops irrelevant requests with `connection.takeRequests`
without checking, as we are only interested in verifying that no request
was sent.

Fixes: https://github.com/zulip/zulip-flutter/issues/945

```

The paragraph about the tests is a pretty small implementation detail that I think doesn't need a special mention in the commit message. On the other hand, it seems maybe helpful as an implementation comment directly on the code it's about.

test/model/message_list_test.dart
```diff
@@ -983,6 +1014,42 @@ void main() {
...
+
+        connection.prepare(delay: const Duration(seconds: 2), json: newestResult(
+          foundOldest: false,
+          messages: initialMessages + movedMessages,
+        ).toJson());
+        await store.handleEvent(eg.updateMessageEventMoveTo(
+          origTopic: movedMessages[0].topic,
+          origStreamId: otherStream.streamId,
+          newMessages: movedMessages,
+        ));
```
chrisbobbe, line 1036: We currently respond to this event by calling `_reset`, but there's a comment suggesting we might stop doing that in some cases:

```dart
  void _messagesMovedIntoNarrow() {
    // If there are some messages we don't have in [MessageStore], and they
    // occur later than the messages we have here, then we just have to
    // re-fetch from scratch.  That's always valid, so just do that always.
    // TODO in cases where we do have data to do better, do better.
    _reset();
    notifyListeners();
    fetchInitial();
  }
```

The test is meant to exercise the `_reset` case (as it says in its description); can we make it fail if it doesn't actually exercise that? Maybe:

```dart
        // check that _reset was called
        check(model).fetched.isFalse();
```

chrisbobbe (commented):
Thanks! Small comments below.

PIG208:
Thanks for the review! This has been updated.

lib/model/message_list.dart
```diff
@@ -528,12 +567,29 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
-      if (this.generation == generation) {
-        _fetchingOlder = false;
-        _updateEndMarkers();
-        notifyListeners();
+      if (this.generation != generation) {
+        // We need the finally block always clean up regardless of errors
+        // occured in the try block, and returning early here is necessary
+        // if such cleanup must be skipped, as the fetch is considered stale.
+        // ignore: control_flow_in_finally
+        return;
```
chrisbobbe, line 577:
```dart
      if (this.generation != generation) {
        // We need the finally block always clean up regardless of errors
        // occured in the try block, and returning early here is necessary
        // if such cleanup must be skipped, as the fetch is considered stale.
        // ignore: control_flow_in_finally
        return;
```

This explanation doesn't make sense to me. The cleanup "always" needs to be done…but there are times when it "must be skipped"? Your reasoning might be correct but I don't understand it yet.

chrisbobbe (commented):
Thanks! One comment below.

lib/model/message_list.dart
chrisbobbe: As a reminder, the task is to explain why the problems flagged by `control_flow_in_finally` either don't exist or are acceptable.

PIG208: How about:
```
        // This lint rule is more helpful for flagging confusing uses of
        // control flow keywords, such as `break` in a block nested in a loop,
        // or `return` a value in a finally block when a value would have been
        // returned in the `try` block.  Returning early to skip the cleanup
        // is none of these.
        // ignore: control_flow_in_finally
```

I was also tempted to refactor the if block into `if (this.generation == generation)` instead of having an early return, but the code will become more nested.

chrisbobbe: Ooh, `if (this.generation == generation)` might actually be nicer. I guess it does basically the same thing, but without making the reader think through a lint rule and whether it's helpful here or not. I'd like to here @gnprice's thoughts; I'll mark for his review.

chrisbobbe:
Marking for Greg's review; Greg, please see in particular a question above: https://github.com/zulip/zulip-flutter/pull/1050#discussion_r1877185303

lib/model/message_list.dart
gnprice: Yeah. The version in `main` uses `if (this.generation == generation)` here instead of our usual early return, and I think this lint rule was the main reason I went for that.

One extra layer of nesting is pretty tolerable, particularly as the indentation size is just 2 spaces.

lib/model/message_list.dart
```diff
@@ -528,9 +567,21 @@ class MessageListView with ChangeNotifier, _MessageSequence {
 
       _insertAllMessages(0, fetchedMessages);
       _haveOldest = result.foundOldest;
+      _fetchOlderCooldownBackoffMachine = null;
```
gnprice, line 570: nit: let's put this line that ends a backoff series right next to the logic that starts a backoff — as an `else` to the `if (hasFetchError)`.

I believe that's equivalent to this way. And that way it's somewhat easier to see how this whole cooldown state machine operates.

lib/model/message_list.dart
```diff
@@ -288,11 +316,14 @@ mixin _MessageSequence {
...
     assert(!(haveOldest && fetchingOlder));
-    final startMarker = switch ((fetchingOlder, haveOldest)) {
-      (true, _) => const MessageListLoadingItem(MessageListDirection.older),
-      (_, true) => const MessageListHistoryStartItem(),
-      (_,    _) => null,
+    assert(!(fetchingOlder && fetchOlderCoolingDown));
+    final startMarker = switch ((fetchingOlder, haveOldest, fetchOlderCoolingDown)) {
+      (true, _, _) => const MessageListLoadingItem(MessageListDirection.older),
+      (_, true, _) => const MessageListHistoryStartItem(),
+      (_, _, true) => const MessageListLoadingItem(MessageListDirection.older),
```
gnprice, line 325: nit: the fact these two fetchOlder-related conditions produce the same result isn't a coincidence; so let's handle them together.

… Oh hmm but is the point that the order of these cases is significant? Is it possible to have `haveOldest` and `fetchOlderCoolingDown` both true?

If that is possible then I guess let's make that explicit. In the existing version in main, the `assert(!(haveOldest && fetchingOlder))` is there precisely to reassure the reader that the order of the existing two cases *doesn't* matter.

gnprice: I believe in fact it's the case in this revision that `haveOldest` and `fetchOlderCoolingDown` can't both be true. Which is good — it seems like that'd be a confused state to be in.

So let's (a) assert that, and (b) combine the two fetchOlder-related cases here. Could have a line like
```dart
  final effectiveFetchingOlder = fetchingOlder || fetchOlderCoolingDown;
```
and then the same `switch` cases as in main.

test/model/message_list_test.dart
```diff
@@ -238,6 +239,42 @@ void main() {
...
+    await prepareMessages(foundOldest: false, messages: initialMessages);
+
+    connection.prepare(httpStatus: 400, json: {
+      'result': 'error', 'code': 'BAD_REQUEST', 'msg': 'Bad request'});
+    check(async.pendingTimers).isEmpty();
+    await check(model.fetchOlder()).throws<ZulipApiException>();
+    checkNotified(count: 2);
+    check(model).fetchOlderCoolingDown.isTrue();
+
+    // Drop irrelevant requests with without checking,
```
gnprice, line 255: nit:
```suggestion
    // Drop irrelevant requests without checking,
```

test/model/message_list_test.dart
```diff
@@ -238,6 +239,42 @@ void main() {
...
+    // as we are only interested in verifying if any request was sent.
+    connection.takeRequests();
```
gnprice, line 257: This line and its comment are a bit puzzling. We're "interested in verifying if any request was sent" — but does this verify that? It looks like it just discards that information.

test/model/message_list_test.dart
```diff
@@ -238,6 +239,42 @@ void main() {
...
+
+    await model.fetchOlder();
+    checkNotNotified();
+    check(model).fetchOlderCoolingDown.isTrue();
+    check(model).fetchingOlder.isFalse();
+    check(connection.lastRequest).isNull();
+
+    // The first backoff is expected to be short enough to complete.
+    async.elapse(const Duration(seconds: 1));
```
gnprice, line 266:
```suggestion
    // Wait long enough that a first backoff is sure to finish.
    async.elapse(const Duration(seconds: 1));
```

"short enough to complete" sounded to me like saying it's possible to complete it. But the point here is that this specific line definitely does complete it.

test/model/message_list_test.dart
```diff
@@ -238,6 +239,42 @@ void main() {
...
+    check(model).fetchOlderCoolingDown.isFalse();
+    checkNotifiedOnce();
+    check(connection.lastRequest).isNull();
+
+    connection.prepare(json: olderResult(
+      anchor: 1000, foundOldest: false, messages: olderMessages).toJson());
+    await model.fetchOlder();
+    checkNotified(count: 2);
+    check(connection.lastRequest).isNotNull();
+  }));
```
gnprice, line 276: I see, this is really the line that comment is most about.

It's fine to not inspect the details of the request. Let's include a check on `model.messages.length`, though — that gives the reader of the test a nice quick legible confirmation that we ended up getting all the messages, and therefore must have made the usual request and handled the response in the usual way.

test/model/message_list_test.dart
```diff
@@ -985,6 +1022,45 @@ void main() {
         checkNotifiedOnce();
       }));
 
+      test('fetchOlder backoff start, _reset, fetchOlder backoff ends, move fetch finishes', () => awaitFakeAsync((async) async {
+        await prepareNarrow(narrow, initialMessages);
+
+        connection.prepare(httpStatus: 400, json: {
+          'result': 'error', 'code': 'BAD_REQUEST', 'msg': 'Bad request'});
+        await check(model.fetchOlder()).throws<ZulipApiException>();
+        final backoffTimer = async.pendingTimers.single;
```
gnprice, line 1031: Hmm, clever!

test/model/message_list_test.dart
```diff
@@ -985,6 +1022,45 @@ void main() {
...
+        check(model).fetched.isTrue();
+        await store.handleEvent(eg.updateMessageEventMoveTo(
+          origTopic: movedMessages[0].topic,
+          origStreamId: otherStream.streamId,
+          newMessages: movedMessages,
+        ));
+        // check that _reset was caleed
+        check(model).fetched.isFalse();
+        check(model).fetchOlderCoolingDown.isFalse();
+        check(async.pendingTimers).contains(backoffTimer);
```
gnprice, line 1049:
```suggestion
        check(backoffTimer.isActive()).isTrue();
```

test/model/message_list_test.dart
```diff
@@ -985,6 +1022,45 @@ void main() {
...
+        checkHasMessages([]);
+        checkNotifiedOnce();
+
+        // The first backoff is expected to be short enough to complete.
+        async.elapse(const Duration(seconds: 1));
+        check(model).fetchOlderCoolingDown.isFalse();
+        check(async.pendingTimers).not((x) => x.contains(backoffTimer));
+        checkNotNotified();
```
gnprice, line 1057: When writing a test case, ideally we think through what scenarios would really stress the logic, and in particular would show clearly buggy symptoms if we had various plausible bugs.

So for example in the existing test just above this one, we let the fetch complete, and then check the model has no messages in its list:
```dart
        await fetchFuture;
        checkHasMessages([]);
        checkNotNotified();
```
(and then call `checkNotNotified`, mainly to ensure that `checkInvariants` gets called). For that test case that's good because if the `fetchOlder` call finished and then carried on, forgetting to check `generation`, then the symptom would be that it puts the fetched messages into the list.

Here, the current revision of this test checks that `fetchOlderCoolingDown` is still false. But that's not very informative, because although false is the value that flag had before this step, it's also exactly what the code under test would set the flag to if it *did* suffer from the race.

So to make a version of this test that really stresses the logic as it should, we should arrange a scenario where by the time the backoff completes, the flag has been set to true. Then we get to check here that it's still true — which confirms we don't have the bug where the code cheerfully sets it to false heedless of the new generation.

test/model/message_list_test.dart
```diff
@@ -985,6 +1022,45 @@ void main() {
         checkNotifiedOnce();
       }));
 
+      test('fetchOlder backoff start, _reset, fetchOlder backoff ends, move fetch finishes', () => awaitFakeAsync((async) async {
```
gnprice, line 1025: nit: let's put this after both the existing "fetchOlder, _reset" test cases, instead of between the two of them. Those two are very similar and I feel like I want to read them together, comparing them.

gnprice (commented):
Thanks to you both! Here's the rest of a full review.

chrisbobbe:
> Then, either here or as a followup, we could give the UI an "error"/"problem" state that's distinct from the loading state. If the request has failed, and especially if it's failed several times, that's an important sign that maybe the user should stop expecting it to succeed.

(Filed #1126 for this.)

test/model/message_list_test.dart
PIG208: Made a new test, "fetchOlder backoff A starts, _reset, move fetch finishes, fetchOlder backoff B starts, fetchOlder backoff A ends", for this. This also needed a refactor to make it easier to test with `BackoffMachine`, by offering a way to override the wait duration.

lib/model/message_list.dart
```diff
@@ -531,6 +570,19 @@ class MessageListView with ChangeNotifier, _MessageSequence {
     } finally {
       if (this.generation == generation) {
         _fetchingOlder = false;
+        if (hasFetchError) {
+          assert(!fetchOlderCoolingDown);
+          _fetchOlderCoolingDown = true;
+          unawaited((_fetchOlderCooldownBackoffMachine ??= BackoffMachine())
+            .wait().then((_) {
+              if (this.generation != generation) return;
```
gnprice, line 578: nit:
```suggestion
              if (this.generation > generation) return;
```

(matching the similar checks elsewhere)

test/model/message_list_test.dart
```diff
@@ -238,6 +240,40 @@ void main() {
...
+    check(model).fetchOlderCoolingDown.isFalse();
+    checkNotifiedOnce();
+    check(connection.lastRequest).isNull();
+
+    connection.prepare(json: olderResult(
+      anchor: 1000, foundOldest: false, messages: olderMessages).toJson());
+    await model.fetchOlder();
+    checkNotified(count: 2);
+    check(connection.takeRequests()).single;
+  }));
```
gnprice, line 275: bump https://github.com/zulip/zulip-flutter/pull/1050#discussion_r1877411417

test/model/message_list_test.dart
```diff
@@ -1020,6 +1056,70 @@ void main() {
...
+        check(backoffTimerB.isActive).isTrue();
+        checkNotNotified();
+
+        // When `backoffTimerB` ends, `fetchOlderCoolingDown` gets reset.
+        async.elapse(const Duration(seconds: 1));
+        check(model).fetchOlderCoolingDown.isFalse();
+        check(backoffTimerA.isActive).isFalse();
+        check(backoffTimerB.isActive).isFalse();
+        checkNotifiedOnce();
+      }));
```
gnprice, line 1121: This new test looks good!

lib/api/backoff.dart
```diff
@@ -66,8 +85,8 @@ class BackoffMachine {
   Future<void> wait() async {
     final bound = _minDuration(maxBound,
                                firstBound * pow(base, _waitsCompleted));
-    final duration = _maxDuration(const Duration(microseconds: 1),
-                                  bound * Random().nextDouble());
+    final duration = debugDuration ?? _maxDuration(const Duration(microseconds: 1),
+                                                   bound * Random().nextDouble());
```
gnprice, line 89: nit: line too long (and in particular the value of `microseconds` is past 80 columns, just)

gnprice (commented):
Thanks for the revision! All looks good now except three small comments.

PIG208:

gnprice:
Thanks! Looks good — rebased, and will watch CI.

PR #1055, opened 2024-11-10
compose box max height changed from static to 30% of screen height

whodivyanshu:


whodivyanshu:
I changed the screen max height to 30% of the screen height

chrisbobbe:
I see you've replaced this PR with #1056. In future, when you have changes to make, please update the existing PR instead of replacing it with a new one.

PR #1056, opened 2024-11-10
changed the compose box max height to 20% of the screen height

whodivyanshu:


chrisbobbe:
Hi @whodivyanshu, thanks for your PR. The compose box is being redesigned, and its max height will be set according to a design from our designer. PR #928 is our implementation of that redesign, so I'll close this PR. Thanks again, though.

PR #1057, opened 2024-11-11
deps: Upgrade Flutter; and handle migrations

rajveermalviya:
deps: Upgrade Flutter to 3.27.0-1.0.pre.443

---

deps: Update CocoaPods to 1.16.2

Changelog:
  https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md#1162-2024-10-31

Upstream changed the recommended CocoaPods version to 1.16.2 in:
  https://github.com/flutter/flutter/commit/dff191625f213a7cfdd1731d6f7b695e638ebfd2

Without this change `flutter run` generates the following warnings:

```
Warning: CocoaPods recommended version 1.16.2 or greater not installed.
Pods handling may fail on some projects involving plugins.
To update CocoaPods, see https://guides.cocoapods.org/using/getting-started.html#updating-cocoapods
```
---

intl: Migrate off of synthetic package; check-in generated files

`flutter` tool deprecated the `package:flutter_gen` synthetic package:
  https://flutter.dev/to/flutter-gen-deprecation

So, instead migrate to referencing source files directly.

Without this change, `flutter run` generates the following warning:

```
Synthetic package output (package:flutter_gen) is deprecated: https://flutter.dev/to/flutter-gen-deprecation. In a future release, synthetic-package will default to `false` and will later be removed entirely.
```


chrisbobbe (changes requested):
Great!

If I understand correctly, the "migrate off synthetic package" commit fixes an issue I run into pretty often, where `flutter_gen` gets removed from `.dart_tool/package_config.json` which causes analysis failures everywhere we use localization, and I have to fix by running `flutter pub get`.

We have a chunk of code in `tools/check` to work around that issue:

```bash
    # The `flutter pub run` disrupts `.dart_tool/package_config.json`,
    # removing `flutter_gen`, which causes analysis failures everywhere
    # that we use localization.  (And `dart run` has the same problem.)
    # Fix it back, with `flutter pub get`.
    # TODO(upstream): Find or file a bug for this `pub run` vs. `flutter_gen` issue.
    if [ -n "${opt_verbose}" ]; then
        flutter pub get \
        || return
    else
        flutter pub get \
            >/dev/null \
        || return
    fi
```

And IIUC we can just delete that code now. :tada:.

`git grep flutter_gen` also finds an instruction in docs/translation.md that should be updated.

l10n.yaml
```diff
@@ -1,7 +1,9 @@
 # Docs on this config file:
 #   https://docs.flutter.dev/ui/accessibility-and-localization/internationalization#configuring-the-l10nyaml-file
 
+synthetic-package: false
 arb-dir: assets/l10n
+output-dir: lib/generated/l10n
```
gnprice, line 6: Neat! Glad to see this fixed.

In addition to the points Chris mentioned above, let's add this directory to `.gitattributes` so that `git log -p` and so on don't become noisy; see the handling there of other generated files.

gnprice:
Merging the first two commits now, to have the Flutter upgrade in when I make a new beta release shortly.

@rajveermalviya please go ahead and send the last commit:
6a495027d intl: Migrate off of synthetic package; check-in generated files

as a follow-up PR, after handling the comments above. Thanks again!

gnprice:
(This means we'll have that warning appear on `flutter run` for the moment:
```
Synthetic package output (package:flutter_gen) is deprecated:
https://flutter.dev/to/flutter-gen-deprecation. In a future release, synthetic-package will default
to `false` and will later be removed entirely.
```
but that's tolerable, and we'll take care of it soon.)

rajveermalviya:
> send the last commit … as a follow-up PR

@gnprice Sent #1058.

PR #1058, opened 2024-11-13
intl: Migrate from `package:flutter_gen`; include generated files

rajveermalviya:
`flutter` tool deprecated the `package:flutter_gen` synthetic package:
  https://flutter.dev/to/flutter-gen-deprecation

The generated localization source files are now included in the
repository and referenced directly.

Without this change, `flutter run` generates the following warning:
```
Synthetic package output (package:flutter_gen) is deprecated:
https://flutter.dev/to/flutter-gen-deprecation. In a future
release, synthetic-package will default to `false` and will
later be removed entirely.
```

### DIFF
--- a/docs/translation.md
+++ b/docs/translation.md
@@ -56,7 +56,7 @@ changes to the ARB file.
 
 To use in our widgets, you need to import the generated bindings:
 ```
-import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
+import '../generated/l10n/zulip_localizations.dart';
 ```
 
 Then in your widget code, pull the localizations object
--- a/docs/translation.md
+++ b/docs/translation.md
@@ -54,12 +54,7 @@ changes to the ARB file.
 
 ### Using a translated string in the code
 
-To use in our widgets, you need to import the generated bindings:
-```
-import '../generated/l10n/zulip_localizations.dart';
-```
-
-Then in your widget code, pull the localizations object
+To use in your widget code, pull the localizations object
 off of the Flutter build context:
 ```
 Widget build(BuildContext context) {
--- a/lib/model/localizations.dart
+++ b/lib/model/localizations.dart
@@ -1,4 +1,4 @@
-import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
+import '../generated/l10n/zulip_localizations.dart';
 
 abstract final class GlobalLocalizations {
   static ZulipLocalizations zulipLocalizations =
--- a/tools/check
+++ b/tools/check
@@ -334,20 +334,6 @@ run_pigeon() {
         > >(filter_flutter_pub_run_output) \
     || return
 
-    # The `flutter pub run` disrupts `.dart_tool/package_config.json`,
-    # removing `flutter_gen`, which causes analysis failures everywhere
-    # that we use localization.  (And `dart run` has the same problem.)
-    # Fix it back, with `flutter pub get`.
-    # TODO(upstream): Find or file a bug for this `pub run` vs. `flutter_gen` issue.
-    if [ -n "${opt_verbose}" ]; then
-        flutter pub get \
-        || return
-    else
-        flutter pub get \
-            >/dev/null \
-        || return
-    fi
-
     check_no_changes "changes to pigeons" "${outputs[@]}"
 }
 
--- a/tools/check
+++ b/tools/check
@@ -26,7 +26,7 @@ this_dir=${BASH_SOURCE[0]%/*}
 
 default_suites=(
     analyze test
-    build_runner drift pigeon icons
+    build_runner l10n drift pigeon icons
     android  # This takes multiple minutes in CI, so do it last.
 )
 
@@ -82,7 +82,7 @@ while (( $# )); do
         --all) opt_files=all; opt_all=1; shift;;
         --fix) opt_fix=1; shift;;
         --verbose) opt_verbose=1; shift;;
-        analyze|test|build_runner|drift|pigeon|icons|android|shellcheck)
+        analyze|test|build_runner|l10n|drift|pigeon|icons|android|shellcheck)
             opt_suites+=("$1"); shift;;
         *) usage;;
     esac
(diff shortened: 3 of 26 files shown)


lib/model/localizations.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
+import '/generated/l10n/zulip_localizations.dart';
```
PIG208, line 1: Should we use relative path like we do in the other files, or is it needed for other reasons?

tools/check
```diff
@@ -241,6 +241,19 @@ should_run_build_runner() {
     return 1
 }
 
+run_l10n() {
+    files_check lib/generated/l10n/'*'.dart \
```
PIG208, line 245: nit: like

    local schema_dir=test/model/schemas/

we can extract a local variable for the output path

PIG208 (commented):
Thanks for the PR. Looks good to me overall! Left some small comments.

rajveermalviya:
Thanks for the review @PIG208! Pushed a new revision.

CI is failing because Flutter upstream changed the generated header for localization files in https://github.com/flutter/flutter/commit/dca37ad17ff0354924f9b5bbbcbb2ce6ed819230, resulting in following diff:
```diff
diff --git a/lib/generated/l10n/zulip_localizations_ar.dart b/lib/generated/l10n/zulip_localizations_ar.dart
index fedc8b74..b7c1ed6d 100644
--- a/lib/generated/l10n/zulip_localizations_ar.dart
+++ b/lib/generated/l10n/zulip_localizations_ar.dart
@@ -1,5 +1,5 @@
+// ignore: unused_import
 import 'package:intl/intl.dart' as intl;
-
 import 'zulip_localizations.dart';
 
 // ignore_for_file: type=lint
…
…
```

Added a head commit to update Flutter, and regenerated the files, that should fix the CI now.

docs/translation.md
```diff
@@ -54,12 +54,8 @@ changes to the ARB file.
 
 ### Using a translated string in the code
 
-To use in our widgets, you need to import the generated bindings:
-```
-import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
```
gnprice, line 59: super nit: the main migration commit should update this line, because otherwise the instruction is incorrect as of that commit

(then the later commit can delete, like in this revision, along with that nice commit message explaining why it used to be needed and no longer is)

tools/check
```diff
@@ -241,6 +241,21 @@ should_run_build_runner() {
     return 1
 }
 
+run_l10n() {
```
gnprice, line 244: nit: move this above `should_run_build_runner`, which is a helper function specific to `run_build_runner`, so as not to split those from each other

(alternatively, move this suite after the `build_runner` suite)

tools/check
```diff
@@ -241,6 +241,21 @@ should_run_build_runner() {
...
+    local output_path=lib/generated/l10n
+
+    files_check "${output_path}"/'*'.dart \
```
gnprice, line 247: nit: pull the whole pathspec `lib/generated/l10n/'*'.dart` out as a local variable

That way there's a natural place for a comment clarifying that this string is meant as a Git pathspec. Otherwise it looks an awful lot like a shell glob (which it would be if the `*` weren't escaped with those single quotes), which would have somewhat different semantics. See `run_pigeon` for an example (though that example also makes it an array, which isn't needed here since there's just one pathspec).

gnprice, line 247: I think this check is too narrow and will miss some cases where this suite should run.

For examples, look back at the other `files_check` call sites.

This should also include an "Omitted from this check" comment like the ones at those other call sites, listing all the other files in the repo that would belong in a 100% version of this check (because changes to them could affect whether this suite passes) but which we're choosing to leave out as described at the `files_check` doc comment. I put those there partly as a way of making sure I explicitly thought through what files could matter to each check, to try to avoid this class of bug :slightly_smiling_face:

gnprice (commented):
Thanks @rajveermalviya for taking care of this! And thanks @PIG208 for the previous review (and @chrisbobbe for the review on the predecessor PR #1057).

Generally this all looks great; a few comments below. All nits except there's one logic error in the new `tools/check l10n` suite.

rajveermalviya:

tools/check
```diff
@@ -277,6 +277,23 @@ run_build_runner() {
     check_no_changes "updates to *.g.dart files" '*.g.dart'
 }
 
+run_l10n() {
+    local output_files=lib/generated/l10n/zulip_localizations'*'.dart
+
+    # Omitted from this check:
+    #   pubspec.{yaml,lock} tools/check
+    files_check assets/l10n/ "${output_files}" \
```
gnprice, line 285: What about `l10n.yaml`?

tools/check
```diff
@@ -277,6 +277,23 @@ run_build_runner() {
     check_no_changes "updates to *.g.dart files" '*.g.dart'
 }
 
+run_l10n() {
+    local output_files=lib/generated/l10n/zulip_localizations'*'.dart
```
gnprice, line 281: nit: I don't love calling this "files" — it makes it sound like it's an actual path (or list of actual paths) to some files. As mentioned at https://github.com/zulip/zulip-flutter/pull/1058#discussion_r1847653173 , I'd like to make explicit that this is instead a Git pathspec.

This would do it:
```suggestion
    local output_pathspec=lib/generated/l10n/zulip_localizations'*'.dart
```

docs/translation.md
gnprice:
```diff
-import 'package:flutter_gen/gen_l10n/zulip_localizations.dart';
+import '/generated/l10n/zulip_localizations.dart';
```

If this nit were the only thing remaining in this revision I'd probably just go ahead and merge it, but since we have another round anyway: this (temporary) advice should match what we actually do in the source files :slightly_smiling_face:, which follows https://github.com/zulip/zulip-flutter/pull/1058#discussion_r1844485771 .

gnprice (commented):

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

gnprice:
Hmmm, the commits I merged earlier today looked like this:
ae8aabc46 deps: Upgrade Flutter to 3.27.0-1.0.pre.518
661adfc7b intl: Migrate from `package:flutter_gen`; include generated files
8293e8172 (TODO fix links) intl docs: Remove instruction to import bindings file
c6a296026 check: Remove workaround for `package:flutter_gen`
d03ec1e32 check: Add suite l10n

Spot the error :slightly_smiling_face: 

Nothing else has been merged since then, and it looks like there aren't even any PRs that have been updated since this was merged. So I think it's not too late to fix it; I'll go and do so.

gnprice:
OK, done. Now the corresponding commits in main are:
ae8aabc46 deps: Upgrade Flutter to 3.27.0-1.0.pre.518
661adfc7b intl: Migrate from `package:flutter_gen`; include generated files
c15bc77ee intl docs: Remove instruction to import bindings file
8148aa47f check: Remove workaround for `package:flutter_gen`
c6c2f33f8 check: Add suite l10n

(If it were just the broken links themselves, I'd leave them there; the reader who wants to follow them always could, just with a small amount of annoyance. But the big capital-letters "TODO" right in the summary line would have kept bugging me into the future when looking at the history. :slightly_smiling_face:)

PR #1059, opened 2024-11-13
internal_link: Always include a "/" after hostname

PIG208:
Meanwhile, reorganized some test groups with the same name.

Supersedes: #874
Fixes: #845

### DIFF
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -96,7 +96,14 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
     fragment.write('/near/$nearMessageId');
   }
 
-  return store.realmUrl.replace(fragment: fragment.toString());
+  Uri result = store.realmUrl.replace(fragment: fragment.toString());
+  if (result.path.isEmpty) {
+    // Always ensure that there is a '/' right after the hostname.
+    // A generated URL without '/' looks odd,
+    // and if used in a Zulip message does not get automatically linkified.
+    result = result.replace(path: '/');
+  }
+  return result;
 }
 
 /// A [Narrow] from a given URL, on `store`'s realm.
--- a/test/model/compose_test.dart
+++ b/test/model/compose_test.dart
@@ -1,8 +1,6 @@
 import 'package:checks/checks.dart';
 import 'package:test/scaffolding.dart';
 import 'package:zulip/model/compose.dart';
-import 'package:zulip/model/narrow.dart';
-import 'package:zulip/model/internal_link.dart';
 
 import '../example_data.dart' as eg;
 import 'test_store.dart';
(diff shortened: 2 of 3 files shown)


test/model/compose_test.dart
```diff
@@ -314,6 +314,28 @@ hello
         '#narrow/dm/1,2-dm/near/12345',
         '#narrow/pm-with/1,2-pm/near/12345');
     });
+
+    test('normalize links to always include a "/" after hostname', () {
+      void checkGeneratedLink({required String realmUrl, required String expectedLink}) {
+        final account = eg.selfAccount.copyWith(realmUrl: Uri.parse(realmUrl));
+        final store = eg.store(account: account);
+        check(narrowLink(store, const CombinedFeedNarrow()))
+          .equals(Uri.parse(expectedLink));
```
chrisbobbe, line 323: I'm slightly inclined to stringify the `narrowLink` result and compare that to the `String expectedLink`. The user's experience of a `narrowLink` result is always through `Uri.toString`. (Quote-and-reply and copy-link-to-message.)

chrisbobbe (commented):
Thanks! Small comments below.

lib/model/internal_link.dart
```diff
@@ -96,7 +96,14 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
...
   }
 
-  return store.realmUrl.replace(fragment: fragment.toString());
+  Uri result = store.realmUrl.replace(fragment: fragment.toString());
+  if (result.path.isEmpty) {
+    // Always ensure that there is a '/' right after the hostname.
+    // A generated URL without '/' looks odd and does not linkify.
+    result = result.replace(path: '/');
+  }
+  assert(result.path.startsWith('/'));
```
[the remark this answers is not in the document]
PIG208: Makes sense. This feels like a duplication of the assertion made in the comments above.

chrisbobbe:

test/model/internal_link_test.dart
```diff
@@ -184,6 +184,36 @@ void main() {
       testExpectedNarrows(testCases, streams: streams);
     });
 
+    group('topic link parsing', () {
```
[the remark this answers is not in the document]
gnprice: Or in fact let's just leave these where they are entirely — it looks like you don't end up making any changes to this test file in the main commit, so it doesn't seem like in this PR we're making use of any energy we spend loading these tests into our heads enough to refactor them.

lib/model/internal_link.dart
```diff
@@ -96,7 +96,13 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
     fragment.write('/near/$nearMessageId');
   }
 
-  return store.realmUrl.replace(fragment: fragment.toString());
+  Uri result = store.realmUrl.replace(fragment: fragment.toString());
+  if (result.path.isEmpty) {
+    // Always ensure that there is a '/' right after the hostname.
+    // A generated URL without '/' looks odd and does not linkify.
```
gnprice, line 102:
```suggestion
    // A generated URL without '/' looks odd,
    // and if used in a Zulip message does not get automatically linkified.
```

The not getting linkified is a specific fact about what Zulip does with URLs it spots in the plain text of a message; it's not a general fact about URLs, and this function isn't specific to generating something to go into a Zulip message.

test/model/compose_test.dart
```diff
@@ -314,6 +314,28 @@ hello
         '#narrow/dm/1,2-dm/near/12345',
         '#narrow/pm-with/1,2-pm/near/12345');
     });
+
+    test('normalize links to always include a "/" after hostname', () {
```
gnprice, line 318: The code this is testing is in `internal_link.dart`, not `compose.dart`, so let's place the test to match. We can do that with a prep commit moving this `narrowLink` test group to the top of `internal_link_test.dart`; really that should have happened as part of 2f944a0125f67aed2df47a02e1f254ac76710179.

test/model/compose_test.dart
```diff
@@ -314,6 +314,28 @@ hello
...
+      void checkGeneratedLink({required String realmUrl, required String expectedLink}) {
+        final account = eg.selfAccount.copyWith(realmUrl: Uri.parse(realmUrl));
+        final store = eg.store(account: account);
```
gnprice, line 321: nit: just make a fresh Account from scratch:
```suggestion
        final store = eg.store(account: eg.account(realmUrl: Uri.parse(realmUrl)));
```

Or is there a reason that doesn't work?

test/model/compose_test.dart
```diff
@@ -314,6 +314,28 @@ hello
...
+        check(narrowLink(store, const CombinedFeedNarrow()).toString())
+          .equals(expectedLink);
+      }
+
+      checkGeneratedLink(
+        realmUrl:     'http://chat.example.com',
+        expectedLink: 'http://chat.example.com/#narrow');
```
gnprice, line 328: nit: always best to make the code transparent about what it's doing, when that doesn't come at a high cost in concision:
```suggestion
      check(generatedLink('http://chat.example.com'))
        .equals(          'http://chat.example.com/#narrow');
```

(And here it actually improves concision, too).

Then further can rename `generatedLink` to something a bit more specific, like `narrowLinkFor`.

test/model/internal_link_test.dart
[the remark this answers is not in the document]
gnprice: … I guess that last remark is made moot by my other comment that the new tests should go in this file.

I'll push a revision with a new commit that fixes these group-name collisions more minimally. We're still not in this PR doing anything with these tests of `parseInternalLink` — the code being modified isn't `parseInternalLink`, but rather `narrowLink` which is roughly its inverse — so we don't need to think too hard about these tests at the moment.

test/model/compose_test.dart
PIG208: It would be:
```dart
      final store = eg.store(account: eg.account(user: eg.selfUser, realmUrl: Uri.parse(realmUrl)));
```
which is fine and actually as short as the other way:
```dart
      final store = eg.store(account: eg.selfAccount.copyWith(realmUrl: Uri.parse(realmUrl)));
```

PIG208:
Thanks for the commit and reviews! I have updated the PR.

test/model/internal_link_test.dart
```diff
@@ -35,6 +35,116 @@ Future<PerAccountStore> setupStore({
...
+        '#narrow/dm/1,2,3,4-group',
+        '#narrow/pm-with/1,2,3,4-group');
+      checkNarrow(allRecipientIds: [1, 2], selfUserId: 1, nearMessageId: 12345,
+        '#narrow/dm/1,2-dm/near/12345',
+        '#narrow/pm-with/1,2-pm/near/12345');
+    });
+
+    test('normalize links to always include a "/" after hostname', () {
+      String narrowLinkFor({required String realmUrl}) {
+        final store = eg.store(account: eg.selfAccount.copyWith(realmUrl: Uri.parse(realmUrl)));
```
gnprice, line 133: bump https://github.com/zulip/zulip-flutter/pull/1059#discussion_r1868495067

gnprice (commented):
Thanks for the revision! Just one surviving comment.

Oh and in a commit message:

5a1caacc3 internal_link test [nfc]: Give test groups distinct names
5acef65ff internal_link test [nfc]: Re-home narrow link tests
1ff2621e7 internal_links: Always include a "/" after hostname

There's no identifier `internal_links`, so that identifier-styled term shouldn't appear in text when there isn't an identifier it refers to.

test/model/compose_test.dart
gnprice: ... Oh, I didn't see this reply when writing https://github.com/zulip/zulip-flutter/pull/1059#discussion_r1881120699 . I think I got got by the GitHub bug where a PR page often/usually live-updates with new comments, but then often doesn't, so I was looking at the page (in an existing tab) and didn't see any reply.

Anyway, if it needs that extra detail, no need to change it. And from your revision I see there's an existing test there that sets this up the same way.

PIG208: I guess the non-copy-with implementation would be a tiny bit more realistic because we supposedly (there is an TODO on `eg.account`) will regenerate a bunch of fields from scratch as well.

gnprice: Hmm true. Not a bunch, I don't think; but the account ID, and maybe the API key too. Sure.

gnprice:
Thanks! Merging, after the subthread at https://github.com/zulip/zulip-flutter/pull/1059#discussion_r1881141813 .

PR #1061, opened 2024-11-14
added the preview content functionality for the compose box 

whodivyanshu:


lib/widgets/compose_box.dart
```diff
@@ -1032,25 +1048,190 @@ class _ComposeBoxLayout extends StatelessWidget {
...
+                text: _buildStyledText(content, context),
+              ),
+      ),
+    );
+  }
+
+  InlineSpan _buildStyledText(String text, BuildContext context) {
+    final boldPattern = RegExp(r'\*\*(.*?)\*\*');
+    final italicPattern = RegExp(r'\*(.*?)\*');
+    final codePattern = RegExp(r'```(.*?)```', dotAll: true); // Match ```code```
```
PIG208: Hmm. I don't think parsing the message content with regular expressions is the right approach. This will eventually leads to the complexity of re-implementing Zulip's markdown processor from scratch, if we want to have full support of all message content features. A correct implementation should use the [render-message](https://zulip.com/api/render-message) endpoint, and reuse Zulip's Content parser.

gnprice:
As discussed above, this approach isn't going to work, so I'll close this PR.

@whodivyanshu if you'd like to contribute to Zulip, I recommend finding a more approachable issue to work on.

Issue #946, opened 2024-09-17
Understand rate-limiting responses (429, RATE_LIMIT_HIT)


gnprice:
The Zulip server rate-limits requests. By default the limit for each user is 200 per minute, in total across all types of requests. For an example where this came up, see #945 and [its original report](https://chat.zulip.org/#narrow/stream/48-mobile/topic/Flutter.20retry.20storm/near/1942930).

We'll resolve #945 by having fetchOlder specifically add backoff for the getMessages requests we make there. But then a further bonus refinement would be for us to understand when the server is specifically giving us rate-limiting errors, and follow its instructions for how long to wait before making another request.

For docs on the rate-limiting errors:
* Partial documentation is here: https://zulip.com/api/rest-error-handling#rate-limit-exceeded
* Some other aspects appear here: https://zulip.com/api/http-headers#rate-limiting-response-headers
* Though we don't seem to have this part documented, the HTTP status code is 429, which is standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

To do this right, this logic belongs on ApiConnection. That way the information can be shared across all the different requests we make on behalf of a given account.

PR #1063, opened 2024-11-16
Comprehensively retry event-poll failures

### LINKED ISSUE #563: Comprehensively retry poll/register even on unforeseen errors
This is a followup to:
* #556 
* #184 
* #185 

Hopefully with #556 (fixed by #561) we're done with needing to add try/catch blocks and retries for conditions that can happen for purely operational reasons like the device going to sleep. But there might be another such condition we've missed; and there's also always the possibility of bugs in our code. In particular the event-poll loop calls PerAccountStore.handleEvent, which can reach quite a lot of our model code because of the variety of different events, so it's probable that at some point we'll have a bug that causes that to throw an exception.

For the bulk of our code, if a bug causes an exception, then the best thing to do is to let it propagate out of our code uncaught — it'll mean that that widget doesn't get built (causing a gray rectangle in release builds), or that user gesture doesn't get acted on, but the user can pick up from there and try again. But the event loop, or `UpdateMachine`, is different: it's the one spot in our code where we have a persistent thread of control flow of our own. It needs to keep itself going no matter what happens, because nothing else will resume it for us.

I have a rough draft for this, which I wrote while working on #556/#561. I put it aside in order to get #561 out the door promptly, but I'll plan to pick it up later. The draft also comes with some user feedback, a form of #555.

Related issues, some of which might end up getting handled at the same time as this one:
* #465 
* #555 
* #186 
* #554

gnprice:
This is a bit of a long branch, as I ended up going into a variety of related cleanups and other prep work.

Fixes: #563 

Selected commit messages:

---

store: On rate-limit error in poll, explicitly back off and retry

This covers a particular case of #946.

This commit is actually NFC as far as the interaction with the server
goes; its only change in behavior is that we treat the error as
"expected", and therefore skip reporting it to the user unless it
follows a string of other errors.

But then this also sets us up so that these rate-limit errors will
continue to be handled with backoff-retry when we change the handling
of unexpected errors, coming next.

---

store: Handle unexpected event-poll errors by reloading store, with backoff

This fixes #563.  We'll follow up with a few more commits that give
more-informative errors in some cases, or change the handling of others
from retrying getEvents to reloading the data from scratch.

The if-disposed and store.isLoading lines have no effect in the case
of a BAD_EVENT_QUEUE_ID error, because those can only come from the
getEvents request, and then the inner catch block will have already
taken the same steps.

Fixes: #563

---

store: Better error message on handleEvent failure

---

store: On non-transient request errors, reload rather than retry

This hopefully gives us a greater chance of getting past whatever
the underlying problem is, by resetting more of our state.

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -1002,119 +1002,130 @@ class UpdateMachine {
 
   void poll() async {
     assert(!_disposed);
+    try {
+      BackoffMachine? backoffMachine;
+      int accumulatedTransientFailureCount = 0;
 
-    BackoffMachine? backoffMachine;
-    int accumulatedTransientFailureCount = 0;
-
-    /// This only reports transient errors after reaching
-    /// a pre-defined threshold of retries.
-    void maybeReportToUserTransientError(Object error) {
-      accumulatedTransientFailureCount++;
-      if (accumulatedTransientFailureCount > transientFailureCountNotifyThreshold) {
-        _reportToUserErrorConnectingToServer(error);
-      }
-    }
-
-    while (true) {
-      if (_debugLoopSignal != null) {
-        await _debugLoopSignal!.future;
-        if (_disposed) return;
-        assert(() {
-          _debugLoopSignal = Completer();
-          return true;
-        }());
+      /// This only reports transient errors after reaching
+      /// a pre-defined threshold of retries.
+      void maybeReportToUserTransientError(Object error) {
+        accumulatedTransientFailureCount++;
+        if (accumulatedTransientFailureCount > transientFailureCountNotifyThreshold) {
+          _reportToUserErrorConnectingToServer(error);
+        }
       }
 
-      final GetEventsResult result;
-      try {
-        result = await getEvents(store.connection,
-          queueId: queueId, lastEventId: lastEventId);
-        if (_disposed) return;
-      } catch (e) {
-        if (_disposed) return;
-
-        store.isLoading = true;
-        bool isUnexpected;
-        bool shouldReportToUser;
-        switch (e) {
-          case ZulipApiException(code: 'BAD_EVENT_QUEUE_ID'):
-            assert(debugLog('Lost event queue for $store.  Replacing…'));
-            // This disposes the store, which disposes this update machine.
-            await store._globalStore._reloadPerAccount(store.accountId);
-            assert(debugLog('… Event queue replaced.'));
-            return;
-
-          case NetworkException(cause: SocketException()):
-            // A [SocketException] is common when the app returns from sleep.
-            isUnexpected = false;
-            shouldReportToUser = false;
-
-          case NetworkException():
-          case Server5xxException():
-            isUnexpected = false;
-            shouldReportToUser = true;
-
-          case ServerException(httpStatus: 429):
-          case ZulipApiException(httpStatus: 429):
-          case ZulipApiException(code: 'RATE_LIMIT_HIT'):
-            // TODO(#946) handle rate-limit errors more generally, in ApiConnection
-            isUnexpected = false;
-            shouldReportToUser = true;
-
-          default:
-            isUnexpected = true;
-            shouldReportToUser = true;
+      while (true) {
+        if (_debugLoopSignal != null) {
+          await _debugLoopSignal!.future;
+          if (_disposed) return;
+          assert(() {
+            _debugLoopSignal = Completer();
+            return true;
+          }());
         }
 
-        if (isUnexpected) {
-          assert(shouldReportToUser);
-          assert(debugLog('Error polling event queue for $store: $e\n'
-              'Backing off and retrying even though may be hopeless…'));
-          // TODO(#186): Handle unrecoverable failures
-          _reportToUserErrorConnectingToServer(e);
-        } else {
-          assert(debugLog('Transient error polling event queue for $store: $e\n'
-              'Backing off, then will retry…'));
-          if (shouldReportToUser) {
-            maybeReportToUserTransientError(e);
+        final GetEventsResult result;
+        try {
+          result = await getEvents(store.connection,
+            queueId: queueId, lastEventId: lastEventId);
+          if (_disposed) return;
+        } catch (e) {
+          if (_disposed) return;
+
+          store.isLoading = true;
+          bool isUnexpected;
+          bool shouldReportToUser;
+          switch (e) {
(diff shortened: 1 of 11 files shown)


test/model/store_test.dart
```diff
@@ -732,7 +732,12 @@ void main() {
       });
     }
 
-    test('retries on NetworkException', () {
+    test('retries on NetworkException from SocketException', () {
+      // We skip reporting errors on these; check we retry them all the same.
```
PIG208, line 787: I see that later we add a test to verify that we indeed ignore such errors, maintaining some sort of "each error that we retries/reloads on gets a corresponding 'report error' test" relationship. But there is this visual gap between each pair of them. Is it feasible to place them next to each other?

lib/model/store.dart
```diff
@@ -1113,13 +1125,41 @@ class UpdateMachine {
...
+          // Hopefully (probably?) the bug only affects our implementation of
+          // the *change* in state represented by the event, and when we get the
+          // new state in a fresh InitialSnapshot we'll handle that just fine.
+          isUnexpected = true;
+      }
+
+      if (isUnexpected) {
+        // We don't know the cause of the failure; it might well keep happening.
+        // Avoid creating a retry storm.
+        await _unexpectedErrorBackoffMachine.wait();
```
PIG208, line 1180: I see that you mentioned
```
The if-disposed and store.isLoading lines have no effect in the case
of a BAD_EVENT_QUEUE_ID error, because those can only come from the
getEvents request, and then the inner catch block will have already
taken the same steps.
```
and for errors other than `BAD_EVENT_QUEUE_ID` where `isUnexpected` is `true`, do we want to do an `if (_disposed)` check right after this?

lib/model/store.dart
```diff
@@ -1044,48 +1044,46 @@ class UpdateMachine {
           if (_disposed) return;
         } catch (e) {
           if (_disposed) return;
-
           store.isLoading = true;
-          bool isUnexpected;
+
+          if (e is! ApiRequestException) {
+            // Some unexpected error, outside even making the HTTP request.
+            // Definitely a bug in our code.
```
PIG208, line 1051: Looks like this check also enables the exhaustive switch later. That's pretty neat!

PIG208:
Thanks for cleaning this up! Looks good to me. Left some comments after reading through the PR.

test/model/store_test.dart
gnprice: Hmm, this is meant to coincide with one of the error types that in main we already test that we skip reporting:
```dart
      test('ignore boring errors', () => awaitFakeAsync((async) async {
        await prepare();

        for (int i = 0; i < UpdateMachine.transientFailureCountNotifyThreshold; i++) {
          connection.prepare(exception: const SocketException('failed'));
          pollAndFail(async);
          check(takeLastReportedError()).isNull();
```

lib/model/store.dart
gnprice: Yeah, good catch, thanks. This is an `await`, so it should be followed by an if-disposed check.

lib/model/store.dart
gnprice: Yep!

Ideally I'd have liked to express this as one more case in the switch, so that the `rethrow` cases were all next to each other. But the patterns minilanguage doesn't seem to have a way to express that. (For example if it had negations of patterns, that would do it.) So this works instead.

gnprice:
Thanks for the review! Pushed a revision with those fixes. See also the question at https://github.com/zulip/zulip-flutter/pull/1063#discussion_r1847674444 .

test/model/store_test.dart
PIG208: Yeah, this is quite a nit. I was referring to the visual gap between this test and
```dart
      test('ignore NetworkException from SocketException', () {
        checkNotReported(prepareNetworkExceptionSocketException);
      });
```
or more in general, the gap between the 'report error` group and the tests before it. We have 10 such pairs:

```dart
    test('reloads on unexpected error within loop', () {
      checkReload(prepareUnexpectedLoopError);
    });

	// ...

    group('report error', () {

	  // ...

      test('report unexpected error within loop', () {
        checkReported(prepareUnexpectedLoopError);
      });
```

I'm not sure if it would be better to group the tests for the same error close to each other, or keep it this way. It does seem that such tests tend to co-occur.

test/model/store_test.dart
```diff
@@ -719,6 +719,30 @@ void main() {
       connection.prepare(httpStatus: 200, body: 'nonsense');
     }
 
+    void prepareRateLimitExceptionCode() {
+      // Example from the Zulip API docs:
+      //   https://zulip.com/api/rest-error-handling#rate-limit-exceeded
+      // (The actual HTTP status should be 429, but that seems undocumented.)
+      connection.prepare(httpStatus: 400, json: {
+        'result': 'error', 'code': 'RATE_LIMIT_HIT',
+        'msg': 'API usage exceeded rate limit,',
```
PIG208, line 738: nit:

```suggestion
        'msg': 'API usage exceeded rate limit',
```

From the API doc linked here.
```
{
    "code": "RATE_LIMIT_HIT",
    "msg": "API usage exceeded rate limit",
    "result": "error",
    "retry-after": 28.706807374954224
}
```

PIG208 (commented):
Thanks for the update! This looks good to me overall. Left a new comment on something I missed last time and a follow-up to https://github.com/zulip/zulip-flutter/pull/1063#discussion_r1847356424.

test/model/store_test.dart
gnprice: Indeed that's a typo, thanks :slightly_smiling_face:

test/model/store_test.dart
gnprice: I see, yeah.

There's a fundamental trade-off to make: these test cases are really a matrix, of (which type of error) x (what to do vs. what to report). So we can group the different checks on a given error type together, but then the different error types get spread farther apart, or vice versa.

I went for this arrangement — different error types together for each kind of check — because I think when considering what we want to do (or report) for a given error type it's helpful to be able to compare across other error types. With this arrangement, all the reload/retry tests fit on the screen at once, or a couple of screenfuls on a smaller display.

PIG208:
The update looks good to me. Thanks!

test/model/store_test.dart
PIG208: That's a great way to put it! Thanks for the explanation.

PR #1069, opened 2024-11-21
Emoji autocomplete

gnprice:
This adds autocomplete for emoji in the compose box, so e.g. typing `:zu` offers an autocomplete option of `:zulip:`.

The first part of the branch makes some refactors to the autocomplete subsystem, in order to prepare to accommodate emoji autocomplete alongside @-mention autocomplete. The changes are all in basically the direction we had in mind when initially building the system; mostly we left some concepts collapsed that we knew we'd eventually need to separate, deferring the details of actually separating them until we had a concrete situation where they came apart.

Fixes: #669 
Fixes: #670

### Selected commit messages

autocomplete [nfc]: Document how query, view-model, and results classes relate

---

autocomplete [nfc]: Separate ComposeAutocompleteQuery/Result from Mention-etc.

The two concepts have meant the same set of possible values so far,
because @-mentions are the only type of autocomplete we've had so far
in the compose box's content input.  As a result we've taken some
shortcuts by conflating them.

But as we introduce other types of autocomplete in the content input,
like for emoji and #-mentions, we have some places that will need to
refer to the more general concept while others refer to the more
specific one.  So separate them out.

---

emoji [nfc]: Factor out ImageEmojiWidget

---

emoji [nfc]: Factor out UnicodeEmojiWidget

---

emoji: Make list of emoji to consider for autocomplete or emoji picker

This leaves the emojiDisplay field of these objects untested.  I
skipped that because it seems like pretty boring low-risk code,
just invoking emojiDisplayFor.  (And emojiDisplayFor has its own
tests.)  But included a TODO comment for completeness in thinking
about what logic there is to test here.

Fixes: #669

---

autocomplete: Identify when the user intends an emoji autocomplete

The "forbid preceding ':'" wrinkle is one I discovered because of
writing tests: I wrote down the '::^' test, was surprised to find
that it failed, and then went back and extended the regexp to
make it pass.

[…]

---

emoji: Finish emoji autocomplete for compose box

Fixes: #670

### DIFF
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -184,37 +184,42 @@ class AutocompleteViewManager {
 /// and the [AutocompleteResult] subclass in `ResultT` describes the
 /// possible results that the user might choose in the autocomplete interaction.
 ///
+/// When an [AutocompleteView] is created, its constructor begins the search
+/// for results corresponding to the initial [query].
+/// The query may later be updated, causing a new search.
+///
 /// The owner of one of these objects must call [dispose] when the object
 /// will no longer be used, in order to free resources on the [PerAccountStore].
 ///
 /// Lifecycle:
-///  * Create an instance of a concrete subtype.
+///  * Create an instance of a concrete subtype, beginning a search
 ///  * Add listeners with [addListener].
-///  * Use the [query] setter to start a search for a query.
+///  * When the user edits the query, use the [query] setter to update the search.
 ///  * On reassemble, call [reassemble].
 ///  * When the object will no longer be used, call [dispose] to free
 ///    resources on the [PerAccountStore].
 abstract class AutocompleteView<QueryT extends AutocompleteQuery, ResultT extends AutocompleteResult> extends ChangeNotifier {
-  AutocompleteView({required this.store});
+  /// Construct a view-model for an autocomplete interaction,
+  /// and begin the search for the initial query.
+  AutocompleteView({required this.store, required QueryT query})
+      : _query = query {
+    _startSearch();
+  }
 
   final PerAccountStore store;
 
-  QueryT? get query => _query;
-  QueryT? _query;
-  set query(QueryT? query) {
+  QueryT get query => _query;
+  QueryT _query;
+  set query(QueryT query) {
     _query = query;
-    if (query != null) {
-      _startSearch();
-    }
+    _startSearch();
   }
 
   /// Called when the app is reassembled during debugging, e.g. for hot reload.
   ///
   /// This will redo the search from scratch for the current query, if any.
   void reassemble() {
-    if (_query != null) {
-      _startSearch();
-    }
+    _startSearch();
   }
 
   Iterable<ResultT> get results => _results;
@@ -271,8 +276,7 @@ abstract class AutocompleteView<QueryT extends AutocompleteQuery, ResultT extend
     required Iterable<T> candidates,
     required List<ResultT> results,
   }) async {
-    assert(_query != null);
-    final query = _query!;
+    final query = _query;
 
     final iterator = candidates.iterator;
     outer: while (true) {
@@ -295,6 +299,7 @@ abstract class AutocompleteView<QueryT extends AutocompleteQuery, ResultT extend
 class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery, MentionAutocompleteResult> {
   MentionAutocompleteView._({
     required super.store,
+    required super.query,
     required this.narrow,
     required this.sortedUsers,
   });
@@ -302,9 +307,11 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
   factory MentionAutocompleteView.init({
     required PerAccountStore store,
     required Narrow narrow,
+    required MentionAutocompleteQuery query,
   }) {
     final view = MentionAutocompleteView._(
       store: store,
+      query: query,
       narrow: narrow,
       sortedUsers: _usersByRelevance(store: store, narrow: narrow),
     );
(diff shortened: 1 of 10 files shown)


lib/widgets/autocomplete.dart
```diff
@@ -247,6 +247,50 @@ class _MentionAutocompleteItem extends StatelessWidget {
...
+
+    final emojiDisplay = candidate.emojiDisplay.resolve(store.userSettings);
+    final Widget? glyph = switch (emojiDisplay) {
+      ImageEmojiDisplay() =>
+        ImageEmojiWidget(size: _size, emojiDisplay: emojiDisplay),
+      UnicodeEmojiDisplay() =>
+        UnicodeEmojiWidget(
+          size: _size, notoColorEmojiTextSize: _notoColorEmojiTextSize,
+          emojiDisplay: emojiDisplay),
+      TextEmojiDisplay() => null, // The text is already shown separately.
```
chrisbobbe, line 271: It's true that the text is shown, but in a way that's more confusing than it needs to be, I think:

image

I'll tap "person_tipping_hand" if it looks appealing, but then I might be confused/unhappy to see ":information_desk_person:" written instead.

Is there a nice way to give the "emoji name" some distinct formatting? Maybe borrowing the label/sublabel design from the Figma on @-mention autocomplete:

image

where the first line says ":information_desk_person:" (or maybe without the colons) and the second says "person_tipping_hand" (plus more if the list is longer)? That same design might be helpful for unicode- and image-display emoji too (though less necessary), so the code could just cleanly apply it unconditionally. This would grow some of the items vertically (splitting what could be one line of text into two), but I think that's an OK tradeoff.

Also I could imagine someone using the "plain text" setting because they don't like seeing emoji everywhere, but they still want to see the emoji in the emoji picker, so they know how it'll look to people who don't use the setting.

lib/model/autocomplete.dart
```diff
@@ -78,6 +105,30 @@ final RegExp mentionAutocompleteMarkerRegex = (() {
...
+  // Specifically forbid a preceding ":", though, to make "::" not a query.
+  const before = r'(?<=^|\s|\p{Punctuation})(?<!:)';
+  // TODO(dart-future): Regexps in ES 2024 have a /v aka unicodeSets flag;
+  //   if Dart matches that, we could combine into one character class
+  //   meaning "whitespace and punctuation, except not `:`":
+  //     r'(?<=^|[[\s\p{Punctuation}]--[:]])'
+
+  /// Characters that might be meant as part of (a query for) an emoji's name,
+  /// other than whitespace.
+  const nameCharacters = r'_\p{Letter}\p{Number}';
```
chrisbobbe, line 119: Hmm, how do we know which characters these are? For the corresponding part in `_mentionIntentRegex` we exclude characters, not include them.

Here, if we don't include `\p{Emoji}` and I guess maybe the ZWJ (maybe more?) then we defeat the part of `EmojiAutocompleteQuery.matches` that's supposed to enable searching by literal emoji.

I don't think that feature is really useful here actually. If I want a specific emoji from the iOS emoji picker, I'll just open that directly, right, instead of typing ":" first. Searching by literal emoji seems more helpful when you're choosing an emoji reaction than when you're typing an emoji into a message.

I guess I don't mind offering the feature. 🤷‍♂️ It might help someone who wants their platform's emoji picker but has already typed ":" because they weren't sure if they needed to.

chrisbobbe (commented):
Thanks, this will be great to have!! Small comments below.

lib/model/autocomplete.dart
gnprice: Yeah, I guess this means searching by literal emoji isn't actually offered in the message-content autocomplete. Which is fine because as you say, it's not that useful here.

What that feature is really for is for adding an emoji reaction — it effectively gives the user the option of using their phone keyboard's emoji picker instead of ours.



> Hmm, how do we know which characters these should be? For the corresponding part in `_mentionIntentRegex` we exclude characters, not include them.

Yeah, which matches the fact that users' names can be almost anything, with a small list of character exclusions. Emoji names are more constrained.

But I hadn't actually looked up what those exact constraints are. Turns out it's more constrained even than I'd guessed:
```py
def check_valid_emoji_name(emoji_name: str) -> None:
    if emoji_name:
        if re.match(r"^[0-9a-z\-_]+(?<![\-_])$", emoji_name):
            return
    # [… otherwise raise one error or other.]

File: lib/widgets/autocomplete.dart [hunk 3]
```diff
@@ -247,6 +247,50 @@ class _MentionAutocompleteItem extends StatelessWidget {
...
+        if (glyph != null) ...[
+          glyph,
+          const SizedBox(width: 8),
+        ],
+        Expanded(
+          child: Text(
+            maxLines: 2,
+            overflow: TextOverflow.ellipsis,
+            label)),
+      ]));
```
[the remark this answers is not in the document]
gnprice: Yeah, that'd all be good. I think of it as basically part of merging these changes with #995. Right now this version matches the design we have for user results, and after #995 we'll want them to continue to match.

gnprice:
Thanks for the thoughtful review! Replied to some subthreads above, and just pushed a revision; PTAL.

lib/widgets/autocomplete.dart
chrisbobbe: Sure, that all sounds fine.

test/model/autocomplete_test.dart
```diff
@@ -125,61 +129,139 @@ void main() {
...
+    doTest(':-p^', null);
+
+    // Avoid interpreting as queries some ways colons appear in source code.
+    doTest('::^', null);
+    doTest(':<^', null);
+    doTest(':=^', null);
+
+    // Emoji names may have letters and numbers in various scripts.
+    // (A few appear in the server's list of Unicode emoji.
+    // many more might be in a given realm's custom emoji.)
```
chrisbobbe, line 219: nit: sentence starting with lowercase

chrisbobbe (approved):
Thanks! One nit below, then please merge at will.

gnprice:
Thanks! Fixed and merged.

Issue #892, opened 2024-08-19
Parse @-topic mentions


gnprice:
When we get a message with an @-topic mention, we should parse the mention and show it in the message list. Currently it appears as "unimplemented".

That's this Zulip feature: https://zulip.com/help/mention-a-user-or-group#mention-all-topic-participants

The appearance should be the same as other wildcard mentions (@-channel, @-stream, @-all, @-everyone).

Umbrella issue:
* #817 

#### Out of scope
* Separately we'll offer these in autocomplete, as part of (or a follow-up to): #234 
* Separately we'll recognize the corresponding per-user message flag `topic_wildcard_mentioned`.
* Later we'll display these differently from mentions of an individual user: #646 
* Later we'll highlight the mention, and the message as a whole, if it refers to the self-user: https://github.com/zulip/zulip-flutter/issues/647

Issue #646, opened 2024-05-07
content: Distinguish direct and wildcard mentions


chrisbobbe:
Currently, we don't distinguish the appearance of direct mentions and wildcard mentions:

<img width="252" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/807e7bbd-f53c-4557-a33b-b9f098c21cc0">

The web app does that—

<img width="187" alt="image" src="https://github.com/zulip/zulip-flutter/assets/22248748/3eee86bd-45af-43a4-b68f-f811feae5c09">

—and we should too, probably using the same colors web does.

I'll file a separate issue for a related but different feature: highlighting non-silent mentions of the self-user. That will take more data-model thinking. -> #647

chrisbobbe:
(Should we add this to #309?)

gnprice:
This is one of the pieces of:
* #157

in that the styling we have matches what was current in web when I started the prototype, but hasn't been updated.

Having this separate issue in the tracker is useful, though — for actually tackling #157, we'll probably want separate issues like this one anyway. And it provides an on-point answer for someone wondering why it doesn't match web. I'll edit #157 to point to this.

gnprice:
We've already done the first part of this: the mentions no longer look like the first screenshot above. Instead they always look like the blue pill in the second screenshot.

The remaining task is to have the distinct styling for group mentions.

Ruhdee:
Now that #2119 is merged, I will begin work on this issue.
I will use the web colors for user group mention and wildcard mention pills. Design discussion: [#mobile-design > mention colors](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/mention.20colors/with/2367260)
I will create `wildcardMentionNode` child class for code clarity.

PR #1071, opened 2024-11-21
 content: Handle @-topic mentions

rajveermalviya:
Fixes: #892

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -623,9 +623,13 @@ class LinkNode extends InlineContainerNode {
   }
 }
 
+sealed class MentionNode extends InlineContainerNode {
+  const MentionNode({super.debugHtmlNode, required super.nodes});
+}
+
 enum UserMentionType { user, userGroup }
 
-class UserMentionNode extends InlineContainerNode {
+class UserMentionNode extends MentionNode {
   const UserMentionNode({
     super.debugHtmlNode,
     required super.nodes,
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -646,6 +646,10 @@ class UserMentionNode extends MentionNode {
   //   final bool isSilent;
 }
 
+class TopicMentionNode extends MentionNode {
+  const TopicMentionNode({super.debugHtmlNode, required super.nodes});
+}
+
 sealed class EmojiNode extends InlineContentNode {
   const EmojiNode({super.debugHtmlNode});
 }
@@ -818,6 +822,13 @@ class _ZulipContentParser {
     return RegExp("^(?:$mentionClass(?: silent)?|silent $mentionClass)\$");
   }();
 
+  static final _topicMentionClassNameRegexp = () {
+    // This matches a class `topic-mention`, plus an optional class `silent`,
+    // appearing in either order.
+    const mentionClass = "topic-mention";
+    return RegExp("^(?:$mentionClass(?: silent)?|silent $mentionClass)\$");
+  }();
+
   static final _emojiClassNameRegexp = () {
     const specificEmoji = r"emoji(?:-[0-9a-f]+)+";
     return RegExp("^(?:emoji $specificEmoji|$specificEmoji emoji)\$");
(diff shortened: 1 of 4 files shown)


rajveermalviya:
(Rebased to main)

test/model/content_test.dart
```diff
@@ -180,6 +180,26 @@ class ContentExample {
     expectedText: 'all',
     '<p><span class="silent user-mention" data-user-id="*">all</span></p>',
     const UserMentionNode(nodes: [TextNode('all')]));
+  static final topicMentionPlain = ContentExample.inline(
```
PIG208, line 183: nit: blank line before this

PIG208 (commented):
Thanks for working on this. Left some comments. This looks good to me!

lib/model/content.dart
```diff
@@ -623,9 +623,13 @@ class LinkNode extends InlineContainerNode {
...
+  const MentionNode({super.debugHtmlNode, required super.nodes});
+}
+
 enum UserMentionType { user, userGroup }
 
-class UserMentionNode extends InlineContainerNode {
+class UserMentionNode extends MentionNode {
   const UserMentionNode({
     super.debugHtmlNode,
     required super.nodes,
```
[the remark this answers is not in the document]
rajveermalviya: Hmm, I'm not sure how the comment about silent mentions would apply to this generic `MentionNode`, since it's being introduced for `@topic` mention and not sure how a silent `@topic` would work. I feel this comment is specific to `UserMentionNode`.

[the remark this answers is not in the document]
PIG208: Oh I see! Yeah that makes sense.

PIG208:
Looks good to me! Marking this for Greg's review.

lib/model/content.dart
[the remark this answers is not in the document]
gnprice: A silent @-topic mention might not be super useful, but it does exist:
https://chat.zulip.org/#narrow/channel/7-test-here/topic/mention/near/2007548

lib/model/content.dart
```diff
@@ -717,6 +721,10 @@ class UserMentionNode extends InlineContainerNode {
   //   final bool isSilent; // TODO(#647)
 }
 
+class TopicMentionNode extends MentionNode {
```
gnprice, line 724: I think the name "topic mention" for this, with no further context, will be confusing — it sounds like it might refer to the `#**channel > topic**` feature, which is unrelated.

lib/widgets/content.dart
```diff
@@ -1117,23 +1124,26 @@ class _InlineContentBuilder {
...
   @override
   Widget build(BuildContext context) {
     final contentTheme = ContentTheme.of(context);
     return Container(
       decoration: BoxDecoration(
         // TODO(#646) different for wildcard mentions
-        color: contentTheme.colorDirectMentionBackground,
+        color: switch (node) {
+          UserMentionNode() => contentTheme.colorDirectMentionBackground,
+          TopicMentionNode() => contentTheme.colorTopicMentionBackground,
```
gnprice, line 1145: It seems like this is beginning to implement #646 — we're now giving some wildcard mentions (namely for topics) a different background vs. direct mentions.

Should @-channel mentions get the same background as @-topic?

gnprice (commented):
Thanks @rajveermalviya, and thanks @PIG208 for the previous review!

Let's actually handle this within the existing class UserMentionNode. It's meant to encompass not only mentions of individual users, but of sets of users either via user groups, the current channel, the current topic, or any other such features we might introduce in the future. And it's already how we currently handle @-channel mentions.

To distinguish @-topic mentions from mentions of individual users, we can use the `mentionType` field that's sketched in a comment, and the currently-unused  UserMentionType enum. That can happen in a separate commit, either before or after the commit that adds @-topic mentions in the first place; if we introduce @-topic mentions first (and just give them the same background as direct user mentions), their treatment isn't any worse than what we have now for @-channel.

lib/widgets/content.dart
alya: Yeah! But personal mentions trump channel and group mentions if a message has both.

Original issue for reference (we've probably tweaked colors since then): https://github.com/zulip/zulip/issues/22059
Corner case that needed fixing, in case this helps avoid an issue: https://github.com/zulip/zulip/issues/27654

gnprice: Thanks for the references!

To possibly clarify, though: the immediate code here is about how the mention is formatted within the message, rather than how the message as a whole gets highlighted when it is/has a mention. The latter is tracked under #647. I think that part will probably call for a separate PR from this one, but the formatting of the mention elements themselves might be easy to fold in here.

rajveermalviya: Updated to remove this part, will open separate PR for #647 later as it's a post-launch issue.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

Issue #1040, opened 2024-11-01
"Whole channel" button from topic narrows


gnprice:
When viewing the message list for a topic narrow, we should have a quick way to get to the enclosing stream/channel narrow.

In the legacy zulip-mobile app, this is just a button in the app bar, with an "up" icon. In a future with a fancier app bar for these message lists (as in #1039), we may do something fancier; but for now, let's just follow the legacy app and have a button.

One update from the legacy app, though: instead of an "up" icon, let's use the "all messages" / "combined feed" icon. That's the same one that web now uses in the channel menu in the left sidebar, for the new "Go to channel feed" option.

I'm giving this the near-term "Beta 4" milestone, because when I was demoing Zulip in person a lot earlier this week, I repeatedly found the lack of it awkward. (Specifically the use case was that I was viewing one conversation and wanted to start a new one. There are other ways we could serve that use case; but any deeper re-imagining of that UX is a post-launch issue.)

Dhanesh-Sawant:
Hi @gnprice 

I've implemented the "Whole channel" button within the MessageListAppBarTitle for the TopicNarrow. Here's a brief overview of the approach I've taken:

In the TopicNarrow case of the MessageListAppBarTitle component, I'm currently returning a Row that includes:
The _buildStreamRow method to display the stream and topic names.

An IconButton (currently with an "up" arrow icon) to navigate to the enclosing channel view.
The IconButton triggers Navigator.pushReplacement to navigate to the MessageListPage with a ChannelNarrow for the given streamId. This way, it directly takes the user back to the channel view without adding a new route to the navigation stack.


class MessageListAppBarTitle extends StatelessWidget {
  const MessageListAppBarTitle({super.key, required this.narrow});

  final Narrow narrow;

  Widget _buildStreamRow(BuildContext context, {
    ZulipStream? stream,
    required String text,
  }) {
    // A null [Icon.icon] makes a blank space.
    final icon = (stream != null) ? iconDataForStream(stream) : null;
    return Row(
      mainAxisSize: MainAxisSize.min,
      // TODO(design): The vertical alignment of the stream privacy icon is a bit ad hoc.
      //   For screenshots of some experiments, see:
      //     https://github.com/zulip/zulip-flutter/pull/219#discussion_r1281024746
      crossAxisAlignment: CrossAxisAlignment.center,
      children: [
        Icon(size: 16, icon),
        const SizedBox(width: 4),
        Flexible(child: Text(text)),
      ]);
  }

[...]

PR #1072, opened 2024-11-22
msglist: Add channel-feed action from topic narrows

gnprice:
Also apply our "icon" color to icon buttons more broadly.

Fixes: #1040

## Selected commit messages

theme: Apply icon color through IconButtonTheme

The most conspicuous change this makes is to the "back" icon
in an AppBar, aka AppBar.leading.  Those have been a white or
black color from Material defaults; now they get our "icon" color.

This matches what's in Figma; and also matches the action icons,
for any app bars that have both.  Currently our only app bar with
an action is also the only one that lacks a "leading": it's on
ChooseAccountPage, which is at the root of navigation so there's
nowhere for a "back" icon to lead to.  But when we add actions in
other pages, the mismatch would become conspicuous.

This change also turns a handful of less-visible icon buttons
from white/black to the icon color: the lightbox's "copy link"
or play/pause buttons, and the show/hide-password button on the
username/password login page.  I think those changes are desirable
too.  All our other icon buttons that I can find (either in the
source, or by browsing around the app) already have a color
specified, and aren't affected by this change.

---

msglist: Add channel-feed action from topic narrows

Fixes: #1040


Dhanesh-Sawant:
Hi @gnprice,

Apologies for not addressing the requested changes sooner—I was tied up with exams. I completely agree with your feedback on commit style, testing, and other improvements.

Thank you for implementing the feature in PR #1072. I'll review it to learn and improve for future contributions.

gnprice:
Cool. Hope your exams went well.

gnprice:
(Oh, I should upload screenshots for the visual changes — I already went to the trouble of taking them. Short on time this morning but I'll try to do so later today.)

chrisbobbe:
The code LGTM and so does the UI (including those changed icon colors); please merge at will, with or without posting those screenshots.

gnprice:
Thanks for the review! For expediency and since the UI changes are all small, I'll just go ahead and merge, then.

Issue #1064, opened 2024-11-18
New channel wildcard mentions are rendered as "unimplemented"


sm-sayedi:
While working on #889, I realized that the channel wildcard mentions (`@all`, `@everyone`, `@channel`, `@stream`) are rendered as "unimplemented". This behavior is only for newer messages (after [#31075](https://github.com/zulip/zulip/pull/31075) is merged in [zulip](https://github.com/zulip/zulip)), the older messages with these wildcard mentions are rendered correctly. This is how the HTML text for older and newer messages looks like:

```html
// older messages
<span class="user-mention" data-user-id="*">@all</span>

// newer messages
<span class="user-mention channel-wildcard-mention" data-user-id="*">@all</span>
```

There seems to have been an additional class (`channel-wildcard-mention`) added. So, I think we have to support both the previous format and the current one.

gnprice:
Hmm! Interesting, thanks for the report.

Can you link to an example message with this format? Or is it not currently possible to generate on chat.zulip.org for some reason?

Using the survey tool from #190, I don't see any examples of this in public messages on chat.zulip.org. See e.g. the output copied in the description of #921.

sm-sayedi:
I couldn't find any such message on chat.zulip.org and because I don't have permission to use `@all` and similar mentions, I couldn't generate one too. I spotted this in my own organization which I use for test purposes.

> Using the survey tool from #190, I don't see any examples of this in public messages on chat.zulip.org.

I am not sure, but maybe that's because there are no new messages with such mentions.

Here are the screenshots (of my organization) both from the web and mobile:

### Web
![image](https://github.com/user-attachments/assets/16130670-60b2-4dc5-bb45-5504305c5045)

### Mobile
<img src="https://github.com/user-attachments/assets/fe81d1a0-fbbd-490e-8ad4-7ccf95c3dc64" width="300" />

gnprice:
Aha, that'd do it. Yeah, we have this permission set: 
![image](https://github.com/user-attachments/assets/3e4ff3de-10ab-4fda-bc87-81e935993a78)

> Who can notify a large number of users with a wildcard mention [(i)](https://chat.zulip.org/help/restrict-wildcard-mentions)
> Admins and moderators

so in most channels, nobody can use @-all and the like except admins and moderators, and I guess in practice we never do.

@rajveermalviya would you put this issue on your queue? Looks like probably a small one.

[...]

PR #1073, opened 2024-11-22
content: Support the new class of channel wildcard mentions

sm-sayedi:
For channel wildcard mentions, the class in the corresponding HTML used to be the same as the user mentions (`class="user-mention"`), but now there's an additional class added. Now it looks like the following: `class="user-mention channel-wildcard-mention"`.

Fixes: #1064

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -881,15 +881,26 @@ class _ZulipContentParser {
     int i = 0;
 
     if (i >= classes.length) return null;
+    bool hasChannelWildcardClass = false;
+    if (classes[i] == 'channel-wildcard-mention') {
+      // Newer channel wildcard mentions have this class; older ones don't.
+      i++;
+      hasChannelWildcardClass = true;
+    }
+
+    if (i >= classes.length) return null;
     if (classes[i] == 'silent') {
       // A silent @-mention.  We ignore this flag; see [UserMentionNode].
       i++;
     }
 
     if (i >= classes.length) return null;
-    if (classes[i] == 'user-mention' || classes[i] == 'user-group-mention') {
+    if (classes[i] == 'user-mention'
+        || (classes[i] == 'user-group-mention' && !hasChannelWildcardClass)) {
       // The class we already knew we'd find before we called this function.
       // We ignore the distinction between these; see [UserMentionNode].
+      // Also, we don't expect "user-group-mention" and "channel-wildcard-mention"
+      // to be in the list at the same time.
       i++;
     }
 
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -139,6 +139,48 @@ class ContentExample {
     '<p><span class="silent user-group-mention" data-user-group-id="186">test-empty</span></p>',
     const UserMentionNode(nodes: [TextNode('test-empty')]));
 
+  static final channelWildcardMentionPlain = ContentExample.inline(
+    'plain channel wildcard @-mention',
+    "@**all**",
+    expectedText: '@all',
+    '<p><span class="user-mention channel-wildcard-mention" data-user-id="*">@all</span></p>',
+    const UserMentionNode(nodes: [TextNode('@all')]));
+
+  static final channelWildcardMentionSilent = ContentExample.inline(
+    'silent channel wildcard @-mention',
+    "@_**everyone**",
+    expectedText: 'everyone',
+    '<p><span class="user-mention channel-wildcard-mention silent" data-user-id="*">everyone</span></p>',
+    const UserMentionNode(nodes: [TextNode('everyone')]));
+
+  static final channelWildcardMentionSilentClassOrderReversed = ContentExample.inline(
+    'silent channel wildcard @-mention, class order reversed',
+    "@_**channel**", // (hypothetical server variation)
+    expectedText: 'channel',
+    '<p><span class="silent user-mention channel-wildcard-mention" data-user-id="*">channel</span></p>',
+    const UserMentionNode(nodes: [TextNode('channel')]));
+
+  static final legacyChannelWildcardMentionPlain = ContentExample.inline(
+    'legacy plain channel wildcard @-mention',
+    "@**channel**",
+    expectedText: '@channel',
+    '<p><span class="user-mention" data-user-id="*">@channel</span></p>',
+    const UserMentionNode(nodes: [TextNode('@channel')]));
+
+  static final legacyChannelWildcardMentionSilent = ContentExample.inline(
+    'legacy silent channel wildcard @-mention',
+    "@_**stream**",
+    expectedText: 'stream',
+    '<p><span class="user-mention silent" data-user-id="*">stream</span></p>',
+    const UserMentionNode(nodes: [TextNode('stream')]));
+
+  static final legacyChannelWildcardMentionSilentClassOrderReversed = ContentExample.inline(
+    'legacy silent channel wildcard @-mention, class order reversed',
+    "@_**all**", // (hypothetical server variation)
+    expectedText: 'all',
+    '<p><span class="silent user-mention" data-user-id="*">all</span></p>',
+    const UserMentionNode(nodes: [TextNode('all')]));
+
   static final emojiUnicode = ContentExample.inline(
     'Unicode emoji, encoded in span element',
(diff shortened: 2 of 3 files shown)


lib/model/content.dart
```diff
@@ -808,10 +808,12 @@ class _ZulipContentParser {
...
-    // This matches a class `user-mention` or `user-group-mention`,
-    // plus an optional class `silent`, appearing in either order.
-    const mentionClass = r"user(?:-group)?-mention";
-    return RegExp("^(?:$mentionClass(?: silent)?|silent $mentionClass)\$");
+    // This matches the classes `user-mention`, `user-group-mention`,
+    // or `user-mention channel-wildcard-mention`, plus an optional
+    // class `silent`, appearing in either order.
+    const mentionClass = r"user(?:-group)?-mention|"
+                          "(?:user-mention channel-wildcard-mention)";
+    return RegExp("^(?:$mentionClass)(?: silent)?|silent (?:$mentionClass)\$");
```
PIG208, line 816: Looks like the non-capturing groups added here are not necessary:
```suggestion
    const mentionClass = r"user(?:-group)?-mention|"
                          "user-mention channel-wildcard-mention";
    return RegExp("^(?:$mentionClass)(?: silent)?|silent $mentionClass\$");
```

PIG208: Seeing this I wondered if regular expressions are the right tool for this issue, but then I read from the commit message of 4a10b3cf8cb286396e034a830993c32544ed0726 mentioning that this is more efficient than `classes`. Makes sense.

PIG208 (commented):
Thanks! This looks good. Left some comments.

lib/model/content.dart
gnprice: The groups around `mentionClass` are needed. The revised version above would have a three-way alternative: either `^(?:$mentionClass)(?: silent)?`, or `silent user(?:-group)?-mention`, or `user-mention channel-wildcard-mention$`.

Probably the cleanest way to express that is by `mentionClass` itself containing the `(?:` and `)`.

In main there's also a group enclosing everything between `^` and `$`. That's necessary so that `^` and `$` apply to all alternatives, instead of just the first and last of them respectively.

gnprice: I've just sent #1086 to refactor the existing logic here, reducing the role of the regexp and hopefully making it easier to add more features. @sm-sayedi PTAL and try rebasing your work atop that.

sm-sayedi:
Thanks @PIG208 and @gnprice for the previous reviews. Pushed the new changes with the branch being rebased on top of #1086. PTAL.

sm-sayedi:
Thanks @PIG208 and @gnprice for the previous comments. Pushed the new changes with the branch being rebased on top of #1086. PTAL.

lib/model/content.dart
```diff
@@ -880,16 +880,28 @@ class _ZulipContentParser {
         || classes.contains('user-group-mention'));
     int i = 0;
 
+    if (i >= classes.length) return null;
+    bool isChannelWildcardClassIncluded = false;
+    if (classes[i] == 'channel-wildcard-mention') {
+      // A newly-added class for channel wildcard mentions.
+      // See: https://github.com/zulip/zulip/pull/31075
```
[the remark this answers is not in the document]
gnprice: (On what goes in a commit message vs. a comment, see: https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#commit-messages-vs-comments )

lib/model/content.dart
```diff
@@ -872,6 +870,53 @@ class _ZulipContentParser {
...
+    assert(element.localName == 'span');
+    final debugHtmlNode = kDebugMode ? element : null;
+
+    final classes = element.className.split(' ')..sort();
+    assert(classes.contains('user-mention')
+        || classes.contains('user-group-mention'));
+    int i = 0;
+
+    if (i >= classes.length) return null;
+    bool isChannelWildcardClassIncluded = false;
```
gnprice, line 884: nit: can simplify "is X included" to "has X":
```suggestion
    bool hasChannelWildcardClass = false;
```

lib/model/content.dart
```diff
@@ -872,6 +870,53 @@ class _ZulipContentParser {
...
+    if (classes[i] == 'channel-wildcard-mention') {
+      // A newly-added class for channel wildcard mentions.
+      // See: https://github.com/zulip/zulip/pull/31075
```
gnprice, line 887: I think a comment is useful here, because there is certain information it can convey that's relevant for the reader trying to understand this function. Here's a version focused on that information:
```suggestion
    if (classes[i] == 'channel-wildcard-mention') {
      // Newer channel wildcard mentions have this class; older ones don't.
```

In particular, in my other review comment above I almost suggested naming the variable `isChannelWildcard` — before I remembered that that would be an inaccurate name, because some genuine channel wildcards don't have this class. That fact will be highly relevant when someone goes to implement #646, as it'd otherwise be tempting to use this class for doing so.

lib/model/content.dart
```diff
@@ -872,6 +870,53 @@ class _ZulipContentParser {
...
+
+    if (i >= classes.length) return null;
+    if (classes[i] == 'silent') {
+      // A silent @-mention.  We ignore this flag; see [UserMentionNode].
+      i++;
+    }
+
+    if (i >= classes.length) return null;
+    if (classes[i] == 'user-mention' ||
+       (classes[i] == 'user-group-mention' && !isChannelWildcardClassIncluded)) {
```
gnprice, line 900: nit: indentation; and operators go after newline:
```suggestion
    if (classes[i] == 'user-mention'
        || (classes[i] == 'user-group-mention' && !isChannelWildcardClassIncluded)) {
```

(See examples in `parseInlineContent` below.)

With less indentation, the second sub-condition looks like it's a peer of the `if` condition as a whole.

For the rationale for `||` after newline, see:
https://github.com/zulip/zulip-mobile/blob/a115df1f71c9dc31e9b41060a8d57b51c017d786/tools/formatting.eslintrc.yaml#L20-L23

test/model/content_test.dart
```diff
@@ -1213,7 +1255,13 @@ void main() {
     testParseExample(ContentExample.groupMentionSilent);
     testParseExample(ContentExample.groupMentionSilentClassOrderReversed);
 
-    // TODO test wildcard mentions
```
gnprice, line 1216: Cool, good to have this handled :slightly_smiling_face:

test/model/content_test.dart
```diff
@@ -139,6 +139,48 @@ class ContentExample {
...
+    const UserMentionNode(nodes: [TextNode('everyone')]));
+
+  static final channelWildcardMentionSilentClassOrderReversed = ContentExample.inline(
+    'silent channel wildcard @-mention, class order reversed',
+    "@_**channel**", // (hypothetical server variation)
+    expectedText: 'channel',
+    '<p><span class="silent user-mention channel-wildcard-mention" data-user-id="*">channel</span></p>',
+    const UserMentionNode(nodes: [TextNode('channel')]));
+
+  static final legacyChannelWildcardMentionPlain = ContentExample.inline(
```
gnprice, line 163: Yeah, this naming is good: the current/new way gets the simple name, and the old way gets the longer "legacy" name.

gnprice (commented):
Thanks @sm-sayedi for taking care of this, and @PIG208 for the previous reviews! This all looks good; some nits below.

sm-sayedi:
Thanks @gnprice and @PIG208 for the reviews! Changes pushed.

lib/model/content.dart
```diff
@@ -880,16 +880,27 @@ class _ZulipContentParser {
...
+    if (classes[i] == 'channel-wildcard-mention') {
+      // Newer channel wildcard mentions have this class; older ones don't.
+      i++;
+      hasChannelWildcardClass = true;
+    }
+
     if (i >= classes.length) return null;
     if (classes[i] == 'silent') {
-      // A silent @-mention.  We ignore this flag; see [UserMentionNode].
+      // A silent @-mention. We ignore this flag; see [UserMentionNode].
```
gnprice, line 893: nit: extraneous change

gnprice (commented):
Thanks for the revision! Looks good, with one nit below. I'll fix that and merge.

PR #1075, opened 2024-11-23
Update message.dart

Adityakk9031:


gnprice:
This doesn't look intended to become something we could merge. In any case there's no description of what it's aiming to accomplish, so there's nothing for us to provide feedback on. Closing this PR as it doesn't serve any purpose to have in our tracker.

You're free to do experiments like this one by modifying our code, of course. If you want them on GitHub, you can push them to your own clone.

Issue #1035, opened 2024-11-01
Bottom tabs and main menu


gnprice:
This is our new design for the overall navigation of the app. It replaces the current placeholder home page. In this design:

1. There is a bottom tab bar (aka ["navigation bar"](https://m3.material.io/components/navigation-bar/overview), in Material 3 terms). It has 6 buttons: Inbox, Combined feed, Mentions, Direct messages, Channels, Menu.
   - All of those buttons except Menu go to screens we already have implemented. Menu is described below.
   - The navigation bar appears only on the screens that it leads directly to. In particular it does not appear on message lists.
   - Changes vs. RN legacy app: **Search** and **Starred** move from the top bar into the menu. There's no top tab bar; instead there's an app bar with the current page's title ("Inbox, "Channels", etc.).
   - For detailed design of the navigation bar, see Figma: [here](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2957-50895&node-type=section&t=ra9oZtMkj37ax6je-0), and [further details here](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2961-53750&node-type=section&t=ra9oZtMkj37ax6je-0). Screenshot: 
![image](https://github.com/user-attachments/assets/28f40cef-7505-40ee-8459-6721bf318c78)

2. The screen with the bottom tab bar is the root of the navigation stack; there's no navigating "back" or "up" from there. For how the user reaches the choose-account page (which we currently put at the root of the navigation stack), see below and 4.
    - We will add a button on the loading screen for the home page. When clicked, it brings the user to the choose-account page. This is necessary in case loading the account data takes a long time. This button should only appear after a few seconds of loading.
    - As a follow-up, we need to support logging out an account without UI interacting with the choose-account page UI. This means that the account can be logged out at any navigation level, which is not possible as of now (until we implement #737 to invalidate API keys). A current side-effect of logging out is that the routes relying on the `PerAccountStore` of the account will be removed from the navigator stack. We need to ensure that the user ends up on choose-account page when that happens.

4. The menu is a modal bottom sheet. It has the same options as the navigation bar (except Menu itself), plus "Starred messages".
   - In the future the menu will gain more options as we implement them:
     - Settings: #97
     - Notifications screen: #661
     - "My profile"
     - Search (but that's a post-launch feature): #252
     - and doubtless others.
[...]

Issue #1039, opened 2024-11-01
Fuse recipient header into app bar for topic narrows


gnprice:
Currently when viewing the message list for a topic narrow, we have an app bar that shows both the channel and the topic, and then a recipient header that shows the topic. (And both app bar and recipient header have the channel's subscription color.)

This looks kind of redundant. Especially so when the thread is long enough to fill the screen, so that the recipient header reaches the top — then the topic name gets displayed twice, one right below the other, at the top of the screen. It'd be good to eliminate the redundancy.

Vlad has made prototype designs for us that do that; see previous issue https://github.com/zulip/zulip-mobile/issues/5511. Here's a screenshot: ![image](https://github.com/user-attachments/assets/d45693b3-6d6f-4e79-b5d7-eda3a6777699)

Design questions we'll want to resolve before implementing this include:
* Have an example where the thread is shorter than the screen, so there's empty space between the app bar and the first message; make sure we're happy with how that looks, or adjust something. (Possibly switch to pinning the earliest message at the top instead of the latest at the bottom, so the empty space goes between the last message and the compose box.)
* Decide how to handle showing the messages' dates, which currently live in the recipient header.

This is a post-launch issue; it doesn't represent any missing functionality, and I think the redundancy is OK for launch.

chrisbobbe:
> * Have an example where the thread is shorter than the screen, so there's empty space between the app bar and the first message; make sure we're happy with how that looks, or adjust something. (Possibly switch to pinning the earliest message at the top instead of the latest at the bottom, so the empty space goes between the last message and the compose box.)

The message list might also be empty, right, and in that case I think we don't show any recipient headers.

Issue #535, opened 2024-02-26
Check for basic usability with screen reader


gnprice:
We should try out the app with iOS's VoiceOver and/or Android's TalkBack, and make sure it's possible to do the essentials to be able to use the app at all:
* log in
* get between the different top-level pages (Inbox, All messages, Subscribed streams, Direct messages)
* use each of those to get to a relevant message list
* read the messages
* send a message

If there's anything that doesn't work, we should file and triage an issue for that so we can fix it.

Generally the fixes will involve Flutter's [`Semantics`](https://main-api.flutter.dev/flutter/widgets/Semantics-class.html) widget and related parts of the Flutter API, to provide the appropriate semantic information for screen readers (like VoiceOver and TalkBack) and other assistive technology to use.

### How to approach this issue

See my discussion in https://github.com/zulip/zulip-mobile/issues/4394 about how to start learning to use these systems and to evaluate the app with them.

### Not in scope

This issue is milestoned "Launch". Out of scope for this issue is to do any more comprehensive review of the rest of the app's features: reactions, less-common message content features, action sheets, lightbox, compose typeaheads, etc., etc. Those features are also important, and I'll file a companion issue to this one for covering all those, which will get the "Post-launch" milestone.

The current state in zulip-mobile — the last I looked, which I see was three years ago — is described in https://github.com/zulip/zulip-mobile/issues/4394: the app basically works, but there are a long list of glitches. We did fix the two issues highlighted as most important (in https://github.com/zulip/zulip-mobile/issues/4232) by an organization with some people who actually use screen readers regularly; but we don't really know how usable the app is for such users.

MINEGHOST007:
I would like to work on this issue I'll be able to provide screenshots and recordings for Android TalkBack .

gnprice:
Sure, please go ahead.

Recordings will be a helpful bonus, but please also include transcriptions, along with those screenshots for context. The transcriptions will help make it easy for others to review your findings.

For detailed discussion, please start a chat thread on chat.zulip.org in #mobile-dev-help. (Zulip makes a better medium for discussion than GitHub does.) Then after we have some concrete findings we're confident in, we can post those here on this issue thread, keeping this thread easy to read.

AshutoshKhadse23:
Hello @gnprice ,
Is someone working on it already if not then I would like to work on it ?
I seen the comment but unable to find the updates on that CZO and here.

gnprice:
[...]

PR #1076, opened 2024-11-23
Implement bottom tabs and main menu

PIG208:
This is stacked on top of some commits cherry-picked from #1041:
* action_sheet [nfc]: Use a generalized name for the cancel button
* action_sheet [nfc]: Extract ActionSheetMenuItemButton

See also: [CZO discussion](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Buttons.20on.20the.20bottom.20tabs.20and.20main.20menu)

<details>

| light | dark |
| - | - |
|   |  |
|  |  |

</details>

Fixes: #1035

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -19,6 +19,19 @@
   "@chooseAccountPageTitle": {
     "description": "Title for the page to choose between Zulip accounts."
   },
+  "switchAccountButton": "Switch account",
+  "@switchAccountButton": {
+    "description": "Label for main-menu button leading to the choose-account page."
+  },
+  "tryAnotherAccountMessage": "Your account at {url} is taking a while to load.",
+  "@tryAnotherAccountMessage": {
+    "description": "Message that appears on the loading screen after waiting for some time.",
+    "url": {"type": "String", "example": "http://chat.example.com/"}
+  },
+  "tryAnotherAccountButton": "Try another account",
+  "@tryAnotherAccountButton": {
+    "description": "Label for loading screen button prompting user to try another account."
+  },
   "chooseAccountPageLogOutButton": "Log out",
   "@chooseAccountPageLogOutButton": {
     "description": "Label for the 'Log out' button for an account on the choose-account page"
@@ -561,6 +574,10 @@
   "@userRoleUnknown": {
     "description": "Label for UserRole.unknown"
   },
+  "inboxPageTitle": "Inbox",
+  "@inboxPageTitle": {
+    "description": "Title for the page with unreads."
+  },
   "recentDmConversationsPageTitle": "Direct messages",
   "@recentDmConversationsPageTitle": {
     "description": "Title for the page with a list of DM conversations."
@@ -576,6 +593,14 @@
   "starredMessagesPageTitle": "Starred messages",
   "@starredMessagesPageTitle": {
     "description": "Page title for the 'Starred messages' message view."
+  },
+  "channelsPageTitle": "Channels",
+  "@channelsPageTitle": {
+    "description": "Title for the page with a list of subscribed channels."
+  },
+  "mainMenuMyProfile": "My profile",
+  "@mainMenuMyProfile": {
+    "description": "Label for main-menu button leading to the user's own profile."
   },
   "channelFeedButtonTooltip": "Channel feed",
   "@channelFeedButtonTooltip": {
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -279,6 +279,7 @@ class ChooseAccountPage extends StatelessWidget {
     final globalStore = GlobalStoreWidget.of(context);
     return Scaffold(
       appBar: AppBar(
+        titleSpacing: 16,
         title: Text(zulipLocalizations.chooseAccountPageTitle),
         actions: const [ChooseAccountPageOverflowButton()]),
       body: SafeArea(
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -13,7 +13,6 @@ import 'about_zulip.dart';
 import 'actions.dart';
 import 'dialog.dart';
 import 'home.dart';
-import 'inbox.dart';
 import 'login.dart';
 import 'page.dart';
 import 'store.dart';
@@ -209,11 +208,10 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
 
           onGenerateInitialRoutes: (_) {
             return [
-              MaterialWidgetRoute(page: const ChooseAccountPage()),
-              if (initialAccountId != null) ...[
+              if (initialAccountId == null)
+                MaterialWidgetRoute(page: const ChooseAccountPage())
+              else
                 HomePage.buildRoute(accountId: initialAccountId),
-                InboxPage.buildRoute(accountId: initialAccountId),
-              ],
             ];
           });
         }));
@@ -271,8 +269,7 @@ class ChooseAccountPage extends StatelessWidget {
         // The default trailing padding with M3 is 24px. Decrease by 12 because
         // IconButton (the "…" button) comes with 12px padding on all sides.
         contentPadding: const EdgeInsetsDirectional.only(start: 16, end: 12),
-        onTap: () => Navigator.push(context,
-          HomePage.buildRoute(accountId: accountId))));
+        onTap: () => HomePage.navigate(context, accountId: accountId)));
   }
 
   @override
@@ -281,13 +278,20 @@ class ChooseAccountPage extends StatelessWidget {
     final zulipLocalizations = ZulipLocalizations.of(context);
     assert(!PerAccountStoreWidget.debugExistsOf(context));
     final globalStore = GlobalStoreWidget.of(context);
+
+    // Borrowed from [AppBar.build].
+    // See documentation on [ModalRoute.impliesAppBarDismissal]:
+    // > Whether an [AppBar] in the route should automatically add a back button or
+    // > close button.
+    final hasBackButton = ModalRoute.of(context)?.impliesAppBarDismissal ?? false;
+
     return MenuButtonTheme(
       data: MenuButtonThemeData(style: MenuItemButton.styleFrom(
         backgroundColor: colorScheme.secondaryContainer,
         foregroundColor: colorScheme.onSecondaryContainer)),
       child: Scaffold(
         appBar: AppBar(
-          titleSpacing: 16,
+          titleSpacing: hasBackButton ? null : 16,
           title: Text(zulipLocalizations.chooseAccountPageTitle),
           actions: const [ChooseAccountPageOverflowButton()]),
         body: SafeArea(
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -229,6 +229,7 @@ class ChooseAccountPage extends StatelessWidget {
     required Widget title,
     Widget? subtitle,
   }) {
+    final colorScheme = ColorScheme.of(context);
     final designVariables = DesignVariables.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
     final materialLocalizations = MaterialLocalizations.of(context);
@@ -237,6 +238,8 @@ class ChooseAccountPage extends StatelessWidget {
       child: ListTile(
         title: title,
         subtitle: subtitle,
+        tileColor: colorScheme.secondaryContainer,
+        textColor: colorScheme.onSecondaryContainer,
         trailing: MenuAnchor(
           menuChildren: [
             MenuItemButton(
@@ -274,36 +277,40 @@ class ChooseAccountPage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final colorScheme = ColorScheme.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
     assert(!PerAccountStoreWidget.debugExistsOf(context));
     final globalStore = GlobalStoreWidget.of(context);
-    return Scaffold(
-      appBar: AppBar(
-        titleSpacing: 16,
-        title: Text(zulipLocalizations.chooseAccountPageTitle),
-        actions: const [ChooseAccountPageOverflowButton()]),
-      body: SafeArea(
-        minimum: const EdgeInsets.fromLTRB(8, 0, 8, 8),
-        child: Center(
-          child: ConstrainedBox(
-            constraints: const BoxConstraints(maxWidth: 400),
-            child: Column(mainAxisSize: MainAxisSize.min, children: [
-              Flexible(child: SingleChildScrollView(
-                padding: const EdgeInsets.only(top: 8),
-                child: Column(mainAxisSize: MainAxisSize.min, children: [
-                  for (final (:accountId, :account) in globalStore.accountEntries)
-                    _buildAccountItem(context,
-                      accountId: accountId,
-                      title: Text(account.realmUrl.toString()),
-                      subtitle: Text(account.email)),
-                ]))),
-              const SizedBox(height: 12),
-              ElevatedButton(
-                onPressed: () => Navigator.push(context,
-                  AddAccountPage.buildRoute()),
-                child: Text(zulipLocalizations.chooseAccountButtonAddAnAccount)),
-            ]))),
-      ));
+    return MenuButtonTheme(
+      data: MenuButtonThemeData(style: MenuItemButton.styleFrom(
+        backgroundColor: colorScheme.secondaryContainer,
+        foregroundColor: colorScheme.onSecondaryContainer)),
+      child: Scaffold(
+        appBar: AppBar(
+          titleSpacing: 16,
+          title: Text(zulipLocalizations.chooseAccountPageTitle),
+          actions: const [ChooseAccountPageOverflowButton()]),
+        body: SafeArea(
+          minimum: const EdgeInsets.fromLTRB(8, 0, 8, 8),
+          child: Center(
+            child: ConstrainedBox(
+              constraints: const BoxConstraints(maxWidth: 400),
+              child: Column(mainAxisSize: MainAxisSize.min, children: [
+                Flexible(child: SingleChildScrollView(
+                  padding: const EdgeInsets.only(top: 8),
+                  child: Column(mainAxisSize: MainAxisSize.min, children: [
+                    for (final (:accountId, :account) in globalStore.accountEntries)
+                      _buildAccountItem(context,
+                        accountId: accountId,
+                        title: Text(account.realmUrl.toString()),
+                        subtitle: Text(account.email)),
+                  ]))),
+                const SizedBox(height: 12),
+                ElevatedButton(
+                  onPressed: () => Navigator.push(context,
+                    AddAccountPage.buildRoute()),
+                  child: Text(zulipLocalizations.chooseAccountButtonAddAnAccount)),
+              ]))))));
   }
 }
 
--- a/lib/widgets/home.dart
+++ b/lib/widgets/home.dart
@@ -24,8 +24,6 @@ class HomePage extends StatelessWidget {
     final store = PerAccountStoreWidget.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
 
-    final colorScheme = ColorScheme.of(context);
-
     InlineSpan bold(String text) => TextSpan(
       style: const TextStyle().merge(weightVariableTextStyle(context, wght: 700)),
       text: text);
@@ -37,61 +35,57 @@ class HomePage extends StatelessWidget {
 
     return Scaffold(
       appBar: ZulipAppBar(title: const Text("Home")),
-      body: ElevatedButtonTheme(
-        data: ElevatedButtonThemeData(style: ButtonStyle(
-          backgroundColor: WidgetStatePropertyAll(colorScheme.secondaryContainer),
-          foregroundColor: WidgetStatePropertyAll(colorScheme.onSecondaryContainer))),
-        child: Center(
-          child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
-            DefaultTextStyle.merge(
-              textAlign: TextAlign.center,
-              style: const TextStyle(fontSize: 18),
-              child: Column(children: [
-                Text.rich(TextSpan(
-                  text: 'Connected to: ',
-                  children: [bold(store.realmUrl.toString())])),
-              ])),
+      body: Center(
+        child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
+          DefaultTextStyle.merge(
+            textAlign: TextAlign.center,
+            style: const TextStyle(fontSize: 18),
+            child: Column(children: [
+              Text.rich(TextSpan(
+                text: 'Connected to: ',
+                children: [bold(store.realmUrl.toString())])),
+            ])),
+          const SizedBox(height: 16),
+          ElevatedButton(
+            onPressed: () => Navigator.push(context,
+              MessageListPage.buildRoute(context: context,
+                narrow: const CombinedFeedNarrow())),
+            child: Text(zulipLocalizations.combinedFeedPageTitle)),
+          const SizedBox(height: 16),
+          ElevatedButton(
+            onPressed: () => Navigator.push(context,
+              MessageListPage.buildRoute(context: context,
+                narrow: const MentionsNarrow())),
+            child: Text(zulipLocalizations.mentionsPageTitle)),
+          const SizedBox(height: 16),
+          ElevatedButton(
+            onPressed: () => Navigator.push(context,
+              MessageListPage.buildRoute(context: context,
+                narrow: const StarredMessagesNarrow())),
+            child: Text(zulipLocalizations.starredMessagesPageTitle)),
+          const SizedBox(height: 16),
+          ElevatedButton(
+            onPressed: () => Navigator.push(context,
+              InboxPage.buildRoute(context: context)),
+            child: const Text("Inbox")), // TODO(i18n)
+          const SizedBox(height: 16),
+          ElevatedButton(
+            onPressed: () => Navigator.push(context,
+              SubscriptionListPage.buildRoute(context: context)),
+            child: const Text("Subscribed channels")),
+          const SizedBox(height: 16),
+          ElevatedButton(
+            onPressed: () => Navigator.push(context,
+              RecentDmConversationsPage.buildRoute(context: context)),
+            child: Text(zulipLocalizations.recentDmConversationsPageTitle)),
+          if (testStreamId != null) ...[
             const SizedBox(height: 16),
             ElevatedButton(
               onPressed: () => Navigator.push(context,
                 MessageListPage.buildRoute(context: context,
-                  narrow: const CombinedFeedNarrow())),
-              child: Text(zulipLocalizations.combinedFeedPageTitle)),
-            const SizedBox(height: 16),
-            ElevatedButton(
-              onPressed: () => Navigator.push(context,
-                MessageListPage.buildRoute(context: context,
-                  narrow: const MentionsNarrow())),
-              child: Text(zulipLocalizations.mentionsPageTitle)),
-            const SizedBox(height: 16),
-            ElevatedButton(
-              onPressed: () => Navigator.push(context,
-                MessageListPage.buildRoute(context: context,
-                  narrow: const StarredMessagesNarrow())),
-              child: Text(zulipLocalizations.starredMessagesPageTitle)),
-            const SizedBox(height: 16),
-            ElevatedButton(
-              onPressed: () => Navigator.push(context,
-                InboxPage.buildRoute(context: context)),
-              child: const Text("Inbox")), // TODO(i18n)
-            const SizedBox(height: 16),
-            ElevatedButton(
-              onPressed: () => Navigator.push(context,
-                SubscriptionListPage.buildRoute(context: context)),
-              child: const Text("Subscribed channels")),
-            const SizedBox(height: 16),
-            ElevatedButton(
-              onPressed: () => Navigator.push(context,
-                RecentDmConversationsPage.buildRoute(context: context)),
-              child: Text(zulipLocalizations.recentDmConversationsPageTitle)),
-            if (testStreamId != null) ...[
-              const SizedBox(height: 16),
-              ElevatedButton(
-                onPressed: () => Navigator.push(context,
-                  MessageListPage.buildRoute(context: context,
-                    narrow: ChannelNarrow(testStreamId!))),
-                child: const Text("#test here")), // scaffolding hack, see above
-            ],
-          ]))));
+                  narrow: ChannelNarrow(testStreamId!))),
+              child: const Text("#test here")), // scaffolding hack, see above
+          ],
+        ])));
   }
 }
--- a/lib/widgets/home.dart
+++ b/lib/widgets/home.dart
@@ -1,91 +1,598 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 
 import '../generated/l10n/zulip_localizations.dart';
 import '../model/narrow.dart';
+import 'action_sheet.dart';
+import 'app.dart';
 import 'app_bar.dart';
+import 'color.dart';
+import 'content.dart';
+import 'icons.dart';
 import 'inbox.dart';
+import 'inset_shadow.dart';
 import 'message_list.dart';
 import 'page.dart';
+import 'profile.dart';
 import 'recent_dm_conversations.dart';
 import 'store.dart';
 import 'subscription_list.dart';
 import 'text.dart';
+import 'theme.dart';
 
-class HomePage extends StatelessWidget {
+enum _HomePageTab {
+  inbox,
+  channels,
+  directMessages,
+}
+
+class HomePage extends StatefulWidget {
   const HomePage({super.key});
 
   static Route<void> buildRoute({required int accountId}) {
     return MaterialAccountWidgetRoute(accountId: accountId,
-        page: const HomePage());
+      loadingPlaceholderPage: _LoadingPlaceholderPage(accountId: accountId),
+      page: const HomePage());
+  }
+
+  /// Navigate to [HomePage], ensuring that its route is at the root level.
+  static void navigate(BuildContext context, {required int accountId}) {
+    final navigator = Navigator.of(context);
+    navigator.popUntil((route) => route.isFirst);
+    unawaited(navigator.pushReplacement(
+      HomePage.buildRoute(accountId: accountId)));
+  }
+
+  @override
+  State<HomePage> createState() => _HomePageState();
+}
+
+class _HomePageState extends State<HomePage> {
+  late final _tab = ValueNotifier(_HomePageTab.inbox);
+
+  @override
+  void initState() {
+    super.initState();
+    _tab.addListener(_tabChanged);
+  }
+
+  @override
+  void dispose() {
+    _tab.dispose();
+    super.dispose();
+  }
+
+  void _tabChanged() {
+    setState(() {
+      // The actual state lives in [_tab].
+    });
+  }
+
+  String get _currentTabTitle {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    switch(_tab.value) {
+      case _HomePageTab.inbox:
+        return zulipLocalizations.inboxPageTitle;
+      case _HomePageTab.channels:
+        return zulipLocalizations.channelsPageTitle;
+      case _HomePageTab.directMessages:
+        return zulipLocalizations.recentDmConversationsPageTitle;
+    }
   }
 
   @override
   Widget build(BuildContext context) {
-    final store = PerAccountStoreWidget.of(context);
-    final zulipLocalizations = ZulipLocalizations.of(context);
+    const pageBodies = [
+      (_HomePageTab.inbox,          InboxPageBody()),
+      (_HomePageTab.channels,       SubscriptionListPageBody()),
+      // TODO(#1094): Users
+      (_HomePageTab.directMessages, RecentDmConversationsPageBody()),
+    ];
 
-    InlineSpan bold(String text) => TextSpan(
-      style: const TextStyle().merge(weightVariableTextStyle(context, wght: 700)),
-      text: text);
-
-    int? testStreamId;
-    if (store.connection.realmUrl.origin == 'https://chat.zulip.org') {
-      testStreamId = 7; // i.e. `#test here`; TODO cut this scaffolding hack
+    _NavigationBarButton button(_HomePageTab tab, IconData icon) {
+      return _NavigationBarButton(icon: icon,
+        selected: _tab.value == tab,
+        onPressed: () {
+          _tab.value = tab;
+        });
     }
 
+    // TODO(a11y): add tooltips for these buttons
+    final navigationBarButtons = [
+      button(_HomePageTab.inbox,          ZulipIcons.inbox),
+      button(_HomePageTab.channels,       ZulipIcons.hash_italic),
+      // TODO(#1094): Users
+      button(_HomePageTab.directMessages, ZulipIcons.user),
+      _NavigationBarButton(        icon: ZulipIcons.menu,
+        selected: false,
+        onPressed: () => _showMainMenu(context, tabNotifier: _tab)),
+    ];
+
+    final designVariables = DesignVariables.of(context);
     return Scaffold(
-      appBar: ZulipAppBar(title: const Text("Home")),
+      appBar: ZulipAppBar(titleSpacing: 16,
+        title: Text(_currentTabTitle)),
+      body: Stack(
+        children: [
+          for (final (tab, body) in pageBodies)
+            // TODO(#535): Decide if we find it helpful to use something like
+            //   [SemanticsProperties.namesRoute] to structure this UI better
+            //   for screen-reader software.
+            Offstage(offstage: tab != _tab.value, child: body),
+        ]),
+      bottomNavigationBar: DecoratedBox(
+        decoration: BoxDecoration(
+          border: Border(top: BorderSide(color: designVariables.borderBar)),
+          color: designVariables.bgBotBar),
+        child: SafeArea(
+          child: SizedBox(height: 48,
+            child: Center(
+              child: ConstrainedBox(
+                // TODO(design): determine a suitable max width for bottom nav bar
+                constraints: const BoxConstraints(maxWidth: 600),
+                child: Row(
+                  crossAxisAlignment: CrossAxisAlignment.stretch,
+                  children: [
+                    for (final navigationBarButton in navigationBarButtons)
+                      Expanded(child: navigationBarButton),
+                  ])))))));
+  }
+}
+
+const kTryAnotherAccountWaitPeriod = Duration(seconds: 5);
+
+class _LoadingPlaceholderPage extends StatefulWidget {
+  const _LoadingPlaceholderPage({required this.accountId});
+
+  final int accountId;
+
+  @override
+  State<_LoadingPlaceholderPage> createState() => _LoadingPlaceholderPageState();
+}
+
+class _LoadingPlaceholderPageState extends State<_LoadingPlaceholderPage> {
+  Timer? tryAnotherAccountTimer;
+  bool showTryAnotherAccount = false;
+
+  @override
+  void initState() {
+    super.initState();
+    tryAnotherAccountTimer = Timer(kTryAnotherAccountWaitPeriod, () {
+      setState(() {
+        showTryAnotherAccount = true;
+      });
+    });
+  }
+
+  @override
+  void dispose() {
+    tryAnotherAccountTimer?.cancel();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final realmUrl = GlobalStoreWidget.of(context)
+      .getAccount(widget.accountId)!.realmUrl;
+
+    return Scaffold(
+      appBar: AppBar(),
       body: Center(
-        child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
-          DefaultTextStyle.merge(
-            textAlign: TextAlign.center,
-            style: const TextStyle(fontSize: 18),
-            child: Column(children: [
-              Text.rich(TextSpan(
-                text: 'Connected to: ',
-                children: [bold(store.realmUrl.toString())])),
-            ])),
-          const SizedBox(height: 16),
-          ElevatedButton(
-            onPressed: () => Navigator.push(context,
-              MessageListPage.buildRoute(context: context,
-                narrow: const CombinedFeedNarrow())),
-            child: Text(zulipLocalizations.combinedFeedPageTitle)),
-          const SizedBox(height: 16),
-          ElevatedButton(
-            onPressed: () => Navigator.push(context,
-              MessageListPage.buildRoute(context: context,
-                narrow: const MentionsNarrow())),
-            child: Text(zulipLocalizations.mentionsPageTitle)),
-          const SizedBox(height: 16),
-          ElevatedButton(
-            onPressed: () => Navigator.push(context,
-              MessageListPage.buildRoute(context: context,
-                narrow: const StarredMessagesNarrow())),
-            child: Text(zulipLocalizations.starredMessagesPageTitle)),
-          const SizedBox(height: 16),
-          ElevatedButton(
-            onPressed: () => Navigator.push(context,
-              InboxPage.buildRoute(context: context)),
-            child: const Text("Inbox")), // TODO(i18n)
-          const SizedBox(height: 16),
-          ElevatedButton(
-            onPressed: () => Navigator.push(context,
-              SubscriptionListPage.buildRoute(context: context)),
-            child: const Text("Subscribed channels")),
-          const SizedBox(height: 16),
-          ElevatedButton(
-            onPressed: () => Navigator.push(context,
-              RecentDmConversationsPage.buildRoute(context: context)),
-            child: Text(zulipLocalizations.recentDmConversationsPageTitle)),
-          if (testStreamId != null) ...[
-            const SizedBox(height: 16),
-            ElevatedButton(
-              onPressed: () => Navigator.push(context,
-                MessageListPage.buildRoute(context: context,
-                  narrow: ChannelNarrow(testStreamId!))),
-              child: const Text("#test here")), // scaffolding hack, see above
-          ],
-        ])));
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            const CircularProgressIndicator(),
+            Visibility(
+              visible: showTryAnotherAccount,
+              maintainSize: true,
+              maintainAnimation: true,
+              maintainState: true,
+              child: Padding(
+                padding: const EdgeInsets.symmetric(horizontal: 8),
+                child: Column(
+                  children: [
+                    const SizedBox(height: 16),
+                    Text(zulipLocalizations.tryAnotherAccountMessage(realmUrl.toString())),
+                    const SizedBox(height: 8),
+                    ElevatedButton(
+                      onPressed: () => Navigator.push(context,
+                        MaterialWidgetRoute(page: const ChooseAccountPage())),
+                      child: Text(zulipLocalizations.tryAnotherAccountButton)),
+                  ]))),
+          ])));
+  }
+}
+
+class _NavigationBarButton extends StatelessWidget {
+  const _NavigationBarButton({
+    required this.icon,
+    required this.selected,
+    required this.onPressed,
+  });
+
+  final IconData icon;
+  final bool selected;
+  final void Function() onPressed;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    final iconColor = WidgetStateColor.fromMap({
+      WidgetState.pressed:  designVariables.iconSelected,
+      ~WidgetState.pressed: selected ? designVariables.iconSelected
+                                     : designVariables.icon,
+    });
+
+    return AnimatedScaleOnTap(
+      scaleEnd: 0.875,
+      duration: const Duration(milliseconds: 100),
+      child: IconButton(
+        icon: Icon(icon, size: 24),
+        onPressed: onPressed,
+        style: IconButton.styleFrom(
+          // TODO(#417): Disable splash effects for all buttons globally.
+          splashFactory: NoSplash.splashFactory,
+          highlightColor: designVariables.navigationButtonBg,
+          shape: const RoundedRectangleBorder(
+            borderRadius: BorderRadius.all(Radius.circular(4))),
+        ).copyWith(foregroundColor: iconColor)));
+  }
+}
+
+void _showMainMenu(BuildContext context, {
+  required ValueNotifier<_HomePageTab> tabNotifier,
+}) {
+  final menuItems = <Widget>[
+    // TODO(#252): Search
+    // const SizedBox(height: 8),
+    _InboxButton(tabNotifier: tabNotifier),
+    // TODO: Recent conversations
+    const _MentionsButton(),
+    const _StarredMessagesButton(),
+    const _CombinedFeedButton(),
+    // TODO: Drafts
+    _ChannelsButton(tabNotifier: tabNotifier),
+    _DirectMessagesButton(tabNotifier: tabNotifier),
+    // TODO(#1094): Users
+    const _MyProfileButton(),
+    const _SwitchAccountButton(),
+    // TODO(#198): Set my status
+    // const SizedBox(height: 8),
+    // TODO(#97): Settings
+    // TODO(#661): Notifications
+    // const SizedBox(height: 8),
+    // TODO(#1095): VersionInfo
+  ];
+
+  final designVariables = DesignVariables.of(context);
+  final accountId = PerAccountStoreWidget.accountIdOf(context);
+  showModalBottomSheet<void>(
+    context: context,
+    // Clip.hardEdge looks bad; Clip.antiAliasWithSaveLayer looks pixel-perfect
+    // on my iPhone 13 Pro but is marked as "much slower":
+    //   https://api.flutter.dev/flutter/dart-ui/Clip.html
+    clipBehavior: Clip.antiAlias,
+    useSafeArea: true,
+    isScrollControlled: true,
+    // TODO: Fix the issue that the color does not respond when the theme
+    //   changes, because `designVariables` was retrieved from a gesture handler,
+    //   not a build method.  Discussion and screenshots:
+    //     https://github.com/zulip/zulip-flutter/pull/1076/files#r1872659043
+    backgroundColor: designVariables.bgBotBar,
+    builder: (BuildContext _) {
+      return PerAccountStoreWidget(
+        accountId: accountId,
+        child: SafeArea(
+          minimum: const EdgeInsets.only(bottom: 8),
+          child: Column(
+            crossAxisAlignment: CrossAxisAlignment.stretch,
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              Flexible(child: InsetShadowBox(
+                top: 8, bottom: 8,
+                color: designVariables.bgBotBar,
+                child: SingleChildScrollView(
+                  padding: const EdgeInsets.symmetric(vertical: 8, horizontal: 8),
+                  child: Column(children: menuItems)))),
+              const Padding(
+                padding: EdgeInsets.symmetric(horizontal: 16),
+                child: AnimatedScaleOnTap(
+                  scaleEnd: 0.95,
+                  duration: Duration(milliseconds: 100),
+                  child: ActionSheetCancelButton())),
+            ])));
+    });
+}
+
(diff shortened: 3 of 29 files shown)


gnprice:
(I turned the link to #1035 into the kind GitHub recognizes specially; that's helpful for example because it makes it appear in the issue's row on the [project board]([Bottom tabs and main menu](https://github.com/zulip/zulip-flutter/issues/1035)	https://github.com/zulip/zulip-flutter/issues/1035). I realize the PR isn't ready to be merged as our solution to the issue, but that's expressed by its draft status; and I believe it is intended to be the PR that fixes the issue.)

lib/widgets/home.dart
```diff
@@ -0,0 +1,461 @@
...
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return 'Attach file';
+  }
+
+  @override
+  void onPressed() {
+    // TODO: implement onPressed
+  }
```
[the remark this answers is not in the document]
gnprice: Oh and I guess there'll also be a method or getter to compute what counter to show, if any.

Maybe a good structure for that is a method like `Widget? buildTrailing(BuildContext context)`. Then most subclasses will return a `_MenuItemCounter`, or null; but e.g. the "Notifications" row (in the future! it's out of scope for this PR) can return that yellow warning icon.

lib/widgets/home.dart
```diff
@@ -0,0 +1,461 @@
...
+      onTapCancel: () => _changeScale(1),
+      child: AnimatedScale(
+        scale: _scale,
+        duration: widget.duration,
+        curve: Curves.easeOut,
+        child: widget.child));
+  }
+}
+
+class NavigationButton extends StatelessWidget {
```
gnprice, line 421: nit: put this just after _HomePageState, since it's closely tied to that

lib/widgets/home.dart
```diff
@@ -0,0 +1,461 @@
...
+  ];
+  final animationController = _PausableAnimationController(
+    vsync: Navigator.of(context),
+    duration: _bottomSheetEnterDuration,
+    reverseDuration: _bottomSheetExitDuration,
+  );
+  bool handleScroll(ScrollNotification notification) {
+    animationController.paused = notification.metrics.extentBefore != 0;
+    return false;
+  }
```
gnprice, line 159: wip; nav: Disable bottom sheet dragging when scrolled

I think the behavior before this commit is fine and actually a bit nicer. So let's leave out this complication.

I gather you're referring to this line in the Figma:
> It would be nice if swipe down can close the menu, but it is important to do that only, when content is scrolled to its top position. And it isn’t just a swipe gesture but full menu dragging 

But I think the thing Vlad means is just: if you've scrolled down in the content and start swiping down, it should scroll you back up in the content (just like normal for a scrollable) instead of having the menu start closing.

When I run with the version before this commit, that all works fine: if you're scrolled down, then swiping anywhere within the scrollable acts on the scrollable before doing anything to the menu as a whole. If you touch the header and drag that, then that does control the menu… but I think that's fine and actually good. After this commit, attempting to drag the header doesn't do anything if the options are scrolled, which to me feels oddly inert like something's not working.

gnprice (commented):
Exciting! Looking forward to getting this landed.

Comments below on the draft.

The important high-level point (of which some corollaries appear below) is: the priority with this PR is to get to something we can merge in time for the blog post about the beta.

That means a version that has all the essential functionality, looks reasonable, and meets our usual standards for the code and tests; but any aspects of the design that turn out to be tricky to implement are a lower priority than getting it merged. Because the issue is complex already, anything that's tricky and not essential should be punted until everything essential is complete.

For a concrete timeline: the target for the blog post is by Dec 12, the week after next. I think that calls for a version ready for review by the first half of next week, to give time for review, revisions, and any follow-ups we realize we need after getting to use this a bit.

lib/widgets/home.dart
PIG208: Yeah, I agree to leaving this out.

> It would be nice if swipe down can close the menu

I think to fully implement this (out of scope) feature, we will need to allow over scrolling the inner content (swiping down the menu items and reaching the top of the menu) to drag the bottom sheet; once you scroll pass a certain extent the menu collapses, as it does when you control it by dragging the header. I guess the menu would otherwise rebound back in position if you release it.

It makes sense to leave this out because it will be quite complicated. An implementation might have to listen to notifications like `OverscrollNotification`, `ScrollEndNotification` and replicate the drag handlers from `_BottomSheetState` and manipulate the animation controller.

lib/widgets/home.dart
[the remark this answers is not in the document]
PIG208: For the record, we decided to defer the feature to #1088.

lib/widgets/home.dart
```diff
@@ -0,0 +1,461 @@
...
+        shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(5))),
+      child: Text(value.toString(), style: TextStyle(
+        fontSize: 18,
+        height: 21 / 18,
+        color: textColor,
+      ).merge(weightVariableTextStyle(context, wght: 600))));
+  }
+}
+
+class _VersionInfo extends StatelessWidget {
```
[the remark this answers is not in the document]
PIG208: Opened #1095 as a follow-up.

lib/widgets/home.dart
```diff
@@ -2,96 +2,489 @@ import 'package:flutter/material.dart';
...
+      button(HomePageTab.channels, ZulipIcons.hash_italic),
+      // TODO(#1094): Users
+      button(HomePageTab.directMessages, ZulipIcons.user),
+      NavigationButton(icon: ZulipIcons.menu, selected: false,
+        onPressed: () => _showMenu(context, tabNotifier: _tab)),
+    };
+
+    final designVariables = DesignVariables.of(context);
     return Scaffold(
-      appBar: ZulipAppBar(title: const Text("Home")),
```
chrisbobbe: There's a top bar in the Figma:

image

Ahh but I see #1035 says "There's no top bar" in the issue description. Fair enough 🤷‍♂️ but let's mention it in the commit message's list of differences from the Figma. Also, now we don't have a spec for what goes in the top inset; here's this revision:

| Not scrolled | Scrolled a bit |
| --- | --- |
| image | image |

The top inset is filled with the same color we use under the content in the y-direction when the list is short. I don't love that; it's like the content just abruptly disappears, instead of sliding under some different surface (what used to be the app bar). I don't know what's best here, perhaps just leave a `TODO(design)` for how to fill the top inset on these pages.

chrisbobbe (commented):
I've read all the commits except the main commit:

```
nav: Add bottom tabs and main menu
```

which I've skimmed a bit. It's time for bed but I wanted to get a review in today. 🙂

Some things I noticed in manual testing:

- If I tap "Switch account" from the new bottom sheet, then go back (with the back button or iOS swipe gesture), the bottom sheet is still open. Nicer to close it when you tap that button, I think. (Same for some other buttons like Mentions.)

- If I do that after logging out the account I was looking at, I get an infinite loading spinner. This is what `PerAccountStoreWidget.routeToRemoveOnLogout` is meant to prevent. Can we use that directly, or do we need some changes to support auto-dismissing an action sheet on logout?

  image

  If I close that action sheet (by dragging down), what's supposed to happen? I get a black screen:

  image

- When you tap an account on the "Choose account" screen, I notice that a route is added to the stack, on top of the "Choose account" screen, which lets me go back to that screen. Instead, can we clear anything on the stack and replace it with just the bottom-tabs screen for the chosen account? The issue says 'there's no navigating "back" or "up" from [the screen with the bottom tab bar].'

assets/l10n/app_en.arb
```diff
@@ -15,8 +15,8 @@
   "@aboutPageTapToView": {
     "description": "Item subtitle in About Zulip page to navigate to Licenses page"
   },
-  "chooseAccountPageTitle": "Choose account",
-  "@chooseAccountPageTitle": {
+  "switchAccountPageTitle": "Switch account",
```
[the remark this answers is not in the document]
PIG208: > Also, what is "i10n", do you mean 'internationalization' ("intl") or localization ("l10n")?

Yeah, that's a typo.

assets/l10n/app_en.arb
```diff
@@ -529,6 +529,10 @@
   "@userRoleUnknown": {
     "description": "Label for UserRole.unknown"
   },
+  "inboxButton": "Inbox",
+  "@inboxButton": {
+    "description": "Label for the menu button switching to inbox."
```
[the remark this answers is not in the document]
PIG208: Updating this with the description altered to match the other existing page titles.

I guess this would also be a reason to also name the button as "Choose account" for the "Choose account" page.

lib/widgets/page.dart
```diff
@@ -159,9 +159,6 @@ class LoadingPlaceholderPage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Scaffold(
-      appBar: AppBar(),
-      body: const LoadingPlaceholder(),
-    );
+    return const Scaffold(body: LoadingPlaceholder());
```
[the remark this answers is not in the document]
gnprice: The main reason this loading page has an app bar is to make it possible to back out of it (on iOS where you don't have a system back gesture/button) so e.g. you can go use some other account.

With the new nav, backing out wouldn't get you to ChooseAccountPage anyway because HomePage is the root of the navigation. But it'll be important to still have some way to switch accounts from the loading page.

I think that means we want two things here:
* LoadingPlaceholderPage, which is used by AccountPageRouteMixin in general, should continue to have an app bar. That way if you wind up on a page that isn't HomePage, e.g. by opening a notification, and it's stuck loading, you can back out.
* HomePage.buildRoute should provide some alternative loading page that's specific to HomePage. That alternative will provide a way to switch accounts, which can be ad hoc like a button in the bottom right.

lib/widgets/home.dart
```diff
@@ -2,96 +2,489 @@ import 'package:flutter/material.dart';
...
-                  MessageListPage.buildRoute(context: context,
-                    narrow: ChannelNarrow(testStreamId!))),
-                child: const Text("#test here")), // scaffolding hack, see above
-            ],
+      body: Stack(
+        children: [
+          for (final (tab, body) in pageBodies)
+            Offstage(offstage: tab != _tab.value, child: body),
+        ]),
+      bottomNavigationBar: SafeArea(
```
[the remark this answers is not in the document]
PIG208: Ah, thanks for testing that out. I found out today that while my Android device has safe area, those area aren't visible at all. Maybe there's still a way to test it without an iOS device?

lib/widgets/home.dart
PIG208: Yeah, that will be a separate redesign project. I recalled that we should have an issue for the top bar other than #1039, but I couldn't find it.

gnprice: Let's have an app bar at the top after all. (It'll be pretty plain — just a title, I think.)

I'm not sure what my thinking was when I said in the issue there's no top bar. But I see that's a bit of a challenge what should happen on scroll, if there's no app bar; and I just went and looked through several apps on my phone that have a bottom nav bar, and all of them have an app bar on those screens. In a couple of cases the app bar goes away as you scroll (making more room for the scrolling content), but that's trickier; let's skip that at least for launch, and have the app bar remain fixed at the top on scroll.

chrisbobbe:
Test, getting 500s from GitHub when trying to post inline replies above. Ugh.

lib/widgets/home.dart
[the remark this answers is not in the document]
gnprice: > Maybe there's still a way to test it without an iOS device?

A good way to do that is to use the Android emulator. It has options to simulate a variety of shapes of insets.

assets/l10n/app_en.arb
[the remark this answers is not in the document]
chrisbobbe: > The documentation for `SemanticsProperties.namesRoute` doesn't seem quite clear on how it can be used. I assume that we need to use `Semantics` with it? Because this might be a non-trivial project on its own, I feel that it might fit better as a separate issue.

Yes, definitely fine to leave this for later 🙂:

> How about adding a `TODO(a11y)` to find out if we're meeting users' expectations when they use Zulip with a screen reader, mentioning `SemanticsProperties.namesRoute`?

You could file an issue if you want, but I think the TODO is fine and better than nothing.

lib/widgets/icons.dart
```diff
@@ -48,65 +48,77 @@ abstract final class ZulipIcons {
   /// The Zulip custom icon "clock".
   static const IconData clock = IconData(0xf108, fontFamily: "Zulip Icons");
 
+  /// The Zulip custom icon "contacts".
+  static const IconData contacts = IconData(0xf109, fontFamily: "Zulip Icons");
```
gnprice, line 52: nit:

    nav: Add icons from Figma
    
    The icons are taken from
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=544-22131&node-type=canvas&t=f4vYPq2CZUpFuf7V-0.

Write instead:

    icons: Add icons for bottom nav bar
    
    The icons are taken from Figma:
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=544-22131&node-type=canvas&t=f4vYPq2CZUpFuf7V-0

Avoiding `.` there avoids ambiguity about what the URL is.

The summary line can be more specific than "nav" about what the icons are; there are other icons involved in nav. And our usual prefix for editing icons is "icons:".

gnprice (commented):
To help accelerate things for this PR, here's a review of some commits that I think we can merge early. @PIG208 would you send a separate PR with these, after revision for the nit below and applicable comments by @chrisbobbe above?

48671b273 nav [nfc]: Extract *PageBody widgets
70777ba26 nav [nfc]: Move HomePage to a new home
848816262 action_sheet [nfc]: Centralize .withValues variants of contextMenuCancelBg
6d9037b49 nav: Add icons from Figma

In particular I think some of those are likely to be conflict-prone — icons changes always are, and theme changes often seem to be — so it's helpful to have them merged early.

test/widgets/home_test.dart
```diff
@@ -0,0 +1,155 @@
...
+      checkIconSelected(tester, channelsMenuIconFinder);
+    });
+
+    testWidgets('navigation menu buttons control navigation states', (tester) async {
+      await prepare(tester);
+
+      await tapOpenMenu(tester);
+      checkIconSelected(tester, inboxMenuIconFinder);
+      checkIconNotSelected(tester, channelsMenuIconFinder);
+      check(find.byType(InboxPageBody)).findsOne();
```
gnprice, line 122:
```suggestion
      check(find.byType(InboxPageBody)).findsOne();
      check(find.byType(SubscriptionListPageBody)).findsNothing();
```

and similarly `check(find.byType(InboxPageBody)).findsNothing();` below. That helps the reader be sure that something really changed as a result of the tap — that it's not just that both finders already would find something and continue to do so (e.g. because we got something wrong in the interaction of the finders with the `Offstage` nature of the pages).

gnprice (commented):
Further in the interest of accelerating things for this PR, here's an initial review from me of the main commit:
6e8de9178 nav: Add bottom tabs and main menu

I skipped points that are already discussed in Chris's review above.

lib/widgets/home.dart
```diff
@@ -0,0 +1,490 @@
...
+        child: DecoratedBox(
+          decoration: BoxDecoration(
+            border: Border(top: BorderSide(color: designVariables.borderBar, width: 0)),
+            color: designVariables.bgBotBar),
+          child: Align(
+            // This is necessary to limit the width of the row of icons,
+            // because [Align] allows the child to be smaller than itself.
+            heightFactor: 1,
```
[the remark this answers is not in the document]
PIG208: Moved this comment above and added another one explaining the `heightFactor`. Also changed it to use `Center` instead.

lib/widgets/home.dart
```diff
@@ -0,0 +1,490 @@
...
+                    mainAxisAlignment: MainAxisAlignment.start,
+                    children: menuItems)))),
+              Padding(
+                padding: const EdgeInsets.symmetric(horizontal: 16),
+                child: AnimatedScaleOnTap(
+                  scaleEnd: 0.95,
+                  duration: const Duration(milliseconds: 100),
+                  child: ConstrainedBox(
+                    constraints: const BoxConstraints(minHeight: 44),
+                    child: const ActionSheetCancelButton()))),
```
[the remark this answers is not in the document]
PIG208: With a smaller font, the button can get a bit shorter than that:

<details>
<summary>Small font</summary>

| w/ minHeight | w/o minHeight |
| - | - |
|  |  |


</details>

PIG208:
Started experimenting with the special placeholder and route changes.

This addresses most of the comment except
https://github.com/zulip/zulip-flutter/pull/1076#discussion_r1868262551 (special loading page) and https://github.com/zulip/zulip-flutter/pull/1076#pullrequestreview-2474679726 (navigation change) and https://github.com/zulip/zulip-flutter/pull/1076#discussion_r1868288529 (app bar). All of these mentioned features are included as WIP for now. Will get back to this tomorrow.

chrisbobbe:
The prep PR #1099 was merged, so this can be updated for that. (Also I see CI is failing.)

PIG208:
Yes. There are some wip changes to be finished before this is ready for review again.

PIG208:
Thanks for the reviews @chrisbobbe @gnprice! The PR is ready again.

Some UI changes:

<details>
<summary>HomePage's loading screen</summary>

Screenshot_20241205_010236

-
Screenshot_20241205_010241


</details>

<details>
<summary>HomePage's app bar</summary>

Screenshot_20241205_005926(1)

</details>

Now we ensure that the HomePage is always at the root level, by removing all the existing routes when pushing it onto the navigator stack.

There are also some minor changes like popping the bottom sheet for all menu buttons.

lib/widgets/home.dart
```diff
@@ -2,96 +2,533 @@ import 'package:flutter/material.dart';
...
+            onPressed: () {
+              Navigator.of(context).push(
+                MaterialWidgetRoute(page: const ChooseAccountPage()));
+            },
+            child: Text(zulipLocalizations.chooseAccountPageTitle)),
+        ]));
+  }
+}
+
+class _NavigationButton extends StatelessWidget {
```
gnprice, line 162: bump https://github.com/zulip/zulip-flutter/pull/1076#discussion_r1868535479

lib/widgets/home.dart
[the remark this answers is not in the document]
gnprice: (Are the labels switched on those screenshots? The one labeled "w/ minHeight" has the button shorter than the other does.)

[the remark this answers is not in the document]
gnprice: Following up on this point:

> I'd think TextButton (which this button uses) would take care of that.

by looking into TextButton, it looks like this is controlled by `ButtonStyle.minimumSize`. The default for which in `TextButton.defaultStyleOf` is:
```dart
  @override
  MaterialStateProperty<Size>? get minimumSize =>
    const MaterialStatePropertyAll<Size>(Size(64.0, 40.0));
```

That's a minimum of 64px wide but only 40px high. So TextButton does indeed take care of this need — but it sets a slightly smaller minimum.

What if we set `TextButton.style` to provide a `minimumSize` property there?

gnprice (commented):
Thanks for the revision!

It's late, so for tonight here's just a review of the points that were in my previous review https://github.com/zulip/zulip-flutter/pull/1076#pullrequestreview-2477006695 . All look good now except for the couple of items below.

lib/widgets/home.dart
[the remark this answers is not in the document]
PIG208: > What if we set TextButton.style to provide a minimumSize property there?

Yeah, that should be cleaner.

<details>
<summary>Updated screenshots</summary>


| font size | before | after |
| - | - | - |
| extra small | small_without_fix | small_with_fix
| normal |  normal | / |

</details>

lib/widgets/home.dart
PIG208: I see. Missed the `Bar` part. Fixed that and updated comment/string references to this name to match.

assets/l10n/app_en.arb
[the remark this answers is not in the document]
PIG208: I found #535 relevant to this:

> We should try out the app with iOS's VoiceOver and/or Android's TalkBack, and make sure it's possible to do the essentials to be able to use the app at all:

> - get between the different top-level pages (Inbox, All messages, Subscribed streams, Direct messages)

Except that the top-level pages become the tabs on the new home screen. With an app bar it might become less of a concern, but still it would be helpful to experiment with this before launch.

test/widgets/login_test.dart
```diff
@@ -118,30 +123,71 @@ void main() {
...
+          (it) => it.isA<MaterialAccountWidgetRoute>()
+            ..accountId.equals(eg.selfAccount.id)
+            ..page.isA<HomePage>(),
+          (it) => it.isA<MaterialWidgetRoute>()
+            ..page.isA<LoginPage>(),
+        ]);
+        pushedRoutes.clear();
+
+        await login(tester, eg.otherAccount);
+        final loggedInAccount = testBinding.globalStore.accounts.singleWhere(
```
gnprice, line 168: nit: "logged-in account" is a confusing name here, because both accounts are logged in.

```suggestion
        final newAccount = testBinding.globalStore.accounts.singleWhere(
```

lib/widgets/home.dart
```diff
@@ -2,96 +2,532 @@ import 'package:flutter/material.dart';
...
-                onPressed: () => Navigator.push(context,
-                  MessageListPage.buildRoute(context: context,
-                    narrow: ChannelNarrow(testStreamId!))),
-                child: const Text("#test here")), // scaffolding hack, see above
-            ],
+      body: Column(
+        crossAxisAlignment: CrossAxisAlignment.end,
+        children: [
+          const Expanded(child: LoadingPlaceholder()),
+          TextButton(
```
gnprice, line 151: This needs to use SafeArea to stay out of the insets. Here's how it looks on my phone:
image

gnprice: Probably cleanest is to make this a floating action button:
```dart
    return Scaffold(
      body: const LoadingPlaceholder(),
      floatingActionButton: TextButton(
        onPressed: () => Navigator.push(context,
          MaterialWidgetRoute(page: const ChooseAccountPage())),
        child: Text(zulipLocalizations.chooseAccountPageTitle)),
    );
```

That takes care of applying SafeArea; and it also means the layout of LoadingPlaceholder and centering of the circular progress indicator doesn't get distorted by the presence of the button.

gnprice (commented):
Thanks for the revision! I played with this UI and am happy to see that the behavior seems to be basically all working.

Partial review below, including one behavior issue in the switch-account button being potentially inaccessible. Posting the partial review now in the interest of pipelining work on this PR.

Another thing I noticed in the UI is that the menu takes up the full height of the screen even when its contents are much shorter. Is that something Vlad specifically addressed in his design? My instinct would be to let the menu be the height of its contents when there's room, and fill the screen only when it needs to.

lib/widgets/home.dart
gnprice: I went to go update #1035 with this change, and rereading the context of that sentence I realized what happened: what I really meant was that there's no top *tab bar*. I probably didn't consider at that moment the question of whether there should or shouldn't be a plain title-bearing app bar instead.

Anyway, fixed there.

test/widgets/app_test.dart
```diff
@@ -23,39 +24,80 @@ void main() {
...
-      check(await initialRoutes(tester)).deepEquals(<Condition<Object?>>[
-        (it) => it.isA<WidgetRoute>().page.isA<ChooseAccountPage>(),
+      check(pushedRoutes).deepEquals(<Condition<Object?>>[
+        (it) => it.isA<MaterialAccountWidgetRoute>()
+          ..accountId.equals(eg.selfAccount.id)
+          ..page.isA<HomePage>(),
+      ]);
+    });
+
+    testWidgets('choosing an account clears the navigator stack', (tester) async {
```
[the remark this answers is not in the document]
PIG208: Moving this to the "ChooseAccountPage" group.

lib/widgets/home.dart
PIG208: In this revision I also added the `AppBar` back because `HomePage` has gained one in the previous update.

PIG208:
> Another thing I noticed in the UI is that the menu takes up the full height of the screen even when its contents are much shorter. Is that something Vlad specifically addressed in his design? My instinct would be to let the menu be the height of its contents when there's room, and fill the screen only when it needs to.

Yeah, it's mentioned [here](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2957-52907&node-type=frame&t=LagtQe63trPehowh-0):

> Menu slides up to full screen state. There is not medium state, and no dragging icon. 

I interpret this as "the menu always fills the entire screen".

lib/widgets/home.dart
```diff
@@ -2,96 +2,528 @@ import 'package:flutter/material.dart';
...
+      strokeAlign: BorderSide.strokeAlignOutside,
+      color: designVariables.borderMenuButtonSelected);
+    final buttonStyle = TextButton.styleFrom(
+      padding: const EdgeInsets.symmetric(vertical: 9, horizontal: 8),
+      foregroundColor: designVariables.labelMenuButton,
+      splashFactory: NoSplash.splashFactory,
+      shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(10)),
+    ).copyWith(
+      backgroundColor: WidgetStateColor.fromMap({
+        WidgetState.pressed: designVariables.bgMenuButtonActive,
```
chrisbobbe, line 308: In dark mode, the pressed button background `bgMenuButtonActive` is black with 20% opacity, and it's on a `#222222` background (`bgBotBar`). So it should be darker than the background. But it's not:

image

…well actually it is 🙂 but only for a frame or two before an animation makes it light like this. I think the culprit is a default `overlayColor` value or something like that. I notice that it gets darker if I pass `overlayColor: Colors.transparent`; that may or may not be the solution but it's an observation.

chrisbobbe (changes requested):
Thanks! Here's another not-quite-thorough review before my bedtime. I still haven't read the tests in the main commit.

I notice you have a commit marked "fixup!" so it's possible this wasn't ready for another review?

I notice this isn't ready to support the situation where the app logs you out without you requesting it in the UI:

https://github.com/zulip/zulip-flutter/pull/1076#pullrequestreview-2474679726
> This way might be nice because it also supports cases where you logged out without requesting it in the UI, like (not yet implemented) when we learn your auth is invalid.

I tested by just changing the "Inbox" menu's press handler so that it called `logOutAccount`, and when I pressed it, it put me on a blank black screen. Since that case isn't currently reachable (right now the only way to be logged out is to request it from the choose-account page), it might be fine to leave for a followup.

I remember one of your commits said this is just a partial implementation, with more things planned as followups—do we have a list of those things somewhere?

lib/widgets/home.dart
```diff
@@ -2,96 +2,528 @@ import 'package:flutter/material.dart';
...
-                    narrow: ChannelNarrow(testStreamId!))),
-                child: const Text("#test here")), // scaffolding hack, see above
-            ],
+      appBar: AppBar(),
+      body: const LoadingPlaceholder(),
+      floatingActionButton: TextButton(
+        onPressed: () => Navigator.push(context,
+          MaterialWidgetRoute(page: const ChooseAccountPage())),
+        child: Text(zulipLocalizations.chooseAccountPageTitle)));
+  }
```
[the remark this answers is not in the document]
gnprice: Yeah, "Choose account" isn't good text here. You made the same point about the button in the menu and I agree. Better text would be "Switch account", or perhaps like "Try another account".

The reason something like this is needed is in case the data for this account takes a long time to load — perhaps isn't ever going to load, because the server is down — so the user can still use some other account if that's what they wanted.

lib/widgets/app.dart
```diff
@@ -209,11 +208,10 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
             return [
-              MaterialWidgetRoute(page: const ChooseAccountPage()),
-              if (initialAccountId != null) ...[
+              if (initialAccountId != null)
                 HomePage.buildRoute(accountId: initialAccountId),
-                InboxPage.buildRoute(accountId: initialAccountId),
-              ],
+              if (initialAccountId == null)
+                MaterialWidgetRoute(page: const ChooseAccountPage()),
             ];
```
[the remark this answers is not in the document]
PIG208: Ah, I tried that before and was wondering why it wasn't working. It turned out that I had the comma at the end of the if expression.

lib/widgets/home.dart
PIG208: I found it helpful to look at the defaults:

```
  /// ## Material 3 defaults
  ///
  /// If [ThemeData.useMaterial3] is set to true the following defaults will
  /// be used:
  ///
  /// {@template flutter.material.text_button.material3_defaults}
  /// * `textStyle` - Theme.textTheme.labelLarge
  /// * `backgroundColor` - transparent
  /// * `foregroundColor`
  ///   * disabled - Theme.colorScheme.onSurface(0.38)
  ///   * others - Theme.colorScheme.primary
  /// * `overlayColor`
  ///   * hovered - Theme.colorScheme.primary(0.08)
  ///   * focused or pressed - Theme.colorScheme.primary(0.1)
  ///   * others - null
  /// * `shadowColor` - Colors.transparent,
```
I guess visually the `pressed` state got affected because `overlayColor` is `Theme.colorScheme.primary(0.1)` by default. We will lose the color distinction for the `hovered`/`focused` state, but we can still control them if we need to.

lib/widgets/home.dart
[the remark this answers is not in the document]
PIG208: In addition to using "Try another account", does it look better as a `FilledButton` or an `OutlinedButton`? (so that user knows that it is a button)

| `TextButton` | `OutlinedButton` | `FilledButton` |
| - | - | - |
| Screenshot from 2024-12-06 15-09-11 | Screenshot from 2024-12-06 15-09-03 | Screenshot from 2024-12-06 15-08-51 |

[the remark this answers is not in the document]
chrisbobbe: Yeah, I think the button's current appearance doesn't really read as a button.

Maybe a nicer alternative (might take longer to implement) would be to wait 5 seconds or so before showing anything, and have it be a message (centered in the page?) like "This is taking longer than usual to load. Try another account?" That way in the happy-path case we don't distract the user by making them figure out UI that they won't need.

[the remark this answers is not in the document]
PIG208: Started a discussion [here](https://chat.zulip.org/#narrow/channel/48-mobile/topic/.23F1076.20accessing.20choose-account.20page.20from.20the.20loading.20screen).

lib/widgets/home.dart
```diff
@@ -2,96 +2,528 @@ import 'package:flutter/material.dart';
...
+  final accountId = PerAccountStoreWidget.accountIdOf(context);
+  showModalBottomSheet<void>(
+    context: context,
+    // Clip.hardEdge looks bad; Clip.antiAliasWithSaveLayer looks pixel-perfect
+    // on my iPhone 13 Pro but is marked as "much slower":
+    //   https://api.flutter.dev/flutter/dart-ui/Clip.html
+    clipBehavior: Clip.antiAlias,
+    useSafeArea: true,
+    isScrollControlled: true,
+    backgroundColor: designVariables.bgBotBar,
```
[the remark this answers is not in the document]
gnprice: Hmm yeah, that's a bug.

I don't see a good way to avoid it with the existing API of `showModalBottomSheet` and friends, though. (If we wanted the same custom background color for all our bottom sheets, or all our modal bottom sheets, we could do that with ThemeData.bottomSheetTheme. But this has a different background color from the message/topic/etc. action sheets.) So let's file that as a follow-up issue for post-launch.

lib/widgets/page.dart
```diff
@@ -152,6 +159,9 @@ class AccountPageRouteBuilder<T extends Object?> extends PageRouteBuilder<T> wit
 
   @override
   final int accountId;
+
+  @override
+  Widget? loadingPlaceholderPage;
```
gnprice, line 164:
```suggestion
  final Widget? loadingPlaceholderPage;
```

lib/widgets/home.dart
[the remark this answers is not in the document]
PIG208: If we do not pass `backgroundColor` on the `showModalBottomSheet` call, we can set the background color on a `DecoratedBox` and have the `DesignVariables.of(context)`'s inline, or do something like:

```dart
  showModalBottomSheet<void>(
    // [...]
    builder: (BuildContext context) {
      return PerAccountStoreWidget(
        accountId: accountId,
        child: DecoratedBox(
          decoration: BoxDecoration(color: designVariables.bgBotBar),
```

This does not set the color of the bottom sheet `Material`. So we also have to set `badckgroundColor` to something like `transparent` to avoid rendering artifacts:

image

(with `backgroundColor: Colors.blue`)

test/widgets/actions_test.dart
```diff
@@ -182,8 +182,10 @@ void main() {
 
       final pushedRoutes = <Route<dynamic>>[];
       testNavObserver.onPushed = (route, prevRoute) => pushedRoutes.add(route);
-      final account1Route = InboxPage.buildRoute(accountId: account1.id);
-      final account2Route = InboxPage.buildRoute(accountId: account2.id);
+      final account1Route = MaterialAccountWidgetRoute(
+        accountId: account1.id, page: const InboxPageBody());
+      final account2Route = MaterialAccountWidgetRoute(
+        accountId: account2.id, page: const InboxPageBody());
```
gnprice, line 188: These pages don't exist in real life. That rather undermines this test, because it's all about how the navigation stack will behave when one account gets logged out.

… In fact I think contemplating what a version of this test would look like that's updated to be realistic after this PR leads to rediscovering the bug that @chrisbobbe mentioned at https://github.com/zulip/zulip-flutter/pull/1076#pullrequestreview-2474679726 and https://github.com/zulip/zulip-flutter/pull/1076#pullrequestreview-2483685141 .

As Chris said in that latter comment (and I agree), that issue can be left as a followup, because it's not a live bug until we make other changes in the future. So I guess this test can just get a TODO comment pointing at that followup issue.

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -49,7 +50,7 @@ Future<void> setupPage(WidgetTester tester, {
   await tester.pumpWidget(TestZulipApp(
     accountId: eg.selfAccount.id,
     navigatorObservers: navigatorObserver != null ? [navigatorObserver] : [],
-    child: const RecentDmConversationsPage()));
+    child: const HomePage(initialTab: HomePageTab.directMessages)));
```
gnprice, line 53: How about making this just `HomePage()`, and following with a `tester.tap(find.byIcon(…))` to get to the right tab?

That'd be more realistic, and (if we do the same for the subscription/channels tab below) would let us avoid having this `initialTab` field that's used only in tests.

lib/widgets/app.dart
```diff
@@ -209,11 +208,10 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
           onGenerateInitialRoutes: (_) {
             return [
-              MaterialWidgetRoute(page: const ChooseAccountPage()),
-              if (initialAccountId != null) ...[
+              if (initialAccountId != null)
                 HomePage.buildRoute(accountId: initialAccountId),
-                InboxPage.buildRoute(accountId: initialAccountId),
-              ],
+              if (initialAccountId == null)
+                MaterialWidgetRoute(page: const ChooseAccountPage()),
```
gnprice, line 214: nit: put the no-account case first; it's the case that always happens first chronologically, and it's also the unusual case so going first is consistent with our usual early-return structure

assets/l10n/app_en.arb
```diff
@@ -529,6 +533,10 @@
   "@userRoleUnknown": {
     "description": "Label for UserRole.unknown"
   },
+  "inboxPageTitle": "Inbox",
+  "@inboxPageTitle": {
+    "description": "Title for the page of inbox."
```
gnprice, line 538: This description doesn't read very well to me. But then I realized the same is true of some of the existing descriptions of similar strings, so I sent a quick PR fixing those:
* #1109 

This can then follow the style of those.

assets/l10n/app_en.arb
```diff
@@ -19,6 +19,10 @@
   "@chooseAccountPageTitle": {
     "description": "Title for ChooseAccountPage"
   },
+  "switchAccountButton": "Switch account",
+  "@switchAccountButton": {
+    "description": "Main menu button for the choose-account page"
```
gnprice, line 24:
```suggestion
    "description": "Label for main-menu button leading to the choose-account page."
```

The string isn't a button, but rather the label for a button.

Separately, without the hyphen it reads to me a bit more like: it's (a label for) a menu button for the choose-account page, namely the main such button.

assets/l10n/app_en.arb
```diff
@@ -545,6 +553,14 @@
   "@starredMessagesPageTitle": {
     "description": "Title for the page of starred messages."
   },
+  "channelsPageTitle": "Channels",
+  "@channelsPageTitle": {
+    "description": "Title for the page of channels."
+  },
+  "profilePageTitle": "My profile",
+  "@profilePageTitle": {
+    "description": "Title for the page of the logged in user's profile."
```
gnprice, line 562: This name and description is inaccurate — the PR doesn't use this string for the title of the page. The page's title is what appears in the app bar, which for this page is the user's name.

Instead a good description would be:
> Label for main-menu button leading to the user's own profile.

And for the name… perhaps `mainMenuMyProfile`.

lib/widgets/app.dart
```diff
@@ -268,17 +266,18 @@ class ChooseAccountPage extends StatelessWidget {
...
-          HomePage.buildRoute(accountId: accountId))));
+        onTap: () => HomePage.navigate(context, accountId: accountId)));
   }
 
   @override
   Widget build(BuildContext context) {
     final zulipLocalizations = ZulipLocalizations.of(context);
     assert(!PerAccountStoreWidget.debugExistsOf(context));
     final globalStore = GlobalStoreWidget.of(context);
+    final hasBackButton = ModalRoute.of(context)?.impliesAppBarDismissal ?? false;
```
gnprice, line 277: This line is pretty mysterious — let's give it a one- or two-line comment pointing at how you determined this was the right way to compute whether there'll be a back button.

lib/widgets/home.dart
```diff
@@ -1,97 +1,539 @@
...
   @override
-  Widget build(BuildContext context) {
-    final store = PerAccountStoreWidget.of(context);
+  State<HomePage> createState() => _HomePageState();
+}
+
+class _HomePageState extends State<HomePage> {
+  late final ValueNotifier<HomePageTab> _tab = ValueNotifier(widget.initialTab);
+
+  String get _currentTabTitle {
```
gnprice, line 55: nit: put this just above `build`, after the state management; it's effectively a private helper for `build`

lib/widgets/home.dart
```diff
@@ -1,97 +1,539 @@
...
+        child: SafeArea(
+          // [Center] is necessary to limit the width of the row of icons,
+          // because it allows the child to be smaller than itself.
+          child: Center(
+            // This shrinks [Center] to be as tall as its child, instead of
+            // growing as tall as [Scaffold] allows, filling the entire screen.
+            heightFactor: 1,
+            child: ConstrainedBox(
+              // TODO(design): determine a suitable max width for bottom nav bar
+              constraints: const BoxConstraints(maxWidth: 600).tighten(height: 48),
```
gnprice, line 138: It's kept bugging me that this feels complicated, so I fiddled a bit to find a simpler version:
```suggestion
          child: SizedBox(height: 48,
            child: Center(
              child: ConstrainedBox(
                // TODO(design): determine a suitable max width for bottom nav bar
                constraints: const BoxConstraints(maxWidth: 600),
```

(The `CrossAxisAlignment.stretch` means the child will indeed fill those 48px of height, and therefore so will the Center widget)

lib/widgets/home.dart
```diff
@@ -1,97 +1,539 @@
...
+          // TODO(#417): Disable splash effects for all buttons globally.
+          splashFactory: NoSplash.splashFactory,
+          highlightColor: designVariables.iconSelected.withValues(alpha: 0.05),
+          shape: const RoundedRectangleBorder(
+            borderRadius: BorderRadius.all(Radius.circular(4))),
+        ).copyWith(foregroundColor: iconColor)));
+  }
+}
+
+void _showMenu(BuildContext context, {
```
gnprice, line 203: nit:
```suggestion
void _showMainMenu(BuildContext context, {
```

Makes it clearer up front which menu it is that this function shows.

lib/widgets/home.dart
```diff
@@ -1,97 +1,539 @@
...
+      ProfilePage.buildRoute(context: context, userId: store.selfUserId));
+  }
+}
+
+class _SwitchAccountButton extends _MenuButton {
+  const _SwitchAccountButton();
+
+  @override
+  // TODO(design): choose an icon
+  IconData get icon => ZulipIcons.globe;
```
gnprice, line 482: Yeah, not `globe` because that's the icon we use to mean "web-public channel".

The TODO can remain, but pending that it'd be better to just make this blank — have `buildLeading` return `SizedBox.shrink()`.

lib/widgets/home.dart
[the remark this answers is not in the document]
gnprice: Yeah, and the `Material` also gets involved (and uses its color) in a variety of interactions, which we'd effectively be turning off. So that feels like a workaround approach that's likely to have subtle bugs of its own. I'm OK with just having that glitch if the theme happens to change just at the moment that you have the main menu open, and leaving the fix for it as a future issue.

PIG208:
(will update tests for the loading screen shortly)

PIG208:
<details>
<summary>Light mode</summary>

| before | after |
| - | - |
| light-1 | light-1 |
| light-2 | light-2 |
| light-3 | light-3 |
| light-4 | light-4 |
| light-5 | light-5 |
| light-6 | light-6 |

</details>

<details>
<summary>Dark mode</summary>

| before | after |
| - | - |
|  dark-1 | dark-1 |
| dark-2 | dark-2 |
| dark-4 | dark-3 |
| dark-3 | dark-4 |
| dark-5 | dark-5 |
| dark-6 | dark-6 |


</details>

PIG208:
Looks like `SafeArea` was dropped (`git diff b92737fe..0037df92 -- lib/widgets/home.dart`). Adding it back.

lib/widgets/theme.dart
```diff
@@ -82,10 +89,7 @@ ThemeData zulipThemeData(BuildContext context) {
     //   https://api.flutter.dev/flutter/material/ColorScheme/ColorScheme.fromSeed.html
     // Or try this tool to see the whole palette:
     //   https://m3.material.io/theme-builder#/custom
-    colorScheme: ColorScheme.fromSeed(
-      brightness: brightness,
-      seedColor: kZulipBrandColor,
-    ),
+    colorScheme: colorScheme,
```
gnprice, line 92: nit: the comment above here is about the construction of the ColorScheme, so should move along with that constructor call

test/widgets/actions_test.dart
gnprice: bump: should get that TODO comment

gnprice (commented):
Thanks for the revision!

I did a quick re-review of mostly just the items that were in my last review; comments from that below, and otherwise those items all look good.

(These are minor and this is definitely lower-priority than the investigation you're doing at the moment on #924 — posting it now just to flush the state from my head and my GitHub pending comments.)

assets/l10n/app_en.arb
```diff
@@ -17,7 +17,20 @@
...
   "chooseAccountPageTitle": "Choose account",
   "@chooseAccountPageTitle": {
-    "description": "Title for ChooseAccountPage"
+    "description": "Title for the page to choose between Zulip accounts."
+  },
+  "switchAccountButton": "Switch account",
+  "@switchAccountButton": {
+    "description": "Label for main-menu button leading to the choose-account page."
+  },
+  "tryAnotherAccountMessage": "You account at {url} is taking a while to load.",
```
gnprice, line 26:
```suggestion
  "tryAnotherAccountMessage": "Your account at {url} is taking a while to load.",
```

assets/l10n/app_en.arb
```diff
@@ -17,7 +17,20 @@
...
+  "@tryAnotherAccountMessage": {
+    "description": "Message that appears on the loading screen after waiting for some time.",
+    "url": {"type": "String", "example": "http://example.com/"}
```
gnprice, line 29: nit:
```suggestion
  "tryAnotherAccountMessage": "You account at {url} is taking a while to load.",
  "@tryAnotherAccountMessage": {
    "description": "Message that appears on the loading screen after waiting for some time.",
    "url": {"type": "String", "example": "http://chat.example.com/"}
```

lib/widgets/app.dart
```diff
@@ -268,41 +266,50 @@ class ChooseAccountPage extends StatelessWidget {
...
         // IconButton (the "…" button) comes with 12px padding on all sides.
         contentPadding: const EdgeInsetsDirectional.only(start: 16, end: 12),
-        onTap: () => Navigator.push(context,
-          HomePage.buildRoute(accountId: accountId))));
+        onTap: () => HomePage.navigate(context, accountId: accountId)));
   }
 
   @override
   Widget build(BuildContext context) {
+    final colorScheme = ColorScheme.of(context);
```
gnprice, line 274: This commit also edits colors for the individual `ListTile`s above:
```diff
       child: ListTile(
         title: title,
         subtitle: subtitle,
+        tileColor: colorScheme.secondaryContainer,
+        textColor: colorScheme.onSecondaryContainer,
```

but then the main commit reverts that.

It doesn't look like the main commit is making some other change that compensates for removing it, so that change probably doesn't belong in the main commit. Are those lines not actually needed in the first place?

lib/widgets/app.dart
```diff
@@ -268,41 +266,50 @@ class ChooseAccountPage extends StatelessWidget {
...
-                onPressed: () => Navigator.push(context,
-                  AddAccountPage.buildRoute()),
-                child: Text(zulipLocalizations.chooseAccountButtonAddAnAccount)),
-            ]))),
-      ));
+    // Borrowed from [AppBar.build].
+    // See documentation on [ModalRoute.impliesAppBarDismissal]:
+    /// > Whether an [AppBar] in the route should automatically add a back button or
+    /// > close button.
+    final hasBackButton = ModalRoute.of(context)?.impliesAppBarDismissal ?? false;
```
gnprice, line 282: nit: set off this stanza with blank lines:
```suggestion

    // Borrowed from [AppBar.build].
    // See documentation on [ModalRoute.impliesAppBarDismissal]:
    /// > Whether an [AppBar] in the route should automatically add a back button or
    /// > close button.
    final hasBackButton = ModalRoute.of(context)?.impliesAppBarDismissal ?? false;

```

In particular that helps make clear what code the comment is describing.

gnprice: nit: those `///` should be `//`, right?

gnprice (commented):
And here's a full review… up to partway through the main commit, just before starting on `home.dart`. (So not covering the bulk of the PR.) Posting now for pipelining, as today's a day where I'm switching between multiple tasks.

lib/widgets/app.dart
PIG208: Those lines should be kept to fix the appearance of the account cards on choose-acocunt page.

alya:
The "Log in with X" buttons look really squashed together in the screenshots above. Should we add a bit of vertical space?

PIG208:
Looks like we are not getting the padded buttons on the Linux build because the buttons are padded to 48px in size by default on mobile devices:


<details>
<summary>Details</summary>

| light | dark |
| - | - |
|  |  |

</details>

We can override that default so it is padded explicitly on all platforms.

test/widgets/app_test.dart
```diff
@@ -162,6 +159,45 @@ void main() {
         ..bottom.isLessThan(2 / 3 * screenHeight);
     });
 
+    testWidgets('choosing an account clears the navigator stack', (tester) async {
+      addTearDown(testBinding.reset);
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+      await testBinding.globalStore.add(eg.otherAccount, eg.initialSnapshot());
+
+      final pushedRoutes = <Route<dynamic>>[];
+      final poppedRoutes = <Route<dynamic>>[];
```
gnprice, line 168: nit: avoid `dynamic`:
```suggestion
      final pushedRoutes = <Route<void>>[];
      final poppedRoutes = <Route<void>>[];
```

test/widgets/login_test.dart
```diff
@@ -118,30 +124,57 @@ void main() {
...
+        check(pushedRoutes).isEmpty();
+        check(testBinding.globalStore.accounts).isEmpty();
+
+        await login(tester, eg.selfAccount);
         check(testBinding.globalStore.accounts).single
           .equals(eg.selfAccount.copyWith(
             id: testBinding.globalStore.accounts.single.id));
       });
 
+      testWidgets('logging into a different account', (tester) async {
```
gnprice, line 154: nit:
```suggestion
      testWidgets('logging into a second account', (tester) async {
```

Otherwise it raises the question, different from what? The point of this particular test is that it's one where there was already an account when logging into another one.

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+import '../flutter_checks.dart';
+import '../model/binding.dart';
+import '../model/test_store.dart';
+import 'page_checks.dart';
+import 'test_app.dart';
+
+void main () {
+  TestZulipBinding.ensureInitialized();
+
+  late PerAccountStore store;
```
gnprice, line 28: nit: make this final non-late, inside `prepare` — we don't actually use having it shared like this

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+    await store.addUsers([eg.selfUser, eg.otherUser]);
+    final stream = eg.stream();
+    await store.addStream(stream);
+    await store.addSubscription(eg.subscription(stream));
+
+    await store.handleEvent(MessageEvent(
+      id: 0, message: eg.streamMessage(stream: stream, sender: eg.otherUser)));
```
gnprice, line 41: This feels awfully specific. Is it generally needed?

Looks like just one test fails if I delete it. Can that test do this for itself?

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+
+    await tester.pumpWidget(TestZulipApp(
+      accountId: eg.selfAccount.id, child: const HomePage()));
+    await tester.pumpAndSettle();
```
gnprice, line 45: Avoid pumpAndSettle:
```suggestion
    await tester.pump();
```

https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-using-pumpandsettle

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+        matching: find.text('Direct messages'))).findsOne();
+    });
+  });
+
+  group('menu', () {
+    final designVariables = DesignVariables.light();
+
+    final inboxMenuIconFinder = find.descendant(
+      of: find.byType(SingleChildScrollView),
+      matching: find.byIcon(ZulipIcons.inbox));
```
gnprice, line 100: How about:
```suggestion
    final inboxMenuIconFinder = find.descendant(
      of: find.byType(BottomSheet),
      matching: find.byIcon(ZulipIcons.inbox));
```

That feels a bit closer to the UI as seen by the user. After all, for users not on small screens the menu doesn't even scroll.

(The point of the ancestor condition IIUC is to exclude the corresponding icon in the nav bar itself.)

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+
+    void checkOnChooseAccountPage() {
+      // Ignore the possible loading page in the background.
+      check(find.byType(CircularProgressIndicator).hitTestable()).findsNothing();
+      check(find.byType(ChooseAccountPage)).findsOne();
+      check(find.byType(HomePage)).findsNothing();
+    }
+
+    ModalRoute<dynamic>? getRouteOf(WidgetTester tester, Finder elementFinder) =>
+      ModalRoute.of(tester.element(elementFinder));
```
gnprice, line 236: nit: it's just the normal type of finder, no need to be more specific than usual:
```suggestion
    ModalRoute<void>? getRouteOf(WidgetTester tester, Finder finder) =>
      ModalRoute.of(tester.element(finder));
```

(and separately, avoid dynamic)

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+      await tapChooseAccount(tester);
+
+      await tester.tap(find.byType(BackButton));
+      await tester.pump(Duration.zero); // tap the button
+      await tester.pump(const Duration(milliseconds: 350)); // wait for pop animation
+      checkOnLoadingPage();
+
+      await tester.pump(loadPerAccountDuration);
+      checkOnHomePage(tester, expectedAccount: eg.selfAccount);
+    });
```
gnprice, line 305: I like these nice clear tests! The various helper functions really contribute to being able to tell these different stories cleanly.

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+      testBinding.globalStore.loadPerAccountDuration = loadPerAccountDuration * 2;
+      await chooseAccountWithEmail(tester, eg.otherAccount.email);
+
+      // The second loadPerAccount is still pending.
+      await tester.pump(loadPerAccountDuration);
+      checkOnLoadingPage();
+
+      // The second loadPerAccount finished.
+      await tester.pump(loadPerAccountDuration);
+      checkOnHomePage(tester, expectedAccount: eg.otherAccount);
```
gnprice, line 322: nit: comment that says something is true, in the present tense, should go at a point where it's become true

```suggestion
      await tester.pump(loadPerAccountDuration);
      // The second loadPerAccount finished.
      checkOnHomePage(tester, expectedAccount: eg.otherAccount);
```

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+      await prepare(tester);
+      await tester.pump(kTryAnotherAccountWaitPeriod);
+      await tapChooseAccount(tester);
+
+      testBinding.globalStore.loadPerAccountDuration = loadPerAccountDuration * 2;
+      await chooseAccountWithEmail(tester, eg.otherAccount.email);
+
+      // The second loadPerAccount is still pending.
+      await tester.pump(loadPerAccountDuration);
+      checkOnLoadingPage();
```
gnprice, line 318: nit: similarly:
```suggestion
      await tester.pump(loadPerAccountDuration);
      // The second loadPerAccount is still pending.
      checkOnLoadingPage();
```

Here the statement in the comment was already true before the pump, but it wasn't interesting — of course it's still pending, only a few seconds have passed. It's after the await that that fact becomes interesting (which corresponds to the fact that that's the point where you're checking it).

test/widgets/home_test.dart
```diff
@@ -0,0 +1,409 @@
...
+
+      // The second loadPerAccount finished.
+      await tester.pump(loadPerAccountDuration);
+      checkOnHomePage(tester, expectedAccount: eg.otherAccount);
+    });
+
+    testWidgets('while loading, choosing an account disallow going back', (tester) async {
```
gnprice, line 325: nit:
```suggestion
    testWidgets('while loading, choosing an account disallows going back', (tester) async {
```

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -49,10 +50,16 @@ Future<void> setupPage(WidgetTester tester, {
...
-    child: const RecentDmConversationsPage()));
+    child: const HomePage()));
 
   // global store, per-account store, and page get loaded
   await tester.pumpAndSettle();
+
+  // Switch to direct messages tab.
+  await tester.tap(find.descendant(
+    of: find.byType(Center),
+    matching: find.byIcon(ZulipIcons.user)));
```
gnprice, line 61: Hmm, the use of `Center` is awkward — it's rather an implementation detail, both the fact that the desired copy of the icon does have a Center ancestor and the fact that the undesired copies don't.

I don't have a better idea to suggest, though, so I guess just leave it. No need for a TODO, even; really it just becomes actionable if the test breaks (because of some innocent change to either the nav bar, or how DMs appear in the inbox, or something else) and we have to find a different solution.

I feel like the API I'd want here is to be able to say "tap the such-and-such icon, specifically the one that's in the bottom part of the screen". But I don't see an easy way to do that with the existing `Finder` API.

test/widgets/home_test.dart
PIG208: It's for preparing a message in the inbox. We can either access `PerAccountStore` from that test, keep the `store.handleEvent` inside, or find a different state to test.

gnprice: Yeah, accessing the store from that test sounds good.

gnprice: (and then I guess it'll stay a late, shared variable)

lib/widgets/home.dart
```diff
@@ -1,97 +1,597 @@
...
+      appBar: AppBar(),
+      body: Center(
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            const CircularProgressIndicator(),
+            IgnorePointer(
+              ignoring: !showTryAnotherAccount,
+              child: Opacity(opacity: showTryAnotherAccount ? 1 : 0,
+                child: Padding(
```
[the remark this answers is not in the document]
PIG208: Needed to change this to:
```dart
            Visibility(
              visible: showTryAnotherAccount,
              maintainSize: true,
              maintainAnimation: true,
              maintainState: true,
```

```
  /// Whether to maintain space for where the widget would have been.
  ///
  /// To set this, [maintainAnimation] and [maintainState] must also be set.
  /// [...]
  final bool maintainSize;
```

test/widgets/home_test.dart
PIG208: Skipping this for https://github.com/zulip/zulip-flutter/pull/1076#discussion_r1879243089.

test/widgets/recent_dm_conversations_test.dart
PIG208: Yeah, I was a bit reluctant when writing this, but some alternatives would be customizing the initial view, adding a debug feature, or anything that only gets used in tests like this, which are not really realistic/useful. Leaving it seems fine to me.

test/widgets/actions_test.dart
```diff
@@ -182,8 +182,11 @@ void main() {
 
       final pushedRoutes = <Route<dynamic>>[];
       testNavObserver.onPushed = (route, prevRoute) => pushedRoutes.add(route);
-      final account1Route = InboxPage.buildRoute(accountId: account1.id);
-      final account2Route = InboxPage.buildRoute(accountId: account2.id);
+      // TODO(#737): switch to a realistic setup:
+      final account1Route = MaterialAccountWidgetRoute(
```
gnprice, line 186: bump https://github.com/zulip/zulip-flutter/pull/1076#discussion_r1879099666 :slightly_smiling_face:

lib/widgets/home.dart
```diff
@@ -1,97 +1,598 @@
...
+  /// Must be non-null unless [buildLeading] is overridden.
+  IconData? get icon;
+
+  static const _iconSize = 24.0;
+
+  Widget buildLeading(BuildContext context) {
+    assert(icon != null);
+    final designVariables = DesignVariables.of(context);
+    return Icon(icon, size: _iconSize, color: selected ? designVariables.iconSelected
+                                                : designVariables.icon);
```
gnprice, line 328: nit: misaligned

(probably just move `color` to second line)

lib/widgets/home.dart
```diff
@@ -1,97 +1,598 @@
...
+  const _MyProfileButton();
+
+  @override
+  IconData? get icon => null;
+
+  @override
+  Widget buildLeading(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    return Avatar(
+      userId: store.selfUserId, size: _MenuButton._iconSize, borderRadius: 4);
```
gnprice, line 517: Ah, bonus factoring-out! :slightly_smiling_face:

gnprice (commented):
Thanks for the swift revision! Just two nits now.

test/widgets/actions_test.dart
PIG208: Ohh! I was wondering why there was only a colon added (which made some sense to me anyway).

gnprice:
@PIG208 would you rebase? Looks like the last commit has a non-local conflict with #1041 (should have merged this one before that one, I guess, oops).

gnprice:
Otherwise all LGTM!

PIG208:
Rebased and fixed the conflict with `$ git rebase -x 'tools/check l10n --fix'`. Additionally addressed the issue that `InboxPage` no longer exists for a test from #1041.

gnprice:
Thanks!

CI has reached the `android` suite (which is the last one because it's slow), and it looks like the others pass. That one shouldn't be affected by these changes, so merging.

Issue #980, opened 2024-10-03
Case-insensitive topics in unreads data


gnprice:
Topic names in Zulip are case-insensitive.

Currently, however, our `Unreads` data structure keeps the data for each channel/stream as just a `Map` keyed by topic name. It should instead treat messages where the topic names differ only in case (and that have the same stream ID) as belonging to the same topic.

There's actually a server bug where this doesn't properly happen in the unreads data we get in the initial snapshot:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Case.20sensitivity.20in.20topics/near/1954334
But we should also handle this properly in events, and in looking up data. For example handleMessageEvent should put the new message into an existing topic if it differs only in case.

If fixing this issue for events has a side effect of working around the server-side issue, then that's nice. But if it doesn't, we don't need to add a workaround. The issue seems fairly uncommon, so it's OK if the symptoms continue when using a server that still has the bug.

#### Related issues
* #739 
* #1205

Adityakk9031:
i am going to work on this issue

chrisbobbe:
Please take a look at the Zulip project's guide to getting involved with the code:

https://zulip.readthedocs.io/en/latest/contributing/contributing.html#your-first-codebase-contribution
https://zulip.readthedocs.io/en/latest/contributing/contributing.html#picking-an-issue-to-work-on

Adityakk9031:
ok

AshutoshKhadse23:
Hello @gnprice ,
I know @chrisbobbe is working on it but could you once take a look at my implement, if yes should I create a PR for that?
And in my implementation the changes are only in ``` unread.dart and unread_check.dart```
<p align="center">
  <img width="365" alt="Image" src="https://github.com/user-attachments/assets/036f855a-e4f6-4042-9061-8b367e6a792a" />
  <img width="364" alt="Image" src="https://github.com/user-attachments/assets/0d4c1899-9a26-41e4-8c85-88fa805f3277" />
</p>

gnprice:
Well, if it only touches those two files then it definitely isn't ready for review 🙂 — as you've heard before, a PR needs tests.

AshutoshKhadse23:
Ok got it !!
```
class CaseInsensitiveTopicMap<V> implements Map<TopicName, V> {
  final Map<String, (TopicName, V)> _internal = {};

  @override
  V? operator [](Object? key) {
    if (key is! TopicName) return null;
[...]

PR #1081, opened 2024-11-24
#980 update the message.dart

Adityakk9031:


gnprice:
Closing for the same reason as #1075. Please keep scratch experiments in your own space; send a PR only when there's something you're asking us to read.

Issue #1082, opened 2024-11-25
App fails to start on chat.zulip.org with new `can_remove_subscribers_group` format


gnprice:
Since a recent deployment on chat.zulip.org, the app fails to load server data for many users there. Chat thread:
https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Infinite.20loading.20flutter.20app/near/1987288

The cause is that feature level 320 made `can_remove_subscribers_group` a group-settings value instead of always an integer group ID, and we expect an integer. So the server's response to `registerQueue` is malformed.

The immediate fix within the app is that we'll stop deserializing that field at all — we don't actually use it. Further discussion:
https://chat.zulip.org/#narrow/channel/378-api-design/topic/migrating.20settings.20to.20user-groups.20framework/near/1987303

Issue #814, opened 2024-07-16
Implement group-based permission checks


gnprice:
In the Zulip API, we've recently introduced a new group-based permission system, where a realm's admins can define arbitrary user groups to use for permissions. This is more flexible than the old roles-based system, which has a fixed handful of roles like "admin", "moderator", or "member".

Mostly this is a matter for the server, and for admin settings UIs, and we generally implement very little of an admin settings UI in the Zulip mobile apps (leaving users to go to the web/desktop app to handle those tasks). But in some places a client also needs to check if the user has a given permission, in order to give them a good experience in the case where they don't. For example, #791.

So we need to implement permission checks, and we should do so with the new API. In particular, new permission settings in Zulip (like #791) are being expressed only in the new system, so it's the only way to properly handle those.

This issue is for implementing the infrastructure of that system, so that we can easily add new permissions checks that use it when we need them. That means:
* First, as a prerequisite, implement #662.

* Then implement a method for "user A is transitively in group X". The desired semantics are transitive: if user A is in group Z, which is in group Y, which is in group X, then A is transitively in X and Y as well as Z.

  API reference for this: https://zulip.com/api/group-setting-values
  See also the web implementation, at `is_user_in_group` in `user_groups.ts`.

* Implement some particular setting that uses the system, so we can demonstrate end-to-end that the implementation works.

  I think the one current candidate for this is `direct_message_permission_group` and `direct_message_initiator_group` (for #791).
  The server also has `can_mention_group` ([this feature](https://zulip.com/api/update-user-group#parameter-can_mention_group)), which we'd use as part of #233.
  I'd thought I'd seen mention of `stream_post_policy` (for #674) getting turned into a group-based setting; but I don't see that [in the server API](https://zulip.com/api/update-stream#parameter-stream_post_policy).
  By the time you read this, there may be more: see https://zulip.com/api/changelog after feature level 270 (the latest as I write this).

chrisbobbe:
> - Implement some particular setting that uses the system, so we can demonstrate end-to-end that the implementation works.

The edit-topic permission is another candidate for this; we'd use that for #1159.

PR #1084, opened 2024-11-25
api: Cut ZulipStream.canRemoveSubscribersGroup as it changes format

gnprice:
This field is becoming a group-setting value.  We haven't yet
implemented those (that's #814), so we can no longer handle this field
until we do that work.  Conveniently we weren't actually making any
use of this field, though -- so for now, just ignore it.

Fixes: #1082


chrisbobbe:
LGTM, thanks! Merging.

PR #1086, opened 2024-11-25
content [nfc]: Pull out parseUserMention, give regexp less to do

gnprice:
This gives us more breathing room for making this handling more
complicated, as we'll need for the new "channel-wildcard-mention"
class (#1064) and for distinguishing different types of mentions (#646, #647).

Prompted by:
https://github.com/zulip/zulip-flutter/pull/1073#discussion_r1857421404
(/cc @sm-sayedi, @PIG208).

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -878,12 +878,24 @@ class _ZulipContentParser {
     final classes = element.className.split(' ')..sort();
     assert(classes.contains('user-mention')
         || classes.contains('user-group-mention'));
-    switch (classes) {
-      case ['user-mention' || 'user-group-mention']:
-      case ['silent', 'user-mention' || 'user-group-mention']:
-        break;
-      default:
-        return null;
+    int i = 0;
+
+    if (i >= classes.length) return null;
+    if (classes[i] == 'silent') {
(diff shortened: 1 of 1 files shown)


lib/model/content.dart
```diff
@@ -878,12 +878,24 @@ class _ZulipContentParser {
...
+
+    if (i >= classes.length) return null;
+    if (classes[i] == 'user-mention' || classes[i] == 'user-group-mention') {
+      // The class we already knew we'd find before we called this function.
+      // We ignore the distinction between these; see [UserMentionNode].
+      i++;
+    }
+
+    if (i != classes.length) {
+      // There was some class we didn't expect.
```
PIG208, line 897: nit: "class" -> "classes"

PIG208 (commented):
Left one comment. Overall this looks good to me! Feel free to merge after fixing it.

lib/model/content.dart
gnprice: The singular is intentional: there's at least one class we didn't expect, namely the one at position `i`.

(Grammatically it'd be either "was some class" or "were some classes".)

PR #1087, opened 2024-11-26
test: Handle emails more realistically in example data

gnprice:
Prompted by chat thread:
https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/ComposeAutocomplete.20test.20emailAddressVisbility.20help/near/1988541
(/cc @fombalang)

## Commit messages

api test: Fix a test to use Account.copyWith when it means that

The "auth headers sent by default" test below assumes that the
request is made with the email and API key of `eg.selfAccount`.
It happens to be true that `eg.account(user: eg.selfUser, …)`
will get that email, but that's a bit of an accident about
how eg.account works and what data is on eg.selfUser.

Instead of `eg.account`, use `eg.selfAccount.copyWith`, to make
the intention more explicit.

---

test: In eg.account use deliveryEmail or generate one, not user.email

This way we avoid using a source that in real data would be guaranteed
to be a fake email address: namely `user.email` for a user where
`user.deliveryEmail` is null.

---

test: Cut specific emails in eg.selfUser, eg.otherUser, and friends

Most of our tests don't and shouldn't care what the specific email
address of the user objects is.

I think these were originally here because `eg.user` didn't generate
fresh emails, so the caller needed to invent them in order to keep
them distinct.  But since d1a253856 it does, and the caller needn't.

---

test: Make eg.user more realistic about email and delivery email

This better reflects the fact that a User object will often have
null `deliveryEmail`, and that when that's the case it will have
a server-generated fake email address in `email`.

### DIFF
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -24,8 +24,7 @@ void main() {
     Future<http.BaseRequest> makeRequest(String realmUrl, String requestUrl, {
       bool? useAuth,
     }) {
-      final account = eg.account(user: eg.selfUser, apiKey: eg.selfAccount.apiKey,
-        realmUrl: Uri.parse(realmUrl));
+      final account = eg.selfAccount.copyWith(realmUrl: Uri.parse(realmUrl));
       return FakeApiConnection.with_(account: account, (connection) async {
         connection.prepare(json: {});
         final request = http.Request('GET', Uri.parse(requestUrl));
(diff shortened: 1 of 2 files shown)


chrisbobbe:
Thanks! Sorry for the delayed review. LGTM, please merge at will.

gnprice:

Issue #1089, opened 2024-11-27
compose: Error banner for deactivated DM recipient has lost its top margin


chrisbobbe:
In 3adb39557c51604d85741664026c85225fba52fa (fyi @PIG208), we regressed on the layout of the error banner shown when a DM recipient is deactivated:

| Before | After |
| --- | --- |
| <img width="675" alt="image" src="https://github.com/user-attachments/assets/32de2f0e-109c-40bb-872e-6dfa273fb19f"> | <img width="675" alt="image" src="https://github.com/user-attachments/assets/3dc3047e-8e0d-4faa-b5aa-b31dbe272d2e"> |

The top margin disappeared, and it looks buggy.

I plan to fix this in a PR that's mostly focused on proposing some prep refactors to make a cleaner starting point for #720. (If those refactors seem good, Zixuan can rebase #924 past them.)

PIG208:
Thanks to sending this issue! eb9fccd7515b50ec30148eb688b3d6c2b6bb3b31 (from #924) intersects with this by using the same `_ErrorBanner` class for the error banner here and the one that appears above the compose box.

I noticed that I didn't post a screenshot for this. Added it to #924. Also starting a discussion on the design of this.

PR #1090, opened 2024-11-27
compose: Refactor some logic; implement redesigned error banner

chrisbobbe:
Prompted by reviewing Zixuan's #924, this is meant to make the compose-box code better prepared for that work than it is in `main`; details at https://github.com/zulip/zulip-flutter/pull/924#pullrequestreview-2450028109 and in the commits.

Please let me know as soon as anything doesn't look clear. I'm tired (it's almost bedtime 🙂) but I've spent a while on this and I wanted to get it in today if possible. Since this makes a UI change, screenshots coming soon.

Fixes: #1089

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1032,18 +1032,21 @@ class _ComposeBoxContainer extends StatelessWidget {
 
 class _ComposeBoxBody extends StatelessWidget {
   const _ComposeBoxBody({
+    required this.narrow,
+    required this.controller,
     required this.topicInput,
     required this.contentInput,
     required this.sendButton,
-    required this.contentController,
-    required this.contentFocusNode,
   });
+
+  /// The narrow on view in the message list.
+  final Narrow narrow;
+
+  final ComposeBoxController controller;
 
   final Widget? topicInput;
   final Widget contentInput;
   final Widget sendButton;
-  final ComposeContentController contentController;
-  final FocusNode contentFocusNode;
 
   @override
   Widget build(BuildContext context) {
@@ -1069,10 +1072,11 @@ class _ComposeBoxBody extends StatelessWidget {
         shape: const RoundedRectangleBorder(
           borderRadius: BorderRadius.all(Radius.circular(4)))));
 
+    final ComposeBoxController(:content, :contentFocusNode) = controller;
     final composeButtons = [
-      _AttachFileButton(contentController: contentController, contentFocusNode: contentFocusNode),
-      _AttachMediaButton(contentController: contentController, contentFocusNode: contentFocusNode),
-      _AttachFromCameraButton(contentController: contentController, contentFocusNode: contentFocusNode),
+      _AttachFileButton(contentController: content, contentFocusNode: contentFocusNode),
+      _AttachMediaButton(contentController: content, contentFocusNode: contentFocusNode),
+      _AttachFromCameraButton(contentController: content, contentFocusNode: contentFocusNode),
     ];
 
     return _ComposeBoxContainer(
@@ -1149,8 +1153,8 @@ class _StreamComposeBoxBody extends StatelessWidget {
     ) = controller;
 
     return _ComposeBoxBody(
-      contentController: content,
-      contentFocusNode: contentFocusNode,
+      narrow: narrow,
+      controller: controller,
       topicInput: _TopicInput(
         streamId: narrow.streamId,
         controller: topic,
@@ -1207,8 +1211,8 @@ class _FixedDestinationComposeBoxBody extends StatelessWidget {
     ) = controller;
 
     return _ComposeBoxBody(
-      contentController: content,
-      contentFocusNode: contentFocusNode,
+      narrow: narrow,
+      controller: controller,
       topicInput: null,
       contentInput: _FixedDestinationContentInput(
         narrow: narrow,
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1134,25 +1134,13 @@ class FixedDestinationComposeBoxController extends ComposeBoxController {}
 ///
 /// This offers a text input for the topic to send to,
 /// in addition to a text input for the message content.
-class _StreamComposeBox extends StatefulWidget {
-  const _StreamComposeBox({super.key, required this.narrow});
+class _StreamComposeBox extends StatelessWidget {
+  const _StreamComposeBox({required this.narrow, required this.controller});
 
   /// The narrow on view in the message list.
   final ChannelNarrow narrow;
 
-  @override
-  State<_StreamComposeBox> createState() => _StreamComposeBoxState();
-}
-
-class _StreamComposeBoxState extends State<_StreamComposeBox> {
-  StreamComposeBoxController get controller => _controller;
-  final _controller = StreamComposeBoxController();
-
-  @override
-  void dispose() {
-    _controller.dispose();
-    super.dispose();
-  }
+  final StreamComposeBoxController controller;
 
   @override
   Widget build(BuildContext context) {
@@ -1164,13 +1152,13 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> {
       contentController: content,
       contentFocusNode: contentFocusNode,
       topicInput: _TopicInput(
-        streamId: widget.narrow.streamId,
+        streamId: narrow.streamId,
         controller: topic,
         focusNode: topicFocusNode,
         contentFocusNode: contentFocusNode,
       ),
       contentInput: _StreamContentInput(
-        narrow: widget.narrow,
+        narrow: narrow,
         topicController: topic,
         controller: content,
         focusNode: contentFocusNode,
@@ -1179,7 +1167,7 @@ class _StreamComposeBoxState extends State<_StreamComposeBox> {
         topicController: topic,
         contentController: content,
         getDestination: () => StreamDestination(
-          widget.narrow.streamId, topic.textNormalized),
+          narrow.streamId, topic.textNormalized),
       ));
   }
 }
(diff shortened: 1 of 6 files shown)


chrisbobbe:
| `main` | #924 (revision at 01c91ce50) | This PR |
| --- | --- | --- |
| image | image | image |
| image | image | image |
| image | image | image |

lib/widgets/compose_box.dart
```diff
@@ -1069,103 +1103,127 @@ class _ComposeBoxLayout extends StatelessWidget {
...
+    controller: controller.topic,
+    focusNode: controller.topicFocusNode,
+    contentFocusNode: controller.contentFocusNode,
+  );
+
+  @override Widget contentInput() => _StreamContentInput(
+    narrow: narrow,
+    topicController: controller.topic,
+    controller: controller.content,
+    focusNode: controller.contentFocusNode,
```
gnprice, line 1161: A further step that I think could be helpful is if ComposeBox grew an InheritedWidget and a static `of` method. Then instead of passing these controller pieces down through each layer, the descendant widgets that need them could say `ComposeBox.of(context).controller` and get what they need.

Or maybe the `of` method would even just return the controller ­— that's the meaningful part of the state anyway.

gnprice: Alternatively, the widgets could continue taking this information as parameters — but as a single parameter that is the whole controller. That seems like it'd also clean this up a lot.

Either way, one thing I'm thinking of in particular is how #924 adds `enabled` and `sendMessageError`, and has to pass them down through layers in a bunch of places (especially `enabled`). So ideally I'd like to see a refactor that makes it so a change like #924 can add a field to the controller and then not have to add references to it at a bunch of widget constructors but still be able to use it at various widgets deeper inside the compose box, like the various buttons.

gnprice (commented):
Neat, thanks! I agree this direction of abstracting and deduplicating this logic looks useful.

Quick comments below since I'll be out for the next couple of workdays (plus the holiday).

lib/widgets/theme.dart
```diff
@@ -375,10 +375,13 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     }
     return DesignVariables._(
       background: Color.lerp(background, other.background, t)!,
+      bannerBgIntDanger: Color.lerp(bannerBgIntDanger, other.bannerBgIntDanger, t)!,
       bgContextMenu: Color.lerp(bgContextMenu, other.bgContextMenu, t)!,
       bgCounterUnread: Color.lerp(bgCounterUnread, other.bgCounterUnread, t)!,
       bgTopBar: Color.lerp(bgTopBar, other.bgTopBar, t)!,
       borderBar: Color.lerp(borderBar, other.borderBar, t)!,
+      btnLabelAttLowIntDanger: Color.lerp(btnLabelAttLowIntDanger, other.btnLabelAttLowIntDanger, t)!,
+      btnLabelAttMediumDanger: Color.lerp(btnLabelAttMediumDanger, other.btnLabelAttMediumDanger, t)!,
```
chrisbobbe, line 384: note to self: same comment as https://github.com/zulip/zulip-flutter/pull/924#discussion_r1859498710 ; this should be `btnLabelAttMediumIntDanger`; it's `btn-label-att_medium-int_danger` in the Figma

lib/widgets/compose_box.dart
```diff
@@ -1291,12 +1342,18 @@ class _ComposeBoxState extends State<ComposeBox> implements ComposeBoxState {
...
   Widget build(BuildContext context) {
-    final Widget body;
+    final Widget? body;
+    Widget? errorBanner;
 
-    final errorBanner = _errorBanner(context);
+    errorBanner = _errorBanner(context);
     if (errorBanner != null) {
-      body = errorBanner;
+      body = null;
```
PIG208, line 1351: I feel that having to do this makes only calling `_ComposeBoxContainer` once less appealing.

Using an early return here when the `errorBanner` is present can make it easier to reason about the value of `body` and `errorBanner`. And we can comfortably pass `errorBanner: null` at the end, or `body: null` at the beginning, for clarity.

How about:
```diff
diff --git a/lib/widgets/compose_box.dart b/lib/widgets/compose_box.dart
index 1ca2e4ae..6514a331 100644
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1343,12 +1343,12 @@ class _ComposeBoxState extends State<ComposeBox> implements ComposeBoxState {
   @override
   Widget build(BuildContext context) {
     final Widget? body;
-    Widget? errorBanner;
 
-    errorBanner = _errorBanner(context);
+    final errorBanner = _errorBanner(context);
     if (errorBanner != null) {
-      body = null;
-    } else {
+      return _ComposeBoxContainer(body: null, errorBanner: errorBanner);
+    }
+
     // TODO(#720) dismissable message-send error, maybe something like:
     //     if (controller.sendMessageError.value != null) {
     //       errorBanner = _ErrorBanner(label:
@@ -1374,8 +1374,7 @@ class _ComposeBoxState extends State<ComposeBox> implements ComposeBoxState {
         assert(false);
         body = null;
     }
-    }
 
-    return _ComposeBoxContainer(body: body, errorBanner: errorBanner);
+    return _ComposeBoxContainer(body: body, errorBanner: null);
   }
 }
```

PIG208 (commented):
Thanks for working on this. Looks good overall! Left some comments.

lib/widgets/compose_box.dart
PIG208: With any of these approaches, we still need to listen to the possible updates. Our current setup for listener has a fine granularity — `_ContentInputState` listens to `ComposeContentController` and a `FocusNode`, `_StreamContentInputState` just listens to `ComposeTopicController`…

Continuing this trend, we may have individual listeners for `enabled` and `sendMessageError` as well, while turning `_TopicInput` and probably some others into `StatefulWidget`s.

With `InheritedWidget`, maybe we can use `BuildContext.dependOnInheritedWidgetOfExactType` to simplify that part.

gnprice: I think the existing ways we have things listen don't need to change.

The one thing worth avoiding with fine vs. coarse granularity here is that it'd be good to minimize how much needs to rebuild every time the user types another character. So things that don't need to depend on the text controllers should avoid doing so. But if things depend on `enabled` or `sendMessageError` when they don't strictly need to, that isn't a problem - those change much less frequently.

chrisbobbe: > InheritedWidget

I almost made a commit like this, but then noticed the branch was getting pretty long. :) I also noticed the `InheritedWidget` API prompting me to think about correctly handling changes to the value (the controller), such as not calling `dependOnInheritedWidgetOfExactType` in an `initState` method. It seemed like something to consider carefully, but I didn't have a clear idea of when the controller would change (or if/when we'd want it to), so I didn't go ahead with it.

test/widgets/compose_box_test.dart
```diff
@@ -198,41 +197,43 @@ void main() {
...
         expectTopicTextField: true);
     });
 
     testWidgets('_FixedDestinationComposeBox', (tester) async {
       final channel = eg.stream();
-      final key = await prepareComposeBox(tester,
-        narrow: TopicNarrow(channel.streamId, 'topic'), streams: [channel]);
-      checkComposeBoxTextFields(tester, controllerKey: key,
+      final controller = (await prepareComposeBox(tester,
+        narrow: TopicNarrow(channel.streamId, 'topic'), streams: [channel]))!;
```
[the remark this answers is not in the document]
chrisbobbe: Ooh nice idea, thanks.

chrisbobbe:

lib/widgets/compose_box.dart
```diff
@@ -891,43 +886,62 @@ class _SendButtonState extends State<_SendButton> {
...
   bool get _hasValidationErrors {
-    return (widget.topicController?.hasValidationErrors.value ?? false)
-      || widget.contentController.hasValidationErrors.value;
+    bool result = false;
+    final controller = widget.controller;
+    if (controller is StreamComposeBoxController) {
+      result = controller.topic.hasValidationErrors.value;
+    }
+    result |= controller.content.hasValidationErrors.value;
+    return result;
```
PIG208, line 930: I feel that this is an indicator to further abstract away logic around validation errors, probably as a future follow-up.

PIG208 (commented):
Thanks for the revision! This looks good to me. The final commit sequence seems to imply that there is the 6/6 commit passing `controller` down. I went through the file and found no other places left to refactor, so we should be all good on this, except for updating the commit summaries.

chrisbobbe:
Oh, oops! Looks like I did 1, 2, 3, 3, 4, 5 😅—will fix.

lib/widgets/compose_box.dart
chrisbobbe: I'm not yet picturing what that might look like, so I won't add a TODO in this revision, but I'm happy to if you help me understand better (or you can push that TODO yourself, wiki-style, after this is merged).

chrisbobbe:
(Done.)

gnprice (commented):
Thanks! This all looks good — just nits below. Also pushing a small added commit for one other nit:
ff85cef6e compose [nfc]: Simplify controller getters to final fields

lib/widgets/compose_box.dart
```diff
@@ -1070,102 +1096,118 @@ class _ComposeBoxLayout extends StatelessWidget {
...
   FocusNode get topicFocusNode => _topicFocusNode;
   final _topicFocusNode = FocusNode();
 
   @override
   void dispose() {
-    _topicController.dispose();
-    _contentController.dispose();
-    _contentFocusNode.dispose();
+    _topic.dispose();
+    _topicFocusNode.dispose();
```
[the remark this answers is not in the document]
chrisbobbe: Ooh thanks for the catch; I'll put the bugfix in a separate commit just before this one.

lib/widgets/compose_box.dart
```diff
@@ -1010,40 +1002,74 @@ class _SendButtonState extends State<_SendButton> {
...
 
-  final Widget child;
+  /// An error bar that goes at the top.
+  ///
+  /// This may be present on its own or with a [body].
+  /// If [body] is null this must be present.
+  ///
+  /// This widget should use a [SafeArea] to pad the left/right device insets.
+  /// If [body] is null it will also receive a bottom inset to pad;
+  /// it should pad that too.
```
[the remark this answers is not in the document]
chrisbobbe: Yep, looks good!

chrisbobbe:

gnprice:

Issue #1068, opened 2024-11-21
Rank emoji autocomplete results


gnprice:
For the initial implementation of #670, there won't be a useful ordering to the results. This means a user might have to type a longer, more specific query to find their desired emoji than they might with good ranking.

This issue is for ranking the results. The ranking should match what the web app does:
https://github.com/zulip/zulip/blob/65f05794e/web/shared/src/typeahead.ts#L322-L382
(We might make exceptions and depart from that ordering if there are areas where we see an alternative that's substantially better in either quality of results or clarity of code, without being particularly worse on the other axis.)

Broadly, that ranking looks like:
* "Popular emoji", before realm emoji, before (other) Unicode emoji, where "popular emoji" means a hard-coded list of six emoji like :+1:.
* Names that exactly match the query, before names that have it as a prefix, before other matches.

For details, see the web implementation.

### Related issues
* Prerequisite:
  #670 
* Similar features for other types of autocomplete:
  #228
  #618 
  #881

gnprice:
I decided this is a nearer-term priority after trying out the emoji picker from #1103 — the lack of ranking was bugging me basically every time I used the emoji picker. So I've now sent #1112 to implement it.

PR #1103, opened 2024-12-04
Support adding arbitrary reactions

rajveermalviya:
| Action sheet | Action sheet (with reactions) | Emoji picker |
| - | - | - |
| Screenshot 2024-12-07 at 02 12 44 | Screenshot 2024-12-07 at 02 13 02 | Screenshot 2024-12-07 at 02 13 11 |
| Screenshot 2024-12-07 at 02 12 41 | Screenshot 2024-12-07 at 02 12 58 | Screenshot 2024-12-07 at 02 13 09 |

Fixes: #388


lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +364,234 @@ class _TextEmoji extends StatelessWidget {
...
+
+    availableEmojiCandidates = allEmojiCandidates
+      .where((candidate) =>
+        !isSelfVoted(candidate) && !isPopularUnicodeEmoji(candidate.emojiDisplay))
+      .toList(growable: false);
+    searchResults = null;
+  }
+
+  // Adapted from web/src/emoji_picker.ts
+  void _filterEmojis(String query) {
```
gnprice, line 435: I'd like to have this use the emoji autocomplete machinery instead, so it works the same as in the compose box.

That has some useful features this version doesn't; and I think it's also useful just for them to feel consistent with each other for the user.

gnprice: For excluding the popular and already-reacted emoji, I think a good approach would be to extend EmojiAutocompleteQuery so it takes a parameter with a set of emoji to exclude.

But I'd also be fine with not handling that exclusion in this PR, and making it an M5 or M6 follow-up task. I think it's pretty OK if those emoji appear in the list.

lib/model/emoji.dart
```diff
@@ -30,7 +30,14 @@ class UnicodeEmojiDisplay extends EmojiDisplay {
   /// The actual Unicode text representing this emoji; for example, "🙂".
   final String emojiUnicode;
 
-  UnicodeEmojiDisplay({required super.emojiName, required this.emojiUnicode});
+  /// The Zulip "emoji code" for this emoji; for example, "1f642" for "🙂" emoji.
+  final String emojiCode;
```
gnprice, line 34: Ideally this doesn't belong here — this class is meant to be strictly the information used for displaying the emoji.

What are the places that would need to change without this field? I suspect they can pass around an EmojiCandidate instead.

lib/model/emoji.dart
```diff
@@ -54,9 +61,36 @@ class ImageEmojiDisplay extends EmojiDisplay {
 /// We do this based on a user preference,
 /// and as a fallback when the Unicode or image approaches fail.
 class TextEmojiDisplay extends EmojiDisplay {
-  TextEmojiDisplay({required super.emojiName});
+  const TextEmojiDisplay({required super.emojiName});
 }
 
+const popularUnicodeEmojis = [
+  UnicodeEmojiDisplay(
```
gnprice, line 68: Yeah, it looks like basically this one spot needs to switch to EmojiCandidate, and then that'll supply an `emojiCode` at the one place that needs it, in ReactionButtons.

EmojiCandidate is the right class here conceptually too. In principle, and in the future, the emoji we show in the quick list at the top of the action sheet should depend on actual usage and may include realm emoji (or perhaps even sometimes the Zulip extra emoji, `:zulip:`).

lib/widgets/emoji.dart
```diff
@@ -71,7 +71,9 @@ class UnicodeEmojiWidget extends StatelessWidget {
           SizedBox(height: boxSize, width: boxSize),
           PositionedDirectional(start: 0, child: Text(
             textScaler: textScaler,
-            style: TextStyle(fontSize: size),
+            style: TextStyle(
+              fontFamily: 'Apple Color Emoji',
```
gnprice, line 75: emoji: Use "Apple Color Emoji" font on iOS/macOS for UnicodeEmojiWidget
    
    Some unicode characters, like U+2764 (❤) or U+00AE (®) can
    have glyphs in non-Emoji fonts, resulting in incorrect
    rendering of such characters, where we specifically want an
    emoji to be displayed.
    
    So, explicitly mention "Apple Color Emoji" to be the font used on
    iOS/macOS for displaying the unicode emoji.

Interesting — I reproduce the issue.

I think the right fix is actually somewhat broader. I'll file an issue and send a PR.

gnprice: Well, filed #1104. Then the broad fix I had in mind turned out not to work: https://github.com/zulip/zulip-flutter/issues/1104#issuecomment-2518782312

I have a draft branch that will fix that issue for emoji in message content, which is the symptom I described in the issue. But we'll still need exactly this fix for this UnicodeEmojiWidget class, found in reaction chips and autocomplete results. So go ahead and leave this commit in your PR; I might also cherry-pick it into the PR I send for #1104.

gnprice:
I added the "Fixes" link to the issue this is a draft PR for fixing. Same reasoning as here:
https://github.com/zulip/zulip-flutter/pull/1076#issuecomment-2499064435

alya:
Let's use "more" rather than "other" for the additional reactions.

gnprice:
OK, and a couple of comments from playing with the UI just now:
* Can we get the search box to be focused (and the keyboard to appear) as soon as the user opens the search modal? The list of options is always going to be huge, and at least at present it's going to be rare that the desired emoji is in the first screenful or two — the user's almost always going to want to type a query to search.
* The results always show only a single name per emoji. (With the exception of before the user types anything — but as mentioned, that's going to be a small part of people's experience.) The design calls instead for showing a list of names.

gnprice:
Excited to see this! Looking forward to getting it completed and merged.

Some high-level comments above from an initial review yesterday. (After leaving individual comments piecemeal I forgot to leave an overall comment like this one.)

Our autocomplete machinery is definitely somewhat complicated, but I think it should work well for this task. If you have any questions as you go to use it, please don't hesitate to ask in #mobile-team.

lib/widgets/emoji.dart
gnprice: OK, sent #1108. It includes a cherry-pick of this commit — the only change I made to the diff was to add a test.

rajveermalviya:
Thanks for the initial comments @gnprice @alya!
This is now ready for review, I've also updated the screenshots.

lib/widgets/action_sheet.dart
```diff
@@ -247,30 +248,56 @@ class ReactionButtons extends StatelessWidget {
...
+            }))))),
+        TextButton(
+          onPressed: () {
+            showEmojiPickerSheet(pageContext: pageContext, message: message);
+          },
+          style: TextButton.styleFrom(
+            padding: const EdgeInsets.fromLTRB(12, 12, 4, 12),
+            splashFactory: NoSplash.splashFactory,
+            foregroundColor: designVariables.contextMenuItemText,
+            shape: const RoundedRectangleBorder(borderRadius: BorderRadius.zero)
```
chrisbobbe, line 286: Why `RoundedRectangleBorder` where the rectangle is not rounded (border radius is zero)? 🙂 (Here and in another place in the PR.)

lib/widgets/action_sheet.dart
```diff
@@ -247,30 +248,56 @@ class ReactionButtons extends StatelessWidget {
...
+                ).copyWith(backgroundColor: WidgetStateColor.resolveWith((states) =>
+                  states.any((e) => e == WidgetState.pressed || e == WidgetState.selected)
+                    ? designVariables.contextMenuItemBg.withFadedAlpha(0.20)
+                    : Colors.transparent)),
+                icon: UnicodeEmojiWidget(
+                  emojiDisplay: emoji.emojiDisplay as UnicodeEmojiDisplay,
+                  notoColorEmojiTextSize: 20.1,
+                  size: 24));
+            }))))),
+        TextButton(
```
chrisbobbe, line 278: I think `TextButton` is adding 2px padding at the top and 2px at the bottom, to make the button's height 48px instead of 44px as in the Figma. It looks like we don't use `TextButton` in the emoji autocomplete. What if we followed that example?

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,225 @@ class _TextEmoji extends StatelessWidget {
...
+      Expanded(child: InsetShadowBox(
+        top: 8, bottom: 8,
+        color: designVariables.bgContextMenu,
+        child: ListView.builder(
+          padding: const EdgeInsets.all(8),
+          itemCount: _resultsToDisplay.length,
+          itemBuilder: (context, i) => EmojiPickerListEntry(
+            pageContext: widget.pageContext,
+            emoji: _resultsToDisplay[i].candidate,
+            message: widget.message)))),
```
chrisbobbe, line 501: The touch feedback is colored by a Material default, `overlayColor`:

image

I see we don't have a spec for these items' touch feedback in the Figma, but is there something similar there you can infer a good color from?

Also, to me it seems like the touch feedback should extend out to the edge of the bottom action sheet.

chrisbobbe: Also, if the user has already reacted with one of the emoji in the results, how about showing that state by marking the item somehow? Maybe similarly (or the same) as the on-press feedback, would that be reasonable?

chrisbobbe (commented):
Thanks! Very excited for this feature. Comments below.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,225 @@ class _TextEmoji extends StatelessWidget {
...
+      return SafeArea(
+        // For _EmojiPickerItem, and RealmContentNetworkImage used in ImageEmojiWidget.
+        child: PerAccountStoreWidget(
+          accountId: store.accountId,
+          child: Padding(
+            // By default, when software keyboard is opened, the ListView
+            // expands behind the software keyboard — resulting in some
+            // list entries being covered by the keyboard. Add explicit
+            // bottom padding the size of the keyboard, which fixes this.
+            padding: EdgeInsets.only(bottom: MediaQuery.viewInsetsOf(context).bottom),
```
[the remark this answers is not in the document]
gnprice: Confusing, yeah.

Let's put this next to the SafeArea widget — their jobs are closely related.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,225 @@ class _TextEmoji extends StatelessWidget {
...
+    super.initState();
+    _controller = TextEditingController();
+  }
+
+  @override
+  void onNewStore() {
+    final store = PerAccountStoreWidget.of(context);
+
+    if (_viewModel != null) {
+      _controller.removeListener(_handleViewModelUpdate);
```
gnprice, line 428: doesn't match `addListener`

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,225 @@ class _TextEmoji extends StatelessWidget {
...
+      _viewModel!.dispose();
+      _resultsToDisplay = const [];
+    }
+    _viewModel = EmojiAutocompleteView.init(store: store, query: EmojiAutocompleteQuery(''))
+      ..addListener(_handleViewModelUpdate);
+    _controller.addListener(_handleControllerUpdate);
```
gnprice, line 434: This can go in `initState` (like in `AutocompleteField`), so the listener gets added only once even if the store gets replaced.

The listener won't be called before this point anyway — it won't be called until there's that TextField attached to the controller, which happens only after this is mounted and built, and `onNewStore` will get called after the state is mounted and before it's built.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,225 @@ class _TextEmoji extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+      Padding(padding: const EdgeInsets.only(left: 8, top: 8),
```
gnprice, line 463: nit: indent
```suggestion
      children: [
        Padding(padding: const EdgeInsets.only(left: 8, top: 8),
```

lib/widgets/action_sheet.dart
```diff
@@ -183,27 +178,39 @@ class MessageActionSheetCancelButton extends StatelessWidget {
...
+  /// A context within the [MessageListPage] this action sheet was
+  /// triggered from.
+  final BuildContext pageContext;
 
-  @override void onPressed() async {
+  /// List of popular emoji reaction buttons to display.
+  /// Each emoji must be a unicode emoji.
+  final List<EmojiCandidate> popularEmojis;
+
+  void _onPressed(
```
gnprice, line 199: Let's abstract this out into a function that both this and the emoji picker can call. Probably `emoji_reaction.dart` is a good home.

lib/widgets/action_sheet.dart
```diff
@@ -216,9 +222,57 @@ class AddThumbsUpButton extends MessageActionSheetMenuItemButton {
...
+    bool hasSelfVote(EmojiCandidate emoji) {
+      return message.reactions?.aggregated.any((reactionWithVotes) {
+        return reactionWithVotes.reactionType == ReactionType.unicodeEmoji
+          && reactionWithVotes.emojiCode == emoji.emojiCode
+          && reactionWithVotes.userIds.contains(store.selfUserId);
+      }) ?? false;
+    }
+
+    return Container(
+      padding: const EdgeInsets.all(8),
```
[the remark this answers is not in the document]
rajveermalviya: Unfortunately, adding 10px horizontal padding on each button will results in overflow on the minimum supported screen width (320px) with default text size:
| `horizontal: 10` | `horizontal: 5` |
| - | - |
| Screenshot 2024-12-07 at 09 42 29 | Screenshot 2024-12-07 at 09 43 04 |

Minimum padding that doesn't overflow is 5px.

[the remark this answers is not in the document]
rajveermalviya: Another option is to have no horizontal padding, and make the container take as much width possible and aligning the emoji in center:
| padding `horizontal: 5` | padding `horizontal: 0`, width `double.infinity` |
| - | - |
| Screenshot 2024-12-07 at 09 53 28 | Screenshot 2024-12-07 at 09 51 58 |
| Screenshot 2024-12-07 at 09 53 22 | Screenshot 2024-12-07 at 09 51 43 |

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,237 @@ void main() {
...
+        required String emojiName,
+      }) {
+        return (Subject<Object?> it) => it.isA<EmojiPickerListEntry>()
+          ..emoji.which((it) => it
+            ..emojiType.equals(emojiType)
+            ..emojiCode.equals(emojiCode)
+            ..emojiName.equals(emojiName));
+      }
+
+      testWidgets('show, search', (tester) async {
```
gnprice, line 247: This test will need updates when merging with #1112.

lib/model/emoji.dart
```diff
@@ -88,6 +88,60 @@ final class EmojiCandidate {
...
+// See: https://github.com/zulip/zulip/blob/3bad36ef8cf07cd57d0b257a739bae635a8527ac/web/shared/src/typeahead.ts#L22-L29
+const zulipPopularEmojis = [
+  EmojiCandidate(
+    emojiType: ReactionType.unicodeEmoji,
+    emojiCode: '1f44d',
+    emojiName: '+1',
+    aliases: ['thumbs_up', 'like'],
+    emojiDisplay: UnicodeEmojiDisplay(
+      emojiName: '+1',
+      emojiUnicode: '\u{1f44d}')), // '👍'
```
gnprice, line 102: I ended up providing this same data as part of #1112. Probably most convenient is if you cherry-pick the relevant commit:
bce508489 emoji: Add list of the "popular" emoji

into this PR.

gnprice (commented):
Trying out this PR's UI on Friday, I decided ranking the results (#1068) is something we'll need soon too — the lack of ranking was bugging me basically every time I use the emoji picker. So I've just sent #1112 for that.

Happily, that combines perfectly smoothly with the emoji picker: now that this PR uses the emoji autocomplete machinery, it gets the benefit of #1112's ranking as soon as the two branches are merged :slightly_smiling_face: 

Trying `git merge pr/1112` from the tip of this branch, the merge is clean, and the emoji picker works fine and has the new ranking. Running `flutter test` shows one test that needs to be updated, which I've marked below.

@rajveermalviya It's probably cleanest if for now you don't rebase this PR on top of that one, since each PR has a fair amount of complexity going on already. (Instead just cherry-pick one commit as mentioned below.) But it'd be good to try that `git merge` and see what changes you have to make to fix that one test, at any point when you're otherwise waiting for the next review; then you can keep that version around so it's ready when the time comes.

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
gnprice: That second option LGTM.

An ideal version would have a minimum width for each item, and when there isn't room for them all at that width (which can happen if the user has set their system display-size setting to make everything big) we'd just drop items from the end of the list. That way :+1: and :tada: and perhaps :slightly_smiling_face: would appear at the appropriate size, at the cost of :octopus: and perhaps :hammer_and_wrench: being evicted from this quick-pick line and having to be found under the "more" button instead.

But implementing that will be more complicated, so let's keep it out of scope for this PR.

rajveermalviya:
Thanks for the reviews @chrisbobbe, @gnprice! Pushed a new revision, PTAL.

gnprice (commented):
Recording for the thread here some context from my DMs with @rajveermalviya: the first five commits here are rebased from #1041, after I asked Rajesh to rebase atop those refactors in order to facilitate merging the PRs with each other.

The merge worked! So I expect to include #1041 in today's [preview beta](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/beta.20release.20management/near/1997112), as well as this PR.

lib/widgets/action_sheet.dart
```diff
@@ -183,41 +147,158 @@ class MessageActionSheetCancelButton extends StatelessWidget {
...
+          && reactionWithVotes.userIds.contains(store.selfUserId);
+      }) ?? false;
     }
+
+    return Container(
+      decoration: BoxDecoration(color: designVariables.contextMenuCancelBg),
+      child: Row(children: [
+        Flexible(child: Row(
+          spacing: 1,
+          children: List.unmodifiable(EmojiStore.popularEmojiCandidates.mapIndexed((index, emoji) {
```
gnprice, line 254: nit: pull this callback out as a helper method:
```suggestion
       children: List.unmodifiable(EmojiStore.popularEmojiCandidates
            .mapIndexed(_buildButton)))),
```

That'll dedent it and I think also help simplify understanding the structure.

lib/widgets/action_sheet.dart
```diff
@@ -183,41 +147,158 @@ class MessageActionSheetCancelButton extends StatelessWidget {
...
+                color: isSelfVoted
+                  ? designVariables.contextMenuItemBg.withFadedAlpha(0.20)
+                  : null,
+                child: UnicodeEmojiWidget(
+                  emojiDisplay: emoji.emojiDisplay as UnicodeEmojiDisplay,
+                  notoColorEmojiTextSize: 20.1,
+                  size: 24))));
+          })))),
+        InkWell(
+          onTap: _onMorePressed,
```
gnprice, line 279: nit: "onFoo" for callbacks, "handleFoo" for the method that is called:
```suggestion
          onTap: _handleTapMore,
```

https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables

Basically it's useful to have some convention for naming the two sides differently, because lots of places have one of each of these kinds of things in the same scope, both being about the same thing, and so without such a convention you end up with name collisions that have to be resolved by hacks like picking a less-good name for one of the two.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,261 @@ class _TextEmoji extends StatelessWidget {
...
+    _viewModel = EmojiAutocompleteView.init(store: store, query: query)
+      ..addListener(_handleViewModelUpdate);
+  }
+
+  @override
+  void onNewStore() {
+    if (_viewModel != null) {
+      _viewModel!.dispose();
+    }
+    _initViewModel(EmojiAutocompleteQuery(_controller.text));
```
gnprice, line 477:
```suggestion
    if (_viewModel != null) {
      assert(_viewModel!.query == _controller.text);
      _viewModel!.dispose();
    }
    _initViewModel(EmojiAutocompleteQuery(_controller.text));
```

That should be true, right? And if somehow it's not true, then I think this would have a glitch as it'd change what query the view-model was working on.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,261 @@ class _TextEmoji extends StatelessWidget {
...
+  }
+
+  void _handleControllerUpdate() {
+    if (_viewModel != null) {
+      _viewModel!.query = EmojiAutocompleteQuery(_controller.text);
+    }
```
gnprice, line 483: It should be impossible for `_viewModel` to be null at this point, right? (Similar reasoning to at https://github.com/zulip/zulip-flutter/pull/1103#discussion_r1874217404 .) So:
```suggestion
  void _handleControllerUpdate() {
    _viewModel!.query = EmojiAutocompleteQuery(_controller.text);
```

lib/widgets/emoji_reaction.dart
gnprice, line 477: Looks like this is the only call site of `_initViewModel`, so we can simplify by inlining it.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,261 @@ class _TextEmoji extends StatelessWidget {
...
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return Column(
+      mainAxisSize: MainAxisSize.min,
```
gnprice, line 505: Does this do anything when there's an `Expanded` child? I'd think it doesn't.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,261 @@ class _TextEmoji extends StatelessWidget {
...
+  static const _emojiSize = 24.0;
+  static const _notoColorEmojiTextSize = 20.1;
+
+  void _onPressed() {
+    // Dismiss the enclosing action sheet immediately,
+    // for swift UI feedback that the user's selection was received.
+    Navigator.pop(pageContext);
+
+    doAddOrRemoveReaction(
+      context: pageContext,
```
gnprice, line 574: I see, this is why we end up needing to pass `pageContext` down through all these layers (rather than each widget using its own BuildContext like usual). It's so that if the add/remove request fails, `doAddOrRemoveReaction` has a (probably-)still-mounted context it can use for showing the error.

I feel like there's probably some better pattern we could use here. But I'm not sure offhand what it'd look like — will probably take a bit of experimenting — and this is the natural extension of the pattern we already use throughout the action sheets. So this pattern is good to stick with for this PR.

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,261 @@ class _TextEmoji extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    final emojiDisplay = emoji.emojiDisplay.resolve(store.userSettings);
+    final Widget? glyph = switch (emojiDisplay) {
+      ImageEmojiDisplay() =>
+        ImageEmojiWidget(size: _emojiSize, emojiDisplay: emojiDisplay),
+      UnicodeEmojiDisplay() =>
```
gnprice, line 591:
```suggestion
    // TODO deduplicate this logic with [_EmojiAutocompleteItem]
    final emojiDisplay = emoji.emojiDisplay.resolve(store.userSettings);
    final Widget? glyph = switch (emojiDisplay) {
      ImageEmojiDisplay() =>
        ImageEmojiWidget(size: _emojiSize, emojiDisplay: emojiDisplay),
      UnicodeEmojiDisplay() =>
```

And then a similar comment in the other place pointing back to here.

That way at least if we start making changes in one place or the other, we'll have something to remind us that there's another copy it'd be good to keep in sync.

And also a good followup sometime would be to find a way to deduplicate the two.

lib/widgets/emoji_reaction.dart
rajveermalviya: Yeah it doesn't, removed.

rajveermalviya:

gnprice (commented):
Thanks for the revision!

The implementation code all looks good now.

Switching to another task for the moment, but here's a couple of things I spotted when starting to check against the Figma design. I didn't complete a comparison, so there may be more such things you could find.

(I also haven't yet looked at the tests.)

lib/widgets/emoji_reaction.dart
```diff
@@ -360,3 +366,256 @@ class _TextEmoji extends StatelessWidget {
...
+                hintText: zulipLocalizations.emojiPickerSearchEmoji,
+                contentPadding: const EdgeInsetsDirectional.only(start: 10, top: 6),
+                filled: true,
+                fillColor: designVariables.bgSearchInput,
+                border: OutlineInputBorder(
+                  borderRadius: BorderRadius.circular(10),
+                  borderSide: BorderSide.none),
+                hintStyle: TextStyle(color: designVariables.textMessage)),
+              style: const TextStyle(fontSize: 19, height: 26 / 19)))),
+          TextButton(
```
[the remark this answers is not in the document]
gnprice: Looks like this does it (and puts the right padding at the end too):
```diff
             style: TextButton.styleFrom(
-              padding: EdgeInsets.zero,
+              padding: const EdgeInsets.symmetric(horizontal: 8),
```

lib/widgets/action_sheet.dart
```diff
@@ -194,41 +190,128 @@ abstract class MessageActionSheetMenuItemButton extends ActionSheetMenuItemButto
...
+    bool hasSelfVote(EmojiCandidate emoji) {
+      return message.reactions?.aggregated.any((reactionWithVotes) {
+        return reactionWithVotes.reactionType == ReactionType.unicodeEmoji
+          && reactionWithVotes.emojiCode == emoji.emojiCode
+          && reactionWithVotes.userIds.contains(store.selfUserId);
+      }) ?? false;
     }
+
+    return Container(
+      decoration: BoxDecoration(color: designVariables.contextMenuItemBg.withFadedAlpha(0.12)),
```
gnprice, line 280: nit: overlong line

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+        final emojiDisplay = emoji.emojiDisplay as UnicodeEmojiDisplay;
+
+        Future<void> tapButton(WidgetTester tester) async {
+          await tester.tap(find.descendant(
+            of: find.descendant(
+              of: find.descendant(
+                of: find.byType(ReactionButtons),
+                matching: find.byType(InkWell)),
+              matching: find.byType(UnicodeEmojiWidget)),
+            matching: find.text(emojiDisplay.emojiUnicode)));
```
gnprice, line 143: We can put this more in terms of how the user sees it, and simplify:
```suggestion
        Future<void> tapButton(WidgetTester tester) async {
          await tester.tap(find.descendant(
            of: find.byType(BottomSheet),
            matching: find.text(emojiDisplay.emojiUnicode)));
```

(I see just `find.text` alone doesn't work in the "removing" tests, because the same emoji appears in a reaction chip.)

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
-      await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
+    group('popular emoji reactions;', () {
+      testWidgets('ensure all are shown', (tester) async {
+        final message = eg.streamMessage();
+        await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
 
-      connection.prepare(json: {});
-      await tapButton(tester);
-      await tester.pump(Duration.zero);
+        // Ensure all popular emoji buttons are shown.
```
gnprice, line 118: This is something we already check with the tests below, right? And they do so in a nice end-to-end way: they check that the actual emoji is shown (with an appropriate `Text` widget), and that tapping it works. So I don't think we need an extra test checking that.

One thing I guess this test case does is check that we don't have any *other* options offered here in the same way. But I'd be happy not having a test for that.

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+        final emojis = tester.widgetList<UnicodeEmojiWidget>(find.descendant(
+          of: find.descendant(
+            of: find.byType(ReactionButtons),
+            matching: find.byType(InkWell)),
+          matching: find.byType(UnicodeEmojiWidget)));
```
gnprice, line 123: This feels pretty deep into the implementation details of this UI. Given the comment above, let's just leave out this test case.

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+            expectedTitle: 'Adding reaction failed',
+            expectedMessage: 'Invalid message(s)')));
+        });
+      }
     });
 
-    testWidgets('request has an error', (tester) async {
-      final message = eg.streamMessage();
-      await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message));
+    group('emoji picker;', () {
```
gnprice, line 207: nit: the code this is really testing is in `lib/widgets/emoji_reaction.dart`, so let's put the test in the corresponding file `test/widgets/emoji_reaction_test.dart`

(I see it's using the `setupToMessageActionSheet` helper; but it can import that from this file)

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+        store.setServerEmojiData(ServerEmojiData(codeToNames: {
+          '1f4a4': ['zzz', 'sleepy'], // (just 'zzz' in real data)
+        }));
+        await store.handleEvent(RealmEmojiUpdateEvent(id: 1, realmEmoji: {
+          '1': eg.realmEmojiItem(emojiCode: '1', emojiName: 'buzzing'),
+        }));
+
+        await tester.tap(find.ancestor(
+          of: find.byIcon(ZulipIcons.chevron_right),
+          matching: find.byType(InkWell)));
```
gnprice, line 228: nit: simplify and put more into user-visible terms:
```suggestion
        await tester.tap(find.byIcon(ZulipIcons.chevron_right));
```

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+      }) {
+        return (Subject<Object?> it) => it.isA<EmojiPickerListEntry>()
+          ..emoji.which((it) => it
+            ..emojiType.equals(emojiType)
+            ..emojiCode.equals(emojiCode)
+            ..emojiName.equals(emojiName));
+      }
+
+      List<Condition<Object?>> arePopularCandidates = popularCandidates.map((c) =>
+        conditionEmojiListEntry(
```
gnprice, line 248: nit: these aren't candidates (i.e. `EmojiCandidate`), but are entries:
```suggestion
      List<Condition<Object?>> arePopularEntries = popularCandidates.map((c) =>
        conditionEmojiListEntry(
```

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+            emojiCode: 'zulip',
+            emojiName: 'zulip'),
+          conditionEmojiListEntry(
+            emojiType: ReactionType.unicodeEmoji,
+            emojiCode: '1f4a4',
+            emojiName: 'zzz'),
+        ]);
+
+        tester.widget(searchFieldFinder);
+        await tester.enterText(searchFieldFinder, 'z');
```
gnprice, line 274:
```suggestion
        await tester.enterText(searchFieldFinder, 'z');
```

Or is there an effect the `tester.widget` line has which I'm missing?

test/widgets/action_sheet_test.dart
```diff
@@ -99,46 +107,244 @@ void main() {
...
+        final message = eg.streamMessage();
+        await setupEmojiPicker(tester, message: message, narrow: TopicNarrow.ofMessage(message));
+
+        connection.prepare(httpStatus: 400, json: {
+          'code': 'BAD_REQUEST',
+          'msg': 'Invalid message(s)',
+          'result': 'error',
+        });
+        await tester.tap(find.text('\u{1f4a4}')); // 'zzz' emoji
+        await tester.pump(Duration.zero); // error arrives; error dialog shows
```
gnprice, line 340: Let's adapt this test so that it would successfully catch if we accidentally broke in the future the careful work you did to properly pass down `pageContext`: https://github.com/zulip/zulip-flutter/pull/1103#discussion_r1877511231

(As Hixie likes to say: the purpose of a test is to make sure we don't accidentally revert your changes.)

As is, if you apply this diff, no tests fail:
```diff
           itemBuilder: (context, i) => EmojiPickerListEntry(
-            pageContext: widget.pageContext,
+            pageContext: context,
```

You can make the test catch that by doing this:
```suggestion
        await tester.pump(); // register tap
        await tester.pump(Duration(seconds: 1)); // emoji picker animates away
        await tester.pump(Duration(seconds: 1)); // error arrives; error dialog shows
```

and adding `delay: Duration(seconds: 2)` to the `prepare` call.

gnprice (commented):
OK, finished comparing with Figma and also reading through the tests. Comments below, in addition to the two above.

This is getting quite close to merge. Looking forward to having it in!

rajveermalviya:

test/widgets/emoji_reaction_test.dart
```diff
@@ -284,4 +296,184 @@ void main() {
...
+    });
+
+    testWidgets('adding success', (tester) async {
+      final message = eg.streamMessage();
+      await setupEmojiPicker(tester, message: message, narrow: TopicNarrow.ofMessage(message));
+
+      connection.prepare(json: {});
+      await tester.tap(find.descendant(
+        of: find.byType(BottomSheet),
+        matching: find.text('\u{1f4a4}')), warnIfMissed: false); // 'zzz' emoji
```
gnprice, line 434: Huh, why warnIfMissed false?
```suggestion
      await tester.tap(find.descendant(
        of: find.byType(BottomSheet),
        matching: find.text('\u{1f4a4}'))); // 'zzz' emoji
```

Seems like if this tap doesn't actually hit, that'd be a bug somewhere.

gnprice (commented):
Thanks for the revision! All looks good now except one question in the tests.

test/widgets/emoji_reaction_test.dart
rajveermalviya: Whoops; this wasn't meant to be included, it's an itermediate debug leftover.
Fixed now.

gnprice:
Thanks! LGTM; rebased and will watch CI, since there were a few conflicts to resolve (in `test/widgets/action_sheet_test.dart`, icons, and lib/generated/l10n/.)

gnprice:
CI passed, and then I merged #1076 and there's a fresh conflict. That one is trivial, though — so resolved, running `tools/check` locally to be sure, and then I'll merge.

PR #1106, opened 2024-12-05
Make topic headers case-insensitive

### LINKED ISSUE #739: Extra recipient header when topic changes case
If you visit the thread `#translation > Missing string` on chat.zulip.org, around this message (the first one in today's thread):
https://chat.zulip.org/#narrow/stream/58-translation/topic/Missing.20string/near/1828362

you'll find it gets its own recipient header, even in a topic narrow, even though there are previous messages. Specifically there's a recipient header "missing string" for the older messages (which are from 2021), and another one "Missing string" for that message and the ones after it.

In general Zulip topics are case-insensitive. Specifically, if you look at that conversation in Zulip web, there's just one recipient header; it reflects the older messages, so "missing string". So we should do the same.

shivanshsharma13:
Topics in Zulip are case-insensitive. This change makes the message list's topic headers match that behavior, so messages whose topics differ only in case (like "missing string" and "Missing string") share a single header.  This brings the behavior in line with Zulip web.

### What is the change?

The change modifies the topic comparison logic in `haveSameRecipient()` to use case-insensitive comparison when determining whether to show a new recipient header.

### Screenshots

| Before                                                                                       | After                                                                                        |
|---------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------|
|  |  |

Fixes #739

### DIFF
--- a/test/model/message_list_test.dart
+++ b/test/model/message_list_test.dart
@@ -1803,6 +1803,26 @@ void main() {
         }
       }
     });
+
+    group('topics compared case-insensitively', () {
+      void doTest(String description, String topicA, String topicB, bool expected) {
+        test(description, () {
+          final stream = eg.stream();
+          final messageA = eg.streamMessage(stream: stream, topic: topicA);
+          final messageB = eg.streamMessage(stream: stream, topic: topicB);
+          check(haveSameRecipient(messageA, messageB)).equals(expected);
+        });
+      }
+
+      doTest('same case, all lower',               'abc',  'abc',  true);
+      doTest('same case, all upper',               'ABC',  'ABC',  true);
+      doTest('same case, mixed',                   'AbC',  'AbC',  true);
+      doTest('same non-cased chars',               '嗎',    '嗎',    true);
+      doTest('different case',                     'aBc',  'ABC',  true);
+      doTest('different case, same diacritics',    'AbÇ',  'aBç',  true);
+      doTest('same letters, different diacritics', 'ma',   'mǎ',   false);
+      doTest('having different CJK characters',    '嗎', '馬', false);
+    });
   });
 
   test('messagesSameDay', () {
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -352,7 +352,7 @@ mixin _MessageSequence {
 bool haveSameRecipient(Message prevMessage, Message message) {
   if (prevMessage is StreamMessage && message is StreamMessage) {
     if (prevMessage.streamId != message.streamId) return false;
-    if (prevMessage.topic != message.topic) return false;
+    if (prevMessage.topic.toLowerCase() != message.topic.toLowerCase()) return false;
   } else if (prevMessage is DmMessage && message is DmMessage) {
     if (!_equalIdSequences(prevMessage.allRecipientIds, message.allRecipientIds)) {
       return false;


chrisbobbe:
Hi, thanks for the PR! Please revise the commit message to follow our style (we have a useful [tip](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/git.md#2-the-secret-to-using-git-log--p) that will help you see examples of mergeable commit messages in the project's history). Also, please write a test for this bugfix.

shivanshsharma13:
Implemented the requested changes. Please let me know if there’s anything else that needs adjustment. Thanks

test/model/message_list_test.dart
```diff
@@ -1703,6 +1703,23 @@ void main() {
         }
       }
     });
+    test('stream messages match with case-insensitive topics', () {
```
chrisbobbe, line 1706: nit: put this new test just under the test 'stream messages match just if same stream/topic'

chrisbobbe: Also here's a version that has different tests for each case (instead of multiple checks in the same test), along with a descriptive label for each test, which will appear in output when it fails. This also tests some additional cases that are interesting:

```dart
    group('topics compared case-insensitively', () {
      void doTest(String description, String topicA, String topicB, bool expected) {
        test(description, () {
          final stream = eg.stream();
          final messageA = eg.streamMessage(stream: stream, topic: topicA);
          final messageB = eg.streamMessage(stream: stream, topic: topicB);
          check(haveSameRecipient(messageA, messageB)).equals(expected);
        });
      }

      doTest('same case, all lower',               'abc',  'abc',  true);
      doTest('same case, all upper',               'ABC',  'ABC',  true);
      doTest('same case, mixed',                   'AbC',  'AbC',  true);
      doTest('same non-letter chars',              '嗎',    '嗎',    true);
      doTest('different case',                     'aBc',  'ABC',  true);
      doTest('different case, same diacritics',    'AbÇ',  'aBç',  true);
      doTest('same letters, different diacritics', 'ma',   'mǎ',   false);
      doTest('having different non-letter chars',  'ma 嗎', 'mǎ 馬', false);
    });
```

chrisbobbe (commented):
Thanks!

For the commit message:

```
Message_list: Make topic headers case-insensitive

This change in the commit makes the message list's topic headers match that behavior, so messages whose topics differ only in case (like "missing string" and "Missing string") share a single header.  This brings the behavior in line with Zulip web. (Added the relevant test case)
```

We use the prefix `msglist:` for commits about the message list; see examples using the tip I shared earlier for browsing history.

Also let's make the summary line more explicit about the bug it fixes. How about:

```
msglist: Remove extra recipient header when topic changes case
```

Also your commit is missing a `Fixes:` line and isn't word-wrapped properly.

test/model/message_list_test.dart
shivanshsharma13: Yes, Thanks for the help. This test looks promising and checks for each case. I'll just replaces this with mine.

chrisbobbe:
Thanks! Please wrap the commit-message body to 68 characters and format the "Fixes" line the same way we do in other commits (see the project's history for examples).

shivanshsharma13:
> Thanks! Please wrap the commit-message body to 68 characters and format the "Fixes" line the same way we do in other commits (see the project's history for examples).

Done, Please review.
- [X] Wrapped the cotent in 68 chars a line
- [X] Used `Fixes:` format as done on other commits (sorry for overlooking previously)

test/model/message_list_test.dart
```diff
@@ -1704,6 +1704,25 @@ void main() {
       }
     });
   });
+  group('topics compared case-insensitively', () {
```
gnprice, line 1707: nit: separate with blank line

gnprice, line 1707: nit: nest this inside the haveSameRecipient group

test/model/message_list_test.dart
```diff
@@ -1704,6 +1704,25 @@ void main() {
...
+        final messageA = eg.streamMessage(stream: stream, topic: topicA);
+        final messageB = eg.streamMessage(stream: stream, topic: topicB);
+        check(haveSameRecipient(messageA, messageB)).equals(expected);
+      });
+    }
+
+    doTest('same case, all lower',               'abc',  'abc',  true);
+    doTest('same case, all upper',               'ABC',  'ABC',  true);
+    doTest('same case, mixed',                   'AbC',  'AbC',  true);
+    doTest('same non-letter chars',              '嗎',    '嗎',    true);
```
gnprice, line 1720:
```suggestion
    doTest('same non-cased chars',               '嗎',    '嗎',    true);
```

Calling this character "non-letter" is confusing, because Unicode in fact classifies it in its "Letter" category, specifically "Lo (Letter, Other)":
```
$ unicode --long 嗎
U+55CE CJK UNIFIED IDEOGRAPH-55CE
UTF-8: e5 97 8e UTF-16BE: 55ce Decimal: &#21966; Octal: \052716
嗎
Category: Lo (Letter, Other); East Asian width: W (wide)
Unicode block: 4E00..9FFF; CJK Unified Ideographs
Bidi: L (Left-to-Right)
```

(By comparison, cased letters like `a` and `A` are in categories "Ll (Letter, Lowercase)" and "Lu (Letter, Uppercase)" respectively.)

test/model/message_list_test.dart
```diff
@@ -1704,6 +1704,25 @@ void main() {
...
+    doTest('different case',                     'aBc',  'ABC',  true);
+    doTest('different case, same diacritics',    'AbÇ',  'aBç',  true);
+    doTest('same letters, different diacritics', 'ma',   'mǎ',   false);
+    doTest('having different non-letter chars',  'ma 嗎', 'mǎ 馬', false);
```
gnprice, line 1724:
```suggestion
    doTest('same letters, different diacritics', 'ma',   'mǎ',   false);
    doTest('having different non-letter chars',  '嗎',    '馬',    false);
```

(plus the renaming above)

Otherwise the last test doesn't really exercise the comparison of the two sinographs — the two names are already going to be different because of the "a" vs "ǎ".

gnprice (commented):
Thanks @shivanshsharma13, and thanks @chrisbobbe for the previous reviews!

Generally this looks good; comments below.

shivanshsharma13:
Thanks for the detailed review. I acknowledged the suggested changes. Please let me know if anything needs to be changed!🙂

test/model/message_list_test.dart
```diff
@@ -1703,6 +1703,25 @@ void main() {
         }
       }
     });
+    group('topics compared case-insensitively', () {
```
gnprice, line 1706: bump https://github.com/zulip/zulip-flutter/pull/1106#discussion_r1889259739

gnprice (commented):
Thanks for the revision! All looks good except one nit below.

(Reviews are a bit slow this week and next, due to holidays.)

test/model/message_list_test.dart
shivanshsharma13: Oh, sorry for overlooking that. Done now!

shivanshsharma13:
Thanks! Totally understandable with the holiday season—Christmas and New Year can be a busy time. Wishing you happy holidays.

gnprice:
Looks good — merged. Thanks again @shivanshsharma13 for the contribution!

PR #1114, opened 2024-12-09
Exclude deactivated realm emoji from autocomplete

### LINKED ISSUE #1113: Deactivated custom emoji offered in autocomplete
A custom emoji can be deactivated:
* https://zulip.com/help/custom-emoji#deactivate-custom-emoji

When that happens, it's still valid for existing messages and reactions; but the server will reject new uses, and so it shouldn't be offered in the UI.

Currently we don't take account of that, so we offer deactivated custom emoji the same way as those still active.

The effect is conspicuous on chat.zulip.org, because there are a substantial number of deactivated emoji (57 of them):
```
$ <tmp/register.json jq '.realm_emoji[] | .deactivated' | sort | uniq -c
    131 false
     57 true
```
Particularly so after #1112 (implementing #1068), since custom emoji are ranked ahead of Unicode emoji (other than the "popular" emoji).

gnprice:
Fixes #1113.

The first commit (toString on some emoji-related data classes) is shared with #1112.

When combining with #1112, the three new test cases will each need a tweak like this one:
```diff
       check(store.allEmojiCandidates()).deepEquals([
+        ...arePopularCandidates,
```
just like #1112 itself does (in its last commit) for several existing test cases.

# Selected commit messages

#### 2c276b8cb emoji [nfc]: Make realmEmoji internal to EmojiStoreImpl

It's a bit of a trap for the unwary (... like me last month, in
ecd2cb5d4, causing #1113), because it includes deactivated realm emoji
as well as active ones.

#### e56480648 api [nfc]: Rename RealmEmojiItem.id to emojiCode

This way the binding encapsulates an important fact about this
bit of the API -- namely that these values are the same as the
"emoji code" values seen elsewhere in the API -- so that we don't
have to re-verify it when working elsewhere in the codebase and
using this class.

Also edit a doc comment in the model code to take advantage of
that understanding.

#### f09aab1a3 emoji: Exclude deactivated realm emoji from autocomplete

Fixes #1113.

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -115,7 +115,8 @@ class CustomProfileFieldExternalAccountData {
 /// in <https://zulip.com/api/register-queue>.
 @JsonSerializable(fieldRename: FieldRename.snake)
 class RealmEmojiItem {
-  final String id;
+  @JsonKey(name: 'id')
+  final String emojiCode;
   final String name;
   final String sourceUrl;
   final String? stillUrl;
@@ -123,7 +124,7 @@ class RealmEmojiItem {
   final int? authorId;
 
   RealmEmojiItem({
-    required this.id,
+    required this.emojiCode,
     required this.name,
     required this.sourceUrl,
     required this.stillUrl,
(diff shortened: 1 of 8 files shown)


chrisbobbe:
Thanks! LGTM, merging.

Issue #1110, opened 2024-12-07
Icon font differs if built on Windows vs. Linux or macOS


gnprice:
Running `tools/icons/build-icon-font` produces different results when run on Windows vs. when run on Linux or macOS. Specifically, the icons get placed in a different order: `at_sign` precedes `attach_file` on Linux and macOS, but follows it on Windows.

Any ordering is fine, but it needs to be consistent — that's what lets CI get the same results as a developer got on their own machine.

Reported by @sm-sayedi [in chat](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/CI.20failure.20related.20to.20Zulip.20icons.20-.20.23F889/near/1994820).

### Diagnosis / implementation

It looks like the issue is that we take the icons in the order that Node's `fs.readdirSync` gives them to us, and don't sort the list ourselves. That's seemed to mostly work, because Node made the regrettable choice to have `fs.readdir` and `fs.readdirSync` sort the list, unlike the underlying OS primitives:
* https://github.com/nodejs/node/issues/3232

… and, predictably (https://github.com/nodejs/node/issues/3232#issuecomment-146336643), it's ended up sorting inconsistently.

So to fix the issue it should suffice to sort the list ourselves. Let's just sort with `Array.sort`'s default ordering on the icon names, i.e. lexicographic on the raw code units of the strings.

PR #1115, opened 2024-12-09
tools: Make `ZulipIcons` results explicitly sorted

sm-sayedi:
The custom icon entries in "ZulipIcons.dart" used to be generated/sorted based on the default OS order of the related ".svg" files in "assets\icons" directory. This default order behavior is different in different platforms (Windows vs. Linux or macOS) thus resulting in inconsistent output of "ZulipIcons.dart".
This PR adds an explicit sorting mechanism; so the output is consistent across different platforms.

Fixes: #1110

### DIFF
--- a/tools/icons/build-icon-font.js
+++ b/tools/icons/build-icon-font.js
@@ -34,7 +34,8 @@ async function main() {
   const iconFiles = fs
     .readdirSync(srcDir)
     .filter(name => name.endsWith('.svg'))
-    .map(name => path.join(srcDir, name));
+    .map(name => path.join(srcDir, name))
+    .sort(); // prevent inconsistent results across platforms
 
   const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'build-icon-font-'));
 


gnprice:
Thanks! LGTM; merging, with some edits to the commit message:

```
$ git range-diff origin pr/1115 @
1:  a6edbc22e ! 1:  db6310b62 tools: Make `ZulipIcons` results explicitly sorted
    @@ Metadata
     Author: Sayed Mahmood Sayedi <smahmood.sayedi@gmail.com>
     
      ## Commit message ##
    -    tools: Make `ZulipIcons` results explicitly sorted
    +    icons: Explicitly sort in build-icon-font
     
    -    The custom icon entries in "ZulipIcons.dart" used to be generated/sorted
    -    based on the default OS order of the related ".svg" files in
    +    The custom icon entries in "icons.dart" used to be generated/sorted
    +    based on the order returned by Node for the related ".svg" files in
         "assets\icons" directory. This default order behavior is different in
    -    different platforms (Windows vs. Linux or macOS) thus resulting in
    -    inconsistent output of "ZulipIcons.dart".
    -    This PR adds an explicit sorting mechanism; so the output is consistent
    -    across different platforms.
    +    Node's implementation on different platforms (Windows vs. Linux or
    +    macOS) thus resulting in inconsistent output of "icons.dart".
    +    This PR adds an explicit sorting mechanism; so the output is
    +    consistent across different platforms.
[…]
```

In particular the ordering we had been getting doesn't come from the OS; it comes from Node. (The fact that Node inserts such behavior is what the author of that upstream Node issue was lamenting, https://github.com/nodejs/node/issues/3232#issuecomment-146336643 .)

For example here's the order that the files in that directory appear on my machine right now, as returned by the operating system:
```
$ ls --sort=none --format=single-column assets/icons/
arrow_down.svg
ZulipIcons.ttf
star_filled.svg
menu.svg
camera.svg
copy.svg
inbox.svg
read_receipts.svg
language.svg
format_quote.svg
arrow_right.svg
contacts.svg
bot.svg
attach_file.svg
unmute.svg
hash_sign.svg
globe.svg
topic.svg
user.svg
mute.svg
hash_italic.svg
smile.svg
at_sign.svg
clock.svg
star.svg
chevron_right.svg
group_dm.svg
send.svg
lock.svg
message_feed.svg
share_ios.svg
share.svg
image.svg
```

The `--sort=none` option to `ls` tells it to skip the sorting that by default it normally does. You can see that the order looks pretty arbitrary.

Issue #1116, opened 2024-12-09
Can't upgrade firebase_core to 3.4.0, because macOS deployment target


gnprice:
When I try upgrading our dependencies with `tools/upgrade pub`, I get the following error from CocoaPods:

```
+ pod update --project-directory=macos/
Update all pods
Updating local specs repositories
Analyzing dependencies
firebase_core: Using Firebase SDK version '11.0.0' defined in 'firebase_core'
firebase_messaging: Using Firebase SDK version '11.0.0' defined in 'firebase_core'
[!] CocoaPods could not find compatible versions for pod "Firebase/Messaging":
  In Podfile:
    firebase_messaging (from `Flutter/ephemeral/.symlinks/plugins/firebase_messaging/macos`) was resolved to 15.1.0, which depends on
      Firebase/Messaging (~> 11.0.0)

Specs satisfying the `Firebase/Messaging (~> 11.0.0)` dependency were found, but they required a higher minimum deployment target.
```

This appears to correspond to the following release note:
https://firebase.google.com/support/release-notes/ios#version_1100_-_july_30_2024
saying version 11.0.0 of the Firebase Apple SDK started requiring a minimum deployment target of macOS 10.15.

Naturally we're happy to bump up our minimum macOS deployment target — 10.15 is from 2019, and even increasing to last year's macOS 14 would be perfectly fine. But I spent a few minutes just now attempting to do so — editing the `MACOSX_DEPLOYMENT_TARGET = 10.14;` lines in the `project.pbxproj` file so they say 10.15 (or 13.0) instead, then also tried setting it in Xcode which added some new lines of that form — and the error persisted. (Perhaps there's somewhere *else* that `pod update` is looking? Perhaps it has a cache of some kind that it fails to update?)

So further debugging is required.

Meanwhile, we'll be holding back `firebase_core` at 3.3.0 rather than 3.4.0 (and so holding back the other Firebase-related packages too), because that's the upgrade that triggers this.

sher999:
<details><summary>Log</summary>

```pod update
Update all pods
Updating local specs repositories
Analyzing dependencies
firebase_core: Using Firebase SDK version '11.0.0' defined in 'firebase_core'
firebase_messaging: Using Firebase SDK version '11.0.0' defined in 'firebase_core'
Downloading dependencies
Installing Firebase (11.0.0)
Installing FirebaseCore (11.0.0)
Installing FirebaseCoreInternal (11.6.0)
[...]

Issue #773, opened 2024-06-28
Run iOS build in CI


gnprice:
Similar to #772, but for `flutter build ipa` instead of `flutter build apk`.

I don't think we've yet had any problems that this would have caught, unlike the Android case, but clearly in principle it can happen for the same kinds of reasons. It is somewhat less likely, though, because we currently write much less code of our own on the native iOS side than the native Android side.

This might be more annoying to do than the Android, because Apple's developer tools are often harder to wrangle. Or it might not be; we'll find out.

Given the costs and benefits, probably the right approach is:
* Let's spend a small amount of time — like no more than a couple of hours' work — exploring this in the near term, while the question is fresh in mind. Try to see if we can easily arrange in GitHub Actions to have a working `flutter build ipa` as part of our CI, and without it making our CI runs take vastly longer.
* If that small amount of time doesn't lead to success, then leave this issue open and move it to a later milestone.

gnprice:
Here's an example where a version of this would have been useful — a deps upgrade broke the iOS build:
https://github.com/zulip/zulip-flutter/pull/1117#issuecomment-2529389420

There is a wrinkle, though: when I try `flutter build ipa` at the same commit, it actually succeeds. Meanwhile `flutter run` targeting an iOS device fails. So ideally we'd either include `flutter run` in CI, or find some other command that reproduces that PR's build failure and include that in CI.

gnprice:
Moved this to a nearer milestone: the iOS build will become more interesting soon with #1379 introducing some nontrivial Swift code, and that increases the value of having it in CI (as @rajveermalviya points out at https://github.com/zulip/zulip-flutter/pull/1444#discussion_r2015664596).

If we further add a check that none of the files in the repo got modified, then that will also help catch side effects that should be included in an upgrade commit (as also pointed out in https://github.com/zulip/zulip-flutter/pull/1444#discussion_r2015664596).

thet1nk3r3r:
Hey @gnprice !

I went through the earlier context around this (#1117 and #1444). In #1117 we saw an iOS issue that showed up via `flutter run` while `flutter build ipa` still succeeded (and was resolved by `flutter clean`), so a simple archive build wouldn’t have caught that specific case. On the other hand, #1444 showed that iOS builds can introduce important generated-file updates (Podfile.lock, Xcode project files) that were missed because iOS wasn’t built as part of the upgrade flow something CI could plausibly help catch, especially as more nontrivial Swift code is added.

[...]

PR #1117, opened 2024-12-09
Upgrade Flutter and packages; hold back firebase_core; record Flutter commit IDs

### LINKED ISSUE #1118: Record Flutter commit ID as well as version number
This issue can be seen as a much-scoped-down version of pinning the Flutter version:
* #15

That issue has stalled because the solutions to it tend to involve inserting themselves into the day-to-day development workflow — after all, they need to control what version of Flutter gets used — and none of the known solutions have felt totally satisfactory to me. Also because, in practice, our existing workflow without actual pinning generally works just fine.

There is one aspect that's bugged me a few times so far in practice, though, and which doesn't call for changing anything in the day-to-day workflow so it should be much easier to resolve: I'd like to know the Git commit ID for the Flutter version that we were using at a specific point in the repo's history. Specifically in practice, I want to know what Flutter commit ID was used for a given published (beta) release.

I already make a practice of building each release with the exact minimum version recorded in pubspec.yaml, in order to make it reproducible. So this information is there in principle, encoded in that version number. But from those Flutter version numbers (`3.27.0-1.0.pre.518`, in our `pubspec.yaml` in main at the moment) it's somewhat annoying to identify the actual Git commit ID (`01590aa27ac4d3a904d25ae4c6b792f933a3da6f` in that example, i.e. flutter/flutter@01590aa27ac4d3a904d25ae4c6b792f933a3da6f).

We can solve this by just having the commit ID also in pubspec.yaml, e.g. in a comment on the same line. Then `tools/upgrade` can maintain the comment. Ideally, we can even have a check in `tools/check` that verifies it's accurate.

### LINKED ISSUE #1119: Dart SDK change breaks background notifications
On upgrading to Flutter from current main, and running in release mode, the app no longer shows notifications on Android when it was in the background. This burst of error messages appears in logcat:

```
DartVM   E  ERROR: To access 'package:zulip/notifications/receive.dart::NotificationService' from native code, it must be annotated.
            ERROR: See https://github.com/dart-lang/sdk/blob/master/runtime/docs/compiler/aot/entry_point_pragma.md
flutter  E  [ERROR:flutter/lib/ui/dart_runtime_hooks.cc(38)] Dart Error: ERROR: To access 'package:zulip/notifications/receive.dart::NotificationService' from native code, it must be annotated.
            ERROR: See https://github.com/dart-lang/sdk/blob/master/runtime/docs/compiler/aot/entry_point_pragma.md
flutter  E  [ERROR:flutter/shell/common/shell.cc(115)] Dart Error: ERROR: To access 'package:zulip/notifications/receive.dart::NotificationService' from native code, it must be annotated.
            ERROR: See https://github.com/dart-lang/sdk/blob/master/runtime/docs/compiler/aot/entry_point_pragma.md
flutter  E  [ERROR:flutter/shell/common/shell.cc(115)] Dart Error: ERROR: To access 'package:zulip/notifications/receive.dart::NotificationService' from native code, it must be annotated.
                                                                                                    ERROR: See https://github.com/dart-lang/sdk/blob/master/runtime/docs/compiler/aot/entry_point_pragma.md
```

It looks like the relevant upstream commit is probably this one:
https://github.com/dart-lang/sdk/commit/9b06e26620a081c89931e5618336ff23dc6083be
which is also the latest commit to touch that linked docs page.

AFAICT those docs don't actually say that we should need to annotate the class. We already do annotate the static *method* that we ask native code to call:
```dart
  @pragma('vm:entry-point')
  static Future<void> _onBackgroundMessage(FirebaseRemoteMessage message) async {
```
Well, at least the error message is nice and clear, though. We'll annotate the class too.

gnprice:
Fixes #1118.
Fixes #1119.

This upgrades Flutter and some of our other dependencies.

OTOH `firebase_core` and the other Firebase packages can't immediately be upgraded; I ran into #1116, which I just filed. So hold those back.

And at the start of the branch, I augment `tools/upgrade` a bit so that it records the actual commit ID in the Flutter tree in a comment in pubspec.yaml, and give `tools/check` a tiny new suite `flutter_version` that validates that information.

On several occasions I've wanted the ability to look back and see what Flutter commit was used in a given published (beta) release, in particular. I already make a practice of doing so with the exact minimum version recorded in `pubspec.yaml`, in order to make it reproducible; but it's somewhat annoying to identify that commit from the version number that normally appears there. So this solves that problem. (… OK, and filed #1118 to describe that.)

# Selected commit messages

#### f5395d990 tools/upgrade: Record commit ID of Flutter into pubspec.yaml

Fixes #1118, or anyway the main part of it.

Upcoming commits will add a suite in tools/check that validates
this information, in case this line gets updated manually rather
than using this script.

#### a29b6143a notif: Add vm:entry-point pragma on class, too

Fixes #1119.
[…]

#### b1e77284e deps: Upgrade packages within constraints (tools/upgrade pub)

This required a manual fixup step to update generated files
for the Drift upgrade:
  $ tools/check --fix --all-files build_runner drift

### DIFF
--- a/tools/check
+++ b/tools/check
@@ -434,8 +434,6 @@ describe_git_head() {
 }
 
 print_header() {
-    local flutter_executable flutter_tree
-
     echo "Test command: ${orig_cmdline}"
 
     echo "Time now: $(date --utc +'%F %T %z')"
@@ -445,9 +443,7 @@ print_header() {
     # We avoid `flutter --version` because, weirdly, when run in a
     # GitHub Actions step it takes about 30 seconds.  (The first time;
     # it's fast subsequent times.)  That's even after `flutter precache`.
-    flutter_executable=$(readlink -f "$(type -p flutter)")
-    flutter_tree=${flutter_executable%/bin/flutter}
-    describe_git_head "flutter/flutter" "${flutter_tree}"/.git
+    describe_git_head "flutter/flutter" "$(flutter_tree)"/.git
 
     dart --version
 }
--- a/tools/check
+++ b/tools/check
@@ -26,6 +26,7 @@ this_dir=${BASH_SOURCE[0]%/*}
 
 default_suites=(
     analyze test
+    flutter_version
     build_runner l10n drift pigeon icons
     android  # This takes multiple minutes in CI, so do it last.
 )
@@ -82,7 +83,7 @@ while (( $# )); do
         --all) opt_files=all; opt_all=1; shift;;
         --fix) opt_fix=1; shift;;
         --verbose) opt_verbose=1; shift;;
-        analyze|test|build_runner|l10n|drift|pigeon|icons|android|shellcheck)
+        analyze|test|flutter_version|build_runner|l10n|drift|pigeon|icons|android|shellcheck)
             opt_suites+=("$1"); shift;;
         *) usage;;
     esac
@@ -216,6 +217,49 @@ run_analyze() {
 run_test() {
     # no `flutter test --verbose` even when $opt_verbose; it's *very* verbose
     flutter test
+}
+
+# Check the Flutter version in pubspec.yaml is commented with a commit ID,
+# and the commit is from upstream main.
+run_flutter_version() {
+    # Omitted from this files check:
+    #   tools/check
+    files_check pubspec.yaml \
+    || return 0
+
+    local flutter_tree flutter_git
+    flutter_tree=$(flutter_tree)
+    flutter_git=( git --git-dir="${flutter_tree}"/.git )
+
+    # Parse our Flutter version spec with its commit-ID comment.
+    local flutter_commit
+    flutter_commit=$(
+        perl <pubspec.yaml -0ne '
+             print $1 if (/^  sdk: .*\n  flutter: \S+\s*# ([0-9a-f]{40})$/m)'
+    ) || return
+    if [ -z "${flutter_commit}" ]; then
+        echo >&2 "error: Flutter commit spec not found in pubspec.yaml"
+        return 1
+    fi
+
+    # Check the commit is an acceptable commit.
+    local commit_count
+    commit_count=$(
+        "${flutter_git[@]}" rev-list --count origin/main.."${flutter_commit}"
+    ) || return
+    if (( "${commit_count}" )); then
+        cat >&2 <<EOF
+error: Flutter commit spec in pubspec.yaml is not from upstream main.
+Commit ${flutter_commit} has ${commit_count} commits not in main:
+EOF
+        "${flutter_git[@]}" log --oneline --reverse \
+            origin/main.."${flutter_commit}"
+        return 1
+    fi
+
+    if_verbose echo "OK Flutter commit ${flutter_commit}"
+
+    return 0
 }
 
 # Whether the build_runner suite should run, given $opt_files.
@@ -456,6 +500,7 @@ for suite in "${opt_suites[@]}"; do
     case "$suite" in
         analyze)      run_analyze ;;
         test)         run_test ;;
+        flutter_version) run_flutter_version ;;
         build_runner) run_build_runner ;;
         l10n)         run_l10n ;;
         drift)        run_drift ;;
--- a/tools/check
+++ b/tools/check
@@ -220,7 +220,7 @@ run_test() {
 }
 
 # Check the Flutter version in pubspec.yaml is commented with a commit ID,
-# and the commit is from upstream main.
+# which agrees with the version, and the commit is from upstream main.
 run_flutter_version() {
     # Omitted from this files check:
     #   tools/check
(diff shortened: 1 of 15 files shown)


gnprice:
CI failed initially but should be fixed now. The fix was adding this commit:
9618f40aa check: Fix branch name for clone of Flutter repo in CI

tools/check
```diff
@@ -218,6 +219,84 @@ run_test() {
...
+We therefore expect the Flutter version name to be: ${predicted_version}
+But the Flutter version bound in pubspec.yaml is: ${flutter_version}
+EOF
+        return 1
+    fi
+
+    # Check the commit is an acceptable commit.
+    local commit_count
+    commit_count=$(
+        "${flutter_git[@]}" rev-list --count origin/main.."${flutter_commit}"
```
chrisbobbe, line 283: I've named the remote "upstream" instead of "origin", so "origin/main" doesn't work for me (here and below). I've named it back to "origin" for testing this new code; should I keep it that way or should the code adapt to handle the "upstream" name?

chrisbobbe (approved):
Thanks! LGTM with one small question below.

chrisbobbe:
Oh I did notice an auto-generated change:

```
diff --git ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
index 5e31d3d34..c53e2b314 100644
--- ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -48,6 +48,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
+      enableGPUValidationMode = "1"
       allowLocationSimulation = "YES">
       <BuildableProductRunnable
          runnableDebuggingMode = "0">
```

tools/check
gnprice: I'd suggest just switching back to "origin". It'd be possible to make this handle both conventions, but would make the logic more complicated.

I don't think the Flutter install steps do anything to override Git's default here, so most contributors will have "origin".

gnprice: (Personally I switched back from using "upstream" to using "origin" a little while ago, for Git repos in general, partly just because I find "origin" is quicker for me to reliably type — "upstream" has long runs on the same hand.)

chrisbobbe:
Hmm, seeing some iOS build issues actually—please hold off merging while I investigate. Here is the output:

```
Launching lib/main.dart on iPhone 15 Pro in debug mode...
Running pod install...
Running Xcode build...
Xcode build done.                                           30.5s
Failed to build iOS app
Semantic Issue (Xcode): Duplicate interface definition for class 'Sqlite3FlutterLibsPlugin'
/Users/chrisbobbe/dev/zulip-flutter/build/ios/Debug-iphonesimulator/sqlite3_flutter_libs/sqlite3_flutter_libs.framework/Headers/Sqlite3FlutterLibsPlugin.h:2:0

Modules Issue (Xcode): Definition of 'Sqlite3FlutterLibsPlugin' must be imported from module 'sqlite3_flutter_libs.Swift' before it is required
/Users/chrisbobbe/dev/zulip-flutter/ios/Runner/GeneratedPluginRegistrant.m:105:3

Could not build the application for the simulator.
Error launching application on iPhone 15 Pro.
```

chrisbobbe:
Hmm, the build succeeded just now when I requested it from the Xcode gui. The output above is from `flutter run` as triggered from the "Run 'main.dart'" button (with a green "play" icon) in Android Studio.

chrisbobbe:
> The output above is from flutter run as triggered from the "Run 'main.dart'" button (with a green "play" icon) in Android Studio.

OK, and it succeeds for me now after running `flutter clean`. 🤷‍♂️ Please merge at will, then.

gnprice:
Thanks for the review! Evidently I had not tried `flutter run` for iOS. <_< >_>

I reproduce that build failure, and also reproduce that `flutter clean` resolves it — good find.

Before `flutter clean`, I tried `flutter build ipa`, prompted by thinking of #773. That actually succeeded. More data points FTR though I'm not sure they'll be useful:
* `flutter run --release` succeeded too
* but OTOH `flutter build ipa --debug` still succeeded, so wouldn't have been a way to detect the issue.

gnprice:
Merging, after adding this:

> Oh I did notice an auto-generated change from `flutter run`:

There was also a similar change in `macos/` after `flutter run` for that. I'd noticed that change yesterday — I was experimenting with the macOS target to try to resolve #1116 and unblock upgrading Firebase — but had assumed it was an existing discrepancy that we just hadn't noticed because we don't use that target as often.

PR #1125, opened 2024-12-10
Series of prep commits for topic action sheet related changes

PIG208:
Taken from #1041.

Other PRs that rely on this:
- #1103
- #1076

Requested by @gnprice [here](https://github.com/zulip/zulip-flutter/pull/1041#issuecomment-2530404897).

### DIFF
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -88,7 +88,7 @@ void showMessageActionSheet({required BuildContext context, required Message mes
                     borderRadius: BorderRadius.circular(7),
                     child: Column(spacing: 1,
                       children: optionButtons))))),
-              const MessageActionSheetCancelButton(),
+              const ActionSheetCancelButton(),
             ])));
     });
 }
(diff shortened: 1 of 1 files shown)


lib/widgets/message_list.dart
```diff
@@ -1273,7 +1273,7 @@ class MessageWithPossibleSender extends StatelessWidget {
 
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
-      onLongPress: () => showMessageActionSheet(context: context, message: message),
+      onLongPress: () => showMessageActionSheet(context, message: message),
```
gnprice, line 1276: I guess this one last commit doesn't convince me:
53a24b043 action_sheet [nfc]: Make showMessageActionSheet's context an unnamed param

The existing API there matches that of upstream's `showModalBottomSheet`. I think in turn the idea there is that it's similar in flavor to a widget constructor; and those always take all named arguments.

gnprice (commented):
Thanks!

I'll go ahead and merge this. @chrisbobbe I believe you were done with reviews on this part of #1041, based on your last couple of reviews there.

The tiny last commit at the end, I'm not sure I like the API change it makes. (Comment below.) So I'll leave that one out.

gnprice:
Merged. @rajveermalviya FYI — this version has fixed that issue you discovered when combining #1103 with the corresponding commits in a previous revision of #1041, so I expect a merge conflict but it should be easy to fix by just dropping the commits that came from #1041 when you rebase #1103 atop the new main.

Issue #1123, opened 2024-12-10
Put emoji picker's query and best matches at bottom instead of top, maybe


gnprice:
That way they'd be in easy reach. In particular when using a phone with one's thumbs, the top of the screen can often be significantly harder to reach than the bottom (or middle, just above the keyboard).

This is a trickier cousin of:
* #1121 

Trickier because it's not obvious just how the resulting UI should look, visually.

Tricky also because there's a shortage of familiar patterns for it in other apps. I've seen a couple of examples of mobile apps putting the *query* at the bottom in an interaction like this — in particular both Firefox and Safari's mobile browsers put the location bar there. But even in those two examples, the top *result* is still at the top.

Letmedecodeher:
Can i try to resolve this?

PR #1131, opened 2024-12-11
autocomplete: Put best matches near input field.

### LINKED ISSUE #1121: Put best autocomplete matches at bottom of list, not top
That way they're in easy reach. In particular when using a phone with one's thumbs, the top of the screen can often be significantly harder to reach.

Because the query is at the bottom of the screen (in the compose box) and the list of autocomplete options opens upward from there, putting the best matches at the bottom is also logical because it puts them right next to the query (the one they're the best match for).

This applies to both ComposeAutocomplete (for autocompleting @-mentions, emoji, and so on from the content input) and TopicAutocomplete (for autocompleting topics from the topic input).

apoorvapendse:
This commit reverses the list that was originally
presented to the user while showing the typeahead
menu.

This makes sense since on mobile its easier to click on options closer to the input box, i.e. where your fingers are currently present, instead of pressing arrow keys to navigate through the options on a keyboard which is convenient on a desktop setup.

Hence we place the best matching options not at the top of the typeahead menu, but instead at the bottom for better reachability and convenience of the user.

[CZO](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Put.20best.20autocomplete.20matches.20at.20bottom.20of.20list.20.23F1121/near/1999427)

Fixes #1121.

Here is a demo:
[Screencast from 12-11-2024 12:38:19 PM.webm](https://github.com/user-attachments/assets/9a299840-9f9e-4f90-ae87-4ec9d2deb05d)

image
image

### DIFF
--- a/test/widgets/autocomplete_test.dart
+++ b/test/widgets/autocomplete_test.dart
@@ -177,6 +177,38 @@ void main() {
 
       debugNetworkImageHttpClientProvider = null;
     });
+
+    testWidgets('options are shown in reversed order', (tester) async {
+      final users = List.generate(8, (i) => eg.user(fullName: 'A$i', avatarUrl: 'user$i.png'));
+      final composeInputFinder = await setupToComposeInput(tester, users: users);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(composeInputFinder, 'hello @');
+      await tester.enterText(composeInputFinder, 'hello @A');
+      await tester.pump();
+      // Add an extra pump to account for any potential frame delays introduced
+      // by the post frame callback in RawAutocomplete's implementation.
+      await tester.pump();
+      final initialPosition = tester.getTopLeft(find.text(users.first.fullName)).dy;
+      // Initially, all but the last autocomplete options are visible.
+      checkUserShown(users.last, store, expected: false);
+      users.take(7).forEach((user) => checkUserShown(user, store, expected: true));
+
+      // Can't scroll down because the options grow from the bottom.
+      await tester.drag(find.byType(ListView), const Offset(0, -50));
+      await tester.pump();
+      check(tester.getTopLeft(find.text(users.first.fullName)).dy)
+        .equals(initialPosition);
+
+      // The last autocomplete option becomes visible after scrolling up.
+      await tester.drag(find.byType(ListView), const Offset(0, 200));
+      await tester.pump();
+      users.skip(1).forEach((user) => checkUserShown(user, store, expected: true));
+      checkUserShown(users.first, store, expected: false);
+
+      debugNetworkImageHttpClientProvider = null;
+    });
   });
 
   group('emoji', () {
@@ -249,6 +281,66 @@ void main() {
 
       debugNetworkImageHttpClientProvider = null;
     });
+
+ testWidgets('emoji options appear in the reverse order and do not scroll down', (tester) async {
+    final composeInputFinder = await setupToComposeInput(tester);
+    final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+    store.setServerEmojiData(
+      ServerEmojiData(
+        codeToNames: {
+          '1f4a4': ['zzz'], // Unicode emoji for "zzz"
+          '1f52a': ['biohazard'],
+          '1f92a': ['zany_face'],
+          '1f993': ['zebra'],
+          '0030-fe0f-20e3': ['zero'],
+          '1f9d0': ['zombie'],
+          }));
+
+    await store.handleEvent(
+      RealmEmojiUpdateEvent(
+        id: 1,
+        realmEmoji: {
+          '1': eg.realmEmojiItem(emojiCode: '1', emojiName: 'buzzing')}));
+
+    final emojiNames = ['zulip', 'zany_face', 'zebra', '💤', 'zombie', 'zero', 'buzzing', 'zzz', 'biohazard'];
+
+    // Enter a query; options appear, of all three emoji types.
+    // TODO(#226): Remove this extra edit when this bug is fixed.
+    await tester.enterText(composeInputFinder, 'hi :');
+    await tester.enterText(composeInputFinder, 'hi :z');
+    await tester.pump();
+    // Add an extra pump to account for any potential frame delays introduced
+    // by the post frame callback in RawAutocomplete's implementation.
+    await tester.pump();
+
+    final firstEmojiInitialPosition = tester.getTopLeft(find.text(emojiNames[0])).dy;
+    final listViewFinder = find.byType(ListView);
+
+    emojiNames.take(8).forEach((emojiName) => check(find.text(emojiName)).findsOne());
+    check(find.text(emojiNames.last)).findsNothing();
+
+    await tester.drag(listViewFinder, const Offset(0, -50));
+    await tester.pump();
+    final firstEmojiPositionAfterScrollDown = tester.getTopLeft(find.text(emojiNames[0])).dy;
+    check(
+      because: "ListView options should not scroll down further than initial position",
+      firstEmojiInitialPosition,
+    ).equals(firstEmojiPositionAfterScrollDown);
+
+    check(
+      because: "The last emoji should not be visible before scrolling up",
+      find.text(emojiNames.last),
+    ).findsNothing();
+
+    await tester.drag(listViewFinder, const Offset(0, 50));
+    await tester.pump();
+
+    check(because: "The last emoji should be visible after scrolling up",
+      find.text(emojiNames.last)).findsOne();
+
+    debugNetworkImageHttpClientProvider = null;
+  });
 
     testWidgets('text emoji means just show text', (tester) async {
       final composeInputFinder = await setupToComposeInput(tester);
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -134,6 +134,7 @@ class _AutocompleteFieldState<QueryT extends AutocompleteQuery, ResultT extends 
               constraints: const BoxConstraints(maxHeight: 300), // TODO not hard-coded
               child: ListView.builder(
                 padding: EdgeInsets.zero,
+                reverse: true,
                 shrinkWrap: true,
                 itemCount: _resultsToDisplay.length,
                 itemBuilder: _buildItem))));


PIG208 (commented):
Thanks for the PR @apoorvapendse! Left some comment.

This doesn't fix #1123 because the emoji picker (added in https://github.com/zulip/zulip-flutter/pull/1103) is a feature separate from autocompletion.

test/widgets/autocomplete_test.dart
```diff
@@ -177,6 +177,67 @@ void main() {
...
+        eg.user(userId: 7, fullName: 'Asif', avatarUrl: 'user7.png'),
+        eg.user(userId: 8, fullName: 'Asim', avatarUrl: 'user8.png')
+      ];
+
+      final composeInputFinder = await setupToComposeInput(tester, users:users);
+      final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+
+      final expectedUserSequence = users;
+      // Options are filtered correctly for query
+      // TODO(#226): Remove this extra edit when this bug is fixed.
```
PIG208, line 198: In https://github.com/zulip/zulip-flutter/pull/1131/commits/b35ddbf51f9efacbde2dbe84660dbe36aba1ebf0#r1889246753, I was referring to this comment here.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +308,96 @@ void main() {
...
+      zanyFaceOptionLabel,
+      zebraOptionLabel,
+      zeroOptionLabel,
+      zombieOptionLabel,
+      buzzingOptionLabel,
+      // biohazardOptionLabel, this won't be rendered in the list initally since it is the 7th option.
+    ];
+
+    await tester.enterText(composeInputFinder, 'hi :');
+    await tester.enterText(composeInputFinder, 'hi :z');
```
PIG208, line 360: Oh I see. I didn't include the second `tester.enterText` in the comment. I believe that #227 also applies here because two `tester.enterText` calls are needed.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +308,98 @@ void main() {
...
+
+    check(because:"ListView should not scroll down because it is already at the bottom",
+    scrollOffsetAfterDragDown).equals(initialScrollOffset);
+
+    final biohazardFinder = find.text(biohazardOptionLabel);
+    check(because: "The biohazard emoji should not be visible before scrolling up",biohazardFinder).findsNothing();
+
+    // Scroll up
+    await tester.drag(listViewFinder, const Offset(0, 50));
+    await tester.pumpAndSettle();
```
PIG208, line 392: I tried adding `print`'s before and after this, but it seems like `tester.getTopLeft(listViewFinder).dy` doesn't actually get the scroll offset, as the value also remains the same:
```dart
    print(tester.getTopLeft(listViewFinder).dy); // prints 196.0
    await tester.drag(listViewFinder, const Offset(0, 50));
    await tester.pumpAndSettle();
    print(tester.getTopLeft(listViewFinder).dy); // prints 196.0
```
This probably means that all the other checks for `scrollOffset` aren't really effective. It might be good to ask about this in #mobile-dev-help.

PIG208, line 392: This can be replaced with `await tester.pump()`.

PIG208 (commented):
Thanks for the update! Left some more comments. I read the first test and skimmed the second one. It looks like the current approach of getting scroll offset is not working properly.

For the next revision, please do a round of self-review addressing the style issues mentioned in the comments, and apply it to other places as well.

test/widgets/autocomplete_test.dart
```diff
@@ -177,6 +177,71 @@ void main() {
...
+      // 8th user should be above 7th user
+      check(because: "8th user should be above 7th user",
+      tester.getTopLeft(
+        find.text(users.last.fullName)).dy
+      ).isLessThan(
+        tester.getTopLeft(
+        find.text(users[users.length - 2].fullName)
+      ).dy);
+      debugNetworkImageHttpClientProvider = null;
+    });
```
PIG208, line 244: The formatting looks good, but I think this test can be simplified.

The purpose our testing is to make sure we don't break a feature change. In this case, it is the change that the autocomplete options are reversed. So if the test fails when I remove the `reversed: true` line, which it does, the test is effective.

However, it would also be nice if we can simplify the test. Currently, it does 3 checks:
1. verify that initially, the first 7 options are shown and we cannot scroll down; 2. check that the options appear before one another, matching the expected order; 3. check that scrolling up, the last option is shown and the first one is hidden.

Among them, I think only 1 and 3 are necessary for the sake of testing `reversed: true`. They happen to check scrolling up and down, which are quite relevant to this feature. 2 would enforce the test at the cost of making it a bit more complicated, but the test is already sufficient without it.

To make the test more readable, we should try to minimize lines of code that set up the test, and highlight those that are more interesting, like the ones checking if an option is visible or not. Here's an updated version of this test:

```dart
    testWidgets('options are shown in reversed order', (tester) async {
      final users = List.generate(8, (i) => eg.user(fullName: 'A$i', avatarUrl: 'user$i.png'));
      final composeInputFinder = await setupToComposeInput(tester, users: users);

      // TODO(#226): Remove this extra edit when this bug is fixed.
      await tester.enterText(composeInputFinder, 'hello @');
      await tester.enterText(composeInputFinder, 'hello @A');
      await tester.pump();

      final initialPosition = tester.getTopLeft(find.text(users.first.fullName)).dy;
      // Initially, all but the last autocomplete options are visible.
      checkUserShown(users.last, store, expected: false);
      users.take(7).forEach((user) => checkUserShown(user, store, expected: true));

      // Can't scroll down because the options grow from the bottom.
      await tester.drag(find.byType(ListView), const Offset(0, -50));
      await tester.pump();
      check(tester.getTopLeft(find.text(users.first.fullName)).dy)
        .equals(initialPosition);

      // The last autocomplete option becomes visible after scrolling up.
      await tester.drag(find.byType(ListView), const Offset(0, 200));
      await tester.pump();
      users.skip(1).forEach((user) => checkUserShown(user, store, expected: true));
      checkUserShown(users.first, store, expected: false);

    })
```

PIG208 (commented):
Thanks for the update! I revised the @-mentions test as an example of what we usually expect for testing. Please read it and try to apply the same approach to the similar test for emojis.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,68 @@ void main() {
...
+
+    // Enter a query; options appear, of all three emoji types.
+    // TODO(#226): Remove this extra edit when this bug is fixed.
+    await tester.enterText(composeInputFinder, 'hi :');
+    await tester.enterText(composeInputFinder, 'hi :z');
+    await tester.pump();
+
+    final positions = emojiSequence.take(8).map((icon) {
+      final finder = find.text(icon);
+      check(because:"Each emoji option should be rendered", finder).findsOne();
```
PIG208, line 311: Let's remove the `because` for this check. `findsOne` should be evident that we expect the option to be rendered.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,68 @@ void main() {
...
+      return tester.getTopLeft(finder).dy;
+    }).toList();
+
+    await tester.drag(find.byType(ListView), const Offset(0, -50));
+    await tester.pump();
+    final firstEmojiPositionAfterScrollDown = tester.getTopLeft(find.text(emojiSequence[0])).dy;
+    check(
+      because:"ListView options should not scroll down further than initial position",
```
PIG208, line 319: There is still the spacing issue — there should be a space here `because: "..."`

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,68 @@ void main() {
...
+      positions.first
+    ).equals(firstEmojiPositionAfterScrollDown);
+
+    check(
+      because: "The last emoji should not be visible before scrolling up"
+      ,find.text(emojiSequence.last)
```
PIG208, line 325: Usually we don't start lines with `,`. Format list of arguments using trailing commas instead.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,68 @@ void main() {
...
+    // Enter a query; options appear, of all three emoji types.
+    // TODO(#226): Remove this extra edit when this bug is fixed.
+    await tester.enterText(composeInputFinder, 'hi :');
+    await tester.enterText(composeInputFinder, 'hi :z');
+    await tester.pump();
+
+    final positions = emojiSequence.take(8).map((icon) {
+      final finder = find.text(icon);
+      check(because:"Each emoji option should be rendered", finder).findsOne();
+      return tester.getTopLeft(finder).dy;
```
PIG208, line 312: We only need the position of the first element, so this `map` can be simplified, similar to the other test case added.

PIG208 (commented):
Thanks for the update. I think the test for emojis needs to be more concise. Maybe try to follow the other one closer as they are testing the same aspect of autocompletion but in different contexts.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,66 @@ void main() {
...
+
+    final emojiSequence = ['zulip', 'zany_face', 'zebra', 'zzz', '💤', 'zombie', 'zero', 'buzzing', 'biohazard'];
+
+    // Enter a query; options appear, of all three emoji types.
+    // TODO(#226): Remove this extra edit when this bug is fixed.
+    await tester.enterText(composeInputFinder, 'hi :');
+    await tester.enterText(composeInputFinder, 'hi :z');
+    await tester.pump();
+
+    final firstEmojiInitialPosition = tester.getTopLeft(find.text(emojiSequence[0]),warnIfMissed: true).dy;
```
PIG208, line 309: `true` should be the default value, so we can omit `warnIfMissed: true` here.

PIG208: I see that you removed the `emojiSequence.take(8).map` altogether. We do still need the check for `(icon) => find.text(icon).findsOne()`, (and additionally a `find.text(emojiSequence.last)).findsNothing()`, moving the check from line 320 to here should work).

Having all these check help the reader understand that only the first few options are initially visible, so it makes sense to check again later after scrolling the list. See the other test for reference, as they are quite similar.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,66 @@ void main() {
...
+      because: "The last emoji should not be visible before scrolling up",
+      find.text(emojiSequence.last)
+    ).findsNothing();
+
+    // Scroll up
+    await tester.drag(listViewFinder, const Offset(0, 50));
+    await tester.pump();
+
+    check(because: "The biohazard emoji should be visible after scrolling up",
+    find.text(emojiSequence.last)).findsOne();
```
PIG208, line 330: nit: indentation

This `find` is a part of the `check` and should be indented.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,66 @@ void main() {
...
+
+    final firstEmojiPositionAfterScrollUp = tester.getTopLeft(find.text(emojiSequence[0])).dy;
+    check(because: "Scrolling up should reveal other emoji matches",
+    firstEmojiPositionAfterScrollUp).isGreaterOrEqual(firstEmojiInitialPosition);
```
PIG208, line 334: nit: similar indentation issue here too

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,66 @@ void main() {
...
+
+    debugNetworkImageHttpClientProvider = null;
+
```
PIG208, line 337: nit: remove extra newline

PIG208 (commented):
Thanks for the update! Left some more comments.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,69 @@ void main() {
...
+    // Scroll up
+    await tester.drag(listViewFinder, const Offset(0, 50));
+    await tester.pump();
+
+    check(because: "The biohazard emoji should be visible after scrolling up",
+      find.text(emojiSequence.last)).findsOne();
+
+    final firstEmojiPositionAfterScrollUp = tester.getTopLeft(find.text(emojiSequence[0])).dy;
+    check(because: "Scrolling up should reveal other emoji matches",
+      firstEmojiPositionAfterScrollUp).isGreaterOrEqual(firstEmojiInitialPosition);
```
PIG208, line 338: Let's skip the check for the position here because it overlaps with the `findsOne` check above.

test/widgets/autocomplete_test.dart
```diff
@@ -247,6 +277,69 @@ void main() {
...
+      because: "ListView options should not scroll down further than initial position",
+      firstEmojiInitialPosition
+    ).equals(firstEmojiPositionAfterScrollDown);
+
+    check(
+      because: "The last emoji should not be visible before scrolling up",
+      find.text(emojiSequence.last)
+    ).findsNothing();
+
+    // Scroll up
```
PIG208, line 329: nit: this comment does not seem quite useful, as it just repeats the code below it

PIG208 (commented):
Commit message nit:

```
Tests have been added to verify the
emoji and mention render behavior.
```

Normally we include tests whenever there is behavior change, so we can remove this paragraph.

```
Co-authored-by: Zixuan James Li <zixuan@zulip.com>
Fixes: https://github.com/zulip/zulip-flutter/issues/1121
```
should be:
```
Co-authored-by: Zixuan James Li <zixuan@zulip.com>
Fixes: #1121
```
The lines started with `Something:` are conventionally placed at the end as an agreed upon metadata structure for commits; there is no strict rules on what types of metadata you can have, but usually they are in this form.

Left some comments, mostly nits.

test/widgets/autocomplete_test.dart
```diff
@@ -250,6 +282,66 @@ void main() {
...
+
+    emojiNames.take(8).forEach((emojiName) => check(find.text(emojiName)).findsOne());
+    check(find.text(emojiNames.last)).findsNothing();
+
+    await tester.drag(listViewFinder, const Offset(0, -50));
+    await tester.pump();
+    final firstEmojiPositionAfterScrollDown = tester.getTopLeft(find.text(emojiNames[0])).dy;
+    check(
+      because: "ListView options should not scroll down further than initial position",
+      firstEmojiInitialPosition
```
PIG208, line 328: nit: need a trailing comma for all parameters when there are multiple lines of them:
```suggestion
      firstEmojiInitialPosition,
```

test/widgets/autocomplete_test.dart
```diff
@@ -250,6 +282,66 @@ void main() {
...
+    ).equals(firstEmojiPositionAfterScrollDown);
+
+    check(
+      because: "The last emoji should not be visible before scrolling up",
+      find.text(emojiNames.last)
```
PIG208, line 333: Similar issue with the trailing comma here.

test/widgets/autocomplete_test.dart
```diff
@@ -250,6 +282,66 @@ void main() {
...
+    ).findsNothing();
+
+    await tester.drag(listViewFinder, const Offset(0, 50));
+    await tester.pump();
+
+    check(because: "The biohazard emoji should be visible after scrolling up",
```
PIG208, line 339: nit: it is more significant that the option is the "last emoji" than that it is the "biohazard emoji"

PIG208:
> I am unable to understand how you see the resolved link https://github.com/zulip/zulip-flutter/pull/1131#pullrequestreview-2551256125:~:text=Co%2Dauthored%2Dby%3A%20Zixuan%20James%20Li%20%3Czixuan%40zulip.com%3E%0AFixes%3A%20https%3A//github.com/zulip/zulip%2Dflutter/issues/1121.

Oh right. That link (`#1121`) was converted into the full link when I drafted that message.

Let's change

```
Co-authored-by: Zixuan James Li <zixuan@zulip.com>
Fixes #1121.
```

to

```
Co-authored-by: Zixuan James Li <zixuan@zulip.com>
Fixes: #1121
```

PIG208 (commented):
All my remaining comments are small. Marking this for Greg's review.

gnprice:
Thanks @apoorvapendse and @PIG208 for all your work on this!

There's still an open question here of whether we want to make this change in the UX. That was Alya's feedback on the thread (in #mobile at the time because #mobile-design didn't exist yet; I've just moved it there) last month:
https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Put.20best.20autocomplete.20matches.20at.20bottom.20of.20list.20.23F1121/near/1999854

Given that state, I think it won't make sense to invest more effort into the implementation before further design discussion.

gnprice:
Given the state described above, and because we're now [focusing more strictly on launch issues](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556), we aren't going to find time to reach a decision on this feature and take this PR forward until after the app's launch is ready.

So I'll close this PR to get it out of our queue. The actual changes will remain in Git — and I'll make a note over on the issue that if we decide we do want this UX change, this is the implementation we'd presumably use. Thanks again for building it.

PR #1132, opened 2024-12-11
Issue#930 Hero animation for images should run only to/from lightbox, not between message lists

sher999:
Inside ` _LightBoxHeroTag` added `narrow` to identify the message list


chrisbobbe:
This needs to follow our Git style before a core team member will review it.

Also, a narrow doesn't uniquely identify a message list. There could be two message lists open showing the same narrow.

PR #1141, opened 2024-12-12
app_bar [nfc]: Centralize _getEffectiveCenterTitle in our wrapper

chrisbobbe:
And make `TestZulipApp` more helpful, too, since I noticed an opportunity for that.

### DIFF
--- a/lib/widgets/app_bar.dart
+++ b/lib/widgets/app_bar.dart
@@ -6,14 +6,66 @@ import 'store.dart';
 ///
 /// This should be used for most of the pages with access to [PerAccountStore].
 class ZulipAppBar extends AppBar {
+  /// Creates our Zulip custom app bar based on [AppBar].
+  ///
+  /// [buildTitle] is passed a boolean `willCenterTitle` that answers
+  /// whether the underlying [AppBar] will decide to center [title]
+  /// based on [centerTitle], the theme, the platform, and [actions].
+  /// Useful if [title] is a container whose children should align the same way,
+  /// such as a [Column] with multiple lines of text.
+  // TODO(upstream) send a PR to replace our `willCenterTitle` code
   ZulipAppBar({
     super.key,
     super.titleSpacing,
-    required super.title,
+    Widget? title,
+    Widget Function(bool willCenterTitle)? buildTitle,
+    super.centerTitle,
     super.backgroundColor,
     super.shape,
     super.actions,
-  }) : super(bottom: _ZulipAppBarBottom(backgroundColor: backgroundColor));
+  }) :
+    assert((title == null) != (buildTitle == null)),
+    super(
+      bottom: _ZulipAppBarBottom(backgroundColor: backgroundColor),
+      title: title ?? _Title(centerTitle: centerTitle, actions: actions, buildTitle: buildTitle!)
+    );
+}
+
+class _Title extends StatelessWidget {
+  const _Title({
+    required this.centerTitle,
+    required this.actions,
+    required this.buildTitle,
+  });
+
+  final bool? centerTitle;
+  final List<Widget>? actions;
+  final Widget Function(bool centerTitle) buildTitle;
+
+  // A copy of [AppBar._getEffectiveCenterTitle].
+  bool _getEffectiveCenterTitle(ThemeData theme) {
+    bool platformCenter() {
+      switch (theme.platform) {
+        case TargetPlatform.android:
+        case TargetPlatform.fuchsia:
+        case TargetPlatform.linux:
+        case TargetPlatform.windows:
+          return false;
+        case TargetPlatform.iOS:
+        case TargetPlatform.macOS:
+          return actions == null || actions!.length < 2;
+      }
+    }
+
+    return centerTitle ?? theme.appBarTheme.centerTitle ?? platformCenter();
+  }
+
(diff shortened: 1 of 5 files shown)


lib/widgets/app_bar.dart
```diff
@@ -6,14 +6,44 @@ import 'store.dart';
...
+  /// whether the underlying [AppBar] will decide to center [title]
+  /// based on [centerTitle], the theme, the platform, and [actions].
+  /// Useful if [title] is a container whose children should align the same way,
+  /// such as a [Column] with multiple lines of text.
+  // TODO send a PR upstream to replace our `willCenterTitle` code
   ZulipAppBar({
     super.key,
     super.titleSpacing,
-    required super.title,
+    required Widget? Function(bool Function(ThemeData) willCenterTitle) buildTitle,
```
PIG208, line 20: It took me a while to parse how this works. Maybe we can try to implement this as a widget? A concept would be like:

```dart
class _TitleBuilder extends StatelessWidget {
  const _TitleBuilder({
    required this.centerTitle,
    required this.actionsCount,
    required this.buildTitle,
  });

  final bool? centerTitle;
  final int actionsCount;
  final Widget Function(bool centerTitle) buildTitle;

  // Copied from [AppBar._getEffectiveCenterTitle].
  bool willCenterTitle(BuildContext context) {
     final theme = Theme.of(context);
     bool platformCenter() {
       switch (theme.platform) {
         case TargetPlatform.android:
         case TargetPlatform.fuchsia:
         case TargetPlatform.linux:
         case TargetPlatform.windows:
           return false;
         case TargetPlatform.iOS:
         case TargetPlatform.macOS:
           return actionsCount == 0 || actionsCount < 2;
       }
     }

     return centerTitle
       ?? theme.appBarTheme.centerTitle
       ?? platformCenter();
  }

  @override
    return buildTitle(willCenterTitle(context));
  }
}
```

Letting `buildTitle` take a `bool` simplifies the API a bit and with a widget we can replace the callback constructed below with a method.

PIG208:
Thanks for working on this! Left a comment on the proposed `buildTitle` API. Otherwise this looks good to me.

lib/widgets/app_bar.dart
chrisbobbe: Oh nice, this is great! Thanks for the idea.

chrisbobbe:
Thanks for the review! Revision pushed. I arranged it so it includes an exact copy of `_getEffectiveCenterTitle`, to make it as easy as possible to verify that it behaves the same.

chrisbobbe:
Note to self; I think there's a comment in the message-list code I've forgotten to remove in this revision:

```dart
      // The helper [_getEffectiveCenterTitle] relies on the fact that we
      // have at most one action here.
```

gnprice (commented):
Thanks to you both!

Some small comments below, and then a further adjustment to the API.

test/widgets/test_app.dart
```diff
@@ -27,8 +44,20 @@ class TestZulipApp extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return GlobalStoreWidget(
-      child: MaterialApp(
+    return GlobalStoreWidget(child: Builder(builder: (context) {
+      if (accountId != null && skipAssertAccountExists != true) {
+        final account = GlobalStoreWidget.of(context).getAccount(accountId!);
+        assert(account != null,
+          'TestZulipApp() was called with [accountId] but a corresponding '
```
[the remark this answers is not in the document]
chrisbobbe: I've put the conditionals inside an IIFE, and actually moved the `GlobalStoreWidget.of` call so it's outside the whole `assert`, since it has a side effect (of creating a dependency on the global store).

chrisbobbe:

test/widgets/app_bar_test.dart
```diff
@@ -35,4 +37,124 @@ void main() {
...
+
+      // test assumes LTR text direction
+      check(tester.platformDispatcher.locale).equals(const Locale('en', 'US'));
+      assert(actions == null || actions.isNotEmpty);
+      final titleAreaRightEdgeOffset = actions == null
+        ? (tester.view.physicalSize / tester.view.devicePixelRatio).width
+        : tester.getTopLeft(find.byWidget(actions.first)).dx;
+      final titlePosition = tester.getCenter(find.text('a')).dx;
+      final isCentered = titlePosition > ((1 / 3) * titleAreaRightEdgeOffset);
+      check(titlePosition).isLessThan((2 / 3) * titleAreaRightEdgeOffset);
```
gnprice, line 69: How about:
```suggestion
      final titlePosition = tester.getTopLeft(find.text('a')).dx;
      final isCentered = titlePosition > ((1 / 3) * titleAreaRightEdgeOffset);
      check(titlePosition).isLessThan((2 / 3) * titleAreaRightEdgeOffset);
```

Otherwise this test implicitly relies on the title having its size shrink to its contents, rather than filling its parent's size.

gnprice (commented):
Thanks! This looks good now; and I've also now read the tests, and I appreciate the thorough set of test cases. One small comment in the tests.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

PR #1143, opened 2024-12-12
notif: package info indirection

### LINKED ISSUE #407: notif: Use live value for app ID on registering APNs token
In my draft for #321 I have a TODO like this:
```dart
      case TargetPlatform.iOS:
        const appBundleId = 'com.zulip.flutter'; // TODO find actual value live
        await registerApnsToken(connection, token: token, appid: appBundleId);
```

This will be just fine as long as the app ID is indeed `com.zulip.flutter`. That means it will basically be fine right through the beta period, until we start rolling this app out as the new version of `org.zulip.Zulip` (the app ID currently occupied by the zulip-mobile RN app).

The fix is easy, locally:
```dart
        final appBundleId = (await PackageInfo.fromPlatform()).packageName;
```

The one wrinkle is that that version breaks tests, because it's directly calling a plugin. To fix it, we'll want to indirect `PackageInfo.fromPlatform` through a method on `ZulipBinding`, like we've done for other plugins. This still isn't hard, but given the impending beta, I'm deferring it for now.

Abhisheksainii:
In this PR, I made the changes to indirect PackageInfo.fromPlatform through a method on ZulipBinding.

Steps I followed :

1. Update the ZulipBinding to provide package info: Add a method in ZulipBinding to return the package name. This method will wrap the call to PackageInfo.fromPlatform() and make it test-friendly.

2. Override the method in TestZulipBinding: In the TestZulipBinding, override getAppBundleId to return a mock or predefined value for testing.

3. Refactor the code to use getAppBundleId: Replace the direct calls with the indirection from Zulip Binding class.

4. Write a test for the above logic under a group called 'tokens'. The allow the grouping of all tests related to tokens, I created a separate group : tokens.

Fixes #407

### DIFF
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -303,10 +303,12 @@ class LinuxDeviceInfo implements BaseDeviceInfo {
 class PackageInfo {
   final String version;
   final String buildNumber;
+  final String packageName;
 
   const PackageInfo({
     required this.version,
     required this.buildNumber,
+    required this.packageName,
   });
 }
 
@@ -411,6 +413,7 @@ class LiveZulipBinding extends ZulipBinding {
       _syncPackageInfo = PackageInfo(
         version: info.version,
         buildNumber: info.buildNumber,
+        packageName: info.packageName,
       );
     } catch (e, st) {
       assert(debugLog('Failed to prefetch package info: $e\n$st')); // TODO(log)
--- a/lib/notifications/receive.dart
+++ b/lib/notifications/receive.dart
@@ -149,8 +149,7 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
-        await addApnsToken(connection, token: token, appid: appBundleId);
+        await addApnsToken(connection, token: token, appid: (await ZulipBinding.instance.packageInfo)?.packageName?? 'com.zulip.flutter');
 
       case TargetPlatform.linux:
       case TargetPlatform.macOS:
--- a/test/model/binding.dart
+++ b/test/model/binding.dart
@@ -228,8 +228,8 @@ class TestZulipBinding extends ZulipBinding {
   BaseDeviceInfo? get syncDeviceInfo => deviceInfoResult;
 
   /// The value that `ZulipBinding.instance.packageInfo` should return.
-  PackageInfo packageInfoResult = _defaultPackageInfo;
-  static const _defaultPackageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+  PackageInfo? packageInfoResult = _defaultPackageInfo;
+  static const _defaultPackageInfo = PackageInfo(version: '0.0.1', buildNumber: '1', packageName: 'com.zulip.flutter.test');
 
   void _resetPackageInfo() {
     packageInfoResult = _defaultPackageInfo;
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -1032,7 +1032,7 @@ void main() {
       if (defaultTargetPlatform == TargetPlatform.android) {
         checkLastRequestFcm(token: '012abc');
       } else {
-        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
+        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter.test');
       }
 
       if (defaultTargetPlatform == TargetPlatform.android) {
@@ -1071,7 +1071,7 @@ void main() {
       if (defaultTargetPlatform == TargetPlatform.android) {
         checkLastRequestFcm(token: '012abc');
       } else {
-        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
+        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter.test');
       }
 
       if (defaultTargetPlatform == TargetPlatform.android) {
@@ -1082,6 +1082,22 @@ void main() {
         checkLastRequestFcm(token: '456def');
       }
     }));
+
+
+    test('fallback to default appBundleId in case packageInfo is null', ()async{
+      // This tests the case where packageInfo is null
+      // and default appBundleId gets passed to the [NotificationService.registerToken] method
+
+      prepareStore();
+      connection.prepare(json: {});
+
+      if(defaultTargetPlatform == TargetPlatform.iOS){
+        testBinding.packageInfoResult=null;
+        await NotificationService.registerToken(connection, token: '012abc');
+        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
+        addTearDown(testBinding.reset);
+      }
+    });
   });
 }
 
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -460,7 +460,7 @@ void main() {
       });
     }
 
-    const packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+    const packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1', packageName: 'com.zulip.flutter.test');
 
     const testCases = [
       ('ZulipFlutter/0.0.1+1 (Android 14)',             AndroidDeviceInfo(release: '14', sdkInt: 34),                      ),


chrisbobbe:
Hi @Abhisheksainii, thanks for the PR. It will need to follow our Git style before it gets reviewed by a core team member.

Abhisheksainii:
@chrisbobbe As per what I learned from the guide, I tried explaining the changes in my PR and also made the commits in such a way that explains what I did and why I did those changes. If there is anything else, for example, drafting this PR, please guide me.
Thanks

lib/model/binding.dart
```diff
@@ -197,6 +197,10 @@ abstract class ZulipBinding {
   ///   https://developer.apple.com/documentation/uikit/uiapplication/1623070-idletimerdisabled
   ///   https://github.com/fluttercommunity/wakelock_plus/blob/5ca5243e7894830ce289fc367bc5fdec27c7f0cf/wakelock_plus/ios/Classes/WakelockPlusPlugin.m
   Future<void> toggleWakelock({required bool enable});
+
+
```
PIG208, line 201: nit: remove the extra newline

lib/model/binding.dart
```diff
@@ -197,6 +197,10 @@ abstract class ZulipBinding {
...
+  // Method for wrapping the call to PackageInfo
+  Future<String> getAppBundleId();
```
PIG208, line 203: We already have access to `PackageInfo` through `ZulipBinding.packageInfo`, so it looks like we can just use that instead?

lib/notifications/receive.dart
```diff
@@ -149,7 +149,7 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
+        final appBundleId = await ZulipBinding.instance.getAppBundleId(); // TODO(#407) find actual value live
```
PIG208, line 152: Then to access the value, we can do something like `await ZulipBinding.instance.packageInfo.packageName`

PIG208 (commented):
Thanks for the PR @Abhisheksainii! Looks like there is a simplification to be made, because we already access and store `packageInfo`. It will probably require you to modify our copy of `PackageInfo` class by adding `packageName` to it.

lib/notifications/receive.dart
Abhisheksainii: I was thinking what if that pacakgeInfo variable is null?
What we can do instead is this ->

Future<String> getAppBundleId() async{
    if(syncPackageInfo !=null ){
      return syncPackageInfo!.packageName;
    }

    final packageInfo = await package_info_plus.PackageInfo.fromPlatform();
    return packageInfo.packageName;
  }

PIG208: > I was thinking what if that pacakgeInfo variable is null?

Yeah, good thought. I think we handled the same issue for `AboutZulipPage`. 

Checking the dartdoc, `packageInfo` can resolve to `null` if there was an error, but otherwise awaiting it should give us the needed information.

```dart
  /// Provides application package information,
  /// via package:package_info_plus.
  ///
  /// The returned Future resolves to null if an error is
  /// encountered while fetching the data.
  ///
  /// This wraps [package_info_plus.PackageInfo.fromPlatform].
  Future<PackageInfo?> get packageInfo;
```

Perhaps something like:
```suggestion
        final appBundleId = (await ZulipBinding.instance.packageInfo)?.packageName ?? 'some fallback value';
```

I'm not entirely sure what fallback value or error handling is appropriate here. Feel free to start a discussion in the [#mobile-dev-help channel](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help) if you need help navigate this.

Abhisheksainii: Okay, this approach looks simple and subtle. I will adding the fallback value as 'com.zulip.flutter' for the binding class and 'com.zulip.flutter.test' for the test binding class.

Abhisheksainii:
@PIG208 changes done as requested, kindly take a look at them.

PIG208:
Thanks for the update. Please review the "commit style guide" referenced in the README and update your commits: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request

Also, it looks like the CI is failing. Could you fix that?

Abhisheksainii:
> Thanks for the update. Please review the "commit style guide" referenced in the README and update your commits: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request
> 
> ~Also, it looks like the CI is failing. Could you fix that?~ Looks like the CI failure [is not relevant](https://github.com/zulip/zulip-flutter/issues/1177) to this PR.

@PIG208 I checked the commits and they pretty much explain what they do. Am I supposed to add some more description?

gnprice:
Please read the Zulip commit style guide, which is linked from the README where Zixuan pointed to.

Abhisheksainii:
> Thanks for the update. Please review the "commit style guide" referenced in the README and update your commits: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request
> 
> ~Also, it looks like the CI is failing. Could you fix that?~ Looks like the CI failure [is not relevant](https://github.com/zulip/zulip-flutter/issues/1177) to this PR.

@PIG208 commits updated according to the style guide.

Abhisheksainii:
@PIG208 kindly review the changes.

test/model/binding.dart
```diff
@@ -380,6 +380,7 @@ class TestZulipBinding extends ZulipBinding {
   Future<void> toggleWakelock({required bool enable}) async {
     _wakelockEnabled = enable;
   }
+
```
PIG208, line 383: Looks like this is unintentional.

lib/model/binding.dart
```diff
@@ -303,10 +303,12 @@ class LinuxDeviceInfo implements BaseDeviceInfo {
 class PackageInfo {
```
PIG208, line 303:
```
binding: add packageName to the packageInfo class in the binding class 
made changes to access packageName from the PackageInfo class.
remove getAppBundleId() from Zulip Binding class
Add packageName variable to PackageInfo class
Fixes #407
```

The commit claims that `getAppBundleId()` got removed, but it is not present in the commit itself. For changes that address reviews on a previous revision of the commit, the PR thread is more appropriate.

The commit summary and the "Fixes" line also need to be updated. This commit doesn't fix #407. We should move this line to the next commit where we start using `packageName`.

For the commit message, changes that one can tell from reading the diffs generally don't need to be reiterated: "made changes to access packageName from the PackageInfo class"
"Add packageName variable to PackageInfo class".

Please review Zulip's commit discipline linked in the README https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request

test/notifications/receive_test.dart
```diff
@@ -30,4 +30,12 @@ void main() {
...
     }, variant: const TargetPlatformVariant({TargetPlatform.android, TargetPlatform.iOS}));
   });
+
+  group('tokens', () {
+    test('APNs token registration using correct app bundle ID', () async {
+      await init();
+      check(await testBinding.getAppBundleId())
+        .equals('com.zulip.flutter.test');
+    }, );
+  });
```
PIG208, line 40: Hmm, this commit should be dropped as `getAppBundleId` doesn't exist anymore.

This is one reason that we should include the tests in the same commit that implements the feature tested.

lib/notifications/receive.dart
```diff
@@ -149,8 +149,8 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
-        await addApnsToken(connection, token: token, appid: appBundleId);
+         // TODO(#407) find actual value live
```
PIG208, line 152: Since this resolves the TODO, we can remove this line.

test/model/store_test.dart
```diff
@@ -1032,7 +1032,7 @@ void main() {
       if (defaultTargetPlatform == TargetPlatform.android) {
         checkLastRequestFcm(token: '012abc');
       } else {
-        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
+        checkLastRequestApns(token: '012abc', appid:(await testBinding.packageInfo)?.packageName?? 'com.zulip.flutter.test');
```
PIG208, line 1035: The test should encode the clear expectation of what the expected appid is. Let's avoid looking up the value and use a literal string instead.

test/model/store_test.dart
```diff
@@ -1071,7 +1071,7 @@ void main() {
...
```
PIG208, line 1074: Same as the other case.

lib/notifications/receive.dart
```diff
@@ -149,8 +149,8 @@ class NotificationService {
...
+        await addApnsToken(connection, token: token, appid: (await ZulipBinding.instance.packageInfo)?.packageName ?? "com.zulip.flutter");
```
PIG208, line 153: I think there is an underlying issue if `packageInfo` is missing, so let's include a `// TODO(log)` comment here.

test/notifications/receive_test.dart
```diff
@@ -34,7 +34,7 @@ void main() {
   group('tokens', () {
     test('APNs token registration using correct app bundle ID', () async {
       await init();
-      check(await testBinding.getAppBundleId())
+      check((await testBinding.packageInfo)?.packageName?? 'com.zulip.flutter.test')
```
PIG208, line 37: Falling back to `com.zulip.flutter.test` defeats the purpose of this test. I think we don't expect `packageInfo` to be missing in this case, right?

lib/notifications/receive.dart
PIG208: This case is interesting so it might worth a test for it. Let's find a way to test when we fallback to this default value, if `packageInfo` is somehow missing.

PIG208:
Thanks for the update @Abhisheksainii! Left some comments.

test/notifications/receive_test.dart
Abhisheksainii: You are correct, falling back to the literal defeats the purpose. Will update.

lib/notifications/receive.dart
Abhisheksainii: So, should we test for packageInfo being null just incase of packageName?, because packageInfo also have other members like build number and version.

PIG208: `packageInfo` itself is nullable, so it makes sense to test that.

Abhisheksainii: @PIG208 I thought of simulating the packageInfo being null when fetching from TestBinding class. For instance, creating a function in binding class that returns null for PackageInfo and then taking its value and checking against the fallback value (i.e. "com.zulip.flutter")

Abhisheksainii:
@PIG208 made changes as requested. Kindly review.

PIG208 (commented):
Thanks for the update! Left some comments. I think we should squash all commits into one, so that the implementation and the tests packed together for [coherency](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#each-commit-must-be-coherent).

Abhisheksainii:
@PIG208 made changes as requested. Kindly review.

PIG208 (commented):
Thanks for the update!

As I [mentioned](https://github.com/zulip/zulip-flutter/pull/1143#pullrequestreview-2539208206) earlier, let's squash all commits into one.

I see that there are some comments marked as resolved that didn't actually get addressed in this update. Please be sure to double-check your changes before pushing.

test/notifications/receive_test.dart
```diff
@@ -34,8 +34,19 @@ void main() {
   group('tokens', () {
     test('APNs token registration using correct app bundle ID', () async {
```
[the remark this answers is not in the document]
Abhisheksainii: I see, then why not add test in notification_test.dart?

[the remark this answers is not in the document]
Abhisheksainii: @PIG208 would like to know your opinion on this :
 test('Fallback to default appBundleId incase packageInfo is null', () async {
     await init();
      // setting package Info to null
      testBinding.packageInfoResult=null;

      const defaultAppId = 'com.zulip.flutter.test';

      return FakeApiConnection.with_((connection) async {
        await checkAddApnsToken(connection, token: 'asdf', appid:(await testBinding.packageInfo)?.packageName?? defaultAppId);
      });
    }, );

Abhisheksainii:
@PIG208 changes made as requested. Kindly review

test/notifications/receive_test.dart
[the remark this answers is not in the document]
PIG208: The reason we want to test it from `test/model/store_test.dart` is that we want to have a test set up in a way to verify that the call-site (the change you made in `lib/notifications/receive.dart`) of `addApnsToken` handles the case when `packageInfo` is null correctly.

Using the `checkAddApnsToken` helper calls `addApnsToken` directly, which introduces a code-site on its own. That doesn't make `addApnsToken` called through `lib/notifications/receive.dart`. This test also uses a fallback value, which is not ideal when we are supposed to know what to expect.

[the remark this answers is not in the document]
Abhisheksainii: so if we call checkLastRequestApns in store_test file, wouldn't it introduce a code-site on its own as well?

Abhisheksainii:
@PIG208 changes done as requested. Please review them

lib/notifications/receive.dart
```diff
@@ -143,14 +143,13 @@ class NotificationService {
     token.value = value;
   }
 
-  static Future<void> registerToken(ApiConnection connection, {required String token}) async {
-    switch (defaultTargetPlatform) {
+  static Future<void> registerToken(ApiConnection connection,{required String token,  TargetPlatform? targetPlatform, }) async {
+    switch (targetPlatform ?? defaultTargetPlatform) {
```
PIG208, line 147: I think we shouldn't need to modify this method for the sake of testing it.

test/model/store_test.dart
```diff
@@ -1082,6 +1082,16 @@ void main() {
...
       }
     }));
+
+     test('Fallback to default appBundleId incase packageInfo is null', () async {
+
+        prepareStore();
+        connection.prepare(json: {});
+
+        await NotificationService.registerToken(connection, token: '012abc', targetPlatform: TargetPlatform.iOS);
+        checkLastRequestApns(token: '012abc', appid:'com.zulip.flutter.test');
```
PIG208, line 1092: Thanks! It's helpful to have a test like this because it exercises the `registerToken` code path that we also changed in this commit.

We generally use 2 spaces for indentation, but there are 3 spaces here.

`'Fallback to default appBundleId incase packageInfo is null'` -> `'fallback to default appBundleId in case packageInfo is null'`

For this test to work, we need to first ensure that `packageInfo` is actually `null`, before we call `reigsterToken`. We can do that with `testBinding.packageInfoResult = null;` Because this modifies `testBinding`, which is shared across tests, we also need to reset it when this test end to avoid interfering with others: `addTearDown(testBinding.reset);`

One other issue is the set up of the test, as mentioned in the other comment. See the other tests in the file for reference. In this case, we only need to test the situation when the target platform is iOS. The other tests use the `testAndroidIos` helper; see if you can borrow only the relevant part from it to set up your test.

test/model/store_test.dart
```diff
@@ -1071,7 +1071,7 @@ void main() {
       if (defaultTargetPlatform == TargetPlatform.android) {
         checkLastRequestFcm(token: '012abc');
       } else {
-        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
+        checkLastRequestApns(token: '012abc', appid:'com.zulip.flutter.test');
```
PIG208, line 1074: nit: space before the string
```suggestion
        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter.test');
```

test/model/store_test.dart
```diff
@@ -1032,7 +1032,7 @@ void main() {
...
```
PIG208, line 1035: similar issue here

gnprice:
The commit in this PR is still quite messy and hard to read, even after four rounds of review (plus a fifth posted just now above). @Abhisheksainii in order for us to continue spending time reviewing your work, you need to start doing a better job of self-reviewing it first. See our guidelines:
https://zulip.readthedocs.io/en/latest/contributing/reviewable-prs.html

For your next revision, I recommend you rewrite the change from scratch. (There's not very much code here anyway, so that isn't much work.) You can look back at the current revision for reference (and you should definitely pay attention to @PIG208's latest round of comments), but I recommend not copy-pasting anything and instead typing it out. The important thing is to think through each change you're making based on what it's accomplishing and what's needed.

Then when you're writing text like a test description or a commit message, make sure it accurately describes what the code is actually doing. In this revision they don't.

When you've finished writing a new version, re-read the whole thing carefully and edit as needed to make everything as clear and accurate as you can, and to make sure you've addressed all review comments. For detailed guidance, see our docs that I've linked above.

test/model/store_test.dart
Abhisheksainii: Really grateful for the explanation. 
Apart from adding addTearDown(testBinding.reset) after the test ends, we can also add it before testBinding.packageInfoResult = null, to ensure we are using a fresh instance of testBinding. 

I have two thoughts here:

1. Even if we don't use addTearDown(testBinding.reset) before setting packageInfoResult to null, it would hardly make any difference, as we will be setting packageInfoResult to null only. It would matter incase if we would be using multiple fields from testBinding.

2. It is helpful to add addTearDown(testBinding.reset) before setting packageInfoResult to null only if we are concerned about the addition of extra steps or enhancements in this test in future.

@PIG208 Please let me know your thoughts if I am thinking correctly or not.

Abhisheksainii: Also, registerToken function is using Zulip Binding, How can we make sure that it use TestZulipBinding? So that testBinding.packageInfoResult = null have some relevance.

  static Future<void> registerToken(ApiConnection connection,{required String token}) async {
    switch (defaultTargetPlatform) {
      case TargetPlatform.android:
        await addFcmToken(connection, token: token);

      case TargetPlatform.iOS:
        await addApnsToken(connection, token: token, appid: (await ZulipBinding.instance.packageInfo)?.packageName ?? "com.zulip.flutter");

      case TargetPlatform.linux:
      case TargetPlatform.macOS:
      case TargetPlatform.windows:
      case TargetPlatform.fuchsia:
        assert(false);
    }
  }

Abhisheksainii: > 
> I'm not sure if I understood the question. `packageInfoResult` has a non-null default value. Setting it to `null` changes its value to something we might not expect from other tests.

My question was should we also add addTearDown(testBinding.reset)) method before setting packageInfoResult to null?

Abhisheksainii: > That's a good question. We have two implementations for the abstract `ZulipBinding` class, `LiveZulipBinding` and `TestZulipBinding`. `ZulipBinding` is abstract, and in runtime, one of the implementations is used. You can find the answer by reading the dartdoc of `TestZulipBinding`. As with many other things in this codebase, we have extensive documentation explaining how something works, so it's always worth a read if you find something confusing.

I got my answer, whichever of the two gets Initialized first, is used as an instance of ZulipBinding. Thanks!

Abhisheksainii:
@PIG208 made changes from scratch. Kindly review

test/model/store_test.dart
```diff
@@ -1082,6 +1082,22 @@ void main() {
...
+
+
+    test('fallback to default appBundleId in case packageInfo is null', ()async{
+      // This tests the case where packageInfo is null
+      // and default appBundleId gets passed to the [NotificationService.registerToken] method
+
+      prepareStore();
+      connection.prepare(json: {});
+
+      if(defaultTargetPlatform == TargetPlatform.iOS){
```
PIG208, line 1094: This check doesn't seem right. Because the `debugDefaultTargetPlatformOverride` doesn't get updated, the block inside never gets executed. The other tests use this because they use `testAndroidIos` instead of `test`.

PIG208: nit:

```suggestion
      if (defaultTargetPlatform == TargetPlatform.iOS) {
```

Please also check for other spots for issues like this.

test/model/store_test.dart
```diff
@@ -1082,6 +1082,22 @@ void main() {
         checkLastRequestFcm(token: '456def');
       }
     }));
+
+
+    test('fallback to default appBundleId in case packageInfo is null', ()async{
+      // This tests the case where packageInfo is null
+      // and default appBundleId gets passed to the [NotificationService.registerToken] method
```
PIG208, line 1089: Ideally the comment shouldn't be needed here, as the test code itself would be self-explanatory in most cases.

It would be better to have comments on the exact places where something might not be obvious to a reader.

test/model/store_test.dart
```diff
@@ -1082,6 +1082,22 @@ void main() {
...
+
+      prepareStore();
+      connection.prepare(json: {});
+
+      if(defaultTargetPlatform == TargetPlatform.iOS){
+        testBinding.packageInfoResult=null;
+        await NotificationService.registerToken(connection, token: '012abc');
+        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
+        addTearDown(testBinding.reset);
```
PIG208, line 1098: Usually this is called at the beginning of the test. `addTearDown` registers a callback function that gets called when the test is finished and "teared down", so there is no need to place it near the end.

lib/notifications/receive.dart
```diff
@@ -149,8 +149,7 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
-        await addApnsToken(connection, token: token, appid: appBundleId);
+        await addApnsToken(connection, token: token, appid: (await ZulipBinding.instance.packageInfo)?.packageName?? 'com.zulip.flutter');
```
PIG208, line 152: Bumping https://github.com/zulip/zulip-flutter/pull/1143#discussion_r1896163485

Let's include a `assert(debugLog(…))` here with a `TODO(log)` comment.

It would be helpful to structure this with an if-statement first.
```suggestion
        final packageInfo = await ZulipBinding.instance.packageInfo;
        if (packageInfo == null) {
          assert(debugLog('missing packageInfo')); // TODO(log)
        }
        await addApnsToken(connection,
          token: token, appid: packageInfo?.packageName ?? 'com.zulip.flutter');
```

PIG208 (commented):
Thanks for the revision! Left a review. Since it doesn't seem ready yet, we won't prioritize this PR for some time until we move on to the next milestone.

test/model/store_test.dart
Abhisheksainii: I found these instructions in testAndroidIos function :
final origTargetPlatform = debugDefaultTargetPlatformOverride;
    addTearDown(() => debugDefaultTargetPlatformOverride = origTargetPlatform);
    debugDefaultTargetPlatformOverride = TargetPlatform.iOS;
  
  what do you think of replicating them in the above test. It will look something like this:
  
  prepareStore();
      connection.prepare(json: {});

      final origTargetPlatform = debugDefaultTargetPlatformOverride;
      addTearDown(() {
        testBinding.reset();
        debugDefaultTargetPlatformOverride = origTargetPlatform;
      });

      debugDefaultTargetPlatformOverride = TargetPlatform.iOS;

      if (debugDefaultTargetPlatformOverride == TargetPlatform.iOS) {
        testBinding.packageInfoResult = null;
        await NotificationService.registerToken(connection, token: '012abc');
        checkLastRequestApns(token: '012abc', appid: 'com.zulip.flutter');
      }

lib/notifications/receive.dart
Abhisheksainii: makes sense! I was supposed to think like that at first😅

Abhisheksainii: can we write this(assert(packageInfo == null, debugLog('missing packageInfo')) instead of an if statement ?

gnprice:
Thanks @Abhisheksainii for working on this. Because it looks like work here isn't continuing, I'm closing this PR to open up the issue for someone else to give it a try.

Issue #43, opened 2023-03-31
lightbox: Add "share" button in bottom app bar


chrisbobbe:
The `share_plus` plugin has an API for sharing a file: https://pub.dev/documentation/share_plus/latest/share_plus/Share/shareXFiles.html

We should try it out, for a button to share the image.

As with #42, I didn't include this in the prototype because I think it'll involve an API request; see there for details.

gyoussef55:
I would like to work on it

chrisbobbe:
We got a [request](https://chat.zulip.org/#narrow/channel/48-mobile/topic/sharing.20content.20from.20Zulip/near/2009779) for this in beta feedback, emphasizing that the legacy app's "share" button in the lightbox is pretty unhelpful because it shares a string of text instead of the image file itself.

From the issue description, it seems like there's a straightforward way to share the file itself; we should do that.

rohitarer:
hello, i would like to work on this

chrisbobbe:
Please take a look at the Zulip project's guide to getting involved with the code:

https://zulip.readthedocs.io/en/latest/contributing/contributing.html#your-first-codebase-contribution
https://zulip.readthedocs.io/en/latest/contributing/contributing.html#picking-an-issue-to-work-on

gnprice:
Requested the other day in chat: [#mobile > Sharing an image from mobile app as an actual image @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Sharing.20an.20image.20from.20mobile.20app.20as.20an.20actual.20image/near/2316590)

PR #1144, opened 2024-12-12
Add download button to lightbox to bottom app bar

### LINKED ISSUE #42: lightbox: Add "download" button in bottom app bar
The bottom app bar currently has just one button, to copy a link to the image. It should also have a button to download the image, in the way that feels most familiar for the platform.

I didn't include this in the prototype because I think it'll involve a network request, like the one we do in the RN app's `api.tryGetFileTemporaryUrl`. As usual for network requests, that means we should consider what to do when the request succeeds, fails, or goes on forever without succeeding or failing.

chimnayajith:
# Pull Request

## Description

This merge request introduces a new **Download Image** button to the bottom app bar, enabling users to download images. Upon tapping the button, the image is downloaded via a network request and saved to the device's storage. Additionally, the media scanner is triggered to ensure the image appears in the gallery. The functionality includes error handling and displays a SnackBar notification to inform the user of success or failure.

A `MethodChannel` (`"gallery_saver"`) has been added to invoke the media scanner after the download is completed.

## Related Issues

- Closes #42 

## Screenshots 

| Before | After  |
| ------ | ------ |
|   before     |  after      |

## Additional Notes

- The network request for downloading the image handles success, failure, and timeouts. 
- A test has been implemented to verify the download button functionality, ensuring the correct behavior is triggered and the SnackBar is displayed appropriately.


lib/widgets/lightbox.dart
```diff
@@ -89,6 +93,157 @@ class _CopyLinkButton extends StatelessWidget {
   }
 }
 
+// class _DownloadImageButton extends StatelessWidget {
```
chrisbobbe, line 96: We don't merge large, unexplained blocks of commented-out code; please remove all this.

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
```diff
@@ -1,6 +1,39 @@
...
+
+    private fun scanFile(filePath: String) {
+        MediaScannerConnection.scanFile(
+            applicationContext,
+            arrayOf(filePath),
+            null
+        ) { _, _ ->
+            // Intentionally left empty
+        }
+    }
```
chrisbobbe, line 38: Why is all this new code needed? The answer should be made clear to someone reading the code.

lib/widgets/lightbox.dart
```diff
@@ -89,6 +93,157 @@ class _CopyLinkButton extends StatelessWidget {
...
+    return IconButton(
+      tooltip: zulipLocalizations.lightboxDownloadImageTooltip,
+      icon: const Icon(Icons.download),
+      onPressed: () async {
+        final scaffoldMessenger = ScaffoldMessenger.of(context);
+        String message = zulipLocalizations.lightboxDownloadImageFailed;
+
+        try {
+          // Fetch the image with a timeout
+          final response = await http.get(url).timeout(
```
chrisbobbe, line 186: Will this work if the image is protected by authentication to the Zulip org? What does zulip-mobile do?

pubspec.yaml
```diff
@@ -64,6 +64,7 @@ dependencies:
   wakelock_plus: ^1.2.8
   zulip_plugin:
     path: ./packages/zulip_plugin
+  permission_handler: ^11.3.1
   # Keep list sorted when adding dependencies; it helps prevent merge conflicts.
```
chrisbobbe, line 68: Please follow the request in the comment. Also, we generally use a separate commit to add dependencies; see the [project history](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/git.md#2-the-secret-to-using-git-log--p) for examples.

chrisbobbe (changes requested):
Comments below from a quick skim. Please also update your commit message to follow our Git style. The part after `lightbox: ` should start with a capital; also add an empty line and then `Fixes: ` with the issue fixed in the commit.

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
chimnayajith: I've added necessary comments for this file,

lib/widgets/lightbox.dart
chimnayajith: That was a mistake, have removed it now,

pubspec.yaml
chimnayajith: changed the order maintaining the list sorted

chimnayajith:
@chrisbobbe I've made the changes as mentioned, Please take a look when you have time, Thank you.

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
```diff
@@ -1,6 +1,57 @@
...
+
+            // Handle the method call when the "scanFile" method is invoked from Flutter
+            if (call.method == "scanFile") {
+                // Get the file path passed from Flutter
+                val filePath = call.argument<String>("path")
+
+                // Check if the file path is not null
+                if (filePath != null) {
+                    // If the file path is valid, trigger the scanning of the file
+                    scanFile(filePath)
```
gnprice, line 30: These comments aren't helpful — they're just repeating the "what" of what the code is doing.

The code itself says the same thing just as clearly, on the very next line after the comment. And reading the code is a reliable way to see what the code is doing, but reading a comment is unreliable because comments can get out of date when the code changes. So a comment like these that doesn't add any useful information is much worse than not having a comment.

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
```diff
@@ -1,6 +1,57 @@
...
+                result.notImplemented()
+            }
+        }
+    }
+
+    // This function triggers the Android MediaScanner to refresh the media library
+    // so that the file is visible in the gallery or other media apps.
+    private fun scanFile(filePath: String) {
+        // Use the MediaScannerConnection to scan the file and add it to the device's media library
+        MediaScannerConnection.scanFile(
```
gnprice, line 48: These are the key comments that do have useful information:
> to refresh the media library so that the file is visible in the gallery or other media apps.

> to scan the file and add it to the device's media library

They contain "why" information about the code, rather than just "what". The "why" information overlaps with "what" information about what some external API is doing, which is also relevant information. That's because someone reading our codebase (e.g., a reviewer of this PR) may well not know what that external API does, whereas they certainly know language features like that `foo != null` checks if foo is not null.

The key information that should go in a comment in this code is
* the information that's in these two comments
* but said just once, not twice
* plus I'd like to know more about both the "why", and the external "what". Let's start with the latter: what's a good link for authoritative documentation about this piece of API? And then if that page doesn't already answer this next question: what's an explanation of why this is the step we need to be taking as part of downloading a file on Android?

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
gnprice: Most of these comments look like what you might get if you asked ChatGPT to add comments everywhere in this file. If you are using some LLM, you need to guide it more to get the output focused on what's relevant.

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
```diff
@@ -1,6 +1,57 @@
...
+
+    // Define a channel name for communication between Flutter and native Android
+    private val CHANNEL = "gallery_saver"
+
+    // Override the configureFlutterEngine method to set up the MethodChannel and handle method calls
+    override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
+        super.configureFlutterEngine(flutterEngine)
+
+        // Set up a method channel for communication between Flutter and native Android code
+        MethodChannel(flutterEngine.dartExecutor.binaryMessenger, CHANNEL).setMethodCallHandler { call, result ->
```
gnprice, line 20: The other high-level feedback I have on this part of the code is that in our app we don't do manual method channels like this. Instead, we use Pigeon to automate the boilerplate and get reliable end-to-end type safety.

For examples, see the other files in this directory, and see the `pigeon/` directory.

android/app/src/main/kotlin/com/zulip/flutter/MainActivity.kt
chimnayajith: will make the commit according to the standards. thank you for the detailed response!

gnprice:
See also the chat thread I started for sorting out the available APIs here:
https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Android.20APIs.20for.20downloading.20an.20image/near/2010826

I'd like to hear there much the same information I asked for at https://github.com/zulip/zulip-flutter/pull/1144#discussion_r1893430841 above — that'd help us compare this API to alternatives and work out what solution we want.

chimnayajith:
@gnprice I've worked on the reviews you've given, 

- Implemented the same with pigeon ( still not sure if its the right way ). 
- Reduced the comments to just the needed ones.

gnprice:
Thanks!

On a high-level scan, this structure looks good. The use of a new file `pigeon/media_scanner.dart` and the `MediaScannerHost` class look like the right structure for implementing a call to the Android side.

The main thing this will need is tests. In this version, the test case checks that a snack bar gets shown, but not any part of the actual functionality of the button. There's significant logic in that `onPressed` callback, and we'll want tests for that logic.

To do that, the main ingredient you'll need is to call the host `scanFile` method via a method you'll add to `ZulipBinding`, rather than calling it directly. See the advice in this section of our README:
https://github.com/zulip/zulip-flutter/?tab=readme-ov-file#writing-tests

For examples using Pigeon, see our notifications tests. For a simpler example though not using Pigeon, try looking at ZulipBinding.pickFiles, and the place where that's called, and the test in test/widgets/compose_box_test.dart that tests the place where pickFiles is called.

You might also need to add a ZulipBinding method for getExternalStorageDirectory, which should be a fairly simple binding method to add.

chimnayajith:
Modified the download implementation to use the `DownloadManager` API instead of the `File` API and `MediaScannerConnection.scanFile()` from the last review. Also wrote tests to verify the functionality of the download using TestZulipBinding.

@gnprice , could you please review it?

gnprice:
Thanks @chimnayajith for all your work on this!

I like this idea of using DownloadManager — I think that's likely to be the right strategy for us to use. The structure of this code looks broadly good, though I haven't reviewed it in detail.

This issue is for the M6 Post-launch milestone, though, and I've determined we need to start [focusing more strictly on launch issues](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556) in order to get the app launched to our users on schedule. And there's still significant work ahead to get this completed, partly because there's always an extra level of care and validation required when dealing with platform APIs.

Given that, let's put this PR on hold for now. We can plan for you to pick this issue up from here when we're ready to resume work on post-launch issues.

chrisbobbe:
Hi @chimnayajith, would you be interested in picking this up again?

chimnayajith:
Yep, I'm interested to continue this.

chrisbobbe:
Great, thanks! When it's ready for review again, please unmark it as a draft.

From a quick skim, there's one thing I don't see yet in this revision: code corresponding to `api.tryGetFileTemporaryUrl` in zulip-mobile; please add that. :) Actually, would you send the "temporary file URL" work as a separate PR? It'll be needed for #43 too.

chimnayajith:
@chrisbobbe Sorry for the delay. Created a separate PR for the `tryGetFileTemporaryUrl` function. #1780

chrisbobbe:
> From a quick skim, there's one thing I don't see yet in this revision: code corresponding to api.tryGetFileTemporaryUrl in zulip-mobile; please add that.

OK, and now, with #1909 merged, we have that! If you'd like to pick this back up again, you can use `ZulipAction.getFileTemporaryUrl`. As mentioned above, when this is ready for another review, please comment and unmark as draft.

chimnayajith:
@chrisbobbe This PR is ready for review. PTAL

lib/widgets/lightbox.dart
```diff
@@ -113,6 +118,67 @@ class _CopyLinkButton extends StatelessWidget {
...
+            downloadUrl = tempUrl ?? url;
+          }
+
+          if (Platform.isAndroid) {
+            final fileName = url.pathSegments.last.trim();
+            await downloader.downloadFile(downloadUrl.toString(), fileName);
+            message = zulipLocalizations.lightboxDownloadImageStarted;
+          } else {
+            message = zulipLocalizations.lightboxDownloadImageError;
+          }
```
chrisbobbe: If `Platform.isAndroid` is false, we show a snackbar saying "An error occurred while downloading the image."? That doesn't seem like the right behavior for our iOS users. 🙂

chimnayajith: Is the intended behavior to simply hide the download icon if not on android instead?

chimnayajith: @chrisbobbe Pushed a revision that conditionally shows the download button - it's now only visible on Android devices where the download functionality is actually supported.

chrisbobbe: > Is the intended behavior […]?

You've marked the commit as a fix for #42, so that's the first place you should look for an answer to this question. 🙂 I took a quick look at #42, and it doesn't say it's Android-only, so that means it's about both Android and iOS.

If you'd like to implement the feature for Android only for now, that's very reasonable. In that case please file an issue for the Android implementation (and link to that instead of #42 in your commit and PR description), so we can leave #42 open to track the iOS implementation.

chimnayajith:
@chrisbobbe Created a new issue #2011  to handle the Android implementation alone. PTAL,

lib/model/binding.dart
```diff
@@ -9,8 +9,8 @@ import 'package:image_picker/image_picker.dart' as image_picker;
 import 'package:package_info_plus/package_info_plus.dart' as package_info_plus;
 import 'package:url_launcher/url_launcher.dart' as url_launcher;
 import 'package:wakelock_plus/wakelock_plus.dart' as wakelock_plus;
-
```
chrisbobbe, line 12: nit: don't delete this blank line; the two sections don't become more closely related in this commit

lib/widgets/lightbox.dart
```diff
@@ -295,8 +354,8 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
       elevation: elevation,
       child: Row(children: [
         _CopyLinkButton(url: widget.src),
+        if (Platform.isAndroid) _DownloadImageButton(url: widget.src),
         // TODO(#43): Share image
-        // TODO(#42): Download image
```
chrisbobbe, line 302: nit: put the `_DownloadImageButton` after the share-image TODO, matching the order of the TODOs

lib/widgets/lightbox.dart
```diff
@@ -295,8 +354,8 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
       elevation: elevation,
       child: Row(children: [
         _CopyLinkButton(url: widget.src),
+        if (Platform.isAndroid) _DownloadImageButton(url: widget.src),
```
chrisbobbe, line 357:
```suggestion
        if (defaultTargetPlatform == TargetPlatform.android) 
          _DownloadImageButton(url: widget.src),
```

because `defaultTargetPlatform` can be controlled in the test environment

test/widgets/lightbox_test.dart
```diff
@@ -447,6 +448,35 @@ void main() {
       debugNetworkImageHttpClientProvider = null;
     });
 
+    testWidgets('download button triggers download', (tester) async {
```
chrisbobbe, line 451: Let's split this into two tests:

- one for Android, expecting the button to be present and respond correctly
- one for iOS, expecting the button to be absent

test/widgets/lightbox_test.dart
```diff
@@ -447,6 +448,35 @@ void main() {
...
+      prepareBoringImageHttpClient();
+      final message = eg.streamMessage(sender: eg.otherUser);
+      await setupPage(tester, message: message, thumbnailUrl: null);
+
+      final downloadButton = find.byIcon(Icons.download);
+
+      if (Platform.isAndroid){
+        expect(downloadButton, findsOneWidget);
```
chrisbobbe, line 459: We use `check` instead of `expect`:

```suggestion
        check(findDownloadButton).findsOne();
```

(see other tests for examples)

test/widgets/lightbox_test.dart
```diff
@@ -447,6 +448,35 @@ void main() {
       debugNetworkImageHttpClientProvider = null;
     });
 
+    testWidgets('download button triggers download', (tester) async {
+      prepareBoringImageHttpClient();
+      final message = eg.streamMessage(sender: eg.otherUser);
+      await setupPage(tester, message: message, thumbnailUrl: null);
+
+      final downloadButton = find.byIcon(Icons.download);
```
chrisbobbe, line 456: nit: let's name this `findDownloadButton`

assets/l10n/app_en.arb
```diff
@@ -751,6 +751,22 @@
   "@lightboxVideoDuration": {
     "description": "The total duration of the video playing in the lightbox."
   },
+  "lightboxDownloadImageTooltip": "Download image",
```
chrisbobbe, line 754: Sometimes we show a video in the lightbox. Do we offer a download button in that case? If so, let's just say "Download", because "image" would be inaccurate in that case. (And update the names of these strings to match, so "lightboxDownloadTooltip" instead of "lightboxDownloadImageTooltip", and similarly for the others. Also let's move these so they come immediately after the related entry "lightboxCopyLinkTooltip".

lib/widgets/lightbox.dart
```diff
@@ -113,6 +118,60 @@ class _CopyLinkButton extends StatelessWidget {
...
+
+          final fileName = url.pathSegments.last.trim();
+          await downloader.downloadFile(downloadUrl.toString(), fileName);
+          message = zulipLocalizations.lightboxDownloadImageStarted;
+        } catch (e) {
+          if (e is TimeoutException || e is SocketException) {
+            message = zulipLocalizations.lightboxDownloadImageError;
+          } else {
+            message = zulipLocalizations.lightboxDownloadImageFailed;
+          }
```
chrisbobbe, line 155: Could you explain why you chose

- "An error occurred while downloading the image." for `TimeoutException` or `SocketException`
- "Failed to download the image." for other errors

?

lib/widgets/lightbox.dart
```diff
@@ -113,6 +118,60 @@ class _CopyLinkButton extends StatelessWidget {
...
+          final internalLink = parseInternalLink(url, store);
+          if (internalLink is UserUploadLink){
+            if (!context.mounted) return;
+            final tempUrl = await ZulipAction.getFileTemporaryUrl(context, internalLink);
+            downloadUrl = tempUrl ?? url;
+          }
+
+          final fileName = url.pathSegments.last.trim();
+          await downloader.downloadFile(downloadUrl.toString(), fileName);
+          message = zulipLocalizations.lightboxDownloadImageStarted;
```
chrisbobbe, line 149: What does the Future returned by `downloader.downloadFile` represent? Does it resolve when the download finishes? If so, it seems wrong to show a "Starting download..." snackbar at that time; perhaps it should say "Download complete." or similar.

pubspec.lock
```diff
@@ -822,6 +822,54 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "2.3.0"
+  permission_handler:
```
chrisbobbe, line 825: Is `permission_handler` actually needed? From one of your messages on CZO, it seems like you discovered that it was not necessary after all:

[#mobile-team > Android APIs for downloading an image @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Android.20APIs.20for.20downloading.20an.20image/near/2064216)

> To avoid this [an issue with Android permissions], I was thinking of switching to the DownloadManager API ([Docs](https://developer.android.com/reference/android/app/DownloadManager)) as it handles permissions better across different Android versions. And so far as I have tested, it seems to be working on Android 10 and below too.

lib/widgets/lightbox.dart
```diff
@@ -113,6 +118,60 @@ class _CopyLinkButton extends StatelessWidget {
...
+        String message = zulipLocalizations.lightboxDownloadImageFailed;
+        try {
+          await requestStoragePermission();
+          Uri downloadUrl = url;
+          final internalLink = parseInternalLink(url, store);
+          if (internalLink is UserUploadLink){
+            if (!context.mounted) return;
+            final tempUrl = await ZulipAction.getFileTemporaryUrl(context, internalLink);
+            downloadUrl = tempUrl ?? url;
+          }
```
chrisbobbe, line 145: If the link is a `UserUploadLink` and `ZulipAction.getFileTemporaryUrl` returns null, we don't expect it to be possible to download the file. In that case, `ZulipAction.getFileTemporaryUrl` will have already shown a dialog saying "Could not access uploaded file". We should stop there, without trying actions we don't expect to work (like `downloader.downloadFile`) and without showing additional UI feedback like a snackbar.

lib/widgets/lightbox.dart
```diff
@@ -113,6 +118,60 @@ class _CopyLinkButton extends StatelessWidget {
...
+        }
+
+        scaffoldMessenger.showSnackBar(
+          SnackBar(behavior: SnackBarBehavior.floating, content: Text(message)));
+      });
+  }
+
+  Future<void> requestStoragePermission() async {
+    final androidInfo = await DeviceInfoPlugin().androidInfo;
```
chrisbobbe, line 164: Isn't there a method on `ZulipBinding` for this? (Also see a previous comment on whether `permission_handler` is needed at all)

chrisbobbe (commented):
Thanks! Comments below, focused on the parts other than the Pigeon / Android code, which I'd like to save for a future round.

assets/l10n/app_en.arb
chimnayajith: I want to double-check my understanding before changing this.

At the moment, the download button is only added in `_ImageLightboxPageState`, and videos are rendered via `VideoLightboxPage`, which doesn’t currently expose a download action.

Given that, should the button and strings be treated as media-generic already (i.e. renamed to just “Download” and `lightboxDownloadTooltip`), even though the functionality is currently image-only? Or would you prefer to keep the naming image-specific until video downloads are actually supported?

chrisbobbe: I don't feel strongly; I guess let's use the more general string "Download", and we can reuse it in a future where we offer a similar button for videos.

alya:
What's the source for the download icon shown in the PR description? It doesn't seem like it matches our icon style.

lib/widgets/lightbox.dart
chimnayajith: I wanted to distinguish between different types of failures:

- Network errors (`TimeoutException`/`SocketException`): "An error occurred while downloading the image."
- Other errors: "Failed to download the image."
My reasoning was that network errors are often temporary and the message "An error occurred..." suggests it might work on retry, while other errors (permissions, storage, etc.) are more definitive so "Failed" indicates something more systematic.

However, looking at it from a user's perspective, I'm not entirely confident this distinction adds value. 

Should it be simplified to a single generic message for all errors?

alya: I don't think that wording distinction would be meaningful for users. If there's something useful to communicate here, we'd need to be much more specific about it.

Issue #1080, opened 2024-11-23
i18n: Localize lists in prose (e.g. of names) instead of just calling `.join(', ')`


chrisbobbe:
For example, I think in Japanese you would want "1位、2位、3位" instead of "1位, 2位, 3位".

JavaScript has API for this:

```js
new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya', 'Zixuan'])
// 'Chris、Greg、Alya、Zixuan'
```

I'm not aware of a nice way to do this in Flutter, though, but I haven't checked in a while.

PR #397, opened 2023-11-20
Subscriptions page

sirpengi:
Adds a Subscriptions page showing the user's subscribed streams. Out of scope are collapsible topics and the ability to browse to all streams. There was no design spec for the situation where the user has no subscriptions so implemented a simple text notice here.

Thanks @chrisbobbe for the color helpers. I've found a lot that could be shared between this and the inbox code in #381 -- besides the collapsible bit, the design is exactly the same save the `fontSize` here is 18 instead of 17 in the inbox. I think we have a lot of space to consolidate design in the future.

Figma design: https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=171-12359&mode=design&t=4d0vykoYQ0KGpFuu-0

Simulator Screenshot - iPhone 14 - 2023-11-20 at 16 26 18
Simulator Screenshot - iPhone 14 - 2023-11-20 at 16 27 15
Simulator Screenshot - iPhone 14 - 2023-11-17 at 15 02 32
Simulator Screenshot - iPhone 14 - 2023-11-20 at 16 27 31

Fixes #187

PR #1148, opened 2024-12-12
 Sweep through UI strings to internationalize

PIG208:
Partial work for:
- #277

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -351,6 +351,13 @@
       "filename": {"type": "String", "example": "file.txt"}
     }
   },
+  "composeBoxLoadingMessage": "(loading message {messageId})",
+  "@composeBoxLoadingMessage": {
+    "description": "Placeholder in compose box showing the quoted message is currently loading.",
+    "placeholders": {
+      "messageId": {"type": "int", "example": "1234"}
+    }
+  },
   "unknownUserName": "(unknown user)",
   "@unknownUserName": {
     "description": "Name placeholder to use for a user when we don't know their name."
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -746,6 +746,13 @@
   "@messageIsMovedLabel": {
     "description": "Label for a moved message. (Use ALL CAPS for cased alphabets: Latin, Greek, Cyrillic, etc.)"
   },
+  "pollVoterNames": "({voterNames})",
+  "@pollVoterNames": {
+    "description": "The list of people who voted for a poll option, wrapped in parentheses.",
+    "placeholders": {
+      "voterNames": {"type":  "String", "example": "Alice, Bob, Chad"}
+    }
+  },
   "pollWidgetQuestionMissing": "No question.",
   "@pollWidgetQuestionMissing": {
     "description": "Text to display for a poll when the question is missing"
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -325,8 +325,8 @@
   "@composeBoxSendTooltip":  {
     "description": "Tooltip for send button in compose box."
   },
-  "composeBoxUnknownChannelName": "(unknown channel)",
-  "@composeBoxUnknownChannelName": {
+  "unknownChannelName": "(unknown channel)",
+  "@unknownChannelName": {
     "description": "Replacement name for channel when it cannot be found in the store."
   },
   "composeBoxTopicHintText": "Topic",
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -403,6 +403,14 @@
   "@lightboxCopyLinkTooltip": {
     "description": "Tooltip in lightbox for the copy link action."
   },
+  "lightboxVideoCurrentPosition": "Current position",
+  "@lightboxVideoCurrentPosition": {
+    "description": "The current playback position of the video playing in the lightbox."
+  },
+  "lightboxVideoDuration": "Video duration",
+  "@lightboxVideoDuration": {
+    "description": "The total duration of the video playing in the lightbox."
+  },
   "loginPageTitle": "Log in",
   "@loginPageTitle": {
     "description": "Title for login page."
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -586,6 +586,10 @@
   "@recentDmConversationsPageTitle": {
     "description": "Title for the page with a list of DM conversations."
   },
+  "recentDmConversationsSectionHeader": "Direct messages",
+  "@recentDmConversationsSectionHeader": {
+    "description": "Heading for direct messages section on the 'Inbox' message view."
+  },
   "combinedFeedPageTitle": "Combined feed",
   "@combinedFeedPageTitle": {
     "description": "Page title for the 'Combined feed' message view."
@@ -617,6 +621,18 @@
       "senderFullName": {"type": "String", "example": "Alice"},
       "numOthers": {"type": "int", "example": "4"}
     }
+  },
+  "pinnedSubscriptionsLabel": "Pinned",
+  "@pinnedSubscriptionsLabel": {
+    "description": "Label for the list of pinned subscribed channels."
+  },
+  "unpinnedSubscriptionsLabel": "Unpinned",
+  "@unpinnedSubscriptionsLabel": {
+    "description": "Label for the list of unpinned subscribed channels."
+  },
+  "subscriptionListNoChannels": "No channels found",
+  "@subscriptionListNoChannels": {
+    "description": "Text to display on subscribed-channels page when there are no subscribed channels."
   },
   "notifSelfUser": "You",
   "@notifSelfUser": {
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -355,6 +355,13 @@
       "others": {"type": "String", "example": "Alice, Bob"}
     }
   },
+  "dmsWithOthersPageTitle": "DMs with {others}",
+  "@dmsWithOthersPageTitle": {
+    "description": "Message list page title for a DM group with others.",
+    "placeholders": {
+      "others": {"type": "String", "example": "Alice, Bob"}
+    }
+  },
   "messageListGroupYouWithYourself": "You with yourself",
   "@messageListGroupYouWithYourself": {
     "description": "Message list recipient header for a DM group that only includes yourself."
@@ -748,5 +755,13 @@
   "emojiPickerSearchEmoji": "Search emoji",
   "@emojiPickerSearchEmoji": {
     "description": "Hint text for the emoji picker search text field."
+  },
+  "noEarlierMessages": "No earlier messages",
+  "@noEarlierMessages": {
+    "description": "Text to show at the start of a message list if there are no earlier messages."
+  },
+  "scrollToBottomTooltip": "Scroll to bottom",
+  "@scrollToBottomTooltip": {
+    "description": "Tooltip for button to scroll to bottom."
   }
 }
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -56,6 +56,10 @@
   "@profileButtonSendDirectMessage": {
     "description": "Label for button in profile screen to navigate to DMs with the shown user."
   },
+  "errorCouldNotShowUserProfile": "Could not show user profile.",
+  "@errorCouldNotShowUserProfile": {
+    "description": "Message that appears on the user profile page when the profile cannot be shown."
+  },
   "permissionsNeededTitle": "Permissions needed",
   "@permissionsNeededTitle": {
     "description": "Title for dialog asking the user to grant additional permissions."
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -344,6 +344,10 @@
   "@unknownUserName": {
     "description": "Name placeholder to use for a user when we don't know their name."
   },
+  "dmsWithYourselfPageTitle": "DMs with yourself",
+  "@dmsWithYourselfPageTitle": {
+    "description": "Message list page title for a DM group that only includes yourself."
+  },
   "messageListGroupYouAndOthers": "You and {others}",
   "@messageListGroupYouAndOthers": {
     "description": "Message list recipient header for a DM group with others.",
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -672,6 +672,10 @@
   "@notifSelfUser": {
     "description": "Display name for the user themself, to show after replying in an Android notification"
   },
+  "reactedEmojiSelfUser": "You",
+  "@reactedEmojiSelfUser": {
+    "description": "Display name for the user themself, to show on an emoji reaction added by the user."
+  },
   "onePersonTyping": "{typist} is typing…",
   "@onePersonTyping": {
     "description": "Text to display when there is one user typing.",
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -234,6 +234,17 @@
       "event": {"type": "String", "example": "UpdateMessageEvent(id: 123, messageIds: [2345, 3456], newTopic: 'dinner')"}
     }
   },
+  "errorCouldNotOpenLinkTitle": "Unable to open link",
+  "@errorCouldNotOpenLinkTitle": {
+    "description": "Error title when opening a link failed."
+  },
+  "errorCouldNotOpenLink": "Link could not be opened: {url}",
+  "@errorCouldNotOpenLink": {
+    "description": "Error message when opening a link failed.",
+    "placeholders": {
+      "url": {"type":  "String", "example":  "https://chat.example.com"}
+    }
+  },
   "errorMuteTopicFailed": "Failed to mute topic",
   "@errorMuteTopicFailed": {
     "description": "Error message when muting a topic failed."
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -204,7 +204,7 @@ class NotificationChannelManager {
 
     await _androidHost.createNotificationChannel(NotificationChannel(
       id: kChannelId,
-      name: 'Messages', // TODO(i18n)
+      name: 'Messages', // TODO(#1284)
       importance: NotificationImportance.high,
       lightsEnabled: true,
       soundUrl: defaultSoundUrl,
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -263,7 +263,7 @@ class NotificationDisplayManager {
       FcmMessageChannelRecipient(:var streamName?, :var topic) =>
         '#$streamName > ${topic.displayName}',
       FcmMessageChannelRecipient(:var topic) =>
-        '#(unknown channel) > ${topic.displayName}', // TODO get stream name from data
+        '#${zulipLocalizations.unknownChannelName} > ${topic.displayName}', // TODO get stream name from data
       FcmMessageDmRecipient(:var allRecipientIds) when allRecipientIds.length > 2 =>
         zulipLocalizations.notifGroupDmConversationLabel(
           data.senderFullName, allRecipientIds.length - 2), // TODO use others' names, from data
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -1570,6 +1570,7 @@ InlineSpan _errorUnimplemented(UnimplementedNode node, {required BuildContext co
   // because release mode isn't yet about general users but developer demos,
   // and we want to keep the demos honest.
   // TODO(#194) think through UX for general release
+  // TODO(#1285) translate this
   final htmlNode = node.htmlNode;
   if (htmlNode is dom.Element) {
     return TextSpan(children: [
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -1319,10 +1319,11 @@ class MessageTableCell extends StatelessWidget {
 
 void _launchUrl(BuildContext context, String urlString) async {
   DialogStatus showError(BuildContext context, String? message) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
     return showErrorDialog(context: context,
-      title: 'Unable to open link',
+      title: zulipLocalizations.errorCouldNotOpenLinkTitle,
       message: [
-        'Link could not be opened: $urlString',
+        zulipLocalizations.errorCouldNotOpenLink(urlString),
         if (message != null) message,
       ].join("\n\n"));
   }
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -147,6 +147,7 @@ class ReactionChip extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final store = PerAccountStoreWidget.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
 
     final reactionType = reactionWithVotes.reactionType;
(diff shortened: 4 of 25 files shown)


PIG208:
Marking this for review as I think it would be a good checkpoint. I did a sweep with `git grep "'" lib | grep -v // | grep -v 'import'` and `git grep '"' lib | grep -v // | grep -v 'import'` (as well as `git grep "'" lib | grep //` and `git grep '"' lib | grep //`), and these seem to be the ones left:

```
lib/widgets/content.dart
1576:      TextSpan(text: "(unimplemented:", style: errorStyle),
1578:      TextSpan(text: ")", style: errorStyle),
1582:      TextSpan(text: "(unimplemented: text «", style: errorStyle),
1584:      TextSpan(text: "»)", style: errorStyle),
1588:      text: "(unimplemented: DOM node type ${htmlNode.nodeType})",

lib/licenses.dart
26:      return '$licenseFileText\n\nAUTHORS file follows:\n\n$authorsFileText';

lib/model/compose.dart
148:  return '${mention(sender!, silent: true)} ${inlineLink('said', url)}: ' // TODO(i18n) ?
149:    '*(loading message ${message.id})*\n'; // TODO(i18n) ?
172:    return '${mention(sender!, silent: true)} ${inlineLink('said', url)}:\n' // TODO(i18n) ?
```

As well as the `.join` onese (#1080) that are out-of-scope for #277.

The licenses one can probably be translated with `GlobalLocalization`; the `compose.dart` ones involve links; and the `unimplemented` text uses different styles. Leaving those as follow-ups to the PR.

chrisbobbe (commented):
Thanks! Small comments below.

I see some remaining `TODO(i18n)`s that look relevant for #277; would these be easy to do?

- Some in lib/model/compose.dart
- One in lib/notifications/display.dart
- In poll.dart, `TODO(i18n): Localize parenthesis characters`—maybe we want to give this string to translators, rather than do it programmatically? This case seems similar to '#$streamName > $topic' which this PR would give to translators.

PIG208:
For the TODOs in `lib/model/compose.dart`, the reason for skipping those was given in https://github.com/zulip/zulip-flutter/pull/1148#issuecomment-2578100728.

For the one in lib/notifications/display.dart, the string to be translated appears in the Android settings menu, outside of the Flutter app, so it seems to be tricky.

For `TODO(i18n): Localize parenthesis characters`, it makes sense to have a string for that, which should be straightforward.

chrisbobbe:
> For the TODOs in `lib/model/compose.dart`, the reason for skipping those was given in [#1148 (comment)](https://github.com/zulip/zulip-flutter/pull/1148#issuecomment-2578100728).

Oops, I seem to have skipped right past that :) thanks for the bump. Looks like that reason is: "the `compose.dart` ones involve links".

Yeah, for the quote-and-reply string, we really want something like:

```
{user} <z>said</z>:
```

where we get handed the translation for "said" and do the Markdown link syntax around that. In zulip-mobile we have a few places like this:

```js
          text={{
            text: 'Suggestion: <z-link>{suggestedServerUrl}</z-link>',
            values: {
              suggestedServerUrl: suggestion,
              'z-link': chunks => (
                <ZulipText
                  inheritFontSize
                  style={styles.suggestionTextLink}
                  onPress={() => handlePressSuggestion(suggestion)}
                >
                  {chunks}
                </ZulipText>
              ),
            },
          }}
```

but in a quick skim I don't see something like that with the APIs we're using.

For this one, perhaps also worth a mention in the followup issue that we might want to translate this into the org's language, not the user's language.

chrisbobbe:
The other string to translate in `lib/model/compose.dart` is, I think:

```
(loading message {messageId})
```

That seems straightforward, and seems appropriate to translate into the user's language rather than the org's language, since it's meant as feedback to the user. Still fine to file as a followup issue :) but it'll be a tiny one.

chrisbobbe:
> For the one in lib/notifications/display.dart, the string to be translated appears in the Android settings menu, outside of the Flutter app, so it seems to be tricky.

Does `GlobalLocalizations` have any restrictions on where it can be used? If so, those would be good to document. Adding that documentation could be a followup issue if it's complicated or not yet understood, along with the issue of using it in this file.

If it's available to use there in `lib/notifications/display.dart`, I guess we can just go for it? 🤷‍♂️

assets/l10n/app_en.arb
```diff
@@ -618,6 +618,14 @@
   "@channelFeedButtonTooltip": {
     "description": "Tooltip for button to navigate to a given channel's feed"
   },
+  "notifChannelConversationLabel": "#{channel} > {topic}",
```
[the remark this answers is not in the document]
PIG208: Yeah. That will be up to the translator to handle then.

assets/l10n/app_en.arb
```diff
@@ -708,5 +708,13 @@
   "emojiPickerSearchEmoji": "Search emoji",
   "@emojiPickerSearchEmoji": {
     "description": "Hint text for the emoji picker search text field."
+  },
+  "videoCurrentPositionLabel": "Current position",
+  "@videoCurrentPositionLabel": {
+    "description": "The current playback position."
```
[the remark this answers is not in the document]
PIG208: Renaming this to `lightboxVideoCurrentPositionLabel` and updated the description to `The current playback position of the video playing in the lightbox` to emphasize that it appears in the lightbox.

assets/l10n/app_en.arb
```diff
@@ -234,6 +234,17 @@
...
+  "@errorOpenLinkTitle": {
+    "description": "Error title when opening a link failed."
+  },
+  "errorOpenLinkDetails": "Link could not be opened: {url}",
+  "@errorOpenLinkDetails": {
+    "description": "Error details when opening a link failed.",
+    "placeholders": {
+      "url": {"type":  "String", "example":  "https://chat.example.com"}
+    }
+  },
```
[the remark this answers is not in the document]
PIG208: Looks like we use `fooDialogMessage` and `fooDialogTitle`, but they are not so common (1 and 2 occurences, respectively). Updating the names to use these suffixes instead.

assets/l10n/app_en.arb
```diff
@@ -664,9 +664,9 @@
   "@noSubscriptions": {
     "description": "Text to display when there are no subscribed channels."
   },
-  "notifSelfUser": "You",
-  "@notifSelfUser": {
-    "description": "Display name for the user themself, to show after replying in an Android notification"
+  "selfUser": "You",
+  "@selfUser": {
+    "description": "Display name for the user themself, to show after replying in an Android notification or on an emoji reaction added by themself."
```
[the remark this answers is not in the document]
PIG208: Having additional context for where each "You" might appear could be helpful, as I find "You" a bit obscure without context, so I'm adding another string.

PIG208:
Thanks for the review! The PR has been updated. I also went through the existing i18n TODOs, and the outputs of `git grep "'" lib | grep -vxFf <(git grep '//' lib) | grep -vxFf <(git grep 'import.*'"'" lib)` and `git grep '"' lib | grep -vxFf <(git grep '//' lib)` again to double-check.

chrisbobbe:

assets/l10n/app_en.arb
```diff
@@ -618,10 +667,26 @@
...
+  "@pinnedSubscriptionsLabel": {
+    "description": "Label for the list of pinned subscribed channels."
+  },
+  "unpinnedSubscriptionsLabel": "Unpinned",
+  "@unpinnedSubscriptionsLabel": {
+    "description": "Label for the list of unpinned subscribed channels."
+  },
+  "noSubscriptions": "No channels found",
+  "@noSubscriptions": {
+    "description": "Text to display when there are no subscribed channels."
```
gnprice, line 680:
```suggestion
  "noSubscriptions": "No channels found",
  "@noSubscriptions": {
    "description": "Text to display on subscribed-channels page when there are no subscribed channels."
```

lib/widgets/inbox.dart
```diff
@@ -312,7 +314,8 @@ class _AllDmsHeaderItem extends _HeaderItem {
     required super.sectionContext,
   });
 
-  @override String get title => 'Direct messages'; // TODO(i18n)
+  @override String title(ZulipLocalizations zulipLocalizations) =>
+    zulipLocalizations.recentDmConversationsPageTitle;
```
gnprice, line 318: Let's make this a separate string — I feel like it's not a given that the heading in this list will be the same as the title atop the other page.

lib/widgets/subscription_list.dart
```diff
@@ -77,10 +78,8 @@ class _SubscriptionListPageBodyState extends State<SubscriptionListPageBody> wit
 
     // TODO: Implement collapsible topics
 
-    // TODO(i18n): localize strings on page
-    //   Strings here left unlocalized as they likely will not
-    //   exist in the settled design.
```
gnprice, line 82: Huh, what was this about?

lib/widgets/subscription_list.dart
```diff
@@ -124,11 +123,12 @@ class _NoSubscriptionsItem extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
 
     return SliverToBoxAdapter(
       child: Padding(
         padding: const EdgeInsets.all(10),
-        child: Text("No channels found",
+        child: Text(zulipLocalizations.noSubscriptions,
```
gnprice, line 131: This feels like a mismatch between the name and the actual string. That risks being confusing when looking at the code and thinking about how it relates to the UI. Let's try `subscriptionListNoChannels`.

That name starts by scoping where this is found, and then gives the gist of the string's actual content.

gnprice (commented):
Thanks for taking care of these! And thanks @chrisbobbe for the previous reviews.

Here's an initial review, on the first few commits:
32caefa38 display [nfc]: Mention issue for notification channel name's i18n TODO
2523d9cbc i18n [nfc]: Mention issue for translating styled strings
1494908ed app: Translate the "about" page button on popup menu
16211c918 inbox: Make header titles localizable

lib/widgets/subscription_list.dart
PIG208: This was from #397, which followed [an old design](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=171-12359&t=ilEg6ymsxO6JLNQ5-0). The comment is no longer accurate as we localize the strings in the PR. It doesn't seem useful to revise this, either.

PIG208:
Updated the PR, thanks for the review!

lib/widgets/profile.dart
```diff
@@ -121,17 +121,18 @@ class _ProfileErrorPage extends StatelessWidget {
...
+          padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 32),
           child: Row(
             mainAxisAlignment: MainAxisAlignment.center,
             children: [
-              Icon(Icons.error),
-              SizedBox(width: 4),
-              Text('Could not show user profile.'),
+              const Icon(Icons.error),
+              const SizedBox(width: 4),
+              Text(zulipLocalizations.errorShowUserProfile),
```
gnprice, line 135: Let's name this to match the other error strings we have — the name describes the failure, not (or not just) the attempted operation. Here's existing examples:
```
$ git grep -ohP '^  "error.*?"' assets/l10n/app_en.arb 
  "errorShowUserProfile"
  "errorWebAuthOperationalErrorTitle"
  "errorWebAuthOperationalError"
  "errorAccountLoggedInTitle"
  "errorAccountLoggedIn"
  "errorCouldNotFetchMessageSource"
  "errorCopyingFailed"
  "errorFailedToUploadFileTitle"
  "errorFilesTooLarge"
  "errorFilesTooLargeTitle"
  "errorLoginInvalidInputTitle"
  "errorLoginFailedTitle"
  "errorMessageNotSent"
  "errorLoginCouldNotConnect"
  "errorLoginCouldNotConnectTitle"
  "errorMessageDoesNotSeemToExist"
  "errorQuotationFailed"
  "errorServerMessage"
  "errorConnectingToServerShort"
  "errorConnectingToServerDetails"
  "errorHandlingEventTitle"
  "errorHandlingEventDetails"
  "errorOpenLinkDialogTitle"
  "errorOpenLinkDialogMessage"
  "errorMuteTopicFailed"
  "errorUnmuteTopicFailed"
  "errorFollowTopicFailed"
  "errorUnfollowTopicFailed"
  "errorSharingFailed"
  "errorStarMessageFailedTitle"
  "errorUnstarMessageFailedTitle"
  "errorBannerDeactivatedDmLabel"
  "errorBannerCannotPostInChannelLabel"
  "errorDialogContinue"
  "errorDialogTitle"
  "errorInvalidResponse"
  "errorNetworkRequestFailed"
  "errorMalformedResponse"
  "errorMalformedResponseWithCause"
  "errorRequestFailed"
  "errorVideoPlayerFailed"
  "errorMarkAsReadFailedTitle"
  "errorMarkAsUnreadFailedTitle"
  "errorNotificationOpenTitle"
  "errorNotificationOpenAccountMissing"
  "errorReactionAddingFailedTitle"
  "errorReactionRemovingFailedTitle"
```

A minimal fix would be `errorShowingUserProfile` (it's an error in showing a user profile), or `errorCouldNotShowUserProfile`.

lib/widgets/message_list.dart
```diff
@@ -413,10 +415,13 @@ class MessageListAppBarTitle extends StatelessWidget {
       case DmNarrow(:var otherRecipientIds):
         final store = PerAccountStoreWidget.of(context);
         if (otherRecipientIds.isEmpty) {
-          return const Text("DMs with yourself");
+          return Text(zulipLocalizations.messageListYouWithYourselfTitle);
```
gnprice, line 418: Would you do a pass through the rest of the branch to find any other places where the string name doesn't align with its value? Similar to https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1936668186 above, and to this one.

For this one, let's also make the name parallel the existing examples in the other cases of this `switch` — so `dmsWithYourselfPageTitle`.

gnprice (commented):
Thanks. A reply above, and here's from reading the next two commits:
d6d85e4fb profile: Translate strings on _ProfileErrorPage
e1e35f503 msglist: Localize message list title for "DMs with yourself"

LMK when you've done the pass requested below, and I'll return to review the rest of the branch.

lib/widgets/profile.dart
PIG208: Noting the use of `-oh` to show only the matches. That is helpful for finding naming examples.

lib/widgets/message_list.dart
PIG208: `messageListDMsWithOthers` -> `dmsWithOthersPageTitle`
`errorOpenLinkDialogMessage` -> `errorCouldNotOpenLinkDialogMessage`
`errorOpenLinkDialogTitle` -> `errorCouldNotOpenLinkDialogTitle`

I kept "reactedEmojiSelfUser" as-is:

```
  "reactedEmojiSelfUser": "You",
  "@reactedEmojiSelfUser": {
    "description": "Display name for the user themself, to show on an emoji reaction added by the user."
  },
```

to match `notifSelfUser`:

```
  "notifSelfUser": "You",
  "@notifSelfUser": {
    "description": "Display name for the user themself, to show after replying in an Android notification"
  },
```

`composeBoxLoadingMessage` is left as-is to match `composeBoxUploadingFilename`, but maybe we actually want to end both with a `Placeholder` suffix?

lib/widgets/content.dart
```diff
@@ -1319,10 +1319,11 @@ class MessageTableCell extends StatelessWidget {
 
 void _launchUrl(BuildContext context, String urlString) async {
   DialogStatus showError(BuildContext context, String? message) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
     return showErrorDialog(context: context,
-      title: 'Unable to open link',
+      title: zulipLocalizations.errorCouldNotOpenLinkDialogTitle,
       message: [
-        'Link could not be opened: $urlString',
+        zulipLocalizations.errorCouldNotOpenLinkDialogMessage(urlString),
```
gnprice, line 1326: The "dialog title" and "dialog message" suffixes here don't match any existing string names, as listed at https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1937976609 . They also make these names pretty long and I think risk making them hard for a reader to parse.

Instead, you can see examples there like these:
```
  "errorWebAuthOperationalErrorTitle"
  "errorWebAuthOperationalError"
  "errorAccountLoggedInTitle"
  "errorAccountLoggedIn"
  "errorFilesTooLarge"
  "errorFilesTooLargeTitle"
```

You can also find examples with `git grep -A4 showErrorDialog`. For example:
```
lib/widgets/compose_box.dart
661:    showErrorDialog(
662-      context: context,
663-      title: zulipLocalizations.errorFilesTooLargeTitle(tooLargeFiles.length),
664-      message: zulipLocalizations.errorFilesTooLarge(
665-        tooLargeFiles.length,
```

lib/notifications/display.dart
```diff
@@ -261,9 +261,10 @@ class NotificationDisplayManager {
     // the first.
     messagingStyle.conversationTitle = switch (data.recipient) {
       FcmMessageChannelRecipient(:var streamName?, :var topic) =>
-        '#$streamName > ${topic.displayName}',
+        zulipLocalizations.notifChannelConversationLabel(streamName, topic.displayName),
```
gnprice, line 264: I don't think translating this makes sense: the "#" and ">" come from how Zulip displays channels and topics, not from any convention of English punctuation. We use the "#" in icons, and we use both the "#" and ">" in the Markdown syntax.

Because they don't come from English punctuation conventions, it doesn't make sense to try to find their equivalents in other languages. And because they do come from how Zulip displays channels and topics in other contexts, if they got translated here to anything else it would just cause confusion.

lib/widgets/lightbox.dart
```diff
@@ -385,13 +385,14 @@ class _VideoPositionSliderControlState extends State<_VideoPositionSliderControl
...
   Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
     final currentPosition = _isSliderDragging
       ? _sliderValue
       : widget.controller.value.position;
 
     return Row(children: [
       VideoDurationLabel(currentPosition,
-        semanticsLabel: "Current position"),
+        semanticsLabel: zulipLocalizations.lightboxVideoCurrentPositionLabel),
```
gnprice, line 395:
```suggestion
        semanticsLabel: zulipLocalizations.lightboxVideoCurrentPosition),
```

The "label" part doesn't really add any meaning here.

assets/l10n/app_en.arb
```diff
@@ -395,6 +428,14 @@
   "@lightboxCopyLinkTooltip": {
     "description": "Tooltip in lightbox for the copy link action."
   },
+  "lightboxVideoCurrentPositionLabel": "Current position",
+  "@lightboxVideoCurrentPositionLabel": {
+    "description": "The current playback position of the video playing in the lightbox."
+  },
+  "lightboxVideoDurationLabel": "Video duration",
+  "@lightboxVideoDurationLabel": {
+    "description": "The total duration of the video playing in the lightbox."
```
gnprice, line 437: This descriptions were adapted from [VideoPlayerValue].

Good thought.

lib/widgets/emoji_reaction.dart
```diff
@@ -160,8 +161,8 @@ class ReactionChip extends StatelessWidget {
       //   // 'Chris、Greg、Alya、Shu'
       ? userIds.map((id) {
           return id == store.selfUserId
-            ? 'You'
-            : store.users[id]?.fullName ?? '(unknown user)'; // TODO(i18n)
+            ? zulipLocalizations.reactedEmojiSelfUser
```
gnprice, line 164: Yeah, this name sounds good (re https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1938033435 ).

assets/l10n/app_en.arb
```diff
@@ -685,6 +754,13 @@
   "@messageIsMovedLabel": {
     "description": "Label for a moved message. (Use ALL CAPS for cased alphabets: Latin, Greek, Cyrillic, etc.)"
   },
+  "pollVoterNames": "({voterNames})",
+  "@pollVoterNames": {
+    "description": "The list of people voted for a poll option, wrapped in parentheses.",
```
gnprice, line 759:
```suggestion
    "description": "The list of people who voted for a poll option, wrapped in parentheses.",
```

gnprice (commented):
Thanks for the revision! I've now read the whole thing; remaining comments below.

lib/widgets/message_list.dart
gnprice: Thanks. Replied on specific items in separate threads:
https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941981911
https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1942000026

lib/widgets/content.dart
PIG208: Yeah, it stemmed from the previous discussion here: https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1915759999.

> They also make these names pretty long and I think risk making them hard for a reader to parse.

is a good point. I thought going a bit verbose can be helpful for showing the context where they appear.

gnprice: Yeah, I think we can leave it to the description to spell that out. The description is in normal prose instead of an identifier, so it has a bit more room :slightly_smiling_face: 

The one existing use of "…DialogMessage" is:
```
  "logOutConfirmationDialogMessage": "To use this account in the future, you will have to re-enter the URL for your organization and your account information.",
```

One operative difference there is that it doesn't say "error" at the start (because it's not an error). So there isn't that bit of context already suggesting this is either a dialog or some other such interaction where an error is presented. Instead that name communicates that by saying "confirmation dialog", which is the name for where that message appears.

PIG208: For these errors, our descriptions are mostly "Error message when…" and "Error title when…" Will update these two strings to be consistent with them.

PIG208:
Also adding a fixes comment now that we have the follow-up issues left out of scope for #277.

> Filed #1285 for `said` and `(unimplemented`, and #1284 for notification channel name.

gnprice:
Thanks! These changes all look good.

Since this is marked now as fixing the whole issue, I went back through the same searches described at https://github.com/zulip/zulip-flutter/pull/1148#issuecomment-2549318850 . Here's one hit that also appeared there:
```
lib/licenses.dart
26:      return '$licenseFileText\n\nAUTHORS file follows:\n\n$authorsFileText';
```

Would you fix that, and also look through the rest of the search results from that point on?

Meanwhile I'll merge this but leave out the "Fixes:" for now.

PIG208:
I went through both queries before the previous push, and didn't find more occurrences since last time.

For the licenses one, we discussed [here](https://chat.zulip.org/#narrow/channel/176/topic/Pygment.20license) and decided to leave it as-is.

gnprice:
Yep, that should get a comment :-). Thanks.

Cross-linking, here's what I found when I went and completed the grep sweep just now:
https://github.com/zulip/zulip-flutter/issues/277#issuecomment-2638297174

lib/notifications/display.dart
chrisbobbe: My thought was:

https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1915755966
> Huh, yeah I guess maybe this could look different in RTL languages, or languages that use a space character other than " "?

but

- I'm not confident about the RTL point; I'd want feedback from a regular user of RTL interfaces
- The space-character thing is a nit, especially since (AIUI) this string doesn't appear in a paragraph (where localized space characters would be used)

and indeed, incorrect/inconsistent translations is a possibility; helpful to prevent when possible.

Issue #744, opened 2024-06-18
Offer "mark as resolved"


gnprice:
We should offer the "mark as resolved" and "mark as unresolved" actions in the UI, to operate on a topic. Help Center page here:
https://zulip.com/help/resolve-a-topic

This will be part of the topic action sheet. At the moment we don't yet have a topic action sheet; this can be the first item in it, or this issue can come after #348 which will also add items to it. See #348 for where the topic action sheet should appear in the UI.

Related chat discussion:
https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/resolving.20topics/near/1829650

Prerequisite:
* #150

PR #1150, opened 2024-12-13
unreads: Fix false positives in a debug check; log in another case when helpful

chrisbobbe:
In manual testing for my work on resolve/unresolve #744, I noticed some surprising debug-log lines that turned out to be false positives, so this fixes those.

It also makes the debug check run in a helpful case where it wasn't running before.

### DIFF
--- a/test/model/unreads_test.dart
+++ b/test/model/unreads_test.dart
@@ -240,6 +240,49 @@ void main() {
     });
   });
 
+  group('isUnread', () {
+    final unreadDmMessage = eg.dmMessage(
+      from: eg.otherUser, to: [eg.selfUser], flags: []);
+    final readDmMessage = eg.dmMessage(
+      from: eg.otherUser, to: [eg.selfUser], flags: [MessageFlag.read]);
+    final unreadChannelMessage = eg.streamMessage(flags: []);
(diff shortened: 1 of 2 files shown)


test/model/unreads_test.dart
```diff
@@ -240,6 +240,49 @@ void main() {
...
+      assert(expected == true || expected == null);
+      test('oldUnreadsMissing; $description', () {
+        prepare();
+        model.oldUnreadsMissing = true;
+        fillWithMessages(allMessages);
+        check(model.isUnread(messageId)).equals(expected);
+      });
+    }
+
+    doTestCommon('unread DM message',      unreadDmMessage.id,      true);
```
PIG208, line 275: nit: Should we make the third parameter use a keyword? Just by looking at it it is not so obvious what this is supposed to mean.

PIG208 (commented):
Looks good! Left a comment.

test/model/unreads_test.dart
chrisbobbe: Sure; made this a named param.

chrisbobbe:
Thanks for the review! Revision pushed. Since you had just one nit, I'll go ahead and mark for Greg's review.

gnprice:
Thanks! Looks good; merging.

PR #1154, opened 2024-12-13
Hero animation for images should run only to/from lightbox, not between message lists

sher999:
Passed `narrow` and `message.topic` augment to the tag parameter we pass to Hero so that it identifies not only the message and the image within the message, but the message list the message is shown in

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -15,6 +15,7 @@ import '../model/avatar_url.dart';
 import '../model/binding.dart';
 import '../model/content.dart';
 import '../model/internal_link.dart';
+import '../model/narrow.dart';
 import 'code_block.dart';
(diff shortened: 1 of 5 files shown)


PIG208:
Thanks for working on this @sher999! Before we can review this change, it will need a test: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request

gnprice:
Closing this PR because it doesn't look like there's work continuing on it.

Thanks @sher999 for making a go at this issue. If you're still interested in contributing to Zulip, you're welcome to make another try at this or another issue. Do take a look at Zulip's Git style guide, and for your next PR make sure to revise the sequence of commits accordingly — that's one of the next things we'd say about this PR in its current revision.

Issue #1130, opened 2024-12-11
Handle TeX blocks with blank lines


gnprice:
Sylwester Cyba [discovered](https://chat.zulip.org/#narrow/channel/48-mobile/topic/some.20TeX.20blocks.20unimplemented/near/1998668) that some LaTeX math blocks are unimplemented syntax.

In particular this message reproduces the issue ([example](https://chat.zulip.org/#narrow/channel/7-test-here/topic/.E2.9C.94.20mathsmaths/near/1998681)):
````markdown
```math
a

b
```
````

Without the blank line, the issue doesn't repro.

This situation doesn't seem super common, but we should handle it.

I haven't yet investigated to determine what's different in the HTML that causes it to not match the structures we're looking for.

PR #1156, opened 2024-12-13
content: Handle multiple math blocks in `<p>`

rajveermalviya:
Fixes: #1130

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -1055,13 +1055,6 @@ class _ZulipContentParser {
     return inlineParser.parseBlockInline(nodes);
   }
 
-  BlockContentNode parseMathBlock(dom.Element element) {
-    final debugHtmlNode = kDebugMode ? element : null;
-    final texSource = _parseMath(element, block: true);
-    if (texSource == null) return UnimplementedBlockContentNode(htmlNode: element);
-    return MathBlockNode(texSource: texSource, debugHtmlNode: debugHtmlNode);
-  }
-
   BlockContentNode parseListNode(dom.Element element) {
     ListStyle? listStyle;
     switch (element.localName) {
@@ -1453,6 +1446,64 @@ class _ZulipContentParser {
     return tableNode ?? UnimplementedBlockContentNode(htmlNode: tableElement);
   }
 
+  void parseMathBlocks(dom.NodeList nodes, List<BlockContentNode> result) {
+    assert(nodes.isNotEmpty);
+    assert((() {
+      final first = nodes.first;
+      return first is dom.Element
+        && first.localName == 'span'
+        && first.className == 'katex-display';
+    })());
+
+    final firstChild = nodes.first as dom.Element;
+    final texSource = _parseMath(firstChild, block: true);
+    if (texSource != null) {
+      result.add(MathBlockNode(
+        texSource: texSource,
+        debugHtmlNode: kDebugMode ? firstChild : null));
+    } else {
+      result.add(UnimplementedBlockContentNode(htmlNode: firstChild));
+    }
+
+    // Skip further checks if there was only a single child.
+    if (nodes.length == 1) return;
+
+    // The case with the `<br>\n` can happen when at the end of a quote;
+    // it seems like a glitch in the server's Markdown processing,
+    // so hopefully there just aren't any further such glitches.
+    bool hasTrailingBreakNewline = false;
+    if (nodes case [..., dom.Element(localName: 'br'), dom.Text(text: '\n')]) {
+      hasTrailingBreakNewline = true;
+    }
+
+    final length = hasTrailingBreakNewline
+      ? nodes.length - 2
+      : nodes.length;
+    for (int i = 1; i < length; i++) {
+      final child = nodes[i];
+      final debugHtmlNode = kDebugMode ? child : null;
+
+      // If there are multiple <span class="katex-display"> nodes in a <p>
+      // each node is interleaved by '\n\n'. Whitespaces are ignored in HTML
+      // on web but each node has `display: block`, which renders each node
+      // on a new line. Since the emitted MathBlockNode are BlockContentNode,
+      // we skip these newlines here to replicate the same behavior as on web.
+      if (child case dom.Text(text: '\n\n')) continue;
+
+      if (child case dom.Element(localName: 'span', className: 'katex-display')) {
+        final texSource = _parseMath(child, block: true);
+        if (texSource != null) {
+          result.add(MathBlockNode(
+            texSource: texSource,
+            debugHtmlNode: debugHtmlNode));
+          continue;
+        }
+      }
+
+      result.add(UnimplementedBlockContentNode(htmlNode: child));
+    }
+  }
+
   BlockContentNode parseBlockContent(dom.Node node) {
     final debugHtmlNode = kDebugMode ? node : null;
     if (node is! dom.Element) {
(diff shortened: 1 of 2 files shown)


lib/model/content.dart
```diff
@@ -1448,6 +1448,43 @@ class _ZulipContentParser {
...
+    })());
+    final debugHtmlNode = kDebugMode ? element : null;
+
+    final blocks = <BlockContentNode>[];
+    for (final child in element.nodes) {
+      // The case with the `<br>\n` can happen when at the end of a quote;
+      // it seems like a glitch in the server's Markdown processing,
+      // so hopefully there just aren't any further such glitches.
+      if (child case dom.Element(localName: 'br')) continue;
+      if (child case dom.Text(text: '\n' || '\n\n')) continue;
```
PIG208, line 1468: Looks like this no longer checks if the `<br>\n` is at the end. Would it be unexpected to have them appear in the middle?

`\n\n` on the other hand, appear to be somehow separate from the quirk commented here.

PIG208:
My [test message](https://chat.zulip.org/#narrow/channel/7-test-here/topic/.E2.9C.94.20mathsmaths/near/2023185) renders like this:

<details>
<summary>Screenshot</summary>

image

</details>

I think this is a result of parsing each `span` as a `MathBlockNode`, which is probably fine, but visually looks a bit off with all the borders. Making an alternative design should be out of scope for the fix, though.

test/model/content_test.dart
```diff
@@ -485,6 +485,23 @@ class ContentExample {
...
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mi>a</mi></mrow>'
+          '<annotation encoding="application/x-tex">a</annotation></semantics></math></span>'
+        '<span class="katex-html" aria-hidden="true"><span class="base"><span class="strut" style="height:0.4306em;"></span><span class="mord mathnormal">a</span></span></span></span></span>\n\n'
+      '<span class="katex-display"><span class="katex">'
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mi>b</mi></mrow>'
+          '<annotation encoding="application/x-tex">b</annotation></semantics></math></span>'
+        '<span class="katex-html" aria-hidden="true"><span class="base"><span class="strut" style="height:0.6944em;"></span><span class="mord mathnormal">b</span></span></span></span></span></p>', [
+      MathBlockNode(texSource: 'a'),
+      MathBlockNode(texSource: 'b'),
+    ]);
```
PIG208, line 503: Let's also have a test with multiple math blocks in a quote.

PIG208:
Thanks for the PR! Apparently I've missed this one as it has been sitting for some time. Overall this works fine for me. Left some comments.

lib/model/content.dart
rajveermalviya: Separated the handling of both these cases.

rajveermalviya:
Thanks for the review @PIG208. Pushed a new revision, PTAL.

lib/model/content.dart
```diff
@@ -1451,6 +1451,60 @@ class _ZulipContentParser {
...
+      ? element.nodes.length - 2
+      : element.nodes.length;
+    for (var i = 0; i < length; i++) {
+      final child = element.nodes[i];
+
+      // If there are multiple <span class="katex-display"> nodes in a <p>
+      // each node is interleaved by '\n\n'. Whitespaces are ignored in HTML
+      // on web but each node has `display: block`, which renders each node
+      // on a newline. So do the same here.
+      if (child case dom.Text(text: '\n\n')) continue;
```
PIG208, line 1487: I can see how the first sentence is relevant. Regarding "do the same here", `continue` itself does not seem to address the part about rendering each node with `display: block`, as that is more relevant to `blocks.add(MathBlockNode(texSource: texSource, debugHtmlNode: debugHtmlNode));`.

PIG208 (commented):
Looks good to me! Left some small comments.

PIG208:
Looks good to me! Thanks for the update.

lib/model/content.dart
```diff
@@ -1451,6 +1451,61 @@ class _ZulipContentParser {
...
+    assert(element.localName == 'p');
+    assert((() {
+      final first = element.nodes.first;
+      return first is dom.Element
+        && first.localName == 'span'
+        && first.className == 'katex-display';
+    })());
+    final debugHtmlNode = kDebugMode ? element : null;
+
+    final blocks = <BlockContentNode>[];
```
gnprice, line 1465: My performance-paranoia for this parsing code doesn't like the idea of allocating this list, just to immediately empty it into the surrounding list, particularly in the common case where there's only going to be a single math block in the `<p>` element. (And here "common case" means "everything but the weird edge case".)

But I think avoiding that would require a larger refactor, so let's go ahead with this way. It's probably fine (that is, it's probably not a big performance impact in practice). And at least it only affects messages with math blocks in them, which isn't a super common feature in the first place.

lib/model/content.dart
```diff
@@ -1451,6 +1451,61 @@ class _ZulipContentParser {
...
+      // we skip these newlines here to replicate the same behavior as on web.
+      if (child case dom.Text(text: '\n\n')) continue;
+
+      if (child is! dom.Element) {
+        blocks.add(UnimplementedBlockContentNode(htmlNode: child));
+        continue;
+      }
+      if (child.className != 'katex-display') {
+        blocks.add(UnimplementedBlockContentNode(htmlNode: child));
+        continue;
```
gnprice, line 1496: My performance-paranoia about this code also doesn't like having all these checks for the first child, where we already know the answers to them; particularly since that'll almost always be the only child.

But for this one, there is a clean solution, so let's do that. Before the loop, operate on just the first child; it'll need an `as dom.Element`, but none of the rest of these checks. And it'll duplicate the last few lines of the loop body, but that's OK.

Then the loop runs from 1 to `< length`, so will normally have no iterations at all.

gnprice (commented):
Thanks @rajveermalviya for taking care of this, and thanks @PIG208 for the previous reviews!

Generally this looks good. Comments below, including one fun edge case in the actual content syntax.

lib/model/content.dart
rajveermalviya: Added the suggested optimization for the first child.

gnprice (commented):
Thanks for the revision! This logic generally all looks good; small comments below.

Would you also split the changes into two commits? I think that would help with thinking through exactly what cases are affected. Specifically:
* A first commit is nearly NFC, and moves the logic from `parseBlockContent` up to its callers. I think this can be NFC apart from some unimplemented-content cases. (It wouldn't change which possible messages contain unimplemented content, but might change the details of what the unimplemented nodes look like.)
* Then the second commit makes the desired substantive change here, supporting multiple blocks in one `p` element.

rajveermalviya:

lib/model/content.dart
```diff
@@ -1599,6 +1584,30 @@ class _ZulipContentParser {
...
+          // so hopefully there just aren't any further such glitches.
+          if (currentParagraph.isNotEmpty) consumeParagraph();
+          if (imageNodes.isNotEmpty) consumeImageNodes();
+          result.add(parseMathBlock(child));
+          continue;
+        } else {
+          if (currentParagraph.isNotEmpty) consumeParagraph();
+          if (imageNodes.isNotEmpty) consumeImageNodes();
+          result.add(UnimplementedBlockContentNode(htmlNode: node));
+          continue;
```
gnprice: I believe this first commit is not quite NFC because the behavior changes in this case. In the old code, I believe this would end up with a ParagraphNode containing an UnimplementedInlineContentNode (and maybe other nodes), instead of with this one UnimplementedBlockContentNode.

That's the non-NFC-ness I was expecting at https://github.com/zulip/zulip-flutter/pull/1156#pullrequestreview-2613718416 . The change is fine; just the commit message should be adjusted: remove the NFC tag, and have the body explain that it's NFC except for this particular behavior change.

(I did work through the logic just now and satisfy myself that there's no *other* behavior change in this commit. So this splitting of commits was helpful, thanks.)

lib/model/content.dart
```diff
@@ -1599,6 +1584,30 @@ class _ZulipContentParser {
...
+                         || [_, dom.Element(localName: 'br'),
+                                dom.Text(text: "\n")]) {
+          // This might be too specific; we'll find out when we do #190.
+          // The case with the `<br>\n` can happen when at the end of a quote;
+          // it seems like a glitch in the server's Markdown processing,
+          // so hopefully there just aren't any further such glitches.
+          if (currentParagraph.isNotEmpty) consumeParagraph();
+          if (imageNodes.isNotEmpty) consumeImageNodes();
+          result.add(parseMathBlock(child));
+          continue;
```
gnprice: nit: How about moving the two "if not empty, consume" lines above this if/else, and the `continue` below it?

That would deduplicate them between the `if` and the `else` cases, showing that those parts don't vary. And it would also reduce the diff in the second commit.

gnprice (commented):
Thanks for the revision! This all looks great; just a couple of small comments below about the commit structure.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

Issue #1095, opened 2024-12-02
Display user facing version information in menu


PIG208:
#1035 is a prerequisite of this.

Figma design [here](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=2957-50165&node-type=frame&t=aQozr0u9uz0CWx6G-0)

Gaurav-Kushwaha-1225:
Hi, 
Can you assign this issue to me?

I tried implementing this feat and got some good results and I want to proceed further and discuss more about it.

<img src="https://github.com/user-attachments/assets/78317cbb-a467-4882-bd39-09689b758559" width=50%>

chrisbobbe:
You have some other PRs currently open; I'd prefer that you focus on those: https://github.com/zulip/zulip-flutter/pulls/Gaurav-Kushwaha-1225

It looks like you're working on a new revision for #1213, and I've just sent a review for #1152.

PR #1157, opened 2024-12-13
nav: Add an About Zulip button to the main menu

### LINKED ISSUE #1128: "About Zulip" in main menu
The choose-account page has a menu with one item, "About Zulip", taking you to a page with that title. That page in turn has the version of the Zulip app, and an option to see the list of licenses for various open-source software used in the app.

Now that we have (after #1035 / #1076) a new overall navigation scheme, we should also add an "About Zulip" option in the main menu. That seems a more natural place for people to find it.

(It should also remain available in its existing location on the choose-account page. That ensures it's always possible to reach, even in a scenario where you can't get to the home page because the data can't be loaded from the server.)

Reported in [a chat thread](https://chat.zulip.org/#narrow/channel/48-mobile/topic/mobile.20beta.20feedback/near/1998560).

tomlin7:
fix #1128

## Changes
- Added About Zulip button to main menu
- Added localizations and tests for the button

## Screenshots

---

_outdated screenshot (icon changed):_

https://github.com/user-attachments/assets/acda9775-1772-478d-9b5b-ae5c337e86f7

The tests are passing:

image


lib/widgets/home.dart
```diff
@@ -259,6 +260,7 @@ void _showMainMenu(BuildContext context, {
     // TODO(#1094): Users
     const _MyProfileButton(),
     const _SwitchAccountButton(),
+    const _AboutZulipButton(),
```
chrisbobbe, line 263: nit: Let's move this down a few lines, so it comes after the TODOs for buttons that should appear earlier in the list. Probably just above `// TODO(#1095): VersionInfo` is the right place.

```
nav: Add about-zulip button to menu bar
```

Also, a commit-message nit: this isn't really a "menu bar"; you can just call it "the main menu".

assets/l10n/app_en.arb
```diff
@@ -23,6 +23,10 @@
   "@switchAccountButton": {
     "description": "Label for main-menu button leading to the choose-account page."
   },
+  "aboutZulipMenuItem": "About Zulip",
+  "@aboutZulipMenuItem": {
+    "description": "Label for the 'About Zulip' option in the main menu."
+  },
```
chrisbobbe, line 29: We already have one "About Zulip" button. Does its label have an entry in this file? If so, probably good to reuse that instead of adding a new entry. If its description is specific to that other place where the button appears, please adjust it because it won't be correct after reusing it in this new place.

lib/widgets/home.dart
chrisbobbe: Another commit-message nit:

```
nav: Add about-zulip button to menu bar

This is an implementation for issue #1128

This commit includes localizations and tests for the button as well.
```

Just say `Fixes: #1128`; you don't need a whole sentence for that.

Also no need to mention localization and tests. Those are always required, so they don't need a special mention in the commit message.

chrisbobbe (commented):
Thanks! Small comments below.

The other thing this PR needs is screenshots. Please include screenshots of the new button in the PR description.

assets/l10n/app_en.arb
tomlin7: @chrisbobbe `aboutPageTitle` exists already
```json
"aboutPageTitle": "About Zulip",
  "@aboutPageTitle": {
    "description": "Title for About Zulip page."
  },
```
May I adjust the description to "Title and main menu option label for About Zulip page"

chrisbobbe: Since its description is not specific to that other place where the button appears—it's about the page itself, not the button that leads to it—it's still correct after adding a button in the new place. The description doesn't need to change.

tomlin7:
@chrisbobbe I have made the code changes and added screenshots! please check

chrisbobbe:
I see a video; please include a still image because videos are harder to review.

chrisbobbe:
For which icon to use, let's use the same icon that web uses for a similar button:

image

Looks like that's this SVG file: https://github.com/zulip/zulip/blob/74d7f1192/web/shared/icons/info.svg

Please add that file to the project in a separate prep commit, using the instructions in lib/widgets/icons.dart. The commit message should say that we didn't find a suitable icon in the [Icons page of the Figma design](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=544-22131&p=f&m=dev) (with a link to that page), which is why we're using the one from web. The commit should include this link, which points to the SVG file in web: https://github.com/zulip/zulip/blob/74d7f1192/web/shared/icons/info.svg

tomlin7:
@chrisbobbe I have added a still image of the About Zulip button as well. Please check now.

chrisbobbe:
Great, thanks! I see it's not the button from the web app that I mentioned in my comment above: https://github.com/zulip/zulip-flutter/pull/1157#issuecomment-2542525306

Please update the PR with that change. 🙂

tomlin7:
@chrisbobbe Pardon, I saw the second message after sending previous one! 
I have the changes ready, how would I go about making a separate preparatory commit? Do I just push to the same branch with a new commit?

tomlin7:
@chrisbobbe After reading more docs, and having an understanding of what to do, I have done a separate commit because it's an entirely different kind of change compared to previous ones. Also updated screenshots, please review!

tomlin7:
@apoorvapendse thanks for pointing this out! I have fixed the test now, please review it.

chrisbobbe:
Thanks, this is quite close! Please switch the order of the two commits: the commit adding the icon should come first, so the new button uses it from the beginning.

You can use an [interactive rebase](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#reordering-commits) to reorder the commits. If you need help, please ask on `#git help` in the [Zulip development community](https://zulip.com/development-community/).

tomlin7:
@chrisbobbe I have reordered the commits, thanks for the docs. Please check!

chrisbobbe:

test/widgets/home_test.dart
```diff
@@ -215,6 +219,16 @@ void main () {
       check(find.byType(ProfilePage)).findsOne();
       check(find.text(eg.selfUser.fullName)).findsAny();
     });
+
+    testWidgets('_AboutZulipButton', (tester) async {
+      await prepare(tester);
+      await tapOpenMenu(tester);
+
+      await tester.tap(aboutZulipMenuIconFinder);
```
gnprice, line 227: nit:
```suggestion
      await tester.tap(find.byIcon(ZulipIcons.info));
```

The extra complication of being a descendant of BottomSheet is needed for three other items because they share their icons with corresponding buttons in the bottom nav, but isn't here.

lib/widgets/icons.dart
```diff
@@ -78,53 +78,56 @@ abstract final class ZulipIcons {
   /// The Zulip custom icon "inbox".
   static const IconData inbox = IconData(0xf112, fontFamily: "Zulip Icons");
 
+  /// The Zulip custom icon "info".
+  static const IconData info = IconData(0xf113, fontFamily: "Zulip Icons");
```
gnprice, line 82: nit in commit message:

    The icon is taken from Zulip main repo:
    https://github.com/zulip/zulip/blob/74d7f1192/web/shared/icons/info.svg

"Zulip main repo" -> "Zulip web"

(I'd usually call the repo the Zulip server/web repo; here we can be more specific as it's from the web client and not the server)

lib/widgets/home.dart
```diff
@@ -551,6 +553,23 @@ class _SwitchAccountButton extends _MenuButton {
   }
 }
 
+class _AboutZulipButton extends _MenuButton {
```
gnprice, line 556: nit in commit message:

    nav: Add an About Zulip button to the main menu
    Fixes: #1128

blank line between summary and body

gnprice (commented):
Thanks @tomlin7 for the contribution! And thanks @chrisbobbe for the previous reviews.

Generally this looks good. Small comments below.

tomlin7:
@gnprice I have fixed the commit messages and fixed the redundancy as well! please check. 
I'm really not sure if the tests are failing due to the change made.

gnprice:
Thanks for the revision! All LGTM.

The CI failure looks like #1177. I've rebased, and `tools/check` passes for me locally — merging.

(CI will take a few minutes longer because it runs `tools/check --all`, including things like building an APK which are very unlikely to be affected by this change.)

PR #1163, opened 2024-12-14
bump up minimum macOS deployment target to 10.15 from 10.14

sher999:
https://github.com/zulip/zulip-flutter/issues/1116#issue-2726129973

changing the minimum deployment target inside the podfile is working as expected

### DIFF
--- a/macos/Podfile
+++ b/macos/Podfile
@@ -1,4 +1,4 @@
-platform :osx, '10.14'
+platform :osx, '10.15'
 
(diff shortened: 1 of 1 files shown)


PIG208:
Thanks for filing the fix! Did you try to test this after upgrading firebase_core to 3.4.0 before running `tools/upgrade pub`? Pushing a commit upgrading it here will be helpful.

PR #1167, opened 2024-12-17
Support some client settings; add settings page

### LINKED ISSUE #1216: ui: Allow selecting a different dark/light-theme setting than device setting
From [beta feedback](https://chat.zulip.org/#narrow/channel/48-mobile/topic/dark.20theme.20in-app.20setting/near/2018387):
> The beta is great, especially the responsiveness compared to the old app. Just a minor question - is there a way to turn off dark mode, despite it being my OS setting?

Currently, the app UI follows the device setting for dark/light theme.

This issue is for an in-app setting to choose one or the other for the app, independently of the device setting.

An instance of:
- #97.

PIG208:
Work towards #97.

Fixes #1216.

<details>
<summary>screenshots</summary>

| | |
| - | - |
| non-default | default |
</details>


gnprice:
Thanks for taking this on! Looking forward to it.

One first high-level comment: in the branch we merge, let's have at least one of these two settings get added as its own separate commit, complete with a migration. That way we have a nice demonstration of what it'll look like to add each new setting in the future.

lib/model/database.dart
```diff
@@ -115,12 +134,20 @@ class AppDatabase extends _$AppDatabase {
...
+          steps: migrationSteps(
+            from1To2: (m, schema) async {
+              await m.addColumn(schema.accounts, schema.accounts.ackedPushToken);
+            },
+            from2To3: (m, schema) async {
+              await m.createTable(schema.globalSettings);
+            },
+            from3To4: (m, schema) async {
+              await m.addColumn(
+                schema.globalSettings, schema.globalSettings.browserPreference);
```
gnprice, line 147: db: Start generating step by step migration helper
    
    https://drift.simonbinder.eu/migrations/step_by_step/#manual-generation

Hmm, interesting. Yeah, this seems useful — I think this migration to 4 is already an example where this structure lets the new migration get added as just a separate thing, and with the more manual structure we'd have to have logic for it to interact with the migration to 3.

lib/model/schema_versions.dart
```diff
@@ -0,0 +1,238 @@
+import 'package:drift/internal/versioned_schema.dart' as i0;
+import 'package:drift/drift.dart' as i1;
+import 'package:drift/drift.dart'; // ignore_for_file: type=lint,unused_import
+
+// GENERATED BY drift_dev, DO NOT MODIFY.
```
gnprice, line 5: Let's give this a `.g.dart` name, so it's clear in a more uniform way that it's generated.

analysis_options.yaml
```diff
@@ -3,6 +3,9 @@
 include: package:flutter_lints/flutter.yaml
 
 analyzer:
+  exclude:
+    # Generated code from drift can contain unused local variables.
+    - lib/model/database.g.dart
```
gnprice, line 8: This should be reported as a bug in Drift — it's fine for the generated code to do this but then it should contain an ignore comment for it.

lib/model/database.dart
```diff
@@ -39,29 +36,61 @@ class Accounts extends Table {
...
-      ackedPushToken: data.ackedPushToken.present ? data.ackedPushToken.value : ackedPushToken,
-    );
-  }
+/// The visual theme of the app.
+///
+/// See [zulipThemeData] for how themes are determined.
+///
+/// Renaming existing enum values will invalidate the database.
+/// Write a migration if such a change is necessary.
+enum ThemeSetting {
```
gnprice, line 52: Let's put these enum definitions in a new file `lib/model/settings.dart`.

Then that can also be the home of other logic we add in the future for individual settings.

gnprice (commented):
I've skimmed now through more of this code. A few other high-level comments below.

analysis_options.yaml
PIG208: Filed https://github.com/simolus3/drift/issues/3384 for that.

test/widgets/theme_test.dart
```diff
@@ -97,6 +100,27 @@ void main() {
         check(() => a.lerp(b, 0.5)).returnsNormally();
       });
     });
+
+    testWidgets('follow globalSettings.themeSetting if not unset', (tester) async {
```
chrisbobbe, line 104: > follow globalSettings.themeSetting if not unset

I know what you mean by "if not unset", but it feels slightly unsatisfying: we do still want to follow `globalSettings.themeSetting` if it's unset 🙂, but the details are just a bit more complicated for that value.

I also see that this test exercises the "unset" value, but incompletely: with "unset", let's also test that the theme responds when the device setting is changed:

```dart
      tester.platformDispatcher.platformBrightnessTestValue = Brightness.dark;
      await tester.pump();
      check(zulipThemeData(element)).brightness.equals(Brightness.dark);
```

test/widgets/settings_test.dart
```diff
@@ -24,6 +24,30 @@ void main() {
     await tester.pump();
   }
 
+  group('BrowserPreference', () {
+    Finder useExternalBrowserSwitchFinder = find.ancestor(
+      of: find.text('Use external browser'),
+      matching: find.byType(SwitchListTile));
+
+    testWidgets('smoke', (tester) async {
```
chrisbobbe, line 32: Let's also test that the switch is in the right place, and responds correctly, when the setting hasn't been set yet (the part meant by "effective" in `effectiveBrowserPreference`)

chrisbobbe (commented):
This is exciting!

One question: what UI text does zulip-mobile use for the browser setting? Would it make sense to copy that?

analysis_options.yaml
PIG208: I hope that there will be a release with this patch soon (last one was 2 weeks ago). If not we can just ask politely.

chrisbobbe:
Oh, and we have an issue for the theme setting 🙂 so the PR description and relevant commit can be marked as fixing that:

- #1216

lib/widgets/app.dart
```diff
@@ -212,41 +212,48 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+            onGenerateTitle: (BuildContext context) {
+              return ZulipLocalizations.of(context).zulipAppTitle;
+            },
+            localizationsDelegates: ZulipLocalizations.localizationsDelegates,
+            supportedLocales: ZulipLocalizations.supportedLocales,
+            // The context has to be taken from the [Builder] because
+            // [zulipThemeData] requires access to [GlobalStoreWidget] in the tree.
+            // TODO: any way to remove the [Builder], like how we pulled out
+            //   [_handleGenerateInitialRoutes]?
+            theme: zulipThemeData(context),
```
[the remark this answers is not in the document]
PIG208: Chose to name it `builder` to match `Builder.builder` and refactored places that use `GlobalStoreWidget` (one in lib and a handful in tests).

PIG208:
There is also one for browser:

- #1228 

I didn't fix them because they do not implement a final design. Perhaps we can have a new design issue for the settings page.

chrisbobbe:
> I didn't fix them because they do not implement a final design. Perhaps we can have a new design issue for the settings page.

Sure, that sounds good: a new issue for redesigning the settings page as a whole. Then the individual issues, #1216 and #1228, don't need to stay open to track that redesign, and they can be marked as fixed in this PR; what do you think?

PIG208:
Opened #1375.

PIG208:
Forgot to post a message here. This PR is ready for review.

> what UI text does zulip-mobile use for the browser setting? Would it make sense to copy that?

Good point! The string is "Open links with in-app browser"; switched to that instead.

test/model/store_checks.dart
```diff
@@ -24,6 +26,7 @@ extension GlobalStoreChecks on Subject<GlobalStore> {
   Subject<Iterable<int>> get accountIds => has((x) => x.accountIds, 'accountIds');
   Subject<Iterable<({ int accountId, Account account })>> get accountEntries => has((x) => x.accountEntries, 'accountEntries');
   Subject<Account?> getAccount(int id) => has((x) => x.getAccount(id), 'getAccount($id)');
+  Subject<GlobalSettingsData> get globalSettings => has((x) => x.globalSettings, 'globalSettings');
```
gnprice, line 29: nit: put this in the commit where the property appears on GlobalStore

test/widgets/theme_test.dart
```diff
@@ -97,6 +101,48 @@ void main() {
         check(() => a.lerp(b, 0.5)).returnsNormally();
       });
     });
+
+    testWidgets('when globalSettings.themeSetting is unset, follow system setting', (tester) async {
```
gnprice, line 105: nit: this is grouped under DesignVariables but doesn't seem to belong there

PIG208:
Still need to make some updates the browser settings changes. Will push again once I'm done.

PIG208:
Updated the PR. This should be ready for review again!

lib/model/database.dart
```diff
@@ -45,6 +46,20 @@ class Accounts extends Table {
   ];
 }
 
+/// The table of the user's chosen settings independent of account, on this
+/// client.
+///
+/// These apply across all the user's accounts on this client (i.e. on this
+/// install of the app on this device).
+@DataClassName('GlobalSettingsData')
+class GlobalSettings extends Table {
```
gnprice, line 55: nit: order the class definitions matching https://github.com/zulip/zulip-flutter/pull/1167#discussion_r1978462412

lib/widgets/app.dart
```diff
@@ -212,41 +212,45 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
-        // `onGenerateRoute`, or `onUnknownRoute`.  Make it `onGenerateRoute`.
-        // It never actually gets called, though: `onGenerateInitialRoutes`
-        // handles startup, and then we always push whole routes with methods
-        // like [Navigator.push], never mere names as with [Navigator.pushNamed].
-        onGenerateRoute: (_) => null,
-
-        onGenerateInitialRoutes: _handleGenerateInitialRoutes));
+      child: Builder(
+        builder: (context) {
+          return MaterialApp(
```
gnprice, line 218: nit: keep a bit more compact (and less indented):
```suggestion
      child: Builder(builder: (context) {
        return MaterialApp(
```

lib/model/settings.dart
```diff
@@ -0,0 +1,67 @@
+import 'package:flutter/foundation.dart';
+
+import '../generated/l10n/zulip_localizations.dart';
+import 'binding.dart';
+import 'database.dart';
+
+/// The user's choice of visual theme for the app
```
gnprice, line 7: nit: summary line of dartdoc gets a period:
```suggestion
/// The user's choice of visual theme for the app.
```

test/model/store_checks.dart
```diff
@@ -47,3 +51,9 @@ extension PerAccountStoreChecks on Subject<PerAccountStore> {
   Subject<RecentDmConversationsView> get recentDmConversationsView => has((x) => x.recentDmConversationsView, 'recentDmConversationsView');
   Subject<AutocompleteViewManager> get autocompleteViewManager => has((x) => x.autocompleteViewManager, 'autocompleteViewManager');
 }
+
+extension GlobalSettingsDataChecks on Subject<GlobalSettingsData> {
```
gnprice, line 55: nit: keep in logical order, matching the corresponding code

lib/model/store.dart
```diff
@@ -53,8 +53,31 @@ export 'database.dart' show Account, AccountsCompanion, AccountAlreadyExistsExce
...
+    required Iterable<Account> accounts,
+  })
+    : _globalSettings = globalSettings,
+      _accounts = Map.fromEntries(accounts.map((a) => MapEntry(a.id, a)));
+
+  /// A cache of the [GlobalSettingsData] singleton in the underlying data store.
+  GlobalSettingsData get globalSettings => _globalSettings;
+  GlobalSettingsData _globalSettings;
+
+  /// Update the global settings in the store, return the new version.
```
gnprice, line 67: nit: the syntax in the existing `updateAccount` doc is correct:
```suggestion
  /// Update the global settings in the store, returning the new version.
```

Other examples would include "do A, resulting in B" or "do A, arriving at B".

test/model/store_checks.dart
gnprice: nit: match order of the underlying properties

test/model/test_store.dart
```diff
@@ -22,7 +23,15 @@ import '../example_data.dart' as eg;
 ///
 /// See also [TestZulipBinding.globalStore], which provides one of these.
 class TestGlobalStore extends GlobalStore {
-  TestGlobalStore({required super.accounts});
+  TestGlobalStore({required super.globalSettings, required super.accounts})
+    : _globalSettings = globalSettings;
+
+  GlobalSettingsData? _globalSettings;
```
gnprice, line 29:
```suggestion
  GlobalSettingsData _globalSettings;
```

right?

gnprice: Hmm or: what if we just leave this field out? Seems like the base class takes care of it.

And in the case of accounts we don't find we need a corresponding field here.

gnprice (commented):
Thanks for the revision! Here's comments on the same prefix as before, 6 commits:
6230f5d1d do not merge; disable checks
f28eeda38 db: Store GlobalSettings in database
e6f98090b store [nfc]: Update outdated references to loadGlobalStore
3ef35072d test [nfc]: Use eg.globalStore when possible
9897cff67 store: Store global settings on global store
c54b92a43 theme: Track theme through global settings

All small, so I'll also take a look next at the subsequent commits.

test/flutter_checks.dart
```diff
@@ -74,6 +74,18 @@ extension ValueListenableChecks<T> on Subject<ValueListenable<T>> {
   Subject<T> get value => has((c) => c.value, 'value');
 }
 
+extension ThemeDataChecks on Subject<ThemeData> {
```
gnprice, line 77: nit: for logical ordering, put these Material-specific items down among the other Material-specific items near the end of the file

lib/model/store.dart
```diff
@@ -53,8 +53,31 @@ export 'database.dart' show Account, AccountsCompanion, AccountAlreadyExistsExce
...
+  ///
+  /// The global settings must already exist in the store.
+  Future<GlobalSettingsData> updateGlobalSettings(GlobalSettingsCompanion data) async {
```
gnprice, line 70: This should get a couple of tests. See the GlobalStore.updateAccount tests for examples.

test/model/database_test.dart
```diff
@@ -174,6 +206,25 @@ void main() {
...
+          themeSetting: Value(ThemeSetting.light.name)));
+      await before.close();
+
+      final db = AppDatabase(schema.newConnection());
+      await verifier.migrateAndValidate(db, 4);
+      await db.close();
+
+      final after = v4.DatabaseAtV4(schema.newConnection());
+      final globalSettings = await after.select(after.globalSettings).getSingle();
+      check(globalSettings.themeSetting).equals(ThemeSetting.light.name);
```
gnprice, line 224: nit:
```suggestion
      check(globalSettings).themeSetting.equals(ThemeSetting.light.name);
```

(then cascade as usual)

gnprice: … Oh I see, the type here is `v4.GlobalSettingsData`, not something more widely shared. OK, no need to have a checks-extension for it.

test/model/database_test.dart
```diff
@@ -4,10 +4,14 @@ import 'package:drift/native.dart';
 import 'package:drift_dev/api/migrations_native.dart';
 import 'package:test/scaffolding.dart';
 import 'package:zulip/model/database.dart';
+import 'package:zulip/model/settings.dart';
 
 import 'schemas/schema.dart';
 import 'schemas/schema_v1.dart' as v1;
 import 'schemas/schema_v2.dart' as v2;
+import 'store_checks.dart';
+import 'schemas/schema_v3.dart' as v3;
```
gnprice, line 13: nit: keep sorted

lib/widgets/settings.dart
```diff
@@ -0,0 +1,86 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    return Scaffold(
+      appBar: ZulipAppBar(
+        title: Text(zulipLocalizations.settingsPageTitle)),
+      body: Column(children: [
+        const _BrowserPreferenceSetting(),
+        const _ThemeSetting(),
```
gnprice, line 28: Seems like theme should come first — it's going to be a much more popular setting to consider.

gnprice (commented):
OK, and here's comments from reading the next two commits:
91992c245 settings: Add settings page with vanilla Flutter widgets
242103639 db: Store browser preference in database

test/model/test_store.dart
PIG208: Yeah. It should be very similar to `doUpdateAccount`:

```dart
  @override
  Future<void> doUpdateAccount(int accountId, AccountsCompanion data) async {
    // Nothing to do.
  }
```

PIG208:
Updated the PR. This should be ready now.

lib/model/database.g.dart
```diff
@@ -133,7 +384,6 @@ class $AccountsTable extends Accounts with TableInfo<$AccountsTable, Account> {
     if (data.containsKey('id')) {
       context.handle(_idMeta, id.isAcceptableOrUnknown(data['id']!, _idMeta));
     }
-    context.handle(_realmUrlMeta, const VerificationResult.success());
```
gnprice, line 136: nit: generated code should update in tandem with the deps upgrade that causes it

test/model/store_test.dart
```diff
@@ -29,6 +31,36 @@ import 'test_store.dart';
...
+      final globalStore = eg.globalStore();
+      globalStore.addListener(() => notifyCount++);
+      check(notifyCount).equals(0);
+
+      await globalStore.updateGlobalSettings(
+        GlobalSettingsCompanion(themeSetting: Value(ThemeSetting.light)));
+      check(notifyCount).equals(1);
+    });
+
+    // TOOD integration tests with sqlite
```
gnprice, line 61: nit:
```suggestion
    // TODO integration tests with sqlite
```

test/model/store_test.dart
```diff
@@ -29,6 +31,36 @@ import 'test_store.dart';
...
+      check(globalStore).globalSettings.themeSetting.equals(null);
+      check(globalStore).globalSettings.browserPreference.equals(null);
+
+      final result = await globalStore.updateGlobalSettings(
+        GlobalSettingsCompanion(
+          themeSetting: Value(ThemeSetting.dark),
+          browserPreference: Value(BrowserPreference.external)));
+      check(globalStore).globalSettings.themeSetting.equals(ThemeSetting.dark);
+      check(globalStore).globalSettings.browserPreference.equals(
+        BrowserPreference.external);
```
gnprice, line 46: This doesn't need to test each setting as we add it; at that point we'd just be testing the generated code from Drift.

lib/model/settings.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  /// Use the user's default browser app.
+  external,
+}
+
+extension GlobalSettingsHelpers on GlobalSettingsData {
+  UrlLaunchMode get urlLaunchMode {
+    if (browserPreference != null) {
+      return switch (browserPreference!) {
+        BrowserPreference.embedded => UrlLaunchMode.platformDefault,
```
gnprice, line 51: This line doesn't look right, does it. :slightly_smiling_face: If the user has explicitly chosen in-app, let's make it in-app even if that isn't the default on a given platform.

(I think you may have asked me about this the other day, but if I said to keep this I was probably thinking of the spot a few lines below where we're defining our own defaults.)

Or is there a reason why that translation doesn't work?

test/widgets/content_test.dart
```diff
@@ -798,6 +799,19 @@ void main() {
...
+      await testBinding.globalStore.updateGlobalSettings(
+        eg.globalSettings(
+          browserPreference: BrowserPreference.embedded).toCompanion(false));
+      await prepare(tester,
+        '<p><a href="https://example/">hello</a></p>');
+
+      await tapText(tester, find.text('hello'));
+      check(testBinding.takeLaunchUrlCalls())
+        .single.equals((
+          url: Uri.parse('https://example/'), mode: LaunchMode.platformDefault));
```
gnprice, line 812: nit: `.single` compacts nicely onto previous line:
```suggestion
      check(testBinding.takeLaunchUrlCalls()).single.equals((
        url: Uri.parse('https://example/'), mode: LaunchMode.platformDefault));
```

lib/model/settings.dart
```diff
@@ -0,0 +1,64 @@
...
+/// What browser the user has set to use for opening links in messages.
+///
+/// See https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser
+/// for the reasoning behind these options.
+///
+/// Renaming existing enum values will invalidate the database.
+/// Write a migration if such a change is necessary.
+enum BrowserPreference {
+  /// Use the in-app browser.
+  embedded,
```
gnprice, line 41: How about:
```suggestion
  /// Use the in-app browser.
  inApp,
```

Seems like that would best align the name with how we're calling it in other code and in the UI.

lib/widgets/settings.dart
```diff
@@ -0,0 +1,86 @@
...
+
+import '../generated/l10n/zulip_localizations.dart';
+import '../model/binding.dart';
+import '../model/database.dart';
+import '../model/settings.dart';
+import 'app_bar.dart';
+import 'page.dart';
+import 'store.dart';
+
+class SettingsPage extends StatelessWidget {
```
gnprice, line 12: This does not aim to fully implement the wip design for the settings
    page.  We offer it mainly to test settings, so the implementation is
    kept as simple as possible.
    
    WIP design:
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=446-21372&t=BZKpTQPSiBDNxwvB-0

Let's replace "WIP design" with more of a complete sentence — it keeps catching my eye as if the "WIP" is a marker about the state of this commit itself.

gnprice (commented):
Thanks! I've now read the whole thing; comments below.

lib/model/settings.dart
PIG208: For Android and iOS, `platformDefault` for HTTP URLs is already in-app browser view; I remember that we decided to keep `platformDefault` because that setting, "open links with in-app browser", applies to those links to the end-user as expected, but does not make sense for other types of links like mailto.

Testing this, it looks like attempting to open mailto links in-app results in an error:

```
E/flutter (14030): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Invalid argument (url): To use an in-app web view, you must provide an http(s) URL.: Instance of '_SimpleUri'
E/flutter (14030): #0      launchUrl (package:url_launcher/src/url_launcher_uri.dart:49:5)
E/flutter (14030): #1      LiveZulipBinding.launchUrl (package:zulip/model/binding.dart:355:25)
E/flutter (14030): #2      _launchUrl (package:zulip/widgets/content.dart:1446:44)
```

The comment on the rationale of this is in the build method of the widget that controls this setting, but you bringing it up makes me feel that it probably fits better in the dartdoc of this extension method.

PIG208: Or perhaps better, as the dartdoct for the `embedded` (or `inApp`) enum.

PIG208:
This should be ready for review. Made some follow-up comments on https://github.com/zulip/zulip-flutter/pull/1167#discussion_r1988113409.

lib/model/settings.dart
```diff
@@ -0,0 +1,68 @@
...
+///
+/// Renaming existing enum values will invalidate the database.
+/// Write a migration if such a change is necessary.
+enum BrowserPreference {
+  /// Use the in-app browser for HTTP links.
+  ///
+  /// This translates to [UrlLaunchMode.platformDefault], so that non-in-app
+  /// browsers/apps will still be used for other types of links (e.g. mailto)
+  /// on both iOS and Android.
+  inApp,
```
gnprice, line 45: Cool, putting it here on the enum's doc is helpful for explanation (continuing from https://github.com/zulip/zulip-flutter/pull/1167#discussion_r1988201367). Still this translation doesn't seem right.

The [launchUrl](https://pub.dev/documentation/url_launcher/latest/url_launcher/launchUrl.html) doc says:
> The default behavior of [LaunchMode.platformDefault](https://pub.dev/documentation/url_launcher/latest/url_launcher/LaunchMode.html) is up to each platform, and its behavior for a given platform may change over time as new modes are supported, so clients that want a specific mode should request it rather than rely on any currently observed default behavior.

so it's explicitly warning not to make the assumption this is making.

The logic in `launchUrl` that can reject inAppWebView is pretty specific:
```dart
  if ((mode == LaunchMode.inAppWebView ||
          mode == LaunchMode.inAppBrowserView) &&
      !(url.scheme == 'https' || url.scheme == 'http')) {
    throw ArgumentError.value(url, 'url',
        'To use an in-app web view, you must provide an http(s) URL.');
  }
```

so I think we can just have our own bit of logic that checks if the scheme is `http` or `https`, and if not then it replaces `inAppBrowserView` with `platformDefault`.

lib/widgets/settings.dart
gnprice: Cool, this is helpful:

    This does not aim to fully implement the wip design for the settings
    page.  We offer it mainly to test settings, so the implementation is
    kept as simple as possible.
    
    Old design for reference:
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=446-21372&t=BZKpTQPSiBDNxwvB-0

Let's make a couple of adjustments: s/wip design/draft design/, s/Old design/Draft design (old)/. That way (a) it more fully avoids that "wip" keyword :slightly_smiling_face:, (b) it's a bit clearer that the two references are to the same design.

gnprice (commented):
Thanks for the revision!

The first 7/10 commits, up through the settings UI, are virtually ready to merge:
7629e443b fix url; deps: Update drift to 2.26.0
98a1a5ae2 db: Store GlobalSettings in database
2ab1c5ccf store [nfc]: Update outdated references to loadGlobalStore
c84aed02e test [nfc]: Use eg.globalStore when possible
80da3ae7e store: Store global settings on global store
b214218dc theme: Track theme through global settings
ed0afcbed settings: Add settings page with vanilla Flutter widgets

but for the browser preference, there's still a substantive comment below. So let's split that part out to a followup PR.

For those first 7 commits, other than nits below, the first commit message just needs to be updated for merge.

gnprice:
Thanks! Looks good; merging.

Issue #737, opened 2024-06-17
Handle invalid API key on register-queue


gnprice:
When the user has logged in, but then their API key changes, the server requests we make will start failing because they're using the old API key.

Currently the main symptom of this is that you open the app, try to use that account, and the spinner keeps going forever as the app just keeps retrying the registerQueue request. The logs look like:
```
flutter: POST https://chat.zulip.org/api/v1/register
flutter: Error fetching initial snapshot: Invalid API key
Backing off, then will retry…
flutter: … Backoff wait complete, retrying initial fetch.
flutter: POST https://chat.zulip.org/api/v1/register
flutter: Error fetching initial snapshot: Invalid API key
Backing off, then will retry…
flutter: … Backoff wait complete, retrying initial fetch.
flutter: POST https://chat.zulip.org/api/v1/register
flutter: Error fetching initial snapshot: Invalid API key
```

Instead, we should detect that this error is non-transient and won't be resolved by retrying. When that happens:
* Don't retry.
* Show an error dialog, after popping from the navigation stack any screens that were on the affected account. (This and the previous step are the same as at https://github.com/zulip/zulip-flutter/issues/267#issuecomment-2159428504 .)
* Forget about the affected account, akin to #463, so that the user can log in afresh.

  (This differs from what we do for unsupported servers #267; for an unsupported server the only thing to do is to retry after the server has been upgraded, and logging in again won't help, whereas for this situation the only thing to do is to log in again and retrying without that won't help.)

As a followup, we might introduce some sort of streamlined flow that saves the user from entering the server URL again. But that's out of scope for this PR.

Thanks to @PIG208 for reporting this issue; see [chat thread](https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Changing.20primary.20email.20breaks.20existing.20added.20accounts/near/1828866).

gnprice:
The status quo of this is extremely annoying when it comes up — the app just sits at the loading page forever, and all you can do is switch to a different account. For example see [this long-running thread](https://chat.zulip.org/#narrow/channel/48-mobile/topic/0.2E0.2E19.20Flutter.20.3A.20Cant.20connect.20to.20self.20hosted.20instance/near/1922932), where we just today figured out that the cause is probably that the API key (or email / API key pair, anyway) was invalid.

That would be substantially mitigated by #890, telling the user what the problem is. I'm leaving this with a later milestone only because we have that one in the current milestone M5 Launch.

[...]

Issue #890, opened 2024-08-16
Show detailed register-queue feedback


gnprice:
This is like #555 but for the step where we call `registerQueue`, rather than `getEvents`.

On the implementation, some details in chat thread:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/0.2E0.2E19.20Flutter.20.3A.20Cant.20connect.20to.20self.20hosted.20instance/near/1923246

This will let us distinguish between:
* there's some kind of network issue (including a TLS-aka-SSL incompatibility, which can be [subtle to diagnose](https://zulip.readthedocs.io/en/latest/production/ssl-certificates.html#the-android-app-can-t-connect-to-the-server))
* the data from the server is malformed — so either a bug in the client or in the server, and either way something we need to diagnose
* some other issue

and will give the user details they can send us in a screenshot. Particularly in the case of malformed server data, we'll want the stack trace so we can debug.

### Related issues:
* #737

gnprice:
A new issue today is a reminder of the malformed-server-data case:
* #1082

gnprice:
One case where register-queue naturally fails is if the user's API key, or email, changed so that the credentials the app had saved are no longer valid. We [realized today](https://chat.zulip.org/#narrow/channel/48-mobile/topic/0.2E0.2E19.20Flutter.20.3A.20Cant.20connect.20to.20self.20hosted.20instance/near/2004042) that that's probably what was happening in a long-running thread where things mysteriously weren't working for a user — in fact, the same one that originally prompted this issue.

So we should be sure to handle at least that case in the nearer term. The full-service version of handling it would be #737; but if at a minimum we give the user a message clearly saying the problem is their login isn't working, that would give them the information they need to go delete the account from the list and log in again.

gnprice:
With #1183 we've now handled the case where the problem was that the credentials the app had saved are no longer valid; we went ahead and resolved #737 fully.

Given that, I'll move this issue back to a later milestone. We still want to handle the remaining cases, but I think their urgency is lower.

Issue #1054, opened 2024-11-08
Handle auth failures on registerQueue and other requests


gnprice:
When we make a request and it fails because the server rejects the credentials, we currently treat that like any miscellaneous unexpected failure:
* For registerQueue, we retry (with backoff).
* For getEvents, we notify the user and retry (with backoff).
* For most requests (which correspond to the user directly taking some action in the UI), we ignore the failure. The user can always try the action again if they choose.

But it's normal and expected that sometimes the credentials will be rejected — for example it happens if the user resets their API key. (Ideally that in turn would happen on changing their password; that's https://github.com/zulip/zulip/issues/21636 .) We should notice this has happened and do something to help the user get reasonably smoothly to logging in again.

The most important case is for registerQueue, because that's the first request we make (to a previously-logged-in account) when the app starts up, and therefore by far the most likely request to be the first one we make after something makes the old credentials invalid. The next most important is getEvents, because that will typically be the first request we make after opening the app if it was in the background.

Once we have #463 (coming soon), the user will always be able to forget the existing logged-in account in order to try logging in again. Because this is an uncommon situation, I think that's good enough for launch. So milestoning this as post-launch.

PR #1183, opened 2024-12-19
store: Handle invalid API key on register-queue

PIG208:
~~This is near term fix for a user-reported issue:
  https://chat.zulip.org/#narrow/channel/48-mobile/topic/0.2E0.2E19.20Flutter.20.3A.20Cant.20connect.20to.20self.20hosted.20instance/near/2004042~~

~~It is not intended to be the full fix.  With a better UX, we would bring the user back to the choose-account page without them manually doing so.  That's covered by #737 but out-of-scope for this commit.~~

This cherry-picks https://github.com/zulip/zulip-flutter/pull/1235, and partially addresses #890, by handling `INVALID_API_KEY` errors from the `PerAccountStoreWidget`.

Fixes: #737

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -523,6 +523,13 @@
   "@topicValidationErrorMandatoryButEmpty": {
     "description": "Topic validation error when topic is required but was empty."
   },
+  "errorInvalidApiKeyMessage": "Your account at {url} could not be authenticated. Please try logging in again or use another account.",
+  "@errorInvalidApiKeyMessage": {
+    "description": "Error message in the dialog for invalid API key.",
+    "placeholders": {
+      "url": {"type": "String", "example": "http://chat.example.com/"}
+    }
+  },
   "errorInvalidResponse": "The server sent an invalid response",
   "@errorInvalidResponse": {
     "description": "Error message when an API call returned an invalid response."
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -214,8 +214,8 @@
       "url": {"type": "String", "example": "http://example.com/"}
     }
   },
-  "errorLoginCouldNotConnectTitle": "Could not connect",
-  "@errorLoginCouldNotConnectTitle": {
+  "errorCouldNotConnectTitle": "Could not connect",
+  "@errorCouldNotConnectTitle": {
     "description": "Error title when the app could not connect to the server."
   },
   "errorMessageDoesNotSeemToExist": "That message does not seem to exist.",
--- a/lib/log.dart
+++ b/lib/log.dart
@@ -31,7 +31,10 @@ bool debugLog(String message) {
   return true;
 }
 
-typedef ReportErrorCallback = void Function(String? message, {String? details});
+// This should only be used for error reporting functions that allow the error
+// to be cancelled programmatically.  The implementation is expected to handle
+// `null` for the `message` parameter and promptly dismiss the reported errors.
+typedef ReportErrorCancellablyCallback = void Function(String? message, {String? details});
 
 /// Show the user an error message, without requiring them to interact with it.
 ///
@@ -48,7 +51,7 @@ typedef ReportErrorCallback = void Function(String? message, {String? details});
 // This gets set in [ZulipApp].  We need this indirection to keep `lib/log.dart`
 // from importing widget code, because the file is a dependency for the rest of
 // the app.
-ReportErrorCallback reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
+ReportErrorCancellablyCallback reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
 
 void defaultReportErrorToUserBriefly(String? message, {String? details}) {
   // Error dismissing is a no-op to the default handler.
--- a/lib/log.dart
+++ b/lib/log.dart
@@ -54,7 +54,11 @@ typedef ReportErrorCancellablyCallback = void Function(String? message, {String?
 ReportErrorCancellablyCallback reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
 
 void defaultReportErrorToUserBriefly(String? message, {String? details}) {
-  // Error dismissing is a no-op to the default handler.
+  _reportErrorToConsole(message, details);
+}
+
+void _reportErrorToConsole(String? message, String? details) {
+  // Error dismissing is a no-op for the console.
   if (message == null) return;
   // If this callback is still in place, then the app's widget tree
   // hasn't mounted yet even as far as the [Navigator].
--- a/lib/log.dart
+++ b/lib/log.dart
@@ -36,6 +36,8 @@ bool debugLog(String message) {
 // `null` for the `message` parameter and promptly dismiss the reported errors.
 typedef ReportErrorCancellablyCallback = void Function(String? message, {String? details});
 
+typedef ReportErrorCallback = void Function(String title, {String? message});
+
 /// Show the user an error message, without requiring them to interact with it.
 ///
 /// Typically this shows a [SnackBar] containing the message.
@@ -53,8 +55,23 @@ typedef ReportErrorCancellablyCallback = void Function(String? message, {String?
 // the app.
 ReportErrorCancellablyCallback reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
 
+/// Show the user a dismissable error message in a modal popup.
+///
+/// Typically this shows an [AlertDialog] with `title` as the title, `message`
+/// as the body.  If called before the app's widget tree is ready
+/// (see [ZulipApp.ready]), then we give up on showing the message to the user,
+/// and just log the message to the console.
+// This gets set in [ZulipApp].  We need this indirection to keep `lib/log.dart`
+// from importing widget code, because the file is a dependency for the rest of
+// the app.
+ReportErrorCallback reportErrorToUserModally = defaultReportErrorToUserModally;
+
 void defaultReportErrorToUserBriefly(String? message, {String? details}) {
   _reportErrorToConsole(message, details);
+}
+
+void defaultReportErrorToUserModally(String title, {String? message}) {
+  _reportErrorToConsole(title, message);
 }
 
 void _reportErrorToConsole(String? message, String? details) {
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -1220,8 +1220,14 @@ class UpdateMachine {
       if (_disposed) return;
     }
 
-    await store._globalStore._reloadPerAccount(store.accountId);
-    assert(_disposed);
+    try {
+      await store._globalStore._reloadPerAccount(store.accountId);
+    } on AccountNotFoundException {
+      assert(debugLog('… Event queue not replaced; account was logged out.'));
+      return;
+    } finally {
+      assert(_disposed);
+    }
     assert(debugLog('… Event queue replaced.'));
   }
 
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -918,7 +918,7 @@ class UpdateMachine {
         // at 1 kiB (at least on Android), and stack can be longer than that.
         assert(debugLog('Stack:\n$s'));
         assert(debugLog('Backing off, then will retry…'));
-        // TODO tell user if initial-fetch errors persist, or look non-transient
+        // TODO(#890): tell user if initial-fetch errors persist, or look non-transient
         await (backoffMachine ??= BackoffMachine()).wait();
         assert(debugLog('… Backoff wait complete, retrying initial fetch.'));
       }
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -19,6 +19,7 @@ import '../api/backoff.dart';
 import '../api/route/realm.dart';
 import '../log.dart';
 import '../notifications/receive.dart';
+import 'actions.dart';
 import 'autocomplete.dart';
 import 'database.dart';
 import 'emoji.dart';
@@ -149,7 +150,34 @@ abstract class GlobalStore extends ChangeNotifier {
   /// and/or [perAccountSync].
   Future<PerAccountStore> loadPerAccount(int accountId) async {
     assert(_accounts.containsKey(accountId));
-    final store = await doLoadPerAccount(accountId);
+    final PerAccountStore store;
+    try {
+      store = await doLoadPerAccount(accountId);
+    } catch (e) {
+      switch (e) {
+        case HttpException(httpStatus: 401):
+          // The API key is invalid and the store can never be loaded
+          // unless the user retries manually.
+          final account = getAccount(accountId);
+          if (account == null) {
+            // The account was logged out during `await doLoadPerAccount`.
+            // Here, that seems possible only by the user's own action;
+            // the logout can't have been done programmatically.
+            // Even if it were, it would have come with its own UI feedback.
+            // Anyway, skip showing feedback, to not be confusing or repetitive.
+            throw AccountNotFoundException();
+          }
+          final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+          reportErrorToUserModally(
+            zulipLocalizations.errorCouldNotConnectTitle,
+            message: zulipLocalizations.errorInvalidApiKeyMessage(
+              account.realmUrl.toString()));
+          await logOutAccount(this, accountId);
+          throw AccountNotFoundException();
+        default:
+          rethrow;
+      }
+    }
     if (!_accounts.containsKey(accountId)) {
       // TODO(#1354): handle this earlier
       // [removeAccount] was called during [doLoadPerAccount].
@@ -914,12 +942,19 @@ class UpdateMachine {
       try {
         return await registerQueue(connection);
       } catch (e, s) {
-        assert(debugLog('Error fetching initial snapshot: $e'));
-        // Print stack trace in its own log entry; log entries are truncated
-        // at 1 kiB (at least on Android), and stack can be longer than that.
-        assert(debugLog('Stack:\n$s'));
-        assert(debugLog('Backing off, then will retry…'));
         // TODO(#890): tell user if initial-fetch errors persist, or look non-transient
+        switch (e) {
+          case HttpException(httpStatus: 401):
+            // We cannot recover from this error through retrying.
+            // Leave it to [GlobalStore.loadPerAccount].
+            rethrow;
+          default:
+            assert(debugLog('Error fetching initial snapshot: $e'));
+            // Print stack trace in its own log entry; log entries are truncated
+            // at 1 kiB (at least on Android), and stack can be longer than that.
+            assert(debugLog('Stack:\n$s'));
+        }
+        assert(debugLog('Backing off, then will retry…'));
         await (backoffMachine ??= BackoffMachine()).wait();
         assert(debugLog('… Backoff wait complete, retrying initial fetch.'));
       }
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -1177,6 +1177,7 @@ class UpdateMachine {
     store.isLoading = true;
 
     bool isUnexpected;
+    // TODO(#1054): handle auth failure
     switch (error) {
       case ZulipApiException(code: 'BAD_EVENT_QUEUE_ID'):
         assert(debugLog('Lost event queue for $store.  Replacing…'));
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -151,6 +151,7 @@ abstract class GlobalStore extends ChangeNotifier {
     assert(_accounts.containsKey(accountId));
     final store = await doLoadPerAccount(accountId);
     if (!_accounts.containsKey(accountId)) {
+      // TODO(#1354): handle this earlier
       // [removeAccount] was called during [doLoadPerAccount].
       store.dispose();
       throw AccountNotFoundException();
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -223,7 +223,11 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
         theme: themeData,
 
         navigatorKey: ZulipApp.navigatorKey,
-        navigatorObservers: widget.navigatorObservers ?? const [],
+        navigatorObservers: [
+          if (widget.navigatorObservers != null)
+            ...widget.navigatorObservers!,
+          _PreventEmptyStack(),
+        ],
         builder: (BuildContext context, Widget? child) {
           if (!ZulipApp.ready.value) {
             SchedulerBinding.instance.addPostFrameCallback(
@@ -243,6 +247,33 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
         onGenerateRoute: (_) => null,
 
         onGenerateInitialRoutes: _handleGenerateInitialRoutes));
+  }
+}
+
+/// Pushes a route whenever the observed navigator stack becomes empty.
+class _PreventEmptyStack extends NavigatorObserver {
+  void _pushRouteIfEmptyStack() async {
+    final navigator = await ZulipApp.navigator;
+    bool isEmptyStack = true;
+    // TODO: find a better way to inspect the navigator stack
+    navigator.popUntil((route) {
+      isEmptyStack = false;
+      return true; // never actually pops
+    });
+    if (isEmptyStack) {
+      unawaited(navigator.push(
+        MaterialWidgetRoute(page: const ChooseAccountPage())));
+    }
+  }
+
+  @override
+  void didRemove(Route<void> route, Route<void>? previousRoute) async {
+    _pushRouteIfEmptyStack();
+  }
+
+  @override
+  void didPop(Route<void> route, Route<void>? previousRoute) async {
+    _pushRouteIfEmptyStack();
   }
 }
 
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -85,6 +85,7 @@ class ZulipApp extends StatefulWidget {
   static void debugReset() {
     _snackBarCount = 0;
     reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
+    reportErrorToUserModally = defaultReportErrorToUserModally;
     _ready.dispose();
     _ready = ValueNotifier(false);
   }
@@ -128,10 +129,21 @@ class ZulipApp extends StatefulWidget {
     newSnackBar.closed.whenComplete(() => _snackBarCount--);
   }
 
+  /// The callback we normally use as [reportErrorToUserModally].
+  static void _reportErrorToUserModally(String title, {String? message}) {
+    assert(_ready.value);
+
+    showErrorDialog(
+      context: navigatorKey.currentContext!,
+      title: title,
+      message: message);
+  }
+
   void _declareReady() {
     assert(navigatorKey.currentContext != null);
     _ready.value = true;
     reportErrorToUserBriefly = _reportErrorToUserBriefly;
+    reportErrorToUserModally = _reportErrorToUserModally;
   }
 
   @override
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -39,6 +39,18 @@ ZulipApiException apiBadRequest({
     data: {}, message: message);
 }
 
+/// The error for the "events" route when the target event queue has been
+/// garbage collected.
+///
+/// https://zulip.com/api/get-events#bad_event_queue_id-errors
+ZulipApiException apiExceptionBadEventQueueId({
+  String queueId = 'fb67bf8a-c031-47cc-84cf-ed80accacda8',
+}) {
+  return ZulipApiException(
+    routeName: 'events', httpStatus: 400, code: 'BAD_EVENT_QUEUE_ID',
+    data: {'queue_id': queueId}, message: 'Bad event queue ID: $queueId');
+}
+
 /// The error the server gives when the client's credentials
 /// (API key together with email and realm URL) are no longer valid.
 ///
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -879,6 +879,7 @@ ChannelUpdateEvent channelUpdateEvent(
 TestGlobalStore globalStore({List<Account> accounts = const []}) {
   return TestGlobalStore(accounts: accounts);
 }
+const _globalStore = globalStore;
 
 InitialSnapshot initialSnapshot({
   String? queueId,
@@ -949,10 +950,14 @@ InitialSnapshot initialSnapshot({
 }
 const _initialSnapshot = initialSnapshot;
 
-PerAccountStore store({Account? account, InitialSnapshot? initialSnapshot}) {
+PerAccountStore store({
+  GlobalStore? globalStore,
+  Account? account,
+  InitialSnapshot? initialSnapshot,
+}) {
   final effectiveAccount = account ?? selfAccount;
   return PerAccountStore.fromInitialSnapshot(
-    globalStore: globalStore(accounts: [effectiveAccount]),
+    globalStore: globalStore ?? _globalStore(accounts: [effectiveAccount]),
     accountId: effectiveAccount.id,
     initialSnapshot: initialSnapshot ?? _initialSnapshot(),
   );
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -757,11 +757,8 @@ void main() {
     }
 
     void prepareExpiredEventQueue() {
-      connection.prepare(httpStatus: 400, json: {
-        'result': 'error', 'code': 'BAD_EVENT_QUEUE_ID',
-        'queue_id': updateMachine.queueId,
-        'msg': 'Bad event queue ID: ${updateMachine.queueId}',
-      });
+      connection.prepare(apiException: eg.apiExceptionBadEventQueueId(
+        queueId: updateMachine.queueId));
     }
 
     Future<void> prepareHandleEventError() async {
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -13,6 +13,7 @@ import 'package:zulip/api/route/events.dart';
 import 'package:zulip/api/route/messages.dart';
 import 'package:zulip/api/route/realm.dart';
 import 'package:zulip/log.dart';
+import 'package:zulip/model/actions.dart';
 import 'package:zulip/model/store.dart';
 import 'package:zulip/notifications/receive.dart';
 
@@ -967,6 +968,65 @@ void main() {
         ));
       });
     });
+  });
+
+  group('UpdateMachine.poll reload failure', () {
+    late LoadingTestGlobalStore globalStore;
+
+    List<Completer<PerAccountStore>> completers() =>
+      globalStore.completers[eg.selfAccount.id]!;
+
+    Future<void> prepareReload(FakeAsync async) async {
+      globalStore = LoadingTestGlobalStore(accounts: [eg.selfAccount]);
+      final future = globalStore.perAccount(eg.selfAccount.id);
+      final store = eg.store(globalStore: globalStore, account: eg.selfAccount);
+      completers().single.complete(store);
+      await future;
+      completers().clear();
+      final updateMachine = globalStore.updateMachines[eg.selfAccount.id] =
+        UpdateMachine.fromInitialSnapshot(
+          store: store, initialSnapshot: eg.initialSnapshot());
+      updateMachine.debugPauseLoop();
+      updateMachine.poll();
+
+      (store.connection as FakeApiConnection).prepare(
+        apiException: eg.apiExceptionBadEventQueueId());
+      updateMachine.debugAdvanceLoop();
+      async.elapse(Duration.zero);
+      check(store).isLoading.isTrue();
+    }
+
+    void checkReloadFailure({
+      required Future<void> Function() completeLoading,
+    }) {
+      awaitFakeAsync((async) async {
+        await prepareReload(async);
+        check(completers()).single.isCompleted.isFalse();
+
+        await completeLoading();
+        check(completers()).single.isCompleted.isTrue();
+        check(globalStore.takeDoRemoveAccountCalls()).single.equals(eg.selfAccount.id);
+
+        async.elapse(TestGlobalStore.removeAccountDuration);
+        check(globalStore.perAccountSync(eg.selfAccount.id)).isNull();
+
+        async.flushTimers();
+        // Reload never succeeds and there are no unhandled errors.
+        check(globalStore.perAccountSync(eg.selfAccount.id)).isNull();
+      });
+    }
+
+    Future<void> logOutAndCompleteWithNewStore() async {
+      // [PerAccountStore.fromInitialSnapshot] requires the account
+      // to be in the global store when called; do so before logging out.
+      final newStore = eg.store(globalStore: globalStore, account: eg.selfAccount);
+      await logOutAccount(globalStore, eg.selfAccount.id);
+      completers().single.complete(newStore);
+    }
+
+    test('user logged out before new store is loaded', () => awaitFakeAsync((async) async {
+      checkReloadFailure(completeLoading: logOutAndCompleteWithNewStore);
+    }));
   });
 
   group('UpdateMachine.registerNotificationToken', () {
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -121,6 +121,29 @@ void main() {
     check(completers(1)).length.equals(1);
   });
 
+  test('GlobalStore.perAccount loading fails with HTTP status code 401', () => awaitFakeAsync((async) async {
+    final globalStore = LoadingTestGlobalStore(accounts: [eg.selfAccount]);
+    final future = globalStore.perAccount(eg.selfAccount.id);
+
+    globalStore.completers[eg.selfAccount.id]!
+      .single.completeError(eg.apiExceptionUnauthorized());
+    await check(future).throws<AccountNotFoundException>();
+  }));
+
+  test('GlobalStore.perAccount account is logged out while loading; then fails with HTTP status code 401', () => awaitFakeAsync((async) async {
+    final globalStore = LoadingTestGlobalStore(accounts: [eg.selfAccount]);
+    final future = globalStore.perAccount(eg.selfAccount.id);
+
+    await logOutAccount(globalStore, eg.selfAccount.id);
+    check(globalStore.takeDoRemoveAccountCalls())
+      .single.equals(eg.selfAccount.id);
+
+    globalStore.completers[eg.selfAccount.id]!
+      .single.completeError(eg.apiExceptionUnauthorized());
+    await check(future).throws<AccountNotFoundException>();
+    check(globalStore.takeDoRemoveAccountCalls()).isEmpty();
+  }));
+
   // TODO test insertAccount
 
   group('GlobalStore.updateAccount', () {
@@ -997,7 +1020,7 @@ void main() {
     }
 
     void checkReloadFailure({
-      required Future<void> Function() completeLoading,
+      required FutureOr<void> Function() completeLoading,
     }) {
       awaitFakeAsync((async) async {
         await prepareReload(async);
@@ -1026,6 +1049,14 @@ void main() {
 
     test('user logged out before new store is loaded', () => awaitFakeAsync((async) async {
       checkReloadFailure(completeLoading: logOutAndCompleteWithNewStore);
+    }));
+
+    void completeWithApiExceptionUnauthorized() {
+      completers().single.completeError(eg.apiExceptionUnauthorized());
+    }
+
+    test('new store is not loaded, gets HTTP 401 error instead', () => awaitFakeAsync((async) async {
+      checkReloadFailure(completeLoading: completeWithApiExceptionUnauthorized);
     }));
   });
 
--- a/test/model/test_store.dart
+++ b/test/model/test_store.dart
@@ -129,6 +129,7 @@ class TestGlobalStore extends GlobalStore {
 
   static const Duration removeAccountDuration = Duration(milliseconds: 1);
   Duration? loadPerAccountDuration;
+  Object? loadPerAccountException;
 
   /// Consume the log of calls made to [doRemoveAccount].
   List<int> takeDoRemoveAccountCalls() {
@@ -149,6 +150,9 @@ class TestGlobalStore extends GlobalStore {
   Future<PerAccountStore> doLoadPerAccount(int accountId) async {
     if (loadPerAccountDuration != null) {
       await Future<void>.delayed(loadPerAccountDuration!);
+    }
+    if (loadPerAccountException != null) {
+      throw loadPerAccountException!;
     }
     final initialSnapshot = _initialSnapshots[accountId]!;
     final store = PerAccountStore.fromInitialSnapshot(
--- a/test/stdlib_checks.dart
+++ b/test/stdlib_checks.dart
@@ -5,6 +5,7 @@
 /// part of the Dart standard library.
 library;
 
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:checks/checks.dart';
@@ -72,6 +73,10 @@ Object? deepToJson(Object? object) {
       return (null, false);
   }
   return (result, true);
+}
+
+extension CompleterChecks<T> on Subject<Completer<T>> {
+  Subject<bool> get isCompleted => has((x) => x.isCompleted, 'isCompleted');
 }
 
 extension JsonChecks on Subject<Object?> {
--- a/test/widgets/app_test.dart
+++ b/test/widgets/app_test.dart
@@ -245,7 +245,9 @@ void main() {
       check(ZulipApp.scaffoldMessenger).isNotNull();
       check(ZulipApp.ready).value.isTrue();
     });
+  });
 
+  group('error reporting', () {
     Finder findSnackBarByText(String text) => find.descendant(
       of: find.byType(SnackBar),
       matching: find.text(text));
@@ -307,7 +309,7 @@ void main() {
       check(findSnackBarByText(message).evaluate()).single;
     }
 
-    testWidgets('reportErrorToUser dismissing SnackBar', (tester) async {
+    testWidgets('reportErrorToUserBriefly dismissing SnackBar', (tester) async {
       const message = 'test error message';
       const details = 'error details';
       await prepareSnackBarWithDetails(tester, message, details);
--- a/test/widgets/app_test.dart
+++ b/test/widgets/app_test.dart
@@ -363,5 +363,24 @@ void main() {
       await tester.pumpAndSettle();
       check(findSnackBarByText('unrelated').evaluate()).single;
     });
+
+    testWidgets('reportErrorToUserModally', (tester) async {
+      addTearDown(testBinding.reset);
+      await tester.pumpWidget(const ZulipApp());
+      const title = 'test title';
+      const message = 'test message';
+
+      // Prior to app startup, reportErrorToUserModally only logs.
+      reportErrorToUserModally(title, message: message);
+      check(ZulipApp.ready).value.isFalse();
+      await tester.pump();
+      checkNoErrorDialog(tester);
+
+      check(ZulipApp.ready).value.isTrue();
+      // After app startup, reportErrorToUserModally displays an [AlertDialog].
+      reportErrorToUserModally(title, message: message);
+      await tester.pump();
+      checkErrorDialog(tester, expectedTitle: title, expectedMessage: message);
+    });
   });
 }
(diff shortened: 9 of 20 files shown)


gnprice:
Thanks!

Would you post a screenshot of what the error looks like? Then we can also wordsmith the error message.

PIG208:
Sure!

<details>
<summary>Details</summary>

image

</details>

gnprice:
Cool, thanks.

I think "Invalid API key" is probably too technical for this context. @alya do we have an appropriate message handy from elsewhere in the product?

PIG208:
Perhaps we can reuse the

PIG208:
Perhaps we can rename and reuse the title string we have for login failure:

```json
  "errorLoginCouldNotConnectTitle": "Could not connect",
```

This sort of feels like an instance of #741 the way we currently handle it, but this message will not be needed after #737.

PIG208:
Updated the PR to implement Chris' suggestion [here](https://chat.zulip.org/#narrow/channel/48-mobile/topic/.23F1183.20Invalid.20API.20key.20response/near/2018994). We now handle invalid API key by calling `logOutAccount`, which removes the routes associated with the affected account, pushing a choose-account page route onto the stack if it becomes empty, and showing an error dialog.

assets/l10n/app_en.arb
```diff
@@ -186,8 +186,8 @@
       "url": {"type": "String", "example": "http://example.com/"}
     }
   },
-  "errorLoginCouldNotConnectTitle": "Could not connect",
-  "@errorLoginCouldNotConnectTitle": {
+  "errorCouldNotConnectTitle": "Could not connect",
+  "@errorCouldNotConnectTitle": {
```
chrisbobbe, line 190:
```
l10n [nfc]: Use a generalize name for errorCouldNotConnectTitle
```

commit-message nit: "generalized"

lib/widgets/home.dart
```diff
@@ -180,9 +185,15 @@ class _LoadingPlaceholderPageState extends State<_LoadingPlaceholderPage> {
...
-      .getAccount(widget.accountId)!.realmUrl;
+    final account = GlobalStoreWidget.of(context).getAccount(widget.accountId);
+
+    if (account == null) {
+      // We should only reach this state very briefly.
+      // See [_LoadingPlaceholderPage.accountId].
+      return Scaffold(
+        appBar: AppBar(),
+        body: const SizedBox.shrink());
+    }
```
chrisbobbe, line 196: This feels like it duplicates more details than it needs to, like the `appBar` param and the whole set of `Scaffold` params that aren't passed. How about applying the condition more precisely to the things that are supposed to be different?

lib/log.dart
```diff
@@ -50,6 +51,17 @@ typedef ReportErrorCancellablyCallback = void Function(String? message, {String?
...
+/// Show the user a dismissable error message in a modal popup.
+///
+/// Typically this shows a [AlertDialog] containing the message.
+/// If called before the app's widget tree is ready (see [ZulipApp.ready]),
+/// then we give up on showing the message to the user,
+/// and just log the message to the console.
+// This gets set in [ZulipApp].  We need this indirection to keep `lib/log.dart`
+// from importing widget code, because the file is a dependency for the rest of
+// the app.
+ReportErrorCallback reportErrorToUserModally = defaultReportErrorToUserBriefly;
```
chrisbobbe, line 63: Setting `reportErrorToUserModally` to `defaultReportErrorToUserBriefly` looks odd. How about renaming `defaultReportErrorToUserBriefly` to something that seems like an appropriate value for both `reportErrorToUserBriefly` and `reportErrorToUserModally`?

Maybe `dumpErrorToConsole`? That name makes it sound like a potential problem if it actually gets called (errors should be reported, not dropped on the floor 🙂)—but that's actually appropriate in this case.

test/widgets/app_test.dart
```diff
@@ -361,5 +362,24 @@ void main() {
...
+      addTearDown(testBinding.reset);
+      await tester.pumpWidget(const ZulipApp());
+      const message = 'test error message';
+      const details = 'details';
+
+      // Prior to app startup, reportErrorToUserModally only logs.
+      reportErrorToUserModally(message, details: details);
+      check(ZulipApp.ready).value.isFalse();
+      await tester.pump();
+      check(find.byType(AlertDialog)).findsNothing();
```
chrisbobbe, line 376: I have a commit

```
test [nfc]: Factor out checkNoErrorDialog helper
```

in my revision for #1239; how about cherry-picking that as a prep commit and using it here? (See reasoning in the commit)

chrisbobbe (commented):
Thanks! Here's a review of the first five commits:

06c554338 l10n [nfc]: Use a generalize name for errorCouldNotConnectTitle
d27b46cc2 log [nfc]: Rename ReportErrorCallback to ReportErrorCancellablyCallback
c5fc2d93e example_data: Start generating account id
652c3328e home: Stop assuming account existence from loading page
ab9148088 log: Add reportErrorModally

which leaves the last three that I haven't read yet:

58fd4b160 store: Handle invalid API key on register-queue
7a632b2c7 display test [nfc]: Make openNotification public
2e0bcb944 actions test: Use a realistic example

For the last commit, it's the one with the `Fixes: ` line but it doesn't actually change any app code, it only touches test code. Is that intentional? I'd like to understand if/how these test changes are related to your app-code changes for #737.

lib/log.dart
PIG208: Good idea! Renaming to `reportErrorToConsole` because we don't usually dump the error object.

PIG208:
The "Fixes" comment gets added to the last commit because it resolves a TODO item for the issue.

lib/widgets/store.dart
```diff
@@ -245,11 +261,28 @@ class _PerAccountStoreWidgetState extends State<PerAccountStoreWidget> {
...
+              // The API key is invalid and the store can never be loaded
+              // unless the user retries manually.
+              if (!mounted) return;
+              final zulipLocalizations = ZulipLocalizations.of(context);
+              reportErrorToUserModally(
+                zulipLocalizations.errorCouldNotConnectTitle,
+                details: zulipLocalizations.errorInvalidApiKeyMessage(
+                  globalStore.getAccount(widget.accountId)!.realmUrl.toString()));
+              unawaited(logOutAccount(context, widget.accountId));
+              return;
```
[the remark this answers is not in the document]
chrisbobbe: We also need tests to check that the account is in fact logged out / removed, and that some subtle related things are dealt with, like `_perAccountStores` and `_perAccountStoresLoading`, calling `PerAccountStore.dispose`, and notifying listeners. How about using `TestGlobalStore.takeDoRemoveAccountCalls` for this, because `GlobalStore.removeAccount` already has tests for those things.

test/widgets/store_test.dart
```diff
@@ -113,6 +119,45 @@ void main() {
...
+    await tester.pump(); // start to load account
+    check(find.byType(CircularProgressIndicator)).findsOne();
+    check(find.byType(ChooseAccountPage)).findsNothing();
+    check(find.byType(BackButton)).findsNothing();
+
+    await tester.pump(kTryAnotherAccountWaitPeriod);
+    await tester.tap(find.text('Try another account'));
+    await tester.pump(); // tap the button
+    await tester.pump(const Duration(milliseconds: 250)); // wait for animation
+    check(find.byType(CircularProgressIndicator)).findsOne();
```
chrisbobbe, line 143: I don't yet understand this `CircularProgressIndicator` check.
- Does the test expect it to be on a specific page? (meaning the check stops being as useful if another page adds a `CircularProgressIndicator`)
- If we're expecting it on a non-active/foregrounded page, do we need to use `skipOffstage: false`?

test/widgets/actions_test.dart
```diff
@@ -168,57 +169,46 @@ void main() {
...
       await tester.pumpWidget(ZulipApp(navigatorObservers: [testNavObserver]));
       await tester.pump();
-      final navigator = await ZulipApp.navigator;
-      navigator.popUntil((_) => false); // clear starting routes
-      await tester.pumpAndSettle();
-
-      final pushedRoutes = <Route<dynamic>>[];
-      testNavObserver.onPushed = (route, prevRoute) => pushedRoutes.add(route);
-      // TODO(#737): switch to a realistic setup:
-      //   https://github.com/zulip/zulip-flutter/pull/1076#discussion_r1874124363
```
chrisbobbe, line 186: Reading that linked discussion and links from there, I'm not finding an explanation for why the TODO involves #737 at all. What Greg meant by not "realistic setup" is just that "These pages don't exist in real life", meaning two `MaterialAccountWidgetRoute`s with page `InboxPageBody`. Routes like that don't exist in the app because, since the bottom-tabs update, the Inbox page is never alone on its own route, it's always part of the route returned by `HomePage.buildRoute`.

The direct fix is to use realistic routes for `account1Route` and `account2Route`, e.g. by using `MessageListPage.buildRoute` instead, making follow-on changes to `findAccount{1,2}PageContent`, and renaming the `makeUnreadTopicInInbox` helper. Let's have a commit that just does that and removes the TODO, explaining in the commit message that it's not related to 737 after all.

If you'd like to go further in a follow-up commit (which I think is not a high priority), I'd have some feedback on that direction:
- Take care to include `skipOffstage: false` where necessary when checking for something on a route that's not the top of the stack.
- I intentionally put `account1Route` (the one to be removed) under `account2Route` in the stack, to guard against buggy implementations that might use something like `NavigatorState.popUntil` (which starts from the top and stops at the first route that shouldn't be removed, even if there are ones that should be removed under it). This would be a nice feature to keep, but I guess not essential.

chrisbobbe (changes requested):
Thanks! Comments below.

lib/widgets/store.dart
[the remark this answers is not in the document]
chrisbobbe: For example (just for the first-`/register` case):

```dart
  test('GlobalStore.perAccount on INVALID_API_KEY', () => awaitFakeAsync((async) async {

    await testBinding.globalStore.insertAccount(eg.selfAccount.toCompanion(false));
    testBinding.globalStore.loadPerAccountException = ZulipApiException(
      routeName: '/register', code: 'INVALID_API_KEY', httpStatus: 400,
      data: {}, message: '');
    await check(testBinding.globalStore.perAccount(eg.selfAccount.id))
      .throws<ZulipApiException>();

    // fails
    check(testBinding.globalStore.takeDoRemoveAccountCalls())
      .single.equals(eg.selfAccount.id);
  }));
```

lib/widgets/store.dart
```diff
@@ -228,7 +235,16 @@ class _PerAccountStoreWidgetState extends State<PerAccountStoreWidget> {
...
-          (_) => Navigator.of(context).removeRoute(widget.routeToRemoveOnLogout!));
+          (_) {
+            final navigator = Navigator.of(context);
+            navigator.removeRoute(widget.routeToRemoveOnLogout!);
+            if (!navigator.canPop()) {
+              // This ensures that the navigator stack is non-empty after the
+              // removal of the route.
+              unawaited(navigator.push(
+                MaterialWidgetRoute(page: const ChooseAccountPage())));
+            }
```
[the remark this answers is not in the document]
PIG208: Yeah, a check for [both `isFirst` and `isCurrent`](https://api.flutter.dev/flutter/widgets/Route/isFirst.html) would be more appropriate.

lib/widgets/store.dart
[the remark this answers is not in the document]
PIG208: I think the right place for calling `logOutAccount` and `reportErrorModally` should be `loadPerAccount`. We throw an `AccountNotFoundException` after processing the error, and make sure that all call-sites handles that correctly. `_PerAccountStoreWidgetState` already expects the exception.

For the other call-site, we need to handle it at least for one of `_reloadPerAccount`, `UpdateMachine._handlePollError`, or `UpdateMachine.poll`.

It doesn't seem right to catch it from `_reloadPerAccount` without rethrowing or at least indicating that the reload was a failure; `_handlePollError`'s caller assume that the store is disposed after calling it (because of a reloaded `PerAccountStore`), so we should still either rethrow the error or indicate the failure reloading the store. So we probably should just catch it from `poll`, within another `catch` block.

---

Because `logOutAccount` is no longer purely a user action, we might want to revisit https://github.com/zulip/zulip-flutter/pull/1010#discussion_r1817457287 and figure out a new home to this. While `GlobalStore` is logically a good fit, I agree that `lib/model/store.dart` has become too crowded for us to add more to it. Maybe splitting `store.dart` would be a helpful prep/follow-up change.

test/widgets/store_test.dart
PIG208: It's looking for the route that is not on the top of the stack. I'm switching to a rewrite of this using `NavigatorObserver`'s, so that we can check the navigator stack more explicitly.

lib/widgets/store.dart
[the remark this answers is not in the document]
PIG208: We now implement a `NavigatorObserver` based solution. It implements `didPop` and `didRemove` to push a choose-account page route proactively if it observes that the navigator stack is becoming empty.

test/widgets/actions_test.dart
PIG208: Thanks for the detailed explanation! This has become #1257, but probably the refactor is not going to be a blocker for #737 so we can hold off reviewing that for now.

PIG208:
The PR has been updated to use a different approach maintaining the navigator stack. We are reusing a code path that throws `AccountNotFoundException` as a result of logging out the account. To ensure that the removal of a route does not end up with an empty navigator stack, we add a navigator observer to `ZulipApp` that pushes a choose-account page route when it becomes empty.

image

lib/model/store.dart
```diff
@@ -19,6 +19,7 @@ import '../api/backoff.dart';
 import '../api/route/realm.dart';
 import '../log.dart';
 import '../notifications/receive.dart';
+import '../widgets/actions.dart';
```
gnprice, line 22: Prompted to look at this by https://github.com/zulip/zulip-flutter/pull/1257#discussion_r1917744040 . As I said there just now, let's avoid importing widgets/ from models/.

As the lib/widgets/actions.dart file's dartdoc says:
```dart
/// Methods that act through the Zulip API and show feedback in the UI.
///
/// The methods in this file can be thought of as higher-level wrappers for
/// some of the Zulip API endpoint binding methods in `lib/api/route/`.
/// But they don't belong in `lib/api/`, because they also interact with widgets
/// in order to present success or error feedback to the user through the UI.
```

the point of that file being in widgets/ is that its functions use widgets to give the user feedback. Since `logOutAccount` and its callee `unregisterToken` don't do that, they can move elsewhere.

An easy solution is to move those two functions to a file `lib/model/actions.dart`, i.e. under `model/`.

I think ultimately the right organization is probably that these belong as methods on GlobalStore, after all the more local-storage-focused methods that also act on accounts.

(Then in the future, as a matter of code organization, all those accounts-oriented methods' implementations will live on an AccountStore or AccountsStore, but the way one calls them will still be via GlobalStore — much like the existing ChannelStore and MessageStore etc. help keep PerAccountStore organized.)

PIG208: Echoing https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1904045285, I assume the reason that we don't move it to `GlobalStore` right away is to prevent cluttering the class before we carry out the code reorganization you mentioned. The plan of moving it to `lib/model/actions.dart` sounds good to me (along with the `BuildContext` parameter to `GlobalStore` refactor).

PIG208:
Updated the PR to resolve conflicts and move `logOutAccount` and `unregisterToken` to `lib/model/actions`. The relevant tests weren't moved. We can if we want. Those require a bit more care to be refactored from `testWidgets` tests to `test` tests, but it might not be worth it at the moment.

lib/model/store.dart
```diff
@@ -912,7 +931,13 @@ class UpdateMachine {
...
         assert(debugLog('Stack:\n$s'));
         assert(debugLog('Backing off, then will retry…'));
-        // TODO tell user if initial-fetch errors persist, or look non-transient
+        // TODO(#890): tell user if initial-fetch errors persist, or look non-transient
+        switch (e) {
+          case ZulipApiException(code: 'INVALID_API_KEY'):
+            // We cannot recover from this error through retrying.
+            // Leave it to [GlobalStore.loadPerAccount].
+            rethrow;
+        }
```
chrisbobbe, line 940: The log line 'Backing off, then will retry…' becomes inaccurate in some cases; can you adjust so that doesn't happen?

lib/model/store.dart
```diff
@@ -912,7 +931,13 @@ class UpdateMachine {
         // at 1 kiB (at least on Android), and stack can be longer than that.
         assert(debugLog('Stack:\n$s'));
         assert(debugLog('Backing off, then will retry…'));
-        // TODO tell user if initial-fetch errors persist, or look non-transient
+        // TODO(#890): tell user if initial-fetch errors persist, or look non-transient
+        switch (e) {
+          case ZulipApiException(code: 'INVALID_API_KEY'):
```
chrisbobbe, line 936: In manual testing on a dev server, where I changed an account's API key, the code I got was actually 'UNAUTHORIZED', not 'INVALID_API_KEY'. Have you seen 'INVALID_API_KEY' in manual testing? If you haven't done manual testing of the new functionality, please do that. 🙂

chrisbobbe (commented):
Thanks! Comments below, including one where I got something unexpected in manual testing.

lib/model/store.dart
```diff
@@ -1044,7 +1069,13 @@ class UpdateMachine {
...
     } catch (e) {
       if (_disposed) return;
-      await _handlePollError(e);
+      try {
+        await _handlePollError(e);
+      } on AccountNotFoundException {
+        // Cannot recover by replacing the store because the account
+        // was logged out.
+        return;
+      }
```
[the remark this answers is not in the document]
PIG208: Cannot `assert(_disposed)` here because the reload might not have happened if there was an error (likely during `_reloadPerAccount`) that prevents the store from being disposed.

I chose to not handle it in `_handlePollError` because of the `assert(_disposed)` that follows the `_handlePollError` call. This way we also maintain that `_handlePollError` has the store replaced. An alternative would be returning a `bool` from `_handlePollError`, indicating if the store gets disposed.

lib/model/store.dart
PIG208: Oh, I see! Didn't realize that my prior testing was limited to throwing `ZulipApiException` from certain places, but not with a dev server. Will do that. Thanks for pointing it out! :smiling_face_with_tear:

lib/model/store.dart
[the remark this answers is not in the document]
PIG208: Having tested this further, there are two cases `_handlePollError` should be aware of:

- The `_reloadPerAccount` call (`POST /register`) can fail due to an invalid API key. Handling this is necessary because otherwise it is a runtime error.
- The `Object error` parameter, i.e. the error from polling (`GET /events`) can be a `ZulipApiException` with HTTP status code 401, indicating the invalid API key error.
  Without handling it, it is marked as unexpected in the `default` branch under the `switch`, but we are still able to recover as long as we have error handling for the `_reloadPerAccount` call.

To handle the second case, we just need to acknowledge that such an error is expected and will be handled later.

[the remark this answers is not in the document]
PIG208: Let's defer the handling of `GET /events` failure to #1054.

PIG208:
This is now stacked on top of #1340, which fixes a related issue that you may run into when invalidating the API key while on a message list page.

Also made some other changes apart from addressing the review comments on commit messages and code organization.

lib/model/store.dart
```diff
@@ -149,7 +150,29 @@ abstract class GlobalStore extends ChangeNotifier {
...
+          final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+          reportErrorToUserModally(
+            zulipLocalizations.errorCouldNotConnectTitle,
+            details: zulipLocalizations.errorInvalidApiKeyMessage(
+              realmUrl.toString()));
+          // The account could be logged out from the choose-account page.
+          if (_accounts.containsKey(accountId)) {
+            await logOutAccount(this, accountId);
+          }
+          throw AccountNotFoundException();
```
chrisbobbe, line 171: Ooh actually: I think it would make sense to also make the  `reportErrorToUserModally` conditional on whether the account is still present, right?

Then, we don't give "could-not-authenticate" feedback about an account that has already been logged out. That feedback could be confusing or repetitive: confusing, if the account was logged out by the user's own action; repetitive, if the app chose to log out the account, because the app gives UI feedback when it does that.

Then also, no need to store `realmUrl` in a variable:

```dart
        case HttpException(httpStatus: 401):
          // The API key is invalid and the store can never be loaded
          // unless the user retries manually.
          final account = getAccount(accountId);
          if (account == null) {
            // The account was logged out during `await doLoadPerAccount`.
            // Here, that seems possible only by the user's own action;
            // the logout can't have been done programmatically.
            // Even if it were, it would have come with its own UI feedback.
            // Anyway, skip showing feedback, to not be confusing or repetitive.
            throw AccountNotFoundException();
          }
          reportErrorToUserModally(
            zulipLocalizations.errorCouldNotConnectTitle,
            details: zulipLocalizations.errorInvalidApiKeyMessage(
              account.realmUrl.toString()));
          await logOutAccount(this, accountId);
          throw AccountNotFoundException();
```

lib/model/store.dart
```diff
@@ -1218,8 +1248,17 @@ class UpdateMachine {
...
 
-    await store._globalStore._reloadPerAccount(store.accountId);
-    assert(_disposed);
+    try {
+      await store._globalStore._reloadPerAccount(store.accountId);
+    } on AccountNotFoundException {
+      // The account was logged out while we retry to replace the store,
+      // but that's OK as long as other code will be removing it from the UI
+      // (usually by using [routeToRemoveOnLogout]).
+      assert(debugLog('… Event queue not replaced; account logged out.'));
```
chrisbobbe, line 1257: nit: rather than "replace the store", how about "reload the store", as more directly what we mean, and matched with the name `_reloadPerAccount` in the code in the `try`.

test/widgets/app_test.dart
```diff
@@ -57,6 +59,99 @@ void main() {
...
+
+    testWidgets('push route when removing last route on stack', (tester) async {
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+      await prepare(tester);
+
+      final future = logOutAccount(testBinding.globalStore, eg.selfAccount.id);
+      await tester.pump(TestGlobalStore.removeAccountDuration);
+      await future;
+      check(pushedRoutes).single.isA<WidgetRoute>().page.isA<ChooseAccountPage>();
+      check(removedRoutes).single.isA<WidgetRoute>().page.isA<HomePage>();
```
chrisbobbe, line 149: Since a route is removed before a new one is pushed, how about ordering the `removedRoutes` check first? (Here and in other tests.)

chrisbobbe (commented):
Thanks! Comments below. I haven't yet tested the reload-on-dead-queue case manually yet; did you do that too?

lib/log.dart
```diff
@@ -48,7 +48,7 @@ typedef ReportErrorCallback = void Function(String? message, {String? details});
 // This gets set in [ZulipApp].  We need this indirection to keep `lib/log.dart`
 // from importing widget code, because the file is a dependency for the rest of
 // the app.
-ReportErrorCallback reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
+ReportErrorCancellablyCallback reportErrorToUserBriefly = defaultReportErrorToUserBriefly;
```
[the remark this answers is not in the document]
PIG208: Yeah. I think it was not clear without additional comments that this is describing a callback that accepts `null` for `message` when the error needs to be programmatically dismissed, instead of something UI specific.

lib/model/store.dart
PIG208: This is to match the other `debugLog` lines in the function, such as the ones that can lead to this line:
```dart
        assert(debugLog('Lost event queue for $store.  Replacing…'));
```

```dart
        assert(debugLog('BUG: Unexpected error in event polling: $error\n' // TODO(log)
          'Replacing event queue…'));
```

I think it is appropriate to use either "replace the event queue" or "reload the store" consistently. I feel that "replace the event queue" is less vague here because this attempted fix only makes sense when you relate it to the API layer.

chrisbobbe: Ah yeah; "replace the queue" or "replace the event queue" would be fine for the code comment.

PIG208:
The PR has been updated. Thanks for the review!

> I haven't yet tested the reload-on-dead-queue case manually yet; did you do that too?

Do you mean the case when the event queue expires but the auth keys aren't valid? It works for me when I leave the app in the background after invalidating the API key.

chrisbobbe:
> Do you mean the case when the event queue expires but the auth keys aren't valid?

Yeah; (1) in your commit message:

```
The method loadPerAccount has two call sites, i.e. places
where we send register-queue requests:

1. _reloadPerAccount through [UpdateMachine._handlePollError]
2. perAccount through [PerAccountStoreWidget] (the common case)
```

Which I guess, thinking more about it, doesn't actually require the event queue expiring, right, because polling could fail for other reasons—including an invalid-API-key error :)

> It works for me when I leave the app in the background after invalidating the API key.

Great!

lib/log.dart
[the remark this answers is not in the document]
chrisbobbe: Ah, I think I understand now: `reportErrorToUserBriefly(null)` is a way to clear the snackbars programmatically.

chrisbobbe:
> Yeah, a `GET /events` invalid-API-key error is typically a precursor to the invalid-API-key from the `_reloadPerAccount` code path, among other non-transient errors.

Yeah, that makes sense.

If `GET /events` has an invalid-API-key error, am I reading correctly that this code in `_handlePollError` is run?:

```dart
      default:
        assert(debugLog('BUG: Unexpected error in event polling: $error\n' // TODO(log)
          'Replacing event queue…'));
        _reportToUserErrorConnectingToServer(error);
        // Similar story to the _EventHandlingException case;
        // separate only so that that other case can print more context.
        // The bug here could be in the server if it's an ApiRequestException,
        // but our remedy is the same either way.
        isUnexpected = true;
```

If so, that seems wrong; this is something we expect to happen, though rarely, and the `_reportToUserErrorConnectingToServer` looks like redundant UI feedback.

PIG208:
> am I reading correctly that this code in _handlePollError is run
Yes. I was thinking of deferring that to #1054 (https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1947256331). For now, there is a TODO comment for this: 7334ccac31a7ea186af765a67530576854e9a2a4.

chrisbobbe:
Ah I see—thanks for reminding me of where you discussed this. :) I think deferring it to #1054 is fine.

chrisbobbe:
So, LGTM; marking for Greg's review.

lib/model/store.dart
```diff
@@ -1218,8 +1254,17 @@ class UpdateMachine {
...
     }
 
-    await store._globalStore._reloadPerAccount(store.accountId);
-    assert(_disposed);
+    try {
+      await store._globalStore._reloadPerAccount(store.accountId);
+    } on AccountNotFoundException {
+      // The account was logged out while we retry to replace the event queue,
+      // but that's OK as long as other code will be removing it from the UI
+      // (usually by using [routeToRemoveOnLogout]).
```
gnprice, line 1262: I don't understand how the second half of this sentence relates to the first. Why does that make it OK that the account was logged out — why would that be not-OK in the first place?

And what does "removing [the account] from the UI" mean?

… Hmm, it looks like this was copy-pasted from a comment in _PerAccountStoreWidgetState. But that one has a totally different meaning:
```
          // This widget will be showing [placeholder] perpetually,
          // but that's OK as long as other code will be removing it from the UI
          // (usually by using [routeToRemoveOnLogout]).
```

Showing a placeholder perpetually sure sounds like a bad thing — it's good to reason through why it'd be OK after all. And "removing [this widget] from the UI" makes sense because the widget is part of the UI in the first place.

lib/model/store.dart
```diff
@@ -1218,8 +1254,17 @@ class UpdateMachine {
       if (_disposed) return;
     }
 
-    await store._globalStore._reloadPerAccount(store.accountId);
-    assert(_disposed);
+    try {
+      await store._globalStore._reloadPerAccount(store.accountId);
+    } on AccountNotFoundException {
```
gnprice, line 1259: Is this try/catch something we should have even before the new handling of auth errors?

I think it is: suppose the app loses its event queue and goes to reload initial data, and while it's doing that the user navigates to the choose-account page and logs out the account. Then I think `loadPerAccount` would hit its existing `throw AccountNotFoundException` line.

So this can be moved to a separate prep commit. I think doing that will also help simplify the explanation in the main commit's commit message.

lib/model/store.dart
```diff
@@ -149,7 +150,34 @@ abstract class GlobalStore extends ChangeNotifier {
   /// and/or [perAccountSync].
   Future<PerAccountStore> loadPerAccount(int accountId) async {
```
gnprice, line 151: nit in commit message:

    The method loadPerAccount has two call sites, i.e. places
    where we send register-queue requests:
    
    1. _reloadPerAccount through [UpdateMachine._handlePollError]
    2. perAccount through [PerAccountStoreWidget] (the common case)

I wouldn't call either of these "the common case" (which implies all other cases are rare or edge cases) — I think both of these are common. When I open the app it quite often was already in the background, but had been for more than 10 minutes, and that hits case 1.

lib/model/store.dart
```diff
@@ -149,7 +150,34 @@ abstract class GlobalStore extends ChangeNotifier {
...
     assert(_accounts.containsKey(accountId));
-    final store = await doLoadPerAccount(accountId);
+    final PerAccountStore store;
+    try {
+      store = await doLoadPerAccount(accountId);
+    } catch (e) {
+      switch (e) {
+        case HttpException(httpStatus: 401):
+          // The API key is invalid and the store can never be loaded
+          // unless the user retries manually.
```
gnprice, line 160: Let's raise a thread on `#api design` asking for feedback on this condition `httpStatus: 401`. It seems quite reasonable, but it'd be good to make sure that everyone's on the same page about the semantics we'll be counting on from the server.

(For most things in the API we just work from the API docs with no further consultation — but this is an area of the API that's not clearly documented. The closest is https://chat.zulip.org/api/rest-error-handling , but for most errors that doesn't specify HTTP status codes, and indeed it specifies a Zulip API error code of INVALID_API_KEY which we learned at https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1945865983 isn't what the server actually produces in the most obvious case of an API key being invalid.)

gnprice (commented):
Thanks @PIG208 for building this, and @chrisbobbe for the previous reviews! Comments below. I didn't finish reading the last commit's tests, but read everything else.

lib/model/store.dart
PIG208: [CZO conversation](https://chat.zulip.org/#narrow/channel/378-api-design/topic/general.20handling.20HTTP.20status.20code.20401)

From that thread, we confirmed that checking for 401 is the desired approach.

lib/model/store.dart
PIG208: Yeah, I wanted to note that it is not bad when the event queue cannot be replaced. How about:

```
      // The event queue didn't get replaced because the account was logged out,
      // but that's OK as long as other code will ensure that the user
      // eventually navigates from the route associated with the old queue
      // (usually by using [routeToRemoveOnLogout]).
```

lib/widgets/app.dart
```diff
@@ -211,7 +223,13 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
         theme: themeData,
 
         navigatorKey: ZulipApp.navigatorKey,
-        navigatorObservers: widget.navigatorObservers ?? const [],
+        navigatorObservers: [
+          if (widget.navigatorObservers != null)
+            ...widget.navigatorObservers!,
+            // This must be the last item to maintain the invariant
+            // that the navigator stack is always non-empty.
+            _EmptyStackNavigatorObserver(),
```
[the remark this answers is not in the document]
PIG208: Leaving this comment out because we don't have other observers right now and that there's currently no issue prompting us to investigate further.

test/widgets/app_test.dart
```diff
@@ -57,6 +59,99 @@ void main() {
...
+        expectedTitle: 'Could not connect',
+        expectedMessage:
+          'Your account at https://chat.example/ could not be authenticated.'
+          ' Please try logging in again or use another account.')));
+      // No more routes are pushed after dismissing the error dialog,
+      // because the navigator stack was non-empty.
+      check(pushedRoutes).isEmpty();
+    });
+
+    testWidgets('push route when popping last route on stack', (tester) async {
```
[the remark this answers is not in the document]
PIG208: It does get popped (a dialog route is popped when dismissed), but we didn't track `poppedRoutes`. Adding it to the tests so that we have that knowledge as well.

lib/model/store.dart
PIG208: I found it pretty hard to hit this existing `throw AccountNotFoundException` line with `TestGlobalStore` or `LiveGlobalStore1, as we have an assertion at the start of `PerAccountStore.fromInitialSnapshot`:

```dart
    final account = globalStore.getAccount(accountId)!;
```

For both implementations, `PerAccountStore.fromInitialSnapshot` is called at the end with no asynchronous gap, meaning that the null-check will always fail ahead of `AccountNotFoundException` being raised, if the user logs out during the asynchronous gap from the /api/register-queue request or `globalStore.updateAccount`:

```dart
  static Future<UpdateMachine> load(GlobalStore globalStore, int accountId) async {
    Account account = globalStore.getAccount(accountId)!;
    final connection = globalStore.apiConnectionFromAccount(account);

    final initialSnapshot = await _registerQueueWithRetry(connection);
    final t = (stopwatch..stop()).elapsed;
    assert(debugLog("initial fetch time: ${t.inMilliseconds}ms"));

    if (initialSnapshot.zulipVersion != account.zulipVersion
        || initialSnapshot.zulipMergeBase != account.zulipMergeBase
        || initialSnapshot.zulipFeatureLevel != account.zulipFeatureLevel) {
      account = await globalStore.updateAccount(accountId, AccountsCompanion(
        zulipVersion: Value(initialSnapshot.zulipVersion),
        zulipMergeBase: Value(initialSnapshot.zulipMergeBase),
        zulipFeatureLevel: Value(initialSnapshot.zulipFeatureLevel),
      ));
      connection.zulipFeatureLevel = initialSnapshot.zulipFeatureLevel;
    }

    final store = PerAccountStore.fromInitialSnapshot(
      globalStore: globalStore,
      accountId: accountId,
      connection: connection,
      initialSnapshot: initialSnapshot,
    );
    final updateMachine = UpdateMachine.fromInitialSnapshot(
      store: store, initialSnapshot: initialSnapshot);
    updateMachine.poll();
    if (initialSnapshot.serverEmojiDataUrl != null) {
      // TODO(server-6): If the server is ancient, just skip trying to have
      //   a list of its emoji.  (The old servers that don't provide
      //   serverEmojiDataUrl are already unsupported at time of writing.)
      unawaited(updateMachine.fetchEmojiData(initialSnapshot.serverEmojiDataUrl!));
    }
    // TODO do registerNotificationToken before registerQueue:
    //   https://github.com/zulip/zulip-flutter/pull/325#discussion_r1365982807
    unawaited(updateMachine.registerNotificationToken());
    return updateMachine;
  }
```

I think this is a bug that the `if (!_accounts.containsKey(accountId))` check was meant to catch. The error should be thrown earlier, ideally every time before the account is accessed, similar to the `if (mounted)` checks and `BuildContext`.

Modulo the bug, it would be helpful to have the try/catch here, except that it is unreachable until later.

chrisbobbe: > Opened https://github.com/zulip/zulip-flutter/pull/1183 for this.

Did you mean a different issue/PR number here? That number refers to the current PR.

PIG208: Ah. It should be #1354.

gnprice (commented):
Thanks for the revision! Still reading, but posting comments now, since I see you've been replying to my related comments over on #1354.

test/model/test_store.dart
```diff
@@ -158,6 +163,9 @@ class TestGlobalStore extends GlobalStore {
     );
     updateMachines[accountId] = UpdateMachine.fromInitialSnapshot(
       store: store, initialSnapshot: initialSnapshot);
+    if (loadPerAccountPostDuration != null) {
+      await Future<void>.delayed(loadPerAccountPostDuration!);
+    }
     return Future.value(store);
```
gnprice, line 169: Hypothetically, this can happen when `removeAccount` was called
    during `doLoadPerAccount`, possibly due to the user logging out
    from choose-account page while the store is getting reloaded.
    However, it is currently not reachable because of a bug in
    `UpdateMachine.load`.  See #1354.

We already have a test that reaches the analogous situation when initially loading an account: https://github.com/zulip/zulip-flutter/issues/1354#issuecomment-2658121787

So we should be able to use the same technique to test this change, without adding a new complication to TestGlobalStore. You'll see in that that existing test case that it uses the subclass LoadingTestGlobalStore, which gives the test much more direct hands-on control of how `doLoadPerAccount` behaves.

lib/widgets/app.dart
```diff
@@ -234,6 +252,39 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+      unawaited(navigator.push(
+        MaterialWidgetRoute(page: const ChooseAccountPage())));
+    }
+  }
+
+  @override
+  void didRemove(Route<void> route, Route<void>? previousRoute) async {
+    if (previousRoute == null) {
+      // The route removed is the bottom-most one.
+      _pushRouteIfEmptyStack();
```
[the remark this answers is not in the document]
PIG208: There is no distinction because I think there is no difference in user-facing behavior. Both implementation will be able to tell empty and non-empty stack apart correctly.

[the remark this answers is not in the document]
gnprice: If there's no difference in behavior, then that means we can use either one of the two checks instead of using both of them, right?

In that case it'd be better to use just the `previousRoute == null` check — that one is a lot simpler.

test/widgets/app_test.dart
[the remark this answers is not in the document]
gnprice: Great, that clears it up, thanks.

lib/model/store.dart
gnprice: Cool, thanks — as discussed on #1354, I agree that there should be a check earlier that throws AccountNotFoundException.

For exercising this case in a test (before the main changes at the tip of this PR), see new comment below: https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1955499803

test/model/store_test.dart
```diff
@@ -7,11 +7,13 @@ import 'package:flutter/foundation.dart';
...
 import 'package:zulip/api/core.dart';
+import 'package:zulip/api/exception.dart';
 import 'package:zulip/api/model/events.dart';
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/route/events.dart';
 import 'package:zulip/api/route/messages.dart';
 import 'package:zulip/api/route/realm.dart';
+import 'package:zulip/model/actions.dart';
 import 'package:zulip/log.dart';
 import 'package:zulip/model/store.dart';
```
gnprice, line 18: nit: keep imports sorted

test/model/store_test.dart
```diff
@@ -120,6 +122,39 @@ void main() {
     check(completers(1)).length.equals(1);
   });
 
+  test('GlobalStore.perAccount loading fails with HTTP status code 401', () => awaitFakeAsync((async) async {
+    addTearDown(testBinding.reset);
+
+    await testBinding.globalStore.insertAccount(eg.selfAccount.toCompanion(false));
+    testBinding.globalStore.loadPerAccountException = ZulipApiException(
+      routeName: '/register', code: 'UNAUTHORIZED', httpStatus: 401,
+      data: {}, message: '');
```
gnprice, line 131: Hmm, seems like there ends up being a lot of copies of producing this same exception object. It'd be good to have a shared helper to encode the details — to help with making them consistently realistic, as well as making the test cases simpler to read.

I'll write a quick commit to do that, and push it to the branch or perhaps as a separate small PR.

test/widgets/app_test.dart
```diff
@@ -57,6 +59,106 @@ void main() {
...
+        .single.equals(eg.selfAccount.id);
+      check(removedRoutes).single.isA<WidgetRoute>().page.isA<HomePage>();
+      check(poppedRoutes).isEmpty();
+      check(pushedRoutes).single.isA<DialogRoute<void>>();
+      pushedRoutes.clear();
+
+      await tester.tap(find.byWidget(checkErrorDialog(tester,
+        expectedTitle: 'Could not connect',
+        expectedMessage:
+          'Your account at https://chat.example/ could not be authenticated.'
```
gnprice, line 112: nit:
```suggestion
          'Your account at ${eg.selfAccount.realmUrl} could not be authenticated.'
```

otherwise this test case depends on the details of what realm URL eg.selfAccount happens to have

test/widgets/app_test.dart
```diff
@@ -57,6 +59,106 @@ void main() {
...
+      const loadPerAccountDuration = Duration(seconds: 30);
+      assert(loadPerAccountDuration > kTryAnotherAccountWaitPeriod);
+      testBinding.globalStore.loadPerAccountDuration = loadPerAccountDuration;
+      testBinding.globalStore.loadPerAccountException = ZulipApiException(
+        routeName: '/register', code: 'UNAUTHORIZED', httpStatus: 401,
+        data: {}, message: '');
+      await testBinding.globalStore.insertAccount(eg.selfAccount.toCompanion(false));
+      await prepare(tester);
+
+      await tester.pump(kTryAnotherAccountWaitPeriod);
```
gnprice, line 94: These test cases make sense, but they have enough complexity to them that I think they'd benefit from a line or two of comments narrating what each of these stanzas is doing. Would you try writing those?

test/widgets/app_test.dart
```diff
@@ -57,6 +59,106 @@ void main() {
...
+      testNavObserver.onRemoved = (route, prevRoute) => removedRoutes.add(route);
+      testNavObserver.onPopped = (route, prevRoute) => poppedRoutes.add(route);
+
+      await tester.pumpWidget(ZulipApp(navigatorObservers: [testNavObserver]));
+      await tester.pump(); // start to load account
+      check(pushedRoutes).single.isA<WidgetRoute>().page.isA<HomePage>();
+      pushedRoutes.clear();
+    }
+
+    testWidgets('do not push route to non-empty navigator stack', (tester) async {
```
gnprice, line 84: nit: I think the narrative would be clearer if these three test cases:

```dart
    testWidgets('do not push route to non-empty navigator stack', (tester) async {
    // …
    testWidgets('push route when popping last route on stack', (tester) async {
    // …
    testWidgets('push route when removing last route on stack', (tester) async {
```

appeared in the opposite order.

In particular the "non-empty stack" case is a variation on the "popping last route" case, with some extra wrinkles added. So it's clearest to understand when taking that other case as a point of departure.

gnprice (commented):
OK, and done reading the new revision.

In addition to the new round of comments starting at https://github.com/zulip/zulip-flutter/pull/1183#pullrequestreview-2616610466 , a couple of subthreads continue on the previous round:
https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1955469383
https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1955491660

lib/widgets/app.dart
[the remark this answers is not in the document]
PIG208: Oh I see! Sorry for the confusion. I was comparing an implementation with just `_pushRouteIfEmptyStack` and one with both `_pushRouteIfEmptyStack` and `previousRoute == null`. Both implementations would work.

Just having `_pushRouteIfEmptyStack` gives us the correct behavior, but only having `previousRoute == null` is insufficient because the bottom-most route can be removed while there are other routes on top of it.

test/model/store_test.dart
gnprice: Sent as #1355.

test/widgets/app_test.dart
PIG208: Added some comments. Mostly structured like "The navigator stack should contain…" with an optional explanation on what is expected to happen.

Because I feel like tracking the stack is essential for this kind of tests, and the `check(…Routes)` are useful checks, but not super intuitive for the reader.

test/model/test_store.dart
PIG208: Thanks! That's really helpful. The navigator tests are kept as-is, because they use `ZulipApp`, but the more bare-bone tests in `test/model/store_test.dart` are all switched to use this technique.

PIG208:
Thanks for the review! I have updated the PR and stacked it atop #1355 (with the analyzer error fixed).

test/model/store_test.dart
```diff
@@ -970,6 +994,68 @@ void main() {
...
+          await preparePoll();
+
+          prepareExpiredEventQueue();
+          updateMachine.debugAdvanceLoop();
+          async.elapse(Duration.zero);
+          check(store).isLoading.isTrue();
+          check(completers()).single.isCompleted.isFalse();
+
+          await completeLoading(async);
+          check(store).isLoading.isTrue();
```
gnprice, line 1028: This isLoading check and the one after it don't really check anything — `store` is still the original store, and would remain that way even if the reload succeeded.

test/stdlib_checks.dart
```diff
@@ -144,3 +145,7 @@ extension HttpMultipartFileChecks on Subject<http.MultipartFile> {
 extension MediaTypeChecks on Subject<MediaType> {
   Subject<String> get asString => has((x) => x.toString(), 'toString'); // TODO(checks): what's a good convention for this?
 }
+
+extension CompleterChecks<T> on Subject<Completer<T>> {
```
gnprice, line 149: nit: put this above the JSON and URL/HTTP stuff — it's about futures, which are a core part of the language whereas those are more peripheral

test/model/store_test.dart
```diff
@@ -970,6 +994,68 @@ void main() {
...
+        globalStore = LoadingTestGlobalStore(accounts: [eg.selfAccount]);
+        final future = globalStore.perAccount(eg.selfAccount.id);
+        completers()
+          ..single.complete(
+              eg.store(globalStore: globalStore, account: eg.selfAccount))
+          ..clear();
+        await future;
+        updateFromGlobalStore();
+        updateMachine.debugPauseLoop();
+        updateMachine.poll();
```
gnprice, line 1012: I think this group of tests would be easier to understand if it didn't try to use the system of shared locals that `updateFromGlobalStore` and the outer `preparePoll` are part of.

There's always a risk with that sort of setup that it becomes necessary to trace through a tangle of different layers to see what's going on. Here, these tests are using those variables but then don't actually want the whole package of how they're set up, so they also modify other aspects of how that system works, and it starts requiring a lot of flipping back and forth to understand.

Instead, this group can have its own local `globalStore`. It doesn't even need the other shared locals — `updateMachine` can be within this "prepare" helper, which can prepare the reload (the next few lines of `checkReloadFailure`) as well as the polling loop.

test/model/store_test.dart
```diff
@@ -1070,10 +1156,13 @@ class LoadingTestGlobalStore extends TestGlobalStore {
...
   @override
-  Future<PerAccountStore> doLoadPerAccount(int accountId) {
+  Future<PerAccountStore> doLoadPerAccount(int accountId) async {
     final completer = Completer<PerAccountStore>();
     (completers[accountId] ??= []).add(completer);
-    return completer.future;
+    final store = await completer.future;
+    updateMachines[accountId] = UpdateMachine.fromInitialSnapshot(
+      store: store, initialSnapshot: eg.initialSnapshot());
+    return store;
```
gnprice, line 1165: This is a bit unfortunate in that it complicates the picture for the existing tests that use this class. I think it's unnecessary if the new tests above just let go completely of the outer group's `preparePoll` and friends (as they've already half-done), as discussed in my preceding comment.

gnprice (commented):
Thanks for the revision!

I haven't yet gotten through the revised tests on the last commit, but here's comments on the rest.

PIG208:
Thanks for the review! I have updated the PR and stacked it atop #1359 to make use of `apiException`.

test/example_data.dart
```diff
@@ -18,10 +19,52 @@ void _checkPositive(int? value, String description) {
...
+    httpStatus: 400, code: 'BAD_REQUEST',
+    data: {}, message: message);
+}
+
+/// The error for the "events" route when the target event queue has been
+/// garbage collected.
+///
+/// https://zulip.com/api/get-events#bad_event_queue_id-errors
+ZulipApiException apiExceptionBadEventQueueId({
+  String queueId = 'fb67bf8a-c031-47cc-84cf-ed80accacda8',
```
gnprice, line 47: Is this what a queue ID looks like? It differs from what we have in `eg.initialSnapshot`:
```dart
    queueId: queueId ?? '1:2345',
```

test/model/store_test.dart
```diff
@@ -757,11 +781,7 @@ void main() {
     }
 
     void prepareExpiredEventQueue() {
-      connection.prepare(httpStatus: 400, json: {
-        'result': 'error', 'code': 'BAD_EVENT_QUEUE_ID',
-        'queue_id': updateMachine.queueId,
-        'msg': 'Bad event queue ID: ${updateMachine.queueId}',
-      });
+      connection.prepare(apiException: eg.apiExceptionBadEventQueueId());
```
gnprice, line 784: nit: Seems like this is more realistic when it uses the update machine's actual queue ID.

test/model/store_test.dart
```diff
@@ -972,6 +992,74 @@ void main() {
...
+      globalStore.completers[eg.selfAccount.id]!;
+
+    Future<void> prepareReload(FakeAsync async) async {
+      globalStore = LoadingTestGlobalStore(accounts: [eg.selfAccount]);
+      final future = globalStore.perAccount(eg.selfAccount.id);
+      completers()
+        ..single.complete(
+            eg.store(globalStore: globalStore, account: eg.selfAccount))
+        ..clear();
+      final store = await future;
```
gnprice, line 1008: nit: I think this is a bit clearer if it's made explicit that the `store` we end up with is the same one created with that `eg.store` call:
```suggestion
      final store = eg.store(globalStore: globalStore, account: eg.selfAccount);
      completers().single.complete(store);
      await future;
      completers().clear();
```

test/model/store_test.dart
```diff
@@ -972,6 +992,74 @@ void main() {
...
+
+    void checkReloadFailure({
+      required FutureOr<void> Function() completeLoading,
+    }) {
+      awaitFakeAsync((async) async {
+        await prepareReload(async);
+
+        check(completers()).single.isCompleted.isFalse();
+        await completeLoading();
+        check(completers()).single.isCompleted.isTrue();
```
gnprice, line 1030: nit:
```suggestion
        await prepareReload(async);
        check(completers()).single.isCompleted.isFalse();

        await completeLoading();
        check(completers()).single.isCompleted.isTrue();
```

that way the rhythm of prep, check; then prep, check; is reflected visually in the stanzas between blank lines

test/model/store_test.dart
```diff
@@ -972,6 +992,74 @@ void main() {
...
+        check(globalStore.takeDoRemoveAccountCalls()).single.equals(eg.selfAccount.id);
+
+        async.elapse(TestGlobalStore.removeAccountDuration);
+        check(globalStore.perAccountSync(eg.selfAccount.id)).isNull();
+
+        async.flushTimers();
+        // Reload never succeeds and there is no unhandled errors.
```
gnprice, line 1037: nit: "errors" is plural, so "there are":
```suggestion
        // Reload never succeeds and there are no unhandled errors.
```

gnprice (commented):
Thanks! Comments below on the parts that changed in the latest revision, all small.

Looking now at the last commit's tests, per https://github.com/zulip/zulip-flutter/pull/1183#pullrequestreview-2619186063 .

test/example_data.dart
PIG208: Yeah. This comes from [/api/get-events](https://zulip.com/api/get-events#bad_event_queue_id-errors). We also use "abc:123" in another test. It might be good to have such a realistic value in test/example_data.dart to share.

gnprice:
OK, finished reading. All looks good except (the comments above and) I felt the _PreventEmptyStack tests could use a bit more narrative beyond what you added at https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1956829014 ; so I've pushed an additional to-be-squashed commit with some edits there. Please take a look and then go ahead and squash that in.

PIG208:
Thanks for the reviews and edition! Updated the PR. For https://github.com/zulip/zulip-flutter/pull/1183#discussion_r1960629449, we can probably do a small sweep to factor out a shared example value, but that can be a follow-up.

test/example_data.dart
gnprice: I see, and in an actual /api/register-queue response I see a similar value.

I guess it must have changed at some point. Pretty sure in the past (years ago) it had the other form.

gnprice:
Thanks for the revision! All LGTM except I guess #1355 and #1359 should get merged first.

chrisbobbe:
I've just merged those :)

PIG208:
Thanks! Rebased the PR.

gnprice:
Great — merging.

PR #1187, opened 2024-12-19
Factor error-handling out of poll method

gnprice:
This is a follow-up to #1063 — it's very lightly revised from commits that were in my draft branch leading to that PR, and which I left out at the time because the PR was getting long and complex enough already.

In this PR, factoring this logic out of the [UpdateMachine.poll] method isn't for the sake of deduplicating anything: the methods the logic gets moved into are still private helpers tuned specifically for the needs of that method. But it allows that method to get quite a lot shorter, so its overall structure is more apparent. And it also allows each of the two phases of error-handling to be read in a bit more of a self-contained context, by having them in their own methods.

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -996,17 +996,22 @@ class UpdateMachine {
     }());
   }
 
+  Future<void> _debugLoopWait() async {
+    await _debugLoopSignal!.future;
+    if (_disposed) return;
+    assert(() {
+      _debugLoopSignal = Completer();
+      return true;
+    }());
+  }
+
   void poll() async {
     assert(!_disposed);
     try {
       while (true) {
         if (_debugLoopSignal != null) {
-          await _debugLoopSignal!.future;
+          await _debugLoopWait();
           if (_disposed) return;
-          assert(() {
-            _debugLoopSignal = Completer();
-            return true;
-          }());
         }
 
         final GetEventsResult result;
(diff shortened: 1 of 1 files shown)


lib/model/store.dart
```diff
@@ -1142,30 +1130,40 @@ class UpdateMachine {
...
+  /// Deal with an error in [poll]: reload server data to replace the store,
+  /// after reporting the error as appropriate to the user and/or developer.
   ///
-  /// Reports the error if appropriate to the user and/or developer;
-  /// then returns true just if the error was unexpected.
-  bool _triagePollError(Object error) {
+  /// See also:
+  ///  * [_handlePollRequestError], which handles certain errors
+  ///    and causes them not to reach this method.
+  Future<void> _handlePollError(Object error) async {
```
PIG208, line 1165: Since this helper becomes the only user of `_unexpectedErrorBackoffMachine`, would it be better to move that `BackoffMachine` here?

PIG208 (commented):
Thanks! The result looks cleaner. I found the commit `store [nfc]: Move polling backoff/reload inside error helpers` to be slightly harder to read, but overall makes sense to me. Left a comment.

lib/model/store.dart
gnprice: Hmm, good thought.

I'll actually move it next to `_pollBackoffMachine` (which is also moving in this commit — so in fact this basically means they stay together). I think it's useful context when looking at one backoff machine to see that there's also this other related one.

gnprice:

PIG208:
Thanks! LGTM

PR #1188, opened 2024-12-20
Unify httpStatus field on ApiRequestException subclasses

gnprice:
Like #1187, this is a follow-up to #1063. It's a thought I had while working on those changes, but didn't pursue then because the PR was long and complex enough already.

#### ab291ad06 api [nfc]: Reorder NetworkException to before sibling classes

We're about to group the others together under a new sibling of
this class, so it'll be cleanest if this one doesn't appear in the
midst of the others.

#### 2eea63516 store [nfc]: Use HttpException to simplify error cases

This expresses more directly what we really mean here: if the
response had either the HTTP status code or the Zulip API error code
that mean a rate-limit error, treat it as a rate-limit error.

### DIFF
--- a/lib/api/exception.dart
+++ b/lib/api/exception.dart
@@ -28,7 +28,10 @@ sealed class ApiRequestException implements Exception {
   String toString() => message;
 }
 
-/// A network-level error that prevented even getting an HTTP response.
+/// A network-level error that prevented even getting an HTTP response
+/// to some Zulip API network request.
+///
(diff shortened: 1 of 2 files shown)


chrisbobbe:
LGTM, thanks! There's a conflict but I think it's just an indentation thing; please merge at will after resolving that.

gnprice:
Thanks! Done.

Yeah, #1187 moved some of the same code this edits, and made it a few levels less deeply indented.

PR #1192, opened 2024-12-20
github: Fix weblate action's rights, and update generated code.

alexmv:


.github/workflows/update-translations.yml
```diff
@@ -8,12 +11,30 @@ jobs:
...
+        # pubspec is satisfied.  It's uncommon for flutter/flutter to go
+        # more than 100 commits between tags.  Fetch 1000 for good measure.
+        run: |
+          git clone --depth=1000 -b main https://github.com/flutter/flutter ~/flutter
+          TZ=UTC git --git-dir ~/flutter/.git log -1 --format='%h | %ci | %s' --date=iso8601-local
+          echo ~/flutter/bin >> "$GITHUB_PATH"
+
+      - name: Update generated code
+        run: |
+          ./tools/check l10n --fix
```
gnprice, line 34: nit:
```suggestion
          tools/check l10n --fix
```

The rule is that once there's a slash in the command name, it means a file by that name, and there's no search in PATH (or for shell functions or builtins). Docs:
https://www.gnu.org/software/bash/manual/bash.html#Command-Search-and-Execution
(and POSIX says the same thing but is more annoying to reference:
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/V3_chap02.html#tag_18_09_01 )

So a leading `./` is useful only as a trick to get a slash in there when the file's name would otherwise have no slash.

.github/workflows/update-translations.yml
```diff
@@ -8,12 +11,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
       - name: Fetch and merge from Weblate
```
gnprice, line 15: nit: once these steps are separated by blank lines, I feel like I want the enclosing `jobs:` to be separated by a blank line from its siblings — otherwise it feels like a bit of a visual-hierarchy inversion

gnprice (commented):
Thanks! One substantive comment, one interesting nit, and one boring one.

gnprice:
Also it'd be great to get to test this (and similar PRs) before merging it to main.

I floated a possible solution for that just now in the chat thread:
https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/weblate.20translations.20.28.23F276.29/near/2011576

.github/workflows/update-translations.yml
alexmv: I'm aware that it's not _required_ -- however, I like its specificity, and that it makes the behaviour consistent and correct if someone naïvely moves the file up a directory.

Regardless, I'll defer to your preference here.

alexmv:
Closing in preference to #1262.

PR #1195, opened 2024-12-21
compose: Skip failing test for windows.

apoorvapendse:
The `attach from camera` test fails on
Windows due to the the path separator
used in the split function of the XFile.name
getter in `cross_file` being `\` instead of `/`.
This results in the getter returning the
entire path on Windows instead of the
last word present after the last '/' in the path.

Read this excellent explanation by
@chrisbobbe for better understanding:
https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/failing.20composebox.20test.20while.20running.20check.20script/near/2000301

### DIFF
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:checks/checks.dart';
 import 'package:file_picker/file_picker.dart';
@@ -581,7 +582,13 @@ void main() {
       });
 
       // TODO test what happens when capturing/uploading fails
-    });
+    },
+    // This test fails on Windows because [XFile.name] splits on
+    // [Platform.pathSeparator], corresponding to the actual host platform
+    // the test is running on, instead of the path separator for the
+    // target platform the test is simulating.
+    // TODO(upstream): unskip after fix to https://github.com/flutter/flutter/issues/161073
+    skip: Platform.isWindows);
   });
 
   group('error banner', () {


test/widgets/compose_box_test.dart
```diff
@@ -581,7 +582,7 @@ void main() {
       });
 
       // TODO test what happens when capturing/uploading fails
-    });
+    },skip: Platform.isWindows);
```
chrisbobbe, line 585: nit: space after `,`

Let's explain this in a code comment instead of the commit message; that's where it will be most useful to people who are just reading the code to understand why it is the way it is.

chrisbobbe (changes requested):
Thanks! Small comment below.

Also a commit message nit: let's make this the summary line:

```
compose_box [nfc]: On Windows, skip a test that fails
```

chrisbobbe:
Thanks! Marking for Greg's review; it's possible he'll want to tweak the explanation in the code comment, but the choice to skip this test on Windows seems fine to me.

test/widgets/compose_box_test.dart
```diff
@@ -581,7 +582,12 @@ void main() {
       });
 
       // TODO test what happens when capturing/uploading fails
-    });
+
+      // This test fails on Windows due to the Platform.pathSeparator.
+      // used in the split function of the XFile.name getter in cross_file being '\' instead of '/'.
+      // This results in the getter returning the entire path on Windows instead of the
+      // last word present after the last '/' in the path.
+    }, skip: Platform.isWindows);
```
gnprice, line 590: nit: this puts the comment inside the test body function, which is potentially confusing about what the comment is referring to; instead, put it after the `},` so it's clearly attached to the `skip:` argument

gnprice (commented):
Thanks to you both!

> it's possible he'll want to tweak the explanation in the code comment

Indeed. :-) In fact I also went and filed an upstream issue for the underlying bug in [XFile.name]:
* https://github.com/flutter/flutter/issues/161073

Merging, after those adjustments:
```diff
       // TODO test what happens when capturing/uploading fails
-
-      // This test fails on Windows due to the Platform.pathSeparator.
-      // used in the split function of the XFile.name getter in cross_file being '\' instead of '/'.
-      // This results in the getter returning the entire path on Windows instead of the
-      // last word present after the last '/' in the path.
-    }, skip: Platform.isWindows);
+    },
+    // This test fails on Windows because [XFile.name] splits on
+    // [Platform.pathSeparator], corresponding to the actual host platform
+    // the test is running on, instead of the path separator for the
+    // target platform the test is simulating.
+    // TODO(upstream): unskip after fix to https://github.com/flutter/flutter/issues/161073
+    skip: Platform.isWindows);
   });
```

Issue #45, opened 2023-03-31
lightbox: Use a friendlier format for the date, like "Yesterday at 4:47 PM"


chrisbobbe:
…or "A few seconds ago", etc.

Currently it shows something like "Mar 31, 2023 15:09:51".

gnprice:
Sure. Do see our guide to picking an issue:
https://zulip.readthedocs.io/en/latest/contributing/contributing.html#picking-an-issue-to-work-on

so in particular before claiming an issue, you should have already gotten started working on it, finding the relevant piece of code and getting an idea of how you'll approach it.

Also note this issue is in the "Post-launch" milestone, not "Launch" or earlier. That means the core team will have limited time to spend helping with this issue for the next few months. But I think the issue is simple enough that that should be fine.

mofirojean:
@gnprice I had a shot at this issue and is the result

The device i am using is my Techno Spark 7 - Android 11
Here is the result, the time stamps were changed to a more readable date format

![Screenshot_20241005-040206](https://github.com/user-attachments/assets/7d75823a-3401-4dc3-afb7-73c7cac8e6b4)
![Screenshot_20241005-041103](https://github.com/user-attachments/assets/1ec442d5-97c3-4aaa-a94f-0b688fe6c6b3)
![Screenshot_20241005-031245](https://github.com/user-attachments/assets/e33d2645-dfa4-4900-b397-e6f43324b65d)
![Screenshot_20241005-031431](https://github.com/user-attachments/assets/175454fe-019a-4a98-8d05-5d9f325e9df8)
![Screenshot_20241005-031134](https://github.com/user-attachments/assets/84812dfe-0bd3-4a5a-a6e5-2863b3dd9794)

Before making a PR, I would like to verify if a change like this would require that tests be made or updated? since i just added a method to the widget to format the date.

lakshya1goel:
Hello, I have worked on the issue, Please review the changes and let me know if anything else is required.
@chrisbobbe

lakshya1goel:
Hello, It would be great to receive some reviews on my PR. Looking forward for your response.
Thanks!

bebialen:
I noticed this issue is still open and labeled as "help wanted". Has the existing PR been abandoned, or can I contribute to improving the solution?

gnprice:
Thanks. In January we learned more about the options for how to implement this, which is reflected now in the issue description. Based on those updates, I think this no longer belongs as a "help wanted" issue

[...]

PR #1197, opened 2024-12-22
lightbox: Use a friendlier format for the date

lakshya1goel:
### Key Changes

Changed the timestamp format from `Mar 31, 2023 15:09:51` to a more readable `Mar 31, 2023 at 3:09 PM.`
Added proper localization support to adapt timestamps to the user's language and region preferences.
Optimized the timestamp logic to ensure accurate categorization into `Today`, `Yesterday`, or specific dates.
Modified test also to ensure the correct format of date.

Issue: #45

**Screenshots:**

| WhatsApp Image 2024-12-22 at 7 43 22 AM |WhatsApp Image 2024-12-22 at 7 43 22 AM(1) |
|-----------------------------|-----------------------------|
|WhatsApp Image 2024-12-22 at 7 43 23 AM| WhatsApp Image 2024-12-22 at 7 43 23 AM(1) |

### DIFF
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -152,11 +152,42 @@ class _LightboxPageLayoutState extends State<_LightboxPageLayout> {
 
     PreferredSizeWidget? appBar;
     if (_headerFooterVisible) {
-      // TODO(#45): Format with e.g. "Yesterday at 4:47 PM"
-      final timestampText = DateFormat
-        .yMMMd(/* TODO(#278): Pass selected language here, I think? */)
-        .add_Hms()
-        .format(DateTime.fromMillisecondsSinceEpoch(widget.message.timestamp * 1000));
+      final zulipLocalizations = ZulipLocalizations.of(context);
+
+      String formatLocalizedTimestamp(DateTime date) {
+        final now = DateTime.now();
+        final nowDateOnly = DateTime(now.year, now.month, now.day);
+        final messageDateOnly = DateTime(date.year, date.month, date.day);
+
+        final differenceInSeconds = now.difference(date).inSeconds;
+        final differenceInMinutes = now.difference(date).inMinutes;
+        final differenceInDays = nowDateOnly.difference(messageDateOnly).inDays;
+
+        if (differenceInSeconds < 60) {
+          return zulipLocalizations.aFewSecondsAgo;
+        } else if (differenceInMinutes < 60) {
+          return Intl.plural(
+            differenceInMinutes,
+            one: zulipLocalizations.oneMinuteAgo,
+            other: zulipLocalizations.minutesAgo(differenceInMinutes),
+            locale: zulipLocalizations.localeName,
+          );
+        } else if (differenceInDays == 0) {
+          final time = DateFormat.jm(zulipLocalizations.localeName).format(date);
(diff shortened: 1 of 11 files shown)


lakshya1goel:
Hi @gnprice,
Could you please take a look at this PR when you get a chance? Let me know if any changes are needed.
Thank you!

lib/widgets/lightbox.dart
```diff
@@ -152,11 +152,42 @@ class _LightboxPageLayoutState extends State<_LightboxPageLayout> {
...
+          );
+        } else if (differenceInDays == 0) {
+          final time = DateFormat.jm(zulipLocalizations.localeName).format(date);
+          return zulipLocalizations.todayAt(time);
+        } else if (differenceInDays == 1) {
+          final time = DateFormat.jm(zulipLocalizations.localeName).format(date);
+          return zulipLocalizations.yesterdayAt(time);
+        } else {
+          final dateStr = DateFormat('MMM d, yyyy', zulipLocalizations.localeName).format(date);
+          final timeStr = DateFormat('hh:mm a', zulipLocalizations.localeName).format(date);
```
chrisbobbe, line 183: Some locales prefer 24-hour format; we should follow that. In French, this should give "22:00" but instead it gives "10:00 PM".

test/widgets/lightbox_test.dart
```diff
@@ -249,7 +250,90 @@ void main() {
...
             eg.otherUser.fullName)));
       check(labelTextWidget.text.toPlainText())
-        .contains('Jul 23, 2024 23:12:24');
+        .contains('Jul 23, 2024 at 11:12 PM');
+
+      debugNetworkImageHttpClientProvider = null;
+    });
+
+    testWidgets('app bar shows localized timestamp: a few seconds ago', (tester) async {
+      prepareBoringImageHttpClient();
```
chrisbobbe, line 259: Why is this needed? (And the corresponding `debugNetworkImageHttpClientProvider = null;`.)

Similarly in the other added tests.

chrisbobbe (commented):
Thanks!

Comments below from a quick skim. Also, `DateFormat` has some instructions for initializing it so that it works with locales other than 'en_US'. It looks like you haven't done that yet; could you take a look? https://pub.dev/documentation/intl/latest/intl/DateFormat-class.html

Also, the tests need to exercise the app in some locale other than English, otherwise they're not really effectively testing that it's localized. 🙂 I think our way of simulating a specific locale in a test is:

```dart
                tester.platformDispatcher.localeTestValue = locale;
                tester.platformDispatcher.localesTestValue = [locale];
                addTearDown(tester.platformDispatcher.clearLocaleTestValue);
                addTearDown(tester.platformDispatcher.clearLocalesTestValue);
```

The chosen `locale` needs to be one that the app declares support for.

chrisbobbe:
On the issue (#45), I've also added details of a more satisfying future implementation using `ffi`, which I believe isn't available yet but hopefully will be this year. It should mean less complexity and less work for our volunteer translators. If you feel discouraged at the idea that your implementation may eventually be replaced, it's OK if you want to close this PR and focus on other work. 🙂

lakshya1goel:
Thanks for the detailed review, I think I should close this PR for now and focus on other issues.

Issue #1179, opened 2024-12-19
Tap target for channel in recipient header is shorter than header


gnprice:
When a channel message appears in a multi-channel narrow (like the combined feed), the recipient header has the channel name as well as topic.

Expected: Tapping anywhere in the channel part of the header should lead to the channel narrow.

Actual: Tapping directly on the text of the channel name leads to the channel narrow. But tapping slightly above or below leads instead to the topic narrow.

aurora500:
I would like to contribute to this issue,kindly assign me

The problem is that the entire recipient header (including the channel area) is wrapped in a single GestureDetector that navigates to the topic narrow. We need to properly separate the gesture areas for the channel and topic sections.

aurora500:
Here is solution Proof of Concept of the above issue:

![screen-recording](https://github.com/user-attachments/assets/46cacd9c-3fe8-4030-8435-6f5b4dd82c3b)

gnprice:
Sure, please go ahead.

E-m-i-n-e-n-c-e:
Hi I have worked on this. Please see #F1370 and [this comment](https://github.com/zulip/zulip-flutter/issues/1368#issuecomment-2674813239

PR #1200, opened 2024-12-23
Fix channel navigation when tapping recipient header areas

aurora500:
This PR fixes the tap target behavior in the recipient header of channel messages. Previously, tapping above or below the channel name would navigate to the topic view instead of the channel view. 

The fix adds `behavior: HitTestBehavior.opaque` to the GestureDetector which makes the entire area (including padding spaces) respond to taps.

 The fix improves the header's usability by ensuring the entire channel section responds correctly to taps, making navigation more intuitive.

Issue: #1179 

Screencast:
<video src="https://github.com/user-attachments/assets/46cacd9c-3fe8-4030-8435-6f5b4dd82c3b" controls></video>

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1054,6 +1054,7 @@ class StreamMessageRecipientHeader extends StatelessWidget {
         ?? '(unknown channel)'; // TODO(log)
 
       streamWidget = GestureDetector(
+	behavior: HitTestBehavior.opaque,
         onTap: () => Navigator.push(context,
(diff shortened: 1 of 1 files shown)


gnprice:
Thanks. Before we can review this, it will need a test — see our README: https://github.com/zulip/zulip-flutter#submitting-a-pull-request

gnprice:
Closing this PR to get it out of our queue, since it doesn't look like work is continuing on it. Thanks @aurora500 for giving the issue a go.

Issue #1173, opened 2024-12-18
Paste an image into the compose box, on Android


gnprice:
After last week's beta, a user [wrote](https://chat.zulip.org/#narrow/channel/48-mobile/topic/sending.20images/near/2002385):
> New app, as the old one, lacks one great feature - insert media content from buffer. For example […] you copy an image from anywhere (from web or gallery) and you just go paste it in compose box.
> That's how it works in Telegram, Discord, Whatsapp and pretty much everywhere else.
> Also would be cool to not see urls for images and video, but just thumbnails. Tho it's not only about flutter app, but about Zulip in general - it looks very outdated when you touch media content.

That person referred to iOS, but several of the other users who echoed or thumbs-upped the remark have mentioned elsewhere that they use Android, so it's desired there too.

On Android, Flutter added this feature last year (see #419), so I think it may be fairly quick to just wire that up.

### Related issue
* The iOS counterpart, which awaits upstream work: #925

Issue #419, opened 2023-11-27
Support images from keyboard


gnprice:
This serves some of the same purposes as:
* #418 

but may be a lot easier to implement.

Android has built-in support for inserting images into a text box as a feature of the keyboard, and Flutter supports that since https://github.com/flutter/flutter/issues/20796 . So we just need to wire that up for the message-content input box.

Chat thread: https://chat.zulip.org/#narrow/stream/137-feedback/topic/Mobile.20app.20GIF.20integration/near/1689275

PR #1203, opened 2024-12-24
compose: Support images from keyboard for Android

PIG208:
<details>

<summary>
Preview error message
</summary>

image

</details>

Fixes: #419
Fixes: #1173

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -4,6 +4,7 @@ import 'package:app_settings/app_settings.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:mime/mime.dart';
+import 'package:path/path.dart' as path;
 
 import '../api/exception.dart';
 import '../api/model/model.dart';
@@ -465,6 +466,39 @@ class _ContentInputState extends State<_ContentInput> with WidgetsBindingObserve
     }
   }
 
+  void _handleContentInserted(KeyboardInsertedContent content) async {
+    if (content.data == null || content.data!.isEmpty) {
+      // As of writing, the engine implementation never leaves `content.data` as
+      // `null`, but ideally it should be when the data cannot be read for
+      // errors.
+      //
+      // When `content.data` is empty, the data is not literally empty — this
+      // can also happen when the data can't be read from the input stream
+      // provided by the Android SDK because of an IO exception.
+      //
+      // See Flutter engine implementation that prepares this data:
+      //   https://github.com/flutter/flutter/blob/0ffc4ce00/engine/src/flutter/shell/platform/android/io/flutter/plugin/editing/InputConnectionAdaptor.java#L497-L548
+      // TODO(upstream): improve the API for this
+      final zulipLocalizations = ZulipLocalizations.of(context);
+      showErrorDialog(context: context,
+        title: zulipLocalizations.errorContentNotInsertedTitle,
+        message: zulipLocalizations.errorContentToInsertIsEmpty);
+      return;
+    }
+
+    final file = _File(
+      content: Stream.fromIterable([content.data!]),
+      length: content.data!.length,
+      filename: path.basename(content.uri),
+      mimeType: content.mimeType);
+
+    await _uploadFiles(
+      context: context,
+      contentController: widget.controller.content,
+      contentFocusNode: widget.controller.contentFocusNode,
+      files: [file]);
+  }
+
   static double maxHeight(BuildContext context) {
     final clampingTextScaler = MediaQuery.textScalerOf(context)
       .clamp(maxScaleFactor: 1.5);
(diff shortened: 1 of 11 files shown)


lib/widgets/compose_box.dart
```diff
@@ -362,6 +363,32 @@ class _ContentInputState extends State<_ContentInput> with WidgetsBindingObserve
     }
   }
 
+  void _handleContentInserted(KeyboardInsertedContent content) async {
+    if (!content.hasData) {
+      // The data can be empty when the URL is associated with an empty
+      // resource.  See Flutter engine implementation that provides this data:
+      //   https://github.com/flutter/flutter/blob/0ffc4ce00ea7bb912e379adf39354644eab2c17e/engine/src/flutter/shell/platform/android/io/flutter/plugin/editing/InputConnectionAdaptor.java#L497-L548
```
chrisbobbe, line 370: This comment is specifically about the case where `data` ends up being present as an empty array, right? In the code you linked to, I assume that specifically means `readStreamFully` came up empty in line 531:

https://github.com/flutter/flutter/blob/0ffc4ce00/engine/src/flutter/shell/platform/android/io/flutter/plugin/editing/InputConnectionAdaptor.java#L531
```java
final byte[] data = this.readStreamFully(is, 64 * 1024);
```

(Right?)

How about a different case: what could cause `data` to be null as opposed to present-but-empty? I noticed [`KeyboardInsertedContent.data`](https://api.flutter.dev/flutter/services/KeyboardInsertedContent/data.html) is optional. This would be helpful to document upstream there. 🙂

Also, looking around the whole `commitContent` function you linked to, it looks like there are various unhappy paths there, where the function can't do something desirable and returns `false`. Like if the Android version is too old. What's the user-facing behavior in those unhappy paths?

chrisbobbe (commented):
Exciting! Small question from a quick skim; I know this is still a draft.

lib/widgets/compose_box.dart
PIG208: Having swept through the Java code, I couldn't find other places where `data` is `null` and gets passed via a `textInputChannel.commitContent` call. The link points to the only current non-test call site. The Dart code constructor doesn't have code paths where `data` is explicitly `null` either.

`data` can be empty if (a) the file is empty or (b) there was an `IOException` during the read. So `data` should probably be `null` if there is an error reading it, so that we can handle it properly.

---

When returning `false`, no `textInputChannel.commitContent` call is made and our handler does not get called at all. But the return value is probably read by some part of the Android SDK code: https://developer.android.com/reference/android/view/inputmethod/BaseInputConnection

> Default implementation which invokes [View.performReceiveContent](https://developer.android.com/reference/android/view/View#performReceiveContent(android.view.ContentInfo)) on the target view if the view [allows](https://developer.android.com/reference/android/view/View#getReceiveContentMimeTypes()) content insertion; otherwise returns false without any side effects.

> true if this request is accepted by the application, whether the request is already handled or still being handled in background, false otherwise.

I assume the behavior is akin to when we do not provide `ContentInsertionConfiguration` at all. That also results in this method returning `false` because the `allowedMimeTypes` default to an empty list. This gives you an OS-specific error message:
image

test/widgets/compose_box_test.dart
```diff
@@ -464,20 +465,24 @@ void main() {
...
+      final narrow = ChannelNarrow(channel.streamId);
+      await prepareComposeBox(tester, narrow: narrow, streams: [channel]);
 
-        final channel = eg.stream();
-        final narrow = ChannelNarrow(channel.streamId);
-        await prepareComposeBox(tester, narrow: narrow, streams: [channel]);
+      // (When we check that the send button looks disabled, it should be because
+      // the file is uploading, not a pre-existing reason.)
+      await enterTopic(tester, narrow: narrow, topic: 'some topic');
+      await tester.enterText(contentInputFinder, 'see image: ');
```
chrisbobbe, line 479:
```suggestion
      await enterTopic(tester, narrow: narrow, topic: 'some topic');
      await enterContent(tester, 'see image: ');
```

chrisbobbe (commented):
Thanks! The implementation and tests look reasonable; some nits below. Could you please post screen recordings, showing both #419 and #1173?

PIG208:
https://github.com/user-attachments/assets/10efc2d5-eee7-4f6c-971d-b635cc2eb8ad

https://github.com/user-attachments/assets/826acb99-905e-475c-9f33-06fc19855716

PIG208:

chrisbobbe:
Thanks! Oh, also: could you check if this feature is available on all Android versions we support, and if it isn't, put a comment somewhere? (I guess with a `TODO(android-sdk-<n>)` to remove the comment when it becomes unnecessary.)

Otherwise LGTM, marking for Greg's review.

PIG208:
Looks like the support requires at least API level 25, which corresponds to Android 7.1:

https://github.com/flutter/flutter/blob/0ffc4ce00ea7bb912e379adf39354644eab2c17e/engine/src/flutter/shell/platform/android/io/flutter/plugin/editing/InputConnectionAdaptor.java#L498-L500

([image keyboard](https://developer.android.com/develop/ui/views/touch-and-input/image-keyboard) was added at the same API level)

This is way lower than Android 9, so it should be available on all versions we support.

chrisbobbe:
Cool; yeah, if 9 is our min-supported, then there's nothing to do here; a comment won't be needed.

PIG208:
Closing in favor of #1643.

PR #1214, opened 2024-12-27
Fix unread topic case sensitivity.

### LINKED ISSUE #980: Case-insensitive topics in unreads data

apoorvapendse:
Fixes #980.

| Before  | After  |
|---------|--------|
| before | after

[CZO Discussion](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Expected.20behaviour.20for.20case.20insensitivity.20for.20unreads.20.23F980).

### DIFF
--- a/lib/model/unreads.dart
+++ b/lib/model/unreads.dart
@@ -40,14 +40,14 @@ class Unreads extends ChangeNotifier {
     required int selfUserId,
     required ChannelStore channelStore,
   }) {
-    final streams = <int, Map<TopicName, QueueList<int>>>{};
+    final streams = <int, TopicMap<QueueList<int>>>{};
     final dms = <DmNarrow, QueueList<int>>{};
     final mentions = Set.of(initial.mentions);
 
     for (final unreadChannelSnapshot in initial.channels) {
       final streamId = unreadChannelSnapshot.streamId;
       final topic = unreadChannelSnapshot.topic;
-      (streams[streamId] ??= {})[topic] = QueueList.from(unreadChannelSnapshot.unreadMessageIds);
+      (streams[streamId] ??= TopicMap<QueueList<int>>()).set(topic,QueueList.from(unreadChannelSnapshot.unreadMessageIds));
     }
 
     for (final unreadDmSnapshot in initial.dms) {
@@ -86,7 +86,7 @@ class Unreads extends ChangeNotifier {
   // int count;
 
   /// Unread stream messages, as: stream ID → topic → message IDs (sorted).
-  final Map<int, Map<TopicName, QueueList<int>>> streams;
+  final Map<int, TopicMap<QueueList<int>>> streams;
 
   /// Unread DM messages, as: DM narrow → message IDs (sorted).
   final Map<DmNarrow, QueueList<int>> dms;
@@ -187,7 +187,9 @@ class Unreads extends ChangeNotifier {
 
   int countInTopicNarrow(int streamId, TopicName topic) {
     final topics = streams[streamId];
-    return topics?[topic]?.length ?? 0;
+    if (topics == null) return 0;
+    final qlist = topics.get(topic);
+    return qlist?.length ?? 0;
   }
 
   int countInDmNarrow(DmNarrow narrow) => dms[narrow]?.length ?? 0;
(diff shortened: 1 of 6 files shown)


lib/model/unreads.dart
```diff
@@ -91,6 +96,9 @@ class Unreads extends ChangeNotifier {
   /// Unread DM messages, as: DM narrow → message IDs (sorted).
   final Map<DmNarrow, QueueList<int>> dms;
 
+  // Maps lowercase topic names for lookup to one of the variants of the topic (case preserving).
+  final Map<String, TopicName> topicMapper;
```
gnprice, line 100: This won't have the right behavior because this map is shared between channels. So if there's one conversation [at](https://en.wikipedia.org/wiki/Ios) `#islands > Ios` and another at `#mobile > iOS`, then one or the other would get wrongly rendered with the other's casing.

gnprice (commented):
Thanks @apoorvapendse for building this! Sorry this PR fell through the cracks.

One high-level comment below. Then [see chat](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Expected.20behaviour.20for.20case.20insensitivity.20for.20unreads.20.23F980/near/2076806) on what I think would be a good strategy for the solution, if you'd like to take that forward.

gnprice:
> Sorry this PR fell through the cracks.

Ah, I guess it was marked as draft until yesterday — so that at least explains why we didn't review it sooner :slightly_smiling_face:. Anyway, thanks for sending it and for the update yesterday (and [in chat](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Expected.20behaviour.20for.20case.20insensitivity.20for.20unreads.20.23F980/near/2074480)).

gnprice:
Thanks for the revision! I see you mentioned [in chat](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Expected.20behaviour.20for.20case.20insensitivity.20for.20unreads.20.23F980/near/2083759) that this is ready for review again.

This looks like broadly the right structure. A few high-level comments:
* Let's put TopicMap in `lib/model/channel.dart`. Topics live within channels, and that file is already the home of other data about topics (namely the visibility policies).
* Let's have a good set of unit tests for TopicMap itself. I [mentioned](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Expected.20behaviour.20for.20case.20insensitivity.20for.20unreads.20.23F980/near/2076809) "a quick test case or two in `unreads_test.dart`", but that was in the context of what comes on top of already having TopicMap :slightly_smiling_face: 
* Do we need all those methods on TopicMap? If there are fewer, that's less to test.
* The names and signatures of the methods should match `Map`. So e.g. a wrapper for `Map.remove` should be called `remove`, not `delete`. That helps make it easy to remember and find the name of each method.

gnprice:
After a revision addressing that high-level feedback, this PR can go to maintainer review to get more detailed feedback.

gnprice:
Thanks for your work on this. It feels like there's still significant changes needed before we'd want to merge this; so in the interest of getting the issue resolved soon, I've asked @chrisbobbe to take over writing an implementation.

For a brief high-level picture of some of the things that still need revision:
* There are still several examples of this comment, from above:
  > * The names and signatures of the methods should match `Map`. So e.g. a wrapper for `Map.remove` should be called `remove`, not `delete`. That helps make it easy to remember and find the name of each method.

  In addition to keeping it easy to remember and find the names, this would simplify the diff.
* The implementation can be simplified by using the stdlib's handy [MapBase](https://api.dart.dev/dart-collection/MapBase-class.html) mixin. That also makes `toMap` unnecessary and simplifies tests.
* The comment in `_addAllInStreamTopic` shouldn't get deleted by this change, since it isn't related to what the comment is about.

Issue #1215, opened 2024-12-27
compose: Insert a <time> element ("global times")


chrisbobbe:
From [beta feedback](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20beta.3A.20Missing.20date.20picker.20support/near/2011823).

This issue is for offering the ["global times" feature](https://zulip.com/help/global-times) as part of the message-composing UI.

We already parse and render these elements in messages (that was #354; we still have #922 open for an uncommon edge case). We should offer a date-picker UI that helps the user insert the Markdown for one of these in the compose box.

This will be another button in the row beneath the content input; see [Figma](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3897-9239&m=dev):

<img width="373" alt="image" src="https://github.com/user-attachments/assets/3b6d92d6-953d-4f08-b86d-19b72a40eb04" />

The icon should come from the Figma [icons](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=557-24868&m=dev) page.

chrisbobbe:
Milestoning post-launch because the legacy apps don't offer this.

AshutoshKhadse23:
Hello,
I would like to work on it


https://github.com/user-attachments/assets/ec732456-dfdf-4c99-bda8-43a4e07a540c

chrisbobbe:
Sure, I've just assigned you @AshutoshKhadse23; please send a PR when your work is ready for review. Since this issue has the "post-launch" milestone, the core team won't have much time to help in the near term. As always, please self-review your work to check that it's clear and correct.

AshutoshKhadse23:
Hello @chrisbobbe ,
Sorry to close the first PR but could you review other and tell the changes is any ?
I had resolve conflict and unnecessary changes.

chrisbobbe:
See Greg's reply: https://github.com/zulip/zulip-flutter/pull/1232#issuecomment-2565856627

yash-agarwa-l:
Hi! If no one is currently working on this, I would love to pick it up.
I plan to implement a `_AttachGlobalTimeButton` in the compose box that chains `showDatePicker` and `showTimePicker` to capture the timestamp. Once selected, I'll handle the logic to format it into the required `<time:2025-12-31T13:00:00+05:30>` syntax and automatically insert it into the text field.

chrisbobbe:
[...]

PR #1231, opened 2024-12-28
Insert time element global

AshutoshKhadse23:
I had added the global time element 

Related #1215 

https://github.com/user-attachments/assets/6777034d-a77c-429d-8f60-52042af844b6


chrisbobbe:
Thanks for the PR! This needs to follow our Git style and exclude unrelated changes before we can review it.

Issue #1207, opened 2024-12-26
Compose-box border is subtle in dark mode


gnprice:
We got [this feedback](https://chat.zulip.org/#narrow/channel/48-mobile/topic/compose.20box.20design/near/1997848) on the current design of the compose box, in dark mode:
> I'm using the latest version and I notice that the box design is not highlighted compared to the text you read and it's difficult to understand where you have to start typing the text, you need a higher contrast or change the design […]

Indeed the compose box has a similar background to the message list, and the boundary is a bit hard to see.

### Implementation

We'll use this [updated design](https://chat.zulip.org/#narrow/channel/48-mobile/topic/compose.20box.20design/near/2015213) from @terpimost:

> if we change "border-bar" in dark mode from 50% black to 10% white I think we have a better contrast of the panel: […]
> Also in dark mode we could turn on a shadow […]

Details are [in Figma](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5908-64038&t=alSmAmdRXFDwT4IT-1). See also screenshots [in the chat thread](https://chat.zulip.org/#narrow/channel/48-mobile/topic/compose.20box.20design/near/2015213).

lakshya1goel:
Hello, I want to work on this issue. I have implemented it as well.
| Before                                    | After                                     |
|-------------------------------------------|-------------------------------------------|
| ![WhatsApp Image 2024-12-28 at 1 36 19 PM](https://github.com/user-attachments/assets/fc8a265b-6c5f-4c4d-80d3-a27bf0abac7d) | ![WhatsApp Image 2024-12-28 at 1 36 20 PM](https://github.com/user-attachments/assets/48df459f-3068-4d41-9ea6-5529057e1a50) |
| ![WhatsApp Image 2024-12-28 at 1 36 20 PM(2)](https://github.com/user-attachments/assets/34ccb768-d9a1-4aa7-8bc8-659ddc1e5003) | ![WhatsApp Image 2024-12-28 at 1 36 20 PM(1)](https://github.com/user-attachments/assets/2d7d8ef2-6aef-4d75-8b9b-286d69f23309) |

vats004:
@zulipbot claim

yashrajpu:
Please give me this assignment, it will only take about two minutes to complete. Thank you!

PR #1233, opened 2024-12-29
compose: Make compose-box border more prominent in dark mode

lakshya1goel:
This PR will update the top border and add shadow effect to the top of the compose box to make it more visible in dark mode

For reference, see the Figma design:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5908-64038&t=alSmAmdRXFDwT4IT-1

Fixes: #1207

|WhatsApp Image 2024-12-29 at 2 28 00 PM | WhatsApp Image 2024-12-29 at 2 28 01 PM |

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -22,6 +22,59 @@ import 'store.dart';
 import 'text.dart';
 import 'theme.dart';
 
+/// Compose-box styles that differ between light and dark theme.
+///
+/// These styles will animate on theme changes (with help from [lerp]).
+class ComposeBoxTheme extends ThemeExtension<ComposeBoxTheme> {
+  static final light = ComposeBoxTheme._(
+    boxShadow: null,
+  );
+
+  static final dark = ComposeBoxTheme._(
+    boxShadow: [BoxShadow(
+      color: DesignVariables.dark.bgTopBar,
+      offset: const Offset(0, -4),
+      blurRadius: 16,
+      spreadRadius: 0,
+    )],
+  );
+
+  ComposeBoxTheme._({
+    required this.boxShadow,
+  });
+
+  /// The [ComposeBoxTheme] from the context's active theme.
+  ///
+  /// The [ThemeData] must include [ComposeBoxTheme] in [ThemeData.extensions].
+  static ComposeBoxTheme of(BuildContext context) {
+    final theme = Theme.of(context);
+    final extension = theme.extension<ComposeBoxTheme>();
+    assert(extension != null);
+    return extension!;
+  }
+
+  final List<BoxShadow>? boxShadow;
+
+  @override
+  ComposeBoxTheme copyWith({
+    List<BoxShadow>? boxShadow,
+  }) {
+    return ComposeBoxTheme._(
+      boxShadow: boxShadow ?? this.boxShadow,
+    );
+  }
+
+  @override
+  ComposeBoxTheme lerp(ComposeBoxTheme other, double t) {
+    if (identical(this, other)) {
+      return this;
+    }
+    return ComposeBoxTheme._(
+      boxShadow: BoxShadow.lerpList(boxShadow, other.boxShadow, t)!,
+    );
+  }
+}
+
 const double _composeButtonSize = 44;
 
 /// A [TextEditingController] for use in the compose box.
@@ -1090,7 +1143,9 @@ class _ComposeBoxContainer extends StatelessWidget {
     //   the message list itself; if so, remember to update ComposeBox's dartdoc.
     return Container(width: double.infinity,
       decoration: BoxDecoration(
-        border: Border(top: BorderSide(color: designVariables.borderBar))),
+        border: Border(top: BorderSide(color: designVariables.borderBar)),
+        boxShadow: ComposeBoxTheme.of(context).boxShadow,
+      ),
       // TODO(#720) try a Stack for the overlaid linear progress indicator
       child: Material(
         color: designVariables.composeBoxBg,
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 
 import '../api/model/model.dart';
+import 'compose_box.dart';
 import 'content.dart';
 import 'emoji_reaction.dart';
 import 'message_list.dart';
@@ -32,6 +33,7 @@ ThemeData zulipThemeData(BuildContext context) {
         designVariables,
         EmojiReactionTheme.light,
         MessageListTheme.light,
+        ComposeBoxTheme.light,
       ];
     }
     case Brightness.dark: {
@@ -41,6 +43,7 @@ ThemeData zulipThemeData(BuildContext context) {
         designVariables,
         EmojiReactionTheme.dark,
         MessageListTheme.dark,
+        ComposeBoxTheme.dark,
       ];
     }
   }
@@ -175,7 +178,7 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     bgMenuButtonActive: Colors.black.withValues(alpha: 0.2),
     bgMenuButtonSelected: Colors.black.withValues(alpha: 0.25),
     bgTopBar: const Color(0xff242424),
-    borderBar: Colors.black.withValues(alpha: 0.5),
+    borderBar: const Color(0xffffffff).withValues(alpha: 0.1),
     borderMenuButtonSelected: Colors.white.withValues(alpha: 0.1),
     btnLabelAttLowIntDanger: const Color(0xffff8b7c),
     btnLabelAttMediumIntDanger: const Color(0xffff8b7c),
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -111,6 +111,15 @@ void main() {
     await tester.pump(Duration.zero);
   }
 
+  group('ComposeBoxTheme', () {
+    test('lerp light to dark, no crash', () {
+      final a = ComposeBoxTheme.light;
+      final b = ComposeBoxTheme.dark;
+
+      check(() => a.lerp(b, 0.5)).returnsNormally();
+    });
+  });
+
   group('ComposeContentController', () {
     group('insertPadded', () {
       // Like `parseMarkedText` in test/model/autocomplete_test.dart,


lib/widgets/theme.dart
```diff
@@ -182,6 +184,8 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
       btnLabelAttLowIntDanger: const Color(0xffff8b7c),
       btnLabelAttMediumIntDanger: const Color(0xffff8b7c),
       composeBoxBg: const Color(0xff0f0f0f),
+      composeBoxBorderBar: Colors.white.withValues(alpha: 0.1),
```
chrisbobbe, line 187: The Figma doesn't add a new design variable for the compose box border bar, it just changes the dark-theme value of an existing variable, `border-bar`. We should do the same.

lib/widgets/theme.dart
```diff
@@ -182,6 +184,8 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
...
+      composeBoxShadow: const Color(0xff242424),
```
chrisbobbe, line 188: The Figma uses an existing variable for the shadow color, `bg-top-bar`; we should do the same.

chrisbobbe (changes requested):
Thanks! I see this is a draft, but I gave it a quick skim and I noticed a different direction I'd like to take. 🙂

Since the shadow is only meant to be applied in dark theme, I'd like to avoid asking Flutter to do any computations for it when in light mode. For that, let's add a `ComposeBoxTheme` class in lib/widgets/compose_box.dart, with just one field `boxShadow` that's null in light theme and non-null in dark theme. `ContentTheme` and `MessageListTheme` are examples of the pattern I have in mind. So, something like this:

<details>
<summary>Proposal</summary>

```dart
/// Compose-box styles that differ between light and dark theme.
///
/// These styles will animate on theme changes (with help from [lerp]).
class ComposeBoxTheme extends ThemeExtension<ComposeBoxTheme> {
  factory ComposeBoxTheme.light(BuildContext context) {
    return ComposeBoxTheme._(
      boxShadow: null,
    );
  }

  factory ComposeBoxTheme.dark(BuildContext context) {
    return ComposeBoxTheme._(
      boxShadow: [BoxShadow(
        color: DesignVariables.dark.bgTopBar,
        offset: const Offset(0, -4),
        blurRadius: 16,
        spreadRadius: 0,
      )],
    );
  }

  ComposeBoxTheme._({
    required this.boxShadow,
  });

  /// The [ComposeBoxTheme] from the context's active theme.
  ///
  /// The [ThemeData] must include [ComposeBoxTheme] in [ThemeData.extensions].
  static ComposeBoxTheme of(BuildContext context) {
    final theme = Theme.of(context);
    final extension = theme.extension<ComposeBoxTheme>();
    assert(extension != null);
    return extension!;
  }

  final List<BoxShadow>? boxShadow;

  @override
  ComposeBoxTheme copyWith({
    List<BoxShadow>? boxShadow,
  }) {
    return ComposeBoxTheme._(
      boxShadow: boxShadow ?? this.boxShadow,
    );
  }

  @override
  ComposeBoxTheme lerp(ComposeBoxTheme other, double t) {
    if (identical(this, other)) {
      return this;
    }
    return ComposeBoxTheme._(
      boxShadow: BoxShadow.lerpList(boxShadow, other.boxShadow, t)!,
    );
  }
}
```

</details>

Then the caller can say `boxShadow: ComposeBoxTheme.of(context).boxShadow`.

Note the use of `DesignVariables.dark.bgTopBar` instead of adding a new design variable. The Figma doesn't add any new design variables. Also, it's `DesignVariables.dark.bgTopBar` instead of `DesignVariables.dark().bgTopBar` because of a small optimization, #1236; I suggest rebasing your work atop that PR to get that optimization.

lakshya1goel:
Thanks for the review, I am working on it and will update it as required ASAP.
Thanks!

PIG208:
There is a merge conflict. Could you resolve it? Thanks!

0pni (approved):
🥳

lakshya1goel:
Pushed the changes atop https://github.com/zulip/zulip-flutter/pull/1290 for the sake of CI.

lib/widgets/compose_box.dart
```diff
@@ -22,6 +22,63 @@ import 'store.dart';
...
+  factory ComposeBoxTheme.dark(BuildContext context) {
+    return ComposeBoxTheme._(
+      boxShadow: [BoxShadow(
+        color: DesignVariables.dark.bgTopBar,
+        offset: const Offset(0, -4),
+        blurRadius: 16,
+        spreadRadius: 0,
+      )],
+    );
+  }
```
chrisbobbe, line 44: Ah a small improvement on my proposal above 🙂 these don't actually need a `BuildContext` so they can just be static final fields:

```dart
  static final light = ComposeBoxTheme._(
    boxShadow: null,
  );

  static final dark = ComposeBoxTheme._(
    boxShadow: [BoxShadow(
      color: DesignVariables.dark.bgTopBar,
      offset: const Offset(0, -4),
      blurRadius: 16,
      spreadRadius: 0,
    )],
  );
```

0pni (approved):

lakshya1goel:
Hello, @chrisbobbe thanks for mentioning the changes. I have updated it accordingly. 
Please have a look and let me know if anything else is required.
Thanks!

lib/widgets/compose_box.dart
```diff
@@ -22,6 +22,59 @@ import 'store.dart';
 import 'text.dart';
```
PIG208, line 22: commit message nit:

```
compose: Compose-box border is subtle in dark mode 
```

Let's [start the commit summary line with an imperative verb](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commit-summary-part-2) instead.

lib/widgets/compose_box.dart
```diff
@@ -22,6 +22,59 @@ import 'store.dart';
...
+  }
+
+  final List<BoxShadow>? boxShadow;
+
+  @override
+  ComposeBoxTheme copyWith({
+    List<BoxShadow>? boxShadow,
+  }) {
+    return ComposeBoxTheme._(
+      boxShadow: boxShadow ?? this.boxShadow,
```
PIG208, line 63: This is quite a nit and not a blocker for this PR: there is no way for a caller of `copyWith` to override the value of `boxShadow` to `null`, which is a valid value.

`ThemeExtension` documents the method this way:

```
  /// Creates a copy of this theme extension with the given fields
  /// replaced by the non-null parameter values.
  ThemeExtension<T> copyWith();
```

I guess using nullable values for theme extension fields is uncommon but not incorrect. But there can be instances where the distinction between absent/null arguments matter.

lib/widgets/compose_box.dart
```diff
@@ -22,6 +22,59 @@ import 'store.dart';
...
+    );
+  }
+
+  @override
+  ComposeBoxTheme lerp(ComposeBoxTheme other, double t) {
+    if (identical(this, other)) {
+      return this;
+    }
+    return ComposeBoxTheme._(
+      boxShadow: BoxShadow.lerpList(boxShadow, other.boxShadow, t)!,
```
PIG208, line 73: I wonder how `lerpList` works if one of the `BoxShadow` is `null`, and what happens when the animation happens.

<details>
<summary>Implementation</summary>

```
  /// Linearly interpolate between two box shadows.
  ///
  /// If either box shadow is null, this function linearly interpolates from
  /// a box shadow that matches the other box shadow in color but has a zero
  /// offset and a zero blurRadius.
  ///
  /// {@macro dart.ui.shadow.lerp}
  static BoxShadow? lerp(BoxShadow? a, BoxShadow? b, double t) {
    if (identical(a, b)) {
      return a;
    }
    if (a == null) {
      return b!.scale(t);
    }
    if (b == null) {
      return a.scale(1.0 - t);
    }
    return BoxShadow(
      color: Color.lerp(a.color, b.color, t)!,
      offset: Offset.lerp(a.offset, b.offset, t)!,
      blurRadius: ui.lerpDouble(a.blurRadius, b.blurRadius, t)!,
      spreadRadius: ui.lerpDouble(a.spreadRadius, b.spreadRadius, t)!,
      blurStyle: a.blurStyle == BlurStyle.normal ? b.blurStyle : a.blurStyle,
    );
  }

  /// Linearly interpolate between two lists of box shadows.
  ///
  /// If the lists differ in length, excess items are lerped with null.
  ///
  /// {@macro dart.ui.shadow.lerp}
  static List<BoxShadow>? lerpList(List<BoxShadow>? a, List<BoxShadow>? b, double t) {
    if (identical(a, b)) {
      return a;
    }
    a ??= <BoxShadow>[];
    b ??= <BoxShadow>[];
    final int commonLength = math.min(a.length, b.length);
    return <BoxShadow>[
      for (int i = 0; i < commonLength; i += 1) BoxShadow.lerp(a[i], b[i], t)!,
      for (int i = commonLength; i < a.length; i += 1) a[i].scale(1.0 - t),
      for (int i = commonLength; i < b.length; i += 1) b[i].scale(t),
    ];
  }
```

</details>

So effectively, the `lerpList` only applies to the non-null `BoxShadow`, and it `lerp`s its offset and blurRadius without changing its color. Visually it looks fine to me.

PIG208 (commented):
This looks good to me! Left some comments, where only the commit message nit is actionable right now.

lib/widgets/compose_box.dart
chrisbobbe, line 73: Oh, and it would be good to have a test for this lerping logic, basically just to make sure it doesn't have an unexpected error. It doesn't have to be long, like:

```dart
      testWidgets('lerp light to dark, no crash', (tester) async {
        final a = ComposeBoxTheme.light;
        final b = ComposeBoxTheme.dark;
        check(() => a.lerp(b, 0.5)).returnsNormally();
      });
```

lakshya1goel:
Thank you for the reviews @PIG208 and @chrisbobbe , I have made the required changes. Please let me know if anything else is required.

test/widgets/compose_box_test.dart
```diff
@@ -37,6 +37,13 @@ import 'test_app.dart';
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  testWidgets('lerp light to dark, no crash', (tester) async {
```
PIG208, line 40: nit: tester is unused; this can use `test` instead

PIG208:
LGTM with one nit! Adding the integration tag but leaving the maintainer review in case Chris has anything to add.

lakshya1goel:
Thanks for the review @PIG208, pushed the revision.

test/widgets/compose_box_test.dart
PIG208: Just saw your update! I meant instead of using `testWidgets`, we should use `test` the function, not renaming the parameter to `test`.

lakshya1goel: Ohh, sorry for misinterpreting. Thanks for pointing out again. Pushed the revision.

lib/widgets/compose_box.dart
```diff
@@ -22,6 +22,59 @@ import 'store.dart';
 import 'text.dart';
 import 'theme.dart';
 
+/// Compose-box styles that differ between light and dark theme.
```
chrisbobbe, line 25:
```
compose: Make compose-box border more prominent in dark mode

Updated the top border and added shadow effect to the top of
the compose box to ensure it is subtle in dark mode.

For reference, see the Figma design:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5908-64038&t=alSmAmdRXFDwT4IT-1

Fixes: #1207
```

Commit-message nit: "[…] to ensure it is subtle in dark mode" isn't right 🙂—"subtle" would mean the visual distinction is _less_ obvious, and the point is to make it more obvious.

Instead, how about "to make it more visible in dark mode", or something like that?

test/widgets/compose_box_test.dart
```diff
@@ -37,6 +37,13 @@ import 'test_app.dart';
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  test('lerp light to dark, no crash', () {
+    final a = ComposeBoxTheme.light;
+    final b = ComposeBoxTheme.dark;
+
+    check(() => a.lerp(b, 0.5)).returnsNormally();
+  });
+
```
chrisbobbe, line 46: Let's wrap this in a `group('ComposeBoxTheme'` and put it near the other `group`s.

chrisbobbe (commented):
Thanks! Very small comments below.

lib/widgets/compose_box.dart
lakshya1goel: Ahh, sorry missed it last time.

lib/widgets/compose_box.dart
```diff
@@ -1089,7 +1142,9 @@ class _ComposeBoxContainer extends StatelessWidget {
     //   the message list itself; if so, remember to update ComposeBox's dartdoc.
     return Container(width: double.infinity,
       decoration: BoxDecoration(
-        border: Border(top: BorderSide(color: designVariables.borderBar))),
+        border: Border(top: BorderSide(color: designVariables.borderBar)),
+        boxShadow: ComposeBoxTheme.of(context).boxShadow,
```
gnprice, line 1146: The commit message says:

    For reference, see the Figma design:
    https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile\?node-id\=5908-64038\&t\=alSmAmdRXFDwT4IT-1

That link doesn't work, though. I think the `\` characters are undesired.

gnprice (commented):
Thanks @lakshya1goel for taking care of this, and @chrisbobbe and @PIG208 for the previous reviews!

These changes all look good. Just a couple of comments — one on the commit message, below, and two on the PR text:

The PR description is backward from what the PR does:
> to ensure it is subtle in dark mode.

I think this is the same issue as https://github.com/zulip/zulip-flutter/pull/1233#discussion_r1932799377 .

The PR title says:
> compose: Compose-box border is subtle in dark mode

which matches the issue title. But a PR title should say what the PR does, not describe the status quo from before the PR.

A good solution for both would be to edit the PR title and description to match the commit message. In general a PR title and description do a very similar job to a commit message; so when a PR has just one commit, they should usually be very similar to that one commit's message.

lib/widgets/compose_box.dart
lakshya1goel: Ohh, thanks for pointing. Corrected it.

lakshya1goel:
Thanks for the review @gnprice, updated the PR accordingly, PTAL.

gnprice:
Thanks! Looks good; merging.

Issue #1219, opened 2024-12-27
ui: Incorrect null-check operator in `_LoadingPlaceholderPageState.build`


chrisbobbe:
To reproduce:

1. Use a debug build so you can see errors printed to the console
2. Log into an account
3. Log out of the account ("Choose account" > three-dots menu > "Log out")
4. See the following in the console:

```
======== Exception caught by widgets library =======================================================
The following _TypeError was thrown building _LoadingPlaceholderPage(dirty, dependencies: [_GlobalStoreInheritedWidget, _LocalizationsScope-[GlobalKey#c7674]], state: _LoadingPlaceholderPageState#4549b):
Null check operator used on a null value

The relevant error-causing widget was: 
  _LoadingPlaceholderPage _LoadingPlaceholderPage:file:///Users/chrisbobbe/dev/zulip-flutter/lib/widgets/home.dart:36:31
When the exception was thrown, this was the stack: 
#0      _LoadingPlaceholderPageState.build (package:zulip/widgets/home.dart:184:36)
#1      StatefulElement.build (package:flutter/src/widgets/framework.dart:5841:27)
#2      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5733:15)
#3      StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5892:11)
#4      Element.rebuild (package:flutter/src/widgets/framework.dart:5445:7)
#5      ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:5715:5)
#6      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:5883:11)
#7      ComponentElement.mount (package:flutter/src/widgets/framework.dart:5709:5)
```

This is a latent bug introduced in a97ac4218 (fyi @PIG208). This widget shouldn't assume that an account for its `accountId` param exists; in particular, it won't exist in the short period after the account is removed from the database (for logout) and before `routeToRemoveOnLogout` is processed in a post-frame callback.

I see two plausible fixes:
- Remove the `accountId` param and pass…hmm, I guess a `realmUrl` param, for the "try another account" message. Or:
- Keep the `accountId` param but give it dartdoc with a reminder that the corresponding account might not exist, saying why.

In either case we should add a test that would fail before the fix but passes after it.

chrisbobbe:
Tentatively assigning @PIG208 as the author of a97ac4218.

PR #1235, opened 2024-12-30
home: Stop assuming account existence from loading page

PIG208:
We could pass realmUrl when initializing the `_LoadingPlaceholderPage`, but that will still require a check for the existence of the account.

The loading page will be blank when the account does not exist. The user can't easily reach this page because they can only logout from `ChooseAccountPage`, until we start invalidating API keys. Even then, they will only see the blank page briefly before getting navigated, so we should not show any text at all.

<details>
<summary>Screenshot</summary>

image

</details>

Fixes: #1219

### DIFF
--- a/test/widgets/home_test.dart
+++ b/test/widgets/home_test.dart
@@ -6,6 +6,7 @@ import 'package:zulip/api/model/events.dart';
 import 'package:zulip/model/narrow.dart';
 import 'package:zulip/model/store.dart';
 import 'package:zulip/widgets/about_zulip.dart';
+import 'package:zulip/widgets/actions.dart';
 import 'package:zulip/widgets/app.dart';
 import 'package:zulip/widgets/app_bar.dart';
 import 'package:zulip/widgets/home.dart';
@@ -21,6 +22,7 @@ import '../api/fake_api.dart';
 import '../example_data.dart' as eg;
 import '../flutter_checks.dart';
 import '../model/binding.dart';
+import '../model/store_checks.dart';
 import '../model/test_store.dart';
 import '../test_navigation.dart';
 import 'message_list_checks.dart';
@@ -441,5 +443,40 @@ void main () {
       // No additional wait for loadPerAccount.
       checkOnHomePage(tester, expectedAccount: eg.selfAccount);
     });
+  });
+
+  testWidgets('logging out while still loading', (tester) async {
+    // Regression test for: https://github.com/zulip/zulip-flutter/issues/1219
+    addTearDown(testBinding.reset);
+    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    await tester.pumpWidget(const ZulipApp());
+    await tester.pump(); // wait for the loading page
+    checkOnLoadingPage();
+
+    final element = tester.element(find.byType(MaterialApp));
+    final future = logOutAccount(element, eg.selfAccount.id);
+    await tester.pump(TestGlobalStore.removeAccountDuration);
+    await future;
+    // No error expected from briefly not having
+    // access to the account being logged out.
+    check(testBinding.globalStore).accountIds.isEmpty();
+  });
+
+  testWidgets('logging out after fully loaded', (tester) async {
+    // Regression test for: https://github.com/zulip/zulip-flutter/issues/1219
+    addTearDown(testBinding.reset);
+    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    await tester.pumpWidget(const ZulipApp());
+    await tester.pump(); // wait for the loading page
+    await tester.pump(); // wait for store
+    checkOnHomePage(tester, expectedAccount: eg.selfAccount);
+
+    final element = tester.element(find.byType(HomePage));
+    final future = logOutAccount(element, eg.selfAccount.id);
+    await tester.pump(TestGlobalStore.removeAccountDuration);
+    await future;
+    // No error expected from briefly not having
+    // access to the account being logged out.
+    check(testBinding.globalStore).accountIds.isEmpty();
   });
 }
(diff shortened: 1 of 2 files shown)


PIG208:
Not sure if we still show the loading indicator there, though. This page can be visible for a frame only after #737, but if we don't have the realmUrl, there will potentially be a layout shift as the circular indicator's position is affected by the hidden text.

gnprice:
> Not sure if we still show the loading indicator there, though. This page can be visible for a frame only after #737, but if we don't have the realmUrl, there will potentially be a layout shift as the circular indicator's position is affected by the hidden text.

Yeah, that's a fine reason to leave out the loading indicator. It's probably less jarring to have it disappear while the page is animating out than to have it suddenly shift while the page is animating out.

Note it's more than a frame — the animation duration for leaving a page is more like 200ms.

chrisbobbe:

test/widgets/home_test.dart
```diff
@@ -441,5 +443,22 @@ void main () {
...
       checkOnHomePage(tester, expectedAccount: eg.selfAccount);
     });
+
+    testWidgets('logging out', (tester) async {
+      // Regression test for: https://github.com/zulip/zulip-flutter/issues/1219
+      addTearDown(testBinding.reset);
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+      await tester.pumpWidget(const ZulipApp());
+      await tester.pump(Duration.zero); // wait for the loading page
+      await tester.pump(loadPerAccountDuration);
```
gnprice, line 453: What's the role of `loadPerAccountDuration` here? It doesn't look like anything else in the test refers to it.

test/widgets/home_test.dart
```diff
@@ -441,5 +443,22 @@ void main () {
...
+
+      final element = tester.element(find.byType(HomePage));
```
gnprice, line 455: This test feels like it could easily stop having any effect if something otherwise unrelated changes about the timing of the steps here. It could end up not getting to the loading placeholder at all — or going past it, to the normal contents of HomePage.

How about having it check that it can find a widget with text containing, say, "taking a while"? That should be pretty identifying of the loading page.

gnprice (commented):
Thanks! The fix looks good. A couple of comments below on the test.

test/widgets/home_test.dart
PIG208: Yeah it is not needed.

test/widgets/home_test.dart
PIG208: We can't check for "taking a while" because the text cannot be rendered without the `realmUrl`. The fix replaced the page with `SizedBox.shrink()` which doesn't really stand out as a widget to look for. I ended up adding `CircularProgressIndicator` back instead. It might result in a [layout shift](https://github.com/zulip/zulip-flutter/pull/1235#issuecomment-2565863123) as mentioned earlier.

gnprice: Making the test easier to write is definitely not a good reason to change the actual UX in a way we previously decided we didn't like. :slightly_smiling_face: 

The `SizedBox.shrink()` only comes after logging out, right? The step I'm asking for above is to check that we got to the loading placeholder page in the first place, right before logging the user out.

PIG208: Hmm, I see. Previously the account would finish loading before getting logged out. This setup relies on the behavior that after logging out, the app is briefly brought back to the loading page after removing the `PerAccountStore` and `Account`, and stays there until the route is popped from the navigator.

By removing the extra pump, checking it before logging out works.

Looking for another way to check for the alternative loading placeholder page after logging out, I found that we might do something like this without modifying the UX:

```dart
      // Missing AppBar title currently only applies to the loading pages,
      // but the check is still somewhat brittle.
      check(tester.widget<AppBar>(find.byType(AppBar))).title.isNull();

      check(tester.widget<SizedBox>(find.byType(SizedBox)))
        ..width.equals(0)
        ..height.equals(0);
```

For now, I'm pushing a revision that checks for the loading placeholder page before logging out, with the UX restored.

gnprice: I see. Rereading #1219, it's really about what happens after the normal contents of HomePage have already loaded, and the account subsequently gets logged out. The involvement of `_LoadingPlaceholderPage` is really just a detail of how the bug happened — there's no reason there needs to be a "loading placeholder" involved in the process of removing an account.

OK, so in order to function as a regression test for that issue, the test should exercise that situation: the data fully loads, and then the account gets logged out. I think that's what an earlier revision of this PR (before last week) did.

But because the issue is about HomePage, not _LoadingPlaceholderPage, there's no need to have it check that a loading page is involved at that point… and it should go outside of the _LoadingPlaceholderPage test group. I think the test group is what confused me earlier — it's a way of saying the test is about _LoadingPlaceholderPage, so I noticed the test didn't really ensure _LoadingPlaceholderPage was involved in what it tested and that seemed like a problem.

In particular, consider what should happen if in the future we change the behavior of HomePage (perhaps via MaterialAccountWidgetRoute) so that when the account gets logged out, the `loadingPlaceholderPage` doesn't get built — instead it's just an empty page or something. This test will still be doing a useful job by checking that when the user gets logged out from HomePage, there's no crash — effectively checking that that new arrangement doesn't reintroduce this bug. But there's no need for it to complain about the fact that _LoadingPlaceholderPage is no longer present.

This new version of the test, where the loading page is still in place when the user gets logged out, is potentially useful too — so we might as well keep it, in addition to restoring the test that's a more direct regression test for #1219.

PIG208:
Thanks for the review! Updated the PR by restoring the old version of this test, with a minor tweak that checks that HomePage is fully loaded, and removed the check for loading page after logging out. The new version of the test is kept too.

gnprice:
Thanks! Looks good; merging.

PR #1236, opened 2024-12-30
theme [nfc]: Make DesignVariables.{light,dark} final fields, not constructors

chrisbobbe:
Nicer to compute each of these just once.


gnprice:
(nit: the status quo is they're constructors, not methods)

chrisbobbe:
Indeed :) thanks for the catch

PIG208:
LGTM! We should be able to apply this to `MessageListTheme` and etc., right?

Issue #1210, opened 2024-12-26
android: When opening a notification, "back" should not leave the account's context


chrisbobbe:
From [beta feedback](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Beta.3A.20after.20open.20via.20notification.2C.20back.20opens.20old.20server/near/2013981):

> Steps to reproduce: 
> - have at least two servers configured in the app 
> - use the app on one server, then switch away from it
> - receive a notification for the other server 
> - tap the notification to open zulip 
> - hit the back button

> Expected behavior: the back button should go to rhe next level up on the server that the notification came from. (For instance, if browsing a topic it should go up to the stream, or similar.)

> Observed behavior: the back button goes to a view of the other server.

-----

If the app is showing UI for Account A and you open a notification for Account B, we push a message-list route for Account B onto the nav stack that's supposed to belong to Account A. This means the "back" button will take you out of Account B's context, which can be confusing.

When we navigate on opening a notification, we should switch to a new nav stack for the notification's account (if different) before pushing the message-list route.

(The iOS counterpart is covered by #1147, for navigating at all when a notification is tapped.)

## Implementation

Probably we want to call `HomePage.navigate`—

```dart
  /// Navigate to [HomePage], ensuring that its route is at the root level.
  static void navigate(BuildContext context, {required int accountId}) {
    final navigator = Navigator.of(context);
    navigator.popUntil((route) => route.isFirst);
    unawaited(navigator.pushReplacement(
      HomePage.buildRoute(accountId: accountId)));
  }
```

—and then push the result of `MessageListPage.buildRoute` after that.

lakshya1goel:
Hi, I have started working on this, will raise a PR shortly. Thanks!

gnprice:
To update the status of this issue: #1240 fixed one case of this. The description says it applies when "opened from background", but thinking again about the code I think it actually applies to a situation I would term differently:
* That PR applies when the app was not running, and starts running because the user opened a notification.
* If the app was already running but in the background, I believe that PR doesn't apply. That would instead go through `didPushRouteInformation`, which wasn't affected.
* The third possible case is when the app was already running and was in the foreground. I believe that also goes through `didPushRouteInformation`.

So I believe the current state is that this issue is resolved for the case where the app wasn't running, but remains open for the case when the app was already running, either in the background or the foreground. @rajveermalviya does that sound right to you?

rajveermalviya:
[...]

Issue #1147, opened 2024-12-12
ios notifs: Support tapping a notification to open the conversation


chrisbobbe:
Like #123, but for iOS.

When you tap a new-message notification, we should open the message's conversation.

chrisbobbe:
Beta feedback requesting this: https://chat.zulip.org/#narrow/channel/48-mobile/topic/beta.20feedback.3A.20.28ios.29.20notifications.20should.20open.20conversation/near/2067876

PR #1240, opened 2024-12-30
notif: Use associated account as initial account; if opened from background

rajveermalviya:
Previously, when two accounts (Account-1 and Account-2) were logged in, the app always defaulted to showing the home page of Account-1 on launch. If the app was closed and the user opened a notification from Account-2, the navigation stack would be:
  HomePage(Account-1) -> MessageListPage(Account-2)

This commit fixes that behaviour, now when a notification is opened while the app is closed, the home page will correspond to the account associated with the notification's conversation.

This addresses #1210 for background notifications.

### DIFF
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -470,8 +470,9 @@ class NotificationDisplayManager {
     assert(url.scheme == 'zulip' && url.host == 'notification');
     final payload = NotificationOpenPayload.parseUrl(url);
 
-    final account = globalStore.accounts.firstWhereOrNull((account) =>
-      account.realmUrl == payload.realmUrl && account.userId == payload.userId);
+    final account = globalStore.accounts.firstWhereOrNull(
+      (account) => account.realmUrl.origin == payload.realmUrl.origin
+                && account.userId == payload.userId);
     if (account == null) { // TODO(log)
       final zulipLocalizations = ZulipLocalizations.of(context);
       showErrorDialog(context: context,
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -453,6 +453,39 @@ class NotificationDisplayManager {
 
   static String _personKey(Uri realmUrl, int userId) => "$realmUrl|$userId";
 
+  /// Provides the route and the account ID by parsing the notification URL.
+  ///
+  /// The URL must have been generated using [NotificationOpenPayload.buildUrl]
+  /// while creating the notification.
+  ///
+  /// Returns null and shows an error dialog if the associated account is not
+  /// found in the global store.
+  static AccountRoute<void>? routeForNotification({
+    required BuildContext context,
+    required Uri url,
+  }) {
+    final globalStore = GlobalStoreWidget.of(context);
+
+    assert(debugLog('got notif: url: $url'));
+    assert(url.scheme == 'zulip' && url.host == 'notification');
+    final payload = NotificationOpenPayload.parseUrl(url);
+
+    final account = globalStore.accounts.firstWhereOrNull((account) =>
+      account.realmUrl == payload.realmUrl && account.userId == payload.userId);
+    if (account == null) { // TODO(log)
+      final zulipLocalizations = ZulipLocalizations.of(context);
+      showErrorDialog(context: context,
+        title: zulipLocalizations.errorNotificationOpenTitle,
+        message: zulipLocalizations.errorNotificationOpenAccountMissing);
+      return null;
+    }
+
+    return MessageListPage.buildRoute(
+      accountId: account.id,
+      // TODO(#82): Open at specific message, not just conversation
+      narrow: payload.narrow);
+  }
+
   /// Navigates to the [MessageListPage] of the specific conversation
   /// given the `zulip://notification/…` Android intent data URL,
   /// generated with [NotificationOpenPayload.buildUrl] while creating
@@ -460,29 +493,16 @@ class NotificationDisplayManager {
   static Future<void> navigateForNotification(Uri url) async {
     assert(debugLog('opened notif: url: $url'));
 
-    assert(url.scheme == 'zulip' && url.host == 'notification');
-    final payload = NotificationOpenPayload.parseUrl(url);
-
     NavigatorState navigator = await ZulipApp.navigator;
     final context = navigator.context;
     assert(context.mounted);
     if (!context.mounted) return; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
 
-    final zulipLocalizations = ZulipLocalizations.of(context);
-    final globalStore = GlobalStoreWidget.of(context);
-    final account = globalStore.accounts.firstWhereOrNull((account) =>
-      account.realmUrl == payload.realmUrl && account.userId == payload.userId);
-    if (account == null) { // TODO(log)
-      showErrorDialog(context: context,
-        title: zulipLocalizations.errorNotificationOpenTitle,
-        message: zulipLocalizations.errorNotificationOpenAccountMissing);
-      return;
-    }
+    final route = routeForNotification(context: context, url: url);
+    if (route == null) return; // TODO(log)
 
     // TODO(nav): Better interact with existing nav stack on notif open
-    unawaited(navigator.push(MaterialAccountWidgetRoute<void>(accountId: account.id,
-      // TODO(#82): Open at specific message, not just conversation
-      page: MessageListPage(initNarrow: payload.narrow))));
+    unawaited(navigator.push(route));
   }
 
   static Future<Uint8List?> _fetchBitmap(Uri url) async {
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -178,8 +178,6 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
     return GlobalStoreWidget(
       child: Builder(builder: (context) {
         final globalStore = GlobalStoreWidget.of(context);
-        // TODO(#524) choose initial account as last one used
-        final initialAccountId = globalStore.accounts.firstOrNull?.id;
         return MaterialApp(
           onGenerateTitle: (BuildContext context) {
             return ZulipLocalizations.of(context).zulipAppTitle;
@@ -209,6 +207,8 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
           onGenerateRoute: (_) => null,
 
           onGenerateInitialRoutes: (_) {
+            // TODO(#524) choose initial account as last one used
+            final initialAccountId = globalStore.accounts.firstOrNull?.id;
             return [
               if (initialAccountId == null)
                 MaterialWidgetRoute(page: const ChooseAccountPage())
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -140,26 +140,6 @@ class ZulipApp extends StatefulWidget {
 
 class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
   @override
-  Future<bool> didPushRouteInformation(routeInformation) async {
-    switch (routeInformation.uri) {
-      case Uri(scheme: 'zulip', host: 'login') && var url:
-        await LoginPage.handleWebAuthUrl(url);
-        return true;
-      case Uri(scheme: 'zulip', host: 'notification') && var url:
-        await NotificationDisplayManager.navigateForNotification(url);
-        return true;
-    }
-    return super.didPushRouteInformation(routeInformation);
-  }
-
-  Future<void> _handleInitialRoute() async {
-    final initialRouteUrl = Uri.parse(WidgetsBinding.instance.platformDispatcher.defaultRouteName);
-    if (initialRouteUrl case Uri(scheme: 'zulip', host: 'notification')) {
-      await NotificationDisplayManager.navigateForNotification(initialRouteUrl);
-    }
-  }
-
-  @override
   void initState() {
     super.initState();
     WidgetsBinding.instance.addObserver(this);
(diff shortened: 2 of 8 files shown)


chrisbobbe (commented):
Here's a quick review :) in particular I haven't yet tested this manually on the office Android device.

rajveermalviya:
Thanks for the review @chrisbobbe! Pushed a revision, PTAL.

lib/widgets/app.dart
```diff
@@ -152,18 +152,56 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+
+      final zulipLocalizations = ZulipLocalizations.of(navigatorContext);
+      showErrorDialog(context: navigatorContext,
+        title: zulipLocalizations.errorNotificationOpenTitle,
+        message: zulipLocalizations.errorNotificationOpenAccountMissing);
     }
+
+    final globalStore = GlobalStoreWidget.of(context);
+    // TODO(#524) choose initial account as last one used
+    final initialAccountId = globalStore.accounts.firstOrNull?.id;
```
chrisbobbe, line 172: How about doing this (checking the contents of `globalStore.accounts`) inside the returned `InitialRouteListFactory` callback? That's what we do in the open-from-notification case.

Then in particular we don't have to think about what happens if the `firstOrNull` account is logged out between a call to `_ZulipAppState` and the time when Flutter decides to call this `InitialRouteListFactory` callback.

—ah, I see: its position outside the callback is the same in `main`. How about a prep commit that just puts the query on `globalStore.accounts` into the callback, then?

chrisbobbe: Seems also nice if it's below the open-from-notification code in that callback, so it's easier to see that it's not part of that logic.

chrisbobbe (commented):
Thanks! Small comments below, and all looks good in a quick manual test on an Android emulator.

Does this PR change behavior on iOS? We haven't implemented navigation on tapping a notification on iOS yet; that's #1147. I haven't tested on iOS because Apple makes it complicated to test client-side notification changes: https://github.com/zulip/zulip-mobile/blob/main/docs/howto/push-notifications.md#testing-client-side-changes-on-ios

test/notifications/display_test.dart
```diff
@@ -1070,6 +1070,7 @@ void main() {
 
     testWidgets('at app launch', (tester) async {
       addTearDown(testBinding.reset);
+      addTearDown(tester.binding.platformDispatcher.clearDefaultRouteNameTestValue);
```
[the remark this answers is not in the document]
rajveermalviya: Moved this and the formating change in this test to a separate commit.

rajveermalviya:
Thanks for the review @chrisbobbe. Pushed a new revision, PTAL.

lib/notifications/display.dart
```diff
@@ -456,36 +456,58 @@ class NotificationDisplayManager {
...
+  static ({Route<void> route, int accountId})? routeForNotification({
+    required GlobalStore globalStore,
+    required Uri url,
+  }) {
+    assert(debugLog('got notif: url: $url'));
+    assert(url.scheme == 'zulip' && url.host == 'notification');
+    final payload = NotificationOpenPayload.parseUrl(url);
+
+    final account = globalStore.accounts.firstWhereOrNull((account) =>
+      account.realmUrl == payload.realmUrl && account.userId == payload.userId);
```
chrisbobbe, line 475: How likely do we think it is that these realm URLs could differ in boring ways that shouldn't matter? Like `https://chat.zulip.org/` vs. ``https://chat.zulip.org`. Do we need to just compare `.origin` instead, like we do for most (all?) other realm-URL checks?

lib/widgets/app.dart
```diff
@@ -152,6 +152,21 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
     return super.didPushRouteInformation(routeInformation);
   }
 
+  InitialRouteListFactory _handleGenerateInitialRoutes(BuildContext context) {
```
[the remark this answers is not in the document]
chrisbobbe: It would be even clearer if it were split into two commits: one that changes when `globalStore.accounts` is queried, and one for making the helper function (the pure refactor).

chrisbobbe (commented):
Thanks! Small comments below.

lib/notifications/display.dart
rajveermalviya: Right, I've added a commit that changes this query to compare only using `.origin`.

rajveermalviya:

chrisbobbe:

lib/widgets/app.dart
```diff
@@ -152,18 +152,56 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+      if (initialRouteUrl case Uri(scheme: 'zulip', host: 'notification')) {
+        final notificationResult = NotificationDisplayManager.routeForNotification(
+          globalStore: globalStore,
+          url: initialRouteUrl);
+
+        if (notificationResult != null) {
+          return [
+            HomePage.buildRoute(accountId: notificationResult.accountId),
+            notificationResult.route,
+          ];
```
gnprice, line 183: It's a bit unsatisfying to have `routeForNotification` have to return two values like this, when the route already includes the account ID.

I think we can simplify to make it look like this:
```suggestion
          return [
            HomePage.buildRoute(accountId: route.accountId),
            route,
          ];
```

with a small refactor to make the account ID more transparent on the route.

I'll push a small NFC onto your PR branch that makes that refactor. Then you can move it to the front of the branch and make use of it.

lib/widgets/app.dart
```diff
@@ -152,18 +152,56 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+      final navigatorContext = navigator.context;
+      assert(navigatorContext.mounted);
+      // TODO(linter): this is impossible as there's no actual async gap, but
+      //   the use_build_context_synchronously lint doesn't see that.
+      if (!navigatorContext.mounted) return;
+
+      final zulipLocalizations = ZulipLocalizations.of(navigatorContext);
+      showErrorDialog(context: navigatorContext,
+        title: zulipLocalizations.errorNotificationOpenTitle,
+        message: zulipLocalizations.errorNotificationOpenAccountMissing);
```
gnprice, line 169: This is exactly the same error dialog as `navigateForNotification` shows, right?

Which is probably a good thing — they're basically the same situation, differing only in whether the app was already running (including in the background).

So maybe `routeForNotification` should go ahead and show the error. That'd keep the details of this error in one place, over in the notifications code which is good because it's an error about the notification.

test/notifications/display_test.dart
```diff
@@ -1129,6 +1145,45 @@ void main() {
...
+        (it) => it.isA<MaterialAccountWidgetRoute>()
+          ..accountId.equals(accountB.id)
+          ..page.isA<HomePage>(),
+        (it) => it.isA<MaterialAccountWidgetRoute>()
+          ..accountId.equals(accountB.id)
+          ..page.isA<MessageListPage>()
+            .initNarrow.equals(SendableNarrow.ofMessage(message,
+              selfUserId: accountB.userId))
+      ]);
+      pushedRoutes.clear();
```
gnprice, line 1185: I think we can tighten this to make it compare more directly with the neighboring tests, by extending `takeStartingRoutes` a bit:
```suggestion
      takeStartingRoutes(account: accountB);
      matchesNavigation(check(pushedRoutes).single, accountB, message);
```

gnprice (commented):
Thanks both! This approach looks good. Some comments below.

lib/widgets/app.dart
```diff
@@ -177,9 +215,6 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
     final themeData = zulipThemeData(context);
     return GlobalStoreWidget(
       child: Builder(builder: (context) {
-        final globalStore = GlobalStoreWidget.of(context);
-        // TODO(#524) choose initial account as last one used
-        final initialAccountId = globalStore.accounts.firstOrNull?.id;
```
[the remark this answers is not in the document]
chrisbobbe: > […] I'd guess that those happen synchronously in the same frame.
> 
> Not worth spending time investigating that, if you don't already know of a way it can happen after all — Navigator is real complicated.

Correct, I'm not aware of a way it can happen after all.

rajveermalviya:

lib/widgets/app.dart
```diff
@@ -139,6 +139,50 @@ class ZulipApp extends StatefulWidget {
...
+
+      if (route != null) {
+        return [
+          HomePage.buildRoute(accountId: route.accountId),
+          route,
+        ];
+      } else {
+        // The account didn't match any existing accounts,
+        // fallthrough to show default the route below.
+      }
```
gnprice, line 173: nit:
```suggestion
        // The account didn't match any existing accounts,
        // fallthrough to show the default route below.
      }
```

?

lib/notifications/display.dart
```diff
@@ -453,36 +454,70 @@ class NotificationDisplayManager {
...
+    // A helper to avoid making this function an async function,
+    // allowing it to be called from non-async contexts and where
+    // BuildContext hasn't been initialized with localizations yet.
+    void showNotificationErrorDialog() async {
+      final navigator = await ZulipApp.navigator;
+      final navigatorContext = navigator.context;
+      assert(navigatorContext.mounted);
+      // TODO(linter): this is impossible as there's no actual async gap, but
+      //   the use_build_context_synchronously lint doesn't see that.
+      if (!navigatorContext.mounted) return;
```
gnprice, line 477: Instead of having to go look up a context via ZulipApp statics like this, let's have `routeForNotification` take a BuildContext argument. Its callers both have such an argument already.

Then it can also use that to look up the global store. That's how both of its callers are getting the global store themselves. (As a bonus, one caller will no longer need to do that lookup at all.)

lib/notifications/display.dart
```diff
@@ -453,36 +454,70 @@ class NotificationDisplayManager {
...
+  /// Returns null and shows an error dialog if the associated account is not
+  /// found in the global store.
+  static AccountRoute<void>? routeForNotification({
+    required GlobalStore globalStore,
+    required Uri url,
+  }) {
+    // A helper to avoid making this function an async function,
+    // allowing it to be called from non-async contexts and where
+    // BuildContext hasn't been initialized with localizations yet.
+    void showNotificationErrorDialog() async {
```
gnprice, line 471: Ah, I guess this does add a bit of complication to my picture from the previous comment. But:
* One caller, `navigateForNotification`, already does that wait before it calls this function. So it can just pass the context it gets from that.
* In the other caller, `_handleGenerateInitialRoutes` is called from `MaterialApp.onGenerateInitialRoutes`. It's true that that callback doesn't get passed a BuildContext that's under the navigator (and so can be used by `showErrorDialog`).

  But: by the time `onGenerateInitialRoutes` gets called, the navigator has been mounted. In fact here's its call site, in `_WidgetsAppState.build`:
```dart
        child: Navigator(
          // …
          onGenerateInitialRoutes:
              widget.onGenerateInitialRoutes == null
                  ? Navigator.defaultGenerateInitialRoutes
                  : (NavigatorState navigator, String initialRouteName) {
                    return widget.onGenerateInitialRoutes!(initialRouteName);
                  },
```

(That calls `WidgetsApp.onGenerateInitialRoute`, which `MaterialApp` passed directly from `MaterialApp.onGenerateInitialRoute`.)

So by the time it's called, there's a `NavigatorState` object there.

Sadly that NavigatorState doesn't get passed on to our callback. But I believe that still means `ZulipApp.navigator` should work at that point, to get our hands on that `NavigatorState` after all.

lib/widgets/app.dart
gnprice: ah and another nit: "fall through" is the verb, while "fallthrough" is a noun

(similar to "clean up" / "cleanup" and so on, as in https://github.com/zulip/zulip-flutter/pull/1240#discussion_r1936369343 )

gnprice (commented):
Thanks for the revision!

This looks good except one further simplification I think can be made to the logic that interacts with the navigator. Also a couple of nits.

lib/notifications/display.dart
rajveermalviya: Right, updated to use `Navigator` from `ZulipApp.navigatorKey` instead in the commit `app: Pull out _handleGenerateInitialRoutes`.

rajveermalviya:

lib/widgets/app.dart
```diff
@@ -139,6 +139,52 @@ class ZulipApp extends StatefulWidget {
...
+  }
+
+  @override
+  void dispose() {
+    WidgetsBinding.instance.removeObserver(this);
+    super.dispose();
+  }
+
+  List<Route<dynamic>> _handleGenerateInitialRoutes(String initialRoute) {
+    // The `_ZulipAppState.context` lacks the required dependencies. Instead
```
gnprice, line 155: nit:

```suggestion
    // The `_ZulipAppState.context` lacks the required ancestors. Instead
```

A "dependency" in Flutter is a relationship that's created between an element and one of its ancestors when calling a method like `GlobalStoreWidget.of`. For example that method's doc says:
```
  /// The given build context will be registered as a dependency of the
  /// store.  This means that when the data in the store changes,
  /// the element at that build context will be rebuilt.
```

In turn the upstream method that one calls, `BuildContext.dependOnInheritedWidgetOfExactType`, says:
```dart
  /// Returns the nearest widget of the given type `T` and creates a dependency
  /// on it, or null if no appropriate widget is found.
  ///
  /// The widget found will be a concrete [InheritedWidget] subclass, and
  /// calling [dependOnInheritedWidgetOfExactType] registers this build context
  /// with the returned widget. When that widget changes (or a new widget of
  /// that type is introduced, or the widget goes away), this build context is
  /// rebuilt so that it can obtain new values from that widget.
```

So it's fine for the context to not already have a dependency; calling `GlobalStoreWidget.of` is the step that will create the dependency anyway.

What is needed is for the context to have a `GlobalStoreWidget` (and a `Navigator` and some other things) as *ancestors*, so that it can be used to find them.

lib/notifications/display.dart
```diff
@@ -453,36 +453,57 @@ class NotificationDisplayManager {
...
+    final globalStore = GlobalStoreWidget.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    assert(debugLog('got notif: url: $url'));
+    assert(url.scheme == 'zulip' && url.host == 'notification');
+    final payload = NotificationOpenPayload.parseUrl(url);
+
+    final account = globalStore.accounts.firstWhereOrNull(
+      (account) => account.realmUrl.origin == payload.realmUrl.origin
+                && account.userId == payload.userId);
```
gnprice, line 476: nit: this change is now happening in the commit
6c35bea98 notif: Use associated account as initial account; if opened from background

instead of in the intended commit
9f68ba304 notif: Query account by realm URL origin, not full URL

Should move back to the latter commit.

lib/widgets/app.dart
```diff
@@ -139,6 +139,52 @@ class ZulipApp extends StatefulWidget {
...
+    final initialRouteUrl = Uri.tryParse(initialRoute);
+    if (initialRouteUrl case Uri(scheme: 'zulip', host: 'notification')) {
+      final route = NotificationDisplayManager.routeForNotification(
+        context: context,
+        url: initialRouteUrl);
+
+      if (route != null) {
+        return [
+          HomePage.buildRoute(accountId: route.accountId),
+          route,
```
gnprice, line 169: nit:

    notif: Use associated account as initial account; if opened from background

semicolon ";" doesn't fit for this; use ","

gnprice (commented):
Thanks for the revision! This logic all looks good now — just a few nits, then this will be ready to merge.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

PR #1242, opened 2024-12-30
Mark as resolved/unresolved prep 1

chrisbobbe:
My local branch for #744 has gotten pretty long and I think it'll benefit from being split into multiple PRs. 🙂 Here's one.

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -766,14 +766,6 @@ class UpdateMessageEvent extends Event {
   Map<String, dynamic> toJson() => _$UpdateMessageEventToJson(this);
 }
 
-/// As in [UpdateMessageEvent.propagateMode].
-@JsonEnum(fieldRename: FieldRename.snake)
-enum PropagateMode {
-  changeOne,
-  changeLater,
-  changeAll;
-}
-
 /// A Zulip event of type `delete_message`: https://zulip.com/api/get-events#delete_message
 @JsonSerializable(fieldRename: FieldRename.snake)
 class DeleteMessageEvent extends Event {
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -664,6 +664,11 @@ enum MessageFlag {
 //   Using as a Map key is almost certainly a bug because it won't case-fold;
 //   see for example #739, #980, #1205.
 extension type const TopicName(String _value) {
+  /// The canonical form of the resolved-topic prefix.
+  // This is RESOLVED_TOPIC_PREFIX in web:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
+  static const resolvedTopicPrefix = '✔ ';
+
   /// The string this topic is identified by in the Zulip API.
   ///
   /// This should be used in constructing HTTP requests to the server,
@@ -844,11 +849,6 @@ enum MessageEditState {
   edited,
   moved;
 
-  // Adapted from the shared code:
-  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
-  // The canonical resolved-topic prefix.
-  static const String _resolvedTopicPrefix = '✔ ';
-
   /// Whether the given topic move reflected either a "resolve topic"
   /// or "unresolve topic" operation.
   ///
@@ -856,13 +856,13 @@ enum MessageEditState {
   /// but for purposes of [Message.editState], we want to ignore such renames.
   /// This method identifies topic moves that should be ignored in that context.
   static bool topicMoveWasResolveOrUnresolve(TopicName topic, TopicName prevTopic) {
-    if (topic.apiName.startsWith(_resolvedTopicPrefix)
-        && topic.apiName.substring(_resolvedTopicPrefix.length) == prevTopic.apiName) {
+    if (topic.apiName.startsWith(TopicName.resolvedTopicPrefix)
+        && topic.apiName.substring(TopicName.resolvedTopicPrefix.length) == prevTopic.apiName) {
       return true;
     }
 
-    if (prevTopic.apiName.startsWith(_resolvedTopicPrefix)
-        && prevTopic.apiName.substring(_resolvedTopicPrefix.length) == topic.apiName) {
+    if (prevTopic.apiName.startsWith(TopicName.resolvedTopicPrefix)
+        && prevTopic.apiName.substring(TopicName.resolvedTopicPrefix.length) == topic.apiName) {
       return true;
     }
 
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -669,6 +669,14 @@ extension type const TopicName(String _value) {
   //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
   static const resolvedTopicPrefix = '✔ ';
 
+  /// Pattern for an arbitrary resolved-topic prefix.
+  ///
+  /// These always begin with [resolvedTopicPrefix]
+  /// but can be weird and go on longer, like "✔ ✔✔ ".
+  // This is RESOLVED_TOPIC_PREFIX_RE in web:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts#L4-L12
+  static final resolvedTopicPrefixRegexp = RegExp(r'^✔ [ ✔]*');
+
   /// The string this topic is identified by in the Zulip API.
   ///
   /// This should be used in constructing HTTP requests to the server,
@@ -686,6 +694,14 @@ extension type const TopicName(String _value) {
 
   /// The key to use for "same topic as" comparisons.
   String canonicalize() => apiName.toLowerCase();
+
+  /// A [TopicName] with [resolvedTopicPrefixRegexp] stripped if present.
+  TopicName unresolve() =>
+    TopicName(_value.replaceFirst(resolvedTopicPrefixRegexp, ''));
+
+  /// Whether [this] and [other] have the same canonical form,
+  /// using [canonicalize].
+  bool isSameAs(TopicName other) => canonicalize() == other.canonicalize();
 
   TopicName.fromJson(this._value);
 
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -939,3 +939,13 @@ enum MessageEditState {
     return MessageEditState.none;
   }
 }
+
+/// As in [updateMessage] or [UpdateMessageEvent.propagateMode].
+@JsonEnum(fieldRename: FieldRename.snake, alwaysCreate: true)
+enum PropagateMode {
+  changeOne,
+  changeLater,
+  changeAll;
+
+  String toJson() => _$PropagateModeEnumMap[this]!;
+}
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -872,6 +872,14 @@ enum MessageEditState {
   /// but for purposes of [Message.editState], we want to ignore such renames.
   /// This method identifies topic moves that should be ignored in that context.
   static bool topicMoveWasResolveOrUnresolve(TopicName topic, TopicName prevTopic) {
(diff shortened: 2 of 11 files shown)


lib/api/route/messages.dart
```diff
@@ -158,6 +158,19 @@ class GetMessagesResult {
...
 
+/// The canonical resolved-topic prefix.
+// Adapted from the shared code:
+//   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
+const String kResolvedTopicPrefix = '✔ ';
+
+String stripResolvePrefixIfPresent(String topic) =>
+  topic.startsWith(kResolvedTopicPrefix)
+    ? topic.substring(kResolvedTopicPrefix.length)
+    : topic;
```
gnprice, line 169: How does this end up getting used later in your branch?

Its behavior differs from `unresolve_name` in the shared code:
https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts#L22-L29

which I'm guessing (haven't checked) is what gets applied when the user unresolves a topic in web or the legacy mobile app.

lib/widgets/action_sheet.dart
```diff
@@ -64,6 +64,16 @@ void _showActionSheet(
     });
 }
 
+/// A button in an action sheet.
+///
+/// When built from server data, the action sheet ignores changes in that data;
+/// we intentionally don't live-update the buttons on events.
+/// If a button's label, action, or position changes suddenly,
+/// it can be confusing and make the on-tap behavior unexpected.
+/// Better to let the user decide to tap
```
gnprice, line 73: Hmm, yeah.

Looking for `PerAccountStoreWidget.of` calls in this file, I think there's one place we don't currently follow this: the `hasSelfVote` in `ReactionButtons`.

That one shouldn't cause this problem in practice, though, because it changes by the user's own action. (Or by the message getting deleted, but then it doesn't matter whether the user tries to add a reaction or remove one.)

lib/widgets/action_sheet.dart
```diff
@@ -64,6 +64,16 @@ void _showActionSheet(
...
+/// based on information that's comfortably in their working memory,
+/// even if we sometimes have to explain (where we handle the tap)
+/// that that information has changed and they need to decide again.
```
gnprice, line 76: A key point here is that these races are always possible anyway — the change may have happened on the server but not reached us yet.

lib/api/model/model.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 
+import '../route/messages.dart';
```
gnprice, line 3: Hmm, I don't like having lib/api/model/ import from lib/api/route/. If it's needed for the model, that feels like a sign it's part of the model and not part of how particular routes interact with the model.

I think these can just stay in this file.

lib/api/model/events.dart
```diff
@@ -766,14 +767,6 @@ class UpdateMessageEvent extends Event {
   Map<String, dynamic> toJson() => _$UpdateMessageEventToJson(this);
 }
 
-/// As in [UpdateMessageEvent.propagateMode].
-@JsonEnum(fieldRename: FieldRename.snake)
-enum PropagateMode {
```
gnprice, line 771: And this can move from lib/api/model/events.dart to lib/api/model/model.dart, since it's no longer specific to an event.

lib/api/model/model.dart
gnprice: (When this file gets to feel unwieldy, I think the thing to do is to split pieces out by subject area: like a `lib/ap/model/message.dart`, akin to the `narrow.dart` and `submessage.dart` and so on that we already have.)

gnprice (commented):
Thanks! Good idea.

Comments below from an initial review.

lib/api/route/messages.dart
chrisbobbe: Updating to match web's behavior.

lib/widgets/action_sheet.dart
chrisbobbe: In a quick manual test, I don't even see the `hasSelfVote` live-updating in the message action sheet as I add or remove one of the popular emojis from the web app.

`hasSelfVote` uses `PerAccountStoreWidget.of` to get the self-user's ID. My user ID doesn't change when I add/remove a reaction. And although we're still signing up to refresh whenever the `PerAccountStore` notifies listeners, it doesn't notify listeners on `ReactionEvent`s. (Any `MessageListView`s do notify their listeners.)

chrisbobbe:

lib/api/model/model.dart
```diff
@@ -682,6 +695,29 @@ extension type const TopicName(String _value) {
...
+      return TopicName(_value.substring(match.end));
+    }
+  }
+
+  /// Whether [this] and [other] have the same canonical form,
+  /// using [canonicalize].
+  ///
+  /// If [ignoringResolvePrefix] is true,
+  /// the [unresolve]d forms of [this] and [other] are compared instead.
+  bool isSameAs(TopicName other, {required bool ignoringResolvePrefix}) {
```
gnprice, line 715: This feels to me more like two different methods than one method with a boolean flag.

And then I think the "ignoring" case is basically doing much the same job as its one caller `topicMoveWasResolveOrUnresolve`, so can be inlined into there.

lib/widgets/action_sheet.dart
gnprice: > In a quick manual test, I don't even see the `hasSelfVote` live-updating in the message action sheet as I add or remove one of the popular emojis from the web app.

Right, but then it's still your own action — you know that you just made that change (even though via a different client). So there isn't a risk of confusion.

lib/api/model/model.dart
```diff
@@ -915,3 +942,13 @@ enum MessageEditState {
...
 }
+
+/// As in [updateMessage] or [UpdateMessageEvent.propagateMode].
+@JsonEnum(fieldRename: FieldRename.snake, alwaysCreate: true)
+enum PropagateMode {
+  changeOne,
+  changeLater,
+  changeAll;
+
+  String toJson() =>_$PropagateModeEnumMap[this]!;
```
gnprice, line 953: nit:
```suggestion
  String toJson() => _$PropagateModeEnumMap[this]!;
```

lib/api/route/messages.dart
```diff
@@ -258,6 +258,39 @@ class SendMessageResult {
   Map<String, dynamic> toJson() => _$SendMessageResultToJson(this);
 }
 
+/// https://zulip.com/api/update-message
+Future<UpdateMessageResult> updateMessage(
+  ApiConnection connection, {
+  required int messageId,
+  String? topic,
```
gnprice, line 265:
```suggestion
  TopicName? topic,
```

:slightly_smiling_face:

gnprice (commented):
Thanks for the revision! Generally this all looks good. A few comments above, and a couple more below.

lib/api/model/model.dart
```diff
@@ -844,29 +880,20 @@ enum MessageEditState {
...
-        && prevTopic.apiName.substring(_resolvedTopicPrefix.length) == topic.apiName) {
+    if (topic.isSameAs(prevTopic, ignoringResolvePrefix: false)) { // TODO(log)
+      // Not a topic move after all… Return `true`
+      // just because `false` would put a marker for `MessageEditState.moved`
+      // on a message in the message list, which would be conspicuously wrong.
       return true;
     }
-
-    return false;
+    return topic.isSameAs(prevTopic, ignoringResolvePrefix: true);
```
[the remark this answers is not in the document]
chrisbobbe: Oh! Actually this isn't what web does. The code before matches what web does. Compare (zulip-flutter before vs. current web):

```dart
  static bool topicMoveWasResolveOrUnresolve(TopicName topic, TopicName prevTopic) {
    if (topic.apiName.startsWith(TopicName.resolvedTopicPrefix)
        && topic.apiName.substring(TopicName.resolvedTopicPrefix.length) == prevTopic.apiName) {
      return true;
    }

    if (prevTopic.apiName.startsWith(TopicName.resolvedTopicPrefix)
        && prevTopic.apiName.substring(TopicName.resolvedTopicPrefix.length) == topic.apiName) {
      return true;
    }

    return false;
  }
```

```ts
                // We know it has a topic edit. Now we need to determine if
                // it was a true move or a resolve/unresolve.
                if (
                    resolved_topic.is_resolved(edit_history_event.topic) &&
                    edit_history_event.topic.slice(2) === edit_history_event.prev_topic
                ) {
                    // Resolved.
                    resolve_toggled = true;
                    continue;
                }
                if (
                    resolved_topic.is_resolved(edit_history_event.prev_topic) &&
                    edit_history_event.prev_topic.slice(2) === edit_history_event.topic
                ) {
                    // Unresolved.
                    resolve_toggled = true;
                    continue;
                }
                // Otherwise, it is a real topic rename/move.
                moved = true;
```

For this "MOVED"-marker logic, there are pros and cons of following web:

- Pro: performance, as you mentioned
- Pro (I guess): You get a "MOVED" marker if the topic changes from like "✔ ✔ ✔ some topic" to "✔ ✔ some topic"
- Con: The "MOVED" marker is not suppressed on an actual resolve/unresolve, if the resolved-topic prefix is weird: "✔ ✔ some topic" to "some topic"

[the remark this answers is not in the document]
chrisbobbe: It looks like web uses the regex in just one place: to make an unresolve-topic API request.

chrisbobbe:
Thanks for the review! Revision pushed, and please see https://github.com/zulip/zulip-flutter/pull/1242#discussion_r1919302963.

lib/api/model/model.dart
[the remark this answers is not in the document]
gnprice: Cool. Yeah, let's stick with the simpler existing behavior which matches web, then.

The job of this "was resolve or unresolve" function is really just to paper over the fact that our "resolve topic" feature is implemented under the hood by renaming the topic. So it's fine that if someone edits the topic name only by fixing the letter case, then that will be displayed with a "MOVED" marker — the user who made that change indeed did so through a "move this topic" UI, just like for other times we show a "MOVED" marker.

gnprice:
Thanks! All looks good except that open subthread above.

lib/api/model/model.dart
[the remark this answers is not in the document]
chrisbobbe: Oh I see, yeah, that makes sense.

chrisbobbe:
Thanks for the review! Revision pushed, atop #1290 for the sake of CI.

test/api/model/model_test.dart
```diff
@@ -126,6 +126,46 @@ void main() {
...
+      void doTest(TopicName input, TopicName expected, {bool expectIdentity = false}) {
+        final output = input.unresolve();
+        check(output).apiName.equals(expected.apiName);
+        if (expectIdentity) {
+          check(output).identicalTo(expected);
+        }
+      }
+
+      doTest(eg.t('some topic'),       eg.t('some topic'),   expectIdentity: true);
+      doTest(eg.t('Some Topic'),       eg.t('Some Topic'),   expectIdentity: true);
```
gnprice, line 140: This ends up checking that the `'some topic'` on the left is identical to the `'some topic'` on the right. From the docs at [identical] I think that isn't guaranteed to be true.

Conversely, it seems like the point of this identity check is to test the `return this;` case in the implementation. But the test actually still passes if you remove that:
```dart
    if (match == null) {
      return TopicName(_value.substring(0));
    } else {
      return TopicName(_value.substring(match.end));
    }
```

or even replace the whole function with a `String.replaceFirst` call:
```dart
    return TopicName(_value.replaceAll(resolvedTopicPrefixRegexp, ''));
```

Let's simplify to the latter, and leave out the `identicalTo` check. We'll rely on the Dart stdlib to have the optimization where it doesn't unnecessarily copy the string in `replaceFirst` when there were no matches to replace.

gnprice (commented):
Thanks for the revision! One comment now on the details of `unresolve`.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

Issue #1172, opened 2024-12-18
downgraded schema does not always comply with the actual schema version


PIG208:
We try to handle the case when a schema downgrade happens:

https://github.com/zulip/zulip-flutter/blob/28b3536bc81b6665a6014ce72f40c51f83b0c5c4/lib/model/database.dart#L104-L115

This will work in dev because if you checkout and build code with an older schema version, the schema shipped with the build matches the version we are migrating to.

But we can't easily test this:

```dart
    test('downgrading', () async {
      final connection = await verifier.startAt(2);
      final db = AppDatabase(connection);
      await verifier.migrateAndValidate(db, 1);
      await db.close();
    });
```

<details>
<summary>result</summary>

```
package:drift_dev/src/services/schema/verifier_common.dart 25:5  verify
package:drift_dev/src/services/schema/verifier_impl.dart 44:5    VerifierImplementation.migrateAndValidate

Schema does not match
accounts:
 columns:
  additional:
   Contains the following unexpected entries: acked_push_token
```

</details>

This is because `m.createAll()` only has access to the latest schema, so while the version is supposedly v1, additional calls to drop the extra columns are needed.

A good step forward would be moving to [step-by-step migrations](https://drift.simonbinder.eu/migrations/step_by_step/). That won't fix this issue, but it offers a helpful way to access older schemas at specific versions. As this is only a dev-only feature, fixing this will be a low priority.

This feature was added in https://github.com/simolus3/drift/releases/tag/drift-2.10.0, long after we initially added `drift` as a dependency. It is among the many helpful things added since 2.5.2 (as of writing, we are at `2.19.1+1`, and the latest is `2.22.0`).

PIG208:
An issue closely related to this, is that switching from a branch that knows about tables to one that doesn't, will cause a schema downgrade to happen without removing those tables.

If you switch back to the branch with the new tables, attempts to upgrade will always fail, because the tables already exist. To make the app usable again, the database file has to be deleted. Perhaps we should do that by default when downgrading.

gnprice:
> To make the app usable again, the database file has to be deleted. Perhaps we should do that by default when downgrading.

Yeah, that'd more fully accomplish what that code is aimed at doing.

Given the context of this code — we've already opened the database when we learn what the schema version is — it might be annoying to arrange that smoothly, though. It might be easier to instead ask the database what all the tables are, and delete those.

gnprice:
[...]

PR #1248, opened 2025-01-03
Prepare schema migration/testing code for upcoming database changes

PIG208:
This prepares for #97 (PR #1167).

Fixes: #1172


lib/model/store.dart
```diff
@@ -778,19 +778,22 @@ class LiveGlobalStore extends GlobalStore {
...
-    //     on iOS, "Document directory" via https://developer.apple.com/documentation/foundation/nssearchpathdirectory/nsdocumentdirectory
-    //     on Linux, -> `xdg-user-dir DOCUMENTS` -> e.g. ~/Documents
-    //     That Linux answer is definitely not a fit.  Harder to tell about the rest.
+    // path_provider's getApplicationSupportDirectory:
+    //   on Android, -> Flutter's PathUtils.getFilesDir -> https://developer.android.com/reference/android/content/Context#getFilesDir()
+    //     -> empirically /data/data/com.zulip.flutter/files/
+    //   on iOS, -> "Library/Application Support" via https://developer.apple.com/documentation/foundation/nssearchpathdirectory/nsapplicationsupportdirectory
+    //   on Linux, -> "${XDG_DATA_HOME:-~/.local/share}/com.zulip.flutter/"
+    //
+    // This is reasonable for iOS per Apple's recommendation:
```
chrisbobbe, line 787: How about "This is the best fit for iOS per Apple's recommendation", if that's true? That would feel more satisfying than just saying it's a reasonable option, because then you wonder if there are other reasonable options and what those look like.

Same for Android and Linux below, and I don't yet follow your reasoning about Android. (Linux is clear from the removed line "That Linux answer is definitely not a fit.")

lib/model/database.dart
```diff
@@ -53,8 +53,9 @@ class AppDatabase extends _$AppDatabase {
 
   // When updating the schema:
   //  * Make the change in the table classes, and bump schemaVersion.
-  //  * Export the new schema and generate test migrations:
+  //  * Export the new schema and generate test migrations with drift:
   //    $ tools/check --fix drift
+  //    and generate database code with build_runner.
```
chrisbobbe, line 60: Is there a specific command we can put here, or perhaps refer the reader to some documentation, even if it's just `dart run build_runner --help`? (The `dart run` seems helpful to a contributor arriving here for the first time.)

pubspec.yaml
```diff
@@ -40,7 +40,7 @@ dependencies:
   convert: ^3.1.1
   crypto: ^3.0.3
   device_info_plus: ^11.2.0
-  drift: ^2.5.0
+  drift: ^2.23.0
```
chrisbobbe, line 43: Need help checking if podfiles need an update; deps: Upgrade drift to 2.23.0

Confirmed by running `tools/upgrade pod` at this commit, on a Mac, that no changes to `ios/Podfile.lock` or `macos/Podfile.lock` are needed.

chrisbobbe (commented):
Thanks! Small comments below, and I'll defer to Greg for a more detailed review of the substantive changes, as I'm not yet up-to-speed on this system. 🙂

pubspec.yaml
PIG208: Thanks for confirming@

lib/model/store.dart
PIG208: Will put the relevant pieces back here. I dug into the path for Android more, and found that there is [`getDatabasePath`](https://developer.android.com/reference/android/content/Context#getDatabasePath(java.lang.String)), which is not provided by path_provider. It would be nice to have support for that, though.

PIG208: Opened #1321 and refer to the issue instead. We should be good with using `getApplicationSupportDirectory` for this PR.

PIG208: Will be skipping this commit after [a CZO discussion](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.23F1248.20database.20path.20for.20Android).

PIG208:
Rebased on top of #1324 for CI, and adjust the commit message so that this is not really about upgrading drift, but rather utilizing new drift features.

chrisbobbe:
Thanks for the revision! Marking for Greg's review:

> I'll defer to Greg for a more detailed review of the substantive changes, as I'm not yet up-to-speed on this system. 🙂

lib/model/database.dart
```diff
@@ -52,30 +49,42 @@ class UriConverter extends TypeConverter<Uri, String> {
...
+    // This should only ever happen in dev.  As a dev convenience,
+    // drop everything from the database and start over.
+    await m.database.transaction(() async {
+      final query = m.database.customSelect(
+        "SELECT name FROM sqlite_master WHERE type='table'");
+      for (final row in await query.get()) {
+        final data = row.data;
+        final tableName = data['name'] as String;
+        // Skip sqlite-internal tables, https://www.sqlite.org/fileformat2.html#intschema
+        // https://github.com/simolus3/drift/blob/0901c984a987e54ba0b67e227adbfdcf3c08eeb4/drift_dev/lib/src/services/schema/verifier_common.dart#L9-L22
```
gnprice, line 78: nit: I'm not sure how to interpret a comma here. How about:
```suggestion
        // Skip sqlite-internal tables.  See for comparison:
        //   https://www.sqlite.org/fileformat2.html#intschema
        //   https://github.com/simolus3/drift/blob/0901c984a/drift_dev/lib/src/services/schema/verifier_common.dart#L9-L22
```

gnprice: (That bit of Drift code is actually a bit puzzling:
```dart
/// Attempts to recognize whether [name] is likely the name of an internal
/// sqlite3 table (like `sqlite3_sequence`) that we should not consider when
/// comparing schemas.
bool isInternalElement(String name, List<String> virtualTables) {
  // Skip sqlite-internal tables, https://www.sqlite.org/fileformat2.html#intschema
  if (name.startsWith('sqlite_')) return true;
```

The example in the doc would *not* get recognized by the actual implementation. Unless it happens to be in the `virtualTables`, but I don't think that's what that argument is for.)

lib/model/database.dart
```diff
@@ -52,30 +49,42 @@ class UriConverter extends TypeConverter<Uri, String> {
...
   //    $ tools/check --fix drift
+  //    and generate database code with build_runner.
+  //    See ../../README.md#generated-files for more
+  //    information on using the build_runner.
   //  * Write a migration in `onUpgrade` below.
   //  * Write tests.
   @override
   int get schemaVersion => 2; // See note.
 
+  Future<void> _resetDatabase(Migrator m) async {
```
gnprice, line 68: Testing for this is blocked until #1172.

Why does something block testing this?

It seems like we should be able to test it like so:
* Make a database that has the current schema, plus some arbitrary extra column on accounts and some arbitrary extra table. And has a version greater than the current version.
* Attempt a migration.
* Check that the schema now is the correct one.

gnprice: I guess this function doesn't exactly meet the spec of `onUpgrade`, because it assumes `to` is the current latest schema version. That's what #1172 is about.

But that doesn't seem like it's an obstacle for testing that it correctly does the thing it's intended to do, which is to correctly migrate given that assumption.

gnprice (commented):
Thanks @PIG208 for working on this, and @chrisbobbe for the previous review!

Comments below.

I think the PR title also rather undersells what this is about. "Using new features" isn't particularly appealing in itself — lots of our dependencies are adding new features all the time, and most of them aren't anything we want.

The real point, I think, is that this refactors our handling of schema migrations, to better set us up for doing more migrations in the future. And then in order to do that, it relies on some features that Drift didn't yet have in 2023 when this migrations code was first written.

lib/model/database.dart
PIG208: Hmm, good catch. I believe that is a typo. The sqlite.org link does mention a table called `sqlite_sequence`, and nothing relevant comes up in my Google search for `"sqlite3_sequence"`.

test/model/database_test.dart
```diff
@@ -98,37 +98,62 @@ void main() {
...
+    // tables or columns typically don't need these advanced tests. For more
+    // information, see https://drift.simonbinder.eu/migrations/tests/#verifying-data-integrity
+    group('migrate with data', () {
+      test('upgrade to v2', () async {
+        late final v1.AccountsData oldAccountData;
+        await verifier.testWithDataIntegrity(
+          oldVersion: 1, createOld: v1.DatabaseAtV1.new,
+          newVersion: 2, createNew: v2.DatabaseAtV2.new,
+          openTestedDatabase: AppDatabase.new,
+          createItems: (batch, oldDb) async {
```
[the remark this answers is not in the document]
PIG208: The generated test looks like this:

<details>
<summary>test/drift/default/migration_test.dart</summary>

```dart
// dart format width=80
// ignore_for_file: unused_local_variable, unused_import
import 'package:drift/drift.dart';
import 'package:drift_dev/api/migrations_native.dart';
import 'package:zulip/model/database.dart';
import 'package:test/test.dart';
import 'generated/schema.dart';

import 'generated/schema_v1.dart' as v1;
import 'generated/schema_v2.dart' as v2;

void main() {
  driftRuntimeOptions.dontWarnAboutMultipleDatabases = true;
  late SchemaVerifier verifier;

  setUpAll(() {
    verifier = SchemaVerifier(GeneratedHelper());
  });

  group('simple database migrations', () {
    // These simple tests verify all possible schema updates with a simple (no
    // data) migration. This is a quick way to ensure that written database
    // migrations properly alter the schema.
    final versions = GeneratedHelper.versions;
    for (final (i, fromVersion) in versions.indexed) {
      group('from $fromVersion', () {
        for (final toVersion in versions.skip(i + 1)) {
          test('to $toVersion', () async {
            final schema = await verifier.schemaAt(fromVersion);
            final db = AppDatabase(schema.newConnection());
            await verifier.migrateAndValidate(db, toVersion);
            await db.close();
          });
        }
      });
    }
  });

  // The following template shows how to write tests ensuring your migrations
  // preserve existing data.
  // Testing this can be useful for migrations that change existing columns
  // (e.g. by alterating their type or constraints). Migrations that only add
  // tables or columns typically don't need these advanced tests. For more
  // information, see https://drift.simonbinder.eu/migrations/tests/#verifying-data-integrity
  // TODO: This generated template shows how these tests could be written. Adopt
  // it to your own needs when testing migrations with data integrity.
  test("migration from v1 to v2 does not corrupt data", () async {
    // Add data to insert into the old database, and the expected rows after the
    // migration.
    // TODO: Fill these lists
    final oldAccountsData = <v1.AccountsData>[];
    final expectedNewAccountsData = <v2.AccountsData>[];

    await verifier.testWithDataIntegrity(
      oldVersion: 1,
      newVersion: 2,
      createOld: v1.DatabaseAtV1.new,
      createNew: v2.DatabaseAtV2.new,
      openTestedDatabase: AppDatabase.new,
      createItems: (batch, oldDb) {
        batch.insertAll(oldDb.accounts, oldAccountsData);
      },
      validateItems: (newDb) async {
        expect(
            expectedNewAccountsData, await newDb.select(newDb.accounts).get());
      },
    );
  });
}
```

</details>

will also include the detailed steps to generate in the commit message.

test/model/database_test.dart
```diff
@@ -137,6 +137,7 @@ void main() {
         ...accountV1.toJson(),
         'ackedPushToken': null,
       });
+      await after.close();
```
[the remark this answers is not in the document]
gnprice: Great, thanks. Yeah, just mention in the commit message that this otherwise leaks state, and would produce a warning if we had another test after this that also created a database instance.

lib/model/database.dart
```diff
@@ -85,24 +94,18 @@ class AppDatabase extends _$AppDatabase {
...
-        if (from < 2 && 2 <= to) {
-          await m.addColumn(accounts, accounts.ackedPushToken);
-        }
-        // New migrations go here.
-      }
-    );
+        await m.runMigrationSteps(from: from, to: to,
+          steps: migrationSteps(
+            from1To2: (m, schema) async {
+              await m.addColumn(schema.accounts, schema.accounts.ackedPushToken);
```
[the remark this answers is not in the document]
PIG208: Yeah, and the commit message can be clearer if we move the step-by-step helper change to an NFC commit.

test/model/database_test.dart
```diff
@@ -98,37 +98,62 @@ void main() {
...
-        ...accountV1.toJson(),
-        'ackedPushToken': null,
+    // Testing this can be useful for migrations that change existing columns
+    // (e.g. by alterating their type or constraints). Migrations that only add
+    // tables or columns typically don't need these advanced tests. For more
+    // information, see https://drift.simonbinder.eu/migrations/tests/#verifying-data-integrity
+    group('migrate with data', () {
+      test('upgrade to v2', () async {
+        late final v1.AccountsData oldAccountData;
+        await verifier.testWithDataIntegrity(
```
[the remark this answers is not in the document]
PIG208: Ended up dropping the change after https://github.com/zulip/zulip-flutter/pull/1248#discussion_r1960966988.

I too agree that the helper is verbose and this part is mostly NFC.

lib/model/database.dart
PIG208: The test ended up being a bit more complicated, but it turned out quite doable. We can fix #1172 at the same time.

PIG208:
Thanks for the review! This is ready again now.

PIG208:

gnprice:
Thanks! Looks good; merging.

Issue #1247, opened 2025-01-02
msglist: Move star icon 2px away from the edge of the screen


chrisbobbe:
We've had some reactions in [beta feedback](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/stars.20and.20message.20layout/near/2015071) that the star icon is too close to the edge of the screen:

<img width=400 src="https://github.com/user-attachments/assets/5ccf999c-a734-40fa-817d-115df69f06c6">

We've decided (see linked discussion) to move the star 2px inward while leaving the message-content margin unchanged (16px). This means there might be some overlap between the text and the star, but that seems rare and probably not critical. (We can think more about this if we do get reports that it's a problem.)

chrisbobbe:
For the implementation, perhaps we want a [`Stack`](https://api.flutter.dev/flutter/widgets/Stack-class.html) with a [`Positioned`](https://api.flutter.dev/flutter/widgets/Positioned-class.html) child.

lakshya1goel:
Hi, I want to work on this issue.

lakshya1goel:
Hey, I have raised a PR for this issue please review it.
Thanks!

yashrajpu:
Please assign this issue to me. I have already solved it, I just need to commit the changes. Thank you!

PR #1249, opened 2025-01-03
msglist: Move star icon 2px away from the edge of the screen

lakshya1goel:
Fixes: #1247

| WhatsApp Image 2025-01-03 at 9 21 11 AM | WhatsApp Image 2025-01-03 at 9 21 11 AM(1) |

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -3,7 +3,7 @@ import 'dart:math';
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_color_models/flutter_color_models.dart';
-import 'package:intl/intl.dart';
+import 'package:intl/intl.dart' hide TextDirection;
 
 import '../api/model/model.dart';
 import '../generated/l10n/zulip_localizations.dart';
@@ -1387,6 +1387,17 @@ class MessageWithPossibleSender extends StatelessWidget {
       case MessageEditState.none:
     }
 
+    Widget? star;
+    if (message.flags.contains(MessageFlag.starred)) {
+      final starOffset = switch (Directionality.of(context)) {
+        TextDirection.ltr => -2.0,
+        TextDirection.rtl => 2.0,
+      };
+      star = Transform.translate(
+        offset: Offset(starOffset, 0),
+        child: Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star));
+    }
+
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
       onLongPress: () => showMessageActionSheet(context: context, message: message),
@@ -1418,9 +1429,7 @@ class MessageWithPossibleSender extends StatelessWidget {
                           context, 0.05, baseFontSize: 12))),
                 ])),
               SizedBox(width: 16,
-                child: message.flags.contains(MessageFlag.starred)
-                  ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
-                  : null),
+                child: star),
             ]),
         ])));
   }


lib/widgets/message_list.dart
```diff
@@ -1354,32 +1354,37 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                        textAlign: TextAlign.end,
+                        style: TextStyle(
+                          color: designVariables.labelEdited,
+                          fontSize: 12,
+                          height: (12 / 12),
+                          letterSpacing: proportionalLetterSpacing(
+                            context, 0.05, baseFontSize: 12))),
+                  ])),
+              if (message.flags.contains(MessageFlag.starred))
+                Positioned(
```
chrisbobbe, line 1378: We want [`PositionedDirectional`](https://api.flutter.dev/flutter/widgets/PositionedDirectional-class.html) here.

lib/widgets/message_list.dart
```diff
@@ -1354,32 +1354,37 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                  top: 4,
```
chrisbobbe, line 1379: Doesn't this affect the vertical position? I don't think we want that.

lib/widgets/message_list.dart
```diff
@@ -1354,32 +1354,37 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-                        color: designVariables.labelEdited,
-                        fontSize: 12,
-                        height: (12 / 12),
-                        letterSpacing: proportionalLetterSpacing(
-                          context, 0.05, baseFontSize: 12))),
-                ])),
-              SizedBox(width: 16,
-                child: message.flags.contains(MessageFlag.starred)
-                  ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
-                  : null),
```
chrisbobbe, line 1381: Can we make the change more locally, by only replacing this `SizedBox`? I'm not sure we should need to touch anything higher than this in the widget tree, like the `Row` etc.

chrisbobbe (changes requested):
Thanks! Comments below.

lib/widgets/message_list.dart
lakshya1goel: Can't we simply add a right padding of 2 px for the sizedbox? 
```
Padding(
  padding: const EdgeInsets.only(right: 2),
  child: SizedBox(width: 16,
    child: message.flags.contains(MessageFlag.starred)
      ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
      : null),
),
```
I think there is no chance of overlapping of content and star icon because of this sizedbox.

chrisbobbe: This would be a change in the design; would you propose it in the [design discussion](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/stars.20and.20message.20layout/near/2015071)?

Specifically, this proposal would remove 2px of the width currently available for message content, for starred and unstarred messages. That doesn't seem ideal, actually—I'd still be interested to see an implementation of the original design, just without changes that seem like they might not be necessary: https://github.com/zulip/zulip-flutter/pull/1249#discussion_r1919380764

lakshya1goel: Thank you for the feedback! I've adjusted the approach to address this. Instead of making changes higher up in the widget tree (like the Row), I've made the change more localized by replacing the SizedBox logic directly.
Using `Transform.translate`, the adjustment is now applied directly to the Icon, ensuring the widget tree above remains unaffected. This way, we don't unnecessarily modify the available width for message content. 

Let me know if this works better, or if there's anything else you'd like to refine further!

lib/widgets/message_list.dart
```diff
@@ -1381,7 +1381,9 @@ class MessageWithPossibleSender extends StatelessWidget {
                 ])),
               SizedBox(width: 16,
                 child: message.flags.contains(MessageFlag.starred)
-                  ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
+                  ? Transform.translate(
+                    offset: const Offset(-2,0),
```
chrisbobbe, line 1385: When the UI is showing in a right-to-left (RTL) language, this will move the star in the wrong direction:

image

I don't see an `OffsetDirectional` (on the pattern of [`AlignmentDirectional`](https://api.flutter.dev/flutter/painting/AlignmentDirectional-class.html), [`EdgeInsetsDirectional`](https://api.flutter.dev/flutter/painting/EdgeInsetsDirectional-class.html), etc.), so I think the solution is to write a short conditional on the value of [`Directionality.of(context)`](https://api.flutter.dev/flutter/widgets/Directionality/of.html).

lib/widgets/message_list.dart
```diff
@@ -1381,7 +1381,9 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+                    child: Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star))
```
chrisbobbe, line 1386: Also please check that the indentation and spacing matches the patterns we follow in other code.

lakshya1goel:
Thanks for the review @chrisbobbe, pushed the required changes, PTAL.

lakshya1goel:
Hi @chrisbobbe , mentioning you for follow up, PTAL whe you get some time. Thanks!

lib/widgets/message_list.dart
```diff
@@ -4,7 +4,7 @@ import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_color_models/flutter_color_models.dart';
 import 'package:intl/intl.dart';
-
+import 'package:flutter/widgets.dart' as flutter;
```
chrisbobbe, line 7: Instead of this, I would change the `package:intl` line to:

```dart
import 'package:intl/intl.dart' hide TextDirection;
```

lib/widgets/message_list.dart
```diff
@@ -1381,8 +1381,15 @@ class MessageWithPossibleSender extends StatelessWidget {
...
-                  : null),
+                    ? Transform.translate(
+                        offset: Offset(
+                          Directionality.of(context) == flutter.TextDirection.rtl ? 2 : -2,
+                          0,
+                        ),
+                        child: Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
+                      )
+                    : null,
+              ),
```
chrisbobbe, line 1392: I would format it this way (note indentation and exhaustive `switch`):

```dart
              SizedBox(width: 16,
                child: message.flags.contains(MessageFlag.starred)
                  ? Transform.translate(
                      offset: Offset(
                        switch (Directionality.of(context)) {
                          TextDirection.ltr => -2,
                          TextDirection.rtl => 2,
                        },
                        0,
                      ),
                      child: Icon(size: 16, color: designVariables.star,
                        ZulipIcons.star_filled))
                  : null),
```

chrisbobbe (commented):
Thanks! Small comments below.

lib/widgets/message_list.dart
```diff
@@ -1403,7 +1402,16 @@ class MessageWithPossibleSender extends StatelessWidget {
...
                 child: message.flags.contains(MessageFlag.starred)
-                  ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
+                  ? Transform.translate(
+                      offset: Offset(
+                        switch (Directionality.of(context)) {
+                          TextDirection.ltr => -2,
+                          TextDirection.rtl => 2,
+                        },
+                        0,
+                      ),
```
gnprice, line 1412: This logic gets kind of complicated for a single expression. Let's move it out into statement form (a) so that it can use a local variable or two, and (b) to make it less deeply indented.

To do that, let's have a local variable like `Widget? star;` — it'll be either null if we don't want to show a star, or the whole `Transform.translate(…)` widget if we do.

Then the `switch` expression can be pulled out as a local variable within the code that computes that.

gnprice (commented):
Thanks to you both! The logic here looks good. A couple of comments below.

lakshya1goel:
Thanks for the review @gnprice, pushed the revision, PTAL!

lib/widgets/message_list.dart
```diff
@@ -1417,10 +1430,10 @@ class MessageWithPossibleSender extends StatelessWidget {
...
                           context, 0.05, baseFontSize: 12))),
                 ])),
-              SizedBox(width: 16,
-                child: message.flags.contains(MessageFlag.starred)
-                  ? Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
-                  : null),
+              SizedBox(
+                width: 16,
+                child: star,
+              ),
```
gnprice, line 1436: nit: keep compact, following the pattern of the existing code:
```suggestion
              SizedBox(width: 16,
                child: star),
```

lib/widgets/message_list.dart
```diff
@@ -1387,6 +1387,19 @@ class MessageWithPossibleSender extends StatelessWidget {
       case MessageEditState.none:
     }
 
+    final starOffset = switch (Directionality.of(context)) {
+      TextDirection.ltr => -2.0,
+      TextDirection.rtl => 2.0,
+    };
```
gnprice, line 1393: nit: move this inside the if-block — it's only meaningful within that case

lib/widgets/message_list.dart
```diff
@@ -1387,6 +1387,19 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+
+    Widget? star;
+    if (message.flags.contains(MessageFlag.starred)) {
+      star = Transform.translate(
+        offset: Offset(starOffset, 0),
+        child: Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star)
+      );
```
gnprice, line 1400: nit: similarly:
```suggestion
        child: Icon(ZulipIcons.star_filled, size: 16, color: designVariables.star));
```

gnprice (commented):
Thanks! A couple of small comments still.

lakshya1goel:
Pushed the revision, PTAL. Thanks!

gnprice:
Thanks! Looks good; merging.

Issue #385, opened 2023-11-17
RecentDmConversationsPage: Design / implement appearance when empty


chrisbobbe:
When the user doesn't have any recent DM conversations, `RecentDmConversationsPage` (titled "Direct messages") should have a message saying so, instead of just being blank:

<img width=400 src="https://github.com/zulip/zulip-flutter/assets/22248748/15d1f869-d162-42fa-b509-385787ad350f">

### Design

We have a design for this in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5956-115842&t=dosxggnU7UlsrbfI-0

Screenshot:
![image](https://github.com/user-attachments/assets/8157a040-c97d-4a57-a5d6-704f5b5860d3)

For details, see the Figma doc.

The design relies on having a way to start a new DM conversation:
* #127 

So this issue won't be ready to work on until that one is done.

### Related issues
- #386
- #127

chrisbobbe:
Once we do this (or #386, for the inbox), let's revisit `SubscriptionListPage` (titled "Streams") to make sure it looks consistent with what we do here. Here's what that page currently does:

<img width=400 src="https://github.com/zulip/zulip-flutter/assets/22248748/c596c017-0e31-404e-b915-55e5611c97f2">

sm-sayedi:
I would like to work on this issue!

So far that I have worked on it, it just shows a message like this at the center of the page:

<img src="https://github.com/zulip/zulip-flutter/assets/59946442/718771c7-c969-4f9f-af6f-c3647b6acb99" width="50%" />

chrisbobbe:
Thanks for your interest, @sm-sayedi! I think this issue isn't ripe to work on yet, though. We'll want our designer to suggest how this should look.

alya:
We should keep the text consistent with the web app, unless there's a good reason for them to be different:

> You have no direct messages yet! Why not start the conversation?

The rest of the visual design can be as in Figma.

chrisbobbe:
I've split #1551 off from this, for some polish in the Figma (graphics), and sent #1552 for a text-only placeholder, marking that as fixing this issue.

Issue #127, opened 2023-05-26
Support starting a new DM conversation


chrisbobbe:
This corresponds to the "New DM" and "New group DM" buttons which zulip-mobile has at the top of the DM-conversations screen.

### Design
The design for this UI is in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4923-3308&t=dosxggnU7UlsrbfI-0

Here's one screenshot:
![image](https://github.com/user-attachments/assets/39dd1de9-ec18-4164-8825-66c6ae5a897a)

For details, see the Figma doc.

### Related issues
* An issue that will rely on this one: #385
* A previous phase: #119

Khader-1:
Do we have any design expectations here?

chrisbobbe:
I expect we'll want to wait for our designer to propose a design for this feature, yes.

gnprice:
We have a design for this feature now; I've updated the description.

Dhanesh-Sawant:
I would like to work on this issue

I have started it with adding the design when there are no dm's yet as required in [#385 ](385)

![WhatsApp Image 2025-01-02 at 08 39 48](https://github.com/user-attachments/assets/f943ad74-5221-4f43-b7da-86d25421c756)

Dhanesh-Sawant:
this is just the test for search functionality in direct messages, the search box design is just for testing now & will finalize it once I finish doing for groups also
 
https://github.com/user-attachments/assets/fddda99a-053d-4f98-80cd-435ab6ec0f8c

Dhanesh-Sawant:
Here I have implemented the functionality of starting a conversation with a person or in a group after clicking new dm button.
- be able to search for members
- add members to the selected users required for forming a group almost same as the required design(some changes are needed to be done further as this was just for testing).


https://github.com/user-attachments/assets/2e6c0be2-df97-4ec9-a772-60534dccca9d

chimnayajith:
The PR for this seems to be abandoned. I'd like to work on this issue.

Dhanesh-Sawant:
hey @chimnayajith, I was caught up by something else, but I've started working on it again..

PR #1252, opened 2025-01-03
new dm : page to start a new dm conversation

Dhanesh-Sawant:
This intents to add new dm functionality as mentioned in #127 & #385 

1. When the user doesn't have any recent DM conversations, RecentDmConversationsPage this UI gets showed. I have added the 2 svg's for it and placed it according to the required design.

WhatsApp Image 2025-01-02 at 08 39 48
 
2. First the search functionality within the recent dm's, here the search icon in the design is in the appbar, but I think having a separate search bar would be much better and looks very obvious for the users. other than that the margin for the new DM button is properly set by me also with the case of whether it should be sticked or moved up with the keyboard would be taken care by in the next change,  

3. I have used wrap with SingleChildScrollView with chip for displaying the selected users for staring a dm, an addition here is that there is a small cross button also in the chip when pressed will deselect that user, then after adding all desired users there is a Next button on the appbar when pressed will take it to their messageList.

fourth
third
second
first

 kindly suggest the required changes and review my suggestion of search box and cross button is selected user chip.


gnprice:
Thanks for the contribution. Before we can review this, it will need tests. See our README:
https://github.com/zulip/zulip-flutter?tab=readme-ov-file#submitting-a-pull-request

From a quick look at the screenshots, the UI doesn't appear to match the design in Figma. Please take a close look at the details of padding, spacing, color, and other visual elements, and try to exactly match what's in Figma. If there are aspects you don't match or that you're unsure about, please flag them explicitly in the PR.

Dhanesh-Sawant:
@gnprice 

I suggest keeping the search functionality directly within the RecentDmConversationsPageBody page, with the search bar always visible, rather than toggling it from the app bar. This would provide better UX by making the search capability immediately discoverable like in whatsapp, This approach would eliminate the complexity of using GlobalKey and state synchronization as we have our appbar in HomePage.

Dhanesh-Sawant:
image

also i didnt get what''s the use of check_circle/circle_outlined button here, as i thought that once on clicking the ListTile it will get selected and will get appended on the top selected users row.

Dhanesh-Sawant:
https://github.com/user-attachments/assets/dc648f35-f836-48ef-9aa7-5ab631d8d5c1

also needed a clarification here that the new dm button is jumping up when the keyboard opens, so do we need like that or we need to make it stick to its position or it should be just below the last dm row, as it is not that clear from the design.

image

gnprice:
> I suggest keeping the search functionality directly within the RecentDmConversationsPageBody page, with the search bar always visible, rather than toggling it from the app bar.

> I also recommend using the Chip widget for selected users

The main thing this issue is calling for is to implement the UX design we have that's expressed in Figma, not to design a new UX.

There might be details we revise after seeing an implementation. But the starting point should be to implement the design accurately.


> This approach would eliminate the complexity of using GlobalKey and state synchronization as we have our appbar in HomePage.

If there's something you're running into in the implementation that feels more complex than it should be, the #mobile-dev-help channel is a good place to ask. You'll need to give more details of what it is you're doing and what you find is a challenge with it, and then we can discuss solutions.


> also needed a clarification here that the new dm button is jumping up when the keyboard opens, so do we need like that or we need to make it stick to its position or it should be just below the last dm row, as it is not that clear from the design.

This would be a good question to ask in the #mobile-design channel. See this section of our docs:
https://zulip.readthedocs.io/en/latest/contributing/design-discussions.html#guidelines-for-code-contributors
(but for mobile issues and PRs, use the #mobile-design channel rather than #design).

Dhanesh-Sawant:
should I continue to contribute to this issue as https://github.com/zulip/zulip-flutter/pull/1322, this pr also addresses the same issue?

gnprice:
Yeah, probably best to find a different issue to work on, since that PR has gotten further.

I'll close this PR to get it out of our queue. The code will remain available to refer to, or to reopen the PR, if in the future we end up returning to it.

PR #1257, opened 2025-01-07
home: Stop assuming account existence from loading page

PIG208:
This rewrite ended being a bit more substantial than just switching both
accounts initial routes to the result of MessageListPage.buildRoute().

We do not use popUtil because the navigator stack normally should not
become empty, so the HomePage route for account1 stays. Additionally,
because we are pushing a different page route, we no longer need to set
up different messages as the discriminator, further simplifying the
test.

This is rebased on top of #1235, and taken from #1183.


chrisbobbe:

test/widgets/actions_test.dart
```diff
@@ -174,50 +167,42 @@ void main() {
...
       await tester.pump();
+      final account1Route = pushedRoutes.single;
+      check(account1Route).isA<WidgetRoute>().page.isA<HomePage>();
+
+      final account2Connection =
+        (await testBinding.globalStore.perAccount(account2.id)).connection as FakeApiConnection;
+      account2Connection.prepare(json: eg.newestGetMessagesResult(
+        foundOldest: true, messages: []).toJson());
+      final account2Route = MessageListPage.buildRoute(
+        accountId: account2.id, narrow: const CombinedFeedNarrow());
```
gnprice, line 183: This still doesn't seem realistic — how would you get to a CombinedFeedNarrow for account 2 atop a HomePage for account 1?

gnprice (commented):
Thanks. Some initial comments below.

This feels surprisingly complex. Let's hold off investing further engineering into it until we've completed #1183, which seems to have overlapping work and is a priority.

test/widgets/actions_test.dart
PIG208: How about using `openNotification` to get a `MessageListPage` route with a `TopicMessage` narrow?

test/widgets/actions_test.dart
```diff
@@ -174,50 +167,42 @@ void main() {
...
-      check(findLoadingPage).findsNothing();
+      check(findAccount2PageContent).findsOne();
 
-      final removedRoutes = <Route<dynamic>>[];
+      final removedRoutes = <Route<void>>[];
       testNavObserver.onRemoved = (route, prevRoute) => removedRoutes.add(route);
 
-      final context = tester.element(find.byType(MaterialApp));
-      final future = logOutAccount(context, account1.id);
+      final future = logOutAccount(testBinding.globalStore, account1.id);
```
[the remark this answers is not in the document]
PIG208: Ended up leaving this unchanged and access the context the old way, because we no longer need to use the navigator.

PIG208:
Updated the PR to make `openNotification` public, dropped the `logOutAccount` refactor, and use notification to set up the navigator stack.

PIG208:
Marking this as draft until we revisit this topic, so as to prioritize launch tasks.

Issue #1204, opened 2024-12-26
ci: Reinstate `--depth=1000` in cloning Flutter, once Flutter has a new tag


chrisbobbe:
Flutter `main` now has well over a thousand commits since the most recent tag:

```diff
diff --git pubspec.yaml pubspec.yaml
index 780c05765..c021e8d1e 100644
--- pubspec.yaml
+++ pubspec.yaml
@@ -14,8 +14,8 @@ environment:
   # We use a recent version of Flutter from its main channel, and
   # the corresponding recent version of the Dart SDK.
   # Feel free to update these regularly; see README.md for instructions.
-  sdk: '>=3.7.0-224.0.dev <4.0.0'
-  flutter: '>=3.28.0-1.0.pre.35'  # 1c1f35711be76d1d2a9f5e3a730747591e64e14a
+  sdk: '>=3.7.0-267.0.dev <4.0.0'
+  flutter: '>=3.28.0-2.0.pre.38575'  # 65ff060283e19423e9538c18c24e44495b70aeff
 
 # To update dependencies, see instructions in README.md.
 dependencies:
```

These are [migrated engine commits](https://github.com/zulip/zulip-flutter/pull/1184#issuecomment-2560439941) from the merge of flutter/engine into the flutter/flutter monorepo.

Anyway, when we do a shallow clone with `--depth=1000`, CI can't see the most recent tag, which is needed:

```dart
      # […] we need the most recent tag
      # so that Flutter knows its version and sees the constraint in our
      # pubspec is satisfied. […]
```

So we'll be temporarily removing the `--depth=1000`, to do a full clone instead of a shallow clone. See discussion: https://github.com/zulip/zulip-flutter/pull/1184#issuecomment-2560440790

That change will slow down CI, so this issue is to reinstate the `--depth=1000` once Flutter makes a new tag.

PR #1262, opened 2025-01-08
github: Fix weblate action's rights, and update generated code.

alexmv:
This replaces #1192, with the branch pushed to `zulip/zulip-flutter` so that it can possibly be tested before merging.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -1,4 +1,7 @@
 name: Update translations from Weblate
+permissions:
+  contents: write
+  pull-requests: write
 on:
   schedule:
     - cron: "0 10 * * 1"
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -6,17 +6,44 @@ on:
   schedule:
     - cron: "0 10 * * 1"
   workflow_dispatch:
+
 jobs:
   update-translations:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
       - name: Fetch and merge from Weblate
         # The commit message is generated in Weblate; see https://hosted.weblate.org/addon/17163/
         run: |
           git remote add weblate https://hosted.weblate.org/git/zulip/zulip-flutter/
           git fetch weblate
           git merge --ff-only weblate/main
+
+      - name: Clone Flutter SDK
+        # We can't do a depth-1 clone, because we need the most recent tag
+        # so that Flutter knows its version and sees the constraint in our
+        # pubspec is satisfied.  It's uncommon for flutter/flutter to go
+        # more than 100 commits between tags.  Fetch 1000 for good measure.
+        run: |
+          git clone --depth=1000 -b main https://github.com/flutter/flutter ~/flutter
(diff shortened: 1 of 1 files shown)


.github/workflows/update-translations.yml
```diff
@@ -1,22 +1,56 @@
...
       - name: Fetch and merge from Weblate
         # The commit message is generated in Weblate; see https://hosted.weblate.org/addon/17163/
         run: |
+          git checkout -b main
+          git log --oneline --graph --all -n5
           git remote add weblate https://hosted.weblate.org/git/zulip/zulip-flutter/
           git fetch weblate
-          git merge --ff-only weblate/main
+          GIT_COMMITTER_NAME="Hosted Weblate" GIT_COMMITTER_EMAIL="hosted@weblate.org" \
+            git cherry-pick weblate/main ^HEAD
```
gnprice, line 27: Hmm I guess there's a race where if origin/main just got updated and Weblate hasn't taken that update yet, the `merge --ff-only` will fail.

How about `reset --hard`? Then the PR's base will be slightly behind the tip, but that's normal. We'll rebase it forward as part of merge like we do for many other PRs.

gnprice: Ah I see, this change is part of the "run on this PR" don't-merge commit. Never mind, then.

(There still is that race, but it's unlikely.)

alexmv: One option is we can `git reset --hard weblate/main` and let the PR potentially be behind `main` if weblate is lagging.  Which means one might get a PR with "cannot merge" rather than the failure one gets with `git merge --ff-only`, which is that the whole action fails.

PIG208: Yeah, I think lagging behind `main` is fine. We can always re-run the workflow.

.github/workflows/update-translations.yml
```diff
@@ -6,17 +6,41 @@ on:
...
+        run: |
+          git clone -b main https://github.com/flutter/flutter ~/flutter
+          TZ=UTC git --git-dir ~/flutter/.git log -1 --format='%h | %ci | %s' --date=iso8601-local
+          echo ~/flutter/bin >> "$GITHUB_PATH"
+
+          # The Flutter tool assumes the tip of tree is "origin/master"
+          # (or "upstream/master"):
+          #   https://github.com/flutter/flutter/issues/160626
+          # TODO(upstream): make workaround unneeded
+          git --git-dir ~/flutter/.git update-ref refs/remotes/origin/master origin/main
```
PIG208, line 40: Because #1204 has been resolved, we can update this to match.

alexmv:
Repushed, with comments addressed and the `DO NOT MERGE` trailing commit stripped off.

PIG208:
Just manually triggered [a run](https://github.com/zulip/zulip-flutter/actions/runs/13573313244). Let's see how it does.

alexmv:
The runs won't work because I stripped off the final commit which makes this work on non-`main`.

gnprice:
Thanks! This looks good; merging, and then I'll try a run from main.

Issue #1250, opened 2025-01-03
Handle no-topic topic ("general chat")


gnprice:
The server will start allowing channel messages to have an empty topic at the API level, replacing the old "(no topic)" behavior. This should be displayed as something like "*general chat*" — with italics — in the UI.

The exact phrase to use for "general chat" is specified by the server in the API. For compatibility with old clients, the feature is gated by a flag the client sends. For details, see references below.

This is a feature planned for the upcoming Zulip Server 10.

### Plan

We'll want a version of this soon, as work is proceeding on the server side. This will be a task for a member of the core team.

### References
* [#api design](https://chat.zulip.org/#narrow/channel/378-api-design/topic/general.20chat)
* Initial server PR: https://github.com/zulip/zulip/pull/32330
* [#design](https://chat.zulip.org/#narrow/channel/101-design/topic/general.20chat.20.2F.20default.20topic/near/1964986), for user/product motivation and UX choices
* (There doesn't appear to be a corresponding server/web issue.)

gnprice:
I've just sent #1266 as a refactor that will set us up to implement this, by distinguishing a TopicName type from plain String.

Other than that PR, I'd like to not drive this feature myself, in order to focus on other issues.

After that refactor, the places that will need to adapt to handle the empty topic differently from other topics can be found as the references to [TopicName.displayName]. There are about 9 such places. A couple of them will probably raise questions that should be discussed in the #design channel.

I believe the other pieces needed for the implementation are
* we'll need to track the new `realm_empty_topic_display_name` realm setting;
* we'll need to pass the `empty_topic_name` client capability, and the `allow_empty_topic_name` parameter to /api/get-messages and possibly other endpoints.

gnprice:
For when we're closing in on completing this, here's a list of late changes in the web UI for the feature:
* https://github.com/zulip/zulip/issues/32996

Some of those have no analogue in the mobile app, but others do have analogues, so it'd be good to run through that to see if it brings to mind any spots in the UI we might not have handled yet.

PR #1266, opened 2025-01-08
Distinguish a TopicName type from plain String

### LINKED ISSUE #1205: android: Notification grouping should treat topics case-insensitively
We use Android's "inbox style" notifications (this was #128 / PR #718) to group notifications by conversation.

For channel messages, this means grouping messages by topic. Topics in Zulip are meant to be case-insensitive, so we should ignore case for that grouping. For example, messages with the topic "move to Zulip" or the topic "move to zulip" should all be grouped together.[^1]

Other issues where we should be treating topics case-insensitively but we're not:

- #980
- #739

See our notification-display code for getting a message's "conversation key". It just takes the topic literally without normalizing it by case:

```dart
  static String _conversationKey(MessageFcmMessage data, String groupKey) {
    final conversation = switch (data.recipient) {
      FcmMessageChannelRecipient(:var streamId, :var topic) => 'stream:$streamId:$topic',
      FcmMessageDmRecipient(:var allRecipientIds) => 'dm:${allRecipientIds.join(',')}',
    };
    return '$groupKey|$conversation';
  }
```

[^1]: As they are in the message list in the web app. By using "Copy link to message", you can see that [this](https://chat.zulip.org/#narrow/channel/7-test-here/topic/move.20to.20Zulip/near/2016957) and [this](https://chat.zulip.org/#narrow/channel/7-test-here/topic/move.20to.20zulip/near/2016958) message have those topics that only differ by case. But they both appear in the message list as though they both have topic "move to Zulip". image

gnprice:
Toward #1250.
Fixes #1205.

This PR introduces a new type `TopicName` as a Dart "extension type":
https://dart.dev/language/extension-types
At runtime it's just a String — so there's no performance cost to it — but the type-checker tracks the distinction.

This lets us readily find all the spots in the code that need to adapt to show "*general chat*" for #1250: it's exactly the use sites of the new `TopicName.displayName`. (With one possible loophole involving string interpolation; see the last commit message below.)

This series is nearly NFC; each commit is NFC except the first two prep commits, which are nearly so.

## Selected commit messages

api: Explicitly ignore topics on DMs

The topic/subject field on DMs carries no information in the current
API, and we already never actually look at it, as demonstrated by
the lack of changes in the rest of the tree in this commit.  Clean
up the type definitions in the API a bit by making that explicit.

This is NFC except in the case where a server were to not send this
field at all for a DM (or to send a non-string value).  In that case
the old code would reject the server's response as malformed, and
the new code after this commit would accept it because it doesn't
look for the field.

---

api [nfc]: Introduce TopicName extension type

We'll use this to make a type-level distinction between a string
that's being used to name a topic, and just any old string.
That will help us in particular with #1250 the "general chat"
feature, where the way we show a topic name to the user stops being
necessarily just the same string that it appears as in the API.

The next phase will be to migrate a bunch of places in our code
to refer to TopicName when that's what they mean, instead of just
String.

During this phase, a TopicName can be freely used as a String, but
not vice versa.  So we'll do the migrations mostly in order of the
data flow, from upstream to downstream.  That will allow us to do
them over a series of individually coherent commits, with a minimum
of occasions where we temporarily introduce a conversion that won't
be needed in the final result.

That means: first, test data; then topics returned from the API to
our code; then our internal models; then back to the API, this time
for topics passed to the API from our code.

After we have the type in place all over where it belongs, we'll
start making use of the distinction, and then enforcing it.

---

api [nfc]: Use TopicName type across request parameters

This completes the conversion to use the TopicName type in the
places where it belongs.

Next, we'll start making use of this new type distinction: instead
of using a TopicName implicitly as a String, we'll introduce members
corresponding to the different ways a topic name is used which are
going to start having different behavior in the future.

After that's complete, we can start enforcing the distinction in
order to catch accidental future implicit uses as a String.

---

api [nfc]: Introduce TopicName.apiName, and use for making API requests

---

api [nfc]: Use TopicName.apiName for constructing narrow links

---

notif [nfc]: Use TopicName.apiName for encoding NotificationOpenPayload

This isn't part of the Zulip server API -- it's entirely between the
`buildUrl` and `parseUrl` of this class, and we could make up whatever
encoding we like.  But the API name of the topic is the natural thing
to use.

---

api [nfc]: Factor out TopicName.canonicalize

---

api [nfc]: Introduce TopicName.displayName, and use where needed

Each of these call sites will need to be updated for #1250.  Those
sites can now be found by listing the references to this getter.

As a bonus the type-checker will point out most of the places that
need updates, because this getter will become nullable.  (The
exceptions are where it's used in a string interpolation, because
`null.toString()` is valid and returns "null".)

---

api [nfc]: Seal the TopicName migration by making non-transparent

We're now explicitly using apiName, canonicalize, or displayName in
each of the places where we had been implicitly treating a topic name
as a String.  Let the type-checker catch any future regressions on
that front, by removing "implements String".

More precisely, I *think* we've covered each such place.  The
loophole is that Object methods can still be called, including
toString, so an interpolation like "${channelName} > ${topic}" won't
be flagged by the analyzer.  I found some such interpolations with a
grep, but that's necessarily heuristic and there could be more.
So the analyzer won't give us quite as much help here as we'd like;
but it'll give quite a bit, and this is the most I see how to do.

Without the "implements String", a TopicName value can no longer be
implicitly converted to String or have String methods called on it.
Instead the type-checker will require any code that has such a value
to call one of the members we've declared on TopicName (or a member
of Object) in order to do anything with it.  That way the code will
be explicit about whether it needs the API name, or the display
name, or the canonicalized form for making equality comparisons.

And that in turn will enable us to make "display name" behave
differently from "API name", for #1250 the "general chat" feature,
with a reliable way of tracking down which sites need which version.

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -655,6 +655,14 @@ enum MessageFlag {
   String toJson() => _$MessageFlagEnumMap[this]!;
 }
 
+/// The name of a Zulip topic.
+// TODO(#1250): Migrate all implicit uses as String; remove "implements String".
+extension type const TopicName(String _value) implements String {
+  TopicName.fromJson(this._value);
+
+  String toJson() => _value;
+}
+
 @JsonSerializable(fieldRename: FieldRename.snake)
 class StreamMessage extends Message {
   @override
@@ -674,7 +682,7 @@ class StreamMessage extends Message {
   // that will need new UI that we'll design then as part of that feature,
   // and ignoring the topics seems as good a fallback behavior as any.
   @JsonKey(name: 'subject')
-  String topic;
+  TopicName topic;
 
   StreamMessage({
     required super.client,
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -661,6 +661,9 @@ extension type const TopicName(String _value) implements String {
   /// The string this topic is identified by in the Zulip API.
   String get apiName => _value;
 
+  /// The key to use for "same topic as" comparisons.
+  String canonicalize() => apiName.toLowerCase();
+
   TopicName.fromJson(this._value);
 
   String toJson() => apiName;
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -556,11 +556,11 @@ sealed class Message {
   final String senderFullName;
   final int senderId;
   final String senderRealmStr;
-  @JsonKey(name: 'subject')
-  String topic;
+
   /// Poll data if "submessages" describe a poll, `null` otherwise.
   @JsonKey(name: 'submessages', readValue: _readPoll, fromJson: Poll.fromJson, toJson: Poll.toJson)
   Poll? poll;
+
   final int timestamp;
   String get type;
 
@@ -613,7 +613,6 @@ sealed class Message {
     required this.senderFullName,
     required this.senderId,
     required this.senderRealmStr,
-    required this.topic,
     required this.timestamp,
     required this.flags,
     required this.matchContent,
@@ -667,7 +666,15 @@ class StreamMessage extends Message {
   // invalidated.
   @JsonKey(required: true, disallowNullValue: true)
   String? displayRecipient;
+
   int streamId;
+
+  // The topic/subject is documented to be present on DMs too, just empty.
+  // We ignore it on DMs; if a future server introduces distinct topics in DMs,
+  // that will need new UI that we'll design then as part of that feature,
+  // and ignoring the topics seems as good a fallback behavior as any.
+  @JsonKey(name: 'subject')
+  String topic;
 
   StreamMessage({
     required super.client,
@@ -683,13 +690,13 @@ class StreamMessage extends Message {
     required super.senderFullName,
     required super.senderId,
     required super.senderRealmStr,
-    required super.topic,
     required super.timestamp,
     required super.flags,
     required super.matchContent,
     required super.matchTopic,
     required this.displayRecipient,
     required this.streamId,
+    required this.topic,
   });
 
   factory StreamMessage.fromJson(Map<String, dynamic> json) =>
@@ -786,7 +793,6 @@ class DmMessage extends Message {
     required super.senderFullName,
     required super.senderId,
     required super.senderRealmStr,
-    required super.topic,
     required super.timestamp,
     required super.flags,
     required super.matchContent,
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -659,7 +659,19 @@ enum MessageFlag {
 // TODO(#1250): Migrate all implicit uses as String; remove "implements String".
 extension type const TopicName(String _value) implements String {
   /// The string this topic is identified by in the Zulip API.
+  ///
+  /// This should be used in constructing HTTP requests to the server,
+  /// but rarely for other purposes.  See [displayName] and [canonicalize].
   String get apiName => _value;
+
+  /// The string this topic is displayed as to the user in our UI.
+  ///
+  /// At the moment this always equals [apiName].
+  /// In the future this will become null for the "general chat" topic (#1250),
+  /// so that UI code can identify when it needs to represent the topic
+  /// specially in the way prescribed for "general chat".
+  // TODO(#1250) carry out that plan
+  String get displayName => _value;
 
   /// The key to use for "same topic as" comparisons.
   String canonicalize() => apiName.toLowerCase();
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -835,13 +835,13 @@ enum MessageEditState {
   /// but for purposes of [Message.editState], we want to ignore such renames.
   /// This method identifies topic moves that should be ignored in that context.
   static bool topicMoveWasResolveOrUnresolve(TopicName topic, TopicName prevTopic) {
-    if (topic.startsWith(_resolvedTopicPrefix)
-        && topic.substring(_resolvedTopicPrefix.length) == prevTopic) {
+    if (topic.apiName.startsWith(_resolvedTopicPrefix)
+        && topic.apiName.substring(_resolvedTopicPrefix.length) == prevTopic.apiName) {
       return true;
     }
 
-    if (prevTopic.startsWith(_resolvedTopicPrefix)
-        && prevTopic.substring(_resolvedTopicPrefix.length) == topic) {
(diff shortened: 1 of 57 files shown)


chrisbobbe:
Oh nice!

```
More precisely, I think we've covered each such place. The
loophole is that Object methods can still be called, including
toString, so an interpolation like "${channelName} > ${topic}" won't
be flagged by the analyzer. I found some such interpolations with a
grep, but that's necessarily heuristic and there could be more.
So the analyzer won't give us quite as much help here as we'd like;
but it'll give quite a bit, and this is the most I see how to do.
```

I see several with `git grep '\$topic'` 🙂:

<details>

```
$ git grep '\$topic'
lib/api/model/model.dart://   In particular an interpolation "foo > $topic" is a bug we'd like to catch.
Binary file lib/generated/l10n/zulip_localizations_ar.dart matches
Binary file lib/generated/l10n/zulip_localizations_en.dart matches
Binary file lib/generated/l10n/zulip_localizations_fr.dart matches
Binary file lib/generated/l10n/zulip_localizations_ja.dart matches
Binary file lib/generated/l10n/zulip_localizations_pl.dart matches
Binary file lib/generated/l10n/zulip_localizations_ru.dart matches
lib/notifications/display.dart:      FcmMessageChannelRecipient(:var streamId, :var topic) => 'stream:$streamId:$topic',
test/model/autocomplete_test.dart:        eg.streamMessage(id: 60, sender: users[5-1], stream: stream, topic: 'other $topic'),
test/notifications/display_test.dart:      final expectedTitle = '#${stream.name} > $topic';
test/notifications/display_test.dart:      final expectedTagComponent = 'stream:${stream.streamId}:$topic';
test/notifications/display_test.dart:        expectedTitle: '#${stream.name} > $topicA',
test/notifications/display_test.dart:        expectedTagComponent: 'stream:${stream.streamId}:$topicA');
test/notifications/display_test.dart:        expectedTitle: '#${stream.name} > $topicB',
test/notifications/display_test.dart:        expectedTagComponent: 'stream:${stream.streamId}:$topicB');
test/notifications/display_test.dart:        expectedTitle: '#${stream.name} > $topicA',
test/notifications/display_test.dart:        expectedTagComponent: 'stream:${stream.streamId}:$topicA');
test/notifications/display_test.dart:        expectedTitle: '#Before > $topic',
test/notifications/display_test.dart:        expectedTagComponent: 'stream:${stream.streamId}:$topic');
test/notifications/display_test.dart:        expectedTitle: '#After > $topic',
test/notifications/display_test.dart:        expectedTagComponent: 'stream:${stream.streamId}:$topic');
test/notifications/display_test.dart:        conditionActiveNotif(data3, 'stream:${stream.streamId}:$topicA'),
test/notifications/display_test.dart:        conditionActiveNotif(data3, 'stream:${stream.streamId}:$topicA'),
test/notifications/display_test.dart:        conditionActiveNotif(data1, 'stream:${stream.streamId}:$topicA'),
test/notifications/display_test.dart:        conditionActiveNotif(data2, 'stream:${stream.streamId}:$topicB'),
test/notifications/display_test.dart:        conditionActiveNotif(data2, 'stream:${stream.streamId}:$topicB'),
test/notifications/display_test.dart:        conditionActiveNotif(data1, 'stream:${stream.streamId}:$topic'),
test/notifications/display_test.dart:        conditionActiveNotif(data2, 'stream:${stream.streamId}:$topic'),
test/notifications/display_test.dart:        conditionActiveNotif(data2, 'stream:${stream.streamId}:$topic'),
test/notifications/display_test.dart:        conditionActiveNotif(data1, 'stream:${stream.streamId}:$topic'),
test/notifications/display_test.dart:        conditionActiveNotif(data2, 'stream:${stream.streamId}:$topic'),
test/notifications/display_test.dart:        conditionActiveNotif(data2, 'stream:${stream.streamId}:$topic'),
test/widgets/message_list_test.dart:        ..decoration.isNotNull().hintText.equals('Message #${channel.name} > $topic')
```

</details>

(The l10n results are red herrings; they don't involve `TopicName`)

Just one result in non-test code. But it's #1205, so probably helpful to process that one too, perhaps adding a `TODO(#1205)` because I see we don't have one yet. 🙂 I like that the fix there is simple; just need to call `.canonicalize()`, right?

lib/api/model/model.dart
```diff
@@ -656,6 +655,38 @@ enum MessageFlag {
...
 }
 
+/// The name of a Zulip topic.
+// TODO(dart): Can we forbid calling Object members on this extension type?
+//   (The lack of "implements Object" ought to do that, but doesn't.)
+//   In particular an interpolation "foo > $topic" is a bug we'd like to catch.
+// TODO(dart): Can we forbid using this extension type as a key in a Map?
+//   (The lack of "implements Object" arguably should do that, but doesn't.)
+//   Using as a Map key is almost certainly a bug because it won't case-fold;
+//   see for example #739, #980, #1205.
```
chrisbobbe, line 665: Hmm, yeah. One way this PR is helpful is by making it easy to find topic-keyed maps, where we're being incorrectly case-sensitive 🙂 with `git grep 'Map<TopicName,'`

- `Unreads.streams` (tracked as #980)
- `ChannelStoreImpl.topicVisibility` (no issue yet!)
- `RecentSenders.topicSenders` (no issue yet!)

Do you know if we can make a nice drop-in replacement for `Map<TopicName, …>` that uses the canonicalized form of `TopicName`?

chrisbobbe:
Small comments above, otherwise LGTM! Thanks for doing this!

gnprice:
> Just one result in non-test code. But it's #1205, so probably helpful to process that one too, perhaps adding a `TODO(#1205)` because I see we don't have one yet. 🙂 I like that the fix there is simple; just need to call `.canonicalize()`, right?

Hmm yeah, indeed.

I'll just fix that one — it's a small fix.

lib/api/model/model.dart
gnprice: Yeah, that is handy.

> Do you know if we can make a nice drop-in replacement for `Map<TopicName, …>` that uses the canonicalized form of `TopicName`?

Yeah. I think we'll want a class with a name like `TopicMap`, which maybe even implements `Map<TopicName, …>`. Its implementation will resemble web's `FoldDict`:
https://github.com/zulip/zulip/blob/e963d5eaa9d73faa31b9c9e1463a2ac0a5169e7a/web/src/fold_dict.ts

gnprice:
Updated — PTAL.

chrisbobbe:
Thanks! LGTM, merging.

PR #1267, opened 2025-01-09
Max message limit changed from static value to server value

Spyyy004:
1. Extracted max_message_limit from the register-queue API endpoint response
2. Added the max_message_limit property in the initial_snapshot model
3. Created a maxMessageLimit variable in the compose_box widget and set the value of it to the value fetched from the api endpoint while rendering the sendButton. 
4. Added dummy value for maxMessageLimit in example_data.dart file. 
5. Generated a new .g file for initial_snapshot as I made a new addition to the model.


chrisbobbe:
This will need to follow the project's Git style before it can be reviewed.

gnprice:
It looks like there isn't further work happening on this PR; so closing it to get it out of our queue.

PR #1270, opened 2025-01-10
Split inline from block parser class, and other small refactors

gnprice:
This series of commits has a few small refactors of the content-parsing code, prompted by #1156.

@rajveermalviya feel free to rebase #1156 atop this if it seems helpful — the last two commits in particular might simplify that PR a bit. If the conflict resolution seems more annoying than the refactor is helpful, no need, and I can resolve it after either PR is merged.

## Selected commit messages

content [nfc]: Split out _ZulipInlineContentParser class

This helps organize the parsing code a bit more cleanly -- not to
mention more statically, by making the dynamic _debugParserContext
assertions redundant.

---

content [nfc]: Encapsulate parsing math nodes a bit more

---

content [nfc]: Factor out consumeImageNodes in block-content parse loops

This hopefully makes the logic of these loops a bit more readable.

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -786,7 +786,7 @@ class MathInlineNode extends InlineContentNode {
 class GlobalTimeNode extends InlineContentNode {
   const GlobalTimeNode({super.debugHtmlNode, required this.datetime});
 
-  /// Always in UTC, enforced in [_ZulipContentParser.parseInlineContent].
+  /// Always in UTC, enforced in [_ZulipInlineContentParser.parseInlineContent].
   final DateTime datetime;
 
   @override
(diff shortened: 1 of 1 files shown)


lib/model/content.dart
```diff
@@ -1054,8 +1055,11 @@ class _ZulipContentParser {
     return inlineParser.parseBlockInline(nodes);
   }
 
-  String? parseMathBlock(dom.Element element) {
-    return _parseMath(element, block: true);
+  BlockContentNode parseMathBlock(dom.Element element) {
+    final debugHtmlNode = kDebugMode ? element : null;
```
PIG208, line 1059: Looks like we will end up with a different `debugHtmlNode` here, but that's a debug-mode-only behavior so it should be fine as a part of an [nfc] commit.

PIG208 (approved):
Looks good to me! Merging so that @rajveermalviya can rebase atop main to make use of this for #1156.

lib/model/content.dart
gnprice: Ah, yeah, would have been good for me to make that explicit.

There's also a related change that if it fails to parse, the UnimplementedNode will get the `span.katex-display` child instead of the `p` parent. I think that change is good (and it's definitely needed in the #1156 context, and in fact the current revision of #1156 effectively already makes that change) but it's not quite NFC.

I actually had that thought at one point as I was writing these changes; but I guess I forgot at the moment I was making that commit.

gnprice:
Thanks for the review!

Issue #1225, opened 2024-12-27
Offer "Mark topic as read" in topic action sheet


chrisbobbe:
In the legacy app, when a topic has unreads, we offer "Mark topic as read" in the topic action sheet. We should do that in zulip-flutter too.

Marking as "beta feedback" because this was requested for the (not yet built) channel action sheet [on CZO](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20beta.3A.20no.20access.20to.20channel.20menu.20via.20long.20press/near/2011844):
> In the channel list, a long press on the channel name should bring up a menu with options like "Mark all read" and "unsubscribe" and "copy link". That menu was available in the old app.

## Related issues
- #1226

AshutoshKhadse23:
Hello,
I would like to work on it 

https://github.com/user-attachments/assets/dd6b3219-d0d4-43e2-a705-3e7ac15e6b14

gnprice:
That screen capture appears to show a "Mark stream as read" option. That'd be #1226 (and should say "channel"). See the title of this issue.

AshutoshKhadse23:
Thanks for pointing that out! You're correct that the terminology should align with 'channel' instead of 'stream' for consistency.I will update accordingly.

lakshya1goel:
Hi, I have worked on this issue. Here is the screen recording for the same. Please assign the issue to me. 

https://github.com/user-attachments/assets/140c529b-41f8-47c6-b3af-9e4c58eca51c

Also what icon should I use for this?

gnprice:
Thanks @lakshya1goel! Assigned to you.

> Also what icon should I use for this?

Good question. The place to ask that is in the `#mobile-design` channel. See details here:
https://zulip.readthedocs.io/en/latest/contributing/design-discussions.html#guidelines-for-code-contributors

(That section says `#design`, but for the mobile app the right channel is `#mobile-design`.)

lakshya1goel:
Hello @gnprice, the PR is now ready for review, PTAL.
Thanks!

chrisbobbe:
When you need to mention a PR in its corresponding issue, please include the PR number; that helps us find which PR you're referring to. In this case, just commenting on the PR itself, instead, would have been effective, and it's nice to avoid unnecessary comments on GitHub issues. GitHub's UI doesn't accommodate long conversations very well.

PR #1274, opened 2025-01-12
action_sheet: Add "Mark topic as read" button

lakshya1goel:
This PR adds a "Mark Topic as Read" button to the topic action sheet. The button appears only when there are unread messages in the topic.
  
Fixes: #1225
Fixes: #1357

## Screenshot
WhatsApp Image 2025-02-06 at 3 05 22 AM

### Screen Recording

https://github.com/user-attachments/assets/aaa03f37-a162-4aef-93ba-7115541a5d6a

### Additional
- Updated icon for "Mark all messages as read button" from `Icons.playlist_add_check` to `Icons.mark_chat_read_outlined`

See [CZO Discussion](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Icon.20for.20.22Mark.20topic.20as.20read.22.20in.20topic.20action.20sheet)

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>


lakshya1goel:
Hello, I want to know that should I commit the change of icon of  "Mark all messages as read button" separately?

PIG208:
Yeah, committing it separately sounds good.

lib/widgets/action_sheet.dart
```diff
@@ -372,6 +381,61 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+    return zulipLocalizations.actionSheetOptionMarkTopicAsRead;
+  }
+
+  @override void onPressed() async {
+    final store = PerAccountStoreWidget.of(pageContext);
+    final connection = store.connection;
+    final zulipLocalizations = ZulipLocalizations.of(pageContext);
+
+    try {
+      if (connection.zulipFeatureLevel! >= 155) {
```
PIG208, line 408: `markNarrowAsRead` handles feature level compatibility and only takes a `BuildContext` and a `Narrow`. We can use that here.

PIG208 (commented):
Thanks for the PR! Took a quick look at the implementation and left a comment.

alya:
Why is the icon different from the one we have in the web app?

Screenshot 2025-01-31 at 11 01 39

lakshya1goel:
Thanks Alya for pointing. I have updated the correct icon and also updated the PR description with latest screenshots.

lib/widgets/action_sheet.dart
lakshya1goel: Thanks for the review, I have updated the implementation and thus, removed the tests as well targeting the API requests.

lakshya1goel:
Thanks for the review @PIG208, pushed the revision, PTAL.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
       });
     });
 
+    group('MarkTopicAsReadButton', () {
+      Future<void> setupToTopicActionSheetWithUnreadMessages(WidgetTester tester, {
```
PIG208, line 1065: Looks like this and `setupToTopicActionSheetWithNoUnreadMessages` are each only called once in their respective test. Let's reorganize this for clarity.

The logic tested here whether `MarkTopicAsReadButton` is hidden in the action sheet, where the only variable is the presence of unread messages in the store, so it would make sense to have a single helper that's capable of controlling that.

If you look at the "topic action sheet" test group, you can find a helper called `prepare` that should accomplish all of these for you. You can use the "UserTopicUpdateButton" test group as a reference.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
...
+        ZulipStream? channel,
+      }) async {
+        addTearDown(testBinding.reset);
+
+        final effectiveChannel = channel ?? eg.stream();
```
PIG208, line 1070: Because `channel` is unused, we can just have `final channel = eg.stream()` here.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
       });
     });
 
+    group('MarkTopicAsReadButton', () {
+      Future<void> setupToTopicActionSheetWithUnreadMessages(WidgetTester tester, {
+        ZulipStream? channel,
```
PIG208, line 1066: We can remove this because it is unused.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
...
+        store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+        connection = store.connection as FakeApiConnection;
+
+        connection.prepare(json: eg.newestGetMessagesResult(
+          foundOldest: true, messages: [message]).toJson());
+
+        await store.addMessage(message);
+        store.unreads.streams[effectiveChannel.streamId] ??= {};
+        store.unreads.streams[effectiveChannel.streamId]![topicName] ??= QueueList<int>();
+        store.unreads.streams[effectiveChannel.streamId]![topicName]!.add(message.id);
```
PIG208, line 1086: We shouldn't need to modify `store.unreads.streams` here. A convenient way to add unread messages to the store is through `store.handleEvent(MessageEvent(id: 0, message: message))`. There should be some examples in other tests.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
...
+        final message = eg.streamMessage(stream: effectiveChannel, topic: 'test topic');
+        await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot(
+          realmUsers: [eg.selfUser],
+          streams: [effectiveChannel],
+          subscriptions: [eg.subscription(effectiveChannel)]));
+        store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+        connection = store.connection as FakeApiConnection;
+
+        connection.prepare(json: eg.newestGetMessagesResult(
+          foundOldest: true, messages: [message]).toJson());
```
PIG208, line 1081: Looking at `setupToMessageActionSheet`, the `connection.prepare` call is right before `tester.pumpWidget`. This is because this call prepares for a request that will be made by `MessageListPage`. Moving it there help highlight their relationship.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
...
+
+        await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
+          child: MessageListPage(initNarrow: TopicNarrow(effectiveChannel.streamId, topicName))));
+        await tester.pumpAndSettle();
+
+        await tester.longPress(find.byType(ZulipAppBar));
+        await tester.pump(const Duration(milliseconds: 250));
+      }
+
+      Future<void> setupToTopicActionSheetWithNoUnreadMessages(WidgetTester tester) async {
```
PIG208, line 1096: The other helper can be removed once we refactor the first one to handle both situations.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
...
+        await tester.longPress(find.byType(ZulipAppBar));
+        await tester.pump(const Duration(milliseconds: 250));
+      }
+
+      group('visibility', () {
+        testWidgets('shows button when topic has unread messages', (tester) async {
+          await setupToTopicActionSheetWithUnreadMessages(tester);
+
+          final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+          check(find.text(zulipLocalizations.actionSheetOptionMarkTopicAsRead)).findsOne();
```
PIG208, line 1128: Let's use the string directly, without referencing `GlobalLocalizations`, because we don't need tests to respond to locale changes and that we can be clear about what is expected exactly.

test/widgets/action_sheet_test.dart
```diff
@@ -1061,6 +1061,82 @@ void main() {
       });
     });
 
+    group('MarkTopicAsReadButton', () {
```
PIG208, line 1064: This test should be moved to the "topic action sheet" group.

test/widgets/action_sheet_test.dart
PIG208: The following comments apply to this helper, but you can skip most of them when working on the next revision.

PIG208 (commented):
Thanks for the PR! Went through the changes and left some comments, mainly on the tests.

lakshya1goel:
Thanks for the review @PIG208, I have updated the tests and the commit message accordingly. Please have a look and let me know if anything else is required.

test/widgets/action_sheet_test.dart
```diff
@@ -450,6 +450,39 @@ void main() {
         }
       });
     });
+
+    group('MarkTopicAsReadButton', () {
+      Future<void> setupToTopicActionSheet(WidgetTester tester, {
+        required bool hasUnreadMessages,
+      }) async {
+        addTearDown(testBinding.reset);
```
PIG208, line 458: We can skip the `addTearDown` call because `prepare` already calls it for us.

test/widgets/action_sheet_test.dart
```diff
@@ -450,6 +450,39 @@ void main() {
...
+          await store.handleEvent(MessageEvent(id: 0, message: message));
+        }
+        await showFromAppBar(tester,
+          channel: channel, topic: topic, messages: [message]);
+      }
+
+      testWidgets('visible if topic has unread messages', (tester) async {
+        await setupToTopicActionSheet(tester, hasUnreadMessages: true);
+        check(find.text('Mark Topic As Read')).findsOne();
+      });
```
PIG208, line 479: Let's try to simplify away the helper, we can always add it back if we find a lot of tests reusing the same logic:

```suggestion
      testWidgets('visible if topic has unread messages', (tester) async {
        await prepare();
        final message = eg.streamMessage(stream: someChannel, topic: someTopic);
        await store.handleEvent(MessageEvent(id: 0, message: message));
        await showFromAppBar(tester, messages: [message]);
        check(find.text('Mark Topic As Read')).findsOne();
      });
```

This way we can also skip the flag and the if-statement.

PIG208 (commented):
Thanks for the update! Left a review comment.

lakshya1goel:
Pushed the changes, Please have a look now @PIG208 , Thanks!

test/widgets/action_sheet_test.dart
```diff
@@ -450,6 +450,39 @@ void main() {
...
+
+      testWidgets('not visible if topic has no unread messages', (tester) async {
+        await setupToTopicActionSheet(tester, hasUnreadMessages: false);
+        check(find.text('Mark Topic As Read')).findsNothing();
+      });
+    });
```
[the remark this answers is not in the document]
lakshya1goel: Should I add 2 tests for both:  
- For FL < 155: /api/v1/mark_topic_as_read
- For FL ≥ 155: /api/v1/messages/flags/narrow

test/widgets/action_sheet_test.dart
```diff
@@ -450,6 +450,53 @@ void main() {
...
+            'num_before': '0',
+            'num_after': '1000',
+            'narrow': jsonEncode([
+              ...TopicNarrow(someChannel.streamId, TopicName(someTopic)).apiEncode(),
+              {'operator': 'is', 'operand': 'unread'}
+            ]),
+            'op': 'add',
+            'flag': 'read',
+          });
+      });
```
PIG208, line 516: For this new test, we just need to focus on checking the field most interesting to us, i.e. 'narrow'. This is exactly what we are doing in 'on topic move, acts on new topic':

```dart
        testWidgets('on topic move, acts on new topic', (tester) async {
          // ...
          check(connection.lastRequest).isA<http.Request>()
            ..method.equals('POST')
            ..url.path.equals('/api/v1/messages/flags/narrow')
            ..bodyFields['narrow'].equals(
                jsonEncode(eg.topicNarrow(newStream.streamId, newTopic).apiEncode()));
```

This is helpful to highlight what is important for this particular test.

PIG208:
Could you also post a screenshot of the topic action sheet with the mark topic as read button on there? For design changes, we prefer screenshots over screen recording, unless it is specific about some interaction. Even then, having both included is helpful.

lakshya1goel:
Updated the description of PR.

test/widgets/action_sheet_test.dart
lakshya1goel: Thanks for pointing, I will make sure to include the necessary part only from now on.

test/widgets/action_sheet_test.dart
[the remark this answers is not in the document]
PIG208: Here we don't need to, because the tests for `markNarrowAsRead` cover that.

alya:
The capitalization doesn't match the web app / issue. Please be mindful not to introduce unexpected changes from the spec.

lakshya1goel:
> The capitalization doesn't match the web app / issue. Please be mindful not to introduce unexpected changes from the spec.

Sorry for missing out that, updated it. Also changed the screenshot and recording in PR description.

lib/widgets/action_sheet.dart
```diff
@@ -372,6 +380,30 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionMarkTopicAsRead;
+  }
+
+  @override void onPressed() async {
+    if (!pageContext.mounted) return;
+    await markNarrowAsRead(pageContext, TopicNarrow(channelId, topic));
+  }
```
chrisbobbe, line 404: This should have error handling, like other buttons that make API requests

lakshya1goel: Since the `markNarrowAsRead` itself has its error handling, I haven't handled that here. Following the [CZO discussion](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/.23F1317.20showErrorDialog), I have pushed the revision.

PIG208:
There are some conflicts since the previous update. Could you resolve that? Thanks!

test/widgets/action_sheet_test.dart
```diff
@@ -568,6 +568,44 @@ void main() {
...
+          foundOldest: true, foundNewest: true).toJson());
+        await tester.tap(find.text('Mark topic as read'));
+        await tester.pumpAndSettle();
+
+        check(connection.lastRequest).isA<http.Request>()
+          ..url.path.equals('/api/v1/messages/flags/narrow')
+          ..bodyFields['narrow'].equals(jsonEncode([
+            ...TopicNarrow(someChannel.streamId, TopicName(someTopic)).apiEncode(),
+            {'operator': 'is', 'operand': 'unread'}
+          ]));
```
PIG208, line 606: nit: indentation
```suggestion
              ...TopicNarrow(someChannel.streamId, TopicName(someTopic)).apiEncode(),
              {'operator': 'is', 'operand': 'unread'}
            ]));
```

PIG208 (commented):
Thanks for the update! This mostly looks good to me, with some nits.

lib/widgets/message_list.dart
```diff
@@ -852,7 +852,7 @@ class _MarkAsReadWidgetState extends State<MarkAsReadWidget> {
                 backgroundColor: WidgetStatePropertyAll(messageListTheme.unreadMarker),
               ),
               onPressed: _loading ? null : () => _handlePress(context),
-              icon: const Icon(Icons.playlist_add_check),
+              icon: const Icon(ZulipIcons.message_checked),
```
chrisbobbe, line 855: Ah: I missed the relevant discussion and ended up opening an issue for this change 🙂 please add `Fixes #1357.` to the commit message and PR description. Thanks!

lakshya1goel:
Pushed the revision @PIG208, PTAL. Thanks!

test/widgets/action_sheet_test.dart
```diff
@@ -568,6 +568,44 @@ void main() {
...
+          firstProcessedId: message.id, lastProcessedId: message.id,
+          foundOldest: true, foundNewest: true).toJson());
+        await tester.tap(find.text('Mark topic as read'));
+        await tester.pumpAndSettle();
+
+        check(connection.lastRequest).isA<http.Request>()
+          ..url.path.equals('/api/v1/messages/flags/narrow')
+          ..bodyFields['narrow'].equals(jsonEncode([
+              ...TopicNarrow(someChannel.streamId, TopicName(someTopic)).apiEncode(),
+              {'operator': 'is', 'operand': 'unread'}
```
PIG208, line 605: nit: trailing comma for list elements

PIG208:
Looks good except for one more nit! Marking it for Greg's review.

lakshya1goel:
Hi @gnprice, please review the PR when you have some time, thanks!

lib/widgets/actions.dart
```diff
@@ -18,49 +18,59 @@ import '../model/narrow.dart';
...
-      if (!context.mounted) return;
-      showErrorDialog(context: context,
-        title: zulipLocalizations.errorMarkAsReadFailedTitle,
-        message: e.toString()); // TODO(#741): extract user-facing message better
-      return;
+/// High-level operations that interact with the Zulip API and handle UI feedback.
+///
+/// Methods in this class show progress and error feedback in the UI,
+/// and handle error cases internally rather than expecting the caller to handle them.
+abstract final class ZulipAction {
```
chrisbobbe, line 25: `markNarrowAsUnreadFromMessage` should also go in this class; see PR #1317 for a revised version of this commit.

test/widgets/action_sheet_test.dart
```diff
@@ -568,6 +568,44 @@ void main() {
...
+          foundOldest: true, foundNewest: true).toJson());
+        await tester.tap(find.text('Mark topic as read'));
+        await tester.pumpAndSettle();
+
+        check(connection.lastRequest).isA<http.Request>()
+          ..url.path.equals('/api/v1/messages/flags/narrow')
+          ..bodyFields['narrow'].equals(jsonEncode([
+              ...TopicNarrow(someChannel.streamId, TopicName(someTopic)).apiEncode(),
+              {'operator': 'is', 'operand': 'unread'},
+            ]));
```
gnprice, line 606: This checks it's doing *something* to the message flags in this narrow — but not that it's marking as read. In particular this test would equally well pass if the implementation were changed to mark the messages as *unread* instead of read.

Let's add the checks that ensure what it's doing is marking as read, rather than marking as unread or setting or unsetting some other flag.

lib/widgets/actions.dart
```diff
@@ -18,49 +18,59 @@ import '../model/narrow.dart';
...
+  /// Marks all messages in the given [narrow] as read, showing feedback in the UI.
+  ///
+  /// Shows progress feedback for long-running operations, and error feedback
+  /// if the operation fails. Updates unread counts in the UI on success.
+  static Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
```
gnprice, line 30: This last sentence isn't accurate — this function doesn't do anything to unread counts in the UI.

(It wouldn't, because it's the model's job to drive the data that appears in the UI. But this doesn't do anything to unread counts in the model, either. If you're thinking of the `PerAccountStoreWidget.of(context).unreads.handleAllMessagesReadSuccess()` call at the end, see that method's doc.)

lib/widgets/actions.dart
```diff
@@ -18,49 +18,59 @@ import '../model/narrow.dart';
...
-    try {
-      await _legacyMarkNarrowAsRead(context, narrow);
-      return;
-    } catch (e) {
-      if (!context.mounted) return;
-      showErrorDialog(context: context,
-        title: zulipLocalizations.errorMarkAsReadFailedTitle,
-        message: e.toString()); // TODO(#741): extract user-facing message better
-      return;
+/// High-level operations that interact with the Zulip API and handle UI feedback.
```
gnprice, line 21: "handle feedback" sounds like something that *receives* feedback and does something about it. These do the opposite, showing feedback to the user.

gnprice (commented):
Thanks @lakshya1goel for taking care of this, and thanks @PIG208 and all for the previous reviews! Comments below.

Let's leave out the last commit introducing ZulipAction; that change is mostly a matter of writing docs, so I think it'll be simpler if I just do that change myself rather than going through cycles of editing.

test/widgets/action_sheet_test.dart
gnprice: This needs more checks than just the narrow — see https://github.com/zulip/zulip-flutter/pull/1274#discussion_r1960914494.

One key point about that "on topic move, acts on new topic" test is that it comes after a smoke-test for the same button, which checks all the details including that the request is marking the messages in question as (in that case) unread. So that test is focused specifically on the part that's different when the topic moved, and that difference is all about the narrow that's being acted on.

gnprice:
> Let's leave out the last commit introducing ZulipAction; that change is mostly a matter of writing docs, so I think it'll be simpler if I just do that change myself rather than going through cycles of editing.

I've just sent #1366 for this part.

lib/widgets/action_sheet.dart
```diff
@@ -461,6 +469,30 @@ class ResolveUnresolveButton extends ActionSheetMenuItemButton {
...
+  @override IconData get icon => ZulipIcons.message_checked;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionMarkTopicAsRead;
+  }
+
+  @override void onPressed() async {
+    if (!pageContext.mounted) return;
+    await ZulipAction.markNarrowAsRead(pageContext, TopicNarrow(channelId, topic));
```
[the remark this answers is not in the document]
lakshya1goel: No, there was no discussion done for this in previous round. Also I will not mark the comments as resolved form now on.

lakshya1goel:
Pushed the revision - PTAL @gnprice , Thanks!

lakshya1goel:
Hello @gnprice, whenever you have some time, could you kindly take a look at the PR? Thank you!

assets/icons/message_checked.svg
gnprice, line 1: nit: commit message has stray leading space

     msglist: Update "Mark all messages as read" button icon
    
    Icon taken from the Figma design:
    https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3512-7&t=FEydAZe5DNKVlKnV-0
    
    Fixes: #1357

test/widgets/action_sheet_test.dart
```diff
@@ -562,6 +563,47 @@ void main() {
...
+        await prepare();
+        final message = eg.streamMessage(stream: someChannel, topic: someTopic, flags: []);
+        await store.addMessage(message);
+        await showFromAppBar(tester, messages: [message]);
+        check(find.text('Mark topic as read')).findsOne();
+      });
+
+      testWidgets('not visible if topic has no unread messages', (tester) async {
+        await prepare();
+        final message = eg.streamMessage(stream: someChannel, topic: someTopic, flags: [MessageFlag.read]);
```
gnprice, line 578: nit: line too long: key information is past 80 columns

gnprice (commented):
Thanks for the revision! All looks good, modulo two nits below. I'll fix them up and merge.

test/widgets/action_sheet_test.dart
```diff
@@ -562,6 +563,47 @@ void main() {
...
+        await store.addMessage(message);
+        await showFromAppBar(tester, messages: [message]);
+        check(find.text('Mark topic as read')).findsNothing();
+      });
+
+      testWidgets('marks topic as read when pressed', (tester) async {
+        await prepare();
+        final message = eg.streamMessage(stream: someChannel, topic: someTopic);
+        await store.handleEvent(MessageEvent(id: 0, message: message));
+        await showFromAppBar(tester, messages: [message]);
```
gnprice, line 588: Hmm I guess there's one more-substantive comment — this test should be fixed for some of the same comments made in the last review above.

I've already fixed the two nits (and also rebased past changes I merged earlier today), so I'll push those fixes back to the PR branch. But then @lakshya1goel  please make a revision to update this test case too based on the feedback given above on other tests.

lakshya1goel:
Thanks @gnprice for updating the PR. I have also pushed the revision for the test case. PTAL, Thanks!

gnprice:
Thanks! Looks good; merging.

PR #1275, opened 2025-01-13
l10n: Add stub translation for nb (Norwegian Bokmål).

alexmv:
This is represented in Weblate as `app_nb_NO.arb`, which is not supported without the existence of a bare `app_nb.arb`.  Adding a bare `app_nb.arb` for now may allow Weblate's aliases[^1] to map the `nb_NO` to a file name `app_nb.arb`.

[^1]: https://docs.weblate.org/en/latest/admin/projects.html#project-language-aliases


gnprice:
Thanks!

CI is failing because the generated files need a corresponding update. I'll run `tools/check l10n --fix` and merge.

PR #1276, opened 2025-01-13
Make {MessageList,EmojiReaction}Theme.{light,dark} static final fields

chrisbobbe:
Followup to https://github.com/zulip/zulip-flutter/pull/1236#issuecomment-2576750264:

> We should be able to apply this to MessageListTheme and etc., right?

We can, for `MessageListTheme` and `EmojiReactionTheme`, but not for `ContentTheme` because the light and dark variants need a `BuildContext`.

### DIFF
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -16,38 +16,36 @@ import 'theme.dart';
 
 /// Emoji-reaction styles that differ between light and dark themes.
 class EmojiReactionTheme extends ThemeExtension<EmojiReactionTheme> {
-  EmojiReactionTheme.light() :
-    this._(
-      bgSelected: Colors.white,
+  static final light = EmojiReactionTheme._(
+    bgSelected: Colors.white,
 
-      // TODO shadow effect, following web, which uses `box-shadow: inset`:
-      //   https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#inset
-      //   Needs Flutter support for something like that:
-      //     https://github.com/flutter/flutter/issues/18636
-      //     https://github.com/flutter/flutter/issues/52999
-      //   Until then use a solid color; a much-lightened version of the shadow color.
-      //   Also adapt by making [borderUnselected] more transparent, so we'll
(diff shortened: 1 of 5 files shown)


lib/widgets/content.dart
```diff
@@ -38,91 +38,91 @@ import 'text.dart';
 class ContentTheme extends ThemeExtension<ContentTheme> {
```
PIG208, line 38: nit: perhaps it's intentional, but don't we usually use `content:` for the commit summary of changes like this?

PIG208 (commented):
Thanks! I did a quick grep for the old variable names (in the final commit) and found no hanging references in comments. This LGTM.

chrisbobbe:
Thanks for the review! Revision pushed, and I see you've marked for Greg's review.

lib/widgets/content.dart
```diff
@@ -1092,7 +1092,7 @@ class _InlineContentBuilder {
 
     return _buildNodes(
       style: widget.style
-        .merge(ContentTheme.of(_context!).textStyleInlineCode)
+        .merge(ContentTheme.of(_context!).inlineCode)
```
gnprice, line 1095: I'm not sure about this last commit. It feels confusing to me to have some of these be colors and others text styles and not have any identification of which is which — you can't really guess from the names, as it's arbitrary that e.g. `pollNames` is only a color while `inlineCode` is a style. (The poll names are text just as much as the inline code is.)

It's true that the types are distinct and so the type-checker will prevent any real disasters. But it's useful to be able to read and understand the code without having to go track down (or consult the IDE about) the types of things all the time.

The other theme classes don't have this issue because they're entirely colors.

gnprice (commented):
Thanks! Comment below; otherwise looks good.

lib/widgets/content.dart
chrisbobbe: Oh, yeah I don't mind dropping the last commit :)

chrisbobbe:
Thanks for the review! Revision pushed, atop #1290 for the sake of CI.

Issue #942, opened 2024-09-13
Use new null-safe generics in Pigeon


gnprice:
The Pigeon feature request to handle types like `List<Foo>`, as opposed to only `List<Foo?>`, is complete:
* https://github.com/flutter/flutter/issues/97848#event-14254355682

We have a couple of types in our `pigeon/notifications.dart` that could benefit from that. One has a TODO comment; for the full list, look for `List<` and `Map<`.

ArpanMondalGITHUB:
Can I work on it

skylon07:
I'd also add that [the docs](https://github.com/flutter/flutter/issues/97848) should probably be updated too, since #5 is now obsolete.

gnprice:
@skylon07 Good catch, yeah.

That'd be a good point to make as an issue in the Flutter upstream tracker: https://github.com/flutter/flutter/issues . This thread here is in the issue tracker of a particular app that uses Flutter, namely Zulip.

skylon07:
Oh, haha, I clicked into this issue from the Flutter repo and didn't realize I had switched to Zulip. My apologies -- I'll make a ticket there instead.

PR #1278, opened 2025-01-13
Upgrade Flutter and packages

rajveermalviya:
Fixes: #942


test/model/binding.dart
```diff
@@ -508,6 +509,20 @@ typedef FirebaseMessagingRequestPermissionCall = ({
...
+  @override
+  // ignore: non_constant_identifier_names
+  final BinaryMessenger? pigeonVar_binaryMessenger = null;
+
+  // TODO(?): Find a better way to handle this. This member is exported from
+  //   the Pigeon generated class but are not used for this fake class,
+  //   so return the default value.
+  @override
+  // ignore: non_constant_identifier_names
+  final String pigeonVar_messageChannelSuffix = '';
```
PIG208, line 528: I wonder if `dartHostTestHandler` helps, but it is also subject to removal (https://github.com/flutter/flutter/issues/159886). Mocking `HostApi` this way seems fine to me. Maybe these `pigeonVar` prefixed variables should be private in the generated code.

PIG208 (commented):
Thanks for the PR! I checked the changelogs from the upgrades. None of the breaking changes packages from the `tools/upgrade pub-major` upgrade affects us. The app also works fine on my Android device. It would be great to have it tested on someone's iOS device too.

pubspec.yaml
```diff
@@ -39,7 +39,7 @@ dependencies:
   collection: ^1.17.2
   convert: ^3.1.1
   crypto: ^3.0.3
-  device_info_plus: ^10.0.1
+  device_info_plus: ^11.2.0
```
gnprice, line 42: Looks like the one "breaking" change in this changelog is:
> `BREAKING FIX(device_info_plus): fixed webasm compliance (#3254). (e35e2123)`

so I'll assume that's web-only and has no effect on us.

pubspec.yaml
```diff
@@ -50,7 +50,7 @@ dependencies:
   http_parser: ^4.0.2
   image_picker: ^1.0.0
   json_annotation: ^4.9.0
-  mime: ^1.0.5
+  mime: ^2.0.0
```
gnprice, line 53: The breaking change here is in `extensionFromMime`.

Looks like we don't use that method:
```
$ git grep extensionFromMime
$
```
so doesn't affect us.

pubspec.yaml
```diff
@@ -97,11 +97,11 @@ dev_dependencies:
   drift_dev: ^2.5.2
   fake_async: ^1.3.1
   flutter_checks: ^0.1.1
-  flutter_lints: ^4.0.0
+  flutter_lints: ^5.0.0
```
gnprice, line 100: By nature if this doesn't cause CI to fail, it can't have broken anything.

There is one interesting change:
> * Removes the following lints (see https://github.com/dart-lang/lints/issues 205):
>   * `prefer_const_constructors`
>   * `prefer_const_declarations`
>   * `prefer_const_literals_to_create_immutables`

So that ends the regular nagging from the analyzer to add `const` everywhere. As discussed in the thread, it's not clear that the performance benefit of it is/was material.

gnprice (commented):
Thanks @rajveermalviya for taking care of this, and @PIG208 for the previous review and the manual testing!

Generally this all looks good — one request below to split up a commit. The other comments are all just me reading the changelogs, the results of which are all that there's nothing more we need to do. I can add that information to the commit messages before merge.

> The app also works fine on my Android device. It would be great to have it tested on someone's iOS device too.

For this set of upgrades I'm content with that testing you did, and not doing additional manual testing on iOS.

The main area where I'd be concerned about platform-specific issues that call for manual testing on both platforms is notifications. In this upgrade, the changelogs for the notification-related (Firebase-related) upgrades look quite benign.

pubspec.yaml
```diff
@@ -14,8 +14,8 @@ environment:
   # We use a recent version of Flutter from its main channel, and
   # the corresponding recent version of the Dart SDK.
   # Feel free to update these regularly; see README.md for instructions.
-  sdk: '>=3.7.0-267.0.dev <4.0.0'
-  flutter: '>=3.28.0-2.0.pre.38575'  # 65ff060283e19423e9538c18c24e44495b70aeff
+  sdk: '>=3.7.0-312.0.dev <4.0.0'
+  flutter: '>=3.28.0-2.0.pre.38699'  # d14140f85439c517c98b0c40f8eaf942fcb46c74
```
chrisbobbe, line 18: Huh, I guess we still can't do #1204 then? 🤷‍♂️

rajveermalviya:

android/app/src/main/kotlin/com/zulip/flutter/ZulipPlugin.kt
```diff
@@ -221,7 +221,7 @@ private class AndroidNotificationHost(val context: Context)
                 val style = NotificationCompat.MessagingStyle(toAndroidPerson(messagingStyle.user))
                     .setConversationTitle(messagingStyle.conversationTitle)
                     .setGroupConversation(messagingStyle.isGroupConversation)
-                messagingStyle.messages.forEach { it?.let {
+                messagingStyle.messages.forEach { it.let {
```
gnprice, line 224: After this change, the `it.let {` isn't doing anything, so that layer of wrapping can be cut out.

I'll do that in a separate commit, since I'm revising this to add text to the commit messages anyway.

gnprice (commented):
Thanks for the revision!

Looks good; one nit below. I'll fix that, add to those commit messages what I found above, and merge.

PR #1282, opened 2025-01-15
action_sheet [nfc]: Remove an assumption that popular emoji are Unicode

chrisbobbe:
Currently the popular emoji are all Unicode emoji. But in a possible future where they might not be (e.g. if done by "recently used") we won't want this to break.


PIG208:
This seems helpful. Was it intentionally closed?

chrisbobbe:
Yeah—it turns out there's more code that relies on it being Unicode emoji, and there's already an `assert` to check the assumption, so it won't silently start being buggy. Rather than start reworking that other code, I decided to focus on higher priorities :)

PR #1288, opened 2025-01-17
Started working #83

### LINKED ISSUE #82: Open a message list in the middle of history, when requested

### LINKED ISSUE #83: Control scroll position on new and newly-fetched messages
Currently if you're looking at the message list and then a new message comes in, you'll see the messages on the screen jump. This is exactly the right thing (*) if you're scrolled to the bottom — the new message appears at the bottom, and the existing messages all move up to make room. But if you're scrolled further up, then we should keep steady on the screen the messages you're looking at.

Instead, the current behavior is that you see all the messages jump up.

Fundamentally what's happening is that we use a (variant of a) [ListView](https://api.flutter.dev/flutter/widgets/ListView-class.html), which models the contents as a sequence of items indexed 0, 1, 2, 3, etc.; and then we pass `reverse: true` so that item 0 is shown at the bottom, and interpret item index `i` as "the i'th message in the list, counting zero-based from the end". So if we're looking around item 37 (the message 37 slots from the end), and then a new message arrives, then as far as the ListView knows we should still be looking around item 37 — but that's now the message that was item 36 a moment ago.

A good 90% solution here would be to build the message list out of two list slivers facing back to back, rather than one list sliver facing upward. This is a standard Flutter solution for having a list that grows in both directions; see [example in the CustomScrollView doc](https://api.flutter.dev/flutter/widgets/CustomScrollView-class.html#widgets.CustomScrollView.2).

(The list sliver class in that example is `SliverList`, which corresponds to a basic `ListView`. We'd use our `SliverStickyHeaderList`, which corresponds to our `StickyHeaderListView`.)

(*) Well, possibly it'd be better to have some kind of sliding animation rather than a sudden jump. But that'd be its own issue. This issue is about the new state of things that would be the endpoint of any animation.

---

Notes on possible followups:

That two-sliver solution would completely solve the problem for adding more messages (or removing messages) at the very top or bottom, which covers (a) `message` events and (b) fetching messages on scroll, #78. We'd still have some jumps on (c) deleting messages, or moving them away, when they're not the very oldest or latest; (d) moving messages into the middle of the list; (e) likely on some cases of editing messages so that their height changes; (f) similarly on a message going from no reactions to some reactions or back, or other changes that affect its displayed height.

For (c) and (d), I think the Flutter API has the hooks we'd need (we'd make our own implementation of `RenderSliverBoxChildManager`), but fundamentally it's a subtle problem; in particular we'd need a richer data structure for the list of messages than a `List` or `Queue`.

For (e) and (f), and for that matter as a possible alternative solution for (c) and (d) or even (a) and (b), I believe we could use a [ScrollController](https://api.flutter.dev/flutter/widgets/ScrollController-class.html). When something changes, we'd calculate what the right scroll position is in the new state of things, and call `jumpTo` to go there. Or perhaps we'd override [`ScrollController.createScrollPosition`](https://api.flutter.dev/flutter/widgets/ScrollController/createScrollPosition.html)? The example of `PageController`, mentioned in that doc, sounds potentially promising as a model.

E-m-i-n-e-n-c-e:
Closes #83

 Closes #82
 
 Started working on this.
 Here's what i did

- Added an optional parameter anchor to MessageList and started fetching newer Messages similar to how older messages were being fetched around the anchor(if one is given)

- Changed the single sliver to two slivers that grow in opposite directions. The center is after the anchor message with newer messages growing downward in the bottom sliver

- Handled the case of when user is looking at the latest messages by adding a condition that when the scroll is less than (viewport size + 300 px) away from the bottom it scrolls down to the bottom. This is required now as adding new messages doesn't cause the messages to jump up. Fetching newer messages causes no jump or scroll though

- The navigateToBottom button was initially set to scroll down to 0 which was after the changes was only scrolling down to the zero index of the toplist so i changed it to scrolldown to maxScrollExtent.

But still the navigate button works correctly when new messages are fetched as part of the scrolling process but when a single message is added to the list the maxScrollExtent does not scroll all the way down as it (for some reason) calculates the maxScrollExtent wrong. 

What I did to fix the above problem was to call the scrollDown function repeatedly until the end is reached. 

Videos:

https://github.com/user-attachments/assets/a379256e-3208-48b5-960c-7740a06fe497

https://github.com/user-attachments/assets/9feec67c-380e-4e92-9631-37a58ec8d8ab

Chat:

https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Control.20scroll.20position.20on.20new.20and.20newly-fetched.20messages

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -56,9 +56,13 @@ class MessageListLoadingItem extends MessageListItem {
   final MessageListDirection direction;
 
   const MessageListLoadingItem(this.direction);
+
 }
 
-enum MessageListDirection { older }
+enum MessageListDirection {
+  older,
+  newer
+}
 
 /// Indicates we've reached the oldest message in the narrow.
 class MessageListHistoryStartItem extends MessageListItem {
@@ -85,9 +89,7 @@ mixin _MessageSequence {
   bool _fetched = false;
 
   /// Whether we know we have the oldest messages for this narrow.
-  ///
-  /// (Currently we always have the newest messages for the narrow,
-  /// once [fetched] is true, because we start from the newest.)
+
   bool get haveOldest => _haveOldest;
   bool _haveOldest = false;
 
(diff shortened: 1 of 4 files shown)


gnprice:
Thanks @E-m-i-n-e-n-c-e for giving this a try. Closing in favor of #1517 and the series of PRs that led up to it (including #1507 which fixed most of #83).

PR #1296, opened 2025-01-21
Respect realm setting for mandatory topics

PIG208:
This tracks `mandatory_topics` from realm settings and set up the compose box to read it. This does not yet track for updates to the setting (that'd be #668, out-of-scope for this PR).

This prepares for #1250.


test/widgets/compose_box_test.dart
```diff
@@ -75,6 +62,25 @@ void main() {
     controller = tester.state<ComposeBoxState>(find.byType(ComposeBox)).controller;
   }
 
+  Future<void> prepare(WidgetTester tester, {
```
chrisbobbe, line 67:
```
compose test [nfc]: Extract prepareComposeBox
```

In a file that's entirely about compose-box tests (compose_box_test.dart), and in the same scope (just at the top of the `main` method), it feels confusing to have one function named `prepare` and another named `prepareComposeBox`. If I were writing a test, I don't think I would know which function to use, except by reading their implementations.

How about instead of this refactor, we make `prepare` more flexible (if needed) so it can be used in the new tests?

test/widgets/compose_box_test.dart
```diff
@@ -48,6 +48,8 @@ void main() {
     if (narrow case ChannelNarrow(:var streamId) || TopicNarrow(: var streamId)) {
       assert(store.streams.values.any((stream) => stream.streamId == streamId),
         'Add a channel with "streamId" the same as of $narrow.streamId to the store.');
+      assert(store.users.values.any((user) => user.userId == account.userId),
+        'Add an user with "userId" the same as of $account.userId to the store.');
```
chrisbobbe, line 52: nit: "a user"

lib/widgets/compose_box.dart
```diff
@@ -89,13 +89,15 @@ enum TopicValidationError {
...
 class ComposeTopicController extends ComposeController<TopicValidationError> {
-  ComposeTopicController() {
+  ComposeTopicController({required this.store}) {
     _update();
   }
 
+  PerAccountStore store;
+
   // TODO: subscribe to this value:
   //   https://zulip.com/help/require-topics
```
chrisbobbe, line 99: What's the condition for this existing TODO to be finished, and can we be more concrete about it? It looks like it's meant to be a `TODO(#668)`, right? Does `ComposeTopicController` need to add a `PerAccountStore` listener?

lib/widgets/compose_box.dart
```diff
@@ -1304,16 +1309,17 @@ abstract class ComposeBoxState extends State<ComposeBox> {
...
 
   @override
-  void initState() {
-    super.initState();
+  void onNewStore() {
     switch (widget.narrow) {
       case ChannelNarrow():
-        _controller = StreamComposeBoxController();
+        final store = PerAccountStoreWidget.of(context);
+        _controller ??= StreamComposeBoxController(store: store);
```
chrisbobbe, line 1321: Does the old value of `_controller` (if non-null) need to be `dispose()`d?

chrisbobbe (commented):
Thanks! Small comments below.

lib/widgets/compose_box.dart
PIG208: I think `_controller` shouldn't be disposed until the `ComposeBox` is disposed; we only set it once when initializing, and update its reference to the store if there is a new store.

chrisbobbe: Ah right, I see; I read too fast and thought we were replacing the controller here, rather than updating its reference to the store.

lib/widgets/compose_box.dart
PIG208: > Does ComposeTopicController need to add a PerAccountStore listener?

Right. We may include that as a part of #668 or its follow-up. This would be something for the future us to triage. I will update to mention the issue.

test/widgets/compose_box_test.dart
PIG208: The assertion will be removed as `account` is replaced with `zulipFeatureLevel`.

PIG208:

test/widgets/compose_box_test.dart
```diff
@@ -44,7 +44,6 @@ void main() {
   Future<void> prepareComposeBox(WidgetTester tester, {
     required Narrow narrow,
     User? selfUser,
-    int? realmWaitingPeriodThreshold,
```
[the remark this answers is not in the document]
chrisbobbe: Bump on truncating commit IDs to 9 characters :)

test/widgets/compose_box_test.dart
```diff
@@ -44,7 +44,7 @@ void main() {
   Future<void> prepareComposeBox(WidgetTester tester, {
     required Narrow narrow,
     User? selfUser,
-    List<User> users = const [],
+    List<User> otherUsers = const [],
```
chrisbobbe, line 47: nit: the commit message says `otherUser` which doesn't match this name

chrisbobbe (commented):
Thanks! LGTM except some nits; I'll mark for Greg's review.

test/widgets/compose_box_test.dart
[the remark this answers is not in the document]
PIG208: Oops, missed that. Fixed!

gnprice:
Looks like CI is failing, and there are also a number of added commits that weren't in yesterday's revision of the branch. Are those meant for a later PR?

PIG208:
Ahh yes. Those are for general chat (whose base is the tip of this PR).

test/widgets/compose_box_test.dart
```diff
@@ -560,6 +561,64 @@ void main() {
...
+      check(connection.lastRequest).isA<http.Request>()
+        ..method.equals('POST')
+        ..url.path.equals('/api/v1/messages')
+        ..bodyFields.deepEquals({
+          'type': 'stream',
+          'to': channel.streamId.toString(),
+          'topic': '(no topic)',
+          'content': 'test content',
+          'read_by_sender': 'true',
+        });
```
gnprice, line 604: Let's simplify this to focus on just the part that this test is about:
```suggestion
        ..bodyFields['topic'].equals('(no topic)');
```

(The `.method` and `.url.path` checks probably still belong, as they're part of confirming that the request was to the expected endpoint — effectively they provide the namespace that the parameter name `'topic'` draws its meaning from.)

gnprice (commented):
Thanks both! Generally this all looks good. A few comments, mainly in the tests.

PIG208:

gnprice:
Thanks! Looks good; merging.

PR #1297, opened 2025-01-22
general chat #1: add support at most places

PIG208:
~~Stacked on top of #1342.~~

This addresses part of #1250.

This is the bare minimum needed for making `TopicName.displayName` nullable before we can work on the more intricate details on how parts like sending from compose box or live hint text updates work.

The last three commits enable support for empty topics for testing. When merging, we leave those three commits out because we haven't yet finished working other other parts mentioned above. This allows us to review and merge related changes without interrupting beta releases with user-visible changes.

See the [screenshots](https://github.com/zulip/zulip-flutter/pull/1297#issuecomment-2644408826) for what changes are included here.

### DIFF
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -921,7 +921,7 @@ class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, Top
   }
 
   TopicAutocompleteResult? _testTopic(TopicAutocompleteQuery query, TopicName topic) {
-    if (query.testTopic(topic)) {
+    if (query.testTopic(topic, store)) {
       return TopicAutocompleteResult(topic: topic);
     }
     return null;
@@ -939,10 +939,17 @@ class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, Top
 class TopicAutocompleteQuery extends AutocompleteQuery {
   TopicAutocompleteQuery(super.raw);
 
-  bool testTopic(TopicName topic) {
+  bool testTopic(TopicName topic, PerAccountStore store) {
     // TODO(#881): Sort by match relevance, like web does.
+
+    // ignore: unnecessary_null_comparison // null topic names soon to be enabled
+    if (topic.displayName == null) {
+      return store.realmEmptyTopicDisplayName.toLowerCase()
+        .contains(raw.toLowerCase());
+    }
     return topic.displayName != raw
-      && topic.displayName.toLowerCase().contains(raw.toLowerCase());
+      // ignore: unnecessary_non_null_assertion // null topic names soon to be enabled
+      && topic.displayName!.toLowerCase().contains(raw.toLowerCase());
   }
 
   @override
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -302,6 +302,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
       realmMandatoryTopics: initialSnapshot.realmMandatoryTopics,
       realmWaitingPeriodThreshold: initialSnapshot.realmWaitingPeriodThreshold,
       maxFileUploadSizeMib: initialSnapshot.maxFileUploadSizeMib,
+      realmEmptyTopicDisplayName: initialSnapshot.realmEmptyTopicDisplayName,
       realmDefaultExternalAccounts: initialSnapshot.realmDefaultExternalAccounts,
       customProfileFields: _sortCustomProfileFields(initialSnapshot.customProfileFields),
       emailAddressVisibility: initialSnapshot.emailAddressVisibility,
@@ -344,6 +345,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
     required this.realmMandatoryTopics,
     required this.realmWaitingPeriodThreshold,
     required this.maxFileUploadSizeMib,
+    required String? realmEmptyTopicDisplayName,
     required this.realmDefaultExternalAccounts,
     required this.customProfileFields,
     required this.emailAddressVisibility,
@@ -362,6 +364,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
        assert(realmUrl == connection.realmUrl),
        assert(emoji.realmUrl == realmUrl),
        _globalStore = globalStore,
+       _realmEmptyTopicDisplayName = realmEmptyTopicDisplayName,
        _emoji = emoji,
        _users = users,
        _channels = channels,
@@ -414,6 +417,18 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
   /// For docs, please see [InitialSnapshot.realmWaitingPeriodThreshold].
   final int realmWaitingPeriodThreshold;  // TODO(#668): update this realm setting
   final int maxFileUploadSizeMib; // No event for this.
+
+  /// The display name to use for empty topics.
+  ///
+  /// This should only be accessed when FL >= 334, since topics cannot
+  /// be empty otherwise.
+  // TODO(server-10) simplify this
+  String get realmEmptyTopicDisplayName {
+    assert(_realmEmptyTopicDisplayName != null); // TODO(log)
+    return _realmEmptyTopicDisplayName ?? 'general chat';
+  }
+  final String? _realmEmptyTopicDisplayName; // TODO(#668): update this realm setting
+
   final Map<String, RealmDefaultExternalAccount> realmDefaultExternalAccounts;
   List<CustomProfileField> customProfileFields;
   /// For docs, please see [InitialSnapshot.emailAddressVisibility].
--- a/lib/widgets/inbox.dart
+++ b/lib/widgets/inbox.dart
@@ -532,12 +532,15 @@ class _TopicItem extends StatelessWidget {
                 style: TextStyle(
                   fontSize: 17,
                   height: (20 / 17),
+                  // ignore: unnecessary_null_comparison // null topic names soon to be enabled
+                  fontStyle: topic.displayName == null ? FontStyle.italic : null,
                   // TODO(design) check if this is the right variable
                   color: designVariables.labelMenuButton,
                 ),
                 maxLines: 2,
                 overflow: TextOverflow.ellipsis,
-                topic.displayName))),
+                // ignore: dead_null_aware_expression // null topic names soon to be enabled
+                topic.displayName ?? store.realmEmptyTopicDisplayName))),
             const SizedBox(width: 12),
             if (hasMention) const _IconMarker(icon: ZulipIcons.at_sign),
             // TODO(design) copies the "@" marker color; is there a better color?
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -339,8 +339,11 @@ class MessageListAppBarTitle extends StatelessWidget {
     return Row(
       mainAxisSize: MainAxisSize.min,
       children: [
-        Flexible(child: Text(topic.displayName, style: const TextStyle(
+        // ignore: dead_null_aware_expression // null topic names soon to be enabled
+        Flexible(child: Text(topic.displayName ?? store.realmEmptyTopicDisplayName, style: TextStyle(
           fontSize: 13,
+          // ignore: unnecessary_null_comparison // null topic names soon to be enabled
+          fontStyle: topic.displayName == null ? FontStyle.italic : null,
         ).merge(weightVariableTextStyle(context)))),
         if (icon != null)
           Padding(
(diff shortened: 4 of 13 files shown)


PIG208:
I expect #1301 to be merged pretty soon, but the rest of the PR should be ready for review.

PIG208:
Will rebase and add screenshots shortly.

PIG208:
Screenshots:

<details>
<summary>Autocomplete</summary>

| | |
| - | - |
| autocomplete filtered | auto complete topic |

</details>

<details>
<summary>Message list page</summary>

| | |
| - | - |
| channel narrow empty compose | topic narrow |

</details>

<details>
<summary>Inbox</summary>

inbox

</details>

<details>
<summary>No resolve button</summary>

lack of resolve button

</details>

PIG208:
Will update the PR to match the behavior from https://github.com/zulip/zulip/pull/33338:
If topics are not mandatory:
- show "test general chat" in solid color when topic input is blurred;
- show the default placeholder in original placeholder color is focused.

PIG208:
Starting [a discussion on CZO](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/general.20chat.20design.20.23F1297) to nail the design details.

lib/model/store.dart
```diff
@@ -413,6 +416,11 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, ChannelStore, Mess
   /// For docs, please see [InitialSnapshot.realmWaitingPeriodThreshold].
   final int realmWaitingPeriodThreshold;  // TODO(#668): update this realm setting
   final int maxFileUploadSizeMib; // No event for this.
+  final String? _realmEmptyTopicDisplayName; // TODO(#668): update this realm setting
+  String get realmEmptyTopicDisplayName {
+    assert(_realmEmptyTopicDisplayName != null); // TODO(log)
+    return _realmEmptyTopicDisplayName ?? 'general chat';
+  }
```
chrisbobbe, line 423: nit: getter first: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense

test/widgets/autocomplete_test.dart
```diff
@@ -352,24 +352,24 @@ void main() {
       await tester.pumpAndSettle();
 
       // "topic three" and "topic two" appear, but not "topic one"
-      checkTopicShown(topic1, store, expected: false);
-      checkTopicShown(topic2, store, expected: true);
-      checkTopicShown(topic3, store, expected: true);
+      checkTopicShown('Topic one',   store, expected: false);
+      checkTopicShown('Topic two',   store, expected: true);
+      checkTopicShown('Topic three', store, expected: true);
```
chrisbobbe: autocomplete test [nfc]: Use plain strings for finding topic
    
    instead of inferring it from a TopicName
    
Commit-message nit: I think "inferring" isn't quite the right word; if you have a `TopicName` and you want its display name, you just look at its `displayName` field; no inference is needed

lib/model/autocomplete.dart
```diff
@@ -921,7 +921,7 @@ class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, Top
   }
 
   TopicAutocompleteResult? _testTopic(TopicAutocompleteQuery query, TopicName topic) {
-    if (query.testTopic(topic)) {
+    if (query.testTopic(store, topic)) {
```
chrisbobbe, line 924: autocomplete: Handle autocompleting to empty topic
    
This commit is NFC, right?

lib/model/autocomplete.dart
```diff
@@ -947,6 +947,8 @@ class TopicAutocompleteQuery extends AutocompleteQuery {
       return store.realmEmptyTopicDisplayName.toLowerCase()
         .contains(raw.toLowerCase());
     }
+    // This checks for inequality because there is nothing to autocomplete to
+    // when `raw` already matches the topic exactly.
     return topic.displayName != raw
```
chrisbobbe, line 951: How about:

```dart
    // Skip this option if it matches the query exactly
```

(a little less verbose)

lib/widgets/inbox.dart
```diff
@@ -537,13 +537,16 @@ class _TopicItem extends StatelessWidget {
               child: Text(
                 style: TextStyle(
                   fontSize: 17,
+                  // ignore: unnecessary_null_comparison // null topic names soon to be enabled
+                  fontStyle: topic.displayName == null ? FontStyle.italic : null,
```
chrisbobbe, line 540: msglist: Display realmEmptyTopicDisplayName where empty topic appears
    
This is another NFC commit, right? (And it doesn't actually do what the summary line says it does, because empty topics don't appear anywhere yet?)

lib/widgets/message_list.dart
```diff
@@ -366,8 +366,11 @@ class MessageListAppBarTitle extends StatelessWidget {
     return Row(
       mainAxisSize: MainAxisSize.min,
       children: [
-        Flexible(child: Text(topic.displayName, style: const TextStyle(
```
chrisbobbe, line 369: msglist: Display realmEmptyTopicDisplayName where empty topic appears
    
(Same comment about this being an NFC commit.)

lib/api/route/messages.dart
```diff
@@ -90,6 +92,7 @@ Future<GetMessagesResult> getMessages(ApiConnection connection, {
   bool? applyMarkdown,
   // bool? useFirstUnreadAnchor // omitted because deprecated
 }) {
+  final supportsEmptyTopics = connection.zulipFeatureLevel! >= 334; // TODO(server-10)
```
chrisbobbe, line 95: `clientGravatar` looks like another client-capability param, but that one is exposed to the caller of this binding. I think we should maybe also expose `supportsEmptyTopics`, to make the binding as thin as possible? (Here and in other bindings)

chrisbobbe (commented):
Thanks, and sorry for the delay in my review! Small comments below.

lib/widgets/inbox.dart
PIG208: Yeah, these commits are structured in a way so that they are not user visible without the final "do not merge" series. This way we can start merging part of it while still have a way to test them.

It will do what the summary says, but only after we allow empty topics in our data structures.

lib/api/route/messages.dart
PIG208: Ended up rewriting most of the last commit, and added a bunch of new tests.

PIG208:
Thanks for the review! This should be ready again.

chrisbobbe:
Thanks, looks mostly good! For the NFC commits:

autocomplete [nfc]: Handle autocompleting to empty topic

inbox [nfc]: Display realmEmptyTopicDisplayName where empty topic appears
msglist [nfc]: Display realmEmptyTopicDisplayName where empty topic appears

I see that you've explained them in the PR discussion:

https://github.com/zulip/zulip-flutter/pull/1297#discussion_r1970805946
> Yeah, these commits are structured in a way so that they are not user visible without the final "do not merge" series. This way we can start merging part of these changes while still have a way to test them.
> 
> It will do what the summary says, but only after we allow empty topics in our data structures.

But the commit messages themselves look confusing; they say NFC but they also read as descriptions of a user-facing change. How about adding a brief note in each commit-message body saying that the described change doesn't actually happen until we allow empty topics in our data structures, which will come in a later series of commits.

lib/model/store.dart
```diff
@@ -414,6 +417,11 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
   /// For docs, please see [InitialSnapshot.realmWaitingPeriodThreshold].
   final int realmWaitingPeriodThreshold;  // TODO(#668): update this realm setting
   final int maxFileUploadSizeMib; // No event for this.
+  String get realmEmptyTopicDisplayName {
+    assert(_realmEmptyTopicDisplayName != null); // TODO(log)
```
gnprice, line 421: This calls for a comment explaining why this condition is supposed to be unreachable — it doesn't seem obvious.

lib/model/store.dart
```diff
@@ -414,6 +417,11 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
...
+    return _realmEmptyTopicDisplayName ?? 'general chat';
+  }
+  final String? _realmEmptyTopicDisplayName; // TODO(#668): update this realm setting
```
gnprice, line 424: nit: set this group off with blank lines

lib/model/store.dart
gnprice: Or maybe clearest: a bit of dartdoc saying not to access this unless zulipFeatureLevel is at least such-and-such. That's the idea for why this shouldn't happen, right?

test/widgets/autocomplete_test.dart
```diff
@@ -406,5 +401,44 @@ void main() {
...
+    testWidgets('match general chat in autocomplete', (tester) async {
+      final topic = eg.getStreamTopicsEntry(name: '');
+      final topicInputFinder = await setupToTopicInput(tester, topics: [topic]);
+
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(topicInputFinder, 'general ch');
+      await tester.enterText(topicInputFinder, 'general cha');
+      await tester.pumpAndSettle();
+
+      check(find.text(eg.defaultRealmEmptyTopicDisplayName)).findsOne();
```
gnprice, line 426: This test is already relying on the value of `eg.defaultRealmEmptyTopicDisplayName` being what it is (or something like it), namely that it contains "general chat", in order for the search steps to work.

So the test should instead give explicitly the value of `realmEmptyTopicDisplayName` that it's counting on. Then it can use that in this check too.

lib/model/autocomplete.dart
```diff
@@ -939,10 +943,16 @@ class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, Top
...
-  bool testTopic(TopicName topic) {
+  bool testTopic(PerAccountStore store, TopicName topic) {
     // TODO(#881): Sort by match relevance, like web does.
+
+    if (topic.displayName == null) {
+      return store.realmEmptyTopicDisplayName.toLowerCase()
+        .contains(raw.toLowerCase());
+    }
+    // Skip this option if it matches the query exactly.
     return topic.displayName != raw
```
gnprice, line 954: This comment seems to just restate what the line after it says. Cleanest to leave it out.

lib/widgets/inbox.dart
```diff
@@ -531,13 +531,14 @@ class _TopicItem extends StatelessWidget {
               child: Text(
                 style: TextStyle(
                   fontSize: 17,
+                  fontStyle: topic.displayName == null ? FontStyle.italic : null,
                   height: (20 / 17),
```
gnprice, line 535: nit: keep fontSize and height next to each other — the "17" in `height` is a reference to the "17" in `fontSize`

(And even if we hadn't written `height` that way, those two properties are closely related and one naturally wants to read them together)

test/widgets/inbox_test.dart
```diff
@@ -307,6 +307,16 @@ void main() {
...
     });
 
+    testWidgets('empty topic', (tester) async {
+      final channel = eg.stream();
+      await setupPage(tester,
+        streams: [channel],
+        subscriptions: [(eg.subscription(channel))],
+        unreadMessages: [eg.streamMessage(stream: channel, topic: '')]);
+
+      check(find.text(eg.defaultRealmEmptyTopicDisplayName)).findsOne();
```
gnprice, line 317: nit: keep checks joined into same stanza after the setup that leads to them
```suggestion
        unreadMessages: [eg.streamMessage(stream: channel, topic: '')]);
      check(find.text(eg.defaultRealmEmptyTopicDisplayName)).findsOne();
```

lib/widgets/message_list.dart
```diff
@@ -1092,11 +1093,13 @@ class StreamMessageRecipientHeader extends StatelessWidget {
...
         children: [
           Flexible(
-            child: Text(topic.displayName,
+            child: Text(topic.displayName ?? store.realmEmptyTopicDisplayName,
               // TODO: Give a way to see the whole topic (maybe a
               //   long-press interaction?)
               overflow: TextOverflow.ellipsis,
-              style: recipientHeaderTextStyle(context))),
+              style: recipientHeaderTextStyle(context).copyWith(
+                fontStyle: topic.displayName == null ? FontStyle.italic : null,
```
gnprice, line 1101: Instead of copying the style object right after making it, let's have this `recipientHeaderTextStyle` helper take a parameter. Can just be `fontStyle` directly.

lib/widgets/message_list.dart
```diff
@@ -339,8 +339,9 @@ class MessageListAppBarTitle extends StatelessWidget {
     return Row(
       mainAxisSize: MainAxisSize.min,
       children: [
-        Flexible(child: Text(topic.displayName, style: const TextStyle(
+        Flexible(child: Text(topic.displayName ?? store.realmEmptyTopicDisplayName, style: TextStyle(
```
gnprice, line 342: Let's have a test for this part too. Should be quick, I think; can go in the "app bar" group.

gnprice (commented):
Thanks to you both! Generally this looks good; comments below.

(I haven't read the "do not merge" commits.)

PIG208:

Issue #1008, opened 2024-10-19
msglist: For bulk moves with `PropagateMode.changeLater`, don't renarrow if user was reading earlier messages


chrisbobbe:
When messages are moved out of the current narrow with `PropagateMode.changeLater`, we always follow the move by renarrowing `MessageListPage`. If the user is scrolled up to much earlier messages than the ones that were moved, the context change will be unhelpful and confusing.

To fix, we can guess which onscreen message the user is reading and only renarrow if that message was moved. We can't always guess correctly, and that's fine. (When more than one message is visible, the user's eyes could be on any of them.) But a simple algorithm would be to choose the first message that's at least partly shown in the viewport. If screen-reader software is active and a message is selected, that message is probably a better choice.

Marking post-launch because zulip-mobile doesn't do this.

Issue #901, opened 2024-08-21
Handle moved messages in unreads data


gnprice:
With #787 we implemented most of #150, handling moved messages. But as that PR description said:
> Fixes most of https://github.com/zulip/zulip-flutter/issues/150. (After this lands we should do a quick scan to identify anything else that needs updating on message moves — there are several new data structures in the store that we've merged in recent weeks.)

Then we didn't do that scan.

Places that still need updating on message moves include:
- [x] `Unreads` (which has a TODO comment for it)
- [x] ~~`RecentSenders` (that's one of those recently-merged data structures)~~ → split to its own issue: #1460 
- … OK and I think that's it, having now scanned through what's in the store. (Another recently-merged data structure is `RecentDmConversationsView`, but that's only about DMs and those can't be moved.)

Related issues:

- #862

gnprice:
The key user-facing scenario where this comes up, which I happened to see again earlier today:
* You're reading messages in Combined feed.
* A topic gets resolved, containing some unread messages.
* The messages are still unread, and shown with the blue unread marker at left. If you tap the recipient header to navigate to their conversation, the unread marker appears, as it should.
* But the conversation *doesn't* have a mark-as-read button. As a result there's no way for you to dismiss those unread markers (short of marking read the entire combined feed, or entire channel).

What's happening is that the `Unreads` data structure still has the messages under their old, unresolved topic; so when viewing the new resolved topic, it finds no unreads for that topic.

PIG208:
I will work on this next as with #1301 this scenario might come up more often.

PIG208:
`TypingStatus` also refers to data that can be moved (`TopicNarrow`s), but it appears that the web app doesn't move an active typing indicator even if propagate mode is 'all'. It could be a deliberate UX or a known/unknown issue. Anyway, this is pretty subtle and we shouldn't need to consider it when working on this.

gnprice:
I've split out the recent-senders part of this issue as its own, M6 post-launch issue #1460, for the reasons at https://github.com/zulip/zulip-flutter/pull/1418#issuecomment-2774549920 .

This issue is now purely about the unreads data, which was an M5a launch-blocker issue. It was fixed by #1311.

Issue #862, opened 2024-08-02
Update topic input in channel compose box when topic moved


gnprice:
This is a followup to #150 / #787. After that issue/PR, when you're looking at a TopicNarrow and the topic gets moved (with PropagateMode.changeAll or PropagateMode.changeLater), we'll update the page's narrow to follow along, and this will also mean that the send button in the compose box will point to the new narrow.

This issue is: If you're looking instead at a ChannelNarrow and you happen to have the moved topic selected in the compose box, then we should also follow along.
* If the move happened within the channel, we should update the compose box's topic input.
* If the move was to a different channel… TBD but we should probably switch to either the new TopicNarrow or its surrounding ChannelNarrow. The point is that if you hit send just after the move happens, we want the message to go to the conversation in its new location.

This is a behavior zulip-mobile doesn't have; instead we have a TODO comment (referring to zulip/zulip-mobile#5251):
```js
          } else if (isStreamNarrow(narrow) && streamIdOfNarrow(narrow) === move.orig_stream_id) {
            // A ChatScreen showing the stream that contained the moved messages.
            //
            // TODO(#5251): Update topic input, if it matches.  (If the stream
            //   changed too, unclear what to do.  Possibly change the
            //   screen's narrow, as if narrowed to the moved topic?)
          }
        }
```

This point was previously described at https://github.com/zulip/zulip-flutter/pull/787#discussion_r1702297052 .

PR #1301, opened 2025-01-24
action_sheet: Implement resolve/unresolve in topic action sheet

chrisbobbe:
Fixes: #744 

The long-awaited resolve/unresolve button! If it's helpful, feel free to review just some of the prep commits, for a first round. I thought about sending them as a separate PR, but figured it might be helpful to see them with the context of the work they're preparing for.

Screenshots coming soon.


test/widgets/action_sheet_test.dart
```diff
@@ -143,6 +144,35 @@ void main() {
...
+        final channelIdsWithUnreads = store.unreads.streams.keys;
+        final topicWithUnreads = channelIdsWithUnreads.firstWhereOrNull((streamId) =>
+          store.unreads.countInTopicNarrow(streamId, TopicName(topic)) > 0);
+        if (topicWithUnreads == null) {
+          throw FlutterError.fromParts([
+            ErrorSummary('showFromInbox called without an unread message'),
+            ErrorHint(
+              'Before calling showFromInbox, ensure that [Unreads] '
+              'has an unread message in the relevant topic. ',
+            ),
```
PIG208, line 160: The error looks useful! Would it be necessary to wrap it in the `assert((){})`, though, as this only gets run in tests?

test/widgets/action_sheet_test.dart
```diff
@@ -108,21 +108,26 @@ void main() {
...
+
     group('showTopicActionSheet', () {
-      final channel = eg.stream();
-      const topic = 'my topic';
       final message = eg.streamMessage(
-        stream: channel, topic: topic, sender: eg.otherUser);
+        stream: someChannel, topic: someTopic, sender: eg.otherUser);
 
       Future<void> prepare({
+        ZulipStream? channel,
```
PIG208: I think none of the callers of `prepare` ends up needing `channel` to be a specific value other than `someChannel`, so maybe we can get this removed?

test/widgets/action_sheet_test.dart
```diff
@@ -113,37 +113,37 @@ void main() {
...
-        connection = store.connection as FakeApiConnection;
-      }
+    Future<void> prepare({
+      ZulipStream? channel,
+      String topic = someTopic,
+      bool isChannelSubscribed = true,
+      bool? isChannelMuted,
+      UserTopicVisibilityPolicy? visibilityPolicy,
+      UnreadMessagesSnapshot? unreadMsgs,
+      int? zulipFeatureLevel,
```
PIG208, line 124: I agree that these will potentially become useful as we add more. Maybe at some point it will be helpful to extract some of this to `test/example_data.dart` as a helper to set up global store, along with some invariant checks. Out of scope for this PR, though.

PIG208 (commented):
Thanks for the PR! I went through all the commits before the last two. Generally this looks good to me.

* f200ff9c4 action_sheet test [nfc]: Have showFromAppBar take `messages`, not `message`
* 792179074 action_sheet test: Move some logic out of checkButtons helper
* 271ae08c4 test [nfc]: Comment on where to find topic-action-sheet tests
* 34740585d action_sheet test [nfc]: Pull out showFromRecipientHeader helper
* c918101ec action_sheet test: Cut assumption of topic recipient header in topic narrow
* 3c405174b action_sheet test [nfc]: Pull out showFromAppBar helper
* d0517df41 action_sheet test [nfc]: Pull out showFromInbox helper
* 56b9f82b3 action_sheet test [nfc]: Simplify some setup by using someChannel
* ad3c6f60f action_sheet test [nfc]: Pull out a `prepare` function and reuse
* 3fc94054b action_sheet test: Use eg.otherUser for a sender in setupToTopicActionSheet
* efe13817d action_sheet test [nfc]: Add `visibilityPolicy` to a `prepare`
* 1933a2580 action_sheet test [nfc]: Add subscription-related params to a `prepare`
* 754caac25 action_sheet test [nfc]: Add zulipFeatureLevel param to a `prepare`
* 3331c2553 action_sheet test [nfc]: Pull out `someMessage` variable
* d108b96ac action_sheet test [nfc]: Parameterize a `prepare` by channel and topic
* 3c1818bf3 action_sheet test: Make some show-from-inbox setup optional in a `prepare`
* 4e92e91a4 action_sheet test [nfc]: Add new '{topic,message} action sheet' groups
* b7ddabce6 action_sheet test [nfc]: Make checkButtons helper for showTopicActionSheet
* 41c7e3d98 action_sheet test: Make the app-bar topic-row finder more precise
* e205d7433 icons: Add resolve/unresolve-topic icons, from Figma

I also tested this manually. It works great except that #150 and #862 become more noticeable. So maybe we should prioritize handling those as well (#150 is already M5-a, #862 is currently M7). Will look at the implementation during the next round of review.

chrisbobbe:
> I also tested this manually. It works great except that #150 and #862 become more noticeable.

Is #150 the issue you meant? That issue is closed as fixed.

lib/widgets/inbox.dart
```diff
@@ -13,14 +13,34 @@ import 'text.dart';
...
+  ///
+  /// Uses the inefficient [BuildContext.findAncestorStateOfType];
+  /// don't call this in a build method.
+  // If we do find ourselves wanting this in a build method, it won't be hard
+  // to enable that: we'd just need to add an [InheritedWidget] here.
+  static InboxPageState ancestorOf(BuildContext context) {
+    final state = context.findAncestorStateOfType<_InboxPageState>();
+    assert(state != null, 'No InboxPageBody ancestor');
+    return state!;
+  }
```
gnprice, line 37: As we discussed just now in the office, the problem this is solving makes sense but I'd like to avoid having to add these for all the different page widgets when there isn't another reason for them. And one way to handle that generically would be:
* make a marker widget with a name like `PageRoot` — can even be an inherited widget
* put that in `AccountPageRouteMixin.buildPage`, inside the PerAccountStoreWidget and just outside the child widget built for the page contents
* also put that in `_MessageListPageState.build`, around everything
* then just look up the ancestor widget of that type.

gnprice (commented):
Thanks! Two comments from a high-level skim. One comment below, and:

Let's make a first PR have the series of test refactors, and then a second PR on top have the other commits. I think that'll be a good split.

(Sending both PRs at once can be a good way to handle the point you mentioned about seeing the context of how the early changes end up getting used.)

PIG208:
> Is https://github.com/zulip/zulip-flutter/issues/150 the issue you meant? That issue is closed as fixed.

Ah, I meant #901. The issue for updating topic input on message move.

test/widgets/action_sheet_test.dart
chrisbobbe: Oh indeed; I think it doesn't need the `assert`.

test/widgets/action_sheet_test.dart
chrisbobbe: I think I slightly prefer leaving it in, in case later tests want to use it? It would be easy to remove, but harder to add back correctly (without forgetting e.g. to use `effectiveChannel` everywhere instead of just `channel`, which the analyzer wouldn't help with because they're both in scope).

chrisbobbe:
> Let's make a first PR have the series of test refactors, and then a second PR on top have the other commits. I think that'll be a good split.

Sure! I've just sent #1306 with the test refactors, revised after Zixuan's review above.

lib/widgets/action_sheet.dart
```diff
@@ -377,6 +387,106 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+    final store = PerAccountStoreWidget.of(pageContext);
+    final message = store.messages[someMessageIdInTopic] as StreamMessage?;
+
+    if (message != null && !message.topic.isSameAs(topic)) {
+      // TODO also interrupt on a channel move, with a specific message?
+
+      // The message's topic doesn't match what it was when we opened the
+      // action sheet. So either the resolve/unresolve action was already done
+      // or the topic was renamed; either way, tell the user and don't make a
+      // request.
```
PIG208, line 434: This is a good thought. I haven't gone through the PR yet, but we don't already do this for other topic actions like changing topic visibility. Maybe that can be a follow-up issue? Some bits from this if-statement might be reusable in that case.

chrisbobbe: Added a comment about this.

chrisbobbe:

PIG208 (commented):
Thanks for the update! This looks promsing. Posted some comments about certain edge cases, and a question on the implementation of `PageRoot`.

lib/widgets/message_list.dart
```diff
@@ -277,48 +277,51 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
...
-                  key: _messageListKey,
-                  narrow: narrow,
-                  onNarrowChanged: _narrowChanged,
-                ))),
-            if (ComposeBox.hasComposeBox(narrow))
-              ComposeBox(key: _composeBoxKey, narrow: narrow)
-          ])));
+    // The Builder enables MessageListPage.ancestorOf
+    // on the context provided to the message action sheet by PageRoot.
+    return Builder(builder: (context) => PageRoot(context: context,
```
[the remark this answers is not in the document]
chrisbobbe: > `PageRoot` sounds like a neat concept!

Yeah! Greg's idea: https://github.com/zulip/zulip-flutter/pull/1301#discussion_r1929349531

After talking with him in the office just now, the conclusion is that we don't need `contextOf` to add a build dependency (with something like `dependOnInheritedWidgetOfExactType`). 

So just `getElementForInheritedWidgetOfExactType` should suffice.

lib/api/model/model.dart
```diff
@@ -695,6 +695,12 @@ extension type const TopicName(String _value) {
   /// The key to use for "same topic as" comparisons.
   String canonicalize() => apiName.toLowerCase();
 
+  /// Whether the topic starts with [resolvedTopicPrefixRegexp].
+  bool get isResolved => resolvedTopicPrefixRegexp.hasMatch(_value);
```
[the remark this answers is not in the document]
chrisbobbe: Indeed, thanks for the catch!!

lib/widgets/action_sheet.dart
```diff
@@ -377,6 +387,106 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+      ? zulipLocalizations.actionSheetOptionResolveTopic
+      : zulipLocalizations.actionSheetOptionUnresolveTopic;
+  }
+
+  @override void onPressed() async {
+    final zulipLocalizations = ZulipLocalizations.of(pageContext);
+    final store = PerAccountStoreWidget.of(pageContext);
+    final message = store.messages[someMessageIdInTopic] as StreamMessage?;
+
+    if (message != null && !message.topic.isSameAs(topic)) {
```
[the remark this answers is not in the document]
chrisbobbe: This sounds very uncommon.

One thought that this case prompts: while this revision correctly shows these error dialogs when they apply:
- 'This topic is already marked as unresolved.'
- 'This topic is already marked as resolved.'
- 'This topic has been renamed.'

it's missing a condition for an error dialog like:

- 'One or more messages in this topic were moved.'

That's a valid thing that can happen, and in principle I think it matters as much as the other cases do. But it's not practical to implement fully correctly, because we'd need to check each message with the old topic, and we don't have all that message data. We don't have that data for good reasons, and we don't want to bother fetching it. (For one thing, it could be a huge number of messages.)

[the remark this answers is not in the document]
chrisbobbe: In your repro recipe, it seems like the unexpected behavior (from the user's view) doesn't involve any messages being set to a topic the user didn't intend, right? The old (not-moved) message(s) get their topic set appropriately, and the later (moved) messages don't. But the user is now looking at the later messages (or might be—see #1008), and they can choose to resolve that new topic too if they want to.

chrisbobbe:
Thanks for the review, revision pushed!

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
PIG208: > The old (not-moved) message(s) get their topic set appropriately, and the later (moved) messages don't.

This was unexpected for me because I though the topic action sheet (when opened from a MessageListPage) refers to the topic of the active MessageListPage, whether it has been moved or not. So anything that happens as a result of an interaction with the action sheet is immediately visible to the user, but in this case that happens in background on a topic that is no longer on screen.

assets/l10n/app_en.arb
```diff
@@ -88,6 +88,42 @@
...
+  "@topicAlreadyResolvedMessage": {
+    "description": "Error message when trying to 'Mark as resolved' a topic that was already marked by someone else."
+  },
+  "topicAlreadyUnresolvedMessage": "This topic is already marked as unresolved.",
+  "@topicAlreadyUnresolvedMessage": {
+    "description": "Error message when trying to 'Mark as unresolved' a topic that was already marked by someone else."
+  },
+  "topicRenamedMessage": "This topic has been renamed.",
+  "@topicRenamedMessage": {
+    "description": "Error title when trying to 'Mark as resolved' or 'Mark as unresolved' a topic that was renamed by someone else."
```
PIG208, line 117: nit: "Error title" -> "Error message"

PIG208 (commented):
Thanks for the revision! Looks good to me except one more nit.

lib/widgets/action_sheet.dart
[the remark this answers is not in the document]
chrisbobbe: Hmm, I see; the change of context (changing the narrow in response to the bulk-move) makes things confusing at the moment the action completes, because the effect isn't visible in the new context.

chrisbobbe:
Thanks! Revision pushed, and I see you've marked it for Greg's review. Highlighting the above thread for Greg in case he has thoughts: https://github.com/zulip/zulip-flutter/pull/1301#discussion_r1934944417

lib/widgets/action_sheet.dart
```diff
@@ -372,37 +391,145 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
   final isComposeBoxOffered = messageListPage.composeBoxController != null;
 
   final isMessageRead = message.flags.contains(MessageFlag.read);
   final markAsUnreadSupported = store.connection.zulipFeatureLevel! >= 155; // TODO(server-6)
   final showMarkAsUnreadButton = markAsUnreadSupported && isMessageRead;
 
   final optionButtons = [
-    ReactionButtons(message: message, pageContext: context),
-    StarButton(message: message, pageContext: context),
+    ReactionButtons(message: message, pageContext: pageRootContext),
```
gnprice, line 521: nit: let's just call this "pageContext" — it's shorter and matches the name we're already using
```suggestion
    ReactionButtons(message: message, pageContext: pageContext),
```

lib/widgets/action_sheet.dart
```diff
@@ -372,37 +391,145 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+      showErrorDialog(context: pageContext, title: title, message: errorMessage);
+    }
+  }
+}
+
 /// Show a sheet of actions you can take on a message in the message list.
 ///
-/// Must have a [MessageListPage] ancestor.
+/// Must have a [MessageListPage] ancestor and a [PageRoot] ancestor.
 void showMessageActionSheet({required BuildContext context, required Message message}) {
```
gnprice, line 504: I think specifically this implementation needs a PageRoot ancestor that *has* a MessageListPage ancestor.

But given that MessageListPage inserts a PageRoot, that just means a MessageListPage ancestor, right?

lib/widgets/message_list.dart
```diff
@@ -277,7 +277,7 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
             narrow: ChannelNarrow(streamId)))));
     }
 
-    return Scaffold(
+    return PageRoot(child: Scaffold(
```
gnprice, line 280: This probably deserves a brief comment saying why this needs a PageRoot here. (It's so that the PageRoot context can be used to find the MessageListPageState.)

test/widgets/action_sheet_test.dart
```diff
@@ -160,9 +160,13 @@ void main() {
         ]);
       }
 
-      await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
-        child: const HomePage()));
+      // ZulipApp instead of TestZulipApp so that we can push a route to the nav,
+      // exercising HomePage.buildRoute which (via AccountPageRouteMixin)
+      // makes a PageRoot, which the action sheet needs.
+      await tester.pumpWidget(ZulipApp());
```
gnprice, line 166: Hmm, do we push a route to the nav in these tests? I don't see where.

lib/widgets/action_sheet.dart
```diff
@@ -372,37 +391,145 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
 
+class ResolveUnresolveButton extends ActionSheetMenuItemButton {
+  ResolveUnresolveButton({
+    super.key,
+    required this.topic,
+    required this.someMessageIdInTopic,
+    required super.pageContext,
+  }) : _actionIsResolve = !topic.isResolved;
+
+  final bool _actionIsResolve;
```
gnprice, line 402: nit: order this private field after the public fields that appear in the constructor

lib/widgets/action_sheet.dart
```diff
@@ -372,37 +391,145 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+    final message = store.messages[someMessageIdInTopic] as StreamMessage?;
+
+    if (message != null && !message.topic.isSameAs(topic)) {
+      // TODO also interrupt on a channel move, with a specific message?
+
+      // The message's topic doesn't match what it was when we opened the
+      // action sheet. So either the resolve/unresolve action was already done,
+      // the topic was renamed, or at least one message in it including
+      // someMessageIdInTopic was moved. In any case, tell the user and don't
+      // make a request.
```
gnprice, line 439: What happens if we leave out this whole conditional, and this situation arises?

Whatever happens then is the same thing that will happen in this version if the user is a bit quicker to hit the resolve/unresolve button (so that they do so before the event reaches the client to say the message was moved), or if the other move happens a bit later (ditto). So we'll need to be sure something reasonable happens in that case.

lib/widgets/action_sheet.dart
```diff
@@ -372,37 +391,145 @@ class UserTopicUpdateButton extends ActionSheetMenuItemButton {
...
+        propagateMode: PropagateMode.changeAll,
+        sendNotificationToOldThread: false,
+        sendNotificationToNewThread: true,
+      );
+    } catch (e) {
+      if (!pageContext.mounted) return;
+
+      String? errorMessage;
+
+      switch (e) {
```
gnprice, line 485: nit:
```suggestion
      String? errorMessage;
      switch (e) {
```

since this `switch` is the block that sets this variable

gnprice (commented):
Thanks both! Comments below.

About to switch to another task, so posting this without having yet read the tests or the small last commit.

test/widgets/action_sheet_test.dart
chrisbobbe: Ah indeed; yeah it turned out we don't need to push a route in order to show the inbox, since it's on HomePage, which automatically appears. We still need `ZulipApp` instead of `TestZulipApp`, though, to get the `PageRoot` included, since that happens via `HomePage.buildRoute`.

lib/widgets/action_sheet.dart
chrisbobbe: We proceed with the request, and then it's up to the server how to handle it.

On CZO, when I let a second resolve-topic request go through, I get an API error, which we show this way. I think it's reasonable:

image

I produced this by commenting out this whole conditional, also commenting out the line in `ActionSheetMenuItemButton._handlePressed` that immediately closes the action sheet, and tapping "Resolve topic" twice.

Here's an example with unresolve-topic:

image

-----

No such luck in the case where the topic was renamed, though. My steps were, still with this conditional commented out:

- (With the topic unresolved)
- Open the topic action sheet, from a recipient header in a topic narrow
- Rename the topic from the web app ("Move topic" in the UI)
- Wait for event to be processed in the app (topic-narrow msglist renarrows)
- Tap "Resolve topic" in the action sheet

Actual: No API error, and all the messages in the post-rename topic get moved to the original topic plus the "✔ " prefix.

Better, I guess, if the server gave an API error, but I don't see how it could do that with the current request params. Maybe if it let clients say what they assume the topic is at the time of applying the update, so it could give an error if it's not a match?

chrisbobbe:
Thanks for the review! Revision pushed, and replied above.

test/widgets/action_sheet_test.dart
gnprice: I think it'd make sense to have TestZulipApp provide a PageRoot too. It's setting `home`, which is basically a simplified version of the job that `onGenerateInitialRoutes` does.

gnprice:
Thanks! The revision looks good modulo the two open subthreads above, and the nit https://github.com/zulip/zulip-flutter/pull/1301#discussion_r1936171983 . (And I haven't yet read further than before.)

lib/widgets/action_sheet.dart
chrisbobbe: > Does the web app have logic like this for trying to catch the race on the client side? I guess that'd influence how to interpret the observation that we don't seem to see the race in practice.

No I don't think so. There's a function `message_edit.toggle_resolve_topic`, which takes an `old_topic_name` and makes the request with the toggled version of that:

<details>

```js
export function toggle_resolve_topic(
    message_id: number,
    old_topic_name: string,
    report_errors_in_global_banner: boolean,
    $row?: JQuery,
): void {
    let new_topic_name;
    const topic_is_resolved = resolved_topic.is_resolved(old_topic_name);
    if (topic_is_resolved) {
        new_topic_name = resolved_topic.unresolve_name(old_topic_name);
    } else {
        new_topic_name = resolved_topic.resolve_name(old_topic_name);
    }

    if (
        !topic_is_resolved &&
        onboarding_steps.ONE_TIME_NOTICES_TO_DISPLAY.has("intro_resolve_topic")
    ) {
        show_intro_resolve_topic_modal(old_topic_name, () => {
            do_toggle_resolve_topic(
                message_id,
                new_topic_name,
                topic_is_resolved,
                report_errors_in_global_banner,
                $row,
            );
        });
        onboarding_steps.post_onboarding_step_as_read("intro_resolve_topic");
        return;
    }

    do_toggle_resolve_topic(
        message_id,
        new_topic_name,
        topic_is_resolved,
        report_errors_in_global_banner,
        $row,
    );
}

function do_toggle_resolve_topic(
    message_id: number,
    new_topic_name: string,
    topic_is_resolved: boolean,
    report_errors_in_global_banner: boolean,
    $row?: JQuery,
): void {
    if ($row) {
        show_toggle_resolve_topic_spinner($row);
    }

    const request = {
        propagate_mode: "change_all",
        topic: new_topic_name,
        send_notification_to_old_thread: false,
        send_notification_to_new_thread: true,
    };

    void channel.patch({
        url: "/json/messages/" + message_id,
        data: request,
        success() {
            if ($row) {
                const $spinner = $row.find(".toggle_resolve_topic_spinner");
                loading.destroy_indicator($spinner);
            }
        },
        error(xhr) {
            if ($row) {
                const $spinner = $row.find(".toggle_resolve_topic_spinner");
                loading.destroy_indicator($spinner);
            }

            if (xhr.responseJSON) {
                const {code} = z.object({code: z.string()}).parse(xhr.responseJSON);
                if (code === "MOVE_MESSAGES_TIME_LIMIT_EXCEEDED") {
                    handle_resolve_topic_failure_due_to_time_limit(topic_is_resolved);
                    return;
                }

                if (report_errors_in_global_banner) {
                    const {msg} = z.object({msg: z.string()}).parse(xhr.responseJSON);
                    ui_report.generic_embed_error(msg, 3500);
                }
            }
        },
    });
}
```

</details>

The `old_topic_name` is always taken straight from a UI element (specifically an element's `data-topic-name` attribute). In one case, there's a round-trip to fetch a message ID when there isn't one on hand (what we call `someMessageIdInTopic`).

test/widgets/action_sheet_test.dart
chrisbobbe: Sure, seems reasonable.

chrisbobbe:
Thanks! Revision pushed.

lib/widgets/action_sheet.dart
gnprice: OK. What happens if (on web) a user opens the menu for a topic in the left sidebar, then lets that sit for a while, and then comes back and hits "Resolve topic"? Does that use the data the client had when they opened the menu, or the data the client has when they hit the button?

gnprice:
These changes all look good modulo the remaining open subthread above:
https://github.com/zulip/zulip-flutter/pull/1301#discussion_r1936363685

I've also read the tests, and they look good except I skipped the bottom half (starting at `setupTopicChangeAfterSheetOpened`) which are about that question. It looks like the small last commit is also downstream of that.

lib/widgets/action_sheet.dart
chrisbobbe: 1. A channel ID and topic are taken from client data when the popover menu is opened:
  
  web/src/topic_popover.ts
  ```js
              onMount(instance) {
                  const $popper = $(instance.popper);
                  const {stream_id, topic_name, url} = get_conversation(instance);
  ```

2. When the resolve/unresolve button in the popover is hit, the channel ID and topic from step 1 are used to fetch a message ID for a message (the newest message) in the conversation. (The fetch is done by `message_edit.with_first_message_id`.)

3. The edit-messages request is made, with the message ID from step 2 and the result of toggling the topic from step 1.

One important observation though: when the message list renarrows to follow a resolve-topic, bulk-message-move, or whole-topic-rename, the popover menu closes, which thwarts most cases where a stale topic might actually go into the resolve/unresolve request.

chrisbobbe: > […] which should thwart most cases where a stale topic might actually go into the resolve/unresolve request.

Hmm, I guess then the answer to your question above:

> Does the web app have logic like this for trying to catch the race on the client side? I guess that'd influence how to interpret the observation that we don't seem to see the race in practice.

is yes, the web app does have such logic :)

gnprice: > the channel ID and topic from step 1 are used to fetch a message ID for a message (the newest message) in the conversation. (The fetch is done by `message_edit.with_first_message_id`.)

Huh interesting. From that function's doc-ish comment:
```js
    // The API endpoint for editing messages to change their
    // content, topic, or stream requires a message ID.
    //
    // Because we don't have full data in the browser client, it's
    // possible that we might display a topic in the left sidebar
    // (and thus expose the UI for moving its topic to another
    // stream) without having a message ID that is definitely
    // within the topic.  (The comments in stream_topic_history.ts
    // discuss the tricky issues around message deletion that are
    // involved here).
    //
    // To ensure this option works reliably at a small latency
    // cost for a rare operation, we just ask the server for the
    // latest message ID in the topic.
```

The mention of deletion suggests that there's something tracking only the latest one message ID for each topic, so that if that message gets deleted the client maybe doesn't know what other message IDs might be there.

I think we don't have any cases of that problem in this PR, conveniently, so no need for that extra fetch.

---

> One important observation though: when the message list renarrows to follow a resolve-topic, bulk-message-move, or whole-topic-rename, the popover menu closes, which should thwart most cases where a stale topic might actually go into the resolve/unresolve request.

Hmm, I see. What if the popover was just opened from the left sidebar, though? The user might not be narrowed to that conversation in the first place.

chrisbobbe: Ah indeed; yeah, if you're viewing a topic narrow you can open the popover for a different topic, and if you're viewing a channel narrow you can open the popover for one of the channel's topics.

In this case, if all the messages in the topic get their topics changed after the popover is opened, then the popover will look odd for a bit because it won't have a corresponding topic in the list (notice "Chris asdfjkl" vs "Chris asdfjkl;"):

image

Then if you click the resolve/unresolve button, it'll silently do nothing, with an assertion error in the console, because the single-message-fetch didn't produce a message ID (the old topic now being empty). From the `assert` here:

```js
                $popper.one("click", ".sidebar-popover-toggle-resolved", () => {
                    message_edit.with_first_message_id(stream_id, topic_name, (message_id) => {
                        assert(message_id !== undefined);
                        message_edit.toggle_resolve_topic(message_id, topic_name, true);
                    });

                    popover_menus.hide_current_popover_if_visible(instance);
                });
```

chrisbobbe:
In the office just now, we decided to skip the client-side race checks for if the topic was changed after the action sheet was opened. Will send that revision soon :)

chrisbobbe:
OK, pushed that revision!

I've left in the PageRoot commit because it's still theoretically helpful, but now (with those client-side checks removed) it makes a difference more rarely; I can say more about this if needed. 🙂

gnprice:
Thanks! Looks great — merging.

PR #1306, opened 2025-01-24
Mark as resolved/unresolved prep 2

chrisbobbe:
Here are the commits from #1301 that are just about refactoring the action-sheet tests. I've revised them for Zixuan's review https://github.com/zulip/zulip-flutter/pull/1301#pullrequestreview-2573663606 ; in particular by just removing an `assert(() {}())` wrapper (which isn't needed because it's only running in tests).

### DIFF
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -110,11 +110,10 @@ void main() {
   group('topic action sheet', () {
     final someChannel = eg.stream();
     const someTopic = 'my topic';
+    final someMessage = eg.streamMessage(
+      stream: someChannel, topic: someTopic, sender: eg.otherUser);
 
     group('showTopicActionSheet', () {
-      final message = eg.streamMessage(
-        stream: someChannel, topic: someTopic, sender: eg.otherUser);
-
       Future<void> prepare({
         ZulipStream? channel,
         String topic = someTopic,
@@ -151,7 +150,7 @@ void main() {
           channels: [eg.unreadChannelMsgs(
             streamId: someChannel.streamId,
             topic: someTopic,
-            unreadMessageIds: [message.id],
+            unreadMessageIds: [someMessage.id],
           )]));
         await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
           child: const HomePage()));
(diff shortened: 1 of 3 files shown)


test/widgets/action_sheet_test.dart
```diff
@@ -106,27 +108,58 @@ void main() {
...
 
-    testWidgets('show from inbox', (tester) async {
-      await prepare();
+    Future<void> showFromInbox(WidgetTester tester, {
+      String topic = someTopic,
+    }) async {
+      final channelIdsWithUnreads = store.unreads.streams.keys;
+      final topicWithUnreads = channelIdsWithUnreads.firstWhereOrNull((streamId) =>
+        store.unreads.countInTopicNarrow(streamId, TopicName(topic)) > 0);
+      if (topicWithUnreads == null) {
```
gnprice, line 153: The name `topicWithUnreads` doesn't seem to match what this does — it ends up taking the value of a channel ID.

Perhaps use `any` instead of `firstWhereOrNull`?

test/widgets/action_sheet_test.dart
```diff
@@ -135,31 +168,42 @@ void main() {
...
-      check(find.text('Follow topic')).findsOne();
-    });
+    }
+
+    Future<void> showFromRecipientHeader(WidgetTester tester, {
+      String topic = someTopic,
+      StreamMessage? message,
+    }) async {
+      final effectiveMessage = message ?? someMessage;
+      assert(effectiveMessage.topic.apiName == topic);
```
gnprice, line 203: Can this function skip taking `topic` as a parameter, and just get it from the message?

test/widgets/action_sheet_test.dart
```diff
@@ -135,31 +168,42 @@ void main() {
...
       await tester.pump(const Duration(milliseconds: 250));
-      check(find.byType(BottomSheet)).findsOne();
-      check(find.text('Follow topic')).findsOne();
-    });
+    }
+
+    Future<void> showFromAppBar(WidgetTester tester, {
+      ZulipStream? channel,
+      String topic = someTopic,
+      Iterable<StreamMessage>? messages,
```
gnprice, line 176:
```suggestion
      List<StreamMessage>? messages,
```

Then we can skip the `toList` below. The callers are all passing lists anyway.

gnprice (commented):
Thanks! A few small comments below — otherwise these test changes all look good.

Then I think you meant to leave the last 3 commit out of this PR, right? They'll be for #1301.

test/widgets/action_sheet_test.dart
chrisbobbe: Ah indeed, and `hasTopicWithUnreads` for the variable name.

chrisbobbe:
> Then I think you meant to leave the last 3 commit out of this PR, right? They'll be for #1301.

Ah indeed!

chrisbobbe:
Thanks for the review, revision pushed!

gnprice:
Thanks! Looks good; merging.

Issue #649, opened 2024-05-08
Update unreads data on fetching messages


gnprice:
When a new message comes in, we update the `Unreads` data structure, `store.unreads`. But when we go and fetch a bunch of messages for a message list, we don't.

Mostly this works out, because the initial snapshot has an unreads digest that already covers the most recent unreads, up to a large number of them. But if you go look at some very old history where you have unreads, we'll end up with messages we know are unread but `store.unreads` doesn't reflect it. This is odd and would be good to fix.

This will most naturally come after a central message store, #648 / #455.

Milestoned post-launch because we don't currently do this in zulip-mobile.

Issue #81, opened 2023-04-21
Mark messages as read on scroll, configurably


gnprice:
This Zulip feature: https://zulip.com/help/marking-messages-as-read#configure-whether-messages-are-automatically-marked-as-read

That means:
* When scrolling through messages that are unread, by default mark them as read with a request to the server.
* Control that behavior with a local setting "Mark messages as read on scroll", with three values: https://zulip.com/help/marking-messages-as-read#configure-whether-messages-are-automatically-marked-as-read
* Unlike most everything else in the Zulip data model, apply "latency compensation" when marking messages as read on scroll, meaning eagerly apply the change to the local UI rather than wait for the server to acknowledge it with an event.
  * Specifically, eagerly fade out the unread-message markers in the message list.

That local setting should go in the app's database on the device, from #13.

This should come after:
* #79
* #80

gnprice:
> Specifically, eagerly fade out the unread-message markers in the message list.

We might go further and eagerly update the model itself.
* The zulip-mobile RN app doesn't do this; the consequence is that it looks kind of glitchy as the "N unreads" banner at the top of the message list doesn't update or disappear in sync with the unread markers.
* I'm not actually sure what the web app does.
* The downside is that if we eagerly update the model, then it's no longer part of a coherent snapshot together with the rest of our model — we'd be giving up some of the consistency provided by the Zulip event system. 

There are a couple of approaches we could take to that consistency question, but we should think through the consequences. So, leaving that out of scope for this issue.

chrisbobbe:
We've had a request for this as beta feedback: https://chat.zulip.org/#narrow/channel/48-mobile/topic/mobile.20beta.20feedback/near/2000430

chrisbobbe:
And [some more requests](https://chat.zulip.org/#narrow/channel/48-mobile/topic/marking.20read/near/2018451):

> here's another vote for prioritizing "Mark messages as read on scroll" support in the Flutter app.

> […] FWIW this is the main reason why I have not switched to the beta app. I'm subscribed to all channels here, and the fact that I have to continually mark messages as read manually makes the beta app very tedious for me to use -- I am manually marking hundreds if not thousands of messages as read every day.

PR #1311, opened 2025-01-27
Handle moved messages for unreads

PIG208:
Work toward #901

I prioritized updating unreads data for moved messages because it is more visible after supporting the resolve/unresolve topic actions. Updating recent senders data will be a follow-up to this.


chrisbobbe:
Actually @gnprice, I'd appreciate any thoughts you have from a skim of the first three commits:

model [nfc]: Add PropagateMode.fromRawString
api [nfc]: Extract and parse UpdateMessageMoveData from UpdateMessageEvent
api: Parse UpdateMessageMoveData with inter-related constraints

before I start my review; I think you designed/implemented similar logic in zulip-mobile.

gnprice:
Didn't catch the GitHub notification for this earlier, sorry — thanks for the ping in chat today.

The general strategy in those commits looks good. The high-level comments I have are:
* Here in the API parsing code, when the response is malformed, instead of returning null, let's just throw — for example if something shouldn't be null, then just the `!` operator is enough.

  That's what already happens for all kinds of other ways it can be malformed, like if `json['stream_id']` is a string value (instead of a number or absent/null). Such exceptions get turned into MalformedServerResponseException by the generic code in ApiConnection.send, and in the case of events those get handled appropriately by the long-poll loop. (They cause a reload of the store, because if there's an event we weren't able to handle then we may no longer have accurate state.)

* Let's have an invariant that we don't wind up with an `UpdateMessageMoveData` that doesn't represent a move: i.e. where the old and new channel ID are equal and the old and new topic are equal. In our code that updates data structures we might rely on that assumption (for example so that we're not mutating the same data structure twice concurrently as both the source and target of a move).

  That can be enforced by both an assert on the constructor, and the try-parse method either throwing or returning null before getting to the constructor in that case. (Compare the [zulip-mobile logic](https://github.com/zulip/zulip-mobile/blob/e352f563e/src/api/misc.js#L40-L42) quoted below.)

* This logic about in what circumstances the various "orig" and "new" fields might be null, or are required to be present, feels pretty unsatisfying. Even though it's meant to accurately reflect guarantees that are in the API docs (https://zulip.com/api/get-events#update_message), and probably it does, it doesn't feel *solid*.

  In particular, facts like this one feel like quirks:
  > `orig_subject` (aka origTopic) is documented to be present when either the stream or topic changed.

  and are asymmetric between the topic and the channel fields. Indeed this sort of fact has changed in the past (origStreamId aka `stream_id` [became more widely present](https://github.com/zulip/zulip-mobile/blob/e352f563e/src/api/misc.js#L30-L33) in Zulip Server 5), and just last year we had to edit the logic because the docs were wrong (020bfcb9bd6dbd94e45a0cf8274e1a46891a2793) about a fact of this kind which we had relied on.

  This logic [in the zulip-mobile version](https://github.com/zulip/zulip-mobile/blob/e352f563e/src/api/misc.js#L40-L42) feels more robust, because it's based on natural facts about the semantics the event is about:

  ```js
  if (new_topic === orig_topic && new_stream_id === orig_stream_id) {
    // Stream and topic didn't change.
    return null;
  ```

  Perhaps better yet, though: we could just look to see if `propagate_mode` is present. (The zulip-mobile version didn't consider `propagate_mode` as part of this logic.) That's something that is necessary if there was a move, and clearly meaningless if there wasn't — so it seems like a very crisp and reliable indicator of whether there was a move. (And the API docs agree that it, too, is present just if there was a move.)

gnprice:
(So probably the right next steps are: @PIG208 go ahead and act on the feedback above, and then after that revision @chrisbobbe will do maintainer review.)

PIG208:

chrisbobbe (commented):
Thanks, I'm excited for more correctness in the Unreads model! Small comments below.

test/api/model/events_test.dart
```diff
@@ -99,23 +99,32 @@ void main() {
       'message_ids': [message.id],
       'flags': <String>[],
       'edit_timestamp': 1718741351,
-      'stream_id': eg.stream().streamId,
     };
 
-    test('stream_id -> origStreamId', () {
+    test('stream_id -> origStreamId, subject = orig_subject', () {
```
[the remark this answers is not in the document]
PIG208: It's meant to represent that the topic does not change in this move. Changing it to a short phrase should help make it clear.

test/model/message_list_test.dart
```diff
@@ -777,11 +777,14 @@ void main() {
...
+        final thirdStream = eg.stream();
         await prepareNarrow(narrow, initialMessages);
+        await store.addStream(thirdStream);
+        await store.addSubscription(eg.subscription(thirdStream));
 
         await store.handleEvent(eg.updateMessageEventMoveFrom(
           origMessages: otherChannelMovedMessages,
-          newStreamId: otherStream.streamId,
+          newStreamId: thirdStream.streamId,
         ));
```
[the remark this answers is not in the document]
PIG208: It would be helpful before we pull out `UpdateMessageMoveData`, which actually helped me discover this issue in the first place. Having the assertions and check in one place (`UpdateMessageMoveData`) appears to be easier to maintain.

test/api/model/events_test.dart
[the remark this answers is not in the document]
gnprice: This new check seems like it'd be clearest as a new separate test case. I'm not sure how it relates to the existing test, or really what it's meant to check.

lib/api/model/events.dart
```diff
@@ -730,50 +730,33 @@ class UpdateMessageMoveData {
...
     final propagateMode = propagateModeString == null ? null
       : PropagateMode.fromRawString(propagateModeString);
-    final origTopic = json['orig_subject'] == null ? null
-      : TopicName.fromJson(json['orig_subject'] as String);
-    final newTopic = json['subject'] == null ? null
-      : TopicName.fromJson(json['subject'] as String);
 
-    if (origTopic == null) {
+    if (propagateMode == null) {
       // There was no move.
```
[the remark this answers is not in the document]
gnprice: > That would be easy to add back in, right, in the `if (propagateMode == null)`? Just that we'd throw if any of the four stream / topic fields are present.

When there isn't a move, some of those four fields may still be present and it's complicated what exact subset is expected. (See "feel like quirks" and "changed in the past" in this comment: https://github.com/zulip/zulip-flutter/pull/1311#issuecomment-2667327271) So such a check would have to be more complicated than that.

lib/model/unreads.dart
```diff
@@ -296,13 +296,38 @@ class Unreads extends ChangeNotifier {
         madeAnyUpdate |= mentions.add(messageId);
     }
 
-    // TODO(#901) handle moved messages
+    madeAnyUpdate |= _handleMessageMove(event);
```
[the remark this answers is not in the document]
PIG208: We will get there soon! #901 also requires updates to recent senders data model.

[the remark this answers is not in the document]
chrisbobbe: Oh oops! Indeed!

lib/api/model/events.dart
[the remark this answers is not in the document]
PIG208: We can have a check like:

```dart
// pseudo representation of the types of the fields
TopicName? origTopic = ...;
int? origStreamId = ...;
final newTopic ??= origTopic;
final newStreamId ??= origStreamId;

if (propagateMode == null) {
  if (origTopic != newTopic || origStreamId != newStreamId) {
    throw FormatException('move but no propagateMode'); // or "maybe move but no propagateMode"
  }
  // There was no move.
  return null;
}
```

This still throws if we have `origTopic == null` or `origStreamId == null` with a non-null `newTopic` or `newStreamId`, but that's malformed data anyway.

It is the negation of `origStreamId == newStreamId && origTopic == newTopic`, which is used later in this method.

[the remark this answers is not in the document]
gnprice: Yeah, that could be fine if it doesn't involve making the other logic any more complicated to support it.

chrisbobbe:
It turns out there's a move-message edge case that we'll need to handle, either here or as a followup issue.

It's this TODO in `Unreads.handleUpdateMessageEvent`:

```dart
    // We assume this event can't signal a change in a message's 'read' flag.
    // TODO can it actually though, when it's about messages being moved into an
    //   unsubscribed stream?
    //   https://chat.zulip.org/#narrow/stream/378-api-design/topic/mark-as-read.20events.20with.20message.20moves.3F/near/1639957
```

The conclusion from an API-design [discussion](https://chat.zulip.org/#narrow/channel/378-api-design/topic/mark-as-read.20events.20with.20message.20moves.3F/near/2102257) is that clients need to check if the messages were moved into a channel that is unsubscribed, and if so, drop them from unreads, since we won't be getting a separate event about that. (This is currently undocumented on `update_message` but should be soon.)

chrisbobbe (commented):
Thanks! Small comments below, and I'll follow up on CZO where you [pointed out](https://chat.zulip.org/#narrow/channel/378-api-design/topic/mark-as-read.20events.20with.20message.20moves.3F/near/2103924) that the move-messages event isn't arriving for a move to an unsubscribed channel, in your testing.

test/model/unreads_test.dart
```diff
@@ -469,6 +469,189 @@ void main() {
         }
       }
     });
+
+    group('moves', () {
```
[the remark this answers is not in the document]
PIG208: "to another subscribed channel" and "to unsubscribed channel" under the "move unread messages" subgroup do test this scenario. Do you mean testing a move to a different channel and a different topic name?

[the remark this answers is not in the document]
chrisbobbe: > Do you mean testing a move to a different channel and a different topic name?

Oh, oops: yes, I assume it would be helpful to test both cases.

lib/api/model/events.dart
```diff
@@ -747,18 +803,20 @@ class UpdateMessageEvent extends Event {
...
     required this.origContent,
     required this.origRenderedContent,
     required this.content,
     required this.renderedContent,
     required this.isMeMessage,
   });
 
+  static Map<String, Object?> _readMoveData(Map<Object?, Object?> json, String key) {
+    // Parsing [UpdateMessageMoveData] requires `json`, not the default `json[key]`.
+    assert(json is Map<String, Object?>); // value came through `fromJson` with this type
```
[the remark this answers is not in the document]
PIG208: I noticed that in the generated code, while the return type of `_readMoveData` is already `Map<String, Object?>`, the generator still uses a type cast in response to `tryParseFromJson` accepting `Map<String, Object?>`:

```
      moveData: UpdateMessageMoveData.tryParseFromJson(
          UpdateMessageEvent._readMoveData(json, 'move_data')
              as Map<String, Object?>),
```

So we can actually skip the type cast in the readValue callback. And then we can just have:

```
  static Object? _readMoveData(Map<dynamic, dynamic> json, String key) {
    // Parsing [UpdateMessageMoveData] requires `json`, not the default `json[key]`.
    return json;
  }
```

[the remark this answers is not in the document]
chrisbobbe: I don't yet understand why to prefer `Object?` for a value that we definitely know is a string-keyed Map. It's helpful if the type hints at what kind of value we're dealing with. With `Object?`, it seems harder than it needs to be to understand that the value is `Map`-ified JSON, not something else, like null, or an int, string, boolean, or any other `Object?`.

lib/model/unreads.dart
```diff
@@ -488,20 +525,28 @@ class Unreads extends ChangeNotifier {
...
     final messageIds = topics[topic];
-    if (messageIds == null) return;
+    if (messageIds == null) return null;
 
     // ([QueueList] doesn't have a `removeAll`)
-    messageIds.removeWhere((id) => incomingMessageIds.contains(id));
+    final removedMessageIds = QueueList<int>();
+    messageIds.removeWhere((id) {
+      if (incomingMessageIds.contains(id)) {
+        removedMessageIds.add(id);
```
gnprice, line 538: In the case where the whole conversation got moved — a very common case — this will end up unnecessarily making a new copy of the list. Let's adjust so that in that case it can just return the existing list, without copying it.

gnprice: Even more common: this code path also occurs every time the user marks some messages as read. That case doesn't need a list of the removed message IDs at all, so let's have it avoid making such a copy even when it's only some of the messages. (Which is common when reading on web, or reading in this app in the future after #81.)

That probably means these two cases should be separate methods, since they basically want different return types. It's OK that they'll effectively duplicate each other's first and last few lines.

gnprice (commented):
Thanks @PIG208 for building this, and @chrisbobbe for the previous reviews!

Comments below. For this round I've read all but the tests of the last/main commit.

lib/model/unreads.dart
PIG208: I think shortcuting to removing all unreads for the original topic when propagate mode is 'change_all' is subject to the message fetch race, which will cause some unreads to be incorrectly removed.

gnprice: Yeah, to clarify, I don't mean to say we should start interpreting `propagateMode` here. (This model should be responding to fetches, and it's a bug that it doesn't yet: #649. So the fetch/event race will apply to it in the future.) I'm just referring to the case where all the message IDs we have for the conversation turn out to be in the list of message IDs to move.

It may be cleanest to replace `removeWhere` with an explicit loop. If you click through to its definition, you'll see an example of what that loop can look like. (And we don't need complications like `ConcurrentModificationError`, because unlike the general `removeWhere` this code isn't calling out to arbitrary other code.)

test/api/model/events_checks.dart
```diff
@@ -48,18 +48,23 @@ extension UpdateMessageEventChecks on Subject<UpdateMessageEvent> {
...
+  Subject<UpdateMessageMoveData?> get moveData => has((e) => e.moveData, 'moveData');
   Subject<String?> get origContent => has((e) => e.origContent, 'origContent');
   Subject<String?> get origRenderedContent => has((e) => e.origRenderedContent, 'origRenderedContent');
   Subject<String?> get content => has((e) => e.content, 'content');
   Subject<String?> get renderedContent => has((e) => e.renderedContent, 'renderedContent');
   Subject<bool?> get isMeMessage => has((e) => e.isMeMessage, 'isMeMessage');
 }
 
+
+extension UpdateMessageMoveDataChecks on Subject<UpdateMessageMoveData> {
```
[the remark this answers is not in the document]
PIG208: Ordered `UpdateMessageMoveData` after `UpdateMessageEvent` instead (similar to how `UpdateMessageFlagsMessageDetail` comes after `UpdateMessageFlagsRemoveEvent`).

PIG208:
Thanks for the review! This should be ready again. I made a new revision of tests in the last commit.

PIG208:
I'm planning to update the test cases for the main implementation again, after having a call with Greg.

PIG208:
This should be ready now.

lib/model/unreads.dart
```diff
@@ -504,6 +542,53 @@ class Unreads extends ChangeNotifier {
...
+    if (topics == null) return null;
+    final messageIds = topics[topic];
+    if (messageIds == null) return null;
+
+    final retainedMessageIds = <int>[];
+    for (final id in messageIds) {
+      if (!incomingMessageIds.contains(id)) {
+        retainedMessageIds.add(id);
+      }
+    }
```
gnprice, line 563: This is just `messageIds.where(…).toList()`, right?

lib/model/unreads.dart
```diff
@@ -504,6 +542,53 @@ class Unreads extends ChangeNotifier {
...
+
+    if (retainedMessageIds.isEmpty) {
+      // This is an optimization for the case when all messages in the
+      // conversation are removed, which avoids making a copy of `messageId`
+      // unnecessarily.
```
gnprice, line 568:
```suggestion
      // conversation are removed, which avoids making a copy of `messageIds`
      // unnecessarily.
```

right?

test/model/unreads_test.dart
```diff
@@ -469,6 +469,145 @@ void main() {
...
+      final origChannel = eg.stream();
+      const origTopic = 'origTopic';
+      const newTopic = 'newTopic';
+
+      final readMessages = List<StreamMessage>.generate(10,
+        (_) => eg.streamMessage(
+          stream: origChannel, topic: origTopic, flags: [MessageFlag.read]));
+      final unreadMessages = List<StreamMessage>.generate(10,
+        (_) => eg.streamMessage(stream: origChannel, topic: origTopic));
```
gnprice, line 482: nit: make the parallel parts visibly parallel:
```suggestion
        (_) => eg.streamMessage(stream: origChannel, topic: origTopic,
                 flags: [MessageFlag.read]));
      final unreadMessages = List<StreamMessage>.generate(10,
        (_) => eg.streamMessage(stream: origChannel, topic: origTopic));
```

test/model/unreads_test.dart
```diff
@@ -469,6 +469,145 @@ void main() {
...
       }
     });
+
+    group('moves', () {
+      final origChannel = eg.stream();
+      const origTopic = 'origTopic';
+      const newTopic = 'newTopic';
+
+      final readMessages = List<StreamMessage>.generate(10,
+        (_) => eg.streamMessage(
```
gnprice, line 479: Messages can get mutated to handle events. So let's have these get generated fresh by each test that needs them.

(In particular if we refactored so that instead of calling `Unreads.handleUpdateMessageEvent` like other tests in this file, these called `PerAccountStore.handleEvent` like we do in some other areas of test/model/, I believe these would break as written.)

test/model/unreads_test.dart
```diff
@@ -469,6 +469,145 @@ void main() {
...
+      }) {
+        assert(newChannel != null || newTopic != null);
+        return messages.map((message) => StreamMessage.fromJson(
+          message.toJson()
+            ..['stream_id'] = newChannel?.streamId ?? message.streamId
+            ..['subject'] = newTopic ?? message.topic
+        )).toList();
+      }
+
+      test('smoke', () async {
```
gnprice, line 502:
```suggestion
      test('moving a conversation exactly', () async {
```

gnprice: I think this can also check we have the optimization at https://github.com/zulip/zulip-flutter/pull/1311#discussion_r1978634423 , using `identicalTo`.

test/model/unreads_test.dart
```diff
@@ -469,6 +469,145 @@ void main() {
         }
       }
     });
+
+    group('moves', () {
```
gnprice, line 473: Other cases to test here, in order to properly exercise `_popAllInStreamTopic`:
* the moved messages are an (inhabited) proper subset of the existing unreads in the conversation
* the moved messages are a proper superset of the (inhabited) set of existing unreads in the conversation

gnprice (commented):
Thanks for the revision! Comments below. In this round I've now read the whole thing.

PIG208:
Thanks for the review! Reorganized the move data change into 1 commit followed by 3 nfc ones, and addressed the comments on the main commit.

* eeb89b4a2 message test [nfc]: Move tests to a new home
* cf816efc7 api [nfc]: Simplify move data parsing with null-checks
* 40ccd50ed api [nfc]: Extract and parse UpdateMessageMoveData from UpdateMessageEvent
* b9eafeee9 message: Preface message move handling with more comprehensive null-checks

gnprice (commented):
Thanks!

Here's a partial review focused on that stretch of refactoring commits. I decided some of this would be most efficient for me to write instead of trying to describe — so I've just pushed a revision where those 4 commits you mentioned above are now these 5:
95fbd04d5 message: Identify moves by inequality new vs orig, not just non-nullity
4392beb90 message [nfc]: In moves, give newStreamId/newTopic their fallback values sooner
6b75bae94 wip api [nfc]: Extract and parse UpdateMessageMoveData from UpdateMessageEvent
5d2a8bbcb api [nfc]: Simplify move data parsing with null-checks
e79ea1ea3 message test [nfc]: Move tests to a new home

The first two are new commits I just wrote (and one of them has you as a co-author), replacing your commit:
b9eafee message: Preface message move handling with more comprehensive null-checks

The third one I marked as "wip" because I'm not sure it's entirely coherent after I rebased it atop the other two.

The two after that I didn't touch.

I think this revision handles my comments below, except for the last one. So the open items for you are that last small comment, plus please take a look at the "wip" commit  6b75bae94 and see what adjustments it needs.

PIG208:
Thanks for helping with this! Pushed an update to the PR.

lib/api/model/events.dart
```diff
@@ -747,25 +739,95 @@ class UpdateMessageEvent extends Event {
...
+  /// Try to extract [UpdateMessageMoveData] from the JSON object for an
+  /// [UpdateMessageEvent].
+  ///
+  /// Returns `null` if there was no message move.
+  ///
+  /// Throws an error if the data is malformed.
+  // When parsing this, 'stream_id', which is also present when there was only
+  // a content edit, cannot be recovered if this ends up returning `null`.
+  // This may matter if we ever need 'stream_id' when no message move occurred.
+  static UpdateMessageMoveData? tryParseFromJson(Map<String, Object?> json) {
```
gnprice, line 790: nit:

    This is mostly NFC except that we were logging and ignoring malformed
    event data before, and now start to throw errors with the parsing code.

s/with/within/ — i.e., we're now throwing errors from inside the parsing code

lib/api/model/events.dart
```diff
@@ -747,25 +739,95 @@ class UpdateMessageEvent extends Event {
...
+
+    final origTopic = json['orig_subject'] == null ? null
+      : TopicName.fromJson(json['orig_subject'] as String);
+    final newTopicRaw = json['subject'] == null ? null
+      : TopicName.fromJson(json['subject'] as String);
+    final newTopic = newTopicRaw ?? origTopic;
+
+    if (newStreamId == origStreamId && newTopic == origTopic) {
+      if (propagateMode != null) {
+        throw FormatException(
```
gnprice, line 807: nit in commit message:

    (It might also change what particular message gets printed to the
    debug log; but that doesn't count as a behavior change. As a part of
    that, we add an additional check for propagate mode, to succeed the
    block of assertion.)

I see this second sentence as not so closely related to the first one; the first sentence is already true without any caveats related to the `propagateMode` check, since the only effect of that check is a log message.

Instead let's have a short paragraph at the end, after the "This change is useful" paragraph:

    While here, we add a debug log message on unexpected propagateMode,
    to further bring this closer to how it will look in the API parsing
    code.

lib/api/model/events.dart
```diff
@@ -747,25 +739,95 @@ class UpdateMessageEvent extends Event {
...
 
+/// Data structure representing a message move.
+class UpdateMessageMoveData {
+  final int origStreamId;
+  final int newStreamId;
+
+  final PropagateMode propagateMode;
+
+  final TopicName origTopic;
+  final TopicName newTopic;
```
gnprice, line 771: nit: let's follow the existing code in main by putting these in a more logical order, rather than this order (which I realize reflects the API docs). After all, by making newStreamId and newTopic non-null we're already deviating from the server API, and in a more substantive way :slightly_smiling_face: 

The existing code's order of stream, stream, topic, topic, propagateMode seems good.

(Then similarly for the parsing code below, and so on.)

test/model/unreads_test.dart
```diff
@@ -469,6 +469,192 @@ void main() {
...
+        model.handleUpdateMessageEvent(eg.updateMessageEventMoveFrom(
+          origMessages: unreadMessages,
+          newStreamId: newChannel.streamId,
+          newTopicStr: newTopic));
+        checkNotifiedOnce();
+        checkMatchesMessages(copyMessagesWith(unreadMessages,
+          newChannel: newChannel, newTopic: newTopic));
+        final newMessageIds =
+          model.streams[newChannel.streamId]![TopicName(newTopic)]!;
+        check(originalMessageIds).identicalTo(newMessageIds);
```
gnprice, line 522:
```suggestion
        // Check we successfully avoided making a copy of the list.
        check(originalMessageIds).identicalTo(newMessageIds);
```

(otherwise it seems not quite self-explanatory why this check is relevant / why we'd care if it started failing)

test/model/unreads_test.dart
```diff
@@ -469,6 +469,192 @@ void main() {
...
+        checkNotifiedOnce();
+        checkMatchesMessages([
+          ...copyMessagesWith(unreadMessages, newTopic: newTopic),
+        ]);
+        final newMessageIds =
+          model.streams[origChannel.streamId]![TopicName(newTopic)]!;
+        check(originalMessageIds).identicalTo(newMessageIds);
+      });
+
+      test('moving all read and unread messages from a conversation', () async {
```
gnprice, line 586: This test seems redundant with the one above it, "moving some read and all unread messages from a conversation". So we can leave it out.

(The only difference is whether there are 2 or 10 read messages included in the list to move. In both cases there are 10 read messages included in `fillWithMessages` in setup; but read messages have no effect in `fillWithMessages`, so that doesn't matter.) 

I guess in particular this means that for one of the two tests I asked for at https://github.com/zulip/zulip-flutter/pull/1311#discussion_r1988053653 , all that was needed was a rename to make clearer that an existing test covered that scenario :slightly_smiling_face:

test/model/unreads_test.dart
```diff
@@ -469,6 +469,192 @@ void main() {
...
+        final originalMessageIds =
+          model.streams[origChannel.streamId]![TopicName(origTopic)]!;
+
+        model.handleUpdateMessageEvent(eg.updateMessageEventMoveFrom(
+          origMessages: messagesToMove,
+          newTopicStr: newTopic));
+        checkNotifiedOnce();
+        checkMatchesMessages([
+          ...copyMessagesWith(unreadMessages, newTopic: newTopic),
+        ]);
```
gnprice, line 580: nit: no need for a list literal, can just use the `copyMessagesWith` result directly

gnprice (commented):
Thanks! This is quite close now — just small comments below.

test/model/unreads_test.dart
PIG208: Ha, I see. Yeah it occurred to me that the current test names aren't accurate enough about what they are doing, so it was no so clear whether they cover all the cases we care about.

Updated the names to express the relationships between the messages in terms of sets.

PIG208 (commented):

test/model/unreads_test.dart
```diff
@@ -469,6 +469,175 @@ void main() {
...
+          origMessages: messagesToMove,
+          newTopicStr: newTopic));
+        checkNotifiedOnce();
+        checkMatchesMessages([
+          ...copyMessagesWith(messagesToMove, newTopic: newTopic),
+          ...unreadMessages.skip(2),
+        ]);
+      });
+
+      test('moved messages ∩ unread messages ≠ Ø, moved messages ∩ read messages ≠ Ø, moved messages ⊅ unread messages', () async {
```
gnprice, line 553:
```suggestion
      test('moved messages ∩ unread messages ≠ Ø, moved messages ∩ read messages ≠ Ø, moved messages ⊅ unread messages', () async {
```

:slightly_smiling_face: I'm not convinced this way is clearer than using words, but it does work.

gnprice (commented):
Thanks! Looks good; merging.

PR #1315, opened 2025-01-30
emoji: Fix bottom padding of emoji picker

PIG208:
Previously, the body of the bottom sheet was wrapped in `SafeArea`. This pads the bottom unconditionally, shifting the bottom of the list view above the device bottom inset.

This is undesirable, because the area beneath the bottom inset is still visible, and the list view should extend to the bottom regardless of the bottom inset.

By just removing the `SafeArea`, the list view extends to the bottom. However, if the bottom inset is more than 8px, we can't scroll past the last item in the list view.  Essentially, we want the behavior of `SafeArea.minimum` with a `ListView` — the bottom of the list should always be padded by at least 8px, so that we can scroll past the shadow; and if the bottom inset is more than 8px, use that as the padding instead — which is achieved through the use `ListView.padding` and `MediaQuery`.

### DIFF
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -530,7 +530,7 @@ class _EmojiPickerState extends State<EmojiPicker> with PerAccountStoreAwareStat
               style: const TextStyle(fontSize: 20, height: 30 / 20))),
         ])),
       Expanded(child: InsetShadowBox(
-        top: 8, bottom: 8,
+        top: 8,
         color: designVariables.bgContextMenu,
         child: CustomScrollView(
           slivers: [
--- a/lib/widgets/emoji_reaction.dart
+++ b/lib/widgets/emoji_reaction.dart
@@ -417,20 +417,21 @@ void showEmojiPickerSheet({
     // on my iPhone 13 Pro but is marked as "much slower":
     //   https://api.flutter.dev/flutter/dart-ui/Clip.html
     clipBehavior: Clip.antiAlias,
+    // The bottom inset is left for [builder] to handle;
+    // see [EmojiPicker] and its [CustomScrollView] for how we do that.
     useSafeArea: true,
     isScrollControlled: true,
     builder: (BuildContext context) {
-      return SafeArea(
-        child: Padding(
-          // By default, when software keyboard is opened, the ListView
-          // expands behind the software keyboard — resulting in some
-          // list entries being covered by the keyboard. Add explicit
-          // bottom padding the size of the keyboard, which fixes this.
-          padding: EdgeInsets.only(bottom: MediaQuery.viewInsetsOf(context).bottom),
-          // For _EmojiPickerItem, and RealmContentNetworkImage used in ImageEmojiWidget.
-          child: PerAccountStoreWidget(
-            accountId: store.accountId,
-            child: EmojiPicker(pageContext: pageContext, message: message))));
+      return Padding(
+        // By default, when software keyboard is opened, the ListView
+        // expands behind the software keyboard — resulting in some
+        // list entries being covered by the keyboard. Add explicit
+        // bottom padding the size of the keyboard, which fixes this.
+        padding: EdgeInsets.only(bottom: MediaQuery.viewInsetsOf(context).bottom),
+        // For _EmojiPickerItem, and RealmContentNetworkImage used in ImageEmojiWidget.
+        child: PerAccountStoreWidget(
+          accountId: store.accountId,
+          child: EmojiPicker(pageContext: pageContext, message: message)));
     });
 }
 
@@ -531,13 +532,19 @@ class _EmojiPickerState extends State<EmojiPicker> with PerAccountStoreAwareStat
       Expanded(child: InsetShadowBox(
         top: 8, bottom: 8,
         color: designVariables.bgContextMenu,
-        child: ListView.builder(
-          padding: const EdgeInsets.symmetric(vertical: 8),
-          itemCount: _resultsToDisplay.length,
-          itemBuilder: (context, i) => EmojiPickerListEntry(
-            pageContext: widget.pageContext,
-            emoji: _resultsToDisplay[i].candidate,
-            message: widget.message)))),
+        child: CustomScrollView(
+          slivers: [
+            SliverPadding(
+              padding: EdgeInsets.only(top: 8),
+              sliver: SliverSafeArea(
+                minimum: EdgeInsets.only(bottom: 8),
+                sliver: SliverList.builder(
+                  itemCount: _resultsToDisplay.length,
+                  itemBuilder: (context, i) => EmojiPickerListEntry(
+                    pageContext: widget.pageContext,
+                    emoji: _resultsToDisplay[i].candidate,
+                    message: widget.message)))),
+          ]))),
     ]);
   }
 }
--- a/test/flutter_checks.dart
+++ b/test/flutter_checks.dart
@@ -11,6 +11,11 @@ extension PaintChecks on Subject<Paint> {
   Subject<Shader?> get shader => has((x) => x.shader, 'shader');
 }
 
+extension OffsetChecks on Subject<Offset> {
+  Subject<double> get dx => has((x) => x.dx, 'dx');
+  Subject<double> get dy => has((x) => x.dy, 'dy');
+}
+
 extension RectChecks on Subject<Rect> {
   Subject<double> get left => has((d) => d.left, 'left');
   Subject<double> get top => has((d) => d.top, 'top');
--- a/test/widgets/emoji_reaction_test.dart
+++ b/test/widgets/emoji_reaction_test.dart
@@ -351,6 +351,9 @@ void main() {
 
     final searchFieldFinder = find.widgetWithText(TextField, 'Search emoji');
 
+    Finder findInPicker(Finder finder) =>
+      find.descendant(of: find.byType(EmojiPicker), matching: finder);
+
     Condition<Object?> conditionEmojiListEntry({
       required ReactionType emojiType,
       required String emojiCode,
@@ -429,9 +432,7 @@ void main() {
       await setupEmojiPicker(tester, message: message, narrow: TopicNarrow.ofMessage(message));
 
       connection.prepare(json: {});
-      await tester.tap(find.descendant(
-        of: find.byType(BottomSheet),
-        matching: find.text('\u{1f4a4}'))); // 'zzz' emoji
+      await tester.tap(findInPicker(find.text('\u{1f4a4}'))); // 'zzz' emoji
       await tester.pump(Duration.zero);
 
       check(connection.lastRequest).isA<http.Request>()
@@ -453,9 +454,8 @@ void main() {
       connection.prepare(
         delay: const Duration(seconds: 2),
         apiException: eg.apiBadRequest(message: 'Invalid message(s)'));
-      await tester.tap(find.descendant(
-        of: find.byType(BottomSheet),
-        matching: find.text('\u{1f4a4}'))); // 'zzz' emoji
+
+      await tester.tap(findInPicker(find.text('\u{1f4a4}'))); // 'zzz' emoji
       await tester.pump(); // register tap
       await tester.pump(const Duration(seconds: 1)); // emoji picker animates away
       await tester.pump(const Duration(seconds: 1)); // error arrives; error dialog shows
--- a/test/widgets/emoji_reaction_test.dart
+++ b/test/widgets/emoji_reaction_test.dart
@@ -466,6 +466,92 @@ void main() {
 
       debugNetworkImageHttpClientProvider = null;
     });
+
+    group('handle view paddings', () {
+      const screenHeight = 400.0;
+
+      late Rect scrollViewRect;
+      final scrollViewFinder = findInPicker(find.bySubtype<ScrollView>());
+
+      Rect getListEntriesRect(WidgetTester tester) =>
+        tester.getRect(find.byType(EmojiPickerListEntry).first)
+          .expandToInclude(tester.getRect(find.byType(EmojiPickerListEntry).last));
+
+      Future<void> prepare(WidgetTester tester, {
+        required FakeViewPadding viewPadding,
+      }) async {
+        addTearDown(tester.view.reset);
+        tester.view.physicalSize = Size(640, screenHeight);
+        // This makes it easier to convert between device pixels used for
+        // [FakeViewPadding] and logical pixels used in tests.
+        // If needed, there is a clearer way to implement this generally.
+        // See comment: https://github.com/zulip/zulip-flutter/pull/1315/files#r1962703436
+        tester.view.devicePixelRatio = 1.0;
+
+        tester.view.viewPadding = viewPadding;
+        tester.view.padding = viewPadding;
+
+        final message = eg.streamMessage();
+        await setupEmojiPicker(tester,
+          message: message, narrow: TopicNarrow.ofMessage(message));
+
+        scrollViewRect = tester.getRect(scrollViewFinder);
+        // The scroll view should expand all the way to the bottom of the
+        // screen, even if there is device bottom padding.
+        check(scrollViewRect)
+          ..bottom.equals(screenHeight)
+          // There should always be enough entries to overflow the scroll view.
+          ..height.isLessThan(getListEntriesRect(tester).height);
+      }
+
+      testWidgets('no view padding', (tester) async {
+        await prepare(tester, viewPadding: FakeViewPadding.zero);
+
+        // The top edge of the list entries is padded by 8px from the top edge
+        // of the scroll view; the bottom edge is out of view.
+        Rect listEntriesRect = getListEntriesRect(tester);
+        check(scrollViewRect)
+          ..top.equals(listEntriesRect.top - 8)
+          ..bottom.isLessThan(listEntriesRect.bottom);
+
+        // Scroll to the very bottom of the list with a large offset.
+        await tester.drag(scrollViewFinder, Offset(0, -500));
+        await tester.pump();
+        // The top edge of the list entries is out of view;
+        // the bottom is padded by 8px, the minimum padding, from the bottom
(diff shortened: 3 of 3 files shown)


PIG208:
| | before | after |
| - | - | - |
| with keyboard | Screenshot_20250130_172020 | Screenshot_20250130_172056 |

lib/widgets/emoji_reaction.dart
```diff
@@ -416,20 +418,21 @@ void showEmojiPickerSheet({
     // on my iPhone 13 Pro but is marked as "much slower":
     //   https://api.flutter.dev/flutter/dart-ui/Clip.html
     clipBehavior: Clip.antiAlias,
+    // This does remove the bottom inset.
+    // Leave it to the descendent [ListView].
     useSafeArea: true,
```
chrisbobbe, line 423: I don't understand this yet—what do you mean by "remove the bottom inset"?

Here's the [doc](https://api.flutter.dev/flutter/material/showModalBottomSheet.html):

```dart
/// The [useSafeArea] parameter specifies whether the sheet will avoid system
/// intrusions on the top, left, and right. If false, no [SafeArea] is added;
/// and [MediaQuery.removePadding] is applied to the top,
/// so that system intrusions at the top will not be avoided by a [SafeArea]
/// inside the bottom sheet either.
/// Defaults to false.
```

It doesn't say anything about controlling what's done about the bottom inset.

chrisbobbe (commented):
Thanks! Comments below.

I stopped reading the tests when I realized they might not work with an alternative implementation I thought of. (Actually would it be possible to write a set of tests that would work with either/any implementation? That's generally preferable.)

lib/widgets/emoji_reaction.dart
```diff
@@ -530,13 +534,22 @@ class _EmojiPickerState extends State<EmojiPicker> with PerAccountStoreAwareStat
...
-          itemCount: _resultsToDisplay.length,
-          itemBuilder: (context, i) => EmojiPickerListEntry(
-            pageContext: widget.pageContext,
-            emoji: _resultsToDisplay[i].candidate,
-            message: widget.message)))),
+        child: MediaQuery.removePadding(
+          context: context,
+          // The bottom is padded below with `padding` within the [ListView].
+          removeBottom: true,
+          child: ListView.builder(
```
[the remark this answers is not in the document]
PIG208: Indeed! This works for me:

```dart
      Expanded(child: InsetShadowBox(
        top: 8, bottom: 8,
        color: designVariables.bgContextMenu,
        child: CustomScrollView(
          slivers: [
            SliverPadding(
              padding: EdgeInsets.only(top: 8),
              sliver: SliverSafeArea(
                minimum: EdgeInsets.only(bottom: 8),
                sliver: SliverList.builder(
                  itemCount: _resultsToDisplay.length,
                  itemBuilder: (context, i) => EmojiPickerListEntry(
                    pageContext: widget.pageContext,
                    emoji: _resultsToDisplay[i].candidate,
                    message: widget.message)))),
          ]))),
```

PIG208:
Updated the PR, with new screenshots here: https://github.com/zulip/zulip-flutter/pull/1315#issuecomment-2625720854.

The new implementation does require reworking the tests. I have updated them with an approach that should work just fine with the emoji picker implementation on `main`.

chrisbobbe:
Thanks! Reviewing now.

One nit: I think this is the wrong column label in the table of screenshots for the current revision:

> after `max(8px, bottom inset)` bottom padding

because the `max` approach isn't used in this revision. I think just a copy-paste error from the other table of screenshots.

test/widgets/emoji_reaction_test.dart
```diff
@@ -471,6 +469,87 @@ void main() {
...
+
+        scrollViewTop = tester.getTopLeft(scrollViewFinder).dy;
+        scrollViewBottom = tester.getBottomLeft(scrollViewFinder).dy;
+        // The scroll view should expand all the way to the bottom of the
+        // screen, even if there is device bottom padding.
+        check(scrollViewBottom).equals(screenHeight);
+
+        final options = tester.widgetList(find.byType(EmojiPickerListEntry));
+        firstOption = options.first;
+        lastOption = options.last;
```
chrisbobbe, line 511:
```suggestion
        final listEntryFinder = find.byType(EmojiPickerListEntry);
        firstOption = tester.widget(listEntryFinder.first);
        lastOption = tester.widget(listEntryFinder.last);
```

with the goal of leaning on `flutter_test` just a little bit more, in case it gives richer / more helpful feedback when a test fails—I don't know if it actually would in this case though.

Actually, with these chained finders (`foo.first` and `foo.last`), would we still need to store `firstOption` and `lastOption` to pass to a `find.byWidget`, or could we just used the chained finders directly?

test/widgets/emoji_reaction_test.dart
```diff
@@ -471,6 +469,87 @@ void main() {
...
+      late double scrollViewTop;
+      late double scrollViewBottom;
+      final scrollViewFinder = findInPicker(find.bySubtype<ScrollView>());
+
+      late Widget firstOption;
+      late Widget lastOption;
+      double getScrollChildrenTop(WidgetTester tester) =>
+        tester.getTopLeft(find.byWidget(firstOption)).dy;
+      double getScrollChildrenBottom(WidgetTester tester) =>
+        tester.getBottomLeft(find.byWidget(lastOption)).dy;
```
chrisbobbe, line 485: The name, return type, and implementation don't seem to fit with each other; I can say more if helpful.

test/widgets/emoji_reaction_test.dart
```diff
@@ -471,6 +469,87 @@ void main() {
...
+        lastOption = options.last;
+      }
+
+      testWidgets('no view padding', (tester) async {
+        await prepare(tester, viewPadding: FakeViewPadding.zero);
+
+        // The top of the list of children is padded by 8px;
+        // the bottom is out of view.
+        check(scrollViewTop).equals(getScrollChildrenTop(tester) - 8);
+        check(scrollViewBottom).isLessThan(getScrollChildrenBottom(tester));
```
chrisbobbe, line 520: It looks like this assumes there's a very long list of children, but I'm not seeing that clearly in the setup code. Reading `setupEmojiPicker`, called by `prepare`, I think I see two items being set up?

```dart
      store.setServerEmojiData(ServerEmojiData(codeToNames: {
        '1f4a4': ['zzz', 'sleepy'], // (just 'zzz' in real data)
      }));
      await store.handleEvent(RealmEmojiUpdateEvent(id: 1, realmEmoji: {
        '1': eg.realmEmojiItem(emojiCode: '1', emojiName: 'buzzing'),
      }));
```

lib/widgets/emoji_reaction.dart
```diff
@@ -417,20 +417,21 @@ void showEmojiPickerSheet({
     // on my iPhone 13 Pro but is marked as "much slower":
     //   https://api.flutter.dev/flutter/dart-ui/Clip.html
     clipBehavior: Clip.antiAlias,
+    // This removes insets from all sides except bottom.
+    // Leave it to the descendent [CustomScrollView].
```
chrisbobbe, line 421: I don't know what "removes insets" means. Taken literally, it doesn't make sense; the insets are features of the device (which might include hardware) that our app isn't able to remove.

How about:

```dart
    // The bottom inset is left for [builder] to handle;
    // see [EmojiPicker] and its [CustomScrollView] for how we do that.
```

It's also possible that we don't need a comment at all, and the `useSafeArea` dartdoc is sufficient if readers are curious.

chrisbobbe (commented):
Thanks! One design-level comment I noticed: now that the content can scroll all the way to the bottom edge of the screen (which I also think is an improvement), how about removing the `InsetShadowBox ` shadow effect at the bottom? We don't have that effect on similar scrollables, such as the "Combined feed" message list.

The shadow effect makes the viewport edge more graceful when there's not already a clear boundary that coincides with it, but the screen edge is such a boundary. Similarly the top edge of the keyboard when that's open. (Clear definition of the top edge of the keyboard is the platform's job, not ours.)

test/widgets/emoji_reaction_test.dart
PIG208: Yeah, I kind of struggled naming these helpers. I meant to express that this fetches y-coord of the top-most/bottom-most edge of the scrollable content (scroll items/scroll list?). 

Are `double getScrollElementTopEdgeDy` and `double getScrollElementBottomEdgeDy` better?

test/widgets/emoji_reaction_test.dart
PIG208: The rest of them come from `EmojiStore.popularEmojiCandidates`.

The test controls the screen height so that the scroll view can be filled. I think we can add some additional checks in `prepare` to verify that's.

PIG208:
Thanks for the review! Just pushed a new revision addressing the comments, and updated the [screenshots](https://github.com/zulip/zulip-flutter/pull/1315#issuecomment-2625720854).

PIG208:
Hmm, this

```dart
tester.getTopLeft(tester.widget(listEntryFinder.first));
```

is causing a `_TypeError`

```dart
The following _TypeError was thrown running a test:
type 'EmojiPickerListEntry' is not a subtype of type 'Never' in type cast
```

test/widgets/emoji_reaction_test.dart
```diff
@@ -469,6 +469,87 @@ void main() {
...
+        tester.getBottomLeft(listEntryFinder.last).dy;
+
+      Future<void> prepare(WidgetTester tester, {
+        required FakeViewPadding viewPadding,
+      }) async {
+        addTearDown(tester.view.reset);
+        tester.view.physicalSize = Size(640, screenHeight);
+        // This makes it easier to convert between device pixels used for
+        // [FakeViewPadding] and logical pixels used in tests.
+        tester.view.devicePixelRatio = 1.0;
```
chrisbobbe, line 493: This is a way to let callers use less code, but it doesn't really remove the caller's burden to think about the physical/logical pixel distinction. That's because callers are still asked to make a `FakeViewPadding`, and the dartdoc of that says it's in physical pixels, while the rest of the test code is naturally in logical pixels. Callers are asked to assume the pixel ratio is 1, but they're not asked explicitly, in `prepare`'s interface (it has no dartdoc), so they have to look at this line in the implementation, to be sure.

It's helpful that we're not asking callers to do math with `devicePixelRatio`, but I think there's a more satisfying way to do that, and with something reusable. (Helpful because this is likely to come up again.) What if we did this:

- New file, like test/test_window.dart
- New private subclass of `FakeViewPadding`:
  - It implements a method `FakeViewPadding resolve(double devicePixelRatio)` that returns a plain `FakeViewPadding`, scaled from `this` by `devicePixelRatio`
  - Its `left`, `top`, `right`, and `bottom` getters throw errors saying that `resolve` must be called. (Or maybe call the superclass's getter just if the value is zero.) See for example the `operator` getter on `ApiNarrowDm`.
- An extension on `FakeViewPadding`, adding a static method with return type `FakeViewPadding`, named something like `fromLogicalPx`, that creates and returns an instance of the private subclass

Then for these tests, callers would say:

```dart
FakeViewPadding.fromLogicalPx(bottom: 10)
```

and `prepare` would say

```dart
        tester.view.viewPadding = viewPadding.resolve(devicePixelRatio);
```

test/widgets/emoji_reaction_test.dart
```diff
@@ -469,6 +469,87 @@ void main() {
...
+
+      late double scrollViewTopEdgeOffset;
+      late double scrollViewBottomEdgeOffset;
+      final scrollViewFinder = findInPicker(find.bySubtype<ScrollView>());
+
+      final listEntryFinder = find.byType(EmojiPickerListEntry);
+      double getListEntriesTopEdgeOffset(WidgetTester tester) =>
+        tester.getTopLeft(listEntryFinder.first).dy;
+      double getListEntriesBottomEdgeOffset(WidgetTester tester) =>
+        tester.getBottomLeft(listEntryFinder.last).dy;
```
chrisbobbe, line 484: Are the finders restricted to list entries that are at least partially visible in the viewport, and do the tests depend on that? If so, maybe that should be reflected in these names. I agree naming these is hard 😅 https://github.com/zulip/zulip-flutter/pull/1315#discussion_r1955384584 —maybe Greg would have some thoughts, having recently worked with the sticky-header code.

chrisbobbe (commented):
Thanks! Small comments on the tests, which actually I'm curious for Greg's thoughts on: one is a naming question, the other is about adding a helper that might be helpful but maybe not necessary.

test/widgets/emoji_reaction_test.dart
```diff
@@ -350,6 +350,8 @@ void main() {
     }
 
     final searchFieldFinder = find.widgetWithText(TextField, 'Search emoji');
+    Finder findInPicker(Finder finder) =>
```
gnprice, line 353: I like the name of this.

It should have a blank line to separate from the other declaration, because they're not both one-line declarations (nor closely related):
```suggestion
    final searchFieldFinder = find.widgetWithText(TextField, 'Search emoji');

    Finder findInPicker(Finder finder) =>
```

test/flutter_checks.dart
```diff
@@ -169,3 +169,8 @@ extension TableChecks on Subject<Table> {
 extension IconButtonChecks on Subject<IconButton> {
   Subject<bool?> get isSelected => has((x) => x.isSelected, 'isSelected');
 }
+
+extension OffsetChecks on Subject<Offset> {
```
gnprice, line 173: nit: keep the file organized more or less logically, rather than in the order we needed things; so put this near the top because it's from `dart:ui`, i.e. the engine; specifically just above Rect, because it's closely related to Rect but simpler

test/widgets/emoji_reaction_test.dart
```diff
@@ -466,6 +464,87 @@ void main() {
...
+        tester.view.viewPadding = viewPadding;
+        tester.view.padding = viewPadding;
+
+        final message = eg.streamMessage();
+        await setupEmojiPicker(tester,
+          message: message, narrow: TopicNarrow.ofMessage(message));
+
+        scrollViewTopEdgeOffset = tester.getTopLeft(scrollViewFinder).dy;
+        scrollViewBottomEdgeOffset = tester.getBottomLeft(scrollViewFinder).dy;
+        final scrollViewHeight = tester.getSize(scrollViewFinder).height;
```
gnprice, line 499: nit: These facts can all be bundled together as a Rect:
```suggestion
        scrollViewRect = tester.getRect(scrollViewFinder);
```

Then references below to these three variables can instead use `.top`, `.bottom`, `.height` on the Rect.

test/widgets/emoji_reaction_test.dart
```diff
@@ -466,6 +464,87 @@ void main() {
...
+        check(getListEntriesBottomEdgeOffset(tester) - getListEntriesTopEdgeOffset(tester))
+          .isGreaterThan(scrollViewHeight);
+      }
+
+      testWidgets('no view padding', (tester) async {
+        await prepare(tester, viewPadding: FakeViewPadding.zero);
+
+        // The top edge of the list entries is padded by 8px from the top edge
+        // of the scroll view; the bottom edge is out of view.
+        check(scrollViewTopEdgeOffset).equals(getListEntriesTopEdgeOffset(tester) - 8);
```
gnprice, line 513: nit: line too long, key information past 80 columns
```suggestion
        check(scrollViewTopEdgeOffset)
          .equals(getListEntriesTopEdgeOffset(tester) - 8);
```

test/widgets/emoji_reaction_test.dart
gnprice: These will find only list items that have actually been built and are present in the element tree.

The list will build items until either it runs out or it fills the viewport, plus 250px in either direction — the default value of [ScrollView.cacheExtent]. It looks like these tests never care about how far beyond the viewport the list items extend, as long as they extend beyond it at all — so the fact that cacheExtent is positive is enough to make things work.

test/widgets/emoji_reaction_test.dart
gnprice: This sounds useful but I think it's OK to skip it for these tests — this private `prepare` helper is used only by a couple of test cases right next to it, so it's OK that they depend on some details of it which aren't clearly described.

It'd be good to add a little comment here linking to the above PR comment to describe a cleaner way this could be done. That way if in the future we're doing something similar and consulting this code as an example, we'll see that and have the option to act on it.

gnprice (commented):
Thanks @PIG208 for fixing this, and @chrisbobbe for the previous reviews! Generally this looks good. Comments below.

test/widgets/emoji_reaction_test.dart
PIG208: Neat! I think we can make a helper for the list entries that returns a `Rect` too:

```dart
      Rect getListEntriesRect(WidgetTester tester) =>
        tester.getRect(find.byType(EmojiPickerListEntry).first)
          .expandToInclude(tester.getRect(find.byType(EmojiPickerListEntry).last));
```

PIG208:
Thanks! This is ready for review now.

test/widgets/emoji_reaction_test.dart
```diff
@@ -466,6 +466,88 @@ void main() {
...
+          .isGreaterThan(scrollViewRect.height);
+      }
+
+      testWidgets('no view padding', (tester) async {
+        await prepare(tester, viewPadding: FakeViewPadding.zero);
+
+        // The top edge of the list entries is padded by 8px from the top edge
+        // of the scroll view; the bottom edge is out of view.
+        Rect listEntriesRect = getListEntriesRect(tester);
+        check(scrollViewRect.top).equals(listEntriesRect.top - 8);
```
gnprice, line 513: nit:
```suggestion
        check(scrollViewRect).top.equals(listEntriesRect.top - 8);
```

then combine with the next check as a cascade

gnprice (commented):
Thanks! One nit — otherwise all now looks good.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Looks good; merging.

PR #1318, opened 2025-02-03
profile: Display user's local time

Komyyy:
I'll add docs, tests and PR message later.


tannal:
It seems that you are using full database now.
The 10y database doesn't contain time zone abbreviations.
And `user.timezone` can be `CET` or other abbreviations.
If you are going to use the 10y database for better performance, you might need some kind of mapping.

See https://github.com/zulip/zulip/blob/42bc59b8027a074b31054f5fc0f1e6440fdd5252/zerver/lib/timezone.py#L22-L24

> We don't provide any functions to get locations by time zone abbreviations because of the ambiguities.

See https://pub.dev/packages/timezone#location.

PR #1322, opened 2025-02-04
new_dm: Add UI for starting new DM conversations

### LINKED ISSUE #127: Support starting a new DM conversation

chimnayajith:
# Pull Request

## Description

This pull request adds the UI to starting new DM conversations.. A floating action button has been added to the `RecentDmConversationsPage`, which opens a bottom modal sheet, allowing users to select conversation participants and proceed to the `MessageListPage`.

Design reference: [Figma](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4903-31879&p=f&t=pQP4QcxpccllCF7g-0)

## Related Issues

- Closes #127 

### Light mode
<div>
  
  
  
  
  
</div>

### Dark mode
<div>
  
  
  
  
  
</div>

## Additional Notes

The user list is currently sorted based on the recency of direct messages.


PIG208:
Thanks for working on this @chimnayajith!

I took a quick look at the implementation and checked the design. There are places where it currently does not match the Figma, for example:

```dart
![image](https://github.com/user-attachments/assets/e3b4f690-f883-47ee-9252-b69b293c0e47)
          Container(
            width: MediaQuery.of(context).size.width,
            constraints: const BoxConstraints(
              minHeight: 44,
              maxHeight: 124,
            ),
            padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 8),
            decoration: BoxDecoration(
              color: designVariables.bgSearchInput,
            ),
            child: SingleChildScrollView(
              controller: scrollController,
              child: Row(
                children: [
                  Expanded(
                    child: Wrap(
                      spacing: 4,
                      runSpacing: 4,
```

where the horizontal padding should be 14px;

image

and the spacing between the pills should be 6px.

I think there might be more places like this, so please sure to check your PR with the design and make sure that they match exactly.

While working on the next revision, please also try to break down the sheet into reasonable widgets, instead of a single one that encompasses the entire new DM page. There are also things like collapsing the closing brackets into a single line; you can find examples of that throughout the codebase:

```
      // [...]
        child: SafeArea(
          child: SizedBox(height: 48,
            child: Center(
              child: ConstrainedBox(
                // TODO(design): determine a suitable max width for bottom nav bar
                constraints: const BoxConstraints(maxWidth: 600),
                child: Row(
                  children: [
                    for (final navigationBarButton in navigationBarButtons)
                      Expanded(child: navigationBarButton),
                  ])))))));
```

All those changes will help make your PR more reviewable.

At the high-level, I think the user filtering code should re-use the `AutocompleteView` API, similar to what `EmojiPicker` does. However, We don't have a implementation for user auto-completion. While `MentionAutocompleteView` is close, it is also responsible for other tasks. Supporting that might require a good amount of refactoring and is less suited for a new contributor. So let's just try to focus on getting the UI right for this PR.

chimnayajith:
@PIG208 Thanks for the review! I’ve made the necessary changes to match the Figma, and refactored the sheet into smaller widgets. I also followed the code style guidelines you mentioned.

I’ve pushed the revision—PTAL.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+          duration: const Duration(milliseconds: 200),
+          curve: Curves.easeOut);
+      }
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
+      height: MediaQuery.of(context).size.height * 0.95,
```
PIG208, line 104: It doesn't seem right to size it this way. If the goal is to avoid the top of the bottom sheet overlapping with the device's top inset, see `useSafeArea` on `showModalBottomSheet`.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+              } else {
+                selectedUserIds.add(userId);
+              }
+              scrollToSearchField();
+            });
+          }))]));
+  }
+}
+
+class NewDmHeader extends StatelessWidget {
```
PIG208, line 129: For helper widgets like this, let's make them private (`_NewDmHeader`) unless it is otherwise necessary.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Container(
+      padding: const EdgeInsets.all(8),
+      child: Row(
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        children: [
+          _buildBackButton(context, designVariables, zulipLocalizations),
+          Text(zulipLocalizations.newDmSheetScreenTitle,
+            style: TextStyle(color: designVariables.title,fontSize: 20, fontWeight: FontWeight.w600)),
```
PIG208, line 151: nit:
```suggestion
style: TextStyle(color: designVariables.title, fontSize: 20, fontWeight: FontWeight.w600)),
```

PIG208: To control font weights, we usually use `weightVariableTextStyle`. See its dartdoc and existing places that use it for reference.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+          _buildNextButton(context, store, designVariables, zulipLocalizations)]));
+  }
+
+  Widget _buildBackButton(BuildContext context, DesignVariables designVariables,
+      ZulipLocalizations zulipLocalizations) => TextButton(
```
PIG208, line 156: If the helpers already have access to a `BuildContext`, there is generally no need to pass `designVariables` etc around.

PIG208: Since this method is already pretty long, we should [use braces](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#use-braces-for-long-functions-and-methods).

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+    onPressed: onClose,
+    child: Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        Icon(ZulipIcons.chevron_left, size: 24, color: designVariables.icon),
+        Text(zulipLocalizations.newDmSheetBackButtonLabel,
+          style: TextStyle(fontSize: 20, color: designVariables.icon))]));
+
+  Widget _buildNextButton(BuildContext context, PerAccountStore store,
+      DesignVariables designVariables, ZulipLocalizations zulipLocalizations) => TextButton(
```
PIG208, line 166: Similar here. Removing the other parameters will help make this more readable.

lib/widgets/new_dm_sheet.dart
PIG208: I think it might actually be cleaner to extract these helpers as local variables in `NewDmHeader.build`.

If you look at the other `Widget _buildSomething` style helpers in the code base with `git grep -A2 'Widget _build'`, you may find that they often have other parameters not tied to the `BuildContext`.

lib/widgets/new_dm_sheet.dart
PIG208: The design also specifies the line-height for this element.

image

The Flutter equivalent of this is `height`/`fontSize`. You can find some examples like:

```dart
      child: Text(label(zulipLocalizations),
        style: const TextStyle(fontSize: 20, height: 24 / 20)
```

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+    onPressed: selectedUserIds.isEmpty ? null : () {
+      final narrow = DmNarrow.withOtherUsers(
+        selectedUserIds,
+        selfUserId: store.selfUserId);
+      Navigator.pop(context);
+      Navigator.push(context, MessageListPage.buildRoute(
+        context: context,
+        narrow: narrow));
+    },
+    child: Row(
```
PIG208, line 176: Similar to a previous comment, we should pay close attention to the padding and reflect that in the code.

image

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        Text(zulipLocalizations.newDmSheetNextButtonLabel,
+          style: TextStyle(
+            fontSize: 20,
```
PIG208, line 181: Similar comment on line-height here. Be sure to check other `TextStyle`'s as well.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+            fontWeight: FontWeight.w600,
+            color: selectedUserIds.isEmpty
```
PIG208, line 183: It might be useful to extract the color as a local variable so we don't have to repeat it later.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+  final ScrollController scrollController;
+  final Set<int> selectedUserIds;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Container(
+      width: MediaQuery.of(context).size.width,
```
PIG208, line 210: Is `MediaQuery` necessary here? I guess this is for avoiding the device insets, but that's better handled with safe areas. See my other comment on using `useSafeArea`.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+      constraints: const BoxConstraints(
+        minHeight: 44,
+        maxHeight: 124),
+      padding: const EdgeInsets.symmetric(horizontal: 14,vertical:11),
```
PIG208, line 214: nit:
```suggestion
      padding: const EdgeInsets.symmetric(horizontal: 14, vertical: 11),
```

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+          Expanded(child: Wrap(
+            spacing: 6,
+            runSpacing: 4,
+            crossAxisAlignment: WrapCrossAlignment.center,
+            children: [
+              ...selectedUserIds.map((userId) => SelectedUserChip(userId: userId)),
+              _buildSearchField(designVariables, zulipLocalizations)]))])));
+  }
+
+  Widget _buildSearchField(DesignVariables designVariables,
```
PIG208, line 228: Similar issue on helper vs. local variable. Usually we don't pass `designVariables` and etc. around.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+      ZulipLocalizations zulipLocalizations) => IntrinsicWidth(
```
PIG208, line 229: Is `IntrinsicWidth` necessary for this? Looking at the constructor's dartdoc:

```dart
  /// Creates a widget that sizes its child to the child's intrinsic width.
  ///
  /// This class is relatively expensive. Avoid using it where possible.
```

In general it would be good to avoid expensive widgets, unless there is a reason to do so, documented with comment on why it is necessary.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+        child: SizedBox(
+          height: 22,
+          child: TextField(
+            controller: searchController,
+            cursorColor: designVariables.foreground,
+            style: const TextStyle(
+              fontSize: 17,
+              height: 1.0,
```
PIG208, line 237: In the design, the line-height is 22px:

image

See my other comment on `height`. We usually express them as fractions.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+            ),
+            scrollPadding: EdgeInsets.zero,
+            decoration: InputDecoration(
+              hintText: zulipLocalizations.newDmSheetSearchHint,
+              hintStyle: TextStyle(
+                fontSize: 17,
+                height: 1.0,
+                color: designVariables.textInput.withFadedAlpha(0.5)),
```
PIG208, line 245: Looks like we do have a Figma variable for this: `label-search-prompt`.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+
+    return Container(
+      height: 22,
+      decoration: BoxDecoration(color: designVariables.bgMenuButtonSelected, borderRadius: BorderRadius.circular(3)),
+      child: Row(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          Avatar(userId: userId, size: 22, borderRadius: 3),
+          Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 5, vertical: 3),
```
PIG208, line 271: The end inset is not 5px:

image

```suggestion
            padding: EdgeInsetsDirectional.fromSTEB(5, 3, 4, 3),
```

We use `EdgeInsetsDirectional` because the start/end insets are different. This is for RTL where left and right are flipped.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+class NewDmUserList extends StatelessWidget {
+  const NewDmUserList({
+    super.key,
+    required this.filteredUsers,
+    required this.selectedUserIds,
+    required this.onUserSelected});
+
+  final List<User> filteredUsers;
+  final Set<int> selectedUserIds;
+  final void Function(int) onUserSelected;
```
PIG208, line 289: The name of this is a bit misleading, because "selected" seems to indicate that the user is added to the list of selected user, when in reality it is more like toggling the selection of the user.

How about something like `void Function(int userId) onUserTapped`, with the parameter name.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    return ListView.builder(
```
PIG208, line 295: There is an 8px top padding that applies to the scrollable:

image

I assume that's something that allows you to scroll over. Perhaps `ListView.padding` will be helpful.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+      itemBuilder: (context, index) {
+        final user = filteredUsers[index];
+        final isSelected = selectedUserIds.contains(user.userId);
+
+        return InkWell(
+          onTap: () => onUserSelected(user.userId),
+          child: Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 2),
+            child: Container(
+              padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 6),
```
PIG208, line 306: Is the outer `Padding` redundant?

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+            return bLatestMessageId.compareTo(aLatestMessageId);
+          }
+          if (aLatestMessageId != null) return -1;
+          if (bLatestMessageId != null) return 1;
+          return 0;
+        });
+    });
+  }
+
+  // Scroll to the search field when the user selects a user
```
PIG208, line 89: Hmm, is this part of the UX specified in the design? It would be good to link to the relevant discussion/Figma if that's the case. Either way, I feel that it might be the best to leave this out from the initial implementation, to simplify things as we work it out.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+            fontSize: 20,
+            fontWeight: FontWeight.w600,
+            color: selectedUserIds.isEmpty
+              ? designVariables.icon.withFadedAlpha(0.5)
+              : designVariables.icon)),
+        Icon(ZulipIcons.chevron_right,
+          size: 24,
+          color: selectedUserIds.isEmpty
+            ? designVariables.icon.withFadedAlpha(0.5)
+            : designVariables.icon)]));
```
PIG208, line 190: Square brackets are usually not wrapped, because they help indicate the end of a list and that the items before them are parallel.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,327 @@
...
+}
+
+class NewDmSearchBar extends StatelessWidget {
+  const NewDmSearchBar({
+    super.key,
+    required this.searchController,
+    required this.scrollController,
+    required this.selectedUserIds});
```
PIG208, line 198: Similarly, curly brackets are not wrapped, because they indicate the end of a parameter list, function, etc.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -61,7 +65,25 @@ class _RecentDmConversationsPageBodyState extends State<RecentDmConversationsPag
...
+        })),
+      floatingActionButton: SizedBox(
+        width: 137,
+        height: 48,
+        child: FloatingActionButton.extended(
+          onPressed: () {
+            showNewDmSheet(context);
+          },
+          icon: Icon(Icons.add, size: 24),
+          label: Text(zulipLocalizations.newDmFabButtonLabel, style: TextStyle(fontSize: 20, fontWeight: FontWeight.w500)),
```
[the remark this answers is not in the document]
chimnayajith: The `extendedIconLabelSpacing` field for extended floating action buttons has a default value of 8.0. Should it still be specified?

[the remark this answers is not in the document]
PIG208: Yeah. Because the default value can change, but we have a specific value in mind.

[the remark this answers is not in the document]
chimnayajith: Have given it in the revision pushed.

alya:
@chimnayajith Are the screenshots in the PR description up to date? Please update them if not.

chimnayajith:
I've updated the screenshots.

alya:
Let's change "Add Person" to "Add user". I'm not sure why your screenshots have varied capitalization; only the first word should be capitalized.

alya:
Actually, it would probably be better to match the web app more fully, if it doesn't feel too long in the mobile context.

- "Add one or more users" before anyone is selected
- "Add another user…" once someone is selected
-

alya:
Why is there no "Add Person" in your last screenshot? What is different vs. the other ones?

chimnayajith:
Continuing the discussion in [CZO](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/New.20DM.20Sheet.20.23F1322/with/2107851)

chimnayajith:
@PIG208 Pushed a revision with the mentioned changes in UI. PTAL

lib/widgets/recent_dm_conversations.dart
```diff
@@ -48,8 +51,11 @@ class _RecentDmConversationsPageBodyState extends State<RecentDmConversationsPag
 
   @override
   Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
     final sorted = model!.sorted;
-    return SafeArea(
+    return Scaffold(
+      body: SafeArea(
```
gnprice, line 58: This will mean a `Scaffold` nested inside a `Scaffold`. See the Scaffold docs on that:
https://main-api.flutter.dev/flutter/material/Scaffold-class.html#nested-scaffolds

Have you tried alternatives to nesting the Scaffolds?

lib/widgets/recent_dm_conversations.dart
```diff
@@ -60,7 +66,28 @@ class _RecentDmConversationsPageBodyState extends State<RecentDmConversationsPag
...
+          onPressed: () {
+            showNewDmSheet(context);
+          },
+          extendedPadding: const EdgeInsetsDirectional.fromSTEB(16, 12, 20, 12),
+          icon: const Icon(Icons.add, size: 24),
+          label: Text(
+            zulipLocalizations.newDmFabButtonLabel,
+            style: const TextStyle(fontSize: 20, height: 24 / 20)
+              .merge(weightVariableTextStyle(context, wght: 500))),
+          backgroundColor: designVariables.fabBg,
```
gnprice, line 85: Trying out the PR, this color doesn't look right.

It looks like the screenshots in the PR description agree with what I see running it (good), but don't agree on this color with what's in Figma.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -60,7 +66,28 @@ class _RecentDmConversationsPageBodyState extends State<RecentDmConversationsPag
             narrow: narrow,
             unreadCount: unreadsModel!.countInDmNarrow(narrow),
           );
-        }));
+        })),
+      floatingActionButton: Container(
+        constraints: const BoxConstraints(
+          minWidth: 137,
+          minHeight: 48),
```
gnprice, line 73: Where do these numbers come from?

When I run the PR, the button looks too big, with oversized padding:
image

gnprice (commented):
Thanks @chimnayajith!

This is still in maintainer review with @PIG208, but I took a quick skim just now and also tried playing with this in the running app. Here's one high-level comment on the code, and a couple of things that jumped out at me in the UX.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+
+  final Set<int> selectedUserIds;
+  final VoidCallback onClose;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Row(
```
PIG208, line 130: The height of this in the Figma design is 44px, but the height I got is 60px:

image

The devtool shows that both "Back" and "Next" are 60px in height, so finding where the paddings come from in those widgets will be a good start:

image

A debugging tip is to utilize useful information about the layout hierarchy, the widget configurations, etc from the Flutter devtools.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+        _buildNextButton(context),
+      ]);
+  }
+
+  Widget _buildBackButton(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Padding(
+      padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 6),
```
PIG208, line 147: The distance to the top of the modal from the top edge of the back button should be 10px:

image.

The padding on the back button and the next button probably fits better inline in the build method, because those are related to how the buttons are positioned in the header, not a part of the buttons themselves (after all, these padded areas are not tappable)

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+  const _NewDmHeader({
+    required this.selectedUserIds,
+    required this.onClose,
+  });
+
+  final Set<int> selectedUserIds;
+  final VoidCallback onClose;
+
+  @override
+  Widget build(BuildContext context) {
```
PIG208, line 126: nit: move `build` to the end of the class.

See also: [Flutter style guide](https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense)

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        _buildBackButton(context),
+        Text(zulipLocalizations.newDmSheetScreenTitle,
+          style: TextStyle(color: designVariables.title,
+            fontSize: 20, height: 30 / 20)
+            .merge(weightVariableTextStyle(context, wght: 600))),
```
PIG208, line 137: This `Text` should have the same vertical padding that matches the design. Since all these elements have the same padding, it might be cleaner to apply it on the entire `Row`.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+        hintText: hintText,
+        hintStyle: TextStyle(color: designVariables.labelSearchPrompt,
+          fontSize: 17, height: 22 / 17),
+        border: InputBorder.none));
+  }
+}
+
+class _SelectedUserChip extends StatelessWidget {
+  const _SelectedUserChip({
+    required this.userId
```
PIG208, line 252: nit: this can be reduced to a single line, since the argument list is short enough

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+  });
+
+  final int userId;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final store = PerAccountStoreWidget.of(context);
+    final user = store.getUser(userId)!;
```
PIG208, line 261: Hmm, I don't feel confident that the `userId` will always present. If it does, we need a comment justifying why it is always present; if it doesn't, we should be more defensive and find a way to not throw.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+
+    return Container(
+    decoration: BoxDecoration(color: designVariables.bgMenuButtonSelected, borderRadius: BorderRadius.circular(3)),
```
PIG208, line 264: nit: missing indentation; this line should also be wrapped because it is too long (>80 characters)

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        _buildBackButton(context),
+        Text(zulipLocalizations.newDmSheetScreenTitle,
+          style: TextStyle(color: designVariables.title,
+            fontSize: 20, height: 30 / 20)
+            .merge(weightVariableTextStyle(context, wght: 600))),
+        _buildNextButton(context),
```
PIG208, line 138: This overflows on the right with a large font size setting:

image

The simplest workaround is probably just make the center element `Expanded`. It might be tricky to get it right, so let's leave the full fix for later.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+
+    return ListView.builder(
+      padding: EdgeInsets.all(8),
+      itemCount: filteredUsers.length,
+      itemBuilder: (context, index) {
+        final user = filteredUsers[index];
+        final isSelected = selectedUserIds.contains(user.userId);
+
+        return InkWell(
+          onTap: () => onUserTapped(user.userId),
```
PIG208, line 303: Let's use `splashFactory: NoSplash.splashFactory,` here; in general we avoid having Material's splash effect in the app.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,331 @@
...
+                Expanded(
+                  child: Padding(
+                    padding: const EdgeInsetsDirectional.fromSTEB(8, 6.5, 0, 6.5),
+                    child: Text(user.fullName,
+                      style: TextStyle(
+                        fontSize: 17,
+                        height: 19 / 17,
+                        color: designVariables.textMessage)
+                      .merge(weightVariableTextStyle(context, wght: 500))))),
+              ])));
```
PIG208, line 328: This group of widgets should have some vertical and right paddings:

image

I see that the leading check_circle icon is the tallest widget in the row by default, but that might not be the case with non-default font setting.

PIG208 (commented):
Thanks for the update! I went through the implementation but haven't looked at the tests in this round. Most of the colors look good to me. The font height/size/weight also looks good. Once all the comments are addressed, this should match the design quite well. Using devtool will be helpful to confirm the actual sizes/paddings.

Gaurav-Kushwaha-1225:
Hii @chimnayajith, 
The PR seems to be silent for 2 weeks. Are you still working on it?

lib/widgets/new_dm_sheet.dart
chimnayajith: Isn't the height 46 px?

chimnayajith:
@PIG208 Pushed a revision. PTAL.

lib/widgets/new_dm_sheet.dart
PIG208: Oops. Yeah, that's right.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,345 @@
...
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Container(
+      padding: const EdgeInsetsDirectional.fromSTEB(8, 10, 8, 6),
+      child: Row(
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        children: [
+          _buildBackButton(context),
+          Expanded(
+            child: Center(
```
PIG208, line 189: Looks like we can skip `Center` because `Text` itself is centered.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,146 @@
...
+      );
+
+      await tester.tap(userTile);
+      await tester.pump();
+      check(find.byIcon(Icons.check_circle)).findsOne();
+      check(find.byIcon(Icons.circle_outlined)).findsNothing();
+
+      final nextButton = tester.widget<GestureDetector>(
+        find.widgetWithText(GestureDetector, 'Next'));
+      check(nextButton.onTap).isNotNull();
```
PIG208, line 105: It might be a good idea to check `onTap` before selecting and after deselecting the user tile.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,146 @@
...
+      await tester.tap(userTile);
+      await tester.pump();
+      await tester.tap(find.widgetWithText(GestureDetector, 'Next'));
+
+      check(pushedRoutes).last.isA<WidgetRoute>().page
+        .isA<MessageListPage>()
+        .initNarrow.equals(DmNarrow.withOtherUsers(
+          [user.userId],
+          selfUserId: eg.selfUser.userId));
+    }
```
PIG208, line 139: We can probably simplify this (and `setupSheet`) by just checking the DM message list title. A useful principle when writing widget tests is to aim for checking what a user might see while using the app.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,146 @@
...
+
+    testWidgets('navigates to message list on Next', (tester) async {
+      final user = eg.user(userId: 1, fullName: 'Test User');
```
PIG208, line 142: We can skip specifying `userId` because we don't rely on it being this particular value for the test.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,146 @@
...
+      final nextButton = tester.widget<GestureDetector>(
+        find.widgetWithText(GestureDetector, 'Next'));
+      check(nextButton.onTap).isNull();
+    });
+
+    testWidgets('shows filtered users based on search', (tester) async {
+      final users = [
+        eg.user(userId: 1, fullName: 'Alice Anderson'),
+        eg.user(userId: 2, fullName: 'Bob Brown'),
+        eg.user(userId: 3, fullName: 'Alice Carter'),
```
PIG208, line 75: nit: we can skip specifying user IDs here

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,146 @@
...
+  required List<User> users,
+  NavigatorObserver? navigatorObserver,
+}) async {
+  addTearDown(testBinding.reset);
+  await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  await store.addUser(eg.selfUser);
+  for (final user in users) {
+    await store.addUser(user);
+  }
```
PIG208, line 31: nit: use `store.addUsers` instead

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,146 @@
...
+      check(find.text('Next')).findsOne();
+      check(find.byIcon(ZulipIcons.chevron_left)).findsOne();
+      check(find.byIcon(ZulipIcons.chevron_right)).findsOne();
+
+      final nextButton = tester.widget<GestureDetector>(
+        find.widgetWithText(GestureDetector, 'Next'));
+      check(nextButton.onTap).isNull();
+    });
+
+    testWidgets('shows filtered users based on search', (tester) async {
```
PIG208, line 71: I think there are some other interesting filtering logic that can be tested. Such as checking that the self user is excluded from the list, the search is case-insensitive. Another useful test would be checking if unknown users are handled correctly.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,345 @@
...
+@visibleForTesting
+class NewDmPicker extends StatefulWidget {
+  const NewDmPicker({
+    super.key,
+    required this.pageContext,
+    required this.store,
+  });
+
+  final BuildContext pageContext;
+  final PerAccountStore store;
```
PIG208, line 42: I think we can skip passing this by replacing the `_updateFilteredUsers` call with an overridden `didChangeDependencies`.

```
  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    _handleSearchUpdate();
  }
```

PIG208 (commented):
Thanks! This looks a lot closer to the design now and I left some small comments on the implementation. This time, I also reviewed the tests.

PIG208:
Hi @chimnayajith! Thanks for working on this. Since this issue is a launch blocker, we would prioritize reviewing this/getting this done first; the other issue that you are working on (#1344) is a launch goal, which means that it will likely receive less attention until we clear the blockers.

test/widgets/new_dm_sheet_test.dart
chimnayajith: I wasn't able to write tests specifically for handling unknown users, but the logic relies on `userDisplayName`, which is already covered in `user_test.dart`. I also didn’t find any other tests that handle this specifically.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,360 @@
...
+
+    if (filteredUsers.isEmpty) {
+      return Center(
+        child: Padding(
+          padding: const EdgeInsets.all(24),
+          child: Text(
+            zulipLocalizations.newDmSheetNoUsersFound,
+            style: TextStyle(
+              color: designVariables.labelMenuButton,
+              fontSize: 16))));
```
PIG208, line 318: This will need a `TODO(design)` comment since it is not a part of the Figma design.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,360 @@
...
+        final user = filteredUsers[index];
+        final isSelected = selectedUserIds.contains(user.userId);
+
+        return InkWell(
+          splashFactory: NoSplash.splashFactory,
+          onTap: () => onUserTapped(user.userId),
+          child: Container(
+            decoration: isSelected ? BoxDecoration(
+              color: designVariables.bgMenuButtonSelected,
+              borderRadius: BorderRadius.circular(10)) : null,
```
PIG208, line 334: nit: `Container` shouldn't be necessary; we can use `DecoratedBox`

another nit: I think we usually have the `null` case first, for reasons similar to why early returns are preferred.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,360 @@
...
+                  Avatar(userId: user.userId, size: 32, borderRadius: 3),
+                  Expanded(
+                    child: Padding(
+                      padding: const EdgeInsetsDirectional.fromSTEB(8, 6.5, 0, 6.5),
+                      child: Text(store.userDisplayName(user.userId),
+                        style: TextStyle(
+                          fontSize: 17,
+                          height: 19 / 17,
+                          color: designVariables.textMessage)
+                            .merge(weightVariableTextStyle(context, wght: 500))))),
```
PIG208, line 356: nit:
```suggestion
                          color: designVariables.textMessage,
                        ).merge(weightVariableTextStyle(context, wght: 500))))),
```
This way the end boundary of `TextStyle` is clear.

test/widgets/new_dm_sheet_test.dart
PIG208: I realized that it's a product question as to whether we want to allow selecting the self user. I think we should, since that's allowed on web and the legacy app.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,257 @@
...
+        if (users.length == 1) {
+          check(find.descendant(
+            of: find.byType(ZulipAppBar),
+            matching: find.text('DMs with ${users[0].fullName}'))).findsOne();
+        } else {
+          final names = users.map((user) => user.fullName).join(', ');
+          check(find.descendant(
+            of: find.byType(ZulipAppBar),
+            matching: find.text('DMs with $names'))).findsOne();
+        }
```
PIG208, line 238: It will probably be easier to pass the expected app bar title from the tests.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -138,3 +149,55 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+    final fabBgColor = _pressed ? designVariables.fabBgPressed : designVariables.fabBg;
+    final fabLabelColor = _pressed ? designVariables.fabLabelPressed : designVariables.fabLabel;
+
+    return InkWell(
+      onTap: widget.onPressed,
+      onTapDown: (_) => setState(() => _pressed = true),
+      onTapUp: (_) => setState(() => _pressed = false),
+      onTapCancel: () => setState(() => _pressed = false),
+      borderRadius: BorderRadius.circular(28),
+      child: Container(
```
PIG208, line 178: Let's use the animated version `AnimatedContainer` here. The Figma design has specified the duration and curve for transitions.

Looks like it also handles the box shadow animation properly, so we can just complete the TODO from https://github.com/zulip/zulip-flutter/pull/1322#discussion_r2031996733, now that we do have a AnimatedScaleOnTap style widget.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -138,3 +149,55 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+        child: Row(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            Icon(Icons.add, size: 24, color: fabLabelColor),
+            const SizedBox(width: 8),
+            Text(
+              zulipLocalizations.newDmFabButtonLabel,
+              style: TextStyle(fontSize: 20, height: 24 / 20)
+                .merge(weightVariableTextStyle(context, wght: 500))
+                .copyWith(color: fabLabelColor)),
```
PIG208, line 200: It should be fine to just specify the color when constructing `TextStyle`, right?

lib/widgets/recent_dm_conversations.dart
```diff
@@ -138,3 +149,55 @@ class RecentDmConversationsItem extends StatelessWidget {
           ]))));
   }
 }
+
+class _NewDmButton extends StatefulWidget {
+  final VoidCallback onPressed;
```
PIG208, line 154: nit: inline the callback in the `_NewDmButtonState.build` so that we can remove this parameter; it seems fitting that this widget is responsible for calling `showNewDmSheet`

lib/widgets/recent_dm_conversations.dart
```diff
@@ -138,3 +149,55 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+
+class _NewDmButtonState extends State<_NewDmButton> {
+  bool _pressed = false;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final fabBgColor = _pressed ? designVariables.fabBgPressed : designVariables.fabBg;
+    final fabLabelColor = _pressed ? designVariables.fabLabelPressed : designVariables.fabLabel;
```
PIG208, line 170: nit: these lines are too long; wrap them so that they stay under 80 characters in width

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,257 @@
...
+import 'package:zulip/widgets/store.dart';
+
+import '../api/fake_api.dart';
+import '../example_data.dart' as eg;
+import '../model/binding.dart';
+import '../model/test_store.dart';
+import 'test_app.dart';
+
+Future<void> setupSheet(WidgetTester tester, {
+  required List<User> users
```
PIG208, line 20: nit: missing trailing comma

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,257 @@
...
+        final userTile = find.ancestor(
+          of: find.text(user.fullName),
+          matching: find.byType(InkWell));
+        await tester.tap(userTile);
+        await tester.pump();
+        check(find.byIcon(Icons.check_circle)).findsOne();
+        check(find.byIcon(Icons.circle_outlined)).findsNothing();
+        nextButton = tester.widget<GestureDetector>(findNextButton());
+        check(nextButton.onTap).isNotNull();
+      });
```
PIG208, line 163: nit:

```suggestion
      testWidgets('selecting a user', (tester) async {
        final user = eg.user(fullName: 'Test User');
        final userTileFinder = find.ancestor(
          of: find.text(user.fullName),
          matching: find.byType(InkWell));
        await setupSheet(tester, users: [user]);

        final nextButton = tester.widget<GestureDetector>(
          find.widgetWithText(GestureDetector, 'Next'));
        check(find.byIcon(Icons.circle_outlined)).findsOne();
        check(find.byIcon(Icons.check_circle)).findsNothing();
        check(nextButton.onTap).isNull();

        await tester.tap(userTileFinder);
        await tester.pump();
        check(find.byIcon(Icons.circle_outlined)).findsNothing();
        check(find.byIcon(Icons.check_circle)).findsOne();
        check(nextButton.onTap).isNotNull();
      });
```

By breaking the checks into two stanzas, we can make this a bit more readable. (We don't need to call `tester.widget()` twice to see updates to `nextButton.onTap`.)

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -106,6 +108,24 @@ void main() {
       await tester.pumpAndSettle();
       check(tester.any(oldestConversationFinder)).isTrue(); // onscreen
     });
+
+    testWidgets('opens new DM sheet on FAB tap', (tester) async {
+      await setupPage(tester, users: [], dmMessages: []);
+      final fab = find.ancestor(
```
PIG208, line 114: nit: Let's [avoid abbreviations](github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-abbreviations) like this (here and in the test name) when possible, unless it comes from an outside source like the Figma design. In this case, we are just referring to the new DM button widget.

PIG208 (commented):
Thanks for the update! We have gone through the design in previous rounds. In this review I focused mostly on new changes like `_NewDmButton` and the tests.

test/widgets/new_dm_sheet_test.dart
chimnayajith: > We don't need to call `tester.widget()` twice to see updates to `nextButton.onTap`.

I tried implementing this change, but I'm running into some issues - the tests are failing when I remove the second call.
When I only get the button once before the tap and then check its `onTap` property after the UI update, the test fails as if the property isn't being updated.
Could you help me understand the correct approach here?

chimnayajith:
@PIG208 Pushed a revision. PTAL!

test/widgets/new_dm_sheet_test.dart
PIG208: I see. Yeah, it looks like the return values of the `tester.widget` call before and after the pump are different. Let's keep calling them again then.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,385 @@
...
+          && !selectedUserIds.contains(store.selfUserId);
+
+        if (user.userId != store.selfUserId) return true;
+        return !shouldExcludeSelfUser;
+      }
+
+      filteredUsers = store.allUsers
+        .where((user) =>
+          shouldIncludeSelfUser(user) &&
+          !user.isBot &&
```
PIG208, line 89: I think we shouldn't exclude bot users. Some bots might actually utilize DMs as a way of receiving commands from the user.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,385 @@
...
+              style: TextStyle(
+                color: designVariables.title,
+                fontSize: 20,
+                height: 30 / 20,
+              ).merge(weightVariableTextStyle(context, wght: 600)),
+              overflow: TextOverflow.ellipsis,
+              textAlign: TextAlign.center)),
+          _buildNextButton(context),
+        ])
+    );
```
PIG208, line 220: nit: move trailing parenthesis to the previous line

lib/widgets/recent_dm_conversations.dart
```diff
@@ -138,3 +148,64 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+        curve: Curves.easeOut,
+        padding: const EdgeInsetsDirectional.fromSTEB(16, 12, 20, 12),
+        decoration: BoxDecoration(
+          color: fabBgColor,
+          borderRadius: BorderRadius.circular(28),
+          boxShadow: [
+            BoxShadow(
+              color:_pressed
+                ? const Color(0xff2B0E8A).withValues(alpha: 0.3)
+                : const Color(0xff2B0E8A).withValues(alpha: 0.4),
```
PIG208, line 190: Hmm, so these are not named variables in the Figma design. In dark mode, the shadow is barely visible. Let's also have a TODO(design) comment here, mentioning dark mode colors. Perhaps it is also helpful to discussion in #mobile-design on whether we want the shadows in dark mode or not.

PIG208 (commented):
Thanks for the update! Left some comments, most are small, except for ones on adding more test cases.

chimnayajith:
Started a [CZO thread](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Excluding.20self.20user.20in.20new.20dm.20sheet.2E/with/2163150) to discuss the filtering logic.

lib/widgets/recent_dm_conversations.dart
chimnayajith: Opened a design thread for this in CZO [here](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Button.20shadows.20in.20dark.20mode/with/2164710)

chimnayajith:
@PIG208 Pushed a revision. PTAL!

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,293 @@
...
+      testWidgets('deselecting a user', (tester) async {
+        final user = eg.user(fullName: 'Test User');
+        final userTileFinder = find.ancestor(
+          of: find.text(user.fullName),
+          matching: find.byType(InkWell));
+        await setupSheet(tester, users: [user]);
+
+        await tester.tap(userTileFinder);
+        await tester.pump();
+
```
PIG208, line 161: We should check `nextButton` here as well, so that we know that it's precisely the second tap that toggles the user selection

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,293 @@
...
+        final selfUserTileFinder = find.ancestor(
+          of: find.text(eg.selfUser.fullName),
+          matching: find.byType(InkWell));
+        await setupSheet(tester, users: [otherUser]);
+
+        await tester.tap(selfUserTileFinder);
+        await tester.pump();
+        check(find.descendant(
+          of: selfUserTileFinder,
+          matching: find.byIcon(Icons.check_circle))).findsOne();
```
PIG208, line 184: We should check `eg.selfUser.fullName` here too.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,293 @@
...
+      });
+
+      testWidgets('navigates to 1:1 DM on Next', (tester) async {
+        final user = eg.user(fullName: 'Test User');
+        await runAndCheck(
+          tester,
+          users: [user],
+          expectedAppBarTitle: 'DMs with Test User');
+      });
+      testWidgets('navigates to group DM on Next', (tester) async {
```
PIG208, line 280: nit: missing newline

test/widgets/new_dm_sheet_test.dart
```diff
@@ -0,0 +1,293 @@
...
+
+        check(find.byType(ComposeBox)).findsOne();
+      }
+
+      testWidgets('navigates to self DM on Next', (tester) async {
+        await runAndCheck(
+          tester,
+          users: [],
+          expectedAppBarTitle: 'DMs with yourself',
+          isSelfDm: true);
```
PIG208, line 270: Hmm, how about just passing `users: [eg.selfUser]`? That way we don't need to introduce a separate parameter.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -138,3 +148,66 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+          boxShadow: [
+            BoxShadow(
+              // TODO(design): Shadow colors barely visible in dark mode
+              color: fabShadowColor,
+              blurRadius: _pressed ? 12 : 16,
+              spreadRadius: 0,
+              offset: _pressed
+                ? const Offset(0, 2)
+                : const Offset(0, 4)),
+          ]),
```
PIG208, line 198: Let's make `boxShadow` (normal and pressed states) design variables. We can use these values in light mode, and simply `null` in dark mode. See `ComposeBoxTheme` for example.

PIG208 (commented):
Thanks for the revisions! Just a handful of comments.

test/widgets/new_dm_sheet_test.dart
chimnayajith: Wouldn't the suggested change make the 'deselecting a user' test redundant with the 'selecting a user' test, since the latter already checks that nextButton.onTap becomes non-null after selection?

PIG208: Yeah, that's true. I think this probably suggests that these two tests can be combined together.

lib/widgets/recent_dm_conversations.dart
chimnayajith: Looking at the design discussion on [CZO thread](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Button.20shadows.20in.20dark.20mode/near/2171594) it appears the decision is to have shadows in dark mode as well.

Given this, do I still need to follow the `ComposeBoxTheme` pattern with a separate theme extension class, or should I simply continue using designVariables.fabShadow directly in the button code as currently implemented?

chimnayajith:
@PIG208 Pushed a revision. PTAL!

assets/l10n/app_en.arb
```diff
@@ -345,6 +345,34 @@
...
+  "@newDmFabButtonLabel": {
+    "description": "Label for the floating action button (FAB) that opens the new DM sheet."
+  },
+  "newDmSheetSearchHintEmpty": "Add one or more users",
+  "@newDmSheetSearchHintEmpty": {
+    "description": "Hint text for the search bar when no users are selected"
+  },
+  "newDmSheetSearchHintSomeSelected": "Add another user…",
+  "@newDmSheetSearchHintSomeSelected": {
+    "description": "Hint text for the search bar when at least one user is selected"
```
PIG208, line 370: nit: missing period at the end of the sentence

PIG208 (commented):
Thanks! Just a comment on the l10n descriptions. Marking this for Greg's review.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+      scrollPadding: EdgeInsets.zero,
+      decoration: InputDecoration(
+        isDense: true,
+        contentPadding: EdgeInsets.zero,
+        hintText: hintText,
+        hintStyle: TextStyle(
+          color: designVariables.labelSearchPrompt,
+          fontSize: 17,
+          height: 22 / 17),
+        border: InputBorder.none));
```
gnprice, line 251: nit: group properties logically; `border` is closely related to `contentPadding` (much more so than either of them is to the hint), so make them adjacent

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+    final user = store.getUser(userId);
+
+    return Container(
+      decoration: BoxDecoration(
+        color: designVariables.bgMenuButtonSelected,
+        borderRadius: BorderRadius.circular(3)),
+      child: Row(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          Avatar(userId: userId, size: 22, borderRadius: 3),
```
gnprice, line 299: nit: can make compact and reduce indentation:
```suggestion
      child: Row(mainAxisSize: MainAxisSize.min, children: [
        Avatar(userId: userId, size: 22, borderRadius: 3),
```

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+          Padding(
+            padding: const EdgeInsetsDirectional.fromSTEB(5, 3, 4, 3),
+            child: Text(user?.fullName ?? zulipLocalizations.unknownUserName,
```
gnprice, line 302: We have a centralized getter for this. (It might be new since you began work on this feature.) Search for references to `unknownUserName`.

gnprice (commented):
Thanks @chimnayajith for building this, and thanks @PIG208 for the previous reviews!

Comments below. For this round, I've read the main new file down through the header and serach bar. Left for a future round are the user list, the changes in other widgets files, and the tests.

gnprice:
I have this running now in the build on my device.

One quick comment from playing with it: when I select a user from the list of results, the search text I've entered should get cleared. Effectively by choosing a user I'm completing the name I had started typing as text.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+      child: SingleChildScrollView(
+        reverse: true,
+        child: Row(children: [
+          Expanded(child: Wrap(
+            spacing: 6,
+            runSpacing: 4,
+            crossAxisAlignment: WrapCrossAlignment.center,
+            children: [
+              ...selectedUserIds.map((userId) => _SelectedUserChip(userId: userId)),
+              IntrinsicWidth(child: _buildSearchField(context)),
```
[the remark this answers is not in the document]
chimnayajith: The IntrinsicWidth forces the TextField to be sized based on its content width rather than expanding to fill its parent. This allows the field to appear naturally sized alongside the user chips.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+        ]));
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return Container(
+      padding: const EdgeInsetsDirectional.fromSTEB(8, 10, 8, 6),
```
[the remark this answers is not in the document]
gnprice: Great, that's helpful, thanks.

It looks like the layout is structured differently within the Figma design: there's 4px top padding, then a 42px height, then the other things are centered vertically within that. But this version is simpler and seems fine.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+        children: [
+          _buildBackButton(context),
+          Expanded(
+            child: Text(zulipLocalizations.newDmSheetScreenTitle,
+              style: TextStyle(
+                color: designVariables.title,
+                fontSize: 20,
+                height: 30 / 20,
+              ).merge(weightVariableTextStyle(context, wght: 600)),
+              overflow: TextOverflow.ellipsis,
```
gnprice, line 212: Does this `overflow` parameter have any effect without `maxLines`? Reading docs, I believe it doesn't.

lib/widgets/new_dm_sheet.dart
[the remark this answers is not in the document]
gnprice: Is there an image of how it would look different without the IntrinsicWidth?

I think there was a discussion of this earlier; a link to that might be helpful.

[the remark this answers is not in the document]
chimnayajith: This is how it'll look without the `IntrinsicWidth`
image

Link to the CZO Discussion: [#mobile-dev-help>New DM Search Bar](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/New.20DM.20Search.20Bar/with/2114561).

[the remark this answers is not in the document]
gnprice: Great, OK: so the point is that without IntrinsicWidth, the place where the user types always winds up on a new line, after any pills for users already chosen. With IntrinsicWidth, it participates effectively in the Wrap and can go after a pill on the same line when there's room.

IIUC that happens because the Wrap widget passes loose constraints down to its child the text field, and the text field uses the max width available, which then forces the Wrap to put it on its own line. With an IntrinsicWidth involved, the latter gets those loose constraints from its parent Wrap, then asks its child the text field how wide its actual contents are (the "intrinsic width"), and forces the child to that size instead.

Let's include a link to this GitHub subthread in a comment, like this:
```suggestion
              // The IntrinsicWidth lets the text field participate in the Wrap
              // when its content fits on the same line with a user chip,
              // by preventing it from expanding to fill the available width.  See:
              //   https://github.com/zulip/zulip-flutter/pull/1322#discussion_r2094112488
              IntrinsicWidth(child: _buildSearchField(context)),
```

gnprice:
Vlad offered some updates and feedback for the design of this page:
[#mobile > Flutter DMs flow design review @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20DMs.20flow.20design.20review/near/2181176)

Specifically I think items 0, 1, 2, 3, 4, 5, 6 are things to be updated on this PR. (Item 5 is the same as my comment https://github.com/zulip/zulip-flutter/pull/1322#issuecomment-2878199404 above.)

gnprice:
Thanks again @chimnayajith for all your work on this PR!

The new app's launch is just a few weeks away now ([#mobile-team > Flutter beta timeline @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2174273)), and this is one of the M5a launch-blocker issues we want to be sure to have completed before then. Since it looks like you don't currently have time to make revisions to this PR, I've asked @chrisbobbe to pick it up from here so we can be sure to have it finished soon.

lib/widgets/new_dm_sheet.dart
```diff
@@ -0,0 +1,380 @@
...
+    return Container(
+      constraints: const BoxConstraints(
+        minHeight: 44,
+        maxHeight: 124),
+      padding: const EdgeInsets.symmetric(horizontal: 14, vertical: 11),
+      decoration: BoxDecoration(color: designVariables.bgSearchInput),
+      child: SingleChildScrollView(
+        reverse: true,
+        child: Row(children: [
+          Expanded(child: Wrap(
```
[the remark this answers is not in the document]
chimnayajith: But that would mean `Expanded` directly inside `SingleChildScrollView` which results in an error.

chimnayajith:
Pushed a revision. PTAL.
> Specifically I think items 0, 1, 3, 4, 5, 6 are things to be updated on this PR.

Had a few doubts about some of these items. Started a [CZO thread](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Follow.20up.20on.20comments.20.20on.20new.20Dm.20Sheet/with/2184971) for the same.


Also the `app_en.arb` and other related files seems to already have an older version of the strings coming from this commit -> [`bb1ca88`](https://github.com/zulip/zulip-flutter/commit/bb1ca8868489b439d89ec729f0764a6cf709370e)

lib/widgets/new_dm_sheet.dart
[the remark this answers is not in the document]
gnprice: Yeah, the Expanded is really part of the Row protocol — what I meant was to use the child of the Expanded directly.

It looks like Chris cleared this up as part of #1554:
https://github.com/zulip/zulip-flutter/commit/a8751e40424a0ea1f1c685693b63178f671ce107
So I guess what this Row-plus-Expanded was accomplishing was to make the Wrap take up the full width. That can be done more directly by setting `CrossAxisAlignment.stretch` on the parent Column.

gnprice:
Thanks @chimnayajith for the revision!

It looks like @chrisbobbe has sent a new version, based on this revision, as #1554. So I'll close this PR in order to continue the thread there. (See also https://github.com/zulip/zulip-flutter/pull/1554#issuecomment-2961170657 for Chris's detailed feedback on this revision, and https://github.com/zulip/zulip-flutter/pull/1554#pullrequestreview-2915241380 for my comments, many of which apply to this version.)

Issue #1319, opened 2025-02-03
Custom headers for messages with yourself


alya:
To fix the TODO below, we should label DMs with yourself as "Messages with yourself" in the DM header bar.

```
    if (message.allRecipientIds.length > 1) {
      title = // …
    } else {
      // TODO pick string; web has glitchy "You and $yourname"
      title = zulipLocalizations.messageListGroupYouWithYourself;
    }
```

Parallel web app issue: https://github.com/zulip/zulip/issues/33321

The "DMs with yourself" label in the Flutter app navbar is fine as is.

PR #1323, opened 2025-02-04
msglist: Update label to 'Messages with yourself' in DM header

lakshya1goel:
Fixes: #1319

## Screenshot
Image

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -404,7 +404,7 @@
       "others": {"type": "String", "example": "Alice, Bob"}
     }
   },
-  "messageListGroupYouWithYourself": "You with yourself",
+  "messageListGroupYouWithYourself": "Messages with yourself",
   "@messageListGroupYouWithYourself": {
     "description": "Message list recipient header for a DM group that only includes yourself."
   },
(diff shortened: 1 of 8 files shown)


chrisbobbe (changes requested):
```
i18n: Update label to "Messages with yourself" in DM header
```

commit-message nit: choose a different prefix than `i18n`. This commit involves the i18n subsystem, because it's changing a UI string and we translate those—but it isn't a change in the i18n subsystem itself. How about `msglist`.

chrisbobbe:

gnprice (commented):
Thanks for taking care of this! A few small comments.

---

There's a TODO comment where this is used:

```dart
    } else {
      // TODO pick string; web has glitchy "You and $yourname"
      title = zulipLocalizations.messageListGroupYouWithYourself;
```

We're taking care of that TODO with this change, so the comment should be deleted.

---

There was a chat thread where we decided to make this change, right? Let's have the commit message link to that thread.

---

In the commit message:

    Fixes: zulip#1319

It should say just `#1319`, not `zulip#1319` — the "zulip" part is redundant.

lakshya1goel:
Thanks for the review @gnprice, pushed the revision, PTAL.

gnprice:

PR #1325, opened 2025-02-05
Wrap up translating all strings

PIG208:
Fixes:
- #277

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -812,5 +812,9 @@
   "appVersionUnknownPlaceholder": "(…)",
   "@appVersionUnknownPlaceholder": {
     "description": "Placeholder to show in place of the app version when it is unknown."
+  },
+  "zulipAppTitle": "Zulip",
+  "@zulipAppTitle": {
+    "description": "The name of Zulip. This should be either 'Zulip' or a transliteration."
   }
 }
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -185,7 +185,7 @@
     "placeholders": {
       "num": {"type": "int", "example": "2"},
       "maxFileUploadSizeMib": {"type": "int", "example": "15"},
-      "listMessage": {"type": "String", "example": "foo.txt\nbar.txt"}
+      "listMessage": {"type": "String", "example": "foo.txt: 10.1 MiB\nbar.txt 20.2 MiB"}
     }
   },
   "errorFilesTooLargeTitle": "{num, plural, =1{File} other{Files}} too large",
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -483,9 +483,9 @@
   "@loginAddAnAccountPageTitle": {
     "description": "Title for page to add a Zulip account."
   },
-  "loginServerUrlInputLabel": "Your Zulip server URL",
-  "@loginServerUrlInputLabel": {
-    "description": "Input label in login page for Zulip server URL entry."
+  "loginServerUrlLabel": "Your Zulip server URL",
+  "@loginServerUrlLabel": {
+    "description": "Label in login page for Zulip server URL entry."
   },
   "loginHidePassword": "Hide password",
   "@loginHidePassword": {
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -808,5 +808,9 @@
   "scrollToBottomTooltip": "Scroll to bottom",
   "@scrollToBottomTooltip": {
     "description": "Tooltip for button to scroll to bottom."
+  },
+  "appVersionUnknownPlaceholder": "(…)",
+  "@appVersionUnknownPlaceholder": {
+    "description": "Placeholder to show in place of the app version when it is unknown."
   }
 }
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -171,6 +171,14 @@
       "filename": {"type": "String", "example": "file.txt"}
     }
   },
+  "filenameAndSizeInMiB": "{filename}: {size} MiB",
+  "@filenameAndSizeInMiB": {
+    "description": "The name of a file, and its size in mebibytes.",
+    "placeholders": {
+      "filename": {"type": "String", "example": "foo.txt"},
+      "size": {"type": "String", "example": "20.2"}
+    }
+  },
   "errorFilesTooLarge": "{num, plural, =1{File is} other{{num} files are}} larger than the server's limit of {maxFileUploadSizeMib} MiB and will not be uploaded:\n\n{listMessage}",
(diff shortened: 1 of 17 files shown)


chrisbobbe (commented):
Thanks, LGTM!

lib/widgets/content.dart
```diff
@@ -1586,8 +1586,9 @@ InlineSpan _errorUnimplemented(UnimplementedNode node, {required BuildContext co
       TextSpan(text: "»)", style: errorStyle),
     ]);
   } else {
+    final zulipLocalizations = ZulipLocalizations.of(context);
     return TextSpan(
-      text: "(unimplemented: DOM node type ${htmlNode.nodeType})",
+      text: zulipLocalizations.errorUnimplementedPlaceholder(htmlNode.nodeType),
```
gnprice, line 1591: I think this one is covered by the TODO comments above, so we can leave it out.

(I'm not sure I've ever seen this case happen, so there's no urgency to ask our translators to translate it before we have a solution for its neighbors. The `is dom.Element` case above is by far the most common.)

lib/licenses.dart
```diff
@@ -23,6 +23,9 @@ Stream<LicenseEntry> additionalLicenses() async* {
         rootBundle.loadString('assets/Pygments/AUTHORS.txt'),
       ]);
 
+      // This does not need to be translated, as it is just a small fragment
+      // of text surrounded by a large quantity of English text that isn't
+      // translated anyway.
       return '$licenseFileText\n\nAUTHORS file follows:\n\n$authorsFileText';
```
gnprice, line 29: Great, this is helpful.

The other key piece of the reasoning is the fact that it wouldn't be easy to translate — if it were, we should translate it rather than try to think through whether this is a corner we can get away with cutting. So, drawing on [that chat thread](https://chat.zulip.org/#narrow/channel/176-mobile-core/topic/Pygment.20license/near/2044034) you linked earlier:
```suggestion
      // This does not need to be translated, as it is just a small fragment
      // of text surrounded by a large quantity of English text that isn't
      // translated anyway.
      // (And it would be logistically tricky to translate, as this code is
      // called from the `main` function before the [ZulipApp] widget is built,
      // let alone has updated [GlobalLocalizations].)
      return '$licenseFileText\n\nAUTHORS file follows:\n\n$authorsFileText';
```

gnprice (commented):
Thanks! Comments below.

This also doesn't yet complete #277 — there are four more sites that need translations, as found at https://github.com/zulip/zulip-flutter/issues/277#issuecomment-2638297174 , in addition to the sites that need new or updated TODO markers which I've just sent in #1332.

PIG208:
Updated the PR to include the missed strings.

assets/l10n/app_en.arb
```diff
@@ -155,13 +155,21 @@
       "filename": {"type": "String", "example": "file.txt"}
     }
   },
+  "filenameAndSizeInMiB": "{filename}: {size} MiB",
+  "@filenameAndSizeInMiB": {
+    "description": "Message to display the name of a file and its size on a single line.",
```
gnprice, line 160:
```suggestion
    "description": "The name of a file, and its size in mebibytes.",
```

The abbreviation "MiB" may not be recognizable to all translators, so spelling it out gives them a better shot at successfully looking it up.

(That unit's name is one of the things translators will edit — for example in French it'll be "Mio", for "mébioctets".)

assets/l10n/app_en.arb
```diff
@@ -459,9 +467,13 @@
...
-  "loginServerUrlInputLabel": "Your Zulip server URL",
-  "@loginServerUrlInputLabel": {
-    "description": "Input label in login page for Zulip server URL entry."
+  "loginServerUrlLabel": "Your Zulip server URL",
+  "@loginServerUrlLabel": {
+    "description": "Label in login page for Zulip server URL entry."
+  },
+  "loginServerUrlHint": "your-org.zulipchat.com",
+  "@loginServerUrlHint": {
+    "description": "Hint text in login page for Zulip server URL entry."
```
gnprice, line 476: … Hmm. I started writing in my head a bit more for this description (the translation should keep the ".zulipchat.com" part), then realized: we actually can't have translators substituting their own translations of the phrase "your org" either.

That's because it's important that an example like this be [a name we've reserved](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/flutter.3A.20login.20URL/near/1570360) so that it won't be some actual Zulip organization.

So I'll drop this commit and instead write a comment saying why this isn't translated.

gnprice (commented):

assets/l10n/app_en.arb
```diff
@@ -792,5 +804,13 @@
...
   "@scrollToBottomTooltip": {
     "description": "Tooltip for button to scroll to bottom."
+  },
+  "appVersionUnknownPlaceholder": "(…)",
+  "@appVersionUnknownPlaceholder": {
+    "description": "Placeholder to show in place of the app version when it is unknown."
+  },
+  "zulipAppTitle": "Zulip",
+  "@zulipAppTitle": {
+    "description": "Title of the application that appears above the task manager's app snapshot when the user presses the \"recent apps\" button on Android. This is not used for iOS."
```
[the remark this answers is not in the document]
PIG208: "Зулип"

Ah, I see.

PIG208:

assets/l10n/app_en.arb
PIG208: Interesting. TIL!

gnprice:
Thanks for the revision! Merging, after updating a couple of the commit messages to match:
```
3:  4760127f2 ! 3:  0ac905c2b app: Translate app title
    @@ Metadata
     Author: Zixuan James Li <zixuan@zulip.com>
     
      ## Commit message ##
    -    app: Translate app title
    -
    -    The title right now is just "Zulip", which on its own should translate to
    -    "Zulip" in other languages. But because this is user-visible, we make it
    -    translatable, as we can always change it to a phrase or anything else.
    -
    -    This string is applicable for Android, not iOS. The description,
    -    adapted from the dartdoc of [WidgetsApp.title], for translator mentions
    -    this fact to describe context in which the string appears.
    -
    -    While this might also appear on other platforms like web, there is no
    -    need to bring this up as we do not currently support them.
    +    app: Translate (well, transliterate) app title
     
4:  31446d55e ! 4:  a08dcda2f compose: Translate message for a filename with size
    @@ Commit message
         The string is used at the end of the "errorFilesTooLarge" message,
         which includes a list of files with its size that are too large.
     
    -    We didn't mention that it is a part of the error message because this
    -    part of the context is not as relevant as that it fits in a single
    -    line.
    -
```

gnprice:
This fixed #277, but didn't cause that issue to be closed because the issue description didn't use the special GitHub syntax for it:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#mention-fixed-issue

I've edited the description and will close the issue.

Issue #59, opened 2023-04-10
Handle `<ol start=…>`


gnprice:
An `<ol>` element in Zulip HTML can have the `start` attribute, like so:
```html
<ol start="9999">
<li>Fjfjdj</li>
<li>Fjdjdj</li>
</ol>
```

On Zulip web, this [renders](https://chat.zulip.org/#narrow/stream/7-test-here/topic/large.20lists/near/1544557) like so:
![image](https://user-images.githubusercontent.com/28173/230997092-aac3638f-3119-45e1-b0e7-c7a10270c15b.png)

(The fact that the numbers are cut off at the left is a bug that's getting fixed: https://github.com/zulip/zulip/pull/25063.)

Currently we ignore `start` and always start the numbering at 1. We should instead respect `start` when we see it.

This means:
* When we parse the HTML, `ListNode` should get a bit more information to record `start`. Possibly make it abstract and give it subclasses `OrderedListNode` and `UnorderedListNode`, since `start` won't make sense on the latter.
* Then `ListNodeWidget` should act on that information.

gnprice:
The Zulip Markdown message source that produces the example HTML above is:
```markdown
9999. Fjfjdj
10000. Fjdjdj
```

gnprice:
As part of this issue, the changes to HTML parsing should get tests in `test/model/content_test.dart`. This will probably involve adding another case to `ContentNodeChecks.equalsNode` (in `content_checks.dart`) for `ListNode`, similar to the existing case for `ParagraphNode`.

Ideally the changes to `ListNodeWidget` should also get tests. But we don't yet really have an example of that kind of test (it'd go in `test/widgets/content_test.dart`), so it's OK to leave that out for this issue.

chplalex:
Hi, Greg.

I do not know all details of the environment for this task. 
That's why I have done it as simple utils extensions for html document and element. 

Please, see the result here:

https://github.com/chplalex/zulip_59

I would like feedback from you.

A little later I will look at your second task

Have a nice weekend!

gnprice:
Thanks @chplalex!

I think a separate standalone piece of code like that won't be very helpful for us, though.  What this issue is about is to add this feature in the context of our existing codebase.

[...]

PR #1329, opened 2025-02-05
content: Add start attribute support for ordered list

lakshya1goel:
Fixes: #59
Fixes: #1356

## Screenshot
| WhatsApp Image 2025-02-05 at 8 46 10 PM | WhatsApp Image 2025-02-14 at 11 32 54 PM |

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -251,20 +251,10 @@ class HeadingNode extends BlockInlineContainerNode {
   }
 }
 
-enum ListStyle { ordered, unordered }
+sealed class ListNode extends BlockContentNode {
+  const ListNode(this.items, {super.debugHtmlNode});
 
-class ListNode extends BlockContentNode {
-  const ListNode(this.style, this.items, {super.debugHtmlNode});
-
-  final ListStyle style;
   final List<List<BlockContentNode>> items;
-
-  @override
-  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
-    super.debugFillProperties(properties);
-    properties.add(FlagProperty('ordered', value: style == ListStyle.ordered,
-      ifTrue: 'ordered', ifFalse: 'unordered'));
-  }
 
   @override
   List<DiagnosticsNode> debugDescribeChildren() {
@@ -272,6 +262,22 @@ class ListNode extends BlockContentNode {
       .mapIndexed((i, nodes) =>
         _BlockContentListNode(nodes).toDiagnosticsNode(name: 'item $i'))
       .toList();
+  }
+}
+
+class UnorderedListNode extends ListNode {
+  const UnorderedListNode(super.items, {super.debugHtmlNode});
+}
+
+class OrderedListNode extends ListNode {
+  const OrderedListNode(super.items, {required this.start, super.debugHtmlNode});
+
+  final int start;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(IntProperty('start', start));
   }
 }
 
@@ -1108,12 +1114,7 @@ class _ZulipContentParser {
   }
 
   BlockContentNode parseListNode(dom.Element element) {
-    ListStyle? listStyle;
-    switch (element.localName) {
-      case 'ol': listStyle = ListStyle.ordered; break;
-      case 'ul': listStyle = ListStyle.unordered; break;
-    }
-    assert(listStyle != null);
+    assert(element.localName == 'ol' || element.localName == 'ul');
     assert(element.className.isEmpty);
 
     final debugHtmlNode = kDebugMode ? element : null;
@@ -1126,7 +1127,15 @@ class _ZulipContentParser {
       items.add(parseImplicitParagraphBlockContentList(item.nodes));
     }
 
-    return ListNode(listStyle!, items, debugHtmlNode: debugHtmlNode);
+    if (element.localName == 'ol') {
+      final startAttr = element.attributes['start'];
+      final start = startAttr == null ? 1
+        : int.tryParse(startAttr, radix: 10);
+      if (start == null) return UnimplementedBlockContentNode(htmlNode: element);
+      return OrderedListNode(items, start: start, debugHtmlNode: debugHtmlNode);
+    } else {
+      return UnorderedListNode(items, debugHtmlNode: debugHtmlNode);
+    }
   }
 
   BlockContentNode parseSpoilerNode(dom.Element divElement) {
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -503,33 +503,24 @@ class ListNodeWidget extends StatelessWidget {
         case UnorderedListNode(): marker = "•   "; break;
         case OrderedListNode(:final start): marker = "${start + index}. "; break;
       }
-      return ListItemWidget(marker: marker, nodes: item);
+      return TableRow(children: [
+        Align(
+          alignment: AlignmentDirectional.topEnd,
+          child: Text(marker)),
+        BlockContentList(nodes: item),
+      ]);
     });
+
     return Padding(
       padding: const EdgeInsets.only(top: 2, bottom: 5),
-      child: Column(children: items));
-  }
-}
-
-class ListItemWidget extends StatelessWidget {
-  const ListItemWidget({super.key, required this.marker, required this.nodes});
-
-  final String marker;
-  final List<BlockContentNode> nodes;
-
-  @override
-  Widget build(BuildContext context) {
-    return Row(
-      mainAxisAlignment: MainAxisAlignment.start,
-      crossAxisAlignment: CrossAxisAlignment.baseline,
-      textBaseline: localizedTextBaseline(context),
-      children: [
-        SizedBox(
-          width: 20, // TODO handle long numbers in <ol>, like https://github.com/zulip/zulip/pull/25063
-          child: Align(
-            alignment: AlignmentDirectional.topEnd, child: Text(marker))),
-        Expanded(child: BlockContentList(nodes: nodes)),
-      ]);
+      child: Table(
+        defaultVerticalAlignment: TableCellVerticalAlignment.baseline,
+        textBaseline: localizedTextBaseline(context),
+        columnWidths: const <int, TableColumnWidth>{
+          0: IntrinsicColumnWidth(),
+          1: FlexColumnWidth(),
+        },
+        children: items));
   }
 }
 
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -492,7 +492,7 @@ class ListNodeWidget extends StatelessWidget {
     final items = List.generate(node.items.length, (index) {
       final item = node.items[index];
       String marker;
-      switch (node.style) {
+      switch (node) {
         // TODO(#161): different unordered marker styles at different levels of nesting
         //   see:
         //     https://html.spec.whatwg.org/multipage/rendering.html#lists
@@ -500,9 +500,8 @@ class ListNodeWidget extends StatelessWidget {
         // TODO proper alignment of unordered marker; should be "• ", one space,
         //   but that comes out too close to item; not sure what's fixing that
         //   in a browser
-        case ListStyle.unordered: marker = "•   "; break;
-        // TODO(#59) ordered lists starting not at 1
-        case ListStyle.ordered: marker = "${index+1}. "; break;
+        case UnorderedListNode(): marker = "•   "; break;
+        case OrderedListNode(:final start): marker = "${start + index}. "; break;
       }
       return ListItemWidget(marker: marker, nodes: item);
     });
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -269,6 +269,26 @@ class ContentExample {
       url: '/#narrow/channel/378-api-design/topic/notation.20for.20near.20links/near/1972281',
       nodes: [TextNode('#api design > notation for near links @ 💬')]));
 
+  static const orderedListCustomStart = ContentExample(
+    'ordered list with custom start',
+    '5. fifth\n6. sixth',
+    '<ol start="5">\n<li>fifth</li>\n<li>sixth</li>\n</ol>',
+    [OrderedListNode(start: 5, [
+      [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('fifth')])],
+      [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('sixth')])],
+    ])],
+  );
+
+  static const orderedListLargeStart = ContentExample(
+    'ordered list with large start number',
+    '9999. first\n10000. second',
+    '<ol start="9999">\n<li>first</li>\n<li>second</li>\n</ol>',
+    [OrderedListNode(start: 9999, [
+      [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('first')])],
+      [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('second')])],
+    ])],
+  );
+
   static const spoilerDefaultHeader = ContentExample(
     'spoiler with default header',
     '```spoiler\nhello world\n```',
@@ -306,8 +326,8 @@ class ContentExample {
         '<p><em>italic</em> <a href="https://zulip.com/">zulip</a></p>\n'
         '</div></div>',
     [SpoilerNode(
-      header: [ListNode(ListStyle.ordered, [
-        [ListNode(ListStyle.unordered, [
+      header: [OrderedListNode(start: 1, [
+        [UnorderedListNode([
           [HeadingNode(level: HeadingLevel.h2, links: null, nodes: [
             TextNode('hello'),
           ])]
@@ -836,7 +856,7 @@ class ContentExample {
         '<div class="message_inline_image">'
           '<a href="https://chat.zulip.org/user_avatars/2/realm/icon.png">'
             '<img src="https://chat.zulip.org/user_avatars/2/realm/icon.png"></a></div></li>\n</ul>', [
-    ListNode(ListStyle.unordered, [[
+    UnorderedListNode([[
       ImageNodeList([
         ImageNode(srcUrl: 'https://chat.zulip.org/user_avatars/2/realm/icon.png',
           thumbnailUrl: null, loading: false,
@@ -858,7 +878,7 @@ class ContentExample {
         '<div class="message_inline_image">'
           '<a href="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=2" title="icon.png">'
             '<img src="https://chat.zulip.org/user_avatars/2/realm/icon.png?version=2"></a></div></li>\n</ul>', [
-    ListNode(ListStyle.unordered, [[
+    UnorderedListNode([[
       ParagraphNode(wasImplicit: true, links: null, nodes: [
         LinkNode(url: 'https://chat.zulip.org/user_avatars/2/realm/icon.png', nodes: [TextNode('icon.png')]),
         TextNode(' '),
@@ -887,7 +907,7 @@ class ContentExample {
           '<a href="https://chat.zulip.org/user_avatars/2/realm/icon.png" title="icon.png">'
             '<img src="https://chat.zulip.org/user_avatars/2/realm/icon.png"></a></div>'
         'more text</li>\n</ul>', [
-    ListNode(ListStyle.unordered, [[
+    UnorderedListNode([[
       const ParagraphNode(wasImplicit: true, links: null, nodes: [
         LinkNode(url: 'https://chat.zulip.org/user_avatars/2/realm/icon.png', nodes: [TextNode('icon.png')]),
         TextNode(' '),
@@ -1559,7 +1579,7 @@ void main() {
     testParse('<ol>',
       // "1. first\n2. then"
       '<ol>\n<li>first</li>\n<li>then</li>\n</ol>', const [
-        ListNode(ListStyle.ordered, [
+        OrderedListNode(start: 1, [
           [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('first')])],
           [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('then')])],
         ]),
@@ -1568,7 +1588,7 @@ void main() {
     testParse('<ul>',
       // "* something\n* another"
       '<ul>\n<li>something</li>\n<li>another</li>\n</ul>', const [
-        ListNode(ListStyle.unordered, [
+        UnorderedListNode([
           [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('something')])],
           [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('another')])],
         ]),
@@ -1577,7 +1597,7 @@ void main() {
     testParse('implicit paragraph with internal <br>',
       // "* a\n  b"
       '<ul>\n<li>a<br>\n  b</li>\n</ul>', const [
-        ListNode(ListStyle.unordered, [
+        UnorderedListNode([
           [ParagraphNode(wasImplicit: true, links: null, nodes: [
             TextNode('a'),
             LineBreakInlineNode(),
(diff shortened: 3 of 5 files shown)


lib/model/content.dart
```diff
@@ -1081,7 +1103,13 @@ class _ZulipContentParser {
       items.add(parseImplicitParagraphBlockContentList(item.nodes));
     }
 
-    return ListNode(listStyle!, items, debugHtmlNode: debugHtmlNode);
+    if (listStyle == ListStyle.ordered) {
+      final startAttr = element.attributes['start'];
+      final start = startAttr != null ? (int.tryParse(startAttr) ?? 1) : 1;
```
rajveermalviya, line 1108: nit: Can we simplify this condition, it's fine to introduce an extra variable.

test/widgets/content_test.dart
```diff
@@ -1158,4 +1158,19 @@ void main() {
...
     });
   });
+
+  group('ListNodeWidget', () {
+    testWidgets('ordered list with custom start', (tester) async {
+      await prepareContent(tester, plainContent('''
+        <ol start="3">
+          <li>third</li>
+          <li>fourth</li>
+        </ol>'''));
```
rajveermalviya, line 1168: nit: Use single-quouted strings
```suggestion
      await prepareContent(tester, plainContent('<ol start="3">\n<li>third</li>\n<li>fourth</li>\n</ol>'));
```

test/widgets/content_test.dart
```diff
@@ -1158,4 +1158,19 @@ void main() {
       check(linkText.textAlign).equals(TextAlign.center);
     });
   });
+
+  group('ListNodeWidget', () {
```
rajveermalviya, line 1162: nit: Let's keep order same as in `widget/content.dart` file, move this test after the `Quotation` test.

rajveermalviya (changes requested):
Thanks for working on this @lakshya1goel! Some small comments below, otherwise looks good.

lib/widgets/content.dart
```diff
@@ -516,7 +515,6 @@ class ListItemWidget extends StatelessWidget {
       textBaseline: localizedTextBaseline(context),
       children: [
         SizedBox(
-          width: 20, // TODO handle long numbers in <ol>, like https://github.com/zulip/zulip/pull/25063
```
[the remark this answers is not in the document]
lakshya1goel: I have a doubt regarding this, can we continue the discussion for this in [CZO Discussion](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/handle.20long.20numbers.20in.20.3Col.3E/with/2087964)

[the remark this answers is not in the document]
rajveermalviya: Commented there.

lakshya1goel:
Thanks for the review @rajveermalviya, pushed the revision. PTAL.

lib/widgets/content.dart
```diff
@@ -493,34 +493,37 @@ class ListNodeWidget extends StatelessWidget {
         case UnorderedListNode(): marker = "•   "; break;
         case OrderedListNode(:final start): marker = "${start + index}. "; break;
       }
-      return ListItemWidget(marker: marker, nodes: item);
+      return buildListItemRow(context: context, marker: marker, nodes: item);
```
rajveermalviya, line 496: nit: Let's just inline this function here.

lib/widgets/content.dart
```diff
@@ -479,49 +479,51 @@ class ListNodeWidget extends StatelessWidget {
...
-        Expanded(child: BlockContentList(nodes: nodes)),
-      ]);
-  }
+TableRow buildListItemRow({
+  required BuildContext context,
+  required String marker,
+  required List<BlockContentNode> nodes,
+}) {
+  return TableRow(
+    children: [
```
rajveermalviya, line 519: nit:
```suggestion
  return TableRow(children: [
```

lib/widgets/content.dart
```diff
@@ -479,49 +479,51 @@ class ListNodeWidget extends StatelessWidget {
...
+      Align(
+        alignment: AlignmentDirectional.topEnd,
+        child: Text(marker, textAlign: TextAlign.right),
+      ),
+      BlockContentList(nodes: nodes),
+    ],
+  );
```
rajveermalviya, line 526: nit: formatting
```suggestion
    ]);
```

lib/widgets/content.dart
```diff
@@ -479,49 +479,51 @@ class ListNodeWidget extends StatelessWidget {
...
+TableRow buildListItemRow({
+  required BuildContext context,
+  required String marker,
+  required List<BlockContentNode> nodes,
+}) {
+  return TableRow(
+    children: [
+      Align(
+        alignment: AlignmentDirectional.topEnd,
+        child: Text(marker, textAlign: TextAlign.right),
```
rajveermalviya, line 522: Here, `alignment: AlignmentDirectional.topEnd` should suffice, with `textAlign: TextAlign.right` there is no space between marker text and item content. 

```suggestion
        child: Text(marker),
```

lib/widgets/content.dart
```diff
@@ -479,49 +479,51 @@ class ListNodeWidget extends StatelessWidget {
...
+      ),
```
rajveermalviya, line 523: nit: formatting
```suggestion
        child: Text(marker, textAlign: TextAlign.right)),
```

lib/widgets/content.dart
```diff
@@ -479,49 +479,51 @@ class ListNodeWidget extends StatelessWidget {
...
+      child: Table(
+        textBaseline: localizedTextBaseline(context),
+        defaultVerticalAlignment: TableCellVerticalAlignment.baseline,
+        columnWidths: const <int, TableColumnWidth>{
+          0: IntrinsicColumnWidth(),
+          1: FlexColumnWidth(),
+        },
+        children: items,
+      ),
+    );
```
rajveermalviya, line 509: nit: formatting

```suggestion
        children: items));
```

lib/widgets/content.dart
```diff
@@ -479,49 +479,51 @@ class ListNodeWidget extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     // TODO(#162): p+ul and p+ol interactions
-    final items = List.generate(node.items.length, (index) {
+    final items = List<TableRow>.generate(node.items.length, (index) {
```
rajveermalviya, line 482: nit: Shouldn't be needed, right?
```suggestion
    final items = List.generate(node.items.length, (index) {
```

rajveermalviya (changes requested):
Thanks for the revision @lakshya1goel! Left some small comments, mostly nits.

lib/widgets/content.dart
lakshya1goel: Yes, right!

lib/widgets/content.dart
lakshya1goel: Yes, right!

lib/widgets/content.dart
```diff
@@ -482,23 +482,34 @@ class ListNodeWidget extends StatelessWidget {
...
         // TODO proper alignment of unordered marker; should be "• ", one space,
         //   but that comes out too close to item; not sure what's fixing that
         //   in a browser
-        case ListStyle.unordered: marker = "•   "; break;
-        // TODO(#59) ordered lists starting not at 1
-        case ListStyle.ordered: marker = "${index+1}. "; break;
+        case UnorderedListNode(): marker = "•   "; break;
+        case OrderedListNode(:final start): marker = "${start + index}. "; break;
       }
-      return ListItemWidget(marker: marker, nodes: item);
```
rajveermalviya, line 497: `ListItemWidget` becomes unused after this right? Let's remove it.

rajveermalviya (approved):
Thanks for the revision @lakshya1goel!
Apart from one comment, this LGTM! Marking for Greg's review.

lib/widgets/content.dart
lakshya1goel: Thanks for the review, pushed the revision!

lib/model/content.dart
```diff
@@ -275,6 +272,22 @@ class ListNode extends BlockContentNode {
   }
 }
 
+class OrderedListNode extends ListNode {
+  const OrderedListNode({this.start = 1, required super.items, super.debugHtmlNode});
```
gnprice, line 276: nit: keep parameters required on these constructors

Any call site of this constructor in the actual app should definitely make an explicit choice of this value. So giving it a default is purely a test-code convenience; and generally conveniences for test code should be kept within the test code. In this case, with only 6 call sites even in the tests (of which 2 are passing this anyway), it's pretty painless to just provide an explicit value at each call site.

You can see this pattern at work if you search for the string " = " when viewing this file in your editor — you'll see there are virtually no matches in the 800 lines or so that are defining these node classes.

gnprice, line 276: nit: follow the patterns of the other classes nearby in the ordering of the named parameters

lib/model/content.dart
```diff
@@ -275,6 +272,22 @@ class ListNode extends BlockContentNode {
...
+  final int start;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(IntProperty('start', start));
+  }
+}
+
+class UnorderedListNode extends ListNode {
```
gnprice, line 287: nit: put this class before OrderedListNode, since it's simpler

lib/model/content.dart
```diff
@@ -253,17 +253,14 @@ class HeadingNode extends BlockInlineContainerNode {
 
 enum ListStyle { ordered, unordered }
```
gnprice, line 254: This enum is no longer really pulling its weight, as it's redundant with the two subclasses; let's cut it out.

The remaining references to it, all within `parseListNode`, can just refer to `element.localName` directly.

lib/model/content.dart
```diff
@@ -1081,7 +1094,16 @@ class _ZulipContentParser {
       items.add(parseImplicitParagraphBlockContentList(item.nodes));
     }
 
-    return ListNode(listStyle!, items, debugHtmlNode: debugHtmlNode);
+    if (listStyle == ListStyle.ordered) {
+      final startAttr = element.attributes['start'];
+      final start = startAttr != null
+        ? int.tryParse(startAttr, radix: 10)
+        : 1;
```
gnprice, line 1101: nit:
```suggestion
      final start = startAttr == null ? 1
        : int.tryParse(startAttr, radix: 10);
```

That puts the simpler case first, getting it out of the way.

test/model/content_test.dart
```diff
@@ -1412,13 +1412,43 @@ void main() {
...
           [
             ParagraphNode(links: null, nodes: [TextNode('a')]),
             ParagraphNode(links: null, nodes: [TextNode('b')]),
           ],
         ]),
       ]);
+
+    testParse('ordered list - large start number (9999)',
+      // "9999. first\n10000. second"
+      '<ol start="9999">\n<li>first</li>\n<li>second</li>\n</ol>', const [
```
gnprice, line 1425: Let's put these new test cases in the new-style pattern that we use for most of the tests in this file: add new ContentExample instances as statics on ContentExample.

For a guide, see the comment at the top of that class.

test/model/content_test.dart
```diff
@@ -1412,13 +1412,43 @@ void main() {
...
+        OrderedListNode(
+          start: 9999,
+          items: [
+            [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('first')])],
+            [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('second')])],
+          ]),
+      ]);
+
+    testParse('ordered list - default start (1)',
+      '<ol>\n<li>first</li>\n<li>second</li>\n</ol>', const [
```
gnprice, line 1435: I think this is covered already by the example at the top of this group.

gnprice (commented):
Thanks @lakshya1goel, and thanks @rajveermalviya for the previous reviews!

Generally this looks good. Comments below, most of them small.

lib/widgets/content.dart
```diff
@@ -482,45 +482,34 @@ class ListNodeWidget extends StatelessWidget {
...
-      children: [
-        SizedBox(
-          width: 20, // TODO handle long numbers in <ol>, like https://github.com/zulip/zulip/pull/25063
-          child: Align(
-            alignment: AlignmentDirectional.topEnd, child: Text(marker))),
-        Expanded(child: BlockContentList(nodes: nodes)),
-      ]);
+      child: Table(
+        textBaseline: localizedTextBaseline(context),
+        defaultVerticalAlignment: TableCellVerticalAlignment.baseline,
```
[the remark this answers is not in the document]
lakshya1goel: I think we didn't have one previously so I have filed [#F1356](https://github.com/zulip/zulip-flutter/issues/1356) for it.

lakshya1goel:
Thanks for the detailed review @gnprice, I have filed issue [#F1356](https://github.com/zulip/zulip-flutter/issues/1356) and mentioned that in commit and PR as well, also pushed the revision resolving all the comments. PTAL!

test/model/content_test.dart
```diff
@@ -1155,6 +1158,32 @@ class ContentExample {
       ], isHeader: false),
     ]),
   ]);
+
+  static const orderedListLargeStart = ContentExample(
```
gnprice, line 1162: nit: in general, try to keep test code in the same order as the code it's testing

Here, that means let's order these examples to match the parsing code — so these list examples go above the spoiler examples, just after the many inline examples.

test/model/content_test.dart
```diff
@@ -1503,16 +1534,20 @@ void main() {
...
   testParse('parse nested lists, quotes, headings, code blocks',
     // "1. > ###### two\n   > * three\n\n      four"
     '<ol>\n<li>\n<blockquote>\n<h6>two</h6>\n<ul>\n<li>three</li>\n'
         '</ul>\n</blockquote>\n<div class="codehilite"><pre><span></span>'
         '<code>four\n</code></pre></div>\n\n</li>\n</ol>', const [
-      ListNode(ListStyle.ordered, [[
+      OrderedListNode(
+        start: 1,
+        items: [[
         QuotationNode([
```
gnprice, line 1548: nit: keep the existing formatting:
```suggestion
      OrderedListNode(start: 1, items: [[
        QuotationNode([
```

In particular the version in the current revision of the PR doesn't have correct indentation: the QuotationNode inside `items` needs to be indented more deeply than the line `items` starts on.

test/model/content_test.dart
```diff
@@ -1155,6 +1158,32 @@ class ContentExample {
...
+    [OrderedListNode(
+        start: 9999,
+        items: [
+          [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('first')])],
+          [ParagraphNode(wasImplicit: true, links: null, nodes: [TextNode('second')])],
+        ])
+    ],
+  );
+
+  static const orderedListCustomStart = ContentExample(
```
gnprice, line 1175: nit: put this "custom" example before "large" — after all, the "large" value is a custom value too, so it's a particular case of being "custom"

test/widgets/content_test.dart
```diff
@@ -230,6 +230,33 @@ void main() {
...
+    testWidgets('ordered list with custom start', (tester) async {
+      await prepareContent(tester, plainContent('<ol start="3">\n<li>third</li>\n<li>fourth</li>\n</ol>'));
+
+      expect(find.text('3. '), findsOneWidget);
+      expect(find.text('4. '), findsOneWidget);
+      expect(find.text('third'), findsOneWidget);
+      expect(find.text('fourth'), findsOneWidget);
+    });
+
+    testWidgets('list uses correct text baseline alignment', (tester) async {
```
gnprice, line 243: Hmm, I guess this is good to check too but it's not the thing I was hoping for in https://github.com/zulip/zulip-flutter/pull/1329#discussion_r1955622801 — it doesn't check that #1356 was (and remains) fixed.

In particular if I make this edit:
```diff
--- lib/widgets/content.dart
+++ lib/widgets/content.dart
@@ -507,7 +507,7 @@ class ListNodeWidget extends StatelessWidget {
         defaultVerticalAlignment: TableCellVerticalAlignment.baseline,
         textBaseline: localizedTextBaseline(context),
         columnWidths: const <int, TableColumnWidth>{
-          0: IntrinsicColumnWidth(),
+          0: FixedColumnWidth(20),
           1: FlexColumnWidth(),
         },
```

then that should reintroduce #1356 — it'd make the new code behave pretty much just like the old code before your fix — but there isn't currently a test that would detect that and fail.

Can you find a way to write a test that checks that #1356 is fixed?

test/widgets/content_test.dart
```diff
@@ -230,6 +230,33 @@ void main() {
...
+      check(table.defaultVerticalAlignment).equals(TableCellVerticalAlignment.baseline);
+      check(table.textBaseline).equals(localizedTextBaseline(tester.element(find.byType(Table))));
+
+      await prepareContent(tester, Directionality(
+        textDirection: TextDirection.rtl,
+        child: plainContent(ContentExample.orderedListLargeStart.html)));
+
+      final tableRtl = tester.widget<Table>(find.byType(Table));
+      check(tableRtl.defaultVerticalAlignment).equals(TableCellVerticalAlignment.baseline);
+      check(tableRtl.textBaseline).equals(localizedTextBaseline(tester.element(find.byType(Table))));
```
gnprice, line 256: Does the baseline logic interact with the text direction?

I think it doesn't, and it doesn't seem likely that it would. So let's leave this part of the test out, as being redundant.

gnprice (commented):
Thanks for the revision! Mostly small comments now.

lakshya1goel:
Pushed the revision @gnprice, PTAL. Thanks!

test/widgets/content_test.dart
```diff
@@ -230,6 +230,49 @@ void main() {
...
+      final table = tester.widget<Table>(find.byType(Table));
+      check(table.defaultVerticalAlignment).equals(TableCellVerticalAlignment.baseline);
+      check(table.textBaseline).equals(localizedTextBaseline(tester.element(find.byType(Table))));
+    });
+
+    testWidgets('long ordered list markers render completely and are right-aligned', (tester) async {
+      await prepareContent(tester, plainContent(ContentExample.orderedListLargeStart.html));
+
+      final markerTexts = tester.widgetList<Text>(
+        find.descendant(of: find.byType(Align), matching: find.byType(Text)));
```
gnprice, line 255: Looking for `Align` here makes this tied into some fairly internal details of how ListNodeWidget is implemented.

Instead, let's use `find.textContaining('9999')`. That should find the same `Text` widget in a way that's more focused on what the user sees, rather than how the implementation works. See:

(It does make this test dependent on the details of the test data it's using, from `ContentExample.orderedListLargeStart`. That's OK — that test data exists basically for the sake of this test and the related parsing test.)

gnprice: Also we only need to check one marker, not both.

test/widgets/content_test.dart
```diff
@@ -230,6 +230,49 @@ void main() {
...
+        final renderParagraph = tester.renderObject(find.text(text.data!)) as RenderParagraph;
+        final textHeight = renderParagraph.size.height;
+        final lineHeight = renderParagraph.text.style!.height! * renderParagraph.text.style!.fontSize!;
+        check(textHeight).equals(lineHeight);
+
+        final textWidth = renderParagraph.size.width;
+        final renderBox = tester.renderObject(find.ancestor(
+            of: find.text(text.data!),
+            matching: find.byType(Align))) as RenderBox;
+        check(renderBox.size.width >= textWidth).isTrue();
```
gnprice, line 267: I think this check can't really fail — it's checking that the parent is at least as big as the child, and that's always true unless one uses some fairly special widgets that aren't likely to get involved here. We can leave it out.

test/widgets/content_test.dart
```diff
@@ -230,6 +230,49 @@ void main() {
...
+
+        final align = tester.widget<Align>(
+          find.ancestor(of: find.text(text.data!), matching: find.byType(Align)));
+        check(align.alignment).equals(AlignmentDirectional.topEnd);
```
gnprice, line 271: Let's adapt this so that it's in terms of what the user sees, rather than how the implementation code works. (Per https://zulip.readthedocs.io/en/latest/testing/philosophy.html#integration-testing-or-unit-testing again.)

Concretely: the content in this test has two markers. They aren't going to be the same width. So if the alignment is correct, they'll have their right edges at the same x-coordinate, and if the alignment is wrong they'll almost surely have their right edges at different x-coordinates. So to write a test, get the locations of the two marker `Text`s, confirm the widths are different, and then check the right edges match.

test/widgets/content_test.dart
```diff
@@ -230,6 +230,49 @@ void main() {
...
+
+    testWidgets('list uses correct text baseline alignment', (tester) async {
+      await prepareContent(tester, plainContent(ContentExample.orderedListLargeStart.html));
+
+      final table = tester.widget<Table>(find.byType(Table));
+      check(table.defaultVerticalAlignment).equals(TableCellVerticalAlignment.baseline);
+      check(table.textBaseline).equals(localizedTextBaseline(tester.element(find.byType(Table))));
+    });
+
+    testWidgets('long ordered list markers render completely and are right-aligned', (tester) async {
```
gnprice, line 251: nit: this should be two test cases (two `testWidgets` calls) — one for having enough room, and one for being end-aligned

(also it's not right-aligned in general — it's *end*-aligned, which means the left side when using an RTL language, like Persian or Arabic or Hebrew)

(it's fine not to add a test for the RTL case, just adjust this test name so it's accurate)

lib/model/content.dart
```diff
@@ -251,21 +251,11 @@ class HeadingNode extends BlockInlineContainerNode {
   }
 }
 
-enum ListStyle { ordered, unordered }
+sealed class ListNode extends BlockContentNode {
+  const ListNode({required this.items, super.debugHtmlNode});
```
gnprice, line 255: Ah I missed this earlier — let's keep the constructor calls a bit simpler:
```suggestion
  const ListNode(this.items, {super.debugHtmlNode});
```

(Just like in the existing version, and like in some other classes in this file: QuotationNode, CodeBlockNode, TextNode, and others where it's clear what a positional argument should mean.)

gnprice (commented):

lakshya1goel:
Thanks for the detailed review, I have pushed the revision, PTAL.

gnprice:
(I've just edited the PR description so that it's marked as fixing #1356 too — the GitHub "fixes" syntax doesn't understand "and", and needs the word "fixes" to be right next to the issue number.)

test/widgets/content_test.dart
```diff
@@ -230,6 +230,48 @@ void main() {
...
+      await prepareContent(tester, plainContent(ContentExample.orderedListLargeStart.html));
+
+      final marker9999 = tester.renderObject(find.textContaining('9999.')) as RenderParagraph;
+      final marker10000 = tester.renderObject(find.textContaining('10000.')) as RenderParagraph;
+      check(marker9999.size.width != marker10000.size.width).isTrue();
+
+      final marker9999Pos = marker9999.localToGlobal(Offset.zero);
+      final marker10000Pos = marker10000.localToGlobal(Offset.zero);
+      check(marker9999Pos.dx + marker9999.size.width)
+        .equals(marker10000Pos.dx + marker10000.size.width);
```
gnprice, line 271: Cool, this test is accurate, but can be simplified using `getRect`:
```dart
      final marker9999 = tester.getRect(find.textContaining('9999.'));
```
then `marker9999.width` and `marker9999.right`.

gnprice: (If you dig into the implementation of `getRect`, you'll see it's doing something similar to this test's logic.)

test/model/content_test.dart
```diff
@@ -306,13 +326,13 @@ class ContentExample {
...
-            TextNode('hello'),
+      header: [OrderedListNode([[
+          UnorderedListNode([
+            [HeadingNode(level: HeadingLevel.h2, links: null, nodes: [
+              TextNode('hello'),
+            ])]
           ])]
-        ])],
-      ])],
+        ], start: 1)],
```
gnprice, line 335: nit: keep `start: 1` as the first argument, before the positional `items` argument

Dart has the handy feature of allowing named and positional arguments to be interleaved in any order, including named first. That's useful to do in a case like this where there's a named argument that's short and fits easily on the first line and then a longer multi-line argument that's positional.

gnprice (commented):
Thanks for the revision! Just a couple of small comments now.

lakshya1goel:
Pushed the revision, Please have a look @gnprice . Thanks!

gnprice:
Thanks! Looks good; merging.

I made a few tweaks in the first commit along the lines of https://github.com/zulip/zulip-flutter/pull/1329#discussion_r1966349901, for two-space indentation and to avoid other formatting changes. Also added one commit on top which I recommend reading:
e0ee70ed0 content test [nfc]: Use checks-extensions in list-marker tests; explain a bit more

since it points to an important bit of test-writing style which I'd missed in previous reviews.

In particular, a change like this:
```diff
-      check(marker9999.right).equals(marker10000.right);
+      check(marker9999).right.equals(marker10000.right);
```
means that if that check fails, the resulting error will be more informative because it will have the whole Rect value `marker9999` rather than just the one piece of it that is `marker9999.right`. To see the difference, try sabotaging the test to fail and then trying both versions, with `.right` inside and outside the parens, to see the output.

PR #1331, opened 2025-02-05
Attempt to fix JVM high memory usage in CI

PIG208:
Will trigger the CI with this.


gnprice:
Thanks for all the investigation!

Pushed a revision that adjusts the commit message in a few places, and that also makes the diff do the same number the commit message says :slightly_smiling_face: (it was actually reverting the previous bump).

I'll let CI run on that to be sure I haven't missed something, and then merge.

PR #1332, opened 2025-02-05
i18n: Add and update TODO markers after a sweep

gnprice:
Toward #277.

Background: https://github.com/zulip/zulip-flutter/issues/277#issuecomment-2638297174


chrisbobbe:
LGTM, thanks! ~~Merging~~ I'll let Zixuan review as I see he's assigned (guess I skipped looking at that 🙂 field before reviewing)

As noted in #277, there's more work before we can declare that issue complete: https://github.com/zulip/zulip-flutter/issues/277#issuecomment-2638297174

PIG208:
Thanks for catching these! Merging.

PR #1305, opened 2025-01-24
msglist: Use distinct background colors for DM messages.

E-m-i-n-e-n-c-e:
This PR modifies DmMessage and its DateSeparator to use a different background color from StreamMessage. This matches the web app's behavior where DM messages have a distinct background color.

Added new theme colors to MessageListTheme(These colors match the web app version)

* Light theme: hsl(45deg 20% 96%)

* Dark theme: hsl(46deg 7% 16%)

Fixes #681.

**Screenshots**:
 
Image

Image

Image

Image

PR #1336, opened 2025-02-07
msglist: Fix DM header color lerp bug.

E-m-i-n-e-n-c-e:
Fixed a bug in MessageListTheme.lerp() where dmRecipientHeaderBg was using streamMessageBgDefault instead of dmRecipientHeaderBg for interpolation.


chrisbobbe:
(Edited the description to remove "Part of #1305", because this is an independent bugfix that's not part of fixing that PR's issue #681. The connection to #1305 is just that that's when you found the bug. 🙂)

chrisbobbe:

gnprice:
Thanks for the fix! Looks good — merging.

PR #1340, opened 2025-02-08
store: Ensure sole ownership of MessageListView

PIG208:
PerAccountStore shouldn't be an owner of the MessageListView elements.

Its relationship to MessageListView is similar to that of AutocompleteViewManager to MentionAutocompleteView (#645).

With two owners, the MessageListView can be disposed twice:

1. before the frame is rendered, `removeAccount` disposes the `PerAccountStoreWidget`, which disposes the `MessageListView`; `_MessageListState` is not yet disposed;

2. during build, because `store` is set to `null`, `PerAccountStoreWidget` gets rebuilt. `_MessageListState`, a descendent of it, is no longer in the render tree;

3. during finalization, `_MessageListState` tries to dispose the `MessageListView`.

This removes regression tests added for #810, because `MessageStoreImpl.dispose` no longer exists.  `MessageListView` does not get disposed unless there is a `_MessageListState` owner.

For reproduction, I created a testing branch with #1183 stacked on top of this, but has the fix reverted:

- Checkout https://github.com/PIG208/zulip-flutter/pull/1.
- Go to any `MessageListPage`
- Invalidate the API key (by regenerating it on web, for example)

You should see the following error:

```
======== Exception caught by widgets library =======================================================
The following assertion was thrown while finalizing the widget tree:
'package:zulip/model/message.dart': Failed assertion: line 54 pos 12: 'removed': is not true.

#2      MessageStoreImpl.unregisterMessageList (package:zulip/model/message.dart:54:12)
#3      PerAccountStore.unregisterMessageList (package:zulip/model/store.dart:540:15)
#4      MessageListView.dispose (package:zulip/model/message_list.dart:427:11)
#5      _MessageListState.dispose (package:zulip/widgets/message_list.dart:491:12)
#6      StatefulElement.unmount (package:flutter/src/widgets/framework.dart:5940:11)
#7      _InactiveElements._unmount (package:flutter/src/widgets/framework.dart:2085:13)
#8      _InactiveElements._unmount.<anonymous closure> (package:flutter/src/widgets/framework.dart:2083:7)
#9      ComponentElement.visitChildren (package:flutter/src/widgets/framework.dart:5781:14)
(elided 2 frames from class _AssertionError)
====================================================================================================
```

[CZO discussion](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.60MentionAutocompleteView.2Edispose.60/near/2083074)


lib/model/actions.dart
```diff
@@ -0,0 +1,36 @@
...
+
+Future<void> unregisterToken(GlobalStore globalStore, int accountId) async {
+  final account = globalStore.getAccount(accountId);
+  if (account == null) return; // TODO(log)
+
+  // TODO(#322) use actual acked push token; until #322, this is just null.
+  final token = account.ackedPushToken
+      // Try the current token as a fallback; maybe the server has registered
+      // it and we just haven't recorded that fact in the client.
+      ?? NotificationService.instance.token.value;
```
chrisbobbe, line 25: nit: indentation

chrisbobbe (commented):
Thanks! LGTM, with an indentation nit below. Another small comment: since there's no issue to refer to (with a "Fixes" line), could the commit message either link to the discussion or say something like "see discussion linked in a code comment"?

lib/model/message.dart
```diff
@@ -60,14 +60,12 @@ class MessageStoreImpl with MessageStore {
...
 
-  void dispose() {
-    // When a MessageListView is disposed, it removes itself from the Set
-    // `MessageStoreImpl._messageListViews`. Instead of iterating on that Set,
-    // iterate on a copy, to avoid concurrent modifications.
-    for (final view in _messageListViews.toList()) {
-      view.dispose();
-    }
-  }
+  // No `dispose` method, because there's nothing for it to do.
```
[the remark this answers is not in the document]
gnprice: 2. during build, because `store` is set to `null`,
      `PerAccountStoreWidget` gets rebuilt. `_MessageListState`,

This isn't quite right — if we didn't have this line (which I believe is what you're referring to as "`store` is set to `null`"):
```dart
      _setStore(null);
```

then that element would still get rebuilt. That's because that line is within that state's `didChangeDependencies`, and the base method's doc says:
```dart
  /// Subclasses rarely override this method because the framework always
  /// calls [build] after a dependency changes. Some subclasses do override
```

So if that method was getting called, then `build` was going to get called next anyway.

It's a good small exercise to work out exactly what does cause `build` to get called in this scenario.

gnprice, line 63: `PerAccountStoreWidget` gets rebuilt. `_MessageListState`,
      a descendent of it, is no longer in the render tree;

nit: element tree, not render tree — the render tree is a different thing:
https://main-api.flutter.dev/flutter/rendering/RenderObject-class.html

and a `_MessageListState` doesn't directly correspond to anything in the render tree (i.e. to any render object).

gnprice, line 63: nit:

    PerAccountStore shouldn't be an owner of the MessageListView elements.

A MessageListView isn't an "element"; that term has a specific meaning in Flutter:
https://main-api.flutter.dev/flutter/widgets/Element-class.html
and is also used (following math terminology) for an item in a collection data structure, like a List, but it's not a general synonym for "object" or "value".

Instead of "elements", "objects" would be accurate.

lib/model/message.dart
```diff
@@ -60,14 +60,12 @@ class MessageStoreImpl with MessageStore {
...
+  // The [MessageListView]s are owned by (i.e., they get [dispose]d by)
+  // the [_MessageListState], including in the case where the [PerAccountStore]
+  // is replaced.  Discussion:
+  //   https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.60MentionAutocompleteView.2Edispose.60/near/2083074
+  // void dispose() { … }
```
gnprice, line 68: This makes me a bit nervous that it would be easy for us to spring a leak where MessageListView objects stick around. They can be memory-hungry, I think, so that's good to avoid.

That risk is highlighted by the fact that you had to add a `model?.dispose();` line elsewhere in this commit — we had such a line when the state gets disposed, but had omitted to add one for when it gets a new store.

How about we instead have a `dispose` method here but it just asserts that its set of MessageListView objects to manage is empty? I believe that's implied by the [logic](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.60MentionAutocompleteView.2Edispose.60/near/1791292) we're relying on in order to justify not needing to dispose them here:
> […] But it also is tied to a given `PerAccountStore`, and so shouldn't outlive that store. So it should get disposed when the store goes away, too.
>
> I think that might already be getting accomplished by the `dispose` calls in `_ComposeAutocompleteState`. If the store is getting disposed of, then it shouldn't be the case that a `_ComposeAutocompleteState` sticks around in the tree under that store — any such `_ComposeAutocompleteState` should either be told of a new store with `onNewStore`, or be leaving the tree entirely and so have its own `dispose` called. Both of those call `dispose` on the view-model.

test/model/store_test.dart
```diff
@@ -824,25 +822,6 @@ void main() {
       checkReload(prepareHandleEventError);
     });
 
-    test('expired queue disposes registered MessageListView instances', () => awaitFakeAsync((async) async {
```
gnprice, line 827: This scenario is still highly relevant — see my discussion above of the potential for a leak.

So we should still test it. The change to make is just that the test now needs appropriate widgets as part of the setup, because the MessageList widget is now part of what makes this behave correctly.

lib/widgets/message_list.dart
```diff
@@ -483,6 +483,7 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
 
   @override
   void onNewStore() { // TODO(#464) try to keep using old model until new one gets messages
+    model?.dispose();
```
gnprice, line 486: If this line is removed, it looks like all tests still pass. There should be a test that would notice if we dropped this.

(I think this is exactly what the test discussed in my previous comment would do.)

gnprice (commented):
Thanks for spotting and fixing this!

I believe the implementation logic here is all correct. But I'd like to see an assert, and one of these pieces still needs a test. See also other comments below.

PIG208:
Updated the PR to address the comments. Thanks for the review! The insights on the Flutter's widgets/elements are really helpful.

lib/model/message.dart
gnprice: Yeah. As a nit, let's keep the method, though — it'll just have an empty body containing that TODO. That provides a bit more framing to make clear exactly what the TODO calls for, and it also means we preserve the record of what's meant to call this method.

lib/model/message.dart
```diff
@@ -60,14 +60,17 @@ class MessageStoreImpl with MessageStore {
...
 
-  void dispose() {
-    // When a MessageListView is disposed, it removes itself from the Set
-    // `MessageStoreImpl._messageListViews`. Instead of iterating on that Set,
-    // iterate on a copy, to avoid concurrent modifications.
-    for (final view in _messageListViews.toList()) {
-      view.dispose();
-    }
-  }
+  // No `dispose` method, because there's nothing for it to do.
```
gnprice, line 63: nit:

    2. Being dependent on `GlobalStore`, `PerAccountStoreWidget` is rebuilt.
       This time, the StatefulElement corresponding to the `MessageList`
       widget, is no longer in the element tree because `PerAccountStoreWidget`
       cannot find the account and builds a placeholder instead.

It'd be good to call out explicitly at the start of this scenario (before the numbered steps) that you're talking about a case where `removeAccount` is called, like if the user logs out. I read this step a couple of times and initially thought it was mistaken, because the main scenario that comes to my mind for a store getting disposed is when it's replaced by a new store.

lib/model/message.dart
```diff
@@ -60,14 +60,17 @@ class MessageStoreImpl with MessageStore {
...
+  // Not disposing the [MessageListView]s here, because they are owned by
+  // (i.e., they get [dispose]d by) the [_MessageListState], including in the
+  // case where the [PerAccountStore] is replaced.
+  //
+  // TODO: Add assertions that the [MessageListView]s are indeed disposed, by
+  //   first ensuring that [PerAccountStore] is only disposed after those with
+  //   references to it are disposed, then reinstating this `dispose` method.
+  //   Discussion:
+  //     https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.60MentionAutocompleteView.2Edispose.60/near/2083074
```
gnprice, line 72: nit: let's have this point here:
https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/MessageListView.20lifecycle/near/2086893

since that's where the discussion relevant to this particular TODO begins.

test/widgets/message_list_test.dart
```diff
@@ -130,6 +134,48 @@ void main() {
       final state = MessageListPage.ancestorOf(tester.element(find.text("a message")));
       check(state.composeBoxController).isNull();
     });
+
+    testWidgets('dispose MessageListView when event queue expired', (tester) async {
+      // Regression test for: https://github.com/zulip/zulip-flutter/issues/810
```
gnprice, line 139: This isn't a regression test for #810 anymore — to do that, there'd have to be more than one MessageListView.

That's fine, though. We don't really need a regression test for that, now that the code that the bug was in (the loop through view-models in MessageStoreImpl.dispose) is gone. Instead this is just testing the thing the test's name says it is.

```suggestion
```

gnprice (commented):
Thanks for tracking down these subtle interactions!

This all looks good now, modulo a few small comments.

PIG208:
Thanks! Updated the PR.

lib/model/message.dart
```diff
@@ -61,12 +61,20 @@ class MessageStoreImpl with MessageStore {
...
+    // TODO: Add assertions that the [MessageListView]s are indeed disposed, by
+    //   first ensuring that [PerAccountStore] is only disposed after those with
+    //   references to it are disposed, then reinstating this `dispose` method.
+    //
+    //   We can't add the assertions as-is because the sequence of events
+    //   guarantees that `PerAccountStore` is disposed (when that happens,
+    //   [GlobalStore] notifies its listeners, causing widgets dependent on the
+    //   [InheritedNotifier] to rebuild in the next frame) before the owner's
+    //   `dispose` or `onNewStore` is called.  Discussion:
+    //     https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.60MentionAutocompleteView.2Edispose.60/near/2083074
```
gnprice, line 77: bump https://github.com/zulip/zulip-flutter/pull/1340#discussion_r1951802928 :slightly_smiling_face:

PIG208: Hmm, I read that as "have this point, here" and incorporated this new paragraph. I think you are meant to have the link to "point here", right?

PIG208:
Updated the references to the CZO discussion with the link from https://github.com/zulip/zulip-flutter/pull/1340#discussion_r1951802928.

lib/model/message.dart
gnprice: I see. Yeah, I meant "have this link point to here", i.e. to the following URL.

gnprice:
Thanks! Looks good; merging.

PR #1341, opened 2025-02-09
solve ndk issue 

DevloperMahak:
Add ndk correct version in app/build.gradle, add a path in local.properties and in android/build.gradle and also correct a structure in android/build.gradle.


chrisbobbe:
Is there an issue in the tracker that this is meant to solve? Please add a link in the PR description if so.

DevloperMahak:
> Is there an issue in the tracker that this is meant to solve? Please mention it in the PR description if so.
No, there is no existing issue in the tracker related to this. I encountered this issue while setting up the project locally, as the NDK version was not specified in app/build.gradle. The project was expected to specify the NDK version but does not, which may indicate a missing dependency issue.

gnprice:
I don't think there's anything we need to change here. The build works fine for many other people, and in our CI.

If you're having trouble setting up to develop the app, ask for help in chat and we can help debug. See the README for details. When asking for help, be sure to describe what you're doing and exactly what symptom you're seeing.

That's also how we would determine if there is something we should change in the setup — we'd need to see specifically what's going wrong for you so we can determine the right fix.

PR #1342, opened 2025-02-11
Get compose box hint texts ready for general chat

PIG208:
Prepares for:
- #1297

| mandatory topics | non-mandatory topics |
| - | - |
| pre-334, mandatory | Screenshot_20250211_143906 |

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -348,12 +348,11 @@
   "@composeBoxSelfDmContentHint": {
     "description": "Hint text for content input when sending a message to yourself."
   },
-  "composeBoxChannelContentHint": "Message #{channel} > {topic}",
+  "composeBoxChannelContentHint": "Message {destination}",
   "@composeBoxChannelContentHint": {
-    "description": "Hint text for content input when sending a message to a channel",
+    "description": "Hint text for content input when sending a message to a channel.",
     "placeholders": {
-      "channel": {"type": "String", "example": "channel name"},
-      "topic": {"type": "String", "example": "topic name"}
+      "destination": {"type": "String", "example": "#channel name > topic name"}
     }
   },
   "composeBoxSendTooltip": "Send",
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -160,10 +160,17 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
     return trimmed.isEmpty ? kNoTopicTopic : trimmed;
   }
 
+  /// Whether [textNormalized] would fail a mandatory-topics check
+  /// (see [mandatory]).
+  ///
+  /// The term "Vacuous" draws distinction from [String.isEmpty], in the sense
+  /// that certain strings are not empty but also indicate the absence of a topic.
+  bool get isTopicVacuous => textNormalized == kNoTopicTopic;
+
   @override
   List<TopicValidationError> _computeValidationErrors() {
     return [
-      if (mandatory && textNormalized == kNoTopicTopic)
+      if (mandatory && isTopicVacuous)
         TopicValidationError.mandatoryButEmpty,
 
       if (
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -577,13 +577,13 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     final zulipLocalizations = ZulipLocalizations.of(context);
     final streamName = store.streams[widget.narrow.streamId]?.name
       ?? zulipLocalizations.unknownChannelName;
+    final topic = TopicName(widget.controller.topic.textNormalized);
     return _ContentInput(
       narrow: widget.narrow,
-      destination: TopicNarrow(widget.narrow.streamId,
-        TopicName(widget.controller.topic.textNormalized)),
+      destination: TopicNarrow(widget.narrow.streamId, topic),
       controller: widget.controller,
-      hintText: zulipLocalizations.composeBoxChannelContentHint(streamName,
-        widget.controller.topic.textNormalized));
+      hintText: zulipLocalizations.composeBoxChannelContentHint(
+        streamName, topic.displayName));
   }
 }
 
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -590,7 +590,11 @@ class _StreamContentInputState extends State<_StreamContentInput> {
       destination: TopicNarrow(widget.narrow.streamId, topic),
       controller: widget.controller,
       hintText: zulipLocalizations.composeBoxChannelContentHint(
-        streamName, topic.displayName));
+        // No i18n of this use of "#" and ">" string; those are part of how
+        // Zulip expresses channels and topics, not any normal English punctuation,
+        // so don't make sense to translate. See:
+        //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
+        '#$streamName > ${topic.displayName}'));
   }
 }
 
@@ -649,7 +653,11 @@ class _FixedDestinationContentInput extends StatelessWidget {
         final streamName = store.streams[streamId]?.name
           ?? zulipLocalizations.unknownChannelName;
         return zulipLocalizations.composeBoxChannelContentHint(
-          streamName, topic.displayName);
+          // No i18n of this use of "#" and ">" string; those are part of how
+          // Zulip expresses channels and topics, not any normal English punctuation,
+          // so don't make sense to translate. See:
+          //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
+          '#$streamName > ${topic.displayName}');
 
       case DmNarrow(otherRecipientIds: []): // The self-1:1 thread.
         return zulipLocalizations.composeBoxSelfDmContentHint;
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -78,14 +78,17 @@ void main() {
     controller = tester.state<ComposeBoxState>(find.byType(ComposeBox)).controller;
   }
 
+  /// A [Finder] for the topic input.
+  ///
+  /// To enter some text, use [enterTopic].
+  final topicInputFinder = find.byWidgetPredicate(
+    (widget) => widget is TextField && widget.controller is ComposeTopicController);
+
   /// Set the topic input's text to [topic], using [WidgetTester.enterText].
   Future<void> enterTopic(WidgetTester tester, {
     required ChannelNarrow narrow,
     required String topic,
   }) async {
-    final topicInputFinder = find.byWidgetPredicate(
-      (widget) => widget is TextField && widget.controller is ComposeTopicController);
-
     connection.prepare(body:
       jsonEncode(GetStreamTopicsResult(topics: [eg.getStreamTopicsEntry()]).toJson()));
     await tester.enterText(topicInputFinder, topic);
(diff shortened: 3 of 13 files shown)


test/widgets/compose_box_test.dart
```diff
@@ -318,6 +321,81 @@ void main() {
...
+          contentHintText: 'Message #${channel.name} > (no topic)');
+      });
+
+      testWidgets('with non-empty topic', (tester) async {
+        final narrow = ChannelNarrow(channel.streamId);
+        await prepare(tester, narrow: narrow);
+        await enterTopic(tester, narrow: narrow, topic: 'new topic');
+        await tester.pump();
+        checkComposeBoxHintTexts(tester,
+          topicHintText: 'Topic',
```
chrisbobbe, line 368: Huh, I'm surprised this test passes. When using the app, when the topic input is empty, I see the hint text "Topic" there—but I don't see (or expect to see) hint text after I've typed something; I just see the text I've typed. Do you know why this is passing?

test/widgets/compose_box_test.dart
```diff
@@ -366,6 +370,25 @@ void main() {
       });
     });
 
+    testWidgets('to ChannelNarrow, mandatory topics with empty topic', (tester) async {
```
chrisbobbe, line 373:
```
compose: Use alternative placeholder if topics are mandatory and missing
```

We have this pair of tests:

- group 'to ChannelNarrow, topics not mandatory'
  - test 'with empty topic'
  - test 'with non-empty topic'

And we have this pair of tests:

- test 'to ChannelNarrow, mandatory topics with empty topic'
- test 'to ChannelNarrow, mandatory topics with non-empty topic'

It seems cleaner if, for both pairs, it either uses `group` or doesn't, with consistent naming.

chrisbobbe (commented):
Thanks! Small comments below.

test/widgets/compose_box_test.dart
PIG208: This helper checks `TextField.decoration.hintText` directly, but does not test if `hintText` is invisible when the input is non-empty (it does check if the input exists or not at all). We could figure out a way to construct the finder that distinguishes the content and the hint text, but I felt that upstream covers this aspect already. Maybe we can add a dartdoc for this helper?

PIG208:

test/widgets/compose_box_test.dart
PIG208: Ah! I was mistaken. It doesn't check for the controller but `find.text` still ignores that the hint text can be invisible.

chrisbobbe (commented):

test/widgets/compose_box_test.dart
```diff
@@ -318,6 +321,86 @@ void main() {
...
+        narrow: narrow,
+        otherUsers: [eg.otherUser, eg.thirdUser],
+        streams: [channel]);
+    }
+
+    /// Check if the topic and content inputs use the expected hint texts.
+    ///
+    /// If `topicHintText` is `null`, check that the topic input is not present.
+    ///
+    /// This does not check if the hint texts are present but invisible.
```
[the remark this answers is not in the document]
PIG208: This comes from [`_InputDecoratorState.build`](https://github.com/flutter/flutter/blob/5365993ff52ed863d8eb474517addf2da7128bcc/packages/flutter/lib/src/material/input_decorator.dart#L2270-L2301):

```dart
    final TextStyle hintStyle = _getInlineHintStyle(themeData, defaults);
    final String? hintText = decoration.hintText;
    final bool maintainHintSize = decoration.maintainHintSize;
    Widget? hint;
    if (decoration.hint != null || hintText != null) {
      final Widget hintWidget =
          decoration.hint ??
          Text(
            hintText!,
            style: hintStyle,
            textDirection: decoration.hintTextDirection,
            overflow:
                hintStyle.overflow ??
                (decoration.hintMaxLines == null ? null : TextOverflow.ellipsis),
            textAlign: textAlign,
            maxLines: decoration.hintMaxLines,
          );
      final bool showHint = isEmpty && !_hasInlineLabel;
      hint =
          maintainHintSize
              ? AnimatedOpacity(
                opacity: showHint ? 1.0 : 0.0,
                duration: decoration.hintFadeDuration ?? _kHintFadeTransitionDuration,
                curve: _kTransitionCurve,
                child: hintWidget,
              )
              : AnimatedSwitcher(
                duration: decoration.hintFadeDuration ?? _kHintFadeTransitionDuration,
                transitionBuilder: _buildTransition,
                child: showHint ? hintWidget : const SizedBox.shrink(),
              );
    }
```

So for our test to work as-is it would require `maintainHintSize` to be `true` (the current default), and that they continue to just change the opacity when hiding it. It appears to be an implementation detail.

We could also change the behavior of the test, by making sure that the hint text is visible. I tried Using `.hitTestable` when appropriate, but it doesn't work (the check always fails). I think it has to do with how the hint text is rendered.

Alternatively, we could switch to checking `.decoration` on the `TextField`, so that we don't have to rely on this implementation detail. Will be pushing this in the next revision.

chrisbobbe:
Thanks! LGTM, marking for Greg's review.

lib/widgets/compose_box.dart
```diff
@@ -160,10 +160,17 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
     return trimmed.isEmpty ? kNoTopicTopic : trimmed;
   }
 
+  /// Whether [textNormalized] would fail a mandatory-topics check
+  /// (see [mandatory]).
+  ///
+  /// The term "Vacuous" draws distinction from [String.isEmpty], in the sense
+  /// that certain strings are empty but also indicate the absence of a topic.
+  bool get isTopicVacuous => textNormalized == kNoTopicTopic;
```
gnprice, line 168:
```suggestion
  /// The term "Vacuous" draws distinction from [String.isEmpty], in the sense
  /// that certain strings are not empty but also indicate the absence of a topic.
  bool get isTopicVacuous => textNormalized == kNoTopicTopic;
```

?  Or am I misunderstanding what you mean?

lib/widgets/compose_box.dart
```diff
@@ -580,11 +587,22 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     final zulipLocalizations = ZulipLocalizations.of(context);
     final streamName = store.streams[widget.narrow.streamId]?.name
       ?? zulipLocalizations.unknownChannelName;
+    final topic = TopicName(_topicTextNormalized);
+    final String? topicDisplayName;
+    if (store.realmMandatoryTopics && widget.controller.topic.isTopicVacuous) {
+      topicDisplayName = null;
+    } else {
+      topicDisplayName = topic.displayName;
+    }
```
gnprice, line 596: compose: Use alternative placeholder if topics are mandatory and missing
    
    Previously, the placeholder reads "#channel name > (no topic)" even if
    topics are mandatory for the realm.

This change in behavior seems fine, but not especially helpful. It's fine for the hint to echo the thing the user has asked to do, even if that's something they won't ultimately be able to do.

Does this set up this code in a way that will be convenient in a later PR for "general chat"? That'd be a good reason for it.

assets/l10n/app_en.arb
```diff
@@ -348,12 +348,18 @@
   "@composeBoxSelfDmContentHint": {
     "description": "Hint text for content input when sending a message to yourself."
   },
-  "composeBoxChannelContentHint": "Message #{channel} > {topic}",
+  "composeBoxChannelContentHint": "Message #{channel}",
```
gnprice, line 351: The reasoning in the last commit:

    compose: Avoid translating Zulip message destinations
    
    The '#channel > topic' style strings are not supposed to be translated
    into different languages as they are Zulip's language of expressing the
    desintation, not something bound to the English language.
    
    See also:
      https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585

applies equally well to this string, right?

So maybe let's keep this as a single translated string, and just pass either "#general" or "#general > lunch" to it from app code.

lib/widgets/compose_box.dart
```diff
@@ -580,11 +587,22 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     final zulipLocalizations = ZulipLocalizations.of(context);
     final streamName = store.streams[widget.narrow.streamId]?.name
       ?? zulipLocalizations.unknownChannelName;
+    final topic = TopicName(_topicTextNormalized);
+    final String? topicDisplayName;
+    if (store.realmMandatoryTopics && widget.controller.topic.isTopicVacuous) {
```
gnprice, line 592: It looks like at least as of #1297, this line and the definition of `isTopicVacuous` are the same as in this PR. Is that what you're expecting in your current plan for how "general chat" will work in this file?

If that's the case, I think the `isTopicVacuous` getter isn't really doing much for us — it'd be clearer to just inline it here as well as its one other use site.

We discussed a getter along those lines last week, and the idea then was that in a "general chat" world the `isVacuous` getter would gain additional logic related to "general chat". If there's no such logic to put there, it's not needed.

gnprice (commented):
Thanks both!

Have a few minutes at the moment, so I've read all but the tests; comments below.

lib/widgets/compose_box.dart
PIG208: We start adding logic to this in #1364 (https://github.com/zulip/zulip-flutter/pull/1364/commits/6886bc59), when implementing send for empty topics.

lib/widgets/compose_box.dart
PIG208: It's mainly to later prevent hinting the existence of "general chat" at all, if topics are required for the realm.

This is also meant to match web's behavior. On web, "general chat" never shows up in the hint text if topics are required.

gnprice: > It's mainly to later prevent hinting the existence of "general chat" at all, if topics are required for the realm.

Cool, that's certainly good. Where can I read the code that builds on this to do that?

PIG208: The change is on a draft branch, #1365 (https://github.com/zulip/zulip-flutter/pull/1365/commits/51772ec51fbacabd3a431d636c83a58a2ca13adf), where we extend the checks to initialize `topicDisplayName`, but some more work is needed to make this display name stateful.

lib/widgets/compose_box.dart
```diff
@@ -580,11 +587,22 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+      topicDisplayName = null;
+    } else {
+      topicDisplayName = topic.displayName;
```
gnprice, line 595: The link in https://github.com/zulip/zulip-flutter/pull/1342#discussion_r1964053113 is helpful, thanks.

Can you write down what `topicDisplayName` being null is intended to mean? After staring at that later code for a few minutes, I think the answer is: it's null when the user hasn't yet picked a valid topic.

But then that doesn't seem to agree with the remark here:
https://github.com/zulip/zulip-flutter/pull/1342#discussion_r1964125345
If the user hasn't yet picked a topic at all, but topics aren't mandatory, it looks like the logic in the current version of #1364 would show "general chat" in the hint text here, which as mentioned in that comment isn't what we want.

lib/widgets/compose_box.dart
gnprice, line 592: The mixing here of some data that this class takes from the topic controller and then stores locally (`_topicTextNormalized`) and other data that it gets directly from the topic controller (`isTopicVacuous`) makes me nervous — what if those are out of sync?

I think that isn't actually a possibility, because I think `_topicTextNormalized` should always be the latest value from whenever the topic controller last changed. But it'd be good to not have to worry about it.

Looking at the history, I think the right solution is that there's no reason this widget state needs to be keeping `_topicTextNormalized` around at all — sent https://github.com/zulip/zulip-flutter/pull/1367 to fix that. (It was needed originally, but that was before we started keeping the normalized value around on the controller.)

lib/widgets/compose_box.dart
gnprice: Separately: I think it would be cleanest for this `topicDisplayName` logic to live on `ComposeTopicController` (as a getter there).

When I try reasoning through what it's doing — particularly in the eventual version found in #1364 — it's pretty intricately tied together with the details of what `isTopicVacuous` does. So it'd be helpful for their definitions to live next to each other.

PIG208:
Thanks for the review! I decided to drop [the alternative placeholder commit](https://github.com/zulip/zulip-flutter/commit/1ea0fcd613c20a05d3b811d47178e7a353b15ac3) because I think implementing the required refactoring in the later PRs helps with finding a better solution, as indicated in https://github.com/zulip/zulip-flutter/pull/1342#discussion_r1964389878.

lib/widgets/compose_box.dart
```diff
@@ -577,13 +584,13 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
     return _ContentInput(
       narrow: widget.narrow,
-      destination: TopicNarrow(widget.narrow.streamId,
-        TopicName(widget.controller.topic.textNormalized)),
+      destination: TopicNarrow(widget.narrow.streamId, topic),
       controller: widget.controller,
-      hintText: zulipLocalizations.composeBoxChannelContentHint(streamName,
-        widget.controller.topic.textNormalized));
+      hintText: zulipLocalizations.composeBoxChannelContentHint(
+        '#$streamName > ${topic.displayName}'));
```
gnprice, line 593: nit: I think this is a case where much of the information in the commit message is also relevant to understanding the resulting code (not only to understanding its history). So:
```suggestion
        // No i18n of this use of "#" and ">" string; those are part of how
        // Zulip expresses channels and topics, not any normal English punctuation,
        // so don't make sense to translate. See:
        //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
        '#$streamName > ${topic.displayName}'));
```

test/widgets/compose_box_test.dart
```diff
@@ -318,6 +321,85 @@ void main() {
...
+      required Narrow narrow,
+    }) async {
+      await prepareComposeBox(tester,
+        narrow: narrow,
+        otherUsers: [eg.otherUser, eg.thirdUser],
+        streams: [channel]);
+    }
+
+    /// This checks the input's configured hint text without regard to whether
+    /// it's currently visible, as it won't be if the user has entered some text.
```
gnprice, line 337: Hmm, this is a bit unsatisfying because it's checking a condition the user doesn't see. Ideally we'd be checking whether the user actually sees a hint, and if so then what it is.

From the thread at https://github.com/zulip/zulip-flutter/pull/1342#discussion_r1960414632 it sounds like getting that ideal behavior might be more complicated, though. In that case it's fine to let this be — the question of whether a hint is visible isn't critical for us to test, particularly as it's mostly the responsibility of the framework's underlying text-input code anyway.

gnprice (commented):
Thanks for the revision! I've now read the whole thing, and this all looks good now; just two small comments.

PIG208:
Updated the PR. I kept the commit message from the last commit, and added the proposed comments at the places that use the new string.

gnprice:
Thanks! Looks good; merging.

PR #1344, opened 2025-02-11
Better management of archived channel names

### LINKED ISSUE #800: Support better management of names of archived streams
In https://github.com/zulip/zulip/issues/26661, we plan to rework how archived streams are handled. As part of making these changes, the mobile apps should ideally match the web app in how messages in archived streams are displayed.

### Viewing messages in archived streams
- [ ] When viewing messages in an archived stream (e.g.,in search results), append " (archived)" to the end of the stream name. This should be marked for translation. (We can also consider using a different font or a pill in the header bar to indicate archived streams, but just "(archived)" should be fine to start with.)
- [ ] Topic links for archived streams should work. Currently, the server doesn't send data on archived streams to clients. See related discussion [here](https://chat.zulip.org/#narrow/stream/101-design/topic/archived.20streams.20.28renaming.29/near/1636412).

Web app / server issue:
- https://github.com/zulip/zulip/issues/26661

chimnayajith:
# Pull Request

## Description

This PR improves archived stream support in the mobile app to match the web app’s behavior. Key updates include:
- Appending " (archived)" to stream names when viewing messages.
- Updating UI elements (e.g., stream headers) to clearly indicate archived streams.
- Filtering out archived channels from relevant pages.
- Displaying the archived tag in a distinct font and color, consistent with the web app.
- Replacing the compose box with a "no permission" error banner instead of offering message composition.

## Related Issues

- Closes #800 

## Screenshots
<div>

</div>


lib/api/model/model.dart
```diff
@@ -347,6 +347,8 @@ class ZulipStream {
   // TODO(server-8): added in FL 199, was previously only on [Subscription] objects
   int? streamWeeklyTraffic;
 
+  final bool isArchived;
```
PIG208, line 350: nit: Make sure to match the order of these fields in the API documentation. See the dartdoc of `ZulipStream` to find the relevant piece.

lib/widgets/inbox.dart
```diff
@@ -284,16 +285,33 @@ abstract class _HeaderItem extends StatelessWidget {
...
+                    text: title(zulipLocalizations),
+                    style: TextStyle(
+                      fontSize: 17,
+                      height: (20 / 17),
+                      color: designVariables.labelMenuButton,
+                    ).merge(weightVariableTextStyle(context, wght: 600)),
+                  ),
+                  if (isArchived)
+                    TextSpan(
+                      text: ' ${zulipLocalizations.channelArchivedLabel}',
```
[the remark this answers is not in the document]
PIG208: We should use padding here to better support RTL languages, because this hard-coded space assumes that the text reads from left to right.

PIG208 (commented):
Thanks for working on this! One more thing that I think needs to be handled is the topic actions sheet. None of the supported actions (follow/mute/unmute/resolve/unresolve) works on a topic in an archived channel. Not sure what's the best way to deal with this, though. It would be a good question to ask on #mobile-design.

lib/api/model/model.dart
chimnayajith: this seems to be the order in the API documentation as I checked. Please let me know if its not

https://zulip.com/api/register-queue and go to `subscriptions` return value

PIG208: I see. The location the dartdoc points to has a different order. Let's match that:

```dart
/// For docs, search for "if stream"
/// in <https://zulip.com/api/register-queue>.
class ZulipStream {
```

lib/widgets/message_list.dart
```diff
@@ -351,6 +352,18 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+            padding: const EdgeInsets.symmetric(horizontal: 4),
+            child: Text(
+              zulipLocalizations.channelArchivedLabel,
+              style: TextStyle(
+                fontSize: 18,
+                color: messageListTheme.streamRecipientHeaderChevronRight,
+                fontStyle: FontStyle.italic,
+              ),
+            ),
+          ),
```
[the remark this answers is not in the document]
chimnayajith: There isn't a particular design that i could reference. I'm trying to keep it consistent with web.

lib/widgets/subscription_list.dart
```diff
@@ -187,10 +189,13 @@ class _SubscriptionList extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    // Filtering out the archived subscriptions.
+    final activeSubscriptions = subscriptions.where((sub) => !sub.isArchived).toList();
```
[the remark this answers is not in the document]
chimnayajith: Ah now that I check, its already being handled. removed this part.

chimnayajith:
The above comments were addressed and pushed. Waiting for clarification on [comment](https://github.com/zulip/zulip-flutter/pull/1344#pullrequestreview-2619162205) above no [CZO](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/archived.20channel.20topic.20actions/with/2109050)

lib/widgets/inbox.dart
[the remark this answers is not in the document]
PIG208: Bumping this; I see that you added the padding, but it is meant to replace the space in the string.

lib/api/model/model.dart
```diff
@@ -328,6 +328,7 @@ class ZulipStream {
 
   final int streamId;
   String name;
+  bool isArchived;
```
PIG208, line 331: We should make this optional. On older servers, this field will not be sent.

> is_archived: boolean
> A boolean indicating whether the channel is [archived](https://zulip.com/help/archive-a-channel).
> Changes: New in Zulip 10.0 (feature level 315). Previously, this endpoint never returned archived channels.

A clean way to do this might be defaulting this to `false`, since its value can be indicated from the fact that no archived channels were returned from this endpoint before, with a `// TODO(server-10) remove default` comment.

lib/widgets/inbox.dart
```diff
@@ -284,16 +286,32 @@ abstract class _HeaderItem extends StatelessWidget {
...
+                    // TODO(#1285): Avoid concatenating translated strings
+                    WidgetSpan(
+                      child: Padding(
+                        padding: EdgeInsetsDirectional.only(start: 4),
+                        child: Text(
+                          zulipLocalizations.channelArchivedLabel,
+                          style: TextStyle(
+                            fontSize: 17,
+                            height: 20 / 17,
+                            color: MessageListTheme.of(context).streamRecipientHeaderChevronRight,
```
PIG208, line 312: This line is too long. Let's just extract the `MessageListTheme` as a variable at the start.

PIG208: It also doesn't seem right that we are using `streamRecipientHeaderChevronRight` as the color here. Web uses `--color-text-message-header-archived`; how about adding that to `DesignVariables`'s "Not named variables in Figma; taken from older Figma drafts, or elsewhere." section

Let's also have a TODO(design) comment to check if this is the right variable.

test/widgets/inbox_test.dart
```diff
@@ -118,17 +118,29 @@ void main() {
...
 
-    return tester.widget(
-      find.ancestor(
-        of: find.byWidget(rowLabel),
-        matching: find.byType(Row)));
+    final richTextFinder = find.byWidgetPredicate((widget) =>
+      widget is RichText &&
+      widget.text.toPlainText().contains(label)
+    );
+
```
PIG208, line 134: nit: remove newline since these stanzas are closely related

test/widgets/inbox_test.dart
```diff
@@ -595,6 +607,38 @@ void main() {
...
+          check(headerRowFinder).isNotNull();
+
+          final richTextFinder = find.descendant(
+            of: find.byWidget(headerRowFinder!),
+            matching: find.byWidgetPredicate((widget) =>
+              widget is RichText &&
+              widget.text.toPlainText().contains(stream.name)
+            ),
+          );
+          expect(richTextFinder, findsOneWidget);
```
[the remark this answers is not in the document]
PIG208: This comment also applies to other similar places in the PR, so please look for them and update.

test/widgets/inbox_test.dart
```diff
@@ -595,6 +607,38 @@ void main() {
           check(rectAfterTap).equals(rectBeforeTap);
         });
 
+        testWidgets('shows archived label for archived streams', (tester) async {
+          final stream = eg.stream(streamId: 1, isArchived: true);
```
PIG208, line 611: Looks like `streamId` being `1` is not required for this test. We can remove that to keep this test boring and focused.

test/widgets/inbox_test.dart
PIG208, line 629: Looking at the implementation of `findStreamHeaderRow`, this stanza seems to be repetitive. Let's just remove this check. Another benefit of removing this is that we can focus more on checking the presence of the "(archived)" label.

test/widgets/inbox_test.dart
```diff
@@ -595,6 +607,38 @@ void main() {
...
+
+          final archivedLabelFinder = find.descendant(
+            of: find.byWidget(headerRowFinder),
+            matching: find.text(' (archived)')
+          );
+          expect(archivedLabelFinder, findsOneWidget);
+
+          // Check that the Text widget has italic style
+          final archivedLabelText = tester.widget<Text>(archivedLabelFinder);
+          expect(archivedLabelText.style?.fontStyle, FontStyle.italic);
```
PIG208, line 639: We usually leave out checks for specific styling unless there is some extra logic to it (changing style dynamically to different inputs, etc.)

test/widgets/message_list_test.dart
```diff
@@ -1101,6 +1101,28 @@ void main() {
         await tester.pump();
         check(pushedRoutes).isEmpty();
       });
+
+      testWidgets('shows archived label for archived streams', (tester) async {
+        final stream = eg.stream(streamId: 1, name: 'stream name', isArchived: true);
```
PIG208, line 1106: Similarly, we can remove `streamId` here.

test/widgets/compose_box_test.dart
```diff
@@ -951,6 +951,34 @@ void main() {
...
+      final narrowTestCases = [
+        ('channel', const ChannelNarrow(1)),
+        ('topic',   eg.topicNarrow(1, 'topic')),
+      ];
+
+      for (final (String narrowType, Narrow narrow) in narrowTestCases) {
+        testWidgets('error banner is shown in $narrowType narrow', (tester) async {
+          await prepareComposeBox(tester,
+            narrow: narrow,
+            selfUser: eg.user(),
```
PIG208, line 967: We can remove `selfUser` if the tests do not rely on it.

test/widgets/subscription_list_test.dart
```diff
@@ -75,6 +75,25 @@ void main() {
...
+    check(getItemCount()).equals(2);
+    check(isPinnedHeaderInTree()).isTrue();
+    check(isUnpinnedHeaderInTree()).isTrue();
+
+    check(find.text('stream 1')).findsNothing();
+    check(find.text('stream 2')).findsNothing();
+
+    check(find.text('stream 3')).findsOne();
+    check(find.text('stream 4')).findsOne();
+  });
```
PIG208, line 95: nit: Order this after "only unpinned subscriptions" because "basic subscriptions", "only pinned subscriptions" are more similar to that test and they should be next to each other.

PIG208 (commented):
Thanks for the update! Left some more comments.

PIG208:
While addressing the review comments, a useful thing to do (and often necessary) is to try to find other places where the comment applies to during your self-reviews. This way we can save some back-and-forth revisions when there are some remaining known issues.

chimnayajith:
@PIG208 Pushed a revision. PTAL. Removed unsupported options from the channel and message action sheets for archived streams.

PIG208 (commented):
Thanks for the update! I just went through it again. Left some comments.

PIG208:
Hi @chimnayajith, I saw that there are some recent pushes. Please let me know if this is ready for review! Thanks.

PIG208 (commented):
I haven't gone through the tests yet. Here are some comments on the implementation.

Relatively recently, there was a server change (https://github.com/zulip/zulip/pull/34258) that causes stream update events to be sent when channels are archived. We should handle that to support live-updates properly. See `ChannelStore.handleChannelEvent`, where we handle the channel events, and [#api design > archived channels events @ 💬](https://chat.zulip.org/#narrow/channel/378-api-design/topic/archived.20channels.20events/near/2137668) for reference.

A client should only pass `archived_channels` when sending a register-queue request when it actually supports handling them. So let's move the first commit (1ba56c566) to the end.

chimnayajith:
Marking this as Draft for now since the issue is marked MXA:Later

PR #1347, opened 2025-02-12
doc: Add libdrm troubleshooting entry

gnprice:
I ran into this recently when I tried running the Linux app.

(It looks like the reason I hadn't seen it before is that until recently I had this package installed; it was installed "automatically", i.e. only because it was a dependency of something else, and then I upgraded my machine, it was no longer such a dependency, and got autoremoved.  Now that I've installed it directly, it'll stay.)


gnprice:
Merging directly, because this is a pure docs PR.

PR #1348, opened 2025-02-12
lightbox: Prevent hero animation between message lists

### LINKED ISSUE #44: lightbox: Fix Hero when multiple images in one message have same URL
When you tap an image preview in a message, the Hero animation needs a "tag" that uniquely identifies the image on the route (i.e., among all other images in the message list).

Currently, `_LightboxHeroTag` fails to uniquely identify an image when there are multiple image previews with the same URL in a message. The effect is that there is no Hero animation, and an error is logged to the console:

```
======== Exception caught by scheduler library =====================================================
The following assertion was thrown during a scheduler callback:
There are multiple heroes that share the same tag within a subtree.

Within each subtree for which heroes are to be animated (i.e. a PageRoute subtree), each Hero must have a unique non-null tag.
In this case, multiple heroes had the following tag: […]
```

The bug is a bit tricky to reproduce: the open-lightbox-on-tap handler is on the image preview widget itself, not on the image link text, and (empirically) servers try to deduplicate image previews in a message with the same URL. Try something like this to thwart that deduplication:

```
[image](/user_uploads/2/fd/AogAK3meEt4GkXJaDvDXNwyy/image.png)
[image](https://chat.zulip.org/user_uploads/2/fd/AogAK3meEt4GkXJaDvDXNwyy/image.png)
```

This TODO on `_LightboxHeroTag` gives a recipe for a potential fix:

```dart
// TODO: Add index of the image preview in the message, to not break if
//   there are multiple image previews with the same URL in the same
//   message. Maybe keep `src`, so that on exit the lightbox image doesn't
//   fly to an image preview with a different URL, following a message edit
//   while the lightbox was open.
```

### LINKED ISSUE #930: Hero animation for images should run only to/from lightbox, not between message lists
When a message has an image and you tap the image to see it in the lightbox, we use a `Hero` widget to cause the image to animate from its old position and size in the message list to its new position and size in the lightbox, and to do the same in reverse when you navigate back.

Currently, though, our implementation of this causes those "hero" animations to happen also when you navigate from one message list to another that happens to contain the same image-bearing message. That doesn't make so much sense, and tends to look glitchy. It can look extra glitchy when the message is off screen either before or after the navigation (or even both), as spotted by @alexmv:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/Images.20have.20the.20wrong.20z-index/near/1932012

So we should fix that.

I think the way to implement this will be to augment the `tag` parameter we pass to `Hero` so that it identifies not only the message and the image within the message, but the message list the message is shown in.

lakshya1goel:
Fixes #930.
Fixes #44.

### Before

https://github.com/user-attachments/assets/37a832c2-87d6-4223-a47e-f5549fccfead

### After

https://github.com/user-attachments/assets/206af9a4-f487-49ad-adca-54faa70d5783

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -651,6 +651,7 @@ class MessageImage extends StatelessWidget {
         Navigator.of(context).push(getImageLightboxRoute(
           context: context,
           message: message,
+          messageImageContext: context,
           src: resolvedSrcUrl,
           thumbnailUrl: resolvedThumbnailUrl,
           originalWidth: node.originalWidth,
@@ -659,7 +660,7 @@ class MessageImage extends StatelessWidget {
       child: node.loading
         ? const CupertinoActivityIndicator()
         : resolvedSrcUrl == null ? null : LightboxHero(
-            message: message,
+            messageImageContext: context,
             src: resolvedSrcUrl,
             child: RealmContentNetworkImage(
               resolvedThumbnailUrl ?? resolvedSrcUrl,
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -15,45 +15,55 @@ import 'dialog.dart';
 import 'page.dart';
 import 'store.dart';
 
-// TODO(#44): Add index of the image preview in the message, to not break if
-//   there are multiple image previews with the same URL in the same
-//   message. Maybe keep `src`, so that on exit the lightbox image doesn't
-//   fly to an image preview with a different URL, following a message edit
-//   while the lightbox was open.
 class _LightboxHeroTag {
-  _LightboxHeroTag({required this.messageId, required this.src});
+  _LightboxHeroTag({
+    required this.messageImageContext,
+    required this.src,
+  });
 
-  final int messageId;
+  /// The [BuildContext] of the image in the message list that's being expanded
+  /// into the lightbox. Used to coordinate the Hero animation between this specific
+  /// image and the lightbox view.
+  ///
+  /// This helps ensure the animation only happens between the correct image instances,
+  /// preventing unwanted animations between different message lists or between
+  /// different images that happen to have the same URL.
+  // TODO: write a regression test for #44, duplicate images within a message
+  final BuildContext messageImageContext;
+
+  /// The image source URL. Used to match the source and destination images
+  /// during the Hero animation, ensuring the animation only occurs between
+  /// images showing the same content.
   final Uri src;
 
   @override
   bool operator ==(Object other) {
     return other is _LightboxHeroTag &&
-      other.messageId == messageId &&
+      other.messageImageContext == messageImageContext &&
       other.src == src;
   }
 
   @override
-  int get hashCode => Object.hash('_LightboxHeroTag', messageId, src);
+  int get hashCode => Object.hash('_LightboxHeroTag', messageImageContext, src);
 }
 
 /// Builds a [Hero] from an image in the message list to the lightbox page.
 class LightboxHero extends StatelessWidget {
   const LightboxHero({
     super.key,
-    required this.message,
+    required this.messageImageContext,
     required this.src,
     required this.child,
   });
 
-  final Message message;
+  final BuildContext messageImageContext;
   final Uri src;
   final Widget child;
 
   @override
   Widget build(BuildContext context) {
     return Hero(
-      tag: _LightboxHeroTag(messageId: message.id, src: src),
+      tag: _LightboxHeroTag(messageImageContext: messageImageContext, src: src),
       flightShuttleBuilder: (
         BuildContext flightContext,
         Animation<double> animation,
@@ -226,6 +236,7 @@ class _ImageLightboxPage extends StatefulWidget {
   const _ImageLightboxPage({
     required this.routeEntranceAnimation,
     required this.message,
+    required this.messageImageContext,
     required this.src,
     required this.thumbnailUrl,
     required this.originalWidth,
@@ -234,6 +245,7 @@ class _ImageLightboxPage extends StatefulWidget {
 
   final Animation<double> routeEntranceAnimation;
   final Message message;
+  final BuildContext messageImageContext;
   final Uri src;
   final Uri? thumbnailUrl;
   final double? originalWidth;
@@ -317,7 +329,7 @@ class _ImageLightboxPageState extends State<_ImageLightboxPage> {
         child: InteractiveViewer(
           child: SafeArea(
             child: LightboxHero(
-              message: widget.message,
+              messageImageContext: widget.messageImageContext,
               src: widget.src,
               child: RealmContentNetworkImage(widget.src,
                 filterQuality: FilterQuality.medium,
@@ -599,6 +611,7 @@ Route<void> getImageLightboxRoute({
   int? accountId,
   BuildContext? context,
   required Message message,
+  required BuildContext messageImageContext,
   required Uri src,
   required Uri? thumbnailUrl,
   required double? originalWidth,
@@ -611,6 +624,7 @@ Route<void> getImageLightboxRoute({
       return _ImageLightboxPage(
         routeEntranceAnimation: animation,
         message: message,
+        messageImageContext: messageImageContext,
         src: src,
         thumbnailUrl: thumbnailUrl,
         originalWidth: originalWidth,
--- a/lib/widgets/lightbox.dart
+++ b/lib/widgets/lightbox.dart
@@ -15,25 +15,37 @@ import 'dialog.dart';
 import 'page.dart';
 import 'store.dart';
 
+/// Identifies which [LightboxHero]s should match up with each other
+/// to produce a hero animation.
+///
+/// See [Hero.tag], the field where we use instances of this class.
+///
+/// The intended behavior is that when the user acts on an image
+/// in the message list to have the app expand it in the lightbox,
+/// a hero animation goes from the original view of the image
+/// to the version in the lightbox,
+/// and back to the original upon exiting the lightbox.
 class _LightboxHeroTag {
   _LightboxHeroTag({
     required this.messageImageContext,
     required this.src,
   });
 
-  /// The [BuildContext] of the image in the message list that's being expanded
-  /// into the lightbox. Used to coordinate the Hero animation between this specific
-  /// image and the lightbox view.
+  /// The [BuildContext] for the [MessageImage] being expanded into the lightbox.
   ///
-  /// This helps ensure the animation only happens between the correct image instances,
-  /// preventing unwanted animations between different message lists or between
-  /// different images that happen to have the same URL.
+  /// In particular this prevents hero animations between
+  /// different message lists that happen to have the same message.
+  /// It also distinguishes different copies of the same image
+  /// in a given message list.
   // TODO: write a regression test for #44, duplicate images within a message
   final BuildContext messageImageContext;
 
-  /// The image source URL. Used to match the source and destination images
-  /// during the Hero animation, ensuring the animation only occurs between
-  /// images showing the same content.
+  /// The image source URL.
+  ///
+  /// This ensures the animation only occurs between matching images, even if
+  /// the message was edited before navigating back to the message list
+  /// so that the original [MessageImage] has been replaced in the tree
+  /// by a different image.
   final Uri src;
 
   @override
--- a/test/widgets/lightbox_test.dart
+++ b/test/widgets/lightbox_test.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:math';
 
 import 'package:checks/checks.dart';
 import 'package:clock/clock.dart';
@@ -10,12 +11,18 @@ import 'package:video_player_platform_interface/video_player_platform_interface.
 import 'package:video_player/video_player.dart';
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/model/localizations.dart';
+import 'package:zulip/model/narrow.dart';
+import 'package:zulip/model/store.dart';
 import 'package:zulip/widgets/app.dart';
 import 'package:zulip/widgets/content.dart';
 import 'package:zulip/widgets/lightbox.dart';
+import 'package:zulip/widgets/message_list.dart';
 
+import '../api/fake_api.dart';
 import '../example_data.dart' as eg;
 import '../model/binding.dart';
+import '../model/content_test.dart';
+import '../model/test_store.dart';
 import '../test_images.dart';
 import 'dialog_checks.dart';
 import 'test_app.dart';
@@ -197,6 +204,113 @@ class FakeVideoPlayerPlatform extends Fake
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  group('LightboxHero', () {
+    late PerAccountStore store;
+    late FakeApiConnection connection;
+
+    final channel = eg.stream();
+    final message = eg.streamMessage(stream: channel,
+      topic: 'test topic', contentMarkdown: ContentExample.imageSingle.html);
+
+    // From ContentExample.imageSingle.
+    final imageSrcUrlStr = 'https://chat.example/user_uploads/thumbnail/2/ce/nvoNL2LaZOciwGZ-FYagddtK/image.jpg/840x560.webp';
+    final imageSrcUrl = Uri.parse(imageSrcUrlStr);
+    final imageFinder = find.byWidgetPredicate(
+      (widget) => widget is RealmContentNetworkImage && widget.src == imageSrcUrl);
+
+    Future<void> setupMessageListPage(WidgetTester tester) async {
+      addTearDown(testBinding.reset);
+      final subscription = eg.subscription(channel);
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot(
+        streams: [channel], subscriptions: [subscription]));
+      store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+      connection = store.connection as FakeApiConnection;
+      await store.addUser(eg.selfUser);
+
+      connection.prepare(json:
+        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
+      await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
+        child: MessageListPage(initNarrow: const CombinedFeedNarrow())));
+      await tester.pumpAndSettle();
+    }
+
+    testWidgets('Hero animation occurs smoothly when opening lightbox from message list', (tester) async {
+      double dist(Rect a, Rect b) =>
+        sqrt(pow(a.top - b.top, 2) + pow(a.left - b.left, 2));
+
+      prepareBoringImageHttpClient();
+
+      await setupMessageListPage(tester);
+
+      final initialImagePosition = tester.getRect(imageFinder);
+      await tester.tap(imageFinder);
+      await tester.pump();
+      // pump to start hero animation
+      await tester.pump();
+
+      const heroAnimationDuration = Duration(milliseconds: 300);
+      const steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      final animatedPositions = <Rect>[];
+      for (int i = 1; i <= steps; i++) {
+        await tester.pump(stepDuration);
+        animatedPositions.add(tester.getRect(imageFinder));
+      }
+
+      final totalDistance = dist(initialImagePosition, animatedPositions.last);
+      Rect previousPosition = initialImagePosition;
+      double maxStepDistance = 0.0;
+      for (final position in animatedPositions) {
+        final stepDistance = dist(previousPosition, position);
+        maxStepDistance = max(maxStepDistance, stepDistance);
+        check(position).not((pos) => pos.equals(previousPosition));
+
+        previousPosition = position;
+      }
+      check(maxStepDistance).isLessThan(0.03 * totalDistance);
+
+      debugNetworkImageHttpClientProvider = null;
+    });
+
+    testWidgets('no hero animation occurs between different message list pages for same image', (tester) async {
+      Rect getElementRect(Element element) =>
+        tester.getRect(find.byElementPredicate((e) => e == element));
+
+      prepareBoringImageHttpClient();
+
+      await setupMessageListPage(tester);
+
+      final firstElement = tester.element(imageFinder);
+      final firstImagePosition = getElementRect(firstElement);
+
+      connection.prepare(json:
+        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
+      await tester.tap(find.descendant(
+        of: find.byType(StreamMessageRecipientHeader),
+        matching: find.text('test topic')));
+      await tester.pumpAndSettle();
+
+      final secondElement = tester.element(imageFinder);
+      final secondImagePosition = getElementRect(secondElement);
+
+      await tester.tap(find.byType(BackButton));
+      await tester.pump();
+
+      const heroAnimationDuration = Duration(milliseconds: 300);
+      const steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+        check(tester.elementList(imageFinder))
+          .unorderedEquals([firstElement, secondElement]);
+        check(getElementRect(firstElement)).equals(firstImagePosition);
+        check(getElementRect(secondElement)).equals(secondImagePosition);
+      }
+
+      debugNetworkImageHttpClientProvider = null;
+    });
+  });
+
   group('_ImageLightboxPage', () {
     final src = Uri.parse('https://chat.example/lightbox-image.png');
 
@@ -216,6 +330,7 @@ void main() {
       unawaited(navigator.push(getImageLightboxRoute(
         accountId: eg.selfAccount.id,
         message: message ?? eg.streamMessage(),
+        messageImageContext: navigator.context,
         src: src,
         thumbnailUrl: thumbnailUrl,
         originalHeight: null,


gnprice:
Thanks! Chat thread here:
https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Hero.20animation/near/2092791
with comments particularly on how to write a good test for this fix.

lakshya1goel:
Hi @PIG208, PR is ready for a review now, PTAL, Thanks!

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +601,51 @@ void main() {
       check(platform.position).equals(position);
     });
   });
-}
+
+  group('LightboxHero', () {
+    testWidgets('no hero animation occurs between different message list pages for same image', (tester) async {
+      final channel = eg.stream(streamId: eg.defaultStreamMessageStreamId);
+      final message = eg.streamMessage(stream: channel,
+        contentMarkdown: ContentExample.spoilerHeaderHasImage.html, topic: 'test topic');
```
PIG208, line 609: nit: This example seems a bit too atypical. How about using something like `imageSingle` to replicate a message that a user is more likely to encounter.

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +601,51 @@ void main() {
...
+
+      final imageFinder = find.byType(RealmContentNetworkImage);
+      final initialImageRect = tester.getRect(imageFinder);
+
+      await tester.tap(imageFinder);
+      await tester.pump();
+      await tester.pump(const Duration(milliseconds: 150));
+
+      final imageInTransition = tester.getRect(imageFinder);
+      check(imageInTransition.top != initialImageRect.top).isTrue();
```
PIG208, line 646: nit: utilize checks API when possible:
```suggestion
      check(imageInTransition.top).not((it) => it.equals(initialImageRect.top));
```
same below

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +601,51 @@ void main() {
...
+      check(imageInTransition.left != initialImageRect.left).isTrue();
+      await tester.pumpAndSettle();
+    });
+  });
+}
```
PIG208, line 651: Need a newline at the end of the file.

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +601,51 @@ void main() {
...
+      await tester.tap(find.descendant(
+        of: find.byType(StreamMessageRecipientHeader),
+        matching: find.text('test topic')));
+      await tester.pump();
+      await tester.pump(const Duration(milliseconds: 150));
+
+      final imageInTransition = tester.getRect(imageFinder);
+      check(imageInTransition.top).equals(initialImageRect.top);
+      check(imageInTransition.left).equals(initialImageRect.left);
+      await tester.pumpAndSettle();
```
PIG208, line 628: Is this supposed to be mid transition time? While this works, I feel that it can miss bugs if the duration somehow changes.

We can have a while-loop that checks `tester.hasRunningAnimations`, which pumps repeatedly with a fixed duration, and repeats the check:

```suggestion

      int timeElapsed = 0;
      const interval = 50;
      while (timeElapsed < interval || tester.hasRunningAnimations) {
        final imageInTransition = tester.getRect(imageFinder);
        check(imageInTransition.top).equals(initialImageRect.top);
        check(imageInTransition.left).equals(initialImageRect.left);
        await tester.pump(const Duration(milliseconds: interval));
        timeElapsed += interval;
      }
      check(timeElapsed).isGreaterOrEqual(interval);
```

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,37 @@ class FakeVideoPlayerPlatform extends Fake
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  late PerAccountStore store;
+  late FakeApiConnection connection;
+
+  Future<void> setupMessageListPage(WidgetTester tester, {
```
PIG208, line 210: Let's keep this helper close to where it is used in the 'LightBoxHero' group.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,37 @@ class FakeVideoPlayerPlatform extends Fake
...
+    Narrow narrow = const CombinedFeedNarrow(),
+    List<Message>? messages,
+    List<ZulipStream>? streams,
```
PIG208, line 213: Because none of the two callers actually rely on having different messages/channels, it looks like we can remove these parameters. An advantage of keeping the helper close to where it's used is that we can more easily spot what properties the tests rely on.

For things that are constant to the tests (like the topic name, finders), we can have a shared local variable within the group.

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +601,51 @@ void main() {
...
+  group('LightboxHero', () {
+    testWidgets('no hero animation occurs between different message list pages for same image', (tester) async {
+      final channel = eg.stream(streamId: eg.defaultStreamMessageStreamId);
+      final message = eg.streamMessage(stream: channel,
+        contentMarkdown: ContentExample.spoilerHeaderHasImage.html, topic: 'test topic');
+      await setupMessageListPage(tester, narrow: const CombinedFeedNarrow(),
+        streams: [channel], messages: [message]);
+
+      connection.prepare(json:
+        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
```
PIG208, line 614: nit: move this to the line just before `tester.tap`, because it prepares an API response for that

PIG208 (commented):
Thanks for fixing this! The approach looks fine to me. Left a comment on the implementation and some more on the tests.

test/widgets/lightbox_test.dart
PIG208: Just saw the [CZO comment](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Hero.20animation/near/2095334)[#mobile-dev-help > Hero animation @ 💬](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Hero.20animation/near/2095334):

> In order to have that benefit, it's important to keep the two test cases similar to each other as far as possible. It should be easy for the reader to convince themself that the two tests are checking the same thing, and just expecting opposite outcomes.

I agree that having these two tests similar to each other does mitigate the concern of one of them not working properly. So it's probably also fine to leave them as-is. Either way, we should comment that the duration is specifically chosen such that we are in the middle of a hero animation (if there is one).

lib/widgets/lightbox.dart
```diff
@@ -44,16 +46,18 @@ class LightboxHero extends StatelessWidget {
...
   final Message message;
   final Uri src;
   final Widget child;
+  final BuildContext pageContext;
 
   @override
   Widget build(BuildContext context) {
     return Hero(
-      tag: _LightboxHeroTag(messageId: message.id, src: src),
+      tag: _LightboxHeroTag(messageId: message.id, src: src, pageContext: pageContext),
```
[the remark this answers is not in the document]
lakshya1goel: Things would go wrong since using PageRoot.contextOf(context) here would break the hero animation because the widget exists in two different page contexts simultaneously in the `MessageListPage` and the `lightbox`. By explicitly passing the source page's context (pageContext), we ensure both the source and destination heroes use the same context value in their tags, allowing Flutter to properly match them for the animation.

test/widgets/lightbox_test.dart
```diff
@@ -207,14 +245,19 @@ void main() {
...
+      // 2. The PageRoot widget to provide context for Hero animations between
+      //    the message list and lightbox.
+      await tester.pumpWidget(PageRoot(
+        child: const ZulipApp()
+      ));
       await tester.pump();
       final navigator = await ZulipApp.navigator;
       unawaited(navigator.push(getImageLightboxRoute(
         accountId: eg.selfAccount.id,
+        pageContext: PageRoot.contextOf(navigator.context),
```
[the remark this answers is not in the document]
lakshya1goel: No, I think we can't skip passing `pageContext` because during the hero animation, the transitioning widget lives in Flutter's overlay layer outside our PageRoot hierarchy. The hero animation needs the same tag value at both ends (source and destination) to know which widgets to animate between, and since we can't get the PageRoot context during the transition, we must pass it explicitly.

lib/widgets/lightbox.dart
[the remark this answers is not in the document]
PIG208: Indeed. Thanks for the explanation.

lakshya1goel:
Pushed the revision, PTAL @PIG208. Thanks!

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +571,77 @@ void main() {
...
     });
   });
+
+  group('LightboxHero', () {
+    late PerAccountStore store;
+    late FakeApiConnection connection;
+
+    final channel = eg.stream();
+    final message = eg.streamMessage(stream: channel,
+        contentMarkdown: ContentExample.imageSingle.html, topic: 'test topic');
```
PIG208, line 581: nit: indentation
```suggestion
    final message = eg.streamMessage(stream: channel,
      contentMarkdown: ContentExample.imageSingle.html, topic: 'test topic');
```

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +571,77 @@ void main() {
...
+
+    Future<void> setupMessageListPage(WidgetTester tester) async {
+      addTearDown(testBinding.reset);
+      final subscription = eg.subscription(channel);
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot(
+        streams: [channel], subscriptions: [subscription]));
+      store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+      connection = store.connection as FakeApiConnection;
+
+      await store.addUser(eg.selfUser);
```
PIG208, line 591: nit:
```suggestion
      await store.addUser(eg.selfUser);
```
because `store.addUser` is a part of the setup code stanza

lib/widgets/content.dart
```diff
@@ -663,12 +665,14 @@ class MessageImage extends StatelessWidget {
           src: resolvedSrcUrl,
           thumbnailUrl: resolvedThumbnailUrl,
           originalWidth: node.originalWidth,
-          originalHeight: node.originalHeight));
+          originalHeight: node.originalHeight,
+          pageContext: pageContext));
```
PIG208, line 669: While testing this, I tried passing `context` to `pageContext` (and below), and it appears to work. Let's try this so that we don't need to rely on having a PageRoot.

PIG208 (commented):
Thanks for the update! I found a place where we can potentially shake off the dependency on PageRoot. Let me know what you think!

lakshya1goel:
Pushed the revision @PIG208, please have a look. Thanks!

PIG208:
Thanks! I think because we are not using the context of `PageRoot`, there is probably a better name than `pageConext`. (Maybe `messageImageContext`?) Marking this for Greg's review.

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +566,76 @@ void main() {
...
+      await tester.tap(find.descendant(
+        of: find.byType(StreamMessageRecipientHeader),
+        matching: find.text('test topic')));
+      await tester.pump();
+      // Pump halfway through the hero animation duration (300ms)
+      await tester.pump(const Duration(milliseconds: 150));
+
+      final imageInTransition = tester.getRect(imageFinder);
+      check(imageInTransition).top.equals(initialImageRect.top);
+      check(imageInTransition).left.equals(initialImageRect.left);
```
gnprice, line 616: Given `imageFinder = find.byType(RealmContentNetworkImage).first`, this doesn't make an entirely convincing check that there isn't a hero animation going on. This effectively says that the original image is still the *first* RealmContentNetworkImage in the tree — but if there were a hero animation going on, there's no reason that necessarily has to be the first image in the tree. Maybe it comes later in the tree than the original image.

test/widgets/lightbox_test.dart
```diff
@@ -558,4 +566,76 @@ void main() {
...
+      final initialImageRect = tester.getRect(imageFinder);
+
+      await tester.tap(imageFinder);
+      await tester.pump();
+      // Pump halfway through the hero animation duration (300ms)
+      await tester.pump(const Duration(milliseconds: 150));
+
+      final imageInTransition = tester.getRect(imageFinder);
+      check(imageInTransition).top.not((it) => it.equals(initialImageRect.top));
+      check(imageInTransition).left.not((it) => it.equals(initialImageRect.left));
```
gnprice, line 636: Similarly, this says that the first RealmContentNetworkImage in the tree isn't at the place the original image was. But
* there's the new route, the lightbox, which has a version of the image — maybe that's the first one in the tree now
* there's the old route, the message list, which may have a version of the image — and although right now that won't have moved, maybe in the future we'll use a different navigation transition here so that the old route moves away while the new one is moving in. Again maybe that's the first one in the tree now.

So this should get more specific about our expectations in order to be convincing that it wouldn't end up passing even in some future where we've broken the hero animation entirely.

gnprice (commented):
Thanks @lakshya1goel for taking this on, and @PIG208 for the previous reviews! Comments below.

lib/widgets/lightbox.dart
```diff
@@ -21,20 +21,26 @@ import 'store.dart';
 //   fly to an image preview with a different URL, following a message edit
 //   while the lightbox was open.
```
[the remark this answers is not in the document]
lakshya1goel: I think the current implementation handles this scenario well. When a message is edited while its image is open in the lightbox, the behavior depends on whether the image URL changes:
- If the edit changes the image URL:
The Hero animation won't find a matching tag (because src is part of the tag and has changed)
It will gracefully fall back to a fade transition instead of attempting to animate to the wrong image
- If the edit doesn't change the image URL:
All three tag components (messageId, src, and messageImageContext) still match
The Hero animation will work normally

test/widgets/lightbox_test.dart
```diff
@@ -300,6 +308,100 @@ void main() {
...
+      await tester.tap(find.descendant(
+        of: find.byType(StreamMessageRecipientHeader),
+        matching: find.text('test topic')));
+      await tester.pump();
+      // Pump halfway through the hero animation duration (300ms)
+      await tester.pump(const Duration(milliseconds: 150));
+
+      final imageInTransition = tester.getRect(imageFinder);
+      check(imageInTransition).top.equals(initialImageRect.top);
+      check(imageInTransition).left.equals(initialImageRect.left);
```
gnprice, line 399: This test is also less specific than the previous revision. In fact this one no longer tests anything about the hero animation at all: even if you edit the code so that the tags are completely defeated and the hero animation happens all the time:
```diff
   bool operator ==(Object other) {
-    return other is _LightboxHeroTag &&
-      other.messageId == messageId &&
-      other.src == src &&
-      other.messageImageContext == messageImageContext;
+    return true;
   }
 
   @override
-  int get hashCode => Object.hash('_LightboxHeroTag', messageId, src, messageImageContext);
+  int get hashCode => '_LightboxHeroTag'.hashCode;
```

this test still passes.

Like with the other test case, can you describe what condition this verifies is true, in terms that reflect what the user sees on the screen?

gnprice (commented):
Thanks for the revision — comments below.

For detailed how-to discussion, let's use [#mobile-dev-help > Hero animation](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Hero.20animation/with/2117577), since Zulip will be better for discussion than GitHub is.

gnprice:
Thanks! Chat thread here about the tests: [#mobile-dev-help > Hero animation @ 💬](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Hero.20animation/near/2131953)

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,152 @@ class FakeVideoPlayerPlatform extends Fake
...
+        .where((element) => (element.widget as Hero).tag == heroWidgetTag)
+        .where((element) => element.findAncestorWidgetOfExactType<InteractiveViewer>() == null);
+
+        final destinationHeroElement = tester.elementList(find.byType(Hero))
+        .where((element) => (element.widget as Hero).tag == heroWidgetTag)
+        .where((element) => element.findAncestorWidgetOfExactType<InteractiveViewer>() != null);
+
+        check(sourceHeroElement).isNotEmpty();
+        //without hero animation, destination hero element should not exist
+        check(destinationHeroElement).isEmpty();
```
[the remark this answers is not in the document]
gnprice: And more fundamentally: this test does not succeed at doing its job as a test.

The purpose of a test is so that we don't accidentally revert your change. If we do revert your change, the test should fail.

Here:
* Revert your change, but keep the new tests, with the command `git checkout @~ lib/`.
* A different test below needs to be tweaked so it still compiles: comment out the `messageImageContext: navigator.context,` in a call to `getImageLightboxRoute`.
* Run this test file: `flutter test test/widgets/lightbox_test.dart`.
* This test "no hero animation occurs between different message list pages for same image" still passes, even though in reality a hero animation now *does* occur between such pages.
* (And the other test "Hero animation occurs smoothly when opening lightbox from message list" also gets the wrong answer: that one fails, even though in reality the thing it's testing for still works. That's because of that test's use of the `key` that was artificially added for the test — which is a good illustration of why that's not a good way for that test to work.)

gnprice (commented):
Thanks for the revision. This is closer, but there are still significant things that need to be changed before we can merge it.

In particular, this version still doesn't have a test that successfully exercises the issue this PR is meant to fix. See the last pair of comments below.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,152 @@ class FakeVideoPlayerPlatform extends Fake
...
+        final positionChangeLeft = animatedFlightImagePosition.left - previousAnimatedPosition.left;
+        check(positionChangeTop.abs()).isLessThan(0.1*totalTopDistance);
+        check(positionChangeLeft.abs()).isLessThan(0.1*totalLeftDistance);
+
+        previousAnimatedPosition = animatedFlightImagePosition;
+      }
+
+      await tester.pumpAndSettle();
+
+      final lightBoxImageFinder = find.byKey(Key(message.id.toString()));
```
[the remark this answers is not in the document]
lakshya1goel: actually this was done because when I was finding the image using the src url it is finding two images , this could be due to generation of  source and destination hero image during hero animation , how could I uniquely find out the image element without key ?

lakshya1goel:
Hi @gnprice, thanks for the detailed reviews so far. 
I have revised the implementation as:
- Removed the hero related implementation and focused purely on the image now. (Used that earlier to check that the animation is actually a hero animation, nevertheless). 
- Removed the key used for the identification of the image rather used the image src now. 
- Moreover, sorry for the 2nd test, now I have checked and verified that it is failing on reverting my changes and passing after the fix. 
- Resolved the other comments mentioned above.

PTAL and let me know if anything else is required.
Thanks!

lib/widgets/lightbox.dart
```diff
@@ -21,39 +21,55 @@ import 'store.dart';
...
   final Message message;
+  final BuildContext messageImageContext;
   final Uri src;
   final Widget child;
 
   @override
   Widget build(BuildContext context) {
     return Hero(
-      tag: _LightboxHeroTag(messageId: message.id, src: src),
+      tag: _LightboxHeroTag(messageImageContext: messageImageContext, src: src),
```
gnprice, line 72: Is the `message` field on this class used for anything after this change?

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +203,136 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final heroAnimationDuration = const Duration(milliseconds: 300);
+      final steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      List<Rect> imagePositions = [];
+      for (int i = 1; i <= steps; i++) {
+        await tester.pump(stepDuration);
+
+        final imageFinder = find.byWidgetPredicate((widget) =>
+          widget is RealmContentNetworkImage && widget.src == imageSrcUrl
+        ).first;
```
gnprice, line 323: Why `.first`?

The purpose of this test is to check that there isn't an image animating around. What if there's one copy that's animating around, and another that's staying in the initial position? The one that stays in the initial position could very well happen to be the first one found in the tree, and then this test would wrongly pass.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +203,136 @@ class FakeVideoPlayerPlatform extends Fake
...
+      for (final position in animatedPositions) {
+        final positionChangeTop = position.top - previousPosition.top;
+        final positionChangeLeft = position.left - previousPosition.left;
+        check(positionChangeTop.abs()).isLessThan(0.1 * totalTopDistance);
+        check(positionChangeLeft.abs()).isLessThan(0.1 * totalLeftDistance);
+        check(position).not((pos) => pos.equals(previousPosition));
+
+        previousPosition = position;
+      }
+      await tester.pumpAndSettle();
```
gnprice, line 281: If this is needed, that suggests that the animation isn't actually over when this test stops checking.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +203,136 @@ class FakeVideoPlayerPlatform extends Fake
...
+        initialImagePosition.top - animatedPositions.last.top).abs();
+      final totalLeftDistance = (
+        initialImagePosition.left - animatedPositions.last.left).abs();
+
+      Rect previousPosition = initialImagePosition;
+      for (final position in animatedPositions) {
+        final positionChangeTop = position.top - previousPosition.top;
+        final positionChangeLeft = position.left - previousPosition.left;
+        check(positionChangeTop.abs()).isLessThan(0.1 * totalTopDistance);
+        check(positionChangeLeft.abs()).isLessThan(0.1 * totalLeftDistance);
```
gnprice, line 276: 10% of the distance traveled in the whole animation is quite a lot for a duration of 1/150th of the total duration.

We can measure this better by using the Pythagorean theorem to compute the overall distance moved in the two-dimensional plane, rather than looking at motion in each axis separately. That way if the start and end points happen to be close in one axis, this doesn't expect unreasonably tight bounds on how much the image might move in that axis during the animation.

gnprice, line 276: Instead of checking each of these separately in a loop against the same upper bound, it's better to compute the max across all the steps and then check the comparison of just that max against the upper bound.

That will have the same behavior as to whether the test passes or fails. But if it fails, it will give more informative output: it'll print the number from the one step where the image moved the most, instead of just the one that happened to come first out of the steps that were over the bound. That's especially helpful when picking what the bound should be.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +203,136 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final stepDuration = heroAnimationDuration ~/ steps;
+      List<Rect> animatedPositions = [];
+      for (int i = 1; i <= steps; i++) {
+        await tester.pump(stepDuration);
+
+        final animatedFlightImageFinder = find.byWidgetPredicate((widget) =>
+          widget is RealmContentNetworkImage && widget.src == imageSrcUrl
+        );
+        final animatedFlightImagePosition = tester.getRect(animatedFlightImageFinder);
+        animatedPositions.add(animatedFlightImagePosition);
```
gnprice, line 263: This can be simpler and shorter by inlining.

gnprice (commented):
Thanks for the revision. This is significantly closer; some comments below.

test/widgets/lightbox_test.dart
lakshya1goel: Here I am getting 2 exactly same images, I have checked the position of both the images inside the loop for 150 steps using the logs and the position is same for both the images for a particular step.
This is probably due to the navigation, on the press of back button the top most route is being popped and we are moving to the older route so at a particular instance we are having the image due to both of these routes.

lib/widgets/lightbox.dart
lakshya1goel: yeah, its not required now.

test/widgets/lightbox_test.dart
gnprice: Yeah, that sounds like the behavior we expect to see here.

The job of a test is to convince a skeptical reader that the behavior we expect to see is the behavior that's actually happening (given that the test passes). So in particular this test should be written so that it rules out the possibility I described above, and other alternative possibilities of that flavor.

gnprice:
Thanks for the revision. This looks good now, except that the test needs to be more specific as described in this subthread above: https://github.com/zulip/zulip-flutter/pull/1348#discussion_r2019569770

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,146 @@ class FakeVideoPlayerPlatform extends Fake
...
+        }
+      }
+
+      for (final element in imagePositionsOverTime.keys) {
+        final positions = imagePositionsOverTime[element]!;
+        check(positions).isNotEmpty();
+
+        final initialPos = positions[0];
+        for (final position in positions) {
+          check(position).equals(initialPos);
```
gnprice, line 339: This still isn't very specific, because it doesn't check that the set of elements at one step has any relationship to the set of elements at another step.

For example, one plausible buggy behavior would be that at each step, there's a new image element and the old image element is gone. From the user's perspective this could produce an animation, because the new element could be at a different position along an animation path. But that would pass this test, because each list would have just one entry in it.

gnprice: Instead, try this:
* Before navigating to the second page, find the one image element and remember its location.
* While on the second page, find the one image element that's now visible and remember its location.
* Once the animation starts, at each step, check that the set of image elements is exactly those two image elements.
* Also at each step, check that each image element is in the same place as it was at the start.

Does that pass? If so, great — that's very specific and makes clear what the behavior is.

If not, then loosen the conditions a little in order to make it pass. But experiment to find just how and why the specific version fails, so that you can loosen it in only the minimal way that makes it pass, keeping it as specific as possible.

lakshya1goel:
Hi, @gnprice just pushed the revision following the comments above. The test is passing without loosening the constraints.
PTAL, Thanks!

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,153 @@ class FakeVideoPlayerPlatform extends Fake
...
+
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+
+        final currentImages = find.byWidgetPredicate((widget) =>
+          widget is RealmContentNetworkImage && widget.src == imageSrcUrl).evaluate();
+        check(currentImages)
+          ..length.equals(2)
+          ..contains(firstElement)
+          ..contains(secondElement);
```
gnprice, line 335: Fun shorthand for this:
```suggestion
        check(currentImages).unorderedEquals([firstElement, secondElement]);
```

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,153 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      final firstImagePositions = <Rect>[];
+      final secondImagePositions = <Rect>[];
+
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+
+        final currentImages = find.byWidgetPredicate((widget) =>
+          widget is RealmContentNetworkImage && widget.src == imageSrcUrl).evaluate();
```
gnprice, line 331: This is the same finder as `secondImageFinder`, right? It's really "find images with this URL". So it'd be good to combine them, under a name that reflects that shared meaning.

They can also replace `firstImageFinder`, if we just identify the URL in advance.

And we can do that, because this test controls what the URL is going to be — it set up the message list and supplied the message HTML.

Probably cleanest is to just duplicate the URL from that ContentExample value. It's unlikely to change; if it does change, the tests will fail because they won't find elements they expect. So like:
```dart
    final message = eg.streamMessage(stream: channel,
      contentMarkdown: ContentExample.imageSingle.html, topic: 'test topic');
    
    // From ContentExample.imageSingle.
    final imageSrcUrlStr = '/user_uploads/2/ce/nvoNL2LaZOciwGZ-FYagddtK/image.jpg';
    final imageSrcUrl = …;
    final imageFinder = …;
```

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,153 @@ class FakeVideoPlayerPlatform extends Fake
...
+      await tester.pumpAndSettle();
+
+      final secondImageFinder = find.byWidgetPredicate((widget) =>
+        widget is RealmContentNetworkImage && widget.src == imageSrcUrl
+      );
+      final secondImagePosition = tester.getRect(secondImageFinder);
+      final secondElement = secondImageFinder.evaluate().single;
+
+      final backButton = find.byType(BackButton);
+      await tester.tap(backButton);
```
gnprice, line 318: nit: simpler if inlined

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,153 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final firstImagePosition = tester.getRect(firstImageFinder);
+      final firstElement = firstImageFinder.evaluate().single;
+
+      connection.prepare(json:
+        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
+
+      await tester.tap(find.descendant(
+        of: find.byType(StreamMessageRecipientHeader),
+        matching: find.text('test topic')));
+      await tester.pumpAndSettle();
```
gnprice, line 309: nit: these are logically one operation, so join as one stanza:
```suggestion
      connection.prepare(json:
        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
      await tester.tap(find.descendant(
        of: find.byType(StreamMessageRecipientHeader),
        matching: find.text('test topic')));
      await tester.pumpAndSettle();
```

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,153 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final initialImagePosition = tester.getRect(messageListImageFinder);
+      await tester.tap(messageListImageFinder);
+      await tester.pump();
+      // pump to start hero animation
+      await tester.pump();
+
+      const heroAnimationDuration = Duration(milliseconds: 300);
+      const steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      List<Rect> animatedPositions = [];
```
gnprice, line 256: nit: final

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,153 @@ class FakeVideoPlayerPlatform extends Fake
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  group('LightboxHero', () {
+    late PerAccountStore store;
+    late FakeApiConnection connection;
+
+    final channel = eg.stream();
+    final message = eg.streamMessage(stream: channel,
+      contentMarkdown: ContentExample.imageSingle.html, topic: 'test topic');
```
gnprice, line 213: nit: order elements logically: so stream, topic, then content, because the stream and topic give the more general context where the content appears

lakshya1goel:
Hi, @gnprice I have pushed the revision. PTAL, Thanks!

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  group('LightboxHero', () {
+    late PerAccountStore store;
+    late FakeApiConnection connection;
+
+    final channel = eg.stream();
+    final imageSrcUrlStr = 'https://chat.example/user_uploads/thumbnail/2/ce/nvoNL2LaZOciwGZ-FYagddtK/image.jpg/840x560.webp';
```
gnprice, line 212: nit: this should have a comment saying where it comes from, as it otherwise looks quite random — cf https://github.com/zulip/zulip-flutter/pull/1348#discussion_r2029628420

test/widgets/lightbox_test.dart
gnprice: `const` (which is stronger than `final`) was fine for the locals where it applied — this comment was meant for `animatedPositions` :slightly_smiling_face: 

(In GitHub's review interface, if you make a comment targeting just a single line — like this one was — it shows the 4 lines up through that one as the context. That does makes it look a bit ambiguous, though in cases where the difference matters it is possible to find if you look closely: some comments say "Comment on lines MMM to NNN" above them.)

The principle here is that when a value is marked as immutable, that makes the code easier for the reader to reason about. So whenever we don't actually need to mutate it (like we do for loop variables like `previousPosition`), it's best to declare it `final` or `const`.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
...
+      await setupMessageListPage(tester);
+
+      final messageContentFinder = find.byWidgetPredicate((widget) =>
+        widget is MessageContent && widget.message.id == message.id
+      );
+      final messageListImageFinder = find.descendant(
+        of: messageContentFinder,
+        matching: find.byType(RealmContentNetworkImage)
+      );
+      final initialImagePosition = tester.getRect(messageListImageFinder);
```
gnprice, line 247: This can be simplified to work the same way as the corresponding part of the other test case. (And then `imageFinder` can be shared between the test cases.)

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+
+        final currentImages = imageFinder.evaluate();
+        check(currentImages).unorderedEquals([firstElement, secondElement]);
+
+        check(tester.getRect(
+          find.byElementPredicate((e) => e == firstElement))).equals(firstImagePosition);
```
gnprice, line 320: nit: line too long; instead:
```suggestion
        check(tester.getRect(find.byElementPredicate((e) => e == firstElement)))
          .equals(firstImagePosition);
```

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
...
+      await tester.pump();
+
+      final heroAnimationDuration = Duration(milliseconds: 300);
+      final steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+
+        final currentImages = imageFinder.evaluate();
+        check(currentImages).unorderedEquals([firstElement, secondElement]);
```
gnprice, line 317: nit: simpler if inlined

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
...
+    testWidgets('no hero animation occurs between different message list pages for same image', (tester) async {
+      prepareBoringImageHttpClient();
+
+      await setupMessageListPage(tester);
+
+      final imageFinder = find.byWidgetPredicate((widget) =>
+        widget is RealmContentNetworkImage && widget.src == imageSrcUrl
+      );
+      final firstImagePosition = tester.getRect(imageFinder);
+      final firstElement = imageFinder.evaluate().single;
```
gnprice, line 295: Bit cleaner (as `evaluate` is meant really as internal API for implementing these finders):
```suggestion
      final firstElement = tester.element(imageFinder);
```

test/widgets/lightbox_test.dart
gnprice: Better yet, make a little helper function within this test case, and then:
```suggestion
        check(getElementRect(firstElement)).equals(firstImagePosition);
```

Then the same `getElementRect` can be used for defining `firstImagePosition` in terms of `firstElement` in the first place. That way this check works the same way as `firstImagePosition` got computed originally — so we're doing the same thing the same way both times, instead of two somewhat different ways, and that means one fewer thing for the reader to understand and think through.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
...
+      final totalDistance = sqrt(
+        pow(initialImagePosition.top - animatedPositions.last.top, 2) +
+        pow(initialImagePosition.left - animatedPositions.last.left, 2));
+
+      Rect previousPosition = initialImagePosition;
+      double maxStepDistance = 0.0;
+      for (final position in animatedPositions) {
+        final stepDistance = sqrt(
+          pow(position.top - previousPosition.top, 2) +
+          pow(position.left - previousPosition.left, 2));
```
gnprice, line 275: The logic here can become a bit easier to see by making a small helper function inside the test case:
```suggestion
        final stepDistance = dist(previousPosition, position);
```

Particularly for comparison with `totalDistance` above.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,128 @@ class FakeVideoPlayerPlatform extends Fake
...
+      connection = store.connection as FakeApiConnection;
+      await store.addUser(eg.selfUser);
+
+      connection.prepare(json:
+        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
+
+      await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
+        child: MessageListPage(initNarrow: const CombinedFeedNarrow())));
+
+      await tester.pumpAndSettle();
```
gnprice, line 232: nit: these are basically one operation; join them as one stanza:
```suggestion

      connection.prepare(json:
        eg.newestGetMessagesResult(foundOldest: true, messages: [message]).toJson());
      await tester.pumpWidget(TestZulipApp(accountId: eg.selfAccount.id,
        child: MessageListPage(initNarrow: const CombinedFeedNarrow())));
      await tester.pumpAndSettle();
```

gnprice (commented):
Thanks for the revision! This is getting closer; here's comments on this version.

(This review is a bit delayed because I was out sick part of last week.)

lakshya1goel:
Hi, I have pushed the revision. PTAL @gnprice. Thanks!

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,112 @@ class FakeVideoPlayerPlatform extends Fake
...
+
+      await tester.tap(find.byType(BackButton));
+      await tester.pump();
+
+      const heroAnimationDuration = Duration(milliseconds: 300);
+      const steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+        check(imageFinder.evaluate()).unorderedEquals([firstElement, secondElement]);
```
gnprice, line 304: nit: similar comment as: https://github.com/zulip/zulip-flutter/pull/1348#discussion_r2045447107

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,112 @@ class FakeVideoPlayerPlatform extends Fake
...
+
+    final channel = eg.stream();
+    // From ContentExample.imageSingle.
+    final imageSrcUrlStr = 'https://chat.example/user_uploads/thumbnail/2/ce/nvoNL2LaZOciwGZ-FYagddtK/image.jpg/840x560.webp';
+    final imageSrcUrl = Uri.parse(imageSrcUrlStr);
+    final message = eg.streamMessage(stream: channel,
+      topic: 'test topic', contentMarkdown: ContentExample.imageSingle.html);
+
+    final imageFinder = find.byWidgetPredicate(
+      (widget) => widget is RealmContentNetworkImage && widget.src == imageSrcUrl);
```
gnprice, line 219: nit: this is clearer if rearranged (like at https://github.com/zulip/zulip-flutter/pull/1348#discussion_r2029628420):
```suggestion
    final channel = eg.stream();
    final message = eg.streamMessage(stream: channel,
      topic: 'test topic', contentMarkdown: ContentExample.imageSingle.html);

    // From ContentExample.imageSingle.
    final imageSrcUrlStr = 'https://chat.example/user_uploads/thumbnail/2/ce/nvoNL2LaZOciwGZ-FYagddtK/image.jpg/840x560.webp';
    final imageSrcUrl = Uri.parse(imageSrcUrlStr);
    final imageFinder = find.byWidgetPredicate(
      (widget) => widget is RealmContentNetworkImage && widget.src == imageSrcUrl);
```

There's basically two narratives (or mini-narratives) here: setting up `message`, and setting up `imageFinder`. This way the reader gets through one, then goes through the other, rather than interleaving them.

This way also makes the comment on `imageSrcUrlStr` make more sense — the reason `ContentExample.imageSingle` is relevant is that it's what's used in `message`. When this comes shortly after `message` rather than before it, the reader has that context.

gnprice (commented):
Thanks for the revision! Down now to a couple of small comments, both of them echoes of comments from the last round.

test/widgets/lightbox_test.dart
```diff
@@ -197,6 +204,115 @@ class FakeVideoPlayerPlatform extends Fake
...
+
+      const heroAnimationDuration = Duration(milliseconds: 300);
+      const steps = 150;
+      final stepDuration = heroAnimationDuration ~/ steps;
+      for (int i = 0; i < steps; i++) {
+        await tester.pump(stepDuration);
+        check(tester.element(imageFinder.first)).anyOf([
+          (e) => e.equals(firstElement), (e) => e.equals(secondElement)]);
+        check(tester.element(imageFinder.last)).anyOf([
+          (e) => e.equals(firstElement), (e) => e.equals(secondElement)]);
```
gnprice, line 307: This no longer checks that the situation is the way we expect: it allows there to be three or more elements matching `imageFinder`.

It's also a lot more code to read than an `unorderedEquals` check. (Cf. my previous comment https://github.com/zulip/zulip-flutter/pull/1348#discussion_r2029614529 .)

Using `unorderedEquals` is the right way to do this. You'll want something different from `tester.element`, though.

Take a look around the API surrounding `tester.element` to see what variations of it you can find that would be useful here. I recommend jumping in your IDE to the definition of `tester.element` — that will let you browse around the neighboring code.

pubspec.lock
```diff
@@ -650,26 +650,26 @@ packages:
     dependency: transitive
     description:
       name: leak_tracker
-      sha256: "6bb818ecbdffe216e81182c2f0714a2e62b593f4a4f13098713ff1685dfb6ab0"
+      sha256: "8dcda04c3fc16c14f48a7bb586d4be1f0d1572731b6d81d51772ef47c02081e0"
```
gnprice, line 653: This diff doesn't seem related to what this commit is doing. To keep the commits [clear and coherent](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html), this change should be kept to a separate commit.

It should also go with a corresponding change in `pubspec.yaml`. See our README about upgrading Flutter.

lakshya1goel:
Hi @gnprice, I have pushed the correction please have a look. Thanks!

gnprice (commented):
Thanks @lakshya1goel for all your work on this!

The tests all look good now, modulo one nit.

See also one more-substantive comment below. But I think it doesn't call for any code changes, only comments and metadata.

So I'll make those edits and merge. I'll also add a commit that revises the docs from this PR.

From reading the TODO comment that I remarked on in an earlier review, I see that this in fact fixes not only #930 but also another issue #44. A good followup for you would be to do next would be to write a regression test to ensure that #44 stays fixed; see details below.

Issue #1264, opened 2025-01-08
Don't show notifications for logged-out accounts, on Android


gnprice:
When we get a notification-message from a Zulip server, but it turns out to be for an account that isn't actually in our list of accounts, we should skip showing any actual notification to the user.

The situation where this can happen is when the user has logged out of an account (by removing it from the list of accounts). When they do, we make a request to the server to stop sending notifications for that account to this device; see [implementation](https://github.com/zulip/zulip-flutter/blob/0b90e421dbff8b35fee3e8768cbf0e6d2f0cccba/lib/widgets/actions.dart#L30-L50). But that request can fail — for example, maybe the device doesn't happen to have an Internet connection at the time the user removes the account. In that case the server will keep sending notifications.

It's annoying to the user if notifications keep showing up after they've logged out. (Also alarming: it suggests the logout didn't fully work.) The user didn't want those notifications, so we should avoid showing them.

Right now this will only benefit Android, because that's where we have our own code involved between the device receiving a notification-message and a notification getting shown in the UI.

Reported today [in chat](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20beta.3A.20Notification.20when.20logged.20out.20and.20during.20DND/near/2033294).

### Implementation

In NotificationDisplayManager, when a MessageFcmMessage comes in, we should look up the corresponding account; if none is found, don't show anything.

Also when logging out we should remove any existing notifications for that account from the UI.

### Related issues

* Previous report for the zulip-mobile legacy app, with some discussion:
  https://github.com/zulip/zulip-mobile/issues/5763

  (This would have been much harder to fix in the legacy app than in this Flutter app, because notifications were handled there in Kotlin code that didn't have access to the app's stored data.)

* We should fix this symptom on iOS too. I think the way to do that will be to start having our own code involved in deciding what notification to show in the UI, just like it is for Android:
  #1265
  (We'll need that for some other nice notification features, too.)

* Separately, we should retry the "stop sending notifications" request later if it didn't succeed the first time, so that the server stops sending notification-messages that aren't going to be shown to the user.

  This is tricky security-wise, because we don't want to hold onto the user's API key after they've logged out. Doing this may require server-side changes, so that we can hold onto some credential that carries only the ability to do this one request.

tomlin7:
Hi, I was able to implement the first part (not showing notifications for logged out accounts), but my logic for second (removing existing notifications of an account upon logout) is failing atm.

[...]

PR #1317, opened 2025-02-02
action_sheet: Offer "Mark channel as read" in channel action sheet

chimnayajith:
# Pull Request

## Description

Adds a channel action sheet with a "Mark channel as read" option, accessible via long-press from both the inbox and subscription list pages. The action sheet is triggered by long-pressing on a channel header.

## Related Issues

- Closes #1226 

## Additional Information
Currently the action sheet only appears in subscription list when the channel has unread messages, since the sheet doesn't have any default buttons.

Issue #1265, opened 2025-01-08
ios notif: Control the showing of notifications


gnprice:
When a notification-message comes in from the server on iOS, we should have some of our own Dart code run, and have that code control the notification that gets shown in the UI.

That's something we already do on Android: see NotificationDisplayManager. Most of the code of that class is specific to Android's notification API, so some refactoring and/or renaming will be needed.

For this issue, there doesn't need to be any change to how the notifications look or behave. If the notifications appear exactly the same way they did before, that's success.

One immediate way we can make use of having this control is to do the equivalent of #1264: check if the account is still logged in, and if not then suppress the notification.

### Related issues

* An easy thing we can do as soon as we have this: the iOS equivalent of
  #1264

* We can also use this control to improve the wording and formatting of notifications, similar to things we already do for Android notifications. (As far as possible, those changes should happen as separate commits or separate PRs, after the commit that resolves this issue.)

* Other issues this might help us with:
  #153 
  #408

jefferypalmer:
@gnprice We have just built and tested the latest tag v30.0.264 to evaluate the notifications behavior on iOS. We have enabled the server side code to push a badge update on receipt of a new message to a user, and generally show a badge with the unread message count. This works great as is to keep unread message counts on the app, but, the new flutter app doesn't clear the badge. I believe what we did previously is to simply clear the badge when the app opens, which doesn't show a true unread message count until the server side pushes another notification. Sharing this with you to get your input and feelings on a more comprehensive solution in place for full server/client side message count accuracy, or we utilize a bit of a hack for now just by clearing the badge when the app opens?

gnprice:
@jefferypalmer Thanks for the feedback! It sounds like the feature you're talking about is #153. And to close the loop, it looks like you found the chat thread for it: [#mobile-team > iOS badge count @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/iOS.20badge.20count/near/2250105).

PR #1349, opened 2025-02-12
notif: Ignore notifications for logged out accounts

tomlin7:
Fixes: #1264

## Changes
- Ignores notifications received for accounts that were logged out while the app was offline
- Removes notifications (if there exist) of the user when the user logs out

### DIFF
--- a/lib/model/actions.dart
+++ b/lib/model/actions.dart
@@ -1,5 +1,8 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
+
+import '../notifications/display.dart';
 import '../notifications/receive.dart';
 import 'store.dart';
 
@@ -10,6 +13,10 @@ Future<void> logOutAccount(GlobalStore globalStore, int accountId) async {
 
   // Unawaited, to not block removing the account on this request.
   unawaited(unregisterToken(globalStore, accountId));
+
+  if (defaultTargetPlatform == TargetPlatform.android) {
+    unawaited(NotificationDisplayManager.removeNotificationsForAccount(account.realmUrl, account.userId));
+  }
 
   await globalStore.removeAccount(accountId);
 }
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -1,10 +1,10 @@
 import 'dart:async';
 import 'dart:io';
 
-import 'package:http/http.dart' as http;
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart' hide Notification;
+import 'package:http/http.dart' as http;
 
 import '../api/model/model.dart';
 import '../api/notifications.dart';
@@ -231,7 +231,7 @@ class NotificationDisplayManager {
   static Future<void> _onMessageFcmMessage(MessageFcmMessage data, Map<String, dynamic> dataJson) async {
     assert(debugLog('notif message content: ${data.content}'));
     final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
-    final groupKey = _groupKey(data);
+    final groupKey = _groupKey(data.realmUrl, data.userId);
     final conversationKey = _conversationKey(data, groupKey);
 
     final oldMessagingStyle = await _androidHost
@@ -365,7 +365,7 @@ class NotificationDisplayManager {
     // There may be a lot of messages mentioned here, across a lot of
     // conversations.  But they'll all be for one account, so they'll
     // fall under one notification group.
-    final groupKey = _groupKey(data);
+    final groupKey = _groupKey(data.realmUrl, data.userId);
 
     // Find any conversations we can cancel the notification for.
     // The API doesn't lend itself to removing individual messages as
@@ -445,10 +445,10 @@ class NotificationDisplayManager {
     return '$groupKey|$conversation';
   }
 
-  static String _groupKey(FcmMessageWithIdentity data) {
+  static String _groupKey(Uri realmUrl, int userId) {
     // The realm URL can't contain a `|`, because `|` is not a URL code point:
     //   https://url.spec.whatwg.org/#url-code-points
-    return "${data.realmUrl}|${data.userId}";
+    return "$realmUrl|$userId";
   }
 
   static String _personKey(Uri realmUrl, int userId) => "$realmUrl|$userId";
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -234,6 +234,18 @@ class NotificationDisplayManager {
     final groupKey = _groupKey(data.realmUrl, data.userId);
     final conversationKey = _conversationKey(data, groupKey);
 
+    final globalStore = await ZulipBinding.instance.getGlobalStore();
+    final account = globalStore.accounts.firstWhereOrNull((account) =>
+      account.realmUrl.origin == data.realmUrl.origin && account.userId == data.userId);
+
+    // Skip showing notifications for a logged-out account. This can occur if
+    // the unregisterToken request failed previously. It would be annoying
+    // to the user if notifications keep showing up after they've logged out.
+    // (Also alarming: it suggests the logout didn't fully work.)
+    if (account == null) {
+      return;
+    }
+
     final oldMessagingStyle = await _androidHost
       .getActiveNotificationMessagingStyleByTag(conversationKey);
 
@@ -418,6 +430,18 @@ class NotificationDisplayManager {
       // during creation, the summary notification doesn't get auto canceled if
       // child notifications are canceled programatically as done above.
       await _androidHost.cancel(tag: groupKey, id: kNotificationId);
+    }
+  }
+
+  static Future<void> removeNotificationsForAccount(Uri realmUrl, int userId) async {
+    final groupKey = _groupKey(realmUrl, userId);
+    final activeNotifications = await _androidHost.getActiveNotifications(
+      desiredExtras: []);
+    for (final statusBarNotification in activeNotifications) {
+      if (statusBarNotification.notification.group == groupKey) {
+        await _androidHost.cancel(
+          tag: statusBarNotification.tag, id: statusBarNotification.id);
+      }
     }
   }
 
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -217,10 +217,12 @@ class NotificationChannelManager {
 /// Service for managing the notifications shown to the user.
 class NotificationDisplayManager {
   static Future<void> init() async {
+    assert(defaultTargetPlatform == TargetPlatform.android);
     await NotificationChannelManager.ensureChannel();
   }
 
   static void onFcmMessage(FcmMessage data, Map<String, dynamic> dataJson) {
+    assert(defaultTargetPlatform == TargetPlatform.android);
     switch (data) {
       case MessageFcmMessage(): _onMessageFcmMessage(data, dataJson);
       case RemoveFcmMessage(): _onRemoveFcmMessage(data);
@@ -434,6 +436,8 @@ class NotificationDisplayManager {
   }
 
   static Future<void> removeNotificationsForAccount(Uri realmUrl, int userId) async {
+    assert(defaultTargetPlatform == TargetPlatform.android);
+
     final groupKey = _groupKey(realmUrl, userId);
     final activeNotifications = await _androidHost.getActiveNotifications(
       desiredExtras: []);
@@ -488,6 +492,8 @@ class NotificationDisplayManager {
     required BuildContext context,
     required Uri url,
   }) {
+    assert(defaultTargetPlatform == TargetPlatform.android);
+
     final globalStore = GlobalStoreWidget.of(context);
 
     assert(debugLog('got notif: url: $url'));
@@ -516,6 +522,7 @@ class NotificationDisplayManager {
   /// generated with [NotificationOpenPayload.buildUrl] while creating
   /// the notification.
   static Future<void> navigateForNotification(Uri url) async {
+    assert(defaultTargetPlatform == TargetPlatform.android);
     assert(debugLog('opened notif: url: $url'));
 
     NavigatorState navigator = await ZulipApp.navigator;
--- a/test/model/actions_test.dart
+++ b/test/model/actions_test.dart
@@ -1,7 +1,11 @@
+import 'dart:io';
+
 import 'package:checks/checks.dart';
+import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:http/http.dart' as http;
+import 'package:http/testing.dart' as http_testing;
 import 'package:zulip/model/actions.dart';
 import 'package:zulip/model/store.dart';
 import 'package:zulip/notifications/receive.dart';
@@ -12,7 +16,9 @@ import '../fake_async.dart';
 import '../model/binding.dart';
 import '../model/store_checks.dart';
 import '../model/test_store.dart';
+import '../notifications/display_test.dart';
 import '../stdlib_checks.dart';
+import '../test_images.dart';
 import 'store_test.dart';
 
 void main() {
@@ -20,6 +26,24 @@ void main() {
 
   late PerAccountStore store;
   late FakeApiConnection connection;
+
+  http.Client makeFakeHttpClient({http.Response? response, Exception? exception}) {
+    return http_testing.MockClient((request) async {
+      assert((response != null) ^ (exception != null));
+      if (exception != null) throw exception;
+      return response!; // TODO return 404 on non avatar urls
+    });
+  }
+
+  final fakeHttpClientGivingSuccess = makeFakeHttpClient(
+    response: http.Response.bytes(kSolidBlueAvatar, HttpStatus.ok));
+
+  T runWithHttpClient<T>(
+    T Function() callback, {
+    http.Client Function()? httpClientFactory,
+  }) {
+    return http.runWithClient(callback, httpClientFactory ?? () => fakeHttpClientGivingSuccess);
+  }
 
   Future<void> prepare({String? ackedPushToken = '123'}) async {
     addTearDown(testBinding.reset);
@@ -120,6 +144,24 @@ void main() {
 
       async.elapse(unregisterDelay - TestGlobalStore.removeAccountDuration);
       check(newConnection.isOpen).isFalse();
+    }));
+
+    test('notifications are removed after logout', () => awaitFakeAsync((async) async {
+      await prepare();
+      testBinding.firebaseMessagingInitialToken = '123';
+      addTearDown(NotificationService.debugReset);
+      NotificationService.debugBackgroundIsolateIsLive = false;
+      await runWithHttpClient(NotificationService.instance.start);
+
+      // Create a notification to check that it's removed after logout
+      final message = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
+      testBinding.firebaseMessaging.onMessage.add(
+        RemoteMessage(data: messageFcmMessage(message).toJson()));
+      async.flushMicrotasks();
+      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
+
+      await logOutAccount(testBinding.globalStore, eg.selfAccount.id);
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
     }));
   });
 
--- a/test/notifications/display_test.dart
+++ b/test/notifications/display_test.dart
@@ -26,9 +26,9 @@ import 'package:zulip/widgets/message_list.dart';
 import 'package:zulip/widgets/page.dart';
 import 'package:zulip/widgets/theme.dart';
 
+import '../example_data.dart' as eg;
 import '../fake_async.dart';
 import '../model/binding.dart';
-import '../example_data.dart' as eg;
 import '../model/narrow_checks.dart';
 import '../stdlib_checks.dart';
 import '../test_images.dart';
@@ -114,7 +114,10 @@ void main() {
     return http.runWithClient(callback, httpClientFactory ?? () => fakeHttpClientGivingSuccess);
   }
 
-  Future<void> init() async {
+  Future<void> init({bool addSelfAccount = true}) async {
+    if (addSelfAccount) {
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    }
     addTearDown(testBinding.reset);
     testBinding.firebaseMessagingInitialToken = '012abc';
     addTearDown(NotificationService.debugReset);
@@ -872,7 +875,8 @@ void main() {
     })));
 
     test('remove: different realm URLs but same user-ids and same message-ids', () => runWithHttpClient(() => awaitFakeAsync((async) async {
-      await init();
+      await init(addSelfAccount: false);
+
       final stream = eg.stream();
       const topic = 'Some Topic';
       final conversationKey = 'stream:${stream.streamId}:some topic';
@@ -881,6 +885,7 @@ void main() {
         realmUrl: Uri.parse('https://1.chat.example'),
         id: 1001,
         user: eg.user(userId: 1001));
+      await testBinding.globalStore.add(account1, eg.initialSnapshot());
       final message1 = eg.streamMessage(id: 1000, stream: stream, topic: topic);
       final data1 =
         messageFcmMessage(message1, account: account1, streamName: stream.name);
@@ -890,6 +895,7 @@ void main() {
         realmUrl: Uri.parse('https://2.chat.example'),
         id: 1002,
         user: eg.user(userId: 1001));
+      await testBinding.globalStore.add(account2, eg.initialSnapshot());
       final message2 = eg.streamMessage(id: 1000, stream: stream, topic: topic);
       final data2 =
         messageFcmMessage(message2, account: account2, streamName: stream.name);
@@ -917,19 +923,21 @@ void main() {
     })));
 
     test('remove: different user-ids but same realm URL and same message-ids', () => runWithHttpClient(() => awaitFakeAsync((async) async {
-      await init();
+      await init(addSelfAccount: false);
       final realmUrl = eg.realmUrl;
       final stream = eg.stream();
       const topic = 'Some Topic';
       final conversationKey = 'stream:${stream.streamId}:some topic';
 
       final account1 = eg.account(id: 1001, user: eg.user(userId: 1001), realmUrl: realmUrl);
+      await testBinding.globalStore.add(account1, eg.initialSnapshot());
       final message1 = eg.streamMessage(id: 1000, stream: stream, topic: topic);
       final data1 =
         messageFcmMessage(message1, account: account1, streamName: stream.name);
       final groupKey1 = '${account1.realmUrl}|${account1.userId}';
 
       final account2 = eg.account(id: 1002, user: eg.user(userId: 1002), realmUrl: realmUrl);
+      await testBinding.globalStore.add(account2, eg.initialSnapshot());
       final message2 = eg.streamMessage(id: 1000, stream: stream, topic: topic);
       final data2 =
         messageFcmMessage(message2, account: account2, streamName: stream.name);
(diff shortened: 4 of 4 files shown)


gnprice:
Thanks @tomlin7 for the contribution! Before we can review this, it will need a test — see the repo's README file.

lib/model/actions.dart
```diff
@@ -11,6 +12,8 @@ Future<void> logOutAccount(GlobalStore globalStore, int accountId) async {
   // Unawaited, to not block removing the account on this request.
   unawaited(unregisterToken(globalStore, accountId));
 
+  await NotificationDisplayManager.removeNotificationsForAccount(account.realmUrl, account.userId);
```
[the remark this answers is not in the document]
chrisbobbe: Also, what's the effect when this runs on iOS? Does it error or do nothing?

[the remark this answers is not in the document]
chrisbobbe: Also let's have this be `unawaited`, like `unregisterToken` above; we still want to do the `removeAccount`, as soon as possible, if this fails or takes a long time for some reason.

chrisbobbe (commented):
Thanks! Comments below from a brief review.

lib/notifications/display.dart
```diff
@@ -518,6 +525,16 @@ class NotificationDisplayManager {
     }
     return null;
   }
+
+  static Future<void> removeNotificationsForAccount(Uri realmUri, int userId) async {
+    final groupKey = "$realmUri|$userId";
```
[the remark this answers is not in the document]
tomlin7: @chrisbobbe but the current implementation of `_groupKey()` takes a subtype of `FcmMessageWithIdentity` and I don't have sufficient data at this point. What should be done?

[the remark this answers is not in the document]
chrisbobbe: I see two options:

1. Change `_groupKey` to take a realm URL and user ID instead of an `FcmMessageWithIdentity`.
2. Keep `_groupKey`'s signature but add a separate helper that takes a realm URL and user ID; have `_groupKey` and this new function call that new helper.

Option 1 is less code; let's try that for now, and perhaps Greg will have thoughts in his review.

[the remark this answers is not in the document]
tomlin7: @chrisbobbe I noticed `_personKey` does exactly what I needed here, and it works fine. What do you think about using it instead?

lib/model/actions.dart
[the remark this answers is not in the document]
tomlin7: @chrisbobbe The tests are passing, I can't test on iOS atm since I'm on Android :)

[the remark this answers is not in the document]
chrisbobbe: Yeah; I see this output in the console when I log out on iOS:

```dart
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: PlatformException(channel-error, Unable to establish connection on channel: "dev.flutter.pigeon.zulip.AndroidNotificationHostApi.getActiveNotifications"., null, null)
#0      AndroidNotificationHostApi.getActiveNotifications (package:zulip/host/android_notifications.g.dart:721:7)
<asynchronous suspension>
#1      NotificationDisplayManager.removeNotificationsForAccount (package:zulip/notifications/display.dart:531:33)
<asynchronous suspension>
```

I'll start a CZO discussion on what to do for iOS here.

lib/notifications/display.dart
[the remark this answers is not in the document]
chrisbobbe: Hmm, I'd appreciate @gnprice's thoughts on that question.

lib/model/actions.dart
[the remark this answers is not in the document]
chrisbobbe: Following that discussion, yeah, we will need to put the new call inside an Android platform conditional.

Then as an optional, independent improvement, you could add a commit that does this:

> Could be helpful to put asserts at the top of [`NotificationDisplayManager`'s] public methods, checking the target platform is indeed Android — because that's basically an assumption the implementation is making

lib/notifications/display.dart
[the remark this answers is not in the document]
chrisbobbe: I found some 🙂—https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Removing.20notifications.20for.20logged.20out.20account/near/2088813 :

> As a matter of code organization, we'll want to keep centralized in one place the details of how the group key is formatted — the fact that it's the realm URL, then user ID, and separated by `|`. That's currently in this method:
> ```dart
>   static String _groupKey(FcmMessageWithIdentity data) {
>     // The realm URL can't contain a `|`, because `|` is not a URL code point:
>     //   https://url.spec.whatwg.org/#url-code-points
>     return "${data.realmUrl}|${data.userId}";
>   }
> ```
> 
> So you'll want to refactor that a bit so that it can be used by this code path as well as the existing code paths that use it.

I read that as suggesting my option (1) above.

chrisbobbe:
I believe the two questions above that were open now have answers:

- https://github.com/zulip/zulip-flutter/pull/1349#discussion_r1968595783
- https://github.com/zulip/zulip-flutter/pull/1349#discussion_r1968599602

So you should be able to proceed, incorporating those.

tomlin7:
@chrisbobbe I have made the requested changes, please review again!

chrisbobbe:
Quoting my comment https://github.com/zulip/zulip-flutter/pull/1349#discussion_r1968595783:

> Following that discussion, yeah, we will need to put the new call inside an Android platform conditional.

You haven't done this in this revision; please do that. In `logOutAccount`, we should only call `NotificationDisplayManager .removeNotificationsForAccount` if the platform is Android.

> Then as an optional, independent improvement, you could add a commit that does [this](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/NotificationDisplayManager.20Android-only/near/2099196):
> 
> > Could be helpful to put asserts at the top of [`NotificationDisplayManager`'s] public methods, checking the target platform is indeed Android — because that's basically an assumption the implementation is making

I see some added asserts, but this needs to be done in a separate commit. The current revision only has one commit.

tomlin7:
@chrisbobbe I have implemented all the requested changes. Let me know if there are any further improvements needed.

lib/notifications/display.dart
```diff
@@ -217,10 +217,12 @@ class NotificationChannelManager {
 /// Service for managing the notifications shown to the user.
 class NotificationDisplayManager {
   static Future<void> init() async {
+    assert(defaultTargetPlatform == TargetPlatform.android);
```
chrisbobbe, line 220:
```
notif: Ensure NotificationDisplayManager only supports Android
```

Here's an attempt at a more helpful commit message; see what you think:

```
    notif [nfc]: Add Android-only asserts in NotificationDisplayManager methods
    
    NotificationDisplayManager's public methods only get invoked on
    Android, and their implementations basically assume the platform is
    Android. So, add these asserts, to make that assumption clearer and
    to fail early if we accidentally call one of these on iOS.
    
    When we eventually take more hands-on control of notifications on
    iOS, we might either adapt this class to support iOS too, or make a
    new class for iOS and make this one's name and interface clearer
    that it's Android-only.
    
    Discussion:
      https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/NotificationDisplayManager.20Android-only/near/2099196
```

lib/model/actions.dart
```diff
@@ -1,5 +1,8 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
```
chrisbobbe, line 3: notif: Ignore notifications for logged out accounts
    Fixes: #1264

Commit-message nit: add empty line after the summary line

chrisbobbe: Also the commit message should be clear that two things are happening: notifications are ignored for logged-out accounts, and notifications are removed at the time an account is logged out.

chrisbobbe (commented):
Thanks! Comments below.

Also, have you tested this with an Android device or emulator?

test/model/actions_test.dart
```diff
@@ -92,6 +92,10 @@ void main() {
       check(connection.isOpen).isFalse();
       check(newConnection.isOpen).isTrue(); // still busy with unregister-token
 
+      // Check that notifications were removed
+      final activeNotifications = await testBinding.androidNotificationHost.getActiveNotifications(desiredExtras: []);
+      check(activeNotifications).isEmpty();
+
```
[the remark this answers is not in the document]
tomlin7: @chrisbobbe I faced some issues attempting to generate notifications in this scope. Explaining the steps below:

```dart
      // Create a notification to check that it's removed after logout
      final message = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
      final data = messageFcmMessage(message);

      // this won't help
      testBinding.firebaseMessaging.onMessage.add(
        RemoteMessage(data: data.toJson()));
      async.flushMicrotasks();

      testBinding.firebaseMessaging.onBackgroundMessage.add(
        RemoteMessage(data: data.toJson()));
      async.flushMicrotasks();

      // 2. this won't help either
      NotificationDisplayManager.onFcmMessage(data, RemoteMessage(data: data.toJson()).data);

      // Check that notifications were removed
      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
```

modified `prepare` for this change as well:
```dart
  Future<void> prepare({String? ackedPushToken = '123'}) async {
    ...

    testBinding.firebaseMessagingInitialToken = '012abc';
    NotificationService.debugBackgroundIsolateIsLive = false;
  }
```

can you please tell me where I am going wrong.

test/notifications/display_test.dart
```diff
@@ -493,6 +497,9 @@ void main() {
       const topic = 'topic 1';
       final message1 = eg.streamMessage(topic: topic, stream: stream);
       final data1 = messageFcmMessage(message1, streamName: stream.name);
+      final account = eg.account(id: data1.userId, user: eg.user(userId: data1.userId));
+      await testBinding.globalStore.add(account, eg.initialSnapshot());
```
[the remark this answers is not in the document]
tomlin7: @chrisbobbe I tried that, but some tests are failing afterwards that change, so I'm having to add an extra boolean parameter to the `init`

```dart
  Future<void> init({bool addSelfAccount = true}) async {
    if (addSelfAccount) {
      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
    }
    ...
  }
```

tomlin7:
@chrisbobbe I tested this with an Android device (both debug and release).

lib/notifications/display.dart
```diff
@@ -234,6 +234,13 @@ class NotificationDisplayManager {
     final groupKey = _groupKey(data.realmUrl, data.userId);
     final conversationKey = _conversationKey(data, groupKey);
 
+    final globalStore = await ZulipBinding.instance.getGlobalStore();
+    final account = globalStore.accounts.firstWhereOrNull((account) =>
+      account.realmUrl.origin == data.realmUrl.origin && account.userId == data.userId);
+    if (account == null) {
+      return;
+    }
```
chrisbobbe, line 244: This deserves a code comment, explaining that it's about logged-out accounts where the unregister-token request failed for some reason.

chrisbobbe (commented):
Thanks! A few comments from a brief review—also, some tests are failing; please fix those.

tomlin7:
@chrisbobbe It's ready, please review the changes!

I got stuck a bit while generating the notifications in `test/models/actions_test`, but that is resolved now. [CZO](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Removing.20notifications.20for.20logged.20out.20account/near/2127632)[#mobile-dev-help > Removing notifications for logged out account @ 💬](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Removing.20notifications.20for.20logged.20out.20account/near/2127632)

lib/notifications/display.dart
```diff
@@ -231,9 +233,21 @@ class NotificationDisplayManager {
...
+    final globalStore = await ZulipBinding.instance.getGlobalStore();
+    final account = globalStore.accounts.firstWhereOrNull((account) =>
+      account.realmUrl.origin == data.realmUrl.origin && account.userId == data.userId);
+
+    // If the account is null, it means the user is logged out. This can happen
+    // if the unregisterToken request failed previously, leaving the account
+    // in a state where it's no longer valid in our system. In this scenario,
+    // we simply return and do nothing, as there's no active account to process.
+    if (account == null) {
+      return;
```
chrisbobbe, line 248: I'd like this comment to focus more on the experience for the user.

How about this, adapted from the issue description on #1264:

```dart
    // Skip showing notifications for a logged-out account. It would be annoying
    // to the user if notifications keep showing up after they've logged out.
    // (Also alarming: it suggests the logout didn't fully work.)
```

chrisbobbe (commented):
Thanks! One small comment below, and I'll mark for Greg's review.

test/model/actions_test.dart
```diff
@@ -21,12 +27,35 @@ void main() {
...
     addTearDown(testBinding.reset);
     final selfAccount = eg.selfAccount.copyWith(ackedPushToken: Value(ackedPushToken));
     await testBinding.globalStore.add(selfAccount, eg.initialSnapshot());
     store = await testBinding.globalStore.perAccount(selfAccount.id);
     connection = store.connection as FakeApiConnection;
+
+    testBinding.firebaseMessagingInitialToken = '123';
+    addTearDown(NotificationService.debugReset);
+    NotificationService.debugBackgroundIsolateIsLive = false;
+    await NotificationService.instance.start();
```
gnprice, line 58: Let's move this part inside the tests that need it. That way it's clearer how this setup relates to the content of the tests.

That also means that when someone's trying to understand one of the tests in this file that use this `prepare` method and *don't* in fact need this NotificationService setup, they aren't faced with the mystery of whether this is related to those tests.

gnprice:
@rajveermalviya, would you take a look at this and
* check for any interactions with your upcoming #1379 which might mean something should be done differently here (as well as any other points that jump out at you);
* do a round of manual testing to confirm this behaves correctly? (I think the implementation is already very close to what we'll merge; the more substantive comments above are on the tests.)

rajveermalviya (commented):
I tested the following cases and verified they work correctly:
- The notifications being removed when logging-out of the account.
- Receiving a notification from FCM but not displaying it when the account wasn't found in GlobalStore (simulating that the API call to `unregisterToken` failed, by commenting it out).

Also since this PR removes the notification by matching `groupKey`, it should also include the group summary notification and remove it, avoiding a lingering summary notification.

As for the interaction with #1379, apart from little conflicts, there shouldn't be anything that requires changing here, as this PR only checks for the notification group key (to find and remove notifications of the specific account) and there's no change in regarding to that in #1379.  And for iOS this shouldn't matter for now (but remains unresolved), as we don't have a way to control displaying notifications there yet (#1265), and also because #1264 is only about Android.

gnprice:
Thanks @rajveermalviya! Sounds good.

lib/notifications/display.dart
tomlin7: Looks good to me! But also, that doesn't explain the actual situation well, though

gnprice: Can you say more about how that differs from the actual situation?

tomlin7: Surely, the comment focuses on the user's experience, but then hides why we had to check this for logged out accounts in the first place. It's because `unregisterToken` failed due to some reason.

gnprice: Thanks. Rereading the context, I think two things are both helpful to explain here: (a) why this is a situation that can arise (and isn't a sign of something buggy happening before this point, and (b) why just cheerfully doing nothing is an appropriate way to handle it.

So I think one reason Chris made the remark above is likely that this:
> In this scenario, we simply return and do nothing, as there's no active account to process.

can naturally make the reader wonder about question (b): do nothing? is that really what the user wants? we got a notification for them, after all.

I think it'd be quite reasonable to add back (relative to Chris's draft above) a mention of the unregisterToken request failing, though. That helps clarify the answer to question (a).

test/model/actions_test.dart
```diff
@@ -121,6 +145,45 @@ void main() {
...
+      final newConnection = separateConnection()
+        ..prepare(delay: unregisterDelay, json: {'msg': '', 'result': 'success'});
+
+      // Create a notification to check that it's removed after logout
+      final message = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
+      testBinding.firebaseMessaging.onMessage.add(
+        RemoteMessage(data: messageFcmMessage(message).toJson()));
+      async.flushMicrotasks();
+
+      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
```
gnprice, line 168: nit: keep checks joined to the steps that set up the state they're checking
```suggestion
      async.flushMicrotasks();
      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
```

(see https://github.com/zulip/zulip-flutter/pull/1317#discussion_r1980467256 and links from there)

test/model/actions_test.dart
```diff
@@ -121,6 +145,45 @@ void main() {
...
+      testBinding.firebaseMessagingInitialToken = '123';
+      addTearDown(NotificationService.debugReset);
+      NotificationService.debugBackgroundIsolateIsLive = false;
+      await runWithHttpClient(NotificationService.instance.start);
+
+      check(testBinding.globalStore).accountIds.single.equals(eg.selfAccount.id);
+      const unregisterDelay = Duration(seconds: 5);
+      assert(unregisterDelay > TestGlobalStore.removeAccountDuration);
+      final newConnection = separateConnection()
+        ..prepare(delay: unregisterDelay, json: {'msg': '', 'result': 'success'});
```
gnprice, line 160: I think this logically goes after the "create a notification" steps. This step is about preparing the response that `logOutAccount` will get when it unregisters the token.

test/model/actions_test.dart
```diff
@@ -121,6 +145,45 @@ void main() {
...
+      async.elapse(TestGlobalStore.removeAccountDuration);
+      await future;
+      // Account removal not blocked on unregister-token response
+      check(testBinding.globalStore).accountIds.isEmpty();
+      check(connection.isOpen).isFalse();
+      check(newConnection.isOpen).isTrue(); // for the unregister-token request
+
+      async.elapse(unregisterDelay - TestGlobalStore.removeAccountDuration);
+      check(newConnection.isOpen).isFalse();
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
```
gnprice, line 185: Quite a lot of these steps feel like they're repeating the work that the two sibling tests already do.

The main point of this test is that after the `logOutAccount` completes, there should be no notifications. So we need the `logOutAccount` call and the `activeNotifications`/`isEmpty` check. Are any of the steps in between necessary?

The goal of a test like this is to tell a story, and make the story convincing for a skeptical reader. The test code needs to have all the details that are necessary in order for the reader not to say "ah, but maybe such-and-such is happening and that causes the test to pass even though the desired behavior isn't really there". But details that aren't relevant to the story are best left out, to keep the story easier to read and understand.

Here the story is: calling `logOutAccount` caused notifications to get removed. So, for example, the check above on `activeNotifications` being non-empty is needed so that the reader doesn't wonder if maybe the test's setup failed to create any notifications in the first place. But the reader of this test doesn't need to be convinced that, for example, `newConnection` got closed — that's not part of thinking about whether `logOutAccount` really did remove notifications like it should.

test/notifications/display_test.dart
```diff
@@ -955,6 +963,68 @@ void main() {
...
       check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
     })));
+
+    test('removeNotificationsForAccount: removes notifications', () => runWithHttpClient(() => awaitFakeAsync((async) async {
+      await init();
+      final message = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
+
+      receiveFcmMessage(async, messageFcmMessage(message));
+      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
+      await NotificationDisplayManager.removeNotificationsForAccount(
```
gnprice, line 973: nit: while joining checks to the steps that set them up, separate from the steps that come after them (which set up later checks):
```suggestion
      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();

      await NotificationDisplayManager.removeNotificationsForAccount(
```

See related comment above. (There are also other examples below.)

test/notifications/display_test.dart
```diff
@@ -955,6 +963,68 @@ void main() {
...
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      receiveFcmMessage(async, messageFcmMessage(message1, account: account1));
+      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
+
+      check(testBinding.androidNotificationHost.activeNotifications).length.equals(4);
+      await NotificationDisplayManager.removeNotificationsForAccount(realmUrl, account1.userId);
+      check(testBinding.androidNotificationHost.activeNotifications).length.equals(2);
+      check(testBinding.androidNotificationHost.activeNotifications.first.notification.group)
```
gnprice, line 996: nit: use getters on the Subject rather than the value itself:
```suggestion
      check(testBinding.androidNotificationHost.activeNotifications).first.notification.group
```

That produces more-informative error output when the test fails.

test/notifications/display_test.dart
```diff
@@ -955,6 +963,68 @@ void main() {
...
+    })));
+
+    test('removeNotificationsForAccount does nothing if there are no notifications', () => runWithHttpClient(() => awaitFakeAsync((async) async {
+      await init();
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+      await NotificationDisplayManager.removeNotificationsForAccount(eg.selfAccount.realmUrl, eg.selfAccount.userId);
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+    })));
+
   });
```
gnprice, line 1028: nit: no blank line at end of block
```suggestion
    })));
  });
```

test/notifications/display_test.dart
```diff
@@ -955,6 +963,68 @@ void main() {
...
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      receiveFcmMessage(async, messageFcmMessage(message1, account: account1));
+      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
+
+      check(testBinding.androidNotificationHost.activeNotifications).length.equals(4);
+      await NotificationDisplayManager.removeNotificationsForAccount(account1.realmUrl, userId);
+      check(testBinding.androidNotificationHost.activeNotifications).length.equals(2);
+      check(testBinding.androidNotificationHost.activeNotifications.first.notification.group)
+        .equals('${account2.realmUrl}|$userId');
```
gnprice, line 1018: nit: lines too long — key info beyond 80 columns

(e.g. those expected lengths, which are the main thing this test is watching for!)

gnprice (commented):
Thanks for the revision! These tests look complete now. Comments below.

test/model/actions_test.dart
```diff
@@ -121,6 +145,30 @@ void main() {
...
+      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
+
+      check(testBinding.globalStore).accountIds.single.equals(eg.selfAccount.id);
+      final future = logOutAccount(testBinding.globalStore, eg.selfAccount.id);
+      check(testBinding.globalStore.takeDoRemoveAccountCalls())
+        .single.equals(eg.selfAccount.id);
+      async.elapse(TestGlobalStore.removeAccountDuration);
+      await future;
+
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
```
gnprice, line 170: This can be further simplified, on the lines described in https://github.com/zulip/zulip-flutter/pull/1349#discussion_r2021897865 :
```suggestion
      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();

      await logOutAccount(testBinding.globalStore, eg.selfAccount.id);
      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
```

(with also a tweak to join checks up to the steps they're checking, https://github.com/zulip/zulip-flutter/pull/1349#discussion_r2021889137)

gnprice: This way the narrative is: look, there were notifications; then we called logout; now there aren't any, so whatever was there got removed.

test/notifications/display_test.dart
```diff
@@ -955,6 +963,70 @@ void main() {
...
+      check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      receiveFcmMessage(async, messageFcmMessage(message1, account: account1));
+      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
+
+      check(testBinding.androidNotificationHost.activeNotifications).length.equals(4);
+      await NotificationDisplayManager.removeNotificationsForAccount(realmUrl, account1.userId);
+      check(testBinding.androidNotificationHost.activeNotifications).length.equals(2);
+      check(testBinding.androidNotificationHost.activeNotifications).first.notification.group
```
gnprice, line 997: nit: bump https://github.com/zulip/zulip-flutter/pull/1349#discussion_r2021905895

test/notifications/display_test.dart
```diff
@@ -955,6 +963,70 @@ void main() {
...
+        .equals('$realmUrl|${account2.userId}');
+    })));
+
+    test('removeNotificationsForAccount leaves notifications for other accounts (same user-ids)', () => runWithHttpClient(() => awaitFakeAsync((async) async {
+      await init(addSelfAccount: false);
+
+      final userId = 1001;
+      final account1 = eg.account(id: 1001, user: eg.user(userId: userId), realmUrl: Uri.parse('https://realm1.example'));
+      final account2 = eg.account(id: 1002, user: eg.user(userId: userId), realmUrl: Uri.parse('https://realm2.example'));
```
gnprice, line 1006: also bump https://github.com/zulip/zulip-flutter/pull/1349#discussion_r2021905895

test/notifications/display_test.dart
```diff
@@ -955,6 +963,70 @@ void main() {
...
+      receiveFcmMessage(async, messageFcmMessage(message1, account: account1));
+      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
+
+      check(testBinding.androidNotificationHost.activeNotifications)
+        .length.equals(4);
+      await NotificationDisplayManager.removeNotificationsForAccount(account1.realmUrl, userId);
+      check(testBinding.androidNotificationHost.activeNotifications)
+        .length.equals(2);
+      check(testBinding.androidNotificationHost.activeNotifications)
+        .first.notification.group.equals('${account2.realmUrl}|$userId');
```
gnprice, line 1021: nit: can simplify with method-cascade syntax:
```suggestion
      check(testBinding.androidNotificationHost.activeNotifications)
        ..length.equals(2)
        ..first.notification.group.equals('${account2.realmUrl}|$userId');
```

test/notifications/display_test.dart
```diff
@@ -955,6 +963,70 @@ void main() {
...
+      await testBinding.globalStore.add(account2, eg.initialSnapshot());
+
+      final message1 = eg.streamMessage();
+      final message2 = eg.streamMessage();
+      receiveFcmMessage(async, messageFcmMessage(message1, account: account1));
+      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
+
+      check(testBinding.androidNotificationHost.activeNotifications)
+        .length.equals(4);
+      await NotificationDisplayManager.removeNotificationsForAccount(account1.realmUrl, userId);
```
gnprice, line 1017: nit: bump https://github.com/zulip/zulip-flutter/pull/1349#discussion_r2021900257

```suggestion
      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
      check(testBinding.androidNotificationHost.activeNotifications)
        .length.equals(4);

      await NotificationDisplayManager.removeNotificationsForAccount(account1.realmUrl, userId);
```

gnprice (commented):
Thanks for the revision! This is very close now. Comments below.

Some of these comments are additional cases of things that I mentioned in the previous review. When you get code-review feedback on a PR, one key thing you can do to help keep the PR moving swiftly along to merge is to look for other places where the same comments would apply, and act on the feedback in those places too.

tomlin7:
Thanks for the suggestions! It's improving my code quality significiantly. I made sure to run the suggestions through all changes made.

test/notifications/display_test.dart
```diff
@@ -955,6 +963,74 @@ void main() {
       receiveFcmMessage(async, removeFcmMessage([message2], account: account2));
       check(testBinding.androidNotificationHost.activeNotifications).isEmpty();
     })));
+
+    test('removeNotificationsForAccount: removes notifications', () => runWithHttpClient(() => awaitFakeAsync((async) async {
+      await init();
+      final message = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
+
+      receiveFcmMessage(async, messageFcmMessage(message));
+      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
```
gnprice, line 972: nit: these are closely tied, so group together in one stanza:
```suggestion
      final message = eg.dmMessage(from: eg.otherUser, to: [eg.selfUser]);
      receiveFcmMessage(async, messageFcmMessage(message));
      check(testBinding.androidNotificationHost.activeNotifications).isNotEmpty();
```

(e.g., that `message` local exists only to be used on that next line, so those two lines belong in the same stanza)

test/notifications/display_test.dart
```diff
@@ -955,6 +963,74 @@ void main() {
...
+    })));
+
+    test('removeNotificationsForAccount leaves notifications for other accounts (same user-ids)', () => runWithHttpClient(() => awaitFakeAsync((async) async {
+      await init(addSelfAccount: false);
+
+      final userId = 1001;
+      final account1 = eg.account(
+        id: 1001, user: eg.user(userId: userId), realmUrl: Uri.parse('https://realm1.example'));
+      final account2 = eg.account(
+        id: 1002, user: eg.user(userId: userId), realmUrl: Uri.parse('https://realm2.example'));
```
gnprice, line 1010: nit: the key info here (the parts of the realm URLs that differ) is still past 80 columns

test/notifications/display_test.dart
```diff
@@ -955,6 +963,74 @@ void main() {
...
+      final message2 = eg.streamMessage();
+      receiveFcmMessage(async, messageFcmMessage(message1, account: account1));
+      receiveFcmMessage(async, messageFcmMessage(message2, account: account2));
+      check(testBinding.androidNotificationHost.activeNotifications)
+        .length.equals(4);
+
+      await NotificationDisplayManager.removeNotificationsForAccount(realmUrl, account1.userId);
+      check(testBinding.androidNotificationHost.activeNotifications)
+        ..length.equals(2)
+        ..first.notification.group.equals('$realmUrl|${account2.userId}');
```
gnprice, line 1000: nit:
```suggestion
      await NotificationDisplayManager.removeNotificationsForAccount(
        realmUrl, account1.userId);
      check(testBinding.androidNotificationHost.activeNotifications)
        ..length.equals(2)
        ..first.notification.group.equals('$realmUrl|${account2.userId}');
```

otherwise "`account1.userId`" is hidden off past 80 columns, and that's the key information for interpreting the check at the end of the test

gnprice (commented):
Thanks for the revision! Just a few nits remain.

gnprice:
All looks good — merging! Thanks again @tomlin7 for all your work on this.

PR #1359, opened 2025-02-16
fake_api: Prepare API exceptions conveniently, distinct from HTTP exceptions

gnprice:
Stacked atop #1355. Prompted by this review thread (thanks @chrisbobbe for the observation):
https://github.com/zulip/zulip-flutter/pull/1355#discussion_r1956718605

#### a93323261 fake_api [nfc]: Clarify prepared exception is at HTTP layer

And leave a couple of TODO comments in the one test file that was
accidentally preparing exceptions here that won't turn out in the
intended way.  The next commit will provide a clean way for these
tests to do what they intend instead.

#### 2ed83f3f0 test [nfc]: Introduce eg.apiBadRequest

This changes the message string in a few of the call sites, where
the string was already an arbitrary one rather than a realistic
specific string.  That's still NFC because the tests weren't
depending on the specific string.

### DIFF
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -504,7 +504,7 @@ Future<T> tryRequest<T extends Object?>({
   fromJson ??= (((Map<String, dynamic> x) => x) as T Function(Map<String, dynamic>));
   return FakeApiConnection.with_((connection) {
     connection.prepare(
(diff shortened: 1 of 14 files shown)


chrisbobbe:
Thanks! LGTM, merged.

Issue #1360, opened 2025-02-16
CI flaky in Android build as out of memory


gnprice:
We're continuing to occasionally see the symptom of #1330 — CI runs sometimes fail with a "Java heap space" error in the `android` suite, as Gradle runs out of memory. Here's [an example today](https://github.com/zulip/zulip-flutter/actions/runs/13152142737/job/36817594452?pr=1327):
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:compileDebugKotlin'.
> Error while evaluating property 'friendPathsSet$kotlin_gradle_plugin_common' of task ':app:compileDebugKotlin'.
   > Could not resolve all files for configuration ':app:debugCompileClasspath'.
      > Failed to transform armeabi_v7a_debug-1.0.0-1e035cc693f067ed5033f180e0b6169d2c8a3e58.jar (io.flutter:armeabi_v7a_debug:1.0.0-1e035cc693f067ed5033f180e0b6169d2c8a3e58) to match attributes {artifactType=android-classes-jar, org.gradle.category=library, org.gradle.libraryelements=jar, org.gradle.status=release, org.gradle.usage=java-api}.
         > Execution failed for JetifyTransform: /home/runner/.gradle/caches/modules-2/files-2.1/io.flutter/armeabi_v7a_debug/1.0.0-1e035cc693f067ed5033f180e0b6169d2c8a3e58/1d4de013e75bd976543eba257de04ed6657c9c28/armeabi_v7a_debug-1.0.0-1e035cc693f067ed5033f180e0b6169d2c8a3e58.jar.
            > Java heap space
```

Before #1324, this failure was happening in 100% of CI runs — that was #1330.

With #1324 we bumped the Gradle JVM max heap size from 1.5 GiB to 3 GiB. That made CI mostly pass, but it continued to sometimes fail:
https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Gradle.20out.20of.20memory/near/2082072

With #1331 we bumped it further to 4 GiB. The failure today, quoted above, was with 4 GiB.

gnprice:
The obvious next thing to try, and which I guess we should try, is to bump the max heap size further.

[...]

PR #1361, opened 2025-02-16
android build: Bump Gradle max memory use to 6 GiB

gnprice:
Fixes, potentially, #1360.

Even after the bump to 4 GiB, CI is still occasionally flaking with a "Java heap space" error in the `android` suite:
  https://github.com/zulip/zulip-flutter/actions/runs/13152142737/job/36817594452?pr=1327

So try bumping the memory limit still further.  It's not entirely clear this will be enough (or that an even bigger number would be):
  https://github.com/zulip/zulip-flutter/issues/1360#issuecomment-2661191343
but we'll find out.

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx4G
+org.gradle.jvmargs=-Xmx6G
 android.useAndroidX=true
 android.enableJetifier=true
 


gnprice:
Turns out this apparent flake was actually just a classic bug in the GitHub web app:
https://github.com/zulip/zulip-flutter/issues/1360#issuecomment-2661198124

So we don't need this PR. (Or at least we don't yet have any indication that we need it.)

PR #1363, opened 2025-02-17
test: testable `DateTime.now`

Komyyy:


chrisbobbe:
Thanks! fa5dfd56 added `ZulipBinding.utcNow`; can we use that, or is there a need to add a new field?

PIG208:
I that believe this is just a timezone conversion away from what we have in main. The part about replacing DateTime.now with the ZulipBinding equivalent is a nice follow-up we mentioned in 0c03009.

chrisbobbe:
So it is something we want, separately from `utcNow`? If so, Komyyy let's put the two fields side-by-side and make their dartdocs more parallel/similar, with bidirectional "see also" notes that help the reader decide which one to use.

Komyyy:
> So it is something we want, separately from `utcNow`? If so, Komyyy let's put the two fields side-by-side and make their dartdocs more parallel/similar, with bidirectional "see also" notes that help the reader decide which one to use.

I wasn't aware that this function had been added.
The `ZulipBinding.now` function can be replaced with `DateTime.toLocal` and `ZulipBinding.utcNow`.
Therefore, I will update this PR to replace `DateTime.now` with `ZulipBinding.utcNow`.

PR #1364, opened 2025-02-19
general chat #2: support sending to empty topic from channel narrow, hide resolve/unresolve conditionally

PIG208:
Similar to #1297, this PR focuses on part of the general chat feature, with some final "do not merge" commits to make it testable.

This includes the following commits:

* 5ad98d9bb compose: Support sending to empty topic
* 1e31660a3 compose: Change content input hint text if topic is empty and mandatory
* 073042a69 compose [nfc]: Make isTopicVacuous private
* dd47653d3 compose [nfc]: Handle empty topics for fixed destination input hint text
* 73c46b5e3 compose test [nfc]: Extract ChannelNarrow variable
* 0f4e7948d action_sheet [nfc]: Hide resolve/unresolve button for empty topic

The remaining commits should be reviewed in #1297, but neither PR blocks each other, as they each focuses on a specific part of the feature.

<details>

| mandatory topics | non-empty topic | content is focused | content isn't focused |
| - | - | - | - |
| true | mandatory-channel-non-empty | mandatory-channel-empty-focused | mandatory-channel-empty-non-focused |
| false | non-mandatory-channel-non-empty | non-mandatory-channel-empty-focused | non-mandatory-channel-empty-non-focused |
| false (legacy, FL<334) | / | non-mandatory-channel-empty-focused-legacy | non-mandatory-channel-empty-non-focused-legacy |
| hide resolve | resolve | |
</details>

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -157,7 +157,12 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
   @override
   String _computeTextNormalized() {
     String trimmed = text.trim();
-    return trimmed.isEmpty ? kNoTopicTopic : trimmed;
+    // TODO(server-10): simplify
+    if (store.zulipFeatureLevel < 334) {
+      return trimmed.isEmpty ? kNoTopicTopic : trimmed;
+    }
+
+    return trimmed;
   }
 
   /// Whether [textNormalized] would fail a mandatory-topics check
@@ -165,7 +170,20 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
   ///
   /// The term "Vacuous" draws distinction from [String.isEmpty], in the sense
   /// that certain strings are not empty but also indicate the absence of a topic.
-  bool get isTopicVacuous => textNormalized == kNoTopicTopic;
+  ///
+  /// See also: https://zulip.com/api/send-message#parameter-topic
+  bool get isTopicVacuous {
+    if (textNormalized.isEmpty) return true;
+
+    if (textNormalized == kNoTopicTopic) return true;
+
+    // TODO(server-10): simplify
+    if (store.zulipFeatureLevel >= 334) {
+      return textNormalized == store.realmEmptyTopicDisplayName;
+    }
+
+    return false;
+  }
 
   @override
   List<TopicValidationError> _computeValidationErrors() {
@@ -558,10 +576,17 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     });
   }
 
+  void _contentFocusChanged() {
+    setState(() {
+      // The relevant state lives on widget.controller.contentFocusNode itself.
+    });
+  }
+
   @override
   void initState() {
     super.initState();
     widget.controller.topic.addListener(_topicChanged);
+    widget.controller.contentFocusNode.addListener(_contentFocusChanged);
   }
 
   @override
@@ -571,11 +596,16 @@ class _StreamContentInputState extends State<_StreamContentInput> {
       oldWidget.controller.topic.removeListener(_topicChanged);
       widget.controller.topic.addListener(_topicChanged);
     }
+    if (widget.controller.contentFocusNode != oldWidget.controller.contentFocusNode) {
+      oldWidget.controller.contentFocusNode.removeListener(_contentFocusChanged);
+      widget.controller.contentFocusNode.addListener(_contentFocusChanged);
+    }
   }
 
   @override
   void dispose() {
     widget.controller.topic.removeListener(_topicChanged);
+    widget.controller.contentFocusNode.removeListener(_contentFocusChanged);
     super.dispose();
   }
 
@@ -584,6 +614,13 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     if (widget.controller.topic.isTopicVacuous) {
       if (widget.controller.topic.mandatory) {
         // The chosen topic can't be sent to, so don't show it.
+        return null;
+      }
+      if (!widget.controller.contentFocusNode.hasFocus) {
+        // Do not fall back to a vacuous topic unless the user explicitly chooses
+        // to do so (by skipping topic input and moving focus to content input),
+        // so that the user is not encouraged to use vacuous topic when they
+        // have not interacted with the inputs at all.
         return null;
       }
     }
@@ -604,7 +641,9 @@ class _StreamContentInputState extends State<_StreamContentInput> {
       // Zulip expresses channels and topics, not any normal English punctuation,
       // so don't make sense to translate. See:
       //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
-      ? '#$streamName' : '#$streamName > ${hintTopic.displayName}';
+      ? '#$streamName'
+      // ignore: dead_null_aware_expression // null topic names soon to be enabled
+      : '#$streamName > ${hintTopic.displayName ?? store.realmEmptyTopicDisplayName}';
 
     return _ContentInput(
       narrow: widget.narrow,
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -579,8 +579,15 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     super.dispose();
   }
 
-  /// The topic name to use in the hint text.
-  TopicName _hintTopic() {
+  /// The topic name to show in the hint text, or null to show no topic.
+  TopicName? _hintTopic() {
+    if (widget.controller.topic.isTopicVacuous) {
+      if (widget.controller.topic.mandatory) {
+        // The chosen topic can't be sent to, so don't show it.
+        return null;
+      }
+    }
+
     return TopicName(widget.controller.topic.textNormalized);
   }
 
@@ -591,12 +598,13 @@ class _StreamContentInputState extends State<_StreamContentInput> {
 
     final streamName = store.streams[widget.narrow.streamId]?.name
       ?? zulipLocalizations.unknownChannelName;
-    final hintDestination =
+    final hintTopic = _hintTopic();
+    final hintDestination = hintTopic == null
       // No i18n of this use of "#" and ">" string; those are part of how
       // Zulip expresses channels and topics, not any normal English punctuation,
       // so don't make sense to translate. See:
       //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
-      '#$streamName > ${_hintTopic().displayName}';
+      ? '#$streamName' : '#$streamName > ${hintTopic.displayName}';
 
     return _ContentInput(
       narrow: widget.narrow,
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -658,7 +658,8 @@ class _FixedDestinationContentInput extends StatelessWidget {
           // Zulip expresses channels and topics, not any normal English punctuation,
           // so don't make sense to translate. See:
           //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
-          '#$streamName > ${topic.displayName}');
+          // ignore: dead_null_aware_expression // null topic names soon to be enabled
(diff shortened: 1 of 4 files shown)


PIG208:
(pending an update to fix an issue with handling "(no topic)")

test/widgets/compose_box_test.dart
```diff
@@ -371,11 +373,22 @@ void main() {
...
-    testWidgets('to TopicNarrow', (tester) async {
-      await prepare(tester,
-        narrow: TopicNarrow(channel.streamId, TopicName('topic')));
-      checkComposeBoxHintTexts(tester,
-        contentHintText: 'Message #${channel.name} > topic');
+    group('to TopicNarrow', () {
+      testWidgets('with non-empty topic', (tester) async {
+        await prepare(tester,
+          narrow: TopicNarrow(channel.streamId, TopicName('topic')),
+          mandatoryTopics: false);
```
chrisbobbe, line 467: Does `mandatoryTopics` make any difference for a topic-narrow compose box? A topic-narrow compose box doesn't offer a topic input.

lib/widgets/compose_box.dart
```diff
@@ -167,6 +167,32 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
   /// that certain strings are not empty but also indicate the absence of a topic.
   bool get _isTopicVacuous => textNormalized == kNoTopicTopic;
 
+  /// The send destination as a string.
```
chrisbobbe, line 189:
```
compose: Change content input hint text if topic is empty and mandatory

Previously, "Message #stream > (no topic)" would appear as the hint text
when the topic input is empty but mandatory.

[…]
```

The commit message says what the hint text used to be; how about also saying what it is now?

test/widgets/compose_box_test.dart
PIG208: Yes, because the hint text of the content input displays the topic name (e.g.: "Message #channel > general chat").

test/widgets/compose_box_test.dart
```diff
@@ -726,10 +777,21 @@ void main() {
         expectedMessage: 'Topics are required in this organization.');
     }
 
-    testWidgets('empty topic -> "(no topic)"', (tester) async {
+    testWidgets('empty topic -> empty topic', (tester) async {
```
chrisbobbe, line 780: How about calling this 'empty topic -> ""'? The left side is about the state of a text input, and the right is about a literal string, right?

test/widgets/compose_box_test.dart
```diff
@@ -743,6 +805,13 @@ void main() {
       checkMessageNotSent(tester);
     });
 
+    testWidgets('if topics are mandatory, reject `realmEmptyTopicDisplayName`', (tester) async {
+      await setupAndTapSend(tester,
+        topicInputText: eg.defaultRealmEmptyTopicDisplayName,
+        mandatoryTopics: true);
+      checkMessageNotSent(tester);
+    }, skip: true); // null topic names soon to be enabled
```
chrisbobbe: Does this need to be skipped?

lib/widgets/compose_box.dart
```diff
@@ -157,20 +157,80 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
...
+      // TODO(server-10): simplify away conditional
+      textNormalized.isEmpty ? store.realmEmptyTopicDisplayName : textNormalized;
+    return '#$streamName > $vacuousTopicDisplayName';
+  }
+
+  @override List<TopicValidationError> _computeValidationErrors() {
     return [
-      if (mandatory && isTopicVacuous)
+      if (mandatory && _isTopicVacuous)
         TopicValidationError.mandatoryButEmpty,
```
chrisbobbe, line 234: Looking again at this, it looks like `TopicValidationError.mandatoryButEmpty` should really be called `TopicValidationError.mandatoryButVacuous`, is that right?

chrisbobbe: Perhaps similarly for `zulipLocalizations.topicValidationErrorMandatoryButEmpty`?

chrisbobbe (commented):
And now some comments on that more-complicated commit. 🙂

lib/widgets/compose_box.dart
PIG208: I think it makes sense to rename the enum value, because the distinction between empty and vacuous matters to the developer. Not too sure about translator/user-facing strings, but I understand the benefit of having variable names and string names match.

test/widgets/compose_box_test.dart
```diff
@@ -47,15 +47,18 @@ void main() {
...
   }) async {
     if (narrow case ChannelNarrow(:var streamId) || TopicNarrow(: var streamId)) {
       assert(streams.any((stream) => stream.streamId == streamId),
         'Add a channel with "streamId" the same as of $narrow.streamId to the store.');
     }
     addTearDown(testBinding.reset);
     selfUser ??= eg.selfUser;
-    final selfAccount = eg.account(user: selfUser);
+    zulipFeatureLevel ??= eg.futureZulipFeatureLevel;
+    final selfAccount = eg.account(user: selfUser, zulipFeatureLevel: zulipFeatureLevel);
```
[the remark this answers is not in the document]
PIG208: Usually the benefit of `copyWith` is the brevity as you don't need to refer to the user. Here `selfUser` can be overridden, so it seems better to create a fresh account.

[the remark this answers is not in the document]
chrisbobbe: I see; makes sense.

PIG208:
Thanks for the review! I have updated the PR. With a question on https://github.com/zulip/zulip-flutter/pull/1364#discussion_r1992365465 about whether we should use "vacuous" in user-facing strings.

chrisbobbe (commented):
Thanks! One small comment below, and I'll mark for Greg's review.

lib/widgets/compose_box.dart
```diff
@@ -170,6 +170,32 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
...
+  ///
+  /// This returns a string formatted like "#stream name" when topics are
+  /// [mandatory] but the trimmed input is empty.
+  ///
+  /// Otherwise, returns a string formatted like "#stream name > topic name".
+  // No i18n of the use of "#" and ">" strings; those are part of how
+  // Zulip expresses channels and topics, not any normal English punctuation,
+  // so don't make sense to translate. See:
+  //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
+  String getDestinationString({required String streamName}) {
```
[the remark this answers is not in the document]
PIG208: Yeah, good point. This was placed next to `_isTopicVacuous` because its implementation logic used to be more closely related to `_isTopicVacuous`. But now moving it to `_StreamContentInput` as a private helper should be a good call.

PIG208:
Will push an update later addressing https://github.com/zulip/zulip-flutter/pull/1364#discussion_r2004335904.

gnprice (commented):
Thanks both!

Here's a review from reading up through:
85104e149 action_sheet [nfc]: Hide resolve/unresolve button for empty topic
b612c794c compose test [nfc]: Extract ChannelNarrow variable
77fa5a0d7 compose [nfc]: Handle empty topics for topic-narrow input hint text
a7ff94cb4 compose [nfc]: Make isTopicVacuous private and move some of its dartdoc

and part of the next commit:
442d11296 compose: Change content input hint text if topic is empty and mandatory

I haven't yet read:
dcaf1659a compose: Support sending to empty topic

I think a good solution for a couple of my comments below may be clearest by illustration. I'll send a draft PR shortly to do that.

gnprice:
> I think a good solution for a couple of my comments below may be clearest by illustration. I'll send a draft PR shortly to do that.

→ #1425

PIG208:
Thanks! #1425 is super helpful. The PR has been updated.

test/widgets/compose_box_test.dart
```diff
@@ -370,11 +420,58 @@ void main() {
...
+          topic: eg.defaultRealmEmptyTopicDisplayName);
+        await tester.pump();
+        checkComposeBoxHintTexts(tester,
+          topicHintText: 'Topic',
+          contentHintText: 'Message #${channel.name}');
+      });
+
+      testWidgets('legacy: with empty topic', (tester) async {
+        await prepare(tester, narrow: narrow, mandatoryTopics: true,
+          zulipFeatureLevel: 333);
```
gnprice, line 445:
```suggestion
          zulipFeatureLevel: 333); // TODO(server-10)
```

Usually tests don't need TODO-server comments, because usually (if they're about old-server code at all) their point is to test some behavior that we're going to delete when we drop support for the old server — so the test will break, and we'll naturally find it so as to clean it out too.

But here this is identical to a non-legacy test above, except for the feature level it sets. (I believe its purpose is to test that some old-server code we have *doesn't* have any effect in a certain case.) So it looks like it would continue to pass if we deleted that code and didn't notice this test. Therefore that normal logic doesn't apply, and it should get a TODO-server comment.

Similarly one of the tests above.

test/widgets/compose_box_test.dart
```diff
@@ -370,11 +420,58 @@ void main() {
...
+        await prepare(tester, narrow: narrow, mandatoryTopics: true);
+        await enterTopic(tester, narrow: narrow,
+          topic: eg.defaultRealmEmptyTopicDisplayName);
+        await tester.pump();
+        checkComposeBoxHintTexts(tester,
+          topicHintText: 'Topic',
+          contentHintText: 'Message #${channel.name}');
+      });
+
+      testWidgets('legacy: with empty topic', (tester) async {
```
gnprice, line 443: nit: the ordering of this among the other test cases doesn't seem logical: it's separating these two non-legacy test cases from each other, but without getting the benefit of being right next to the non-legacy test case that directly corresponds to it.

Either after all three cases, or right after the non-legacy "with empty topic", would be a logical home.

test/widgets/compose_box_test.dart
```diff
@@ -370,11 +420,58 @@ void main() {
...
+      final narrow = ChannelNarrow(channel.streamId);
+
+      testWidgets('with empty topic', (tester) async {
+        await prepare(tester, narrow: narrow, mandatoryTopics: true);
+        checkComposeBoxHintTexts(tester,
+          topicHintText: 'Topic',
+          contentHintText: 'Message #${channel.name}');
+      });
+
+      testWidgets('with non-empty but vacuous topic', (tester) async {
```
gnprice, line 433: It looks like this case is missing a legacy variant (with `(no topic)` as the contents of the topic input).

gnprice (commented):
Thanks! Generally this all looks good now; a few comments below.

lib/widgets/compose_box.dart
```diff
@@ -157,15 +157,34 @@ class ComposeTopicController extends ComposeController<TopicValidationError> {
...
   /// The term "Vacuous" draws distinction from [String.isEmpty], in the sense
   /// that certain strings are not empty but also indicate the absence of a topic.
-  bool get isTopicVacuous => textNormalized == kNoTopicTopic;
+  bool get isTopicVacuous {
+    bool result = textNormalized.isEmpty
+      // We keep checking for '(no topic)' regardless of the feature level
+      // because it remains equivalent to an empty topic even when FL >= 334.
+      // This can change in the future:
+      //   https://chat.zulip.org/#narrow/channel/412-api-documentation/topic/.28realm_.29mandatory_topics.20behavior/near/2062391
+      || textNormalized == kNoTopicTopic;
```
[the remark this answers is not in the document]
PIG208: Added a link to the updated API documentation ([send-message#parameter-topic](https://zulip.com/api/send-message#parameter-topic)) and remove the block of comment about how "(no topic)" check can change "in the future".

PIG208:
Pushed an update and replied to some comments. Thanks for the review!

gnprice:
Thanks! Looks good; merging, with those "do not merge" commits at the end stripped off:
cca33c4a4 do not merge; api: Make displayName nullable
7646cb44f do not merge; api: Indicate support for handling empty topics

PR #1365, opened 2025-02-19
general chat #3: support alternative styling for topic input on edits/focus/blur

PIG208:


lib/widgets/compose_box.dart
```diff
@@ -654,41 +669,121 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+          defaultTopicDisplayName);
+      } else if (widget.controller.hasChosenTopic.value) {
+        // The topic has been chosen. Show the default topic display name in
+        // the input as if the user has entered that when they left it empty.
+        hintText = defaultTopicDisplayName;
+        hintStyle = topicTextStyle.copyWith(
+          // TODO(server-10) simplify
+          fontStyle: store.zulipFeatureLevel >= 334 ? FontStyle.italic : null);
+      }
+    }
```
PIG208, line 765: Alternatively, with `switch`:
```dart
final defaultTopicDisplayName = store.zulipFeatureLevel >= 334
  ? store.realmEmptyTopicDisplayName : kNoTopicTopic;

final decoration = switch ((
  store.realmMandatoryTopics,
  widget.controller.hasChosenTopic.value,
  widget.controller.topicFocusNode.hasFocus,
)) {
  (false, true, _) => InputDecoration(
    hintText: defaultTopicDisplayName,
    hintStyle: topicTextStyle.copyWith(
      fontStyle: store.zulipFeatureLevel >= 334 ? FontStyle.italic : null)),
  (false, false, true) => InputDecoration(
    hintText: zulipLocalizations.composeBoxEnterTopicOrSkipHintText(
      defaultTopicDisplayName),
    hintStyle: hintStyle),
  (_, _, _) => InputDecoration(
    hintText: zulipLocalizations.composeBoxTopicHintText,
    hintStyle: hintStyle),
};
```

gnprice: I think that `switch` would be easier to understand as an if/else chain — I feel like as the reader I'd be basically transforming it into that in my head in order to understand what it's saying.

lib/widgets/compose_box.dart
```diff
@@ -1368,10 +1418,18 @@ class StreamComposeBoxController extends ComposeBoxController {
   final ComposeTopicController topic;
   final topicFocusNode = FocusNode();
 
+  /// Whether the user has made up their mind choosing a topic.
+  ///
+  /// Empirically, this should be set to `false` whenever the user focuses on
+  /// the topic input, and set to `true` whenever the user focuses on the
+  /// content input.
+  ValueNotifier<bool> hasChosenTopic = ValueNotifier(false);
```
chrisbobbe, line 1461: How about "Whether the user seems to have made up their mind choosing a topic." or similar? It's impossible to know for sure, right :) they could come back and change it again before sending, or they could send accidentally before making up their mind.

Also I don't understand what "empirically" means here. The sentence sounds like we're an observer making notes from what we've seen of the app's behavior, rather than the ones deciding what the behavior is and causing it. 🙂

lib/widgets/compose_box.dart
```diff
@@ -576,16 +576,26 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+    setState(() {
+      // The relevant state lives on widget.controller.hasChosenTopic itself.
+    });
+  }
+
   void _contentFocusChanged() {
     setState(() {
       // The relevant state lives on widget.controller.contentFocusNode itself.
     });
+    if (widget.controller.contentFocusNode.hasFocus){
```
chrisbobbe, line 588: nit: spacing

lib/widgets/compose_box.dart
```diff
@@ -576,16 +576,26 @@ class _StreamContentInputState extends State<_StreamContentInput> {
     });
   }
 
+  void _hasChosenTopicChanged() {
```
chrisbobbe, line 578:
```
compose: Track user interaction with inputs statefully
```

What's the behavior change in this commit?

lib/widgets/compose_box.dart
```diff
@@ -712,15 +718,52 @@ class _TopicInputState extends State<_TopicInput> {
...
+        // The user is actively interacting with the input.  Show a long and
+        // engaging hint text.
+        hintText: zulipLocalizations.composeBoxEnterTopicOrSkipHintText(
+          defaultTopicDisplayName),
+        hintStyle: hintStyle),
+
+      (_, _, _) => InputDecoration(
+        // In all other cases, show a short hint text for less distraction.
+        hintText: zulipLocalizations.composeBoxTopicHintText,
+        hintStyle: hintStyle),
```
chrisbobbe, line 761: Can we assert that cases `(_, false, false)` and `(_, true, true)` never happen? If those states can happen, they probably won't persist for more than a frame—but that would be a frame where the text would be flickering to something different, and it would look janky.

Also would it be helpful to mention `realmMandatoryTopics` in some of these comments?

I wonder if an if/else might be a better fit for this logic than a `switch`.

chrisbobbe (commented):
Thanks! Small comments below.

lib/widgets/compose_box.dart
PIG208: I think `(_ false, false)` can represent the initial state when the user open a page without interacting with any of the inputs, so it is definitely possible.

`(_, true, true)` seems possible as a really intermediate state, but it is hard to tell without staring at the conditions for some time.

These three cases might be better represented with an enum:
- `none` when the user hasn't interacted with the inputs;
- `isEditing` when the topic input has focus;
- `hasEdited` when the content input has focus/topic input lost focus previously.

We can tweak the behavior a bit so `none` -> `isEditing`, `none` -> `hasEdited`, and `isEditing` <-> `hasEdited` are the only possible state transitions. This eliminates the possible flickering when more than one transition can happen between frames (namely the case when topic input is unfocused and content input gains focus, which can be represented as the `isEditing` -> `hasEdited` transition).

PIG208:
Thanks for the review! I have updated the PR. This revision rewrites the feature using `enum`s and a state machine.

chrisbobbe:
Thanks! This seems to have gathered a rebase conflict; could you resolve that please?

lib/widgets/compose_box.dart
```diff
@@ -1383,17 +1466,63 @@ sealed class ComposeBoxController {
...
+///    topic input loses focus
+///    and content input has no focus
+/// ```
+///
+/// This state machine offers the following invariants:
+/// - When topic input has focus, the edit status must be [isEditing].
+/// - When content input has focus, the edit status must be [hasChosen].
+/// - Otherwise, the edit status can be either of [none] or [hasChosen],
+///   but never [isEditing].
+enum ComposeTopicEditStatus {
```
chrisbobbe, line 1499: I think "how a user has edited the topic" isn't the best description for what this represents. If I use this to try to straightforwardly describe each state, the results are misleading:
  - "The user has edited the topic in the 'none' way"—or perhaps "The user has not edited the topic"…but looking at the diagram, that's definitely not what this state means.

Or awkward/confusing:
  - "The user has edited the topic in the 'has chosen' way."
  - "The user has edited the topic in the 'is editing' way."

Or from another angle: it's equally possible to be in any of the states after any given sequence of edits (if an edit means changing the text in the input).

chrisbobbe: Is the "topic input gains focus" label meant to apply equally to the `hasChosen` -> `isEditing` transition? It's positioned so it looks like it's just meant to apply to `none` -> `isEditing`.

chrisbobbe:
```dart
/// - Otherwise, the edit status can be either of [none] or [hasChosen],
///   but never [isEditing].
```

"Otherwise" means neither the topic input nor the content input has focus, right? Looking at the diagram, I don't see how the state could be `hasChosen` in that "otherwise" case.

chrisbobbe (commented):
Thanks! Here's a review of the first two commits:

90770e123 compose: Unskip test; handle TopicName with empty displayName for now
ec6622fda compose [nfc]: Convert _TopicInput to a StatefulWidget

and I've read part of the third:

513298058 compose: Change topic input hint text

lib/widgets/compose_box.dart
```diff
@@ -655,8 +655,10 @@ class _StreamContentInputState extends State<_StreamContentInput> {
       // so don't make sense to translate. See:
       //   https://github.com/zulip/zulip-flutter/pull/1148#discussion_r1941990585
       ? '#$streamName'
-      // ignore: dead_null_aware_expression // null topic names soon to be enabled
-      : '#$streamName > ${hintTopic.displayName ?? store.realmEmptyTopicDisplayName}';
+      // null topic names soon to be enabled
+      : '#$streamName > '
+        '${hintTopic.displayName.isEmpty ? store.realmEmptyTopicDisplayName
+                                         : hintTopic.displayName}';
```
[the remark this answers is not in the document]
PIG208: Yeah. the value of `TopicName` usually comes from the server, and before we start passing the "empty_topic_names" client capability, `displayName` being empty is unexpected. I think it should be fine to work with strings here.

Not sure if we want to switch back to `TopicName` once `TopicName.displayName` becomes nullable. In UI code, we usually handle empty topic names by checking if `TopicName.displayName` is `null`, which should also make grepping easier.

lib/widgets/compose_box.dart
```diff
@@ -682,15 +682,101 @@ class _TopicInput extends StatefulWidget {
...
+  void initState() {
+    super.initState();
+    widget.controller.topicFocusNode.addListener(_topicFocusChanged);
+    widget.controller.contentFocusNode.addListener(_contentFocusChanged);
+    widget.controller.topicEditStatus.addListener(_topicEditStatusChanged);
+  }
+
+  @override
+  void didUpdateWidget(covariant _TopicInput oldWidget) {
+    super.didUpdateWidget(oldWidget);
```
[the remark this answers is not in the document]
PIG208: I'm not sure what to include in the `onNewStore` method. `_TopicInputState` doesn't have listeners on `widget.controller.topic`. Its only reference to `widget.controller.topic` is in the build method, where it passes the `widget.controller.topic` instance as-is.

`widget.controller` comes from and is owned by `_ComposeBoxState`. It does implement a `onNewStore` method that swaps out `widget.controller.topic.store` with the new store. Maybe that should be sufficient?

lib/widgets/compose_box.dart
PIG208: Yeah. How about maybe `ComposeTopicInteractionStatus`? I'm also not happy with the `none` name. We can be a bit more literal about what that means. Updated the diagram:

```dart
/// ```
///                       (default)
///    Topic input            │          Content input
///    lost focus.            ▼          gained focus.
///   ┌────────────► notEditingNotChosen ────────────┐
///   │                                 │            │
///   │         Topic input             │            │
///   │         gained focus.           │            │
///   │       ◄─────────────────────────┘            ▼
/// isEditing ◄───────────────────────────── hasChosen
///   │         Focus moved from                     ▲
///   │         content to topic.                    │
///   │                                              │
///   └──────────────────────────────────────────────┘
///    Focus moved from
///    topic to content.
/// ```
```

The state can be `hasChosen` even after content input has lost focus. Will revise that to make it a bit clearer.

lib/widgets/compose_box.dart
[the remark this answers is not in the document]
gnprice: TopicName is for requests we make (or might make) to the server, as much as it is for values we get from the server. That's the reason for its `apiName` getter:
```dart
  /// The string this topic is identified by in the Zulip API.
  ///
  /// This should be used in constructing HTTP requests to the server,
  /// but rarely for other purposes.  See [displayName] and [canonicalize].
```
and is why it appears in a number of places in the compose-box code.

As of the tip of the branch, I think it's helpful that this `hintTopic` local is a TopicName, because that allows the logic for falling back to "general chat" to look the same as the corresponding logic across the rest of the codebase:
```dart
      : '#$streamName > ${hintTopic.displayName ?? store.realmEmptyTopicDisplayName}';
```
Looking at references to `realmEmptyTopicDisplayName`, one sees that almost all of them are gated behind a condition of some `TopicName.displayName` being null, and the majority are in exactly this form with a `??`.

I haven't looked closely at the individual commits in this branch to have a suggestion on the clearest way to break the changes down into steps, though.

PIG208:
Thanks for the review! Pushed an update.

test/model/autocomplete_test.dart
```diff
@@ -1026,6 +1028,38 @@ void main() {
     check(done).isTrue();
   });
 
+  test('TopicAutocompleteView allow empty topic name on modern servers', () async {
+    final account = eg.account(user: eg.selfUser, zulipFeatureLevel: 334);
```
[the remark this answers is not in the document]
chrisbobbe: Relatedly, when I write tests, I think I generally omit "on modern servers" in a test description, leaving that as implied, and I write the description (and the test) so that it won't need changes when we eventually clear out the legacy code. Then the legacy-servers test is the thing that gets a marker in its description, like "on legacy servers", and a `TODO(server` comment to remove the test.

So here, maybe "TopicAutocompleteView getStreamTopics request"? (And add `..method.equals('GET')`, for completeness in matching that description.)

test/widgets/action_sheet_test.dart
```diff
@@ -1155,6 +1159,31 @@ void main() {
         await setupToMessageActionSheet(tester, message: message, narrow: const StarredMessagesNarrow());
         check(findQuoteAndReplyButton(tester)).isNull();
       });
+
+      testWidgets('handle empty topic', (tester) async {
+        final message = eg.streamMessage();
+        await setupToMessageActionSheet(tester, message: message, narrow: TopicNarrow.ofMessage(message),
+          zulipFeatureLevel: 334);
```
chrisbobbe, line 1166: (same comment about using `eg.recentZulipFeatureLevel`)

chrisbobbe (commented):
Thanks! I've now read through all the commits. Small comments below.

test/model/autocomplete_test.dart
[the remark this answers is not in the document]
PIG208: Yeah. We should be able to skip a lot of changes in the first commit if we bump `eg.recentZulipFeatureLevel`.

lib/widgets/compose_box.dart
```diff
@@ -669,41 +679,132 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+        hintText: zulipLocalizations.composeBoxEnterTopicOrSkipHintText(
+          defaultTopicDisplayName),
+        hintStyle: hintStyle),
+
+      (false, ComposeTopicInteractionStatus.notEditingNotChosen) ||
+      (true, _) => InputDecoration(
+        // Otherwise, show a short hint text for less distraction.
+        hintText: zulipLocalizations.composeBoxTopicHintText,
+        hintStyle: hintStyle),
+    };
```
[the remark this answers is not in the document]
PIG208: Thanks! This looks good to me (removed `late` from `late final String hintText`).

test/model/autocomplete_test.dart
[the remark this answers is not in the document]
PIG208: The description changes make sense to be, but I thought that we usually don't have `TODO(server-x)` comments on tests (since they ought to fail when we drop legacy server support)?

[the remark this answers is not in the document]
chrisbobbe: Mm, yeah, reasonable to leave them out on the tests I think.

PIG208:
Thanks for the review! Pushed a new update.

chrisbobbe (commented):
Thanks! Small comments below, and I'll mark for Greg's review.

lib/widgets/compose_box.dart
```diff
@@ -670,41 +680,140 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+class _TopicInputState extends State<_TopicInput> {
+  void _topicFocusChanged() {
+    setState(() {
+      if (widget.controller.topicFocusNode.hasFocus) {
+        widget.controller.topicInteractionStatus.value =
+          ComposeTopicInteractionStatus.isEditing;
+      } else if (!widget.controller.contentFocusNode.hasFocus) {
+        widget.controller.topicInteractionStatus.value =
+          ComposeTopicInteractionStatus.notEditingNotChosen;
+      }
```
gnprice, line 702:
```suggestion
      if (widget.controller.topicFocusNode.hasFocus) {
        widget.controller.topicInteractionStatus.value =
          ComposeTopicInteractionStatus.isEditing;
      } else if (!widget.controller.contentFocusNode.hasFocus) {
        widget.controller.topicInteractionStatus.value =
          ComposeTopicInteractionStatus.notEditingNotChosen;
      } else {
        assert(widget.controller.topicInteractionStatus.value
          == ComposeTopicInteractionStatus.hasChosen);
      }
```

Does that look right?

Otherwise, it's a bit confusing because this appears to be missing one of the transitions in the state diagram. IIUC the point is that if the content input already has focus, then _contentFocusChanged should have already set the status.

gnprice (commented):
Thanks to you both! Generally this is looking good.

Going AFK in a few minutes, so here's a review of what I've read so far:
d4cd45578 test: Bump recentZulipFeatureLevel to FL 334, from FL 278
c193b7b4e store [nfc]: Assert that FL>=334 when accessing realmEmptyTopicName
a39dc3cdc compose: Unskip test; show realmEmptyDisplayName in hint text
51de0bef9 compose [nfc]: Convert _TopicInput to a StatefulWidget
8992d1d16 compose: Change topic input hint text

That leaves for a future round:
607ddb351 compose: Change content input hint text statefully
3a315b92c api: Make displayName nullable
5fa846c34 api: Indicate support for handling empty topics

lib/widgets/compose_box.dart
PIG208: This `else` does represent the case when the focus moves from topic input to content input, however, the assertion will only succeed if `_contentFocusChanged` is called before `_topicFocusChanged`. Otherwise, the state is still `isEditing`; this seems to be the case consistently when I run all the tests.

> IIUC the point is that if the content input already has focus, then _contentFocusChanged should have already set the status.

Yeah, it appears that we have the up-to-date value of the content focus node (suggesting that `_topicFocusChanged` is called in the same frame?), but `_contentFocusChanged` hasn't been called yet. 

From the stacktrace:
```
#0      debugPrintStack (package:flutter/src/foundation/assertions.dart:1224:29)
#1      _TopicInputState._topicFocusChanged.<anonymous closure> (package:zulip/widgets/compose_box.dart:695:11)
#2      State.setState (package:flutter/src/widgets/framework.dart:1199:30)
#3      _TopicInputState._topicFocusChanged (package:zulip/widgets/compose_box.dart:685:5)
#4      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:435:24)
#5      FocusNode._notify (package:flutter/src/widgets/focus_manager.dart:1131:5)
#6      FocusManager.applyFocusChangesIfNeeded (package:flutter/src/widgets/focus_manager.dart:1996:12)
```

`applyFocusChangesIfNeeded`:

```dart
    if (previousFocus != _primaryFocus) {
      assert(_focusDebug(() => 'Updating focus from $previousFocus to $_primaryFocus'));
      if (previousFocus != null) {
        _dirtyNodes.add(previousFocus);
      }
      if (_primaryFocus != null) {
        _dirtyNodes.add(_primaryFocus!);
      }
    }
    for (final FocusNode node in _dirtyNodes) {
      node._notify();
    }
```

I guess because the topic focus node is the `previousFocus`, and content focus node is the `_primaryFocus`, `_topicFocusChanged` always gets called before `_contentFocusChanged` when the focus moves from topic to content.

PIG208: It should be fine to set it to `hasChosen` twice; once in the else branch, another time in `_contentFocusChanged`.

However, `_contentFocusChanged` still needs to be kept. It handles the case when the focus moves from nothing to content input.

lib/widgets/compose_box.dart
```diff
@@ -630,11 +641,11 @@ class _StreamContentInputState extends State<_StreamContentInput> {
         // The chosen topic can't be sent to, so don't show it.
         return null;
       }
-      if (!widget.controller.contentFocusNode.hasFocus) {
-        // Do not fall back to a vacuous topic unless the user explicitly chooses
-        // to do so (by skipping topic input and moving focus to content input),
-        // so that the user is not encouraged to use vacuous topic when they
-        // have not interacted with the inputs at all.
+      if (widget.controller.topicInteractionStatus.value !=
+          ComposeTopicInteractionStatus.hasChosen) {
```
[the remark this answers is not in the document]
gnprice: bump

lib/widgets/compose_box.dart
```diff
@@ -630,11 +641,11 @@ class _StreamContentInputState extends State<_StreamContentInput> {
...
+        // Do not fall back to a vacuous topic unless the user explicitly
```
gnprice, line 646: compose: Change content input hint text statefully
    
    Before, the content input shows the "#stream > topic" hint text
    as long as it has focus, and set the hint text to "#stream" when it
    loses focus.
    
    Now, the content input still shows "#stream > topic" when it gains
    focus, except that it will keep showing it even after losing focus,
    until the user moves focus to the topic input.

This change only affects the case where the topic is vacuous, right? As written, the commit message sounds like it applies to an arbitrary topic.

I think replacing "topic" with "general chat" in the examples would go most of the way to clearing that up. Also add a condition in the first sentence (the "Before, …").

lib/model/autocomplete.dart
```diff
@@ -904,7 +904,11 @@ class TopicAutocompleteView extends AutocompleteView<TopicAutocompleteQuery, Top
   Future<void> _fetch() async {
      assert(!_isFetching);
     _isFetching = true;
-    final result = await getStreamTopics(store.connection, streamId: streamId);
+    final result = await getStreamTopics(store.connection, streamId: streamId,
+      allowEmptyTopicName:
+        // TODO(server-10): simplify this condition away
+        store.zulipFeatureLevel >= 334 ? true : null,
```
gnprice, line 910: We can simplify up front by just passing this unconditionally — generally the server ignores unexpected parameters in requests.

Similar story to what I wrote here (to Chris):
https://github.com/zulip/zulip-flutter/pull/1484#discussion_r2072200834

lib/api/route/channels.dart
```diff
@@ -7,8 +7,12 @@ part 'channels.g.dart';
 /// https://zulip.com/api/get-stream-topics
 Future<GetStreamTopicsResult> getStreamTopics(ApiConnection connection, {
   required int streamId,
+  bool? allowEmptyTopicName,
 }) {
-  return connection.get('getStreamTopics', GetStreamTopicsResult.fromJson, 'users/me/$streamId/topics', {});
+  assert(allowEmptyTopicName != false, '`allowEmptyTopicName` should only be true or null');
```
gnprice, line 12: Then let's have these actually assert that it's true. Otherwise it feels a bit error-prone to have to remember to pass this flag at each call site — because if you don't, things will *mostly* work and you'll just get different semantics in a edge case, semantics incompatible with what the rest of the app is doing.

gnprice (commented):
Thanks for the revision! Those first 5 commits now all look good. I've finished reading the rest; just a few comments below.

lib/model/autocomplete.dart
PIG208: Oh, that's neat! I suppose we make `allowEmptyTopicName` required for all bindings that have it, and always pass `true`.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Looks good; merging.

PR #1366, opened 2025-02-19
actions [nfc]: Namespace actions as statics on a ZulipAction class, and document

gnprice:
This makes it explicit at each call site that the method being called
is an action, not (for example) an API endpoint binding.  The
difference is important in particular because it affects how --
really, whether -- the caller should handle errors.

See discussion from when the similarity in appearance to API endpoint
bindings caused confusion:
  https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/.23F1317.20showErrorDialog/near/2080570

Also document the two methods here (other than the private helper) that didn't already have docs.


chrisbobbe:
Thanks, LGTM! Merged.

PR #1367, opened 2025-02-20
compose [nfc]: Simplify content-input state, deduplicating topic text

gnprice:
Ever since 5827512af made the text input's controller start memoizing `textNormalized` as part of its own state, there's nothing to gain by keeping a copy of it on the widget state too.

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -544,18 +544,15 @@ class _StreamContentInput extends StatefulWidget {
 }
 
 class _StreamContentInputState extends State<_StreamContentInput> {
-  late String _topicTextNormalized;
-
   void _topicChanged() {
     setState(() {
-      _topicTextNormalized = widget.controller.topic.textNormalized;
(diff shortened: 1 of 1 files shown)


lib/widgets/compose_box.dart
```diff
@@ -544,18 +544,15 @@ class _StreamContentInput extends StatefulWidget {
...
 
 class _StreamContentInputState extends State<_StreamContentInput> {
-  late String _topicTextNormalized;
-
   void _topicChanged() {
     setState(() {
-      _topicTextNormalized = widget.controller.topic.textNormalized;
+      // The relevant state lives on widget.controller.topic itself.
     });
   }
```
PIG208, line 551: nit: while we are on this, the lifecycle methods (`initState` and etc.) should probably go before `_topicChanged`

gnprice: Yeah, I think either way is fine. It's a helper method used by all the lifecycle methods, so either before or after them are reasonable homes.

PIG208:
LGTM! Merged.

PR #1370, opened 2025-02-21
combined feed: Fix channel header tap behaviour in combined feed

### LINKED ISSUE #1368: combined feed: Tap behavior inconsistent in the channel portion of recipient headers
When you tap anywhere in the channel part of a recipient header—

Image

it should take you to the channel feed. If your tap is in that area but not on the channel icon "#", the channel name, or the ">", then we actually take you to the page for the specific topic. That's confusing, because you meant to go to the channel.

This will also be relevant for the channel action sheet, which will be offered on long-press in this area:

- #347 
- #1223 
- #1224 
- #1226 
- #1227 

## Implementation

I believe the bug can be fixed by adding a line to the relevant `GestureDetector`:

```diff
diff --git lib/widgets/message_list.dart lib/widgets/message_list.dart
index 48c0d70b5..7d51d6fbd 100644
--- lib/widgets/message_list.dart
+++ lib/widgets/message_list.dart
@@ -1080,6 +1080,7 @@ class StreamMessageRecipientHeader extends StatelessWidget {
         ?? zulipLocalizations.unknownChannelName; // TODO(log)
 
       streamWidget = GestureDetector(
+        behavior: HitTestBehavior.opaque,
         onTap: () => Navigator.push(context,
           MessageListPage.buildRoute(context: context,
             narrow: ChannelNarrow(message.streamId))),
```

We also need a widget test for the fix; we won't merge a PR without this. The test should pass after the fix and fail before the fix.

E-m-i-n-e-n-c-e:
# Description

This PR fixes the channel header navigation in the combined feed view

## Changes

- Modified the `GestureDetector` in `StreamMessageRecipientHeader` to use `HitTestBehavior.opaque`
- Added a test that verifies tapping empty space (3px above the chevron icon) in the channel header area navigates to the channel feed

## Related Issues

Fixes #1368

I have verified that the test fails before the changes and the test passes after the changes


lib/widgets/message_list.dart
```diff
@@ -1082,6 +1082,7 @@ class StreamMessageRecipientHeader extends StatelessWidget {
         ?? zulipLocalizations.unknownChannelName; // TODO(log)
```
PIG208, line 1082: commit message nit:

```
    msglist: Fix channel header tap behaviour in multi-channel narrows
    
    Set gesture detecter behavior of streamWidget to HitTestBehavior.opaque
    to handle taps in empty space around the header.
    
    Also added two tests to verify the behaviour
    
    Fixes #1368.
```

Because all behavior change are expected to come with tests, the comment on "Also added two tests…" can be dropped.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+          navObservers: [navObserver]);
+
+        assert(pushedRoutes.length == 1);
+        pushedRoutes.clear();
+
+        connection.prepare(json: eg.newestGetMessagesResult(
+          foundOldest: true, messages: [message]).toJson());
+
+        final recipientHeaderFinder = find.byType(StreamMessageRecipientHeader);
+        final recipientHeaderRect = tester.getRect(recipientHeaderFinder);
```
PIG208, line 1495: Similarly, these two are shared between the two tests, so we can pull them out too.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+
+        final channelNameFinder = find.descendant(
+          of: recipientHeaderFinder,
+          matching: find.text(channel.name));
+        final channelNameRect = tester.getRect(channelNameFinder);
+
+        // First tap: 1 px below the top of recepient header at channelName x coordinate
+        await tester.tapAt(Offset(channelNameRect.center.dx, recipientHeaderRect.top + 1));
```
PIG208, line 1503:
```suggestion
        await tester.tapAt(Offset(
          channelNameRect.center.dx, recipientHeaderRect.top + 1));
```

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+
+        final recipientHeaderFinder = find.byType(StreamMessageRecipientHeader);
+        final recipientHeaderRect = tester.getRect(recipientHeaderFinder);
+
+        final channelNameFinder = find.descendant(
+          of: recipientHeaderFinder,
+          matching: find.text(channel.name));
+        final channelNameRect = tester.getRect(channelNameFinder);
+
+        // First tap: 1 px below the top of recepient header at channelName x coordinate
```
PIG208, line 1502: nit: line too long (>80):
```suggestion
        // First tap: 1 px below the top of recepient header
        // at channelName's x coordinate
```

This applies to other similar comments too.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+        final message = eg.streamMessage(stream: channel);
+        await setupMessageListPage(tester,
+          narrow: const CombinedFeedNarrow(),
+          streams: [channel],
+          subscriptions: [eg.subscription(channel)],
+          messages: [message],
+          navObservers: [navObserver]);
+
+        assert(pushedRoutes.length == 1);
+        pushedRoutes.clear();
```
PIG208, line 1489: The bulk of this setup code might fit better in a `Future<void> prepare` helper shared within the group. An advantage of doing this, other than deduplicating code, is that the tests themselves can be more focused on what's being tested.

Things like `pushedRoutes` can be defined as late variables, or like `message` as final variables.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+
+        connection.prepare(json: eg.newestGetMessagesResult(
+          foundOldest: true, messages: [message]).toJson());
```
PIG208, line 1492: Looks like this prepares for the message list page before we tap the channel name, right? Let's move it to just before the comment for that line to make it clear what this prepared response is for.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+        await tester.pumpAndSettle();
+
+        // Navigate back to original page and clear routes
+        await tester.pageBack();
+        await tester.pumpAndSettle();
+        pushedRoutes.clear();
+
+        connection.prepare(json: eg.newestGetMessagesResult(
+          foundOldest: true, messages: [message]).toJson());
+
```
PIG208, line 1517: nit: we can delete the gap because this prepares a response for what happens next immediately.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
...
+        final channel = eg.stream();
+        final message = eg.streamMessage(stream: channel);
+        await setupMessageListPage(tester,
+          narrow: const CombinedFeedNarrow(),
+          streams: [channel],
+          subscriptions: [eg.subscription(channel)],
+          messages: [message],
+          navObservers: [navObserver]);
+
+        assert(pushedRoutes.length == 1);
```
PIG208, line 1488: nit: `assert(pushedRoutes.single);`

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,108 @@ void main() {
         ..status.equals(AnimationStatus.dismissed);
     });
   });
+
+  group('channel header navigation in multi-channel narrows', () {
```
PIG208, line 1474: nit: it would be more accurate to say "recipient header" instead of "channel header"

PIG208 (commented):
Thanks for the PR! Left some comments.

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: ```pushedRoutes.single``` does not return a boolean value. I checked that all the other tests that were already written in the file also use ```assert(pushedRoutes.single);```

E-m-i-n-e-n-c-e:
Pushed the revision. PTAL., Thanks.

lib/widgets/message_list.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
PIG208: Ah right. We can do pushedRoutes.single; as an
 statement.

E-m-i-n-e-n-c-e: We could but that doesn't feel intuitive and wouldn't it be better to keep this consistent with all the other tests and just use ```assert(pushed routes.length)==1);```

PIG208: Yeah, that should be fine.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+
+      final channelNameFinder = find.descendant(
+        of: recipientHeaderFinder,
+        matching: find.text(channel.name));
+      final channelNameRect = tester.getRect(channelNameFinder);
+
+      connection.prepare(json: eg.newestGetMessagesResult(
+      foundOldest: true, messages: [message]).toJson());
+      // First tap: 1 px below the top of recepient header
+      // at channelName x coordinate
```
PIG208, line 1516: nit: typo "recepient" -> "recipient"

PIG208: Re-reading this, I feel that "channelName x coordinate" is a bit ambiguous. We can tell from the code below that this refers to the center of the `channelNameRect`.

Perhaps we want to emphasize that the tap position is above the channel name and below the recipient header. How about:
```dart
      // Tap just right below the top of recepient header, above and outside of
      // its channel name component.
```
and similarly for the other places.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+      await tester.tapAt(Offset(
+        channelNameRect.center.dx, recipientHeaderRect.top + 1));
+
+      await tester.pump();
+      check(pushedRoutes).single.isA<WidgetRoute>().page.isA<MessageListPage>()
+        .initNarrow.equals(ChannelNarrow(channel.streamId));
+      await tester.pumpAndSettle();
+
+      // Navigate back to original page and clear routes
```
PIG208, line 1525: nit: need a period at the end since it is a full sentence

PIG208 (commented):
Thanks for the update! Did another round with mostly small comments.

lib/widgets/message_list.dart
PIG208, line 1082: commit message nit:

```
msglist: Fix channel header tap behaviour in multi-channel narrows
    
    Set gesture detecter behavior of streamWidget to HitTestBehavior.opaque
    to handle taps in empty space around the header.
    
    Fixes #1368.
```

The issue number should #1179 because #1368 was closed as a duplicate.

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+      await tester.pageBack();
+      await tester.pumpAndSettle();
+      pushedRoutes.clear();
+
+      connection.prepare(json: eg.newestGetMessagesResult(
+        foundOldest: true, messages: [message]).toJson());
+      // Second tap: 1 px above the bottom of the recipient header
+      // at channelName x coordinate
+      await tester.tapAt(Offset(
+        channelNameRect.center.dx, channelNameRect.bottom - 1));
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: No its the rectangle of the channelName. We are using the same rect the second time as well tapping below the name this time

```dart
      final channelNameFinder = find.descendant(
        of: recipientHeaderFinder,
        matching: find.text(channel.name));
      final channelNameRect = tester.getRect(channelNameFinder);
   ```

File: test/widgets/message_list_test.dart [hunk 13]
```diff
@@ -1470,4 +1470,110 @@ void main() {
         ..status.equals(AnimationStatus.dismissed);
     });
   });
+
+  group('recipient header navigation in multi-channel narrows', () {
+    late List<Route<void>> pushedRoutes;
+    late TestNavigatorObserver navObserver;
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: I don't understand how that would work. Both the tests use the ```pushedRoutes``` and ```navObserver``` multiple times. If they were local to prepare(), the tests wouldn't have access to them when needed.

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Yeah this seems to be a mistake on my part. Thanks for catching it

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+    late ZulipStream channel;
+    late StreamMessage message;
+    late Finder recipientHeaderFinder;
+    late Rect recipientHeaderRect;
+    const testTopic = 'testTopic';
+
+    Future<void> prepare(WidgetTester tester) async {
+      pushedRoutes = <Route<void>>[];
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
E-m-i-n-e-n-c-e: Alright

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
   });
+
+  group('recipient header navigation in multi-channel narrows', () {
+    late List<Route<void>> pushedRoutes;
+    late TestNavigatorObserver navObserver;
+    late ZulipStream channel;
+    late StreamMessage message;
+    late Finder recipientHeaderFinder;
+    late Rect recipientHeaderRect;
+    const testTopic = 'testTopic';
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Done

lib/widgets/message_list.dart
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+
+      assert(pushedRoutes.length == 1);
+      pushedRoutes.clear();
+
+      recipientHeaderFinder = find.byType(StreamMessageRecipientHeader);
+      recipientHeaderRect = tester.getRect(recipientHeaderFinder);
+    }
+
+    testWidgets("navigates to ChannelNarrow when tapping above or below channel name in recipient header",
+     (tester) async {
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: OK

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
         ..status.equals(AnimationStatus.dismissed);
     });
   });
+
+  group('recipient header navigation in multi-channel narrows', () {
+    late List<Route<void>> pushedRoutes;
+    late TestNavigatorObserver navObserver;
+    late ZulipStream channel;
+    late StreamMessage message;
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Alright

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+        matching: find.text(channel.name));
+      final channelNameRect = tester.getRect(channelNameFinder);
+
+      connection.prepare(json: eg.newestGetMessagesResult(
+      foundOldest: true, messages: [message]).toJson());
+      // First tap: 1 px below the top of recepient header
+      // at channelName x coordinate
+      await tester.tapAt(Offset(
+        channelNameRect.center.dx, recipientHeaderRect.top + 1));
+
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+        narrow: const CombinedFeedNarrow(),
+        streams: [channel],
+        subscriptions: [eg.subscription(channel)],
+        messages: [message],
+        navObservers: [navObserver]);
+
+      assert(pushedRoutes.length == 1);
+      pushedRoutes.clear();
+
+      recipientHeaderFinder = find.byType(StreamMessageRecipientHeader);
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Done

test/widgets/message_list_test.dart
```diff
@@ -1470,4 +1470,110 @@ void main() {
...
+     (tester) async {
+      await prepare(tester);
+
+      final channelNameFinder = find.descendant(
+        of: recipientHeaderFinder,
+        matching: find.text(channel.name));
+      final channelNameRect = tester.getRect(channelNameFinder);
+
+      connection.prepare(json: eg.newestGetMessagesResult(
+      foundOldest: true, messages: [message]).toJson());
```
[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Done

E-m-i-n-e-n-c-e:
I have pushed a revision addressing all the comments except for one comment I couldn't understand which I left unresolved. PTAL. Thanks!

test/widgets/message_list_test.dart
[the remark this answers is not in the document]
PIG208: I meant that we can make `navObserver` local to the helper, since the tests do not need access to it.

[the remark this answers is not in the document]
E-m-i-n-e-n-c-e: Done

E-m-i-n-e-n-c-e:
Pushed another update. This one addresses all the review comments

PIG208:
Thanks for the update! This LGTM; marking for Greg's review. Because this resolves a post-launch issue, expect that it might take longer for us to eventually merge this while we prioritize launch tasks.

gnprice:
Thanks @E-m-i-n-e-n-c-e for taking care of this, and @PIG208 for all the previous reviews!

Just picking this up now, and it looks good; merging. I appreciate the careful tests which should make sure that any future changes to this layout don't break this without us noticing.

Fixed one nit in the commit message:

    -    msglist: Fix channel header tap behaviour in multi-channel narrows
    +    msglist: Fix channel header tap behavior in multi-channel narrows

The American spelling "behavior" is what's in the API we're using, and it's generally the standard in Flutter upstream and in Zulip.

Issue #1356, opened 2025-02-14
Fix ordered list layout to handle long numbers in `<ol>`


lakshya1goel:
The ordered list layout currently uses a **fixed width of 20px** for the marker column:  

```dart
width: 20, // TODO handle long numbers in <ol>, like https://github.com/zulip/zulip/pull/25063
```

This can cause layout issues when the list starts with large numbers (e.g., "9999. first")

### Observed Behavior:

The **dot (.) moves to the next line** when the marker has **2 or more digits**.
Example: **"10." appears misaligned**, with the dot shifting below the number in the screenshot below.

<img src="https://github.com/user-attachments/assets/bf47e73e-b8cc-46fb-8c67-4da14b428892" height="500">

### Expected Behavior:

The marker column should **adjust dynamically to accommodate longer numbers** without breaking the layout.

shkshreyas:
I want to handle this

shkshreyas:
[https://github.com/zulip/zulip-flutter/pull/1372](Solved The Issue)

lakshya1goel:
Hi @shkshreyas, there is already an open PR #1329 under review to fix this issue, Please have a look. Thanks!

PR #1372, opened 2025-02-22
issue 1356 fixed

shkshreyas:
Before
image

After
image


gnprice:
Before we can review a PR, it needs to have tests, and it needs to have clear and coherent commits free of unrelated changes. See our README, which also links to resources with more details.

Before asking other people to spend time reviewing your work, you should also take time to review it yourself:
In addition to the problems mentioned above, this PR has an unused duplicate copy of its code in another file, complete with a comment that looks like it's something an LLM wrote for you to read — not something that makes sense in the context of a PR:
```dart
// Example usage in a widget
Widget buildOrderedList(List<String> items) {
```
And the "After" screenshot shows the two-digit numbers very far out of alignment with the one-digit numbers. Both of those are things you definitely should have been able to catch for yourself.

In this case, there's already a PR #1329 which fixes this issue #1356 and is near merge. So there's no need for another PR for this, and I'm closing this one.

PR #1373, opened 2025-02-22
notif: Preserve target account context on back navigation after opening a notification

### LINKED ISSUE #1210: android: When opening a notification, "back" should not leave the account's context

lakshya1goel:
Fixes: #1210

### Description
When opening a notification from a different account B while in account A and pressing back, the app should stay in account B's context rather than returning to account A. This PR implements this behavior by:

1. Checking if the current route belongs to a different account than the notification's target account
2. If so, navigating to the target account's homepage before pushing the message list
3. This ensures the navigation stack has the proper account context, making back navigation intuitive

### Video Demonstration
https://github.com/user-attachments/assets/b0d41ef9-324c-45e6-8cab-af8116633c2a

### DIFF
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -17,6 +17,7 @@ import '../model/narrow.dart';
 import '../widgets/app.dart';
 import '../widgets/color.dart';
 import '../widgets/dialog.dart';
+import '../widgets/home.dart';
 import '../widgets/message_list.dart';
 import '../widgets/page.dart';
 import '../widgets/store.dart';
@@ -214,8 +215,34 @@ class NotificationChannelManager {
   }
 }
 
+/// Tracks the currently active account in the navigation stack
+class AccountNavigationObserver extends NavigatorObserver {
+  int? _activeAccountId;
+  int? get activeAccountId => _activeAccountId;
+
+  @override
+  void didPush(Route<dynamic> route, Route<void>? previousRoute) {
+    _updateActiveAccount(route);
+  }
+
+  @override
+  void didReplace({Route<dynamic>? newRoute, Route<void>? oldRoute}) {
+    if (newRoute != null) {
+      _updateActiveAccount(newRoute);
+    }
+  }
+
+  void _updateActiveAccount(Route<void> route) {
+    if (route is AccountRoute) {
+      _activeAccountId = route.accountId;
+    }
+  }
+}
+
 /// Service for managing the notifications shown to the user.
 class NotificationDisplayManager {
+  static final accountObserver = AccountNavigationObserver();
+
   static Future<void> init() async {
     await NotificationChannelManager.ensureChannel();
   }
@@ -502,7 +529,14 @@ class NotificationDisplayManager {
     final route = routeForNotification(context: context, url: url);
     if (route == null) return; // TODO(log)
 
-    // TODO(nav): Better interact with existing nav stack on notif open
+    final currentAccountId = accountObserver.activeAccountId;
+
+    if (currentAccountId != route.accountId) {
+      navigator.popUntil((r) => r.isFirst);
+      HomePage.navigate(context, accountId: route.accountId);
+      navigator = await ZulipApp.navigator;
+    }
+
     unawaited(navigator.push(route));
   }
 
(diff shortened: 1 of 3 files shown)


lib/notifications/display.dart
```diff
@@ -502,6 +503,18 @@ class NotificationDisplayManager {
...
+    navigator.popUntil((r) {
+      topRoute = r;
+      return true;
+    });
+
+    if (topRoute == null || topRoute is! AccountRoute
+        || (topRoute as AccountRoute).accountId != route.accountId) {
+      HomePage.navigate(context, accountId: route.accountId);
+      navigator = await ZulipApp.navigator;
+    }
```
PIG208, line 516: The `popUntil` above unconditionally clears the navigator stack until there is only one route left. I believe that this is equivalent to `HomePage.navigate`, which also pops all routes beforehand.

Either way, we should not pop all routes unconditionally, because it doesn't feel like great user experience when the existing routes already belong to the same account.

#F1210 suggests that we 

> switch to a new nav stack for the notification's account (if different) before pushing the message-list route.

I think what we need here is just a way to keep track of the currently active account, or a way to inspect the navigator stack without actually popping the routes.

PIG208: See discussion here: [#mobile > Beta: after open via notification, back opens old server @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Beta.3A.20after.20open.20via.20notification.2C.20back.20opens.20old.20server/near/2105586). Feel free to ask questions if you have any!

lakshya1goel: Based on my current implementation, I have addressed the following:
1.  Inspect the navigation stack without modifying it by using `popUntil` with `return true`.
2. Only switch accounts (via `HomePage.navigate`) when the notification is for a different account than the current one (`currentAccountId != route.accountId`).
3. When the notification is for the current account, preserve the existing navigation stack and simply push the new route on top, providing a better user experience.

lakshya1goel:
Hello @PIG208, I have pushed the revision. PTAL, Thanks!

lakshya1goel:
Hi @PIG208 , whenever you get some time, please take a look at the PR. Thanks!

lib/notifications/display.dart
```diff
@@ -502,7 +503,19 @@ class NotificationDisplayManager {
...
-    // TODO(nav): Better interact with existing nav stack on notif open
+    int? currentAccountId;
+    navigator.popUntil((r) {
+      if (r is AccountRoute) {
+        currentAccountId = r.accountId;
+      }
+      return true;
+    });
+
+    if (currentAccountId != route.accountId) {
```
PIG208, line 514: When `currentAccountId` (i.e., account ID associated with the top-most route) is `null`, routes below the top-most route can still belong to the same account. `HomePage.navigate` will pop those routes even though they belong to same account that receives this notification.

PIG208 (commented):
Thanks for the update! I think a full implementation might need to be a bit more complex, since the top-most route might not always tell us which account "context" we are currently in.

lakshya1goel:
Hi @PIG208, I have pushed the revision. PTAL, Thanks!

chrisbobbe:
Closing as superseded by #1993. Thanks for your work @lakshya1goel!

Issue #1083, opened 2024-11-25
Include stack trace when reporting MalformedServerResponseException


gnprice:
When a request to the server fails, we typically show an error to the user. (The main gap in that is #890, which we should fix.) The details of the error can be useful, particularly if the user takes a screenshot of them to include in reporting the issue to us.

If the cause of the error is that the server's response doesn't match our expectations — a MalformedServerResponseException — then a key thing we want to know is where in the schema the mismatch occurred: what field of what type of object. That information isn't in the error message itself, but it is in the stack trace. For example see [this error](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Infinite.20loading.20flutter.20app/near/1987298) caused by #1082.

So we should show the stack trace when reporting a MalformedServerResponseException. This is most important for a registerQueue request (#890) just because that's the most sprawling data schema; but it can matter for other request types too.

lakshya1goel:
Hello, I have opened PR #1374 for fixing this issue - PTAL, Thanks!

PR #1374, opened 2025-02-23
api: Include stack trace when reporting MalformedServerResponseException

lakshya1goel:
Fixes: #1083

### Changes:
* Add `causeStackTrace` field to `MalformedServerResponseException`
* Include stack trace in the exception's `toString()` output
* Add test to verify stack trace is included in error messages

### Testing:
* Added unit test that verifies stack trace is included when a type mismatch occurs
* Manually tested with registerQueue to confirm stack trace shows in error dialog

### DIFF
--- a/lib/api/exception.dart
+++ b/lib/api/exception.dart
@@ -149,16 +149,34 @@ class MalformedServerResponseException extends ServerException {
   /// in order to preserve the underlying exception's stack trace, which
   /// may be more informative than the exception object itself.
   final Object? causeException;
+  final StackTrace? causeStackTrace;
 
   MalformedServerResponseException({
     required super.routeName,
     required super.httpStatus,
     required super.data,
     this.causeException,
+    this.causeStackTrace,
   }) : super(message: causeException == null
          ? GlobalLocalizations.zulipLocalizations
             .errorMalformedResponse(httpStatus)
          : GlobalLocalizations.zulipLocalizations
             .errorMalformedResponseWithCause(
               httpStatus, causeException.toString()));
+
+  @override
+  String toString() {
+    final sb = StringBuffer();
+    sb.write('${objectRuntimeType(this, 'MalformedServerResponseException')}:');
+    sb.write(" $httpStatus");
+    sb.write(" $routeName");
+    sb.write(": $message");
+    if (causeException != null) {
+      sb.write('\nCaused by: $causeException');
+      if (causeStackTrace != null) {
+        sb.write('\n$causeStackTrace');
+      }
+    }
+    return sb.toString();
+  }
 }
--- a/test/api/exception_test.dart
+++ b/test/api/exception_test.dart
@@ -1,4 +1,7 @@
+import 'dart:convert';
+
 import 'package:checks/checks.dart';
+import 'package:test/expect.dart';
 import 'package:test/scaffolding.dart';
 import 'package:zulip/api/exception.dart';
 
@@ -25,4 +28,40 @@ void main() {
   });
 
   // NetworkException.toString: see "API network errors" test in core_test.dart
+
+  test('MalformedServerResponseException shows JSON parsing location', () {
+    const jsonResponse = '''
+      {
+        "result": "success",
+        "queue_id": "1234567890",
+        "last_event_id": -1,
+        "properties": {
+          "realm_uri": "https://example.zulipchat.com"
+        }
+      }
+    ''';
+
+    try {
+      Map<String, String> wrongType =
+        (jsonDecode(jsonResponse) as Map<String, dynamic>)
+          .map((key, value) => MapEntry(key, value.toString()));
+
+      wrongType['properties']! as Map<String, dynamic>;
+      fail('Should have thrown');
+    } catch (e, stackTrace) {
+      final exception = MalformedServerResponseException(
+        routeName: 'registerQueue',
+        httpStatus: 200,
+        data: jsonDecode(jsonResponse) as Map<String, dynamic>,
+        causeException: e,
+        causeStackTrace: stackTrace,
+      );
+
+      final string = exception.toString();
+      check(string)
+        ..contains('type \'String\'')
+        ..contains('Map<String, dynamic>')
+        ..contains(stackTrace.toString());
+    }
+  });
 }


gnprice:
It doesn't look like this new parameter `causeStackTrace` ever gets passed to the constructor. So it doesn't actually do anything in the live app.

Please test your change end to end in the app, and revise it so it solves the problem described in the issue:
> When a request to the server fails, we typically show an error to the user. (The main gap in that is https://github.com/zulip/zulip-flutter/issues/890, which we should fix.) The details of the error can be useful, particularly if the user takes a screenshot of them to include in reporting the issue to us.
>
> If the cause of the error is that the server's response doesn't match our expectations — a MalformedServerResponseException — then a key thing we want to know is where in the schema the mismatch occurred: what field of what type of object. […] So we should show the stack trace when reporting a MalformedServerResponseException.

test/api/exception_test.dart
```diff
@@ -25,4 +28,40 @@ void main() {
...
+
+    try {
+      Map<String, String> wrongType =
+        (jsonDecode(jsonResponse) as Map<String, dynamic>)
+          .map((key, value) => MapEntry(key, value.toString()));
+
+      wrongType['properties']! as Map<String, dynamic>;
+      fail('Should have thrown');
+    } catch (e, stackTrace) {
+      final exception = MalformedServerResponseException(
```
PIG208, line 52: A more end-to-end test will be more valuable. `MalformedServerResponseException` already has a bunch of tests in `test/api/core_test.dart`.

lib/api/exception.dart
```diff
@@ -149,16 +149,34 @@ class MalformedServerResponseException extends ServerException {
...
     required super.httpStatus,
     required super.data,
     this.causeException,
+    this.causeStackTrace
   }) : super(message: causeException == null
          ? GlobalLocalizations.zulipLocalizations
             .errorMalformedResponse(httpStatus)
          : GlobalLocalizations.zulipLocalizations
             .errorMalformedResponseWithCause(
               httpStatus, causeException.toString()));
```
PIG208, line 165: We already stringify `causeException` and make it a part of `message`. Including it again in the `toString` implementation will repeat that in the error message. This is what I got when printing out the error from the test:

```
MalformedServerResponseException: 200 registerQueue: Server gave malformed response; HTTP status 200; type 'String' is not a subtype of type 'Map<String, dynamic>' in type cast
Caused by: type 'String' is not a subtype of type 'Map<String, dynamic>' in type cast
#0      main.<anonymous closure> (file:///home/zixuan/zulip-stuff/zulip-flutter/test/api/exception_test.dart:49:32)
#1      Declarer.test.<anonymous closure>.<anonymous closure> (package:test_api/src/backend/declarer.dart:229:19)
<asynchronous suspension>
#2      Declarer.test.<anonymous closure> (package:test_api/src/backend/declarer.dart:227:7)
<asynchronous suspension>
#3      Invoker._waitForOutstandingCallbacks.<anonymous closure> (package:test_api/src/backend/invoker.dart:258:9)
<asynchronous suspension>
true
```

lib/api/exception.dart
```diff
@@ -149,16 +149,34 @@ class MalformedServerResponseException extends ServerException {
...
+
+  @override
+  String toString() {
+    final sb = StringBuffer();
+    sb.write('${objectRuntimeType(this, 'MalformedServerResponseException')}:');
+    sb.write(" $httpStatus");
+    sb.write(" $routeName");
+    sb.write(": $message");
+    if (causeException != null) {
+      sb.write('\nCaused by: $causeException');
```
PIG208, line 175: New strings added should be localized. We already have `errorMalformedResponseWithCause` and `errorMalformedResponse`, so it might be a good idea to reuse them/add one.

lib/api/exception.dart
```diff
@@ -149,16 +149,34 @@ class MalformedServerResponseException extends ServerException {
...
+      if (causeStackTrace != null) {
```
PIG208, line 176: Looks like it would be a developer error if `causeException` is present but `causeStackTrace` is not. Let's have an assertion in the constructor to validate that.

test/api/exception_test.dart
PIG208: We should then leave a comment similar to this one above:

```
// NetworkException.toString: see "API network errors" test in core_test.dart
```

to indicate that the tests live elsewhere.

PIG208 (commented):
Thanks for working on this! Left some comments.

lakshya1goel:
Hi @PIG208, thanks for the above reviews. I have pushed the revision. PTAL.

assets/l10n/app_en.arb
```diff
@@ -552,6 +552,24 @@
       "details": {"type": "String", "example": "type 'Null' is not a subtype of type 'String' in type cast"}
     }
   },
+  "errorMalformedResponseFormat": "{className}: {httpStatus} {routeName}: {message}",
+  "@errorMalformedResponseFormat": {
+    "description": "Format for error messages from malformed server responses. The message parameter is already localized.",
```
PIG208, line 557: nit: the audience of the string description is mainly the translators; so there is nothing actionable for them knowing that `message` is localized. Similarly, we can leave out "format" in both the name of the string and its description.

How about naming it "errorMalformedServerResponse"? For rewriting the description, you can find soem examples with:

`git grep -hEA2 'error.+\",' assets/l10n/app_en.arb`

You can check out `git grep --help` for what each of these options mean.

assets/l10n/app_en.arb
```diff
@@ -552,6 +552,24 @@
...
+    }
+  },
+  "errorMalformedServerResponseExceptionName": "MalformedServerResponseException",
+  "@errorMalformedServerResponseExceptionName": {
+    "description": "Name of the error type for malformed server responses"
+  },
+  "errorMalformedServerResponseExceptionAssertMessage": "If causeException is provided, causeStackTrace must also be provided",
+  "@errorMalformedServerResponseExceptionAssertMessage": {
+    "description": "Developer error message when creating a MalformedServerResponseException with incomplete error details"
+  },
```
PIG208, line 572: Because assertions are only run in debug builds, messages from them are usually only visible to the developer. We can skip translating those strings.

assets/l10n/app_en.arb
```diff
@@ -552,6 +552,24 @@
...
+      "className": {"type": "String", "example": "MalformedServerResponseException"},
+      "httpStatus": {"type": "int", "example": "200"},
+      "routeName": {"type": "String", "example": "registerQueue"},
+      "message": {"type": "String", "example": "Server gave malformed response; HTTP status 200"}
+    }
+  },
+  "errorMalformedServerResponseExceptionName": "MalformedServerResponseException",
+  "@errorMalformedServerResponseExceptionName": {
+    "description": "Name of the error type for malformed server responses"
+  },
```
PIG208, line 568: Hmm, it doesn't seem that a class name that is only meaningful to our codebase can be translated.

lib/api/exception.dart
```diff
@@ -149,16 +149,30 @@ class MalformedServerResponseException extends ServerException {
...
               httpStatus, causeException.toString()));
+
+  @override
+  String toString() {
+    final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+    return zulipLocalizations.errorMalformedResponseFormat(
+      zulipLocalizations.errorMalformedServerResponseExceptionName,
+      httpStatus,
+      routeName,
+      message);
```
PIG208, line 176: Looks like the stack trace is not used in this revision. We should include it in the error message.

I think the structure we had in the previous revision is good. https://github.com/zulip/zulip-flutter/pull/1374#discussion_r1978046314 was for the untranslated "Cause by". Since we have removed it in favor of `errorMalformedResponseWithCause`, the remaining parts of the error message probably don't need translations.

PIG208:
Thanks for the update! Left some comments.

lakshya1goel:
Hi @PIG208, thanks for the review. I have pushed the revision. PTAL!

lakshya1goel:
Hi @PIG208 , whenever you get some time, please take a look at the PR. Thanks!

lib/api/exception.dart
PIG208: Hmm, I see that you added "Caused by" back. It's an English phrase that should be translated.

lib/api/exception.dart
```diff
@@ -149,16 +149,32 @@ class MalformedServerResponseException extends ServerException {
...
+  String toString() {
+    final StringBuffer sb = StringBuffer('MalformedServerResponseException');
+    sb.write(" $routeName");
+    sb.write(": $message");
+    if (causeException != null) {
+      sb.write('\nCaused by: $causeException');
+      sb.write('\n$causeStackTrace');
+    }
+    return sb.toString();
+  }
```
PIG208, line 179: Hmm, I see that you added "Caused by" back. It's an English phrase that should be translated. However, maybe we don't need to override `toString` at all? `ZulipApiException` overrides it because it has multiple checks on fields like `httpStatus`, `code`, and `data`. But for `MalformedServerResponseException`, we are only checking `causeException`.

The original code seems capable of handling this:

```
  MalformedServerResponseException({
    required super.routeName,
    required super.httpStatus,
    required super.data,
    this.causeException,
  }) : super(message: causeException == null
         ? GlobalLocalizations.zulipLocalizations
            .errorMalformedResponse(httpStatus)
         : GlobalLocalizations.zulipLocalizations
            .errorMalformedResponseWithCause(
              httpStatus, causeException.toString()));
```

We just need to add the stack trace to the case where `causeException` is non-null, and have the assertions here to ensure that both of these fields are non-null or null at the same time.

test/api/core_test.dart
```diff
@@ -430,6 +430,29 @@ void main() {
     }
   });
 
+  test('MalformedServerResponseException includes cause details in toString', () async {
+    try {
+      throw FormatException("Expected field 'stream_id' to be an integer, but got string 'abc123'");
+    } catch (e, st) {
+      final exception = MalformedServerResponseException(
```
PIG208, line 437: Hmm, it would be ideal for the `MalformedServerResponseException` to be thrown by non-test code, as this is the point of having this test in `core_test.dart`.

I think an existing test, "malformed API success responses: exception preserves details", has all the setup needed to test this correctly.

PIG208 (commented):
Thanks for the update! Left some more comments

test/api/core_test.dart
lakshya1goel: Thanks for pointing. So should I update the exiting test only to check for the `causeStackTrace` as well or need to add a separate test for it?

Updated test can be:
```
test('malformed API success responses: exception preserves details', () async {
    int distinctivelyNamedFromJson(Map<String, dynamic> json) {
      throw DistinctiveError("something is wrong");
    }

    try {
      await tryRequest(json: {}, fromJson: distinctivelyNamedFromJson);
      assert(false);
    } catch (e, st) {
      check(e).isA<MalformedServerResponseException>()
        ..causeException.isA<DistinctiveError>()
        ..message.contains("something is wrong")
        ..has((it) => it.causeStackTrace, "causeStackTrace").isNotNull()
        ..has((it) => it.causeStackTrace.toString(), "causeStackTrace.toString()")
          .contains("distinctivelyNamedFromJson");
      check(st.toString()).contains("distinctivelyNamedFromJson");
    }
  });
```

PIG208: Updating the existing test sounds good to me. I would add `causeStackTrace` to the …Checks extension of the error, to avoid inlining the `has`' here.

lakshya1goel:
Hi @PIG208, I have pushed the revision PTAL, Thanks!

PIG208 (commented):
Thanks for the update! Could you also include a screenshot of what an error message would look like with the next revision? Left some comments.

lakshya1goel:
Hi @PIG208, I have pushed the revision as well as updates the PR description with the screenshot of error. PTAL, thanks!

test/api/exception_checks.dart
```diff
@@ -28,4 +28,5 @@ extension Server5xxExceptionChecks on Subject<Server5xxException> {
 
 extension MalformedServerResponseExceptionChecks on Subject<MalformedServerResponseException> {
   Subject<Object?> get causeException => has((e) => e.causeException, 'causeException');
+  Subject<String> get causeStackTrace => has((e) => e.causeStackTrace.toString(), 'causeStackTrace');
```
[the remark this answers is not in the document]
PIG208: And then, we can have `StackTraceChecks` next to `ErrorChecks` in `test/stdlib_checks.dart` that offers `asString` on `Subject<StackTrace>`.

test/api/core_test.dart
```diff
@@ -425,7 +425,8 @@ void main() {
     } catch (e, st) {
       check(e).isA<MalformedServerResponseException>()
         ..causeException.isA<DistinctiveError>()
-        ..message.contains("something is wrong");
+        ..message.contains("something is wrong")
+        ..causeStackTrace.toString().contains("distinctivelyNamedFromJson");
```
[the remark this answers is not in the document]
PIG208: Calling `toString()` on a `Subject` also causes this to not fail when the expected substring is not found, because we intended to call `Subject.contains`, but this actually calls `String.contains`, which just returns a `bool` and does not throw.

PIG208 (commented):
Thanks for the update! I think the test still needs some tweaks, but looks a lot closer now.

One other note: I think I wasn't clear in my previous https://github.com/zulip/zulip-flutter/pull/1374#pullrequestreview-2727138368 that I was referring to taking a screenshot of the error the user might see while using the app.

PIG208:
I think we shouldn't need to add a snack bar/dialog, since the error should have been handled from the polling loop in `UpdateMachine`.

Such an error might look like this:

image

test/stdlib_checks.dart
```diff
@@ -30,6 +30,10 @@ extension ErrorChecks on Subject<Error> {
   Subject<String> get asString => has((x) => x.toString(), 'toString'); // TODO(checks): what's a good convention for this?
 }
 
+extension StackTraceChecks on Subject<StackTrace> {
+  Subject<String> get asString => has((x) => x.toString(), 'toString');
```
PIG208, line 34: nit:
```suggestion
  Subject<String> get asString => has((x) => x.toString(), 'toString'); // TODO(checks): what's a good convention for this?
```

(removed trailing space and copied the comment from the other `asString` check)

PIG208 (commented):
Thanks for the update! Marking this for Greg's review.

gnprice:
As @PIG208 said above, I think we shouldn't need to add any new snack bar or dialog in order for the error to get shown to the user — the existing error dialog should do it.

But as the issue #1083 says, and as I emphasized in https://github.com/zulip/zulip-flutter/pull/1374#issuecomment-2679677776 at the top of this thread, the point of the issue is about what information we show to the user.

So, does this PR cause the stack trace to get shown to the user when they hit one of these errors? Please post a screenshot of what that looks like.

(The PR description currently has a screenshot of something shown in your terminal, as a developer. That isn't informative about what the user sees. If it were relevant information, then copy-paste into a code block would be a clearer way to present it than a screenshot.)

chrisbobbe:
Hi @lakshya1goel, please let us know if you're planning to update this following Greg's comment above.

gnprice:
We've resolved this issue, in #1949, so closing this PR.

PR #1377, opened 2025-02-25
Upgrade Flutter and packages

rajveermalviya:
There are some code formatting changes in generated files because of `package:dart_style` being bumped, which affects the diff a bit.


chrisbobbe:
Thanks, LGTM! Have you done a quick manual smoke-test of the app, including code that exercises Pigeon and `file_picker`, since those were major-version bumps?

rajveermalviya:
Thanks for the review @chrisbobbe! Yes, I'd done manual smoke tests.

On iOS Simulator, verified that APNs token is retrieved to test Firebase SDK updates, tested file picker and some small tests (browsing through, reopen app, login) to test other updates (drift, sqlite, video player).

On Android device, tested notifications (foreground and background) to test pigeon and firebase updates, tested file picker and similar small tests as done for iOS.

gnprice:
Thanks for taking care of these!

For the pigeon upgrade and the resulting dart_style upgrade, I took a look and found I could split out the dart_style upgrade as a prep commit, so I've done that. See the resulting commits:
a15d6b42f deps: Update dart_style to 3.0.1, from 2.x
249875dc5 deps: Update pigeon to 24.2.1, from 22.7.4

Otherwise this all looks great. Merging, with that change and a couple of added notes in commit messages with the result of reading the changelogs.

PR #1261, opened 2025-01-07
notif ios: Handle opening of conversation on tap

rajveermalviya:
Fixes: #1147

PR #1379, opened 2025-02-26
notif ios: Handle opening of conversation on tap; take 2

### LINKED ISSUE #1147: ios notifs: Support tapping a notification to open the conversation

rajveermalviya:
Fixes #1147.

2nd attempt, first attempt was #1261. This one uses pigeon to move most of the notification payload handling to dart side (and doesn't use/rely on `zulip://notification` URL).


chrisbobbe:
Ah this has gathered a conflict in lib/widgets/app.dart; could you resolve it please? (I see you did a few days ago, but looks like it's happened again; thanks. 🙂)

rajveermalviya:
(Rebased to main, Thanks!)

chrisbobbe (commented):
Thanks! There's a lot here, and I haven't gotten around to it all today. But here are some comments from an initial review.

test/widgets/store_test.dart
```diff
@@ -246,17 +273,23 @@ void main() {
...
-              accountId: accountId,
-              child: MyWidgetWithMixin(key: widgetWithMixinKey)))));
+          home: DeferrredBuilderWidget(
+            future: globalStoreFuture,
+            builder: (context, store) {
+              return GlobalStoreWidget(
+                store: store,
+                child: PerAccountStoreWidget(
+                  accountId: accountId,
+                  child: MyWidgetWithMixin(key: widgetWithMixinKey)));
```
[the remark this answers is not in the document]
rajveermalviya: Used `TestZulipApp` in tests where possible, but kept this unchanged because I was getting the same extraneous dep changes mentioned in the above TODO.

rajveermalviya:
Thanks for the review @chrisbobbe! Pushed a new revision, PTAL.

lib/notifications/open.dart
```diff
@@ -88,6 +91,24 @@ class NotificationOpenManager {
...
+
+    NavigatorState navigator = await ZulipApp.navigator;
+    final context = navigator.context;
+    assert(context.mounted);
+    if (!context.mounted) return; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
+
+    final route = _routeForNotification(context, payload);
+    if (route == null) return; // TODO(log)
+
+    // TODO(nav): Better interact with existing nav stack on notif open
```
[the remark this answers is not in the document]
rajveermalviya: After this PR is merged the potential fix for that issue would work on both platforms, because this PR consolidates the notification routing implementation on both iOS and Android.

ios/Runner/AppDelegate.swift
```diff
@@ -33,3 +53,22 @@ private class NotificationHostApiImpl: NotificationHostApi {
...
+
+  override func onListen(withArguments arguments: Any?, sink: PigeonEventSink<NotificationPayloadForOpen>) {
+    eventSink = sink
+  }
+
+  func onNotificationTapEvent(data: NotificationPayloadForOpen) {
+    if let eventSink = eventSink {
+      eventSink.success(data)
+    }
+  }
```
chrisbobbe, line 68: Instead of the one class `NotificationPayloadForOpen`, could we have two separate classes `NotificationDataFromLaunch` and `NotificationTapEvent`? Perhaps with a `Payload` typedef helper:

```dart
typedef Payload = Map<Object?, Object?>;

class NotificationDataFromLaunch {
  const NotificationDataFromLaunch({required this.payload});

  /// The raw payload that is attached to the notification,
  /// holding the information required to carry out the navigation.
  final Payload payload;
}

class NotificationTapEvent {
  const NotificationTapEvent({required this.payload});

  /// The raw payload that is attached to the notification,
  /// holding the information required to carry out the navigation.
  final Payload payload;
}
```

I think the current naming makes the event-channel code harder to read than it needs to be. When reading the [Pigeon example code](https://github.com/flutter/packages/blob/main/packages/pigeon/example/README.md#HostApi_Example), I see "event" used pretty consistently in the names of things. Here, we have both "data" (`onNotificationTapEvent`'s param) and "payload", and "payload" is used ambiguously for a payload (`NotificationPayloadForOpen.payload`) and something that contains a payload (`NotificationPayloadForOpen` itself).

That would mean, for this method:

```swift
  func onNotificationTapEvent(event: NotificationTapEvent) {
    if let eventSink = eventSink {
      eventSink.success(event)
    }
  }
```

pigeon/notifications.dart
```diff
@@ -26,3 +26,16 @@ abstract class NotificationHostApi {
   /// See: https://developer.apple.com/documentation/uikit/uiapplication/launchoptionskey/remotenotification
   NotificationPayloadForOpen? getNotificationDataFromLaunch();
 }
+
+@EventChannelApi()
+abstract class NotificationHostEvents {
```
chrisbobbe, line 44: Would `NotificationEventChannelApi` be a better name for this, to make it clearer what kind of thing it is?

lib/model/binding.dart
```diff
@@ -310,6 +314,19 @@ class PackageInfo {
   });
 }
 
+// Pigeon generates methods under `@EventChannelApi` annotated classes
+// in global scope of the generated file. This is a helper class to
+// namespace the notification related Pigeon API under a single class.
+class NotificationPigeonApi {
+  final _notifInteractionHost = notif_pigeon.NotificationHostApi();
```
chrisbobbe, line 321: Is `_notifInteractionHost` the best name for this? How about `_hostApi`? We might add more to `NotificationHostApi` that's not about interacting with notifications (such as a method to query the current notification permissions). Also, `NotificationHostApi` isn't the only code that's about interacting with notifications; `notif_pigeon.notificationTapEvents` is too.

lib/notifications/open.dart
```diff
@@ -95,6 +98,24 @@ class NotificationOpenManager {
       // TODO(#82): Open at specific message, not just conversation
       narrow: openData.narrow);
   }
+
+  /// Navigates to the [MessageListPage] of the specific conversation
+  /// for the provided payload that was attached while creating the
+  /// notification.
+  Future<void> _navigateForNotification(NotificationPayloadForOpen payload) async {
```
chrisbobbe, line 101: Another reason the `NotificationPayloadForOpen` rename would be helpful: currently, it's pretty unclear that this private method is only about notifications that come while the app is open. It'll be helpful for debugging if it's easier to see what code is about the launch notification vs. not.

test/notifications/open_test.dart
```diff
@@ -115,6 +125,107 @@ void main() {
...
 
+    Future<void> checkOpenNotification(WidgetTester tester, Account account, Message message) async {
+      await openNotification(tester, account, message);
+      matchesNavigation(check(pushedRoutes).single, account, message);
+      pushedRoutes.clear();
+    }
+
+    testWidgets('(iOS) stream message', (tester) async {
+      addTearDown(testBinding.reset);
+      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
```
chrisbobbe, line 163: This add-self-account looks boring; can we put it in `prepare`?

chrisbobbe (commented):
Thanks! Here's a more thorough review of the first four commits:

d1f648cba app: Use DeferredBuilderWidget while loading GlobalStore
a5a0fff85 pigeon [nfc]: Rename pigeon file to `notification` -> `android_notifications`
ab3ff8434 notif ios: Navigate when app launched from notification
4b2ade09d notif ios: Navigate when app running but in background

That leaves the last two commits:

28ea77fcf notif android: Migrate to cross-platform Pigeon API for navigation
616defef0 docs: Document testing push notifications on iOS Simulator

Actually for your next revision, could you send a new PR with everything except the "Migrate to cross-platform" commit? That one's pretty large, so makes sense to review separately.

test/model/binding.dart
```diff
@@ -724,6 +733,20 @@ class FakeAndroidNotificationHostApi implements AndroidNotificationHostApi {
...
+
+  /// Populates the notification data for launch to be returned
+  /// by [getNotificationDataFromLaunch].
+  void setNotificationDataFromLaunch(NotificationPayloadForOpen? data) {
+    _notificationDataFromLaunch = data;
+  }
+
+  @override
+  Future<NotificationPayloadForOpen?> getNotificationDataFromLaunch() async =>
+    _notificationDataFromLaunch;
```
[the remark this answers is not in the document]
rajveermalviya: Right, on the platform side too it's just a getter. And we already handle the awaiting this future using `DeferredBuilderWidget`.

test/notifications/open_test.dart
```diff
@@ -0,0 +1,160 @@
...
+      "user_id": account.userId,
+      "sender_id": zulipMessage.senderId,
+      "sender_email": zulipMessage.senderEmail,
+      "time": zulipMessage.timestamp,
+      "message_ids": [zulipMessage.id],
+      ...(switch (zulipMessage) {
+        StreamMessage(:var streamId, :var topic) => {
+          "recipient_type": "stream",
+          "stream_id": streamId,
+          if (streamName != null) "stream": streamName,
```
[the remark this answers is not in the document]
rajveermalviya: Hmm, that's not what I see in the payloads. Also see [the similar code](https://github.com/zulip/zulip-mobile/blob/4d9d46d7f852af3fb4d44463654a141141e7bac3/src/notification/extract.js#L115-L116) for handling this in zulip-mobile.

test/notifications/open_test.dart
```diff
@@ -0,0 +1,160 @@
...
+
+    Future<void> prepare(WidgetTester tester,
+        {bool early = false, bool withAccount = true}) async {
+      await init();
+      pushedRoutes = [];
+      final testNavObserver = TestNavigatorObserver()
+        ..onPushed = (route, prevRoute) => pushedRoutes.add(route);
+      // This uses [ZulipApp] instead of [TestZulipApp] because notification
+      // logic uses `await ZulipApp.navigator`.
+      await tester.pumpWidget(ZulipApp(navigatorObservers: [testNavObserver]));
```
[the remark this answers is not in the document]
rajveermalviya: Right this commit doesn't use it, but the following commit does. So, removed the comment from the first commit but still using `ZulipApp` for simplicity.

test/notifications/open_test.dart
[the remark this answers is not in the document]
chrisbobbe: Indeed, thanks for reading more carefully than me :) awkward that that code reads the value from a "stream" key and stores it in a `stream_name` variable.

rajveermalviya:

test/notifications/open_test.dart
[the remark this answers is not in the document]
gnprice: Yeah, the `stream_name` name is partly there because of the work to redo the legacy app's internal APIs so as to track streams (now channels) by their ID, rather than by their name. There had been a lot of variables, parameters, etc., that were named just "stream" and meant a stream name, and I wanted to make those stand out because a name shouldn't be the canonical way to refer to a stream.

Now that we're pretty good about that, I'd be less determined to 100% avoid "stream" for a stream name; but the longer name is probably still better.

chrisbobbe (commented):
Thanks! Just small comments below.

chrisbobbe:

test/widgets/store_test.dart
```diff
@@ -58,43 +58,17 @@ extension MyWidgetWithMixinStateChecks on Subject<MyWidgetWithMixinState> {
 void main() {
   TestZulipBinding.ensureInitialized();
 
-  testWidgets('GlobalStoreWidget loads data while showing placeholder', (tester) async {
```
gnprice, line 61: This functionality should continue to get tested, though the test will now be about ZulipApp (and live in the corresponding test file) instead.

lib/widgets/app.dart
```diff
@@ -212,44 +242,110 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
+          }));
+      });
+  }
+}
+
+/// A widget that defers the builder until the provided [future] completes.
+///
+/// It shows a placeholder widget while it waits for the [future]
+/// to complete.
+class DeferredBuilderWidget<T> extends StatefulWidget {
```
gnprice, line 304: Seeing this widget class with this quite generic name and description makes me think of the widget Flutter provides for this upstream, FutureBuilder, and wonder about what the differences are and why that's not used.

The FutureBuilder API is somewhat more complicated, partly in order to handle errors, whereas we ignore errors here. I think that simplification would be straightforward to package up as a DeferredBuilderWidget that is implemented via FutureBuilder, though.

Then looking at FutureBuilder causes me to notice one subtlety: if the `DeferredBuilderWidget`'s parent gets rebuilt and provides a new future, the `DeferredBuilderWidget` never actually looks at the new future. That seems buggy. It doesn't affect our use case because the new future is always substantively the same; but on a nice generic interface like this, it's a bug waiting to happen with a future use of this class.

We could fix that with more effort, but I think the cleanest thing may be to just make this logic specialized again, without the generality. As is, all the call sites of the DeferredBuilderWidget constructor are wrapped around a GlobalStoreWidget, and vice versa.

I'll describe a way to do that in the next comment.

lib/widgets/app.dart
```diff
@@ -151,10 +153,13 @@ class ZulipApp extends StatefulWidget {
 }
 
 class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
+  late final Future<GlobalStore> _globalStoreFuture;
+
   @override
   void initState() {
     super.initState();
     WidgetsBinding.instance.addObserver(this);
+    _globalStoreFuture = ZulipBinding.instance.getGlobalStoreUniquely();
```
gnprice, line 162: This can do the job of both DeferredBuilderWidget and the Future.wait, in just a few lines of code because specialized to our use case:
```suggestion
    () async {
      final globalStore = await ZulipBinding.instance.getGlobalStoreUniquely();
      final notifFuture = NotificationOpenManager.instance.initializationFuture;
      if (notifFuture != null) await notifFuture;
      if (!mounted) return;
      setState(() {
        this.globalStore = globalStore;
      });
    }();
```

Relative to the status quo in main (or to a version after a refactor like the one in your first commit, moving the loading up to ZulipApp), this differs only in adding the two lines for `notifFuture`.

Because the underlying asynchronous work behind `initializationFuture` is already started by `main` before reaching this point, the only thing this needs to do to make the loading happen concurrently is to be sure to call `getGlobalStoreUniquely()` before awaiting `initializationFuture`.

test/widgets/store_test.dart
```diff
@@ -149,13 +120,18 @@ void main() {
...
-          child: Builder(
-            builder: (context) {
-              final store = PerAccountStoreWidget.of(context);
-              return Text('found store, account: ${store.accountId}');
-            }))));
+        child: DeferredBuilderWidget(
+          future: testBinding.getGlobalStoreUniquely(),
+          builder: (context, store) {
+            return GlobalStoreWidget(
+              store: store,
```
gnprice, line 127: For these tests that aren't about GlobalStoreWidget and how it would formerly load data, I think they don't need DeferredBuilderWidget or an analogue; they can just use (the new) GlobalStoreWidget directly, and have the caller do its own await to get the actual global store to pass in.

lib/widgets/store.dart
```diff
@@ -75,43 +72,6 @@ class GlobalStoreWidget extends StatefulWidget {
...
-    final store = this.store;
-    if (store == null) return widget.placeholder;
-    return _GlobalStoreInheritedWidget(store: store, child: widget.child);
-  }
-}
-
-// This is separate from [GlobalStoreWidget] only because we need
-// a [StatefulWidget] to get hold of the store, and an [InheritedWidget] to
-// provide it to descendants, and one widget can't be both of those.
-class _GlobalStoreInheritedWidget extends InheritedNotifier<GlobalStore> {
```
gnprice, line 107: There's a comment below which refers to this class, so needs to be updated :slightly_smiling_face:

gnprice (commented):

Here's a review just on the first part, as a start:
5eb2d1bd2 app: Use DeferredBuilderWidget while loading GlobalStore
06394114b pigeon [nfc]: Rename `notifications.dart` to `android_notifications.dart`
5a462cb56 dialog [nfc]: Document required ancestors for BuildContext in showErrorDialog
be9018362 notif: Fix error message when account not found in store
e9ce0d877 binding test [nfc]: Reorder androidNotificationHost getter

(i.e. mainly on the first commit). I'll look closer at the subsequent, main commits in a next round.

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,233 @@
...
+
+```shell-session
+$ echo '{"alert":{"title": ...' | jq '{aps: {alert: .alert, sound: .sound, badge: .badge}, zulip: .custom.zulip}'
+```
+
+## 6. Push APNs payload to iOS Simulator
+
+_If you skipped steps 2-5, you'll need pre-forged APNs payloads for
+existing messages in a default development server messages for the
+user `iago@zulip.com`:_
```
gnprice, line 88: These worked great for me (because I haven't gotten the dev server working with the bouncer — see question above), with one change:

I replaced the string `localhost` throughout with the IP address my dev server was using in its EXTERNAL_HOST, per the instructions at https://github.com/zulip/zulip-mobile/blob/main/docs/howto/dev-server.md .

(If my dev server were located on the Mac where the simulator is running, then `localhost` would work fine, as per https://github.com/zulip/zulip-mobile/blob/main/docs/howto/dev-server.md#ios-simulator-macos-only . But instead I have it running on a different machine, my Linux desktop, so it needs to use the address of that machine.)

For me that looked like:
`$ perl -i -0pe s/localhost/192.168.0.113/g tmp/*.json`
to make the strings say `"http://192.168.0.113:9991"` instead of `"http://localhost:9991"`.

gnprice (commented):
Here's a review on the handy new page of docs.

Using the canned payloads, I successfully tested this PR, and opening a notification works!

I didn't get to testing it more end-to-end, with a real (dev) server; see questions below.

gnprice:
Oh also let's put the doc commit before the two main "notif ios:" commits. It's basically describing how to manually test those commits and the functionality they're changing; and the instructions apply just as well before those changes as they do after them.

gnprice (commented):
I started reviewing the new revision, but then realized you hadn't actually said it was ready for review yet. :slightly_smiling_face: Anyway, here's the comments I wrote down so far, all on the first commit:
e6e19e1e5 app: Move initialization of GlobalStore from GlobalStoreWidget to ZulipApp

But I'll leave further re-review until you say it's ready.

ios/Runner/AppDelegate.swift
```diff
@@ -3,11 +3,72 @@ import Flutter
...
+  var eventSink: PigeonEventSink<NotificationTapEvent>?
+
+  override func onListen(withArguments arguments: Any?, sink: PigeonEventSink<NotificationTapEvent>) {
+    eventSink = sink
+  }
+
+  func onNotificationTapEvent(event: NotificationTapEvent) {
+    if let eventSink = eventSink {
+      eventSink.success(event)
+    }
```
gnprice, line 67: This would do the same thing, right?
```suggestion
    eventSink?.success(event)
```

ios/Runner/AppDelegate.swift
```diff
@@ -3,11 +3,72 @@ import Flutter
...
   }
+
+  override func userNotificationCenter(
+    _ center: UNUserNotificationCenter,
+    didReceive response: UNNotificationResponse,
+    withCompletionHandler completionHandler: @escaping () -> Void
+  ) {
+    if let listener = notificationTapEventListener {
+      let userInfo = response.notification.request.content.userInfo
+      listener.onNotificationTapEvent(event: NotificationTapEvent(payload: userInfo))
```
gnprice, line 39: I think this API would be a bit cleaner if we adjust it like so:
```suggestion
      listener.onNotificationTapEvent(payload: userInfo)
```

Then the NotificationTapEvent gets constructed by the method implementation.

That's also consistent with the Pigeon docs' [example](https://github.com/flutter/packages/blob/e6e814c1ee1894efff93a17293c0d5e3169db05d/packages/pigeon/example/README.md#swift-2):
```swift
  func onIntEvent(event: Int64) {
    if let eventSink = eventSink {
      eventSink.success(IntEvent(data: event))
    }
  }

  func onStringEvent(event: String) {
    if let eventSink = eventSink {
      eventSink.success(StringEvent(data: event))
    }
  }
```

ios/Runner/AppDelegate.swift
```diff
@@ -3,11 +3,72 @@ import Flutter
...
+  private var notificationTapEventListener: NotificationTapEventListener?
+
   override func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
     GeneratedPluginRegistrant.register(with: self)
+    guard let controller = window?.rootViewController as? FlutterViewController else {
+      fatalError("rootViewController is not type FlutterViewController")
+    }
```
gnprice, line 15: What did you look at to determine this was the way to set up the Pigeon channels?

(I'm not finding a clear answer as I look now in the Pigeon docs.)

gnprice (commented):
OK, and I've now sat down and spent some time studying the Swift code (and reading docs to understand how it should work). Here's a round of review on that.

ios/Runner/AppDelegate.swift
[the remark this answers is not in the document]
rajveermalviya: Hmm, I tested this and found that dismissing a notification doesn't cause that method to be called. (Even with the [small func implemented](https://github.com/zulip/zulip-mobile/blob/4d9d46d7f852af3fb4d44463654a141141e7bac3/ios/ZulipMobile/AppDelegate.mm#L130-L145) to support foreground notifications on iOS)

The the docs for [UNNotificationDismissActionIdentifier](https://developer.apple.com/documentation/usernotifications/unnotificationdismissactionidentifier) say:
> The system delivers this action only if your app configured the notification’s category object with the [customDismissAction](https://developer.apple.com/documentation/usernotifications/unnotificationcategoryoptions/customdismissaction) option.
> …

And from [UNNotificationDefaultActionIdentifier](https://developer.apple.com/documentation/usernotifications/unnotificationdefaultactionidentifier):
> The delivery of this action doesn’t require any special configuration of notification categories. Use the [userNotificationCenter(_:didReceive:withCompletionHandler:)](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:didreceive:withcompletionhandler:)) method of your delegate object to receive this action.

So, since we do not setup a notification category using [setNotificationCategories](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/setnotificationcategories(_:)), we don't need to handle `UNNotificationDismissActionIdentifier` currently.

But adding a check anyway (`response.actionIdentifier == UNNotificationDefaultActionIdentifier`) in case in future that changes.

ios/Runner/AppDelegate.swift
rajveermalviya: I had adapted this implementation by refering to [Flutter's PlatformChannel examples](https://github.com/flutter/flutter/blob/2615ab6c1932e44e6802aaba8dc715b387ff155f/examples/platform_channel_swift/ios/Runner/AppDelegate.swift#L30-L32) and [Pigeon's example](https://github.com/flutter/packages/blob/267ac7b66308ad34b6ce14c1f5399ab0691f9ede/packages/pigeon/example/app/ios/Runner/AppDelegate.swift#L104).

ios/Runner/AppDelegate.swift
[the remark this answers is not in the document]
gnprice: Ah I see, good to know. Sounds good, though.

ios/Runner/AppDelegate.swift
gnprice: Excellent, thanks. This logic looks good, then.

For the style of it, let's use the terser `as!` form, rather than throwing with a custom error message. It should be an invariant of our app that there'll be a FlutterViewController here, and I think that's a boring enough fact that we don't need to call extra attention to it.

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,233 @@
...
+
+This doc describes how to test client side changes on iOS Simulator.
+It will demonstrate how to use APNs payloads the server sends to
+Apple's Push Notification service to show notifications on iOS
+Simulator.
+
+## 1. Setup dev server
+
+[Follow the steps in this section to setup a development server, and
+register the development server to the production bouncer](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/push-notifications.md#server).
```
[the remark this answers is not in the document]
rajveermalviya: It looks like after the commit https://github.com/zulip/zulip/commit/4a9314943525ad001c46a8ab44a0ca42e26bbb80 there is no need to do the push notifications registration manually, just running the dev server should work fine. Tested with fresh dev environment and verified that it works.

rajveermalviya:

gnprice (commented):
Thanks for the revision!

Here's comments on just the first commit:
60577d6f1 app: Move initialization of GlobalStore from GlobalStoreWidget to ZulipApp

It's felt unsatisfying how that change concentrates more functionality on the omnibus widget ZulipApp, particularly as that causes it to get duplicated on TestZulipApp (which introduces the risk of divergence between test and live code). After the comment below about comments that refer to the store getting loaded, I noticed that it seemed like some of those comments were going to need to get more complicated, too, to explain the new situation.

So I went and tried implementing the solution I think I suggested when we looked at this task in a call a few weeks ago: have GlobalStoreWidget keep responsibility for loading the store, and add a parameter to have it await an additional future. I think that works out well. Just pushed back to the PR branch a revision with that change:
9426b8298 store: Add "blocking future" option on GlobalStoreWidget

The effect on the subsequent "Navigate when app launched from notification" commit is small:
```
    -@@ lib/widgets/app.dart: class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
    -     WidgetsBinding.instance.addObserver(this);
    -     () async {
    -       final globalStore = await ZulipBinding.instance.getGlobalStoreUniquely();
    -+      final notifFuture = NotificationOpenManager.instance.initializationFuture;
    -+      if (notifFuture != null) await notifFuture;
    -       if (!mounted) return;
    -       setState(() {
    -         _globalStore = globalStore;
    +@@ lib/widgets/app.dart: class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
    +   @override
    +   Widget build(BuildContext context) {
    +     return GlobalStoreWidget(
    ++      blockingFuture: NotificationOpenManager.instance.initializationFuture,
    +       child: Builder(builder: (context) {

File: docs/howto/push-notifications-ios-simulator.md [hunk 21]
```diff
@@ -0,0 +1,235 @@
...
+```
+
+## 6. Push APNs payload to iOS Simulator
+
+_If you skipped steps 2-5, you'll need pre-forged APNs payloads for
+existing messages in a default development server messages for the
+user `iago@zulip.com`:_
+
+The canned payloads were generated from Zulip Server 11.0-dev+git
+8fd04b0f, API Feature Level 377, in April 2025.
```
gnprice, line 90: That's a version from this week, after you originally wrote this doc — did you confirm that it still produces the same output?

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,235 @@
...
+Apple's Push Notification service to show notifications on iOS
+Simulator.
+
+## 1.(Optional) Setup dev server
+
+_Skip to step 6 in which you will use canned notification payloads,
+these intermediate steps records how to get those payloads, which may
+be useful in future._
+
+Setup and run Zulip dev server following the [setup tutorial](https://zulip.readthedocs.io/en/latest/development/setup-recommended.html).
```
gnprice, line 17: I think this isn't enough on its own — for getting one's phone to be able to talk to the dev server you'll want the instructions at https://github.com/zulip/zulip-mobile/blob/main/docs/howto/dev-server.md (per https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2023621262).

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,235 @@
...
+
+For documentation on testing push notifications on Android or a real
+iOS Device, see https://github.com/zulip/zulip-mobile/blob/main/docs/howto/push-notifications.md
+
+This doc describes how to test client side changes on iOS Simulator.
+It will demonstrate how to use APNs payloads the server sends to
+Apple's Push Notification service to show notifications on iOS
+Simulator.
+
+## 1.(Optional) Setup dev server
```
gnprice, line 11: nit:
```suggestion
## 1. (Optional) Setup dev server
```

docs/howto/push-notifications-ios-simulator.md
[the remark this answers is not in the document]
gnprice: Huh interesting. That is a bit puzzling, though — looking at the `dev_settings.py` in that commit, it seems like it'll end up trying to talk to the bouncer at `http://push.zulipdev.com:9991` (exactly like the signup command was doing when I tried above), or at an even more unlikely URL like `http://push.192.168.0.101:9991` when one sets EXTERNAL_HOST to enable one's phone to get to the server.

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,235 @@
...
+
+We'll use the devlogin user `iago@zulip.com` to test notifications,
+log in to that user by going to `/devlogin` on that server on Web.
+
+And then follow the steps [here](https://zulip.com/help/mobile-notifications)
+to enable Mobile Notifications for "Channels".
+
+## 3. (Optional) Login to the dev user on zulip-flutter.
+
+<!-- TODO(405) Guide to use the new devlogin page instead -->
```
gnprice, line 29:
```suggestion
<!-- TODO(#405) Guide to use the new devlogin page instead -->
```

That's the convention we use for TODO links to issues. See existing examples:
```
$ git grep -P 'TODO.\d+'
$ git grep -P 'TODO.#\d+'
```

Keeping the convention consistent helps keep the references greppable.

docs/howto/push-notifications-ios-simulator.md
gnprice: bump — I'd like this information to not get lost for the next person.

Here's a version that could go in verbatim:
> These canned payloads assume that EXTERNAL_HOST has its default value for the dev server. If you've [set EXTERNAL_HOST to use an IP address](https://github.com/zulip/zulip-mobile/blob/main/docs/howto/dev-server.md#4-set-external_host) in order to enable your device to connect to the dev server, you'll need to adjust the `realm_url` fields. You can do this by a find-and-replace for `localhost`; for example, `perl -i -0pe s/localhost/10.0.2.2/g tmp/*.json` after saving the canned payloads to files `tmp/*.json`.

gnprice (commented):
And here's a re-review on the docs commit.

docs/howto/push-notifications-ios-simulator.md
rajveermalviya: I had changed to that link because these instructions were specific to testing on iOS Simulator, and expectation was to do these intermediate (optional) steps on the Simulator too.

Changed back to this link, in case reader tries these steps on an actual device, or in your case where server is running on a different machine than the Mac.

docs/howto/push-notifications-ios-simulator.md
gnprice: Great, thanks!

docs/howto/push-notifications-ios-simulator.md
gnprice: Ah true, the doc does say it's about the simulator. And the key `xcrun simctl push` step at the end sure sounds like it'll only work on the simulator.

It's probably not real common to be running the dev server on a different machine from the Mac hosting the simulator. So it'd be reasonable to add a remark here saying that if you'll be running the dev server on the Mac that's also hosting the simulator, then those steps boil down to just the normal Zulip dev server instructions (and include a direct link to the latter).

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+## 6. Push APNs payload to iOS Simulator
+
+_If you skipped steps 2-5, you'll need pre-forged APNs payloads for
+existing messages in a default development server messages for the
+user `iago@zulip.com`:_
+
+These canned payloads were generated from Zulip Server 11.0-dev+git
+8fd04b0f0, API Feature Level 377, in April 2025.
```
gnprice, line 90: ah, thanks for adding the 9th digit here :-)

gnprice (commented):
OK, completed re-review of the things covered in the last pair of reviews. All looks good; one comment here above:
https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2043473092

Still on my plate in reviewing this are:
* re-review the Swift code, revised since https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2741346660
* re-review the small early commits, revised since https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2730602297 (I believe little changed here and I didn't have much to say the first time, so should be quick)
* retry the testing instructions on a fresh dev server, following https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2034079271
* review the Dart changes in those two main commits (having covered the Swift changes earlier)

docs/howto/push-notifications-ios-simulator.md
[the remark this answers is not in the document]
gnprice: Started a chat thread to debug in more detail: [#mobile-team > notifications from dev server @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/notifications.20from.20dev.20server/near/2159855)

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,253 @@
...
+         trigger,
+         mentioned_user_group_id,
+         mentioned_user_group_name,
+         can_access_sender,
+     )
+     gcm_payload, gcm_options = get_message_payload_gcm(
+         user_profile, message, mentioned_user_group_id, mentioned_user_group_name, can_access_sender
+     )
+     logger.info("Sending push notifications to mobile clients for user %s", user_profile_id)
++    logger.info("APNS payload %s", orjson.dumps(apns_payload))
```
gnprice, line 70:
```suggestion
+    logger.info("APNS payload %s", orjson.dumps(apns_payload).decode())
```

That way it gets printed in the log as verbatim JSON. Otherwise the byte-string gets encoded the way it would be in Python source code, like `b'…'` — which means that if you try to copy-paste it to use directly, it can get mangled depending on the contents. For example if you send a message whose content is `"`, you get:
`APNS payload: b'{"alert":{"title":"Desdemona","subtitle":"","body":"\\""},"sound":"default","badge":0,"custom":{"zulip":{"server":"192.168.0.113:9991","realm_id":2,"realm_uri":"http://192.168.0.113:9991","realm_url":"http://192.168.0.113:9991","realm_name":"Zulip Dev","user_id":11,"sender_id":9,"sender_email":"user9@zulipdev.com","time":1746576414,"recipient_type":"private","message_ids":[104]}}}'`

with `"\\""` , which fails to parse as JSON.

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,253 @@
...
+        "message_ids": [
+            40
+        ]
+    }
+}
+```
+
+</details>
+
+To receive a notification on the iOS Simulator, we need to push
```
gnprice, line 216: Now that I've gotten the full instructions working end to end for me and then scanned back through them, I think the structure of what the reader is intended to do here can be expressed more clearly with a bit of reordering:
* First, a `##` section consisting of the steps from here to the end.

  This is the main thing the document is saying to do — everything else is optional, and/or only needs to be done occasionally before then doing this portion many times.
* Then, a `##` section with the three pre-canned payloads, i.e. the portion of Step 6 above this point.
* Then, a third `##` section describing how to produce new canned payloads. That consists of Steps 1–5, with each one adjusted to a `###`-level section instead of `##`-level so that it's a sub-section of this overall section.

Then the structure is: you do what's in the first section. To get input to use in that, you can use either the second section or the third section.

And then in particular the individual steps of the third section don't get labeled as optional; they're all required if you want to produce payloads. So this structure makes that relationship clearer, whereas the individual "optional" labels look as if they're saying one might pick and choose.

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,253 @@
...
+
+For documentation on testing push notifications on Android or a real
+iOS Device, see https://github.com/zulip/zulip-mobile/blob/main/docs/howto/push-notifications.md
+
+This doc describes how to test client side changes on iOS Simulator.
+It will demonstrate how to use APNs payloads the server sends to
+Apple's Push Notification service to show notifications on iOS
+Simulator.
+
+## 1. (Optional) Setup dev server
```
gnprice, line 11:
```suggestion
## 1. (Optional) Set up dev server
```

"setup" is a noun, "set up" the verb

docs/howto/push-notifications-ios-simulator.md
```diff
@@ -0,0 +1,253 @@
...
+
+## 2. (Optional) Setup the dev user to receive mobile notifications.
+
+We'll use the devlogin user `iago@zulip.com` to test notifications,
+log in to that user by going to `/devlogin` on that server on Web.
+
+And then follow the steps [here](https://zulip.com/help/mobile-notifications)
+to enable Mobile Notifications for "Channels".
+
+## 3. (Optional) Login to the dev user on zulip-flutter.
```
gnprice, line 36:
```suggestion
## 3. (Optional) Log in as the dev user on zulip-flutter.
```

Similarly "login" is a noun, "log in" the verb.

And as a smaller point: one logs in "to" a server, a realm, or even an account, but logs in "as" a user. The user isn't a space one is entering, but an identity one is taking on.

gnprice (commented):
OK, I finally came back and tested out these instructions after the debugging described in that chat thread (from https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2055151244 ).

This version works great, with one correction below.

Then I also have one high-level comment on the structure (doesn't call for much new text, mostly just moving things around plus a bit of new section-intro text), and a couple of small comments.

My remaining review to-do list for this PR is the other three points in https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2766578116 .

ios/Runner/AppDelegate.swift
```diff
@@ -3,11 +3,64 @@ import Flutter
...
+  init(_ maybeDataFromLaunch: NotificationDataFromLaunch?) {
+    self.maybeDataFromLaunch = maybeDataFromLaunch
+  }
+
+  func getNotificationDataFromLaunch() -> NotificationDataFromLaunch? {
+    maybeDataFromLaunch
+  }
+}
+
+class NotificationTapEventListener: NotificationTapEventsStreamHandler {
```
gnprice, line 56: Let's include comments somewhere in here with links to the Flutter examples that you found helpful, from the thread https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2027866613 , since those examples were nontrivial to find.

Here at the top of this class is probably a fine place.

gnprice (commented):
Here's re-review on the Swift changes. The revisions look good; small comments below.

lib/widgets/dialog.dart
```diff
@@ -49,10 +49,13 @@ class DialogStatus<T> {
 ///
 /// The [DialogStatus.result] field of the return value can be used
 /// for waiting for the dialog to be closed.
+///
+/// The context argument is used to look up [ZulipLocalizations],
+/// the [Navigator] and [Theme] for the dialog.
```
gnprice, line 54: I'm not sure how to use this information, when writing or editing a call site of this function. What sort of context does this mean is acceptable or not acceptable?

I believe the main upshot is: this context should be a descendant of the app's main [Navigator]. (That's enough, because the [ZulipLocalizations] and [Theme] are provided by [MaterialApp] and go above the [Navigator].)

Does that cover it? Or are there situations where further guidance beyond that would be helpful?

gnprice: Also the same information should probably go on `showSuggestedActionDialog` too, since that's so similar.

gnprice (commented):
OK, and re-reviewed the small commits from early in the branch. Just a couple of small comments below.

With this and the reviews just above, the remaining items for me to review in this PR are, after https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2766578116, down to just:
* review the Dart changes in those two main commits at the end of the branch (having covered the Swift changes above).

lib/notifications/open.dart
```diff
@@ -0,0 +1,229 @@
...
+          'topic': final String topic,
+        } =>
+          TopicNarrow(streamId, TopicName(topic)),
+
+        {'recipient_type': 'private', 'pm_users': final String pmUsers} =>
+          DmNarrow(
+            allRecipientIds: pmUsers
+              .split(',')
+              .map((e) => int.parse(e, radix: 10))
+              .toList(growable: false)..sort(),
```
gnprice, line 211: nit: the sort is a significant step, so best to keep it visible and not tuck it in:
```suggestion
              .toList(growable: false)
              ..sort(),
```

lib/notifications/open.dart
```diff
@@ -0,0 +1,229 @@
...
+    final notifNavData = _tryParsePayload(context, data.payload);
+    if (notifNavData == null) return null; // TODO(log)
+    return _routeForNotification(context, notifNavData);
+  }
+
+  /// Provides the route to open by parsing the notification payload.
+  ///
+  /// Returns null and shows an error dialog if the associated account is not
+  /// found in the global store.
+  AccountRoute<void>? _routeForNotification(BuildContext context, NotificationNavigationData data) {
```
gnprice, line 94: It looks like the logic in this method all duplicates most of what's in `NotificationDisplayManager.routeForNotification`.

It'd be good to avoid duplicating that. Instead, can have a method with this logic (could live on either class), and have `NotificationDisplayManager.routeForNotification` call that.

This does take its input in a different form, as `NotificationNavigationData` instead of `NotificationOpenPayload`. The simplest way to share the logic, starting from this version, would be to have the method that has this logic just take the three fields as separate arguments — since those two classes have the exact same three fields.

It also feels like these two classes are really two ways of describing the same thing, though. How about making `fromIosApnsPayload` be a new constructor on the `NotificationOpenPayload` class? The doc comment on that class would change a bit:
```dart
/// The information contained in 'zulip://notification/…' internal
/// Android intent data URL, used for notification-open flow.
class NotificationOpenPayload {
```
but I think the real point remains the same: it's the data about a notification that describes what to do on opening it.

gnprice (commented):
OK, and started on those two main commits at the end.

I think in this round I've read the first of them:
37683dbda notif ios: Navigate when app launched from notification

except its tests.

rajveermalviya:

gnprice:
Thanks for the revision!

High-level comment on the main code changes in this revision: this rename commit looks good:
082f95e42 notif [nfc]: Rename NotificationOpenPayload -> NotificationNavigationData

and the structure of the refactors (along the lines of https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2078691844) within the next commit looks good too:
add4f1732 notif ios: Navigate when app launched from notification

But that latter commit is harder to read than I'd like it to be, because those refactors (moving a lot of code around) are combined in the same commit with substantial other changes (introducing new logic for iOS). Let's split that into two:
* One commit that moves the NotificationNavigationData class and moves/refactors the routeForNotification method, as a pure NFC change. This will create the new file and new NotificationNavigationService class, even though those won't yet do much else. It'll also make the corresponding changes on ZulipApp to follow those moves.
* Then one commit that does the rest of the work of the existing commit, including adding the rest of NotificationNavigationService's functionality like `init` and `routeForNotificationFromLaunch`.

This is an example of [clear and coherent commits](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html#commits-should-generally-be-minimal). I think each of those new commits will be easier to read once split up than the same changes are as part of the combined commit.

lib/widgets/dialog.dart
```diff
@@ -49,10 +49,12 @@ class DialogStatus<T> {
 ///
 /// The [DialogStatus.result] field of the return value can be used
 /// for waiting for the dialog to be closed.
+///
+/// The context argument is used to look up the [Navigator] for the dialog.
```
gnprice, line 53: This is a bit misleading: the context might indeed be used to look that up, but it's also used to look up other things. The reader can see one of those things a few lines below, which casts doubt on the trustworthiness of the whole function's doc.

My suggestion above at https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2076610382 was a bit different; here it is concretely:
```suggestion
/// The context argument should be a descendant of the app's main [Navigator].
```

By being less specific about concrete details (that the caller of this function doesn't need to care about), this version is actually more complete: it covers all the ways the context ends up getting used, because all the ancestors it needs are either the main Navigator or one of its ancestors.

lib/widgets/app.dart
```diff
@@ -168,27 +169,39 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
...
-        ];
-      } else {
-        // The account didn't match any existing accounts,
-        // fall through to show the default route below.
-      }
+    final route = defaultTargetPlatform == TargetPlatform.iOS
+        ? _initialRouteIos(context)
+        : _initialRouteAndroid(context, initialRoute);
+    if (route != null) {
+      return [HomePage.buildRoute(accountId: route.accountId), route];
```
gnprice, line 204: nit: I think keeping this list vertically expanded (like it is in main) is helpful for keeping both of the routes visually salient. Similar to how the children of `Row` and `Column` widgets almost always get listed on separate lines.

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,317 @@
...
+      assert(debugLog('$e\n$st'));
+      final zulipLocalizations = ZulipLocalizations.of(context);
+      showErrorDialog(context: context,
+        title: zulipLocalizations.errorNotificationOpenTitle);
+      return null;
+    }
+  }
+}
+
+class NotificationNavigationData {
```
gnprice, line 170: Let's give this class a bit of dartdoc to say what it's intended to mean, like the old `NotificationOpenPayload` had. Like this (cf https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2078691844):
```suggestion
/// The data from a notification that describes what to do
/// when the user opens the notification.
class NotificationNavigationData {
```

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,317 @@
...
+import '../widgets/app.dart';
+import '../widgets/dialog.dart';
+import '../widgets/message_list.dart';
+import '../widgets/page.dart';
+import '../widgets/store.dart';
+
+NotificationPigeonApi get _notifPigeonApi => ZulipBinding.instance.notificationPigeonApi;
+
+/// Service for handling notification navigation.
+class NotificationNavigationService {
```
gnprice, line 23: What's your thinking behind naming this class and this file with "navigate", rather than "open" as in the previous revision?

(Either name seems plausible to me; but I'm curious to hear what shifted your thinking to cause you to rename it.)

gnprice (commented):
OK, and here's a re-review on the changes related to my last three review groups above: the Swift changes https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2820053475; the small commits from early in the branch https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2820079416; and the Dart changes in the first of the two main commits https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2823467569 . Just a handful of small comments from those, below.

Next I'll look at the docs commit, following up on https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2820030986 , and that will complete my re-review.

After you split up that first main commit as described in
https://github.com/zulip/zulip-flutter/pull/1379#issuecomment-2887861487 , I'll be able to read that more completely, followed by the final commit.

gnprice:
OK, and the docs changes generally all look good. I think my remaining comments on the doc are all best expressed by just making the edits — so I've pushed an additional commit to the branch:

commit 5df6033301ecabf57e58b8df7559fb2d49cdf757
Author: Greg Price <greg@zulip.com>
Date:   Sat May 17 21:20:02 2025 -0700

    docs: Clarify and expand a few spots in the iOS simulator notif doc
    
    Also make use of a handy shorthand within `jq`.


Let's leave that unsquashed, but reorder it to just after the commit that introduces the doc.

Then the remaining steps are as described in my review just above: https://github.com/zulip/zulip-flutter/pull/1379#pullrequestreview-2848669553

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,320 @@
...
+    _instance = null;
+  }
+
+  NotificationDataFromLaunch? _notifDataFromLaunch;
+
+  /// A [Future] that completes to signal that the initialization of
+  /// [NotificationNavigationService] has completed or errored.
+  ///
+  /// Returns null if [start] wasn't called yet.
+  Future<void>? get initializationFuture => _initializedSignal?.future;
```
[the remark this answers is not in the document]
gnprice: For the name, how about just `initialized`? It's common for things of type `Future` to have names that describe the future's result, letting the type speak for the fact that it's a future.

lib/notifications/navigate.dart
gnprice: (Re `NotificationNavigationData`, we'll probably want that to end up covering the more general "open" concept too: in particular we'll need the message ID from the notification, and that will naturally become another field on that class. That more general concept is reflected in the name of the existing `NotificationOpenPayload`, and in the dartdoc this version has which I supplied at https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2094343971 above.)

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,320 @@
...
+    final notifNavData = _tryParseIosApnsPayload(context, event.payload);
+    if (notifNavData == null) return; // TODO(log)
+    final route = routeForNotification(context, notifNavData);
+    if (route == null) return; // TODO(log)
+
+    // TODO(nav): Better interact with existing nav stack on notif open
+    unawaited(navigator.push(route));
+  }
+
+  NotificationNavigationData? _tryParseIosApnsPayload(
```
gnprice, line 140: Can this be static? Seems similar to `tryParseAndroidNotificationUrl` just below it.

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,320 @@
...
+  /// notification.
+  Future<void> _navigateForNotification(NotificationTapEvent event) async {
+    assert(debugLog('opened notif: ${jsonEncode(event.payload)}'));
+
+    NavigatorState navigator = await ZulipApp.navigator;
+    final context = navigator.context;
+    assert(context.mounted);
+    if (!context.mounted) return; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
+
+    final notifNavData = _tryParseIosApnsPayload(context, event.payload);
```
gnprice, line 131: Looks like this `_navigateForNotification` method should get an assert that the platform is iOS, to clarify why this line makes sense.

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,320 @@
...
+    return MessageListPage.buildRoute(
+      accountId: account.id,
+      // TODO(#82): Open at specific message, not just conversation
+      narrow: data.narrow);
+  }
+
+  /// Navigates to the [MessageListPage] of the specific conversation
+  /// for the provided payload that was attached while creating the
+  /// notification.
+  Future<void> _navigateForNotification(NotificationTapEvent event) async {
```
gnprice, line 123: Can this method be static? Doesn't look like it's using any instance data.

gnprice (commented):
Thanks for splitting up that commit! That was very helpful for my reading those changes. Here's a full review of the resulting two commits:
80a89e337 notif [nfc]: Introduce NotificationNavigationServer
b8cfff734 notif ios: Navigate when app launched from notification

except the tests.

I'll take a look at those tests next. Then the remaining to-do list for review will be the final commit:
1c0395087 notif ios: Navigate when app running but in background

test/model/store_test.dart
```diff
@@ -1290,8 +1291,9 @@ void main() {
       // (This is probably the common case.)
       addTearDown(testBinding.reset);
       testBinding.firebaseMessagingInitialToken = '012abc';
-      addTearDown(NotificationService.debugReset);
       testBinding.packageInfoResult = eg.packageInfo(packageName: 'com.zulip.flutter');
+      addTearDown(NotificationService.debugReset);
+      addTearDown(NotificationNavigationService.debugReset);
       await NotificationService.instance.start();
```
gnprice, line 1297: Hmm yeah, good thought reordering these while you're here editing them.

test/notifications/display_test.dart
```diff
@@ -121,6 +122,7 @@ void main() {
     addTearDown(testBinding.reset);
     testBinding.firebaseMessagingInitialToken = '012abc';
     addTearDown(NotificationService.debugReset);
+    addTearDown(NotificationNavigationService.debugReset);
```
gnprice, line 125: It'd be good if we can avoid going and adding another detail like this that tests have to know about.

How about having `NotificationService.debugReset` call the new service's `debugReset`? Seems appropriate for the main service to take responsibility for resetting the new one, given that the main service is also the one place responsible for starting up the new service.

test/notifications/navigate_test.dart
```diff
@@ -0,0 +1,276 @@
...
+  };
+}
+
+void main() {
+  TestZulipBinding.ensureInitialized();
+  final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
+
+  Future<void> init() async {
+    addTearDown(testBinding.reset);
+    testBinding.firebaseMessagingInitialToken = '012abc';
```
gnprice, line 73: Is this line needed? I'd think this data wouldn't be involved in any of these tests.

test/notifications/display_test.dart
```diff
@@ -1256,39 +1258,39 @@ void main() {
     });
   });
 
-  group('NotificationOpenPayload', () {
+  group('NotificationNavigationData', () {
```
gnprice, line 1261: Since the code this is testing is getting moved between files, the tests should move to remain in a corresponding file.

lib/notifications/navigate.dart
```diff
@@ -0,0 +1,320 @@
...
+
+  NotificationNavigationData({
+    required this.realmUrl,
+    required this.userId,
+    required this.narrow,
+  });
+
+  /// Parses the iOS APNs payload and retrieves the information
+  /// required for navigation.
+  factory NotificationNavigationData.fromIosApnsPayload(Map<Object?, Object?> payload) {
```
gnprice, line 186: Hmm, does this function have tests? I don't see them. (I do see a variety of test cases for the logic downstream of here, which is good.)

This logic is fairly densely packed with fiddly details, like parsing logic tends to be, so it's logic that's valuable to have good tests for. And, again typically for parsing logic, it has a pretty crisp interface for both input and output, which makes it nicely amenable to writing plenty of unit tests. So let's go ahead and do that.

For a point of comparison, see the existing tests of `NotificationOpenPayload` in `display_test.dart`.

test/notifications/navigate_test.dart
```diff
@@ -0,0 +1,276 @@
...
+      await prepare(tester, dropStartingRoutes: false);
+      check(pushedRoutes).isEmpty(); // GlobalStore hasn't loaded yet
+
+      // Once the app is ready, we navigate to the conversation.
+      await tester.pump();
+      takeStartingRoutes(account: eg.selfAccount);
+      matchesNavigation(check(pushedRoutes).single, eg.selfAccount, message);
+    }, variant: const TargetPlatformVariant({TargetPlatform.iOS}));
+
+    testWidgets('(iOS) uses associated account as initial account; if initial route', (tester) async {
```
gnprice, line 255: I don't understand what a ";" means here.

test/notifications/navigate_test.dart
```diff
@@ -0,0 +1,276 @@
...
+    }, variant: const TargetPlatformVariant({TargetPlatform.iOS}));
+
+    testWidgets('(iOS) at app launch', (tester) async {
+      addTearDown(testBinding.reset);
+      // Set up a value for `PlatformDispatcher.defaultRouteName` to return,
+      // for determining the initial route.
+      final message = eg.streamMessage();
+      final payload = messageApnsPayload(message, account: eg.selfAccount);
+      testBinding.notificationPigeonApi.setNotificationDataFromLaunch(
+        NotificationDataFromLaunch(payload: payload));
```
gnprice, line 243: Is this comment accurate? I don't see anything about `PlatformDispatcher.defaultRouteName` here.

It looks like this test case is copied from an existing one in `display_test.dart`. But that one says:
```dart
      // Set up a value for `PlatformDispatcher.defaultRouteName` to return,
      // for determining the initial route.
      final account = eg.selfAccount;
      final message = eg.streamMessage();
      final data = messageFcmMessage(message, account: account);
      final intentDataUrl = NotificationNavigationData(
        /* … */).buildAndroidNotificationUrl();
      addTearDown(tester.binding.platformDispatcher.clearDefaultRouteNameTestValue);
      tester.binding.platformDispatcher.defaultRouteNameTestValue = intentDataUrl.toString();
```

so `PlatformDispatcher.defaultRouteName` really is what it's about.

test/notifications/display_test.dart
```diff
@@ -1198,19 +1200,19 @@ void main() {
     testWidgets('at app launch', (tester) async {
       addTearDown(testBinding.reset);
       // Set up a value for `PlatformDispatcher.defaultRouteName` to return,
-      // for determining the intial route.
+      // for determining the initial route.
       final account = eg.selfAccount;
       final message = eg.streamMessage();
       final data = messageFcmMessage(message, account: account);
-      final intentDataUrl = NotificationOpenPayload(
+      final intentDataUrl = NotificationNavigationData(
```
gnprice, line 1207: Speaking of this test, it looks like it (and several of its neighbors) should also move to `navigate_test.dart` / `open_test.dart` as part of the code-moving commit — the code they're testing is part of what's getting moved.

(In particular the way these tests exercise the code they're testing is via ZulipApp — and the relevant call sites in ZulipApp switch from NotificationDisplayManager methods to NotificationNavigationService methods.)

As a bonus, those moves may allow you to outright move some of the helpers from this test file, rather than copy them as the current version does.

lib/notifications/display.dart
```diff
@@ -530,7 +491,12 @@ class NotificationDisplayManager {
     assert(context.mounted);
     if (!context.mounted) return; // TODO(linter): this is impossible as there's no actual async gap, but the use_build_context_synchronously lint doesn't see that
 
-    final route = routeForNotification(context: context, url: url);
+    assert(url.scheme == 'zulip' && url.host == 'notification');
+    final payload =
+      NotificationNavigationService.tryParseAndroidNotificationUrl(context, url);
```
gnprice, line 496: This `navigateForNotification` method seems like it should move too — it's logically closely related to the other methods like `routeForNotification` which are moving.

It's also quite similar to the `_navigateForNotification` method which the last commit introduces on the other class. It might be fine to have some code duplication between these two methods — but if we do that, it's definitely better if they're right next to each other, so that the parallelism is more visible.

gnprice (commented):
OK, and here's a review on the tests in that next-to-last commit:
b8cfff734 notif ios: Navigate when app launched from notification

I think some of the changes requested here will help make those tests easier to read and understand, so for this round I haven't yet read every line.

rajveermalviya:

test/notifications/open_test.dart
```diff
@@ -0,0 +1,577 @@
...
+      );
+      url = payload.buildAndroidNotificationUrl();
+      check(NotificationOpenPayload.parseAndroidNotificationUrl(url))
+        ..realmUrl.equals(payload.realmUrl)
+        ..userId.equals(payload.userId)
+        ..narrow.equals(payload.narrow);
+    });
+
+    group('parseIosApnsPayload', () {
+      test('smoke one-one DM', () {
```
gnprice, line 350: Cool, these tests are a good start.

Because of the nature of this function as parsing code (echoing https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2099235693), and because notifications are important, I'd like to have more thorough testing for it than only smoke tests.

But let's make that a follow-up issue we can pursue after launch. Even if there are edge cases this code doesn't do anything on, that won't be a regression from what's in main. (And if there are edge cases where it does something but it's the wrong thing, going to the wrong narrow is probably not much worse than doing nothing.)

test/notifications/open_test.dart
```diff
@@ -0,0 +1,577 @@
...
+
+    Future<void> openNotification(WidgetTester tester, Account account, Message message) async {
+      switch (defaultTargetPlatform) {
+        case TargetPlatform.android:
+          final intentDataUrl = androidNotificationUrlForMessage(account, message);
+          unawaited(
+            WidgetsBinding.instance.handlePushRoute(intentDataUrl.toString()));
+          await tester.idle(); // let navigateForNotification find navigator
+
+        case TargetPlatform.iOS:
```
gnprice, line 149: Cool, I like this mechanism for reusing the same test code for both platforms.

lib/notifications/open.dart
```diff
@@ -0,0 +1,346 @@
...
+
+  Future<void> start() async {
+    assert(_initializedSignal == null);
+    _initializedSignal = Completer<void>();
+    try {
+      switch (defaultTargetPlatform) {
+        case TargetPlatform.iOS:
+          _notifDataFromLaunch = await _notifPigeonApi.getNotificationDataFromLaunch();
+          _notifPigeonApi.notificationTapEventsStream()
+            .listen(_navigateForNotification);
```
gnprice, line 53: What happens if the user opens the app from the home screen, and then immediately taps a notification, while the app is still just beginning to start up and so before it reaches this `listen` call?

In that case the notification won't get passed at startup to `application(_:didFinishLaunchingWithOptions:)`. Instead it'll go to `userNotificationCenter(_:didReceive:withCompletionHandler:)`. Will it make it to the app's Dart code so that the right conversation gets opened, or will it get dropped because there isn't a listener yet?

Looking at the Swift code, I think it will get dropped:
```swift
  override func onListen(withArguments arguments: Any?, sink: PigeonEventSink<NotificationTapEvent>) {
    eventSink = sink
  }

  func onNotificationTapEvent(payload: [AnyHashable : Any]) {
    eventSink?.success(NotificationTapEvent(payload: payload))
  }
```

I think the ideal behavior would be that instead it gets buffered, stored in the stream, and then when the app reaches this point it consumes any notification that's been buffered there.

If there isn't an obvious tweak to make that would provide that behavior, we can cheerfully leave it as a followup. I think this isn't a super common case.

gnprice (commented):
Cool, thanks for splitting these out as separate commits — that's very helpful for reading them. They all look good.
529265367 notif [nfc]: Rename NotificationOpenPayload methods
d130717db notif [nfc]: Refactor NotificationOpenService.routeForNotification
b2e0ed07f notif: Show a dialog if received malformed Android Notification URL

I've now finished reading the whole thing, and generally it all looks good; just a handful of comments below. A couple of them are for follow-ups we can pursue after launch.

There's one commit making a bunch of test changes which I'd like split up to make clearer — as is, I don't yet fully understand all the things it's doing, which means it's hard to be sure whether it's accidentally undermining some of the tests. (Which in turn would mean there could be a bug that the tests are no longer able to catch.)

chrisbobbe:
There's a breaking change coming from upstream that I think might affect this work: https://github.com/zulip/zulip-flutter/pull/1546#issuecomment-2941064551

test/notifications/open_test.dart
```diff
@@ -0,0 +1,577 @@
...
+      if (!dropStartingRoutes) {
+        check(pushedRoutes).isEmpty();
+        return;
+      }
+      await tester.pump();
+      takeStartingRoutes(account: account);
+      check(pushedRoutes).isEmpty();
+    }
+
+    Uri androidNotificationUrlForMessage(Account account, Message message) {
```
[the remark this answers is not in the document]
rajveermalviya: I've removed the unnecessary refactors in the latest revision but kept the changes for `androidNotificationUrlForMessage`, `openNotification` and `setupNotificationDataForLaunch` in the test refactor commit (075bae8c0) as they were essential for simpler diff for these tests in the later commits.

test/notifications/open_test.dart
rajveermalviya: Filed: https://github.com/zulip/zulip-flutter/issues/1563

lib/notifications/open.dart
rajveermalviya: I tried the buffering implementation but weirdly I was observing some repeated navigations when tapped on a notification. So, I think I will leave this as a follow up, to investigate more about it later.

rajveermalviya:

lib/notifications/navigate.dart
gnprice: > For example, we might want to take you directly to the relevant message in history, rather than to the first unread or the newest message. (Maybe it was an @-mention in the middle of a long-running conversation that you weren't previously reading.)

(Just filed an issue for this: #1565.)

lib/notifications/open.dart
gnprice: OK, sounds good. Please go ahead and file that follow-up issue, then.

test/notifications/open_test.dart
```diff
@@ -0,0 +1,577 @@
...
+      if (early) {
+        check(pushedRoutes).isEmpty();
+        return;
+      }
+      await tester.pump();
+      takeStartingRoutes(withAccount: withAccount);
+      check(pushedRoutes).isEmpty();
+    }
+
+    Uri androidNotificationUrlForMessage(Account account, Message message) {
```
gnprice, line 124: notif test: Refactor tests for NotificationOpenService

Cool, this commit is pretty crisp to read, thanks. (Continuing the thread from https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2114567679 .)

It's an NFC change, right? Even to the tests. (Or does it have an effect on the tests' behavior which I'm missing?) If so, let's specify that in the commit message.

We can also make the message more specific about what the refactor is doing. Like:
> notif test [nfc]: Pull out androidNotificationUrlForMessage and setupNotificationDataForLaunch

That's information the reader basically has to work out one way or another in order to understand what the commit is doing and confirm it doesn't affect how the tests behave. So it's helpful to give them that information up front.

gnprice (commented):
Thanks for the revision! This all looks good. One nit below, and see also one comment above: https://github.com/zulip/zulip-flutter/pull/1379#discussion_r2143208877

Then this is otherwise ready to merge.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

PR #1380, opened 2025-02-27
check: Ensure GNU coreutils are used on macOS

rajveermalviya:
Without this, there are failures when running `./tools/check` script with `--verbose` flag:

    Test command: ./tools/check pigeon --all-files --fix --verbose
    date: illegal option -- -
    usage: date [-jnRu] [-I[date|hours|minutes|seconds]] [-f input_fmt]
                [-r filename|seconds] [-v[+|-]val[y|m|w|d|H|M|S]]
                [[[[mm]dd]HH]MM[[cc]yy][.SS] | new_date] [+output_fmt]
    Time now:
    zulip-flutter e0df0ed10 • 2025-02-26 05:01:14 +0000

### DIFF
--- a/tools/check
+++ b/tools/check
@@ -18,6 +18,9 @@ set -euo pipefail
 
 this_dir=${BASH_SOURCE[0]%/*}
 
+# shellcheck source=tools/lib/ensure-coreutils.sh
+. "${this_dir}"/lib/ensure-coreutils.sh
+
 # shellcheck source=tools/lib/git.sh
 . "${this_dir}"/lib/git.sh
 


gnprice:
Good catch, thanks.

Merging with one style tweak:
```diff
 
 # shellcheck source=tools/lib/ensure-coreutils.sh
 . "${this_dir}"/lib/ensure-coreutils.sh
+
 # shellcheck source=tools/lib/git.sh
 . "${this_dir}"/lib/git.sh
 
```

I think I'd be happy to have the `. …` lines right next to each other if they didn't need those `# shellcheck …` comments; but since they do, it feels a bit easier to read with a blank line there to visually group that comment with the next line rather than the previous.

Issue #1354, opened 2025-02-13
Runtime error if user logs out when reloading PerAccountStore


PIG208:
`loadPerAccount` has a check for whether the account exists or not at the end:
```dart
  Future<PerAccountStore> loadPerAccount(int accountId) async {
    assert(_accounts.containsKey(accountId));
    final store = await doLoadPerAccount(accountId);
    if (!_accounts.containsKey(accountId)) {
      // [removeAccount] was called during [doLoadPerAccount].
      store.dispose();
      throw AccountNotFoundException();
    }
    return store;
  }
```

Hypothetically, this check can only be reached if the account is logged out while loading the `PerAccountStore`. However, because of a runtime error that would happen right after the inner request returns, we cannot reach the if-block here.

The error comes from a null-check at the start of `PerAccountStore.fromInitialSnapshot`:

```dart
    final account = globalStore.getAccount(accountId)!;
```

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Null check operator used on a null value
#0      new PerAccountStore.fromInitialSnapshot (package:zulip/model/store.dart:285:54)
#1      UpdateMachine.load (package:zulip/model/store.dart:913:35)
<asynchronous suspension>
#2      LiveGlobalStore.doLoadPerAccount (package:zulip/model/store.dart:836:27)
<asynchronous suspension>
#3      GlobalStore.loadPerAccount (package:zulip/model/store.dart:155:15)
<asynchronous suspension>
#4      GlobalStore._reloadPerAccount (package:zulip/model/store.dart:135:19)
```

`PerAccountStore.fromInitialSnapshot` is called near the end of `LiveGlobalStore.doLoadPerAccount` with no asynchronous gap, meaning that the null-check will always fail ahead of `AccountNotFoundException` being raised if the user logs out during the asynchronous gap from the /api/register-queue request or `globalStore.updateAccount`:

This is a bug that the `if (!_accounts.containsKey(accountId))` check was meant to catch. The error should be thrown earlier, ideally every time before the account is accessed, similar to the `if (mounted)` checks and `BuildContext`.

gnprice:
> Hypothetically, this check can only be reached if the account is logged out while loading the `PerAccountStore`. However, because of a runtime error that would happen right after the inner request returns, we cannot reach the if-block here.

Hmm, I don't understand. If we can't ever reach that case, that should mean that that conditional has no effect, right?

[...]

PR #1386, opened 2025-03-07
store: Have UpdateMachine.load check account still exists after async gaps

chrisbobbe:
Fixes: #1354

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -170,6 +170,8 @@ abstract class GlobalStore extends ChangeNotifier {
 
   /// Load per-account data for the given account, unconditionally.
   ///
+  /// The account for `accountId` must exist.
+  ///
   /// This method should be called only by the implementation of [perAccount].
   /// Other callers interested in per-account data should use [perAccount]
   /// and/or [perAccountSync].
@@ -213,6 +215,8 @@ abstract class GlobalStore extends ChangeNotifier {
   }
 
   /// Load per-account data for the given account, unconditionally.
+  ///
+  /// The account for `accountId` must exist.
   ///
   /// This method should be called only by [loadPerAccount].
   Future<PerAccountStore> doLoadPerAccount(int accountId);
@@ -264,6 +268,8 @@ abstract class GlobalStore extends ChangeNotifier {
   Future<void> doUpdateAccount(int accountId, AccountsCompanion data);
 
   /// Remove an account from the store.
+  ///
+  /// The account for `accountId` must exist.
   Future<void> removeAccount(int accountId) async {
     assert(_accounts.containsKey(accountId));
     await doRemoveAccount(accountId);
@@ -942,7 +948,10 @@ class UpdateMachine {
     store.updateMachine = this;
   }
 
-  /// Load the user's data from the server, and start an event queue going.
+  /// Load data for the given account from the server,
+  /// and start an event queue going.
+  ///
+  /// The account for `accountId` must exist.
   ///
   /// In the future this might load an old snapshot from local storage first.
   static Future<UpdateMachine> load(GlobalStore globalStore, int accountId) async {
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -154,6 +154,7 @@ abstract class GlobalStore extends ChangeNotifier {
   }
 
   Future<void> _reloadPerAccount(int accountId) async {
+    assert(_accounts.containsKey(accountId));
     assert(_perAccountStores.containsKey(accountId));
     assert(!_perAccountStoresLoading.containsKey(accountId));
     final store = await loadPerAccount(accountId);
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -147,10 +147,13 @@ abstract class GlobalStore extends ChangeNotifier {
     // It's up to us.  Start loading.
     future = loadPerAccount(accountId);
     _perAccountStoresLoading[accountId] = future;
-    store = await future;
-    _setPerAccount(accountId, store);
-    unawaited(_perAccountStoresLoading.remove(accountId));
-    return store;
+    try {
+      store = await future;
+      _setPerAccount(accountId, store);
+      return store;
+    } finally {
+      unawaited(_perAccountStoresLoading.remove(accountId));
+    }
   }
 
   Future<void> _reloadPerAccount(int accountId) async {
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -175,6 +175,9 @@ abstract class GlobalStore extends ChangeNotifier {
   ///
   /// The account for `accountId` must exist.
   ///
+  /// Throws [AccountNotFoundException] if after any async gap
+  /// the account has been removed.
+  ///
   /// This method should be called only by the implementation of [perAccount].
   /// Other callers interested in per-account data should use [perAccount]
   /// and/or [perAccountSync].
@@ -189,37 +192,29 @@ abstract class GlobalStore extends ChangeNotifier {
           // The API key is invalid and the store can never be loaded
           // unless the user retries manually.
           final account = getAccount(accountId);
-          if (account == null) {
-            // The account was logged out during `await doLoadPerAccount`.
-            // Here, that seems possible only by the user's own action;
-            // the logout can't have been done programmatically.
-            // Even if it were, it would have come with its own UI feedback.
-            // Anyway, skip showing feedback, to not be confusing or repetitive.
-            throw AccountNotFoundException();
-          }
+          assert(account != null); // doLoadPerAccount would have thrown AccountNotFoundException
           final zulipLocalizations = GlobalLocalizations.zulipLocalizations;
           reportErrorToUserModally(
             zulipLocalizations.errorCouldNotConnectTitle,
             message: zulipLocalizations.errorInvalidApiKeyMessage(
-              account.realmUrl.toString()));
+              account!.realmUrl.toString()));
           await logOutAccount(this, accountId);
           throw AccountNotFoundException();
         default:
           rethrow;
       }
     }
-    if (!_accounts.containsKey(accountId)) {
-      // TODO(#1354): handle this earlier
-      // [removeAccount] was called during [doLoadPerAccount].
-      store.dispose();
-      throw AccountNotFoundException();
-    }
+    // doLoadPerAccount would have thrown AccountNotFoundException
+    assert(_accounts.containsKey(accountId));
     return store;
   }
 
   /// Load per-account data for the given account, unconditionally.
   ///
   /// The account for `accountId` must exist.
+  ///
+  /// Throws [AccountNotFoundException] if after any async gap
+  /// the account has been removed.
   ///
   /// This method should be called only by [loadPerAccount].
   Future<PerAccountStore> doLoadPerAccount(int accountId);
@@ -956,13 +951,26 @@ class UpdateMachine {
   ///
   /// The account for `accountId` must exist.
   ///
+  /// Throws [AccountNotFoundException] if after any async gap
+  /// the account has been removed.
+  ///
   /// In the future this might load an old snapshot from local storage first.
   static Future<UpdateMachine> load(GlobalStore globalStore, int accountId) async {
     Account account = globalStore.getAccount(accountId)!;
     final connection = globalStore.apiConnectionFromAccount(account);
 
+    void stopAndThrowIfNoAccount() {
+      final account = globalStore.getAccount(accountId);
+      if (account == null) {
+        assert(debugLog('Account logged out during UpdateMachine.load'));
+        connection.close();
+        throw AccountNotFoundException();
+      }
+    }
+
     final stopwatch = Stopwatch()..start();
-    final initialSnapshot = await _registerQueueWithRetry(connection);
+    final initialSnapshot = await _registerQueueWithRetry(connection,
+      stopAndThrowIfNoAccount: stopAndThrowIfNoAccount);
     final t = (stopwatch..stop()).elapsed;
     assert(debugLog("initial fetch time: ${t.inMilliseconds}ms"));
 
@@ -1004,13 +1012,20 @@ class UpdateMachine {
 
   bool _disposed = false;
 
+  /// Make the register-queue request, with retries.
+  ///
+  /// After each async gap, calls [stopAndThrowIfNoAccount].
   static Future<InitialSnapshot> _registerQueueWithRetry(
-      ApiConnection connection) async {
+    ApiConnection connection, {
+    required void Function() stopAndThrowIfNoAccount,
+  }) async {
     BackoffMachine? backoffMachine;
     while (true) {
+      InitialSnapshot? result;
       try {
-        return await registerQueue(connection);
+        result = await registerQueue(connection);
       } catch (e, s) {
+        stopAndThrowIfNoAccount();
         // TODO(#890): tell user if initial-fetch errors persist, or look non-transient
         switch (e) {
           case HttpException(httpStatus: 401):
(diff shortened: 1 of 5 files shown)


test/model/test_store.dart
```diff
@@ -1,29 +1,16 @@
...
-/// there is no automatic event-polling loop or other automated requests.
-/// For each account loaded, there is a corresponding [UpdateMachine]
-/// in [updateMachines], which tests can use for invoking that logic
-/// explicitly when desired.
-///
-/// See also [TestZulipBinding.globalStore], which provides one of these.
-class TestGlobalStore extends GlobalStore {
-  TestGlobalStore({required super.accounts});
-
+mixin _ApiConnectionsMixin on GlobalStore {
```
PIG208, line 13: This is pretty cool! I think the layout of these mixins will probably be helpful guidance to later break down `GlobalStore` itself as well; similar to what we did with `ChannelStore`, etc.

test/model/test_store.dart
```diff
@@ -142,9 +107,55 @@ class TestGlobalStore extends GlobalStore {
...
+/// Unlike with [LiveGlobalStore] and the associated [UpdateMachine.load],
+/// there is no automatic event-polling loop or other automated requests.
+/// For each account loaded, there is a corresponding [UpdateMachine]
+/// in [updateMachines], which tests can use for invoking that logic
+/// explicitly when desired.
+///
+/// See also:
+///   * [TestZulipBinding.globalStore], which provides one of these.
+///   * [UpdateMachineTestGlobalStore], which prepares per-account data
+///     using [UpdateMachine.load] (like [LiveGlobalStore] does)..
```
PIG208, line 131: nit: is the extra "." intentional here?

test/model/store_test.dart
```diff
@@ -966,69 +1026,69 @@ void main() {
...
-      updateMachine.debugPauseLoop();
+      final store = await globalStore.perAccount(eg.selfAccount.id);
+      final updateMachine = store.updateMachine!;
       updateMachine.poll();
 
-      (store.connection as FakeApiConnection).prepare(
+      final connection = store.connection as FakeApiConnection;
+      connection.prepare(
         apiException: eg.apiExceptionBadEventQueueId());
+      globalStore.clearCachedApiConnections();
```
PIG208, line 1043: I'm not sure if I understand how this works. We prepare an error with the current `connection`, but call `clearCachedApiConnections` immediately after.

Is it expected that the update machine still has the connection with the prepared response when polling, but it won't be after reloading?

test/model/store_test.dart
```diff
@@ -130,18 +131,70 @@ void main() {
...
+    check(connection.takeRequests()).isEmpty();
+    check(connection).isOpen.isFalse();
+  }));
+
   test('GlobalStore.perAccount account is logged out while loading; then fails with HTTP status code 401', () => awaitFakeAsync((async) async {
-    final globalStore = LoadingTestGlobalStore(accounts: [eg.selfAccount]);
+    final globalStore = UpdateMachineTestGlobalStore(accounts: [eg.selfAccount]);
+    globalStore.prepareRegisterQueueResponse = (connection) =>
+      connection.prepare(
+        delay: TestGlobalStore.removeAccountDuration + Duration(seconds: 1),
```
PIG208, line 159: Because both `UpdateMachineTestGlobalStore` and `TestGlobalStore` share `removeAccountDuration` through `_DatabaseMixin`, perhaps we should find a new home to this constant. Maybe a "k"-prefixed const?

lib/model/store.dart
```diff
@@ -893,15 +896,31 @@ class UpdateMachine {
...
+        throw AccountNotFoundException();
+      }
+    }
+
     final stopwatch = Stopwatch()..start();
-    final initialSnapshot = await _registerQueueWithRetry(connection);
+    final initialSnapshot = await _registerQueueWithRetry(connection,
+      stopAndThrowIfNoAccount: stopAndThrowIfNoAccount);
     final t = (stopwatch..stop()).elapsed;
     assert(debugLog("initial fetch time: ${t.inMilliseconds}ms"));
```
PIG208, line 926: Further down there is another async gap at `await globalStore.updateAccount`; should we call `stopAndThrowIfNoAccount` there as well?

PIG208 (commented):
Thanks for working on this! I like the new `UpdateMachineTestGlobalStore`. Left some thoughts and comments.

lib/model/store.dart
```diff
@@ -160,35 +169,29 @@ abstract class GlobalStore extends ChangeNotifier {
...
-    }
+    // doLoadPerAccount would have thrown AccountNotFoundException
+    assert(_accounts.containsKey(accountId));
     return store;
   }
 
-  /// Load per-account data for the given account, unconditionally.
+  /// Load per-account data for the given account ID, unconditionally.
+  ///
+  /// The account for `accountId` must exist.
```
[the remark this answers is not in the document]
chrisbobbe: Hmm, I think I wouldn't be satisfied with an approach like that.

Within some synchronous code, say point A is where we retrieve the Account data (`getAccount`), and points B, C, and D are where we need to make decisions about it or pass it across some boundary where point A couldn't go (like if we call a platform API). I'd be happier if we could freely put point A where it makes the most sense for code organization, which will depend on the context; maybe we want to shorten the paths to B, C, and D, or maybe to put it neatly at the top of a function body. If the placement of point A affects behavior (because that's where we do the null-check), It feels much less flexible and harder to reason about. We can get that flexibility by doing `stopAndThrowIfNoAccount` after the async gaps.

> only check for its existence on use

This wouldn't answer the need to interrupt the retry loop in `_registerQueueWithRetry` (the symptom I mentioned at https://github.com/zulip/zulip-flutter/issues/1354#issuecomment-2705336628 ); we need to know if the account has been logged out, so we can stop retrying, but we don't actually use the account's data.

test/model/store_test.dart
chrisbobbe: Yeah, thanks for pointing out that this is confusing. I wasn't sure where would be the least confusing place to put `globalStore.clearCachedApiConnections();`.

My idea in putting it here was that it's right next to the `globalStore.prepareRegisterQueueResponse =(newConnection) { /* … */ }`.

The "cache" is test infrastructure; see `_ApiConnectionsMixin.useCachedApiConnections`. When we look at app code that runs during a test, the cache is only relevant when the code asks to create a new `ApiConnection`, which it does with `GlobalStore.apiConnection` or `GlobalStore.apiConnectionFromAccount`.

In the app code exercised by these tests, that happens twice:
1. When the test calls `await globalStore.perAccount` (near the top of `prepareReload`)
2. When the `UpdateMachine` responds to the bad-event-queue error by re-registering.

The test 'user logged out before new store is loaded' would break if step 2 picked up a connection that was cached in step 1. That's because the connection from step 1 is disposed during logout, and it's not meant to be used after that. So, we clear the cache between step 1 and step 2.

The `UpdateMachineTestGlobalStore` uses the caching behavior to make `doLoadPerAccount` work, but it's still fine to clear the cache between calls to `doLoadPerAccount`.

chrisbobbe: I'll try to make it more clear in my next revision :)

test/model/store_test.dart
chrisbobbe: Maybe; let's see if @gnprice has thoughts on that. I showed him in the office that a static `_DatabaseMixin.removeAccountDuration` wouldn't work (callers would have to access it using the mixin's name, like `FooMixin.removeAccountDuration`, and it doesn't make sense for the mixin to be public).

lib/model/store.dart
chrisbobbe: I believe there isn't a bug that we could catch by doing so.
1. We hope and expect that `await globalStore.updateAccount` always fails if the account is removed concurrently. If it doesn't, that's a bug in the database implementation.
2. If `await globalStore.updateAccount` succeeds, that means there's still an `Account` object in `GlobalStore._accounts`. Since `stopAndThrowIfNoAccount` looks at `GlobalStore._accounts` too, it would find it there and wouldn't throw. So even if the (1) bug exists, `stopAndThrowIfNoAccount` wouldn't help us catch it.

test/model/store_test.dart
gnprice: Yeah, I think this is fine. If we find ourselves with more items like this, maybe we'll find some other way to organize them.

test/model/store_test.dart
gnprice: I think a good comment on this `clearCachedApiConnections` call can probably replace the need for this bit below:
```dart
        // Assert this is actually a new connection, not the one on
        // globalStore._perAccountStores, which at least one test will close
        // via logOutAccount.
        assert(!identical(connection, newConnection));
```

(Because that assert is basically checking that this `clearCached…` happened, right?)

test/model/store_test.dart
```diff
@@ -966,69 +1026,69 @@ void main() {
...
-      completers().single.complete(store);
-      await future;
-      completers().clear();
-      final updateMachine = globalStore.updateMachines[eg.selfAccount.id] =
-        UpdateMachine.fromInitialSnapshot(
-          store: store, initialSnapshot: eg.initialSnapshot());
-      updateMachine.debugPauseLoop();
+      final store = await globalStore.perAccount(eg.selfAccount.id);
+      final updateMachine = store.updateMachine!;
       updateMachine.poll();
```
gnprice, line 1038: Should this `poll` call be here? It looks like it now duplicates one that `UpdateMachine.load` makes.

test/model/test_store.dart
```diff
@@ -166,6 +177,59 @@ class TestGlobalStore extends GlobalStore {
...
+/// Callers can set [prepareRegisterQueueResponse]
+/// to prepare a register-queue payload or an exception.
+/// The implementation pauses the event-polling loop
+/// to avoid being a nuisance and does a boring
+/// [FakeApiConnection.prepare] for the register-token request.
+///
+/// See also:
+///   * [TestGlobalStore], which prepares per-account data
+///     without using [UpdateMachine.load].
+class UpdateMachineTestGlobalStore extends GlobalStore with _ApiConnectionsMixin, _DatabaseMixin {
```
gnprice, line 196: I spent some time staring at this vs. TestGlobalStore and thinking about the relationship between them. I didn't see any big changes I'd want to make; but the `updateMachines` field over on that other class was one thing that was a bit incongruous with the naming of the classes, and I realized it wasn't needed.

I'll push an extra commit that cuts that out, as well as a prep commit in a different area that this brought up:
dd9b29bb8 test [nfc]: Simplify out TestGlobalStore.updateMachines
247bf7ffe store test [nfc]: Clarify what prepareReload is doing

test/model/store_test.dart
chrisbobbe: Indeed; thanks.

test/model/test_store.dart
chrisbobbe: Great; thanks!

chrisbobbe:
Thanks for the reviews! Revision pushed, but GitHub seems stuck and I'm not sure if it took—the branch should be at 1ea55a9439b03c9e5dc45e11d644c2ecec88095d.

PIG208:
I haven't seen this message before. Looks like it's just taking longer than usual: https://github.blog/changelog/2023-09-26-more-details-provided-when-a-pull-request-is-merged-indirectly-or-is-still-processing-updates/#pushed-commits-are-still-being-processed

gnprice:
Weird — that's a long delay at this point (40 minutes since Chris's comment above).

Try pushing again (with some trivial change)? Possibly that'd unwedge it.

lib/model/store.dart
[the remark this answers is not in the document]
PIG208: Bumping https://github.com/zulip/zulip-flutter/pull/1386#discussion_r1985661741; I guess we still want to mention the constraint to its dartdoc?

---

Thanks for the explanation. I think `Account` will go well with a version of [use_build_context_synchronously](https://dart.dev/tools/linter-rules/use_build_context_synchronously), if we have a way to annotate this type somehow.

test/model/store_test.dart
PIG208: Thanks! The new ordering and block of comments makes it quite clear to me.

PIG208:
Looks good to me other than a nit. Thanks for the updates!

gnprice:
I fetched https://github.com/zulip/zulip-flutter/commit/1ea55a9439b03c9e5dc45e11d644c2ecec88095d from Chris's fork, made a no-op change, and pushed that. This time the push seems to have worked.

lib/model/store.dart
```diff
@@ -908,15 +911,31 @@ class UpdateMachine {
...
   }
 
-  /// Load the user's data from the server, and start an event queue going.
+  /// Load data for the given account ID from the server,
+  /// and start an event queue going.
+  ///
+  /// The account for `accountId` must exist.
+  ///
+  /// Throws [AccountNotFoundException] after an async gap
+  /// if the account has been removed.
```
gnprice, line 920:
```suggestion
  /// Throws [AccountNotFoundException] if after any async gap
  /// the account has been removed.
```

Otherwise it sounds like it's saying: if the account has been removed, throws said exception, but first causes an async gap.

gnprice (commented):
Thanks! Generally this looks great. One added test I'd like to see, and otherwise nits.

test/model/store_test.dart
```diff
@@ -130,18 +131,70 @@ void main() {
...
+
+    await logOutAccount(globalStore, eg.selfAccount.id);
+    check(globalStore.takeDoRemoveAccountCalls())
+      .single.equals(eg.selfAccount.id);
+
+    await check(future).throws<AccountNotFoundException>();
+    check(globalStore.takeDoRemoveAccountCalls()).isEmpty();
+    // no poll, server-emoji-data, or register-token requests
+    check(connection.takeRequests()).isEmpty();
+    check(connection).isOpen.isFalse();
```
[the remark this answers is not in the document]
chrisbobbe: > these tests

You mean a test where the loading succeeds, right, so just one test? That's the only condition where the poll, server-emoji-data, and register-token requests are made.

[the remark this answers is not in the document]
gnprice: Right, exactly.

chrisbobbe:

test/model/store_test.dart
```diff
@@ -157,18 +158,89 @@ void main() {
...
+    final globalStore = UpdateMachineTestGlobalStore(
+      globalSettings: eg.globalSettings(), accounts: [eg.selfAccount]);
+    final connection = globalStore.apiConnectionFromAccount(eg.selfAccount) as FakeApiConnection;
+
+    NotificationService.instance.token = ValueNotifier('asdf');
+    addTearDown(NotificationService.debugReset);
+
+    final future = globalStore.perAccount(eg.selfAccount.id);
+    check(connection.takeRequests()).length.equals(1); // register request
+    await future;
```
gnprice, line 171: nit: let's tighten the comparison with the tests below by rearranging slightly, like so:
```suggestion
    NotificationService.instance.token = ValueNotifier('asdf');
    addTearDown(NotificationService.debugReset);

    final globalStore = UpdateMachineTestGlobalStore(
      globalSettings: eg.globalSettings(), accounts: [eg.selfAccount]);
    final connection = globalStore.apiConnectionFromAccount(eg.selfAccount) as FakeApiConnection;
    final future = globalStore.perAccount(eg.selfAccount.id);
    check(connection.takeRequests()).length.equals(1); // register request

    await future;
```

gnprice (commented):
Thanks! Small comments below.

test/model/store_test.dart
```diff
@@ -157,18 +158,89 @@ void main() {
     await check(future).throws<AccountNotFoundException>();
   }));
 
+  test('GlobalStore.perAccount loading succeeds', () => awaitFakeAsync((async) async {
+    final globalStore = UpdateMachineTestGlobalStore(
+      globalSettings: eg.globalSettings(), accounts: [eg.selfAccount]);
```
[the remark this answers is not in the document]
gnprice: Sent #1403 with that prep commit as a quick PR.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

PR #1388, opened 2025-03-07
Revert ios build workaround for Firebase packages

rajveermalviya:
This reverts commit a9b20167ccf13ebaf6bcd6a28060ca4438f8c8a8.

### DIFF
--- a/ios/Flutter/Zulip.xcconfig
+++ b/ios/Flutter/Zulip.xcconfig
@@ -1,10 +1,2 @@
 // Configuration settings file format documentation can be found at:
 // https://help.apple.com/xcode/#/dev745c5c974
-
-// TODO(firebase/flutterfire#13323): remove this flag
-//
-// This flag is added to work around the iOS build failing
-// on Xcode 16.  Remove it when `package:firebase_messaging`
-// is patched with a fix to build successfully on Xcode 16:
-//   https://github.com/firebase/flutterfire/issues/13323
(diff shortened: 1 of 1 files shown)


PIG208 (commented):
Thanks! Looks like I missed this in my review list. I wasn't able to test this manually; from reading the changes and the linked context, this LGTM. Marking this for Greg's review.

gnprice (commented):
Thanks for cleaning this up!

Let's keep around the Zulip.xcconfig file and the bits of config that wire it up — it'll just become empty except for the general comment at top.

That way the next time there's some Xcode config we want to add, we have all the infrastructure in place to just drop it in that file and have it get applied. Effectively we'll get to re-use the work that we did in #958 to figure those details out.

rajveermalviya:

gnprice:
Thanks! Looks good; merging, with a few commit-message tweaks:

    -    Revert "ios: Add workaround for package:firebase_messaging when building with Xcode 16"
    +    ios build: Revert workaround for package:firebase_messaging
     
    -    This reverts part of commit a9b20167ccf13ebaf6bcd6a28060ca4438f8c8a8,
    -    keeping the `Zulip.xcconfig` file and it's references.
    +    This reverts part of commit a9b20167c (#958),
    +    keeping the `Zulip.xcconfig` file and its references.

(In particular, 'Revert "ios: …"' is no longer accurate because it's not reverting that whole commit.)

PR #1389, opened 2025-03-07
mailmap: Add Tomlin7's mailmap entry

tomlin7:
This pr adds my (`tomlin7`) entry to the `.mailmap` file because my name appears as `billy` in the git log. This wasn't intended, and it is my bad that I didn't keep my local git config up-to-date.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -18,6 +18,7 @@ K Akhil <kssakhilraj@gmail.com>
 Lalit Kumar Singh <lalitkumarsingh3716@gmail.com>
 Rajesh Malviya <rmalviya@zulip.com> <rajveer0malviya@gmail.com>
 Shu Chen <shu@zulip.com> <shu.chen@freelancedreams.com>
+Tomlin7 <billydevbusiness@gmail.com>
 
 # The goal when editing this file is to group all of a given person's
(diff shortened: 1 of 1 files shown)


gnprice:
Cool. Just to confirm: this writes the name with a capital "T", as in:
```
$ git log -1
commit cf952715e8a66600f059ab5bf635cb2248b63a9c (HEAD -> main)
Author: Tomlin7 <billydevbusiness@gmail.com>
Date:   Fri Mar 7 23:56:34 2025 +0530

    mailmap: Add tomlin7's entry
```

Is that intended? Checking only because you use a lower-case version of the same name in the commit message and PR title.

gnprice:
The capitalization is up to you. Most contributors use a name that has the shape of a full name, first and last, which is therefore capitalized. Others use something more in the style of a username, and those are usually lowercase.

Whatever name you use, my suggestion would be to capitalize it the same way you do when you use the same name elsewhere. So if "tomlin7" is the name you'd like to be credited by, I'd suggest using it in lowercase.

tomlin7:
Thanks. I will use the capitalized version. I have also updated the commit message and PR title.

gnprice:
Sounds good. Merged.

PR #1391, opened 2025-03-08
Support saved snippets

PIG208:
- [x] Viewing and inserting saved snippets
- [ ] Adding new saved snippets
- [ ] Handle saved snippet events


chrisbobbe:
This has a failing test, could you take a look please?

PIG208:
Oh right. Thanks! Pushed an update. I realized that I didn't make this clear — let's get #1511 reviewed first, since that's the prerequisite of this PR.

PR #1400, opened 2025-03-11
Support setting browser preference

### LINKED ISSUE #1228: Offer "Open links with in-app browser" setting
Like we do in the legacy apps.

An instance of:
- #97

PIG208:
Fixes #1228.

This is taken from #1167, addressing the reviews there.


lib/widgets/settings.dart
```diff
@@ -54,3 +56,26 @@ class _ThemeSetting extends StatelessWidget {
...
+        newOpenLinksWithInAppBrowser ? BrowserPreference.inApp
+                                     : BrowserPreference.external)));
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    final openLinksWithInAppBrowser =
+      GlobalStoreWidget.of(context).globalSettings.urlLaunchMode
+      == UrlLaunchMode.platformDefault;
```
gnprice, line 75: Given that we'll be presenting the setting as either "in-app: yes" or "in-app: no", let's make the semantics of our default match that presentation. IOW on Android let's adjust our default behavior so that it's the same as the behavior we end up with if the database says `inApp`.

That'd be good to do as a prep commit, before introducing the setting.

PIG208:
Thanks! This is ready for review.

<details>

<summary>Screenshot</summary>

in-app

</details>

test/widgets/content_test.dart
```diff
@@ -836,27 +849,27 @@ void main() {
...
+        .single.equals((url: Uri.parse('${eg.realmUrl}a/b?c#d'), mode: LaunchMode.inAppBrowserView));
     });
 
     testWidgets('link inside HeadingNode', (tester) async {
       await prepare(tester,
         '<h6><a href="https://a/">word</a></h6>');
       await tapText(tester, find.text('word'));
       check(testBinding.takeLaunchUrlCalls())
-        .single.equals((url: Uri.parse('https://a/'), mode: LaunchMode.platformDefault));
+        .single.equals((url: Uri.parse('https://a/'), mode: LaunchMode.inAppBrowserView));
```
gnprice, line 872: setting: Make in-app explicitly the default browser preference on Android

    This way, the default behavior when opening links on Android is
    equivalent as if the browser preference is set to in-app.

I think both these two commits:
d1242a27d content [nfc]: Follow browser preference setting when opening links
f2a8d6a0a setting: Make in-app explicitly the default browser preference on Android

will be clearer to understand if the "explicitly in-app" commit comes first, as I suggested at https://github.com/zulip/zulip-flutter/pull/1400#discussion_r1990290299 .

Let's save making that revision for our call tomorrow. I can demonstrate how I'd go about re-splitting these changes (I think this is a case where that will be easier than trying to treat it as a rebase), and I think it'll be a good example.

PIG208:
Thanks for the help! I have updated the PR.

lib/model/settings.dart
```diff
@@ -24,3 +28,44 @@ enum ThemeSetting {
...
+      //  https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser/near/1169118
+      TargetPlatform.iOS => BrowserPreference.external,
+      _ => BrowserPreference.inApp,
+    };
+  }
+
+  UrlLaunchMode getUrlLaunchMode(Uri url) {
+    return switch (effectiveBrowserPreference) {
+      BrowserPreference.inApp => url.scheme == 'http' || url.scheme == 'https'
+        ? UrlLaunchMode.inAppBrowserView : UrlLaunchMode.platformDefault,
```
gnprice: This condition calls for a comment to explain why we need this (and why it takes the particular shape it does).

Much of the needed information is in the commit message, so just needs to be moved from there. I also had some changes I wanted to make to the framing of that, so I made a revision which I'll push along with this review.

test/model/settings_test.dart
```diff
@@ -0,0 +1,53 @@
...
+  final httpLink = Uri.parse('http://chat.zulip.org');
+  final nonHttpLink = Uri.parse('mailto:chat@zulip.org');
+
+  group('getUrlLaunchMode', () {
+    testAndroidIos('globalSettings.browserPreference is null; use our per-platform defaults for HTTP links', () {
+      final globalStore = eg.globalStore(globalSettings: eg.globalSettings(
+        browserPreference: null));
+      check(globalStore).globalSettings.getUrlLaunchMode(httpLink).equals(
+        defaultTargetPlatform == TargetPlatform.iOS
+          ? UrlLaunchMode.externalApplication : UrlLaunchMode.inAppBrowserView);
```
gnprice, line 21: nit: the conditions in these two tests are confusing because they go opposite ways (so the expected values come in opposite orders) — either way seems fine, just make them consistent

test/model/settings_test.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'store_test.dart';
+
+void main() {
+  final httpLink = Uri.parse('http://chat.zulip.org');
+  final nonHttpLink = Uri.parse('mailto:chat@zulip.org');
+
+  group('getUrlLaunchMode', () {
+    testAndroidIos('globalSettings.browserPreference is null; use our per-platform defaults for HTTP links', () {
+      final globalStore = eg.globalStore(globalSettings: eg.globalSettings(
+        browserPreference: null));
```
gnprice, line 18: nit: this bit of setup should arrive in the "follow setting" commit, rather than the preceding commit where the setting isn't yet involved

assets/l10n/app_en.arb
```diff
@@ -811,6 +811,10 @@
   "@themeSettingSystem": {
     "description": "Label for system theme setting."
   },
+  "openLinksWithInAppBrowser": "Open links with in-app browser",
```
gnprice, line 814: The translation for the label is taken from zulip-mobile. The user
    facing string refers to "in in-app browser" when the actual value is
    `UrlLaunchMode.platformDefault`. This is consistent with the behavior
    the user sees when the link is a HTTP URL on Android and iOS, which
    should cover the majority of use cases of this setting (and probably
    what the user would expect).

This apologetic paragraph can get a lot simpler now :slightly_smiling_face: — I think really can just be cut.

The exact semantics are now "Open links with in-app browser where possible", and the implementation code now makes clear that those are the semantics. And having a "where possible" qualifier be implicit is pretty normal for UI text (outside of UIs for quite technical domains).

gnprice (commented):
Thanks!

This is quite close now — a handful of nits below, following the revision I just pushed which expands some of the explanation in this commit:
f06b14c83 setting [nfc]: Make in-app explicitly the browser preference on Android

lib/model/settings.dart
```diff
@@ -24,3 +28,44 @@ enum ThemeSetting {
     };
   }
 }
+
+/// What browser the user has set to use for opening links in messages.
+///
+/// See https://chat.zulip.org/#narrow/stream/48-mobile/topic/in-app.20browser
```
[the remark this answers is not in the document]
PIG208: Removed the link because I think the dartdoc on the individual enum options explain them well enough, and that we have links to this conversation from `effectiveBrowserPreference`.

This line was originally from https://github.com/zulip/zulip-mobile/pull/4679#discussion_r638422880, where the code that interprets the enum values live elsewhere.

lib/model/settings.dart
PIG208: Ah I see! It makes sense that the upstream behavior needs to change.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Looks good; merging.

PR #1403, opened 2025-03-12
test [nfc]: Make globalSettings optional on TestGlobalStore

gnprice:
This has a natural default, at least in tests, because most tests don't care about the specific value of the global settings.

(By contrast most tests that interact with the store at all will care what set of accounts exist, so there isn't as natural a default for `accounts`.)

Originally noticed this here:
  https://github.com/zulip/zulip-flutter/pull/1386#discussion_r1990358363

### DIFF
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -885,16 +885,12 @@ GlobalSettingsData globalSettings({
     themeSetting: themeSetting,
   );
 }
-const _globalSettings = globalSettings;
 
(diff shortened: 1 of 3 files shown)


gnprice:
(FYI @PIG208 since we just introduced this in https://github.com/zulip/zulip-flutter/pull/1167.)

chrisbobbe:
Thanks, LGTM! Merging.

PR #1408, opened 2025-03-13
KaTeX (1/n): Initial support for displaying basic KaTeX content

rajveermalviya:
This initial implementation include:

 - Displaying the characters in their corresponding
   custom text styles (fonts, font weight, font style).

 - Character and symbol sizing.

 - And some subset of inline styles.

This results in support for displaying some simple KaTeX functions.

Related: #46


lib/model/katex.dart
```diff
@@ -0,0 +1,409 @@
...
+            if (marginRightEm != null) continue;
+
+          case 'padding-left':
+            paddingLeftEm = _getEm(expression);
+            if (paddingLeftEm != null) continue;
+
+          default:
+            // TODO handle more CSS properties
+            assert(debugLog('Unsupported CSS property: $property of type ${expression.runtimeType}'));
+        }
```
gnprice, line 279: It seems like this code has three levels of how well it supports a given KaTeX blob:
* The HTML doesn't follow any structure we've yet anticipated. The parser in this file throws KatexHtmlParseError, producing `nodes` of null.
* The HTML contains something we know we don't yet support, but there's no exception; `nodes` ends up with some kind of parse result.
* The HTML consists only of constructs we believe are fully supported — meaning we believe the current version of the code will produce widgets that make the intended math show up exactly the same as with KaTeX on the web.

This `default` case is an example of that middle state.

The middle state is very useful for development: you want to let the parser do its thing as best it currently can, and see how the resulting widgets look, while you work on supporting that case.

But I'd also really like to be able to run the parser in a mode where it will only accept constructs we believe are fully supported. That way we can run it on a corpus of public messages collected by the scripts in `tools/content/`, and get a survey of what remains to be implemented. That mode could also be useful for turning KaTeX support on by default, but only for expressions we believe will show up exactly right, and for other expressions falling back to the raw TeX like we do in main today.

I think the most effective way to draw that boundary will be to do so starting in this first PR, so that when we include a given construct on the "fully supported" side of it we do so at the same time as we're implementing support for that construct and thinking about it in detail.

Maybe have two experimental flags?
* One flag enables showing KaTeX at all. Things that are fully supported get rendered; things that are incompletely supported, just like those not supported, get the fallback.
* A second flag enables showing KaTeX even where it's incomplete — so the behavior this revision has when `renderKatex` is true. (When the first flag is false, this one would just do nothing.)

Then when support is mostly complete, we might turn the first flag into always-true, while keeping the second flag as experimental (and default-false).

gnprice (commented):
Exciting!

Generally this structure looks good. Here's a high-level review — so just a handful of comments mostly about things that I think will help make the changes clearer to understand.

Then once these aspects look good (should be quick, I think), we'll do maintainer review as usual.

lib/model/katex.dart
rajveermalviya: OK, added another flag (`forceRenderKatex`) to control if the KaTeX content should be rendered even if there were some errors encountered while parsing the HTML. Currently, the "ignorable errors" happen either when the parser encounters an unknown CSS class or an unknown CSS property in the inline styles.

rajveermalviya:

gnprice (commented):
Thanks for the revision!

This resolves the high-level feedback I had — @PIG208, over to you for maintainer review.

PIG208 (commented):
Thanks for working on this! I just went through the entire PR and played with this a bit on my device. It looks pretty good!

With regards with testing, I'm not sure what the current plan is. It doesn't seem quite useful to have tests structured basically a duplicate of all the KaTeX classes we support. Maybe it would be better to test with examples crawled online?

gnprice:
> With regards with testing

Yeah, good question. It looks like the current version of the PR doesn't add tests for most of the functionality. Let's at a minimum have tests that cover the interesting logic like for `sizing` and `delimsizing`.

> It doesn't seem quite useful to have tests structured basically a duplicate of all the KaTeX classes we support. Maybe it would be better to test with examples crawled online?

I think these are two complementary types of tests which are both useful:
* In the test suite, we have examples that are simplified as much as possible — to make them clear to understand, and easy to update for refactorings when needed — while still being complex enough to exercise the desired logic.
* Outside the test suite, but at times like when we're doing major development on this area, we systematically test a corpus of public real-world examples. When that finds things we'd previously missed, we take those examples and reduce them into new test cases to add to the test suite.

We'll definitely want to be doing the second kind once we're a couple of PRs in to this current effort and feel that we've covered the majority of usage — it'll help us prioritize which remaining areas to do next.

lib/model/katex.dart
```diff
@@ -0,0 +1,481 @@
...
+      );
+    }
+    return null;
+  }
+
+  double? _getEm(css_visitor.Expression expression) {
+    if (expression is css_visitor.EmTerm && expression.value is num) {
+      return (expression.value as num).toDouble();
+    }
+    return null;
```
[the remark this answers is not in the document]
rajveermalviya: We only support `em` values currently because, they are the only ones observed in the Katex HTML currently, but if we encounter any other than this will return `null` and we handle that later in `switch` block. Error-ing if we get an unexpected type (non-em).

lib/widgets/content.dart
```diff
@@ -804,11 +806,152 @@ class MathBlock extends StatelessWidget {
...
+        KatexSpanTextAlign.left => TextAlign.left,
+        KatexSpanTextAlign.center => TextAlign.center,
+        KatexSpanTextAlign.right => TextAlign.right,
+      };
+    }
+
+    var margin = EdgeInsets.zero;
+    final marginRightEm = styles.marginRightEm;
+    final marginLeftEm = styles.marginLeftEm;
+    if (marginRightEm != null && !marginRightEm.isNegative) {
```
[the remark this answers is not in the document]
rajveermalviya: Negative margin require special handling they are implemented in the next PR, anyways, removed handling for `margin` altogether.

rajveermalviya:
Thanks for the review @PIG208! Pushed an update, PTAL.

PIG208 (commented):
Thanks for the revision! I did another pass and left some small comments.

lib/model/katex.dart
```diff
@@ -0,0 +1,563 @@
...
+        for (final declaration in rule.declarationGroup.declarations) {
+          if (declaration case css_visitor.Declaration(
+            :final property,
+            expression: css_visitor.Expressions(
+              expressions: [css_visitor.Expression() && final expression]),
+          )) {
+            switch (property) {
+              case 'height':
+                heightEm = _getEm(expression);
+                if (heightEm != null) continue;
```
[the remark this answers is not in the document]
rajveermalviya: Sorry, I should've replied to the above comment how I addressed it in the revision.

I had addressed that bug by moving the `_logError` call outside of the `switch` block. If `_getEm` returns `null` then the condition here will be `false`, in which case it would fall down to `_logError` where the error will be logged – reporting we got something other than `em` here.

rajveermalviya:
Thanks for the review @PIG208! Pushed an update, PTAL.

lib/model/katex.dart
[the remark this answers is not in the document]
PIG208: Ah that makes sense. Thanks.

PIG208:
It's a bit puzzling to me that the first parenthesis (span with class "mopen" in HTML) appears differently in Flutter than on web.

| web | Flutter |
| - | - |
| Screenshot from 2025-04-10 15-28-13 | Screenshot from 2025-04-10 15-28-03 |

````

```math
( \big( \Big( \bigg( \Bigg(
```

```math
( (((( ( ( ( ( (
```

````

**Comment by PIG208** [comment]
Thanks for the revision! I just tried this out locally again. Marking this for Greg's review. Left a question on the font style.

File: test/model/content_test.dart [hunk 5]
```diff
@@ -615,6 +699,233 @@ class ContentExample {
       ]),
     ]);
 
+  // The font sizes can be compared using the katex.css generated
+  // from katex.scss :
+  //   https://unpkg.com/katex@0.16.21/dist/katex.css
+  static final mathBlockKatexSizing = ContentExample(
+    'math block; KaTeX different sizing',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2147558
+    "```math\n\\tiny {\\lambda \\Huge \\lambda}\n\n\\Huge \\lambda\n\\huge \\lambda\n\\LARGE \\lambda\n\\Large \\lambda\n\\large \\lambda\n\\normalsize \\lambda\n\\small \\lambda\n\\footnotesize \\lambda\n\\scriptsize \\lambda\n\\tiny \\lambda\n```",
```
gnprice, line 708: Instead of `\lambda` here, how about just `x`? That'd make this shorter and easier to read, given how many copies of it this has.

Some other test cases say `\lambda` because that's a way of making the rendered output have different text from the input, so that it's easy to check which is being displayed. But I think those test cases cover that aspect, so this one doesn't need that wrinkle.

lib/model/katex.dart
```diff
@@ -0,0 +1,576 @@
...
+
+        case 'sizing':
+        case 'fontsize-ensurer':
+          // .sizing,
+          // .fontsize-ensurer { ... }
+          if (index + 2 < spanClass.length) {
+            final resetSizeClass = spanClasses[index + 1];
+            final sizeClass = spanClasses[index + 2];
+
+            final resetSizeClassSuffix =_resetSizeClassRegExp.firstMatch(resetSizeClass)?.group(1);
```
gnprice, line 314: nit:
```suggestion
            final resetSizeClassSuffix = _resetSizeClassRegExp.firstMatch(resetSizeClass)?.group(1);
```

lib/model/katex.dart
```diff
@@ -0,0 +1,576 @@
...
+              if (resetSizeIdx <= sizes.length && sizeIdx <= sizes.length) {
+                styles.fontSizeEm = sizes[sizeIdx - 1] / sizes[resetSizeIdx - 1];
+                index += 3;
+                continue;
+              }
+            }
+          }
+
+          // Should be unreachable.
+          throw KatexHtmlParseError();
```
gnprice, line 333: This totally will be reached if the input we get is malformed, right? (Or just from some future version of KaTeX with surprise changes to the classes it generates.)

So I wouldn't call that "unreachable" — I'd reserve that for where the logic of the code itself should mean the case can't ever be reached, regardless of input.

lib/model/katex.dart
```diff
@@ -0,0 +1,576 @@
...
+}
+
+class KatexSpanStyles {
+  double? heightEm;
+  double? verticalAlignEm;
+
+  String? fontFamily;
+  double? fontSizeEm;
+  KatexSpanFontStyle? fontStyle;
+  KatexSpanFontWeight? fontWeight;
```
gnprice, line 494: nit: put these in same order as the enums are defined

(And I think weight, then style, is the preferred way around: one says "bold italic", e.g. in the names of the fonts this PR adds, not "italic bold")

lib/model/katex.dart
```diff
@@ -0,0 +1,576 @@
...
+      other.fontStyle == fontStyle &&
+      other.fontWeight == fontWeight &&
+      other.textAlign == textAlign;
+  }
+
+  static final _zero = KatexSpanStyles();
+
+  @override
+  String toString() {
+    if (this == _zero) return '${objectRuntimeType(this, 'KatexSpanStyles')}()';
```
gnprice, line 535: This is just an optimization, right?

Can leave it out in favor of simplifying the code — `toString` on these should only come up in errors or debugging.

lib/widgets/content.dart
```diff
@@ -804,11 +805,134 @@ class MathBlock extends StatelessWidget {
...
+    }
+
+    final styles = span.styles;
+    TextStyle? textStyle;
+    TextAlign? textAlign;
+
+    if (styles.fontFamily != null) {
+      textStyle ??= TextStyle();
+      textStyle = textStyle.copyWith(fontFamily: styles.fontFamily);
+    }
```
gnprice, line 892: How about:
```suggestion
    final fontFamily = styles.fontFamily;
```

and then construct `textStyle` and `textAlign` at the end after defining several locals like that?

I think that would be simpler. It'd also avoid copying a `TextStyle` repeatedly when there are a couple of different properties being set.

lib/widgets/content.dart
```diff
@@ -804,11 +805,134 @@ class MathBlock extends StatelessWidget {
...
+}
+
+class _KatexSpan extends StatelessWidget {
+  const _KatexSpan(this.span);
+
+  final KatexNode span;
+
+  @override
+  Widget build(BuildContext context) {
+    final em = DefaultTextStyle.of(context).style.fontSize!;
```
gnprice, line 871: This lookup is only used in one condition, so can move inside that condition. That way we don't do the lookup when it doesn't end up being needed.

test/model/content_test.dart
```diff
@@ -615,6 +699,233 @@ class ContentExample {
...
+            '<span class="mord mathnormal sizing reset-size6 size8">λ</span>'
+            '<span class="mord mathnormal sizing reset-size6 size7">λ</span>'
+            '<span class="mord mathnormal sizing reset-size6 size6">λ</span>'
+            '<span class="mord mathnormal sizing reset-size6 size5">λ</span>'
+            '<span class="mord mathnormal sizing reset-size6 size4">λ</span>'
+            '<span class="mord mathnormal sizing reset-size6 size3">λ</span>'
+            '<span class="mord mathnormal sizing reset-size6 size1">λ</span></span></span></span></span></p>',
+    [
+      MathBlockNode(
+        texSource: "\\tiny {\\lambda \\Huge \\lambda}",
```
gnprice, line 737: Can you write a widget test for this?

I think that'll actually be more pertinent than the parsing test (though we'll naturally have this parsing test as a consequence given the way the ContentExample system is structured, and that's fine): the important thing isn't that one layer says "0.5em" and then a layer inside it says "4.976em" so much as it is that the text winds up at the size that would be 2.488em if found at the root of the math tree.

For the widget test it may be helpful to have this example split into two examples: one for each of the two blocks in this example. That way the widget test can render only the one that it intends to inspect.

test/model/content_test.dart
gnprice: Or instead of "x x x x" could be like "a b c d" etc., if that helps make things convenient for inspecting in a widget test.

gnprice (commented):
Thanks @rajveermalviya for building this, and @PIG208 for the previous reviews!

Exciting to have this feature underway. Here's a review of the first 7/8 commits:
f7a4721c8 deps: Add csslib as a direct dependency
b4b38d75b fonts: Add KaTeX custom fonts
0106b95f2 content [nfc]: Alias package:intl import as `intl`
671fa127f content [nfc]: Use Dart patterns for parsing math content
1fca68613 binding: Add getGlobalStoreSync method
760985214 content: Add KaTeX spans parser, initial rendering; w/o styles
c273b0aa1 content: Support basic text styles for KaTeX content

so leaving one more for a later round:
0a8c846f5 content: Support parsing and handling inline styles for KaTeX content

(I see Zixuan also has one open comment just above.)

lib/model/katex.dart
```diff
@@ -0,0 +1,481 @@
...
+    // TODO maybe check if the sequence of ancestors matter for spans.
+
+    final spanClasses = List<String>.unmodifiable(element.className.split(' '));
+
+    // Aggregate the CSS styles that apply, in the same order as the CSS
+    // classes specified for this span, mimicking the behaviour on web.
+    //
+    // Each case in the switch blocks below is a separate CSS class definition
+    // in the same order as in katex.scss :
+    //   https://github.com/KaTeX/KaTeX/blob/2fe1941b7e6c0603680ef6edd799bd8a8b46871a/src/styles/katex.scss
```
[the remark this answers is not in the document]
gnprice: (FTR the length I normally use is 9 hex digits: https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#mentioning-commits

> **Use 9 (or 10) hex digits to identify a commit.** To help make this convenient, you can tell Git to routinely print 9-digit abbreviations for you by setting `git config core.abbrev 9`.
>
> Rationale: The full 40 hex digits is a lot, and generally doesn't flow well in prose. On the other hand 7 hex digits, which is what GitHub shows in its UI whenever it doesn't show 40, is short enough that there's a material risk of collisions: in zulip.git there are already a handful of commits that are ambiguous when identified by just 7 digits, and in a very large project like the Linux kernel such collisions can become routine.
>
> A 9-digit abbreviation is still short enough to fit well in running text; and it's long enough that it's extremely unlikely any given such abbreviation will ever be ambiguous.

For KaTeX 8 is fine and probably 7 would be fine, because it's a smaller repo; but I just use 9 everywhere to simplify.)

lib/model/content.dart
```diff
@@ -341,22 +342,56 @@ class CodeBlockSpanNode extends ContentNode {
...
+    required this.text,
+    required this.nodes,
+    super.debugHtmlNode,
+  }) : assert((text != null) ^ (nodes != null));
+
+  final KatexSpanStyles styles;
+
+  /// The text or a single character this KaTeX node contains, generally
+  /// observed to be the leaf node in the KaTeX HTML tree.
+  /// It will be null if [nodes] is non-null.
```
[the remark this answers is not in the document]
rajveermalviya: No, if `text` is non-null then it will always mean that that node is a leaf node, removed the unnecessary sentence.

rajveermalviya:
> first parenthesis (span with class "mopen" in HTML) appears differently in Flutter than on web.

@PIG208, that turned out to be a weird bug in upstream Flutter when displaying text with `KaTeX_Math` font with `FontStyle.italic`.

For now, I've added a workaround that should fix this behavior (forcing `FontStyle.normal` for this font on Android), and also reported the issue upstream — https://github.com/flutter/flutter/issues/167474.

rajveermalviya:

gnprice (commented):
Thanks for the revision! Just a couple of nits below, on those same first 7/8 commits.

Then let's save the last commit:
864490350 content: Support parsing and handling inline styles for KaTeX content

for the next PR. That way we can go ahead and merge the rest of them very soon.

lib/model/katex.dart
```diff
@@ -0,0 +1,566 @@
...
+        // TODO handle skipped class declarations between .mainrm and
+        //   .sizing .
+
+        case 'sizing':
+        case 'fontsize-ensurer':
+          // .sizing,
+          // .fontsize-ensurer { ... }
+          if (index + 2 < spanClass.length) {
+            final resetSizeClass = spanClasses[index + 1];
+            final sizeClass = spanClasses[index + 2];
```
gnprice, line 313: Oh, one bug here:
```suggestion
          if (index + 2 < spanClasses.length) {
            final resetSizeClass = spanClasses[index + 1];
            final sizeClass = spanClasses[index + 2];
```

:slightly_smiling_face:

lib/widgets/content.dart
```diff
@@ -804,11 +806,148 @@ class MathBlock extends StatelessWidget {
...
+    }
+
+    return _Katex(inline: false, nodes: nodes);
+  }
+}
+
+// Base text style from .katex class in katex.scss :
+//   https://github.com/KaTeX/KaTeX/blob/613c3da8/src/styles/katex.scss#L13-L15
+const kBaseKatexTextStyle = TextStyle(
+  fontSize: kBaseFontSize * 1.21,
```
gnprice, line 826: A separate bug: this doesn't correctly handle scaling the font size up when the TeX appears inside a heading. For example: [#test here > math @ 💬](https://chat.zulip.org/#narrow/channel/7-test-here/topic/math/near/2159072)

To do that, this factor of 1.21 should instead get applied in the same way as the scaling factor for the font size of code spans. See references to `kInlineCodeFontSizeFactor`.

Let's make that fix as a separate commit, though, which can go in a follow-up PR. That way we can still merge this PR's changes very soon, as a baseline for further progress.

rajveermalviya: Added a fix for this in ffcdfb0c83c8a8083bc827152a93141d6ed7bce9, in #1452.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

test/widgets/content_test.dart
```diff
@@ -553,7 +554,114 @@ void main() {
...
+      checkKatexText(tester, '1',
+        fontFamily: 'KaTeX_Main',
+        fontSize: fontSize,
+        fontHeight: kBaseKatexTextStyle.height!);
+
+      fontSize = 4.976 * fontSize;
+      checkKatexText(tester, '2',
+        fontFamily: 'KaTeX_Main',
+        fontSize: fontSize,
+        fontHeight: kBaseKatexTextStyle.height!);
```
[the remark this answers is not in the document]
gnprice: > There are a couple of other refactors I'm thinking I'll try doing upon merging this PR

Sent most of those just now as #1478.

Others are still a draft, as I mentioned there:
> This doesn't include the changes I've sketched for passing the font size down directly through the recursion in the parser and widgets. I'll want to look a bit more at how those would interact with the changes in https://github.com/zulip/zulip-flutter/pull/1452 (which introduce a number of additional length-valued properties) before I'm confident just what shape that part of the API should take.

and I haven't yet tried revising these tests. (That would naturally go along with some of the changes still in draft.)

Issue #462, opened 2023-12-25
Support Apple auth


gnprice:
AKA "Sign in with Apple". This will be a requirement for launch, because Apple makes it a requirement for the App Store. (That doesn't affect the beta, which is in TestFlight and not in the App Store.)

Upstream docs here:
https://developer.apple.com/sign-in-with-apple/
but I hope we can complete this feature without (re-)reading those in any detail — instead we can crib from the work we did back in 2020, when we read up on this closely in order to implement it the first time.

~~I doubt any significant number of people are actually using this feature, so there's probably no need for it in the beta period. But we'll see if we hear feedback asking for it.~~ Indeed I don't think we've heard from anyone wanting to use it. But now that launch time is getting close, we'll need to do this sometime soon.

### Implementation

Because this issue is closely tied to our interaction with Apple, it will need to be done by a member of the core team.

There are two UX flows for Sign in with Apple:
* a web-based flow (this issue);
* a native iOS flow: #1442.

For this issue, we'll concentrate on the web-based flow, which is available on iOS and on Android. The UX this provides is the same kind of UX as when people log in via GitHub, or Google, or any other form of SSO.

I believe the web flow may actually already work, as part of our support for web-based auth in general. In that case the key task remaining for this issue is to test the web flow manually to confirm that it works smoothly.

### Testing

An important part of the work on this issue will be to test it end to end. The only practical way to do that is manually: fire up the app with your draft implementation of the feature, and try using it.

To test the implementation that's in the published beta app, try running the app and logging in on chat.zulip.org.

For instructions on how to test Sign in with Apple in a development build of the app, see zulip/zulip-mobile@c19cbe2f2c0bda9cbef89a9a4b3435c297f74b41, or perhaps more readably:
https://github.com/zulip/zulip-mobile/blob/main/docs/howto/ios-tips.md#sign-in-with-apple
See also this 2025 thread which has some updates to those instructions:
[#mobile-dev-help > Enable "Sign in with Apple" for Flutter App](https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/Enable.20.22Sign.20in.20with.20Apple.22.20for.20Flutter.20App)

### Related issues

* From when we originally implemented this feature in the server and the legacy mobile app, in 2020 when Apple's mandate of it took effect:
  https://github.com/zulip/zulip/issues/14168
  https://github.com/zulip/zulip-mobile/issues/3964
[...]

PR #1410, opened 2025-03-15
api: Don't allow connecting to servers <4.0

chrisbobbe:
From the README:

https://github.com/zulip/zulip-flutter?tab=readme-ov-file#server-compatibility
> We support Zulip Server 4.0 and later.

This implementation isn't ideal because it logs you out, instead of
just taking down the account's UI as we discussed:
  https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/.23F267.20Disallow.20connecting.20to.20unsupported.20ancient.20servers/near/2104182
But it was simpler this way since programmatically logging out is
already an action we handle.

Fixes: #267

### Screenshot (with `kMinSupportedZulipFeatureLevel` fudged to 9999):

image


lib/widgets/content.dart
```diff
@@ -1414,20 +1412,13 @@ class MessageTableCell extends StatelessWidget {
...
-  }
-
   final store = PerAccountStoreWidget.of(context);
   final url = store.tryResolveUrl(urlString);
   if (url == null) { // TODO(log)
-    showError(context, null);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    showErrorDialog(context: context,
+      title: zulipLocalizations.errorCouldNotOpenLinkTitle,
+      message: zulipLocalizations.errorCouldNotOpenLink(urlString));
```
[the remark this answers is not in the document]
PIG208: While we do have a test "error dialog if invalid link" that covers a similar error in `PlatformActions.launchUrl`, I saw that this branch does not have test coverage.

We should probably move that test to the corresponding file for `PlatformActions`.

lib/widgets/login.dart
```diff
@@ -343,6 +343,9 @@ class _LoginPageState extends State<LoginPage> {
       // Could set [_inProgress]… but we'd need to unset it if the web-auth
       // attempt is aborted (by the user closing the browser, for example),
       // and I don't think we can reliably know when that happens.
+
+      // Not using [PlatformActions.launchUrl] because web auth needs special
+      // error handling.
```
PIG208, line 348: Perhaps as a part of #462, we can get rid of the block of iOS specific error handling code and reuse `PlatofmActions.launchUrl`, but I'm hesitant because of the question mark from `// TODO(#462) remove this?`.

lib/model/store.dart
```diff
@@ -1452,6 +1489,48 @@ class UpdateMachine {
...
+    required this.zulipFeatureLevel,
+  });
+
+  factory ZulipVersionData.fromInitialSnapshot(InitialSnapshot initialSnapshot) =>
+    ZulipVersionData(
+      zulipVersion: initialSnapshot.zulipVersion,
+      zulipMergeBase: initialSnapshot.zulipMergeBase,
+      zulipFeatureLevel: initialSnapshot.zulipFeatureLevel);
+
+  /// A [ZulipVersionData] from a [MalformedServerResponseException],
```
PIG208, line 1507: I think it might be better to start this with a verb.

test/model/store_test.dart
```diff
@@ -1127,6 +1210,46 @@ void main() {
...
+      async.elapse(const Duration(seconds: 1));
+      check(globalStore.takeDoRemoveAccountCalls()).single.equals(eg.selfAccount.id);
+
+      async.elapse(TestGlobalStore.removeAccountDuration);
+      check(globalStore.perAccountSync(eg.selfAccount.id)).isNull();
+
+      async.flushTimers();
+      // Reload never succeeds and there are no unhandled errors.
+      check(globalStore.perAccountSync(eg.selfAccount.id)).isNull();
+    }));
```
PIG208, line 1252: Other than these tests, I think it would be valuable to have an end-to-end widget test that checks the error dialog when connecting to an ancient server, but I feel like it can be difficult to set up.

PIG208 (commented):
Thanks for the PR! I went through all the commits and left some comments. This will need a rebase because of some conflicts with #1400.

lib/widgets/content.dart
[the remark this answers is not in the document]
chrisbobbe: > nit: While the main motivation from [fe3538b](https://github.com/zulip/zulip-flutter/commit/fe3538bcb448803951a4d85e26e01aa6578645ff) does apply here, would it make sense to use `url.toString` here to be consistent?

No, not here :) because this is in an `if (url == null) {}`, and `url.toString()` would always be "null".

lib/widgets/login.dart
chrisbobbe: Yeah; that sounds like it would need investigating when we get to #462.

test/model/store_test.dart
chrisbobbe: I agree. We'd need change `TestZulipBinding` so that it could work with a `UpdateMachineTestGlobalStore`, and I'm not sure how best to do that.

The `TestZulipBinding.globalStore` getter has a `TestGlobalStore` return type, and `UpdateMachineTestGlobalStore` doesn't extend `TestGlobalStore`. Both classes extend `GlobalStore`, but that doesn't make sense as a return type; it would break a lot of callers that use things like `TestGlobalStore.add`.

chrisbobbe:

chrisbobbe:
(Just rebased past #1416.)

test/model/binding.dart
```diff
@@ -161,9 +161,18 @@ class TestZulipBinding extends ZulipBinding {
...
+  ///   * [launchUrlException]
+  ///   * [takeLaunchUrlCalls]
   bool launchUrlResult = true;
 
+  /// The [PlatformException] that `ZulipBinding.instance.launchUrl()` should throw.
+  ///
+  /// See also:
+  ///   * [launchUrlResult]
+  ///   * [takeLaunchUrlCalls]
+  PlatformException? launchUrlException;
```
PIG208, line 174: Looks like this should be reset in `_resetLaunchUrl`.

lib/api/core.dart
```diff
@@ -10,6 +10,24 @@ import '../model/binding.dart';
 import '../model/localizations.dart';
 import 'exception.dart';
 
+/// The Zulip Server version below which we should refuse to connect.
+///
+/// When updating this, also update [kMinSupportedZulipFeatureLevel]
+/// and the README.
+const kMinSupportedZulipVersion = '4.0';
+
+/// The Zulip feature level reserved for the [minSupportedZulipVersion] release.
```
PIG208, line 19: nit: the only `minSupportedZulipVersion` I can find is the parameter of `errorServerVersionUnsupportedMessage`, but it is not a defined symbol within the dartdoc's context; I think this is supposed to be the "k"-prefixed constant?

lib/widgets/actions.dart
```diff
@@ -239,3 +241,70 @@ abstract final class ZulipAction {
...
+      _                              => true,
+    };
+    if (shouldShowSnackbar) {
+      ScaffoldMessenger.of(context).showSnackBar(
+        SnackBar(behavior: SnackBarBehavior.floating, content: successContent));
+    }
+  }
+
+  /// Opens a URL with [ZulipBinding.launchUrl], with an error dialog on failure.
+  // TODO widget tests
```
PIG208, line 286: Now that we have a dedicated test group, we should be able to remove the TODO, unless there is something else that needs tests here?

PIG208 (commented):
Thanks for the update! I read the new commit and have some small comments. Marking this for Greg's review.

lib/api/core.dart
chrisbobbe: Indeed, thanks for the catch!

lib/widgets/actions.dart
chrisbobbe: Indeed!

chrisbobbe:

lib/model/store.dart
```diff
@@ -993,22 +1009,34 @@ class UpdateMachine {
...
     final stopwatch = Stopwatch()..start();
-    final initialSnapshot = await _registerQueueWithRetry(connection,
-      stopAndThrowIfNoAccount: stopAndThrowIfNoAccount);
+    InitialSnapshot? initialSnapshot;
+    try {
+      initialSnapshot = await _registerQueueWithRetry(connection,
+        stopAndThrowIfNoAccount: stopAndThrowIfNoAccount);
+    } on _ServerVersionUnsupportedException catch (e) {
+      connection.close();
+      await updateZulipVersionData(e.data);
```
gnprice, line 1032: This will mutate `connection` after closing it. Probably doesn't throw, but incorrect in principle.

With the suggestion in the previous comment, this can inline `updateZulipVersionData`, and just skip the line that updates `connection`.

Maybe further clean up by having a method on `GlobalStore`:
```dart
if (!zulipVersionData.matchesAccount(globalStore.getAccount(accountId)!)) {
  await globalStore.updateZulipVersionData(accountId, zulipVersionData);
}
```

lib/model/store.dart
```diff
@@ -1477,6 +1514,48 @@ class UpdateMachine {
...
+
+  /// Make a [ZulipVersionData] from a [MalformedServerResponseException],
+  /// if the body was readable/valid JSON and contained the data, else null.
+  static ZulipVersionData? fromMalformedServerResponseException(MalformedServerResponseException e) {
+    try {
+      final data = e.data!;
+      return ZulipVersionData(
+        zulipVersion: data['zulip_version'] as String,
+        zulipMergeBase: data['zulip_merge_base'] as String?,
+        zulipFeatureLevel: data['zulip_feature_level'] as int);
```
gnprice, line 1540: If there's a `zulip_version` but no `zulip_feature_level`, I think we can infer it's indeed Zulip just ancient — so the feature level is zero. (I.e. it's from before Zulip 3.0 introduced feature levels.)

gnprice (commented):
Thanks! I've read most of the branch:
e13084a56 store test [nfc]: Replace old method name in a test group
8228d97fc store [nfc]: Move some declarations up out of a switch case
ebe5f7094 store [nfc]: Re-query account after _registerQueueWithRetry async gap
2983cdb34 store [nfc]: Helper code for UpdateMachine.load's update-feature-level
413bf1b62 store [nfc]: Stop using mutable Account variable in UpdateMachine.load
fb73e4169 content [nfc]: Inline showError helper in _launchUrl
50379e2e2 actions test [nfc]: Put ZulipActions tests in new 'ZulipActions' group
b03c30aa5 actions [nfc]: Add PlatformActions, alongside ZulipActions
ae8a3f757 content: Use `url.toString()` instead of `urlString` in error message
eb9883f36 actions [nfc]: Move open-link logic here, from lib/widgets/content

and part of the final/main commit:
2468db2d2 api: Don't allow connecting to servers <4.0

That leaves:
367a165b6 actions test: Cover PlatformActions.launchUrl
ad292ca25 dialog: Add "Learn more"-button param to showErrorDialog
and the rest of the final commit.

chrisbobbe:

test/widgets/actions_test.dart
```diff
@@ -1,341 +1,510 @@
...
+                await PlatformActions.launchUrl(context, url);
+              },
+              child: const Text('link'))))));
+        await tester.pump();
+        await tester.tap(find.text('link'));
+        await tester.pump(Duration.zero);
+      }
+
+      final httpUrl = Uri.parse('http://chat.zulip.org');
+      final nonHttpUrl = Uri.parse('mailto:chat@zulip.org');
```
gnprice, line 435: nit:
```suggestion
      final httpUrl = Uri.parse('https://chat.example');
      final nonHttpUrl = Uri.parse('mailto:chat@example');
```

* use reserved example domains (RFC 6761; https://en.wikipedia.org/wiki/Special-use_domain_name)
* use `https:` rather than `http:`, since `https:` should be the norm and cleartext `http:` rare

test/widgets/dialog_test.dart
```diff
@@ -0,0 +1,29 @@
...
+
+  group('showErrorDialog', () {
+    testWidgets('tap "Learn more" button', (tester) async {
+      addTearDown(testBinding.reset);
+      await tester.pumpWidget(TestZulipApp());
+      await tester.pump();
+      final element = tester.element(find.byType(Placeholder));
+
+      showErrorDialog(context: element, title: 'hello',
+        learnMoreButtonUrl: Uri.parse('foo.example'));
```
gnprice, line 21: nit: make it a proper absolute URL, for realism
```suggestion
        learnMoreButtonUrl: Uri.parse('https://foo.example'));
```

(if it isn't, launching is unlikely to work)

lib/model/store.dart
```diff
@@ -994,20 +1021,27 @@ class UpdateMachine {
...
-      stopAndThrowIfNoAccount: stopAndThrowIfNoAccount);
+    InitialSnapshot? initialSnapshot;
+    try {
+      initialSnapshot = await _registerQueueWithRetry(connection,
+        stopAndThrowIfNoAccount: stopAndThrowIfNoAccount);
+    } on _ServerVersionUnsupportedException catch (e) {
+      // `!` is OK because _registerQueueWithRetry would have thrown a
+      // not-_ServerVersionUnsupportedException if no account
+      if (!e.data.matchesAccount(globalStore.getAccount(accountId)!)) {
+        await globalStore.updateZulipVersionData(accountId, e.data);
```
gnprice, line 1032: I guess since the caller is going to log the account out anyway, this update isn't really needed. (Realized that when reading tests and seeing there doesn't seem to be a test exercising this.)

Having already done the work to write it, though, may as well keep it. It's more correct in principle, and will simplify if in the future we have it not log out in this case.

lib/api/core.dart
```diff
@@ -10,6 +10,24 @@ import '../model/binding.dart';
...
+///
+/// For this value, see the API changelog:
+///   https://zulip.com/api/changelog
+const kMinSupportedZulipFeatureLevel = 65;
+
+/// The doc stating our oldest supported server version.
+// TODO: Instead, link to new Help Center doc once we have it:
+//   https://github.com/zulip/zulip/issues/23842
+final kServerSupportDocUrl = Uri.parse(
+  'https://zulip.readthedocs.io/en/latest/overview/release-lifecycle.html#compatibility-and-upgrading');
```
gnprice, line 29:
```suggestion
  'https://zulip.readthedocs.io/en/latest/overview/release-lifecycle.html#client-apps');
```

The other URL's fragment doesn't seem to exist on that page now.

(It's fine, the page still loads, but the user doesn't get taken to any particular spot on it.)

gnprice (commented):
Thanks! Looks good, and I've now read the whole thing. Just a few small comments below.

chrisbobbe:

gnprice:
Thanks! Looks good; merging. Glad to have this complete. :slightly_smiling_face:

PR #1411, opened 2025-03-15
Fixed a google sign-in issue

SandaruEARL:
Screenshot_20250315_224042
Fix Google Sign-In "Open With" Pop-Up Issue in _beginWebAuth

This PR fixes a misbehavior in the _beginWebAuth method in login.dart where clicking on a Google account during sign-in would trigger an "Open With" dialog, prompting users to choose an email app (Gmail, Outlook, etc.) instead of proceeding with authentication.

Issue: On the first click, the "Open With" menu appears, causing confusion. However, on the second click, authentication proceeds correctly.
Fix: The authentication URL now always opens in an external browser (LaunchMode.externalApplication), preventing the unwanted pop-up and ensuring a smooth login experience.
Impact: Users can now sign in with Google without being interrupted by the "Open With" pop-up.


gnprice:
Thanks for the report. Would you please open an issue for this, and describe in more detail exactly what you did that led up to getting this unexpected prompt?

I just tried to reproduce the issue, and it didn't reproduce — the flow worked normally as expected, with no prompt for an email app. A detailed set of steps will be helpful for making it possible to reproduce what you're seeing.

After we better understand the issue, we can determine the right fix. A priori, I don't see why changing from having the browser appear in the app to externally would have any effect on this — and it would make the UX more cumbersome for everyone who's not running into this issue.

SandaruEARL:
What is the point when someone uses google sign-in and select an account and instead not directly login to the app it pops a "open with" dialog box to open that select ed email to open in these apps, gmail/outlook, i think its not good for the UX..
________________________________

SandaruEARL:
Made a issue in github, could you refer it...

-- regards --
________________________________

gnprice:
For cross-reference, it looks like this became the following issue:
* #1412

Issue #1409, opened 2025-03-14
Settings for experimental features


gnprice:
It would be useful to have feature flags to control enabling features that aren't yet complete. This is a common technique for making it possible to merge PRs that make partial progress toward a complex multi-PR feature: those working on the feature can have the feature flag enabled in order to see the current incomplete behavior, while for everyone else it remains disabled and so (barring bugs in the use of the flag itself) has no effect.

Specifically I think a good design for this for the Zulip mobile app would be to have a collection of local boolean settings on the device, with UI for them in a subsidiary settings screen, labeled like "Experimental features". Plus probably some text at the top of that screen stressing that these are half-developed features and may not work properly.

Example features where we would likely want to use this include:
 
* #46

  (At a later stage this might get promoted to a normal setting, with the default flipped to being enabled, if we find that at launch time there remain some TeX constructs we don't yet support. Alternatively it might stop needing a setting if we're able to reliably detect when parsing a given math expression whether we're able to support it, and so fall back automatically.)

* #1250

  Our current approach there has the feature enabled by a handful of "do not merge" commits that go at the tips of the PRs. A feature flag could provide a more easily accessible way to try out the partial support.

I expect there will be more examples in the future — these are just the two we're currently working on right now.

gnprice:
BTW one important risk when using feature flags is that if they're allowed to accumulate, they can add more and more complexity to a codebase in the form of conditionals that are in practice always true, or always false (or sometimes worse: almost always true or almost always false). These make the code harder to understand because it's not obvious whether the flag is always one value; and the dead code on the never-used side of the conditional contributes to the maintenance burden of making changes.

This accumulation can happen when a feature gets finished, and the flag turned on for everyone, but the flag is never cleaned up from the code. It can also happen when work on a feature stalls, when it's tempting to keep the feature-flag-gated partial implementation around in the code out of hope that it'll get resumed and finished.

So once we have this system, we'll still want to be selective about using it. Most features won't ever use it, even those that take two or three PRs to implement. And when we do use it for a given feature, we'll want to make sure to drive that project to completion so the flag can be eliminated one way or another within a period of a few weeks or months.

PR #1416, opened 2025-03-17
GlobalSettingsStore, and other refactors

gnprice:
This branch makes several refactors related to global settings (following up on #1167) which came up as part of work on #1409. Some of these are things I set out to do; others are things I saw the opportunity for as I got my hands in this code.

In summary:
* There's now a GlobalSettingsStore, moving out from GlobalStore the responsibility for the details of how global settings are managed. This also makes a convenient home for what had been extension members on GlobalSettingsData.
* The initialization of the GlobalSettings table to have a row now happens as a migration (for existing installs) or at database creation time (for new installs), instead of as a check made when fetching from the table at each app launch.
* Several refactors to the API that app code uses for global settings, and added dartdoc.

#### 86be5c418 test [nfc]: Inline eg.globalSettings

This isn't really any longer than invoking it, and cuts out a
layer of indirection.

#### 32a6e80ce db test: Test updating GlobalSettings

This can be seen as another important aspect of what it means
for ensureGlobalSettings to be functioning properly.

#### 3cf002e33 settings [nfc]: Drop return value from updateGlobalSettings

We don't currently end up using this anywhere.

#### f350c1920 db [nfc]: Extract _migrationSteps as a static

This helpfully makes it impossible for these migration steps
to accidentally call methods on this database instance directly,
rather than on their Migrator parameters `m`.

As a bonus, the code gets three levels less indented.

#### 8ab2c98f4 db: Create GlobalSettings row as part of schema setup

This is a little more complicated in the code overall; but it
simplifies the logic that runs each time the app starts up.

#### 8ed6915d6 settings [nfc]: Split out a GlobalSettingsStore, and GlobalStoreWidget.settingsOf

This allows widgets to listen separately for changes to settings or
to the rest of the global store's data, which is a bit nice.

More fundamentally, as a matter of code organization, it gives us a
place to put code related to the store of global settings where
there's more room to spread out than if we kept it in GlobalStore,
and where the namespace is already specific to global settings.

Those two points in turn will also help us adjust to be more
ergonomic the API the rest of the app uses for interacting with
these settings.

The TODO comments added in this commit mark places where this API
will be more coherent after upcoming commits in this series.

#### ed595fb1b settings [nfc]: Have settingsOf return whole store, not just data

To save code that consults the settings from an extra indirection
".data", we equip GlobalSettingsStore with getters for all the
fields of the data class.  This also makes the store now the natural
home for the additional members that had been in an extension on the
data class.

#### a50426634 store [nfc]: Simplify name GlobalStore.settings, from globalSettings

It's natural that the settings found directly on the global store
are the global settings.

#### 4ef3437a5 store [nfc]: Separate "backend" class from store itself

This commit makes GlobalSettingsStore more self-contained, taking
over from GlobalStore the responsibility for updating settings in
the database at the same time as they get updated in the in-app
cache [GlobalSettingsStore._data].

To do that, it needs to call doUpdateGlobalSettings, the abstract
method that's implemented separately to use a live database in the
real app vs. a vacuous fake backend in tests.  But that means that
method can no longer live on GlobalStore: for GlobalStore needs a
GlobalSettingsStore, but then GlobalSettingsStore would need a
GlobalStore, forming a cycle.  In particular when calling the
GlobalSettingsStore constructor from within the initializer list of
the GlobalStore constructor, we can't pass `this` as an argument
because `this` hasn't yet been constructed.

(If we were determined to make such a cycle work, we could do so by
having a late or nullable field.  But let's take the cycle as a
signal that the design can be cleaned up.)

So, make a new class GlobalStoreBackend as the home for this
abstract method.  A GlobalStore now "has a" GlobalStoreBackend,
rather than "is a" GlobalStoreBackend as it effectively was before.
As a result we can readily construct a "backend" instance first,
and pass that to the GlobalSettingsStore constructor, before going
on to finish constructing the overall GlobalStore.

Probably most or all of the other functionality that LiveGlobalStore
implements should get moved to the new "backend" class too, but
we'll leave that for possible follow-up changes.

#### 6d438f714 settings [nfc]: Use `update` method directly on GlobalSettingsStore

This completes the transition to GlobalSettingsStore as the
self-contained store for the global settings.

#### df70e0462 settings [nfc]: Give settings specific setter methods

This makes the call sites significantly nicer to read.

Relatedly, it turns the specifics of how the settings are
represented in the database into more of an implementation detail
of the settings store.


lib/model/settings.dart
```diff
@@ -44,7 +45,39 @@ enum BrowserPreference {
...
+  /// See also [setThemeSetting].
+  ThemeSetting? get themeSetting => _data.themeSetting;
+
+  /// The user's choice of [BrowserPreference];
+  /// null means use our default choice.
+  ///
+  /// Consider using [effectiveBrowserPreference] or [getUrlLaunchMode].
+  ///
+  /// See also [setBrowserPreference].
+  BrowserPreference? get browserPreference => _data.browserPreference;
```
PIG208, line 75: A reason that `effectiveBrowserPreference` is separate from this was that we didn't opt into implementing a custom `GlobalSettingsData`.

Now that `GlobalSettingsData _data` is nicely hidden as a part of the implementation detail, we can rename `effectiveBrowserPreference` to `browserPreference`, replacing it, and skip the part of the dartdoc redirecting to `effectiveBrowserPreference`.

lib/widgets/store.dart
```diff
@@ -51,6 +53,28 @@ class GlobalStoreWidget extends StatefulWidget {
     return widget!.store;
   }
 
+  /// The user's [GlobalSettings] data within the app's global data store.
```
PIG208, line 56: `GlobalSettings` is the table class; would it be better to refer to `GlobalSettingsStore` since that's what returned by this method?

lib/model/settings.dart
PIG208: I guess keeping `browserPreference` nullable is consistent with `setBrowserPreference`, but I think we don't need to expose a way to set it to `null`.

lib/model/store.dart
```diff
@@ -29,13 +29,30 @@ import 'message_list.dart';
...
 export 'package:drift/drift.dart' show Value;
 export 'database.dart' show Account, AccountsCompanion, AccountAlreadyExistsException;
 
+/// An underlying data store that can support a [GlobalStore],
+/// possibly storing the data to persist between runs of the app.
+///
+/// In the real app, the implementation used is [LiveGlobalStoreBackend],
+/// which stores data persistently in a database on the user's device.
+/// This interface enables tests to use a different implementation.
+abstract class GlobalStoreBackend {
```
PIG208, line 46: This abstraction looks clean! I think it will be helpful for sqlite integration tests that use `LiveGlobalStoreBackend` with a non-`LiveGlobalStore`.

PIG208 (commented):
Thanks! I have read the whole PR and left some comments.

lib/model/settings.dart
```diff
@@ -85,3 +85,18 @@ extension GlobalSettingsHelpers on GlobalSettingsData {
     }
   }
 }
+
```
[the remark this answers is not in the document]
gnprice: How's this?

    Those two points in turn will also help us adjust the API the rest
    of the app uses for interacting with these settings, so as to make
    it more ergonomic.

Same thought but perhaps clearer syntax.

lib/model/settings.dart
gnprice: Yeah, I think hiding the raw `browserPreference` behind (a renamed) `effectiveBrowserPreference` could be a good follow-up change. Then  `setBrowserPreference` would narrow its parameter type to match.

I'll leave that out of this PR, though, because I think it's not on the path to #1409.

lib/widgets/store.dart
gnprice: The return type doesn't need a link in the doc, since it's shown directly in the code. By contrast a link to related other types can be helpful for navigating around. (The same reason is why e.g. the doc on the table class `Accounts` mentions the data class `Account`, which doesn't otherwise appear in the code there and so wouldn't otherwise be easy to navigate to.)

lib/model/store.dart
gnprice: Yeah, that's definitely a possible application of this structure.

gnprice:
Thanks for the review! Pushed a revision, and replied to subthreads above; PTAL.

PIG208:
Thanks! LGTM.

gnprice:

PR #1418, opened 2025-03-18
Handle message moves for recent senders data model

PIG208:
Fixes: #901.


lib/model/recent_senders.dart
```diff
@@ -68,6 +68,52 @@ class RecentSenders {
       [senderId] ??= MessageIdTracker()).add(messageId);
   }
 
+  /// Handles channel/topic updates to the data resulted from a move.
```
chrisbobbe, line 71: I having trouble parsing this sentence. Perhaps: "Handles channel/topic updates when messages are moved."?

test/model/recent_senders_test.dart
```diff
@@ -142,6 +145,169 @@ void main() {
     });
   });
 
+  group('RecentSenders.handleUpdateMessageUpdate', () {
```
chrisbobbe, line 148: 'RecentSenders.handleUpdateMessageEvent'

test/model/recent_senders_test.dart
```diff
@@ -142,6 +145,169 @@ void main() {
...
+        stream: origChannel, topic: origTopic, sender: userX));
+      await prepare(messages);
+      final unknownMessages = List.generate(10, (i) => eg.streamMessage(
+        stream: origChannel, topic: origTopic, sender: userX));
+      checkMatchesMessages(model, messages);
+
+      final streamSenderIdsBefore = model.streamSenders
+        [origChannel.streamId]![userX.userId]!.ids;
+      final topicSenderIdsBefore = model.topicSenders
+        [origChannel.streamId]![TopicName(origTopic)]![userX.userId]!.ids;
```
chrisbobbe, line 190: I read "…SenderIds…" in these names, and it sounds like user IDs but the value is actually message IDs. (Same in several places in these tests)

chrisbobbe (commented):
Thanks! Small comments below.

PIG208:

test/model/recent_senders_test.dart
```diff
@@ -142,6 +145,164 @@ void main() {
...
+        stream: origChannel, topic: origTopic, sender: userX));
+      await prepare(messages);
+      final unknownMessages = List.generate(10, (i) => eg.streamMessage(
+        stream: origChannel, topic: origTopic, sender: userX));
+      checkMatchesMessages(model, messages);
+
+      final messageIdsByUserInStreamBefore = model.streamSenders
+        [origChannel.streamId]![userX.userId]!.ids;
+      final messageIdsByUserInTopicBefore = model.topicSenders
+        [origChannel.streamId]![TopicName(origTopic)]![userX.userId]!.ids;
```
chrisbobbe, line 190: These line breaks feel a little awkward; can we try something like:

```suggestion
      final messageIdsByUserInStreamBefore =
        model.streamSenders[origChannel.streamId]![userX.userId]!.ids;
      final messageIdsByUserInTopicBefore =
        model.topicSenders[origChannel.streamId]![eg.t(origTopic)]![userX.userId]!.ids;
```

(here and elsewhere)

chrisbobbe (approved):
Thanks! Just one small comment below, and I'll mark for Greg's review.

test/model/recent_senders_test.dart
PIG208: Didn't know that `eg.t` exists! Thanks, that's helpful (it still went over by 7 characters `!.ids);`, but it's a bit better than 12 with `TopicName`).

lib/model/recent_senders.dart
```diff
@@ -68,21 +68,61 @@ class RecentSenders {
...
+      // with the same sender, instead of all of them in `event.messageIds`.
+      messages.sort();
+
+      if (newStreamId != origStreamId) {
+        final streamTracker = sendersInChannel?[senderId];
+        final movedMessagesInStreamTracker = streamTracker?.popAll(messages);
+        if (streamTracker?.maxId == null) sendersInChannel?.remove(senderId);
+        if (movedMessagesInStreamTracker != null) {
+          ((streamSenders[newStreamId] ??= {})
+            [senderId] ??= MessageIdTracker()).addAll(movedMessagesInStreamTracker);
```
gnprice, line 97: Should this last line add only `movedMessagesInStreamTracker`, or the whole `messages` list?

I think it should be the whole `messages`. These are messages we know were sent by the given sender, and now have the given channel, so they belong in the new channel's tracker for that sender.

(Also I think it should be an invariant that `movedMessagesInStreamTracker` has all the same messages as `messages` — i.e. that all the messages in `messages` are known to the RecentSenders data structure. That's because `messages` only has messages that were found in `store.messages`, and every message that makes it there should also make it into here.)

This differs from the situation for unreads because a message only belongs in the unreads data structure if it's unread, and the way to find out whether it's unread is by looking in the unreads data structure's own existing data.

lib/model/recent_senders.dart
```diff
@@ -68,21 +68,61 @@ class RecentSenders {
...
+  /// This is a no-op if no message move happened.
+  void handleUpdateMessageEvent(UpdateMessageEvent event, Map<int, Message> cachedMessages) {
+    if (event.moveData == null) {
+      return;
+    }
+    final UpdateMessageMoveData(
+      :origStreamId, :newStreamId, :origTopic, :newTopic) = event.moveData!;
+
+    final messagesBySender = _groupStreamMessageIdsBySender(event.messageIds, cachedMessages);
+    final sendersInChannel = streamSenders[origStreamId];
```
gnprice, line 82:
```suggestion
    final sendersInStream = streamSenders[origStreamId];
```

Given all the other "stream" names here, plus the otherwise very similar `handleDeleteMessageEvent` below which it's useful to compare this code to, I think it's clearest if these particular locals stick with the "stream" names for now. We'll switch them all to "channel" names as part of the post-launch sweep we have in the tracker.

lib/model/recent_senders.dart
```diff
@@ -68,21 +68,61 @@ class RecentSenders {
...
+    final topicsInChannel = topicSenders[origStreamId];
+    final sendersInTopic = topicsInChannel?[origTopic];
+    for (final MapEntry(key: senderId, value: messages) in messagesBySender.entries) {
+      // The later `popAll` calls require the message IDs to be sorted in
+      // ascending order.  Only sort as many as we need: the message IDs
+      // with the same sender, instead of all of them in `event.messageIds`.
+      messages.sort();
```
gnprice, line 89: This feels like something we should be able to count on the server doing — it's an easy guarantee for the server to make, and I'd guess that the server does sort the list and has since forever.

So at a minimum we should leave a `TODO(server)` comment about getting that as an API guarantee. It'd also be good to ask about it in `#api documentation`.

gnprice (commented):
Thanks! Comments below.

In this round I haven't yet read the main commit's tests; I think they can probably be made simpler given one of my comments below.

lib/model/store.dart
```diff
@@ -767,6 +767,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
 
       case UpdateMessageEvent():
         assert(debugLog("server event: update_message ${event.messageId}"));
+        recentSenders.handleUpdateMessageEvent(event, messages);
         _messages.handleUpdateMessageEvent(event);
```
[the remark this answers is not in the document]
PIG208: I think it shouldn't matter because we only need the message IDs and their corresponding sender user IDs, but those are not subject to updates.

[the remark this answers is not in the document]
gnprice: OK, that makes sense.

Let's I guess write that conclusion down in dartdoc on that `handleUpdateMessageEvent` method: the messages parameter should be `store.messages`, and it doesn't matter whether the method is called before or after MessageStore itself is updated.

lib/model/recent_senders.dart
PIG208: I dug a bit but didn't find solid guarantee that this is sorted. The web app suggests otherwise, albeit with quite an [old](https://github.com/zulip/zulip/commit/59e5f2d8fc48ce8b5cb27d97875c7db59a8032c9) comment:

https://github.com/zulip/zulip/blob/82f4ec0d6c1c30e1d1ab72d175deea23530b8ff1/web/src/message_events.ts#L535-L537

Leaving a `TODO(server)` comment and follow up sounds to me. The sort is still needed until we have that guarantee.

lib/model/recent_senders.dart
PIG208: Good catch! I agree that this invariant holds. It would be helpful to add an assertion here, but I think we should still use the return value from `popAll` here, because we otherwise need to construct a `QueueList` from `messages` even when there is one that can be reused.

PIG208:
This has been updated. Thanks for the review!

lib/model/recent_senders.dart
gnprice: I think constructing a QueueList isn't materially more expensive than a plain List, and this implementation is already doing that in `_groupStreamMessageIdsBySender`.

If the invariant gets broken, it doesn't matter which version of the list we use (`messages` vs `movedMessagesInStreamTracker`); that case shouldn't happen, and if it does then bigger things are probably wrong elsewhere than the difference between those lists. Which one to use can be governed by whatever makes the code better for the non-buggy case.

But zooming out… → well, I ended up posting my next thoughts on this below: https://github.com/zulip/zulip-flutter/pull/1418#issuecomment-2774549920

PR #1422, opened 2025-03-19
Upgrade Flutter, packages, and Android build deps

rajveermalviya:


gnprice:
Thanks!

The first three commits (the non-draft commits) look good; merged as #1426.

PR #1424, opened 2025-03-19
[Flutter Migration] migrate emoji_reaction_test.dart

LongCatIsLooong:
Flutter PR https://github.com/flutter/flutter/pull/159999 (fixes a bug where `MediaQuery` provides an incorrect `TextScaler`) breaks a test in `emoji_reaction_test.dart`.

Currently the `isSystemTextScaler` matcher from the `flutter_test` package can be used to match the system text scaler and is intended to be used for migration. Unfortunately it does not have a `checks` equivalent since flutter / flutter_tests is still using matchers and not importing `checks`.

### DIFF
--- a/test/widgets/emoji_reaction_test.dart
+++ b/test/widgets/emoji_reaction_test.dart
@@ -9,6 +9,7 @@ import 'package:flutter_checks/flutter_checks.dart';
 
 import 'package:flutter_test/flutter_test.dart';
 import 'package:http/http.dart' as http;
+import 'package:legacy_checks/legacy_checks.dart';
 import 'package:zulip/api/model/events.dart';
 import 'package:zulip/api/model/model.dart';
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks!

> Currently the `isSystemTextScaler` matcher from the `flutter_test` package can be used to match the system text scaler and is intended to be used for migration. Unfortunately it does not have a `checks` equivalent since flutter / flutter_tests is still using matchers and not importing `checks`.

We can use `check` with `package:legacy_matchers` for this:

```dart
                check(MediaQuery.of(reactionChipsList).textScaler).legacyMatcher(
                  isSystemTextScaler(withScaleFactor: textScaleFactor));
```

I've just pushed a revision doing that, with a new commit message; @gnprice PTAL. :)

LongCatIsLooong:
Thank you for the quick review! Sorry for the breakage.

PR #1428, opened 2025-03-21
Fix DB downgrades to match creating from scratch

### LINKED ISSUE #1427: Schema downgrades fail (from future schema versions)
In #1416, we should have adjusted `AppDatabase._dropAndCreateAll` to reflect the change that PR made to the `onCreate` in `migration`, making it insert the GlobalSettings row.

(Originally noticed this at https://github.com/zulip/zulip-flutter/issues/1421#issuecomment-2737882493 .)

Without that adjustment, if you switch to a branch (like #1421) with a later schema version, then back to schema version 5 — only really possible as a developer — you get this error at startup:
```
E/flutter (15620): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Bad state: No element
E/flutter (15620): #0      List.single (dart:core-patch/growable_array.dart:358:22)
E/flutter (15620): #1      Selectable.getSingle (package:drift/src/runtime/query_builder/statements/query.dart:239:26)
E/flutter (15620): <asynchronous suspension>
E/flutter (15620): #2      AppDatabase.getGlobalSettings (package:zulip/model/database.dart:186:12)
E/flutter (15620): <asynchronous suspension>
E/flutter (15620): #3      LiveGlobalStore.load (package:zulip/model/store.dart:886:28)
E/flutter (15620): <asynchronous suspension>
E/flutter (15620): #4      _GlobalStoreWidgetState.initState.<anonymous closure> (package:zulip/widgets/store.dart:89:21)
E/flutter (15620): <asynchronous suspension>
```
and the app fails to get past the loading screen.

### Workaround

To unwedge from that state, you can switch to a version before schema version 5 (e.g., to this week's v0.0.27 release) and then forward again; or uninstall the app and reinstall.

gnprice:
Fixes #1427.

Prerequisite for #1421.

#### f64450168 db test: Fix migration tests to use intended source versions

The `fromVersion` reference inside this test body, for each of the
test cases, was referring to the same one, mutable `fromVersion`
variable.  So by the time any of them ran, its value was 5, and they
tested migrating from 5 to 2, 5 to 3, 5 to 4, 5 to 5 respectively.
Meanwhile the test names reflected the intended migrations, because
the names were computed eagerly, reading the value of `fromVersion`
before moving on to the next iteration.

To fix the issue, make a fresh `fromVersion` binding inside each
iteration of the loop.

#### 8ed6c5b8e db [nfc]: Assert downgrades are always to current version

And adjust a test to make that true in tests too.

#### e90ea1055 db [nfc]: Simplify downgrades by using that target is latest

As explained in the parent commit, and asserted just above the
_dropAndCreateAll call site, this fact always holds.  That lets
us simplify how downgrades work, and as a bonus cut one step
from the checklist for making schema updates.

This reverts some of the changes made in 601936da7.

#### 4d8a667fb db: Cut transaction wrappers within migrations

No transactions here will do any good if the changes aren't in a
transaction with the `PRAGMA user_version =` statements that update
the database's record of what version the schema is at, and therefore
of which migrations might be needed.

And if Drift is setting up for us such a transaction, it'll
necessarily enclose the whole migration anyway.

#### f03630805 db: Fix downgrades so they match creating from scratch

Fixes #1427.

### DIFF
--- a/lib/model/database.dart
+++ b/lib/model/database.dart
@@ -92,17 +92,19 @@ class AppDatabase extends _$AppDatabase {
   AppDatabase(super.e);
 
   // When updating the schema:
-  //  * Make the change in the table classes, and bump schemaVersion.
+  //  * Make the change in the table classes, and bump latestSchemaVersion.
   //  * Export the new schema and generate test migrations with drift:
   //    $ tools/check --fix drift
   //    and generate database code with build_runner.
   //    See ../../README.md#generated-files for more
   //    information on using the build_runner.
-  //  * Update [_getSchema] to handle the new schemaVersion.
+  //  * Update [_getSchema] to handle the new latestSchemaVersion.
   //  * Write a migration in `_migrationSteps` below.
   //  * Write tests.
+  static const int latestSchemaVersion = 5; // See note.
+
   @override
-  int get schemaVersion => 5; // See note.
+  int get schemaVersion => latestSchemaVersion;
 
   static Future<void> _dropAndCreateAll(Migrator m, {
     required int schemaVersion,
@@ -175,7 +177,7 @@ class AppDatabase extends _$AppDatabase {
           await _dropAndCreateAll(m, schemaVersion: to);
           return;
         }
-        assert(1 <= from && from <= to && to <= schemaVersion);
+        assert(1 <= from && from <= to && to <= latestSchemaVersion);
 
         await m.runMigrationSteps(from: from, to: to, steps: _migrationSteps);
       });
--- a/lib/model/database.dart
+++ b/lib/model/database.dart
@@ -68,25 +68,6 @@ class UriConverter extends TypeConverter<Uri, String> {
   @override Uri fromSql(String fromDb) => Uri.parse(fromDb);
 }
 
-// TODO(drift): generate this
-VersionedSchema _getSchema({
-  required DatabaseConnectionUser database,
-  required int schemaVersion,
-}) {
-  switch (schemaVersion) {
-    case 2:
-      return Schema2(database: database);
-    case 3:
-      return Schema3(database: database);
-    case 4:
-      return Schema4(database: database);
-    case 5:
-      return Schema5(database: database);
-    default:
-      throw Exception('unknown schema version: $schemaVersion');
-  }
-}
-
 @DriftDatabase(tables: [GlobalSettings, Accounts])
 class AppDatabase extends _$AppDatabase {
   AppDatabase(super.e);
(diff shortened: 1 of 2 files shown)


lib/model/database.dart
```diff
@@ -145,37 +124,45 @@ class AppDatabase extends _$AppDatabase {
...
+          RawValuesInsertable({}));
+      }
     },
   );
 
+  Future<void> _createLatestSchema(Migrator m) async {
+    await m.createAll();
+    // Corresponds to `from4to5` above.
+    await into(globalSettings).insert(GlobalSettingsCompanion());
+  }
```
PIG208, line 142: I briefly looked into `beforeOpen` since it is [introduced](https://drift.simonbinder.eu/migrations/#post-migration-callbacks) as a way to "populate data after the database has been created", but I think this PR's approach is still better. Because it fully utilizes when the row needs to be inserted (`beforeOpen` is "called whenever the database is opened, regardless of whether a migration actually ran or not").

lib/model/database.dart
```diff
@@ -68,65 +68,44 @@ class UriConverter extends TypeConverter<Uri, String> {
...
-      }
-    });
+  int get schemaVersion => latestSchemaVersion;
+
+  /// Drop all tables, indexes, etc., in the database.
+  ///
+  /// This includes tables that aren't known to the schema, for example because
+  /// they were defined by a future (perhaps experimental) version of the app
+  /// before switching back to the version currently running.
+  static Future<void> _dropAll(Migrator m) async {
```
PIG208, line 94:
```
    db: Cut transaction wrappers within migrations
```

Makes sense to me. I think drift doesn't set up a transaction for the entire migration.

If we really need one, we can probably wrap the entirety of `onUpgrade` (the line responsible for `PRAGMA user_version =` can be found in `VersionedSchema.runMigrationSteps`, which is called through `Migrator.runMigrationSteps`). While the transaction might prevent corrupting the database if a migration goes wrong, it's not so useful unless the app can tolerate/recover from said migration failure, which is not a goal of this PR.

PIG208 (commented):
Thanks! I tested this manually by changing `latestSchemaVersion` to an earlier version and back, hot restart in between; and reproduced the issue by dropping the fix. Left some comments for discussions. This LGTM.

test/model/database_test.dart
```diff
@@ -127,41 +127,45 @@ void main() {
...
     });
 
     group('migrate without data', () {
       const versions = GeneratedHelper.versions;
       final latestVersion = versions.last;
 
-      int fromVersion = versions.first;
+      int prev = versions.first;
       for (final toVersion in versions.skip(1)) {
+        final fromVersion = prev;
```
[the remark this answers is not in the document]
gnprice: Yeah, exactly — that's another way to handle the same sort of situation.

gnprice:

lib/model/database.dart
chrisbobbe: Filed #2460 for this.

Issue #126, opened 2023-05-26
compose: Support editing an already-sent message


chrisbobbe:
Design from @terpimost: 
![image](https://github.com/user-attachments/assets/54cc1017-4e88-4ce7-a6d9-effdf93f1dfc)

Details are in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4010-6347&m=dev

Updates from the details in Figma:
* For the text when saving the edit is in progress let's say "saving edit…" rather than "sending message…".
* For the text when saving the edit fails let's say "edit wasn't saved" rather than "message wasn't sent".
* Further updates in design thread in chat: [#mobile-design > edit message @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/edit.20message/near/2129881)

---

Original description:

We should probably offer this as an in-place text field within the message list. As Greg points out about this feature in `zulip-mobile` ([link](https://chat.zulip.org/#narrow/stream/412-api-documentation/topic/Sending.20typing.20notifications.20while.20editing.20a.20message/near/1569555)):

> In general the edit-message UI in the mobile app is somewhat unloved; as you noted, the UX for it is borrowed from composing a message (and rather awkwardly so), and the code to implement editing a message is likewise rather jammed into the code for composing. One consequence is that it's easy to accidentally pick up behaviors that compose has even when they aren't intended.

gnprice:
When we implement this, we should aim to offer editing on all message lists, including "All messages" (https://github.com/zulip/zulip-mobile/issues/5619). If we successfully keep at least the code cleanly separated from that for composing a new message, as described above, then that should be straightforward.

(Potentially we'll decide that we want the UX for editing to remain similar to that for composing a message after all. In that case the code will probably have significant pieces in common. But the key will just be to have the differences between edit and compose cleanly factored out, rather than the edit logic being jammed in awkwardly as it is in zulip-mobile.)

gnprice:
This now has a design — thanks @terpimost! I've edited the issue description to show the design.

Chat thread on a couple of specific points in the design:
https://chat.zulip.org/#narrow/stream/48-mobile/topic/edit-message.20design/near/1930070

I think this is already ripe to work on, though. Both of those are things that will be easy to change later.

chrisbobbe:
[...]

PR #1430, opened 2025-03-21
Edit-message (1/n): Refactor compose-box banner logic

chrisbobbe:
These commits help get the code ready to implement the blue "Edit message" banner, with the "Cancel" and "Save" buttons; see Figma:

https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3988-38201&m=dev
image

The most interesting change is to extract a base class `_Banner` out of `_ErrorBanner`:

We'll make a new subclass with a name like `_EditMessageBanner` for #126; here's a rough sketch of what that'll look like:

```dart
class _EditMessageBanner extends _Banner {
  const _EditMessageBanner();

  @override
  String getLabel(ZulipLocalizations zulipLocalizations) =>
    zulipLocalizations.composeBoxBannerLabelEditMessage;

  @override
  Color getLabelColor(DesignVariables designVariables) =>
    designVariables.bannerTextIntInfo;

  @override
  Color getBackgroundColor(DesignVariables designVariables) =>
    designVariables.bannerBgIntInfo;

  @override
  Widget? buildTrailing(context) {
    return Row(mainAxisSize: MainAxisSize.min, spacing: 8, children: [
      ZulipWebUiKitButton(label: 'Cancel',
        onPressed: () {}),
      ZulipWebUiKitButton(label: 'Save',
        attention: ZulipWebUiKitButtonAttention.high,
        onPressed: () {}),
    ]);
  }
}
```

Related: #126

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1389,7 +1389,7 @@ abstract class _Banner extends StatelessWidget {
   /// To control the element's distance from the end edge, override [padEnd].
   Widget? buildTrailing(BuildContext context);
 
-  /// Whether to add 8px trailing padding.
+  /// Whether to apply `end: 8` in [SafeArea.minimum].
   ///
   /// Subclasses can use `false` when the [buildTrailing] element
   /// is meant to abut the edge of the screen
@@ -1410,7 +1410,7 @@ abstract class _Banner extends StatelessWidget {
       decoration: BoxDecoration(
         color: getBackgroundColor(designVariables)),
       child: SafeArea(
-        minimum: const EdgeInsetsDirectional.only(start: 8)
+        minimum: EdgeInsetsDirectional.only(start: 8, end: padEnd ? 8 : 0)
           // (SafeArea.minimum doesn't take an EdgeInsetsDirectional)
           .resolve(Directionality.of(context)),
         child: Padding(
@@ -1426,7 +1426,6 @@ abstract class _Banner extends StatelessWidget {
                 const SizedBox(width: 8),
                 trailing,
               ],
-              if (padEnd) const SizedBox(width: 8),
             ]))));
   }
 }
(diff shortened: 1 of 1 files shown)


lib/widgets/compose_box.dart
```diff
@@ -1320,38 +1320,85 @@ class StreamComposeBoxController extends ComposeBoxController {
...
+
+class _ErrorBanner extends _Banner {
+  const _ErrorBanner({
+    required String Function(ZulipLocalizations) getLabel,
+  }) : _getLabel = getLabel;
+
+  @override
+  String getLabel(ZulipLocalizations zulipLocalizations) =>
+    _getLabel(zulipLocalizations);
+  final String Function(ZulipLocalizations) _getLabel;
```
PIG208, line 1385: I can see how this is needed for `_EditMessageBanner` to just be able to override `getLabel`. What's the drawback of having an API similar to Flutter widgets like `Container` and make `_ErrorBanner` and `_EditMessageBanner` direct subclasses of `StatelessWidget`? For example:

```dart
class _EditMessageBanner extends StatelessWidget {
  const _EditMessageBanner();

  @override
    final trailing = Row(
      // …
    );
    return _Banner(
      label: zulipLocalizations.errorDialogTitle,
      labelColor: designVariables.btnLabelAttMediumIntDanger,
      backgroundColor: designVariables.bannerBgIntDanger,
      trailing: trailing);
  }
}
```

Maybe keeping `_Banner` as an abstract base class is needed for code structuring reasons/later changes?

PIG208 (commented):
Thanks! Left a comment; this otherwise LGTM.

lib/widgets/compose_box.dart
chrisbobbe: I think the idea was just to have one fewer widget in the widget tree.

gnprice: That alternate API looks like a good one to me. Maybe we'll refactor to it later.

I think an extra widget in the tree isn't a big deal. The style Flutter goes for in general is that widgets are cheap; there are lots of widgets; and they're configured mostly by composition (like in the example code above), rather than by inheritance.

gnprice:
Thanks to you both! Looks good; merged.

Issue #831, opened 2024-07-22
Get Zulip color palette from Figma design


gnprice:
There's a new color palette for Zulip's UIs, designed recently by @terpimost:
https://www.figma.com/design/msWyAJ8cnMHgOMPxi7BUvA/Zulip-Web-UI-kit?node-id=900-1861&t=7OnwOYAxWXBEtYXW-0

We should export that and turn it into some generated Dart code, so that our UI code can refer to it.

gnprice:
The steps for implementing this will basically be:
* Export the variables of the color palette into a file in any halfway-reasonable format, using some Figma plugin.
* Have a little script of our own that transforms that into the desired generated Dart code.

For the plugin, after searching through Figma's store of plugins for a bit, we're basically doomed to using some code that a random person on the Internet put together and decided to publish. That's a lot more palatable if the code is at least open source. There don't seem to be very many open-source Figma plugins, but here's two that should be able to do the job:
https://github.com/tokens-bruecke/figma-plugin
https://www.figma.com/community/plugin/1254538877056388290/tokensbrucke
and
https://github.com/lukasoppermann/design-tokens
https://www.figma.com/community/plugin/888356646278934516/design-tokens

Here's another plugin I ran across that is open source and does almost the right job, but in reverse:
https://github.com/microsoft/figma-variables-import
https://www.figma.com/community/plugin/1253424530216967528/variables-import

gnprice:
Oh and about running plugins that are from a random person on the Internet:

Figma has a pretty credible-seeming story for how plugins are contained in what they can do:
https://www.figma.com/blog/an-update-on-plugin-security/

But even the intended security model means that when you choose to run a plugin, it can "modify any data in your Figma document" and "communicate with any server over the internet":
https://www.figma.com/blog/an-update-on-plugin-security/#background

So that remains quite a lot of trust. It's not *more* trust than any of the numerous tools one ordinarily runs on a development machine, published by a variety of people on the Internet to pub.dev or npmjs.com or pypi.python.org… but it's not clear it's really any less trust, either. So it's good for the plugins we use to be open source, just like those other tools are, as that's a key layer of protection we're used to relying on.

alya:
The colors were added in the web app in https://github.com/zulip/zulip/pull/30663/commits/44051107a19ef25ac5c174c5d36a759bec75f89d , in case that's helpful.

terpimost:
[...]

PR #1432, opened 2025-03-21
Edit-message (3/n): Implement ZulipWebUiKitButton, for "Cancel" / "Save" buttons

chrisbobbe:
This PR implements some variants of the "Button" component in the "Zulip Web UI kit" in Figma: https://www.figma.com/design/msWyAJ8cnMHgOMPxi7BUvA/Zulip-Web-UI-kit?node-id=1-8&p=f&m=dev

The Figma uses this component for the "Cancel" and "Save" buttons in the edit-message compose box:

https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3988-38200&m=dev
| Light | Dark |
| --- | --- |
| image | image |

Not all values of the "intent" and "attention" params are implemented here; just the "info" intent with "medium" and "high" attention. In the edit-message compose box banner, both buttons have the "info" intent; the cancel button has "medium" attention and the save button has "high" attention.

Here are screenshots of the buttons, made with these commits and the ones from #1430 (plus the `_EditMessageBanner` class in the PR description):

## Cancel button ("info" intent, "medium" attention)

| Light; normal state | Light; pressed state | Dark; normal state | Dark; pressed state |
| --- | --- | --- | --- |
| image | image | image | image |

## Save button ("info" intent, "high" attention)

| Light; normal state | Light; pressed state | Dark; normal state | Dark; pressed state |
| --- | --- | --- | --- |
| image | image | image | image |

Related: #126

### DIFF
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -1,0 +1,47 @@
+import 'package:flutter/widgets.dart';
+
+/// Apply [Transform.scale] to the child widget when tapped, and reset its scale
+/// when released, while animating the transitions.
+class AnimatedScaleOnTap extends StatefulWidget {
+  const AnimatedScaleOnTap({
+    super.key,
+    required this.scaleEnd,
+    required this.duration,
+    required this.child,
+  });
+
+  /// The terminal scale to animate to.
+  final double scaleEnd;
+
+  /// The duration over which to animate the scale change.
+  final Duration duration;
+
+  final Widget child;
+
+  @override
+  State<AnimatedScaleOnTap> createState() => _AnimatedScaleOnTapState();
+}
+
+class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
+  double _scale = 1;
+
+  void _changeScale(double scale) {
+    setState(() {
+      _scale = scale;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      behavior: HitTestBehavior.translucent,
+      onTapDown: (_) =>  _changeScale(widget.scaleEnd),
+      onTapUp: (_) =>    _changeScale(1),
+      onTapCancel: () => _changeScale(1),
+      child: AnimatedScale(
+        scale: _scale,
+        duration: widget.duration,
+        curve: Curves.easeOut,
+        child: widget.child));
+  }
+}
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -1,4 +1,126 @@
-import 'package:flutter/widgets.dart';
+import 'package:flutter/material.dart';
+
+import 'color.dart';
+import 'text.dart';
+import 'theme.dart';
+
+/// The "Button" component from Zulip Web UI kit.
+///
+/// The Figma uses this for the "Cancel" and "Save" buttons in the compose box
+/// for editing an already-sent message.
+///
+/// See Figma:
+///   * Component: https://www.figma.com/design/msWyAJ8cnMHgOMPxi7BUvA/Zulip-Web-UI-kit?node-id=1-2780&t=Wia0D0i1I0GXdD9z-0
+///   * Edit-message compose box: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3988-38201&m=dev
+class ZulipWebUiKitButton extends StatelessWidget {
+  const ZulipWebUiKitButton({
+    super.key,
+    this.attention = ZulipWebUiKitButtonAttention.medium,
+    this.intent = ZulipWebUiKitButtonIntent.info,
+    required this.label,
+    required this.onPressed,
+  });
+
+  final ZulipWebUiKitButtonAttention attention;
+  final ZulipWebUiKitButtonIntent intent;
+  final String label;
+  final VoidCallback? onPressed;
+
+  WidgetStateColor _backgroundColor(DesignVariables designVariables) {
+    switch ((attention, intent)) {
+      case (ZulipWebUiKitButtonAttention.medium, ZulipWebUiKitButtonIntent.info):
+        return WidgetStateColor.fromMap({
+          WidgetState.pressed: designVariables.btnBgAttMediumIntInfoActive,
+          ~WidgetState.pressed: designVariables.btnBgAttMediumIntInfoNormal,
+        });
+      case (ZulipWebUiKitButtonAttention.high, ZulipWebUiKitButtonIntent.info):
+        return WidgetStateColor.fromMap({
+          WidgetState.pressed: designVariables.btnBgAttHighIntInfoActive,
+          ~WidgetState.pressed: designVariables.btnBgAttHighIntInfoNormal,
+        });
+    }
+  }
+
+  Color _labelColor(DesignVariables designVariables) {
+    switch ((attention, intent)) {
+      case (ZulipWebUiKitButtonAttention.medium, ZulipWebUiKitButtonIntent.info):
+        return designVariables.btnLabelAttMediumIntInfo;
+      case (ZulipWebUiKitButtonAttention.high, ZulipWebUiKitButtonIntent.info):
+        return designVariables.btnLabelAttHigh;
+    }
+  }
+
+  TextStyle _labelStyle(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    return TextStyle(
+      color: _labelColor(designVariables),
+      fontSize: 17,
+      height: 1.20,
+      letterSpacing: proportionalLetterSpacing(context,
+        0.006, baseFontSize: 17),
+    ).merge(weightVariableTextStyle(context, wght: 600));
+  }
+
+  BorderSide _borderSide(DesignVariables designVariables) {
+    switch (attention) {
+      case ZulipWebUiKitButtonAttention.medium:
+        // TODO inner shadow effect like `box-shadow: inset`, following Figma;
+        //   needs Flutter support for something like that:
+        //     https://github.com/flutter/flutter/issues/18636
+        //     https://github.com/flutter/flutter/issues/52999
+        //   For now, we just use a solid-stroke border with half the opacity
+        //   and half the width.
+        return BorderSide(
+          color: designVariables.btnShadowAttMed.withFadedAlpha(0.5),
+          width: 0.5);
+      case ZulipWebUiKitButtonAttention.high:
+        return BorderSide.none;
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    return AnimatedScaleOnTap(
+      scaleEnd: 0.96,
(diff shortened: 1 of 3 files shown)


lib/widgets/button.dart
```diff
@@ -0,0 +1,169 @@
...
+  // low,
+}
+
+enum ZulipWebUiKitButtonIntent {
+  // neutral,
+  // warning,
+  // danger,
+  info,
+  // success,
+  // brand,
```
gnprice, line 122: One good reason to add these only lazily, as you're doing here: to keep down how many different colors we have to maintain in theme.dart, until we finish #831.

lib/widgets/theme.dart
```diff
@@ -246,8 +260,15 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
...
     required this.borderBar,
     required this.borderMenuButtonSelected,
+    required this.btnBgAttHighIntInfoActive,
+    required this.btnBgAttHighIntInfoNormal,
+    required this.btnBgAttMediumIntInfoActive,
+    required this.btnBgAttMediumIntInfoNormal,
+    required this.btnLabelAttHigh,
     required this.btnLabelAttLowIntDanger,
+    required this.btnLabelAttMediumIntInfo,
     required this.btnLabelAttMediumIntDanger,
```
PIG208, line 270: nit: order
```suggestion
    required this.btnLabelAttMediumIntDanger,
    required this.btnLabelAttMediumIntInfo,
```

lib/widgets/button.dart
```diff
@@ -0,0 +1,169 @@
...
+          shape: RoundedRectangleBorder(
+            side: _borderSide(designVariables),
+            borderRadius: BorderRadius.circular(4)),
+          splashFactory: NoSplash.splashFactory,
+          tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+          minimumSize: Size(kMinInteractiveDimension, 28),
+        ).copyWith(backgroundColor: _backgroundColor(designVariables)),
+        onPressed: onPressed,
+        child: ConstrainedBox(
+          constraints: BoxConstraints(maxWidth: 240),
```
PIG208, line 101: I wonder if we need a different value for the mobile app. [Presumably](https://www.figma.com/design/msWyAJ8cnMHgOMPxi7BUvA/Zulip-Web-UI-kit?node-id=667-7292&m=dev) this is for web.

lib/widgets/button.dart
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    return AnimatedScaleOnTap(
+      scaleEnd: 0.96,
+      duration: Duration(milliseconds: 50),
```
PIG208, line 87: Looks like duration in Figma is 100ms:

image

lib/widgets/button.dart
```diff
@@ -0,0 +1,169 @@
...
+    this.attention = ZulipWebUiKitButtonAttention.medium,
+    this.intent = ZulipWebUiKitButtonIntent.info,
+    required this.label,
+    required this.onPressed,
+  });
+
+  final ZulipWebUiKitButtonAttention attention;
+  final ZulipWebUiKitButtonIntent intent;
+  final String label;
+  final VoidCallback? onPressed;
```
PIG208, line 27: Looks like the Figma design does not offer style for disabled buttons. Widgets like `TextButton` are disabled when `onPressed` is `null`, but we don't have special styling to indicate that (and `AnimatedScaleOnTap` still works).

lib/widgets/theme.dart
```diff
@@ -188,8 +195,15 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     bgTopBar: const Color(0xff242424),
     borderBar: const Color(0xffffffff).withValues(alpha: 0.1),
     borderMenuButtonSelected: Colors.white.withValues(alpha: 0.1),
+    btnBgAttHighIntInfoActive: const Color(0xff1e41d3),
+    btnBgAttHighIntInfoNormal: const Color(0xff1e41d3),
+    btnBgAttMediumIntInfoActive: const Color(0xff97b6fe).withValues(alpha: 0.22),
```
PIG208, line 200: This should be 0.12 as well.

PIG208 (commented):
This is neat. Thanks for working on this! Left some comments.

lib/widgets/button.dart
chrisbobbe: If 240 doesn't seem OK, could you say more about why? If it's OK but just not ideal, I think it's fine to leave it as-is and tune later based on feedback.

lib/widgets/button.dart
chrisbobbe: Thanks for spotting that! I didn't see it, so instead found an animation in the web app with Chrome dev tools; probably ought to have made a comment about that :)

lib/widgets/theme.dart
chrisbobbe: 🫣

lib/widgets/theme.dart
chrisbobbe: Thanks for the catch!

chrisbobbe:

lib/widgets/button.dart
PIG208: Usually when we have wider-than-usual buttons, they occupy the screen-width or the width of the parent; for my phone a 240px button would be a bit narrower than my screen. I think leaving it this way and tuning later sounds like a good plan.

PIG208:
Thanks for the update! This LGTM.

lib/widgets/button.dart
```diff
@@ -0,0 +1,169 @@
...
+      child: TextButton(
+        style: TextButton.styleFrom(
+          padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 4),
+          foregroundColor: _labelColor(designVariables),
+          shape: RoundedRectangleBorder(
+            side: _borderSide(designVariables),
+            borderRadius: BorderRadius.circular(4)),
+          splashFactory: NoSplash.splashFactory,
+          tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+          minimumSize: Size(kMinInteractiveDimension, 28),
```
gnprice, line 97: Hmm, 28px is awfully small for a touch target.

The [Figma design says](https://www.figma.com/design/msWyAJ8cnMHgOMPxi7BUvA/Zulip-Web-UI-kit?node-id=667-7293&m=dev):
> People might say that 44px is recommended minimum for touch screens but we assume that this is for web app.

Here we are indeed talking about a touch screen, though :slightly_smiling_face: So being just 28px high, rather than 44px or 48px, seems likely to be frustrating.

It'd be fine if the visual button is on the short side, as long as the tap target is a good height. So in the [edit-message design](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3988-38201&m=dev), it calls for the buttons to be 28px high but then the whole banner is 40px. So making the banner only slightly taller than the design would make room for a 44px tap target, which could still enclose a 28px visual button.

These Material buttons have a facility for padding out the tap target beyond the visual button — see the call site of `_InputPadding` in `button_style_button.dart` (in `_ButtonStyleState.build`). Can we use that to get that effect?

gnprice (commented):
Thanks! This structure looks good; a couple of comments below.

lib/widgets/button.dart
```diff
@@ -0,0 +1,169 @@
...
+
+  TextStyle _labelStyle(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    return TextStyle(
+      color: _labelColor(designVariables),
+      fontSize: 17,
+      height: 1.20,
+      letterSpacing: proportionalLetterSpacing(context,
+        0.006, baseFontSize: 17),
+    ).merge(weightVariableTextStyle(context, wght: 600));
```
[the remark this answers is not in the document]
chrisbobbe: Odd, I took these details from the buttons as they appear in the compose-box frame:

https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3988-38201&m=dev

```css
overflow: hidden;
color: var(--btn-label-att_high, #FFF);
text-align: center;
text-overflow: ellipsis;
font-family: "Source Sans 3";
font-size: 17px;
font-style: normal;
font-weight: 600;
line-height: 120%; /* 20.4px */
letter-spacing: 0.102px;
```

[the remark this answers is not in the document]
gnprice: Huh. Yeah, I guess leave a comment, then. We can leave this code specialized to what that bit of the mobile design calls for, unless we find that other designs call for something different in which case we can generalize as needed.

chrisbobbe:

test/widgets/button_test.dart
```diff
@@ -0,0 +1,69 @@
...
+      await tester.pump();
+
+      final element = tester.element(buttonFinder);
+      final renderObject = element.renderObject as RenderBox;
+      final width = renderObject.size.width;
+      final buttonTopLeft = tester.getTopLeft(buttonFinder);
+      final buttonCenter = tester.getCenter(buttonFinder);
+      check(element).size.isNotNull().height.equals(44); // includes outer padding
+
+      // Outer padding responds to taps, not just the painted part.
```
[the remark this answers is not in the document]
gnprice: Or maybe what I really want here is to move `renderObject` and `width` down to where they're used.

test/widgets/button_test.dart
```diff
@@ -0,0 +1,69 @@
...
+      int numTaps = 0;
+      for (double y = 0; y < 44; y++) {
+        await tester.tapAt(Offset(buttonCenter.dx, y + buttonTopLeft.dy));
+        numTaps++;
+      }
+      check(numTapsHandled).equals(numTaps);
+
+      final textScaler = TextScaler.linear(textScaleFactorVariants.currentValue!)
```
gnprice, line 50: I think this test would break up pretty cleanly into two tests at this spot: one checks the response to taps, and one checks the painting. That would probably make them both a little easier to read.

lib/widgets/button.dart
```diff
@@ -0,0 +1,197 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    // With [MaterialTapTargetSize.padded],
+    // make [TextButton] set 44 instead of 48 for the touch-target height.
+    final visualDensity = VisualDensity(vertical: -1);
+    // A value that [TextButton] adds to some of its layout parameters;
+    // we can cancel out those adjustments by subtracting it.
+    final densityVerticalAdjustment = visualDensity.baseSizeAdjustment.dy;
```
gnprice, line 99: Cool, glad you found a way to control this.

gnprice (commented):
Thanks for the revision! The implementation all looks good. Small comments from reading through the tests.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

PR #1435, opened 2025-03-25
anchors 4/n: Control which sliver paints over which

gnprice:
This is the next round after #1316, toward #82.

Since that PR, the sticky header has been properly laid out and painted even in cases where it needs to overflow on top of another sliver… so long as the sliver with the header gets painted after the sliver it needs to overflow upon. This PR subclasses CustomScrollView and related classes (most pertinently RenderViewport) so that it's always possible to arrange for that to happen.

One welcome consequence is to clear out all the skipped tests that #1316 left in `sticky_header_test.dart`.

I've also sent a PR upstream to add this same functionality to CustomScrollView (and RenderViewport etc.) itself:
https://github.com/flutter/flutter/pull/164818
Hopefully that will land and let us cut out a lot of this code. In the meantime this lets us move forward without waiting; and this custom local version is actually how I drafted this feature in the first place, before adapting it into an upstream PR, so it wasn't much work to polish it up to land here.

#### a1cd62ad6 sticky_header example: Add back-to-back list with headers at top

This example demonstrates a bug we'll shortly fix: an overflowing
sliver getting painted before the sliver it's meant to overflow onto.

#### 83cdc117e scrolling: Add CustomPaintOrderScrollView

This lets us control the paint order between the slivers, so that
the top sliver is able to overflow over the bottom sliver when a
sticky header calls for that.

I've sent a PR upstream to add the same feature to CustomScrollView
itself:
  https://github.com/flutter/flutter/pull/164818

That PR has been delayed by infra issues in Google testing, though.
So doing this in our tree lets us move ahead without waiting.

(I actually wrote this version first, then adapted it to produce the
upstream PR.  So landing this version isn't significant extra work.)

#### 1ce40cf8e sticky_header test: Use CustomPaintOrderScrollView in test, example, and doc

By letting us control the paint order between slivers, this
fixes all the skipped tests in this test file.

#### 8f72cbd2b msglist [nfc]: Use SliverPaintOrder.firstIsTop

This changes the paint order of these slivers.  Right now that has
no observable effect, because the second sliver has zero height
and paints nothing.

In the future when we put the messages in back-to-back slivers,
this is the paint order we'll need so that the top sliver can have
a sticky header overflow properly over the bottom sliver.

### DIFF
--- a/lib/widgets/scrolling.dart
+++ b/lib/widgets/scrolling.dart
@@ -1,0 +1,39 @@
+import 'package:flutter/material.dart';
+
+/// A [SingleChildScrollView] that always shows a Material [Scrollbar].
+///
+/// This differs from the behavior provided by [MaterialScrollBehavior] in that
+/// (a) the scrollbar appears even when [scrollDirection] is [Axis.horizontal],
+/// and (b) the scrollbar appears on all platforms, rather than only on
+/// desktop platforms.
+// TODO(upstream): SingleChildScrollView should have a scrollBehavior field
+//   and pass it on to Scrollable, just like ScrollView does; then this would
+//   be covered by using that.
+// TODO: Maybe show scrollbar only on mobile platforms, like MaterialScrollBehavior
+//   and the base ScrollBehavior do?
+class SingleChildScrollViewWithScrollbar extends StatefulWidget {
+  const SingleChildScrollViewWithScrollbar(
+    {super.key, required this.scrollDirection, required this.child});
+
+  final Axis scrollDirection;
+  final Widget child;
+
+  @override
+  State<SingleChildScrollViewWithScrollbar> createState() =>
+    _SingleChildScrollViewWithScrollbarState();
+}
+
+class _SingleChildScrollViewWithScrollbarState
+    extends State<SingleChildScrollViewWithScrollbar> {
+  final ScrollController controller = ScrollController();
+
+  @override
+  Widget build(BuildContext context) {
+    return Scrollbar(
+      controller: controller,
+      child: SingleChildScrollView(
+        controller: controller,
+        scrollDirection: widget.scrollDirection,
+        child: widget.child));
+  }
+}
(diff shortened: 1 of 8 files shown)


gnprice:
I've now also sent the next step, #1436. That PR is a draft because it still needs tests.

lib/widgets/scrolling.dart
```diff
@@ -37,3 +38,209 @@ class _SingleChildScrollViewWithScrollbarState
...
+//     https://github.com/flutter/flutter/issues/121173
+//     https://github.com/flutter/flutter/issues/121173#issuecomment-1914959184
+//     https://github.com/flutter/flutter/issues/70836
+//   A delegate would give enough flexibility for that and much else,
+//   but I'm not sure how many use cases wouldn't be covered by a small enum.
+//
+//   Ah, and here's a more on-point issue (more recently):
+//     https://github.com/flutter/flutter/issues/145592
+//
+// TODO: perh sticky_header should configure a CustomPaintOrderScrollView automatically?
```
chrisbobbe, line 102: nit: "perh" for "perhaps", right?

chrisbobbe (commented):
Thanks! Two small comments below, otherwise LGTM! 🎉

test/widgets/message_list_test.dart
```diff
@@ -103,9 +103,11 @@ void main() {
       .findsOne();
   }
 
+  ScrollView findScrollView(WidgetTester tester) =>
+    tester.widget<ScrollView>(find.byWidgetPredicate((w) => w is ScrollView));
```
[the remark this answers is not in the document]
gnprice: Sure, good thought

lib/widgets/scrolling.dart
gnprice: Ha, yeah — that's a bit of shorthand I use in notes to myself, which I guess this started as and then I didn't catch that before turning it into a PR.

gnprice:
Thanks for the review! Merged, with those fixes.

PR #1444, opened 2025-03-27
ios build: Take auto updates to Podfile.lock and an Xcode file

chrisbobbe:
I think the Podfile.lock changes are needed because #1426 didn't include the `pod` step of `tools/upgrade`.

I suspect the Xcode-file update is due to an Xcode upgrade.

Done by building for iOS, with Xcode 16.2 (16C5032a), then (when I remembered that macOS/Podfile.lock needs updating too) `tools/upgrade pod`.

-----

cc @rajveermalviya as the author of #1426.

### DIFF
--- a/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ b/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -26,6 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      customLLDBInitFile = "$(SRCROOT)/Flutter/ephemeral/flutter_lldbinit"
       shouldUseLaunchSchemeArgsEnv = "YES">
       <MacroExpansion>
          <BuildableReference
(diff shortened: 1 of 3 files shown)


gnprice:
(Ah, I suggested "ios build" as a summary-line prefix because I was thinking of the Xcode config changes as a separate commit. I agree with your first thought that "ios deps" is a good prefix for the Podfile.lock changes.)

ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -26,6 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      customLLDBInitFile = "$(SRCROOT)/Flutter/ephemeral/flutter_lldbinit"
```
gnprice, line 29: Looking at this line more closely, the content of it sounds like something the `flutter` tool might put there.

And indeed, `git log -G flutter_lldbinit` in the Flutter tree finds this from earlier this month, 2025-03-06:
flutter/flutter@eb07c51230c70a0ecaf010a02e7862f69be3b32b

So this is a side effect of the Flutter upgrade in #1426.

rajveermalviya: Right, I'd only ran those commits in the draft (#1422) on Android, sorry, forgot to note that in #1426.

(Guess this could've been caught if we had the iOS build in CI (#773), which I think should be good to have especially now, as #1379 introduces some Swift code.)

gnprice: Cool, makes sense.

And yeah, agreed on both points re #773.

gnprice: Made a note of that on #773, and bumped it up to M5b (from M6).

rajveermalviya (approved):
Tested on iOS Simulator, and don't see any automated changes being made when running on macOS.

chrisbobbe:

PR #1445, opened 2025-03-27
Insert time element global

AshutoshKhadse23:
I had added the global time element

Related #1215

https://github.com/user-attachments/assets/2f394c52-8eda-4d43-9b37-6654e67e8d9a

Apologies for two unsuccessful PR


gnprice:
As you know from other PRs, we're focusing on launch issues until the app is launched; and this is a post-launch issue. So I'm marking it as draft to get it out of our queue.

Before we can review it (once we do move on to post-launch issues), it will also need to include tests. See our README for details.

PR #1447, opened 2025-03-27
notif android: Migrate to cross-platform Pigeon API for navigation

rajveermalviya:
Split from #1379.

Migrates the Android notification routing implementation from URL-encoded payloads to instead implement and use the same Pigeon API that iOS uses in #1379.


chrisbobbe:
Is this superseded by #2043; should we close this?

rajveermalviya:
Correct, closing this.

Issue #1158, opened 2024-12-13
List of topics in channel


gnprice:
In the legacy app, if you're looking at the messages in a topic, we have a button in the app bar (a "list" icon) that gets you a list of all the topics in that channel. It's sorted by recency, and shows the number of unreads, whether you have an @-mention, and whether the topic is muted etc.

Basically it's similar to what you get in the web left sidebar after hitting ["Show all topics"](https://zulip.com/help/finding-a-conversation-to-read#from-the-left-sidebar) under a given channel.

We should offer the same thing in the new app. I wasn't sure people really use that feature; but it [came up in user feedback](https://chat.zulip.org/#narrow/channel/48-mobile/topic/topics.20in.20channel/near/2002291) after yesterday's beta announcement, so that's a sign that people do.

### Design

Design is in Figma:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6819-35869&m=dev

![Image](https://github.com/user-attachments/assets/25dbafb8-3a4c-4f82-ada6-89f7e2b590fe)

For this issue, differences from the design there:
* The design shows a "pinned topics" feature, which is a possible future Zulip feature but doesn't currently exist. Pending that feature, we'll leave out the "pinned" icon, and avoid using up horizontal space for it.
* For this first version, let's leave out the "filter" feature. I'll file a post-launch followup issue for that. → #1384 
* For this issue, we'll leave out the "New topic" button and the modal it leads to. I'll file a separate, post-launch issue for that. → #1385

### Implementation

The list of topics will come from making a request to the server, after opening the page.

As a follow-up task, after getting the list for a given channel we might store it (on `ChannelStore`), and keep it up to date when handling events, so that coming back a second time to this topic-list page for a given channel is fast. For this PR, though, that's out of scope. Instead the list of topics can just live on a `StatefulWidget` for the page.

### References
* Design thread: [#mobile-design > Topic list in channel](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Topic.20list.20in.20channel)
* A different feature serving some of the same use cases: #412

chrisbobbe:
Tim and others propose featuring this page more prominently than the interleaved channel view:

Tim [said](https://chat.zulip.org/#narrow/channel/48-mobile/topic/topics.20in.20channel/near/2005980):
[...]

PR #1449, opened 2025-03-29
topicList: Add topic list page for each channel

lakshya1goel:
fixes: #1158

Related CZO Discussion: [#mobile-design > Topic list in channel](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Topic.20list.20in.20channel)
Figma link: [Link](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6819-35869&t=OXYuFM8H0DFWWYdr-0)

Light Theme:

| Image 1 | Image 2 |
|---------|---------|
|WhatsApp Image 2025-05-01 at 7 57 45 PM | WhatsApp Image 2025-03-30 at 9 20 37 PM |

Dark Theme:

| Image 1 | Image 2 |
|---------|---------|
| WhatsApp Image 2025-05-01 at 7 57 45 PM(1) | WhatsApp Image 2025-03-30 at 9 20 38 PM(1) |

Video Demonstration:

https://github.com/user-attachments/assets/7671d243-e826-466f-8340-c45ac1fc1d8f


alya:
Why are the topics italicized? We generally use italics only for the "general chat" topic.

lakshya1goel:
Hi @PIG208 I have pushed the revision and also updated the PR description. Please have a look. Thanks!

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+            child: Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 8),
+              child: Row(
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  Expanded(
+                    child: Text(
+                      topic.name.displayName.startsWith('✔')
+                          ? topic.name.displayName.substring(2).trim()
+                          : topic.name.displayName,
```
PIG208, line 273: We can just use `topic.name.unresolve()` instead.

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+        child: Text(ZulipLocalizations.of(context).noTopicsInChannel,
+          style: TextStyle(
+            color: DesignVariables.of(context).labelMenuButton,
+            fontSize: 16,
+          )),
+      );
+    }
+
+    final sortedTopics = List<GetStreamTopicsEntry>.from(_model!.topics!);
+    sortedTopics.sort((a, b) => b.maxId.compareTo(a.maxId));
```
PIG208, line 193: It probably makes more sense to sort the topics with the model. Because if the topics didn't change since last build, re-sorting them is not necessary.

lib/model/topic_list.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:flutter/foundation.dart';
+import '../api/route/channels.dart';
+import '../model/store.dart';
+
+class TopicListView extends ChangeNotifier {
+  TopicListView({required this.store, required this.streamId});
```
PIG208, line 6: This feels like a store since it is where the data actually lives on the client, whereas `MessageListView`'s data comes from `MessageStore`.

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+      }
+    } catch (e) {
+      if (!mounted) return;
+      final zulipLocalizations = ZulipLocalizations.of(context);
+      showErrorDialog(
+        context: context,
+        title: zulipLocalizations.errorFetchingTopics,
+        message: e.toString());
+    }
+  }
```
PIG208, line 160: I think with the way how `fetchTopics` is used right now, having a separate class for the data does not feel like a helpful abstraction, since it adds a layer of indirection as a trade-off.

Right now, we only fetch topics `onNewStore`, and there is no retry logic yet. For error handling, The `_fetchTopics` helper does not really care if `errorMessage` lives on as a part of the model. This also requires as to manage an additional listener of a `ChangeNotifier`.

It would probably be cleaner to inline the `getStreamTopics` here, and only consider extracting a model when there is more complexity that warrants a thicker data model. The only mutable states needed are `isLoading` and `topics`.

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+            padding: const EdgeInsets.only(top: 10.0),
+            child: SizedBox(
+              height: 16,
+              width: 16,
+              child: topic.name.displayName.startsWith('✔')
+                ? Icon(
+                    ZulipIcons.check,
+                    size: 12,
+                    color: designVariables.textMessage,
+                  )
```
PIG208, line 259: The opacity of the icon does not match the design:

image

PIG208: The size of the icon should be 16px.

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+            context,
+            channelId: streamId,
+            topic: topic.name,
+            someMessageIdInTopic: null,
+          );
+        },
+        child: Row(crossAxisAlignment: CrossAxisAlignment.start, children: [
+          const SizedBox(width: 28),
+          Padding(
+            padding: const EdgeInsets.only(top: 10.0),
```
PIG208, line 250: The bottom padding of the icon is missing.

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+                  ),
+                  const SizedBox(width: 8),
+                  if (hasUnreads) ...[
+                    UnreadCountBadge(
+                      count: unreads,
+                      backgroundColor: designVariables.bgCounterUnread,
+                      bold: true,
+                    ),
+                  ],
+                ],
```
PIG208, line 293: There is 12px padding before the end:

image

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,301 @@
...
+            context: context,
+            narrow: TopicNarrow(streamId, topic.name),
+          ));
+        },
+        onLongPress: () {
+          showTopicActionSheet(
+            context,
+            channelId: streamId,
+            topic: topic.name,
+            someMessageIdInTopic: null,
```
PIG208, line 244: Looks like we can conveniently use `topic.maxId` for this.

PIG208 (commented):
Thanks! I did a partial review (didn't go through the design/implementation thoroughly), and leave out the tests for this round. There are multiple changes needed for this to match the design.

When working on the next revision, please make sure that you self-review your PR carefully, since a lot of these issues can be caught with this process.

lakshya1goel:
Hi @PIG208, thanks for the review. I have pushed the update. PTAL. Thanks!

test/widgets/topic_list_test.dart
```diff
@@ -0,0 +1,375 @@
...
+    for (final row in rows) {
+      if (tester.widgetList(find.descendant(
+        of: find.byWidget(row),
+        matching: find.text(label),
+      )).isNotEmpty) {
+        return row;
+      }
+    }
+    return null;
+  }
```
PIG208, line 46: Given that we are always using `find.byWidget` on the result of this, how about just returning a finder?

```suggestion
  Finder findRowByLabel(WidgetTester tester, String label) {
    return find.descendant(
      of: find.byType(TopicListItem),
      matching: find.widgetWithText(Row, label));
  }
```

test/widgets/topic_list_test.dart
```diff
@@ -0,0 +1,375 @@
...
+      await tester.pumpAndSettle();
+
+      check(pushedRoutes).single
+        .isA<MaterialAccountWidgetRoute>()
+        .page.isA<MessageListPage>()
+        .initNarrow.equals(ChannelNarrow(stream.streamId));
+    });
+
+    bool hasIcon(WidgetTester tester, {
+      required Widget? parent,
```
PIG208, line 211: It might cleaner to just pass `label`, since `parent` is always the result from `findRowByLabel`.

test/widgets/topic_list_test.dart
```diff
@@ -0,0 +1,375 @@
...
+      required IconData icon,
+    }) {
+      check(parent).isNotNull();
+      return tester.widgetList(find.descendant(
+        of: find.byWidget(parent!),
+        matching: find.byIcon(icon),
+      )).isNotEmpty;
```
PIG208, line 218: It seems a bit unnecessary to convert the finder result to a `bool`. Let's just return the finder from this helper (we should then rename this too), and use `check(…).findsOne` etc on the result as appropriate.

This way we have more information about the find result when the check fails.

test/widgets/topic_list_test.dart
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    group('mentions', () {
```
PIG208, line 221: Another interesting case for this group would be topic with a mention that is not unread.

PIG208 (commented):
Thanks for the update! I went through the implementation and took a look at the tests. Left some more comments.

I think the various markers (unreads, mentions, topic visibility) are non-trivial to implement, but that do share a lot of commonality with inbox topic items, so I wonder if it's better to figure out a way to share code, at least for the markers, between the topic list items and inbox topic items. @gnprice might have thoughts on this.

An advantage of that would be not needing to maintain a separate set of tests for a very similar feature.

gnprice:
> I think the various markers (unreads, mentions, topic visibility) are non-trivial to implement, but that do share a lot of commonality with inbox topic items, so I wonder if it's better to figure out a way to share code, at least for the markers, between the topic list items and inbox topic items. @gnprice might have thoughts on this.

Yeah, this TopicListItem class clearly has a lot in common with the _TopicItem in the inbox.

At a minimum we should keep the code here as similar as possible to that code, so that it only differs in ways that are deliberately intended because they mean real differences in the user-facing behavior. That way we don't put up unnecessary barriers to ourselves for trying to unify the code later.

After making that revision to isolate the intentional differences, it'd be good to post a list of them, and then we can see which ones we want as differences between the two screens and which ones would be good changes to just make to the inbox too. For example it looks like one difference is that this one presents resolved topics with an icon — that's something that would be a good change for the inbox too.

I think probably in the end what we'll want is to have
* a commit that refactors `_TopicItem` so that most of the work is done by a widget that can be shared with this screen;
* and then a commit that adds this screen and re-uses that widget.

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,326 @@
...
+              child: topic.name.isResolved
+                ? Opacity(
+                  opacity: 0.4,
+                  child: Icon(
+                      ZulipIcons.check,
+                      size: 16,
+                      color: designVariables.textMessage,
+                    ),
+                )
+                : null,
```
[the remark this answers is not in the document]
lakshya1goel: Wouldn't removing the SizedBox around the resolved checkmark icon cause layout misalignment between resolved and unresolved topics?

lib/widgets/topic_list.dart
```diff
@@ -0,0 +1,326 @@
...
+                  ),
+                  const SizedBox(width: 8),
+                  if (hasMention) const _IconMarker(icon: ZulipIcons.at_sign),
+                  if (visibilityIcon != null) _IconMarker(icon: visibilityIcon),
+                  if (hasUnreads) Padding(
+                    padding: const EdgeInsetsDirectional.only(end: 4),
+                    child: UnreadCountBadge(count: unreads,
+                      backgroundColor: designVariables.bgCounterUnread, bold: true),
+                  ),
+                  const SizedBox(width: 8),
```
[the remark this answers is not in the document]
lakshya1goel: We have 8px padding between text and markers thus added the 1st 8px width Sizedbox after the text widget.
Then we have 4px gap between any 2 markers thus added 4px end padding for each marker and at last added this 8px sizedbox so that when any of the marker is there the total padding will be 4px + 8px = 12px.
But yes when there is no marker in that case we are having 8xp + 8px = 16px

lib/widgets/topic_list.dart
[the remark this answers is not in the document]
PIG208: Oh yeah. I meant that we can have either the SizedBox or the Icon. We shouldn't need to wrap the Icon inside a SizedBox.

PIG208:
Hi @lakshya1goel! Do you have any unpushed changes to this draft? Since quite a bit of rewrite is still required to get this merged, we might be taking over this PR to prioritize finishing #1158 sooner, given that it is a launch-blocker issue.

lakshya1goel:
Hey @PIG208, I have pushed the revision. Please have a look.
(Going through some college examinations that's why not able to update it)

gnprice:
Thanks @lakshya1goel for your work on this issue. I think we'll be able to get this to something we're happy to merge much more quickly by having one of us on the core team drive it, so I've asked @PIG208 to take on the issue.

I believe @PIG208 will be sending a PR for this soon — once that's posted, I recommend you take a look at it and compare.

PIG208:
Thank you for your contribution! We will be closing this, but the work will continue in #1500.

PR #1451, opened 2025-03-31
notif: Use live value for app ID on registering APNs token

### LINKED ISSUE #407: notif: Use live value for app ID on registering APNs token

AhmedTareek:
Fixes #407 

Changes made:
- Replaced the hardcoded app bundle ID with `packageInfo.packageName`.
- Added a new test that sets a specific appId via packageInfo and confirms it’s used in a notification request.

The first change caused the "token initially unknown" test in `store_test.dart` to fail due to the added asynchronous call to `ZulipBinding.instance.packageInfo`, which delayed the completion of `registerToken`.

To resolve this, a `flushMicrotasks` was added after `await startFuture` in the test to ensure `_registerNotificationToken` completes before the test checks for the request.

### DIFF
--- a/lib/notifications/receive.dart
+++ b/lib/notifications/receive.dart
@@ -149,8 +149,10 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
-        await addApnsToken(connection, token: token, appid: appBundleId);
+        final packageInfo = await ZulipBinding.instance.packageInfo;
+        await addApnsToken(connection,
+          token: token,
+          appid: packageInfo!.packageName);
 
       case TargetPlatform.linux:
       case TargetPlatform.macOS:
--- a/test/api/core_test.dart
+++ b/test/api/core_test.dart
@@ -460,7 +460,7 @@ void main() {
       });
     }
 
-    const packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+    final packageInfo = eg.packageInfo(version: '0.0.1', buildNumber: '1');
 
     const testCases = [
       ('ZulipFlutter/0.0.1+1 (Android 14)',             AndroidDeviceInfo(release: '14', sdkInt: 34),                      ),
--- a/test/example_data.dart
+++ b/test/example_data.dart
@@ -10,6 +10,7 @@ import 'package:zulip/api/model/submessage.dart';
 import 'package:zulip/api/route/messages.dart';
 import 'package:zulip/api/route/realm.dart';
 import 'package:zulip/api/route/channels.dart';
+import 'package:zulip/model/binding.dart';
 import 'package:zulip/model/database.dart';
 import 'package:zulip/model/narrow.dart';
 import 'package:zulip/model/settings.dart';
@@ -993,4 +994,16 @@ UpdateMachine updateMachine({
     account: account, initialSnapshot: initialSnapshot);
   return UpdateMachine.fromInitialSnapshot(
     store: store, initialSnapshot: initialSnapshot);
+}
+
+PackageInfo packageInfo({
+  String? version,
+  String? buildNumber,
+  String? packageName,
+}) {
+  return PackageInfo(
+    version: version ?? '1.0.0',
+    buildNumber: buildNumber ?? '1',
+    packageName: packageName ?? 'com.example.app',
+  );
 }
--- a/test/model/binding.dart
+++ b/test/model/binding.dart
@@ -260,7 +260,7 @@ class TestZulipBinding extends ZulipBinding {
 
   /// The value that `ZulipBinding.instance.packageInfo` should return.
   PackageInfo packageInfoResult = _defaultPackageInfo;
-  static const _defaultPackageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+  static final _defaultPackageInfo = eg.packageInfo();
 
   void _resetPackageInfo() {
     packageInfoResult = _defaultPackageInfo;
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -1291,6 +1291,7 @@ void main() {
       addTearDown(testBinding.reset);
       testBinding.firebaseMessagingInitialToken = '012abc';
       addTearDown(NotificationService.debugReset);
+      testBinding.packageInfoResult = eg.packageInfo(packageName: 'com.zulip.flutter');
       await NotificationService.instance.start();
 
       // On store startup, send the token.
@@ -1318,6 +1319,7 @@ void main() {
       addTearDown(testBinding.reset);
       testBinding.firebaseMessagingInitialToken = '012abc';
       addTearDown(NotificationService.debugReset);
+      testBinding.packageInfoResult = eg.packageInfo(packageName: 'com.zulip.flutter');
       final startFuture = NotificationService.instance.start();
 
       // TODO this test is a bit brittle in its interaction with asynchrony;
@@ -1336,6 +1338,7 @@ void main() {
       // When the token later appears, send it.
       connection.prepare(json: {});
       await startFuture;
+      async.flushMicrotasks();
       if (defaultTargetPlatform == TargetPlatform.android) {
         checkLastRequestFcm(token: '012abc');
       } else {
@@ -1349,6 +1352,22 @@ void main() {
         async.flushMicrotasks();
         checkLastRequestFcm(token: '456def');
       }
+    }));
+
+    test('on iOS, use provided app ID from packageInfo', () => awaitFakeAsync((async) async {
+      final origTargetPlatform = debugDefaultTargetPlatformOverride;
+      addTearDown(() => debugDefaultTargetPlatformOverride = origTargetPlatform);
+      debugDefaultTargetPlatformOverride = TargetPlatform.iOS;
+      addTearDown(testBinding.reset);
+      testBinding.firebaseMessagingInitialToken = '012abc';
+      testBinding.packageInfoResult = eg.packageInfo(packageName: 'com.example.test');
+      addTearDown(NotificationService.debugReset);
+      await NotificationService.instance.start();
+
+      prepareStore();
+      connection.prepare(json: {});
+      await updateMachine.registerNotificationToken();
+      checkLastRequestApns(token: '012abc', appid: 'com.example.test');
     }));
   });
 
--- a/lib/model/binding.dart
+++ b/lib/model/binding.dart
@@ -310,10 +310,12 @@ class LinuxDeviceInfo implements BaseDeviceInfo {
 class PackageInfo {
   final String version;
   final String buildNumber;
+  final String packageName;
 
   const PackageInfo({
     required this.version,
     required this.buildNumber,
+    required this.packageName,
   });
 }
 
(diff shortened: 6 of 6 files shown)


gnprice:
Thanks for the contribution!

Before we can review this in detail, it will need tests. See the guidance in our README file about writing tests.

AhmedTareek:
Tests are added.

test/model/store_test.dart
```diff
@@ -1191,6 +1194,20 @@ void main() {
         checkLastRequestFcm(token: '456def');
       }
     }));
+
+    testAndroidIos('use default appId if packageInfo is missing', () => awaitFakeAsync((async) async {
```
rajveermalviya, line 1198: This test won't be needed after removing the fallback.

So, instead let's have a test that uses sets different app ID at start (using `testBinding.packageInfoResult = PackageInfo(…)`) and ensure that that app ID is observed in the request.

rajveermalviya (changes requested):
Thanks for working on this @AhmedTareek! Some comments below.

lib/notifications/receive.dart
```diff
@@ -149,9 +149,10 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
-        await addApnsToken(connection, token: token, appid: appBundleId);
-
+        final packageInfo = await ZulipBinding.instance.packageInfo;
+        await addApnsToken(connection,
+            token: token,
+            appid: packageInfo!.packageName);
```
rajveermalviya, line 155: nit: indent
```suggestion
        await addApnsToken(connection,
          token: token,
          appid: packageInfo!.packageName);
```

test/model/binding.dart
```diff
@@ -230,8 +230,9 @@ class TestZulipBinding extends ZulipBinding {
   BaseDeviceInfo? get syncDeviceInfo => deviceInfoResult;
 
   /// The value that `ZulipBinding.instance.packageInfo` should return.
-  PackageInfo packageInfoResult = _defaultPackageInfo;
-  static const _defaultPackageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+  PackageInfo? packageInfoResult = _defaultPackageInfo;
+  static const _defaultPackageInfo = PackageInfo(version: '0.0.1',
+      buildNumber: '1', packageName: 'com.zulip.flutter');
```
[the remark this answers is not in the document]
rajveermalviya: Bump, it should be 2-space indentation.

rajveermalviya (approved):
Thanks for the revision @AhmedTareek! Some nits otherwise looks good, moving to integration review from Greg.

test/model/binding.dart
```diff
@@ -230,8 +230,11 @@ class TestZulipBinding extends ZulipBinding {
   BaseDeviceInfo? get syncDeviceInfo => deviceInfoResult;
 
   /// The value that `ZulipBinding.instance.packageInfo` should return.
-  PackageInfo packageInfoResult = _defaultPackageInfo;
-  static const _defaultPackageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+  PackageInfo? packageInfoResult = _defaultPackageInfo;
+  static const _defaultPackageInfo = PackageInfo(
+    version: '0.0.1',
+    buildNumber: '1',
+    packageName: 'com.zulip.flutter');
```
gnprice, line 237: Having this value here in the central default/example test data is fine and helpful.

It looks like there are a couple of existing tests that would break if this changed, though. That means those tests are now effectively relying implicitly on this data having the particular value that it has — which makes it hard to read those tests and understand what they're doing.

Instead, the tests should make explicit all the details they depend on in the test data they're using, so that the data here in a central location can remain totally boring. See:
https://github.com/zulip/zulip-flutter/pull/1317#discussion_r1970791192
[#frontend > typescript node tests @ 💬](https://chat.zulip.org/#narrow/channel/6-frontend/topic/typescript.20node.20tests/near/1896899)

gnprice: To write those changes to existing tests, you'll probably find it helpful to add a function `eg.packageInfo` in `example_data.dart`. That way a test that only cares about `packageName` can specify only that and let the helper function boringly take care of the other fields.

test/api/core_test.dart
```diff
@@ -460,7 +460,7 @@ void main() {
       });
     }
 
-    const packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+    const packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1', packageName: 'com.zulip.flutter');
```
gnprice, line 463: Similarly, this can look like:
```suggestion
    const packageInfo = eg.packageInfo(version: '0.0.1', buildNumber: '1');
```

so that it doesn't have to make up a value for `packageName`, which this test doesn't care about. (And so that the reader therefore doesn't have to read that value and wonder if and how it affects the test after all.)

gnprice (commented):
Thanks @AhmedTareek for picking this up, and @rajveermalviya for the previous reviews! Comments below.

lib/notifications/receive.dart
```diff
@@ -149,9 +149,10 @@ class NotificationService {
         await addFcmToken(connection, token: token);
 
       case TargetPlatform.iOS:
-        const appBundleId = 'com.zulip.flutter'; // TODO(#407) find actual value live
```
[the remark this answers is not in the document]
gnprice: notif: Use live value for app ID on registering APNs token

    Fixes #407

Closer, but make it a complete sentence:

    notif: Use live value for app ID on registering APNs token

    Fixes #407.

test/api/core_test.dart
```diff
@@ -460,7 +460,7 @@ void main() {
       });
     }
 
-    const packageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+    PackageInfo packageInfo = eg.packageInfo(version: '0.0.1', buildNumber: '1');
```
gnprice, line 463: nit: declare as final (or const) whenever possible

test/model/binding.dart
```diff
@@ -231,7 +231,7 @@ class TestZulipBinding extends ZulipBinding {
 
   /// The value that `ZulipBinding.instance.packageInfo` should return.
   PackageInfo packageInfoResult = _defaultPackageInfo;
-  static const _defaultPackageInfo = PackageInfo(version: '0.0.1', buildNumber: '1');
+  static final PackageInfo _defaultPackageInfo = eg.packageInfo();
```
gnprice, line 234: nit: omit type where it's obvious (as it is here)

(That doesn't apply to mutable fields/variables, because when it's mutable the initializer at the definition site isn't the last word on its value and so it's never obvious from the definition what the type should be. It does apply to things declared final or const.)

test/example_data.dart
```diff
@@ -985,3 +986,15 @@ UpdateMachine updateMachine({
...
+
+PackageInfo packageInfo({
+  String? version,
+  String? buildNumber,
+  String? packageName,
+}) {
+  return PackageInfo(
+    version: version ?? '1.0.0',
+    buildNumber: buildNumber ?? '1',
+    packageName: packageName ?? 'com.package.name',
```
gnprice, line 998: Let's make this value slightly more like a representative value, instead of a description of what it is:
```suggestion
    packageName: packageName ?? 'com.example.app',
```

That also uses a proper RFC 6761 example domain — see https://github.com/zulip/zulip-flutter/pull/1410#discussion_r2006629148 and [#frontend > typescript node tests @ 💬](https://chat.zulip.org/#narrow/channel/6-frontend/topic/typescript.20node.20tests/near/1910932).

test/model/store_test.dart
```diff
@@ -1191,6 +1194,21 @@ void main() {
         checkLastRequestFcm(token: '456def');
       }
     }));
+
+    test('use provided appId from packageInfo', () => awaitFakeAsync((async) async {
```
gnprice, line 1198: nit: a camel-case term like "appId" isn't something that ever appears in normal English capitalization, so it should only be used when it's a proper name — when it's the exact name of something in particular, i.e. usually some identifier in code.

There isn't any identifier `appId` in the code related to this, so that spelling shouldn't appear here either.

Instead, in a plain-English context like a test description, use normal English spacing and capitalization, like "app ID":

```suggestion
    test('use provided app ID from packageInfo', () => awaitFakeAsync((async) async {
```

test/model/store_test.dart
```diff
@@ -1191,6 +1194,21 @@ void main() {
...
+      final origTargetPlatform = debugDefaultTargetPlatformOverride;
+      addTearDown(() => debugDefaultTargetPlatformOverride = origTargetPlatform);
+      debugDefaultTargetPlatformOverride = TargetPlatform.iOS;
+      addTearDown(testBinding.reset);
+      testBinding.firebaseMessagingInitialToken = '012abc';
+      testBinding.packageInfoResult = eg.packageInfo(packageName: 'com.example.test');
+      addTearDown(NotificationService.debugReset);
+      await NotificationService.instance.start();
+      prepareStore();
```
gnprice, line 1207: nit: use a blank line to organize the steps, similar to the neighboring tests

```suggestion
      await NotificationService.instance.start();

      prepareStore();
```

test/model/store_test.dart
gnprice: Separately, the behavior this is testing is specific to iOS, so let's have the test description say that:
```suggestion
    test('on iOS, use provided app ID from packageInfo', () => awaitFakeAsync((async) async {
```

gnprice (commented):
Thanks for the revision! This is close to merge — various small comments below.

gnprice:
Thanks @AhmedTareek for the revision! These changes all look good.

Before merging, I'd like to manually test this end to end, to check that notifications still work as expected. I believe they should — but the changes this PR is making are the kind where an oversight could easily cause notifications to totally break, and without manual testing we likely wouldn't learn that until it went out in the next release.

I tried doing that manual testing just now, but it seems like the instructions we've previously had for manually testing notification changes aren't currently working. So some debugging will be required to work out how to update them. (We'll need that sort of testing to be working again for other changes too, not only this PR.) I started a chat thread for that here:
[#mobile-team > notifications from dev server @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/notifications.20from.20dev.20server/near/2159855)

Once that's sorted out, I can do a round of manual testing on this PR; and provided that comes out clean as expected, we'll then be able to merge it.

gnprice:
OK, I have those instructions all updated, debugged, and working again! For details, see the chat thread linked above.

Then I just turned back to my Mac to try testing this PR, and...
image

It looks like my testing iPhone updated itself to iOS 18.4 in the hour or two between when I tested the updated instructions and now. That's the threshold where I'll need to take a major-version upgrade of macOS in order to connect again to the device.

So I'll do that upgrade soon, but this PR will wait a bit more until I do.

gnprice:
False alarm -- the update in question didn't require a macOS update. (I was misled by [this table](https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_(since_visionOS_support)).) So it finished quickly.

Everything looks good! Notifications work just as well after this change as before. In particular a fresh install successfully registers the device for notifications.

So I'll go ahead and merge.

PR #1452, opened 2025-04-01
KaTeX (1.75/n): Support horizontal offset and pre-req for vertical offsets  

rajveermalviya:
Pre-req for #1698

Related: #46

### DIFF
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -492,6 +492,8 @@ class _KatexParser {
       if (stylesheet.topLevels case [css_visitor.RuleSet() && final rule]) {
         double? heightEm;
         double? verticalAlignEm;
+        double? marginRightEm;
+        double? marginLeftEm;
 
         for (final declaration in rule.declarationGroup.declarations) {
           if (declaration case css_visitor.Declaration(
@@ -507,6 +509,20 @@ class _KatexParser {
               case 'vertical-align':
                 verticalAlignEm = _getEm(expression);
                 if (verticalAlignEm != null) continue;
+
+              case 'margin-right':
+                marginRightEm = _getEm(expression);
+                if (marginRightEm != null) {
+                  if (marginRightEm < 0) throw _KatexHtmlParseError();
+                  continue;
+                }
+
+              case 'margin-left':
+                marginLeftEm = _getEm(expression);
+                if (marginLeftEm != null) {
+                  if (marginLeftEm < 0) throw _KatexHtmlParseError();
+                  continue;
+                }
             }
 
             // TODO handle more CSS properties
@@ -522,6 +538,8 @@ class _KatexParser {
         return KatexSpanStyles(
           heightEm: heightEm,
           verticalAlignEm: verticalAlignEm,
+          marginRightEm: marginRightEm,
+          marginLeftEm: marginLeftEm,
         );
       } else {
         throw _KatexHtmlParseError();
@@ -560,6 +578,9 @@ class KatexSpanStyles {
   final double? heightEm;
   final double? verticalAlignEm;
 
+  final double? marginRightEm;
+  final double? marginLeftEm;
+
   final String? fontFamily;
   final double? fontSizeEm;
   final KatexSpanFontWeight? fontWeight;
@@ -569,6 +590,8 @@ class KatexSpanStyles {
   const KatexSpanStyles({
     this.heightEm,
     this.verticalAlignEm,
+    this.marginRightEm,
+    this.marginLeftEm,
     this.fontFamily,
     this.fontSizeEm,
     this.fontWeight,
@@ -581,6 +604,8 @@ class KatexSpanStyles {
     'KatexSpanStyles',
     heightEm,
     verticalAlignEm,
+    marginRightEm,
+    marginLeftEm,
     fontFamily,
     fontSizeEm,
     fontWeight,
@@ -593,6 +618,8 @@ class KatexSpanStyles {
     return other is KatexSpanStyles &&
       other.heightEm == heightEm &&
       other.verticalAlignEm == verticalAlignEm &&
+      other.marginRightEm == marginRightEm &&
+      other.marginLeftEm == marginLeftEm &&
       other.fontFamily == fontFamily &&
       other.fontSizeEm == fontSizeEm &&
       other.fontWeight == fontWeight &&
@@ -605,6 +632,8 @@ class KatexSpanStyles {
     final args = <String>[];
     if (heightEm != null) args.add('heightEm: $heightEm');
     if (verticalAlignEm != null) args.add('verticalAlignEm: $verticalAlignEm');
+    if (marginRightEm != null) args.add('marginRightEm: $marginRightEm');
+    if (marginLeftEm != null) args.add('marginLeftEm: $marginLeftEm');
     if (fontFamily != null) args.add('fontFamily: $fontFamily');
     if (fontSizeEm != null) args.add('fontSizeEm: $fontSizeEm');
     if (fontWeight != null) args.add('fontWeight: $fontWeight');
@@ -624,6 +653,8 @@ class KatexSpanStyles {
     return KatexSpanStyles(
       heightEm: other.heightEm ?? heightEm,
       verticalAlignEm: other.verticalAlignEm ?? verticalAlignEm,
+      marginRightEm: other.marginRightEm ?? marginRightEm,
+      marginLeftEm: other.marginLeftEm ?? marginLeftEm,
       fontFamily: other.fontFamily ?? fontFamily,
       fontSizeEm: other.fontSizeEm ?? fontSizeEm,
       fontStyle: other.fontStyle ?? fontStyle,
@@ -635,6 +666,8 @@ class KatexSpanStyles {
   KatexSpanStyles filter({
     bool heightEm = true,
     bool verticalAlignEm = true,
+    bool marginRightEm = true,
+    bool marginLeftEm = true,
     bool fontFamily = true,
     bool fontSizeEm = true,
     bool fontWeight = true,
@@ -644,6 +677,8 @@ class KatexSpanStyles {
     return KatexSpanStyles(
       heightEm: heightEm ? this.heightEm : null,
       verticalAlignEm: verticalAlignEm ? this.verticalAlignEm : null,
+      marginRightEm: marginRightEm ? this.marginRightEm : null,
+      marginLeftEm: marginLeftEm ? this.marginLeftEm : null,
       fontFamily: fontFamily ? this.fontFamily : null,
       fontSizeEm: fontSizeEm ? this.fontSizeEm : null,
       fontWeight: fontWeight ? this.fontWeight : null,
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -909,7 +909,7 @@ class _KatexSpan extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final em = DefaultTextStyle.of(context).style.fontSize!;
+    var em = DefaultTextStyle.of(context).style.fontSize!;
 
     Widget widget = const SizedBox.shrink();
     if (node.text != null) {
@@ -929,6 +929,8 @@ class _KatexSpan extends StatelessWidget {
       double fontSizeEm => fontSizeEm * em,
       null => null,
     };
+    if (fontSize != null) em = fontSize;
+
     final fontWeight = switch (styles.fontWeight) {
       KatexSpanFontWeight.bold => FontWeight.bold,
       null => null,
@@ -973,11 +975,28 @@ class _KatexSpan extends StatelessWidget {
         child: widget);
     }
 
-    return SizedBox(
+    widget = SizedBox(
       height: styles.heightEm != null
-        ? styles.heightEm! * (fontSize ?? em)
+        ? styles.heightEm! * em
         : null,
       child: widget);
+
+    final margin = switch ((styles.marginLeftEm, styles.marginRightEm)) {
+      (null, null) => null,
+      (null, final marginRightEm?) =>
+        EdgeInsets.only(right: marginRightEm * em),
+      (final marginLeftEm?, null) =>
+        EdgeInsets.only(left: marginLeftEm * em),
+      (final marginLeftEm?, final marginRightEm?) =>
+        EdgeInsets.only(left: marginLeftEm * em, right: marginRightEm * em),
+    };
+
+    if (margin != null) {
+      assert(margin.isNonNegative);
+      widget = Padding(padding: margin, child: widget);
+    }
+
+    return widget;
   }
 }
 
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -884,6 +884,65 @@ class ContentExample {
         ]),
     ]);
 
+  static const mathBlockKatexSpace = ContentExample(
+    'math block; KaTeX space',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2214883
+    '```math\n1:2\n```',
+    '<p>'
+      '<span class="katex-display"><span class="katex">'
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mn>1</mn><mo>:</mo><mn>2</mn></mrow><annotation encoding="application/x-tex">1:2</annotation></semantics></math></span>'
+        '<span class="katex-html" aria-hidden="true">'
+          '<span class="base">'
+            '<span class="strut" style="height:0.6444em;"></span>'
+            '<span class="mord">1</span>'
+            '<span class="mspace" style="margin-right:0.2778em;"></span>'
+            '<span class="mrel">:</span>'
+            '<span class="mspace" style="margin-right:0.2778em;"></span></span>'
+          '<span class="base">'
+            '<span class="strut" style="height:0.6444em;"></span>'
+            '<span class="mord">2</span></span></span></span></span></p>', [
+      MathBlockNode(
+        texSource: '1:2',
+        nodes: [
+          KatexSpanNode(
+            styles: KatexSpanStyles(),
+            text: null,
+            nodes: [
+              KatexStrutNode(
+                heightEm: 0.6444,
+                verticalAlignEm: null),
+              KatexSpanNode(
+                styles: KatexSpanStyles(),
+                text: '1',
+                nodes: null),
+              KatexSpanNode(
+                styles: KatexSpanStyles(marginRightEm: 0.2778),
+                text: null,
+                nodes: []),
+              KatexSpanNode(
+                styles: KatexSpanStyles(),
+                text: ':',
+                nodes: null),
+              KatexSpanNode(
+                styles: KatexSpanStyles(marginRightEm: 0.2778),
+                text: null,
+                nodes: []),
+            ]),
+          KatexSpanNode(
+            styles: KatexSpanStyles(),
+            text: null,
+            nodes: [
+              KatexStrutNode(
+                heightEm: 0.6444,
+                verticalAlignEm: null),
+              KatexSpanNode(
+                styles: KatexSpanStyles(),
+                text: '2',
+                nodes: null),
+            ]),
+        ]),
+    ]);
+
   static const imageSingle = ContentExample(
     'single image',
     // https://chat.zulip.org/#narrow/stream/7-test-here/topic/Thumbnails/near/1900103
@@ -1973,6 +2032,7 @@ void main() async {
   testParseExample(ContentExample.mathBlockKatexSizing);
   testParseExample(ContentExample.mathBlockKatexNestedSizing);
   testParseExample(ContentExample.mathBlockKatexDelimSizing);
+  testParseExample(ContentExample.mathBlockKatexSpace);
 
   testParseExample(ContentExample.imageSingle);
   testParseExample(ContentExample.imageSingleNoDimensions);
--- a/test/widgets/content_test.dart
+++ b/test/widgets/content_test.dart
@@ -597,6 +597,11 @@ void main() {
           ('⌈', Offset(27.12, 20.14), Size(11.99, 25.00)),
           ('⌊', Offset(39.11, 20.14), Size(13.14, 25.00)),
         ]),
+        (ContentExample.mathBlockKatexSpace, skip: false, [
+          ('1', Offset(0.00, 2.24), Size(10.28, 25.00)),
+          (':', Offset(16.00, 2.24), Size(5.72, 25.00)),
+          ('2', Offset(27.43, 2.24), Size(10.28, 25.00)),
+        ]),
       ];
 
       for (final testCase in testCases) {


lib/model/content.dart
```diff
@@ -341,22 +342,124 @@ class CodeBlockSpanNode extends ContentNode {
...
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return nodes?.map((node) => node.toDiagnosticsNode()).toList() ?? const [];
+  }
+}
+
+sealed class KatexNode extends ContentNode {
+  const KatexNode({super.debugHtmlNode});
+}
+
+class KatexSpanNode extends KatexNode {
```
gnprice, line 376: One point that I noticed while developing #1478 (and checking how my draft of that branch interacted with this PR): it'd be good for this commit:
e268041ca content: Handle vertical offset spans in KaTeX content

to get split up like so:
* First, an NFC prep commit introduces the distinction between KatexNode and KatexSpanNode. At this stage, the latter is the only subclass of the former.
* Then another commit makes the substantive changes this commit is about, including introducing the sibling subclasses KatexVlistNode and KatexVlistRowNode.

One reason that'd be useful is that the split between KatexNode and KatexSpanNode is somewhat nontrivial in itself: some of the references to KatexNode continue to say KatexNode, while others switch to saying KatexSpanNode, so the commit is expressing meaningful information by its choices of which references go which way.

gnprice:
It seems like this had slipped through the cracks :-) but it looks ready for review, so I applied the label.

Also rebased now that #1478 is merged, so this now contains only the changes that are specific to this PR.

lib/widgets/content.dart
```diff
@@ -850,8 +860,7 @@ class _Katex extends StatelessWidget {
     return Directionality(
       textDirection: TextDirection.ltr,
       child: DefaultTextStyle(
-        style: kBaseKatexTextStyle.copyWith(
-          color: ContentTheme.of(context).textStylePlainParagraph.color),
+        style: mkBaseKatexTextStyle(textStyle),
```
PIG208, line 863: This needs some tests, maybe with cases where the base font size is not `kBaseFontSize`?

lib/model/katex.dart
```diff
@@ -136,6 +154,110 @@ class _KatexParser {
...
 
+    final spanClasses = List<String>.unmodifiable(element.className.split(' '));
+
+    if (element case dom.Element(localName: 'span', :final className)
+        when className.startsWith('vlist')) {
+      switch (element) {
+        case dom.Element(
+          localName: 'span',
+          className: 'vlist-t',
+          attributes: final attributesVlistT,
```
PIG208, line 165: I wonder what "t" from "vlist-t" stands for. The "r" from "vlist-r" I assume is for "row" from the naming the `KatexVlistRowNode`.

lib/model/content.dart
```diff
@@ -402,6 +406,63 @@ class KatexNode extends ContentNode {
...
+    this.nodes = const [],
+  });
+
+  final double verticalOffsetEm;
+  final List<KatexNode> nodes;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(StringProperty('verticalOffsetEm', '$verticalOffsetEm'));
```
PIG208, line 435: Looks like we are doing this to `marginRightEm` as well, what's the reason that we don't use `DoubleProperty` with them?

lib/model/katex.dart
```diff
@@ -136,6 +154,110 @@ class _KatexParser {
   KatexNode _parseSpan(dom.Element element) {
     // TODO maybe check if the sequence of ancestors matter for spans.
 
+    final spanClasses = List<String>.unmodifiable(element.className.split(' '));
```
PIG208, line 157: nit: move this to where we might start to use it, since it looks like vlist handling doesn't require `spanClasses`

lib/widgets/content.dart
```diff
@@ -939,7 +952,75 @@ class _KatexSpan extends StatelessWidget {
...
+          children: List.unmodifiable(row.nodes.map((e) {
+            return WidgetSpan(
+              alignment: PlaceholderAlignment.baseline,
+              baseline: TextBaseline.alphabetic,
+              child: switch (e) {
+                KatexSpanNode() => _KatexSpan(e),
+                KatexVlistNode() => _KatexVlist(e),
+                KatexNegativeMarginNode() => _KatexNegativeMargin(e),
+              });
+          })))));
```
PIG208, line 997: looks like this can be replaced with `_KatexNodeList(nodes: row.nodes)`

lib/widgets/content.dart
```diff
@@ -939,7 +952,75 @@ class _KatexSpan extends StatelessWidget {
...
+        children: List.unmodifiable(node.nodes.map((e) {
+          return WidgetSpan(
+            alignment: PlaceholderAlignment.baseline,
+            baseline: TextBaseline.alphabetic,
+            child: switch (e) {
+              KatexSpanNode() => _KatexSpan(e),
+              KatexVlistNode() => _KatexVlist(e),
+              KatexNegativeMarginNode() => _KatexNegativeMargin(e),
+            });
+        })))));
```
PIG208, line 1023: `_KatexNodeList` also seems helpful here.

lib/widgets/content.dart
```diff
@@ -939,7 +952,75 @@ class _KatexSpan extends StatelessWidget {
...
+                KatexVlistNode() => _KatexVlist(e),
+                KatexNegativeMarginNode() => _KatexNegativeMargin(e),
+              });
+          })))));
+    })));
+  }
+}
+
+class _KatexNegativeMargin extends StatelessWidget {
+  const _KatexNegativeMargin(this.node);
```
PIG208, line 1003: Because the name implies that the margin should be negative, maybe we should also add assertions to ensure that that is true.

lib/model/katex.dart
```diff
@@ -121,13 +123,29 @@ class _KatexParser {
...
+          resultSpans = [];
+          resultSpans.add(KatexNegativeMarginNode(
+            marginRightEm: marginRightEm,
+            nodes: previousSpansReversed));
+        }
+      }
+    }
+
+    return resultSpans.reversed.toList(growable: false);
   }
```
PIG208, line 149: Neat solution! A bit tricky to understand. I think we can use `QueueList` (and with `.addFirst`) to avoid the hassle of reversing and unreversing the list.

PIG208: It will be helpful to have tests for this.

lib/model/katex.dart
```diff
@@ -136,6 +154,110 @@ class _KatexParser {
...
+            !attributesVlistR.containsKey('style'):
+          // TODO vlist element should only have `height` style, which we ignore.
+
+          var styles = _parseSpanInlineStyles(innerSpan)!;
+          final topEm = styles.topEm ?? 0;
+
+          final pstrutStyles = _parseSpanInlineStyles(pstrutSpan)!;
+          final pstrutHeight = pstrutStyles.heightEm ?? 0;
+
+          // TODO handle negative right-margin inline style on row nodes.
```
PIG208, line 197: Do we consider this complete after the final commit ("content: Support negative right-margin on KaTeX spans")? I'm not too sure if something else is left to be done here.

This seems to contradict with the earlier assumption that vlist elements contain only the `height` style.

PIG208 (commented):
Thanks for the PR! I think the last few commits will need some tests for the more complicated parts. I went over the changes and left some comments, but haven't extensively tested the changes manually yet.

lib/model/katex.dart
gnprice: I think the "t" is probably for "table". In the CSS, the `.vlist-t` is the element that gets `display: inline-table`:
https://github.com/KaTeX/KaTeX/blob/2fe1941b/src/styles/katex.scss#L183-L184
so that its internal layout is that of a table.

lib/widgets/content.dart
PIG208: Not entirely sure why, but math nested in headings doesn't seem to be visible:

PIG208: Another issue is that the math content seems to get scaled more than normal text, and the smaller the text is (in normal font size), the more scaled it gets.

| normal | larger font |
| - | - |
| image | image |

lib/model/katex.dart
```diff
@@ -136,6 +133,83 @@ class _KatexParser {
...
+
+                final pstrutStyles = _parseSpanInlineStyles(pstrutSpan)!;
+                final pstrutHeight = pstrutStyles.heightEm ?? 0;
+
+                rows.add(KatexVlistRowNode(
+                  verticalOffsetEm: topEm + pstrutHeight,
+                    node: KatexSpanNode(
+                      styles: styles,
+                      text: null,
+                      nodes: _parseChildSpans(otherSpans))));
```
PIG208, line 195: nit: indentation

lib/widgets/content.dart
```diff
@@ -940,7 +951,64 @@ class _KatexSpan extends StatelessWidget {
         textAlign: textAlign,
         child: widget);
     }
-    return widget;
+
+    if (styles.verticalAlignEm != null && styles.heightEm != null) {
+      assert(widget == defaultWidget);
```
PIG208, line 956: I think this assertion is based on the broader assumption that some inline styles never co-occur to `vertical-align` or `height`. This seems like something we can comment on.

PIG208 (commented):
Thanks! Left some comments.

While testing this, I ran into some issues with text scaling (the relevant commit is "content: Scale inline KaTeX content based on the surrounding text"); posted screenshots in one of the comments below.

Otherwise, this works pretty well!

lib/widgets/content.dart
gnprice: > the math content seems to get scaled more than normal text

This is the same underlying issue as #735: it's basically a bug in WidgetSpan.

@chrisbobbe is looking at sending a PR upstream to fix that issue. In the meantime, I think we can work around it here with something like this (untested) in _KatexNodeList:
```diff
         return WidgetSpan(
           alignment: PlaceholderAlignment.baseline,
           baseline: TextBaseline.alphabetic,
-          child: switch (e) {
-            KatexSpanNode() => _KatexSpan(e),
-            KatexVlistNode() => _KatexVlist(e),
-          });
+          child: MediaQuery(
+            data: MediaQueryData(textScaler: TextScaler.noScaling),
+            child: switch (e) {
+              KatexSpanNode() => _KatexSpan(e),
+              KatexVlistNode() => _KatexVlist(e),
+            }));
```

test/widgets/content_test.dart
```diff
@@ -616,17 +622,21 @@ void main() {
...
 
       fontSize = 4.976 * fontSize;
       checkKatexText(tester, '2',
         fontFamily: 'KaTeX_Main',
         fontSize: fontSize,
-        fontHeight: kBaseKatexTextStyle.height!);
+        fontHeight: baseTextStyle.height!);
     });
 
     testWidgets('displays KaTeX content with different delimiter sizing', (tester) async {
```
gnprice, line 642:
```
+    // TODO: Re-enable this test after adding support for parsing
+    //       `vertical-align` in inline styles. Currently it fails
+    //       because `strut` span has `vertical-align`.
+    //
+    // testWidgets('displays KaTeX content with different delimiter sizing', (tester) async {
```

Instead of commenting out the whole test (which makes the diff big), pass `skip: true`. For examples, see `git grep -A2 skip: test/`.

The other reason `skip:` is better than commenting out is that it makes the test case still exist in the test framework and get counted; the total number of skipped tests is printed after the number of passed and failed tests. That's helpful for occasionally sweeping to confirm we don't have tests accidentally left skipped that might point to live bugs.

test/model/content_test.dart
```diff
@@ -727,59 +727,56 @@ class ContentExample {
...
+              KatexSpanNode(
                 styles: KatexSpanStyles(fontSizeEm: 0.5), // .reset-size6.size1
                 text: '0',
                 nodes: null),
             ]),
         ]),
     ]);
 
-  static final mathBlockKatexNestedSizing = ContentExample(
+  static const mathBlockKatexNestedSizing = ContentExample(
```
gnprice, line 779:
```
+  // TODO: Re-enable this test after adding support for parsing
+  //       `vertical-align` in inline styles. Currently it fails
+  //       because `strut` span has `vertical-align`.
+  //
+  // static const mathBlockKatexDelimSizing = ContentExample(
```

Similarly, use `skip: true` instead of commenting.

For this test, that requires a bit of setup to make this file's helpers and the "all content examples are tested" check handle the `skip:` argument. I've just pushed a small added commit at the end of this PR branch to do that:
f9a79d875 content test [nfc]: Enable skips in testParseExample and testParse

So please rebase that to earlier in the branch as needed, and then write:
```dart
  testParseExample(ContentExample.mathBlockKatexNestedSizing, skip: true);
```

rajveermalviya:
This is ready for review again @chrisbobbe, PTAL.

gnprice:
To simplify managing this series of changes, I'd like to get the parts of it merged soon that are easier to merge, while we follow up with reviews on the rest.

@rajveermalviya would you try reordering this branch so the commits that are refactors or otherwise simple come first, and the more complex commits come at the end? On a quick scan, I think that means these 9 commits first:

d9b1be2d6 content test [nfc]: Use const for math block tests
e0e3ef703 content test [nfc]: Enable skips in testParseExample and testParse
f04faede5 content [nfc]: Inline _logError in _KatexParser._parseSpan
af7d9df9a content [nfc]: Refactor _KatexParser._parseChildSpans to take list of nodes
cfaf3d2df content: Populate `debugHtmlNode` for KatexNode
89dfe0076 content [nfc]: Reintroduce KatexNode as a base sealed class
20b6c93c3 content: Ignore more KaTeX classes that don't have CSS definition
92effd05e content: Handle 'mspace' and 'msupsub' KaTeX CSS classes
bef406885 content [nfc]: Remove the `inline` property in _Katex widget

and so these 6 later:

3e80a0d5e content: Scale inline KaTeX content based on the surrounding text
847c91450 content: Support parsing and handling inline styles for KaTeX content
2bf293daa content: Handle 'strut' span in KaTeX content
ef13eff7b content: Handle positive margin-right and margin-left in KaTeX spans
0341f67b9 content: Handle 'top' property in KaTeX span inline style
13d3101e6 content: Handle vertical offset spans in KaTeX content

Then I guess send the first segment as a separate "1.5/n" PR, so it can be merged first.

rajveermalviya:
I've created #1609 which includes the above 9 commits you've listed  + 0bf40b845 content: Support parsing and handling inline styles for KaTeX content.

As that last commit is needed for #1600 (survey script) to list unsupported inline CSS properties.

lib/widgets/content.dart
```diff
@@ -870,15 +876,21 @@ class _KatexNodeList extends StatelessWidget {
         return WidgetSpan(
           alignment: PlaceholderAlignment.baseline,
           baseline: TextBaseline.alphabetic,
-          child: _KatexSpan(e));
+          child: MediaQuery(
+            data: MediaQueryData(textScaler: TextScaler.noScaling),
```
gnprice, line 896: On the face of it this looks like it's doing something wrong: causing the system text-size setting to have no effect on math expressions.

This bit of code is borrowed from my suggestion at https://github.com/zulip/zulip-flutter/pull/1452#discussion_r2109846601. I think it's likely correct — there's a bug (same as at #735) causing that system setting to get applied repeatedly, compounding, and this is to prevent that. But it definitely looks wrong on its face, so it calls for an explanation, probably in a code comment.

Have you tried out how this looks with that system setting applied? Does it end up correctly applying the setting once, rather than zero times or multiple times?

Can this be done as a separate commit before the rest of this commit? The commit isn't obviously about the same thing:
04348639a content: Scale inline KaTeX content based on the surrounding text

(BTW I didn't actually realize until just now, as I got into reviewing this PR closely, that you'd done something to address that issue Zixuan discovered in that previous thread. I'd assumed we still had in the recent releases that issue where the system font scaling was being applied repeatedly in math expressions, and had just lucked out in that nobody had complained yet. For something complex like this it's a good idea to reply in the GitHub subthread when you apply a suggested fix; the suggestion was untested and therefore tentative, so I was figuring you'd report back your findings on whether it turned out to work.)

lib/widgets/content.dart
```diff
@@ -1263,7 +1348,7 @@ class _InlineContentBuilder {
           : WidgetSpan(
               alignment: PlaceholderAlignment.baseline,
               baseline: TextBaseline.alphabetic,
-              child: _Katex(inline: true, nodes: nodes));
+              child: Katex(textStyle: widget.style, nodes: nodes));
```
gnprice: 04348639a content: Scale inline KaTeX content based on the surrounding text

Let's have a test case for this, too — one that would fail if we re-introduced the bug Zixuan found at https://github.com/zulip/zulip-flutter/pull/1452#discussion_r2098947592.

lib/model/katex.dart
```diff
@@ -136,6 +133,123 @@ class _KatexParser {
...
+        final styles = _parseSpanInlineStyles(element);
+        if (styles == null) throw KatexHtmlParseError();
+
+        final heightEm = styles.heightEm;
+        if (heightEm == null) throw KatexHtmlParseError();
+        final verticalAlignEm = styles.verticalAlignEm;
+
+        // Ensure only `height` and `vertical-align` inline styles are present.
+        if (styles.filter(heightEm: false, verticalAlignEm: false) !=
+            KatexSpanStyles()) {
```
gnprice: nit: binary operator goes after newline, not before (so it doesn't get visually lost off at the ragged right edge of the code):
```suggestion
        if (styles.filter(heightEm: false, verticalAlignEm: false)
            != KatexSpanStyles()) {
```

gnprice: This will allocate two fresh KatexSpanStyles objects, with all their fields, in order to check that the other fields of `styles` are all null.

We can eliminate one of those allocations easily by saying `const KatexSpanStyles()`.

The other one is from `filter`. Let's leave that for now, I guess; it's probably not a big deal in practice.

(Definitely it's good to be doing this check, excluding unanticipated shapes of the tree.)

lib/model/katex.dart
```diff
@@ -136,6 +133,123 @@ class _KatexParser {
...
+          throw KatexHtmlParseError();
+        }
+
+        return KatexStrutNode(
+          heightEm: heightEm,
+          verticalAlignEm: verticalAlignEm);
```
gnprice: nit: should get debugHtmlNode

lib/model/katex.dart
```diff
@@ -136,6 +133,123 @@ class _KatexParser {
...
+                  throw KatexHtmlParseError('unexpected CSS class for '
+                    'vlist inner span: ${innerSpan.className}');
+                }
+
+                var styles = _parseSpanInlineStyles(innerSpan)!;
+                final topEm = styles.topEm ?? 0;
+
+                styles = styles.filter(topEm: false);
+
+                final pstrutStyles = _parseSpanInlineStyles(pstrutSpan)!;
```
gnprice: Looks like these need the same check that happens at the other _parseSpanInlineStyles call site below, to rule out vertical-align.

lib/model/katex.dart
```diff
@@ -136,6 +133,123 @@ class _KatexParser {
...
+                final pstrutHeight = pstrutStyles.heightEm ?? 0;
```
gnprice: It looks like we wouldn't notice if this span had additional inline style properties we weren't acting on. So it should get the same check we have in the `.strut` case above.

lib/model/katex.dart
```diff
@@ -136,6 +133,123 @@ class _KatexParser {
   KatexNode _parseSpan(dom.Element element) {
     // TODO maybe check if the sequence of ancestors matter for spans.
 
+    if (element.className.startsWith('strut')) {
+      if (element.className == 'strut' && element.nodes.isEmpty) {
```
gnprice: Is the `startsWith` logic necessary? Seems like we could just skip to the equality check.

lib/widgets/content.dart
```diff
@@ -941,7 +953,80 @@ class _KatexSpan extends StatelessWidget {
...
+    if (verticalAlignEm == null) {
+      return SizedBox(height: node.heightEm * em);
+    }
+
+    return SizedBox(
+      height: node.heightEm * em,
+      child: Baseline(
+        baseline: (verticalAlignEm + node.heightEm) * em,
+        baselineType: TextBaseline.alphabetic,
+        child: const Text('')),
```
gnprice, line 1022: Can we write a test for this?

What's an example of an expression that would look wrong if we left out this Baseline widget; and what would be the effect on how it looks?

lib/widgets/content.dart
```diff
@@ -941,7 +953,80 @@ class _KatexSpan extends StatelessWidget {
         textAlign: textAlign,
         child: widget);
     }
-    return widget;
+
+    final marginRight = switch (styles.marginRightEm) {
```
gnprice: I believe we're assuming here (in _KatexSpan.build) that `styles.verticalAlignEm` will always be null.

That assumption is meant to be guaranteed by the parser. But it's a very non-local fact — it's not something one can tell by looking at this code here. So let's make it explicit here, with an assert.

That way a reader of this code, if they go consult the list of fields on KatexSpanStyles to compare to what's handled here, doesn't think "wait, what about verticalAlignEm? looks like this has a bug that it's not handling that", and then spend a bunch of time trying to exercise that bug before discovering it can't be exercised.

That way also, if there's a bug where that situation *can* happen after all — for example because the parser is missing a spot where it should check `verticalAlignEm` is null, like in the vlist code as commented above :slightly_smiling_face: — then we have a way to potentially notice that.

lib/widgets/content.dart
```diff
@@ -941,7 +953,80 @@ class _KatexSpan extends StatelessWidget {
...
+        margin += EdgeInsets.only(left: marginLeft);
+      }
+    }
+
+    return Container(
+      margin: margin,
+      height: styles.heightEm != null
+        ? styles.heightEm! * em
+        : null,
+      transform: styles.topEm != null
```
gnprice: It's not obvious to me how these three Container fields interact with each other. Let's use individual widgets for these different effects — that way the nesting is clear.

(This is a general problem with the API of Container. We don't have many use sites of it, and probably most of those would be improved by using individual widgets explicitly. I think its main value is for quick hacks and demos.)

lib/widgets/content.dart
```diff
@@ -941,7 +953,80 @@ class _KatexSpan extends StatelessWidget {
...
+      margin = EdgeInsets.zero;
+      if (marginRight != null) {
+        assert(marginRight >= 0);
+        margin += EdgeInsets.only(right: marginRight);
+      }
+      if (marginLeft != null) {
+        assert(marginLeft >= 0);
+        margin += EdgeInsets.only(left: marginLeft);
+      }
+    }
```
gnprice: Is this equivalent?
```suggestion
    final margin = switch ((styles.marginLeftEm, styles.marginRightEm)) {
      (null, null) => null,
      (null, final marginRightEm) => EdgeInsets.only(right: marginRightEm! * em),
      (final marginLeftEm, null) => EdgeInsets.only(left: marginLeftEm! * em),
      (final marginLeftEm, final marginRightEm) =>
        EdgeInsets.only(left: marginLeftEm! * em, right: marginRightEm! * em),
    };
```

lib/widgets/content.dart
```diff
@@ -941,7 +953,80 @@ class _KatexSpan extends StatelessWidget {
...
+    EdgeInsets? margin;
+    if (marginRight != null || marginLeft != null) {
+      margin = EdgeInsets.zero;
+      if (marginRight != null) {
+        assert(marginRight >= 0);
+        margin += EdgeInsets.only(right: marginRight);
+      }
+      if (marginLeft != null) {
+        assert(marginLeft >= 0);
+        margin += EdgeInsets.only(left: marginLeft);
```
gnprice: Can we write a test for this? (Same questions as for struts.)

lib/widgets/content.dart
```diff
@@ -941,7 +953,80 @@ class _KatexSpan extends StatelessWidget {
...
+    }
+
+    return Container(
+      margin: margin,
+      height: styles.heightEm != null
+        ? styles.heightEm! * em
+        : null,
+      transform: styles.topEm != null
+        ? Matrix4.translationValues(0, styles.topEm! * em, 0)
+        : null,
```
gnprice: Huh — so in particular it looks like `top: 0` would have no effect.

Is that true in CSS that `top: 0` has no effect? Are there assumptions we're making about the structure of the HTML, or about the other CSS around, that makes that true?

I'd like to see an example of the sort of content that this appears in. A parser example would be a good start. Then ultimately, like for struts and margins as mentioned in previous comments in this batch, it'd be good to have a widget test that checks that this logic is doing the right thing.

gnprice (commented):
OK, in addition to the commits from #1609 (reviewed there), I've now read the first 4/6 of these commits:
04348639a content: Scale inline KaTeX content based on the surrounding text
d81008f1b content: Handle 'strut' span in KaTeX content
532b3244f content: Handle positive margin-right and margin-left in KaTeX spans
fef8a60ed content: Handle 'top' property in KaTeX span inline style

Comments below.

With that, left for another time are the last two commits, about vlists:
2c5a28cd9 content: Handle vertical offset spans in KaTeX content
7d76da76b content: Error message for unexpected CSS class in vlist inner span

lib/widgets/content.dart
rajveermalviya: The widget side support for this was only added because the parser started to handle it for vlists.

From the corpus data, looks like there is only one prominent case where a span outside `vlist` has `top` inline style, i.e a span with `delimcenter` class. One example for that is: `$$ \left(1\right) $$`.

We don't support `delimcenter` class currently, but that should be trivial to fix as it seems to be just:
```css
.delimcenter {
  position: relative;
}
```

Should I add a commit that includes a fix for that along with a test in this PR? Or should the parser throw for now if it encounters a `top` outside of vlist?

lib/widgets/content.dart
rajveermalviya: Added a test for `margin-right`. From the corpus data, `margin-left` seems to be only present inside a vlist.

lib/widgets/content.dart
rajveermalviya: The (new) offset/size widget test  for `mathBlockKatexDelimSizing` example covers this.

If the Baseline widget is left out, the Katex content height would be larger (and incorrect) and the characters won't be on the correct alignment and may overflow the message container. For example:

| without Baseline | with Baseline |
| - | - |
| image | image |

lib/widgets/content.dart
gnprice: Cool. Let's keep that added functionality to a separate commit.

gnprice: … err, and to answer your actual question, keep it to a later PR :slightly_smiling_face:

lib/widgets/content.dart
gnprice: Great, that'll do, then.

I see this commit now includes a change to that test's expectations, reflecting that effect:
d29fd4dc9 content: Handle 'strut' span in KaTeX content

Can you adjust the style of the test either before or after that commit so that the test isn't also changing style at the same time? That way the change to the expectations will stand out and make clear that the commit is effectively testing this change.

gnprice: One way to do it that works before this commit and remains clean after it might be with `example` and `characters` as positional items in the record but `skip` named:
```dart
      const testCases = <(ContentExample, List<(String, Offset, Size)>, {bool? skip})>[
        (ContentExample.mathBlockKatexSizing, skip: false, [
          ('1', Offset(0.00, 2.24), Size(25.59, 61.00)),
          // …
```

Then this commit would change a `skip: true` to `skip: false`, and perhaps change some of the expected numbers.

lib/widgets/content.dart
rajveermalviya: Yes this workaround does indeed fix the bug in most cases, but there is still some wierdness on Android.

Here's the test message I'd used before to test this, but with all the different font scale settings on iOS and Android.

| | 1 | 2 | 3 | 4 | 5 | 6 |
| - | - | - | - | - | - | - |
| iOS | image | image | image | image |
| Android | image | image | image | image | image | image |

The table labels indicate the steps of font scaling available on the system, where 1 is default.

As can be seen that this workaround works correctly on iOS. But on Android it looks like the font is not playing well after the second step for some reason.

lib/widgets/content.dart
rajveermalviya: Updated.

lib/widgets/content.dart
gnprice: Huh — very interesting that it's different between the two platforms.

Please go ahead and file a follow-up issue, then, for the fact that font scaling doesn't interact correctly with our KaTeX support. That table of screenshots would be great to include there. Then this workaround can get a comment pointing at that issue, to give the reader more context.

lib/widgets/content.dart
rajveermalviya: Added a widget test for KaTeX content in a quote for this in 0a4b6a660ce90e3daa2aa9f3152f3e17b8fb79f8.

rajveermalviya:

lib/widgets/content.dart
```diff
@@ -820,42 +820,49 @@ class MathBlock extends StatelessWidget {
           children: [TextSpan(text: node.texSource)])));
     }
 
-    return _Katex(inline: false, nodes: nodes);
+    return Center(
+      child: SingleChildScrollViewWithScrollbar(
+        scrollDirection: Axis.horizontal,
+        child: Katex(
```
gnprice: Same comment as at https://github.com/zulip/zulip-flutter/pull/1609#discussion_r2181723874 , from when this commit was in a previous version of #1609.

(But now in this PR presumably I'll see what motivated trying this change.)

lib/widgets/content.dart
```diff
@@ -820,42 +820,49 @@ class MathBlock extends StatelessWidget {
...
+    height: 1.2,
+    fontWeight: FontWeight.normal,
+    fontStyle: FontStyle.normal);
+}
 
-class _Katex extends StatelessWidget {
-  const _Katex({
-    required this.inline,
+@visibleForTesting
+class Katex extends StatelessWidget {
```
gnprice: nit: this name feels a bit general for the public namespace (and fortunately it has only a few mentions, so there's minimal cost to a little verbosity)
```suggestion
@visibleForTesting
class KatexWidget extends StatelessWidget {
```

lib/widgets/content.dart
```diff
@@ -872,9 +879,17 @@ class _KatexNodeList extends StatelessWidget {
         return WidgetSpan(
           alignment: PlaceholderAlignment.baseline,
           baseline: TextBaseline.alphabetic,
-          child: switch (e) {
-            KatexSpanNode() => _KatexSpan(e),
-          });
+          // Workaround a bug where text inside a WidgetSpan could be scaled
```
gnprice: nit: "work around" for the verb

lib/widgets/content.dart
gnprice: Huh, does that do it? I wouldn't expect a quote to interact with this bug — those don't change the text size.

(Rereading, I'm not sure that comment by Zixuan was about the bug this fixes, rather than a different issue. What I'm hoping for in this comment is a test case that would fail if we didn't scale inline TeX to match the surrounding font size. Put another way, a test case that matches what's in the commit message:

    content: Scale inline KaTeX content based on the surrounding text
    
    This applies the correct font scaling if the KaTeX content is
    inside a header.

)

test/widgets/content_test.dart
```diff
@@ -572,98 +573,91 @@ void main() {
...
-        final fontSize = katexNode.styles.fontSizeEm! * kBaseKatexTextStyle.fontSize!;
-        checkKatexText(tester, katexNode.text!,
-          fontFamily: 'KaTeX_Main',
-          fontSize: fontSize,
-          fontHeight: kBaseKatexTextStyle.height!);
+    group('characters render at specific offsets with specific size', () {
+      const testCases = <(ContentExample, List<(String, Offset, Size)>, {bool? skip})>[
+        (ContentExample.mathBlockInQuote, skip: false, [
+          ('λ', Offset(0.00, 2.24), Size(11.99, 25.00)),
+        ]),
```
gnprice: (cont'd from https://github.com/zulip/zulip-flutter/pull/1452#discussion_r2179103903)

This sort of test may work in that it breaks if the bug is re-introduced — but then the next key thing a test needs to do, when it breaks due to a regression, is to convince that future developer that there really is a regression, and to help them work out what it is. If it doesn't convince them, they're most likely to just update the test to expect the new values, and the test didn't succeed in preventing the regression.

So a test is most effective if it can tell a crisp story about what it's checking, and why its expectations really should be expected. These test cases centered on a list of numbers typically aren't good at that, though they can still be the best available strategy for some things.

For this test case, the potential regression we're looking for is that the font size doesn't stay in sync with the font size of the surrounding text. This test file has a helper function for just that sort of check: `checkFontSizeRatio`. Can you write a test using that? Then that would make it a lot clearer what the narrative is.

lib/widgets/content.dart
gnprice: … Also a test that checks a given commit works correctly should be squashed into the same commit :slightly_smiling_face: (or sometimes should come before and get updated in the main commit), rather than come after it.

test/widgets/content_test.dart
```diff
@@ -572,98 +573,91 @@ void main() {
...
+        (ContentExample.mathBlockKatexSizing, skip: false, [
+          ('1', Offset(0.00, 2.24), Size(25.59, 61.00)),
+          ('2', Offset(25.59, 10.04), Size(21.33, 51.00)),
+          ('3', Offset(46.91, 16.55), Size(17.77, 43.00)),
+          ('4', Offset(64.68, 21.98), Size(14.80, 36.00)),
```
gnprice, line 582: Huh, it's interesting that in this commit:

    content: Scale inline KaTeX content based on the surrounding text
    
    This applies the correct font scaling if the KaTeX content is
    inside a header.

the changes that happen in the tests are entirely about changing the vertical *offsets* of these characters — no changes to the sizes, nor the horizontal offsets. Do you understand why the vertical offsets are changing when the sizes aren't?

In any case, the lack of size-related test changes seems to confirm that this commit (which is about getting the sizes right) still needs testing.

lib/widgets/content.dart
```diff
@@ -946,11 +967,74 @@ class _KatexSpan extends StatelessWidget {
...
+    };
+
+    widget = SizedBox(
       height: styles.heightEm != null
-        ? styles.heightEm! * (fontSize ?? em)
+        ? styles.heightEm! * em
         : null,
       child: widget);
+
+    if (margin != null) {
```
gnprice, line 994: nit: put the `margin` computation above next to the spot where `margin` is used, here

lib/widgets/content.dart
```diff
@@ -896,12 +911,18 @@ class _KatexSpan extends StatelessWidget {
...
+    // `strut` span, for which parser explicitly emits `KatexStrutNode`.
+    // So, this should always be null for non `strut` spans.
+    assert(styles.verticalAlignEm == null);
 
     final fontFamily = styles.fontFamily;
     final fontSize = switch (styles.fontSizeEm) {
       double fontSizeEm => fontSizeEm * em,
       null => null,
     };
+    if (fontSize != null) em = fontSize;
```
gnprice, line 932: Cool, this is a good idea for making sure we don't forget `fontSize ??` in any `em` references below.

lib/widgets/content.dart
gnprice: Cool, thanks. Does that mean that the `marginLeftEm` code here never gets exercised? In that case let's have an assert to that effect.

lib/widgets/content.dart
gnprice: Hmm, well, I don't see that yet. :-)

What happens if you drop this commit?

gnprice (commented):
Thanks for the revision! I've again read through all but the last two commits:
f17ab3543 content [nfc]: Remove the `inline` property in _Katex widget
be0b048d6 content test: Add offset and size based widget tests for KaTeX content
b14dac214 content: Add a workaround for incorrect sizing in WidgetSpan, for KaTeX
dd64b6edd content: Scale inline KaTeX content based on the surrounding text
0a4b6a660 content test: Add widget tests for KaTeX content inside a quote
7d6634761 content: Handle 'strut' span in KaTeX content
beec4465a content: Handle positive margin-right and margin-left in KaTeX spans
6f85e1aaa content: Handle 'top' property in KaTeX span inline style

This is a lot closer to merge now — the nontrivial comments below are mostly about asking for more of a test on one of these commits.

Then still ahead are the last two commits, on vlists:
eabeb05f8 content: Handle vertical offset spans in KaTeX content
9dde631be content: Error message for unexpected CSS class in vlist inner span

lib/widgets/content.dart
rajveermalviya: Okay, squashed the separate commit that added support for handling `top` into the vlist commit as it was pre-req for vlist anyway. And made the parser to throw if it encounters `top` inline style on any span that isn't the vlist inner row span. Similarly on the widget side added an assert.

lib/widgets/content.dart
rajveermalviya: The next commit (2e2e4fadead0267ead0449e8ca146003c376b8d8) adds one test (`mathBlockKatexVertical2`) that requires it, so it gets exercised there.

test/widgets/content_test.dart
rajveermalviya: Added the `checkFontSizeRatio` test squashed in 05b51db4ae982a66eada184be2da7e374dfb22c6.

lib/widgets/content.dart
rajveermalviya: This commit is pre-req for 4bea0c3065417ffbb247818e50b809c294b372e3.

Allowing to get the rect for `Katex` widget without the `Center(SingleChildScrollViewWithScrollbar(` widgets affecting it, otherwise the offsets would take into account the whole test window width.

```diff
         (ContentExample.mathBlockKatexSizing, skip: false, [
-          ('1', Offset(0.00, 2.24), Size(25.59, 61.00)),
-          ('2', Offset(25.59, 9.90), Size(21.33, 51.00)),
-          ('3', Offset(46.91, 16.30), Size(17.77, 43.00)),
-          ('4', Offset(64.68, 21.63), Size(14.80, 36.00)),
-          ('5', Offset(79.48, 26.07), Size(12.34, 30.00)),
-          ('6', Offset(91.82, 29.77), Size(10.28, 25.00)),
-          ('7', Offset(102.10, 31.62), Size(9.25, 22.00)),
-          ('8', Offset(111.35, 33.47), Size(8.23, 20.00)),
-          ('9', Offset(119.58, 35.32), Size(7.20, 17.00)),
-          ('0', Offset(126.77, 39.02), Size(5.14, 12.00)),
+          ('1', Offset(334.02, 2.24), Size(25.59, 61.00)),
+          ('2', Offset(359.61, 9.90), Size(21.33, 51.00)),
+          ('3', Offset(380.94, 16.30), Size(17.77, 43.00)),
+          ('4', Offset(398.71, 21.63), Size(14.81, 36.00)),
+          ('5', Offset(413.52, 26.07), Size(12.34, 30.00)),
+          ('6', Offset(425.87, 29.77), Size(10.29, 25.00)),
+          ('7', Offset(436.15, 31.62), Size(9.26, 22.00)),
+          ('8', Offset(445.41, 33.47), Size(8.23, 20.00)),
+          ('9', Offset(453.64, 35.32), Size(7.20, 17.00)),
+          ('0', Offset(460.84, 39.02), Size(5.14, 12.00)),
         ]),
```

test/widgets/content_test.dart
rajveermalviya: The change in offset is because of `TextLeadingDistribution.even`. Moved it to a separate commit c748bcef5f4b3a03ca14c4f5a5aace47759ab075.

rajveermalviya:

lib/widgets/content.dart
```diff
@@ -820,41 +820,53 @@ class MathBlock extends StatelessWidget {
...
+/// Requires the [style.fontSize] to be non-null.
+TextStyle mkBaseKatexTextStyle(TextStyle style) {
+  return style.copyWith(
+    fontSize: style.fontSize! * 1.21,
+    fontFamily: 'KaTeX_Main',
+    height: 1.2,
+    fontWeight: FontWeight.normal,
+    fontStyle: FontStyle.normal,
+    textBaseline: TextBaseline.alphabetic,
+    leadingDistribution: TextLeadingDistribution.even,
```
gnprice, line 850: content: Update base KaTeX text style to be explicit
    
    This fixes two potential bugs:
    
    - First about `leadingDistribution` where previously it was taking
    the default value of `TextLeadingDistribution.proportional`, now it
    uses `TextLeadingDistribution.even` which seems to be the default
    strategy used by CSS, and it doesn't look like `katex.scss`
    overrides it.
    
    The vertical offsets being updated in tests are because of this fix.

This one is an actual bug, right? (Not just potential.) The changes in those vertical offsets demonstrate it was having a real effect — things were sitting a little too high, and also not quite aligned correctly with each other.

It is a *small* bug, though: at least in the test data, the differences are all less than one logical pixel (so up to a couple of physical pixels).

gnprice: - Second about `textBaseline` where on some locale systems the
    default value for this could be `TextBaseline.ideographic`, and for
    KaTeX we always want `TextBaseline.alphabetic`.

This one I guess is "potential" because I'm not sure whether the two baselines end up having a different effect with the fonts that KaTeX uses.

lib/widgets/content.dart
gnprice: OK, sounds good.

lib/widgets/content.dart
gnprice: Hmm, I see.

OK, please file a follow-up issue to track the bug described in https://github.com/zulip/zulip-flutter/pull/1609#discussion_r2181723874 . Then this code can just get a TODO comment mentioning that issue.

test/widgets/content_test.dart
```diff
@@ -1074,6 +1065,32 @@ void main() {
...
+      await globalSettings.setBool(BoolGlobalSetting.renderKatex, true);
+      check(globalSettings.getBool(BoolGlobalSetting.renderKatex)).isTrue();
+
+      const html = '<span class="katex">'
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML"><semantics><mrow><mi>λ</mi></mrow>'
+          '<annotation encoding="application/x-tex"> \\lambda </annotation></semantics></math></span>'
+        '<span class="katex-html" aria-hidden="true"><span class="base"><span class="strut" style="height:0.6944em;"></span><span class="mord mathnormal">λ</span></span></span></span>';
+      await checkFontSizeRatio(tester,
+        targetHtml: html,
+        targetFontSizeFinder: (rootSpan) {
```
gnprice, line 1058: Cool, this looks good.

gnprice (commented):
Thanks for the revision! These commits all look good now, except a few small comments below:
f17ab3543 content [nfc]: Remove the `inline` property in _Katex widget
4bea0c306 content test: Add offset and size based widget tests for KaTeX content
0c05d5867 content: Add a workaround for incorrect sizing in WidgetSpan, for KaTeX
c748bcef5 content: Update base KaTeX text style to be explicit
05b51db4a content: Scale inline KaTeX content based on the surrounding text
646098173 content: Handle 'strut' span in KaTeX content
f8f8d7845 content: Handle positive margin-right and margin-left in KaTeX spans

That leaves the last two commits still ahead for me to review:
2e2e4fade content: Handle vertical offset spans in KaTeX content
04dd75395 content: Error message for unexpected CSS class in vlist inner span

Let's split those to a separate PR, then. That way we can merge the rest of these changes very soon.

lib/widgets/content.dart
rajveermalviya: Hmm, I think wrapping `SingleChildScrollViewWithScrollbar` in `Directionality` again should fix this potential bug.

```diff
diff --git a/lib/widgets/content.dart b/lib/widgets/content.dart
index 396bbbba..27426792 100644
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -821,10 +821,12 @@ class MathBlock extends StatelessWidget {
     }
 
     return Center(
-      child: SingleChildScrollViewWithScrollbar(
-        scrollDirection: Axis.horizontal,
-        child: _Katex(
-          nodes: nodes)));
+      child: Directionality(
+        textDirection: TextDirection.ltr,
+        child: SingleChildScrollViewWithScrollbar(
+          scrollDirection: Axis.horizontal,
+          child: _Katex(
+            nodes: nodes))));
   }
 }
```

gnprice: Sure, that sounds good. Better than leaving a follow-up issue, in fact :-)

rajveermalviya:

I've removed those two last commits from this PR, and sent #1698 for those.

lib/widgets/content.dart
rajveermalviya: Added this fix in 9de7110ff91fa52f643ca3672dbb09e3ad86a9cc.

gnprice:
Thanks for all the revisions!

This looks good — just rebased, and will merge after CI passes.

PR #1453, opened 2025-04-01
local echo (7/7): Support simplified version of local echo

### LINKED ISSUE #1441: Simplified local echo, with retry
### Background

This is a new proposed spec for handling the same user need as the following previous issue:
* #720 

The point is to better handle the case where you type out a message and hit send, and it fails because you don't have working network at that moment. The current behavior is pretty unpleasant when it hits: the message you wrote is just gone.

Our previous plan, described in #720, didn't work out (https://github.com/zulip/zulip-flutter/issues/720#issuecomment-2529905664). I think ultimately the way to solve this will be to have local echo #576, showing the still-being-sent message in the message list, along with an outbox system #133 for retrying send. But doing local echo in full will be a substantial task, and an outbox system is tricky to get right.

So, for launch, here's a simplified approach.

### Spec

* We do a limited, placeholder, form of local echo: just show something like "(Sending message…)" where the message's content would go. (Or maybe the message source, as plain text.) This bypasses all the work of interpreting Zulip Markdown.

  To avoid this causing an annoying flicker whenever you actually do have a good connection to the server, we'll put this on a debounce timer: it only appears once it's been perhaps 300ms since the start of the send request, or sooner if the send request fails (as described below).

  (This part is anticipated in #576 as one of the interim options there.)

* Then if a message fails to send, we show on the local-echo placeholder an option that lets you recover it. (Similarly perhaps if it's been a while, like 10s, since trying to send and the request hasn't completed one way or another.)

  You might want to retry sending, or just copy the text to save elsewhere. To cover both options, it can take the text and just put it back in the compose box. (If there's already something there, the recovered message can go at the end, leaving the user the choice of what to do.) The placeholder in the message list then disappears.

  This is a bit like the sort of outbox system described in #133. It's simpler because there's no internal retry, which cuts out the `should-send` state and all its edges. An outbox-message starts at `sending`; goes to `sent` or `failed`; then gets removed.

### Design

We don't have an exact design for this in Figma. See the existing designs for editing a message:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4010-6347&m=dev

See also discussion at [#mobile > handle failed send @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/handle.20failed.20send/near/2128883); and use `#mobile-design` for discussing design questions that arise.

### Implementation

The data structures for the outstanding send requests will live on the message store. Like with other data there, that means the data structures will be fields on `MessageStoreImpl`, exposed via methods on `MessageStore`, so that code elsewhere in the app invokes the methods directly on the `PerAccountStore`.

(The existing method `sendMessage` should move from `PerAccountStore` to `MessageStore`.)

For a state diagram of the behavior of each send request, see [#mobile > handle failed send @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/handle.20failed.20send/near/2128824).

### References
* Previous issue: #720
* Related issues (for more-expansive versions of the different parts of this plan):
  #576 
  #133
* [#mobile > handle failed send @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/handle.20failed.20send/near/2128777)

PIG208:
Fixes #1441.

Stacked atop #1472.

(This branch can be used to preview the full implementation)

<details>

<summary> screenshots </summary>

| message sent | got message event |
| - | - |
| sending-success-before | sending-success-after |

| message failed to send | message content tapped | recipient headers |
| - | - | - |
| send-failed | sent-failed-interacted | recipient-headers |

</details>

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1941,7 +1941,8 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
     // TODO timeout this request?
     if (!mounted) return;
     if (!identical(controller, emptyEditController)) {
-      // user tapped Cancel during the fetch-raw-content request
+      // During the fetch-raw-content request, the user tapped Cancel
+      // or tapped a failed message edit to restore.
       // TODO in this case we don't want the error dialog caused by
       //   ZulipAction.fetchRawContentWithFeedback; suppress that
       return;
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1290,9 +1290,6 @@ class _SendButtonState extends State<_SendButton> {
     controller.content.clear();
 
     try {
-      // TODO(#720) clear content input only on success response;
-      //   while waiting, put input(s) and send button into a disabled
-      //   "working on it" state (letting input text be selected for copying).
       await store.sendMessage(destination: widget.getDestination(), content: content);
     } on ApiRequestException catch (e) {
       if (!mounted) return;
@@ -1384,7 +1381,6 @@ class _ComposeBoxContainer extends StatelessWidget {
         border: Border(top: BorderSide(color: designVariables.borderBar)),
         boxShadow: ComposeBoxTheme.of(context).boxShadow,
       ),
-      // TODO(#720) try a Stack for the overlaid linear progress indicator
       child: Material(
         color: designVariables.composeBoxBg,
         child: Column(
@@ -1742,10 +1738,10 @@ class _ErrorBanner extends _Banner {
 
   @override
   Widget? buildTrailing(context) {
-    // TODO(#720) "x" button goes here.
-    //   24px square with 8px touchable padding in all directions?
-    //   and `bool get padEnd => false`; see Figma:
-    //     https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4031-17029&m=dev
+    // An "x" button can go here.
+    // 24px square with 8px touchable padding in all directions?
+    // and `bool get padEnd => false`; see Figma:
+    //   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4031-17029&m=dev
     return null;
   }
 }
@@ -2083,11 +2079,6 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
       }
     }
 
-    // TODO(#720) dismissable message-send error, maybe something like:
-    //     if (controller.sendMessageError.value != null) {
-    //       errorBanner = _ErrorBanner(label:
-    //         ZulipLocalizations.of(context).errorSendMessageTimeout);
-    //     }
     return ComposeBoxInheritedWidget.fromComposeBoxState(this,
       child: _ComposeBoxContainer(body: body, banner: banner));
   }
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -13,6 +13,7 @@ import '../api/route/messages.dart';
 import '../generated/l10n/zulip_localizations.dart';
 import '../model/binding.dart';
 import '../model/compose.dart';
+import '../model/message.dart';
 import '../model/narrow.dart';
 import '../model/store.dart';
 import 'actions.dart';
@@ -1840,6 +1841,16 @@ class ComposeBox extends StatefulWidget {
 abstract class ComposeBoxState extends State<ComposeBox> {
   ComposeBoxController get controller;
 
+  /// Fills the compose box with the content of an [OutboxMessage]
+  /// for a failed [sendMessage] request.
+  ///
+  /// If there is already text in the compose box, gives a confirmation dialog
+  /// to confirm that it is OK to discard that text.
+  ///
+  /// [localMessageId], as in [OutboxMessage.localMessageId], must be present
+  /// in the message store.
+  void restoreMessageNotSent(int localMessageId);
+
   /// Switch the compose box to editing mode.
   ///
   /// If there is already text in the compose box, gives a confirmation dialog
@@ -1860,6 +1871,29 @@ abstract class ComposeBoxState extends State<ComposeBox> {
 class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateMixin<ComposeBox> implements ComposeBoxState {
   @override ComposeBoxController get controller => _controller!;
   ComposeBoxController? _controller;
+
+  @override
+  void restoreMessageNotSent(int localMessageId) async {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    final abort = await _abortBecauseContentInputNotEmpty(
+      dialogMessage: zulipLocalizations.discardDraftForOutboxConfirmationDialogMessage);
+    if (abort || !mounted) return;
+
+    final store = PerAccountStoreWidget.of(context);
+    final outboxMessage = store.takeOutboxMessage(localMessageId);
+    setState(() {
+      _setNewController(store);
+      final controller = this.controller;
+      controller
+        ..content.value = TextEditingValue(text: outboxMessage.contentMarkdown)
+        ..contentFocusNode.requestFocus();
+      if (controller is StreamComposeBoxController) {
+        controller.topic.setTopic(
+          (outboxMessage.conversation as StreamConversation).topic);
+      }
+    });
+  }
 
   @override
   void startEditInteraction(int messageId) async {
@@ -1942,7 +1976,7 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
     if (!mounted) return;
     if (!identical(controller, emptyEditController)) {
       // During the fetch-raw-content request, the user tapped Cancel
-      // or tapped a failed message edit to restore.
+      // or tapped a failed message edit or failed outbox message to restore.
       // TODO in this case we don't want the error dialog caused by
       //   ZulipAction.fetchRawContentWithFeedback; suppress that
       return;
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1288,10 +1288,6 @@ class _SendButtonState extends State<_SendButton> {
     final content = controller.content.textNormalized;
 
     controller.content.clear();
-    // The following `stoppedComposing` call is currently redundant,
-    // because clearing input sends a "typing stopped" notice.
-    // It will be necessary once we resolve #720.
-    store.typingNotifier.stoppedComposing();
 
     try {
       // TODO(#720) clear content input only on success response;
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -5,6 +5,7 @@ import 'package:intl/intl.dart' hide TextDirection;
 
 import '../api/model/model.dart';
 import '../generated/l10n/zulip_localizations.dart';
+import '../model/message.dart';
 import '../model/message_list.dart';
 import '../model/narrow.dart';
 import '../model/store.dart';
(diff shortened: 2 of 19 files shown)


gnprice:
I've been running on my phone a build with this PR in it for the past few days (as described at [#mobile-team > beta release management @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/beta.20release.20management/near/2174229)). It works! Glad to see this feature nearing completion.

One observation: the loading indicator is pretty conspicuous, and feels too conspicuous in the normal case where the message promptly succeeds.

* One factor is that it causes the message to be taller, so that it and all the messages above it end up shifting down slightly after the send completes. I think this is why the design in Figma calls for the indicator to encroach on the padding instead: https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2085792800

  Unlike for the editing case, though, this might be easy to fix because `OutboxMessageWithPossibleSender` is already a separate widget. If so, let's fix it.

* Even after fixing that shift, it might be helpful to de-emphasize the indicator by having it only appear after some time has elapsed, like 500ms or 1s.

PIG208:
Thanks! Updated this to use 500ms, and rearranged the paddings to avoid displacement.

alya:
The error text reads a bit funny to me. I think "Message not sent." would be better for the first part.

lib/widgets/message_list.dart
```diff
@@ -1615,22 +1616,108 @@ class OutboxMessageWithPossibleSender extends StatelessWidget {
...
+      child: Padding(
+        padding: const EdgeInsets.only(top: 4),
+        child: Column(children: [
+          if (item.showSender)
+            _SenderRow(message: item.message, showTimestamp: false),
+          Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 16),
+            child: Column(crossAxisAlignment: CrossAxisAlignment.stretch,
+              children: [
+                // This is adapated from [MessageContent].
```
chrisbobbe, line 1677: nit: "adapted"

chrisbobbe (commented):
A few small comments from a very incomplete review, but wanted to post before I go out the door 🙂—

PIG208:
This should address comments other than the edit-message-like change; will get to that later.

Also rebased on top of the updated "local echo (5/7)" PR, fixing some tests that relied on an earlier state machine design.

This branch has some changes to accommodate the recently added `middleMessage`/`middleItem`. The tests that fail in "msglist: Add and manage outbox message objects in message list view" will pass in the commit that follows it. I will need to double-check the changes related to `middleMessage`/`middleItem`, but they should not affect the majority of this implementation.

PIG208:
I temporarily removed the last two commits because the confirmation dialog requires us to rework the implementation and tests more.

chrisbobbe (commented):
Thanks! Here's another round of review—it includes some comments I wrote yesterday on those commits that you temporarily removed in this revision, and they may or may not still apply to your new versions of those commits :)

lib/widgets/message_list.dart
```diff
@@ -1601,3 +1606,123 @@ class _RestoreEditMessageGestureDetector extends StatelessWidget {
...
+        final designVariables = DesignVariables.of(context);
+        return Padding(
+          padding: const EdgeInsetsGeometry.only(top: 1.5, bottom: 0.5),
+          child: LinearProgressIndicator(
+            minHeight: 2,
+            color: designVariables.foreground.withFadedAlpha(0.5),
+            backgroundColor: designVariables.foreground.withFadedAlpha(0.2)));
+
+      case OutboxMessageState.failed:
+      case OutboxMessageState.waitPeriodExpired:
```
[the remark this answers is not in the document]
gnprice: In v0.0.29 we'd enter this state even if we got a success response from the send request, if the event didn't then show up. I suspect that's what happened in the report in #1525.

In the current revision of this PR, a success response will put the outbox-message into `waiting` state indefinitely, until the event arrives. I think showing this is OK if we've gotten neither the event nor a response from the request.

(The exact status in that case is more like "tried sending the message, it's been a while, haven't heard any response yet, so it may or may not have made it through, up to you if you want to retry". Not sure there's a good concise way to communicate that.)

lib/model/message_list.dart
```diff
@@ -163,6 +191,7 @@ mixin _MessageSequence {
   /// Either the bottom slices of both [items] and [messages] are empty,
   /// or the first item in the bottom slice of [items] is a [MessageListMessageItem]
   /// for the first message in the bottom slice of [messages].
+  // TODO(#1453) update this in the context of outbox messages.
```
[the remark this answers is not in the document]
PIG208: Ah, this is just a TODO for myself to update this dart doc, since we can have outbox-message related items in the list.

lib/model/message_list.dart
```diff
@@ -139,14 +156,25 @@ mixin _MessageSequence {
...
   final List<ZulipMessageContent> contents = [];
 
+  /// The messages sent by the self-user, retrieved from
+  /// [MessageStore.outboxMessages].
+  ///
+  /// See also [items].
+  ///
+  /// Usually this should not have that many items, so we do not anticipate
+  /// performance issues with unoptimized O(N) iterations through this list.
+  final List<OutboxMessage> outboxMessages = [];
```
gnprice, line 166: This adds some overhead in magnitude of O(1) (where the constant is
    the number of outbox messages in a view, expected to be small) on
    message event handling.

That is not what the word "constant" means :slightly_smiling_face: 

Instead can say:

> This adds some overhead on message-event handling, linear in the number of outbox messages in a view. We rely on that number being small.

Or could say something like "O(k), where there are k outbox messages in the view".

lib/model/message_list.dart
```diff
@@ -63,6 +64,22 @@ class MessageListMessageItem extends MessageListMessageBaseItem {
...
+  final OutboxMessage message;
+  @override
+  final ZulipContent content;
+
+  MessageListOutboxMessageItem(
+    this.message, {
+    required super.showSender,
+    required super.isLastInBlock,
+  }) : content = ZulipContent(nodes: [
+    ParagraphNode(links: [], nodes: [TextNode(message.contentMarkdown)]),
```
gnprice, line 79:
```suggestion
    ParagraphNode(links: null, nodes: [TextNode(message.contentMarkdown)]),
```

See the doc on the `links` field:
```
  /// A list of all [LinkNode] descendants.
  ///
  /// An empty list is represented as null.
  /// […]
```

lib/widgets/message_list.dart
```diff
@@ -718,7 +718,11 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
         return MessageItem(
           key: ValueKey(data.message.id),
           header: header,
-          trailingWhitespace: 11,
+          item: data);
+      case MessageListOutboxMessageItem():
```
gnprice, line 722: This implements minimal support to display outbox message message item
    widgets in the message list, without indicators for theirs states.
    Retrieving content from failed sent requests and the full UI are
    implemented in a later commit.

Is there — ah, I guess that was in the commits you temporarily removed as mentioned at https://github.com/zulip/zulip-flutter/pull/1453#issuecomment-2903279848 .

lib/model/message_list.dart
```diff
@@ -385,13 +451,67 @@ mixin _MessageSequence {
...
+  ///
+  /// This is designed to be idempotent; repeated calls will not change the
+  /// content of [items].
+  ///
+  /// This is efficient due to the expected small size of [outboxMessages].
+  void _removeOutboxMessageItems() {
+    // This loop relies on the assumption that all items that follow
+    // the last [MessageListMessageItem] are derived from outbox messages.
+    // If there is no [MessageListMessageItem] at all,
+    // this will end up removing end markers.
```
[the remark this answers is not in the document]
PIG208: This should end up removing all items, but that's less relevant now because we used to need to restore the end markers at the end.

lib/widgets/message_list.dart
```diff
@@ -1601,3 +1606,123 @@ class _RestoreEditMessageGestureDetector extends StatelessWidget {
...
+    switch (outboxMessageState) {
+      case OutboxMessageState.hidden:
+        assert(false,
+          'Hidden OutboxMessage messages should not appear in message lists');
+        return SizedBox.shrink();
+
+      case OutboxMessageState.waiting:
+        final designVariables = DesignVariables.of(context);
+        return Padding(
+          padding: const EdgeInsetsGeometry.only(top: 1.5, bottom: 0.5),
```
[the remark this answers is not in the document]
PIG208: Greg mentioned that for the case of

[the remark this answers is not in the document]
PIG208: While we have adjusted the debounce timeout in the more recent version #1472, I think the issue with the layout shift caused by the loading indicator still exists (https://github.com/zulip/zulip-flutter/pull/1453#issuecomment-2885667930), at least for messages that don't have fancy formatting.

We have 4px total in message content's bottom padding, and the height of the progress bar is 2px. How about moving the progress bar to the top half of the bottom padding, so that there is 2px instead of 0.5px after it? Meanwhile, we can adjust the bottom padding of the edit status row to be 2px so that they are consistent.'

| 2px below | 2px below (edit-message) |
| - | - |
| image | image |

Note that the local echo indicator sits inside the bottom padding of the message content, and the one for edit-message doesn't. I think this is fine since there is label that doesn't fit within the padding for edit-message anyway.

test/model/narrow_test.dart
```diff
@@ -220,16 +194,19 @@ void main() {
...
+        eg.streamOutboxMessage(stream: eg.stream(), topic: 'topic'))).isFalse();
       check(narrow.containsMessage(
-        _TestDmMessage(allRecipientIds: [2]))).isFalse();
+        eg.dmOutboxMessage(from: user2, to: []))).isFalse();
       check(narrow.containsMessage(
-        _TestDmMessage(allRecipientIds: [2, 3]))).isFalse();
+        eg.dmOutboxMessage(from: user2, to: [user3]))).isFalse();
       check(narrow.containsMessage(
-        _TestDmMessage(allRecipientIds: [1, 2]))).isTrue();
+        eg.dmOutboxMessage(from: user1, to: [user2]))).isTrue();
```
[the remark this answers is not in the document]
PIG208: Hmm, right. The test was not set up correctly. Adding a prep commit (ideally we would change the old helpers to use `from/to`, but I kept it minimal just so we can fix the test data), so that we can keep the refactor NFC.

PIG208:
Thanks for the reviews! Pushed an update, this time with the commits I removed last time, with a bunch of prep commits. I also went through this PR's interaction with `middleMessage`/`middleItem`, and implemented the behavior discussed in a prior call with Greg, making sure that `middleItem` points to `MessageListOutboxMessageItem` in certain condition (the docs on `middleItem` has been updated to reflect the behavior I intended and implemented).

gnprice:
Thanks for the revision! The changes related to my review at https://github.com/zulip/zulip-flutter/pull/1453#pullrequestreview-2867241573 all look good, and the initial 5 commits all look good to me now:
c1f6b8dca msglist [nfc]: Make trailingWhitespace a constant
0b12cab63 narrow test: Make sure sender is selfUser in tests
e1fa25e8f test [nfc]: Extract {dm,stream}OutboxMessage helpers
33a2e0d04 msglist test [nfc]: Make checkInvariant compatible with MessageBase
93d89a6ac msglist [nfc]: Extract _addItemsForMessage

I haven't yet read the subsequent commits in full:
5488a7a2e msglist: Use store.senderDisplayName for sender row
d64698efa msglist [nfc]: Make _SenderRow accept MessageBase
9c0b95568 msglist: Add and manage outbox message objects in message list view
6604cf49f compose [nfc]: Make confirmation dialog message flexible
b2d6c50c2 compose test [nfc]: Move some edit message helpers out of group
cfbbdc0cc compose [nfc]: Update string to mention "edit" in name and description
c8295ae9b msglist: Adjust edit-message status row bottom padding

@chrisbobbe will be taking over this PR; I'll leave those latter commits until after he's made a pass through them.

chrisbobbe:
Great! Merging those first five, with a small commit-message tweak:

```
$ git range-diff upstream 93d89a6a a4c564b1c
1:  c1f6b8dca = 1:  86f62c9b1 msglist [nfc]: Make trailingWhitespace a constant
2:  0b12cab63 ! 2:  75f0debbf narrow test: Make sure sender is selfUser in tests
    @@ Metadata
     Author: Zixuan James Li <zixuan@zulip.com>
     
      ## Commit message ##
    -    narrow test: Make sure sender is selfUser in tests
    +    narrow test: Make sure sender is selfUser for outbox DM messages
    +
    +    [chris: expanded commit-message summary line]
    +
    +    Co-authored-by: Chris Bobbe <cbobbe@zulip.com>
     
      ## test/model/narrow_test.dart ##
     @@ test/model/narrow_test.dart: void main() {
3:  e1fa25e8f = 3:  815b9d2f4 test [nfc]: Extract {dm,stream}OutboxMessage helpers
4:  33a2e0d04 = 4:  2829bd847 msglist test [nfc]: Make checkInvariant compatible with MessageBase
5:  93d89a6ac = 5:  a4c564b1c msglist [nfc]: Extract _addItemsForMessage
```

chrisbobbe:
And I've sent revised versions of the next chunk, in #1535. So that just leaves the last two commits of this PR:

c8295ae9b msglist: Adjust edit-message status row bottom padding

lib/widgets/message_list.dart
```diff
@@ -1651,3 +1656,126 @@ class _RestoreEditMessageGestureDetector extends StatelessWidget {
...
+class OutboxMessageWithPossibleSender extends StatelessWidget {
+  const OutboxMessageWithPossibleSender({super.key, required this.item});
+
+  final MessageListOutboxMessageItem item;
+
+  @override
+  Widget build(BuildContext context) {
+    final message = item.message;
+    final localMessageId = message.localMessageId;
+    final state = item.message.state;
```
gnprice: Maybe just say `message.state` in the two places this is used. I feel like a bare name `state` could be confusing in the context of a widget — sounds like the state of some stateful widget.

gnprice (commented):
Thanks for picking this up! I've read through the commits after #1538:
72ffbd872 compose: Remove a redundant TypingNotifier.stoppedComposing call
3b9366e4e compose [nfc]: Remove obsoleted TODO(#720)s
d56e20fec compose [nfc]: Expand a comment to include an edge case
419e17bf5 msglist: Support retrieving failed outbox message content

except the tests of the last/main commit. All looks good except one small comment below. (And I should read the tests too.)

chrisbobbe:
Thanks for the review! Revision pushed, atop the current #1538.

(edit: and again, resolving a small conflict)

lib/widgets/compose_box.dart
```diff
@@ -1869,6 +1872,29 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
...
+    if (abort || !mounted) return;
+
+    final store = PerAccountStoreWidget.of(context);
+    final outboxMessage = store.takeOutboxMessage(localMessageId);
+    setState(() {
+      _setNewController(store);
+      final controller = this.controller;
+      controller
+        ..content.value = TextEditingValue(text: outboxMessage.contentMarkdown)
+        ..contentFocusNode.requestFocus();
```
gnprice, line 1890: This bit should ideally get a test.

In the interest of merging this for launch, let's just leave a TODO comment for it in the test code.

test/widgets/message_list_test.dart
```diff
@@ -1551,6 +1553,163 @@ void main() {
...
+      await tester.pump(); // wait for navigation
+      check(contentInputFinder).findsOne();
+
+      await sendMessageAndFail(tester);
+      await dismissErrorDialog(tester);
+      // Navigate back to the message list page without a compose box,
+      // where the failed to send message should be visible.
+      await tester.pageBack();
+      await tester.pump(); // handle tap
+      await tester.pump(const Duration(milliseconds: 500)); // wait for navigation
```
gnprice: Should this be adapted to not specify the duration, per that upstream PR?

test/widgets/message_list_test.dart
```diff
@@ -1551,6 +1553,163 @@ void main() {
...
+      check(contentInputFinder).findsNothing();
+      check(messageNotSentFinder).findsOne();
+
+      // Tap the failed to send message.
+      // This should not remove it from the message list.
+      await checkTapNotRestoreMessage(tester);
+    });
+
+    testWidgets('hidden -> failed, tap to restore message', (tester) async {
```
gnprice: nit: perhaps put this above the "tapping does nothing if compose box is not offered" case — this is a lot simpler and serves as a sort of baseline for understanding the more complex cases

chrisbobbe:
Rebased on current `main`; conflicts resolved.

test/widgets/message_list_test.dart
gnprice: Because customer_testing pins the version of our repo, I guess it's not an immediate problem (even for that upstream PR) if we merge it this way. But if the upstream PR lands and this is already merged, it'd break our CI.

gnprice (commented):
Thanks! Also just read through those tests; a handful of comments below.

lib/widgets/compose_box.dart
chrisbobbe: This is easy to test :) done.

chrisbobbe:

lib/widgets/compose_box.dart
gnprice: Cool, looks good :)

gnprice:
Thanks! Looks good; merging.

… perhaps after #1568 lands, in order to pin that one down with certainty without another CI cycle.

PR #1454, opened 2025-04-01
local echo (1/n): Add API support for local_id/local_message_id and queue_id

PIG208:
Related:
- Work toward #1441 
- Prep for #1453

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -682,7 +682,14 @@ class MessageEvent extends Event {
   @JsonKey(readValue: _readMessageValue, includeToJson: false)
   final Message message;
 
-  MessageEvent({required super.id, required this.message});
+  // When present, this equals the "local_id" parameter
+  // from a previous [sendMessage] call by us.
+  //
+  // This is not yet fully documented.  See CZO discussion for reference:
+  //   https://chat.zulip.org/#narrow/channel/412-api-documentation/topic/local_id.2C.20queue_id.2Fsender_queue_id/near/2135340
+  final String? localMessageId;
+
+  MessageEvent({required super.id, required this.message, required this.localMessageId});
 
   static Map<String, dynamic> _readMessageValue(Map<dynamic, dynamic> json, String key) =>
     {...json['message'] as Map<String, dynamic>, 'flags': json['flags']};
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -665,8 +665,7 @@ class UserTopicEvent extends Event {
 }
 
 /// A Zulip event of type `message`: https://zulip.com/api/get-events#message
-// TODO use [JsonSerializable] here too, using its customization features,
-//   in order to skip the boilerplate in [fromJson] and [toJson].
+@JsonSerializable(fieldRename: FieldRename.snake)
 class MessageEvent extends Event {
   @override
   @JsonKey(includeToJson: true)
@@ -680,24 +679,23 @@ class MessageEvent extends Event {
   // events and in the get-messages results is that `matchContent` and
   // `matchTopic` are absent here.  Already [Message.matchContent] and
   // [Message.matchTopic] are optional, so no action is needed on that.
+  @JsonKey(readValue: _readMessageValue, includeToJson: false)
   final Message message;
 
   MessageEvent({required super.id, required this.message});
 
-  factory MessageEvent.fromJson(Map<String, dynamic> json) => MessageEvent(
-    id: json['id'] as int,
-    message: Message.fromJson({
-      ...json['message'] as Map<String, dynamic>,
-      'flags': (json['flags'] as List<dynamic>).map((e) => e as String).toList(),
-    }),
-  );
+  static Map<String, dynamic> _readMessageValue(Map<dynamic, dynamic> json, String key) =>
+    {...json['message'] as Map<String, dynamic>, 'flags': json['flags']};
+
+  factory MessageEvent.fromJson(Map<String, dynamic> json) =>
+    _$MessageEventFromJson(json);
 
   @override
   Map<String, dynamic> toJson() {
     final messageJson = message.toJson();
     final flags = messageJson['flags'];
     messageJson.remove('flags');
-    return {'id': id, 'type': type, 'message': messageJson, 'flags': flags};
+    return {..._$MessageEventToJson(this), 'message': messageJson, 'flags': flags};
   }
 }
 
--- a/lib/api/route/messages.dart
+++ b/lib/api/route/messages.dart
@@ -193,8 +193,8 @@ Future<SendMessageResult> sendMessage(
         'to': destination.userIds,
       }}),
     'content': RawParameter(content),
-    if (queueId != null) 'queue_id': queueId, // TODO should this use RawParameter?
-    if (localId != null) 'local_id': localId, // TODO should this use RawParameter?
+    if (queueId != null) 'queue_id': RawParameter(queueId),
+    if (localId != null) 'local_id': RawParameter(localId),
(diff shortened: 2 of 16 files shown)


rajveermalviya (approved):
Thanks for working on this! This LGTM.

(Also, for the PR title you probably meant 'local echo' instead of 'locale echo').

test/api/route/messages_test.dart
```diff
@@ -355,15 +355,15 @@ void main() {
...
+          localId: 456,
           readBySender: true,
           expectedBodyFields: {
             'type': 'stream',
             'to': streamId.toString(),
             'topic': topic,
             'content': content,
-            'queue_id': '"abc:123"',
-            'local_id': '"456"',
+            'queue_id': 'abc:123',
```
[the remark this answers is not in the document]
gnprice: Strings in general should not be JSON-encoded in Zulip APIs.

It's true that a string as a top-level parameter to a Zulip API endpoint is typically sent as itself, not JSON-encoded. That isn't always true, though. Right in this same endpoint, when passing a channel name to `to`, that's expected to be as JSON and not a raw string.

(The server does currently accept a raw stream name there, apparently, but that isn't [and fundamentally can't be](https://chat.zulip.org/#narrow/channel/412-api-documentation/topic/POST.20does.20not.20accept.20body/near/1841465) reliable.)

Moreover our docs [don't really cover](https://chat.zulip.org/#narrow/channel/412-api-documentation/topic/POST.20does.20not.20accept.20body/near/1734582) where this is the case or isn't, except via the `curl` examples; and the `curl` example for this endpoint doesn't have `queue_id`.

So for this change, I think the key point is that you've tested and the raw string is what works.

lib/api/model/events.dart
```diff
@@ -680,24 +679,38 @@ class MessageEvent extends Event {
...
+  @JsonKey(readValue: _readMessageValue, includeToJson: false)
   final Message message;
 
-  MessageEvent({required super.id, required this.message});
+  // The format of this is documented to be chosen freely by the client.
+  //
+  // When present, this corresponds to the JSON-encoded int, "local_id",
+  // from a previous [sendMessage] call by us.
+  @JsonKey(fromJson: _localMessageIdFromJson, toJson: _localMessageIdToJson)
+  final int? localMessageId;
```
gnprice, line 690: Since this isn't yet covered by the API docs (which are linked from this class's dartdoc), let's include dartdoc with a link to the best reference we have. That'd be the link in the commit message:
https://chat.zulip.org/#narrow/channel/412-api-documentation/topic/local_id.2C.20queue_id.2Fsender_queue_id/near/2135340

lib/api/route/messages.dart
```diff
@@ -193,8 +193,12 @@ Future<SendMessageResult> sendMessage(
         'to': destination.userIds,
       }}),
     'content': RawParameter(content),
-    if (queueId != null) 'queue_id': queueId, // TODO should this use RawParameter?
-    if (localId != null) 'local_id': localId, // TODO should this use RawParameter?
+    if (queueId != null) 'queue_id': RawParameter(queueId),
+    // This is documented to be an optional String whose format is chosen freely
+    // by the client.  Use a JSON-encoded int consistently, so that we know how
+    // to decode it when we receive the corresponding "message" event.
+    // See also: [MessageEvent.localMessageId]
```
gnprice, line 200: In general I'd rather keep these bindings as close as possible to the underlying Zulip API — adapting the API to reasonable Dart idioms, but generally not encoding our particular choices of how the app uses the API.

One reason is that it saves us from needing this sort of comment to explain how and why the internal API differs from the one that's in the API docs :slightly_smiling_face: (or in this case hopefully will be documented there).

What if we just leave this as a string? The MessageStore code that calls this endpoint, and then consumes the `local_message_id` in the resulting MessageEvent, will end up converting an int message ID to string and back, but that's fine — it's really that code that is responsible for matching those requests and events up with each other, so that's a natural place to put the responsibility for deciding what these strings should look like.

test/example_data.dart
```diff
@@ -620,6 +620,9 @@ UserTopicEvent userTopicEvent(
   );
 }
 
+MessageEvent messageEvent(Message message) =>
```
gnprice, line 623: commit-message nit:

    example_data [nfc]: Add messageEvent

Better:

    test [nfc]: Add eg.messageEvent

For examples of previous summary lines, try this command:
```
$ git log --format='%>(20)%an %cs %h %s' --grep 'eg\.|example' origin \
    | grep -P '\bexample|\beg\b'
```

gnprice (commented):
Thanks! Comments below.

lib/api/route/messages.dart
PIG208: Yeah, I agree it will be cleaner when we handle that encoding at a different layer. In this case it should be self-contained in `MessageStore`.

test/example_data.dart
PIG208: Thanks! I added a `-E` flag to the `git log` command to get this working.

PIG208:
Updated the PR. This should be ready for review. Thanks!

test/example_data.dart
gnprice: Ah yeah, thanks. I have this in my `~/.gitconfig`, which I recommend:
```
[grep]
        # "Basic" regexps are a relic.  Extended would be fine; Perl is better.
        patternType = perl
```

That's equivalent to a `-P` flag, which goes farther in the same direction as `-E`.

lib/api/model/events.dart
```diff
@@ -680,24 +679,32 @@ class MessageEvent extends Event {
...
   // `matchTopic` are absent here.  Already [Message.matchContent] and
   // [Message.matchTopic] are optional, so no action is needed on that.
+  @JsonKey(readValue: _readMessageValue, includeToJson: false)
   final Message message;
 
-  MessageEvent({required super.id, required this.message});
+  // The format of this is documented to be chosen freely by the client.
+  //
+  // When present, this corresponds to the JSON-encoded int, "local_id",
+  // from a previous [sendMessage] call by us.
```
gnprice, line 688:
```suggestion
  // When present, this equals the "local_id" parameter
  // from a previous [sendMessage] call by us.
```

The details about what we choose to put into "local_id" are no longer something this needs in order to justify how this code works, so this can stick to documenting the API itself.

lib/api/route/messages.dart
```diff
@@ -193,8 +193,8 @@ Future<SendMessageResult> sendMessage(
         'to': destination.userIds,
       }}),
     'content': RawParameter(content),
-    if (queueId != null) 'queue_id': queueId, // TODO should this use RawParameter?
-    if (localId != null) 'local_id': localId, // TODO should this use RawParameter?
+    if (queueId != null) 'queue_id': RawParameter(queueId),
+    if (localId != null) 'local_id': RawParameter(localId),
```
gnprice, line 197: The change to `localId` here on this route is now totally parallel to the change to `queueId`, so probably best done in the same commit.

Then the change adding `localMessageId` on the event can remain in its own subsequent commit.

gnprice (commented):
Thanks! This looks good; just small comments below.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! A commit message needs updating:

    api: Use RawParameter when passing queue_id
    
    This endpoint in particular expects queue_id to be a String just
    passed as itself, not JSON-encoded.
    
    We will handle local_id separately because that's a bit different.

as that commit now handles local_id too :-)

Otherwise looks all ready.

PIG208:
Oops. Just pushed another update.

gnprice:
Thanks! Looks good; merging.

PR #1455, opened 2025-04-01
local echo (2/n): Rearrange `queueId` and `sendMessage` within store

PIG208:
Related:
- Work toward #1441 
- Prep for #1453

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -393,7 +393,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
         typingStartedExpiryPeriod: Duration(milliseconds: initialSnapshot.serverTypingStartedExpiryPeriodMilliseconds),
       ),
       channels: channels,
-      messages: MessageStoreImpl(),
+      messages: MessageStoreImpl(connection: connection),
       unreads: Unreads(
         initial: initialSnapshot.unreadMsgs,
         selfUserId: account.userId,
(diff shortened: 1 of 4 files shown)


lib/model/store.dart
```diff
@@ -420,6 +429,7 @@ class PerAccountStore extends ChangeNotifier with EmojiStore, UserStore, Channel
   final GlobalStore _globalStore;
   final ApiConnection connection; // TODO(#135): update zulipFeatureLevel with events
 
+  String queueId;
```
gnprice, line 432:
```suggestion
  final String queueId;
```

Right? (Just like it was when on UpdateMachine.)

lib/model/store.dart
```diff
@@ -1075,7 +1073,7 @@ class UpdateMachine {
   }
 
   final PerAccountStore store;
-  final String queueId;
+  String get queueId => store.queueId;
```
gnprice, line 1076: This getter is fine for keeping the commit small, but let's have a followup commit clean it up by having its callers refer directly to the store. I think most of them won't even get any longer — they'll replace something like `updateMachine.queueId` with `store.queueId`.

test/model/store_test.dart
```diff
@@ -212,6 +212,16 @@ void main() {
     check(connection).isOpen.isFalse();
   }));
 
+  test('GlobalStore.perAccount throws if missing queueId', () async {
+    final globalStore = UpdateMachineTestGlobalStore(accounts: [eg.selfAccount]);
+    globalStore.prepareRegisterQueueResponse = (connection) {
+      connection.prepare(json:
+      deepToJson(eg.initialSnapshot()) as Map<String, dynamic>
```
gnprice, line 219: nit: indentation

gnprice (commented):
Thanks! Looks good, with small comments below.

lib/model/store.dart
PIG208: The change needed removing this ends up being quite small, so I added it to the same commit.

PIG208:
Thanks! I have updated the PR.

gnprice:

Issue #1419, opened 2025-03-18
Image auto compression, when upload image larger than 10MB or more


code959437957:
The current zulip android client , won't let you upload image file larger than 10MB,

I suggest we should auto compress image file to lower than 10MB, then upload it .

some Message or chat app will auto compress image file to lower resolution image , then upload it .

zulip should do it either

Vrushabh102:
May I work on this issue?

Vrushabh102:
Thanks, I have started

umairahmadx:
Should i fix this issue? worked on this problem before only

gnprice:
Thanks @code959437957 for the feedback.

I don't think we currently have in the app a limit like that. So I'd be interested in better understanding the behavior you're seeing. Can you describe more specifically what happens when you try to upload a larger file?

andreasoc:
essential

PR #1462, opened 2025-04-03
fix : image compress above 10mb

Riddesh12:
fix issue #1419


gnprice:
This general direction isn't anything we've decided we want — the issue thread is still awaiting more information on what the original reporter was seeing. So I'm closing this PR.

This PR also doesn't meet the baseline requirements for us to be able to review it: it would need tests, and clean commits. See our README.

Riddesh12:
Now I have made the changes and solved the error, so should I push the repository or not.

gnprice:
Well, would that change the reason why the PR is closed?

Riddesh12:
Ya, that is right.

PR #1463, opened 2025-04-03
local echo (3/n): Pull out MessageBase

PIG208:
Related:
- Work toward #1441 
- Prep for #1453

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -619,7 +619,7 @@ sealed class Message {
     required this.matchTopic,
   });
 
-  factory Message.fromJson(Map<String, dynamic> json) {
+  static Message fromJson(Map<String, dynamic> json) {
     final type = json['type'] as String;
     if (type == 'stream') return StreamMessage.fromJson(json);
     if (type == 'private') return DmMessage.fromJson(json);
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -591,7 +591,7 @@ extension type const TopicName(String _value) {
   String toJson() => apiName;
 }
 
-/// As in [StreamMessage.conversation] and [DmMessage.conversation].
+/// As in [MessageBase.conversation].
 ///
 /// Different from [MessageDestination], this information comes from
 /// [getMessages] or [getEvents], identifying the conversation that contains a
@@ -637,10 +637,35 @@ class DmConversation extends Conversation {
     : assert(isSortedWithoutDuplicates(allRecipientIds.toList()));
 }
 
+/// A message or message-like object, for showing in a message list.
+///
+/// Other than [Message], we use this for "outbox messages",
+/// representing outstanding [sendMessage] requests.
+abstract class MessageBase<T extends Conversation> {
+  /// The Zulip message ID.
+  ///
+  /// If null, the message doesn't have an ID acknowledged by the server
+  /// (e.g.: a locally-echoed message).
+  int? get id;
+
+  final int senderId;
+  final int timestamp;
+
+  /// The conversation that contains this message.
+  ///
+  /// When implementing this, the return type should be either
+  /// [StreamConversation] or [DmConversation]; it should never be
+  /// [Conversation], because we expect a concrete subclass of [MessageBase]
+  /// to represent either a channel message or a DM message, not both.
+  T get conversation;
+
+  const MessageBase({required this.senderId, required this.timestamp});
+}
+
 /// As in the get-messages response.
 ///
 /// https://zulip.com/api/get-messages#response
-sealed class Message {
+sealed class Message<T extends Conversation> extends MessageBase<T> {
   // final String? avatarUrl; // Use [User.avatarUrl] instead; will live-update
   final String client;
   String content;
@@ -650,6 +675,7 @@ sealed class Message {
   @JsonKey(readValue: MessageEditState._readFromMessage, fromJson: Message._messageEditStateFromJson)
   MessageEditState editState;
 
+  @override
   final int id;
   bool isMeMessage;
   int? lastEditTimestamp;
@@ -660,14 +686,12 @@ sealed class Message {
   final int recipientId;
   final String senderEmail;
   final String senderFullName;
-  final int senderId;
   final String senderRealmStr;
 
   /// Poll data if "submessages" describe a poll, `null` otherwise.
   @JsonKey(name: 'submessages', readValue: _readPoll, fromJson: Poll.fromJson, toJson: Poll.toJson)
   Poll? poll;
 
-  final int timestamp;
   String get type;
 
   // final List<TopicLink> topicLinks; // TODO handle
@@ -717,14 +741,16 @@ sealed class Message {
     required this.recipientId,
     required this.senderEmail,
     required this.senderFullName,
-    required this.senderId,
+    required super.senderId,
     required this.senderRealmStr,
-    required this.timestamp,
+    required super.timestamp,
     required this.flags,
     required this.matchContent,
     required this.matchTopic,
   });
 
+  // TODO(dart): This has to be a static method, because factories/constructors
+  //   do not support type parameters: https://github.com/dart-lang/language/issues/647
   static Message fromJson(Map<String, dynamic> json) {
     final type = json['type'] as String;
     if (type == 'stream') return StreamMessage.fromJson(json);
@@ -762,7 +788,7 @@ enum MessageFlag {
 }
 
 @JsonSerializable(fieldRename: FieldRename.snake)
-class StreamMessage extends Message {
+class StreamMessage extends Message<StreamConversation> {
   @override
   @JsonKey(includeToJson: true)
   String get type => 'stream';
@@ -780,6 +806,7 @@ class StreamMessage extends Message {
   @JsonKey(includeToJson: true)
   String? get displayRecipient => conversation.displayRecipient;
 
+  @override
   @JsonKey(readValue: _readConversation, includeToJson: false)
   StreamConversation conversation;
 
@@ -816,7 +843,7 @@ class StreamMessage extends Message {
 }
 
 @JsonSerializable(fieldRename: FieldRename.snake)
-class DmMessage extends Message {
+class DmMessage extends Message<DmConversation> {
   @override
   @JsonKey(includeToJson: true)
   String get type => 'private';
@@ -834,6 +861,7 @@ class DmMessage extends Message {
   @JsonKey(name: 'display_recipient', toJson: _allRecipientIdsToJson, includeToJson: true)
   List<int> get allRecipientIds => conversation.allRecipientIds;
 
+  @override
   @JsonKey(name: 'display_recipient', fromJson: _conversationFromJson, includeToJson: false)
   final DmConversation conversation;
 
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -532,6 +532,65 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
   }
 }
 
+/// The name of a Zulip topic.
+// TODO(dart): Can we forbid calling Object members on this extension type?
+//   (The lack of "implements Object" ought to do that, but doesn't.)
+//   In particular an interpolation "foo > $topic" is a bug we'd like to catch.
+// TODO(dart): Can we forbid using this extension type as a key in a Map?
+//   (The lack of "implements Object" arguably should do that, but doesn't.)
+//   Using as a Map key is almost certainly a bug because it won't case-fold;
+//   see for example #739, #980, #1205.
+extension type const TopicName(String _value) {
+  /// The canonical form of the resolved-topic prefix.
+  // This is RESOLVED_TOPIC_PREFIX in web:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
+  static const resolvedTopicPrefix = '✔ ';
+
+  /// Pattern for an arbitrary resolved-topic prefix.
+  ///
+  /// These always begin with [resolvedTopicPrefix]
+  /// but can be weird and go on longer, like "✔ ✔✔ ".
+  // This is RESOLVED_TOPIC_PREFIX_RE in web:
+  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts#L4-L12
+  static final resolvedTopicPrefixRegexp = RegExp(r'^✔ [ ✔]*');
+
+  /// The string this topic is identified by in the Zulip API.
+  ///
+  /// This should be used in constructing HTTP requests to the server,
+  /// but rarely for other purposes.  See [displayName] and [canonicalize].
+  String get apiName => _value;
+
+  /// The string this topic is displayed as to the user in our UI.
+  ///
+  /// At the moment this always equals [apiName].
+  /// In the future this will become null for the "general chat" topic (#1250),
+  /// so that UI code can identify when it needs to represent the topic
+  /// specially in the way prescribed for "general chat".
+  // TODO(#1250) carry out that plan
+  String get displayName => _value;
+
+  /// The key to use for "same topic as" comparisons.
+  String canonicalize() => apiName.toLowerCase();
+
+  /// Whether the topic starts with [resolvedTopicPrefix].
+  bool get isResolved => _value.startsWith(resolvedTopicPrefix);
+
+  /// This [TopicName] plus the [resolvedTopicPrefix] prefix.
+  TopicName resolve() => TopicName(resolvedTopicPrefix + _value);
+
+  /// A [TopicName] with [resolvedTopicPrefixRegexp] stripped if present.
+  TopicName unresolve() =>
+    TopicName(_value.replaceFirst(resolvedTopicPrefixRegexp, ''));
+
+  /// Whether [this] and [other] have the same canonical form,
+  /// using [canonicalize].
+  bool isSameAs(TopicName other) => canonicalize() == other.canonicalize();
+
+  TopicName.fromJson(this._value);
+
+  String toJson() => apiName;
+}
+
 /// As in [StreamMessage.conversation] and [DmMessage.conversation].
 ///
 /// Different from [MessageDestination], this information comes from
@@ -700,65 +759,6 @@ enum MessageFlag {
   static final _byRawString = _$MessageFlagEnumMap.map((key, value) => MapEntry(value, key));
 
   String toJson() => _$MessageFlagEnumMap[this]!;
-}
-
-/// The name of a Zulip topic.
-// TODO(dart): Can we forbid calling Object members on this extension type?
-//   (The lack of "implements Object" ought to do that, but doesn't.)
-//   In particular an interpolation "foo > $topic" is a bug we'd like to catch.
-// TODO(dart): Can we forbid using this extension type as a key in a Map?
-//   (The lack of "implements Object" arguably should do that, but doesn't.)
-//   Using as a Map key is almost certainly a bug because it won't case-fold;
-//   see for example #739, #980, #1205.
-extension type const TopicName(String _value) {
-  /// The canonical form of the resolved-topic prefix.
-  // This is RESOLVED_TOPIC_PREFIX in web:
-  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts
-  static const resolvedTopicPrefix = '✔ ';
-
-  /// Pattern for an arbitrary resolved-topic prefix.
-  ///
-  /// These always begin with [resolvedTopicPrefix]
-  /// but can be weird and go on longer, like "✔ ✔✔ ".
-  // This is RESOLVED_TOPIC_PREFIX_RE in web:
-  //   https://github.com/zulip/zulip/blob/1fac99733/web/shared/src/resolved_topic.ts#L4-L12
-  static final resolvedTopicPrefixRegexp = RegExp(r'^✔ [ ✔]*');
-
-  /// The string this topic is identified by in the Zulip API.
-  ///
-  /// This should be used in constructing HTTP requests to the server,
-  /// but rarely for other purposes.  See [displayName] and [canonicalize].
-  String get apiName => _value;
-
-  /// The string this topic is displayed as to the user in our UI.
-  ///
-  /// At the moment this always equals [apiName].
-  /// In the future this will become null for the "general chat" topic (#1250),
-  /// so that UI code can identify when it needs to represent the topic
-  /// specially in the way prescribed for "general chat".
-  // TODO(#1250) carry out that plan
-  String get displayName => _value;
-
-  /// The key to use for "same topic as" comparisons.
-  String canonicalize() => apiName.toLowerCase();
-
-  /// Whether the topic starts with [resolvedTopicPrefix].
-  bool get isResolved => _value.startsWith(resolvedTopicPrefix);
-
-  /// This [TopicName] plus the [resolvedTopicPrefix] prefix.
-  TopicName resolve() => TopicName(resolvedTopicPrefix + _value);
-
-  /// A [TopicName] with [resolvedTopicPrefixRegexp] stripped if present.
-  TopicName unresolve() =>
-    TopicName(_value.replaceFirst(resolvedTopicPrefixRegexp, ''));
-
-  /// Whether [this] and [other] have the same canonical form,
-  /// using [canonicalize].
-  bool isSameAs(TopicName other) => canonicalize() == other.canonicalize();
-
-  TopicName.fromJson(this._value);
-
-  String toJson() => apiName;
 }
 
 @JsonSerializable(fieldRename: FieldRename.snake)
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 
+import '../../model/algorithms.dart';
 import 'events.dart';
 import 'initial_snapshot.dart';
 import 'reaction.dart';
@@ -531,6 +532,52 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
   }
 }
 
+/// As in [StreamMessage.conversation] and [DmMessage.conversation].
+///
+/// Different from [MessageDestination], this information comes from
+/// [getMessages] or [getEvents], identifying the conversation that contains a
+/// message.
+sealed class Conversation {}
+
+/// The conversation a stream message is in.
+@JsonSerializable(fieldRename: FieldRename.snake, createToJson: false)
+class StreamConversation extends Conversation {
+  int streamId;
+
+  @JsonKey(name: 'subject')
+  TopicName topic;
+
+  /// The name of the channel with ID [streamId] when the message was sent.
+  ///
+  /// The primary reference for the name of the channel is
+  /// the client's data structures about channels, at [streamId].
+  /// This value may be used as a fallback when the channel is unknown.
+  ///
+  /// This is non-null when found in a [StreamMessage] object in the API,
+  /// but may become null in the client's data structures,
+  /// e.g. if the message gets moved between channels.
+  @JsonKey(required: true, disallowNullValue: true)
+  String? displayRecipient;
+
+  StreamConversation(this.streamId, this.topic, {required this.displayRecipient});
+
+  factory StreamConversation.fromJson(Map<String, dynamic> json) =>
+    _$StreamConversationFromJson(json);
+}
+
+/// The conversation a DM message is in.
+class DmConversation extends Conversation {
+  /// The user IDs of all users in the conversation, sorted numerically.
+  ///
+  /// This lists the sender as well as all (other) recipients, and it
+  /// lists each user just once.  In particular the self-user is always
+  /// included.
+  final List<int> allRecipientIds;
+
+  DmConversation({required this.allRecipientIds})
+    : assert(isSortedWithoutDuplicates(allRecipientIds.toList()));
+}
+
 /// As in the get-messages response.
 ///
 /// https://zulip.com/api/get-messages#response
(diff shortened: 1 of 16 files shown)


lib/api/model/model.dart
```diff
@@ -531,10 +532,29 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
   }
 }
 
+/// A common class for messages that can appear in a [MessageList].
+abstract class DisplayableMessage<T extends MessageDestination> {
```
gnprice, line 536: I feel like the name "displayable message" doesn't really capture what this is about — in particular the term "displayable" doesn't much help in saying how this relates to the `Message` class.

A good name would be `BaseMessage`. Here's also some revised dartdoc:

```suggestion
/// A message or message-like object, for showing in a message list.
///
/// Other than [Message], we use this for "outbox messages",
/// representing outstanding [sendMessage] requests.
abstract class BaseMessage<T extends MessageDestination> {
```

Since you have a whole series of commits that use this name, doing the rename in a manual way might be a pain. Here's a command to do it automatically and I think quite painlessly:

```
$ FILTER_BRANCH_SQUELCH_WARNING=1 git filter-branch \
    --tree-filter 'perl -i -0pe s/DisplayableMessage/MessageBase/g $(git ls-files lib test)' \
    @~4..
Rewrite 282632f9fbafacfc53fa3629be22733ccfe7a791 (1/4) (0 seconds passed, remaining 0 predicted)
Rewrite 1e324588d6bf5791a50e2cd65d4175f4a903e3b0 (2/4) (1 seconds passed, remaining 1 predicted)
Rewrite 9b9265bfd4292728dfdabdea5756b8f1c3b4fd58 (2/4) (1 seconds passed, remaining 1 predicted)
Rewrite be02a05bdff9b4bc1f02a50ed2e4513efb92f305 (2/4) (1 seconds passed, remaining 1 predicted)

Ref 'refs/heads/main' was rewritten
```

Definitely use `git diff --stat -p @{1}` and `git range-diff origin @{1} @` to review the results.

The reason for that environment variable at the start of the command is that `git filter-branch` is deprecated these days. When I tried the command without it, I got:
```
WARNING: git-filter-branch has a glut of gotchas generating mangled history
	 rewrites.  Hit Ctrl-C before proceeding to abort, then use an
	 alternative filtering tool such as 'git filter-repo'
	 (https://github.com/newren/git-filter-repo/) instead.  See the
	 filter-branch manual page for more details; to squelch this warning,
	 set FILTER_BRANCH_SQUELCH_WARNING=1.
```

The warning is accurate. But this rewrite is simple enough that I don't think any gotchas apply; and it's small enough you can review the results to confirm that. Given that that's the case, and that I already know how to use `git filter-branch`, I went with just squelching the warning.

If we needed to do something more complex, I would invest the time to do it with the newer out-of-tree tool `git filter-repo`. (And if it were a big job, I'd also be sure to ask Anders if he had any other recommendations.)

gnprice: Or perhaps better: `MessageBase`. This type alone isn't yet even a message; it's just the base for a message.

PIG208: Yeah. Thanks for the script!

One other thing I'm thinking of renaming is `destination`. The Zulip API uses `to` for [send-message](https://zulip.com/api/send-message#parameter-to), Because this adds a level of nesting, perhaps shortening it to two letters will be more readable? So it might be used like `message.to.streamId`, `message.to.userIds` or `final destination = message.to`. But we don't seem to do that in general when naming things.

lib/model/narrow.dart
```diff
@@ -260,11 +264,12 @@ class DmNarrow extends Narrow implements SendableNarrow {
   late final String _key = otherRecipientIds.join(',');
 
   @override
-  bool containsMessage(Message message) {
-    if (message is! DmMessage) return false;
-    if (message.allRecipientIds.length != allRecipientIds.length) return false;
+  bool containsMessage(MessageBase message) {
+    final destination = message.destination;
+    if (destination is! DmDestination) return false;
+    if (destination.userIds.length != allRecipientIds.length) return false;
```
gnprice, line 270: Seeing this change and many of the other instances of this change, it makes me nervous that we now have something that's vague about exactly which user IDs are included and we're expecting it to match up with something that's specifically *all* the recipient IDs.

As a comment on `DmNarrow` says, Zulip has many ways of representing a DM conversation. In particular the Zulip server API has at least *four* different conventions for whether the self-user is included in the list. (Always; never; only for self-1:1; only for self-1:1 and groups.) This was a recurring source of bugs and confusion in zulip-mobile, until I invested substantial effort in cleaning it up with the abstractions found now in `src/utils/recipient.js`. That's the reason that `DmMessage` and `DmNarrow` have the very explicitly-named `allRecipientIds`, as well as `DmNarrow.otherRecipientIds` for when that's desired.

So I'd like to find a solution that preserves that explicitness about the semantics of each list of users.

I think the key here is that `MessageDestination` isn't quite the right class to be using for this purpose. Its doc says:
```dart
/// Which conversation to send a message to, in [sendMessage].
///
/// This is either a [StreamDestination] or a [DmDestination].
sealed class MessageDestination {
```

Similarly the two subclasses say things like:
```dart
/// A DM conversation, for specifying to [sendMessage].
///
/// The server accepts a list of Zulip API emails as an alternative to
/// a list of user IDs, but this binding currently doesn't.
class DmDestination extends MessageDestination {
```

So they're specifically about [sendMessage] — their job is to describe a parameter in a request to that endpoint. The information one says in [sendMessage] about where a message should go naturally has a lot in common with the information the server says in a [getMessages] or [getEvents] response about where a message did go… but it's also natural that it's not exactly the same information.

Let's therefore make a separate handful of small classes that are specifically for the information that belongs on [Message]. I think we can use the nice short/general name `Recipient` for this (and then `StreamRecipient` and `DmRecipient`).

Then the fields on `StreamMessage` and `DmMessage` can move verbatim onto the `Recipient` subclasses. And maybe leave behind getters on the message subclasses, which just proxy through to there, for convenience / to reduce how much code needs to churn.

gnprice: (I guess there's already a `DmRecipient`. It can get renamed first to something more specific, like `DmDisplayRecipient`. It's already quite local in how it's used.)

PIG208: I suppose that we don't want to move `displayRecipient` to `DmRecipient` (after renaming the original class to `DmDisplayRecipient`), right?

It shouldn't be necessary for `sendMessage` to construct a `DmDisplayRecipient` when it only has convenient access to a `DmDestination` and that other fields on `DmDisplayRecipient` are mostly unused.

I think this means that `DmMessage.recipient` should remain as a getter, much like how `DmMessage.destination` is in the current revision, but returns a `DmRecipient` computed from `DmMessage.allRecipientIds`.

gnprice: Hmm good question.

It looks like in main we never actually consult `DmMessage.displayRecipient`, except in its own test and in the `allRecipientIds` getter. So in any case we shouldn't let it get in the way of a structure we otherwise like — we can always just remove it instead, and deserialize straight to the list of IDs.

One way the further detail could in principle be useful is in `DmRecipientHeader`, if a user is unknown in the store, we could fall back to their name from the `DmDisplayRecipient` instead of to "(unknown user)". But we don't do that now, and it seems pretty marginal in value. So if the continued existence of `DmDisplayRecipient` feels like it's getting in the way, let's just rip it out, and we can decide how to add it back if we do ever implement that fallback functionality.

lib/api/model/model.dart
```diff
@@ -765,78 +765,35 @@ class StreamMessage extends Message {
...
-
 @JsonSerializable(fieldRename: FieldRename.snake)
 class DmMessage extends Message {
   @override
   @JsonKey(includeToJson: true)
   String get type => 'private';
 
-  /// The `display_recipient` from the server, sorted by user ID numerically.
+  /// The user IDs of all users in the thread, sorted numerically, as in
+  /// the `display_recipient from the server.
```
rajveermalviya, line 848: nit:
```suggestion
  /// the `display_recipient` from the server.
```

rajveermalviya (approved):
Thanks for working on this @PIG208! Just one small nit, otherwise looks good.

lib/api/model/model.dart
```diff
@@ -619,7 +682,9 @@ sealed class Message {
     required this.matchTopic,
   });
 
-  factory Message.fromJson(Map<String, dynamic> json) {
+  // TODO(dart): This has to be a static method, because factories/constructors
+  //   do not support type parameters: https://github.com/dart-lang/language/issues/647
+  static Message fromJson(Map<String, dynamic> json) {
```
gnprice, line 687: nit:

    api [nfc]: Make Message.fromJson static

As a (factory) constructor, I'd say it's static already — it's certainly not *non*-static, i.e. it's not an instance method.

Instead can clarify as:

    api [nfc]: Make Message.fromJson a plain static method

lib/api/model/model.dart
```diff
@@ -766,77 +839,38 @@ class StreamMessage extends Message {
 }
 
 @JsonSerializable(fieldRename: FieldRename.snake)
-class DmRecipient {
```
gnprice, line 769: nit:

    This is unused in app code.  This is mostly NFC except that we will
    ignore fields other than "id" from the list of objects from
    "display_recipient" on message events.

s/message events/message objects from the server/

These objects also notably appear in getMessages responses.

lib/api/model/model.dart
```diff
@@ -1,5 +1,7 @@
 import 'package:json_annotation/json_annotation.dart';
 
+import '../../model/algorithms.dart';
+import '../route/messages.dart';
```
gnprice, line 4:
```suggestion
```

This is a wrong-direction import: specific routes should import from the model types, but not vice versa.

I think this was here only for a reference in dartdoc. It's acceptable to have those links not work when the alternative is a wrong-way import.

lib/api/model/model.dart
```diff
@@ -531,10 +533,68 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+
+/// The recipient of a stream message.
+@JsonSerializable(fieldRename: FieldRename.snake, createToJson: false)
+class StreamRecipient extends Recipient {
+  StreamRecipient(this.streamId, this.topic);
+
+  int streamId;
+
+  @JsonKey(name: 'subject')
+  TopicName topic;
```
gnprice, line 551: nit: let's move the definition of TopicName up to just above these Recipient classes; it had been where it is to be just above StreamMessage, but now StreamRecipient is the more pertinent primary place it's used

lib/api/model/model.dart
```diff
@@ -531,10 +533,68 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
   }
 }
 
+/// As in [MessageBase.recipient].
+///
+/// Different from [MessageDestination], this information comes from
+/// [getMessages] or [getEvents], identifying the conversation that contains a
+/// message.
+sealed class Recipient {}
```
gnprice, line 541: I'm having second thoughts about this name, even though I think I suggested it myself.

The trouble is that "recipient" is already the name of a concept in the Zulip data model — and that concept doesn't agree with this one. For a channel message, the recipient (as that term is understood in the Zulip database and server code) is the channel as a whole, not the individual conversation within it.

There's even a conceptual reason that's the right way to use that term: the unit of who's subscribed to what is the channel, not the conversation. A message sent to a given channel will have the same set of users who "receive" it, in the sense that they get UserMessage rows or that they have permission to read the message, regardless of what the topic of the message is.

I think `Conversation` would be a good name. And it's only 3 letters longer than `Recipient`.

lib/api/model/model.dart
```diff
@@ -726,14 +791,23 @@ class StreamMessage extends Message {
...
-  TopicName topic;
+  @JsonKey(name: 'subject', includeToJson: true)
+  TopicName get topic => recipient.topic;
+
+  @override
+  @JsonKey(readValue: _readRecipient, includeToJson: false)
+  StreamRecipient recipient;
+
+  static Map<String, dynamic> _readRecipient(Map<dynamic, dynamic> json, String key) {
+    return {'stream_id': json['stream_id'], 'subject': json['subject']};
```
gnprice, line 809: This can just return `json`. That way we skip allocating a new Map.

lib/api/model/model.dart
```diff
@@ -531,10 +533,68 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+    : assert(isSortedWithoutDuplicates(allRecipientIds.toList()));
+
+  /// The user IDs of all users in the thread, sorted numerically.
+  ///
+  /// This lists the sender as well as all (other) recipients, and it
+  /// lists each user just once.  In particular the self-user is always
+  /// included.
+  ///
+  /// This is required to have an efficient `length`.
+  final List<int> allRecipientIds;
```
gnprice, line 569:
```suggestion
  final List<int> allRecipientIds;
```

For a `List`, it's expected that `length` will be efficient.

(I guess the `List.length` doc doesn't say that. But lots and lots of code assumes it, because it's true for the built-in lists and more generally for any array data structure. Similarly lots of code assumes that `operator []` is efficient on a `List`.)

lib/api/model/model.dart
```diff
@@ -531,10 +533,68 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+
+  int get senderId;
+  int get timestamp;
+
+  /// The recipient of this message.
+  // When implementing this, the return type should be either [StreamRecipient]
+  // or [DmRecipient]; it should never be [Recipient], because we
+  // expect a concrete subclass of [MessageBase] to represent either
+  // a channel message or a DM message, not both.
+  T get recipient;
```
gnprice, line 591: This looks like material for a doc comment. In general, instructions about how subclasses should implement something go in doc comments. (That's common in Flutter APIs.)

lib/api/model/model.dart
```diff
@@ -531,10 +533,68 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+/// representing outstanding [sendMessage] requests.
+abstract class MessageBase<T extends Recipient> {
+  /// The Zulip message ID.
+  ///
+  /// If null, the message doesn't have an ID acknowledged by the server
+  /// (e.g.: a locally-echoed message).
+  int? get id;
+
+  int get senderId;
+  int get timestamp;
```
gnprice, line 584: These always get overridden as fields and in the same way — so they can just be fields here.

lib/api/model/model.dart
gnprice, line 591: I think this can also just be a field.

lib/api/model/model.dart
```diff
@@ -531,10 +533,68 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+  ///
+  /// This is required to have an efficient `length`.
+  final List<int> allRecipientIds;
+}
+
+/// A message or message-like object, for showing in a message list.
+///
+/// Other than [Message], we use this for "outbox messages",
+/// representing outstanding [sendMessage] requests.
+abstract class MessageBase<T extends Recipient> {
```
gnprice, line 576: api [nfc]: Pull out MessageBase and Recipient

This can be split into two commits:
* One introduces Recipient (or Conversation) and subclasses, and switches StreamMessage and DmMessage to use them;
* One introduces MessageBase.

I think those are fairly conceptually distinct changes, even though the first is a prerequisite for what we want from the second.

gnprice (commented):
Thanks! Here's a review from reading the first two commits and most of the third:
66ede2252 api [nfc]: Make Message.fromJson static
e35d0be4e api: Drop DmRecipient
bc2f58412 api [nfc]: Pull out MessageBase and Recipient

PIG208:
Thanks! Updated the PR.

lib/api/model/model.dart
```diff
@@ -531,10 +532,132 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+  /// [Conversation], because we expect a concrete subclass of [MessageBase]
+  /// to represent either a channel message or a DM message, not both.
+  final T conversation;
+}
+
 /// As in the get-messages response.
 ///
 /// https://zulip.com/api/get-messages#response
-sealed class Message {
+sealed class Message<T extends Conversation> implements MessageBase<T> {
```
gnprice, line 660: Can this be "extends"?

Sorry I wasn't clear — that's what I had in mind at https://github.com/zulip/zulip-flutter/pull/1463#discussion_r2038629322, as "implements" with a field on the base class doesn't make a lot of sense. (For a final field I think it ends up being no different from implementing a getter.)

lib/model/message_list.dart
```diff
@@ -155,7 +155,7 @@ mixin _MessageSequence {
         }
       case MessageListRecipientHeaderItem(:var message):
       case MessageListDateSeparatorItem(:var message):
-        return (message.id <= messageId) ? -1 : 1;
+        return message.id != null && message.id! <= messageId ? -1 : 1;
```
gnprice, line 158: nit: I think this condition would be clearer by splitting up onto two lines:
```suggestion
        if (message.id == null)                  return 1;
        return message.id! <= messageId ? -1 : 1;
```

(then the spacing is to match the alignment of other cases in this method)

lib/widgets/message_list.dart
```diff
@@ -901,23 +902,29 @@ class _MarkAsReadAnimationState extends State<MarkAsReadAnimation> {
...
-    };
+    switch (message) {
+      case MessageBase<StreamConversation>():
+        return StreamMessageRecipientHeader(message: message, narrow: narrow);
+      case MessageBase<DmConversation>():
+        return DmRecipientHeader(message: message, narrow: narrow);
+      case MessageBase<Conversation>():
+        assert(false, 'Unexpected concrete subclass of MessageBase<Conversation>:'
+                      ' ${objectRuntimeType(message, 'MessageBase<Conversation>')}');
+        return SizedBox.shrink();
```
gnprice, line 919: I think we can handle this with more confidence:
```suggestion
      case MessageBase<Conversation>():
        throw StateError('Bad concrete subclass of MessageBase');
```

We're not going to be writing lots more of these subclasses, so there won't be many opportunities to get this wrong.

That also lets this go back to a switch-expression.

lib/widgets/message_list.dart
```diff
@@ -1073,16 +1080,17 @@ class StreamMessageRecipientHeader extends StatelessWidget {
     if (!_containsDifferentChannels(narrow)) {
       streamWidget = const SizedBox(width: 16);
     } else {
-      final stream = store.streams[message.streamId];
+      final stream = store.streams[streamId];
+      final message = this.message;
       final streamName = stream?.name
-        ?? message.displayRecipient
+        ?? (message is StreamMessage ? message.displayRecipient : null)
```
gnprice, line 1086: Can this field move instead onto StreamConversation?

If nothing else, I think having this condition here calls for a comment — this is `StreamMessageRecipientHeader`, so it should already be known to be a `StreamMessage`, right? (The answer would be that it might be an outbox-message instead; but that's not something the reader of this widget will necessarily have top of mind, and that's a good thing — if we've designed these internal APIs well, then we should be able to read through most of our code without really thinking about the fact that there are outbox-messages.)

lib/model/narrow.dart
```diff
@@ -260,11 +265,11 @@ class DmNarrow extends Narrow implements SendableNarrow {
   late final String _key = otherRecipientIds.join(',');
 
   @override
-  bool containsMessage(Message message) {
-    if (message is! DmMessage) return false;
-    if (message.allRecipientIds.length != allRecipientIds.length) return false;
+  bool containsMessage(MessageBase message) {
+    if (message is! MessageBase<DmConversation>) return false;
+    if (message.conversation.allRecipientIds.length != allRecipientIds.length) return false;
```
gnprice, line 270: This is probably a bit cleaner with a `conversation` local, like the ChannelNarrow and TopicNarrow implementations above.

gnprice (commented):
Thanks for the revision! I think that split of the commit definitely helps make it easier to read, too.

Here's a review after having read the full branch now.

lib/api/model/model.dart
PIG208: Oh, I see. Then https://github.com/zulip/zulip-flutter/pull/1463#discussion_r2038629322 might not work, since we want to [override](https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=overridden_fields#overridden_fields) `conversation` in the `Message<T>` subclasses for (de)serialization.

lib/widgets/message_list.dart
PIG208: Yeah, we can do that. I think one tricky thing might be coming up with an abstraction that fits both kinds of stream messages. We document the choice of making `displayRecipient` nullable that only makes sense for `StreamMessage`. I think we can just point that comment to the subclass, and have a more generic dartdoc for the field. Maybe something like:

```dart
  /// The name of the stream, found on stream message objects from the server.
  ///
  /// This is not updated when its name changes.  Consider using [streamId]
  /// instead to lookup stream name from the store.
  // This is not nullable API-wise for [StreamMessage], but if the message moves
  // across channels, [displayRecipient] still refers to the original channel
  // and it has to be invalidated.
  @JsonKey(required: true, disallowNullValue: true)
  String? displayRecipient;
```

lib/api/model/model.dart
gnprice: Hmm I see, because we need to annotate with JsonKey.

Maybe turn `conversation` back into a getter, then?

PIG208:
Thanks for the review! The PR is ready again, with a follow-up to https://github.com/zulip/zulip-flutter/pull/1463#discussion_r2040480875.

lib/api/model/model.dart
```diff
@@ -531,10 +532,143 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+///
+/// Different from [MessageDestination], this information comes from
+/// [getMessages] or [getEvents], identifying the conversation that contains a
+/// message.
+sealed class Conversation {}
+
+/// The conversation a stream message is in.
+@JsonSerializable(fieldRename: FieldRename.snake, createToJson: false)
+class StreamConversation extends Conversation {
+  StreamConversation(this.streamId, this.topic);
```
gnprice, line 604:
```suggestion
  StreamConversation(this.streamId, this.topic, {required this.displayRecipient});
```

That way it's possible to construct one of these with all one's desired values directly.

lib/api/model/model.dart
```diff
@@ -531,10 +532,143 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
 
+/// The name of a Zulip topic.
+// TODO(dart): Can we forbid calling Object members on this extension type?
+//   (The lack of "implements Object" ought to do that, but doesn't.)
+//   In particular an interpolation "foo > $topic" is a bug we'd like to catch.
+// TODO(dart): Can we forbid using this extension type as a key in a Map?
+//   (The lack of "implements Object" arguably should do that, but doesn't.)
+//   Using as a Map key is almost certainly a bug because it won't case-fold;
+//   see for example #739, #980, #1205.
+extension type const TopicName(String _value) {
```
gnprice, line 543: nit: how about pulling out this move as a separate prep commit? I think that'd make the Conversation commit a bit easier to read.

(In particular the three Conversation classes would then appear in the same screenful as the changes to at least StreamMessage, rather than have the old copy of TopicName appear in between.)

lib/api/model/model.dart
gnprice, line 604: nit: follow the other API types in ordering fields before constructors

(It's possible it'd be better to switch them all sometime to match what we do for widgets and the rest of our codebase, which in turn follows Flutter's style. But I think for a "data" class, in isolation, the fields-first ordering is better anyway; and there isn't really a consistency gain from having a few API classes this way and not the others.)

lib/model/message.dart
```diff
@@ -211,14 +211,14 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
       }
 
       if (newStreamId != origStreamId) {
-        message.streamId = newStreamId;
+        message.conversation.streamId = newStreamId;
         // See [StreamMessage.displayRecipient] on why the invalidation is
         // needed.
```
gnprice, line 216: nit: reference needs update

test/api/model/model_checks.dart
```diff
@@ -36,23 +52,15 @@ extension MessageChecks on Subject<Message> {
...
 }
 
-extension TopicNameChecks on Subject<TopicName> {
-  Subject<String> get apiName => has((x) => x.apiName, 'apiName');
-  Subject<String> get displayName => has((x) => x.displayName, 'displayName');
-}
-
 extension StreamMessageChecks on Subject<StreamMessage> {
-  Subject<String?> get displayRecipient => has((e) => e.displayRecipient, 'displayRecipient');
   Subject<int> get streamId => has((e) => e.streamId, 'streamId');
```
gnprice, line 64: nit: may as well keep this checks-getter, like the others — the others are definitely handy, and keeping it to match those seems the most symmetrical

test/model/message_list_test.dart
```diff
@@ -526,7 +526,7 @@ void main() {
     test('unmute a topic -> refetch from scratch', () => awaitFakeAsync((async) async {
       await prepare(narrow: const CombinedFeedNarrow());
       await prepareMutes(true);
-      final messages = [
+      final messages = <Message>[
```
gnprice, line 529: nit in commit message:

    This requiring specifying the element type of `List`s in some tests (or
    in some cases, the type of a variable declared).
    
    This is a side effect of making `StreamMessage` and `DmMessage` extend
    `Message<T>` with different `T`'s. When both appear in the same `List`,
    […]

This is probably clearest if joined as one paragraph — that way it becomes clear that the "This" in that second sentence refers to the need to specify element types, unlike the "This" in the sentence before it :slightly_smiling_face:

lib/widgets/message_list.dart
```diff
@@ -1053,11 +1057,11 @@ class StreamMessageRecipientHeader extends StatelessWidget {
     final designVariables = DesignVariables.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
 
-    final topic = message.topic;
+    final StreamConversation(:streamId, :topic) = message.conversation;
```
gnprice, line 1060: nit:
```suggestion
    final streamId = message.conversation.streamId;
    final topic = message.conversation.topic;
```

This way is hardly any longer than the pattern-matching, in this case; and it avoids the bug where when a local is bound by a pattern, it breaks Android Studio's feature of showing references when you put your cursor on the variable.

(This is a long function, so that feature is especially critical here for reading it.)

lib/model/narrow.dart
```diff
@@ -343,7 +350,8 @@ class StarredMessagesNarrow extends Narrow {
   ApiNarrow apiEncode() => [ApiNarrowIs(IsOperand.starred)];
 
   @override
-  bool containsMessage(Message message) {
+  bool containsMessage(MessageBase message) {
+    if (message is! Message) return false;
     return message.flags.contains(MessageFlag.starred);
```
gnprice, line 355: This and the one for MentionsNarrow are a bit nontrivial, semantically — we're saying that pseudo-messages (i.e. outbox messages, as that's the one example of pseudo-messages we plan to have) are never in these narrows.

That feels like it should get tested — just a test case each. As of this commit, the test file could make just its own ad-hoc subclass of `MessageBase`. Later it can drop that in favor of OutboxMessage.

lib/model/narrow.dart
```diff
@@ -105,9 +106,10 @@ class TopicNarrow extends Narrow implements SendableNarrow {
   TopicNarrow sansWith() => TopicNarrow(streamId, topic);
 
   @override
-  bool containsMessage(Message message) {
-    return (message is StreamMessage
-      && message.streamId == streamId && message.topic == topic);
+  bool containsMessage(MessageBase message) {
+    final conversation = message.conversation;
+    return conversation is StreamConversation
```
gnprice, line 111: And similarly I guess these other three `containsMessage` implementations should gain a test case each to verify that they *do* include pseudo-messages when the conversation matches. Or more precisely two cases each: one positive, one negative, to show that the result depends on the conversation.

lib/widgets/message_list.dart
```diff
@@ -1073,16 +1077,16 @@ class StreamMessageRecipientHeader extends StatelessWidget {
...
+        ?? message.conversation.displayRecipient
         ?? zulipLocalizations.unknownChannelName; // TODO(log)
 
       streamWidget = GestureDetector(
         onTap: () => Navigator.push(context,
           MessageListPage.buildRoute(context: context,
-            narrow: ChannelNarrow(message.streamId))),
-        onLongPress: () => showChannelActionSheet(context, channelId: message.streamId),
+            narrow: ChannelNarrow(streamId))),
+        onLongPress: () => showChannelActionSheet(context, channelId: streamId),
```
gnprice, line 1089: This hoists streamId in StreamMessageRecipientHeader.build,
    where we used to access streamId via message instead from the
    onLongPress callback.  Because Message.streamId only changes on
    message moves, we expect the build method to be called again, so
    this should be fine.

Yeah — and if the build method weren't called again, so that this did cause a behavior change, then the old behavior would have been buggy.

In general, a callback that responds to input from the user should always be based on the same values that appeared in the UI the user saw.

lib/model/message_list.dart
```diff
@@ -155,7 +155,8 @@ mixin _MessageSequence {
         }
       case MessageListRecipientHeaderItem(:var message):
       case MessageListDateSeparatorItem(:var message):
-        return (message.id <= messageId) ? -1 : 1;
+        if (message.id == null)                  return 1;
```
gnprice, line 158: This should also get a test.

This one might be less convenient to write a test for at this stage. Could settle for a `TODO(#1441): test` comment, and add the test in a later PR when we have OutboxMessage appearing in these data structures.

gnprice (commented):

All small at this point, except that I realized the last commit needs some tests. If those feel annoying to write at this point, could just drop that commit for now and save for a later PR — it's small compared to the rest of this PR.

PIG208:
Thanks for the review! Updated the PR and left out the test for `lib/model/message_list.dart` (https://github.com/zulip/zulip-flutter/pull/1463#discussion_r2043393977). I added `_TestMessage` and friends for testing `containsMessage` (originally used the name `PseudoMessage`, which is a bit longer).

lib/api/model/model.dart
```diff
@@ -531,10 +532,143 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
...
+    : assert(isSortedWithoutDuplicates(allRecipientIds.toList()));
+}
+
+/// A message or message-like object, for showing in a message list.
+///
+/// Other than [Message], we use this for "outbox messages",
+/// representing outstanding [sendMessage] requests.
+abstract class MessageBase<T extends Conversation> {
+  const MessageBase({
+    required this.senderId,
```
gnprice, line 646: nit: bump https://github.com/zulip/zulip-flutter/pull/1463#discussion_r2043353099

test/model/narrow_test.dart
```diff
@@ -22,13 +48,57 @@ void main() {
...
+      final otherStream = eg.stream();
+      final narrow = eg.topicNarrow(stream.streamId, 'topic');
+      check(narrow.containsMessage(
+        eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]))).isFalse();
+      check(narrow.containsMessage(
+        eg.streamMessage(stream: stream,      topic: 'topic2'))).isFalse();
+      check(narrow.containsMessage(
+        eg.streamMessage(stream: otherStream, topic: 'topic'))).isFalse();
+      check(narrow.containsMessage(
+        eg.streamMessage(stream: stream,      topic: 'topic'))).isTrue();
```
gnprice, line 91: nit: swap the middle two cases to make the order slightly more logical: a DM is farthest away from belonging in this narrow, then a message to some other channel, then one to a different topic in the same narrow, then finally one to this same conversation

lib/model/message_list.dart
```diff
@@ -155,7 +155,9 @@ mixin _MessageSequence {
         }
       case MessageListRecipientHeaderItem(:var message):
       case MessageListDateSeparatorItem(:var message):
-        return (message.id <= messageId) ? -1 : 1;
+        // TODO(#1441): test
+        if (message.id == null)                  return 1;
```
gnprice, line 159: nit:
```suggestion
        if (message.id == null)                  return 1;  // TODO(#1441): test
```

That makes it clear the comment applies to that line and not the whole case. (I believe we already have a test covering the existing line below; if not, that's independent of #1441.)

gnprice (commented):
Thanks! This all looks good; I like the structure of the new tests. Just a few nits.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Looks good; merging.

PR #1464, opened 2025-04-04
store [nfc]: Add PerAccountStoreBase to provide `connection` to substores

chrisbobbe:
And demonstrate its use, with MessageStoreImpl.

This will be a convenient way to provide other common things to substores without needing boilerplate on each substore for each thing.

Co-authored-by: Greg Price <greg@zulip.com>


gnprice:
Thanks! Looks good; merging.

PR #1466, opened 2025-04-04
local echo (4/n): Migrate more fields to PerAccountStoreBase

PIG208:
MessageStore will need to access `queueId` and `selfUserId` later for outbox message support.

### DIFF
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -338,6 +338,7 @@ class CorePerAccountStore {
   CorePerAccountStore._({
     required GlobalStore globalStore,
     required this.connection,
+    required this.queueId,
     required this.accountId,
     required this.selfUserId,
   }) : _globalStore = globalStore,
@@ -345,6 +346,7 @@ class CorePerAccountStore {
 
   final GlobalStore _globalStore;
   final ApiConnection connection; // TODO(#135): update zulipFeatureLevel with events
+  final String queueId;
   final int accountId;
 
   // This isn't strictly needed as a field; it could be a getter
@@ -369,6 +371,8 @@ abstract class PerAccountStoreBase {
   GlobalStore get _globalStore => _core._globalStore;
 
   ApiConnection get connection => _core.connection;
+
+  String get queueId => _core.queueId;
 
   ////////////////////////////////
   // Data attached to the realm or the server.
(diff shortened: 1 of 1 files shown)


gnprice:
I see we were thinking in a similar direction today :slightly_smiling_face:. I just sent #1467 which does the `selfUserId` portion of this (including some changes borrowed from this PR), as well as a number of other fields. How about you rebase the `queueId` commit atop that?

PIG208:
Thanks! This has been rebased atop #1467.

chrisbobbe:
#1467 has been merged, and GitHub says there are some conflicts; could you rebase again?

lib/model/store.dart
```diff
@@ -384,6 +386,8 @@ abstract class PerAccountStoreBase {
   ////////////////////////////////
   // Data attached to the self-account on the realm.
 
+  String get queueId => _core.queueId;
```
chrisbobbe, line 389: Is it intentional to move this from the section

```dart
  ////////////////////////////////
  // Where data comes from in the first place.
```

to the section

```dart
  ////////////////////////////////
  // Data attached to the self-account on the realm.
```

?

chrisbobbe (commented):
Thanks! Small comment below.

lib/model/store.dart
PIG208: I want to position this after `connection` and before `accountId`, and I feel that it fits a bit better as it is associated with the self-account. But keeping it in the original section works too (proximity to `UpdateMachine`, `queueId`'s old home, was a reason why it was there).

chrisbobbe: Thanks for explaining; otherwise it looks like an accident 🙂 perhaps Greg has thoughts about it.

gnprice: Yeah, I think up next to `connection` in the "Where data comes from in the first place" section would be better.

I see the queue ID as much more like the connection than it is like the account. It's part of "where data comes from" — each `getEvents` call passes it in order to get the right events.

Conversely it's not really "attached to" the self-account, in that there may be zero, one, or many queues at any given time for a given account and the queues are fairly ephemeral.

gnprice:
Thanks, and thanks @chrisbobbe for the review!

LGTM modulo the thread above.

lib/model/store.dart
PIG208: Makes sense. Move it up to be right after `connection`.

PIG208:
Thanks! This has been updated.

gnprice:
Thanks! Looks good; merging.

PR #1471, opened 2025-04-10
Edit-message (4/n): Misc. preparation for edit-message feature

chrisbobbe:
Here's a small batch of miscellaneous prep work for the edit-message feature, #126.

### DIFF
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -308,16 +308,17 @@ class ChooseAccountPage extends StatelessWidget {
         trailing: MenuAnchor(
           menuChildren: [
             MenuItemButton(
-              onPressed: () {
-                showSuggestedActionDialog(context: context,
+              onPressed: () async {
+                final dialog = showSuggestedActionDialog(context: context,
                   title: zulipLocalizations.logOutConfirmationDialogTitle,
                   message: zulipLocalizations.logOutConfirmationDialogMessage,
                   // TODO(#1032) "destructive" style for action button
-                  actionButtonText: zulipLocalizations.logOutConfirmationDialogConfirmButton,
-                  onActionButtonPress: () {
-                    // TODO error handling if db write fails?
-                    logOutAccount(GlobalStoreWidget.of(context), accountId);
-                  });
+                  actionButtonText: zulipLocalizations.logOutConfirmationDialogConfirmButton);
+                if (await dialog.result == true) {
+                  if (!context.mounted) return;
+                  // TODO error handling if db write fails?
+                  unawaited(logOutAccount(GlobalStoreWidget.of(context), accountId));
+                }
               },
               child: Text(zulipLocalizations.chooseAccountPageLogOutButton)),
           ],
--- a/lib/widgets/dialog.dart
+++ b/lib/widgets/dialog.dart
@@ -19,13 +19,29 @@ Widget _dialogActionText(String text) {
 
 /// Tracks the status of a dialog, in being still open or already closed.
 ///
+/// Use [T] to identify the outcome of the interaction:
+/// - Pass `void` for an informational dialog with just the option to dismiss.
+/// - For confirmation dialogs with an option to dismiss
+///   plus an option to proceed with an action, pass `bool`.
+///   The action button should pass true for Navigator.pop's `result` argument.
+/// - For dialogs with an option to dismiss plus multiple other options,
+///   pass a custom enum.
+/// For the latter two cases, a cancel button should call Navigator.pop
+/// with null for the `result` argument, to match what Flutter does
+/// when you dismiss the dialog by tapping outside its area.
+///
 /// See also:
 ///  * [showDialog], whose return value this class is intended to wrap.
-class DialogStatus {
+class DialogStatus<T> {
   const DialogStatus(this.result);
 
   /// Resolves when the dialog is closed.
-  final Future<void> result;
+  ///
+  /// If this completes with null, the dialog was dismissed.
+  /// Otherwise, completes with a [T] identifying the interaction's outcome.
+  ///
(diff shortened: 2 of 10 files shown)


lib/widgets/dialog.dart
```diff
@@ -21,11 +21,11 @@ Widget _dialogActionText(String text) {
 ///
 /// See also:
 ///  * [showDialog], whose return value this class is intended to wrap.
-class DialogStatus {
+class DialogStatus<T> {
```
PIG208, line 24: It might be helpful to document the meaning of the type parameter. Specially, from `showDialog`, this part seems relevant:

```dart
/// If the application has multiple [Navigator] objects, it may be necessary to
/// call `Navigator.of(context, rootNavigator: true).pop(result)` to close the
/// dialog rather than just `Navigator.pop(context, result)`.
///
/// Returns a [Future] that resolves to the value (if any) that was passed to
/// [Navigator.pop] when the dialog was closed.
```

We just need to adapt it to fit our wrapper class.

PIG208 (commented):
Thanks! This looks good to me overall. Left a small comment, and marking for Greg's review.

test/widgets/dialog_test.dart
```diff
@@ -26,4 +26,36 @@ void main() {
...
+      await tester.pump();
+      final element = tester.element(find.byType(Placeholder));
+
+      final dialog = showSuggestedActionDialog(context: element,
+        title: 'Continue?',
+        message: 'Do the thing?',
+        actionButtonText: 'Sure',);
+      await tester.pump();
+      await tester.tap(find.text('Sure'));
+      await check(dialog.closed).completes((it) => it.equals(SuggestedActionDialogResult.doAction));
```
gnprice, line 43: nit: line too long: key information past 80 columns (namely the only information that distinguishes this check from its opposite in the other test case :slightly_smiling_face:)

gnprice (commented):
Thanks! Generally this looks good; a couple of small comments (adding to Zixuan's above).

chrisbobbe:

lib/widgets/dialog.dart
```diff
@@ -61,28 +70,41 @@ DialogStatus showErrorDialog({
...
         TextButton(
-          onPressed: () {
-            onActionButtonPress();
-            Navigator.pop(context);
-          },
+          onPressed: () => Navigator.pop<SuggestedActionDialogResult>(context,
+            SuggestedActionDialogResult.doAction),
           child: _dialogActionText(actionButtonText ?? zulipLocalizations.dialogContinue)),
       ]));
+  return DialogStatus(_wrapFutureExpectingNonNull(future));
```
gnprice, line 95: Ah interesting.

What happens if the user tries to dismiss the dialog without tapping either button? E.g. by tapping outside the dialog's area (on the obscured remainder of the page that's under the dialog), or by using the Android "back" gesture/button.

I think we'd want to have those work, and have the same effect as tapping the "Cancel" button.

chrisbobbe: Indeed, thanks for catching this—addressing this in the new revision.

chrisbobbe:

lib/widgets/app.dart
```diff
@@ -308,16 +308,17 @@ class ChooseAccountPage extends StatelessWidget {
...
                   title: zulipLocalizations.logOutConfirmationDialogTitle,
                   message: zulipLocalizations.logOutConfirmationDialogMessage,
                   // TODO(#1032) "destructive" style for action button
-                  actionButtonText: zulipLocalizations.logOutConfirmationDialogConfirmButton,
-                  onActionButtonPress: () {
-                    // TODO error handling if db write fails?
-                    logOutAccount(GlobalStoreWidget.of(context), accountId);
-                  });
+                  actionButtonText: zulipLocalizations.logOutConfirmationDialogConfirmButton);
+                if (await dialog.closed == true) {
```
chrisbobbe, line 317: Oh, hmm, I guess this doesn't look ideal. What it means is "did the user press the 'Log out' button", but what it _looks_ like it means is "was the dialog closed?".

gnprice: Yeah. Maybe rename the field, say to `result`?

lib/widgets/dialog.dart
```diff
@@ -19,13 +19,29 @@ Widget _dialogActionText(String text) {
...
+class DialogStatus<T> {
   const DialogStatus(this.closed);
 
   /// Resolves when the dialog is closed.
-  final Future<void> closed;
+  ///
+  /// If this completes with null, the dialog was dismissed.
+  /// Otherwise, completes with a [T] identifying the interaction's outcome.
+  ///
+  /// See, e.g., [showSuggestedActionDialog] and [SuggestedActionDialogResult].
```
gnprice, line 43: nit: the latter reference no longer exists

lib/widgets/dialog.dart
```diff
@@ -61,28 +77,31 @@ DialogStatus showErrorDialog({
...
-  showDialog<void>(
+  final future = showDialog<bool>(
     context: context,
     builder: (BuildContext context) => AlertDialog(
       title: Text(title),
       content: SingleChildScrollView(child: Text(message)),
       actions: [
         TextButton(
-          onPressed: () => Navigator.pop(context),
+          onPressed: () => Navigator.pop<bool?>(context, null),
```
gnprice, line 100: nit: the types allow `bool` here (and below):
```suggestion
          onPressed: () => Navigator.pop<bool>(context, null),
```

and that seems a bit more consistent with `bool` as the type parameter on `showDialog`

gnprice (commented):
Thanks for the revision! LGTM modulo the subthread above, and two nits below.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

PR #1472, opened 2025-04-10
local echo (5/n): Create outbox messages on send

PIG208:
This is stacked atop #1463, toward #1441.

No UI/user-facing change in this PR.

### DIFF
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -530,12 +530,14 @@ const kSendMessageOfferRestoreWaitPeriod = Duration(seconds: 10);  // TODO(#1441
 /// [MessageStore.sendMessage] call and before its eventual deletion.
 ///
 /// ```
-///              Got an [ApiRequestException].
-///          ┌──────┬──────────┬─────────────► failed
-/// (create) │      │          │                 │
-///    └► hidden   waiting   waitPeriodExpired ──┴──────────────► (delete)
-///          │      ▲   │      ▲                   User restores
-///          └──────┘   └──────┘                   the draft.
+///                             Got an [ApiRequestException].
+///          ┌──────┬────────────────────────────┬──────────► failed
+///          │      │                            │              │
+///          │      │      [sendMessage]         │              │
+/// (create) │      │      request succeeds.     │              │
+///    └► hidden   waiting ◄─────────────── waitPeriodExpired ──┴─────► (delete)
+///          │      ▲   │                     ▲               User restores
+///          └──────┘   └─────────────────────┘               the draft.
 ///         Debounce     [sendMessage] request
 ///         timed out.   not finished when
 ///                      wait period timed out.
@@ -559,7 +561,8 @@ enum OutboxMessageState {
   /// outbox message is shown to the user.
   ///
   /// This state can be reached after staying in [hidden] for
-  /// [kLocalEchoDebounceDuration].
+  /// [kLocalEchoDebounceDuration], or when the request succeeds after the
+  /// outbox message reaches [OutboxMessageState.waitPeriodExpired].
   waiting,
 
   /// The [sendMessage] HTTP request did not finish in time and the user is
@@ -717,7 +720,8 @@ mixin _OutboxMessageStore on PerAccountStoreBase {
     final isStateTransitionValid = switch (newState) {
       OutboxMessageState.hidden => false,
       OutboxMessageState.waiting =>
-        oldState == OutboxMessageState.hidden,
+        oldState == OutboxMessageState.hidden
+        || oldState == OutboxMessageState.waitPeriodExpired,
       OutboxMessageState.waitPeriodExpired =>
         oldState == OutboxMessageState.waiting,
       OutboxMessageState.failed =>
@@ -803,6 +807,14 @@ mixin _OutboxMessageStore on PerAccountStoreBase {
     // Cancel the timer that would have had us start presuming that the
     // send might have failed.
     _outboxMessageWaitPeriodTimers.remove(localMessageId)?.cancel();
+    if (_outboxMessages[localMessageId]!.state
+          == OutboxMessageState.waitPeriodExpired) {
+      // The user was offered to restore the message since the request did not
+      // complete for a while.  Since the request was successful, we expect the
+      // message event to arrive eventually.  Stop inviting the the user to
+      // retry, to avoid double-sends.
+      _updateOutboxMessage(localMessageId, newState: OutboxMessageState.waiting);
+    }
   }
 
   TopicName _processTopicLikeServer(TopicName topic, {
--- a/test/model/message_test.dart
+++ b/test/model/message_test.dart
@@ -217,6 +217,33 @@ void main() {
       await check(outboxMessageFailFuture).throws();
     }));
 
+    test('waiting -> waitPeriodExpired -> waiting and never return to waitPeriodExpired', () => awaitFakeAsync((async) async {
+      await prepare(stream: stream);
+      await prepareMessages([eg.streamMessage(stream: stream)]);
+      // Set up a [sendMessage] request that succeeds after enough delay,
+      // for the outbox message to reach the waitPeriodExpired state.
+      // TODO extract helper to add prepare an outbox message with a delayed
+      //   successful [sendMessage] request if we have more tests like this
+      connection.prepare(json: SendMessageResult(id: 1).toJson(),
+        delay: kSendMessageOfferRestoreWaitPeriod + Duration(seconds: 1));
+      final future = store.sendMessage(
+        destination: streamDestination, content: 'content');
+      async.elapse(kSendMessageOfferRestoreWaitPeriod);
+      checkState().equals(OutboxMessageState.waitPeriodExpired);
+
+      // Wait till the [sendMessage] request succeeds.
+      await future;
+      checkState().equals(OutboxMessageState.waiting);
+
+      // Wait till we reach at least [kSendMessageOfferRestoreWaitPeriod] after
+      // returning to the waiting state.
+      async.elapse(kSendMessageOfferRestoreWaitPeriod);
+      async.flushTimers();
+      // The outbox message should stay in the waiting state;
+      // it should not transition to waitPeriodExpired.
+      checkState().equals(OutboxMessageState.waiting);
+    }));
+
     group('… -> failed', () {
       test('hidden -> failed', () => awaitFakeAsync((async) async {
         await prepareOutboxMessageToFailAfterDelay(Duration.zero);


chrisbobbe (commented):
Thanks! Comments below.

lib/model/message.dart
```diff
@@ -84,17 +268,124 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
...
+  int _nextLocalMessageId;
+
   @override
-  Future<void> sendMessage({required MessageDestination destination, required String content}) {
-    // TODO implement outbox; see design at
-    //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/.23M3881.20Sending.20outbox.20messages.20is.20fraught.20with.20issues/near/1405739
-    return _apiSendMessage(connection,
-      destination: destination,
+  Future<void> sendMessage({required MessageDestination destination, required String content}) async {
+    if (!debugOutboxEnable) {
```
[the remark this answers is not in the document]
PIG208: Hmm, I think we still need to send the message, just without `queueId` and `localId`. Could you elaborate on how you would structure this?

lib/model/message.dart
```diff
@@ -8,12 +12,161 @@ import 'message_list.dart';
...
+        localMessageId: localMessageId,
+        selfUserId: selfUserId,
+        conversation: StreamConversation(
+          destination.streamId,
+          destination.topic.interpretAsServer(
+            zulipFeatureLevel: zulipFeatureLevel,
+            realmEmptyTopicDisplayName: realmEmptyTopicDisplayName),
+          displayRecipient: null),
+        content: content,
+      ),
```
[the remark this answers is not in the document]
PIG208: Yeah, good catch.

If we are to handle the updates, the initial setup probably wouldn't change. We could then have something along the lines of:

```dart
  void handleUpdateRealmSettingsEvent(UpdateRealmSettingsEvent event) {
    final affectedLocalMessageIds = <int>[];
    for (final outboxMessage in outboxMessages.values) {
      if (outboxMessage is! StreamOutboxMessage) continue;
      outboxMessage.conversation = outboxMessage.destination.topic.interpretAsServer(
        zulipFeatureLevel: zulipFeatureLevel,
        realmEmptyTopicDisplayName: realmEmptyTopicDisplayName);
      affectedLocalMessageIds.add(outboxMessage.localMessageId);
    }
    for (final view in _messageListViews) {
      view.handleOutboxMessageUpdate(affectedLocalMessageIds);
    }
  }
```

Just making `conversation` lazily evaluated from the current value of `PerAccountStore` wouldn't by itself be sufficient, since we still need to notify the message list views whenever such a change happens.

Right now, since we don't handle these events, I'm inclined to leave it to the issues you linked to.

test/model/message_test.dart
```diff
@@ -75,6 +93,365 @@ void main() {
...
+
+      // Handle the event after the message is sent but before the debounce
+      // timeout.  The outbox message should remain hidden since the send
+      // request was sucessful.
+      await store.handleEvent(eg.messageEvent(message,
+        localMessageId: outboxMessage.localMessageId));
+      check(store.outboxMessages).isEmpty();
+      check(outboxMessage)
+        ..state.equals(OutboxMessageLifecycle.sent)
+        ..hidden.isTrue();
```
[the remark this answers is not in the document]
PIG208: The second sentence of the comment before `store.handleEvent` is meant to address this check:

```
      // Handle the event after the message is sent but before the debounce
      // timeout.  The outbox message should remain hidden since the send
      // request was successful.
```

Perhaps it will be clearer to move this right before the check.

lib/model/message.dart
```diff
@@ -8,12 +12,161 @@ import 'message_list.dart';
...
+///                              │                  Event received,    │
+///              Send            │                  or we abandoned    │
+///           immediately.       │      200.        the queue.         ▼
+/// (create) ──────────────► sending ────────► sent ──────────────► (delete)
+///                              │4xx,          │                      ▲
+///                              │other error,  │Reached       User    │
+///                              │or reached    │time limit.   cancels.│
+///                              │time limit.   ▼                      │
+///                              └───────────► failed ─────────────────┘
+/// ```
```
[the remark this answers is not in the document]
PIG208: For the diagram update, started a discussion here: [#mobile > #F1441 Handle retry state machine @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/.23F1441.20Handle.20retry.20state.20machine/near/2156104)

PIG208:
Updated the PR! There are some pending questions mainly with regards to the state diagram and the send time limit. I introduced the new `waitPeriodExpired` state in this revision, with some changes to the states.

PIG208:
Will be working on a new revision to reorganize some of the implementation code with further state machine changes.

lib/model/message.dart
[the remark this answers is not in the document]
gnprice: I think having a comment noting this can happen is an adequate resolution. The effect is basically that the predicted ultimate topic can change between "(no topic)" and "general chat" and a different name for "general chat", right?

It should be pretty uncommon for that setting to change between when an attempted send happens and when the outbox-message goes away. There's also an unavoidable race that has the same effect: an admin could change the name of "general chat" (i.e. the value of `realmEmptyTopicDisplayName`) concurrently with the user making the send request, so that the setting in effect by the time the request arrives is different from the setting the client last heard about. So I think the comment doesn't need to be a TODO.

PIG208:
The PR has been updated to implement the state diagram discussed in chat: [#mobile > #F1441 Handle retry state machine @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/.23F1441.20Handle.20retry.20state.20machine/near/2157168)

PIG208:
(pushed another update to make timestamp used in sendMessage testable)

lib/model/message.dart
```diff
@@ -1,19 +1,310 @@
...
+      assert(outboxMessages.containsKey(localMessageId));
+      _outboxMessageDebounceTimers.remove(localMessageId);
+      _updateOutboxMessage(localMessageId, newState: OutboxMessageState.waiting);
+    });
+
+    _outboxMessageWaitPeriodTimers[localMessageId] = Timer(kSendMessageRetryWaitPeriod, () {
+      assert(outboxMessages.containsKey(localMessageId));
+      _outboxMessageWaitPeriodTimers.remove(localMessageId);
+      _updateOutboxMessage(localMessageId, newState: OutboxMessageState.waitPeriodExpired);
+    });
```
chrisbobbe, line 243: The `assert`s don't look right to me; wouldn't they throw if an outbox message was removed before the timer finished?, which can happen in this part of the state diagram:

```dart
///              Event received.
///              Or we abandoned the queue.
/// (any state) ────────────────────────────► (delete)
```

lib/model/message.dart
```diff
@@ -1,19 +1,310 @@
...
+      await _apiSendMessage(connection,
+        destination: destination,
+        content: content,
+        readBySender: true,
+        queueId: queueId,
+        localId: localMessageId.toString());
+    } catch (e) {
+      // `localMessageId` is not necessarily in the store. This is because
+      // message event can still arrive before the send request fails to
+      // networking issues.
```
chrisbobbe, line 255: nit: maybe "fails with networking issues" or "fails because of networking issues"

lib/api/model/model.dart
```diff
@@ -586,6 +595,30 @@ extension type const TopicName(String _value) {
...
+      assert(_value.isNotEmpty);
+      return this;
+    }
+    if (_value == kNoTopicTopic || _value.isEmpty) {
+      // TODO(#1250): this assumes that the 'support_empty_topics'
+      //   client_capability is false; update this when we set it to true
+      return TopicName(realmEmptyTopicDisplayName!);
+    }
+    return TopicName(_value);
+  }
```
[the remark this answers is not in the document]
PIG208: I think the method itself will stay, since the server logic for converting literal `realmEmptyTopicDisplayName` into an empty string still applies when the client supports empty topics. I agree that we should have a `// TODO(server-10)` comment, for the feature level check.

Yeah, making sure that this is aware of whitespace-trimming is a good point.

lib/model/message.dart
```diff
@@ -1,19 +1,310 @@
...
+            // There's also an unavoidable race that has the same effect:
+            // an admin could change the name of "general chat"
+            // (i.e. the value of realmEmptyTopicDisplayName) concurrently with
+            // the user making the send request, so that the setting in effect
+            // by the time the request arrives is different from the setting the
+            // client last heard about.  The realm update events do not have
+            // information about this race for us to update the prediction
+            // correctly.
+            zulipFeatureLevel: zulipFeatureLevel,
+            realmEmptyTopicDisplayName: realmEmptyTopicDisplayName),
```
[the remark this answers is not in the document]
PIG208: The first paragraph is much clearer. Thanks!

(I should remind myself of updating this once #1250 changes how `interpretAsServer` behaves in terms of the topic predicted.)

lib/model/message.dart
PIG208: `(delete)` doesn't have outgoing edges triggered by timeouts. So when the outbox message, for example, transition from `waiting` to `(delete)`, the timers should have been canceled too.

lib/model/message.dart
```diff
@@ -1,19 +1,310 @@
...
+  /// Update the state of the [OutboxMessage] with the given [localMessageId],
+  /// and notify listeners if necessary.
+  ///
+  /// This is a no-op if the outbox message does not exist, or that
+  /// [OutboxMessage.state] already equals [newState].
+  void _updateOutboxMessage(int localMessageId, {
+    required OutboxMessageState newState,
+  }) {
+    final outboxMessage = outboxMessages[localMessageId];
+    if (outboxMessage == null || outboxMessage.state == newState) {
```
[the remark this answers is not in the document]
PIG208: > Is there a legitimate need for callers to pass a new state that's the same as the old state? 

I think not. Removing this so that such an invalid state change can be caught with `set state` assertions.

chrisbobbe:
This is next on my queue; sorry I've let it sit for a bit!

lib/api/model/model.dart
```diff
@@ -604,6 +613,33 @@ extension type const TopicName(String _value) {
...
+  /// For a client that does not support empty topics,
+  /// a modern server (FL>=334) would convert "(no topic)" and empty topics to
+  /// `store.realmEmptyTopicDisplayName`.
+  ///
+  /// See also: https://zulip.com/api/send-message#parameter-topic
+  TopicName interpretAsServer({
+    required int zulipFeatureLevel,
+    required String? realmEmptyTopicDisplayName,
+  }) {
+    assert(_value.trim() == _value);
```
chrisbobbe, line 628: Good to have the assert, but let's mention the restriction on when this can be called. I guess the public-interface way of saying that is: it should only be called if the instance was constructed from a string without leading/trailing whitespace, right? (or "from a trimmed string" or similar)

chrisbobbe (commented):
Actually, could you send the first four commits as a new PR (just a small comment on those, below) and mark it for Greg's review?:

a19064924 api [nfc]: Add TopicName.interpretAsServer
bf8063a9c store [nfc]: Move zulip{FeatureLevel,Version} to PerAccountStoreBase
1f126181c test [nfc]: Generate timestamps
f8e1dab5e binding [nfc]: Add utcNow

That should help move things along faster as I'll need more time for the main commit:

cc0d77328 message: Create an outbox message on send; manage its states

chrisbobbe (commented):
Thanks, this is an exciting feature!!

Some general comments on the tests:

What's the reason for the `late OutboxMessage outboxMessage;`, which we check even after the outbox message is removed from the store? If we removed that variable and all the test code that depends on it, would we risk failing to catch a user-visible bug? I guess it's better if the `OutboxMessage` doesn't get changed after it's removed from the store…but it doesn't seem like it would cause any user-visible bug if it did, and it also seems pretty unlikely. (The app code shouldn't need to keep those objects around to compare them by identity, because they have unique IDs.)

Do the tests cover all possible state transitions and their causes? I think a helpful organization would be one test for each state transition, or more than one if the transition has multiple causes:

- hidden -> failed
- waiting -> failed
- waitPeriodExpired -> failed
- hidden -> waiting
- waiting -> waitPeriodExpired
- failed -> (delete) because draft was restored
- waitPeriodExpired -> (delete) because draft was restored
- hidden -> (delete) because event received
- waiting -> (delete) because event received
- failed -> (delete) because event received
- waitPeriodExpired -> (delete) because event received

I think this would make a really helpful starting point for when we work on the complete outbox system in the future, which will require adding and removing some transitions in the state diagram. It'll be easier to see how to update the tests if they're labeled with the current set of transitions.

lib/model/message.dart
```diff
@@ -73,15 +366,18 @@ class _EditMessageRequestStatus {
...
 
-class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
-  MessageStoreImpl({required super.core})
+class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMessageStore {
+  MessageStoreImpl({required super.core, required this.realmEmptyTopicDisplayName})
     // There are no messages in InitialSnapshot, so we don't have
     // a use case for initializing MessageStore with nonempty [messages].
     : messages = {};
 
+  final String? realmEmptyTopicDisplayName;
```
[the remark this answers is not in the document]
gnprice: (In particular the `TODO(#668)` is the thing that will help us be sure to spot this and bundle it up into that new realm-settings-etc substore when we're about to start updating realm settings, #668, so that it gets the benefit of those updates.)

lib/model/message.dart
```diff
@@ -1,21 +1,309 @@
...
+        localId: localMessageId.toString());
+    } catch (e) {
+      // `localMessageId` is not necessarily in the store. This is because the
+      // message event can still arrive, before the send request fails with
+      // networking issues.
+      _outboxMessageDebounceTimers.remove(localMessageId)?.cancel();
+      _outboxMessageWaitPeriodTimers.remove(localMessageId)?.cancel();
+      _updateOutboxMessage(localMessageId, newState: OutboxMessageState.failed);
+      rethrow;
+    }
```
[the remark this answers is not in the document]
PIG208: Thanks. This is Neat!

Looking at `UpdateMachine`, I think this check will be quite granular.

I suppose a good structure will be adding this flag prior to the main commit, making some or all (?) `editMessage`, `handle…Event`, and friends `assert(!_disposed)` at the beginning, and return conditionally (`if (_disposed) return`) after async-gaps.

And then, I guess for private helpers like `_notifyMessageListViewsForOneMessage`, `assert(!_disposed)` at the start is not necessary.

lib/model/message.dart
```diff
@@ -1,21 +1,309 @@
...
+    if (event.localMessageId != null) {
+      final localMessageId = int.parse(event.localMessageId!, radix: 10);
+      // The outbox message can be missing if the user removes it (to be
+      // implemented in #1441) before the event arrives.
+      // Nothing to do in that case.
+      _outboxMessages.remove(localMessageId);
+      _outboxMessageDebounceTimers.remove(localMessageId)?.cancel();
+      _outboxMessageWaitPeriodTimers.remove(localMessageId)?.cancel();
+    }
+  }
```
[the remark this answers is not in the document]
PIG208: Right, that will cause the message-list views to be noticed twice if there is an outbox message — once from  `removeOutboxMessage`, and another time from `view.handleMessageEvent`. That's perhaps fine.

lib/model/message.dart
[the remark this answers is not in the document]
chrisbobbe: That sounds right to me.

lib/model/message.dart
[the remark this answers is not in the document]
chrisbobbe: Oh I see; yeah, I forgot that `MessageStoreImpl.handleMessageEvent` already calls `view.handleMessageEvent` for all the `_messageListViews`.

I guess `removeOutboxMessage` isn't meant as an internal helper function, as I'd thought, right? Its purpose is just to handle these two state transitions, by responding to a user interaction:

- `failed` -> `(delete)`
- `waitPeriodExpired` -> `(delete)`

How about renaming `removeOutboxMessage` around that public-facing role? It's quite similar to `MessageStore.takeFailedEdit`, for edit-message, right:

```dart
  /// Forgets the failed edit request and returns the attempted new content.
  ///
  /// Should only be called when there is a failed request,
  /// per [getEditMessageErrorStatus].
  String takeFailedMessageEdit(int messageId);
```

except that it can also be called from a "wait-period expired" state, not only a "failed" state. Maybe `takeStaleOutboxMessage` or maybe just `takeOutboxMessage`? And have its dartdoc say when it's allowed to be called, i.e., just from the `failed` or `waitPeriodExpired` state.

lib/model/message.dart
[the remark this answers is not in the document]
PIG208: An interesting case is `unregisterMessageList`. Ultimately I think we want to have assert(!_disposed)` there, i.e. the message-list view should be unregistered before the store is disposed. Leaving a TODO similar to the one we have in `dispose`.

PIG208:
Thanks for the reviews! I have reorganized the tests and addressed the comments, after rebasing this on top of edit message changes :tada:

> What's the reason for the late OutboxMessage outboxMessage;, which we check even after the outbox message is removed from the store? If we removed that variable and all the test code that depends on it, would we risk failing to catch a user-visible bug? I guess it's better if the OutboxMessage doesn't get changed after it's removed from the store…but it doesn't seem like it would cause any user-visible bug if it did, and it also seems pretty unlikely. (The app code shouldn't need to keep those objects around to compare them by identity, because they have unique IDs.)

I agree that checking the state after the outbox message is removed from the store does not seem necessary. Removed those checks, and kept this late variable is mostly a helper for us to check the state of the outbox message.

> Do the tests cover all possible state transitions and their causes? I think a helpful organization would be one test for each state transition, or more than one if the transition has multiple causes

Yeah. It is not straightforward to confirm that these sets of tests cover all the state transitions. Your proposed way of organizing the tests does help a lot! I hope that this revision makes tests look shorter and more concise.

test/model/message_test.dart
```diff
@@ -76,6 +94,303 @@ void main() {
...
+    }) async {
+      message = eg.streamMessage(stream: stream);
+      await prepare(stream: stream, zulipFeatureLevel: zulipFeatureLevel);
+      await prepareMessages([eg.streamMessage(stream: stream)]);
+      connection.prepare(json: SendMessageResult(id: 1).toJson());
+      await store.sendMessage(
+        destination: destination ?? streamDestination, content: 'content');
+      outboxMessage = store.outboxMessages.values.single;
+      check(outboxMessage).state.equals(OutboxMessageState.hidden);
+    }
```
chrisbobbe, line 149: This differs from `prepareSendMessageToFail` in that it awaits the send-message request. Is that intentional? If so, would it be helpful to rename the function or add some dartdoc about that?

test/model/message_test.dart
```diff
@@ -76,6 +94,303 @@ void main() {
...
+        check(outboxMessage).state.equals(OutboxMessageState.failed);
+
+        // Wait till we reach at least [kSendMessageOfferRestoreWaitPeriod] after
+        // the send request was initiated.
+        async.elapse(kSendMessageOfferRestoreWaitPeriod);
+        async.flushTimers();
+        // The outbox message should stay in the failed state;
+        // it should not transition to waiting or waitPeriodExpired.
+        check(outboxMessage).state.equals(OutboxMessageState.failed);
+      }));
```
chrisbobbe, line 210: I see your reply at https://github.com/zulip/zulip-flutter/pull/1472#issuecomment-2892391723 :

> I agree that checking the state after the outbox message is removed from the store does not seem necessary. Removed those checks, and kept this late variable is mostly a helper for us to check the state of the outbox message.

It's actually unhelpful to use the `late outboxMessage` variable here, isn't it (?), because if there's a bug where the outbox message is removed from the store when it shouldn't be, this test won't catch it.

If all the uses of `outboxMessage` were replaced with lookups in the store (via a helper function if needed), I think that would be better, because we'd be testing the interface of this subsystem more directly.

chrisbobbe (commented):
Thanks, I think this is getting close! I like the new organization of the tests! Small comments below.

test/model/message_test.dart
PIG208: > because if there's a bug where the outbox message is removed from the store when it shouldn't be, this test won't catch it

Oh right. That makes sense. Yeah, that will make it more end-to-end.

test/model/message_test.dart
PIG208: Right, because this helper never needs to get a delay, unlike `prepareSendMessageToFail`. Perhaps we can name it to mention its immediate nature.

test/model/message_test.dart
```diff
@@ -76,6 +94,302 @@ void main() {
...
+      for (int i = 0; i < 10; i++) {
+        connection.prepare(json: SendMessageResult(id: 1).toJson());
+        await store.sendMessage(destination: streamDestination, content: 'content');
+      }
+      // [store.outboxMessages] has the same number of keys (localMessageId)
+      // as the number of sent messages, which are guaranteed to be distinct.
+      check(store.outboxMessages).keys.length.equals(10);
+    });
+
+    void checkState(OutboxMessageState state) {
```
chrisbobbe, line 135: nit: how about `expected` or `expectedState` for the param's name, so the caller knows to pass a hard-coded value instead of some actual output from the code under test

chrisbobbe (commented):
Thanks, LGTM! One nit below, and I'll mark for Greg's review.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+
+  @override
+  final DmConversation conversation;
+}
+
+/// Manages the outbox messages portion of [MessageStore].
+mixin _OutboxMessageStore on PerAccountStoreBase {
+  late final UnmodifiableMapView<int, OutboxMessage> outboxMessages =
+    UnmodifiableMapView(_outboxMessages);
+  final Map<int, OutboxMessage> _outboxMessages = {};
```
gnprice, line 127: Neat! We should perhaps use this UnmodifiableMapView wrapper more.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+    final isStateTransitionValid = newState != oldState && switch (newState) {
+      OutboxMessageState.hidden => false,
+      OutboxMessageState.waiting =>
+        oldState == OutboxMessageState.hidden,
+      OutboxMessageState.waitPeriodExpired =>
+        oldState == OutboxMessageState.waiting,
+      OutboxMessageState.failed =>
+        oldState == OutboxMessageState.hidden
+        || oldState == OutboxMessageState.waiting
+        || oldState == OutboxMessageState.waitPeriodExpired,
```
gnprice, line 178: nit: equivalently,
```suggestion
      OutboxMessageState.failed => true,
```

right? Given the `newState != oldState` above.

Alternatively, I think with this condition in place here, that inequality check is redundant.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+  /// [OutboxMessageState.waitPeriodExpired] after a delay,
+  /// indexed by [OutboxMessage.localMessageId].
+  ///
+  /// If the send message request fails within the time limit,
+  /// the outbox message's timer gets removed and cancelled.
+  final Map<int, Timer> _outboxMessageWaitPeriodTimers = {};
+
+  /// A fresh ID to use for [OutboxMessage.localMessageId],
+  /// unique within this instance.
+  int _nextLocalMessageId = 0;
```
gnprice, line 146: Let's start at 1 rather than 0. That matches the practice in all the server's autoincrementing database indexes.

lib/model/message.dart
```diff
@@ -137,21 +478,29 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
...
+    assert(!_disposed);
+    if (!debugOutboxEnable) {
+      return _apiSendMessage(connection,
+        destination: destination,
+        content: content,
+        readBySender: true);
+    }
+    return _outboxSendMessage(
+      destination: destination, content: content,
+      realmEmptyTopicDisplayName: realmEmptyTopicDisplayName);
```
gnprice, line 498: It looks like this doesn't conform to the constraint written on the `realmEmptyTopicDisplayName` getter, and enforced with an assert there. This will eagerly call that getter, before knowing whether the ZFL means its value will be needed.

I think the refactor that would be needed to make that assertion fully hold for this path is probably too much to be worth doing here directly. Instead, let's just bypass it (e.g., look at `_realmEmptyTopicDisplayName` directly) and leave a TODO comment.

I think probably the right solution is ultimately to have `TopicName.processLikeServer` (which is the method that consults this value and causes the need to pass it down here) to instead be a method on the per-account store, and more specifically on the RealmSettingStore (or similar) that we plan to set up in the future. Then it can consult that setting, as well as `zulipFeatureLevel`, directly off of that substore, and call the `realmEmptyTopicDisplayName` for itself only after it determines it's needed. But that substore is a post-launch refactor.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+      StreamDestination(:final streamId, :final topic) => StreamOutboxMessage(
+        localMessageId: localMessageId,
+        selfUserId: selfUserId,
+        timestamp: now,
+        conversation: StreamConversation(
+          streamId,
+          topic.processLikeServer(
+            // Processing this just once on creating the outbox message
+            // allows an uncommon bug, because either of these values can change.
+            // During the outbox message's life, a topic processed from
```
gnprice, line 215: Let's move this `topic.processLikeServer` call out to a little private helper method. That will keep the flow of the rest of this method's logic a bit easier to read.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+            // an admin could change the name of "general chat"
+            // (i.e. the value of realmEmptyTopicDisplayName)
+            // concurrently with the user making the send request,
+            // so that the setting in effect by the time the request arrives
+            // is different from the setting the client last heard about.
+            zulipFeatureLevel: zulipFeatureLevel,
+            realmEmptyTopicDisplayName: realmEmptyTopicDisplayName),
+          displayRecipient: null),
+        content: content),
+      DmDestination(:final userIds) => DmOutboxMessage(
```
gnprice, line 230: How about giving `OutboxMessage` a factory constructor that takes an arbitrary `Conversation`, and dispatches between these two subclass constructors? That would allow this `switch` to get pushed down into computing just `conversation`, and the rest of these two constructor calls to get deduplicated. I think that'd make the structure of what this is doing a bit clearer.

The existing `OutboxMessage` constructor could get renamed to `OutboxMessage._` to make room for the nice factory constructor to have the plain `OutboxMessage` name.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+        content: content),
+    };
+
+    _outboxMessageDebounceTimers[localMessageId] = Timer(kLocalEchoDebounceDuration, () {
+      assert(!_disposed);
+      assert(outboxMessages.containsKey(localMessageId),
+        'The timer should have been canceled when the outbox message was removed.');
+      _outboxMessageDebounceTimers.remove(localMessageId);
+      _updateOutboxMessage(localMessageId, newState: OutboxMessageState.waiting);
+    });
```
gnprice, line 244: Let's pull this callback out to its own method:
```suggestion
    _outboxMessageDebounceTimers[localMessageId] = Timer(
      kLocalEchoDebounceDuration, () => _handleOutboxDebounce(localMessageId));
```

That will help shorten this method and make its overall structure easier to see; and it'll also make clear there's a narrow interface between this overall method and the callback, with just the one bit of data `localMessageId` passing between them (vs. all the other parameters and locals that this method otherwise would offer to the callback's closure).

gnprice (commented):
Thanks @PIG208 for building this, and @chrisbobbe for the previous reviews! Exciting to see this approaching ready.

Generally this logic all looks good; various comments below. I haven't yet read the tests.

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+  ///
+  /// This state can be reached after staying in [hidden] for
+  /// [kLocalEchoDebounceDuration].
+  waiting,
+
+  /// The message was assumed not delivered after some time it was sent.
+  ///
+  /// This state can be reached when the message event hasn't arrived in
+  /// [kSendMessageOfferRestoreWaitPeriod] since the outbox message's creation.
+  waitPeriodExpired,
```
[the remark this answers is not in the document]
PIG208: Yeah, this proposed behavior sounds better to me.

I think an easy implementation will just require us to cancel the timer that leads to `waitPeriodExpired` when the request is successful. At first I thought we might want to re-introduce that `sending` state from a previous draft, but I believe that shouldn't be necessary.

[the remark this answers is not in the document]
gnprice: Sounds good!

PIG208:

lib/model/message.dart
```diff
@@ -1,21 +1,332 @@
...
+    required this.conversation,
+    required super.content,
+  }) : assert(conversation.allRecipientIds.contains(selfUserId));
+
+  @override
+  final DmConversation conversation;
+}
+
+/// Manages the outbox messages portion of [MessageStore].
+mixin _OutboxMessageStore on PerAccountStoreBase {
```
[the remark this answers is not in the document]
gnprice: bump — after MessageStoreImpl, not only MessageStore, and including all the new code so the MessageStore interface remains at the top :slightly_smiling_face:

(That way when the reader is thinking about aspects other than the outbox, for example the maintenance of the `messages` map itself, they can quickly go straight to those without skimming past the outbox logic; and conversely when they want to study how the outbox works, that's still all together in one place, namely the end of the file.)

lib/model/message.dart
```diff
@@ -137,21 +539,30 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
...
+    if (!debugOutboxEnable) {
+      return _apiSendMessage(connection,
+        destination: destination,
+        content: content,
+        readBySender: true);
+    }
+    return _outboxSendMessage(
+      destination: destination, content: content,
+      // TODO move [TopicName.processLikeServer] to a substore, eliminating this
+      realmEmptyTopicDisplayName: _realmEmptyTopicDisplayName);
```
gnprice, line 560: nit: let's add a link to https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2099069276 to expand on the details of this thought:
```suggestion
      // TODO move [TopicName.processLikeServer] to a substore, eliminating this;
      //   see https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2099069276
      realmEmptyTopicDisplayName: _realmEmptyTopicDisplayName);
```

lib/model/message.dart
```diff
@@ -1,21 +1,180 @@
...
+    required this.localMessageId,
+    required int selfUserId,
+    required super.timestamp,
+    required this.contentMarkdown,
+  }) : _state = OutboxMessageState.hidden,
+       super(senderId: selfUserId);
+
+  // TODO(dart): This has to be a plain static method, because factories/constructors
+  //   do not support type parameters: https://github.com/dart-lang/language/issues/647
+  static OutboxMessage fromConversation(Conversation conversation, {
```
gnprice, line 101: Ah right, indeed — so it can't quite have the short `OutboxMessage(…)` name like I suggested at https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2099073411 above.

lib/model/message.dart
```diff
@@ -66,6 +234,219 @@ mixin MessageStore {
...
+      () => _handleOutboxWaitPeriodExpired(localMessageId));
+
+    try {
+      await _apiSendMessage(connection,
+        destination: destination,
+        content: content,
+        readBySender: true,
+        queueId: queueId,
+        localId: localMessageId.toString());
+      _outboxMessageWaitPeriodTimers.remove(localMessageId)?.cancel();
```
gnprice, line 351: Cool, I guess this line is the implementation of https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2099101759 .

gnprice, line 351: This feels a bit mysterious on its own, so let's add a comment explaining why it's here:
```suggestion

      // The send request succeeded, so the message was definitely sent.
      // Cancel the timer that would have had us start presuming that the
      // send might have failed.
      _outboxMessageWaitPeriodTimers.remove(localMessageId)?.cancel();
```

gnprice: and nit: move this outside of the try/catch, to make clear the try/catch is focused on exceptions from `_apiSendMessage` (where exceptions are expected) and not potential exceptions from this line (which shouldn't be possible). Discussion here:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#catch-specific

gnprice: Writing out that comment also caused me to have the thought: what happens if the request succeeds and the state has *already* moved to `waitPeriodExpired`? The expiry means we've already started presuming that it probably failed, but suddenly we know for sure it actually succeeded.

If there isn't a quick change to make that would handle that, it's fine to leave a TODO comment for it instead.

gnprice (commented):
Thanks for the revision! Comments below. I'll also go take a quick initial look at the tests.

test/model/message_test.dart
```diff
@@ -76,6 +94,314 @@ void main() {
...
+      await prepareMessages([eg.streamMessage(stream: stream)]);
+      connection.prepare(apiException: eg.apiBadRequest(), delay: delay);
+      outboxMessageFailFuture = store.sendMessage(
+        destination: streamDestination, content: 'content');
+    }
+
+    test('smoke DM: hidden -> waiting -> (delete)', () => awaitFakeAsync((async) async {
+      await prepareOutboxMessage(destination: DmDestination(
+        userIds: [eg.selfUser.userId, eg.otherUser.userId]));
+      checkState(expectedState: OutboxMessageState.hidden);
```
gnprice, line 166: nit: can put these in more canonical `checks` style:
```suggestion
      checkState().equals(OutboxMessageState.hidden);
```

test/model/message_test.dart
```diff
@@ -76,6 +94,314 @@ void main() {
...
+        localMessageId: store.outboxMessages.keys.single));
+      check(store.outboxMessages).isEmpty();
+    }));
+
+    test('smoke stream message: hidden -> waiting -> (delete)', () => awaitFakeAsync((async) async {
+      await prepareOutboxMessage();
+      checkState(expectedState: OutboxMessageState.hidden);
+
+      async.elapse(kLocalEchoDebounceDuration);
+      checkState(expectedState: OutboxMessageState.waiting);
```
gnprice, line 182: I like this structure where the helpers center on a single outbox-message. It helps keep the tests focused, since the bulk of them are all about a single outbox-message (because there's very little between-message interaction to be tested).

gnprice (commented):
Cool, and generally the tests look good.

I've read the miscellaneous test changes outside the main block in `message_test.dart`, and skimmed over the latter; just a couple of comments below. I'll leave for another day reading through that main block of tests in full.

lib/model/message.dart
PIG208: > and nit: move this outside of the try/catch, to make clear the try/catch is focused on exceptions from `_apiSendMessage` (where exceptions are expected) and not potential exceptions from this line (which shouldn't be possible). Discussion here: https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#catch-specific

Ah. This reminds me of adding a `if (_disposed)` check.

PIG208: > Writing out that comment also caused me to have the thought: what happens if the request succeeds and the state has _already_ moved to `waitPeriodExpired`? The expiry means we've already started presuming that it probably failed, but suddenly we know for sure it actually succeeded.
> 
> If there isn't a quick change to make that would handle that, it's fine to leave a TODO comment for it instead.


I think this would be a matter of implementing a new state transition:

 ```                                                                         
                             Got an [ApiRequestException].                   
          ┌──────┬────────────────────────────┬──────────► failed            
          │      │                            │              │               
          │      │      [sendMessage]         │              │               
 (create) │      │      request succeeds.     │              │               
    └► hidden   waiting ◄─────────────── waitPeriodExpired ──┴─────► (delete)
          │      ▲   │                     ▲               User restores     
          └──────┘   └─────────────────────┘               the draft.        
         Debounce     [sendMessage] request                                  
         timed out.   not finished when                                      
                      wait period timed out.                                 
                                                                             
              Event received.                                                
 (any state) ─────────────────► (delete)                                     
 ```                                                                         

I guess we can't really convey that `waitPeriodExpired` cannot be reached again without having a variant of `waiting`, but we can imply that from the dart doc of `waitPeriodExpired`:

```dart
  /// The [sendMessage] HTTP request did not finish in time and the user is
  /// invited to retry it.
  ///
  /// This state can be reached when the request has not finished
  /// [kSendMessageOfferRestoreWaitPeriod] since the outbox message's creation.
  waitPeriodExpired,
```

PIG208: Added a new commit `message: Avoid double-sends after send-message request succeeds`  addressing this. I found it to be a good way seeing how we can incrementally add features to this state machine.

test/model/message_test.dart
```diff
@@ -76,6 +94,341 @@ void main() {
...
+      checkState().equals(OutboxMessageState.waiting);
+
+      await store.handleEvent(eg.messageEvent(
+        eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]),
+        localMessageId: store.outboxMessages.keys.single));
+      check(store.outboxMessages).isEmpty();
+    }));
+
+    test('smoke stream message: hidden -> waiting -> (delete)', () => awaitFakeAsync((async) async {
+      await prepareOutboxMessage();
```
gnprice, line 177: nit: since the test description specifically calls for a stream message, the test should explicitly make it a stream message

(the bulk of the other tests below can cheerfully go on using the default that `prepareOutboxMessage` supplies, since they don't care about the specifics of the message)

test/model/message_test.dart
```diff
@@ -76,6 +94,341 @@ void main() {
...
+      await prepareOutboxMessage(destination: DmDestination(
+        userIds: [eg.selfUser.userId, eg.otherUser.userId]));
+      checkState().equals(OutboxMessageState.hidden);
+
+      async.elapse(kLocalEchoDebounceDuration);
+      checkState().equals(OutboxMessageState.waiting);
+
+      await store.handleEvent(eg.messageEvent(
+        eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]),
+        localMessageId: store.outboxMessages.keys.single));
```
gnprice, line 172: nit: could tighten these tests further (going in the same direction noted at https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2099318648) with a helper for this step:
```suggestion
      await receiveMessage(eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]));
```

Could even have that helper default to the shared `message` local, since that's what most of these tests use.

test/model/message_test.dart
```diff
@@ -76,6 +94,341 @@ void main() {
...
+
+        // Handle the message event while the message is being sent.
+        await store.handleEvent(eg.messageEvent(message,
+          localMessageId: store.outboxMessages.keys.single));
+        check(store.outboxMessages).isEmpty();
+
+        // Complete the send request.  There should be no error despite
+        // the send request failure, because the outbox message is not
+        // in the store any more.
+        await check(outboxMessageFailFuture).completes();
```
gnprice, line 297: Cool, this is an important subtle case to test.

(Important because it could be fairly common if the user is on a flaky connection — if the rate of requests randomly failing is anywhere between like 20% and 80%, then a good fraction of the time, the send will make it through, the response won't and so the request will eventually throw when the connection times out, and in the meantime the event will make it through. Important because also it'd be pretty annoying to get a useless error dialog when nothing actually went wrong.)

test/model/message_test.dart
```diff
@@ -76,6 +94,341 @@ void main() {
...
+      await store.sendMessage(
+        destination: destination ?? streamDestination, content: 'content');
+    }
+
+    late Future<void> outboxMessageFailFuture;
+    Future<void> prepareOutboxMessageToFailAfterDelay(Duration delay) async {
+      message = eg.streamMessage(stream: stream);
+      await prepare(stream: stream);
+      await prepareMessages([eg.streamMessage(stream: stream)]);
+      connection.prepare(apiException: eg.apiBadRequest(), delay: delay);
```
gnprice, line 157: nit: for better realism, make it an error that's consistent with the request having actually succeeded on the server side:
```suggestion
      connection.prepare(httpException: SocketException('failed'), delay: delay);
```

That way it's consistent with the test cases where a corresponding MessageEvent arrives even though the request failed. That also may help a future reader in understanding why that scenario can even happen (which is important to understand since we have some logic carefully written to specifically handle it).

test/model/message_test.dart
```diff
@@ -76,6 +94,341 @@ void main() {
...
+        check(store.outboxMessages).isEmpty();
+      }));
+    });
+
+    test('when sending to "(no topic)", process topic like the server does when creating outbox message', () => awaitFakeAsync((async) async {
+      await prepareOutboxMessage(
+        destination: StreamDestination(stream.streamId, TopicName('(no topic)')),
+        zulipFeatureLevel: 370);
+      async.elapse(kLocalEchoDebounceDuration);
+      check(store.outboxMessages.values.single)
```
gnprice, line 392: nit: use checks-getters
```suggestion
      check(store.outboxMessages).values.single
```

lib/model/message.dart
```diff
@@ -435,4 +488,414 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
...
+    if (!_outboxMessages.containsKey(localMessageId)) {
+      // The message event already arrived; nothing to do.
+      return;
+    }
+    // The send request succeeded, so the message was definitely sent.
+    // Cancel the timer that would have had us start presuming that the
+    // send might have failed.
+    _outboxMessageWaitPeriodTimers.remove(localMessageId)?.cancel();
+    if (_outboxMessages[localMessageId]!.state
+          == OutboxMessageState.waitPeriodExpired) {
```
gnprice, line 811: Neat, glad that change turned out to be clean to make! (Following up on https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2101148783.)

The diffstat is:
 lib/model/message.dart       | 28 ++++++++++++++++++++--------
 test/model/message_test.dart | 28 ++++++++++++++++++++++++++++
 2 files changed, 48 insertions(+), 8 deletions(-)

and outside the new test, most of that is comments; I count 6+/1- other than tests and comments.

gnprice (commented):
Cool, and here's a full review of the remainder: the tests, and the new commit at the end. Generally all looks good.

test/model/message_test.dart
```diff
@@ -76,6 +94,341 @@ void main() {
...
+      // the send request was initiated.
+      async.elapse(
+        kSendMessageOfferRestoreWaitPeriod - kLocalEchoDebounceDuration);
+      async.flushTimers();
+      // The outbox message should stay in the waiting state;
+      // it should not transition to waitPeriodExpired.
+      checkState().equals(OutboxMessageState.waiting);
+    }));
+
+    test('waitPeriodExpired -> waiting and never return to waitPeriodExpired', () => awaitFakeAsync((async) async {
```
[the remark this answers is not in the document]
PIG208: There are two cases of "waiting -> waitPeriodExpired -> (delete) …" that cover this. But I think including the first edge will still be helpful because it shows that there is a loop.

test/model/message_test.dart
PIG208: Yeah, exactly. This is for covering the change from Chris here: https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2093703191

test/model/message_test.dart
```diff
@@ -76,6 +94,335 @@ void main() {
...
+    for (int i = 0; i < 10; i++) {
+      connection.prepare(apiException: eg.apiBadRequest());
+      await check(store.sendMessage(
+        destination: StreamDestination(stream.streamId, eg.t('topic')),
+        content: 'content')).throws();
+    }
+
+    final localMessageIds = store.outboxMessages.keys.toList();
+    store.takeOutboxMessage(localMessageIds.removeAt(5));
+    check(store.outboxMessages.keys).deepEquals(localMessageIds);
```
gnprice, line 423: nit: use checks-getter (like https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2101317278 )
```suggestion
    check(store.outboxMessages).keys.deepEquals(localMessageIds);
```

test/model/message_test.dart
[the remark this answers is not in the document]
gnprice: Cool. Yeah, I see those as focused on the later arrow that goes to "(delete)"; partly because they're tucked inside the group "… -> (delete)". When looking for the test that covers this arrow of the state diagram, this seems like the more natural spot to find it.

(Ideally in fact I think there'd be a case here that was explicitly focused on just that arrow, without a subsequent arrow.)

gnprice (commented):
Thanks for the revision! Just one nit remaining.

PIG208:
In addition to the nit, added a test that targets "waiting -> waitPeriodExpired" specifically, for https://github.com/zulip/zulip-flutter/pull/1472#discussion_r2106517535, dropping the emphasis on "waiting -> waitPeriodExpired" from those two "waiting -> waitingPeriodExpired -> (delete)" tests.

gnprice:
Thanks! Those changes all look good; merging.

Issue #855, opened 2024-08-01
Switch from file_picker package to first-party file_selector and/or image_picker


gnprice:
As described at https://github.com/zulip/zulip-flutter/issues/114#issuecomment-1619401171 ,
> now we may be able to replace both uses of `file_picker` with first-party plugins

namely `package:file_selector` and `package:image_picker`.

We didn't use those initially because they didn't at the time have all the features we need. It looks like they now do.

Meanwhile the third-party `package:file_picker` seems to not be vigorously maintained. For example see:
* #825

Kavan2003:
Hey, I would like to work on this project.

PR #1474, opened 2025-04-16
Upgrade Flutter, packages and Android build dependencies

rajveermalviya:


ios/Podfile.lock
```diff
@@ -212,35 +212,35 @@ EXTERNAL SOURCES:
     :path: ".symlinks/plugins/wakelock_plus/ios"
 
 SPEC CHECKSUMS:
-  app_settings: 58017cd26b604ae98c3e65acbdd8ba173703cc82
-  device_info_plus: bf2e3232933866d73fe290f2942f2156cdd10342
+  app_settings: 5127ae0678de1dcc19f2293271c51d37c89428b2
+  device_info_plus: 21fcca2080fbcd348be798aa36c3e5ed849eefbe
```
gnprice, line 219: These Podfile.lock changes appear to revert the changes made in 16e4d88f0df8080d7f03d17cf508e45323c712e1. That's puzzling.

rajveermalviya: This commit was result of running `tools/upgrade flutter-local` and then running the app on both iOS and macOS.

I tried `flutter run` after `flutter clean` on this commit again, to make sure there is nothing wrong with build cache, but it still ends up making these changes.

gnprice: What happens if you run the app before this commit (and with Flutter at the previous version)? If that makes the same change, then that change should happen in its own commit.

These are third-party packages, not dependencies of Flutter, so it'd be surprising if the versions we get for them in CocoaPods are affected by the Flutter version.

It'd also be good to sort out what the difference is between the versions before and after this change. Is `device_info_plus` of 21fcca2080fbcd348be798aa36c3e5ed849eefbe a newer or older version than bf2e3232933866d73fe290f2942f2156cdd10342?

It looks like the current pattern may be that when you run the app on your machine, you get the version from after this change (which is also the version from before 16e4d88), but when Chris runs the app on his machine he gets the version that's currently in main. We'd like to instead get to a recipe that everyone can run (maybe running `pod repo update` and then running the app?) that reproduces the same results on each machine, and for those results to represent an up-to-date version of each dependency.

chrisbobbe:
(I see a few comments about `file_picker`; here's a small reminder that #855 is for moving off of that. 🙂)

pubspec.yaml
```diff
@@ -35,13 +35,13 @@ dependencies:
...
 
-  app_settings: ^5.0.0
+  app_settings: ^6.1.1
   collection: ^1.17.2
   convert: ^3.1.1
   crypto: ^3.0.3
   device_info_plus: ^11.2.0
   drift: ^2.23.0
-  file_picker: ^9.0.2
+  file_picker: ^10.1.2
```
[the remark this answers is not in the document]
rajveermalviya: Right, the docs don't say much, but looking at the implementation:
 - On Android it will [manually compress](https://github.com/miguelpruivo/flutter_file_picker/blob/565cc38cbbaf6b9dee4802b71871d35f4a856880/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt#L368) any image picked by the file picker, if `compressionQuality > 0`, then provide the compressed result.
 - On iOS, it will [set presets for image and video](https://github.com/miguelpruivo/flutter_file_picker/blob/565cc38cbbaf6b9dee4802b71871d35f4a856880/ios/file_picker/Sources/file_picker/FilePickerPlugin.m#L271-L279) files if `compressionQuality` is non-zero, but ignoring the actual `compressionQuality` value, where the OS will do the compression instead of plugin doing it manually.

Previously, the defaults were [`allowCompression = true` and `compressionQuality = 30`](https://github.com/miguelpruivo/flutter_file_picker/blob/1d93d9f51a4805983e7ff449a90bd7d6aeddaf72/lib/src/file_picker.dart#L98-L99). I doubt we needed that, so I guess this breaking change is for the better.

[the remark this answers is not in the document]
gnprice: Great. OK, so the net effect is:
* It used to compress the images.
* Now it doesn't.

Looks like the PR that made the change is this: https://github.com/miguelpruivo/flutter_file_picker/pull/1744

And it was described as fixing two issues, which are both about how the manual compression it did on Android caused problems.

The iOS implementation looks like it should work nice and smoothly, though, so that there it comes down to a product question of whether we want the images (and videos) compressed. Compressing uploaded images is a feature request we've gotten a few times, and considered on our own before that:
https://github.com/zulip/zulip-flutter/issues/1419
https://github.com/zulip/zulip-mobile/issues/4340#issuecomment-771650124
https://github.com/zulip/zulip-mobile/issues/3499
https://github.com/zulip/zulip-mobile/issues/2749

(That latest issue was filed in this repo, but I now suspect it's about zulip-mobile — it says "The current zulip android client" and doesn't give any indication of being aware of the distinction.)

Anyway, let's go ahead with the upgrade and take the new defaults. We'll see if we get feedback one way or another. We'll revise this commit message to be clear about the change, though.

ios/Podfile.lock
rajveermalviya: Added a separate base commit that introduces these `Podfile.lock` changes.

android/app/lint-baseline.xml
```diff
@@ -0,0 +1,121 @@
...
+            file="$GRADLE_USER_HOME/caches/modules-2/files-2.1/org.apache.tika/tika-core/3.1.0/6ba44a9ddf8f6f2b4bc88e8bc64269aea0424607/tika-core-3.1.0.jar"/>
+    </issue>
+
+    <issue
+        id="SyntheticAccessor"
+        message="Access to `private` method `deepEqualsNotifications` of class `Notifications_gKt` requires synthetic accessor"
+        errorLine1="    return deepEqualsNotifications(toList(), other.toList())  }"
+        errorLine2="           ~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/main/kotlin/com/zulip/flutter/Notifications.g.kt"
```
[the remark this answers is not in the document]
rajveermalviya: Well, turns out this just got fixed in https://github.com/flutter/packages/pull/9034. Updated past that now.

rajveermalviya:
Thanks for the review @gnprice! PTAL.

ios/Podfile.lock
gnprice: Cool.

The commit message should mention the fact that it's reverting a previous commit — that's quite relevant information for understanding what we're thinking with this change.

pubspec.yaml
gnprice, line 44: Changelog:
      https://pub.dev/packages/[file_picker/changelog#1012](https://github.com/file_picker/changelog/issues/1012)

This link seems to have gotten corrupted :slightly_smiling_face:

android/app/lint-baseline.xml
```diff
@@ -0,0 +1,121 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 8.5.2" type="baseline" client="gradle" dependencies="false" name="AGP (8.5.2)" variant="all" version="8.5.2">
+
+    <issue
+        id="InvalidPackage"
+        message="Invalid package reference in library; not included in Android: `javax.xml.stream`. Referenced from `org.apache.tika.utils.XMLReaderUtils`.">
+        <location
+            file="$GRADLE_USER_HOME/caches/modules-2/files-2.1/org.apache.tika/tika-core/3.1.0/6ba44a9ddf8f6f2b4bc88e8bc64269aea0424607/tika-core-3.1.0.jar"/>
```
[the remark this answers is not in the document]
gnprice: I think you said in today's call that you looked in the package's tracker and didn't find an issue for this.

That's fine and we can leave it there — this *probably* doesn't affect our app's behavior, and we're planning to switch off this library in the future anyway. The commit message should just mention that, though: the fact that you looked and didn't find an upstream issue.

gnprice (commented):
Thanks for the revision!

These changes all look good. Just a couple more things to mention in the commit messages at this point.

rajveermalviya:

PR #1475, opened 2025-04-17
local echo (6/n): Prepare message list for outbox message support

PIG208:
This is a prep PR for #1453, toward #1441.

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -614,7 +614,10 @@ extension type const TopicName(String _value) {
 /// Different from [MessageDestination], this information comes from
 /// [getMessages] or [getEvents], identifying the conversation that contains a
 /// message.
-sealed class Conversation {}
+sealed class Conversation {
+  /// Whether [this] and [other] refer to the same Zulip conversation.
+  bool isSameAs(Conversation other);
+}
 
 /// The conversation a stream message is in.
 @JsonSerializable(fieldRename: FieldRename.snake, createToJson: false)
@@ -640,6 +643,13 @@ class StreamConversation extends Conversation {
 
   factory StreamConversation.fromJson(Map<String, dynamic> json) =>
     _$StreamConversationFromJson(json);
+
+  @override
+  bool isSameAs(Conversation other) {
+    return other is StreamConversation
+      && streamId == other.streamId
+      && topic.isSameAs(other.topic);
+  }
 }
 
 /// The conversation a DM message is in.
@@ -653,6 +663,21 @@ class DmConversation extends Conversation {
 
   DmConversation({required this.allRecipientIds})
     : assert(isSortedWithoutDuplicates(allRecipientIds.toList()));
+
+  bool _equalIdSequences(Iterable<int> xs, Iterable<int> ys) {
+    if (xs.length != ys.length) return false;
+    final xs_ = xs.iterator; final ys_ = ys.iterator;
+    while (xs_.moveNext() && ys_.moveNext()) {
+      if (xs_.current != ys_.current) return false;
+    }
+    return true;
+  }
+
+  @override
+  bool isSameAs(Conversation other) {
+    if (other is! DmConversation) return false;
+    return _equalIdSequences(allRecipientIds, other.allRecipientIds);
+  }
 }
 
 /// A message or message-like object, for showing in a message list.
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -35,18 +35,31 @@ class MessageListDateSeparatorItem extends MessageListItem {
   MessageListDateSeparatorItem(this.message);
 }
 
-/// A message to show in the message list.
-class MessageListMessageItem extends MessageListItem {
-  final Message message;
-  ZulipMessageContent content;
+/// A [MessageBase] to show in the message list.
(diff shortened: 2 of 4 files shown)


chrisbobbe (commented):
Thanks, this looks good! Small comments below.

lib/model/message_list.dart
```diff
@@ -359,12 +372,14 @@ mixin _MessageSequence {
...
-    if (prevMessage.streamId != message.streamId) return false;
-    if (prevMessage.topic.canonicalize() != message.topic.canonicalize()) return false;
-  } else if (prevMessage is DmMessage && message is DmMessage) {
-    if (!_equalIdSequences(prevMessage.allRecipientIds, message.allRecipientIds)) {
+bool haveSameRecipient(MessageBase prevMessage, MessageBase message) {
+  final prevConversation = prevMessage.conversation;
+  final conversation = message.conversation;
+  if (prevConversation is StreamConversation && conversation is StreamConversation) {
+    if (prevConversation.streamId != conversation.streamId) return false;
+    if (prevConversation.topic.canonicalize() != conversation.topic.canonicalize()) return false;
```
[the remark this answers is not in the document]
PIG208: Good idea! I think we can go further and also pull out `MessageBase.inSameConversationAs` to get rid of `haveSameRecipient` altogether, moving its tests to `test/api/model/model_test.dart`.

PIG208:

lib/api/model/model.dart
```diff
@@ -596,7 +596,10 @@ extension type const TopicName(String _value) {
 /// Different from [MessageDestination], this information comes from
 /// [getMessages] or [getEvents], identifying the conversation that contains a
 /// message.
-sealed class Conversation {}
+sealed class Conversation {
```
chrisbobbe, line 599:
```
api [nfc]: Extract Conversation.isSameAs

This will make it easier to support comparing the conversations
between subclasses of MessageBase.

The message list tests on displayRecipient are now mostly exercising the
logic on Conversation.isSameAs, which makes it reasonable to move the
tests.  Keep them here for now since this logic is more relevant to
message lists then it is to the rest of the app.
```

Commit-message nit: 'than it is'

chrisbobbe (approved):

lib/api/model/model.dart
```diff
@@ -635,6 +645,19 @@ class DmConversation extends Conversation {
...
+  bool isSameAs(Conversation other) {
+    if (other is! DmConversation) return false;
+    final xs = allRecipientIds;
+    final ys = other.allRecipientIds;
+    if (xs.length != ys.length) return false;
+    final xs_ = xs.iterator; final ys_ = ys.iterator;
+    while (xs_.moveNext() && ys_.moveNext()) {
+      if (xs_.current != ys_.current) return false;
+    }
+    return true;
```
gnprice, line 659: This is best kept as its own method with its own name. It's written to a low-level API, but has a meaning that can be described very crisply and abstractly.

lib/api/model/model.dart
```diff
@@ -596,7 +596,10 @@ extension type const TopicName(String _value) {
 /// Different from [MessageDestination], this information comes from
 /// [getMessages] or [getEvents], identifying the conversation that contains a
 /// message.
-sealed class Conversation {}
+sealed class Conversation {
+  /// Whether [this] and [other] have the same canonical form.
+  bool isSameAs(Conversation other);
```
gnprice, line 601:
```suggestion
  /// Whether [this] and [other] refer to the same Zulip conversation.
  bool isSameAs(Conversation other);
```

I'd think of a "canonical form" as a strategy for implementing this sort of method, and the intended semantics as something a bit more abstract.

lib/model/message_list.dart
```diff
@@ -36,17 +36,29 @@ class MessageListDateSeparatorItem extends MessageListItem {
 }
 
 /// A message to show in the message list.
```
gnprice, line 38: doc should get updated

lib/widgets/message_list.dart
```diff
@@ -968,25 +968,32 @@ class MessageItem extends StatelessWidget {
     this.trailingWhitespace,
   });
 
-  final MessageListMessageItem item;
+  final MessageListMessageBaseItem item;
```
gnprice, line 971: nit:

    msglist [nfc]: Handle MessageListMessageBaseItem with MessageItem

Can expand slightly to clarify:

    msglist [nfc]: Handle MessageListMessageBaseItem with MessageItem widget

Otherwise, when looking at the commit message before rereading the code, these two classes sound like the same kind of thing, and then their relationship is confusing.

gnprice: Or I think better yet:

    msglist [nfc]: Let MessageItem widget take a MessageListMessageBaseItem

That basically specifies that we're talking about a field on the widget, vs. the wide range of other things that "handle" could mean.

gnprice (commented):
Thanks both!

Generally these refactors look fine. A couple of them I'm not totally convinced of, so the motivation may become clearer with the PR that uses them. Otherwise, just small comments below.

PIG208:

PIG208:
Moved the commit "msglist [nfc]: Extract _maybeAppendAuxiliaryItem" to #1453.

gnprice:
Thanks. Looks good; merging.

PR #1479, opened 2025-04-23
channel: Maintain maps consistently when removing subscriptions

PIG208:
When a subscription is removed, the data in streams and streamsByName are left as instances of `Subscription`.  This is inconsistent with the data store's assumption that unsubscribed channels should have plain `ZulipStream` instances.

This doesn't seem to have user-facing effects, though.  The assertion errors only affect debug builds.

CZO discussion:
  https://chat.zulip.org/#narrow/channel/48-mobile/topic/Unsubscribe.20then.20resubscribe.20to.20channel/with/2160241

### DIFF
--- a/test/model/channel_test.dart
+++ b/test/model/channel_test.dart
@@ -68,6 +68,23 @@ void main() {
       ));
       checkUnified(store);
     });
+
+    test('unsubscribed then subscribed by events', () async {
+      // Regression test for: https://chat.zulip.org/#narrow/channel/48-mobile/topic/Unsubscribe.20then.20resubscribe.20to.20channel/with/2160241
+      final stream = eg.stream();
+      final store = eg.store();
+      await store.addStream(stream);
+      await store.addSubscription(eg.subscription(stream));
+      checkUnified(store);
+
+      await store.handleEvent(SubscriptionRemoveEvent(id: 1,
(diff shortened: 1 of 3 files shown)


chrisbobbe (approved):

test/model/channel_test.dart
```diff
@@ -68,6 +68,22 @@ void main() {
       ));
       checkUnified(store);
     });
+
+    test('removed then added by events', () async {
```
gnprice, line 72:
```suggestion
    test('unsubscribed then subscribed by events', () async {
```

The group is about "stream/sub data" in general, so this otherwise sounds like it's probably about removing the whole stream.

gnprice: Separately, since this is a regression test for a particular bug that we actually had, let's link that context:
```suggestion
    test('removed then added by events', () async {
      // Regression test for: https://chat.zulip.org/#narrow/channel/48-mobile/topic/Unsubscribe.20then.20resubscribe.20to.20channel/with/2160241
```

gnprice (commented):
Thanks for the fix! Just a couple of nits below.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Looks good; merging.

PR #1480, opened 2025-04-23
scrolling test: Skip tests against old default z-order of slivers

gnprice:
These two tests on CustomPaintOrderScrollView compare its behavior to the default z-order / paint order behavior of CustomScrollView. They'd therefore be broken by my upstream PR that gives CustomScrollView the functionality of CustomPaintOrderScrollView and simplifies its default behavior:
  https://github.com/flutter/flutter/pull/164818

So skip the tests.  Once that PR lands, we can update past it and then remove CustomPaintOrderScrollView and its tests entirely.

(The customer_tests failures currently seen on the upstream PR are different, because the Zulip pin in flutter/tests is currently at d302584e5.  The failures there are in the sticky_header tests, which want to exercise a variety of sliver configurations and need to control the z-order when they do so.  As of d302584e5, they did so by relying on the specifics of the existing CustomScrollView behavior.  Since then we merged CustomPaintOrderScrollView, and in 49eff8bb3 adapted those tests to use it, making them immune to changes in the default z-order; but that also meant adding CustomPaintOrderScrollView's own tests, including these two.)

### DIFF
--- a/test/widgets/scrolling_test.dart
+++ b/test/widgets/scrolling_test.dart
@@ -43,12 +43,6 @@ void main() {
       check(paintLog).deepEquals([0, 1, 2, 3, 4]);
     });
 
-    // This test will fail if a corresponding upstream PR lands:
-    //   https://github.com/flutter/flutter/pull/164818
-    // because that eliminates the quirky centerTopFirstBottom behavior.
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Thanks, LGTM! Merged.

Issue #1470, opened 2025-04-09
compose: In topic or DM narrow, compose progress is lost when event queue expires


chrisbobbe:
To reproduce:

1. Open a DM narrow or a topic narrow
2. Type some text into the compose box
3. Put the app in the background and wait at least 10 minutes
4. Foreground the app; wait for new-queue loading indicator to disappear

Expected behavior: The text in the compose box is still there

Actual behavior: The text is cleared

In f1646c4939685c617e55b82bc7e8badd7b7dd037, `_ComposeBoxState` started using `PerAccountStoreAwareStateMixin`. For DM and topic narrows, its `onNewStore` implementation drops the `FixedDestinationComposeBoxController` on the floor (without calling `dispose`) and makes a new one. It should only make a new one if it's the first time it's called.

Code:

<details>

```dart
  @override
  void onNewStore() {
    switch (widget.narrow) {
      case ChannelNarrow():
        final store = PerAccountStoreWidget.of(context);
        if (_controller == null) {
          _controller = StreamComposeBoxController(store: store);
        } else {
          (controller as StreamComposeBoxController).topic.store = store;
        }
      case TopicNarrow():
      case DmNarrow():
        _controller = FixedDestinationComposeBoxController();
      case CombinedFeedNarrow():
      case MentionsNarrow():
      case StarredMessagesNarrow():
        assert(false);
    }
  }
```

</details>

Possible fix (untested):

<details>

```diff
diff --git lib/widgets/compose_box.dart lib/widgets/compose_box.dart
index 966bec5f2..8d4b5fccc 100644
--- lib/widgets/compose_box.dart
+++ lib/widgets/compose_box.dart
@@ -1533,7 +1533,7 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
         }
       case TopicNarrow():
       case DmNarrow():
-        _controller = FixedDestinationComposeBoxController();
+        _controller ??= FixedDestinationComposeBoxController();
       case CombinedFeedNarrow():
       case MentionsNarrow():
       case StarredMessagesNarrow():
```

</details>

Assigning @PIG208 as the author of f1646c4939685c617e55b82bc7e8badd7b7dd037.

chrisbobbe:
I'm refactoring this code on the way to #126, so I'll take this issue as part of that work.

PR #1482, opened 2025-04-24
Edit-message (5/n): Fix bug where compose progress can be lost on new event queue

chrisbobbe:
And factor out a helper for the part that reads `widget.narrow` and sets a new controller accordingly; we'll use this helper for the edit-message UI, to reset the compose box at the end of an edit-message session.

Fixes: #1470

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1523,14 +1523,26 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
 
   @override
   void onNewStore() {
+    final newStore = PerAccountStoreWidget.of(context);
+
+    final controller = _controller;
+    if (controller == null) {
+      _setNewController(newStore);
+      return;
+    }
+
+    switch (controller) {
+      case StreamComposeBoxController():
+        controller.topic.store = newStore;
+      case FixedDestinationComposeBoxController():
+        // no reference to the store that needs updating
+    }
+  }
+
+  void _setNewController(PerAccountStore store) {
     switch (widget.narrow) {
       case ChannelNarrow():
-        final store = PerAccountStoreWidget.of(context);
-        if (_controller == null) {
-          _controller = StreamComposeBoxController(store: store);
-        } else {
-          (controller as StreamComposeBoxController).topic.store = store;
-        }
+        _controller = StreamComposeBoxController(store: store);
       case TopicNarrow():
       case DmNarrow():
         _controller = FixedDestinationComposeBoxController();
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -28,6 +28,7 @@ import '../api/fake_api.dart';
 import '../example_data.dart' as eg;
 import '../flutter_checks.dart';
 import '../model/binding.dart';
+import '../model/store_checks.dart';
 import '../model/test_store.dart';
 import '../model/typing_status_test.dart';
 import '../stdlib_checks.dart';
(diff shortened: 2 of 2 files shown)


test/widgets/compose_box_test.dart
```diff
@@ -1239,4 +1245,41 @@ void main() {
...
+      store.updateMachine!
+        ..debugPauseLoop()
+        ..poll()
+        ..debugPrepareLoopError(
+            eg.apiExceptionBadEventQueueId(queueId: store.queueId))
+        ..debugAdvanceLoop();
+      await tester.pump();
+
+      // new store replaces the old one
+      final newStore = testBinding.globalStore.perAccountSync(eg.selfAccount.id)!;
```
PIG208, line 1272: How about this?

```suggestion
      final newStore = testBinding.globalStore.perAccountSync(store.accountId)!;
```

This way we don't rely on `prepareComposeBox` using a specific value for self-account ID.

test/widgets/compose_box_test.dart
```diff
@@ -1239,4 +1245,41 @@ void main() {
...
+      check(store).not((it) => it.identicalTo(newStore));
+
+      // new store has the same basic boring data so the compose box
+      // has its content input instead of no-posting-permission error banner
+      // TODO instead, add data by preparing new store's initial snapshot
+      await newStore.addStream(channel);
+      await newStore.addUser(eg.selfUser);
```
PIG208, line 1279: Maybe the TODO just calls for moving setup code into the initial snapshot in `prepareComposeBox`?

```diff
     await testBinding.globalStore.add(selfAccount, eg.initialSnapshot(
+      realmUsers: [selfUser, ...otherUsers],
+      streams: streams,
       zulipFeatureLevel: zulipFeatureLevel,
       realmMandatoryTopics: mandatoryTopics,
     ));
 
     store = await testBinding.globalStore.perAccount(selfAccount.id);
-
-    await store.addUsers([selfUser, ...otherUsers]);
-    await store.addStreams(streams);
```

The initial call to `testBinding.globalStore.add` sets the initial snapshot that will also be used when reloading the store.

lib/widgets/compose_box.dart
```diff
@@ -1523,14 +1523,26 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
 
   @override
   void onNewStore() {
+    final newStore = PerAccountStoreWidget.of(context);
+
+    if (_controller == null) {
+      _setNewControllerForNarrow(newStore);
+    } else {
```
PIG208, line 1530: nit: perhaps we can return early here and de-indent the else-block

PIG208 (commented):
Thanks! The fix looks good to me with manual testing. Left some comments.

test/widgets/compose_box_test.dart
chrisbobbe: Ooh, thanks for thinking of this.

test/widgets/compose_box_test.dart
chrisbobbe: > The initial call to `testBinding.globalStore.add` sets the initial snapshot that will also be used when reloading the store.

Ohh, I didn't realize; thanks for pointing this out.

chrisbobbe:

PIG208:
Thanks! Marking for Greg's review.

lib/widgets/compose_box.dart
```diff
@@ -1523,14 +1523,27 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
...
+    final newStore = PerAccountStoreWidget.of(context);
+
+    if (_controller == null) {
+      _setNewControllerForNarrow(newStore);
+      return;
+    }
+
+    switch (_controller) {
+      case StreamComposeBoxController():
+        (_controller as StreamComposeBoxController).topic.store = newStore;
```
gnprice, line 1535: This can be simplified by extracting a local variable as a copy of the field, so that type promotion works:

```suggestion
    final controller = _controller;
    switch (controller) {
      case StreamComposeBoxController():
        controller.topic.store = newStore;
```

lib/widgets/compose_box.dart
```diff
@@ -1523,14 +1523,27 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
...
+      case FixedDestinationComposeBoxController():
+        break; // no reference to the store that needs updating
+      case null:
+        assert(false);
+    }
```
gnprice, line 1540: nit: Similarly if the local is used for the initial null check, then the analyzer sees this case is impossible.

lib/widgets/compose_box.dart
```diff
@@ -1523,14 +1523,27 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
...
+  }
+
+  void _setNewControllerForNarrow(PerAccountStore store) {
```
gnprice, line 1543: nit: drop "for narrow"? Seems like it's just the new controller that's appropriate for this compose box.

gnprice (commented):
Thanks both! Looks good; small comments.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merging.

PR #1483, opened 2025-04-24
Edit-message (6/n): Misc. preparation for edit-message feature

chrisbobbe:
Here's another small batch of miscellaneous prep work for the edit-message feature, #126.


lib/widgets/action_sheet.dart
```diff
@@ -562,7 +562,7 @@ void showMessageActionSheet({required BuildContext context, required Message mes
   // So we rely on the fact that isComposeBoxOffered for any given message list
   // will be constant through the page's life.
   final messageListPage = MessageListPage.ancestorOf(pageContext);
-  final isComposeBoxOffered = messageListPage.composeBoxController != null;
+  final isComposeBoxOffered = messageListPage.composeBoxState?.controller != null;
```
PIG208, line 565: Could this just be

```suggestion
  final isComposeBoxOffered = messageListPage.composeBoxState != null;
```

As the dartdoc of `composeBoxState` states:

```dart
  /// The [ComposeBoxState] for this [MessageListPage]'s compose box,
  /// if this [MessageListPage] offers a compose box and it has mounted,
  /// else null.
  ComposeBoxState? get composeBoxState;
```

Checking `composeBoxState?.controller != null` seems a bit confusing use of the API to me.

test/widgets/message_list_test.dart
```diff
@@ -136,20 +136,20 @@ void main() {
...
 
-    testWidgets('composeBoxController finds compose box', (tester) async {
+    testWidgets('composeBoxState finds compose box, with controller', (tester) async {
       final stream = eg.stream();
       await setupMessageListPage(tester, narrow: ChannelNarrow(stream.streamId),
         streams: [stream],
         messages: [eg.streamMessage(stream: stream, content: "<p>a message</p>")]);
       final state = MessageListPage.ancestorOf(tester.element(find.text("a message")));
-      check(state.composeBoxController).isNotNull();
+      check(state.composeBoxState?.controller).isNotNull();
```
PIG208, line 145: Similar to the earlier comment on a null-check, I feel that checking if `composeBoxState?.controller` is null will make more sense if the controller is nullable. The updated contract seems to be: as long as `composeBoxState` is present, there is a compose box and vice versa.

lib/widgets/action_sheet.dart
```diff
@@ -906,7 +858,7 @@ class CopyMessageTextButton extends MessageActionSheetMenuItemButton {
```
PIG208, line 858: nit: perhaps update the reference here?
```
  @override void onPressed() async {
    // This action doesn't show request progress.
    // But hopefully it won't take long at all; and
    // fetchRawContentWithFeedback has a TODO for giving feedback if it does.
```

PIG208 (commented):
Thanks! Left some small comments. This otherwise looks good to me.

lib/widgets/action_sheet.dart
chrisbobbe: Oh, indeed.

test/widgets/message_list_test.dart
chrisbobbe: Yeah; I think what happened is I did something like a copy-paste and didn't make this refinement from there :)

chrisbobbe:

PIG208:
The updates LGTM. Thanks! Marking for Greg's review.

lib/widgets/content.dart
```diff
@@ -296,13 +296,16 @@ class MessageContent extends StatelessWidget {
...
+    };
+
     return InheritedMessage(message: message,
       child: DefaultTextStyle(
         style: ContentTheme.of(context).textStylePlainParagraph,
-        child: switch (content) {
-          ZulipContent() => BlockContentList(nodes: content.nodes),
-          PollContent()  => PollWidget(messageId: message.id, poll: content.poll),
-        }));
+        child: child));
```
gnprice, line 308: This part might become clearer in the upcoming PR that will use this change. But I'm not sure that here in MessageContent is the right place to implement this behavior:

    content [nfc]: Pull out `child` variable in MessageContent.build
    
    For the edit-message feature, the Figma fades message content during
    the loading state:
      https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4026-8563&m=dev
    This refactor prepares for that.

It feels like it might more naturally live in the parent widget.

(Which could do it just as well, right? I imagine the fading doesn't interact with the DefaultTextStyle or the InheritedMessage.)

gnprice (commented):
Thanks both! All looks good except a comment on the last commit.

lib/widgets/content.dart
chrisbobbe: Indeed; yeah, I'll drop this commit.

chrisbobbe:
Thanks! Revision pushed.

PR #1484, opened 2025-04-24
Edit-message (7/n): Implement edit-message methods on MessageStore

chrisbobbe:
This PR, toward the edit-message feature #126, implements three new methods on `MessageStore`:

`editMessage`, to be called when the user taps the "Edit message" button in the message action sheet:

`getEditMessageErrorStatus`, to be called when rendering the message list; showing a loading state when we haven't gotten the update-message event yet, and an error state for a failed edit request:

image

image

(The UI text isn't final; see updates on CZO: [#mobile-design > edit message @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/edit.20message/near/1930073))

`takeFailedMessageEdit`, to be called when you tap the error message pictured above, so we can put you back in the edit box, starting with the text you'd tried to edit to. This part is from CZO discussion: [#mobile-design > edit message @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/edit.20message/near/2129901)

Related: #126

### DIFF
--- a/lib/model/message.dart
+++ b/lib/model/message.dart
@@ -35,6 +35,36 @@ mixin MessageStore {
   /// All [Message] objects in the resulting list will be present in
   /// [this.messages].
   void reconcileMessages(List<Message> messages);
+
+  /// Whether the current edit request for the given message, if any, has failed.
+  ///
+  /// Will be null if there is no current edit request.
+  /// Will be false if the current request hasn't failed
+  /// and the update-message event hasn't arrived.
+  bool? getEditMessageErrorStatus(int messageId);
+
+  /// Edit a message's content, via a request to the server.
+  ///
+  /// Should only be called when there is no current edit request for [messageId],
+  /// i.e., [getEditMessageErrorStatus] returns null for [messageId].
+  ///
+  /// See also:
+  ///   * [getEditMessageErrorStatus]
+  ///   * [takeFailedMessageEdit]
+  void editMessage({required int messageId, required String newContent});
+
+  /// Forgets the failed edit request and returns the attempted new content.
+  ///
+  /// Should only be called when there is a failed request,
+  /// per [getEditMessageErrorStatus].
+  String takeFailedMessageEdit(int messageId);
+}
+
+class _EditMessageRequestStatus {
+  _EditMessageRequestStatus({required this.hasError, required this.newContent});
+
+  bool hasError;
+  final String newContent;
 }
 
 class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
@@ -132,6 +162,56 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
     }
   }
 
+  @override
+  bool? getEditMessageErrorStatus(int messageId) =>
+    _editMessageRequests[messageId]?.hasError;
+
+  final Map<int, _EditMessageRequestStatus> _editMessageRequests = {};
+
+  @override
+  void editMessage({
+    required int messageId,
+    required String newContent,
+  }) async {
+    if (_editMessageRequests.containsKey(messageId)) {
+      throw StateError('an edit request is already in progress');
+    }
+
+    _editMessageRequests[messageId] = _EditMessageRequestStatus(
+      hasError: false, newContent: newContent);
+    _notifyMessageListViewsForOneMessage(messageId);
+    try {
+      await updateMessage(connection, messageId: messageId, content: newContent);
+      // On success, we'll clear the status from _editMessageRequests
+      // when we get the event.
+    } catch (e) {
+      // TODO(log) if e is something unexpected
+
+      final status = _editMessageRequests[messageId];
+      if (status == null) {
+        // The event actually arrived before this request failed
+        // (can happen with network issues).
+        // Or, the message was deleted.
+        return;
+      }
+      status.hasError = true;
+      _notifyMessageListViewsForOneMessage(messageId);
+    }
+  }
+
+  @override
+  String takeFailedMessageEdit(int messageId) {
+    final status = _editMessageRequests.remove(messageId);
+    _notifyMessageListViewsForOneMessage(messageId);
+    if (status == null) {
+      throw StateError('called takeFailedMessageEdit, but no edit');
+    }
+    if (!status.hasError) {
+      throw StateError("called takeFailedMessageEdit, but edit hasn't failed");
+    }
+    return status.newContent;
+  }
+
   void handleUserTopicEvent(UserTopicEvent event) {
     for (final view in _messageListViews) {
       view.handleUserTopicEvent(event);
@@ -183,6 +263,12 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
       // The message is guaranteed to be edited.
       // See also: https://zulip.com/api/get-events#update_message
       message.editState = MessageEditState.edited;
+
+      // Clear the edit-message progress feedback.
+      // This makes a rare bug where we might clear the feedback too early,
+      // if the user raced with themself to edit the same message
+      // from multiple clients.
+      _editMessageRequests.remove(message.id);
     }
     if (event.renderedContent != null) {
       assert(message.contentType == 'text/html',
@@ -245,6 +331,7 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
   void handleDeleteMessageEvent(DeleteMessageEvent event) {
     for (final messageId in event.messageIds) {
       messages.remove(messageId);
+      _editMessageRequests.remove(messageId);
     }
     for (final view in _messageListViews) {
       view.handleDeleteMessageEvent(event);
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -747,6 +747,21 @@ class PerAccountStore extends PerAccountStoreBase with ChangeNotifier, EmojiStor
     // TODO(#649) notify [unreads] of the just-fetched messages
     // TODO(#650) notify [recentDmConversationsView] of the just-fetched messages
   }
+  @override
+  bool? getEditMessageErrorStatus(int messageId) {
+    assert(!_disposed);
+    return _messages.getEditMessageErrorStatus(messageId);
+  }
+  @override
+  void editMessage({required int messageId, required String newContent}) {
+    assert(!_disposed);
+    return _messages.editMessage(messageId: messageId, newContent: newContent);
+  }
+  @override
+  String takeFailedMessageEdit(int messageId) {
+    assert(!_disposed);
+    return _messages.takeFailedMessageEdit(messageId);
+  }
 
   @override
   Set<MessageListView> get debugMessageListViews => _messages.debugMessageListViews;
--- a/lib/model/store.dart
+++ b/lib/model/store.dart
@@ -737,6 +737,11 @@ class PerAccountStore extends PerAccountStoreBase with ChangeNotifier, EmojiStor
   void unregisterMessageList(MessageListView view) =>
     _messages.unregisterMessageList(view);
   @override
+  Future<void> sendMessage({required MessageDestination destination, required String content}) {
+    assert(!_disposed);
+    return _messages.sendMessage(destination: destination, content: content);
+  }
+  @override
   void reconcileMessages(List<Message> messages) {
     _messages.reconcileMessages(messages);
     // TODO(#649) notify [unreads] of the just-fetched messages
@@ -902,12 +907,6 @@ class PerAccountStore extends PerAccountStoreBase with ChangeNotifier, EmojiStor
       case UnexpectedEvent():
         assert(debugLog("server event: ${jsonEncode(event.toJson())}")); // TODO log better
     }
-  }
-
-  @override
-  Future<void> sendMessage({required MessageDestination destination, required String content}) {
-    assert(!_disposed);
-    return _messages.sendMessage(destination: destination, content: content);
   }
 
   static List<CustomProfileField> _sortCustomProfileFields(List<CustomProfileField> initialCustomProfileFields) {
--- a/test/model/message_test.dart
+++ b/test/model/message_test.dart
@@ -1,10 +1,13 @@
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:checks/checks.dart';
+import 'package:http/http.dart' as http;
 import 'package:test/scaffolding.dart';
 import 'package:zulip/api/model/events.dart';
 import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/model/submessage.dart';
+import 'package:zulip/api/route/messages.dart';
 import 'package:zulip/model/message_list.dart';
 import 'package:zulip/model/narrow.dart';
 import 'package:zulip/model/store.dart';
@@ -13,6 +16,7 @@ import '../api/fake_api.dart';
 import '../api/model/model_checks.dart';
 import '../api/model/submessage_checks.dart';
 import '../example_data.dart' as eg;
+import '../fake_async.dart';
 import '../stdlib_checks.dart';
 import 'message_list_test.dart';
 import 'store_checks.dart';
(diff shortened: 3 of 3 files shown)


chrisbobbe:
Marking as a draft because I still have a TODO for some tests about delete-message events.

test/model/message_test.dart
```diff
@@ -123,6 +127,179 @@ void main() {
...
+      store.editMessage(messageId: message.id, content: 'new content');
+
+      async.elapse(Duration(seconds: 1));
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isTrue();
+
+      await store.handleEvent(eg.updateMessageEditEvent(message));
+      check(store.getEditMessageErrorStatus(message.id)).isNull();
+    }));
+
+    test('request failure before event arrival; take failed edit in between', () => awaitFakeAsync((async) async {
```
PIG208, line 232: nit: the format "request fails, then message deleted" seems preferable to me to help understand the sequence of events

test/model/message_test.dart
```diff
@@ -123,6 +127,179 @@ void main() {
...
+      connection.prepare(apiException: eg.apiBadRequest(), delay: Duration(seconds: 1));
+      store.editMessage(messageId: message.id, content: 'new content');
+      async.elapse(Duration(seconds: 1));
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isTrue();
+
+      check(store.takeFailedMessageEdit(message.id)).equals('new content');
+      check(store.getEditMessageErrorStatus(message.id)).isNull();
+    }));
+
+    test('takeFailedMessageEdit throws StateError when nothing to take', () => awaitFakeAsync((async) async {
```
PIG208, line 191: I think a similarly useful test would be repeated `editMessage` calls that operate on the same message.

test/model/message_test.dart
```diff
@@ -123,6 +127,179 @@ void main() {
...
+  group('edit-message methods', () {
+    late PerAccountStore store;
+    late FakeApiConnection connection;
+    late StreamMessage message;
+
+    Future<void> prepare() async {
+      store = eg.store();
+      connection = store.connection as FakeApiConnection;
+      message = eg.streamMessage();
+      await store.addMessage(message);
```
PIG208, line 139: I think we can also test if the listeners are notified whenever an update is expected.

lib/model/message.dart
```diff
@@ -35,6 +35,34 @@ mixin MessageStore {
...
+  /// Will be false if the current request hasn't failed
+  /// and the update-message event hasn't arrived.
+  bool? getEditMessageErrorStatus(int messageId);
+
+  /// Edits a message's content.
+  ///
+  /// See also:
+  ///   * [getEditMessageErrorStatus]
+  ///   * [takeFailedMessageEdit]
+  void editMessage({required int messageId, required String content});
```
PIG208, line 52: I think we should document the condition when you should not call `editMessage` (like we do for `takeFailedMessageEdit`), i.e., when there is a current edit message request.

PIG208 (commented):
Thanks! Left some comments.

test/model/message_test.dart
```diff
@@ -123,6 +127,179 @@ void main() {
...
+          'content': 'new content',
+        });
+
+      async.elapse(Duration(milliseconds: 500));
+      // Mid-request
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isFalse();
+
+      async.elapse(Duration(milliseconds: 500));
+      // Request has succeeded; event hasn't arrived
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isFalse();
```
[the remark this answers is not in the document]
chrisbobbe: Yeah, there's nothing user-facing to do when the future succeeds (we take the event as signaling edit-message success), so `editMessage` doesn't return it.

chrisbobbe:

test/model/message_test.dart
```diff
@@ -123,6 +127,269 @@ void main() {
...
+        json: UpdateMessageResult().toJson(), delay: Duration(seconds: 1));
+      store.editMessage(messageId: otherMessage.id, content: 'other message new content');
+      checkRequest(otherMessage.id, 'other message new content');
+      checkNotifiedOnce();
+
+      async.elapse(Duration(milliseconds: 500));
+      // First request has succeeded; event hasn't arrived
+      // Mid-second request
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isFalse();
+      check(store.getEditMessageErrorStatus(otherMessage.id)).isNotNull().isFalse();
```
PIG208, line 199: How about having `checkNotNotified` calls here and later? So that we are extra sure that it's the edit event that triggers the notifications.

test/model/message_test.dart
```diff
@@ -123,6 +127,269 @@ void main() {
...
+      await prepareEditMessage();
+
+      connection.prepare(
+        json: UpdateMessageResult().toJson(), delay: Duration(seconds: 1));
+      store.editMessage(messageId: message.id, content: 'new content');
+      async.elapse(Duration(milliseconds: 500));
+      check(connection.takeRequests()).length.equals(1);
+      checkNotifiedOnce();
+
+      connection.prepare(json: UpdateMessageResult().toJson());
```
PIG208, line 260: I'm a bit surprised that this is needed, since I thought `store.editMessage` would fail before firing the request, which is confirmed by the later `.isEmpty` check on `takeRequests()`. If I missed something, maybe a comment here will be helpful?

test/model/message_test.dart
```diff
@@ -123,6 +127,269 @@ void main() {
...
+      await prepareEditMessage();
+      check(store.getEditMessageErrorStatus(message.id)).isNull();
+
+      connection.prepare(apiException: eg.apiBadRequest(), delay: Duration(seconds: 1));
+      store.editMessage(messageId: message.id, content: 'new content');
+      checkNotifiedOnce();
+
+      async.elapse(Duration(milliseconds: 500));
+      // Mid-request
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isFalse();
```
PIG208, line 357: Similarly, having a `checkNotNotified` call here might help.

test/model/message_test.dart
```diff
@@ -123,6 +127,269 @@ void main() {
...
+      check(store.getEditMessageErrorStatus(message.id)).isNull();
+
+      connection.prepare(
+        json: UpdateMessageResult().toJson(), delay: Duration(seconds: 1));
+      store.editMessage(messageId: message.id, content: 'new content');
+      checkNotifiedOnce();
+
+      async.elapse(Duration(milliseconds: 500));
+      // Mid-request
+      check(store.getEditMessageErrorStatus(message.id)).isNotNull().isFalse();
```
PIG208, line 380: Same here.

PIG208 (commented):
Thanks! Just some small comments on tests (and the failing CI might need a fix). Marking this for Greg's review.

test/model/message_test.dart
chrisbobbe: Ah yeah, this can be removed.

lib/model/message.dart
```diff
@@ -35,6 +35,37 @@ mixin MessageStore {
...
+
+  /// Whether the current edit request, if any, has failed.
+  ///
+  /// Will be null if there is no current edit request
+  /// or the message was deleted.
+  /// Will be false if the current request hasn't failed
+  /// and the update-message event hasn't arrived.
+  bool? getEditMessageErrorStatus(int messageId);
+
+  /// Edits a message's content.
```
gnprice, line 47: Given the context of being on [MessageStore] (or [PerAccountStore]), this reads as if it's making an edit to the data we have locally. The fact that it's making an API request over the network is an important part of its meaning, so should be made explicit.

E.g.:
```suggestion
  /// Edit a message's content, via a request to the server.
```

lib/model/message.dart
```diff
@@ -35,6 +35,37 @@ mixin MessageStore {
   /// All [Message] objects in the resulting list will be present in
   /// [this.messages].
   void reconcileMessages(List<Message> messages);
+
+  /// Whether the current edit request, if any, has failed.
```
gnprice, line 39: This reads to me as if there's one "current edit request" on the store. How about:
```suggestion
  /// Whether the current edit request for the given message, if any, has failed.
```

lib/model/message.dart
```diff
@@ -35,6 +35,37 @@ mixin MessageStore {
...
+  ///
+  /// Will be null if there is no current edit request
+  /// or the message was deleted.
```
gnprice, line 42: I think I'd want to conceptualize "the message was deleted" as one way to (re-)enter the state of there being no current edit request. If the message was deleted, then any edit request for it that had been outstanding no longer is.

lib/model/message.dart
```diff
@@ -132,6 +163,55 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
     }
   }
 
+  @override
+  bool? getEditMessageErrorStatus(int messageId) =>
+    _editMessageRequests[messageId]?.hasError;
+  final Map<int, _EditMessageRequestStatus> _editMessageRequests = {};
```
gnprice, line 169: nit: this field is a bit more than the backing for this getter-like method:

```suggestion
  @override
  bool? getEditMessageErrorStatus(int messageId) =>
    _editMessageRequests[messageId]?.hasError;

  final Map<int, _EditMessageRequestStatus> _editMessageRequests = {};
```

because in particular it has more data than that exposes.

lib/model/message.dart
```diff
@@ -132,6 +163,55 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
...
+
+  @override
+  void editMessage({
+    required int messageId,
+    required String content,
+  }) async {
+    if (_editMessageRequests.containsKey(messageId)) {
+      throw StateError('message ID already in editMessageRequests');
```
gnprice, line 177: nit: `editMessageRequests` isn't quite the name of anything; and the nearest name to that is private so might not be clearest for an error message:

```suggestion
    if (_editMessageRequests.containsKey(messageId)) {
      throw StateError('an edit request is already in progress');
```

(though a private identifier in this message wouldn't be terrible either, since it really should be only developer-facing)

lib/model/message.dart
```diff
@@ -132,6 +163,55 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore {
...
+    }
+
+    _editMessageRequests[messageId] = _EditMessageRequestStatus(
+      hasError: false, content: content);
+    _notifyMessageListViewsForOneMessage(messageId);
+    try {
+      await updateMessage(connection, messageId: messageId, content: content);
+      // On success, we'll clear `status` from editMessageRequests
+      // when we get the event.
```
gnprice, line 186: Is there an identifier `status` this refers to? I don't see it.

lib/model/store.dart
```diff
@@ -910,6 +910,24 @@ class PerAccountStore extends PerAccountStoreBase with ChangeNotifier, EmojiStor
...
+  @override
+  bool? getEditMessageErrorStatus(int messageId) {
+    assert(!_disposed);
+    return _messages.getEditMessageErrorStatus(messageId);
+  }
+
+  @override
+  void editMessage({required int messageId, required String content}) {
+    assert(!_disposed);
+    return _messages.editMessage(messageId: messageId, content: content);
```
gnprice, line 922: nit: let's move these up to sit next to the other MessageStore proxy implementations

(Should do the same with `sendMessage` too. I guess that should have happened when we moved `sendMessage` to MessageStore, 942aa87a52dc5b911c1f04b9ed774536e1ce1518 / #1455, oh well.)

test/model/message_test.dart
```diff
@@ -123,6 +127,273 @@ void main() {
...
+        json: UpdateMessageResult().toJson(), delay: Duration(seconds: 1));
+      store.editMessage(messageId: message.id, content: 'new content');
+      async.elapse(Duration(milliseconds: 500));
+      check(connection.takeRequests()).length.equals(1);
+      checkNotifiedOnce();
+
+      // A function marked async needs an async check,
+      // but its return type is void, so, cast the return value to Future
+      final future = store.editMessage(messageId: message.id, content: 'newer content') as Future;
+      await check(future).throws<StateError>();
```
gnprice, line 266: Hmm interesting. How about:
```suggestion
      await check(store.editMessage(messageId: message.id, content: 'newer content'))
        .isA<Future<void>>().throws<StateError>();
```

Effectively that moves the cast into a `package:checks` condition.

test/model/message_test.dart
```diff
@@ -123,6 +127,273 @@ void main() {
...
+      checkNotifiedOnce();
+
+      async.elapse(Duration(milliseconds: 500));
+      await store.handleEvent(eg.updateMessageEditEvent(message));
+      check(store.getEditMessageErrorStatus(message.id)).isNull();
+      checkNotifiedOnce();
+
+      async.elapse(Duration(milliseconds: 500));
+      check(store.getEditMessageErrorStatus(message.id)).isNull();
+      checkNotNotified();
```
gnprice, line 288: This version is slightly delicate, in that these two checks are only effective given that the `elapse` calls add up to at least the delay on the error response. The test is correct, but vulnerable to being defused in a not-entirely-obvious way by a future edit to those numbers.

How about:
```suggestion
      async.flushTimers();
      check(store.getEditMessageErrorStatus(message.id)).isNull();
      checkNotNotified();
```

(It's basically OK as is, though; if there weren't such an easy way to make it more robust in that direction, I wouldn't want to attempt anything complex to guard against that sort of edit.)

gnprice (commented):
Thanks! This looks great; just nits below, and a few comments on the docs.

chrisbobbe:

In this revision, I did a few more things:
- s/content/newContent/
- Added support for `prev_content_sha256`

chrisbobbe:
(updated `takeFailedMessageEdit`'s return type from `String?` to `String`)

chrisbobbe:
Hmm, a CI failure:

```
Run flutter precache --universal
Downloading Linux x64 Dart SDK from Flutter engine ae352af9d0fb96896310e[7](https://github.com/chrisbobbe/zulip-flutter/actions/runs/14802472286/job/41564208589#step:6:8)dff8504c761d45d519...

100   255  100   255    0     0   20[8](https://github.com/chrisbobbe/zulip-flutter/actions/runs/14802472286/job/41564208589#step:6:9)2      0 --:--:-- --:--:-- --:--:--  2090

```

I'll rerun and see if that fixes it.

lib/api/route/messages.dart
```diff
@@ -284,6 +288,8 @@ Future<UpdateMessageResult> updateMessage(
     if (sendNotificationToNewThread != null) 'send_notification_to_new_thread': sendNotificationToNewThread,
     if (content != null) 'content': RawParameter(content),
     if (streamId != null) 'stream_id': streamId,
+    // TODO(server-11) remove FL condition and its mention in the dartdoc
+    if (prevContentSha256 != null && connection.zulipFeatureLevel! >= 379) 'prev_content_sha256': RawParameter(prevContentSha256),
```
gnprice, line 292: We should be able to just send this unconditionally — generally the server ignores unexpected parameters in requests:
https://zulip.com/api/rest-error-handling#ignored-parameters

(That doc is about some feedback the server includes in responses when that happens, and the feedback is new in server-7. But I believe the server has tolerated such parameters since forever.)

lib/api/route/messages.dart
```diff
@@ -276,6 +279,7 @@ Future<UpdateMessageResult> updateMessage(
   bool? sendNotificationToNewThread,
   String? content,
   int? streamId,
+  String? prevContentSha256,
```
gnprice, line 282: nit: follow order in API doc (which is also a logical order — groups the two content-related parameters together):
```suggestion
  String? content,
  String? prevContentSha256,
  int? streamId,
```

gnprice (commented):
Thanks! The revision looks good. A couple of comments both on the new commits at the end.

We could merge the first part:
285c1f467 store [nfc]: Move store.sendMessage up near other MessageStore proxy impls
f1c0e0f24 model: Implement edit-message methods on MessageStore

and then do the next two as a follow-up PR:
2c6936454 api: Add prevContentSha256 param to updateMessage route
1eeb197ef model: Use prevContentSha256 in edit-message method

PR #223, opened 2023-07-18
msglist: add scroll-to-bottom button to MessageList

sirpengi:
Introducing a scroll-to-bottom button in the `MessageList` component that shows up only if the user is not at the bottom of the list. Clicking on that button triggers a scroll to the bottom.

## Notes
- My initial version wrapped the MessageList with another component that controlled scrolling, but I was able to fix the re-rendering issue while putting integrating the functionality right in the component. A benefit of having wrapper component may be to reuse the scroll-to-bottom feature but I don't believe that is used in any other context (the list of streams or users do not have this).
- I did not add a test case yet, I would prefer to wait for a greenlight on the implemented component as the test case would need some knowledge of implementation to trigger and check results

PR #1486, opened 2025-04-25
anchors 6/n: Make scroll-to-end work for split slivers

### LINKED ISSUE #1485: Scroll-to-bottom button is abrupt and janks when scrolling a long way
### Background

In the message list, if you hit the scroll-to-bottom button, the intended behavior is that it scrolls to the bottom at an average speed of at most 8000 px/sec, the same as the top speed of a fling-scroll, and slower if necessary to make the animation take at least 300 ms.

So e.g. if you've scrolled up by 16000 logical pixels — roughly 20 screenfuls, since [160px has a nominal value of one inch](https://github.com/zulip/zulip-mobile/blob/main/docs/background/layout.md#display-size-logical-pixels) — then the scroll to the bottom will take (16000 px)/(8000 px/sec) = 2 seconds. If you've scrolled up by only 1200 px (roughly 1.5 screenfuls), then the scroll to the bottom will be slower, at 4000 px/sec, in order to stretch it over 300 ms.

### Bug

The logic to do that, though — introduced in 5abeb887e (#223) — didn't have any tests. As a result, when an unstated assumption it made got broken a few months later (in bdac26fb1), nothing alerted us to that. It's been broken since then.

The current behavior is instead that the animation is always set up to take just 300 ms, no matter how far it needs to travel. So for example if you're scrolled 16,000 px up (again ~20 screenfuls), then it will try to go a little over 53,000 px per second, which at 60 fps is 889 px/frame, i.e. likely more than a whole screenful per frame. (And in fact faster than that for part of the time: that's the average over the whole animation, but it uses [an easing curve](https://main-api.flutter.dev/flutter/animation/Curves/ease-constant.html), so it goes faster early on and slower later.)

Scrolling at such ludicrous speed looks very flickery at best, i.e. if the app does successfully render each of those frames on time. It's also quite a strain on performance: the app needs to build, lay out, and render an entire new screenful of messages each frame, and if going by more than a screenful it also builds and lays out all the unseen messages in between in order to know how tall they are. So the actual behavior I usually see is that a lot of frames get dropped, i.e. jank.

### Plan

I've spotted this behavior as a user a few times over the past year or so. I don't think we've heard anyone complain about it, though, and I think this button isn't central to the UX, so we haven't prioritized it. But I discovered the cause of the bug while reading the scroll-to-bottom button's code in the course of revising it to work at all in a world of #82, and I'll fix the bug as part of that work.

gnprice:
Fixes #1485.

This is the next round after #1436, toward #82. It fixes the "scroll to bottom" button in the message list so that it will work correctly when we start splitting the message list into back-to-back slivers.

Along the way it also fixes that button so that it works smoothly even in the current form of the message list — it fixes #1485, which I'd encountered occasionally as a user in the year or so since we apparently introduced it, and then discovered in the code in the course of writing these changes.

This work also produced two upstream PRs to the framework:
https://github.com/flutter/flutter/pull/166730
https://github.com/flutter/flutter/pull/166731
One of those is merged, and this PR uses it; the other is still pending, so there's a bit of code effectively copy-pasted from the framework in order to let us make the customization we need.

#### c32edc8b7 deps: Upgrade Flutter to 3.32.0-1.0.pre.257

In particular this pulls in a recent PR of mine so we can use it:
  https://github.com/flutter/flutter/pull/166730

#### 7309915b2 msglist: Fix speed calculations in scroll to bottom

Fixes #1485.

This logic for deciding how long the scroll-to-bottom animation
should take, introduced in 5abeb887e (#223), didn't have any tests.

As a result, when its unstated assumption about the message list --
that scrolling up from the end was represented by positive scroll
positions -- was broken a few months later (in bdac26fb1), nothing
alerted us to that.

I did notice a few times over the past year or so that the effect of
the scroll-to-bottom button seemed jerky, as if it were trying to
move much farther in each frame than it should.  Now I know why.

(Discovered this in the course of revisiting this code in order to
adapt it to the more radical change to the message list's scroll
positions which is coming up: "zero" won't be the end, but somewhere
in the middle.)

#### 3db7323dc msglist: Drop easing curve in scroll-to-bottom

Fundamentally an easing curve like this relies on knowing in advance
how far the animation is going to end up going.  When we start
letting the message list scroll from the middle of history, it'll no
longer be possible to know that.

Switch instead to a behavior that can be based on only what's happened
so far, not on a prediction of the future.  In the future, if we want,
we could get fancy and make the speed change over time; but to start,
keep it simple, and just move at the same speed from start to finish.

#### a0de83f68 scroll [nfc]: Inline animateTo implementation into scrollToEnd

This gets us hands-on control of the ScrollActivity being used,
which we'll want to customize in order to get the behavior we want.

#### e738a0852 scroll [nfc]: Introduce ScrollToEndActivity

We'll use this to customize the behavior.

It also makes a handy marker in tests, when using the testing-only
[ScrollPosition.activity] getter for inspecting what's going on in
the scroll behavior.

#### 5233d6d45 scroll: Avoid overscroll in "scroll to end"

As long as the bottom sliver is size zero (or more generally, as long
as maxScrollExtent does not change during the animation), this is
nearly NFC: I believe the only changes in behavior would come from
differences in rounding.

This change handles the case where the end turns out to be closer
than it looked at the beginning of the animation.  Before this
change, the animation would try to scroll past the end in that case.
Now it stops at exactly the end -- just like it already did in the
case where the end was known exactly in advance, as it currently
always is in the actual message list.

That case is a possibility as soon as there's a bottom sliver with a
message in it: scroll up so the message is offscreen and no longer
built; then have the message edited so it becomes shorter; then scroll
back down.  It's impossible for the viewport to know that the bottom
sliver's content has gotten taller until we actually scroll back down
and cause the message's widget to get built.

#### f09976aa1 scroll [nfc]: Introduce SimulationDrivenScrollActivity

In order to implement the "scroll to bottom" button in a way that
behaves well when scrolling to the growth end of a sliver -- in
particular, when scrolling to the end of the message list after we
split it into back-to-back slivers -- we'll want some differences
from the behavior provided by DrivenScrollActivity, which we've used
up until now (originally via the `animateTo` method).

That calls for our own ScrollActivity subclass, ScrollToEndActivity.
We'll want most of the same behavior as DrivenScrollActivity, with
just a couple of changes; but one of the places we want to change
isn't among the places that DrivenScrollActivity exposes for
subclassing.  So add this class, based on DrivenScrollActivity but
with a customization point added in the additional spot we'll need.

Originally there were two additional customization points needed.
After first drafting this change, I sent those upstream as two PRs:
 https://github.com/flutter/flutter/pull/166730
 https://github.com/flutter/flutter/pull/166731

The first one has already landed in DrivenScrollActivity: the
applyMoveTo method now overridden by ScrollToEndActivity in a
previous commit.  The other one, a `.simulation` constructor, is
pending.  A "TODO(upstream)" comment points to that PR, because
once it also merges we can dispense with this class.

#### ac2104029 scroll: Drive "scroll to end" through uncertainty about endpoint

As long as the bottom sliver is size zero (or more generally, as long
as maxScrollExtent does not change during the animation), this is
nearly NFC: I believe the only changes in behavior would come from
differences in rounding.

By describing the animation in terms of velocity, rather than a
duration and exact target position, this lets us smoothly handle the
case where we may not know exactly what the position coordinate of the
end will be.  A previous commit handled the case where the end comes
sooner than estimated, by promptly stopping when that happens.  This
commit ensures the scroll continues past the original estimate, in the
case where the end comes later.

That case is a possibility as soon as there's a bottom sliver with a
message in it: scroll up so the message is offscreen and no longer
built; then have the message edited so it becomes taller; then scroll
back down.  It's impossible for the viewport to know that the bottom
sliver's content has gotten taller until we actually scroll back down
and cause the message's widget to get built.

And naturally that will become even more salient of an issue when we
enable the message list to jump into the middle of a long history, so
that the bottom sliver may have content that hasn't yet been scrolled
to, has never been built as widgets, and may not even have yet been
fetched from the server.


chrisbobbe (commented):
Thanks! LGTM except I think it's the other way around with the upstream PRs:

```
Originally there were two additional customization points needed.
After first drafting this change, I sent those upstream as two PRs:
https://github.com/flutter/flutter/pull/166730
https://github.com/flutter/flutter/pull/166731

The first one has already landed in DrivenScrollActivity: the
applyMoveTo method now overridden by ScrollToEndActivity in a
previous commit. The other one, a .simulation constructor, is
pending. A "TODO(upstream)" comment points to that PR, because
once it also merges we can dispense with this class.
```

i.e. https://github.com/flutter/flutter/pull/166731 with the `applyMoveTo` method hasn't been merged yet, and the analyzer complains about the `@override` marker:

```
scroll: Avoid overscroll in "scroll to end"
```

```
$ flutter analyze
Analyzing zulip-flutter...                                              

warning • The method doesn't override an inherited method • lib/widgets/scrolling.dart:269:8 • override_on_non_overriding_member
  error • The method 'applyMoveTo' isn't defined in a superclass of 'ScrollToEndActivity' • lib/widgets/scrolling.dart:287:16 •
         undefined_super_member

2 issues found. (ran in 3.4s)
```

chrisbobbe:
Looks like some conflicts with the just-merged #1489.

gnprice:
Yep, resolved those and made the other updates — PTAL.

chrisbobbe:
LGTM, thanks! Merged.

PR #1491, opened 2025-04-30
Compute colorSwatchFor when subscription is null

PIG208:
Looking into the design requirement a bit more, we initially fall back to this color because we didn't have the logic to generate channel colors:
  https://chat.zulip.org/#narrow/channel/48-mobile/topic/All.20channels.20screen.20-.20.23F832/near/1904009

This replaces the ad-hoc colors with generated colors based on a fallback base color.

The base color we fall back to is still a placeholder.  Ideally, we should pick a color even for unsubscribed streams, as shown in the Figma design linked in #188.  That issue is out-of-scope for this, though.

For now, it's sufficient to just remove the ad-hoc color on MessageListTheme.  This way we don't have to reuse it in places where colors for unsubscribed streams are needed, namely the topic list (#1158).

| | before | after |
| - | - | - |
| light | light-before | light-after-3 |
| dark | dark-before | dark-after-3 |

### DIFF
--- a/lib/widgets/theme.dart
+++ b/lib/widgets/theme.dart
@@ -554,10 +554,19 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
   }
 }
 
+// This is taken from:
+//   https://github.com/zulip/zulip/blob/b248e2d93/web/src/stream_data.ts#L40
+const kDefaultChannelColorSwatchBaseColor = 0xffc2c2c2;
+
 /// The theme-appropriate [ChannelColorSwatch] based on [subscription.color].
 ///
+/// If [subscription] is null, [ChannelColorSwatch] will be based on
+/// [kDefaultChannelColorSwatchBaseColor].
+///
 /// For how this value is cached, see [ChannelColorSwatches.forBaseColor].
-ChannelColorSwatch colorSwatchFor(BuildContext context, Subscription subscription) {
+// TODO(#188) pick different colors for unsubscribed channels
+ChannelColorSwatch colorSwatchFor(BuildContext context, Subscription? subscription) {
   return DesignVariables.of(context)
-    .channelColorSwatches.forBaseColor(subscription.color);
+    .channelColorSwatches.forBaseColor(
+      subscription?.color ?? kDefaultChannelColorSwatchBaseColor);
 }
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -43,9 +43,6 @@ class MessageListTheme extends ThemeExtension<MessageListTheme> {
     unreadMarker: const HSLColor.fromAHSL(1, 227, 0.78, 0.59).toColor(),
 
     unreadMarkerGap: Colors.white.withValues(alpha: 0.6),
-
-    // TODO(design) this seems ad-hoc; is there a better color?
-    unsubscribedStreamRecipientHeaderBg: const Color(0xfff5f5f5),
   );
 
   static final dark = MessageListTheme._(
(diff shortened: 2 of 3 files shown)


chrisbobbe (commented):
Thanks, LGTM except for a commit-message nit:

```
The base color we fall back to is still a placeholder.  Ideally, we
should pick a color for the unsubscribed streams, as shown in the
Figma design linked in https://github.com/zulip/zulip-flutter/issues/188.  That issue is out-of-scope for this,
though.
```

I was getting confused by "Ideally, we should pick a color for the unsubscribed streams", because that's exactly what this PR does, isn't it :) it picks `kDefaultChannelColorSwatchBaseColor` i.e. `0xffc2c2c2`.

I think what's meant is that we should generate random colors (within constraints) separately for each unsubscribed channel, client-side, is that right?

(also nit: channels, not streams 🙂)

PIG208:
Thanks! Updated the PR.

chrisbobbe:
Thanks! Marking for Greg's review.

gnprice:
> we initially fall back to this color because we didn't have the logic to generate channel colors: https://chat.zulip.org/#narrow/channel/48-mobile/topic/All.20channels.20screen.20-.20.23F832/near/1904009

Hmm. That's from 2024, and we already had the ChannelColorSwatch system (was StreamColorSwatch) in 2023. So it's not because we didn't have these swatches that we didn't use it then.

Rather I think the point in that message:
> On web, we do show colors even for non-subscribed streams:
> https://chat.zulip.org/#channels/notsubscribed
> but I think generating those colors involves logic we don't currently have in mobile, so adding that would be out of scope here.

was that we haven't implemented the logic to make up a different base color for each non-subscribed channel. (Which I believe draws on the same logic that's used to pick a base color for your subscription if you do subscribe.)

---

The changes in the screenshots seem harmless, though. What's the motivation for this change — is it that in your work on #1158 you're finding you want a whole channel-color swatch for a non-subscribed channel, beyond `.barBackground`?

PIG208:
Initially it was just the alternative solution to moving the ad-hoc color to design variables; but it is helpful for providing `.iconOnBarBackground` too, for the colored channel icon on the app bar.

lib/widgets/theme.dart
```diff
@@ -554,10 +554,19 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
...
+const kDefaultChannelColorSwatchBaseColor = 0xffc2c2c2;
+
 /// The theme-appropriate [ChannelColorSwatch] based on [subscription.color].
 ///
+/// If [subscription] is null, [ChannelColorSwatch] will be based on
+/// [kDefaultChannelColorSwatchBaseColor].
+///
 /// For how this value is cached, see [ChannelColorSwatches.forBaseColor].
-ChannelColorSwatch colorSwatchFor(BuildContext context, Subscription subscription) {
+// TODO(#188) pick different colors for unsubscribed channels
```
gnprice, line 567: Looking into the design requirement a bit more, we initially fall back
    to this color because we didn't have the logic to generate channel
    colors:

As discussed at https://github.com/zulip/zulip-flutter/pull/1491#issuecomment-2845893203 , the thing we didn't have then is exactly the thing we don't have now either, the one this TODO points out :slightly_smiling_face:.

gnprice (commented):
Thanks! The changes look good.

Please revise the commit message to not be inaccurate, though :slightly_smiling_face: — the motivation you mention above is fine.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Merging.

PR #1489, opened 2025-04-28
Use now-upstream ScrollView.paintOrder, replacing CustomPaintOrderScrollView

gnprice:
#### 47273ce47 deps: Upgrade Flutter to 3.32.0-1.0.pre.257

And update Flutter's supporting libraries to match.

In particular this pulls in two recent PRs of mine so we can
use them:
  https://github.com/flutter/flutter/pull/166730
  https://github.com/flutter/flutter/pull/164818

#### 7b5ed74b8 scroll [nfc]: Use now-upstream ScrollView.paintOrder

The functionality provided by CustomPaintOrderScrollView is now
available upstream, since my PR adding it landed:
  https://github.com/flutter/flutter/pull/164818

#### 705db8c26 scroll [nfc]: Cut CustomPaintOrderScrollView now that upstream PR merged

This functionality is now covered by [ScrollView.paintOrder],
after my PR adding that feature upstream.

PR #1494, opened 2025-05-01
ios: MinimumOSVersion 12.0 to 14.0 in ios/Flutter/AppFrameworkInfo.plist

chrisbobbe:
The Flutter doc on deploying an iOS app says this number should match the deployment target we set in Xcode (which we also specify in ios/Podfile):
  https://docs.flutter.dev/deployment/ios#updating-the-apps-deployment-version
; sure, go ahead and update it.

Prompted by noticing that `flutter run` started bumping it to 13.0 (in flutter/flutter@09d4dabd6; see discussion:
  https://github.com/zulip/zulip-flutter/pull/1489#issuecomment-2843657156
) and realizing that 14.0 is actually what it should be for us.

### DIFF
--- a/ios/Flutter/AppFrameworkInfo.plist
+++ b/ios/Flutter/AppFrameworkInfo.plist
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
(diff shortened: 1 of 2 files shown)


gnprice:
Thanks! Looks good; merging.

Issue #1481, opened 2025-04-23
compose: Implement disabled appearance for edit-message "Save" button


chrisbobbe:
When we first implement the edit-message compose box (#126), we plan to skip a feature that the regular compose box has, because we don't have a Figma design for it.

The regular compose box gives the "Send" button a faded appearance when there are validation errors. Compare:

| Validation error (content too long) | No validation error |
| --- | --- |
| <img width="502" alt="Image" src="https://github.com/user-attachments/assets/89b5dcc1-8ee8-4cef-aa19-c27e7bc23421" /> | <img width="500" alt="Image" src="https://github.com/user-attachments/assets/d00acf67-a85f-4cf7-a902-08e330d8d49a" /> |

We should do the same with the "Save" button in the edit-message compose box:

<img width="502" alt="Image" src="https://github.com/user-attachments/assets/d9af0de5-519d-4534-b45c-fabc7c515bfc" />

See `ZulipWebUiKitButton` in the code, and the Figma frames linked from there, which don't include a design for a disabled state.

Tapping the "Save" button still gives a dialog explaining the validation error(s), just like the regular compose box's "Send" button, which mitigates the bug a lot.

Issue #1032, opened 2024-10-29
dialog: Styling for action-button text when the action is destructive


chrisbobbe:
We have a function for a confirmation / suggested-action dialog:

<details>

```dart
void showSuggestedActionDialog({
  required BuildContext context,
  required String title,
  required String message,
  required String? actionButtonText,
  required VoidCallback onActionButtonPress,
}) {
  final zulipLocalizations = ZulipLocalizations.of(context);
  showDialog<void>(
    context: context,
    builder: (BuildContext context) => AlertDialog(
      title: Text(title),
      content: SingleChildScrollView(child: Text(message)),
      actions: [
        TextButton(
          onPressed: () => Navigator.pop(context),
          child: _dialogActionText(zulipLocalizations.dialogCancel)),
        TextButton(
          onPressed: onActionButtonPress,
          child: _dialogActionText(actionButtonText ?? zulipLocalizations.dialogContinue)),
      ]));
}
```

</details>

It should grow a param `bool destructive = false,` and, when that's true, show the action button in a style that's appropriate for a destructive action.

On iOS (post-#996) this should be straightforward: just pass `isDestructiveAction: true` to [`CupertinoDialogAction`](https://api.flutter.dev/flutter/cupertino/CupertinoDialogAction/CupertinoDialogAction.html).

In Material 3 style for Android, it looks like there isn't actually a special "destructive"/"danger" style for destructive actions. See the ["Dialogs"](https://m3.material.io/components/dialogs/specs) M3 doc, which even includes a confirmation dialog for a "Delete" action, and the "Delete" button is styled just the same as the "Cancel" button. So, until/unless we get a design from Vlad that says otherwise, we'll just keep the styling we get by default, and give `showSuggestedActionDialog` a dartdoc that says the new param only has an effect on iOS.

Gaurav-Kushwaha-1225:
@chrisbobbe 
Please close this issue, if it's resolved.

Gaurav-Kushwaha-1225:
Hello @chrisbobbe,

Could you please assign this issue to me? I'm a first-time contributor and have already completed the project setup. 
I've also resolved this issue. 

Looking forward to your response!

Gaurav-Kushwaha-1225:
> Hello @chrisbobbe,
> 
> Could you please assign this issue to me? I'm a first-time contributor and have already completed the project setup. I've also resolved this issue.
> 
> Looking forward to your response!

Hii @chrisbobbe @gnprice,

[...]

PR #1498, opened 2025-05-03
Edit-message (8/n): Implement edit-message!

chrisbobbe:
Edit-message! 🎉

Fixes: #126

## Screenshots

| Light | Dark |
| --- | --- |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -144,6 +144,10 @@
   "@actionSheetOptionUnstarMessage":  {
     "description": "Label for unstar button on action sheet."
   },
+  "actionSheetOptionEditMessage": "Edit message",
+  "@actionSheetOptionEditMessage": {
+    "description": "Label for the 'Edit message' button in the message action sheet."
+  },
   "actionSheetOptionMarkTopicAsRead": "Mark topic as read",
   "@actionSheetOptionMarkTopicAsRead": {
     "description": "Option to mark a specific topic as read in the action sheet."
@@ -218,6 +222,10 @@
   "errorMessageNotSent": "Message not sent",
   "@errorMessageNotSent": {
     "description": "Error message for compose box when a message could not be sent."
+  },
+  "errorMessageEditNotSaved": "Message not saved",
+  "@errorMessageEditNotSaved": {
+    "description": "Error message for compose box when a message edit could not be saved."
   },
   "errorLoginCouldNotConnect": "Failed to connect to server:\n{url}",
   "@errorLoginCouldNotConnect": {
@@ -309,6 +317,10 @@
   "@errorUnstarMessageFailedTitle": {
     "description": "Error title when unstarring a message failed."
   },
+  "errorCouldNotEditMessageTitle": "Could not edit message",
+  "@errorCouldNotEditMessageTitle": {
+    "description": "Error title when an exception prevented us from opening the compose box for editing a message."
+  },
   "successLinkCopied": "Link copied",
   "@successLinkCopied": {
     "description": "Success message after copy link action completed."
@@ -328,6 +340,46 @@
   "errorBannerCannotPostInChannelLabel": "You do not have permission to post in this channel.",
   "@errorBannerCannotPostInChannelLabel": {
     "description": "Error-banner text replacing the compose box when you do not have permission to send a message to the channel."
+  },
+  "composeBoxBannerLabelEditMessage": "Edit message",
+  "@composeBoxBannerLabelEditMessage": {
+    "description": "Label text for the compose-box banner when you are editing a message."
+  },
+  "composeBoxBannerButtonCancel": "Cancel",
+  "@composeBoxBannerButtonCancel": {
+    "description": "Label text for the 'Cancel' button in the compose-box banner when you are editing a message."
+  },
+  "composeBoxBannerButtonSave": "Save",
+  "@composeBoxBannerButtonSave": {
+    "description": "Label text for the 'Save' button in the compose-box banner when you are editing a message."
+  },
+  "editAlreadyInProgressTitle": "Cannot edit message",
+  "@editAlreadyInProgressTitle": {
+    "description": "Error title when a message edit cannot be saved because there is another edit already in progress."
+  },
+  "editAlreadyInProgressMessage": "An edit is already in progress. Please wait for it to complete.",
+  "@editAlreadyInProgressMessage": {
+    "description": "Error message when a message edit cannot be saved because there is another edit already in progress."
+  },
+  "savingMessageEditLabel": "SAVING EDIT…",
+  "@savingMessageEditLabel": {
+    "description": "Text on a message in the message list saying that a message edit request is processing. (Use ALL CAPS for cased alphabets: Latin, Greek, Cyrillic, etc.)"
+  },
+  "savingMessageEditFailedLabel": "EDIT NOT SAVED",
+  "@savingMessageEditFailedLabel": {
+    "description": "Text on a message in the message list saying that a message edit request failed. (Use ALL CAPS for cased alphabets: Latin, Greek, Cyrillic, etc.)"
+  },
+  "discardDraftConfirmationDialogTitle": "Discard the message you’re writing?",
+  "@discardDraftConfirmationDialogTitle": {
+    "description": "Title for a confirmation dialog for discarding message text that was typed into the compose box."
+  },
+  "discardDraftConfirmationDialogMessage": "When you edit a message, the content that was previously in the compose box is discarded.",
+  "@discardDraftConfirmationDialogMessage": {
+    "description": "Message for a confirmation dialog for discarding message text that was typed into the compose box."
+  },
+  "discardDraftConfirmationDialogConfirmButton": "Discard",
+  "@discardDraftConfirmationDialogConfirmButton": {
+    "description": "Label for the 'Discard' button on a confirmation dialog for discarding message text that was typed into the compose box."
   },
   "composeBoxAttachFilesTooltip": "Attach files",
   "@composeBoxAttachFilesTooltip": {
@@ -366,6 +418,10 @@
     "placeholders": {
       "destination": {"type": "String", "example": "#channel name > topic name"}
     }
+  },
+  "preparingEditMessageContentInput": "Preparing…",
+  "@preparingEditMessageContentInput": {
+    "description": "Hint text for content input when the compose box is preparing to edit a message."
   },
   "composeBoxSendTooltip": "Send",
   "@composeBoxSendTooltip":  {
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -168,7 +168,7 @@
       "server": {"type": "String", "example": "https://example.com"}
     }
   },
-  "errorCouldNotFetchMessageSource": "Could not fetch message source",
+  "errorCouldNotFetchMessageSource": "Could not fetch message source.",
   "@errorCouldNotFetchMessageSource": {
     "description": "Error message when the source of a message could not be fetched."
   },
@@ -561,7 +561,7 @@
       "url": {"type": "String", "example": "http://chat.example.com/"}
     }
   },
-  "errorInvalidResponse": "The server sent an invalid response",
+  "errorInvalidResponse": "The server sent an invalid response.",
   "@errorInvalidResponse": {
     "description": "Error message when an API call returned an invalid response."
   },
@@ -591,7 +591,7 @@
       "httpStatus": {"type": "int", "example": "500"}
     }
   },
-  "errorVideoPlayerFailed": "Unable to play the video",
+  "errorVideoPlayerFailed": "Unable to play the video.",
   "@errorVideoPlayerFailed": {
     "description": "Error message when a video fails to play."
   },
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -12,6 +12,7 @@ import '../api/model/model.dart';
 import '../api/route/channels.dart';
 import '../api/route/messages.dart';
 import '../generated/l10n/zulip_localizations.dart';
+import '../model/binding.dart';
 import '../model/emoji.dart';
 import '../model/internal_link.dart';
 import '../model/narrow.dart';
@@ -576,6 +577,8 @@ void showMessageActionSheet({required BuildContext context, required Message mes
     CopyMessageTextButton(message: message, pageContext: pageContext),
     CopyMessageLinkButton(message: message, pageContext: pageContext),
     ShareButton(message: message, pageContext: pageContext),
+    if (_getShouldShowEditButton(pageContext, message))
+      EditButton(message: message, pageContext: pageContext),
   ];
 
   _showActionSheet(pageContext, optionButtons: optionButtons);
@@ -589,6 +592,36 @@ abstract class MessageActionSheetMenuItemButton extends ActionSheetMenuItemButto
   }) : assert(pageContext.findAncestorWidgetOfExactType<MessageListPage>() != null);
 
   final Message message;
+}
+
+bool _getShouldShowEditButton(BuildContext pageContext, Message message) {
+  final store = PerAccountStoreWidget.of(pageContext);
+
+  final messageListPage = MessageListPage.ancestorOf(pageContext);
+  final composeBoxState = messageListPage.composeBoxState;
+  final isComposeBoxOffered = composeBoxState != null;
+  final composeBoxController = composeBoxState?.controller;
+
+  final editMessageErrorStatus = store.getEditMessageErrorStatus(message.id);
+  final editMessageInProgress =
+    // The compose box is in edit-message mode, with Cancel/Save instead of Send.
+    composeBoxController is EditMessageComposeBoxController
+    // An edit request is in progress or the error state.
+    || editMessageErrorStatus != null;
+
+  final now = ZulipBinding.instance.utcNow().millisecondsSinceEpoch ~/ 1000;
+  final editLimit = store.realmMessageContentEditLimitSeconds;
+  final outsideEditLimit =
+    editLimit != null
+    && editLimit != 0 // TODO(server-6) remove (pre-FL 138, 0 represents no limit)
+    && now - message.timestamp > editLimit;
+
+  return message.senderId == store.selfUserId
+    && isComposeBoxOffered
+    && store.realmAllowMessageEditing
+    && !outsideEditLimit
+    && !editMessageInProgress
+    && message.poll == null; // messages with polls cannot be edited
 }
 
 class ReactionButtons extends StatelessWidget {
@@ -953,5 +986,24 @@ class ShareButton extends MessageActionSheetMenuItemButton {
       case ShareResultStatus.dismissed:
         // nothing to do
     }
+  }
+}
+
+class EditButton extends MessageActionSheetMenuItemButton {
+  EditButton({super.key, required super.message, required super.pageContext});
+
+  @override
+  IconData get icon => ZulipIcons.edit;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) =>
+    zulipLocalizations.actionSheetOptionEditMessage;
+
+  @override void onPressed() async {
+    final composeBoxState = findMessageListPage().composeBoxState;
+    if (composeBoxState == null) {
+      throw StateError('Compose box unexpectedly absent when edit-message button pressed');
+    }
+    composeBoxState.startEditInteraction(message.id);
   }
 }
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -493,11 +493,13 @@ class _ContentInput extends StatelessWidget {
     required this.narrow,
     required this.controller,
     required this.hintText,
+    this.enabled = true, // ignore: unused_element_parameter
   });
 
   final Narrow narrow;
   final ComposeBoxController controller;
   final String hintText;
+  final bool enabled;
 
   static double maxHeight(BuildContext context) {
     final clampingTextScaler = MediaQuery.textScalerOf(context)
@@ -540,6 +542,7 @@ class _ContentInput extends StatelessWidget {
             top: _verticalPadding, bottom: _verticalPadding,
             color: designVariables.composeBoxBg,
             child: TextField(
+              enabled: enabled,
               controller: controller.content,
               focusNode: controller.contentFocusNode,
               // Let the content show through the `contentPadding` so that
@@ -957,9 +960,10 @@ Future<void> _uploadFiles({
 }
 
 abstract class _AttachUploadsButton extends StatelessWidget {
-  const _AttachUploadsButton({required this.controller});
+  const _AttachUploadsButton({required this.controller, required this.enabled});
 
   final ComposeBoxController controller;
+  final bool enabled;
 
   IconData get icon;
   String tooltip(ZulipLocalizations zulipLocalizations);
@@ -1001,7 +1005,7 @@ abstract class _AttachUploadsButton extends StatelessWidget {
       child: IconButton(
         icon: Icon(icon, color: designVariables.foreground.withFadedAlpha(0.5)),
         tooltip: tooltip(zulipLocalizations),
-        onPressed: () => _handlePress(context)));
+        onPressed: enabled ? () => _handlePress(context) : null));
   }
 }
 
@@ -1060,7 +1064,7 @@ Future<Iterable<_File>> _getFilePickerFiles(BuildContext context, FileType type)
 }
 
 class _AttachFileButton extends _AttachUploadsButton {
-  const _AttachFileButton({required super.controller});
+  const _AttachFileButton({required super.controller, required super.enabled});
 
   @override
   IconData get icon => ZulipIcons.attach_file;
@@ -1076,7 +1080,7 @@ class _AttachFileButton extends _AttachUploadsButton {
 }
 
 class _AttachMediaButton extends _AttachUploadsButton {
-  const _AttachMediaButton({required super.controller});
+  const _AttachMediaButton({required super.controller, required super.enabled});
 
   @override
   IconData get icon => ZulipIcons.image;
@@ -1093,7 +1097,7 @@ class _AttachMediaButton extends _AttachUploadsButton {
 }
 
 class _AttachFromCameraButton extends _AttachUploadsButton {
-  const _AttachFromCameraButton({required super.controller});
+  const _AttachFromCameraButton({required super.controller, required super.enabled});
 
   @override
   IconData get icon => ZulipIcons.camera;
@@ -1370,6 +1374,7 @@ abstract class _ComposeBoxBody extends StatelessWidget {
 
   Widget? buildTopicInput();
   Widget buildContentInput();
+  bool getComposeButtonsEnabled(BuildContext context);
   Widget buildSendButton();
 
   @override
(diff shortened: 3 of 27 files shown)


chrisbobbe:
Oh, and if you tap "Edit message" in the action sheet but there's already text in the compose box, here's the dialog you get. Giving the "Discard" button a destructive/red style is future work, #1032.

| Light | Dark |
| --- | --- |
| image | image |

alya:
Did we discuss the strings anywhere? I would maybe write something like this:

> Discard the message you're writing?
> When you edit a message, the content that was previously in the compose box is discarded.
[ Cancel] [Discard content and continue]

alya:
I think it's good to put a `.` at the end of user-visible error strings.

chrisbobbe:
Makes sense; I started a discussion on CZO: [#mobile-team > Edit-message: UI strings @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Edit-message.3A.20UI.20strings/near/2168048)

chrisbobbe:
Revision pushed, with those UI string updates, and I've updated the screenshots.

chrisbobbe:
(Oops, fixed a failing test.)

PIG208 (commented):
Thanks for the PR! Looks pretty good to me overall. Left some comments.

lib/widgets/message_list.dart
```diff
@@ -1462,10 +1533,11 @@ class MessageWithPossibleSender extends StatelessWidget {
               Expanded(child: Column(
                 crossAxisAlignment: CrossAxisAlignment.stretch,
                 children: [
-                  MessageContent(message: message, content: item.content),
+                  content,
                   if ((message.reactions?.total ?? 0) > 0)
                     ReactionChipsList(messageId: message.id, reactions: message.reactions!),
-                  if (editStateText != null)
+                  if (editMessageErrorStatusRow != null) editMessageErrorStatusRow
```
gnprice, line 1539: I think this would go well as its own widget:
```suggestion
                  if (editMessageErrorStatus != null)
                    _EditMessageStatusRow(messageId: message.id),
```

That'd help keep this build method mostly clear of details that are specific to a message currently being edited.

lib/widgets/message_list.dart
```diff
@@ -1445,9 +1458,67 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+      };
+    }
+
+    Widget content = MessageContent(message: message, content: item.content);
+
+    if (editMessageErrorStatus != null) {
+      content = Opacity(opacity: 0.6, child: content);
+      switch (editMessageErrorStatus) {
+        case true:
+          content = _withRestoreEditMessageGestureDetector(context, content);
```
gnprice, line 1502: This also seems like a good candidate for its own widget:
```suggestion
          content = _RestoreEditMessageGestureDetector(messageId: item.message.id, child: content);
```

lib/widgets/message_list.dart
```diff
@@ -1445,9 +1458,67 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
-      onLongPress: () => showMessageActionSheet(context: context, message: message),
+      onLongPress: () {
+        final composeBoxState = MessageListPage.ancestorOf(context).composeBoxState;
+        final controller = composeBoxState?.controller;
+        final editMessageInProgress =
+          controller is EditMessageComposeBoxController
+          || editMessageErrorStatus != null;
```
gnprice, line 1515: Can this logic all move into `showMessageActionSheet`? Seems like between `context` and `message` it has all the inputs it needs in order to do the same calculations.

gnprice (commented):
Exciting! A few quick comments from an early skim.

lib/widgets/message_list.dart
```diff
@@ -1445,9 +1458,67 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+
+    Widget content = MessageContent(message: message, content: item.content);
+
+    if (editMessageErrorStatus != null) {
+      content = Opacity(opacity: 0.6, child: content);
+      switch (editMessageErrorStatus) {
+        case true:
+          content = _withRestoreEditMessageGestureDetector(context, content);
+        case false:
+          content = IgnorePointer(child: content);
```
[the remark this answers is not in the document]
chrisbobbe: It's meant to neutralize interactive message content like links, spoilers, etc.; I thought it made sense along with the faded 0.6-opacity appearance in the Figma. I can add a comment about that.

lib/widgets/action_sheet.dart
```diff
@@ -568,6 +574,18 @@ void showMessageActionSheet({required BuildContext context, required Message mes
   final markAsUnreadSupported = store.zulipFeatureLevel >= 155; // TODO(server-6)
   final showMarkAsUnreadButton = markAsUnreadSupported && isMessageRead;
 
+  final now = ZulipBinding.instance.utcNow().millisecondsSinceEpoch ~/ 1000;
+  final editLimit = store.realmMessageContentEditLimitSeconds;
+  final outsideEditLimit =
+    editLimit != null
+    && editLimit != 0 // TODO(server-6) remove (pre-FL 138, 0 represents no limit)
+    && now - message.timestamp > editLimit;
+  final showEditButton = message.senderId == store.selfUserId
```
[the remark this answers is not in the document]
chrisbobbe: Oh, TIL; thanks.

lib/widgets/message_list.dart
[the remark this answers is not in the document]
chrisbobbe: This condition means whether the compose box is in edit-message mode instead of the default write-new-message mode.

lib/widgets/action_sheet.dart
```diff
@@ -568,6 +574,18 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+    && messageListHasComposeBox
```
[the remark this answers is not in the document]
chrisbobbe: Oh indeed, I didn't see that we already have the value we need.

lib/widgets/message_list.dart
```diff
@@ -1445,9 +1458,67 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+            style: baseTextStyle.copyWith(color: designVariables.btnLabelAttLowIntDanger),
+            textAlign: TextAlign.end,
+            zulipLocalizations.savingMessageEditFailedLabel)),
+      };
+    }
+
+    Widget content = MessageContent(message: message, content: item.content);
+
+    if (editMessageErrorStatus != null) {
+      content = Opacity(opacity: 0.6, child: content);
```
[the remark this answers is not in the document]
chrisbobbe: Left a comment about this after asking Greg in the office:

```
      // The Figma also fades the sender row:
      //   https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2076574000
      // We've decided to just fade the message content because that's the only
      // thing that's changing.
```

test/widgets/action_sheet_test.dart
```diff
@@ -1410,6 +1418,174 @@ void main() {
       });
     });
 
+    group('EditButton', () {
+      group('present/absent appropriately', () {
+        /// Test whether the edit-message button is visible, given params.
+        ///
+        /// The message timestamp is 60s before the current time
+        /// ([TestBinding.utcNow]) as of the start of the test run.
```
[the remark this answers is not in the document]
chrisbobbe: Indeed—also, thanks for creating this :)

test/widgets/action_sheet_test.dart
```diff
@@ -1410,6 +1418,174 @@ void main() {
...
+              timestamp: eg.utcTimestamp(testBinding.utcNow()) - 60);
+
+            await setupToMessageActionSheet(tester,
+              message: message,
+              narrow: narrow ?? TopicNarrow.ofMessage(message),
+              realmAllowMessageEditing: allowed,
+              realmMessageContentEditLimitSeconds: limit,
+            );
+
+            if (!composing && errorStatus == null) {
```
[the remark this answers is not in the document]
chrisbobbe: I've added some comments to try to clarify.

chrisbobbe:
Thanks for the reviews! Revision pushed, for Zixuan's review.

PIG208:
Looks like there are some compilation errors in lib/widgets/message_list.dart, probably from some non-local merge conflicts.

test/widgets/action_sheet_test.dart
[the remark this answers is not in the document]
PIG208: Glad that helps!

test/widgets/action_sheet_test.dart
```diff
@@ -1410,6 +1429,196 @@ void main() {
...
+          bool allowed = true,
+          int? limit,
+          bool composing = false,
+          bool? errorStatus,
+          bool poll = false,
+        }) {
+          // It's impossible for the compose box to be in editing mode
+          // while, for the same message, the message-edit request is either
+          // in progress or in the error state.
+          assert(!composing || errorStatus == null);
```
PIG208, line 1462: Oh I see! Yeah this clears up the meaning of `composing`.

I think it is still possible if you initiate the edit in both the channel narrow and the topic narrow containing it, then save the edit in the topic narrow, navigate back and save in the channel narrow, on the same message. If the first request takes long enough to finish. But that's quite an edge case, and is guarded by a containsKey check already.

PIG208:
Thanks for the update! This looks good to me other than the compilation errors.

lib/widgets/compose_box.dart
```diff
@@ -1484,6 +1552,60 @@ class _ErrorBanner extends _Banner {
...
+    final zulipLocalizations = ZulipLocalizations.of(context);
+    return Row(mainAxisSize: MainAxisSize.min, spacing: 8, children: [
+      ZulipWebUiKitButton(label: zulipLocalizations.composeBoxBannerButtonCancel,
+        onPressed: () {
+          composeBoxState.endEditInteraction();
+        }),
+      // TODO(#1481) disabled appearance when there are validation errors
+      ZulipWebUiKitButton(label: zulipLocalizations.composeBoxBannerButtonSave,
+        attention: ZulipWebUiKitButtonAttention.high,
+        onPressed: () {
```
gnprice, line 1584: nit: this callback is about 20 lines; once it's more than a small handful of lines, I think it's a bit easier to read as a separate named function, so here a private method on this class.

lib/widgets/message_list.dart
```diff
@@ -1446,6 +1446,26 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+
+    final editMessageErrorStatus = store.getEditMessageErrorStatus(message.id);
+    if (editMessageErrorStatus != null) {
+      // The Figma also fades the sender row:
+      //   https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2076574000
+      // We've decided to just fade the message content because that's the only
+      // thing that's changing.
+      content = Opacity(opacity: 0.6, child: content);
+      switch (editMessageErrorStatus) {
+        case false:
```
gnprice, line 1459: nit: How about `if (!editMessageErrorStatus)`?

That'll have basically the same exhaustiveness-checking effect, too — `if` requires a bool and will reject a `bool?`.

lib/widgets/message_list.dart
```diff
@@ -1481,3 +1503,77 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+        spacing: 1.5,
+        children: [
+          Text(
+            style: baseTextStyle
+              .copyWith(color: designVariables.btnLabelAttLowIntInfo),
+            textAlign: TextAlign.end,
+            zulipLocalizations.savingMessageEditLabel),
+          // TODO instead place within outer padding; see Figma:
+          //   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4026-8775&m=dev
+          LinearProgressIndicator(
```
gnprice, line 1540: I don't quite follow this comment. Which padding should this be placed within?

lib/widgets/message_list.dart
```diff
@@ -1481,3 +1503,77 @@ class MessageWithPossibleSender extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      behavior: HitTestBehavior.opaque,
+      onTap: () {
+        final store = PerAccountStoreWidget.of(context);
+        final composeBoxState = MessageListPage.ancestorOf(context).composeBoxState;
+        if (composeBoxState == null) return;
+        composeBoxState.startEditInteraction(messageId: messageId,
+          originalRawContent: store.takeFailedMessageEdit(messageId));
```
gnprice, line 1575: Hmm. This `originalRawContent` value is the right value to initialize the text input to — but it's not the right value to use when sending the resulting edit request to the server. For that, we want the *original* original raw content, because that's the value the user is (implicitly) expecting the server will have.

Probably the right fix is to add another field on `_EditMessageRequestStatus`.

test/widgets/action_sheet_test.dart
chrisbobbe: Oh, good find! I can address this edge case in my next revision.

assets/l10n/app_en.arb
```diff
@@ -367,6 +411,10 @@
       "destination": {"type": "String", "example": "#channel name > topic name"}
     }
   },
+  "composeBoxEditMessageHint": "Message content",
+  "@composeBoxEditMessageHint": {
+    "description": "Hint text for content input when editing a message."
```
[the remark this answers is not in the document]
chrisbobbe: Sure; going with no hint text.

lib/widgets/message_list.dart
chrisbobbe: Here are screenshots of the Figma with and without that outer padding highlighted:

image

image

The progress indicator sits inside that outer padding.

I think it looks fine as-is and would be OK to remove the TODO, revisiting if needed in the future.

lib/widgets/message_list.dart
chrisbobbe: I see; yeah, thanks for catching this.

lib/widgets/message_list.dart
gnprice: Hmm, those screenshots don't really help. :slightly_smiling_face: How does this version differ from what's in Figma?

Looking into it a bit more, I think you're referring to the 4px of bottom padding on the MessageWithPossibleSender.

So in the Figma design, the indicator encroaches on that padding and gets closer to whatever's below (like the next message, or a date separator). In this version, when the indicator appears, it instead makes the message taller and stays 4px away from the sibling below.

I see the appeal of the version in Figma, but the difference is subtle. Happy to leave it as a TODO.

lib/widgets/message_list.dart
chrisbobbe: I've had some more thoughts prompted by this—will share later today

chrisbobbe:

As discussed in the office on Friday, this makes some improvements over the last revision—but omits the refetch-raw-content request on restoring a failed edit, which I'd been considering. 🙂 (Omits it because the edge case it addresses is minuscule, and it adds time and risk of failure in bad network connections.)

In particular:
- Now if you tap to restore a failed edit but there's already text in the compose box, and you cancel the "Discard…?" confirmation dialog, the failed edit hasn't been consumed.
- The edit-message compose box has a disabled "Preparing…" state for when the fetch-raw-content request is in progress (which happens just when you start an edit from the action sheet).

Here are screenshots of that "Preparing…" state, with the normal state for comparison:

| Dark | Light |
| --- | --- |
| image | image |
| image | image |

Giving the "Save" button a disabled appearance, in this state, is #1481; we don't have a design for that.

lib/widgets/compose_box.dart
```diff
@@ -1515,12 +1664,141 @@ class ComposeBox extends StatefulWidget {
...
+      return;
+    }
+    // TODO scroll message list to ensure the message is still in view;
+    //   highlight it?
+    setState(() {
+      assert(_controller is EditMessageComposeBoxController);
+      (_controller as EditMessageComposeBoxController)
+        ..originalRawContent = fetchedRawContent
+        ..content.value = TextEditingValue(text: fetchedRawContent)
+        ..contentFocusNode.requestFocus();
```
gnprice, line 1786: This step doesn't seem to work for me empirically (on my Android 15 phone). When I hit "Edit message" on the message action sheet, the input doesn't get focused — I have to tap it to focus (and to cause the keyboard to appear).

OTOH the corresponding step in the `_byRestoringFailedEdit` case *does* work.

Do you reproduce?

This is a perfectly acceptable glitch for launch: it just means an extra tap, and I think users are accustomed to having to sometimes tap an input field to get the keyboard to open when it seems like it should happen automatically. So let's only spend a small amount of effort now on diagnosing or fixing, and if that isn't enough then just file a short issue to record the point for post-launch.

lib/widgets/compose_box.dart
```diff
@@ -1515,12 +1664,141 @@ class ComposeBox extends StatefulWidget {
...
+      case false:
+        // This can happen if you start an edit interaction on one
+        // MessageListPage and then do an edit on a different MessageListPage,
+        // and the second edit is still saving when you return to the first.
+        //
+        // Abort rather than sending a request with a prevContentSha256
+        // that the server might not accept, and don't clear the compose
+        // box, so the user can try again after the request settles.
+        // TODO write a test for this
+        showErrorDialog(context: context,
```
gnprice, line 1711: Yeah, I'm OK with skipping a test for this case.

gnprice (commented):
Thanks for the revision! This behavior all now looks good, except one glitch mentioned below which I'm fine with for launch.

I haven't yet read the tests in the last/main commit.

lib/widgets/compose_box.dart
chrisbobbe: Debugged and fixed! 🎉

chrisbobbe:

lib/widgets/compose_box.dart
```diff
@@ -1675,12 +1825,146 @@ class ComposeBox extends StatefulWidget {
...
+  void startEditInteraction(int messageId) async {
+    if (await _abortBecauseContentInputNotEmpty()) return;
+    if (!mounted) return;
+
+    final store = PerAccountStoreWidget.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    switch (store.getEditMessageErrorStatus(messageId)) {
+      case null:
+        _editFromRawContentFetch(messageId);
```
gnprice, line 1860: bump https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2092232833: put the two non-error cases next to each other

lib/widgets/compose_box.dart
gnprice: (FTR, the fix was the `addPostFrameCallback` here, for the reason its comment explains:
```dart
    setState(() {
      // setState to refresh the input, upload buttons, etc.
      // out of the disabled "Preparing…" state.
      editMessageController.originalRawContent = fetchedRawContent;
    });
    editMessageController.content.value = TextEditingValue(text: fetchedRawContent);
    SchedulerBinding.instance.addPostFrameCallback((_) {
      // post-frame callback so this happens after the input is enabled
      editMessageController.contentFocusNode.requestFocus();
    });
```
)

gnprice (commented):
Thanks! Glad that glitch turned out to have a nice clean fix.

All looks good now except one nit — oh and I should read the tests :slightly_smiling_face:

test/widgets/compose_box_test.dart
```diff
@@ -1354,4 +1400,462 @@ void main() {
...
+        checkNotInEditingMode(tester, narrow: narrow);
+      });
+    }
+    doTestInterruptComposingFromActionSheet(narrow: channelNarrow);
+    doTestInterruptComposingFromActionSheet(narrow: topicNarrow);
+    doTestInterruptComposingFromActionSheet(narrow: dmNarrow);
+
+    // Test the "Discard…?" confirmation dialog when there's text in the compose
+    // box for a new message, but you've said you want to edit a message.
+    void doTestInterruptComposingFromFailedEdit({required Narrow narrow}) {
```
gnprice, line 1666: This comment doesn't seem to match the name. (The comment seems to say the same thing as the comment on doTestInterruptComposingFromActionSheet above.)

test/widgets/compose_box_test.dart
```diff
@@ -1354,4 +1400,462 @@ void main() {
...
+        checkNotInEditingMode(tester, narrow: narrow);
+
+        // …or send a new message.
+        connection.prepare(json: {}); // for typing-start request
+        connection.prepare(json: {}); // for typing-stop request
+        await enterContent(tester, 'new message to send');
+        state.controller.contentFocusNode.unfocus();
+        await tester.pump();
+        check(connection.takeRequests()).deepEquals([
+          (Subject<Object?> it) => it.isA<http.Request>()
```
gnprice, line 1814: nit: this sort of `deepEquals` argument is a bit shorter using `Condition`:
```suggestion
        check(connection.takeRequests()).deepEquals(<Condition<Object?>>[
          (it) => it.isA<http.Request>()
```

test/widgets/compose_box_test.dart
```diff
@@ -1354,4 +1400,462 @@ void main() {
...
+
+        // Save; check that the request is made and the compose box resets.
+        connection.prepare(json: UpdateMessageResult().toJson());
+        await tester.tap(find.widgetWithText(ZulipWebUiKitButton, 'Save'));
+        checkRequest(messageId, prevContent: 'foo', content: 'baz');
+        await tester.pump(Duration.zero);
+        checkNotInEditingMode(tester, narrow: narrow);
+      });
+    }
+    doTestInterruptComposingFromFailedEdit(narrow: channelNarrow);
```
gnprice, line 1714: nit: for helpers that define a test case like `test` and `testWidgets` do, name them `testFoo` to match those:
```suggestion
    testInterruptComposingFromFailedEdit(narrow: channelNarrow);
```

We're mostly consistent about that practice already; for examples, see `git grep -P '\btest[^W]\w+\('`. (Or refine the pattern to possibly find more examples; it's excluding `testWidgets` but crudely so.)

gnprice (commented):
OK, and here's a review of the remaining portion: the tests. They look good! Only small comments below.

These are definitely thorough. :sweat_smile: It's good to know that we can write such thorough tests of complex UI flows when we need to, and that we can do so while keeping them readable as these tests are. In general, though, I think our balance should be more in the direction of completing features faster, by spending less effort on comprehensive tests, than this represents.

One other aspect of these tests is that they have a noticeable runtime cost when running this test file. Here's timings on my office desktop (which is somewhat less speedy than my home desktop) with and without the `edit message` test group (knocking it out by inserting a `return;`):
```
$ time flutter test test/widgets/compose_box_test.dart 
00:17 +116: All tests passed!                                                              

time: 19.781s wall (25.325s u, 3.213s s)


$ time flutter test test/widgets/compose_box_test.dart 
00:10 +92: All tests passed!                                                               

time: 13.379s wall (16.966s u, 2.759s s)
```

So the 24 new test cases add about 6 seconds, or 50%, to the time to run the test file. Hardly a catastrophe, but enough to be a noticeable drag in the future when working on unrelated tests in this file (or unrelated changes to the compose box) and so repeatedly running this file.

I think one contributor to that runtime is that there's the beginnings of a combinatorial explosion: each test scenario gets run under three different narrows, including for interactions that seem unlikely to behave differently depending on the narrow.

Are there any of these where you could delete the `doTestFoo` calls for some of the narrows and it'd feel like all the same codepaths are effectively tested? If so, it'd be worth doing that.

If not, then the next point might be too much change to be worth doing before merging this, but would be good to have in mind for future tests: some of the additional checks like these, within some of the `doTestFoo` functions:
```dart
        // Now that we have the raw content, check the input is interactive
        // but no typing notifications are sent…
        // […]
        await enterContent(tester, 'some new content');
        // […]

        // …and the upload buttons work.
        // […]
        await tester.tap(find.byIcon(ZulipIcons.attach_file), warnIfMissed: false);
```
seem like they'd be just as effective if they were run for only one narrow, and only one value of other parameters like `start`. Separating them out is basically a form of making test cases more focused; and among the benefits of that sort of separation is that it can help a cluster of test cases end up doing N + M work, instead of N * M, while still effectively covering all the logic.

lib/widgets/compose_box.dart
chrisbobbe: Ah! I put `null` and `false` together as the "two non-error cases", thinking of "error" as it's used in `getEditMessageErrorStatus` which produces this value:

```dart
  @override
  bool? getEditMessageErrorStatus(int messageId) =>
    _editMessageRequests[messageId]?.hasError;
```

But I see now what you mean, that the error case is the one where we show an error dialog. 😛

test/widgets/action_sheet_test.dart
```diff
@@ -1421,6 +1440,174 @@ void main() {
...
+              // We're testing the request-in-progress state. Prepare a delay
+              // and tap Save.
+              connection.prepare(apiException: eg.apiBadRequest(), delay: Duration(seconds: 1));
+              await tester.tap(find.widgetWithText(ZulipWebUiKitButton, 'Save'));
+              await tester.pump(Duration.zero);
+            } else if (errorStatus == false) {
+              // We're testing the request-failed state. Prepare a failure
+              // and tap Save.
+              connection.prepare(
+                json: UpdateMessageResult().toJson(), delay: Duration(seconds: 1));
```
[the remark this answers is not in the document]
chrisbobbe: Eep, thanks!

gnprice:
Thanks for the revision! Merging, with a couple of tweaks in those tests:
```diff
--- test/widgets/compose_box_test.dart
+++ test/widgets/compose_box_test.dart
@@ -1538,3 +1538,3 @@ void main() {
 
-    void smokeTest({required Narrow narrow, required _EditInteractionStart start}) {
+    void testSmoke({required Narrow narrow, required _EditInteractionStart start}) {
       testWidgets('smoke: $narrow, ${start.message()}', (tester) async {
@@ -1579,2 +1579,3 @@ void main() {
         // TODO could also check that quote-and-reply and autocomplete work
+        //   (but as their own test cases, for a single narrow and start)
 
@@ -1589,8 +1590,8 @@ void main() {
     }
-    smokeTest(narrow: channelNarrow, start: _EditInteractionStart.actionSheet);
-    smokeTest(narrow: topicNarrow,   start: _EditInteractionStart.actionSheet);
-    smokeTest(narrow: dmNarrow,      start: _EditInteractionStart.actionSheet);
-    smokeTest(narrow: channelNarrow, start: _EditInteractionStart.restoreFailedEdit);
-    smokeTest(narrow: topicNarrow,   start: _EditInteractionStart.restoreFailedEdit);
-    smokeTest(narrow: dmNarrow,      start: _EditInteractionStart.restoreFailedEdit);
+    testSmoke(narrow: channelNarrow, start: _EditInteractionStart.actionSheet);
+    testSmoke(narrow: topicNarrow,   start: _EditInteractionStart.actionSheet);
+    testSmoke(narrow: dmNarrow,      start: _EditInteractionStart.actionSheet);
+    testSmoke(narrow: channelNarrow, start: _EditInteractionStart.restoreFailedEdit);
+    testSmoke(narrow: topicNarrow,   start: _EditInteractionStart.restoreFailedEdit);
+    testSmoke(narrow: dmNarrow,      start: _EditInteractionStart.restoreFailedEdit);
 
```

So being complete about that pattern from https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2093814509; and making sure that TODO doesn't point someone in the wrong direction with respect to https://github.com/zulip/zulip-flutter/pull/1498#issuecomment-2892069490 .

Issue #1495, opened 2025-05-02
msglist: Adding 🙂 emoji reaction fails


chrisbobbe:
Reproduced on CZO just now:

1. Long-press a message
2. Tap the 🙂 emoji in the list of emoji at the top of the message action sheet
3. See error dialog (unexpected):

<img width="559" alt="Image" src="https://github.com/user-attachments/assets/b472bea4-6348-42c3-bf8c-26ef212213dc" />

-----

```
"zulip_version":"10.0-550-g2182f40bd5"
```

```
"zulip_feature_level":380
```

gnprice:
Yeah, we should fix this.

Presumably this is due to the recent server change to make U+1F642 🙂 be called `:slight_smile:` instead of `:smile:`: 
* https://github.com/zulip/zulip/issues/32712
* https://github.com/zulip/zulip/pull/34177

The fix will probably involve updating this code:
```dart
  static final _popularCandidates = _generatePopularCandidates();

  static List<EmojiCandidate> _generatePopularCandidates() {
    // …
    return [
      // This list should match web:
      //   https://github.com/zulip/zulip/blob/83a121c7e/web/shared/src/typeahead.ts#L22-L29
      candidate('1f44d', '👍', ['+1', 'thumbs_up', 'like']),
      candidate('1f389', '🎉', ['tada']),
      candidate('1f642', '🙂', ['smile']),
```

The first step will be some refactoring to make that no longer static — it'll need to depend on `zulipFeatureLevel`.

PR #1506, opened 2025-05-08
emoji: Generate popular candidates using names from server data

chrisbobbe:
The server change in zulip/zulip#34177, renaming `:smile:` to
`:slight_smile:`, broke the corresponding reaction button in the
message action sheet. We've been sending the add/remove-reaction
request with the old name 'smile', which modern servers reject.

To fix, take the popular emoji names from ServerEmojiData, so that
we'll use the correct name on servers before and after the change.

API-design discussion:
  https://chat.zulip.org/#narrow/channel/378-api-design/topic/.23F1495.20smile.2Fslight_smile.20change.20broke.20reaction.20button/near/2170354

Fixes: #1495

### DIFF
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -698,11 +698,12 @@ class ReactionButtons extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    assert(EmojiStore.popularEmojiCandidates.every(
+    final store = PerAccountStoreWidget.of(pageContext);
+    final popularEmojiCandidates = store.popularEmojiCandidates;
+    assert(popularEmojiCandidates.every(
       (emoji) => emoji.emojiType == ReactionType.unicodeEmoji));
 
     final zulipLocalizations = ZulipLocalizations.of(context);
-    final store = PerAccountStoreWidget.of(pageContext);
     final designVariables = DesignVariables.of(context);
 
     bool hasSelfVote(EmojiCandidate emoji) {
@@ -718,7 +719,7 @@ class ReactionButtons extends StatelessWidget {
         color: designVariables.contextMenuItemBg.withFadedAlpha(0.12)),
       child: Row(children: [
         Flexible(child: Row(spacing: 1, children: List.unmodifiable(
-          EmojiStore.popularEmojiCandidates.mapIndexed((index, emoji) =>
+          popularEmojiCandidates.mapIndexed((index, emoji) =>
             _buildButton(
               context: context,
               emoji: emoji,
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -699,7 +699,7 @@ class ReactionButtons extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final store = PerAccountStoreWidget.of(pageContext);
-    final popularEmojiCandidates = store.popularEmojiCandidates;
+    final popularEmojiCandidates = store.popularEmojiCandidates();
     assert(popularEmojiCandidates.every(
       (emoji) => emoji.emojiType == ReactionType.unicodeEmoji));
 
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -555,6 +555,8 @@ void showMessageActionSheet({required BuildContext context, required Message mes
   final pageContext = PageRoot.contextOf(context);
   final store = PerAccountStoreWidget.of(pageContext);
 
+  final popularEmojiLoaded = store.popularEmojiCandidates().isNotEmpty;
+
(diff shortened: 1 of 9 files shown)


test/widgets/action_sheet_test.dart
```diff
@@ -811,72 +815,87 @@ void main() {
...
+              ..bodyFields.deepEquals({
+                  'reaction_type': 'unicode_emoji',
+                  'emoji_code': emoji.emojiCode,
+                  'emoji_name': emoji.emojiName,
+                });
+          });
 
-      for (final emoji in popularCandidates) {
-        final emojiDisplay = emoji.emojiDisplay as UnicodeEmojiDisplay;
+          testWidgets('${emoji.emojiName} removing success', (tester) async {
```
PIG208, line 855: nit: maybe add a suffix with the value of `useLegacy` like the previous test

lib/widgets/action_sheet.dart
```diff
@@ -569,7 +571,8 @@ void showMessageActionSheet({required BuildContext context, required Message mes
   final showMarkAsUnreadButton = markAsUnreadSupported && isMessageRead;
 
   final optionButtons = [
-    ReactionButtons(message: message, pageContext: pageContext),
+    if (popularEmojiLoaded)
```
PIG208, line 574: I think we can have a test checking when this should not be offered, like we do with quote-and-reply button.

PIG208 (commented):
Thanks! I went through the PR and the fix looks good to me overall. Left some comments mainly about testing.

test/model/emoji_test.dart
```diff
@@ -118,13 +121,17 @@ void main() {
...
-        store.setServerEmojiData(ServerEmojiData(codeToNames: unicodeEmoji));
-      }
+
+      final extraEmojiData = ServerEmojiData(codeToNames: unicodeEmoji ?? {});
+      final emojiData = addServerDataForPopular
+        ? eg.serverEmojiDataPopularPlus(extraEmojiData)
+        : extraEmojiData;
+      store.setServerEmojiData(emojiData);
       return store;
     }
```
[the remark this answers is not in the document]
chrisbobbe: I think that test should just be deleted, because it's intended that the popular-emoji candidates do not appear when there is no server emoji data.

chrisbobbe:

PIG208:
Looks good to me. Thanks! Marking this for Greg's review.

test/model/emoji_test.dart
```diff
@@ -10,12 +10,28 @@ import 'package:zulip/model/store.dart';
...
+    Map<String, List<String>>? unicodeEmoji,
+  }) {
+    final store = eg.store(
+      initialSnapshot: eg.initialSnapshot(realmEmoji: realmEmoji));
+
+    final extraEmojiData = ServerEmojiData(codeToNames: unicodeEmoji ?? {});
+    final emojiData = addServerDataForPopular
+      ? eg.serverEmojiDataPopularPlus(extraEmojiData)
+      : extraEmojiData;
+    store.setServerEmojiData(emojiData);
```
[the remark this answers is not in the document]
gnprice: Oh and actually the change in default (with `addServerDataForPopular = true`) means it affects at least this test:
```dart
    test('updates on setServerEmojiData', () {
      final store = prepare();
      check(store.allEmojiCandidates()).deepEquals([
        ...arePopularCandidates,
        isZulipCandidate(),
      ]);

      store.setServerEmojiData(eg.serverEmojiDataPopularPlus(
        ServerEmojiData(codeToNames: {
          '1f516': ['bookmark'],
        })));
      check(store.allEmojiCandidates()).deepEquals([
        ...arePopularCandidates,
        isUnicodeCandidate('1f516', ['bookmark']),
        isZulipCandidate(),
      ]);
```

In the app, `setServerEmojiData` gets called only once for a given store. So the scenario that's important to test here is that `allEmojiCandidates` gets updated when it's called before the first (and only) `setServerEmojiData` call and then after that call. This test therefore needs `prepare` to not call `setServerEmojiData`.

gnprice (commented):
Thanks for taking care of this! Just a few comments below.

chrisbobbe:

test/model/emoji_test.dart
```diff
@@ -246,15 +248,16 @@ void main() {
     });
 
     test('updates on setServerEmojiData', () {
-      final store = prepare();
+      final store = prepare(unicodeEmoji: null, addServerDataForPopular: false);
+      check(store.debugServerEmojiData).isNull();
```
gnprice, line 252: Ah, good thought checking this is null (following up on https://github.com/zulip/zulip-flutter/pull/1506#discussion_r2093758975 ).

gnprice (commented):
Thanks! This all looks good; merging.

PR #1507, opened 2025-05-08
anchors 8/n: Let new messages accumulate in bottom sliver

gnprice:
This is the next round after #1468.

After this PR, not only does the last message from the initial fetch go into the bottom sliver, but it remains there even while new messages arrive as events, and those new messages go into the bottom sliver below it.

The main motivation for this change is that it brings us closer to having a `fetchNewer` method, and therefore to being able to have the message list start out in the middle of history, #82. Along the way it also gives us the "90% solution" described in #83, so that new messages don't cause everything on the screen to jump up when you're already scrolled up in history.

#### f4b6e41fb msglist test: Ensure later errors get reported in full too

Otherwise, if several different test cases in this file fail due to
checks failing inside checkInvariants, then only the first one gets
reported in detail with the comparison and the stack trace.

#### 48983a347 msglist [nfc]: Move sliver boundary into the view-model

This will allow the model to maintain it over time as newer messages
arrive or get fetched.

#### e21f6012f msglist: Let new messages accumulate in bottom sliver

This is NFC for the behavior at initial fetch.  But thereafter, with
this change, messages never move between slivers, and new messages
go into the bottom sliver.

I believe the main user-visible consequence of this change is that if
the user is scrolled up in history and then a new message comes in,
the new message will no longer cause all the messages to shift upward.
This is the "90% solution" to #83.

On the other hand, if the user is scrolled all the way to the end,
then they still remain that way when a new message comes in --
there's specific logic to ensure that in MessageListScrollPosition,
and an existing test in test/widgets/message_list_test.dart verifies
it end to end.

The main motivation for this change is that it brings us closer to
having a `fetchNewer` method, and therefore to being able to have the
message list start out in the middle of history.

This change also allows us to revert a portion of fca651bf5, where
a test had had to be weakened slightly because messages began to
get moved between slivers.

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -163,7 +163,7 @@ mixin _MessageSequence {
   /// Either the bottom slices of both [items] and [messages] are empty,
   /// or the first item in the bottom slice of [items] is a [MessageListMessageItem]
   /// for the first message in the bottom slice of [messages].
-  int get middleItem => items.isEmpty ? 0 : items.length - 1;
+  int middleItem = 0;
 
   int _findMessageWithId(int messageId) {
     return binarySearchByKey(messages, messageId,
@@ -295,6 +295,7 @@ mixin _MessageSequence {
     _fetchOlderCooldownBackoffMachine = null;
     contents.clear();
     items.clear();
+    middleItem = 0;
   }
 
   /// Redo all computations from scratch, based on [messages].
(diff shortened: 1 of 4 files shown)


gnprice:
Just rebased past the merge of #1512. That also slightly simplifies one of these commits:
489f48a71 msglist [nfc]: Maintain middleItem as a field

compared with its previous revision 09758b23f.

chrisbobbe:
> Just rebased past the merge of #1512. That also slightly simplifies one of these commits:
> [489f48a](https://github.com/zulip/zulip-flutter/commit/489f48a71b0c88847a8c014536fcf46100d48b31) msglist [nfc]: Maintain middleItem as a field


Thanks! Could you also fix an earlier commit so it passes `flutter analyze`:

7401fd743 msglist: Always place the sliver boundary at a message

gnprice:
Ah indeed, thanks — in fact it turns out #1512 lets us simplify that commit away entirely.

chrisbobbe:
Thanks, LGTM! Merged!

Issue #1499, opened 2025-05-06
Keep track of topics in ChannelStore


PIG208:
This is follow-up to #1158.

The initial implementation of the topic-list will fetch topics with API calls. Using the data we have, by listening to changes to `PerAccountStore` and `Unreads`, we only know about the topic visibility and unreads changes. But not all of its data is up-to-date.

We need to make sure that the data store is specialized for managing what we know about topics, by channels. Some examples of events that need to be handled:

- message events, for new topics;
- update-message events, for moved or (un)resolved topics.

Since the topic-list page orders the topic by the last message sent, the data structure should keep track of the message IDs in each topic.

Ideally, as a follow-up to this, the data store will also keep track of whether previous requests to fetch topics have failed, but that's of lower priority than handling a similar issue with fetch requests errors from `MessageListView.fetchInitial`.

gnprice:
Specifically I'd think of the data structure here as basically recording the results of past calls to https://zulip.com/api/get-stream-topics (the same endpoint that the initial implementation of #1158 will also use) — but then keeping those results up to date with events.

gnprice:
The main user-facing impact of this, compared to the version in #1500 where those /api/get-stream-topics results are just kept on the state of the _TopicList widget, is:

With the simpler initial version, if the user keeps a topic-list page open while they go and make changes, its list of topics won't get updated. It'll get updated only when they navigate out from the page and then navigate to it afresh, causing a new state to be created and make a new request.

This applies to the names of the topics and their ordering, i.e. the information that comes from /api/get-stream-topics . It doesn't apply to the unread counts, or the muted/followed/etc. states, because those are already part of PerAccountStore and kept up to date with events.

gnprice:
One example symptom of the current state is that if you're looking at the topic list and you resolve a topic from there, the update doesn't appear, making it look as if it didn't work. Then if you try again, the resulting error (from the server) is confusing:
[#mobile > List topics and mark as resolved @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/List.20topics.20and.20mark.20as.20resolved/near/2194420)

PR #1508, opened 2025-05-09
Keep topics data up-to-date by handling events

PIG208:
Fixes: #1499


lib/model/channel.dart
```diff
@@ -369,4 +429,45 @@ class ChannelStoreImpl with ChannelStore {
...
+  /// Handle a [MessageEvent], returning whether listeners should be notified.
+  bool handleMessageEvent(MessageEvent event) {
+    if (event.message is! StreamMessage) return false;
+    final StreamMessage(:streamId, :topic) = event.message as StreamMessage;
+
+    final latestMessageIdsByTopic = _latestMessageIdsByStreamTopic[streamId];
+    if (latestMessageIdsByTopic == null) return false;
+    assert(!latestMessageIdsByTopic.containsKey(topic)
+           || latestMessageIdsByTopic[topic]! < event.message.id);
+    latestMessageIdsByTopic[topic] = event.message.id;
```
gnprice, line 498: One thing we'll want to think through in how this part of the model should work is how to handle the fetch/event race which it introduces.

An example of that is that I think this `assert` isn't correct — we could do a fetch that learns about a topic with a new message, and only later get the MessageEvent about that message.

For an example of where we've reasoned through this sort of fetch/event race before, see the implementation of `reconcileMessages` in lib/model/message.dart .

lib/model/channel.dart
```diff
@@ -203,6 +235,34 @@ class ChannelStoreImpl with ChannelStore {
...
+  /// For example: `_latestMessageIdsByStreamTopic[stream.id][topic] = maxId`
+  ///
+  /// In some cases, the same message IDs, when affected by message moves, can
+  /// be present for mutliple stream-topic keys.
+  final Map<int, Map<TopicName, int>> _latestMessageIdsByStreamTopic = {};
+
+  @override
+  Future<void> fetchTopics(int streamId) async {
+    final result = await _apiGetStreamTopics(connection, streamId: streamId,
+      allowEmptyTopicName: true);
```
gnprice, line 306: What if we've already successfully fetched the list for this channel? In that case I think we want to show the existing list (which, after all, this class will have been keeping up to date) rather than make a new fetch.

lib/model/channel.dart
```diff
@@ -369,4 +429,45 @@ class ChannelStoreImpl with ChannelStore {
...
+    final UpdateMessageMoveData(
+      :origStreamId, :origTopic, :newStreamId, :newTopic, :propagateMode,
+    ) = event.moveData!;
+    bool shouldNotify = false;
+
+    final origLatestMessageIdsByTopics = _latestMessageIdsByStreamTopic[origStreamId];
+    if (propagateMode == PropagateMode.changeAll
+        && origLatestMessageIdsByTopics != null) {
+      shouldNotify = origLatestMessageIdsByTopics.remove(origTopic) != null;
+    }
```
gnprice, line 515: I think this might be the right logic to have, but there are some edge cases it doesn't handle. So this code should have comments pointing out those cases and explaining why we're choosing to not handle them.

lib/widgets/topic_list.dart
```diff
@@ -231,6 +225,14 @@ class _TopicItem extends StatelessWidget {
...
     final store = PerAccountStoreWidget.of(context);
     final designVariables = DesignVariables.of(context);
+    // The message with `maxId` might not remain in `topic` since we last fetch
+    // the list of topics.  Make sure we check for that before passing `maxId`
+    // to the topic action sheet.
+    // See also: [ChannelStore.getStreamTopics]
+    final message = store.messages[maxId];
+    final isMaxIdInTopic = message is StreamMessage
+      && message.streamId == streamId
+      && message.topic.isSameAs(topic);
```
gnprice, line 235: Hmm. This will mean that if we just don't happen to have any messages from that topic in MessageStore — for example, because it's not a topic that's included in the very last messages — then we won't give the user the option to resolve or unresolve the topic from this screen. That seems unfortunate.

What happens if we go ahead and use this message ID even though it might turn out to no longer be in the topic?

gnprice (commented):
Thanks @PIG208 for drafting this!

Generally this approach looks good. We'll have someone else pick it up from here.

I skimmed through the key points of the logic; some high-level comments below.

(For whoever picks this up next: do read through all the details yourself in addition to the points I've commented on, and follow the steps in [Zulip's guide to continuing unfinished work](https://zulip.readthedocs.io/en/latest/contributing/continuing-unfinished-work.html).)

gnprice:
Hmm, infra failure in CI:
```
Downloading Linux x64 Dart SDK from Flutter engine 2d053c11840d49924aefe84dabb94545a3f9282e...

100   255  100   255    0     0   3658      0 --:--:-- --:--:-- --:--:--  3695

```

That's the same symptom as at https://github.com/zulip/zulip-flutter/pull/1688#issuecomment-3050644920 , when Flutter's `origin/main` diverged from its `origin/master`. But I just fetched in my Flutter clone to check, and both those branches are at the same commit flutter/flutter@f277865a8 — so the cause this time is different.

And in fact there doesn't seem to be a commit in the upstream repo with the commit ID from that failure output:
flutter/flutter@2d053c11840d49924aefe84dabb94545a3f9282e

gnprice:
Discussion of that upstream issue: https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/Flutter.20upgrade.20-.20downloding.20Dark.20SDK.20fails/near/2228299

Apparently fixed now.

lib/widgets/topic_list.dart
sm-sayedi: If the message ID is not in the topic anymore, then resolve/unresolve will not be successful and will display a dialog with the following message: `You only have permission to move the 1/7 most recent messages in this topic`.
(the topic where the ID now actually belongs to has 7 messages in total)

gnprice:
Continued as #1951. Thanks again @PIG208!

PR #1509, opened 2025-05-10
 local echo (4.5/n): Prep for creating outbox messages on send

PIG208:
This prepares for #1472, toward #1441.

Requested by @chrisbobbe [here](https://github.com/zulip/zulip-flutter/pull/1472#pullrequestreview-2830021645).

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -550,6 +550,15 @@ String? tryParseEmojiCodeToUnicode(String emojiCode) {
   }
 }
 
+/// The topic servers understand to mean "there is no topic".
+///
+/// This should match
+///   https://github.com/zulip/zulip/blob/6.0/zerver/actions/message_edit.py#L940
+/// or similar logic at the latest `main`.
+// This is hardcoded in the server, and therefore untranslated; that's
+// zulip/zulip#3639.
+const String kNoTopicTopic = '(no topic)';
+
 /// The name of a Zulip topic.
 // TODO(dart): Can we forbid calling Object members on this extension type?
 //   (The lack of "implements Object" ought to do that, but doesn't.)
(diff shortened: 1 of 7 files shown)


lib/api/model/model.dart
```diff
@@ -604,6 +613,37 @@ extension type const TopicName(String _value) {
...
+    required String? realmEmptyTopicDisplayName,
+  }) {
+    assert(_value.trim() == _value);
+    // TODO(server-10) simplify this away
+    if (zulipFeatureLevel < 334) {
+      assert(_value.isNotEmpty);
+      return this;
+    }
+
+    if (_value == kNoTopicTopic || _value.isEmpty) {
```
gnprice, line 639: The "(no topic)" case here looks like it started more recently than the other, at FL 370:
https://zulip.com/api/changelog

lib/api/model/model.dart
```diff
@@ -604,6 +613,37 @@ extension type const TopicName(String _value) {
   /// using [canonicalize].
   bool isSameAs(TopicName other) => canonicalize() == other.canonicalize();
 
+  /// Convert this topic to match how it would appear on a message object from
+  /// the server.
+  ///
+  /// This assumes that the topic is originally for a send-message
+  /// request — such a topic must be constructed from a string without
+  /// leading/trailing whitespace.
```
gnprice, line 621: Hmm, is that a requirement of the server endpoint? I don't see it here:
https://zulip.com/api/send-message#parameter-topic

Fine for this method to have this requirement; just shouldn't attribute it to the server.

test/api/model/model_test.dart
```diff
@@ -161,6 +161,27 @@ void main() {
...
+      final emptyTopicDisplayName = eg.defaultRealmEmptyTopicDisplayName;
+      void doCheck(TopicName topicA, TopicName expected, int zulipFeatureLevel) {
+        check(topicA.interpretAsServer(
+          zulipFeatureLevel: zulipFeatureLevel,
+          realmEmptyTopicDisplayName: emptyTopicDisplayName),
+        ).equals(expected);
+      }
+
+      check(() => doCheck(eg.t(''),        eg.t(''),                    333))
+        .throws<void>();
```
gnprice, line 175: The `throws` here would end up catching an error from the `check` itself failing. That doesn't seem intended — I gather what you want is to check that `interpretAsServer` itself throws.

So for example this test as written still passes if you change this 333 to 334, at which point `interpretAsServer` doesn't throw but just returns an answer different from `eg.t('')`.

lib/api/model/model.dart
```diff
@@ -604,6 +613,37 @@ extension type const TopicName(String _value) {
...
+  ///
+  /// For a client that does not support empty topics,
+  /// a modern server (FL>=334) would convert "(no topic)" and empty topics to
+  /// `store.realmEmptyTopicDisplayName`.
+  ///
+  /// See also: https://zulip.com/api/send-message#parameter-topic
+  TopicName interpretAsServer({
```
gnprice, line 628: This name doesn't feel right to me — "interpret as server" sounds like it's trying to interpret the topic name as being a server (whatever that would mean).

How about `processLikeServer`?

test/example_data.dart
```diff
@@ -660,7 +674,7 @@ UpdateMessageEvent updateMessageEditEvent(
     messageId: messageId,
     messageIds: [messageId],
     flags: flags ?? origMessage.flags,
-    editTimestamp: editTimestamp ?? 1234567890, // TODO generate timestamp
+    editTimestamp: editTimestamp ?? utcTimestamp(),
```
gnprice, line 677: I guess I commented on this change earlier today over on #1498 where Chris had cherry-picked it:
https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2085155042

gnprice (commented):
Thanks, and thanks @chrisbobbe for the previous reviews! Generally this looks good; a few comments.

PIG208:
Thanks! Updated the PR.

lib/api/model/model.dart
```diff
@@ -604,6 +613,51 @@ extension type const TopicName(String _value) {
   /// using [canonicalize].
   bool isSameAs(TopicName other) => canonicalize() == other.canonicalize();
 
+  /// Process this topic to match how it would appear on a message object from
+  /// the server.
+  ///
+  /// This assumes that the topic is constructed from a string without
+  /// leading/trailing whitespace.
```
gnprice, line 620:
```suggestion
  /// This returns the [TopicName] the server would be predicted to include
  /// in a message object resulting from sending to this [TopicName]
  /// in a [sendMessage] request.
  ///
  /// This [TopicName] is required to have no leading or trailing whitespace.
```

The other part is needed too (I think it came from a comment @chrisbobbe made on the original PR): it says how this topic relates to the hypothetical scenario where the returned topic appears in a message object from the server.

lib/api/model/model.dart
```diff
@@ -604,6 +613,51 @@ extension type const TopicName(String _value) {
...
+
+    // TODO(server-10) simplify this away
+    if (zulipFeatureLevel < 370 && _value == kNoTopicTopic) {
+      // From the API docs:
+      // > Before Zulip 10.0 (feature level 370), "(no topic)" was not
+      // > interpreted as an empty string.
+      return TopicName(kNoTopicTopic);
+    }
+
+    // TODO(#1250): This assumes that the 'empty_topic_name' client capability
```
gnprice, line 651: to update now that #1365 is merged; the doc above, too

gnprice (commented):
Thanks for the revision! The code now all looks good, modulo rebasing atop #1365.

gnprice:
Thanks! Looks good; merging.

Issue #863, opened 2024-08-02
Add saved replies


alya:
In https://github.com/zulip/zulip/issues/31227, we are implementing saved replies in the web app.

It would be useful to be able to insert a saved reply in the mobile app as well.

[API discussion](https://chat.zulip.org/#narrow/stream/378-api-design/topic/saved.20replies)

gnprice:
Our current plan is that we'll skip this feature for the legacy app, but will have it in the new app at launch time. That way in the blog post for the upcoming Zulip Server 10 release, we can say that mobile support for it is coming soon as part of the new app.

Ideally this means we'd have the feature in the new app, or at least have started implementing it here, by the time of the Server 10 release in order to be sure there are no surprises before launch in how much work it takes.

PIG208:
API documentation:
- https://zulip.com/api/get-saved-snippets
- https://zulip.com/api/create-saved-snippet
- https://zulip.com/api/delete-saved-snippet

PIG208:
As mentioned [here](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/.23F863.20saved.20replies/near/2110089)[#mobile-design > #F863 saved replies @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/.23F863.20saved.20replies/near/2110089), the search bar on the bottom sheet will be replaced with the title "Saved snippets".

[And](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/.23F863.20saved.20replies/near/2169460)[#mobile-design > #F863 saved replies @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/.23F863.20saved.20replies/near/2169460) we will leave editing saved snippets as a follow-up.

PR #1511, opened 2025-05-13
Add and handle live-updates to saved snippets data

PIG208:
This prepares the data model and API bindings for #1391 towards #863.

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -37,6 +37,13 @@ sealed class Event {
           case 'update': return RealmUserUpdateEvent.fromJson(json);
           default: return UnexpectedEvent.fromJson(json);
         }
+      case 'saved_snippets':
+        switch (json['op'] as String) {
+          case 'add': return SavedSnippetsAddEvent.fromJson(json);
+          case 'update': return SavedSnippetsUpdateEvent.fromJson(json);
+          case 'remove': return SavedSnippetsRemoveEvent.fromJson(json);
+          default: return UnexpectedEvent.fromJson(json);
+        }
       case 'stream':
         switch (json['op'] as String) {
           case 'create': return ChannelCreateEvent.fromJson(json);
@@ -334,6 +341,68 @@ class RealmUserUpdateEvent extends RealmUserEvent {
   // TODO make round-trip (see _readFromPerson)
   @override
   Map<String, dynamic> toJson() => _$RealmUserUpdateEventToJson(this);
+}
+
+/// A Zulip event of type `saved_snippets`: https://zulip.com/api/get-events#saved_snippets-add
+sealed class SavedSnippetsEvent extends Event {
+  @override
+  @JsonKey(includeToJson: true)
+  String get type => 'saved_snippets';
+
+  String get op;
+
+  SavedSnippetsEvent({required super.id});
+}
+
+/// A [SavedSnippetsEvent] with op `add`: https://zulip.com/api/get-events#saved_snippets-add
+@JsonSerializable(fieldRename: FieldRename.snake)
+class SavedSnippetsAddEvent extends SavedSnippetsEvent {
+  @override
+  String get op => 'add';
+
+  final SavedSnippet savedSnippet;
+
+  SavedSnippetsAddEvent({required super.id, required this.savedSnippet});
+
+  factory SavedSnippetsAddEvent.fromJson(Map<String, dynamic> json) =>
+    _$SavedSnippetsAddEventFromJson(json);
+
+  @override
+  Map<String, dynamic> toJson() => _$SavedSnippetsAddEventToJson(this);
+}
+
+/// A [SavedSnippetsEvent] with op `update`: https://zulip.com/api/get-events#saved_snippets-update
+@JsonSerializable(fieldRename: FieldRename.snake)
+class SavedSnippetsUpdateEvent extends SavedSnippetsEvent {
+  @override
+  String get op => 'update';
+
+  final SavedSnippet savedSnippet;
+
+  SavedSnippetsUpdateEvent({required super.id, required this.savedSnippet});
+
+  factory SavedSnippetsUpdateEvent.fromJson(Map<String, dynamic> json) =>
+    _$SavedSnippetsUpdateEventFromJson(json);
+
+  @override
+  Map<String, dynamic> toJson() => _$SavedSnippetsUpdateEventToJson(this);
+}
+
+/// A [SavedSnippetsEvent] with op `remove`: https://zulip.com/api/get-events#saved_snippets-remove
+@JsonSerializable(fieldRename: FieldRename.snake)
+class SavedSnippetsRemoveEvent extends SavedSnippetsEvent {
+  @override
+  String get op => 'remove';
+
+  final int savedSnippetId;
+
+  SavedSnippetsRemoveEvent({required super.id, required this.savedSnippetId});
+
+  factory SavedSnippetsRemoveEvent.fromJson(Map<String, dynamic> json) =>
+    _$SavedSnippetsRemoveEventFromJson(json);
+
+  @override
+  Map<String, dynamic> toJson() => _$SavedSnippetsRemoveEventToJson(this);
 }
 
 /// A Zulip event of type `stream`.
--- a/lib/api/route/saved_snippets.dart
+++ b/lib/api/route/saved_snippets.dart
@@ -1,0 +1,31 @@
+import 'package:json_annotation/json_annotation.dart';
+
+import '../core.dart';
+
+part 'saved_snippets.g.dart';
+
+/// https://zulip.com/api/create-saved-snippet
+Future<CreateSavedSnippetResult> createSavedSnippet(ApiConnection connection, {
+  required String title,
+  required String content,
+}) {
+  assert(connection.zulipFeatureLevel! >= 297); // TODO(server-10)
+  return connection.post('createSavedSnippet', CreateSavedSnippetResult.fromJson, 'saved_snippets', {
+    'title': RawParameter(title),
+    'content': RawParameter(content),
+  });
+}
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class CreateSavedSnippetResult {
+  final int savedSnippetId;
+
+  CreateSavedSnippetResult({
+    required this.savedSnippetId,
+  });
+
+  factory CreateSavedSnippetResult.fromJson(Map<String, dynamic> json) =>
+    _$CreateSavedSnippetResultFromJson(json);
+
+  Map<String, dynamic> toJson() => _$CreateSavedSnippetResultToJson(this);
+}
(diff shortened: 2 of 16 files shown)


lib/api/model/events.dart
```diff
@@ -37,6 +37,13 @@ sealed class Event {
           case 'update': return RealmUserUpdateEvent.fromJson(json);
           default: return UnexpectedEvent.fromJson(json);
         }
+      case 'saved_snippets':
```
chrisbobbe, line 40:
```
api: Add save_snippets events and handle live-updates
```

Commit-message nit: `saved_snippets`

lib/model/saved_snippet.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:collection/collection.dart';
+
+import '../api/model/events.dart';
+import '../api/model/model.dart';
+
+mixin SavedSnippetStore {
+  Map<int, SavedSnippet> get savedSnippets;
+}
+
+class SavedSnippetStoreImpl with SavedSnippetStore {
```
chrisbobbe, line 10: Normally a substore would extend `PerAccountStoreBase`…it seems we don't depend on that here though 🙂 but maybe it's helpful to do it anyway, in case it's needed in the future, and also as an example for future substore implementations?

```diff
diff --git lib/model/saved_snippet.dart lib/model/saved_snippet.dart
index 66b50781d..92cb253ef 100644
--- lib/model/saved_snippet.dart
+++ lib/model/saved_snippet.dart
@@ -2,14 +2,18 @@ import 'package:collection/collection.dart';
 
 import '../api/model/events.dart';
 import '../api/model/model.dart';
+import 'store.dart';
 
 mixin SavedSnippetStore {
   Map<int, SavedSnippet> get savedSnippets;
 }
 
-class SavedSnippetStoreImpl with SavedSnippetStore {
-  SavedSnippetStoreImpl({required Iterable<SavedSnippet> savedSnippets})
-    : _savedSnippets = {
+class SavedSnippetStoreImpl extends PerAccountStoreBase with SavedSnippetStore {
+  SavedSnippetStoreImpl({
+    required super.core,
+    required Iterable<SavedSnippet> savedSnippets,
+  }) :
+      _savedSnippets = {
         for (final savedSnippet in savedSnippets)
           savedSnippet.id: savedSnippet,
       };
```

test/model/saved_snippet.dart
```diff
@@ -0,0 +1,44 @@
...
+import '../example_data.dart' as eg;
+import 'store_checks.dart';
+
+void main() {
+  test('handleSavedSnippetsEvent', () async {
+    final store = eg.store(initialSnapshot: eg.initialSnapshot(
+      savedSnippets: [eg.savedSnippet(id: 101)]));
+    check(store).savedSnippets.values.single.id.equals(101);
+
+    await store.handleEvent(SavedSnippetsAddEvent(id: 1,savedSnippet:
```
chrisbobbe, line 16: nit: space after comma; actually maybe put `savedSnippet:` on the next line, with `eg.savedSnippet(`?

lib/api/model/events.dart
chrisbobbe: Also this is labeled as an "api:" commit, but it also adds to model code; how about separate `api:` and `model:` commits?

(Or perhaps just one `model:` commit with the API-binding additions lumped into that same commit—but if doing that, it would be logical to lump _all_ the new API-binding code into that commit, including the initial-snapshot code.)

chrisbobbe (commented):
Thanks!! Small comments below.

lib/api/route/saved_snippets.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:json_annotation/json_annotation.dart';
+
+import '../core.dart';
+
+part 'saved_snippets.g.dart';
+
+/// https://zulip.com/api/create-saved-snippet
+Future<CreateSavedSnippetResult> createSavedSnippet(ApiConnection connection, {
```
chrisbobbe, line 8:
```
api: Add createSavedSnippet route
```

This needs a simple smoke test in test/api/route/.

lib/model/store.dart
```diff
@@ -868,6 +877,11 @@ class PerAccountStore extends PerAccountStoreBase with ChangeNotifier, EmojiStor
         autocompleteViewManager.handleRealmUserUpdateEvent(event);
         notifyListeners();
 
+      case SavedSnippetsEvent():
+        assert(debugLog('server event: saved_snippets/${event.op}'));
+        _savedSnippets.handleSavedSnippetsEvent(event);
+        notifyListeners();
```
chrisbobbe, line 883: Ah I guess this case needs to be handled (as a no-op) in the earlier commit—

```
api: Add saved_snippets events
```

—to satisfy the analyzer there.

lib/model/saved_snippet.dart
```diff
@@ -0,0 +1,38 @@
...
+
+import '../api/model/events.dart';
+import '../api/model/model.dart';
+import 'store.dart';
+
+mixin SavedSnippetStore {
+  Map<int, SavedSnippet> get savedSnippets;
+}
+
+class SavedSnippetStoreImpl extends PerAccountStoreBase with SavedSnippetStore {
```
chrisbobbe, line 11:
```
model: Handle live-updates to savedSnippets
```

Commit-message nit: this commit does more than just handle live-updates to savedSnippets—it's what adds savedSnippets into the codebase in the first place :)

Could be:

```
model: Add PerAccountStore.savedSnippets, updating with events
```

or similar.

chrisbobbe (commented):
Thanks! Small comments below.

PIG208:
Thanks! Pushed an update.

chrisbobbe (commented):
Thanks! One nit below, and I'll mark for Greg's review.

test/api/model/saved_snippets_test.dart
```diff
@@ -0,0 +1,25 @@
...
+      connection.prepare(
+        json: CreateSavedSnippetResult(savedSnippetId: 123).toJson());
+      final result = await createSavedSnippet(connection,
+        title: 'test saved snippet', content: 'content');
+      check(result).savedSnippetId.equals(123);
+      check(connection.takeRequests()).single.isA<http.Request>()
+        .bodyFields.deepEquals({
+        'title': 'test saved snippet',
+        'content': 'content',
+      });
```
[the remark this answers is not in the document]
PIG208: Oh right. Oops 😬

lib/api/model/events.dart
```diff
@@ -336,6 +343,71 @@ class RealmUserUpdateEvent extends RealmUserEvent {
   Map<String, dynamic> toJson() => _$RealmUserUpdateEventToJson(this);
 }
 
+/// A Zulip event of type `saved_snippets`.
+///
+/// The corresponding API docs are in several places for
+/// different values of `op`; see subclasses.
+sealed class SavedSnippetsEvent extends Event {
```
gnprice, line 350: nit: It looks like these docs are actually next to each other, so this class can usefully link to the one that comes first. (The subclasses can still link to the individual sections.)

This wording was originally for `realm_user`, which is actually scattered across two widely-separated places in https://zulip.com/api/get-events . (And I see I wrote it at the same time for `stream` but it's not true of `stream` now; not sure if that was an oversight or that one got fixed since then.)

gnprice (commented):
Thanks to you both! I've done a quick skim, and this structure looks good. I like the organization with the separate API and model/store commits (as per https://github.com/zulip/zulip-flutter/pull/1511#discussion_r2098733072), with this to decouple the API event commit from the store changes:
```dart
      case SavedSnippetsEvent():
        // TODO handle
        break;
```

I'll aim to do a full review tomorrow.

lib/api/route/saved_snippets.dart
```diff
@@ -0,0 +1,30 @@
...
+  required String title,
+  required String content,
+}) {
+  return connection.post('createSavedSnippet', CreateSavedSnippetResult.fromJson, 'saved_snippets', {
```
gnprice, line 12:
```suggestion
}) {
  assert(connection.zulipFeatureLevel! >= 297); // TODO(server-10)
  return connection.post('createSavedSnippet', CreateSavedSnippetResult.fromJson, 'saved_snippets', {
```

That doubles as a reminder, when writing code that calls this endpoint, of what feature level to check for before offering the feature in the UI.

test/api/model/saved_snippets_test.dart
```diff
@@ -0,0 +1,25 @@
...
+import '../../stdlib_checks.dart';
+import '../fake_api.dart';
+import '../route/route_checks.dart';
+
+void main() {
+  test('smoke', () async {
+    return FakeApiConnection.with_((connection) async {
+      connection.prepare(
+        json: CreateSavedSnippetResult(savedSnippetId: 123).toJson());
+      final result = await createSavedSnippet(connection,
```
gnprice, line 15: this file goes in test/api/route/ :slightly_smiling_face: (not …/model/)

test/api/model/saved_snippets_test.dart
```diff
@@ -0,0 +1,25 @@
...
+        title: 'test saved snippet', content: 'content');
+      check(result).savedSnippetId.equals(123);
+      check(connection.takeRequests()).single.isA<http.Request>()
+        .bodyFields.deepEquals({
```
gnprice, line 19: As long as we're writing this test, it's only a couple more lines to have it check the other key details of the request too: the method and URL path. So let's include those too.

test/api/model/saved_snippets_test.dart
```diff
@@ -0,0 +1,25 @@
...
+
+void main() {
+  test('smoke', () async {
+    return FakeApiConnection.with_((connection) async {
+      connection.prepare(
+        json: CreateSavedSnippetResult(savedSnippetId: 123).toJson());
+      final result = await createSavedSnippet(connection,
+        title: 'test saved snippet', content: 'content');
+      check(result).savedSnippetId.equals(123);
+      check(connection.takeRequests()).single.isA<http.Request>()
```
gnprice, line 18: nit: the order is a bit more logical if we check the request first, then the response — after all, the request had to happen before we would receive the response

gnprice (commented):
OK, and here's a full review — just small comments below.

(And I see you've already revised to handle my one comment above, thanks.)

test/api/model/model_checks.dart
```diff
@@ -21,6 +21,12 @@ extension UserChecks on Subject<User> {
   Subject<bool> get isSystemBot => has((x) => x.isSystemBot, 'isSystemBot');
 }
 
+extension SavedSnippetChecks on Subject<SavedSnippet> {
+  Subject<int> get id => has((x) => x.id, 'id');
+  Subject<String> get title => has((x) => x.title, 'title');
+  Subject<String> get content => has((x) => x.content, 'content');
+}
```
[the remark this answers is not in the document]
PIG208: That's neat! I usually went line by line replacing the symbols and the string. This sounds more efficient when you have multiple fields.

PIG208:
Thanks! Updated the PR.

gnprice:
Thanks! Looks good; merging.

Issue #1139, opened 2024-12-11
i18n: Allow selecting a different language than device setting


chrisbobbe:
Currently, the app UI follows the device setting for what language it should try to display in.

This issue is for an in-app setting to choose any language we have translations for, instead of always following the device.

An instance of:
- #97.

Gaurav-Kushwaha-1225:
Hi @chrisbobbe,

This storing of some client settings isn't yet implemented for Theme Switching among dark/light/system.
Am I right?

If yes. Then, where we might add this feature for language now and theme later on.
I would suggest keeping it here, in the menu button.

<img src="https://github.com/user-attachments/assets/c82baa9e-c530-4033-b831-3b83159e8cfb" width=30%>

Thank you.

gnprice:
As an example of where this is desired, we heard from a user [today](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Default.20language.20in.20Flutter.20app/near/2005465):
> In our team we use English as a default language in communication and collaborative apps, tho our native language is not English. Right now in Flutter app I can't change system language, but in legacy app it's not a problem.
> […] my system language is not English

So they want Zulip in English although they have a different language chosen for the system overall.

PrinceJohn279:
hi this is going to be my first pr! any advice?

chrisbobbe:
Hi, welcome! See the guide in our README and the resources linked from there: https://github.com/zulip/zulip-flutter#picking-an-issue-to-work-on

As I said [above](https://github.com/zulip/zulip-flutter/issues/1139#issuecomment-2542467199):

> So let's wait to work on this until [PIG208] implements a settings UI where this setting will go. For now, if you'd like to contribute, please choose a different issue to work on.

AshutoshKhadse23:
Hello @chrisbobbe ,
Can you assign this to me?

https://github.com/user-attachments/assets/f9d1300c-e17b-49a2-bab4-d94bc76201d9

gnprice:
Thanks. Instead of a popup submenu, let's put this as a screen underneath the main settings screen — that's a UX that will scale better when there are dozens of translations, which I expect there to be soon after the Zulip project as a whole completes our migration to Weblate and asks our community of translation contributors to translate this app there.

For a model of how that UI can work, see the language setting in the legacy app.

To make it available from the choose-account screen, we can add an option to go to the settings screen as one of the options in that menu.

AshutoshKhadse23:
[...]

Issue #1077, opened 2024-11-23
☂️ Upstream issues affecting Zulip Flutter app


gnprice:
This is a tracking issue, collecting notes on potential work in Flutter, Dart, or other upstreams that would benefit the (new) Zulip mobile app.

Some of these are much more important to us than others.

Many of these have corresponding issues in our tracker with the ["upstream" label](https://github.com/zulip/zulip-flutter/issues?q=is%3Aissue+is%3Aopen+label%3Aupstream). This issue should describe all of those (but doesn't yet as it's still being written.)

# Flutter

- Emoji small on iOS: https://github.com/flutter/flutter/issues/28894
- WidgetSpan double-applies text scaling: #735 
  Especially affects a11y, at large font size.
- #226 
  RawAutocomplete should gain a "controller" for customizability.
- `package:video_player` bugs on iOS:
  https://github.com/flutter/flutter/issues/126760
  https://github.com/flutter/flutter/issues/139107
  Because of these, for now we don't try to play video in the message list; we only play if the user opens in lightbox.
  Details here: https://github.com/zulip/zulip-flutter/issues/656
- Padding for text spans, to let us match Zulip web's styling for code spans.
  An overlapping issue: flutter/flutter#30541
  (Is there an exactly on-point issue?)
- Richer `Flex` (aka `Row` and `Column`), with more CSS flexbox features like flex-basis vs flex-grow vs flex-shrink.
  We probably need something like this for:
  #282
  #433
- We want to have scripts that run as a CLI (at dev time) and import our model code.
  - E.g., #190
    where we wound up with a real hack of putting the logic in a "test", run with `flutter test`, and having a wrapper script around that to accept a more reasonable CLI and run `flutter test`. (`flutter test` has a fine CLI for its actual job; but a terrible one for running an arbitrary tool as a script, which it's not intended for.)      
    Potentially also: #617
  - This is a well-known problem in the community. The trouble is our model code imports `package:flutter/foundation.dart` for things like `ChangeNotifier` and `dart:ui` for things like `Color`; the former imports the latter; and `dart:ui` is available only in the Flutter SDK, not `dart run`.
  - One existing proposal is to move some definitions to a package outside the Flutter SDK. Not sure if that will succeed in solving the problem.

# Dart SDK

- Dart needs a way to make code using `DateTime.now` be testable: 
  https://github.com/dart-lang/sdk/issues/28985
- Dart needs a way to make code that uses the local timezone be testable.
[...]

PR #1513, opened 2025-05-16
Support language settings

PIG208:
#1139


chrisbobbe:
When I rebase atop `main`, there's some failures at the first commit:

<details>

```
$ tools/check analyze l10n
Running analyze...
Analyzing zulip-flutter...                                              

  error • Missing concrete implementations of 'getter abstract class
         ZulipLocalizations.languageEn', 'getter abstract class ZulipLocalizations.languagePl',
         'getter abstract class ZulipLocalizations.languageRu', and 'getter abstract class
         ZulipLocalizations.languageUk' • lib/generated/l10n/zulip_localizations_de.dart:8:7 •
         non_abstract_class_inherits_abstract_member

1 issue found. (ran in 3.0s)
Running l10n...
Error: there were updates to l10n:
 M lib/generated/l10n/zulip_localizations_de.dart

FAILED: analyze l10n

To rerun the suites that failed, run:
  $ tools/check analyze l10n
```

</details>

Could you rebase and fix those please?

PIG208:
Thanks. Just updated the PR!

docs/translation.md
```diff
@@ -79,6 +81,21 @@ For example:
 `zulipLocalizations.subscribedToNChannels(store.subscriptions.length)`.
 
 
+## Adding a new language
+
+ARB files for new languages are automatically created in Pull Requests generated
+[the update-translations GitHub workflow](/.github/workflows/update-translations.yml).
+However, this won't make them appear in settings.
+
+When a language has a good percentage of strings translated, we should add it to
```
chrisbobbe, line 90: Does Weblate show the percentage? Let's try to match what zulip-mobile uses for a threshold; I think it might be 5%?

lib/model/localizations.dart
```diff
@@ -15,3 +17,60 @@ abstract final class GlobalLocalizations {
   static ZulipLocalizations zulipLocalizations =
     lookupZulipLocalizations(ZulipLocalizations.supportedLocales.first);
 }
+
+/// The mapping, of the locales we offer, to their display name in their own
+/// locale a.k.a. selfname.
+///
+/// [ZulipLocalizations.supportedLocales] can include languages that only have
+/// a few strings translated.  This map should only inlcude sufficiently
+/// translated locales from [ZulipLocalizations.supportedLocales].
```
chrisbobbe, line 26:
```suggestion
/// This map only includes some of [ZulipLocalizations.supportedLocales];
/// it includes languages that have substantially complete translations.
/// For what counts as substantially translated, see docs/translation.md.
```

(and then make sure docs/translation.md answers the question)

lib/model/localizations.dart
```diff
@@ -15,3 +17,60 @@ abstract final class GlobalLocalizations {
...
+///
+/// The map should be sorted by selfname, to help users find their
+/// language in the UI.  When in doubt how to sort (like between different
+/// scripts, or in scripts you don't know), try to match the order found in
+/// other UIs, like for choosing a language in your phone's system settings.
```
chrisbobbe, line 31: Does a map have an order? Maybe we should hard-code a `List<(Locale, String)>`, which is definitely ordered, and compute `kSelfnamesByLocale` from that?

lib/model/localizations.dart
```diff
@@ -15,3 +17,60 @@ abstract final class GlobalLocalizations {
...
+      case 'en':
+        return languageEn;
+      case 'pl':
+        return languagePl;
+      case 'ru':
+        return languageRu;
+      case 'uk':
+        return languageUk;
+      default:
+        throw ArgumentError.value(locale, 'locale');
```
chrisbobbe, line 73: Ah (following from my comment above): or we have a list where each entry contains the language tag, `Locale`, and `languageEn`/etc.? And from that list, we generate `kSelfnamesByLocale` and a map to support this `localeDisplayName` method?

lib/model/database.dart
```diff
@@ -24,13 +26,66 @@ class GlobalSettings extends Table {
...
+  /// [variant subtags](https://www.unicode.org/reports/tr35/#unicode_variant_subtag).
+  /// nor [extensions](https://www.unicode.org/reports/tr35/#extensions)
+  /// are supported.  It throws when it fails to convert [languageTag] into a
+  /// [Locale].
+  ///
+  /// See also:
+  ///  * https://www.unicode.org/reports/tr35/#Unicode_language_identifier;
+  ///    the method implements a subset of this EBNF grammar.
+  // TODO(upstream): support Locale.fromLanguageTag:
+  //   https://github.com/flutter/flutter/issues/143491
```
chrisbobbe, line 56: Let's remember to add this to

- #1077

lib/model/localizations.dart
chrisbobbe: And indeed the commit message mentions making a list, but I guess the commit doesn't actually do that 🙂:

```
i18n: Maintain a list of supported languages with metadata
```

lib/model/database.dart
```diff
@@ -24,13 +26,66 @@ class GlobalSettings extends Table {
...
+  Locale _fromLanguageTag(String languageTag) {
```
chrisbobbe, line 57: Could you add a comment saying how you produced this implementation? I'm not sure if it was written from scratch looking at a spec, or if it's a simplified version of something in `Locale.fromLanguageTag`, etc.

lib/widgets/settings.dart
```diff
@@ -17,17 +21,28 @@ class SettingsPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final zulipLocalizations = ZulipLocalizations.of(context);
+
+    Widget? subtitle;
```
chrisbobbe, line 25: As a variable in `SettingsPage.build`, this looks like it should have a more specific name—it's not the page's subtitle, but the subtitle for something on a specific item in the page.

chrisbobbe (commented):
Thanks, this is an exciting feature!!

Small comments below.

And a question: what should happen, and what happens, if the language setting in the database isn't one of the languages we offer in the settings page? I think this could happen if we remove a language setting (e.g. if we become more strict about when to offer a language in the UI) or we change the language tag for a set of translations. If we need to do a database migration, let's document that, or perhaps the app can just notice the inconsistency and unset the setting.

lib/model/database.dart
gnprice: This method looks like it's basically an implementation of the desired upstream feature — is that right?

In that case let's keep the TODO but it can instead read like:
```
// TODO(upstream): send this as a factory Locale.fromLanguageTag:
//   https://github.com/flutter/flutter/issues/143491
```

and no need to add it to that umbrella list — the issue isn't really affecting us, it's just an opportunity to send something useful upstream :slightly_smiling_face:

PIG208:
> And a question: what should happen, and what happens, if the language setting in the database isn't one of the languages we offer in the settings page? I think this could happen if we remove a language setting (e.g. if we become more strict about when to offer a language in the UI) or we change the language tag for a set of translations. If we need to do a database migration, let's document that, or perhaps the app can just notice the inconsistency and unset the setting.

Yeah, that's a good point. In case of removal withou replacement, we will just assume the language is not set, until the user chooses one from the settings page — that's the sceneario the test "handle unsupported (but valid) locale stored in database" covers.

When we replace a language tag with another, e.g. replacing `zh-Hant` with `zh-Hant-TW` we can probably maintain maps of names we have used before for each language tag, and write back its latest form to the database if `zh-Hant` is used.

I'm not sure how common it is going to be for us to repurpose a language tag like `zh-Hant` when it goes out of favor. When that happens, we can probably just bump the schema version and write a migration for it, without actually updating the database schema.

lib/model/database.dart
PIG208: This was written from scratch to support parsing Unicode Language Identifier, whose grammar is defined here:
https://www.unicode.org/reports/tr35/#Unicode_language_identifier

How about this:

```dart
  /// Parse a Unicode BCP 47 Language Identifier into [Locale].
  ///
  /// Throw when it fails to convert [languageTag] into a [Locale].
  ///
  /// This supports parsing a Unicode Language Identifier returned from
  /// [Locale.toLanguageTag].
  ///
  /// This implementation refers to a part of
  /// [this EBNF grammar](https://www.unicode.org/reports/tr35/#Unicode_language_identifier),
  /// assuming the identifier is valid without
  /// [unicode_variant_subtag](https://www.unicode.org/reports/tr35/#unicode_variant_subtag).
  ///
  /// This doesn't check if the [languageTag] is a valid identifier, (i.e., when
  /// this returns without errors, the identifier is not necessarily
  /// syntactically well-formed or valid).
```

PIG208:
Thanks for the review! I have updated the PR and left a TODO for migrations. I think the current behavior should be forward-compatible, and we can add migrations or locale-alias mappings later on, if needed.

gnprice:
> to repurpose a language tag like `zh-Hant` when it goes out of favor

I expect it will never happen that we take a language tag which used to mean one thing, and later re-use the same tag to mean a different thing. That's because these tags follow an Internet-wide standard (https://en.wikipedia.org/wiki/IETF_language_tag); and although a migration to free up an old name is an option for our app's local database, it's not an option for such a distributed standard. So those standard tags will never reuse an old name with a new meaning.

PR #1515, opened 2025-05-18
anchors 9/n: Add anchor, fetchNewer, haveNewest to msglist view-model

gnprice:
This is the next round after #1507 (plus #1512), toward #82 and #80.

At the end of this branch, the view-model MessageListView has everything it needs for #82 and #80: the message list can start at an arbitrary anchor, including a numeric message-ID anchor or AnchorCode.firstUnread, and can fetch more history from there in both directions.

This doesn't yet actually use the new functionality outside of tests. As a result this branch is NFC as to the live app, given a well-behaved server (there are a couple of commits that are not NFC in how they act with a server that gives incoherent responses). I have a draft branch (as I mentioned at [#mobile-team > beta release management @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/beta.20release.20management/near/2174229)) that completes both #82 and #80; that'll be for an upcoming PR or two.

#### 83b89a967 msglist [nfc]: Say we'll show "loading" even when fetch is at other end

This is NFC ultimately because we currently only ever fetch, or
show loading indicators, at one end of the message list, namely
the start.

When we do start supporting a message list in the middle of history,
though (#82), and consequently loading newer as well as older
messages, my conclusion after thinking it through is that we'll want
a "busy fetching" state at one end to mean we show a loading
indicator at the other end too, if it still has more to be fetched.

This would look weird if the user actually saw both at the same time
-- but that shouldn't happen, because if both ends (or even either
end) is still open then the original fetch should have found plenty of
messages to separate them, many screenfuls' worth.

And conversely, if the user does kick off a fetch at one end and then
scroll swiftly to the other end and witness how that appears, we want
to show them a "loading" sign.  The situation is exactly like if
they'd had a fetch attempt on that same end and we were backing off
from failure: there's no fetch right now, but there won't be one yet,
so effectively the loading is busy.

#### e3588e99c msglist [nfc]: Use `fetched` getter when reading

Generally this is helpful because it means that viewing references
to the field will highlight specifically the places that set it.

Here it's also helpful because we're about to replace the field
with an enum shared across several getters.

#### 0d5e84bf4 msglist [nfc]: Use an enum for fetched/fetching/backoff state

This makes the relationships between these flags clearer.

It will also simplify some upcoming refactors that change their
semantics.

#### 5e1f2677f msglist [nfc]: Rename backoff state to share between older/newer

If a fetch in one direction has recently failed, we'll want the
backoff to apply to any attempt to fetch in the other direction too;
after all, it's the same server.

We can also drop the term "cooldown" here, which is effectively
redundant with "backoff".

#### e12d706fe msglist: Set haveNewest from response, like haveOldest

This is NFC with a correctly-behaved server: we set `anchor=newest`,
so the server always sets `found_newest` to true.

Conversely, this will be helpful as we generalize `fetchInitial` to
work with other anchor values; we'll use the `found_newest` value
given by the server, without trying to predict it from the anchor.

The server behavior that makes this effectively NFC isn't quite
explicit in the API docs.  Those say:

  found_newest: boolean

  Whether the server promises that the messages list includes the
  very newest messages matching the narrow (used by clients that
  paginate their requests to decide whether there may be more
  messages to fetch).

  https://zulip.com/api/get-messages#response

But with `anchor=newest`, the response does need to include the very
newest messages in the narrow -- that's the meaning of that `anchor`
value.  So the server is in fact promising the list includes those,
and `found_newest` is therefore required to be true.

(And indeed in practice the server does set `found_newest` to true
when `anchor=newest`; it has specific logic to do so.)

#### f9efd71bb msglist [nfc]: Document narrow field; make setter private

Even if the reader is already sure that the field doesn't get mutated
from outside this file, giving it a different name from the getter is
useful for seeing exactly where it does get mutated: now one can look
at the references to `_narrow`, and see the mutation sites without
having them intermingled with all the sites that just read it.

#### 7b0e40509 msglist: Send positive numAfter for fetchInitial

This is effectively NFC given normal server behavior.  In particular,
the Zulip server is smart enough to skip doing any actual work to
fetch later messages when the anchor is already `newest`.

When we start passing anchors other than `newest`, we'll need this.

#### 8ea4c8d3c msglist: Make initial fetch from any anchor, in model

This is NFC as to the live app, because we continue to always set
the anchor to AnchorCode.newest there.

#### 4e7a18843 msglist: Add fetchNewer method to model

This completes the model layer of #82 and #80: the message list can
start at an arbitrary anchor, including a numeric message-ID anchor
or AnchorCode.firstUnread, and can fetch more history from there
in both directions.

Still to do is to work that into the widgets layer.  This change is
therefore NFC as to the live app: nothing calls this method yet.

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -77,7 +77,7 @@ enum FetchingStatus {
   idle,
 
   /// The model has an active `fetchOlder` request.
-  fetchOlder,
+  fetchingMore,
 
   /// The model is in a backoff period from a failed request.
   backoff,
@@ -135,7 +135,7 @@ mixin _MessageSequence {
   /// This is true both when the recent request is still outstanding,
   /// and when it failed and the backoff from that is still in progress.
   bool get busyFetchingMore => switch (_status) {
-    FetchingStatus.fetchOlder || FetchingStatus.backoff => true,
+    FetchingStatus.fetchingMore || FetchingStatus.backoff => true,
     _ => false,
   };
 
@@ -609,7 +609,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       || (narrow as TopicNarrow).with_ == null);
     assert(messages.isNotEmpty);
     assert(_status == FetchingStatus.idle);
-    _status = FetchingStatus.fetchOlder;
+    _status = FetchingStatus.fetchingMore;
     notifyListeners();
     final generation = this.generation;
     bool hasFetchError = false;
@@ -647,7 +647,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       _haveOldest = result.foundOldest;
     } finally {
       if (this.generation == generation) {
-        assert(_status == FetchingStatus.fetchOlder);
+        assert(_status == FetchingStatus.fetchingMore);
         if (hasFetchError) {
           _status = FetchingStatus.backoff;
           unawaited((_fetchBackoffMachine ??= BackoffMachine())
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -79,8 +79,8 @@ enum FetchingStatus {
   /// The model has an active `fetchOlder` request.
   fetchOlder,
 
-  /// The model is in a backoff period from a failed `fetchOlder` request.
-  fetchOlderCoolingDown,
+  /// The model is in a backoff period from a failed request.
+  backoff,
 }
 
 /// The sequence of messages in a message list, and how to display them.
@@ -135,13 +135,13 @@ mixin _MessageSequence {
   /// This is true both when the recent request is still outstanding,
   /// and when it failed and the backoff from that is still in progress.
   bool get busyFetchingMore => switch (_status) {
-    FetchingStatus.fetchOlder || FetchingStatus.fetchOlderCoolingDown => true,
+    FetchingStatus.fetchOlder || FetchingStatus.backoff => true,
     _ => false,
   };
 
   FetchingStatus _status = FetchingStatus.unstarted;
 
-  BackoffMachine? _fetchOlderCooldownBackoffMachine;
+  BackoffMachine? _fetchBackoffMachine;
 
   /// The parsed message contents, as a list parallel to [messages].
   ///
@@ -316,7 +316,7 @@ mixin _MessageSequence {
     middleMessage = 0;
     _haveOldest = false;
     _status = FetchingStatus.unstarted;
-    _fetchOlderCooldownBackoffMachine = null;
+    _fetchBackoffMachine = null;
     contents.clear();
     items.clear();
     middleItem = 0;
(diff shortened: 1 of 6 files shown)


lib/model/message_list.dart
```diff
@@ -94,7 +94,7 @@ mixin _MessageSequence {
   ///
   /// This may or may not represent all the message history that
   /// conceptually belongs in this message list.
-  /// That information is expressed in [fetched] and [haveOldest].
+  /// That information is expressed in [fetched], [haveOldest], [haveNewest].
```
chrisbobbe, line 99:
```
msglist [nfc]: Introduce haveNewest in model, always true for now
```

`haveNewest` should also control how we respond to new-message events, right? When a message event comes in, we shouldn't add to the end of the list unless `haveNewest` is true. If we do add it to the list, then we'll skip messages after the last that had been fetched, and they won't be filled in by subsequent fetches because `fetchNewer` will use the event's message as the anchor.

chrisbobbe (commented):
This is great! One commit-message nit and one substantive comment on an asymmetry between the fetch-older and fetch-newer logics that we should handle.

lib/model/message_list.dart
```diff
@@ -64,6 +66,25 @@ class MessageListMessageItem extends MessageListMessageBaseItem {
...
+
+  /// The model has made a `fetchInitial` request, which hasn't succeeded.
+  fetchInitial,
+
+  /// The model made a successful `fetchInitial` request,
+  /// and has no outstanding requests or backoff.
+  idle,
+
+  /// The model has an active `fetchOlder` request.
+  fetchingMore,
```
chrisbobbe: This should say "an active `fetchOlder` or `fetchNewer` request", I think.

lib/model/message_list.dart
gnprice: Indeed — good catch, thanks!

lib/widgets/message_list.dart
```diff
@@ -716,16 +716,21 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
   }
 
   Widget _buildStartCap() {
-    // These assertions are invariants of [MessageListView].
```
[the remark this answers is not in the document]
gnprice: How's this version?

    And conversely, if the user does kick off a fetch at one end and then
    scroll swiftly to the other end and witness how that appears, we want
    to show them a "loading" sign.  The situation is exactly like if
    they'd had a fetch attempt on that same end and we were backing off
    from failure: there's no fetch right now, but even though a fetch is
    needed (because the user is looking at the incomplete end of the
    known history), the app will hold off from starting one right now.
    Declining to start a fetch when one is needed means effectively that
    the loading is busy.

> Maybe: "there's no fetch right now, but one is queued behind a backoff wait, so effectively the loading is busy."?

That clause after the colon is intended to be a description that applies to both situations (in order to draw the parallel between them): if there's an active fetch at the other end, or if there's an ongoing backoff from a fetch failure at the same end. So there isn't necessarily a backoff wait.

In both cases, it isn't quite true either that a fetch at this end is queued. When the loading is busy, a call to `fetchOlder` or `fetchNewer` just returns, without recording the desired fetch in any sort of queue. When a backoff completes (or when a fetch at the other end completes), what we rely on for causing a fetch to begin is instead that there'll be a new `fetchOlder` or `fetchNewer` request upon any slight movement where the user scrolls even a bit (and at the very first frame of such a scrolling motion).

lib/model/message_list.dart
gnprice: Indeed, thanks.

gnprice:
Thanks for the review! Revision pushed; PTAL.

lib/widgets/message_list.dart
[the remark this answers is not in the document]
chrisbobbe: Ah indeed; yeah. Thanks for explaining! LGTM.

chrisbobbe:
Thanks, LGTM! Merging.

Issue #1510, opened 2025-05-10
App crashes when selecting multiple images from gallery.


Vainyala17:
When selecting multiple images from the gallery, the app crashes.
This issue happens on:

- ✅ Android 11

- ✅ Android 13

- ✅ Android 15

It happens every time multiple images are selected.
Single image selection works fine.

**Steps to reproduce:**

- Tap the gallery icon.

- Select 2 or more images.

- App crashes immediately.

**Expected behavior:**
The app should allow selecting and uploading multiple images without crashing.

vagabond-0:
Can i work on these

Vainyala17:
> Can i work on these

I am working on it.

chrisbobbe:
I imagine this might be addressed by #855.

Issue #1490, opened 2025-04-29
Gallery icon allows all file types and crashes on multiple image selection (Android 11)


Vainyala17:
### Steps to Reproduce:
1. Run the Zulip app on an Android 11 device using `flutter run` .
2. Tap the Gallery icon (🖼).
3. You can select all types of files "including PDFs, videos, etc. (not limited to images)."
4. Try selecting multiple images — the app crashes and exits.

### Expected Behavior:
- Gallery icon should only allow image file selection (.jpg, .png, etc.).
- Should support multiple image selection without crashing.

### Observed Behavior:
- Gallery icon allows all file types (PDFs, videos, etc.).
- Crashes when selecting multiple images.

### Device Info:
- Android Version: 11
- Flutter Version:
Flutter 3.32.0-1.0.pre.203 • channel master • https://github.com/flutter/flutter.git
Framework • revision b73450f8f2 (6 days ago) • 2025-04-23 09:59:49 -0500
Engine • revision b73450f8f2 (6 days ago) • 2025-04-23 09:59:49 -0500
Tools • Dart 3.9.0 (build 3.9.0-44.0.dev) • DevTools 2.45.0

### Additional Info:
This issue does not occur on newer Android versions (like 13, 14 or 15).
Might be related to the `file_picker` plugin as described in:
- #855: Switch from `file_picker` to `image_picker`
- #1308: iOS issue related to video attachment

### Video:

https://github.com/user-attachments/assets/c024308c-a0eb-4190-a0ef-647a4148bcae

### I'd like to help fix this!
If maintainers agree this is a valid issue, I’d be happy to work on a fix.

gnprice:
Thanks for the report. As discussed in the chat thread [#mobile > Issue with Gallery icon behavior](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Issue.20with.20Gallery.20icon.20behavior/with/2162069), these would be good to fix.

This is two issues, not one:
* the picker shows all files, not just images and not a photo-gallery style of UI;
* the crash on selecting multiple images.

They probably have different causes, and might be fixed in different PRs at different times. One strong clue that they have different causes is that (as discussed in the chat thread) the first issue reproduces on Android 13 and Android 15 as well as Android 11, while the second issue doesn't.

So let's track the two issues separately. @Vainyala17 would you please edit this to focus on one of the two issues, and then file a separate issue for the other one?

Vainyala17:
[...]

PR #1516, opened 2025-05-18
Fix: Image picker crash and restrict selection to images only

Vainyala:
### **Purpose / Description**
Fixed the app crash that occurred when selecting multiple images using the image picker.
Restricted the file picker to only allow image files to prevent unintended crashes from unsupported file types.

### **Fixes**
Fixes app crash when picking multiple images (#1510)

Restricts file picker to images only (#1490)

### **Approach**
The fix modifies the image picker logic to handle multiple selections safely and adds validation to restrict the file picker to accept only images.


chrisbobbe:
Thanks for your work on this! Closing, though, as both the linked issues have already been fixed.

PR #1517, opened 2025-05-20
anchors n/n: Open at first unread

### LINKED ISSUE #80: Open message list at first unread, rather than latest message

### LINKED ISSUE #82: Open a message list in the middle of history, when requested

gnprice:
Fixes #80.

This is stacked atop #1566.

This is marked as a draft only because of a lack of tests. The functionality is fully implemented, and otherwise ready for review.

---

This implementation has one key difference from the legacy app and from the original plan for #80:

When fetching at first unread, it turns out the resulting /api/get-messages request can be quite slow, if the user's first unread is far back in history (e.g. in combined feed or a busy channel). This seems to be on the server side — probably it's just a naturally slower query for the database to handle.

In retrospect this is probably a major contributor (though far from the only one) to why the legacy RN app has always felt slow to me.

In this PR, fetching at first unread therefore (a) is controlled by a setting, and (b) by default is done only for conversation views, not interleaved views.

If I turn the setting to "always go to first unread", opening my combined feed on chat.zulip.org takes multiple seconds. But then if I tap the "go to bottom" button, it loads the newest messages like usual, and that is zippy like zulip-flutter has always been. (Also if I scroll around in history, even way back in the vicinity of the first unread, it remains as zippy as always.)

Without that setting (or if the default were "always go to first unread"), this would be a substantial regression relative to zulip-flutter main for users who have old unreads, and would give up a large part of this app's ability to impress users with its zippy handling of messages.

---

Previous notes about the portion that now belongs to #1566:

The main remaining work this branch needs is tests. There are also some directly user-visible aspects still to do:
* When opening a /near/ link, this PR ensures the feed opens with the target message in view. But it'd be good to visibly highlight the specific message, e.g. by having its background blink once or twice.

  That will probably be a post-launch follow-up — the lack of it isn't a regression relative to either zulip-flutter main, or the legacy app.
  (→ Filed as #1564.)

* When opening a /near/ link, this doesn't implement following where the message has been moved to.

  That's #683. Currently M6 but arguably should be M5b; either way, won't block this, and most likely won't happen before launch (which is coming soon).


gnprice:
I've just pushed a small revision with some updates to comments and commit messages, as well as rebasing atop the now-merged #1515.

With this revision, the only remaining TODOs are
* writing a bunch of tests;
* in the final commit, to add a setting as discussed above (and enable it by default).

So @chrisbobbe hopefully that helps with reading the branch.

gnprice:
A little earlier today, I split out the first portion of this, up through handling /near/ links, as #1566.

Just now I finished implementing the setting discussed previously, and pushed that as a revision here. This PR is now fully implemented, except for tests.

lib/widgets/settings.dart
```diff
@@ -85,6 +86,70 @@ class _BrowserPreferenceSetting extends StatelessWidget {
...
+        for (final value in VisitFirstUnreadSetting.values)
+          RadioListTile.adaptive(
+            title: Text(_valueDisplayName(value,
+              zulipLocalizations: zulipLocalizations)),
+            value: value,
+            // TODO(#1545) stop using the deprecated members
+            // ignore: deprecated_member_use
+            groupValue: globalSettings.visitFirstUnread,
+            // ignore: deprecated_member_use
+            onChanged: (newValue) => _handleChange(context, newValue)),
```
chrisbobbe, line 148: #1545 may be resolved soon, by PR #1568

chrisbobbe (commented):
Looks great!! Very excited for this :)

lib/widgets/settings.dart
gnprice: Yeah. So I guess whichever of these lands first, there'll be a bit of a nonlocal merge conflict to resolve by applying that migration to this new bit of code.

gnprice:
OK, updated to make this ready to merge. I did the same tactic as in #1566: I filed a follow-up issue for writing tests:
* #1571

gnprice:
And merged. Thanks for the review!

PR #1520, opened 2025-05-20
Update translations from Weblate with app_zh.arb added, support zh-Hant-TW

PIG208:
This adds app_zh.arb so that locales like "zh-Hans" can have their
code generated.

This is because `flutter gen-l10n` requires a base locale to be present.
Only having "zh-Hans" gives us the following error.

```
  Arb file for a fallback, zh, does not exist, even though
  the following locale(s) exist: [zh_Hans].
  When locales specify a script code or country code, a
  base locale (without the script code or country code) should
  exist as the fallback. Please create a {fileName}_zh.arb
  file.
```

"zh" will not appear on Weblate, though, because we exclude it from
settings.  This is intentional since it only makes sense to translate
the more specific locales like "zh-Hans" and "zh-Hant-TW".

CZO discussion:
  https://chat.zulip.org/#narrow/channel/58-translation/topic/zh_*.20in.20Weblate/near/2174667


gnprice:
Cool, thanks.

I just sent #1522 as a pure cherry-pick of the commit that went into v0.0.29. If we merge that first, then that will facilitate comparing v0.0.29 with upcomng versions in main; it should also simplify this change somewhat by taking care of most of the unrelated changes that had accumulated.

PIG208:
This is also an experiment of how Weblate processes new languages we added. I suppose `ar` (Arabic) was added to Weblate from our GitHub remote as opposed to the other way around. This time we will see if Weblate successfully ignores `zh` with a [language filter](https://docs.weblate.org/en/latest/admin/projects.html#component-language-regex) we set up for this.

gnprice:
> I suppose `ar` (Arabic) was added to Weblate from our GitHub remote as opposed to the other way around.

It was, but as part of the initial setup of Weblate — the file was added last July, and I believe it was a few months later that we first set up Weblate.

More recently, though, we added `nb` in our repo and let Weblate pick it up from there, and that seemed to work: [#mobile-team > weblate translations (#F276) @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/weblate.20translations.20.28.23F276.29/near/2039979) (That one was complicated partly because it involved an alias to connect that to a translation Weblate itself calls `nb_NO`: [#mobile-team > weblate translations (#F276) @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/weblate.20translations.20.28.23F276.29/near/2035361).)

Issue #1456, opened 2025-04-02
Disallow connecting to server 4.x


gnprice:
This is a reduced, pre-launch version of:
* #268 

The Zulip Server 5 release went out [over 3 years ago](https://blog.zulip.com/2022/03/29/zulip-5-0-released/), so the 4.x releases are all long out of our announced [18-month compatibility window](https://zulip.readthedocs.io/en/latest/overview/release-lifecycle.html#client-apps).

As #268 describes, there are already a few places where we rely on the API of Zulip Server 5 and later. If someone tries using the app on a server older than that, then instead of quietly not working correctly it'd be better to give a clear error. Now that #267 is complete, it'll be straightforward to do that.

(The difference between this issue and #268 is that the latter calls for taking advantage of the new minimum version in order to clear out all the `TODO(server-5)` items in our codebase. This issue is purely for updating the minimum version set by #267, so as to give the user an error when they're on such an old version.)

### References
* This issue is a followup to: #267 
* A fuller version of this issue, to be done post-launch: #268

gnprice:
I was about to go also file an issue for us to have the legacy app start warning people that older than Server 5 isn't supported and they should (get their admins to) upgrade… but happily we already did that back in 2023 (thanks @chrisbobbe!):
* https://github.com/zulip/zulip-mobile/pull/5762

The [changelog says](https://github.com/zulip/zulip-mobile/blob/4d9d46d7f852af3fb4d44463654a141141e7bac3/docs/changelog.md#27216-2023-10-10) that went out to all users in v27.216 and earlier, dated 2023-10-10.

gnprice:
This issue isn't yet ripe to carry out (at least not to merge a PR doing it): out of an abundance of caution, I'd like to first get out a beta release containing #267 to disallow servers older than 4.x, and only then merge this change so that it goes out in the next or later beta release.

chrisbobbe:
> the legacy app

I've just sent zulip/zulip-mobile#5920 to nudge more server admins via the legacy app :)

chrisbobbe:
And https://github.com/zulip/zulip-mobile/pull/5922 was merged, to nudge even more server admins via the legacy app :)—when that's released, the nag banner will appear through servers 7.x. It's included in today's v27.235 beta release, which should go out to all users in a few days.

PR #1521, opened 2025-05-21
api: Like zulip-mobile, don't allow connecting to Zulip Server <7.0

chrisbobbe:
This goes rather farther than #1456, which is for bumping this number just to 5.0. This is OK; we recently bumped zulip-mobile to use 7.0, in the v27.235 release.

zulip-mobile shows a "nag banner" to nudge server admins to upgrade. As of v27.235 (PR zulip/zulip-mobile#5922), that banner appears on 7.x servers, the latest 7.x having recently aged out of our 18-month support window. So we should feel comfortable nudging this number in zulip-flutter to 8.0 once v27.235 has been out for a little while.

Fixes: #1456

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,7 +289,7 @@ good time to [report them as issues][dart-test-tracker].
 
 #### Server compatibility
 
-We support Zulip Server 4.0 and later.
+We support Zulip Server 7.0 and later.
 
 For API features added in newer versions, use `TODO(server-N)`
 comments (like those you see in the existing code.)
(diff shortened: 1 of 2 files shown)


chrisbobbe:
I added a

```dart
// TODO(#268) address all TODO(server-5), TODO(server-6), and TODO(server-7)
```

even though #268 really only corresponds to `TODO(server-5)`s. I can update or replace #268 once this is merged, if that's helpful.

gnprice:
Thanks! Looks good; merging.

I think it'll be cleanest to let #268 remain for just the `TODO(server-5)`s, and have separate issues for server-6 and server-7. That'll let us separately track fixing one and fixing the next, which might happen in separate PRs.

It's fine for the TODO to refer initially to just #268, though. When we go to handle the `TODO(server-5)`s, we'll naturally be acting on that line, and we can update it to refer to the successor issues then.

PR #1522, opened 2025-05-21
l10n: Update translations from Weblate, except skip zh_Hans

gnprice:
Cherry-picked from 54664523686a1fa599bab644b515a380853d90e1, which was in the v0.0.29 preview beta.  The cherry-pick required fixup only in rerunning `tools/check l10n --fix`, which changed the set of English fallback strings included in the new `de` translation.

This update required adjustment; straight from the Weblate branch, l10n generation (`tools/check l10n --fix`) failed:

    Arb file for a fallback, zh, does not exist, even though
    the following locale(s) exist: [zh_Hans].
    When locales specify a script code or country code, a
    base locale (without the script code or country code) should
    exist as the fallback. Please create a {fileName}_zh.arb
    file.

The app_zh_Hans.arb file had only two translations in it; it exists at the moment precisely in order to test this sort of workflow. For the release, I just removed it and reran generation. (We'll include it properly in a future change.)


PIG208:
Thanks! While I do see some changes in the Weblate remote since this was written, I think we will just synchronize translations from there after merging this.

PIG208:
Merged. Now I will try to get the Weblate remote up-to-date, since this has some conflicts with the recent updates I mentioned above.

PR #1500, opened 2025-05-06
Add topic-list page

PIG208:
Stacked on top of #1491.

Some non-goals of this change are deferred to #1499. In this implementation, we fetch the topics but do not handle all events to receive live-updates. It's expected that when topics are resolved/unresolved or moved, or when new messages arrived, the changes to the topic-list will not be seen until the next fetch.

We also skip thinning down the app bar, since that will require work on app bars on message-list page too.

The PR is structured to encourage side-by-side comparison with similar existing code. Namely `_TopicItem` from `lib/widgets/inbox.dart` and `MessageListAppBarTitle`.

Fixes: #1158

<details>
<summary>
screenshots (taken on my Android device, hence the left-aligned app bar!)
</summary>

| | light | dark |
| - | - | - |
| regular | image | image |
| unsubscribed | image | image |
| unknown channel | image | image |

| small | regular | large |
| - | - | - |
| image | image | image |

| message-list | channel action sheet |
| - | - |
| image | image |

| non-muted topic | muted topic |
| - | - |
| image | image |

`debugDefaultTargetPlatformOverride = TargetPlatform.iOS`:

</details>

PR #1524, opened 2025-05-21
msglist: Colorize channel icon in the app bar, following Figma

chrisbobbe:
Following the Figma:
  https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6089-28394&m=dev

And while we're adding a test for it, also check that the chosen channel icon is the intended one.


chrisbobbe:
This will avoid an inconsistency with the topic-list page (PR #1500), which colorizes the icon.

chrisbobbe:
| Before | After |
| --- | --- |
| image | image |

lib/widgets/message_list.dart
```diff
@@ -297,17 +297,26 @@ class MessageListAppBarTitle extends StatelessWidget {
...
     final zulipLocalizations = ZulipLocalizations.of(context);
+
     // A null [Icon.icon] makes a blank space.
-    final icon = stream != null ? iconDataForStream(stream) : null;
+    IconData? icon;
+    Color? iconColor;
+    if (stream != null) {
+      icon = iconDataForStream(stream);
+      iconColor = colorSwatchFor(context, store.subscriptions[stream.streamId])
+        .iconOnBarBackground;
```
PIG208, line 309: Should we fall back to the default base color `kDefaultChannelColorSwatchBaseColor` (by passing a possibly null subscription) until we get to fully implementing #188?

test/widgets/message_list_test.dart
```diff
@@ -208,6 +210,36 @@ void main() {
...
+          of: find.byType(ZulipAppBar),
+          matching: find.byIcon(expectedIcon)));
+
+        check(Theme.brightnessOf(iconElement)).equals(Brightness.light);
+        check(iconElement.widget as Icon).color.equals(Color(0xff5972fc));
+      });
+    }
+    testChannelIconInChannelRow(ZulipIcons.globe, isWebPublic: true, inviteOnly: false);
+    testChannelIconInChannelRow(ZulipIcons.lock, isWebPublic: false, inviteOnly: true);
+    testChannelIconInChannelRow(ZulipIcons.hash_sign, isWebPublic: false, inviteOnly: false);
```
PIG208, line 241: Thanks! It's nice to have this tested.

PIG208 (commented):
Thanks! Overall looks good to me. Left some comments. Looks like the CI is failing, but it should be an easy fix.

lib/widgets/message_list.dart
chrisbobbe: I think that's what this code already does; am I reading too quickly? `store.subscriptions[stream.streamId]` might be null, and if it is, `colorSwatchFor` uses a swatch based on `kDefaultChannelColorSwatchBaseColor`.

PIG208: Oh right. I think I meant the case when `stream` is `null`, i.e. we got a "(unknown channel)".

PR #1526, opened 2025-05-22
android build: Disable lint AndroidGradlePluginVersion

chrisbobbe:
android build: Disable lint AndroidGradlePluginVersion
    
    We're successfully managing our Gradle and AGP upgrades without this
    lint rule, and it's annoying that it fails our CI builds.
    Discussion:
      https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/CI.20fail.3A.20lint.20on.20Gradle.20version.3F/near/2179323


gnprice:
Thanks for taking care of this! Looks good; merged.

PR #1535, opened 2025-05-28
local echo (6.5/7): Revised prep commits for local echo

chrisbobbe:
Split from #1453, with some small adjustments. This leaves the two last commits from #1453—

c8295ae9b msglist: Adjust edit-message status row bottom padding

—for later.

Here are before/after screenshots for the commit that adjusts padding:

3f8a10d80 msglist: Put edit-message progress bar in top half of 4px bottom padding

| Before | After |
| --- | --- |
| image | image |


lib/widgets/message_list.dart
```diff
@@ -1580,33 +1585,34 @@ class _EditMessageStatusRow extends StatelessWidget {
...
+            //   https://github.com/zulip/zulip-flutter/pull/1498#discussion_r2087576108
+            LinearProgressIndicator(
+              minHeight: 2,
+              color: designVariables.foreground.withValues(alpha: 0.5),
+              backgroundColor: designVariables.foreground.withValues(alpha: 0.2),
+            ),
+          ]),
+        true => _RestoreEditMessageGestureDetector(
+          messageId: messageId,
+          child: Text(
```
gnprice, line 1610: This case should also get 4px of bottom padding, right?

(Maybe this change would be clearer as first an NFC commit that just distributes the logic for the bottom padding among the different cases, followed by a commit that removes 2px of padding for the one case where that's intended.)

chrisbobbe:

PR #1538, opened 2025-05-30
local echo (6.75/7): Revised commit for local echo (sans retrieving failed outbox content)

chrisbobbe:
Here's a revised version of a substantive commit from #1453:

9c0b95568 msglist: Add and manage outbox message objects in message list view

That just leaves one more substantive commit from that PR:

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -66,6 +66,22 @@ class MessageListMessageItem extends MessageListMessageBaseItem {
   });
 }
 
+/// An [OutboxMessage] to show in the message list.
+class MessageListOutboxMessageItem extends MessageListMessageBaseItem {
+  @override
+  final OutboxMessage message;
+  @override
+  final ZulipContent content;
+
+  MessageListOutboxMessageItem(
+    this.message, {
+    required super.showSender,
+    required super.isLastInBlock,
+  }) : content = ZulipContent(nodes: [
+    ParagraphNode(links: null, nodes: [TextNode(message.contentMarkdown)]),
+  ]);
+}
+
 /// The status of outstanding or recent fetch requests from a [MessageListView].
 enum FetchingStatus {
   /// The model has not made any fetch requests (since its last reset, if any).
@@ -158,14 +174,24 @@ mixin _MessageSequence {
   /// It exists as an optimization, to memoize the work of parsing.
   final List<ZulipMessageContent> contents = [];
 
+  /// The [OutboxMessage]s sent by the self-user, retrieved from
+  /// [MessageStore.outboxMessages].
+  ///
+  /// See also [items].
+  ///
+  /// O(N) iterations through this list are acceptable
+  /// because it won't normally have more than a few items.
+  final List<OutboxMessage> outboxMessages = [];
+
   /// The messages and their siblings in the UI, in order.
   ///
   /// This has a [MessageListMessageItem] corresponding to each element
   /// of [messages], in order.  It may have additional items interspersed
-  /// before, between, or after the messages.
+  /// before, between, or after the messages. Then, similarly,
+  /// [MessageListOutboxMessageItem]s corresponding to [outboxMessages].
   ///
-  /// This information is completely derived from [messages] and
-  /// the flags [haveOldest], [haveNewest], and [busyFetchingMore].
+  /// This information is completely derived from [messages], [outboxMessages],
+  /// and the flags [haveOldest], [haveNewest], and [busyFetchingMore].
   /// It exists as an optimization, to memoize that computation.
   ///
   /// See also [middleItem], an index which divides this list
@@ -177,11 +203,14 @@ mixin _MessageSequence {
   /// The indices 0 to before [middleItem] are the top slice of [items],
   /// and the indices from [middleItem] to the end are the bottom slice.
   ///
-  /// The top and bottom slices of [items] correspond to
-  /// the top and bottom slices of [messages] respectively.
-  /// Either the bottom slices of both [items] and [messages] are empty,
-  /// or the first item in the bottom slice of [items] is a [MessageListMessageItem]
-  /// for the first message in the bottom slice of [messages].
+  /// The top slice of [items] corresponds to the top slice of [messages].
+  /// The bottom slice of [items] corresponds to the bottom slice of [messages]
+  /// plus any [outboxMessages].
+  ///
+  /// The bottom slice will either be empty
+  /// or start with a [MessageListMessageBaseItem].
+  /// It will not start with a [MessageListDateSeparatorItem]
+  /// or a [MessageListRecipientHeaderItem].
   int middleItem = 0;
 
   int _findMessageWithId(int messageId) {
@@ -197,9 +226,10 @@ mixin _MessageSequence {
     switch (item) {
       case MessageListRecipientHeaderItem(:var message):
       case MessageListDateSeparatorItem(:var message):
-        if (message.id == null)                  return 1;  // TODO(#1441): test
+        if (message.id == null)                  return 1;
         return message.id! <= messageId ? -1 : 1;
       case MessageListMessageItem(:var message): return message.id.compareTo(messageId);
+      case MessageListOutboxMessageItem():       return 1;
     }
   }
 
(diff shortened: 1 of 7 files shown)


lib/model/message_list.dart
```diff
@@ -298,10 +328,46 @@ mixin _MessageSequence {
...
+    _reprocessOutboxMessages();
+    return true;
+  }
+
   /// Reset all [_MessageSequence] data, and cancel any active fetches.
   void _reset() {
     generation += 1;
     messages.clear();
+    outboxMessages.clear();
     middleMessage = 0;
```
gnprice: nit: keep `messages` and `middleMessage` together (like they are in their declarations)
```suggestion
    messages.clear();
    middleMessage = 0;
    outboxMessages.clear();
```

lib/model/message_list.dart
```diff
@@ -384,13 +450,67 @@ mixin _MessageSequence {
...
+  /// content of [items].
+  ///
+  /// This is efficient due to the expected small size of [outboxMessages].
+  void _removeOutboxMessageItems() {
+    // This loop relies on the assumption that all items that follow
+    // the last [MessageListMessageItem] are derived from outbox messages.
+    while (items.isNotEmpty && items.last is! MessageListMessageItem) {
+      items.removeLast();
+    }
+    assert(items.none((e) => e is MessageListOutboxMessageItem));
```
gnprice: This will loop through the entire list of items, which may be long with a lot of message history.

Let's avoid that kind of expensive check even in debug builds of the app — the message list already feels somewhat sluggish in debug mode. Instead we rely on the tests and their `checkInvariants` function to confirm these invariants are satisfied.

lib/model/message_list.dart
```diff
@@ -537,12 +659,14 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
 
     // We'll make the bottom slice start at the last visible message, if any.
+    _removeOutboxMessageItems();
     for (final message in result.messages) {
       if (!_messageVisible(message)) continue;
       middleMessage = messages.length;
       _addMessage(message);
       // Now [middleMessage] is the last message (the one just added).
     }
+    _reprocessOutboxMessages();
```
gnprice: It looks like this logic will have `outboxMessages` get filled in before the fetch even completes. That feels kind of odd. This is a view-model for the message list, and what we actually show in the message list in that state is just a loading indicator and no messages.

And in the future with #82, we won't necessarily want them even after the fetch, depending on the value that `haveNewest` ends up with after the fetch. (Just like you pointed out at https://github.com/zulip/zulip-flutter/pull/1515#discussion_r2118295040 for new actual messages.)

So how about we have `_syncOutboxMessages` get called only here in the fetch-result handler, rather than before the fetch? Then in #1515 we can straightforwardly make it conditional on `haveNewest`, and do the same in `fetchNewer`.

Probably that's best done as a separate commit on top, to facilitate merging the main commit.

lib/model/message_list.dart
```diff
@@ -666,10 +825,17 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
-        if (_removeMessagesWhere((message) =>
-            (message is StreamMessage
-             && message.streamId == event.streamId
-             && message.topic == event.topicName))) {
+        bool removed = _removeOutboxMessagesWhere((message) =>
+          message is StreamOutboxMessage
+          && message.conversation.streamId == event.streamId
+          && message.conversation.topic == event.topicName);
+
+        removed |= _removeMessagesWhere((message) =>
```
gnprice: nit: handle the normal messages first, since they come earlier in the list

lib/model/message_list.dart
```diff
@@ -682,6 +848,18 @@ class MessageListView with ChangeNotifier, _MessageSequence {
           notifyListeners();
           fetchInitial();
         }
+
+        outboxMessages.clear();
+        for (final outboxMessage in store.outboxMessages.values) {
+          if (_shouldAddOutboxMessage(
+            outboxMessage,
+            wasUnmuted: outboxMessage is StreamOutboxMessage
+              && outboxMessage.conversation.streamId == event.streamId
```
gnprice: I think this whole stanza goes away if we let `fetchInitial` take care of syncing in the outbox messages when appropriate.

lib/model/message_list.dart
```diff
@@ -695,14 +873,34 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+      return;
+    }
+
+    // We always remove all outbox message items
+    // to ensure that message items come before them.
+    _removeOutboxMessageItems();
+    // TODO insert in middle of [messages] instead, when appropriate
     _addMessage(message);
+    if (event.localMessageId != null) {
+      final localMessageId = int.parse(event.localMessageId!);
```
gnprice: `int.parse` should always use explicit `radix`

test/model/message_test.dart
```diff
@@ -168,32 +168,40 @@ void main() {
...
       checkState().equals(OutboxMessageState.hidden);
+      checkNotNotified();
 
       async.elapse(kLocalEchoDebounceDuration);
       checkState().equals(OutboxMessageState.waiting);
+      checkNotifiedOnce();
 
       await receiveMessage(eg.dmMessage(from: eg.selfUser, to: [eg.otherUser]));
       check(store.outboxMessages).isEmpty();
+      checkNotifiedOnce();
```
gnprice, line 180: Maybe add all these checks as a prep commit? Then if there are any that change with the main change (or if there are none), that will be more visible.

test/model/message_list_test.dart
```diff
@@ -2244,8 +2813,16 @@ void checkInvariants(MessageListView model) {
   check(model).middleItem
     ..isGreaterOrEqual(0)
     ..isLessOrEqual(model.items.length);
-  if (model.middleItem == model.items.length) {
-    check(model.middleMessage).equals(model.messages.length);
+  if (model.middleMessage == model.messages.length) {
+    if (model.outboxMessages.isEmpty) {
+      // the bottom slice of `model.messages` is empty
+      check(model.items).length.equals(model.middleItem);
```
gnprice: nit: this check is really about validating the behavior of `middleItem`, not `items`
```suggestion
      check(model.middleItem).equals(model.items.length);
```

test/model/message_list_test.dart
```diff
@@ -2244,8 +2813,16 @@ void checkInvariants(MessageListView model) {
...
+    } else {
+      check(model.items.sublist(model.middleItem))
+        ..first.isA<MessageListOutboxMessageItem>()
+            .message.identicalTo(model.outboxMessages.first)
+        ..every((item) => item.not((it) => it.isA<MessageListMessageItem>()));
```
gnprice: This "is not MessageListMessageItem" check is redundant with the "is MessageListOutboxMessageItem" check just above, given the overall structure of `items`: all items for outbox-messages come after all items for normal messages. And that latter structure was verified in the big `for (int j = 0; j < allMessages.length; j++)` loop above.

So this `every` can be dropped. And then the `.sublist` plus `.first` can be simplified down.

gnprice (commented):
Thanks! I've read through all the non-test changes and part of the tests. Generally this looks good; comments below.

lib/model/message_list.dart
chrisbobbe: Makes sense; I agree.

> Probably that's best done as a separate commit on top, to facilitate merging the main commit.

Hmm, since #1515 is already merged, the main commit will need to check `haveNewest` before adding outbox messages to `items`.

chrisbobbe:
Thanks for the review! Revision pushed. This revision postpones copying `MessageStore.outboxMessages` to the message-list's `outboxMessages` until `haveNewest` becomes true. It also doesn't add new, just-sent outbox messages unless `haveNewest` is true.

lib/model/message_list.dart
```diff
@@ -756,9 +886,42 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+  }
+
+  /// Reads [MessageStore.outboxMessages] and copies to [outboxMessages]
+  /// the ones belonging to this view.
+  ///
+  /// This should only be called when [haveNewest] is true
+  /// because outbox messages are considered newer than regular messages.
+  ///
+  /// This does not call [notifyListeners].
+  void _syncOutboxMessagesFromStore() {
```
gnprice, line 917: Cool, I like this name change, too (adding "from store").

test/model/message_list_test.dart
```diff
@@ -2459,8 +3042,15 @@ void checkInvariants(MessageListView model) {
...
-  if (model.middleItem == model.items.length) {
-    check(model.middleMessage).equals(model.messages.length);
+  if (model.middleMessage == model.messages.length) {
+    if (model.outboxMessages.isEmpty) {
+      // the bottom slice of `model.messages` is empty
+      check(model).middleItem.equals(model.items.length);
+    } else {
+      check(model.items[model.middleItem])
+        .isA<MessageListOutboxMessageItem>()
+          .message.identicalTo(model.outboxMessages.first);
```
gnprice: nit: can make this a bit more visibly parallel to the normal-message case just below it:
```suggestion
      check(model.items[model.middleItem]).isA<MessageListOutboxMessageItem>()
        .message.identicalTo(model.outboxMessages.first);
```

test/model/message_list_test.dart
```diff
@@ -748,18 +1116,30 @@ void main() {
...
+      check(model).outboxMessages.isEmpty();
 
       connection.prepare(
         json: newestResult(foundOldest: true, messages: messages).toJson());
       await setVisibility(UserTopicVisibilityPolicy.unmuted);
       checkNotifiedOnce();
-      check(model).fetched.isFalse();
+      check(model)
+        ..fetched.isFalse()
+        ..outboxMessages.isEmpty();
```
gnprice: This second check is implied by the first one plus an invariant, right?

Hmm, we could make that invariant more explicitly checked, though, in checkInvariants:
```diff
   if (!model.fetched) {
     check(model)
       ..messages.isEmpty()
+      ..outboxMessages.isEmpty()
```

… I guess the existing test explicitly checks `messages` is empty, even though that's similarly redundant, because it's helpful for parallelism with the steps above and below. So this check makes sense to include for the same reason. For that same parallelism, though, it's probably clearest to leave it as a separate `check` call that comes after the `checkHasMessageIds`, rather than consolidate with the check here on `fetched`.

test/model/message_list_test.dart
```diff
@@ -246,6 +276,105 @@ void main() {
...
+        foundOldest: true,
+        foundNewest: false,
+        messages: [message]).toJson());
+      await model.fetchInitial();
+      checkNotifiedOnce();
+      check(model)..fetched.isTrue()..haveNewest.isFalse()..outboxMessages.isEmpty();
+
+      connection.prepare(json: newestResult(foundOldest: false,
+        messages: [eg.streamMessage(stream: stream, topic: 'topic')]).toJson());
+      final fetchFuture = model.fetchNewer();
```
gnprice: nit for realism: use `newerResult` (the `anchor` will be different)

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

Issue #1543, opened 2025-06-03
msglist: When the initial message fetch comes up empty, auto-focus the compose box


chrisbobbe:
Part of #127 (see below).

Normally, it would be bothersome if we auto-focused the compose box when opening a message-list page, because the keyboard takes up a lot of space when it appears, and that gets in the way of reading messages. Opening the keyboard isn't a problem when there are no messages, though, and it can actually be quite helpful—especially when opening a new DM narrow with no history.

Implementing this feature would address item 6 in Vlad's feedback about the new-DM UI:
- https://github.com/zulip/zulip-flutter/pull/1322#issuecomment-2917492437
- [#mobile > Flutter DMs flow design review @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20DMs.20flow.20design.20review/near/2181176)

> ### 6. Keyboard should be opened once transitioned to compose dialog.
> It is ok to close keyboard if user clicked outside and input is empty, but by default keyboard should be opened and input is focused for me to type a message

Issue #1448, opened 2025-03-28
Focus topic input first when opening compose box


gnprice:
In a channel feed, if you tap the compose box and the topic input is empty (i.e. you haven't already entered something there), we'd like to put focus in the topic input instead of in the content input.

Chat thread here: [#mobile > reply goes to "no topic" @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/reply.20goes.20to.20.22no.20topic.22/near/2138196)

### Background / motivation

This bit of user feedback [started that same thread](https://chat.zulip.org/#narrow/channel/48-mobile/topic/reply.20goes.20to.20.22no.20topic.22/near/2133203):
> Hello, a new user / first time poster with some hopefully useful feedback : we moved a small organisation to Zulip recently and the thing we keep banging our heads on is accidentally posting to [no topic] when we meant to reply to a comment in some topic.

In the legacy app we effectively have the behavior this issue is calling for: when the compose box is closed, the topic input doesn't even appear, but once you touch it so it opens, your focus goes to the topic input.

### References
* [#mobile > reply goes to "no topic" @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/reply.20goes.20to.20.22no.20topic.22/near/2138196)
* An issue that may make this change less effective, until it's resolved:
  #226

shkshreyas:
About improving the user experience in the Zulip mobile app by changing the focus behavior when opening the compose box. my solution for this issue.

# Solution for Compose Box Focus Issue

The issue requires changing the focus behavior when tapping the compose box so that if the topic input is empty, focus should be placed on the topic input instead of the content input.

This change would help prevent users from accidentally posting messages with "no topic" when they intended to reply to a specific topic.

Here's how we can implement this solution:

```dart:c:\Users\shksh\Desktop\OS\oppia-android\lib\compose\compose_box.dart
// ... existing code ...

class ComposeBox extends StatefulWidget {
  // ... existing code ...
}

class ComposeBoxState extends State<ComposeBox> {
  // ... existing code ...
  
  // Add or modify the method that handles tapping on the compose box
  void _handleComposeBoxTap() {
    // Check if the topic input is empty
    if (_topicController.text.isEmpty) {
      // If empty, focus on the topic input
      FocusScope.of(context).requestFocus(_topicFocusNode);
    } else {
      // If topic already has content, focus on the message input as before
      FocusScope.of(context).requestFocus(_messageFocusNode);
    }
[...]

PR #1544, opened 2025-06-03
msglist: When initial message fetch comes up empty, auto-focus compose box

chrisbobbe:
This is part of our plan to streamline the new-DM UI: when you start a new DM conversation with no history, we should auto-focus the content input in the compose box.

Fixes: #1543

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1546,6 +1546,15 @@ sealed class ComposeBoxController {
   final content = ComposeContentController();
   final contentFocusNode = FocusNode();
 
+  /// If no input is focused, requests focus on the appropriate input.
+  ///
+  /// A convenience method to encapsulate choosing the topic or content input
+  /// when both exist (see [StreamComposeBoxController.requestFocusIfUnfocused]).
+  void requestFocusIfUnfocused() {
+    if (contentFocusNode.hasFocus) return;
+    contentFocusNode.requestFocus();
+  }
+
   @mustCallSuper
   void dispose() {
     content.dispose();
@@ -1608,6 +1617,15 @@ class StreamComposeBoxController extends ComposeBoxController {
   final topicFocusNode = FocusNode();
   final ValueNotifier<ComposeTopicInteractionStatus> topicInteractionStatus =
     ValueNotifier(ComposeTopicInteractionStatus.notEditingNotChosen);
+
+  @override void requestFocusIfUnfocused() {
+    if (topicFocusNode.hasFocus || contentFocusNode.hasFocus) return;
+    if (topicInteractionStatus.value == ComposeTopicInteractionStatus.notEditingNotChosen) {
+      topicFocusNode.requestFocus();
+    } else {
+      contentFocusNode.requestFocus();
+    }
+  }
 
   @override
   void dispose() {
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:checks/checks.dart';
+import 'package:collection/collection.dart';
 import 'package:crypto/crypto.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:flutter_checks/flutter_checks.dart';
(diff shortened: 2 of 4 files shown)


chrisbobbe:
(The CI failure is #1545; I've sent #1546 for that.)

lib/widgets/compose_box.dart
```diff
@@ -1609,6 +1618,15 @@ class StreamComposeBoxController extends ComposeBoxController {
...
     ValueNotifier(ComposeTopicInteractionStatus.notEditingNotChosen);
 
+  @override void requestFocusIfUnfocused() {
+    if (topicFocusNode.hasFocus || contentFocusNode.hasFocus) return;
+    if (topicInteractionStatus.value == ComposeTopicInteractionStatus.notEditingNotChosen) {
+      topicFocusNode.requestFocus();
+    } else {
+      contentFocusNode.requestFocus();
+    }
+  }
```
chrisbobbe: This function could also be useful for #1448.

lib/widgets/compose_box.dart
```diff
@@ -1546,6 +1546,15 @@ sealed class ComposeBoxController {
   final content = ComposeContentController();
   final contentFocusNode = FocusNode();
 
+  /// If no input is focused, requests focus on the appropriate input.
+  ///
+  /// A convenience method to encapsulate choosing the topic or content input
+  /// when both exist (see [StreamComposeBoxController.requestFocusIfUnfocused]).
```
gnprice: nit: I'd think of a "convenience method" as doing something the caller could have perfectly reasonably done for itself, but which just makes it a bit easier or shorter at the caller. Because this is encapsulating meaningful logic of its own (at least in the subclass implementation), I wouldn't call it a "convenience method".

```suggestion
  /// This encapsulates choosing the topic or content input
  /// when both exist (see [StreamComposeBoxController.requestFocusIfUnfocused]).
```

lib/widgets/compose_box.dart
```diff
@@ -1609,6 +1618,15 @@ class StreamComposeBoxController extends ComposeBoxController {
   final ValueNotifier<ComposeTopicInteractionStatus> topicInteractionStatus =
     ValueNotifier(ComposeTopicInteractionStatus.notEditingNotChosen);
 
+  @override void requestFocusIfUnfocused() {
+    if (topicFocusNode.hasFocus || contentFocusNode.hasFocus) return;
+    if (topicInteractionStatus.value == ComposeTopicInteractionStatus.notEditingNotChosen) {
+      topicFocusNode.requestFocus();
+    } else {
+      contentFocusNode.requestFocus();
```
gnprice: Let's make this a `switch` — seems like one wants to explicitly think through each of the values of ComposeTopicInteractionStatus in order to understand this logic.

test/widgets/compose_box_test.dart
```diff
@@ -134,6 +148,61 @@ void main() {
...
 
+  group('auto focus', () {
+    testWidgets('ChannelNarrow, non-empty fetch', (tester) async {
+      final channel = eg.stream();
+      await prepareComposeBox(tester,
+        narrow: ChannelNarrow(channel.streamId),
+        streams: [channel],
+        messages: [eg.streamMessage(stream: channel)]);
+      check(controller).isA<StreamComposeBoxController>()
+        .topicFocusNode.hasFocus.isFalse();
```
gnprice: Can make this check a bit stronger, verifying focus doesn't go to the content input either:
```suggestion
      check(controller).isA<StreamComposeBoxController>()
        ..topicFocusNode.hasFocus.isFalse()
        ..contentFocusNode.hasFocus.isFalse();
```

test/widgets/compose_box_test.dart
```diff
@@ -134,6 +148,61 @@ void main() {
...
+        streams: [channel],
+        messages: []);
+      check(controller).isNotNull().contentFocusNode.hasFocus.isTrue();
+    });
+
+    testWidgets('DmNarrow, non-empty fetch', (tester) async {
+      final user = eg.user();
+      await prepareComposeBox(tester,
+        narrow: DmNarrow.withUser(user.userId, selfUserId: store.selfUserId),
+        messages: [eg.dmMessage(from: user, to: [store.selfUser])]);
```
gnprice: `store` is set up by `prepareComposeBox` itself, so these end up using the state left behind by the previous test case. You can see the effect by running the case in isolation:
```
$ flutter test test/widgets/compose_box_test.dart --name DmNarrow
00:06 +0: auto focus DmNarrow, non-empty fetch                                   
The following LateError was thrown running a test:
LateInitializationError: Local 'store' has not been initialized.

#0      LateError._throwLocalNotInitialized (dart:_internal-patch/internal_patch.dart:214:5)
#1      main.<anonymous closure>.<anonymous closure> (file:///home/greg/z/flutterz/test/widgets/compose_box_test.dart)
[…]
```

Looks like `prepareComposeBox` takes a `selfUser` argument, so that can be used to control this.

gnprice (commented):
Thanks! Looks good; some small comments below.

chrisbobbe:
Thanks! Revision pushed.

test/widgets/compose_box_test.dart
```diff
@@ -134,6 +148,63 @@ void main() {
...
+      await prepareComposeBox(tester,
+        selfUser: eg.selfUser,
+        narrow: DmNarrow.withUser(user.userId, selfUserId: eg.selfUser.userId),
+        messages: [eg.dmMessage(from: user, to: [eg.selfUser])]);
+      check(controller).isNotNull().contentFocusNode.hasFocus.isFalse();
+    });
+
+    testWidgets('DmNarrow, empty fetch', (tester) async {
+      await prepareComposeBox(tester,
+        narrow: DmNarrow.withUser(eg.user().userId, selfUserId: store.selfUserId),
```
gnprice: same story as https://github.com/zulip/zulip-flutter/pull/1544#discussion_r2139165959 :slightly_smiling_face:

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merged.

Issue #1545, opened 2025-06-03
Resolve deprecation warnings about radio buttons


chrisbobbe:
This has come up in CI recently:

```
Running analyze...
Analyzing zulip-flutter...                                      

   info • 'groupValue' is deprecated and shouldn't be used. Use a RadioGroup ancestor to manage group value instead. This feature was deprecated after v3.32.0-0.0.pre • lib/widgets/settings.dart:56:13 • deprecated_member_use
   info • 'onChanged' is deprecated and shouldn't be used. Use RadioGroup to handle value change instead. This feature was deprecated after v3.32.0-0.0.pre • lib/widgets/settings.dart:57:13 • deprecated_member_use
   info • 'checked' is deprecated and shouldn't be used. Use RadioGroup.groupValue to find which radio is checked. This feature was deprecated after v3.32.0-0.0.pre • test/flutter_checks.dart:254:45 • deprecated_member_use

3 issues found. (ran in 28.0s)
```

We suppressed the warnings, in #1562, but we should resolve them instead. See #1546 for an approach that effectively resolved them, but required upgrading Flutter past a breaking change.

PR #1546, opened 2025-06-03
settings: Upgrade Flutter; migrate to new RadioGroup API

chrisbobbe:
Fixes: #1545


chrisbobbe:
Hmm actually—it looks like the Info.plist thing is a breaking change: flutter/flutter@3d8fa621c

From my reading of the commit and the migration guide linked from there, it looks like it wouldn't break our `main` but it would interfere with #1379; is that right?

rajveermalviya:
Right, also looks like the migration/fix for that is going to be a bit involved. (See discussion here: https://discord.com/channels/608014603317936148/1380539391671275662)

Also, I [saw a report](https://github.com/flutter/flutter/pull/169365#issuecomment-2954471369) of deep links being broken, which led me to test the web auth flow in our app on iOS Simulator and unfortunately it turns out that we are affected by it too (verified by switching back and forth between that and the parent commit).  I'll file an upstream issue (in a bit, with a reproducer) if the reporter doesn't already have it by then.

Edit: Filed https://github.com/flutter/flutter/issues/170350.

chrisbobbe:
Thanks for investigating!

gnprice:
Given the breakage of web auth, let's plan to not upgrade Flutter past that commit until after the launch. (Even if it gets fixed upstream this week, time is getting short before the launch.)

The migration those upstream changes require in #1379 would also be enough to make it tempting to hold off on upgrading until after the launch.

@chrisbobbe Does the upstream change that's needed for the #1545 fix happen before the UIScene breaking change? If so, let's upgrade to that point. If not, we can just put ignore-comments on those deprecation warnings for now, and update #1545 to be about removing those ignore-comments.

chrisbobbe:
> @chrisbobbe Does the upstream change that's needed for the #1545 fix happen before the UIScene breaking change?

Hmm, no, the change needed for #1545 (flutter/flutter@d0058ec36) happens after the UiScene breaking change (flutter/flutter@3d8fa621c):

```
chrisbobbe:~/dev/flutter (main u-40) 
$ git merge-base --is-ancestor 3d8fa621c d0058ec
chrisbobbe:~/dev/flutter (main u-40) 
$ echo $?
0
```

So yeah, I'll close this and do this:

> If not, we can just put ignore-comments on those deprecation warnings for now, and update https://github.com/zulip/zulip-flutter/issues/1545 to be about removing those ignore-comments.

chrisbobbe:
Sent #1562 for that and edited #1545.

PR #1547, opened 2025-06-04
test: Make home-page tests robust to changes in route animation duration

chrisbobbe:
This would fix some tests to allow an upstream change to land.

Discussion; see Greg's comment: https://github.com/flutter/flutter/pull/165832#issuecomment-2931857108
> We definitely have some tests in Zulip that wait a fixed number of ms when what they really want is "until the page transition finishes", and the real point of the test doesn't care whether that takes 300ms or 800ms or some other value. (I'd expect Zulip to be representative of a lot of apps out there in having some tests of this kind.) So where that's the case, the test would be cleaner if there were some value it could refer to instead.
> 
> It might be tricky to specify the value as a constant, though — different transitions may take different durations, right? So maybe the API that's needed is some way to look up the duration (or the remaining duration?) of the page transition that's currently in progress, or perhaps to get a Future that will complete when the transition finishes.

There are more places we use hard-coded durations, but those don't seem to be affected—the four test failures are all in `home_test.dart`.

### DIFF
--- a/test/widgets/home_test.dart
+++ b/test/widgets/home_test.dart
@@ -37,19 +37,22 @@ void main () {
   late Route<dynamic>? topRoute;
   late Route<dynamic>? previousTopRoute;
   late List<Route<dynamic>> pushedRoutes;
+  late Route<dynamic>? lastPoppedRoute;
 
   final testNavObserver = TestNavigatorObserver()
     ..onChangedTop = ((current, previous) {
         topRoute = current;
         previousTopRoute = previous;
       })
-    ..onPushed = ((route, prevRoute) => pushedRoutes.add(route));
+    ..onPushed = ((route, prevRoute) => pushedRoutes.add(route))
+    ..onPopped = ((route, prevRoute) => lastPoppedRoute = route);
 
   Future<void> prepare(WidgetTester tester) async {
     addTearDown(testBinding.reset);
     topRoute = null;
     previousTopRoute = null;
     pushedRoutes = [];
+    lastPoppedRoute = null;
     await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
     store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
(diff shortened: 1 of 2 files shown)


test/widgets/home_test.dart
```diff
@@ -168,6 +171,44 @@ void main () {
...
+        of: find.byType(BottomSheet),
+        matching: buttonFinder));
+      await tester.pump(Duration.zero);
+
+      final newPushedRoute = pushedRoutes.skip(numPushedRoutesBeforePress)
+        .singleOrNull;
+
+      final sheetPopDuration = (topRouteBeforePress as ModalBottomSheetRoute<void>)
+        .reverseTransitionDuration;
+      // TODO not sure why a 1ms fudge is needed; investigate.
```
rajveermalviya, line 195: Looks like the example provided upstream also has the same 1ms hack: https://github.com/flutter/flutter/pull/165832#issuecomment-2932867868.

Also the value of `sheetPopDuration` here turns out to be `200ms`, different from the previous `250ms`.

rajveermalviya (approved):
Thanks, LGTM! Just one non-blocking comment.

test/widgets/home_test.dart
chrisbobbe: I wonder if we tried 200ms when writing the test, but it failed for the same (unknown) reason it fails now, and we added 50ms to hack around instead of just 1ms.

chrisbobbe:
The CI failure is #1545; I have #1546 open for that. (That failure shouldn't affect the customer-tests run, because that's done with `flutter analyze --no-fatal-infos`, and the analyzer output in #1545 is all info-level.)

I'll go ahead and merge this and send a flutter/tests PR bumping the Zulip commit ID. Thanks for the review! Normally I wouldn't merge without Greg's review, but the changes are only in tests, Greg is OOO through Monday, and I'd like to unblock the upstream work.

PR #1550, opened 2025-06-04
home [nfc]: Clean up some dead safe-area code from before bottom-tabs

chrisbobbe:
The responsibility for consuming the bottom inset is now external to all of these *PageBody widgets; it's handled by
Scaffold.bottomNavigationBar. So we don't need to forward a bottom-inset value down to descendants
(via `SafeArea.bottom: false`), and we don't need to add widgets to pad the bottom inset; the bottom-inset height in all these places is zero.

### DIFF
--- a/lib/widgets/inbox.dart
+++ b/lib/widgets/inbox.dart
@@ -168,9 +168,7 @@ class _InboxPageState extends State<InboxPageBody> with PerAccountStoreAwareStat
         message: zulipLocalizations.inboxEmptyPlaceholder);
     }
 
-    return SafeArea(
-      // Don't pad the bottom here; we want the list content to do that.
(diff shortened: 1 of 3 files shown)


rajveermalviya:
Apologies, I missed the notification for this. Sorry about that.

rajveermalviya:
Thanks for doing this clean up!

I see that there are some conflicts currently, can you please resolve them. Thanks!

chrisbobbe:
No worries; this is just a small cleanup :) revision pushed.

rajveermalviya (approved):
Thanks @chrisbobbe! LGTM.

gnprice:
Thanks for the cleanup. Looks good; merging.

Issue #1551, opened 2025-06-06
home: Add leading/trailing graphics to empty-page placeholders


chrisbobbe:
For launch, we'll simplify the empty-page placeholders for these pages:

- Inbox: #386
- Channels
- Direct messages: #385

by just displaying some plain text. That's much better than just an empty space.

These have designs in Figma, though, with nice-looking SVGs. This issue is for incorporating those graphics.

| Inbox | Channels | Direct messages |
| --- | --- | --- |
| <img width="306" alt="Image" src="https://github.com/user-attachments/assets/cfd040f6-bc88-4476-b1ad-8fac3132b1fe" /> | <img width="378" alt="Image" src="https://github.com/user-attachments/assets/a24df5e3-1057-43b5-b977-87925d045849" /> | <img width="355" alt="Image" src="https://github.com/user-attachments/assets/dc4df048-2a74-4b46-8416-80ab296bde70" /> |

Figma links:

Inbox: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5966-176662&m=dev
Channels: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6205-25948&m=dev
Direct messages: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5956-115842&m=dev

karanbhatt16:
Hi! I'd like to work on this issue. I've started implementing the empty inbox placeholder illustration based on the Figma design and would like to submit a PR once it's ready. Please let me know if someone is already working on it.

PR #1552, opened 2025-06-06
home: Add placeholder text for empty Inbox, Channels, and Direct messages

chrisbobbe:
The "Direct messages" part of this was requested as part of #127 new-DM: it's item 0 in https://github.com/zulip/zulip-flutter/pull/1322#issuecomment-2917492437 . This should block on the rest of #127, because the empty "Direct messages" placeholder says "Why not start the conversation?", and there's no way to do that pre-#127.

Screenshots coming soon.

cc @alya

-----

The Figma has some graphics on all of these, but we'll leave that for later: see issue #1551.

Fixes: #385
Fixes: #386

### DIFF
--- a/assets/l10n/app_en.arb
+++ b/assets/l10n/app_en.arb
@@ -781,6 +781,10 @@
   "@inboxPageTitle": {
     "description": "Title for the page with unreads."
   },
+  "inboxEmptyPlaceholder": "There are no unread messages in your inbox. Use the buttons below to view the combined feed or list of channels.",
+  "@inboxEmptyPlaceholder": {
+    "description": "Centered text on the 'Inbox' page saying that there is no content to show."
+  },
   "recentDmConversationsPageTitle": "Direct messages",
   "@recentDmConversationsPageTitle": {
     "description": "Title for the page with a list of DM conversations."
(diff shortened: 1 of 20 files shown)


chrisbobbe:
Inbox:

| Before | After |
| --- | --- |
| image | image |
| image | image |

chrisbobbe:
Direct messages (when #127 is done, soon, there will be a "New DM" button at the bottom):

| Before | After |
| --- | --- |
| image | image |
| image | image |

chrisbobbe:
Channels:

| Before | After |
| --- | --- |
| image | image |
| image | image |

alya:
I'm pretty sure we had a CZO discussion about what to write as the second sentence in the empty inbox?

chrisbobbe:
Ah yeah, that's right: [#mobile > iOS feedback for Julia community member @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/iOS.20feedback.20for.20Julia.20community.20member/near/1964130)

So e.g.

> There are no unread messages in your inbox. You might be interested in _Combined feed_ or _Channels_.

(but with links instead of italics).

Putting inline links (or italics, icons, etc.) in prose is blocked on getting it to work with internationalization (#1285), though. That version of a second sentence seems kind of useless without the links, so I think we should either leave it off or rewrite so that it calls attention to the nav buttons at the bottom. And file a follow-up issue for putting the links in there, noting that it's blocked on #1285.

alya:
OK, we can do:

> There are no unread messages in your inbox. Use the buttons below to view the combined feed or list of channels.

chrisbobbe:
SGTM; updated (and updated the screenshots), and filed #1553 for the text with links.

gnprice:
Thanks! Looks good; merging.

PR #1559, opened 2025-06-10
KaTeX (3/n): Support negative margins for spans

rajveermalviya:
Stacked on top of #1452.

Related: #46


gnprice:
@rajveermalviya Copying here for visibility what you told me on our call today: the reason this is a draft (not ready to merge) is that you're still working on the widget test. Apart from that, you consider it ready to review.

lib/widgets/katex.dart
```diff
@@ -0,0 +1,202 @@
...
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(DiagnosticsProperty<EdgeInsetsGeometry>('padding', _padding));
+  }
+}
+
+// Like [RenderPadding] but supports negative values.
+class RenderNegativePadding extends RenderShiftedBox {
```
gnprice: Let's leave a TODO tracking our intention to get a version of this upstream:
```suggestion
// Like [RenderPadding] but supports negative values.
// TODO(upstream): give Padding an option to accept negative padding (at cost of hit-testing not working)
class RenderNegativePadding extends RenderShiftedBox {
```

lib/widgets/katex.dart
```diff
@@ -0,0 +1,202 @@
...
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(DiagnosticsProperty<EdgeInsetsGeometry>('padding', padding));
+    properties.add(EnumProperty<TextDirection>('textDirection', textDirection, defaultValue: null));
+  }
+}
+
+extension on EdgeInsetsGeometry {
+  bool get isNegative => !isNonNegative;
```
gnprice: I think this is clearer (and hardly much longer) if just inlined.

In particular there's some ambiguity in that "is negative" sounds like it might require *all* the sides to be negative. So being more transparent, with one fewer layer of indirection, helps mitigate that.

gnprice (commented):
Thanks! Here's small comments from a partial skim.

rajveermalviya:
Thanks for the initial comments @gnprice. Revision pushed.

lib/widgets/content.dart
```diff
@@ -946,14 +958,94 @@ class _KatexSpan extends StatelessWidget {
         child: widget);
     }
 
-    return SizedBox(
+    final marginRight = switch (styles.marginRightEm) {
+      double marginRightEm when marginRightEm >= 0 => marginRightEm * em,
```
gnprice: Is it intended to be possible to get to this point and have a negative marginRightEm?

I believe not — and it looks like if it did happen, this code would just ignore it, which seems like clearly a bug.

So instead of this "when" condition, let's have an assertion:
```dart
    assert((styles.marginLeftEm ?? 0) >= 0);
```

Similarly for right margin.

rajveermalviya: Updated so that parser makes sure not to emit a node with negative margin styles, as it already emits a separate `KatexNegativeMarginNode` for handling that on the widget side. And widget side retains the `assert(margin.isNonNegative);`.

rajveermalviya:

lib/widgets/content.dart
gnprice: Great, thanks — that definitely sounds cleaner.

test/model/content_test.dart
```diff
@@ -943,6 +943,522 @@ class ContentExample {
...
+          KatexSpanNode(
+            styles: KatexSpanStyles(),
+            text: null,
+            nodes: [
+              KatexStrutNode(heightEm: 0.8988, verticalAlignEm: -0.2155),
+              KatexSpanNode(
+                styles: KatexSpanStyles(),
+                text: null,
+                nodes: [
+                  KatexSpanNode(
```
gnprice: nit: Let's make the boring parts of this more compact, and also involve less indentation. Like so:
```suggestion
      MathBlockNode(texSource: '\\KaTeX', nodes: [
        KatexSpanNode(styles: KatexSpanStyles(), text: null, nodes: [
          KatexStrutNode(heightEm: 0.8988, verticalAlignEm: -0.2155),
          KatexSpanNode(styles: KatexSpanStyles(), text: null, nodes: [
            KatexSpanNode(
```

and similarly below.

test/model/content_test.dart
```diff
@@ -943,6 +943,522 @@ class ContentExample {
...
+                  fontStyle: KatexSpanFontStyle.italic),
+                text: 'c',
+                nodes: null),
+            ]),
+        ]),
+    ]);
+
+  static const mathBlockKatexNegativeMargins = ContentExample(
+    'math block katex negative margins',
+    '```math\n\\KaTeX\n```',
```
gnprice: This example is fine to have. But let's also have a more minimal, focused example that's just about negative margins. So:
```suggestion
  static const mathBlockKatexLogo = ContentExample(
    'math block KaTeX logo',
    '```math\n\\KaTeX\n```',
```

and then for a minimal example, try `$$ 1 \! 2 $$`. The minimal example can get a name like "katex negative margins".

For anyone who's investigating this in the future — or who makes a change, the change causes a regression here, and they need to investigate the regression to figure out the right fix — that minimal, focused example should be helpful as the place to look.

test/model/content_test.dart
```diff
@@ -943,6 +943,424 @@ class ContentExample {
...
+                  '<span class="vlist-s">​</span></span>'
+                  '<span class="vlist-r">'
+                    '<span class="vlist" style="height:0.15em;"><span></span></span></span></span></span></span></span></span></span></span></p>', [
+      MathBlockNode(
+        texSource: 'X_n',
+        nodes: [
+          KatexSpanNode(
+            styles: KatexSpanStyles(),
+            text: null,
+            nodes: [
```
gnprice: nit: same story as https://github.com/zulip/zulip-flutter/pull/1559#discussion_r2211697392 and https://github.com/zulip/zulip-flutter/pull/1698#discussion_r2212292471

test/model/content_test.dart
```diff
@@ -943,6 +943,424 @@ class ContentExample {
...
+            styles: KatexSpanStyles(fontFamily: 'KaTeX_Math', fontStyle: KatexSpanFontStyle.italic),
+            text: 'c', nodes: null),
+        ]),
+      ]),
+    ]);
+
+  static const mathBlockKatexNegativeMargin = ContentExample(
+    'math block, KaTeX negative margin',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2223563
+    '```math\n1 \\! 2\n```',
```
gnprice, line 1171: Let's be sure to include this new, minimal example (from https://github.com/zulip/zulip-flutter/pull/1559#discussion_r2211705232) ­in the widget tests too, so that it can fully do its work.

lib/model/katex.dart
```diff
@@ -209,11 +250,142 @@ class _KatexParser {
...
+                  text: null,
+                  nodes: [KatexNegativeMarginNode(
+                    leftOffsetEm: marginLeftEm,
+                    nodes: [KatexSpanNode(
+                      styles: styles.filter(topEm: false, marginLeftEm: false),
+                      text: null,
+                      nodes: _parseChildSpans(otherSpans))])]);
+              } else {
+                innerSpanNode = KatexSpanNode(
+                  styles: styles.filter(topEm: false),
```
gnprice: At this point `styles` has already had `topEm` filtered out, right? So I believe that filtering is redundant.

test/model/content_test.dart
```diff
@@ -2033,6 +2451,13 @@ void main() async {
   testParseExample(ContentExample.mathBlockKatexNestedSizing);
   testParseExample(ContentExample.mathBlockKatexDelimSizing);
   testParseExample(ContentExample.mathBlockKatexSpace);
+  testParseExample(ContentExample.mathBlockKatexSuperscript);
+  testParseExample(ContentExample.mathBlockKatexSubscript);
+  testParseExample(ContentExample.mathBlockKatexSubSuperScript);
+  testParseExample(ContentExample.mathBlockKatexRaisebox);
+  testParseExample(ContentExample.mathBlockKatexNegativeMargin);
+  testParseExample(ContentExample.mathBlockKatexLogo);
+  testParseExample(ContentExample.mathBlockNegativeMarginsOnVlistRow);
```
gnprice: From the last two commits (the commits newly added in yesterday's revision, mentioned at https://github.com/zulip/zulip-flutter/pull/1698#issuecomment-3082035127):
f845c6d0d content test: Add test for negative margins on a vlist row in KaTeX content
0ec2cc556 content: Filter negative margin styles if present on a vlist row

```diff
+  testParseExample(ContentExample.mathBlockNegativeMarginsOnVlistRow,
+    // TODO fix parser to filter negative margins styles from
+    //   vlist row span styles.
+    skip: true);
```

```diff
-  testParseExample(ContentExample.mathBlockNegativeMarginsOnVlistRow,
-    // TODO fix parser to filter negative margins styles from
-    //   vlist row span styles.
-    skip: true);
+  testParseExample(ContentExample.mathBlockNegativeMarginsOnVlistRow);
```

This version doesn't really give any added information compared to just squashing the two commits together. The normal arrangement is to have the tests appear in the same commit as the changes they're testing, because then the tests can be read in the light of those changes and vice versa, which can help for understanding both of them.

There's a variation of this split which would be useful, though:
* The first commit adds the test, but doesn't skip it; instead, the expectation reflects the old wrong behavior, and a TODO comment points out where it's wrong.
* Then the second commit makes the fix, and updates the test to expect the right behavior.

If the updates to the test are short and readable — simpler to read than the whole test itself — then this can be a useful way of highlighting exactly which aspect of the test expectation is the thing getting fixed by the change.

If I'm understanding this change correctly, then this would be a good example of that: the bit that's wrong would be a single line with a `marginLeftEm`.

gnprice: Relatedly:

    content: Filter negative margin styles if present on a vlist row
    
    This fixes a bug where if negative margin on a vlist row is present
    the widget side code would hit an assert. Displaying the error
    (red screen) in debug mode, but in release mode the negative padding
    would be applied twice, once by `_KatexNegativeMargin` and another by
    `Padding` widget.

Since one of the symptoms is that the widget code hits an assert, let's have a widget test that would exercise that assert. :-)

gnprice (commented):
Thanks for the revision! Comments below, mainly on the two new commits.

rajveermalviya:

test/model/content_test.dart
```diff
@@ -1165,6 +1165,180 @@ class ContentExample {
...
+              '<span class="msupsub">'
+                '<span class="vlist-t vlist-t2">'
+                  '<span class="vlist-r">'
+                    '<span class="vlist" style="height:0.1514em;">'
+                      '<span style="top:-2.55em;margin-left:-0.0785em;margin-right:0.05em;">'
+                        '<span class="pstrut" style="height:2.7em;"></span>'
+                        '<span class="sizing reset-size6 size3 mtight">'
+                          '<span class="mord mathnormal mtight">n</span></span></span></span>'
+                  '<span class="vlist-s">​</span></span>'
+                  '<span class="vlist-r">'
```
gnprice: nit:
```suggestion
                '<span class="vlist-t vlist-t2">'
                  '<span class="vlist-r">'
                    '<span class="vlist" style="height:0.1514em;">'
                      '<span style="top:-2.55em;margin-left:-0.0785em;margin-right:0.05em;">'
                        '<span class="pstrut" style="height:2.7em;"></span>'
                        '<span class="sizing reset-size6 size3 mtight">'
                          '<span class="mord mathnormal mtight">n</span></span></span></span>'
                    '<span class="vlist-s">​</span></span>'
                  '<span class="vlist-r">'
```

right?

(I.e., the .vlist-s is a child of the preceding .vlist-r, not of the .vlist-t.)

test/widgets/content_test.dart
```diff
@@ -619,6 +619,21 @@ void main() {
...
+          ('2', Offset(6.85, 3.36), Size(10.28, 25.00)),
+        ]),
+        (ContentExample.mathBlockKatexLogo, skip: false, [
+          ('K', Offset(0.0, 8.64), Size(16.0, 25.0)),
+          ('A', Offset(12.50, 10.85), Size(10.79, 17.0)),
+          ('T', Offset(20.21, 9.36), Size(14.85, 25.0)),
+          ('E', Offset(31.63, 14.52), Size(14.0, 25.0)),
+          ('X', Offset(43.06, 9.85), Size(15.42, 25.0)),
+        ]),
+        (ContentExample.mathBlockKatexNegativeMarginsOnVlistRow, skip: true, [
```
gnprice: This shouldn't be skipped at the tip commit, right?

gnprice (commented):
Thanks for the revision! I think the implementation now all looks good. Two small comments below on the tests; and would you also do the split described at https://github.com/zulip/zulip-flutter/pull/1559#discussion_r2214835996 ? I think it should be easy to do, and as discussed there I think this is a type of change where that split would be helpful for understanding the change.

rajveermalviya:

gnprice:
Thanks again for all your work on this! All looks good now — merging :grinning:

PR #1429, opened 2025-03-21
Distinguish muted users in UI

sm-sayedi:
**_Note: This PR is rebased on top of #1530._**

This PR aims to respect the user's settings for muting other users by distinguishing its related UI appearance. From the list of places in UI mentioned in https://zulip.com/help/mute-a-user, the following are included as they're compatible with the current scope of mobile:

- Combined feed sent by muted users, including the name, profile picture, and message content, are hidden behind a Click here to reveal banner. A revealed message can later be [re-hidden](https://zulip.com/help/mute-a-user#re-hide-a-message-that-has-been-revealed).
<table>
    <tr>
        <td><b>Message from muted user (Before)</b></td>
        <td><b>Message from muted user (After)</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
<tr>
        <td><b>Revealed muted message</b></td>
        <td><b>Revealed muted message (Long-pressed)</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
</table>

- Muted users have their name displayed as "Muted user" for [emoji reactions](https://zulip.com/help/emoji-reactions#view-who-reacted-to-a-message), [polls](https://zulip.com/help/create-a-poll), and when displaying the recipients of group direct messages.
<table>
    <tr>
        <td><b>Emoji reactions | Poll (Before)</b></td>
        <td><b>Emoji reactions | Poll (After)</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
<tr>
        <td><b>Group DMs with muted user (Before)</b></td>
        <td><b>Group DMs with muted user (After)</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
<tr>
        <td><b>Recent DMs view (Before)</b></td>
        <td><b>Recent DMs view (After)</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
</table>

- Recent conversations and other features that display avatars will show a generic user symbol in place of a muted user's profile picture.
<table>
<tr>
        <td><b>Custom profile field (Before)</b></td>
        <td><b>Custom profile field (After)</b></td>
    </tr>
    <tr>
        <td>
        <td></td>
    </tr>
</table>

- Muted users are excluded from the autocomplete for ~composing a direct message or~ [mentioning a user](https://zulip.com/help/mention-a-user-or-group).

Fixes: #296

Issue #296, opened 2023-09-02
Mute muted users


gnprice:
This Zulip feature: https://zulip.com/help/mute-a-user .

This issue only covers respecting the user's existing choice to mute (or unmute) a given user. Offering UI for changing that choice is a feature we don't have in zulip-mobile (https://github.com/zulip/zulip-mobile/issues/4655), and will be a separate issue.

That help-center doc has a handy (and long) list of places in the UI that should respond to user muting. where the user's name and avatar are hidden. One other such place is:
* on a user profile screen (of not necessarily a muted user), in a custom profile field with type `user`, when the muted user is among the users listed there.

  (I happened to notice this when reviewing #287 and comparing a profile screen between the PR and zulip-mobile, which does mute users in this spot.)

Because the list of places that should be affected is so long, we should find ways to encapsulate many of those to go through a much smaller number of distinct places in the source code; otherwise it'll be very easy to miss some, and to forget this feature when adding new pieces of UI that happen to show a user's name or avatar.
* One point of comparison for that encapsulation is the `UserItem` component in zulip-mobile. It's used in 8 other components; and expanding the call tree through the generic `UserList` and `UserPickerCard`, it's ultimately used in 11 different places in the UI.

  Similarly `UserAvatarWithPresence`, which is used in `UserItem` and used directly by several other pieces of UI.
* Another job which is probably convenient to handle as part of the same encapsulated piece of code is to deal with the case where a user is unknown, i.e. not in our data structures (which will become quite normal with https://github.com/zulip/zulip/issues/10970 ). We've been defaulting the avatar to transparent and the name to "(unknown user)". It's already awkward that that logic is duplicated in each piece of UI that needs it — makes it easy for different places to diverge accidentally, as in https://github.com/zulip/zulip-flutter/pull/287#discussion_r1313630648 — and it'll get more so when the "(unknown user)" part needs to gain complexity for i18n. So we need to encapsulate that job too, independently of the need for muting users; and they probably belong in the same abstraction.

gnprice:
An update on this issue's status, since it's an M5a launch blocker issue and launch is imminent:

This issue has been resolved in releases since v0.0.30. The releases have included PRs #1560 and #1561, or previously #1429, which implement this feature. I believe the PRs are close to merge; they're open just because I haven't found time for the last steps of review, as we've approached the launch. I intend to get back to that in the early post-launch period.

PR #1554, opened 2025-06-06
new-dm: Add UI for starting new DM conversation (Chris's revision)

chrisbobbe:
Many thanks to @chimnayajith for doing most of this work, in #1322!

As Greg foreshadowed in https://github.com/zulip/zulip-flutter/pull/1322#issuecomment-2918078351, we're on a tight timeline for getting this feature in. I've gone ahead and made this revision directly, incorporating my own feedback, rather than going through another round of review.

I've sent other PRs for some things that could be separated out:

- #1544
- #1552
  - (This one is actually meant to land after the current PR.)

Fixes: #127

PR #1560, opened 2025-06-10
Mute muted users (Chris's revision, 1 of 2)

chrisbobbe:
Stacked atop #1530.

Thanks @sm-sayedi for your work on muted-users, and for making sure I had your latest revision of #1429 before you left for vacation! 🙂

With the launch coming up, and since you're on vacation, I've gone ahead with taking #1429 forward with my own changes, as we discussed last week, rather than doing another round of review.

This PR is part of that work; more to come. In particular, this PR doesn't include the part of #1429 that excludes muted users and messages. It only affects how they appear in the UI.

Notable differences from #1429:
- Centralized the "Muted user" placeholder condition in `store.userDisplayName` and `store.senderDisplayName`, as suggested in the issue and in https://github.com/zulip/zulip-flutter/pull/1429#pullrequestreview-2884790447. (With some commits beforehand, to use those methods everywhere it makes sense to use them.)
- I got the `InheritedWidget` strategy to work (https://github.com/zulip/zulip-flutter/pull/1429#discussion_r2105542945 ); I think this is simpler and clearer.
- I've extended `ZulipWebUiKitButton` for the "Reveal message" button, rather than making a new ad hoc button. `ZulipWebUiKitButton` handles subtle things like a scale animation on press, and making sure its touch target is 44px tall even if the visible portion of the button is shorter (24px in this case).
- If a muted message is starred, this revision still shows the star, following web.
- Some adjustments to the tests.
- Some splitting of commits (like the rename of some members of `DesignVariables`)

Related: #296


chrisbobbe:
"Reveal message" button and "Muted user" placeholder in the message list, then "Hide muted message again" option in the action sheet:

| Light | Dark |
| --- | --- |
| image | image |
| image | image |

alya:
The slashed eye icon seems to be missing half its iris. It doesn't have that effect on web.
this PR

web

alya:
The screenshots look fine to me otherwise.

lib/model/compose.dart
```diff
@@ -212,14 +229,10 @@ String quoteAndReply(PerAccountStore store, {
...
   required String rawContent,
 }) {
-  final sender = store.getUser(message.senderId);
-  assert(sender != null); // TODO(#716): should use `store.senderDisplayName`
   final url = narrowLink(store,
     SendableNarrow.ofMessage(message, selfUserId: store.selfUserId),
     nearMessageId: message.id);
-    // Could ask `mention` to omit the |<id> part unless the mention is ambiguous…
-    // but that would mean a linear scan through all users, and the extra noise
-    // won't much matter with the already probably-long message link in there too.
```
gnprice, line 222: I think this comment is still relevant for explaining a choice we made about how this logic works.

(The name `mention` needs updating, though — already did, to say `userMention`.)

lib/model/user.dart
```diff
@@ -44,28 +44,47 @@ mixin UserStore on PerAccountStoreBase {
...
+  /// Otherwise this is the user's [User.fullName] if the user is known,
+  /// or (if unknown) [ZulipLocalizations.unknownUserName].
   ///
   /// When a [Message] is available which the user sent,
   /// use [senderDisplayName] instead for a better-informed fallback.
-  String userDisplayName(int userId) {
+  String userDisplayName(int userId, {bool replaceIfMuted = true}) {
+    if (replaceIfMuted && isUserMuted(userId)) {
+      return GlobalLocalizations.zulipLocalizations.mutedUser;
+    }
```
gnprice, line 58: muted-users: Say "Muted user" to replace a user's name, where applicable
    
    (Done by adding an is-muted condition in store.userDisplayName and
    store.senderDisplayName, with an opt-out param.)

Cool, I like this structure.

The inventory of where we say "Muted user" and where we don't is also helpful:

    If Chris is muted, we'll now show "Muted user" where before we would
    show "Chris Bobbe", in the following places:
    
    - Message-list page:
      - DM-narrow app bar.
      - DM recipient headers.
      - The sender row on messages. This and message content will get
        more treatment in a separate commit.
      - Emoji-reaction chips on messages.
      - Typing indicators ("Muted user is typing…"), but we'll be
        excluding muted users, coming up in a separate commit.
      - Voter names in poll messages.
    - DM items in the Inbox page. (Messages from muted users are
      automatically marked as read, but they can end up in the inbox if
      you un-mark them as read.)
    - @-mention autocomplete, but we'll be excluding muted users, coming
      up in a separate commit.
    - Items in the Direct messages ("recent DMs") page. We'll be
      excluding DMs where everyone is muted, coming up in a separate
      commit.
    - User items in custom profile fields.
    
    We *don't* do this replacement in the following places, i.e., we'll
    still show "Chris Bobbe" if Chris is muted:
    
    - Sender name in the header of the lightbox page. (This follows
      web.)
    - The "hint text" for the compose box in a DM narrow: it will still
      say "Message @Chris Bobbe", not "Message @Muted user". (This
      follows web.)
    - The user's name at the top of the Profile page.
    - We won't generate malformed @-mention syntax like
      @_**Muted user|13313**.

lib/widgets/message_list.dart
```diff
@@ -151,13 +165,17 @@ class MessageListPage extends StatefulWidget {
...
-  /// don't call this in a build method.
-  // If we do find ourselves wanting this in a build method, it won't be hard
-  // to enable that: we'd just need to add an [InheritedWidget] here.
+  /// Uses the efficient [BuildContext.dependOnInheritedWidgetOfExactType],
+  /// so this may be called in a build method.
+  ///
+  /// Because this uses [BuildContext.dependOnInheritedWidgetOfExactType],
+  /// it creates a dependency, and [context] will rebuild when the underlying
+  /// [State.setState] is called.
   static MessageListPageState ancestorOf(BuildContext context) {
```
gnprice: nit: Let's rename to just `of`, either in this commit or right after it. The "ancestor of" name was meant to give a bit of pause when one considers calling it, because it suggests the idea of walking up the tree through ancestors, whereas `.of` is the usual convention for finding an InheritedWidget.

lib/widgets/message_list.dart
```diff
@@ -283,7 +318,30 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
...
+    required super.child,
+  });
+
+  final MessageListPageState state;
+
+  @override
+  bool updateShouldNotify(covariant _MessageListPageInheritedWidget oldWidget) {
+    // Ensure that dependent elements using [MessageListPage.ancestorOf]
+    // always rebuild when _MessageListPageState.setState is called.
+    return true;
```
gnprice: This isn't specific to `_MessageListPageState.setState`, is it? I believe this means the dependent elements will get rebuilt any time the MessageListPage gets rebuilt. Which could e.g. be because one of its own dependencies was updated.

gnprice: Maybe the cleanest thing would be to have the InheritedWidget be an InheritedNotifier, specifically for the set of muted messages. That set could be an object like this:
```dart
class RevealedMutedMessagesState extends ChangeNotifier {
  final Set<int> _revealedMessages = {};

  bool isMutedMessageRevealed(int messageId) =>
    _revealedMessages.contains(messageId);

  void _add(int messageId) {
    _revealedMessages.add(messageId);
    notifyListeners();
  }

  // … _remove …
}
```

Then the bit on `_MessageListPageState` would look something like this:
```dart
  final _revealedMutedMessages = RevealedMutedMessagesState();

  @override
  void revealMutedMessage(int messageId) {
    _revealedMutedMessages._add(messageId);
  }
```

And it wouldn't have an `isMutedMessageRevealed` method directly; instead, there'd be a static on MessageListPage that looks like `static RevealedMutedMessagesState revealedMutedMessagesOf(BuildContext context)`. A caller that needs `isMutedMessageRevealed` starts by calling that static, which creates a dependency; then it calls `isMutedMessageRevealed` on that; and the dependency ensures that the caller gets notified if the data changes.

chrisbobbe:
Oops, just noticed in the 0.0.31 beta that I dropped a piece of #1429 🙂 we should disable gestures on hidden messages in the message list (tap sender-row to go to profile; long-press to bring up message action sheet). Fixed.

gnprice:
Ah good thought, thanks. (Looks like the previous revision already did that for tapping the sender row, but not for long-pressing the message.)

Would you also rebase to resolve conflicts? (After #1554 and #1552.) That'd help simplify preparing the next release branch.

lib/widgets/message_list.dart
chrisbobbe: I think this is moot with the `InheritedNotifier` / `RevealedMutedMessagesState` approach: https://github.com/zulip/zulip-flutter/pull/1560#discussion_r2139426438

Leafward widgets won't need `MessageListPageState` in their build methods (they'll use `RevealedMutedMessagesState`), so we don't need to provide `MessageListPageState` via an inherited widget.

chrisbobbe:
Done; revision pushed! This is on Sayed's revision of the prep PR from today; please see updated screenshots: https://github.com/zulip/zulip-flutter/pull/1530#issuecomment-2967878936

chrisbobbe:
I don't think I've seen this CI failure before. I'll trigger it to rerun:

<details>

```
Run flutter precache --universal
Downloading Linux x64 Dart SDK from Flutter engine f[7](https://github.com/zulip/zulip-flutter/actions/runs/15618925511/job/43998675759?pr=1560#step:6:8)9452e3f4ea7b830a00cafa0135c6658e75cd50...

  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
100   30[8](https://github.com/zulip/zulip-flutter/actions/runs/15618925511/job/43998675759?pr=1560#step:6:9)  100   308    0     0     58      0  0:00:05  0:00:05 --:--:--    76
Warning: Problem : HTTP error. Will retry in 1 seconds. 3 retries left.

  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
100   308  100   308    0     0     60      0  0:00:05  0:00:05 --:--:--    74
Warning: Problem : HTTP error. Will retry in 2 seconds. 2 retries left.

  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:06 --:--:--     0
100   308  100   308    0     0     51      0  0:00:06  0:00:06 --:--:--    76
Warning: Problem : HTTP error. Will retry in 4 seconds. 1 retries left.

  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
100   308  100   308    0     0     5[9](https://github.com/zulip/zulip-flutter/actions/runs/15618925511/job/43998675759?pr=1560#step:6:10)      0  0:00:05  0:00:05 --:--:--    73

```

</details>

(edit: it worked!)

lib/widgets/message_list.dart
```diff
@@ -1447,31 +1516,40 @@ class _SenderRow extends StatelessWidget {
     final sender = store.getUser(message.senderId);
     final time = _kMessageTimestampFormat
       .format(DateTime.fromMillisecondsSinceEpoch(1000 * message.timestamp));
+
+    final revealedMutedMessages = MessageListPage.revealedMutedMessagesOf(context);
+    final showAsMuted = store.isUserMuted(message.senderId)
+      && message is Message // i.e., not an outbox message
+      && !revealedMutedMessages.isMutedMessageRevealed((message as Message).id);
```
gnprice: Usually `isUserMuted` will be false, so we can optimize this by saving the `revealedMutedMessagesOf` lookup (and the resulting dependency tracking) in that case.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! That looks good.

So what remains is for me to read the remainder of the last commit:
ba57e4147 msglist: Hide content of muted messages, with a "Reveal message" button

lib/widgets/action_sheet.dart
```diff
@@ -597,6 +599,8 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       QuoteAndReplyButton(message: message, pageContext: pageContext),
     if (showMarkAsUnreadButton)
       MarkAsUnreadButton(message: message, pageContext: pageContext),
+    if (isSenderMuted)
+      UnrevealMutedMessageButton(message: message, pageContext: pageContext),
```
gnprice: I was confused initially on first looking at this line — wait we only show this if the sender is muted, but the button is to *un*reveal the message? I think a short comment would help:
```suggestion
    if (isSenderMuted)
      // The message must have been revealed in order to open this action sheet.
      UnrevealMutedMessageButton(message: message, pageContext: pageContext),
```

lib/widgets/action_sheet.dart
```diff
@@ -902,6 +906,27 @@ class MarkAsUnreadButton extends MessageActionSheetMenuItemButton {
...
+  IconData get icon => ZulipIcons.eye_off;
+
+  @override
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionHideMutedMessage;
+  }
+
+  @override
+  void onPressed() {
+    findMessageListPage().unrevealMutedMessage(message.id);
```
gnprice: And this could gain an assertion:

```suggestion
  void onPressed() {
    // The message should have been revealed in order to reach this action sheet.
    assert(MessageListPage.revealedMutedMessagesOf(pageContext)
      .isMutedMessageRevealed(message.id));
    findMessageListPage().unrevealMutedMessage(message.id);
```

lib/widgets/message_list.dart
```diff
@@ -155,6 +164,21 @@ class MessageListPage extends StatefulWidget {
...
+  /// This is updated via [MessageListPageState.revealMutedMessage]
+  /// and [MessageListPageState.unrevealMutedMessage].
+  ///
+  /// Uses the efficient [BuildContext.dependOnInheritedWidgetOfExactType],
+  /// so this is safe to call in a build method.
+  static RevealedMutedMessagesState revealedMutedMessagesOf(BuildContext context) {
+    final state =
+      context.dependOnInheritedWidgetOfExactType<_RevealedMutedMessagesProvider>()
+      ?.state;
+    assert(state != null, 'No _RevealedMutedMessagesProvider ancestor');
```
gnprice: nit: From the caller's perspective, I think the error is that there's no MessageListPage:
```suggestion
  static RevealedMutedMessagesState revealedMutedMessagesOf(BuildContext context) {
    final state =
      context.dependOnInheritedWidgetOfExactType<_RevealedMutedMessagesProvider>()
      ?.state;
    assert(state != null, 'No MessageListPage ancestor');
```

If there's somehow a MessageListPage but no _RevealedMutedMessagesProvider, that's a bug in MessageListPage.

This is similar to what PerAccountStoreWidget.of says:
```dart
    final widget = context.dependOnInheritedWidgetOfExactType<_PerAccountStoreInheritedWidget>();
    assert(_debugCheckFound(context, widget));
// …
  static bool _debugCheckFound(BuildContext context, Object? ancestor) {
    // Compare [debugCheckHasMediaQuery], and its caller [MediaQuery.of].
    assert(() {
      if (ancestor != null) return true;
      throw FlutterError.fromParts([
        ErrorSummary('No PerAccountStoreWidget ancestor found.'),
```

where the thing it actually looks up is _PerAccountStoreInheritedWidget but that's an implementation detail.

gnprice (commented):
OK, and finished reading. Just a few small comments below.

chrisbobbe:

gnprice:
Thanks! All looks good. Just rebased and resolved conflicts; will merge after running tests.

PR #1561, opened 2025-06-10
Mute muted users (Chris's revision, 2 of 2)

chrisbobbe:
Stacked on #1560.

Many thanks again to @sm-sayedi for your work on this! As with #1560, this is meant to take #1429 across the finish line in time for the upcoming app launch. Thanks again for making sure I had your latest revision of https://github.com/zulip/zulip-flutter/pull/1429 before you left for vacation. 🙂

In these new commits:
- In "Direct messages" (the recent-DMs page), we exclude conversations where everyone is muted.
- In "Combined feed", "Starred", and "Mentions", we exclude messages in those conversations.
- In the typing-status text ("Chris Bobbe is typing…"), we exclude muted users.

Fixes: #296


chrisbobbe:
Revision pushed; this one also excludes muted users from the new new-DM UI, which landed after my last revision.

lib/model/channel.dart
```diff
@@ -147,11 +147,11 @@ enum VisibilityEffect {
...
 
-  factory VisibilityEffect._fromBeforeAfter(bool before, bool after) {
+  factory UserTopicVisibilityEffect._fromBeforeAfter(bool before, bool after) {
     return switch ((before, after)) {
-      (false, true) => VisibilityEffect.unmuted,
-      (true, false) => VisibilityEffect.muted,
-      _             => VisibilityEffect.none,
+      (false, true) => UserTopicVisibilityEffect.unmuted,
+      (true, false) => UserTopicVisibilityEffect.muted,
+      _             => UserTopicVisibilityEffect.none,
```
gnprice, line 154: This is awkwardly verbose, but we're about to add another kind of
    visibility effect, and I think the code will end up clearer if we
    make a separate enum for it.

Yeah, agreed.

Some of the verbosity (like in these lines) will at least get resolved by the upcoming Dart feature of "dot shorthands", expected later this year:
https://github.com/dart-lang/language/blob/main/working/3616%20-%20enum%20value%20shorthand/proposal-simple-lrhn.md

I believe with that feature we'll be able to say just `.unmuted`, `.muted`, `.none` here, without the name of the enum type.

lib/model/message_list.dart
```diff
@@ -504,7 +505,9 @@ class MessageListView with ChangeNotifier, _MessageSequence {
         return switch (message.conversation) {
           StreamConversation(:final streamId, :final topic) =>
             store.isTopicVisible(streamId, topic),
-          DmConversation() => true,
+          DmConversation(:final allRecipientIds) =>
+            !store.shouldMuteDmConversation(DmNarrow(
+              allRecipientIds: allRecipientIds, selfUserId: store.selfUserId)),
```
gnprice: Let's make another DmNarrow constructor for this:
```suggestion
            !store.shouldMuteDmConversation(DmNarrow.ofConversation(
              message.conversation, selfUserId: store.selfUserId)),
```

Compare the comment at the top of that class:
```dart
// Zulip has many ways of representing a DM conversation; for example code
// handling many of them, see zulip-mobile:src/utils/recipient.js .
// Please add more constructors and getters here to handle any of those
// as we turn out to need them.
class DmNarrow extends Narrow implements SendableNarrow {
```

It'd be bad if this logic did something inefficient like copy the list of recipient IDs, because we're doing this for potentially a large fraction of the messages in this feed (if a user uses lots of DMs). I believe this in fact doesn't make such a copy — and a constructor on DmNarrow is a good place to centralize coming to that conclusion.

lib/model/message_list.dart
```diff
@@ -515,45 +518,67 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
+    switch (narrow) {
+      case CombinedFeedNarrow():
+      case ChannelNarrow():
       case MentionsNarrow():
       case StarredMessagesNarrow():
+        return false;
+
+      case TopicNarrow():
+      case DmNarrow():
         return true;
```
gnprice: nit: let's keep these cases in the same order as in _messageVisible, for ease of comparing the logic. It's important that the two align (that's this method's job); and it's potentially a bit subtle to confirm that, so it's good to make it as easy as possible.

It's fine for this to have two separate groups of cases that return the same thing.

lib/model/message_list.dart
```diff
@@ -515,45 +518,67 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
-  /// Whether [_messageVisible] is true for all possible messages.
-  ///
-  /// This is useful for an optimization.
-  bool get _allMessagesVisible {
-    switch (narrow) {
+  /// Whether this event could affect the result that [_messageVisible]
+  /// would ever have returned for any possible message in this message list.
+  MutedUsersVisibilityEffect _mutedUsersEventCanAffectVisibility(MutedUsersEvent event) {
+    switch(narrow) {
       case CombinedFeedNarrow():
```
gnprice, line 745: similarly, put the cases in the same order as in `_messageVisible`

lib/model/user.dart
```diff
@@ -91,6 +145,36 @@ class UserStoreImpl extends PerAccountStoreBase with UserStore {
...
+  @override
+  MutedUsersVisibilityEffect mightChangeShouldMuteDmConversation(MutedUsersEvent event) {
+    final sortedOld = _mutedUsers.toList()..sort();
+    final sortedNew = event.mutedUsers.map((u) => u.id).toList()..sort();
+    assert(isSortedWithoutDuplicates(sortedOld));
+    assert(isSortedWithoutDuplicates(sortedNew));
+    final union = setUnion(sortedOld, sortedNew);
+
+    final willMuteSome = sortedOld.length < union.length;
+    final willUnmuteSome = sortedNew.length < union.length;
```
gnprice, line 164: Neat, yeah, this logic works.

gnprice (commented):
Thanks! I've read so far the first three commits:
8af150c1c recent-dms: Exclude DM conversations where all other recipients are muted
06937e21c msglist [nfc]: Place `_allMessagesVisible` right after `_messageVisible`
8339c640f msglist [nfc]: s/VisibilityEffect/UserTopicVisibilityEffect/

and the next one except for its tests:
898a2ade2 msglist: In combined/mentions/starred, exclude DMs if all recipients muted

Generally all looks good; a few small comments below.

That leaves four more commits that I haven't read (though I've skimmed them):
619eb246a autocomplete: Exclude muted users from user-mention autocomplete
87df78003 msglist [nfc]: Represent nobody-is-typing as null, not empty list
c40cb89e8 msglist: Exclude muted users from typing-status text
0acc63b20 new-dm: Exclude muted users

lib/model/message_list.dart
chrisbobbe: Huh…it looks like `DmNarrow.ofMessage`, though, does copy the list of recipient IDs:

```dart
  factory DmNarrow.ofMessage(MessageBase<DmConversation> message, {
    required int selfUserId,
  }) {
    return DmNarrow(
      allRecipientIds: List.unmodifiable(message.conversation.allRecipientIds),
      selfUserId: selfUserId,
    );
  }
```

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Those changes look good.

There's a CI failure in build_runner.

(And there are parts of the branch I still need to read, per https://github.com/zulip/zulip-flutter/pull/1561#pullrequestreview-2922900623 .)

chrisbobbe:
I just rebased atop current `main` (after the dependency-upgrades PR), in case that fixes CI. 🤔

lib/model/message_list.dart
gnprice: Hmm indeed.

It looks like we don't currently call that one in any place that's as performance-sensitive as this. But it's a bit of a pitfall lying around.

lib/model/autocomplete.dart
```diff
@@ -753,9 +754,10 @@ class MentionAutocompleteQuery extends ComposeAutocompleteQuery {
       || wildcardOption.localizedCanonicalString(localizations).contains(_lowercase);
   }
 
-  bool testUser(User user, AutocompleteDataCache cache) {
+  bool testUser(User user, AutocompleteDataCache cache, UserStore store) {
     // TODO(#236) test email too, not just name
     if (!user.isActive) return false;
+    if (store.isUserMuted(user.userId)) return false;
 
     return _testName(user, cache);
```
gnprice: nit:
```suggestion
    if (!user.isActive) return false;
    if (store.isUserMuted(user.userId)) return false;

    // TODO(#236) test email too, not just name
    return _testName(user, cache);
```

The comment was already a bit separated from what it applied to, so perhaps should have been moved down there sooner; but this would widen the separation, so let's fix it.

(Can stay in the same commit, no need to make a separate prep commit.)

lib/widgets/message_list.dart
```diff
@@ -943,13 +1012,16 @@ class _TypingStatusWidgetState extends State<TypingStatusWidget> with PerAccount
     final store = PerAccountStoreWidget.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
     final typistIds = model!.typistIdsInNarrow(narrow);
-    if (typistIds.isEmpty) return const SizedBox();
-    final text = switch (typistIds.length) {
+    if (typistIds == null) return const SizedBox();
```
gnprice: msglist [nfc]: Represent nobody-is-typing as null, not empty list
    
    Just to skip the bit of extra work of creating new lists.

Hmm, this feels like the API gets a bit more complicated from the caller's perspective. What if `typistIds` is empty, and why is there a different possibility which is that it's null?

(In reality I think it's never empty after this change — _removeTypist removes the sub-maps when empty. But that's not obvious from the outside.)

lib/model/typing_status.dart
```diff
@@ -21,8 +21,8 @@ class TypingStatus extends PerAccountStoreBase with ChangeNotifier {
 
   Iterable<SendableNarrow> get debugActiveNarrows => _timerMapsByNarrow.keys;
 
-  Iterable<int> typistIdsInNarrow(SendableNarrow narrow) =>
-    _timerMapsByNarrow[narrow]?.keys ?? [];
+  Iterable<int>? typistIdsInNarrow(SendableNarrow narrow) =>
+    _timerMapsByNarrow[narrow]?.keys;
```
gnprice: An alternate way of not creating new lists each time:
```dart
  Iterable<int> typistIdsInNarrow(SendableNarrow narrow) =>
    _timerMapsByNarrow[narrow]?.keys ?? const [];
```

lib/widgets/message_list.dart
```diff
@@ -943,13 +1012,16 @@ class _TypingStatusWidgetState extends State<TypingStatusWidget> with PerAccount
...
+    final filteredTypistIds = typistIds
+      .whereNot((userId) => store.isUserMuted(userId));
```
gnprice: nit: can use a tear-off, which is shorter and I think here a bit easier to read:
```suggestion
    final filteredTypistIds = typistIds.whereNot(store.isUserMuted);
```

test/widgets/message_list_test.dart
```diff
@@ -812,9 +832,19 @@ void main() {
         await checkTyping(tester,
           eg.typingEvent(narrow, TypingOp.stop, eg.otherUser.userId),
           expected: 'Third User and Fourth User are typing…');
+        await store.setMutedUsers([eg.thirdUser.userId]);
+        await tester.pump();
+        check(tester.widget<Text>(finder)).data.equals('Fourth User is typing…');
```
gnprice: Let's put this in a separate test case. This one (or this pair of them) is pretty long and a bit unfocused already.

The new test case can be for just a topic narrow; I don't think it needs to exercise both kinds of conversations.

lib/widgets/new_dm_sheet.dart
```diff
@@ -68,7 +69,9 @@ class _NewDmPickerState extends State<NewDmPicker> with PerAccountStoreAwareStat
   }
 
   void _initSortedUsers(PerAccountStore store) {
-    sortedUsers = List<User>.from(store.allUsers)
+    final sansMuted = store.allUsers
+      .whereNot((User user) => store.isUserMuted(user.userId));
```
gnprice: nit: omit boring types
```suggestion
    final sansMuted = store.allUsers
      .whereNot((user) => store.isUserMuted(user.userId));
```

gnprice:
(Just rebased after merging #1560.)

lib/model/typing_status.dart
chrisbobbe: Aha indeed. Thanks!

chrisbobbe:
Thanks for the review and rebase! Revision pushed.

test/widgets/message_list_test.dart
```diff
@@ -893,6 +893,22 @@ void main() {
...
+    testWidgets('muted user typing', (tester) async {
+      await setupMessageListPage(tester,
+        narrow: topicNarrow, users: users, messages: [streamMessage]);
+
+      await store.setMutedUsers([eg.otherUser.userId]);
+      await tester.pump();
+
+      await checkTyping(tester,
+        eg.typingEvent(topicNarrow, TypingOp.start, eg.thirdUser.userId),
+        expected: 'Third User is typing…', // no "Other User"
```
gnprice, line 914: Hmm but is "Other user" typing in the first place? I don't see how they are.

Maybe have both users typing, then make one of them muted and check that only the other one is shown. That should make clear it's the muting that causes the other not to be shown.

chrisbobbe: Eep, indeed, thanks for the catch. Updated!

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merged.

PR #1562, opened 2025-06-10
settings [nfc]: Suppress some new deprecation warnings

chrisbobbe:
See #1546 for a PR that would resolve them, and why we're not doing that just now (it would require upgrading Flutter past a breaking change).


gnprice:
Thanks for taking care of this! Looks good; merging.

PR #1566, opened 2025-06-12
anchors 10/n: Open /near/ links at message

### LINKED ISSUE #82: Open a message list in the middle of history, when requested

gnprice:
Fixes #82.

This is the next round after #1515 — and completes #82, as well as bringing us close to #80. (These changes were previously sent as part of #1517.)

This branch is a draft only because of a lack of tests. The changes it does have are ready for review.

This doesn't implement the other aspect of the behavior of /near/ links, i.e. to follow the message if it's been moved to another conversation. That's tracked as #683.

### DIFF
--- a/lib/model/internal_link.dart
+++ b/lib/model/internal_link.dart
@@ -125,29 +125,33 @@ class NarrowLink extends InternalLink {
   final Narrow narrow;
 }
 
-/// A [Narrow] from a given URL, on `store`'s realm.
+/// Try to parse the given URL as a page in this app, on `store`'s realm.
 ///
 /// `url` must already be a result from [PerAccountStore.tryResolveUrl]
 /// on `store`.
 ///
-/// Returns `null` if any of the operator/operand pairs are invalid.
+/// Returns null if the URL isn't on this realm,
+/// or isn't a valid Zulip URL,
+/// or isn't currently supported as leading to a page in this app.
 ///
+/// In particular this will return null if `url` is a `/#narrow/…` URL
+/// and any of the operator/operand pairs are invalid.
 /// Since narrow links can combine operators in ways our [Narrow] type can't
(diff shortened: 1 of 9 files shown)


lib/model/internal_link.dart
```diff
@@ -221,8 +223,9 @@ NarrowLink? _interpretNarrowSegments(List<String> segments, PerAccountStore stor
         // It is fine to have duplicates of the same [IsOperand].
         isElementOperands.add(IsOperand.fromRawString(operand));
 
-      case _NarrowOperator.near: // TODO(#82): support for near
-        continue;
+      case _NarrowOperator.near:
+        if (nearMessageId != null) return null;
+        nearMessageId = int.tryParse(operand, radix: 10);
```
chrisbobbe: Could reject the link (i.e. return null) if the `int.tryParse` returns null, like we do for `/with/sdlfkjasdlkf` in the `case _NarrowOperator.with_` above this

chrisbobbe (commented):
Looks great—very excited for this! One small comment below, and I see you've just filed #1569 for tests.

lib/model/internal_link.dart
gnprice: Yeah, good thought — will do.

gnprice:
Thanks for the review! Made that tweak, and I also went through and converted the TODOs in commit messages into TODO comments pointing to #1569 or #1570.

(Plus made the actual test changes in one commit, where they were simple:
cb94d4e3a internal_link [nfc]: Propagate InternalLink up to caller
)

That way I think we can go ahead and merge this.

chrisbobbe:
Yep! Please do, maybe waiting till CI passes. (The range-diff looks fine to me.)

gnprice:
Thanks! Merged.

Issue #622, opened 2024-04-06
tools/check: `pigeon` runs old/cached version of Pigeon after Pigeon dependency is upgraded


chrisbobbe:
Part of #621 was to upgrade Pigeon, from 17.2.0 to 18.0.0. After I did that, `tools/check pigeon --all-files` on my machine was running the old Pigeon, while in CI the newer Pigeon was running, as we want it to, and flagging the need to update generated files. I found a workaround: https://github.com/zulip/zulip-flutter/pull/621#issuecomment-2040764190 but it would be nice not to require a manual step.

Can we make it so when we change the version of the Pigeon dependency, the next run of `tools/check pigeon --all-files` uses the updated version?

gnprice:
To make the workaround explicit in this thread: it's that after an upgrade affecting `package:pigeon`, and before running `tools/check pigeon`, one runs `dart run pigeon`.

That just prints the `pigeon` CLI's help message… but first it causes the `pigeon` CLI binary itself to get rebuilt with the new package.

gnprice:
Ran into this again just now (and had forgotten the workaround):
https://github.com/zulip/zulip-flutter/pull/1568#issuecomment-2968698121

Fixing this is probably just a couple of lines in `tools/check`, in the implementation of `run_pigeon`.

PR #1568, opened 2025-06-12
Upgrade Flutter, packages and Android build dependencies

rajveermalviya:
Reverts #1562, and includes cherry-picked commits from Chris's #1546 for RadioGroup API migration.


lib/widgets/settings.dart
```diff
@@ -44,21 +44,19 @@ class _ThemeSetting extends StatelessWidget {
...
-      children: [
-        ListTile(title: Text(zulipLocalizations.themeSettingTitle)),
-        for (final themeSettingOption in [null, ...ThemeSetting.values])
-          RadioListTile<ThemeSetting?>.adaptive(
-            title: Text(ThemeSetting.displayName(
-              themeSetting: themeSettingOption,
-              zulipLocalizations: zulipLocalizations)),
-            value: themeSettingOption,
-            // TODO(#1545) stop using the deprecated members
-            // ignore: deprecated_member_use
```
gnprice: These ignores should get removed in the same commit that makes them no longer needed, rather than in a previous commit; otherwise the previous commit would enter a broken state.

ios/Podfile.lock
```diff
@@ -37,37 +37,37 @@ PODS:
...
+    - "GoogleUtilities/NSData+zlib (~> 8.1)"
+  - FirebaseInstallations (11.13.0):
+    - FirebaseCore (~> 11.13.0)
+    - GoogleUtilities/Environment (~> 8.1)
+    - GoogleUtilities/UserDefaults (~> 8.1)
     - PromisesObjC (~> 2.4)
-  - FirebaseMessaging (11.10.0):
-    - FirebaseCore (~> 11.10.0)
+  - FirebaseMessaging (11.13.0):
+    - FirebaseCore (~> 11.13.0)
```
gnprice, line 64: Notable change is Firebase iOS SDK bump to 11.13.0,
    from 11.10.0, changelog for that is at:
      https://firebase.google.com/support/release-notes/ios
    
    No notable change there for FCM (the only component we use).

By my reading, there's no change there at all, right? Not just no notable change.

android/app/lint-baseline.xml
```diff
@@ -1,12 +1,12 @@
...
     <!--TODO(#855) see if this is no longer needed-->
     <issue
         id="InvalidPackage"
         message="Invalid package reference in library; not included in Android: `javax.xml.stream`. Referenced from `org.apache.tika.utils.XMLReaderUtils`.">
         <location
-            file="$GRADLE_USER_HOME/caches/modules-2/files-2.1/org.apache.tika/tika-core/3.1.0/6ba44a9ddf8f6f2b4bc88e8bc64269aea0424607/tika-core-3.1.0.jar"/>
+            file="$GRADLE_USER_HOME/caches/modules-2/files-2.1/org.apache.tika/tika-core/3.2.0/9232bb3c71f231e8228f570071c0e1ea29d40115/tika-core-3.2.0.jar"/>
     </issue>
 
 </issues>
```
chrisbobbe, line 12:
```
deps: Update file_picker to 10.2.0, from 10.1.9
```

I'm not following why the lint-baseline changes are necessary. At this commit, for me, `tools/check android` still fails with a similar-looking lint with what looks like the original location string (`…3.2.0/9232bb…`). Is there a command I should run?

gnprice (commented):
Thanks for taking care of these! This all looks great, modulo two small comments below.

The #1545 fix will also need a few more changes now that #1517 has merged, as a sort of non-local merge conflict:
https://github.com/zulip/zulip-flutter/pull/1517#discussion_r2143728481

Since it's fine for #1545 to wait until next week, and in the interest of getting the rest merged so it can go into a beta later today, I'll drop those three commits:
ccdacf9bc Revert "settings [nfc]: Suppress some new deprecation warnings"
023f1e648 settings test: Check theme-setting radio buttons by checking what paints
ca3de6aa1 settings: Migrate to new RadioGroup API for theme-setting radio buttons

and make the tweak for the other comment below. Then I'll go ahead and merge.

android/app/lint-baseline.xml
chrisbobbe: I tried `./gradlew clean` and `rm -rf .gradle` (after `cd android`) and reran; same result.

gnprice: Hmm interesting. `3.2.0/9232bb…` is the new location string, though, right?

Do you get the same failure in main?

chrisbobbe: > […] is the new location string, though, right?

Er oops, I think I copy-pasted from the wrong thing (the lint report showed two things (1) a suppression that wasn't used because nothing matched, and (2) an error that wasn't caught by a suppression, and they looked very similar).

Anyway, it's passing now; not sure what changed 🤷‍♂️ :

```
Lint found no errors or warnings

1 errors/warnings were listed in the baseline file (/Users/chrisbobbe/dev/zulip-flutter/android/app/lint-baseline.xml) but not found in the project; perhaps they have been fixed?

Note: The baseline was created using a different target/variant than it was checked against.
Creation variant: lint
Current variant: lintVitalRelease
```

chrisbobbe: (So, LGTM; there's a "Note" there that we might investigate if there were time, but it doesn't seem to be about something broken.)

gnprice: Sounds good; thanks for checking into it!

gnprice:
Made those changes, and rebased resolving conflicts. I'll let CI run on the result before merging.

chrisbobbe:
The CI failures are drift and pigeon; perhaps conflicts with the now-merged #1517 (which adds a setting) and #1379?

gnprice:
Hmm, thanks. Running this now:
`$ git rebase -i -x 'tools/check --fix --all-files drift pigeon'`
(and using the `-i` to delete the tools/check steps after commits that seem like they couldn't be affected)

gnprice:
OK, that command fixed it for Drift.

For Pigeon… it ended up actually reverting the changes (which are only to the Pigeon version number written at the top of the file) in the two files that were already updated in that commit:
android/app/src/main/kotlin/com/zulip/flutter/AndroidNotifications.g.kt
lib/host/android_notifications.g.dart

As we've seen a couple of times before, it seems like there's some unfortunate caching going on in the Pigeon generator, or perhaps in build_runner underneath it.

So I just edited the affected file directly to bump that version number.

Let's see how CI goes on this version.

chrisbobbe:
#622?

gnprice:
Yeah, that looks like it — thanks for pulling it up.

chrisbobbe:
Looks like CI passed! Please merge at will.

ios/Podfile.lock
rajveermalviya: Yes, correct.

gnprice: Thanks for confirming!

Issue #685, opened 2024-05-16
Try latency compensation on marking messages as read


gnprice:
As described in #81, when we implement marking messages as read when you scroll past them, we'll have the unread markers disappear optimistically without waiting for an ack from the server:

> * Unlike most everything else in the Zulip data model, apply "latency compensation" when marking messages as read on scroll, meaning eagerly apply the change to the local UI rather than wait for the server to acknowledge it with an event.
>   * Specifically, eagerly fade out the unread-message markers in the message list.

This issue is for going further, to eagerly update the model itself.

* The zulip-mobile RN app doesn't do this; the consequence is that it looks kind of glitchy as the "N unreads" banner at the top of the message list doesn't update or disappear in sync with the unread markers.
* I'm not actually sure what the web app does.
* The downside is that if we eagerly update the model, then it's no longer part of a coherent snapshot together with the rest of our model — we'd be giving up some of the consistency provided by the Zulip event system. 

There are a couple of approaches we could take to that consistency question, but we should think through the consequences.

When doing this, we should also do the same with the "mark as read" button at the bottom of the message list. As part of that, we'll make its fade-out animation somewhat longer again. See #613.

chrisbobbe:
[#mobile-team > last code changes for launch @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/last.20code.20changes.20for.20launch/near/2193892)
> There's one other gap in that implementation #F1574 relative to our original #F81 plan (and the legacy app), which is that it doesn't eagerly mark the messages as read locally, and instead waits for the event from the server.
> 
> In manual testing, the UX that way seems basically fine when you're on a good network like here in the office.
> 
> So I think that will be fine for launch. We'll file an M5b issue for adding that latency compensation.
> 
> (Adding it is potentially nontrivial without compromising performance, because of how our unreads data structure currently works. Totally solvable problem, but makes it a bigger task than I'd like to include between now and Monday.)

gnprice:
The status quo after #1574 (for #81) will go a bit less far than the original plan: unlike the legacy app, it won't do latency compensation on the unread-message markers within the message list, either. So from a user-impact perspective, this issue is for causing those to update eagerly, as well as everything else in the UI (like the "Mark as read" button at the bottom of the message list, and the inbox page, and so on).

[...]

PR #1574, opened 2025-06-13
Implement mark-as-read-on-scroll, v1

### LINKED ISSUE #81: Mark messages as read on scroll, configurably

chrisbobbe:
(edited)

Fixes #81. 🎉

Still to do, but can be soon-after-launch:

- #1583
- #685


gnprice:
Great, glad to see this progress.

> Can the "range hull" logic end up naively applying its arithmetic to two ranges that actually don't have anything to do with each other, because e.g. the narrow changed, and cause hundreds of messages to get marked as read incorrectly? (when is _modelChanged called?)

If the model's narrow changes, that will always cause its listeners to be notified. (It'd be a bug in the model if data like that could change without notifying listeners.) So in particular _modelChanged will get called.

To verify that listeners do always get called, look at the references to `_narrow` to see where it's mutated. (This is an example of where it comes in handy to have a different name for the field from the name for the getter.) They each… hmm, well, it looks like _handlePropagateMode may have a bug where it lacks that call. At the other sites, though, listeners get notified at the end of fetchInitial, via _setStatus.

lib/widgets/message_list.dart
```diff
@@ -503,24 +523,102 @@ class MessageList extends StatefulWidget {
...
+        }
+        return;
+      }
+      element.visitChildElements(visit);
+    }
+
+    final scrollViewElement = _scrollViewKey.currentContext as Element;
+    // TODO this frequently ends up walking through a few hundred elements,
+    //   only 5-10 of which end up being MessageItem ones.
+    scrollViewElement.visitChildElements(visit);
```
gnprice: Yeah, it should be possible to make this more focused so that it only goes down the relevant paths of the tree — only to the list items (be they MessageItem or date separator or recipient header), not to any of their descendants or to other parts of the tree. We can take a look together.

chrisbobbe:
Ah I do trust that _modelChanged will get called when the model's narrow changes, I'm just not totally sure the `_messagesRecentlyInViewport = null;` I added in _modelChanged will do what I want it to do. Like maybe: what if that runs, but a scroll event arrives and `_messagesRecentlyInViewport` gets set to some messages from the old narrow (because layout/etc. for the new narrow hasn't finished yet); then on the next scroll event, `_messagesRecentlyInViewport` won't have anything to do with the messages currently in view, and convex-hulling the two ranges will be problematic and mark some messages read incorrectly.

I might think more clearly about this in the office tomorrow 😛

lib/widgets/message_list.dart
```diff
@@ -503,24 +523,102 @@ class MessageList extends StatefulWidget {
...
+        result.add(message.id);
+        break;
+      }
+
+      final isMessageBottomVisible = messageBottom > 0 && messageBottom < viewportHeight;
+      if (isMessageBottomVisible) {
+        result.add(message.id);
+      }
+    }
+    // TODO why isn't it sorted already?
```
gnprice: This visits the items in tree order. Within each list sliver, the tree order of the items (their order as children of the sliver) is in the sliver's growth direction. So for the top sliver, the items appear from bottom to top, the reverse of their order as Zulip messages.

lib/widgets/message_list.dart
```diff
@@ -503,24 +523,102 @@ class MessageList extends StatefulWidget {
...
+
+  /// Which messages are onscreen.
+  ///
+  /// Ignores outbox messages.
+  ///
+  /// A message is considered onscreen if
+  /// - it covers the full height of the viewport
+  ///   (i.e. its top is above the viewport top
+  ///   and its bottom is below the viewport bottom) or
+  /// - its bottom is in the viewport (i.e. between the viewport top and bottom)
```
gnprice: Hmm, in the first case I'd think we'd rather wait to mark the message read until the user reaches the bottom of the message.

chrisbobbe: With the new "Open message feeds at" setting, It'll be more common than before to approach unread messages from below. There's also the point about how we'd like there to be at least one message in the result, so we can do the convex-hull thing from one result to the next result.

lib/model/message.dart
```diff
@@ -180,6 +182,37 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+      // TODO mark as read locally for latency compensation
+      //   (in Unreads and on the message objects)
+      try {
+        await updateMessageFlags(connection,
+          messages: batch,
+          op: UpdateMessageFlagsOp.add,
+          flag: MessageFlag.read);
+      } catch (e) {
+        // do nothing
+      }
```
gnprice: This part goes better if the messages only get removed from the queue after the request succeeds. (I believe that's what the legacy app does.)

That way if there's a network hiccup or something, those messages effectively get retried and eventually marked as read.

lib/widgets/message_list.dart
chrisbobbe: > Hmm, in the first case I'd think we'd rather wait to mark the message read until the user reaches the bottom of the message.

My more-awake brain today has come around to agreeing with this :)

timabbott:
When we merge this, we should revert the hunk from https://github.com/zulip/zulip/pull/34887#discussion_r2145924464.

chrisbobbe:
Revision pushed; Greg, PTAL!

lib/model/message.dart
```diff
@@ -180,6 +183,64 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+      for (final messageId in _markReadOnScrollQueue.iterable) {
+        if (toSend.length == _markReadOnScrollBatchSize) {
+          break;
+        }
+        i++;
+        final message = messages[messageId];
+        if (message == null) continue;
+        if (!message.flags.contains(MessageFlag.read)) {
+          toSend.add(message.id);
+        }
```
chrisbobbe: Equivalent and probably better, right:

```dart
        final message = messages[messageId];
        if (message != null && !message.flags.contains(MessageFlag.read)) {
          toSend.add(message.id);
        }
        i++;
```

lib/model/message.dart
```diff
@@ -180,6 +183,64 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+
+    _markReadOnScrollBusy = true;
+    do {
+      final toSend = <int>[];
+      int i = 0;
+      for (final messageId in _markReadOnScrollQueue.iterable) {
+        if (toSend.length == _markReadOnScrollBatchSize) {
+          break;
+        }
+        i++;
```
gnprice: nit: a more descriptive name is probably helpful for this; perhaps `numFromQueue`?

lib/model/message.dart
```diff
@@ -180,6 +183,64 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+
+      if (toSend.isEmpty || await _sendMarkReadOnScrollRequest(toSend)) {
+        if (_disposed) return;
+        _markReadOnScrollQueue.removeFirstN(i);
+      }
+      if (_disposed) return;
+
+      await Future<void>.delayed(_markReadOnScrollDebounceDuration);
+      if (_disposed) return;
+    } while (!_markReadOnScrollQueue.isEmpty);
```
gnprice: nit: isNotEmpty
```suggestion
    } while (_markReadOnScrollQueue.isNotEmpty);
```

lib/model/message.dart
```diff
@@ -180,6 +183,64 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+    }
+    return true;
+  }
+
+  @override
+  void markReadOnScroll(Iterable<int> messageIds) async {
+    assert(!_disposed);
+    final anyAdded = _markReadOnScrollQueue.add(messageIds);
+    if (!anyAdded) return;
+    if (_markReadOnScrollBusy) return;
```
gnprice: Is the anyAdded condition needed? If the queue is nonempty but the "busy" flag is false, then nothing's working on the queue; so then it seems good to start doing so, even if this call didn't happen to add anything new to the queue.

(Also it seems like a bug if that situation happens in the first place.)

gnprice: Without that condition, the `add` implementation can also get a bit simpler.

lib/model/message.dart
```diff
@@ -180,6 +183,64 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+
+  /// Returns true on success, false on failure.
+  Future<bool> _sendMarkReadOnScrollRequest(Iterable<int> toSend) async {
+    assert(toSend.isNotEmpty);
+
+    // TODO(#1581) mark as read locally for latency compensation
+    //   (in Unreads and on the message objects)
+    try {
+      await updateMessageFlags(connection,
+        messages: toSend.toList(),
```
gnprice: Looks like this method could take type `List` for toSend, and then no need to copy here. (The argument it gets passed is already a List, and isn't getting reused to be modified by anything else so there's no semantic need to copy it.)

lib/model/message.dart
```diff
@@ -180,6 +183,64 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+      if (toSend.isEmpty || await _sendMarkReadOnScrollRequest(toSend)) {
+        if (_disposed) return;
+        _markReadOnScrollQueue.removeFirstN(i);
+      }
+      if (_disposed) return;
+
+      await Future<void>.delayed(_markReadOnScrollDebounceDuration);
+      if (_disposed) return;
+    } while (!_markReadOnScrollQueue.isEmpty);
+    _markReadOnScrollBusy = false;
```
gnprice: Oh, speaking of the possibility of bugs causing surprising states. This clearing of the busy flag should be in a `finally`:
```suggestion
      } while (!_markReadOnScrollQueue.isEmpty);
    } finally {
      if (!_disposed) {
        _markReadOnScrollBusy = false;
      }
    }
```
with the `try` starting immediately after setting the flag:
```dart
    _markReadOnScrollBusy = true;
    try {
      do {
```

That way if this loop stops its work for any reason — in particular an unexpected exception — it'll always clear the flag, so that a future call can still try again.

lib/model/message_list.dart
```diff
@@ -222,6 +222,17 @@ mixin _MessageSequence {
...
 
+  List<Message>? getMessagesSublist(int firstMessageId, int lastMessageId) {
+    assert(firstMessageId <= lastMessageId);
+    final firstIndex = _findMessageWithId(firstMessageId);
+    final lastIndex = _findMessageWithId(lastMessageId);
+    if (firstIndex == -1 || lastIndex == -1) {
+      // TODO(log)
+      return null;
+    }
+    return messages.sublist(firstIndex, lastIndex + 1);
```
gnprice: This is making a new List by copying the given range of elements. But it looks like the one caller immediately consumes it as effectively an Iterable. So this could avoid the copy by instead using `List.getRange`.

gnprice (commented):
Exciting! Here's comments from a closer read just now, of the `model/message.dart` and `model/message_list.dart` changes.

lib/model/message.dart
gnprice: Sure.

lib/model/message.dart
```diff
@@ -517,6 +578,40 @@ class MessageStoreImpl extends PerAccountStoreBase with MessageStore, _OutboxMes
...
+
+  final _set = <int>{};
+  final _queue = QueueList<int>();
+
+  /// Add [messageIds] to the end of the queue,
+  /// if they aren't already in the queue.
+  ///
+  /// Returns true if some were added (i.e. weren't already in the queue),
+  /// false otherwise.
+  bool add(Iterable<int> messageIds) {
```
gnprice: nit: containers like List and Set have methods `add` to add one element and `addAll` to add from an iterable, so probably clearest to match that naming:
```suggestion
  bool addAll(Iterable<int> messageIds) {
```

chrisbobbe:
Thanks!! Revision pushed.

lib/model/message_list.dart
```diff
@@ -222,6 +222,17 @@ mixin _MessageSequence {
     return binarySearchByKey(items, messageId, _compareItemToMessageId);
   }
 
+  Iterable<Message>? getMessagesSublist(int firstMessageId, int lastMessageId) {
```
gnprice: nit: rename to match new semantics:
```suggestion
  Iterable<Message>? getMessagesRange(int firstMessageId, int lastMessageId) {
```

lib/main.dart
```diff
@@ -6,10 +6,12 @@ import 'log.dart';
 import 'model/binding.dart';
 import 'notifications/receive.dart';
 import 'widgets/app.dart';
+import 'widgets/message_list.dart';
 
 void main() {
   assert(() {
     debugLogEnabled = true;
+    MessageListPage.debugEnableMarkReadOnScroll = true;
```
gnprice: It doesn't feel great to me to have to set something like this in the main function, vs. having tests disable it.

Trying `flutter test` with the default flipped, it looks like there are just four test files with failing tests: autocomplete, action_sheet, compose_box, message_list (all in test/widgets/). So maybe just set it to false at the top of each of those?

If we did want to do it globally across all tests, there's a handy file `test/flutter_test_config.dart` for that. That's probably cleaner than here, because it means the default where the flag is declared is the default that applies in the live app.

(Potentially `debugLogEnabled` should go there too; I think I added that flag very early and may not have known about `flutter_test_config` at the time. But also it bothers me less to have it here than this flag would, because it's not in the user-facing functionality of the app.)

lib/widgets/message_list.dart
```diff
@@ -139,6 +140,14 @@ abstract class MessageListPageState {
...
   /// This is null if [MessageList] has not mounted yet.
   MessageListView? get model;
+
+  /// This view's decision whether to mark read on scroll,
+  /// overriding [GlobalSettings.markReadOnScroll].
+  ///
+  /// For example, this is set to false after pressing
+  /// "Mark as unread from here" in the message action sheet.
+  bool? get thisViewMarksReadOnScroll;
+  set thisViewMarksReadOnScroll(bool? value);
```
gnprice: Given that this has essentially the same semantics when not null as the global `markReadOnScroll`, just overrides the latter, I actually like the idea of giving it the same name `markReadOnScroll`.

That way matches the pattern `foo ?? theme.foo ?? _defaults.foo` that we use in some other places. It also means fewer names for essentially the same concept.

lib/widgets/message_list.dart
```diff
@@ -172,6 +181,34 @@ class MessageListPage extends StatefulWidget {
...
   State<MessageListPage> createState() => _MessageListPageState();
+
+  /// In debug mode, controls whether mark-read-on-scroll is enabled,
+  /// overriding [MessageListPageState.thisViewMarksReadOnScroll].
+  ///
+  /// Outside of debug mode, this is always true and the setter has no effect.
+  ///
+  /// Defaults to false, so we don't have to set to false all over our tests --
+  /// but main.dart sets `true` so the feature is enabled when you run the app.
+  static bool get debugEnableMarkReadOnScroll {
```
gnprice, line 190: For this one, on the other hand, it's good the name is different (with "enable") — a `true` here has quite a different meaning from in the other variables, and in particular true and false are asymmetric.

lib/widgets/message_list.dart
```diff
@@ -576,7 +713,44 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
...
+    model.store.markReadOnScroll(sublist.map((message) => message.id));
+
+    _messagesRecentlyInViewport = currentRange;
+  }
+
   void _handleScrollMetrics(ScrollMetrics scrollMetrics) {
+    if (
+      widget.thisViewMarksReadOnScroll
+      ?? GlobalStoreWidget.settingsOf(context).markReadOnScrollForNarrow(widget.narrow)
+    ) {
```
gnprice: Likely clearer if the logic that gives effect to `debugEnableMarkReadOnScroll` is consolidated here too, so they're all next to each other to make their relationships clear.

Perhaps as a method (or getter?) named like `_effectiveMarkReadOnScroll`.

lib/widgets/message_list.dart
```diff
@@ -576,7 +713,44 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
     _prevFetched = model.fetched;
   }
 
+  void _markReadOnScroll() {
```
gnprice: Hmmm but possibly confusing that this method (and the one on MessageStore) have the same name as the setting.

I think "on" isn't quite the right preposition here anyway — calling this method is a way of saying "please go and mark read, because scrolling happened", not "please mark read *if and when* scrolling happens".

Perhaps `_markReadFromScroll`? And similarly for the MessageStore method.

lib/widgets/message_list.dart
```diff
@@ -503,24 +551,102 @@ class MessageList extends StatefulWidget {
...
+  final GlobalKey _scrollViewKey = GlobalKey();
+
   MessageListView get model => _model!;
   MessageListView? _model;
 
   final MessageListScrollController scrollController = MessageListScrollController();
 
   final ValueNotifier<bool> _scrollToBottomVisible = ValueNotifier<bool>(false);
 
+  (int, int)? _messagesRecentlyInViewport;
```
gnprice: nit: move this next to _markReadOnScroll, as that's the code that's really responsible for maintaining it (and that consumes it)

lib/widgets/message_list.dart
```diff
@@ -503,24 +551,102 @@ class MessageList extends StatefulWidget {
...
+
+  /// Find the range of message IDs on screen, as a (first, last) tuple,
+  /// or null if no messages are onscreen.
+  ///
+  /// A message is considered onscreen if its bottom edge is in the viewport.
+  ///
+  /// Ignores outbox messages.
+  (int, int)? _findMessagesInViewport() {
```
gnprice, line 646: nit: move this and the other new method `_isMessageItemInViewport` down below the lifecycle methods, and below their close partner `_modelChanged`; I guess that puts them right near `_markReadOnScroll` and the existing scroll-handling methods

lib/widgets/message_list.dart
```diff
@@ -503,24 +551,102 @@ class MessageList extends StatefulWidget {
...
+            if (message.id < first!) {
+              first = message.id;
+            }
+            if (last! < message.id) {
+              last = message.id;
+            }
+          }
+          return; // no need to look for more MessageItems inside this one
+      }
+      element.visitChildElements(visit);
```
gnprice: nit: this is equivalent, right?
```suggestion
      
        default:
          element.visitChildElements(visit);
      }
```

Then it's more immediately clear that the recursion is only for widgets not among the cases above.

lib/widgets/message_list.dart
```diff
@@ -503,24 +551,102 @@ class MessageList extends StatefulWidget {
...
+  (int, int)? _findMessagesInViewport() {
+    final scrollViewElement = _scrollViewKey.currentContext as Element;
+    final scrollViewRenderObject = scrollViewElement.renderObject as RenderBox;
+
+    int? first;
+    int? last;
+    void visit(Element element) {
+      final widget = element.widget;
+
+      switch (widget) {
```
gnprice: nit: these seem closely tied together
```suggestion
      final widget = element.widget;
      switch (widget) {
```

lib/widgets/message_list.dart
```diff
@@ -503,24 +551,102 @@ class MessageList extends StatefulWidget {
...
+        case RecipientHeader():
+        case DateSeparator():
+        case MarkAsReadWidget():
+          // MessageItems won't be descendants of these
+          return;
+
+        case MessageItem():
+          if (widget.item is! MessageListMessageItem) return; // ignore outbox
+
```
gnprice: How about:
```suggestion
        case MessageItem(item: MessageListOutboxMessageItem()):
          return; // ignore outbox

        case MessageItem(item: MessageListMessageItem(:final message)):
```

Then the later line
```dart
            final message = (widget.item as MessageListMessageItem).message;
```
can be dropped.

gnprice (commented):
OK, and here's comments on the widgets/message_list.dart changes. All the suggested changes here are NFC.

lib/widgets/settings.dart
```diff
@@ -150,6 +151,86 @@ class VisitFirstUnreadSettingPage extends StatelessWidget {
...
+    return Scaffold(
+      appBar: AppBar(title: Text(zulipLocalizations.markReadOnScrollSettingTitle)),
+      body: Column(children: [
+        ListTile(title: Text(zulipLocalizations.markReadOnScrollSettingDescription)),
+        for (final value in MarkReadOnScrollSetting.values)
+          RadioListTile.adaptive(
+            title: Text(_valueDisplayName(value,
+              zulipLocalizations: zulipLocalizations)),
+            subtitle: () {
+              final result = _valueDisplayDescription(value,
```
gnprice: nit: "description" seems enough to make "display" redundant:
```suggestion
          RadioListTile.adaptive(
            title: Text(_valueDisplayName(value,
              zulipLocalizations: zulipLocalizations)),
            subtitle: () {
              final result = _valueDescription(value,
```

(by contrast "name" could mean an identifier-style name used in the code, or the name displayed to the user)

lib/widgets/settings.dart
```diff
@@ -150,6 +151,86 @@ class VisitFirstUnreadSettingPage extends StatelessWidget {
   }
 }
 
+class _MarkReadOnScrollSetting extends StatelessWidget {
```
gnprice, line 154: I think this main commit could productively be split in two:
* A final commit adds the setting.
* A preceding commit adds all the logic in model/message and model/message_list and widgets/message_list and so on which the setting will enable, but behaves as if the setting were always false/never.

If we were writing tests for this synchronously, then after the first commit the logic would get exercised by the tests, just never in the live app. Potentially that would require a small extra mechanism, deleted in the second commit, to allow turning the behavior on. (I think not even that, though — the tests could set `thisViewMarksReadOnScroll` to true.)

The main value of this split is that the setting is a fair amount of code, and all of a quite different character from most of the rest of the changes.

lib/main.dart
chrisbobbe: > it looks like there are just four test files with failing tests: autocomplete, action_sheet, compose_box, message_list (all in test/widgets/)

(I also get failures in lightbox and emoji_reaction tests; those are easy to treat the same way)

gnprice: Cool. Yeah, what I did was:
* Run `flutter test`.
* Scan the 158 resulting test failures by eye as they scroll past (each of which is like 20 lines of output, designed to be enough information to actually investigate the failure), and try to spot all the different filenames that appear.

So fairly error-prone.

(I'm not sure `flutter test` / `dart test` has a better way out of the box for collecting more of a summary of which tests failed, without all that detail about each failure. There are some verbosity controls, but I think only for turning *up* the verbosity on non-failing tests.)

chrisbobbe:
Thanks! Revision pushed, and it's time for bed for me :)

lib/widgets/message_list.dart
```diff
@@ -666,6 +840,11 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
     final topSliver = SliverStickyHeaderList(
       headerPlacement: HeaderPlacement.scrollingStart,
       delegate: SliverChildBuilderDelegate(
+        // None of the children ever need to keep themselves alive;
+        // opt out of adding AutomaticKeepAlive widgets that we don't need.
+        // (This helps keep _findMessagesInViewport efficient.)
+        addAutomaticKeepAlives: false,
```
gnprice: Looking into this a bit, I'm nearly convinced we never currently need keep-alive in these list items.

I'm less sure we'll never want it in the future — for example, being able to select text in a message is a repeatedly-requested feature, and selecting text would cause the child to try to keep itself alive. And I'm not entirely sure what the consequences would be of a widget getting discarded after all when it tried to keep itself alive. So this seems like potentially a pitfall for future changes, when we might not be thinking about this feature.

The way I looked for things that would invoke this is: it seems it's all about receiving KeepAliveNotification instances. I looked for call sites of that class's constructor; there's only one, in AutomaticKeepAliveClientMixin. So then I looked for subclasses of that mixin, and there are 5 of them:
_DismissibleState
EditableTextState
_InkResponseState
_SelectionKeepAliveState
_CupertinoTextFieldState

So selection, text fields (plain, Material which uses plain, and Cupertino), InkWell, and Dismissible which is about items the user can swipe out of a list.

Some of those are unlikely to ever appear, like Dismissible. But for InkWell, I'm not 100% sure we don't already have something in the message list that uses it.

---

When we measured the timing, the walk of the element tree took a very short amount of time even in debug mode. With this flag true (the default), it'll have roughly double as many elements to walk through, but that'll still make a negligible amount of time. So let's leave this commit out for launch.

gnprice:
And merging! After dropping that `addAutomaticKeepAlives: false` commit:
2cbb71d78 msglist: Opt out of adding AutomaticKeepAlive widgets to the tree

PR #1575, opened 2025-06-13
msglist: Say "Quote message" for the quote-and-reply button, following web

chrisbobbe:
Thanks Alya for pointing this out:
  https://chat.zulip.org/#narrow/channel/48-mobile/topic/quote.20and.20reply.20-.3E.20quote.20message/near/2193484


gnprice:
Thanks! Looks good; merging.

PR #1577, opened 2025-06-13
logo: Switch app icons to non-"beta" versions

rajveermalviya:
| | Launcher | App Switcher |
| - | - | - |
| Android | Screenshot 2025-06-14 at 00 06 10 | Screenshot 2025-06-13 at 23 17 41 |
| iOS | Screenshot 2025-06-14 at 00 03 06 | Screenshot 2025-06-13 at 23 17 50 |

Fixes: #1537

Issue #715, opened 2024-05-31
Launch-time tasks


gnprice:
This is an issue for recording things we'll want to do around the time of launching the app to GA (replacing the legacy zulip-mobile app), and that we'd otherwise be at risk of forgetting.

- [ ] [Update the names seen in /stats](https://chat.zulip.org/#narrow/stream/378-api-design/topic/zulip-flutter.20in.20analytics.20client.20chart/near/1808961) (currently "Mobile app (React Native)" and "Mobile app beta (Flutter)")
- (There'll be lots of other things to do too; we'll fill out a list of them more fully when the launch gets closer. For now this is only to track things we think of now and feel we should write down in order not to forget.)

gnprice:
The launch is imminent; and the tasks above are all done! (Or at least prepared, with the final deploy steps tracked on short-term checklists.)

PR #1585, opened 2025-06-14
notif: Use a different channel ID from previous values of zulip-mobile

rajveermalviya:
Previous values list is sourced from here:
      https://github.com/zulip/zulip-mobile/blob/eb8505c4a/android/app/src/main/java/com/zulipmobile/notifications/NotificationChannelManager.kt#L22-L24

### DIFF
--- a/lib/notifications/display.dart
+++ b/lib/notifications/display.dart
@@ -37,11 +37,12 @@ enum NotificationSound {
 class NotificationChannelManager {
   /// The channel ID we use for our one notification channel, which we use for
   /// all notifications.
-  // TODO(launch) check this doesn't match zulip-mobile's current or previous
-  //   channel IDs
-  // Previous values: 'messages-1'
+  // Previous values from Zulip Flutter Beta:
+  //   'messages-1'
+  // Previous values from Zulip Mobile:
+  //   'default', "messages-1', (alpha-only: 'messages-2'), 'messages-3'
   @visibleForTesting
-  static const kChannelId = 'messages-2';
+  static const kChannelId = 'messages-4';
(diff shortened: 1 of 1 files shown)


gnprice:
Ah good thought, thanks. Will include this before shipping a version with #1586.

lib/notifications/display.dart
```diff
@@ -37,11 +37,12 @@ enum NotificationSound {
 class NotificationChannelManager {
   /// The channel ID we use for our one notification channel, which we use for
   /// all notifications.
-  // TODO(launch) check this doesn't match zulip-mobile's current or previous
```
gnprice, line 40: And I just checked with a grep, and there are no other `TODO(launch)` comments in the tree :slightly_smiling_face:.

(And the only TODO(#715) comments are the ones resolved by #1577.)

lib/notifications/display.dart
```diff
@@ -37,11 +37,12 @@ enum NotificationSound {
...
-  //   channel IDs
-  // Previous values: 'messages-1'
+  // Previous values from Zulip Flutter Beta:
+  //   'messages-1'
+  // Previous values from Zulip Mobile:
+  //   'default', "messages-1', (alpha-only: 'messages-2'), 'messages-3'
```
gnprice: nit:
```suggestion
  //   'default', 'messages-1', (alpha-only: 'messages-2'), 'messages-3'
```

gnprice:
Merging (see https://github.com/zulip/zulip-flutter/pull/1586#issuecomment-2974679247), with that nit fixed.

Issue #1582, opened 2025-06-13
Switch app ID to those of the main app


gnprice:
As part of launching this app as the new Zulip mobile app (cf #715), we'll need to switch the app ID it uses to the app IDs we use for the main Zulip listings in the Play Store and App Store respectively.

Those are `com.zulipmobile` on Android and `org.zulip.Zulip` on iOS. To date, this app has been using `com.zulip.flutter`.

(The values we'll switch to are a bit funny-looking, yes. This is an aspect that's completely invisible to users, and very costly to change once an app is published, so early legacy values persist forever.)

### Implementation

In addition to being the app ID on all platforms, "com.zulip.flutter" is also the JVM "package" name of our Kotlin code. I believe we can leave that unchanged, and that changing it would be somewhat more complex, so let's try to leave that unchanged. In particular I think this means that in `android/app/build.gradle` we'll change `applicationId` to be "com.zulipmobile", but leave `namespace` as "com.zulip.flutter".

The name "com.zulip.flutter" is also the app ID of the Linux and macOS apps, and I don't see a reason to change those.

In general, run `git grep com.zulip.flutter` and look at each of the results.  Many of those instances will change, and some won't.

PR #1586, opened 2025-06-14
Switch app ID to those of the main app

rajveermalviya:
Fixes: #1582

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -38,7 +38,7 @@ android {
     }
 
     defaultConfig {
-        applicationId "com.zulip.flutter"
+        applicationId "com.zulipmobile"
         minSdkVersion 28
         targetSdkVersion flutter.targetSdkVersion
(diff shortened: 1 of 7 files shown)


gnprice:
Excellent, thanks! Will look at this today, and use it for preparing a beta with #1070.

gnprice:
Cool, this all looks good. I'll include it in assembling a branch for switching to be the main app.

gnprice:
OK, I think we are done making releases as the Flutter beta app! So I'll merge this, and before it #1577 and #1585, to main.

If we do find ourselves wanting to make another release as `com.zulip.flutter` in the future, we can:
* make a branch atop the current main, i.e. from before I merge these three PRs:
  f8677accc version: Sync version and changelog from v0.0.33 release
* or make a branch starting from v0.0.33, the last pre-launch release
* or if we want such a release incorporating significant changes that have happened between now and that future, then make a branch that switches back the app IDs and perhaps app name and app icon.

PR #1588, opened 2025-06-15
Migrate data from legacy app

### LINKED ISSUE #1070: Read legacy app's database, and migrate data

gnprice:
Fixes #1070.

This is a draft because I'd like to write at least some tests before merging. (Unlike some of the other PRs we've merged in the past week with tests left as a follow-up, the job this is doing is especially sensitive to edge cases.)

But otherwise I believe this is ready. I've tested it manually on both Android and iOS, and the data gets carried along as intended. So in parallel with me writing those tests, I'd like to have additional eyes reading the code, to try to spot things I might have missed.


lib/model/legacy_app_data.dart
```diff
@@ -0,0 +1,222 @@
...
+
+/// This and its subclasses correspond to portions of src/storage/replaceRevive.js .
+///
+/// (The rest of the conversions in that file are for types that don't appear
+/// in the portions of the legacy app's state we care about.)
+sealed class _LegacyAppJsonConverter<T> extends JsonConverter<T, Map<String, dynamic>> {
+  const _LegacyAppJsonConverter();
+
+  static const List<_LegacyAppJsonConverter<Object?>> converters = [
+    _LegacyAppUrlJsonConverter(),
```
PIG208: I can see why `_LegacyAppZulipVersionJsonConverter` got left out here, since that converter has `T` as `String`. A comment explaining that this is intentional might be helpful.

(Or perhaps just inline the `_LegacyAppUrlJsonConverter` annotation when needed?)

PIG208:
I probably haven't set this up properly. On my Android debug build, `LegacyAppDatabase._filename` returns `/data/user/0/com.zulip.flutter/files/SQLite/zulip.db`. That path, with "com.zulip.flutter" replaced with "com.zulipmobile", does not exist; instead, I got `Android/data/com.zulipmobile/files` (which I think I don't have access to).

lib/model/legacy_app_data.dart
```diff
@@ -8,13 +8,83 @@ library;
...
+    return;
+  }
+
+  assert(debugLog("Found settings: ${legacyData.settings?.toJson()}"));
+  final settings = legacyData.settings;
+  if (settings != null) {
+    await db.update(db.globalSettings).write(GlobalSettingsCompanion(
+      // TODO(#1139) apply settings.language
+      themeSetting: switch (settings.theme) {
+        LegacyAppThemeSetting.default_ => drift.Value.absent(),
```
PIG208, line 46: I think the "default" in the legacy app [is light mode](https://github.com/zulip/zulip-mobile/blob/eb8505c4a0ccfb7bdd817ca9022c11ce57ab69d8/src/settings/settingsSelectors.js#L5-L6).

There was a migration to rename "default" to "light" (https://github.com/zulip/zulip-mobile/commit/43acc4199919d5ea8ece1d00ac6e1243b4073e67). Oh, but that was ultimately for https://github.com/zulip/zulip-mobile/issues/5533 (Automatically follow user's OS-level dark/light preference). So I think, after all, this completes:

> On existing installs where the user has stuck with the default of light mode, we'll give them the new default.

gnprice:
Thanks for the review!

For investigating the file paths, I started a chat thread: [#mobile-team > migrating legacy data @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/migrating.20legacy.20data/near/2194721). Let's continue there for debugging.

(I'll also reply on the other subthreads above, but that one is the most critical since it could affect whether the migration works at all for a given user.)

lib/model/legacy_app_data.dart
gnprice: Yeah, this is a good question to consider explicitly.

The legacy app has just two values: always light and always dark. In the app's data store, and so in the database, these are called "default" and "night" respectively (that's the `ThemeSetting` enum cited in a comment here). And the default is "default" (i.e. thankfully the name isn't misleading on that point).

The new app has values `light`, `dark`, and null. Null means to use the system-wide setting, and is the default.

So if the value in the legacy app's database is "night", clearly we should translate that to `dark`. The user explicitly chose a dark theme.

If the value in the legacy app's database is "default", though, it's less clear. That's the value it will have if the user didn't make any choice.

So my thinking is that if the value is "default", then we should translate that to the new default, which is to use the system-wide setting. If the system-wide setting is on a daily cycle, then the user will almost certainly experience that as an upgrade: at last Zulip follows the daily cycle the rest of the device follows. If the system-wide setting is light, there'll be no change. And if the system-wide setting is dark, but the user really does prefer Zulip specifically to be in a light theme… well, it's easy for them to change it.

lib/model/legacy_app_data.dart
gnprice: Yeah, this could be clearer. I started with this one; then got to where `ZulipVersion` is needed and the same approach doesn't apply.

Given that there ended up being just one place `_LegacyAppUrlJsonConverter` comes into effect (for `LegacyAppAccount.realm`), and that this set of types is basically complete now and isn't going to expand in the future, I think I'll just apply the converter explicitly in that one spot and drop this list.

lib/model/legacy_app_data.dart
gnprice: Ah, and GitHub was showing me a stale page that didn't have your edit. Yeah, exactly, this effectively completes that issue https://github.com/zulip/zulip-mobile/issues/5533 .

lib/model/legacy_app_data.dart
```diff
@@ -0,0 +1,464 @@
...
+      // (See docs and comments on these properties in src/types.js .)
+      // Drop the account.
+      assert(debugLog("    (account ignored because missing metadata)"));
+      continue;
+    }
+    await db.createAccount(AccountsCompanion.insert(
+      realmUrl: account.realm,
+      userId: account.userId!,
+      email: account.email,
+      apiKey: account.apiKey,
```
PIG208, line 96: Hmm, what happens if `apiKey` is an empty string ([possible for logged-out users](https://github.com/zulip/zulip-mobile/blob/eb8505c4a0ccfb7bdd817ca9022c11ce57ab69d8/src/api/transportTypes.js#L21-L27))?

I think the converted `Account` will show up on the choose-account page, and will get handled by the code responsible for invalid API keys. The Flutter app doesn't remember logged-out accounts right now, similar to how language setting is not yet supported. Perhaps this can be a TODO?

gnprice: Hmm, yeah, good find.

The new app doesn't have a concept of an account that's in the list and yet logged out.

I'm not sure it was all that useful in the legacy app either — it's basically a relic of the fact that, when you were a screen or two into the login flow, the legacy app didn't originally have a better place to keep that information (with the realm URL, and another screen later the email and password, that you'd already entered) than in its global persistent list of accounts. We eventually fixed that, making that information part of the relevant pages' local state instead, and the new app does the same thing.

Since the new app doesn't have this concept, I'll have it drop such records if they appear.

gnprice:
Revision pushed. Thanks @PIG208 for the review!

lib/model/legacy_app_data.dart
```diff
@@ -8,13 +8,99 @@ library;
...
+      zulipVersion: account.zulipVersion!,
+      // no zulipMergeBase; legacy app didn't record it
+      zulipFeatureLevel: account.zulipFeatureLevel!,
+      ackedPushToken: drift.Value(account.ackedPushToken),
```
chrisbobbe: This will be the first time we store a non-null `ackedPushToken` in the new app, since we haven't done #322 yet. We don't update it when the server registers a new push token, so it might become stale…currently I think such a stale token could only flow into an unregister-token request, which we do on logout.

gnprice: Hmm interesting, good catch.

I think I'll drop this value, then — it doesn't really do any good (since until #322 it doesn't cause this app to stop repeatedly sending the token to the server), and at that one spot where it would get used I think it would be more likely to make the unregister-token request not work (by making it use a stale token) than to make it work. (Both of which are unlikely — I believe these tokens rarely change in practice.)

When we eventually do #322, that will cause the app to promptly store the current value, and there won't be any benefit to remembering the value the legacy app had stored in the past.

gnprice:
(Pushed a revision with that change.)

gnprice:
OK, at this point I'm not going to get to writing any significant automated tests for this before the launch. So I'll file a followup issue for those, and go ahead and merge after tweaking the commit messages.

(I also just pushed a revision expanding one comment, with details on when the legacy app's Account records started having a `userId`.)

Effectively we'll just rely on
* the manual testing I did before sending this
* the scrutiny of the code by @PIG208 and @chrisbobbe above (plus my own self-review before that)
* and the manual testing we [got today](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/last.20code.20changes.20for.20launch/near/2194853) from alpha users.

Speaking of that alpha testing, I have a couple more edits to harden the logic that I want to include for launch. I'll send those as a separate follow-up PR, though — that way the merged version of this in main can be the same logic that went into today's v30.0.257 beta, and the difference between the logic in v30.0.257 and the next build (for launch) can be expressed as a discrete commit of its own. That's helpful for looking at the history in Git to understand the history of the releases.

gnprice:
And merged. Thanks again @PIG208 and @chrisbobbe for the reviews!

PR #1590, opened 2025-06-15
Welcome dialog on upgrade from legacy app

### LINKED ISSUE #1580: Show a welcome dialog on upgrading from legacy app
When this app launches for the first time in an install that previously held the legacy app (following the launch, cf #715), we should show a brief dialog box to acknowledge the obvious that the app will look quite different, and to offer more information.

Here's text for the dialog box (written by @alya):

> ### Welcome to the new Zulip app!
>
> You’ll find a familiar experience in a faster, sleeker package. Check out the [announcement blog post](https://)!
>
> [Let’s go]

The link will be to a URL under `https://blog.zulip.com/…` which we've picked in advance, and which will redirect to the blog post once it goes up. (The blog post won't yet be up when we're preparing the release build, so its actual URL won't yet be final, hence the indirection.)

A key step in this issue is to determine whether the app was upgraded from the legacy app. We'll have that information as a side effect of #1070, so I expect to implement this issue next after that one.

gnprice:
Fixes #1580.

Stacked atop #1588.

I believe this works correctly. Some tests would be nice to write, but are less critical than for #1070 / #1588.

### Screenshots
| dark | light |
| --- | --- |
| Screenshot_20250614_235102 | Screenshot_20250614_235205 |

### DIFF
--- a/lib/model/settings.dart
+++ b/lib/model/settings.dart
@@ -119,6 +119,9 @@ enum GlobalSettingType {
   /// we give it a placeholder value which isn't a real setting.
   placeholder,
 
+  /// Describes a pseudo-setting not directly exposed in the UI.
+  internal,
+
   /// Describes a setting which enables an in-progress feature of the app.
   ///
   /// Sometimes when building a complex feature it's useful to merge PRs that
@@ -169,6 +172,10 @@ enum BoolGlobalSetting {
   /// That way when we remove those, this is already here.
   /// (Having one stable value in this enum is also handy for tests.)
   placeholderIgnore(GlobalSettingType.placeholder, false),
+
+  /// A pseudo-setting recording whether the user has been shown the
+  /// welcome dialog for upgrading from the legacy app.
+  upgradeWelcomeDialogShown(GlobalSettingType.internal, false),
 
   /// An experimental flag to toggle rendering KaTeX content in messages.
   renderKatex(GlobalSettingType.experimentalFeatureFlag, false),
--- a/lib/widgets/app.dart
+++ b/lib/widgets/app.dart
@@ -160,6 +160,7 @@ class _ZulipAppState extends State<ZulipApp> with WidgetsBindingObserver {
   void initState() {
     super.initState();
     WidgetsBinding.instance.addObserver(this);
+    UpgradeWelcomeDialog.maybeShow();
   }
 
   @override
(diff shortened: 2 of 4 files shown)


gnprice:
@alya for review on the screenshots. (See thread in #1580 for why the link is slightly different from the original plan.)

gnprice:
I've pushed a revision that adjusts what URL the link goes to:
```dart
  static const String _announcementUrl =
    // TODO(#1580) switch to this URL, which will point to the actual blog post:
    //   'https://blog.zulip.com/flutter-mobile-app-launch';
    // For the moment, that's a 404, so point directly to this pre-announcement:
    'https://chat.zulip.org/#narrow/channel/1-announce/topic/New.20mobile.20app.20launching.20this.20week/near/2194879';
```

We'll want to change that before the actual launch, once the desired URL works. (Which will be before the actual post is up; it can redirect to that same pre-announcement instead, and change to redirecting to the post once the post is up.)

gnprice:
CI failure in `l10n`:
```
Running l10n...
There are uncommitted changes in lib/generated/l10n/zulip_localizations*.dart:

 M lib/generated/l10n/zulip_localizations_de.dart

Aborting, to avoid losing your work.
```

Odd. And that's an *uncommitted* change, so left around by some previous step.

Seems harmless in itself, so I'll ignore for now given the impending launch. If it recurs it'll be good to debug.

gnprice:
And updated that URL; the desired URL works now :slightly_smiling_face:

gnprice:
Well, that CI failure repeated the same way after my next few pushes after seeing it. But with the recent push which rebased this and its dependency #1588 past these commits in main:
67af2ed14 logo: Switch app icons to non-"beta" versions
2ab189d2f (pr/1577) android, ios: Update app name to "Zulip", from "Zulip beta"
40046c85d (pr/1585) notif: Use a different channel ID from previous values of zulip-mobile
a4a96a1c5 notif: Use PackageInfo to generate sound resource file URL
e0e51b1ab android: Switch app ID to that of the main app
378e18758 (pr/1586) ios: Switch app ID to that of the main app
5be0e70ce l10n: Add translatable strings from welcome dialog in v30.0.256 release
fec0071a0 (pr/1591) l10n: Update translations from Weblate.
a6ea88209 version: Sync version and changelog from v30.0.256 release

CI now passes :shrug:.

Of those, the two commits that touched any files in assets/l10n/ or lib/generated/l10n/ were:
5be0e70ce l10n: Add translatable strings from welcome dialog in v30.0.256 release
fec0071a0 l10n: Update translations from Weblate.

so presumably the reason involved one of those.

chrisbobbe:
LGTM, thanks, please merge at will! 🚀

Issue #1593, opened 2025-06-16
Write tests for legacy-data migration


gnprice:
At launch time, the implementation of #1070 that we'll have (mainly from #1588) won't include automated tests of its own.

That logic got extensive manual testing, plus careful scrutiny of the code. But as the codebase changes over time, the informativeness of that manual testing degrades — unless repeated, which is a lot of work. That also creates a barrier to editing the relevant code, for fear of either breaking it or repeating a bunch of manual testing.

So it'd be good to write some tests for it.

gnprice:
At this point, the great majority of the users who will ever run this code have already done so.

We've also never yet touched this code since it was first merged. (The closest was c29934a60e44be45f1b52c7a67b1b114c1f90ec9, adding a pair of comment lines explaining why it hadn't changed any more than that.) It's not clear we ever will, or anyway will do so beyond very small changes in the code's very last phase, to reflect changes to the table in this app's database which this code writes into. Other than that spot, this code has very few points where it could be affected by changes we make in the rest of the app.

So I think it won't make sense to invest time pre-emptively in writing tests for it. If in the future we do end up making substantive changes for some reason, we can invest then in writing any tests we think are needed.

PR #1595, opened 2025-06-16
More error tolerance in migrating legacy data

gnprice:
In particular, if the legacy app's account list contains multiple entries with the same realm and user ID (though different emails), we'll drop all but the most-recently-used one because this app's accounts table keeps (realmUrl, userId) unique. With this change, we still drop the duplicates, but we catch the exception and proceed smoothly to the next account in the list.

Ideally these changes would be tested. We'll leave that as part of the followup issue #1593, which covers this whole file.


chrisbobbe:
LGTM, thanks, please merge at will! 🚀

PR #1601, opened 2025-06-17
KaTeX (4/n): Ignore classes that don't have CSS definition

rajveermalviya:
Stacked on top of #1609.

### DIFF
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -357,6 +357,11 @@ class _KatexParser {
         case 'mop':
         case 'mclose':
         case 'minner':
+        case 'mbin':
+        case 'mpunct':
+        case 'nobreak':
+        case 'allowbreak':
+        case 'mathdefault':
           // Ignore these classes because they don't have a CSS definition
           // in katex.scss, but we encounter them in the generated HTML.
           break;
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -364,6 +364,12 @@ class _KatexParser {
         case 'mathdefault':
           // Ignore these classes because they don't have a CSS definition
           // in katex.scss, but we encounter them in the generated HTML.
+          // (Why are they there if they're not used?  The story seems to be:
+          // they were used in KaTeX's CSS in the past, before 2020 or so; and
+          // they're still used internally by KaTeX in producing the HTML.
+          //   https://github.com/KaTeX/KaTeX/issues/2194#issuecomment-584703052
+          //   https://github.com/KaTeX/KaTeX/issues/3344
+          // )
           break;
 
         default:


gnprice:
Thanks!

The new commit here atop #1559:
cc657f1e3 content: Ignore classes that don't have CSS definition

looks like it's pretty simple and should be fairly easy to merge. After you split up #1452 as described at https://github.com/zulip/zulip-flutter/pull/1452#issuecomment-2985792163, would you rebase this to be atop that first segment?

(It looks like it'd conflict with one of the commits there, so that seems like the simplest way to handle the conflict.)

If these PRs were new, I'd suggest instead squashing this into the similar commit in #1452. But given that previous revisions of #1452 have been out in releases now, I think it'll be good to avoid adding new functionality to #1452, and put it in new PRs instead, like this one — that'll help us keep track of what's changed from one release in a crisp way.

rajveermalviya:
Updated to add `mathdefault` class, as discussed here: [#mobile-team > KaTeX survey results @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/KaTeX.20survey.20results/near/2202488).

gnprice:
I've included this PR in today's release 259: [#announce > mobile releases @ 💬](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2202921). (I cherry-picked its one commit into the release branch; the previous commits' changes were included in the previous release 258, as part of #1452.)

To copy into this thread what we found in chat ([#mobile-team > KaTeX survey results @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/KaTeX.20survey.20results/near/2200833)) using the survey script #1600: on a corpus from an open Zulip community about math,
* Before this change, in release v30.0.258, about 45% of messages using TeX were fully supported.
* After this change, so in release v30.0.259, about 80% of messages using TeX are fully supported.

lib/model/katex.dart
```diff
@@ -329,12 +345,24 @@ class _KatexParser {
...
+        case 'text':
+        case 'mrel':
+        case 'mop':
+        case 'mclose':
+        case 'minner':
+        case 'mbin':
+        case 'mpunct':
+        case 'nobreak':
+        case 'allowbreak':
+        case 'mathdefault':
```
gnprice, line 364: This change looks good, but leaves it still a bit mysterious why these classes are there in the KaTeX output if we really don't need to do anything about them.

As it happens, while I was reviewing #1609 earlier today I came across a couple of threads in the KaTeX tracker that basically answer that question. So I've just pushed a commit on top which adds a comment linking to those:
cbeef340a content [nfc]: Explain why some KaTeX CSS classes are unused in its CSS

gnprice:
This is ready to merge as soon as #1609 is merged.

gnprice:
And merged, after rebasing atop the merged #1609.

Issue #1583, opened 2025-06-13
Write tests for mark-read-on-scroll


chrisbobbe:
This is a follow-up to #81. For the launch we'll be shipping an implementation (from #1574) that doesn't yet have tests, in the interest of timeliness. It'll be good to follow up by writing those tests.

gnprice:
I think at this point we can defer the writing of these tests until the next time we're engaged in making changes to this subsystem. As a result, it's not a task that we benefit from tracking in the issue tracker.

Issue #1571, opened 2025-06-12
Write tests for msglist opening at first unread, and setting


gnprice:
This is a follow-up to #80. For the launch we'll be shipping an implementation (from #1517) that doesn't yet have tests, in the interest of timeliness. It'll be good to follow up by writing those tests.

The code to be tested can be seen by looking at the one commit in #1517. Before merging that PR, I also intend to edit it to add TODO comments, pointing at this issue, in the relevant areas of our test files.

gnprice:
I think at this point we can defer the writing of these tests until the next time we're engaged in making changes to this subsystem. As a result, it's not a task that we benefit from tracking in the issue tracker.

As mentioned above, there are some TODO comments in the test files to outline what areas are untested. So those at least provide some notice of the gaps when one is working in those files.

PR #1602, opened 2025-06-17
settings: Migrate to new RadioGroup API; write widget tests for two just-added settings

### LINKED ISSUE #1545: Resolve deprecation warnings about radio buttons

chrisbobbe:
Fixes #1545.

Fixes part of:
- #1571
- #1583

### DIFF
--- a/test/widgets/settings_test.dart
+++ b/test/widgets/settings_test.dart
@@ -26,11 +26,19 @@ void main() {
     await tester.pump();
   }
 
-  group('ThemeSetting', () {
-    Finder findRadioListTileWithTitle(String title) => find.ancestor(
-      of: find.text(title),
-      matching: find.byType(RadioListTile<ThemeSetting?>));
+  Finder findRadioListTileWithTitle<T>(String title) => find.ancestor(
+    of: find.text(title),
+    matching: find.byType(RadioListTile<T>));
 
+  void checkRadioButtonAppearsChecked<T>(WidgetTester tester, String title, bool expectedIsChecked) {
+    check(tester.semantics.find(findRadioListTileWithTitle<T>(title)))
+      .containsSemantics(
+        label: title,
+        isInMutuallyExclusiveGroup: true,
+        hasCheckedState: true, isChecked: expectedIsChecked);
+  }
+
+  group('ThemeSetting', () {
     void checkThemeSetting(WidgetTester tester, {
       required ThemeSetting? expectedThemeSetting,
     }) {
@@ -40,12 +48,7 @@ void main() {
         ThemeSetting.dark => 'Dark',
       };
       for (final title in ['System', 'Light', 'Dark']) {
-        final expectedIsChecked = title == expectedCheckedTitle;
-        check(tester.semantics.find(findRadioListTileWithTitle(title)))
-          .containsSemantics(
-            label: title,
-            isInMutuallyExclusiveGroup: true,
-            hasCheckedState: true, isChecked: expectedIsChecked);
+        checkRadioButtonAppearsChecked<ThemeSetting?>(tester, title, title == expectedCheckedTitle);
       }
       check(testBinding.globalStore)
         .settings.themeSetting.equals(expectedThemeSetting);
@@ -60,13 +63,13 @@ void main() {
       check(Theme.of(element)).brightness.equals(Brightness.light);
       checkThemeSetting(tester, expectedThemeSetting: ThemeSetting.light);
 
-      await tester.tap(findRadioListTileWithTitle('Dark'));
+      await tester.tap(findRadioListTileWithTitle<ThemeSetting?>('Dark'));
       await tester.pump();
       await tester.pump(Duration(milliseconds: 250)); // wait for transition
       check(Theme.of(element)).brightness.equals(Brightness.dark);
       checkThemeSetting(tester, expectedThemeSetting: ThemeSetting.dark);
 
-      await tester.tap(findRadioListTileWithTitle('System'));
+      await tester.tap(findRadioListTileWithTitle<ThemeSetting?>('System'));
       await tester.pump();
       await tester.pump(Duration(milliseconds: 250)); // wait for transition
       check(Theme.of(element)).brightness.equals(Brightness.light);
--- a/test/widgets/settings_test.dart
+++ b/test/widgets/settings_test.dart
@@ -1,6 +1,7 @@
 import 'package:checks/checks.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_checks/flutter_checks.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:zulip/model/settings.dart';
 import 'package:zulip/widgets/settings.dart';
@@ -39,9 +40,12 @@ void main() {
         ThemeSetting.dark => 'Dark',
       };
       for (final title in ['System', 'Light', 'Dark']) {
-        check(tester.widget<RadioListTile<ThemeSetting?>>(
-          findRadioListTileWithTitle(title)))
-            .checked.equals(title == expectedCheckedTitle);
+        final expectedIsChecked = title == expectedCheckedTitle;
+        check(tester.semantics.find(findRadioListTileWithTitle(title)))
+          .containsSemantics(
+            label: title,
+            isInMutuallyExclusiveGroup: true,
+            hasCheckedState: true, isChecked: expectedIsChecked);
       }
       check(testBinding.globalStore)
         .settings.themeSetting.equals(expectedThemeSetting);
--- a/test/widgets/settings_test.dart
+++ b/test/widgets/settings_test.dart
@@ -4,36 +4,62 @@ import 'package:flutter/material.dart';
 import 'package:flutter_checks/flutter_checks.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:zulip/model/settings.dart';
+import 'package:zulip/widgets/page.dart';
 import 'package:zulip/widgets/settings.dart';
+import 'package:zulip/widgets/store.dart';
 
 import '../flutter_checks.dart';
 import '../model/binding.dart';
 import '../model/store_checks.dart';
 import '../example_data.dart' as eg;
+import '../test_navigation.dart';
+import 'checks.dart';
 import 'test_app.dart';
 
 void main() {
   TestZulipBinding.ensureInitialized();
 
+  late TestNavigatorObserver testNavObserver;
+  late Route<dynamic>? lastPushedRoute;
+  late Route<dynamic>? lastPoppedRoute;
+
   Future<void> prepare(WidgetTester tester) async {
     addTearDown(testBinding.reset);
+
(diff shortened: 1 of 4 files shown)


test/widgets/settings_test.dart
```diff
@@ -127,7 +164,133 @@ void main() {
...
+      VisitFirstUnreadSetting.never => 'Newest message',
+    };
+
+    void checkPage(WidgetTester tester, {
+      required VisitFirstUnreadSetting expectedSetting,
+    }) {
+      for (final setting in VisitFirstUnreadSetting.values) {
+        final thisSettingTitle = settingTitle(setting);
+        checkRadioButtonAppearsChecked<VisitFirstUnreadSetting>(tester,
+          thisSettingTitle, thisSettingTitle == settingTitle(expectedSetting));
```
gnprice: This comparison can be on the values directly instead of the labels, right?
```suggestion
      for (final setting in VisitFirstUnreadSetting.values) {
        final thisSettingTitle = settingTitle(setting);
        checkRadioButtonAppearsChecked<VisitFirstUnreadSetting>(tester,
          thisSettingTitle, setting == expectedSetting);
```

test/widgets/settings_test.dart
```diff
@@ -1,35 +1,74 @@
...
+  late Route<dynamic>? lastPoppedRoute;
+
   Future<void> prepare(WidgetTester tester) async {
     addTearDown(testBinding.reset);
 
+    testNavObserver = TestNavigatorObserver()
+      ..onPushed = ((route, _) => lastPushedRoute = route)
+      ..onPopped = ((route, _) => lastPoppedRoute = route);
+    lastPushedRoute = null;
+    lastPoppedRoute = null;
```
gnprice, line 33: Not for this PR, but: given how much this pattern is recurring now, I'm thinking it'd be good to centralize the code for it.

Perhaps by giving TestNavigatorObserver (or a subclass) some more structure, to track this sort of state itself.

test/widgets/settings_test.dart
```diff
@@ -127,7 +164,133 @@ void main() {
...
+        .visitFirstUnread.equals(VisitFirstUnreadSetting.conversations);
+      checkTileOnSettingsPage(tester,
+        expectedTitle: 'Open message feeds at',
+        expectedSubtitle: settingTitle(VisitFirstUnreadSetting.conversations));
+
+      await tester.tap(find.text('Open message feeds at'));
+      await tester.pump();
+      check(lastPushedRoute).isA<MaterialWidgetRoute>()
+        .page.isA<VisitFirstUnreadSettingPage>();
+      await tester.pump((lastPushedRoute as TransitionRoute).transitionDuration);
```
gnprice, line 195: (again not for this PR)

If the nav observer tracks a log of both pushes and pops in a single list (or I guess even just tracks the last single such action), then this could perhaps even look something like:
```suggestion
      await tester.pump(navObserver.lastTransitionDuration);
```

and exactly the same in the pop case (when the last thing that happened was a pop, the `lastTransitionDuration` getter would look at its `reverseTransitionDuration`).

test/widgets/settings_test.dart
```diff
@@ -127,7 +164,133 @@ void main() {
...
+      checkPage(tester, expectedSetting: VisitFirstUnreadSetting.conversations);
+
+      await tester.tap(findRadioListTileWithTitle<VisitFirstUnreadSetting>(
+        settingTitle(VisitFirstUnreadSetting.always)));
+      await tester.pumpAndSettle(); // TODO why doesn't just `pump` work?
```
gnprice: Hmm, dunno. Does `pump(Duration.zero)` work?

Looking at the implementation of `setThemeSetting`, it'll wait for the (simulated) DB write to complete before it updates the within-Dart store. In real life that's some IO that will take a small but nonzero amount of time, so its completion callback won't happen in a microtask, which is what `pump()` does. So it fits that that also wouldn't suffice in the test.

test/widgets/settings_test.dart
```diff
@@ -127,7 +164,133 @@ void main() {
...
+        settingTitle(VisitFirstUnreadSetting.conversations)));
+      await tester.pumpAndSettle();
+      checkPage(tester, expectedSetting: VisitFirstUnreadSetting.conversations);
+
+      await tester.tap(findRadioListTileWithTitle<VisitFirstUnreadSetting>(
+        settingTitle(VisitFirstUnreadSetting.never)));
+      await tester.pumpAndSettle();
+      checkPage(tester, expectedSetting: VisitFirstUnreadSetting.never);
+
+      await tester.tap(find.byType(BackButton));
```
gnprice: Fun alternative (see its docs) which I happened across recently while browsing other code:
```suggestion
      await tester.tap(find.backButton());
```

test/widgets/settings_test.dart
```diff
@@ -127,7 +164,133 @@ void main() {
...
+      check(GlobalStoreWidget.settingsOf(tester.element(find.byType(SettingsPage))))
+        .visitFirstUnread.equals(VisitFirstUnreadSetting.never);
+
+      checkTileOnSettingsPage(tester,
+        expectedTitle: 'Open message feeds at',
+        expectedSubtitle: settingTitle(VisitFirstUnreadSetting.never));
+    });
+  });
+
+  group('MarkReadOnScrollSetting', () {
```
gnprice, line 226: If we add one or two more of these, we'll likely want to try abstracting them out a bit more — the body of the test case is repetitive enough that it risks getting hard to read without glossing over.

But explicit is always a good way to write the first version, and I think it works fine for two of these.

chrisbobbe:
Hmm, flutter/flutter@a5a3ab3f92f64055855e526634068e8b30bb5a10 breaks the "check what paints" commit. I think probably the fix is to follow a Figma design for radio buttons, instead of using Material widgets. Such a frame exists:

https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5003-65595&t=yxYjWrPU6rpZqcUg-0
image

gnprice:
Hmm. That's probably a good change to do eventually anyway — but it shouldn't be a blocker for this API migration. The Material widgets should have a reasonable way to test a UI that uses them.

I think one good answer would be to switch from inspecting the paint behavior to inspecting the semantics, as in SemanticsNode. That's what people using a screen-reader would encounter, and conveniently it's also by nature more structured. To explore a bit, I tried this edit to the tests:
```diff
   void checkRadioButtonAppearsChecked<T>(WidgetTester tester, String title, bool expectedIsChecked) {
     final element = tester.element(findRadioListTileWithTitle<T>(title));
+    print(element.renderObject!.debugSemantics!.toStringDeep());
```

and then ran the tests. The first couple of trees printed looked like this:
```
SemanticsNode#5
 │ merge boundary ⛔️
 │ Rect.fromLTRB(0.0, 56.0, 800.0, 112.0)
 │ actions: focus, tap
 │ flags: hasEnabledState, isEnabled, isFocusable, hasSelectedState
 │ label: "System"
 │ textDirection: ltr
 │
 └─SemanticsNode#6
     merged up ⬆️
     Rect.fromLTRB(16.0, 8.0, 56.0, 48.0)
     actions: focus, tap
     flags: hasCheckedState, hasEnabledState, isEnabled,
       isInMutuallyExclusiveGroup, isFocusable

SemanticsNode#7
 │ merge boundary ⛔️
 │ Rect.fromLTRB(0.0, 112.0, 800.0, 168.0)
 │ actions: focus, tap
 │ flags: hasEnabledState, isEnabled, isFocusable, hasSelectedState
 │ label: "Light"
 │ textDirection: ltr
 │
 └─SemanticsNode#8
     merged up ⬆️
     Rect.fromLTRB(16.0, 8.0, 56.0, 48.0)
     actions: focus, tap
     flags: hasCheckedState, isChecked, hasEnabledState, isEnabled,
       isInMutuallyExclusiveGroup, isFocusable
```

so it looks like that `isChecked` flag is the thing to inspect.

chrisbobbe:
Interesting; thanks for the `debugSemantics` tip—I definitely tried having the test check semantics before trying the check-what-paints approach, and I couldn't get it to work (the `isChecked` flag wasn't present on the node where I thought it should've been)—but I may have better luck with the help of `debugSemantics` when I come back to this

gnprice:
(Pushed here a branch intended for #1662, oops; then restored the branch that had been here.)

test/widgets/settings_test.dart
chrisbobbe: Indeed; #1668

test/widgets/settings_test.dart
chrisbobbe: Ah a regular `tester.pump` does indeed work when I change to a semantics-based approach instead of checking what paints :)

chrisbobbe:
OK, updated, checking semantics this time. PTAL! 🙂

test/widgets/settings_test.dart
```diff
@@ -1,35 +1,78 @@
...
+        label: subtitle == null
+          ? title
+          : '$title\n$subtitle',
+        isFocusable: true,
+        hasEnabledState: true, isEnabled: true,
+        isInMutuallyExclusiveGroup: true,
+        hasCheckedState: true, isChecked: expectedIsChecked,
+        hasSelectedState: true,
+        hasTapAction: true,
+        hasFocusAction: true));
```
gnprice: Hmm, annoying we have to specify all these flags that seem like not the point of these tests. But I see that's how `matchesSemantics` behaves.

gnprice: … Oho: how about `containsSemantics`? Then I think this whole check statement becomes:
```dart
    check(tester.semantics.find(findRadioListTileWithTitle<T>(title)))
      .containsSemantics(
        label: subtitle == null ? title : '$title\n$subtitle',
        isInMutuallyExclusiveGroup: true,
        hasCheckedState: true, isChecked: expectedIsChecked);
```

gnprice (commented):
Thanks! This approach looks good. One comment below.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! All looks good. I just rebased, and resolved an invisible-to-Git conflict with #1735; will merge after CI passes.

PR #1608, opened 2025-06-18
model: Treat topics case-insensitively in Unreads, topic-visibility, and recent-senders

chrisbobbe:
Fixes: #980

### DIFF
--- a/lib/model/channel.dart
+++ b/lib/model/channel.dart
@@ -1,3 +1,5 @@
+import 'dart:collection';
+
 import 'package:flutter/foundation.dart';
 
 import '../api/model/events.dart';
@@ -369,3 +371,21 @@ class ChannelStoreImpl with ChannelStore {
     }
   }
 }
+
+/// A [Map] with [TopicName] keys and [V] values.
+///
+/// When one of these is created by [makeTopicKeyedMap],
+/// key equality is done case-insensitively; see there.
+///
+/// This type should only be used for maps created by [makeTopicKeyedMap].
+/// It would be nice to enforce that.
+typedef TopicKeyedMap<V> = Map<TopicName, V>;
+
+/// Make a case-insensitive, case-preserving [TopicName]-keyed [LinkedHashMap].
+///
+/// The equality function is [TopicName.isSameAs],
+/// and the hash code is [String.hashCode] of [TopicName.canonicalize].
+TopicKeyedMap<V> makeTopicKeyedMap<V>() => LinkedHashMap<TopicName, V>(
+  equals: (a, b) => a.isSameAs(b),
+  hashCode: (k) => k.canonicalize().hashCode,
+);
--- a/lib/model/channel.dart
+++ b/lib/model/channel.dart
@@ -43,6 +43,8 @@ mixin ChannelStore {
   ///
   /// For policies directly applicable in the UI, see
   /// [isTopicVisibleInStream] and [isTopicVisible].
+  ///
+  /// Topics are treated case-insensitively; see [TopicName.isSameAs].
   UserTopicVisibilityPolicy topicVisibilityPolicy(int streamId, TopicName topic);
 
   /// The raw data structure underlying [topicVisibilityPolicy].
@@ -173,13 +175,13 @@ class ChannelStoreImpl with ChannelStore {
       streams.putIfAbsent(stream.streamId, () => stream);
     }
 
-    final topicVisibility = <int, Map<TopicName, UserTopicVisibilityPolicy>>{};
+    final topicVisibility = <int, TopicKeyedMap<UserTopicVisibilityPolicy>>{};
     for (final item in initialSnapshot.userTopics ?? const <UserTopicItem>[]) {
       if (_warnInvalidVisibilityPolicy(item.visibilityPolicy)) {
         // Not a value we expect. Keep it out of our data structures. // TODO(log)
         continue;
       }
-      final forStream = topicVisibility.putIfAbsent(item.streamId, () => {});
+      final forStream = topicVisibility.putIfAbsent(item.streamId, () => makeTopicKeyedMap());
       forStream[item.topicName] = item.visibilityPolicy;
     }
 
@@ -206,9 +208,9 @@ class ChannelStoreImpl with ChannelStore {
   final Map<int, Subscription> subscriptions;
 
   @override
-  Map<int, Map<TopicName, UserTopicVisibilityPolicy>> get debugTopicVisibility => topicVisibility;
+  Map<int, TopicKeyedMap<UserTopicVisibilityPolicy>> get debugTopicVisibility => topicVisibility;
 
-  final Map<int, Map<TopicName, UserTopicVisibilityPolicy>> topicVisibility;
+  final Map<int, TopicKeyedMap<UserTopicVisibilityPolicy>> topicVisibility;
 
   @override
   UserTopicVisibilityPolicy topicVisibilityPolicy(int streamId, TopicName topic) {
@@ -366,7 +368,7 @@ class ChannelStoreImpl with ChannelStore {
         topicVisibility.remove(event.streamId);
       }
     } else {
-      final forStream = topicVisibility.putIfAbsent(event.streamId, () => {});
+      final forStream = topicVisibility.putIfAbsent(event.streamId, () => makeTopicKeyedMap());
       forStream[event.topicName] = visibilityPolicy;
     }
   }
--- a/lib/model/recent_senders.dart
+++ b/lib/model/recent_senders.dart
@@ -4,6 +4,7 @@ import 'package:flutter/foundation.dart';
 import '../api/model/events.dart';
 import '../api/model/model.dart';
 import 'algorithms.dart';
+import 'channel.dart';
 
 /// Tracks the latest messages sent by each user, in each stream and topic.
 ///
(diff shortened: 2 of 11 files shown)


test/model/unreads_test.dart
```diff
@@ -138,14 +164,16 @@ void main() {
         eg.streamMessage(id: 6, stream: stream2, topic: 'b', flags: [MessageFlag.mentioned]),
         eg.streamMessage(id: 7, stream: stream2, topic: 'c', flags: []),
         eg.streamMessage(id: 8, stream: stream2, topic: 'c', flags: []),
-        eg.dmMessage(id: 9,  from: user1, to: [eg.selfUser], flags: []),
-        eg.dmMessage(id: 10, from: user1, to: [eg.selfUser], flags: []),
```
gnprice, line 142: nit:

    unreads test: Make room for two more message IDs (7, 8) in a test's setup

9, 10, right?

lib/model/channel.dart
```diff
@@ -365,8 +369,18 @@ class ChannelStoreImpl with ChannelStore {
...
   }
 }
+
+/// Make a case-insensitive, case-preserving [TopicName]-keyed [LinkedHashMap].
+///
+/// The equality function is [TopicName.isSameAs],
+/// and the hash code is [String.hashCode] of [TopicName.canonicalize].
+LinkedHashMap<TopicName, V> makeTopicKeyedMap<V>() => LinkedHashMap<TopicName, V>(
+  equals: (a, b) => a.isSameAs(b),
+  hashCode: (k) => k.canonicalize().hashCode,
```
gnprice: I guess the one way this is less efficient than a FoldDict approach is that when looking up a key and finding a match (or a match of the hash code), the FoldDict only needs to call `toLowerCase` on the new key because it already has the lower-case form of the key that's in the map; this version needs to call `toLowerCase` twice.

It's probably fine. If later we find this is hot in profiling, we can optimize.

(And if we do, we might go farther anyway: turn TopicName into a class instead of an extension type, memoizing its canonicalized form. That would make FoldDict unnecessary too.)

lib/model/channel.dart
```diff
@@ -365,8 +369,18 @@ class ChannelStoreImpl with ChannelStore {
...
+  isValidKey: (k) => k is TopicName,
```
gnprice: Is this argument needed? It looks from the docs like this is the default behavior.

lib/model/recent_senders.dart
```diff
@@ -16,7 +19,7 @@ class RecentSenders {
 
   // topicSenders[streamId][topic][senderId] = MessageIdTracker
   @visibleForTesting
-  final Map<int, Map<TopicName, Map<int, MessageIdTracker>>> topicSenders = {};
+  final Map<int, LinkedHashMap<TopicName, Map<int, MessageIdTracker>>> topicSenders = {};
```
gnprice: Hmm, this doesn't really express the meaning we care about. It is a LinkedHashMap, true — but so is what you'd get from just `{}`. If these were some other (reasonable) Map implementation instead, that'd be fine too as long as they used the particular `equals` and `hashCode` etc. that `makeTopicKeyedMap` passes.

I think just `Map` here would be fine (and better than saying LinkedHashMap). It could also be a type alias, like `TopicKeyedMap<Map<int, …>>` — that'd express the desired meaning, even though it wouldn't be enforced.

test/model/recent_senders_checks.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:checks/checks.dart';
+import 'package:collection/collection.dart';
+import 'package:zulip/model/recent_senders.dart';
+
+extension MessageIdTrackerChecks on Subject<MessageIdTracker> {
+  Subject<QueueList<int>> get ids => has((x) => x.ids, 'ids');
```
gnprice: nit: let's put this inside recent_senders_test.dart — for checks-extensions that are unlikely to be used outside a given test file, I've preferred doing it that way

I'm not totally satisfied with that pattern, because sometimes things are used in two or three files, and sometimes they start out used in just one and later are used in another, so that we end up having one test file import another which feels a little unclean.

But the trade-off is that when this file is present, it makes it harder to type a command like `flutter test test/model/recent_senders_test.dart` — the tab-completion has to stop at the `…/recent_senders_` to ask if the next character should be a `t` or a `c` :slightly_smiling_face:

test/model/recent_senders_test.dart
```diff
@@ -125,6 +131,17 @@ void main() {
...
 
+    test('case-insensitive topics', () {
+      checkHandleMessages(
+        [eg.streamMessage(stream: streamA, topic: 'thing', sender: userX)],
+        [eg.streamMessage(stream: streamA, topic: 'ThInG', sender: userX)]);
+      check(model.topicSenders).values.single.entries.single
+        ..key.equals(eg.t('thing'))
+        ..value.entries.single.which((it) => it
+          ..key.equals(userX.userId)
+          ..value.isA<MessageIdTracker>().ids.length.equals(2));
```
gnprice: This felt a bit verbose to me, so I tried simplifying it by using `deepEquals`.

Not sure whether this is an improvement — the need for a Condition in there does dampen things:
```dart
      check(model.topicSenders).values.single.deepEquals(
        {eg.t('thing'): {userX.userId: (Subject<Object?> it) =>
           it.isA<MessageIdTracker>().ids.length.equals(2)}});
```

Maybe better if a bit more expanded in formatting:
```dart
      check(model.topicSenders).values.single.deepEquals(
        {eg.t('thing'):
          {userX.userId: (Subject<Object?> it) =>
             it.isA<MessageIdTracker>().ids.length.equals(2)}});
```

lib/model/unreads.dart
```diff
@@ -41,14 +42,25 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
     required CorePerAccountStore core,
     required ChannelStore channelStore,
   }) {
-    final streams = <int, Map<TopicName, QueueList<int>>>{};
+    final streams = <int, LinkedHashMap<TopicName, QueueList<int>>>{};
```
gnprice: Searching this file for "Map<TopicName", it looks like there's one more spot that probably needs `makeTopicKeyedMap`: in handleUpdateMessageFlagsEvent when marking unread.

lib/model/unreads.dart
```diff
@@ -41,14 +42,25 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
...
+      if (topics[topic] != null) {
+        // Older servers differentiate topics case-sensitively, but shouldn't:
+        //   https://github.com/zulip/zulip/pull/31869
+        // Our topic-keyed map is case-insensitive. When we've seen this
+        // topic before, modulo case, aggregate instead of clobbering.
+        // TODO(server-10) simplify away
+        topics[topic] =
+          setUnion(topics[topic]!, unreadChannelSnapshot.unreadMessageIds);
+      } else {
+        topics[topic] = QueueList.from(unreadChannelSnapshot.unreadMessageIds);
```
gnprice: This will look up `topic` in the map two or three times. Since this is in a potentially hot path (processing the unreads data from the initial snapshot, so often 50k message IDs across perhaps 1000s of conversations), let's avoid that: `topics.update(topic, …)`.

test/model/test_store.dart
```diff
@@ -283,7 +283,7 @@ extension PerAccountStoreTestExtension on PerAccountStore {
     await handleEvent(SubscriptionAddEvent(id: 1, subscriptions: subscriptions));
   }
 
-  Future<void> addUserTopic(ZulipStream stream, String topic, UserTopicVisibilityPolicy visibilityPolicy) async {
+  Future<void> setUserTopic(ZulipStream stream, String topic, UserTopicVisibilityPolicy visibilityPolicy) async {
```
gnprice, line 290: test [nfc]: s/addUserTopic/setUserTopic/ in PerAccountStoreTestExtension
    
    The event's doc says:
    
    https://zulip.com/api/get-events#user_topic
    > Event sent to a user's clients when the user mutes/unmutes a
    > topic, or otherwise modifies their personal per-topic
    > configuration.
    
    The new name is still accurate for "adding" a configuration (i.e.
    setting it to something not-"none"), but now also covers updating a
    configuration (moving between not-"none" configurations) and
    removing one (resetting to "none).

Ah, good thought.

test/model/channel_test.dart
```diff
@@ -158,9 +159,13 @@ void main() {
...
           UserTopicVisibilityPolicy.followed,
         ]) {
-          await store.addUserTopic(stream1, 'topic', policy);
+          await store.setUserTopic(stream1, 'topic', policy);
           check(store.topicVisibilityPolicy(stream1.streamId, eg.t('topic')))
             .equals(policy);
+
+          // Case-insensitive
+          check(store.topicVisibilityPolicy(stream1.streamId, eg.t('ToPiC')))
+            .equals(policy);
```
gnprice, line 168: I think my ideal version of these tests would have most of the new checks in this file be separate test cases, rather than added to existing tests. This also doesn't need a full matrix with the other variables: for example, this check can be done with any one of the non-`none` policies, rather than all three.

These are fine, though. No need to rewrite them — we have higher-priority things to do.

gnprice (commented):
Thanks for taking care of this! Generally this all looks good; mostly small comments below.

(I just saw you'd marked it for maintainer review first, ah well — already wrote this review, so posting it.)

lib/model/recent_senders.dart
gnprice: Apropos of this: the upstream tree just converted places that were saying LinkedHashMap to say plain Map:
https://github.com/flutter/flutter/pull/170713
which apparently will be needed as part of a future lint change there.

test/model/channel_test.dart
chrisbobbe: Yeah, makes sense.

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

PR #1609, opened 2025-06-19
KaTeX (1.5/n): Prep for horizontal and vertical offsets support

rajveermalviya:
Towards #46
Prep for #1452

### DIFF
--- a/lib/widgets/content.dart
+++ b/lib/widgets/content.dart
@@ -948,7 +948,7 @@ class _KatexSpan extends StatelessWidget {
 
     return SizedBox(
       height: styles.heightEm != null
-        ? styles.heightEm! * em
+        ? styles.heightEm! * (fontSize ?? em)
         : null,
       child: widget);
   }


lib/model/katex.dart
```diff
@@ -275,6 +276,21 @@ class _KatexParser {
           fontStyle = KatexSpanFontStyle.normal;
 
         // TODO handle skipped class declarations between .mainrm and
+        //   .mspace .
+
+        case 'mspace':
+          // .mspace { ... }
+          // Do nothing, it has properties that don't need special handling.
+          break;
```
gnprice: It looks like the content of this rule in katex.css is `display: inline-block;`.

On its face that looks like something that has an effect. After some investigation, I'm convinced it doesn't end up having an effect; but that calls for more of a comment here to explain.

I'll add a commit that inserts the result of that investigation.

lib/widgets/content.dart
```diff
@@ -818,7 +818,11 @@ class MathBlock extends StatelessWidget {
           children: [TextSpan(text: node.texSource)])));
     }
 
-    return _Katex(inline: false, nodes: nodes);
+    return Center(
+      child: SingleChildScrollViewWithScrollbar(
+        scrollDirection: Axis.horizontal,
+        child: _Katex(
+          nodes: nodes)));
```
gnprice: content [nfc]: Remove the `inline` property in _Katex widget
    
    And inline the behaviour for `inline: false` in MathBlock widget.

Hmm, this looks like it changes the behavior: the child of the SingleChildScrollViewWithScrollbar was a _KatexNodeList, but is now a _Katex widget. The difference is that the latter adds Directionality and DefaultTextStyle.

The Directionality sounds like a behavior change, when in an RTL locale. Probably a bug fix, in fact; not NFC, and calls for being mentioned in the commit message. Does it indeed change behavior?

The DefaultTextStyle sounds like potentially also a behavior change. Can you determine if it does change behavior, or explain why it doesn't?

lib/model/katex.dart
```diff
@@ -350,14 +373,66 @@ class _KatexParser {
...
-      spans = _parseChildSpans(element);
+      spans = _parseChildSpans(element.nodes);
     }
     if (text == null && spans == null) throw KatexHtmlParseError();
 
-    return KatexNode(
-      styles: styles,
+    return KatexSpanNode(
+      styles: inlineStyles != null
+        ? styles.merge(inlineStyles)
```
gnprice, line 388: It's not obvious to me from this name `merge` whether the recipient `styles` or the argument `inlineStyles` will take precedence when they both have a value for something.

Let's add a bit of dartdoc on the method to make that clear.

How did you choose the name `merge` — is there a similar method you looked at for aligning this with? That might be useful to mention in the dartdoc, as a sort of mnemonic.

lib/model/katex.dart
```diff
@@ -350,14 +373,66 @@ class _KatexParser {
...
+        : styles,
       text: text,
-      nodes: spans);
+      nodes: spans,
+      debugHtmlNode: debugHtmlNode);
+  }
+
+  KatexSpanStyles? _parseSpanInlineStyles(dom.Element element) {
+    if (element.attributes case {'style': final styleStr}) {
```
gnprice, line 396: This seems like it'd be simpler without the if-case:
```suggestion
    final styleStr = element.attributes['style'];
    if (styleStr == null) return null;
```

(Then no indentation added to the rest of the function, and no `return null;` way off at the end that's effectively paired with this as the handler for its failure case.)

… I guess let's plan to do that refactor after merging this and #1452, though, to reduce changes to the later commits there compared with what was in the releases.

lib/model/katex.dart
```diff
@@ -350,14 +373,66 @@ class _KatexParser {
...
+      // `package:csslib` doesn't seem to have a way to parse inline styles:
+      //   https://github.com/dart-lang/tools/issues/1173
+      // So, workaround that by wrapping it in a universal declaration.
```
gnprice: nit: "workaround" is a noun, "work around" is the verb; like "setup"/"set up", etc.

lib/widgets/content.dart
```diff
@@ -941,7 +938,13 @@ class _KatexSpan extends StatelessWidget {
         textAlign: textAlign,
         child: widget);
     }
-    return widget;
+
+    return SizedBox(
+      height: styles.heightEm != null
+        ? styles.heightEm! * em
```
gnprice: Does this do the right thing if this span had both `fontSize` and `heightEm`?

I think it doesn't — the meaning of `em` in CSS is affected by this element's own `font-size` attribute, except when defining `font-size` itself:
https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#relative_length_units
> `em` is relative to the font size of this element, or the font size of the parent element when used for [font-size](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size).

I believe this is enough to fix it:
```suggestion
    return SizedBox(
      height: styles.heightEm != null
        ? styles.heightEm! * (fontSize ?? em)
```

I guess I'll add that as a separate commit, for clarity in comparing the merged version to the last few releases. (It's quite possible it never comes up in practice — that KaTeX never ends up giving us a span that gets both a font-size and a height.)

lib/widgets/content.dart
```diff
@@ -941,7 +938,13 @@ class _KatexSpan extends StatelessWidget {
...
+        : null,
+      child: widget,
+    );
```
gnprice: nit:
```suggestion
      child: widget);
```

gnprice (commented):
OK, reviewed!

I've just pushed a revision that adds two new commits for substantive changes. Comments below for a few nits and also some questions. I think this is close to being merged.

As we've discussed, since previous versions of this PR have been included in releases, I'd like to be mindful of keeping it relatively easy to see how the version we merge compares to those. So for changes to behavior or to a lot of code, let's try to put those into separate commits or separate PRs, even where we typically would squash them in.

lib/widgets/content.dart
rajveermalviya: Hmm, not sure if I am misunderstanding it but before this commit `Directionality` and `DefaultTextStyle` were still in the tree for MathBlock widget, only change is that `Center(SingleChildScrollViewWithScrollbar(` is moving out from `_Katex` to `MathBlock`.

lib/model/katex.dart
rajveermalviya: I guess I had used [`DefaultTextStyle.merge`](https://api.flutter.dev/flutter/widgets/DefaultTextStyle/merge.html) as a reference which does something similar (but a static method instead) for default text style.

rajveermalviya:

lib/widgets/content.dart
gnprice: … Oh aha I was misreading, thanks: the old code said `widget = Center(…)`, not `return Center(…)`.

gnprice: Rereading with that in mind, I believe the effect of this change on the tree is that previously the Center and SingleChildScrollViewWithScrollbar were inside the Directionality and DefaultTextStyle, and now they're outside those. Does that sound right to you?

I think that does have an effect on behavior: the Center doesn't care about either of those, and the SingleChildScrollViewWithScrollbar doesn't care about the DefaultTextStyle, but the latter does care about the Directionality, because the scroll axis is horizontal.

And I think the change there seems undesired. Given that the content of this scroll view is going to be in LTR, I think we want its scrolling to be based on LTR, regardless of what direction is used for the rest of the app.

What was the motivation for this change?

It looks like this doesn't interact with any of the remaining changes in this PR. If that's right, then let's perhaps save it for #1452 — I'm guessing the motivation for it becomes clearer there.

gnprice (commented):
Thanks for the revision! All looks good now, except a question below about the `_Katex.inline` commit.

lib/widgets/content.dart
rajveermalviya: Right this commit is a pre-req for #1452, motivation there was to get the rect for KaTeX content without the `Center` in widget tests.

rajveermalviya: Okay, removed this commit from this PR.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

Issue #196, opened 2023-06-16
Presence data, and report presence


gnprice:
See subsystem doc:
https://zulip.readthedocs.io/en/latest/subsystems/presence.html
and API docs:
https://zulip.com/api/get-events#presence
https://zulip.com/api/get-presence

This is one of the handful of areas where the Zulip API remains underdocumented, so see also the corresponding code in zulip-mobile, and the references linked from https://github.com/zulip/zulip/issues/13519 .

Once we're tracking presence data, it should appear in the UI at:
* #119
* #195
* and possibly other places.

chrisbobbe:
This came up in beta feedback today: [#mobile > Flutter Android 0.0.30 feedback @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20Android.200.2E0.2E30.20feedback/near/2188324)

chrisbobbe:
Someone wrote to support@ in the early launch period (~1 day) requesting this feature.

PR #1614, opened 2025-06-20
api: Prep for supporting presence

chrisbobbe:
This adds some realm/server settings to InitialSnapshot, and presences there, and adds the presence event and updatePresence binding.

Related: #196

### DIFF
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -350,7 +350,7 @@ class PerUserPresence {
   Map<String, dynamic> toJson() => _$PerUserPresenceToJson(this);
 }
 
-/// As in [PerClientPresence.status].
+/// As in [PerClientPresence.status] and [updatePresence].
 @JsonEnum(fieldRename: FieldRename.snake, alwaysCreate: true)
 enum PresenceStatus {
   active,
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -350,6 +350,15 @@ class PerUserPresence {
   Map<String, dynamic> toJson() => _$PerUserPresenceToJson(this);
 }
 
+/// As in [PerClientPresence.status].
+@JsonEnum(fieldRename: FieldRename.snake, alwaysCreate: true)
+enum PresenceStatus {
+  active,
+  idle;
+
+  String toJson() => _$PresenceStatusEnumMap[this]!;
+}
+
 /// An item in `saved_snippets` from the initial snapshot.
 ///
 /// For docs, search for "saved_snippets:"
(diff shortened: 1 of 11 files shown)


lib/api/model/model.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:json_annotation/json_annotation.dart';
 
 import '../../model/algorithms.dart';
+import '../route/users.dart';
```
gnprice: nit: avoid importing routes from models
```suggestion
```

(this was only for a dartdoc reference, so easy to drop)

lib/api/route/users.dart
```diff
@@ -32,3 +33,46 @@ class GetOwnUserResult {
...
+  bool? pingOnly,
+  required PresenceStatus status,
+}) {
+  return connection.post('updatePresence', UpdatePresenceResult.fromJson, 'users/me/presence', {
+    if (lastUpdateId != null) 'last_update_id': lastUpdateId,
+    if (historyLimitDays != null) 'history_limit_days': historyLimitDays,
+    if (newUserInput != null) 'new_user_input': newUserInput,
+    if (pingOnly != null) 'ping_only': pingOnly,
+    'status': RawParameter(status.toJson()),
+    'slim_presence': true,
```
gnprice, line 60: The doc says the client should not pass this on current servers:
> Modern clients should use [last_update_id](https://zulip.com/api/update-presence#parameter-last_update_id), which guarantees that user presence data will be returned in the modern format, and should not pass this parameter as true unless interacting with an older server.

I'm hoping it doesn't really mean that and that the parameter is redundant but harmless. Should check in `#api design`, though. If it does mean it, then I guess we do a zulipFeatureLevel check.

lib/api/route/users.dart
```diff
@@ -32,3 +33,46 @@ class GetOwnUserResult {
...
+  });
+}
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class UpdatePresenceResult {
+  final int? presenceLastUpdateId; // TODO(server-9.0) new in FL 263
+  final double? serverTimestamp; // 1656958539.6287155 in the example response
```
gnprice, line 67: Hmm interesting. Yeah, good to have a test for this.

lib/api/route/users.dart
chrisbobbe: Tim answered this in chat; turns out to be fine, as hoped: [#api design > presence rewrite @ 💬](https://chat.zulip.org/#narrow/channel/378-api-design/topic/presence.20rewrite/near/2201035)

chrisbobbe:
Thanks for the review! Revision pushed, and I plan to send another PR in the next few hours, for reporting and showing presence.

gnprice:
Thanks! Looks good; merging. I'll be on the lookout for that subsequent PR. :-)

Issue #1607, opened 2025-06-18
Show presence (active/inactive) in the app


chrisbobbe:
We should show users' presence in the app, like we do in web:

<img width="254" alt="Image" src="https://github.com/user-attachments/assets/46e99cda-eee6-4366-94d7-b2d4c4e93d16" />

Presence dots do appear in the [Figma](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4992-9046&m=dev), but there's one aspect that should follow web instead: to support color-blind users, we should use different shapes for the dot, not just different colors.

### Related issues

- Prerequisite: #196

gnprice:
@batistein You may be glad to hear (since you filed zulip/zulip-mobile#5925) that the fix for this issue went out in a release a couple of weeks ago, 30.0.259: [#announce > mobile releases @ 💬](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2202921).

Please try out the current release (30.0.261, which has [additional](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2213050) [features](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2217685) as well) and see what you think!

Issue #1618, opened 2025-06-23
presence: Handle `presence` events


chrisbobbe:
In the initial version of presence #196 / #1607, I plan to omit handling the [`presence`](https://zulip.com/api/get-events#presence) event, to get the feature out ASAP. 

The consequence is: when an offline user comes online, it might be up to a minute until the UI updates accordingly. The UI will eventually update, though, which is why I think it's OK to postpone this for a little bit.

We'll need to handle the event in its current form (awkward; I'd like to avoid spending time on it right now) and in a future updated form; see discussion: [#api design > presence rewrite @ 💬](https://chat.zulip.org/#narrow/channel/378-api-design/topic/presence.20rewrite/near/2200812)

PR #1619, opened 2025-06-23
Store, report, and display presence!

chrisbobbe:
TODO tests, but here's a draft for presence!

Screenshots coming soon.

Fixes: #196
Fixes: #1607

### DIFF
--- a/lib/model/presence.dart
+++ b/lib/model/presence.dart
@@ -1,0 +1,182 @@
+import 'dart:async';
+
+import 'package:flutter/scheduler.dart';
+import 'package:flutter/widgets.dart';
+
+import '../api/model/events.dart';
+import '../api/model/model.dart';
+import '../api/route/users.dart';
+import 'store.dart';
+
+/// The model for tracking which users are online, idle, and offline.
+///
+/// Use [presenceStatusForUser]. If that returns null, the user is offline.
+///
+/// This substore is its own [ChangeNotifier],
+/// so callers need to remember to add a listener (and remove it on dispose).
+/// In particular, [PerAccountStoreWidget] doesn't subscribe a widget subtree
+/// to updates.
+class Presence extends PerAccountStoreBase with ChangeNotifier {
+  Presence({
+    required super.core,
+    required this.serverPresencePingInterval,
+    required this.serverPresenceOfflineThresholdSeconds,
+    required this.realmPresenceDisabled,
+    required Map<int, PerUserPresence> initial,
+  }) : _map = initial;
+
+  final Duration serverPresencePingInterval;
+  final int serverPresenceOfflineThresholdSeconds;
+  // TODO(#668): update this realm setting (probably by accessing it from a new
+  //   realm/server-settings substore that gets passed to Presence)
+  final bool realmPresenceDisabled;
+
+  Map<int, PerUserPresence> _map;
+
+  AppLifecycleListener? _appLifecycleListener;
+
+  void _handleLifecycleStateChange(AppLifecycleState newState) {
+    assert(!_disposed); // We remove the listener in [dispose].
+
+    // Since this handler can cause multiple requests within a
+    // serverPresencePingInterval period, we pass `pingOnly: true`, for now, because:
+    // - This makes the request cheap for the server.
+    // - We don't want to record stale presence data when responses arrive out
+    //   of order. This handler would increase the risk of that by potentially
+    //   sending requests more frequently than serverPresencePingInterval.
+    //   (`pingOnly: true` causes presence data to be omitted in the response.)
+    // TODO(#1611) Both of these reasons can be easily addressed by passing
+    //   lastUpdateId. Do that, and stop sending `pingOnly: true`.
+    //   (For the latter point, we'd ignore responses with a stale lastUpdateId.)
+    _maybePingAndRecordResponse(newState, pingOnly: true);
+  }
+
+  bool _hasStarted = false;
+
+  void start() async {
+    if (!debugEnable) return;
+    if (_hasStarted) {
+      throw StateError('Presence.start should only be called once.');
+    }
+    _hasStarted = true;
+
+    _appLifecycleListener = AppLifecycleListener(
+      onStateChange: _handleLifecycleStateChange);
+
+    _poll();
+  }
+
+  Future<void> _maybePingAndRecordResponse(AppLifecycleState? appLifecycleState, {
+    required bool pingOnly,
+  }) async {
+    if (realmPresenceDisabled) return;
+
+    final UpdatePresenceResult result;
+    switch (appLifecycleState) {
+      case null:
+      case AppLifecycleState.hidden:
+      case AppLifecycleState.paused:
+        // No presence update.
+        return;
+      case AppLifecycleState.detached:
+        // > The application is still hosted by a Flutter engine but is
+        // > detached from any host views.
+        // TODO see if this actually works as a way to send an "idle" update
+        //   when the user closes the app completely.
+        result = await updatePresence(connection,
+          pingOnly: pingOnly,
+          status: PresenceStatus.idle,
+          newUserInput: false);
+      case AppLifecycleState.resumed:
+        // > […] the default running mode for a running application that has
+        // > input focus and is visible.
+        result = await updatePresence(connection,
+          pingOnly: pingOnly,
+          status: PresenceStatus.active,
+          newUserInput: true);
+      case AppLifecycleState.inactive:
+        // > At least one view of the application is visible, but none have
+        // > input focus. The application is otherwise running normally.
+        // For example, we expect this state when the user is selecting a file
+        // to upload.
+        result = await updatePresence(connection,
+          pingOnly: pingOnly,
+          status: PresenceStatus.active,
+          newUserInput: false);
+    }
+    if (!pingOnly) {
+      _map = result.presences!;
+      notifyListeners();
+    }
+  }
+
+  void _poll() async {
+    assert(!_disposed);
+    while (true) {
+      // We put the wait upfront because we already have data when [start] is
+      // called; it comes from /register.
+      await Future<void>.delayed(serverPresencePingInterval);
+      if (_disposed) return;
+
+      await _maybePingAndRecordResponse(
+        SchedulerBinding.instance.lifecycleState, pingOnly: false);
+      if (_disposed) return;
+    }
+  }
+
+  bool _disposed = false;
+
+  @override
+  void dispose() {
+    _appLifecycleListener?.dispose();
+    _disposed = true;
+    super.dispose();
+  }
+
+  /// The [PresenceStatus] for [userId], or null if the user is offline.
+  PresenceStatus? presenceStatusForUser(int userId, {required DateTime utcNow}) {
+    final now = utcNow.millisecondsSinceEpoch ~/ 1000;
+    final perUserPresence = _map[userId];
+    if (perUserPresence == null) return null;
+    final PerUserPresence(:activeTimestamp, :idleTimestamp) = perUserPresence;
+
+    if (now - activeTimestamp <= serverPresenceOfflineThresholdSeconds) {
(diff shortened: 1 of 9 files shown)


chrisbobbe:
On the profile page:

TODO (followup): The "offline" status should be represented with an open circle, like on web, not a solid circle, like here. That's in order to show meaning by shape, not just color, for color-blind users.

| Light | Dark |
| --- | --- |
| image | image |
| image | image |
| image | image |

chrisbobbe:
Direct-messages screen. See that "Chris Bobbe (Test Account)" is online, "Alya Abbott" is idle, and everyone else is offline.

| Light | Dark |
| --- | --- |
| image | image |

chrisbobbe:
Message list:

| Light | Dark |
| --- | --- |
| image | image |
| image | image |

lib/model/presence.dart
```diff
@@ -0,0 +1,158 @@
...
+          newUserInput: true);
+      case AppLifecycleState.inactive:
+        // > At least one view of the application is visible, but none have
+        // > input focus. The application is otherwise running normally.
+        // For example, we expect this state when the user is selecting a file
+        // to upload.
+        result = await updatePresence(connection,
+          pingOnly: pingOnly,
+          status: PresenceStatus.active,
+          newUserInput: false);
```
gnprice, line 105: - The newUserInput param is now usually true instead of always
      false. This seems more correct to me, and the change seems
      low-stakes (the doc says it's used to implement usage statistics);
      see the doc:
        https://zulip.com/api/update-presence#parameter-new_user_input

This sounds reasonable to me, but probably good to check with Tim in case he has thoughts based on how that's actually used on the server — perhaps @-mention him in the #mobile-team thread.

lib/model/presence.dart
```diff
@@ -0,0 +1,158 @@
...
+
+    _poll();
+  }
+
+  Future<void> _maybePingAndRecordResponse(AppLifecycleState? appLifecycleState, {
+    required bool pingOnly,
+  }) async {
+    if (realmPresenceDisabled) return;
+
+    late UpdatePresenceResult result;
```
gnprice:
```suggestion
    final UpdatePresenceResult result;
```

right? (I.e., the analyzer should be able to verify this is always initialized before it gets used.)

lib/model/presence.dart
chrisbobbe: [#mobile-team > presence @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/presence/near/2202723)

Quoting Tim:

> Probably true is a better value if one wants to do something quick. Most accurately, you should be passing true if any only if the user has interacted with the app at all since the last presence request.

> I think on a mobile device, always true is a fairly good simulation of that. But I'd definitely an M6 issue for refining it to implement that detail on the spec fully and then move on.

lib/model/store.dart
```diff
@@ -498,8 +502,12 @@ class PerAccountStore extends PerAccountStoreBase with ChangeNotifier, EmojiStor
       ),
       users: UserStoreImpl(core: core, initialSnapshot: initialSnapshot),
       typingStatus: TypingStatus(core: core,
-        typingStartedExpiryPeriod: Duration(milliseconds: initialSnapshot.serverTypingStartedExpiryPeriodMilliseconds),
-      ),
+        typingStartedExpiryPeriod: Duration(milliseconds: initialSnapshot.serverTypingStartedExpiryPeriodMilliseconds)),
+      presence: Presence(core: core,
+        serverPresencePingIntervalSeconds: Duration(seconds: initialSnapshot.serverPresencePingIntervalSeconds),
```
gnprice: nit: once it's a Duration rather than a plain number, the units no longer belong in the name:
```suggestion
        serverPresencePingInterval: Duration(seconds: initialSnapshot.serverPresencePingIntervalSeconds),
```

(cf `typingStartedExpiryPeriod` / `serverTypingStartedExpiryPeriodMilliseconds` above)

lib/model/presence.dart
```diff
@@ -0,0 +1,158 @@
...
+      notifyListeners();
+    }
+  }
+
+  void _poll() async {
+    assert(!_disposed);
+    while (true) {
+      // We put the wait upfront because we already have data when [start] is
+      // called; it comes from /register.
+      await Future<void>.delayed(serverPresencePingIntervalSeconds);
```
gnprice: This is actually the spot that prompted the comment just above about this field's name — this line looks wrong, as it's taking what looks like a number of seconds but then using it in a way that doesn't appear to be about seconds.

lib/widgets/content.dart
```diff
@@ -1662,17 +1664,23 @@ class Avatar extends StatelessWidget {
     required this.userId,
     required this.size,
     required this.borderRadius,
+    this.backgroundColor,
+    this.omitPresenceStatus = false,
```
gnprice: nit: the negative feels more confusing to read; and I think "status" is redundant here
```suggestion
    this.showPresence = true,
```

lib/widgets/content.dart
```diff
@@ -1662,17 +1664,23 @@ class Avatar extends StatelessWidget {
...
   });
```
gnprice: The backgroundColor is only meaningful if presence will be shown, right? Probably worth an assert here to avoid passing it when it wouldn't do anything.

(and/or on AvatarShape)

lib/widgets/message_list.dart
```diff
@@ -1704,7 +1704,12 @@ class _SenderRow extends StatelessWidget {
...
               child: Row(
                 children: [
-                  Avatar(size: 32, borderRadius: 3,
+                  Avatar(
+                    size: 32,
+                    borderRadius: 3,
+                    // TODO(#647) use different color for highlighted messages
+                    // TODO(#681) use different color for DM messages
+                    backgroundColor: designVariables.bgMessageRegular,
                     userId: message.senderId),
```
gnprice: Hmm, do we want to show presence on all the sender avatars in the message list? We don't on web.

OTOH I guess on web you have the right sidebar.

(Did we show these in the legacy app?)

lib/widgets/content.dart
```diff
@@ -1722,26 +1730,163 @@ class AvatarImage extends StatelessWidget {
...
+  final Color? backgroundColor;
+  final int? userIdForPresence;
   final Widget child;
 
   @override
   Widget build(BuildContext context) {
+    Widget result = child;
+
+    if (userIdForPresence != null) {
+      final presenceCircleSize = size / 4; // TODO(design) is this right?
```
gnprice, line 1769: Yeah, good question. Probably good to raise in #mobile-design, as a follow-up.

Given we don't show them on the one place where an avatar is a drastically different size from elsewhere (namely the top of the profile page), this way seems fine to ship.

gnprice: I think one assumption this layout makes is that borderRadius is no bigger than presenceCircleSize. If bigger, it looks like it would cut off part of the circle.

Could make that assumption explicit with an assert… or could also move the ClipRRect to apply only to the child, not this whole Stack. Then the presence circle in that case would stick out a little past the avatar's rounded corner, which I think would be fine.

lib/widgets/content.dart
```diff
@@ -1722,26 +1730,163 @@ class AvatarImage extends StatelessWidget {
...
         borderRadius: BorderRadius.all(Radius.circular(borderRadius)),
         clipBehavior: Clip.antiAlias,
-        child: child));
+        child: result));
+  }
+}
+
+/// The green or orange-gradient circle representing [PresenceStatus].
+///
+/// [backgroundColor] is required and must not be [Colors.transparent].
```
gnprice: This bit appears to not match the code — that parameter is optional.

lib/widgets/content.dart
```diff
@@ -1722,26 +1730,163 @@ class AvatarImage extends StatelessWidget {
...
+    final size = textScaler.scale(fontSize) / 2;
+    return WidgetSpan(
+      alignment: PlaceholderAlignment.middle,
+      child: Padding(
+        padding: const EdgeInsetsDirectional.only(end: 4),
+        child: PresenceCircle(
+          userId: userId,
+          size: size,
+          backgroundColor: backgroundColor,
+          explicitOffline: true)));
```
gnprice, line 1828: This flag seems orthogonal to being used as a WidgetSpan. Probably just make it a parameter of this method, and have the one caller specify it explicitly.

gnprice: … hmm, I guess the padding would look odd if there's nothing there.

Maybe just mention in this method's doc, then.

lib/widgets/content.dart
```diff
@@ -1722,26 +1730,163 @@ class AvatarImage extends StatelessWidget {
...
+    final status = model!.presenceStatusForUser(
+      widget.userId, utcNow: ZulipBinding.instance.utcNow());
+    final designVariables = DesignVariables.of(context);
+    final effectiveBackgroundColor = widget.backgroundColor ?? designVariables.mainBackground;
+    assert(effectiveBackgroundColor != Colors.transparent);
+
+    Color? color;
+    LinearGradient? gradient;
+
+    switch (status) {
```
gnprice: nit: join these to the switch that sets them
```suggestion
    Color? color;
    LinearGradient? gradient;
    switch (status) {
```

lib/widgets/content.dart
```diff
@@ -1662,17 +1664,23 @@ class Avatar extends StatelessWidget {
...
     required this.borderRadius,
+    this.backgroundColor,
+    this.omitPresenceStatus = false,
   });
 
   final int userId;
   final double size;
   final double borderRadius;
+  final Color? backgroundColor;
+  final bool omitPresenceStatus;
```
gnprice: Looking at other call sites of the Avatar constructor, and thinking about whether they should set this flag to omit presence.
* autocomplete → keep
* "My profile" in main menu → probably omit? seems like there's no useful information there; and pending #1618, could be confusing by suggesting you're offline
* lightbox → keep
* msglist senders → see other comment
* new DMs, two places → keep
* top of profile → skip, as this version already does
* users in custom profile fields → keep

gnprice (commented):
Thanks for the swift implementation! Generally all looks great. Comments below.

lib/widgets/message_list.dart
chrisbobbe: IIRC we didn't show them in the message list in zulip-mobile.

gnprice:
CI shows some test failures, in test/model/store_test.dart .

chrisbobbe:
Thanks for the review! Revision pushed. This one omits the presence circle on sender rows in the message list, to match zulip-mobile and web.

lib/model/presence.dart
```diff
@@ -0,0 +1,182 @@
...
+  static set debugEnable(bool value) {
+    assert(() {
+      _debugEnable = value;
+      return true;
+    }());
+  }
+
+  @visibleForTesting
+  static void debugReset() {
+    _debugEnable = true;
```
gnprice: nit:
```suggestion
  static void debugReset() {
    debugEnable = true;
```

That way it goes through the setter above, so it's clear to the compiler that this field never gets written outside debug mode.

gnprice (commented):
Thanks! All looks good except a nit below.

> This one omits the presence circle on sender rows in the message list, to match zulip-mobile and web.

To be explicit for the thread: it also omits it on the "My profile" item in the main menu.

gnprice:
Then I'll want to go ahead and merge, to simplify release management. So I guess let's file one more follow-up issue calling for tests.

chrisbobbe:
Thanks! Done (#1620), and fixed that nit; PTAL.

gnprice:
Thanks! Looks good; merging.

Tweaked one bit in the commit messages:

    -    We plan to write tests for this as a followup: #1620
    +    We plan to write tests for this as a followup: #1620.
     
         Fixes: #1607

because it was looking almost like an analogue of the "Fixes:" lines, and then it was bugging me because the name had spaces in it :slightly_smiling_face:

Issue #457, opened 2023-12-21
Show message timestamps on swipe


gnprice:
With #446 adopting the mobile design from Figma for the layout of messages (as opposed to this client's current design there which was copied from web), we'll lose the timestamps at the right, except on the first message in a row from a given sender, in order to reclaim that horizontal space for the message content.

As a result, as @chrisbobbe explains at https://github.com/zulip/zulip-flutter/pull/446#pullrequestreview-1787600537 , we'll stop showing some useful information up front:
* where the boundaries between a sender's messages are;
* what time they sent each of their messages.

We've had a similar dilemma in zulip-mobile, with the same need to make good use of the available horizontal space. With our current design there, the boundaries are visible if you look closely, and there is a way to get each timestamp. But I would not call that design a great success, as I suspect most users have never discovered it even if they've felt the desire for that information.

However:

> The Figma has a design to support checking the date/time for any message; it looks like it wants us to let the user slide a message to the left, to reveal the timestamp: https://www.figma.com/file/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?type=design&node-id=147-8975&mode=design&t=P97AuMzjt47Mudyg-0
[...]

Issue #10, opened 2023-02-11
feature: Text-selection of message content


chrisbobbe:
In the React Native app, I miss the web app's feature of letting you select a word or phrase from a message in the message list, so I can:

- Copy it to the clipboard
- Look it up in the dictionary
- Translate it into my language
- [etc.; this feature is powered by the browser, so e.g. on Chrome I see 'Search Google for "[text]"']

If we wrap a message's contents in a [`SelectionArea`](https://api.flutter.dev/flutter/material/SelectionArea-class.html) widget, the text becomes selectable:

<img width=400 src="https://user-images.githubusercontent.com/22248748/218236433-8d5ed305-b83f-4fa0-b86e-813d9f94428a.gif" />

And I can copy it! (Though I don't get any other options like a dictionary lookup.)

In my screen recording, the interaction is started by a long-press, as is usual on iOS. I think we'll want to reserve the long-press gesture to open an action sheet, like in the React Native app. I don't think that has to block offering text selection, just that we should find a different way to start the interaction. How about a "select text" item in the message action sheet, which starts a selection session by selecting the whole message's contents? Then the user can move the selection handles where they want them.

At that point I think it would work OK for selecting a short word or phrase, although I should mention that a bunch of formatting is lost when I copy the text. Here's what went to the clipboard in that screen recording:

> timabbott pushed 2 commits to branch main. Commits by alya (1) and drrosa (1).•   help: Document new user setting for configuring email visibility. (9ea3572017e)•   help: Update URL references to new "Configure email visibility" page. (2b9b0acaa4f)

So as a polish issue, we should see if we can change what goes to the clipboard, with information from the parsed Markdown, probably by consulting the widgets whose rendered text appears in the selection. We can re-Markdown links, put linebreaks where they clearly belong, etc.

-----

The React Native app lets you copy the whole raw Markdown content of a message. That's a feature we should offer, but it's not this feature exactly: it's not great when you want to copy just a word or short phrase from a long message, without bringing loads of other stuff into the clipboard that you don't want.

gnprice:
I think this is a useful feature and would be a good thing to look at at some point after the prototype phase.

chrisbobbe:
This would also open the door to quote-and-replying just part of a message. That came up in a beta-feedback email to support@:

> thank you for providing the new Mobile Application. I am a bit disappointed that one of the major missing feature, quoting part of a previous message had not been implemented yet.
> 
[...]

Issue #1142, opened 2024-12-12
msglist: Message bounds hard to see in a run of messages from the same user


chrisbobbe:
When we show a run of consecutive messages from the same user, it can be hard to see where one message ends and the next begins. Compare zulip-flutter and web:

| web | zulip-flutter |
| --- | --- |
| <img width="881" alt="image" src="https://github.com/user-attachments/assets/e0943fa3-c135-4402-a9d9-f6f197287bcd" /> | <img width="675" alt="image" src="https://github.com/user-attachments/assets/20b8e1f3-d151-4b86-ae92-a04857f173ec" /> |
| <img width="885" alt="image" src="https://github.com/user-attachments/assets/5025ae97-7c99-4b6d-9655-42de67137b24" /> | <img width="675" alt="image" src="https://github.com/user-attachments/assets/1a8662b6-62a6-4a14-9309-335e640a3b71" /> |

I noticed this at https://github.com/zulip/zulip-flutter/pull/446#pullrequestreview-1787600537 and it came up again in beta feedback: https://chat.zulip.org/#narrow/channel/48-mobile/topic/mobile.20beta.20feedback/near/2000252

One clear improvement, I think, is to tint the background of a message when you long-press it, which you might do to open the message action sheet. We even have a [Figma frame](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=4010-7127&node-type=text&m=dev) we can follow:

<img width="377" alt="image" src="https://github.com/user-attachments/assets/519348f6-0095-4759-8235-50b190f394a4" />

> Once message is pressed there should be a tint of the background, since the background color could be anything. So pressedTint is a transparent foreground color.
> 
> Pressed state here is kind of selected state, because we want to keep highlighting the message as a contextual information for a dialog of editing the message.

That design has a different motivation (#126 edit message) but would help this message-bounds issue too, without having to add or rearrange anything in the layout. (Space is tight on mobile.)

Gaurav-Kushwaha-1225:
Hi @chrisbobbe,
I would like to work on this issue and implement the proposed solution. 
Could you please assign it to me?
Thank you! 😊

chrisbobbe:
Please take a look at the Zulip project's guide to getting involved with the code:

https://zulip.readthedocs.io/en/latest/contributing/contributing.html#your-first-codebase-contribution
[...]

PR #1624, opened 2025-06-24
Show message content, sender, and timestamp in message action sheet

### LINKED ISSUE #217: Show message text in bottom sheet

chrisbobbe:
Fixes #217.

We've had some complaints about not being able to see the timestamps of messages in a run of messages from the same sender, e.g.:

[#mobile > Sent time not visible @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Sent.20time.20not.20visible/near/2201862)
> There is a "message sent" time shown in an upper right hand corner of the message. But when the messages are sent by bot, the time is only shown for the first message of the day. The old android app at least shown the time when tapped on the message, but I haven't found a way how to display the time here.

We have #457 for how to offer that in a really quick interaction. But this PR lets you get that information, and it aligns with a different part of the Figma, namely #217. As discussed: [#mobile > Sent time not visible @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Sent.20time.20not.20visible/near/2202542)

(edit: all but one of these screenshots are outdated in the timestamp format; the one with the full date and time is current)

| Light | Dark |
| --- | --- |
| image | image |
| image | image |
| image | image |

This revision leaves the message completely interactable; you can open images in the lightbox, open and close spoilers, etc.; that seemed most straightforward to implement. It does not implement text selection of the content; that's requested in the Figma, and #10 is for that, and I've left a TODO for it.

### DIFF
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -13,12 +13,14 @@ import '../api/route/channels.dart';
 import '../api/route/messages.dart';
 import '../generated/l10n/zulip_localizations.dart';
 import '../model/binding.dart';
+import '../model/content.dart';
 import '../model/emoji.dart';
 import '../model/internal_link.dart';
 import '../model/narrow.dart';
 import 'actions.dart';
 import 'color.dart';
 import 'compose_box.dart';
+import 'content.dart';
 import 'dialog.dart';
 import 'emoji.dart';
 import 'emoji_reaction.dart';
@@ -33,6 +35,7 @@ import 'topic_list.dart';
 
 void _showActionSheet(
   BuildContext pageContext, {
+  Widget? header,
   required List<Widget> optionButtons,
 }) {
   // Could omit this if we need _showActionSheet outside a per-account context.
@@ -57,15 +60,31 @@ void _showActionSheet(
             child: Column(
               mainAxisSize: MainAxisSize.min,
               children: [
-                SizedBox(height: 8),
+                if (header != null)
+                  Flexible(
+                    // TODO(upstream) Enforce a flex ratio (e.g. 1:3)
+                    //   only when the header height plus the buttons' height
+                    //   exceeds available space. Otherwise let one or the other
+                    //   grow to fill available space even if it breaks the ratio.
+                    //   Needs support for separate properties like `flex-grow`
+                    //   and `flex-shrink`.
+                    flex: 1,
+                    child: InsetShadowBox(
+                      top: 8, bottom: 8,
+                      color: designVariables.bgContextMenu,
+                      child: SingleChildScrollView(
+                        padding: EdgeInsets.symmetric(vertical: 8),
+                        child: header)))
+                else
+                  SizedBox(height: 8),
                 Flexible(
+                  flex: 3,
                   child: Padding(
                     padding: const EdgeInsets.fromLTRB(16, 0, 16, 0),
                     child: Column(
                       crossAxisAlignment: CrossAxisAlignment.stretch,
                       mainAxisSize: MainAxisSize.min,
                       children: [
-                        // TODO(#217): show message text
                         Flexible(child: InsetShadowBox(
                           top: 8, bottom: 8,
                           color: designVariables.bgContextMenu,
@@ -621,7 +640,42 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       EditButton(message: message, pageContext: pageContext),
   ];
 
-  _showActionSheet(pageContext, optionButtons: optionButtons);
+  _showActionSheet(pageContext,
+    optionButtons: optionButtons,
+    header: _MessageActionSheetHeader(message: message));
+}
+
+class _MessageActionSheetHeader extends StatelessWidget {
+  const _MessageActionSheetHeader({required this.message});
+
+  final Message message;
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+
+    // TODO this seems to lose the hero animation when opening an image;
+    //   investigate.
+    // TODO should we close the sheet before opening a narrow link?
+    //   On popping the pushed narrow route, the sheet is still open.
+
+    return Container(
+      // TODO(#647) use different color for highlighted messages
+      // TODO(#681) use different color for DM messages
+      color: designVariables.bgMessageRegular,
+      padding: EdgeInsets.symmetric(vertical: 4),
+      child: Column(
+        spacing: 4,
+        children: [
+          SenderRow(message: message,
+            timestampStyle: MessageTimestampStyle.full),
+          Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 16),
+            // TODO(#10) offer text selection; the Figma asks for it here:
+            //   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-30210&m=dev
+            child: MessageContent(message: message, content: parseMessageContent(message))),
+        ]));
+  }
 }
 
 abstract class MessageActionSheetMenuItemButton extends ActionSheetMenuItemButton {
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1893,7 +1893,8 @@ class SenderRow extends StatelessWidget {
     final revealedMutedMessagesState =
       MessageListPage.maybeRevealedMutedMessagesOf(context);
     // The "unrevealed" state only exists in the message list,
-    // and we're about to start showing a sender row outside the message list.
+    // and we show a sender row in at least one place outside the message list
+    // (the message action sheet).
     if (revealedMutedMessagesState == null) return false;
     return !revealedMutedMessagesState.isMutedMessageRevealed(message.id);
   }
@@ -1969,9 +1970,23 @@ class SenderRow extends StatelessWidget {
 enum MessageTimestampStyle {
   none,
   timeOnly,
+
+  /// The longest format, with full date and time as numbers, not "Today"/etc.
+  ///
+  /// For UI contexts focused just on the one message,
+  /// or as a tooltip on a shorter-formatted timestamp.
+  ///
+  /// The detail won't always be needed, but this format makes mental timezone
+  /// conversions easier, which is helpful when the user is thinking about
+  /// business hours on a different continent,
+  /// or traveling and they know their device timezone setting is wrong, etc.
+  // TODO(design) show "Today"/etc. after all? Discussion:
+  //   https://github.com/zulip/zulip-flutter/pull/1624#issuecomment-3050296488
+  full,
   ;
 
   static final _timeOnlyFormat = DateFormat('h:mm aa', 'en_US');
+  static final _fullFormat = DateFormat.yMMMd().add_jm();
 
   /// Format a [Message.timestamp] for this mode.
   // TODO(i18n): locale-specific formatting (see #45 for a plan with ffi)
@@ -1982,6 +1997,7 @@ enum MessageTimestampStyle {
     switch (this) {
       case none:     return null;
       case timeOnly: return _timeOnlyFormat.format(asDateTime);
+      case full: return _fullFormat.format(asDateTime);
     }
   }
 }
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -38,6 +38,7 @@ import '../api/fake_api.dart';
 import '../example_data.dart' as eg;
 import '../flutter_checks.dart';
 import '../model/binding.dart';
+import '../model/content_test.dart';
 import '../model/test_store.dart';
 import '../stdlib_checks.dart';
 import '../test_clipboard.dart';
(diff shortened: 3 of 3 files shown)


chrisbobbe:
This also mitigates #1142 because it gives a way to see one message by itself.

lib/widgets/action_sheet.dart
```diff
@@ -57,15 +61,27 @@ void _showActionSheet(
             child: Column(
               mainAxisSize: MainAxisSize.min,
               children: [
-                SizedBox(height: 8),
+                if (header != null)
+                  Flexible(
+                    flex: 1,
```
rajveermalviya, line 71: The figma design [says](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-27050&t=keKKioYTorJvPlnT-0):
> Menu should grow until it fits the whole screen

But currently it doesn't, so maybe let's have a TODO for it.

lib/widgets/action_sheet.dart
```diff
@@ -616,7 +632,36 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+      child: Column(
+        mainAxisSize: MainAxisSize.max,
+        spacing: 4,
+        children: [
+          SenderRow(message: message, showTimestamp: true),
+          Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 16),
+            // TODO(#10) offer text selection; the Figma asks for it here:
+            //   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-30210&m=dev
+            child: MessageContent(message: message, content: parseMessageContent(message))),
```
rajveermalviya, line 676: Couple of things I noticed with interactions in the message content:
- There was no hero animation when opening an image.
- Clicking on a narrow link, would open the narrow, but wouldn't close the action sheet (which we do for some for some cases).

Neither seem blocker for this feature though.

lib/widgets/action_sheet.dart
chrisbobbe: Yeah; there's a note on #1077 that's relevant:

> (TODO upstream issue)
> Richer Flex (aka Row and Column), with more CSS flexbox features like flex-basis vs flex-grow vs flex-shrink. (This is one of the few things we miss from React Native: its Yoga layout engine has those features.)
> We probably need something like this for: [etc.]

chrisbobbe:
Thanks! I've added TODOs for all those points.

test/widgets/action_sheet_test.dart
```diff
@@ -103,7 +104,11 @@ Future<void> setupToMessageActionSheet(WidgetTester tester, {
   // because its render box effectively has HitTestBehavior.deferToChild, and
   // the long-press might land where no child hit-tests as true,
   // like if it's in padding around a Paragraph.
-  await tester.longPress(find.byType(MessageContent), warnIfMissed: false);
+  await tester.longPress(
+    find.descendant(
+      of: find.byType(MessageList),
+      matching: find.byType(MessageContent)),
+    warnIfMissed: false);
```
gnprice: action_sheet test: Narrow a Finder (long-press for message action sheet)
    
    We're about to put a MessageContent in the action sheet itself,
    for #217.

This change seems fine but I'm a bit surprised it's needed. There's no action sheet in the tree at this point, right? This is the step that will cause one to appear.

lib/model/content.dart
```diff
@@ -1887,6 +1887,12 @@ class _ZulipContentParser {
   }
 }
 
+ZulipMessageContent parseMessageContent(Message message) {
```
gnprice: nit: put at end of file, matching the specific-to-general ordering of the other parsing code here

lib/widgets/action_sheet.dart
```diff
@@ -32,39 +34,71 @@ import 'theme.dart';
...
 
 void _showActionSheet(
-  BuildContext context, {
+  BuildContext pageContext, {
   required List<Widget> optionButtons,
+  Widget? header,
+  Color? headerBackgroundColor,
 }) {
+  // Could omit this if we need _showActionSheet outside a per-account context.
+  final store = PerAccountStoreWidget.of(pageContext);
```
gnprice: Let's use accountIdOf instead, to make clear that this doesn't use anything else from the store.

(If it kept this store around and then used much of anything else from it in the build callback below, that'd be a bug because the store might change; in general we don't want to be keeping references to the store around beyond the lifetime of a build method or a gesture callback, and instead should look the store up anew from a context.)

lib/widgets/action_sheet.dart
```diff
@@ -32,39 +34,71 @@ import 'theme.dart';
...
+                    //   and `flex-shrink`.
+                    flex: 1,
+                    child: InsetShadowBox(
+                      top: 8, bottom: 8,
+                      color: designVariables.bgContextMenu,
+                      child: SingleChildScrollView(
+                        padding: EdgeInsets.symmetric(vertical: 8),
+                        child: ColoredBox(
+                          color: headerBackgroundColor ?? Colors.transparent,
+                          child: header))))
```
gnprice: This color is currently always transparent, right? So can simplify this out, if that's the value we want.

lib/widgets/action_sheet.dart
```diff
@@ -604,7 +638,42 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       EditButton(message: message, pageContext: pageContext),
   ];
 
-  _showActionSheet(pageContext, optionButtons: optionButtons);
+  _showActionSheet(pageContext,
+    optionButtons: optionButtons,
+    header: _MessageActionSheetHeader(message: message));
```
gnprice, line 645: nit: put header before option buttons; it appears before them physically, and logically they feel like the payload and this more like some decoration, so those two ways of ordering agree with each other

lib/widgets/action_sheet.dart
```diff
@@ -604,7 +638,42 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+    // TODO should we close the sheet before opening a narrow link?
+    //   On popping the pushed narrow route, the sheet is still open.
+
+    return Container(
+      // TODO(#647) use different color for highlighted messages
+      // TODO(#681) use different color for DM messages
+      color: designVariables.bgMessageRegular,
+      padding: EdgeInsets.symmetric(vertical: 4),
+      child: Column(
+        mainAxisSize: MainAxisSize.max,
```
gnprice: Hmm interesting. What's the effect of this parameter?

lib/widgets/action_sheet.dart
```diff
@@ -604,7 +638,42 @@ void showMessageActionSheet({required BuildContext context, required Message mes
...
+        spacing: 4,
+        children: [
+          SenderRow(message: message, showTimestamp: true),
+          Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 16),
+            // TODO(#10) offer text selection; the Figma asks for it here:
+            //   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3483-30210&m=dev
+            child: MessageContent(message: message, content: parseMessageContent(message))),
```
gnprice, line 676: At first I'd thought this wouldn't be possible without more work
    toward issue #488:
    
    - #488 "content: Support Zulip content outside messages (even
      outside per-account contexts)".
    
    But it turns out that our content widgets don't assume a MessageList
    ancestor; it's sufficient that we provide a per-account context, via
    PerAccountStoreWidget, and a per-message context, via
    InheritedMessage (created by MessageContent).

How does this interact with muted users, #1560? That's one place where some related code, at least, relies on state that lives at MessageList.

(This is probably an instance where we're paying some cost for the technical debt of not yet having merged those muted-users PRs.)

test/widgets/action_sheet_test.dart
```diff
@@ -850,6 +855,44 @@ void main() {
   });
 
   group('message action sheet', () {
+    group('header', () {
+      testWidgets('message sender and content shown', (tester) async {
```
gnprice, line 884: Is it easy to also add a check that the timestamp is shown? That's the motivating use case that prompted us to actually go and implement this feature now, so it'd be good to test it too :slightly_smiling_face: 

(But I know controlling times and time zones is an annoying area in tests still, so if it's not easy then we can skip it in favor of shipping, with I guess a TODO comment here in the tests.)

gnprice (commented):
Thanks for building this! Generally it all looks good; comments below.

lib/widgets/action_sheet.dart
chrisbobbe: Oops, I think this isn't wanted; will remove. (I'm not noticing it doing anything, and can't think of why we'd want it.)

lib/widgets/action_sheet.dart
chrisbobbe: Indeed, this calls for some work to allow `SenderRow` to appear where there isn't an ancestor `MessageListPage`. Will fix.

test/widgets/action_sheet_test.dart
chrisbobbe: I started writing a test for this just now, then realized we don't show the full date :) is that okay, do you think? You could be looking at a message from a different day and see "1:54 PM" here, and I wonder if that might confuse people into thinking it was sent today.

test/widgets/action_sheet_test.dart
chrisbobbe: Oh I think this isn't actually needed here; I should be able to simplify by dropping this commit.

chrisbobbe: Oh wait, not the whole commit, just this one change.

chrisbobbe:

test/widgets/action_sheet_test.dart
```diff
@@ -1674,7 +1750,11 @@ void main() {
             }
 
             // See comment in setupToMessageActionSheet about warnIfMissed: false
-            await tester.longPress(find.byType(MessageContent), warnIfMissed: false);
+            await tester.longPress(
+              find.descendant(
+                of: find.byType(MessageList),
+                matching: find.byType(MessageContent)),
```
gnprice: Same question as at https://github.com/zulip/zulip-flutter/pull/1624#discussion_r2181315307 , I think. In this one, there was an action sheet at a previous step of the test, but we already hit a button on it and then even used the compose box to save a message edit. So it seems like the story of the test calls for the action sheet to not be here now — which is why we need this step, after all, to bring it back.

It looks like nothing above waits for the action sheet to actually disappear after the button is hit, though; instead there's just `await tester.pump(Duration.zero);`. Maybe have that wait longer?

test/widgets/action_sheet_test.dart
gnprice: Hmm, good catch!

That does sound potentially confusing, so I think it'd be better to avoid it. Perhaps show the date (as if in a date separator or recipient header) and then the time? Or perhaps a refinement: skip the date if it's today.

test/widgets/action_sheet_test.dart
```diff
@@ -859,6 +860,81 @@ void main() {
...
+          message: message,
+          narrow: TopicNarrow.ofMessage(message));
+        checkSenderAndTimestampShown(tester, senderId: message.senderId);
+        check(find.descendant(
+          of: find.byType(BottomSheet),
+          matching: find.byType(UserMention))
+        ).findsOne();
+      });
+
+      testWidgets('muted sender', (tester) async {
```
gnprice: nit: clarify the expected outcome (as it's the opposite of what you might expect if you just look at the test name):
```suggestion
      testWidgets('muted sender also shown', (tester) async {
```

lib/widgets/message_list.dart
```diff
@@ -1779,8 +1779,8 @@ class _SenderRow extends StatelessWidget {
 
     final showAsMuted = store.isUserMuted(message.senderId)
       && message is Message // i.e., not an outbox message
-      && !MessageListPage.revealedMutedMessagesOf(context)
-                          .isMutedMessageRevealed((message as Message).id);
+      && (MessageListPage.maybeRevealedMutedMessagesOf(context)
+                          ?.isMutedMessageRevealed((message as Message).id) == false);
```
gnprice: This feels a bit hard to follow; and the behavior when there is no MessageListPage gets kind of buried here, whereas I feel like it should be more explicit and get a brief comment.

I'll send a quick PR to pull this whole condition out as a private method, which you can rebase atop to add this piece.

gnprice: Sent as #1664.

test/widgets/action_sheet_test.dart
chrisbobbe: Good catch; yes, thanks for spotting this.

chrisbobbe:
Thanks! Revision pushed, and I've put some thoughts about the timestamp format in a dartdoc.

lib/widgets/message_list.dart
```diff
@@ -1970,6 +1971,38 @@ class _SenderRow extends StatelessWidget {
   }
 }
 
+// TODO centralize on this for wherever we show message timestamps
+enum MessageTimestampDisplay {
```
gnprice: Interesting, I like this way of organizing it.

lib/widgets/message_list.dart
```diff
@@ -2223,7 +2257,7 @@ class OutboxMessageWithPossibleSender extends StatelessWidget {
       padding: const EdgeInsets.only(top: 4),
       child: Column(children: [
         if (item.showSender)
-          _SenderRow(message: message, showTimestamp: false),
+          SenderRow(message: message, timestampDisplay: null),
```
gnprice: How about:
```suggestion
          SenderRow(message: message, timestampStyle: MessageTimestampStyle.none),
```

(which in the near future will look like:
```dart
          SenderRow(message: message, timestampStyle: .none),
```
)

I think that'd be a bit more explicit than using null for this parameter.

Then the `format` method can return null for that style, so that the `timestamp` local in the build method is still null, which I think is helpful in that logic.

chrisbobbe:
Thanks! Update a screenshot and revision pushed, PTAL.

gnprice:
Thanks! LGTM.

@alya, do you have any feedback on the screenshots? (The one at the bottom left is current; the others have a previous version of the timestamp display but I believe are otherwise the same as the current revision.)

alya:
Generally looks good! Is it an illusion, or do we have different letter spacing in "Jul" vs. "AM" (which seems weird)?

Screenshot 2025-07-08 at 13 51 35

alya:
Do we use "Today at" and "Yesterday at" as appropriate? Not essential, but seems nice.

alya:
Do we need to handle long names running into the date/time, or is max sender name length short enough for that not to be an issue?

gnprice:
I think the letter spacing is just the letters A and M being wider, and so the font letting them get a bit closer together.

It's one string displayed with a single text style (font and all the other parameters).

chrisbobbe:
> Do we use "Today at" and "Yesterday at" as appropriate? Not essential, but seems nice.

Good question; I wrote down my thinking in a code comment (see `full`):

```dart
// TODO centralize on this for wherever we show message timestamps
enum MessageTimestampStyle {
  none,
  timeOnly,

  /// The longest format, with full date and time as numbers, not "Today"/etc.
  ///
  /// For UI contexts focused just on the one message,
  /// or as a tooltip on a shorter-formatted timestamp.
  ///
  /// The detail won't always be needed, but this format makes mental timezone
  /// conversions easier, which is helpful when the user is thinking about
  /// business hours on a different continent,
  /// or traveling and they know their device timezone setting is wrong, etc.
  full,
  ;
```

alya:
I dunno, to me the date is harder to think about than "Today" in the vast majority of cases when there's nothing weird about my time zone. I don't always know today's date.

chrisbobbe:
> Do we need to handle long names running into the date/time, or is max sender name length short enough for that not to be an issue?

A long name will just be truncated with "…" if it runs into the timestamp. We could allow it to wrap on a second line, if helpful?

alya:
Truncating seems fine. Just as long as it looks reasonable. :)

chrisbobbe:
> when there's nothing weird about my time zone

(Or the sender's time zone 🙂)

I can add a TODO for localized "Today"/"Yesterday"-style timestamps here. We have a plan for that in the lightbox (see #45) but it'll take investigating and possibly upstream work. Should the TODO say to follow the #45 format exactly, so with "21 minutes ago" etc.?

gnprice:
I think the TODO doesn't have to be precise about that; and it can link to this discussion :slightly_smiling_face:

gnprice:
I'm planning to include this PR in a v30.0.261 release tomorrow. It'd be great to have it merged first.

chrisbobbe:
Here's a long name getting truncated:

image

chrisbobbe:
Thanks both! Revision pushed, and @alya see screenshot in my previous comment here.

gnprice:
Thanks! Looks good; merging.

(If we find further design tweaks we want to make, we can always do those as follow-ups.)

PR #1627, opened 2025-06-24
KaTeX(5/n): Handle 'position' & 'top' property in KaTeX span inline style

rajveermalviya:
| Flutter | Web |
| - | - |
| image | image |

#### Survey results

```diff
-  …, 31380 of them were KaTeX containing messages and 3817 of those failed.
-  There were 1075 math block nodes out of which 614 failed.
-  There were 164014 math inline nodes out of which 6470 failed.
-  Because of hard fail: unsupported inline CSS property: top:
-    1546 messages failed.
-  Because of unsupported inline css property: position:
-    1356 messages failed.
+  …, 31380 of them were KaTeX containing messages and 2801 of those failed.
+  There were 1075 math block nodes out of which 538 failed.
+  There were 164014 math inline nodes out of which 4710 failed.
   Because of hard fail: unexpected CSS class for vlist inner span: svg-align:
-    644 messages failed.
-  Because of unsupported css class: accent:
-    609 messages failed.
+    648 messages failed.
   Because of unsupported inline css property: border-bottom-width:
-    605 messages failed.
+    631 messages failed.
+  Because of unsupported css class: accent:
+    614 messages failed.
   Because of unsupported css class: op-limits:
-    517 messages failed.
+    521 messages failed.
   Because of unsupported css class: fix:
-    473 messages failed.
+    488 messages failed.
   Because of unsupported css class: inner:
-    473 messages failed.
-  Because of unsupported css class: accent-body:
-    455 messages failed.
-  Because of unsupported inline css property: left:
-    455 messages failed.
+    488 messages failed.
   Because of unsupported css class: rlap:
-    450 messages failed.
+    465 messages failed.
   Because of unsupported css class: thinbox:
-    448 messages failed.
+    463 messages failed.
   Because of unsupported css class: vbox:
-    448 messages failed.
+    463 messages failed.
   Because of unsupported css class: nulldelimiter:
-    437 messages failed.
+    461 messages failed.
+  Because of unsupported css class: accent-body:
+    460 messages failed.
+  Because of unsupported inline css property: left:
+    460 messages failed.
   Because of unsupported css class: mfrac:
-    410 messages failed.
+    433 messages failed.
   Because of unsupported css class: frac-line:
-    389 messages failed.
+    416 messages failed.
   Because of unsupported css class: x-arrow:
-    389 messages failed.
+    392 messages failed.
   Because of unsupported css class: x-arrow-pad:
-    380 messages failed.
+    383 messages failed.
+  Because of hard fail: unsupported inline CSS property "top", when "position: null":
+    235 messages failed.
   Because of unsupported css class: delimcenter:
-    226 messages failed.
+    235 messages failed.
   Because of unsupported css class: mtable:
     209 messages failed.
@@ -53,9 +51,9 @@
     86 messages failed.
   Because of unsupported css class: col-align-r:
-    80 messages failed.
+    83 messages failed.
+  Because of hard fail: unsupported html node:
+    78 messages failed.
   Because of unsupported css class: arraycolsep:
     77 messages failed.
-  Because of hard fail: unsupported html node:
-    75 messages failed.
   Because of unsupported css class: sqrt:
     75 messages failed.
@@ -74,10 +72,10 @@
   Because of hard fail: unsupported html node: svg:
     28 messages failed.
+  Because of unsupported css class: newline:
+    28 messages failed.
   Because of unsupported css class: overlay:
     28 messages failed.
   Because of unsupported css class: llap:
     26 messages failed.
-  Because of unsupported css class: newline:
-    25 messages failed.
   Because of unsupported css class: cjk_fallback:
     20 messages failed.
@@ -127,3 +125,3 @@
     1 messages failed.                                                                                                                                                       
```

Fixes: #1671

### DIFF
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -637,6 +637,7 @@ class _KatexParser {
       marginLeftEm: _takeStyleEm(inlineStyles, 'margin-left'),
       marginRightEm: _takeStyleEm(inlineStyles, 'margin-right'),
       color: _takeStyleColor(inlineStyles, 'color'),
+      position: _takeStylePosition(inlineStyles, 'position'),
       // TODO handle more CSS properties
     );
     if (inlineStyles != null && inlineStyles.isNotEmpty) {
@@ -646,10 +647,10 @@ class _KatexParser {
         _hasError = true;
       }
     }
-    // Currently, we expect `top` to only be inside a vlist, and
-    // we handle that case separately above.
-    if (styles.topEm != null) {
-      throw _KatexHtmlParseError('unsupported inline CSS property: top');
+    if (styles.topEm != null && styles.position != KatexSpanPosition.relative) {
+      // The meaning of `top` would be different without `position: relative`.
+      throw _KatexHtmlParseError(
+        'unsupported inline CSS property "top" given "position: ${styles.position}"');
     }
 
     String? text;
@@ -771,6 +772,34 @@ class _KatexParser {
     _hasError = true;
     return null;
   }
+
+  /// Remove the given property from the given style map,
+  /// and parse as a CSS position value.
+  ///
+  /// If the property is present but is not a valid CSS position value,
+  /// record an error and return null.
+  ///
+  /// If the property is absent, return null with no error.
+  ///
+  /// If the map is null, treat it as empty.
+  ///
+  /// To produce the map this method expects, see [_parseInlineStyles].
+  KatexSpanPosition? _takeStylePosition(Map<String, css_visitor.Expression>? styles, String property) {
+    final expression = styles?.remove(property);
+    if (expression == null) return null;
+    if (expression case css_visitor.LiteralTerm(:final value)) {
+      if (value case css_visitor.Identifier(:final name)) {
+        if (name == 'relative') {
+          return KatexSpanPosition.relative;
+        }
+      }
+    }
+    assert(debugLog('KaTeX: Unsupported value for CSS property $property,'
+      ' expected a CSS position value: ${expression.toDebugString()}'));
+    unsupportedInlineCssProperties.add(property);
+    _hasError = true;
+    return null;
+  }
 }
 
 enum KatexSpanFontWeight {
@@ -786,6 +815,10 @@ enum KatexSpanTextAlign {
   left,
   center,
   right,
+}
+
+enum KatexSpanPosition {
+  relative,
 }
 
 class KatexSpanColor {
@@ -840,6 +873,7 @@ class KatexSpanStyles {
   final KatexSpanTextAlign? textAlign;
 
   final KatexSpanColor? color;
+  final KatexSpanPosition? position;
 
   const KatexSpanStyles({
     this.widthEm,
@@ -853,6 +887,7 @@ class KatexSpanStyles {
     this.fontStyle,
     this.textAlign,
     this.color,
+    this.position,
   });
 
   @override
@@ -869,6 +904,7 @@ class KatexSpanStyles {
     fontStyle,
     textAlign,
     color,
+    position,
   );
 
   @override
@@ -884,7 +920,8 @@ class KatexSpanStyles {
       other.fontWeight == fontWeight &&
       other.fontStyle == fontStyle &&
       other.textAlign == textAlign &&
-      other.color == color;
+      other.color == color &&
+      other.position == position;
   }
 
   @override
@@ -901,6 +938,7 @@ class KatexSpanStyles {
     if (fontStyle != null) args.add('fontStyle: $fontStyle');
     if (textAlign != null) args.add('textAlign: $textAlign');
     if (color != null) args.add('color: $color');
+    if (position != null) args.add('position: $position');
     return '${objectRuntimeType(this, 'KatexSpanStyles')}(${args.join(', ')})';
   }
 
@@ -917,6 +955,7 @@ class KatexSpanStyles {
     bool fontStyle = true,
     bool textAlign = true,
     bool color = true,
+    bool position = true,
   }) {
     return KatexSpanStyles(
       widthEm: widthEm ? this.widthEm : null,
@@ -930,6 +969,7 @@ class KatexSpanStyles {
       fontStyle: fontStyle ? this.fontStyle : null,
       textAlign: textAlign ? this.textAlign : null,
       color: color ? this.color : null,
+      position: position ? this.position : null,
     );
   }
 }
--- a/lib/widgets/katex.dart
+++ b/lib/widgets/katex.dart
@@ -96,10 +96,10 @@ class _KatexSpan extends StatelessWidget {
     }
 
     final styles = node.styles;
-
-    // Currently, we expect `top` to be only present with the
-    // vlist inner row span, and parser handles that explicitly.
-    assert(styles.topEm == null);
+    if (styles.topEm != null) {
+      // The meaning of `top` would be different without `position: relative`.
+      assert(styles.position == KatexSpanPosition.relative);
+    }
 
     final fontFamily = styles.fontFamily;
     final fontSize = switch (styles.fontSizeEm) {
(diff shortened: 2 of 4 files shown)


gnprice:
Thanks for building this! See my comments on #1670 for a few things I'd like to see on all the open KaTeX PRs: https://github.com/zulip/zulip-flutter/pull/1670#pullrequestreview-3045025908, https://github.com/zulip/zulip-flutter/pull/1670#issuecomment-3105055606 .

lib/model/katex.dart
```diff
@@ -765,6 +766,34 @@ class _KatexParser {
     _hasError = true;
     return null;
   }
+
+  /// Remove the given property from the given style map,
+  /// and parse as a literal value of CSS position attribute.
```
gnprice:
```suggestion
  /// and parse as a CSS position value.
```

Here's the relevant spec and MDN doc:
https://drafts.csswg.org/css-position/#position-property
https://developer.mozilla.org/en-US/docs/Web/CSS/position

Scanning those:
* I don't think "literal" adds anything to the meaning here.
* The CSS terminology is that `position` and things like it are "properties", not "attributes".

lib/model/katex.dart
```diff
@@ -640,10 +641,10 @@ class _KatexParser {
...
       }
     }
-    // Currently, we expect `top` to only be inside a vlist, and
-    // we handle that case separately above.
-    if (styles.topEm != null) {
-      throw _KatexHtmlParseError('unsupported inline CSS property: top');
+    // Currently, we expect `top` to be inside a vlist (which we handle it
+    // separately above), and if it's along with a `position: relative`.
+    if (styles.topEm != null && styles.position != KatexSpanPosition.relative) {
+      throw _KatexHtmlParseError('unsupported inline CSS property "top", when "position: ${styles.position}"');
```
gnprice: This comment doesn't parse for me.

I think the vlist part is beside the point now: we do expect a `top` property, but only when accompanied by `position: relative`.

And just saying that isn't useful as a comment, because it repeats what the code in the condition already says. The thing that is useful to say in a comment, though, is *why*. So:
```suggestion
    if (styles.topEm != null && styles.position != KatexSpanPosition.relative) {
      // The meaning of `top` would be different without `position: relative`.
      throw _KatexHtmlParseError(
        'unsupported inline CSS property "top" given "position: ${styles.position}"');
```

I.e.,
> The meaning of `top` would be different without `position: relative`.

As MDN says:
https://developer.mozilla.org/en-US/docs/Web/CSS/top
> The effect of top depends on how the element is positioned (i.e., the value of the [position](https://developer.mozilla.org/en-US/docs/Web/CSS/position) property):
> [… and then a list of 4 quite different meanings for the 5 possible values of `position`.]

lib/widgets/katex.dart
```diff
@@ -97,10 +97,6 @@ class _KatexSpan extends StatelessWidget {
 
     final styles = node.styles;
 
-    // Currently, we expect `top` to be only present with the
-    // vlist inner row span, and parser handles that explicitly.
-    assert(styles.topEm == null);
```
gnprice, line 102: This assert still seems quite relevant — but only in the case where `position` is null.

lib/widgets/katex.dart
```diff
@@ -180,6 +176,23 @@ class _KatexSpan extends StatelessWidget {
...
     }
 
+    switch (styles.position) {
+      case KatexSpanPosition.relative:
+        final offset = switch (styles.topEm) {
+          final topEm? => Offset(0, topEm * em),
+          null => null,
+        };
+
+        if (offset != null) {
```
gnprice: nit: these are closely related; keep joined as one stanza
```suggestion
        };
        if (offset != null) {
```

lib/widgets/katex.dart
```diff
@@ -180,6 +176,23 @@ class _KatexSpan extends StatelessWidget {
...
+          widget = Transform.translate(
+            offset: offset,
```
gnprice: Though actually this seems like it can be tightened further than that: check if `styles.topEm` is non-null, and if so then do the Transform.translate. It's fine to have a `!` null-assertion inside the body of that conditional — it's easy for the reader to see why the assertion will always hold.

gnprice (commented):
Thanks! Generally this all looks good — various small comments below.

Would you also post screenshots of an example where there's a bit more going on? With these screenshots, if we just ignored `top` then it'd be hard to tell from the screenshots that anything was wrong, because there's nothing next to the `\int` that it's getting aligned with. So even `\int dx` would help a lot.

---

It also looks like the value of `top` in this example is quite small:
> `top:-0.0011em;`

And in fact, that's small enough that it defeats the widget test in this PR. If I delete the logic in `_KatexSpan.build` that consumes `styles.topEm`, then the tests all still pass, because 1.1 milli-ems is a very small fraction of a pixel and is below the epsilon the test (appropriately) uses.

Can you find an example where the effect is bigger? Try other big operators, like `\sum`; or limits above or below the operators, like `\int_0^1`.

gnprice:
```diff
-  …, 31380 of them were KaTeX containing messages and 3817 of those failed.
-  There were 1075 math block nodes out of which 614 failed.
-  There were 164014 math inline nodes out of which 6470 failed.
-  Because of hard fail: unsupported inline CSS property: top:
-    1546 messages failed.
-  Because of unsupported inline css property: position:
-    1356 messages failed.
+  …, 31380 of them were KaTeX containing messages and 2801 of those failed.
```

A very nice improvement! Will be glad to have this in.

rajveermalviya:

I wasn't able to find an example with big operators having a larger `top` offset, but I did find another case where `position: relative; top: …` inline style is present i.e for `/colonequals` and it also has a larger `top` offset. So, added a test for that.

test/model/katex_test.dart
```diff
@@ -643,6 +643,72 @@ class KatexExample extends ContentExample {
           text: '∗'),
       ]),
     ]);
+
+  static final bigOperators = KatexExample.block(
+    r'big operators: \int',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2240766
+    r'\int dx',
```
gnprice: Ah, I wasn't clear enough — for making this example more complex, I really did mean just the screenshots :slightly_smiling_face:. In the test code, it's nice to keep it as simple as possible. (And this test doesn't really gain anything from having the additional elements.)

test/widgets/katex_test.dart
```diff
@@ -73,6 +73,15 @@ void main() {
         ('X', Offset(0.00, 7.04), Size(17.03, 25.00)),
         ('n', Offset(17.03, 15.90), Size(8.63, 17.00)),
       ]),
+      (KatexExample.bigOperators, skip: false, [
+        ('∫', Offset(0.00, 12.02), Size(11.43, 25.00)),
+        ('d', Offset(24.00, 12.03), Size(10.69, 25.00)),
+        ('x', Offset(34.70, 12.03), Size(11.76, 25.00)),
+      ]),
```
gnprice: Maybe even leave this one out of this widget test, since the test isn't sensitive enough to pick up whether the code is working.

rajveermalviya:
> spacing between the colon and the equals-sign seems to be noticeably wider in Flutter than on web

Filed #1810 for this, the issue is that the negative-margin ends up being no-op because the span with negative margin is inside another span, and it doesn't have any siblings after it in that subtree. This can be confirmed on Web by toggling the `margin-right` style on `mspace` span via web dev console.

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

gnprice:
> the negative-margin ends up being no-op because the span with negative margin is inside another span, and it doesn't have any siblings after it in that subtree

Aha, makes sense, thanks. I was wondering if/when that limitation in our current implementation would bite us — I guess this is an answer.

gnprice:
(And, to be clear, this `\colonequals` symptom is a pretty minor one.)

PR #1628, opened 2025-06-24
Decrease whitespace at end of message feed

chrisbobbe:
This does two things:

- Shrinks a 36px gap to 12px (the gap below the mark-as-read button)
- Removes 11px bottom padding on the last message in the feed (but keeps it when followed by a recipient header; see https://github.com/zulip/zulip-flutter/pull/1453#discussion_r2106526985)

Proposed by Tim in chat:

[#mobile > space at end of thread @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/space.20at.20end.20of.20thread/near/2203404)

> I bet those two tweaks might fully address this issue.

Screenshots coming soon.

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1167,8 +1167,8 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
         // TODO perhaps offer mark-as-read even when not done fetching?
         MarkAsReadWidget(narrow: widget.narrow),
         // To reinforce that the end of the feed has been reached:
-        //   https://chat.zulip.org/#narrow/stream/243-mobile-team/topic/flutter.3A.20Mark-as-read/near/1680603
-        const SizedBox(height: 36),
+        //   https://chat.zulip.org/#narrow/channel/48-mobile/topic/space.20at.20end.20of.20thread/near/2203391
+        const SizedBox(height: 12),
       ]);
     } else if (model.busyFetchingMore) {
       // See [_buildStartCap] for why this condition shows a loading indicator.
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1073,7 +1073,7 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
 
           final itemIndex = totalItems - 1 - (childIndex + bottomItems);
           final data = model.items[itemIndex];
-          final item = _buildItem(data);
+          final item = _buildItem(data, isLastInFeed: itemIndex == totalItems - 1);
           return item;
         }));
 
@@ -1113,7 +1113,7 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
 
           final itemIndex = topItems + childIndex;
           final data = model.items[itemIndex];
-          return _buildItem(data);
+          return _buildItem(data, isLastInFeed: itemIndex == totalItems - 1);
         }));
 
     if (!ComposeBox.hasComposeBox(widget.narrow)) {
@@ -1178,7 +1178,7 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
     }
   }
 
-  Widget _buildItem(MessageListItem data) {
+  Widget _buildItem(MessageListItem data, {required bool isLastInFeed}) {
     switch (data) {
       case MessageListRecipientHeaderItem():
         final header = RecipientHeader(message: data.message, narrow: widget.narrow);
@@ -1195,12 +1195,14 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
           key: ValueKey(data.message.id),
           narrow: widget.narrow,
           header: header,
+          isLastInFeed: isLastInFeed,
           item: data);
       case MessageListOutboxMessageItem():
         final header = RecipientHeader(message: data.message, narrow: widget.narrow);
         return MessageItem(
           narrow: widget.narrow,
           header: header,
+          isLastInFeed: isLastInFeed,
           item: data);
     }
   }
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Dark mode:

| Current | 11px to 0px | 11px to 5px |
| --- | --- | --- |
| image | image | image |
| image | image | image |
| image | image | image |

chrisbobbe:
Light mode:

| Before | 11px to 0px | 11px to 5px |
| --- | --- | --- |
| image | image | image |
| image | image | image |
| image | image | image |

chrisbobbe:
Updated, this time decreasing the 11px to 5px instead of making it zero. I've kept the screenshots from the previous revision, though, and added ones from this revision. PTAL!

alya:
5px seems like a fine adjustment to me.

lib/model/message_list.dart
```diff
@@ -447,31 +451,37 @@ mixin _MessageSequence {
...
-    if (
-      prevMessage == null
-      || oneMessagePerBlock
-      || !haveSameRecipient(prevMessage, message)
-    ) {
+    if (prevMessage == null) {
       items.add(MessageListRecipientHeaderItem(message));
       canShareSender = false;
     } else {
       assert(items.last is MessageListMessageBaseItem);
```
gnprice: msglist [nfc]: Pull out some asserts that also apply when haveSameRecipient
    
    These invariants apply regardless of `haveSameRecipient`.

That also apply when *not* haveSameRecipient, right?

(The case where they were previously being asserted is one where haveSameRecipient is always true.)

lib/widgets/message_list.dart
```diff
@@ -1546,9 +1546,10 @@ class MessageItem extends StatelessWidget {
             item: item),
           MessageListOutboxMessageItem() => OutboxMessageWithPossibleSender(item: item),
         },
-        // TODO refine this padding; discussion:
-        //   https://github.com/zulip/zulip-flutter/pull/1453#discussion_r2106526985
-        if (item.isLastInBlock) const SizedBox(height: 11),
+        if (item.isLastInBlock && !item.isLastInFeed)
+          const SizedBox(height: 11)
+        else if (item.isLastInFeed)
+          const SizedBox(height: 5),
```
gnprice: nit: can swap these cases to simplify one of the conditions

lib/model/message_list.dart
```diff
@@ -447,31 +451,37 @@ mixin _MessageSequence {
...
       final prevMessageItem = items.last as MessageListMessageBaseItem;
       assert(identical(prevMessageItem.message, prevMessage));
       assert(prevMessageItem.isLastInBlock);
-      prevMessageItem.isLastInBlock = false;
+      assert(prevMessageItem.isLastInFeed);
 
-      if (!messagesSameDay(prevMessageItem.message, message)) {
-        items.add(MessageListDateSeparatorItem(message));
+      prevMessageItem.isLastInFeed = false;
```
gnprice: I'd rather avoid making this _addItemsForMessage method more complicated if we can help it — it has a complicated job already as it is.

The question "is this the last item in the feed" seems like it can be answered easily within the MessageList widget, by looking at the index in the list. After all, the invariant says:
```dart
      ..isLastInFeed.equals(i == model.items.length);
```
(where `i` is the index after the item in question).

So instead how about:
* MessageItem gains a boolean field isLastInFeed.
* _buildItem gains a matching parameter.
* Its two call sites compare `itemIndex` to `totalItems` to determine that parameter.

chrisbobbe:
Thanks! Revision pushed.

I've left in the two NFC commits because they seemed potentially helpful, but are now orthogonal to the interesting changes here:

359b99ac2 msglist [nfc]: Make a common codepath for prevMessage != null
9c17fca46 msglist [nfc]: Pull out asserts that also apply at start of new block

Happy to remove or revise though if helpful.

lib/model/message_list.dart
chrisbobbe: Sure, that's reasonable; I've done this in my new revision.

It puts a bit more of the work in widget code, which doesn't have tests for this new padding logic; would that be a blocker to merging?

lib/widgets/message_list.dart
```diff
@@ -1195,13 +1195,15 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
           key: ValueKey(data.message.id),
           narrow: widget.narrow,
           header: header,
-          item: data);
+          item: data,
+          isLastInFeed: isLastInFeed);
```
gnprice: nit: metadata before payload / `child`-like argument (so here, isLastInFeed before header and item)

lib/model/message_list.dart
gnprice: Yeah, ideally there'd be a widget test or two for this. (I guess in particular one for the usual case, and one for the case where the bottom sliver is empty but the top is nonempty, checking effectively that that first `_buildItem` call site is doing the right thing.)

If that isn't quick to write, though, it's not a blocker for merging.

lib/model/message_list.dart
```diff
@@ -448,25 +448,27 @@ mixin _MessageSequence {
...
+      assert(prevMessageItem.isLastInBlock); // though this may mutate; see below
 
-      if (!messagesSameDay(prevMessageItem.message, message)) {
-        items.add(MessageListDateSeparatorItem(message));
+      if (oneMessagePerBlock || !haveSameRecipient(prevMessage, message)) {
+        items.add(MessageListRecipientHeaderItem(message));
         canShareSender = false;
       } else {
-        canShareSender = prevMessageItem.message.senderId == message.senderId;
+        prevMessageItem.isLastInBlock = false;
```
gnprice: Yeah, I think the refactors in the middle two commits are no longer helpful. The assertions are really there to support the `prevMessageItem` local, and then the last one is for checking the precondition of the assignment right after it:
```dart
      assert(prevMessageItem.isLastInBlock);
      prevMessageItem.isLastInBlock = false;
```

So separating them from that line, and from the uses of `prevMessageItem` in general, isn't an improvement.

gnprice (commented):
Thanks for the revision! Just small comments.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Merged.

Issue #1578, opened 2025-06-13
Support toggling "Invisible mode"


chrisbobbe:
This setting: https://zulip.com/help/status-and-availability#invisible-mode

Implemented in zulip-mobile in https://github.com/zulip/zulip-mobile/pull/5531

### Related issues

- #196

chrisbobbe:
When this is done we should update the Help Center doc. We removed instructions for the legacy app when preparing to launch the new app: zulip/zulip@815a522b86a990f0729ecf79652cff34b42bb891

chrisbobbe:
Someone asked for this feature in the first week of the launch period. (A Play Store review)

timabbott:
This was requested via an issue in the legacy repository here: https://github.com/zulip/zulip-mobile/issues/5927.

chrisbobbe:
Do we have a design for this? @terpimost @alya 

Started a thread: [#mobile-design > presence: invisible mode @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/presence.3A.20invisible.20mode/near/2202874)

gnprice:
@jb2barrels You may be glad to hear (since you filed zulip/zulip-mobile#5927) that the fix for this issue went out in yesterday's release 30.0.261: [#announce > mobile releases @ 💬](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2217685). Please try it out and see what you think!

PR #1629, opened 2025-06-24
Track user status

sm-sayedi:
Support for tracking the user status.

Message List:
<table>
    <tr>
        <td><b>Normal display name</b></td>
        <td><b>Long display name</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>
 
Recent DMs Page:
<table>
    <tr>
        <td><b>Normal display name</b></td>
        <td><b>Long display name</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>
 
New DM Sheet
<table>
    <tr>
        <td><b>Normal display name</b></td>
        <td><b>Long display name</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>
 
User Autocomplete:
<table>
    <tr>
        <td><b>Normal display name</b></td>
        <td><b>Long display name</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

Profile Page:
<table>
    <tr>
        <td><b>Normal display name</b></td>
        <td><b>Normal display name - Dark</b></td>
        <td><b>Long display name</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
        <td></td>
    </tr>
</table>
 
Fixes: #197

Issue #741, opened 2024-06-18
Consistently interpret API errors for user


gnprice:
When the user tries to do something that calls for an API request, and the request fails, we should show them an error dialog with a useful error message.

In some cases we'll have specific logic to interpret the error, but in general we should show them (a) the error message the server sent, if we got one; (b) if we weren't even able to complete the HTTP request, then whatever error message we got from the system about that (via our `NetworkException` type).

We already do this in a lot of places, in particular on the various actions the user can choose from the message action sheet (in `lib/widgets/action_sheet.dart`). But in other places we call `showErrorDialog` with something like `e.toString()`, where `e` is just the exception we caught.

This issue is to
* factor that logic out in a clean way, to the extent it's reusable (which a lot of it is);
* test that factored-out code for all the different cases of what an `ApiRequestException` can look like, so that we're making explicit choices about what the error message looks like in each case;
* use that logic consistently. In particular, search for call sites of `showErrorDialog` and make sure those use this logic where appropriate; and search for `TODO` comments referring to this issue, and resolve those.

Related issues, which may overlap with this one:
* #105 
* #720

PR #1631, opened 2025-06-26
Offer "Invisible mode" toggle switch!

chrisbobbe:
~~Still TODO: tests.~~

Some known differences from the [Figma](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60682&m=dev), which won't be hard to fix but require upstream changes, so not totally trivial:

- When the switch is on:
  - The switch should be 28px by 48px, not 36px by 56px.
  - The "thumb" (white circle with checkmark) should have radius 10px, not 12px.

- When the switch is off:
  - The switch should be 26px by 46px, not 34px by 54px.
  - The "thumb" should have radius 7px, not 8px.

(edit: screenshots of what it would look like with these differences resolved: https://github.com/zulip/zulip-flutter/pull/1631#issuecomment-3009660346 , and I filed #1636 for these fixes)

Fixes: #1578

### DIFF
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -251,6 +251,7 @@ class UserSettings {
   bool twentyFourHourTime;
   bool? displayEmojiReactionUsers; // TODO(server-6)
   Emojiset emojiset;
+  bool presenceEnabled;
 
   // TODO more, as needed. When adding a setting here, please also:
   // (1) add it to the [UserSettingName] enum
@@ -261,6 +262,7 @@ class UserSettings {
     required this.twentyFourHourTime,
     required this.displayEmojiReactionUsers,
     required this.emojiset,
+    required this.presenceEnabled,
   });
 
   factory UserSettings.fromJson(Map<String, dynamic> json) =>
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -68,13 +68,7 @@ class InitialSnapshot {
 
   final List<ZulipStream> streams;
 
-  // Servers pre-5.0 don't have `user_settings`, and instead provide whatever
-  // user settings they support at toplevel in the initial snapshot. Since we're
-  // likely to desupport pre-5.0 servers before wide release, we prefer to
-  // ignore the toplevel fields and use `user_settings` where present instead,
-  // even at the expense of functionality with pre-5.0 servers.
-  // TODO(server-5) remove pre-5.0 comment
-  final UserSettings? userSettings; // TODO(server-5)
+  final UserSettings userSettings;
 
   final List<UserTopicItem>? userTopics; // TODO(server-6)
 
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -257,6 +257,7 @@ class UserSettings {
   // (1) add it to the [UserSettingName] enum
   // (2) then re-run the command to refresh the .g.dart files
   // (3) handle the event that signals an update to the setting
+  // (4) add the setting to the [updateSettings] route binding
 
   UserSettings({
     required this.twentyFourHourTime,
--- a/lib/api/route/settings.dart
+++ b/lib/api/route/settings.dart
@@ -1,0 +1,22 @@
+import '../core.dart';
+import '../model/model.dart';
+
+/// https://zulip.com/api/update-settings
+Future<void> updateSettings(ApiConnection connection, {
+  required Map<UserSettingName, Object?> newSettings,
+}) {
+  final params = <String, Object?>{};
+  for (final entry in newSettings.entries) {
+    final name = entry.key;
+    final valueRaw = entry.value;
+    final value = switch (name) {
+      UserSettingName.twentyFourHourTime => valueRaw as bool,
+      UserSettingName.displayEmojiReactionUsers => valueRaw as bool,
+      UserSettingName.emojiset => RawParameter((valueRaw as Emojiset).toJson()),
+      UserSettingName.presenceEnabled => valueRaw as bool,
+    };
+    params[name.toJson()] = value;
+  }
+
+  return connection.patch('updateSettings', (_) {}, 'settings', params);
+}
--- a/lib/basic.dart
+++ b/lib/basic.dart
@@ -19,6 +19,12 @@ sealed class Option<T> {
 
   /// The value contained in this option, if any; else the given value.
   T or(T optb);
+
+  /// The value contained in this option, if any;
+  /// else the value returned by [fn].
+  ///
+  /// [fn] is called only if its return value is needed.
+  T orElse(T Function() fn);
 }
 
 class OptionNone<T> extends Option<T> {
@@ -26,6 +32,9 @@ class OptionNone<T> extends Option<T> {
 
   @override
   T or(T optb) => optb;
+
+  @override
+  T orElse(T Function() fn) => fn();
 
   @override
   bool operator ==(Object other) => other is OptionNone;
@@ -44,6 +53,9 @@ class OptionSome<T> extends Option<T> {
 
   @override
   T or(T optb) => value;
+
+  @override
+  T orElse(T Function() fn) => value;
 
   @override
   bool operator ==(Object other) => other is OptionSome && value == other.value;
--- a/lib/basic.dart
+++ b/lib/basic.dart
@@ -1,0 +1,56 @@
+
+/// Either a value, or the absence of a value.
+///
+/// An `Option<T>` is either an `OptionSome` representing a `T` value,
+/// or an `OptionNone` representing the absence of a value.
+///
+/// When `T` is non-nullable, this is the same information that is
+/// normally represented as a `T?`.
+/// This class is useful when T is nullable (or might be nullable).
+/// In that case `null` is already a T value,
+/// and so can't also be used to represent the absence of a T value,
+/// but `OptionNone()` is a different value from `OptionSome(null)`.
+///
+/// This interface is small because members are added lazily when needed.
+/// If adding another member, consider borrowing the naming from Rust:
+///   https://doc.rust-lang.org/std/option/enum.Option.html
+sealed class Option<T> {
+  const Option();
+
+  /// The value contained in this option, if any; else the given value.
+  T or(T optb);
+}
+
+class OptionNone<T> extends Option<T> {
+  const OptionNone();
+
+  @override
+  T or(T optb) => optb;
+
+  @override
+  bool operator ==(Object other) => other is OptionNone;
+
+  @override
+  int get hashCode => 'OptionNone'.hashCode;
+
+  @override
+  String toString() => 'OptionNone';
+}
+
+class OptionSome<T> extends Option<T> {
+  const OptionSome(this.value);
+
+  final T value;
+
+  @override
+  T or(T optb) => value;
+
+  @override
+  bool operator ==(Object other) => other is OptionSome && value == other.value;
+
+  @override
+  int get hashCode => Object.hash('OptionSome', value);
+
+  @override
+  String toString() => 'OptionSome($value)';
+}
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -289,11 +289,18 @@ class MenuButton extends StatelessWidget {
     required this.label,
     required this.onPressed,
     this.icon,
+    this.toggle,
   });
 
   final String label;
   final VoidCallback onPressed;
   final IconData? icon;
+
+  /// A [Toggle] to go before [icon], or in its place if it's null.
+  ///
+  /// See Figma:
+  ///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60682&m=dev
+  final Widget? toggle;
 
   static double itemSpacing = 16;
 
@@ -319,26 +326,37 @@ class MenuButton extends StatelessWidget {
     assert(Theme.of(context).visualDensity == VisualDensity.standard);
 
     return MenuItemButton(
-      trailingIcon: icon != null
+      trailingIcon: (icon != null || toggle != null)
         ? Padding(
             // This Material widget gives us 12px padding before the icon --
             // or more or less, depending on Theme.of(context).visualDensity,
             // hence the `assert` above.
             padding: EdgeInsetsDirectional.only(start: itemSpacing - 12),
 
-            child: Icon(icon, color: designVariables.contextMenuItemText))
+            child: Row(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.center,
+              spacing: itemSpacing,
+              children: [
+                if (toggle != null) toggle!,
+                if (icon != null) Icon(icon!, color: designVariables.contextMenuItemText),
+              ]))
         : null,
       style: MenuItemButton.styleFrom(
-        padding: const EdgeInsets.symmetric(vertical: 12, horizontal: 16),
+        minimumSize: Size.fromHeight(48),
+        padding: const EdgeInsets.symmetric(horizontal: 16),
         foregroundColor: designVariables.contextMenuItemText,
         splashFactory: NoSplash.splashFactory,
       ).copyWith(backgroundColor: WidgetStateColor.resolveWith((states) =>
           designVariables.contextMenuItemBg.withFadedAlpha(
             states.contains(WidgetState.pressed) ? 0.20 : 0.12))),
       onPressed: onPressed,
-      child: Text(label,
-        style: const TextStyle(fontSize: 20, height: 24 / 20)
-          .merge(weightVariableTextStyle(context, wght: 600))));
+      child: Padding(
+        padding: const EdgeInsets.symmetric(vertical: 4),
+        child: Text(label,
+          style: const TextStyle(fontSize: 20, height: 24 / 20)
+            .merge(weightVariableTextStyle(context, wght: 600))),
+      ));
   }
 }
 
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -257,3 +257,75 @@ class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
         child: widget.child));
   }
 }
+
+/// The rounded-rectangle shape and 1-pixel spacing for a run of [MenuButton]s.
+class MenuButtonsShape extends StatelessWidget {
+  const MenuButtonsShape({
+    super.key,
+    required this.buttons,
+  });
+
+  final List<Widget> buttons;
+
+  @override
+  Widget build(BuildContext context) {
+    return ClipRRect(
+      borderRadius: BorderRadius.circular(7),
+      child: Column(spacing: 1,
+        children: buttons));
+  }
+}
+
+/// The "menu button" component in Figma.
+///
+/// Must have a [MenuButtonsShape] ancestor.
+///
+/// See Figma:
+///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60681&m=dev
+class MenuButton extends StatelessWidget {
+  const MenuButton({
+    super.key,
+    required this.label,
+    required this.onPressed,
+    this.icon,
+  });
+
+  final String label;
+  final VoidCallback onPressed;
+  final IconData? icon;
+
+  static bool _debugCheckShapeAncestor(BuildContext context) {
+    final ancestor = context.findAncestorWidgetOfExactType<MenuButtonsShape>();
+    assert(() {
+      if (ancestor != null) return true;
+      throw FlutterError.fromParts([
+        ErrorSummary('No MenuButtonsShape ancestor found.'),
+        ErrorDescription('MenuButton widgets require a MenuButtonsShape ancestor.'),
+      ]);
+    }());
+    return true;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    _debugCheckShapeAncestor(context);
+
+    final designVariables = DesignVariables.of(context);
+
+    return MenuItemButton(
+      trailingIcon: icon != null
+        ? Icon(icon, color: designVariables.contextMenuItemText)
+        : null,
+      style: MenuItemButton.styleFrom(
+        padding: const EdgeInsets.symmetric(vertical: 12, horizontal: 16),
+        foregroundColor: designVariables.contextMenuItemText,
+        splashFactory: NoSplash.splashFactory,
+      ).copyWith(backgroundColor: WidgetStateColor.resolveWith((states) =>
+          designVariables.contextMenuItemBg.withFadedAlpha(
+            states.contains(WidgetState.pressed) ? 0.20 : 0.12))),
+      onPressed: onPressed,
+      child: Text(label,
+        style: const TextStyle(fontSize: 20, height: 24 / 20)
+          .merge(weightVariableTextStyle(context, wght: 600))));
+  }
+}
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 
 import 'color.dart';
+import 'icons.dart';
 import 'text.dart';
 import 'theme.dart';
 
@@ -338,5 +339,72 @@ class MenuButton extends StatelessWidget {
       child: Text(label,
         style: const TextStyle(fontSize: 20, height: 24 / 20)
           .merge(weightVariableTextStyle(context, wght: 600))));
+  }
+}
+
+/// The "toggle" component in Figma.
+///
+/// See Figma:
+///    https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60682&m=dev
+class Toggle extends StatelessWidget {
+  const Toggle({
+    super.key,
+    required this.value,
+    required this.onChanged,
+  });
+
+  final bool value;
+  final ValueChanged<bool> onChanged;
+
+  @override
+  Widget build(BuildContext context) {
+    // Figma has this (blue/500) in both light and dark mode.
+    // TODO(#831)
+    final activeColor = Color(0xff4370f0);
+
+    // Figma has this (grey/400) in both light and dark mode.
+    // TODO(#831)
+    final inactiveColor = Color(0xff9194a3);
+
+    // TODO(#1636):
+    //   All of these just need _SwitchConfig to be exposed,
+    //   and there's an upstream issue for that:
+    //     https://github.com/flutter/flutter/issues/131478
+    //
+    //   - active thumb radius should be 10px, not 12px
+    //     (_SwitchConfig.thumbRadiusWithIcon)
+    //   - inactive thumb radius should be 7px, not 8px
+    //     (_SwitchConfig.inactiveThumbRadius)
+    //   - track dimensions before trackOutlineWidth should be 24px by 44px,
+    //     not 32px by 52px (_SwitchConfig.trackHeight and trackWidth).
+
+    return Switch(
+      value: value,
+      onChanged: onChanged,
+      padding: EdgeInsets.zero,
+      splashRadius: 0,
+      thumbIcon: WidgetStateProperty<Icon?>.fromMap({
+        WidgetState.selected: Icon(ZulipIcons.check, size: 16, color: activeColor),
+        ~WidgetState.selected: null,
+      }),
+
+      // Figma has white for "on" and "off" in both light and dark mode.
+      thumbColor: WidgetStatePropertyAll(Colors.white),
+
+      activeTrackColor: activeColor,
+      inactiveTrackColor: inactiveColor,
+      trackOutlineColor: WidgetStateColor.fromMap({
+        WidgetState.selected: activeColor,
+        ~WidgetState.selected: inactiveColor,
+      }),
+      trackOutlineWidth: WidgetStateProperty<double>.fromMap({
+        // The outline is effectively painted with strokeAlignCenter:
+        //   https://api.flutter.dev/flutter/painting/BorderSide/strokeAlignCenter-constant.html
+        WidgetState.selected: 2 * 2,
+        ~WidgetState.selected: 1 * 2,
+      }),
+      overlayColor: WidgetStatePropertyAll(Colors.transparent),
+      materialTapTargetSize: MaterialTapTargetSize.shrinkWrap,
+    );
   }
 }
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -339,7 +339,7 @@ class MenuButton extends StatelessWidget {
               spacing: itemSpacing,
               children: [
                 if (toggle != null) toggle!,
-                if (icon != null) Icon(icon!, color: designVariables.contextMenuItemText),
+                if (icon != null) Icon(icon!, color: designVariables.contextMenuItemIcon),
               ]))
         : null,
       style: MenuItemButton.styleFrom(
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -259,7 +259,7 @@ class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
   }
 }
 
-/// The rounded-rectangle shape and 1-pixel spacing for a run of [MenuButton]s.
+/// The rounded-rectangle shape and 1-pixel spacing for a run of [ZulipMenuItemButton]s.
 class MenuButtonsShape extends StatelessWidget {
   const MenuButtonsShape({
     super.key,
@@ -277,21 +277,25 @@ class MenuButtonsShape extends StatelessWidget {
   }
 }
 
-/// The "menu button" component in Figma.
+/// The "menu button" or "list button" component in Figma.
+///
+/// Use [ZulipMenuItemButtonStyle] to choose between components.
 ///
 /// Must have a [MenuButtonsShape] ancestor.
 ///
 /// See Figma:
 ///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60681&m=dev
-class MenuButton extends StatelessWidget {
-  const MenuButton({
+class ZulipMenuItemButton extends StatelessWidget {
+  const ZulipMenuItemButton({
     super.key,
+    this.style = ZulipMenuItemButtonStyle.menu,
     required this.label,
     required this.onPressed,
     this.icon,
     this.toggle,
   });
 
+  final ZulipMenuItemButtonStyle style;
   final String label;
   final VoidCallback onPressed;
   final IconData? icon;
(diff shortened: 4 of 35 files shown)


chrisbobbe:
Self-profile

| Light | Dark |
| --- | --- |
| image | image |
| image | image |

Non-self profile (no change):

| Light | Dark |
| --- | --- |
| image | image |

chrisbobbe:
Here's how it could look after some upstream work (https://github.com/flutter/flutter/issues/131478). This removes the differences from the Figma that I mentioned in the PR description:

<details>

| Before | After |
| --- | --- |
| image | image |
| image | image |

</details>

chrisbobbe:
And I did go ahead and download GIPHY CAPTURE to record the animation and drag behavior 🙂:

(This is the PR as-is, not with the upstream work from my previous comment)

| The switch responds to taps | The button responds to taps | The "thumb" responds to dragging |
| --- | --- | --- |
| Jun-26-2025 12-28-36 | Jun-26-2025 12-29-31 | Jun-26-2025 12-30-36 |

chrisbobbe:
Error feedback when failing to turn mode on / off, and tapping the "Details" button in the snackbar.

The "Details" dialog isn't user-friendly. We could omit the "Details" button, but it might be helpful for debugging unexplained failures.

| On | Off |
| --- | --- |
| Jun-26-2025 12-33-56 | Jun-26-2025 12-36-06 |

alya:
Yeah, I guess depends on  how much we expect users to actually send us those details. If not, probably not super helpful to show them, unless we can translate them to something more readable (e.g., just say "Network error" or "Couldn't connect to server." or something).

timabbott:
I think Sentry is for debugging, and we want some generic network error as the thing to present in the UI.

chrisbobbe:
Makes sense; here's with the "Details" button removed:

| Light/On | Dark/Off |
| --- | --- |
| image | image |

chrisbobbe:
(And I'll add a TODO(#741) for interpreting the error)

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,217 @@
...
+  ///
+  /// The common case is when the API request fails quickly.
+  ///
+  /// (Another case is when spam-tapping a toggle switch,
+  /// if a user wants to do that.
+  /// The timer resets on [PerAccountSettingBuilderFn.handleRequestNewValue],
+  /// so until the spam-taps are finished, the switch responds only to the taps,
+  /// not to the event stream.
+  /// Then when the taps stop, it settles to the value from the latest event.)
+  static final Duration localEchoMininum = Duration(seconds: 1);
```
rajveermalviya: nit:
```suggestion
  static final Duration localEchoMinimum = Duration(seconds: 1);
```

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,217 @@
...
+      _lowerBoundCompleter!.complete();
+      _lowerBoundCompleter = null;
+    });
+
+    _upperBoundTimer?.cancel();
+    _upperBoundTimer = Timer(PerAccountSettingBuilder.localEchoIdleTimeout, () {
+      value = null;
+    });
+
+  }
```
rajveermalviya: nit:
```suggestion
    });
  }
```

chrisbobbe:
Thanks for the review! Revision pushed, with some tests added.

lib/api/route/settings.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:json_annotation/json_annotation.dart';
+
+import '../core.dart';
+import '../model/model.dart';
+
+part 'settings.g.dart';
+
+/// https://zulip.com/api/update-settings
+Future<UpdateSettingsResult> updateSettings(ApiConnection connection, {
```
gnprice: nit:

    api: Add updateSettings

add the word "route" at the end — it wasn't clear to me before reading the code what sort of a thing "updateSettings" was

lib/api/model/initial_snapshot.dart
```diff
@@ -254,16 +248,19 @@ class UserSettings {
   bool twentyFourHourTime;
   bool? displayEmojiReactionUsers; // TODO(server-6)
   Emojiset emojiset;
+  bool presenceEnabled;
 
   // TODO more, as needed. When adding a setting here, please also:
   // (1) add it to the [UserSettingName] enum
   // (2) then re-run the command to refresh the .g.dart files
   // (3) handle the event that signals an update to the setting
+  // (4) add the setting to the updateSettings binding
```
gnprice: nit: similarly,
```suggestion
  // (4) add the setting to the [updateSettings] route binding
```

lib/api/route/settings.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  return connection.patch('updateSettings', UpdateSettingsResult.fromJson, 'settings',
+    params);
+}
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class UpdateSettingsResult {
+  UpdateSettingsResult();
+
+  factory UpdateSettingsResult.fromJson(Map<String, dynamic> json) =>
```
gnprice: Since there's no content in this result, we can skip it — just have the endpoint function return `Future<void>`. Grep for "Future<void>" in this directory to see other examples.

test/api/route/settings_test.dart
```diff
@@ -0,0 +1,43 @@
...
+void main() {
+  test('smoke updateSettings', () {
+    return FakeApiConnection.with_((connection) async {
+      final response = UpdateSettingsResult();
+      connection.prepare(json: response.toJson());
+
+      final newSettings = <UserSettingName, Object?>{};
+      final expectedBodyFields = <String, String>{};
+
+      for (final name in UserSettingName.values) {
```
gnprice: nit: join declarations with their initializers
```suggestion
      final newSettings = <UserSettingName, Object?>{};
      final expectedBodyFields = <String, String>{};
      for (final name in UserSettingName.values) {
```

test/api/route/settings_test.dart
```diff
@@ -0,0 +1,43 @@
...
+        switch (name) {
+          case UserSettingName.twentyFourHourTime:
+            newSettings[name] = true;
+            expectedBodyFields['twenty_four_hour_time'] = 'true';
+          case UserSettingName.displayEmojiReactionUsers:
```
gnprice, line 22: Interesting, so this ensures that future additions to the enum get tested on this endpoint.

lib/widgets/button.dart
```diff
@@ -257,3 +258,167 @@ class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
...
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.center,
+              spacing: itemSpacing,
+              children: [
+                if (beforeIcon != null) beforeIcon!,
+                if (icon != null) Icon(icon, color: designVariables.contextMenuItemText),
+              ]))
+        : null,
+      style: MenuItemButton.styleFrom(
+        padding: const EdgeInsets.symmetric(horizontal: 16),
```
gnprice: button [nfc]: Factor MenuButton's vertical padding as around just the text
    
    This gives flexibility to pass a switch component with built-in
    vertical padding as `beforeIcon`.

Hmm, is this really NFC?

The way it would change behavior is if trailingIcon were taller than the text. Can that happen when the text scale factor is small? (I'm not immediately finding what controls the size of the icon.)

gnprice: If this can indeed shrink things, then we probably want a ConstrainedBox with minHeight: 48 around the whole thing.

And then in fact it's probably best to have much less vertical padding here on the text, like 4px or 2px or 0px. If the text is at large size, it's not clear that it's helpful to also have a bunch of padding above and below it.

lib/widgets/button.dart
```diff
@@ -257,3 +258,167 @@ class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
...
+  final bool value;
+  final ValueChanged<bool> onChanged;
+
+  @override
+  Widget build(BuildContext context) {
+    // Figma has this (blue/500) in both light and dark mode.
+    final activeColor = Color(0xff4370f0);
+
+    // Figma has this (grey/400) in both light and dark mode.
+    final inactiveColor = Color(0xff9194a3);
```
gnprice: Let's put TODO(#831) markers on these, so we can come back and wire them up to those values when we have them.

test/widgets/profile_test.dart
```diff
@@ -79,18 +92,24 @@ CustomProfileField mkCustomProfileField(
...
+    check(because: 'find user delivery email', find.text('testuser@example.com').evaluate()).isNotEmpty();
+  });
 
+  testWidgets('page builds; error page shows up if data is missing', (tester) async {
+    await setupPage(tester, pageUserId: eg.selfUser.userId + 1989);
+    check(because: 'find no user avatar', find.byType(Avatar).evaluate()).isEmpty();
+    check(because: 'find error icon', find.byIcon(Icons.error).evaluate()).isNotEmpty();
+  });
+
+  group('custom profile fields', () {
```
gnprice, line 137: This grouping makes sense.

There are a couple more test cases in here that aren't about custom profile fields, so should also get moved out:
* 'page builds; dm links to correct narrow'
* 'page builds; ensure long name does not overflow'

(To find those, I put my cursor on a reference to mkCustomProfileField so as to highlight the other references, and scrolled through.)

gnprice, line 137: nit:

    profile test [nfc]: Rename group about custom fields; pull out some tests

I'd describe this more like:
> profile test [nfc]: Split redundant ProfilePage group; make group for custom fields

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,216 @@
...
+  /// and the server may handle them out of order.
+  /// But the local echo minimizes flickering (see [localEchoMinimum])
+  /// while ensuring that the real in-store value is shown soon after the
+  /// user finishes interacting, whether the request(s) succeeded or failed.
+  void Function(T) handleRequestNewValue,
+);
+
+/// A stateful builder widget for toggles/etc.
+/// that control per-account settings on the server,
+/// with time-bounded local echo.
```
gnprice: Interesting! Seems useful.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,216 @@
...
+/// starts or extends a local-echo period.
+/// During local echo, [builder] is passed the new value
+/// instead of the value in the store.
+/// Local echo will continue for at least [localEchoMinimum]
+/// after the current call. After that, it may end
+/// - because the [findValueInStore] value changed after this call
+///   (i.e. the event arrived), or
+/// - because [sendValueToServer] failed, or
+/// - because [localEchoIdleTimeout] elapsed and there wasn't another call.
+class PerAccountSettingBuilder<T> extends StatefulWidget {
```
gnprice: How about `RemoteSettingBuilder`? I think the key for what makes this widget appropriate is less that the setting is per-account than that it lives on the server, and therefore requires a network request (which may be slow, and may fail) in order to set.

Potentially we could in the future have some settings that can be set separately per account even though they live on the client (in our local SQLite database). If we do, those wouldn't want to use this class; instead they'd work the same as our widgets for other local settings.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,216 @@
...
+import '../model/store.dart';
+import 'store.dart';
+
+typedef PerAccountSettingBuilderFn<T> = Widget Function(
+  /// The value, whether from [PerAccountSettingBuilder.findValueInStore]
+  /// or from local echo.
+  T value,
+
+  /// Calls [PerAccountSettingBuilder.sendValueToServer]
+  /// and starts or extends a local-echo period.
```
gnprice: These don't seem to work as dartdoc — they don't appear when I hover on anything relevant.

I think in general the way to document function parameters is in the function's doc; so here on the typedef.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,216 @@
...
+    _disposed = true;
+    super.dispose();
+  }
+}
+
+/// A local-echo value.
+///
+/// May be null, subject to [T].
+/// ("No local echo value" is represented by the absence of one of these.)
+class _LocalEchoValue<T> {
```
gnprice: Is this equivalent to what you mean?
```suggestion
/// A local-echo value.
///
/// The value represented may be null if [T] is nullable.
/// ("No local echo value" is represented by the absence of a [_LocalEchoValue].)
class _LocalEchoValue<T> {
```

I think the referent of "may be null", in particular, is worth spelling out more explicitly.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,216 @@
...
+    final value = switch (_notifier.value) {
+      _LocalEchoValue<T>(:final value) => value,
+      null => widget.findValueInStore(store),
+    };
+
+    return widget.builder(value, _handleRequestNewValue);
+  }
+}
+
+class _LocalEchoNotifier<T> extends ValueNotifier<_LocalEchoValue<T>?> {
```
gnprice: I think this would benefit from a short dartdoc: particularly on what it means to have a value here, or for [value] to be null.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,216 @@
...
+      _lowerBoundCompleter = null;
+    });
+
+    _upperBoundTimer?.cancel();
+    _upperBoundTimer = Timer(PerAccountSettingBuilder.localEchoIdleTimeout, () {
+      value = null;
+    });
+  }
+
+  Future<void> stop() async {
```
gnprice: Also good to document would be what this Future represents — what does it mean to await it?

lib/widgets/profile.dart
```diff
@@ -83,6 +87,28 @@ class ProfilePage extends StatelessWidget {
       // TODO(#196) render active status
       // TODO(#292) render user local time
 
+      if (!store.realmPresenceDisabled && userId == store.selfUserId) ...[
+        const SizedBox(height: 16),
+        MenuButtonsShape(buttons: [
```
gnprice: Let's pull this whole MenuButtonsShape subtree out to its own widget — e.g. call it `_InvisibleModeToggle` — so that this just has the condition, the spacers, and a reference to that widget. I think that'll help keep this more organized to read.

(The same might also be good for some of these other items; but this is more complex than any of the existing items here. And conversely it doesn't share interactions like oh they're all centered, they're all using _TextStyles.primaryFieldText, which are advantages to having the others inline next to each other.)

lib/widgets/profile.dart
```diff
@@ -83,6 +87,28 @@ class ProfilePage extends StatelessWidget {
...
+          // `value: true` means invisible mode is on,
+          // i.e., that presenceEnabled is false.
+          PerAccountSettingBuilder<bool>(
+            findValueInStore: (store) => !store.userSettings.presenceEnabled,
+            sendValueToServer: (value) => updateSettings(store.connection,
+              newSettings: {UserSettingName.presenceEnabled: !value}),
```
gnprice: Hmm, interesting that findValueInStore gets passed the store to use, but then sendValueToServer ends up using the store it got from the parent widget.

(This is fine to ship, doesn't need to be refactored, but feels like it suggests the API could ideally be adjusted)

lib/widgets/button.dart
```diff
@@ -257,3 +258,167 @@ class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
...
+            // hence the `assert` above.
+            padding: EdgeInsetsDirectional.only(start: itemSpacing - 12),
+
+            child: Row(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.center,
+              spacing: itemSpacing,
+              children: [
+                if (beforeIcon != null) beforeIcon!,
+                if (icon != null) Icon(icon, color: designVariables.contextMenuItemText),
```
gnprice: Interesting that this provides for both an icon and some other widget before it. Are there examples of that in the Figma designs?

gnprice (commented):
Thanks! Generally this looks great; comments below.

I haven't yet read all the logic in PerAccountSettingBuilder (pending a couple of questions / requests for dartdoc below), or the tests in the last commit. I've read everything else, so including everything before the last commit:
57da94352 api: Require InitialSnapshot.userSettings, present since Zulip Server 5
ebdb03526 api: Add InitialSnapshot.userSettings.presenceEnabled
f373c9417 api: Add updateSettings
4f114af89 button [nfc]: Extract MenuButton{,sShape} out to here, from action sheet
847b3267d button: Fix an inconsistency with Figma in MenuButton
3e496ec1e button: Give MenuButton a beforeIcon param, to use for a switch
f5598f5a8 button [nfc]: Factor MenuButton's vertical padding as around just the text
428e97e8b button: Implement "toggle" component from Figma, for "Invisible mode"
ed544a7c3 profile test [nfc]: Rename group about custom fields; pull out some tests

and part of that last commit:
a71d75d95 profile: Add toggle switch for "Invisible mode", with local echo

lib/widgets/per_account_settings.dart
chrisbobbe: How about I actually use the `Option` class you wrote for user-status #1629, instead of adding this new `_LocalEchoValue` class? That should work.

lib/widgets/button.dart
chrisbobbe: Yeah at the Figma link in the dartdoc 🙂:

https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60681&m=dev

image

lib/widgets/per_account_settings.dart
gnprice: SGTM!

lib/widgets/button.dart
gnprice: I see, indeed :smile:

chrisbobbe:

lib/widgets/button.dart
gnprice: The name "before icon" feels like it doesn't quite get to the point — I feel like I have to think a bit to work through what it's supposed to end up meaning. And in the current example, as the dartdoc says:
```dart
  /// An element to go before [icon], or in its place if it's null.
  ///
  /// E.g. a [Toggle].
  final Widget? beforeIcon;
```

this "before icon" widget appears even though there is no icon.

How about a name like "control"? It's the visible control in the button, when the button has such a thing.

Or perhaps with less generalization: "toggle"? That's the name it has in that Figma subtree: the children are "label", "toggle", "icon". If we later have a menu button somewhere that has some other kind of widget in here other than a toggle, I'm not sure we'll necessarily want to treat it the same way.

test/widgets/profile_test.dart
```diff
@@ -79,18 +91,50 @@ CustomProfileField mkCustomProfileField(
...
+    check(because: 'find user delivery email', find.text('testuser@example.com').evaluate()).isNotEmpty();
+  });
 
+  testWidgets('page builds; error page shows up if data is missing', (tester) async {
+    await setupPage(tester, pageUserId: eg.selfUser.userId + 1989);
+    check(because: 'find no user avatar', find.byType(Avatar).evaluate()).isEmpty();
+    check(because: 'find error icon', find.byIcon(Icons.error).evaluate()).isNotEmpty();
+  });
+
+  testWidgets('page builds; dm links to correct narrow', (tester) async {
```
gnprice, line 111: nit:

    Most of the tests in this ProfilePage group are about custom profile
    fields. Pull out the two that aren't, and rename the group.

s/two/handful/ or some other update :slightly_smiling_face: (after https://github.com/zulip/zulip-flutter/pull/1631#discussion_r2190923149)

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+/// and starts or extends a local-echo period.
+///
+/// If extending a local-echo period, replaces the old local-echo value.
+///
+/// This may be called any time. API requests are not debounced,
+/// and the server may handle them out of order.
+/// But the local echo minimizes flickering (see [localEchoMinimum])
+/// while ensuring that the real in-store value is shown soon after the
+/// user finishes interacting, whether the request(s) succeeded or failed.
+typedef PerAccountSettingBuilderFn<T> =
```
gnprice: nit: rename this to match the class; rename the file, too

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+///
+/// [value] is the value, whether from [RemoteSettingBuilder.findValueInStore]
+/// or from local echo.
+///
+/// [handleRequestNewValue] calls [RemoteSettingBuilder.sendValueToServer]
+/// and starts or extends a local-echo period.
+///
+/// If extending a local-echo period, replaces the old local-echo value.
+///
+/// This may be called any time. API requests are not debounced,
```
gnprice: "this" should be [handleRequestNewValue], right? Not the function whose type this is documenting.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /// Request that a local-echo session, if any, be stopped as soon as possible.
+  ///
+  /// The session will be stopped either immediately or
+  /// [RemoteSettingBuilder.localEchoMinimum] after the last [startOrExtend] call,
+  /// whichever is later.
+  ///
+  /// The returned [Future] resolves when the session is stopped.
+  Future<void> stop() async {
```
gnprice: Very helpful, thanks.

chrisbobbe:
And pushed again, with the "before icon" / "toggle" change suggestion https://github.com/zulip/zulip-flutter/pull/1631#discussion_r2191170091

gnprice (commented):
Thanks for the revisions! Just read the changes and re-scanned my comments above; all looks good except a few small comments below.

Next I'll read the remaining parts of the last/main commit, as mentioned at https://github.com/zulip/zulip-flutter/pull/1631#pullrequestreview-2995056594.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    // (Create the PerAccountStoreWidget dependency unconditionally.)
+    final store = PerAccountStoreWidget.of(context);
+
+    final value = switch (_notifier.value) {
+      OptionSome(:final value) => value,
+      OptionNone() => widget.findValueInStore(store),
+    };
```
gnprice: I guess this would be a use case for adding an `orElse`:
```suggestion
    final value = _notifier.value.orElse(() => widget.findValueInStore(store)); 
```

(taking API inspiration [from Rust's Option](https://doc.rust-lang.org/std/option/enum.Option.html#method.or_else), per the Option dartdoc)

This version is fine, though.

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+      if (widget.onError != null) {
+        widget.onError!(e, value);
+      }
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    // (Create the PerAccountStoreWidget dependency unconditionally.)
+    final store = PerAccountStoreWidget.of(context);
```
gnprice: Is there a reason we actively want to do this? That's what I take the comment to mean, but then I'm curious what the reason is :slightly_smiling_face: 

(I think it's harmless to do so — in fact we'll always have the dependency anyway before even reaching the build method, because we use it from didChangeDependencies)

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+
+  void _handleRequestNewValue(T value) async {
+    _notifier.startOrExtend(value);
+
+    try {
+      await widget.sendValueToServer(value);
+      if (_disposed) return;
+      // Don't call _notifier.stop(). We do that when the event arrives,
+      // causing the in-store value to change (see didChangeDependencies).
+    } catch (e) {
```
gnprice: This is potentially a real error, so let's leave a TODO(log) comment:
```suggestion
    } catch (e) { // TODO(log)
```

lib/widgets/profile.dart
```diff
@@ -109,6 +119,35 @@ class ProfilePage extends StatelessWidget {
...
+        builder: (value, handleRequestNewValue) => MenuButton(
+          label: zulipLocalizations.invisibleMode,
+          onPressed: () => handleRequestNewValue(!value),
+          toggle: Toggle(value: value, onChanged: handleRequestNewValue))),
+    ]);
+  }
+}
+
+
 class _ProfileErrorPage extends StatelessWidget {
```
gnprice: nit: double blank line

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+/// When it is [OptionSome], local echo is active with [OptionSome.value].
+/// When it is [OptionNone], local echo is not active.
+///
+/// Use [startOrExtend] and [stop] to control local echo.
+class _LocalEchoNotifier<T> extends ValueNotifier<Option<T>> {
+  factory _LocalEchoNotifier() {
+    return _LocalEchoNotifier._(OptionNone());
+  }
+
+  _LocalEchoNotifier._(super.value);
```
gnprice: nit: can simplify:
```suggestion
  _LocalEchoNotifier() : super(OptionNone());
```

lib/widgets/per_account_settings.dart
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  bool _disposed = false;
+
+  @override
+  void dispose() {
+    _lowerBoundTimer?.cancel();
+    _upperBoundTimer?.cancel();
+    _disposed = true;
+    super.dispose();
```
gnprice: Also `_lowerBoundCompleter?.complete()`?

Otherwise, I think if there's an outstanding `stop` call awaiting the completer's future, then that will hang around forever because the future never completes. Which might then keep a `_handleRequestNewValue` call hanging around, and therefore the state object and widget object, and so on.

test/widgets/profile_test.dart
```diff
@@ -365,4 +377,270 @@ void main() {
...
+    void checkAppears(WidgetTester tester) {
+      check(findRow).findsOne();
+      check(findToggle).findsOne();
+    }
+
+    bool getValue(WidgetTester tester) => tester.widget<Toggle>(findToggle).value;
+
+    void checkAppearsActive(WidgetTester tester, bool expected) {
+      final toggle = tester.widget<Toggle>(findToggle);
+      check(toggle.value).equals(expected);
```
gnprice: nit: this is equivalent, right?
```suggestion
      check(getValue(tester)).equals(expected);
```

test/widgets/profile_test.dart
```diff
@@ -365,4 +377,270 @@ void main() {
...
+        checkAppearsActive(tester, newValue);
+        checkRequest(newValue);
+      }
+
+      // Events will be coming in, but those don't control the switch;
+      // only the user interaction does, until there have been no interactions
+      // for [PerAccountSettingBuilder.localEchoMinimum].
+      await doSpamTap();
+      await tester.pump(Duration(milliseconds: 90));
+      await doSpamTap();
```
gnprice: It doesn't seem like these actually check that events don't control the switch.

In fact I tried deleting the key part of `stop`:
```diff
   Future<void> stop() async {
-    if (_lowerBoundCompleter != null) {
-      await _lowerBoundCompleter!.future;
-      if (_disposed) return;
-    }
     value = OptionNone();
   }
```

and it seems tests still pass.

test/widgets/profile_test.dart
```diff
@@ -365,4 +377,270 @@ void main() {
...
+      // The appearance changes and the request is sent, immediately.
+      prepareRequestError(Duration(milliseconds: 100));
+      await doToggle(tester, toggleInteractionMode);
+      await tester.pump();
+      await tester.pump();
+      checkAppearsActive(tester, true);
+      checkRequest(true);
+
+      // After [PerAccountSettingBuilder.localEchoMinimum], changes back
+      await tester.pump(RemoteSettingBuilder.localEchoMinimum);
```
gnprice: Similarly, this should gain a check that the appearance doesn't change *before* that minimum (when the request failure comes back).

gnprice (commented):
OK, and finished reading the rest. Generally this all looks good — comments below, most notably some checks missing in the tests.

test/widgets/profile_test.dart
chrisbobbe: Eep, indeed! Will fix.

lib/widgets/per_account_settings.dart
chrisbobbe: I think my thinking was: 

- `didChangeDependencies` needs to be called when local-echo isn't active, which means the dependency needs to exist when local-echo isn't active
- The (only) way to create a dependency is by calling `PerAccountStoreWidget.of` in the build method; it doesn't count if we call it in `didChangeDependencies`
- The dependency might disappear if a `build` call doesn't call `PerAccountStoreWidget.of` (I remember being pretty sure about this but can't right away find where I read about it)

I think I was indeed wrong on the second point though 🙂—

```dart
  /// This method should not be called from widget constructors or from
  /// [State.initState] methods, because those methods would not get called
  /// again if the inherited value were to change. To ensure that the widget
  /// correctly updates itself when the inherited value changes, only call this
  /// (directly or indirectly) from build methods, layout and paint callbacks,
  /// or from [State.didChangeDependencies] (which is called immediately after
  /// [State.initState]).
```

so yeah, the comment isn't guarding against a potential bug, and I'll remove it.

chrisbobbe:
Thanks! Revision pushed.

lib/widgets/per_account_settings.dart
gnprice: > * The dependency might disappear if a `build` call doesn't call `PerAccountStoreWidget.of` (I remember being pretty sure about this but can't right away find where I read about it)

There've been occasional requests upstream for this behavior. But the current behavior is that once an element has a dependency, it keeps it throughout its lifetime; there's no attempt to clear out dependencies.

See background here:
https://github.com/flutter/flutter/issues/12992#issuecomment-344358873

lib/basic.dart
```diff
@@ -0,0 +1,66 @@
...
+///   https://doc.rust-lang.org/std/option/enum.Option.html
+sealed class Option<T> {
+  const Option();
+
+  /// The value contained in this option, if any; else the given value.
+  T or(T optb);
+
+  /// The value contained in this option, if any;
+  /// else the value returned by [fn].
+  T orElse(T Function() fn);
```
gnprice: nit: specify the other key piece of this method's semantics:
```suggestion
  /// The value contained in this option, if any;
  /// else the value returned by [fn].
  ///
  /// [fn] is called only if its return value is needed.
  T orElse(T Function() fn);
```

gnprice: Oh and I guess test that too, e.g. by having the callback throw.

test/widgets/profile_test.dart
```diff
@@ -365,4 +377,277 @@ void main() {
...
+      await doToggle(tester, toggleInteractionMode);
+      await tester.pump();
+      await tester.pump();
+      checkAppearsActive(tester, true);
+      checkRequest(true);
+
+      // The appearance doesn't change as soon as the request errors,
+      // if it errored quickly…
+      await tester.pump(requestDuration);
+      checkAppearsActive(tester, true);
```
gnprice, line 576: We've seen elsewhere (those route transitions) that sometimes one needs to call `tester.pump` for a duration that's epsilon longer than nominally should be necessary in order to get something to happen.

So when a test is, like here, checking that something *doesn't* happen, it's not entirely convincing as a reader unless it waits for a bit longer than the nominal time, for good measure.

Could say `requestDuration + Duration(milliseconds: 1)`. Or could say like `Duration(milliseconds: 101)`, or `Duration(milliseconds: 200)`, and let the reader compare to the 100ms constant above.

lib/widgets/per_account_settings.dart
chrisbobbe: Good to know, thanks!

test/widgets/profile_test.dart
```diff
@@ -365,4 +377,277 @@ void main() {
...
+      await doSpamTap(expectedCurrentValue: true);
+      await tester.pump(Duration(milliseconds: 45));
+      await doSpamTap(expectedCurrentValue: false);
+      await tester.pump(Duration(milliseconds: 600));
+      await doSpamTap(expectedCurrentValue: true);
+      await tester.pump(Duration(milliseconds: 5));
+      await doSpamTap(expectedCurrentValue: false);
+
+      await tester.pump(RemoteSettingBuilder.localEchoMinimum - Duration(milliseconds: 1));
+      check(getValue(tester)).equals(true);
```
gnprice: nit: I think a bit more explicitness would help understanding here:
```suggestion
      await doSpamTap(expectedCurrentValue: false);
      check(getValue(tester)).equals(true);

      await tester.pump(RemoteSettingBuilder.localEchoMinimum - Duration(milliseconds: 1));
      check(getValue(tester)).equals(true);
```

(That is the right expectation, right?)

alya:
I left a design comment on CZO: [#mobile-design > presence: invisible mode @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/presence.3A.20invisible.20mode/near/2216128)

gnprice:
Copying those comments here (edited a bit):

> Looking at the screnshots in [#1631](https://github.com/zulip/zulip/pull/1631), I think having brand color on that row adds too much emphasis in this context. (Most users are expected to never go invisible at all.)
>
> I think neutral styling (like what we have in [the main menu]?) would work better.

gnprice:
I'm planning to include this PR in a v30.0.261 release tomorrow. It'd be great to have it merged first, and to have that design tweak made.

chrisbobbe:
Thanks for the reviews and design guidance! Revision pushed, with three new commits:

60a3e2b8d button: Fix MenuButton's icon color
84181e8bd button [nfc]: Add "list button" variant, factored into MenuButton widget
36d973515 profile: Change "Invisible mode" toggle to more neutral "list button" style

The first of those fixes a bug in `main` where we had the wrong icon color in action sheets, only in light mode:

| Before | After |
| --- | --- |
| image | image |

New screenshots:

| Switch on | Switch off |
| --- | --- |
| image | image |
| image | image |

test/widgets/profile_test.dart
chrisbobbe: Yep!

chrisbobbe:
Thanks both! Revision pushed.

lib/widgets/button.dart
```diff
@@ -257,3 +258,229 @@ class _AnimatedScaleOnTapState extends State<AnimatedScaleOnTap> {
...
+    ZulipMenuItemButtonStyle.list => 12,
+  };
+
+  static bool _debugCheckShapeAncestor(BuildContext context) {
+    final ancestor = context.findAncestorWidgetOfExactType<MenuButtonsShape>();
+    assert(() {
+      if (ancestor != null) return true;
+      throw FlutterError.fromParts([
+        ErrorSummary('No MenuButtonsShape ancestor found.'),
+        ErrorDescription('MenuButton widgets require a MenuButtonsShape ancestor.'),
```
gnprice: nit: one stray reference here to the MenuButton name

lib/widgets/theme.dart
```diff
@@ -154,6 +154,7 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     composeBoxBg: const Color(0xffffffff),
     contextMenuCancelText: const Color(0xff222222),
     contextMenuItemBg: const Color(0xff6159e1),
+    contextMenuItemIcon: const Color(0xff4f42c9),
     contextMenuItemLabel: const Color(0xff242631),
     contextMenuItemMeta: const Color(0xff626573),
     contextMenuItemText: const Color(0xff381da7),
```
gnprice, line 160: button: Fix MenuButton's icon color
    
    Oops, this wasn't following the Figma. Fixed.

Hmm good catch.

I guess this affects the icons in the action sheets? Is that the existing place this affects?

Looks like the difference is pretty subtle, thankfully. Here's a quick preview, from my IDE showing color swatches in the left gutter next to the source code:
image

Bringing those right next to each other (new above, old below):
image

chrisbobbe: Yep yep, and I posted a screenshot of the change in the action sheet: https://github.com/zulip/zulip-flutter/pull/1631#issuecomment-3051130010

chrisbobbe:
Thanks! Revision pushed.

lib/widgets/theme.dart
gnprice: Ah indeed, thanks :-)

gnprice:
Thanks! Looks good; merging.

PR #1633, opened 2025-06-26
content: Handle 'nulldelimiter' KaTeX CSS

rajveermalviya:
Stacked on top of #1609.
Related: #46


gnprice:
Thanks for building this! See my comments on #1670 for a few things I'd like to see on all the open KaTeX PRs: https://github.com/zulip/zulip-flutter/pull/1670#pullrequestreview-3045025908, https://github.com/zulip/zulip-flutter/pull/1670#issuecomment-3105055606 .

test/model/katex_test.dart
```diff
@@ -643,6 +643,38 @@ class KatexExample extends ContentExample {
           text: '∗'),
       ]),
     ]);
+
+  static final nulldelimiter = KatexExample.block(
+    r'\{left,middle,right}. : spans with "nulldelimiter" class',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2205534
+    r'\left. a \middle. b \right.',
```
gnprice, line 650: Huh interesting — I didn't know you could use `.` that way in TeX :slightly_smiling_face:

test/model/katex_test.dart
```diff
@@ -643,6 +643,38 @@ class KatexExample extends ContentExample {
           text: '∗'),
       ]),
     ]);
+
+  static final nulldelimiter = KatexExample.block(
+    r'\{left,middle,right}. : spans with "nulldelimiter" class',
```
gnprice: Using curly braces for this shell-like syntax here in a TeX formula is confusing, because braces have their own quite different meaning in TeX.

It looks like "null delimiter" is the TeX term for what's going on here — search for that term in the TeXbook:
https://visualmatheditor.equatheque.net/doc/texbook.pdf

So we can use that term in the test description:
```suggestion
    r'null delimiters, like `\left.`',
```

test/model/katex_test.dart
```diff
@@ -643,6 +643,38 @@ class KatexExample extends ContentExample {
...
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2205534
+    r'\left. a \middle. b \right.',
+    '<p>'
+      '<span class="katex-display"><span class="katex">'
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mi>a</mi><mo fence="true" lspace="0.05em" rspace="0.05em">.</mo><mi>b</mi></mrow><annotation encoding="application/x-tex">\\left. a \\middle. b \\right.</annotation></semantics></math></span>'
+      '<span class="katex-html" aria-hidden="true">'
+        '<span class="base">'
+          '<span class="strut" style="height:0.6944em;"></span>'
+            '<span class="minner">'
```
gnprice: nit: indentation; the .minner is a sibling, not child, of the .strut
```suggestion
        '<span class="base">'
          '<span class="strut" style="height:0.6944em;"></span>'
          '<span class="minner">'
```

test/model/katex_test.dart
```diff
@@ -643,6 +643,38 @@ class KatexExample extends ContentExample {
...
+              '<span class="mopen nulldelimiter"></span>'
+              '<span class="mord mathnormal">a</span>'
+              '<span class="nulldelimiter"></span>'
+              '<span class="mord mathnormal">b</span>'
+              '<span class="mclose nulldelimiter"></span></span></span></span></span></span></p>', [
+      KatexSpanNode(nodes: [
+        KatexStrutNode(heightEm: 0.6944, verticalAlignEm: null),
+        KatexSpanNode(nodes: [
+          KatexSpanNode(styles: KatexSpanStyles(widthEm: 0.12), nodes: []),
```
gnprice, line 666: nit: simplify
```suggestion
          KatexSpanNode(styles: KatexSpanStyles(widthEm: 0.12)),
```

gnprice (commented):
Thanks! Generally this all looks good; just nits below.

> No change in total failed messages, so most uses in the wild also depend on some other feature currently unimplemented (probably fractions).

Ah well. I think this is a prerequisite for those, though (#1678), and fractions are certainly a useful feature.

test/model/katex_test.dart
rajveermalviya: Hmm we need to specify an empty list here, because we currently require either `text` or `nodes` to be non-null in `KatexSpanNode`.

rajveermalviya:

test/model/katex_test.dart
gnprice: Ah I see, indeed. OK.

gnprice:
Thanks! Looks good; merging.

Issue #1637, opened 2025-06-26
msglist: Show recipient headers on all messages, in Mentions / Starred


chrisbobbe:
[#mobile > search @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/search/near/2199084)
> Every message [in search] should show a recipient header, even if it's to the same conversation as the one before it (because there might be other non-matching messages in between those two, so we shouldn't give the impression that one was immediately after the other).
This just occurred to me in thinking about search, but I think it should also apply to the starred-messages and mentions views, and is a bug we currently have in those. Not hard to fix.

Related: #252

PR #1638, opened 2025-06-26
msglist: Show recipient headers on all messages, in Mentions / Starred

chrisbobbe:
Fixes: #1637

| Before | After |
| --- | --- |
| image | image |
| image | image |

cc @alya

### DIFF
--- a/lib/model/message_list.dart
+++ b/lib/model/message_list.dart
@@ -105,6 +105,18 @@ enum FetchingStatus {
 ///
 /// This comprises much of the guts of [MessageListView].
 mixin _MessageSequence {
+  /// Whether each message should have its own recipient header,
+  /// even if it's in the same conversation as the previous message.
+  ///
+  /// In some message-list views, notably "Mentions" and "Starred",
+  /// it would be misleading to give the impression that consecutive messages
+  /// in the same conversation were sent one after the other
+  /// with no other messages in between.
+  /// By giving each message its own recipient header (a `true` value for this),
+  /// we intend to avoid giving that impression.
+  @visibleForTesting
+  bool get oneMessagePerBlock;
+
   /// A sequence number for invalidating stale fetches.
   int generation = 0;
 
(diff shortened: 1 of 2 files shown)


alya:
The screenshots look great, thanks!

lib/model/message_list.dart
```diff
@@ -613,6 +619,15 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Anchor get anchor => _anchor;
   Anchor _anchor;
 
+  @override bool get oneMessagePerBlock => switch (narrow) {
+    CombinedFeedNarrow()
+      || ChannelNarrow()
+      || TopicNarrow()
+      || DmNarrow() => false,
+    MentionsNarrow()
+      || StarredMessagesNarrow() => true,
```
gnprice: This logic should get some dartdoc, to guide people adding cases to it in the future.

lib/model/message_list.dart
```diff
@@ -613,6 +619,15 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Anchor get anchor => _anchor;
   Anchor _anchor;
 
+  @override bool get oneMessagePerBlock => switch (narrow) {
```
gnprice: nit: this feels to me like it belongs after the lifecycle methods, next to _messageVisible and friends

lib/model/message_list.dart
gnprice: I'd like to have a test that would fail if this started being all true or all false. Maybe a quick test case for a channel narrow, and another for the mentions narrow, showing different results with the same two messages.

lib/model/message_list.dart
```diff
@@ -105,6 +105,8 @@ enum FetchingStatus {
 ///
 /// This comprises much of the guts of [MessageListView].
 mixin _MessageSequence {
+  bool get oneMessagePerBlock;
```
gnprice, line 118: Conceptually I think this should be private. It is convenient to use it in checkInvariants; but then let's mark it `@visibleForTesting`.

gnprice (commented):
Thanks for taking care of this! Comments below.

chrisbobbe:

test/model/message_list_test.dart
```diff
@@ -2742,6 +2742,53 @@ void main() {
     checkNotifiedOnce();
   }));
 
+  group('one message per block?', () {
+    final channelId = 1;
+    final topic = 'some topic';
+    void doTest({required Narrow narrow, required bool expected}) {
+      test('$narrow: ${expected ? 'yes' : 'no'}', () => awaitFakeAsync((async) async {
+        final sender = eg.user();
+        final channel = eg.stream(streamId: 1);
```
gnprice: nit: implicitly relying on these stream/channel IDs being equal, IIUC; instead mention the variable

gnprice (commented):
Thanks! All looks good modulo one small comment below; then go ahead and merge.

chrisbobbe:
Thanks! Done.

Issue #1621, opened 2025-06-24
msglist: When tapping a message in starred/mentions/search, open conversation view at that message


chrisbobbe:
This will be helpful for:

- #252 

and should be easy to do.

P-yiush07:
heyy can i work on this ?

chrisbobbe:
Updated the description following discussion: [#mobile-team > msglist: anchor when tapping recipient header @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/msglist.3A.20anchor.20when.20tapping.20recipient.20header/near/2211469)

Issue #252, opened 2023-08-03
msglist: Offer search narrows


gnprice:
This is a bit more complex than other narrows (#72, #250, #251) for a couple of reasons:
* It calls for some UI of its own, for the user to enter the search they want.
* The messages themselves should have a bit more data, to highlight the text within them that matched the query. This comes from the server as `match_content` and `match_subject` fields on the message objects; see https://zulip.com/api/get-messages .

For a first version, the search UI can be at the same level as in zulip-mobile, which means fairly crude: it doesn't support anything more than a text search. Ideally we'd support the richer range of searches that Zulip supports in general, complete with UI for the various kinds of filters, but that can be a followup issue.

The UI also calls for some care with concurrency and asynchronicity. See zulip-mobile's implementation.

chrisbobbe:
Probably as a followup, we should make it easy to [search shared history](https://zulip.com/help/search-for-messages#searching-shared-history) in addition to searching personal history. The zulip-flutter equivalent of zulip/zulip-mobile#5732.

gnprice:
A user [writes](https://chat.zulip.org/#narrow/channel/48-mobile/topic/beta.20feedback.3A.20Jonas.20Suni/near/2009768) last week:
> I don't use search that often, and mostly on Desktop, but when I do use it I really do need it. So I really wish it will continue to be available in the future, even if it's not working perfectly.

chrisbobbe:
From a review in the Google Play Console:

> It is nice, but the absence of a search bar (at least to search for topics) makes it frustrating to use.

chrisbobbe:
This came up in beta feedback: [#mobile > Flutter Beta: Some Feedback @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20Beta.3A.20Some.20Feedback/near/2153404)

jyn514:
i find search extremely useful and miss it a lot on the mobile app, it means i have to use zulip through a mobile browser instead some of the time.

gnprice:
Thanks @jyn514 for the feedback!

The legacy mobile app has a search feature but with limitations: it only supports searching for a text query, not limiting by channel or sender or any of Zulip's other search filters. Do you (or would you) similarly miss search when using the legacy app, or would a search feature with those limitations cover many of the times you use search?

jyn514:
[...]

PR #1639, opened 2025-06-26
msglist: On tap recipient header, open msglist anchored at message

chrisbobbe:
Fixes: #1621
Related: #252

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1437,7 +1437,8 @@ class StreamMessageRecipientHeader extends StatelessWidget {
       streamWidget = GestureDetector(
         onTap: () => Navigator.push(context,
           MessageListPage.buildRoute(context: context,
-            narrow: ChannelNarrow(streamId))),
+            narrow: ChannelNarrow(streamId),
+            initAnchorMessageId: message.id)),
         onLongPress: () => showChannelActionSheet(context, channelId: streamId),
         child: Row(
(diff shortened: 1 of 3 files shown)


gnprice:
Looking at this, I'm realizing I don't want this behavior when coming from the combined feed or a channel feed (two things I regularly do). In my common workflows, I'm catching up by scrolling those near the end; I see there's some conversation that looks relevant for me; I want to see the conversation from the beginning, or the beginning of what I haven't previously read. So first-unread is exactly what I want.

We had a chat thread where this behavior came up in the context of search, right? I'm not finding it at the moment — it looks like #1621 didn't get mentioned there. Let's discuss in that thread what the logic should be for this.

chrisbobbe:
I don't think there's a chat thread, there's just what I wrote on the issue:

> This will be helpful for:
> 
> - #252 
> 
> and should be easy to do.

So I've just started one: [#mobile-team > msglist: anchor when tapping recipient header @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/msglist.3A.20anchor.20when.20tapping.20recipient.20header/near/2210354)

gnprice:
Sounds good. I ran across this comment today: https://github.com/zulip/zulip-flutter/issues/252#issuecomment-2968215228 and I think it might have been what I was thinking of.

chrisbobbe:
Closing in favor of #1657 (but I've kept the add-a-missing-test commit).

PR #1646, opened 2025-07-01
content: Inline audio preview

### LINKED ISSUE #1665: Handle inline audio previews
This is an upcoming Zulip feature.

This issue is for a basic version that just shows a link the user can use to get to the file on the web. A follow-up issue will be for showing an actual audio player.

### References
* [#api design > audio previews @ 💬](https://chat.zulip.org/#narrow/channel/378-api-design/topic/audio.20previews/near/2143883)
* Server PR: zulip/zulip#34903

rajveermalviya:
Fixes #1665.

The first commit is ready, which just displays the source URL of `<audio>` element. Later commit adds the actual inline audio player which currently is minimal implementation using [`package:just_audio`](https://pub.dev/packages/just_audio), as I was unsure about the required behaviour and design.

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -1087,6 +1087,22 @@ class _ZulipInlineContentParser {
       return GlobalTimeNode(datetime: datetime, debugHtmlNode: debugHtmlNode);
     }
 
+    if (localName == 'audio' && className.isEmpty) {
+      final srcAttr = element.attributes['src'];
+      if (srcAttr == null) return unimplemented();
+
+      final String title = switch (element.attributes) {
+        {'title': final titleAttr} => titleAttr,
+        _ => Uri.tryParse(srcAttr)?.pathSegments.lastOrNull ?? srcAttr,
+      };
+
+      final link = LinkNode(
+        url: srcAttr,
+        nodes: [TextNode(title)]);
+      (_linkNodes ??= []).add(link);
+      return link;
+    }
+
     if (localName == 'span' && className == 'katex') {
       return parseInlineMath(element) ?? unimplemented();
     }
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -1327,6 +1327,24 @@ class ContentExample {
     InlineVideoNode(srcUrl: '/user_uploads/2/78/_KoRecCHZTFrVtyTKCkIh5Hq/Big-Buck-Bunny.webm'),
   ]);
 
+  static const audioInline = ContentExample(
+    'audio inline',
+    '![crab-rave.mp3](/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3)',
+    '<p><audio controls preload="metadata" src="/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3" title="crab-rave.mp3"></audio></p>', [
+    ParagraphNode(links: null, nodes: [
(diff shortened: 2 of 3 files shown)


rajveermalviya:
Screenshots:
| Web (Firefox) | First commit | Final commit |
| - | - | - |
|  image | image | image |

test/model/content_test.dart
```diff
@@ -1327,6 +1327,24 @@ class ContentExample {
...
+      LinkNode(url: '/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3', nodes: [TextNode('crab-rave.mp3')]),
+    ]),
+  ]);
+
+  static const audioInlineNoTitle = ContentExample(
+    'audio inline no title',
+    '![](/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3)',
+    '<p><audio controls preload="metadata" src="/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3"></audio></p>', [
+    ParagraphNode(links: null, nodes: [
+      LinkNode(url: '/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3', nodes: [TextNode('/user_uploads/2/f2/a_WnijOXIeRnI6OSxo9F6gZM/crab-rave.mp3')]),
```
gnprice: When there's no title, let's display the basename of the "src" URL's path, rather than the whole thing. That means the portion after the last slash.

That way this sort of example will be rather more readable.

lib/model/content.dart
```diff
@@ -1072,6 +1072,19 @@ class _ZulipInlineContentParser {
...
 
+    if (localName == 'audio' && className.isEmpty) {
+      final srcAttr = element.attributes['src'];
+      if (srcAttr == null) return unimplemented();
+
+      final titleAttr = element.attributes['title'];
+
+      final link = LinkNode(
+        url: srcAttr,
+        nodes: [TextNode(titleAttr ?? srcAttr)]);
```
gnprice: It feels a bit messy to be transforming this into a plain link, the same thing that an actual link would have been parsed as. Cleaner would be to give it its own node type, and handle this transform at build time.

But this is fine for getting something quickly that we can ship. The cleaner thing probably requires a bit of refactoring to make it cooperate with LinkNode and that infrastructure.

gnprice (commented):
Thanks! Comments below on the first commit.

rajveermalviya:

gnprice:
Thanks! The first commit looks good.

I haven't yet read the second commit in detail, but I'd like to ship the first commit in a release tomorrow ([#api design > audio previews @ 💬](https://chat.zulip.org/#narrow/channel/378-api-design/topic/audio.20previews/near/2211992)). So let's split the two commits into two separate PRs.

gnprice:
Splitting this PR to contain just its first commit:
e944915bd content: Initial support for inline &lt;audio>, showing source links

Then I'll merge.

@rajveermalviya please go ahead and send a separate draft PR with the second commit:
af7cae9af wip: Initial AudioPlayer using just_audio, rendered inline along with content

Even though you have other priorities before refining that further, having it around as a draft PR will be good just to help us not lose track of it.

gnprice:
OK, merged.

I also filed two issues, to track what this fixes already:
* #1665 

(and edited the commit and PR to point to that)

and what the follow-up PR will fix:
* #1666

PR #1640, opened 2025-06-27
msglist: Friendlier placeholder text when narrow has no messages

chrisbobbe:
msglist: Friendlier placeholder text when narrow has no messages
    
    Fixes #1555.
    
    A "no-messages-here" placeholder will be important for search, #252.
    Might as well implement it properly for the various possible cases,
    using web as a guide; see web/src/narrow_banner.ts in zulip/zulip.

### Screenshots

Screenshot 2025-06-26 at 5 28 34 PM

Starred messages: "Learn more" button opens `/help/star-a-message`
Screenshot 2025-06-26 at 5 32 00 PM

Screenshot 2025-06-26 at 5 31 52 PM

Group DM:
Screenshot 2025-06-26 at 5 31 32 PM

Group DM with a deactivated user:
Screenshot 2025-06-26 at 9 52 33 PM

Screenshot 2025-06-26 at 5 30 27 PM

DM with muted user:
Screenshot 2025-06-26 at 9 48 02 PM

DM with deactivated user:
Screenshot 2025-06-26 at 9 51 07 PM

Screenshot 2025-06-26 at 5 30 11 PM
Screenshot 2025-06-26 at 5 29 57 PM
Screenshot 2025-06-26 at 5 28 46 PM

Unknown channel (unknown because I'm not allowed to know about it):
Screenshot 2025-06-26 at 9 43 11 PM

Private channel I'm not subscribed to: "Learn more" button goes to `/help/channel-permissions#private-channels`
(The error banner replacing the compose box is also new in this PR. It's the same as the one above, for unknown channels, which is in `main`.)
Screenshot 2025-06-26 at 9 36 12 PM

PR #1650, opened 2025-07-02
msglist: Friendlier placeholder text when narrow has no messages (simple version)

### LINKED ISSUE #1555: msglist: Friendlier placeholder text when narrow has no messages
When a message list has no messages, we currently just show "No earlier messages". Screenshots for the narrows currently supported in the message-list page:

| Combined feed | Channel | Topic | DM (1:1) | DM (self) | DM (group) | Starred messages | Mentions |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Image | Image | Image | Image | Image | Image | Image | Image | Image |

This issue is to choose and display a nicer message. It might be different for each type of narrow.

cc @alya

chrisbobbe:
Splitting this out from #1640, so it can be merged soon.

Fixes #1555.

For now, the text simply says "There are no messages here." We'll add per-narrow logic later, but this is an improvement over the current appearance which just says "No earlier messages." (Earlier than what?)

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -859,7 +859,14 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
 
   @override
   Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
     if (!model.fetched) return const Center(child: CircularProgressIndicator());
+
+    if (model.items.isEmpty && model.haveNewest && model.haveOldest) {
+      return PageBodyEmptyContentPlaceholder(
+        message: zulipLocalizations.emptyMessageList);
+    }
 
     // Pad the left and right insets, for small devices in landscape.
     return SafeArea(
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -349,6 +349,36 @@ void main() {
     });
   });
 
+  group('no-messages placeholder', () {
+    final findPlaceholder = find.byType(PageBodyEmptyContentPlaceholder);
+
+    testWidgets('Combined feed', (tester) async {
+      await setupMessageListPage(tester, narrow: CombinedFeedNarrow(), messages: []);
+      check(
+        find.descendant(
+          of: findPlaceholder,
+          matching: find.textContaining('There are no messages here.')),
+      ).findsOne();
+    });
+
+    testWidgets('when `messages` empty but `outboxMessages` not empty, show outboxes, not placeholder', (tester) async {
+      final channel = eg.stream();
+      await setupMessageListPage(tester,
+        narrow: TopicNarrow(channel.streamId, eg.t('topic')),
+        streams: [channel],
+        messages: []);
+      check(findPlaceholder).findsOne();
+
+      connection.prepare(json: SendMessageResult(id: 1).toJson());
+      await tester.enterText(contentInputFinder, 'asdfjkl;');
+      await tester.tap(find.byIcon(ZulipIcons.send));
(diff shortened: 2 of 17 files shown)


chrisbobbe:
| Before | After |
| --- | --- |
| image | image |

lib/widgets/home.dart
```diff
@@ -148,40 +148,6 @@ class _HomePageState extends State<HomePage> {
   }
```
rajveermalviya, line 148: > page: Move no-content placeholder widget to page.dart, from home.dart 
 
 Should this commit be marked as nfc?

rajveermalviya (approved):
Thanks @chrisbobbe! Just one small comment, otherwise LGTM.

chrisbobbe:
Thanks! Revision pushed, and I'll mark for Greg's review.

chrisbobbe:
(Updated just to add a test that I wrote for #1640 but also belongs here.)

test/widgets/message_list_test.dart
```diff
@@ -326,6 +329,36 @@ void main() {
...
+    testWidgets('Combined feed', (tester) async {
+      await setupMessageListPage(tester, narrow: CombinedFeedNarrow(), messages: []);
+      check(
+        find.descendant(
+          of: findPlaceholder,
+          matching: find.textContaining('There are no messages here.')),
+      ).findsOne();
+    });
+
+    testWidgets('when `messages` empty but `outboxMessages` not empty, show outboxes, not placeholder', (tester) async {
```
gnprice, line 364: Yeah, good thought to test this.

test/widgets/message_list_test.dart
```diff
@@ -326,6 +329,36 @@ void main() {
...
+        messages: []);
+      check(findPlaceholder).findsOne();
+
+      connection.prepare(json: SendMessageResult(id: 1).toJson());
+      await tester.enterText(contentInputFinder, 'asdfjkl;');
+      await tester.tap(find.byIcon(ZulipIcons.send));
+      await tester.pump(kLocalEchoDebounceDuration);
+
+      check(findPlaceholder).findsNothing();
+      check(find.text('asdfjkl;')).findsOne();
```
gnprice, line 378: And this is a nice user-focused structure for the test.

lib/widgets/message_list.dart
```diff
@@ -788,8 +788,16 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
 
   @override
   Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
     if (!model.fetched) return const Center(child: CircularProgressIndicator());
 
+    if (model.haveNewest && model.haveOldest
+          && model.messages.isEmpty && model.outboxMessages.isEmpty) {
```
gnprice: Maybe test model.items is empty, rather than messages and outboxMessages separately?
```suggestion
    if (model.items.isEmpty && model.haveNewest && model.haveOldest) {
```

That's the layer the rest of this widget's logic works at: it operates on items and doesn't look at the messages or outboxMessages lists directly.

If in some future we add another way there can be items in the list, with neither messages nor outbox-messages, and don't edit this, then that would mean it'd show the start and end caps, and the items in between. I think that's probably the right outcome: the items are there in the view-model, so should be shown. (If they shouldn't be shown, they shouldn't be in the view-model.)

gnprice (commented):
Thanks! Looks good; one small comment.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merging.

PR #1652, opened 2025-07-02
compose: Don't show compose box for private, unsubscribed channels

### LINKED ISSUE #1651: compose: Hide compose box for private, unsubscribed channels
Currently, if you view a private channel you're not subscribed to, we show the compose box and let you tap "Send", which just gives you a server error:

Instead, we should hide the compose box, like we do for other narrows you're not allowed to post in.

chrisbobbe:
I noticed this while working on #1555.

Fixes #1651.

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -2064,8 +2064,12 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
       case ChannelNarrow(:final streamId):
       case TopicNarrow(:final streamId):
         final channel = store.streams[streamId];
-        if (channel == null || !store.hasPostingPermission(inChannel: channel,
-            user: store.selfUser, byDate: DateTime.now())) {
+        if (
+          channel == null
+          || (channel.inviteOnly && channel is! Subscription)
+          || !store.hasPostingPermission(inChannel: channel,
+               user: store.selfUser, byDate: DateTime.now())
+        ) {
           return _ErrorBanner(getLabel: (zulipLocalizations) =>
             zulipLocalizations.errorBannerCannotPostInChannelLabel);
         }
--- a/test/widgets/compose_box_test.dart
+++ b/test/widgets/compose_box_test.dart
@@ -58,6 +58,7 @@ void main() {
     User? selfUser,
     List<User> otherUsers = const [],
     List<ZulipStream> streams = const [],
+    List<Subscription> subscriptions = const [],
     List<Message>? messages,
     bool? mandatoryTopics,
     int? zulipFeatureLevel,
@@ -81,6 +82,7 @@ void main() {
     await testBinding.globalStore.add(selfAccount, eg.initialSnapshot(
       realmUsers: [selfUser, ...otherUsers],
       streams: streams,
+      subscriptions: subscriptions,
       zulipFeatureLevel: zulipFeatureLevel,
       realmMandatoryTopics: mandatoryTopics,
       realmAllowMessageEditing: true,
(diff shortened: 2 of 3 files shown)


rajveermalviya (approved):
Thanks @chrisbobbe! LGTM.

lib/widgets/compose_box.dart
```diff
@@ -2064,8 +2064,12 @@ class _ComposeBoxState extends State<ComposeBox> with PerAccountStoreAwareStateM
       case ChannelNarrow(:final streamId):
       case TopicNarrow(:final streamId):
         final channel = store.streams[streamId];
-        if (channel == null || !store.hasPostingPermission(inChannel: channel,
-            user: store.selfUser, byDate: DateTime.now())) {
+        if (
+          channel == null
+          || (channel.inviteOnly && channel is! Subscription)
```
gnprice, line 2069: Huh interesting. Is this the right logic?

You could be viewing a channel that's private, you're not subscribed, but you have content access to (now that that's a thing). In that case I'd expect the rule for whether you can post to be the same as it is for a public channel.

gnprice:
(Propagated the milestone from the issue.)

lib/widgets/compose_box.dart
chrisbobbe: > you have content access to (now that that's a thing)

Are you thinking of shared vs. protected history, or is there a newer feature than that that I'm not aware of?

gnprice: This one: https://zulip.com/help/configure-who-can-subscribe

> This permission grants access to channel content: users who are allowed to subscribe to a channel will also be able to read messages in it without subscribing.

It looks like in order to send, one then has to actually subscribe:
> A designer on the team could then follow a [link to a conversation](https://zulip.com/help/link-to-a-message-or-conversation#link-to-a-topic-within-zulip) in the private engineering channel, and read it without subscribing. They could subscribe if they need to send a message there, without asking for help.

Seems like a wart that that's different from the behavior for public channels, though. And ISTR a recent chat thread where either Alya or Tim said that behavior for public channels was likely to change.

I think the upshot is:
* this logic should live somewhere more central, in model code; probably within the hasPostingPermission method (which in turn should ultimately be on ChannelStore, after other refactors);
* the public-vs-private quirk should get a comment pointing to the help page and/or to a relevant chat thread;
* let's finish taking care of search and some other priorities before returning to finish this :slightly_smiling_face:

alya: https://github.com/zulip/zulip/issues/21433 has some work by a current GSoC participant.

chrisbobbe: > It looks like in order to send, one then has to actually subscribe:
> 
> > A designer on the team could then follow a [link to a conversation](https://zulip.com/help/link-to-a-message-or-conversation#link-to-a-topic-within-zulip) in the private engineering channel, and read it without subscribing. They could subscribe if they need to send a message there, without asking for help.
> 
> Seems like a wart that that's different from the behavior for public channels, though.

I think that note in the doc makes sense for frontend users, because frontend hasn't implemented sending at all when unsubscribed, even when you have permission: https://github.com/zulip/zulip/issues/35902. (That issue supersedes the one Alya linked to here.)

I don't think the server actually prohibits sending to all unsubscribed private channels though. There's no note about that in the ["Channel posting policy" Help Center doc](https://zulip.com/help/channel-posting-policy); I read that doc as saying that `can_send_message_group` is the only thing that controls whether you can send messages in a channel.

chrisbobbe:
Indeed, I didn't have the right logic in mind when writing this, but I think we do now have the right logic, implemented in #1863 (see https://github.com/zulip/zulip-flutter/pull/1652#discussion_r2418187393), so I'll close this.

lib/widgets/compose_box.dart
chrisbobbe: As of #1863 being merged, we do now respect `can_send_message_group`, when present, so I think there's nothing more to do here and I'll close this.

PR #1657, opened 2025-07-02
msglist: Tapping a message in Starred or Mentions opens anchored msglist

### LINKED ISSUE #1621: msglist: When tapping a message in starred/mentions/search, open conversation view at that message

chrisbobbe:
Fixes #1621.

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_color_models/flutter_color_models.dart';
@@ -989,11 +991,15 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
         final header = RecipientHeader(message: data.message, narrow: widget.narrow);
         return MessageItem(
           key: ValueKey(data.message.id),
+          narrow: widget.narrow,
           header: header,
           item: data);
       case MessageListOutboxMessageItem():
         final header = RecipientHeader(message: data.message, narrow: widget.narrow);
-        return MessageItem(header: header, item: data);
+        return MessageItem(
+          narrow: widget.narrow,
+          header: header,
+          item: data);
     }
   }
 }
@@ -1315,10 +1321,12 @@ class DateSeparator extends StatelessWidget {
 class MessageItem extends StatelessWidget {
   const MessageItem({
     super.key,
+    required this.narrow,
     required this.item,
     required this.header,
   });
 
+  final Narrow narrow;
   final MessageListMessageBaseItem item;
   final Widget header;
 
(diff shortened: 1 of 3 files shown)


chrisbobbe:
(fixing CI; just an unused import.)

lib/widgets/message_list.dart
```diff
@@ -1331,7 +1351,10 @@ class MessageItem extends StatelessWidget {
       color: designVariables.bgMessageRegular,
       child: Column(children: [
         switch (item) {
-          MessageListMessageItem() => MessageWithPossibleSender(item: item),
+          MessageListMessageItem() => MessageWithPossibleSender(
+            item: item,
+            tapOpensConversation: tapOpensConversation,
+          ),
```
gnprice: nit: `item` is the payload, `tapOpensConversation` is metadata, so keep the payload last in the widget constructor's parameter list

lib/widgets/message_list.dart
```diff
@@ -974,6 +976,16 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
     }
   }
 
+  /// Whether tapping a message should open its conversation.
+  bool get _tapMessageOpensConversation => switch (widget.narrow) {
+    CombinedFeedNarrow()
+      || ChannelNarrow()
+      || TopicNarrow()
+      || DmNarrow() => false,
```
gnprice: nit: I think it's probably cleaner to have this logic live next to where it gets used, on MessageWithPossibleSender; so instead of passing down the bool this returns, we'd pass down the Narrow itself. Just like we do with RecipientHeader now.

test/widgets/message_list_test.dart
```diff
@@ -1653,6 +1653,78 @@ void main() {
...
+
+            check(lastPushedRoute).isNotNull().isA<MaterialAccountWidgetRoute>()
+              .page.isA<MessageListPage>()
+                ..initNarrow.equals(expectedNarrow)
+                ..initAnchorMessageId.equals(message.id);
+          } else {
+            check(lastPushedRoute).isNull();
+          }
+
+          // TODO test tapping whitespace in message
```
gnprice, line 1715: I guess this TODO test is the one that would fail if the `HitTestBehavior.translucent` were changed to `deferToChild`.

Leaving a TODO seems fine.

test/widgets/message_list_test.dart
```diff
@@ -1626,6 +1653,78 @@ void main() {
...
+        });
+      }
+
+      doTest(CombinedFeedNarrow(), false, mkMessage: () => eg.streamMessage());
+      doTest(ChannelNarrow(subscription.streamId), false,
+        mkMessage: () => eg.streamMessage(stream: subscription));
+      doTest(TopicNarrow(subscription.streamId, eg.t(topic)), false,
```
gnprice: nit: can I think clarify by reformatting slightly:
```suggestion
      doTest(expected: false, CombinedFeedNarrow(),
        mkMessage: () => eg.streamMessage());
      doTest(expected: false, ChannelNarrow(subscription.streamId),
        mkMessage: () => eg.streamMessage(stream: subscription));
      doTest(expected: false, TopicNarrow(subscription.streamId, eg.t(topic)),
```

* name the boolean parameter, so its meaning is clear;
* put the short, near-constant-width argument first, so corresponding arguments are aligned;
* put this `mkMessage` on the next line, to match the others.

gnprice (commented):
Thanks! Looks good; small comments below.

test/widgets/message_list_test.dart
chrisbobbe: Yeah; I timed out trying to simulate tapping the back button to clear the state from the previous part of the test (`find.backButton()` wasn't working)

chrisbobbe:

gnprice:
Thanks! Looks good; merging.

Issue #992, opened 2024-10-08
Require Zulip Server 6


gnprice:
This is the next round after:
* #268 

See there for background.

Zulip Server 6 was released 2022-11-17:
https://blog.zulip.com/tag/major-releases/
so it reached 18 months in 2024-05, which means that earlier releases are no longer supported. So the calendar is ripe for us to sweep the codebase for `TODO(server-6)` comments and resolve them.

Prerequisite issues:
* #267 
* #268

PR #1658, opened 2025-07-02
api: Remove backward-compat code for mark-read protocol in FL <155

chrisbobbe:
This code had a switch/case on the Narrow type, so I discovered it while implementing keyword-search narrows.

We support Zulip Server 7 and later (see README) and refuse to connect to older servers. Since we haven't been using this protocol for servers FL 155+, this should be NFC for all servers we connect to, as long as we know their feature level accurately.

Related: #992


chrisbobbe:
Bumping to integration review, given this should be low-risk (removing code that's dead as long as we have accurate FLs) and would help toward search, #1662.

lib/widgets/actions.dart
```diff
@@ -26,25 +26,7 @@ abstract final class ZulipAction {
   /// This is mostly a wrapper around [updateMessageFlagsStartingFromAnchor];
   /// for details on the UI feedback, see there.
   static Future<void> markNarrowAsRead(BuildContext context, Narrow narrow) async {
-    final store = PerAccountStoreWidget.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
-    final useLegacy = store.zulipFeatureLevel < 155; // TODO(server-6)
-    if (useLegacy) {
```
gnprice, line 32: We support Zulip Server 7 and later (see README) and refuse to
    connect to older servers. Since we haven't been using this protocol
    for servers FL 155+, this should be NFC for all servers we connect
    to, as long as we know their feature level accurately.

I think this can be a bit stronger: it's just NFC.

The condition on these lines is based on the `zulipFeatureLevel` we have. If that feature level were below our supported threshold, we wouldn't have gotten to this point: UpdateMachine.load would have (caught and re-)thrown _ServerVersionUnsupportedException, rather than getting as far as constructing a PerAccountStore. So this feature level is always at least 185, and this condition is always false.

Because this is false, the rest of the affected lib/ code is dead (except from tests), so the change is NFC.

gnprice: I guess more specifically it's NFC in the actual app (where the store comes from `UpdateMachine.load`), though not in tests: we have various tests that construct stores with older zulipFeatureLevel, pending #992 and similar cleanups.

chrisbobbe:
Thanks! Revision pushed.

gnprice:
Thanks! Looks good; merging.

Issue #1655, opened 2025-07-02
msglist: Temporary state "this view does not mark messages as read on scroll", with alert and button to leave


chrisbobbe:
Sometimes (as when following a /near/ link) you want to go straight to a specific message, even if it's unread, and you don't want the app to mess with your read state in doing so. Web handles this; see discussion:

[#mobile-team > msglist: anchor when tapping recipient header @ 💬](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/msglist.3A.20anchor.20when.20tapping.20recipient.20header/near/2210470)
> > * File an issue to implement "this view does not mark messages as read" mode; use that mode when following /near/ links […]
> 
> I think some key features of the new mode are:
> - UI that makes it clear what's happening
> - UI to exit the temporary state
> 
> Web has those UI features (an alert with a "Mark as read" button):
> 
> ![Image](https://github.com/user-attachments/assets/6ca5bb98-6827-45e7-9704-b52beac62486)
> 
> (I guess we should only show those if we'd ordinarily be marking read as scroll, given the narrow and the "Mark messages as read on scroll" setting.)
> 
> And I guess we're already missing those features when we enter a "not-marking-read" state on tapping "Mark as unread from here" in the action sheet.

### Related issues

- #1654

PR #1659, opened 2025-07-03
msglist: Softly override mark-read-on-scroll setting in starred/mentions

chrisbobbe:
This isn't so "soft" right now, because we don't offer a way to undo the override -- that's #1655, as the new comment says -- but conceptually this seems correct and seems to match web.

We'll do this for search, too, coming up.


lib/widgets/message_list.dart
```diff
@@ -235,6 +238,18 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
...
+    switch (narrow) {
+      case CombinedFeedNarrow():
+      case ChannelNarrow():
+      case TopicNarrow():
+      case DmNarrow():
+        break;
+      case MentionsNarrow():
+      case StarredMessagesNarrow():
+        // TODO(#1583) write tests for this
+        _markReadOnScroll = false;
```
gnprice, line 251: Huh, is this really what web does? Even if you have the "Automatically mark messages as read" setting set to "Always"? (If that's either of the other values, then it's as if this were false anyway.)

As I said in the office earlier (probably a bit after you sent this PR), it seems fine to me if choosing "Always" for that setting means always. In particular that was the behavior Zulip web had, with no option to do something else, for a number of years.

In any event I think this isn't a blocker for search. With the default value of the auto-mark-read setting, at "conversations only", it'll have no effect.

PR #1662, opened 2025-07-03
Implement basic keyword search

### LINKED ISSUE #252: msglist: Offer search narrows

chrisbobbe:
Fixes #252.

Stacked on #1658 

------

Still TODO before shipping:

- Tests.

-----

### Screenshots

| Light | Dark |
| --- | --- |
| image | image |
| image | image |
| image | image |
| image | image |


lib/widgets/message_list.dart
```diff
@@ -281,8 +281,7 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
     });
   }
 
-  @override
-  Widget build(BuildContext context) {
+  AppBar _buildAppBar(BuildContext context) {
```
gnprice: msglist [nfc]: Pull out _buildAppBar helper

How about making it a separate widget entirely? Seems like that'd make the separation a bit crisper; and it doesn't look like this method is interacting with this widget's state beyond looking at `narrow`.

lib/model/message_list.dart
```diff
@@ -939,7 +960,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   bool _shouldAddOutboxMessage(OutboxMessage outboxMessage) {
     assert(haveNewest);
     return !outboxMessage.hidden
-      && narrow.containsMessage(outboxMessage)
+      && narrow.containsMessage(outboxMessage) == true
```
gnprice, line 966: The app-code changes show just small behavior changes for a
    message-list page in a search narrow, which we'll implement soon:
    - Message events are ignored
    - Outbox messages never appear

nit: s/changes/choices/ — before this commit, there was no answer to the question of how these would behave on a null `containsMessage` result, because the types ruled that out

lib/model/message_list.dart
```diff
@@ -618,6 +618,14 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Narrow get narrow => _narrow;
   Narrow _narrow;
 
+  /// Set [narrow] to [newNarrow], reset, [notifyListeners], and [fetchInitial].
+  void renarrowAndFetch(Narrow newNarrow) {
```
gnprice, line 623: nit: mention this from the [narrow] doc, since it's a lot like a setter for that property

lib/api/model/narrow.dart
```diff
@@ -245,3 +245,16 @@ class ApiNarrowMessageId extends ApiNarrowElement {
     negated: json['negated'] as bool? ?? false,
   );
 }
+
+class ApiNarrowKeywordSearch extends ApiNarrowElement {
+  @override String get operator => 'search';
```
gnprice: nit: let's call this `ApiNarrowSearch`, to keep it closely aligned with the underlying API

(unlike the Narrow subclass, which is more about how we organize the UI within the app)

lib/api/model/narrow.dart
```diff
@@ -245,3 +245,16 @@ class ApiNarrowMessageId extends ApiNarrowElement {
     negated: json['negated'] as bool? ?? false,
   );
 }
+
+class ApiNarrowKeywordSearch extends ApiNarrowElement {
```
gnprice: nit: let's put this before `with` and `id`; those feel more specific than this because they point to an individual message, and naturally go at the end of a list of filters

(I guess that calls for `with` to go after `is`, too, next to `id`.)

lib/model/internal_link.dart
```diff
@@ -92,6 +92,8 @@ Uri narrowLink(PerAccountStore store, Narrow narrow, {int? nearMessageId}) {
         fragment.write(element.operand.toString());
       case ApiNarrowMessageId():
         fragment.write(element.operand.toString());
+      case ApiNarrowKeywordSearch():
+        fragment.write(element.operand.toString());
```
gnprice: This operand is an arbitrary string, right? So it probably needs some encoding to appear here.

lib/model/message_list.dart
```diff
@@ -725,6 +737,15 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Future<void> fetchInitial() async {
     assert(!fetched && !haveOldest && !haveNewest && !busyFetchingMore);
     assert(messages.isEmpty && contents.isEmpty);
+
+    if (narrow case KeywordSearchNarrow(:final keyword) when keyword.isEmpty) {
+      // The server would reject an empty keyword search; skip the request.
+      _haveOldest = true;
+      _haveNewest = true;
+      _setStatus(FetchingStatus.idle, was: FetchingStatus.unstarted);
+      return;
```
gnprice: Hmm, this seems like an awkward layer to handle this at — probably better if the UI code doesn't take it to this point.

This is fine to ship, though, if getting to the alternative is complicated.

lib/model/message_list.dart
```diff
@@ -725,6 +737,15 @@ class MessageListView with ChangeNotifier, _MessageSequence {
   Future<void> fetchInitial() async {
     assert(!fetched && !haveOldest && !haveNewest && !busyFetchingMore);
     assert(messages.isEmpty && contents.isEmpty);
+
+    if (narrow case KeywordSearchNarrow(:final keyword) when keyword.isEmpty) {
```
gnprice: nit: can simplify a bit by converting further into patterns:
```suggestion
    if (narrow case KeywordSearchNarrow(keyword: '')) {
```

lib/model/message_list.dart
```diff
@@ -1124,6 +1143,7 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       case CombinedFeedNarrow():
       case MentionsNarrow():
       case StarredMessagesNarrow():
+      case KeywordSearchNarrow():
         // The messages were and remain in this narrow.
```
gnprice: This comment seems not true of this case: the match may have been on the topic name, or messages that hadn't matched might gain a match on the topic name.

Same outcome, I think, but should get a separate case block so it can have a comment with separate reasoning: we're punting on trying to update the message list, effectively assuming that the match didn't change, because we aren't equipped to evaluate the match without asking the server.

gnprice: … Also this comment isn't quite right for two of these three cases already. I suspect it was written for CombinedFeedNarrow, and not updated when the other two were added.

For those three, though, I think a quick reword suffices:
> The messages didn't enter or leave this narrow.

lib/model/narrow.dart
```diff
@@ -19,7 +19,9 @@ sealed class Narrow {
   /// This does not necessarily mean the message list would show this message
   /// when navigated to this narrow; in particular it does not address the
   /// question of whether the stream or topic, or the sending user, is muted.
-  bool containsMessage(MessageBase message);
+  ///
+  /// Will be null for search narrows.
+  bool? containsMessage(MessageBase message);
```
gnprice: I'd like this to give a bit more of a semantics for what it means for this to be null:
```suggestion
  ///
  /// Null when the client is unable to predict whether the message
  /// satisfies the filters of this narrow, e.g. when this is a search narrow.
  bool? containsMessage(MessageBase message);
```

lib/widgets/message_list.dart
```diff
@@ -591,10 +607,97 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    return TextField(
+      controller: _controller,
+      autocorrect: false,
+
+      // Current servers seem to require straight quotes for the "exact match"-
```
gnprice: [{{when}}](https://en.wikipedia.org/wiki/Template:When)

lib/widgets/message_list.dart
```diff
@@ -591,10 +607,97 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+      // style query. (N.B. the doc says this param is iOS-only.)
+      smartQuotesType: SmartQuotesType.disabled,
+
+      autofocus: true,
+      onSubmitted: (value) => widget.onSubmitted(_valueToNarrow(value)),
```
gnprice: nit: maybe have one method do both these steps (since they always occur together):
```suggestion
      onSubmitted: _handleSubmitted,
```

lib/widgets/message_list.dart
```diff
@@ -591,10 +607,97 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+          icon: Icon(ZulipIcons.remove)),
+        suffixIconColor: designVariables.textMessageMuted,
+        suffixIconConstraints: BoxConstraints(minWidth: 42, minHeight: 42),
+        contentPadding: const EdgeInsetsDirectional.symmetric(vertical: 7),
+        filled: true,
+        fillColor: designVariables.bgSearchInput,
+        border: OutlineInputBorder(
+          borderRadius: BorderRadius.circular(10),
+          borderSide: BorderSide.none),
+        hintStyle: TextStyle(color: designVariables.labelSearchPrompt)));
```
gnprice: nit: should go next to `hintText`

test/model/message_list_test.dart
```diff
@@ -3001,7 +3007,17 @@ void checkInvariants(MessageListView model) {
...
+    switch (model.narrow) {
+      case CombinedFeedNarrow():
+      case ChannelNarrow():
+      case TopicNarrow():
+      case DmNarrow():
+      case MentionsNarrow():
+      case StarredMessagesNarrow():
+        check(model.narrow.containsMessage(message)).isNotNull().isTrue();
+      case KeywordSearchNarrow():
+        check(model.narrow.containsMessage(message)).isNull();
```
gnprice: If it's a KeywordSearchNarrow, then there's nothing that needs to be checked here; it's a fact about that narrow type, not about the list of messages, that `containsMessage` will always be null.

Then I think also this is one place where we don't need to be considering for each new narrow type how it should be handled. We can let the default be that the check is done; if at some point we add another narrow type where the check would fail, then we'll naturally notice at test time, and can fix it then. So this could look like:
```dart
    if (model.narrow is! KeywordSearchNarrow) {
      check(model.narrow.containsMessage(message)).isNotNull().isTrue();
    }
```

Hmm or maybe better yet: this could just check that `containsMessage` is *either* null or true. If it's null, we're unable to confirm whether the message really belongs here, so `checkInvariants` has to let it slide. What it's really looking for is cases where it's false, and therefore MessageListView really should have omitted the message.

gnprice:
(All the comments above are small, except https://github.com/zulip/zulip-flutter/pull/1662#discussion_r2181744174 about `narrowLink`; and I believe that bit of code never actually gets exercised in the app, so it'd be fine to have it just throw UnimplementedError or something.

… Yeah, confirmed: the app only ever calls that function for generating a message link, so with a SendableNarrow.)

alya:
Just a simple "No search results." is the default web app message for no results. We also show stopwords that were excluded, but I don't know if we exclude stopwords in mobile searches?

gnprice:
The stopwords behavior is server-side (I believe it happens within the database, even), so it'll be the same for mobile as web.

lib/widgets/message_list.dart
chrisbobbe: This might be feasible, but the reason I didn't is that Scaffold's `appBar` param is actually typed as `PreferredSizeWidget?`, not `Widget?`. Maybe I can subclass `ZulipAppBar` (which extends `AppBar` / `PreferredSizeWidget`) for this app bar?

lib/model/internal_link.dart
chrisbobbe: Ah indeed, thanks for the catch.

lib/widgets/message_list.dart
gnprice: Hmm I see, yeah. That bit of the Scaffold/AppBar API is kind of a pain.

I don't think there's a great way to do it by subclassing, either. Most of what this build-helper method is doing is computing values that get passed to the AppBar constructor, so those computations would have to happen in the subclass's constructor… and they're using the context, so that's no good.

I have an idea which I'll send as an added commit, with a comment.

gnprice: ac6c7503d msglist [nfc]: Move _buildAppBar to a widget-like class

chrisbobbe:
Thanks for the reviews! Revision pushed. I still intend to write some tests today, but I can do that while you're reviewing this revision. Also @alya I put screenshots in the issue description.

> Just a simple "No search results." is the default web app message for no results.

Works for me for now, but I think ideally we'd show a different message (or no message?) if you haven't requested results yet, i.e. you haven't submitted a search query.

lib/widgets/message_list.dart
gnprice: Then I also went into a bit of a digression and sent a PR upstream that would clean up AppBar's API in this respect:
* https://github.com/flutter/flutter/issues/171603
* https://github.com/flutter/flutter/pull/171604

And I have a small draft commit that would use that to turn `_MessageListAppBar` into a proper widget class:
3ecadf2ba wip use AppBar.preferredSizeFor to clean up _MessageListAppBar

lib/widgets/message_list.dart
```diff
@@ -283,55 +283,10 @@ class _MessageListPageState extends State<MessageListPage> implements MessageLis
 
   @override
   Widget build(BuildContext context) {
-    final store = PerAccountStoreWidget.of(context);
-    final messageListTheme = MessageListTheme.of(context);
-    final zulipLocalizations = ZulipLocalizations.of(context);
-
-    final Color? appBarBackgroundColor;
```
gnprice, line 290: nit: commit message needs edit after squashing

    msglist [nfc]: Move _buildAppBar to a widget-like class
    
    Conceptually this should be a widget, but can't quite be, for the
    reasons explained in the comment.  Still we can organize the code
    in nearly the same way as if it were a widget.

lib/model/message_list.dart
```diff
@@ -1124,12 +1146,20 @@ class MessageListView with ChangeNotifier, _MessageSequence {
       case CombinedFeedNarrow():
       case MentionsNarrow():
       case StarredMessagesNarrow():
-        // The messages were and remain in this narrow.
-        // TODO(#421): … except they may have become muted or not.
+        // The messages didn't enter or leave this narrow.
+        // TODO(#1255): … except they may have become muted or not.
```
gnprice, line 1150: msglist [nfc]: Renumber a TODO
    
    The UserTopic part of this is done, but handling channel
    muting/unmuting is still TODO.

Looks like there's a couple other TODO comments for this issue that should get the same update:
```
lib/model/channel.dart
325:            // TODO(#421) update [MessageListView] if affected
357:    // TODO(#421) update [MessageListView] if affected
```

… Or actually I guess only the first of those two is about channel mutes; the other is topic mutes. Is the latter resolved, then?

gnprice: also nit: make the summary line a bit more specific, like "a TODO about muting"

lib/generated/l10n/zulip_localizations.dart
```diff
@@ -1203,6 +1209,24 @@ abstract class ZulipLocalizations {
   /// **'Unknown'**
   String get userRoleUnknown;
 
+  /// Page title for the 'Search' message view.
+  ///
+  /// In en, this message translates to:
+  /// **'Search'**
+  String get searchMessagesPageTitle;
```
gnprice, line 1216: nit: these look to be happening in the wrong commit

lib/api/model/model.dart
```diff
@@ -865,9 +865,7 @@ sealed class Message<T extends Conversation> extends MessageBase<T> {
   // final string type; // handled by runtime type of object
   @JsonKey(fromJson: _flagsFromJson)
   List<MessageFlag> flags; // Unrecognized flags won't roundtrip through {to,from}Json.
-  final String? matchContent;
-  @JsonKey(name: 'match_subject')
-  final String? matchTopic;
+  // TODO(#1663) matchContent and matchTopic
```
gnprice: Let's leave some kind of breadcrumb in this TODO comment to say there's a commit we'll want to revert to restore these. (It has some useful bits in it beyond what one gets mechanically from the API docs: the rename, and the test for that.)

lib/model/message_list.dart
```diff
@@ -1124,12 +1146,20 @@ class MessageListView with ChangeNotifier, _MessageSequence {
...
 
+      case KeywordSearchNarrow():
+        // This might not be quite true, since matches can be determined by
+        // the topic alone, and topics change. Punt on trying to add/remove
+        // messages, though, because we aren't equipped to evaluate the match
+        // without asking the server.
+        _messagesMovedInternally(messageIds);
+
+
       case ChannelNarrow(:final streamId):
```
gnprice: nit: double blank line

lib/widgets/message_list.dart
```diff
@@ -591,10 +619,101 @@ class MessageListAppBarTitle extends StatelessWidget {
...
+          iconSize: 24,
+          icon: Icon(ZulipIcons.remove)),
+        suffixIconColor: designVariables.textMessageMuted,
+        suffixIconConstraints: BoxConstraints(minWidth: 42, minHeight: 42),
+        contentPadding: const EdgeInsetsDirectional.symmetric(vertical: 7),
+        filled: true,
+        fillColor: designVariables.bgSearchInput,
+        border: OutlineInputBorder(
+          borderRadius: BorderRadius.circular(10),
+          borderSide: BorderSide.none),));
```
gnprice: nit:
```suggestion
          borderSide: BorderSide.none),
      ));
```

lib/widgets/message_list.dart
```diff
@@ -864,8 +983,15 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
     if (!model.fetched) return const Center(child: CircularProgressIndicator());
 
     if (model.items.isEmpty && model.haveNewest && model.haveOldest) {
-      return PageBodyEmptyContentPlaceholder(
-        message: zulipLocalizations.emptyMessageList);
+      final String message;
+
+      if (widget.narrow is KeywordSearchNarrow) {
```
gnprice: nit: group declaration with initializing code
```suggestion
      final String message;
      if (widget.narrow is KeywordSearchNarrow) {
```

gnprice (commented):
Thanks for the revision! All looks good, modulo nits below, and it'd be good to have those tests you're working on.

I also haven't yet read the commit at the start which is #1658. I'll do that next.

lib/widgets/message_list.dart
chrisbobbe: Interesting, thanks for doing that!

chrisbobbe:
Thanks! Revision pushed, and filed #1667 for tests as a followup, for timeliness.

gnprice:
Thanks! Looks good; merging.

PR #1664, opened 2025-07-03
msglist [nfc]: Pull out a _showAsMuted method

gnprice:
This condition is a bit complex already, and will become more so soon (https://github.com/zulip/zulip-flutter/pull/1624#discussion_r2183802267).  So making it a method instead of an expression gives its logic a bit more room to breathe and spread out.

### DIFF
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1767,6 +1767,14 @@ class _SenderRow extends StatelessWidget {
   final MessageBase message;
   final bool showTimestamp;
 
+  bool _showAsMuted(BuildContext context, PerAccountStore store) {
+    final message = this.message;
+    if (!store.isUserMuted(message.senderId)) return false;
+    if (message is! Message) return false; // i.e., if an outbox message
+    return !MessageListPage.revealedMutedMessagesOf(context)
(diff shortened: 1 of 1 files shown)


lib/widgets/message_list.dart
```diff
@@ -1767,6 +1767,14 @@ class _SenderRow extends StatelessWidget {
   final MessageBase message;
   final bool showTimestamp;
 
+  bool _showAsMuted(BuildContext context, PerAccountStore store) {
+    final message = this.message;
+    if (!store.isUserMuted(message.senderId)) return false;
+    if (message is! Message) return false; // i.e., if an outbox message
+    return MessageListPage.revealedMutedMessagesOf(context)
+      .isMutedMessageRevealed(message.id);
```
chrisbobbe: This should be negated, I think: we show as muted if the message is _not_ revealed.

gnprice: Indeed, thanks :slightly_smiling_face: — fixed.

(Which I guess also shows I didn't run tests; we had tests that failed with this, as we should.)

chrisbobbe:
Thanks! Merging.

PR #1670, opened 2025-07-04
KaTeX(8/n): Support colored text in KaTeX content

rajveermalviya:
Related: #46

### DIFF
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -1,7 +1,7 @@
+import 'package:convert/convert.dart';
 import 'package:csslib/parser.dart' as css_parser;
 import 'package:csslib/visitor.dart' as css_visitor;
 import 'package:flutter/foundation.dart';
-import 'package:flutter/widgets.dart';
 import 'package:html/dom.dart' as dom;
 
 import '../log.dart';
@@ -463,6 +463,7 @@ class _KatexParser {
       final stylesheet = css_parser.parse('*{$styleStr}');
       if (stylesheet.topLevels case [css_visitor.RuleSet() && final rule]) {
         double? heightEm;
+        KatexSpanColor? color;
 
         for (final declaration in rule.declarationGroup.declarations) {
           if (declaration case css_visitor.Declaration(
@@ -474,6 +475,18 @@ class _KatexParser {
               case 'height':
                 heightEm = _getEm(expression);
                 if (heightEm != null) continue;
+
+              case 'color':
+                final valueStr = _getRawValue(expression);
+                if (valueStr != null) {
+                  if (valueStr.startsWith('#')) {
+                    color = parseCssHexColor(valueStr);
+                    if (color != null) continue;
+                  }
+
+                  color = _cssNamedColorsMap[valueStr];
+                  if (color != null) continue;
+                }
             }
 
             // TODO handle more CSS properties
@@ -488,6 +501,7 @@ class _KatexParser {
 
         return KatexSpanStyles(
           heightEm: heightEm,
+          color: color,
         );
       } else {
         throw _KatexHtmlParseError();
@@ -503,6 +517,10 @@ class _KatexParser {
       return (expression.value as num).toDouble();
     }
     return null;
+  }
+
+  String? _getRawValue(css_visitor.Expression expression) {
+    return expression.span?.text;
   }
 }
 
@@ -521,6 +539,32 @@ enum KatexSpanTextAlign {
   right,
 }
 
+class KatexSpanColor {
+  const KatexSpanColor(this.r, this.g, this.b, this.a);
+
+  final int r;
+  final int g;
+  final int b;
+  final int a;
+
+  @override
+  bool operator ==(Object other) {
+    return other is KatexSpanColor &&
+      other.r == r &&
+      other.g == g &&
+      other.b == b &&
+      other.a == a;
+  }
+
+  @override
+  int get hashCode => Object.hash('KatexSpanColor', r, g, b, a);
+
+  @override
+  String toString() {
+    return '${objectRuntimeType(this, 'KatexSpanColor')}($r, $g, $b, $a)';
+  }
+}
+
 @immutable
 class KatexSpanStyles {
   final double? heightEm;
@@ -531,6 +575,8 @@ class KatexSpanStyles {
   final KatexSpanFontStyle? fontStyle;
   final KatexSpanTextAlign? textAlign;
 
+  final KatexSpanColor? color;
+
   const KatexSpanStyles({
     this.heightEm,
     this.fontFamily,
@@ -538,6 +584,7 @@ class KatexSpanStyles {
     this.fontWeight,
     this.fontStyle,
     this.textAlign,
+    this.color,
   });
 
   @override
@@ -549,6 +596,7 @@ class KatexSpanStyles {
     fontWeight,
     fontStyle,
     textAlign,
+    color,
   );
 
   @override
@@ -559,7 +607,8 @@ class KatexSpanStyles {
       other.fontSizeEm == fontSizeEm &&
       other.fontWeight == fontWeight &&
       other.fontStyle == fontStyle &&
-      other.textAlign == textAlign;
+      other.textAlign == textAlign &&
+      other.color == color;
   }
 
   @override
@@ -571,6 +620,7 @@ class KatexSpanStyles {
     if (fontWeight != null) args.add('fontWeight: $fontWeight');
     if (fontStyle != null) args.add('fontStyle: $fontStyle');
     if (textAlign != null) args.add('textAlign: $textAlign');
+    if (color != null) args.add('color: $color');
     return '${objectRuntimeType(this, 'KatexSpanStyles')}(${args.join(', ')})';
   }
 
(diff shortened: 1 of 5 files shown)


lib/model/katex.dart
```diff
@@ -523,6 +526,38 @@ class _KatexParser {
...
+
+                final valueStr = _getRawValue(expression);
+                if (valueStr != null) {
+                  if (valueStr.startsWith('#')) {
+                    color = parseCssHexColor(valueStr);
+                    if (color != null) continue;
+                  }
+
+                  color = _cssNamedColorsMap[valueStr];
+                  if (color != null) continue;
```
gnprice: On its face, this looks like even if the value starts with `#` it could overwrite that with what it finds in the map.

That won't happen in practice, because no names start with `#`. But it'd be cleaner to not have to think about that — once we decide it looks like hex notation, we stick with that and don't try something else. Maybe just wrap this second case in an `else`?

lib/model/katex.dart
```diff
@@ -523,6 +526,38 @@ class _KatexParser {
                   if (marginLeftEm < 0) throw _KatexHtmlParseError();
                   continue;
                 }
+
+              case 'color':
+                // `package:csslib` parser emits a HexColorTerm for the `color`
+                // attribute. It automatically resolves the named CSS colors to
+                // their hex values. The `HexColorTerm.value` is the hex
+                // encoded in an integer in the same sequence as the input hex
```
gnprice: This is a helpful comment, thanks.

Since the logic in this case gets somewhat long, let's push all that logic, along with this accompanying comment, into a dedicated private helper function. That will help keep it easy to read in isolation, and also keep this surrounding code easy to read when not thinking specifically about colors.

lib/model/katex.dart
```diff
@@ -556,6 +592,10 @@ class _KatexParser {
     }
     return null;
   }
+
+  String? _getRawValue(css_visitor.Expression expression) {
+    return expression.span?.text;
```
gnprice: This name doesn't seem quite right — this isn't really the value of the expression, so much as the source code of the expression.

When we call this, it does represent the value of a *property*… but only because this expression is the value of a property. So that's not a fact that this method itself expresses.

I think "get raw text" would be a good name. Or just inline it — that's not really longer, and makes transparent what it does.

lib/model/katex.dart
```diff
@@ -688,6 +761,196 @@ class KatexSpanStyles {
...
+
+/// Parses the CSS hex color notation.
+///
+/// See: https://drafts.csswg.org/css-color/#hex-notation
+KatexSpanColor? parseCssHexColor(String hexStr) {
+  final match = _hexColorRegExp.firstMatch(hexStr);
+  if (match == null) return null;
+
+  String hexValue = match.group(1)!;
+  hexValue = hexValue.toLowerCase();
```
gnprice: Is this line needed? I'd have thought that `hex.decode` would accept upper as well as lower case.

lib/model/katex.dart
```diff
@@ -688,6 +761,196 @@ class KatexSpanStyles {
   }
 }
 
+final _hexColorRegExp =
+  RegExp(r'^#([0-9a-fA-F]{3}|[0-9a-fA-F]{4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$');
+
+/// Parses the CSS hex color notation.
+///
+/// See: https://drafts.csswg.org/css-color/#hex-notation
+KatexSpanColor? parseCssHexColor(String hexStr) {
```
gnprice: Thanks for including that spec link — very helpful!

lib/model/katex.dart
```diff
@@ -523,6 +526,38 @@ class _KatexParser {
...
+                // encoded in an integer in the same sequence as the input hex
+                // string. But it also allows some non-conformant CSS hex color
+                // notations, like #f, #ff, #fffff, #fffffff.
+                // See:
+                //   https://drafts.csswg.org/css-color/#hex-notation.
+                //   https://github.com/dart-lang/tools/blob/2a2a2d611/pkgs/csslib/lib/parser.dart#L2714-L2743
+                //
+                // Also the generated integer value will be in 0xRRGGBBAA
+                // sequence (CSS notation), whereas `dart:ui` Color
+                // requires 0xAARRGGBB.
```
gnprice: This doesn't seem material — even in this version, we break it up into bytes and then say
`Color.fromARGB(color.a, color.r, color.g, color.b)`, which isn't a single integer anyway.

gnprice (commented):
Thanks for building this! I especially appreciate the clear comments and the spec links.

Generally this all looks good. See comments below; and please also rebase it atop main, and adjust the style to match the changes that were in #1722 and #1729.

gnprice:
For this and the other TeX content PRs, would you also include two things in the PR description:
* Screenshots of example messages; for this one, I think just one message like the one in the test would be good.
* Some numbers on the effect in an empirical corpus. Specifically, let's use the diff/grep commands shown in #1722 — that should be easy to do and produce a nice clear picture of the overall effect. (Then where it seems appropriate go ahead and include more detail, too; but I imagine that won't be needed for this PR.)

rajveermalviya:

test/model/katex_test.dart
```diff
@@ -566,6 +566,83 @@ class KatexExample extends ContentExample {
         ]),
       ]),
     ]);
+
+  static final mathBlockKatexColorFunction = KatexExample.block(
+    'math block, KaTeX color function, hex color',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2232197
+    r'\color{#f00} 0',
```
gnprice: nit: TeX has no concept of "function"; `\color` here is a LaTeX command, which is a type of TeX macro.

test/model/katex_test.dart
```diff
@@ -566,6 +566,83 @@ class KatexExample extends ContentExample {
...
+        KatexSpanNode(
+          styles: KatexSpanStyles(color: KatexSpanColor(223, 0, 48, 255)),
+          text: '2', nodes: null),
+      ]),
+    ]);
+
+  static final mathBlockKatexPhantomFunction = KatexExample.block(
+    'math block, KaTeX phantom function',
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2229515
+    r'\phantom{*}',
```
gnprice, line 631: For example (apropos of the "function" question): one fun TeX fact is that `\phantom{*}` means exactly the same thing as `\phantom *` would; in both cases, `*` is the argument consumed by `\phantom`.

Similarly, one can write `\frac12` — $$\frac12$$ — as a lazy/quick-to-type way of saying the same thing as `\frac{1}{2}`. With or without curly braces, the `\frac` command will consume the next two things as its arguments; the braces just serve to group a whole string of characters and/or commands into a single item, so that they get consumed together as a single argument.

```suggestion
    r'\phantom{*}',
```

test/model/katex_test.dart
```diff
@@ -566,6 +566,83 @@ class KatexExample extends ContentExample {
...
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mstyle mathcolor="#f00"><mn>0</mn></mstyle></mrow><annotation encoding="application/x-tex">\\color{#f00} 0</annotation></semantics></math></span>'
+        '<span class="katex-html" aria-hidden="true">'
+          '<span class="base">'
+            '<span class="strut" style="height:0.6444em;"></span>'
+            '<span class="mord" style="color:#f00;">0</span></span></span></span></span></p>', [
+      KatexSpanNode(styles: KatexSpanStyles(), text: null, nodes: [
+        KatexStrutNode(heightEm: 0.6444, verticalAlignEm: null),
+        KatexSpanNode(
+          styles: KatexSpanStyles(color: KatexSpanColor(255, 0, 0, 255)),
+          text: '0', nodes: null),
```
gnprice: nit:
```suggestion
          text: '0'),
```

test/model/katex_test.dart
```diff
@@ -566,6 +566,83 @@ class KatexExample extends ContentExample {
...
+    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2232197
+    r'\color{#f00} 0',
+    '<p>'
+      '<span class="katex-display"><span class="katex">'
+        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mstyle mathcolor="#f00"><mn>0</mn></mstyle></mrow><annotation encoding="application/x-tex">\\color{#f00} 0</annotation></semantics></math></span>'
+        '<span class="katex-html" aria-hidden="true">'
+          '<span class="base">'
+            '<span class="strut" style="height:0.6444em;"></span>'
+            '<span class="mord" style="color:#f00;">0</span></span></span></span></span></p>', [
+      KatexSpanNode(styles: KatexSpanStyles(), text: null, nodes: [
```
gnprice: nit: simplify
```suggestion
      KatexSpanNode(nodes: [
```

test/model/katex_test.dart
```diff
@@ -604,4 +685,49 @@ void main() async {
   }, skip: Platform.isWindows, // [intended] purely analyzes source, so
        // any one platform is enough; avoid dealing with Windows file paths
   );
+
+  group('parseCssHexColor', () {
```
gnprice, line 667: nit: let's keep the "all examples are tested" test at the end — it's a rather special kind of test, in that it reads the test file's own source code

test/model/katex_test.dart
gnprice: I just sent #1767 to simplify a different aspect of these tests' names and descriptions. Combining that and my comment above, one good version of this would be:
```suggestion
  static final color = KatexExample.block(
    r'\color: 3-digit hex color',
    // https://chat.zulip.org/#narrow/channel/7-test-here/topic/Rajesh/near/2232197
    r'\color{#f00} 0',
```

rajveermalviya:

gnprice:
Thanks! Looks good; merging.

Made a tweak to clarify one test name:
```diff
   // KaTeX custom color macros, see https://github.com/KaTeX/KaTeX/blob/9fb63136e/src/macros.js#L977-L1033
   static final customColorMacro = KatexExample.block(
-    r'\red: custom color macro',
+    r'\red, custom KaTeX color macro: CSS 6-digit hex color',
```
since the meaning of "custom" here is that it's something KaTeX made up (according to that handy link you included in the comment).

gnprice:
Oh and also tweaked the commit message:

    -    content: Support colored text in KaTeX content
    +    katex: Support colored text in KaTeX content

Now that we've got the KaTeX logic all into its own files, I think a corresponding specific prefix is helpful.

PR #1686, opened 2025-07-07
store: Pass `dont_block` in first poll attempt after a failure

### LINKED ISSUE #979: Loading indicator does not go away when the event queue appears to be active

chrisbobbe:
Fixes #979.

Discussion:
  https://github.com/zulip/zulip-flutter/issues/979#issuecomment-2993296322

------

Tested manually, with and without this change:

1. Log out from CZO (too active; events coming in frequently); connect to a not-very-active org

2. Make an artificial network error in the `getEvents` binding:

```diff
diff --git lib/api/route/events.dart lib/api/route/events.dart
index bd14521c7..da7bf4fea 100644
--- lib/api/route/events.dart
+++ lib/api/route/events.dart
@@ -1,6 +1,7 @@
 import 'package:json_annotation/json_annotation.dart';
 
 import '../core.dart';
+import '../exception.dart';
 import '../model/events.dart';
 import '../model/initial_snapshot.dart';
 
@@ -27,6 +28,7 @@ Future<InitialSnapshot> registerQueue(ApiConnection connection) {
 Future<GetEventsResult> getEvents(ApiConnection connection, {
   required String queueId, int? lastEventId, bool? dontBlock,
 }) {
+  throw NetworkException(routeName: 'foo', message: 'foo', cause: 'foo');
   return connection.get('getEvents', GetEventsResult.fromJson, 'events', {
     'queue_id': RawParameter(queueId),
     if (lastEventId != null) 'last_event_id': lastEventId,
```

3. Watch the logs to see retries going by
4. Comment out the `throw NetworkException` line and do a hot-reload
5. Watch the logs for `dont_block` in the `GET /events`, and watch the app to see if the loading indicator goes away quickly
   - Without the change, I don't see `dont_block`, and the loading indicator doesn't go away until the next event comes in
   - With the change, I see `dont_block`, and the loading indicator goes away quickly

### DIFF
--- a/test/model/store_test.dart
+++ b/test/model/store_test.dart
@@ -841,7 +841,7 @@ void main() {
         await prepareError();
         updateMachine.debugAdvanceLoop();
         async.elapse(Duration.zero);
-        check(store).isLoading.isTrue();
+        check(store).isRecoveringEventStream.isTrue();
 
         if (expectBackoff) {
           // The reload doesn't happen immediately; there's a timer.
@@ -853,7 +853,7 @@ void main() {
         // The global store has a new store.
         check(globalStore.perAccountSync(store.accountId)).not((it) => it.identicalTo(store));
         updateFromGlobalStore();
-        check(store).isLoading.isFalse();
+        check(store).isRecoveringEventStream.isFalse();
 
         // The new UpdateMachine updates the new store.
         updateMachine.debugPauseLoop();
@@ -879,7 +879,7 @@ void main() {
         updateMachine.debugAdvanceLoop();
         async.elapse(Duration.zero);
         checkLastRequest(lastEventId: 1);
-        check(store).isLoading.isTrue();
+        check(store).isRecoveringEventStream.isTrue();
 
         // Polling doesn't resume immediately; there's a timer.
         check(async.pendingTimers).length.equals(1);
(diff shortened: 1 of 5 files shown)


chrisbobbe:
I found a comment that needed updating; revision pushed.

test/model/store_test.dart
```diff
@@ -1054,21 +1057,26 @@ void main() {
         return awaitFakeAsync((async) async {
           await prepare();
 
+          bool expectDontBlock = false; // (Not on the first attempt, but after that)
```
gnprice: This is a bit confusing because of what I think is a double negative. This value takes effect precisely on the first attempt, right? Then it's set to a different value for the next attempt.

gnprice: I think probably just dropping this comment clears things up. The reader can see a few lines below how this gets mutated after the first attempt.

chrisbobbe: > This value takes effect precisely on the first attempt, right? Then it's set to a different value for the next attempt.

Yeah, that's right.

chrisbobbe:
Thanks!! Revision pushed.

gnprice:
Thanks! Looks good; merging.

Issue #233, opened 2023-07-21
autocomplete: Support user-group @-mentions


chrisbobbe:
With #207 being merged, we now have an @-mention autocomplete UI for user mentions! 🎉

We should extend it to support autocompleting user-_group_ mentions, like `@*mobile*`.

### Related issues

- Prerequisite: #662

gnprice:
When implementing this, we should be sure to exclude deactivated groups from the results.

Until recently, such groups weren't included in the data the server gave us in the first place. Now they are:
* https://chat.zulip.org/#narrow/stream/378-api-design/topic/deactivated.20groups.20in.20.60GET.60.20endpoints/near/1940339
* https://github.com/zulip/zulip-flutter/issues/662#issuecomment-2344818170
* https://github.com/zulip/zulip/pull/30504

chrisbobbe:
Requested on CZO as a "mini-blocker" for a user's team: [#mobile > Flutter app: group tags @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20app.3A.20group.20tags/near/2196872)

alya:
Requested [in the Rust community](https://rust-lang.zulipchat.com/#narrow/channel/122653-zulip/topic/new.20mobile.20app.20feedback/near/526415430).

PR #1688, opened 2025-07-08
Add user groups to store and API binding

### LINKED ISSUE #662: Store user groups
We should store the list of user groups in the realm, with their metadata and their memberships, and keep the data up to date.

This data originates as `realm_user_groups` in the initial snapshot at https://zulip.com/api/register-queue#response , and is updated by events.

As a preparatory refactor, we'll move `PerAccountStore.users` (which is a `Map<int, User>`, indexed by user ID) to live on a new type `UserStore`. This will closely follow the example of `streams` and `StreamStore`.

Then the user groups will be added as an additional data structure on `UserStore`.

This is a prerequisite for features involving user groups, in particular:
* #233

gnprice:
Fixes #662, toward #233 user-group @-mentions.


lib/api/model/model.dart
```diff
@@ -199,6 +199,38 @@ enum Emojiset {
...
+  // List<int> directSubgroupIds; // TODO(#1687) track group members
+
+  String name;
+  String description;
+
+  // final int? dateCreated; // not using; ignore
+  // final int? creatorId; // not using; ignore
+
+  final bool isSystemGroup;
+  bool deactivated;
```
chrisbobbe, line 221: The doc says "New in Zulip 10.0 (feature level 290)"; I think we want to make this optional?

chrisbobbe (commented):
Thanks! Just one comment below, otherwise LGTM please merge at will.

I wonder if we want to sometimes do a smoke test with the earliest supported server version (7.0 now) to make sure we're not accidentally breaking support for some servers.

lib/api/model/model.dart
gnprice: Oh good catch, thanks. And I guess we can default it to false — I think before that feature level, there just was no concept of a deactivated group, so all groups were effectively active.

chrisbobbe: > before that feature level, there just was no concept of a deactivated group

Hmm, `realm_user_groups` in [register-queue](https://zulip.com/api/register-queue) says:

> Changes: Prior to Zulip 10.0 (feature level 294), deactivated groups were included for all the clients.

I'm not sure whether or not that's consistent with that.

gnprice:
> I wonder if we want to sometimes do a smoke test with the earliest supported server version (7.0 now) to make sure we're not accidentally breaking support for some servers.

Hmm, with a live server, then?

Yeah, in principle that would probably be a good thing to do. I think that's been a rare class of bug for us to have, though, and it's an expensive type of test to set up and maintain (and to run), so I don't expect it to be a priority any time soon.

Effectively we've been relying on Zulip's high-quality API docs, including its meticulous change notices; and on both author and reviewer reading those API docs, like you just demonstrated :slightly_smiling_face:

lib/api/model/model.dart
gnprice: Yeah, that caused me to have to reread a couple of times. But I think the story is:
* Starting in 290, there was such a thing as a deactivated group.
* Then starting in 294, deactivated groups were kept out of the list by default.

See also the 290 and 294 entries in the changelog https://zulip.com/api/changelog . In particular 290 includes:
> * [POST /user_groups/{user_group_id}/deactivate](https://zulip.com/api/deactivate-user-group): Added new API endpoint to deactivate a user group.
> * [POST /register](https://zulip.com/api/register-queue), [GET /user_groups](https://zulip.com/api/get-user-groups): Added deactivated field in the user group objects to identify deactivated user groups.

chrisbobbe: Cool. This makes sense then:

> And I guess we can default it to false

gnprice:
Huh, infra failure in CI:
```
Downloading Linux x64 Dart SDK from Flutter engine 36b77984e2ee2b734a5ec59fe1193da311888f01...

100   255  100   255    0     0   411[8](https://github.com/zulip/zulip-flutter/actions/runs/16157122950/job/45601779716?pr=1688#step:6:9)      0 --:--:-- --:--:-- --:--:--  4180

```

Rerunning.

gnprice:
Retried a couple of times; same error on all three tries so far.

gnprice:
… Well huh. That commit is Flutter's `origin/main`… but not its `origin/master`:
```
$ git log --oneline --graph --boundary 36b77984e2ee2b734a5ec59fe1193da311888f01...origin/master
* 36b77984e (origin/main) fix minor spelling mistake in roadmap (#167888)
| * 43312d1cd (origin/master, origin/HEAD) Roll Fuchsia Linux SDK from AinHuT0vgOelA1g7_... to 0-xqmXWc4cXzw3tfe... (#171823)
|/  
o e80b19126 Marks Linux_android_emu android_display_cutout to be unflaky (#171140)
```

That shouldn't happen. And that probably explains why we're failing to fetch an engine artifact from there.

gnprice:
OK, worked around that infra issue in #1690. Now CI is failing only here:
```
Running flutter_version...
fatal: ambiguous argument 'origin/main..d35bde5363d5e25b71d69a81e8c93b0ee3272609': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

I'll fix that too; but it's unrelated to this PR, so I'll go ahead and merge.

Issue #1663, opened 2025-07-03
search: Highlight matches in content and topic


chrisbobbe:
Web highlights matches in the topic and content when you search:

<img width="807" alt="Image" src="https://github.com/user-attachments/assets/de6e71a2-8f0c-46f9-8729-ef941a44b0d9" />

We should do the same.

### Implementation

See `match_content` and `match_subject` in the API doc: https://zulip.com/api/get-messages

See "Sub-issues" in the issue metadata.

PR #1689, opened 2025-07-08
model: Restore matchContent/matchTopic but remove them in reconcileMessages

chrisbobbe:
This reverts commit ac7be1666 while still keeping these per-keyword-search fields out of MessageStore (by removing them in reconcileMessages).

Related: #1663


gnprice:
Sure, this approach LGTM. Merged.

Issue #1149, opened 2024-12-13
[iOS] Missing iOS splash configuration


chyiiiiiiiiiiii:
Now: If we open the app for the first time, you will see the bland and white screen during 200ms.

Goal: Add splash setting for iOS. If users open the apps, the screen will show a specific color bg, or specific effect first, and smooth to the first screen in the app.

Version: 0.0.24

<img src="https://github.com/user-attachments/assets/df913257-3337-46a6-a8bd-d087b2636319" width="250"/>

Sabrina902-prog:
Ok

gnprice:
Thanks for the report. I think the existing behavior is OK if you're in light mode — the splash screen is plain white, and then the app's first screen is close to that.

As the screen capture shows, though, it's jarring when in dark mode. Instead, in dark mode we should have the splash screen be a dark background, ideally the exact same as the app's overall background in dark mode.

chrisbobbe:
> As the screen capture shows, though, it's jarring when in dark mode. Instead, in dark mode we should have the splash screen be a dark background, ideally the exact same as the app's overall background in dark mode.

This came up in launch feedback:

[#mobile > white screen when starting up, also in dark mode @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/white.20screen.20when.20starting.20up.2C.20also.20in.20dark.20mode/near/2213365)
> When I launch the new app on iOS, it loads with a bright white screen. If possible please make this startup screen black when dark mode is set. The white screen is a bit of an eye-hurter 😵😂

chrisbobbe:
Here are Apple's design guidelines for the launch screen: https://developer.apple.com/design/human-interface-guidelines/launching#Launch-screens

I think a plain solid color would be pretty much consistent with that, as long as it's light in light mode and dark in dark mode.

gnprice:
The fix for this issue went out in yesterday's release 30.0.261: [#announce > mobile releases @ 💬](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2217685). @chyiiiiiiiiiiii thanks again for the report — please try out the new release and see what you think!

PR #1697, opened 2025-07-09
ios: Use main background color as the splash screen

rajveermalviya:
Use `DesignVariables.mainBackground` as the iOS launch screen colors.

Discussion:
  https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/splash.20screen.20on.20iOS/near/2215883

Fixes: #1149


test/widgets/store_test.dart
```diff
@@ -98,14 +98,15 @@ void main() {
...
     // Even after the store must have loaded,
     // still shows loading page while blockingFuture is pending.
-    check(find.byType(CircularProgressIndicator)).findsOne();
+    check(find.byType(BlankLoadingPlaceholder)).findsOne();
     check(find.text('done')).findsNothing();
 
     // Once blockingFuture completes…
     completer.complete();
     await tester.pump();
+    await tester.pump();
```
rajveermalviya: I am not sure why the extra `tester.pump()` is needed now.

chrisbobbe:
Thanks, this looks great! Can we add bidirectional comments on the new "colorset" and on `DesignVariables.mainBackground`, saying that if one is changed, the other should be changed too? (Not sure offhand if the "colorset" file format supports comments)

rajveermalviya:
> Not sure offhand if the "colorset" file format supports comments

Unfortunately it doesn't, anyway added the one-way comment on `DesignVariables.mainBackground` for now.

ios/Runner/Assets.xcassets/LaunchBackground.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
+  "colors" : [
+    {
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "1.000",
+          "blue" : "0.941",
+          "green" : "0.941",
+          "red" : "0.941"
```
chrisbobbe: nit: Ah I think we can use a hex representation, to make it easier to compare with `DesignVariables.mainBackground`?

Like in zulip-mobile:

https://github.com/zulip/zulip-mobile/blob/eb8505c4a0ccfb7bdd817ca9022c11ce57ab69d8/ios/ZulipMobile/Assets.xcassets/Brand.colorset/Contents.json#L8-L10

```
          "blue" : "0xFE",
          "green" : "0x92",
          "red" : "0x64"
```

chrisbobbe: Hopefully that would be an NFC change; hopefully it wasn't a factor in my failing to get things to work (as I mentioned in discussion)

rajveermalviya: Updated, and tested that the behaviour is same.

chrisbobbe:
Thanks, LGTM! @gnprice PTAL.

chrisbobbe:
Also GitHub shows a rebase conflict in theme.dart

rajveermalviya:
(Rebased to main.)

test/widgets/store_test.dart
gnprice: Puzzling, yeah. I just spent a few minutes investigating; and I don't see why this would have made a difference, but it does.

This potentially indicates a real bug, though purely in performance and not correctness: it may mean that the loading screen sticks around for a frame longer than it needs to before things get going, so that the user experiences the app starting up a bit more slowly.

I'll add a TODO comment, here and the other instance below. This code *looks* wrong — why is this extra pump needed? — and code that looks wrong should have a comment so the reader doesn't have to wonder about it.

PR #1698, opened 2025-07-09
KaTeX (2/n): Support for vertical offsets in spans

rajveermalviya:
Stacked on top of: #1452

| Web | Flutter |
| - | - |
| Screenshot 2025-05-19 at 22 50 20 | Screenshot 2025-05-19 at 22 50 43 |

Related: #46

### DIFF
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -429,6 +429,37 @@ class KatexStrutNode extends KatexNode {
   }
 }
 
+class KatexVlistNode extends KatexNode {
+  const KatexVlistNode({
+    required this.rows,
+    super.debugHtmlNode,
+  });
+
+  final List<KatexVlistRowNode> rows;
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return rows.map((row) => row.toDiagnosticsNode()).toList();
+  }
+}
+
+class KatexVlistRowNode extends ContentNode {
+  const KatexVlistRowNode({
+    required this.verticalOffsetEm,
+    required this.node,
+    super.debugHtmlNode,
+  });
+
+  final double verticalOffsetEm;
+  final KatexSpanNode node;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(DoubleProperty('verticalOffsetEm', verticalOffsetEm));
+  }
+}
+
 class MathBlockNode extends MathNode implements BlockContentNode {
   const MathBlockNode({
     super.debugHtmlNode,
--- a/lib/model/content.dart
+++ b/lib/model/content.dart
@@ -458,6 +458,11 @@ class KatexVlistRowNode extends ContentNode {
     super.debugFillProperties(properties);
     properties.add(DoubleProperty('verticalOffsetEm', verticalOffsetEm));
   }
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return [node.toDiagnosticsNode()];
+  }
 }
 
 class MathBlockNode extends MathNode implements BlockContentNode {
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -246,13 +246,17 @@ class _KatexParser {
           for (final innerSpan in vlist.nodes) {
             if (innerSpan case dom.Element(
               localName: 'span',
-              className: '',
               nodes: [
                 dom.Element(localName: 'span', className: 'pstrut') &&
                     final pstrutSpan,
                 ...final otherSpans,
               ],
             )) {
+              if (innerSpan.className != '') {
+                throw _KatexHtmlParseError('unexpected CSS class for '
+                  'vlist inner span: ${innerSpan.className}');
+              }
+
               var styles = _parseSpanInlineStyles(innerSpan);
               if (styles == null) throw _KatexHtmlParseError();
               if (styles.verticalAlignEm != null) throw _KatexHtmlParseError();
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -224,10 +224,18 @@ class _KatexParser {
           dom.Element(localName: 'span', className: 'vlist-r', nodes: [
             dom.Element(localName: 'span', className: 'vlist', nodes: [
               dom.Element(localName: 'span', className: '', nodes: []),
-            ]),
+            ]) && final vlist,
           ]),
         ]) {
-          // Do nothing.
+          // In the generated HTML the .vlist in second .vlist-r span will have
+          // a "height" inline style which we ignore, because it doesn't seem
+          // to have any effect in rendering on the web.
+          // But also make sure there aren't any other inline styles present.
+          final vlistStyles = _parseSpanInlineStyles(vlist);
+          if (vlistStyles != null
+            && vlistStyles.filter(heightEm: false) != const KatexSpanStyles()) {
+            throw _KatexHtmlParseError();
+          }
         } else {
           throw _KatexHtmlParseError();
         }
@@ -241,6 +249,17 @@ class _KatexParser {
         if (vlistR.nodes.first
             case dom.Element(localName: 'span', className: 'vlist') &&
                 final vlist) {
+          // Same as above for the second .vlist-r span, .vlist span in first
+          // .vlist-r span will have "height" inline style which we ignore,
+          // because it doesn't seem to have any effect in rendering on
+          // the web.
+          // But also make sure there aren't any other inline styles present.
+          final vlistStyles = _parseSpanInlineStyles(vlist);
+          if (vlistStyles != null
+            && vlistStyles.filter(heightEm: false) != const KatexSpanStyles()) {
+            throw _KatexHtmlParseError();
+          }
+
           final rows = <KatexVlistRowNode>[];
 
           for (final innerSpan in vlist.nodes) {
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -209,11 +209,99 @@ class _KatexParser {
         debugHtmlNode: debugHtmlNode);
     }
 
+    if (element.className == 'vlist-t'
+        || element.className == 'vlist-t vlist-t2') {
+      final vlistT = element;
+      if (vlistT.nodes.isEmpty) throw _KatexHtmlParseError();
+      if (vlistT.attributes.containsKey('style')) throw _KatexHtmlParseError();
+
+      final hasTwoVlistR = vlistT.className == 'vlist-t vlist-t2';
+      if (!hasTwoVlistR && vlistT.nodes.length != 1) throw _KatexHtmlParseError();
+
+      if (hasTwoVlistR) {
+        if (vlistT.nodes case [
+          _,
+          dom.Element(localName: 'span', className: 'vlist-r', nodes: [
+            dom.Element(localName: 'span', className: 'vlist', nodes: [
+              dom.Element(localName: 'span', className: '', nodes: []),
+            ]),
+          ]),
+        ]) {
+          // Do nothing.
+        } else {
+          throw _KatexHtmlParseError();
+        }
+      }
+
+      if (vlistT.nodes.first
+          case dom.Element(localName: 'span', className: 'vlist-r') &&
+              final vlistR) {
+        if (vlistR.attributes.containsKey('style')) throw _KatexHtmlParseError();
+
+        if (vlistR.nodes.first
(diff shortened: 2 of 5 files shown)


lib/model/katex.dart
```diff
@@ -209,11 +209,102 @@ class _KatexParser {
...
+
+      if (hasTwoVlistR) {
+        if (vlistT.nodes case [
+          _,
+          dom.Element(localName: 'span', className: 'vlist-r', nodes: [
+            dom.Element(localName: 'span', className: 'vlist', nodes: [
+              dom.Element(localName: 'span', className: '', nodes: []),
+            ]),
+          ]),
+        ]) {
```
gnprice: Interesting — do we not need any information from this second row of the table?

It looks like an example of how that second row looks in the HTML is (from ContentExample.mathBlockKatexVertical2 added here):
```html
                  <span class="vlist-r">
                    <span class="vlist" style="height:0.15em;"><span></span></span></span>
```

What's the effect of that `height` value when the whole vlist-t gets rendered in a browser?

lib/model/katex.dart
```diff
@@ -209,11 +209,102 @@ class _KatexParser {
         debugHtmlNode: debugHtmlNode);
     }
 
+    if (element.className == 'vlist-t'
+        || element.className == 'vlist-t vlist-t2') {
+      final vlistT = element;
+      if (vlistT.nodes.isEmpty) throw _KatexHtmlParseError();
+      if (vlistT.attributes.containsKey('style')) throw _KatexHtmlParseError();
+
+      final hasTwoVlistR = vlistT.className == 'vlist-t vlist-t2';
```
gnprice, line 218: There's also a CSS rule for .vlist-t2 in the KaTeX CSS:
```css
    .vlist-t2 {
        margin-right: -2px;
    }
```

What's the net effect of that? On its face it looks like something we should have to take account of.

lib/model/katex.dart
```diff
@@ -209,11 +209,102 @@ class _KatexParser {
...
+                  text: null,
+                  nodes: _parseChildSpans(otherSpans))));
+            } else {
+              throw _KatexHtmlParseError();
+            }
+          }
+
+          return KatexVlistNode(
+            rows: rows,
+            debugHtmlNode: kDebugMode ? vlistT : null,
```
gnprice: nit:
```suggestion
            debugHtmlNode: debugHtmlNode,
```

This is equivalent, right? Seems simpler for the reader — makes it the same as most of our `debugHtmlNode` arguments.

lib/model/katex.dart
```diff
@@ -209,11 +209,102 @@ class _KatexParser {
...
+          case dom.Element(localName: 'span', className: 'vlist-r') &&
+              final vlistR) {
+        if (vlistR.attributes.containsKey('style')) throw _KatexHtmlParseError();
+
+        if (vlistR.nodes.first
+            case dom.Element(localName: 'span', className: 'vlist') &&
+                final vlist) {
+          final rows = <KatexVlistRowNode>[];
+
+          for (final innerSpan in vlist.nodes) {
```
gnprice, line 265: Hmm, what's meant by calling these "rows"?

In CSS terms, the element that's a "row" here is the `.vlist-r` — that's the one with `display: table-row`. But that's the grandparent of these elements — these are the children of the `.vlist`, which is one cell of the table (having `display: table-cell`).

lib/model/katex.dart
```diff
@@ -209,11 +209,102 @@ class _KatexParser {
...
+              nodes: [
+                dom.Element(localName: 'span', className: 'pstrut') &&
+                    final pstrutSpan,
+                ...final otherSpans,
+              ],
+            )) {
+              if (innerSpan.className != '') {
+                throw _KatexHtmlParseError('unexpected CSS class for '
+                  'vlist inner span: ${innerSpan.className}');
+              }
```
gnprice, line 277: I guess this points at a different, sometimes-applicable downside of if-case syntax (other than the fact that it puts the failure case way at the bottom on the other side of the main code, which we can hope to someday have a "guard let" feature to address): it doesn't make room for emitting specific errors for specific types of failures to match.

test/model/content_test.dart
```diff
@@ -943,6 +943,387 @@ class ContentExample {
...
+                              ])),
+                        ]),
+                    ]),
+                ]),
+            ]),
+        ]),
+    ]);
+
+  static const mathBlockKatexVertical2 = ContentExample(
+    'math block katex vertical 2',
```
gnprice: Can you give these examples names that reflect what interesting situations they're each exercising that isn't covered by the other examples?

For example, this one introduces `vlist-t2`, which is definitely an interesting case that's good to be sure to cover.

test/model/content_test.dart
```diff
@@ -943,6 +943,387 @@ class ContentExample {
...
+                              ])),
+                        ]),
+                    ]),
+                ]),
+            ]),
+        ]),
+    ]);
+
+  static const mathBlockKatexVertical3 = ContentExample(
+    'math block katex vertical 3',
```
gnprice: (And I'm not sure what this example demonstrates that isn't covered by the "vertical 1" example above. So that would be good to make explicit.)

gnprice (commented):
Thanks for all your work on this! Now that #1452 is merged, I'm reaching these commits for detailed review :-)

Comments below. Posting this now because I'm going AFK, but I've read most of the branch so far: partway through the tests, and read all the other changes.

gnprice:
(As we discussed on #1452 earlier, let's avoid major revisions to these commits until this PR and #1559 are merged — that will help reduce the work of tracking what changed between the last few releases, which have already included versions of these, and the version we eventually merge. Instead let's stick to small edits for the moment where possible, and make larger changes as follow-up PRs.)

lib/model/katex.dart
rajveermalviya: The value of `height` on `span.vlist-r span.vlist` specifies the height of the whole `span.vlist-t`. And if there are two `span.vlist-r` then the addition of `height` value of both `span.vlist-r span.vlist` specifies the height of the whole `span.vlist-t`.

From the web dev console, disabling `height` on either `span.vlist-r span.vlist` doesn't seem to affect the rendering.

lib/model/katex.dart
rajveermalviya: This will need support for negative margins, so will add a commit in the next PR for this. (It will also need changes for supporting non-`em` values in KatexSpanStyles.)

lib/model/katex.dart
rajveermalviya: IIRC, I called them "rows" because they seem like rows in a column with different vertical offset.

test/model/content_test.dart
rajveermalviya: Removed this example as it was technically same as "vertical 1", just different characters.

rajveermalviya:

Also, I've added commits for 2 bug fixes:
- ca001586 content: Implement debugDescribeChildren for KatexVlistRowNode
- (in [#1559](https://github.com/zulip/zulip-flutter/pull/1559)) 2e0c092c content test: Add test for negative margins on a vlist row in KaTeX content 
- (in [#1559](https://github.com/zulip/zulip-flutter/pull/1559)) d4ccfc8b content: Filter negative margin styles if present on a vlist row

rajveermalviya:
CI failure is unrelated, will investigate later.
Logs:
```
2025-07-17T00:50:46.9768068Z Resolving dependencies...
2025-07-17T00:50:49.0495710Z The current Flutter SDK version is 0.0.0-unknown.
2025-07-17T00:50:49.0496135Z 
2025-07-17T00:50:49.0496587Z Because zulip requires Flutter SDK version >=3.33.0-1.0.pre.832, version solving failed.
2025-07-17T00:50:49.0497167Z 
2025-07-17T00:50:49.0497175Z 
2025-07-17T00:50:49.0497468Z You can try the following suggestion to make the pubspec resolve:
2025-07-17T00:50:49.0498127Z * Try using the Flutter SDK version: 3.35.0-0.0.pre. 
2025-07-17T00:50:49.1161302Z Failed to update packages.
```

lib/model/katex.dart
gnprice: OK, we can go with that.

lib/model/katex.dart
gnprice: Cool, sounds good. Please file a brief follow-up issue for that, then; and then this code can have a TODO comment mentioning that issue.

lib/model/content.dart
```diff
@@ -429,6 +429,42 @@ class KatexStrutNode extends KatexNode {
...
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(DoubleProperty('verticalOffsetEm', verticalOffsetEm));
+  }
+
+  @override
+  List<DiagnosticsNode> debugDescribeChildren() {
+    return [node.toDiagnosticsNode()];
```
gnprice, line 464: content: Implement debugDescribeChildren for KatexVlistRowNode
    
    Turns out that anything under KatexVlistRowNode wasn't being
    tested by content tests, fix that by implementing this method.

Hmm indeed, good to catch this!

This meant it wasn't appearing in either the actual or expected tree, so it just got missed.

I went and scanned through all the other node classes in this file, and I think this is the only such omission.

lib/model/katex.dart
gnprice: Hmm, OK. A bit odd that there's that information there and it's not getting used; but maybe it's another example where it either used to get used, or gets used within KaTeX's internal processing.

Let's definitely mark that with a comment, though, because otherwise it looks wrong to a reader who's comparing this parsing code to the HTML. In fact we should do that for both of the `.vlist` cases — this one for the second `.vlist-r`'s child, and below for the first/main `.vlist-r`'s child.

test/model/content_test.dart
```diff
@@ -943,6 +943,317 @@ class ContentExample {
...
+                          '<span class="mord mtight">'
+                            '<span class="mord mtight">′</span></span></span></span></span></span></span></span></span></span></span></span></span></p>',
+    [
+      MathBlockNode(
+        texSource: 'a\'',
+        nodes: [
+          KatexSpanNode(
+            styles: KatexSpanStyles(),
+            text: null,
+            nodes: [
```
gnprice: Let's make these examples more compact in the same way I described today at https://github.com/zulip/zulip-flutter/pull/1559#discussion_r2211697392 . I think that will help making it easier to visually scan for the interesting parts.

gnprice:
> CI failure is unrelated, will investigate later. Logs:
> 
> ```
> 2025-07-17T00:50:46.9768068Z Resolving dependencies...
> 2025-07-17T00:50:49.0495710Z The current Flutter SDK version is 0.0.0-unknown.
> […]

I think the problem is that the last tag is more than 1000 commits behind origin/main (or, equally, master):

```
$ flutter --version
Flutter 3.33.0-1.0.pre-1005 • channel main • https://github.com/flutter/flutter.git
Framework • revision 9c626d9f9a (3 hours ago) • 2025-07-16 23:03:27 -0400
Engine • hash de6ba3d359194bd9ddc0c917c2d1f4c531a63ed0 (revision 9c626d9f9a) (2 hours ago)
• 2025-07-17 03:03:27.000Z
Tools • Dart 3.10.0 (build 3.10.0-6.0.dev) • DevTools 2.48.0
```

The quick fix is to bump the `--depth` in `ci.yml` so that it fetches, say, 2000 commits instead of 1000.

Then we should probably also follow up to get the upstream Flutter release team to start putting tags in main again, so that version numbers in the main/master channel are accurate. (There's been a 3.34 and a 3.35 at this point, so the version number shouldn't be saying "3.33".) That's an issue whenever using the main (or master) channel, even when one makes a full clone without using `--depth`.

rajveermalviya:

lib/model/katex.dart
```diff
@@ -209,11 +209,122 @@ class _KatexParser {
...
+                node: KatexSpanNode(
+                  styles: styles,
+                  text: null,
+                  nodes: _parseChildSpans(otherSpans))));
+            } else {
+              throw _KatexHtmlParseError();
+            }
+          }
+
+          // TODO(1716) Handle styling for .vlist-t2 spans
```
gnprice: nit: follow pattern of our other TODO comments by making more explicit this is an issue reference:
```suggestion
          // TODO(#1716) Handle styling for .vlist-t2 spans
```

gnprice (commented):
Thanks! One nit below — then this is all ready to merge.

rajveermalviya:

gnprice:
Thanks! Looks good; merging :tada:

PR #1700, opened 2025-07-10
msglist: Support viewing who reacted to a message

### LINKED ISSUE #740: See who left an emoji reaction
I.e., this Zulip feature:
https://zulip.com/help/emoji-reactions#viewing-who-reacted-to-a-message

This is less critical than it would have been years ago, because we already show the names of who reacted when there are up to 3 of them. But once there are more, we just show counts, and the user may want to see specifically who left the reactions.

chrisbobbe:
Fixes #740.

Followup:

- #1747

Notable changes from previous revision:

- "See who reacted" button omitted in action sheet when no reactions
- Emojis centered horizontally when few
- Emojis scroll into view when tapped
- Reasonable UI with VoiceOver (at least in my testing)
- Wrote some basic tests

-----

### Screenshots

I've included some screenshots where there are
- many emoji
- many users

and in those, I set the scroll position such that you can see the shadow effect.

| Light | Dark |
| --- | --- |
| image | image |
| image | image |
| image | image |
| image | image |

Scroll-on-select animation:

Jul-31-2025 17-27-36


gnprice:
Thanks! I skimmed through this and I'm pretty confident it won't break any existing functionality. I also saw you demonstrate this running on your device, and it seemed to work well. So I plan to include this in today's upcoming release, without yet merging to main.

chrisbobbe:
Thanks! This is now ready for review, and I've updated the issue description with changes since the draft.

gnprice:
Thanks! Would you also post a few screenshots?

chrisbobbe:
Sure! Done.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    return Padding(
+      padding: const EdgeInsets.only(top: 16, bottom: 4),
+      child: InsetShadowBox(start: 8, end: 8,
+        color: designVariables.bgContextMenu,
+        child: SingleChildScrollView(
+          // TODO(upstream) we want to pass excludeFromSemantics: true
+          //    to the underlying Scrollable to remove an unwanted node
+          //    in accessibility focus traversal.
+          scrollDirection: Axis.horizontal,
+          physics: ClampingScrollPhysics(),
```
gnprice: Interesting — why this instead of the default?

(I believe the default would be this on Android, but `BouncingScrollPhysics` on iOS.)

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    return SizedBox(
+      width: double.infinity,
+      child: Column(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          ViewReactionsHeader(widget.pageContext,
+            messageId: widget.messageId,
+            reactionType: reactionType,
+            emojiCode: emojiCode,
+            onRequestSelect: (r) => _setSelection(r),
```
gnprice: nit: tearoff
```suggestion
            onRequestSelect: _setSelection,
```

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    });
+  }
+
+  void _storeChanged() {
+    _reconcile();
+  }
+
+  /// Check that the given reaction still has votes;
+  /// if not, select a different one if possible or clear the selection.
+  void _reconcile() {
```
gnprice, line 710: It looks like this always concludes with `_setSelection`. It might be a little clearer to read by making that explicit up front:
```suggestion
  void _reconcile() {
    _setSelection(_findMatchingReaction());
  }
```

Then _findMatchingReaction can directly return whenever it's found its answer.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+  void onNewStore() {
+    // TODO(#1747) listen for changes in the message's reactions
+    store?.removeListener(_storeChanged);
+    store = PerAccountStoreWidget.of(context);
+    store!.addListener(_storeChanged);
+    if (reactionType == null && widget.initialReactionType != null) {
+      assert(emojiCode == null);
+      assert(widget.initialEmojiCode != null);
+      reactionType = widget.initialReactionType!;
+      emojiCode = widget.initialEmojiCode!;
```
gnprice: This doesn't depend on the store, does it? It looks like it doesn't use the context at all; so it could run as early as initState.

And conversely it doesn't sound like it'd be desirable for this to run repeatedly, or after a new store — these are the *initial* emoji type and code, after all.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    // TODO could pull out this layout/appearance code,
+    //   focusing this widget only on state management
+    return SizedBox(
+      width: double.infinity,
+      child: Column(
+        mainAxisSize: MainAxisSize.min,
```
gnprice: Would this be equivalent?
```suggestion
    return Column(
      mainAxisSize: MainAxisSize.min,
      crossAxisAlignment: CrossAxisAlignment.stretch,
```

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+  const ViewReactionsHeader(
+    this.pageContext, {
+    super.key,
+    required this.messageId,
+    required this.reactionType,
+    required this.emojiCode,
+    required this.onRequestSelect,
+  });
+
+  final BuildContext pageContext;
```
gnprice: nit: this doesn't seem to get used

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    )?.userIds.toList();
+
+    // (No filtering of muted or deactivated users.)
+
+    if (userIds == null) {
+      // This reaction lost all its votes, or the message was deleted.
+      return SizedBox.shrink();
+    }
+
+    // TODO sort userIds?
```
gnprice: As is, I believe it comes in the order in which people left reactions. That seems as good an order as any.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+              role: SemanticsRole.tabBar,
+              container: true,
+              explicitChildNodes: true,
+              label: zulipLocalizations.seeWhoReactedSheetHeaderLabel(reactions.total),
+              child: Row(
+                children: reactions.aggregated.map((r) =>
+                  _ViewReactionsEmojiItem(
+                    reactionWithVotes: r,
+                    selected: r.reactionType == reactionType && r.emojiCode == emojiCode,
+                    onRequestSelect: () => onRequestSelect(r)),
```
gnprice: nit: can avoid making a new closure here:
```suggestion
                    reactionWithVotes: r,
                    selected: r.reactionType == reactionType && r.emojiCode == emojiCode,
                    onRequestSelect: onRequestSelect),
```

since each child is getting passed the same data `r` already, and so can pass that directly to the original callback if it gets invoked.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    final count = reactionWithVotes.userIds.length;
+
+    final emojiName = reactionWithVotes.emojiName;
+    final emojiDisplay = store.emojiDisplayFor(
+      emojiType: reactionWithVotes.reactionType,
+      emojiCode: reactionWithVotes.emojiCode,
+      emojiName: emojiName);
+
+    // Don't use a :text_emoji:-style display here.
+    final placeholder = SizedBox.fromSize(size: Size.square(emojiSize));
```
gnprice:
```suggestion
    final placeholder = SizedBox.square(dimension: emojiSize);
```

That'd be equivalent, right?

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+      message: emojiName,
+      child: GestureDetector(
+        behavior: HitTestBehavior.opaque,
+        onTap: () => _handleTap(context),
+        child: Container(
+          decoration: BoxDecoration(
+            border: Border.all(
+              // This border seems to affect the layout
+              // (I thought it was normally paint-only?),
+              // so always include a border, so positions don't shift.
```
gnprice: Looks like that mystery is answered at [[DecoratedBox]](https://main-api.flutter.dev/flutter/widgets/DecoratedBox-class.html):
> [Container](https://main-api.flutter.dev/flutter/widgets/Container-class.html) insets its child by the widths of the borders; this widget does not.

TIL. Might be another reason to skip Container and use specific widgets, here DecoratedBox and Padding.

assets/l10n/app_en.arb
```diff
@@ -136,6 +136,37 @@
...
   },
+  "actionSheetOptionSeeWhoReacted": "See who reacted",
+  "@actionSheetOptionSeeWhoReacted":  {
+    "description": "Label for the 'See who reacted' button in the message action sheet."
+  },
+  "seeWhoReactedSheetNoReactions": "This message has no reactions.",
+  "@seeWhoReactedSheetNoReactions":  {
+    "description": "Explanation on the 'See who reacted' sheet when the message has no reactions (because they were removed after the sheet was opened)."
+  },
+  "seeWhoReactedSheetHeaderLabel": "Emoji reactions ({num})",
```
gnprice: Perhaps "total"?
```suggestion
  "seeWhoReactedSheetHeaderLabel": "Emoji reactions ({num} total)",
```

Otherwise it kind of sounds like it's saying there are this many different reactions; at least I think that'd be my first guess as to the meaning.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    final store = PerAccountStoreWidget.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    final message = store.messages[messageId];
+
+    final userIds = message?.reactions?.aggregated.firstWhereOrNull(
+      (x) => x.reactionType == reactionType && x.emojiCode == emojiCode
+    )?.userIds.toList();
+
+    // (No filtering of muted or deactivated users.)
```
gnprice:
```suggestion
    // (No filtering of muted or deactivated users.
    //  Muted users will be shown as muted.)
```

My first reaction to this comment was "is that right? seems like we shouldn't show muted users here." Then I remembered that that's probably handled at a different layer, and indeed it looks like it is.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+
+    // TODO sort userIds?
+
+    Widget result = InsetShadowBox(
+      top: 8, bottom: 8,
+      color: designVariables.bgContextMenu,
+      child: SizedBox(
+        height: 400, // TODO(design) tune
+        child: ListView.builder(
+          padding: EdgeInsets.symmetric(vertical: 8),
```
gnprice: nit: move SizedBox outward, so the InsetShadowBox is right next to (in the source code) the padding that needs to match it

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+          itemCount: userIds.length,
+          itemBuilder: (context, index) =>
+            ViewReactionsUserItem(context, userId: userIds[index]))));
```
gnprice: Same comment as I told Sayed last week :slightly_smiling_face: : https://github.com/zulip/zulip-flutter/pull/1706#discussion_r2211840301

(on code that I guess was modeled in part on this PR, so no coincidence)

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+
+  final BuildContext pageContext;
+  final int userId;
+
+  void _onPressed() {
+    // Dismiss the action sheet.
+    Navigator.pop(pageContext);
+
+    Navigator.push(pageContext,
+      ProfilePage.buildRoute(context: pageContext, userId: userId));
```
gnprice: I think the reason this works even though this `pageContext` isn't really a page context is that there's no `await` here.

The reason our action-sheet buttons often need a real page context is that they'll dismiss the action sheet, then do something that takes time like a network request, and then after that completes they need to use a context to do something with the result, or to show an error. So if the request takes longer than the few hundred ms of the dismiss animation, the action sheet's own context may be unmounted by that point.

For just synchronously acting on the navigation like this, the widget's own context is fine.

lib/widgets/emoji_reaction.dart
```diff
@@ -614,3 +627,411 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return InkWell(
+      onTap: _onPressed,
+      splashFactory: NoSplash.splashFactory,
+      overlayColor: WidgetStateColor.resolveWith((states) =>
+        states.any((e) => e == WidgetState.pressed)
```
gnprice:
```suggestion
        states.contains(WidgetState.pressed)
```

Looks like it's a Set, so that should be equivalent (and potentially faster)

gnprice (commented):
Thanks!

Generally this looks good. Here's a full review except the tests.

lib/widgets/emoji_reaction.dart
chrisbobbe: Ah yeah I think the default is fine actually :)

I was reading a tab-bar implementation (when doing the Semantics work for this PR) and I think it opted out of the "bouncing" behavior. But I'm not sure this is really a "tab bar" in the same sense as that code I was reading.

lib/widgets/emoji_reaction.dart
chrisbobbe: Ah no actually, and I think this is how it is in the released revision 🙂—it start-aligns the emoji items instead of center-aligning them.

Looks like I can get the desired behavior with `CrossAxisAlignment.center` instead of stretch, though.

lib/widgets/emoji_reaction.dart
chrisbobbe: Ah but I think it should actually; we should be using this for `Navigator.push` to open a user's profile; will fix.

chrisbobbe: Oh actually I see your comment here about why `pageContext` isn't actually needed for that: https://github.com/zulip/zulip-flutter/pull/1700#discussion_r2232259547

chrisbobbe:
Thanks for the review! Revision pushed. I noticed on reaction chips that the tooltip-on-long-press was sometimes winning out over the show-view-reactions-sheet, so I added a commit to get rid of the tooltip, with a test change for some tests that had been relying on it.

I also added a tweak to the scroll-into-view logic for the horizontal list of emojis, as a separate commit on top.

chrisbobbe:
And I've just updated the screenshots.

lib/widgets/emoji_reaction.dart
```diff
@@ -194,74 +220,88 @@ class ReactionChip extends StatelessWidget {
...
+                        label))),
+                ]);
+              }))));
+
+    if (MediaQuery.accessibleNavigationOf(context)) {
+      // Avoid a _voterNames call with potentially many users,
+      // except in this case where a computation lag would be very small
+      // compared to the time saved in accessing the information.
+      result = Semantics(
+        label: _semanticsLabelWithVoterNames(store, zulipLocalizations),
```
gnprice: Hmm. I agree the computation time might not be a big deal here. But this feels like it's making the UI more complicated to think about — if the user is using TalkBack or VoiceOver, it's presenting different information that isn't present in the UI we regularly use ourselves. I'd prefer to avoid that complication.

Can this instead just present the same information that's there in the main UI? It's true that getting the list of people who reacted is then an extra step or two for people using a screen-reader; but it's an extra step or two for everyone else, too, so that doesn't seem terrible.

Here's what the Flutter doc on the relevant MediaQueryData field says:
```dart
  /// Whether the user is using an accessibility service like TalkBack or
  /// VoiceOver to interact with the application.
  ///
  /// When this setting is true, features such as timeouts should be disabled or
  /// have minimum durations increased.
```

So this conditional seems rather afield from that.

gnprice: (Or for now just present the same information that had been in the tooltip, namely the emoji name.)

lib/widgets/emoji_reaction.dart
gnprice: Sorry, I don't quite understand the first sentence — can you spell out which version does what? (I.e. in "this is how" and "it start-aligns", what do "this" and "it" refer to?) Is the behavior in the release different from what was in this PR?

lib/widgets/emoji_reaction.dart
```diff
@@ -595,3 +635,446 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+      .firstWhereOrNull((x) =>
+        x.reactionType == reactionType && x.emojiCode == emojiCode)
+      // first item will exist; early-return above on reactions.isEmpty
+      ?? reactions.first;
+  }
+
+  @override
+  void initState() {
+    super.initState();
+    if (reactionType == null && widget.initialReactionType != null) {
```
gnprice: The `reactionType == null` condition is always true now that this is in initState, right?

lib/widgets/emoji_reaction.dart
```diff
@@ -595,3 +635,446 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+  final ReactionType? reactionType;
+  final String? emojiCode;
+  final void Function(ReactionWithVotes) onRequestSelect;
+
+  /// A [double] between 0.0 and 1.0 for an emoji's position in the list.
+  ///
+  /// When auto-scrolling an emoji into view,
+  /// this is where the scroll position will land
+  /// (the min- and max- scroll extent lerped to this value).
+  double _emojiItemPosition(int index, int aggregatedLength) {
```
gnprice: lerped "at"?

Or some other/further rewording. I'd read "lerped to X" as meaning that X was the second argument of a `lerp` method — conceptually the arguments are start, end, and time.

lib/widgets/emoji_reaction.dart
```diff
@@ -595,3 +635,446 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+  /// - By animating, draws attention to the fact that this is a scrollable list
+  ///   and there may be more items in view. (In particular, does this when
+  ///   any item is tapped, because each item has a different [position].)
+  void _scrollIntoView(BuildContext context) {
+    final scrollPosition = Scrollable.of(context, axis: Axis.horizontal).position;
+    final destination = lerpDouble(
+      scrollPosition.minScrollExtent,
+      scrollPosition.maxScrollExtent,
+      position);
+    if (destination == null) return; // TODO(log)
```
gnprice: This condition is impossible, right? (From the definition of lerpDouble.) So just use `!`.

lib/widgets/emoji_reaction.dart
chrisbobbe: Ah sorry, yes that was confusing!

In [30.0.262](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2229437), the items are start-aligned:

image

Here's code pasted from the 30.0.262 tag:

```dart
    // TODO could pull out this layout/appearance code,
    //   focusing this widget only on state management
    return SizedBox(
      width: double.infinity,
      child: Column(
        children: [
          ViewReactionsHeader(widget.pageContext,
```

I want the items to be center-aligned, not start aligned. `Column`'s default `crossAxisAlignment` is `CrossAxisAlignment.center`, so I can accomplish that with

```dart
    return SizedBox(
      width: double.infinity,
      child: Column(
```

It turned out that the `SizedBox` isn't necessary, and I can spell out `CrossAxisAlignment.center` if that's helpful. So, as I've had in recent revisions of this PR:

```dart
    return Column(
```

which center-aligns the items.

chrisbobbe:
Thanks! Revision pushed.

lib/widgets/emoji_reaction.dart
```diff
@@ -595,3 +641,447 @@ class EmojiPickerListEntry extends StatelessWidget {
...
+    // (No filtering of muted or deactivated users.
+    //  Muted users will be shown as muted.)
+
+    if (userIds == null) {
+      // This reaction lost all its votes, or the message was deleted.
+      return SizedBox.shrink();
+    }
+
+    Widget result = SizedBox(
+      height: 400,
```
chrisbobbe: Oh oops, I meant to leave a `TODO(design) tune` here (I think I had it in an earlier revision)—will put it back in my next revision

lib/widgets/emoji_reaction.dart
gnprice: Very helpful, thanks! OK, so an earlier version of this PR (as seen in v30.0.262) had "stretch" alignment. That makes the children wide, including the header; and then I think the Row in the header, with defaults MainAxisSize.max and MainAxisAlignment.start, ends up putting the items at the start.

The revised version looks good.

gnprice:
Thanks! All looks good to me. Please go ahead and merge after that comment tweak.

chrisbobbe:
Done, thanks!

Issue #198, opened 2023-06-16
Set own user status


gnprice:
This is the flip side of #197, as part of this Zulip feature: https://zulip.com/help/status-and-availability
API docs: https://zulip.com/api/update-status

#197 is a prerequisite, so that we have the current value of the self-user's own status.

### Design

In the UI, this will be its own screen, reached from the "My profile" screen. For details, see Figma, starting here:
https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5000-52611&m=dev
and in the screens to the right of that.

chrisbobbe:
This came up in beta feedback today: [#mobile > Flutter Android 0.0.30 feedback @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/Flutter.20Android.200.2E0.2E30.20feedback/near/2188324)

chrisbobbe:
When this is done we should update the Help Center doc. We removed instructions for the legacy app when preparing to launch the new app: zulip/zulip@815a522b86a990f0729ecf79652cff34b42bb891

schewara:
Hi all

We also have many users in our Company complaining about this missing functionality and it would be nice if this would be implemented sooner then later.

sm-sayedi:
This came up in [#mobile > set user status](https://chat.zulip.org/#narrow/channel/48-mobile/topic/set.20user.20status/with/2195296) today and about a week before.

gnprice:
@schewara You may be glad to hear that this feature is added in the new release v30.0.262: [#announce > mobile releases @ 💬](https://chat.zulip.org/#narrow/channel/1-announce/topic/mobile.20releases/near/2229437). Please try it out and let us know what you think.

(The PR #1701 that implements it isn't yet merged to main, which is why we've kept this issue thread open; but it's included in the release.)

PR #1699, opened 2025-07-10
Show user status

gnprice:
This is the second half of @sm-sayedi's #1629:
https://github.com/zulip/zulip-flutter/pull/1629#issuecomment-3054609695

Sending this so these commits are in a PR, to help track what I'm including in today's release.

@sm-sayedi please go ahead and send these as your own PR (so you can update it), and then I'll close this one.

PR #1701, opened 2025-07-10
Set user status

sm-sayedi:
Support for setting the user status.

Figma design: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=5000-52611&t=ZkYDTwsnzsgZJRYr-0

### Profile Page

<table>
    <tr>
        <td><b>No status set</b></td>
        <td><b>Full status set</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
    <tr>
        <td><b>Only emoji set</b></td>
        <td><b>Only text set</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

### Set Status Page

<table>
    <tr>
        <td><b>No status set - Light</b></td>
        <td><b>No status set - Dark</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
    <tr>
        <td><b>Full status set - Light</b></td>
        <td><b>Full status set - Dark</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

### Screen recordings

Set status - Success

https://github.com/user-attachments/assets/5b72c70d-9ec0-47dc-bb9b-204616ce5100

Set status - Failure

https://github.com/user-attachments/assets/3c46156a-b661-4f2a-8f04-63667b57f3fc

Fixes: #198


gnprice:
Great!

Let's make this two PRs: one for the remaining portion of #1629 (i.e. same commits as #1699), and one for setting status. That will help keep the reviews more organized and will help us merge the first part earlier.

sm-sayedi:
Sure, on it.

assets/l10n/app_en.arb
```diff
@@ -809,6 +809,70 @@
...
   },
+  "status": "Status",
+  "@status": {
+    "description": "The status button label in self-user profile page when status is set."
+  },
+  "setStatus": "Set status",
+  "@setStatus": {
+    "description": "The status button label in self-user profile page when status is not set."
+  },
+  "noStatusText": "Not status text",
```
alya:
```suggestion
  "noStatusText": "No status text",
```

alya:
The screenshots look good to me, other than the small note above!

lib/widgets/button.dart
```diff
@@ -396,10 +398,23 @@ class ZulipMenuItemButton extends StatelessWidget {
...
+        child: Row(
+          spacing: 8,
+          crossAxisAlignment: CrossAxisAlignment.baseline,
+          textBaseline: localizedTextBaseline(context),
+          children: [
+            Text(label,
+              style: const TextStyle(fontSize: 20, height: 24 / 20)
+                .merge(weightVariableTextStyle(context, wght: _labelWght()))),
+            if (subLabel != null)
+              Text.rich(subLabel!,
```
gnprice: What happens when there isn't room for both of these? (Always a question with a `Row`.)

lib/widgets/profile.dart
```diff
@@ -119,6 +139,46 @@ class ProfilePage extends StatelessWidget {
   }
 }
 
+class _SetStatusButton extends StatelessWidget {
+  const _SetStatusButton({required this.userId});
+
+  final int userId;
```
gnprice: This is always the self-user ID, right? (Otherwise, it wouldn't make sense to *set* status.)

Seems clearest to not take it as a parameter, then; this widget's build method can look it up for itself.

assets/l10n/app_en.arb
```diff
@@ -809,6 +809,62 @@
   "@userRoleUnknown": {
     "description": "Label for UserRole.unknown"
   },
+  "status": "Status",
```
gnprice: This key is too general — there could easily be several places where we use the same string "Status" in English, and then the right translations might not be the same in some languages.

Instead, can say:
```suggestion
  "statusButton": "Status",
```

assets/l10n/app_en.arb
```diff
@@ -809,6 +809,62 @@
...
+  "@status": {
+    "description": "The status button label in self-user profile page when status is set."
+  },
+  "setStatus": "Set status",
+  "@setStatus": {
+    "description": "The status button label in self-user profile page when status is not set. Also, title for the page where user status is set."
```
gnprice: This "also" is a good sign that this should be two separate strings :slightly_smiling_face: even though they happen to have the same value in English.

For the first… perhaps `statusButtonAction`.

For the second, `setStatusPageTitle`. (See examples of other "…PageTitle" strings.)

assets/l10n/app_en.arb
```diff
@@ -809,6 +809,62 @@
...
+  "@statusSave": {
+    "description": "Label for the button that saves the user status."
+  },
+  "yourStatus": "Your status",
+  "@yourStatus": {
+    "description": "Hint text for the status text input field in setting user status page."
+  },
+  "userStatusBusy": "Busy",
+  "@userStatusBusy": {
+    "description": "Label for one of the suggested user statuses with status text 'Busy', in setting user status page."
```
gnprice: This isn't so much a label as a suggested actual value for the status. (If the user chooses it, it'll set their status text to the value of this string, right? Not to the English "Busy".)

So:
```suggestion
  "userStatusBusy": "Busy",
  "@userStatusBusy": {
    "description": "A suggested user status text, 'Busy'."
```

assets/l10n/app_en.arb
```diff
@@ -809,6 +809,62 @@
...
+  "@statusClear": {
+    "description": "Label for the button that clears the user status."
+  },
+  "statusSave": "Save",
+  "@statusSave": {
+    "description": "Label for the button that saves the user status."
+  },
+  "yourStatus": "Your status",
+  "@yourStatus": {
+    "description": "Hint text for the status text input field in setting user status page."
```
gnprice: We usually name this sort of string by its role, not its value:
```suggestion
  "statusTextHint": "Your status",
  "@statusTextHint": {
    "description": "Hint text for the status text input field in setting user status page."
```

The main reason is the point above about a key being too general — potentially we could use the same English string "Your status" for, say, a label when showing your own status, and then the translation in a given language might not be the same as the translation for this one.

lib/widgets/profile.dart
```diff
@@ -119,6 +139,46 @@ class ProfilePage extends StatelessWidget {
...
+    final localizations = ZulipLocalizations.of(context);
+    final store = PerAccountStoreWidget.of(context);
+    final userStatus = store.getUserStatus(userId);
+
+    return ZulipMenuItemButton(
+      style: ZulipMenuItemButtonStyle.list,
+      label: userStatus != UserStatus.zero
+        ? localizations.status : localizations.setStatus,
+      subLabel: userStatus != UserStatus.zero ? TextSpan(
+        children: [
```
gnprice: nit: put the zero case first, because that's the one that's quick to get out of the way (just `null`); that way the reader isn't in suspense about how that will turn out, and can move on to focus on the main case

lib/widgets/profile.dart
```diff
@@ -119,6 +139,46 @@ class ProfilePage extends StatelessWidget {
...
+            fontSize: 16,
+            textScaler: MediaQuery.textScalerOf(context),
+            position: StatusEmojiPosition.before,
+            neverAnimate: false,
+          ),
+          TextSpan(
+            text: userStatus.text ?? localizations.noStatusText,
+            style: TextStyle(
+              fontStyle: userStatus.text == null ? FontStyle.italic : null),
+          ),
```
gnprice: nit: this would probably benefit from having a single `userStatus.text == null` condition around the whole span

lib/widgets/profile.dart
```diff
@@ -119,6 +139,46 @@ class ProfilePage extends StatelessWidget {
...
+        ],
+      ) : null,
+      icon: ZulipIcons.chevron_right,
+      onPressed: () {
```
gnprice, line 170: At an intermediate commit, this says:
```dart
      icon: ZulipIcons.chevron_right,
      onPressed: () {},
```

That's a bad state: the button will look like it's tappable, it'll also report to a11y tools that it's tappable (because `onPressed` is non-null), but tapping won't do anything.

Instead, let's have it omit both the icon and onPressed at that stage. (You'll need a small NFC change to have ZulipMenuItemButton make onPressed optional.)

lib/widgets/content.dart
```diff
@@ -2053,23 +2062,25 @@ class UserStatusEmoji extends StatelessWidget {
...
         neverAnimate: neverAnimate));
   }
 
   @override
   Widget build(BuildContext context) {
     final store = PerAccountStoreWidget.of(context);
-    final emoji = store.getUserStatus(userId).emoji;
+    final effectiveEmoji = userId != null
+      ? store.getUserStatus(userId!).emoji
+      : emoji;
```
gnprice: nit:
```suggestion
    final effectiveEmoji = emoji ?? store.getUserStatus(userId!).emoji;
```

Bit shorter this way; and if we did end up with both `emoji` and `userId` non-null, I think the semantics I'd expect is this way around, where `emoji` took precedence for determining the emoji to show.

lib/widgets/action_sheet.dart
```diff
@@ -738,14 +738,22 @@ class ReactionButtons extends StatelessWidget {
...
     // TODO(design): have emoji picker slide in from right and push
     //   action sheet off to the left
 
     // Dismiss current action sheet before opening emoji picker sheet.
     Navigator.of(pageContext).pop();
 
-    showEmojiPickerSheet(pageContext: pageContext, message: message);
+    final emoji = await showEmojiPickerSheet(pageContext: pageContext);
+    if (emoji == null || !pageContext.mounted) return;
+    unawaited(doAddOrRemoveReaction(
```
gnprice, line 801: emoji: Make emoji picker return the selected emoji, for reuse
    
    Instead of using the selected emoji deep down the widget tree, simply
    return it where the emoji picker sheet is opened, to use it for
    different purposes.

This change is NFC, right? (Or is there a behavior change I'm missing?)

That's very helpful for reading the changes. So in that case let's mark it as such in the commit message.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+  final UserStatus oldStatus;
+
+  static AccountRoute<void> buildRoute({
+    required BuildContext context,
+    required UserStatus oldStatus,
+  }) {
+    return MaterialAccountWidgetRoute(
+      context: context,
+      page: SetStatusPage(oldStatus: oldStatus),
+    );
```
gnprice: nit: make compact; `page` is a "child"-like, payload argument:
```suggestion
    return MaterialAccountWidgetRoute(context: context,
      page: SetStatusPage(oldStatus: oldStatus));
```

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+  }
+
+  @override
+  State<SetStatusPage> createState() => _SetStatusPageState();
+}
+
+class _SetStatusPageState extends State<SetStatusPage> {
+  List<UserStatus> _statusSuggestions(ZulipLocalizations localizations) => [
+    UserStatus(text: localizations.userStatusBusy, emoji: StatusEmoji(emojiName: 'working_on_it', emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji)),
+    UserStatus(text: localizations.userStatusInAMeeting, emoji: StatusEmoji(emojiName: 'calendar', emojiCode: '1f4c5', reactionType: ReactionType.unicodeEmoji)),
```
gnprice: Let's avoid hard-coding emoji names here. See 34a562a09d8d441f5b6c662542a8859f69950651 fixing #1495, where doing the same thing with the "popular emoji" was a bug.

Instead, use the emoji code to look up the name.

If it helps simplify the code, go ahead and use the fact that `reactionType` is always `.unicodeEmoji` for these. That's unlikely to change; and if it does, we'll be editing this code anyway (since the list of emoji codes is hard-coded), so we can deal with it then.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+      page: SetStatusPage(oldStatus: oldStatus),
+    );
+  }
+
+  @override
+  State<SetStatusPage> createState() => _SetStatusPageState();
+}
+
+class _SetStatusPageState extends State<SetStatusPage> {
+  List<UserStatus> _statusSuggestions(ZulipLocalizations localizations) => [
```
gnprice: nit: put this method down below the state fields and the lifecycle methods; it's basically part of the work of the build method, so put it next to that

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+            }),
+        ],
+      ),
+      body: Column(children: [
+        Padding(
+          padding: const EdgeInsets.only(
+            // In Figma design, this is 16px, but we compensate for that in
+            // the icon button below.
+            left: 8,
+            top: 8, right: 10,
```
gnprice: left/right should be start/end (so it works correctly in RTL locales)

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+                    builder: (_, change, icon) {
+                      final emoji = switch(change.emoji) {
+                        OptionNone<StatusEmoji?>()           => oldStatus.emoji,
+                        OptionSome<StatusEmoji?>(:var value) => value,
+                      };
+                      return emoji != null
+                        ? UserStatusEmoji(emoji: emoji, size: 24, neverAnimate: false)
+                        : icon!;
+                    },
+                    child: Icon(ZulipIcons.smile, size: 24)),
```
gnprice: This icon doesn't really play the role of "child", since it commonly won't even appear at all.

The small optimization that comes from using ValueListenableBuilder.child here can be matched (and exceeded) by using `const`:
```suggestion
                        : const Icon(ZulipIcons.smile, size: 24);
                    }),
```

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+                  SizedBox(width: 4),
+                  Icon(ZulipIcons.chevron_down, size: 16),
+                ]),
+              ),
+              Expanded(child: TextField(
+                controller: statusTextController,
+                minLines: 1,
+                maxLines: 2,
```
gnprice, line 225: Is this different from omitting `minLines`?

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,338 @@
...
+                    // provide it in order to create a proper tap target size.
+                    horizontal: 8)),
+                icon: Row(children: [
+                  ValueListenableBuilder(
+                    valueListenable: statusChange,
+                    builder: (_, change, icon) {
+                      final emoji = switch(change.emoji) {
+                        OptionNone<StatusEmoji?>()           => oldStatus.emoji,
+                        OptionSome<StatusEmoji?>(:var value) => value,
+                      };
```
gnprice: nit:
```suggestion
                      final emoji = change.emoji.or(oldStatus.emoji);
```

That's equivalent, right?

gnprice (commented):
Thanks for building this! Comments below; and I see your PR description has a TODO for tests, which I believe is the only reason this is currently marked as a draft.

I've read through the first 5 commits:
a604e2260 button: Add ZulipMenuItemButton.subLabel
e3b091aaa profile: Add button for setting/showing user status in self-user profile
a90f83ea4 model [nfc]: Add UserStatusChange.copyWith method
0104cd1b9 content: Add `emoji` property to UserStatusEmoji widget
649232a19 emoji: Make emoji picker return the selected emoji, for reuse

and part of the 6th:
1ca2ceb4c user-status: Add page for setting own user status

lib/widgets/set_status.dart
sm-sayedi: Ahh, indeed; thanks!

lib/widgets/set_status.dart
sm-sayedi: Omitting `minLines` will make the field take `maxLines` height.

https://github.com/flutter/flutter/blob/e2441683275879df1ac0311e02ff868410599ab1/packages/flutter/lib/src/widgets/editable_text.dart#L1225-L1228

https://github.com/flutter/flutter/blob/e2441683275879df1ac0311e02ff868410599ab1/packages/flutter/lib/src/widgets/editable_text.dart#L1246-L1250

gnprice: I see, interesting. I looked at the doc on `minLines` before making my comment, and it seemed pretty unclear on this point, but didn't see those bits of the doc on `maxLines`.

It looks like the key logic in the implementation is this:
https://github.com/flutter/flutter/blob/e2441683275879df1ac0311e02ff868410599ab1/packages/flutter/lib/src/rendering/editable.dart#L2425-L2433
```dart
    final double preferredHeight = switch (maxLines) {
      null => math.max(_textPainter.height, preferredLineHeight * (minLines ?? 0)),
      1 => _textPainter.height,
      final int maxLines => clampDouble(
        _textPainter.height,
        preferredLineHeight * (minLines ?? maxLines),
        preferredLineHeight * maxLines,
      ),
    };
```

So when `maxLines` is non-null, the effective default for `minLines` is that it equals `maxLines`. When `maxLines` is null, though, the effective default for `minLines` is 0.

lib/widgets/button.dart
sm-sayedi: Thanks for catching. They would overflow. Fixed in the new revision, with both of them taking half the space and ellipsized.

lib/widgets/set_status.dart
sm-sayedi: Interesting piece of code. However, I think that, at least for me, it didn't make sense the first time to know that I had to specify `minLines`, for `maxLines` to work as the maximum number of lines.😀

sm-sayedi:
Thanks @gnprice for the review. New revision pushed. Please have a look.

lib/widgets/button.dart
```diff
@@ -393,13 +395,29 @@ class ZulipMenuItemButton extends StatelessWidget {
         foregroundColor: _labelColor(designVariables),
         splashFactory: NoSplash.splashFactory,
       ).copyWith(backgroundColor: _backgroundColor(designVariables)),
+      overflowAxis: Axis.vertical,
```
gnprice, line 398: Hmm interesting. What's the effect of this? I'm finding its doc a bit opaque :slightly_smiling_face:

lib/widgets/set_status.dart
gnprice: Yeah agreed — the behavior is a bit nonintuitive, and the docs aren't super clear about it.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+    statusChange.dispose();
+    super.dispose();
+  }
+
+  List<UserStatus> statusSuggestions(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final localizations = ZulipLocalizations.of(context);
+
+    final values = [
+      (localizations.userStatusBusy,            '1f6e0', 'working_on_it'),
```
gnprice: If the data we have from the server is missing one of these emoji, I think better to leave it out from the suggestions than to try to fall back to a hard-coded name. It's likely in that case that the hard-coded name won't work either.

(continuing from https://github.com/zulip/zulip-flutter/pull/1701#discussion_r2211785018)

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+      (localizations.userStatusAtTheOffice,     '1f3e2', 'office'),
+    ];
+
+    return [
+      for (final (statusText, emojiCode, fallbackEmojiName) in values)
+        UserStatus(
+          text: statusText,
+          emoji: StatusEmoji(
+            emojiName: store.allEmojiCandidates()
+                         .firstWhereOrNull((e) => e.emojiCode == emojiCode)
```
gnprice: Hmm, this iterates through the whole list of all known emoji, repeatedly for each of the suggestions we want to offer. Let's avoid doing that. :slightly_smiling_face: 

Instead, have a prep commit add a little method to EmojiStore to offer the API we need in order to do this efficiently. I think that can look like `String? getUnicodeEmojiNameByCode(String emojiCode)`.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+                    }),
+                  SizedBox(width: 4),
+                  Icon(ZulipIcons.chevron_down, size: 16),
+                ]),
+              ),
+              Expanded(child: TextField(
+                controller: statusTextController,
+                minLines: 1,
+                maxLines: 2,
+                maxLength: 60,
```
gnprice, line 228: Ah interesting. Where does this max length come from?

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+                cursorColor: designVariables.textInput,
+                textCapitalization: TextCapitalization.sentences,
+                style: TextStyle(fontSize: 19, height: 24 / 19),
+                decoration: InputDecoration(
+                  counterText: '', // TODO: should we show counter?
```
gnprice: Sounds like a question for `#mobile-design` :-)

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+            ]),
+        ),
+        Expanded(child: InsetShadowBox(
+          top: 6, bottom: 6,
+          color: designVariables.mainBackground,
+          child: SingleChildScrollView(
+            physics: AlwaysScrollableScrollPhysics(), // TODO: necessary?
+            padding: EdgeInsets.symmetric(vertical: 6),
+            child: Column(children: [
+              for (final status in statusSuggestions(context))
```
gnprice: Let's pull this out as a local variable: `final suggestions = statusSuggestions(context);`, and then this iterates through that local.

That way the call to this helper method is a bit more visible when reading this build method's logic.

lib/api/route/users.dart
```diff
@@ -34,6 +34,18 @@ class GetOwnUserResult {
   Map<String, dynamic> toJson() => _$GetOwnUserResultToJson(this);
 }
 
+/// https://zulip.com/api/update-status
+Future<void> updateStatus(ApiConnection connection, {
```
gnprice, line 39: nit: add API bindings in a separate commit

lib/api/route/users.dart
```diff
@@ -34,6 +34,18 @@ class GetOwnUserResult {
...
+  required UserStatus status,
+}) {
```
gnprice: In the server API, both the text and the emoji are optional — either can be omitted and then only the other one will be set. So `UserStatusChange` seems like a closer fit for that.

gnprice (commented):

Like in the previous round above, I've read the first N-1 commits:
88f8299bb button: Add ZulipMenuItemButton.subLabel
03d399900 button [nfc]: Make ZulipMenuItemButton.onPressed optional
f65ee2536 profile: Add button for setting/showing user status in self-user profile
5d50ee561 model [nfc]: Add UserStatusChange.copyWith method
c9b5972ef content: Add `emoji` property to UserStatusEmoji widget
4feb05b96 emoji [nfc]: Make emoji picker return the selected emoji, for reuse

and part of the last:
ea1d7111f user-status: Add page for setting own user status

lib/widgets/button.dart
sm-sayedi: Normally, the child widget of `MenuItemButton` would overflow horizontally, even if it's a simple widget like `Text` with a longer string. Setting `overflowAxis: Axis.vertical` wraps the child in an `Expanded` widget internally, thus preventing the overflow.

https://github.com/flutter/flutter/blob/440713c3b2878048be2661a6705b219c51f72c95/packages/flutter/lib/src/material/menu_anchor.dart#L2821-L2858

lib/widgets/set_status.dart
sm-sayedi: > Note: The limit on the size of the message is 60 characters.

It's in the [updateStatus](https://zulip.com/api/update-status#parameter-status_text) API docs, and I forgot to include it in the previous revisions.😀

lib/widgets/set_status.dart
sm-sayedi: Asked: [#mobile-design > Set user status @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Set.20user.20status/near/2224547).

lib/widgets/button.dart
gnprice: What would "overflow horizontally" mean concretely?

I guess the other part I'm trying to understand is: the argument sounds like it means the button now overflows vertically, instead of horizontally. What does that look like?

lib/widgets/set_status.dart
gnprice: Cool, thanks. This comment can link to that, then.

lib/widgets/set_status.dart
gnprice: Aha, I see. I guess let's have a short comment here with that link. That way it's clear why that's there and has the value it has; and, crucially, it's easy for someone reading in the future to check whether 60 is still the right answer then.

lib/widgets/button.dart
sm-sayedi: By "overflow horizontally", I meant that it shows that striped pattern:
Screenshot 2025-07-19 at 8 53 07 PM
And when `overflowAxis: Axis.vertical` is set, it will expand vertically:
Screenshot 2025-07-19 at 8 56 41 PM
Now `Text.overflow` will also work:
Screenshot 2025-07-19 at 11 01 41 PM

For more context, here's the upstream PR that added `MenuItemButton.overflowAxis`: https://github.com/flutter/flutter/pull/143932

sm-sayedi:
Thanks @gnprice for the review. New revision pushed.

lib/widgets/button.dart
gnprice: Cool, thanks. The name still seems confusing to me (since the two behaviors don't seem at all symmetric), but that explanation is helpful.

It looks like this comment was the origin of the name that got merged: https://github.com/flutter/flutter/pull/143932#discussion_r1541600556
but the previous drafts already had the "direction" naming that doesn't seem to match what it really does.

gnprice:
Thanks! Those changes all look good.

As you mention in the description, this PR still needs some tests. :-) There's also more for me to read in the final/main commit.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+  void initState() {
+    super.initState();
+    statusTextController = TextEditingController(text: oldStatus.text)
+      ..addListener(() {
+        final trimmedValue = statusTextController.text.trim();
+        final text = trimmedValue.isNotEmpty ? trimmedValue : null;
+
+        // Ignore updating [statusChange] for the additional updates with the
+        // same value from TextField. For example, when a character is deleted.
+        if (text == newStatus.text) return;
```
gnprice: I don't understand this comment. When a character is deleted, that's a new value for the text, right?

I believe the point here is for when something other than the text changes: namely the selection, or the composing range.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+          // Ignore updating the status text field if it already has the same
+          // text. It can happen in the following cases:
+          //   1. Only the emoji is changed.
+          //   2. The same status is chosen consecutively from the suggested
+          //      statuses list.
+          //   3. This listener is called as a result of the change in status
+          //      text field.
+          if (text == statusTextController.text) return;
+
+          statusTextController.text = text;
```
gnprice, line 75: Hmm yeah, this structure is doing something a bit delicate — there's two `ChangeNotifier`s here that both have a copy of the same data, namely the text the user has entered in the text field, and they're updating each other in both directions. So it ends up needing to carefully avoid an infinite loop of updating each other.

gnprice: This version works, and it's not necessary to rework it. But FWIW I think my solution here would have been:
* Give each piece of data a single source of truth, responsible for notifying listeners on changes; the text lives in the TextEditingController, so that's what things listen to for updates to the text.
* Then for the emoji, put it directly on the state and use `setState` when updating.

The main thing this design is accomplishing relative to that one is that it avoids rebuilding the overall page when the text or emoji changes, instead using ValueListenableBuilder on the few smaller widgets that actually need to update. Those rebuilds are probably OK — I think this build method isn't doing anything all that expensive — but it's true that it's nice to avoid them.

A variation of my above solution which would accomplish that is:
* For the emoji, give it its own ValueNotifier (and don't call setState on updates).
* Then to avoid having to double-wrap with ValueListenableBuilder (for both the text and emoji), say something like `emojiAndTextListenable = Listenable.merge([statusTextController, statusEmoji])` to make one listenable for both pieces, and use ListenableBuilder with that.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+    super.dispose();
+  }
+
+  List<UserStatus> statusSuggestions(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
+    final localizations = ZulipLocalizations.of(context);
+
+    final values = [
+      (localizations.userStatusBusy,            '1f6e0'),
+      (localizations.userStatusInAMeeting,      '1f4c5'),
```
gnprice: Thanks for nicely aligning these columns. One trick that can make that simpler: when one of the columns is all the same width, like the emoji codes here, put that column first. Then the subsequent columns are naturally aligned.

(Or when one of the columns is nearly a constant width, put it first, and then less spacing is required to align the next column.)

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+
+  void chooseStatusSuggestion(UserStatus status) {
+    statusChange.value = UserStatusChange(
+      text: toOption(recent: status.text, old: oldStatus.text),
+      emoji: toOption(recent: status.emoji, old: oldStatus.emoji)
+    );
+  }
+
+  Option<T> toOption<T>({required T recent, required T old}) =>
+    recent == old ? OptionNone() : OptionSome(recent);
```
gnprice: "to option" seems much more general than what this is really about — the meaning of this is specific to the "change" concept found in UserStatusChange.

How about `asChange`?

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  void chooseStatusSuggestion(UserStatus status) {
+    statusChange.value = UserStatusChange(
+      text: toOption(recent: status.text, old: oldStatus.text),
+      emoji: toOption(recent: status.emoji, old: oldStatus.emoji)
+    );
+  }
+
+  Option<T> toOption<T>({required T recent, required T old}) =>
```
gnprice: "recent" doesn't feel right for this parameter — it suggests this is a value that has already become the actual value, and in fact has maybe been replaced by some other value since then.

The right name would be "new". That's a keyword, but could call this `new_` — compare `TopicNarrow.with_`.

I think it'd actually make most sense for it to be positional: it's the direct argument which this is interpreting, and the other argument is more of a background fact. So (including my other comment about the method's own name):

```suggestion
  Option<T> asChange<T>(T new_, {required T old}) =>
```

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+    recent == old ? OptionNone() : OptionSome(recent);
+
+  @override
+  Widget build(BuildContext context) {
+    final designVariables = DesignVariables.of(context);
+    final localizations = ZulipLocalizations.of(context);
+    final suggestions = statusSuggestions(context);
+
+    return Scaffold(
```
gnprice: nit: this is a different kind of thing from the steps above (they're quick O(1) data lookups, while this is doing a bit of computation of its own), so set it off with a blank line:
```suggestion
    final designVariables = DesignVariables.of(context);
    final localizations = ZulipLocalizations.of(context);

    final suggestions = statusSuggestions(context);

    return Scaffold(
```

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+      body: Column(children: [
+        Padding(
+          padding: const EdgeInsetsDirectional.only(
+            // In Figma design, this is 16px, but we compensate for that in
+            // the icon button below.
+            start: 8,
+            top: 8, end: 10,
+            // In Figma design, this is 4px, be we compensate for that in
+            // [SingleChildScrollView.padding] below.
+            bottom: 0),
```
gnprice, line 195: Hmm but that doesn't do the same thing, does it? Once you scroll, that other padding scrolls away.

Why not do this the same way as in the design?

… Ah, the point is that there's also an InsetShadowBox for those same 6px. Which requires the scroll view to handle the padding; and also means that even when scrolled, the shadow keeps the scroll content visually separated from the text field.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+                minLines: 1,
+                maxLines: 2,
+                // The limit on the size of the status text is 60 characters.
+                // See: https://zulip.com/api/update-status#parameter-status_text
+                maxLength: 60,
+                cursorColor: designVariables.textInput,
+                textCapitalization: TextCapitalization.sentences,
+                style: TextStyle(fontSize: 19, height: 24 / 19),
+                decoration: InputDecoration(
+                  counterText: '',
```
gnprice, line 235: include a 1- or 2-line comment with a link to the thread about this (per https://github.com/zulip/zulip-flutter/pull/1701#discussion_r2216825570)

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+                    borderRadius: BorderRadius.circular(10),
+                    borderSide: BorderSide.none,
+                  )))),
+            ]),
+        ),
+        Expanded(child: InsetShadowBox(
+          top: 6, bottom: 6,
+          color: designVariables.mainBackground,
+          child: SingleChildScrollView(
+            physics: AlwaysScrollableScrollPhysics(), // TODO: necessary?
```
gnprice: How might a future reader resolve this TODO — what are you uncertain about in figuring out whether this is necessary?

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,339 @@
...
+        (states) => states.any((e) => e == WidgetState.pressed)
+          ? designVariables.contextMenuItemBg.withFadedAlpha(0.20)
+          : Colors.transparent,
+      ),
+      child: Padding(
+        padding: EdgeInsets.symmetric(vertical: 7, horizontal: 16),
+        child: Row(
+          spacing: 8,
+          children: [
+            UserStatusEmoji(emoji: status.emoji!, size: 19),
```
gnprice: These should be 24px in size, according to the Figma design.

gnprice: In particular that makes these candidate emoji the same size as the chosen emoji at the top. See the "Full status set" screenshots, where these are smaller, and compare with the corresponding screen in Figma, where they're the same size.

gnprice (commented):
OK, and I've finished reading all but the tests in the last commit. Comments below.

The only behavior change that these comments call for is in the size of the candidate emoji. So I plan to include this in today's release; I'll add a squashable commit to fix that emoji size, unless you're online and have made a revision by then.

gnprice:
(I rebased atop the merged version of #1702, and then atop main past the refactors in #1735 and #1736.)

gnprice:
Added two commits at the end:
862b48a23 (squash) user-status: Update a test on profile page
9d3abbd6e (squash) user-status: Adjust candidate emoji size to 24 on set-status screen

lib/widgets/set_status.dart
sm-sayedi: When a character is deleted, the listener is called twice with the same value. But the point you mentioned is also an important case for that. So I will include it in the comment.

lib/widgets/set_status.dart
sm-sayedi: When I was first writing this method, I wanted to use `new_`, but I thought it may not look good as a named parameter. But now that you suggested it to be a positional parameter, looks good.

lib/widgets/set_status.dart
sm-sayedi: In the Figma design, there is in fact no shadow. But I thought it may look better when scrolled, especially that now we use the shadow in multiple places in UI. 🙂

lib/widgets/set_status.dart
sm-sayedi: I think a better place for me to write that comment was in a GitHub comment. What I meant was that in most devices, the suggestions list occupies a portion of the screen height, so it doesn't provide that nice scroll animation when dragged. So will it be better to make it always scrollable?! We have a somewhat similar case in the "See who reacted" sheet, where if there is only one reactor, when the list is dragged, it has the scroll animation.

lib/widgets/set_status.dart
gnprice: I see. My guess would be that when that happens, it's because one of those other properties changed too — and the listener is getting one call for the text changing, and another for the selection or the composing range changing. (Or perhaps one of the latter is changing at the same time as the text, then promptly changing back.)

lib/widgets/set_status.dart
gnprice: Yeah.

FWIW I think even with the name visible to callers, it'd be better to have a name like `new_` that has the right word with a bit of funny-looking punctuation than a name like `recent` where the word doesn't mean the right thing.

lib/widgets/set_status.dart
gnprice: Cool. Yeah, I agree, a GitHub comment is a good place for that question.

And then now that we're here in a GitHub subthread, I think actually the ideal place for it is in #mobile-design :slightly_smiling_face: — that way Alya or Vlad or others can readily see it and chime in.

lib/widgets/set_status.dart
sm-sayedi: Yeah, in Figma the box size of those emojis is 24px. But looking at the "Typography" section of the "Properties" panel, there are two properties with different values: `Size: 19px` and `Line height: 24px`. The same property values are for the actual text of status suggestions. So I thought I would use 19px as the size of the emoji, as that's what we did in other parts of UI where status emoji is shown, so we gave it the same size as the text it appeared with.

Screenshot 2025-07-26 at 4 49 36 AM

And for the chosen emoji in the top, in the before-current revision of the Figma design, it was 24px, the same as the "smile" placeholder icon. But in the new revision, the placeholder icon stayed 24px, and the selected emoji size became 19px. But even then, I preferred to use the previous 24px size for the emoji too, to align it with the placeholder icon. 🙂

lib/widgets/set_status.dart
sm-sayedi: Yeah, this makes sense.

lib/widgets/set_status.dart
gnprice: Right, so UserStatusEmoji.size (and UnicodeEmojiWidget.size) specify the size the emoji's box will have in the layout. That corresponds to the 24x24 square in that screenshot, so `size: 24`.

The "Typography" section says "Size: 19px" referring to a font size. But the font size to use to get the desired box size depends on the font. See the implementation of UnicodeEmojiWidget — it'll use a font size of 24px with Apple Color Emoji, and (14.5/17) * 24px =~ 20.5px with Noto Color Emoji (so on iOS and Android respectively). I'm not sure what font is being used for the emoji when Figma renders the design.

lib/widgets/set_status.dart
sm-sayedi: Posted: [#mobile-design > Set user status @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/Set.20user.20status/near/2230333).

lib/widgets/set_status.dart
sm-sayedi: Hmm. With a 24px size for the status suggestion emoji, it fills the 24x24 box entirely, unlike the design in Figma, which is smaller than that due to `size: 19px`. 
Screenshot 2025-07-26 at 6 55 55 AM

In other places of UI where we show emoji, we used the font size of the neighboring text for the emoji size, instead of its line height. For example:
https://github.com/zulip/zulip-flutter/blob/11ffac283f058542d96f6e10d7518dc83c2338b3/lib/widgets/message_list.dart#L1920-L1934

https://github.com/zulip/zulip-flutter/blob/11ffac283f058542d96f6e10d7518dc83c2338b3/lib/widgets/new_dm_sheet.dart#L419-L428

Or maybe to stay consistent and use line height for the emoji size, we may consider changing other places in the UI to also use the same criterion. 🙂

gnprice: This would be a good question to take to #mobile-design 🙂

gnprice: (with screenshots of the different alternatives)

sm-sayedi:
Thanks @gnprice for all the previous reviews. This is now out of Draft with tests included. Please have a look.

~Note: The CI is failing because of the Flutter upgrade in #1755 which this PR is rebased on.~

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,545 @@
...
+        await tester.pump();
+
+        final connection = store.connection as FakeApiConnection;
+        connection.prepare(json: {}, delay: Duration(milliseconds: 100));
+
+        checkButtonEnabled(tester, saveButtonFinder, expected: true);
+        await tester.tap(saveButtonFinder);
+        await testNavObserver.pumpPastTransition(tester);
+        // check(currentRoute).isNotNull().isA<MaterialAccountWidgetRoute>()
+        //   .page.isA<ProfilePage>();
```
sm-sayedi: This commented-out check is not working for no obvious reason. It shows the "Profile page" route to be a `MaterialPageRoute` instead of a `MaterialAccountWidgetRoute`.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,346 @@
...
+              style: TextStyle(fontSize: 19, height: 24 / 19),
+              maxLines: 1,
+              overflow: TextOverflow.ellipsis)),
+          ])),
+    );
+  }
+}
+
+@visibleForTesting
+Map<String, String> statusCodesToText(ZulipLocalizations localizations) {
```
gnprice: This name sounds more general than what the function actually is about. That makes it potentially confusing when it appears in the global namespace.

In general a good way to address that is by making it a method on some relevant class, which effectively puts the name into a namespace that clarifies the context. Here, a good class would be SetStatusPage.

test/widgets/set_status_test.dart
gnprice: Well, where does this route come from, and why would it be one or the other?

It looks like this will be the route that `setupPage` put there with TestZulipApp, in this code:
```dart
    await tester.pumpWidget(TestZulipApp(
      accountId: eg.selfAccount.id,
      navigatorObservers: [testNavObserver, ?navigatorObserver],
      child: ProfilePage(userId: eg.selfUser.userId)));
```

I think you should be able to get the intended effect of this check by just looking for a widget, not at routes: like `check(find.byType(ProfilePage)).findsOne()`.

lib/model/emoji.dart
```diff
@@ -396,6 +402,11 @@ class EmojiStoreImpl extends PerAccountStoreBase with EmojiStore {
     return _allEmojiCandidates ??= _generateAllCandidates();
   }
 
+  @override
+  String? getUnicodeEmojiNameByCode(String emojiCode) =>
+    // TODO(log) if null; not supposed to happen
+    _serverEmojiData?[emojiCode]?.firstOrNull;
```
gnprice: Hmm. Well, why is it not supposed to happen? :slightly_smiling_face: 

I think there are actually three different answers for the three different ways this can be null.
* _serverEmojiData still null: this means this method is being called when we don't yet have the server emoji data. That's something that can happen, but it's a problem if it happens often; so good to log. Compare _generateAllCandidates.
* `emojiCode` isn't a key there. Given the way this method is used, that means the hard-coded list of suggested status emoji is out of date. Definitely would want to log that. Similar to a case in _generatePopularCandidates.
* `emojiCode` is present, but the value is an empty list. That is an error by the server. The format of this data is documented at https://zulip.com/api/register-queue under `server_emoji_data_url`; and
  > each value is the corresponding [emoji names](https://zulip.com/api/add-reaction#parameter-emoji_name) for this emoji, with the canonical name for the emoji always appearing first.

  So there must be a first value.

  I think we can just ignore the case where the list is empty; in general we're not trying to be a validator for the server's conformance to the Zulip API, and this is an unlikely bug for the server to have. On the ["crunchy shell" principle](https://github.com/zulip/zulip-mobile/blob/main/docs/architecture/crunchy-shell.md), we could also have ServerEmojiData.fromJson enforce that.

test/model/emoji_test.dart
```diff
@@ -331,6 +331,37 @@ void main() {
...
+      });
+      check(store.getUnicodeEmojiNameByCode('1f34a')).isNull();
+    });
+
+    test('server emoji data present, emoji code corresponds to empty emoji name list', () {
+      final store = prepare(unicodeEmoji: {
+        '1f4c5': [],
+      });
+      check(store.getUnicodeEmojiNameByCode('1f516')).isNull();
+    });
```
gnprice: In particular let's not have this test case; an empty list shouldn't reach this code in the first place.

test/test_navigation.dart
```diff
@@ -1,11 +1,12 @@
 import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
 
 // Inspired by test code in the Flutter tree:
 //   https://github.com/flutter/flutter/blob/53082f65b/packages/flutter/test/widgets/observer_tester.dart
 //   https://github.com/flutter/flutter/blob/53082f65b/packages/flutter/test/widgets/navigator_test.dart
 
 /// A trivial observer for testing the navigator.
-class TestNavigatorObserver extends NavigatorObserver {
+class TestNavigatorObserver extends TransitionDurationObserver{
```
gnprice, line 9: nav test: Make TestNavigatorObserver extend TransitionDurationObserver
    
    This way, we can use the instance of TestNavigatorObserver to wait for
    the route transition to complete, instead of creating a different object
    of TransitionDurationObserver for the same purpose.

Cool, good idea.

test/test_navigation.dart
```diff
@@ -36,6 +39,7 @@ class TestNavigatorObserver extends NavigatorObserver {
 
   @override
   void didReplace({ Route<dynamic>? oldRoute, Route<dynamic>? newRoute }) {
+    super.didReplace();
```
gnprice: Let's have all our overrides in this class call super; that way we don't have to wonder if we're potentially missing one that should.

test/example_data.dart
```diff
@@ -166,16 +166,9 @@ final ServerEmojiData serverEmojiDataPopular = _immutableServerEmojiData(codeToN
...
-  final a = serverEmojiDataPopular;
-  final b = data;
-  final result = ServerEmojiData(
-    codeToNames: {...a.codeToNames, ...b.codeToNames},
-  );
-  assert(
-    result.codeToNames.length == a.codeToNames.length + b.codeToNames.length,
-    'eg.serverEmojiDataPopularPlus called with data that collides with eg.serverEmojiDataPopular',
+  return ServerEmojiData(
+    codeToNames: {...data.codeToNames, ...serverEmojiDataPopular.codeToNames},
```
gnprice: The assert was for not letting the coming emoji data collide with
    popular emoji data. This change will make it easier for callers to
    provide data of any emojis. If the data contains one/more of popular
    emojis, the data from serverEmojiDataPopular will replace it, as it is
    the source of truth for popular emojis in test code.

Hmm. That sounds like a recipe for someone to have a confusing debugging session: they specifically said such-and-such emoji should have such-and-such names, so why does just that one emoji have a different set of names?

In the new version, what this function does is very simple. If there are callers that really want to permit this sort of collision, how about just having them do this directly? That way at least it's transparent at the call site exactly what sources of data are being combined, and how one of them overrides the other one.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+
+  testWidgets('set status page renders', (tester) async {
+    await setupPage(tester, emojiData: suggestedEmojiData);
+
+    check(find.text('Set status')).findsOne();
+    check(clearButtonFinder).findsOne();
+    check(saveButtonFinder).findsOne();
+    check(findEmojiButton()).findsOne();
+    check(findStatusTextField()).findsOne();
+    checkSuggestionsVisible(suggestedUnicodeEmoji.keys.toList());
```
gnprice: I think this is clearer if it's spelled out with a couple of examples:
```suggestion
    check(findSuggestion(code: '1f6e0', text: 'Busy')).findsOne();
    check(findSuggestion(code: '1f4c5', text: 'In a meeting')).findsOne();
```

Doesn't have to be all of them. The canonical list of all of them is the one in the implementation, and I don't think there's much value in the test checking that the same list was used.

Then that also means `checkSuggestionsVisible` isn't needed, and `statusCodesToText` can become private again (e.g. a private method on the state class, right near where it's used), which would be nice.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+    addTearDown(testBinding.reset);
+
+    Route<dynamic>? currentRoute;
+    final testNavObserver = TestNavigatorObserver()
+      ..onPushed = (route, _) => currentRoute = route;
+
+    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+    await store.addUser(eg.selfUser);
+    await store.changeUserStatus(eg.selfUser.userId, change);
```
gnprice: With the default value of `change`, the server wouldn't be expected to send such an event, right? So it'd be more realistic to skip this step when the caller didn't pass a nontrivial change.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+
+    await tester.tap(find.widgetWithText(ZulipMenuItemButton,
+      switch (change) {
+        UserStatusChange(text: OptionNone(), emoji: OptionNone())
+          => 'Set status',
+        _ => 'Status',
+      }));
+    await tester.pump();
+    await testNavObserver.pumpPastTransition(tester);
+    check(currentRoute).isNotNull().isA<MaterialAccountWidgetRoute>()
```
gnprice: nit: `isNotNull` redundant with `isA`
```suggestion
    check(currentRoute).isA<MaterialAccountWidgetRoute>()
```

(here and other examples below)

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+          emoji: OptionSome(StatusEmoji(emojiName: 'slight_smile',
+            emojiCode: '1f642', reactionType: ReactionType.unicodeEmoji))));
+
+        checkButtonEnabled(tester, clearButtonFinder, expected: true);
+        checkButtonEnabled(tester, saveButtonFinder, expected: false);
+      });
+    });
+
+    group('edit status', () {
+      Future<void> chooseEmojiFromPicker(WidgetTester tester, String code, {
```
gnprice:
```suggestion
      Future<void> chooseUnicodeEmojiFromPicker(WidgetTester tester, String code, {
```

That makes this function's assumptions clear — it relies on this being a Unicode emoji.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+        text: OptionSome('Happy'),
+        emoji: OptionSome(StatusEmoji(emojiName: 'slight_smile',
+          emojiCode: '1f642', reactionType: ReactionType.unicodeEmoji))));
+
+      check(findEmojiButton(emojiSelected: true)).findsOne();
+      check(findStatusTextField()).findsOne();
+      check(switch(tester.widget<TextField>(findStatusTextField())) {
+        TextField(controller: TextEditingController(text: 'Happy')) => true,
+        _                                                           => false,
+      }).isTrue();
```
gnprice: This can be a lot clearer by being more direct :slightly_smiling_face: :
```suggestion
      check(tester.widget<TextField>(findStatusTextField()).controller!.text)
        .equals('Happy');
```

gnprice: Also then the `findsOne` check above for the same finder seems unnecessary — it's implied by `tester.widget` on that finder not throwing.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+
+      check(findEmojiButton(emojiSelected: false)).findsOne();
+      check(findStatusTextField()).findsOne();
+      check(switch(tester.widget<TextField>(findStatusTextField())) {
+        TextField(controller: TextEditingController(text: '')) => true,
+        _                                                      => false,
+      }).isTrue();
+    });
+
+    group('"Save" button returns to Profile page, saves the status', () {
```
gnprice, line 401: The description says "saves the status", and I agree that's a good thing to test. But I don't see anything that actually checks for that.

In fact if I delete that line of the implementation:
```diff
     try {
-      await updateStatus(store.connection, change: statusChange.value);
     } catch (e) {
```

it looks like none of the tests fail.

You'll want FakeApiConnection.takeRequests, or lastRequest; look at their call sites in existing widget tests for examples.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+          emojiData: suggestedEmojiData, navigatorObserver: testNavObserver);
+
+        await tester.tap(findSuggestion(code: '1f6e0', text: 'Busy'));
+        await tester.pump();
+
+        final connection = store.connection as FakeApiConnection;
+        connection.prepare(json: {}, delay: Duration(milliseconds: 100));
+
+        checkButtonEnabled(tester, saveButtonFinder, expected: true);
+        await tester.tap(saveButtonFinder);
```
gnprice: The "enabled" check is redundant:
```suggestion
        await tester.tap(saveButtonFinder);
```

If it weren't, then the tap wouldn't have any effect; and the next lines after this check for effects it should have had.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+        // check(currentRoute).isNotNull().isA<MaterialAccountWidgetRoute>()
+        //   .page.isA<ProfilePage>();
+
+        await store.changeUserStatus(eg.selfUser.userId, UserStatusChange(
+          text: OptionSome('Busy'),
+          emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+            emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji))));
+        await tester.pump();
+        check(find.text('\u{1f6e0}')).findsOne();
+        check(findText('Busy', includePlaceholders: false)).findsOne();
```
gnprice: This doesn't really check anything about the set-status page. Instead it's effectively repeating a test you've added separately for the profile page. So it can be left out.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+        await tester.tap(find.descendant(of: find.byType(EmojiPicker),
+          matching: find.text(tryParseEmojiCodeToUnicode(code)!)));
+        await tester.pump();
+        await navObserver.pumpPastTransition(tester);
+        check(currentRoute()).isNotNull().isA<MaterialAccountWidgetRoute>()
+          .page.isA<SetStatusPage>();
+      }
+
+      group('no status set, buttons are disabled', () {
+        testWidgets('emoji is added -> buttons are enabled', (tester) async {
```
gnprice, line 191: There are a lot of test cases in this group and the next one; and they feel pretty repetitive, totalling about 260 lines of code. It feels like a lot to read compared to the complexity of what it's testing, so I skimmed (until I got to the test cases for what the buttons do) rather than spend the time to read them all fully.

I don't think this area is critical to test, so let's not spend effort really polishing this set of tests; it's OK for them to go in while being somewhat repetitive and hard to read, because it'd be OK for this to go in without having most of this set of tests. But I'll leave one or two comments where I think it might be easy to cut them down.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+          checkButtonEnabled(tester, saveButtonFinder, expected: false);
+
+          await tester.tap(findSuggestion(code: '1f6e0', text: 'Busy'));
+          await tester.pump();
+
+          checkButtonEnabled(tester, clearButtonFinder, expected: true);
+          checkButtonEnabled(tester, saveButtonFinder, expected: true);
+        });
+
+        testWidgets('emoji & text are added, then removed -> buttons are enabled, then disabled', (tester) async {
```
gnprice: There's a test just below for the buttons getting disabled on hitting clear; and other tests above in this group for them getting enabled in the first place on adding emoji or text. So it seems like everything this test is covering is already covered by more-specific tests, and it can be cut.

(Which is especially nice since it's the longest one in the group, at just over 30 lines.)

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,552 @@
...
+                emojiCode: '1f3e0', reactionType: ReactionType.unicodeEmoji))));
+
+          checkButtonEnabled(tester, clearButtonFinder, expected: true);
+          checkButtonEnabled(tester, saveButtonFinder, expected: false);
+
+          await tester.tap(findSuggestion(code: '1f6e0', text: 'Busy'));
+          await tester.pump();
+
+          checkButtonEnabled(tester, clearButtonFinder, expected: true);
+          checkButtonEnabled(tester, saveButtonFinder, expected: true);
```
gnprice: These two helpers almost always get called in pairs, for the clear and save buttons. How about combining those into one helper?

Then those call sites can generally all be joined to the stanzas above them, following the usual pattern I've mentioned before where a round of "set up, then check" is one stanza. (In fact that'd be a good change even with the two separate helper calls, but especially so with just one.) So:
```suggestion
          await tester.pump();
          checkButtonsEnabled(tester, expectClear: true, expectSave: true);
```

gnprice (commented):
Thanks! Comments below.

lib/model/emoji.dart
sm-sayedi: That "not supposed to happen" phrase was actually for the third case you mentioned. The first revision used `List.first`, but in our last week's check-in call, Chris and I decided to handle the bad situation from the server by using `List.firstOrNull`. When putting that comment, I should have specified the specific case we meant, but now as you mentioned we shouldn't worry about it, I think we can remove that last part of the comment.

test/widgets/set_status_test.dart
sm-sayedi: Ok, so it is not a `MaterialAccountPageRoute` becuase the page is shown by TestZulipApp! Now it makes sense!

sm-sayedi:
Thanks @gnprice for the previous review. New revision pushed.

lib/api/route/realm.dart
```diff
@@ -137,6 +138,16 @@ class ServerEmojiData {
...
   ServerEmojiData({required this.codeToNames});
 
+  /// Returns this emoji data merged with [other].
+  ///
+  /// Returns a new `ServerEmojiData` object combining [codeToNames] of [this]
+  /// and [other]. If there is any duplicate code key, its resulting names value
+  /// will be from [other.codeToNames].
+  @visibleForTesting
+  ServerEmojiData operator +(ServerEmojiData other) {
+    return ServerEmojiData(codeToNames: {...codeToNames, ...other.codeToNames});
```
gnprice: This operator which right now is only intended for test usage, makes it
    easy to combine two emoji data objects with the control of which one
    overrides which.

Since this logic is only needed in tests, best to keep it in test code.

It's actually not easy for me to quickly check just where this is used — "find references" in the IDE doesn't work, because it's an operator override instead of a normal method. That's one of the reasons that operator overrides are good to avoid except where they're really indispensable.

As I said at https://github.com/zulip/zulip-flutter/pull/1701#discussion_r2238064377, best to have the callers just do this directly. That's not much code, and it makes it transparent what's happening.

lib/widgets/set_status.dart
```diff
@@ -0,0 +1,352 @@
...
+              overflow: TextOverflow.ellipsis)),
+          ])),
+    );
+  }
+}
+
+// @visibleForTesting
+// Map<String, String> statusCodesToText(ZulipLocalizations localizations) {
+//   return {
+//     '1f6e0': localizations.userStatusBusy,
```
gnprice: :slightly_smiling_face:

test/test_navigation.dart
```diff
@@ -16,36 +17,43 @@ class TestNavigatorObserver extends NavigatorObserver {
...
 
   @override
   void didRemove(Route<dynamic> route, Route<dynamic>? previousRoute) {
+    super.didRemove(route, previousRoute);
     onRemoved?.call(route, previousRoute);
   }
 
   @override
   void didReplace({ Route<dynamic>? oldRoute, Route<dynamic>? newRoute }) {
+    super.didReplace();
```
gnprice: This super call is missing arguments.

That's a recipe for a confused debugging session if we ever exercise this method — e.g. if we use a method like Navigator.replace and then try to call `pumpPastTransition`.

test/widgets/set_status_test.dart
```diff
@@ -0,0 +1,458 @@
...
+
+    Route<dynamic>? currentRoute;
+    final testNavObserver = TestNavigatorObserver()
+      ..onPushed = (route, _) => currentRoute = route;
+
+    await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+    store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+    await store.addUser(eg.selfUser);
+    if (change.text != OptionNone<String?>() || change.emoji != OptionNone<StatusEmoji?>()) {
+      await store.changeUserStatus(eg.selfUser.userId, change);
```
gnprice: nit: can make this simpler:
```suggestion
    if (change != null) {
      await store.changeUserStatus(eg.selfUser.userId, change!);
```

and let `change` default to null.

gnprice (commented):
Thanks for the revision! Generally this looks good; a couple of important comments below, and some nits.

lib/api/route/realm.dart
sm-sayedi: Dropped this commit in the new revision and used it directly. 🙂

lib/widgets/set_status.dart
sm-sayedi: 🤦🏻‍♂️

sm-sayedi:
Thanks for the review. New changes pushed.

gnprice:
Thanks! All looks good now; merging.

One commit-message nit I noticed and fixed:
```
      ## Commit message ##
    -    api: Add updateStatus
    +    api: Add route updateStatus
```

That makes it clearer what sort of thing it is that's being added.

sm-sayedi:
@laurynmm I think this is now ready to be documented in the help center.

laurynmm:
Thanks for the ping! We got that documentation update done early; see [#35515](https://github.com/zulip/zulip/pull/35515). It's not yet up on Zulip Cloud, but you can see it in the CZO help center docs: https://chat.zulip.org/help/status-and-availability.

sm-sayedi:
Ahh, I see. Thanks for mentioning.

Issue #197, opened 2023-06-16
Track user status


gnprice:
This Zulip feature: https://zulip.com/help/status-and-availability
but only for tracking the data; setting one's own status is a separate issue #198.

In the API this is:
https://zulip.com/api/get-events#user_status
and search for "user_status" on https://zulip.com/api/register-queue .

This information should then appear in the UI at:
* #119
* #195
* and possibly other places.

mofirojean:
@gnprice , I went through the codebase and discovered this was a todo, Can you assign me to it?

gnprice:
@mofirojean Take a look at the steps described in our guide to finding an issue to work on:
https://zulip.readthedocs.io/en/latest/contributing/contributing.html#picking-an-issue-to-work-on

So before claiming an issue, you should have already gotten started working on it, finding the relevant piece of code and getting an idea of how you'll approach it.

In the case of this issue, it's not one that I think would be suitable for a contributor new to Zulip.

mofirojean:
Thank you @gnprice , I will start with something suitable and then come to this when i get better.

mofirojean:
@gnprice , I have been taking a look at the various issues from the project, and i read the documentation you sent to me concerning picking an issue to work on; i haven't seen an issue tagged with `help wanted.` . I wanted to ask if issues could have more clarity so the other participants can understand what to do and also a way to indicate that a particular issue could be attempted by someone new to the project. The past few days I have been going through one issue at a time following the documents you sent above, but I later realized they involve many parts of the code-based being updated, and it has been confusing.

gnprice:
From [our guide](https://zulip.readthedocs.io/en/latest/contributing/contributing.html#where-to-look-for-an-issue):
> Mobile apps: no “help wanted” label, but see the [project board](https://github.com/orgs/zulip/projects/5/views/4) for the upcoming Flutter-based app. Look for issues up through the “Launch” milestone, and that aren’t already assigned.

Does that answer the question?

If not, the best venue for follow-up questions would be in #mobile-dev-help or another public channel on chat.zulip.org. A discussion on an unrelated GitHub issue thread is less visible — which means fewer people are able to help you out — and it also makes the issue thread harder to read when someone eventually picks up this issue in order to implement it.

chrisbobbe:
[...]

PR #1702, opened 2025-07-10
Show user status in UI

sm-sayedi:
The remaining half of #1629. For related images, please see #1629 description.

Fixes: #197

### DIFF
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -223,7 +223,7 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
     final designVariables = DesignVariables.of(context);
 
     final child = switch (option) {
-      MentionAutocompleteResult() => _MentionAutocompleteItem(
+      MentionAutocompleteResult() => MentionAutocompleteItem(
         option: option, narrow: narrow),
       EmojiAutocompleteResult() => _EmojiAutocompleteItem(option: option),
     };
@@ -238,8 +238,13 @@ class ComposeAutocomplete extends AutocompleteField<ComposeAutocompleteQuery, Co
   }
 }
 
-class _MentionAutocompleteItem extends StatelessWidget {
-  const _MentionAutocompleteItem({required this.option, required this.narrow});
+@visibleForTesting
+class MentionAutocompleteItem extends StatelessWidget {
+  const MentionAutocompleteItem({
+    super.key,
+    required this.option,
+    required this.narrow,
+  });
 
   final MentionAutocompleteResult option;
   final Narrow narrow;
--- a/lib/widgets/autocomplete.dart
+++ b/lib/widgets/autocomplete.dart
@@ -277,29 +277,35 @@ class MentionAutocompleteItem extends StatelessWidget {
 
     Widget avatar;
     String label;
+    Widget? emoji;
     String? sublabel;
     switch (option) {
       case UserMentionAutocompleteResult(:var userId):
         avatar = Avatar(userId: userId, size: 36, borderRadius: 4);
         label = store.userDisplayName(userId);
+        emoji = UserStatusEmoji(userId: userId, size: 18,
+          padding: const EdgeInsetsDirectional.only(start: 5.0));
         sublabel = store.userDisplayEmail(userId);
       case WildcardMentionAutocompleteResult(:var wildcardOption):
         avatar = SizedBox.square(dimension: 36,
           child: const Icon(ZulipIcons.three_person, size: 24));
         label = wildcardOption.canonicalString;
+        emoji = null;
         sublabel = wildcardSublabel(wildcardOption, context: context, store: store);
     }
 
-    final labelWidget = Text(
-      label,
-      style: TextStyle(
-        fontSize: 18,
-        height: 20 / 18,
-        color: designVariables.contextMenuItemLabel,
-      ).merge(weightVariableTextStyle(context,
-          wght: sublabel == null ? 500 : 600)),
-      overflow: TextOverflow.ellipsis,
-      maxLines: 1);
+    final labelWidget = Row(children: [
+      Flexible(child: Text(label,
+        style: TextStyle(
+          fontSize: 18,
+          height: 20 / 18,
+          color: designVariables.contextMenuItemLabel,
+        ).merge(weightVariableTextStyle(context,
+            wght: sublabel == null ? 500 : 600)),
+        overflow: TextOverflow.ellipsis,
+        maxLines: 1)),
+      ?emoji,
+    ]);
 
     final sublabelWidget = sublabel == null ? null : Text(
       sublabel,
@@ -318,10 +324,7 @@ class MentionAutocompleteItem extends StatelessWidget {
         Expanded(child: Column(
           mainAxisSize: MainAxisSize.min,
           crossAxisAlignment: CrossAxisAlignment.start,
-          children: [
-            labelWidget,
-            if (sublabelWidget != null) sublabelWidget,
-          ])),
+          children: [labelWidget, ?sublabelWidget])),
       ]));
   }
 }
--- a/lib/widgets/new_dm_sheet.dart
+++ b/lib/widgets/new_dm_sheet.dart
@@ -317,6 +317,8 @@ class _SelectedUserChip extends StatelessWidget {
                   fontSize: 16,
                   height: 16 / 16,
                   color: designVariables.labelMenuButton)))),
+          UserStatusEmoji(userId: userId, size: 16,
+            padding: EdgeInsetsDirectional.only(end: 4)),
         ])));
   }
 }
@@ -415,7 +417,11 @@ class _NewDmUserListItem extends StatelessWidget {
             Avatar(userId: userId, size: 32, borderRadius: 3),
             SizedBox(width: 8),
             Expanded(
-              child: Text(store.userDisplayName(userId),
+              child: Text.rich(
+                TextSpan(text: store.userDisplayName(userId), children: [
+                  UserStatusEmoji.asWidgetSpan(userId: userId, fontSize: 17,
+                    textScaler: MediaQuery.textScalerOf(context)),
+                ]),
                 style: TextStyle(
                   fontSize: 17,
                   height: 19 / 17,
--- a/lib/widgets/profile.dart
+++ b/lib/widgets/profile.dart
@@ -16,6 +16,7 @@ import 'page.dart';
 import 'remote_settings.dart';
 import 'store.dart';
 import 'text.dart';
+import 'theme.dart';
 
 class _TextStyles {
   static const primaryFieldText = TextStyle(fontSize: 20);
@@ -51,6 +52,7 @@ class ProfilePage extends StatelessWidget {
     final nameStyle = _TextStyles.primaryFieldText
       .merge(weightVariableTextStyle(context, wght: 700));
 
+    final userStatus = store.getUserStatus(userId);
     final displayEmail = store.userDisplayEmail(userId);
     final items = [
       Center(
@@ -73,9 +75,20 @@ class ProfilePage extends StatelessWidget {
           ),
           // TODO write a test where the user is muted; check this and avatar
           TextSpan(text: store.userDisplayName(userId, replaceIfMuted: false)),
+          UserStatusEmoji.asWidgetSpan(
+            userId: userId,
+            fontSize: nameStyle.fontSize!,
+            textScaler: MediaQuery.textScalerOf(context),
+            neverAnimate: false,
+          ),
         ]),
         textAlign: TextAlign.center,
         style: nameStyle),
+      if (userStatus.text != null)
+        Text(userStatus.text!,
+          textAlign: TextAlign.center,
+          style: TextStyle(fontSize: 18, height: 22 / 18,
+            color: DesignVariables.of(context).userStatusText)),
       if (displayEmail != null)
         Text(displayEmail,
           textAlign: TextAlign.center,
@@ -83,7 +96,6 @@ class ProfilePage extends StatelessWidget {
       Text(roleToLabel(user.role, zulipLocalizations),
         textAlign: TextAlign.center,
         style: _TextStyles.primaryFieldText),
-      // TODO(#197) render user status
       // TODO(#196) render active status
       // TODO(#292) render user local time
 
--- a/lib/widgets/recent_dm_conversations.dart
+++ b/lib/widgets/recent_dm_conversations.dart
@@ -104,23 +104,29 @@ class RecentDmConversationsItem extends StatelessWidget {
     final store = PerAccountStoreWidget.of(context);
     final designVariables = DesignVariables.of(context);
 
-    final String title;
+    final InlineSpan title;
     final Widget avatar;
     int? userIdForPresence;
     switch (narrow.otherRecipientIds) { // TODO dedupe with DM items in [InboxPage]
       case []:
-        title = store.selfUser.fullName;
+        title = TextSpan(text: store.selfUser.fullName, children: [
+          UserStatusEmoji.asWidgetSpan(userId: store.selfUserId,
+            fontSize: 17, textScaler: MediaQuery.textScalerOf(context)),
+        ]);
         avatar = AvatarImage(userId: store.selfUserId, size: _avatarSize);
       case [var otherUserId]:
-        title = store.userDisplayName(otherUserId);
+        title = TextSpan(text: store.userDisplayName(otherUserId), children: [
+          UserStatusEmoji.asWidgetSpan(userId: otherUserId,
+            fontSize: 17, textScaler: MediaQuery.textScalerOf(context)),
+        ]);
         avatar = AvatarImage(userId: otherUserId, size: _avatarSize);
         userIdForPresence = otherUserId;
       default:
-        // TODO(i18n): List formatting, like you can do in JavaScript:
-        //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya'])
-        //   // 'Chris、Greg、Alya'
-        title = narrow.otherRecipientIds.map(store.userDisplayName)
-          .join(', ');
+        title = TextSpan(
+          // TODO(i18n): List formatting, like you can do in JavaScript:
+          //   new Intl.ListFormat('ja').format(['Chris', 'Greg', 'Alya'])
+          //   // 'Chris、Greg、Alya'
+          text: narrow.otherRecipientIds.map(store.userDisplayName).join(', '));
         avatar = ColoredBox(color: designVariables.avatarPlaceholderBg,
           child: Center(
             child: Icon(color: designVariables.avatarPlaceholderIcon,
@@ -148,7 +154,7 @@ class RecentDmConversationsItem extends StatelessWidget {
             const SizedBox(width: 8),
             Expanded(child: Padding(
               padding: const EdgeInsets.symmetric(vertical: 4),
-              child: Text(
+              child: Text.rich(
                 style: TextStyle(
                   fontSize: 17,
                   height: (20 / 17),
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -14,6 +14,7 @@ import 'package:zulip/api/model/model.dart';
 import 'package:zulip/api/model/narrow.dart';
 import 'package:zulip/api/route/channels.dart';
 import 'package:zulip/api/route/messages.dart';
+import 'package:zulip/basic.dart';
 import 'package:zulip/model/actions.dart';
 import 'package:zulip/model/localizations.dart';
 import 'package:zulip/model/message.dart';
@@ -1770,6 +1771,74 @@ void main() {
       checkUser(users[2], isBot: false);
 
       debugNetworkImageHttpClientProvider = null;
+    });
+
+    group('User status', () {
+      void checkFindsStatusEmoji(WidgetTester tester, Finder emojiFinder) {
+        final statusEmojiFinder = find.ancestor(of: emojiFinder,
+          matching: find.byType(UserStatusEmoji));
+        check(statusEmojiFinder).findsOne();
+        check(tester.widget<UserStatusEmoji>(statusEmojiFinder)
+          .neverAnimate).isTrue();
+        check(find.ancestor(of: statusEmojiFinder,
+          matching: find.byType(SenderRow))).findsOne();
+      }
+
+      testWidgets('emoji (unicode) & text are set -> emoji is displayed, text is not', (tester) async {
+        final user = eg.user();
+        await setupMessageListPage(tester,
+          users: [user], messages: [eg.streamMessage(sender: user)]);
+        await store.changeUserStatus(user.userId, UserStatusChange(
+          text: OptionSome('Busy'),
+          emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+            emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji))));
+        await tester.pump();
+
+        checkFindsStatusEmoji(tester, find.text('\u{1f6e0}'));
+        check(find.textContaining('Busy')).findsNothing();
+      });
+
+      testWidgets('emoji (image) & text are set -> emoji is displayed, text is not', (tester) async {
+        prepareBoringImageHttpClient();
+
+        final user = eg.user();
+        await setupMessageListPage(tester,
+          users: [user], messages: [eg.streamMessage(sender: user)]);
+        await store.changeUserStatus(user.userId, UserStatusChange(
+          text: OptionSome('Coding'),
+          emoji: OptionSome(StatusEmoji(emojiName: 'zulip',
+            emojiCode: 'zulip', reactionType: ReactionType.zulipExtraEmoji))));
+        await tester.pump();
+
+        checkFindsStatusEmoji(tester, find.byType(Image));
+        check(find.textContaining('Coding')).findsNothing();
+
+        debugNetworkImageHttpClientProvider = null;
+      });
+
+      testWidgets('longer user name -> emoji stays visible', (tester) async {
+        final user = eg.user(fullName: 'User with a very very very long name to check if emoji is still visible');
+        await setupMessageListPage(tester,
+          users: [user], messages: [eg.streamMessage(sender: user)]);
+        await store.changeUserStatus(user.userId, UserStatusChange(
+          text: OptionNone(),
+          emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+            emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji))));
+        await tester.pump();
+
+        checkFindsStatusEmoji(tester, find.text('\u{1f6e0}'));
+      });
+
+      testWidgets('emoji is not set, text is set -> text is not displayed', (tester) async {
+        final user = eg.user();
+        await setupMessageListPage(tester,
+          users: [user], messages: [eg.streamMessage(sender: user)]);
+        await store.changeUserStatus(user.userId, UserStatusChange(
+          text: OptionSome('Busy'), emoji: OptionNone()));
+        await tester.pump();
+
+        check(find.textContaining('Busy')).findsNothing();
+      });
     });
 
     group('Muted sender', () {
--- a/test/widgets/new_dm_sheet_test.dart
+++ b/test/widgets/new_dm_sheet_test.dart
@@ -3,6 +3,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_checks/flutter_checks.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:zulip/api/model/model.dart';
+import 'package:zulip/basic.dart';
+import 'package:zulip/model/store.dart';
 import 'package:zulip/widgets/app_bar.dart';
 import 'package:zulip/widgets/compose_box.dart';
 import 'package:zulip/widgets/content.dart';
@@ -17,7 +19,10 @@ import '../flutter_checks.dart';
 import '../model/binding.dart';
 import '../model/test_store.dart';
 import '../test_navigation.dart';
+import 'finders.dart';
 import 'test_app.dart';
+
+late PerAccountStore store;
 
 Future<void> setupSheet(WidgetTester tester, {
   required List<User> users,
@@ -30,7 +35,7 @@ Future<void> setupSheet(WidgetTester tester, {
     ..onPushed = (route, _) => lastPushedRoute = route;
 
   await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
-  final store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
+  store = await testBinding.globalStore.perAccount(eg.selfAccount.id);
   await store.addUsers(users);
   if (mutedUserIds != null) {
     await store.setMutedUsers(mutedUserIds);
@@ -65,7 +70,8 @@ void main() {
   }
 
   Finder findUserTile(User user) =>
-    find.widgetWithText(InkWell, user.fullName).first;
+    find.ancestor(of: findText(user.fullName, includePlaceholders: false),
+      matching: find.byType(InkWell)).first;
 
   Finder findUserChip(User user) {
     final findAvatar = find.byWidgetPredicate((widget) =>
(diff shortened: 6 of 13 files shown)


gnprice:
Copying the review status from #1629 (per https://github.com/zulip/zulip-flutter/pull/1629#pullrequestreview-299892783): I've read the first commit except for its tests:
463b938a7 msglist: Show user status emoji

and the comments I had there have been resolved. Remaining to review are those tests, and the other 4 commits:
14093b28c recent-dms: Show user status emoji in recent DMs page
646f9d962 new-dm: Show user status emoji
846261a6c autocomplete: Show user status emoji in user-mention autocomplete
e9f682eaf profile: Show user status

test/widgets/message_list_test.dart
```diff
@@ -50,6 +52,22 @@ import 'message_list_checks.dart';
...
 
+/// Finder for [UserStatusEmoji] widget.
+///
+/// Use [type] to specify the exact emoji child widget. It can be either
+/// [UnicodeEmojiWidget] or [ImageEmojiWidget].
+Finder findStatusEmoji(Type type) {
+  assert(type == UnicodeEmojiWidget || type == ImageEmojiWidget);
+  return find.ancestor(
+    of: find.byType(type),
+    matching: find.byType(UserStatusEmoji));
```
gnprice: This seems like it's clearer if just inlined at its call sites.

The implementation (minus the assert, which becomes unnecessary) isn't much longer than the call; and it's more transparent about what it's doing. I also don't feel like this is particularly encapsulating any knowledge of the details of how the UserStatusEmoji widget works.

gnprice: Also, after doing that: how about just `find.byType(UserStatusEmoji)`? It's not clear to me what's gained by adding the condition that the widget have a UnicodeEmojiWidget descendant, or an ImageEmojiWidget descendant.

test/widgets/message_list_test.dart
```diff
@@ -50,6 +52,22 @@ import 'message_list_checks.dart';
...
+}
+
+void checkUserStatusEmoji(Finder emojiFinder, {required bool isAnimated}) {
+  check((emojiFinder
+    .evaluate().first.widget as UserStatusEmoji).neverAnimate).equals(!isAnimated);
```
gnprice: This also feels like it'd be clearer inlined. The actual check is all about `neverAnimate`, but the name doesn't sound like that — it sounds like it's going to be doing some other, more general, check.

gnprice: Instead of `finder.evaluate()`, use `tester` explicitly:
```suggestion
  check(tester.firstWidget<UserStatusEmoji>(emojiFinder).neverAnimate)
    .equals(!isAnimated);
```

That way it's conceptually clearer what's going on. (The implementation of `finder.evaluate` will end up using the same WidgetTester instance anyway, getting hold of it behind the scenes via a singleton.)

test/widgets/message_list_test.dart
```diff
@@ -91,6 +110,7 @@ void main() {
     if (mutedUserIds != null) {
       await store.setMutedUsers(mutedUserIds);
     }
+    await store.changeUserStatuses(userStatuses ?? []);
```
gnprice: How about having the individual test cases do this? That way we avoid adding yet another feature to this shared helper. (It probably should have fewer features already ­— that'd help make fewer things for the reader of each test case to potentially have to think about.)

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+
+      testWidgets('status emoji is set -> emoji is displayed', (tester) async {
+        await setupMessageListPage(tester,
+          users: [user1, user2],
+          messages: [eg.streamMessage(sender: user1), eg.streamMessage(sender: user2)],
+          userStatuses: [
+            (
+              user1.userId,
+              UserStatusChange(
+                text: OptionSome('Busy'),
```
gnprice: Hmm, these tuples are a bit of a pain to read.

How about making `changeUserStatuses` instead take a Map? That seems semantically appropriate, as there's no reason to have the same user ID more than once in the same call. Then:
```suggestion
          userStatuses: {
            user1.userId: UserStatusChange(
              text: OptionSome('Busy'),
```

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+              user2.userId,
+              UserStatusChange(
+                text: OptionSome('Coding'),
+                emoji: OptionSome(StatusEmoji(emojiName: 'zulip',
+                  emojiCode: 'zulip', reactionType: ReactionType.zulipExtraEmoji)))
+            ),
+          ],
+        );
+        checkStatusEmoji(user1, UnicodeEmojiWidget, isPresent: true);
+        checkStatusEmoji(user2, ImageEmojiWidget, isPresent: true);
```
gnprice: I see, so this pair of call sites is using the emoji type to select which UserStatusEmoji widget to look at.

I think the simplest fix for that is to make these two separate test cases. The test doesn't seem to be really gaining anything from having both at once.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+              user1.userId,
+              UserStatusChange(
+                text: OptionSome('Busy'),
+                emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+                  emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji)))
+            ),
+            (
+              user2.userId,
+              UserStatusChange(
+                text: OptionSome('Coding'),
```
gnprice: It'd also be good to check that the status text does not appear.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+        check(statusEmojiFinder).findsExactly(isPresent ? 1 : 0);
+        if (isPresent) {
+          checkUserStatusEmoji(statusEmojiFinder, isAnimated: false);
+        }
+
+        final senderRowFinder = find.ancestor(of: nameFinder,
+          matching: find.ancestor(of: statusEmojiFinder,
+            matching: find.byType(Row)));
+        isPresent
+          ? check(senderRowFinder).findsAny()
```
gnprice: This doesn't really say very much — I think it'd be better to leave it out.

It looks like it's intended to check that the sender's status emoji appears in the right row with the sender's name. But it doesn't really have that effect, because `Row` is a quite generic layout widget which could be used in a lot of contexts: we could easily end up having a `Row` near the top of the widget tree, for whatever reason, and then that would cause this check to always pass (whenever we get as far as this line in the first place).

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+          : check(senderRowFinder).findsNothing();
```
gnprice: Instead of these three separate if-isPresent conditions, I think it'd be clearer to have one if/else on isPresent.

In particular, when isPresent is false, only the first check does anything. So with an if/else, the rest can just be left out in that branch.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+      }
+
+      final user1 = eg.user(userId: 1,
+        fullName: 'User with a very very very long name to check if emoji is still visible');
```
gnprice: This point deserves its own test case with its own name. That'd help make the point of what that test case is doing.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1775,64 @@ void main() {
...
+      final user2 = eg.user(userId: 2, fullName: 'User 2');
```
gnprice: The specific user IDs don't matter, do they? I don't think they do. So they can be left out, to make less for the reader to have to read and think about.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -148,16 +148,31 @@ class RecentDmConversationsItem extends StatelessWidget {
...
+              child: Text.rich(
+                TextSpan(
+                  children: [
+                    TextSpan(text: title),
+                    ...(switch (narrow.otherRecipientIds) {
+                      // self-DM
+                      [] => [UserStatusEmoji.asWidgetSpan(userId: store.selfUserId,
+                              fontSize: 17, textScaler: MediaQuery.textScalerOf(context))],
+                      // 1:1-DM
+                      [final otherUserId] =>
```
gnprice: Let's pull all this out of the return-expression and into a local variable, to reduce the amount of logic that's in the one expression.

In fact, looking at where `title` comes from: it's a switch on exactly the same three cases as this. So how about making `title` an InlineSpan that includes this?

lib/widgets/new_dm_sheet.dart
```diff
@@ -415,7 +417,12 @@ class _NewDmUserListItem extends StatelessWidget {
             Avatar(userId: userId, size: 32, borderRadius: 3),
             SizedBox(width: 8),
             Expanded(
-              child: Text(store.userDisplayName(userId),
+              child: Text.rich(
+                TextSpan(
+                  children: [
+                    TextSpan(text: store.userDisplayName(userId)),
+                    UserStatusEmoji.asWidgetSpan(userId: userId, fontSize: 17,
+                      textScaler: MediaQuery.textScalerOf(context))]),
```
gnprice: nit: for list elements, give the last one a trailing comma and newline like the others, because it's not a distinguished "child" or "payload" argument; OTOH can compact this by putting the "children" argument on the first line:
```suggestion
                TextSpan(children: [
                  TextSpan(text: store.userDisplayName(userId)),
                  UserStatusEmoji.asWidgetSpan(userId: userId, fontSize: 17,
                    textScaler: MediaQuery.textScalerOf(context)),
                ]),
```

lib/widgets/autocomplete.dart
```diff
@@ -314,7 +314,11 @@ class _MentionAutocompleteItem extends StatelessWidget {
           mainAxisSize: MainAxisSize.min,
           crossAxisAlignment: CrossAxisAlignment.start,
           children: [
-            labelWidget,
+            Row(children: [
+              Flexible(child: labelWidget),
+              if (option case UserMentionAutocompleteResult(:var userId))
+                UserStatusEmoji(userId: userId, size: 18,
+                  padding: const EdgeInsetsDirectional.only(start: 5.0))]),
```
gnprice: nit: similarly:
```suggestion
                  padding: const EdgeInsetsDirectional.only(start: 5.0)),
            ]),
```

lib/widgets/autocomplete.dart
```diff
@@ -314,7 +314,11 @@ class _MentionAutocompleteItem extends StatelessWidget {
...
             if (sublabelWidget != null) sublabelWidget,
           ])),
```
gnprice: I think this new logic is best included as part of `labelWidget`. Conceptually it's closely tied to the user's name, which is part of the label.

Probably in fact the cleanest home for this is in the `switch (option)` above, which is already the place where we inspect the details of `option`.

lib/widgets/profile.dart
```diff
@@ -73,17 +75,28 @@ class ProfilePage extends StatelessWidget {
           ),
           // TODO write a test where the user is muted; check this and avatar
           TextSpan(text: store.userDisplayName(userId, replaceIfMuted: false)),
+          UserStatusEmoji.asWidgetSpan(
+            userId: userId,
+            fontSize: 20,
```
gnprice: nit: this says `fontSize: 20` but the presence circle above says `fontSize: nameStyle.fontSize!`; should say the same thing both places, since the intent is for them to be the same

lib/widgets/profile.dart
```diff
@@ -47,6 +48,7 @@ class ProfilePage extends StatelessWidget {
     if (user == null) {
       return const _ProfileErrorPage();
     }
+    final userStatus = store.getUserStatus(userId);
```
gnprice: nit: put this next to `displayEmail`; conceptually they're here for the very same reason

lib/widgets/profile.dart
```diff
@@ -73,17 +75,28 @@ class ProfilePage extends StatelessWidget {
...
         ]),
         textAlign: TextAlign.center,
         style: nameStyle),
+      if (userStatus.text != null)
+        Text(userStatus.text!,
+          textAlign: TextAlign.center,
+          style: TextStyle(fontSize: 18, height: 22 / 18,
+            color: DesignVariables.of(context).userStatusText)),
+
       if (displayEmail != null)
```
gnprice: Is this blank line intended?

The items just above here feel to me closely related to the items just below, and in particular just as closely related as other items that are grouped next to each other.

test/widgets/message_list_test.dart
sm-sayedi: > Also, after doing that: how about just `find.byType(UserStatusEmoji)`?

Copying https://github.com/zulip/zulip-flutter/pull/1629#discussion_r2181040133 here: 🙂
> As we discussed in this week’s check-in call, `UserStatusEmoji` for many cases can contain only `SizedBox.shrink()`, including when there’s no status set for a user. So only using `find.byType(UserStatusEmoji)` will pass the tests even when we don’t pass status for a user.

test/widgets/message_list_test.dart
sm-sayedi: What about the following?
```dart
final senderRowFinder = find.ancestor(of: nameFinder,
  matching: find.ancestor(of: statusEmojiFinder,
    matching: find.byType(SenderRow)));
check(senderRowFinder).findsExactly(isPresent ? 1 : 0);
```
I think this is more accurate and helps make sure the emoji is displayed in the correct place.

test/widgets/message_list_test.dart
sm-sayedi: The main motivation for not using `tester` was to minimize the number of params passed to the helper methods (`checkStatusEmoji` and others in the next commits). Now that it helps in making the tests clearer, so going to use that in the new revision.🙂

sm-sayedi:
Thanks @gnprice for the detailed review. Revision pushed, PTAL.

Note: CI is failing with the following issue:
```
Run flutter pub get
Resolving dependencies...
The current Flutter SDK version is 0.0.0-unknown.

Because zulip requires Flutter SDK version >=3.33.0-1.0.pre.832, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Try using the Flutter SDK version: 3.35.0-0.0.pre. 
Failed to update packages.
```

I tried pushing several times, but it didn't solve the problem.

test/widgets/message_list_test.dart
sm-sayedi: I think when using `SenderRow`, we can ignore `nameFinder`:
```dart
final senderRowFinder = find.ancestor(of: statusEmojiFinder,
    matching: find.byType(SenderRow));
check(senderRowFinder).findsExactly(isPresent ? 1 : 0);
```

gnprice:
Yeah, that CI failure is unrelated. Filed as #1710, and I'll fix it.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1757,104 @@ void main() {
...
+
+      testWidgets('status emoji (unicode) & text are set -> emoji is displayed, text is not', (tester) async {
+        final user = eg.user();
+        await setupMessageListPage(tester,
+          users: [user],
+          messages: [eg.streamMessage(sender: user)],
+        );
+        await store.changeUserStatuses({
+          user.userId: UserStatusChange(
+            text: OptionSome('Busy'),
```
gnprice: nit: can be a bit simpler and more compact by calling `changeUserStatus` directly:
```suggestion
        await store.changeUserStatus(user.userId, UserStatusChange(
          text: OptionSome('Busy'),
```

Also nice is that that corresponds more directly to what happens in the real server API: each user's status comes in a separate UserStatusEvent.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1757,104 @@ void main() {
...
+          check(statusEmojiFinder).findsNothing();
+        }
+      }
+
+      testWidgets('status emoji (unicode) & text are set -> emoji is displayed, text is not', (tester) async {
+        final user = eg.user();
+        await setupMessageListPage(tester,
+          users: [user],
+          messages: [eg.streamMessage(sender: user)],
+        );
```
gnprice: nit: can compact a bit vertically:
```suggestion
        await setupMessageListPage(tester,
          users: [user], messages: [eg.streamMessage(sender: user)]);
```

Making each test case more compact is helpful for the reader scanning through a given test case to fully understand what it's doing, and especially for scanning through several neighboring test cases to compare them and to think about what situations are covered and what situations there might be that aren't yet.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1757,104 @@ void main() {
...
+        await store.changeUserStatuses({
+          user.userId: UserStatusChange(
+            text: OptionSome('Coding'),
+            emoji: OptionSome(StatusEmoji(emojiName: 'zulip',
+              emojiCode: 'zulip', reactionType: ReactionType.zulipExtraEmoji))),
+        });
+        await tester.pump();
+
+        checkStatusEmoji(tester, type: ImageEmojiWidget, isPresent: true);
```
gnprice: It's kind of puzzling to me that this isn't giving an error, actually: if there's an image emoji here, that should be a RealmContentNetworkImage, which should mean the widget tries to load an image from the network. And I'd expect that to throw an error, given that this test case hasn't set debugNetworkImageHttpClientProvider (e.g. by calling prepareBoringImageHttpClient).

Would you try to get to the bottom of that? I wonder if perhaps this widget tree isn't getting fully built in one frame, and another `tester.pump` is needed in order to fully exercise the code.

test/widgets/message_list_test.dart
```diff
@@ -1755,6 +1757,104 @@ void main() {
...
+        );
+        await store.changeUserStatuses({
+          user.userId: UserStatusChange(
+            text: OptionSome('Busy'),
+            emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+              emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji))),
+        });
+        await tester.pump();
+
+        checkStatusEmoji(tester, type: UnicodeEmojiWidget, isPresent: true);
```
gnprice: Instead of `find.byType(UnicodeEmojiWidget)`, how about something like `find.text("\u{1f6e0}")`? (Continuing from https://github.com/zulip/zulip-flutter/pull/1702#discussion_r2212256700 .)

That's more nicely end-to-end — it expresses what we're looking for in terms of what the user sees, rather than the way we've organized our code. In this case it makes it a bit more specific; it also means that if we refactor how those widgets work, these tests wouldn't need to change. (Except insofar as we changed how the UI actually *looks* to the user in ways this is checking for, which is a good reason to need to change it.)

Then I think that opens up a further refactor that would make these a bit more transparent:
```suggestion
        checkFindsStatusEmoji(tester, find.text('\u{1f6e0}'));
```

The `isPresent: false` call sites could become like:
```dart
        check(find.text('\u{1f6e0}')).findsNothing();
```

…. But actually, looking at them, the test data in those test cases doesn't supply any emoji in the first place. I think it's fine to not have any such check in that case, then — instead we can rely on an assumption that our code isn't going to invent out of whole cloth an emoji to display as the user's status emoji.

test/widgets/message_list_test.dart
gnprice: Sure, that works.

It does introduce some degree of dependence on the internals of how we've organized our widgets (the existence of a SenderRow widget class, and the fact that it should contain the status emoji). But I think the concept of a "sender row" is a pretty stable aspect of our UI, so that the only way that assumption is likely to break is if we're changing the real user-visible UI in a way that alters the user-facing fact that the sender's status emoji belongs on a row together with their name etc.

lib/widgets/recent_dm_conversations.dart
```diff
@@ -104,23 +104,31 @@ class RecentDmConversationsItem extends StatelessWidget {
...
 
-    final String title;
+    final InlineSpan title;
     final Widget avatar;
     int? userIdForPresence;
     switch (narrow.otherRecipientIds) { // TODO dedupe with DM items in [InboxPage]
       case []:
-        title = store.selfUser.fullName;
+        title = TextSpan(children: [
+          TextSpan(text: store.selfUser.fullName),
```
gnprice: nit: can simplify these a bit by using both `text` and `children`:
```suggestion
        title = TextSpan(text: store.selfUser.fullName, children: [
```

(See the docs for those two TextSpan fields.)

Looking at the TextSpan implementation, I think that saves a small amount of work by eliminating that indirection. There's potentially a lot of these on the screen, so that savings is potentially useful too.

lib/widgets/autocomplete.dart
```diff
@@ -272,29 +277,35 @@ class _MentionAutocompleteItem extends StatelessWidget {
...
         avatar = Avatar(userId: userId, size: 36, borderRadius: 4);
         label = store.userDisplayName(userId);
+        emoji = UserStatusEmoji(userId: userId, size: 18,
+          padding: const EdgeInsetsDirectional.only(start: 5.0));
         sublabel = store.userDisplayEmail(userId);
       case WildcardMentionAutocompleteResult(:var wildcardOption):
         avatar = SizedBox.square(dimension: 36,
           child: const Icon(ZulipIcons.three_person, size: 24));
+        emoji = null;
         label = wildcardOption.canonicalString;
```
gnprice: nit: order label vs. emoji consistently

(probably label then emoji, to correspond to how they're displayed)

test/widgets/message_list_test.dart
gnprice: (continued in https://github.com/zulip/zulip-flutter/pull/1702#discussion_r2214862085 below)

test/widgets/message_list_test.dart
sm-sayedi: The first problem is that it needs another `tester.pump`, but even with that, it will not throw an error because we have `ImageEmojiWidget.errorBuilder` set. Nevertheless, added `prepareBoringImageHttpClient`.

sm-sayedi:
Thanks for the review. New revision pushed.

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -231,6 +247,33 @@ void main() {
...
+          testWidgets('status emoji & text are set -> emoji is displayed, text is not', (tester) async {
+            final message = eg.dmMessage(from: eg.selfUser, to: []);
+            await setupPage(tester, dmMessages: [message], users: []);
+            await store.changeUserStatus(eg.selfUser.userId, UserStatusChange(
+              text: OptionSome('Busy'),
+              emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+                emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji))));
+            await tester.pump();
+
+            checkFindsStatusEmoji(tester, find.text('\u{1f6e0}'));
```
sm-sayedi, line 258: For this commit and the following ones, should we make `find.text('\u{1f6e0}')` an implementation detail of `checkFindsStatusEmoji` as it stays the same?

lib/widgets/new_dm_sheet.dart
```diff
@@ -415,7 +417,12 @@ class _NewDmUserListItem extends StatelessWidget {
             Avatar(userId: userId, size: 32, borderRadius: 3),
             SizedBox(width: 8),
             Expanded(
-              child: Text(store.userDisplayName(userId),
+              child: Text.rich(
+                TextSpan(children: [
+                  TextSpan(text: store.userDisplayName(userId)),
```
gnprice: same as https://github.com/zulip/zulip-flutter/pull/1702#discussion_r2214875090

test/widgets/message_list_test.dart
```diff
@@ -1758,6 +1759,79 @@ void main() {
...
+          users: [user], messages: [eg.streamMessage(sender: user)]);
+        await store.changeUserStatus(user.userId, UserStatusChange(
+          text: OptionSome('Busy'),
+          emoji: OptionSome(StatusEmoji(emojiName: 'working_on_it',
+            emojiCode: '1f6e0', reactionType: ReactionType.unicodeEmoji))));
+        await tester.pump();
+
+        checkFindsStatusEmoji(tester, find.text('\u{1f6e0}'));
+        check(find.descendant(of: find.byType(SenderRow),
+          matching: find.text('Busy'))).findsNothing();
```
gnprice: This test is a bit stronger if the "finds nothing" check isn't limited by SenderRow:
```suggestion
        checkFindsStatusEmoji(tester, find.text('\u{1f6e0}'));
        check(find.text('Busy')).findsNothing();
```

With the SenderRow restriction, the reader has to wonder if perhaps the status text is getting shown after all, and just doesn't happen to be within a widget called SenderRow. (Perhaps something changed in the organization of the UI code.)

This is a bit different from the situation at https://github.com/zulip/zulip-flutter/pull/1702#discussion_r2214870406 (where introducing SenderRow was OK), because the check has the opposite polarity. In that test, adding a SenderRow condition makes the check more demanding — so if the code gets reorganized in a way that breaks the test's implicit assumptions about how it's organized, then the test will fail and we'll know we need to update it. But here, adding SenderRow makes it *less* demanding — so in that situation the check wouldn't fail, but instead would become vacuous, less able to fail when it should.

test/widgets/message_list_test.dart
```diff
@@ -1758,6 +1759,79 @@ void main() {
...
+        prepareBoringImageHttpClient();
+
+        final user = eg.user();
+        await setupMessageListPage(tester,
+          users: [user], messages: [eg.streamMessage(sender: user)]);
+        await store.changeUserStatus(user.userId, UserStatusChange(
+          text: OptionSome('Coding'),
+          emoji: OptionSome(StatusEmoji(emojiName: 'zulip',
+            emojiCode: 'zulip', reactionType: ReactionType.zulipExtraEmoji))));
+        await tester.pumpAndSettle();
```
gnprice: Best to avoid pumpAndSettle: https://github.com/flutter/flutter/blob/main/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-using-pumpandsettle

Can this be two pumps? Better to be explicit that way.

test/widgets/message_list_test.dart
gnprice: Hmm, I see.

How did you determine that it needs another `tester.pump` — what's missing after the first pump?

test/widgets/message_list_test.dart
```diff
@@ -1758,6 +1759,79 @@ void main() {
...
+      void checkFindsStatusEmoji(WidgetTester tester, Finder emojiFinder) {
+        final statusEmojiFinder = find.ancestor(of: emojiFinder,
+          matching: find.byType(UserStatusEmoji));
+        check(statusEmojiFinder).findsOne();
+        check(tester.firstWidget<UserStatusEmoji>(statusEmojiFinder)
+          .neverAnimate).isTrue();
+
+        final senderRowFinder = find.ancestor(of: statusEmojiFinder,
+          matching: find.byType(SenderRow));
+        check(senderRowFinder).findsOne();
```
gnprice: nit: I think this would be just as clear if inlined, and a bit shorter. It's saying the status emoji has a SenderRow as an ancestor.

test/widgets/message_list_test.dart
```diff
@@ -1758,6 +1759,79 @@ void main() {
       debugNetworkImageHttpClientProvider = null;
     });
 
+    group('User status', () {
+      void checkFindsStatusEmoji(WidgetTester tester, Finder emojiFinder) {
+        final statusEmojiFinder = find.ancestor(of: emojiFinder,
+          matching: find.byType(UserStatusEmoji));
+        check(statusEmojiFinder).findsOne();
+        check(tester.firstWidget<UserStatusEmoji>(statusEmojiFinder)
```
gnprice: nit: since we're expecting this to find just one widget anyway, can simplify `firstWidget` to `widget`:
```suggestion
        check(statusEmojiFinder).findsOne();
        check(tester.widget<UserStatusEmoji>(statusEmojiFinder)
```

test/widgets/message_list_test.dart
gnprice: Then also can drop the blank line above, and join this function's whole body in one stanza.

gnprice (commented):
Thanks! Comments below on the changes.

test/widgets/message_list_test.dart
sm-sayedi: Examining the code again, if `ImageEmojiWidget.errorBuilder` is not provided, there is no need for the second pump. Only with the first pump will the code throw the error. 
The last time that I concluded that it needs two pumps is because I used a `print` statement inside `ImageEmojiWidget.errorBuilder` with the assumption that if there is an error, then `errorBuilder` is called. But it is [triggered](https://github.com/flutter/flutter/blob/4b5a9367619c87e3c8acd8356d533b597901357f/packages/flutter/lib/src/widgets/image.dart#L1346-L1353) after a [`setState`](https://github.com/flutter/flutter/blob/4b5a9367619c87e3c8acd8356d533b597901357f/packages/flutter/lib/src/widgets/image.dart#L1195-L1198), which requires the second pump in the test code to see if it is called.

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -231,6 +245,31 @@ void main() {
...
+            check(find.textContaining('Busy')).findsNothing();
```
sm-sayedi, line 259: In places where status emoji is used as a span, I used `find.textContaining` instead of `find.text`, so if status text is shown in a text span, these tests can detect it.

test/widgets/new_dm_sheet_test.dart
```diff
@@ -285,6 +290,75 @@ void main() {
...
+      await tester.tap(findUserTile(user));
+      await tester.pump();
+
+      checkFindsTileStatusEmoji(tester, user, find.text('\u{1f6e0}'));
+      check(find.descendant(of: findUserTile(user),
+        matching: find.textContaining('Busy'))).findsNothing();
+      check(findUserChip(user)).findsOne();
+      checkFindsChipStatusEmoji(tester, user, find.text('\u{1f6e0}'));
+      check(find.descendant(of: findUserChip(user),
+        matching: find.text('Busy'))).findsNothing();
```
sm-sayedi: I still kept this check for the status texts the same as before (didn't remove `find.descendant`), as there could be status texts in two places on the same page (one for the user tile, another for the user chip).

sm-sayedi (commented):
Thank @gnprice for the review. New changes pushed.

test/widgets/recent_dm_conversations_test.dart
gnprice: Hmm, I thought I replied to this in the last round, but I don't see the reply.

It's an important part of how this test works that the emoji it's looking for here is the same one as the emoji the user has as their status emoji. So this line and the `emojiCode: '1f6e0'` a few lines above need to agree.

If this were moved inside `checkFindsStatusEmoji`, then someone reading that function would see a mysterious `\u{1f6e0}` and have to wonder why that particular code point was the one to check for; and someone reading this test case would be unaware that the `checkFindsStatusEmoji` function it's calling has a secret dependence on this detail of how the test set up its data. So both sides of that division would become impossible for a reader to properly understand without going and reading the other side too.

In short, it wouldn't really be an implementation detail — it'd be an essential part of the interface that `checkFindsStatusEmoji` presents to its callers.

If on the other hand we had a helper function that included both the work of `checkFindsStatusEmoji` and the setup step above, then it'd make sense for that helper function's source to specify which emoji to use, because that really would be an implementation detail.

There are a few ways one could organize the code so as to still ensure that both the test cases and the helpers are readable and understandable without having to go search for related code and look for hidden dependencies. But I think the organization in the current revision works well.

test/widgets/recent_dm_conversations_test.dart
```diff
@@ -176,7 +180,7 @@ void main() {
         // TODO(#232): syntax like `check(find(…), findsOneWidget)`
         final widget = tester.widget(find.descendant(
           of: find.byType(RecentDmConversationsItem),
-          matching: find.text(expectedText),
+          matching: find.textContaining(expectedText),
```
gnprice: Hmm, it's unfortunate to make all the existing tests here less specific. We do want the title that the user sees to be the title we intend — and not, for example, the intended title plus some whitespace or other added text.

I'm guessing that the reason you made this less specific is because the title now has a WidgetSpan in it for the status emoji, and that doesn't match. In that case, it'd be good to keep the check so that it allows that part to differ, but no more.

It looks like Flutter upstream already has most of the work done to support ignoring WidgetSpans in these matches — it's just missing the last step. I just drafted up an implementation of that last step, and I'll push it as an added commit to this branch. (I'll also see about sending the same functionality as a PR upstream.) Then this line can say:
```dart
          matching: findText(expectedText, includePlaceholders: false),
```

And after the functionality is upstream, it can be:
```dart
          matching: find.text(expectedText, includePlaceholders: false),
```

test/widgets/recent_dm_conversations_test.dart
gnprice: Cool — that's good for these negative checks (`.findsNothing()`), as it makes them stronger.

test/widgets/new_dm_sheet_test.dart
gnprice: Hmm. But these are still both negative checks — so it's still true that there shouldn't be "Busy" anywhere at all, right?

And saying "no 'Busy' anywhere" is a stronger requirement than "no 'Busy' in place A and no 'Busy' in place B". So I think all the same logic in https://github.com/zulip/zulip-flutter/pull/1702#discussion_r2216792820 still applies.

test/widgets/profile_test.dart
```diff
@@ -96,9 +97,20 @@ void main() {
...
+    await tester.pump();
 
     check(because: 'find user avatar', find.byType(Avatar).evaluate()).length.equals(1);
     check(because: 'find user name', find.text('test user').evaluate()).isNotEmpty();
+    final statusEmojiFinder = find.ancestor(of: find.text('\u{1f6e0}'),
+      matching: find.byType(UserStatusEmoji));
+    check(because: 'find user status emoji', statusEmojiFinder).findsOne();
+    check(tester.widget<UserStatusEmoji>(statusEmojiFinder)
+      .neverAnimate).isFalse();
+    check(because: 'find user status text', find.text('Busy')).findsOne();
```
gnprice: Let's put these checks into their own new test case, and leave this one as it is.

That will make both sets of test steps easier to understand than when they're combined into one does-everything test scenario like this.

That will also make this test a bit stronger: the existing test case will effectively check that if there is no user status, that doesn't cause this page to crash (e.g. because of a misplaced `!`, as in `userStatus.text!`).

gnprice (commented):
Thanks for the revision! I've now read through the whole thing.

Comments below on a few things that came up only in those remaining tests. I've also just pushed a couple of additional commits, as described in one of those below comments:
9aa3c6293 test: Add find.text analogue with includePlaceholders option
4eca2897e (squash) Demo using the new findText and includePlaceholders

The first should get reordered to earlier in the branch, before the point it's needed; the second does part of the conversion to use the new finder, which you can squash into the appropriate commits before doing the rest.

test/widgets/message_list_test.dart
gnprice: Cool; thanks for the explanation!

test/widgets/recent_dm_conversations_test.dart
sm-sayedi: Got it, thanks for the explanation.

sm-sayedi:
Thanks for the review @gnprice. New revision pushed; PTAL.

gnprice:
Thanks for the revision! All looks good now — merging.

PR #1703, opened 2025-07-13
unreads: Add/use locatorMap, to efficiently locate messages

### LINKED ISSUE #332: Optimize unreads model on mark-as-read

chrisbobbe:
Fixes #332.

Really the same optimization as described for zulip-mobile in
  https://github.com/zulip/zulip-mobile/issues/4684
, and makes mark-as-read take 0ms (to the nearest ms) down from 3 or 4, in my testing, and so fixes #332.

### DIFF
--- a/lib/model/unreads.dart
+++ b/lib/model/unreads.dart
@@ -41,6 +41,7 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
     required ChannelStore channelStore,
     required UnreadMessagesSnapshot initial,
   }) {
+    final locatorMap = <int, SendableNarrow>{};
     final streams = <int, TopicKeyedMap<QueueList<int>>>{};
     final dms = <DmNarrow, QueueList<int>>{};
     final mentions = Set.of(initial.mentions);
@@ -57,23 +58,34 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
         // TODO(server-10) simplify away
         (value) => setUnion(value, unreadChannelSnapshot.unreadMessageIds),
         ifAbsent: () => QueueList.from(unreadChannelSnapshot.unreadMessageIds));
+      final narrow = TopicNarrow(streamId, topic);
+      for (final messageId in unreadChannelSnapshot.unreadMessageIds) {
+        locatorMap[messageId] = narrow;
+      }
     }
 
     for (final unreadDmSnapshot in initial.dms) {
       final otherUserId = unreadDmSnapshot.otherUserId;
       final narrow = DmNarrow.withUser(otherUserId, selfUserId: core.selfUserId);
       dms[narrow] = QueueList.from(unreadDmSnapshot.unreadMessageIds);
+      for (final messageId in dms[narrow]!) {
+        locatorMap[messageId] = narrow;
+      }
     }
 
     for (final unreadHuddleSnapshot in initial.huddles) {
       final narrow = DmNarrow.ofUnreadHuddleSnapshot(unreadHuddleSnapshot,
           selfUserId: core.selfUserId);
       dms[narrow] = QueueList.from(unreadHuddleSnapshot.unreadMessageIds);
+      for (final messageId in dms[narrow]!) {
+        locatorMap[messageId] = narrow;
+      }
     }
 
     return Unreads._(
       core: core,
       channelStore: channelStore,
+      locatorMap: locatorMap,
       streams: streams,
       dms: dms,
       mentions: mentions,
@@ -84,6 +96,7 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
   Unreads._({
     required super.core,
     required this.channelStore,
+    required this.locatorMap,
     required this.streams,
     required this.dms,
     required this.mentions,
@@ -91,6 +104,12 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
   });
 
   final ChannelStore channelStore;
+
+  /// All unread messages, as: message ID → narrow ([TopicNarrow] or [DmNarrow]).
+  ///
+  /// Enables efficient [isUnread] and efficient lookups in [streams] and [dms].
+  @visibleForTesting
+  final Map<int, SendableNarrow> locatorMap;
 
   // TODO excluded for now; would need to handle nuances around muting etc.
   // int count;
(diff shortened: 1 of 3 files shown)


lib/model/unreads.dart
```diff
@@ -248,6 +263,12 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
       return;
     }
 
+    final narrow = switch (message) {
+      StreamMessage() => TopicNarrow.ofMessage(message),
+      DmMessage() => DmNarrow.ofMessage(message, selfUserId: selfUserId),
+    };
+    locatorMap[event.message.id] = narrow;
+
```
rajveermalviya: nit: maybe merge both switches? To avoid a copy of `DmNarrow` and more specifically of `allRecipientIds` list in  `DmNarrow.ofMessage`.
```dart
    switch (message) {
      case StreamMessage():
        final narrow = TopicNarrow.ofMessage(message);
        locatorMap[event.message.id] = narrow;
        _addLastInStreamTopic(message.id, message.streamId, message.topic);
      case DmMessage():
        final narrow = DmNarrow.ofMessage(message, selfUserId: selfUserId);
        locatorMap[event.message.id] = narrow;
        _addLastInDm(message.id, narrow);
    }
```

lib/model/unreads.dart
```diff
@@ -517,26 +546,23 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
...
-  void _slowRemoveAllInStreams(Set<int> idsToRemove) {
-    final newlyEmptyStreams = <int>[];
-    for (final MapEntry(key: streamId, value: topics) in streams.entries) {
-      final newlyEmptyTopics = <TopicName>[];
-      for (final MapEntry(key: topic, value: messageIds) in topics.entries) {
-        messageIds.removeWhere((id) => idsToRemove.contains(id));
-        if (messageIds.isEmpty) {
-          newlyEmptyTopics.add(topic);
-        }
+  /// Remove any of [idsToRemove] that are in  [streams].
```
rajveermalviya: nit:
```suggestion
  /// Remove any of [idsToRemove] that are in [streams].
```

chrisbobbe:

lib/model/unreads.dart
```diff
@@ -92,6 +105,11 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
 
   final ChannelStore channelStore;
 
+  /// All unread messages, as: message ID → narrow ([TopicNarrow] or [DmNarrow]).
+  ///
+  /// Enables efficient [isUnread] and efficient lookups in [streams] and [dms].
+  final Map<int, Narrow> locatorMap;
```
gnprice: This can be a bit more specific, I think:
```suggestion
  final Map<int, SendableNarrow> locatorMap;
```

Not sure the optimizer will get any useful information from that. But it seems good for matching our intent — the doc says basically the same thing a few lines above.

gnprice: In principle this is a private detail of this class. It might be convenient to access it in tests; but let's make it `@visibleForTesting`, if not private.

lib/model/unreads.dart
```diff
@@ -517,26 +543,23 @@ class Unreads extends PerAccountStoreBase with ChangeNotifier {
...
+  void _removeAllInStreams(Set<int> idsToRemove) {
+    for (final messageId in idsToRemove) {
+      final narrow = locatorMap[messageId];
+      if (narrow == null) continue;
+      if (narrow is! TopicNarrow) continue;
+
+      final messageIds = streams[narrow.streamId]?[narrow.topic];
+      if (messageIds == null) continue;
+
+      messageIds.remove(messageId);
```
gnprice: It's common that a mark-read event has all its messages from the same conversation, or many messages in each of a small number of conversations. When that happens, this version will do some repeated work on those:
* looking up the list in the two maps under `streams`;
* more significantly, scanning through the list for the conversation to find the message to remove, and then going through the remainder of the list to move the other entries up to keep the list contiguous.

The second part in particular makes it take quadratic time to mark all the messages in a long conversation as read.

To avoid that, let's have this method first go through the list of message IDs and look them up in locatorMap, and assemble a to-do list grouped by conversation. Then it can go through that to-do list and act on each conversation just once.

For an example, see the implementation of mark-as-unread above.

gnprice: (similarly for `_removeAllInDms`)

test/model/unreads_test.dart
```diff
@@ -83,14 +83,17 @@ void main() {
     final Set<int> expectedMentions = {};
     for (final message in messages) {
       if (message.flags.contains(MessageFlag.read)) {
+        check(model.locatorMap).not((it) => it.containsKey(message.id));
         continue;
       }
       switch (message) {
         case StreamMessage():
+          check(model.locatorMap)[message.id].equals(TopicNarrow.ofMessage(message));
```
gnprice: These checks don't fully cover the invariant that I'd want locatorMap to maintain: it should contain exactly the messages that are in `streams` and `dms`, and agree with those data structures about where each message is. These checks appear to not cover the possibility that some messages linger in locatorMap that aren't in this `messages` argument.

I think the cleanest way to check that would be as a separate piece of logic at the end of this function. Let's build up the whole expected locatorMap, based on the other data structures; then `deepEquals` to compare.

chrisbobbe:
Thanks!! Revision pushed.

gnprice:
Thanks! Looks good; merging.

gnprice:
Merged this earlier as:
b50f6bbab unreads: Add/use locatorMap, to efficiently locate messages

Must have been distracted, though, because apparently (says my shell history) I used a plain `git push` instead of the usual command which would have been `~/z/zulip/tools/push-to-pull-request --merge 1703`. The result is that GitHub sees it as closed instead of merged, oops.

Issue #667, opened 2024-05-13
Show read receipts


gnprice:
This Zulip feature:
https://zulip.com/help/read-receipts

The API endpoint for this feature is:
https://zulip.com/api/get-read-receipts
Unlike most Zulip features, this data isn't controlled by the event system; we just fetch it separately from the server.

See the zulip-mobile implementation for a model to work from, in `src/message/ReadReceiptsScreen.js`. We can use similar logic: in particular, the data won't live on `PerAccountStore` at all, but only in the state of the `StatefulWidget` for the read-receipts screen.

chrisbobbe:
When this is done we should update the Help Center doc. We removed instructions for the legacy app when preparing to launch the new app: zulip/zulip@a10bcdfc6394b9521fdf0b71765ba38c60f997ec

chrisbobbe:
Someone wrote to support@ in the early launch period (~1 day) requesting this feature. Bumping to M6, since this was supported in the legacy app.

Issue #1285, opened 2025-01-15
Internationalize styled strings


PIG208:
For strings like `'${mention(sender!, silent: true)} ${inlineLink('said', url)}: '` (the beginning of a quote-and-reply text), where part of it is styled with markdown syntax, or strings like

```dart
    return TextSpan(children: [
      TextSpan(text: "(unimplemented: text «", style: errorStyle),
      TextSpan(text: htmlNode.text, style: errorCodeStyle),
      TextSpan(text: "»)", style: errorStyle),
    ]);

```

where the string is fragmented with `TextStyle`'s, expressing them with the regular placeholders is challenging.

For the quote-and-reply string, we could define a string `"{user} [said]({link}): "`, but that requires the translators to write valid Markdown. We want something like `{user} <z>said</z>:`, where only `said` is translatable. In zulip-mobile, we have:

```javascript
          text={{
            text: 'Suggestion: <z-link>{suggestedServerUrl}</z-link>',
            values: {
              suggestedServerUrl: suggestion,
              'z-link': chunks => (
                <ZulipText
                  inheritFontSize
                  style={styles.suggestionTextLink}
                  onPress={() => handlePressSuggestion(suggestion)}
                >
                  {chunks}
                </ZulipText>
              ),
            },
          }}
```

(See https://github.com/zulip/zulip-flutter/pull/1148#issuecomment-2591411828)

To resolve this, we need to search for a similar API.

PIG208:
Another use case for this is the hint text for sending to general chat, which can be displayed as:

Message #channel_name > _general chat_

This also requires only part of the message to be stylized, and the current string does not support that:

```json
  "composeBoxChannelContentHint": "Message #{channel} > {topic}",
  "@composeBoxChannelContentHint": {
    "description": "Hint text for content input when sending a message to a channel",
    "placeholders": {
      "channel": {"type": "String", "example": "channel name"},
      "topic": {"type": "String", "example": "topic name"}
    }
  },
```

PR #1706, opened 2025-07-15
Show read receipts

sm-sayedi:
Support for viewing read receipts of a message.

Rebased on top of #1700 to borrow some code from.

TODO:
  - Tests

### Screenshots

<table>
    <tr>
        <td><b>Message action sheet</b></td>
        <td><b>Read receipts - Many</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

<table>
    <tr>
        <td><b>Read receipts - One</b></td>
        <td><b>Read receipts - Zero</b></td>
        <td><b>Read receipts - Failed</b></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
        <td></td>
    </tr>
</table>

### Screen recording

https://github.com/user-attachments/assets/6e6f3a50-4a90-47f4-83b5-a127389d81a4

Fixes: #667


alya:
We should probably use the same icon as in the web app here:

Screenshot 2025-07-15 at 09 59 19@2x

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,243 @@
...
+        crossAxisAlignment: CrossAxisAlignment.stretch,
+        children: [
+          Text(localizations.actionSheetReadReceipts,
+            style: TextStyle(
+              fontSize: 20,
+              height: 20 / 20,
+              color: designVariables.title,
+            ).merge(weightVariableTextStyle(context, wght: 600))),
+          if (status == FetchStatus.success && receiptCount > 0)
+            StyledText(
```
chrisbobbe: Very interesting, I didn't expect to be able to do this! Is this a solution to #1285?

sm-sayedi: Looking at the issue, I think this could be a solution for it!

lib/widgets/action_sheet.dart
```diff
@@ -839,6 +891,36 @@ class ReactionButtons extends StatelessWidget {
...
+  String label(ZulipLocalizations zulipLocalizations) {
+    return zulipLocalizations.actionSheetOptionSeeWhoReacted;
+  }
+
+  @override void onPressed() {
+    showViewReactionsSheet(pageContext, messageId: message.id);
+  }
+}
+
+class ViewReadReceiptsButton extends MessageActionSheetMenuItemButton {
```
gnprice, line 1176: An intermediate commit has:
```dart
  @override void onPressed() {
    // TODO: open read receipts sheet
  }
```

So it shows the button but the button doesn't do anything. That's a buggy state, so let's avoid it.

That commit is quite small, so we can just squash it into the later commit that gives this button something to do.

lib/widgets/action_sheet.dart
```diff
@@ -92,12 +93,43 @@ void _showActionSheet(
...
     });
 }
 
+/// A plain text widget for a bottom sheet with a multiline UI string.
+///
+/// Comes with built-in 16px horizontal padding.
+///
+/// Figma:
+///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-26993&m=dev
+class BottomSheetPlainText extends StatelessWidget {
```
gnprice: This name is very general. It's hard for me to tell what sort of situations this is meant for, or what it does.

If I want plain text, that's what `Text` does, right? It's not clear how the bottom sheet interacts with that.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,243 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final localizations = ZulipLocalizations.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return Center(
+      child: status == FetchStatus.loading
+        ? CircularProgressIndicator()
+        : status == FetchStatus.error
+          ? BottomSheetPlainText(
```
gnprice: This chain of `?:` would be best as a `switch (status)`. That way it's clear we handle all the possible values.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,243 @@
...
+
+  final List<int> userIds;
+  final FetchStatus status;
+
+  @override
+  Widget build(BuildContext context) {
+    final localizations = ZulipLocalizations.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return Center(
```
gnprice: Does this `Center` end up doing anything in the three cases other than `CircularProgressIndicator`? I suspect it doesn't.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,243 @@
...
+                  itemBuilder: (context, index) =>
+                    ReadReceiptsUserItem(context, userId: userIds[index]))));
+  }
+}
+
+
+// TODO: deduplicate the code with [ViewReactionsUserItem]
+@visibleForTesting
+class ReadReceiptsUserItem extends StatelessWidget {
+  const ReadReceiptsUserItem(this.pageContext, {
```
gnprice: nit: widget parameters should always be named (with very rare exceptions like `Text`)

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,243 @@
...
+                textAlign: TextAlign.center)
+            : InsetShadowBox(
+                top: 8,
+                bottom: 8,
+                color: designVariables.bgContextMenu,
+                child: ListView.builder(
+                  padding: EdgeInsets.symmetric(vertical: 8),
+                  itemCount: userIds.length,
+                  itemBuilder: (context, index) =>
+                    ReadReceiptsUserItem(context, userId: userIds[index]))));
```
gnprice: The child widget here calls this argument `pageContext`, but the context we're passing looks like just its own parent's context. If this context works, then I expect the child's own context would work equally well.

(This is also an example of why it's good for widget parameters to be named — if this said `pageContext: context`, then the contrast would jump out more.)

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,243 @@
...
+        ])));
+  }
+}
+
+/// https://zulip.com/api/get-read-receipts
+Future<GetReadReceiptsResult> getReadReceipts(ApiConnection connection, {
+  required int messageId,
+}) {
+  return connection.get('getReadReceipts', GetReadReceiptsResult.fromJson,
+    'messages/$messageId/read_receipts', null);
```
gnprice: This should go in lib/api/ :slightly_smiling_face:

lib/widgets/read_receipts.dart
gnprice: Hmm interesting.

This wouldn't be a satisfactory solution to all of #1285, because this appears to fire up a whole XML parser for each one of these widgets, which is going to be expensive. That's acceptable in this context, where there's only one of these at a time (and it isn't a very common part of the UI anyway). It wouldn't be OK for something that could appear in, say, message content, where you might scroll past a lot of them.

gnprice (commented):
Thanks! Here's comments from an initial skim.

lib/widgets/read_receipts.dart
sm-sayedi: It also centers the plain text in the sheet, especially vertically.🙂

lib/widgets/action_sheet.dart
sm-sayedi: What about the new name `BottomSheetInfoText`?

sm-sayedi:
Thanks @gnprice for the previous review. New changes pushed. Marked for Chris's review.

lib/widgets/action_sheet.dart
gnprice: That does suggest more information. Can you explain a bit more in the widget's doc?

sm-sayedi:
Pushed a new revision with tests included.

lib/widgets/action_sheet.dart
sm-sayedi: Updated its doc.

lib/widgets/icons.dart
```diff
@@ -48,143 +48,146 @@ abstract final class ZulipIcons {
   /// The Zulip custom icon "check".
   static const IconData check = IconData(0xf108, fontFamily: "Zulip Icons");
 
+  /// The Zulip custom icon "check_check".
+  static const IconData check_check = IconData(0xf109, fontFamily: "Zulip Icons");
```
chrisbobbe, line 52:
```
icons: Add check_check icon, from Web Figma file
```

I started a discussion on this; I don't think there's a change to make right now, but please link to it in the commit message for context: [#mobile-design > read-receipts icon in #F1706 @ 💬](https://chat.zulip.org/#narrow/channel/530-mobile-design/topic/read-receipts.20icon.20in.20.23F1706/near/2234770)

lib/widgets/action_sheet.dart
```diff
@@ -98,26 +98,31 @@ void _showActionSheet(
...
+/// A plain text widget for a bottom sheet with a multiline UI string.
 ///
-/// Assumes 8px padding below the top of the bottom sheet.
+/// Use it to present short, non-interactive explanatory messages to the user,
+/// such as an error message or other feedback.
+///
+/// Comes with built-in 16px horizontal padding.
 ///
 /// Figma:
 ///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-26993&m=dev
```
chrisbobbe:
```
action_sheet: Rename BottomSheetHeaderPlainText to BottomSheetInfoText
```

This commit decreases the number of widgets that 
- declare the job of following one specific source in the Figma,
- do that job correctly, and
- have an implementation that's easy to compare to that source. 🙂

Let's make a new widget for the new not-header thing, and keep this widget for a header. This would be an example of the principle ["prefer duplication over the wrong abstraction"](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction), I think 🙂, which Greg taught me about a while ago: https://chat.zulip.org/#narrow/sender/2187-Greg-Price/search/wrong.20abstraction

<details>
<summary>aside</summary>

I could more easily imagine `BottomSheetHeaderPlainText` growing a "title/body" variant, if that's helpful for following this part of the Figma—

image

; it would still have the "header" abstraction, which seems right. In fact, that might be helpful—the Figma's 18px horizontal padding kind of looks like a mistake to me; the plain-text version has 16px. A shared widget would be a natural place for a code comment that mentions both kinds of headers. We'd want to make sure the widget still has a clear, coherent interface, with a link to the Figma source for the new behavior.

</details>

assets/l10n/app_en.arb
```diff
@@ -171,6 +171,29 @@
...
   },
+  "actionSheetOptionViewReadReceipts": "View read receipts",
+  "@actionSheetOptionViewReadReceipts": {
+    "description": "Label for the 'View read receipts' button in the message action sheet."
+  },
+  "actionSheetReadReceipts": "Read receipts",
+  "@actionSheetReadReceipts": {
+    "description": "Title for the \"Read receipts\" bottom sheet."
+  },
+  "actionSheetReadReceiptsReadCount": "This message has been <link href=\"https://chat.zulip.org/help/read-receipts\">read</link> by {count} {count, plural, =1{person} other{people}}:",
```
chrisbobbe: We shouldn't give translators control of the URL. If the Help Center page moves, we should only have to make a one-line change in our source code, instead of going through and changing language data for all the languages. 🙂 

See web's string, can we do something like `<z-link>` there?

```
{num_of_people, plural, one {This message has been <z-link>read</z-link> by {num_of_people} person:} other {This message has been <z-link>read</z-link> by {num_of_people} people:}}
```

Also I think we should link to `https://zulip.com/help/read-receipts` instead of that page on CZO.

chrisbobbe: Also IIUC that plural syntax works, but I think we'd normally do

> {count, plural, =1{1 person} other{{count} people}}

in case that's a more helpful model when translating into a language that varies the word order among the one/two/few/many/other cases? (I'm not actually sure if such languages exist actually.)

lib/widgets/action_sheet.dart
```diff
@@ -720,6 +721,7 @@ void showMessageActionSheet({required BuildContext context, required Message mes
       ReactionButtons(message: message, pageContext: pageContext),
     if (hasReactions)
       ViewReactionsButton(message: message, pageContext: pageContext),
+    ViewReadReceiptsButton(message: message, pageContext: pageContext),
```
chrisbobbe: We should store and check the org's ["Enable read receipts" setting](https://zulip.com/help/read-receipts#configure-read-receipts-for-your-organization) before including the button; see `realm_enable_read_receipts` in https://zulip.com/api/register-queue

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+    final store = PerAccountStoreWidget.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return InkWell(
+      onTap: () => _onPressed(context),
+      splashFactory: NoSplash.splashFactory,
+      overlayColor: WidgetStateColor.resolveWith((states) =>
+        states.any((e) => e == WidgetState.pressed)
+          ? designVariables.contextMenuItemBg.withFadedAlpha(0.20)
+          : Colors.transparent),
```
chrisbobbe: How about

```dart
      overlayColor: WidgetStateColor.fromMap({
        WidgetState.pressed: designVariables.contextMenuItemBg.withFadedAlpha(0.20)
      }),
```

?

I updated my #1700 with that in response to https://github.com/zulip/zulip-flutter/pull/1700#discussion_r2232261004 🙂 (there might be other things that would be helpful to bring over from `ViewReactionsUserItem` in my current revision there)

assets/l10n/app_en.arb
gnprice: Yeah, in fact best to have the whole sentence inside the plural block — like it is in that example from web.

That way if e.g. the form of the verb needs to change too, the translator has that flexibility.

chrisbobbe: > That way if e.g. the form of the verb needs to change too, the translator has that flexibility.

nit: IIUC the translator has that flexibility already; they can choose to write the translated string however they want, including by re-scoping the plural block (or even omitting it which would probably be wrong). But by putting the whole sentence in the plural block in English, we're potentially helping by minimizing the things a translator has to think through in order to make a correct translation for their language.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+      status = FetchStatus.error;
+    } finally {
+      setState(() {});
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
+      height: 500,
```
chrisbobbe: In the view-reactions sheet, I used 400 with a `TODO(design) tune`. I'd like to be consistent between the two places unless there's a reason not to, but I don't mind which value is used; 500 is fine if you think it's better.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+class _ReadReceiptsState extends State<ReadReceipts> {
+  List<int> userIds = [];
+  FetchStatus status = FetchStatus.loading;
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    tryFetchReadReceipts();
+  }
```
chrisbobbe: This means we can end up fetching multiple times, so I think my ideal version would have `generation` logic like in lib/model/message_list.dart, to invalidate stale fetches. If results arrive out of order, we don't want to get stuck showing old results when we know we have results from a fetch we made more recently.

I think that can be a followup though—I can write up a more detailed proposal later, or perhaps just send a draft to demonstrate, once this work is merged.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+    builder: (_) {
+      return PerAccountStoreWidget(
+        accountId: accountId,
+        child: SafeArea(
+          minimum: const EdgeInsets.only(bottom: 16),
+          child: ReadReceipts(messageId: messageId)));
+    });
+}
+
+class ReadReceipts extends StatefulWidget {
```
chrisbobbe, line 41: Let's link to the Figma, so it's easier for readers to check the implementation against the spec.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+  final int receiptCount;
+  final FetchStatus status;
+
+  @override
+  Widget build(BuildContext context) {
+    final localizations = ZulipLocalizations.of(context);
+    final designVariables = DesignVariables.of(context);
+
+    return Padding(
+      padding: EdgeInsetsDirectional.fromSTEB(18, 16, 18, 8),
```
chrisbobbe: I think this is a rare case where I disagree with the Figma ([link](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=11367-20684&m=dev)). Can we change the horizontal padding to 16 and the bottom padding to 4, like in `BottomSheetHeaderPlainText` (which in `main` follows [this](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-26825&t=4L137jFprQR7UfZY-0) in the Figma)?

Here's a before/after of my proposal (from the current revision):

| Before | After |
| --- | --- |
| image | image |

You can leave a code comment explaining that we prefer consistency with that other header, with a link to this GitHub comment.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+
+class _ReadReceiptsHeader extends StatelessWidget {
+  const _ReadReceiptsHeader({required this.receiptCount, required this.status});
+
+  final int receiptCount;
+  final FetchStatus status;
+
+  @override
+  Widget build(BuildContext context) {
+    final localizations = ZulipLocalizations.of(context);
```
chrisbobbe: nit: we generally write `zulipLocalizations` instead of `localizations`; that way it's distinct from other localizations like `MaterialLocalizations`.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+            StyledText(
+              text: localizations.actionSheetReadReceiptsReadCount(receiptCount),
+              tags: {
+                'link': StyledTextActionTag((_, attrs) {
+                    PlatformActions.launchUrl(context, Uri.parse(attrs['href']!));
+                  },
+                  style: TextStyle(
+                    decoration: TextDecoration.underline,
+                    color: designVariables.link,
+                    decorationColor: designVariables.link),
```
chrisbobbe: The Figma specifies some other details about the underline style. Let's follow any of those that we can (e.g. [`decorationThickness`](https://api.flutter.dev/flutter/painting/TextStyle/decorationThickness.html), I think) and write `TODO(upstream)`s for the others.

lib/widgets/theme.dart
```diff
@@ -171,6 +171,8 @@ class DesignVariables extends ThemeExtension<DesignVariables> {
     labelEdited: const HSLColor.fromAHSL(0.35, 0, 0, 0).toColor(),
     labelMenuButton: const Color(0xff222222),
     labelSearchPrompt: const Color(0xff000000).withValues(alpha: 0.5),
+    labelTime: const Color(0x00000000).withValues(alpha: 0.49),
+    link: const Color(0xff066bd0),
```
chrisbobbe: Hmm, I didn't find this in the list of Figma variables that I usually look at.

Turns out it's defined in the "Zulip Web UI kit": https://www.figma.com/design/msWyAJ8cnMHgOMPxi7BUvA/Zulip-Web-UI-kit?node-id=0-1&p=f&vars=1&var-id=b9c211a7c212568bc84139413454dd8dd07321fb%2F833-56&m=dev

image

It has distinct light/dark values; we should use both. And maybe add a comment like

```dart
    link: const Color(0xff066bd0), // From "Zulip Web UI kit"
```

test/widgets/read_receipts_test.dart
```diff
@@ -0,0 +1,179 @@
...
+      matching: find.byType(BottomSheet))).findsOne(); // receipts sheet opened
+    check(find.byType(CircularProgressIndicator)).findsOne();
+    check(find.text('Read receipts')).findsOne();
+    check(find.text('Close')).findsOne();
+
+    await tester.pumpAndSettle();
+  }
+
+  Finder findUserItem(String fullName) => find.ancestor(of: find.text(fullName),
+    matching: find.byType(ReadReceiptsUserItem));
```
chrisbobbe: How about:

```suggestion
  Finder findUserItem(String fullName) =>
    find.widgetWithText(ReadReceiptsUserItem, fullName);
```

?

test/widgets/read_receipts_test.dart
```diff
@@ -0,0 +1,179 @@
...
+  group('success', () {
+    testWidgets('message read by many people', (tester) async {
+      final user1 = eg.user(userId: 1, fullName: 'User 1');
+      final user2 = eg.user(userId: 2, fullName: 'User 2');
+      await setupReceiptsSheet(tester, messageId: 100, users: [user1, user2],
+        prepareReceiptsResponse: () {
+          connection.prepare(
+            json: GetReadReceiptsResult(userIds: [1, 2]).toJson(),
+            delay: transitionDurationObserver.transitionDuration
+                   + const Duration(milliseconds: 100));
```
chrisbobbe: It takes some nonlocal reasoning to work out which nav-transition this is about, and I notice that we're always passing the same duration here.

How about replacing `prepareReceiptsResponse` with `prepareReceiptsResponseSuccess(List<int> userIds)` and `prepareReceiptsResponseError(Object? httpException)`, and control the delay in `setupReceiptsSheet`?

test/widgets/read_receipts_test.dart
```diff
@@ -0,0 +1,179 @@
...
+      check(connection.lastRequest).isA<http.Request>()
+        ..method.equals('GET')
+        ..url.path.equals('/api/v1/messages/100/read_receipts');
+
+      await tester.tap(findUserItem('User 1'));
+      await transitionDurationObserver.pumpPastTransition(tester);
+      check(find.byWidgetPredicate((widget) => switch(widget) {
+        ProfilePage(userId: 1) => true,
+        _                      => false,
+      })).findsOne();
```
chrisbobbe: Can be two lines instead of four I think 🙂

```dart
      check(find.byWidgetPredicate((widget) => widget is ProfilePage && widget.userId == 1))
        .findsOne();
```

chrisbobbe (commented):
And here's a full review of that last commit. Generally looks great; small comments below.

assets/l10n/app_en.arb
gnprice: True. So it's more that we're giving them a model that makes it easier to see how to apply that flexibility.

lib/widgets/read_receipts.dart
gnprice: What are the reasons we would want to do a new fetch because a dependency changed?

I think it may only be if there's a new store. So this could go in an onNewStore method instead.

Then the generation-like logic can just be: when getting the response, check if the store we had at the start of the request is still the current store (with `identical`). If not, ignore the response.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,215 @@
...
+    tryFetchReadReceipts();
+  }
+
+  Future<void> tryFetchReadReceipts() async {
+    final store = PerAccountStoreWidget.of(context);
+    try {
+      final result = await getReadReceipts(store.connection, messageId: widget.messageId);
+      // TODO(i18n): add locale-aware sorting
+      userIds = result.userIds.sortedByCompare(
+        (id) => store.userDisplayName(id),
```
gnprice: Relatedly: in general we never want to keep a reference to the store across an await and use it again after, because the old store might already have been disposed. Instead, after an await we should look up the store afresh using a context.

lib/widgets/read_receipts.dart
sm-sayedi: Yeah. This`SizedBox` is a little bit different from the `SizedBox` in view-reactions sheet. Here, the height of the view-receipts sheet is 500, and in view-reactions sheet, the height of the user-item list is 400 (excluding the reactions-header and dismiss-button). In fact, the height of that sheet is 518px (I can make this sheet's size 518 to make it consistent). I wrapped `SizedBox` on the entire sheet (instead of the user-item list) so that when there are no read receipts, the sheet's height is not changed.

Maybe the same approach could be applied to view-reactions sheet to address the following comment:
https://github.com/zulip/zulip-flutter/blob/98b94bd2a0685606272882e089441000f8680952/lib/widgets/emoji_reaction.dart#L770-L779

lib/widgets/read_receipts.dart
sm-sayedi: I think there's an important difference between the two thicknesses in Figma and Flutter, and I am not sure how we can map between them. In Figma, it's a percentage of the font size of the text to which the decoration belongs. In Flutter, it is a multiplier of the thickness defined by the font.

lib/widgets/action_sheet.dart
sm-sayedi: > I could more easily imagine BottomSheetHeaderPlainText growing a "title/body" variant,...

Should we create a new widget for the header in this PR, with a name like `BottomSheetHeader`? One distinction of this from `BottomSheetHeaderPlainText` is that the latter has a different styling than the title/body of this header widget.

(Thanks for the link to that interesting blog post! 🙂)

sm-sayedi:
Thanks @chrisbobbe for the review! New revision pushed, PTAL.

lib/widgets/read_receipts.dart
chrisbobbe: Sure, that's reasonable. In zulip-mobile I implemented polling to refresh this page every 15s (IIRC), and generic `generation` logic would help if we did that as a followup.

But arguably the polling wasn't really needed in zulip-mobile, and it's maybe less compelling in zulip-flutter, where the view is more clearly ephemeral, not something you'd necessarily expect to stay up-to-date if you left it foregrounded in the app. (A modal bottom sheet vs. a whole full-screen page with a back button.)

lib/widgets/read_receipts.dart
gnprice: I've written up a small TextWithLink widget which does the thing we need here more directly. Pushing that to this branch as a pair of additional commits on top:
c33ee6183 text: Introduce TextWithLink widget
93e8919d7 (squash) Use our TextWithLink instead of package:styled_text


I think for the other cases we need for #1285, I'll be happier with further developing that widget than with what package:styled_text is doing. In particular the thing that I feel least confident about in what's missing from TextWithLink is this:
```dart
/// TODO(#1285): Integrate this with l10n so that the markup can be parsed
///   from the constant translated string, with placeholders for any variables,
///   rather than the string that results from interpolating variables.
///   That way it'll be fine to interpolate variables with arbitrary text.
```
but package:styled_text doesn't handle that either.

gnprice:
Posted a comment a few minutes ago on a subthread above which should also be made visible here in chronological order on the PR thread. So here's a link https://github.com/zulip/zulip-flutter/pull/1706#discussion_r2252435627 and a copy:

> I've written up a small TextWithLink widget which does the thing we need here more directly. Pushing that to this branch as a pair of additional commits on top:
> [c33ee61](https://github.com/zulip/zulip-flutter/commit/c33ee61835f406bd75b419018edb5f8a50982ece) text: Introduce TextWithLink widget
> [93e8919](https://github.com/zulip/zulip-flutter/commit/93e8919d75db545df45dfe4d36b34c887dc2299c) (squash) Use our TextWithLink instead of package:styled_text
> 
> I think for the other cases we need for #1285, I'll be happier with further developing that widget than with what package:styled_text is doing. In particular the thing that I feel least confident about in what's missing from TextWithLink is this:
> 
> ```dart
> /// TODO(#1285): Integrate this with l10n so that the markup can be parsed
> ///   from the constant translated string, with placeholders for any variables,
> ///   rather than the string that results from interpolating variables.
> ///   That way it'll be fine to interpolate variables with arbitrary text.
> ```
> 
> but package:styled_text doesn't handle that either.

lib/widgets/read_receipts.dart
chrisbobbe: Odd, thanks for noticing that.

From Flutter's [`decorationThickness` doc](https://api.flutter.dev/flutter/painting/TextStyle/decorationThickness.html), it does seem like you can provide a constant (like 1.5 or 2.0 or whatever) and the thickness you actually get will change with the font size. That seems desirable.

But if it's true that font files always include a preferred underline stroke-width value (as that doc says), I haven't been able to find that value for the font we're using here, Source Sans 3. 🤷‍♂️

sm-sayedi:
Thanks @gnprice for the new code. Changes pushed.

lib/widgets/text.dart
```diff
@@ -415,3 +418,108 @@ TextBaseline localizedTextBaseline(BuildContext context) {
...
+            decorationStyle: TextDecorationStyle.solid,
+            // We use the default value for this, as there's no obvious
+            // way to map the thickness value from Figma design as it is
+            // a percentage of the font size.
+            decorationThickness: 1,
+            // decorationOffset: // TODO(upstream): https://github.com/flutter/flutter/issues/30541
+            color: designVariables.link,
+            decorationColor: designVariables.link)),
+        TextSpan(text: match.group(3)),
+      ]);
```
chrisbobbe: This parent/child structure feels odd; can we make the three parts be siblings instead?

```dart
      span = TextSpan(children: [
        TextSpan(text: match.group(1)),
        TextSpan(text: match.group(2), /* etc. */),
        TextSpan(text: match.group(3)),
      ]);
```

lib/widgets/action_sheet.dart
```diff
@@ -126,6 +127,41 @@ class BottomSheetHeaderPlainText extends StatelessWidget {
...
+///
+/// Use it to present short, non-interactive explanatory messages to the user,
+/// such as an error message or other feedback.
+///
+/// Comes with built-in 16px horizontal padding.
+///
+/// Style-wise, this mostly follows the design of [BottomSheetHeaderPlainText],
+/// as there is no design for this in Figma right now.
+// TODO(design) create
+class BottomSheetInfoText extends StatelessWidget {
```
chrisbobbe: Thinking more about this, I think my preference is:

- Declare an API that's explicitly about "empty content". One sign of a nice API is that, when you read about it, it's clear early on whether it's something you want to use or not. This `BottomSheetInfoText` widget catches the attention of a would-be caller who wants to show some "info text", but then it's not clear whether this or `BottomSheetHeaderPlainText` is the better way to do that—especially because this says the style is modeled on that header widget :)
- Instead of `BottomSheetInfoText`, I suggest basically duplicating `PageBodyEmptyContentPlaceholder`, adapted for use in a bottom sheet, and with bidirectional links. Then:
  - "Empty content placeholder" answers the "should-I-use-this" question more quickly than "info text".
  - It actually sort of comes from a Figma frame, though indirectly 🙂

For example:

<details>

```dart
/// A "no content here" message for when a bottom sheet has no content to show.
///
/// Show this below a [BottomSheetHeader] if present.
///
/// See also:
///  * [PageBodyEmptyContentPlaceholder], for a similar element to use in
///    pages on the home screen.
// TODO(design) we don't yet have a design for this;
//   it was ad-hoc and modeled on [PageBodyEmptyContentPlaceholder].
class BottomSheetEmptyContentPlaceholder extends StatelessWidget {
  const BottomSheetEmptyContentPlaceholder({super.key, required this.message});

  final String message;

  @override

    return Padding(
      padding: EdgeInsets.fromLTRB(24, 48, 24, 16),
      child: Align(
        alignment: Alignment.topCenter,
        child: Text(
          textAlign: TextAlign.center,
          style: TextStyle(
            color: designVariables.labelSearchPrompt,
            fontSize: 17,
            height: 23 / 17,
          ).merge(weightVariableTextStyle(context, wght: 500)),
          message)));
  }
}
```

</details>

Here's how it would look:

| Before | After |
| --- | --- |
| image | image |
| image | image |

I'll push a draft commit with that example.

chrisbobbe (commented):
Thanks! Comments below, and I've pushed four new commits:

f92e7f9bb action_sheet [nfc]: Make BottomSheetHeaderPlainText more flexible; rename
9ee2e3cef action_sheet: Implement BottomSheetEmptyContentPlaceholder, to use soon
bad574730 read_receipts [nfc]: Use BottomSheetHeader for read-receipts header
4eacb89e1 read_receipts: Use BottomSheetEmptyContentPlaceholder for empty-content

These are suggestions; if they make sense (see my comments), please rearrange the branch
- to put the first two of those before the main commit that implements the feature, and
- squash the last two commits into that main commit.

Excited to have this soon! 🎉

chrisbobbe:
Oh, another thought: I mentioned before that the link on the word "read" should be to the Help Center page on zulip.com, but…how about the Help Center page on the current realm, so using `store.tryResolveUrl`?

sm-sayedi:
Thanks @chrisbobbe for the review and the code. New revision pushed.

Also, it would be great to have your thoughts on https://github.com/zulip/zulip-flutter/pull/1706#discussion_r2249350524 about making the height of view-reactions and view-read-receipts action sheets the same. 🙂

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,213 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
+
+    WidgetBuilderFromTextStyle? headerSubtitleBuilder;
+    if (status == FetchStatus.success && receiptCount > 0) {
+      headerSubtitleBuilder = (TextStyle style) => TextWithLink(
+        onTap: () {
+          final url = PerAccountStoreWidget.of(context).tryResolveUrl(ReadReceipts._helpCenterReference)
+            ?? Uri.parse(ReadReceipts._helpCenterUrl);
```
chrisbobbe: Can resolving the URL ever actually fail? I don't think it can; we control the path part ("/help/read-receipts"), and we have the realm URL as a `Uri` object, which means it passed some level of validation. I think we can do `tryResolveUrl(…)!` (throwing an error if it fails), instead of opening a page on zulip.com that might have wrong information because it was written to correspond with a different Zulip server version.

lib/widgets/read_receipts.dart
chrisbobbe: In the current revision, I believe this `Center` only has an effect in the `FetchStatus.loading` state, and otherwise it does nothing, which is confusing. (`BottomSheetEmptyContentPlaceholder` takes care of positioning the "no-read-receipts" message in the error and success cases.)

I think it would be reasonable to give that new widget the job of showing a loading indicator as appropriate, and then this `Center` can go away. Could give it a `loading` param, requiring either that or `message` but not both`, and have it show the loading spinner in the place where message would go; something like:

<details>

```diff
--- lib/widgets/action_sheet.dart
+++ lib/widgets/action_sheet.dart
@@ -182,7 +182,11 @@ class BottomSheetHeader extends StatelessWidget {
   }
 }
 
-/// A "no content here" message for when a bottom sheet has no content to show.
+/// A placeholder for when a bottom sheet has no content to show.
+///
+/// Pass [message] for a "no-content-here" message,
+/// or pass true for [loading] if the content hasn't finished loading yet,
+/// but don't pass both.
 ///
 /// Show this below a [BottomSheetHeader] if present.
 ///
@@ -192,26 +196,35 @@ class BottomSheetHeader extends StatelessWidget {
 // TODO(design) we don't yet have a design for this;
 //   it was ad-hoc and modeled on [PageBodyEmptyContentPlaceholder].
 class BottomSheetEmptyContentPlaceholder extends StatelessWidget {
-  const BottomSheetEmptyContentPlaceholder({super.key, required this.message});
+  const BottomSheetEmptyContentPlaceholder({
+    super.key,
+    this.message,
+    this.loading = false,
+  }) : assert(message == null || !loading);
 
-  final String message;
+  final String? message;
+  final bool loading;
 
   @override
   Widget build(BuildContext context) {
     final designVariables = DesignVariables.of(context);
 
-    return Padding(
-      padding: EdgeInsets.fromLTRB(24, 48, 24, 16),
-      child: Align(
-        alignment: Alignment.topCenter,
-        child: Text(
+    final child = loading
+      ? CircularProgressIndicator()
+      : Text(
           textAlign: TextAlign.center,
           style: TextStyle(
             color: designVariables.labelSearchPrompt,
             fontSize: 17,
             height: 23 / 17,
           ).merge(weightVariableTextStyle(context, wght: 500)),
-          message)));
+          message!);
+
+    return Padding(
+      padding: EdgeInsets.fromLTRB(24, 48, 24, 16),
+      child: Align(
+        alignment: Alignment.topCenter,
+        child: child));
   }
 }
```

</details>

chrisbobbe (commented):
Ah yeah, thanks for the ping on the height question :). I think the height and small inconsistency is fine to merge; I don't want to block this on further changes about the height, and I'm working on something locally that I might send as a followup.

Small comments below.

sm-sayedi:
Thanks for the review. New changes pushed.

chrisbobbe:

lib/widgets/action_sheet.dart
```diff
@@ -102,31 +103,128 @@ void _showActionSheet(
...
 ///
-/// Figma:
+/// Figma; just text no title:
 ///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=3481-26993&m=dev
-class BottomSheetHeaderPlainText extends StatelessWidget {
-  const BottomSheetHeaderPlainText({super.key, required this.text});
-
-  final String text;
+///
+/// Figma; title and text:
```
gnprice: Looks like this and just above say "text" for what the code calls "subtitle".

lib/widgets/emoji_reaction.dart
```diff
@@ -822,7 +822,7 @@ class ViewReactionsHeader extends StatelessWidget {
     if (reactions == null || reactions.aggregated.isEmpty) {
       return Padding(
         padding: const EdgeInsets.only(top: 8),
-        child: BottomSheetHeaderPlainText(text: zulipLocalizations.seeWhoReactedSheetNoReactions),
+        child: BottomSheetHeader(subtitle: zulipLocalizations.seeWhoReactedSheetNoReactions),
```
gnprice: It feels odd to me to have something called a "subtitle" when there's no title.

Maybe "note"? "message"?

I guess "message" might sound like it's referring to a Zulip message; but we already use that name for dialogs and it seems fine. In any case "note" doesn't have that issue.

lib/widgets/action_sheet.dart
```diff
@@ -102,31 +103,128 @@ void _showActionSheet(
...
+      _                                        => null,
+    };
+
+    final baseSubtitleStyle = TextStyle(
+      color: designVariables.labelTime,
+      fontSize: 17,
+      height: 22 / 17);
+
+    final effectiveSubtitle = switch ((buildSubtitle, subtitle)) {
+      (WidgetBuilderFromTextStyle build, null) => build(baseSubtitleStyle),
```
gnprice: Hmm, interesting idea passing the desired style as an argument to the builder callback.

lib/widgets/action_sheet.dart
```diff
@@ -102,31 +103,128 @@ void _showActionSheet(
...
+      (null, String data)                      => Text(style: baseSubtitleStyle, data),
+      _                                        => null,
```
gnprice: nit: can make less verbose by leaving out the types
```suggestion
      (final build?, null) => build(baseSubtitleStyle),
      (null,  final data?) => Text(style: baseSubtitleStyle, data),
      _                    => null,
```

lib/widgets/action_sheet.dart
```diff
@@ -102,31 +103,128 @@ void _showActionSheet(
...
+    };
+
+    // (should have been caught by `assert` in constructor)
+    if (effectiveTitle == null && effectiveSubtitle == null) throw ArgumentError();
```
gnprice: Hmm, I don't like throwing from a build method, in release mode, if we can avoid it. Certainly not going out of our way to detect a reason to throw.

What happens if this case somehow occurs and we don't have this line? The Column will have no children — so probably it's just empty, right? That's a lot better than crashing.

OTOH it'd be fine to have an assert here. That'd affect debug mode only, the same way as the assert in the constructor.

lib/widgets/action_sheet.dart
```diff
@@ -102,31 +103,128 @@ void _showActionSheet(
...
+          textAlign: TextAlign.center,
           style: TextStyle(
-            color: designVariables.labelTime,
+            color: designVariables.labelSearchPrompt,
             fontSize: 17,
-            height: 22 / 17),
-          text)));
+            height: 23 / 17,
+          ).merge(weightVariableTextStyle(context, wght: 500)),
+          message!);
```
gnprice, line 218: This `!` should get a corresponding assertion earlier, at the constructor.

lib/widgets/text.dart
```diff
@@ -415,3 +418,109 @@ TextBaseline localizedTextBaseline(BuildContext context) {
...
+      span = TextSpan(children: [
+        TextSpan(text: match.group(1)),
+        TextSpan(text: match.group(2), recognizer: _recognizer,
+          style: TextStyle(
+            decoration: TextDecoration.underline,
+            decorationStyle: TextDecorationStyle.solid,
+            // We use the default value for this, because there's no obvious
+            // way to map the thickness value from the Figma design as it is
+            // a percentage of the font size.
+            decorationThickness: 1,
```
gnprice: This seems pretty solvable to me if we wanted to, but not a priority to solve.

Since 1 is the default, we also don't need to specify it here. So:
```suggestion
            // TODO(design): work out what decorationThickness to use;
            //   the Figma design calls for 4% of the font sise, but Flutter
            //   expects it as a ratio of the font's default stroke thickness.
            // decorationThickness: 1, // (the default)
```

lib/widgets/text.dart
```diff
@@ -415,3 +418,109 @@ TextBaseline localizedTextBaseline(BuildContext context) {
...
+      // (It's important not to try to interpret any markup here, in case it
+      // comes buggily from user-controlled text.)
+      span = TextSpan(text: widget.markup);
+    } else {
+      span = TextSpan(children: [
+        TextSpan(text: match.group(1)),
+        TextSpan(text: match.group(2), recognizer: _recognizer,
+          style: TextStyle(
+            decoration: TextDecoration.underline,
+            decorationStyle: TextDecorationStyle.solid,
```
gnprice: This is the default, right? Can leave it that way.

(We're not going to have a DefaultTextStyle setting a `decorationStyle`; this field doesn't make sense without `decoration`.)

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,208 @@
...
+    builder: (_) {
+      return PerAccountStoreWidget(
+        accountId: accountId,
+        child: SafeArea(
+          minimum: const EdgeInsets.only(bottom: 16),
+          child: ReadReceipts(messageId: messageId)));
+    });
+}
+
+/// The read receipts sheet.
```
gnprice: nit:
```suggestion
/// The read-receipts sheet.
```

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,208 @@
...
+///
+/// Figma link:
+///   https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=11367-20647&t=lSnHudU6l7NWx0Fa-0
+class ReadReceipts extends StatefulWidget {
+  const ReadReceipts({super.key, required this.messageId});
+
+  final int messageId;
+
+  static const _helpCenterReference = '/help/read-receipts';
```
gnprice: nit: "reference" is a bit vague about what sort of thing this is.

I think "relative URL" is a good term. The official term in the spec would be "relative-URL string":
https://url.spec.whatwg.org/#relative-url-string

and here the type says it's a string so we don't need to repeat that part. Therefore:

```suggestion
  static const _helpCenterRelativeUrl = '/help/read-receipts';
```

(In the now-obsolete RFCs that previously defined the concept of a URL, [this was](https://www.rfc-editor.org/rfc/rfc3986#section-4) a "URI-reference" and more specifically a "relative reference". Perhaps that's what you had in mind with this name.)

gnprice: separate nit: this private const can go on the one class that uses it, and then the reference to it doesn't need to repeat the class name.

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,208 @@
...
+    try {
+      final result = await getReadReceipts(store.connection, messageId: widget.messageId);
+
+      if (!context.mounted) return;
+      final storeNow = PerAccountStoreWidget.of(context);
+      if (!identical(store, storeNow)) return;
+
+      // TODO(i18n): add locale-aware sorting
+      userIds = result.userIds.sortedByCompare(
+        (id) => storeNow.userDisplayName(id),
```
gnprice: Since `store` and `storeNow` are identical, this is equivalent:
```suggestion
        store.userDisplayName,
```

and keeps to our convention of always calling the store just `store`.

(This version also incorporates a nit that this can use a tearoff instead of a fresh closure.)

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,208 @@
...
+              padding: EdgeInsets.symmetric(vertical: 8),
+              itemCount: userIds.length,
+              itemBuilder: (context, index) =>
+                ReadReceiptsUserItem(userId: userIds[index])))
+    };
+  }
+}
+
+
+// TODO: deduplicate the code with [ViewReactionsUserItem]
```
gnprice: nit: double blank line

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,208 @@
...
+@visibleForTesting
+class ReadReceiptsUserItem extends StatelessWidget {
```
gnprice, line 163: This comment should get a twin at the other end, pointing back to this one.

That way when editing the other one, we might notice and look at this one to edit it the same way (or to go ahead and do this deduplication).

chrisbobbe:
Ah, thanks Greg! @sm-sayedi would you mind taking care of Greg's feedback on the code I wrote? That seems simpler process-wise than if we both try to make our own edits to the branch, and I'd appreciate it :)

sm-sayedi:
@chrisbobbe Sure, doing it now.

lib/widgets/action_sheet.dart
sm-sayedi: Can we remove the assert here, as it will already be caught in the constructor anyway?

lib/widgets/text.dart
sm-sayedi: > This seems pretty solvable to me if we wanted to,...

Interesting, would love to know about the solution, maybe in one of our weekly calls, if possible.🙂

sm-sayedi:
Thanks @gnprice for the review. New changes pushed, PTAL.

lib/widgets/action_sheet.dart
```diff
@@ -102,31 +103,125 @@ void _showActionSheet(
...
+      _                    => null,
+    };
+
+    final baseSubtitleStyle = TextStyle(
+      color: designVariables.labelTime,
+      fontSize: 17,
+      height: 22 / 17);
+
+    final effectiveMessage = switch ((buildMessage, message)) {
+      (final build?, null) => build(baseSubtitleStyle),
```
gnprice: mismatched name:
```suggestion
    final effectiveMessage = switch ((buildMessage, message)) {
      (final build?, null) => build(baseMessageStyle),
```

lib/widgets/text.dart
gnprice: I think the key step is just to go inspect the font and determine what its default stroke thickness is, relative to the font size.

Here's a thread from last year where Karl inspected our standard font for some of the other details found in it:
[#design > info density: inline emoji presentation @ 💬](https://chat.zulip.org/#narrow/channel/101-design/topic/info.20density.3A.20inline.20emoji.20presentation/near/1798391)

lib/widgets/read_receipts.dart
```diff
@@ -0,0 +1,207 @@
...
+        },
+        style: style,
+        markup: zulipLocalizations.actionSheetReadReceiptsReadCount(receiptCount));
+    }
+
+    return Padding(
+      padding: const EdgeInsets.only(top: 8),
+      child: BottomSheetHeader(
+        title: zulipLocalizations.actionSheetReadReceipts,
+        buildMessage: headerSubtitleBuilder));
```
gnprice: mismatched names

lib/widgets/emoji_reaction.dart
```diff
@@ -1036,7 +1036,8 @@ class ViewReactionsUserList extends StatelessWidget {
   }
 }
 
-@visibleForTesting
+// TODO: deduplicate the code with [ReadReceiptsUserItem]
+  @visibleForTesting
 class ViewReactionsUserItem extends StatelessWidget {
```
gnprice: nit: stray indent

sm-sayedi:
Thanks for the review. New revision pushed.

gnprice:
Thanks! Looks good; merged.

Issue #1717, opened 2025-07-18
Message list on "general chat" fails to get new messages


gnprice:
The original reported symptom is: you open a notification that takes you to a conversation, then try sending messages there, and the messages don't appear. This only happens when the conversation is in the "general chat" topic (or more generally, the value of `realmEmptyTopicDisplayName`).

The same behavior can also happen without a notification, if you follow a link like [#test here > general chat](https://chat.zulip.org/#narrow/channel/7-test-here/topic/general.20chat), as in `#**test here>general chat**`. (Things work fine if the link is instead [#test here > *general chat*](https://chat.zulip.org/#narrow/channel/7-test-here/topic/), as in `#**test here>**` — the form the Zulip clients prefer to generate.)

To fix, when opening a message list to a topic narrow, we should normalize the topic. We want the topic to be such that when a message is sent to it, it comes back on that same topic — so probably this means calling `TopicName.processLikeServer`.

(That also suggests that the same symptom would affect a link to "(no topic)" the same way as a link to "general chat". I haven't tested to confirm that.)

### References
* User report, and debugging thread: [#mobile > flutter app didn't show message after posting @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/flutter.20app.20didn't.20show.20message.20after.20posting/near/2215499)

PR #1718, opened 2025-07-18
Normalize topic name when opening message list

rajveermalviya:
Fixes: #1717

### DIFF
--- a/test/widgets/message_list_test.dart
+++ b/test/widgets/message_list_test.dart
@@ -158,14 +158,17 @@ void main() {
       check(state.narrow).equals(ChannelNarrow(stream.streamId));
     });
 
-    testWidgets('MessageListPageState.narrow (general chat)', (tester) async {
+    testWidgets('narrow gets normalized from "general chat"', (tester) async {
+      // Regression test for: https://github.com/zulip/zulip-flutter/issues/1717
       final stream = eg.stream();
+      // Open the page on a topic with the literal name "general chat".
       final topic = eg.defaultRealmEmptyTopicDisplayName;
       final topicNarrow = eg.topicNarrow(stream.streamId, topic);
       await setupMessageListPage(tester, narrow: topicNarrow,
         streams: [stream],
         messages: [eg.streamMessage(stream: stream, topic: topic, content: "<p>a message</p>")]);
       final state = MessageListPage.ancestorOf(tester.element(find.text("a message")));
+      // The page's narrow has been updated; the topic is "", not "general chat".
       check(state.narrow).equals(eg.topicNarrow(stream.streamId, ''));
     });
 


lib/notifications/open.dart
```diff
@@ -112,10 +112,19 @@ class NotificationOpenService {
...
       accountId: account.id,
-      // TODO(#1565): Open at specific message, not just conversation
-      narrow: data.narrow);
+      page: Builder(builder: (context) {
+        final store = PerAccountStoreWidget.of(context);
+        var narrow = data.narrow;
+        if (narrow is TopicNarrow) {
+          narrow = narrow.processTopicLikeServer(
+            zulipFeatureLevel: store.zulipFeatureLevel,
+            realmEmptyTopicDisplayName: store.realmEmptyTopicDisplayName);
```
gnprice: Hmm, I see. The normalization is a bit sticky to do at this spot, because when this method is running we don't yet have a per-account store. (In particular the app may have really not yet loaded the data from the server.) So we don't know realmEmptyTopicDisplayName, and have to make a route before we've done the normalization.

I think it'd be best to let this continue to use a proper MessageListPage route, though, and let MessageListPage or MessageList take the responsibility for normalizing the narrow. They already have the capability to do something similar, because the initial fetch might discover the topic has been moved. That's even later than the point where we have the store and can do this normalization.

Specifically I think _MessageListState.onNewStore looks like a good place to handle this.

lib/widgets/message_list.dart
```diff
@@ -772,8 +772,15 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
   }
 
   void _initModel(PerAccountStore store, Anchor anchor) {
+    // Normalize topic name if this is a TopicNarrow. See #1717.
+    var narrow = widget.narrow;
+    if (narrow is TopicNarrow) {
+      narrow = narrow.processTopicLikeServer(
```
gnprice, line 779: This should also notify the MessageListPage about the change.

In particular that will cause the app bar to put "general chat" in italics.

rajveermalviya: We already notify (albeit a bit later) as a side effect of `model.addListener(_modelChanged)` below.
If we want to do it here we would need to use something like, because ancestor calls `setState` in  `onNarrowChanged`:
```dart
      SchedulerBinding.instance.scheduleFrameCallback((_) {
        widget.onNarrowChanged(narrow);
      });
```

gnprice: I see. Yeah, I think that would be better — that will make sure it happens for the next frame, rather than waiting until the model updates for some other reason.

That explicit approach is also good because it means a reader of this code doesn't have to wonder about how and when the change makes it to the page widget.

rajveermalviya: Updated.

lib/widgets/message_list.dart
```diff
@@ -772,8 +773,20 @@ class _MessageListState extends State<MessageList> with PerAccountStoreAwareStat
...
+        zulipFeatureLevel: store.zulipFeatureLevel,
+        realmEmptyTopicDisplayName: store.zulipFeatureLevel > 334
+          ? store.realmEmptyTopicDisplayName
+          : null);
+      SchedulerBinding.instance.scheduleFrameCallback((_) {
+        widget.onNarrowChanged(narrow);
```
gnprice: This is only needed if the narrow is different from its original value — let's avoid calling it when nothing changed.

rajveermalviya: Done.

lib/widgets/message_list.dart
gnprice: (In manual testing, I tried commenting that call back out to see the effect — the page ends up continuing to show the old value until the message list finishes fetching the initial messages from the server. That means "general chat" isn't in italics; and a bit more substantively, any applicable topic-visibility icon is missing.

So indeed calling `onNarrowChanged` explicitly here prevents a user-visible glitch, in addition to making it easier for a reader to become confident that things work.)

test/widgets/message_list_test.dart
```diff
@@ -158,6 +158,17 @@ void main() {
...
 
+    testWidgets('MessageListPageState.narrow (general chat)', (tester) async {
+      final stream = eg.stream();
+      final topic = eg.defaultRealmEmptyTopicDisplayName;
+      final topicNarrow = eg.topicNarrow(stream.streamId, topic);
+      await setupMessageListPage(tester, narrow: topicNarrow,
+        streams: [stream],
+        messages: [eg.streamMessage(stream: stream, topic: topic, content: "<p>a message</p>")]);
+      final state = MessageListPage.ancestorOf(tester.element(find.text("a message")));
+      check(state.narrow).equals(eg.topicNarrow(stream.streamId, ''));
```
gnprice: I think this test works, at the level that if we were to re-introduce this bug, the test would fail.

There's another layer of what a test needs to do to be fully successful at its job: when the test does fail in that situation, it needs to tell its story. (I go into more detail on this in the talk I gave last month at Fluttercon; hoping for the talk video to get posted soon so I can share it.) The person who's made the change needs to be able to look at the test and figure out why the test's author thought it was important for things to behave this way; they need to be able to determine whether the breakage is pointing at a real bug in their change, or just means the test needs to be updated to match their change. If they can't tell that, then it's likely they'll just edit the test to match the new (broken) behavior, and the test won't have done its job of preventing the regression.

I think this version doesn't really tell the story. It looks very much like the test above it — but that test is just testing some boring plumbing. In this test, there's a critical contrast where a piece of data (the topic name) is different at one stage from an earlier stage; and there's a reason why that particular change should happen and why it's important. The contrast is easy to miss when scanning this code; and once spotted, without the reason, it almost looks like a bug instead of an important feature.

I'll add a commit to fix this aspect up and demonstrate telling the test's story.

gnprice (commented):
Thanks for fixing this! The logic looks good now. Two comments below.

I'll add that commit clarifying the test, and merge.

PR #1720, opened 2025-07-19
Render KaTeX by default

gnprice:
Here's a version of the commit I've included in each release recently, atop (previous versions of) #1559, to enable our KaTeX support. Now that #1559 and its preceding PRs (#1698, and others before that) are merged to main, this is ready to go to main too.

This was a minimal version for the release, just flipping the flag's default to true. But in general we should remove experimental flags from the codebase as soon as they're no longer relevant, to keep complexity from building up. (See the doc on [GlobalSettingType.experimentalFeatureFlag].) @rajveermalviya would you write a follow-up commit that removes this flag entirely, and cleans up the logic elsewhere which that makes possible to clean up?

#### 5287d39b9 content: Render KaTeX by default, the way we have in recent releases

This is cherry-picked from commits we included in each of the last
several releases, starting with 8f3723768 in v0.0.31.

### DIFF
--- a/lib/model/settings.dart
+++ b/lib/model/settings.dart
@@ -178,7 +178,7 @@ enum BoolGlobalSetting {
   upgradeWelcomeDialogShown(GlobalSettingType.internal, false),
 
   /// An experimental flag to toggle rendering KaTeX content in messages.
-  renderKatex(GlobalSettingType.experimentalFeatureFlag, false),
+  renderKatex(GlobalSettingType.experimentalFeatureFlag, true),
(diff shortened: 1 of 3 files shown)


gnprice:
> @rajveermalviya would you write a follow-up commit that removes this flag entirely, and cleans up the logic elsewhere which that makes possible to clean up?

I've filed #1728 to track this follow-up task.

Merging this change now, in the interest of further syncing up the main branch with what we've had in recent releases.

PR #1722, opened 2025-07-19
Refactor KaTeX parsing of inline styles and vlists; normalize survey output a bit more

gnprice:
Now that the KaTeX support we've had in recent releases is merged into main (with #1559) (*), I've gone and made a number of the refactors that I'd been contemplating for this code but hadn't wanted to do before those were merged. In this branch:
* Parse each element's inline styles in a more direct way, letting the parsing logic for different types of elements apply their respective context of what styles are expected. This also means less building and copying of generic KatexSpanStyles objects.
* Simplify some of the vlist parsing code, notably around margins.
* Split the parsing of struts and vlists into their own respective methods.
* At the start of the branch, a few adjustments to make the output of `tools/content/check-features katex-check` more stable and easier to compare across changes to the implementation. This helped me validate that on an empirical corpus, none of the above changes caused any KaTeX expressions to stop being successfully parsed.

In particular I ran `tools/content/check-features katex-check` near the start of the branch, after the first three commits:
8c4106d84 katex [nfc]: Show line numbers only on unknown hard-fails, not others
d2696287f katex [nfc]: Add messages for remaining hard-fail cases seen in corpus
e74557bba tools/content [nfc]: Tie-break on reason text when number of failures equal

and again at the end of the branch. I then compared the output with commands like:
```
$ diff -U2 <(grep -P 'failed|Because' tmp/tex.old) \
           <(grep -P 'failed|Because' tmp/tex.new)
```

There were no changes in the total number of messages or KaTeX expressions where the parser failed. For a small number of expressions, there were changes in the set of soft failures reported before reaching a hard failure; the reasons for those are in a couple of individual commit messages. The full diff was (with a denominator of "28109 of them were KaTeX containing messages and 3370 of those failed"):
```
@@ -38,4 +38,6 @@
   Because of unsupported css class: frac-line:
     335 messages failed.
+  Because of unsupported css class: delimcenter:
+    184 messages failed.
   Because of unsupported css class: mtable:
     177 messages failed.
@@ -90,6 +92,4 @@
   Because of unsupported css class: cd-arrow-pad:
     8 messages failed.
-  Because of unsupported inline css property: border-right-width:
-    5 messages failed.
   Because of unsupported css class: boxpad:
     4 messages failed.
@@ -102,4 +102,6 @@
   Because of unsupported css class: rule:
     3 messages failed.
+  Because of unsupported inline css property: border-right-width:
+    3 messages failed.
   Because of unsupported inline css property: border-top-width:
     3 messages failed.
@@ -120,5 +122,5 @@
   Because of unsupported css class: textup:
     2 messages failed.
-  Because of unsupported inline css property: border-right-style:
+  Because of unsupported css class: vertical-separator:
     2 messages failed.
   Because of unsupported css class: mover:
```

(*) One change is still outstanding, in #1720. But it won't much interact with the main logic.

#### 8c4106d84 katex [nfc]: Show line numbers only on unknown hard-fails, not others

This makes the output of the survey script more stable as our KaTeX
parser gets refactored and otherwise edited.

#### e74557bba tools/content [nfc]: Tie-break on reason text when number of failures equal

This helps make the output more stable from run to run, so that
it's easier to spot changes (or confirm the absence of changes)
when editing the code.

#### 16deb86bb katex [nfc]: Cut stray import of widgets library

This was here only for a reference in a doc.  In general we try to
avoid imports of widgets from model code; it's an inversion of layers.

#### e223229e5 katex: Require height on pstrut spans

If this were missing, it's not clear to me that zero would be an
appropriate default.

In any case, in an empirical corpus, it's always present.
So just require that.

#### b62979c58 katex [nfc]: Separate _parseInlineStyles from constructing a KatexSpanStyles

Also document the existing _parseSpanInlineStyles method, and
describe our plan for eliminating most of its call sites.

#### a88de1450 katex [nfc]: Push more parsing into _parseInlineStyles; return Map

This has a small effect on the survey script's list of failure
reasons: in the rare case that the "unexpected shape of inline CSS"
error appears, it now fires before any of the CSS properties in the
same inline style are processed.  That can mean fewer entries added
to unsupportedInlineCssProperties.

This difference is only possible, though, on a KaTeX expression that
is going to reach the same hard failure either way.  So it has no
effect on behavior seen by a user.

#### ade7f6662 katex: Fix a misleading log line: unexpected CSS property, not value

Until the previous commit, this bit of code was handling both the case
where the value was unexpected (for which this message was accurate)
and the case where the property itself was unexpected (for which it
wasn't).  Now the first case is handled elsewhere, so fix the
remaining case.

#### 9c32bd6b6 katex [nfc]: Skip building whole KatexSpanStyles for struts' two properties

We know at this spot that there are just two specific CSS properties
we expect to see, and we'll end up handling them directly rather than
through a KatexSpanStyles object.  So parse them directly, rather
than build a whole KatexSpanStyles object (and then another one
with `.filter()`).

#### 6c2e55cb7 katex [nfc]: Cut vertical-align from generic style properties

All the remaining call sites of _parseSpanInlineStyles would throw
anyway if this property were actually found.  We only expect it
in a specific context, namely a strut.

#### f12c749f4 katex [nfc]: Check "only has height" directly, without making KatexSpanStyles

This lets us skip allocating these objects (two in each case -- the
second one comes from `.filter()`).  We also get to skip parsing the
value of `height`, since we don't intend to use it.

#### 7443e1ae6 katex [nfc]: Directly handle expected inline styles on vlist child

There's only a handful of specific properties we expect to see on this
type of span; so handle those explicitly.

In fact, making this list explicit brings to light that there's one
property here which doesn't actually appear on KaTeX's vlist children:
height.  We'll cut that in a separate non-NFC commit.

This will also open up ways to simplify the interaction between this
and the margin-handling logic below.

#### 8e059b3c6 katex: Don't expect height on vlist child spans

These spans are highly structured; the only properties that go
into their inline styles are top, margin-left, and margin-right.

#### e61c6c7c4 katex [nfc]: Consolidate logic for computing overall styles of KatexSpanNode

This just pulls these three pieces of closely-related logic next to
each other.  That will make it easier to refactor them further.

This causes one change in the survey script's list of failure reasons:
when the `delimcenter` class occurs with an inline `top` property, we
now record the unsupported class before reaching the hard fail for the
unsupported property.  This has no user-visible effect, though,
because it can only happen when the expression is going to reach that
hard failure either way.

#### 77b1c2fd4 katex [nfc]: Inline remaining/main use of _parseSpanInlineStyles

And inline the effect of the `merge` method, eliminating that
method too.

This way we get to construct just one KatexSpanStyles object, rather
than constructing three of them when inline styles are present.

#### 26ac991aa katex [nfc]: Split out _parseStrut, _parseVlist, _parseGenericSpan

Each of these swathes of logic has no interaction with the others.
Splitting them into their own methods makes that structure easy for
the reader to see.

### DIFF
--- a/lib/model/katex.dart
+++ b/lib/model/katex.dart
@@ -612,23 +612,32 @@ class _KatexParser {
           _hasError = true;
       }
     }
-    final classStyles = KatexSpanStyles(
+
+    final inlineStyles = _parseInlineStyles(element);
+    final styles = KatexSpanStyles(
       fontFamily: fontFamily,
       fontSizeEm: fontSizeEm,
       fontWeight: fontWeight,
       fontStyle: fontStyle,
       textAlign: textAlign,
+      heightEm: _takeStyleEm(inlineStyles, 'height'),
+      topEm: _takeStyleEm(inlineStyles, 'top'),
+      marginLeftEm: _takeStyleEm(inlineStyles, 'margin-left'),
+      marginRightEm: _takeStyleEm(inlineStyles, 'margin-right'),
+      // TODO handle more CSS properties
     );
-    final inlineStyles = _parseSpanInlineStyles(element);
-    if (inlineStyles != null) {
-      // Currently, we expect `top` to only be inside a vlist, and
-      // we handle that case separately above.
-      if (inlineStyles.topEm != null) {
-        throw _KatexHtmlParseError('unsupported inline CSS property: top');
+    if (inlineStyles != null && inlineStyles.isNotEmpty) {
+      for (final property in inlineStyles.keys) {
+        assert(debugLog('KaTeX: Unexpected inline CSS property: $property'));
+        unsupportedInlineCssProperties.add(property);
+        _hasError = true;
       }
     }
-    final styles = inlineStyles == null ? classStyles
-      : classStyles.merge(inlineStyles);
+    // Currently, we expect `top` to only be inside a vlist, and
+    // we handle that case separately above.
+    if (styles.topEm != null) {
+      throw _KatexHtmlParseError('unsupported inline CSS property: top');
+    }
 
     String? text;
     List<KatexNode>? spans;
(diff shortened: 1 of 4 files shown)


lib/model/katex.dart
```diff
@@ -749,12 +725,27 @@ class _KatexParser {
...
+    if (expression == null) return null;
     if (expression is css_visitor.EmTerm && expression.value is num) {
       return (expression.value as num).toDouble();
     }
+    assert(debugLog('KaTeX: Unsupported value for CSS property $property,'
+      ' expected a length in em: ${expression.toDebugString()}'));
+    unsupportedInlineCssProperties.add(property);
+    _hasError = true;
     return null;
   }
```
rajveermalviya, line 724: This is really neat!

gnprice:

Issue #1636, opened 2025-06-26
Refine `Toggle` dimensions to follow Figma spec


chrisbobbe:
We plan to ship "Invisible mode" with a toggle switch that doesn't quite match the Figma. Quoting the PR: 

https://github.com/zulip/zulip-flutter/pull/1631
> Some known differences from the [Figma](https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6070-60682&m=dev), which won't be hard to fix but require upstream changes, so not totally trivial:
> 
> - When the switch is on:
>   - The switch should be 28px by 48px, not 36px by 56px.
>   - The "thumb" (white circle with checkmark) should have radius 10px, not 12px.
> 
> - When the switch is off:
>   - The switch should be 26px by 46px, not 34px by 54px.
>   - The "thumb" should have radius 7px, not 8px.

This issue is to resolve those differences.

### Implementation

The upstream issue is https://github.com/flutter/flutter/issues/131478 ; we should investigate and send a PR, then use the exposed configuration to resolve the TODO in the `Toggle` code.

PR #1723, opened 2025-07-19
Fix/toggle dimensions 1636

loveucifer:
Fixes: #1636

This pull request resolves the issue where the "Invisible mode" toggle switch did not match the Figma design specifications. The standard Flutter Switch widget does not provide the necessary customization options to achieve the required dimensions.
to address this, this PR introduces a new custom widget, FigmaSwitch, which is built from the ground up to precisely match the specified dimensions for both its on and off states.

Changes I made""

New FigmaSwitch Widget:  highly-customizable w, highly-customizable switch widget that accurately reflects the Figma design for the "Invisible mode" toggle.

New FigmaSwitchListTile Widget: A ListTile that incorporates the FigmaSwitch for easy integration into settings pages.

Settings Page Integration: The "Invisible mode" setting now uses the new FigmaSwitchListTile to ensure it matches the design.

Testing""

Navigate to the settings page.

See the "Invisible mode" toggle switch.

Verify that its dimensions match the Figma specifications in both the "on" and "off" states.

On: 28px by 48px, with a 10px radius thumb.

Off: 26px by 46px, with a 7px radius thumb.

Confirm that the toggle switch functions correctly, updating the user's presence setting.             

Hope this resolves the issue 

also attaching before and after screenshots 

BEFORE
BEFORE

AFTER
AFTER

### DIFF
--- a/lib/widgets/settings.dart
+++ b/lib/widgets/settings.dart
@@ -34,7 +34,7 @@ class FigmaToggle extends StatelessWidget {
     final theme = Theme.of(context);
     final colorScheme = theme.colorScheme;
 
-    // Figma-specified dimensions
+    // Exact Figma-specified dimensions
     final trackWidth = value ? 48.0 : 46.0;
     final trackHeight = value ? 28.0 : 26.0;
     final thumbRadius = value ? 10.0 : 7.0;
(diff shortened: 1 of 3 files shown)


loveucifer:
uhh I think I messed up with the test file

chrisbobbe:
Thanks for the PR!

(I fixed your PR description to point to #1636; it was pointing to #1631 by mistake.)

See the issue description:

> The upstream issue is [flutter/flutter#131478](https://github.com/flutter/flutter/issues/131478) ; we should investigate and send a PR, then use the exposed configuration to resolve the TODO in the `Toggle` code.

Our plan to fix this upstream will help other projects besides Zulip. This PR doesn't do that 🙂 it just reimplements a toggle switch with different dimensions. I suspect it also drops behavior that the toggle switch currently has, like usability with a screen reader via the [Semantics](https://api.flutter.dev/flutter/widgets/Semantics-class.html) API.

Since the issue has a clear path forward and this PR isn't on that path, I'll go ahead and close this PR. In future please [ask](https://github.com/zulip/zulip-flutter?tab=readme-ov-file#asking-questions-getting-help) if you have doubts about how we'd like things to be done; it can save you some time.

PR #1725, opened 2025-07-20
Visually disable Save button on validation error.

loveucifer:
**Fixes: #1481**

This change implements the disabled visual state for the "Save" button in the edit-message compose box, providing immediate feedback to the user when the content is invalid.
also
**The Problem:**

Previously, when editing a message, if the user entered content that was too long (exceeding the server's character limit), the "Save" button remained visually active. Although tapping it would correctly show a validation error dialog and prevent saving, the button itself did not appear disabled. This could lead to a confusing user experience, as there was no visual cue that the save action was invalid.

**The Solution:**

This PR resolves the issue by wrapping the "Save" button (`ZulipWebUiKitButton`) inside a `ValueListenableBuilder`.

- The `ValueListenableBuilder` listens to the `hasValidationErrors` notifier on the `ComposeContentController`.
- When `hasValidationErrors` is `true`, or when the original message content is still loading, the button is now visually disabled in two ways:
    1. It is wrapped in an `Opacity` widget to give it a faded appearance.
    2. The `onPressed` callback is set to an empty function `() {}`, preventing any action on tap and satisfying the non-nullable requirement of the button's API.

This ensures the UI reacts instantly to the validation state of the text input, providing clear and correct feedback to the user.

**Testing:**

I have manually tested this change with the following scenarios:

1.  **Edit a message and enter content that exceeds the 10,000 character limit.**
    - **Result:** The "Save" button correctly fades out and becomes unresponsive.
2.  **Delete characters to bring the content back within the valid limit.**
    - **Result:** The "Save" button correctly becomes fully opaque and active again.
3.  **Edit a message and delete all content.**
    - **Result:** The "Save" button correctly remains enabled, preserving the ability to delete a message's content by saving an empty edit.

The fix works as expected and resolves the visual inconsistency.        
I hope this solves the error

### DIFF
--- a/lib/widgets/compose_box.dart
+++ b/lib/widgets/compose_box.dart
@@ -1799,14 +1799,24 @@ class _EditMessageBanner extends _Banner {
   @override
   Widget buildTrailing(context) {
     final zulipLocalizations = ZulipLocalizations.of(context);
+    final controller = composeBoxState.controller as EditMessageComposeBoxController;
     return Row(mainAxisSize: MainAxisSize.min, spacing: 8, children: [
       ZulipWebUiKitButton(label: zulipLocalizations.composeBoxBannerButtonCancel,
         onPressed: composeBoxState.endEditInteraction),
-      // TODO(#1481) disabled appearance when there are validation errors
-      //   or the original raw content hasn't loaded yet
-      ZulipWebUiKitButton(label: zulipLocalizations.composeBoxBannerButtonSave,
-        attention: ZulipWebUiKitButtonAttention.high,
-        onPressed: () => _handleTapSave(context)),
+      ValueListenableBuilder<bool>(
+        valueListenable: controller.content.hasValidationErrors,
+        builder: (context, hasErrors, child) {
+          final bool disabled = hasErrors || controller.originalRawContent == null;
+          return Opacity(
+            opacity: disabled ? 0.5 : 1.0,
+            child: ZulipWebUiKitButton(
+              label: zulipLocalizations.composeBoxBannerButtonSave,
+              attention: ZulipWebUiKitButtonAttention.high,
+              onPressed: disabled ? () {} : () => _handleTapSave(context),
+            ),
+          );
+        },
+      ),
     ]);
   }
 }


chrisbobbe:
Thanks for the PR! I'm closing this because the issue isn't yet ripe to work on. Quoting the top of the issue description:

> When we first implement the edit-message compose box ([#126](https://github.com/zulip/zulip-flutter/issues/126)), we plan to skip a feature that the regular compose box has, because we don't have a Figma design for it.

When we get a Figma design from our designer, we'll put a link in the issue, and then the task will be to follow that design.

gnprice:
BTW, one thing I notice is that your description for this PR has a lot of text that doesn't say very much. (Most of it is just repeating what's in either the issue or the code.) It looks a lot like the sort of text you might get by asking an LLM like ChatGPT to write it.

Generally it's not helpful to include text you had an LLM write for you. It just makes more text a reviewer needs to read through if they're going to review your work, without any benefit; the LLM doesn't know anything about this code beyond what was in its input. I'd rather read your own actual thoughts about what you were doing.

loveucifer:
> BTW, one thing I notice is that your description for this PR has a lot of text that doesn't say very much. (Most of it is just repeating what's in either the issue or the code.) It looks a lot like the sort of text you might get by asking an LLM like ChatGPT to write it.
> 
> Generally it's not helpful to include text you had an LLM write for you. It just makes more text a reviewer needs to read through if they're going to review your work, without any benefit; the LLM doesn't know anything about this code beyond what was in its input. I'd rather read your own actual thoughts about what you were doing.

ohnokay I have a lot of typos while writing so I send to llm to rephrase or something , totally understand will not make this mistake again

gnprice:
Yeah, typos are fine. :slightly_smiling_face: We're much more interested in the content of what you're saying.

(We're pickier about typos when it comes to the parts that get committed into the repo. But in PR descriptions, GitHub comments, chat messages, etc., spelling and grammar and so on are no problem as long as we can understand what you mean.)

loveucifer:
sure, will totally try my best

PR #1729, opened 2025-07-22
Split out KaTeX tests; simplify a bit; split KaTeX widgets from other content widgets

gnprice:
This is the next round, after #1722, of the refactors I'd been contemplating for the KaTeX code but hadn't wanted to do before the logic we'd had in recent releases was merged. In this branch:
* Split model/content_test.dart to match the implementation code; make a KatexExample subclass of ContentExample, with conveniences for KaTeX examples specifically.
* Move KaTeX widgets to a new widgets/katex.dart, from widgets/content.dart. (Move their tests to a new file correspondingly.)
* Simplify writing test data in a few different ways.
* Add a bit of docs on each of the KaTeX-related content-node classes — particularly to help clarify the relationship between types like MathNode, KatexNode, and KatexSpanNode, which can sound similar on the face of things.

#### 20eb9af7e katex test [nfc]: Split parsing tests to their own file, like implementation

This way we maintain our usual parallel between the files that tests
live in and the files with the code the tests are about.

There's more we can do to make use of having a specific class for
these.  But we'll do that in separate commits, so that this massive
move commit is as pure of a move as possible.

#### 286fede3a katex [nfc]: Move KaTeX widgets to their own file

This is a pretty self-contained swath of our content widgets -- the
only interface the others use from these is KatexWidget.  So it works
well to split into a separate file, paralleling the model files.

#### f14e318d3 katex test [nfc]: Simplify a bit with a KatexExample.block constructor

This removes some repetitive noisy bits from these examples,
as well as a level of indentation.

#### 18112661a katex test [nfc]: Use Dart raw strings to avoid double-backslash

This is available to us now that these strings don't include the
leading "```math" and trailing "```" lines, and the newlines
separating them from the main content.

#### afb3692a6 katex [nfc]: On KatexSpanNode constructor let text or nodes be omitted

We already have an assert enforcing that exactly one of these is
passed with a non-null (i.e. non-default) value.  So there's no
risk of forgetting to specify a detail that should be specified.

#### af80c5258 katex [nfc]: Let KatexSpanNode.styles default to empty

We have a lot of calls to the KatexSpanNode constructor in our tests,
and a large fraction of them pass an empty `styles`.  Let that be the
default, reducing some noise.  It's a natural default, since it
effectively means "do nothing".


gnprice:
@rajveermalviya the code moves in this PR will call for some conflict resolution in order to rebase your open KaTeX PRs atop this one. I'd like those PRs to also match the style changes here, even where Git doesn't flag a merge conflict.

The good news on that front is that I think this completes all the changes I had in mind that were likely to have merge conflicts with lots of other KaTeX work. :slightly_smiling_face: I still have other refactors in mind, but they affect narrower swaths of code and I think mostly won't call for any changes to existing open PRs.

Issue #891, opened 2024-08-17
msglist: Refresh msglist recipient-header date text periodically to respond to current-date change


chrisbobbe:
Untested, but from reading code, it seems like if it's near midnight and you're looking at a recipient header that says "TODAY", it won't update to "YESTERDAY" after midnight unless something unrelated to the clock causes it to update. We can fix this by running a check every minute or so, to refresh the current-time value that controls that text.

(Same for the transition from "YESTERDAY" to an absolute date.)

gnprice:
Yeah, agreed.

This is a bug that web has even today in a much more conspicuous place, namely "Recent conversations" — I regularly get fooled by a conversation being shown there as "Just now", when in reality it was like 20 minutes ago and that just happens to be the last time anything exogenously caused the table to get updated. So I agree this is a post-launch issue.

Issue #293, opened 2023-08-31
profile: Keep "current" local time current


gnprice:
This will be a followup to #292, showing a user's current local time on their profile page.

It'd be good for that "current" local time to remain current as time advances. That way, if our user sets down the app and does something else, and then comes back some minutes or hours later, we're not showing them something that's untrue. (That's a bug that zulip-mobile currently has here, and that Zulip web has for the times in the "Recent conversations" screen.)

This is probably not real complicated; we just need the relevant widget to rebuild at the start of every minute.

gnprice:
Thanks @Komyyy for picking this up!

This issue has the M7 Future milestone, which means we're not going to have time to help move it forward within the next couple of months while we [focus on getting the app launched](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20beta.20timeline/near/2078556). If you're interested in contributing to Zulip, I'd be glad for you to take a look through issues in the M5-a and M5 launch milestones and take on one of those.

Komyyy:
Thanks, @gnprice 
After I opened this draft PR, I will contribute to M5 issues.

Vedanshi27vishu:
Hello Sir, Please assign me this issue.

rajveermalviya:
@Vedanshi27vishu, Please stop spamming on multiple issue threads.

Read the [contributing guidelines](https://github.com/zulip/zulip-flutter?tab=readme-ov-file#contributing) in the README, especially the section of ["picking an issue to work on"](https://github.com/zulip/zulip-flutter?tab=readme-ov-file#picking-an-issue-to-work-on).

PR #1730, opened 2025-07-22
Follow user's time format setting (12 or 24 hour)

### LINKED ISSUE #1015: ui: Follow `user_settings.twenty_four_hour_time`
When we show timestamps, for example on messages in the message list, we should format them according to the user's setting:

- `user_settings.twenty_four_hour_time` true: "13:52"
- `user_settings.twenty_four_hour_time` false: "1:52 PM"

This is something we did in the zulip-mobile legacy app.

In addition to the timestamps on messages in the message list (which is the most visible and most important place), this should apply to everywhere else that we show a time of day to the user. To find other such places, try `git grep DateFormat`.

### References

- [Help center doc](https://zulip.com/help/change-the-time-format)
- `UserSettings.twentyFourHourTime` in lib/api/model/initial_snapshot.dart
- `UserSettingsUpdateEvent` in lib/api/model/events.dart

chrisbobbe:
Fixes #1015.

cc @alya for screenshots below.

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -175,6 +175,7 @@ class UserSettingsUpdateEvent extends Event {
     final value = json['value'];
     switch (UserSettingName.fromRawString(json['property'] as String)) {
       case UserSettingName.twentyFourHourTime:
+        return TwentyFourHourTimeMode.fromApiValue(value as bool?);
       case UserSettingName.displayEmojiReactionUsers:
         return value as bool;
       case UserSettingName.emojiset:
--- a/lib/api/model/initial_snapshot.dart
+++ b/lib/api/model/initial_snapshot.dart
@@ -259,7 +259,12 @@ class RecentDmConversation {
 /// in <https://zulip.com/api/register-queue>.
 @JsonSerializable(fieldRename: FieldRename.snake, createFieldMap: true)
 class UserSettings {
-  bool twentyFourHourTime;
+  @JsonKey(
+    fromJson: TwentyFourHourTimeMode.fromApiValue,
+    toJson: TwentyFourHourTimeMode.staticToJson,
+  )
+  TwentyFourHourTimeMode twentyFourHourTime;
+
   bool? displayEmojiReactionUsers; // TODO(server-6)
   Emojiset emojiset;
   bool presenceEnabled;
--- a/lib/api/model/model.dart
+++ b/lib/api/model/model.dart
@@ -298,6 +298,35 @@ enum UserSettingName {
   String toJson() => _$UserSettingNameEnumMap[this]!;
 }
 
+/// A value from [UserSettings.twentyFourHourTime].
+enum TwentyFourHourTimeMode {
+  twelveHour(apiValue: false),
+  twentyFourHour(apiValue: true),
+
+  /// The locale's default format (12-hour for en_US, 24-hour for fr_FR, etc.).
+  // TODO(#1727) actually follow this
+  // Not sent by current servers, but planned when most client installs accept it:
+  //   https://chat.zulip.org/#narrow/channel/378-api-design/topic/.60user_settings.2Etwenty_four_hour_time.60/near/2220696
+  // TODO(server-future) Write down what server N starts sending null;
+  //   adjust the comment; leave a TODO(server-N) to delete the comment
+  localeDefault(apiValue: null),
+  ;
+
+  const TwentyFourHourTimeMode({required this.apiValue});
+
+  final bool? apiValue;
+
+  static bool? staticToJson(TwentyFourHourTimeMode instance) => instance.apiValue;
+
+  bool? toJson() => TwentyFourHourTimeMode.staticToJson(this);
+
+  static TwentyFourHourTimeMode fromApiValue(bool? value) => switch (value) {
+    false => twelveHour,
+    true => twentyFourHour,
+    null => localeDefault,
+  };
+}
+
 /// As in [UserSettings.emojiset].
 @JsonEnum(fieldRename: FieldRename.kebab, alwaysCreate: true)
 enum Emojiset {
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -2011,7 +2011,7 @@ enum MessageTimestampStyle {
       return DateFormat.yMMMd().format(dateTime);
     }
   }
-  static final _timeOnlyFormat = DateFormat('h:mm aa', 'en_US');
+  static final _timeOnlyFormat = DateFormat('h:mm aa');
   static final _lightboxFormat = DateFormat.yMMMd().add_Hms();
   static final _fullFormat = DateFormat.yMMMd().add_jm();
 
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -1915,12 +1915,14 @@ class SenderRow extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final zulipLocalizations = ZulipLocalizations.of(context);
     final store = PerAccountStoreWidget.of(context);
     final messageListTheme = MessageListTheme.of(context);
     final designVariables = DesignVariables.of(context);
 
     final sender = store.getUser(message.senderId);
-    final timestamp = timestampStyle.format(message.timestamp);
+    final timestamp = timestampStyle.format(
+      message.timestamp, now: DateTime.now(), zulipLocalizations: zulipLocalizations);
 
     final showAsMuted = _showAsMuted(context, store);
 
@@ -2010,7 +2012,11 @@ enum MessageTimestampStyle {
 
   /// Format a [Message.timestamp] for this mode.
   // TODO(i18n): locale-specific formatting (see #45 for a plan with ffi)
-  String? format(int messageTimestamp) {
+  String? format(
+    int messageTimestamp, {
+    required DateTime now,
+    required ZulipLocalizations zulipLocalizations,
+  }) {
     final asDateTime =
       DateTime.fromMillisecondsSinceEpoch(1000 * messageTimestamp);
 
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -2011,7 +2011,7 @@ enum MessageTimestampStyle {
       return DateFormat.yMMMd().format(dateTime);
     }
   }
-  static final _timeOnlyFormat = DateFormat('h:mm aa');
+  static final _timeFormat = DateFormat('h:mm aa');
   static final _lightboxFormat = DateFormat.yMMMd().add_Hms();
   static final _fullFormat = DateFormat.yMMMd().add_jm();
 
@@ -2031,7 +2031,7 @@ enum MessageTimestampStyle {
       case dateOnlyRelative:
         return _formatDateOnlyRelative(asDateTime,
           now: now, zulipLocalizations: zulipLocalizations);
-      case timeOnly: return _timeOnlyFormat.format(asDateTime);
+      case timeOnly: return _timeFormat.format(asDateTime);
       case lightbox: return _lightboxFormat.format(asDateTime);
       case full: return _fullFormat.format(asDateTime);
     }
--- a/lib/widgets/message_list.dart
+++ b/lib/widgets/message_list.dart
@@ -2011,9 +2011,30 @@ enum MessageTimestampStyle {
       return DateFormat.yMMMd().format(dateTime);
     }
   }
-  static final _timeFormat = DateFormat('h:mm aa');
-  static final _timeFormatWithSeconds = DateFormat('Hms');
-  static final _lightboxFormat = DateFormat.yMMMd().addPattern(_timeFormatWithSeconds.pattern);
+
+  static final _timeFormat12 =                       DateFormat('h:mm aa');
+  static final _timeFormat24 =                       DateFormat('Hm');
+  static final _timeFormatLocaleDefault =            DateFormat('jm');
+  static final _timeFormat12WithSeconds =            DateFormat('h:mm:ss aa');
+  static final _timeFormat24WithSeconds =            DateFormat('Hms');
+  static final _timeFormatLocaleDefaultWithSeconds = DateFormat('jms');
+
+  // ignore: unused_element
+  static DateFormat _resolveTimeFormat(TwentyFourHourTimeMode mode) => switch (mode) {
+    TwentyFourHourTimeMode.twelveHour => _timeFormat12,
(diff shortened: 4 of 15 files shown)


chrisbobbe:
Conveniently for making screenshots, I could fit three things on the screen at once:
- A [global time](https://zulip.com/help/global-times#global-times) in message content
- A short message timestamp in the message list
- A long message timestamp in the message action sheet

## `user_settings.twenty_four_hour_time: false`

| Before | After |
| --- | --- |
| image | image |
| image | image |

## `user_settings.twenty_four_hour_time: true`

| Before | After |
| --- | --- |
| image | image |
| image | image |

chrisbobbe:
Thanks for the review!

lib/widgets/message_list.dart
```diff
@@ -1826,8 +1827,14 @@ class DateText extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final store = PerAccountStoreWidget.of(context);
     final messageListTheme = MessageListTheme.of(context);
     final zulipLocalizations = ZulipLocalizations.of(context);
+    final formattedTimestamp = MessageTimestampStyle.dateOnlyRelative.format(
+      timestamp,
+      now: ZulipBinding.instance.utcNow().toLocal(),
```
gnprice, line 1849: Using the binding is a good idea. How about having the callee invoke it? Then it doesn't need to be a parameter that each caller passes.

gnprice: (Potentially that happens after the other commits — it might be simpler to do that change after this one:
26d5ef54b msglist [nfc]: Finish centralizing on MessageTimestampStyle

than vice versa.)

test/widgets/message_list_test.dart
```diff
@@ -1624,27 +1625,132 @@ void main() {
...
-      ("2023-01-10 12:00", zulipLocalizations.today),
-      ("2023-01-10 00:00", zulipLocalizations.today),
-      ("2023-01-10 23:59", zulipLocalizations.today),
-      ("2023-01-09 23:59", zulipLocalizations.yesterday),
-      ("2023-01-09 00:00", zulipLocalizations.yesterday),
-      ("2023-01-08 00:00", "Jan 8"),
-      ("2022-12-31 00:00", "Dec 31, 2022"),
-      // Future times
-      ("2023-01-10 19:00", zulipLocalizations.today),
-      ("2023-01-11 00:00", "Jan 11, 2023"),
```
gnprice, line 1653: For these test cases of the "relative date" logic, it's quite useful that the different times are all right next to each other and next to the value of "now". When you look at "now" and then read through the different cases, they tell a legible story about how the intended logic works and what edge cases are successfully covered.

test/widgets/message_list_test.dart
```diff
@@ -1624,27 +1625,132 @@ void main() {
...
+    });
+    doTests("2023-01-10 23:59", (style, expectTwelveHour) => switch (style) {
+      MessageTimestampStyle.none => null,
+      MessageTimestampStyle.dateOnlyRelative => zulipLocalizations.today,
+      MessageTimestampStyle.timeOnly =>
+        expectTwelveHour ? '11:59 PM' : '23:59',
+      MessageTimestampStyle.lightbox =>
+        expectTwelveHour ? 'Jan 10, 2023 11:59:00 PM' : 'Jan 10, 2023 23:59:00',
+      MessageTimestampStyle.full =>
+        expectTwelveHour ? 'Jan 10, 2023 11:59 PM' : 'Jan 10, 2023 23:59',
```
gnprice: Meanwhile, all these non-relative formats don't really benefit from most of the different test values that are there for the relative-date format. After seeing how `timeOnly` comes out on one timestamp that's at 11:59pm, there's nothing new learned from seeing another such timestamp.

test/widgets/message_list_test.dart
```diff
@@ -1624,27 +1625,132 @@ void main() {
...
+            // (This will change when we do #1727.)
+            TwentyFourHourTimeMode.localeDefault => true,
+          });
+          final description = '${style.name}, ${mode.name}: $messageTimestampStr returns $expected';
+          test(description, () {
+            addTearDown(testBinding.reset);
+
+            withClock(Clock.fixed(DateTime.parse("2023-01-10 12:00")), () {
+              // TODO take the message timestamp as an int, like [Message.timestamp]?
+              final timestamp = DateTime.parse(messageTimestampStr).millisecondsSinceEpoch ~/ 1000;
```
gnprice: No, it's useful that the test source code has it as a readable date/time string. That's what makes it possible for a human reading the tests to judge whether a given expected answer is right, and also to think about what edge cases there might be and then look to see if they're covered by the given examples.

lib/api/model/initial_snapshot.dart
```diff
@@ -259,7 +259,13 @@ class RecentDmConversation {
 /// in <https://zulip.com/api/register-queue>.
 @JsonSerializable(fieldRename: FieldRename.snake, createFieldMap: true)
 class UserSettings {
-  bool twentyFourHourTime;
+  @JsonKey(
+    readValue: JsonNullable.readFromJsonAssertingPresent<bool>,
+    fromJson: TwentyFourHourTimeMode.fromApiValue,
+    toJson: TwentyFourHourTimeMode.staticToJson,
+  )
+  TwentyFourHourTimeMode twentyFourHourTime;
```
gnprice: How about without the readValue?

That'd be simpler in the code. I believe the only difference in behavior is that if a server didn't send the value at all, this would treat it the same as null, whereas the PR's current version would treat it as malformed.

I don't think we gain anything by rejecting a case where the server doesn't send the value. In general across our API binding we treat that the same as sending null; the only exceptions are in a small number of places where the API's semantics actually give different meanings to absent and null. (In fact it looks like just one so far: `deliveryEmail` on `RealmUserUpdateEvent`.) But it's not doing that here; if it were, "asserting present" would be wrong.

lib/api/model/events.dart
```diff
@@ -175,6 +175,8 @@ class UserSettingsUpdateEvent extends Event {
     final value = json['value'];
     switch (UserSettingName.fromRawString(json['property'] as String)) {
       case UserSettingName.twentyFourHourTime:
+        final apiValue = JsonNullable.readFromJsonAssertingPresent<bool>(json, key);
+        return TwentyFourHourTimeMode.fromApiValue(apiValue);
```
gnprice: Similarly:
```suggestion
        return TwentyFourHourTimeMode.fromApiValue(value as bool?);
```

lib/api/model/initial_snapshot.dart
gnprice: (And then can we also eliminate the other JsonKey arguments? What if we rename or alias `TwentyFourHourTimeMode.fromApiValue` as `fromJson`?)

lib/api/model/model.dart
```diff
@@ -298,6 +298,35 @@ enum UserSettingName {
   String toJson() => _$UserSettingNameEnumMap[this]!;
 }
 
+/// A value from the bool [UserSettings.twentyFourHourTime].
+enum TwentyFourHourTimeMode {
```
gnprice: This doc seems confusing — if it's a bool, why is it an enum? :upside_down_face: 

I think s/the bool // would be fine.

lib/widgets/store.dart
```diff
@@ -213,6 +215,21 @@ class PerAccountStoreWidget extends StatefulWidget {
...
+  /// [PerAccountStoreWidget] ancestor.
+  ///
+  /// Useful in particular for content widgets that have special functionality
+  /// when they appear in a per-account context, which is the common case
+  /// (the message list).
+  ///
+  /// This will be useful for #488, supporting Zulip content rendering
+  /// in e.g. an org description shown before you log in.
+  // TODO(#488) Remove #488 comment.
+  static PerAccountStore? maybeOf(BuildContext context) {
```
gnprice: This will be useful when we do that, but I think it's not helpful to start half-doing it in advance.

There are already several content widgets that use PerAccountStore.of; when we start supporting Zulip content rendering in such contexts, we'll want to sweep through all of those and work out what they should each do instead. So GlobalTime can do the same until then.

test/widgets/content_test.dart
```diff
@@ -139,9 +140,16 @@ void main() {
   Future<void> prepareContent(WidgetTester tester, Widget child, {
     List<NavigatorObserver> navObservers = const [],
     bool wrapWithPerAccountStoreWidget = false,
+    TwentyFourHourTimeMode? twentyFourHourTimeMode
```
gnprice: nit:
```suggestion
    TwentyFourHourTimeMode? twentyFourHourTimeMode,
```

test/widgets/content_test.dart
```diff
@@ -139,9 +140,16 @@ void main() {
...
   }) async {
     if (wrapWithPerAccountStoreWidget) {
-      await testBinding.globalStore.add(eg.selfAccount, eg.initialSnapshot());
+      final initialSnapshot = eg.initialSnapshot();
+      if (twentyFourHourTimeMode != null) {
+        initialSnapshot.userSettings.twentyFourHourTime = twentyFourHourTimeMode;
+      }
+      await testBinding.globalStore.add(eg.selfAccount, initialSnapshot);
```
gnprice: How about taking an optional `initialSnapshot` parameter instead?

That way it's up to the caller to deal with whatever details it wants to add, and this shared prep function doesn't need to grow for each feature that only a small number of test cases will interact with.

test/widgets/content_test.dart
```diff
@@ -1102,11 +1110,45 @@ void main() {
...
 
     testWidgets('smoke', (tester) async {
-      await prepareContent(tester, plainContent('<p>$timeSpanHtml</p>'));
+      await prepareContent(tester,
+        wrapWithPerAccountStoreWidget: false,
+        plainContent('<p>$timeSpanHtml</p>'));
       tester.widget(find.textContaining(renderedTextRegexp));
+
+      // TODO(#1727) test with different locales
+      tester.widget(find.textContaining(renderedTextRegexpTwelveHour));
```
gnprice:
```suggestion
      // TODO(#1727) test with different locales
      tester.widget(find.textContaining(renderedTextRegexpTwelveHour));
```

This is equivalent, right?

test/widgets/content_test.dart
```diff
@@ -1102,11 +1110,45 @@ void main() {
...
+
+    testWidgets('TwentyFourHourTimeMode.localeDefault', (tester) async {
+      await prepareContent(tester,
+        wrapWithPerAccountStoreWidget: true,
+        twentyFourHourTimeMode: TwentyFourHourTimeMode.localeDefault,
+        plainContent('<p>$timeSpanHtml</p>'));
+      tester.widget(find.textContaining(renderedTextRegexp));
+
+      // TODO(#1727) test with different locales
+      check(find.textContaining(renderedTextRegexpTwelveHour)).findsOne();
```
gnprice: It'd be good to have a short comment saying why this expectation is right — it looks counterintuitive.

test/widgets/content_test.dart
```diff
@@ -1102,11 +1110,45 @@ void main() {
...
     // `test/widgets/message_list_test.dart`.
-    final renderedTextRegexp = RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d [AP]M$');
+    final renderedTextRegexp = RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d(?: [AP]M)?$');
+    final renderedTextRegexpTwelveHour = RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d [AP]M$');
+    final renderedTextRegexpTwentyFourHour = RegExp(r'^(Tue, Jan 30|Wed, Jan 31), 2024, \d+:\d\d$');
 
     testWidgets('smoke', (tester) async {
-      await prepareContent(tester, plainContent('<p>$timeSpanHtml</p>'));
+      await prepareContent(tester,
+        wrapWithPerAccountStoreWidget: false,
```
gnprice: This test case is now doing something very unusual (indeed something that doesn't yet happen in the live app) — so it should definitely get a more specific name than this :-)

lib/widgets/message_list.dart
chrisbobbe: I was thinking we'd want `format` to be a pure function, so not a right place to check the clock. And I guess that callers would be more likely to take care of refreshing relative times (e.g. #293, #891) if they're responsible for reading the clock. What do you think?

lib/api/model/initial_snapshot.dart
chrisbobbe: > (And then can we also eliminate the other JsonKey arguments? What if we rename or alias `TwentyFourHourTimeMode.fromApiValue` as `fromJson`?)

I think when the .g.dart files are generated, an enum's `fromJson` or `toJson` methods (when present) are never actually used, unless you ask for it with those JsonKey arguments.

I'd normally just do `@JsonEnum(valueField: "apiValue")`, but when I do that, `dart run build_runner build` gives an error:

```
Failed to build with build_runner in 18s; wrote 381 outputs.                          
                                                                                      
log output for json_serializable on lib/api/model/initial_snapshot.dart               
E `JsonEnum.valueField` was set to "apiValue", but that field does not have a type of 
  String, int, or null.                                                               
  package:zulip/api/model/model.dart:303:6                                            
      ╷                                                                               
  303 │ enum TwentyFourHourTimeMode {                                                 
      │      ^^^^^^^^^^^^^^^^^^^^^^                                                   
      ╵                 
```

(The `apiValue` field is typed `bool?`.)

lib/widgets/message_list.dart
gnprice: Hmm, refreshing relative times is an interesting point. I guess let's leave it this way for now, and we'll see what we come up with to organize the code when we go to handle that.

lib/api/model/initial_snapshot.dart
gnprice: Hmm, I see. OK, then.

chrisbobbe:

The last commit isn't needed for this feature—

5152d41f1 json: Implement JsonNullable.toJson and readFromJsonAssertingPresent

—but I left it in so we have the option to merge it in case it might be helpful in the future.

gnprice:
Thanks for the revision!

Let's skip that last commit — I think for the same reasons as at https://github.com/zulip/zulip-flutter/pull/1730#discussion_r2223794172 , "readFromJsonAssertingPresent" doesn't really make sense for the situations that JsonNullable is for.

Other than that, all looks good to me; please go ahead and merge.

chrisbobbe:
Mm, `flutter analyze` is failing in CI—sent #1739 for that.

chrisbobbe:
Thanks for the review and merge!

Issue #408, opened 2023-11-21
Show notifications when in foreground too, on iOS


gnprice:
In my draft implementation of #321, we show notifications only if they arrive while the app is either in the background or not running; we don't show them if they arrive while the app is in the foreground.

This is the default behavior iOS provides, which is why that's what we get in that draft. But it doesn't seem like the right behavior for Zulip, for reasons described at https://github.com/zulip/zulip-mobile/issues/3114#issuecomment-585966226 . So we should arrange to have notifications be shown while the app is in the foreground, too.

This is done by implementing a certain method on our iOS app delegate, the class found at `ios/Runner/AppDelegate.swift`. See zulip-mobile's implementation at https://github.com/zulip/zulip-mobile/commit/d4606ab26c650fae661dd52cfe41f4691ae5d374 .

In this case Flutter provides its own implementation of that method, with the `FlutterAppDelegate` class that our `AppDelegate` inherits from. That implementation in turn has a protocol for Flutter plugins to provide implementations to be called. I haven't yet looked much into this, and I'm not sure if it'd be better to (a) use that protocol, providing an implementation to be called by that Flutter-provided mechanism, or (b) override that Flutter-provided implementation with our own.

loveucifer:
was looking for issues to fix and came across this and I did some search about this and came up with this not sure if im on the right but here goes , overriding flutter provided is recommended most places right or have you guys already made a choice on whatmethod to choose --https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:willpresent:withcompletionhandler:).  this and also firebase --- https://firebase.google.com/docs/cloud-messaging/flutter/receive?authuser=1  talks about using the AppDelegate.swift  method right if that's the case then we only need to fix something in that file and not add an extra dependency , idk if all these are correct I just looked all these up today

loveucifer:
also for some reason we can't see notifications on simulator devices I am trying to overcome this pls suggest any alt if anyone is aware

loveucifer:
and for some reason even if I try with my iOS device since I am on personal team it won't allow me to do this -- 

<img width="595" height="398" alt="Image" src="https://github.com/user-attachments/assets/f0c5ec6b-62cf-473e-ae1e-8ec6cd7956c7" />.  I will still send a pr since I think this will fix it but please do note that I haven't done testing on the same since there is no way to figure this out

gnprice:
[...]

PR #1734, opened 2025-07-22
hopefully fixes ios foreground notifications issue #408

loveucifer:
as previously mentioned in the comment of this issue #408  I read up on some community notes and decided to go with this method Idk if you guys were having trouble deciding on what to use , and I couldn't test this out since I didn't have a "team apple dev acc" but from what I read and understood from the docs this might be able to fix it 
here what I did
first of all i went with thist method because it's the official and most direct way Apple provides to solve this issue I think and I got most of the code from developer documentation for their UserNotifications framework. not sure if it works since as I said im not sure about how to test this but I have worked on something like this and I remember doing something exactly similar that's why I picked this issue , hope this fixes the issue, simulator showed no errors.

### DIFF
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -29,13 +29,13 @@ import Flutter
     UNUserNotificationCenter.current().delegate = self
 
     // This is needed to get a push token from APNs.
-    // The user will be prompted for permission only the first time.
+    // The user will be prompted for permission only the first timee.
     application.registerForRemoteNotifications()
 
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
   }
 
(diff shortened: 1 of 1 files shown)


loveucifer:
just noticed I added extra e on the time lol my bad was trying to click the hot reload button

gnprice:
Thanks. This is an issue, though, where an essential part of the work is going to be to try running your PR in different situations and verify that it behaves as desired.

This also is an issue in a future milestone (M6), which means it's not one we're planning to spend time on in the near term. If you'd like to contribute to Zulip, please take a look at our guide to picking an issue to work on:
https://github.com/zulip/zulip-flutter?#picking-an-issue-to-work-on

PR #1735, opened 2025-07-22
Split avatar, presence, status widgets out from content.dart

gnprice:
These started here with Avatar, one little widget which we didn't have a more natural home for and so tucked in with the content widgets.

That gradually grew from 34 lines to 159 lines, across several widgets.  Then when we built widgets for presence and user status, we put them here as they're natural companions of the avatar. Now it's 367 lines in total, none of it really about message content; so give this all its own separate file.

(I had this thought a few weeks ago, but didn't want to create merge conflicts with the PRs adding presence and user status. Now that #1702 is merged, it's a good time for this move.)

Along the way, the need to move a few bits from `content_checks.dart` reminded me of an old recurring annoyance and prompted me to think of a new solution:

#### 608a24f39 test [nfc]: Consolidate a widgets/checks.dart for trivial checks-extensions

I've long had a recurring small annoyance that the "foo_checks.dart"
file interferes with tab-completion for "foo_test.dart".  For example
as mentioned in this comment:
  https://github.com/zulip/zulip-flutter/pull/1608#discussion_r2155918312

So here's an approach to solve that, beginning with the "checks" files
in test/widgets/.

For updating the imports, I used a bit of Perl:

    $ perl -i -0pe '
          s,import .\K(?!dialog)\w*_checks.dart,checks.dart,g
        ' test/widgets/*_test.dart


chrisbobbe:
Thanks, LGTM! Merged.

PR #1736, opened 2025-07-23
Refactor store: RealmStore; proxy mixins; move more methods to UserStore etc

gnprice:
This PR makes a series of refactors to how PerAccountStore and its substores are organized, which I hope will help keep things clean to understand as the app's model has continued to gain complexity to handle the features we've been adding.

Specifically:
* There's a new RealmStore substore, for realm settings.
* The proxy boilerplate which forwards method calls from PerAccountStore to the various substores has been pushed out of the central PerAccountStore definition, to each substore's respective file.
* A new class HasRealmStore makes it easy for other substores to use data from RealmStore, much like PerAccountStoreBase does for CorePerAccountStore. Similarly a new HasUserStore for data from UserStore.
* The definitions for methods like hasPassedWaitingPeriod, userDisplayEmail, and hasPostingPermission have been moved out of PerAccountStore to individual substores. The reason they'd been on PerAccountStore was because they need data from realm settings; now that that's available from substores, they can move to appropriate substores.
* The logic that was in TopicName.processLikeServer has been moved to RealmStore. It needs data from realm settings (and from CorePerAccountStore) in order to do its work; we'd successfully handled that by having its caller look up that data in advance, but that solution required some contortions to the logic. Now its implementation can look up the particular details it needs directly.
* There's now a crisp rule about what methods can live on substore interfaces like FooStore: their implementations on the substore FooStoreImpl and on the overall PerAccountStore must have identical behavior. (Where that's not the case, the method should instead live separately on FooStoreImpl and PerAccountStore, without claiming to implement a shared interface.) This rule had been very nearly true, with just two exceptions — setServerEmojiData and reconcileMessages — which risked being awfully confusing.

## Selected commit messages

(Most of the most important commits are covered above, so omitted here.)

#### 86ce0252a dartdoc: Stop section-divider comments from getting absorbed into
 docs

Before this change, if you hover in your IDE over an identifier whose
declaration happens to be the first one after one of these section
dividers, the doc shown would begin with a long string of slashes.
After all, the divider line did start with "///".

The choice of "|" as the substitute was fairly arbitrary.  I tried a
space first, and felt it interrupted the divider too much and made it
look like just another line of comments.

Done with a bit of Perl and Git:

    $ perl -i -0pe 's,^\s*//\K(?=//),|,gm' $(git grep -l //// '*.dart')

#### 2b045d259 realm [nfc]: Organize realm settings, from current API doc

To identify which fields are realm settings in the API, so that #668
applies to them, I looked at the current API docs:
  https://zulip.com/api/get-events#realm-update_dict

It turns out a couple of items that had been realm settings are now
something else; one that was a server setting (and even lacked "realm"
in the name: maxFileUploadSizeMib) is now a realm setting updated by
events.  So the comments on the implementation fields need updating.
I'll do that in a separate commit, though, so that this one is
more purely moving code around.

#### db8809bbf emoji [nfc]: Drop setServerEmojiData from main interface

Instead, leave it as a method on EmojiStoreImpl and on the overall
PerAccountStore.

Each of these FooStore types, such as EmojiStore, is implemented by
both a FooStoreImpl class and the overall PerAccountStore.  Nearly all
of these types' methods behave exactly the same on both those types;
the one implementation just forwards to the other.

But this method behaves slightly differently between them: the
PerAccountStore implementation will call notifyListeners, while the
EmojiStoreImpl implementation doesn't (because it can't, not being the
ChangeNotifier itself).

That mismatch, when nearly all the other similar methods have an
exact match, risks confusion.  Mitigate that by removing the method
from the EmojiStore interface, so that the two implementations become
unrelated methods.

(The one other such mismatch is `reconcileMessages`.  We'll deal
with that one too, in a later commit.)

#### f83c9c104 message test [nfc]: Move sendMessage smoke test to follow sendMessage

Ideally this would have been done when sendMessage itself was moved,
in 942aa87a5 (#1455), oops.

#### 9499dbe61 message [nfc]: Consolidate "disposed" checks onto substore methods

For most of these methods, we check the store isn't disposed
within the substore's implementation; for a few, we check in
the proxy implementation.  Simplify the proxies by always checking
in the substore's implementation.


chrisbobbe:
Very helpful, thanks much!! LGTM; merging.

sm-sayedi:
CI is failing, and interestingly, the same problem occurred to me now when I fetched from upstream and then ran `flutter doctor -v` or `flutter --version`!

```
Downloading Darwin arm64 Dart SDK from Flutter engine 2d053c11840d49924aefe84dabb94545a3f9282e...
100   258  100   258    0     0    451      0 --:--:-- --:--:-- --:--:--   451
[/Users/sm-sayedi/Library/Flutter/main/bin/cache/dart-sdk-darwin-arm64.zip]
unzip:  cannot find zipfile directory in one of /Users/sm-sayedi/Library/Flutter/main/bin/cache/dart-sdk-darwin-arm64.zip or
        /Users/sm-sayedi/Library/Flutter/main/bin/cache/dart-sdk-darwin-arm64.zip.zip, and cannot find /Users/sm-sayedi/Library/Flutter/main/bin/cache/dart-sdk-darwin-arm64.zip.ZIP, period.

```

gnprice:
Thanks — seems like definitely an upstream issue, then. Let's discuss in #mobile-team.

sm-sayedi:
The problem seems to be solved. Here's the discussion link: [#mobile-team > Flutter upgrade - downloding Dark SDK fails](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/Flutter.20upgrade.20-.20downloding.20Dark.20SDK.20fails).

PR #1738, opened 2025-07-23
deps: Upgrade Flutter to 3.33.0-1.0.pre-1086

sm-sayedi:

With this upgrade, we can now use the route transition duration in tests to wait the minimum amount of time needed for the route animation to complete.

Related upstream PR: https://github.com/flutter/flutter/pull/171109


gnprice:
Thanks! Looks good; merging.

PR #1739, opened 2025-07-23
analyze: Fix new lints on latest Flutter main

chrisbobbe:
`flutter analyze` has started giving the following, which breaks CI:

   info • Unnecessary comparison to a boolean literal • lib/api/model/events.dart:1177:10 •
          no_literal_bool_comparisons
   info • Unnecessary comparison to a boolean literal • lib/model/channel.dart:158:13 •
          no_literal_bool_comparisons

Fixed.

### DIFF
--- a/lib/api/model/events.dart
+++ b/lib/api/model/events.dart
@@ -1172,10 +1172,7 @@ class UpdateMessageFlagsRemoveEvent extends UpdateMessageFlagsEvent {
   factory UpdateMessageFlagsRemoveEvent.fromJson(Map<String, dynamic> json) {
     final result = _$UpdateMessageFlagsRemoveEventFromJson(json);
     // Crunchy-shell validation
-    if (
-      result.flag == MessageFlag.read
-      && true // (we assume `event_types` has `message` and `update_message_flags`)
-    ) {
+    if (result.flag == MessageFlag.read) {
       result.messageDetails as Map<int, UpdateMessageFlagsMessageDetail>;
     }
     return result;
--- a/lib/model/channel.dart
+++ b/lib/model/channel.dart
@@ -153,9 +153,10 @@ mixin ChannelStore on UserStore {
       case ChannelPostPolicy.any:             return true;
       case ChannelPostPolicy.fullMembers:     {
         if (!role.isAtLeast(UserRole.member)) return false;
-        return role == UserRole.member
-          ? hasPassedWaitingPeriod(user, byDate: byDate)
-          : true;
+        if (role == UserRole.member) {
+          return hasPassedWaitingPeriod(user, byDate: byDate);
+        }
+        return true;
       }
       case ChannelPostPolicy.moderators:      return role.isAtLeast(UserRole.moderator);
       case ChannelPostPolicy.administrators:  return role.isAtLeast(UserRole.administrator);


lib/model/channel.dart
```diff
@@ -153,9 +153,8 @@ mixin ChannelStore on UserStore {
       case ChannelPostPolicy.any:             return true;
       case ChannelPostPolicy.fullMembers:     {
         if (!role.isAtLeast(UserRole.member)) return false;
-        return role == UserRole.member
-          ? hasPassedWaitingPeriod(user, byDate: byDate)
-          : true;
+        return role != UserRole.member
+          || hasPassedWaitingPeriod(user, byDate: byDate);
```
gnprice: The new code here seems more confusing than the old. I feel like to understand it, I have to unpack it back into basically what we have now.

So this looks like an overzealous lint rule to me. But probably easier to accommodate it than to ignore it.

How about:
```dart
        if (!role.isAtLeast(UserRole.member)) return false;
        if (role == UserRole.member) {
          return hasPassedWaitingPeriod(user, byDate: byDate);
        }
        return true;
```

lib/api/model/events.dart
```diff
@@ -1172,10 +1172,8 @@ class UpdateMessageFlagsRemoveEvent extends UpdateMessageFlagsEvent {
   factory UpdateMessageFlagsRemoveEvent.fromJson(Map<String, dynamic> json) {
     final result = _$UpdateMessageFlagsRemoveEventFromJson(json);
     // Crunchy-shell validation
-    if (
-      result.flag == MessageFlag.read
-      && true // (we assume `event_types` has `message` and `update_message_flags`)
-    ) {
+    // (we assume `event_types` has `message` and `update_message_flags`)
+    if (result.flag == MessageFlag.read) {
       result.messageDetails as Map<int, UpdateMessageFlagsMessageDetail>;
```
gnprice: Here I don't really understand what the comment was saying about this `true` :slightly_smiling_face:, so the lint seems helpful.

Looking at the API docs (https://zulip.com/api/get-events#update_message_flags-remove), I don't understand the comment now either. It looks like messageDetails is supposed to be there if the flag is "read". What's the other condition about?

gnprice (commented):
Thanks for taking care of this! Comments below.

sm-sayedi:
Thanks for fixing this. I think this is the reason #1738 is also failing. I didn't expect that "info-level" warning would also break CI!

gnprice:
Yeah, when the PR itself introduces an info-level warning, we want to know that so we fix it before merging the PR — that's how we keep the tree in a state where it's totally clean of warning noise.

But then this becomes one of several ways that a change upstream can cause our CI to start failing for all PRs. The root cause of that really is the fact that we run CI on the latest upstream. Started a chat thread on that: [#mobile-team > CI on latest upstream](https://chat.zulip.org/#narrow/channel/243-mobile-team/topic/CI.20on.20latest.20upstream/with/2228868)

lib/api/model/events.dart
chrisbobbe: Ah: looks like the doc used to say `message_details` was "Present if `message` and `update_message_flags` are both present in `event_types` and the `flag` is `read` and the `op` is `remove`." But the `event_types` part was deleted in zulip/zulip@48a1cf04d07d6c7e1d823571f45b9c0a72d382f3.

chrisbobbe:
Thanks for the review and that explanation! Revision pushed.

lib/api/model/events.dart
gnprice: Aha — thanks for tracking that down.

gnprice:
Thanks! Looks good; merging.

PR #1741, opened 2025-07-24
lightbox test: Skip a no-hero test, pending rework for new page transitions

gnprice:
This test relies on background assumptions about how the "back" page transition work which are true of the current default transitions on Android (the default target platform of tests), but are broken by the upcoming introduction of predictive back:
  https://github.com/flutter/flutter/pull/165832#issuecomment-3111641360

For a quick unblocking of the upstream change, skip the test.

In principle there's no reason a test like this can't work with the upcoming upstream behavior; it'll just need to work a little differently.  But leave that for a follow-up change.


chrisbobbe:
LGTM! Please merge at will once CI passes.

justinmc:
Thank you guys!

Issue #1728, opened 2025-07-22
Remove renderKatex feature flag, and simplify code


gnprice:
We've had KaTeX supported in the last several releases, with the `renderKatex` experimental feature flag defaulting to true; with #1720 we're doing the same thing in main. We don't plan to ever turn the flag back off.

It's therefore time we should remove the flag and make the simplifications that that enables, in particular cutting out the code that was conditioned on the flag being false. See the doc on [GlobalSettingType.experimentalFeatureFlag]:
```dart
  /// Describes a setting which enables an in-progress feature of the app.
  ///
  /// Sometimes when building a complex feature it's useful to merge PRs that
  /// make partial progress, and then to have the feature's logic gated behind
  /// a setting that serves as a "feature flag".
  /// This enables those working on the feature to enable the flag in order to
  /// see the current incomplete behavior, while for everyone else it remains
  /// disabled and so (barring bugs in the use of the flag itself) has no effect.
  ///
  /// These settings are primarily meant for people developing Zulip to use,
  /// and so appear in an out-of-the-way part of the settings UI.
  ///
  /// Settings of this kind are costly to the health of the codebase if
  /// allowed to accumulate.  Most features don't need one, even features that
  /// take two or three PRs to implement.  See discussion at:
  ///   https://github.com/zulip/zulip-flutter/issues/1409#issuecomment-2725793787
  /// When a feature flag is introduced, take care to drive the project to
  /// completion, either by merge or removal, so that the flag can be retired
  /// within a period of a few weeks or months.
  experimentalFeatureFlag,
```

PR #1742, opened 2025-07-24
katex: Remove renderKaTex experimental flag

rajveermalviya:
This flag has been enabled since v0.0.29 (beta) release, and we don't plan to disable it. So remove the flag completely, simplifying related code.

Fixes: #1728

### DIFF
--- a/test/model/content_test.dart
+++ b/test/model/content_test.dart
@@ -6,7 +6,6 @@ import 'package:stack_trace/stack_trace.dart';
 import 'package:test/scaffolding.dart';
 import 'package:zulip/model/code_block.dart';
 import 'package:zulip/model/content.dart';
-import 'package:zulip/model/settings.dart';
 import 'package:zulip/model/katex.dart';
 
 import 'binding.dart';
@@ -528,6 +527,20 @@ class ContentExample {
       ]),
     ]));
 
+  // A test message to test the fallback behaviour of KaTeX implementation.
+  static final mathInlineUnknown = ContentExample.inline(
+    'inline math',
+    null, // r"$$ \lambda $$" (hypothetical server variation)
+    expectedText: r'\lambda',
+    '<p><span class="katex">'
+      '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML"><semantics><mrow><mi>λ</mi></mrow>'
+        '<annotation encoding="application/x-tex"> \\lambda </annotation></semantics></math></span>'
+      '<span class="katex-html" aria-hidden="true">'
+        '<span class="base unknown">' // Server doesn't generate this 'unknown' class.
+        '<span class="strut" style="height:0.6944em;"></span>'
+        '<span class="mord mathnormal">λ</span></span></span></span></p>',
+    MathInlineNode(texSource: r'\lambda', nodes: null));
+
   static const mathBlock = ContentExample(
     'math block',
     "```math\n\\lambda\n```",
@@ -546,6 +559,20 @@ class ContentExample {
           text: 'λ'),
       ]),
     ])]);
+
+  // A test message to test the fallback behaviour of KaTeX implementation.
+  static const mathBlockUnknown = ContentExample(
+    'math block unknown, fallback to TeX source',
+    null, // r"```math\n\lambda\n```" (hypothetical server variation)
+    expectedText: r'\lambda',
+    '<p><span class="katex-display"><span class="katex">'
+      '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mi>λ</mi></mrow>'
+        '<annotation encoding="application/x-tex">\\lambda</annotation></semantics></math></span>'
+      '<span class="katex-html" aria-hidden="true">'
+        '<span class="base unknown">' // Server doesn't generate this 'unknown' class.
+          '<span class="strut" style="height:0.6944em;"></span>'
+          '<span class="mord mathnormal">λ</span></span></span></span></span></p>',
+    [MathBlockNode(texSource: r'\lambda', nodes: null)]);
 
   static const mathBlocksMultipleInParagraph = ContentExample(
     'math blocks, multiple in paragraph',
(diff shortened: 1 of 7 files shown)


test/widgets/content_test.dart
```diff
@@ -1017,50 +1002,27 @@ void main() {
...
-      const html = '<span class="katex">'
-        '<span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML"><semantics><mrow><mi>λ</mi></mrow>'
-          '<annotation encoding="application/x-tex"> \\lambda </annotation></semantics></math></span>'
-        '<span class="katex-html" aria-hidden="true"><span class="base"><span class="strut" style="height:0.6944em;"></span><span class="mord mathnormal">λ</span></span></span></span>';
-      await checkFontSizeRatio(tester,
-        targetHtml: html,
-        targetFontSizeFinder: mkTargetFontSizeFinderFromPattern(r'λ'));
-    }, skip: true // TODO(#46): adapt this test
-                  //   (it needs a more complex targetFontSizeFinder;
-                  //    see other uses in this file for examples.)
```
gnprice, line 1038: nit:

    content test [nfc]: Drop duplicate skipped KaTeX font-size ratio test

rather, it's a skipped duplicate such test — there's another KaTeX font-size ratio test, which this is a duplicate of, but there's no other *skipped* KaTeX font-size ratio test, so it can't be a duplicate within that category.

test/widgets/content_test.dart
```diff
@@ -1017,50 +1002,27 @@ void main() {
...
-      final globalSettings = testBinding.globalStore.settings;
-      await globalSettings.setBool(BoolGlobalSetting.renderKatex, false);
-
+    testWidgets('displays KaTeX content', (tester) async {
       await prepareContent(tester, plainContent(ContentExample.mathInline.html));
-      tester.widget(find.text(r'\lambda', findRichText: true));
+      tester.widget(find.text('λ', findRichText: true));
     });
 
-    testWidgets('displays KaTeX content; experimental flag enabled', (tester) async {
```
gnprice, line 1064: nit: make summary line more specific

    content test [nfc]: Reorder KaTeX widget tests
    
    With renderKatex flag enabled (and soon to be default behaviour)
    being first.

There's room in that summary line to put some of that information from the body, to give an idea of what the reordering is about. Like:
> content test [nfc]: Reorder KaTeX widget tests, enabled before disabled

test/widgets/content_test.dart
gnprice: Also please add a sentence to the commit message clarifying that the existing test this duplicates is the one above it, not the one below. I initially read it as the latter (because the descriptions match — my conflict-resolution error, probably, somewhere in the chain of cherry-picking that led to b073c6b26f785ff72530e23e768e58d33d30211b / #1270), and was going to comment that it's not actually a duplicate of that :slightly_smiling_face:

test/model/content_test.dart
```diff
@@ -528,6 +527,20 @@ class ContentExample {
       ]),
     ]));
 
+  // A test message to test the fallback behaviour of KaTeX implementation.
+  static final mathInlineUnknown = ContentExample.inline(
+    'inline math',
+    r"$$ \lambda $$",
```
gnprice: Since this HTML isn't something the server actually generates, this `markdown` field should be null. See its doc:
```dart
  /// The Zulip Markdown source, if any, that the server renders as [html].
  ///
  /// This is useful for reproducing the example content for live use in the
  /// app, and as a starting point for variations on it.
  ///
  /// Currently the test suite does not verify the relationship between
  /// [markdown] and [html].
  ///
  /// If there is no known Markdown that a Zulip server can render as [html],
  /// then this should be null and a comment should explain why the test uses
  /// such an example.
  final String? markdown;
```

So e.g., following the example `emojiUnicodeClassesFlipped`:

```suggestion
  static final mathInlineUnknown = ContentExample.inline(
    'inline math',
    null, // r"$$ \lambda $$" (hypothetical server variation)
```

test/widgets/content_test.dart
```diff
@@ -1017,50 +1002,27 @@ void main() {
...
-                  //    see other uses in this file for examples.)
-    );
-
-    testWidgets('maintains font-size ratio with surrounding text, when showing TeX source', (tester) async {
-      addTearDown(testBinding.reset);
-      final globalSettings = testBinding.globalStore.settings;
-      await globalSettings.setBool(BoolGlobalSetting.renderKatex, false);
-
-      const html = '<span class="katex">'
+    testWidgets('maintains font-size ratio with surrounding text, when fallback to TeX source', (tester) async {
```
gnprice: nit: "fallback" is a noun, "fall back" the verb:
```suggestion
    testWidgets('maintains font-size ratio with surrounding text, when falling back to TeX source', (tester) async {
```

gnprice (commented):
Thanks for cleaning this up! Generally all looks good; small comments below.

rajveermalviya:

test/model/content_test.dart
```diff
@@ -511,7 +510,7 @@ class ContentExample {
 
   static final mathInline = ContentExample.inline(
     'inline math',
-    r"$$ \lambda $$",
+    null, // r"$$ \lambda $$" (hypothetical server variation)
```
gnprice: I think you made this edit on a different example than intended :slightly_smiling_face:

rajveermalviya: Ahh, should be fixed now.

rajveermalviya:

gnprice:
Thanks! Merging, with a tweak in one commit message (cf https://github.com/zulip/zulip-flutter/pull/1742#discussion_r2229237016):
```
$ git range-diff origin pr/1742 @
1:  edfb68198 ! 1:  8c9a7b0ce content test [nfc]: Drop skipped duplicate KaTeX font-size ratio test
    @@ Commit message
         This skipped test is a duplicate of the one above it, but
         incorrectly has the description of the one below it. Probably a
         result of conflict-resolution error in b073c6b26 which introduced
    -    this duplicate.
    +    this duplicate, or another in the chain of cherry-picks that led to
    +    that commit.
     
      ## test/widgets/content_test.dart ##
     @@ test/widgets/content_test.dart: void main() {
2:  95c25c6c3 = 2:  6459af689 content test [nfc]: Reorder KaTeX widget tests, enabled before disabled
3:  ccf262f4c = 3:  da5c1f706 katex: Remove renderKaTex experimental flag
```

PR #1744, opened 2025-07-24
new-dm: Exclude deactivated users

### LINKED ISSUE #1743: Exclude deactivated users in new-DM screen
When composing a DM on the web and searching for the users you want to send the DM to, deactivated users are omitted from the list. We should do the same.

gnprice:
Fixes #1743.


gnprice:
CI failing only with the pair of lint issues fixed in #1739.

chrisbobbe:
Thanks, LGTM! Merged.

PR #1745, opened 2025-07-24
Prep for user-group autocomplete

chrisbobbe:
These three commits add a `rank` field to `MentionAutocompleteResult`, simply choosing 0 for wildcard results and 1 for user results, and a `bucketSort` on that new field. This is to prepare for more complicated ranking, with more ranks:
- by match quality
- with user-group results interleaved with user results, for #233 

Discussion: [#mobile > user-group mentions #F233 @ 💬](https://chat.zulip.org/#narrow/channel/48-mobile/topic/user-group.20mentions.20.23F233/near/2216353)

@gnprice, I'm curious how you'd like the tests to change, to support that more complicated ranking. Would you be up for demonstrating that by writing a few commits? I think you could do that by converting the `compareByBotStatus` logic into the new bucket-sort-by-rank logic. If we're following web (`sort_recipients` in web/src/typeahead_helper.ts), then some user-group results will need to be interleaved between human- and bot-user results, which calls for humans and bots to be in different buckets. As a first step, I think we'd go from two ranks (wildcard then users) to three ranks (wildcard then humans then bots); is that right?

### DIFF
--- a/lib/model/autocomplete.dart
+++ b/lib/model/autocomplete.dart
@@ -650,10 +650,7 @@ class MentionAutocompleteView extends AutocompleteView<MentionAutocompleteQuery,
   }
 
   MentionAutocompleteResult? _testUser(MentionAutocompleteQuery query, User user) {
-    if (query.testUser(user, store.autocompleteViewManager.autocompleteDataCache, store)) {
-      return UserMentionAutocompleteResult(userId: user.userId);
-    }
-    return null;
+    return query.testUser(user, store.autocompleteViewManager.autocompleteDataCache, store);
   }
 
   @override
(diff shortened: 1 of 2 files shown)


lib/model/autocomplete.dart
```diff
@@ -748,25 +747,47 @@ class MentionAutocompleteQuery extends ComposeAutocompleteQuery {
...
 
   bool _testName(User user, AutocompleteDataCache cache) {
     return _testContainsQueryWords(cache.nameWordsForUser(user));
   }
 
+  /// A measure of a wildcard result's quality in the context of the query,
+  /// from 0 (best) to one less than [_numResultRanks].
+  ///
+  /// See also [_rankUserResult].
+  static const _rankWildcardResult = 0;
```
gnprice, line 779: This name is a verb phrase (well, it could be extra elliptical and mean "the rank for a wildcard result", but I want to read it as a verb phrase: "rank a wildcard result"), so it's a bit odd for it to be a constant value.

Maybe go ahead and make it a function that returns a constant? That'll better demonstrate the future structure you intend anyway.

gnprice (commented):
Sure, that plan sounds good. I'll take a look.

gnprice:
> Would you be up for demonstrating that by writing a few commits? I think you could do that by converting the `compareByBotStatus` logic into the new bucket-sort-by-rank logic.

Sent as #1748 — please take a look.

chrisbobbe:
Closing, since this has served its purpose (#1775 was merged). Thanks much! 🙂

PR #1746, opened 2025-07-24
deps: Update CocoaPods pods (tools/upgrade pod)

rajveermalviya:
Looks like #1738 also upgraded some pub dependencies, so CocoaPods dependencies are now out-of-date, resulting in the following error when building on iOS:
```
…
Error: CocoaPods's specs repository is too out-of-date to satisfy dependencies.
To update the CocoaPods specs, run:
  pod repo update



Error running pod install
Error launching application on iPhone 16 Pro.
```

This PR fixes that by updating CocoaPods dependencies.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -42,10 +42,10 @@ PODS:
   - Firebase/Messaging (11.15.0):
     - Firebase/CoreOnly
     - FirebaseMessaging (~> 11.15.0)
-  - firebase_core (3.15.1):
+  - firebase_core (3.15.2):
     - Firebase/CoreOnly (= 11.15.0)
     - Flutter
-  - firebase_messaging (15.2.9):
+  - firebase_messaging (15.2.10):
     - Firebase/Messaging (= 11.15.0)
     - firebase_core
     - Flutter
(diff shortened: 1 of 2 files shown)


gnprice:
Hmm I see — #1738's description said it was upgrading Flutter (and supporting libraries), but in fact it upgraded a bunch of third-party libraries as well. It looks a lot like the output of `flutter pub upgrade`. Thanks for catching that!

I'll start by reverting that commit; then we can do the upgrade again but in individual pieces.

If `tools/upgrade flutter-local` produced that commit in #1738 (/cc @sm-sayedi), then there's probably also something we can do in the script to make it more robust against whatever situation led to that.

gnprice:
Did that revert as #1749. Thanks again for spotting this.

sm-sayedi:
Sorry that it caused the inconvenience. It was not `tools/upgrade flutter-local` that made changes to the other package. When I first opened that PR and CI failed, one of the messages there was that the analyzer needs to be updated by using `flutter pub upgrade`. So I quickly amended that commit and pushed to see if it fixes CI which it didn't and turned out that #1739 was the fix. I wanted to leave a comment in #1738 saying that l will push again (and not include additional package changes) when that fix is merged, but I totally forgot it.

gnprice:
Got it, thanks.

The one step I'd ask that you do differently next time, then, is to do the `flutter pub upgrade` as a separate commit. (And if you're not sure whether it should be squashed into the original commit, mention that on the PR — it's easy to squash later.)

That way it's clear where the changes came from, which makes it easier to review them quickly without missing something.

PR #1748, opened 2025-07-25
Start using `rank` field on mention autocomplete

gnprice:
Stacked atop #1745. This makes some refactors to the tests for the ordering of autocomplete results, and then demonstrates changing that ordering by moving the human-vs-bot distinction into the "ranking" layer, as suggested in #1745.

Now that I've spelled that out, complete with its consequences as seen in the tests, it doesn't actually sound desirable as a product change: if there's a bot participating in the current thread and I start typing its name, wouldn't we want that as a top result? The distinctions in the "ranking" layer take priority over those in the initial sorting of candidates, so this change causes bots to always come after all other matches.

Anyway, this demonstrates how such a change would work out in the tests, though. And if that really is the behavior that web has, then we should probably just do it — in order to minimize avoidable differences from web's logic, to better clear the way to proceeding with user groups as autocomplete results, #233 — and consider coming back and changing it on both platforms later.

### DIFF
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -409,7 +409,9 @@ void main() {
     void doCheck(String rawQuery, User user, bool expected) {
       final result = MentionAutocompleteQuery(rawQuery)
         .testUser(user, AutocompleteDataCache(), store);
-      expected ? check(result).isTrue() : check(result).isFalse();
+      expected
+        ? check(result).isA<UserMentionAutocompleteResult>()
+        : check(result).isNull();
     }
 
     test('user is always excluded when not active regardless of other criteria', () {
--- a/test/model/autocomplete_test.dart
+++ b/test/model/autocomplete_test.dart
@@ -599,26 +599,6 @@ void main() {
       });
     });
 
-    group('compareByBotStatus', () {
-      final humanUser = eg.user(isBot: false);
-      final botUser = eg.user(isBot: true);
-
-      int compareAB(User a, User b) => MentionAutocompleteView.compareByBotStatus(a, b);
-
-      test('userA is human, userB is bot -> favor userA', () {
-        check(compareAB(humanUser, botUser)).isLessThan(0);
-      });
-
-      test('userA is bot, userB is human -> favor userB', () {
-        check(compareAB(botUser, humanUser)).isGreaterThan(0);
-      });
-
-      test('both users have the same bot status -> favor none', () {
-        check(compareAB(humanUser, humanUser)).equals(0);
-        check(compareAB(botUser, botUser)).equals(0);
-      });
-    });
-
     group('compareByAlphabeticalOrder', () {
       int compareAB(String aName, String bName) => MentionAutocompleteView.compareByAlphabeticalOrder(
         eg.user(fullName: aName), eg.user(fullName: bName), store: store);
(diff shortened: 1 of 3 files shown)


test/model/autocomplete_test.dart
```diff
@@ -701,23 +622,23 @@ void main() {
...
+      test('ChannelNarrow: human/bot > stream recency > DM recency > name', () async {
         // Same principle as for TopicNarrow; see that test case above.
         final users = [
-          eg.user(fullName: 'Z', isBot: true), // wins by stream recency
-          eg.user(fullName: 'Y', isBot: true), // runner-up, by DM recency
-          eg.user(fullName: 'X', isBot: false), // runner-up, by human-vs-bot
+          eg.user(fullName: 'Z', isBot: false), // wins by human-vs-bot
+          eg.user(fullName: 'Y', isBot: true), // runner-up, by stream recency
+          eg.user(fullName: 'X', isBot: true), // runner-up, by DM recency
           eg.user(fullName: 'A', isBot: true), // runner-up, by name
```
gnprice, line 631: These test changes happen when putting human-vs-bot in the match ranking, because there are these existing tests that check how that distinction compares in ranking priority with other criteria. The outcome there changes, so it seems good to reflect that in the tests.

But for introducing user groups to these results (#233), as from #1745, I wouldn't recommend adding more tests of this kind. Instead…

test/model/autocomplete_test.dart
```diff
@@ -1000,15 +980,15 @@ void main() {
...
     }
 
-    test('wildcards, then users', () {
+    test('wildcards, non-bot users, then bots', () {
       checkSameRank('', WildcardMentionOption.all, WildcardMentionOption.topic);
-      checkPrecedes('', WildcardMentionOption.topic, eg.user());
-      checkSameRank('', eg.user(), eg.user());
+      checkPrecedes('', WildcardMentionOption.topic, eg.user(isBot: false));
+      checkPrecedes('', eg.user(isBot: false), eg.user(isBot: true));
     });
```
gnprice: … I think the ranking of user groups vs. users can be tested well with just this class of test. The tests in `test/model/emoji_test.dart` like "ranks match quality exact/prefix/word-aligned/other" and its neighbors provide a useful set of examples.

The structure of the result ordering (IIUC) is that the rank prevails, and when the ranks are equal the other ordering is effectively used as a tie-breaker. So the tests can rely on that structure: tests of this kind focus purely on the ranks; and then for all the criteria that sort users of equal match rank, the tests above (the ones that call the various `compareByFoo` methods and `debugCompareUsers`) can remain in place covering those.

test/model/autocomplete_test.dart
```diff
@@ -1000,15 +980,15 @@ void main() {
...
 
     test('wildcard-vs-user more significant than match quality', () {
```
gnprice, line 989: Oh, and rereading those rank tests in `test/model/emoji_test.dart` reminds me of one other type of test that would be good to include:
```dart
    test('full list of ranks', () {
      check([
        rankOf('o', unicode(['o'])),              // exact (generic)
        rankOf('o', octopus),                     // prefix popular
        rankOf('o', workingOnIt),                 // word-aligned popular
        rankOf('o', realmCandidate('open_book')), // prefix realm
        rankOf('z', zulipCandidate()),            //  == prefix :zulip:
        rankOf('y', realmCandidate('thank_you')), // word-aligned realm
            // (word-aligned :zulip: is impossible because the name is one word)
        rankOf('o', unicode(['ok'])),             // prefix generic
        rankOf('o', unicode(['squared_ok'])),     // word-aligned generic
        rankOf('o', realmCandidate('yo')),        // other realm
        rankOf('p', zulipCandidate()),            //  == other :zulip:
        rankOf('o', unicode(['book'])),           // other generic
        rankOf('o', love),                        //  == other popular
      ]).deepEquals([0, 1, 2, 3, 3, 4, 5, 6, 7, 7, 8, 8]);
    });
```

That can be introduced at the outset even though it'll be trivial, with just two items; then it can grow as more cases and criteria are added.

chrisbobbe:
Closing, since this has served its purpose (#1775 was merged). Thanks much! 🙂

PR #1752, opened 2025-07-25
Omit unread button unsubscribed

loveucifer:
I am sorry for doing a pr on this , I already saw that someone was assigned and I saw that he hasn't done a pr yet so I am doing this basically I noticed that If store.subscriptions[message.streamId] returns null, it means the user is not subscribed to that stream. so I just fixed that , please ignore this if the pr is to be completed by the assignee Idk how this works im still learning 
im doing the table thing for the first time not sure if its needed but here goes
<table>
<thead>
<tr>
<th>Before (Subscribed Channel)</th>
<th>After (Unsubscribed Channel)</th>
</tr>
</thead>
<tbody>
<tr>
<td>
A screenshot showing the message action sheet with a 'Mark as unread' option.
</td>
<td>
A screenshot showing the message action sheet without the 'Mark as unread' option.
</td>
</tr>
</tbody>
</table>

### DIFF
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -28,20 +28,7 @@ import Flutter
 
     UNUserNotificationCenter.current().delegate = self
 
-    // This is needed to get a push token from APNs.
-    // The user will be prompted for permission only the first time.
-    application.registerForRemoteNotifications()
-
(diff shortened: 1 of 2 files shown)


loveucifer:
somehow the earlier pr for appdelgate keeps sneaking In that's why I closed it but all seems good now

Issue #1587, opened 2025-06-14
Emoji-reaction picker looks odd with plain-text-emojis setting


chrisbobbe:
The touch targets aren't tall enough:

| Default setting | Plain-text setting |
| --- | --- |
| ![image](https://github.com/user-attachments/assets/247cd29e-b5e5-436c-9298-5094dc627c18) | ![image](https://github.com/user-attachments/assets/2b52024f-6a7b-48b2-80bb-baa36b479d02) |

MunibullahShah:
Please assign it to me, I am working on it.

MunibullahShah:
https://github.com/zulip/zulip-flutter/pull/1592

avinashgarlaa:
Hi, I’d like to work on this — could you please assign it to me?

omkaware12:
hi iam currently working on the same issue in my project i know how to deal with this issue please assign it to me

chrisbobbe:
Please take a look at the Zulip project's guide to getting involved with the code:

https://zulip.readthedocs.io/en/latest/contributing/contributing.html#your-first-codebase-contribution
https://zulip.readthedocs.io/en/latest/contributing/contributing.html#picking-an-issue-to-work-on

yash-agarwa-l:
Hi! I’d like to work on this issue.

I noticed the row height becomes inconsistent for text-only emoji entries. My approach is to enforce a minimum row height using a ConstrainedBox (e.g., _emojiSize + 20) so glyph and text-only items render consistently.

I've tested this locally and it fixes the layout.
Please let me know if this approach looks good, and I can proceed with a PR.

chrisbobbe:
Sure, that sounds reasonable, I'll assign this to you.

shreyas-kamat:
Hi, if this issue isn't resolved yet, I have increased the vertical padding and wrapped it in a rounded container and have tested the changes locally in both dark & light mode. Please assign it to me, will submit a PR.

<img width="444" height="898" alt="Image" src="https://github.com/user-attachments/assets/7921ff63-7222-464a-b9dc-88a4edf1b443" />

gnprice:
@shreyas-kamat What does our [contributing guide](https://zulip.readthedocs.io/en/latest/contributing/contributing.html) say about issues that are assigned?

PR #1759, opened 2025-07-28
fix(emoji): Adjust picker layout for plain-text-emojis setting #1587

loveucifer:
| Before | After |
| --- | --- |
| before | after |


chrisbobbe (changes requested):
Hi @loveucifer, thanks for the PR.

- It needs a test, in the "EmojiPicker" group in test/widgets/emoji_reaction_test.dart.
  - The test should have a comment like "Regression test for #1587". It should pass with the change in this commit, and it should fail without the change.
  - In the test, set the "emojiset" to `Emojiset.text`, show the emoji picker, find a list item in there, and check that its height is correct. To check its height, try getting the render object with `tester.renderObject` and looking at its `.size.height`.
- `EmojiPickerListEntry.build` has this comment: "// TODO deduplicate this logic with [_EmojiAutocompleteItem]
  - That can be out of scope for this PR, but the comment is a hint that some other code (in `_EmojiAutocompleteItem`) might have the same problem, with a similar fix. That seems to be the case; could you please make that fix and post screenshots of it too, and write a test for it?
- Please revise the commit message according to the project style (follow links in the README)

loveucifer:
> Hi @loveucifer, thanks for the PR.
> 
> 
> 
> - It needs a test, in the "EmojiPicker" group in test/widgets/emoji_reaction_test.dart.
> 
>   - The test should have a comment like "Regression test for #1587". It should pass with the change in this commit, and it should fail without the change.
> 
>   - In the test, set the "emojiset" to `Emojiset.text`, show the emoji picker, find a list item in there, and check that its height is correct. To check its height, try getting the render object with `tester.renderObject` and looking at its `.size.height`.
> 
> - `EmojiPickerListEntry.build` has this comment: "// TODO deduplicate this logic with [_EmojiAutocompleteItem]
> 
>   - That can be out of scope for this PR, but the comment is a hint that some other code (in `_EmojiAutocompleteItem`) might have the same problem, with a similar fix. That seems to be the case; could you please make that fix and post screenshots of it too, and write a test for it?
> 
> - Please revise the commit message according to the project style (follow links in the README)

oh okay will work on this

loveucifer:
> Hi @loveucifer, thanks for the PR.
> 
> 
> 
> - It needs a test, in the "EmojiPicker" group in test/widgets/emoji_reaction_test.dart.
> 
>   - The test should have a comment like "Regression test for #1587". It should pass with the change in this commit, and it should fail without the change.
> 
>   - In the test, set the "emojiset" to `Emojiset.text`, show the emoji picker, find a list item in there, and check that its height is correct. To check its height, try getting the render object with `tester.renderObject` and looking at its `.size.height`.
> 
> - `EmojiPickerListEntry.build` has this comment: "// TODO deduplicate this logic with [_EmojiAutocompleteItem]
> 
>   - That can be out of scope for this PR, but the comment is a hint that some other code (in `_EmojiAutocompleteItem`) might have the same problem, with a similar fix. That seems to be the case; could you please make that fix and post screenshots of it too, and write a test for it?
> 
> - Please revise the commit message according to the project style (follow links in the README)

oh okay will work on this

loveucifer:
wrote the test for emojI-reaction but having a hard time with autocomplete test dont know why will try to send one asap

lib/widgets/emoji_reaction.dart
```diff
@@ -577,12 +577,12 @@ class EmojiPickerListEntry extends StatelessWidget {
...
     // TODO deduplicate this logic with [_EmojiAutocompleteItem]
     final emojiDisplay = emoji.emojiDisplay.resolve(store.userSettings);
-    final Widget? glyph = switch (emojiDisplay) {
+    final Widget glyph = switch (emojiDisplay) {
       ImageEmojiDisplay() =>
         ImageEmojiWidget(size: _emojiSize, emojiDisplay: emojiDisplay),
       UnicodeEmojiDisplay() =>
         UnicodeEmojiWidget(size: _emojiSize, emojiDisplay: emojiDisplay),
-      TextEmojiDisplay() => null, // The text is already shown separately.
+      TextEmojiDisplay() => Text(' ', style: TextStyle(fontSize: _emojiSize)),
```
chrisbobbe: > When using the text-only emojis the emoji picker list items did not scale correctly with the system's text scale factor in the earlier pr . i think this was because the placeholder for the emoji glyph had a fixed, unscalable height, preventing the row from growing as we wanted.

Image and Unicode emoji don't scale with the system text-size setting (try it), so we shouldn't scale the empty-square placeholder either.

loveucifer:
> wrote the test for emojI-reaction but having a hard time with autocomplete test dont know why will try to send one asap

I had some help from llm to write the test for autocomplete but it did pass for now , couldn't figure out why the one I wrote kept falling even though I applied almost the same logic

loveucifer:
> > wrote the test for emojI-reaction but having a hard time with autocomplete test dont know why will try to send one asap
> 
> I had some help from llm to write the test for autocomplete but it did pass for now , couldn't figure out why the one I wrote kept falling even though I applied almost the same logic

actually circling back and seeing why it failed I think it was a syntax error I think I used `check(finder).findsOneWidget()` instead of `check(finder).findsOne()`

chrisbobbe:
OK, please comment when this is ready for another review. If you need help, see our README, in particular the "asking great questions" article linked from there: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#asking-questions-getting-help

loveucifer:
> OK, please comment when this is ready for another review. If you need help, see our README, in particular the "asking great questions" article linked from there: https://github.com/zulip/zulip-flutter?tab=readme-ov-file#asking-questions-getting-help 

i have done it , it is running through the checks rn , please do check and give a review when possible

lib/widgets/autocomplete.dart
```diff
@@ -342,14 +342,15 @@ class _EmojiAutocompleteItem extends StatelessWidget {
     final designVariables = DesignVariables.of(context);
     final candidate = option.candidate;
 
-    // TODO deduplicate this logic with [EmojiPickerListEntry]
+    // This logic is now corrected to match EmojiPickerListEntry as per todo
```
chrisbobbe: This TODO still applies; please don't delete it. The comment doesn't need to be changed. ("Deduplicate" means make a helper function or something that can be invoked in both places. As I mentioned earlier, you don't have to write that helper function as part of this PR.)

lib/widgets/autocomplete.dart
```diff
@@ -342,14 +342,15 @@ class _EmojiAutocompleteItem extends StatelessWidget {
...
     final emojiDisplay = candidate.emojiDisplay.resolve(store.userSettings);
-    final Widget? glyph = switch (emojiDisplay) {
+    final Widget glyph = switch (emojiDisplay) {
       ImageEmojiDisplay() =>
         ImageEmojiWidget(size: _size, emojiDisplay: emojiDisplay),
       UnicodeEmojiDisplay() =>
         UnicodeEmojiWidget(size: _size, emojiDisplay: emojiDisplay),
-      TextEmojiDisplay() => null, // The text is already shown separately.
+      // Use a simple SizedBox for a perfect, non-scaling placeholder.
+      TextEmojiDisplay() => const SizedBox.square(dimension: _size),
```
chrisbobbe:
```suggestion
      // The text is already shown; just put a placeholder here to set the row height.
      TextEmojiDisplay() => const SizedBox.square(dimension: _size),
```

lib/widgets/autocomplete.dart
```diff
@@ -367,11 +368,10 @@ class _EmojiAutocompleteItem extends StatelessWidget {
     return Padding(
       padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
       child: Row(children: [
-        if (glyph != null) ...[
-          Padding(padding: const EdgeInsets.all(6),
-            child: glyph),
-          const SizedBox(width: 6),
-        ],
+        // The `if (glyph != null)` check is removed, as the glyph is always present.
```
chrisbobbe: This comment isn't needed; it refers to something that isn't visible in the code.

lib/widgets/emoji_reaction.dart
```diff
@@ -577,12 +577,13 @@ class EmojiPickerListEntry extends StatelessWidget {
...
     final emojiDisplay = emoji.emojiDisplay.resolve(store.userSettings);
-    final Widget? glyph = switch (emojiDisplay) {
+    final Widget glyph = switch (emojiDisplay) {
       ImageEmojiDisplay() =>
         ImageEmojiWidget(size: _emojiSize, emojiDisplay: emojiDisplay),
       UnicodeEmojiDisplay() =>
         UnicodeEmojiWidget(size: _emojiSize, emojiDisplay: emojiDisplay),
-      TextEmojiDisplay() => null, // The text is already shown separately.
+      TextEmojiDisplay() => Text(' ',
+        style: TextStyle(fontSize: _emojiSize), textScaler: TextScaler.noScaling),
```
chrisbobbe: We should use an empty square placeholder here too.

test/widgets/autocomplete_test.dart
```diff
@@ -372,6 +372,56 @@ void main() {
...
+      final composeInputFinder = await setupToComposeInput(tester);
+
+      // Set up a realm emoji so we have a reliable image emoji to test against.
+      await store.handleEvent(RealmEmojiUpdateEvent(id: 1, realmEmoji: {
+        '1': eg.realmEmojiItem(emojiCode: '1', emojiName: 'an_image_emoji'),
+      }));
+      await tester.pumpAndSettle(); // Ensure store update is fully processed.
+
+      // Phase 1: Get the height of the item when it has an image.
+      // This two-step `enterText` call is a more reliable way to trigger the autocomplete view.
```
chrisbobbe:
```suggestion
      // TODO(#226): Remove this extra edit when this bug is fixed.
```

(here and the other place(s) where it applies)

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,43 @@ void main() {
...
+      await tester.pumpAndSettle(); // Let dismiss animations finish.
+
+      // We are now back at the message list. Let's start over to see the setting change.
+      // now we Long-press the message to open the action sheet again.
+      await tester.longPress(find.byType(MessageContent));
+      await tester.pumpAndSettle();
+
+      // now we Tap the button to open the emoji picker, which will now use the new setting.
+      await tester.tap(find.byIcon(ZulipIcons.chevron_right));
+      await tester.pumpAndSettle();
```
chrisbobbe: Is it necessary to close and re-open the emoji picker? What happens if the test doesn't do that?

chrisbobbe (changes requested):
Comments below. Please squash the two commits together and do another pass on the commit message to make sure it matches our style.

test/widgets/emoji_reaction_test.dart
loveucifer: I thought that byy completely closing the emoji picker and re-opening it, we could be be  sure that it was a brand-new widget built with the updated "Text only" setting, is that bad?

loveucifer:
> Comments below. Please squash the two commits together and do another pass on the commit message to make sure it matches our style.

as mentioned in my other pr im not sure if im doing the rebase thing correctly I havent used it yet ,will try another go again, also trying another pass on the commit style and as an update the test still works without reopening the emoji-picker

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,32 @@ void main() {
...
+          value: Emojiset.text));
+        // The picker is already open and will rebuild with the new setting.
+        await tester.pumpAndSettle();
+
+        // Find a list entry and check its height.
+        final listEntryFinder = find.byType(EmojiPickerListEntry).first;
+        await tester.ensureVisible(listEntryFinder);
+
+        final renderBox = tester.renderObject<RenderBox>(listEntryFinder);
+        check(renderBox.size.height).isLessThan(45.0);
```
chrisbobbe: Why check that the height is _less than_ a number? The point of this test is to make sure the height is large enough. In other words (as you've written in a comment at the top), it's a regression test for #1587.

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,32 @@ void main() {
         debugNetworkImageHttpClientProvider = null;
       });
 
+      testWidgets('text emoji list items must have correct height', (tester) async {
+        // Regression test for #1587
+        tester.platformDispatcher.textScaleFactorTestValue = 2.0;
+        addTearDown(tester.platformDispatcher.clearTextScaleFactorTestValue);
```
chrisbobbe: The reproduction recipe for #1587 doesn't involve setting the system text-size setting, so I don't think these lines are needed in a regression test for #1587. What happens if we remove them?

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,32 @@ void main() {
...
+
+        final message = eg.streamMessage();
+        // Set up the emoji picker, which leaves it open.
+        await setupEmojiPicker(tester, message: message, narrow: TopicNarrow.ofMessage(message));
```
chrisbobbe: nit: Remove the comment; it describes something that's already obvious from reading the code

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,32 @@ void main() {
...
+
+        // Now, change the setting to text-only emoji.
```
chrisbobbe: nit: also remove this comment for the same reason

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,32 @@ void main() {
...
+        await store.handleEvent(UserSettingsUpdateEvent(id: 1,
+          property: UserSettingName.emojiset,
+          value: Emojiset.text));
+        // The picker is already open and will rebuild with the new setting.
```
chrisbobbe: nit: and this one

test/widgets/emoji_reaction_test.dart
```diff
@@ -556,6 +556,32 @@ void main() {
...
+        await tester.pumpAndSettle();
+
+        // Find a list entry and check its height.
```
chrisbobbe: nit: and this one

test/widgets/autocomplete_test.dart
```diff
@@ -372,6 +372,53 @@ void main() {
...
     }
 
+    testWidgets('text emoji autocomplete item must have correct height', (tester) async {
+      // Regression test for #1587
+      tester.platformDispatcher.textScaleFactorTestValue = 2.0;
+      addTearDown(tester.platformDispatcher.clearTextScaleFactorTestValue);
+
+      final composeInputFinder = await setupToComposeInput(tester);
+
+      // Set up a realm emoji so we have a reliable image emoji to test against.
```
chrisbobbe: nit: Remove this comment; it explains something that's already obvious from reading the code

test/widgets/autocomplete_test.dart
```diff
@@ -372,6 +372,53 @@ void main() {
...
+      await store.handleEvent(RealmEmojiUpdateEvent(id: 1, realmEmoji: {
+        '1': eg.realmEmojiItem(emojiCode: '1', emojiName: 'an_image_emoji'),
+      }));
+      await tester.pumpAndSettle(); // Ensure store update is fully processed.
```
chrisbobbe: Can this just be an `await tester.pump()`? Also the comment doesn't seem necessary.

test/widgets/autocomplete_test.dart
```diff
@@ -372,6 +372,53 @@ void main() {
...
+      await tester.pumpAndSettle();
+
+      final imageItemFinder = find.ancestor(
+        of: find.text('an_image_emoji'),
+        matching: find.byType(InkWell),
+      );
+      check(imageItemFinder).findsOne();
+      final imageItemHeight = tester.getSize(imageItemFinder).height;
+
+      //TODO(#226): Remove this extra edit when this bug is fixed.
```
chrisbobbe: nit: space after `//`

test/widgets/autocomplete_test.dart
```diff
@@ -372,6 +372,53 @@ void main() {
...
+      // Set up a realm emoji so we have a reliable image emoji to test against.
+      await store.handleEvent(RealmEmojiUpdateEvent(id: 1, realmEmoji: {
+        '1': eg.realmEmojiItem(emojiCode: '1', emojiName: 'an_image_emoji'),
+      }));
+      await tester.pumpAndSettle(); // Ensure store update is fully processed.
+
+      // TODO(#226): Remove this extra edit when this bug is fixed.
+      await tester.enterText(composeInputFinder, ':an_imag');
+      await tester.enterText(composeInputFinder, ':an_image');
+      await tester.pumpAndSettle();
```
chrisbobbe, line 387: nit: Can this just be an `await tester.pump()`? What do the other tests do?

test/widgets/autocomplete_test.dart
```diff
@@ -372,6 +372,53 @@ void main() {
...
+
+      final imageItemFinder = find.ancestor(
+        of: find.text('an_image_emoji'),
+        matching: find.byType(InkWell),
+      );
```
chrisbobbe: Can this be `find.widgetWithText(InkWell, 'an_image_emoji')`? (and similarly below)

chrisbobbe (changes requested):
Thanks, comments below.

Please find the place in our commit-discipline doc where it says how to wrap the commit-message body text. If that's difficult for you, please explain where you looked and how we can make it more obvious for future contributors.

Looking back at the issue #1587 and your screenshots (thanks for posting those!), I've also realized that we don't actually want to add any horizontal whitespace, only vertical whitespace. The empty space on the left looks odd. So in the text-emoji case, instead of creating a padded `SizedBox.square`, we should put a zero-width `SizedBox(height: n)` where `n` is whatever would make the rows the same height as in the non-text-emoji case.

loveucifer:
> So in the text-emoji case, instead of creating a padded `SizedBox.square`, we should put a zero-width `SizedBox(height: n)` where `n` is whatever would make the rows the same height as in the non-text-emoji case.




I gave this code initially -- ` final imageHeight = imageRenderBox.size.height;` but the test failed and gave output `he following TestFailure was thrown running a test:
Expected: a double that:
  is within <1.0> of <40.0>
Actual: <44.0>
Which: differs by <4.0>` so im changing it to this for now to see how it will go 
`40.0 (Target Height) - 10.0 (Top Padding) - 10.0 (Bottom Padding) = 20.0`

loveucifer:
> Thanks, comments below.
> 
> Please find the place in our commit-discipline doc where it says how to wrap the commit-message body text. If that's difficult for you, please explain where you looked and how we can make it more obvious for future contributors.
> 
> Looking back at the issue #1587 and your screenshots (thanks for posting those!), I've also realized that we don't actually want to add any horizontal whitespace, only vertical whitespace. The empty space on the left looks odd. So in the text-emoji case, instead of creating a padded `SizedBox.square`, we should put a zero-width `SizedBox(height: n)` where `n` is whatever would make the rows the same height as in the non-text-emoji case.

also updating the image addressing this new change for autocomplete

pubspec.lock
```diff
@@ -1079,26 +1079,26 @@ packages:
     dependency: "direct dev"
     description:
       name: test
-      sha256: "65e29d831719be0591f7b3b1a32a3cda258ec98c58c7b25f7b84241bc31215bb"
+      sha256: "75906bf273541b676716d1ca7627a17e4c4070a3a16272b7a3dc7da3b9f3f6b7"
```
chrisbobbe: Remove these changes in pubspec.lock that aren't part of the change this commit is about

lib/widgets/autocomplete.dart
```diff
@@ -344,12 +344,13 @@ class _EmojiAutocompleteItem extends StatelessWidget {
 
     // TODO deduplicate this logic with [EmojiPickerListEntry]
     final emojiDisplay = candidate.emojiDisplay.resolve(store.userSettings);
-    final Widget? glyph = switch (emojiDisplay) {
+    final Widget glyph = switch (emojiDisplay) {
```
chrisbobbe: The commit message says:

```
Text-only emoji list items in the emoji picker and autocomplete
views had incorrect row heights, causing visual inconsistencies. This
was due to the placeholder  not correctly accounting for
the vertical padding applied by its parent widget, making text-only
rows taller than rows with image emoji.
```

No, "making text-only rows taller than rows with image emoji" is not true: look at the screenshots in the issue, #1587, which clearly show that the text-only rows are shorter than rows with image emoji.

Also I don't know what "placeholder" you're referring to in `main`.

lib/widgets/emoji_reaction.dart
```diff
@@ -580,8 +580,7 @@ class EmojiPickerListEntry extends StatelessWidget {
       child: Padding(
         padding: const EdgeInsets.symmetric(horizontal: 8),
         child: Row(spacing: 4, children: [
-          if (glyph != null)
-            Padding(
+          Padding(
               padding: const EdgeInsets.all(10),
               child: glyph),
```
chrisbobbe: This adds horizontal padding (`const EdgeInsets.all(10)`). We don't want to do that, as I mentioned in https://github.com/zulip/zulip-flutter/pull/1759#pullrequestreview-3076631735.

I suggest keeping `null` for `glyph` in the text-emoji case, and adding an `else` here that makes a `SizedBox(height: _emojiSize + 10 + 10)`. Similarly in the other case.

lib/widgets/autocomplete.dart
loveucifer: i was trying to write better commit messages using llms since mine were not upto mark , i will rebase and try to write a better one

lib/widgets/emoji_reaction.dart
loveucifer: okay will work on that

lib/widgets/autocomplete.dart
chrisbobbe: Please read this message: [#general > using LLMs in Zulip work @ 💬](https://chat.zulip.org/#narrow/channel/2-general/topic/using.20LLMs.20in.20Zulip.20work/near/2046012)

loveucifer: will try to keep the use to a minimum and try to polish everything to a point I can , also sorry for keep messing this up and having you review this all the time

loveucifer:
@chrisbobbe could you help me out in this thread--https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/merge.20conflicts.20help

Issue #792, opened 2024-07-06
Offer "Copy link to topic"


gnprice:
We should offer a "Copy link to topic" option.

This will be part of the topic action sheet. At the moment we don't have a topic action sheet; this can be the first item in it, or this issue can come after one of the other open issues for options to add to the topic action sheet:
* #348 
* #744

In particular see #348 for a description of where the topic action sheet should appear.

When we do offer this option, the links it generates should be topic permalinks, when the server supports it (it's part of the upcoming Zulip Server 9.0) and once we're ready for Zulip clients in general to be generating topic permalinks. In other words, the zulip-flutter implementation of this feature should match this zulip-mobile issue:
* https://github.com/zulip/zulip-mobile/issues/5876

and the eventual version (i.e. the version that gets merged, after any revisions) of the web side of this PR:
* https://github.com/zulip/zulip/pull/30644

lakshya1goel:
Hi, I’m interested in working on the "Copy link to topic" feature for the topic action sheet. Before proceeding with the UI and integration, I wanted to confirm whether Zulip Server 9.0, which supports topic permalinks, is already available or if there’s a specific timeline for its release.

Since the feature involves generating topic links as permalinks, I’d like to ensure the server is ready to handle this before implementing it on the Flutter side.

Looking forward to any updates!
Thanks!

gnprice:
Yes, if you do a web search for "zulip server 9.0", you'll find the announcement:
https://blog.zulip.com/2024/07/25/zulip-9-0-released/

In any case a PR for this issue will need to support generating both permalinks and the older non-permalinks. As I said above, we'll want to generate permalinks only when the server supports it; and not all servers have upgraded to 9.0. See our README at https://github.com/zulip/zulip-flutter#server-compatibility, and see the existing code for examples of how we handle differences between server versions.

chrisbobbe:
When this is done we should update the Help Center doc. We removed instructions for the legacy app when preparing to launch the new app: zulip/zulip@3a626036d1d2da8c0955516f43b4a90ea46a957e

gnprice:
This was requested yesterday [in the Rust community](https://rust-lang.zulipchat.com/#narrow/channel/122653-zulip/topic/new.20mobile.20app.20feedback/near/527993981).

PR #1762, opened 2025-07-28
"Copy link to topic" in topic action sheet

sm-sayedi:
### Screenshots
| Topic action sheet | Copied link pasted |
| --- | --- |
| Copy link to topic | Topic link pasted

### Screen recording

https://github.com/user-attachments/assets/59d864bb-169e-497f-9722-46f7d7d52359

Fixes: #792

### DIFF
--- a/lib/widgets/action_sheet.dart
+++ b/lib/widgets/action_sheet.dart
@@ -389,15 +389,9 @@ void showTopicActionSheet(BuildContext context, {
       pageContext: context));
   }
 
-  if (optionButtons.isEmpty) {
-    // TODO(a11y): This case makes a no-op gesture handler; as a consequence,
-    //   we're presenting some UI (to people who use screen-reader software) as
-    //   though it offers a gesture interaction that it doesn't meaningfully
-    //   offer, which is confusing. The solution here is probably to remove this
-    //   is-empty case by having at least one button that's always present,
-    //   such as "copy link to topic".
-    return;
-  }
+  optionButtons.add(CopyTopicLinkButton(
+    narrow: TopicNarrow(channelId, topic, with_: someMessageIdInTopic),
+    pageContext: context));
 
   _showActionSheet(pageContext, optionButtons: optionButtons);
 }
@@ -615,6 +609,32 @@ class MarkTopicAsReadButton extends ActionSheetMenuItemButton {
 
   @override void onPressed() async {
     await ZulipAction.markNarrowAsRead(pageContext, TopicNarrow(channelId, topic));
+  }
+}
+
+class CopyTopicLinkButton extends ActionSheetMenuItemButton {
+  const CopyTopicLinkButton({
+    super.key,
+    required this.narrow,
+    required super.pageContext,
+  });
+
+  final TopicNarrow narrow;
+
+  @override IconData get icon => ZulipIcons.link;
+
+  @override
+  String label(ZulipLocalizations localizations) {
+    return localizations.actionSheetOptionCopyTopicLink;
+  }
+
+  @override void onPressed() async {
+    final localizations = ZulipLocalizations.of(pageContext);
+    final store = PerAccountStoreWidget.of(pageContext);
+
+    PlatformActions.copyWithPopup(context: pageContext,
+      successContent: Text(localizations.successTopicLinkCopied),
+      data: ClipboardData(text: narrowLink(store, narrow).toString()));
   }
 }
 
--- a/test/widgets/action_sheet_test.dart
+++ b/test/widgets/action_sheet_test.dart
@@ -493,6 +493,7 @@ void main() {
 
         checkButton('Follow topic');
         checkButton('Mark as resolved');
+        checkButton('Copy link to topic');
       }
 
       testWidgets('show from inbox; message in Unreads but not in MessageStore', (tester) async {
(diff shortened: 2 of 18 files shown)


lib/widgets/action_sheet.dart
```diff
@@ -587,6 +583,32 @@ class MarkTopicAsReadButton extends ActionSheetMenuItemButton {
...
+class CopyTopicLinkButton extends ActionSheetMenuItemButton {
+  const CopyTopicLinkButton({
+    super.key,
+    required this.narrow,
+    required super.pageContext,
+  });
+
+  final TopicNarrow narrow;
+
+  @override IconData get icon => Icons.link;
```
chrisbobbe: I notice that this is `Icons.link` instead of `ZulipIcons.link`. Do we have a custom icon we can use here?

…Ah I think we do. Here's the SVG: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=544-22131&p=f&m=dev

I see it being used for "Copy link to channel" in Figma: https://www.figma.com/design/1JTNtYo9memgW7vV6d0ygq/Zulip-Mobile?node-id=6326-96513&m=dev

Would you add a prep commit that adds that new custom icon, and another commit that changes the one existing `Icons.link` (in "Copy link to message") to use the new icon?

chrisbobbe (commented):
Thanks! This looks good; one small comment on the icon in the action sheet.

For the tests, let's include a test for the pre-271 behavior of omitting "/with/" (see `resolveApiNarrowForServer`).

alya:
Would it be a follow-up to do the equivalent of https://github.com/zulip/zulip/issues/29136?

sm-sayedi:
Thanks @chrisbobbe for the review! New revision pushed, with tests included. Also added the new "link" icon, with screenshots updated. PTAL.

sm-sayedi:
> Would it be a follow-up to do the equivalent of [zulip/zulip#29136](https://github.com/zulip/zulip/issues/29136)?

Yeah, I think doing that as a follow-up would be the better option and will make it easy to get this PR merged sooner.

lib/widgets/action_sheet.dart
```diff
@@ -358,15 +358,11 @@ void showTopicActionSheet(BuildContext context, {
...
 
-  if (optionButtons.isEmpty) {
-    // TODO(a11y): This case makes a no-op gesture handler; as a consequence,
-    //   we're presenting some UI (to people who use screen-reader software) as
-    //   though it offers a gesture interaction that it doesn't meaningfully
-    //   offer, which is confusing. The solution here is probably to remove this
-    //   is-empty case by having at least one button that's always present,
-    //   such as "copy link to topic".
-    return;
-  }
```
chrisbobbe, line 400: 🎉

lib/widgets/action_sheet.dart
```diff
@@ -358,15 +358,11 @@ void showTopicActionSheet(BuildContext context, {
...
+  optionButtons.add(CopyTopicLinkButton(
+    narrow: TopicNarrow(channelId, topic,
+      // TODO(#1499): use the/a max message id of the topic
```
chrisbobbe: Is there still a TODO here? A previous commit started using `lastOrNull` of a list of message IDs.

sm-sayedi:
@chrisbobbe Pushed a new revision with the small changes you suggested in today's call.

test/widgets/action_sheet_test.dart
```diff
@@ -861,6 +862,51 @@ void main() {
...
+      });
+
+      Future<void> tapCopyTopicLinkButton(WidgetTester tester) async {
+        await tester.ensureVisible(find.byIcon(ZulipIcons.link, skipOffstage: false));
+        await tester.tap(find.byIcon(ZulipIcons.link));
+        await tester.pump(); // [MenuItemButton.onPressed] called in a post-frame callback: flutter/flutter@e4a39fa2e
+      }
+
+      group('copies topic link to clipboard', () {
+        testWidgets('FL >= 271 -> link contains "with" operator', (tester) async {
```
chrisbobbe: nit: When there's a pair of tests for "legacy" vs. "current" behavior, I like to consider the "current" behavior as boring/default/unmarked, so for example naming the test like

```dart
testWidgets('copies message link to clipboard', (tester) async {
```

, and consider the "legacy" one as interesting/unusual/marked:

```dart
testWidgets('FL < 271 -> link doesn\'t contain "with" operator', (tester) async {
```


In the future, when we stop supporting servers with the legacy behavior,  "message link" will always include "with" in its meaning, and when we clear out the backwards-compatibility code and its corresponding test, the remaining test won't need any changes to make it look natural.

test/widgets/action_sheet_test.dart
```diff
@@ -861,6 +862,51 @@ void main() {
...
+          await prepare(channel: someChannel, topic: someTopic,
+            zulipFeatureLevel: eg.recentZulipFeatureLevel);
+          await showFromAppBar(tester, channel: someChannel,
+            topic: TopicName(someTopic), messages: [message]);
+
+          await tapCopyTopicLinkButton(tester);
+          await tester.pump(Duration.zero);
+          final expectedLink = narrowLink(store,
+            TopicNarrow(someChannel.streamId, TopicName(someTopic), with_: message.id)
+          ).toString();
```
chrisbobbe: How about a quick `check(expectedLink.toString().contains('with')).isTrue()`, and `isFalse` in the "FL < 271" test? Just to satisfy a reader that `narrowLink` handles the feature-level condition correctly and they don't need to check that detail in its implementation or tests.

(No need for a TODO to remove that check in the default/current test when clearing out backwards-compatibility code; it won't be as necessary in that future, but it's just one line so the cost is pretty minimal.)

chrisbobbe (commented):
Thanks, this looks great! Just small comments below, and I see you've already updated the `lastOrNull` commit as discussed on our call today.

lib/widgets/action_sheet.dart
sm-sayedi: Yeah, I think we will still need that TODO as the previous commit partially addressed the "max message id" case and did it only for when the topic action sheet is opened through the app bar title (it's still partial becuase the last message of topic list view is still not gauranteed to be the most recent message of topic). Then for other places where the sheet is opened from, like through the recipient header or in inbox view (which currently don't pass the max message id), the best option will be to get the most recent message id from a central place.

sm-sayedi:
Thanks @chrisbobbe for the review. New changes pushed.

test/widgets/action_sheet_test.dart
```diff
@@ -888,6 +889,51 @@ void main() {
...
+          zulipFeatureLevel: eg.recentZulipFeatureLevel);
+        await showFromAppBar(tester, channel: someChannel,
+          topic: TopicName(someTopic), messages: [message]);
+
+        await tapCopyTopicLinkButton(tester);
+        await tester.pump(Duration.zero);
+        final expectedLink = narrowLink(store,
+          TopicNarrow(someChannel.streamId, TopicName(someTopic), with_: message.id)
+        ).toString();
+        check(expectedLink.toString().contains('/with/')).isTrue();
```
gnprice: nit: redundant toString

I'd probably remove the first one, and let expectedLink be a `Uri` object.

test/widgets/action_sheet_test.dart
```diff
@@ -888,6 +889,51 @@ void main() {
...
+        check((await Clipboard.getData('text/plain'))!).text.equals(expectedLink);
+      });
+
+      testWidgets('FL < 271 -> link doesn\'t contain "with" operator', (tester) async {
+        final message = eg.streamMessage(stream: someChannel, topic: someTopic);
+        await prepare(channel: someChannel, topic: someTopic,
+          zulipFeatureLevel: 270);
+        await showFromRecipientHeader(tester, message: message);
```
gnprice: Is there a reason this test uses showFromRecipientHeader while its counterpart above uses showFromAppBar?

Because these two are meant to demonstrate a contrast in the behavior depending on the feature level, it's best if their setup is all the same except for the feature level.

lib/widgets/action_sheet.dart
gnprice: I think no further change is actually needed in this direction.
* The recipient header uses the message ID of the message directly under it. When that's how the user got to the topic action sheet, having the /with/ operand point to that message seems fine and good — it means the link will continue to link to the conversation that that message belongs to, if things get split up.
* The inbox uses the latest unread message ID in the conversation, right? When there are any unreads, the latest unread will almost always be the latest message. In the unusual situations where it isn't, and if the latest (but not unread) message later gets split to a different conversation from the latest unread, I'm not sure which one would be better for the link to point to.

sm-sayedi:
Thanks @gnprice for the review. New revision pushed. PTAL.

gnprice:
Thanks! Looks good; merging.

PR #1755, opened 2025-07-26
deps: Upgrade Flutter to 3.33.0-1.0.pre.1149

sm-sayedi:
And update Flutter's supporting libraries to match.

With this upgrade, we can now use the route transition duration in tests to wait the minimum amount of time needed for the route animation to complete.

Related upstream PR: https://github.com/flutter/flutter/pull/171109

PR #1763, opened 2025-07-28
deps: Upgrade Flutter to 3.33.0-1.0.pre-1085

gnprice:

In particular this gets us TransitionDurationObserver:
  https://github.com/flutter/flutter/pull/171109

We stop just before a certain Dart SDK roll:
  https://github.com/flutter/flutter/commit/fa80cbcbdbdd
because that commit breaks completion in the IDE, which would be pretty annoying for doing development.  Thanks to Sayed for spotting that breakage and bisecting it to that commit:
  https://github.com/zulip/zulip-flutter/pull/1755#issuecomment-3124335590
  https://chat.zulip.org/#narrow/channel/516-mobile-dev-help/topic/New.20Flutter.20upgrade.20degrades.20code.20autocompletion/near/2230721

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -714,10 +714,10 @@ packages:
     dependency: transitive
     description:
       name: meta
-      sha256: e3641ec5d63ebf0d9b41bd43201a66e3fc79a65db5f61fc181f04cd27aab950c
+      sha256: "23f08335362185a5ea2ad3a4e597f1375e78bce8a040df5c600c8d3552ef2394"
       url: "https://pub.dev"
     source: hosted
-    version: "1.16.0"
+    version: "1.17.0"
   mime:
     dependency: "direct main"
     description:
@@ -1355,5 +1355,5 @@ packages:
     source: path
     version: "0.0.1"
 sdks:
-  dart: ">=3.9.0-293.0.dev <4.0.0"
-  flutter: ">=3.33.0-1.0.pre.832"
+  dart: ">=3.10.0-15.0.dev <4.0.0"
+  flutter: ">=3.33.0-1.0.pre-1085"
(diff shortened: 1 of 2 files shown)


gnprice:
CI failed in our `flutter_version` check:
```
Running flutter_version...
error: Flutter commit in pubspec.yaml seems to differ from version bound
Commit b46957fb06ffbd8590caa8590f1a0b06a71fc609 was described as: 3.33.0-0.0.pre-1085-gb46957fb
We therefore expect the Flutter version name to be: 3.33.0-1.0.pre.1085
But the Flutter version bound in pubspec.yaml is: 3.33.0-1.0.pre-1085
```

The dash is how `flutter --version` now formats the version, though:
```
$ flutter --version | cat
Flutter 3.33.0-1.0.pre-1085 • channel main • https://github.com/flutter/flutter.git
Framework • revision b46957fb06 (5 days ago) • 2025-07-23 13:35:08 -0400
Engine • hash 9aa877987921faf059f2bad643175ba79b1dc283 (revision 1f65ef4e60) (5 days ago) • 2025-07-23 17:21:09.000Z
Tools • Dart 3.10.0 (build 3.10.0-15.0.dev) • DevTools 2.48.0
```

So the right fix might be to update that bit of `tools/check` to expect the new format.

I'll first spend a few minutes trying to pin down how and where the format changed, though, mostly to validate that this seems like an intentional change which we can expect to persist.

gnprice:
Aha, yep: https://github.com/flutter/flutter/pull/172141

    [ Tool ] Fix `flutter upgrade` stating that an upgrade is available on `main` when up to date (#172141)
    
    The `frameworkVersion` string written to the version files wasn't
    actually parsable by `GitTagVersion` as it didn't match the format
    output by `git`.
    
    This change updates the `frameworkVersion` format to use a `-` instead
    of a `.` before the commit count and adds support to the version parsing
    regex to handle both `.` and `-` separators before the commit count.
    
    Fixes https://github.com/flutter/flutter/issues/172091

Seems reasonable, and a reason that's likely to persist.

I'll probably have `tools/check` accept either format, though, in case it changes back.

chrisbobbe:
Thanks, LGTM! Please merge at will.

PR #1768, opened 2025-07-29
Clear out some backwards-compat code; inline store.userDisplayEmail

chrisbobbe:


gnprice:
Thanks for the cleanup! Looks good; merging, with one commit-message tweak:

      ## Commit message ##
    -    store [nfc]: Inline store.userDisplayEmail, which has become trivial
    +    user [nfc]: Inline store.userDisplayEmail, which has become trivial

For changes that belong to a particular substore, I've been using the substore for the summary-line prefix to make it a bit more specific.

chrisbobbe:
Great, thanks!

